PR #61815, opened 2024-05-21
Inspector summary rows: Make tooltips appear middle-left

jameskoster:
## What?
Update the Buttons in Inspector summary rows so that tooltips appear to the left of the button.

## Why?
There exists a small usability issue with the current placement. As you cursor up and down the list of buttons, the tooltips can annoyingly obscure the other buttons. Video demo:

https://github.com/WordPress/gutenberg/assets/846565/c258245f-f85b-4a41-b0d6-e24cfb84ae6d

Note that when the date tooltip is visible the template button becomes inaccessible until you cursor away.

## Result
The tooltips now appear left of the button so that you can cursor up and down the list without distraction.

Screenshot 2024-05-21 at 10 08 55

## Testing Instructions
1. Edit a post or page
2. Cursor over buttons in the summary panel
3. Ensure all tooltips appear middle-left

### DIFF
--- a/packages/editor/src/components/post-schedule/panel.js
+++ b/packages/editor/src/components/post-schedule/panel.js
@@ -72,6 +72,7 @@ export default function PostSchedulePanel() {
 							size="compact"
 							className="editor-post-schedule__dialog-toggle"
 							variant="tertiary"
+							tooltipPosition="middle left"
 							onClick={ onToggle }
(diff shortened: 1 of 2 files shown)


youknowriad (approved):
LGTM 👍

jasmussen (approved):
Good use of the prop, given it exists. Nice attention to detail.

PR #61629, opened 2024-05-13
E2E: Ensure SE canvas loader appears before waiting for it to disappear

WunderBart:
## What?
When the Site Editor takes longer to load (e.g., due to large canvas content), the `visitSiteEditor` promise resolves prematurely as it incorrectly assumes the canvas loader has already disappeared. In reality, it hasn't even appeared yet. It can lead to, e.g., subsequent locator assertions timing out as their 10s timeout gets eaten up by the canvas loading time. This PR fixes that by attempting to wait for the loader element immediately after the page is loaded so that we can reliably wait for it to disappear.

**Failure example**: [trace.zip](https://github.com/WordPress/gutenberg/files/15306958/trace.zip) (from WooCommerce Blocks)

## Testing Instructions

All tests should pass.

PR #61816, opened 2024-05-21
E2E: Fix canvas waiter in visitSiteEditor

WunderBart:
Reported in: https://github.com/WordPress/gutenberg/pull/61629#issuecomment-2120893082
Related convo: https://wordpress.slack.com/archives/C02QB2JS7/p1716226087226579

## What?
Fix the canvas loading bar waiter step so that it only applies when we open a page for editing, e.g., by passing the `pageId` param.

## Why?
The [current implementation](https://github.com/WordPress/gutenberg/pull/61629) adds 2 minutes to each performance test that visits the site editor's page that doesn't load the canvas, e.g., the `All Templates` view. It also does that for E2E tests, only there the `actionTimeout` is 5 seconds instead of 2 minutes, so we don't really notice that.

## How?
By checking if the `pageId` or the `canvas` query params are present, so that we know the intent is to actually edit content.

## Testing Instructions
All E2E tests should pass, and the total time of the performance tests should be reduced from ~50 to ~30 minutes.

### DIFF
--- a/packages/e2e-test-utils-playwright/src/admin/visit-site-editor.ts
+++ b/packages/e2e-test-utils-playwright/src/admin/visit-site-editor.ts
@@ -45,7 +45,7 @@ export async function visitSiteEditor(
 	 * content underneath the loading overlay should be marked inert until the
 	 * loading is done.
 	 */
-	if ( postId || canvas === 'edit' ) {
+	if ( ! query.size || postId || canvas === 'edit' ) {
 		const canvasLoader = this.page.locator(
 			// Spinner was used instead of the progress bar in an earlier
 			// version of the site editor.
--- a/packages/e2e-test-utils-playwright/src/admin/visit-site-editor.ts
+++ b/packages/e2e-test-utils-playwright/src/admin/visit-site-editor.ts
@@ -37,17 +37,7 @@ export async function visitSiteEditor(
 		query.set( 'canvas', canvas );
 	}
 
-	const canvasLoader = this.page.locator(
-		// Spinner was used instead of the progress bar in an earlier version of
-		// the site editor.
-		'.edit-site-canvas-loader, .edit-site-canvas-spinner'
-	);
-
 	await this.visitAdminPage( 'site-editor.php', query.toString() );
-
-	// Try waiting for the canvas loader to appear first, so that the locator
-	// that waits for it to disappear doesn't resolve prematurely.
-	await canvasLoader.waitFor().catch( () => {} );
 
 	/**
 	 * @todo This is a workaround for the fact that the editor canvas is seen as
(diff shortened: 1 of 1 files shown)


ellatrix (approved):
Great, last run was again 34mins 👍

WunderBart:
Retriggering the whole thing, [something's off](https://github.com/WordPress/gutenberg/actions/runs/9173858853/job/25225334888?pr=61816) with the E2E runner (likely unrelated to this PR). 🤞

packages/e2e-test-utils-playwright/src/admin/visit-site-editor.ts
```diff
@@ -37,31 +37,32 @@ export async function visitSiteEditor(
...
+	if ( ! query.size || postId || canvas === 'edit' ) {
+		const canvasLoader = this.page.locator(
+			// Spinner was used instead of the progress bar in an earlier
+			// version of the site editor.
+			'.edit-site-canvas-loader, .edit-site-canvas-spinner'
+		);
+
+		// Wait for the canvas loader to appear first, so that the locator that
+		// waits for the hidden state doesn't resolve prematurely.
+		await canvasLoader.waitFor( { state: 'visible' } );
```
ellatrix, line 57: Just wondering: why did it previously pass the check for the existence of the canvas loader, and then continue on to never disappear?

WunderBart: It only passed the check because [there was an empty `catch`](https://github.com/WordPress/gutenberg/pull/61629#discussion_r1599825668) for the scenarios in which the loader disappeared before the test reached that line. It was a premature optimization on my end – we didn't need it, and it only made things worse by swallowing the error that would prevent the original PR from being merged. 😅

packages/e2e-test-utils-playwright/src/admin/visit-site-editor.ts
```diff
@@ -37,31 +37,32 @@ export async function visitSiteEditor(
...
 	 * loading is done.
 	 */
-	await canvasLoader.waitFor( {
-		state: 'hidden',
-		// Bigger timeout is needed for larger entities, like the Large Post
-		// HTML fixture that we load for performance tests, which often doesn't
-		// make it under the default timeout value.
-		timeout: 60_000,
-	} );
+	if ( ! query.size || postId || canvas === 'edit' ) {
```
youknowriad, line 48: I think this check is probably a bit fragile and not something that will stay consistent over time. A better check would be to check if the "preview area" exist in the DOM or something like that.

WunderBart: I agree, although checking if something exists is what we had before this PR, and it didn't work well. We were checking whether the loader was there, and if not, we were swallowing the `waitFor()` promise error and continuing with the test. The downside of such a solution is that whenever the loader element is not there, we add the current `actionTimeout` value to the test (before the error is swallowed), which is 10s for E2E tests and 120s for perf tests. We could explicitly fix that timeout for that `waitFor()` call to, e.g., 5s, but it would still add 5s for no-loader situations and wouldn't guarantee stability. 

IMO, the most stable solution for these situations is to use `locator.or` or parallel locator waits with `Promise.any`, which immediately returns the element first on the page. We need to know which element is there when the canvas is not there and vice-versa, though. For example:

```js
await Promise.any([
  page.locator('.canvas-loader').waitFor(),
  page.locator('.some-panel-instead-of-canvas').waitFor(),
]);
```

I couldn't quickly trace an element like that when making this fix. Do you know any we could use here?

youknowriad: I can't think of a "or" condition that wouldn't break the expected behavior :P

PR #61817, opened 2024-05-21
Block Bindings: Move logic to core functions instead of using a hook

SantosGuillamot:
## What?
Move the bindings logic to get and update the value of the connected source to `getBlockAttributes` and `updateBlockAttributes` instead of having a custom hook.

## Why?
This is part of core and it is where it belongs. Apart from that, it ensure it works as expected in all the uses cases and not only when getting the attributes from the props and updating them with `setAttributes`.

## How?
* Remove the old hook implementation.
* As part of that, I moved a couple of utils to a new file and changed the imports of the necessary files.
* To change the attribute with the value returned by the source with `getValue`, hook into the `getBlockAttribute` selector to read the bindings.
* To allow editing of the source value through `setValue`, instead of updating the attribute itself, hook into the `updateBlockAttributes` action.

This last step requires changing the implementation to use `dispatch`, so I had to modify the e2e tests.

## Testing Instructions
e2e tests passing should be enough. If you want to test it manually you can:

<details>

```php

	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
		)
	);
}
```

</details>

<details>

<summary>2. Add a paragraph block bound to the custom field using the Code Editor</summary>

```
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Check that the value of the custom field is shown.


ellatrix:
Yes, so far this is what I had in mind :)

SantosGuillamot:
It seems there are (at least) two problems with the current refactoring:

## 1. Pattern inner blocks are not updated properly when parent block attributes change

When I modify a pattern value and I click on Reset, the attribute "content" of the parent is changed correctly, but the inner blocks are not refreshed. I made a quick video to show what I mean:

https://github.com/WordPress/gutenberg/assets/34552881/00bb2205-4b4f-4e52-b2e3-20b58cce633b

## 2. Tests with blocks with bindings using `insertBlock` update the value before being inserted

When a block attribute is connected to a binding, we don't actually modify the values, we just show it in the editor. This means that if the paragraph content is connected to a custom field, the editor shows the value of the custom field. But if you go to the code editor, you can still see the original value.

However, in e2e tests, if we use `insertBlock` like in [this case](https://github.com/WordPress/gutenberg/blob/0e91129ee1b9fa7cf34db6567b9dd70e1d6599c6/test/e2e/specs/editor/various/block-bindings.spec.js#L1217-L1231), it seems that it actually modifies the value of the block attribute. This means that instead of `<p>Fallback value</p>` it has `<p>Value of the custom field</p>`.

---
I will keep taking a look, but any guidance is welcome 🙂

packages/editor/src/bindings/pattern-overrides.js
```diff
@@ -9,18 +9,18 @@ const CONTENT = 'content';
...
 	name: 'core/pattern-overrides',
 	label: _x( 'Pattern Overrides', 'block bindings source' ),
-	getValue( { registry, clientId, attributeName } ) {
-		const { getBlockAttributes, getBlockParentsByBlockName } =
-			registry.select( blockEditorStore );
-		const currentBlockAttributes = getBlockAttributes( clientId );
+	getValue( { select, clientId, attributeName } ) {
+		const { getBlock, getBlockParentsByBlockName } =
+			select( blockEditorStore );
+		const { attributes: currentBlockAttributes } = getBlock( clientId );
```
ellatrix: Why use `getBlock`? That does not contain the attributes right? We'll somehow have to make sure it's not recursive, or pass the attributes to this function.

packages/block-editor/src/store/selectors.js
```diff
@@ -114,19 +115,76 @@ export function isBlockValid( state, clientId ) {
...
+		) ) {
+			const source = sources[ boundAttribute.source ];
+			if (
+				! source?.getValue ||
+				! canBindAttribute( block.name, attributeName )
+			) {
+				continue;
+			}
+
+			const context = {};
```
ellatrix: This is still a to do, right?

packages/editor/src/bindings/pattern-overrides.js
SantosGuillamot: > Why use getBlock? That does not contain the attributes right?

From my tests, it seems it actually includes the attributes. I decided to use that because calling `getBlockAttributes` causes an infinite loop. Maybe it is better to pass the attributes as an argument or prevent the loop, as you suggest. Any preference?

packages/block-editor/src/store/selectors.js
SantosGuillamot: Yes. I added a TODO comment at [this point](https://github.com/WordPress/gutenberg/pull/61817/files/0e91129ee1b9fa7cf34db6567b9dd70e1d6599c6#diff-aa92c48095c35f17feb834dd919baf52274a73c7b1180ff5acce7b6087b126cfR127-R128).

packages/editor/src/bindings/pattern-overrides.js
ellatrix: I think somehow we'll have to make `getBlockAttributes` work, someone might decide to use it and not know it's not allowed. Also, we might have to filter `getBlock` as well at some point?

ellatrix:
I think number (2) was the reason for not doing this earlier. This approach also "modifies" the original attribute value.

packages/editor/src/bindings/pattern-overrides.js
SantosGuillamot: Yes, I think you're right. The simplest solution that come to my mind is having an external `isProcessingBindings` variable which by default is `false`, it is set to `true` when bindings processing starts, and it is set to `false` when bindings processing finishes. Could something like that be enough? I quickly tested it and it seems to work but I don't know if there are any implications.

> we might have to filter getBlock as well at some point?

Mmm, I guess so.

SantosGuillamot:
> I think number (2) was the reason for not doing this earlier. This approach also "modifies" the original attribute value.

It doesn't seem to be an issue for existing blocks with bindings or when bindings are added manually, but it seems it is broken when using `insertBlock` because `getBlockAttributes` runs and change the attributes before that. Any idea how we can overcome that limitation? I've been talking a look at the code and I am not sure how to approach it. I thought of waiting somehow until the block is inserted to update the bindings, but I am not convinced about that.

SantosGuillamot:
A quick summary of the challenges we need to solve before this pull request could be merged:

## 1. How to access the block context

Right now, we are passing the block context needed by the sources in the existing hook. However, it isn't available in this new implementation while using the store. In [this Slack conversation](https://wordpress.slack.com/archives/C02QB2JS7/p1716285854418259?thread_ts=1715935369.206469&cid=C02QB2JS7), Ella suggested syncing the context to the store, but it needs to be discussed if that's the right approach.

EDIT: We started exploring this in [this pull request](https://github.com/WordPress/gutenberg/pull/61895).

## 2. Pattern inner blocks are not updated properly when parent block attributes change

When I modify a pattern value and I click on Reset, the attribute "content" of the parent is changed correctly, but the inner blocks are not refreshed. I made a quick video to show what I mean:

https://github.com/WordPress/gutenberg/assets/34552881/00bb2205-4b4f-4e52-b2e3-20b58cce633b

## 3. Inserting blocks with bindings with `insertBlock` doesn't keep the original attribute value

While it is not an issue while adding the bindings manually, inserting blocks programmatically remove the original value because the `getBlockAttributes` logic to replace it with the bindings run before it is inserted.

packages/editor/src/bindings/pattern-overrides.js
SantosGuillamot: I've tried to make it work with `getBlockAttributes` with [this logic](https://github.com/WordPress/gutenberg/pull/61817/commits/ed11e1ad90c10a063a5facf8bea4b3f50caae79c). While it works, I am not sure if it is the best approach. Could something like that work or should I look for a better solution?

ellatrix:
It would be good to have a list of the cases where we do NOT want bindings in block attributes, and a list where we DO.

SantosGuillamot:
> It would be good to have a list of the cases where we do NOT want bindings in block attributes, and a list where we DO.

Until now, I would say that we wanted the bindings in all cases but in the save, because we didn't want to save the bindings value in the database. But I don't have a clear picture of all the uses cases where block attributes are used.

ellatrix:
> Until now, I would say that we wanted the bindings in all cases but in the save, because we didn't want to save the bindings value in the database. But I don't have a clear picture of all the uses cases where block attributes are used.

I guess that's something we really need to understand before continuing imo. Maybe in certain React contexts it shows the sourced value? Idk. Maybe we should also think about how Bits would work, it should probably be very similar?

SantosGuillamot:
I believe I made good progress and "solve" most of the issues to make it work as it is working right now, plus supporting functions like `getBlockAttributes`.

After rebasing the branch to [the one where we are syncing the context with the store](https://github.com/WordPress/gutenberg/pull/61895), I made some changes and fixed the bugs I encountered.

Let's wait to see if tests pass (they passed locally), but if so I guess the first step is to decide if the syncing context is the right approach.

> I guess that's something we really need to understand before continuing IMO. 

I agree we need to work on that. On the other hand, it is an issue with the current implementation as well because we are hooking in the edit. I feel this pull request would be more reliable and we can build on top of it.

gziolo:
This one needs to be revisited and potentially closed. My impression is that we have everything in place for Block Bindings to work efficiently on the client, but I don't remember exactly what was explored here, so I'm happy to discuss the opportunity.

SantosGuillamot:
We were exploring the possibility of directly returning the "bindings value" in `getBlockAttributes`, which would avoid creating a hook. However, I believe that's not always the wanted scenario. I feel we should probably keep `getBlockAttributes` as it is and explore adding a new function to get the "bindings value," at least until it is clearer in which use cases we want the original block attribute and in which ones we want the value of the binding.

I'm closing this pull request based on that.

Issue #61158, opened 2024-04-26
LinkControl: Provide better way to see the full URLs


afercia:
### Description

Splitting this out from https://github.com/WordPress/gutenberg/issues/60901 and https://github.com/WordPress/gutenberg/pull/60908

As mentioned in the linked issue and PR, using the 'Copy link' button tooltip to display the _full URL_ of the link is a no-go for accessibility and, I would say, also for usability.

> [!IMPORTANT]
> Please do not ever use values or states for a control's tooltip. A control's tooltip is meant to visually expose the control's name provided by its `aria-label`. A label must be used to communicate what the control does. It's not the place for values or states.

By using extraneous content for the aria-label and tooltip:
- Screen reader users will hear long, noisy, confusing information.
- Speech recognition software users will have troubles issuing a voice command, the voice command may nnot work at all.

I'd add that also visually, a very long URL displayed within a tooltip isn't a great user experience. A couple example screenshots:

![Screenshot 2024-04-24 at 14 02 47](https://github.com/WordPress/gutenberg/assets/1682452/d54807cc-322b-4b85-abd9-4e40c1467f8f)

![Screenshot 2024-04-24 at 14 29 55](https://github.com/WordPress/gutenberg/assets/1682452/75e876bd-f8a7-46d5-8021-7d04594e3ae0)

## Proposal
Displaying of long URLs is not a new problem in WordPress. It has been discussed across the year and addressed, for example, for the Permalink editing UI in the Classic editor.

As suggested in https://github.com/WordPress/gutenberg/pull/60908#issuecomment-2074885973 I'd like to propose to use the same displaying mechanism used in the Classic Permalink editing UI.

- Important information should be shown with visible text. A tooltip isn't a great option in the first place.
- The Classic Permalink editing UI uses some smart handling of long URLs to attempt to display only the relevant parts. See [the Core implementation](https://github.com/WordPress/wordpress-develop/blob/d898a25d1a37f493b7050b263cb9c5b29565792a/src/js/_enqueues/vendor/tinymce/plugins/wplink/plugin.js#L11-L51).
- The Link Preview popover width is currently limited to 350 pixels. I don't see a good reason to limit the width of this specific popover to such a small width. This popover may potentially contain _long content_. A wider width could help.

Screenshots to illustrate what Classic Editor does to display long URLs in the link 'preview':

[...]

PR #61819, opened 2024-05-21
LinkControl: refined the display of the link preview title and url when both are same

### LINKED ISSUE #60900: LinkControl: refine the display of the link preview title and url
### Description

Normally, the LinkControl 'link preview' displayes two lines:
1. the `displayTitle`, which can be a few things:
  - the document title of the fetched 'rich data' of the link destination page
  - or, the entered link text
  - or, the entered link URL
2. The `displayURL` which is the link URL, with the protocol part stripped out.

When the rich data info can not be fetched, the 'title' is, normally, the linked text. Screenshot:

normally

When the link text is the same as the `displayURL`, there is some logic in place to avoid to display two lines with the same identical information. That makes totally sense, and only one line is displayed. Screenshot:

Screenshot 2024-04-19 at 14 03 49

Without this logic, there would be two lines that would repeat the same, identical, information. Screenshot:

otherwise

So far, so good.

However, the logic in place doesn't take into account the case when the _link text_ contains the protocol. In this case, the preview displayes both lines and they contain the same information, the only difference is the displayed protocol. Screenshot:

Screenshot 2024-04-19 at 14 04 09

Visually and semantically, in this case the repeated information is redundant. It would be best to show only the `displayTitle`.

- Add a Paragraph that contains the text `test.org`.
- Select the text `test.org` and add a link to `test.org` without the protocol part. 
- Once the link is added, click the link in the Paragraph to open the link control preview.
- Observe the link preview displays only one line: `test.org`.
- Add a Paragraph with some text and paste `https://test.org` within the text. The pasted link should be automatically transformed to a link. If not, make sure to add a link to it with the value `https://test.org`. It doesn't matter whether the URL contains the protocol part.
- Once the link is added, click the link in the Paragraph to open the link control preview.
- Observe the link preview displays two lines: `https://test.org` and `test.org`.

_No response_

### Environment info

_No response_

Yes

Yes

amitraj2203:
## What?
Fixes: #60900

## Why?
When the link text contains the protocol, the preview displays both lines (text and link) and they contain the same information, the only difference is the displayed protocol. So instead of displaying it twice its better to show it only once.

## How?
This PR adds a check that weather the URL and Text is same after removing the protocol from the text.

## Testing Instructions
1. Add a Paragraph that contains the text `https://test.org`.
2. Select the text `https://test.org` and add a link to it `test.org` without the protocol part.
3. Once the link is added, click the link in the Paragraph to open the link control preview.
4. Observe the link preview displays only one line: test.org.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/77401999/a9d01146-bd23-42af-8f8f-9503d09dd395


packages/url/src/filter-url-for-display.js
```diff
@@ -53,3 +53,14 @@ export function filterURLForDisplay( url, maxLength = null ) {
...
+
+/**
+ * Removes the protocol from a URL.
+ *
+ * @param {string} url - The URL to remove the protocol from.
+ *
+ * @return {string} The URL without the protocol.
+ */
+export function removeProtocol( url ) {
+	return url.replace( /^(?:https?:)\/\/(?:www\.)?/, '' );
```
afercia: Question: why this regex is slightly different from [the one used in core](https://github.com/WordPress/wordpress-develop/blob/4b91a597b7a91d8ce099314a744fca5309dd0134/src/js/_enqueues/vendor/tinymce/plugins/wplink/plugin.js#L19)?

afercia (commented):
Only one small question about the regex.

packages/url/src/filter-url-for-display.js
tyxla: I'd expect this to use more or less the same regex as `isValidProtocol` or `getProtocol`.

packages/url/src/filter-url-for-display.js
```diff
@@ -53,3 +53,14 @@ export function filterURLForDisplay( url, maxLength = null ) {
...
 }
+
+/**
+ * Removes the protocol from a URL.
+ *
+ * @param {string} url - The URL to remove the protocol from.
+ *
+ * @return {string} The URL without the protocol.
+ */
+export function removeProtocol( url ) {
```
tyxla: As a new public API, this function should come with some unit tests.

packages/url/src/filter-url-for-display.js
tyxla: This won't work for protocols different than HTTP/HTTPS, like `tel:` or `file:`.

tyxla: We're also quite generous towards what we consider a valid protocol, see the implementation and unit tests for `getProtocol()` and `isValidProtocol()`

amitraj2203: Hi @afercia, I took the reference of the regex from [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/url/src/filter-url-for-display.js#L17) I don't know the exact reason why this is diferent from the one used in core

amitraj2203: Hi @tyxla, I have made the changes as per your suggestion. Please take a look at it as per your convenience 🙇

packages/url/src/filter-url-for-display.js
```diff
@@ -53,3 +53,19 @@ export function filterURLForDisplay( url, maxLength = null ) {
...
 }
+
+/**
+ * Removes domain prefixes from a URL.
+ *
+ * @param {string} url The URL to strip domain prefixes from.
+ *
+ * @return {string} The URL without domain prefixes.
+ */
+export function stripDomainPrefixes( url ) {
```
tyxla: I find it a bit confusing that this function promises to remove the domain prefix (`www.`) but it also removes the protocol. I wonder if we could think of a more suitable name?

packages/url/src/filter-url-for-display.js
```diff
@@ -53,3 +53,19 @@ export function filterURLForDisplay( url, maxLength = null ) {
...
+	if ( ! url ) {
+		return '';
+	}
+	return url
+		.replace( /^[a-z\-.\+]+[0-9]*:(\/\/)?/i, '' )
```
tyxla: If this is for display purposes and link preview, it feels odd that we remove the protocol and the prefix, but we don't remove the `#hash` portion of the URL. And I can say the same about the port, and the query args.

amitraj2203: Yes it makes sense to remove the fragment, query args and port too from the link. Just wanted to confirm, should I change the regex accordingly?

afercia: > If this is for display purposes and link preview, it feels odd that we remove the protocol and the prefix, but we don't remove the `#hash` portion of the URL. And I can say the same about the port, and the query args.

@tyxla I'd totally agree. Wouldn't that be better addressed in the scope of https://github.com/WordPress/gutenberg/issues/61158 though? Your expertise would greatly help there to provide new utility functions for displaying purposes. I feel like this PR is going a little out of scope, as the original intent here was only to avoid to show duplicate information in the link control preview.

tyxla: > Yes it makes sense to remove the fragment, query args and port too from the link. Just wanted to confirm, should I change the regex accordingly?

Yes, I think it's worth considering how to name that helper function though, in order to reflect everything that it does. Technically, we're talking about returning the `hostname` (see [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname)), so the function could be called `getHostname()` or something similar.

> @tyxla I'd totally agree. Wouldn't that be better addressed in the scope of https://github.com/WordPress/gutenberg/issues/61158 though? Your expertise would greatly help there to provide new utility functions for displaying purposes. I feel like this PR is going a little out of scope, as the original intent here was only to avoid to show duplicate information in the link control preview.

To be fair, I don't mind where it will be addressed, but if this PR introduces a new API to a package, we need to ensure that it makes sense and is a reasonable API, and I was suggesting ways to get there. I'm happy to help with reviewing an alternative fix and handling this separately. If you feel like this PR is going too far, feel free to limit the fix to the affected component, avoiding introducing a new API, and that should make it more straightforward. We can then introduce the new function in a separate PR, if necessary.

amitraj2203:
Hi @afercia @tyxla 👋 
Based on the above discussion, please let me know what changes should be made with this PR from my side.

Thanks!

tyxla:
> Hi @afercia @tyxla 👋 Based on the above discussion, please let me know what changes should be made with this PR from my side.
> 
> Thanks!

Thanks for your hard work @amitraj2203 🙌 

Introducing new public API functions to another package could take more time as we want to make sure to design them well. 

In the meantime, I don't think that's a necessity to fix this issue. So I suggested [above](https://github.com/WordPress/gutenberg/pull/61819#discussion_r1622271182), feel free to fix this inline in the component (by introducing a local, unexported helper function), and we can consider abstracting this to public API in another PR. That will help fix the issue at hand and not delay it further.

amitraj2203:
Thanks for having a look at it. I have made the changes and considered the regex which only removes the protocol and domain prefix `(www.)`

packages/url/src/filter-url-for-display.js
```diff
@@ -13,8 +13,14 @@
...
 export function filterURLForDisplay( url, maxLength = null ) {
+	if ( ! url ) {
+		return '';
+	}
+
 	// Remove protocol and www prefixes.
-	let filteredURL = url.replace( /^(?:https?:)\/\/(?:www\.)?/, '' );
+	let filteredURL = url
+		.replace( /^[a-z\-.\+]+[0-9]*:(\/\/)?/i, '' )
+		.replace( /^www\./i, '' );
```
tyxla, line 23: Looks like these changes can now be reverted and/or suggested in another PR?

packages/url/src/filter-url-for-display.js
```diff
@@ -13,8 +13,14 @@
  * @return {string} Displayed URL.
  */
 export function filterURLForDisplay( url, maxLength = null ) {
+	if ( ! url ) {
+		return '';
+	}
+
 	// Remove protocol and www prefixes.
-	let filteredURL = url.replace( /^(?:https?:)\/\/(?:www\.)?/, '' );
```
tyxla, line 17: Again, outside of this PR, but this regex could likely use some love either way, because it doesn't correspond to the standards right now, which according to [RFC 2396](http://www.ietf.org/rfc/rfc2396.txt), Appendix A, is:

```
scheme        = alpha *( alpha | digit | "+" | "-" | "." )
```

meaning that digits, plus, minus or dots can be everywhere in the scheme, except for the first character (which has to be an alphabetic one).

Not something to fix here, but definitely something we should consider improving at some point.

packages/block-editor/src/components/link-control/link-preview.js
```diff
@@ -59,6 +59,22 @@ export default function LinkPreview( {
...
+	/**
+	 * Filters the title for display. Removes the protocol and www prefix.
+	 *
+	 * @param {string} title - The title to be filtered.
+	 *
+	 * @return {string} The filtered title.
+	 */
+	function filterTitleForDisplay( title ) {
+		return title
+			.replace( /^[a-z\-.\+]+[0-9]*:(\/\/)?/i, '' )
```
tyxla: Worth adding a comment that this is derived from `@wordpress/url`'s `isValidProtocol()` utility.

packages/block-editor/src/components/link-control/link-preview.js
```diff
@@ -59,6 +59,22 @@ export default function LinkPreview( {
...
 		stripHTML( richData?.title || value?.title || displayURL );
 
+	/**
+	 * Filters the title for display. Removes the protocol and www prefix.
+	 *
+	 * @param {string} title - The title to be filtered.
+	 *
+	 * @return {string} The filtered title.
+	 */
+	function filterTitleForDisplay( title ) {
```
tyxla: Can we declare this function outside of the component, so it doesn't get redeclared on every render?

tyxla (commented):
Thanks for your hard work on it @amitraj2203 🙌 

I think there's some last-minute cleanup to do before moving forward with this one.

packages/url/src/filter-url-for-display.js
amitraj2203: Thanks for reviewing the PR 🙇.
I wanted to clarify that if we revert this change and keep the new regex [here](https://github.com/WordPress/gutenberg/pull/61819/files#diff-905d5565ba99e656d55d9a7ee9e9d1a2a7b876e0e50a700dafdae1b9249a8e3eR70-R72) in link-preview.js, it will only work for URLs with `http or https` as the protocol.

Since the [issue](https://github.com/WordPress/gutenberg/issues/60900) is to avoid showing duplicate information in the link control preview, this approach will not be effective. We need to use the same regex in both places where the check is performed to [determine if the URL and title are the same](https://github.com/WordPress/gutenberg/pull/61819/files#diff-905d5565ba99e656d55d9a7ee9e9d1a2a7b876e0e50a700dafdae1b9249a8e3eL115). (We can also use the old regex at both the places)

For example, these URLs will fail if we use different regex at both places:
image
image

tyxla: I understand, thanks for elaborating. 

If changes to the `filterURLForDisplay` are necessary, can we accompany them with unit tests accordingly? Ideally, unit tests that will cover the changes would fail if run on `trunk`.

packages/block-editor/src/components/link-control/link-preview.js
amitraj2203: Since this regex is added in `filterURLForDisplay()` so I have added this function name in the inline comment.

packages/url/src/filter-url-for-display.js
amitraj2203: The updated regex works with all different kinds of protocol instead of working only with `http or https` so I have added few more test cases. (The old test cases are already passing)

packages/url/src/test/index.js
```diff
@@ -1030,18 +1036,28 @@ describe( 'filterURLForDisplay', () => {
...
+		let url = filterURLForDisplay(
 			'http://www.wordpress.org/wp-content/uploads/myimage.jpg',
 			20
 		);
 		expect( url ).toBe( '…/uploads/myimage.jpg' );
+		url = filterURLForDisplay(
+			'file:///home/user/documents/project/reports/2024/myfile.pdf',
+			20
+		);
+		expect( url ).toBe( '…orts/2024/myfile.pdf' );
```
tyxla: This assertion shouldn't be necessary, since the truncation is being tested in the above assertion, and the protocol removal is being tested in another test.

packages/url/src/test/index.js
```diff
@@ -1030,18 +1036,28 @@ describe( 'filterURLForDisplay', () => {
...
+		let url = filterURLForDisplay(
 			'http://www.wordpress.org/wp-content/uploads/superlongtitlewithextension.jpeg',
 			20
 		);
 		expect( url ).toBe( 'superlongti…ion.jpeg' );
+		url = filterURLForDisplay(
+			'file:///home/user/documents/project/reports/2024/superlongtitlewithextension.pdf',
+			20
+		);
+		expect( url ).toBe( 'superlongtit…ion.pdf' );
```
tyxla: Same as above - this may be redundant.

packages/url/src/test/index.js
```diff
@@ -1058,11 +1074,16 @@ describe( 'filterURLForDisplay', () => {
...
+		let url = filterURLForDisplay(
 			'http://www.wordpress.org/wp-content/uploads/filename.2020.12.20.png',
 			20
 		);
 		expect( url ).toBe( 'filename.202….20.png' );
+		url = filterURLForDisplay(
+			'file:///home/user/documents/project/reports/2024/filename.2020.12.20.png',
+			20
+		);
+		expect( url ).toBe( 'filename.202….20.png' );
```
tyxla: Same as above.

packages/block-editor/src/components/link-control/link-preview.js
```diff
@@ -27,6 +27,20 @@ import { ViewerSlot } from './viewer-slot';
...
 
+/**
+ * Filters the title for display. Removes the protocol and www prefix.
+ *
+ * @param {string} title The title to be filtered.
+ *
+ * @return {string} The filtered title.
+ */
+function filterTitleForDisplay( title ) {
+	// Dervied from `filterURLForDisplay` in `@wordpress/url`.
```
tyxla:
```suggestion
	// Derived from `filterURLForDisplay` in `@wordpress/url`.
```

packages/block-editor/src/components/link-control/link-preview.js
```diff
@@ -59,6 +73,9 @@ export default function LinkPreview( {
 		! isEmptyURL &&
 		stripHTML( richData?.title || value?.title || displayURL );
 
+	const isUrlRedundant =
+		value?.url && filterTitleForDisplay( displayTitle ) === displayURL;
```
tyxla: I think this might have a couple of issues: 

1. If `value` doesn't have an `url` prop, shouldn't `isUrlRedundant` be `true`? Currently, it will be `false,` unless we invert the `value?.url` logic. 

2. If I understand correctly, the URL will be redundant if any of the two conditions is independently `false`, meaning we likely have to change `&&` to `||`.

Here's how the changed logic would look like:

```suggestion
		! value?.url || filterTitleForDisplay( displayTitle ) === displayURL;
```

Alternatively, feel free to keep the original logic, refraining from inverting it, if you feel that makes it extra confusing or less readable.

tyxla (commented):
I think this is really close, and left a last round of feedback. 

Let me know what you think @amitraj2203 and thanks again for working on this one!

packages/block-editor/src/components/link-control/link-preview.js
amitraj2203: @tyxla Thanks for your continuous review and valuable feedback on the PR 🙇. I agree and have updated it as per your suggestion.

packages/url/src/filter-url-for-display.js
```diff
@@ -13,8 +13,14 @@
  * @return {string} Displayed URL.
  */
 export function filterURLForDisplay( url, maxLength = null ) {
+	if ( ! url ) {
+		return '';
+	}
```
tyxla, line 18: Worth adding a test case for this one, too.

tyxla (approved):
This looks great, thanks @amitraj2203 🙌 

Just one last thing before we can ship this.

packages/url/src/filter-url-for-display.js
amitraj2203: Added the test case for it 🙇

tyxla (approved):
LGTM 👍 

Thanks again @amitraj2203 🙌 

I'll let @afercia ship this after confirming #60900 is resolved as expected.

afercia:
@amitraj2203 @tyxla I loved seeing your collaboration and perseverance on this PR ❤️ Thanks both.

I can't speak for the code part and I'm totally confident in your expertise. Thanks for adding a few tests. 
Quickly tested and it does solve the issue.

PR #61820, opened 2024-05-21
Data views: Update grid layout on small screens

jameskoster:
## What?
Reduce the minimum number of columns a grid can be arranged into to 1.

## Why?
With the addition of bulk actions (checkbox) and the lock icon for patterns, a 2 column layout doesn't work so well at small screen sizes.

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2024-05-21 at 14 14 00 | Screenshot 2024-05-21 at 14 14 17 |

## Testing Instructions
1. Open a data view e.g. Patterns and ensure Grid layout is selected
2. Resize window to <480px
3. See grid arranged into single column

Note: This PR also includes a small fix to stop checkboxes shrinking and breaking the layout at smaller screen sizes.

### DIFF
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -277,9 +277,13 @@
 
 .dataviews-view-grid {
 	margin-bottom: auto;
-	grid-template-columns: repeat(2, minmax(0, 1fr)) !important;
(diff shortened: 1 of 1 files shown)


jasmussen (approved):
We need some better responsive mixins at some point. But this works.

PR #61741, opened 2024-05-16
Site Editor: Remove the details pages

youknowriad:
Related to #59689 

## What?

This PR is kind of the latest step of the unification of the "details panels" and the "inspector panel". Now that all/most of these panels unification is done, we can actually remove the "details pages" entirely.

That said, it's not without challenges when it comes to "routing" and "urls": Pages in the site editor can be loaded using two URLs today:

 - postType=page&postId=someID  This URL loads the page with the "details" panel open
 - path=%2Fpage&postId=270 This URL loads the page without the "details" panel (in list view)
 
As we're removing the details panel step, we need to unify. We have to pick one URL and redirect the other. I decided to keep the former as it's more friendly. 

This also means a change in all places where we link to these pages...

That said, it brings a lot of consistency and simplicity to our pages IMO, so I think it's a good change anyway.

## Testing Instructions

This is a very impactful change in the site editor, we need to test all the flows properly. E2E tests should help us catch some but probably not all the potential regressions.

PR #61230, opened 2024-04-30
Site Editor: useLocation instead of window.location.href

jsnajdr:
Noticed this when reviewing #61174: some Site Editor components look at `window.location.href`, but they should use the more "abstract" `useLocation` hook provided by the router.

This also reveals an API gap in the `useHistory` hook: it should expose the non-reactive location, via a `history.getLocationWithParams()` method. We don't need the reactive location (triggering a rerender on location change) when it's used only in an event handler. I'll address this separately in some follow-up PR.

PR #61823, opened 2024-05-21
History: add getLocationWithParams method

jsnajdr:
Add a new method to `history` called `getLocationWithParams`. The method is not really new, I'm just moving the existing function of the same name from the `RouterProvider` to the `history` object.

The `RouterProvider` can now just use a very straightforward `useSyncExternalStore` without any complications.

The method is memoized using a `WeakMap` so that an identical input (`location`) produces an identical output (`locationWithParams`). In https://github.com/WordPress/gutenberg/pull/61741#discussion_r1608040443 @youknowriad writes about the weakmap implementation:

> initially when I tried using weak map and use the entire location object as a key, the component was re-rendering way too much (location didn't seem stable for some reason)

but I cannot confirm this. I put logpoints at the place where `locationWithParams` is recomputed, and didn't notice any extra rerenders that shouldn't be there.

I'm also updating many places where the `locationWithParams` value was used in a non-reactive way, in event handlers and callbacks. Here I'm migrating reactive `useLocation` calls to imperative `history.getLocationWithParams()` calls. That should save a lot of redundant rerenders. I first noticed this when working on #61230 and now I'm fixing it.

This overlaps with #61741 -- depending on which one is merged first, there will be conflicts. But I believe they will be easy to resolve.

### DIFF
--- a/packages/edit-site/src/components/add-new-pattern/index.js
+++ b/packages/edit-site/src/components/add-new-pattern/index.js
@@ -25,14 +25,13 @@ import {
 	TEMPLATE_PART_POST_TYPE,
 } from '../../utils/constants';
 
-const { useHistory, useLocation } = unlock( routerPrivateApis );
+const { useHistory } = unlock( routerPrivateApis );
 const { CreatePatternModal, useAddPatternCategory } = unlock(
 	editPatternsPrivateApis
 );
 
 export default function AddNewPattern() {
 	const history = useHistory();
-	const { params } = useLocation();
 	const [ showPatternModal, setShowPatternModal ] = useState( false );
 	const [ showTemplatePartModal, setShowTemplatePartModal ] =
 		useState( false );
@@ -141,16 +140,17 @@ export default function AddNewPattern() {
 						return;
 					}
 					try {
+						const {
+							params: { postType, categoryId },
+						} = history.getLocationWithParams();
 						let currentCategoryId;
 						// When we're not handling template parts, we should
 						// add or create the proper pattern category.
-						if ( params.postType !== TEMPLATE_PART_POST_TYPE ) {
+						if ( postType !== TEMPLATE_PART_POST_TYPE ) {
 							const currentCategory = categoryMap
 								.values()
-								.find(
-									( term ) => term.name === params.categoryId
-								);
-							if ( !! currentCategory ) {
+								.find( ( term ) => term.name === categoryId );
+							if ( currentCategory ) {
 								currentCategoryId =
 									currentCategory.id ||
 									( await findOrCreateTerm(
@@ -170,7 +170,7 @@ export default function AddNewPattern() {
 						// category.
 						if (
 							! currentCategoryId &&
-							params.categoryId !== 'my-patterns'
+							categoryId !== 'my-patterns'
 						) {
 							history.push( {
 								postType: PATTERN_TYPES.user,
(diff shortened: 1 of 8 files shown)


packages/edit-site/src/components/add-new-pattern/index.js
```diff
@@ -26,14 +26,13 @@ import {
...
 
-const { useHistory, useLocation } = unlock( routerPrivateApis );
+const { useHistory } = unlock( routerPrivateApis );
 const { CreatePatternModal, useAddPatternCategory } = unlock(
 	editPatternsPrivateApis
 );
 
 export default function AddNewPattern() {
 	const history = useHistory();
-	const { params } = useLocation();
```
youknowriad, line 35: I personally like the direct hook to retrieve the current route's params. Most routers have a similar hook. React Router has `useParams` for instance.

youknowriad:
I can reproduce the crash that forced me to avoid using the location object as cache in this PR too.

 - Open the site editor
 - Open "pages"
 - Select a random page 
 - Switch to "drafts" on the sidebar or "trash"
 - The editor explodes because of a race condition.

There's a `useEffect` hook in the DataViews component that is problematic in a lot of ways but it's not the only issue. The caching of the params is also an issue here.

jsnajdr:
> There's a useEffect hook in the DataViews component that is problematic in a lot of ways but it's not the only issue.

Yes, I can reproduce this, the hook is infinity-loopy. That's fixable.

> The caching of the params is also an issue here.

What are the other problems? What's different with the weakmap is that each `history.push` creates a new `location` object, even if it's the same. For example, a `history.push( { postId } )` where the new `postId` is the same as the previous one.

That can be fixed either by fixing the effect hook that calls `history.push` in a loop, or patching `history.push` to detect noop changes and act as noop.

youknowriad:
> What are the other problems?

No other problems, it's just that using the map trick instead of weak-map fixed the crash above (while keeping the hook...)

youknowriad:
> That can be fixed either by fixing the effect hook that calls history.push in a loop, or patching history.push to detect noop changes and act as noop.

If that's the solution we want to go with, that's totally good for me, I'm not opinionated on how we do the caching, mainly want to avoid the crash.

jsnajdr:
> If that's the solution we want to go with, that's totally good for me

Pushed a fix in 29291e1. The PR should be in a mergeable condition now.

> I personally like the direct hook to retrieve the current route's params.

In a callback, however, retrieving the params iteratively is technically more correct. It avoids an unneeded rerender. And more importantly, see how in my PR I was able to eliminate `params` from dependencies of several `useEffect` and `useCallback` hooks. These effects are not supposed to be triggered by `params` change.

ntsekouras (approved):
Personally I like this imperative approach with `getLocationWithParams`,  couldn't spot any additional renders, and it tests well for me.

youknowriad:
@jsnajdr can we apply a fix at the framework level instead of the application level, as I assume it might be easy to create this issue again inadvertently.

jsnajdr:
@youknowriad One framework-level fix that comes to mind would be `history.push` accepting a callback, similar to `setState`:
```js
history.push( ( params ) => ( {
  ...params,
  postId
} );
```
That's elegant and you don't need the extra `history.getLocationWithParams()` call.

Many Site Editor navigations are of the kind where you want to preserve all query params and only add or remove some of them. This is probably not very typical in mainstream router applications. For example, in Calypso I never encountered a problem of this kind.

PR #61518, opened 2024-05-09
Try drill down for pattern inserter

draganescu:
## What?

Enables on large viewports the same drill down navigation we have in the site editor sidebar and in the small viewport for inserter categories, and in the process removes the preview pane.

## Why?

The preview pane is not very useful: it creates another content area to skip for focus management, and it uses a lot of viewport area which is quite precious for assembling patterns (e.g. it should be possible on a large viewport to have both the inserter and the inspector open and assemble patterns). All that just to keep the category list available.

## How?

- Copies over the component used for the small viewport and adjusts it to also work on larger viewports.
- Removes some zoom out layout tweaks.
- Enables the assembling in zoom out mode when the pattern tab of the inserter is accessed (as opposed to when a category is accessed)
- Tries to make the media and patterns behave the same on category navigation

## Testing Instructions

1. Go to the site editor
2. Go to Pages and edit a page
3. Open the inserter
4. Open the patterns tab
5. **Notice zoom out mode is engaged**
6. **Notice the list of pattern categores**
7. Click a category
8. **Notice the patterns appear in the same panel**
9. Enable responsive mode
10. **Small viewport experience should be unchanged**

### Testing Instructions for Keyboard

N/A

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/107534/ca660195-4876-4df4-8c12-180bab94ddbf

**Known Issues**

- [ ] There are some styling refinements required so that all panels look well spaced and similar
- [ ] There is a wild super scroll appearing when searching for patterns
- [ ] There is a sneaky disappearing scroll when the pattern list is rendering
- [ ] Somehow switching viewport size (?) may result in zoom out mode stuck to on
- [ ] We should move up the search in media if possible so that all tabs have the same UI structure

PR #61831, opened 2024-05-21
Adjust pattern list items resting, hover, focus styles

richtabor:
## What?

Tweaks to fix the various implementations of pattern list items to have the same resting, hover, and focus styles as other related elements. Also forwards https://github.com/WordPress/gutenberg/pull/61518. 

## Testing Instructions

1. Open a post or page.
2. Open the Inserter.
3. Select the "Patterns" tab. 
4. Select a category. 
5. See changes. 

## Screenshots or screencast 
Focus state is the admin blue. 

| Before  | After |
|CleanShot 2024-05-21 at 14 25 09CleanShot 2024-05-21 at 14 26 23 |CleanShot 2024-05-21 at 14 19 07CleanShot 2024-05-21 at 14 17 35|

### DIFF
--- a/packages/block-editor/src/components/block-patterns-list/style.scss
+++ b/packages/block-editor/src/components/block-patterns-list/style.scss
@@ -38,21 +38,22 @@
 		border-radius: $radius-block-ui;
 
 		&::after {
-			outline: $border-width solid rgba($black, 0.075);
+			outline: $border-width solid rgba($black, 0.1);
 			outline-offset: -$border-width;
 			border-radius: $radius-block-ui;
-			transition: outline 0.1s linear;
 		}
 	}
 
 	&:hover:not(:focus) .block-editor-block-preview__container::after {
-		outline-color: rgba($black, 0.2);
+		outline-color: rgba($black, 0.3);
 	}
 
 	&:focus .block-editor-block-preview__container::after {
 		outline-color: var(--wp-admin-theme-color);
 		outline-width: var(--wp-admin-border-width-focus);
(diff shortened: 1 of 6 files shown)


jasmussen (approved):

t-hamano (commented):
In this PR, the focus style of the synced pattern also changes from the purple color to the primary color, which is intentional, right?

image

If so, the `.block-editor-block-patterns-list__list-item-synced` class is no longer used, so we should be able to remove it here as well.

https://github.com/WordPress/gutenberg/blob/5c04dec2971f8f2ee27d4b917bf6de0df7a65a24/packages/block-editor/src/components/block-patterns-list/index.js#L123

packages/block-editor/src/components/block-patterns-list/style.scss
```diff
@@ -26,33 +26,33 @@
...
+		display: flex;
+		align-items: center;
+		overflow: hidden;
+		border-radius: $radius-block-ui;
+
+		&::after {
+			outline: $border-width solid rgba($black, 0.075);
+			outline-offset: -$border-width;
+			border-radius: $radius-block-ui;
+			transition: outline 0.1s linear;
```
t-hamano:
```suggestion
			transition: outline 0.1s linear;
			@include reduce-motion("transition");
```

richtabor:
> In this PR, the focus style of the synced pattern also changes from the purple color to the primary color, which is intentional, right?

Yes, I don't think the focus color of previews should change based on the synced status. Within the canvas it's fine as you're selecting the actual entity, but within the inserter it's an element that adds the pattern/synced pattern.

PR #60561, opened 2024-04-08
Editor: Animate opening and closing editor right sidebar

youknowriad:
Related to #60423

## What?

This PR animates the right sidebar of the post and site editors (opening and closing the sidebar). It takes a different approach than #60423 (animates the width of the complementary area which automatically pushes the content so we don't need to animate the content).

**Notes**

 - I'm not entirely sure about the performance impact of this PR. Something to be checked.
 - I wanted to avoid animating the sidebar if we're replacing the sidebar with another one (like when switching between global styles and post sidebar) but failed to do so, so far.

## Testing Instructions

- Try the post and site editors.
- Try opening and closing sidebars there.

PR #60286, opened 2024-03-28
Exploration: Simpler outlines and synced pattern interactions

richtabor:
**Not intended for direct merge.** 

This is an exploration of a number of ideas around block selection, outlines and related interactions, combined to get a solid sense of how these will work together. 

My goal is to try and make it less "weighty" selecting blocks in the Site Editor, more predictable, and less concepts to understand. Less moving parts if you would. 

If we like the direction we can split it into proper PRs. 

## What?
1. Changed the editable block indicator to render only for synced patterns, as you can now select headers and footers when editing pages in the Site Editor. These editable indicators are now background color, rather than the existing dotted outline. 
2. Try `outline` and `outline-offset` as the mechanism for visible outlines, rather than shadows. Allows for borders to be clearly seen on fullwidth block selection, while also reducing clashing borders when hovering/selecting blocks near each other.
3. Use a consistent `var(--wp-admin-border-width-focus)` width for outlines, rather than `1px` on hover and `1.5px` on selection. 
4. Does not render the `.is-selected` outline when selecting RichText, making text editing less disruptive by outlines.
5. Moves the block toolbar 8px away from the selection, rather than 12px away—just to get a bit less coverage over other blocks.
6. Another pass at using shadows for interface borders that are edge-to-edge with the editable canvas, reducing visual blurry edges when having any color that's not white or black against the header toolbar or either editor sidebar.
7. Try removing border radius from outlines, to reduce visual conflicts when against or over non-radii elements (images, fullwidth blocks, edge of the canvas, etc).

## Testing Instructions

1. Open the Site Editor.
2. Explore the changes listed above by selecting blocks. 
3. Add a synced pattern with overrides, to see the editable block indicators.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/1813435/798051a9-ffcd-4a59-b3b6-c3ee3db45b9d

CleanShot 2024-03-28 at 09 51 52

PR #42282, opened 2022-07-08
Try using shadows instead of borders on interface skeleton

richtabor:
## What?
Replaces gray interface skeleton borders with opaque shadows, which tightens up the UI when darker background colors are used. 

## Why?
This PR explores using box-shadows instead of borders on the interface skeleton, to tighten up the interface on sites that have darker background colors. Currently if the background color of the site is darker than the border, you end up with a slight distortion around the edges of the interface. 

By using shadows, we can maintain an edge with lighter backgrounds, but on darker backgrounds (where the edge is already defined by the background color) it's sharp - instead of distorted. 

## How?
Replacing borders with box shadows. 

## Testing Instructions
1. Open a Post or Page.
2. View the interface skeleton

---

## Before
Notice the distortion around the gray borders against the colored background:

before

But there is distortion on lighter backgrounds:

CleanShot 2022-07-08 at 17 03 08@2x

Zoomed in: 

CleanShot 2022-07-08 at 17 06 56@2x

---

## After
No distortion around the edges of the interface:

after

Zoomed in: 

CleanShot 2022-07-08 at 17 07 23@2x

Borders with lighter backgrounds still look as they do today:

CleanShot 2022-07-08 at 17 01 00@2x

PR #61835, opened 2024-05-21
Shadows instead of borders on interface skeleton

richtabor:
## What?
Part of the exploration in #60286 and an alternate to https://github.com/WordPress/gutenberg/pull/42282, attempting to reduce visual conflicts when viewing sites with background colors that are not lighter than $gray-200. Currently if the background color of the site is darker than the border, you end up with a slight distortion around the edges of the interface.

Also fixes a 1px offset when the Inspector is closed, from the animation adjusting the canvas based on the width of the sidebar (not accounting for the 1px border) in https://github.com/WordPress/gutenberg/pull/60561. 

## How?
Replace borders with box shadows. 

## Testing Instructions

1. Open a Post or Page.
2. View the interface skeleton

## Screenshots or screencast 

| Before  | After |
|CleanShot 2024-05-21 at 15 50 58|CleanShot 2024-05-21 at 15 44 34|

### DIFF
--- a/packages/edit-post/src/components/header/fullscreen-mode-close/style.scss
+++ b/packages/edit-post/src/components/header/fullscreen-mode-close/style.scss
@@ -14,10 +14,9 @@
 		background: $gray-900;
 		color: $white;
 		border-radius: 0;
-		height: $header-height + $border-width;
+		height: $header-height;
 		width: $header-height;
 		position: relative;
-		margin-bottom: - $border-width;
 
 		&:active {
 			color: $white;
--- a/packages/edit-site/src/components/layout/style.scss
+++ b/packages/edit-site/src/components/layout/style.scss
@@ -162,15 +162,13 @@
 	position: relative;
 	color: $white;
 	border-radius: 0;
-	height: $header-height +  $border-width;
+	height: $header-height;
 	width: $header-height;
-	margin-bottom: -  $border-width;
 	overflow: hidden;
 	padding: 0;
 	display: flex;
 	align-items: center;
 	justify-content: center;
-	border-bottom: 1px solid transparent;
 
 	&:hover,
 	&:active {
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -129,7 +129,7 @@ html.interface-interface-skeleton__html-container {
 	overflow: hidden;
 
 	@include break-medium() {
-		box-shadow: -$border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+		box-shadow: -$border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
 	}
 }
 
@@ -138,7 +138,7 @@ html.interface-interface-skeleton__html-container {
 	right: 0;
 
 	@include break-medium() {
-		box-shadow: $border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+		box-shadow: $border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
 	}
 }
 
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -125,24 +125,27 @@ html.interface-interface-skeleton__html-container {
 }
 
 .interface-interface-skeleton__sidebar {
+	border-top: $border-width solid $gray-200;
 	overflow: hidden;
 
 	@include break-medium() {
-		border-left: $border-width solid $gray-200;
+		box-shadow: -$border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
 	}
 }
 
 .interface-interface-skeleton__secondary-sidebar {
+	border-top: $border-width solid $gray-200;
 	right: 0;
+
 	@include break-medium() {
-		border-right: $border-width solid $gray-200;
+		box-shadow: $border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
 	}
 }
 
 .interface-interface-skeleton__header {
 	flex-shrink: 0;
 	height: auto;  // Keep the height flexible.
-	border-bottom: $border-width solid $gray-200;
+	box-shadow: 0 $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
 	z-index: z-index(".interface-interface-skeleton__header");
 	color: $gray-900;
 }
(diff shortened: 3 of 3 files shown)


richtabor:
@t-hamano what would you suggest for windows high contrast?

jasmussen:
Something's going on here:

Screenshot 2024-05-22 at 10 28 31

The shadow stacks, likely due to the use of opacity. Any neat trick to improve this?

Sidenote, if we go with this, you'll need to combine it with something like an `outline: 1px solid transparent;`. In Windows high contrast mode, all box-shadows are eliminated, but conversely, all _transparent_ or semi-opaque colors are given full opacity. In effect that will mean the transparent color will take the place of the box-shadow, which is important for the boundaries of the block editor to still be visible in that context.

t-hamano:
When it comes to high contrast mode in Windows, we certainly want outlines as a fallback.

### trunk

| trunk

### This PR

pr

From what I've tested, you can make the following changes to this current PR:

```diff
diff --git a/packages/interface/src/components/interface-skeleton/style.scss b/packages/interface/src/components/interface-skeleton/style.scss
index 9d8c40aaee..e4af3d01a4 100644
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -130,6 +130,7 @@ html.interface-interface-skeleton__html-container {
 
        @include break-medium() {
                box-shadow: -$border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+               outline: 1px solid transparent;
        }
 }
 
@@ -139,6 +140,7 @@ html.interface-interface-skeleton__html-container {
 
        @include break-medium() {
                box-shadow: $border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+               outline: 1px solid transparent;
        }
 }
 
@@ -148,6 +150,7 @@ html.interface-interface-skeleton__html-container {
        box-shadow: 0 $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
        z-index: z-index(".interface-interface-skeleton__header");
        color: $gray-900;
+       outline: 1px solid transparent;
 }
 
 .interface-interface-skeleton__footer {
```

richtabor:
> The shadow stacks, likely due to the use of opacity. Any neat trick to improve this?

I have an idea, though it's a long shot I think.

richtabor:
> The shadow stacks, likely due to the use of opacity. Any neat trick to improve this?

Fixed 🎉 

CleanShot 2024-05-22 at 17 32 21

richtabor:
> When it comes to high contrast mode in Windows, we certainly want outlines as a fallback.

@t-hamano Added via https://github.com/WordPress/gutenberg/pull/61835/commits/5519af045dee0e1a2bd2f698077ca80b4ba89b49. Thanks!

jasmussen:
Nice work. Here are some of the effects of this. On colored backgrounds, the shadow through opacity absorbs part of the color:

overlay colors on colored backgrounds

Consequently on dark backgrounds, that means the canvas can feel like it almost bleeds into the shadow:

dark background

It also appears to give 1px of "extra available space" all around the canvas, so in-canvas borders now actually overlap the shadow:

overlap

It's not clear to me that's problematic or not, I don't mind it — it's UI after all.

One thing that's slightly more visible is the inconsistency around the site logo. Compare these upscaled versions, post editor:
post editor

Site editor:

site editor

Both are 61px tall, that's the header height plus a 1px border below. But there's something different in the visual, not quite sure what it is. In both cases, should they now be 60px, since this PR effectively removes the discrepancy that was previously in place with the non-opacity border?

richtabor:
> Consequently on dark backgrounds, that means the canvas can feel like it almost bleeds into the shadow:

Yes, the idea is that the shadow is visible when it's needed to define the separation between UI and editor (the lighter the background, the more definition is needed). The same effect is employed on color palette swatches, so that you can see where the color selector is contained to, when you don't see a visual difference between the UI background and the color itself. 


> It also appears to give 1px of "extra available space" all around the canvas, so in-canvas borders now actually overlap the shadow:

Yes, this is also intentional, making the selection styles on fullwidth blocks much sharper and clear, as they're not right up against the light gray. 

Here's a before/after: 

| Before  | After |
|CleanShot 2024-05-21 at 15 50 58|CleanShot 2024-05-21 at 15 44 34|

> One thing that's slightly more visible is the inconsistency around the site logo. Compare these upscaled versions, post editor:

I'll take a look.

richtabor:
@jasmussen updated the fullscreen close/site logo to be 1:1 across both: 

CleanShot 2024-05-24 at 11 50 30
CleanShot 2024-05-24 at 11 52 12

jasmussen:
Very nice.

t-hamano (approved):
LGTM!

I tested several scenarios on Windows OS.

Chrome:

chrome

Firefox:

firefox

Chrome, High Contrast Mode:

chrome-high-contrast

Firefox, High Contrast Mode:

firefox-high-contrast

---

Nit: We may want to avoid hard-coding color values.

```diff
diff --git a/packages/interface/src/components/interface-skeleton/style.scss b/packages/interface/src/components/interface-skeleton/style.scss
index be8eced71d..5b77075b95 100644
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -129,7 +129,7 @@ html.interface-interface-skeleton__html-container {
        overflow: hidden;
 
        @include break-medium() {
-               box-shadow: -$border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+               box-shadow: -$border-width $border-width 0 0 rgba($color: $black, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
                outline: 1px solid transparent; // Shown for Windows 10 High Contrast mode.
        }
 }
@@ -139,7 +139,7 @@ html.interface-interface-skeleton__html-container {
        right: 0;
 
        @include break-medium() {
-               box-shadow: $border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+               box-shadow: $border-width $border-width 0 0 rgba($color: $black, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
                outline: 1px solid transparent; // Shown for Windows 10 High Contrast mode.
        }
 }
@@ -147,7 +147,7 @@ html.interface-interface-skeleton__html-container {
 .interface-interface-skeleton__header {
        flex-shrink: 0;
        height: auto;  // Keep the height flexible.
-       box-shadow: 0 $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+       box-shadow: 0 $border-width 0 0 rgba($color: $black, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
        z-index: z-index(".interface-interface-skeleton__header");
        color: $gray-900;
        outline: 1px solid transparent; // Shown for Windows 10 High Contrast mode.
```

stokesman:
Pardon me for being too lazy to make an issue at the moment and if I'm attributing this to the wrong PR. I'm seeing the block toolbar somehow not being shadowed:
shadow-fail-on-block-toolbar

Can anyone confirm? Though I've been seeing it through various restarts of `npm run dev` and think it must be legit. If so, hopefully it’s quite a simple fix.

t-hamano:
I was able to reproduce the issue as well. I haven't been able to fully investigate if this PR is related, but the issue does not occur in WP6.5.

In the video below, the first half tests the latest Gutenberg, and the second half tests WP6.5.4.

https://github.com/WordPress/gutenberg/assets/54422211/567d57b2-5b38-41aa-97f6-48ff271420b4

t-hamano:
I used `git bisect` to investigate this issue further and it appears that this issue did indeed occur in this PR.

I've reported the issue on #62639.

PR #61839, opened 2024-05-22
Docs: Fix missing braces into code snippet

HelloAlberuni:
The parameters of useBlockProps should be an object otherwise it generates error.

## What?

It adds the missing braces to code snippets using the `useBlockProps` hook.

## Why?
The parameters of the useBlockProps hook must be an object; otherwise, it will throw an error in the console. I've seen another code snippet [in this doc](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-edit-save/#block-wrapper-props passed) passed an object which is the right way.

### DIFF
--- a/packages/block-editor/README.md
+++ b/packages/block-editor/README.md
@@ -926,13 +926,13 @@ import { useBlockProps } from '@wordpress/block-editor';
 
 export default function Edit() {
 
-  const blockProps = useBlockProps(
+  const blockProps = useBlockProps( {
     className: 'my-custom-class',
     style: {
       color: '#222222',
       backgroundColor: '#eeeeee'
     }
-  )
+  } );
 
   return (
     <div { ...blockProps }>


Mamaduka:
Thanks for fixing the docs, @HelloAlberuni!

This particular example is auto-generated, and you'll need to fix it in this file https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/block-list/use-block-props/index.js#L52-L58.

Before committing the changes, make sure to run - the `npm run docs:build` command.

shail-mehta:
@HelloAlberuni, Thanks for the PR. 

Could you please Resolve the Conflict and incorporate the suggested changes mentioned in the comment above?

Mamaduka:
It looks like the issue has been fixed via https://github.com/WordPress/gutenberg/pull/64363.

PR #58409, opened 2024-01-29
Add defaultFontSizes theme.json (theme.json v3)

ajlende:
## Dev Note 📝 

See https://github.com/WordPress/gutenberg/pull/58409#issuecomment-2078077477

## What?

Reimplementation of #56661 with a proper theme.json migration.

Adds a `defaultFontSizes` option to disable showing the default font sizes.

When `defaultFontSizes === false` the options are removed from the UI.

In theme.json v3, by default `defaultFontSizes` is `true`.

In theme.json v2, `defaultFontSizes` isn't a valid key, but is treated as `false` where it would be used when a theme has font sizes defined for backwards compatibility of generated CSS.

## Why?

Having an option for `defaultFontSizes` would give themes more control over which options are available. This option brings consistency to font sizes to match colors, gradients, duotone filters, and shadow presets.

## How?

Goals are to migrate the behavior of the font sizes presets to be like other presets that can be disabled via `default*` settings.

## Testing Instructions

Screenshots below contain example modified `fontSizes` JSON for TT4 to cover various cases of having font sizes and overriding default ones.

For each modified theme.json test that each step works as expected (see screenshots):

1. Test with theme.json `"version": 2`. (trunk and this PR should be the same.)
2. Update your theme.json to `"version": 3`. (Don't forget to update the style variations too.)
3. Update your theme.json (again) with `settings.typography.defaultFontSizes` to `false`.

Then test custom origin styles migration.

1. Start on trunk with a `"version": 2` theme.json with no associated `wp_global_styles` `post_type`s in the DB.
2. Make changes to the global styles UI such as adding a new color.
3. See the `"version": 2` entry in the DB.
4. Switch to this branch.
5. Make more changes to the global styles UI such removing the color.
6. See that the entry is updated to `"version": 3` in the DB.
7. See that `"defaultFontSizes": false` is _not_ added to the entry in the DB.
8. Update the theme's theme.json to `"version": 3`.
9. Make more updates in the UI.
10. See that `"defaultFontSizes": false` is still not added.

### Testing Instructions for Keyboard

N/A

## Screenshots or screencast 

Each set of screenshots shows a different example theme.json modification to TT4.

- v2 trunk: Screenshot of previous behavior in v2 theme.json.
- v2: Screenshot from this branch for v2 theme.json.
- v3 (default): Screenshot with v3 theme.json. (Should be the same as `"defaultFontSizes": true`.)
- v3 `false`: Screenshot with `"defaultFontSizes": false` set in v3 theme.json.

[...]

PR #44247, opened 2022-09-19
Spacing presets: switch to using numbers instead of t-shirt sizes for labels

glendaviesnz:
## What?
Replaces `X-Small, Small, Medium` labels with simple `1,2,3,4` labels for spacing presets

## Why?
Requested as [a follow-up change here](https://github.com/WordPress/gutenberg/issues/43412#issuecomment-1249403414).

## How?
Updates the `set_spacing_sizes()` to assign numbers as the `name` value of each spacing size

## Testing Instructions

- Add a Group block and set the padding and margin spacing and check that the number labels display as expected
- Check that other spacing size related functionality still works as expected in editor, frontend and global styles
- Update `settings.spacing.spacingScale.steps` to be greater than `7` and check that the select list version of the components still shows t-shirt sizes

## Screenshots or screencast 

Before:

https://user-images.githubusercontent.com/3629020/190936381-dd72e659-d47e-41eb-bbab-85b8a47ded1d.mp4

After:

https://user-images.githubusercontent.com/3629020/190936395-a393026e-542c-41f5-86cb-419effe028f2.mp4

PR #61842, opened 2024-05-22
Add defaultSpacingSizes option (theme.json v3)

### LINKED ISSUE #55636: Spacing: Spacing Sizes are added
### Description

Spacing sizes are getting added to the list, even though I have defined different ones in my theme.
This seems to happen to any theme that does not have `1`, `2`, `3`… defined as spacing size.

1. Select a theme with custom spacing sizes
2. Wonder where the new spacing sizes are coming from

image

### Environment info

- WP  6.4-RC2
- GB 16.9.0

No

No

ajlende:
## Dev Note 📝 

Since #58409 is so similar, I've updated the dev note there to include these changes too. See https://github.com/WordPress/gutenberg/pull/58409#issuecomment-2078077477.

## What?

As a follow-up to https://github.com/WordPress/gutenberg/pull/58409, to get the migration in the same theme.json version bump.

Adds a `defaultSpacingSizes` option to disable showing the default spacing sizes.

When `defaultSpacingSizes === false` the options are removed from the UI.

In theme.json v3, by default `defaultSpacingSizes` is `true`.

In theme.json v2, `defaultSpacingSizes` isn't a valid key, but is treated as `false` where it would be used when a theme has spacing sizes defined for backwards compatibility of generated CSS.

In v3, the sizes generated with `spacingScale` are merged with the custom `spacingSizes` with the custom `spacingSizes` taking precedence when it uses the same slug as one of the generated ones.

## Why?

Fixes #55636. See https://github.com/WordPress/gutenberg/issues/55636#issuecomment-1784540462 since the suggested resolution changed from the original description.

This option brings consistency to spacing sizes to match colors, gradients, duotone filters, and shadow presets.

## How?

Goals are to migrate the behavior of the spacing sizes presets to be like other presets that can be disabled via `default*` settings.

## Testing Instructions

Test various combinations of the following settings:

- Previous or current behavior
- More or fewer than 7 presets (dropdown vs range control)
- Theme.json `version` set to `2` or `3`
- `defaultSpacingSizes` set to `true` or `false`
- `spacing.spacingScale` set to `{ "steps": 0 }` or empty
- Presets with same or different slugs (default slugs are 20,30,40,50,60,70,80)
- Partial or complete overlap of default slugs
- Setting custom user slugs (may need to write database directly)
- Setting `spacingScale` vs `spacingSizes` or both

N/A

## Screenshots or screencast 

TODO


lib/class-wp-theme-json-resolver-gutenberg.php
```diff
@@ -603,7 +603,6 @@ public static function get_merged_data( $origin = 'custom' ) {
 		$result = new WP_Theme_JSON_Gutenberg();
 		$result->merge( static::get_core_data() );
 		if ( 'default' === $origin ) {
-			$result->set_spacing_sizes();
```
ajlende, line 603: Moved to [`WP_Theme_JSON_Gutenberg->__construct()`](#diff-b03597cc3da199e5aa5a94950e8241135904853e7c3b82d758e42744878afae7R782) so that the generated presets for each origin are merged properly rather than always being set as `default` origin.

packages/block-editor/src/private-apis.js
```diff
@@ -40,6 +40,7 @@ import {
 import { requiresWrapperOnCopy } from './components/writing-flow/utils';
 import { PrivateRichText } from './components/rich-text/';
 import { PrivateBlockPopover } from './components/block-popover';
+import useSpacingSizes from './components/spacing-sizes-control/hooks/use-spacing-sizes';
```
ajlende, line 45: The spacer block was using the list of presets to conditionally show UI elements. For the sake of not changing too many things to get this into 6.6, I just exported the same `useSpacingSizes` hook that the block editor package uses as a private API rather than refactoring the spacer block to not need it.

packages/block-editor/src/components/spacing-sizes-control/hooks/use-spacing-sizes.js
```diff
@@ -1,28 +1,57 @@
...
+						{
+							name: __( 'Default' ),
+							slug: 'default',
+							size: undefined,
+						},
+				  ]
+				: null;
+
+		return [
+			...maybeUnsetSize,
```
MaggieCabrera: I'm getting `Uncaught TypeError: maybeUnsetSize is null` when I insert a spacer in the post editor. `sizesLength` for me is 7 and my theme.json uses (v2): 
```

"spacing": {
	"spacingScale": {
		"steps": 0
	},
	"spacingSizes": [
		{
			"name": "1",
			"size": "1rem",
			"slug": "10"
		},
		{
			"name": "2",
			"size": "2rem",
			"slug": "20"
		},
		{
			"name": "3",
			"size": "3rem",
			"slug": "30"
		},
		{
			"name": "4",
			"size": "4rem",
			"slug": "40"
		},
		{
			"name": "5",
			"size": "5rem",
			"slug": "50"
		},
		{
			"name": "6",
			"size": "6rem",
			"slug": "60"
		}
	],
	"units": ["%", "px", "em", "rem", "vh", "vw"]
},
```

ajlende: So I thought that I could simplify this to not use an `EMPTY_ARRAY` constant because `null` should expand to nothing. And it does for objects, but not arrays...

image

Should be fixed by 3203c48ae95f780fed7d9d45d2fd475496cb44a2.

ellatrix:
Regarding the "Verify Core Backport Changlog", it'd be good to make a backport PR with that combines these and the previous changes, then add them as you see here: https://github.com/WordPress/gutenberg/blob/trunk/backport-changelog/readme.md

MaggieCabrera:
Given the nature of this change, I think this "feature" should be blessed, because 6.6 is already going to ship with theme.json v3, this change should make it in too, but we need to make sure we are not breaking anything with it. /cc @priethor

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -3682,6 +3794,97 @@ public function set_spacing_sizes() {
...
+			! isset( $spacing_scale['unit'] ) ||
+			! in_array( $spacing_scale['unit'], static::VALID_SPACING_UNITS, true ) ||
+			! isset( $spacing_scale['operator'] ) ||
+			! in_array( $spacing_scale['operator'], static::VALID_SPACING_OPERATORS, true ) ||
+			! isset( $spacing_scale['increment'] ) ||
+			! is_numeric( $spacing_scale['increment'] ) ||
+			$spacing_scale['increment'] <= 0
+		) {
+			return array();
+		}
```
ajlende: I had to reimplement [this](https://github.com/WordPress/gutenberg/pull/61842/files#diff-b03597cc3da199e5aa5a94950e8241135904853e7c3b82d758e42744878afae7L3664-R3790) to fail silently for partials that were expected in earlier stages of merging that didn't have access to the default presets. So while I was rewriting it, I updated the check to be more precise.

One of the features I added was a limit to the number of steps that could be generated. The algorithm didn't work correctly above 10 steps. It no longer generated the correct number of presets. Anything below 3X-Small didn't get added because the slug was less than 10. For example, `"steps": 20` only generated slugs 10-150 (or 15 steps).

With this change as-is, nothing gets output instead of bad output, and we probably can't have that because of backwards compatibility.

I think there are a few options for solving this without breaking old themes in order of increasing effort for the immediate PR.
1. Continue to use the old check for v3 even though it produces bad output.
2. "Fix" the algorithm for v2 and v3 by just shoving those extra presets to the XL end. Problem is that I'm not sure if people's intention is to have a limit of 3X-Small on the low end, but unbounded on the high end, especially if there is room for the increment to make sizes smaller than 3X-Small before reaching zero.
3. Add a special case for v2 to use the old check there, but use the new check in v3. Not as easy as it sounds because at this point in the code everything is already v3, but probably doable by setting an internal `"migratedFrom": [2]` key or something during the migration to list all the migrations that have happened to trigger the behavior.

packages/block-editor/src/components/spacing-sizes-control/hooks/use-spacing-sizes.js
```diff
@@ -1,28 +1,58 @@
...
+	const themeSizes = themeSpacingSizes ?? EMPTY_ARRAY;
+
+	const defaultSizes =
+		defaultSpacingSizes && defaultSpacingSizesEnabled !== false
+			? defaultSpacingSizes
+			: EMPTY_ARRAY;
+
+	return useMemo( () => {
+		const sizes = [
+			{ name: __( 'None' ), slug: '0', size: 0 },
```
ajlende, line 41: I updated a small UX oddity where the dropdown would show the number `0` when the rest of the generated presets used t-shirt sizing names.

lib/class-wp-theme-json-gutenberg.php
ajlende: > Given the nature of this change, I think this "feature" should be blessed, because 6.6 is already going to ship with theme.json v3, this change should make it in too, but we need to make sure we are not breaking anything with it. /cc @priethor

If the feature gets blessed, then I'll just fix the spacing scale algorithm because I know I'll have enough time.

Alternatively, we could merge with this known bug and add a follow-up during the RC to fix it.

MaggieCabrera: If this doesn't get blessed it would be best to just do 1 and leave the fix for later (and the decision). From what I've gathered so far investigating themes, they usually have 10 spacing sizes tops, but they are always custom values. It feels like using `spacingScale` for that many steps would be very much an edge case.

ajlende: I just reverted the check to the old one and added a comment.

MaggieCabrera:
Some testing:

`spacingSizes` defined (8 values, some use the preset slugs, some don't)
`spacingScale` defined (5 steps)

On trunk:

**v2**
Result: only the 8 spacingSizes show on the dropdown

On this PR:

**v2**
Result: only the 8 spacingSizes show on the dropdown

**v3**
Result: merges both and I only see the spacingSizes that don't have the same slug as the default presets, the rest show t-shirt sizes, but I see 7 values (from 2XS to 2XL), when I was expecting 5

If I set defaultSpacingSizes to false, then I go back to seeing just the custom spacingSizes only, as expected

schemas/json/theme.json
```diff
@@ -463,16 +468,22 @@
...
 									"type": "integer",
+									"minimum": 1,
+									"maximum": 10,
 									"default": 7
 								},
 								"mediumStep": {
 									"description": "The value to medium setting in the scale.",
 									"type": "number",
+									"exclusiveMinimum": true,
+									"minimum": 0,
```
ajlende, line 487: Even though I've removed the check in the code for these ranges, I think it's a good idea to keep them in the schema so people at least get warned about the potential issues.

docs/reference-guides/theme-json-reference/theme-json-migrations.md
```diff
@@ -93,3 +93,22 @@ In theme.json v2, the default font sizes were only shown when theme sizes were n
...
+- When set to `true` it will show the default spacing sizes and prevent them from being overridden by the theme.
+- When set to `false` it will hide the default spacing sizes and allow the theme to use the default slugs.
+
+It is `true` by default when switching to v3. This is to be consistent with how other `default*` options work such as `settings.color.defaultPalette`, but differs from the behavior in v2.
+
+In theme.json v2, the default spacing sizes were only shown when theme `spacingSizes` presets and `spacingScale` config were not defined. A theme providing font sizes with the same slugs as the defaults would always override the default ones.
+
+To keep behavior similar to v2 with a v3 theme.json:
+* If you do not have any `spacingSizes` presets or `spacingScale` config defined, `defaultSpacingSizes` can be left out or set to `true`.
+* If you disabled default spacing sizes by setting `spacingScale` to `{ "steps": 0 }`, remove the `spacingScale` config and set `defaultSpacingSizes` to `false`.
```
MaggieCabrera, line 112: Should we add a comment about the merge behaviour on v3?

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -3762,6 +3830,94 @@ public function set_spacing_sizes() {
...
+			! isset( $spacing_scale['steps'] ) ||
+			! is_numeric( $spacing_scale['steps'] ) ||
+			0 === $spacing_scale['steps'] ||
+			! isset( $spacing_scale['mediumStep'] ) ||
+			! is_numeric( $spacing_scale['mediumStep'] ) ||
+			! isset( $spacing_scale['unit'] ) ||
+			! isset( $spacing_scale['operator'] ) ||
+			( '+' !== $spacing_scale['operator'] && '*' !== $spacing_scale['operator'] ) ||
+			! isset( $spacing_scale['increment'] ) ||
+			! is_numeric( $spacing_scale['increment'] )
```
MaggieCabrera, line 3916: Not a blocker but I wonder if we should have unit tests for this mammoth of a condition? Can be a follow up

MaggieCabrera (approved):
This is working as expected. Thank you for the extensive comments, they helped a lot. Let's bring this in for 6.6 since we are already incrementing the theme.json version.

lib/class-wp-theme-json-gutenberg.php
ajlende: Let's leave it for a follow-up. There are already some tests for the existence of these conditions, but not the things it doesn't catch like `"steps": 20` only generating 15 steps.

phpunit/class-wp-theme-json-test.php
```diff
@@ -4466,27 +4466,27 @@ public function data_set_spacing_sizes() {
...
 					),
 					array(
-						'name' => '4',
+						'name' => 'Large',
 						'slug' => '60',
 						'size' => '6rem',
 					),
 					array(
-						'name' => '5',
+						'name' => 'X-Large',
```
ajlende, line 4489: These renames are all done in the UI instead in case the UI wants to show the numbers in the tooltips in the slider, but an actually useful name elsewhere. See https://github.com/WordPress/gutenberg/pull/44247.

docs/how-to-guides/themes/global-settings-and-styles.md
```diff
@@ -336,16 +336,7 @@ The following presets can be defined via `theme.json`:
...
-    - `steps`: the number of steps to generate in the spacing scale. The default is 7. To prevent the generation of the spacing presets, and to disable the related UI, this can be set to `0`.
-    - `mediumStep`: the steps in the scale are generated descending and ascending from a medium step, so this should be the size value of the medium space, without the unit. The default medium step is `1.5rem` so the mediumStep value is `1.5`.
-    - `unit`: the unit the scale uses, eg. `px, rem, em, %`. The default is `rem`.
-- `spacing.spacingSizes`: themes can choose to include a static `spacing.spacingSizes` array of spacing preset sizes if they have a sequence of sizes that can't be generated via an increment or multiplier.
-    - `name`: a human readable name for the size, eg. `Small, Medium, Large`.
-    - `slug`: the machine readable name. In order to provide the best cross site/theme compatibility the slugs should be in the format, "10","20","30","40","50","60", with "50" representing the `Medium` size value.
-    - `size`: the size, including the unit, eg. `1.5rem`. It is possible to include fluid values like `clamp(2rem, 10vw, 20rem)`.
```
ajlende, line 348: This isn't the documentation page where keys are defined, just the one that explains what gets generated in the stylesheet, so I removed this chunk. The actual documentation exists in the [Theme Handbook](https://developer.wordpress.org/themes/global-settings-and-styles/settings/spacing/) which isn't part of Gutenberg source code.

richtabor:
This shouldn't have any effect on V2 `spacingSizes`, correct? 

I'm noting a few issues with the "Powder" theme for example, which registers [it's own spacing values](https://github.com/bgardner/powder/blob/main/theme.json#L146): 

### Trunk (from this PR)

https://github.com/WordPress/gutenberg/assets/1813435/8ad2520f-abdf-44ba-a90f-0738cf0e9c2f

### Without Gutenberg

https://github.com/WordPress/gutenberg/assets/1813435/f0b7af2d-3358-4db6-b6a3-e6f2eae84026

Notice the values and labels are not accurate in trunk.

richtabor:
All of these are rendered, which still properly include the theme's custom slugs (which seem to be the issue). The control is applying the presets inaccurately, with missing `name` values. 

CleanShot 2024-05-31 at 14 44 30

bgardner:
Wanted to note that I created an issue https://github.com/WordPress/gutenberg/issues/62194

ajlende:
Responded on @bgardner's issue https://github.com/WordPress/gutenberg/issues/62194#issuecomment-2142933855. Thanks for raising this. 😄

PR #61844, opened 2024-05-22
Move default redering mode setting to `block_editor_settings_all` hook in PHP to allow customizing it

### LINKED ISSUE #58038: Add an option to change default renderingMode of Post Editor
In #56817 the ability to preview the template of a post right in the Post Editor. This is a really nice update that is a huge win for making the editor look like the frontend. 

Ideally, I would love to be able to have this turned on by default. Either as a user setting, or even a global option that could be applied via a filter somewhere. 

Another option would be for the setting to simply get remembered so that when you enable the `template-locked` rendereing mode and for example reload the editor the setting gets remembered. Kind of like the code editor vs. visual editor setting gets remembered.

fabiankaegy:
## What?

Moves the `defaultRenderingMode` property from being hardcoded in JS to being an option set in PHP via the `block_editor_settings_all` hook.

## Why?

To allow extenders to change the default rendering mode for their custom post types.

## How?

By moving the property to the `block_editor_settings_all` hook.

## Testing Instructions

To test that this functions properly you can either add your own custom filter on the `block_editor_settings_all` hook and modify the `defaultRenderingMode` through that, or you can change this line here: https://github.com/WordPress/gutenberg/blob/9ef296a5165320e03735dc2b659b4ce7462e883a/lib/block-editor-settings.php#L149 to 

```php
	$settings['defaultRenderingMode'] = 'template-locked';
```


youknowriad:
I understand the reason for this, but I'm not sure I agree with the solution personally. For instance I don't think all CPTs from the site editor should have the same default rendering mode. I personally would prefer a CPT property instead (like we do for template)

fabiankaegy:
@youknowriad I agree with the fact that it should be a property of the post types. Looking at the `template` / `template_lock` properties however they already work the exact same way: https://github.com/WordPress/wordpress-develop/blob/6662cb0bef6bdd3fd2aa02db92f5781a4b5db657/src/wp-admin/edit-form-blocks.php#L229-L232

They may be a property of the post type. But they get piped to the editor via the block editor settings hook. 

So I think this PR is the first step of getting us there :)

youknowriad:
I think using block editor settings hook for this is not the right thing (and it's not the right thing for template/templateLock either but it's history now :P). Can we instead add these properties to the the post types REST API instead?

backport-changelog/6.6/6601.md
```diff
@@ -0,0 +1,3 @@
+https://github.com/WordPress/wordpress-develop/pull/6601
+
+* https://github.com/WordPress/gutenberg/pull/61844
```
ellatrix, line 3: Thank you ❤️

fabiankaegy:
Closing in favor of https://github.com/WordPress/gutenberg/pull/62304

PR #61845, opened 2024-05-22
Add a lock snackbar for modifying content lock blocks

### LINKED ISSUE #61554: Proposal: Add snackbar notification when unlocking content only blocks
Perhaps there is a persistent snackbar that renders when you opt to unlock a content-only locked block. 

This would allow you to easily re-lock a temporarily unlocked block, instead of having to find the "Done" control on the parent locked block. I'd expect that selecting "Done" in the block toolbar of the parent block would dismiss the snackbar, and clicking the "Lock" call the action would re-lock the block and dismiss the snackbar. 

Perhaps the copy needs refinement, but I think it's an interesting approach to be able to easily re-lock a recently unlocked content-only block. 

CleanShot 2024-05-10 at 13 47 52

I thought of this while reviewing https://github.com/WordPress/gutenberg/pull/61127.

kevin940726:
## What?

Close #61554.

Add a "Done editing? Lock" snackbar for easier access to re-lock a modifying content-locked block.

> [!NOTE]
> This UI is still TBD. This PR is created to help with testing.

## Why?

See #61554. This is a UI enhancement to improve the discoverability of the feature.

## How?

Add a snackbar once users enter the temporary modifying mode.

## Testing Instructions

1. Add a template-locked block or insert the content below.
```html
<!-- wp:group {"templateLock":"contentOnly","layout":{"type":"constrained"}} -->
<p>template lock</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```
2. Go to the list view and select the Group block.
3. Open the "More" menu.
4. Click on "Modify".
5. The snackbar should appear and announce the message.
6. Clicking on the "Lock" action in the snackbar should re-lock the group and return the focus to the block.
7. Repeat steps 2~4. Clicking on the "Done" menu item on the block toolbar should behave the same, and it should dismiss the snackbar.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/7753001/d21e9a8f-438f-483a-8dde-a765f0b0a522

### DIFF
--- a/packages/block-editor/src/hooks/content-lock-ui.js
+++ b/packages/block-editor/src/hooks/content-lock-ui.js
@@ -4,12 +4,13 @@
 import { ToolbarButton, MenuItem } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
-import { useCallback } from '@wordpress/element';
+import { store as noticesStore } from '@wordpress/notices';
 
 /**
  * Internal dependencies
  */
 import { store as blockEditorStore } from '../store';
+import { getModifyContentLockSnackbarId } from '../store/utils';
 import { BlockControls, BlockSettingsMenuControls } from '../components';
 import { unlock } from '../lock-unlock';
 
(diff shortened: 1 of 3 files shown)


ramonjd:
Closing as stale

PR #61710, opened 2024-05-16
Visually hide 'Actions' column header

jameskoster:
## What?
Visually hide the 'Actions' column header in data views table layout.

## Why?
We can reduce some visual clutter in the top-right corner of the UI. 

Unlike other column headers the 'Actions' `th` doesn't surface a menu on click, you can also argue that the items in this column are quite clearly actions, so the heading doesn't add a lot of value.

The text is only hidden visually, so will still be announced by screen readers.

| Before | After |
| --- | --- |
| Screenshot 2024-05-16 at 10 38 10 | Screenshot 2024-05-16 at 10 37 20 |

Issue #42505, opened 2022-07-18
Improve the accessibility of the Templates browser 'Actions' column


afercia:
### Description

As a low vision user, I struggle to see the ellipsis in the 'Actions' column of the Templates browser.

At first, I didn't notice there were some buttons in this column. Even now that I know there are buttons, the ellipsis icon is so tiny that I struggle to see it and click reliably on it.

Besides the very light icon, there are other factors that make these buttons hard to see for low vision users:
- The third column header text is visually hidden. Actually, there's an 'Actions' text rendered within a `VisuallyHidden` component. I guess that's because the buttons in the column cells are displayed conditionally so that sometimes the cells are empty. 
-  The buttons are displayed conditionally. Under some conditions, there are no buttons at all so that it's hard to discover the third column _might_ actually have some content.
- The ellipsis button opens a dropdown menu, even when the menu contains just one item. In this case, users are forced to  click twice to actually take the _only_ available action. Seems to me this isn't ideal. When there's only one action, there's no need for a dropdown menu in the first place. Placing the only available action control directly within the table cell would probably be a better option.

Proposal:
- Always display the column header 'Actions'.
- When the action cell is empty, display some text to clarify there are no actions e.g. 'None'.
- Reconsider the usage of the dropdown menu and use buttons within the cell instead.
- Alternatively, keep the icon button but use an icon with a more visible shape.

### Step-by-step reproduction instructions

- Go to the Site editor.
- Create a Taxonomy template.
- Create a Custom template.
- Make sure you have a default template with some saved customizations.
- Click on 'Toggle navigation' to open the Navigation left sidebar.
- Click on 'Templates to open the templates browser.
- Observe the content of the table third column.
- Observe the third column header text is visually hidden.
- Observe your customized templates have an ellipsis icon button in the third column.
- Explore the content of the ellipsis dropdown menu.

### Screenshots, screen recording, code snippet

The ellipsis icon button are barely visible for low vision users:

<img width="1127" alt="Screenshot 2022-07-18 at 15 06 18" src="https://user-images.githubusercontent.com/1682452/179521737-b8257176-f5e0-473d-ba85-8fb6fe57e431.png">

Dropdown menu with only one available action:

<img width="1216" alt="Screenshot 2022-07-15 at 10 55 54" src="https://user-images.githubusercontent.com/1682452/179522159-f3c41208-0837-4d07-a39c-e5f0636d1d60.png">

One more case where the dropdown menu contains only one action:

[...]

PR #42555, opened 2022-07-20
Improve accessibility of the templates list Actions column

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/42505

## What?

Changes the Actions button in the templates list to use text instead of only an icon. Also makes the column header visible to as there's no reason to hide it.

## Why?

The ellipsis icon of the Actions button is barely visible for low vision users. In other places, for example in the block toolbar, the ellipsis icon is more visible because of the context and the borders. Instead, in the templates list is very light.

From an accessibility perspective, the golden rule for 'icon-only' buttons should be: use them only when there's really no space to use visible text. In this case, there's enough space to use tedxt and there's no reason to use only an icon.

## How?

- Changes the dropdown menu toggle button to use visible text + a chevron icon instead of only the ellipsis icon.
- Makes the column header text visible and changes it from 'Actions' to 'Manage' to avoid repetition.
- **Removes the flex styles used for the column**.  This is necessary to make the column header text and the buttons within the cells align correctly. I'm not sure why flexbox was used in the first place.
- Since this is a _data_ table, make sure to not have empty cells and use visually hidden text for the 'no actions' case.

## Testing Instructions
- Go to the Site Editor.
- Make sure to have a default template with some customization and a custom template.
- Click Toggle navigation, and then click Templates.
- Observe the third column header text is now visible.
- Observe the customized templates have a button in the third column. The button text is 'Actions' plus a chevron icon.
- Observe the templates with no actions available have an icon representing a line in the third column and that there's a visually hidden text 'No actions available'. This pattern is consistent with the one already used in the WordPress list tables.
- Check there are no breaking style changes.
- Worth noting this table was not fully responsive already. The CSS changes don't improve responsiveness but seems to me they don't make it worse either.

## Screenshots or screencast 

Before:

before

After: 

after

PR #61847, opened 2024-05-22
Data views: Always show Actions table header

afercia:
See https://github.com/WordPress/gutenberg/pull/61710#issuecomment-2124002994
Reverts https://github.com/WordPress/gutenberg/pull/61710

## What?

Makes the 'Actions' table header in the data views always visible.

## Why?

Hiding table heders is less than ideal for the reasons explained in https://github.com/WordPress/gutenberg/pull/61710#issuecomment-2124002994
Previously: https://github.com/WordPress/gutenberg/issues/42505

## How?

Reverts the changes from https://github.com/WordPress/gutenberg/pull/61710

## Testing Instructions

- Go to the Site editor > Pages > View options > Layout > Table
- Observe the 'Actions' table header is always visible.

### DIFF
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -107,7 +107,8 @@
 		th:first-child {
 			padding-left: $grid-unit-40;
 
-			.dataviews-view-table-header-button {
+			.dataviews-view-table-header-button,
+			.dataviews-view-table-header {
 				margin-left: - #{$grid-unit-10};
 			}
 		}
@@ -221,6 +222,10 @@
 				display: none;
 			}
 		}
+	}
+
+	.dataviews-view-table-header {
+		padding-left: $grid-unit-05;
 	}
 
 	.dataviews-view-table__actions-column {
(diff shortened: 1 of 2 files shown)


t-hamano (commented):
Thanks for the PR!

I didn't know that not being able to see table header text was an issue before. Various approaches were considered in #42555, but I think it's okay to ship this PR for now.

@jameskoster what do you think?

afercia:
Thanks @t-hamano. The visibility of the ellipsis button is still a problem though. However it is now a broader issue, not limited to the data views. In https://github.com/WordPress/gutenberg/pull/42555#issuecomment-1831734289 I promised to create a specific issue about the ellipsis button but failed to do so. Will do it now.

jameskoster:
> A visible table header helps all users e.g. low vision users to draw their attention and identify the Actions column

Is it not adequate that numerous actions appear on row hover/focus?

afercia:
> Is it not adequate that numerous actions appear on row hover/focus?

As a low vision user, I can't even visually identify the column.

From a semantics perspective, I'm not sure I understand why this UI uses a table and then it hides one of the table headers, which are essential part of a table. Semantics and functionality should come first, befor any other considerations related to visual clutter or aesthetics.

jameskoster (approved):
The thinking was that given the cells clearly contain actions, and that the header had no menu (like other headers in the table), that we could afford to visually hide it, in this case. If it's not the case, let's revert :)

PR #61802, opened 2024-05-20
SlotFills: Use state for registry initialization

Mamaduka:
## What?
PR updates SlotFill providers to initialize context values via `useState` lazily. The code used the same initialization method but changed during TS refactoring - 821f4d4344f630b337fed4c5beb10eee346899b0.

## Why?
* Fixes `react-compiler/react-compiler` [error](https://github.com/WordPress/gutenberg/actions/runs/9158471368/job/25176908344?pr=61788#step:5:625) - `Expected the first argument to be an inline function expression`.
* This is a valid pattern for initializing values. See https://tkdodo.eu/blog/use-state-for-one-time-initializations and https://x.com/dan_abramov2/status/1792555601106551229.

## Testing Instructions
CI checks should be green.

### Testing Instructions for Keyboard
Same.

PR #61848, opened 2024-05-22
Style Book: Use state to initialize examples

Mamaduka:
## What?
PR updates method for initializing examples for the `StyleBook` component.

This is similar to #61802.

## Why?

- Fixes `react-compiler/react-compiler` [error](https://github.com/WordPress/gutenberg/actions/runs/9173492334/job/25222306709?pr=61788#step:5:779) - `Expected the first argument to be an inline function expression.`
- This is a valid pattern for initializing values. See https://tkdodo.eu/blog/use-state-for-one-time-initializations and https://x.com/dan_abramov2/status/1792555601106551229.

## Testing Instructions
1. Open the Site Editor.
2. Navigate to the Style Book.
3. It should work as before.

Same.


tyxla (approved):
Thanks, looks good 🙌

PR #61850, opened 2024-05-22
Block settings: Update variant of "Apply globally" Button component to secondary

### LINKED ISSUE #61834: Block settings: Change "Apply globally" to secondary button styling to lower prominence
While testing overrides in synced patterns, I noticed that the prominence of "apply globally" really stands out even more when we have now the option to enable/disable overrides: 

Screenshot 2024-05-21 at 12 33 12 PM

To create more of a balance here, I propose we change the "apply globally" option to be a secondary button styling rather than primary like so:

Screenshot 2024-05-21 at 12 33 50 PM

If anything, I see creating overrides as being more of a common action than taking an individual styling and applying it globally but, at the least, we should have equal weighting here in my opinion. cc @WordPress/gutenberg-design for thoughts!

amitraj2203:
## What?
Fixes: #61834

## Why?
The prominence of "Apply globally" button stands out even more when we have now the option to enable/disable overrides.

## How?
This PR changes the variant of `Apply globally` button to secondary

## Testing Instructions

1. Activate any FSE theme. For example TT4
2. Go to Appearance > Editor
3. Go to Patterns > My Patterns and click on any pattern to edit it.
4. On the right-hand side, go to the Advanced panel and see the `Apply globally` button at the bottom. (Make sure to change any styles so that the button comes in active state)

## Screenshots
image


jasmussen (approved):
Nice one.

t-hamano (approved):
LGTM !

PR #61185, opened 2024-04-29
DataViews: Enable types

youknowriad:
Related #55083 

## What?

The DataViews package is a types heavy package. There's a lot of structures that need to be documented properly. So this PR introduces types and typescript to the package. It's not exhaustive by any means, The idea is to add the config and start typing bit by bit. This one adds a "Field" type which is one of the most important ones and only types a single file in the package. We can expand the coverage in follow-ups.

## Testing instructions

There's no functional change, it's essentially a code quality + documentation change.

Issue #55083, opened 2023-10-05
DataViews component


youknowriad:
This document keeps track of all the work and tasks to build a powerful DataViews component capable of rendering all kind of lists of data in WordPress Admin and beyond: from Post Lists, taxonomies, patterns, templates...

DataViews is based on a Fields abstraction and support searching, filtering, reordering fields, multiple layouts, pagination and more.

<img width="1132" height="804" alt="Image" src="https://github.com/user-attachments/assets/c74bf6b6-17c0-4cae-8400-d48e54d55022" />

> See also issues and PRs tagged with the [`[Feature] Data Views`](https://github.com/WordPress/gutenberg/labels/%5BFeature%5D%20Data%20Views) label.

## DataViews component

 - [x] Support defining actions per row. [#55096](https://github.com/WordPress/gutenberg/issues/55096)
 - [x] Add support for reordering fields. [#63414](https://github.com/WordPress/gutenberg/issues/63414)
 - [x] Implement a new layout (or view type): Grid/Gallery.
 - [x] Iterations on the grid layout: [#56643](https://github.com/WordPress/gutenberg/issues/56643)
 - [x] Implement the list layout. [#56746](https://github.com/WordPress/gutenberg/issues/56746)
 - [x] List View Iterations [#56983](https://github.com/WordPress/gutenberg/issues/56983) [#57116](https://github.com/WordPress/gutenberg/issues/57116) [#57578](https://github.com/WordPress/gutenberg/issues/57578) [#57523](#57523) [#56602](https://github.com/WordPress/gutenberg/issues/56602) [#57649](https://github.com/WordPress/gutenberg/issues/57649) [#56746](https://github.com/WordPress/gutenberg/issues/56746)
 - [x] Keyboard navigation: potentially use arrow up/down to navigate through the list.  [#57895](https://github.com/WordPress/gutenberg/issues/57895) [#59637](https://github.com/WordPress/gutenberg/issues/59637)
 - [x] Iterations on search and filtering [#55905](https://github.com/WordPress/gutenberg/issues/55905) [#60696](https://github.com/WordPress/gutenberg/issues/60696) [#63152](https://github.com/WordPress/gutenberg/issues/63152)
 - [x] Bulk actions [#55098](https://github.com/WordPress/gutenberg/issues/55098) [#63193](https://github.com/WordPress/gutenberg/issues/63193) [#63150](https://github.com/WordPress/gutenberg/issues/63150)
 - [x] Field registration API
[...]

PR #61854, opened 2024-05-22
DataViews: Full type the dataviews package

youknowriad:
Related #55083 
Follow-up to #61185

## What?

The DataViews package is a types heavy package. There's a lot of structures that need to be documented properly. So this continues the effort to add explicit types to the package. This PR types the root level DataViews component which basically makes the package fully typed.

## Testing instructions

There should be no real change here, it's mostly code quality and documentation.


packages/dataviews/src/filters.tsx
```diff
@@ -117,6 +117,9 @@ const Filters = memo( function Filters< Item extends AnyItem >( {
 			{ filterComponents }
 		</HStack>
 	);
-} );
+}
+
+// A type assertion is used here to keep the type argument.
```
youknowriad, line 122: I wonder why React doesn't do this by default using generics.

youknowriad:
I know we shouldn't be using "as" too much but I'm not sure how to fix the ViewComponent thing, so I've used that for now. Consider updating it if you have a better solution.

sirreal (approved):
This seems reasonable. I'd love to find a way to avoid the type assertion, but that may be more effort than it's worth at this time.

PR #61004, opened 2024-04-23
Zoomed Out View: Don't close the inserter

scruffian:
## What?
Make the inserter not a dialog anymore so that it doesn't close when we focus outside of it.

## Why?
The other sidebar panels don't operate as dialogs, so this brings consistency with them

## How?
Just remove the `useDialog` hook from the component.

## Testing Instructions
1. Open the block inserter
2. Insert a block/pattern
3. Check that the inserter panel stays open

## Screenshots or screencast 
https://github.com/WordPress/gutenberg/assets/275961/6bcfbc7d-02cf-4d69-9593-6024a7bb59cc

PR #61856, opened 2024-05-22
Inserter: Auto-close the inserter unless  the zoom out experiment is on.

scruffian:
## What?
This reverts the change in https://github.com/WordPress/gutenberg/pull/61004 and hides it behind the experiment flag.

## Why?

This is necessary because without zoom out mode on, the pattern selector overlaps the canvas:
Screenshot 2024-05-22 at 11 23 25

The solution to this is to move the pattern categories inside the drill down as proposed in https://github.com/WordPress/gutenberg/pull/61518, but this PR has hit some blockers, so for now I suggest we revert the behaviour.

## How?
We output different wrappers for the inserter depending on whether the experiment is enabled.

## Testing Instructions
1. Open the site editor
2. Open the inserter
3. Click on the canvas
4. The inserter should close
5. Open the Gutenberg > Experiments screen
6. Enable the Zoom Out experiment
1. Open the site editor
2. Open the inserter
3. Click on the canvas
4. The inserter should not close

### DIFF
--- a/packages/editor/src/components/inserter-sidebar/index.js
+++ b/packages/editor/src/components/inserter-sidebar/index.js
@@ -6,7 +6,10 @@ import {
 	__experimentalLibrary as Library,
 	store as blockEditorStore,
 } from '@wordpress/block-editor';
-import { useViewportMatch } from '@wordpress/compose';
+import {
+	useViewportMatch,
+	__experimentalUseDialog as useDialog,
+} from '@wordpress/compose';
 import { useCallback, useRef } from '@wordpress/element';
 import { store as preferencesStore } from '@wordpress/preferences';
 import { ESCAPE } from '@wordpress/keycodes';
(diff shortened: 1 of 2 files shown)


richtabor:
> This is necessary because without zoom out mode on, the pattern selector overlaps the canvas

And doesn't auto-close when inserting patterns. When zoomed out, it's great. Without zoom out, it's unexpected (as it covers content).

ellatrix:
My brain crashed at "only always" 😆

Remind me again why we completely hide zoom out when opening patterns, instead of just removing the restrictions around section selection etc.?

scruffian:
> Remind me again why we completely hide zoom out when opening patterns, instead of just removing the restrictions around section selection etc.?

The main reasons are:

1. There are a lot of bugs with the flow (you can insert patterns in to patterns, the inserters jump around, etc.)
2. We don't have a clear way forward on how to exit zoom out

draganescu (approved):
This tests OK for me 🚢 it!

ellatrix:
@scruffian I meant: why not _just_ zoom out the canvas without any other changes (when patterns is open). Same as editor mode, but just scaling the canvas.

Issue #61787, opened 2024-05-20
DataViews: The correct action menus are not displayed for patterns and template parts


t-hamano:
### Description

On the Patterns page, the DataView's action dropdown displays unexpected action menus or expected action menus are missing.

### Step-by-step reproduction instructions

The following are unnecessary actions as far as I've researched.

### Patterns bundled with themes

![image](https://github.com/WordPress/gutenberg/assets/54422211/414b678c-4b11-40de-a1cf-8cbc94d42431)

Patterns bundled with themes should only be allowed to be duplicated. The following three actions should not be displayed because they cause unintended behavior.

- Edit: The editor canvas opens, but with the template instead of the selected pattern.
- Rename: The message `Patterns bundled with themes should only be allowed to be duplicated. ` is displayed.
- Move to Trash: The action is executed, but the pattern is not removed (and of course it shouldn't be).

### User-created patterns

![image](https://github.com/WordPress/gutenberg/assets/54422211/370e38c9-6f2b-4aeb-a2c4-fc5288733d3e)

"Move to trash" is not necessary because the "Delete" menu already exists.

### User-created template parts

![image](https://github.com/WordPress/gutenberg/assets/54422211/1272e6dd-7313-4133-b09c-e8fefccbfdcb)

The following actions are missing:

- Rename
- Delete

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

jameskoster:
Just noting that the extraneous 'Move to trash' action on patterns also appears in the Inspector:

<img width="280" alt="Screenshot 2024-05-20 at 16 55 00" src="https://github.com/WordPress/gutenberg/assets/846565/c531a798-3f6f-4cfb-8eff-e627e51a06dc">

ntsekouras:
I'll take a look into this.

ntsekouras:
Closing this as I think it's fully fixed with this [last PR](https://github.com/WordPress/gutenberg/pull/61912). Feel free to reopen if I've missed something.

PR #61857, opened 2024-05-22
Consolidate and fix `rename` post action

ntsekouras:
## What?
Part of: https://github.com/WordPress/gutenberg/issues/61787

After a recent refactor to post actions some regressions were introduced, where we show more actions than we should in certain cases. An example is the `rename` post action which has different conditions based on some core post types (like templates, template parts, etc..). Additionally the goal is to use a single action for all post types where we can, like `rename, delete, etc..`.

This PR consolidates the `rename` action and adds all needed checks for core post types. In the future when we'll have an API for third party devs to register actions, it should be possible to also remove core actions too. This way they would be able to implement a custom rename action if they want..

I've changed only one action here to discuss the approach and if deemed fine, I'll update more either here or in follow ups.

## Testing Instructions
1. Check that `rename` action works as expected and is shown only when it should. For example it shouldn't be shown in theme patterns or templates or template parts. 
2. Same action is reused in inspector controls and lists view in site editor(Data Views).

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -518,9 +518,16 @@ const renamePostAction = {
 		// Make necessary checks for template parts and patterns.
 		const isTemplatePart = post.type === TEMPLATE_PART_POST_TYPE;
 		const isUserPattern = post.type === PATTERN_TYPES.user;
+		// In patterns list page we map the templates parts to a different object
+		// than the one returned from the endpoint. This is why we need to check for
+		// two props whether is custom or has a theme file.
 		const isCustomPattern =
-			isUserPattern || ( isTemplatePart && post.isCustom );
-		const hasThemeFile = isTemplatePart && post.templatePart.has_theme_file;
+			isUserPattern ||
+			( isTemplatePart &&
+				( post.isCustom || post.source === TEMPLATE_ORIGINS.custom ) );
+		const hasThemeFile =
+			isTemplatePart &&
+			( post.templatePart?.has_theme_file || post.has_theme_file );
 		return isCustomPattern && ! hasThemeFile;
 	},
 	RenderModal: ( { items, closeModal, onActionPerformed } ) => {
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -498,7 +498,30 @@ const renamePostAction = {
 	id: 'rename-post',
 	label: __( 'Rename' ),
 	isEligible( post ) {
-		return post.status !== 'trash';
+		if ( post.status === 'trash' ) {
+			return false;
+		}
+		// Templates, template parts and patterns have special checks for renaming.
+		if (
+			! [
+				TEMPLATE_POST_TYPE,
+				TEMPLATE_PART_POST_TYPE,
+				...Object.values( PATTERN_TYPES ),
+			].includes( post.type )
+		) {
+			return true;
+		}
+		// In the case of templates, we can only remove custom templates.
+		if ( post.type === TEMPLATE_POST_TYPE ) {
+			return isTemplateRemovable( post ) && post.is_custom;
+		}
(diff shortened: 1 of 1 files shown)


youknowriad:
I like this approach.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -498,7 +498,30 @@ const renamePostAction = {
...
+		// In the case of templates, we can only remove custom templates.
+		if ( post.type === TEMPLATE_POST_TYPE ) {
+			return isTemplateRemovable( post ) && post.is_custom;
+		}
+		// Make necessary checks for template parts and patterns.
+		const isTemplatePart = post.type === TEMPLATE_PART_POST_TYPE;
+		const isUserPattern = post.type === PATTERN_TYPES.user;
+		const isCustomPattern =
+			isUserPattern || ( isTemplatePart && post.isCustom );
+		const hasThemeFile = isTemplatePart && post.templatePart.has_theme_file;
```
Mamaduka:
```suggestion
		const hasThemeFile = isTemplatePart && post.has_theme_file;
```

I think this was a typo. It triggers an error when I open the template part for editing.

ntsekouras: Good catch! It seems it wasn't a typo but we change the template part object in patterns list. Fixed [here](https://github.com/WordPress/gutenberg/pull/61857/commits/9c4b3cbc3837caf049da62f8248882488e04da3c).

Mamaduka (approved):
Thanks, Nik!

The fix tests well for me ✅

Also, while testing this fix, I noticed that theme patterns (`PATTERN_TYPES.theme`) have "Edit" and "Move to Trash" actions available in DataViews. These actions do nothing and should probably only be available for user patterns.

youknowriad:
`PATTERN_TYPES.theme` as a "type" doesn't really make sense, they should all have the same "type" (aka wp_block) and have a separate flag to indicate whether they come from the theme or not.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -1196,13 +1121,12 @@ export function usePostActions( postType, onActionPerformed ) {
...
+			renamePostAction,
 			isPattern && exportPatternAsJSONAction,
-			isTemplateOrTemplatePart && resetTemplateAction,
-			! isTemplateOrTemplatePart && restorePostAction,
-			isTemplateOrTemplatePart && deleteTemplateAction,
-			! isTemplateOrTemplatePart && permanentlyDeletePostAction,
+			isTemplateOrTemplatePart ? resetTemplateAction : restorePostAction,
+			isTemplateOrTemplatePart
+				? deleteTemplateAction
+				: permanentlyDeletePostAction,
```
Mamaduka, line 1129: Sidenote: The way we build the `actions` array here is difficult to parse. Maybe we should convert it into separate if conditions.

Example:

```js
if ( condition ) {
    actions.push( action );
}
```

ntsekouras:
>PATTERN_TYPES.theme as a "type" doesn't really make sense, they should all have the same "type" (aka wp_block) and have a separate flag to indicate whether they come from the theme or not.

I agree. We should have a flag in REST response.

>Also, while testing this fix, I noticed that theme patterns (PATTERN_TYPES.theme) have "Edit" and "Move to Trash" actions available in DataViews. These actions do nothing and should probably only be available for user patterns.

Yeah, I'll handle them in follow ups.

packages/editor/src/components/post-actions/actions.js
ntsekouras: The order matters for UI for now and would need multiple `if` conditions. So I didn't do it here, even though I thought about it too. Will check it in follow ups where more actions will be consolidated.

Mamaduka: It's definitely not a blocker, just something that comes to my mind every time I see it 😄

PR #61788, opened 2024-05-20
ESLint: Enable `eslint-plugin-react-compiler`

tyxla:
## What?
This PR introduces the `eslint-plugin-react-compiler` ESLint plugin to the repo.

Even if the plugin is still beta, it will help us to have it in the repo in order to maintain a friendlier environment for React 19 and the React Compiler.

We're experimenting with adopting React Compiler through the Babel plugin in #66361.

## Why?

[It's recommended](https://react.dev/learn/react-compiler) that this ESLint plugin is used to help improve the quality of your codebase.

## How?
The `eslint-plugin-react-compiler` plugin can be used without the React Compiler itself. We're installing it and enabling it. Also ignoring a few files that include top-level `await` which seems to break the linter right now.

## Testing Instructions
Verify there are no ESLint errors and the static analysis check passes. 

### Testing Instructions for Keyboard
None

## Screenshots or screencast 
None

PR #57202, opened 2023-12-19
Tooltip: no-op when nested inside another Tooltip component

ciampo:
Requires #57345
Pre-requisite for #56490

## What?

Allow nesting 2 or more `Tooltip` components without breaking the tooltip functionality. The contents of the tooltip will be the ones defined in the outer-most `Tooltip` instance.

## Why?

In #56490, we're trying to fix [an issue](https://github.com/WordPress/gutenberg/issues/56522) where `Button` could trash and re-create its children when the internal tooltip logic flips from showing to hiding the tootlip (or vice-versa).

The solution proposed in #56490 involves always rendering (in the react sense) the `Tooltip` component inside `Button`. But doing so would cause breakages, since the `Tooltip` component doesn't support that scenario.

This PR allows that scenario to work without causing breakages.

## How?

By using the internal context system to inform instances of `Tooltip` whether they are being rendered inside another instance of `Tooltip`. When that happens, the `Tooltip` component only renders the anchor and ignores the tooltip part.

## Testing Instructions

- Test the temporary "Nested" storybook example
- Make sure unit tests pass

PR #60992, opened 2024-04-23
Components: upgrade Ariakit to latest

DaniGuardiola:

PR #61858, opened 2024-05-22
Tooltip: Fix Ariakit tooltip store usage

tyxla:
## What?
This PR fixes how we use the `Ariakit.useTooltipStore` hook in the `Tooltip` component.

## Why?
Previously, we were circumventing an ESLint error because the hook was conditionally called. I recall I suggested this workaround, so I'm glad I'm the one to suggest removing it now (see https://github.com/WordPress/gutenberg/pull/57202#discussion_r1443170682).

The workaround is no longer necessary since the hook is no longer called conditionally.

The motivation is that I'm fixing it to resolve a couple of ESLint errors that were raised by the React Compiler ESLint plugin in https://github.com/WordPress/gutenberg/pull/61788

## How?
Just using the hook directly.

## Testing Instructions
Verify the static analysis check is still green. 

None

## Screenshots or screencast 
None

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Internal
 
 -   Remove usage of deprecated spreading of `key` prop in JSX in CustomSelectControl and FormTokenField components ([#61692](https://github.com/WordPress/gutenberg/pull/61692)).
+-   Tooltip: Fix Ariakit tooltip store usage ([#61858](https://github.com/WordPress/gutenberg/pull/61858)).
 
 ## 27.6.0 (2024-05-16)
 
--- a/packages/components/src/tooltip/index.tsx
+++ b/packages/components/src/tooltip/index.tsx
@@ -87,10 +87,7 @@ function UnforwardedTooltip(
 	}
 	computedPlacement = computedPlacement || 'bottom';
 
-	// Removing the `Ariakit` namespace from the hook name allows ESLint to
-	// properly identify the hook, and apply the correct linting rules.
-	const useAriakitTooltipStore = Ariakit.useTooltipStore;
-	const tooltipStore = useAriakitTooltipStore( {
+	const tooltipStore = Ariakit.useTooltipStore( {
 		placement: computedPlacement,
 		showTimeout: delay,
 	} );


diegohaz (approved):

Mamaduka (approved):
Thanks, @tyxla!

We also need to figure out how to handle warnings/errors for Ariakit's store hooks.

```
const compositeStore = useCompositeStore();
const activeId = compositeStore.useState( 'activeId' );
```

tyxla:
> We also need to figure out how to handle warnings/errors for Ariakit's store hooks.
> 
> ```
> const compositeStore = useCompositeStore();
> const activeId = compositeStore.useState( 'activeId' );
> ```

Definitely. Might need some additional wrapper, in Ariakit or here, in the package. One simple solution I can think of is having a separate `useCompositeStoreState( store )` hook that accepts a store argument and will return the internal state of that store based on that store argument.  Might also need some name shuffling to prevent ESLint from being sad again.

tyxla:
@diegohaz @DaniGuardiola is this a problem you've encountered and/or thought about already in Ariakit?

diegohaz:
I'm not sure I'm aware of that problem. Could you elaborate?

tyxla:
We're trying out the new React Compiler ESLint plugin in https://github.com/WordPress/gutenberg/pull/61788

It will complain that hooks can't be just dynamically declared or assigned somewhere. So in code like @Mamaduka [provided above](https://github.com/WordPress/gutenberg/pull/61858#pullrequestreview-2070931233):

```
const compositeStore = useCompositeStore();
const activeId = compositeStore.useState( 'activeId' );
```

it will complain because the `useState` hook reference can potentially change, and that's against the rules of hooks - more details [here](https://react.dev/reference/rules/react-calls-components-and-hooks#dont-dynamically-use-hooks). 

To solve this issue, `useState` must be recognized by the React Compiler as a stable, static reference. My conclusion is that we might need a different API for that (as I roughly suggested above), but I'm curious if you have any other ideas.

Mamaduka:
@diegohaz, the following code triggers a warning/error when enabling the new  React compiler ESLint rules.

```
[2] /home/runner/work/gutenberg/gutenberg/packages/components/src/alignment-matrix-control/index.tsx
[2]   71:19  error  Hooks must be the same function on every render, but this value may change over time to a different function. See https://react.dev/reference/rules/react-calls-components-and-hooks#dont-dynamically-use-hooks  react-compiler/react-compiler
```

https://github.com/WordPress/gutenberg/blob/6e8c26100818b7db463e1536cc1fc5a8893de467/packages/components/src/alignment-matrix-control/index.tsx#L59-L71

diegohaz:
I see. Ariakit has an experimental `useStoreState` function (currently exported by `@ariakit/react-core/utils/store`) that can be used like this:

```js
const activeId = useStoreState(compositeStore, 'activeId');
```

This is essentially what `compositeStore.useState` does internally. There are no plans to promote this API to `@ariakit/react` soon, but it might happen if this is a general issue.

tyxla:
>  There are no plans to promote this API to @ariakit/react soon, but it might happen if this is a general issue.

I think that might be the case. Thanks for considering it. Also, cc @DaniGuardiola who's working on the Ariakit upgrade in #60992

DaniGuardiola:
@tyxla I'm not sure this PR is doing anything other than trick eslint into thinking there's no hook to analyze. The React linting tule is just too dumb to identify namespaces hooks as such, but it is still a hook and it should still respect the rules of hooks that eslint complained about.

Separately, we should wrap these hooks into ones that throw if the context is not present, so that no conditional logic is necessary (this is also what we discussed regarding Tabs).

Edit: I think I misunderstood, will check again once I'm back from lunch.

PR #61859, opened 2024-05-22
withRegistryProvider: prevent intermediate state with no children

### LINKED ISSUE #61436: Site Editor: navigating to Patterns produces and unpleasant flash effect
Steps to reproduce:
1. Open Site Editor on the home page (labeled as "Design").
2. Click on "Patterns".

There is a large white content area on the page, and during the navigation to "Patterns" it becomes all black for a moment, before a new white content area is rendered again:

https://github.com/WordPress/gutenberg/assets/664258/1749b4de-5a5e-40bc-93ef-53aa9ec48c73

This is visually very unpleasant, it's like your screen flashed.

jsnajdr:
Fixes #61436. When the `BlockEditorProvider` mounts, and creates a new data subregistry, it renders `null` for the brief moment between initial render and the effect that creates the subregistry. That creates the flash of no UI rendered reported in #61436.

This PR fixes that by creating and memoizing the subregistry directly in the render function. There are no dangerous side effects or race conditions involved.

The `withRegistryProvider` now uses `useRegistry` instead of the `withRegistry` HOC, and correctly reacts to all changes in props: both `useSubRegistry` and `registry` changes lead to passing the correct subregistry.

I'm also fixing the `EditorProvider` component, which also creates a subregistry, just with different stores.

**How to test:**
Verify that the flash reported in #61436 disappears when navigating to Patterns.

### DIFF
--- a/packages/block-editor/src/components/provider/with-registry-provider.js
+++ b/packages/block-editor/src/components/provider/with-registry-provider.js
@@ -1,12 +1,8 @@
 /**
  * WordPress dependencies
  */
-import { useState, useEffect } from '@wordpress/element';
-import {
-	withRegistry,
-	createRegistry,
-	RegistryProvider,
-} from '@wordpress/data';
+import { useState } from '@wordpress/element';
+import { useRegistry, createRegistry, RegistryProvider } from '@wordpress/data';
 import { createHigherOrderComponent } from '@wordpress/compose';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/components/provider/with-registry-provider.js
```diff
@@ -15,41 +11,40 @@ import { createHigherOrderComponent } from '@wordpress/compose';
...
+}
 
-				if ( ! subRegistry ) {
-					return null;
-				}
+const withRegistryProvider = createHigherOrderComponent(
+	( WrappedComponent ) =>
+		( { useSubRegistry = true, ...props } ) => {
+			const registry = useRegistry();
+			const [ subRegistries ] = useState( () => new WeakMap() );
```
tyxla, line 31: Nice to see the `useState` [pattern for one-time init](https://tkdodo.eu/blog/use-state-for-one-time-initializations) more often recently 👍

youknowriad (approved):
Oh nice one :)
Confirmed that it feels better.

tyxla (approved):
Nice solution 👍 

Fixes the screen flash issue too.

🚀

Mamaduka:
This is a nice fix, indeed! Thanks, @jsnajdr!

Issue #52632, opened 2023-07-14
Align page edit features in Site Editor and Post Editor


jameskoster:
There are a number of smaller edit actions that are available in the post editor, but not the site editor. The mis-alignment means users sometimes need to hop between editors, which can be frustrating.

In addition to the Page Inspector, we should consider the new details panel, and the Command Palette when working out the placement for these items.

- [x] Access to revisions: https://github.com/WordPress/gutenberg/issues/51343
- [x] Switch page template: https://github.com/WordPress/gutenberg/issues/51347
- [x] Create new template while editing the page.
- [x] https://github.com/WordPress/gutenberg/issues/52763
- [x] Make pages trash-able in the full screen editor.
- [x] https://github.com/WordPress/gutenberg/issues/53188
- [x] Page attributes (parent page, order).
- [x] Edit slug.
- [x] Change Author.
- [x] Unify preferences
  - [x] right click override #57468
  - [x] top toolbar #57531
  - [x] spotlight mode #57533
  - [x] whether the list view opens by default #57504
  - [x] show/hide block breadcrumbs #57506
  - [x] distraction free #57590
  - [x] which panels to show on the sidebar #57529
  - [x] showing button text labels #57480
  - [x] constrain text cursor #57479
  - [x] show most used blocks in the inserter #57637
  - [x] block visibility #57639
  - [x] default editor (code/visual) #57642
- [x] Unify commands.
- [x] Unify publishing flow #59878

youknowriad:
Hello folks, so I've been thinking a little bit about it and shared some comments here https://github.com/WordPress/gutenberg/pull/55844#issuecomment-1794318027

I think we've been going about this in the wrong way. We're duplicating more and more components while we could unify both post editor and site editor to use the same code and share components through the existing `@wordpress/editor` package and ultimately just use the exact same "editor" entirely.

I'm going to start working on some of this and my plan is the following:

 - Use the EditorProvider in the site editor (to set the correct postType/postId...)
 - Move the UI components one by one from the site editor to the @wordpress/editor package. (which might allow us to remove duplicated components)
 - Move the shell of the editor (probably the most complex task) as it requires thinking about extensibilty/preferences...
 - Final step: Once all UI components have moved, remove the duplicated state from the site-editor and store and proxy the selectors/actions.

jameskoster:
Should we close this issue and work on refining the document inspector design in the post editor instead?

youknowriad:
[...]

PR #61860, opened 2024-05-22
Editor: Unify the content area of the post and site editors

youknowriad:
Related #52632 

## What?

This is just a small PR that moves some components/elements around in order to prepare for a future potential unification of the whole "content" area of the editor skeleton between post and site editors.

## Testing Instructions

There should be no functional change here.

### DIFF
--- a/packages/edit-post/src/components/layout/index.js
+++ b/packages/edit-post/src/components/layout/index.js
@@ -12,7 +12,6 @@ import {
 	UnsavedChangesWarning,
 	EditorNotices,
 	EditorKeyboardShortcutsRegister,
-	EditorKeyboardShortcuts,
 	EditorSnackbars,
 	store as editorStore,
 	privateApis as editorPrivateApis,
@@ -24,7 +23,6 @@ import {
 	privateApis as blockEditorPrivateApis,
 	store as blockEditorStore,
 } from '@wordpress/block-editor';
-import { ScrollLock } from '@wordpress/components';
 import { useViewportMatch } from '@wordpress/compose';
 import { PluginArea } from '@wordpress/plugins';
 import { __, _x, sprintf } from '@wordpress/i18n';
@@ -335,7 +333,6 @@ function Layout( { initialPost } ) {
 			<LocalAutosaveMonitor />
 			<EditPostKeyboardShortcuts />
 			<EditorKeyboardShortcutsRegister />
-			<EditorKeyboardShortcuts />
 			<BlockKeyboardShortcuts />
 
 			<InterfaceSkeleton
@@ -367,7 +364,9 @@ function Layout( { initialPost } ) {
 						{ ( mode === 'text' || ! isRichEditingEnabled ) && (
 							<TextEditor />
 						) }
-						{ ! isLargeViewport && <BlockToolbar hideDragHandle /> }
+						{ ! isLargeViewport && mode === 'visual' && (
+							<BlockToolbar hideDragHandle />
+						) }
 						{ isRichEditingEnabled && mode === 'visual' && (
 							<VisualEditor styles={ styles } />
 						) }
@@ -376,9 +375,6 @@ function Layout( { initialPost } ) {
 								<MetaBoxes location="normal" />
 								<MetaBoxes location="advanced" />
 							</div>
-						) }
-						{ isMobileViewport && sidebarIsOpened && (
-							<ScrollLock />
 						) }
 					</>
 				}
--- a/packages/edit-site/src/components/editor/index.js
+++ b/packages/edit-site/src/components/editor/index.js
@@ -25,7 +25,6 @@ import {
 } from '@wordpress/block-editor';
 import {
 	EditorKeyboardShortcutsRegister,
-	EditorKeyboardShortcuts,
 	EditorNotices,
 	privateApis as editorPrivateApis,
 	store as editorStore,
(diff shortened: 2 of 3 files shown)


packages/edit-post/src/components/layout/index.js
```diff
@@ -367,7 +364,9 @@ function Layout( { initialPost } ) {
 						{ ( mode === 'text' || ! isRichEditingEnabled ) && (
 							<TextEditor />
 						) }
-						{ ! isLargeViewport && <BlockToolbar hideDragHandle /> }
+						{ ! isLargeViewport && mode === 'visual' && (
```
youknowriad, line 367: Extra check to align with site editor.

packages/edit-post/src/components/layout/index.js
```diff
@@ -377,9 +376,6 @@ function Layout( { initialPost } ) {
 								<MetaBoxes location="advanced" />
 							</div>
 						) }
-						{ isMobileViewport && sidebarIsOpened && (
-							<ScrollLock />
```
youknowriad, line 381: Seems like component was added a long time ago for mobile but it's not present in the site editor and I'm not sure why we need it. Maybe @jasmussen or @ellatrix would know if this removal is impactful or not.

packages/edit-site/src/components/editor/index.js
```diff
@@ -265,6 +264,15 @@ export default function Editor( { isLoading, onClick } ) {
...
 		<>
+			<GlobalStylesRenderer />
+			<EditorKeyboardShortcutsRegister />
+			{ isEditMode && <BlockKeyboardShortcuts /> }
+			{ showVisualEditor && (
+				<>
+					<TemplatePartConverter />
+					<PatternModal />
+				</>
+			) }
```
youknowriad, line 275: Most of these don't need to be in the "content" area. So I just moved them to align the content area with the post editor.

packages/edit-site/src/components/editor/index.js
```diff
@@ -265,6 +264,15 @@ export default function Editor( { isLoading, onClick } ) {
...
 	return (
 		<>
+			<GlobalStylesRenderer />
+			<EditorKeyboardShortcutsRegister />
+			{ isEditMode && <BlockKeyboardShortcuts /> }
+			{ showVisualEditor && (
+				<>
+					<TemplatePartConverter />
+					<PatternModal />
+				</>
```
youknowriad, line 274: I think both of these template part converter and pattern modal components need to move to the editor package. They provide features that are site editor only for now for no obvious reasons.

packages/editor/src/components/provider/index.js
```diff
@@ -273,6 +274,7 @@ export const ExperimentalEditorProvider = withRegistryProvider(
 									{ type === 'wp_navigation' && (
 										<NavigationBlockEditingMode />
 									) }
+									<EditorKeyboardShortcuts />
```
ntsekouras, line 277: Should we also move `EditorKeyboardShortcutsRegister`?

youknowriad: No, because we want the shortcuts to be registered regardless of whether they are actually applied. Registering the shortcuts allow them to be visible in the keyboard shortcuts modal.

packages/edit-site/src/components/editor/index.js
ntsekouras: `PatternModal` can definitely move IMO and also move the commands that use them. The template part converter though is probably because we do not allow inserting template parts in posts. Now that both editors are converging, I'm curious if the decision to exclude template parts in post editor should be revisited too..

youknowriad: We do allow template parts in the template editor which is also available in the post editor. So it should move too.

For the pattern modal, I'm moving it here #61862
For the template part converter, I'm waiting for @jorgefilipecosta's PR that moves the template part creation modal first.

packages/editor/src/components/provider/index.js
ntsekouras: I still don't get how it would affect the modal if we moved them in Editor provider. Will do some testing to verify 😄 

--edit. I tested and it works for me. Maybe I'm testing a different thing? I moved the register in `editor/EditorProvider` and shortcuts work..

ntsekouras (approved):
Besides the open question about `ScrollLock`, this LGTM. Thanks!

packages/editor/src/components/provider/index.js
youknowriad: It works right now because the keyboard shortcuts modal is rendered only when the editor is rendered but in theory the keyboard shortcuts modal should be accessible every time (even in the site editor shell).

I think it's something we can revisit though and the separation between registration and availability in the modal can be reconsidered.

youknowriad: If we ever reconsider, I think we should just unify the components/registration...

jasmussen:
There appears to be a small issue with this PR. The root padding broke, somehow:

after

I found that out by checking the previous commit in https://github.com/WordPress/gutenberg/commits/trunk/?before=f49e0b5958adad9fbd211d5ded07b4cabf5bfec3+51, and that one is fine:

before

youknowriad:
Thanks for catching that @jasmussen I'll take a look.

youknowriad:
@jasmussen Fix here #61906

packages/edit-site/src/components/editor/index.js
jorgefilipecosta: Hi @youknowriad, that PR is available at https://github.com/WordPress/gutenberg/pull/61879.

PR #61862, opened 2024-05-22
Commands: Unify the editor context between post and site editors

youknowriad:
Related #52632 

## What?

This PR moves the "PatternModal" component and their corresponding commands to the editor package rather than leaving this as a site editor only feature. This means that these two commands will be available to folks still loading patterns in the post editor.

It also unifies the commands context between post and site editors opening the door for more cross-editor entity commands.

## Testing Instructions

There should be no functional change.
You should see the "duplicate" and "rename" commands when opening patterns in the editor.


ntsekouras (approved):
LGTM, thanks!

PR #61863, opened 2024-05-22
Global Styles: Move type presets below elements

scruffian:
## What?
Moves the type presets lower in the global styles sidebar.

## Why?

It is possible for there to be a lot presets - having them at the top would push down all the other settings, so it's better to have them lower so that they can scroll.

## How?
Just move the component.

## Testing Instructions
0. Using TT4...
1. Open the Site Editor
2. Open Global Styles
3. Open Typography
4. Check that the presets are at the bottom of the list.

## Screenshots or screencast 
|Before|After|
|---|----|
|Screenshot 2024-05-22 at 14 11 44|Screenshot 2024-05-22 at 14 10 46|


MaggieCabrera (approved):
This is looking good to me

Issue #43743, opened 2022-08-31
Blocks: Add support for block name aliases for variations


gziolo:
## What problem does this address?

Part of Block API roadmap #41236.

Let's explore whether some block variations can be read as a block type. For example, instead of `core/template-part/header` be able to just have `core/header` to still get resolved by the template part functions. We used to have a similar granular naming schema for Embed and Social Link blocks before we converted them to block variations:

https://github.com/WordPress/gutenberg/blob/bdb6a41099529541ce1f35d05037fef066674150/packages/blocks/src/api/parser/convert-legacy-block.js#L23-L29

https://github.com/WordPress/gutenberg/blob/bdb6a41099529541ce1f35d05037fef066674150/packages/blocks/src/api/parser/convert-legacy-block.js#L31-L50

### Benefits of using aliases

- Aliased block names would get converted into unique class names assigned to block variations, giving more options to style their sites.
- Manipulate in `theme.json` granularly (like attach custom CSS that only is used when the block is used, etc.).
- Aliases used would give easier access for site owners to the content analysis of which blocks and variations are used in posts and templates.
- Benefit from all the block API tools in existence (like [`multiple`](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#multiple) block support).

## What is your proposed solution?

Add handling for block type name aliases used with block variations. They should fall back if necessary to the original block type name, e.g. `woocommerce/product-list` that can get resolved to the `product-list` variation added to the `core/query` block.

Some examples to consider:

```html
<!-- wp:template-part {"slug":"header","theme":"twentytwenty"} /-->
<!-- wp:meta { "name": "mykey", "type": "heading" } /-->
```

becomes

```html
<!-- wp:template-part/header {"theme":"twentytwenty"} /-->
<!-- wp:meta/mykey { "type": "heading" } /-->
```

This could also remove the need for complex logic used with block variations when detecting the variation based on the result of the `isActive` callback executed.

louwie17:
Hey @gziolo has there been any more progress around this idea/issue?
I think this would be a great feature to have and may really benefit us in WooCommerce when laying out the template for the new product editor.
We currently make active use of attributes and re-use blocks quite often, I think something like this would offer two big benefits for us:
- Clean up the actual template by being able to remove a lot of our attribute values
- Make better use of block hooks, as this will allow 3rd party plugins to hook into variations*

*This will depend if the above proposed solution will also work with block hooks.

A current example would be:
[...]

PR #56469, opened 2023-11-23
Add block class name to the list block

carolinan:
## What?

Adds the `wp-block-list` class to the list block
For https://github.com/WordPress/gutenberg/issues/12420

## Why?

The missing class makes it difficult to style the block separately from other lists.

## How?

List:
- Removes `"className": false,` from block.json.
- Removes the "`__experimentalSelector": "ol,ul",` from block.json so that the styles are applied to `.wp-block-list` instead of plain ul and ol.

List item:
- Changes the selector for the list-item block to `.wp-block-list > li`

## Testing Instructions
Activate a block theme.

Open theme.json in your code editor and add a visible style to the list and list-item blocks, for example a background color.
If you are using Twenty Twenty-Four, the theme already has a styles for the list block, find them on line 515 in theme.json to avoid validation errors.

Example:
```
			"core/list": {
				"color": {
					"background": "pink"
				},
				"spacing": {
					"padding": {
						"left": "var(--wp--preset--spacing--10)"
					}
				}
			},
			"core/list-item":{
				"color": {
					"background": "yellow"
				}
			},
```

Create a new post or page.
Add an HTML block with a list inside.
Example:
```
<!-- wp:html -->
<ul><li>hello</li><li>world</li></ul>
<!-- /wp:html -->
```
Select preview from the HTML block toolbar and confirm that the list does not have a background color.

Add a list block with some inner list items.
- Confirm that the block background is pink and the list item background is yellow.
- View the page source and confirm that the `wp-list-block` is present on the block.
- Open the code editor mode and confirm that the `wp-list-block` is present on the block.

Save, and view the front of the website.

- Confirm that the list in the HTML block does not have a background color.
- Confirm that the list block has a pink background, and list item has a yellow background.
- View the page source and confirm that the wp-list-block is present on the block.

Return to the editor and style the blocks via the settings in the sidebar; adjust color, typography and spacing. Confirm that the custom styles are applied correctly in the editor and on the front.

Bonus:
If you are using Twenty Twenty-Four, you can also confirm that the navigation block (which uses a list element) in the theme's header and footer does not have a pink background.

Issue #12420, opened 2018-11-29
Add class to <ul> in List block.


pagelab:
**Is your feature request related to a problem? Please describe.**

The subject of adding classes to default blocks was discussed at [#6639](https://github.com/WordPress/gutenberg/issues/6639), and I know [you can filter the List block to add a new class](https://wordpress.org/gutenberg/handbook/designers-developers/developers/filters/block-filters/), but the `wp-block-list` class should be present by default.

Without a specific class, any style applied to the generic `<ul>` will inadvertently affect elements from third-party plugins or custom HTML added elsewhere, and we can't know beforehand if they will be affected.

One practical example happens when you try, for instance, to style the List block in a theme to add new bullets and you are also using the Elementor plugin.

Without a class, the generic `.ul` style will affect `<li>` elements used by Elementor in their control handles – unless you explicitly filter the class or resort to a convoluted solution as `ul:not(.elementor-editor-element-settings):not(.wp-block-gallery):not(.wp-block-categories):not(.wp-block-latest-posts)`... and so on.

This will surely happen with lots of other plugins that use the `<ul>` tag for tasks like this.

**Image 1:** Custom style added to `<ul>` tag.

![image](https://user-images.githubusercontent.com/1451087/49225833-37cf2380-f3cc-11e8-9a02-44841590160a.png)

**Image 2:** Custom style affecting a third-party plugin.

![image](https://user-images.githubusercontent.com/1451087/49225770-0d7d6600-f3cc-11e8-9ea6-dee3f430f179.png)

**Describe the solution you'd like**
Just add the `wp-block-list` class to the `<ul>` tag by default, in the List block.

**Describe alternatives you've considered**
* [Filter the List block to add a new class](https://wordpress.org/gutenberg/handbook/designers-developers/developers/filters/block-filters/)
* Add multiple `:not()` to the generic `.ul` style to target only the List block. 

Tested with 5.0-RC1-43946

swissspidy:
Note that using non-specific selectors like `ul` is a bad idea in general. I'm sure you could use something like `.content-area ul` in your theme.

pagelab:
That was just an example. 

The problem would persist if you utilize `.content-area ul.` – or other more specific selector.

The only way to solve the issue is to add a class to the `<ul>` tag itself.

designsimply:
[...]

PR #61864, opened 2024-05-22
Block variations: Add block-supports flag to add variation specific classname

tjcafferkey:
Related to: https://github.com/WordPress/gutenberg/issues/43743

## What?
For a given variation of a static block that is registered it will add a unique className specific to the variation to the block markup. For example [`group-row`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/group/variations.js#L22-L33) which is a variant of `core/group` would have the resulting CSS className: `wp-block-group wp-block-group-group-row`.

## Why?
As part of the project to support block aliases for variations ([tracking issue](https://github.com/WordPress/gutenberg/issues/43743)) one of the requirements is to provide an additional CSS className to provide more styling options for these blocks.

## How?

We chose to update the existing `className` block-support flag. Previously, setting `className: true` would cause the default block classname (`wp-block-my-block`) to be added. This continues to be the default value and behavior for that flag.

Additionally, the following syntax is now supported:
```js
className: {
    block: true,
    variation: true,
}
```

The `block` property controls adding the default block classname, whereas the `variation` property is in charge of adding the variation-specific classname (`wp-block-my-variation`).

**Question:** Are we happy with the choice of names for these properties (`block`, `variation`)? Should they be called differently?

## Testing Instructions

* Apply the patch below, which will opt the Group block into variation classname block support.
* Using the TT4 theme, go to the Site Editor.
* In the DevTools console, verify that deprecations have been run for all instances of the Group block, and that they've injected the new variation-specific classnames. (See screenshot at the bottom of the PR desc.)
* In the Code Editor, verify that all Group blocks have a classname added for their respective variation (e.g. `wp-block-group-group-row`).
* Save the template (to persist changes made by block deprecations).
* Reload. Verify that no blocks show up as invalid.
* View the frontend.
* In the developer tools, verify that the classname is present on the frontend as well.
* Go to the post editor and insert a few new Group blocks. Use different variations, and verify in the Code Editor that each of them gets the correct variation classname added.

<details>
<summary>
Patch
</summary>

```diff
diff --git a/packages/block-library/src/group/block.json b/packages/block-library/src/group/block.json
index 3c7d8d3ce13..bb958b0cce6 100644
--- a/packages/block-library/src/group/block.json
+++ b/packages/block-library/src/group/block.json
@@ -91,6 +91,9 @@
 		},
 		"interactivity": {
 			"clientNavigation": true
+		},
+		"className": {
+			"variation": true
 		}
 	},
 	"editorStyle": "wp-block-group-editor",
diff --git a/packages/block-library/src/group/deprecated.js b/packages/block-library/src/group/deprecated.js
index fcc2b249d22..b97af894eeb 100644
--- a/packages/block-library/src/group/deprecated.js
+++ b/packages/block-library/src/group/deprecated.js
@@ -42,6 +42,100 @@ const migrateAttributes = ( attributes ) => {
 };
 
 const deprecated = [
+	// Version without block support 'className: { variation: true }'.
+	{
+		attributes: {
+			tagName: {
+				type: 'string',
+				default: 'div',
+			},
+			templateLock: {
+				type: [ 'string', 'boolean' ],
+				enum: [ 'all', 'insert', 'contentOnly', false ],
+			},
+			allowedBlocks: {
+				type: 'array',
+			},
+		},
+		supports: {
+			__experimentalOnEnter: true,
+			__experimentalOnMerge: true,
+			__experimentalSettings: true,
+			align: [ 'wide', 'full' ],
+			anchor: true,
+			ariaLabel: true,
+			html: false,
+			background: {
+				backgroundImage: true,
+				backgroundSize: true,
+				__experimentalDefaultControls: {
+					backgroundImage: true,
+				},
+			},
+			color: {
+				gradients: true,
+				heading: true,
+				button: true,
+				link: true,
+				__experimentalDefaultControls: {
+					background: true,
+					text: true,
+				},
+			},
+			shadow: true,
+			spacing: {
+				margin: [ 'top', 'bottom' ],
+				padding: true,
+				blockGap: true,
+				__experimentalDefaultControls: {
+					padding: true,
+					blockGap: true,
+				},
+			},
+			dimensions: {
+				minHeight: true,
+			},
+			__experimentalBorder: {
+				color: true,
+				radius: true,
+				style: true,
+				width: true,
+				__experimentalDefaultControls: {
+					color: true,
+					radius: true,
+					style: true,
+					width: true,
+				},
+			},
+			position: {
+				sticky: true,
+			},
+			typography: {
+				fontSize: true,
+				lineHeight: true,
+				__experimentalFontFamily: true,
+				__experimentalFontWeight: true,
+				__experimentalFontStyle: true,
+				__experimentalTextTransform: true,
+				__experimentalTextDecoration: true,
+				__experimentalLetterSpacing: true,
+				__experimentalDefaultControls: {
+					fontSize: true,
+				},
+			},
+			layout: {
+				allowSizingOnChildren: true,
+			},
+			interactivity: {
+				clientNavigation: true,
+			},
+		},
+		save( { attributes: { tagName: Tag } } ) {
+			return (
+				<Tag { ...useInnerBlocksProps.save( useBlockProps.save() ) } />
+			);
+		},
+	},
 	// Version with default layout.
 	{
 		attributes: {
```
</details>

## Screenshot

image


packages/blocks/src/api/serializer.js
```diff
@@ -41,7 +41,7 @@ export function getBlockDefaultClassName( blockName ) {
 	// Generated HTML classes for blocks follow the `wp-block-{name}` nomenclature.
 	// Blocks provided by WordPress drop the prefixes 'core/' or 'core-' (historically used in 'core-embed/').
 	const className =
-		'wp-block-' + blockName.replace( /\//, '-' ).replace( /^core-/, '' );
+		'wp-block-' + blockName.replace( /\//g, '-' ).replace( /^core-/, '' );
```
tjcafferkey: This was previously only replacing the first '/' with a '-' in the block name. So `namespace/block-name/variation-name` would result in `wp-block-namespace-block-name/variation-name`.

This is different to how its PHP counterpart function (`wp_get_block_default_classname()`) behaves which replaces all occurrences in the string. This change brings consistency with the PHP function.

tjcafferkey: An alternative to this change would be to alter the `blockName` we pass in to this function _knowing_ that it's only replacing the first `/`

ockham (changes requested):
I think the code that adds the variation class name should be colocated with the code that adds the auto-generated block class name 🤔 The latter is only added conditionally to the block [if it has `className` block-supports](https://github.com/WordPress/gutenberg/blob/58b454c05172c51a023e36ff3b7078ee63086e09/packages/block-editor/src/hooks/generated-class-name.js#L19).

The relevant code is encapsulated in `generated-class-name.js`, so it looks like adding the variation class name there would also preserve parity with [what we're doing on the server](https://github.com/WordPress/wordpress-develop/pull/6602).

ntsekouras:
>I think the code that adds the variation class name should be colocated with the code that adds the auto-generated block class name 🤔 

I'd echo that. We should inject the class name and it should be auto generated. We shouldn't preserve the className in the html markup as it will create various issues if we switch to a different variation.

tjcafferkey:
>The relevant code is encapsulated in generated-class-name.js, so it looks like adding the variation class name there would also preserve parity with https://github.com/WordPress/wordpress-develop/pull/6602.

I agree with colocating the code, but AFAIK the code in `generated-class-name.js` is only run for blocks with a `save` function. If we take `core/social-link` as an example this isn't the case. I'll keep digging.

ockham:
👋 I've only now started reviewing this, and I think it'll take me a bit longer to let it all sink in.

Coincidentally, I came across #56469 today, which changed the List block -- which previously had `className` block-support set to `false`, and thus no `wp-block-list` class added -- enable `className` block support, and thus add the missing class name.

That PR is interesting -- and potentially relevant to the problem we're trying to solve.
It includes a [deprecation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/) for [the previous version of the List block](https://github.com/WordPress/gutenberg/pull/56469/files#diff-1aa2274435a9d35d6102b9ab016ac0747cdbd7f1947412c1cb1e14438fafbeea); these are run in the editor and are the established way to avoid parse errors for blocks whose attributes have changed. I just didn't think of them as a potential solution, since they're always block-specific -- there's no block deprecation mechanism that I'm aware of that would apply the same kind of deprecation to a whole class of blocks. I do like that this approach uses an established mechanism to make the necessary changes to existing blocks. 🤔 

More surprisingly to me, the PR adds a server-side [`render_callback` that injects the `wp-block-list` class into the `ul` or `ol` wrapper, respectively](https://github.com/WordPress/gutenberg/pull/56469/files#diff-a4183240aa6b9f8c06c5f32967febdad97df2b17ee01b89c5c543e2b4b6da1bb). The rationale for this is found in a code comment: https://github.com/WordPress/gutenberg/blob/d80b5a96f297629c832a7b59ee5ffcc68d67eb92/packages/block-library/src/list/index.php#L9-L11

...and possibly in a [prior discussion on a GH issue](https://github.com/WordPress/gutenberg/issues/12420#issuecomment-2039736605):

> I think @mtias's idea was to add it to both the frontend and the editor but to not serialize it in the saved markup. I think it's a given that the markup of the editor should match the frontend.

Some ideas from that other PR might help us reduce some of the more "magical" parts from our approach. Anyway, I'll need a bit more time to fully wrap my head around this.

ockham:
Another observation: If I revert 83ddbfb (i.e. #56469), the `wp-block-list` class is indeed not added on the frontend, but it _is_ present in the editor. (I wonder if historically, we first introduced the `wp-block-` classnames in the editor to be able to style blocks there, and only later carried the classname over to the frontend as it turned out handy to have?)

| Frontend | Editor |
|----------|-------|
| image | image |

packages/block-editor/src/components/block-edit/edit.js
```diff
@@ -75,8 +76,22 @@ const EditWithGeneratedProps = ( props ) => {
...
+
+	// Generate a class name for the block's editable form.
+	const generatedVariationClassName =
+		hasBlockSupport( blockType, 'className', true ) && activeVariation
+			? getBlockDefaultClassName( `${ name }/${ activeVariation.name }` )
+			: null;
+
 	const className = clsx(
 		generatedClassName,
+		generatedVariationClassName,
```
ntsekouras: In my mind this should be part of `getBlockDefaultClassName`. Is there a reason for not adding the logic in that function?

tjcafferkey: I think my initial approach did involve doing this, but IIRC @ockham preferred to keep them separated which I can also see an argument for. Not sure if you'd prefer to elaborate on this @ockham :)

ockham: I'm thinking `getBlockDefaultClassName` is used in a number of places where the expectation is that it only returns _one_ class name. For example, the [documentation of the corresponding filter](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#blocks-getblockdefaultclassname) states

> Generated HTML classes for blocks follow the `wp-block-{name}` nomenclature. This filter allows to provide an alternative class name.

Now it might be possible to change that, but I felt like there's too much potential for breakage of existing code. Instead, I thought it might make sense to keep `getBlockDefaultClassName` as the low-level util that generates the default `wp-block-` classname that's used by higher-level functions such as `addGeneratedClassName` to compose the final list of class names, ~and add `generatedVariationClassName` as another low-level util for use by the latter.~

_Edit:_ It might make sense to add a new `getBlockVariationDefaultClassName` util.

ockham (dismissed):

ockham:
I'll rebase.

packages/block-editor/src/components/block-edit/edit.js
ntsekouras: Yeah,  I get what you mean, but still in my mind we are adding an extra default classname for every block. By having them separated it means we'd have to duplicate the checks for it, new similar filter etc.. I'd love some more thoughts maybe from @mcsf or @youknowriad ?

packages/blocks/src/api/serializer.js
mcsf: Given the discrepancy with the PHP counterpart, I think adding the `/g` flag is the best thing to do.

Otherwise I might've suggested encoding the variation as a new argument. Block names are actually [well defined](https://github.com/WordPress/gutenberg/blob/bd6e764e1ba988d35ab17b5cfcb678715051fdd0/packages/blocks/src/api/registration.js#L232-L237) (`namespace-with-hyphens/name-with-hyphens`), so I worried about the possibility of these variation-aware names (`namespace/block/variation`) subtly breaking things.

packages/block-editor/src/hooks/generated-variation-class-name.js
```diff
@@ -0,0 +1,66 @@
...
+			extraProps.className = getBlockDefaultClassName( variationName );
+		}
+	}
+
+	return extraProps;
+}
+
+addFilter(
+	'blocks.getSaveContent.extraProps',
+	'core/generated-class-name/save-props',
```
mcsf:
```suggestion
	'core/generated-variation-class-name/save-props',
```

packages/block-editor/src/hooks/generated-variation-class-name.js
```diff
@@ -0,0 +1,66 @@
...
+		if ( ! activeVariation ) {
+			return extraProps;
+		}
+
+		const variationName = `${ blockType.name }/${ activeVariation.name }`;
+
+		if ( typeof extraProps.className === 'string' ) {
+			// We have some extra classes and want to add the default classname
+			// We use uniq to prevent duplicate classnames.
+
```
mcsf:
```suggestion
```

packages/block-editor/src/components/block-edit/edit.js
mcsf: I think it's fine to **not** fit everything into `getBlockDefaultClassName` (but ultimately I don't care much). 

However, here's my nit: I would consolidate the support check. It's not just saving a function call, but more importantly I think it reads better:

```js
// Generate class names for the block's editable form.
const generatedNames = [];
if ( hasBlockSupport( blockType, 'className', true ) ) {
	generatedNames.push( getBlockDefaultClassName( name ) );
	if ( activeVariation ) {
		generatedNames.push(
			getBlockDefaultClassName(
				`${ name }/${ activeVariation.name }`
			)
		);
	}
}

const className = clsx(
	generatedNames,
	attributes.className,
	props.className
);
```

tjcafferkey:
Just noting some thoughts / discoveries down here:

* If we load a template and the markup doesn't already have the variation classNames on the blocks, they will not be present if you switch to Code Editor. But they will pass block validation.
    * Making an update (e.g. hitting return to line down) in the Visual Editor tab will force the variation classNames to get added to all variations in the Code Editor
* These variation classNames won't be visible in the Visual Editor if you were to inspect the DOM. (Fixed in https://github.com/WordPress/gutenberg/pull/61864/commits/0a12f45c9030fc69f1635c77c21c0c8dcfe841e9)
    * **Note**: This is important because if there is some CSS associated with the variation className it wont be applied in the Visual Editor, but it will be applied on the frontend if saved, this means theres not visual parity between the two.

ockham:
> Just noting some thoughts / discoveries down here:
> 
> * If we load a template from the database or filesystem and the markup doesn't already have the variation classNames on the blocks, they will not get added. But they will pass block validation.
> * Making an update (e.g. hitting return to line down) in the Visual Editor tab will force the variation classNames to get added to all variations in the Code Editor in both cases (filesystem / database).

This means that any modification and subsequent saving of the template will cause the variation className to be added, right? If that's the case, then it's pretty much the behavior I'd expect 👍 

> * These variation classNames won't be visible in the Visual Editor if you were to inspect the DOM. (Fixed in [0a12f45](https://github.com/WordPress/gutenberg/commit/0a12f45c9030fc69f1635c77c21c0c8dcfe841e9))
>   * **Note**: This is important because if there is some CSS associated with the variation className it wont be applied in the Visual Editor, but it will be applied on the frontend if saved, this means theres not visual parity between the two.

I understand that's fixed now? 😄

tjcafferkey:
When a template is loaded into the Visual Editor, variation classNames are added to existing static blocks (visible through DevTools by inspecting the DOM). However, if you switch immediately to the Code Editor, the block markup will appear without these classNames. Making any change in the Visual Editor will then force the variation classNames to appear in the Code Editor. These new classNames will not show on the frontend until you save your template.

Initially, I thought this discrepancy between the Visual and Code Editors needed fixing, but this behavior is already present in the deprecation/migration API.

For example, when [deprecating a block to change its tag name](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/), the Visual Editor shows the new tag name in the DOM, while the old tag name appears in the Code Editor until the user makes a change and saves the post/template.

packages/blocks/src/api/serializer.js
ockham: Echoing what Miguel said -- and given the uncertain future of the `/variation` suffix, which seems to be somewhat controversial -- let's maybe revert the `/g` change and instead append the `-variation` manually to the blockname whenever we need it.

packages/block-editor/src/components/block-edit/edit.js
```diff
@@ -71,12 +72,23 @@ const EditWithGeneratedProps = ( props ) => {
...
+
+	if ( hasBlockSupport( blockType, 'className', true ) ) {
+		generatedClassNames.push( getBlockDefaultClassName( name ) );
+
+		const activeVariation = getActiveBlockVariation( name, attributes );
+		if ( activeVariation && activeVariation?.name ) {
+			generatedClassNames.push(
+				getBlockDefaultClassName(
+					`${ name }/${ activeVariation.name }`
+				)
```
ockham: Per https://github.com/WordPress/gutenberg/pull/61864/files#r1655059595, let's maybe change this to

```suggestion
				getBlockDefaultClassName( name ) + `-${ activeVariation.name }`
```

(There are probably a few more instances of this in this PR.)

packages/block-editor/src/components/block-edit/test/edit.js
```diff
@@ -84,6 +86,47 @@ describe( 'Edit', () => {
...
+		registerBlockVariation( 'core/test-block', {
+			name: 'variation',
+			title: 'block variation title',
+			attributes: {
+				fruit: {
+					type: 'string',
+					default: 'Bananas',
+				},
+			},
+			isActive: ( { fruit } ) => fruit === 'Bananas',
```
ockham: Nitpick, I think it's possible to [define variations during block registration](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#creating-a-block-variation), so it should be possible to drop the extra `registerBlockVariation` call (and hopefully also the `unregister`?)

```suggestion
		variations: [
			{
				name: 'variation',
				title: 'block variation title',
				attributes: {
					fruit: {
						type: 'string',
						default: 'Bananas',
					},
				},
				isActive: ( { fruit } ) => fruit === 'Bananas',
			}
		],
```

ockham: (Not too important. If you think the separate registration reads better, feel free to keep it.)

packages/block-editor/src/hooks/generated-variation-class-name.js
```diff
@@ -0,0 +1,64 @@
...
+import {
+	hasBlockSupport,
+	getBlockDefaultClassName,
+	getActiveBlockVariation,
+} from '@wordpress/blocks';
+
+/**
+ * Override props assigned to save component to inject generated className if
+ * block supports it. This is only applied if the block's save result is an
+ * element and not a markup string.
```
ockham:
```suggestion
 * Override props assigned to save component to inject generated variation
 * className if block supports it. This is only applied if the block's save
 * result is an element and not a markup string.
```

packages/block-editor/src/components/block-edit/edit.js
```diff
@@ -71,12 +72,23 @@ const EditWithGeneratedProps = ( props ) => {
 		return <EditWithFilters { ...props } context={ context } />;
 	}
 
-	// Generate a class name for the block's editable form.
-	const generatedClassName = hasBlockSupport( blockType, 'className', true )
-		? getBlockDefaultClassName( name )
-		: null;
+	const generatedClassNames = [];
```
ockham, line 78: Alternatively, we could use a string that we append to 🤷‍♂️ 

```suggestion
	let generatedClassName = '';
```

packages/block-editor/src/hooks/test/__snapshots__/align.native.js.snap
```diff
@@ -2,19 +2,19 @@
 
 exports[`Align options for group block sets Full width option 1`] = `
 "<!-- wp:group {"align":"full","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull"></div>
+<div class="wp-block-group-group wp-block-group alignfull"></div>
```
ockham: It'd be great if we were able to keep the default block class name as the first classname, and add the variation as the second one (rather than vice versa). Is this happening due to the order that filters are run? Can we change that? 🤔

packages/block-editor/src/components/block-edit/test/edit.js
```diff
@@ -84,6 +86,47 @@ describe( 'Edit', () => {
...
+		} );
+
+		registerBlockVariation( 'core/test-block', {
+			name: 'variation',
+			title: 'block variation title',
+			attributes: {
+				fruit: {
+					type: 'string',
+					default: 'Bananas',
+				},
```
ockham: I just noticed that this looks like a _block_ attribute definition (with a type and default). However, a variation is defined by the concrete values for its block attributes, i.e.

```suggestion
				fruit: 'Bananas',
```

It should be possible to verify that that's correct by using the shorthand notation for `isActive`, i.e. `[ 'Bananas']`.

packages/block-editor/src/hooks/test/generated-variation-class-name.js
```diff
@@ -0,0 +1,132 @@
...
+	};
+
+	const variationBlockSettings = {
+		attributes: {
+			fruit: {
+				type: 'string',
+				default: 'Bananas',
+				source: 'text',
+				selector: 'div',
+			},
```
ockham: Same as https://github.com/WordPress/gutenberg/pull/61864/files#r1655105267

```suggestion
			fruit: 'Apples',
```

packages/block-editor/src/hooks/test/generated-variation-class-name.js
```diff
@@ -0,0 +1,132 @@
...
+		},
+		save: ( { attributes } ) => (
+			// eslint-disable-next-line react/no-unknown-property
+			<div class={ attributes.className }>{ attributes.fruit }</div>
```
ockham:
```suggestion
			<div className={ attributes.className }>{ attributes.fruit }</div>
```

packages/block-editor/src/hooks/test/__snapshots__/align.native.js.snap
ockham: Ah, I think I see it now. We're adding the filters in the correct order, but each of them is [prepending](https://github.com/WordPress/gutenberg/pull/61864/files#diff-5f22902b77331fe4326ed58f63bef95005aef70a958db2de34e2d39a763ade80R41-R49) the generated block and variation class name, respectively -- meaning that when the variation class name filter is called last, the variation class name is prepended to everything else, including the block class name.

packages/block-editor/src/hooks/test/generated-variation-class-name.js
```diff
@@ -0,0 +1,132 @@
...
+			const blockType = getBlockType( 'core/test-block' );
+			const variationBlockType = {
+				...blockType,
+				...variationBlockSettings,
+			};
+			const extraProps = addSaveProps(
+				{ className: 'foo' },
+				variationBlockType,
+				attributes
+			);
```
ockham: I don't think we should merge `blockType` and `variationBlockSettings` into one object -- they're actually different in nature, see https://github.com/WordPress/gutenberg/pull/61864/files#r1655105267. AFAICT, the filter just needs the block type definition as its second argument:

```suggestion
			const extraProps = addSaveProps(
				{ className: 'foo' },
				blockType,
				attributes
			);
```

ockham: Or even just

```suggestion
			const extraProps = addSaveProps(
				{ className: 'foo' },
				{ name: 'core/test-block', ...defaultBlockSettings },
				attributes
			);
```

packages/block-editor/src/hooks/test/generated-variation-class-name.js
```diff
@@ -0,0 +1,132 @@
...
+						selector: 'div',
+					},
+				},
+				isActive: ( { fruit } ) => fruit === 'Apples',
+			};
+			const extraProps = addSaveProps(
+				{ className: 'foo wp-block-test-block-variation' },
+				variationBlockType,
+				attributes
+			);
```
ockham:
```suggestion
			const extraProps = addSaveProps(
				{ className: 'foo wp-block-test-block-variation' },
				blockType,
				attributes
			);
```

packages/blocks/src/api/parser/test/index.js
```diff
@@ -82,6 +84,52 @@ describe( 'block parser', () => {
...
+			registerBlockVariation( 'core/test-block', {
+				name: 'variation',
+				title: 'Variation',
+				attributes: {
+					fruit: {
+						type: 'string',
+						default: 'Apples',
+						source: 'text',
+						selector: 'div',
+					},
```
ockham:
```suggestion
					fruit: 'Apples',
```

packages/blocks/src/api/test/serializer.js
```diff
@@ -448,4 +449,30 @@ describe( 'block serializer', () => {
...
+		it( 'should return the default class name for a block without the core namespace', () => {
+			expect( getBlockDefaultClassName( 'core/test-block' ) ).toBe(
+				'wp-block-test-block'
+			);
+		} );
+
+		it( 'should return the default class name for a block', () => {
+			expect( getBlockDefaultClassName( 'core/test-block' ) ).toBe(
+				'wp-block-test-block'
+			);
```
ockham: This is kinda duplicated from above, isn't it?

```suggestion
```

packages/blocks/src/api/parser/fix-classnames.js
```diff
@@ -69,3 +69,49 @@ export function fixCustomClassname( blockAttributes, blockType, innerHTML ) {
...
+				return innerHTML.replace(
+					/(<\w+[^>]*\s+class=")([^"]*)"/,
+					`$1$2 ${ variationClassName }"`
+				);
+			}
+		}
+	}
+
+	return innerHTML;
+}
```
ockham: I stared at this for a while, and I think we should move it somewhere else. Right now, it's kind of modeled after `fixCustomClassname`, but it's really different in nature: `fixCustomClassname` scans for classes that are present on the HTML wrapper element but not present in the block's `className` attribute (nor auto-generated), and adds them to that attribute, allowing for some fault tolerance. This is reflected in the function signature, where the HTML is only an input (which is never modified), and only the "fixed" attributes are returned. It's also reflected by the callsite (`applyBuiltInValidationFixes`), where the HTML is passed as `originalHTML` -- so there's arguably some expectation that that won't be changed 😬 

By contrast, `fixVariationClassname` doesn't need to modify the attributes but the HTML, and it's less about fault tolerance and more about migrating block markup globally. So I'd suggest we call it directly from somewhere inside of [`applyBlockDeprecatedVersion`](https://github.com/WordPress/gutenberg/blob/1998f1f9837dca4e5186198b4e45565c97fc3245/packages/blocks/src/api/parser/apply-block-deprecated-versions.js#L19-L124) (which is also where `applyBuiltInValidationFixes`  is called from).

Edit: And from `applyBlockValidation` 😬 

Edit 2: Actually, only from `applyBlockDeprecatedVersion` should be enough 🤔

ockham:
To address https://github.com/WordPress/gutenberg/pull/61864#discussion_r1656640733, 64efaec273b encapsulates the variation class name in its own block-supports flag (`variationClassName`). For now, it's more of a proof-of-concept. As discussed on a call with @tjcafferkey, we should make it a nested property of the [`className` block-supports flag](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#classname), i.e.

```js
supports: {
    className: {
        blockName: true,
        variationName: true,
    }
}
```

(or maybe `blockType` and `variation`?)

For back-compat, `className: true` would then need to be equivalent with

```js
    className: {
        blockName: true,
        variationName: false,
    }
```

Since `className: true` is the default, the slightly unpleasant side effect is that all Core blocks that would like to opt into having variation class names added would need to add those four lines explicitly to their `block.json`s; they'll also need to add block deprecations. (But then again, not all Core blocks support variations.)

ockham:
Here's a (sloppy) patch to test this with the Group block:

<details>
<summary>
Diff
</summary>

```diff
diff --git a/packages/block-library/src/group/block.json b/packages/block-library/src/group/block.json
index db7d09c55d2..ac3f95b490e 100644
--- a/packages/block-library/src/group/block.json
+++ b/packages/block-library/src/group/block.json
@@ -90,6 +90,10 @@
 		},
 		"interactivity": {
 			"clientNavigation": true
+		},
+		"className": {
+			"block": true,
+			"variation": true
 		}
 	},
 	"editorStyle": "wp-block-group-editor",
diff --git a/packages/block-library/src/group/deprecated.js b/packages/block-library/src/group/deprecated.js
index fcc2b249d22..6c2e9e9fcf5 100644
--- a/packages/block-library/src/group/deprecated.js
+++ b/packages/block-library/src/group/deprecated.js
@@ -43,6 +43,96 @@ const migrateAttributes = ( attributes ) => {
 
 const deprecated = [
 	// Version with default layout.
+	{
+		attributes: {
+			tagName: {
+				type: 'string',
+				default: 'div',
+			},
+			templateLock: {
+				type: [ 'string', 'boolean' ],
+				enum: [ 'all', 'insert', false ],
+			},
+		},
+		supports: {
+			__experimentalOnEnter: true,
+			__experimentalSettings: true,
+			align: [ 'wide', 'full' ],
+			anchor: true,
+			ariaLabel: true,
+			html: false,
+			color: {
+				gradients: true,
+				link: true,
+				__experimentalDefaultControls: {
+					background: true,
+					text: true,
+				},
+			},
+			spacing: {
+				margin: [ 'top', 'bottom' ],
+				padding: true,
+				blockGap: true,
+				__experimentalDefaultControls: {
+					padding: true,
+					blockGap: true,
+				},
+			},
+			__experimentalBorder: {
+				color: true,
+				radius: true,
+				style: true,
+				width: true,
+				__experimentalDefaultControls: {
+					color: true,
+					radius: true,
+					style: true,
+					width: true,
+				},
+			},
+			typography: {
+				fontSize: true,
+				lineHeight: true,
+				__experimentalFontStyle: true,
+				__experimentalFontWeight: true,
+				__experimentalLetterSpacing: true,
+				__experimentalTextTransform: true,
+				__experimentalDefaultControls: {
+					fontSize: true,
+				},
+			},
+			layout: true,
+			className: {
+				block: true,
+				variation: true,
+			}
+		},
+		save( { attributes: { tagName: Tag } } ) {
+			return (
+				<Tag { ...useInnerBlocksProps.save( useBlockProps.save() ) } />
+			);
+		},
+		isEligible: ( { layout } ) =>
+			! layout ||
+			layout.inherit ||
+			( layout.contentSize && layout.type !== 'constrained' ),
+		migrate: ( attributes ) => {
+			const { layout = null } = attributes;
+			if ( ! layout ) {
+				return attributes;
+			}
+			if ( layout.inherit || layout.contentSize ) {
+				return {
+					...attributes,
+					layout: {
+						...layout,
+						type: 'constrained',
+					},
+				};
+			}
+		},
+	},
+	// Version without variation class name.
 	{
 		attributes: {
 			tagName: {
```
</details>

peterwilsoncc (commented):
Rather than `wp-block-block-name-variation` is it possible to set the class names to `wp-block-block-name_variation`?

This will help avoid naming collisions if the variation matches the suffix of an existing block name. 

It's an edge case so not a big problem if it can't be done.

ockham:
> Rather than `wp-block-block-name-variation` is it possible to set the class names to `wp-block-block-name_variation`?
> 
> This will help avoid naming collisions if the variation matches the suffix of an existing block name.
> 
> It's an edge case so not a big problem if it can't be done.

Yeah, that's something we can consider! It's true that we have some precedent -- even in Core -- of blocks whose name fully contains the name of another block, e.g. `comments`, `comments-pagination`, and `comments-pagination-next`. I'm not sure how big the risk is that something like that might also overlap with a block variation (for example, a fictitious `next` variation of the `comments-pagination` block), but it's conceivable.

ockham:
Updated TODO:

- [x] Fix the order of generated class names. We want the variation class name inserted _after_ the block class name (which is currently not the case). This is non-trivial, since both the `addGeneratedClassName` and the `addGeneratedVariationClassName` filters are hooked to `blocks.getSaveContent.extraProps` (and there's no order/priority arg for client-side filters).
  - Maybe the best solution is to absorb `addGeneratedVariationClassName` into `addGeneratedClassName`.
- [x] Make sure `className: { block: false }` is handled correctly.
- [x] Add test coverage
  - Verify that a block with block-supports for the variation class name gets the correct classname added, in the correct place. Cover different scenarios:
  - `className: true`
  - `className: { block: false, variation: true }`
  - `className: { variation: true }`
  - etc.
- [ ] ~Prefix `variation` block-support with `__experimental` (or use private APIs)?~ _See https://github.com/WordPress/gutenberg/pull/61864#issuecomment-2220473054._
- [x] Add docs

peterwilsoncc:
> Yeah, that's something we can consider!... I'm not sure how big the risk is that something like that might also overlap with a block variation

I can't either (which is why I didn't use an example :)) so don't stress too much if it's a significant refactor.

> Prefix variation block-support with __experimental?

Is it possible to set block-supports up with the private APIs or did that prove too complex as it comes from the json config file?

packages/blocks/src/api/serializer.js
ockham: 59c23577a56

packages/block-editor/src/components/block-edit/test/edit.js
```diff
@@ -84,6 +86,50 @@ describe( 'Edit', () => {
...
+			},
+		} );
+
+		registerBlockVariation( 'core/test-block', {
+			name: 'variation',
+			title: 'block variation title',
+			attributes: {
+				fruit: 'Bananas',
+			},
+			isActive: ( { fruit } ) => fruit === 'Bananas',
```
ockham:
```suggestion
			isActive: [ 'fruit' ],
```

ockham:
> > Prefix variation block-support with __experimental?
> 
> Is it possible to set block-supports up with the private APIs or did that prove too complex as it comes from the json config file?

Ah, good point! I hadn't thought of that but will look into that 😊

ockham:
> > > Prefix variation block-support with __experimental?
> > 
> > Is it possible to set block-supports up with the private APIs or did that prove too complex as it comes from the json config file?
> 
> Ah, good point! I hadn't thought of that but will look into that 😊

I looked into the [docs](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/#legacy-experimental-apis-plugin-only-apis-and-private-apis), and it seems that private APIs can't be used for block-supports flags. Per discussion with @gziolo, we're thinking to go with unprefixed values instead. This way, a decision on the new flag(s) can be made during PR review, and we won't leave another `__experimental` flag behind.

ockham:
There's still a console error when following the test instructions (as soon as the Code Editor is opened).
Furthermore, saving the template (after migrations have run) and then reloading it shows a number of blocks as invalid.

Need to investigate 😕

ockham:
> There's still a console error when following the test instructions (as soon as the Code Editor is opened). Furthermore, saving the template (after migrations have run) and then reloading it shows a number of blocks as invalid.
> 
> Need to investigate 😕

Looks like I'd made a mistake in the deprecation that I'd used to test adding the variation className block support to the Group block. I've updated the diff in the PR desc now, and I believe the error is gone.

ockham:
I believe that this PR is finally ready for review. Please test using the instructions and patch from the PR description. Furthermore, please help us answer the following questions:

- **Are the names I chose for the new sub-properties of the `className` block support fine?**

```js
className: {
    block: true,
    variation: true,
}
```

- **Should we add code to inject the variation-specific class name on the frontend** (i.e. on the server side) when a block has opted into variation class name support, even if the block hasn't been touched in the editor? (We have some precedent for that with some blocks that only added `className` support for the block-specific class name later, see e.g. [List](https://github.com/WordPress/gutenberg/blob/93d16e52ac410fddcef43d31f34bc11749211331/packages/block-library/src/list/index.php#L32) and [Heading](https://github.com/WordPress/gutenberg/blob/93d16e52ac410fddcef43d31f34bc11749211331/packages/block-library/src/heading/index.php#L34).)
  - Note that this is a bit more complicated for variation class names than for block class names, as the server needs to be able to determine which variation is active. This is being explored in https://github.com/WordPress/wordpress-develop/pull/6602; there are some more related thoughts in this [make/core proposal](https://make.wordpress.org/core/2024/06/24/proposal-block-variation-aliases/).
  - If we decide to implement this, I'd strongly advocate to do it in a follow-up PR, as this one has been sitting for a while, and it's going to take some more time to figure out the server-side part.

ockham:
👋 Pinging @ntsekouras and @gziolo for review in case y'all can spare some time to review this later this week or next week 😊

ndiego:
> Should we add code to inject the variation-specific class name on the frontend?

I think this would be an excellent follow-up. Not being able to directly style block variations has come up a fair amount in my discussions with developers, so this PR is an amazing improvement. 

But, having the classes added automatically to all existing variations on a site will make this enhancement much better. Otherwise, you would have to go in and update every page/post, if I am understanding this implementation correctly.

packages/block-editor/src/hooks/supports.js
```diff
@@ -339,3 +339,33 @@ export const getLayoutSupport = ( nameOrType ) =>
...
+/**
+ * Returns true if the block defines support for block class name.
+ *
+ * @param {string|Object} nameOrType Block name or type object.
+ * @return {boolean} Whether the block supports the feature.
+ */
+export const hasBlockClassNameSupport = ( nameOrType ) => {
+	const classNameSupport = getBlockSupport( nameOrType, 'className', true );
+
+	if ( classNameSupport === true || classNameSupport === false ) {
```
ntsekouras: Nit: 
```suggestion
	if ( typeof classNameSupport === 'boolean' ) {
```

packages/block-editor/src/hooks/supports.js
```diff
@@ -339,3 +339,33 @@ export const getLayoutSupport = ( nameOrType ) =>
...
+		return classNameSupport;
+	}
+
+	// classNameSupport can be an object. If it doesn't have a block key,
```
ntsekouras:
```suggestion
	// classNameSupport can be an object. If it doesn't have a `block` key,
```

packages/block-editor/src/hooks/supports.js
```diff
@@ -339,3 +339,33 @@ export const getLayoutSupport = ( nameOrType ) =>
...
+	// we default to true.
+	return (
+		! Object.hasOwn( classNameSupport, 'block' ) ||
```
ntsekouras, line 359: Would this be equivalent with `hasBlockSupport( nameOrType, 'className.block', true );` ?

packages/block-editor/src/hooks/supports.js
```diff
@@ -339,3 +339,33 @@ export const getLayoutSupport = ( nameOrType ) =>
...
+	);
+};
+
+/**
+ * Returns true if the block defines support for variation class name.
+ *
+ * @param {string|Object} nameOrType Block name or type object.
+ * @return {boolean} Whether the block supports the feature.
+ */
+export const hasVariationClassNameSupport = ( nameOrType ) =>
```
ntsekouras, line 370: Should we default to `true` for this feature? Have you discussed this? This would also affect the default handling when we have `supports.className: true`.

packages/block-editor/src/hooks/test/supports.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * Internal dependencies
+ */
+import {
+	hasBlockClassNameSupport,
+	hasVariationClassNameSupport,
+} from '../supports';
+
+describe( 'hasBlockClassNameSupport', () => {
```
ntsekouras, line 9: Do you think these tests add value? Aren't these cases covered extensively at `generated className` tests?

gziolo: What problem do you anticipate with these tests? They document the behavior of the util. Should it be treated as private?

ntsekouras: No problem at all. I just thought we could skip them as the actual behavior is tested with the others. No strong opinion for sure.

packages/block-editor/src/hooks/generated-class-name.js
```diff
@@ -11,30 +22,41 @@ import { hasBlockSupport, getBlockDefaultClassName } from '@wordpress/blocks';
...
+	if ( hasVariationClassNameSupport( blockType ) ) {
+		const variationClassName = getBlockVariationClassName(
+			blockType.name,
+			attributes
+		);
+		if ( variationClassName ) {
+			generatedClassNames.push( variationClassName );
 		}
 	}
+	if ( typeof extraProps.className === 'string' ) {
```
ntsekouras, line 48: If we don't have either classname support at this point, we can bail early and return the props.

packages/blocks/src/api/serializer.js
```diff
@@ -50,6 +51,26 @@ export function getBlockDefaultClassName( blockName ) {
...
+ *
+ * @return {string|null} The block variation classname, or null if the block doesn't match any variation.
+ */
+export function getBlockVariationClassName( blockName, attributes ) {
+	const activeVariation = getActiveBlockVariation( blockName, attributes );
+	if ( ! activeVariation ) {
+		return null;
+	}
+
+	return getBlockDefaultClassName( blockName ) + '-' + activeVariation.name;
```
ntsekouras: Maybe worth mentioning at the docs that the prefix is from `getBlockDefaultClassName`, which means its value is also affected by the filter? 🤔

packages/blocks/src/api/index.js
```diff
@@ -136,6 +137,7 @@ export {
 	getBlockSupport,
 	hasBlockSupport,
 	getBlockVariations,
+	getActiveBlockVariation,
```
ntsekouras: Do we need to export this one?

packages/block-editor/src/components/block-list/block.js
```diff
@@ -592,12 +594,27 @@ function BlockListBlockProvider( props ) {
...
 			} = select( blocksStore );
+
 			const attributes = getBlockAttributes( clientId );
 			const { name: blockName, isValid } = blockWithoutAttributes;
 			const blockType = getBlockType( blockName );
 			const { supportsLayout, __unstableIsPreviewMode: isPreviewMode } =
 				getSettings();
 			const hasLightBlockWrapper = blockType?.apiVersion > 1;
+			const defaultClassNames = [];
+			if ( hasLightBlockWrapper && blockName ) {
```
ntsekouras: Is there a case where `blockName` isn't set?

ntsekouras:
Thanks for all your work here! 

I've left some comments but I have a couple more questions.

So it seems this only works for now for static blocks and not dynamic blocks, right? This should be documented if we follow an iterative approach.

It seems we would need a deprecation for every block that needs to add this support, because we indirectly changing the `save` function. This should be documented too..

packages/block-editor/src/hooks/supports.js
ockham: I think so, yeah. Are you suggesting we replace it with that? 🙂 

(Personally, I find it a bit easier to read if we keep it as-is: We've fetched the value of `className` block-support and have determined that it's not a boolean, so now we check if it has a `block` property and return its value if it does; otherwise, we default to `true`. It also removes the need for separately calling the `hasBlockSupport` selector, since we already have fetched `classNameSupport`.)

packages/block-editor/src/hooks/supports.js
ockham: I was originally hoping to do that (and discussed it with @gziolo again today), but I've come to the conclusion that it's too much of a risk. As you mentioned [elsewhere](https://github.com/WordPress/gutenberg/pull/61864#issuecomment-2236340867), setting this to `true` will change the serialized markup for existing blocks, so we'll need to add migrations for them.

Having to add those for any block with variations that wants to use this new feature seemed like a bit of a hassle at first, but I now think it's an okay tradeoff. It's probably worse for plugins (and authors) that include a large number of blocks -- and for Core itself -- but even then, not all blocks come with variations, and it's a one time change. IMO, it's reasonable to ask extenders to enable this feature manually (and to add a deprecation) in exchange for a useful enhancement.

Since we might also need [a server-side part](https://github.com/WordPress/gutenberg/pull/61864#issuecomment-2233381017) (which will require a robust way of [determining the active variation](https://github.com/WordPress/wordpress-develop/pull/6602)), it makes even more sense to require extenders to opt into this new feature explicitly, as it gives them control over adding this new class name, and allows them to fix their `isActive` fields before they do so.

packages/block-editor/src/hooks/test/supports.js
ockham: It actually took me a while to get `hasBlockClassNameSupport` and `hasVariationClassNameSupport` right, since they're a bit subtle, especially with regard to their default behavior (e.g. no `supports.className` field in `block.json` specified means block class name is `true`, whereas variation class name is `false`). I found that the tests helped me cover the various cases; at some point, I changed a default behavior, and the tests proved hugely helpful to get it right. So I'd like to keep them 😄

packages/block-editor/src/hooks/generated-class-name.js
ockham: Good point! Addressed in 1cbb0612b4a.

packages/blocks/src/api/index.js
ockham: Ah, we probably don't! I think that's a leftover from an earlier iteration. Good spot! I'll remove it.

ockham: 9965b64f17e

packages/blocks/src/api/serializer.js
ockham: Yeah, good point!

ockham: 455a7f4f35a

packages/block-editor/src/components/block-list/block.js
ockham: Hmm, good point. Seems unlikely. Probably another leftover from a previous iteration of the PR.

I've removed it in 8ea9ac35be31089fd97546012bfb9368c5d3f200.

ockham:
> Thanks for all your work here!
> 
> I've left some comments but I have a couple more questions.
> 
> So it seems this only works for now for static blocks and not dynamic blocks, right? This should be documented if we follow an iterative approach.

Yes, at least for now. [This comment](https://github.com/WordPress/gutenberg/pull/61864#issuecomment-2233381017) has some more on that.

> It seems we would need a deprecation for every block that needs to add this support, because we indirectly changing the `save` function. This should be documented too..

Yeah, that's correct. But isn't that the case for other block-support flags too? 🤔 
Anyway, I'm happy to document this somewhere. Should I add it to `docs/reference-guides/block-api/block-supports.md`?

packages/blocks/src/api/serializer.js
ockham: And two more notes (right in the JSDoc) here: 19f2e00fdfb

ockham:
Rebased.

ockham:
Rebased.
@ntsekouras When you get a chance, would you mind giving this another look? 😊

packages/blocks/src/api/serializer.js
gziolo: I saw a comment from @peterwilsoncc suggesting a different char than `-` as a delimiter in https://github.com/WordPress/gutenberg/pull/61864#pullrequestreview-2152376947. I like that idea as it removes the probability of having the same class for two different blocks and their variations:

`plugin/foo` + `boo` -> `wp-block-plugin-foo-boo`
`plugin/foo-boo` -> `wp-block-plugin-foo-boo`

Maybe we could follow [BEM rules](https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/coding-guidelines.md#css) and use `__`?

schemas/json/block.json
```diff
@@ -256,7 +256,22 @@
...
+						{ "type": "boolean" },
+						{
+							"type": "object",
+							"properties": {
+								"block": {
+									"description": "Add the .wp-block-your-block-name class to the block's wrapper element",
+									"type": "boolean"
+								},
+								"variation": {
+									"description": "Add the .wp-block-your-block-name-your-block-variation class to the block's wrapper element",
```
gziolo: Should it get clarified that this class name gets injected only when the active variations can be detected during saving?

packages/blocks/src/api/serializer.js
gziolo: One more thing worth noting here. `getBlockDefaultClassName( blockName )` after filters get applied might change to anything, so it could also be multiple class names: `foo boo bar`. In that case, the last one will get changed extended, and the reset repeated. Should it pick only the first class name returned so that doesn't end up looking as follows:
`foo boo bar foo boo bar-variaiton`

packages/block-editor/src/hooks/test/supports.js
```diff
@@ -0,0 +1,97 @@
...
+	const blockName = 'block/name';
+
+	it( 'should default to true', () => {
+		const block = {
+			name: blockName,
+		};
+		expect( hasBlockClassNameSupport( block ) ).toEqual( true );
```
gziolo: Nitpick: when testing against boolean value the `===` strict equality operator is more appropriate:
```suggestion
		expect( hasBlockClassNameSupport( block ) ).toBe( true );
```
It applies to all test cases added.

gziolo:
Overall, it looks very close to ready. As noted in https://github.com/WordPress/gutenberg/pull/61864#discussion_r1747508276, the implementation might be a little bit brittle in the case when the `blocks.getBlockDefaultClassName` filter append some random class, example:

```js
addFilter( 'blocks.getBlockDefaultClassName', 'my-plugin/append-class-name', ( value ) => value + ' random-class' );
```

It's been like that forever for default class names, but it's hard to tell what should happen for variations in this case.

We should take it to the finish line and treat it as a progressive enhancement that gets applied when the blocks that opt to use this option get saved to the database. I'm also not entirely sure what will happen when you enable class names for variations for existing core blocks on your WP instance. Would it invalidate all saved blocks of that type that match the active variation? It's something I would like to test next.

schemas/json/block.json
ockham: Hmm, but isn't that equivalent to active variation detection working at all on the client side? I.e. if a block is correctly shown e.g. in the block inspector as a variation, then the class name will get injected correctly as well. (As opposed to the server side, where active variation detection is a bit trickier.)

If it's equivalent to "the active variation is detected correctly [on the client]", then IMO we don't need to point that out 🤔

packages/block-editor/src/hooks/test/supports.js
ockham: Ah, good spot! I'll change those 👍

ockham: 9b5f0b60f3a98ee5129b701f76ab8218a572b7fe and 611dd128dac

packages/blocks/src/api/serializer.js
ockham: > One more thing worth noting here. `getBlockDefaultClassName( blockName )` after filters get applied might change to anything, so it could also be multiple class names: `foo boo bar`. In that case, the last one will get changed extended, and the reset repeated. Should it pick only the first class name returned so that doesn't end up looking as follows: `foo boo bar foo boo bar-variaiton`

Yeah, that's a tricky one. TBH I'm not sure what the answer should be here; multiple options come to mind:
1. Appending the variation to the first class name, as you suggested, i.e. `foo-variation boo bar`.
2. Appending the variation to the last class name only: `foo boo bar-variation`
3. Appending it to the string, splitting it at whitespaces, deduplicating, and rejoining:
   1. Attach: `foo boo bar foo boo bar-variation`
   2. Split: `[ 'foo', 'boo', 'bar', 'foo', 'boo', 'bar-variation' ]`
   3. Deduplicate: `[ 'foo', 'boo', 'bar', 'bar-variation' ]`
   4. Rejoin: `foo boo bar bar-variation`
 This is equivalent to keeping a copy of the last class name, and appending another copy with the variation name attached.

TBH none of these stands out to me as the "right" answer.

My original thinking was to treat the "multiple classnames" case as unsupported (as both the function name `getBlockDefaultClassName` (singular!) and the examples in the docs indicate that it should be _one_ class name only); i.e. people would indeed end up with `foo boo bar foo boo bar-variation`. I'm not sure how much of a problem that would be in practice.

ockham: Finally, we could consider bypassing the `getBlockDefaultClassName()` function altogether when creating the variation class name, and instead introduce a dedicated one (plus a filter) for the variation.

However, using that function for both the block class name and the variation class name has one big advantage. Consider the following example:

```js
// Our filter function.
function setBlockCustomClassName( className, blockName ) {
	return blockName === 'core/code' ? 'my-plugin-code' : className;
}

// Adding the filter.
wp.hooks.addFilter(
	'blocks.getBlockDefaultClassName',
	'my-plugin/set-block-custom-class-name',
	setBlockCustomClassName
);
```

In this case, if we use `getBlockDefaultClassName` for both the block and variation class name, we get `my-plugin-code my-plugin-code-myvariation` as result.

If we don't use `getBlockDefaultClassName` to generate the variation, we'd get `my-plugin-code wp-code-myvariation`, i.e. we'd lose consistency between the block and variation class names; the user would have to manually duplicate the logic used in `setBlockCustomClassName` for the variation-specific filter.

ockham:
> Overall, it looks very close to ready. As noted in [#61864 (comment)](https://github.com/WordPress/gutenberg/pull/61864#discussion_r1747508276), the implementation might be a little bit brittle in the case when the `blocks.getBlockDefaultClassName` filter append some random class, example:
> 
> ```js
> addFilter( 'blocks.getBlockDefaultClassName', 'my-plugin/append-class-name', ( value ) => value + ' random-class' );
> ```
> 
> It's been like that forever for default class names, but it's hard to tell what should happen for variations in this case.

Indeed. I've replied to your comment in https://github.com/WordPress/gutenberg/pull/61864#discussion_r1756990859 and https://github.com/WordPress/gutenberg/pull/61864#discussion_r1756996341.

> We should take it to the finish line and treat it as a progressive enhancement that gets applied when the blocks that opt to use this option get saved to the database. I'm also not entirely sure what will happen when you enable class names for variations for existing core blocks on your WP instance. Would it invalidate all saved blocks of that type that match the active variation? It's something I would like to test next.

Yeah, it would invalidate them. If you look at the patch that's part of the testing instructions in the PR description, it's three lines to opt the Group block into the auto-generated variation-specific class name, and another ~100 lines to add the deprecation 😬

ockham:
> We should take it to the finish line and treat it as a progressive enhancement that gets applied when the blocks that opt to use this option get saved to the database.

Yeah, I'd also love to get this one done; it's been open for a couple of months now.

I have some concerns around committing to a format for both the [block-supports semantics](https://github.com/WordPress/gutenberg/pull/61864/files#diff-cb2052b95e3c5b5fa711f6519d1e3c376597cd1747e6a0a781dc51447255ef16R181-R183) I've chosen (especially since they're non-experimental), the usage of the `blocks.getBlockDefaultClassName` that you pointed out, and the [choice of character(s)](https://github.com/WordPress/gutenberg/pull/61864#discussion_r1747473914) we use to append the variation name.

Furthermore, I think there's a real chance that folks will want us to add the variation class names on the server side as well (as Nick [pointed out](https://github.com/WordPress/gutenberg/pull/61864#issuecomment-2233394708)); if people feel fairly strongly about this, then we should probably _not_ get the client-side part into 6.7 without the server-side part (which is unfortunately a bit more complex to get right, as discussed). So timing is a bit tricky here 😕

packages/blocks/src/api/serializer.js
ockham: > I saw a comment from @peterwilsoncc suggesting a different char than `-` as a delimiter in [#61864 (review)](https://github.com/WordPress/gutenberg/pull/61864#pullrequestreview-2152376947). I like that idea as it removes the probability of having the same class for two different blocks and their variations:
> 
> `plugin/foo` + `boo` -> `wp-block-plugin-foo-boo` `plugin/foo-boo` -> `wp-block-plugin-foo-boo`
> 
> Maybe we could follow [BEM rules](https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/coding-guidelines.md#css) and use `__`?

Trying this out in d369a978111.

gziolo:
> I have some concerns around committing to a format for both the [block-supports semantics](https://github.com/WordPress/gutenberg/pull/61864/files#diff-cb2052b95e3c5b5fa711f6519d1e3c376597cd1747e6a0a781dc51447255ef16R181-R183) I've chosen (especially since they're non-experimental), the usage of the blocks.getBlockDefaultClassName that you pointed out, and the https://github.com/WordPress/gutenberg/pull/61864#discussion_r1747473914 we use to append the variation name.

I'm not worried about the semantics of `supports.className` as it's fully backward compatible and enabling class names for variations is optional.

I'm starting to think that we should unbundle the variations class name generation from the default class name, example implementation:

```js
function generateDefaultClassName( blockName, variationName = '' ) {
	// Generated HTML classes for blocks follow the `wp-block-{name}` nomenclature.
	// Blocks provided by WordPress drop the prefixes 'core/' or 'core-' (historically used in 'core-embed/').
	const className = 'wp-block-' + blockName.replace( /\//, '-' ).replace( /^core-/, '' );
	return variationName?  className + '__' + variationName : className;
}

export function getBlockDefaultClassName( blockName ) {
	const className = generateDefaultClassName( blockName );

	return applyFilters(
		'blocks.getBlockDefaultClassName',
		className,
		blockName
	);
}

// This one is more nuanced, but the simplified version is to illustrate the idea.
export function getBlockDefaultVariationClassName( blockName, variationName ) {
	const className = generateDefaultClassName( blockName, variationName );

	return applyFilters(
		'blocks.getBlockDefaultVariationClassName',
		className,
		blockName,
		variationName
	);
}
```

This way, we would start fresh for block variations. In case, folks don't like the pattern used when generating the class name for block variation they have another filter to use in addition to preexisting `blocks.getBlockDefaultClassName`.

ockham:
> I'm starting to think that we should unbundle the variations class name generation from the default class name, example implementation: [...]

Yeah, I think that's a good idea.

As discussed on our call earlier this week, I'll pick this up after GB 19.3 RC1 is out (which is due today; however, I'll be gone for a week, starting tomorrow). I will try to land it soon after that, though. This way, it won't be in WP 6.7, and we'll have a bit more time to tweak it.

t-hamano:
This issue has been punted to 6.8. Let's move this issue to the 6.9 Editor Tasks board and see what's possible in 6.9.

Firstly, I'd be happy to confirm if this PR is still valid.

t-hamano:
I'd suggest punting this PR to 7.0, as we may need more discussion. What do you think?

PR #61865, opened 2024-05-22
Post Actions: Hide the trash action for auto-drafts

Mamaduka:
## What?
PR updates the post status check condition for the "Move to Trash" action and hides it for freshly created posts.

## Why?
It matches the condition from the legacy `PostTrash` component, which used the `isEditedPostNew` selector for this check.

## Testing Instructions
1. Create a new post.
2. Confirm that the trash action is hidden.
3. Add a title to the post and save it.
4. Confirm that the trash action is available.

Same.


ntsekouras (approved):
Looks good, thanks!

Mamaduka:
Thanks for testing, Nik!

PR #61866, opened 2024-05-22
Zoom Out: Hide inserters behind the experiment flag

scruffian:
## What?
Hides the zoom out inserters behind the next experiment flag.

## Why?
Until zoom out mode works well with the patterns tab we should disable these inserters. You still should be able to access them by enabling the experiment for testing purposes.

## How?
Only outputs the inserters when the experiment is enabled.

## Testing Instructions
1. Open the Site Editor
2. Open Global Styles
3. Open Browse Styles
4. Confirm that the view zooms out
5. Confirm that no inserters are visible
6. Enable the zoom out experiment from Gutenberg > Experiments
1. Open the Site Editor
2. Open Global Styles
3. Open Browse Styles
4. Confirm that the view zooms out
5. Confirm that inserters are now visible

## Screenshots or screencast 

|Before|After|
|-|-|
|Screenshot 2024-05-22 at 15 04 01|Screenshot 2024-05-22 at 15 05 11|

### DIFF
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -231,11 +231,12 @@ export default function BlockTools( {
 					name="__unstable-block-tools-after"
 					ref={ blockToolbarAfterRef }
 				/>
-				{ isZoomOutMode && (
-					<ZoomOutModeInserters
-						__unstableContentRef={ __unstableContentRef }
-					/>
-				) }
+				{ window.__experimentalEnableZoomedOutPatternsTab &&
+					isZoomOutMode && (
+						<ZoomOutModeInserters
+							__unstableContentRef={ __unstableContentRef }
+						/>
+					) }
 			</InsertionPointOpenRef.Provider>
 		</div>
 	);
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/components/block-tools/index.js
```diff
@@ -231,11 +231,12 @@ export default function BlockTools( {
 					name="__unstable-block-tools-after"
 					ref={ blockToolbarAfterRef }
 				/>
-				{ isZoomOutMode && (
-					<ZoomOutModeInserters
-						__unstableContentRef={ __unstableContentRef }
-					/>
-				) }
+				{ window.__experimentalEnableZoomedOutPatternsTab &&
```
ellatrix, line 234: Is this how all experiments work? I don't love the window global 😊

ellatrix (approved):
Its usefulness is anyway arguable in browse styles?

packages/block-editor/src/components/block-tools/index.js
scruffian: Yeah it's how we do it elsewhere.

Issue #59689, opened 2024-03-07
Tracking issue: Normalize `details` view with inspector controls in site editor


ntsekouras:
In site editor currently the details view in navigation sidebar about the selected entity(page, template, etc..)  shares a lot of information with contents/panels in the inspector sidebar. 

Details               |  Inspector
-----------------------------------------|-------------------------
<img width="356" alt="Screenshot 2024-03-07 at 4 14 39 PM" src="https://github.com/WordPress/gutenberg/assets/16275880/cf2f1393-6057-40dd-bc95-deffd78e39c4"> |  <img width="356" alt="Screenshot 2024-03-07 at 4 14 51 PM" src="https://github.com/WordPress/gutenberg/assets/16275880/7e8a8be4-3498-4f4d-8efb-96f868da98be">

It would be good to merge them and see how we can eventually make a reusable component that can used in other places too. An example would be to show the inspector controls somehow in the entities list that use the new Data Views(pages, patterns, etc..).

The process should be additive and not reductive. We should **add whatever is missing from the details view to the inspector controls**. Some duplication already exists and it's fine to add a bit more until the changes have completed.

Details view contain a lot of `readonly` information in contrast to inspector controls. If some item can be made editable(ex Featured Image), it should be added with `edit` functionality. 

The issue is not exhaustive and design can be discussed either here or on the related PRs. It's meant to be an initial break down of tasks to parallelize work and can be updated.

## Design

<img width="4313" alt="panels" src="https://github.com/WordPress/gutenberg/assets/846565/3d4dda52-8e06-45f1-9ec2-057f38563939">

## Tasks
- [x] Move the readonly items https://github.com/WordPress/gutenberg/pull/60672
- [x] Move the Featured Image 
- [x] Consolidate the ellipsis menu currently in details view with the main entity info(page card, etc..). Related to that could be to check the actions we have in Data Views and also consolidate with them.([PR](https://github.com/WordPress/gutenberg/pull/59849))
- [x] Move revisions info
- [x] Provide a way to make the `title` editable, either with rich text or a `rename` action in ellipsis menu(will need design discussions)
- [x] Move excerpt/description https://github.com/WordPress/gutenberg/pull/60894
- [x] When the complete merge has completed, we should remove/deprecate the details interstitial page for content views. Related [issue](https://github.com/WordPress/gutenberg/issues/59659) for this is the advancing site editor index views.

jameskoster:
[...]

PR #61867, opened 2024-05-22
Inspector: Remove revisions panel

jameskoster:
Follow-up to https://github.com/WordPress/gutenberg/issues/59689.

## What?
Remove the revisions panel from the document inspector.

## Why?
Revisions are now accessible from the ellipsis menu which offers more appropriate prominence.

## Testing Instructions
1. Edit documents like pages, posts, templates, etc.
2. Observe the 'View revisions' action in the ellipsis menu
3. Notice the 'Revisions' panel is no longer visible in the Inspector

| Before | After |
| --- | --- |
| Screenshot 2024-05-22 at 15 17 14 | Screenshot 2024-05-22 at 15 17 42 |

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -1108,6 +1108,7 @@ export function usePostActions( postType, onActionPerformed ) {
 	].includes( postType );
 	const isPattern = postType === PATTERN_POST_TYPE;
 	const isLoaded = !! postTypeObject;
+	const supportsRevisions = !! postTypeObject?.supports?.revisions;
 	return useMemo( () => {
 		if ( ! isLoaded ) {
 			return [];
@@ -1115,7 +1116,7 @@ export function usePostActions( postType, onActionPerformed ) {
 
 		const actions = [
 			postTypeObject?.viewable && viewPostAction,
-			postRevisionsAction,
+			supportsRevisions && postRevisionsAction,
 			globalThis.IS_GUTENBERG_PLUGIN
 				? ! isTemplateOrTemplatePart &&
 				  ! isPattern &&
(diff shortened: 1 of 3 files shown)


youknowriad:
Pushed a small commit to ensure the action is hidden from post types that don't support revisions (just like the component). Also, with this, we're losing the revision count from the UI (just want to make sure that it's ok).

youknowriad (approved):

jameskoster:
@youknowriad Ideally the revisions menu item would include the count as a `suffix`. I wasn't able to get that working. Feel free to push if it's a trivial change.

youknowriad:
It's not trivial I think, the "label" of the action need to be a component and not sure we support that @ntsekouras @jorgefilipecosta

jameskoster:
I could get it to work by adding `suffix={ action.suffix }` here: https://github.com/WordPress/gutenberg/blob/f49e0b5958adad9fbd211d5ded07b4cabf5bfec3/packages/editor/src/components/post-actions/index.js#L92-L95

But I wasn't able to update `postRevisionsAction` so that it returned the count. 

Seems as though it'd be possible given the count is already available:

https://github.com/WordPress/gutenberg/blob/f49e0b5958adad9fbd211d5ded07b4cabf5bfec3/packages/editor/src/components/post-actions/actions.js#L481

I feel like I'm missing something obvious :D

ntsekouras:
>It's not trivial I think, the "label" of the action need to be a component and not sure we support that @ntsekouras @jorgefilipecosta

Yeah, we don't for now it would involve some exploring to see what a good API for that would be. Is that a blocker for this PR and the removal of the panel?

--edit

I'll open a separate PR to support labels as components.

jameskoster:
Not a blocker.

jameskoster:
I guess some tests will need updating here.

ntsekouras (approved):
Looks good and I updated the e2e tests. Thank you!

PR #61868, opened 2024-05-22
Fix broken license check script

swissspidy:
## What?

While working on another PR I realized that the license check script isn't working correctly.

This fixes it.

## Why?

Due to early `return` statements in the `for` loops, numerous dependencies were not actually checked, which led to the false assumption that all licenses are compatible when they are in fact not.

## How?

Replace early returns with `continue` instead

Also fixes some incorrect output when `dep` is an empty object for instance.

## Testing Instructions

Run license check, see failing status.


gziolo:
There are some incompatible licenses reported that need to be further investigated:

Screenshot 2024-05-24 at 11 46 44

swissspidy:
For reference:
 https://github.com/WordPress/gutenberg/blob/8c3a688af79c37adcb89c32aab92c3a7fd3772f1/packages/scripts/scripts/check-licenses.js#L68-L73

Looking at the reported errors and the packages based on `npm ls`, the following are clearly used in build tooling and can be ignored:

* `@ampproject/remapping`
* `human-signals`
* `chromium-edge-launcher`
* `human-signals`
* `fb-watchman`
* `bser`
* `walker`
* `jackspeak`
* `path-scurry`

The following ones need more clarification:

* `lighthouse`:
	* Used by `@wordpress/e2e-test-utils-playwright`
	* `lighthouse` and many of its dependencies are licensed under Apache 2.0
		* Examples: `lighthouse-logger`, `marky`, `chrome-launcher` (also used by RN build tools)

gziolo:
> Looking at the reported errors and the packages based on npm ls, the following are clearly used in build tooling and can be ignored:

Maybe they are dependencies of the React Native-related CLI tools that are listed as production dependencies and, therefore, explicitly ignored.

gziolo:
`@wordpress/e2e-test-utils-playwright` should get recognized as a development package, so therefore, all its dependencies, like Lighthouse, are able to use an `Apache 2.0` license. It's all complex to configure all these checks correctly.

swissspidy:
`chrome-launcher`, `lighthouse-logger`, and `marky` (which is a dep of `lighthouse-logger`) are flagged for prod dependencies because of `root package.json` -> `@wordpress/react-native-aztec` -> `react-native`. Ignoring them for now.

swissspidy:
@gziolo This seems to be good to go now

gziolo (approved):
It isn't ideal to keep growing the list of ignored libraries, but it's all very difficult to find a good middle-ground with React Native dependencies used during the development that are at the same part of the platform that produces production code 🤷🏻

PR #61870, opened 2024-05-22
DataViews: Remove unnecessary dependency for pattern fields memo

Mamaduka:
## What?
PR removes unnecessary dependency for the `fields` memo in the `DataviewsPatterns` component. Fixes ESLint warning:

```
React Hook useMemo has an unnecessary dependency: 'categoryId'. Either exclude it or remove the dependency array.
```

## Testing Instructions
CI checks should be green.


ellatrix (approved):

t-hamano (approved):
Nice catch 👍

PR #61871, opened 2024-05-22
Automatically select group variation if there is only one available

### LINKED ISSUE #62239: Don't show group block's placeholder when there is only one variation available

When someone limits a group block's variations so that there is only one left. The block still shows a placeholder and asks a user to pick a variation. The placeholder step could be skipped and the only variation could be used directly.

The real-world use-case is the email editor, where we are not able to support Flex or Grid because these are not supported by email clients.

Screen Capture on 2024-05-22 at 16-50-25

The block could select the single variation automatically and skip the placeholder step.

Screen Capture on 2024-05-22 at 16-52-43

There is a PR that adds this behavior https://github.com/WordPress/gutenberg/pull/61871.

costasovo:
## What?
When a core/group block has only one variation available, let's skip the placeholder step and select the variation.

## Why?
We are building an email editor, but many email clients don't support modern grids or flex layouts. Therefore, we've disabled most variations, leaving us only with one. It feels weird to display the placeholder step with only one option.

## How?
I added an effect into the placeholder component in which I select the variation if there is only one available. I was thinking about adding this logic directly to the Edit component, but it would require calling select for variants for every Edit render so I rather added the change directly into the Placeholder. I'm happy to change it based on the feedback.

### Default state (multiple variations)
1. Insert a group block via inserter
2. See placeholder with multiple variations

### With only one variation

1. Disable all group variations but one:
```js
addFilter(
    'blocks.registerBlockType',
    'mailpoet-email-editor/disable-group-variations',
    ( settings, name ) => {
      if ( name === 'core/group' ) {
        return {
          ...settings,
          variations: settings.variations.filter(
            ( variation ) => variation.name === 'group',
          ),
        };
      }
      return settings;
    },
  );
```
2. Insert the group block via the inserter
3. See that the placeholder step was skipped

The appender in the empty group block is accessible using arrows in the canvas

### Before
Screen Capture on 2024-05-22 at 16-50-25

### After
Screen Capture on 2024-05-22 at 16-52-43

Closes #62239

### DIFF
--- a/packages/block-library/src/group/placeholder.js
+++ b/packages/block-library/src/group/placeholder.js
@@ -139,6 +139,13 @@ function GroupPlaceHolder( { name, onSelect } ) {
 	const blockProps = useBlockProps( {
 		className: 'wp-block-group__placeholder',
 	} );
+
+	useEffect( () => {
+		if ( variations && variations.length === 1 ) {
+			onSelect( variations[ 0 ] );
+		}
+	}, [ onSelect, variations ] );
+
 	return (
 		<div { ...blockProps }>
 			<Placeholder


ockham (approved):
Thank you! 

> I added an effect into the placeholder component in which I select the variation if there is only one available. I was thinking about adding this logic directly to the Edit component, but it would require calling select for variants for every Edit render so I rather added the change directly into the Placeholder. I'm happy to change it based on the feedback.

Ah yes, this is an interesting question. Looking at the Group block's code, it seems to have been structured to contain the checks that determine if a placeholder should been shown in the `useShouldShowPlaceHolder` hook, whereas the Placeholder component doesn't perform any such checks.

Consequently, it's somewhat tempting to preserve this division -- i.e. move the variations check into `useShouldShowPlaceHolder`. However, that's unfortunately not the end of the story; we'll also need to add logic to the Edit component itself (probably another logical branch [here](https://github.com/WordPress/gutenberg/blob/bdd11d5b1341769981594d4c87b56a9604f0c74c/packages/block-library/src/group/edit.js#L112-L122)). The Edit component will then also need to know the block variations, as will the `useShouldShowPlaceHolder` hook. I'm not too worried about the performance impact, but it would significantly balloon the code that's needed for this otherwise simple change. It's dubious if shoehorning this new check into the existing structure of the block will really make the code that much more readable if all these changes are needed.

So let's be pragmatic instead and go with your approach 👍

ockham:
Rebased (to get tests to pass 🤞)

costasovo:
@ockham Thank you for the review, rebasing and the feedback 🙇 

> Consequently, it's somewhat tempting to preserve this division -- i.e. move the variations check into useShouldShowPlaceHolder. However, that's unfortunately not the end of the story; we'll also need to add logic to the Edit component itself (probably another logical branch [here](https://github.com/WordPress/gutenberg/blob/bdd11d5b1341769981594d4c87b56a9604f0c74c/packages/block-library/src/group/edit.js#L112-L122)). The Edit component will then also need to know the block variations, as will the useShouldShowPlaceHolder hook.

I started with this approach, and after I made a couple of these changes, I decided to go with the simple approach. My thoughts were pretty similar.

PR #60560, opened 2024-04-08
Tabs: indicator animation

DaniGuardiola:
## What?

Makes the `Tabs` indicator animated (except when the user prefers reduced motion).

## Why?

Looks cool. @mirka can probably add more context :)

## How?

See styles. Position and length is synced through the Ariakit state store.

## Testing Instructions

Play with it on Storybook. Optionally simulate the "prefers reduced motion" setting through devtools.

### Testing Instructions for Keyboard

Tab into the component, then use arrow keys to change tabs.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/8862967/c03f6abc-f331-4711-b94d-8c801b41deaf

In Windows high-contrast mode:

https://github.com/WordPress/gutenberg/assets/8862967/9c11703a-7fda-4a4d-8445-e704543d4457

PR #61872, opened 2024-05-22
Try: Add CSS Custom Properties to CSS types

DaniGuardiola:
Context: https://github.com/WordPress/gutenberg/pull/60560#discussion_r1610059640

Through this augmentation, CSS Custom Properties are supported (type-wise) in the `style` prop of React elements. They need to start with `--`. This is only applied to the components package, and could be extended to other packages by just copying the ambient declaration file.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Internal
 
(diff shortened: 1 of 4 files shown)


mirka (approved):
Good one 👍 I'm fine with keeping it components-only for now, since there doesn't seem to be any relevant usage outside of the package.

tyxla:
Thank you @DaniGuardiola, this has been bugging me for ages 🙌

Issue #60991, opened 2024-04-23
Improve Inserter experience when blocks cannot be inserted


jameskoster:
### The problem

When opened, the Inserter checks where to place blocks and patterns. When the logic is unable to find an appropriate position then the UI turns up empty. 

A common example is page editing in the site editor with the template preview turned on; if you select the Title or Featured Image block, then open the Inserter, you see this:

<img width="363" alt="Screenshot 2024-04-22 at 14 50 49" src="https://github.com/WordPress/gutenberg/assets/846565/50010f22-ff97-4b51-8b36-280229fe6276">

It's quite unexpected to be presented with an empty Inserter, and the lack of explanation further confuses things.

### Solution – always show blocks / patterns in the Inserter

- [ ] When the block/pattern cannot be inserted in the target location (based on existing logic), place it in the 'closest' possible location instead.
- [ ] When there are legitimately no insertable locations in the current document, show a modal when a block/pattern is clicked/dragged to inform the user.
- [ ] Give additional prominence `allowedBlocks`. E.g. when a List block (or a descendant thereof) is selected, opening the Inserter could show:

<img width="349" alt="A screenshot of the block Inserter with a new section titled 'Suggested' above all other block categories" src="https://github.com/WordPress/gutenberg/assets/846565/1d83ea20-f297-4256-a582-ca4c44008b5d">

The label can be polished in a PR. "Based on your selection" could be an alternative to "Suggested".

<details><summary>Original idea</summary>
Ideally the Inserter should never appear empty, so as to eliminate that initial point of confusion.

One option could be to update the logic so that it falls back to the _closest_ insertable location based on the current parameters. In the scenario above, 95% of the time that would result in blocks being added to the Content block which seems reasonable and expected. 

When there are genuinely no insertable locations in the current document, a dialog could appear when a block / pattern is clicked or dragged that explains the situation.</details>

cc @WordPress/gutenberg-design for other ideas.

jasmussen:
There's a bit more to it than that, not only does nothing show up, but if you search for blocks, you won't find anything, presumably because you can't insert in that container:
![screenshot of the site editor with the title selected and an empty inserter](https://github.com/WordPress/gutenberg/assets/1204802/c9f6f94c-c9a3-4c88-92d7-3e7e9b4e59ad)

Both aspects of the same problem. One path forward is, as you say, finding the nearest insertable location, and then applying both to the inserter and search results, as well as _where the block will be inserted_. 

[...]

PR #61873, opened 2024-05-22
Show all blocks all the times to avoid confusing UX

draganescu:
## What?

Explores #60991 

## Why?

Too often the inserter is empty, for instance when opening the inserter while a list block list item is selected or a button in a buttons block, the only available block is the list item , or the button respectively.

This is confusing because it's unclear why the other blocks are not available.

## How?

Try to show all the blocks all the time and insert blocks in the nearest possible location.

## Testing Instructions

1. In the site editor or the post editor add a list block
2. Add a list item
3. Open the inserter
4. **Notice all blocks are listed**
5. Insert a block that is _not_ the list item block, e.g. an image block
6. **Notice the image block is inserted right after the parent list block of the currently selected list item block**

N/A

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/107534/ffd8aa77-f1fe-4be2-a3f3-77134bc5904a

# Known limitations

- the inserter is still empty when a block in a pattern with overrides is selected and no insertion is possible
- https://github.com/WordPress/gutenberg/pull/61873/files#r1611877561

### DIFF
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -31,15 +31,6 @@ export function BlockTypesTab(
 	{ rootClientId, onInsert, onHover, showMostUsedBlocks },
 	ref
 ) {
-	// Find the root of what we _can_ insert into for this block.
-	// const { rootContentBlockId } = useSelect( ( select ) => {
-	// 	// Try the empty root first.
-	// 	const { getAllowedBlockTypes } = select( 'core/block-editor' );
-
-	// 	select( 'core/block-editor' ).getInserterItems( rootClientId );
-
-	// }, [] );
-
 	const [ allItems, items, categories, collections, onSelectItem ] =
 		useBlockTypesState( rootClientId, onInsert );
 
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -3,20 +3,17 @@
  */
 import { __, _x } from '@wordpress/i18n';
 import { useMemo, useEffect, forwardRef } from '@wordpress/element';
-import { useSelect } from '@wordpress/data';
 import { pipe, useAsyncList } from '@wordpress/compose';
 
 /**
  * Internal dependencies
  */
-import { store as blockEditorStore } from '../../store';
 import BlockTypesList from '../block-types-list';
 import InserterPanel from './panel';
 import useBlockTypesState from './hooks/use-block-types-state';
 import InserterListbox from '../inserter-listbox';
 import { orderBy } from '../../utils/sorting';
 import InserterNoResults from './no-results';
-import useInsertionPoint from './hooks/use-insertion-point';
 
 const getBlockNamespace = ( item ) => item.name.split( '/' )[ 0 ];
 
@@ -34,38 +31,8 @@ export function BlockTypesTab(
 	{ rootClientId, onInsert, onHover, showMostUsedBlocks },
 	ref
 ) {
-	const [
-		items,
-		categories,
-		collections,
-		onSelectItem,
-		allItems,
-		createNewBlock,
-	] = useBlockTypesState( rootClientId, onInsert );
-
-	const insertionIndex = useSelect( ( select ) => {
-		const { getBlockOrder } = select( blockEditorStore );
-
-		// Last item.
-		return getBlockOrder().length;
-	} );
-
-	const [ , onInsertBlocks ] = useInsertionPoint( { insertionIndex } );
-
-	const onSelectGeneralItem = (
-		{ name, initialAttributes, innerBlocks, syncStatus, content },
-		shouldFocusBlock
-	) => {
-		const newBlock = createNewBlock( {
-			content,
-			name,
-			initialAttributes,
-			innerBlocks,
-			syncStatus,
-		} );
-
-		onInsertBlocks( newBlock, undefined, shouldFocusBlock );
-	};
+	const [ items, categories, collections, onSelectItem, allItems ] =
+		useBlockTypesState( rootClientId, onInsert );
 
 	const suggestedItems = useMemo( () => {
 		return orderBy( items, 'frecency', 'desc' ).slice(
@@ -223,7 +190,7 @@ export function BlockTypesTab(
 						<InserterPanel title={ _x( 'All blocks', 'blocks' ) }>
 							<BlockTypesList
 								items={ allItems }
-								onSelect={ onSelectGeneralItem }
+								onSelect={ onSelectItem }
 								onHover={ onHover }
 								label={ _x( 'All blocks', 'blocks' ) }
 							/>
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -31,8 +31,31 @@ export function BlockTypesTab(
 	{ rootClientId, onInsert, onHover, showMostUsedBlocks },
 	ref
 ) {
-	const [ items, categories, collections, onSelectItem, allItems ] =
-		useBlockTypesState( rootClientId, onInsert );
+	const [ items, categories, collections, onSelectItem ] = useBlockTypesState(
+		rootClientId,
+		onInsert
+	);
+
+	// TODO: Split this into its own component that has all items?
+	const [ allItems, allCategories ] = useBlockTypesState( '', () => {} );
+
+	const allItemsPerCategory = useMemo( () => {
+		return pipe(
+			( itemList ) =>
+				itemList.filter(
+					( item ) => item.category && item.category !== 'reusable'
+				),
+			( itemList ) =>
+				itemList.reduce( ( acc, item ) => {
+					const { category } = item;
+					if ( ! acc[ category ] ) {
+						acc[ category ] = [];
+					}
+					acc[ category ].push( item );
+					return acc;
+				}, {} )
+		)( allItems );
+	}, [ allItems ] );
 
 	const suggestedItems = useMemo( () => {
 		return orderBy( items, 'frecency', 'desc' ).slice(
@@ -133,7 +156,11 @@ export function BlockTypesTab(
 					return (
 						<InserterPanel
 							key={ category.slug }
-							title={ category.title }
+							title={
+								categoryItems.length === 1
+									? __( 'Allowed' ) // Maybe check the parent block name and use that instead, like "List Block"?
+									: category.title
+							}
 							icon={ category.icon }
 						>
 							<BlockTypesList
@@ -186,15 +213,25 @@ export function BlockTypesTab(
 
 				{ ( items.length === 0 ||
 					availableCategories.length === 1 ) && (
-					<div className="block-editor-inserter__tips">
-						<InserterPanel title={ _x( 'All blocks', 'blocks' ) }>
-							<BlockTypesList
-								items={ allItems }
-								onSelect={ onSelectItem }
-								onHover={ onHover }
-								label={ _x( 'All blocks', 'blocks' ) }
-							/>
-						</InserterPanel>
+					<div className="block-editor-inserter__all-blocks">
+						{ allCategories.map( ( category ) => {
+							const categoryItems =
+								allItemsPerCategory[ category.slug ];
+							return (
+								<InserterPanel
+									key={ category.slug }
+									title={ category.title }
+									icon={ category.icon }
+								>
+									<BlockTypesList
+										items={ categoryItems }
+										onSelect={ onSelectItem }
+										onHover={ onHover }
+										label={ category.title }
+									/>
+								</InserterPanel>
+							);
+						} ) }
 					</div>
 				) }
 			</div>
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -14,6 +14,7 @@ import useBlockTypesState from './hooks/use-block-types-state';
 import InserterListbox from '../inserter-listbox';
 import { orderBy } from '../../utils/sorting';
 import InserterNoResults from './no-results';
+import useInsertionPoint from './hooks/use-insertion-point';
 
 const getBlockNamespace = ( item ) => item.name.split( '/' )[ 0 ];
 
@@ -31,10 +32,14 @@ export function BlockTypesTab(
 	{ rootClientId, onInsert, onHover, showMostUsedBlocks },
 	ref
 ) {
-	const [ items, categories, collections, onSelectItem ] = useBlockTypesState(
-		rootClientId,
-		onInsert
-	);
+	const [ items, categories, collections, onSelectItem, allItems ] =
+		useBlockTypesState( rootClientId, onInsert );
+
+	const [ , onInsertBlocks ] = useInsertionPoint( {} );
+
+	const onSelectGeneralItem = ( blocks, meta, shouldForceFocusBlock ) => {
+		onInsertBlocks( blocks, meta, shouldForceFocusBlock );
+	};
 
 	const suggestedItems = useMemo( () => {
 		return orderBy( items, 'frecency', 'desc' ).slice(
@@ -104,6 +109,14 @@ export function BlockTypesTab(
 	if ( ! items.length ) {
 		return <InserterNoResults />;
 	}
+
+	const availableCategories = categories.filter( ( category ) => {
+		const categoryItems = itemsPerCategory[ category.slug ];
+		if ( ! categoryItems || ! categoryItems.length ) {
+			return false;
+		}
+		return category;
+	} );
 
 	return (
 		<InserterListbox>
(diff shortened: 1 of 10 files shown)


packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
```diff
@@ -119,10 +120,36 @@ function useInsertionPoint( {
...
+					: [ blocks ];
+
+				let _destinationIndex = destinationIndex;
+				let _destinationRootClientId = destinationRootClientId;
+				// Check if it's possible to insert the block in the current location
+				const canInsertBlocks = normalizedBlocks.every( ( block ) =>
+					canInsertBlockType( block.name, _destinationRootClientId )
+				);
+
+				if ( ! canInsertBlocks && normalizedBlocks.length === 1 ) {
```
jeryj: Why is this only if the block length is 1? What if we're inserting a group of blocks?

draganescu: For now I think we should default to this case until it's ironed out.

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
```diff
@@ -119,10 +120,36 @@ function useInsertionPoint( {
...
+							_destinationRootClientId
+						)
+					) {
+						_destinationIndex =
+							getBlockIndex( _destinationRootClientId ) + 1;
+						_destinationRootClientId = getBlockRootClientId(
+							_destinationRootClientId
+						);
+					}
+				}
```
draganescu: This procedure will crash the browser if there really is no place to insert in. This needs to be handled.

annezazu:
Testing this out and it works well! The only thing I'm confused by is the labeling. In this, when adding a list block, it has the label "TEXT" when I'd expect something like "ALLOWED" "SUGGESTED" or "LIST BLOCK" as [shown in the original designs](https://github.com/WordPress/gutenberg/issues/60991#issuecomment-2114746204). 

Screenshot 2024-05-23 at 11 17 40 AM

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
jeryj: Should be handled by checking [null on the `_destinationRootClientId`](https://github.com/WordPress/gutenberg/pull/61873/files#diff-16ab90c30e615c2cfba050dda8bfa30642629bb53adcb73195fc5053ad576f30R145-R148)

packages/block-editor/src/components/inserter/hooks/use-block-types-state.js
```diff
@@ -23,6 +23,26 @@ import { store as blockEditorStore } from '../../../store';
...
+		const { getBlocksByName, getInserterItems } =
+			select( blockEditorStore );
+
+		// Try the empty root first.
+		const rootInserterItems = getInserterItems( '' );
+		if ( rootInserterItems.length ) {
+			return [ rootInserterItems ];
+		}
+
+		const postContentBlock = getBlocksByName( 'core/post-content' );
```
draganescu: Yea this adds a soft dependency to the WordPress block library in the agnostic block editor package. That should be another better way, ideally "if we can't find insertable items up let's look down until we can".

I tried to use current block selection instead of referring to a block by name, assuming this means "check for this particular block which somehow has locked parents"

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
draganescu: I think this would suffice and treat a fallback. 
```
const _previousDestinationRootClientId = _destinationRootClientId;
... // get index, get new destination
if ( _previousDestinationRootClientId === _destinationRootClientId) {
        break;
}
```

packages/block-editor/src/components/inserter/hooks/use-block-types-state.js
jeryj: Agreed. It'll probably be useful to move something like this into a selector in the block editor store like "getClosestInsertionPoint"

draganescu: for now I think something like:
```
const [ allItems ] = useSelect( ( select ) => {
               let availableItems = select( blockEditorStore ).getInserterItems( '' );
               // use current selection as root for situations like
               // template locked mode
               if ( availableItems ) {
                       const selectedBlock =
                               select( blockEditorStore ).getSelectedBlockClientId();
                       availableItems = select( blockEditorStore ).getInserterItems( selectedBlock );
               }
               return [ availableItems ];
       } );
```

works good enough.

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
```diff
@@ -142,9 +143,17 @@ function useInsertionPoint( {
 						_destinationRootClientId
 					);
 
-					// If we are at the root, there's nothing more to check
 					if ( _destinationRootClientId === null ) {
-						break;
+						// If we are at the root, try to find a post content block
+						const postContentBlock =
+							getBlocksByName( 'core/post-content' );
+						if ( postContentBlock.length ) {
```
draganescu: This too should go away.

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
draganescu: The end result should be like:
```
if ( ! canInsertBlocks && normalizedBlocks.length === 1 ) {
        // If it's not possible to insert the block in the current 
location, try to find a new location up the tree
        while (
                ! canInsertBlockType(
                        normalizedBlocks[ 0 ].name,
                        _destinationRootClientId
                )
        ) {
                // save the index of the previous destination root client ID
                const _previousDestinationRootClientId = _destinationRootClientId;
                _destinationIndex = getBlockIndex( _previousDestinationRootClientId ) + 1;

                // get the parent of the current destination root client ID
                _destinationRootClientId = getBlockRootClientId(
                        _destinationRootClientId
                );

                // Break the loop if we searched the whole tree
                if (
                        _previousDestinationRootClientId ===
                        _destinationRootClientId
                ) {
                        break;
                }
        }
}
```

jeryj:
We have gotten it to work, but it needs refactoring to be more stable and easier to maintain.


This code in [`useBlockTypesState` to get allItems](https://github.com/WordPress/gutenberg/pull/61873/files#diff-fd6c860a990e16ac7433b91dafc257b517e854d1800bc60f296250eeb752e14bR26-R43) is trying to find all possible blocks. Can we get a block list of all possible blocks, independent of any root block? Or can we only get insertion items in relation to a block? If we can return all possible registered blocks, this would be ideal, as we don't really care about the currently selected block or root block at this point in the code. 


- The code at the top of `use-insertion-point` that sets the `destinationRoot` and `destinationRootIndex` is in relation to the current block, and it is assuming that only allowed blocks will be showing in the inserter. This is no longer the case. The insertion point needs to be adapted to whatever block in the inserter is being hovered or focused. We need a method like `getClosestInsertionPoint( selectedBlockId, blockTypeToBeInserted )` so we can show the correct insertion point for the block that is going to be inserted.

`getClosestInsertionPoint()` can be used within `onInsertBlocks`  in `use-insertion-point` so it inserts in the same location. Or, maybe `getClosestInsertionPoint` sets the `destinationRoot` and `destinationRootIndex` we can rely on that. 

So, to refactor all this effectively, I think we need:
- ~`getBlockTypes()` independent of any block. (I believe this already exists).~ This does exist, but the inserter list item does a lot of work that is related to being able to insert it into the canvas somewhere. It needs more context than `getBlockTypes` can provide it. I think we'll need `getRootInsertionPoint` or something similar to has a stable insertion point where any block is allowed to be inserted.
- `getClosestInsertionPoint( selectedBlockId, blockTypeToBeInserted )`

jeryj:
@draganescu Maybe we can modify [getInsertionPoint](https://github.com/WordPress/gutenberg/blob/try/show-all-blocks/packages/editor/src/store/private-selectors.js#L38) for our needs. It's also passed into the inserter already. It's always returning undefined for me at the moment, but maybe it'll be useful.

packages/block-editor/src/components/inserter/hooks/use-block-types-state.js
```diff
@@ -23,6 +23,25 @@ import { store as blockEditorStore } from '../../../store';
...
  */
 const useBlockTypesState = ( rootClientId, onInsert ) => {
+	const [ allItems ] = useSelect( ( select ) => {
+		let availableItems = select( blockEditorStore ).getInserterItems( '' );
+
+		// use current selection as root for situations like
+		// template locked mode
+		const rootBlocks = select( blockEditorStore ).getBlocks();
+		while ( availableItems.length === 0 ) {
+			for ( const block of rootBlocks ) {
```
youknowriad, line 33: This loop feels like it can be very bad perf wise in nested templates and things like that. (not sure how true though).

The other thing that feels odd, is that it will end up with a random client id and the behavior for the user can be confusing: where am I going to insert when I click that thing?

It feels like a better path, would be to try to find the "rootClientId" that we want to insert in before hand and update `useBlockTypesState` to take a rootClientId as an argument. Now that I'm writing that, it seems it already has a rootClientId, so it's a matter of extracting a "BlockList" component from the current BlockTab component and doing something like:

```
const { rootClientId, alternativeRootClientId } = useSelect(() => {
   return {
       rootClientId: getInsertionPoint().rootClientId,
       alternativeRootClientId: findAlternativeRootClientId() // we can discuss implementation of this function separately.
   }
}, [] );

<BlockList rootClientId={ rootClientID } />
<BlockList rootClientId={ alternativeRootClientId } />
```

draganescu: Yes by all means there a couple of risky loops but they're in place just to make things "work" and to show conceptual limitations we still have.

For example a loop later in the PR will make the action of inserting a block while a post content block is selected in a template locked edit mode and it shows that we can't tell where to insert without:

- bringing knowledge of template locked mode to the block editor package
- relying on something else in a higher level package finding "post content" and its client id and pass that down to lower levels

... both of which are kinda weird.

Separately, the theoretical `findAlternativeRootClientId` will also have to try a sort of heuristic since we don't have a clear picture of how to find the next best insertable spot.

Maybe the block list should advertise this in a separate precomputed part of the store? As in top down places to insert blocks if the next thing is unavailable ... or something.

youknowriad: There are two separate discussions here:

 - First; I don't think the way this PR is implemented is correct. It's just adding code to make things work as it want but without organizing things properly. This is for instance forcing us to have this loop in two places (find the right place). 
 - Second: how `findAlternativeRootClientId` should be implemented. There are options: smart as the PR is trying to do (find the closest parent or something), or more controlled: (use the same wrapper the zoom-out would use for instance)

draganescu: > I don't think the way this PR is implemented is correct. It's just adding code to make things work as it want but without organizing things properly

Yes, that's why I said: "by all means there a couple of risky loops but they're in place just to make things "work" and to show conceptual limitations we still have." and @jeryj said "We have gotten it to work, but it needs refactoring to be more stable and easier to maintain." 😂

youknowriad: Yeah, sorry I misread the original reply :)

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -83,6 +102,19 @@ export function BlockTypesTab(
...
 	useEffect( () => () => onHover( null ), [] );
 
+	const getRootBlockTitle = () => {
+		const blockName = getBlockName( rootClientId );
+		// Find the title within the list of all possible blocks.
+		const block = allItems.find( ( item ) => item.name === blockName );
+		return block
+			? sprintf(
+					// translators: %s: Block title, e.g: "List", "Buttons", "Group"
+					__( '%s Block' ),
```
richtabor, line 112: We don't need "Block" appended here. The block name itself will be enough context. "List", "Columns", "Buttons", etc.

richtabor:
UX wise, this feels pretty good. You don't end up hung up with no blocks to insert, like this: 

CleanShot 2024-05-30 at 21 25 11

richtabor:
Odd border when the allowed blocks is not necessary: 

CleanShot 2024-05-30 at 21 25 58

It may be fine to omit the border even. None of the other categories have one.

richtabor:
Noticed that "Patterns" renders at the bottom as well: 

CleanShot 2024-05-30 at 21 27 31

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -31,10 +33,27 @@ export function BlockTypesTab(
...
+		useBlockTypesState( rootClientId, onInsert );
+
+	const allItemsPerCategory = useMemo( () => {
+		return pipe(
+			( itemList ) =>
+				itemList.filter(
+					( item ) => item.category && item.category !== 'reusable'
+				),
+			( itemList ) =>
+				itemList.reduce( ( acc, item ) => {
```
ellatrix, line 47: Why loop over this twice and add `pipe` if it can be done in a single loop?

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -83,6 +102,19 @@ export function BlockTypesTab(
 	// Hide block preview on unmount.
 	useEffect( () => () => onHover( null ), [] );
 
+	const getRootBlockTitle = () => {
+		const blockName = getBlockName( rootClientId );
+		// Find the title within the list of all possible blocks.
+		const block = allItems.find( ( item ) => item.name === blockName );
```
ellatrix, line 108: Are we getting the block title here? Why not check the block type instead of looping through all blocks, like we normally do?

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -127,7 +164,11 @@ export function BlockTypesTab(
 					return (
 						<InserterPanel
 							key={ category.slug }
-							title={ category.title }
+							title={
+								categoryItems.length === 1
+									? getRootBlockTitle()
+									: category.title
```
ellatrix, line 170: I don't understand this condition. Why do we fall back to the title of the parent of the selected block when there's 1 category?

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -177,6 +218,30 @@ export function BlockTypesTab(
 						);
 					}
 				) }
+
+				{ ( items.length === 0 ||
+					availableCategories.length === 1 ) && (
```
ellatrix, line 223: Would be great to comment on this condition. Why show this when there's 1 category available and not 2?

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
```diff
@@ -145,18 +167,21 @@ function useInsertionPoint( {
...
+			setLastFocus,
 		]
 	);
 
 	const onToggleInsertionPoint = useCallback(
 		( show ) => {
 			if ( show ) {
-				showInsertionPoint( destinationRootClientId, destinationIndex );
+				showInsertionPoint( destinationRootClientId, destinationIndex, {
+					block: blockToInsert,
```
ellatrix, line 183: I assume we need to pass the block so that we can change the target if necessary. Can't be change `destinationRootClientId` and `destinationIndex` so we avoid modifying `showInsertionPoint`?

packages/block-editor/src/store/actions.js
```diff
@@ -603,22 +603,76 @@ export const insertBlocks =
...
+			const canInsertBlocks = () =>
+				normalizedBlocks.every( ( blockToInsert ) =>
+					select.canInsertBlockType(
+						blockToInsert.name,
+						_destinationRootClientId
+					)
+				);
+			if ( ! canInsertBlocks() && normalizedBlocks.length === 1 ) {
+				// If it's not possible to insert the block in the current location, try to find a new location up the tree
+				const rootBlocks = select.getBlocks();
```
ellatrix, line 630: Why do we jump straight from the selected block to the root block instead of attempting to insert into parent blocks? Or am I missing something?

packages/block-editor/src/components/inserter/block-types-tab.js
ellatrix: Ah, I got confused by `categoryItems`, I'm guessing this is just the total amount of items, not the amount of categories.

youknowriad:
Where does this PR stand at the moment? What's the logic being implemented? Any thoughts on the proposal here? https://github.com/WordPress/gutenberg/pull/61873#discussion_r1618577527

packages/block-editor/src/components/inserter/block-types-tab.js
draganescu: > Why show this when there's 1 category available and not 2?

Pure convention 🤷🏻

draganescu:
Let's close this in favor of #62169 - it's the same idea overall but solves the spread of changes to a simpler location. Testing that PR it achieves the same result. It also improves upon our solution here around inserting when a template locked situation exists, by defaulting to the section root (nice!) as the place to insert into. It also solves the same problem for patterns with overrides. So it's a better continuation of the work here.

@priethor heads up that #62169 is now the bug fix for #60991

PR #61876, opened 2024-05-22
Editor: Ensure Copy button in sidebar copies whole permalink, *with* URL protocol

mcsf:
`usePostURLLabel` is meant to be used in labels and intentionally strips the protocol of a URL. As a result, the copy button in the `PostURL` component wrote something like "example.org/foobar" to the clipboard, instead of "https://example.org/foobar".

As a fix, just grab the permalink directly.

Screenshot 2024-05-22 at 19 20 29

### DIFF
--- a/packages/editor/src/components/post-url/index.js
+++ b/packages/editor/src/components/post-url/index.js
@@ -21,7 +21,6 @@ import { useCopyToClipboard } from '@wordpress/compose';
 /**
  * Internal dependencies
  */
-import { usePostURLLabel } from './label';
 import { store as editorStore } from '../../store';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/editor/src/components/post-url/index.js
```diff
@@ -62,8 +61,11 @@ export default function PostURL( { onClose } ) {
 	const { editPost } = useDispatch( editorStore );
 	const { createNotice } = useDispatch( noticesStore );
 	const [ forceEmptyField, setForceEmptyField ] = useState( false );
-	const postUrlLabel = usePostURLLabel();
-	const copyButtonRef = useCopyToClipboard( postUrlLabel, () => {
+	const permalink = useSelect(
+		( select ) => select( editorStore ).getPermalink(),
+		[]
+	);
+	const copyButtonRef = useCopyToClipboard( permalink, () => {
```
ntsekouras, line 71: We should also use `safeDecodeURIComponent` for the copied permalink.

ntsekouras (commented):
Good catch, thank you! I've left a small comment and it's good to 🚢

packages/editor/src/components/post-url/index.js
mcsf: @ntsekouras: Is it needed in practice? I had a quick look at `getPermalink`, `getPermalinkParts` and I don't see where we'd get encoded components from. Also:

Screenshot 2024-05-23 at 10 24 02

In the above scenario I see `http://localhost:8881/2024/05/ola-φίλε/` in my clipboard, as expected.

Finally, I also just refactored to use the existing `useSelect` call in that component.

ntsekouras: Hm.. yesterday I tested with some greek and accents and it produced a URL that needed it. Don't remember what I tried exactly though and couldn't reproduce today with random tests, that could be similar to the yesterday ones..

mcsf: So I should just add the decoder to be safe?

ntsekouras (approved):
Thanks Miguel!

Issue #55299, opened 2023-10-12
Mentions Autocomplete is re-rendeing on each keypress


WunderBart:
### Description

Originally mentioned in https://github.com/WordPress/gutenberg/issues/55118#issuecomment-1751040571

The constant rerendering of the results popup is causing issues for accessibility. 

The responsible part seems to be the [popup cloning via `createPortal`](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/autocomplete/autocompleter-ui.tsx#L169-L174) as it does not rerender when that part is removed. Kudos to @chad1008 for pinning it down.

cc: @ellatrix, since you worked on this part recently.

### Step-by-step reproduction instructions

1. create a new post
2. type "@" to trigger the mentions autocomplete popup
3. navigate the results with arrow down/up and notice the rendering

### Screenshots, screen recording, code snippet

https://github.com/WordPress/gutenberg/assets/1451471/a8ab4374-6d5e-48df-a94e-6e4c8d17cd81

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

No

alexstine:
Labeling this as low priority. Not urgent that this gets fixed, just something to improve on later.

jsnajdr:
Fixed by #61877.

PR #61877, opened 2024-05-22
getAutocompleterUI: don't redefine ListBox component on every render

### LINKED ISSUE #50562: Autocomplete: Clicking to insert item does nothing when inside a table cell
### Description

When the autocomplete dropdown is opened inside a table cell, clicking on one of the items inside the dropdown does not insert the item. It also doesn't close the dropdown, but I can no longer navigate the dropdown with my keyboard either.

1. Go to https://wordpress.org/gutenberg/
2. Insert a table. The default dimensions (2x2) are sufficient.
3. In one of the table cells, type `@` to begin inserting a mention. The autocomplete dropdown will open with two users shown.
4. Click on one of the items in the dropdown. Nothing happens.

Screen Recording 2023-05-11 at 10 12 34 PM

### Environment info

- Firefox 113 and Chrome 113 (yes, the version numbers happen to be the same -- not a typo!)
- Macbook Pro with MacOS 13.3.1

Yes

Yes

jsnajdr:
Fixes #50562. The `AutocompleterUI` component defined a nested `ListBox` component inside its render function, but this is an antipattern! The `ListBox` component has different identity on every render, and that causes it to completely unmount and remount on every render.

One of the symptoms is that the `onClick` handler on the `Button` subcomponents in the `ListBox` is never called when the autocompleter is inside the table. Don't ask me why 🙂 There is probably some chaotic unstable rerendering that destroys the `onClick` handler before it gets a chance to be called.

Another symptom that you can easily reproduce is that the gravatar image on the user autocomplete is blinking as you type. Because the items are not rendered in a stable way, the `img` element gets repeatedly destroyed and created and loaded again.

**How to test:**
Verify that the symptoms reported in #50562 (user autocomplete inside table cell) are fixed. Also, the gravatar images should be stabilized.

### DIFF
--- a/packages/components/src/autocomplete/autocompleter-ui.tsx
+++ b/packages/components/src/autocomplete/autocompleter-ui.tsx
@@ -37,48 +37,47 @@ type ListBoxProps = {
 	Component?: React.ElementType;
 };
 
-export function getAutoCompleterUI( autocompleter: WPCompleter ) {
-	const useItems = autocompleter.useItems
-		? autocompleter.useItems
-		: getDefaultUseItems( autocompleter );
+function ListBox( {
+	items,
+	onSelect,
+	selectedIndex,
+	instanceId,
+	listBoxId,
+	className,
+	Component = 'div',
+}: ListBoxProps ) {
+	return (
+		<Component
+			id={ listBoxId }
+			role="listbox"
+			className="components-autocomplete__results"
+		>
+			{ items.map( ( option, index ) => (
+				<Button
+					key={ option.key }
+					id={ `components-autocomplete-item-${ instanceId }-${ option.key }` }
+					role="option"
+					aria-selected={ index === selectedIndex }
+					disabled={ option.isDisabled }
+					className={ clsx(
+						'components-autocomplete__result',
+						className,
+						{
+							'is-selected': index === selectedIndex,
+						}
+					) }
+					onClick={ () => onSelect( option ) }
+				>
+					{ option.label }
+				</Button>
+			) ) }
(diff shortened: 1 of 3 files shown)


ellatrix (approved):
Wouldn't be bad if it also adds an e2e test to the existing suite of autocomplete tests

packages/components/src/autocomplete/autocompleter-ui.tsx
```diff
@@ -27,11 +27,59 @@ import { VisuallyHidden } from '../visually-hidden';
...
+							}
+						) }
+						onClick={ () => onSelect( option ) }
+					>
+						{ option.label }
+					</Button>
+				) ) }
+			</Component>
+		);
+	}
```
ellatrix: Do we need it to be inside of `getAutoCompleterUI`?

alexstine (approved):
Works fine for me. Also fixes this issue: https://github.com/WordPress/gutenberg/issues/55299.

packages/components/src/autocomplete/autocompleter-ui.tsx
jsnajdr: It can be outside! Moved 🙂

jsnajdr:
> Wouldn't be bad if it also adds an e2e test to the existing suite of autocomplete tests

Good idea, added a test that autocompletes inside a table. Verified that it fails on trunk and succeeds after the fix.

I think I addressed all feedback, going to merge now.

tyxla (approved):
Really nice one, @jsnajdr 🙌

PR #61874, opened 2024-05-22
Try having less of the tree within `EditorProvider`

stokesman:
## What?
Trys a naive fix for some animations that are currently broken. It probably can generally make parts of the interface a teeny bit more snappy (less re-renders).

## Why?
Fixes #60875 and fixes #61875.

## How?
Moves `EditorProvider` to wrap less of the interface—just the content.

## Testing Instructions
- See the linked issues and try reproducing them
- Try finding something else this may break

PR #61882, opened 2024-05-22
EntityProvider: Avoid remounts and simplify

### LINKED ISSUE #61875: Document Bar missing the initial animation changing from page/template
### Description

It used to look like this:

https://github.com/WordPress/gutenberg/assets/6458278/279577f4-5b29-4a38-a30e-a718b82d4190

~~A `git bisect` points at #60408 as the change that regressed it.~~

1. Open the (Site) editor
2. Open a page for editing
3. Select “Edit template” in the sidebar
4. Note the lack of animation in the Document Bar

_No response_

### Environment info

- Gutenberg trunk
- Chrome or Safari on Mac OS

Yes

No

youknowriad:
closes #61875 
Alternative to #61874 

## What?

While debugging #61875 I noticed that the EditorProvider remounts its children when the "entity" changes (for instance when moving from a template to a template part). The reason for this is a poor implementation of the `EntityProvider` component. It was basically using some complex code to use a separate React context per entity.

The current PR simplifies things a bit by using the same provider and hold all the "entity providers" in it. Personally, I think ultimately, we should seek to deprecate the EntityProviders entirely in favor of Block Context. It was a bad idea that stuck but that's for another time.

## Testing Instructions

1- Open the site editor
2- Select a template part and click "edit"
3- The document bar should animate forward
4- Click "back" on the document bar and it should animate backwards.

### DIFF
--- a/packages/core-data/src/entity-provider.js
+++ b/packages/core-data/src/entity-provider.js
@@ -20,39 +20,7 @@ import { updateFootnotesFromMeta } from './footnotes';
 
 const EMPTY_ARRAY = [];
 
-/**
- * Internal dependencies
- */
-import { rootEntitiesConfig, additionalEntityConfigLoaders } from './entities';
-
-const entityContexts = {
-	...rootEntitiesConfig.reduce( ( acc, loader ) => {
-		if ( ! acc[ loader.kind ] ) {
-			acc[ loader.kind ] = {};
-		}
-		acc[ loader.kind ][ loader.name ] = {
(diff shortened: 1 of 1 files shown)


stokesman (approved):
I tested and it fixes the issue. Code looks much nicer too.

This is just the fix I hoped for! Those remounts could have been potential problems in many other cases. 🚀

youknowriad:
Thanks for raising these issues @stokesman

PR #61883, opened 2024-05-22
Enable shadow support for cover block

vcanales:
## What?
On previous attempts to do this, the cover block was displaying inconsistencies between how it was rendered in the editor vs the frontend; that doesn't seem to be the case anymore, so we should be enable shadow support without any modifications to the block itself.

I'd still appreciate some eyes on this, because I don't know what exactly has changed since the first time I attempted this.

## Testing Instructions
1. Add a cover block
2. Add a shadow to it
3. Modify the block's border radius and size; modifications should be consistent between frontend and editor.

## Screenshots or screencast 
image


madhusudhand (approved):
Shadow applies to cover block both in global styles and block settings.

LGTM! 🚢

PR #61885, opened 2024-05-23
Introduce `wp-on-async` directive as performant alternative over synchronous `wp-on` directive

### LINKED ISSUE #61634: Event directives can only be synchronous without the option for them to be asynchronous/passive
Please see discussion in https://github.com/WordPress/gutenberg/discussions/60574#discussioncomment-9060900.

Interactivity performance on the web is heavily impeded by too many event handlers running synchronously in response to a user interaction. This is made evident by poor responsiveness on web pages which is also reflected in poor [Interaction to Next Paint (INP)](https://web.dev/articles/inp) scores. Too much JavaScript running at once results in long tasks (>50ms) which a user can experience as jank (e.g. stuttering animations) or worse: rage clicks. To help guard against this as much as possible, event handlers should be broken up into chunks, each of which take less than 50 ms to complete on modest hardware. Nevertheless, even if one event handler is doing its part to take less than 50 ms, if there are multiple event handlers which are also running (and which may be less than 50 ms), it is most likely that they will all add up to a long task that hurts the user experience. (This is a real possibility as of https://github.com/WordPress/gutenberg/pull/60661.)

The issue is that by default an event handler runs synchronously and blocks the main thread. In the Interactivity API, this is the `wp-on` directive. Sometimes an event handler must be synchronous in order to be able to call `event.preventDefault()`. However, for many use cases there is no need for synchronous access. For example, a common problem on the web is that analytics trackers will attach their event handlers that run synchronously at the same time as critical 1st party code. These should be safely made asynchronous. The majority of directives in the Image block can also be made asynchronous, particularly those which don't call `event.preventDefault()`:

Directive | Handler | Async-eligible?
--|--|--
`data-wp-on--click` | `actions.hideLightbox` | ✅
`data-wp-on--click` | `actions.showLightbox` | ✅
`data-wp-on--keydown` | `actions.handleKeydown` | ❌
`data-wp-on--load` | `callbacks.setButtonStyles` | ✅
`data-wp-on--touchend` | `actions.handleTouchEnd` | ✅
`data-wp-on--touchmove` | `actions.handleTouchMove` | ❌
`data-wp-on--touchstart` | `actions.handleTouchStart` | ✅
`data-wp-on-window--resize` | `callbacks.setButtonStyles` | ✅
`data-wp-on-window--resize` | `callbacks.setOverlayStyles` | ✅
`data-wp-on-window--scroll` | `actions.handleScroll` | ✅

As for implementation, it could look very similar to the [existing `data-wp-on` directive logic](https://github.com/WordPress/gutenberg/blob/af0394916c13a9f250db7cdf3f521852752b8e4a/packages/interactivity/src/directives.js#L282-L302) but simply adding in a statement to yield before invoking each handler:

```js
	// data-wp-on-async--[event]
	directive( 'on-async', ( { directives: { on }, element, evaluate } ) => {
		const events = new Map();
		on.filter( ( { suffix } ) => suffix !== 'default' ).forEach(
			( entry ) => {
				const event = entry.suffix.split( '--' )[ 0 ];
				if ( ! events.has( event ) ) {
					events.set( event, new Set() );
				}
				events.get( event ).add( entry );
			}
		);

		events.forEach( ( entries, eventType ) => {
			element.props[ `on${ eventType }` ] = async ( event ) => {
				entries.forEach( ( entry ) => {
					await yieldToMain(); // 👈👈👈 KEY ADDITION HERE
					evaluate( entry, event );
				} );
			};
		} );
	} );
```

As discussed in https://github.com/WordPress/gutenberg/discussions/60574#discussioncomment-9060900, having such a `yieldToMain()` helper function would be useful for actions and callbacks to manually insert their own yield points before or after some expensive logic. For an example implementation of `yieldToMain()`, see [Optimize long tasks](https://web.dev/articles/optimize-long-tasks#async-await), which was already implemented in the Interactivity API during hydration in #58227.

Eventually once Preact supports `passive` event listeners (https://github.com/WordPress/gutenberg/issues/58705), such async event handlers could be made passive as well.

Once this is implemented, the core blocks can be updated to use `data-wp-async` instead of `data-wp`. Additionally, @luisherranz noted in https://github.com/WordPress/gutenberg/discussions/60574#discussioncomment-9393724, the Interactivity API docs should be updated to make use of this directive whenever possible, and the create-block template can be updated to use it as well.

westonruter:
Fixes #61634.

## What?

This adds an alternative async variation to the `data-wp-on` directives by tagging on an `-async` to the directive name. So instead of `data-wp-on--click` it can be `data-wp-on-async--click`. This is also done for the `data-wp-on-window` and `data-wp-on-document` directives with `data-wp-on-async-window` and`data-wp-on-async-document` respectively (which here can also be made `passive`).

## Why?

This ensures that when there are multiple directives for the same event, they do not add up to a long task.

## How?

This directive yields to the main thread immediately before invoking the action/callback.

This is implemented for the eligible directives in the Image block. 

## Testing Instructions

1. Install and activate this [Try Slow Events](https://gist.github.com/westonruter/59f7c58230bd5d1e82733e5b87d245dc) plugin which adds 10 click event directives to the `img` in an Image block.
2. Add an Image to a post and enable the click-to-expand feature.
3. Navigate to the post on the frontend.
4. First add `?try-slow-events=sync` to the URL and try opening the image in a lightbox. Notice the latency.
5. Second, try `?try-slow-events=async` to the URL and try opening the image in a lightbox. Notice the latency goes away.

### Before (`?try-slow-events=sync`)

[Screen recording 2024-05-22 17.34.29.webm](https://github.com/WordPress/gutenberg/assets/134745/5ffb9ab4-1706-44f7-989f-812b8c8a35d2)

image

### After (`?try-slow-events=async`)

[Screen recording 2024-05-22 17.34.52.webm](https://github.com/WordPress/gutenberg/assets/134745/19457cf1-a9d0-43a3-8f8e-731007be6cf4)

image

(Aside: The `yieldToMain()` seems it could be improved yet further since a long task is still sneaking in, even though it is much better.)

### DIFF
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -188,12 +188,12 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// Image.
 	$p->next_tag( 'img' );
 	$p->set_attribute( 'data-wp-init', 'callbacks.setButtonStyles' );
-	$p->set_attribute( 'data-wp-on--load', 'callbacks.setButtonStyles' );
-	$p->set_attribute( 'data-wp-on-window--resize', 'callbacks.setButtonStyles' );
+	$p->set_attribute( 'data-wp-on-async--load', 'callbacks.setButtonStyles' );
+	$p->set_attribute( 'data-wp-on-async-window--resize', 'callbacks.setButtonStyles' );
 	// Sets an event callback on the `img` because the `figure` element can also
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
-	$p->set_attribute( 'data-wp-on--click', 'actions.showLightbox' );
+	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
 
 	$body_content = $p->get_updated_html();
 
@@ -209,7 +209,7 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 			aria-haspopup="dialog"
 			aria-label="' . esc_attr( $aria_label ) . '"
 			data-wp-init="callbacks.initTriggerButton"
-			data-wp-on--click="actions.showLightbox"
+			data-wp-on-async--click="actions.showLightbox"
 			data-wp-style--right="context.imageButtonRight"
 			data-wp-style--top="context.imageButtonTop"
 		>
@@ -258,12 +258,12 @@ function block_core_image_print_lightbox_overlay() {
 			data-wp-class--show-closing-animation="state.showClosingAnimation"
 			data-wp-watch="callbacks.setOverlayFocus"
 			data-wp-on--keydown="actions.handleKeydown"
-			data-wp-on--touchstart="actions.handleTouchStart"
+			data-wp-on-async--touchstart="actions.handleTouchStart"
 			data-wp-on--touchmove="actions.handleTouchMove"
-			data-wp-on--touchend="actions.handleTouchEnd"
-			data-wp-on--click="actions.hideLightbox"
-			data-wp-on-window--resize="callbacks.setOverlayStyles"
-			data-wp-on-window--scroll="actions.handleScroll"
+			data-wp-on-async--touchend="actions.handleTouchEnd"
+			data-wp-on-async--click="actions.hideLightbox"
+			data-wp-on-async-window--resize="callbacks.setOverlayStyles"
+			data-wp-on-async-window--scroll="actions.handleScroll"
 			tabindex="-1"
 			>
 				<button type="button" aria-label="$close_button_label" style="fill: $close_button_color" class="close-button">
--- a/packages/interactivity/CHANGELOG.md
+++ b/packages/interactivity/CHANGELOG.md
@@ -2,14 +2,18 @@
 
 ## Unreleased
 
-### Bug fixes
+### New Features
 
--	Fix null and number strings as namespaces runtime error. ([#61960](https://github.com/WordPress/gutenberg/pull/61960/))
+-   Introduce `wp-on-async` directive as performant alternative over synchronous `wp-on` directive. ([#61885](https://github.com/WordPress/gutenberg/pull/61885))
 
 ### Breaking Changes
 
 -   Variables like `process.env.IS_GUTENBERG_PLUGIN` have been replaced by `globalThis.IS_GUTENBERG_PLUGIN`. Build systems using `process.env` should be updated ([#61486](https://github.com/WordPress/gutenberg/pull/61486)).
 -   Increase the minimum required Node.js version to v18.12.0 matching long-term support releases ([#31270](https://github.com/WordPress/gutenberg/pull/61930)). Learn more about [Node.js releases](https://nodejs.org/en/about/previous-releases).
+
+### Bug fixes
+
+-	Fix null and number strings as namespaces runtime error. ([#61960](https://github.com/WordPress/gutenberg/pull/61960/))
 
 ## 5.7.0 (2024-05-16)
 
--- a/packages/interactivity/src/directives.tsx
+++ b/packages/interactivity/src/directives.tsx
@@ -11,7 +11,13 @@ import { deepSignal, peek, type DeepSignal } from 'deepsignal';
 /**
  * Internal dependencies
  */
-import { useWatch, useInit, kebabToCamelCase, warn } from './utils';
+import {
+	useWatch,
+	useInit,
+	kebabToCamelCase,
+	warn,
+	yieldToMain,
+} from './utils';
 import type { DirectiveEntry } from './hooks';
 import { directive, getScope, getEvaluate } from './hooks';
 
@@ -226,6 +232,33 @@ const getGlobalEventDirective = ( type: 'window' | 'document' ) => {
 	};
 };
 
+/**
+ * Creates a directive that adds an async event listener to the global window or
+ * document object.
+ *
+ * @param type 'window' or 'document'
+ */
+const getGlobalAsyncEventDirective = ( type: 'window' | 'document' ) => {
+	return ( { directives, evaluate } ) => {
+		directives[ `on-async-${ type }` ]
+			.filter( ( { suffix } ) => suffix !== 'default' )
+			.forEach( ( entry: DirectiveEntry ) => {
+				const eventName = entry.suffix.split( '--', 1 )[ 0 ];
+				useInit( () => {
+					const cb = async ( event: Event ) => {
+						await yieldToMain();
+						evaluate( entry, event );
+					};
+					const globalVar = type === 'window' ? window : document;
+					globalVar.addEventListener( eventName, cb, {
+						passive: true,
+					} );
+					return () => globalVar.removeEventListener( eventName, cb );
+				} );
+			} );
+	};
+};
+
 export default () => {
 	// data-wp-context
 	directive(
@@ -294,18 +327,60 @@ export default () => {
 		);
 
 		events.forEach( ( entries, eventType ) => {
+			const existingHandler = element.props[ `on${ eventType }` ];
 			element.props[ `on${ eventType }` ] = ( event: Event ) => {
 				entries.forEach( ( entry ) => {
+					if ( existingHandler ) {
+						existingHandler( event );
+					}
 					evaluate( entry, event );
 				} );
 			};
 		} );
 	} );
 
+	// data-wp-on-async--[event]
+	directive(
+		'on-async',
+		( { directives: { 'on-async': onAsync }, element, evaluate } ) => {
+			const events = new Map< string, Set< DirectiveEntry > >();
+			onAsync
+				.filter( ( { suffix } ) => suffix !== 'default' )
+				.forEach( ( entry ) => {
+					const event = entry.suffix.split( '--' )[ 0 ];
+					if ( ! events.has( event ) ) {
+						events.set( event, new Set< DirectiveEntry >() );
+					}
+					events.get( event )!.add( entry );
+				} );
+
+			events.forEach( ( entries, eventType ) => {
+				const existingHandler = element.props[ `on${ eventType }` ];
+				element.props[ `on${ eventType }` ] = ( event: Event ) => {
+					if ( existingHandler ) {
+						existingHandler( event );
+					}
+					entries.forEach( async ( entry ) => {
+						await yieldToMain();
+						evaluate( entry, event );
+					} );
+				};
+			} );
+		}
+	);
+
 	// data-wp-on-window--[event]
 	directive( 'on-window', getGlobalEventDirective( 'window' ) );
 	// data-wp-on-document--[event]
 	directive( 'on-document', getGlobalEventDirective( 'document' ) );
+
+	// data-wp-on-async-window--[event]
+	directive( 'on-async-window', getGlobalAsyncEventDirective( 'window' ) );
+	// data-wp-on-async-document--[event]
+	directive(
+		'on-async-document',
+		getGlobalAsyncEventDirective( 'document' )
+	);
 
 	// data-wp-class--[classname]
 	directive(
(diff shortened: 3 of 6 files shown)


cbravobernal:
Thanks @westonruter for the PR!

My main concern is that developers won't use the async mode as such as needed.

What could be the implications of making it async by default and switching to `wp-on-sync-` for the rest of cases?

How can we guide developers for the best approach, apart from updating the docs?

cbravobernal (dismissed):
Code looks good, and tests are working as expected.

It would be nice to add some e2e or check if we can update the ones we have for the document and window directives.

We still need to update the documentation to include these new directives.

cbravobernal (approved):

westonruter:
@cbravobernal:

> My main concern is that developers won't use the async mode as such as needed.
> 
> What could be the implications of making it async by default and switching to `wp-on-sync-` for the rest of cases?

That would be ideal, if the back-compat breakage is acceptable. Basically, synchronous events are required when passing in an `event` object _and_ when the handler calls `event.preventDefault()`, `event.stopPropagation()`, or `event.stopImmediatePropagation()`. Two of the actions on the Image block call `event.preventDefault()` (as listed in #61634). For example, if `actions.handleKeydown` were to be made async, then no longer would Tab key presses get trapped in the opened lightbox since `event.preventDefault()` would be called too late.

> How can we guide developers for the best approach, apart from updating the docs?

One possibility is that we warn users about switching to use `on-async` when the associated action lacks any argument. Perhaps when `SCRIPT_DEBUG` is enabled, when looping over the directives to add to the `events` Map, it could look up the referenced function and stringify it to then parse out the function's argument list. If it is empty and `on` is used, it could warn that the user should use `on-async` instead. (Maybe the function body could also be looked at to see if it contains `preventDefault`, `stopPropagation`, or `stopImmediatePropagation`.)

westonruter:
> One possibility is that we warn users about switching to use `on-async` when the associated action lacks any argument. Perhaps when `SCRIPT_DEBUG` is enabled, when looping over the directives to add to the `events` Map, it could look up the referenced function and stringify it to then parse out the function's argument list. If it is empty and `on` is used, it could warn that the user should use `on-async` instead. (Maybe the function body could also be looked at to see if it contains `preventDefault`, `stopPropagation`, or `stopImmediatePropagation`.)

Humm. This doesn't seem to work as I expected because the actions/callbacks are getting wrapped in additional functions.

<details><summary>My failed patch</summary>

```diff
diff --git a/packages/interactivity/src/directives.tsx b/packages/interactivity/src/directives.tsx
index a4a320eda57..7af341da558 100644
--- a/packages/interactivity/src/directives.tsx
+++ b/packages/interactivity/src/directives.tsx
@@ -19,7 +19,7 @@ import {
 	yieldToMain,
 } from './utils';
 import type { DirectiveEntry } from './hooks';
-import { directive, getScope, getEvaluate } from './hooks';
+import { directive, getScope, getEvaluate, getDereference } from './hooks';
 
 // Assigned objects should be ignored during proxification.
 const contextAssignedObjects = new WeakMap();
@@ -314,7 +314,7 @@ export default () => {
 	} );
 
 	// data-wp-on--[event]
-	directive( 'on', ( { directives: { on }, element, evaluate } ) => {
+	directive( 'on', ( { directives: { on }, element, evaluate, dereference } ) => {
 		const events = new Map< string, Set< DirectiveEntry > >();
 		on.filter( ( { suffix } ) => suffix !== 'default' ).forEach(
 			( entry ) => {
@@ -328,6 +328,14 @@ export default () => {
 
 		events.forEach( ( entries, eventType ) => {
 			const existingHandler = element.props[ `on${ eventType }` ];
+
+			if ( typeof SCRIPT_DEBUG === 'boolean' && SCRIPT_DEBUG ) {
+				for ( const entry of entries ) {
+					const handler = dereference( entry ).value;
+					console.info(handler.toString());
+				}
+			}
+
 			element.props[ `on${ eventType }` ] = ( event: Event ) => {
 				entries.forEach( ( entry ) => {
 					if ( existingHandler ) {
diff --git a/packages/interactivity/src/hooks.tsx b/packages/interactivity/src/hooks.tsx
index d8aa43d2c0e..9fc403c7801 100644
--- a/packages/interactivity/src/hooks.tsx
+++ b/packages/interactivity/src/hooks.tsx
@@ -53,6 +53,11 @@ interface DirectiveArgs {
 	 * context.
 	 */
 	evaluate: Evaluate;
+	/**
+	 * Function that dereferences a given path to a value either in the store or the
+	 * context.
+	 */
+	dereference: Dereference;
 }
 
 interface DirectiveCallback {
@@ -71,6 +76,7 @@ interface DirectiveOptions {
 
 interface Scope {
 	evaluate: Evaluate;
+	dereference: Dereference;
 	context: object;
 	ref: RefObject< HTMLElement >;
 	attributes: createElement.JSX.HTMLAttributes;
@@ -80,10 +86,18 @@ interface Evaluate {
 	( entry: DirectiveEntry, ...args: any[] ): any;
 }
 
+interface Dereference {
+	( entry: DirectiveEntry ): any;
+}
+
 interface GetEvaluate {
 	( args: { scope: Scope } ): Evaluate;
 }
 
+interface GetDereference {
+	( args: { scope: Scope } ): Dereference;
+}
+
 type PriorityLevel = string[];
 
 interface GetPriorityLevels {
@@ -307,6 +321,24 @@ export const getEvaluate: GetEvaluate =
 		return hasNegationOperator ? ! result : result;
 	};
 
+// Generate the dereference function.
+export const getDereference: GetDereference =
+	( { scope } ) =>
+	( entry ) => {
+		let { value: path, namespace } = entry;
+		if ( typeof path !== 'string' ) {
+			throw new Error( 'The `value` prop should be a string path' );
+		}
+		// If path starts with !, remove it and save a flag.
+		const hasNegationOperator =
+			path[ 0 ] === '!' && !! ( path = path.slice( 1 ) );
+		setScope( scope );
+		const value = resolve( path, namespace );
+		const result = { value, hasNegationOperator };
+		resetScope();
+		return result;
+	};
+
 // Separate directives by priority. The resulting array contains objects
 // of directives grouped by same priority, and sorted in ascending order.
 const getPriorityLevels: GetPriorityLevels = ( directives ) => {
@@ -338,6 +370,7 @@ const Directives = ( {
 	// element ref, state and props.
 	const scope = useRef< Scope >( {} as Scope ).current;
 	scope.evaluate = useCallback( getEvaluate( { scope } ), [] );
+	scope.dereference = useCallback( getDereference( { scope } ), [] );
 	scope.context = useContext( context );
 	/* eslint-disable react-hooks/rules-of-hooks */
 	scope.ref = previousScope?.ref || useRef( null );
@@ -367,6 +400,7 @@ const Directives = ( {
 		element,
 		context,
 		evaluate: scope.evaluate,
+		dereference: scope.dereference,
 	};
 
 	setScope( scope );
```

</details> 

When dereferencing a function, I'm seeing its source as a closure coming from `handlers` in `store.ts`:

```js
(...args) => {
  (0,_hooks__WEBPACK_IMPORTED_MODULE_1__.setNamespace)(ns);
  try {
    return result(...args);
  } finally {
    (0,_hooks__WEBPACK_IMPORTED_MODULE_1__.resetNamespace)();
  }
}
```

cbravobernal:
We only have this week to land this change before beta 1.

Should we merge this, go with the `data-wp-on-async` directives and work later on the warning @westonruter ?

westonruter:
@cbravobernal That makes sense to me!

gziolo:
Let's remember to document the new API. It should be very similar to:

https://github.com/WordPress/gutenberg/blob/trunk/docs/reference-guides/interactivity-api/api-reference.md#wp-on

It would be also worth explaining the nuances and when to use `wp-on-async` instead of `wp-on`. The way I read it is - always use the async version when there is no `event` consumed.

cbravobernal:
I fixed the changelog conflict.

adamsilverstein:
> Let's remember to document the new API. It should be very similar to:
> 
> [`trunk`/docs/reference-guides/interactivity-api/api-reference.md#wp-on](https://github.com/WordPress/gutenberg/blob/trunk/docs/reference-guides/interactivity-api/api-reference.md?rgh-link-date=2024-05-28T10%3A36%3A06Z#wp-on)
> 
> It would be also worth explaining the nuances and when to use `wp-on-async` instead of `wp-on`. The way I read it is - always use the async version when there is no `event` consumed.

I added the "Needs Dev Note label here - this feature is worth calling out for the 6.6 core release.

westonruter:
I've opened #62160 to implement `data-wp-on-async` for the other core blocks.

cbravobernal:
> Dev note

WordPress 6.6 includes three new directives aimed at improving performance:
 - `data-wp-on-async`
 - `data-wp-on-async-document`
 - `data-wp-on-async-window`
 
These `async` directives optimize event callbacks by first yielding to the main thread. That way, complex interactions won't contribute to long tasks, improving the [Interaction to Next Paint](https://web.dev/articles/inp) (INP). You can read more about it in [Optimize Interaction to Next Paint](https://web.dev/articles/optimize-inp#yield-to-allow-rendering-work-to-occur-sooner).

These directives are recommended over the sync ones (`data-wp-on`, `data-wp-on-document`, and `data-wp-on-window`), but you can use them only when you don't need synchronous access to the `event` object, specifically if you need to call `event.preventDefault()`, `event.stopPropagation()`, or `event.stopImmediatePropagation()`. The directives in core blocks have been updated to use async where available.

When you must resort to using the non-async directives, the `@wordpress/interactivity` package now exports a `splitTask` function which can be used to manually split yield an action to the main thread after calling the synchronous event API. Please see the documentation on [async actions](https://github.com/WordPress/gutenberg/blob/trunk/docs/reference-guides/interactivity-api/api-reference.md#async-actions) for how to implement this.

westonruter:
@cbravobernal Thanks for drafting that dev note! I've made some edits.

westonruter:
I just realized that I didn't export `yieldToMain`. Proposed in https://github.com/WordPress/gutenberg/pull/62665

cbravobernal:
> @cbravobernal Thanks for drafting that dev note! I've made some edits.

I initially put "that way" and GTP4 corrected me to "this way" 😆

westonruter:
@cbravobernal I've updated the dev note comment above to incorporate discussion of the new `splitTask` function and to fix the names of the directives (also updated in #62759).

janusqa:
Hi there. I've been looking at the docs here:
https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/api-reference/

I am using wp-on-window--message to listen for messages sent from an IFRAME to the parent window.
wp-on-window--message seems to work ok.  I get the message, but when i switch over to wp-on-async-window--message no messages arrive.

I've also read this from the PR. Does it mean I'm too early to the party to be using this? The docs indicate 6.5 or higher.
```
WordPress 6.6 includes three new directives aimed at improving performance:

    data-wp-on-async
    data-wp-on-async-document
    data-wp-on-async-window
```

**The particulars:**
```
wordpress 6.5.5
"@wordpress/interactivity": "^6.3.0",
"@wordpress/scripts": "^27.9.0",
```


**The basics**

render.php
```
<div data-wp-on-async-window--message="actions.processPostMessage" id="<?php echo $block_unique_id ?>" class="the-view" data-wp-interactive="create-block" <?php echo wp_interactivity_data_wp_context($block_context) ?>>
<!-- some html goes here -->
</div>
<iframe class="iframe-container" name="theFrame"></iframe>
```

view.js
```
/**
 * WordPress dependencies
 */
import { store, getContext } from '@wordpress/interactivity';
import { longRunningAsyncFunction } from './thefunction';

store('create-block', {
    actions: {
        processPostMessage: function* (event) {
            if (event.origin !== 'http://localhost:9000') return;
            const context = getContext();
            event.data.elementArr.push({ theThing: context.the_thing });
            yield longRunningAsyncFunction(event);
        },
    },
    callbacks: {},
});
```

westonruter:
The async directives are new in 6.6.

Issue #61494, opened 2024-05-08
Global styles: Group "Background image" in "Layout" section


jameskoster:
The global styles inspector is growing:

![new items shadows and background image](https://github.com/WordPress/gutenberg/assets/1204802/85db922e-d2a2-498e-86b6-eb4608a031a9)

Those categories are all related to presets or design tools, except "Background image", which is a tool. So conceptually it doesn't fit at the top level.

A previous version of this issue put it inside a renamed "Layout" section, but per discussion it's not clear that this is actually where we want to be either, given there's a larger revisit we need to do as part of #47369.

To that end, it would be good to — for 6.6 — simply move the Background image tool inside inside Layout. It can even have a subheading, "Background". This would keep the existing, if slightly unclear, IA in place for 6.6, and not preclude a better structure in 6.7.

<details><summary>Previous version of this issue</summary>

<img width="1621" alt="site" src="https://github.com/WordPress/gutenberg/assets/846565/6efed2d5-769f-44ba-bb7a-1419af6e28ae">

As a pathway to a control for a single property, the newly added "Background image" section in global styles is an outlier. 

To improve the iA for 6.6 we could combine it with Layout controls in a new "Site" section. This would also allow the panel title to be consistent across global styles and the block inspector.

</details>

ramonjd:
I'll take a look at this soon, with a view to renaming "Layout" to "Site" - hopefully it'll be a single PR. 👍🏻

jasmussen:
> I'll take a look at this soon, with a view to renaming "Layout" to "Site" - hopefully it'll be a single PR. 👍🏻

@jameskoster are we sure this is the right path forward? Does it fit with the larger IA pieces from #47369? Another path would be to keep the Layout option, holding content and wide widths, as well as block spacing. Then "Site" could be a place for both shadows and background, and any other presets we create until the larger pieces in 47369 come to pass.

Alternately, perhaps the Layout icon works better than the globe? 🤔 

This is mostly a question of the dev time involved. If it's easy to try "Site" in a plugin cycle, let's do that, with the option of reverting to Plan B, which is to just put background image inside the exiting "Layout", and instead make progress on 47369 for 6.7.

jameskoster:
It's really hard to say because the revised iA isn't fully fleshed out. We need to properly define all the concepts and group accordingly. 

[...]

PR #59454, opened 2024-02-29
Global styles: background UI controls

ramonjd:
## Dev Note 📓 

https://github.com/WordPress/gutenberg/pull/59354#issuecomment-2138583222

## What?

Part of:

- https://github.com/WordPress/gutenberg/issues/54336

Follow up to:

- https://github.com/WordPress/gutenberg/pull/59354

Related:

- https://github.com/WordPress/gutenberg/pull/60100

This PR add background image UI controls to top-level global styles. 

### TODO

- Maybe E2E tests...

## Why?

Background styles can now be added to theme.json since https://github.com/WordPress/gutenberg/pull/59354.

The styles are currently only " top-level", which means they're added to `body` selector. 

This PR allows users to edit these styles in the site editor.

## How?

Similar to dimensions and typography panels, this PR refactors the background block supports component so that it can be shared between the site and post editors.

It moves all relevant tests and styles to the `/global-styles` directory.

## Testing Instructions

This is a big one, so get comfortable.

Firstly, check out this branch.

At this stage it's probably good to check if there are any regressions to background block supports. So, add a group block to a post, and make sure you can:

1. Add/remove/upload/reset a new background image (See https://github.com/WordPress/gutenberg/pull/53934 for test instructions)
2. Edit/reset the background position values (https://github.com/WordPress/gutenberg/pull/58592)
3. Edit/reset background size and repeat values (https://github.com/WordPress/gutenberg/pull/57005)

If that's all good, head over to the site editor.

I've been using empty theme to test. Here is some test theme.json, with a relative path to the updated theme.json schema:

```json
{
	"$schema": "../../schemas/json/theme.json",
	"version": 2,
	"settings": {
		"appearanceTools": true,
		"background": {
			"backgroundImage": true,
			"backgroundSize": true
		}
	},
	"styles": {
		"color": {
			"background": "tomato"
		},
		"background": {
			"backgroundImage": {
				"url": "https://i0.wp.com/wordpress.org/files/2023/12/sotw-dotorg-drawer.png?w=1807&ssl=1",
				"source": "file"
			},
			"backgroundPosition": "30% 30%",
			"backgroundRepeat": "repeat",
			"backgroundSize": "cover"
		}
	}
}
```

Ensure that:

1. Any inherited values from theme.json are present.
2. You can interact with the background controls in a similar way to above, e.g., at the block level, and
3. Saving the styles affects the body styles of your site. 
[...]

PR #61886, opened 2024-05-23
Global styles: relocate Background Image controls to sit under Layout

ramonjd:
## What?

This PR relocates the Background Image controls to sit under Layout in the global styles sidebar.

It also removes the Background image top-level navigation item.

Background image controls were originally added to global styles in https://github.com/WordPress/gutenberg/pull/59454

There are no modifications to functionality.

## Why?

The global styles taxonomy is under review and will likely change between now and WordPress 6.7.

See discussion:

- https://github.com/WordPress/gutenberg/issues/61494#issuecomment-2116865577

The ambition in this PR is to make a low-impact change to the global styles taxonomy until these changes are implemented.

## How?
Just movin' stuff around. 

## Testing Instructions

The main concern here is whether the location of Background controls, despite it being temporary, is a reasonable step and won't introduce too much confusion.

Furthermore, as no functionality is being introduced, it would be great to smoke test that the background controls work as they should, e.g., 

- you should be able to add and edit a site-wide background image by heading to **Global styles > Layout > Background**
- you should still be able to set a background image in your theme.json and have it appear in the controls

## Screenshots or screencast 

Screenshot 2024-05-23 at 10 36 48 am

Screenshot 2024-05-23 at 10 37 30 am

### DIFF
--- a/packages/edit-site/src/components/global-styles/background-panel.js
+++ b/packages/edit-site/src/components/global-styles/background-panel.js
@@ -59,7 +59,7 @@ export default function BackgroundPanel() {
 			value={ style }
 			onChange={ setStyle }
 			settings={ settings }
-			headerLabel={ __( 'Image' ) }
+			headerLabel={ __( 'Background' ) }
 			defaultValues={ BACKGROUND_DEFAULT_VALUES }
 			defaultControls={ defaultControls }
 			themeFileURIs={ _links?.[ 'wp:theme-file' ] }
--- a/packages/edit-site/src/components/global-styles/root-menu.js
+++ b/packages/edit-site/src/components/global-styles/root-menu.js
@@ -6,7 +6,6 @@ import {
 	typography,
 	color,
 	layout,
-	image,
 	shadow as shadowIcon,
 } from '@wordpress/icons';
 import { __ } from '@wordpress/i18n';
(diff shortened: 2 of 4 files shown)


jasmussen (approved):
This feels like a good interim state for the feature. It allows it to ship, be iterated, and it's grouped with styles that are perhaps also not ideally placed, but nevertheless are related in a "global" context.

background inside styles

Notably it doesn't preclude future improvements and a clearer hierarchy, as we've discussed in other tickets.

youknowriad:
Hi Ramon, just noting that the commit titles appear on #core slack, so we should try to make the commit titles a bit more comprehensible as much as we can when merging PRs. Thank you.

ramonjd:
> Hi Ramon, just noting that the commit titles appear on #core slack, so we should try to make the commit titles a bit more comprehensible as much as we can when merging PRs. Thank you.

Thanks for the ping @youknowriad 

This appears to be a limitation with the slack bot. I see that it only [picked up the first line](https://wordpress.slack.com/archives/C02RQBWTW/p1716522139828739), but my commit message was sufficiently comprehensive in my opinion.

> Global styles: 
> 
> - remove Background image top-level navigation item
> - relocate background panel to sit next to Layout controls


So that means we need to use one liners in our commit messages to appease the core slack bot 😆

ramonjd:
Oh, the slack bot picks up the PR title. 

The title was always `Global styles: relocate Background Image controls to sit under Layout` so it seems to be an error with the slack bot in this case? Maybe the semi colon 🤷🏻 

In any case I've added a comment in a thread in #core to elaborate.

youknowriad:
Definitely a bug in slack bot. Sorry for not noticing :)

PR #61889, opened 2024-05-23
Writing flow: remove first empty paragraph on Backspace

ellatrix:
## What?

Allows you to remove the first paragraph in the editor (or post content) with Backspace.

## Why?

Currently it's impossible to remove the first paragraph with Backspace, you'd have to click the delete button in the options menu.

## How?

Extends the block merging behaviour.

## Testing Instructions

See e2e test.

* Create a post in the post editor, add an empty paragraph and a second non empty paragraph, Backspace from the first. The first paragraph should be deleted, selection should move to the start of the second paragraph.
* Do the same thing but in the post content block in the site editor (create a new page, or use an existing).

### DIFF
--- a/packages/block-editor/src/components/block-list/block.js
+++ b/packages/block-editor/src/components/block-list/block.js
@@ -306,6 +306,28 @@ const applyWithDispatch = withDispatch( ( dispatch, ownProps, registry ) => {
 				canInsertBlockType,
 			} = registry.select( blockEditorStore );
 
+			function switchToDefaultOrRemove() {
+				const block = getBlock( clientId );
+				const defaultBlockName = getDefaultBlockName();
+				if ( getBlockName( clientId ) !== defaultBlockName ) {
+					const replacement = switchToBlockType(
+						block,
+						defaultBlockName
+					);
+					if ( replacement && replacement.length ) {
+						replaceBlocks( clientId, replacement );
+					}
+				} else if ( isUnmodifiedDefaultBlock( block ) ) {
+					const nextBlockClientId = getNextBlockClientId( clientId );
+					if ( nextBlockClientId ) {
+						registry.batch( () => {
+							removeBlock( clientId );
+							selectBlock( nextBlockClientId );
+						} );
+					}
+				}
+			}
+
 			/**
 			 * Moves the block with clientId up one level. If the block type
 			 * cannot be inserted at the new location, it will be attempted to
@@ -345,7 +367,16 @@ const applyWithDispatch = withDispatch( ( dispatch, ownProps, registry ) => {
 								getDefaultBlockName()
 							);
 
-							if ( replacement && replacement.length ) {
+							if (
+								replacement &&
+								replacement.length &&
+								replacement.every( ( block ) =>
+									canInsertBlockType(
+										block.name,
+										targetRootClientId
+									)
+								)
+							) {
 								insertBlocks(
 									replacement,
 									getBlockIndex( _clientId ),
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/components/block-list/block.js
```diff
@@ -344,14 +367,25 @@ const applyWithDispatch = withDispatch( ( dispatch, ownProps, registry ) => {
...
+							) {
 								insertBlocks(
 									replacement,
 									getBlockIndex( _clientId ),
 									targetRootClientId,
 									changeSelection
 								);
 								removeBlock( firstClientId, false );
+							} else {
+								switchToDefaultOrRemove();
```
ellatrix, line 388: I'll add a test for the site editor as well.

andrewserong (approved):
Thanks for the ping! This is an interesting one.

When it comes to backspace, my expectation was that it moves the caret up a line, and when we're in the first position within the list of blocks, backspace doing nothing feels a bit more correct to me (it works that way in Google Docs for example). In order to remove the first line, I'd reach for the delete key rather than backspace, as the action we're performing applies to what follows from the current caret position rather than what precedes.

I notice that in both trunk and this PR, if I press delete from the first line, there's a focus loss, so the focus is no longer within the editor canvas:

https://github.com/WordPress/gutenberg/assets/14988353/e361416e-8232-4eae-b49c-f2f29b50c358

Would it be worth fixing that up in this PR, or in a separate one? I imagine it's a case of using the `selectBlock( nextBlockClientId );` logic again somewhere?

In this PR, when using the backspace key, the focus is correctly preserved within the editor canvas:

https://github.com/WordPress/gutenberg/assets/14988353/91c9f3c3-bea5-44ea-8c41-63686d6c99e0

---

So, from my perspective I have a very weakly held opinion that I slightly prefer the behaviour on trunk. Since this PR exists, though, I imagine this is a feature folks have requested, or a pain point that's been encountered?

The code change looks good to me — I'm not terribly familiar with this part of the code, but it makes sense to me. It looks like the `else` conditions are slightly more permissive than beforehand, but from the looks of it that's intentional.

As we're close to the beta freeze, I think if this is a desired feature it's worth trying and the added test coverage gives me confidence, along with the fact that it's testing as expected in the post and site editors, so I'll give this the ✅.

It might be worth getting a second review, though, if there's any doubt about it!

t-hamano (approved):
I tested it on Windows OS and it works as expected.

> I notice that in both trunk and this PR, if I press delete from the first line, there's a focus loss, so the focus is no longer within the editor canvas:

I was able to reproduce this issue too. It might not be necessary to address it in this PR, but perhaps we could make the following changes, for example?

```diff
diff --git a/packages/block-editor/src/components/block-list/block.js b/packages/block-editor/src/components/block-list/block.js
index 030d08b42c..5e191af76b 100644
--- a/packages/block-editor/src/components/block-list/block.js
+++ b/packages/block-editor/src/components/block-list/block.js
@@ -453,7 +453,11 @@ const applyWithDispatch = withDispatch( ( dispatch, ownProps, registry ) => {
                                if ( getBlockOrder( nextBlockClientId ).length ) {
                                        moveFirstItemUp( nextBlockClientId, false );
                                } else {
+                                       const block = getBlock( clientId );
                                        mergeBlocks( clientId, nextBlockClientId );
+                                       if ( isUnmodifiedDefaultBlock( block ) ) {
+                                               selectBlock( nextBlockClientId );
+                                       }
                                }
                        } else {
```

ellatrix:
@andrewserong Since it's in trunk I can do a follow-up. Let's merge this so we have it in the release.

PR #61891, opened 2024-05-23
Writing flow: split heading into default block

ellatrix:
## What?

|Before|After|
|-|-|
|split-before|split-after|

## Why?

* Headings can have an anchor that shouldn't be duplicated.
* It never makes sense to split an h1 in two, there should be no two h1 tags.
* Most likely other headings should not be duplicated either.

One can always switch it back to a heading.

### DIFF
--- a/packages/block-editor/src/store/actions.js
+++ b/packages/block-editor/src/store/actions.js
@@ -1002,6 +1002,9 @@ export const __unstableSplitSelection =
 		// the default block type cannot be inserted.
 		const defaultBlockName = getDefaultBlockName();
 		if (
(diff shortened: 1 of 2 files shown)


ellatrix:
Just need to adjust the test expectations still

erikiva (approved):
Tested and it looks good and behaves as expected. Thank you!

ntsekouras (approved):
Looks good, thanks!

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: c17a892bc2

PR #61892, opened 2024-05-23
Editor: Use edited entity for post actions

### LINKED ISSUE #61292: Site Editor > Pages: renaming shows old title in Inspector if you don't save
While testing GB 18.2 and some of the newer efforts, I noticed that if I rename the page via the page title, don't hit save, and then select the option to rename the page from the inspector, it shows the old title:

https://github.com/WordPress/gutenberg/assets/26996883/c99380d6-5bb7-402f-8d97-4b82ebd0721c

To replicate:

1. Create a new page in the site editor.
2. Edit the page title block and add a new title.
3. Don't save.
4. Select the three dot menu > rename in the Inspector.
5. Notice the old name is in place. 

cc @ntsekouras since you're working on this: https://github.com/WordPress/gutenberg/issues/59689

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/61292

it's better to use the edited entity, because many actions might be better to use the current info, even if not saved. Examples would be the rename Post action, or the duplicate post action.

## Testing Instructions
1. Go to a page
3. Edit the page title block and add a new title.
4. Don't save.
5. Select the three dot menu > rename in the Inspector.
6. Observe the edited name is rendered in the rename control

### DIFF
--- a/packages/editor/src/components/post-actions/index.js
+++ b/packages/editor/src/components/post-actions/index.js
@@ -10,6 +10,7 @@ import {
 	Modal,
 } from '@wordpress/components';
 import { moreVertical } from '@wordpress/icons';
+import { store as coreStore } from '@wordpress/core-data';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


youknowriad (approved):
LGTM

PR #61893, opened 2024-05-23
Editor: Fix canvas padding in post editor

youknowriad:
## What?

I noticed a small regression in the post editor where the width of the canvas is not correct (no padding on the right) when you move to "template mode". This PR fixes the issue but I'm not entirely sure when this regression happened.

Screenshot 2024-05-23 at 9 32 09 AM

## Testing Instructions

1- Open the post editor
2- Click "Template" on the sidebar and "Edit template"
3- There should be a padding around the canvas.


ellatrix:
Could this be related to zoom-out changes? Maybe that's at least a good thing to retest, cc @MaggieCabrera

ellatrix (approved):

PR #61753, opened 2024-05-17
Block Bindings: Allow editing in post meta source

SantosGuillamot:
## What?
Use the latest changes to the block bindings API in the editor to allow users to edit the value of the custom fields directly through the blocks when they are connected. For example, when the paragraph content is bound to a specific custom field, and the user start typing, they would be modifying the custom field and not the content.

At least in this first iteration, it should be limited to:
* Users with the permissions to edit the field.
* When users are editing a post, and not a template.
* Custom fields should not be protected and available through the REST API.

## How?
Using the `setValue` function based on the code created by @ellatrix  [here](https://github.com/WordPress/gutenberg/pull/60725). Apart from that, we should use the `lockAttributesEditing` callback added in this pull request to cover the mentioned limitation.

Additionally, it ensure that the split selection work as expected in bound blocks.

## Testing Instructions
I added a couple of e2e tests for this use case. You can test it manually by:

## Testing Instructions

<details>
<summary>1. Register post meta by adding this snippet to your theme's functions.php</summary>

```php
add_action( 'init', 'test_block_bindings' );

function test_block_bindings() {
	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'type'              => 'string',
			'default'           => 'default text value',
		)
	);
}
```

</details>

<details>

<summary>2. Add a paragraph block bound to the custom field using the Code Editor</summary>

```
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_field"}}}}} -->
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Check that the value of the custom field is shown.
4. Check you can edit it when you are in the post editor.
5. Check that modifying the value actually changes the custom field and not the paragraph content.

PR #60826, opened 2024-04-17
Block Bindings: rely on broader context instead of requiring direct block context

ellatrix:
## What?

This PR removes the need for bound block to have the context that a binding needs. Instead we can look at the broader context. I've opted to require a binding source to declare the context it needs, just to avoid passing all context down to the callbacks (not absolutely required though). This is very similar to how a block needs to declare what context it needs.

## Why?

Block bindings shouldn't have to affect the block registration, instead we can add the right context just in time.

I could see Bits work similarly: it can declare what context it needs without requiring the parent block to consume the context as well.

PR #61895, opened 2024-05-23
Try: Sync block context with store

SantosGuillamot:
## What?
The idea behind this pull request is to sync the block context in the store.

## Why?
To be able to access it without using a hook for uses cases like [bindings modifying the block attributes](https://github.com/WordPress/gutenberg/pull/61817).

## How?
Creating a new action `updateBlockContext` to modify the block context in the store and a selector `getBlockContext` to retrieve it.

Apart from that, I'm syncing the context in the edit component using `useContext`.

## Testing Instructions
* In any block, use the `getBlockContext` selector and check that the block context is present.

### DIFF
--- a/packages/block-editor/src/components/block-context/index.js
+++ b/packages/block-editor/src/components/block-context/index.js
@@ -2,7 +2,12 @@
  * WordPress dependencies
  */
 import { useDispatch } from '@wordpress/data';
-import { createContext, useContext, useMemo } from '@wordpress/element';
+import {
+	createContext,
+	useContext,
+	useLayoutEffect,
+	useMemo,
+} from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -35,7 +40,12 @@ export function BlockContextProvider( { value, children } ) {
 		() => ( { ...context, ...value } ),
 		[ context, value ]
 	);
-	useDispatch( blockEditorStore ).updateBlockContext( nextValue );
+
+	// Synced the block context in the store.
+	const { updateBlockContext } = useDispatch( blockEditorStore );
+	useLayoutEffect( () => {
+		updateBlockContext( nextValue );
+	}, [ nextValue ] );
 
 	return <Context.Provider value={ nextValue } children={ children } />;
 }
--- a/packages/block-editor/src/components/block-context/index.js
+++ b/packages/block-editor/src/components/block-context/index.js
@@ -1,7 +1,13 @@
 /**
  * WordPress dependencies
  */
+import { useDispatch } from '@wordpress/data';
 import { createContext, useContext, useMemo } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { store as blockEditorStore } from '../../store';
 
 /** @typedef {import('react').ReactNode} ReactNode */
 
@@ -29,6 +35,7 @@ export function BlockContextProvider( { value, children } ) {
 		() => ( { ...context, ...value } ),
 		[ context, value ]
 	);
+	useDispatch( blockEditorStore ).updateBlockContext( nextValue );
 
 	return <Context.Provider value={ nextValue } children={ children } />;
 }
--- a/packages/block-editor/src/components/block-context/index.js
+++ b/packages/block-editor/src/components/block-context/index.js
@@ -1,18 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { useDispatch } from '@wordpress/data';
-import {
-	createContext,
-	useContext,
-	useLayoutEffect,
-	useMemo,
-} from '@wordpress/element';
-
-/**
- * Internal dependencies
- */
-import { store as blockEditorStore } from '../../store';
+import { createContext, useContext, useMemo } from '@wordpress/element';
 
 /** @typedef {import('react').ReactNode} ReactNode */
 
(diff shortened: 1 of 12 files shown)


packages/block-editor/src/components/block-edit/edit.js
```diff
@@ -48,9 +50,14 @@ const Edit = ( props ) => {
...
 const EditWithGeneratedProps = ( props ) => {
-	const { attributes = {}, name } = props;
+	const { clientId, attributes = {}, name } = props;
 	const blockType = getBlockType( name );
 	const blockContext = useContext( BlockContext );
+	// Sync the block context with the block editor store.
+	useDispatch( blockEditorStore ).updateBlockContext(
+		clientId,
+		blockContext
+	);
```
ellatrix: We can't call actions on render, must be used to an effect, see https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/inner-blocks/use-nested-settings-update.js

ellatrix: Also, we shouldn't do this for every block, we should only do if where context it provided

ellatrix: Actually, looking at https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/inner-blocks/use-block-context.js, it appears block context is just a selector? So we could derive the whole thing the getBlockAttributes selector?

ellatrix: So in `getBlockAttributes`, we should loop up the parents, see if they provide context, and if so set their attributes as context?

SantosGuillamot: Good points 🙂 

I thought we needed to get the value from `useContext( BlockContext )`, as you were doing [here](https://github.com/WordPress/gutenberg/pull/60826/files#diff-99d79bdba99fe4517413b0fd9503eef924e439c03d44504c1ad0606b727cebb6R62), but maybe you are right and we can do that looping up the parents. I can give that a try.

SantosGuillamot: That wouldn't include context provided by `BlockContextProvider` that doesn't belong to a block, right?

I'm thinking that, if I am not mistaken, we are providing `postId` and `postType` as context in the post template: [link](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/post-template/edit.js#L212-L220). But that wouldn't be available iterating through the parents, right?

ellatrix: Hm, I guess then it is easier to sync the context. I'd sync it inside of `BlockContextProvider`. Then create a selector that gets the context which would loop through the parents and finds/merges the context.

SantosGuillamot: Do you mean something like [this try](https://github.com/WordPress/gutenberg/pull/61895/commits/c93906123cedcd6aaed84d443b3eb0f60af308cb)?

packages/block-editor/src/components/block-context/index.js
```diff
@@ -29,6 +35,7 @@ export function BlockContextProvider( { value, children } ) {
 		() => ( { ...context, ...value } ),
 		[ context, value ]
 	);
+	useDispatch( blockEditorStore ).updateBlockContext( nextValue );
```
ellatrix: Must be done in an effect

SantosGuillamot:
I've been taking a deeper look at this and I am not sure how the context syncing should work. Let me use the query loop & post template example to explain my concerns.

When we have a query loop with a post template, the post template provides the `postId` and `postType` using `BlockContextProvider`: [link](https://github.com/WordPress/gutenberg/blob/93d45e807519a2e5eda92744a706d175e51afd49/packages/block-library/src/post-template/edit.js#L278-L302). If I understand it correctly, `BlockContextProvider` adds or overwrite the context **ONLY** for the children. This means that a block will have a different context depending on the post template it is inside.

If I am not mistaken, the `useContext` hook takes care of knowing which context belongs where, but I am not sure how to make that differentiation in the store. In the current implementation of this pull request, we have a single store, but that isn't right.

Maybe we should rely on `useContext` and divide the store by clientId. But that probably means not doing the syncing in the `BlockContextProvider`, right?

EDIT: I went back to the previous implementation of syncing the store in the edit component and it seems to work better.

SantosGuillamot:
I've been testing this pull request, plus one to [move the bindings logic](https://github.com/WordPress/gutenberg/pull/61817) and to [allow post meta editing](https://github.com/WordPress/gutenberg/pull/61753), and everything seems to work as expected.

I'm marking this pull request as ready for review, but there are some outstanding questions:
* Is it okay to expose the context in the store?
* If so, is this the proper way?
* Should we make the action and selector private?

SantosGuillamot:
As it seems it is still not clear if we should expose the block context in the store, I've moved the action and selector to private because it seems safer to me.

gziolo:
It looks like we won't need it after all. Based on the progress on other tasks planned for WordPress 6.7 outlined in https://github.com/WordPress/gutenberg/issues/63018, I suggest we close this PR.

SantosGuillamot:
Yes, let's close this one and revisit it if considered necessary in the future.

PR #61896, opened 2024-05-23
Editor: Move the resizing of the editor to the EditorCanvas component

youknowriad:
Related #52632 

## What?

In the site editor, some entities offer a way to resize the canvas (like template parts and patterns) but in the post editor it's not the case, this PR aligns this behavior by moving the resizing feature to the "editor" package sharing it between both post and site editors.

This involves moving a lot of stuff around so there's a small potential for regressions. Please help test the different entities.

## Testing Instructions

- Try opening and navigating templates, template parts, patterns, navigation in site and post editor and ensure the canvas size is correct and that you can resize the canvas properly for template parts, patterns and navigation.


jameskoster:
It works, but there's something off with the height and spacing.

### Site Editor (correct)
Screenshot 2024-05-23 at 12 11 54

### Post editor
Screenshot 2024-05-23 at 12 12 05

The canvas is too tall, and the padding is incorrect.

youknowriad:
I think the padding is probably unrelated to this PR but yeah I'm not entirely sure why the height is not the same. I'll try to debug.

youknowriad:
@jameskoster Would you mind tracking the padding bug separately. I think it exists in trunk already and would be good to add to the 6.6 list.

jameskoster:
Okay.

I'm still seeing the height issue though, in fact now it's a bit worse. Seems the canvas fills the available space however large it is:

Screenshot 2024-05-23 at 13 01 35

youknowriad:
@jameskoster 🤔 I can't reproduce, would you mind doing another test?

jameskoster:
Still not working for me. I think it's the inline height style here:

Screenshot 2024-05-23 at 13 15 44

For some reason it's calculated much too large. 

Curiously, if I toggle it off in the browser inspector then it gets immediately recalculated and re-added with the correct value 🤔 

Screenshot 2024-05-23 at 13 18 26

youknowriad:
Which theme are you using?

jameskoster:
Twenty Twenty-Four

youknowriad:
Since this is just the post editor and that I'm not able to reproduce. We have two options:

 - Don't consider this a blocker (not sure what the post editor does for you in trunk)
 - Help me figure out why I can't reproduce :P

youknowriad:
@ntsekouras or @ellatrix Are you able to reproduce the issue @jameskoster raised above: Opening a template part in focus mode in the post editor makes the height 100% rather than picking the actual height of the template part.

jameskoster:
Okay I figured it out. Well, partially. I don't know why, but when I deactivate Jetpack it works as expected.

jameskoster (approved):

PR #39286, opened 2022-03-08
Add focus mode for Navigation Menus

getdave:
## What

Adds a focus mode to Navigations. Can be accessed via Browse Mode or directly via URL.

Also refactors the Site Editor code to enable this.

Part of https://github.com/WordPress/gutenberg/issues/50396.

## Why

In https://github.com/WordPress/gutenberg/issues/37950 we learnt that a "focus mode" for editing Navigation Menus in isolation could be useful.

This PR implements that experience allowing you to edit a menu's _items_ in the block editor canvas. 

Note that because a Navigation Menu can be assigned to multiple _blocks_, this experience must necessarily be presentationally agonistic (i.e. it's intended to focus more on manipulating the menu _items_ rather than the specific Navigation block instance).

## How

Extends the existing infrastructure around Template Parts to also handle Navigation Menu entities. This requires some interesting "hacks" such as using a "ghost" Navigation block to provide the editing context. I anticipate these will be removed in future in favour of a more unified experience but this will be a wider refactor and much more invovled.

Closes https://github.com/WordPress/gutenberg/issues/37950.

## Testing Instructions

- Create a new Navigation Menus (with some items) using the Nav _block_.
- Load the Site Editor and open Browse Mode.
- Click `Navigation` and then select one of your menus.
- See the editor window change to a isolated version of the block.
- Click on the editor to enter focus mode. 
- See that the Nav block is there but it's locked down and doesn't show any controls.
- Verify you can still access list view editing mode on the Navigation block.
- Make some changes to the menu _items_ and save.
- Open Browse Mode again and see changes reflected there.

Try this with some Themes with Global Styles which effect the styling of the Navigation. You should see the styles inherited in the focus mode.

## Todo

- [x] Refactor around bias toward templates and template parts.
- [x] Add spacing/padding around Navigation block when not in full "focus" mode (i.e. when browsing via Navigation in Browse Mode).
- [x] Restrict block editor to only allow the current root block (i.e. the Navigation block) when in focus mode for the `wp_navigation` entity.
- [x] Remove controls coming from block supports on the Navigation block. Note this will require adding a hook which adds an "Alignment" control to the Nav block just for focus mode.
- [ ] Consider adding `focus` as one of [the editor modes](https://github.com/WordPress/gutenberg/blob/04f09823091360b4251d89af477dd5498169483f/packages/block-editor/src/store/reducer.js#L1694). We can then toggle this when entering focus mode and allow us to conditionalise various things about the UI based on whether it is active.

## Screenshots 

[...]

PR #61897, opened 2024-05-23
Navigation Focus Mode: Remove leftover code

youknowriad:
## What?

When trying to align post and site editors, I noticed that the Navigation block focus mode has a lot of specific custom code to make it work as it should. But after digging a little bit more, I found that most of that code is actually useless and already handled elsewhere. This PR just cleans these things a little bit:

 - No specific className to add padding (this is already handled by the editor package)
 - No specific code to disable the appender (it was already disabled at the root level for me)

This code was first introduced in the initial PR that introduced Navigation focus mode #39286 

## Testing Instructions

1- Open the site editor
2- Click navigation
3- Click the "edit" link in the dropdown in the sidebar
4- You should be able to edit the navigation menu properly, a padding is visible around the canvas and there's no root level appender.

### DIFF
--- a/packages/edit-site/src/components/block-editor/editor-canvas.js
+++ b/packages/edit-site/src/components/block-editor/editor-canvas.js
@@ -6,7 +6,6 @@ import clsx from 'clsx';
 /**
  * WordPress dependencies
  */
-import { store as blockEditorStore } from '@wordpress/block-editor';
 import { useSelect, useDispatch } from '@wordpress/data';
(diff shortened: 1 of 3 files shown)


draganescu (approved):
I tested the "navigation editor" and with all this code removed the experience is 100% the same as it is on `trunk`:

- no appender in both edit and view outside the nav block
- not sure what the padding was doing b/c there is no padding on trunk either.

PR #61280, opened 2024-05-01
Fix: Remove color styles from typography variation controls - #61217

hbhalodia:
## What?

- Issue - https://github.com/WordPress/gutenberg/issues/61217

## Why?

- This would remove the global style colors being added to the typography variation in the style's editor page.

## How?

- This PR adds the extra props `addGlobalStyles`, if this is passed as false, it would not add the style to the iframe preview.
- This is necessary because, the same component reused at multiple instance, hence in order to not make effect to other place, such as the style variations we need to pass it as a prop and do action based on the prop provided.

## Testing Instructions

1. Checkout to this branch.
2. Run the NPM run build process.
3. Open the site → editor → styles → typography.
4. You would see that, the typography options have now natural black and white color combination, and it would be the same for each of the style variations.

### Testing Instructions for Keyboard

- NIL

## Screenshots or screencast 

[GB Issue- 61217.webm](https://github.com/WordPress/gutenberg/assets/58802366/ce8b8ea3-a8d7-456b-823c-f2c942c27e6c)

PR #61898, opened 2024-05-23
Global Styles: Don't apply the background and text colors to typography previews

### LINKED ISSUE #61217: Remove color styles from typography variation controls
Currently typography variations have the active theme’s styles applied to the controls.

Let’s instead not apply those colors, but only the typography styles—making these controls feel more integrated with the rest of the UI, potentially reducing confusion. These don't need the colors to communicate the typography values, and may be confused with full style variations (which do include color and typography styles).   

### Visual 

CleanShot 2024-04-29 at 15 39 33
CleanShot 2024-04-29 at 15 41 58

scruffian:
## What?
This overrides the site text and background colors on typography previews. Fixes #61217

## Why?
It's easier to preview the typography without the additional color information.

## How?
Pass styles to the Iframe Preview component.

## Testing Instructions
1. Open Global Styles
2. Open Typography and check that the background color is white and text color is black
3. Also check this on the site view.

## Screenshots or screencast 
Screenshot 2024-05-23 at 13 04 29
Screenshot 2024-05-23 at 13 04 17

### DIFF
--- a/packages/edit-site/src/components/global-styles/preview-iframe.js
+++ b/packages/edit-site/src/components/global-styles/preview-iframe.js
@@ -38,6 +38,7 @@ export default function PreviewIframe( {
 	label,
 	isFocused,
 	withHoverView,
+	previewStyles,
 } ) {
 	const [ backgroundColor = 'white' ] = useGlobalStyle( 'color.background' );
 	const [ gradientValue ] = useGlobalStyle( 'color.gradient' );
@@ -109,6 +110,15 @@ export default function PreviewIframe( {
 	}, [ styles ] );
 	const isReady = !! width;
 
+	const calculatedStyles = {
+		height: normalizedHeight * ratio,
+		width: '100%',
+		background: gradientValue ?? backgroundColor,
+		cursor: withHoverView ? 'pointer' : undefined,
+		// Override these styles with the previewStyles prop.
+		...previewStyles,
+	};
+
 	return (
 		<>
 			<div style={ { position: 'relative' } }>
@@ -126,12 +136,7 @@ export default function PreviewIframe( {
 				>
 					<EditorStyles styles={ editorStyles } />
 					<motion.div
-						style={ {
-							height: normalizedHeight * ratio,
-							width: '100%',
-							background: gradientValue ?? backgroundColor,
-							cursor: withHoverView ? 'pointer' : undefined,
-						} }
+						style={ calculatedStyles }
 						initial="start"
 						animate={
 							( isHovered || isFocused ) &&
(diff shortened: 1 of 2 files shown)


draganescu (commented):
This tests out very well!

packages/edit-site/src/components/global-styles/preview-iframe.js
```diff
@@ -109,6 +110,15 @@ export default function PreviewIframe( {
 	}, [ styles ] );
 	const isReady = !! width;
 
+	const calculatedStyles = {
```
draganescu: Since this is not a calculation of this vector but a computation of its values (we have if conditions hence steps) maybe:
```suggestion
	const computedStyles = {
```

packages/edit-site/src/components/global-styles/preview-iframe.js
```diff
@@ -38,6 +38,7 @@ export default function PreviewIframe( {
 	label,
 	isFocused,
 	withHoverView,
+	previewStyles,
```
draganescu, line 41: At first I thought it was odd that a component named `PreviewIframe` didn't already receive `previewStyles` before this PR. Then I noticed that this is used to imperatively set CSS on the iframe. I am not a fan of this, although it works.

Why not tell the component what we're previewing and let it internally decide how to style itself? This way we also avoid magic values coming from other components.

scruffian: That's the approach in https://github.com/WordPress/gutenberg/pull/61280

ajlende: I agree with @draganescu, and I don't like the boolean `addGlobalStyles` prop from #61280.

We have a [`useGlobalStylesOutputWithConfig`](https://github.com/WordPress/gutenberg/blob/851e2a051b5c7a14669b83b83c1f9bd4b1e0bd86/packages/block-editor/src/components/global-styles/use-global-styles-output.js#L1311) function that allows you to pass exactly what is needed for styles. Or you could move the change up to the [`Variation`](https://github.com/WordPress/gutenberg/blob/962bd063bd6c75662b5facf5bc3408549816af93/packages/edit-site/src/components/global-styles/variations/variation.js#L25) component which provides the [`GlobalStylesContext`](https://github.com/WordPress/gutenberg/blob/962bd063bd6c75662b5facf5bc3408549816af93/packages/edit-site/src/components/global-styles/variations/variation.js#L28-L40).

It seems like a simpler solution to just change the source of the styles rather than overriding them with a new prop at this level.

scruffian:
Closing in favour of https://github.com/WordPress/gutenberg/pull/62578

PR #54543, opened 2023-09-18
Writing Flow/Rich Text: unify split logic

ellatrix:
## What/why?

Fixes https://github.com/WordPress/gutenberg/issues/53076: when pasting blocks, the split block should be merged with the pasted blocks.

This PR aims to unify split logic (enter key and paste). Currently we have logic _both_ in the block editor generally and rich text specifically when it contains selection. This PR moves all the logic to the block editor (writing flow), adding a block.json setting that tells us if we can split a block.

This removes the need for the block to pass an `onSplit` function, the behaviour can be turned on with just a setting. This is also quite convenient for extenders wishing to recreate the core blocks.

This will also allow us to solve splitting for paste more easily, which is actually a split operation + merging the ends together.

## How?

One implementation details is that change of how rich text listens for key events. Right now in trunk, the event listener is attached to the rich text node, so writing flow (listening on a wrapper node) cannot prevent default behaviour. To fix this, rich text can listen to the document instead. This is how it works for browser behaviour too. Default behaviour is cancelable until the event bubbled all the way through the DOM. Rich text is meant to be a sort of polyfill for the native `contenteditable` API: actions should be cancelable.

## Testing Instructions

Copy a few paragraphs and paste them into another paragraph. The result should be:

```
[target paragraph][pasted paragraph

pasted paragraph

pasted paragraph][target paragraph]
```

## Screenshots or screencast 

paste-between

PR #61900, opened 2024-05-23
Writing flow: fix heading crash on split (via paste)

ellatrix:
## What?

Currently the two resulting heading blocks crash because the `level` attribute is not defined. This is because the `merge` function does not merge any attributes other than `content`.

## Why?

Regression introduced in #54543.

## How?

We should use the current attributes as a fallback.

## Testing Instructions

Create a heading and paste two paragraphs in the middle.


SantosGuillamot (approved):
I've tested it and it seems to solve the issue 🙂

PR #61901, opened 2024-05-23
Global Styles: Restore the default variation to the color and typography style tiles

scruffian:
## What?
This restores the default variation to the style tiles for color and typography sets.

## Why?
This is necessary to allow users to put their color and typography settings back to how they were before they made changes.

## How?
Don't filter out Default as an empty variation - treat it as an exception.

## Testing Instructions
1. Open the Site Editor
2. Open Global Styles
3. Go to Typography
4. Confirm that the default styles are first in the list
3. Go to Colors
4. Confirm that the default styles are first in the list
5. Also confirm at the default styles are visible on the site view

## Screenshots or screencast 
Screenshot 2024-05-23 at 13 28 27
Screenshot 2024-05-23 at 13 28 12
Screenshot 2024-05-23 at 13 27 57

### DIFF
--- a/packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
+++ b/packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
@@ -160,8 +160,12 @@ export default function useThemeStyleVariationsByProperty( {
 					cloneDeep( variation ),
 					property
 				);
+
 				// Remove variations that are empty once the property is filtered out.
-				if ( Object.keys( variationFilteredByProperty ).length === 0 ) {
+				if (
+					variation.title !== 'Default' &&
+					Object.keys( variationFilteredByProperty ).length === 0
+				) {
 					return accumulator;
 				}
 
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -160,8 +160,12 @@ export default function useThemeStyleVariationsByProperty( {
 					cloneDeep( variation ),
 					property
 				);
+
 				// Remove variations that are empty once the property is filtered out.
-				if ( Object.keys( variationFilteredByProperty ).length === 0 ) {
+				if (
+					variation.title !== 'Default' &&
```
youknowriad: If I'm not wrong, this is a translated string so it might break. We could test `__( 'Default' )` instead but I wonder if there's a better way to identify this (a flag or something)

scruffian: Good catch. We do this elsewhere too: https://github.com/WordPress/gutenberg/blob/79d79ff565daa0a88746eb85bccd2a0778ac7a93/packages/edit-site/src/components/global-styles/hooks.js#L126

scruffian: Done in 2cfcf6ba4e

youknowriad (approved):
LGTM 👍 

Would be good to consider moving away from checking labels though. Someone can register a variable called "Default".

PR #61903, opened 2024-05-23
Navigation block: Check for update_ignored_hooked_blocks_postmeta in core

tjcafferkey:
Related PR: https://github.com/WordPress/wordpress-develop/pull/6604

## What?

Adding another `has_filter` check before adding the Navigation blocks filter to store `ignoredHookedBlocks` metadata as this functionality is being moved into WP core.

## Why?

The logic to add the `ignoredHookedBlocks` metadata has been moved to core to consolidate and haromise the functionality.

### DIFF
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -1620,7 +1620,7 @@ $rest_insert_wp_navigation_core_callback = 'block_core_navigation_' . 'update_ig
  * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
  * that are not present in Gutenberg's WP 6.5 compatibility layer.
  */
-if ( function_exists( 'set_ignored_hooked_blocks_metadata' ) && ! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) ) {
+if ( function_exists( 'set_ignored_hooked_blocks_metadata' ) && ( ! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) && ! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' ) ) ) {
 	add_filter( 'rest_pre_insert_wp_navigation', 'block_core_navigation_update_ignore_hooked_blocks_meta' );
 }
 
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -1621,10 +1621,9 @@ $rest_insert_wp_navigation_core_callback = 'block_core_navigation_' . 'update_ig
  * that are not present in Gutenberg's WP 6.5 compatibility layer.
  */
 if (
-	function_exists( 'set_ignored_hooked_blocks_metadata' ) && (
-		! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) ||
-		! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' )
-	)
+	! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' ) &&
+	function_exists( 'set_ignored_hooked_blocks_metadata' ) &&
+	! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback )
 ) {
 	add_filter( 'rest_pre_insert_wp_navigation', 'block_core_navigation_update_ignore_hooked_blocks_meta' );
 }
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -1620,7 +1620,7 @@ $rest_insert_wp_navigation_core_callback = 'block_core_navigation_' . 'update_ig
  * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
  * that are not present in Gutenberg's WP 6.5 compatibility layer.
  */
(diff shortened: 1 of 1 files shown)


packages/block-library/src/navigation/index.php
```diff
@@ -1620,7 +1620,7 @@ function block_core_navigation_update_ignore_hooked_blocks_meta( $post ) {
  * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
  * that are not present in Gutenberg's WP 6.5 compatibility layer.
  */
-if ( function_exists( 'set_ignored_hooked_blocks_metadata' ) && ! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) ) {
+if ( function_exists( 'set_ignored_hooked_blocks_metadata' ) && ( ! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) || ! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' ) ) ) {
```
ockham: Nitpick, I think we should wrap this line 😬 

```suggestion
if (
	function_exists( 'set_ignored_hooked_blocks_metadata' ) && (
		! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) ||
		! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' )
	)
) {
```

packages/block-library/src/navigation/index.php
```diff
@@ -1620,7 +1620,12 @@ function block_core_navigation_update_ignore_hooked_blocks_meta( $post ) {
  * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
  * that are not present in Gutenberg's WP 6.5 compatibility layer.
  */
-if ( function_exists( 'set_ignored_hooked_blocks_metadata' ) && ! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) ) {
+if (
+	function_exists( 'set_ignored_hooked_blocks_metadata' ) && (
+		! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback ) ||
+		! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' )
+	)
+) {
```
ockham, line 1628: Oh, I had another thought; we can re-shuffle the criteria a bit.

ockham: Reading the PHPDoc and the criterion, we might want to re-shuffle the criteria a bit to better map the intent here.
Specifically, if `update_ignored_hooked_blocks_postmeta` is hooked to `rest_pre_insert_wp_navigation`, we don't even need to check if `set_ignored_hooked_blocks_metadata` exists; that check was solely to ascertain we can hook `block_core_navigation_update_ignore_hooked_blocks_meta`, but for WP 6.6, we won't need to do that anymore.

ockham: Also, shouldn't the `||` be an `&&`? 🤔

ockham: Like so?

```suggestion
if (
	! has_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' ) &&
	function_exists( 'set_ignored_hooked_blocks_metadata' ) &&
	! has_filter( 'rest_pre_insert_wp_navigation', $rest_insert_wp_navigation_core_callback )
) {
```

So we add the filter here if

- Core hasn't added `update_ignored_hooked_blocks_postmeta`, and
- there's a `set_ignored_hooked_blocks_metadata` function (meaning that we're running on WP 6.5), and
- Core hasn't added `$rest_insert_wp_navigation_core_callback`.

Seem about right? 🤔

tjcafferkey: Ah you've saved me a commit there. I was about to commit this myself. Yes you're right, I came up with the same condition myself locally.

ockham: Great minds! 🧠🧠 😆

packages/block-library/src/navigation/index.php
```diff
@@ -1620,7 +1620,11 @@ function block_core_navigation_update_ignore_hooked_blocks_meta( $post ) {
  * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
  * that are not present in Gutenberg's WP 6.5 compatibility layer.
```
ockham:
```suggestion
 * Do not add the `block_core_navigation_update_ignore_hooked_blocks_meta` filter in the following cases:
 * - If Core has added the `update_ignored_hooked_blocks_postmeta` filter already (WP >= 6);
 * - or if the `set_ignored_hooked_blocks_metadata` function is unavailable (which is required for the filter to work. It was introduced by WP 6.5 but is not present in Gutenberg's WP 6.5 compatibility layer);
 * - or if the `$rest_insert_wp_navigation_core_callback` filter has already been added.
```

Fun 😬

packages/block-library/src/navigation/index.php
```diff
@@ -1617,10 +1617,16 @@ function block_core_navigation_update_ignore_hooked_blocks_meta( $post ) {
 $rest_insert_wp_navigation_core_callback = 'block_core_navigation_' . 'update_ignore_hooked_blocks_meta'; // phpcs:ignore Generic.Strings.UnnecessaryStringConcat.Found
 
 /*
- * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
- * that are not present in Gutenberg's WP 6.5 compatibility layer.
+ * Do not add the `block_core_navigation_update_ignore_hooked_blocks_meta` filter in the following cases:
+ * - If Core has added the `update_ignored_hooked_blocks_postmeta` filter already (WP >= 6);
```
ockham: ugh I typo'd

```suggestion
 * - If Core has added the `update_ignored_hooked_blocks_postmeta` filter already (WP >= 6.6);
```

tjcafferkey: It has now got to the point where I feel bad getting this merge stat once this gets approved 😂

ockham (approved):
Thanks a lot! 🚀

PR #61906, opened 2024-05-23
Site Editor: Fix the Root Padding styles

youknowriad:
Fix regression introduced in #61860 

## What?

The GlobalStyles styles generation hook had a hidden dependency towards the "BlockContext". The hook required that it would be applied within that context to generate the right styles for templates. That is a weird dependency. So when we moved the component to generate the styles, the context was not wrapping it anymore, so things started to break.

This PR solves that by removing this dependency in favor of an explicit argument.

## Testing Instructions

1- Open a template or a page in the site editor: root padding should be applied
2- Open a template part or a pattern: root padding should not be applied.

### DIFF
--- a/packages/block-editor/src/components/global-styles/use-global-styles-output.js
+++ b/packages/block-editor/src/components/global-styles/use-global-styles-output.js
@@ -33,7 +33,6 @@ import { getGapCSSValue } from '../../hooks/gap';
 import { store as blockEditorStore } from '../../store';
 import { LAYOUT_DEFINITIONS } from '../../layouts/definitions';
 import { getValueFromObjectPath, setImmutably } from '../../utils/object';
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/components/global-styles/use-global-styles-output.js
```diff
@@ -1224,10 +1232,6 @@ export function useGlobalStylesOutputWithConfig( mergedConfig = {} ) {
 		return !! getSettings().disableLayoutStyles;
 	} );
 
-	const blockContext = useContext( BlockContext );
```
youknowriad, line 1227: This was the bug, who says that useGlobalStylesOutput need to be run within "BlockContext" and why?

jasmussen (approved):
Nice. Works.

PR #61908, opened 2024-05-23
Block Editor: Fix `ZoomOutModeInserters` dependencies

tyxla:
## What?
This PR fixes the dependencies of one of `ZoomOutModeInserters`'s `useEffect` calls.

## Why?
From what I understand, there is no good reason not to provide `setInserterIsOpened` as a dependency:

- The effect will be called only once because of the mounted check
- The `setInserterIsOpened` reference is a stable reference to a block editor store action function. 

The motivation is resolving an ESLint error that was raised by the React Compiler ESLint plugin in https://github.com/WordPress/gutenberg/pull/61788

## How?
By adding `setInserterIsOpened` as a dependency and removing the ESLint rule disabling.

## Testing Instructions
* Enable the "Enable zoomed out view when patterns are browsed in the inserter" experiment
* Open the post editor
* Open the inserter and go to Patterns
* Select a pattern and verify that the inserters still look and work the same way as in `trunk` (they look broken btw, but I'm not 100% sure how they are expected to look and work, cc @youknowriad) 

None

## Screenshots or screencast 
None

### DIFF
--- a/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
+++ b/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
@@ -51,8 +51,7 @@ function ZoomOutModeInserters() {
 		}
 		// reset insertion point when the block order changes
 		setInserterIsOpened( true );
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, [ blockOrder ] );
+	}, [ blockOrder, setInserterIsOpened ] );
 
 	// Defer the initial rendering to avoid the jumps due to the animation.
 	useEffect( () => {


packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
```diff
@@ -51,8 +51,7 @@ function ZoomOutModeInserters() {
 		}
 		// reset insertion point when the block order changes
 		setInserterIsOpened( true );
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, [ blockOrder ] );
+	}, [ blockOrder, setInserterIsOpened ] );
```
ellatrix, line 54: I'm pretty sure I added this because it was misbehaving, but I can't remember after all the changes. What happens theoretically when the function changes? We only want to reset it when the order changes, so maybe we have to check that in the effect.

tyxla: Will this ever happen? It looks like a stable reference to me, so I don't understand when it will change. Plus we also have the mounted check. Happy to abandon it if I'm missing something of course.

Mamaduka: Right, actions should be safe to list in dependencies. If they're not anymore, we might have a bigger problem 😅

P.S. It would be cool if linter treated Redux actions as state setters, but IIRC that involved a lot of work.

ellatrix: Sure, if everything works fine. I could swear I added it for a reason 😅

ellatrix: We don't know it's an action though, it's a setting, right? I don't love these "impure" dependencies that much tbh

ellatrix (approved):

packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
Mamaduka: Yes. It's a setting pointing to the Editor package's `setIsInserterOpened` [action](https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/components/provider/use-block-editor-settings.js#L314).

For reference, this was the commit that disabled ESLint - ac1bbadb47308ffd0e1ffea7f0a7a91f254b9a0e.

tyxla: What George said. Thanks for verifying it, folks 🙌

PR #61912, opened 2024-05-23
Consolidate and fix `delete` and `edit` post actions

ntsekouras:
## What?
Part of: https://github.com/WordPress/gutenberg/issues/61787
Similar to: https://github.com/WordPress/gutenberg/pull/61857

This PR does two things:
1. Fix the `edit` action in site editor to be shown only in user patterns.
2. Fixes some logic of `delete` post action and keep one `delete` action for templates, template parts and patterns, and one `move to trash` action for the rest post types.

## Notes
I've updated the messages when removing a template, template part, pattern to have the generic `item` label, to simplify the logic there. For single item deletions nothing changes because we use the name of the item. It changes though for multiple items and for example the `An error occurred while deleting the templates.` is now `An error occurred while deleting the items.`

## Testing Instructions
1. Try the edit action in list views of `patterns, pages and templates`. In patterns it should be shown only for user created patterns and all template parts
2. Test that the delete action is working properly and that is shown only when it should(for example you cannot delete a theme template)
    - in site editor in all list views with single and multiple selection
    - in both editors for all post types(post, page, template, pattern, template part) in inspector controls

### DIFF
--- a/packages/edit-site/src/components/dataviews-actions/index.js
+++ b/packages/edit-site/src/components/dataviews-actions/index.js
@@ -26,16 +26,8 @@ export const useEditPostAction = () => {
 				if ( post.status === 'trash' ) {
 					return false;
 				}
-				// It's eligible for all post types except patterns.
-				if (
-					! [ ...Object.values( PATTERN_TYPES ) ].includes(
-						post.type
-					)
-				) {
-					return true;
-				}
-				// We can only edit user patterns.
-				return post.type === PATTERN_TYPES.user;
+				// It's eligible for all post types except theme patterns.
+				return post.type !== PATTERN_TYPES.theme;
 			},
 			callback( items ) {
 				const post = items[ 0 ];
--- a/packages/edit-site/src/components/dataviews-actions/index.js
+++ b/packages/edit-site/src/components/dataviews-actions/index.js
@@ -9,6 +9,7 @@ import { privateApis as routerPrivateApis } from '@wordpress/router';
 /**
  * Internal dependencies
  */
+import { PATTERN_TYPES } from '../../utils/constants';
 import { unlock } from '../../lock-unlock';
 
 const { useHistory } = unlock( routerPrivateApis );
@@ -21,8 +22,20 @@ export const useEditPostAction = () => {
 			label: __( 'Edit' ),
 			isPrimary: true,
 			icon: edit,
-			isEligible( { status } ) {
-				return status !== 'trash';
+			isEligible( post ) {
+				if ( post.status === 'trash' ) {
+					return false;
+				}
+				// It's eligible for all post types except patterns.
+				if (
+					! [ ...Object.values( PATTERN_TYPES ) ].includes(
+						post.type
+					)
+				) {
+					return true;
+				}
+				// We can only edit user patterns.
+				return post.type === PATTERN_TYPES.user;
 			},
 			callback( items ) {
 				const post = items[ 0 ];
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -38,10 +38,9 @@ const { PATTERN_TYPES } = unlock( patternsPrivateApis );
 
 /**
  * Check if a template is removable.
- * Copy from packages/edit-site/src/utils/is-template-removable.js.
  *
  * @param {Object} template The template entity to check.
- * @return {boolean} Whether the template is revertable.
+ * @return {boolean} Whether the template is removable.
  */
 function isTemplateRemovable( template ) {
 	if ( ! template ) {
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -70,31 +70,96 @@ function getItemTitle( item ) {
 	return decodeEntities( item.title?.rendered || '' );
 }
 
+// This action is used for templates, patterns and template parts.
+// Every other post type uses the similar `trashPostAction` which
+// moves the post to trash.
 const deletePostAction = {
 	id: 'delete-post',
 	label: __( 'Delete' ),
 	isPrimary: true,
 	icon: trash,
 	isEligible( post ) {
-		if ( [ 'auto-draft', 'trash' ].includes( post.status ) ) {
-			return false;
-		}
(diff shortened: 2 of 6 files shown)


packages/edit-site/src/components/dataviews-actions/index.js
```diff
@@ -21,8 +22,20 @@ export const useEditPostAction = () => {
...
 			icon: edit,
-			isEligible( { status } ) {
-				return status !== 'trash';
+			isEligible( post ) {
+				if ( post.status === 'trash' ) {
+					return false;
+				}
+				// It's eligible for all post types except patterns.
+				if (
+					! [ ...Object.values( PATTERN_TYPES ) ].includes(
```
youknowriad: Can't we just merge the two conditions into one. `post.type !== PATTERN_TYPES.theme`

packages/editor/src/components/post-actions/actions.js
```diff
@@ -64,23 +110,36 @@ const trashPostAction = {
 		const { createSuccessNotice, createErrorNotice } =
 			useDispatch( noticesStore );
 		const { deleteEntityRecord } = useDispatch( coreStore );
+		const { __experimentalDeleteReusableBlock } =
+			useDispatch( reusableBlocksStore );
+		const isPattern = items[ 0 ].type === PATTERN_POST_TYPE;
+		const deleteItem = isPattern
+			? ( item ) => __experimentalDeleteReusableBlock( item.id )
```
youknowriad: This action removes the reusable block but also updates the current block editor to remove that block. It feels a bit misplaced for me. A delete reusable block action should just delete the reusable block and not impact the block editor IMO. So maybe we should just unify the calls here.

ntsekouras: You mean using only the delete entity call, right? I didn't investigate why it was introduced, but it seemed it has some specific logic that might be still needed. I'll check it more.

packages/edit-site/src/components/dataviews-actions/index.js
ntsekouras: Of course 😄

packages/editor/src/components/post-actions/actions.js
```diff
@@ -37,25 +36,71 @@ import { exportPatternAsJSONAction } from './export-pattern-action';
 // Patterns.
 const { PATTERN_TYPES } = unlock( patternsPrivateApis );
 
+/**
+ * Check if a template is removable.
+ * Copy from packages/edit-site/src/utils/is-template-removable.js.
```
t-hamano: Since this PR has added new logic, should we remove this line?

ntsekouras: Good catch, thanks!

t-hamano (approved):
As far as I've tested, everything is working as expected.

One slight concern is that the "Move to Trash" has been changed to "Delete" in posts and pages. Some users may be confused as to whether the new "Delete" text, which previously read "Move to Trash," means "Move to Trash" or "Delete Permanently".

By the way, a "Delete Permanently" action may be added in the future, right?

ntsekouras:
>By the way, a "Delete Permanently" action may be added in the future, right?

There is one already, but the only way to access that action is through the pages list in `Trash` view. So personally I don't think it will create confusion.

richtabor:
> One slight concern is that the "Move to Trash" has been changed to "Delete" in posts and pages. Some users may be confused as to whether the new "Delete" text, which previously read "Move to Trash," means "Move to Trash" or "Delete Permanently".

I agree, this may be confusing. 

Deleting a post doesn't actually delete it, but moves it to the trash where it may be recovered; I do think that's an important distinction. For items that cannot be recovered, "Delete" is fine, but for items that can be, I think using "Move to Trash" is expected.

ntsekouras:
It seems there is no correlation between the ability to trash posts and how the post type is registered, so I'll keep two separate actions for now. One for template, patterns and template parts, and one for all other posts.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -202,7 +313,7 @@ const trashPostAction = {
 						disabled={ isBusy }
 						__experimentalIsFocusable
 					>
-						{ __( 'Delete' ) }
+						{ __( 'Trash' ) }
```
t-hamano, line 316:
```suggestion
						{ __( 'OK' ) }
```

I think "OK" is easier to understand, just like in WP6.5, what do you think?

### WP6.5

image

### This PR

image

jorgefilipecosta (approved):
Worked well on my tests 👍

Mamaduka (approved):
Testing well for me as well ✅

packages/editor/src/components/post-actions/actions.js
```diff
@@ -37,13 +36,123 @@ import { exportPatternAsJSONAction } from './export-pattern-action';
...
+	}
+	// In patterns list page we map the templates parts to a different object
+	// than the one returned from the endpoint. This is why we need to check for
+	// two props whether is custom or has a theme file.
+	return (
+		[ template.source, template.templatePart?.source ].includes(
+			TEMPLATE_ORIGINS.custom
+		) &&
+		! template.has_theme_file &&
+		! template.templatePart?.has_theme_file
```
youknowriad, line 57: I don't understand why the template part object is actually template.templatePart and not just template?

ntsekouras: In patterns list page we map the templates parts to a different object than the one returned from the endpoint. This is why we need to check two props.

youknowriad: Can we avoid mapping and just return the right objects?

Mamaduka: I encountered the same issue when fixing a bug for the "Reset" action - #62951. The fact that action might receive differently shaped objects when viewing a single or list of entities can be confusing.

youknowriad: I'm fixing that in https://github.com/WordPress/gutenberg/pull/62927

ntsekouras: > Can we avoid mapping and just return the right objects?

Probably.. I hadn't implemented that and will check your PR @youknowriad

PR #61915, opened 2024-05-23
Use available font weights and styles in FontAppearanceControl

### LINKED ISSUE #49090: Typography > Appearance controls should show available weights only

Currently adding Typography > Appearance controls to a block will render the entire range of font weight (100 - 900), plus their italic options: 

Screenshot 2023-03-15 at 11 35 17 AM

Some fonts do not have all of these weights, so unexpected results occur when an Appearance selection is made and the font does not have that weight supplied.

Would it be possible or a good idea to limit these options to only the weights that are available? 

I guess an alternative is to disable entirely the Typography > Appearance controls. 

cc @jasmussen @beafialho if you have any thoughts on this, thanks in advance!

mikachan:
## What?

Currently in Global Styles > Typography, the font Appearance dropdown includes a static list of options that is hard-coded in the component. Often many of these options aren't available to actually use, depending on the installed font families and the active font. 

This PR swaps out this list of static options and replaces them with a list of available font weights/styles based on the currently active font.

## Why?

Fixes https://github.com/WordPress/gutenberg/issues/49090.

## How?

Passes the list of font faces from the active font family to the `FontAppearanceControl` component. This can then be used instead of the static `FONT_WEIGHTS` and `FONT_STYLES` constants currently used to populate the Appearance dropdown menu.

## Testing Instructions

1. Open Global Styles > Typography
2. Open the Appearance dropdown menu
3. The menu should include only the weights and styles made available by the currently active font
4. Additionally, Bold is added as an Appearance option if the font doesn't already have a weight option of 600 or above, along with an italic option for each available weight
5. Change the active font by using the Font dropdown menu (ideally to a font that has different weights/styles)
6. Ensure that the Appearance dropdown menu is updated to reflect the new font

Please also test changing to a system font and a variable font (e.g. AR One Sans). System fonts should have all possible Appearance options, and variable fonts should have the range of weights included with the font shown in the Appearance dropdown.

## Screenshots or screencast 

| Before | After |
| ----- | ----- |
| Screenshot 2024-05-24 at 00 25 38 | Screenshot 2024-05-24 at 00 15 37 |


packages/block-editor/src/components/font-appearance-control/index.js
```diff
@@ -98,12 +104,48 @@ export default function FontAppearanceControl( props ) {
...
+	// Build font style and weight options based on the font family faces.
+	let fontWeights = [];
+	let fontStyles = [];
+
+	fontFamilyFaces?.forEach( ( face ) => {
+		if ( face.fontWeight ) {
+			if (
+				fontWeights.findIndex(
+					( weight ) => weight.value === face.fontWeight
+				) === -1
```
vcanales: Could we do something a bit more idiomatic with the conditions that are currently using `findIndex`? Such as:

```suggestion
				! fontWeight.some( ( weight ) => weight.value === face.fontWeight )
```

mikachan: Yes, nice idea! Done in https://github.com/WordPress/gutenberg/pull/61915/commits/a1bc596c1aad4622c0394cec9d75c763a582ace5.

mikachan:
When there is only one font style/weight option available, it looks a little odd having both a "Default" option and the single available option in the dropdown menu, for example:

image

In this case, Default is always going to be the same as the single available option.

Should we remove the Default option if there is only one font style/weight option? This also doesn't feel ideal as that means the Default option sometimes shows for other font faces, and then disappears when there is only one option.

packages/block-editor/src/utils/format-font-weight.js
```diff
@@ -0,0 +1,64 @@
...
+		return fontWeight;
+	}
+
+	let name;
+
+	switch ( fontWeight ) {
+		case 'normal':
+			name = _x( 'Regular', 'font weight' );
+			break;
+		case 'bold':
```
t-hamano: Is it necessary to consider weights other than `normal` and `bold`, such as `thin` and `black`?

Then, I think the case statements can be put together.

```javascript
		case 'normal':
		case '400':
			name = _x( 'Regular', 'font weight' );
			break;
```

packages/block-editor/src/utils/format-font-style.js
```diff
@@ -0,0 +1,43 @@
...
+			name = _x( 'Oblique', 'font style' );
+			break;
+
+		default:
+			name = fontStyle;
+			break;
+	}
+
+	// Ensure first letter of font style name is capitalized.
+	name = name.charAt( 0 ).toUpperCase() + name.slice( 1 );
```
t-hamano: I'm wondering if this is even necessary, since always capitalizing strings may not be appropriate in all locales. What do you think about removing this process?

t-hamano (commented):
In my testing, some fonts seem to display weights incorrectly. Is this because they are variable fonts?

image

packages/block-editor/src/utils/format-font-weight.js
mikachan: I didn't realise both `400` and `normal`, and `700` and `bold` could be combined, thanks! I've done that here: https://github.com/WordPress/gutenberg/pull/61915/commits/672ae3afdf3be1be1369220ce8c65b4e95668104.

Does this address your original comment? Or do you think we should remove `thin`, `extra light`, `light`, `medium`, `semi bold`, `extra bold`, and `black`? I added these as this follows the [CSS spec for `font-weight`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight), and it's closer to how we're currently showing these values.

packages/block-editor/src/utils/format-font-style.js
mikachan: I'm not sure it's necessary, when I added it I wasn't too sure about it. I'll remove it for now and we can always add this functionality back if needed.

Removed in https://github.com/WordPress/gutenberg/pull/61915/commits/dc60306de757f164f959a307779629c473b4b281.

packages/block-editor/src/utils/format-font-weight.js
vcanales: > Or do you think we should remove thin, extra light, light, medium, semi bold, extra bold, and black

The question was not for me, but I think those are worth keeping; fonts do offer that level of customization.

mikachan:
> In my testing, some fonts seem to display weights incorrectly. Is this because they are variable fonts?

Thanks for highlighting this, @t-hamano. I tested using the variable version of AR One Sans and this seemed to work fine, but I see the same options as you with Inter. 

I've added a fix [here](https://github.com/WordPress/gutenberg/pull/61915/commits/3ba9ce861847fd7d8d6acebb3460dc518063659a) that checks if the font weight includes a space, as in the case of Inter, the font weight is coming through as `100 900`. I'm thinking that if the weight includes a space, then it must be a variable font. But I'm not sure if this is good enough for checking for variable fonts - maybe we should check something like the `fvar.axes` table records from the font file.

packages/block-editor/src/components/font-appearance-control/index.js
```diff
@@ -98,12 +105,46 @@ export default function FontAppearanceControl( props ) {
...
+
+	fontFamilyFaces?.forEach( ( face ) => {
+		if ( face.fontWeight ) {
+			if (
+				! fontWeights.some(
+					( weight ) => weight.value === face.fontWeight
+				)
+			) {
+				// Check if font weight includes a space, if so it must be a variable font.
+				if ( /\s/.test( face.fontWeight ) ) {
```
creativecoder: Might be best to trim `face.fontWeight` before testing, to make sure any detected space isn't stray padding at the start or end of the string.

packages/block-editor/src/utils/test/format-font-weight.js
```diff
@@ -0,0 +1,54 @@
...
+			value: '700',
+		} );
+		expect( formatFontWeight( '800' ) ).toEqual( {
+			name: 'Extra Bold',
+			value: '800',
+		} );
+		expect( formatFontWeight( '900' ) ).toEqual( {
+			name: 'Black',
+			value: '900',
+		} );
```
creativecoder, line 52: Does it make sense to add additional tests for `normal` and `bold` here?

creativecoder (approved):
Overall this is testing well for me.

- System fonts: I see all appearance options
- Non-variable fonts: I see the weight/style appearance options change depending upon which font faces are activated for the font
- Variable fonts: I see all weight/style options

A nice enhancement for variable fonts would be checking which weights and styles are available and limiting the options, if possible. For example, right now if I install a variable font Cabin that covers only italics style with 400 - 700 weight, I still see all styles and appearance options. But we can address that in a follow-up.

I'll go ahead and approve this as it seems to be working as expected for me, but let's keep testing to make sure there aren't more edge-cases/bugs.

creativecoder:
> Should we remove the Default option if there is only one font style/weight option? This also doesn't feel ideal as that means the Default option sometimes shows for other font faces, and then disappears when there is only one option.

No, I think we need to keep the "Default" option. While it's a bit awkward, it has a specific meaning: don't override the font appearance, and instead inherit from the user global styles and/or theme.json.

creativecoder:
> I've added a fix [here](https://github.com/WordPress/gutenberg/pull/61915/commits/3ba9ce861847fd7d8d6acebb3460dc518063659a) that checks if the font weight includes a space, as in the case of Inter, the font weight is coming through as 100 900. I'm thinking that if the weight includes a space, then it must be a variable font. But I'm not sure if this is good enough for checking for variable fonts - maybe we should check something like the fvar.axes table records from the font file.

I don't think we need to check the font file itself. At this point, we're depending on the font-face rules that have been saved for the file--that's what the browser will use to match the element CSS to the font-face rule.

packages/block-editor/src/utils/format-font-weight.js
creativecoder: `thin` and `black` aren't valid CSS values for font-weight, but only labels that are mapped to numerical weights by convention. So to me this now looks good to go: check for the numerical values 100 - 900 or the text values `normal`/`bold`, and map those to a common label, when possible.

packages/block-editor/src/components/font-appearance-control/index.js
mikachan: Nice idea, thanks! Done here: https://github.com/WordPress/gutenberg/pull/61915/commits/7677b53099f7572434dffd91080c980f89b5b044.

packages/block-editor/src/utils/test/format-font-weight.js
mikachan: Yes I think it does - done here: https://github.com/WordPress/gutenberg/pull/61915/commits/362d1c7afccc2841b165165c77e4f750cf5fb21f.

packages/block-editor/src/components/font-appearance-control/index.js
```diff
@@ -116,6 +117,11 @@ export default function FontAppearanceControl( props ) {
 					( weight ) => weight.value === face.fontWeight
 				)
 			) {
+				// Check if font weight includes a space, if so it must be a variable font.
+				if ( /\s/.test( face.fontWeight ) ) {
```
matiasbenedetto: We should probably trim left and right, split by space, and try to get the first and second numbers.

mikachan:
Thanks @creativecoder!

> A nice enhancement for variable fonts would be checking which weights and styles are available and limiting the options, if possible.

That would be a great follow-up. For now, this is similar to the current behaviour in trunk, but it's at least limited to variable fonts.

packages/block-editor/src/components/font-appearance-control/index.js
mikachan: But in the case of Inter, doesn't the font weight string `100 900` represent a range of weights from 100-900? So in this case, it would only list 100 and 900 in the editor.

Maybe we should split by space and then try to work out the list of available fonts between the two numbers? e.g. in the case of `100 900`, we can assume we want all the weights between 100-900: 100, 200, 300, 400, 500, 600, 700, 800, and 900.

t-hamano: I think the tricky part is that CSS Fonts level 4 supports weight 1000, and some Google web fonts support that weight too. For example, it seems like there are fonts with a weight 1000 in [this list of variable fonts](https://fonts.google.com/variablefonts) as well.

Also, I don't know much about variable fonts, but is it possible that the weights are not in units of 100? If so, it's too complicated for this PR and we might be able to keep the logic as it is now.

mikachan: I agree, I think for now, at least the behaviour in this PR is better than trunk, and we can follow-up with more detailed handling for variable fonts.

packages/block-editor/src/components/font-appearance-control/index.js
```diff
@@ -98,12 +105,46 @@ export default function FontAppearanceControl( props ) {
...
+				fontStyles.push( formatFontStyle( face.fontStyle ) );
+			}
+		}
+	} );
+
+	fontWeights =
+		fontWeights.length === 0 || variableFont ? FONT_WEIGHTS : fontWeights;
+	fontStyles =
+		fontStyles.length === 0 || variableFont ? FONT_STYLES : fontStyles;
+
```
matiasbenedetto: This logic is complex enough to be worth extracting as a pure function and adding test cases for it. The tests would also help to document the expected output for each theme.json definition. I think that's useful for the follow-up tasks regarding variable typographies, too.

packages/block-editor/src/components/font-appearance-control/index.js
matiasbenedetto: I think I wasn't clear enough. I reformulate: the presence of a space character is not a great way to tell if a font face is variable. 

Look at this example. Notice the trailing space in the fontWeight of the first fontFace. The space breaks the implementation, and the result is listing all the rage of font weights. Probably we should apply `trim()` before checking for spaces.
```
{
					"fontFace": [
						{
							"fontFamily": "Cardo",
							"fontStyle": "normal",
							"fontWeight": "400 ",
							"src": [
								"file:./assets/fonts/cardo/cardo_normal_400.woff2"
							]
						},
						{
							"fontFamily": "Cardo",
							"fontStyle": "italic",
							"fontWeight": "400",
							"src": [
								"file:./assets/fonts/cardo/cardo_italic_400.woff2"
							]
						},
					],
					"fontFamily": "Cardo",
					"name": "Cardo",
					"slug": "heading"
				}
```

mikachan: Ah I see! Moved the `.trim()` before we check for spaces here: https://github.com/WordPress/gutenberg/pull/61915/commits/aba764dc91a26685b3a4f51420f26b94ed42f935.

packages/block-editor/src/components/font-appearance-control/index.js
mikachan: Makes sense! I've abstracted this to a function here: https://github.com/WordPress/gutenberg/pull/61915/commits/e6d2e67d0898b361c13fd092bae6497e7529b010.

packages/block-editor/src/components/font-appearance-control/index.js
creativecoder: @mikachan I think what we had before https://github.com/WordPress/gutenberg/commit/aba764dc91a26685b3a4f51420f26b94ed42f935 was actually what we need. `String.trim()` doesn't modify the value in place; it returns a new value.

mikachan: Thanks @creativecoder - I've tested @matiasbenedetto's example above with both doing `.trim()` first separately, and having it within the if statement, and both work the same. So I think the version with less code is best (i.e. before https://github.com/WordPress/gutenberg/commit/aba764dc91a26685b3a4f51420f26b94ed42f935). Thanks! Updated here: https://github.com/WordPress/gutenberg/pull/61915/commits/7bd07ab08120159cd65df8087605687bd36a439a. Also improved the inline comment.

matiasbenedetto (commented):
The listing of the alternative font styles (italic) seems erroneous.

Using TT4 default 'Inter' font family definition:
```json
				{
					"fontFace": [
						{
							"fontFamily": "Inter",
							"fontStretch": "normal",
							"fontStyle": "normal",
							"fontWeight": "300 900",
							"src": [
								"file:./assets/fonts/inter/Inter-VariableFont_slnt,wght.woff2"
							]
						}
					],
					"fontFamily": "\"Inter\", sans-serif",
					"name": "Inter",
					"slug": "body"
				},
```
the italic styles are added to the list when the font family doesn't define any font face for italic.

image

Another example.
Install 'Piazzolla' font with normal-100, nomal-200 and italic-800 and italic-900 styles.
This is the list rendered:
image

**Expected:**
The font styles available should be added only if there are font faces available for those styles (variable or not) or if it's a system font.

vcanales:
> the italic styles are added to the list when the font family doesn't define any font face for italic.

The default installation of Inter does seem to support those styles, since it's a variable type.

https://github.com/WordPress/gutenberg/assets/1157901/6736733e-28e2-4610-a250-527b8e4795d3

> Install 'Piazzolla' font with normal-100, nomal-200 and italic-800 and italic-900 styles.

I was able to confirm this. It seems like the UI is currently rendering a range of options, instead of only the 2 available, both for weight and style.

packages/block-editor/src/utils/test/get-font-styles-and-weights.js
```diff
@@ -0,0 +1,97 @@
...
+					name: 'Extra Bold',
+					value: '800',
+				},
+				{
+					name: 'Black',
+					value: '900',
+				},
+			],
+		} );
+	} );
```
vcanales, line 512: Wrote a test for [this](https://github.com/WordPress/gutenberg/pull/61915#pullrequestreview-2094185074) — it does fail.

> Install 'Piazzolla' font with normal-100, nomal-200 and italic-800 and italic-900 styles.
> Expected:
> The font styles available should be added only if there are font faces available for those styles (variable or not) or if it's a system font.

```suggestion
	} );

	it( 'should return only provided font styles and weights', () => {
		const fontFamilyFaces = [
			{
				fontFamily: 'Piazzolla',
				fontStyle: 'normal',
				fontWeight: '400',
				src: 'http://www.wordpress.org/wp-content/uploads/fonts/N0b52SlTPu5rIkWIZjVKKtYtfxYqZ4RJBFzFfYUjkSDdlqZgy7LYxnL31AHfAAy5.woff2',
			},
			{
				fontFamily: 'Piazzolla',
				fontStyle: 'normal',
				fontWeight: '900',
				src: 'http://www.wordpress.org/wp-content/uploads/fonts/N0b52SlTPu5rIkWIZjVKKtYtfxYqZ4RJBFzFfYUjkSDdlqZgy7JxwXL31AHfAAy5.woff2',
			},
			{
				fontFamily: 'Piazzolla',
				fontStyle: 'italic',
				fontWeight: '300',
				src: 'http://www.wordpress.org/wp-content/uploads/fonts/N0b72SlTPu5rIkWIZjVgI-TckS03oGpPETyEJ88Rbvi0_TzOzKcQhcSx3gD9BRy5m5M.woff2',
			},
			{
				fontFamily: 'Piazzolla',
				fontStyle: 'italic',
				fontWeight: '900',
				src: 'http://www.wordpress.org/wp-content/uploads/fonts/N0b72SlTPu5rIkWIZjVgI-TckS03oGpPETyEJ88Rbvi0_TzOzKcQhTO23gD9BRy5m5M.woff2',
			},
		];

		expect( getFontStylesAndWeights( fontFamilyFaces ) ).toEqual( {
			fontStyles: [
				{
					name: 'Regular',
					value: 'normal',
				},
				{
					name: 'Italic',
					value: 'italic',
				},
			],
			fontWeights: [
				{
					name: 'Light',
					value: '300',
				},
				{
					name: 'Black',
					value: '900',
				},
			],
		} );
	} );
```

mikachan: Thanks for this. I've switched the less complicated `ABeeZee` test with your `Piazzolla` test above in https://github.com/WordPress/gutenberg/pull/61915/commits/1ca996bfd54ada3246596aab3ae22586652227b3.

mikachan:
Thanks @matiasbenedetto and @vcanales! It looks like this PR was only addressing part of the issue. Basically what you said here:

> It seems like the UI is currently rendering a range of options, instead of only the 2 available, both for weight and style.

Currently, there is a function that combines all font styles and weights regardless of which font faces are active/installed, and this function combines these values and then creates a list with all the combined options (although it's combining fewer options compared to trunk 😅)

I've updated this so that the styles and weights are only combined into a larger list if the font is a system font or a variable font, otherwise, the listed options will only include the font faces that are installed. To confirm, the logic is now:

- If the font is a system font, the default font styles and weights will be used.
- If the font is a variable font, the default font styles and weights will be used until we improve this in a follow-up PR.
- If the font is a specific weight/style, then these weights and styles will be included in the Appearance dropdown and will not be combined to make new styles/weight combinations.

Here is the same example from above with 'Piazzolla', using normal-100, nomal-200 and italic-800 and italic-900 styles:

image

I think this is ready for another review, thanks all 🙇

vcanales (commented):
I've noticed some behaviors that might not bet expected:

1. I select a font for Text, that includes a Style such as Regular Italic, and set that style.
Result: The style changes for all Text, even the text that is not in the selected font. Not sure if this has always been the case, or if it's actually desired.
3. I go to a font that does not include that style and select it.
Result: I can't unset the Italics if I select a font that does not have italics.
The font I selected, System font in this case, is still in Italics, even though it's not listed as available.


https://github.com/WordPress/gutenberg/assets/1157901/085557ce-e057-4198-aafc-6859e7f32ca4

mikachan:
Thanks for testing again, @vcanales.

> I select a font for Text, that includes a Style such as Regular Italic, and set that style.
Result: The style changes for all Text, even the text that is not in the selected font. Not sure if this has always been the case, or if it's actually desired.

I _think_ this has always been the case, but I don't think changing the style of even the text that is in a different font should be the desirable behavior. I'll do some more investigation, but perhaps this would be a good fix in a follow-up PR.

> I go to a font that does not include that style and select it.
Result: I can't unset the Italics if I select a font that does not have italics.
The font I selected, System font in this case, is still in Italics, even though it's not listed as available.

Oh good spot! So I _think_ again, this may have always been the case, but now that the available font weights/styles are limited to the activated font faces, it's made this issue more obvious. Previously, "italic" would have remained in the dropdown for all font faces, and I believe the browser will attempt to make any font face italic/bold (faux italic or faux bold), which makes this difficult to notice (i.e. it'll look like it works for most fonts, even if the font face isn't installed). I'm guessing the `font-style` property isn't being updated correctly in the editor when you switch between styles and I think this should be fixed as part of this PR - I'll look into this next.

mikachan:
I've updated this so that the Appearance dropdown is updated whenever the font family is changed. This ensures that the correct font styles and weights are applied according to the active and installed font faces. If the same style and weight combination exists across multiple font families, then the same appearance option should be applied. Otherwise, the appearance dropdown will be reset to Default.

To test:

1. Select a font family that has many appearance options, such as a system font.
2. Select an appearance that doesn't exist on another installed font family (e.g. Semi Bold)
3. Select the font family that does not have the same style/weight combination (Semi Bold)
4. The appearance dropdown should reset to Default and the Default styles should be applied in the Editor
5. Select an appearance option that does exist for another font family
6. Select the other font family where this option exists
7. Ensure that the appearance option does not change and the styles are still successfully applied in the Editor

https://github.com/WordPress/gutenberg/assets/1645628/c6638734-6dc0-4ea1-94d3-22cd8e8654ac

I think this is ready for another review. Thanks all 🙇

vcanales (commented):
> 5. Select an appearance option that does exist for another font family
> 6. Select the other font family where this option exists
> 7. Ensure that the appearance option does not change and the styles are still successfully applied in the Editor
> 

These steps work as described when I move from a Semi-Bold Font into the System Font, but not the other way around. It's also resetting the style when moving between two custom fonts that do include variations such as Semi Bold. 

https://github.com/WordPress/gutenberg/assets/1157901/a3be95ea-a83e-4eac-a407-9f746dff5adc


I think this is happening because the weight comparison is missing some nuance in how the weights are defined in variable fonts:

| fontFamilyFaces | useEffect |
| -- | -- |
| image | image|

mikachan:
Thanks so much @vcanales 🙏 

> when I move from a Semi-Bold Font into the System Font, but not the other way around. 

I thought I'd tested this and it was working, but I don't think I tried with a variable font that had a limited range of weights. I'm able to reproduce this and I've attempted a fix that compares the formatted font weights and styles from the new `getFontStylesAndWeights` function (which formats the values into individual font weights), rather than the raw font faces array (which includes values like `400 700`). Does this work better for you now, after https://github.com/WordPress/gutenberg/pull/61915/commits/ec2cf86ffa15435da8c925b7c70a93de75fda14e?

vcanales (approved):
> Does this work better for you now, after https://github.com/WordPress/gutenberg/commit/ec2cf86ffa15435da8c925b7c70a93de75fda14e?

Yes, that took care of that case :)

This is looking good to me now.

creativecoder (commented):
I've tested this, trying to switch between different font families in as many ways I can think of: switching the font family to/from a system font, a web font with one font face, a web font with multiple font faces, and variable fonts both with and without italics.

Everything seems to be working as described, mainly:

- ✅ System fonts and variable fonts have all weights and styles available, so switching to them never changes the font appearance style. The same when the web font has a font face with the current appearance style available--it stays the same.
- ✅ Regular web fonts (not variable) show only the appearance styles for the font faces they have activated, so switching to a web font when the appearance style is not available resets the appearance to "Default"

---

That said, while testing and thinking about the problem more, the way this takes away flexibility gives me some pause at merging it. 

First, **font weight**: showing only the weight options the font has available is generally a good improvement. Before this PR, changing the Appearance from "Thin" to "Medium" or from "Black" to "Bold" for a font that only has Regular and Bold faces available has no effect--the extra choices are just clutter.

However, for a font with only a Regular weight font face active, many browsers will create a bold style as well--it seems you can turn any font bold, and this PR takes away that option.

Additionally, with this PR, changing the font family will reset Appearance to "Default" if the font doesn't have a face that matches exactly. For example, if I want the Appearance to be Bold, but the font only has a Black weight available, the Appearance setting goes back to Default when I select the new font. I loose the Bold setting and have to manually change it every time I change the font if I want to preview a bold-like appearance.

Second, the **italic** setting: before this PR, it you select italic appearance for a font that doesn't have one, many browsers will automatically create an italic font style. So this change takes away the option to make any font italic from the block settings.

Another reason I find this change awkward is how the Global Styles settings can be different. For example, you can set the default Appearance for a block to be Bold in Global Styles settings. Then when the individual block setting Appearance is "Default", the font is Bold, even when that font doesn't have an Bold font face. But you can't find Bold in the list of Appearance options unless the font has the face installed. I'm worried this inconsistency will cause confusion.

Some potential changes to address these problems:

- Keep the "Bold" Appearance option available for all fonts, even when the font face isn't explicitly installed
- Keep an Italic Appearance option available for each font weight available, even when the font face isn't explicitly installed (e.g. if a font has only Black font weight installed, also show "Black Italic" as an option)
- Try to match the closest weight option available when switching fonts, rather than going back to "Default" (e.g. if the Appearance settings is Bold, and switching to a font that only has Black available, change the Appearance to Black rather than default)

What do you think?

mikachan:
Thanks for the extensive testing, @creativecoder! I'm glad to see the green ticks 😄 

> many browsers will create a bold style as well--it seems you can turn any font bold

I believe when browsers apply bold and italic to a regular font, this is called "faux bold" and "faux italic", and sometimes the results are not ideal. For example, some characters will look weird/wrong and may be hard to read, and each browser will apply these styles slightly differently. 

However, having thought about this more after your comment, I think there are more reasons to include these faux options:

- It allows for more flexibility (and maintains the current flexibility)
- These faux options can still be applied via the block toolbar, theme.json, and custom CSS
- Does not create any additional confusion between what can be applied via theme.json/CSS vs global styles

I've included them in https://github.com/WordPress/gutenberg/pull/61915/commits/d359d9ec7406e124a43d0d1f02c45664ebabfd4b, let me know what you think!

> Try to match the closest weight option available when switching fonts, rather than going back to "Default"

I think this is a good idea! I've tried this out in https://github.com/WordPress/gutenberg/pull/61915/commits/ad00060e33fcdf4690aa15227d738cc0e6a24c3f.

I've also gone ahead and handled variable font weights in this commit: https://github.com/WordPress/gutenberg/pull/61915/commits/0ece8f5bee1a58ef5521f3c795451911bec1d614. I was working on this anyway as a follow-up and since it seems to be in a good state, I've included it here.

---

Some new testing steps following these changes:

- Ensure bold and italic options are available on every font
- When testing with a variable font, ensure that the correct range of font weights is available. e.g. if the range is 400-700, the font weights from 400-700 should only be included.
- When testing with a font that includes a font weight of 1000, ensure that this option is included ("Extra Black"). An example font is variable DM Sans.

creativecoder:
> I believe when browsers apply bold and italic to a regular font, this is called "faux bold" and "faux italic", and sometimes the results are not ideal. For example, some characters will look weird/wrong and may be hard to read, and each browser will apply these styles slightly differently.

Yeah, I have mixed feelings about the faux settings. Ideally you would install the font faces you need, but having bold/italic appearance always available is handy if you just want one thing on the page to look a bit different without the overhead of another font face. I hope that in the future we can help users reconcile this more transparently in some way.

> I've also gone ahead and handled variable font weights in this commit: https://github.com/WordPress/gutenberg/commit/0ece8f5bee1a58ef5521f3c795451911bec1d614. I was working on this anyway as a follow-up and since it seems to be in a good state, I've included it here.

Nice. I noticed this as well, but my review was already getting quite long. Glad to see it added--will test!

creativecoder:
I've updated the logic in https://github.com/WordPress/gutenberg/pull/61915/commits/d7bba77c9f0368a3b4176d3d87389b65eb86e65e to only add faux bold if there isn't already a font weight available of 600 or above.

In my testing, the addition of faux Bold does not display any differently for a font face that already has a Semi Bold weight or above. With this change faux Bold is only added if it will actually change the appearance. (Of course, it's still dependent on the browser supporting faux bold text).

creativecoder (approved):
Thanks for the quick updates @mikachan !

With the recent changes, I think this is ready to merge, as this PR now only removes font Appearance options that have no effect. If users are currently depending on faux bold or faux italic options, those will still be available.

- ✅ All fonts have a bold Appearance option, as well as italic options for each weight
- ✅ System fonts have all weights and styles available
- ✅ Regular web fonts (not variable) show only the Appearance styles for the font faces they have activated (plus added bold and italic options)
- ✅ Variable fonts only show Appearance options for the weight range included with the font (plus added italic options)

mikachan:
> I've updated the logic in https://github.com/WordPress/gutenberg/commit/d7bba77c9f0368a3b4176d3d87389b65eb86e65e to only add faux bold if there isn't already a font weight available of 600 or above.

Ah nice, I didn't realise this about Semi Bold. Thank you!

> With the recent changes, I think this is ready to merge

Woop, thank you for reviewing again! And thank you to everyone who has helped out here! I'll go ahead and merge this 🎉

packages/block-editor/src/components/global-styles/typography-panel.js
```diff
@@ -252,6 +255,39 @@ export default function TypographyPanel( {
...
+			setFontAppearance( {
+				fontStyle,
+				fontWeight,
+			} );
+		}
+
+		// Reset font appearance if the font family does not have the selected font style
+		if ( ! hasFontStyle ) {
+			resetFontAppearance();
+		}
```
talldan, line 288: @mikachan I noticed here that if `isSystemFont` is `true` and if `hasFontWeight` has a value, then if first calls `setFontAppearance` to set the font weight, but will straight away call `resetFontAppearance` and unsets the value that was just set.

The background is that I've been working on some e2e tests (#61932) that programatically set `fontWeight` a bit like this:
```js
var block = wp.blocks.createBlock('core/paragraph', {
  content: 'Hello',
  style: { typography: { fontWeight: '300' } } }
);
wp.data.dispatch('core/block-editor').insertBlock( block )
```

But that recently stopped working, and I think it's due to the `resetFontAppearance` here. To reproduce, if you run that in the console the block that's inserted ends up with no appearance styles.

I'm not familiar with the code, but thought I'd mention it.

mikachan: Thanks for highlighting, @talldan! I'll take a look and open a follow-up PR. I'm guessing the `hasFontStyle` and `hasFontWeight` combined logic needs to be adjusted.

mikachan: @talldan I've opened a follow-up here that hopefully fixes this: https://github.com/WordPress/gutenberg/pull/63215 🤞

packages/block-editor/src/components/global-styles/typography-panel.js
```diff
@@ -252,6 +255,39 @@ export default function TypographyPanel( {
 		setFontAppearance( {} );
 	};
 
+	// Check if previous font style and weight values are available in the new font family
+	useEffect( () => {
```
noisysocks, line 259: This effect runs when the component first mounts and will potentially create undo levels when `setFontAppearance` calls `onChange` which in turn calls `setAttributes`. Is that what we want? 

I'm also running into a problem with some new code I'm writing where this effect is clobbering `style` because it's calling `onChange` on mount with an outdated `value`.

Is an effect necessary? Could the logic be in the callback that triggers when a new font family is selected?

mikachan: > This effect runs when the component first mounts and will potentially create undo levels when setFontAppearance calls onChange which in turn calls setAttributes. Is that what we want?

Hmm, no, we want this to update when a new font family is set or when a new block is inserted with font appearance settings.

> Is an effect necessary? Could the logic be in the callback that triggers when a new font family is selected?

Which callback do you mean here? Currently, `setFontFamily` isn't a callback, but maybe it should be one 🤔 

I'll investigate and see if we can avoid the `useEffect`.

mikachan: @noisysocks I think I've got something that avoids running the effect when the component first mounts - using a `useRef` to store the `fontFamily` value and then checking against that before running the logic. I've added this to the follow-up PR in this commit: https://github.com/WordPress/gutenberg/pull/63215/commits/18e1bfebf9b50cae3bb9fc81d1bf3814681b7922. Please let me know if this works better with what you're working on!

noisysocks: Thanks I'll give that PR a spin!

PR #61917, opened 2024-05-23
Components: Fix React Warning triggers by the new JSX transform

youknowriad:
Just a tiny PR to get rid of a React warning that is triggered when you open one of the color element dropdowns in the global styles sidebar. If you have SCRIPT_DEBUG enabled, you can see a warning in trunk when opening components panels in global styles, this PR solves that.

### DIFF
--- a/packages/block-editor/src/components/global-styles/color-panel.js
+++ b/packages/block-editor/src/components/global-styles/color-panel.js
@@ -209,7 +209,7 @@ function ColorPanelDropdown( {
 	panelId,
 } ) {
 	const currentTab = tabs.find( ( tab ) => tab.userValue !== undefined );
-
+	const { key: firstTabKey, ...firstTab } = tabs[ 0 ] ?? {};
 	return (
 		<ToolsPanelItem
 			className="block-editor-tools-panel-color-gradient-settings__item"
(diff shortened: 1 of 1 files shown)


packages/block-editor/src/components/global-styles/color-panel.js
```diff
@@ -251,7 +251,8 @@ function ColorPanelDropdown( {
 						<div className="block-editor-panel-color-gradient-settings__dropdown-content">
 							{ tabs.length === 1 && (
 								<ColorPanelTab
-									{ ...tabs[ 0 ] }
+									keu={ firstTabKey }
```
ellatrix: Typo? `kue`?

ellatrix: I'm not seeing a warning when opening the Colors tab in global styles 🤔

youknowriad: Are you sure you have SCRIPT_DEBUG true

youknowriad: Also it happens only when there's just one tab in the color dropdown.

Mamaduka (approved):
I can reproduce the issue.

Should we try running e2e tests in a similar mode and see how many warnings it will turn up?

## Screenshot
CleanShot 2024-05-24 at 11 57 39

packages/block-editor/src/components/global-styles/color-panel.js
ellatrix: Yes, it's true. I guess I'm not sure what you mean with dropdown, but I have one palette and multiple "elements"

youknowriad:
> Should we try running e2e tests in a similar mode and see how many warnings it will turn up?

Locally, that could be a good idea, in CI I think e2e tests should probably run in prod.

Mamaduka:
I meant test PR just to see warnings and fix them.

Issue #60291, opened 2024-03-28
Polish design of edit interfaces for meta in page inspector


jameskoster:
As a part of the work in https://github.com/WordPress/gutenberg/issues/59689 it would be nice to tidy up the edit interface for page meta data. These UIs looking a little rough around the edges, and the designs in 59689 call for some options like discussion and parent selection to move into the Summary panel.

Here's the current state of affairs:

![current-panels](https://github.com/WordPress/gutenberg/assets/846565/ecc1ad17-fe92-4fb4-8ebd-65b0419773c5)

- [x] Slug https://github.com/WordPress/gutenberg/pull/60632
- [x] Page attributes: https://github.com/WordPress/gutenberg/issues/61869
- [x] Discussion
- [x] Author
- [x] Post status https://github.com/WordPress/gutenberg/pull/60456

jameskoster:
Status panel is being worked on in https://github.com/WordPress/gutenberg/pull/60456.

jameskoster:
Mockups for the other panels:

<img width="1404" alt="Screenshot 2024-04-04 at 14 40 25" src="https://github.com/WordPress/gutenberg/assets/846565/eb5ce261-c7eb-455c-910f-5767bb2b5421">

ntsekouras:
@jameskoster  what about the pingbacks & trackbacks in posts? How do they fit with the comments setting? You can see the current checkbox in post editor in `posts`.

jameskoster:
It could live in the discussion popover as a checkbox for now, I suppose that would be least disruptive. But it _might_ be too prominent there – seems like a setting more likely to be configured at the global level, and rarely overridden locally. We could consider placing it in the ellipsis menu down the road:

<img width="1178" alt="Screenshot 2024-04-10 at 09 47 03" src="https://github.com/WordPress/gutenberg/assets/846565/3f02c3af-405b-46d7-9269-4348afc17ff6">

ntsekouras:
>It could live in the discussion popover as a checkbox for now,

What about the label of the button though? It can't just say `comments open/closed`, right?

Also there is the possibility that we only have the trackbaacks checkbox available and not the comments, if the post type supports only that..

jameskoster:
Perhaps it moves to the ellipsis now, then :) It doesn't feel worthy of a main item in the summary panel to me, but happy to hear other opinions about that.

jasmussen:
[...]

PR #61918, opened 2024-05-23
Update: Implement new parent and order design.

jorgefilipecosta:
This PR implements the new parent and order design.
Related https://github.com/WordPress/gutenberg/issues/60291, https://github.com/WordPress/gutenberg/issues/59689.

## Screenshot
Screenshot 2024-05-23 at 19 48 35

## Testing Instructions
Verified I could still change the parent and order of a page.

### DIFF
--- a/packages/editor/src/components/page-attributes/panel.js
+++ b/packages/editor/src/components/page-attributes/panel.js
@@ -1,161 +1,17 @@
 /**
  * WordPress dependencies
  */
-import { __, sprintf } from '@wordpress/i18n';
-import { Dropdown, Button } from '@wordpress/components';
 import { useSelect } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
-import { useMemo, useState } from '@wordpress/element';
-import { decodeEntities } from '@wordpress/html-entities';
-import { __experimentalInspectorPopoverHeader as InspectorPopoverHeader } from '@wordpress/block-editor';
-
 /**
  * Internal dependencies
  */
-import PostPanelRow from '../post-panel-row';
 import { store as editorStore } from '../../store';
 import PageAttributesCheck from './check';
-import PageAttributesOrder from './order';
-import PageAttributesParent from './parent';
+import { OrderRow } from './order';
+import { ParentRow } from './parent';
 
 const PANEL_NAME = 'page-attributes';
-
-function getTitle( post ) {
-	return post?.title?.rendered
-		? decodeEntities( post.title.rendered )
-		: `#${ post.id } (${ __( 'no title' ) })`;
-}
-
-function PostParentToggle( { isOpen, onClick } ) {
-	const parentPost = useSelect( ( select ) => {
-		const { getEditedPostAttribute } = select( editorStore );
-		const parentPostId = getEditedPostAttribute( 'parent' );
-		if ( ! parentPostId ) {
-			return null;
-		}
-		const { getEntityRecord } = select( coreStore );
-		const postTypeSlug = getEditedPostAttribute( 'type' );
-		return getEntityRecord( 'postType', postTypeSlug, parentPostId );
-	}, [] );
-	const parentTitle = useMemo(
-		() => ( ! parentPost ? __( 'None' ) : getTitle( parentPost ) ),
-		[ parentPost ]
-	);
-	return (
-		<Button
-			size="compact"
-			className="editor-post-parent__panel-toggle"
-			variant="tertiary"
-			aria-expanded={ isOpen }
-			// translators: %s: Current post parent.
-			aria-label={ sprintf( __( 'Change parent: %s' ), parentTitle ) }
-			onClick={ onClick }
-		>
-			{ parentTitle }
-		</Button>
-	);
-}
-
-function PostOrderToggle( { isOpen, onClick } ) {
-	const order = useSelect(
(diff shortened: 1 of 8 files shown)


test/e2e/specs/editor/plugins/custom-post-types.spec.js
```diff
@@ -36,16 +36,7 @@ test.describe( 'Test Custom Post Types', () => {
...
-		const pageAttributes = page.getByRole( 'button', {
-			name: 'Page Attributes',
-		} );
-		const isClosed =
-			( await pageAttributes.getAttribute( 'aria-expanded' ) ) ===
-			'false';
-		if ( isClosed ) {
-			await pageAttributes.click();
-		}
+		await page.locator( '.editor-post-parent__panel-toggle' ).click();
```
ntsekouras, line 39: I believe it's better to use `getByRole` if we can.

packages/editor/src/components/page-attributes/parent.js
```diff
@@ -173,6 +173,7 @@ export function PageAttributesParent() {
 			options={ parentOptions }
 			onFilterValueChange={ debounce( handleKeydown, 300 ) }
 			onChange={ handleChange }
+			hideLabelFromVision
```
ntsekouras, line 184: The new designs shared [here](https://github.com/WordPress/gutenberg/issues/60291#issuecomment-2037258428) include some helping text as well. Is this still the case @jameskoster ?

packages/editor/src/components/page-attributes/panel.js
```diff
@@ -1,54 +1,182 @@
...
 
 const PANEL_NAME = 'page-attributes';
 
+function getTitle( post ) {
+	return post?.title?.rendered
+		? decodeEntities( post.title.rendered )
+		: `#${ post.id } (${ __( 'no title' ) })`;
+}
+
+function PostParentToggle( { isOpen, onClick } ) {
```
ntsekouras: Should we move these components inside `order.js and parent.js` respectively?

ntsekouras (commented):
This tests well for me, thanks! I've left a design question (help text in parent controls) and some suggestions to move around some components.

packages/editor/src/components/page-attributes/parent.js
jameskoster: Yeah I think we should include the help text:

Screenshot 2024-05-28 at 11 18 49

jameskoster:
Re-ordering feels a little prominent given it's a seldom-used feature. It's an improvement on trunk but do you think it would be feasible to move it into a modal, accessed from the ellipsis menu?

Screenshot 2024-05-28 at 11 20 29

jorgefilipecosta:
Hi @jameskoster,

> Re-ordering feels a little prominent given it's a seldom-used feature. It's an improvement on trunk but do you think it would be feasible to move it into a modal, accessed from the ellipsis menu?
> 
> Screenshot 2024-05-28 at 11 20 29


This would involve implementing a reorder page action. A little bit like the rename. Are we ok with it being an action e.g.: something that could be done from the dataviews page list? I think it seems fine but I want to confirm.

jameskoster:
I think that's okay, but I'd appreciate a gut-check from @jasmussen.

jasmussen:
The change seems okay, especially for Friday, but we should probably improve it further, it's not clear at all what Order does, and "parent" is ambiguous.

jameskoster:
@jasmussen sorry I should have been more specific. It's a question about where to place access to re-ordering. See the mockup [here](https://github.com/WordPress/gutenberg/pull/61918#issuecomment-2134863600) as an alternative to having it appear as a row in the summary panel.

jasmussen:
Ah, thanks for clarifying.

Yeah it's tricky, it reduces the prominence, but that may be fine? In a block editor context, you'd most likely reorder your menus through a custom menu rather than changing the page order. But it is definitely harder to find. 

If we go with this and get confusion in the beta period, are we okay with a full revert back to a panel, pending better ideas for 6.7?

jameskoster:
Yeah that seems reasonable.

@jorgefilipecosta do you think there's time to try the modal approach in this PR before the beta deadline? Or should we merge this and follow up with the modal separately?

test/e2e/specs/editor/plugins/custom-post-types.spec.js
jorgefilipecosta: In this case, we can't use getByRole because the aria-label is dynamic.

packages/editor/src/components/page-attributes/panel.js
jorgefilipecosta: Done 👍

jorgefilipecosta:
> Yeah that seems reasonable.
> 
> @jorgefilipecosta do you think there's time to try the modal approach in this PR before the beta deadline? Or should we merge this and follow up with the modal separately?

Hi @jameskoster, I think ideally we should merge this component as the action approach for order may need more discussion (it needs some checks in addition to the other actions).
All the feedback on this PR was applied, let me know if I can merge this PR as a temporary step and I will follow up with the move of order to the top menu.

jameskoster:
@jorgefilipecosta okay, that sounds good to me. I pushed a couple of small tweaks.

jameskoster (approved):

Issue #47738, opened 2023-02-03
Add prefix / suffix to Post Date block


druxstr:
## What problem does this address?

The Post Date block outputs a date with no explicit context. Example :  

> 2 Febuary 2023

This could be a "Last modified date" or a "Published on" date. 

There's no way to tell.

## What is your proposed solution?

Add an optional prefix / suffix to Post Date, the same way the Category block does (see PR #40559).

> prefix date suffix

For instance : 

> Published on : 2 Febuary 2023

This would allow to use 2 Post Date blocks on the same post, which may solve #46645 : 

> Published on : 2 Febuary 2023 
> Last Modified : 3 Febuary 2023

iamtakashi:
+1 This will be a nice addition to the block because we wouldn't see the prefix if the post hasn't been modified.

benimub:
+1

jenxi:
+1 This would definitely make it easier to label published/modified dates.

remkus:
Quite stumped why this hasn't been given any prio. This block is basically useless with being able to prefix/suffix it.

carolinan:
@remkus Reading the comments on the linked PR, I am assuming that this is blocked because it is expected to use a block binding instead of a prefix that uses a rich text input.

carolinan:
I mention block bindings because afaik the "bits" proposal has not seen any movement since June 2024.

remkus:
Thanks, @carolinan

westonruter:
https://github.com/WordPress/gutenberg/pull/61920#issuecomment-3771214715:

> I just created a stopgap/prototype plugin to implement this conditional display of the modified date after the published date if it differs: [Post Date Block: Published & Modified](https://github.com/westonruter/post-date-modified). This includes settings for adding a prefix/suffix to both the publish date and modified date. The UI is not suitable for core proposal, but the plugin is a prototype/stopgap which I hope will provide some inspiration. I also published a [blog post](https://weston.ruter.net/2026/01/19/post-date-block-published-modified/) which ends as follows:
> 
[...]

Issue #46645, opened 2022-12-19
Post date block: to display post date and modified date side by side.


iidastudio:
I want to display the post date and modified date side by side.
To do this, I need the option of not displaying the modified date on the front if it has never been modified.
I think we also need to add a class to distinguish each of them.

carolinan:
Do we also need to add a prefix to the block, similar to the post term block?
So that we can easily add "Published (current block text here)" and "Modified (current block text here)"

carolinan:
If there is consensus to change the behavior of the modified date, this PR should solve it https://github.com/WordPress/gutenberg/pull/46839
But we need to decide what we consider a modification, because if we create a post, and spend say half an hour writing the content, WordPress will save a modified date: I assume it is common to not publish the post immediately.

PR #40559, opened 2022-04-22
Post terms: Add prefix & suffix

dsas:
## What?

Allowing an optional prefix and suffix to be added to the Post Categories and Post Tags

## Why?

A common pattern in themes is to display categories and tags prepended with a text. This is the case for all default themes, including TwentyTwentyOne.

image

Fixes #29909

## How?

 * Add prefix and suffix attributes to the post-terms block. 
 * Allow these attributes to be set inside the block editor via a RichText component

## Testing Instructions
 1. Open a Post or page
 2. Insert a Post Tags or Post Categories block
 3. Click the placeholder and start typing
 8. Press Publish
 9. Look at the site on the frontend

## Screenshots or screencast 
image

https://user-images.githubusercontent.com/93301/164788672-b7899105-19d2-4481-ad9b-b5f1153b3e8c.mp4

PR #61508, opened 2024-05-08
Feat: Adds the editing capability to the search title.

vipul0425:
## What?
Fixes issue #60701 
This PR adds the editing capability to the search results title block.

## Why?
- As mentioned in issue #60701 the search result title is not editable, so the user can't edit the text for the search results. 

## How?
- Adds the `richtext `component to make the title editable.

## Testing Instructions
1. Open the Site editor and edit the search template.
2. There you will find the "Search Results Title" Block.
3. You should be able to add and edit the text there.

### Testing Instructions for Keyboard
Nil

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/55375170/5ddc366a-8724-409d-b0d0-2d08b221026d

PR #61920, opened 2024-05-23
Add prefix / suffix to Post Date block

TimBHowe:
## What?

Add prefix / suffix to Post Date block

## Why?

https://github.com/WordPress/gutenberg/issues/47738

The [Post Date](https://wordpress.org/documentation/article/post-date-block/) block doesn't let you add a prefix or suffix. You can manually add a text before the block but you are using the Last Modified Date and it is empty then the text you added will appear with no date output.

This would allow using 2 Post Date blocks on the same post, which may solve https://github.com/WordPress/gutenberg/issues/46645 :

> Published on : 2 Febuary 2023
> Last Modified : 3 Febuary 2023

## How?

This pull request adds an optional prefix / suffix to Post Date, the same way the Category block does (see PR https://github.com/WordPress/gutenberg/pull/40559).

> prefix date suffix

For instance :

> Published on : 2 Febuary 2023

This would allow to use 2 Post Date blocks on the same post, which may solve https://github.com/WordPress/gutenberg/issues/46645 :

Published on : 2 Febuary 2023
Last Modified : 3 Febuary 2023

## Testing Instructions

1. Create a new blog post
2. Add a Post Date block with the "Display last modified date" toggle disabled
3. Enter a prefix. example "Posted On: "
4. Enter a suffix. example "."
2. Add a Post Date block with the "Display last modified date" toggle enabled
5. Enter a prefix. example "Updated On: "
6. Enter a suffix. example "." 
7. Save the post and view the post. As it is a new post it should show the "Posted On: 'Date'." and **Not** the "Updated On: 'Date'." as they are the same.
8. Update the publish date to be in the past to trigger the Published and Last Modified date to be different. This will allow the Post Date block output to pass the follow statement.  https://github.com/dsas/gutenberg/blob/5b95e96a90aac522c6df73349930144b89b53dfa/packages/block-library/src/post-date/index.php#L38
9. Save the post and view the post.  It should show the "Posted On: 'Date'." and  the "Updated On: 'Date'."
10. Party

## Screenshots or screencast 
image
image
image
image
image

### DIFF
--- a/packages/block-library/src/post-date/index.php
+++ b/packages/block-library/src/post-date/index.php
@@ -50,7 +50,7 @@ function render_block_core_post_date( $attributes, $content, $block ) {
 
 	$prefix = '';
 	if ( isset( $attributes['prefix'] ) && $attributes['prefix'] ) {
-		$prefix = '<span class="wp-block-post-date__prefix">' . $attributes['prefix'] . '</span>';
+		$prefix = wp_kses_post( '<span class="wp-block-post-date__prefix">' . $attributes['prefix'] . '</span>' );
 	}
 
 	if ( isset( $attributes['isLink'] ) && $attributes['isLink'] ) {
@@ -59,7 +59,7 @@ function render_block_core_post_date( $attributes, $content, $block ) {
 
 	$suffix = '';
 	if ( isset( $attributes['suffix'] ) && $attributes['suffix'] ) {
-		$suffix = '<span class="wp-block-post-date__suffix">' . $attributes['suffix'] . '</span>';
+		$suffix = wp_kses_post( '<span class="wp-block-post-date__suffix">' . $attributes['suffix'] . '</span>' );
 	}
 
 	return sprintf(
--- a/packages/block-library/src/post-date/index.php
+++ b/packages/block-library/src/post-date/index.php
@@ -48,15 +48,27 @@ function render_block_core_post_date( $attributes, $content, $block ) {
 
 	$wrapper_attributes = get_block_wrapper_attributes( array( 'class' => implode( ' ', $classes ) ) );
 
+	$prefix = '';
+	if ( isset( $attributes['prefix'] ) && $attributes['prefix'] ) {
+		$prefix = '<span class="wp-block-post-date__prefix">' . $attributes['prefix'] . '</span>';
+	}
+
 	if ( isset( $attributes['isLink'] ) && $attributes['isLink'] ) {
 		$formatted_date = sprintf( '<a href="%1s">%2s</a>', get_the_permalink( $post_ID ), $formatted_date );
 	}
 
+	$suffix = '';
+	if ( isset( $attributes['suffix'] ) && $attributes['suffix'] ) {
+		$suffix = '<span class="wp-block-post-date__suffix">' . $attributes['suffix'] . '</span>';
+	}
+
 	return sprintf(
-		'<div %1$s><time datetime="%2$s">%3$s</time></div>',
+		'<div %1$s>%2$s<time datetime="%3$s">%4$s</time>%5$s</div>',
 		$wrapper_attributes,
+		$prefix,
 		$unformatted_date,
-		$formatted_date
+		$formatted_date,
+		$suffix
 	);
 }
 
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -625,7 +625,7 @@ Display the publish date for an entry such as a post or page. ([Source](https://
 -	**Name:** core/post-date
 -	**Category:** theme
 -	**Supports:** color (background, gradients, link, text), interactivity (clientNavigation), spacing (margin, padding), typography (fontSize, lineHeight), ~~html~~
--	**Attributes:** displayType, format, isLink, textAlign
+-	**Attributes:** displayType, format, isLink, prefix, suffix, textAlign
 
 ## Excerpt
 
--- a/packages/block-library/src/post-date/block.json
+++ b/packages/block-library/src/post-date/block.json
@@ -10,8 +10,16 @@
 		"textAlign": {
 			"type": "string"
 		},
+		"prefix": {
+			"type": "string",
+			"default": ""
+		},
 		"format": {
 			"type": "string"
+		},
+		"suffix": {
+			"type": "string",
+			"default": ""
 		},
 		"isLink": {
 			"type": "boolean",
--- a/packages/block-library/src/post-date/edit.js
+++ b/packages/block-library/src/post-date/edit.js
@@ -13,6 +13,7 @@ import {
 	AlignmentControl,
 	BlockControls,
 	InspectorControls,
+	RichText,
 	useBlockProps,
 	__experimentalDateFormatPicker as DateFormatPicker,
 	__experimentalPublishDateTimePicker as PublishDateTimePicker,
@@ -30,7 +31,7 @@ import { DOWN } from '@wordpress/keycodes';
 import { useSelect } from '@wordpress/data';
 
 export default function PostDateEdit( {
-	attributes: { textAlign, format, isLink, displayType },
+	attributes: { textAlign, format, prefix, suffix, isLink, displayType },
 	context: { postId, postType: postTypeSlug, queryId },
 	setAttributes,
 } ) {
@@ -148,7 +149,6 @@ export default function PostDateEdit( {
 						</ToolbarGroup>
 					) }
 			</BlockControls>
-
 			<InspectorControls>
 				<PanelBody title={ __( 'Settings' ) }>
 					<DateFormatPicker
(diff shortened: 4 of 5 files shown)


fabiankaegy (commented):
This works exactly like one would expect and the code looks good to me. 

I personally think this is a very valuable enhancement. If you try to accomplish the same thing today in a block theme you would need to add a paragraph before / after the post date block. But that comes with the downside of the label still getting rendered when the block is not getting rendered. Like in the case when the modified date is the same as the publish date. 

Having said all that I definitely am looking for additional reviews from the core maintainers.

One downside of the prefix / suffix always is internationalization. What may be a prefix in one language is a suffix in another language etc. 

Pinging @WordPress/gutenberg-core for reviews :)

packages/block-library/src/post-date/index.php
```diff
@@ -48,15 +48,27 @@ function render_block_core_post_date( $attributes, $content, $block ) {
...
 	}
 
+	$suffix = '';
+	if ( isset( $attributes['suffix'] ) && $attributes['suffix'] ) {
+		$suffix = '<span class="wp-block-post-date__suffix">' . $attributes['suffix'] . '</span>';
+	}
+
 	return sprintf(
-		'<div %1$s><time datetime="%2$s">%3$s</time></div>',
+		'<div %1$s>%2$s<time datetime="%3$s">%4$s</time>%5$s</div>',
```
cbravobernal, line 66: I think prefix and suffix would need `esc_html`. A contributor user could add scripts there.

cbravobernal:
Maybe we could achieve a similar solution in a future by using block bindings or bits (see https://github.com/WordPress/gutenberg/discussions/39831) instead of adding a prefix and a suffix. 🤔 

Pinging @dmsnell or @SantosGuillamot get some opinions (if they are able to provide it, of course)

youknowriad:
To be honest, this one feels weird to me. Not because it's not valuable but because it sets a precedent. Is this the only block where you would need a prefix/suffix.

There are two alternatives to this:

 - Use a row block (possible today)
 - Consider solving this using "bits" which I know @dmsnell is exploring.

packages/block-library/src/post-date/index.php
fabiankaegy: It would need to be `wo_kses_post` because the rich text string may contain formats such as bold / italics etc.

fabiankaegy:
@youknowriad thats not technically true. There are other blocks that already use this approach: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/post-terms/edit.js

And the example you mention about using a row with a paragraph to achieve this is also sadly not really a good option. Like described above the issue there is that the modified date variation of the post data block doesn't render any output if the modified date is the same as the publish date. In that case the label you add via another block in a row before the post date still renders but the actual information doesn't render. 

We have that same issue with block bindings. And block pattern overrides. Not all data always exists. And currently there are no solutions for possibly empty fields which should get omitted entirely. 

Having said all that this doesn't mean I think this is the perfect approach. I'm not sure how to solve it. But the other solutions all have issues that this implementation here doesn't have.

packages/block-library/src/post-date/index.php
TimBHowe: Applied `wp_kses_post` to the Prefix and Suffix for the Post Date block output https://github.com/WordPress/gutenberg/pull/61920/commits/4d1ca75f6223cee84f742d48d7ddea8fa286ad0b

dmsnell:
> Pinging @dmsnell or @SantosGuillamot get some opinions (if they are able to provide it, of course)

Thanks! *Bits* definitely are going to solve this need, as the current system is overly cumbersome and inconvenient. Working on a Make proposal as we speak, but it won't be here in WordPress 6.6 for sure.

youknowriad:
@fabiankaegy How would you achieve this design?

Screenshot 2024-05-24 at 9 18 09 AM

>  Like described above the issue there is that the modified date variation of the post data block doesn't render any output if the modified date is the same as the publish date.

The solution in this PR is a bit misguided if this is the main argument for us because it achieves the result for a very small set of designs.

I agree that this is a gap in our APIs and code but don't you think something like the Block visibility plugin would be a better approach for this kind of use-cases?

fabiankaegy:
@youknowriad yeah that is exactly why I didn't approve the PR but instead pinged the room. I don't think it is the right solution because it doesn't allow for these kinds of layouts without custom CSS. 

But I did want to raise it because it is a clear gap that I am seeing in many projects. And currently my solution on those projects is essentially adding a prefix via the render block filter in PHP. 

I think this PR showcases the gap in our current API well and I think it is a problem worth looking into :)

jameskoster:
Potentially related: https://github.com/WordPress/gutenberg/pull/61508

Marc-pi:
Hello, i faced this situation this week on some projects, but it's a daily need.
Here are the workarounds i used

1/ About prefix, i use a csss solution, just for the pleasure to test if it still works :-)
indeed we can play with a php filter on the block rendering

```
.datepubli::before {
  content: "Publié le ";
    }

.datemodif::before {
  content: "| Edité le ";
    }

```

2/ Hide Created when Edited is available
- Not everyone will want such a behavior
- I quickly fixed this with the Nick's Block Visilbility plugin
- I did not investigate how the modified date is stored, 
but if it's in a dedicated value,  the case, would be nice to have a checkbox "don't render if empty", like the Phi's MFB plugin is doing

3/ About internationalization, i fixed this with the Block Visiblity plugin also, and another css class (see 1/ via a dedicated class per locale)
- Perhaps we have to wait Phase 4 to know how to improve this...
- For the short term, Nick's plugin does not recognize the localized content at least from Polylang (ex: categories) so we are not able to filter by category (filtering by URL is the workaround solution), but that's another story :-) Will push the need in his plugin forum.


Hope my feedback helps :-)

TimBHowe:
I created an alternative solution to also review.
It seem the options are...

1. Add Prefix/Suffix
2. Always show Post Date blocks when using Last Modified Date
3. Wait for Gutenberg Bits solution.
4. Be there a 4th option....?

Hopefully this help, and let me know if there is anything else I can do to contribute

paaljoachim:
This is the third block I believe that needs the prefix / suffix.
The PR James linked to above: 
Potentially related: https://github.com/WordPress/gutenberg/pull/61508

All of these should be handled in a similar way.

paaljoachim:
@vipul0425 and @t-hamano

t-hamano:
Does anyone know where there is a search/discussion going on about "Bits"? I remember seeing it but couldn't track it down...

dmsnell:
@t-hamano #39831 

Also there was some preliminary exploration in #42015, which was replaced later by #60735 and https://github.com/WordPress/gutenberg/pull/60735

noisysocks:
@dmsnell how confident are you in bits landing eventually? If it's "medium" or higher then I'd say let's hold off on this PR and its alternatives and wait for bits.

dmsnell:
@noisysocks I feel like Bits are inevitable, and while I don't know if I'd say that they will be fully in 6.7, I think some start of them will be. It's a broad-spanning project that essentially began when we introduced Blocks and proverbially kicked out Shortcodes.

What's most likely to appear first is the ability for WordPress to find and render them, since this is all basically possible today. It can be sped up with help to push forward https://github.com/WordPress/wordpress-develop/pull/6395, which is a fix to prevent Core from corrupting the syntax that is now almost definitely going to be the Bits syntax.

paaljoachim:
For various blocks where it would be helpful to add prefix and a suffix could the default Pagination block Previews/Next Page feature be used? Do we need to wait for a totally new Bits syntax? When Bits is ready in perhaps 6-10 months perhaps one can convert over where the Previews/Net Page feature is used into using the new Bits syntax?

dmsnell:
There's now [an official proposal for Bits](https://make.wordpress.org/core/2024/06/08/proposal-bits-as-dynamic-tokens/), where I encourage we all participate to move the project forward.

It's probably not too late to have the server preserve Bits syntax on save for 6.6, because the change necessary to do that is fixing an existing bug. If it's in, then the Post Date block can start exploring with others (such as the work @ellatrix is doing, or @SantosGuillamot with Block Bindings).

westonruter:
I just created a stopgap/prototype plugin to implement this conditional display of the modified date after the published date if it differs: [Post Date Block: Published & Modified](https://github.com/westonruter/post-date-modified). This includes settings for adding a prefix/suffix to both the publish date and modified date. The UI is not suitable for core proposal, but the plugin is a prototype/stopgap which I hope will provide some inspiration. I also published a [blog post](https://weston.ruter.net/2026/01/19/post-date-block-published-modified/) which ends as follows:

> The ideal solution perhaps would [involve Bits](https://github.com/WordPress/gutenberg/pull/61920#issuecomment-2156095190) (now [Shortblocks](https://github.com/WordPress/gutenberg/issues/73929)). Maybe the default view for a Post Date block when editing a block template would be to show the view when the modified date is the same as the published date. But then there could potentially be a popover that could allow configuring how the block should render when the modified date is different. With Shortblocks, these tokens could be inserted inline with the content and moved around inside of a paragraph block. This would avoid the need to supply a suffix/prefix in the detached block settings sidebar.

See https://github.com/WordPress/gutenberg/issues/53099.

PR #61127, opened 2024-04-26
Add content only descriptions in dropdown menus for patterns and templates

kevin940726:
## What?

Based on #61012. Implement the design in https://github.com/WordPress/gutenberg/issues/60834#issuecomment-2063662703.

Add descriptions for content-only blocks inside patterns or templates in the dropdown menu.

## Why?

See #60834.

This is a UI enhancement for content-only blocks. Most of the actions in the original dropdown aren't needed for content-only blocks. We remove them and provide descriptions and links to edit the entities instead.

## How?

Hide original actions and inject descriptions for content-only blocks. The wording and the design are still TBD and can be tweaked.

## Testing Instructions

**For patterns**:
- Create a synced pattern with overridable blocks.
- Insert the pattern into a post.
- Use list view to select the overridable blocks and open the options dropdown.
- Expect to see the description and the "Edit pattern" menu item.
- Clicking on "Edit pattern" should navigate to the pattern editor.
- The same flow should also work in the site editor.

**For templates**:
- Enable twentytwentyfour theme.
- Go to Site editor -> Pages -> Privacy Policy.
- Open list view and select the "Featured Image" block.
- Open the Options dropdown and expect to see the description and the "Edit template" menu item.
- Clicking on it should navigate to the template editor.

## Screenshots or screencast 

**Patterns**:
pattern

**Templates**:
template

PR #61922, opened 2024-05-24
Remove lock icons from Content blocks inner blocks when editing a page in the site editor

talldan:
## What?
It was mentioned in https://github.com/WordPress/gutenberg/pull/61127#issuecomment-2127615997 that there was a regression around how lock icons are shown when editing a page in the site editor.

Previously, blocks within the 'Content' block were shown without the lock icon as they can be freely edited, but after #61127 they show a lock icon.

## Why?
The regression seems to have happened because of a change to the `canMoveBlock` selector, which is called as part of the logic for displaying the lock icon. The change considered any block with a parent that has `blockEditingMode === 'contentOnly'` as not moveable, but that's not the case.

My understanding is there are some differences between `blockEditingMode === 'contentOnly'` and `templateLock === 'contentOnly'` that can be hard to understand. `blockEditingMode` doesn't affect inner blocks, while `templateLock` does.

The selector is probably missing some logic for `templateLock === 'contentOnly'`, but I'd consider that for a different PR.

## How?
Revert the changes to the selector. This may bring back some other issues that the change was intended to solve.

## Testing Instructions

In addition to the steps below, it's also worth smoke testing other editing modes (editing template, template parts, posts etc.)

#### Page editing
1. Edit a page
2. Open List View

Expected: The lock icons shouldn't be displayed on the inner blocks of the 'Content' block. The templated blocks (Title, Featured Image, Content) are also not expected to show a lock icon since #61127, the block options dropdown menu explains why those blocks can't be moved.

#### Synced patterns with overrides
1. Insert a synced pattern that has overrides
2. Open List View

Expected: The lock icons are not displayed on overriden block with a synced pattern. The block options dropdown menu explains why those blocks can't be moved.

#### contentOnly locked groups
1. Insert a group that has inner blocks (easiest way is to insert an unsynced pattern that has a group as the wrapping block)
2. Switch to code editor mode and add the `"templateLock":"contentOnly"` attribute to the group block
3. Switch back to visual mode and open list view

Expected: The lock icons are not displayed on the inner blocks of the group. The block options dropdown menu explains why those blocks can't be moved.

#### Page editing in the site editor
Screenshot 2024-05-24 at 12 38 48 pm

#### Synced patterns with overrides
Screenshot 2024-05-24 at 12 42 40 pm

#### Groups with `templateLock === 'contentOnly'`
Screenshot 2024-05-24 at 12 41 20 pm

### DIFF
--- a/packages/block-editor/src/store/selectors.js
+++ b/packages/block-editor/src/store/selectors.js
@@ -1759,16 +1759,13 @@ export function canMoveBlock( state, clientId, rootClientId = null ) {
 	if ( attributes === null ) {
 		return true;
 	}
-	if ( getBlockEditingMode( state, rootClientId ) !== 'default' ) {
-		return false;
-	}
 	if ( attributes.lock?.move !== undefined ) {
 		return ! attributes.lock.move;
 	}
 	if ( getTemplateLock( state, rootClientId ) === 'all' ) {
 		return false;
 	}
-	return true;
+	return getBlockEditingMode( state, rootClientId ) !== 'disabled';
 }
 
 /**


talldan:
Seems this breaks the "Zoom Out › Clicking on inserter while on zoom-out should open the patterns tab on the inserter" test, which is strange. Will look into it.

talldan:
> Seems this breaks the "Zoom Out › Clicking on inserter while on zoom-out should open the patterns tab on the inserter" test, which is strange. Will look into it.

This test fails on `trunk` for me locally, so I don't think it's related to this PR.

I think the test is very flaky (https://github.com/WordPress/gutenberg/issues/61806) 🤷

youknowriad:
Before https://github.com/WordPress/gutenberg/pull/61127 top level blocks were considered locked (which they are) in template-locked mode but not in this PR.

talldan:
> Before https://github.com/WordPress/gutenberg/pull/61127 top level blocks were considered locked (which they are) in template-locked mode but not in this PR.

@youknowriad Yes, that's by design, there was a design decision to remove the locked icon for content only locked blocks.

jameskoster:
This fix is working for me 👍

youknowriad (approved):
ok LGTM 👍 Definitely worth an e2e test as a follow-up or something.

youknowriad:
Actually, in template-lock mode, these top level blocks are entirely locked and not just "content only". So I think they should have the lock icon.

talldan:
> Actually, in template-lock mode, these top level blocks are entirely locked and not just "content only". So I think they should have the lock icon.

I might be misunderstanding what you mean by 'template-locked' mode. I'm testing by editing a page. For me I can still change the title and featured image content. The Header/Footer I can't, but they have an 'Edit' button.

youknowriad:
Ok, so I did more tests, it seems that "non post blocks" like random paragraphs present in the template are hidden entirely from the list view, rather than kept with the lock icon. Isn't that a bit confusing?

youknowriad:
Maybe it's out of scope of this PR though, so I'm fine if this is treated separately but I wonder if this was discussed and something that was agreed upon. Don't want to block this PR.

talldan:
Yeah, I think it was a design decision to try to simplify the hierarchy of blocks. If all the various layout blocks that make up a template are included it becomes quite complicated. I think the idea started from contentOnly mode, and it's been that way for page editing going back to the original work that @noisysocks was involved in when `useBlockEditingMode` was introduced.

Synced patterns now also work the same way, hiding the completely locked blocks and only showing what's editable in List View.

talldan:
The failing test is now skipped in trunk - https://github.com/WordPress/gutenberg/pull/61925.

So I'll merge this even though it's failing.

PR #61929, opened 2024-05-24
Try: Enable script debug for the test env

Mamaduka:
## What?
🚧  DO NOT MERGE! 🚧

PR enables `SCRIPT_DEBUG` for the testing environment to catch any warnings triggered by new JSX runtime. See: #61917.


Mamaduka:
@youknowriad, @jsnajdr, we should probably use the same testing environment when enabling `StrictMode`. It should help us catch any possible issues with double-effect runs.

PR #61692, opened 2024-05-15
React: Upgrade to the new JSX transform 

youknowriad:
Alternative to #34221 

## What

React 17 introduced a new JSX runtime but we are still using the old JSX build process. As part of the React 19 upgrade, the React team is deprecating the old JSX so this is the second attempt to move away from it.

The initial PR #34221 tried to include the runtime into the "@wordpress/element" package but since, we've moved away from using @wordpress/element as the JSX target of our builds, which means we can take a simpler approach and just "externalize" the react/jsx-runtime script into a new WP script handle and global variable and use it while building our scripts.

We should be able to deprecate our package `@wordpress/babel-plugin-import-jsx-pragma`.

To learn more about the new JSX transform, [read this post](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).

This definitely requires a dev note to nudge plugin authors to also switch to the upgraded "@wordpress/scripts".

Todo:

 - [ ] Deprecate @wordpress/babel-plugin-import-jsx-pragma
 - [x] Document the breaking changes in our dev tools packages.

Anything I'm missing?

## Testing instructions.

The e2e tests got us covered here in general.

PR #61486, opened 2024-05-08
Build: Use globalThis over process.env and enable TS lib checking

sirreal:
## What?

There have been some problems with types in some packages. This resulted in build errors and in producing types that were not valid. This was observed in several other PRs recently.

This PR does two things:
- Remove all `skipLibCheck` from tsconfig.
- Update usage of custom `process.env.NAME` usage in the application to `globalThis.NAME`. This has the benefit of being safer in third party code and valid in the type system.

There are a few problems with using `process.env`, although it's common to use some names for historic reasons - primarily `process.env.NODE_ENV`.

- It only exists in Node.js. It does not exist in the browser. This means that additional checks must be added in application code to make sure that it doesn't error. Using `process.env` in general is error prone.
- [`process.env` in Node.js is unusual. It's a key-value structure where _the values are always strings_.](https://nodejs.org/docs/latest-v20.x/api/process.html#processenv) Gutenberg used non-string values which was wrong in environments where process.env existed, but also did compile-time replacement sometimes. This makes everything confusing.
- The inconsistent process.env values caused problems in the type system. They conflicted with the Node.js types that _correctly assert that process.env values must be `strings`_.

Instead of using the problematic `process.env` "object" for these values, we change to use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis). This is the global object that works natively in both browsers and Node.js. It is an object and can be used as such, so assigning arbitrary properties is straightforward.

One big benefit is that `globalThis` is much easier to use, `globalThis.nonExistentProperty` is `undefined` instead of a `ReferenceError` like when trying to access `process.env.nonExistentProperty` in the browser.

This eliminates the type issues around using `process.env` and is generally more coherent. The build tooling and code using WordPress and Gutenberg variables around `process.env` has been updated in this PR. Documentation has also been updated where I found it.

Some eslint rules have also been added, changed, or removed:

- `@wordpress/wp-global-usage` replaces `@wordpress/is-gutenberg-plugin`. It applies the same logic to all the globals that Gutenberg uses - namely that they should be used in simple conditional tests to help with dead code elimination in compiled packages.
- `@wordpress/wp-process-env` is a new rule that makes the usage of `process.env` WordPress/Gutenberg variables a violation. It includes a fixer, so e.b. `process.env.SCRIPT_DEBUG` becomes `globalThis.SCRIPT_DEBUG`.
- `@wordpress/gutenberg-phase` rule is removed - this variable has not been used for some time.

[...]

PR #61930, opened 2024-05-24
Packages: Increase the minimum required Node.js version to v18.12.0

gziolo:
## What?

Node.js v18.12.0 was the first long-term supported version introduced in v18.x line (https://nodejs.org/en/blog/release/v18.12.0). It is now the lowest version that is listed as Maintenance LTS release at https://nodejs.org/en/about/previous-releases. The same version is bundled with npm 8.19.2. This PR tries to unify the minimum required Node.js version defined in all npm packages published to npm.

## Why?

We are approaching another major WordPress release - 6.6.0. It's a good opportunity to ship all the breaking changes batched together with some other changes that landed recently:
- https://github.com/WordPress/gutenberg/pull/61486
- https://github.com/WordPress/gutenberg/pull/61692

While Gutenberg and WordPress core are more strict in defining the minimum required Node.js and npm versions:

```json
"engines": {
	"node": ">=20.10.0",
	"npm": ">=10.2.3"
},
```

I propose we also allow the 18.x line for WordPress packages, which will be actively maintained until the middle of 2025. This will give more time to folks who might still be using Node.js 18 in their projects.

## How?

I ensured that all non-private packages have the minimum version defined with the following:

```json
"engines": {
	"node": ">=18.12.0",
	"npm": ">=8.19.2"
},
```

I also updated all corresponding documentation and add changelog entries.

The only exception is `@wordpress/create-block,` which aligns more closely with the versions that WordPress and Gutenberg use, as decided separately some time ago.

## Testing Instructions

There isn't that much to test because `engines` works mostly as a hint for Node.js, unless the strict check is explicitly enabled in `.npmrc`. More in https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53072,7 +53072,8 @@
 				"@wordpress/i18n": "file:../i18n"
 			},
 			"engines": {
-				"node": ">=12"
+				"node": ">=18.12.0",
+				"npm": ">=8.19.2"
 			}
 		},
 		"packages/annotations": {
@@ -53088,7 +53089,8 @@
 				"uuid": "^9.0.1"
 			},
 			"engines": {
-				"node": ">=12"
+				"node": ">=18.12.0",
+				"npm": ">=8.19.2"
 			},
 			"peerDependencies": {
 				"react": "^18.0.0"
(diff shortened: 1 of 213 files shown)


sirreal (approved):
Yes please!

I've reviewed a sample of the changes and they look good.

youknowriad (approved):
LGTM 👍 

Thanks for being on top of this.

Issue #61846, opened 2024-05-22
Layout Styles: Block list appender causes inconsistent application of layout margin styles


aaronrobertshaw:
### Description

When a block type has margins, defined either through theme.json or global styles, they can be overridden by layout styles e.g. the constrained layout which removes the start and end margins on the first and last child elements respectively.

When the block list appender is inserted within a container with such a layout, the style removing the last child's margin now targets the appender instead of the last block causing a layout shift

### Step-by-step reproduction instructions

1. Navigate to Appearance > Editor > Global Styles > Blocks > Cover
2. Apply vertical margins to Cover blocks
3. Add a Cover block within a Group block to a page or post
4. Note the Group block's default constrained layout removes the Cover block's margins
5. Select the Group block and see the bottom margin suddenly applied 

### Screenshots, screen recording, code snippet

https://github.com/WordPress/gutenberg/assets/60436221/f5f83822-16da-4b10-9c8b-debe11328866

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

PR #61638, opened 2024-05-14
Styles: try wrapping with :root to fix reset styles

ellatrix:
## What?

In order to apply global styles _after_ any theme reset styles, e.g. `ol, ul { ... }`, increase the specificity _slightly_, while _keeping_ CSS overall specificity levelling and specificity decreases.

To do: figure out where else exactly we need `:root`, I'm not super familiar with this code.

_**Note: Block style variations and their selectors will be handled via a separate follow-up as they haven't yet had their specificity reduction merged in Gutenberg (see https://github.com/WordPress/gutenberg/pull/61032).**_

## Why?

To make sure reset styles not wrapped in `:where` do not break anything.

Why `:root` and not `html` or `body`? Because `:root` has the same specificity of attributes and would catch all kinds of reset styles including those with attributes ([example](https://github.com/necolas/normalize.css/blob/fc091cce1534909334c1911709a39c22d406977b/normalize.css)).

## How?

Wrapping all our specificity normalised/levelled rules in `:root`, so that they run after these reset styles.

## Testing Instructions

The only unit tests we have are covering the generation of styles rather than their application, so we'll have to rely on some manual testing or adding new tests. Any help here is appreciated! :)

#### Manual Testing

- Activate emptytheme
- Add all blocks to a post or page in both editors, ensure they display correctly and block library styles are applied
- Switch to the switch editor and apply global styles for each block, check these all take affect both in the editor and frontend
- Test various classic themes making sure the blocks get all the correct styles on the frontend or in the editors

PR #61932, opened 2024-05-24
Paragraph: Add e2e tests for computed block styles

talldan:
## What?
One of the big gaps in our automated test coverage is that we lack assertions that the generated styles for blocks are correct.

The block editor's styling system is a layered complex system with global element styles, global styles for block types, block instance styles, block style variations, and probably more, all of which have an order of precedence. It's often too easy to cause a regression as this as the generated CSS can depend upon specificity for the style rules, as well as the ordering for the CSS. 

At the moment this PR only adds a test for the paragraph block, but once the approach is agreed upon it should be easier to replicate these tests across different block types, and also add some combinations of blocks for layouts and section styling.

## How?
The tests are structured like this:
- One file for a block (`test.describe`)
- Within that, one `test` for each type of style (e.g. color, typography, spacing)
- Within that, one `test.step` for each layer of style starting from the top (usually global element styles), and working down to the bottom (usually block instance styles) and for each step checking that it overrides the previous layer.

I've tried to keep the assertions as simple as possible, they just use `toHaveCss` which runs `getComputedStyle` in the background.

A new `updateGlobalStyles` util is added, so the test doesn't have to tinker with the UI. I think there need to be separate e2e tests for the global styles UI, as that seems like something else where there's little coverage.


packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
```diff
@@ -0,0 +1,70 @@
...
+	} );
+}
+
+/**
+ * Deletes all post revisions using the REST API.
+ *
+ * @param {}              this     RequestUtils.
+ * @param {string|number} parentId Post attributes.
+ */
+async function getThemeGlobalStylesRevisions(
```
talldan, line 66: I moved this from another file, but it seems the description is wrong, so I'll update it 👍

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
```diff
@@ -0,0 +1,70 @@
...
+		if ( globalStylesURL ) {
+			themeGlobalStylesId = globalStylesURL?.split(
+				'rest_route=/wp/v2/global-styles/'
+			)[ 1 ];
+		}
+	}
+	return themeGlobalStylesId;
+}
+
+async function updateGlobalStyles( this: RequestUtils, styles: Object ) {
```
talldan, line 46: I need to add doc blocks.

Hopefully I don't have to add types for the global styles object 🙏 

If there's a type somewhere else that can be imported that would be nice.

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
```diff
@@ -0,0 +1,402 @@
...
+			} );
+
+			const block = editor.canvas.getByRole( 'document', {
+				name: 'Block: Paragraph',
+			} );
+
+			// fontStyle, fontWeight, fontSize.
+			await expect( block ).toHaveCSS(
+				'font',
+				'100 11px / 12.1px Times'
```
talldan: It seems like the test is failing, probably because on linux it evaluates a different font name (Times vs Times New Roman).

I'll break this out into multiple assertions for each property to fix that.

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
ramonjd: Probably one of my copy-paste doozies, thanks for updating.

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
ramonjd: I checked in core-data, and it's mostly using `any` or `Record< string, Object >` as return values for selectors 😄 . 

I did create an interface for [GlobalStylesRevisions](https://github.com/WordPress/gutenberg/blob/8eb63ddb51525164860e785900d7d86544f37afc/packages/core-data/src/entity-types/global-styles-revision.ts), I guess it would also be helpful to have one for global styles.

This comment is zero helpful I know.

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
```diff
@@ -0,0 +1,402 @@
...
+					content: 'Paragraph 3',
+				},
+			} );
+
+			const paragraphs = editor.canvas.getByRole( 'document', {
+				name: 'Block: Paragraph',
+			} );
+
+			// Left/right margin is set to auto by alignments, there's no reliable way to test this.
+			// First block will have its top margin set to zero by the layout.
```
aaronrobertshaw, line 325: We could wrap the paragraph in a group block and change the group's layout (to `default`) to allow the paragraph's left/right margins to take effect.

Doing this manually in the editor is just toggling off the "Inner blocks use content width" option. For these tests it could just be adding  `layout: { type: 'default' }` to the group block's attribtues.

For example:
```html
<!-- wp:group {"layout":{"type":"default"}} -->
```

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
```diff
@@ -0,0 +1,402 @@
...
+				'margin-bottom',
+				'0px'
+			);
+			await expect( paragraphs.last() ).toHaveCSS(
+				'padding',
+				'3px 4px 5px 6px'
+			);
+		} );
+
+		await test.step( 'Block instance styles', async () => {
```
aaronrobertshaw, line 419: For the paragraph block, it might be good to test the padding against a couple of paragraph block instances. That way we can confirm:

- When a background is set on the instance, the padding is overridden as per the `p.has-background` style
- When the instance has background and padding styles, the instance padding styles take precedence

This test would then also cover the [change](https://github.com/WordPress/gutenberg/pull/61638/files#diff-06e5cc378a588f30c09b6d62f89ad8abc373044c2acdf93e20e9ecf3e619aaebL41) to `p.has-background` specificity in https://github.com/WordPress/gutenberg/pull/61638.

aaronrobertshaw (commented):
Nice work @talldan 👍 

The general approach LGTM. It's easy to follow and looks easy to spin up new tests for other blocks or edge cases, which should help increase adoption and provide greater confidence in our style system. 

I can definitely see these sorts of tests catching a few regressions if we can get sufficient coverage.

Approach aside, there are a couple of tweaks I think we can make for the paragraph block specific tests which I've left as inline comments.

Thanks for getting the ball rolling on style tests!

aaronrobertshaw:
In case it helps, below is a quick diff relating to the suggestions I left in my review above:

<details>
<summary>Suggested tweaks</summary>

```diff
diff --git a/test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js b/test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
index 29bfa5dadc..512d38916f 100644
--- a/test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
+++ b/test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
@@ -365,38 +365,85 @@ test.describe( 'Paragraph computed styles', () => {
 		} );
 
 		await test.step( 'Block instance styles', async () => {
-			const instanceStyles = {
-				spacing: {
-					margin: {
-						top: '11px',
-						bottom: '12px',
-					},
-					padding: {
-						top: '13px',
-						right: '14px',
-						bottom: '15px',
-						left: '16px',
-					},
-				},
+			const marginStyles = {
+				top: '11px',
+				right: '12px',
+				bottom: '17px',
+				left: '18px',
+			};
+			const paddingStyles = {
+				top: '13px',
+				right: '14px',
+				bottom: '15px',
+				left: '16px',
 			};
 
 			await admin.createNewPost();
+
+			// Paragraph is wrapped in a group block with a default layout so
+			// that left/right margins will not be overridden by the layout.
+			await editor.insertBlock( {
+				name: 'core/group',
+				layout: { type: 'default' },
+				innerBlocks: [
+					{
+						name: 'core/paragraph',
+						attributes: {
+							content:
+								'Hello <a href="https://wordpress.org">world</a>',
+							backgroundColor: 'primary',
+							style: {
+								spacing: { margin: marginStyles },
+							},
+						},
+					},
+				],
+			} );
 			await editor.insertBlock( {
 				name: 'core/paragraph',
 				attributes: {
 					content: 'Hello <a href="https://wordpress.org">world</a>',
-					style: instanceStyles,
+					style: {
+						color: { background: '#2d2d2d' },
+						spacing: {
+							padding: paddingStyles,
+						},
+					},
 				},
 			} );
 
-			const block = editor.canvas.getByRole( 'document', {
+			const paragraphs = editor.canvas.getByRole( 'document', {
 				name: 'Block: Paragraph',
 			} );
 
-			// Left/right margin is set to auto by alignments, there's no reliable way to test this.
-			await expect( block ).toHaveCSS( 'margin-top', '11px' );
-			await expect( block ).toHaveCSS( 'margin-bottom', '12px' );
-			await expect( block ).toHaveCSS( 'padding', '13px 14px 15px 16px' );
+			await expect( paragraphs.first() ).toHaveCSS(
+				'margin-top',
+				'11px'
+			);
+			await expect( paragraphs.first() ).toHaveCSS(
+				'margin-right',
+				'12px'
+			);
+			await expect( paragraphs.first() ).toHaveCSS(
+				'margin-bottom',
+				'17px'
+			);
+			await expect( paragraphs.first() ).toHaveCSS(
+				'margin-left',
+				'18px'
+			);
+			// p.has-background padding is 1.25em 2.375. At 16px font size padding should be 20px 38px.
+			await expect( paragraphs.first() ).toHaveCSS(
+				'padding',
+				'20px 38px'
+			);
+
+			// Block instance padding styles should override the default padding
+			// when background is applied.
+			await expect( paragraphs.nth( 1 ) ).toHaveCSS(
+				'padding',
+				'13px 14px 15px 16px'
+			);
 		} );
 	} );
 } );

```

</details>

I've tested these against this branch and everything passes. After cherry picking this PR's commits to #61638, they also failed at the expected styles.

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
```diff
@@ -0,0 +1,501 @@
...
+			await expect( nestedParagraph2 ).toHaveCSS(
+				'margin',
+				'1px 2px 3px 4px'
+			);
+			// Note: This is different to what I reproduce locally.
+			// Bottom margin should be 0px for the last block.
+			await expect( nestedParagraph3 ).toHaveCSS(
+				'margin',
+				'1px 2px 3px 4px'
+			);
```
talldan: This is the only oddity I'm seeing, when I tested this manually the bottom margin was `0px` for this paragraph 🤔

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
talldan: Good idea, there was already one assertion for the first case in the colors tests, but I moved them to the 'dimensions' section and added an extra assertion for the instance padding overriding the `has-background` padding.

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
talldan: Nice idea. I've added extra blocks and assertions for that.

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
talldan: Thanks, I didn't know those existed. Could be useful for the utils if we can import the types from there. I'd feel better about adding a global styles type in that location.

test/e2e/specs/editor/block-computed-styles/paragraph-styles.spec.js
talldan: I can reproduce this now. It seems like the `last-child` style doesn't get applied straight away. Steps to reproduce:
1. Set some non-zero random margin values for the paragraph block type in global styles
2. Create a group with two paragraphs
3. Select the group and toggle 'Inner blocks use content width'
4. Without selecting it, inspect the styles of the last paragraph, observe that it has the margin you specified in global styles
5. Now select and deselect the last paragraph
6. Inspect the styles again, observe that there's now a new style for the `last-child` that sets bottom margin to zero.

talldan: Actually, I think the issue is that the group block is selected, and that means a `block-list-appender` is inserted at the end of the group block. The appender is the `last-child` so it receives the `:last-child` margin and not the last paragraph.

It's very noticeable when the global styles bottom margin is set to a huge value.

I'll make a separate issue for it. In the editor, the `:last-child` style should I guess only be applied to blocks.

For the test, I'll make it deselect blocks before making the assertion.

edit: I see @aaronrobertshaw beat me to it (https://github.com/WordPress/gutenberg/issues/61846) 😄

talldan:
E2E tests now passing. I'm ignoring the 'Compressed Size' failure, which isn't relevant to this PR anyway. 😄 

The big omission from this PR is testing the frontend styles. I'd like to spend a bit more time try to find a nice way to do that.

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
```diff
@@ -0,0 +1,83 @@
...
+		params: {
+			per_page: 100,
+		},
+	} );
+}
+
+export {
+	getCurrentThemeGlobalStylesPostId,
+	getThemeGlobalStylesRevisions,
+	updateGlobalStyles,
```
Mamaduka, line 82: @talldan, do you think these utils will be helpful in other tests? The best practice is to avoid overloading global utils in favor of local utils - https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/e2e/README.md#dont-overload-test-utils-inline-simple-utils.

talldan: Two of them were already present, I just moved them to a different file.

`updateGlobalStyles` is the only new one, which I think will be very useful, as I plan to add some more of these tests for different blocks and combinations of blocks.

It might also be useful for clearing global styles changes as part of the test cleanup, though perhaps we also want to clear revisions when doing that.

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
```diff
@@ -0,0 +1,83 @@
...
+	let themeGlobalStylesId: string = '';
+	if ( themes && themes.length ) {
+		const currentTheme: ThemeItem | undefined = themes.find(
+			( { status } ) => status === 'active'
+		);
+
+		const globalStylesURL =
+			currentTheme?._links?.[ 'wp:user-global-styles' ]?.[ 0 ]?.href;
+		if ( globalStylesURL ) {
+			themeGlobalStylesId = globalStylesURL?.split(
```
kevin940726, line 31: Nit: Should we throw an error to indicate there's no post id found? Maybe `undefined` or `null` would be better than an empty string too.

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return themeGlobalStylesId;
+}
+
+/**
+ * Update the user's global styles. Note that this overwrites and previously
+ * set styles. Styles that you want to keep should be merged in manually.
+ *
+ * @param {RequestUtils} this   RequestUtils.
+ * @param {Object}       styles A theme.json compatible object.
```
kevin940726, line 44: Nit: The jsdoc type isn't needed if we're using typescript, it's annoying to keep them in sync anyway 😅.

kevin940726 (commented):
Looks good! There's a test failure though. Any clue?

packages/e2e-test-utils-playwright/src/request-utils/global-styles.ts
ramonjd: `updateGlobalStyles` is potentially helpful for testing revisions, see: https://github.com/WordPress/gutenberg/blob/add/e2e-tests-for-block-styles/test/e2e/specs/site-editor/user-global-styles-revisions.spec.js#L304 👍🏻

PR #61936, opened 2024-05-24
More block: Fix React warning when adding custom text

Mamaduka:
## What?
PR fixes a React warning triggered by the "More" block when updating the placeholder text.

```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen.
```

## Why?
This can happen when the value changes from `undefined` to other primitive values like string.

## How?
Set the default value to an empty string instead of `undefined`.

## Testing Instructions
1. Open a post or page.
2. Insert the "More" block.
3. Change the placeholder "Read More" text to something else.
4. Confirm that no React warnings are logged in the console.

Same.

## Screenshots or screencast 
CleanShot 2024-05-24 at 14 11 17

### DIFF
--- a/packages/block-library/src/more/edit.js
+++ b/packages/block-library/src/more/edit.js
@@ -10,14 +10,13 @@ import { getDefaultBlockName, createBlock } from '@wordpress/blocks';
 const DEFAULT_TEXT = __( 'Read more' );
 
 export default function MoreEdit( {
-	attributes: { customText = '', noTeaser },
+	attributes: { customText, noTeaser },
 	insertBlocksAfter,
 	setAttributes,
 } ) {
 	const onChangeInput = ( event ) => {
 		setAttributes( {
-			customText:
-				event.target.value !== '' ? event.target.value : undefined,
+			customText: event.target.value,
 		} );
 	};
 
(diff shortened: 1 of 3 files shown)


packages/block-library/src/more/edit.js
```diff
@@ -10,7 +10,7 @@ import { getDefaultBlockName, createBlock } from '@wordpress/blocks';
 const DEFAULT_TEXT = __( 'Read more' );
 
 export default function MoreEdit( {
-	attributes: { customText, noTeaser },
+	attributes: { customText = '', noTeaser },
```
ellatrix: Should we add a default to the attribute definition? I think that's what a lot of blocks do

Mamaduka: Sometimes, attribute updates require deprecations. I think undefined values don't get serialized, while empty does.

I can double-check, but adding deprecation for a simple fix like this is probably not worth it. What do you think?

Mamaduka: @ellatrix, I was wrong. It looks like there's no need for deprecation, at least for the current case.

ellatrix (approved):

PR #61939, opened 2024-05-24
Clean up packages build-types when cleaning types

sirreal:
## What?

Stale generated types can be left sitting around the repo which can cause build issues. Gutenberg recommends using the `clean:package-types` command to clean types and fix this problem, but the generated declaration files remain.

This can cause issues with builds that are resolved by removing the stale declaration files (put in the `build-types` directories) and getting a fresh build.

## Why?

It can break the types builds.

## How?

By removing all the stale generated files.

## Testing Instructions

After running

```sh
npm run clean:package-types
```

No packages should have `build-types` directories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -273,7 +273,7 @@
 		"build:packages": "npm run --silent build:package-types && node ./bin/packages/build.js",
 		"build:plugin-zip": "bash ./bin/build-plugin-zip.sh",
 		"clean:package-types": "tsc --build --clean; rimraf \"./packages/*/build-types\"",
-		"clean:packages": "rimraf \"./packages/*/@(build|build-module|build-style)\"",
+		"clean:packages": "rimraf \"./packages/*/@(build|build-module|build-style|build-types)\"",
 		"dev": "npm run build:packages && concurrently \"wp-scripts start\" \"npm run dev:packages\"",
 		"dev:packages": "concurrently \"node ./bin/packages/watch.js\" \"tsc --build --watch\"",
 		"distclean": "rimraf node_modules packages/*/node_modules",
--- a/package.json
+++ b/package.json
@@ -272,7 +272,7 @@
 		"prebuild:packages": "npm run clean:packages && lerna run build",
 		"build:packages": "npm run --silent build:package-types && node ./bin/packages/build.js",
 		"build:plugin-zip": "bash ./bin/build-plugin-zip.sh",
-		"clean:package-types": "tsc --build --clean",
+		"clean:package-types": "tsc --build --clean; rm -rf packages/*/build-types",
 		"clean:packages": "rimraf \"./packages/*/@(build|build-module|build-style)\"",
 		"dev": "npm run build:packages && concurrently \"wp-scripts start\" \"npm run dev:packages\"",
 		"dev:packages": "concurrently \"node ./bin/packages/watch.js\" \"tsc --build --watch\"",
--- a/package.json
+++ b/package.json
@@ -272,7 +272,7 @@
 		"prebuild:packages": "npm run clean:packages && lerna run build",
 		"build:packages": "npm run --silent build:package-types && node ./bin/packages/build.js",
 		"build:plugin-zip": "bash ./bin/build-plugin-zip.sh",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -272,7 +272,7 @@
 		"prebuild:packages": "npm run clean:packages && lerna run build",
 		"build:packages": "npm run --silent build:package-types && node ./bin/packages/build.js",
 		"build:plugin-zip": "bash ./bin/build-plugin-zip.sh",
-		"clean:package-types": "tsc --build --clean",
+		"clean:package-types": "tsc --build --clean; rm -rf packages/*/build-types",
```
ellatrix: Does the order matter? Just asking because you suggested running `rm -rf packages/*/build-types` before this command to me

ellatrix (approved):

package.json
sirreal: No, it shouldn't matter.

t-hamano: `rm -rf` does not work on Windows OS. Probably not many contributors are building Gutenberg on a Windows host OS, but is there a way to change the command to work on Windows/Mac/Linux all?

t-hamano: I noticed that `rimraf` is already added as a dependency. We can use this library. This library works in a cross-platform.

sirreal: Sure, I used `rm -rf` because I noticed it already in another script.

package.json
```diff
@@ -272,8 +272,8 @@
 		"prebuild:packages": "npm run clean:packages && lerna run build",
 		"build:packages": "npm run --silent build:package-types && node ./bin/packages/build.js",
 		"build:plugin-zip": "bash ./bin/build-plugin-zip.sh",
-		"clean:package-types": "tsc --build --clean",
-		"clean:packages": "rimraf \"./packages/*/@(build|build-module|build-style)\"",
+		"clean:package-types": "tsc --build --clean; rimraf \"./packages/*/build-types\"",
```
t-hamano:
```suggestion
		"clean:package-types": "tsc --build --clean && rimraf \"./packages/*/build-types\"",
```

This format caused an error on Windows. Perhaps `&&` is better?

image

t-hamano (approved):
Both scripts `clean:package-types` and `clean:packages` seem to work on Windows OS if we apply the changes in [this comment](https://github.com/WordPress/gutenberg/pull/61939/files#r1613754104).

However, it may be an issue with my environment, but the latest Gutenberg seems to cause the build to fail on the Windows host OS 🤔 However, this problem is limited and Windows developers should be fine as long as they use [WSL](https://learn.microsoft.com/en-us/windows/wsl/about).

image

package.json
sirreal: Would it work on Windows if there were space before the `;`?

```suggestion
		"clean:package-types": "tsc --build --clean ; rimraf \"./packages/*/build-types\"",
```

t-hamano: No 😅
image

package.json
```diff
@@ -272,8 +272,8 @@
 		"prebuild:packages": "npm run clean:packages && lerna run build",
 		"build:packages": "npm run --silent build:package-types && node ./bin/packages/build.js",
 		"build:plugin-zip": "bash ./bin/build-plugin-zip.sh",
-		"clean:package-types": "tsc --build --clean",
-		"clean:packages": "rimraf \"./packages/*/@(build|build-module|build-style)\"",
+		"clean:package-types": "tsc --build --clean ; rimraf \"./packages/*/build-types\"",
```
sirreal:
```suggestion
		"clean:package-types": "tsc --build --clean && rimraf \"./packages/*/build-types\"",
```

t-hamano:
@sirreal I am concerned that this PR may have caused two problems.

- ["Compressed Size" github action](https://github.com/WordPress/gutenberg/actions/workflows/bundle-size.yml) fails.
- An error will occur if I run `npm run dev` or `npm run build` consecutively locally. I have to run `clean:package-types` every time.

image

sirreal:
This should not have added `build-types/` to the package clean script: https://github.com/WordPress/gutenberg/pull/62008

PR #61940, opened 2024-05-24
Add postcss-modules to support local selectors

jsnajdr:
Adds support for marking selectors or keyframes as `:local`, and preventing them from conflicting with other global CSS definitions.

See how I'm using it to locally scope the `slide-from-right` sidebar animations in Site Editor. The animation names no longer pollute the global namespace. They are mangled into names like `_slide-from-left_1uetc_2516`.

There are still open problems. Stylelint doesn't understand the `:local` syntax and complains:
```
49:20  ✖  Unexpected unknown pseudo-class selector ":local"  selector-pseudo-class-no-unknown
```

Also, VSCode complains about the syntax:

Screenshot 2024-05-24 at 13 22 36

Here I'm not even sure where the error comes from. Is it a built-in VSCode SCSS syntax checker? Or is it a lint plugin that the Gutenberg repo can configure?

**How to test:**
Go to Site Editor and verify that the sidebar animations (sliding as you navigate back and forth) still work:

Screenshot 2024-05-24 at 13 26 29

### DIFF
--- a/packages/postcss-plugins-preset/lib/index.js
+++ b/packages/postcss-plugins-preset/lib/index.js
@@ -1,1 +1,7 @@
-module.exports = [ require( 'autoprefixer' )( { grid: true } ) ];
+module.exports = [
+	require( 'autoprefixer' )( { grid: true } ),
+	require( 'postcss-modules' )( {
+		scopeBehaviour: 'global',
+		getJSON: () => {},
+	} ),
+];
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -23,6 +23,10 @@
 		],
 		"rule-empty-line-before": null,
 		"selector-class-pattern": null,
+		"selector-pseudo-class-no-unknown": [
+			true,
+			{ "ignorePseudoClasses": [ "local" ] }
+		],
 		"value-keyword-case": null,
 		"scss/operator-no-unspaced": null,
 		"scss/selector-no-redundant-nesting-selector": null,
--- a/package-lock.json
+++ b/package-lock.json
@@ -29068,6 +29068,24 @@
 				"node": "^16.13.0 || >=18.0.0"
 			}
 		},
+		"node_modules/generic-names": {
+			"version": "4.0.0",
+			"resolved": "https://registry.npmjs.org/generic-names/-/generic-names-4.0.0.tgz",
+			"integrity": "sha512-ySFolZQfw9FoDb3ed9d80Cm9f0+r7qj+HJkWjeD9RBfpxEVTlVhol+gvaQB/78WbwYfbnNh8nWHHBSlg072y6A==",
+			"dev": true,
+			"dependencies": {
+				"loader-utils": "^3.2.0"
+			}
+		},
+		"node_modules/generic-names/node_modules/loader-utils": {
+			"version": "3.2.1",
+			"resolved": "https://registry.npmjs.org/loader-utils/-/loader-utils-3.2.1.tgz",
+			"integrity": "sha512-ZvFw1KWS3GVyYBYb7qkmRM/WwL2TQQBxgCK62rlvm4WpVQ23Nb4tYjApUlfjrEGvOs7KHEsmyUn75OHZrJMWPw==",
+			"dev": true,
+			"engines": {
+				"node": ">= 12.13.0"
+			}
+		},
 		"node_modules/gensync": {
 			"version": "1.0.0-beta.2",
 			"resolved": "https://registry.npmjs.org/gensync/-/gensync-1.0.0-beta.2.tgz",
@@ -35832,6 +35850,12 @@
 			"resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz",
 			"integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg=="
 		},
+		"node_modules/lodash.camelcase": {
+			"version": "4.3.0",
+			"resolved": "https://registry.npmjs.org/lodash.camelcase/-/lodash.camelcase-4.3.0.tgz",
+			"integrity": "sha512-TwuEnCnxbc3rAvhf/LbG7tJUDzhqXyFnv3dtzLOPgCG/hODL7WFnsbwktkD7yUV0RrreP/l1PALq/YSg6VvjlA==",
+			"dev": true
+		},
 		"node_modules/lodash.clonedeep": {
 			"version": "4.5.0",
 			"resolved": "https://registry.npmjs.org/lodash.clonedeep/-/lodash.clonedeep-4.5.0.tgz",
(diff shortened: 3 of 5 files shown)


tyxla:
This seems like a sensible solution to me 👍 

> Stylelint doesn't understand the `:local` syntax and complains:

I've added [a commit](https://github.com/WordPress/gutenberg/pull/61940/commits/b69414392e5c286f23ead7fd2918cdd247b50994) to resolve that - seems like we need to ignore it specifically, because the `selector-pseudo-class-no-unknown` rule doesn't recognize it as a valid pseudo-class selector:

https://github.com/stylelint/stylelint/blob/aec445c3a3a8aff474847bf529886ee52a1c4ff5/lib/reference/selectors.mjs

Not sure about the VSCode error at a first glance 🤔

jsnajdr:
> Not sure about the VSCode error at a first glance 🤔

Turns out to be a problem with the VSCode's CSS language server and its CSS parser. Reported it here: https://github.com/microsoft/vscode-css-languageservice/issues/393

So, the `:local` scopes work, but from the VSCode problem it seems that I'm the first person to every try this 😄 Or the last person to use CSS modules.

packages/postcss-plugins-preset/lib/index.js
```diff
@@ -1 +1,7 @@
-module.exports = [ require( 'autoprefixer' )( { grid: true } ) ];
+module.exports = [
+	require( 'autoprefixer' )( { grid: true } ),
+	require( 'postcss-modules' )( {
+		scopeBehaviour: 'global',
+		getJSON: () => {},
+	} ),
```
mirka, line 6: I'm not entirely sure if it's appropriate/necessary to add this to `@wordpress/postcss-plugins-preset`, rather than just for the Gutenberg project. I feel like most WP extenders would not need this, or will have their own setup if they do. It's also non-obvious that `@wordpress/postcss-plugins-preset` would support CSS Modules by default. Thoughts?

There is also a possibility that we will be using CSS Modules for wp-components, and in that case we will need to add additional config (to un-noop the `getJSON`, at the very least) for the Gutenberg project anyway. Not super relevant to the decision we make here, but just adding as an FYI.

jsnajdr: I agree that we can start with enabling `postcss-modules` only for the Components package :+1:

DaniGuardiola (approved):
Love it! Simple and effective.

I do wish that we had a way to implement this without the need for `:local` at all. What I mean by that is something like having everything be global by default, and then have specific things like animations be local by default. That way, the source would just look like normal CSS, and we wouldn't need the `:local` pseudoclass in the declaration nor in the usage.

For now I think this is perfectly fine, though I might take a stab at implementing a simple PostCSS plugin that does what I described at some point.

jsnajdr:
> I do wish that we had a way to implement this without the need for `:local` at all. What I mean by that is something like having everything be global by default, and then have specific things like animations be local by default.

This is what the PR mostly does: all styles are still global, and only these explicitly marked as `:local` are mangled.

Do you mean a variation of this that just mangles all keyframes names automatically, without tagging them? For that we could indeed implement our own PostCSS plugin. Modeled after `postcss-modules-scope` -- it's just one file:

https://github.com/css-modules/postcss-modules-scope/blob/master/src/index.js

For our purposes we could heavily simplify this. Just walk the AST, find all `@keyframes` and `animation-name` nodes and mangle their values.

DaniGuardiola:
@jsnajdr yes, that is what I propose, have those be local by _default_ (and then possibly allow `:global` to revert it). If you see value in this, I will take a stab at it. I love writing AST related plugins :D

DaniGuardiola:
@jsnajdr I found this plugin that does what we need, more or less. I do not love that the prefix is fixed, since that could potentially result in someone re-using a local animation. It also does not have a way to opt out of it. It could be a better starting point for a fork though. Let me do a quick attempt and see what I can come up with.

Edit: forgot to link the plugin lol - https://github.com/VitaliyR/postcss-prefix-keyframe

jsnajdr:
Yes, `postcss-prefix-keyframe` is exactly what I had in mind: simple AST visitor with just two rules, for `AtRule.keyframes` and `Declaration.animation-name`.

> I do not love that the prefix is fixed

`postcss-modules` has a smarter algorithm for that: `generateScopedName` function that includes some hashes (generated from the file path maybe?) and configurable by the user.

> If you see value in this, I will take a stab at it.

Yes, I think it would be a much better solution that this PR currently is :+1: Go for it!

DaniGuardiola:
WIP plugin here, already does what I proposed so far: https://github.com/DaniGuardiola/postcss-local-keyframes

If there are no objections or further improvements, I will publish to npm soon and then create a PR here that uses it.

fullofcaffeine (approved):
I prefer making variables local by default and global when needed, as it's safer and provides better DevEX. However, considering the current feature requirements, I'd say shipping this now is fine while we evaluate @DaniGuardiola's plugin, as migrating to a new approach would be simple 👍🏻

DaniGuardiola:
@fullofcaffeine I fully understand your reasoning, though to be clear, these "local" and "global" concepts are fully dependent on a commitment to using CSS modules, which in my understanding is not what we're doing here. In fact, I personally don't think that's the best idea, and have some ideas to handle style scoping in the future that I will be proposing and experimenting with (in a nutshell, use the native CSS feature for scoping, `@scope`, since it makes non-standard things like CSS modules unnecessary).

This PR is (again, in my understanding) about `@keyframes` specifically, and the suggested approach of using CSS modules is just a means to that end. Hopefully my plugin will prove even simpler. In the long run, I'm hoping that `@scope` will allow local `@keyframes` declarations so we can remove my plugin entirely and just use vanilla, native CSS fully.

jsnajdr:
> I prefer making variables local by default and global when needed

That means full commitment to CSS modules and major rewrite of all CSS files. This PR is much less ambitious: it's using CSS modules only to mark a very small numbers of selectors as `:local`, keeping everything else the same.

packages/postcss-plugins-preset/lib/index.js
DaniGuardiola: What's the appropriate place to add this config only to Gutenberg?

tyxla: Gutenberg's webpack config in `tools/webpack` perhaps?

jsnajdr: I think all you need is to change the config passed to `postcss()` here:

https://github.com/WordPress/gutenberg/blob/add/postcss-modules/bin/packages/build-worker.js#L126

It should configure `postcss` with both the preset and the additional plugin.

DaniGuardiola: Great thank you, now I'm gonna try to test it in my branch to see if it has the desired effect. If you have suggestions on how to find the output that'd be a helpful shortcut :)

jsnajdr: Run `npm run build:packages`. That should build all packages, including styles.

Then look at `packages/edit-site/build-style/style.css`. That's the output of the build, all `edit-site` styles processed and concatenated together.

There is no bundler in this build pipeline, it's just the `bin/packages/build.js` script spawning a farm of `bin/packages/build-worker.js` scripts.

fullofcaffeine:
>  I fully understand your reasoning, though to be clear, these "local" and "global" concepts are fully dependent on a commitment to using CSS modules, 

> That means full commitment to CSS modules and major rewrite of all CSS files. This PR is much less ambitious: it's using CSS modules only to mark a very small numbers of selectors as :local, keeping everything else the same.

Understood, thanks for elaborating!

jsnajdr:
Closing in favor of #62476. It was a nice experiment 🙂

PR #61942, opened 2024-05-24
DataViews: `label` prop in Actions API can be either a `string` or a `function`

ntsekouras:
Related: https://github.com/WordPress/gutenberg/pull/61867

This PR allows `label` prop in Actions API can be either a `string` or a `function`, in case we want to use information from the selected items. The function  accepts the selected items, in case an action wants to modify the label based on that information.

It makes more sense for **non** bulk actions and in this PR is used for the post revisions action to also display the number of available revisions.

Since the action's label is used mostly in places that expect a string, I think it makes sense not to have a component, but just a callback that accepts the selected items and returns a `string`.

## Testing Instructions
Ensure that actions and their labels work properly: 
1. in lists pages in site editor where we use Data Views
2. in inspector controls for any given entity

The only visible change from this PR is that post revisions action label includes the number of the revisions.

### DIFF
--- a/packages/dataviews/README.md
+++ b/packages/dataviews/README.md
@@ -14,7 +14,6 @@ npm install @wordpress/dataviews --save
 
 ```jsx
 const Example = () => {
-
 	// Declare data, fields, etc.
 
 	return (
@@ -27,7 +26,7 @@ const Example = () => {
 			paginationInfo={ paginationInfo }
 		/>
 	);
-}
+};
 ```
 
 ## Properties
@@ -42,12 +41,14 @@ Example:
 const data = [
 	{
 		id: 1,
-		title: "Title",
-		author: "Admin",
-		date: "2012-04-23T18:25:43.511Z"
+		title: 'Title',
+		author: 'Admin',
+		date: '2012-04-23T18:25:43.511Z',
 	},
-	{ /* ... */ }
-]
+	{
+		/* ... */
+	},
+];
 ```
 
 By default, dataviews would use each record's `id` as an unique identifier. If it's not, the consumer should provide a `getItemId` function that returns one.
@@ -125,8 +126,8 @@ Each field is an object with the following properties:
 -   `enableSorting`: whether the data can be sorted by the given field. True by default.
 -   `enableHiding`: whether the field can be hidden. True by default.
 -   `filterBy`: configuration for the filters.
-    - `operators`: the list of operators supported by the field.
-    - `isPrimary`: whether it is a primary filter. A primary filter is always visible and is not listed in the "Add filter" component, except for the list layout where it behaves like a secondary filter.
+    -   `operators`: the list of operators supported by the field.
+    -   `isPrimary`: whether it is a primary filter. A primary filter is always visible and is not listed in the "Add filter" component, except for the list layout where it behaves like a secondary filter.
 
 ### `view`: `object`
 
@@ -140,7 +141,7 @@ const view = {
 	search: '',
 	filters: [
 		{ field: 'author', operator: 'is', value: 2 },
-		{ field: 'status', operator: 'isAny', value: [ 'publish', 'draft'] }
+		{ field: 'status', operator: 'isAny', value: [ 'publish', 'draft' ] },
 	],
 	page: 1,
 	perPage: 5,
@@ -150,7 +151,7 @@ const view = {
 	},
 	hiddenFields: [ 'date', 'featured-image' ],
 	layout: {},
-}
+};
 ```
 
 Properties:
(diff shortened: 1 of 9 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -470,7 +470,15 @@ const viewPostAction = {
 
 const postRevisionsAction = {
 	id: 'view-post-revisions',
-	label: __( 'View revisions' ),
+	getLabel( items ) {
```
youknowriad: What if instead of a new property getLabel, we just make the label `string | ReactComponent`

youknowriad: A React component would allow calling selectors, hooks ... in additional to using the "items".

ntsekouras: I didn't want to mix a prop with such different types, but don't have a strong opinion on this. A component could do that, but we use the label in specific internal components that render the actions and they expect a string returned value. 

Do you think it would be enough to add docs that we support a label component, but expect a string value?

youknowriad: > they expect a string returned value.

What are these places?

youknowriad: I've just looked into the dataviews package, and I think we can support components in all the places where action.label is used now. That said, I would not mind having other opinions here. For block types, we have `__experimentalLabel` which is basically always a component that is kind of equivalent. 

Any thoughts @mcsf @ellatrix @jorgefilipecosta

ntsekouras: >they expect a string returned value.

Sorry, I should rephrase.. What I meant is that the label is used in places (menu items, tooltips, modal header title, etc..) that expect a final 'string' value - not per JS type per se. So while we can support components technically, personally I'd rather not do it from the start to try to keep consistency of the UI that is handled by the internal packages and not third party consumers. 

`__experimentalLabel` is equivalent yes, and also returns a `string` value in our codebase. If we returned something different - that we could - it would definitely create problems in UI. if you see most of the implementations of this callback it's either passed the block attributes, which is like our selected items or have a hook to get the current entity.

So, while I don't have a strong opinion to use a component, I'd rather we are just as flexible as we need right now. I don't think we have a use case right now for needing components, do we?

packages/editor/src/components/post-actions/actions.js
```diff
@@ -470,7 +470,15 @@ const viewPostAction = {
...
+		const revisionsCount =
+			items[ 0 ]._links?.[ 'version-history' ]?.[ 0 ]?.count ?? 0;
+		return sprintf(
+			/* translators: %s: number of revisions */
+			__( 'View revisions (%s)' ),
```
jorgefilipecosta, line 589: This approach works but I'm not sure from the design point of view if the idea was to use a suffix in the button, something like:
Screenshot 2024-05-27 at 12 38 20

It would make things complex in terms of API, as we would need something besides the label.

jorgefilipecosta (commented):
Things work well on this PR, and the code looks good. I'm ok with a getLabel function for now that just returns a string if needed in the future we can always switch to a component to allow hooks usage etc.
My only doubt is if the number of revisions should be shown as part of the label or as suffix but that's a design question, from the code point of view things look good.

ntsekouras:
>My only doubt is if the number of revisions should be shown as part of the label or as suffix but that's a design question, from the code point of view things look good.

The label is not used only in `menuitems`. It can be used in tooltips etc.. So I'm not sure we should complicate things to tightly couple the label with the `menuitem` implementation detail.

TBH, having a component or getter is a small detail (although I prefer a simple function). What I'm not sure after @youknowriad 's comment is whether we should keep two props for getting the label or having one that could be a string or a function. I'm still leaning towards two props for clarity, but no strong opinion.

jorgefilipecosta (approved):
I think we can merge things as they are and if needed iterate later on the design or API.

youknowriad:
I don't like the two properties for the same thing but it's fine by me.

packages/dataviews/README.md
```diff
@@ -241,7 +244,7 @@ Collection of operations that can be performed upon each record.
 Each action is an object with the following properties:
 
 -   `id`: string, required. Unique identifier of the action. For example, `move-to-trash`.
--   `label`: string, required. User facing description of the action. For example, `Move to Trash`.
+-   `label`: string|function, required. User facing description of the action. For example, `Move to Trash`. In case we want to adjust the label based on the selected items, a getter function which accepts the selected records as input can be provided. The getter function should always return a `string` value.
```
ntsekouras: Besides this change, everything else was from auto formatting.

jameskoster:
Apologies for not catching this, but do you think it would be feasible to use the `suffix` affordance on `DropdownMenuItem` for the revisions count?

ntsekouras:
>Apologies for not catching this, but do you think it would be feasible to use the suffix affordance on DropdownMenuItem for the revisions count?

@jameskoster 

>The label is not used only in menuitems. It can be used in tooltips etc.. So I'm not sure we should complicate things to tightly couple the label with the menuitem implementation detail.

jameskoster:
Would it be massively over-complicating to include a separate prop for the tooltip content, so that you could specify:

* Label: View revisions
* Suffix: $count
* Tooltip: View revisions ($count)

ntsekouras:
> Would it be massively over-complicating to include a separate prop for the tooltip content, so that you could specify:

I think it will add complexity at the API and TBH I'm not sure it will add much value in general for every action. Do we see many more actions to have something similar? If we update the actions rendering in the future, would that API make sense? 

So, I think while we could explore this and try to evaluate an API for splitting labels, suffix etc.., I don't think it's something that could be done for 6.6, given the timeline.

PR #47701, opened 2023-02-02
Disable Strict Mode, breaks dev site editor

jsnajdr:
Reverts #47639, only the `<StrictMode>` wrappers, and keeps all the other changes -- e2e and effect fixes.

PR #61943, opened 2024-05-24
Try: Re-enable React StrictMode

Mamaduka:
## What?

PR re-enable React StrictMode for editors. It was removed in #47701 and never restored.

## Why?
* Could be helpful with the React 19 upgrade.
* The StrictMode is mandatory for the React Compiler.

## Bugs Discovered
* https://github.com/WordPress/gutenberg/pull/62051
* https://github.com/WordPress/gutenberg/pull/62053
* https://github.com/WordPress/gutenberg/pull/62059
* https://github.com/WordPress/gutenberg/pull/62140
* https://github.com/WordPress/gutenberg/issues/62089
* https://github.com/WordPress/gutenberg/issues/62105

## Testing Instructions
* Smoke test the editors.
* CI checks should be green.

Same.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -4,6 +4,9 @@
 	"themes": [ "./test/emptytheme" ],
 	"env": {
 		"tests": {
+			"config": {
+				"SCRIPT_DEBUG": true
+			},
 			"mappings": {
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -4,9 +4,6 @@
 	"themes": [ "./test/emptytheme" ],
 	"env": {
 		"tests": {
-			"config": {
-				"SCRIPT_DEBUG": true
-			},
 			"mappings": {
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",
--- a/packages/customize-widgets/src/index.js
+++ b/packages/customize-widgets/src/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { createRoot } from '@wordpress/element';
+import { createRoot, StrictMode } from '@wordpress/element';
 import {
 	registerCoreBlocks,
 	__experimentalGetCoreBlocks,
@@ -92,11 +92,13 @@ export function initialize( editorName, blockEditorSettings ) {
 		} );
 
 		createRoot( container ).render(
-			<CustomizeWidgets
-				api={ wp.customize }
-				sidebarControls={ sidebarControls }
-				blockEditorSettings={ blockEditorSettings }
-			/>
+			<StrictMode>
+				<CustomizeWidgets
+					api={ wp.customize }
+					sidebarControls={ sidebarControls }
+					blockEditorSettings={ blockEditorSettings }
+				/>
+			</StrictMode>
 		);
 	} );
 }
--- a/packages/edit-post/src/editor.js
+++ b/packages/edit-post/src/editor.js
@@ -8,7 +8,7 @@ import {
 	store as editorStore,
 	privateApis as editorPrivateApis,
 } from '@wordpress/editor';
-import { StrictMode, useMemo } from '@wordpress/element';
+import { useMemo } from '@wordpress/element';
 import { SlotFillProvider } from '@wordpress/components';
 import { store as coreStore } from '@wordpress/core-data';
 import { CommandMenu } from '@wordpress/commands';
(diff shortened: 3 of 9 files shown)


Mamaduka:
The e2e run with `SCRIPT_DEBUG` [2a60f04](https://github.com/WordPress/gutenberg/pull/61943/commits/2a60f04d43501ee13f05972c8c7d81e6431b872f) should give us a better idea about `StrictMode` and related issues.

.wp-env.json
```diff
@@ -4,6 +4,9 @@
 	"themes": [ "./test/emptytheme" ],
 	"env": {
 		"tests": {
+			"config": {
+				"SCRIPT_DEBUG": true
+			},
```
youknowriad: I'm not entirely sure about this. I think it's good for the dev environment to have this enabled by I believe that ideally the tests environment should be as close to prod as possible.

Also, for performance tests, this will have an impact.

Mamaduka: This is temporary to see if anything fails in the development environment when StictMode is enabled.

youknowriad (approved):
Testing locally, I didn't see anything obvious. LGTM

tyxla:
Thanks @Mamaduka!

I'd love to do some more thorough testing early next week, but if y'all are confident, go for it.

Mamaduka:
As you can see, some e2e tests are failing when they're run in development and strict mode. This probably means we've subtle bugs in effects or anything that runs twice in `StrictMode`.

tyxla:
> As you can see, some e2e tests are failing when they're run in development and strict mode. This probably means we've subtle bugs in effects or anything that runs twice in `StrictMode`.

Sounds familiar 😅

Mamaduka:
Technically, I guess we can merge this since it doesn’t affect production apps.

Then create a separate tracking PR, where we try to fix each failure. Just like we do for React Compiler.

What do you think?

youknowriad:
Yeah, since these bugs are actually already on trunk. I'm ok with merging.

The downside is that some folks might start noticing the bugs in their dev environments and wonder what's happening.

youknowriad:
Actually, let's not merge this in 6.6 (unless we fix all the bugs).

Mamaduka:
Sounds good. I’ll keep temp commit and we can use this as tracking PR.

Mamaduka:
I started fixing the dev environment bugs. The "good news" is that some failures have related causes (e.g., #62053).

ellatrix:
Is the list of bugs in the PR description complete?

Mamaduka:
> Is the list of bugs in the PR description complete?

Not yet. I was trying to isolate and fix the ones that greatly affected the app, like the `useFocusOnMount` bug.

I'll try to complete it today.

Mamaduka:
I think it's a complete list of bugs (caught by e2e tests). The tests report two false positives, which fail due to warnings generated in development mode but aren't real issues.

I'll create separate issues for the remaining problems.

Mamaduka:
The only known bug that remains to be fixed is https://github.com/WordPress/gutenberg/pull/62140. I would appreciate review/testing.

youknowriad:
@Mamaduka Are you certain about the link in your last comment?

Mamaduka:
@youknowriad, it should be fixed now. GitHub autocorrected my PR number :/

Mamaduka:
This is ready to be merged, but we will do it after this week's RC. It is probably not worth adding this as a minute change to WP  6.6.

tyxla (commented):
Noting that at this time this needs to be rebased to take [this fix](https://github.com/WordPress/gutenberg/pull/62154) into consideration otherwise there are PHP fatals. 

I'm giving it a spin with a locally rebased branch to the latest `trunk`.

tyxla (commented):
Do we also want to add strict mode to the reusable blocks?

https://github.com/WordPress/gutenberg/blob/00cfeee55e19be3f590748dd5d5f8a23c93d0a00/packages/list-reusable-blocks/src/index.js#L47

tyxla (approved):
Tested all editors, working pretty well. Shipping after the next release sounds sensible and will give us a good window to fix any incoming problems 👍

Mamaduka:
> Noting that at this time this needs to be rebased to take https://github.com/WordPress/gutenberg/pull/62154 into consideration otherwise there are PHP fatals.

So, just rebase on top of the latest trunk, or are there any additional fixes required?

tyxla:
> > Noting that at this time this needs to be rebased to take #62154 into consideration otherwise there are PHP fatals.
> 
> So, just rebase on top of the latest trunk, or are there any additional fixes required?

Yes, just rebasing fixes it.

kevin940726 (approved):
LGTM 👍. I'd expect to see warnings once this is merged but better sooner than later 😆.

andrewserong:
It looks like this change caused a regression in the customizer's widget editor (reported in #62898). Just sharing here in case anyone has an idea for how to fix it? I wasn't sure if it was safe to revert the `<StrictMode>` in `customize-widgets/src/index.js` or if there's a better approach.

t-hamano:
Strangely, two commands seem to no longer work when `StrictMode` is enabled. Please see #64004.

Issue #61763, opened 2024-05-17
Improve implementation of the 'Show button text labels' preference


afercia:
### Description

The 'Show button text labels' preference is an usability and accessibility features that makes icon buttons show visible text instead of icons.

It's not just accessibility. It's general usability, as some users may just _prefer_ visible text. It has been inspired by the similar macOS finder preference, where users may choose to show buttons with: Icon and text, Icon only, Text only. Screenshot:

![Screenshot 2024-05-17 at 15 17 04](https://github.com/WordPress/gutenberg/assets/1682452/fea9149d-b07f-4b7a-8c8b-297aad002f53)

So far, this feature isn't 'built-in' in the components. Rather, it is implemented by the means of 'local overrides' of the buttons styling. It is also worth reminding that it is not implemented for _all_ buttons but only for some of them. The related CSS is scattered all around in several files and compoennts. It's difficult to maintain and it's duplicated code.

Over time, this feature faced several breakages as it is often missed in the design of new features in the first place and often missed during developing and manual testing. https://github.com/WordPress/gutenberg/issues/61761 is the most recent breakage.

The `Button` component isn't aware of this preference, as it's more a preference of the environment the component lives in. As such, an architectural choice should me made first on whether the Button componetn should support this feature natively or keeping the current approach based on local overrides. As I see it, considering the several breakages in the history of this project related to this feature, the former option would be preferable.

I would appreciate any thoughts and feedback on the best way to make this feature more solid, incorporated in any new design, and better testable.

In the meantime, I'd think there's a few points that can be addressed soon to make this feature better meintenable. For example, looking at the CSS used for local overrides, there's a few things that are duplicated all around and could use a mixin:

What the various CSS overrides always do is basically the following:
- Hide the button SVG icon.
- Generate the button visible text by the means of CSS pseudo generated content that rargets the button `aria-label` attribute.
- Reset the button width to `auto`.

As a start, a new mixin could be used to have this rule in a centralized place and avoid code duplication. The mixin could be then extended to incorporate other styling, as much as possible. Small adjustments on a per component base would be fine, I guess, but the main rules for the styling should be centralized in an unique place.

### Step-by-step reproduction instructions

N/A

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

afercia:
[...]

PR #61949, opened 2024-05-24
InspectorControls: Text not displayed when "Show button text labels" is enabled

### LINKED ISSUE #61947: Inspector tabs: text no longer visible when 'Show button text labels' preference is enabled
### Description

This appears to be a regression from WordPress 6.5 and should be fixed soon.

Depending on the selected block, the Inspector panel may show content organized within tabs. By default these tabs use icon buttons. Example screenshot:

Screenshot 2024-05-24 at 15 48 21

When the 'Show button text labels' preference is enabled, the icons are supposed to be replacred by visible text. Example screenshot from WordPress 6.5. for a Navigation block:

On latest trunk, the text is not visible any longer. Screenshot for a Navigation block:

trunk navigaton

Screenshot for an Image block:

trunk image

It appears that some recent changes to a CSS transition introduced this bug, as the `::before` pseudo element that is supposed to generate the visible text has now an `opacity: 0`.

Worth noting the technique to reveal the text for the 'Show button text labels' preference  is fragile to start with, as it uses pseudo elements that _may_ be already used for other purposes. In this case, I think it's used also for the tabs focus style.

Regardless the change to the animation should be either fixed or reverted, as WP 6.6 Beta is very close.

- Go to Options > Preferences > Accessibility, and enable the 'Show button text labels' preference.
- Select any block that is supposed to render tabs in its Inspector panel e.g. an Image block or a Navigation block.
- Observe the tabs text is no longer visible.

_No response_

### Environment info

_No response_

Yes

Yes

t-hamano:
Fixes #61947
Seems to be caused by #60560

## What?

This PR fixes an issue where tab text in InspectorControls is not displayed when "Show button text labels" is enabled.

image

## Why?

When "Show button text labels" is enabled, text is often displayed using the following approach:

```scss
.show-icon-labels {
	.some-element {
		.components-button.has-icon {
			// Hide the button icons when labels are set to display...
			svg {
				display: none;
			}
			// ... and display labels.
			&::before { // Or `&::after`
				content: attr(aria-label);
			}
		}
	}
}
```

However, #60560 applied `opacity: 0;` to the `before` pseudo-element, so the text is no longer visible.

## How?

Fortunately #60560 removed the `after` pseudo-element, so I changed `before` to `after`. But this is _a temporary fix_ for WP6.6. It may not be an ideal approach as it may be affected by changes to the `Tab` component in the future. Overall, we will need to consider the implementation itself if "Show button text labels" is enabled, as discussed in #61763.

## Testing Instructions

- Enable "Show button text labels".
- Select some block.
- Make sure the tab text is visible.

### DIFF
--- a/packages/block-editor/src/components/inspector-controls-tabs/index.js
+++ b/packages/block-editor/src/components/inspector-controls-tabs/index.js
@@ -5,6 +5,8 @@ import {
 	Button,
 	privateApis as componentsPrivateApis,
 } from '@wordpress/components';
+import { store as preferencesStore } from '@wordpress/preferences';
+import { useSelect } from '@wordpress/data';
 
 /**
  * Internal dependencies
@@ -24,6 +26,10 @@ export default function InspectorControlsTabs( {
 	hasBlockStyles,
 	tabs,
 } ) {
+	const showIconLabels = useSelect( ( select ) => {
+		return select( preferencesStore ).get( 'core', 'showIconLabels' );
+	}, [] );
+
 	// The tabs panel will mount before fills are rendered to the list view
 	// slot. This means the list view tab isn't initially included in the
 	// available tabs so the panel defaults selection to the settings tab
@@ -43,10 +49,16 @@ export default function InspectorControlsTabs( {
 							tabId={ tab.name }
 							render={
 								<Button
-									icon={ tab.icon }
-									label={ tab.title }
+									icon={
+										! showIconLabels ? tab.icon : undefined
+									}
+									label={
+										! showIconLabels ? tab.title : undefined
+									}
 									className={ tab.className }
-								/>
+								>
+									{ showIconLabels && tab.title }
+								</Button>
 							}
 						/>
 					) ) }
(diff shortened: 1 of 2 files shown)


tyxla:
cc @DaniGuardiola who worked on https://github.com/WordPress/gutenberg/pull/60560

t-hamano:
As an alternative approach, I changed the implementation to one that does not rely on pseudo-elements. Please check the two commits.

- [5f42aa5](https://github.com/WordPress/gutenberg/pull/61949/commits/5f42aa508ddfd4ed212b10e45b0bbd9099249395): Simply changed the pseudo element from `before` to `after`
- [e21b144](https://github.com/WordPress/gutenberg/pull/61949/commits/e21b144ee3fa4d5e9a9dd4d63fd5aa16b59150cb): Controls button behavior by respecting the showIconLabels preference

DaniGuardiola (approved):
Thank you @t-hamano, I agree that not relying in pseudo-elements is a better solution. The code looks good to me 👍

mirka:
This is a really good example of the problems discussed in #61763. I understand this is a temporary fix, pending a more overarching solution, but I like how it moves/isolates the responsibility into a more robust place. Thanks @t-hamano 🙏

t-hamano:
Thanks for the review!

Issue #57915, opened 2024-01-17
Post Date block doesn't show the modified date for newly published posts


bugnumber9:
### Description

When a new post is published and hasn't been modified after publishing, the Post Date block doesn't show the modified date.

I confirmed that `get_the_modified_date` function works as expected in a non-FSE theme (Hello theme) by altering the code in single.php as shown here:
![hello](https://github.com/WordPress/gutenberg/assets/1769548/3aa7b431-5f6b-4bde-8327-25137de27bf6)
The output is expected:
![frontend2](https://github.com/WordPress/gutenberg/assets/1769548/0bfb17d8-4b84-41dc-9893-d1f9b709b215)

However, when I change a single post template in site editor (standard 2024 theme) as shown here:
![site-editor](https://github.com/WordPress/gutenberg/assets/1769548/a237969c-513a-4efb-a14a-98485e9c1dfa)
The modified date on the frontend is empty:
![frontend](https://github.com/WordPress/gutenberg/assets/1769548/1bf9e89f-1576-4435-a9d9-444bdb1bf25a)

### Step-by-step reproduction instructions

Shown above.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

benniledl:
I see, that this was the most recent change in the post date block and was intentional: #46839

benniledl:
@bugnumber9 I recommend using 
`get_the_modified_date() ? get_the_modified_date() : get_the_date();`
This returns the modified date and if that is empty it returns the normal date

bugnumber9:
Well, this is a bit weird.
Let's say, I want to show the "last updated" not "created" date in the postmeta section (using FSE) - it won't show for posts that weren't modified after publishing.
I was able to create a workaround using Block Logic and a custom PHP function, but it's surely not ideal.

benniledl:
@bugnumber9 you don't need an extra function you can just do 
`echo 'Modified: ' .  get_the_modified_date() ? get_the_modified_date() : get_the_date();`

bugnumber9:
> @bugnumber9 I recommend using `get_the_modified_date() ? get_the_modified_date() : get_the_date();` This returns the modified date and if that is empty it returns the normal date

[...]

PR #46839, opened 2023-01-02
Only display the modified post date if the post has been modified

carolinan:
## What?

The post date block has a setting that displays the modified date. When the setting is enabled, it displays the post date
if the post has not been modified.

This PR updates the block to only display the modified date if the setting is enabled and the post has been modified.
If the setting is enabled, but the post has not been modified, nothing is shown on the front.
It also adds a new CSS class to identify that the setting is enabled so that developers can style it differently.

Fixes [#46645](https://github.com/WordPress/gutenberg/issues/46645)

## Why?

Without this PR, a template that displays both the published date and the modified date could show two identical dates,
and there was no way to hide the modified date if the post was not modified.

## How?

The PR compares the modified date and the post date and returns early if the time difference is a less than a minute.

## Testing Instructions

1. Create a new post.
2. Add a post date block.
3. _Wait a minute or two_ and add a second post date block. Enable the "Display last modified date" setting.
4. Save and view the front. Both blocks should show, with different dates.

5. Create a second new post.
6. Add a post date block. Enable the "Display last modified date" setting.
7. Save and view the front. If you were fast enough, the block should not show.
8. Go back to the editor, _wait a minute_ and make a change to the post.
9. Save and view the front, confirm that the block shows.

PR #61953, opened 2024-05-24
Post Date block always show the accurate Last Modified Date

TimBHowe:
## What?
Remove the if statement preventing the Post Date block from returning an empty value when the Last Modified Date is used and it matches the Published Date.

## Why?
If you add a label before the [Post Date](https://wordpress.org/documentation/article/post-date-block/) block and you are using the Last Modified Date if is the same as the publish date then it doesn't display on the front end despite being shown in the editor on the back end.

This is another alternative solution to the issue mentioned above and the other pull request I made here: https://github.com/WordPress/gutenberg/pull/61920
For this issue: https://github.com/WordPress/gutenberg/issues/47738

This being subtractive should be simpler to merge and act as a temporary solution until a Bit solution can be created.
https://github.com/WordPress/gutenberg/discussions/39831
https://github.com/WordPress/gutenberg/pull/61920#issuecomment-2128136026

**This does cause an issue...**
If the publish date is set in the future to schedule the post the Publish Date and Last Modified Date will show the post was last modified prior to being published. While this is true it may not be ideal for UX/UI.

## How?
Removed the if statement here: https://github.com/dsas/gutenberg/blob/5b95e96a90aac522c6df73349930144b89b53dfa/packages/block-library/src/post-date/index.php#L38

## Testing Instructions

1. Create a new blog post
2. Add a Group block to create a Row
3. Add a Paragraph block with the text: "Posted On:"
4. Add a Post Date block with the "Display last modified date" toggle disabled
3. Add a Paragraph block with the text: "Updated On:"
6. Add a Post Date block with the "Display last modified date" toggle enabled
9. Save the post and view the post. Both should appear with the same date.
10. Update the publish date to the past. Both should appear with the same date.
11. Update the publish date to the future will schedule the post. Both should appear with the same date.
12. Update your test servers system clock to past the future date. Both should appear with the same date, but _this might be seen as an issue with the Publish date is after the Modified date._ 

## Screenshots or screencast 
image
image
image
image
image
image
image

Example of potential issue:
image

### DIFF
--- a/packages/block-library/src/post-date/edit.js
+++ b/packages/block-library/src/post-date/edit.js
@@ -181,9 +181,7 @@ export default function PostDateEdit( {
 							} )
 						}
 						checked={ displayType === 'modified' }
-						help={ __(
-							'Only shows if the post has been modified'
-						) }
+						help={ __( "Show the post's last modified date" ) }
 					/>
 				</PanelBody>
 			</InspectorControls>
--- a/packages/block-library/src/post-date/index.php
+++ b/packages/block-library/src/post-date/index.php
@@ -33,17 +33,12 @@ function render_block_core_post_date( $attributes, $content, $block ) {
 	}
 
 	/*
-	 * If the "Display last modified date" setting is enabled,
-	 * only display the modified date if it is later than the publishing date.
+	 * If the "Display last modified date" setting is enabled.
 	 */
 	if ( isset( $attributes['displayType'] ) && 'modified' === $attributes['displayType'] ) {
-		if ( get_the_modified_date( 'Ymdhi', $post_ID ) > get_the_date( 'Ymdhi', $post_ID ) ) {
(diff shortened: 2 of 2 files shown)


Mamaduka:
I think this was resolved by recent refactoring such as #70585.

cc @ockham

ockham:
> I think this was resolved by recent refactoring such as #70585.

@Mamaduka I'm afraid not -- I tried to keep the block as consistent as possible with its previous behavior. In practice, while the logic to provide the last modified date was moved to the newly introduced `core/post-data` source, I included the [conditional](https://github.com/WordPress/gutenberg/pull/70585/files#diff-fbcd0626135c09a8c23244f67553e50b44c1993b0dbaf645b764ecda8588b8efR42-R47) that will hide it if it is before (or equal to) the publish date.

Mamaduka:
It looks like the condition was introduced in #46839. There's also an issue to resolve this particular case - #57915.

I think the patch needs refreshing, so it's applied to the correct place after recent refactorings.

PR #61703, opened 2024-05-16
Fix long template name UI break issue

itzmekhokan:
Fixes #61702 

### Before
Before-post-template-UI-1

### After applying PR - https://github.com/WordPress/gutenberg/pull/61703
After-post-templte-UI-1

## Use of AI Tools
AI assistance: Yes
Tool(s): Claude

PR #61492, opened 2024-05-08
Try: Wrap date-time stamp well for publish status

jasmussen:
## What?

The date-stamp in the summary panel can wrap poorly when it's long, usually putting the timezone on its own line out of context with the time itself:
scheduled-wrapping-before

This PR adds text-wrap: pretty; to address this so the timezone is more likely to be in context of the timestamp:

scheduled wrapping after

## Testing Instructions

Write a post, change its status to scheduled, and observe the date-stamp in the summary panel. It should wrap well.

PR #61954, opened 2024-05-24
Make post meta row button treatment consistent

jameskoster:
Alternative to https://github.com/WordPress/gutenberg/pull/61492 and https://github.com/WordPress/gutenberg/pull/61703

## What?
Apply the same treatment to all buttons in `.editor-post-panel__row-control`. Specifically; truncate to one line, and use a tooltip to display the full value where helpful (IE the date field).

## Why?
Currently button treatments are inconsistent and in some cases broken. Here's a comparison between trunk and this branch:

| Trunk | This branch |
| --- | --- |
| Screenshot 2024-05-24 at 16 27 13 | Screenshot 2024-05-24 at 16 24 56 |

It's not uncommon for find long date strings, template names, slugs, author names etc. The UI should be better prepared to handle these scenarios.

Another option to try could be truncating to X number of lines so that details like the full date are always visible, but I prefer truncating to a single line as it keeps the panel neat. The full value is always accessible (either via tooltip, or by opening the associated panel).

### DIFF
--- a/packages/editor/src/components/post-panel-row/style.scss
+++ b/packages/editor/src/components/post-panel-row/style.scss
@@ -11,6 +11,8 @@
 	min-height: $grid-unit-40;
 	display: flex;
 	align-items: center;
+	padding: 6px 0; // Matches button to ensure alignment
+	line-height: $grid-unit-05 * 5;
 }
 
 .editor-post-panel__row-control {
--- a/packages/editor/src/components/post-panel-row/style.scss
+++ b/packages/editor/src/components/post-panel-row/style.scss
@@ -29,8 +29,7 @@
 		min-height: $button-size-compact;
 	}
 
-	.components-dropdown,
-	.editor-post-author__panel-dropdown {
+	.components-dropdown {
 		max-width: 100%;
 	}
 }
--- a/packages/editor/src/components/post-panel-row/style.scss
+++ b/packages/editor/src/components/post-panel-row/style.scss
@@ -13,6 +13,7 @@
 	align-items: center;
 	padding: 6px 0; // Matches button to ensure alignment
 	line-height: $grid-unit-05 * 5;
+	hyphens: auto;
 }
 
 .editor-post-panel__row-control {
--- a/packages/editor/src/components/post-panel-row/style.scss
+++ b/packages/editor/src/components/post-panel-row/style.scss
@@ -20,11 +20,11 @@
 	align-items: center;
 
 	.components-button {
-		text-overflow: ellipsis;
-		white-space: nowrap;
-		overflow: hidden;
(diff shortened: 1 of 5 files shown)


jasmussen:
I'm personally leaning towards not eliding. Not strongly. But so long as it's top aligned, and wraps using `text-wrap: pretty`, it seems okay to wrap.

jameskoster:
[a282328](https://github.com/WordPress/gutenberg/pull/61954/commits/a28232894024e8e2e111703136972959659309ca) removes truncation but fixes alignment. I fear this will make the panel feel overwhemling in situations where there are several long labels, but I acknowledge those may be edge cases. It still fixes the underlying issue which is most important as we approach beta.

Screenshot 2024-05-24 at 16 58 16

jameskoster:
Pushed a small tweak to ensure alignment when the label wraps:

Screenshot 2024-05-28 at 16 12 58

jasmussen:
Coming back to this one, and keeping in mind we need to make an initial call here, and then we can always revisit, I'd recommend we do this:

* Top/left align all items, left and right.
* Allow wrapping, but apply `text-wrap: pretty;` to each.
* If it's not too hard, we can elide if the content spans 2 lines of text or more.

packages/editor/src/components/post-panel-row/style.scss
```diff
@@ -11,11 +11,26 @@
...
+	.components-button {
+		max-width: 100%;
+		text-align: left;
+		text-wrap: pretty;
+		height: auto;
+		min-height: $button-size-compact;
+	}
+
+	.components-dropdown,
+	.editor-post-author__panel-dropdown {
```
ntsekouras: This would be better to be part of the post author component's styles and not part of post panel row.

jameskoster:
> If it's not too hard, we can elide if the content spans 2 lines of text or more.

This will need further investigation and is probably better handled separately.

jasmussen:
Thanks for the work. This is very close. In most cases it looks like this:

Screenshot 2024-05-30 at 09 01 30

I had to use the inspector to fake some longer titles and results:

Screenshot 2024-05-30 at 09 02 59

In general, that's good too, though note how we probably need some `word-break: break-word;` action here.

jameskoster:
Good catch. Pushed a fix:

Screenshot 2024-05-30 at 11 30 16

This approach uses `hyphens` which should aid readability.

jameskoster:
The failing tests seem unrelated to this PR 🤔 Perhaps @ntsekouras can take a look? 🙏

jasmussen (approved):
Great work. Tested, works.

jameskoster:
Thanks all ❤️

PR #46752, opened 2022-12-22
Iframe: skip scoping styles

ellatrix:
## What?

This is unnecessary when the content is iframed so we can skip all this work for better performance.

Fixes https://github.com/WordPress/gutenberg/pull/52767#issuecomment-1643025927 (CSS parser stumbling over escaped characters)

## Why?

Right! Why scope if we don't have to? :)

## How?

Iframes!

## Testing Instructions

Check content styles are not prefix and still work in iframed editors.

PR #56649, opened 2023-11-29
Editor styles: if scope is needed, don't increase specificity

ellatrix:
## What?

Previously:
- #46752
- #56564

Currently, when the editor is iframed (most cases), we don't scope editor styles. Now the problem is that when the styles _do_ need to be scoped in a non iframed editor, all the editor styles' specificity is bumped as well.

## Why?

There's cases in block editor.css stylesheets where the extra `.editor-styles-wrapper` is needed to override other styles.

It's also better to keep the specificity of the editor styles css rules the same between the iframed and non iframed editor.

## How?

We can use `:where()` to avoid adding specificity.

## Testing Instructions

To trigger a non iframed editor, run this in the console:

```
wp.blocks.registerBlockType( 'test/v2', { apiVersion: 2, title: 'test' } );
```

Alternatively you could enable a plugin that adds meta boxes.

PR #61959, opened 2024-05-24
Restore `html` scope in `reset.scss` selectors

### LINKED ISSUE #57176: Regression: iframed post editor can show some admin styles with classic themes
### Description

In classic themes, styles such as the `background-color` and `line-height` from the admin `common.css` can apply to the `body` element within the iframed post editor canvas.

1. Activate Twenty Nineteen or another classic theme that does not set its own background color to the body (without using the Customizer, that also includes bundled themes from Twenty Ten to Twenty Sixteen).
2. Visit the post editor, and check that it is inside an iframe.
3. Note that the background is a light gray (`#f0f0f1`) instead of white.

Twenty Nineteen is one of the themes that shows the admin's light gray background color instead of white.

post editor uses admin CSS for the body background color

In an uncommon and extreme example, a Calendar block has a `line-height` of 18.2 pixels—even when the font size is much bigger—in Twenty Twenty.

font size is 36 pixels and the line-height is only 18.2

### Environment info

The admin CSS bleed can happen with WordPress 6.4.2, either without any plugins active or with Gutenberg 17.2.3. I also experienced it with WordPress 6.3.2, but only when Gutenberg was active (version 16.9.0).

For the screenshots, I used:
- WordPress 6.4.2 with Gutenberg 17.2.3
- Firefox 120.0.1 in Windows 10 Home

Yes

Yes

sabernhardt:
Fixes: #57176

(simplest option)

## What?
Raise the specificity of reset selectors to what they were before [#46752](https://github.com/WordPress/gutenberg/pull/46752) so that the CSS reset overrides styles in [wp-admin/css/common.css](https://github.com/WordPress/wordpress-develop/blob/96de28cc29774aeb4577b0d314e6b1724c1804ae/src/wp-admin/css/common.css#L223-L230).

## Why?
The admin CSS bleed affects the iframed post editor in 'classic' themes that do not specify `font-size`, `line-height` and/or `background-color` for the `body`. In addition to the extra discrepancies between editor and front end, the smaller text is more difficult to read when editing post content, and the background reduces color contrast.

## How?
Revert the `reset.scss` change from [#46752](https://github.com/WordPress/gutenberg/pull/46752).

## Testing Instructions
1. Activate a 'classic' theme such as Twenty Thirteen or Twenty Fourteen (those two currently have the wrong font size and background color).
2. Create a new post and add blocks including a Paragraph and/or List.
3. Check whether the post editor is using an iframe by using the [browser inspector](https://blog.hubspot.com/website/how-to-inspect) on text in a Paragraph block. If it is **not** inside an iframe, you may need to save the post and then deactivate plugins or go to Preferences and remove a panel such as Custom Fields.
4. In the browser inspector, look for styles on the `body` element coming from `common.css` (or `load-styles.php` if `SCRIPT_DEBUG` is `false`). With the PR applied, the `background`, `color`, `font-family`, `font-size` and `line-height` properties should have a line through them (the `min-width` is not reset, but that seems to be fine).
admin body styles crossed out, except for min-width


carolinan:
This change solves the problem with the unexpected grey background color in classic themes.
What negative side effects do we need to test for?

fabiankaegy:
@aaronrobertshaw I think this one may be of interest to you.

packages/block-library/src/reset.scss
```diff
@@ -7,7 +7,7 @@
 
 // We use :where to keep specificity minimal.
 // https://css-tricks.com/almanac/selectors/w/where/
-:where(.editor-styles-wrapper) {
+html :where(.editor-styles-wrapper) {
```
fabiankaegy, line 10: From recent other PR's such as #61638 I think `:root` would be a better selector here.

fabiankaegy, line 10:
```suggestion
:root :where(.editor-styles-wrapper) {
```

sabernhardt: I used `html` so this would be exactly as safe as it was in WordPress 6.3.

The only selector that I know needs an increased specificity is `:where(.editor-styles-wrapper)`, so it can override the admin styles on `body`, and `0-0-1` would be enough for that element.

aaronrobertshaw: Given this is a reset stylesheet, I'm inclined towards the `0-0-1` specificity as well. 

Perhaps we could further isolate the change here by extracting only the rules needed to override the `common.css` `body` styles under the restored selector. For example:

```css
html :where(.editor-styles-wrapper) {
	/**
	* The following styles revert to the browser defaults overriding the WPAdmin styles.
	* This is only needed while the block editor is not being loaded in an iframe.
	*/
	font-family: serif; // unfortunately initial doesn't work for font-family.
	font-size: initial;
	line-height: initial;
	color: initial;
	// Many themes with white backgrounds load editor styles but fail to also provide
	// an explicit white background color, assuming a white editing canvas.
	// So to match browser defaults, we provide a white default here as well.
	background: #fff;
}
```

The rest could remain under `:where(.editor-styles-wrapper)`.

There is also [another PR](https://github.com/WordPress/gutenberg/pull/56649) around the editor styles wrapper that might be worth taking a look at to see how these changes might be affected.

sabernhardt: If you want to isolate the one element, you could use the `html &` method I mentioned in https://github.com/WordPress/gutenberg/issues/57176#issuecomment-1861193823

However, a PR like that is more complex than I should try to create within the browser, so someone else should make it.

Plus, the comment about not being in an iframe is inaccurate (with classic themes), so that sentence could be removed too.

carolinan:
How can we arrive at a final decision about how to solve?

sabernhardt:
#62350 is the version that only creates `html :where(.editor-styles-wrapper)` once in the compiled stylesheet.

sabernhardt:
https://github.com/WordPress/gutenberg/pull/62350 has been merged instead of this one.

PR #61960, opened 2024-05-24
Interactivity API: Fix null and number strings as namespaces runtime error.

cbravobernal:
## What?

Right now, creating a namespace called `null` or an integer `24` will kill the runtime.

## Why?

We are doing a JSON.parse() function that returns `null` type or `number` type if receives `null` or `number` strings.

## How?

By checking if it is an object.

## Testing Instructions

1) Create an interactive block.
2) Use null or a number for the namespace in render.php
3) The interactive block should work.

### DIFF
--- a/packages/e2e-tests/plugins/interactive-blocks/namespace/render.php
+++ b/packages/e2e-tests/plugins/interactive-blocks/namespace/render.php
@@ -29,3 +29,17 @@
 <div data-wp-interactive>
 	<a data-wp-bind--href="other::state.url" data-testid="other namespace"></a>
 </div>
+
+<div data-wp-interactive="true">
+	<a data-wp-bind--href="state.url" data-testid="true namespace"></a>
+</div>
+
+<div data-wp-interactive="false">
+	<a data-wp-bind--href="state.url" data-testid="false namespace"></a>
+</div>
+<div data-wp-interactive="[]">
+	<a data-wp-bind--href="state.url" data-testid="[] namespace"></a>
+</div>
+<div data-wp-interactive='"quoted string"'>
+	<a data-wp-bind--href="state.url" data-testid="quoted namespace"></a>
+</div>
--- a/packages/e2e-tests/plugins/interactive-blocks/namespace/render.php
+++ b/packages/e2e-tests/plugins/interactive-blocks/namespace/render.php
@@ -18,6 +18,14 @@
 	<a data-wp-bind--href="state.url" data-testid="object namespace"></a>
 </div>
 
+<div data-wp-interactive="null">
+	<a data-wp-bind--href="state.url" data-testid="null namespace"></a>
+</div>
+
+<div data-wp-interactive="2">
+	<a data-wp-bind--href="state.url" data-testid="number namespace"></a>
+</div>
+
 <div data-wp-interactive>
 	<a data-wp-bind--href="other::state.url" data-testid="other namespace"></a>
 </div>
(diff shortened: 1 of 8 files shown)


packages/e2e-tests/plugins/interactive-blocks/namespace/render.php
```diff
@@ -18,6 +18,14 @@
 	<a data-wp-bind--href="state.url" data-testid="object namespace"></a>
 </div>
 
+<div data-wp-interactive="null">
+	<a data-wp-bind--href="state.url" data-testid="null namespace"></a>
+</div>
+
+<div data-wp-interactive="2">
+	<a data-wp-bind--href="state.url" data-testid="number namespace"></a>
+</div>
```
sirreal, line 27: These would've had similar interesting behavior, all the JSON values:

```html
<div data-wp-interactive="true"></div>
<div data-wp-interactive="false"></div>
<div data-wp-interactive="[]"></div>
<div data-wp-interactive='"my namespace is a quoted string"'></div>
```

sirreal (commented):
I checked the server side, it seems to handle this well (decodes to associative array and checks for is_array):

https://github.com/WordPress/wordpress-develop/blob/581f8d5eebcc57c8f2e8fe05e44db6cc183e1c26/src/wp-includes/interactivity-api/class-wp-interactivity-api.php#L558-L565

packages/interactivity/src/vdom.ts
```diff
@@ -100,7 +100,8 @@ export function toVdom( root: Node ): Array< ComponentChild > {
 					const namespace = regexResult?.[ 1 ] ?? null;
 					let value: any = regexResult?.[ 2 ] ?? attributeValue;
 					try {
-						value = value && JSON.parse( value );
+						const parsedValue = JSON.parse( value );
+						value = isObject( parsedValue ) ? parsedValue : value;
 					} catch ( e ) {}
```
sirreal: Nit: we can omit `( e )` when not using the error:
```suggestion
					} catch {}
```

cbravobernal: Fixed in https://github.com/WordPress/gutenberg/pull/61960/commits/b7b6da8f447a974daf807c8034119ccd6824fcaa

packages/e2e-tests/plugins/interactive-blocks/namespace/view.js
```diff
@@ -14,3 +21,50 @@ store( 'other', {
 		url: '/other-store-url',
 	},
 } );
+
+store( 'null', {
+	state: {
+		url: '/some-url',
```
sirreal: I'd like to see different values in the different stores. I think it's unlikely, but one could imagine falsy values evaluating to the same store.

sirreal (approved):
I tested this and it does allow me to make a (terribly named) store with the namespace `null`.

packages/e2e-tests/plugins/interactive-blocks/namespace/view.js
cbravobernal: TBH, I'm not 100% sure if we should allow this values to be namespaces 😅

cbravobernal:
> I tested this and it does allow me to make a (terribly named) store with the namespace `null`.

TBH, I'm not 100% sure if we should allow this values to be namespaces 😅

packages/e2e-tests/plugins/interactive-blocks/namespace/view.js
cbravobernal: Done in https://github.com/WordPress/gutenberg/pull/61960/commits/64e275d4fdd6ff47e30a45e5933a0dba5a92f892

Issue #60358, opened 2024-04-02
Documentation: Add missing docs in `editor` package


ntsekouras:
Recently, we [added](https://github.com/WordPress/gutenberg/pull/60356) auto-generated docs for the `editor` package. 

There are lots of undocumented APIs, so let's add what's missing.

- [x] AutosaveMonitor [#60882](https://github.com/WordPress/gutenberg/pull/60882), [#60905](https://github.com/WordPress/gutenberg/pull/60905)
- [x] CharacterCount [#60906](https://github.com/WordPress/gutenberg/pull/60906)
- [x] DocumentBar [#61733](https://github.com/WordPress/gutenberg/pull/61733)
- [x] DocumentOutline [#61129](https://github.com/WordPress/gutenberg/pull/61129)
- [x] DocumentOutlineCheck [#61129](https://github.com/WordPress/gutenberg/pull/61129)
- [x] EditorHistoryRedo [#60932](https://github.com/WordPress/gutenberg/pull/60932)
- [x] EditorHistoryUndo [#60932](https://github.com/WordPress/gutenberg/pull/60932)
- [x] EditorKeyboardShortcuts [#60933](https://github.com/WordPress/gutenberg/pull/60933)
- [x] EditorKeyboardShortcutsRegister [#60933](https://github.com/WordPress/gutenberg/pull/60933)
- [x] EditorNotices [#61736](https://github.com/WordPress/gutenberg/pull/61736)
- [x] EditorProvider [#61739](https://github.com/WordPress/gutenberg/pull/61739)
- [x] EditorSnackbars [#61110](https://github.com/WordPress/gutenberg/pull/61110)
- [x] EntitiesSavedStates [#62377](https://github.com/WordPress/gutenberg/pull/62377)
- [x] ErrorBoundary [#61950](https://github.com/WordPress/gutenberg/pull/61950)
- [x] LocalAutosaveMonitor [#61951](https://github.com/WordPress/gutenberg/pull/61951)
- [x] PageAttributesCheck [#60977](https://github.com/WordPress/gutenberg/pull/60977)
- [x] PageAttributesOrder [#60977](https://github.com/WordPress/gutenberg/pull/60977)
- [x] PageAttributesPanel [#60977](https://github.com/WordPress/gutenberg/pull/60977)
- [x] PageAttributesParent [#60977](https://github.com/WordPress/gutenberg/pull/60977)
[...]

PR #61966, opened 2024-05-24
Documentation: PostDiscussionPanel editor component

colorful-tones:
## What? & Why?
Addresses one item in #60358

- mitigates risk

## How?
Add a JSDoc comment to the `PostDiscussionPanel` component and run `npm run docs:build` to populate the `README` with the newly added documents.


Mamaduka:
I think some of these panel components are becoming absolete with recent refactoring and probably will be deprecated. The question is if we still want to document those.

packages/editor/src/components/post-discussion/panel.js
```diff
@@ -86,6 +86,16 @@ function PostDiscussionToggle( { isOpen, onClick } ) {
 	);
 }
 
+/**
+ * Uses the `useSelect` hook to determine if the discussion panel is enabled.
```
ntsekouras: I think we should have a higher level small description of the component and not describe its current implementation, with maybe the exception of mentioning the panel name that can be used to programmatically hide the panel. Something like:

> This component allows to update comment and pingback settings for the current post. Internally there are checks whether the current post has support for the above and if the `discussion-panel` panel is enabled.

ntsekouras:
>I think some of these panel components are becoming absolete with recent refactoring and probably will be deprecated. The question is if we still want to document those.

That's true and if we find components that are only exist for back compat we should deprecate them. Having some high level docs even for them though, until they are removed (if they are removed sometime in the future 😄 ). I think it's fine.

Having said that the specific component in this PR is used in our codebase too.

packages/editor/src/components/post-discussion/panel.js
Mamaduka: Agreed. The implementation details aren't crucial for similar documents.

colorful-tones:
Thanks for all the guidance and clarification @Mamaduka @ntsekouras 👍 
I streamlined the docs.

Mamaduka (approved):
Thanks for the follow-up, @colorful-tones!

PR #61967, opened 2024-05-24
Documentation: PostExcerptPanel editor component

colorful-tones:
## What? & Why?
Addresses one item in #60358

- mitigates risk

## How?
Add a JSDoc comment to the `PostExcerptPanel` component and run `npm run docs:build` to populate the `README` with the newly added documents.


packages/editor/src/components/post-excerpt/panel.js
```diff
@@ -79,6 +79,12 @@ function ExcerptPanel() {
 	);
 }
 
+/**
+ * `PostExcerptPanel` is a wrapper component that uses the `PostExcerptCheck`
```
ntsekouras: Maybe something like:

> Is rendered if the post type supports excerpts and allows editing the excerpt.

We can omit the name at the start of the docs, since it's going to be a bit duplicative in the docs..

Mamaduka (approved):

PR #61138, opened 2024-04-26
Replace classnames with clsx

DaniGuardiola:
PR originally started as a test, see: https://github.com/WordPress/gutenberg/pull/61091#issuecomment-2078823298

Then it replaced #61091, see: https://github.com/WordPress/gutenberg/pull/61091#issuecomment-2088118267

There are no significant changes in this PR that weren't already part of the previous PR, I only synced with trunk and ensured that all usage of classnames was still consistently replaced with cslx.

PR #61969, opened 2024-05-24
Remove useless clsx calls.

DaniGuardiola:
Context: https://github.com/WordPress/gutenberg/pull/61138#discussion_r1586061700

### DIFF
--- a/packages/block-library/src/site-logo/edit.js
+++ b/packages/block-library/src/site-logo/edit.js
@@ -73,7 +73,6 @@ const SiteLogo = ( {
 	const [ { naturalWidth, naturalHeight }, setNaturalSize ] = useState( {} );
 	const [ isEditingImage, setIsEditingImage ] = useState( false );
 	const { toggleSelection } = useDispatch( blockEditorStore );
-	const classes = 'custom-logo-link';
 	const { imageEditing, maxWidth, title } = useSelect( ( select ) => {
 		const settings = select( blockEditorStore ).getSettings();
 		const siteEntities = select( coreStore ).getEntityRecord(
@@ -136,7 +135,7 @@ const SiteLogo = ( {
 			/* eslint-disable jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/click-events-have-key-events */
 			<a
 				href={ siteUrl }
-				className={ classes }
+				className="custom-logo-link"
 				rel="home"
 				title={ title }
 				onClick={ ( event ) => event.preventDefault() }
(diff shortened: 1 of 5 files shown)


fullofcaffeine (approved):
LGTM!

storybook/decorators/with-global-css.js
```diff
@@ -59,7 +54,7 @@ export const WithGlobalCSS = ( Story, context ) => {
 	}, [ context.globals.css ] );
 
 	return (
-		<div className={ clsx( classes ) }>
+		<div className={ classes }>
```
talldan: This change feels risky in comparison to the others, is this variable guaranteed to be a string?

packages/block-library/src/site-logo/edit.js
```diff
@@ -73,7 +73,7 @@ const SiteLogo = ( {
 	const [ { naturalWidth, naturalHeight }, setNaturalSize ] = useState( {} );
 	const [ isEditingImage, setIsEditingImage ] = useState( false );
 	const { toggleSelection } = useDispatch( blockEditorStore );
-	const classes = clsx( 'custom-logo-link' );
+	const classes = 'custom-logo-link';
```
tyxla: I think we can go a step further and just inline such cases without the need to maintain a constant.

packages/components/src/focal-point-picker/focal-point.tsx
```diff
@@ -15,7 +14,7 @@ export default function FocalPoint( {
 	top = '50%',
 	...props
 }: WordPressComponentProps< FocalPointProps, 'div' > ) {
-	const classes = clsx( 'components-focal-point-picker__icon_container' );
+	const classes = 'components-focal-point-picker__icon_container';
```
tyxla: Same as above, I'd inline that.

packages/edit-site/src/components/save-panel/index.js
```diff
@@ -151,9 +151,7 @@ export default function SavePanel() {
 			>
 				<Button
 					variant="secondary"
-					className={ clsx(
-						'edit-site-editor__toggle-save-panel-button'
-					) }
+					className={ 'edit-site-editor__toggle-save-panel-button' }
```
tyxla: Nit: double quotes are preferred when it's a string literal. 

```suggestion
					className="edit-site-editor__toggle-save-panel-button"
```

storybook/decorators/with-global-css.js
tyxla: Yeah, I'd likely keep that `clsx` call just in case, since it's not invoked with a static string literal.

packages/edit-site/src/components/save-panel/index.js
DaniGuardiola: Agreed. I believe there's an autofixable eslint rule for this we can enable. I'll send a separate PR.

storybook/decorators/with-global-css.js
DaniGuardiola: Good catch!

tyxla (approved):
LGTM, thanks 👍

packages/edit-site/src/components/save-panel/index.js
DaniGuardiola: PR: https://github.com/WordPress/gutenberg/pull/62026

Issue #59013, opened 2024-02-14
Remove extraneous elements placed between Tabs and Tab panels


afercia:
### Description

Splitting this out from https://github.com/WordPress/gutenberg/pull/58942/

Recent work improved the consistency of various tabbed interfaces in the editor by using the [Tabs](https://github.com/WordPress/gutenberg/blob/abdde7a6c5ce7f58e3c924b6d3e3ca25005ccd88/packages/components/src/tabs/index.tsx) component. That's a very welcomed improvement.

However, in a few places the Tabs component is used incorrectly. The whole purpose of a tabbed interface is to allow keyboard users to save tab key presses and be able to tab directly from a selected tab to its associated tab panel.

Image courtesy of Heydon Pickering, from [his accessible components](https://inclusive-components.design/tabbed-interfaces/), to visually illustrate the expected keyboard interaction:

![tabs interaction](https://github.com/WordPress/gutenberg/assets/1682452/e3de1cf1-cee3-4e11-808b-59f56d3e189c)

Placing extraneous focusable elements between the tabs and the tab panels defeats the purpose of a tabbed interface.

There's a few places in the editor where a 'Close' button is placed in the between and adds an unexpected tab stop between the tabs and the tab panels. For example, in all the Settings panels:

![Screenshot 2024-02-14 at 12 50 28](https://github.com/WordPress/gutenberg/assets/1682452/be4045f1-21df-48b8-9a75-e918ca2849d3)

![Screenshot 2024-02-14 at 12 50 51](https://github.com/WordPress/gutenberg/assets/1682452/0dcca6aa-a310-47d1-9b5d-7f7b127df212)

![Screenshot 2024-02-14 at 13 52 31](https://github.com/WordPress/gutenberg/assets/1682452/103faaa2-190a-425f-a886-ee8c10cfabc5)

And https://github.com/WordPress/gutenberg/pull/58942/ would reproduce the same problem in the List View panel.

There's two requirements to follow:
- There must be no focusable or otherwise interactive elements between Tabs and Tab panels.
- The visual order must match the DOM order.

As such, the only solution to this is to remove the Close button from that spot and place it somewhere else.

It would be great to have the Tabs component to enforce this via code: no extraneous element can be placed between Tabs and Tab panels.

### Step-by-step reproduction instructions

- Go to the Post editor > Settings panel.
- Use a keyboard.
- Tab to the first tab.
- Use the right or left arrow keys to select a tab.
[...]

Issue #58940, opened 2024-02-12
List View: tabs and close button visual order and DOM order mismatch


afercia:
### Description

In the process of reviewing all the CSS `order` properties in use in the editor, I noticed the List View tabs and close button visual order and DOM order mismatch. For accessibility, the order must always match when it affects 'meaning and operation'. basically, only purely decorative elements can be reordered. Reference:

WCAG 2.2
1.3.2 Meaningful Sequence (Level A) https://www.w3.org/TR/WCAG22/#meaningful-sequence
2.4.3 Focus Order (Level A) https://www.w3.org/TR/WCAG22/#focus-order

Screenshot to illustrate the reading / DOM order mismatch.

![Screenshot 2024-02-12 at 14 50 40](https://github.com/WordPress/gutenberg/assets/1682452/a67411f8-8def-47c1-a66c-dc1b32cd45e9)

In this specific case, these elements are also interactive, which makes the problem even more relevant.

Instead of using `order`, the DOM order must be adjusted to match the desired visual order.

Introduced in https://github.com/WordPress/gutenberg/commit/a3a55bb21718833db88d516bb22edfd6021263ee so this `order` property is new on trunk and should be fixed for the WordPress 6.5 release.
Worth noting that the order mismatch occurs also in previous versione but for a different reason. The Close button used to be absollutely positioned.

The fix appears to be trivial as the button just needs to be moved after the tabs. Code:

https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/components/list-view-sidebar/index.js/#L133-L155

Since it appeaers to be a simple, self-contained fix, I'd like to propose it for WP 6.5 consideration.

### Step-by-step reproduction instructions

- Go to te post editor.
- Open the List VIew.
- Inspect the elements in the list view top part.
- Observe the Close button is first in the DOM but it's visually shown after the tabs.

Test with a screen reader, for example Safari + VoiceOver.
- Open the list view. Initial focus is set on the first item in the list (which is an a11y problem I will report separately).
- Press Control + Option + Left arrow key until you reach toe top part of the list view.
- Observe the first element VoiceOver moves to within the top part of the list view is the 'Outline' tab.
- Press Control + Option + Left arrow key again.
- Observe VoiceOver moves to the 'List View' tab.
- Press Control + Option + Left arrow key again.
- Observe VoiceOver moves to the Close button, unexpectedly jumping to the right.

### Please confirm that you have searched existing issues in the repo.

Yes

[...]

PR #61974, opened 2024-05-24
Try: Improved tabbed sidebar styles

DaniGuardiola:
I've removed the "balanced" tabs styles, as it overrides the built-in Tabs styles - we're trying to avoid and remove style overrides in the components package for many reasons. @afercia discussed tabs styles here: https://github.com/WordPress/gutenberg/issues/62025

If we want these styles to be an option for Tabs, then it should be proposed, discussed and implemented in the Tabs component directly. These styles, though, present some problems beyond aesthetic concerns, notably it could behave badly in other languages since the length of the labels is unpredictable (see @afercia's comment: https://github.com/WordPress/gutenberg/pull/61974#issuecomment-2149719264)

Before:

https://github.com/WordPress/gutenberg/assets/8862967/6756cd14-05fb-4b59-8486-6860f42cc3dd

After:

image

image

---

Note that this PR also removes a few styles that were completely redundant (basically all but the negative bottom margin).


tyxla (commented):
To be fair, I prefer how it was before. 

From what I understand, we try to keep the tabs occupying all the horizontal space, which made even more visual sense when the X button wasn't there. Even with the X button though, it's OK for them to be wider and occupying more space, there's not much else to do with that space and folks are likely used to have larger inserter category tabs.

This is something @WordPress/gutenberg-design should have a good sense about.

jasmussen:
Thanks for your attention to detail. There's an ongoing conversation about these changes, maybe @richtabor has more info.

richtabor:
There is a discrepancy between the tabs in the Inserter and the Inspector, but I am cautious to reduce the surface area of the Inserter tabs. 

Here's the state of the tabs: 

### Current:

| Inserter  | Inspector |
|CleanShot 2024-05-30 at 16 47 38|CleanShot 2024-05-30 at 16 47 45|

CleanShot 2024-05-30 at 17 03 38


### Proposed:

| Inserter  | Inspector |
|CleanShot 2024-05-30 at 17 00 23|CleanShot 2024-05-30 at 17 00 38|

CleanShot 2024-05-30 at 17 01 37

richtabor:
And here's a counter, if we applied the same positioning logic that exists in the Inserter to the Inspector tabs: 

CleanShot 2024-05-30 at 17 04 51

Not a formal proposal; an exploration.

DaniGuardiola:
Thanks for the visuals, very helpful @richtabor. I think I strongly prefer the first option, personally.

afercia:
> There is a discrepancy between the tabs in the Inserter and the Inspector,

Yes, and I reported the inconsistencies in https://github.com/WordPress/gutenberg/issues/62025

To me, tabs should have the same alignment / layout everywhere.

Worth reminding the X close button is candidate to be removed from there, as it breaks the ARIA tabs pattern. See https://github.com/WordPress/gutenberg/issues/59013 / https://github.com/WordPress/gutenberg/issues/58940

As per @DaniGuardiola proposal in this PR, I support the left alignment option. To me, it doesn't make much sense attempting to have a balanced equal width because the length of the tabs content is not predictable. It may look nice in English, which usually has shorter strings, but it's a bit pointless with longer / uneven strings in other languages.

tyxla:
If the consensus is in favor of the left-aligned instead of balanced-width tabs, I'm fine with that 👍

scruffian:
> To me, tabs should have the same alignment / layout everywhere.

This is the aim of the component I am proposing in https://github.com/WordPress/gutenberg/pull/62343

DaniGuardiola:
I've updated the PR after @scruffian's work to unify sidebar, and also edited the description to more clearly state the issues and how this PR addresses it. I'd like to seek some consensus around this to hopefully make a decision.

cc @WordPress/gutenberg-design @WordPress/gutenberg-components @richtabor @tyxla @afercia

jameskoster:
There are also these tabs in the Inspector:

Screenshot 2024-07-25 at 09 46 48

I agree the behavior should be consistent across. The example above would look quite strange if the icons were aligned left.

Looking at the component objectively, centrally aligning the label seems simplest, likely with a prop to specify when the row should fill the container or not.

| Fill | Do not fill |
| --- | --- |
| Screenshot 2024-07-25 at 09 45 11 | Screenshot 2024-07-25 at 09 45 24 |
| Used by: Block settings/styles | Used by: Inserter, Document/Block tabs in the Inspector |

There could also be a prop to left-align the label/icon, but I'm not entirely convinced it's worth the added complexity.

afercia:
> There are also these tabs in the Inspector:

Screenshot 2024-07-25 at 09 46 48

Yes, and the Navigation block shows 3 icon-tabs:

Screenshot 2024-07-25 at 14 15 21

I have a solution for that: use text instead of icons by default 🥳
That would be better for accessibility and would also be consistent with the tabs at the top of the panel.

Worth reminding that enabling the 'Show button text labels' preference already switches those tab icons to text, and the alignment is still inconsistent:

Screenshot 2024-07-25 at 14 18 51

DaniGuardiola:
I think those tabs are a different (but definitely related) issue. This PR focuses on the tabbed sidebar layout, currently only present in inserter and list view, but planned for other similar sidebars across the app (see https://github.com/WordPress/gutenberg/pull/62343 for full context). I personally feel like what needs to be consistent is this component (tabbed sidebar), and it doesn't feel weird if other tabs like that example have a different layout (especially since it is already quite different by having icon labels instead of text).

With that in mind, and if we agree on that, I'm happy to discuss more about those tabs. My hunch is that it might justify a "balanced" variant for icon-based tabs, some notes:

> centrally aligning the label seems simplest

I agree. I think for this "balanced" variant, labels (in this case icons) should indeed be centered.

> Worth reminding that enabling the 'Show button text labels' preference already switches those tab icons to text

Agreed. I think the variant should probably revert to the default "left-aligned" layout when the preference is enabled. Feels like it should be implemented at the usage/product level though, and not at the component level, since the preference is a product-level feature.

afercia:
>  I personally feel like what needs to be consistent is this component (tabbed sidebar), and it doesn't feel weird if other tabs like that example have a different layout

I'd agree the inconsistent layout with other tabs isn't a blocker for this PR, which is an improvement. I'd encourage to moev on with this PR as long as there is a follow-up to better consider consistent alignment for _all_ tabs across the entire editor UI.

tyxla (commented):
I'm generally in favor of the approach here, since it contributes to a more consistent tabs behavior across the editor and less custom styles and overrides. 

However, before we move forward, it would be nice to have a consensus on the design approach, especially as far as those icon-only tabs.

ciampo:
> likely with a prop to specify when the row should fill the container or not.

@jameskoster  Probably something to add to a hypothetical issue on whether `Tabs` should allow for tabs overflowing its layout. We need to find the right balance between using props, and just empowering users of the components to apply their layout preferences via CSS to the individual compound components.

jameskoster:
In terms of design, the question to answer is whether to apply the 'fill' prop in these contexts or not. In other words which is preferred between:

| Fill | No fill |
| --- | --- |
| Screenshot 2024-08-06 at 11 13 22 | Screenshot 2024-08-06 at 11 13 51 |

Personally I wouldn't consider that decision a blocker to this PR.

jasmussen:
Although for icon-only tabs, they should be centered, the centered text in the fullwide tabs feels awkward. That's either an argument for no fill, or for left-aligned in the tabs by default, with the centered option an exception, or a combination of both.

jameskoster:
Ideally we should figure out which options the component should offer and go from there.

We know we want fill-with-centered contents for the block settings toolbar so this should be built into the component.

The fill-with-left-aligned-contents is dubious imo. It only really seems to exist to align the tab first label with the contents below in the Inserter / Inspector. This is an exception, which is fine, but doesn't really make sense as a component option.

To unlock this work I'd propose;

- Update the tab component so consumers are able to use a 'fill' layout. 
  - In this layout the tab contents are centrally aligned.
- By default tabs will 'hug' their contents.
- Update the Inserter tab instance to either:
  - Use the 'fill' layout with an override to left-align the text.
  - Use the default 'hug' layout.

DaniGuardiola:
Thank you for your comments. I like your proposal @jameskoster, though practically speaking if we go for "Update the Inserter tab instance to > Use the default 'hug' layout" then implementing the "fill" variant is not a blocker for this specific PR. I strongly prefer this resolution personally, and it's also the simplest:

image

jameskoster:
The issue with merging this first is that there will be an interim where the inserter tabs are narrower, which could be seen as a visual regression. Personally I think it's okay, so long as we address the other issues reasonably quickly, but I'd welcome more thoughts from @WordPress/gutenberg-design.

DaniGuardiola:
@jameskoster yes, to be clear I'm proposing that we go with that option ("Update the Inserter tab instance to > Use the default 'hug' layout"), not just temporarily. If so, "fill" is not a blocker here, and tabs being narrower is not a regression but just a "permanent" update. I hope that clarifies what I meant?

jameskoster:
> not just temporarily

Right, that's the part we need design consensus on. Is the following acceptable:

Screenshot 2024-08-07 at 12 41 07

It's fine with me, to be honest, but I don't have a full history around the design of this UI. Hopefully @jasmussen can help make this decision.

jasmussen:
Sure let's try it. 

In general we should avoid _centered text_, so I think that does need followups. But all these pieces need followup, so that wouldn't be a blocker.

jameskoster (approved):

tyxla (approved):
Glad to see some consensus on the design side 👍 

Just confirming, while the text tabs are no longer balanced, the tabs with icons remain balanced, which makes sense and looks good IMHO:

Screenshot 2024-08-07 at 16 26 45

Thanks for the cleanup @DaniGuardiola, looks great 🚀

DaniGuardiola:
@WordPress/gutenberg-design "fill" implementation https://github.com/WordPress/gutenberg/pull/64371

cc @jameskoster @jasmussen

PR #61975, opened 2024-05-24
Don't steal focus when opening browse all blocks

### LINKED ISSUE #61361: Navigation block: browse all blocks auto closes without inserting anything
[Originally reported in #core-editor](https://wordpress.slack.com/archives/C02QB2JS7/p1714749597293399) by @talksina, there's a bug with the navigation block in Gutenberg 18.2 where, when trying to search to add additional blocks, the inserter disappears. cc @alexstine for tagging me to investigate.

**Steps to replicate**

1. Create a navigation block, no matter if in post, page or template.
2. Add a new menu item. "add block", the block inserter pops up.
3. Click on "browse all blocks"
4. Type a desired block name in the search field.
5. The inserter closes, without adding anything. 

I can replicate both with clicking and via keyboard navigation.

**Video**

https://github.com/WordPress/gutenberg/assets/26996883/e52effd7-2a41-4d2a-b65a-0a9874e750bc

**Environment**

GB 18.2, WP 6.5.2, TT4

jeryj:
Fixes https://github.com/WordPress/gutenberg/issues/61361

## What?

When using the browse all blocks button from adding a new navigation link block, the inserter would steal focus from the popover (correctly) and then the popover would close, stealing focus back. 

## Why?

Fix using the inserter from the navigabion block.

## How?

To prevent this bug, we check to see if focus is still within the popover at the time of closing. If focus is no longer within it, we don't move focus back to the selected block and instead let whatever stole focus keep the focus.

## Testing Instructions

- Use the navigation link inserter (+ button)
- Select Add Block
- Select Browse All
- Focus should be in the inserter and the link inserter popover should be closed

- Use the navigation link inserter (+ button)
- Select Add Block
- Select Browse All
- Focus should be in the inserter and the link inserter popover should be closed

Escape from Adding block
- Use the navigation link inserter (+ button)
- Select Add Block
- Select Browse All
- Press Escape
- Focus should be on the previous navigation block

### DIFF
--- a/packages/block-library/src/navigation-link/edit.js
+++ b/packages/block-library/src/navigation-link/edit.js
@@ -267,6 +267,7 @@ export default function NavigationLinkEdit( {
 	// Change the label using inspector causes rich text to change focus on firefox.
 	// This is a workaround to keep the focus on the label field when label filed is focused we don't render the rich text.
 	const [ isLabelFieldFocused, setIsLabelFieldFocused ] = useState( false );
+
 	const {
 		isAtMaxNesting,
 		isTopLevelLink,
@@ -587,8 +588,6 @@ export default function NavigationLinkEdit( {
 											window.document.activeElement
 										)
 									) {
-										// Where is focus right now? is it within the modal?
-										// get the active focused element
 										// Select the previous block to keep focus nearby
 										selectPreviousBlock( clientId, true );
 									}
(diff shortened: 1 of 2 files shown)


t-hamano (approved):
From what I've tested, it works as expected.

Adding E2E tests might help, but since I believe this issue is critical, let's ship this PR to WP6.6.

PR #61062, opened 2024-04-25
Components: Make the `ProgressBar` public

fullofcaffeine:
Issue: https://github.com/WordPress/gutenberg/issues/59418.

## What?

This is part of a [larger effort](https://github.com/WordPress/gutenberg/issues/59418) to remove `__experimental` prefix from all "experimental" components, effectively promoting them to regular stable components. See the related issue for more context.

## Why?

The strategy of prefixing exports with `__experimental` has become deprecated after the introduction of private APIs. 

## How?

1. Export it from components without the `__experimental` prefix;
1. Keep the old `__experimental` export for backwards compatibility;
1. Change all imports of the old `__experimental` in GB and components to the one without the prefix (including in storybook stories). Also, update the docs to refer to the new unprefixed component;
1. Add the component storybook `id` (get it from the storybook URL) to the `PREVIOUSLY_EXPERIMENTAL_COMPONENTS` const array in `manager-head.html` so that old experimental story paths are redirected to the new one;
1. Add a changelog for the change.
1. Make the component public if needed, and update consumers to import it directly instead of using the private API.

- [x] Add changelog

PR #53030, opened 2023-07-27
Components: Introduce a basic ProgressBar component

tyxla:
## What?
This PR introduces a basic horizontal `ProgressBar` component.

I'd love to get some design feedback since this is done from scratch and is using some component package colors by default, for consistency with the rest of the components. We're introducing some props to allow easier changing of the colors. 

## Why?
While it can be useful in various places, we're aiming to use this new component to replace the `Spinner` component that is currently utilized in the site editor loading experience. I've started work on this in #53032.

## How?
We're introducing a basic component with tests, stories, and README - all the basic pieces for a new component.

## Testing Instructions
* Run `npm run storybook:dev`
* Navigate to http://localhost:50240/?path=/docs/components-experimental-progressbar--default
* Play with the controls and verify the progress bar works well.
* Verify all tests pass.

### Testing Instructions for Keyboard
None

## Screenshots or screencast 

Indeterminate:

https://github.com/WordPress/gutenberg/assets/8436925/6d5fe03f-f420-4681-92e5-9fda885ce111

Determinate:

https://github.com/WordPress/gutenberg/assets/8436925/d868e20d-8430-4a20-8d48-85fb4854f38a

## ✍️ Dev Note

**Introducing a new experimental `ProgressBar` component**

A new experimental `ProgressBar` component has been introduced to the `@wordpress/components` package. 

The `ProgressBar` component supports two modes: 
* determinate - when a particular progress value has been specified (from 0 to 100)
* indeterminate - when a value hasn't been specified

The component is capable of switching between modes depending on the progress `value` prop that's passed to it.

The primary need and motivation for creating this new component is currently in the Site Editor, where it's already utilized for displaying the loading progress of the editor.

PR #61976, opened 2024-05-24
`ProgressBar`: Simplify default `width` implementation and make it more easily overridable

fullofcaffeine:
Depends on: https://github.com/WordPress/gutenberg/pull/61062.

This is a follow-up to https://github.com/WordPress/gutenberg/pull/61062. See this comment for more details on how the work has been split: https://github.com/WordPress/gutenberg/pull/61062#issuecomment-2130015327. You can also read the discussion there to better understand how we arrived at this solution. You can also check the edit history to see the original description for this PR.

## What?

Better align the default width behavior with the [original component specs ](https://github.com/WordPress/gutenberg/pull/61976#issuecomment-2134537064)and make it easier to override it.

This PR also adds a new story to illustrate a custom width and updates the docs/jsdoc.

## Why?

This progress bar was designed to be of a fixed width, so we should keep it. Existing consumers already expect it, too. However, if there are any scenarios where a custom width size/behavior is required, we should provide the means to do so. 

## How?

* Keep the default width of `160px` for the progress bar. It's meant to have a fixed size, [see this comment](https://github.com/WordPress/gutenberg/pull/61976#issuecomment-2134537064);
* Stop using `max-width` to set it. It's not meant to adapt to the parent size, by default;
* Instead, use the `:where` pseud-element (which has zero specificity) to set a default width. This makes it trivial for consumers to override the width with a custom CSS class, if they wish.

## Testing Instructions

1. All tests should pass;
2. Check the storybook stories for the `ProgressBar`.
3. The component should be rendered the same in existing consumers.


packages/components/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Enhancements
 
 - Components: Make the `ProgressBar` public ([#61062](https://github.com/WordPress/gutenberg/pull/61062)).
+- Components: Improve `ProgressBar` width control ([#61976](https://github.com/WordPress/gutenberg/pull/61976).
```
tyxla: Same as the other PR - this needs to use the existing Enhancements section.

tyxla (commented):
To be fair, I think an extra prop that toggles a particular style rule on or off is a worse API than just relying on a classname or inline style.

I also read the discussion in #61062 but couldn't find where and why we decided to have an `hasUnconstrainedWidth` prop. 

I agree with @jasmussen when [he said](https://github.com/WordPress/gutenberg/pull/61062#issuecomment-2098532169) that we should have a default width for consistency. 

I also think we should provide a way to override the `width`, but maybe a `size/width` prop isn't the best idea since it will prevent the `ProgressBar` from being fluid and flexible. We do provide consumers the option to add a `classname` where they can add `max-width: unset;` or whatever works best for them. I don't think we need an extra prop on top of that. 

So I think we should rely on the `classname`, document how to change the `width` better (including with fluid and non-fluid options), and either:

1. Keep the default `max-width` for consistency
2. Remove the `max-width` and rely on the parent width across the editor. 

As I agree with @jasmussen that a default width makes sense for consistency reasons, I'd lean on option 1. I know @mirka preferred option 2, but I believe this will add additional costs for maintaining consistency (having to specify `max-width` or `width` in multiple places) so I'm curious to hear other thoughts on it.

mirka:
> I think we should rely on the `classname`

The concern I have with the "just override with classname" approach with things like width, is that it's non-obvious _which properties exactly_ need to be overridden. Is it `width`, `max-width`, or both? Plus, as a package we always emphasize that we do not guarantee these styling implementation details to not change. So it's inherently an unstable approach that we'd have to recommend to consumers. Maybe I'm being too idealistic, but it doesn't look clean to me.

For me, it's either:

1. Have a default width for consistency, but provide a way to _safely_ change it. (Which a prop like `hasUnconstrainedWidth` achieves)
2. Have no width constraints.

Though I understand that there are no clear winners here. Both have downsides.

Also for what it's worth, I just realized that the native `<progress>` element has a default width of 160px in Chrome and Firefox, so that makes me feel a bit better about us having the same default width as that.

packages/components/src/progress-bar/styles.ts
```diff
@@ -21,11 +21,12 @@ const animateProgressBar = keyframes( {
 // Width of the indicator for the indeterminate progress bar
 export const INDETERMINATE_TRACK_WIDTH = 50;
 
-export const Track = styled.div`
+export const Track = styled.div< { hasUnconstrainedWidth?: boolean } >`
 	position: relative;
 	overflow: hidden;
 	width: 100%;
```
mirka:
```suggestion
```

Could we remove this too, maybe? Should default to `auto`, which should behave the same in our case.

tyxla:
> Have a default width for consistency, but provide a way to safely change it. (Which a prop like `hasUnconstrainedWidth` achieves)

Seems like this is where we disagree. I view this `hasUnconstrainedWidth` prop as unnecessarily exposing an internal implementation detail. For many folks, the default width will be OK. For others, the `hasUnconstrainedWidth` feels like unnecessary complexity, because they might have to use the prop and still apply custom CSS through a classname in some cases. I don't see why a boolean prop about a style difference makes it a better API, I only see unnecessarily added complexity.

> The concern I have with the "just override with classname" approach with things like width, is that it's non-obvious which properties exactly need to be overridden. Is it width, max-width, or both?

For those who want to customize it, it will be easy to spot what they need to override. Also, we can make it more obvious through docs and storybook. I don't see an issue with that.

> Also for what it's worth, I just realized that the native <progress> element has a default width of 160px in Chrome and Firefox, so that makes me feel a bit better about us having the same default width as that.

Nice. This is yet another good reason to keep it as a reasonable, consistent default.

DaniGuardiola:
My humble opinion is that I don't really see this potential consistency issue as important enough to grant a default fixed size plus this prop API.

Doing a fixed size comes with important trade-offs, and in this case having such a prop just to be able to change the size feels very wrong to me. In some components with content (such as buttons) it makes sense for a default sizing strategy to be `fit-content` ([example from a design system I built in the past](https://atlas.guide.co/?path=/story/components-button--width-examples)).

However, in most cases, the width I would expect by default is always 100%, because that's how block elements behave by default. Even if it isn't 100%, I would at least expect to be able to style them with CSS, directly. Having to pass a prop feels cumbersome, and also comes with a discoverability issue.

I think this is a case of YAGNI, and would strongly suggest not having a default fixed size at all.

Regarding HTML `<progress>`, while it's a very valid reference, I don't think it really matters that much. Take text inputs, for instance. They also come with a fixed size, yet we apply 100% size by default. Not only that, but they have `display: inline-block` by default, which we don't do at all (we do `block` instead). In fact, `<progress>` also does `inline-block`!

If lack of sizing consistency ever becomes a problem worth addressing, I'm sure we'll be able to find a satisfactory solution then. Until then, I'd suggest not committing to this kind of API. I think simpler and more consistent (with other components) should win this one. My two cents!

fullofcaffeine:
Thanks y'all for taking the time to share your thoughts about this!

From what I see, the common theme here is that this API smells, and there are two paths forward:
1. Keep the default `max-width` of 160px without this boolean flag to disable it;
2. Remove the default `max-width` and let it adapt to the space around it with `width: 100%`.

And In both cases, we would also document that if a different default width size is needed, the consumer can use CSS to override it.

I'm thinking that, given the low number of consumers we have now, we could go with 2). Thoughts?

fullofcaffeine:
About this:

> I also read the discussion in https://github.com/WordPress/gutenberg/pull/61062 but couldn't find where and why we decided to have an hasUnconstrainedWidth prop.

It came up in my last call with @mirka. I should probably have mentioned it here in the description. I will do it next time.

jasmussen:
Thanks for sweating the details. To echo [my comment from here](https://github.com/WordPress/gutenberg/pull/61062#issuecomment-2134438092), the progressbar was designed to be a specfic width. It can be contextually larger if you need it, but it's not meant to be smaller than a specific width, lest the actual progress will be so small it's not legible. If you need something smaller, use the spinner.

tyxla:
> Thanks for sweating the details. To echo [my comment from here](https://github.com/WordPress/gutenberg/pull/61062#issuecomment-2134438092), the progressbar was designed to be a specfic width. It can be contextually larger if you need it, but it's not meant to be smaller than a specific width, lest the actual progress will be so small it's not legible. If you need something smaller, use the spinner.

Thanks for adding some context, @jasmussen. The `max-width` was intentionally added as per [your recommendation here](https://github.com/WordPress/gutenberg/pull/53030#issuecomment-1657779723) back in the day, and we never actually added a minimum width, although it was discussed. As stated [above ](https://github.com/WordPress/gutenberg/pull/61976#issuecomment-2133549336) I'm also in favor of keeping a reasonable default (this isn't a generic component library after all), especially given that we've provided reasonable options for overriding and further customization. We may only have to work on documenting those customization options to be clear and straightforward.

tyxla:
> Thanks y'all for taking the time to share your thoughts about this!
> 
> From what I see, the common theme here is that this API smells, and there are two paths forward:
> 
> 1. Keep the default `max-width` of 160px without this boolean flag to disable it;
> 2. Remove the default `max-width` and let it adapt to the space around it with `width: 100%`.
> 
> And In both cases, we would also document that if a different default width size is needed, the consumer can use CSS to override it.
> 
> I'm thinking that, given the low number of consumers we have now, we could go with 2). Thoughts?

As agreed with @jasmussen and stated [above](https://github.com/WordPress/gutenberg/pull/61976#issuecomment-2133549336) and [here](https://github.com/WordPress/gutenberg/pull/61976#issuecomment-2134864823) I'm leaning towards option 1), which is more or less what we currently have.

DaniGuardiola:
If we're going with default fixed size then I'd strongly suggest simply setting a width using :where for proper, uncomplicated overridability. No min or max width, they wouldn't be necessary. This also more or less mimics what native components with default width (such as text inputs or progress) do.

I've successfully used this strategy in the past for overridable default fixed sizes in components with total success :)

mirka:
> If we're going with default fixed size then I'd strongly suggest simply setting a width using :where for proper, uncomplicated overridability. No min or max width, they wouldn't be necessary.

I can agree to this 👍 Let's go with this then.

@fullofcaffeine Again, thank you for your patience here 🙏

tyxla:
Sounds like a good plan to me as well 👍

packages/components/src/progress-bar/styles.ts
```diff
@@ -38,6 +36,10 @@ export const Track = styled.div`
 	// Windows high contrast mode.
 	outline: 2px solid transparent;
 	outline-offset: 2px;
+
+	:where( & ) {
```
fullofcaffeine, line 40: @DaniGuardiola Is that what you had in mind when you suggested using `:where`? AFAIK, the main benefit is the zero specificity, right?

packages/components/src/progress-bar/stories/index.story.tsx
```diff
@@ -29,3 +29,34 @@ const Template: StoryFn< typeof ProgressBar > = ( { ...args } ) => {
...
+ * ```
+ */
+export const WithCustomWidth: StoryFn< typeof ProgressBar > = () => {
+	return (
+		<>
+			<style>{ withCustomWidthCustomCSS }</style>
+			<ProgressBar className="custom-progress-bar" />
+		</>
+	);
+};
```
mirka: Here is a cleaner way we could do this, which keeps the autogenerated code snippet clean:

```suggestion
export const WithCustomWidth = Template.bind( {} );
WithCustomWidth.args = {
	className: 'custom-progress-bar',
};
WithCustomWidth.decorators = [
	( Story ) => (
		<>
			<style>{ withCustomWidthCustomCSS }</style>
			<Story />
		</>
	),
];
```

Another thing we need to be careful about is to make sure all the controls still work when adding a new story. So we can't just do `const NewStory = () => (<Foo />)`, but instead need to pass down props like `const NewStory = (props) => (<Foo {...props} />)`.

packages/components/src/progress-bar/stories/index.story.tsx
```diff
@@ -29,3 +29,34 @@ const Template: StoryFn< typeof ProgressBar > = ( { ...args } ) => {
...
+	.custom-progress-bar {
+		width: 100%;
+	}
+`;
+
+/**
+ * A progress bar with a custom overriden width.
+ *
+ * You can override the default `width` or further customize the behavior
+ * and visuals by passing a custom CSS class via the `cssName` prop.
```
mirka:
```suggestion
 * You can override the default `width` by passing a custom CSS class via the `className` prop.
```

I don't want to recommend _too much_ customization 🫣

mirka (approved):
I think all our deliberation paid off! We thought through every detail, and I'd say we landed on the best solution 🎉 (Fingers crossed!)

Thank you everyone for the great collaboration.

packages/components/src/progress-bar/stories/index.story.tsx
fullofcaffeine: > Here is a cleaner way we could do this, which keeps the autogenerated code snippet clean:

Ah, nice! Thanks for the suggestion. It's nice that we can just inherit from the main story.

> we can't just do const NewStory = () => (<Foo />), but instead need to pass down props like const NewStory = (props) => (<Foo {...props} />).

Ah, true, I had done this in past iterations, but forgot here. Thanks for catching that!

packages/components/src/progress-bar/stories/index.story.tsx
fullofcaffeine: Fair enough :)

packages/components/src/progress-bar/stories/index.story.tsx
```diff
@@ -29,3 +29,37 @@ const Template: StoryFn< typeof ProgressBar > = ( { ...args } ) => {
...
 Default.args = {};
+
+const withCustomWidthCustomCSS = `
+	.custom-progress-bar {
+		width: 100%;
+	}
+`;
+
+/**
+ * A progress bar with a custom overriden width.
```
tyxla: We can remove some redundancy:

```suggestion
 * A progress bar with a custom width.
```

tyxla (approved):
Looking good 👍 🚀

packages/components/src/progress-bar/styles.ts
```diff
@@ -38,6 +36,10 @@ export const Track = styled.div`
...
+		width: 160px;
```
DaniGuardiola, line 41: Not specific to this PR, but is there any reason we're not using `rem` almost everywhere? It is good practice for proper zoom support, which is an important accessibility feature. @WordPress/gutenberg-components

packages/components/src/progress-bar/stories/index.story.tsx
```diff
@@ -29,3 +29,37 @@ const Template: StoryFn< typeof ProgressBar > = ( { ...args } ) => {
...
+ * .custom-progress-bar {
+ *   width: 100%;
+ * }
+ * ```
+ */
+export const WithCustomWidth = Template.bind( {} );
+WithCustomWidth.args = {
+	className: 'custom-progress-bar',
+};
+WithCustomWidth.decorators = [
```
DaniGuardiola, line 59: Another thing that's not specific to this PR, but I wonder why we're still using a legacy CSF version, even for new stuff. Haven't checked our current version, but CSF 3 has been supported for the last couple of versions and it makes Storybook much easier to work with. @WordPress/gutenberg-components

DaniGuardiola (approved):
Looks good to me! Pending @tyxla's comments, I think this is ready to merge.

fullofcaffeine:
Thanks for the collab and thorough review folks! ❤️

packages/components/src/progress-bar/styles.ts
mirka: I've always assumed that this was a "that ship has sailed" situation, since WordPress/Gutenberg is already built entirely on a non-rem basis 😓

packages/components/src/progress-bar/stories/index.story.tsx
mirka: We probably should! Nobody hadn't really prioritized it enough.

PR #61980, opened 2024-05-25
Shadow Panel: Improve a11y and fix browser console error

t-hamano:
## What?

This PR makes the following two changes in the shadow editing panel.

1. Resolve browser warning errors regarding `ariaExpanded`
  image
2. Remove invalid `arialabel` attributes
  arialabel

## Why? / How?

> Resolve browser warning errors regarding `ariaExpanded`

`aria-*` HTML should be hyphen-cased ([Source](https://legacy.reactjs.org/docs/accessibility.html#wai-aria)).

> Remove invalid `arialabel` attributes

This attribute itself should not be needed, since the button has proper text.

## Testing Instructions

- Go to the Site Editor > Global Styles > Shadows > Natural.
- Confirm that no browser warnings are displayed.
- Confirm that the toggle button's `aria-expanded` attribute toggles when the shadow popup is opened and closed.


madhusudhand (approved):
Good catch. Thank you for your work on this.

Changes looks good! 🎉

t-hamano:
Thanks for the review!

PR #56241, opened 2023-11-17
DataViews: remove title from pages

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/55083

## What?

Removes the title from the pages.

| Page | Before | After |
| --- | --- | --- |
| Templates | Captura de ecrã 2023-11-17, às 11 33 54 | Captura de ecrã 2023-11-17, às 11 33 22 |
| Pages | Captura de ecrã 2023-11-17, às 11 33 46 | Captura de ecrã 2023-11-17, às 11 33 33 |

## Why?

The page title is redundant, as the sidebar already anchors the user, telling them where they are. Removing the title leaves also more space for the main interaction of the page (filtering, search, pagination, etc.).

## How?

Removes the title from the `<Page>` component.

## Testing Instructions

- Enable the "admin views" experiment and visit the "Manage all templates" and "Manage all pages" pages in "Appaerance > Editor".
- See how it looks like.

Issue #60468, opened 2024-04-04
Improve UX around filter customisation of bundled data views


jameskoster:
Based on the discussion below, let's update the UX conventions around customising views like so:

- [x] Filters that are part of the view config (for example `status = trash` in the 'Trash' view) are locked – they cannot be edited or removed. They can potentially be hidden from the UI, as they are in pattern category views.
- [x] https://github.com/WordPress/gutenberg/issues/60472
- [x] When additional filters are added, the 'Reset' button should restore to view to its original state rather than removing all filters as it does now.
- [ ] When additional filters are added, mark the view as modified:

<img width="900" alt="modified" src="https://github.com/WordPress/gutenberg/assets/846565/5d1f5e41-84ea-4fa2-8f56-0ec4078eee74">

- [ ] Add actions for modified views in the sidebar; 'Reset', 'Save' (for user-generated views), 'Save as...' (for bundled views):

<img width="900" alt="actions" src="https://github.com/WordPress/gutenberg/assets/846565/52d09a93-8fdd-456d-819a-8254f3cb88a2">

---

### Original issue

It can sometimes be confusing when you edit the filters associated with a bundled view so that the visible records no longer reflect the purpose of the view.
 
A great example of this is the "Trash" view in page management. If you remove the status filter in this view you see _all_ pages, yet the active view suggests you're viewing trash:

https://github.com/WordPress/gutenberg/assets/846565/382c6a96-b47f-4fca-9e13-da8e197f6a3e

jameskoster:
Three options to consider:

### 1. Lock default filters for bundled data views
Perhaps too restrictive, but one option could be to simply lock the filters associated with a view. So in the Trash example users simply wouldn't be able to modify or remove the status filter.

### 2. De-select active view once modified
Another option could be to deselect the active view once it has been customised. This would eliminate some of the confusion, but potentially introduce more as you effectively end up in 'limbo'

### 3. Mark views as modified
Thinking further ahead to custom views, this option would involve marking views as customised somehow, so the user is able to understand _why_ the records on display may not match the view label. Such a UI could facilitate the creation of custom views down the road, as well as a reset affordance.

Initially it could be purely informational, with a simple reset action:

<img width="900" alt="customised-view" src="https://github.com/WordPress/gutenberg/assets/846565/2281eaea-66b3-4f07-a696-3e7432dd69da">

Later it could be a place for other view actions to live:

[...]

PR #61983, opened 2024-05-25
DataViews Pages: Dynamic Frame title as per active view for Pages and Templates editor pages

### LINKED ISSUE #56199: DataViews: Frame title should match current view
For instance in this view:

Screenshot 2023-11-16 at 10 22 04

The frame title should be "Drafts" instead of "Pages":

Screenshot 2023-11-16 at 10 23 09

Souptik2001:
## What?

In DataList views for pages, add a subtitle to display the activeView.

## Why?

Closes: #56199 

This requirement is first mentioned in the above issue.

But a more refined requirement can be found in this comment - https://github.com/WordPress/gutenberg/pull/56241#issuecomment-1816762258

Just looking at the whitespace of the DataList Views of pages, it might be a little unclear what activeView you are on. That's why we can add a sub-title below the title "Pages" to display the activeView over there.

## How?

This PR adds a sub-title for DataViews screen for pages, which displays the activeView.

## Testing Instructions

Please see the below screencast section or refer the steps below.

- Open the site-editor.
- Go to "Pages".
- Now switch the active view by clicking on the left sidebar.
- As you change the views you will see that the subtitle under the title "Pages" also changes accordingly.

N/A

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/39427067/1019c127-dcf1-42db-9b99-0159163b54b6

**OLD VIDEO:**

https://github.com/WordPress/gutenberg/assets/39427067/94712b02-8a93-4527-a957-7c93f2816073


jameskoster:
Thank you for opening the PR :) 

I think it probably makes sense to use the main title, rather than a subtitle, at least initially. This would align with the Pattern management page, and leave the subtitle free for additional context to be added later if needed.

What do you think?

Souptik2001:
> Thank you for opening the PR :)
> 
> I think it probably makes sense to use the main title, rather than a subtitle, at least initially. This would align with the Pattern management page, and leave the subtitle free for additional context to be added later if needed.
> 
> What do you think?

@jameskoster Yes I align with your thought! That would be better!

I made it like this as per this comment - https://github.com/WordPress/gutenberg/pull/56241#issuecomment-1816762258

Just taking a confirmation, that should I change it?

jameskoster:
> Just taking a confirmation, that should I change it?

I think so, yes! Thank you.

Souptik2001:
Thanks @jameskoster for confirming!

I have made the update! Thanks!

jameskoster:
This works well for me. I think it would make sense to apply the same treatment to the templates page. What do you think?

Souptik2001:
Thanks @jameskoster! Yes we can also update templates to have the same behaviour! 👍 

I have made that change!
I have also updated the video in the description, you can check now for both page and template.
Also I have updated the PR title to make it more appropriate.

jameskoster:
Thanks for updating the Templates page :)

There is a quirk (as I think you mentioned on the issue): if you reset the filters on a bundled page, the title can be misleading. E.g. Go to trashed pages, remove the status filter, and you're no longer viewing trashed pages:

Screenshot 2024-05-30 at 11 09 31

This will be fixed when https://github.com/WordPress/gutenberg/issues/60468 is addressed. So the question here is whether we should include this change in the upcoming 6.6 release, or whether we should consider #60468 a blocker.

The main argument for including it is to establish alignment with the Patterns page, where the frame title matches the menu item:

Screenshot 2024-05-30 at 11 16 50


I'm keen to gather thoughts from others on that cc @WordPress/gutenberg-design.

t-hamano:
> There is a quirk (as I think you mentioned on the issue): if you reset the filters on a bundled page, the title can be misleading. E.g. Go to trashed pages, remove the status filter, and you're no longer viewing trashed pages:

To avoid this discrepancy, how about excluding the sidebar items from the data view filter?

For example, on a page, if you select "Draft" in the sidebar, the data view will only list posts with a Draft status. However, it does not add any filters and we cannot reset the post status in the DataViews.

jameskoster:
@t-hamano that sounds like the first point in https://github.com/WordPress/gutenberg/issues/60468. I think the question is whether or not we consider that a blocker to this PR.

t-hamano:
Ah, I didn't read that issue thoroughly enough 😅 Personally, I would lean towards not including this PR in 6.6 and moving the discussion forward in #60468.

Souptik2001:
@jameskoster @t-hamano I can get that task rolling. 🙂 

If I understood correctly, for this first version we just want a visual feedback of whether the view is editor or not right(as per the design is provided)? Like the "Save view" feature, etc. will be introduced later, correct?

If you give the go-ahead I can work on that one. 🙂 
Thanks!

jameskoster:
Yup, I think the first three tasks in https://github.com/WordPress/gutenberg/issues/60468 are related, and can potentially be tackled in a single PR. The fourth and fifth tasks can come later 👍

Souptik2001:
Thanks @jameskoster! I have raised a PR for first three points of #60468 !

PR #61994, opened 2024-05-26
Update matchers.js DOM Text Interpreted As HTML

Shivam7-1:
## What?
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

## Why?

This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

### DIFF
--- a/packages/blocks/src/api/matchers.js
+++ b/packages/blocks/src/api/matchers.js
@@ -43,7 +43,7 @@ export function html( selector, multilineTag ) {
 			return value;
 		}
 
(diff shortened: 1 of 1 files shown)


Mamaduka:
As far as I know, the current behavior is intentional. Closing this as not-applicable.

Issue #61987, opened 2024-05-25
Add UI to Global styles to allow modification of font size presets


matiasbenedetto:
## What?
Add UI to Global styles to allow modification of font size presets

## Why?
This will allow users to edit the font size presets using the editor.

matiasbenedetto:
I started a potential implementation of this here: https://github.com/WordPress/gutenberg/pull/62328

annezazu:
Just wanted to bring in feedback as the PR above gets underway that being able to control the font sizes on smaller devices is a major responsiveness related request I hear from folks. The latest look showing how you can [toggle on an option to set custom fluid sizes](https://github.com/WordPress/gutenberg/pull/62328#issuecomment-2182672359) will go a long way in addressing that! Excited to see this land and to get some testing done.

Issue #61986, opened 2024-05-25
Add UI to Global styles to allow managing spacing presets


matiasbenedetto:
## What?
Add UI to Global styles to allow modification and creation of spacing scale preset values.

## Why?
This will allow users to edit the spacing presets using the editor.

christopherabalos:
My apologies if this has been covered somewhere else but I wanted to mention it on this issue due to cross concerns.

I make use of clamp in spacings. It's been working great to dynamically adjust content density. If this could be accounted for and handled exactly how font sizes work, it would be amazing. This would require an adjustment to theme.json schema to accept a min and max value. The UI would then be very similar to the new fonts UI.

t-hamano:
This issue has been punted to 6.8 and has not seen any activity since. Let's move this issue to the 6.9 Editor Task Board and see what can be done in 6.9.

ryanwelcher:
This issues hasn't seen any activity since 6.7. Is it still relevant?

juanfra:
I believe this is still relevant. It feels like one of the last remaining pieces of UI needed to fully unlock website presentation management. I’m not sure if this will make it into 6.9, but I do think it’s worth keeping on the radar. We could likely build on what was done for font presets. The main challenge will be figuring out how to handle themes that define spacing with `clamp()`, which means designing a suitable UI control for that.

ryanwelcher:
> I believe this is still relevant. It feels like one of the last remaining pieces of UI needed to fully unlock website presentation management. I’m not sure if this will make it into 6.9, but I do think it’s worth keeping on the radar. We could likely build on what was done for font presets. The main challenge will be figuring out how to handle themes that define spacing with `clamp()`, which means designing a suitable UI control for that.

Thanks for the reply! I think I should have been more clear, I was not advocating for closure rather to see if this would be part of 6.9 which unless someone picks it up, it will most likely not be.

t-hamano:
> if this would be part of 6.9 which unless someone picks it up, it will most likely not be.

I agree, I think this task is a "nice to have" and not required for 6.9.

juanfra:
Thanks for clarifying! That makes sense and I agree.

karmatosed:
I am happy to look at this now it has needs design on it. I will try and do so within next few days to at least get a first pass.

khoipro:
Let's me collab with @karmatosed to audit this issue.

t-hamano:
[...]

PR #61997, opened 2024-05-26
Shadow Panel: Generates unique shadow slugs by finding max suffix and incrementing it

### LINKED ISSUE #61989: Shadow Panel: slugs may conflict
### Description

The shadow slug uses the current number of shadows plus 1 as a suffix.

https://github.com/WordPress/gutenberg/blob/278810c1b987a6926fc41248f567d8bfa4230b70/packages/edit-site/src/components/global-styles/shadows-panel.js#L87

But with this logic, slugs can be duplicates.

I think we need to find the maximum value from all the shadow suffixes and add 1 to that value, [like in the `PaletteEdit` component](https://github.com/WordPress/gutenberg/blob/278810c1b987a6926fc41248f567d8bfa4230b70/packages/components/src/palette-edit/index.tsx#L92-L103).

- Add three custom shadows. The labels are "Shadow 1", "Shadow 2", and "Shadow3" respectively.
- Delete "Shadow 2".
- Add shadow.
- A shadow labeled "Shadow 3" is added. This will cause a browser warning error because the slug matches the existing "Shadow 3".
- Also, editing one will affect the other.

https://github.com/WordPress/gutenberg/assets/54422211/9517b4c7-7a4f-4afe-956a-840ec87ee0b9

### Environment info

_No response_

Yes

Yes

amitraj2203:
## What?
Fixes: #61989 

## Why?
Shadow slugs were getting duplicated when we try to delete any custom shadow and add a new. See [this](https://github.com/WordPress/gutenberg/issues/61989#issue-2317416856) for more details.

## How?
This PR adds a function to generate unique names and slugs for shadows by finding the highest suffix number used and incrementing it.

## Testing Instructions
1. Add three custom shadows. The labels are "Shadow 1", "Shadow 2", and "Shadow3" respectively.
2. Delete "Shadow 2".
3. Add a new shadow.
4. Now the new shadow will have unique slug and name.

## Screenshots or screencast 
https://github.com/WordPress/gutenberg/assets/77401999/d8ebb043-16e9-423b-a314-880606dafce6

### DIFF
--- a/packages/edit-site/src/components/global-styles/utils.js
+++ b/packages/edit-site/src/components/global-styles/utils.js
@@ -11,6 +11,22 @@ export function getVariationClassName( variation ) {
 	return `is-style-${ variation }`;
 }
 
+export function getNewIndexFromPresets( presets, slugPrefix ) {
+	const nameRegex = new RegExp( `^${ slugPrefix }([\\d]+)$` );
+	return presets.reduce( ( previousValue, currentValue ) => {
+		if ( typeof currentValue?.slug === 'string' ) {
+			const matches = currentValue?.slug.match( nameRegex );
+			if ( matches ) {
+				const id = parseInt( matches[ 1 ], 10 );
+				if ( id >= previousValue ) {
+					return id + 1;
+				}
+			}
+		}
+		return previousValue;
+	}, 1 );
+}
+
 function getFontFamilyFromSetting( fontFamilies, setting ) {
 	if ( ! Array.isArray( fontFamilies ) || ! setting ) {
 		return null;
--- a/packages/edit-site/src/components/global-styles/utils.js
+++ b/packages/edit-site/src/components/global-styles/utils.js
@@ -11,6 +11,14 @@ export function getVariationClassName( variation ) {
 	return `is-style-${ variation }`;
 }
 
+/**
+ * Returns a new index based on the presets and slug prefix.
+ *
+ * @param {Array}  presets    The array of presets.
+ * @param {string} slugPrefix The prefix for the slug.
+ *
+ * @return {number} The new index.
+ */
 export function getNewIndexFromPresets( presets, slugPrefix ) {
 	const nameRegex = new RegExp( `^${ slugPrefix }([\\d]+)$` );
 	return presets.reduce( ( previousValue, currentValue ) => {
--- a/packages/edit-site/src/components/global-styles/utils.js
+++ b/packages/edit-site/src/components/global-styles/utils.js
@@ -12,27 +12,30 @@ export function getVariationClassName( variation ) {
 }
 
 /**
- * Returns a new index based on the presets and slug prefix.
+ * Iterates through the presets array and searches for slugs that start with the specified
+ * slugPrefix followed by a numerical suffix. It identifies the highest numerical suffix found
+ * and returns one greater than the highest found suffix, ensuring that the new index is unique.
  *
- * @param {Array}  presets    The array of presets.
- * @param {string} slugPrefix The prefix for the slug.
+ * @param {Array}  presets    The array of preset objects, each potentially containing a slug property.
+ * @param {string} slugPrefix The prefix to look for in the preset slugs.
  *
- * @return {number} The new index.
+ * @return {number} The next available index for a preset with the specified slug prefix, or 1 if no matching slugs are found.
  */
 export function getNewIndexFromPresets( presets, slugPrefix ) {
 	const nameRegex = new RegExp( `^${ slugPrefix }([\\d]+)$` );
-	return presets.reduce( ( previousValue, currentValue ) => {
-		if ( typeof currentValue?.slug === 'string' ) {
-			const matches = currentValue?.slug.match( nameRegex );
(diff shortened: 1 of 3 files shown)


t-hamano (commented):
Thanks for the PR!

I think the approach makes sense, but personally I think this function needs to be made more generic. Similar problems will be faced when the UI for customizing [font size presets](https://github.com/WordPress/gutenberg/issues/61987) and [spacing scale values](https://github.com/WordPress/gutenberg/issues/61986) are implemented.

I haven't tested it, but what about a function that simply returns just the `index`, like below? This allows us to apply it to other presets in the future.

```javascript
export function getNewIndexFromPresets( presets, slugPrefix ) {
	const nameRegex = new RegExp( `^${ slugPrefix }([\\d]+)$` );
	return presets.reduce( ( previousValue, currentValue ) => {
		if ( typeof currentValue?.slug === 'string' ) {
			const matches = currentValue?.slug.match( nameRegex );
			if ( matches ) {
				const id = parseInt( matches[ 1 ], 10 );
				if ( id >= previousValue ) {
					return id + 1;
				}
			}
		}
		return previousValue;
	}, 1 );
}

const handleAddShadow = () => {
	const newIndex = getNewIndexFromPresets( shadows, 'shadow-' );
	onCreate( {
		name: sprintf(
			/* translators: %s: is an index for a preset */
			__( 'Shadow %s' ),
			newIndex
		),
		slug: `shadow-${ newIndex }`,
		shadow: defaultShadow,
	} );
};
```

Additionally, we may want to consider:

- Considering the possibility that this function will be used for general purposes in the future, move the function to `packages/edit-site/src/components/global-styles/utils.js`.
- Add JSDoc to the function
- Add some unit tests to `packages/edit-site/src/components/global-styles/test`

amitraj2203:
Hi @t-hamano, as suggested I have created a generic function for getting the `Index` and added some unit tests. Let me know if there are any further improvements you'd like to make.

t-hamano:
@amitraj2203 

Thanks for the update!

@madhusudhand @jorgefilipecosta What do you think about this approach? I'm sure you've implemented shadow UI, so I'd love to hear your advice 🙇‍♂️

packages/edit-site/src/components/global-styles/utils.js
```diff
@@ -11,6 +11,30 @@ export function getVariationClassName( variation ) {
...
 
+/**
+ * Returns a new index based on the presets and slug prefix.
+ *
+ * @param {Array}  presets    The array of presets.
+ * @param {string} slugPrefix The prefix for the slug.
+ *
+ * @return {number} The new index.
+ */
+export function getNewIndexFromPresets( presets, slugPrefix ) {
```
talldan: The logic in the function looks good to me - find the highest preset and increment it by one, nice work putting that together and adding the associated tests.

I think the docblock could be a little more descriptive about what the function does. It mentions 'new index', but the reader has no idea what kind of value that new index will be. It should be possible to understand exactly what the function returns without reading any of the code.

packages/edit-site/src/components/global-styles/utils.js
```diff
@@ -11,6 +11,30 @@ export function getVariationClassName( variation ) {
...
+			const matches = currentValue?.slug.match( nameRegex );
+			if ( matches ) {
+				const id = parseInt( matches[ 1 ], 10 );
+				if ( id >= previousValue ) {
+					return id + 1;
+				}
+			}
+		}
+		return previousValue;
+	}, 1 );
```
talldan: Minor optional feedback (it doesn't need to be addressed), but I think the function can be more self documenting with some small changes to the variable naming and if the increment happens at the end:
```jsx
	const highestPresetValue = presets.reduce( ( currentHighest, preset ) => {
		if ( typeof preset?.slug === 'string' ) {
			const matches = preset?.slug.match( nameRegex );
			if ( matches ) {
				const id = parseInt( matches[ 1 ], 10 );
				if ( id > currentHighest ) {
					return id;
				}
			}
		}
		return currentHighest;
	}, 0 );
	return highestPresetValue + 1;
```

I haven't tested it though, so not sure if this code still works 😄

packages/edit-site/src/components/global-styles/utils.js
amitraj2203: Acknowledged! I've updated the docblock to provide a clearer description of the function's purpose and its return value.

packages/edit-site/src/components/global-styles/utils.js
amitraj2203: Thanks for the review. I completely agree with your suggestion and also I've tested these changes and it is working as expected.

t-hamano (approved):
LGTM! I think it's working as expected.

https://github.com/WordPress/gutenberg/assets/54422211/8a59500f-0dc0-499d-bc54-28252c9f186c

Could you rebase this PR with the latest trunk? Then the failing GitHub Actions should pass.

I would be happy if you could give us a second re-review, @talldan.

talldan (approved):
Looks good to me too, thanks for your contribution 🎉 

I'll go ahead and merge the PR, as I don't think it will have had an impact on bundle size.

PR #61271, opened 2024-05-01
Background image: add support for relative theme path URLs in top-level theme.json styles

ramonjd:
## Dev Note 📓 

https://github.com/WordPress/gutenberg/pull/59354#issuecomment-2138583222

## What?

This PR resolves relative paths in top-level `background.backgroundImage.url` values in theme.json and theme variation JSON files using [get_theme_file_uri](https://developer.wordpress.org/reference/functions/get_theme_file_uri/).

Part of:

- #54336

Based on explorations in:

- https://github.com/WordPress/gutenberg/pull/61401

Related trac ticket:

- https://core.trac.wordpress.org/ticket/61205

> [!NOTE]
> This PR doesn't enable theme relative path resolution for blocks in templates/patterns. That'll be a follow up.

## Why?

To ensure that themes with background image values in theme.json are portable — a theme developer can set a relative path to an image asset as a background image value, and that path will resolve correctly to absolute path of the file in the theme directory no matter where the theme is installed.

Contrast this with absolute image paths: every installation of the theme would point to single image files sitting somewhere on the internet.

## How?

Paths to images residing in the theme directory can be set in theme.son using the `file:` syntax, [a convention already used in Core for web fonts](https://developer.wordpress.org/themes/global-settings-and-styles/settings/typography/#registering-web-fonts-font-faces).

The path is relative to the theme root. For example:

```json
	"styles": {
		"background": {
			"backgroundImage": {
				"url": "file:./img/untitled.png"
			}
		}
	}
```

#### For the frontend:

Relative paths will be resolved when building the global styles sheet in `gutenberg_get_global_stylesheet()`.

A new public method `WP_Theme_JSON_Resolver_Gutenberg::resolve_theme_file_uris` accepts a `WP_Theme_JSON_Gutenberg` object and will resolves any `file:` paths it finds using [get_theme_file_uri](https://developer.wordpress.org/reference/functions/get_theme_file_uri/).

#### For the editor:

Global styles are built using merged theme, core and user theme.json configs in the block editor. All this is done using Javascript. 

So that the editor has access to the resolved paths before building editor global styles, Gutenberg resolves the file paths in the backend and passes them to the browser via the `_links` objects in global styles controller REST responses (Global styles and Global styles revisions controllers).

Global styles are built using the `useGlobalStylesOutputWithConfig` hook. This PR adds a mutating function to that hook that replaces the relative paths with their absolute equivalents using the data in the `_links` object.

[...]

PR #59354, opened 2024-02-26
Global styles: add background image to top-level theme.json styles

ramonjd:
## Dev Note 📓 

https://github.com/WordPress/gutenberg/pull/59354#issuecomment-2138583222

## What?

Part of:

- https://github.com/WordPress/gutenberg/issues/54336

Enabling background image block supports at the root level of global styles.

### TODO
- [x] Check santization and parsing. Could we assume relative path to the theme and use `get_theme_file_uri( $path )`?
- [x] Unit tests
- [x] Ensure styles appear in site editor (New PR?)
- [ ] Documentation
- [x] Add UI controls to global styles (top level) https://github.com/WordPress/gutenberg/pull/59454

## How?
Add `style.background` properties as valid top-level styles in theme.json.

## Testing Instructions

Check this branch out and enable emptytheme for ease of testing.

Below are some test theme.jsons. 

Ensure that the `background-*` CSS appears in the  body selector and that everything looks fine. 

```json
{
	"$schema": "../../schemas/json/theme.json",
	"version": 2,
	"styles": {
		"color": {
			"background": "tomato"
		},
		"background": {
			"backgroundImage": {
				"url": "https://i0.wp.com/wordpress.org/files/2023/12/sotw-dotorg-drawer.png?w=1807&ssl=1",
				"source": "file"
			},
			"backgroundPosition": "center center",
			"backgroundRepeat": "repeat",
			"backgroundSize": "cover"
		}
	}
}
```

```json
{
	"$schema": "../../schemas/json/theme.json",
	"version": 2,
	"styles": {
		"color": {
			"background": "tomato"
		},
		"background": {
			"backgroundImage": "url('https://i0.wp.com/wordpress.org/files/2023/12/sotw-dotorg-drawer.png?w=1807&ssl=1')",
			"backgroundPosition": "center center",
			"backgroundRepeat": "no-repeat",
			"backgroundSize": "cover"
		}
	}
}
```

Issue #54336, opened 2023-09-11
Background Image block support follow-up tasks


andrewserong:
Related to #16479, following on from #14744.

Now that the background image block support has been introduced in #53934, this issue captures some ideas for follow-up tasks and features.

## Prioritized tasks for WP 6.8 and beyond
* [x] https://github.com/WordPress/gutenberg/issues/60401 - WIP https://github.com/WordPress/gutenberg/pull/60739
   * [ ] https://github.com/WordPress/gutenberg/issues/32787 - Get the block support to work with gradient backgrounds in a logical way. See related comment: https://github.com/WordPress/gutenberg/pull/59354#issuecomment-1975782474 
   * [ ] Update Core safecss_filter_attr/kses to support linear gradient/rgb(a). See: https://github.com/WordPress/gutenberg/pull/60739#discussion_r1575567729
* [ ] https://github.com/WordPress/gutenberg/issues/60939
* [ ] Allow for background-attachment and other controls to be disabled via theme.json (?)
* [ ] https://github.com/WordPress/gutenberg/issues/63096
    - https://github.com/WordPress/gutenberg/pull/63216
    - https://github.com/WordPress/gutenberg/pull/64242
    - https://github.com/WordPress/gutenberg/pull/65304
    - https://github.com/WordPress/gutenberg/pull/66736
* [ ] https://github.com/WordPress/gutenberg/issues/62254
* [ ] Add support for relative theme paths in variation styles and variation block styles. See:  https://github.com/WordPress/gutenberg/pull/60100/files#r1639128861
* [ ] Fix default controls support for background support and panel https://github.com/WordPress/gutenberg/pull/68085
* [ ] Add ability to skip background support serialization in line with other block supports

## Prioritized tasks for WP 6.7
* [x] https://github.com/WordPress/gutenberg/pull/64565
* [x] https://github.com/WordPress/gutenberg/pull/64128
* [x] https://github.com/WordPress/gutenberg/pull/64328
* [x] https://github.com/WordPress/gutenberg/pull/64192
* [x] https://github.com/WordPress/gutenberg/pull/64182
* [x] https://github.com/WordPress/gutenberg/pull/60151
    * [x] https://github.com/WordPress/gutenberg/pull/63225 
* [x] https://github.com/WordPress/gutenberg/pull/61382  - adds support for [background-attachment ](https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment) 
[...]

PR #61998, opened 2024-05-27
Global styles: add option to remove site-wide theme background image

ramonjd:
## Dev Note 📓 

https://github.com/WordPress/gutenberg/pull/59354#issuecomment-2138583222

## What?

Part of:

- https://github.com/WordPress/gutenberg/issues/54336

Hello.

This PR implements @andrewserong's idea and adds functionality to remove a theme's default, site-wide background image in the editor.

See: https://github.com/WordPress/gutenberg/pull/61271#issuecomment-2095236976

## Why?

Check it out: 

https://github.com/WordPress/gutenberg/assets/6458278/6b76f775-7813-49a5-b220-1c402ad9e137

There's no way to remove a background image if it's been defined in theme.json.

## How?

By adding a "Remove" option to the image dropdown that sets the value of `background.backgroundImage` to `"none"`, when:

1. A user value has not been set, and
2. There's a theme.json value

## Testing Instructions

1. Define a background image for your site in theme.json
2. Head over to the site editor **Global Styles > Layout**
3. Under **Background**, click on the image control and then click on "Remove"
4. The background image should disappear! Save and check the frontend. The value of `background-image` should be `"none"`. 
5. In the site editor again, reset the background image value by clicking on the image control and then click on "Reset"
6. You should see the default background image again.

Here's some test theme.json!

```json
{
	"version": 3,
	"settings": {
		"appearanceTools": true
	},
	"styles": {
		"background": {
			"backgroundImage": {
				"url": "https://images.pexels.com/photos/22484288/pexels-photo-22484288/free-photo-of-the-circular-stone-terraces-of-the-inca-ruins.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr=2"
			}
		}
	}
}
```

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/6458278/476e9125-c031-4183-b257-688b350ce5c4

### DIFF
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -83,6 +83,8 @@ export function hasBackgroundSizeValue( style ) {
 export function hasBackgroundImageValue( style ) {
 	return (
 		!! style?.background?.backgroundImage?.id ||
+		// Supports url() string values in theme.json.
+		'string' === typeof style?.background?.backgroundImage ||
 		!! style?.background?.backgroundImage?.url
 	);
 }
@@ -279,6 +281,23 @@ function BackgroundImageToolsPanelItem( {
 
 	const hasValue = hasBackgroundImageValue( style );
 
+	const closeAndFocus = () => {
+		const [ toggleButton ] = focus.tabbable.find(
+			replaceContainerRef.current
+		);
+		// Focus the toggle button and close the dropdown menu.
+		// This ensures similar behaviour as to selecting an image, where the dropdown is
+		// closed and focus is redirected to the dropdown toggle button.
+		toggleButton?.focus();
+		toggleButton?.click();
+	};
+
+	const onRemove = () =>
+		onChange(
+			setImmutably( style, [ 'background', 'backgroundImage' ], 'none' )
+		);
+	const canRemove = ! hasValue && hasBackgroundImageValue( inheritedValue );
+
 	return (
 		<ToolsPanelItem
 			className="single-column"
@@ -311,17 +330,20 @@ function BackgroundImageToolsPanelItem( {
 					}
 					variant="secondary"
 				>
+					{ canRemove && (
+						<MenuItem
+							onClick={ () => {
+								closeAndFocus();
+								onRemove();
+							} }
+						>
+							{ __( 'Remove' ) }
+						</MenuItem>
+					) }
 					{ hasValue && (
 						<MenuItem
 							onClick={ () => {
-								const [ toggleButton ] = focus.tabbable.find(
-									replaceContainerRef.current
-								);
-								// Focus the toggle button and close the dropdown menu.
-								// This ensures similar behaviour as to selecting an image, where the dropdown is
-								// closed and focus is redirected to the dropdown toggle button.
-								toggleButton?.focus();
-								toggleButton?.click();
+								closeAndFocus();
 								resetBackgroundImage();
 							} }
 						>


packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -279,6 +281,23 @@ function BackgroundImageToolsPanelItem( {
 
 	const hasValue = hasBackgroundImageValue( style );
 
+	const closeAndFocus = () => {
```
ramonjd, line 284: An abstraction since we need this in two places

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -83,6 +83,8 @@ export function hasBackgroundSizeValue( style ) {
 export function hasBackgroundImageValue( style ) {
 	return (
 		!! style?.background?.backgroundImage?.id ||
+		// Supports url() string values in theme.json.
+		'string' === typeof style?.background?.backgroundImage ||
```
ramonjd, line 87: This is so we can detect strings values with `url()`

```json

	"styles": {
		"background": {
			"backgroundImage": "url('https://images.pexels.com/photos/22484288/pexels-photo-22484288/free-photo-of-the-circular-stone-terraces-of-the-inca-ruins.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr=2')"
		}
	}

```

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -311,17 +330,20 @@ function BackgroundImageToolsPanelItem( {
...
 					variant="secondary"
 				>
+					{ canRemove && (
+						<MenuItem
+							onClick={ () => {
+								closeAndFocus();
+								onRemove();
+							} }
+						>
+							{ __( 'Remove' ) }
```
ramonjd, line 340: Is "Remove" clear enough? 

It should only appear where there's a theme.json value AND there's no user style defined - basically so users can remove the default theme background image and nothing else.

tellthemachines: Yeah that's a good question. I think visually it's pretty clear when you open the dropdown by clicking on tile with the image name and thumbnail; in context "Remove" could only be related to that image. But a quick test with VoiceOver shows the experience is pretty flaky if you don't have the visual context: you tab past a menu popup button named "Background options" (this is the only clue you get that you're in a section related to background), then you tab into another menu popup button labeled with the the image file name, plus "Selected image: Untitled". That's a wider problem than this PR addresses though, so I'm not suggesting tying to solve it here 😅 

Really tools panels were never accessible to begin with, which is the main issue, but perhaps for this particular instance we could improve the experience by switching the order of file name and "Selected image: Untitled", and maybe changing "Selected image" to "Site background image" (I'm not sure where the "Untitled" is coming from? but maybe we could dispense with it altogether?). But given this is an existing issue it should probably be addressed in a follow-up.

tellthemachines (approved):
Code LGTM and it's working well in testing! Plus it's a really useful enhancement, because currently there's no way of getting rid of a theme background image short of editing the theme files 😅

ramonjd:
Thanks for the feedback and testing @tellthemachines 

> But a quick test with VoiceOver shows the experience is pretty flaky if you don't have the visual context
> we could improve the experience by switching the order of file name and "Selected image: Untitled", and maybe changing "Selected image" to "Site background image" (I'm not sure where the "Untitled" is coming from? but maybe we could dispense with it altogether?). But given this is an existing issue it should probably be addressed in a follow-up.

Good idea. "untitled" is the fallback where the image has no title.

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/global-styles/background-panel.js#L305

I have a [tweak PR](https://github.com/WordPress/gutenberg/pull/62000) going so I can update that over there.

Cheers!

jasmussen:
Good starting point, I'm assuming we can also reset it using the toolspanel mechanism as is the case for the group bg option?

Screenshot 2024-05-27 at 08 23 43

There may be a larger question, across all design-tools and perhaps related to the inheritance visualization work, to improve how items are generally reset to their theme.json definitions. It's ultimately about user customizations. The curious piece here is that you may be resetting an image, and are then surprised to still see an image, expecting there to be no image. The closest equivalent to mimic is likely going to be having an explicit "no background image" option, similar to how we have explicit text-decoration definitions, or explicit "no color" definitions.

tellthemachines:
> I'm assuming we can also reset it using the toolspanel mechanism as is the case for the group bg option?

If the background image is defined by the theme (if comes from the theme.json) then the tools panel reset shouldn't remove it, right? At least I always assumed that "reset" means returning to whatever the theme defaults are.

jasmussen:
> At least I always assumed that "reset" means returning to whatever the theme defaults are.

Global styles should always show all tools in a tools panel, and yes, clicking reset should reset to "default", which would be whatever default theme.json provides.

So I'm not suggesting an action item, I'm mainly thinking through what the longer term solution is for adding an explicit "none" option background images, similar to explicit "no background color" or explicit "no text-decoration" options, for the case where you want to unset a default. What we have in this branch is good, just wondering if there are clearer labels: "remove" and "reset" are a bit confusing. I wonder if "Reset" and "No background image" would be clearer?

Issue #30725, opened 2021-04-12
Feedback: It is confusing that adding background color also adds padding.


carolinan:
During a user testing session on April 9, several users asked why adding a background color to a block also adds padding, this result was unexpected for them.

Comments included:
_Coloring the background of my text, the text automatically moves to the right. I can’t get it back._

_Why does the column change size when changing the color in the settings? At least it definitely seemed like it happened that way. That’s unnecessary and unexpected_

annezazu:
[The fifth call for testing for the FSE program](https://make.wordpress.org/test/2021/04/14/fse-program-testing-call-5-query-quest/#comment-1501) also led to similar feedback being shared:

> padding added when applying a background seems quite extreme. Could we provide an option to customise the padding? – https://d.pr/i/HQQoof/D4ciOBn8gY

I imagine this will be implemented as more padding controls are explored but wanted to pass this feedback along!

carolinan:
Yes I wonder if both margin and padding would be enabled by default for all blocks.

paaljoachim:
Associated:
Full Site Editor: Applying background color also updates content padding (possible duplicate?)
https://github.com/WordPress/gutenberg/issues/36586

beckej13820:
Just want to chime in here with a +1. The confusing thing about this behavior for me is that the padding changed, but wasn't communicated to me in any way. I'm not sure it should work this way, but even if it was, it would be really nice if somehow the new padding number would be displayed to me. 

Because my user experience was padding was blank before there was a background color and it was blank after there was a background color.

kathrynwp:
Definitely agree that this is confusing behaviour. It came up in this forum thread, when a user was confused by the large amount of padding added automatically the moment a background colour was put on a paragraph block within a column:

https://wordpress.org/support/topic/how-can-i-make-all-blocks-within-a-row-with-one-height/

annezazu:
This continues to come up in the FSE Outreach Program as well, [including the fifteenth call for testing:](https://make.wordpress.org/test/2022/07/11/fse-program-testing-call-15-category-customization/#comment-2612)

>I noticed when changing the background color of the post title (in the query block), it meaningfully change the spacing as well (and made the title take up many lines).

Since we have more dimension controls, it would be interesting to iterate on this!

richiecarey:
+1 on this ... I was very confused why adding a background color to a column would also add padding.

annezazu:
[...]

PR #61999, opened 2024-05-27
Styles: Allow global styles padding to apply to blocks with background

aaronrobertshaw:
Related:
- https://github.com/WordPress/gutenberg/pull/61638

## What?

This is a follow-up to https://github.com/WordPress/gutenberg/pull/61638 to allow global styles set padding for blocks with backgrounds to apply.

## Why?

Currently, the Paragraph and List blocks apply default padding when those blocks have a background color set. This is done by selectors that have a higher specificity for these blocks than the global styles. The result is the global styles set padding is overridden.

Theme authors using theme.json or users attempting to tweak spacing via global styles shouldn't have to resort to hacks setting custom css for these blocks to achieve their desired result e.g. 

```json
	"core/paragraph": {
		"css": ".has-background {padding: 20px 30px;}"
	},
```

## Important

This change, if accepted, could mean global styles rules might unintentionally override individual paragraph blocks that have a background color set. More context can be found in https://github.com/WordPress/gutenberg/pull/61638#issuecomment-2132528042. As noted there, this PR would likely deliver the _expected_ behaviour and can be considered a "fix".

> I'd imagine if someone has set the list or paragraph blocks to have a particular value in global styles they might actually want it to override that default .has-background padding

## How?

- Modifies the `p.has-background` and `ul.has-background, ol.has-background` selectors so they have `0-1-0` specificity to match global styles after #61638 lands.
- As global styles are loaded after the block library styles, global styles would now apply correctly

## Testing Instructions

1. Add Paragraph and List blocks to the page
2. Apply a background color to each of the blocks
3. Confirm the default `.has-background` padding for each is applied
4. In Global Styles, set padding styles for the Paragraph and List block types
5. Confirm the global styles padding values are now applied to the blocks


carolinan:
There is a 3 year old issue about removing the automatic padding that is added when the background color is added, but it is not so simple. https://github.com/WordPress/gutenberg/issues/30725

aaronrobertshaw:
Closing this as an alternative approach has now been merged to controlling padding on classes that get default padding when they get a background color applied. 

See https://github.com/WordPress/gutenberg/pull/82024

PR #61382, opened 2024-05-06
Background: add background attachment to top level styles

ramonjd:
## What?

Adds a toggle to allow background images to scroll with the page, or remain fixed in position.

Part of:

- https://github.com/WordPress/gutenberg/issues/54336

Core backport PR:

- https://github.com/WordPress/wordpress-develop/pull/6991

## Why?

To increase the possibilities of background images.

## How?

Adding support for `background-attachment`.

## TODO

- [x] Add backport PR
- [x] Get advice on placement/design

## Testing Instructions

1. Insert a Group or Verse block.
2. Add a background image to the block
3. Check that image scrolls with the page by default
4. Toggle the "Scroll with page" option off.
5. Test that the background image is now fixed.
6. Save and check the frontend
7. Repeat with a site background image in the Site Editor (under Styles > Layout) 

### Testing Instructions for Keyboard

1. Ensure you can tab through the sidebar controls to the background attachment toggle
2. To activate/deactivate hit the space bar

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/6458278/8067a8ce-f2ad-4101-a470-33bae4dd6015

Issue #60939, opened 2024-04-22
Background image: should it integrate with background-image theme mod?


ramonjd:
### Site background images in WordPress pre-block editor

in this issue a "site background image" refer to an image set via the `background-image` CSS property to the body tag.

WordPress themes have long been able to set site background images and other background properties such as size and repeat (on the HTML body element) either by the Customizer UI or, more directly, using `custom-background` theme support. For example:

```php
add_theme_support(
    'custom-background',
    array(
        // As well as default image, there's default-position-x, default-position-y, default-size, default-repeat, default-attachment
        'default-image' => get_stylesheet_directory_uri() . '/img/bird.jpg',
    )
);
```

Or like this:

```php
add_theme_support( 'custom-background' );
// As well as background_image, there's background_preset, background_size, background_position_x, background_position_y
set_theme_mod( 'background_image',  get_stylesheet_directory_uri() . '/img/bird.jpg' );
```

Enabling background support adds a link to the Customizer in the admin menu. Users can add/edit the background image using Customizer controls:

| Link in sidebar | Customizer controls |
|--------|--------|
|  <img width="150" alt="Screenshot 2024-04-22 at 1 44 09 pm" src="https://github.com/WordPress/gutenberg/assets/6458278/a88d3066-22ba-46da-9b9f-05c88d4eb6d1"> | <img width="150" alt="Screenshot 2024-04-22 at 1 43 36 pm" src="https://github.com/WordPress/gutenberg/assets/6458278/63053d7b-17c7-402e-be51-6aab7928c398"> | 

### Site background images in block themes/block editor

Since https://github.com/WordPress/gutenberg/pull/59354 and https://github.com/WordPress/gutenberg/pull/59454, users can add site background images via theme.json or global styles in the site editor.

However, if a site background image is set via theme support/mods, it will overwrite any theme.json or global styles image set in the site editor because of CSS specificity, that is `body.custom-background` will trump `:where(body)`.

<img width="644" alt="Screenshot 2024-04-22 at 11 50 44 am" src="https://github.com/WordPress/gutenberg/assets/6458278/25716cab-b83b-4036-8b57-c2605d76bc9f">

## What is the question? Or your proposed solution?

The main question is whether Gutenberg should take values set by theme support/mods into account by:

- using any `background-` values from theme support/mods as defaults for top-level background property values.
[...]

PR #62000, opened 2024-05-27
Background image: update controls defaults and layout

### LINKED ISSUE #61928: Background image tool: Improvements, bugfixes, find alternative to "fixed"
There's a neat new background image tool in global styles, allowing you to place a site wide background:

site wide bg

The tool is similar to that of Group or Cover background images:

| Site  | Group |
| site  | group  |

It diverges in a few places, and both can benefit from a few improvements:

* In both cases, the tool should be called "Background image". It's currently just "Background" on the site wide option.
* At the moment adding a background image to either have different defaults. Site wide it defaults to "Fixed". This is likely a bug, because it holds helptext that suggests of "Cover". Both should default to "Cover".

Finally, the "Fixed" option is easy to confuse with the "Fixed background" property of the Cover block:

fixed background

One is a fixed size, another is a fixed background attachment. In both cases, the word "Fixed" is the source of confusion. What's a good option there?

One option is to rephrase the option to "tile":

tile option

* "Cover" would remain as is, and default. Same with "Contain".
* Tile would still reveal width and repeat controls, but would these would be output as new row, with a "width" label, and instead of the help text saying "Specify a fixed width." it would say "Image has a fixed width." 
* A default width of 50% would be assigned. This would also immediately have a visual effect in the canvas, whereas now, you have to enter a number first. Percent may also be a neater default unit, than px.

[Figma](https://www.figma.com/design/hrrqkZCCRmbO8A2VX0aKun/Background-Image?node-id=0%3A1&t=8nMtTZX4wik8S2K5-1).

ramonjd:
## What?

Maybe resolves https://github.com/WordPress/gutenberg/issues/61928

This PR:

- harmonizes headings between block supports and global styles
- updates copy including changing the label "Fixed" to "Tile"
- uses horizontal layout for image size and repeat
- fixes bug where width input is cleared and the size defaults to "cover"
- tidies up the visually hidden context labels

## Why?

Trying to bring some consistency to the controls.

## Testing Instructions
... To come

## Screenshots or screencast 
Screenshot 2024-05-27 at 3 28 10 pm

### DIFF
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -172,13 +172,13 @@ function InspectorImagePreview( { label, filename, url: imgUrl } ) {
 						{ imgLabel }
 					</Truncate>
 					<VisuallyHidden as="span">
-						{ filename
+						{ imgUrl
 							? sprintf(
 									/* translators: %s: file name */
-									__( 'Selected image: %s' ),
-									filename
+									__( 'Background image: %s' ),
+									filename || imgLabel
 							  )
-							: __( 'No image selected' ) }
+							: __( 'No background image selected' ) }
 					</VisuallyHidden>
 				</FlexItem>
 			</HStack>
@@ -302,7 +302,7 @@ function BackgroundImageToolsPanelItem( {
 					name={
 						<InspectorImagePreview
 							label={ title }
-							filename={ title || __( 'Untitled' ) }
+							filename={ title }
 							url={ getResolvedThemeFilePath(
 								url,
 								themeFileURIs
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -100,7 +100,7 @@ function backgroundSizeHelpText( value ) {
 	if ( value === 'contain' ) {
 		return __( 'Image is contained without distortion.' );
 	}
-	return __( 'Specify a fixed width.' );
+	return __( 'Image has a fixed width.' );
 }
 
 /**
@@ -421,6 +421,14 @@ function BackgroundSizeToolsPanelItem( {
 			nextRepeat = undefined;
 		}
 
+		/*
+		 * Next will be null when the input is cleared,
+		 * in which case the value should be 'auto'.
+		 */
+		if ( ! next && currentValueForToggle === 'auto' ) {
+			next = 'auto';
+		}
+
 		onChange(
 			setImmutably( style, [ 'background' ], {
 				...style?.background,
@@ -484,7 +492,9 @@ function BackgroundSizeToolsPanelItem( {
 				value={ currentValueForToggle }
 				onChange={ updateBackgroundSize }
 				isBlock
-				help={ backgroundSizeHelpText( sizeValue ) }
+				help={ backgroundSizeHelpText(
+					sizeValue || defaultValues?.backgroundSize
+				) }
 			>
 				<ToggleGroupControlOption
 					key={ 'cover' }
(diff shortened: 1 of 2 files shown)


jasmussen (approved):
Fast work, big improvement. Bug is fixed, and this is coherent with the Group option as well:

group block background image in inspector

[As I argue in this comment](https://github.com/WordPress/gutenberg/issues/61928#issuecomment-2132746466), I think we should also still change the default from Tile to Cover. The short summary of why is: until we have background-size and background-position tools to enable 2x resolution, and top/centered as opposed to top/left tiling position, "Tile" as a default is not going to be a good or expected experience. 

That's not a blocker, so we can discuss this separately if need be.

andrewserong (approved):
This is testing well for me, too, and adds a nice level of polish. (Great idea @jasmussen to use the name Tile, too, it fits much better than Fixed, and allows for a true fixed option further down the track).

> I think we should also still change the default from Tile to Cover ... That's not a blocker, so we can discuss this separately if need be.

On this one, I still feel like Cover for site wide backgrounds as a default risks leading users astray, so IMO a separate PR for switching the defaults would be a good way to go if we're planning to change it, as that way we've got the change clearly documented in its own PR.

This one LGTM! 🚀

ramonjd:
I appreciate the test and feedback, folks!

> I still feel like Cover for site wide backgrounds as a default risks leading users astray

I'm also for leaving it set to "auto" by default, but for different reasons. 

For theme.json, I think we should be wary of inserting any default values that a user might not expect. My view is that theme.json should primarily represent positive intent, that is, "turn stuff on", rather than override opinionated CMS rules.

ramonjd:
> For theme.json, I think we should be wary of inserting any default values that a user might not expect.

Oh, and the backwards compat "fun" there'd be if folks ever wanted to change/roll back the default. 🤣

jasmussen:
Thanks for moving fast on this one, also carefully. I did mention that I think Cover may be a better default experience, but as it turns out, I was misremembering aspects of this implementation, which in more in-depth testing reveals that, well, there probably really isn't any good default here, unless we add another bit of bundled default code, which I'll get back to. 

My instinct was for Cover to solve these issues:

* It solves the low resolution issue without adding a background-size tool, and without precluding such a tool landing in the future.
* It solves the problem of defaulting to top/left alignment, where arguably most sites are top/center aligned, without adding a background-position tool.
* It solves the problem of images being huge on mobile, small on desktop, because it doesn't scale across.

Turns out I was thinking of the behavior of _Contain_, and even in that case, it only works with Contain + repeat, and then only for landscape images. Here's a case where that works well:

contain-desktop

contain-mobile

But even that falls apart if the image is portrait oriented. At that point it's still top/left aligned, but tiles horizontally.

For reference, here's the current default, Tile, which as noted has its own problems:
tile

tile-mobile

I'll dive back into #62046 with another suggestion, yet another new default!

burnuser:
"Cover" would be the best default, when used like in good old Customizer as cover the screen (not the whole webpage!) in combination with "background-attachment: fixed"
(In fact I've never seen a website with a scrolling background image in real live.)

So, if there is no comparable solution to Customizer default, this feature is - in my opinion - not ready for WP 6.6

ramonjd:
Thanks for the input @burnuser 🙇🏻 

> "Cover" would be the best default, when used like in good old Customizer as cover the screen (not the whole webpage!) in combination with "background-attachment: fixed"

Background attachment controls are in the works:

- https://github.com/WordPress/gutenberg/pull/61382

But it won't be part of 6.6 unfortunately. I'm working on this stuff as fast as I can 😄 

As for the `background-size: cover` as the default, it's already the default for blocks (Group Block), but not for the site. It can be set in theme.json by specifying `"backgroundSize": "cover"`, or via the Site Editor tools however.

In my opinion, we should be cautious about assuming defaults in theme.json in the absence of any values. 

In the case of background size, the CSS default is `"auto"`. For WordPress to assume "cover" universally might look fine for some images, but it won't for all. Furthermore, if, for any reason, that default needs to change in the future, then WordPress would have to continue to honor "cover" due to backwards compatibility forever. 

> So, if there is no comparable solution to Customizer default, this feature is - in my opinion - not ready for WP 6.6

There's a discussion going on how best background images in block themes can play with the current Customizer tool.

- https://github.com/WordPress/gutenberg/issues/60939

The current state is that background images set via the Customizer will take precedence over those set in theme.json due to CSS specificity, so you can continue to use it without any interference from this new feature.

PR #62010, opened 2024-05-27
Dataviews: Remove unused dependencies

sirreal:
## What?

Remove some apparently unused dependencies form the dataviews package.

## Why?

I noticed that the dataviews package claims more dependencies than it seems to use in code.

## Testing Instructions

Does this PR build and work as expected? Search for the removed dependencies in the dataviews package.


oandregal (approved):
I don't see them in use, either.

PR #59702, opened 2024-03-08
Implement pattern overrides behind `IS_GUTENBERG_PLUGIN` flag

talldan:
## What?
See https://github.com/WordPress/gutenberg/issues/53705#issuecomment-1983576920.

Puts the pattern overrides feature behind the `IS_GUTENBERG_PLUGIN` flag to prevent it shipping to core.

## How?
I've taken the approach of using `IS_GUTENBERG_PLUGIN` where the `core/pattern-overrides` block bindings source is registered.

In other places in the codebase, we check for the existence of that `core/pattern-overrides` block binding source to disable the feature.

## Testing Instructions
This needs some thorough testing to ensure all traces of pattern overrides are no longer present.

To replicate `IS_GUTENBERG_PLUGIN` being `false` you can do the following:
- Edit the `webpack/shared.js` file, changing this line to `'process.env.IS_GUTENBERG_PLUGIN': false`. You can usually also modify this in the project package.json itself, but for me this caused a build failure.
https://github.com/WordPress/gutenberg/blob/ed7edbcc3db3155831239092830557df21a73b3a/tools/webpack/shared.js#L67-L68 
- Edit the `lib/load.php` file, changing this line to `define( 'IS_GUTENBERG_PLUGIN', false );`:
https://github.com/WordPress/gutenberg/blob/ed7edbcc3db3155831239092830557df21a73b3a/lib/load.php#L12

Then finally rebuild.

PR #62011, opened 2024-05-27
Remove `IS_GUTENBERG_PLUGIN` check to ensure pattern overrides ship in 6.6.

talldan:
Reverses #59702

## What?
In the build up to WordPress 6.5, the pattern overrides feature was bumped from the release.

The `IS_GUTENBERG_PLUGIN` flag was utilized for this. In both PHP and JS the `pattern overrides` source would only be registered in the gutenberg plugin.

## Why?
The plan is to ship it in 6.6.

## How?
Removes the use of `IS_GUTENBERG_PLUGIN`. While #59702 added the flag in both `lib/load.php` and `packages/editor/src/bindings/index.js`, the `lib/load.php` flag has already been removed in another change prior to this PR.

Only the `js` flag remains, which this PR removes.

### Prerequisites
1. In the root `package.json`, change the `IS_GUTENBERG_PLUGIN` property to `false`.
2. Build from scratch (if you have a watch task running, stop and restart it)

### Test pattern overrides
3. Insert some paragraph/image/button/heading blocks into the post editor
4. Select them and from the block options menu choose 'Create pattern
5. Name the pattern and make it synced
6. Click 'Edit original' from the block toolbar
7. Select one of the paragraph/image/button/heading blocks and in the advanced section of the inspector choose 'Enable Overrides'
8. In the resulting modal, Name the block and click the Enable button
9. Save and select the back button from the editor top bar
10. Back in the post, change the block that you enabled overrides for
11. Preview the post
12. Check that the changes are reflected

### DIFF
--- a/packages/editor/src/bindings/index.js
+++ b/packages/editor/src/bindings/index.js
@@ -12,7 +12,4 @@ import postMeta from './post-meta';
 
 const { registerBlockBindingsSource } = unlock( dispatch( blocksStore ) );
 registerBlockBindingsSource( postMeta );
-
-if ( globalThis.IS_GUTENBERG_PLUGIN ) {
-	registerBlockBindingsSource( patternOverrides );
-}
+registerBlockBindingsSource( patternOverrides );


SantosGuillamot (approved):
Looks good to me 🙂 

Should we also remove [this check](https://github.com/WordPress/gutenberg/blob/remove/pattern-overrides-feature-flag/packages/blocks/src/api/parser/convert-legacy-block.js#L80-L111) in `conver-legacy-block.js` file? (I am not familiar with that, so I am not sure)

talldan:
> Should we also remove [this check](https://github.com/WordPress/gutenberg/blob/remove/pattern-overrides-feature-flag/packages/blocks/src/api/parser/convert-legacy-block.js#L80-L111) in conver-legacy-block.js file? (I am not familiar with that, so I am not sure)

This is back compat for something that was only ever shipped in the Gutenberg plugin, so it doesn't need to be shipped to core. Eventually we can remove that code completely from the plugin.

PR #62012, opened 2024-05-27
Post Sticky Toggle: Improve the design

youknowriad:
## What?

Looking at the design of the Post Sticky control in the post editor sidebar, it looks completely broken and out of place compared to the other controls. This PR tries to bring some harmony there.

That said, there's some a11y challenges there, so in order to keep the design correctness while also keeping things accessible, I used a visually hidden label.

Before:

Screenshot 2024-05-27 at 9 30 00 AM

After:

Screenshot 2024-05-27 at 9 29 13 AM

## Testing Instructions

 - Open the post editor sidebar.
 - Check that you can toggle the "sticky" control both visually and using a screen reader.
 - Check that it looks properly like shown in the screenshot.

### DIFF
--- a/packages/editor/src/components/post-sticky/index.js
+++ b/packages/editor/src/components/post-sticky/index.js
@@ -2,7 +2,7 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { CheckboxControl } from '@wordpress/components';
+import { ToggleControl, VisuallyHidden } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
 
 /**
@@ -10,6 +10,7 @@ import { useDispatch, useSelect } from '@wordpress/data';
  */
 import PostStickyCheck from './check';
 import { store as editorStore } from '../../store';
+import PostPanelRow from '../post-panel-row';
 
 export default function PostSticky() {
 	const postSticky = useSelect( ( select ) => {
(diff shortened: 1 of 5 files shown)


ntsekouras (approved):
Looks good, thanks!

packages/editor/src/components/post-sticky/index.js
```diff
@@ -21,12 +22,18 @@ export default function PostSticky() {
...
-				checked={ postSticky }
-				onChange={ () => editPost( { sticky: ! postSticky } ) }
-			/>
+			<PostPanelRow label={ __( 'Sticky' ) }>
+				<ToggleControl
+					className="editor-post-sticky__toggle-control"
+					label={
+						<VisuallyHidden>
+							{ __( 'Stick to the top of the blog' ) }
+						</VisuallyHidden>
```
Mamaduka: I thought most of the components supported the `hideLabelFromVision` prop.

Mamaduka (approved):
Looks good, but it would be nice to get design feedback as well.

packages/editor/src/components/post-sticky/index.js
youknowriad: Mmm good point, I could update the component to support this prop. 🤔

Mamaduka: It's not a blocker for this PR. Let's create a separate issue for it.

joedolson:
If the visual control label is "Sticky", then the accessible name of the control should also be "Sticky". Having a different visually hidden label makes the control harder to use for voice command. Additionally, WCAG 2.5.3 Label in Name requires that the accessible name of a control must contain the visual label, which this does not.

If we believe that "Sticky" is sufficient labeling to convey the purpose of this control for sighted users, then it is also sufficient for users of screen readers.

huzaifaalmesbah:
The PR looks good. I believe this enhancement will offer users a more pleasant and improved experience. Thanks!

jameskoster:
This seems like a step in the right direction in that it conforms to the updated summary panel format.
 
I do wonder if it remains a little too prominent though. Something to explore separately could be moving it to the Status dialog.

PR #61795, opened 2024-05-20
DataViews: Type all the filters components

youknowriad:
Related #55083 
Follow-up to #61185

## What?

The DataViews package is a types heavy package. There's a lot of structures that need to be documented properly. So this continues the effort to add explicit types to the package. This PR focuses on typing the several filters components we have. This is probably the latest PR before typing the root "DataViews" component.

## Testing instructions

There should be no real change here, it's mostly code quality and documentation.

PR #62013, opened 2024-05-27
Changelog: add note about removing legacy operators

oandregal:
Follow-up to https://github.com/WordPress/gutenberg/pull/61795

Adds a changelog note to dataviews so consumers are aware of important changes. Namely, we removed the legacy support for `in` and `notIn` operators that were replaced by `is` and `isNot`, respectively.


youknowriad (approved):
Thanks :)

PR #62015, opened 2024-05-27
Block Bindings: Filter pattern overrides source in bindings panel

SantosGuillamot:
## What?
Don't show the bindings connected to `core/pattern-overrides` in the bindings panel.

## Why?
Pattern overrides have their own UI, so they don't need to be in the panel. At least right now.

## How?
I just filtered the bindings connected to that source. In the future, we should explore if we want to abstract it or is it okay to keep it specific to pattern overrides.

## Testing Instructions
1. Add a new synced pattern.
2. Add a paragraph and allow overrides in the Advanced panel.
3. After selecting the paragraph, check that the bindings panel is not present.

### DIFF
--- a/packages/block-editor/src/hooks/block-bindings.js
+++ b/packages/block-editor/src/hooks/block-bindings.js
@@ -33,6 +33,19 @@ export const BlockBindingsPanel = ( { metadata } ) => {
 		return null;
 	}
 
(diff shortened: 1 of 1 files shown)


jasmussen (approved):
Works well, thank you!

talldan (approved):
LGTM

PR #62016, opened 2024-05-27
Editor: Remove extra div container and unify the container between post and site editors

youknowriad:
## What?

Small refactoring PR that removes and unifies `edit-post-visual-editor` and `edit-site-visual-editor` containers. There's already a shared `editor-visual-editor` container. So this PR removes that extra DOM node and moves some styles around. I'm keeping the classnames mostly for backward compatibility though.

## Testing Instructions

1- Ensure that the editor looks properly with and without metaboxes (with and without iframe).
2- Ensure that the site editor looks properly. 
3- Ensure that stylebook looks properly.


jameskoster:
> Ensure that the editor looks properly with and without metaboxes (with and without iframe)

Something isn't right here:


https://github.com/WordPress/gutenberg/assets/846565/ccecc31d-404a-454d-ac41-5b9903f0f423

When editing a long post the metabox appears to be floating in the middle of it, and the post styles (background color at least) disappear beneath it.

youknowriad:
Good catch @jameskoster I think I fixed it in the last commit.

jameskoster (approved):

PR #62017, opened 2024-05-27
Block Library: Remove useless styles

youknowriad:
## What?

While working on some unrelated work, I've found a couple of CSS styles in the block library that looked very suspicious and IMO useless.

These styles target `edit-post-visual-editor` className and specific blocks. `edit-post-visual-editor` is rendered outside the iframe, which means that these CSS rules are only applied to these blocks if they are rendered outside the iframe (for instance when you have some metaboxes enabled in the post editor), so these styles are not used in the site editor and not in the default post editor.

In my quick testing, removing them doesn't really impact the non-iframed mode either but would love help testing.


packages/block-library/src/latest-posts/editor.scss
```diff
@@ -14,10 +14,6 @@
 	display: inline;
 }
 
-.edit-post-visual-editor .wp-block-latest-posts.is-grid li {
-	margin-bottom: 20px;
-}
```
youknowriad, line 19: This style was applied in my testing in non-iframe mode but there's a very similar style in the style.scss which for me means that removing it is fine.

You can test by using the latest posts block in the post editor, switch to "grid" and enable custom fields (to enable non-iframe mode).

packages/block-library/src/block/editor.scss
```diff
@@ -1,35 +0,0 @@
-.edit-post-visual-editor .block-library-block__reusable-block-container {
```
youknowriad, line 1: These styles target reusable blocks in non-iframe mode. In my testing reusable blocks still behave and look properly without these (just like they do in iframe mode)

jasmussen (approved):

PR #62022, opened 2024-05-27
Shadow Panel: Make subtitle translatable

t-hamano:
Apply the translation function to the text below.

shadows-subtitle


madhusudhand (approved):
LGTM!

PR #61253, opened 2024-04-30
Add eslint autofix commit to ignored git commits

sirreal:
## What?

Ignore automated commit 0221522f253e094b277a1485b7a2d186cb172632.

Follow up to https://github.com/WordPress/gutenberg/pull/61204

## Why?

There are not meaningful changes in the commit, it's some tooling-applied changes.

PR #62026, opened 2024-05-27
Add eslint rule for curly brace presence in JSX

DaniGuardiola:
Context: https://github.com/WordPress/gutenberg/pull/61969#discussion_r1615987959

Rule docs: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md

This rule is autofixable and, with the current configuration, has this effect:

```diff
-<div class={ "my-class" } />
+<div class="my-class" />
```

This PR enables the rule (see `.eslintrc` and applies the fix throughout the repo. This is a completely safe fix as the semantics are the same.

**Note: the lint CI errors showing up in this PR are, to the best of my knowledge, pre-existing. They are probably being triggered simply because the PR is touching these files. They should be safe to ignore for the purposes of this PR, though I think it would be beneficial to take a full pass at the whole repo and squash (or mindfully silence) these errors entirely, in a separate PR.**


packages/components/src/animate/stories/index.story.tsx
```diff
@@ -27,7 +27,10 @@ export const Default = Template.bind( {} );
 Default.args = {
 	children: ( { className } ) => (
 		<Notice className={ className } status="success">
-			<p>{ `No default animation. Use one of type = "appear", "slide-in", or "loading".` }</p>
+			<p>
+				No default animation. Use one of type = "appear", "slide-in", or
```
tyxla, line 32: Just using these quotes like this will create some `react/no-unescaped-entities` errors - can we use entities instead?

packages/block-library/src/page-list/convert-to-links-modal.js
```diff
@@ -13,12 +13,12 @@ export function ConvertToLinksModal( { onClick, onClose, disabled } ) {
...
 			onRequestClose={ onClose }
 			title={ __( 'Edit Page List' ) }
-			className={ 'wp-block-page-list-modal' }
+			className="wp-block-page-list-modal"
 			aria={ {
 				describedby: 'wp-block-page-list-modal__description',
 			} }
 		>
-			<p id={ 'wp-block-page-list-modal__description' }>
+			<p id="wp-block-page-list-modal__description">
```
tyxla: Should be quite straightforward to use `useInstanceId()` or something similar here, no? Otherwise, this will create a `no-restricted-syntax` that wasn't reported previously.

tyxla (commented):
Thanks for this 🙌 

Getting the static analysis check to pass should be pretty straightforward, I've left a few suggestions. I'd prefer getting it fixed here and now, since the errors are actually introduced (or start to be reported) since this PR. 

Other than that, good to go, this brings some nice consistency to the codebase 👍

packages/components/src/animate/stories/index.story.tsx
DaniGuardiola: I actually really dislike this rule. I have never found a good argument on why these entities could ever be a problem in a react JSX context, and I've never experienced any problems myself. I'd suggest just disabling them. Entity codes in the source are really annoying and hard to work with.

The official explanation is the following: "This rule prevents characters that you may have meant as JSX escape characters from being accidentally injected as a text node in JSX statements."

While this could theoretically happen and make someone confused for about 3 seconds, with code highlighting and linting this is virtually a non-issue IMHO, and the pain of working with something like `&#34;` in the source -rather than simply `"`- feels like a really bad trade-off to me (talking from experience).

tyxla: I see, and I sympathize with you about this one. However, linting rules often come with some trade-offs, it's tricky for them to serve 100% of the use cases well. I'm pretty open about where this goes, but it should be fixed somehow, even if you decide to do `eslint-disable` right there to avoid potential confusion and keep this more readable.

Mamaduka:
Note: After merging the PR, let's add the merge commit hash to the `.git-blame-ignore-revs` file. Here's the latest example: #61253.

packages/components/src/animate/stories/index.story.tsx
DaniGuardiola: Went with simply disabling the rule globally 👍

tyxla: This seems like a separate, unrelated change though. Mind doing it in another PR where we can discuss it separately?

packages/block-library/src/page-list/convert-to-links-modal.js
```diff
@@ -15,10 +16,18 @@ export function ConvertToLinksModal( { onClick, onClose, disabled } ) {
 			title={ __( 'Edit Page List' ) }
 			className="wp-block-page-list-modal"
 			aria={ {
-				describedby: 'wp-block-page-list-modal__description',
+				describedby: useInstanceId(
+					ConvertToLinksModal,
+					'wp-block-page-list-modal__description'
+				),
```
Mamaduka, line 22: Let's assign the returned value to a variable and reuse it. Otherwise, the new `react-compiler` ESLint rules will complain here.

Update: This would be the warning/error - https://react.dev/reference/rules/react-calls-components-and-hooks#never-pass-around-hooks-as-regular-values

DaniGuardiola: We're not breaking these rules. Using hooks inline is perfectly fine, as long as the traditional rules of hook are not broken. That rule refers to passing the hook itself (as in the hook function value) around.

Mamaduka: IIRC, the current version of the `react-compiler` ESLint rules complains about similar usage. They could fix this in the future or leave it as is.

I don't have a strong opinion, but it's not a significant change, and we can consider it whenever possible.

DaniGuardiola: I personally have strong feelings against unnecessary dis-location (the term I've coined for the opposite of co-location 😝) without good enough reasons. I'm open to being wrong, but it would be surprising to me if the linter complains about it, and if it does it seems like a clear bug or temporary limitation of the React compiler as there's no significant semantic difference between the two styles. But even if it's a bug, as long as the linter complains I'm 100% willing to change it and apply your suggestion.

Can we maybe make sure of this before taking action though? Do you have a reproduction I can look at at hand?

tyxla: @Mamaduka I ran this change through https://github.com/WordPress/gutenberg/pull/61788 and I don't see an ESLint error triggered there. Am I perhaps missing something?

DaniGuardiola: Based on @tyxla's observations I think this is fine to merge as is. Can always go back and change it, if we realize this test was incorrect for some reason. Thanks for chiming in!

packages/element/src/test/serialize.js
```diff
@@ -322,12 +322,11 @@ describe( 'renderElement()', () => {
...
 					<Consumer>{ ( context ) => context.value }</Consumer>
 				</Provider>
-				{ '|' }
+				|
 				<Provider value={ { value: '2nd provided' } }>
 					<Consumer>{ ( context ) => context.value }</Consumer>
 				</Provider>
-				{ '|' }
-				<Consumer>{ ( context ) => context.value }</Consumer>
+				|<Consumer>{ ( context ) => context.value }</Consumer>
```
tyxla, line 329: Should these `|` have been on separate lines?

packages/element/src/test/serialize.js
```diff
@@ -344,8 +343,7 @@ describe( 'renderElement()', () => {
 				<Provider value={ { value: 'inner provided' } }>
 					<Consumer>{ ( context ) => context.value }</Consumer>
 				</Provider>
-				{ '|' }
-				<Consumer>{ ( context ) => context.value }</Consumer>
+				|<Consumer>{ ( context ) => context.value }</Consumer>
```
tyxla, line 346: Similar to above

tyxla (approved):
LGTM 👍 

Let's make sure to follow up with git-blame ignoring this commit as @Mamaduka [suggested](https://github.com/WordPress/gutenberg/pull/62026#issuecomment-2134521109). 

Also, feel free to follow up on polishing how we deal with `react/no-unescaped-entities`.

🚀

packages/element/src/test/serialize.js
DaniGuardiola: I was also a bit weirded out by this, and ran a full pass of formatting just in case. Seems to be just a bit of a weird edge case, but it conforms with auto-formatting so 🤷‍♂️

packages/element/src/test/serialize.js
DaniGuardiola: https://github.com/WordPress/gutenberg/pull/62026#discussion_r1618829728

packages/block-library/src/page-list/convert-to-links-modal.js
Mamaduka: Thanks for double-checking, @tyxla, and sorry for the false alarm, @DaniGuardiola!

packages/element/src/test/serialize.js
tyxla: cc @jsnajdr: Prettier is being unpretty here 😅

jsnajdr: I need some more info about what exactly is the problem. Prettier didn't do the change from `{ '|' }` to `|`, did it? It was done manually to fix (or autofix) the curly brace rule, right?

Is the new syntax incorrect? Does it produce, for example, more or less whitespace that expected in the rendered HTML?

Or is it just unpleasant to look at?

tyxla: Looks like it was auto-fixed but I'll let @DaniGuardiola confirm. 

Looks unpleasant to me, but I'm happy to just not look at it too much. 😅

DaniGuardiola: It's correct, just unpretty :)

The unnecessary curly was eslint autofix, but as I said above I did a full formatting pass afterwards just in case and there were no changes, so this is how prettier likes it.

Mamaduka:
It looks like base branch was little out of the date and we missed a spot, which currently causing CI check to fail - #62109.

tyxla:
Followed up to ignore this commit in git blame in: #62144.

PR #62027, opened 2024-05-27
Make Tabs have a fluid height

### LINKED ISSUE #62018: Tabs: focus style overlaps text when 'Show button text labels' preference is enabled
### Description

The various panels in the editor may use content organized in two or three horizontal tabs.
Typically this is for some blocks inspector panel e.g. the Image and Navigation block but this design pattern is used also in the Glogabl styles panel and elsewhere.

The well known problem with horizontal tabs is that the layout may break when the horizontal space is limited. The editor panels do have limted space, as they are rendered in a narrow column.

As such, when the 'Show button text labels' preference is enabled _and_ the editor is set to languages other than English, some translations provide longer strings that go in two or even three lines. 

The focus style for the Tabs uses a CSS pseudo element that is positioned absolutely and draws the focus blue outline.
Additionally:
- The tabs have a fixed height.
- The vertical padding is insufficient.

The current CSS doesn't take into account longer strings and it appears it hasn't ever been tested with languages other than English.

Having an outline that draws a line on top of text isn't ideal as it partially obscures the text.

Fixed height on the web are often problematic. The desire of making a 'pixel perfect' design often clashes with the reality where content on the web is supposed to re-flow and adapt when strings are longer and / or the viewport changes.

Overall, the styling of the tabs should be adjusted to take into account longer strings by:
- Avoiding to use a fixed height.
- Provide sufficient vertical padding to reserve space for the focus style outline.

This problem can be reproduced when testing with languages that typically provide longer strings or do take more vertical space. Here's a couple screenshots witht the editor language set to Italian and Japanese:

focus style

focus style japanese

It would be worth to start a broader discussion about ways to make sure the Editor layout doesn't break with languages other than English. At the very lesst:
- Testing with other languages should be part of any development that touches the UI and the base components.
- Any design should take inco account languages other than English to start with.

- Switch WordPress language to Italian or Japanese.
- Go to the post editor and add a navigation block.
- Go to Options > Preferences > Accessibility, and enable the 'Show button text labels' preference.
- With the navigation block selected, observe the block inspector panel.
- Use the Tab key to focus the first tab.
- Use the right and left arrow keys to move through the tabs.
- Observe the focus style appears on top of the text.

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/62018

## What?

The styling of the Tabs component doesn't take into account longer strings that wrap to multiple lines. The focus style overlaps the text and partially hides it.

## Why?

Text should always be fully readable.
Fixed heights should be avoided as much as possible, to allow text to re-flow and adapt.

## How?

- Makes the tabs height fluid.
- Simplifies the focus style: in this case there's no need to use both `box-shadow` _and_ `outline`. The latter always works, also for Windows Hight Contrast Mode, and cam be styled exactly like the previous box-shadow.
- The _default_ height is now a `min-height` of 48 pixels.

## Testing Instructions

- Repeat the steps from the issue https://github.com/WordPress/gutenberg/issues/62018
- Observe the tabs height adapts to the content.
- Observe the focus style does not overlap text any longer.
- Thoroughly test all other instances of the Tab component e.g. but not limited to: Main inserter, Styles, Preferences modal dialog, Fonts modal dialog.

## Screenshots or screencast 

Before: 

before

After:

after

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 ### Bug Fixes
 
+-   `Tabs`: Make Tabs have a fluid height ([#62027](https://github.com/WordPress/gutenberg/pull/62027)).
 -   `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
 -   `useUpdateEffect`: Correctly track mounted state in strict mode. ([#62974](https://github.com/WordPress/gutenberg/pull/62974))
 -   `UnitControl`: Fix an issue where keyboard shortcuts unintentionally shift focus on Windows OS. ([#62988](https://github.com/WordPress/gutenberg/pull/62988))
--- a/packages/components/src/tabs/styles.ts
+++ b/packages/components/src/tabs/styles.ts
@@ -58,12 +58,16 @@ export const Tab = styled( Ariakit.Tab )`
 		align-items: center;
 		position: relative;
 		border-radius: 0;
-		height: ${ space( 12 ) };
+		min-height: ${ space(
+			12
+		) }; // Avoid fixed height to allow for long strings that go in multiple lines.
+		height: auto;
 		background: transparent;
 		border: none;
 		box-shadow: none;
 		cursor: pointer;
-		padding: 3px ${ space( 4 ) }; // Use padding to offset the [aria-selected="true"] border, this benefits Windows High Contrast mode
+		line-height: 1.2; // Some languages characters e.g. Japanese may have a native higher line-height.
+		padding: ${ space( 4 ) };
 		margin-left: 0;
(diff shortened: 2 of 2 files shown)


huzaifaalmesbah:
This is looking good to me.

ciampo:
Thank you for working on this, i'll take a look tomorrow

mirka (approved):
I went ahead and merged current `trunk` to test it with the most recent indicator styles for `orientation="vertical"`.

I found two pre-existing problems which I broke out into a separate issue/PR, if anyone else notices them too:

- #63271
- #63272

Otherwise, code changes look good to me 👍

afercia:
Thanks all for your review and feedback.

PR #62028, opened 2024-05-27
DataViews Sidebar: Display item count on DataViews sidebar

### LINKED ISSUE #57072: Data views: Show total count in menu item
In the Patterns page, each view (pattern category) has an associated count:

Screenshot 2023-12-14 at 16 22 10

It would be nice to bring this UI detail to other views, e.g.:

Screenshot 2023-12-14 at 16 21 32

Souptik2001:
## What?

Display item count on DataViews sidebar.

## Why?

Closes: https://github.com/WordPress/gutenberg/issues/57072

As per the issue -

In the Pattern management sidebar each category is represented by a menu item. Inside each menu item is a count indicating how many patterns there are in the category.
Therefore, it would be good to also display that on the "Pages" management sidebar, so that anyone can get the count of each view without visiting it.

## How?

Display the page count beside each view.
Acheived by fetching all the pages and then doing a filter using Javascript.

## Testing Instructions

Check the screencast below or the steps mentioned -

- Go to site-editor.
- Go to "Pages".
- On the sidebar, you will see that you have the count of each item/view.

## Screenshots or screencast 

**There is one problem as you can see in the below video, that the "All Pages" count displays three, that's because it just "All the pages" but inside it, it doesn't display the trash pages, because of which it displays only two items. So, should this be the correct behaviour, or the count should also display 2, or the trash items should also be displayed under all pages.**

https://github.com/WordPress/gutenberg/assets/39427067/11c0302f-372f-417b-a79f-c4c26cf4d0e4


packages/edit-site/src/components/sidebar-dataviews/use-data-views.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Internal dependencies
+ */
+import { OPERATOR_IS_ANY, OPERATOR_IS_NONE } from '../../utils/constants';
+import { DEFAULT_VIEWS } from './default-views';
+
+export default function useDataViews( postType ) {
+	// Fetch all the posts of the post type.
+	const data = useEntityRecords( 'postType', postType, {
+		per_page: -1,
```
Souptik2001, line 16: I am a little skeptical about fetching all the results using -1 like this.
But I am not able to think about any better way. Please suggest if you get any better approach. 🙏

jameskoster:
Thanks for the PR :)

> There is one problem as you can see in the below video, that the "All Pages" count displays three, that's because it just "All the pages" but inside it, it doesn't display the trash pages, because of which it displays only two items. So, should this be the correct behaviour, or the count should also display 2, or the trash items should also be displayed under all pages.

Good question, I think the count should be representative of the visible records, as it is in wp-admin. So in this case the count on 'All pages' should be 2.

Souptik2001:
Thanks @jameskoster !

I have updated the logic to exclude trash pages count from All Pages view.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
```diff
@@ -52,7 +53,16 @@ export const DEFAULT_VIEWS = {
...
-			view: DEFAULT_PAGE_BASE,
+			view: {
+				...DEFAULT_PAGE_BASE,
+				filters: [
+					{
+						field: 'status',
+						operator: OPERATOR_IS_NONE,
+						value: 'trash',
+					},
+				],
```
Souptik2001, line 64: Although this filter is not needed for the query which is run to fetch all the pages, because `trash` is excluded by default, but this needed in the JS filter we are doing here - https://github.com/WordPress/gutenberg/pull/62028/files#diff-67163f722d76f852faaff5a7c0c6046d411846fc118479e62b70273363272d5cR27

packages/edit-site/src/components/sidebar-dataviews/dataview-item.js
```diff
@@ -51,6 +52,7 @@ export default function DataViewItem( {
 			<SidebarNavigationItem
 				icon={ iconToUse }
 				{ ...linkInfo }
+				suffix={ <span>{ count }</span> }
```
jameskoster, line 55: Is the span needed? 🤔

jameskoster:
I left one little comment, but in terms of design this seems good. Just needs a code review :)

ellatrix:
This is an enhancement, so punting to WP 6.7.

jameskoster:
I think this just needs a code review 🙏

packages/edit-site/src/components/sidebar-dataviews/use-data-views.js
ramonjd: If you just want the total count, did you try looking at `totalItems` returned by [useEntityRecords](https://github.com/WordPress/gutenberg/blob/2139bc2ceb6101537be196ef3fddc965b90a6b75/packages/core-data/src/hooks/use-entity-records.ts#L86-L86)?

Example:

```js
const { totalItems } = useEntityRecords( 'postType', 'post', { status: [ 'any', 'trash' ] } ) );
```

`useEntityRecords` makes a call to the `getEntityRecordsTotalItems` selector, which returns the count sent via [response headers in X-WP-Total when calling getEntityRecords](https://github.com/WordPress/gutenberg/blob/2139bc2ceb6101537be196ef3fddc965b90a6b75/packages/core-data/src/resolvers.js#L238-L238).

Or you call `getEntityRecordsTotalItems` directly maybe.

jasmussen:
@WordPress/gutenberg-core Would appreciate a review of this one! It might need a rebase at this point, though.

DaniGuardiola (approved):
Looks good to me and tests well locally. Needs a rebase, and there are some open conversations pending response @Souptik2001.

I actually tried to perform the rebase myself, but then realized the branch is in a fork and not this repo, so I can't push to it.

jasmussen:
@DaniGuardiola Would you be up for performing this rebase, then pushing to a branch directly in based on the core repo? We can credit the author using [this](https://github.com/WordPress/gutenberg/pull/62028#issuecomment-2133482231).

ramonjd:
I'm happy to rebase today and implement the [suggestion to use totalItems count](https://github.com/WordPress/gutenberg/pull/62028#discussion_r1655674533) since I made it 😄

 I'll leave it open to make sure @DaniGuardiola is still happy with it, or wants to make further tweaks.

ramonjd:
I have a rebased PR over at https://github.com/WordPress/gutenberg/pull/65223 (I also couldn't push to this branch), based on the recent changes to trunk. The PR is essentially the same as this one in terms of functionality.

In my testing, it does expose some inconsistencies and bugs with the filtering:

All pages doesn't include trashed items, but the count does. Should we include trashed items in the default view?

Toggling status filters (and sometimes author filters) crashes the browser. This is from trunk.

https://github.com/user-attachments/assets/f22fe1ab-a65b-479d-93f8-73f5ab405119

ramonjd:
> Toggling status filters (and sometimes author filters) crashes the browser. 

PR to fix that I think:

https://github.com/WordPress/gutenberg/pull/65227

jasmussen:
> All pages doesn't include trashed items, but the count does. Should we include trashed items in the default view?

@jameskoster I believed you worked on some mockups here. Any instincts? Mine would be to omit the trashed items from the count.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
youknowriad: Why are we doing JS filtering? IMO, we shouldn't because the REST API can be altered and extended in multiple ways so the only way to get a reliable result is to use the same REST API used in the dataviews.

ramonjd: I need to spend more time with data views, so I can't respond to the above point, but I don't think we need this filter here at all. The resulting filter badge, when clicked, does nothing.

jameskoster:
Agreed. I left a comment in https://github.com/WordPress/gutenberg/pull/65223 to that effect 👍

packages/edit-site/src/components/sidebar-dataviews/dataview-item.js
tyxla: The `span` is not needed, but if `count` is `0` it might mess up with the boolean logic:

https://github.com/WordPress/gutenberg/blob/e1aaee2eb146467c0b20f469c2b04f2adc7295ac/packages/edit-site/src/components/sidebar-navigation-item/index.js#L56

Passing an object (which the React element `<span />` essentially is) will always evaluate to `true`, bypassing that problem.

But again, the `span` is not needed. The same can be achieved with a `React.Fragment`:

```suggestion
				suffix={ <>{ count }</> }
```

Alternatively, we'd alter `suffix` to deal better with `0` values, but this sounds like overkill.

noisysocks:
Closing this in favour of https://github.com/WordPress/gutenberg/pull/65223. Let's remember to props @Souptik2001 for the good work here.

packages/edit-site/src/components/sidebar-dataviews/use-data-views.js
ramonjd: Trying something over here with a new route and [wp_count_posts](https://developer.wordpress.org/reference/functions/wp_count_posts/)

- https://github.com/WordPress/gutenberg/pull/66294

Issue #62023, opened 2024-05-27
Media & text block: The featured image is output in the wrong place if the block has a nested Media & text block.


carolinan:
### Description

If you have a media & text block that:

1. Has the featured image option enabled
2. Has the media on the right
3. Has a media & text block inside the content area

Then the featured image is printed in the wrong position in the markup on the front.
The markup that is supposed to be updated with the featured image  prints and empty `<img>` tag.

### Step-by-step reproduction instructions

1. Activate Gutenberg trunk.
2. Create a new post and add a featured image.
3. Insert a media & text block.
4. Enable the featured image.
5. Enable the option to position the media on the right.
6. Now insert another media & text block in the first media & text block. It doesn't matter if this block has any content or has a media selected.
8. View the front of the website, and compare the result in the editor and front.
9. View the source code to see if the outer media & text block also prints an empty `<img>` tag.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

annezazu:
Adding to 6.6 since this impacts a feature for the release (using featured image in media & text block)

Issue #60497, opened 2024-04-05
Media & Text block: The link option does not work when the featured image is used


carolinan:
### Description

When the Media & Text block uses the featured image, and a link is added using the block toolbar option, the link does not work because it does not wrap around the image.

[It is not added to the block in index.php.](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/media-text/index.php#L32)

### Step-by-step reproduction instructions

Activate Gutenberg
Create a new post and add a featured image.
Add a Media & Text block and select the option "Use featured image".
In the block toolbar, select the link option and add a link.
Save and view the front.
Confirm that the image is not linked.

### Environment info

Gutenberg 18.0 or newer, or Gutenberg trunk.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

fabiankaegy:
This has also been brought up in the Core Editor channel in slack: https://wordpress.slack.com/archives/C02QB2JS7/p1716196185562119.

TLDR: We need help wit engineering

PR #62030, opened 2024-05-27
Media & text: Update the image replacement logic

carolinan:
## What?

Updates the logic that inserts the featured image in the media and text block, taking in to account if the media is set to be on the left or right side. 

For https://github.com/WordPress/gutenberg/issues/62023

## Why?

When the media is on the right, and there is a second media & text block inside the block, the featured image was inserted inside the wrong HTML element.

## How?

First, the updated code checks if the block's `mediaPosition` attribute is set to `right`.
If it is, it locates the **last** figure inside the Media & text block, and inserts an `img` tag inside it.
If not, the `img` tag  is inserted inside the first  figure.

It identifies the correct figure element by the class wp-block-media-text__media to not get mixed up with other figure elements.
It identifies the inserted img tag with the class wp-block-media-text__featured_image.

Then the attributes on the img tag are updated with the data for the featured image, such as the `src`.

## Testing Instructions
Apply the PR.
Create a new post and add a featured image.
Insert a media & text block.
Enable using the featured image option on the block.
Enable the option to position the media on the right.

Now insert another media & text block in the first media & text block. It doesn't matter if this block has any content or has a media selected.
View the front of the website, and compare the size and position of the featured image in the editor and front.

Without this PR; the HTML will look something like this on the front.
Note that there are two img elements inside the first figure.

```
<div class="wp-block-media-text has-media-on-the-right is-stacked-on-mobile has-primary-background-color has-background">
<div class="wp-block-media-text__content">
<div class="wp-block-media-text has-media-on-the-right is-stacked-on-mobile is-vertically-aligned-bottom">
<div class="wp-block-media-text__content"></div>
<figure class="wp-block-media-text__media">
<img decoding="async" width="282" height="159" alt="" class="wp-image-240 size-full" src="...">
<img fetchpriority="high" decoding="async" width="775" height="1024" src="..." alt="" class="wp-image-30 size-full" srcset="..." sizes="(max-width: 775px) 100vw, 775px">
</figure>
</div>
</div>
<figure class="wp-block-media-text__media"><img></figure>
</div>
```

And with the PR:

```
<div class="wp-block-media-text has-media-on-the-right is-stacked-on-mobile has-primary-background-color has-background">
<div class="wp-block-media-text__content">
<div class="wp-block-media-text has-media-on-the-right is-stacked-on-mobile is-vertically-aligned-bottom">
<div class="wp-block-media-text__content"></div>
<figure class="wp-block-media-text__media">
<img fetchpriority="high" decoding="async" width="775" height="1024" src="..." alt="" class="wp-image-30 size-full" srcset="..." sizes="(max-width: 775px) 100vw, 775px"></figure>
</div>
</div>
<figure class="wp-block-media-text__media"><img decoding="async" width="282" height="159" alt="" src="..." class="wp-image-240 size-full"></figure>
</div>

```


carolinan:
This PHP change is hacky and not a complete solution.

If anyone is available, I need help with using the HTML tag processor to locate the **last** `<figure class="wp-block-media-text__media">` in the media and text block when the `mediaPosition` attribute is set to `right`.

Why? Because the content area in the Media & text can contain any block, and any number of blocks, including other figure elements and other media & text blocks. When the media is on the right, then the tag processor will first find the elements that are in the content area.

It needs to work with and without `imageFill` enabled:
Without imagefill, the featured image is an  element inside `<figure class="wp-block-media-text__media">`
With imagefill, the featured image is a background-image in a style attribute on the `<figure class="wp-block-media-text__media">`

@dmsnell @Mamaduka @t-hamano @ramonjd

carolinan:
Perhaps a unique identifier needs to be added to the figure that contains the media, that can then be used in the PHP file, but I'm really not sure.

(This identifier can't just be the image ID since it is possible that the same image is used in the content area)

dmsnell:
> If anyone is available, I need help with using the HTML tag processor to locate the last `<figure class="wp-block-media-text__media">` in the media and text block when the mediaPosition attribute is set to right.

It's easy enough to find the first and last figure with this class name, but the HTML API doesn't yet support inserting a node.

We _can_ get around this by subclassing, but so far we've been reluctant to add workaround code in Core/Gutenberg so as to be careful not to share the wrong example code and accidentally encourage dangerous behavior.

For instance, will this always inject an `IMG` element? What if the `FIGURE` already contains an `IMG`?

The general approach for finding the first is easy, and for finding the last there's a kind of bookmark crawl to take.

```php
$media_tag_processor   = new WP_HTML_Tag_Processor( $content );
$wrapping_figure_query = array( 'tag_name' => 'figure', 'class_name' => 'wp-block-media-text__media' );
if ( $has_media_on_right ) {
	while ( $media_tag_processor->next_tag( $wrapping_figure_query ) ) {
		$media_tag_processor->set_bookmark( 'last_figure' );
	}
	
	if ( $media_tag_processor->has_bookmark( 'last_figure' ) ) {
		// Modify last figure here.
	}
} else {
	if ( $media_tag_processor->next_tag( $wrapping_figure_query ) ) {
		// Modify first figure here.
	}
}

$content = $media_tag_processor->get_updated_html();
```

The HTML API tries to avoid handing out string indices, and while design work is still ongoing for insertion/removal operations, there's no built-in support. For subclasses, there's a way to achieve this through bookmarking.

```php
$media_tag_processor = new class( $content ) extends WP_HTML_Tag_Processor {
	public function insert_after( $dangerous_raw_html_reject_this_patch ) {
		if (
			WP_HTML_Tag_Processor::STATE_READY === $this->parser_state ||
			WP_HTML_Tag_Processor::STATE_INCOMPLETE_INPUT === $this->parser_state ||
			WP_HTML_Tag_Processor::STATE_COMPLETE === $this->parser_state
		) {
			return false;
		}

		$this->set_bookmark( 'here' );
		$here = $this->bookmarks['here'];
		$this->lexical_updates[] = new WP_HTML_Text_Replacement(
			$here->start,
			$here->length,
			$dangerous_raw_html_reject_this_patch
		);
	}
};
```

Inserting an `IMG` tag is largely benign.

carolinan:
@dmsnell  Thank you. @sirreal suggested something similar.

With this update I am not extending the class but still using  `preg_replace` to insert the image tag.
I still need someone to review the logic and the code quality with critical eyes.
But also to consider if there are any other scenarios to test where the feature can break, besides adding a media & text block  or image block inside a media & text block.

ramonjd:
I didn't have a lot of time sorry, but I did a smoke test locally and the frontend appears as I'd expect, even with multiple nested media & text blocks. 🎉 

| Before | After |
|--------|--------|
|Screenshot 2024-05-29 at 10 44 24 am | Screenshot 2024-05-29 at 10 44 36 am | 


There was one bug I found with the link but I see there's an [issue](https://github.com/WordPress/gutenberg/issues/60497) for it already.

Given the constraints I think your approach is fine - the use of the unique id seems like a clever way to target the figure element, and `preg_replace` does the job.

> But also to consider if there are any other scenarios to test where the feature can break, besides adding a media & text block or image block inside a media & text block.

It might help other folks to understand the intention behind the code, and also to have confidence that it doesn't break in various scenarios by getting some test coverage in there, focussing on the expected output. 

See other tests for block render functions:

https://github.com/WordPress/gutenberg/tree/7988696ec25a7373337d91a359929aad204307fd/phpunit/blocks

packages/block-library/src/media-text/index.php
```diff
@@ -29,28 +29,76 @@ function render_block_core_media_text( $attributes, $content ) {
...
+	// and update the image attributes in order to display the featured image.
+	if ( ! $image_fill ) {
+		$content = preg_replace(
+			'/(<figure\s+id="' . preg_quote( $unique_id, '/' ) . '"\s+class="wp-block-media-text__media"\s*>)/',
+			'$1' . $image_tag,
+			$content
+		);
+
+		$image_tag_processor = new WP_HTML_Tag_Processor( $content );
+		$image_tag_processor->next_tag(
```
carolinan: Maybe this should be "_if_ next tag, do this".

packages/block-library/src/media-text/index.php
```diff
@@ -29,28 +29,76 @@ function render_block_core_media_text( $attributes, $content ) {
...
 	}
 
-	$image_tag = '<figure class="wp-block-media-text__media"><img>';
-	$content   = preg_replace( '/<figure\s+class="wp-block-media-text__media">/', $image_tag, $content );
+	$has_media_on_right    = isset( $attributes['mediaPosition'] ) && 'right' === $attributes['mediaPosition'];
+	$image_fill            = isset( $attributes['imageFill'] ) && $attributes['imageFill'];
+	$focal_point           = isset( $attributes['focalPoint'] ) ? round( $attributes['focalPoint']['x'] * 100 ) . '% ' . round( $attributes['focalPoint']['y'] * 100 ) . '%' : '50% 50%';
+	$media_size_slug       = isset( $attributes['mediaSizeSlug'] ) ? $attributes['mediaSizeSlug'] : 'full';
+	$unique_id             = 'wp-block-media-text__media-' . wp_unique_id();
+	$image_tag             = '<img class="wp-block-media-text__featured_image">';
```
carolinan: This class name is probably not needed, and does not exist in the editor, which makes it difficult to use it for styling anyway.

carolinan:
> It might help other folks to understand the intention behind the code, and also to have confidence that it doesn't break in various scenarios by getting some test coverage in there, focussing on the expected output.
> 
> See other tests for block render functions:
> 
> https://github.com/WordPress/gutenberg/tree/7988696ec25a7373337d91a359929aad204307fd/phpunit/blocks

I have tried, but I have never written a test class before. Maybe it should be 3 separate tests, if that is easier to read? One for default settings, one for image fill, and one for when the media is on the right. Now they are all in the same test.

commands:
`npm run test:unit:php  `

Or (after starting wp-env)
`wp-env run --env-cwd='wp-content/plugins/gutenberg' tests-wordpress vendor/bin/phpunit -c phpunit.xml.dist --verbose --filter Render_Block_MediaText_Test`

packages/block-library/src/media-text/index.php
carolinan: I forgot that I already removed it from the resulting markup, it is _only_ used for identifying the image tag.

phpunit/blocks/render-block-media-text-test.php
```diff
@@ -0,0 +1,128 @@
...
+		$this->assertStringContainsString( 'background-image:url(' . wp_get_attachment_image_url( self::$attachment_id, 'full' ) . ')', $rendered );
+		$this->assertStringNotContainsString( '<img', $rendered );
+
+		// Assert that the rendered block contains the featured image when media is on the right.
+		$attributes = array(
+			'useFeaturedImage' => true,
+			'mediaPosition'    => 'right',
+		);
+		$rendered   = gutenberg_render_block_core_media_text( $attributes, $content_media_on_right );
+		$this->assertStringContainsString( wp_get_attachment_image_url( self::$attachment_id, 'full' ), $rendered );
```
ramonjd:
```suggestion
		$this->assertStringContainsString( '<img alt="" src="' . wp_get_attachment_image_url( self::$attachment_id, 'full' ) . '"', $rendered );
```

Thanks a lot for the tests! I know these are tricky to write, especially where unique ids are being generated. 

Here we could check that the image is at least part of an IMG element I think. There might be one or two above I missed.

phpunit/blocks/render-block-media-text-test.php
```diff
@@ -0,0 +1,128 @@
...
+
+		$rendered = gutenberg_render_block_core_media_text( $attributes, $content_nested_media_on_right );
+		$this->assertStringContainsString( wp_get_attachment_image_url( self::$attachment_id, 'full' ), $rendered );
```
ramonjd:
```suggestion
		$this->assertStringContainsString( '<img alt="" src="' . wp_get_attachment_image_url( self::$attachment_id, 'full' ) . '"', $rendered );
```

This one too.

packages/block-library/src/media-text/index.php
ramonjd: Yeah, good point. So something like

```php
		if ( $image_tag_processor->next_tag(
			array(
				'tag_name' => 'figure',
				'id'       => $unique_id,
			)
		) ) {
			// The ID is only used to ensure that the correct figure tag is selected,
			// and can now be removed.
			$image_tag_processor->remove_attribute( 'id' );
			$image_tag_processor->next_tag(
				array(
					'tag_name'   => 'img',
					'class_name' => 'wp-block-media-text__featured_image',
				)
			);
			$image_tag_processor->set_attribute( 'src', esc_url( $current_featured_image ) );
			$image_tag_processor->set_attribute( 'class', 'wp-image-' . get_post_thumbnail_id() . ' size-' . $media_size_slug );
			$image_tag_processor->set_attribute( 'alt', trim( strip_tags( get_post_meta( get_post_thumbnail_id(), '_wp_attachment_image_alt', true ) ) ) );

			$content = $image_tag_processor->get_updated_html();
		}
```

phpunit/blocks/render-block-media-text-test.php
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+/**
+ * Tests for the Media & Text block.
+ *
+ * @group blocks
+ *
+ * @covers ::render_block_core_media_text
+ */
+//class Tests_Blocks_Render_MediaText extends WP_UnitTestCase {
```
ramonjd: Tiny nit - can remove

phpunit/blocks/render-block-media-text-test.php
```diff
@@ -0,0 +1,128 @@
...
+		$content_nested                = '<div class="wp-block-media-text is-stacked-on-mobile"><figure class="wp-block-media-text__media"></figure><div class="wp-block-media-text__content"><div class="wp-block-media-text is-stacked-on-mobile"><figure class="wp-block-media-text__media"></figure><div class="wp-block-media-text__content"><p></p></div></div></div></div>';
+		$content_nested_media_on_right = '<div class="wp-block-media-text has-media-on-the-right is-stacked-on-mobile"><div class="wp-block-media-text__content"><div class="wp-block-media-text is-stacked-on-mobile"><div class="wp-block-media-text__content"><p></p></div><figure class="wp-block-media-text__media"></figure></div></div><figure class="wp-block-media-text__media"></figure></div>';
+
+		// Assert that the rendered block contains the featured image.
+		$attributes = array(
+			'useFeaturedImage' => true,
+		);
+		$rendered   = gutenberg_render_block_core_media_text( $attributes, $content );
+		$this->assertStringContainsString( wp_get_attachment_image_url( self::$attachment_id, 'full' ), $rendered );
```
ramonjd: > I have tried, but I have never written a test class before. Maybe it should be 3 separate tests, if that is easier to read? One for default settings, one for image fill, and one for when the media is on the right. Now they are all in the same test.

It's pretty good! Great job.

I would split them up into 3 tests as they'll be easier to read and maintain.

At the very least I'd add the third message argument to the assertion methods so it's easier to know which test fails, e.g., 

```php
$this->assertStringContainsString( wp_get_attachment_image_url( self::$attachment_id, 'full' ), $rendered, 'The rendered block does not contain a featured image blah blah' );
```

Alternatively you could use a data provider ([Example](https://github.com/WordPress/gutenberg/blob/ee92f208805462a13229a4e4d5baee6898dec734/phpunit/block-supports/layout-test.php#L413-L413)) to loop over a series of test arguments.

But I think just splitting them up into separate functions would be find if you feel comfortable doing that.

packages/block-library/src/media-text/index.php
ramonjd: Do you need it at all?

Wouldn't we expect an image to come after the figure? E.g., 

```php
		$image_tag_processor->next_tag(
			array(
				'tag_name'   => 'img',
		);
```

Otherwise you could delete it as you do with the id. Probably best to remove if it's not required 👍🏻

packages/block-library/src/media-text/index.php
ramonjd: Also maybe for the `tag_name' => 'img',` one as well.

ramonjd (commented):
Thanks for the continued work on this! And also for the tests.

I think it'd be good to break the tests up for readability and also to help isolate fails.

carolinan:
Actually removing the class name from the img did break the nested media & text block so I need to revert it.

phpunit/blocks/render-block-media-text-test.php
carolinan: Oh yeah I was trying to figure out what the best practice was for the naming 😺

phpunit/blocks/render-block-media-text-test.php
```diff
@@ -94,21 +87,66 @@ public function test_render_block_core_media_text_featured_image() {
...
 
-		$rendered = gutenberg_render_block_core_media_text( $attributes, $content_nested );
+		global $wp_query;
+
+		// Fake being in the loop.
+		$wp_query->in_the_loop = true;
+		$wp_query->post        = self::$post;
+
+		$wp_query->posts = array( self::$post );
+		$GLOBALS['post'] = self::$post;
```
carolinan: I want to try to see if I can avoid repeating this code but I might not have time until the end of June.

ramonjd: There are only three tests, so I reckon it's fine to iterate. 

Thanks a lot for splitting them up!

packages/block-library/src/media-text/index.php
```diff
@@ -29,28 +29,78 @@ function render_block_core_media_text( $attributes, $content ) {
 		return $content;
 	}
 
-	$image_tag = '<figure class="wp-block-media-text__media"><img>';
-	$content   = preg_replace( '/<figure\s+class="wp-block-media-text__media">/', $image_tag, $content );
+	$has_media_on_right    = isset( $attributes['mediaPosition'] ) && 'right' === $attributes['mediaPosition'];
+	$image_fill            = isset( $attributes['imageFill'] ) && $attributes['imageFill'];
+	$focal_point           = isset( $attributes['focalPoint'] ) ? round( $attributes['focalPoint']['x'] * 100 ) . '% ' . round( $attributes['focalPoint']['y'] * 100 ) . '%' : '50% 50%';
+	$media_size_slug       = isset( $attributes['mediaSizeSlug'] ) ? $attributes['mediaSizeSlug'] : 'full';
```
ramonjd: Could this be moved down, immediately before it's used?

packages/block-library/src/media-text/index.php
```diff
@@ -29,28 +29,78 @@ function render_block_core_media_text( $attributes, $content ) {
...
+	$unique_id             = 'wp-block-media-text__media-' . wp_unique_id();
+	$image_tag             = '<img class="wp-block-media-text__featured_image">';
```
ramonjd: Same here. Could this be moved inside the `if ( ! $image_fill ) {` block below?

packages/block-library/src/media-text/index.php
```diff
@@ -29,28 +29,78 @@ function render_block_core_media_text( $attributes, $content ) {
...
+			$image_tag_processor->remove_attribute( 'id' );
+			if ( $image_tag_processor->next_tag(
+				array(
+					'tag_name'   => 'img',
+					'class_name' => 'wp-block-media-text__featured_image',
+				)
+			) ) {
+				$image_tag_processor->set_attribute( 'src', esc_url( $current_featured_image ) );
+				$image_tag_processor->set_attribute( 'class', 'wp-image-' . get_post_thumbnail_id() . ' size-' . $media_size_slug );
+				$image_tag_processor->set_attribute( 'alt', trim( strip_tags( get_post_meta( get_post_thumbnail_id(), '_wp_attachment_image_alt', true ) ) ) );
```
ramonjd, line 98: Idea for possible follow up: When the image is a background image add accessible alt text via `aria-label` (or whatever's appropriate) on the figure element

phpunit/blocks/render-block-media-text-test.php
```diff
@@ -0,0 +1,201 @@
...
+ * @group blocks
+ *
+ * @covers ::render_block_core_media_text
+ */
+class Render_Block_MediaText_Test extends WP_UnitTestCase {
+
+	/**
+	 * Post object.
+	 *
+	 * @var object
```
ramonjd:
```suggestion
	 * @var WP_Post
```

ramonjd (approved):
I've tested manually with single media and text and nested. I think this is an improvement 👍🏻 

Thanks for sticking with it.

| Before  | After |
| Screenshot 2024-06-04 at 12 04 13 PM | Screenshot 2024-06-04 at 11 59 37 AM  |


If @dmsnell has time, it'd be good to get his keen eye to run over the HTML Processor changes. 



I noticed a bug - a nested media and text block with an alignment of left does not align left in the frontend or the editor. 

Screenshot 2024-06-04 at 11 52 53 AM

I think this is an existing bug however.

carolinan:
> I noticed a bug - a nested media and text block with an alignment of left does not align left in the frontend or the editor.

This should be fixed: https://github.com/WordPress/gutenberg/pull/62184

dmsnell:
> If @dmsnell has time, it'd be good to get his keen eye to run over the HTML Processor changes.

It'd be nice to get rid of the regex, but that's not convenient or easy with the current limits of the HTML API. I'm happy with how the code is because at least the search for the right place to add the classes is robust. IMO it's easier to clean up existing code patterns (the `regex_replace`) than it is to start hunting down custom subclasses of the HTML API that extend its functionality.

In other words, I still think this is appropriate w.r.t. the HTML processing.

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: ab687f7939

Issue #41303, opened 2022-05-25
Group block variations not recognized as active


Humanify-nl:
### Description

Using a custom block variation and selecting it in the block inspector does not show the 'is-pressed' css class.

The selected varation does in fact get shown in the editor, it is just the button press that does not get registered somehow.

### Step-by-step reproduction instructions

Step 1: add a custom variation to the group block.
``` 
import { __ } from '@wordpress/i18n';
import { group, row, stack } from '@wordpress/icons';

wp.domReady(() => {
  wp.blocks.registerBlockVariation(
    'core/group',
    {
      name: 'group-section',
      title: __( 'Section' ),
      description: __( 'Gather blocks in a section container.' ),
      attributes: {
        align: 'full',
        className: "is-section",
        layout: { type: 'default', inherit: true }
      },
      scope: [ 'inserter', 'transform' ],
      isActive: ( blockAttributes ) =>
        ! blockAttributes.layout ||
        ! blockAttributes.layout?.type ||
        blockAttributes.layout?.type === 'default',
      icon: group,
    }
  );
}

```
Step 2: select the variation in the editor.
Step 3: notice that the first button shows is-pressed, _instead_ of the custom variation.

### Screenshots, screen recording, code snippet

![custom-variation](https://user-images.githubusercontent.com/82579622/170223353-1c3bfaf2-ec0c-40e1-80f0-b7b22e2c5d60.gif)

### Environment info

- WordPress 6.0 with theme.json theme
- Gutenberg plugin not installed.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

Mamaduka:
Hi, @Humanify-nl

Your `isActive` check matches the default group variation check, so the logic has no way to tell the difference.

I would recommend updating `isActive` and making it specific to your variation.

https://github.com/WordPress/gutenberg/blob/90995b6b5e5392ebbc900ec6eebe022daa677ccd/packages/block-library/src/group/variations.js#L14-L17

Humanify-nl:
Thanks @Mamaduka ,

I didnt realize that `isActive` logic is needed for a new variation (docs say optional), I guess this has to do with the new icon-based variation selector that is different from regular variation dropdown on which I couldn't find anything in the docs. Apologies for the incorrect bug report!

Sadly, I can't get the logic to work since the group attributes are mostly the same, and adding below code doesn't seem to work.
```
[...]

PR #62031, opened 2024-05-27
Block Variations: Have `getActiveBlockVariation` return variation with highest specificity

ockham:
## What?
Change `getActiveBlockVariation` such that if multiple variations' `isActive` arrays match the given block's attributes, the variation with the highest specificity (i.e. matching the largest number of attributes) is returned.

## Why?
See the "[Caveats to using isActive](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#caveats-to-using-isactive)" section of the Block Variation docs.

Fixes part of https://github.com/WordPress/gutenberg/issues/41303 (see https://github.com/WordPress/gutenberg/issues/41303#issuecomment-2135217685).

## How?
By assigning a specificity score to each match, corresponding to the number of attributes specified by `isActive`, and returning the match with the highest specificity. Take the [example from the docs](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#caveats-to-using-isactive):

```js
wp.blocks.registerBlockVariation(
	'core/paragraph',
	{
		name: 'paragraph-red',
		title: 'Red Paragraph',
		attributes: {
			textColor: 'vivid-red',
		},
		isActive: [ 'textColor' ],
	}
);

wp.blocks.registerBlockVariation(
	'core/paragraph',
	{
		name: 'paragraph-red-grey',
		title: 'Red/Grey Paragraph',
		attributes: {
			textColor: 'vivid-red',
			backgroundColor: 'cyan-bluish-gray'
		},
		isActive: [ 'textColor', 'backgroundColor' ]
	}
);
```

The specificity (i.e. `isActive.length`) of the `paragraph-red` variation is `1`, whereas for the `paragraph-red-grey` variation, it is 2. If the current block matches both variations, the one with the higher specificity (here: `paragraph-red-grey`) is returned.

Note that if any matching variation uses a _function_ rather than an array for its `isActive` field, a specificity is not assigned; as a result, the match cannot be compared to the other matches. In this case, the first matching variation will be returned instead (much like `getActiveBlockVariation` did anyway, prior to this PR). (See https://github.com/WordPress/gutenberg/issues/62068 for a related issue; fixing it should make it possible to use arrays rather than functions for in most cases.)

## Testing Instructions
Do some some testing with block variations in the editor.
Other than that, refer to automated tests.

### DIFF
--- a/docs/reference-guides/block-api/block-variations.md
+++ b/docs/reference-guides/block-api/block-variations.md
@@ -129,9 +129,9 @@ While the `isActive` property is optional, it's recommended. This API is used by
 
 If `isActive` is not set, the Editor cannot distinguish between an instance of the original block and your variation, so the original block information will be displayed.
 
-The property can be set to either a function or an array of strings (`string[]`).
+The property can be set to either an array of strings (`string[]`), or a function. It is recommended to use the string array version whenever possible.
 
-The function version of this property accepts a block instance's `blockAttributes` as the first argument, and the `variationAttributes` declared for a variation as the second argument. These arguments can be used to determine if a variation is active by comparing them and returning a `true` or `false` (indicating whether this variation is inactive for this block instance).
+The `string[]` version is used to declare which of the block instance's attributes should be compared to the given variation's. Each attribute will be checked and the variation will be active if all of them match.
 
 As an example, in the core Embed block, the `providerNameSlug` attribute is used to determine the embed provider (e.g. 'youtube' or 'twitter'). The variations may be declared like this:
 
@@ -162,18 +162,13 @@ const variations = [
 ]
 ```
 
- The `isActive` function can compare the block instance value for `providerNameSlug` to the value declared in the variation's declaration (the values in the code snippet above) to determine which embed variation is active:
-
-```js
-isActive: ( blockAttributes, variationAttributes ) =>
-	blockAttributes.providerNameSlug === variationAttributes.providerNameSlug,
-```
-
-The `string[]` version is used to declare which attributes should be compared as a shorthand. Each attribute will be checked and the variation will be active if all of them match. Using the same example for the embed block, the string version would look like this:
+The `isActive` property would then look like this:
 
 ```js
 isActive: [ 'providerNameSlug' ]
 ```
+
+This will cause the block instance value for `providerNameSlug` to be compared to the value declared in the variation's declaration (the values in the code snippet above) to determine which embed variation is active.
 
 Nested object paths are also supported. For example, consider a block variation that has a `query` object as an attribute. It is possible to determine if the variation is active solely based on that object's `postType` property (while ignoring all its other properties):
 
@@ -181,9 +176,18 @@ Nested object paths are also supported. For example, consider a block variation 
 isActive: [ 'query.postType' ]
 ```
 
-### Caveats to using `isActive`
+The function version of this property accepts a block instance's `blockAttributes` as the first argument, and the `variationAttributes` declared for a variation as the second argument. These arguments can be used to determine if a variation is active by comparing them and returning a `true` or `false` (indicating whether this variation is inactive for this block instance).
 
-The `isActive` property can return false positives if multiple variations exist for a specific block and the `isActive` checks are not specific enough. To demonstrate this, consider the following example:
+Using the same example for the embed block, the function version would look like this:
+
+```js
+isActive: ( blockAttributes, variationAttributes ) =>
+	blockAttributes.providerNameSlug === variationAttributes.providerNameSlug,
+```
+
+### Specificity of `isActive` matches
+
+If there are multiple variations whose `isActive` check matches a given block instance, and all of them are string arrays, then the variation with the highest _specificity_ will be chosen. Consider the following example:
 
 ```js
 wp.blocks.registerBlockVariation(
@@ -212,6 +216,6 @@ wp.blocks.registerBlockVariation(
 );
 ```
 
-The `isActive` check on both variations tests the `textColor`, but each variations uses `vivid-red`. Since the `paragraph-red` variation is registered first, once the `paragraph-red-grey` variation is inserted into the Editor, it will have the title `Red Paragraph` instead of `Red/Grey Paragraph`. As soon as the Editor finds a match, it stops checking.
+If a block instance has attributes `textColor: vivid-red` and `backgroundColor: cyan-bluish-gray`, both variations' `isActive` criterion will match that block instance. In this case, the more _specific_ match will be determined to be the active variation, where specificity is calculated as the length of each `isActive` array. This means that the `Red/Grey Paragraph` will be shown as the active variation.
 
-There have been [discussions](https://github.com/WordPress/gutenberg/issues/41303#issuecomment-1526193087) around how the API can be improved, but as of WordPress 6.3, this remains an issue to watch out for.
+Note that specificity cannot be determined for a matching variation if its `isActive` property is a function rather than a `string[]`. In this case, the first matching variation will be determined to be the active variation. For this reason, it is generally recommended to use a `string[]` rather than a `function` for the `isActive` property.
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -237,10 +237,17 @@ export const getBlockVariations = createSelector(
 export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 	const variations = getBlockVariations( state, blockName, scope );
 
-	const match = variations?.find( ( variation ) => {
+	if ( ! variations ) {
+		return variations;
+	}
+
+	const blockType = getBlockType( state, blockName );
+	const attributeKeys = Object.keys( blockType?.attributes || {} );
+
+	const matches = [];
+
+	for ( const variation of variations ) {
 		if ( Array.isArray( variation.isActive ) ) {
-			const blockType = getBlockType( state, blockName );
-			const attributeKeys = Object.keys( blockType?.attributes || {} );
 			const definedAttributes = variation.isActive.filter(
 				( attribute ) => {
 					// We support nested attribute paths, e.g. `layout.type`.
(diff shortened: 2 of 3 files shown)


ntsekouras:
Thanks for the PR!

I [commented](https://github.com/WordPress/gutenberg/issues/41303#issuecomment-2135511809) on the Group block variations issue and it seems really related:

>That's a hard problem indeed. We had some explorations in the past for this to see how it would be best to extend a core block and its variations. We ended up adding a 'dumb' [attribute(namespace) in Query Loop ](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/extending-the-query-loop-block/)block that's been taken into account in isActive function and can be used in filters, hooks, php, etc.. While it might not be ideal, this approach seems reliable and is similar to the workaround suggested above(className, etc..).

> Maybe something like that could be a reliable solution, by injecting in all blocks an attribute(like the namespace) with a supports flag? 🤔


This approach still has some shortcomings with the `isActive` being a function for one. The most important thing probably though is that sometimes with variations we want to more(like adding extra controls with filters etc..), without actually having many more attributes to define that the variation is active. An example would be the `Posts List` variation of Query Loop and a possible new variation that just wants to add more controls, but still cares only if the `postType` is `post`. Makes sense?

Sorry if I'm focusing more on examples of Query Loop and object attributes, but that's the use case I've discussed/explored most..

ockham:
> Sorry if I'm focusing more on examples of Query Loop and object attributes, but that's the use case I've discussed/explored most..

No, that's fine! It's actually helpful to see other use cases that I hadn't taken into account 🙂 

>> Maybe something like that could be a reliable solution, by injecting in all blocks an attribute(like the namespace) with a supports flag? 🤔

So basically a "silent" attribute that's used to differentiate variations, but doesn't have any other effect on the block? During our explorations related to https://github.com/WordPress/gutenberg/issues/43743, @tjcafferkey and I considered adding a `metadata.variation` attribute to a block to store its currently active variations. Ultimately, we decided against this (at least for now), as we were hoping to infer the variation from attributes instead (to avoid having to sync the attribute, among other things). As a consequence, we're trying to improve the logic that we're using to infer the active variation, with the goal of getting incremental improvements (like this PR, or https://github.com/WordPress/wordpress-develop/pull/6602) into WP 6.6 🙂 

Now while this PR cannot detect the active variation in the situation you're describing, I believe it would cover the specific use case described in #41303 (if we also implement the idea in https://github.com/WordPress/gutenberg/issues/62068) and AFAICS, there wouldn't be any adverse side effects. Do you think it's viable to land those two things before we consider adding a "silent" attribute (like the `namespace` in [this example](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/extending-the-query-loop-block/)) to all variations?

docs/reference-guides/block-api/block-variations.md
```diff
@@ -162,28 +162,32 @@ const variations = [
...
 ```
 
-The `string[]` version is used to declare which attributes should be compared as a shorthand. Each attribute will be checked and the variation will be active if all of them match. Using the same example for the embed block, the string version would look like this:
+This will cause the block instance value for `providerNameSlug` to be compared to the value declared in the variation's declaration (the values in the code snippet above) to determine which embed variation is active.
+
+Nested object paths are also supported. For example, consider a block variation that has a `query` object as an attribute. It is possible to determine if the variation is active solely based on that object's `postType` property (while ignoring all its other properties):
 
 ```js
-isActive: [ 'providerNameSlug' ]
+isActive: [ 'query.postType' ]
```
gziolo, line 176: @justintadlock and @bacoords, I think you will like these change in relation to Block Bindings 😄

ndiego: This is a fantastic PR overall. 🙌

packages/blocks/src/store/selectors.js
```diff
@@ -251,27 +258,49 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
...
+			}
+			return variation;
 		}
+	}
 
-		return variation.isActive?.( attributes, variation.attributes );
-	} );
-
-	return match;
+	let candidate = [ undefined, 0 ];
```
gziolo: This part of the code is hard to follow because of the tuple-like array. However, it works as expected.

packages/blocks/src/store/selectors.js
```diff
@@ -237,10 +237,17 @@ export const getBlockVariations = createSelector(
 export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 	const variations = getBlockVariations( state, blockName, scope );
 
-	const match = variations?.find( ( variation ) => {
+	if ( ! variations ) {
```
gziolo, line 240: I wasn't sure about this like but `getBlockVariations` returns a non-empty array or undefined after converting the array into the return value.

gziolo (commented):
Code-wise, this looks good to me. I will give it some testing as I wanted to confirm it plays nicely with block bindings :)

docs/reference-guides/block-api/block-variations.md
```diff
@@ -129,9 +129,9 @@ While the `isActive` property is optional, it's recommended. This API is used by
 
 If `isActive` is not set, the Editor cannot distinguish between an instance of the original block and your variation, so the original block information will be displayed.
 
-The property can be set to either a function or an array of strings (`string[]`).
+The property can be set to either an array of strings (`string[]`), or a function. It is recommended to use the string array version whenever possible.
 
-The function version of this property accepts a block instance's `blockAttributes` as the first argument, and the `variationAttributes` declared for a variation as the second argument. These arguments can be used to determine if a variation is active by comparing them and returning a `true` or `false` (indicating whether this variation is inactive for this block instance).
+The `string[]` version is used to declare which of the block instance's attributes should be compared to the given variation's. Each attribute will be checked and the variation will be active if all of them match.
```
ntsekouras, line 134:
```suggestion
The `string[]` version is used to declare which of the block instance's attributes should be compared to each variation's attributes. Each attribute will be checked and the variation will be active if all of them match.
```

docs/reference-guides/block-api/block-variations.md
```diff
@@ -162,28 +162,32 @@ const variations = [
...
+
+Nested object paths are also supported. For example, consider a block variation that has a `query` object as an attribute. It is possible to determine if the variation is active solely based on that object's `postType` property (while ignoring all its other properties):
 
 ```js
-isActive: [ 'providerNameSlug' ]
+isActive: [ 'query.postType' ]
 ```
 
-Nested object paths are also supported. For example, consider a block variation that has a `query` object as an attribute. It is possible to determine if the variation is active solely based on that object's `postType` property (while ignoring all its other properties):
+The function version of this property accepts a block instance's `blockAttributes` as the first argument, and the `variationAttributes` declared for a variation as the second argument. These arguments can be used to determine if a variation is active by comparing them and returning a `true` or `false` (indicating whether this variation is inactive for this block instance).
```
ntsekouras, line 179:
```suggestion
The function version of this property accepts a block instance's `blockAttributes` as the first argument, and the `variationAttributes` declared for a variation as the second argument. These arguments can be used to determine if a variation is active by comparing them and returning `true` or `false` (indicating whether this variation is inactive for this block instance).
```

docs/reference-guides/block-api/block-variations.md
```diff
@@ -162,28 +162,32 @@ const variations = [
...
-isActive: [ 'query.postType' ]
+isActive: ( blockAttributes, variationAttributes ) =>
+	blockAttributes.providerNameSlug === variationAttributes.providerNameSlug,
 ```
 
-### Caveats to using `isActive`
+### Specificity of `isActive` matches
 
-The `isActive` property can return false positives if multiple variations exist for a specific block and the `isActive` checks are not specific enough. To demonstrate this, consider the following example:
+If there are multiple variations whose `isActive` check matches a given block instance, and all of them are string arrays, then the variation with the highest _specificity_ will be chosen. Consider the following example:
```
ntsekouras, line 190: Nit: Maybe make bold the `all of them`?

gziolo:
I'm seeing some small inconsistencies after inserting a variation. It gets recognized correctly upon insertion, but after reloading the page, it no longer can be recognized as a variation. Maybe the comparison is based on the object reference which is the same upon insertion but will differ after the page reloads?


https://github.com/WordPress/gutenberg/assets/699132/87658869-455b-4b06-9875-3ecc4637220b

This is the modification that I applied to the Paragraph block:

```diff
diff --git a/packages/block-library/src/paragraph/block.json b/packages/block-library/src/paragraph/block.json
index 6f11baf838..8bd350fb1e 100644
--- a/packages/block-library/src/paragraph/block.json
+++ b/packages/block-library/src/paragraph/block.json
@@ -70,5 +70,25 @@
 		}
 	},
 	"editorStyle": "wp-block-paragraph-editor",
-	"style": "wp-block-paragraph"
+	"style": "wp-block-paragraph",
+	"variations": [
+		{
+			"name": "warning",
+			"title": "Warning",
+			"icon": "warning",
+			"attributes": {
+				"metadata": {
+					"bindings": {
+						"content": {
+							"source": "core/post-meta",
+							"args": {
+								"key": "text_field"
+							}
+						}
+					}
+				}
+			},
+			"isActive": [ "metadata.bindings.content" ]
+		}
+	]
 }
```

ntsekouras (commented):
Thank you @ockham! Code wise looks good! I'm trying something small if we can simplify a bit and @gziolo comment is probably something to look into. I haven't work on bindings, so it would be better for folks who have, to test better.

gziolo:
I'm trying to test this variation added to the Paragraph block:
```json
"variations": [
	{
		"name": "warning-with-color",
		"title": "Warning with color",
		"icon": "warning",
		"attributes": {
			"content": "This is a warning message.",
			"backgroundColor": "luminous-vivid-amber"
		},
		"isActive": [ "content", "backgroundColor" ]
	}
]
```

It isn't recognized as active for some reason I don't fully understand for both `trunk` and this branch. I was wondering if it ever worked this way or regressed at some point. In my testing, I figured out that it works correctly for:

```json
"isActive": [ "backgroundColor" ]
```

but not

```json
"isActive": [ "content" ]
```

In effect, I think we need to fix it separately for attributes represented as RichText objects and cast them to string values when comparing.

gziolo (approved):
Very excited about this functionality, we will have to follow-up with some small fixes to ensure that objects are compared by value, and RichText value properly casted to string value. The same issues existed before so it isn't directly related to this improved functionality but rather related to the way I'm testing it.

ntsekouras:
Hope you don't mind, but pushed a commit that I think it simplifies things a bit and also keeps the best match if a variation has defined an `isActive` function (see the test change). Feel free to revert of course 😄

gziolo:
@ntsekouras, feel free to land it when CI passes so we have it ready on time for WP 6.6 Beta 1 😄

gziolo:
> I'm seeing some small inconsistencies after inserting a variation. It gets recognized correctly upon insertion, but after reloading the page, it no longer can be recognized as a variation. Maybe the comparison is based on the object reference which is the same upon insertion but will differ after the page reloads?

I wanted to report back on my findings. I did some more testing, and everything works as in `trunk` when defining the `isActive` array correctly with the new notation. We can compare at the moment only using the primitive values. I solved the issues reported previously by using more specific selectors for comparison:

```json
"isActive": [
	"backgroundColor",
	"metadata.bindings.content.source",
	"metadata.bindings.content.args.key"
]
```

This is how it looks in action for a few variations correctly defined:



https://github.com/WordPress/gutenberg/assets/699132/374f7f1a-ba9d-4ee9-8c61-453aab12d7ae

ockham:
Thank you both for approving and landing this! 🎉 

> Hope you don't mind, but pushed a commit that I think it simplifies things a bit and also keeps the best match if a variation has defined an `isActive` function (see the test change). Feel free to revert of course 😄

@ntsekouras I was initially considering a similar approach but ultimately decided against. IIUC, your change returns the best `string[]` `isActive` based match if an `isActive` function is found. This silently assumes that the specificity of the string array `isActive` is better than the function one.

It seemed to me that we couldn't make that assumption, as an `isActive` function is kind of opaque. Internally, it could compare one or ten attributes, in which case its de-facto specificity would be very low, or very high; I didn't want to assume either way. Instead, I concluded that the only consistent behavior would be to discard the specificity sorting if there were any `isActive` properties that matched the criterion.

Granted, it is probably an edge case to have a mix of `string[]` and function `isActive` properties, and defaulting to the first match (as I did) is also somewhat arbitrary. I can't seem to come up with a case where the current logic yields "worse" result than the previous method, so it might be just fine 🙂

ockham:
@gziolo I'll file issues and/or PRs to address object and RichText comparison in the next couple of days!

Issue #51472, opened 2023-06-13
Rename the site editor list view to document overview


alexstine:
Currently, the site editor contains a list view of blocks but it is slightly confusing because the post editor has the same sidebar but it is called document overview. The main difference between the two is the post editor also includes characters, words, etc. statistics about the post or page while the site editor only has the list of blocks.

Does it make sense to rename for parity sake the list view to document overview in the site editor to match the post editor?

CC: @annezazu @andrewserong

andrewserong:
Oh, good question! I imagine it was left as List View in the site editor since Document Overview was likely created as a parent name to describe an area that encompasses both the List View and the Outline. It would make sense to me to name the area (and button that opens it) Document Overview, and retain the List View heading where it currently is, as it somewhat matches the List View tab in the post editor.

I'll just add the Needs Design Feedback label as there are likely other folks who might have a bit more insight into some of the strategy surrounding naming with the site editor, but overall I like the idea, personally!

hanneslsm:
I added the idea to the ongoing naming discussion: https://github.com/WordPress/gutenberg/discussions/49837

afercia:
With the ongoing effor to unify the editor, this panel now contains the same features in both the post editor and the site editor:
- The List View.
- The Document Outline.

However, the toggle button that opens it and the navigable region that contains the panel are still labeled differently.

As such, I'm changing this issue 'type' label from Question to Bug.

annezazu:
cc @ntsekouras and @youknowriad as part of the work you all are doing around unification here.

ntsekouras:
@afercia  this has been resolved [here](https://github.com/WordPress/gutenberg/pull/57467). I think we can close this issue, no?

afercia:
Sorry for late reply, yes I think this can be closed ❤️

PR #62032, opened 2024-05-27
Editor: Remove obsolete `listViewLabel` prop from DocumentTools

ntsekouras:
## What?
I was looking into https://github.com/WordPress/gutenberg/issues/51472 and observed that we have kept an obsolete prop, since the unification of Document tools has been made.

## Testing instructions
The document tools should work as before with the same `Document Overview` label in both editors, as before


Mamaduka (approved):
Good catch, @ntsekouras!

PR #62033, opened 2024-05-27
Update: Move template areas into a panel.

jorgefilipecosta:
This PR removes the "Areas" section of templates and makes the content panel available for templates.
It also simplifies a lot the code as it uses BlockQuickNavigation which is generic for any block instead of something specific for template parts.

Part of the remaining items in https://github.com/WordPress/gutenberg/issues/59689#issuecomment-2110493251.

## Screenshots or screencast 
Before:

Screenshot 2024-05-27 at 16 36 33

After:
Screenshot 2024-05-27 at 16 35 28


jasmussen:
This makes sense to me, and appears to work as intended:

2024-06-18 12 43 17 local-wordpress local 71e3efc37eca

"Areas" was always a bit obtuse for me, so I'm here for it, though would love a gut-check by @WordPress/gutenberg-design.

jameskoster:
On this branch the Content panel seems to have disappeared when editing a page.

There's also a small UX quirk; upon loading the Editor, if you click an area the tab switches from Page to Block. However on subsequent clicks this doesn't occur. This also seems to be a problem on trunk when editing a page, so perhaps one to fix separately.


https://github.com/WordPress/gutenberg/assets/846565/c70b5753-bd1b-45d4-b13c-3f0f6fb2538b

Finally I think the icon should match the associated area. IE header template parts should have the `header` icon.

jasmussen:
> There's also a small UX quirk; upon loading the Editor, if you click an area the tab switches from Page to Block. However on subsequent clicks this doesn't occur. This also seems to be a problem on trunk when editing a page, so perhaps one to fix separately.

Speaking of, there's been some changing behavior here. In the past, clicking a block would _always_ switch to the block tab. In the site editor, the Page tab started to become a bit stickier, and now with the unification it feels like we have something in-between, where the page tab will be sticky if you click it, until you deselect all blocks (which is currently hard in the site editor) and then selecting a block again. What do you think should be the ideal behavior here?

jameskoster:
I don't know that there's a hard rule we can apply across, but in this particular case I find it slightly unexpected that clicking an item in the Content panel also switches tabs.

jasmussen:
> in this particular case I find it slightly unexpected that clicking an item in the Content panel also switches tabs.

That's a good point. But could that be the exception? That clicking any block in the _canvas_ still selects the block tab (so the inspector can remain contextual by default), but clicking a content item doesn't?

jameskoster:
Yes, that might work.

How about List View? I noticed the issue around inconsistent behavior exists there too.

jasmussen:
List view is a good question too, because an instance of it lives inside the navigation block for managing items there. I _suspect_ that instance already has local code to not swap tabs when you're just selecting an item there. If that's the case, then I feel like selecting a block in the canvas should be the same as selecting a block in the list view.

It's not because trunk works very poorly at the moment: it's mostly that the behavior is unpredictable. I don't know why sometimes the block tab is invoked, but not others. I think if it's mainly for the content panel, it would feel implied to not switch tabs when using those controls.

jameskoster:
> I think if it's mainly for the content panel, it would feel implied to not switch tabs when using those controls.

Agreed. This feels like a good change for this PR, assuming it isn't a big can of worms 🤞

jorgefilipecosta:
Hi @jasmussen, and @jameskoster, this PR was updated we now keep the template tab when an item is clicked.

jorgefilipecosta:
Regarding the other issues:

> On this branch the Content panel seems to have disappeared when editing a page.

I'm not being able to reproduce this issue. @jameskoster would you be able to provide the steps you are following or a recording? Thank you in advance for the help.

> Finally I think the icon should match the associated area. IE header template parts should have the header icon.

Yes, this one seems like a bug on the list view things. I'm fixing this issue.

jorgefilipecosta:
> > Finally I think the icon should match the associated area. IE header template parts should have the header icon.

This is now fixed.

I think I addressed the pending items. It should be ready for a new look.

jameskoster:
>  @jameskoster would you be able to provide the steps you are following or a recording? Thank you in advance for the help.

No specific steps, just editing a page in the site editor.

| Trunk | This PR |
| --- | --- |
| Screenshot 2024-07-03 at 10 17 23 | Screenshot 2024-07-03 at 10 17 32 |

The other issues appear to be fixed 👍

jorgefilipecosta:
Hi @jameskoster, the issue of the page's content disappearing is fixing not sure why it did not happen probably tested on the wrong branch, but things should work now.

Let me know if there is any other issue or if we can merge. Thank you for there reviews and testing 😊

jameskoster (approved):
Seems to be working now 🚀

packages/editor/src/components/sidebar/index.js
```diff
@@ -112,9 +112,7 @@ const SidebarContent = ( {
 				<Tabs.TabPanel tabId={ sidebars.document } focusable={ false }>
 					<PostSummary onActionPerformed={ onActionPerformed } />
 					<PluginDocumentSettingPanel.Slot />
-					{ renderingMode !== 'post-only' && (
-						<TemplateContentPanel />
-					) }
+					<TemplateContentPanel renderingMode={ renderingMode } />
```
youknowriad, line 115: Why are we passing renderingMode here, can't we just select it within the component?

youknowriad (commented):
I like the unification of the content components.

packages/editor/src/store/private-selectors.js
```diff
@@ -120,29 +118,6 @@ export const getPostIcon = createRegistrySelector(
 	}
 );
 
-/**
- * Returns the template parts and their blocks for the current edited template.
- *
- * @param {Object} state Global application state.
- * @return {Array} Template parts and their blocks in an array.
- */
-export const getCurrentTemplateTemplateParts = createRegistrySelector(
```
Mamaduka, line 129: This selector is still "aliased" in the site editor store and triggers an error now.

Should we also remove it?

https://github.com/WordPress/gutenberg/blob/387acefdeb14d4d8aeb2885b840c2ffa50562b64/packages/edit-site/src/store/selectors.js#L241-L247

PR #62034, opened 2024-05-27
Add: Content panel for template parts.

jorgefilipecosta:
This PR implements a content panel for template parts whose mockup is available here https://github.com/WordPress/gutenberg/issues/59689#issuecomment-2092659485.

Blocks that appear on the content area of template parts are the blocks whose category is 'theme'.

## Testing Instructions
I opened some template parts and verified they included a content panel.

## Screenshots or screencast 
Screenshot 2024-05-27 at 17 11 03

### DIFF
--- a/packages/editor/src/components/sidebar/index.js
+++ b/packages/editor/src/components/sidebar/index.js
@@ -30,6 +30,7 @@ import PostTaxonomiesPanel from '../post-taxonomies/panel';
 import PostTransformPanel from '../post-transform-panel';
 import SidebarHeader from './header';
 import TemplateContentPanel from '../template-content-panel';
+import TemplatePartContentPanel from '../template-part-content-panel';
 import useAutoSwitchEditorSidebars from '../provider/use-auto-switch-editor-sidebars';
 import { sidebars } from './constants';
 import { unlock } from '../../lock-unlock';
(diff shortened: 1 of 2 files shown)


jasmussen:
Nice one. I'm not seeing the panel show up for me, though, am I doing something wrong?
2024-06-18 12 41 22 local-wordpress local 5a4c97038d1b

jorgefilipecosta:
Hi @jasmussen, it seems you opened a template and not a template part, for this to appear you need to open a template part directly in the editor (e.g: go to patterns, header, and open a header template part), it appears under the "Template part" tab, not the block one.
Right now we don't have content panels in the block tab but it may be something to consider 😊

jasmussen (approved):
Works as advertised, good improvement. Could use a code review. Before:

before

After:

after

Your explanation of what I did wrong makes sense, however it does suggest that what exists on the editor for the individual template part could exist on the block tab for a template part as well. Both because that block tab is currently empty:

per block

But also because there's precedence for this content panel in the block tab, with contentOnly patterns, and synced pattern overrides as well.

jorgefilipecosta:
> Your explanation of what I did wrong makes sense, however it does suggest that what exists on the editor for the individual template part could exist on the block tab for a template part as well. Both because that block tab is currently empty

Sure, let me try in a follow-up to add a content panel there 😊

PR #62036, opened 2024-05-27
Documentation: PostPreviewButton editor component

colorful-tones:
## What? & Why?
Addresses one item in #60358

- mitigates risk

## How?
Add a JSDoc comment to the `PostPreviewButton` component and run `npm run docs:build` to populate the `README` with the newly added documents.


packages/editor/src/components/post-preview-button/index.js
```diff
@@ -99,6 +99,18 @@ function writeInterstitialMessage( targetDocument ) {
 	targetDocument.close();
 }
 
+/**
+ * A button for post preview. It uses WordPress data selectors to get the necessary data for the preview, such as the post ID, the current post link, the preview link, and whether the post is saveable and viewable. When the button is clicked, it opens a new window or tab for the preview, writes the interstitial message to this window, and then navigates to the actual preview link. The button is disabled if the post is not saveable.
```
ntsekouras: I think we should have a bit higher level description..

```suggestion
 * Renders a button that opens a new window or tab for the preview, writes the interstitial message to this window, and then navigates to the actual preview link. The button is not rendered if the post is not viewable and disabled if the post is not saveable.
```

ntsekouras: Also just noticed this.. We should probably wrap this comment in more lines to be consistent with our code base. I thought we had a linting rule for the max length of comments line, but it seems we don't 😄

ntsekouras (approved):
Thank you! It seems this PR and some others from you will need an extra rebase, because a linting error slipped in trunk, but already fixed.

PR #60727, opened 2024-04-14
Components: Remove CSS hack for Internet Explorer 11

t-hamano:
Part of #58034

## What?

This PR removes the fallback CSS for IE11 present in the components package.

## Why?

Support for IE11 has been [officially removed in WordPress 5.8](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/). And since the Gutenberg plugin no longer supports WordPress 5.8, we can remove the fallback code for IE11.

## Testing Instructions

- Run Syorybook and access the URL below.
  - `Placeholder`: http://localhost:50240/?path=/story/components-placeholder--default
  - `Toolbar`: http://localhost:50240/?path=/story/components-toolbar--default&args=label: (From my understanding, one scenario for having a legacy toolbar, i.e. `components-toolbar` class, is to have an empty label.)
- The exact same CSS as before should be applied.

## Screenshots or screencast 

Below are the CSS changes applied to the two components affected by this PR.

### Placeholder

| Before | After |
|--------|--------|
| image | image | 

### Toolbar

| Before | After |
|--------|--------|
|  image|  image|

Issue #58034, opened 2024-01-20
Remove fallback code for IE11


t-hamano:
## What problem does this address?

Support for IE11 has been [officially removed in WordPress 5.8](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/). And since the Gutenberg plugin no longer supports WordPress 5.8, we can remove the fallback code for IE11.

## What is your proposed solution?

- Search the codebase for `ie11`, `ie 11`, `internet explorer`, etc.
- Analyze and modify or remove any fallback code found.

t-hamano:
Almost all of the styles and JS code for IE11 have been removed. As far as I can tell, the only ones that remain are the following two places:

- [TypewriterOrIEBypass](https://github.com/WordPress/gutenberg/blob/1f4891ec7c3f2862272cc6571f208d0633c4cc66/packages/block-editor/src/components/typewriter/index.js#L285): This is a public variable, so we may not be able to delete it.
- [Conditional statement in `input-and-selection.js` file](https://github.com/WordPress/gutenberg/blob/1f4891ec7c3f2862272cc6571f208d0633c4cc66/packages/rich-text/src/component/event-listeners/input-and-selection.js#L137-L142): I tried to remove it once, but the e2e test failed, so it may be necessary for browsers other than IE11.

Based on the above, I don't think there is any other code that can be removed, so I will consider this issue closed.

PR #44707, opened 2022-10-05
Cover: Fix erroneous focus style in editor.

jasmussen:
## What?

In the editor, the focus style for the cover block is wrong, showing a gap at the bottom:

before

This PR fixes that gap, by unsetting a min-height applied to the focus style in the editor:

after

As best I can tell from context in style.scss, that min-height exists solely for the frontend, and to fix an older IE issue at that. This PR keeps that hack intact, but adds a comment explaining the context.

## Testing Instructions

Test the cover block in the editor, and observe no gap between image and focus style bottom border.

PR #62043, opened 2024-05-28
Block Library: Remove CSS hack for Internet Explorer 11

t-hamano:
Part of #58034
Similar to #60727

## What?

This PR removes the fallback CSS for IE11 present in the `@wordpress/block-library` package. The CSS hack for IE11 in this package is only present in the Cover block.

## Why?

Support for IE11 has been [officially removed in WordPress 5.8](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/). And since the Gutenberg plugin no longer supports WordPress 5.8, we can remove the fallback code for IE11.

Additionally, the `after` pseudo-element of the block is used for outlines, so this CSS hack for the cover block may have implications in the future.

## How?

Just removed.

## Testing Instructions

This PR does not change the behavior of the Cover block, but please be sure to check that the issue fixed in #44707 is not regressing. That is, the focus outline should exactly match the size of the cover image.

image

### DIFF
--- a/packages/block-library/src/cover/editor.scss
+++ b/packages/block-library/src/cover/editor.scss
@@ -123,10 +123,3 @@
 .color-block-support-panel__inner-wrapper > :not(.block-editor-tools-panel-color-gradient-settings__item) {
 	margin-top: $grid-unit-30;
 }
-
-
-// The frontend style.scss includes a min-height as an IE11 fix.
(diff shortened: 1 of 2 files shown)


Mamaduka (approved):

jasmussen:
Nice, thank you for your attention to detail.

Issue #60975, opened 2024-04-22
Components: remove Framer Motion


DaniGuardiola:
This is a tracking issue. There has been some discussion of removing Framer Motion (the use of `motion` especially), citing a few arguments:

- It's chunky: just the `motion` component [is 40.2kb](https://bundlejs.com/?q=framer-motion%4011.1.7&treeshake=%5B%7B+motion+%7D%5D) or [about x13 production `react`s](https://bundlejs.com/?q=react%4018.2.0&treeshake=%5B*%5D) (all gzipped and minified!).
- Pure CSS animations are more performant.
- CSS is advanced enough to make most animation use-cases simple enough to implement and maintain.
- [Often, it costs more effort than just implementing the vanilla CSS alternative.](https://github.com/WordPress/gutenberg/pull/60560#issuecomment-2078109937)
- [It seems like it won't support React 19 for a good while, as the maintainers have run into a lot of trouble](https://github.com/framer/motion/pull/2667)

## Current situation

Framer Motion is wrapped in the components package (`packages/components/src/animation/index.tsx`). A few utilities are re-exported from it, including the `motion` component.

Those wrappers are then used both internally and externally in a few places. This issue tracks those instances as we transition out of them.

## Strategy

A good start would be removing all instance of internal usage within the components package.

In a second stage, external uses within Gutenberg could be removed as well.

Then, we should set the component as deprecated and nudge potential consumers (if any) towards either using CSS directly (preferred), using a lightweight alternative like [motion.dev](https://motion.dev/), or at least using framer-motion directly, as we don't want to officially support it anymore from the components package.

Finally, the re-exports should be deleted, or if we feel like an intermediate phase is useful, replaced by mock functions that throw errors asking users to migrate before completely deleting a few versions after.

## Internal instances

- [x] DropZone (https://github.com/WordPress/gutenberg/pull/62044 in progress by @fullofcaffeine) https://github.com/WordPress/gutenberg/blob/e8878c19a389f12df941cba9eb44326dafa856f0/packages/components/src/drop-zone/index.tsx#L21-L24 
- [ ] Snackbar https://github.com/WordPress/gutenberg/blob/e8878c19a389f12df941cba9eb44326dafa856f0/packages/components/src/snackbar/list.tsx#L16-L19
- [ ] Popover https://github.com/WordPress/gutenberg/blob/e8878c19a389f12df941cba9eb44326dafa856f0/packages/components/src/popover/utils.ts#L5
[...]

PR #62044, opened 2024-05-28
Components: Remove Framer Motion from `DropZone`

fullofcaffeine:
Project: https://github.com/WordPress/gutenberg/issues/60975.

## What

Remove Framer Motion from the component and implement the previous animation using pure CSS.

## Why?

See the full list of reasons in https://github.com/WordPress/gutenberg/issues/60975. TL;DR is better runtime performance, bundle size, and following web standards, which are often leaner and simpler to maintain.

## How?

* Remove all imports and references to framer motion components. In Gutenberg, these are re-exported from a shallow wrapper called "Animate" in `@wordpress/components`.
* FIgure out how to re-implement the same animation using pure CSS (transitions or keyframes).

It might be acceptable to simplify he animation if it makes sense, provided everyone agrees with it.

## Testing Instructions

* Verify storybook
* Verify consumer features

## Screenshots or screencast 

I'll expand this section later with screencasts comparing the before / after. TBD

- [ ] Add changelog


fullofcaffeine:
This was trickier than I initially thought. :P After experimenting a bit further, I managed to make the animation based on the mount/unmount cycle of the inner `DropInZone`, but for that, I needed to create a helper wrapper component to handle the animation lifecycle. I initially tried using solely transitions, but it was easier to just rely on keyframes. This first iteration animates well but still animates the whole component as a whole:


| `Framer Motion` | CSS keyframes via wrapper component (this PR at the moment) |
|---------|---------|
| <video src="https://github.com/WordPress/gutenberg/assets/81248/ea954e3c-689b-4725-b6ea-4f0c5b7c0985" controls="controls" style="max-width: 730px;"></video> | <video src="https://github.com/WordPress/gutenberg/assets/81248/5ac3085f-c413-4743-ba58-2cb7abf97e90" controls="controls" style="max-width: 730px;"></video> |

I'll experiment with animating the foreground separately tomorrow.

As a related note, and It might be because I'm not very experienced with CSS animations, but one advantage I see with `Framer Motion` is that it's easier to hook into more specific event-based/lifecycle animations as it already abstracts that for you in its API.

fullofcaffeine:
I think this is ready for a first review pass. I managed to implement an animation with CSS that is very close to the original, notice how the inner element (the up arrow and the label) animates in sequence and a bit later than the background:

<video src="https://github.com/WordPress/gutenberg/assets/81248/fc83b972-18ea-4e2f-b602-054c3e2599d3" controls="controls" style="max-width: 730px;"></video>

It also handles the reduced motion scenario just like before:

<video src="https://github.com/WordPress/gutenberg/assets/81248/6e46c7f7-9b51-42d1-8169-fcec593128ce" controls="controls" style="max-width: 730px;"></video>

Here's the original animation when it was implemented with Framer Motion, for comparison:

<video src="https://github.com/WordPress/gutenberg/assets/81248/ea954e3c-689b-4725-b6ea-4f0c5b7c0985" controls="controls" style="max-width: 730px;"></video>

Let me know what you think!

packages/components/src/drop-zone/style.scss
```diff
@@ -1,3 +1,73 @@
+@keyframes fadeIn {
```
fullofcaffeine: I just recalled that these might conflict with other keyframes as this CSS is bundled into the final build. I'm assuming I should wait for and incorporate `:local` after https://github.com/WordPress/gutenberg/ is merged.

tyxla: I guess you mean you're waiting for #61940? If that's the case, should we mark the PR as a draft while waiting?

fullofcaffeine: Yeah, I realized after I set it as ready for review. However, AFAIK, the changes should be simple (only adding `:local`), so I think it's fine to leave it as ready for review, as it shouldn't affect the animation outcome. I can then locally scope the keyframes and do another round of tests before merging once #61940 is available on `trunk`.

packages/components/src/drop-zone/index.tsx
```diff
@@ -126,6 +113,7 @@ export function DropZoneComponent( {
 	const [ isDraggingOverElement, setIsDraggingOverElement ] =
 		useState< boolean >();
 	const [ type, setType ] = useState< DropType >();
+	//const disableMotion = true;
```
fullofcaffeine: Opsie, I just saw that. It's a leftover from a previous test I made. I will remove it.

packages/components/src/drop-zone/style.scss
fullofcaffeine: I namespace the identifiers manually for now: 540eafcda50be0a81cc0685a5321589bd6fadbef.

packages/components/src/drop-zone/animators/index.ts
```diff
@@ -0,0 +1 @@
+export { MaybeFade } from './maybe-fade';
```
fullofcaffeine: I named the module `animators` because the components here should really more of an orchestrator than something what defines the animation per se. The animation is defined in the CSS. The extraction is to make it easier to further abstract it and share it between components later if needed.

packages/components/src/drop-zone/style.scss
```diff
@@ -1,3 +1,67 @@
...
+}
+
+@keyframes components-drop-zone__fadeOut {
+	0% {
+		opacity: 1;
+	}
+	100% {
+		opacity: 0;
+	}
+}
```
tyxla: I don't see a good reason for this to be with `@keyframes`. Can't it just be a simple CSS transition?

packages/components/src/drop-zone/style.scss
```diff
@@ -1,3 +1,67 @@
...
+	}
+	50% {
+		opacity: 1;
+		transform: scale(1);
+	}
+	100% {
+		opacity: 0;
+		transform: scale(0.9);
+	}
+}
```
tyxla: Seems like another case where we could just do it with CSS `transition`. We could change the transition timing function if we want to delay the transition a bit.

packages/components/src/drop-zone/style.scss
```diff
@@ -1,3 +1,67 @@
...
+
+.components-drop-zone__fade-enter {
```
tyxla: There's another downside of doing it with keyframes rather than with a simple `transition`: if the user aborts the animation and quickly restarts it (they drag out of the drop zone and drag into it again), the animation will be reset with `@keyframes`, while with `transition` the browser will start it from where it left off (which was the case with framer motion's implementation IIUC), which will both feel better as a UX and be less work for the browser.

packages/components/src/drop-zone/animators/maybe-fade.tsx
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Handles fade-in/fade-out animation for its children if `reducedMotion` is disabled,
+ * if not, just shows/hides the children without any animation.
+ *
+ * Relies on the animations being defined in CSS styles.
+ */
+export const MaybeFade: React.FC< {
+	show: boolean;
+	children: ReactNode;
+	duration?: number;
```
tyxla: Are we using this prop for anything?

packages/components/src/drop-zone/animators/maybe-fade.tsx
```diff
@@ -0,0 +1,48 @@
...
+} > = ( { show, children } ) => {
+	const [ shouldRender, setRender ] = useState( show );
+	const reducedMotion = useReducedMotion();
```
tyxla: Instead of triggering rerenders if the user changes their reduced motion preference, we could just do it in CSS, which if we end up using transitions as I recommend below, we could do with something as simple as:

```
@media ( prefers-reduced-motion: reduce ) {
	transition: none;
}
```

packages/components/src/drop-zone/index.tsx
```diff
@@ -194,7 +143,9 @@ export function DropZoneComponent( {
 
 	return (
 		<div { ...restProps } ref={ ref } className={ classes }>
-			{ isDraggingOverElement && <DropIndicator label={ label } /> }
+			<MaybeFade show={ isDraggingOverElement || false }>
+				<DropIndicator label={ label } />
+			</MaybeFade>
```
tyxla: With the above suggestion to implement reduced motion preference with CSS, we end up not needing to create the entire `MaybeFade` component if we do something like:

```suggestion
			<div
				className={
					isDraggingOverElement
						? 'components-drop-zone__fade-enter'
						: 'components-drop-zone__fade-exit'
				}
			>
				<DropIndicator label={ label } />
			</div>
```

packages/components/src/drop-zone/animators/maybe-fade.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useState, useEffect } from '@wordpress/element';
+import { useReducedMotion } from '@wordpress/compose';
+
+/**
+ * Handles fade-in/fade-out animation for its children if `reducedMotion` is disabled,
+ * if not, just shows/hides the children without any animation.
+ *
+ * Relies on the animations being defined in CSS styles.
+ */
+export const MaybeFade: React.FC< {
```
tyxla: I think this wrapper component is not necessary, see comments below.

packages/components/src/drop-zone/index.tsx
```diff
@@ -10,84 +10,32 @@ import { __ } from '@wordpress/i18n';
...
-
-const backdrop = {
-	hidden: { opacity: 0 },
-	show: {
-		opacity: 1,
-		transition: {
-			type: 'tween',
-			duration: 0.2,
-			delay: 0,
-			delayChildren: 0.1,
```
tyxla, line 36: Are we taking this into consideration? Could be done with `transition-delay` FWIW.

tyxla (commented):
Thanks for working on it @fullofcaffeine 👍 

I think this can be greatly simplified if we:

* Use CSS transitions instead of `@keyframes`
* Support reduced motion preference with CSS

Left some more detailed comments, let me know what you think.

packages/components/src/drop-zone/animators/maybe-fade.tsx
DaniGuardiola: I feel like we should just settle for global classes with proper names, e.g. `.wp-animation_fade`, something like that. If we go that way, we should be clear on the right naming pattern from the start.

I wanted to experiment with this but you beat me to it (thank you for getting this started!).

packages/components/src/drop-zone/style.scss
fullofcaffeine: I did try with transitions, but it was tricky to get the nested animation to work right, in sequence. With keyframes, I could add an intermediate keyframe to get virtually the same animation we had before. I'll have another go with transitions and report the results here; I'll tinker with the `transition delay` to try to get the same effect.

packages/components/src/drop-zone/style.scss
fullofcaffeine: Cool, yeah, I guess this could work 👍🏻 I'll try it out.

packages/components/src/drop-zone/animators/maybe-fade.tsx
fullofcaffeine: Oh, good catch, I did use it in one of my tests, but not anymore. I'll remove it.

packages/components/src/drop-zone/animators/maybe-fade.tsx
fullofcaffeine: Oh, I had no idea there was a media query for that. Thanks for the suggestion! 👍🏻

packages/components/src/drop-zone/index.tsx
fullofcaffeine: Yes, we are; the additional keyframe at 50% in the scale / scaleOut keyframes is what implements this delay. It's very close enough to the original behavior. However, as you suggest, `transition-delay` + transition might be an easier and more intuitive and explicit way of implementing the same 👍🏻

packages/components/src/drop-zone/animators/maybe-fade.tsx
fullofcaffeine: Sounds good! I'll work on simplifying this to use transition and the media query for reduced motion. Thanks, folks!

packages/components/src/drop-zone/index.tsx
fullofcaffeine: Yep, that makes sense 👍🏻

packages/components/src/drop-zone/style.scss
fullofcaffeine: I didn't notice this in my tests, but maybe that's because the current animation is very fast, making this hard to notice. Good to know 👍🏻

packages/components/src/drop-zone/style.scss
```diff
@@ -5,31 +5,69 @@
...
+		height: 100%;
+		width: 100%;
+		display: flex;
+		background-color: $components-color-accent;
+		align-items: center;
+		justify-content: center;
+		z-index: z-index(".components-drop-zone__content");
+		text-align: center;
+		color: $white;
+		opacity: 0;
```
fullofcaffeine, line 32: If you look at the initial visual state here and for the `__content-inner`, you might notice they might be redundant with the ones defined in the `.inactive` class below, and you're right. However, it was a good compromise that I've found to handle the drop scenario without making the expressions in the component more complicated:

1. In the `shouldActivate` expression, the `isDraggingOverDocument` seems to be needed, or the drag-in sometimes is not detected.
2. However, you can't just negate that expression, as it doesn't work well to detect the **drag out**. For dragging-out, we should only check if `! isDraggingOverElement`.
3. This means that `is-active` stays applied even if you drag-out (because it includes the `isDraggingOverDocument`, which remains `true`). That's okay for dragging out as long as `inactive` is applied, which it is because when you drag out, `isDraggingOverElement` changes to `false`.
4. However, when you drop inside the zone, then `inactive` is not applied. We could change the expression there to also apply it, but it's simpler to just then let it fallback to the property here, which it does, because when you drop, then `is-active` is not applied (nor `inactive`).

fullofcaffeine:
@DaniGuardiola @tyxla  I've changed it to use CSS transition and a media query to handle the reduced motion scenario. I also further simplified it by removing the `DropInZone` component and inlining the markup inside the `DropZone` component as in the implementation, that component became redundant. The animation should behave the same as the original but do let me know if you find any differences worth fixing. Thanks again for your time and suggestions! 

Here's how it looks (and a screencast of the version in `trunk` below it for comparison):


## New implementation

<video src="https://github.com/WordPress/gutenberg/assets/81248/e1afba68-0ac0-47ff-b583-292254a9d796" controls="controls" style="max-width: 730px;"></video>

## Trunk

<video src="https://github.com/WordPress/gutenberg/assets/81248/d742a190-48bf-46f9-9240-8254b40ecfb0" controls="controls" style="max-width: 730px;"></video>

packages/components/src/drop-zone/index.tsx
```diff
@@ -181,20 +109,31 @@ export function DropZoneComponent( {
...
+			( type === 'html' && onHTMLDrop ) ||
+			( type === 'default' && onDrop ) );
+
 	const classes = clsx( 'components-drop-zone', className, {
-		'is-active':
-			( isDraggingOverDocument || isDraggingOverElement ) &&
-			( ( type === 'file' && onFilesDrop ) ||
-				( type === 'html' && onHTMLDrop ) ||
-				( type === 'default' && onDrop ) ),
-		'is-dragging-over-document': isDraggingOverDocument,
```
fullofcaffeine, line 190: I've removed these classes as I couldn't find where they were used. They don't seem to be needed for the component to function. If you know they are used somewhere, let me know, and I'll add them back.

packages/components/src/drop-zone/animators/maybe-fade.tsx
fullofcaffeine: > I feel like we should just settle for global classes with proper names, e.g. .wp-animation_fade

@DaniGuardiola I didn't try to generalize in this iteration, I think the fade transition is simple enough and very specific to this component, but let me know if you think otherwise.

packages/components/src/drop-zone/index.tsx
tyxla: Seems like there is some 3rd party plugin usage in one of BuddyPress's component plugins:

https://github.com/buddypress/bp-attachments/blob/74a4af6c8c102f08716d09453c83d97084a92d15/bp-attachments/assets/media-library/style.css#L63

packages/components/src/drop-zone/index.tsx
```diff
@@ -10,86 +10,14 @@ import { __ } from '@wordpress/i18n';
...
-		<motion.div
-			variants={ backdrop }
-			initial={ disableMotion ? 'show' : 'hidden' }
-			animate="show"
-			exit={ disableMotion ? 'show' : 'exit' }
-			className="components-drop-zone__content"
-			// Without this, when this div is shown,
-			// Safari calls a onDropZoneLeave causing a loop because of this bug
-			// https://bugs.webkit.org/show_bug.cgi?id=66547
-			style={ { pointerEvents: 'none' } }
```
tyxla, line 72: I wonder if this will need to be ported to CSS as well.

tyxla (commented):
@WordPress/gutenberg-components this one seems to be close. Could anyone pick it up and get it to the finish line?

ciampo:
I'll work on this

packages/components/src/drop-zone/index.tsx
ciampo: Added back those classnames

packages/components/src/drop-zone/index.tsx
ciampo: Moved to the Sass file and kept the original comment

ciampo:
@tyxla I updated this PR, tackling the outstanding comments and pushing some more changes with the goal of minimizing the diff with `trunk` and refactor away from `framer-motion`.

This PR should be ready for a new round of review

packages/components/src/drop-zone/style.scss
ciampo: Ended up renaming the class to `has-dragged-out` as it felt like it better represented the meaning behind it. The way I understand this change, is that it allows removing the conditional rendering of the drop zone, and it allows us to achieve the same effect just using CSS classes.

I also managed to remove some CSS associated to the `inactive` classname by applying the active styles only when `&.is-active:not(.has-dragged-out)`.

packages/components/src/drop-zone/index.tsx
tyxla: `is-dragging-over-document` could possibly be unused, I didn't spot any usage when I was doing my wpdirectory.net search. Sounds good as a follow-up cleanup too.

tyxla (approved):
Looks and works great 👍 🚀 

Thanks, @fullofcaffeine for the original work and @ciampo and @DaniGuardiola for finishing this one up 🙌

ciampo:
Apologies for not including the [list of contributors](https://github.com/WordPress/gutenberg/pull/62044#issuecomment-2143195350) to the the commit message — it seems like this change was introduced while I was away. I'll make sure I'll do so from now on

PR #62046, opened 2024-05-28
Background images: add defaults for background size

ramonjd:
## What?

See: https://github.com/WordPress/gutenberg/issues/61928#issuecomment-2134387541

This PR:

- sets a default background position of `"50% 0"` for uploaded images when background size is toggled to "Tile"
- resets background position in between toggling
- defines a placeholder of "Auto" for the background image width image value.

https://github.com/WordPress/gutenberg/assets/6458278/46fc3deb-6988-45ed-9dda-bd4e92cd54b2

## Why?
To increase the chance that the image's focus point is visible.

## Testing Instructions

1. Insert a Group block and add a new background image.
2. Toggle the background size to "Tile"
3. Check that the position is set to "50% 0"
4. Toggle between size options, e.g., Cover, Contains, Tile. 
5. Check that the position resets between toggling (should it?)
6. With the size set to "Tile", upload an image or select a new one from the Media library. Check that the position is set to "50% 0"
7. Change the default position X Y values. Upload an image or select a new one from the Media library. Your position should be preserved.
8. For the Tile option, observe that the placeholder for image width input field underneath is "Auto"

Finally, to ensure that no default background position is set for theme.json-defined background images, set a background image in theme.json and toggle between the size settings in the site editor.

https://github.com/WordPress/gutenberg/assets/6458278/e7b6ec82-e7bd-442a-b498-b796a40cba05

### DIFF
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -245,12 +245,22 @@ function BackgroundImageToolsPanelItem( {
 			return;
 		}
 
+		const sizeValue = style?.background?.backgroundSize;
+		const positionValue = style?.background?.backgroundPosition;
+
 		onChange(
-			setImmutably( style, [ 'background', 'backgroundImage' ], {
-				url: media.url,
-				id: media.id,
-				source: 'file',
-				title: media.title || undefined,
+			setImmutably( style, [ 'background' ], {
+				...style?.background,
+				backgroundImage: {
+					url: media.url,
+					id: media.id,
+					source: 'file',
+					title: media.title || undefined,
+				},
+				backgroundPosition:
+					! positionValue && ( 'auto' === sizeValue || ! sizeValue )
+						? '50% 0'
+						: positionValue,
 			} )
 		);
 	};
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -448,7 +448,7 @@ function BackgroundSizeToolsPanelItem( {
 			 * when the toggle switches to "Tile". This is to increase the chance that
 			 * the image's focus point is visible.
 			 */
-			if ( !! style?.background?.backgroundImage?.id ) {
+			if ( !! style?.background?.backgroundImage?.id && ! nextPosition ) {
 				nextPosition = '50% 0';
 			}
 		}
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -441,6 +441,7 @@ function BackgroundSizeToolsPanelItem( {
 			next === 'auto'
 		) {
 			nextRepeat = undefined;
+			next = !! style?.background?.backgroundImage?.id ? '50%' : next;
 		}
 
 		/*
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -430,10 +430,12 @@ function BackgroundSizeToolsPanelItem( {
 
 		if ( next === 'contain' ) {
 			nextRepeat = 'no-repeat';
+			nextPosition = undefined;
 		}
 
 		if ( next === 'cover' ) {
 			nextRepeat = undefined;
+			nextPosition = undefined;
 		}
 
 		if (
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -22,12 +22,11 @@ import {
 	__experimentalItemGroup as ItemGroup,
 	__experimentalHStack as HStack,
 	__experimentalTruncate as Truncate,
-	__experimentalParseQuantityAndUnitFromRawValue as parseQuantityAndUnitFromRawValue,
 } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
 import { store as noticesStore } from '@wordpress/notices';
 import { getFilename } from '@wordpress/url';
-import { useCallback, Platform, useRef, useState } from '@wordpress/element';
+import { useCallback, Platform, useRef } from '@wordpress/element';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { focus } from '@wordpress/dom';
 import { isBlobURL } from '@wordpress/blob';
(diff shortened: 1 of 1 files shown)


jasmussen:
Thanks for keeping on this, and sweating the details. This is rapidly improving.

This PR solves the issue of background images by default being low resolution. But [reading up on the properties](https://www.w3schools.com/cssref/css3_pr_background-size.php), and [the back compat conversations](https://github.com/WordPress/gutenberg/pull/62000#issuecomment-2134172276), I wonder: is it possible to unset this default value if you wanted _no_ value, i.e. `auto`?

ramonjd:
> is it possible to unset this default value if you wanted no value, i.e. auto?

Good question! I did a quick video and added it to the description - the video suggests that yes, it's possible, but only via clearing the input field. 

I haven't tested it very thoroughly so I could be off track.

jasmussen:
Thanks for the video. Bouncing back [from here](https://github.com/WordPress/gutenberg/pull/62000#issuecomment-2134470630), I'm really wondering what a good default is. I briefly misremembered "Cover" as behaving how "Contain" with repeat did, which I thought was the preferable one, turns out that's not the case. So thinking between Tile and Contain as defaults, there's no clear win.

**Tile by default**

If we add:

* `background-size: 100%;`
* `background-position: 50% 0%;`

That would give us this:

properties

That would be a good experience when you add a large image, and you'd be able to configure it from there.

But if you add a low-resolution image with the specific intent to tile it, you'd want to add a pixel value that's half the dimensions of the image itself, to ensure the 2x resolution. I.e. a 100x100 tiled background should be 50px in background-size, and in fact the percentage would make it fill the screen, behave like the aforementioned Contain + repeat that I mentioned.

**Contain by default**

As noted, this would provide a good experience if you add a large landscape image. But it would fall apart if you wanted a small tiled background, or used a portrait image instead.

**Default values**

For specificity reasons and to not add technical debt, any default extra CSS we output, we have to be very careful. It may provide a better one-click default experience, but it comes at the cost of adding a tiny amount of extra burden when that default is _not_ what you want. To that end, and because of the above test cases, it's not clear that adding a default `background-size` is a good choice, after all:

* By using percentages, we are emulating the Contain behavior.
* 50% does not guarantee at least 2x resolution, neither does 100%.
* The only thing that would guarantee 2x resolution would be for you to manually enter PX values that are half the intrinsic dimensions of the image.

**Best choice?**

At this point, it's not clear to me which value is the best to default to. It's probably either _contain + repeat_, or _tile_, but without the added default background-size property after all.

The only thing that seems clearer after this exercise is that it may be good to add a default background-position, e.g. this:
center top

This should center/top align the background image, which when tiled would go well with a centered layout, which is _arguably_ the most common one. Since the Cover block has 50%/50% values by default, it doesn't seem unreasonable to me to add defaults here as well:

cover defaults

So: maybe we keep the tile, but add 50%/0% background position values. What do you think?

andrewserong:
> So: maybe we keep the tile, but add 50%/0% background position values. What do you think?

If we apply that only when a user adds an image, that sounds good to try to me! From what you've described, of all the available options, this gets as a little closer to pleasing defaults without boxing us in too much, IMO

ramonjd:
Thanks for all the testing @jasmussen 🙇🏻 

 I'll play around in this PR later in the week.

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -426,13 +426,16 @@ function BackgroundSizeToolsPanelItem( {
 	const updateBackgroundSize = ( next ) => {
 		// When switching to 'contain' toggle the repeat off.
 		let nextRepeat = repeatValue;
+		let nextPosition = positionValue;
 
 		if ( next === 'contain' ) {
 			nextRepeat = 'no-repeat';
+			nextPosition = undefined;
```
ramonjd, line 443: Setting to `undefined` between toggles so that `"50% 0"` doesn't persist.

This means however, that any user-defined position values are lost as well.

I'm wondering if we should save the value in local state and only reset when a default value is detected, e.g, `"50% 0"`

andrewserong: IMO it's fine to reset the position when switching between sizes as you're kind of resetting how the image will appear anyway. Could be a follow-up PR if we think it's needed after all?

andrewserong (commented):
This is testing nicely for me when switching between background sizes on an already uploaded image. However, if I go to add an image from scratch, the default isn't added when the image is attached. I'm wondering if we also need to add a line somewhere around here in `onSelectMedia`:

https://github.com/WordPress/gutenberg/blob/f263cf8ec832289fdd676653010734780cbd5b65/packages/block-editor/src/components/global-styles/background-panel.js#L248-L255

I.e. check if `style?.background?.backgroundSize` is unset, and if so, set style.background.backgroundPosition to `50% 0`?

What do you think? I'm not sure if we've added other values when media is selected before, so wasn't sure if there's any potential edge cases to consider 🤔

jasmussen:
Tested this branch per the review, but I'm not seeing the defaults added when I add a background image:

Screenshot 2024-05-30 at 08 40 41

andrewserong:
Snap! Joen and I noticed the same thing at the same time 😄

jasmussen (approved):
Have to step away from this for a bit, so can't follow. But the decision about defaulting to center top position for tile, is one I can support, and this PR moves in that direction. So behavior-wise, a 👍 👍, I'll defer to Andrew and others for a code check.

andrewserong:
> I'll defer to Andrew and others for a code check.

Code-wise it's looking good to me. We can either land as-is, and add the behaviour when the image is first uploaded in a separate PR, or hold off on merging until that's added in. Whichever you prefer, @ramonjd!

ramonjd:
Thanks for the speedy reviews @jasmussen and @andrewserong 

I'll aim to add the [missing functionality](https://github.com/WordPress/gutenberg/pull/62046#pullrequestreview-2087364640) in this PR. Should be a one/two liner I think. 

Cheers!

ramonjd:
> Tested this branch per the review, but I'm not seeing the defaults added when I add a background image:

Okay! I've added the change and updated the test instructions. 

The default of `50% 0` is set when selecting/uploading an image via the media library. 

But if the user has already set a position and not toggled from "Tile", the user value is preserved.

Thanks again!

https://github.com/WordPress/gutenberg/assets/6458278/e84040bf-ab86-4a4e-9958-dbe3e1c154f1

ramonjd:
Thanks for catching that final feature bug. 

I'll merge and we can carry onwards and upwards!

andrewserong:
Nice one, thanks for fixing that up! 🎉

PR #31822, opened 2021-05-14
Replace hardcoded CSS_UNITS and inherit values from theme.json

aristath:
## Description
Currently, most controls have units hardcoded in `CSS_UNITS` constants, and many components don't use the units defined in `theme.json` files.

We already had a `useCustomUnits` function inside `packages/block-editor/src/components/unit-control/`, so this PR does the following:
* Move `useCustomUnits` to `packages/components/src/unit-control/utils.js`
* Expose `useCustomUnits` in `@wordpress/components`
* Remove hardcoded `CSS_UNITS` arrays, and replace them with calls to `useCustomUnits`. The calls check for `theme.json` settings (using `useSetting`), and if no settings are located then fallback to an array of units (this array was extrapolated from the previous `CSS_UNITS` constants).
* Added an array (`ALL_CSS_UNITS`) containing **all** existing CSS units. This way users can define their arrays like `["px","em"]` and all labels will be properly populated - even if they decide to use unusual units like `cm`, `in` or `pc` :shrug: 

While doing the above, I realized that `spacing.units` doesn't make sense for everything... Spacing refers to padding & margins, not widths. So In some instances (like for example Layout settings & column widths), instead of checking for `spacing.units` I'm checking for `layout.units`.

Overall, this PR cleans up the code a bit and feels like an improvement, increasing consistency.

## How has this been tested?
Checked in an FSE theme with and without values defined in `theme.json` for `settings.spacing.units` & `settings.layout.units`.
Confirmed that it works as expected for layout, padding, margin, column-width, cover block, search block. Checked both the post-editor & site-editor.

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #62049, opened 2024-05-28
UnitControl: preserve order of incoming units

ramonjd:
## What?

Testing passing an allowedUnitValues prop to UnitControl, which intends (eventually) to act in some way like `useCustomUnits` hook, but for now just filters the default units and preserves the order.

See: https://github.com/WordPress/gutenberg/issues/61928#issuecomment-2134387541

See also: https://github.com/WordPress/gutenberg/pull/31822#discussion_r633280823

## Why?

Consumers should be able to determine the order of the units in the control.

### DIFF
--- a/packages/components/src/unit-control/index.tsx
+++ b/packages/components/src/unit-control/index.tsx
@@ -25,7 +25,11 @@ import {
 } from './utils';
 import { useControlledState } from '../utils/hooks';
 import { escapeRegExp } from '../utils/strings';
-import type { UnitControlProps, UnitControlOnChangeCallback } from './types';
+import type {
+	UnitControlProps,
+	UnitControlOnChangeCallback,
+	WPUnitControlUnit,
+} from './types';
 import { useDeprecated36pxDefaultSizeProp } from '../utils/use-deprecated-props';
 
 function UnforwardedUnitControl(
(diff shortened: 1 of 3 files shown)


packages/components/src/unit-control/index.tsx
```diff
@@ -71,11 +76,26 @@ function UnforwardedUnitControl(
...
 	const nonNullValueProp = valueProp ?? undefined;
 	const [ units, reFirstCharacterOfUnits ] = useMemo( () => {
-		const list = getUnitsWithCurrentUnit(
+		let list = getUnitsWithCurrentUnit(
 			nonNullValueProp,
 			unitProp,
 			unitsProp
 		);
+		// If allowedUnitValues is provided, we need to filter the list of units.
+		if ( allowedUnitValues?.length ) {
```
ramonjd, line 85: So this is a huge hack and just here to illustrate the proposal.

The goal is for this component to honor the order of units. For example, if I wanted to have `%` appear at the top of the unit list.

I'm just not sure whether to do it in the component or as part of `useCustomUnits`.

https://github.com/WordPress/gutenberg/pull/31822#discussion_r633280823 suggests that `useCustomUnits` might be the wrong place to do it.

ramonjd: cc @ciampo for advice - non-urgent, when you get time 🙇🏻

ramonjd:
Closing this as stale. I might come back to this idea one day.

ciampo:
Sorry for the delay, I was on a parental leave and I'm still catching up with the notification queue.

Happy to discuss the new approach if/when the time comes!

Issue #61084, opened 2024-04-25
DataViews and DataForm: Extensibility


youknowriad:
This document keeps track of all the work and tasks to extend thel DataForm and DataViews components (Fields). These components are used to power all/most screens in Admin Redesign (side editor first and expanded to more screens later). Users should be able to extend these screens in all sort of ways. We should able to allow field, actions registration/unregistration, custom layouts, default layouts for lists and forms and a lot more.

## CPT / Entity based extensibility APIs

The extensibility APIs should be consistent, and PHP first. you should be able to do 80% of the modification using simple PHP APIs and for very specific/advanced cases be able to write React components to extend the default behavior. Most of the following APIs involve extending `register_post_type/register_taxonomy` function with new possibilities. But for WordPress entities that are not post types or taxonomies (users, comments...), a similar PHP API is going to be needed.

 - [ ] Support defining the support layouts per entity (list, table, grid)
 - [ ] Support defining the default fields and default configuration for each layout per entity.
 - [ ] Support defining the default quick edit (form) fields for each entity.
 - [ ] Support defining the default rendering mode for each entity (template-locked vs post only).
 - [ ] Support registering additional fields per entity.
   - [x] client-side https://github.com/WordPress/gutenberg/pull/67175
   - [ ] https://github.com/WordPress/gutenberg/issues/74865
 - [ ] Support registering additional actions per entity. #62052
 - [ ] Support registering "default views" per entity.

## More global extensibility APIs.

 - [ ] Support registering custom DataViews layouts.
 - [ ] https://github.com/WordPress/gutenberg/issues/74858
 - [ ] https://github.com/WordPress/gutenberg/issues/74856

## Common

 - [ ] Versioning the APIs #67274 

### Low priority

 - [ ] Audit the action "ids" and bring some consistency there.
 - [ ] Audit the action code bases and improve things like notices and error handling (most of these can be absorbed into the framework I think)

fabiankaegy:
Thanks for opening this issue :) 

I want to reference back to this comment / thread in the original ["Expanding the `Table` component"](https://github.com/WordPress/gutenberg/issues/53233) issue from a year ago: 

https://github.com/WordPress/gutenberg/issues/53233#issuecomment-1665263796

oandregal:
Related ticket about (duplicate post) actions extensibility https://github.com/WordPress/gutenberg/issues/61083 considering user capabilities, ability to filter by plugins, etc.

youknowriad:
[...]

PR #62052, opened 2024-05-28
DataViews: Bootstrap Actions Extensibility API

youknowriad:
## What?

This PR bootstraps an API to allow third-party developers to register and unregister post type actions. 
Right now post type actions appear in two places:

 - Site editor dataviews.
 - Sidebar of the post editor (or site editor)

Actions allow users to interact with the selected, active or bulk selected entity/post/record. The API will be initially only available in the Gutenberg plugin for some time (until we get testing, feedback, iterations...)

As mentioned on the [issue](https://github.com/WordPress/gutenberg/issues/61084#issuecomment-2134613932), initially I'm adding this API to the editor package but I think a dedicated package is probably going to be needed quickly.

you can test the API by opening the editor (or dataviews) and calling the function like so:

```js
wp.editor.registerEntityAction( 'postType', 'page', {
	id: 'myplugin/test-action',
	label: 'Test Action',
	isPrimary: true,
	icon: 'admin-site',
	supportsBulk: false,
	callback( posts ) {
		alert( 'Test Action: ' + ( posts[0]?.title?.rendered ?? posts[0]?.title ) );
	},
} );
```

The action will appear in the actions dropdown.

**Note**

 - Default actions have not been refactored to use the API, which means you can't unregister them yet. 
 - We'll probably have to support "*" as a "kind" or "name" in order to register an action for multiple post types or entities.

### DIFF
--- a/packages/editor/src/dataviews/store/private-actions.ts
+++ b/packages/editor/src/dataviews/store/private-actions.ts
@@ -1,0 +1,30 @@
+/**
+ * WordPress dependencies
+ */
+import type { Action, AnyItem } from '@wordpress/dataviews';
+
+export function registerEntityAction< Item extends AnyItem >(
+	kind: string,
+	name: string,
+	config: Action< Item >
+) {
+	return {
+		type: 'REGISTER_ENTITY_ACTION' as const,
+		kind,
+		name,
+		config,
+	};
+}
+
+export function unregisterEntityAction(
+	kind: string,
+	name: string,
+	actionId: string
+) {
+	return {
+		type: 'UNREGISTER_ENTITY_ACTION' as const,
+		kind,
+		name,
+		actionId,
+	};
+}
--- a/packages/editor/src/dataviews/store/private-selectors.ts
+++ b/packages/editor/src/dataviews/store/private-selectors.ts
@@ -1,0 +1,8 @@
+/**
+ * Internal dependencies
+ */
+import type { State } from './reducer';
+
+export function getEntityActions( state: State, kind: string, name: string ) {
+	return state.actions[ kind ]?.[ name ] ?? [];
+}
--- a/packages/editor/src/store/reducer.js
+++ b/packages/editor/src/store/reducer.js
@@ -7,6 +7,7 @@ import { combineReducers } from '@wordpress/data';
  * Internal dependencies
  */
 import { EDITOR_SETTINGS_DEFAULTS } from './defaults';
+import dataviewsReducer from '../dataviews/store/reducer';
 
 /**
  * Returns a post attribute value, flattening nested rendered content using its
@@ -402,4 +403,5 @@ export default combineReducers( {
 	listViewPanel,
 	listViewToggleRef,
 	publishSidebarActive,
+	dataviews: dataviewsReducer,
 } );
--- a/packages/dataviews/src/index.ts
+++ b/packages/dataviews/src/index.ts
@@ -1,3 +1,4 @@
 export { default as DataViews } from './dataviews';
 export { VIEW_LAYOUTS } from './layouts';
 export { filterSortAndPaginate } from './filter-and-sort-data-view';
+export type * from './types';
--- a/packages/editor/README.md
+++ b/packages/editor/README.md
@@ -1504,6 +1504,18 @@ _Returns_
 
 Undocumented declaration.
 
+### registerEntityAction
+
+Registers a new DataViews action.
+
+This is an experimental API and is subject to change. it's only available in the Gutenberg plugin for now.
+
+_Parameters_
+
+-   _kind_ `string`: Entity kind.
+-   _name_ `string`: Entity name.
+-   _config_ `Action`: Action configuration.
+
 ### RichText
 
 > **Deprecated** since 5.3, use `wp.blockEditor.RichText` instead.
@@ -1588,6 +1600,18 @@ Undocumented declaration.
 ### transformStyles
 
 Undocumented declaration.
+
+### unregisterEntityAction
+
+Unregisters a DataViews action.
+
+This is an experimental API and is subject to change. it's only available in the Gutenberg plugin for now.
+
+_Parameters_
+
+-   _kind_ `string`: Entity kind.
+-   _name_ `string`: Entity name.
+-   _actionId_ `string`: Action ID.
 
 ### UnsavedChangesWarning
 
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -1030,11 +1030,13 @@ export const duplicateTemplatePartAction = {
 };
 
 export function usePostActions( postType, onActionPerformed ) {
-	const { postTypeObject } = useSelect(
+	const { defaultActions, postTypeObject } = useSelect(
 		( select ) => {
 			const { getPostType } = select( coreStore );
+			const { getEntityActions } = unlock( select( editorStore ) );
 			return {
 				postTypeObject: getPostType( postType ),
+				defaultActions: getEntityActions( 'postType', postType ),
 			};
 		},
 		[ postType ]
@@ -1072,6 +1074,7 @@ export function usePostActions( postType, onActionPerformed ) {
 				? deletePostAction
 				: trashPostAction,
 			! isTemplateOrTemplatePart && permanentlyDeletePostAction,
+			...defaultActions,
 		].filter( Boolean );
 
 		if ( onActionPerformed ) {
@@ -1117,6 +1120,7 @@ export function usePostActions( postType, onActionPerformed ) {
 
 		return actions;
 	}, [
+		defaultActions,
 		isTemplateOrTemplatePart,
 		isPattern,
 		postTypeObject?.viewable,
--- a/packages/editor/src/dataviews/api.js
+++ b/packages/editor/src/dataviews/api.js
@@ -1,0 +1,55 @@
+/**
+ * WordPress dependencies
+ */
+import { dispatch } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../lock-unlock';
+import { store as editorStore } from '../store';
+
+/**
+ * @typedef {import('@wordpress/dataviews').Action} Action
+ */
+
+/**
+ * Registers a new DataViews action.
+ *
+ * This is an experimental API and is subject to change.
+ * it's only available in the Gutenberg plugin for now.
+ *
+ * @param {string} kind   Entity kind.
+ * @param {string} name   Entity name.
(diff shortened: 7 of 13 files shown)


fabiankaegy:
@youknowriad is there any way to be aware of the context where the post action is rendered/added? 

I wonder whether it makes sense to be able to control whether an action should show up in the data view vs in the post actions of a single post? 🤔

youknowriad:
> @youknowriad is there any way to be aware of the context where the post action is rendered/added?

There are some indirect ways, we may want to introduce something like "scope" config later but for now, I believe we should encourage folks to write agnostic actions. Could be one follow-up to consider.

mcsf:
Thanks for getting this started. Seems like a reasonable approach; let's see where it leads us.

Mamaduka:
I love the idea, and this style of API should be more future-proof than using a filter or a SlotFill.

packages/editor/src/store/reducer.js
```diff
@@ -402,4 +403,5 @@ export default combineReducers( {
 	listViewPanel,
 	listViewToggleRef,
 	publishSidebarActive,
+	dataviews: dataviewsReducer,
```
ntsekouras, line 406: This is named like that because the plan is to have a separate package eventually? Because the actions are used beyond the Data Views, so I don't really get why name it like that.

youknowriad: yes, I can rename, what would you suggest? I guess you can think of it as the package name.

ntsekouras:
Thanks Riad! Have you considered registering an action by providing the `id` and the action object with extra prop of `postTypes`. Then this `postTypes` info could be baked in checks like `isEligible` to show/hide actions and we could avoid the wildcard `*` pattern, that I feel it usually creates more problems than it solves.

What is the expectation and gains of having the `kind` as param?

youknowriad:
> Thanks Riad! Have you considered registering an action by providing the id and the action object with extra prop of postTypes. Then this postTypes info could be baked in checks like isEligible to show/hide actions and we could avoid the wildcard * pattern, that I feel it usually creates more problems than it solves.

This is a good alternative suggestion.

> What is the expectation and gains of having the kind as param?

I aligned the API with the entities abstraction in core-data, so you can have actions and fields for things like site, taxonomies,... that are not post types.

---

Based on this, I feel like the entities abstraction mapping make more sense because I see WP providing these dataviews with the same filters as well. WDYT?

ntsekouras:
>I aligned the API with the entities abstraction in core-data, so you can have actions and fields for things like site, taxonomies,... that are not post types.

For start I think the structure we'll use in store, doesn't matter much right now and could be updated anytime. For the consumers facing part though (`register API`). Do we really need to expose these things though?

Maybe we have to answer how related is the core-data abstraction with the actions API. In actions the callbacks are specified inside the action's props and they could do anything (ex view revisions). I think the core-data abstraction is used to help with the entity handling mostly. 

Also, can we have actions that apply for more than one entity? For example could the rename action be almost identical (at least in UI) for post types and taxonomies. Probably it could, but maybe it should be kept separate. How could this affect the API?

My first thoughts would be that we could keep an internal single registration API if we want, but expose specific ones like registerPostTypeAction, registerTaxonomyAction, etc.., with respective fields ( postTypes, terms, etc..) 

I'm rambling a bit, but just thinking out loud to get more thoughts on this.

>Based on this, I feel like the entities abstraction mapping make more sense because I see WP providing these dataviews with the same filters as well. 

Can you expand a bit on this?

packages/editor/src/dataviews/store/private-actions.ts
```diff
@@ -0,0 +1,30 @@
...
+ * WordPress dependencies
+ */
+import type { Action, AnyItem } from '@wordpress/dataviews';
+
+export function registerEntityAction< Item extends AnyItem >(
+	kind: string,
+	name: string,
+	config: Action< Item >
+) {
+	return {
```
ntsekouras, line 11: I know this is the bootstrap PR, but we should add validation, even if it's in a follow up.

youknowriad:
I think the main action is:

 - how to register an action for a taxonomy?
 - how to register an action for a site?
 - how to register an action for a post type?
 - how to register an action for all post types?

Personally, it seems clear to me that we shouldn't try to reuse actions between taxonomies and post types for instance, in that sense a global action with "taxonomies" or "post types" configs, doesn't really make sense to me. This is the main reason that I think the proposed API is a bit better.

Now there's a case that we have where we want the same action for multiple post types. So I was thinking about

```
registerAction( 'postTypes', '*', config )
registerAction( 'postTypes', [ 'post', 'page' ], config )
```

to address this.

packages/editor/src/dataviews/store/private-actions.ts
youknowriad: That's one of the reasons I added the types, for us at least, everything is typed, so there's less need for validation but indeed, validation for third-parties would be great. I wonder if there's a runtime lib that can use typescript types to generate validation automatically.

youknowriad: I added a todo list for this to the issue.

ntsekouras:
I still think the proposed API is a bit more verbose but we can always add some wrapper functions(syntactic sugar) like registerPostTypeAction, etc.. Not that important for now.

We can start with this, but I think these should be addressed here:
1. Support the array of `names` 
2. Support the `all` or `*`.

I'll say again that the wildcard API is something I don't like and I think we'll end up needing extra props eventually to satisfy cases like: include all, except one.. I guess that's fine though, or in such cases the `except` part could be in `isEligible`? 

With actions being extensible, we would also need to add an `order` prop to actions and not depend on order of registration - not for here..

youknowriad:
> With actions being extensible, we would also need to add an order prop to actions and not depend on order of registration - not for here..

We have always refrained from having order in our APIs to avoid plugins competing for positions. So I think for this one, we should be consistent and continue until that decision is reversed.

ntsekouras:
>We have always refrained from having order in our APIs to avoid plugins competing for positions.

Yeah, I know it's a tough one, but usually we would overcome this by having a slot and render the third party components there. We'll see in the future, but I guess for example that the `delete` action should be last and we would have similar cases with other core actions.

mcsf:
I went through @youknowriad's and @ntsekouras's discussion and have some loose thoughts:

* Agreed that the store structure isn't particularly important, whereas the API is. The store is plastic, and it won't be dealing with much data anyway.

* On the relationship between entities, actions and `core-data`:
  * Is the idea to turn this experiment into a module tied to `editor`? If so, what are the pros and cons of keeping it close to `editor`? Unless that choice of package was purely arbitrary for the sake of the experiment.
  * I can maybe get behind the `registerEntityAction` signature mimicking the original entity structure, but in that case this brings the Actions domain much closer to entities — possibly to the point of letting them be absorbed into the `core-data` module, or making actions a tiny "satellite" module of `core-data`.
  * That said, the whole entity concept is a difficult chunk of the codebase for developers to learn. Is it worth exposing them to it just to register actions tied to (initially) posts?
  * Kind-specific functions like `registerPostAction` make a bit more sense:
    * They are still technically a reflection of entities, but for developers it should feel much more intuitive (`(postType, action) => {}`).
    * In the highly dynamic world of WP-style JS, dedicated functions for each kind act as guard rails: the kind-specific function either exists or not. We could of course just have the generic `registerEntityAction` with `kind: 'postType' | 'site' | ...`, but in that case we would need _runtime_ validation too — you know this, it's just a reminder.

> Also, can we have actions that apply for more than one entity? For example could the rename action be almost identical (at least in UI) for post types and taxonomies. Probably it could, but maybe it should be kept separate. How could this affect the API?

Thinking about post types in the traditional PHP sense: CPTs don't manually re-register/re-implement every bit of functionality that they want from the base `'post'` type. Instead, they have `supports` at their disposal, allowing easy opting in and out of sets of features. Is this concept helpful to us here? For instance, could there be a base set of actions automatically applied to CPTs — e.g. view, edit, rename (if viewable, if editable, etc.) so that devs don't need to explicitly register them? Or, if not, if everything is opt-in, there should be a way for third parties to reuse core actions (`registerPostAction( 'my_cpt', CORE_POST_ACTIONS.view )`).

Finally, I don't hold this opinion strongly, but `'*'` might be one of those "clever" things that we come up with that are actually too blunt as tools that don't really help anyone. In PHP land we commonly use basic loops to achieve similar results, thereby avoiding the pitfalls of wildcards, such as accidental contamination of forgotten post types. In other words, this is enough: `for (const type of TYPES) registerPostAction(type, myAction)`.

youknowriad:
> Is the idea to turn this experiment into a module tied to editor? If so, what are the pros and cons of keeping it close to editor? Unless that choice of package was purely arbitrary for the sake of the experiment.

I think inherently the actions (and fields) are closer to entities than "editor". Some entities can be edited in forms outside the block editor for instance, and still have actions and fields. So for me, the usage of "editor" is just a temporary hack. I think this API makes more sense in "core-data" or its own dedicated package.

> That said, the whole entity concept is a difficult chunk of the codebase for developers to learn. Is it worth exposing them to it just to register actions tied to (initially) posts?

I agree but I think we shouldn't try to create a new concept now. How would you tie registering "site actions" into this API without inventing a new concept (or picking entities)

mcsf:
> How would you tie registering "site actions" into this API without inventing a new concept (or picking entities)

We'd either assume the compromise and expose the entities model to users, or mediate via entity-specific functions. Don't see a better way.

mcsf (approved):
I don't want to block this with the illusion of indecision. Mapping the API to the entities domain seems fine, and we can always make adjustments.

youknowriad:
@annezazu @fabiankaegy Going to merge this. It would be cool to try to get some outreach around this at some point.

I wonder whether we should first make the core actions removable.

annezazu:
Working on the outreach piece. Make Core post in the work and will ensure this is shared for this week's dev chat meeting. Awesome job getting this done!

packages/editor/src/dataviews/store/private-selectors.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Internal dependencies
+ */
+import type { State } from './reducer';
+
+export function getEntityActions( state: State, kind: string, name: string ) {
+	return state.actions[ kind ]?.[ name ] ?? [];
```
Mamaduka, line 7:
```
return state.actions[ kind ]?.[ name ] ?? EMPTY_ARRAY;
```

Let's return a stable fallback reference.

Fixes the "The 'useSelect' hook returns different values when called with the same state and parameters. This can lead to unnecessary rerenders. " warning.

ntsekouras: https://github.com/WordPress/gutenberg/pull/62536

Issue #59592, opened 2024-03-05
Block editor should work in "headless" mode, without Edit UI mounted


jsnajdr:
When working on lazy loading block `edit` functions in #55585, I am dealing with a situation where a block already exists (is part of the block tree in the `core/block-editor` store), but its `<Edit>` UI is not yet mounted, because the `edit` component is being lazy loaded. I am trying to come up with a reasonable "universal edit placeholder" that could be displayed while the real `edit` is loading, and that could handle keyboard typing in the interim period.

While working on this, one of the surprising findings was that the `edit` component doesn't only provide UI for editing, but also defines some basic block behavior. In MVC language, it's not only "view", but also implements part of the "model". Architecturally, that's bad. Some examples:

**You can't insert an inner block without edit UI mounted:** If you dispatch the `insertBlocks` or `replaceBlocks` action in the `core/block-editor` store, and the edit UI is not mounted, these actions will fail, they won't insert any new blocks. Because the `canInsertBlockType` permission check will return `false` -- it determines whether the target block supports inner blocks, and it detects that by checking for presence of `getBlockListSettings` for that block. But block list settings are set by the `edit` React UI, namely the `InnerBlocks` component, which calls `updateBlockListSettings` on mount.

**Default templates are not inserted.** Many blocks, when inserted from the slash inserter, will not insert just the block itself, but will also create some default inner structure. List has one inner List Item. Quote and Cover has a Paragraph. Social Links have a set of four default social icons. These are inserted only when the `Edit` UI is mounted. Because the insertion is done by an on-mount effect inside `useInnerBlockTemplateSync` called in `InnerBlocks`.

Inserting blocks with prefix transforms (`* ` -> list, `> ` -> quote, ...) does the same job differently -- creating the inner blocks is directly inside the transform functions. Therefore, prefix transforms work correctly when headless.

**Splitting and merging list items:** List items have custom merge and split behavior, mostly to account for nested sub-lists, but store actions like `mergeBlocks` don't know about it because it's defined inside the `edit` UI and passed as props to `RichText`. Ideally, all this behavior should be present in the `merge` function defined on the block type registration.

Then the merging and splitting could be done headlessly, without `edit` UI mounted. After all, the `core/block-editor` store fully owns and manages the selection, i.e., it knows where the text cursor is, and doesn't need any `edit` UI for that. The `edit` UI should only bind the split/merge operations to keys like Enter or Backspace.

[...]

PR #62059, opened 2024-05-28
useUploadMediaFromBlobURL: Prevent duplicate uploads in StrictMode

Mamaduka:
## What?
PR added an extra check to the  `useUploadMediaFromBlobURL` to prevent triggering the upload process multiple times in development mode.

The bug was discovered via #61943.

## Why?
The hook didn't account for the double effect run in the `StrictMode`; the effect can't cancel the in-flight upload via the cleanup method.

~~P.S. The hooks should use the `useEffectEvent` hook when it ships in React.~~

## Testing Instructions
1. Make sure `SCRIPT_DEBUG` is set to true.
2. Enabled React `StrictMode` for the post editor.
3. Create a post.
4. Drag an image on canvas.
5. Confirm that only one `POST` request is triggered to upload the image.

<details><summary>StrictMode patch</summary>
<p>

```diff
diff --git packages/edit-post/src/index.js packages/edit-post/src/index.js
index 1e0b3fe7d4d..6e4a220f92f 100644
--- packages/edit-post/src/index.js
+++ packages/edit-post/src/index.js
@@ -7,7 +7,7 @@ import {
 	__experimentalRegisterExperimentalCoreBlocks,
 } from '@wordpress/block-library';
 import deprecated from '@wordpress/deprecated';
-import { createRoot } from '@wordpress/element';
+import { StrictMode, createRoot } from '@wordpress/element';
 import { dispatch, select } from '@wordpress/data';
 import { store as preferencesStore } from '@wordpress/preferences';
 import {
@@ -131,12 +131,14 @@ export function initializeEditor(
 	window.addEventListener( 'drop', ( e ) => e.preventDefault(), false );
 
 	root.render(
-		<Editor
-			settings={ settings }
-			postId={ postId }
-			postType={ postType }
-			initialEdits={ initialEdits }
-		/>
+		<StrictMode>
+			<Editor
+				settings={ settings }
+				postId={ postId }
+				postType={ postType }
+				initialEdits={ initialEdits }
+			/>
+		</StrictMode>
 	);
 
 	return root;
``` 

</p>
</details> 

Same.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/240569/f50668fe-2a5f-48a3-aee6-ca6b34d6b31f

### DIFF
--- a/packages/block-library/src/utils/hooks.js
+++ b/packages/block-library/src/utils/hooks.js
@@ -33,6 +33,7 @@ export function useCanEditEntity( kind, name, recordId ) {
  */
 export function useUploadMediaFromBlobURL( args = {} ) {
 	const latestArgs = useRef( args );
+	const hasUploadStarted = useRef( false );
 	const { getSettings } = useSelect( blockEditorStore );
 
 	useLayoutEffect( () => {
@@ -40,6 +41,10 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 	} );
 
 	useEffect( () => {
+		if ( hasUploadStarted.current ) {
+			return;
+		}
+
 		if (
 			! latestArgs.current.url ||
 			! isBlobURL( latestArgs.current.url )
@@ -55,6 +60,8 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 		const { url, allowedTypes, onChange, onError } = latestArgs.current;
 		const { mediaUpload } = getSettings();
 
+		hasUploadStarted.current = true;
+
 		mediaUpload( {
 			filesList: [ file ],
 			allowedTypes,
@@ -65,10 +72,12 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 
 				revokeBlobURL( url );
 				onChange( media );
+				hasUploadStarted.current = false;
 			},
 			onError: ( message ) => {
 				revokeBlobURL( url );
 				onError( message );
+				hasUploadStarted.current = false;
 			},
 		} );
 	}, [ getSettings ] );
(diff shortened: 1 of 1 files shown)


youknowriad:
Do you think we can build a temporary `useEffectEvent` as a private API or something?

Mamaduka:
@youknowriad, I was just looking into that and now I'm unsure if that will solve our use case. The returned callback from `useEffectEvent` will also be called twice in StrictMode unless React implements some internal tracking system to change the behavior.

youknowriad (approved):
Seems fine.

jsnajdr:
Maybe this bug is a signal that we should be looking for a better solution than the `useUpload` on-mount hook. If I understand it correctly, the hook is used only for the drag&drop flow:
1. Drop a file onto the editor canvas.
2. The `onFilesDrop` handler finds a matching block transform of type `files`.
3. This transform creates a block with `url` attribute that is a blob URL created from the file data.
4. The new block's `Edit` component mounts, and on mount it detects the blob `url` attribute, performs the upload and then updates the `url` attribute to the permanent HTTP URL.

There isn't any other use case for `useUploadMediaFromBlobURL`, is there? The drag&drop flow is a combination of imperative and declarative (effect) code. What if the upload was a part of the transform?
```js
createBlock( 'core/image', {
  url: await uploadFile( file )
} )
```

Also this is another example where the block is not "[headless](https://github.com/WordPress/gutenberg/issues/59592)", i.e., the `Edit` component is doing something that's not editing -- here it's doing part of the job of creating the block, and keeping its structural integrity.

Mamaduka:
@jsnajdr, I agree that would be a better solution. This is a known shortcoming, and there's an open discussion in #49466.

> What if the upload was a part of the transform?

There was an issue or PR for a similar proposal, and if I recall correctly, the issue was making transformations async. I'll try to find it later and link it here for the reference.

PR #62061, opened 2024-05-28
Fix: Some jsdoc return types on edit site selector.

jorgefilipecosta:
The correct JSDoc syntax is {?string} instead of {string?} fixes some errors I noticed on edit site selectors.

## Testing Instructions
Verified there are no static analysis errors.


fabiankaegy (approved):
Looks good to me 👍

PR #62063, opened 2024-05-28
Block Bindings: Create utils for block bindings

SantosGuillamot:
## What?
I'm creating a new `bindings` util file to store all the reusable functions instead of having it in the editor hook. As part of that, I'm abstracting the logic to change the block attributes with the bindings values into a function.

## Why?
It looks cleaner this way and we can reuse the same logic to replace the block attributes in the future.

## How?
Just creating a new file and creating a new `transformBlockAttributesWithBindingsValues` function to replace the block attributes with bindings.

## Testing Instructions
Tests should pass. We are not adding new functionalities.

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -39,7 +39,6 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 		} = props;
 		const bindings = blockAttributes?.metadata?.bindings;
 
-		// It seems if I don't wrap this in a useSelect, the reset in pattern overrides doesn't work as expected.
 		const boundAttributes = useSelect( () => {
 			return getBoundAttributesValues( clientId, context, registry );
 		}, [ clientId, context, registry ] );
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -42,9 +42,7 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 		let newAttributes = blockAttributes;
 		if ( bindings ) {
 			newAttributes = transformBlockAttributesWithBindingsValues(
-				blockAttributes,
 				clientId,
-				name,
 				context,
 				registry
 			);
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -3,7 +3,7 @@
  */
 import { store as blocksStore } from '@wordpress/blocks';
 import { createHigherOrderComponent } from '@wordpress/compose';
-import { useRegistry, useSelect } from '@wordpress/data';
+import { useRegistry } from '@wordpress/data';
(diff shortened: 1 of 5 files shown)


packages/block-editor/src/hooks/use-bindings-attributes.js
```diff
@@ -22,90 +27,25 @@ import { unlock } from '../lock-unlock';
...
-					}
-				}
-			}
-
-			return attributes;
-		}, [ bindings, name, clientId, context, registry, sources ] );
-
-		const { setAttributes } = props;
+		const boundAttributes = useSelect( () => {
+			return getBoundAttributesValues( clientId, context, registry );
```
ellatrix, line 43: Why is this a reusable function if it's only used once? I also don't really like this, it becomes unclear what store this is selecting from.

packages/block-editor/src/utils/bindings.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param {string} clientId     - The block clientId.
+ * @param {Object} blockContext - The block context, which is needed for the binding sources.
+ * @param {Object} registry     - The data registry.
+ *
+ * @return {Object} Object with the value obtained from the bindings of each bound attribute.
+ */
+export function getBoundAttributesValues( clientId, blockContext, registry ) {
+	const attributes = registry
+		.select( blockEditorStore )
+		.getBlockAttributes( clientId );
```
ellatrix, line 60: You sneaked in a block editor store subscription here for every block, which will impact performance. Why not use the attributes prop?

ellatrix:
The PR title/description makes its sound like this is just moving code, but it's not?

SantosGuillamot:
> The PR title/description makes its sound like this is just moving code, but it's not?

It was the intention, but I wasn't fully aware of the implications 😄 

Anyway, this pull request was meant to create a reusable function to transform the attributes not only in the Edit hook but in other places if wanted. I thought it was needed for [the pull request to allow editing of post meta in bindings](https://github.com/WordPress/gutenberg/pull/61753), but it isn't with the new approach there.

I am marking it as draft again because I believe it could still be valuable something like that but there are some uncertainties we would need to explore deeper. Happy to close and reopen if you prefer.

packages/block-editor/src/utils/bindings.js
SantosGuillamot: I tried to avoid passing `attributes` as a parameter, but I think it makes sense then.

packages/block-editor/src/hooks/use-bindings-attributes.js
SantosGuillamot: It was meant to be used in other places like [allow editing of post meta](https://github.com/WordPress/gutenberg/pull/61753), but we changed the approach there so it is not reused anymore. See [related comment](https://github.com/WordPress/gutenberg/pull/62063#issuecomment-2138907668).

gziolo:
Is it still actionable? It looks like an alternative approach are being explored, example:
- https://github.com/WordPress/gutenberg/pull/64102

SantosGuillamot:
I will close this and reopen it if consider necessary.

PR #61696, opened 2024-05-15
Fix: Update styles for checkbox and radio controls.

vipul0425:
## What?
Fixes issue: #61652 

## Why?
The current spacing between the checkbox input and its label is inconsistent with that of the radio buttons, even though they have the same size.

## How?
This PR updates the stylings for the checkbox control and radio control to make them consistent.

## Testing Instructions
1. Open any page or post in the editor and open its settings panel.
2. Click on the published or discussions buttons and another tab should appear where one can test the checkbox and radio controls simultaneously.

### Testing Instructions for Keyboard
nil

## Screenshots or screencast 
image

PR #62066, opened 2024-05-28
Editor: Align the Post Format control design with the rest of the post sidebar controls

youknowriad:
## What?

Looking at the design of the Post Format control in the post editor sidebar, it looks out of place compared to the other controls. This PR tries to bring some harmony there and just applies the same behavior we have for Post Author for instance.

Before:

Screenshot 2024-05-28 at 2 19 47 PM

After:

Screenshot 2024-05-28 at 2 19 20 PM

## Testing Instructions

 - Open the post editor sidebar in a theme with post formats (2012 for instance).
 - Check the post format control in the sidebar.

### DIFF
--- a/packages/editor/src/components/post-format/panel.js
+++ b/packages/editor/src/components/post-format/panel.js
@@ -1,20 +1,85 @@
 /**
  * WordPress dependencies
  */
-import { PanelRow } from '@wordpress/components';
+import { Button, Dropdown } from '@wordpress/components';
+import { __, sprintf } from '@wordpress/i18n';
+import { useSelect } from '@wordpress/data';
+import { useState, useMemo } from '@wordpress/element';
+import { __experimentalInspectorPopoverHeader as InspectorPopoverHeader } from '@wordpress/block-editor';
 
 /**
  * Internal dependencies
  */
-import PostFormatForm from './';
+import { default as PostFormatForm, POST_FORMATS } from './';
 import PostFormatCheck from './check';
+import PostPanelRow from '../post-panel-row';
+import { store as editorStore } from '../../store';
 
-export function PostFormat() {
+/**
+ * Renders the Post Author Panel component.
+ *
+ * @return {Component} The component to be rendered.
+ */
+function PostFormat() {
+	const { postFormat } = useSelect( ( select ) => {
+		const { getEditedPostAttribute } = select( editorStore );
+		const _postFormat = getEditedPostAttribute( 'format' );
+		return {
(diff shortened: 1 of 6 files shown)


jameskoster:
To avoid introducing too many extra clicks I think ideally we'd move to something more like the Status dialog, IE radios rather than a select. I appreciate that might be a detail to work on separately though.

If we keep the select for now, would it make sense to auto-focus the select when the dialog is opened? Also could it be 40px rather than 32px, to match the other Inspector inputs?

youknowriad:
I updated the dropdown to use RadioControl like the post status one.

packages/editor/src/components/post-format/style.scss
```diff
@@ -2,8 +2,14 @@
...
-	align-items: stretch;
+.editor-post-format__dialog .editor-post-format__dialog-content {
+	// sidebar width - popover padding - form margin
+	min-width: $sidebar-width - $grid-unit-20 - $grid-unit-20;
+	margin: $grid-unit-10;
+}
+
+.editor-post-format__options {
+	.components-base-control__field > .components-v-stack {
+		gap: $grid-unit-15;
```
youknowriad, line 13: The gap between radio options is way too small by default in RadioControl, I wonder if this style should be something to elevate to the components package. cc @mirka

packages/editor/src/components/post-format/panel.js
```diff
@@ -1,20 +1,85 @@
...
+							variant="tertiary"
+							aria-expanded={ isOpen }
+							aria-label={ sprintf(
+								// translators: %s: Current post format.
+								__( 'Change format: %s' ),
+								activeFormat?.caption
+							) }
+							onClick={ onToggle }
+						>
+							{ activeFormat?.caption }
```
ntsekouras, line 69: Can this be in any case not defined? I think not, but if yes, it will just render an 'empty' button here which would make no much sense, would it?

ntsekouras (approved):
Code wise this looks good. Thanks!

packages/editor/src/components/post-format/panel.js
youknowriad: In theory no, if you change in the UI but this is just defensive coding I guess to avoid breaking the editor if some plugin does something weird in editEntityRecord.

packages/editor/src/components/post-format/style.scss
mirka: In progress at #61696

Issue #61488, opened 2024-05-08
Scheduling: Improve flow of new Status dropdown


jasmussen:
Work is underway to unify the publishing flows between post and site editors. As part of that, there's a "Status" dropdown, which lets you choose between draft, pending, private, scheduled, and published:

![statuses](https://github.com/WordPress/gutenberg/assets/1204802/c16d3356-426c-42c9-a877-cfd0cd1a6838)

If you change the publish date to a time in the future, the "Scheduled" status will automatically be picked for you. 

But if you change the status to "Schedule" first, before picking a time in the future, the publish time will automatically be set for you, to a point in the future, as shown in this GIF:

![scheduling](https://github.com/WordPress/gutenberg/assets/1204802/17c5499c-6265-4474-be18-e8571f7b3f35)

For the near term, let's consider the following solution: we remove "Scheduled" from the Status chip, and rename "Publish" to "Public". So to schedule a post, you'd have to set these properties:

* Status: Public.
* Publish: [future date]

For a draft, you'd have:

* Status: Draft.
* Publish: Immediately.

For an already published post, you'd have:

* Status: Public.
* Publish: [past date]

This flow is similar to what exists in 6.5, but comes with the structural benefits of unifying in a status panel.

Longer term updates could work like the following.

### Editing a draft

<img width="1070" alt="Screenshot 2024-05-13 at 15 27 04" src="https://github.com/WordPress/gutenberg/assets/846565/090999ff-019f-46c8-8bfe-322af43f1d7f">

### Editing a published record

<img width="984" alt="Screenshot 2024-05-13 at 15 30 11" src="https://github.com/WordPress/gutenberg/assets/846565/2ed07dac-0aab-4c72-9c3f-5d51023c7ef2">

* In both cases the Status chip reveals a menu containing actions equating to status changes. 
* The pre-publish UI moves to a modal which enables content to be published from data views. 
* Publish date is specified when you elect to schedule the post. 
* The status change is saved immediately, eliminating ambiguous interaction between status + publish date fields.

annezazu:
I was testing this today with GB 18.2 and was so baffled by this experience, partially because I couldn't always get it to consistently schedule out in advance:

https://github.com/WordPress/gutenberg/assets/26996883/a2715b29-d326-480c-9836-f74937fd5765

When it did schedule out a week later, I hardly noticed it as it's not a hugely visible change and it's disconnected from the options you're in the process of selecting. 

[...]

PR #62070, opened 2024-05-28
Editor: render publish date control when the status is `future`(scheduled)

ntsekouras:
## What?
Related: https://github.com/WordPress/gutenberg/issues/61488#issuecomment-2117560411

There are some suggestion on how to improve the scheduling flow in the editor with the new reused `PostStatus` component.

This PR for now simply does:
1. No setting automatically a future day (a week from current time) when we select the `scheduled` status from the available statuses
2. When a post is scheduled or when we select the option from the PostStatus component, the publish date component is rendered for users to select the date they want.

### Notes
1. This has some nuances similar to the problem it solves, but on the other way around. For example if a user sets the status to `scheduled` and not update the date, the final status will not be `future`. I think that's fine since the rendering of the control will be very prominent to the user to also select a date, but just mentioning.
2. The design suggested in the issue has a different date picker component and I'm reusing the one we use for the publish date. Do we need some designs updates or that's fine? 

## Testing Instructions
1. In a post try combinations with the post status to `scheduled` and the publish date and see how it feels and share feedback

### DIFF
--- a/packages/block-editor/src/components/publish-date-time-picker/index.js
+++ b/packages/block-editor/src/components/publish-date-time-picker/index.js
@@ -11,20 +11,24 @@ import { getSettings } from '@wordpress/date';
  */
 import InspectorPopoverHeader from '../inspector-popover-header';
 
-function PublishDateTimePicker(
-	{ onClose, onChange, ...additionalProps },
+export function PublishDateTimePicker(
+	{ onClose, onChange, showPopoverHeaderActions, ...additionalProps },
 	ref
 ) {
 	return (
 		<div ref={ ref } className="block-editor-publish-date-time-picker">
 			<InspectorPopoverHeader
 				title={ __( 'Publish' ) }
-				actions={ [
-					{
-						label: __( 'Now' ),
-						onClick: () => onChange?.( null ),
-					},
-				] }
+				actions={
+					showPopoverHeaderActions
+						? [
+								{
+									label: __( 'Now' ),
+									onClick: () => onChange?.( null ),
+								},
+						  ]
+						: undefined
+				}
 				onClose={ onClose }
 			/>
 			<DateTimePicker
@@ -36,4 +40,16 @@ function PublishDateTimePicker(
 	);
 }
 
-export default forwardRef( PublishDateTimePicker );
+export const PrivatePublishDateTimePicker = forwardRef( PublishDateTimePicker );
+
+function PublicPublishDateTimePicker( props, ref ) {
+	return (
+		<PrivatePublishDateTimePicker
+			{ ...props }
+			showPopoverHeaderActions
+			ref={ ref }
+		/>
+	);
+}
+
+export default forwardRef( PublicPublishDateTimePicker );
--- a/packages/block-editor/src/components/publish-date-time-picker/index.js
+++ b/packages/block-editor/src/components/publish-date-time-picker/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { DateTimePicker } from '@wordpress/components';
+import { DateTimePicker, TimePicker } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { forwardRef } from '@wordpress/element';
 import { getSettings } from '@wordpress/date';
@@ -12,9 +12,16 @@ import { getSettings } from '@wordpress/date';
 import InspectorPopoverHeader from '../inspector-popover-header';
 
 export function PublishDateTimePicker(
-	{ onClose, onChange, showPopoverHeaderActions, ...additionalProps },
+	{
+		onClose,
+		onChange,
+		showPopoverHeaderActions,
+		isCompact,
+		...additionalProps
+	},
 	ref
 ) {
+	const DatePickerComponent = isCompact ? TimePicker : DateTimePicker;
 	return (
 		<div ref={ ref } className="block-editor-publish-date-time-picker">
 			<InspectorPopoverHeader
@@ -31,7 +38,7 @@ export function PublishDateTimePicker(
 				}
 				onClose={ onClose }
 			/>
-			<DateTimePicker
+			<DatePickerComponent
 				startOfWeek={ getSettings().l10n.startOfWeek }
 				onChange={ onChange }
 				{ ...additionalProps }
@@ -47,6 +54,7 @@ function PublicPublishDateTimePicker( props, ref ) {
 		<PrivatePublishDateTimePicker
 			{ ...props }
 			showPopoverHeaderActions
+			isCompact={ false }
 			ref={ ref }
 		/>
 	);
--- a/packages/block-editor/src/private-apis.js
+++ b/packages/block-editor/src/private-apis.js
@@ -41,6 +41,7 @@ import { requiresWrapperOnCopy } from './components/writing-flow/utils';
 import { PrivateRichText } from './components/rich-text/';
 import { PrivateBlockPopover } from './components/block-popover';
 import { PrivateInserterLibrary } from './components/inserter/library';
+import { PrivatePublishDateTimePicker } from './components/publish-date-time-picker';
 
 /**
  * Private @wordpress/block-editor APIs.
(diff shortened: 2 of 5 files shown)


jameskoster:
> if a user sets the status to scheduled and not update the date, the final status will not be future. I think that's fine since the rendering of the control will be very prominent to the user to also select a date, but just mentioning.

I agree it's okay. Ideally in this case the status would revert to whatever it is currently set to on close. Would that be possible?

> The design suggested in the issue has a different date picker component and I'm reusing the one we use for the publish date. Do we need some designs updates or that's fine?

The date picker definitely needs some design love, but that's one to handle separately. For this PR the only changes I'd suggest:

* Add a horizontal rule (or some additional spacing) above the date picker to separate it from the status radios.
* Hide the 'Now' shortcut – it doesn't make much sense when scheduling a post for the future. 

On a similar note to that last point, would it be possible to disable all dates before `today` when status = scheduled?

ntsekouras:
I addressed these:

>Add a horizontal rule (or some additional spacing) above the date picker to separate it from the status radios.
Hide the 'Now' shortcut – it doesn't make much sense when scheduling a post for the future.

by making private components. Due to the nested components it's a bit more complicated to make specific changes like don't allow the previous dates, etc.. and I'd prefer if we look into them separately.

>I agree it's okay. Ideally in this case the status would revert to whatever it is currently set to on close. Would that be possible?

That would add some complexity too by needing some new side effects and I feel it would probably be better to explore separately. Are the above two a block for this one?

jameskoster:
Not blockers 👍 

Generally I think this is in a decent spot given the constraints.

My only remaining question is whether we could use `TimePicker` instead of `DateTimePicker`. The calendar takes up an enormous amount of space in the dialog. What do you think @jasmussen ?

| TimePicker | DateTimePicker |
| --- | --- |
| Screenshot 2024-05-29 at 14 20 00 | Screenshot 2024-05-29 at 14 21 34 |

jasmussen:
That's a good question. I'd personally be okay with omitting the fuller datepicker; can it be expand/collapsed easily? I'd love @annezazu's instincts here too. 

The main beat here is that the calendar allows you to more easily pick something later in the month, knowing weekends, etc, vs. typing in the number manually.

jameskoster:
> can it be expand/collapsed easily?

Nik will know better but I don't think so. The full date picker with calendar is always available from the dedicated 'Publish' row though:

Screenshot 2024-05-29 at 15 38 54

ntsekouras:
>can it be expand/collapsed easily

Not very easily and I feel like we should dedicate time to that component separately to be more flexible. In this PR just for hiding the `now`, I had to make two components private 😄 . If this is what's wanted of course we could do it, but it will probably take a bit more time to discuss/design I think and the 6.6 deadline is in a couple of days.

jameskoster:
@ntsekouras is using `TimePicker` an option? Unless there's any objection I think that's the final consideration before merging here.

ntsekouras:
>is using TimePicker an option? Unless there's any objection I think that's the final consideration before merging here.

@jameskoster you mean also hide it?

annezazu:
Noting that I ran into a strange flow shown at the beginning of this where after selecting the schedule option, it's almost as if it kicked me over into the "Publish" module where I couldn't go back and change the status. I had a hard time replicating as you'll see in the rest of the video. Other quick notes are around the month only being bolded and not the year. Not sure if that's part of the component but it stood out to me. 

https://github.com/WordPress/gutenberg/assets/26996883/75146d39-5b17-4983-a0e8-a5b523b484c2

In terms of rendering the entire calendar view, I think we should for consistency. It's important to know the day of the week a post is scheduled, especially for large news rooms that have a cadence of themed posts per day of the week.

packages/block-editor/src/components/publish-date-time-picker/index.js
```diff
@@ -11,20 +11,24 @@ import { getSettings } from '@wordpress/date';
...
-				] }
+				actions={
+					showPopoverHeaderActions
+						? [
+								{
+									label: __( 'Now' ),
+									onClick: () => onChange?.( null ),
+								},
+						  ]
+						: undefined
```
Mamaduka, line 37: If we decide to stabilize it, we can allow consumers to pass the `actions` prop to the component.

Mamaduka (approved):
The changes look good and test well for me.

I wasn't able to reproduce @annezazu's issue, but my device has scroll bars enabled by default.

Unrelated note: I've noticed that a post might get accidentally Published/Scheduled if you're playing with the settings and the autosave functionality kicks in. I'm not sure if this is a new or old issue, but I would expect not to publish my posts accidentally.

## Steps to reproduce
1. Open a draft post.
2. Change status to Published or Scheduled.
3. Trigger an autosave - `window.wp.data.dispatch( 'core/editor' ).autosave()`.
4. Reload the page, and notice you'll get unsaved changes warning.
5. Notice that the post isn't a draft anymore.

jasmussen:
I'd support Anne's instincts here. 

This is a sensitive part of the flow, we wouldn't want people accidentally publishing when they meant to schedule. I'm here for this particular flow, it seems to capture the best of both worlds. Question, however: in case we find out in the beta period that this is, after all, disruptive to some of our users, would it be possible in the beta period to change the behavior to the plan B outlined in https://github.com/WordPress/gutenberg/issues/61488?

ntsekouras:
>I wasn't able to reproduce @annezazu's issue

Same here..

>in case we find out in the beta period that this is, after all, disruptive to some of our users, would it be possible in the beta period to change the behavior to the plan B outlined in https://github.com/WordPress/gutenberg/issues/61488?

Is plan B:
Screenshot 2024-05-13 at 15 27 04

and the image below from the issue?

If yes, probably, but it's quite different in terms of UI and when changing the publishing flow, it requires much testing. Personally it would feel to me though that we back tracking to the old flow though..

packages/block-editor/src/components/publish-date-time-picker/index.js
ntsekouras: Of course..

jasmussen:
Plan B is this, and I hope we don't need to go there, I'm just a cautious person because this is a hugely prominent flow:

We remove "Scheduled" from the Status chip, and rename "Publish" to "Public". So to schedule a post, you'd have to set these properties:

* Status: Public.
* Publish: [future date]

For a draft, you'd have:

* Status: Draft.
* Publish: Immediately.

For an already published post, you'd have:

* Status: Public.
* Publish: [past date]

This flow is similar to what exists in 6.5, but comes with the structural benefits of unifying in a status panel.

jameskoster (approved):
The size of the datepicker really bugs me, the dialog feels very large on small screens. 

Still, this removes the ambiguity around scheduling, so I think we should merge and keep 'plan B' up the sleeve in case any major issues arise during testing.

youknowriad:
I agree the size of the date picker feels wrong. I have to scroll the dropdown to actually make sense of it.

youknowriad:
I agree as well that it should be possible to switch to `TimePicker` instead of `DateTimePicker` to solve this. A private prop like "compact" or something could solve the issue here.

packages/block-editor/src/components/publish-date-time-picker/index.js
```diff
@@ -11,20 +11,24 @@ import { getSettings } from '@wordpress/date';
...
+				}
 				onClose={ onClose }
 			/>
 			<DateTimePicker
```
youknowriad: We could use a `compact` prop to switch this one to `TimePicker`

ntsekouras:
I've updated to only use `TimePicker` inside the post status panel. Sorry, but I wasn't aware of that component and its name is misleading. I guess this should be good to land now and will also investigate @Mamaduka comment about autosaving.

annezazu:
I'm going to try to track down some feedback from enterprise-y/news folks for now to get some initial thoughts too. Scheduling content, from what I've seen, is a common action there to help with everything from weekend coverage (scheduling posts to ensure content goes live over the weekend) to going live with a big piece that needs more coordination.

Issue #61106, opened 2024-04-25
Convert to component: Badge


jameskoster:
A recently-merged data views feature allows consumers to display field values as a 'badge' in grid layout. This is utilised in the Patterns page to communicate sync status:

<img width="540" alt="Screenshot 2024-04-25 at 15 12 13" src="https://github.com/WordPress/gutenberg/assets/846565/8eeeb81a-0827-4d67-b2a8-bd9fbc20fbac">

This element will likely have applications beyond data views, so it may be worth extracting and converting to a reusable component.

jasmussen:
Indeed, that seems useful. However it's probably important to do that effort only if and when the need for this chip emerges organically.

jameskoster:
In addition to badge fields in data views, another potential use case is https://github.com/WordPress/gutenberg/issues/57763.

But yes, no need to rush for only one or two applications.

jameskoster:
Another potential use case for the badge: https://github.com/WordPress/gutenberg/issues/38277#issuecomment-2090312644

Vrishabhsk:
Hi @jameskoster 👋 
- I have raised a PR for adding the `Badge` component : https://github.com/WordPress/gutenberg/pull/66555

Let me know your thoughts on this. Thanks

orcungogus:
@jameskoster could we please add a version of this badge component where we have the option to include an icon (before OR after the text)?

rogermattic:
@jameskoster @Vrishabhsk I worked on the component design a bit, and here's my proposal.

To cover a range of usecases we need to have a few different colour `badges`.  So I started exploring the colour mixes using the [WordPress palette](https://github.com/WordPress/gutenberg/blob/26123cd51f16a77c961a38052fb13535ceecf8b7/packages/base-styles/_colors.scss#L4). Came up with fairly simple rules (that hopefully can be algorithmically generated):

**Base colours**
If we take these as our base colours:
![Image](https://github.com/user-attachments/assets/42bea5dc-f01e-46b3-9274-17c9e792428b)

**Background**
We can take the base colours and apply 10% transparency on solid white. 
![Image](https://github.com/user-attachments/assets/87cc3165-a22c-4bfc-a8e3-42a1385364f1)

**Foreground**
We can apply the $gray-700 with a multiply blending mode onto the base colour.
![Image](https://github.com/user-attachments/assets/6fb9f44f-2c5d-4230-a9d5-7781887c38aa)

[...]

PR #62071, opened 2024-05-28
Inspector: Display home / posts page badge

jameskoster:
## What?
Display a badge in the Inspector when editing the front page or the posts page.

## Why?
In data views this badge appears next to the title to help you identify these pages. However there's nothing in the Editor to communicate this detail.

## Testing Instructions
1. In reading settings configure a static homepage and a posts page
2. Open those pages in the Editor
3. Notice the 'Front page' and 'Posts page' chip in the Inspector

| Data views | Inspector | 
| --- | --- |
| Screenshot 2024-05-28 at 13 43 43 | Screenshot 2024-05-28 at 15 25 52 Screenshot 2024-05-28 at 15 29 39  |

As a follow-up it might be nice to update the icon too, but as that will involve touching the command palette/document title I'd prefer to explore it separately.

### DIFF
--- a/packages/editor/src/components/post-card-panel/index.js
+++ b/packages/editor/src/components/post-card-panel/index.js
@@ -28,42 +28,52 @@ import {
 import { unlock } from '../../lock-unlock';
 
 export default function PostCardPanel( { actions } ) {
-	const { title, icon, isSync } = useSelect( ( select ) => {
-		const {
-			getEditedPostAttribute,
-			getCurrentPostType,
-			getCurrentPostId,
-			__experimentalGetTemplateInfo,
-		} = select( editorStore );
-		const { getEditedEntityRecord } = select( coreStore );
-		const _type = getCurrentPostType();
-		const _id = getCurrentPostId();
-		const _record = getEditedEntityRecord( 'postType', _type, _id );
-		const _templateInfo =
-			[ TEMPLATE_POST_TYPE, TEMPLATE_PART_POST_TYPE ].includes( _type ) &&
-			__experimentalGetTemplateInfo( _record );
-		let _isSync = false;
-		if ( GLOBAL_POST_TYPES.includes( _type ) ) {
-			if ( PATTERN_POST_TYPE === _type ) {
-				// When the post is first created, the top level wp_pattern_sync_status is not set so get meta value instead.
-				const currentSyncStatus =
-					getEditedPostAttribute( 'meta' )?.wp_pattern_sync_status ===
-					'unsynced'
-						? 'unsynced'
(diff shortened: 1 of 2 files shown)


jasmussen (approved):
Seems okay, would appreciate a code check before you land.

packages/editor/src/components/post-card-panel/index.js
```diff
@@ -83,8 +93,19 @@ export default function PostCardPanel( { actions } ) {
 					className="editor-post-card-panel__title"
 					weight={ 500 }
 					as="h2"
+					lineHeight={ '20px' }
 				>
 					{ title ? decodeEntities( title ) : __( 'No Title' ) }
+					{ isFrontPage && (
+						<span className="editor-post-card-panel__title-badge">
```
youknowriad, line 100: Is this badge something that can become a generic UI components to move to the components package? We're using it in multiple places now.

packages/editor/src/components/post-card-panel/index.js
```diff
@@ -28,42 +28,52 @@ import {
...
-	}, [] );
+			return {
+				title:
+					_templateInfo?.title || getEditedPostAttribute( 'title' ),
+				icon: unlock( select( editorStore ) ).getPostIcon( _type, {
+					area: _record?.area,
+				} ),
+				isSync: _isSync,
+				isFrontPage: siteSettings?.page_on_front === _id,
+				isPostsPage: siteSettings?.page_for_posts === _id,
```
youknowriad, line 72: In other places the "id" here was typecast into an integer to work. I don't know whether it's needed here. cc @ntsekouras

youknowriad (approved):

packages/editor/src/components/post-card-panel/index.js
richtabor: Yes, probably so. Also in use for anchor tags in list view and within data views. 
 
CleanShot 2024-05-29 at 09 46 40
CleanShot 2024-05-29 at 09 47 05

packages/editor/src/components/post-card-panel/index.js
ntsekouras: There was definitely an issue with some endpoint or selector in the past that would return a `string` id, but  couldn't pinpoint it now.. By debugging and looking at the code, at least in these checks, both values are numbers. So it's safe to leave as is.

packages/editor/src/components/post-card-panel/index.js
jameskoster: Yes please https://github.com/WordPress/gutenberg/issues/61106

Issue #61196, opened 2024-04-29
Restore the terms 'Permalink' and 'Slug' in the PostURL user interface


afercia:
### Description

Splitting this out from https://github.com/WordPress/gutenberg/issues/56381#issuecomment-2061235862

In https://github.com/WordPress/gutenberg/pull/60632 some long-established temrinology regartding the 'post permalink' has been changed. I'm not sure there has been a broader discussion to validate whether these changes are any good for users. I'd liek to start a broader discussion about that.

Current state on trunk, where the new UI uses the term 'Link':

![Screenshot 2024-04-29 at 14 43 43](https://github.com/WordPress/gutenberg/assets/1682452/c3d87bfc-9fb6-4a46-9ea8-170e891b1bc7)

While I'd agree that, if I had to explain tp a complete newbie or to non-tech-savvy users what to edit here, I would use the most simple phrasing e.g. 'Edit the post link', I'd also note that, _for ages_, WordPress has been using its own specific terminology:

## Permalink
The term 'permalink' has been for ages one of the most prominent terminology used in WordPress. For years, it appeared right below the post title, in a very prominent place:

![Screenshot 2024-04-29 at 14 21 32](https://github.com/WordPress/gutenberg/assets/1682452/34f49dce-79c5-44cc-80fb-3c57791b2c01)

Even though it's a somewhat 'technical' term, WordPress users are very familiar with what a 'Permalink' is.

The admin menu uses the same term under Settings > Permalinks at wp-admin/options-permalink.php

## Slug 
The term 'Slug' or 'URL Slug' has been used in WordPress for ages as well. I can count around 30 occurrences of the term 'slug' used within transtlatable strings.

Sometimes this term is used for objects other than a post, such as plugins, widgets, patterns, font collections, etc. Regardless, it's a widely used term and WordPress users are familiar with it.

## Concerns 
https://github.com/WordPress/gutenberg/pull/60632 changed the terms used in the 'edit permalink' user interface after very little discussion. I'm not sure such changes should be done that lightly. I'm not aprioristically opposed to changes but I definitely think _changes that impact long established conventions should be validated by user testing_. In absence of user testing that validates the proposed change performs better, we should just avoid to introduce such changes.

In this specific case, personally I don't see a good reason for this change.

[...]

PR #62073, opened 2024-05-28
Inspector: Add '/' prefix to Link button

jameskoster:
## What?
Adds a `/` prefix to the Link button label in the Inspector.

## Why?
The slash prefix helps make the connection with the URL. Otherwise the uncapitalised, hyphenated text string could look like a mistake. Also aligns with the input you find in the dialog which also includes the `/` prefix.

| Before | After |
| --- | --- |
| Screenshot 2024-05-28 at 15 38 01 | Screenshot 2024-05-28 at 15 37 40 |

### DIFF
--- a/packages/editor/src/components/post-url/panel.js
+++ b/packages/editor/src/components/post-url/panel.js
@@ -73,7 +73,7 @@ function PostURLToggle( { isOpen, onClick } ) {
 			aria-label={ sprintf( __( 'Change link: %s' ), decodedSlug ) }
 			onClick={ onClick }
 		>
-			{ decodedSlug }
+			/{ decodedSlug }
 		</Button>
 	);
 }


ntsekouras (approved):
Looks good, thanks!

jasmussen:
We can try this.

t-hamano:
Not a blocker, but I've noticed that the Link `slug` always appears, even when the permalink is something other than Post Name. If we add a prefix slash, some users might mistakenly think it's the actual URL.

This might be something to fix in a separate PR.

### Gutenberg

image

### WP6.5

image

richtabor:
> Not a blocker, but I've noticed that the Link slug always appears, even when the permalink is something other than Post > Name. If we add a prefix slash, some users might mistakenly think it's the actual URL.

> This might be something to fix in a separate PR.

Yes, let's follow-up to make sure they're the same.

richtabor (approved):
LGTM

t-hamano:
> Not a blocker, but I've noticed that the Link slug always appears, even when the permalink is something other than Post Name. If we add a prefix slash, some users might mistakenly think it's the actual URL.
> 
> This might be something to fix in a separate PR.

I have discovered that this issue may be covered under #61196.

paaljoachim:
In general I would think that seeing the / (forward slash) in front of the link can be somewhat confusing for folks.
As what is seen as the link to me is associated with the title of the page/post and does not have much of an association with the URL structure of the page/post.

That is just an outside view looking in....

Issue #59659, opened 2024-03-07
Advancing site editor index views


SaxonF:
Related https://github.com/WordPress/gutenberg/issues/55083 https://github.com/WordPress/gutenberg/issues/53322

## What problem does this address?
As part of the [admin track of work](https://github.com/WordPress/gutenberg/issues/53322) we want to start adopting some of the new UX paradigms within the site editor index views. 

https://github.com/WordPress/gutenberg/assets/1072756/8fe4bd09-694c-44d7-b9a9-c91d6c87d56c

Pages and templates within the site editor currently drill-down the navigation sidebar showing a list of records, but what we want to work towards is always using the "content frame" for lists, including the "side by side" layout type which lets you quickly preview posts before editing. This also solves the weird "manage pages" and "manage templates" links that currently exist. The goal of this work is bring those "manage" pages forward so they are visible when visiting /pages or /templates, much like /patterns.

### Site editor responsiveness

- [ ] https://github.com/WordPress/gutenberg/issues/60847

### Replace current pages with manage pages
Replace the existing pages view that shows pages in sidebar and update with the "manage pages" view. For this piece of work we can keep the "custom views" as experimental.

Old
<img width="1512" alt="image" src="https://github.com/WordPress/gutenberg/assets/1072756/e4185722-2c11-406e-ba48-6590ae72a32a">

New
<img width="1512" alt="image" src="https://github.com/WordPress/gutenberg/assets/1072756/2a087496-e970-4423-a3d1-9d19c074a176">

Criteria:

- [x] Clicking "pages" in top level navigation opens pages data views with "side by side" layout type enabled by default. https://github.com/WordPress/gutenberg/pull/59950
- [x] Default views for pages include [All, Published, Draft, Scheduled, Utility, Trash] https://github.com/WordPress/gutenberg/pull/62021
- [ ] Pages: "Trash" view should default to `table` layout
- [ ] https://github.com/WordPress/gutenberg/issues/60772
- [ ] https://github.com/WordPress/gutenberg/issues/41716
- [x] When in side by side layout, clicking an item previews in the frame on right.
- [x] Clicking the frame expands into edit mode
- [x] Each item in side by side has an "edit" icon that expands frame into edit mode
- [x] Each item in side by side has an ellipsis icon that triggers page actions [Delete, rename, edit]
- [x] When in table view, clicking a page title should open editor
- [x] Mobile: show the content frame. https://github.com/WordPress/gutenberg/pull/60409
[...]

PR #61577, opened 2024-05-10
Add plugin template registration API

Aljullu:
## What?

This PR introduces a new API so plugins can easily register templates.

Closes https://github.com/WordPress/gutenberg/issues/41401.

Backport PR in core: https://github.com/WordPress/wordpress-develop/pull/7125.

## Why?

Currently, it's not easy for plugins to register templates, as they need to hook into several filters.

## How?

This PR creates a new function: `wp_register_block_template()` as well as a `WP_Block_Templates_Registry` class. They are based on `register_block_type()` and `WP_Block_Type_Registry`.

You can read more about the API shape in this comment: https://github.com/WordPress/gutenberg/issues/41401#issuecomment-2085542661.

## Testing Instructions

**Basic template customization:**

1. Add this code to an existing plugin or in a code snippet using the [Code Snippets plugin](https://wordpress.org/plugins/code-snippets/):

```PHP
add_action(
	'init',
	function () {
		wp_register_block_template(
			'gutenberg//plugin-template',
			array(
				'title'       => 'Plugin Template',
				'description' => 'A template registered by a plugin.',
				'content'     => '<!-- wp:template-part {"slug":"header","tagName":"header"} /--><!-- wp:group {"tagName":"main","layout":{"inherit":true}} --><main class="wp-block-group"><!-- wp:paragraph --><p>This is a plugin-registered template.</p><!-- /wp:paragraph --></main><!-- /wp:group -->',
			)
		);
		add_action(
			'category_template_hierarchy',
			function () {
				return array( 'plugin-template' );
			}
		);
	}
);
```
2. Go to a post category in the frontend (ie: `/category/test-category/`) and verify the _Plugin Template_ contents are rendered.
3. Go to Appearance > Editor > Templates and verify the `Plugin Template` appears with the correct title and description.
4. Make some edits to the template. Verify they are applied in the frontend.
5. Revert the edits and verify edits are reverted in the frontend as well.

**With theme customization:**

6. Add a `/templates/plugin-template.html` template to your active theme and repeat steps 2-5.
7. Remove `/templates/plugin-template.html` from your theme before proceeding with the next steps.

**Using the template in a post type:**

8. Update the code snippet from step 1 to this:
```PHP
add_action(
	'init',
	function () {
		wp_register_block_template(
			'gutenberg//plugin-template',
			array(
				'title'       => 'Plugin Template',
				'description' => 'A template registered by a plugin.',
[...]

PR #36911, opened 2021-11-26
Templates Controller: Add missing 'is_custom' prop

Mamaduka:
## Description
The prop was originally introduced in #35802. It returns a boolean value whether a template is a custom template and isn't part of WP default templates.

## How has this been tested?
Unit tests should pass.

## Types of changes
Bugfix

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #62075, opened 2024-05-28
Templates: add `postTypes` and `isCustom` fields & filters

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/59659

## What?

Adds two new fields to the templates (`postTypes` & `isCustom`).

| Post types (`postTypes` field) | Type (`isCustom` field) |
| --- | --- |
| Captura de ecrã 2024-05-28, às 17 14 22 | Captura de ecrã 2024-05-28, às 17 14 05 |

## Why?

This is a preliminary step to use this to provide a new sidebar filters that include "Posts", "Pages", and "Custom", as per https://github.com/WordPress/gutenberg/issues/59659

## How?

- c821e44c89e01b1c45350f577c55fe49d554dbd5 Adds `isCustom` field & filter in the client. Displays it as a badge in the grid layout, and only renders the `Custom` value.
- 4cea71191e740259ae88c6789447d116f52320ab Adds `postType` field & filter in the client.
- f7b953e418dc127fc024b7cdda028e1669b6987f Declares `post_types` in the templates REST endpoint.
- 16518e70095d6517a42f17f1326399d8ed86ca68 Backport changelog. 

## Testing Instructions

Visit Site Editor > Templates and verify there's two new fields (hidden by default) and filters.

## TODO

- [ ] Verify `postType` logic. The current logic only accounts for templates the theme intended to register for a given post type. However, when switching templates (Post editor > swap templates), the user-created templates are also listed.

### DIFF
--- a/lib/compat/wordpress-6.6/rest-api.php
+++ b/lib/compat/wordpress-6.6/rest-api.php
@@ -87,3 +87,43 @@ function gutenberg_register_global_styles_revisions_endpoints() {
 }
 
 add_action( 'rest_api_init', 'gutenberg_register_global_styles_revisions_endpoints' );
+
+/**
+ * Registers a new post_type field for the Templates REST API route.
+ */
+function gutenberg_register_template_post_type_field() {
+	register_rest_field(
+		'wp_template',
+		'post_types',
+		array(
+			'get_callback' => 'gutenberg_rest_template_post_type_callback',
+			'schema'       => array(
+				'description' => __( 'The post types the template is intended for.', 'gutenberg' ),
+				'type'        => 'array',
+				'readonly'    => true,
+			),
+		)
+	);
+}
+add_action( 'rest_api_init', 'gutenberg_register_template_post_type_field' );
+
+/**
+ * Callback for the post_type field in the Templates REST API route.
+ *
+ * @param array $template The template object.
+ *
+ * @return array The post type the template is intended for.
+ */
+function gutenberg_rest_template_post_type_callback( $item ) {
+
+	$template_metadata = _get_block_template_file( 'wp_template', $item['slug'] );
+	if ( null === $template_metadata ) {
+		return array();
+	}
+
+	if ( isset( $template_metadata['postTypes'] ) ) {
+		return $template_metadata['postTypes'];
+	}
+
+	return array();
+}
--- a/lib/compat/wordpress-6.6/rest-api.php
+++ b/lib/compat/wordpress-6.6/rest-api.php
@@ -118,12 +118,12 @@ function gutenberg_rest_template_post_type_callback( $item ) {
 
 	$template_metadata = _get_block_template_file( 'wp_template', $item['slug'] );
 	if ( null === $template_metadata ) {
-		return array();
+		return null;
 	}
 
 	if ( isset( $template_metadata['postTypes'] ) ) {
 		return $template_metadata['postTypes'];
 	}
 
-	return array();
+	return null;
 }
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -380,9 +380,6 @@ export default function PageTemplates() {
 					value: key,
 					label: POST_TYPES[ key ],
 				} ) ),
-				filterBy: {
-					operators: [ OPERATOR_IS_ANY ],
-				},
 			},
 			{
 				header: __( 'Type' ),
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -63,7 +63,6 @@ const defaultConfigPerViewType = {
 		mediaField: 'preview',
 		primaryField: 'title',
 		columnFields: [ 'description' ],
-		badgeFields: [ 'isCustom' ],
 	},
 	[ LAYOUT_LIST ]: {
 		primaryField: 'title',
@@ -82,8 +81,7 @@ const DEFAULT_VIEW = {
 	},
 	// All fields are visible by default, so it's
 	// better to keep track of the hidden ones.
-	hiddenFields: [ 'preview' ],
-	// hiddenFields: [ 'preview', 'postTypes' ],
+	hiddenFields: [ 'preview', 'postTypes', 'isCustom' ],
 	layout: defaultConfigPerViewType[ LAYOUT_GRID ],
 	filters: [],
 };
@@ -205,6 +203,9 @@ const TEMPLATE_TO_POST_TYPE = {
 	// 2. Secondary templates.
 	'single-post': [ 'post' ],
 };
+
+const CUSTOM_TEMPLATE = __( 'Custom' );
+const NOT_CUSTOM_TEMPLATE = __( 'Not custom' );
 
 export default function PageTemplates() {
 	const { params } = useLocation();
@@ -388,12 +389,12 @@ export default function PageTemplates() {
 			{
 				header: __( 'Type' ),
 				id: 'isCustom',
-				getValue: ( { item } ) => item.is_custom,
+				getValue: ( { item } ) => !! item.is_custom,
 				render: ( { item } ) =>
-					item.is_custom ? 'Custom' : undefined,
+					!! item.is_custom ? CUSTOM_TEMPLATE : NOT_CUSTOM_TEMPLATE,
 				elements: [
-					{ value: true, label: 'Custom' },
-					{ value: false, label: 'Default' },
+					{ value: true, label: CUSTOM_TEMPLATE },
+					{ value: false, label: NOT_CUSTOM_TEMPLATE },
 				],
 				filterBy: {
 					operators: [ OPERATOR_IS ],
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -36,6 +36,7 @@ import { useAddedBy } from './hooks';
 import {
 	TEMPLATE_POST_TYPE,
 	OPERATOR_IS_ANY,
+	OPERATOR_IS,
 	LAYOUT_GRID,
 	LAYOUT_TABLE,
 	LAYOUT_LIST,
@@ -62,6 +63,7 @@ const defaultConfigPerViewType = {
 		mediaField: 'preview',
 		primaryField: 'title',
 		columnFields: [ 'description' ],
+		badgeFields: [ 'isCustom' ],
 	},
 	[ LAYOUT_LIST ]: {
 		primaryField: 'title',
@@ -337,6 +339,20 @@ export default function PageTemplates() {
 				],
 				filterBy: {
 					operators: [ OPERATOR_IS_ANY ],
+				},
+			},
+			{
+				header: __( 'Type' ),
+				id: 'isCustom',
+				getValue: ( { item } ) => item.is_custom,
+				render: ( { item } ) =>
+					item.is_custom ? 'Custom' : undefined,
+				elements: [
+					{ value: true, label: 'Custom' },
+					{ value: false, label: 'Default' },
+				],
+				filterBy: {
+					operators: [ OPERATOR_IS ],
 				},
 			},
 		],
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -185,12 +185,6 @@ function Preview( { item, viewType } ) {
 	);
 }
 
-// TODO: how about CPTs added by plugins, etc.?
-const POST_TYPES = {
-	post: __( 'Post' ),
-	page: __( 'Page' ),
-};
-
 // This maps the template slug to the post types it should be available for.
 // https://developer.wordpress.org/themes/basics/template-hierarchy/#visual-overview
 // It only addresses primary and secondary templates, but not tertiary (aka variable) templates.
@@ -252,6 +246,23 @@ export default function PageTemplates() {
 			per_page: -1,
 		}
 	);
+	const { records: types } = useEntityRecords( 'root', 'postType', {
+		per_page: -1,
+		context: 'edit',
+	} );
+
+	const registeredPostTypes = useMemo( () => {
+		const result =
+			types
+				?.filter( ( type ) => type.viewable )
+				.map( ( { name, slug } ) => ( { name, slug } ) )
+				.reduce( ( acc, current ) => {
+					acc[ current.slug ] = current.name;
+					return acc;
+				}, {} ) || {};
+		return result;
+	}, [ types ] );
+
 	const history = useHistory();
 	const onSelectionChange = useCallback(
 		( items ) => {
@@ -294,7 +305,7 @@ export default function PageTemplates() {
 
 		return (
 			item.post_types ||
-			( item.is_custom && Object.keys( POST_TYPES ) ) ||
+			( item.is_custom && Object.keys( registeredPostTypes ) ) ||
 			TEMPLATE_TO_POST_TYPE[ item.slug ] ||
 			[]
 		);
@@ -370,20 +381,22 @@ export default function PageTemplates() {
 					}
 
 					if (
-						postTypes.length === Object.keys( POST_TYPES ).length
+						postTypes.length ===
+						Object.keys( registeredPostTypes ).length
 					) {
 						return __( 'Any' );
 					}
 
 					return postTypes
 						.map(
-							( postType ) => POST_TYPES[ postType ] || postType
+							( postType ) =>
+								registeredPostTypes[ postType ] || postType
 						)
 						.join( ',' );
 				},
-				elements: Object.keys( POST_TYPES ).map( ( key ) => ( {
+				elements: Object.keys( registeredPostTypes ).map( ( key ) => ( {
 					value: key,
-					label: POST_TYPES[ key ],
+					label: registeredPostTypes[ key ],
 				} ) ),
 			},
 			{
@@ -401,7 +414,7 @@ export default function PageTemplates() {
 				},
 			},
 		],
-		[ authors, view.type ]
+		[ authors, view.type, registeredPostTypes, getPostTypesFromItem ]
 	);
 
 	const { data, paginationInfo } = useMemo( () => {
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -80,7 +80,7 @@ const DEFAULT_VIEW = {
 	},
 	// All fields are visible by default, so it's
 	// better to keep track of the hidden ones.
-	hiddenFields: [ 'preview' ],
+	hiddenFields: [ 'preview', 'postTypes' ],
 	layout: defaultConfigPerViewType[ LAYOUT_GRID ],
 	filters: [],
 };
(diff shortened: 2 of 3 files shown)


packages/edit-site/src/components/page-templates/index.js
```diff
@@ -315,6 +323,38 @@ export default function PageTemplates() {
...
+				elements: [
+					{ value: 'post', label: POST_TYPES.post },
+					{ value: 'page', label: POST_TYPES.page },
+				],
+				filterBy: {
+					operators: [ OPERATOR_IS_ANY ],
+				},
+			},
+			{
+				header: __( 'Type' ),
```
oandregal, line 403: I've made the following decisions about how to present the `isCustom` field to the user:

- Called it `Type`. Alternatives: `Template type`, `Is custom?`.
- Displayed it as a badge field in the grid layout, as we do for other boolean-type fields (sync status in patterns).
- In any layout, it'd only render "Custom" if the template is custom. If it is not, it'll render an empty value. Alternatively, it could say "Default".

This is a concept that requires a certain knowledge of WordPress templates, so I was unsure about how to simplify it further. Happy to hear thoughts.

jameskoster: Yeah this is very tricky. 

* I'm not sure about displaying the type as a badge; it might not be obvious what this refers to without the label? 
* 'Type' suggests multiple options... It's odd that the column is either empty or 'custom'. Would it be feasible to add other types? E.g. Archive, Single, Utility? Combined with the post type filter this could potentially work quite well.
* I'd probably hide this field by default.

jameskoster:
Unsure how it works, but the post type filter turned up some unexpected results for me:

* post type = page didn't return the `page.html` template
* post type = post didn't return `single-post.html`

When the post types field is visible, it's a little strange that some templates show the field, but others do not, especially in table layout. Should a `–` appear for templates not associated with any particular post type? Similarly should there be an `Any` value for agnostic templates like `Archive.html`?

oandregal:
> Unsure how it works, but the post type filter turned up some unexpected results for me:

I need to verify this logic to be the same we use when the user switches templates:

https://github.com/WordPress/gutenberg/assets/583546/bddf7fd9-95ca-459a-b8f5-321d5b27e9a0

It seems that:

- All the templates registered via `customTemplates` in the theme's `theme.json` ([example](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentytwentyfour/theme.json#L935) for TwentyTwentyFour) should be displayed for its corresponding post type.
- All the templates created by the user should be displayed for both `post` and `page` post types.
- Some extra templates, for example:
  - `single` and `singular` for `post`, if they exist
  -  `page` and `singular` for `page`, if they exist

Does that sound good or am I missing something?

oandregal:
_Edit: I've updated this comment with the latest I've learned about [how the `post_type` query works](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/block-template-utils.php#L1077)._

I'll look into this tomorrow morning, but at the implementation level this is a bit tricky. It seems the next steps are:
 
- On the one hand, the `post_types` field in the endpoint should behave the same as if the endpoint was [queried](https://developer.wordpress.org/rest-api/reference/wp_templates/#retrieve-a-template) with `post_type`. This is, only the templates registered through the `customTemplates` field in `theme.json` would have post type information. Any other template would return null in `postTypes` field.

- At the DataViews level, in addition to the templates with post type information we should add some logic to match other templates that are relevant to the search:
  - The custom templates that don't have post types info attached. These templates are also returned by the endpoint as per [this query logic](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/block-template-utils.php#L1077).
  - The "extra" templates that are relevant for the search (single, page, singular). This are not returned by the templates endpoint but they are the "defaults".

packages/edit-site/src/components/page-templates/index.js
```diff
@@ -315,6 +323,38 @@ export default function PageTemplates() {
...
+				render: ( { item } ) =>
+					item.post_types
+						?.map(
+							( postType ) => POST_TYPES[ postType ] || postType
+						)
+						?.join( ',' ),
+				elements: [
+					{ value: 'post', label: POST_TYPES.post },
+					{ value: 'page', label: POST_TYPES.page },
+				],
```
youknowriad: Are these the only possible post types?

packages/edit-site/src/components/page-templates/index.js
oandregal: > It's odd that the column is either empty or 'custom'. Would it be feasible to add other types? E.g. Archive, Single, Utility? Combined with the post type filter this could potentially work quite well.

To be honest, I'm struggling to see the value of this field/filter in isolation: having the `postTypes` one enables users to search for templates bound that those post types — which are considered custom already, so this `Type` is a redundant. Is there any other value to it I'm not seeing?

For adding more types: is there any stablished categorization of templates I can look at? That could be useful, though I wouldn't want to introduce a new concept to templates in this PR.

If this field doesn't provide much value right now, I'd rather start with `postTypes` only.

oandregal:
I've implemented the [logic above](https://github.com/WordPress/gutenberg/pull/62075#issuecomment-2135669182) and this is how it looks like:

| Templates for post | Templates for pages |
| --- | --- |
| Captura de ecrã 2024-05-29, às 11 19 15 | Captura de ecrã 2024-05-29, às 11 21 47 |

The results make sense to me, though it's a complex topic and I'm not sure the information display helps to make sense of it if you're not familiar with all kind of templates.

- I think that it'd be better if we don't add the `custom` field for now, see [related conversation](https://github.com/WordPress/gutenberg/pull/62075#discussion_r1618358179).
- Note that I've made the `postType` field visible to make it obvious how the filtered value and the rendered value are different. Perhaps this is fine, given the field is hidden by default. I'm posting more thoughts about this as a [code review](https://github.com/WordPress/gutenberg/pull/62075#discussion_r1618567149), as it's a bit different from what we've been doing so far (filter<=>display was a 1:1 match).

packages/edit-site/src/components/page-templates/index.js
```diff
@@ -315,6 +329,50 @@ export default function PageTemplates() {
 				elements: authors,
 				width: '1%',
 			},
+			{
+				header: __( 'Post types' ),
+				id: 'postTypes',
+				getValue: ( { item } ) =>
```
oandregal: This `getValue` function is used by the filters. Note how the values returned here are different from the values returned by `render`, which is not what we've been doing so far.

This enables us to implement [this logic](https://github.com/WordPress/gutenberg/pull/62075#issuecomment-2136968494), in line with [the server logic](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/block-template-utils.php#L1077). For example, when [querying the templates](https://developer.wordpress.org/rest-api/reference/wp_templates/#retrieve-a-template) endpoint with `post_type: post` the server returns:

1. Templates whose `templates->post_types` values contain `post`.
2. User templates that are custom but don't have any post type information.

Additionally, there's a few default templates with special meaning that are not returned by the endpoint in that scenario because they're the default and we should probably display here when the user filters by post type:

3. The `single` & `singular` templates, as per the [template hierarchy](https://developer.wordpress.org/themes/basics/template-hierarchy/#visual-overview).

oandregal: An alternative to this approach would be allowing the field to provide a custom filter callback. I'm investigating this next, though I wanted to share as early as possible in case anyone has different ideas.

packages/edit-site/src/components/page-templates/index.js
```diff
@@ -315,6 +329,50 @@ export default function PageTemplates() {
...
+					// This logic would be used by the filter
+					// and replicates what we have in the server at
+					// https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/block-template-utils.php#L1077
+					//
+					// 1. Consider the post types defined by the item, if any
+					item.post_types ||
+					// 2. Otherwise, if a template is custom add it for any CPT.
+					( item.is_custom ? [ 'post', 'page' ] : item.post_types ) || // TODO: take any CPT.
```
oandregal: A downside of this approach is that you have to provide all the values that _could_ match. For example, in this case, we should provide all possible CPTs in this array.

packages/edit-site/src/components/page-templates/index.js
oandregal: No, they are not. There's a TODO note over there to look at this later, when the custom filtering was addressed.

packages/edit-site/src/components/page-templates/index.js
jameskoster: I think the main motivator was to provide a way to replicate the 'Custom' section in 6.5's Templates sidebar:

Screenshot 2024-05-29 at 14 49 17

Without this filter it wouldn't be possible to view custom templates easily which could be seen as a regression? Possibly a bad idea, but could the filter be hidden in the UI but still function as part of a 'Custom' view? I think pattern categories work a similar way.

---

The [template hierarchy](https://developer.wordpress.org/themes/basics/template-hierarchy/#visual-overview) resolves by page type, essentially; "Archive", "Singular", "Front Page", "Posts Page", "Search Results", "404". 

Grouping by "Archive", "Singular", and "Utility" seems reasonable, but I don't think it's defined in the code. Agree it seems late to be adding this complexity.

jameskoster:
> I've implemented the https://github.com/WordPress/gutenberg/pull/62075#issuecomment-2135669182 and this is how it looks like:

I think this looks good. But shouldn't the `Post types` field be visible on each record? IE the 'Pages' template should have "Post type: Page".

For user-generated custom templates, should the post types value be `Any` as technically these templates could be applied to CPTs too?

We'll need to think about how this works for other template types like archives. E.g. `archive-posts.html` should have "Post types: Post".

Finally, should templates not associated with any particular post type still show the field but with an `n/a` value? It's a bit strange to see an empty cell in table layout.

oandregal:
James, I've pushed some changes that implement your suggestions for post types.

So far, only [primary and secondary templates](https://developer.wordpress.org/themes/basics/template-hierarchy/#visual-overview) are mapped to a post type, the rest will display "n/a". Do we consider support for variable templates (`archive-${posttype}`, `single-${posttype}`, etc) required? It adds a bit of complexity, I can revisit it later if it's a requirement.

I'm looking at "Custom" field/filter & CPTs support next.

https://github.com/WordPress/gutenberg/assets/583546/ba7c449a-90ac-45a2-92fb-62b519e1b4b4

packages/edit-site/src/components/page-templates/index.js
oandregal: I've extracted a `getPostTypesFromItem` method that centralizes all this logic. I feel better about this now. Still need to review/add support for CPTs to assess whether this is a good enough approach.

jameskoster:
> Do we consider support for variable templates (archive-${posttype}, single-${posttype}, etc) required?

For these templates I fear "Post types: n/a" is probably too disingenuous:

Screenshot 2024-05-30 at 11 47 43

It's unclear to users why `pages.html` and `single-post.html` declare the correct post type, but these do not.

There's one other quirk we should probably consider here; the `home.html` template ("Blog Home") will only ever display `post`s, so it could potentially be hard-coded as "Post types: Post".

packages/edit-site/src/components/page-templates/index.js
oandregal: > The [template hierarchy](https://developer.wordpress.org/themes/basics/template-hierarchy/#visual-overview) resolves by page type, essentially; "Archive", "Singular", "Front Page", "Posts Page", "Search Results", "404".

I feel this is the best solution, and I'm also wary to introduce it this close to the beta.

If we agree exposing those types would be best long-term, a plan for 6.6 can be:

- naming the field "Type", so we can reuse it later
- its values for now are "Custom" / "Not custom" — we can add more granularity in 6.7 ("Archive", etc.)

Thoughts?

Pushed this approach at [895214a](https://github.com/WordPress/gutenberg/pull/62075/commits/895214a792352a33e1f91d6e713c8d792c41a9e6). Made the field hidden by default and not a badge as well.

jameskoster: How do you feel about implementing the filter with `custom` / `not custom` as values, but hiding it in the UI for 6.6?

We can then use it to add a 'Custom' view, including a description about what custom templates are. This would be similar (I think) to categories in the Patterns page.

Screenshot 2024-05-30 at 12 23 21

This would be dependent on updating the frame titles, but there's a [PR](https://github.com/WordPress/gutenberg/pull/61983) for that ready to go.

packages/edit-site/src/components/page-templates/index.js
oandregal: Pushed support for any registered post type that is viewable & declares editor support (as a proxy for template support, otherwise, media post type will be listed as well). https://github.com/WordPress/gutenberg/pull/62075/commits/35300bb6c21899af6775caba4b0a2eae8d2ff0c9 and https://github.com/WordPress/gutenberg/pull/62075/commits/18756b3321a7268d5d9cf39f83d2111efca268c4

oandregal:
> For these templates I fear "Post types: n/a" is probably too disingenuous:

I've done some research and find this is plugin territory and has to do with how they provide template metadata. The gist of the issue is that core & plugins have a different understanding of what `post_types` and `is_custom` fields are for. This is what I see:

- templates come from:
  - the theme, automatically picked up by core
  - plugins that register them by using the `get_block_templates` hook
  - the user, when they save changes to existing templates or create their own templates

- `$template->is_custom`:
  - core defines this as "any template that is not one of the [default types](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/block-template-utils.php#L141)". Original PRs: [#35802](https://github.com/WordPress/gutenberg/pull/35802) and [#36911](https://github.com/WordPress/gutenberg/pull/36911).
  - plugins define this as "any template coming from the DB" ([learnpress](https://github.com/LearnPress/learnpress/blob/develop/inc/block-template/class-block-template-handle.php#L99), [sensei](https://github.com/Automattic/sensei/blob/af62fb1115daf2063bc56331a7d8b1b3ea805866/includes/course-theme/class-sensei-course-theme-templates.php#L469)) or always set it to false ([woo](https://github.com/woocommerce/woocommerce/blob/3d3945652e02a6c39fce6494159cf9628a7f467a/plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php#L242), relevant PRs: [introduced `is_custom`](https://github.com/woocommerce/woocommerce-blocks/pull/5062), [set it to always false](https://github.com/woocommerce/woocommerce-blocks/pull/5385)).
  - I'm not sure what `is_custom` is for. The core definition only accounts for [primary & secondary templates](https://developer.wordpress.org/themes/basics/template-hierarchy/#visual-overview) defined by core, but it is not extensible or says anything about variable templates (templates for a CPT: `archive-${post-type}`, `single-${post-type}`, etc.).

- `$template->post_types`:
  - themes provide this info via `customTemplates` in `theme.json`
  - plugins provide this info via the filter and do not to provide any info (learnpress, sensei, [woo](https://github.com/woocommerce/woocommerce/blob/3d3945652e02a6c39fce6494159cf9628a7f467a/plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php#L172)).
  - My expectation would be that plugins fill this property with the relevant CPTs their templates are for.

This PR is working as expected for the templates coming from the theme. However, given the different interpretations plugins have for `is_custom` and `post_types` the feature is not working as an _user would expect_ (e.g.: _give me the templates for the Woo-defined Product CPT_) due to how plugins provide metadata.

For me, the next steps here are:

- Clarify what `is_custom` is for.
- Let the plugins know that we're using `post_types` in more places (dataviews), so they have time to provide data for their templates — or provide a subpar experience in this screen. This means this PR won't be part of 6.6 but we'd land it early in the 6.7 cycle so plugins have time to adapt.

Pinging people that has been involved on this for thoughts and good measure:

- original authors: @Mamaduka @youknowriad @ntsekouras 
- plugins: @Aljullu @opr @tjcafferkey @donnapep

Aljullu:
> `$template->post_types`: plugins provide this info via the filter and do not to provide any info (learnpress, sensei, [woo](https://github.com/woocommerce/woocommerce/blob/3d3945652e02a6c39fce6494159cf9628a7f467a/plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php#L172)).

I'm a bit confused by the _Single product_ template having an empty array as the `post_types` value, honestly. I think this was an overlook and it should be `product` instead. My guess is that as we don't allow switching templates for individual products (https://github.com/woocommerce/woocommerce/issues/42393), having it set to an empty array didn't make any difference functionality-wise.

> Clarify what `is_custom` is for.

Thanks for raising this! `is_custom` is something that has always confused me and I guess other devs as well, seeing several plugins used it in different ways. It's even more confusing considering that a template can have `origin = "custom"`, but that's completely unrelated to the `is_custom` attribute. :sweat_smile: 

My guess is that most plugins have set `is_custom` based on the specific needs they had for their templates, instead of adhering to the formal definition. We could investigate setting `is_custom` to true for WooCommerce templates, but we need to keep an eye on how that value is used across the codebase. Ie, if `is_custom` is true that means templates can be renamed by users (something we tried to avoid)?

> Let the plugins know that we're using `post_types` in more places (dataviews), so they have time to provide data for their templates — or provide a subpar experience in this screen. This means this PR won't be part of 6.6 but we'd land it early in the 6.7 cycle so plugins have time to adapt.

This would work for us. :+1: 

---

Heads-up that we are also working on introducing a template registration API for plugins (https://github.com/WordPress/gutenberg/pull/61577), so we have the opportunity to unify how this is handled across different plugins.

oandregal:
> Heads-up that we are also working on introducing a template registration API for plugins (https://github.com/WordPress/gutenberg/pull/61577), so we have the opportunity to unify how this is handled across different plugins.

Oh, nice, I'll review that! I'm focused on backports today, so I'll take a look in a day or two.

lib/compat/wordpress-6.6/rest-api.php
```diff
@@ -87,3 +87,43 @@ function gutenberg_register_global_styles_revisions_endpoints() {
...
+ * @param array $template The template object.
+ *
+ * @return array The post type the template is intended for.
+ */
+function gutenberg_rest_template_post_type_callback( $item ) {
+
+	$template_metadata = _get_block_template_file( 'wp_template', $item['slug'] );
+	if ( null === $template_metadata ) {
+		return null;
+	}
```
mcsf, line 122: Super minor, but this check isn't needed before `isset`.

```suggestion
```

packages/edit-site/src/components/page-templates/index.js
```diff
@@ -256,6 +290,27 @@ export default function PageTemplates() {
 		} ) );
 	}, [ records ] );
 
+	const getPostTypesFromItem = ( item ) => {
+		// This logic replicates querying the REST templates endpoint with a post_type parameter.
+		// https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/block-template-utils.php#L1077
```
mcsf, line 295: This needs to be a permalink and not just relative to `trunk`, otherwise it will go (and already has gone) stale.

packages/edit-site/src/components/page-templates/index.js
```diff
@@ -256,6 +290,27 @@ export default function PageTemplates() {
 		} ) );
 	}, [ records ] );
 
+	const getPostTypesFromItem = ( item ) => {
```
mcsf, line 293: This should be a `useCallback`, no?

mcsf (commented):
Thanks for the research into what "custom" and "post types" actually mean in this context!

I suppose that, as far as "custom" is concerned, those two different definitions ("is not a default template" / "lives in the database") overlapped until recently. It was only through the improvement of templating infrastructure (including the `get_block_templates` hook) that it became gradually easier to produce new realities for "custom" templates.

Anyway, I don't really have an answer for you on that, and I think it'll take a few discussions and competing opinions before we can untangle this.

Issue #62038, opened 2024-05-27
Update alt text decision tree links to be translatable


joedolson:
### Description

In WordPress 6.6, the core alt text decision tree links will become translatable, so that translators can provide a link to the W3C WAI documentation in their own language, when it exists. 

[Locations where this link is referenced in Gutenberg](https://github.com/search?q=repo%3AWordPress%2Fgutenberg+https%3A%2F%2Fwww.w3.org%2FWAI%2Ftutorials%2Fimages%2Fdecision-tree&type=code)

[Relevant trac ticket 60975](https://core.trac.wordpress.org/ticket/60975)

### Step-by-step reproduction instructions

No reproduction steps relevant; this does not change anything in default behavior.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

t-hamano:
I removed the milestone from this issue because it will be automatically applied when the PR that resolves this issue is merged.

narenin:
@joedolson I am working on this.

colorful-tones:
@narenin, please attach and/or reference a pull request for this when you have a moment. Even if it is still a Draft. Thanks for working on this!

narenin:
@colorful-tones Yes sure, it's done please check.

https://github.com/WordPress/gutenberg/pull/62076

t-hamano:
Fixed by #62076

PR #62076, opened 2024-05-28
Fixed : Update `alt text decision tree` links to be translatable

narenin:
Fixes a bug raised here https://github.com/WordPress/gutenberg/issues/62038

What?

Updated alt text decision tree links to be translatable.


michalczaplinski:
Thank you form working on this @narenin !

This seems fine to me but perhaps @joedolson would like to confirm?

joedolson:
These links should have translator comments to provide context for the link. In the related core patch, you can see an example of the comment in core.

narenin:
@joedolson I have implemented the suggestions, could you please check.

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -211,9 +211,14 @@ export default function CoverInspectorControls( {
 								}
 								help={
 									<>
-										<ExternalLink href={
-												__('https://www.w3.org/WAI/tutorials/images/decision-tree/')
-												}>
+										<ExternalLink
+											href={
+												// translators: Localized tutorial, if one exists. Link includes list of existing translations.
+												__(
```
joedolson, line 217: This is the comment from an earlier version of the patch; we should name the source in the comment, e.g.

`// translators: Localized tutorial, if one exists. W3C Web Accessibility Initiative link has list of existing translations. `

joedolson (changes requested):
Additional comment text change; applies to all files.

packages/block-library/src/cover/edit/inspector-controls.js
narenin: @joedolson  Thanks for the feedback, I have implemented the suggestions, please check.

joedolson (approved):
Thanks, @narenin!

t-hamano:
@narenin

It seems there is a conflict in one file, so could you resolve that conflict? Also, GitHub Actions should be stable on the latest trunk, so I would appreciate it if you could rebase your PR again.

Thank you!

narenin:
@t-hamano Thanks for the feedback. I have resolved the conflict, please have a look.

t-hamano (approved):
@narenin Thanks for the update!

PR #62077, opened 2024-05-28
Update: Show Pre Publish Panel when are are publishing a private post.

enejb:
## What?
Currently when you are publishing a private post it seems like we are just saving the post but the post does show up for logged in users as if it is published even though it is not visible in a lot of contexts in the front end. 

## Why?
We want to treat publishing a private post as if you are actually publishing the post.

## How?
The Pr addresses the issue by adding the prepublish flow as well updates the button to tell the use that they are publishing the post vs just saving it. 

## Testing Instructions
Create a new private post. Does it tell you to save the post or publish it? 

Create a new private post. Does it tell you to save the post or publish it? 

## Screenshots or screencast 

Before:

https://github.com/WordPress/gutenberg/assets/115071/67fc09ec-5b99-4bff-9629-919bd7b41cf7

After:

https://github.com/WordPress/gutenberg/assets/115071/915d0040-b4ac-4c7f-8b29-228988f2de31

### DIFF
--- a/packages/editor/src/components/post-publish-button/label.js
+++ b/packages/editor/src/components/post-publish-button/label.js
@@ -71,7 +71,7 @@ export default function PublishButtonLabel() {
 		hasNonPostEntityChanges ||
 		isPublished ||
 		( postStatusHasChanged &&
-			! [ 'future', 'publish' ].includes( postStatus ) ) ||
+			! [ 'future', 'publish', 'private' ].includes( postStatus ) ) ||
 		( ! postStatusHasChanged && postStatus === 'future' )
 	) {
 		return __( 'Save' );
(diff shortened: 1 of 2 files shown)


ntsekouras:
Thanks for the PR! There are so many nuances with the flow that I can't be sure, but by also testing the previous flow (by deactivating GB). Changing the status to `private` doesn't show the Pre Publish panels. I'm not sure we should add this, but would love some other feedback from others.

https://github.com/WordPress/gutenberg/assets/16275880/98560f0c-10ec-431a-8476-cc71b91abd4e

jameskoster:
Switching from draft/pending to published will invoke the checks, but draft → private → published will bypass them. It’s the same for draft → private → scheduled. 

This could be considered an oversight, so I do think there's a case to be made that switching from an unpublished status (pending, draft) to a published status (private, scheduled, published) should invoke the checks. 

But I'm also curious to hear more thoughts on this, given things have been this way for so long.

PR #59518, opened 2024-03-01
ES Lint: Add a rule to prevent use of the disabled attribute

scruffian:
## What?
Adds a rule to prevent the use of the disabled attribute.

- @alexstine 

## How?
Adds a new ES Lint rule to catch this.

## Testing Instructions
1. Add `disabled` to any React component
2. Try to commit your change
3. The linter should stop you

Issue #56547, opened 2023-11-27
Make sure disabled buttons are always focusable by default


afercia:
### Description

Splitting this out from https://github.com/WordPress/gutenberg/issues/56396#issuecomment-1824180314 for a broader discussion. Cc @ciampo @andrewhayward @alexsanford @joedolson 

As noted in [this comment](https://github.com/WordPress/gutenberg/issues/55110#issuecomment-1822629840), focusable elements that are dynamically added or removed / hidden / disabled potentially trigger a focus loss and impact other utilities, components and hooks like the `@wordpress/dom` utilities `focusable` and `tabbable`, `useConstrainedTabbing` and `useFocusOutside`. There have been several cases so far where focus losses were triggered because of this. When using a keyboard, focus losses are a terrible user experience that needs to be avoided.

RIght now, the Button component has an `__experimentalIsFocusable` prop that is meant to render `aria-disabled` instead of the `disabled` attribute to keep disabled buttons focusable. In https://github.com/WordPress/gutenberg/issues/56396#issuecomment-1824180314 a proposal was made to make this prop `true` bu default:

> any potential drawbacks if the Button component had the `__experimentalIsFocusable` prop true by default?

I'm not sure this would be ideal. If I understand correctly, the current situation is something like this, in pseudo-code:

`<Button disabled __experimentalIsFocusable />`

Conceptually, I find this confusing. `disabled` is a HTML attribute, it should not be used as a prop name. I think this is source of confusion for many new-generation contributors that are used to just use components props without a deep understanding of what happens in terms of the DOM and the rendered HTML.

If I understand correctly, what has been proposed in https://github.com/WordPress/gutenberg/issues/56396#issuecomment-1824180314 would be, in pseudo-code, something like this:

`<Button disabled __experimentalIsFocusable={ true } />` 

And when both props are true, under the hood the button would use `aria-disabled`. I'd think this would be even more confusing because it dosn't explain what it does and it doesn't educate contributors.

## What we need for accessibility

In 90% of the cases, the `disabled` attribute should not be used. In fact, in the editor we established a pattern for 'disabled' buttons to:
- use `aria-disabled` to keep buttons focusable
- `noop` the disabled button
 
This is in place for two reasons:

- avoid focus losses
- discoverablity of disabled controls

See W3C: ARIA Authoring Practices Guide (APG) > Developing a Keyboard Interface > **Focusability of disabled controls**
[...]

PR #9815, opened 2018-09-12
Code Editor: unselect blocks and disable inserter

youknowriad:
closes #9807 

This PR does two things:

 - Always unselect blocks when switching to the code editor to avoid lingering controls in the inspector/toolbar
 - Disable the inserter when we switch to the code editor. It's a very confusing experience to allow it in the Code editor (see #9807)

PR #62080, opened 2024-05-28
Add lint rule for inaccessible disabled `Button`

mirka:
Alternative to #59518
Part of #56547

## What?

Adds an eslint error for when the `disabled` prop is used without the `__experimentalIsFocusable` prop in the `Button` component.

## Why?

We have long been discussing how to prevent accessibility issues stemming from `disabled` controls (#56547). Although there are [situations](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) where you do want to completely disable a control, we more often find cases where this was not a conscious decision but simply not considered at all. In these cases, we often end up with inaccessible disabled controls, which are not perceivable by screen readers, or cause focus loss after closing a popover.

### Next steps

- Stabilize the `__experimentalIsFocusable` prop. We should consider renaming it to `accessibleWhenDisabled`, which is clear and matches the naming in Ariakit. (#62282)
- Add a `accessibleWhenDisabled` prop to other relevant components, and add those components to the eslint rule.

## How?

Other strategies considered include:

- Changing the behavior of the `disabled` prop to be accessible disabled by default. This is not ideal because it goes against the standard behavior of the `disabled` HTML attribute.
- Changing the name of the `disabled` prop to `isDisabled`, and then make that accessible disabled by default. This is not ideal because it will require a lot of API changes, and the behavioral difference between the two props would be unclear.
- Adding a blanket lint rule for `disabled`, and nudge people toward using `aria-disabled` instead. This is [not ideal](https://github.com/WordPress/gutenberg/pull/59518#issuecomment-1977130497) because using `aria-disabled` is not necessarily going to address all issues, depending on the element/component.
- [Automatically enable](https://github.com/WordPress/gutenberg/issues/56547#issuecomment-1895725290) accessible disabled when a control is dynamically disabled while focused. This is not ideal because it [isn't foolproof](https://github.com/WordPress/gutenberg/pull/59518#issuecomment-2090525451), and also does not address the screen reader perceivability problem.

## Testing Instructions

✅ The linter should error in places where a `Button` has a `disabled` prop without a `__experimentalIsFocusable` prop.
✅ The error message should be clear and actionable.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -199,6 +199,12 @@ module.exports = {
 				message:
 					'Avoid truthy checks on length property rendering, as zero length is rendered verbatim.',
 			},
+			{
+				selector:
+					'JSXOpeningElement[name.name="Button"]:not(:has(JSXAttribute[name.name="__experimentalIsFocusable"])) JSXAttribute[name.name="disabled"]',
+				message:
+					'`disabled` used without the `__experimentalIsFocusable` prop. Disabling a control without maintaining focusability can cause accessibility issues, by hiding their presence from screen reader users, or preventing focus from returning to a trigger element. (Ignore this error if you truly mean to disable.)',
+			},
 		],
 	},
 	overrides: [
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -199,12 +199,6 @@ module.exports = {
 				message:
 					'Avoid truthy checks on length property rendering, as zero length is rendered verbatim.',
 			},
-			{
-				selector:
-					'JSXOpeningElement[name.name="Button"]:not(:has(JSXAttribute[name.name="__experimentalIsFocusable"])) JSXAttribute[name.name="disabled"]',
-				message:
-					'`disabled` used without the `__experimentalIsFocusable` prop. Disabling a control without maintaining focusability can cause accessibility issues, by hiding their presence from screen reader users, or preventing focus from returning to a trigger element. (Ignore this error if you truly mean to disable.)',
-			},
 		],
 	},
 	overrides: [
@@ -251,6 +245,21 @@ module.exports = {
 								message: `use cross-platform <${ componentName } /> component instead.`,
 							};
 						} ),
+					},
+				],
+			},
+		},
+		{
+			files: [ 'packages/*/src/**/*.[tj]s?(x)' ],
+			excludedFiles: [ '**/*.native.js' ],
+			rules: {
+				'no-restricted-syntax': [
+					'error',
+					{
+						selector:
+							'JSXOpeningElement[name.name="Button"]:not(:has(JSXAttribute[name.name="__experimentalIsFocusable"])) JSXAttribute[name.name="disabled"]',
+						message:
+							'`disabled` used without the `__experimentalIsFocusable` prop. Disabling a control without maintaining focusability can cause accessibility issues, by hiding their presence from screen reader users, or preventing focus from returning to a trigger element. (Ignore this error if you truly mean to disable.)',
 					},
 				],
 			},
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -250,7 +250,10 @@ module.exports = {
 			},
 		},
 		{
-			files: [ 'packages/*/src/**/*.[tj]s?(x)' ],
+			files: [
+				'packages/*/src/**/*.[tj]s?(x)',
+				'storybook/stories/**/*.[tj]s?(x)',
+			],
 			excludedFiles: [ '**/*.native.js' ],
 			rules: {
 				'no-restricted-syntax': [
--- a/packages/block-directory/src/plugins/get-install-missing/install-button.js
+++ b/packages/block-directory/src/plugins/get-install-missing/install-button.js
@@ -42,6 +42,7 @@ export default function InstallButton( { attributes, block, clientId } ) {
 					}
 				} )
 			}
+			__experimentalIsFocusable
 			disabled={ isInstallingBlock }
 			isBusy={ isInstallingBlock }
 			variant="primary"
--- a/packages/block-editor/src/components/button-block-appender/index.js
+++ b/packages/block-editor/src/components/button-block-appender/index.js
@@ -60,7 +60,7 @@ function ButtonBlockAppender(
 						onClick={ onToggle }
 						aria-haspopup={ isToggleButton ? 'true' : undefined }
 						aria-expanded={ isToggleButton ? isOpen : undefined }
-						// TODO: Investigate whether this button should be accessible when disabled.
+						// Disable reason: There shouldn't be a case where this button is disabled but not visually hidden.
 						// eslint-disable-next-line no-restricted-syntax
 						disabled={ disabled }
 						label={ label }
--- a/packages/block-editor/src/components/button-block-appender/index.js
+++ b/packages/block-editor/src/components/button-block-appender/index.js
@@ -60,6 +60,8 @@ function ButtonBlockAppender(
 						onClick={ onToggle }
 						aria-haspopup={ isToggleButton ? 'true' : undefined }
 						aria-expanded={ isToggleButton ? isOpen : undefined }
+						// TODO: Investigate whether this button should be accessible when disabled.
+						// eslint-disable-next-line no-restricted-syntax
 						disabled={ disabled }
 						label={ label }
 					>
--- a/packages/block-editor/src/components/link-control/link-preview.js
+++ b/packages/block-editor/src/components/link-control/link-preview.js
@@ -149,6 +149,7 @@ export default function LinkPreview( {
 						isEmptyURL || showIconLabels ? '' : ': ' + value.url
 					) }
 					ref={ ref }
+					__experimentalIsFocusable
 					disabled={ isEmptyURL }
 					size="compact"
 				/>
--- a/packages/block-library/src/gallery/v1/gallery-image.js
+++ b/packages/block-library/src/gallery/v1/gallery-image.js
@@ -222,6 +222,8 @@ class GalleryImage extends Component {
 						onClick={ isFirstItem ? undefined : onMoveBackward }
 						label={ __( 'Move image backward' ) }
 						aria-disabled={ isFirstItem }
+						// Disable reason: Truly disable when image is not selected.
+						// eslint-disable-next-line no-restricted-syntax
 						disabled={ ! isSelected }
 					/>
 					<Button
@@ -229,6 +231,8 @@ class GalleryImage extends Component {
 						onClick={ isLastItem ? undefined : onMoveForward }
 						label={ __( 'Move image forward' ) }
 						aria-disabled={ isLastItem }
+						// Disable reason: Truly disable when image is not selected.
+						// eslint-disable-next-line no-restricted-syntax
 						disabled={ ! isSelected }
 					/>
 				</ButtonGroup>
@@ -237,12 +241,16 @@ class GalleryImage extends Component {
 						icon={ edit }
 						onClick={ this.onEdit }
 						label={ __( 'Replace image' ) }
+						// Disable reason: Truly disable when image is not selected.
+						// eslint-disable-next-line no-restricted-syntax
 						disabled={ ! isSelected }
 					/>
 					<Button
 						icon={ closeSmall }
 						onClick={ onRemove }
 						label={ __( 'Remove image' ) }
+						// Disable reason: Truly disable when image is not selected.
+						// eslint-disable-next-line no-restricted-syntax
 						disabled={ ! isSelected }
 					/>
 				</ButtonGroup>
--- a/packages/block-library/src/page-list/convert-to-links-modal.js
+++ b/packages/block-library/src/page-list/convert-to-links-modal.js
@@ -27,6 +27,7 @@ export function ConvertToLinksModal( { onClick, onClose, disabled } ) {
 				</Button>
 				<Button
 					variant="primary"
+					__experimentalIsFocusable
 					disabled={ disabled }
 					onClick={ onClick }
 				>
--- a/packages/block-library/src/page-list/edit.js
+++ b/packages/block-library/src/page-list/edit.js
@@ -338,6 +338,7 @@ export default function PageListEdit( {
 						<p>{ convertDescription }</p>
 						<Button
 							variant="primary"
+							__experimentalIsFocusable
 							disabled={ ! hasResolvedPages }
 							onClick={ convertToNavigationLinks }
 						>
--- a/packages/block-library/src/template-part/edit/title-modal.js
+++ b/packages/block-library/src/template-part/edit/title-modal.js
@@ -43,8 +43,8 @@ export default function TitleModal( { areaLabel, onClose, onSubmit } ) {
 						<Button
 							variant="primary"
 							type="submit"
+							__experimentalIsFocusable
 							disabled={ ! title.length }
-							aria-disabled={ ! title.length }
 						>
 							{ __( 'Create' ) }
 						</Button>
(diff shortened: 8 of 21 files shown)


mirka:
When you run `npm run lint:js -- --quiet`, we can see that there are currently 26 violations of this error.

Is anyone available to do a quick sweep of these and fix or ignore as appropriate? If not, I can go in and ignore them with some kind of TODO comment to fix when possible.

.eslintrc.js
```diff
@@ -249,6 +249,21 @@ module.exports = {
...
 		},
+		{
+			files: [ 'packages/*/src/**/*.[tj]s?(x)' ],
+			excludedFiles: [ '**/*.native.js' ],
+			rules: {
+				'no-restricted-syntax': [
+					'error',
+					{
+						selector:
+							'JSXOpeningElement[name.name="Button"]:not(:has(JSXAttribute[name.name="__experimentalIsFocusable"])) JSXAttribute[name.name="disabled"]',
```
tyxla, line 263: Confirmed that the rule works well 👍

.eslintrc.js
```diff
@@ -249,6 +249,21 @@ module.exports = {
 				],
 			},
 		},
+		{
+			files: [ 'packages/*/src/**/*.[tj]s?(x)' ],
```
tyxla: Should we also enable this for the `storybook` root dir, which may contain some relevant usages?

Example:
https://github.com/WordPress/gutenberg/blob/77a8b55446c7a92ebdd430f2f09eaa804c32f66b/storybook/stories/playground/with-undo-redo/index.js#L51

tyxla (commented):
Nice. Haven't approved yet as we still need to deal with the failures.

DaniGuardiola:
FWIW, Ariakit handles a lot of this out of the box. I feel like before doing something like this we should just consider re-writing the internals to use Ariakit, keeping the same API. Then, it should be easier to figure out a strategy to nudge users to keep things accessible. This includes some ideas that I have explored in the past like making buttons with tooltips or binded as popover disclosure triggers automatically `accessibleWhenDisabled`, which could take care of most, if not all of this issue, while remaining simpler.

mirka:
> FWIW, Ariakit handles a lot of this out of the box. I feel like before doing something like this we should just consider re-writing the internals to use Ariakit, keeping the same API. Then, it should be easier to figure out a strategy to nudge users to keep things accessible. This includes some ideas that I have explored in the past like making buttons with tooltips or binded as popover disclosure triggers automatically `accessibleWhenDisabled`, which could take care of most, if not all of this issue, while remaining simpler.

I have [suggested](https://github.com/WordPress/gutenberg/issues/56547#issuecomment-1895725290) this type of automatic mitigation as a preferred first step as well, but apparently there have already been a couple cases where this would not have been sufficient. Your idea of targeting all buttons bound as popover triggers sounds good, and would work in certain integrated components, but probably not all possible instances of a popover trigger.

My feeling now is that these linting and automating strategies are not mutually exclusive, and would complement each other. We should continue to look into automating strategies, since that would benefit all consumers and not just Gutenberg where this lint rule would be in effect.

People are already growing weary of the repeated accessibility issues that this introduces, so I'm hoping we can at least prevent new problematic instances and improve developer awareness, while working on some automated strategies in parallel. If they work well enough, maybe we can remove the lint rule. Does that sound reasonable?

.eslintrc.js
mirka: Yes, good catch!

packages/editor/src/components/post-preview-button/test/index.js
```diff
@@ -144,7 +144,10 @@ describe( 'PostPreviewButton', () => {
 
 		render( <PostPreviewButton /> );
 
-		expect( screen.getByRole( 'button' ) ).toBeDisabled();
+		expect( screen.getByRole( 'button' ) ).toHaveAttribute(
+			'aria-disabled',
+			'true'
+		);
```
mirka, line 151: This is the recommended way to test for `aria-disabled` (see https://github.com/testing-library/jest-dom/issues/144#issuecomment-577235097).

packages/block-directory/src/plugins/get-install-missing/install-button.js
```diff
@@ -42,6 +42,7 @@ export default function InstallButton( { attributes, block, clientId } ) {
 					}
 				} )
 			}
+			__experimentalIsFocusable
```
mirka, line 45: Clearly a busy state, which can cause focus loss.

packages/block-editor/src/components/button-block-appender/index.js
```diff
@@ -60,6 +60,8 @@ function ButtonBlockAppender(
 						onClick={ onToggle }
 						aria-haspopup={ isToggleButton ? 'true' : undefined }
 						aria-expanded={ isToggleButton ? isOpen : undefined }
+						// TODO: Investigate whether this button should be accessible when disabled.
+						// eslint-disable-next-line no-restricted-syntax
```
mirka: This was the only instance I didn't feel confident enough to make a decision without deeper investigation. This is a publicly exported component with both a `tabIndex` and `disabled` prop exposed. (But no usages of these props within the Gutenberg repo.)

packages/block-editor/src/components/link-control/link-preview.js
```diff
@@ -149,6 +149,7 @@ export default function LinkPreview( {
 						isEmptyURL || showIconLabels ? '' : ': ' + value.url
 					) }
 					ref={ ref }
+					__experimentalIsFocusable
```
mirka, line 152: Alleviates confusion, since this is a "standard" button usually present in the view.

packages/block-library/src/gallery/v1/gallery-image.js
```diff
@@ -222,13 +222,17 @@ class GalleryImage extends Component {
...
 						disabled={ ! isSelected }
 					/>
 					<Button
 						icon={ chevronRight }
 						onClick={ isLastItem ? undefined : onMoveForward }
 						label={ __( 'Move image forward' ) }
 						aria-disabled={ isLastItem }
+						// Disable reason: Truly disable when image is not selected.
+						// eslint-disable-next-line no-restricted-syntax
 						disabled={ ! isSelected }
```
mirka, line 236: I was wondering whether a component having both `disabled` and `aria-disabled` props was something to flag with another lint, but this I think is a legitimate use case. (Though the same logic can also be expressed with a combination of the `disabled` and the `__experimentalIsFocusable` prop.)

packages/block-library/src/page-list/convert-to-links-modal.js
```diff
@@ -27,6 +27,7 @@ export function ConvertToLinksModal( { onClick, onClose, disabled } ) {
 				</Button>
 				<Button
 					variant="primary"
+					__experimentalIsFocusable
```
mirka, line 30: Should be perceivable, and doesn't clutter tab sequence.

packages/block-library/src/page-list/edit.js
```diff
@@ -338,6 +338,7 @@ export default function PageListEdit( {
 						<p>{ convertDescription }</p>
 						<Button
 							variant="primary"
+							__experimentalIsFocusable
```
mirka, line 341: Should be perceivable, especially because the description is always present.

packages/block-library/src/template-part/edit/title-modal.js
```diff
@@ -43,8 +43,8 @@ export default function TitleModal( { areaLabel, onClose, onSubmit } ) {
 						<Button
 							variant="primary"
 							type="submit"
+							__experimentalIsFocusable
 							disabled={ ! title.length }
-							aria-disabled={ ! title.length }
```
mirka, line 47: `disabled` and `aria-disabled` were set with an identical condition, which is clearly a mistake, but signals the intent to keep it focusable.

packages/dataviews/src/item-actions.tsx
```diff
@@ -254,6 +254,7 @@ function CompactItemActions< Item extends AnyItem >( {
 					size="compact"
 					icon={ moreVertical }
 					label={ __( 'Actions' ) }
+					__experimentalIsFocusable
```
mirka, line 257: This is a dropdown trigger, so it can potentially cause focus loss.

packages/dataviews/src/view-list.tsx
```diff
@@ -255,6 +255,7 @@ function ListItem< Item extends AnyItem >( {
 												size="compact"
 												icon={ moreVertical }
 												label={ __( 'Actions' ) }
+												__experimentalIsFocusable
```
mirka, line 258: An empty action menu should still be perceivable to alleviate confusion, and does not clutter tab sequence due to `Composite` use.

packages/edit-post/src/components/preferences-modal/enable-custom-fields.js
```diff
@@ -42,6 +42,7 @@ export function CustomFieldsConfirmation( { willEnable } ) {
 				className="edit-post-preferences-modal__custom-fields-confirmation-button"
 				variant="secondary"
 				isBusy={ isReloading }
+				__experimentalIsFocusable
```
mirka, line 45: Clearly a busy state, which can cause focus loss.

packages/edit-site/src/components/global-styles-sidebar/index.js
mirka, line 1: Both buttons should be perceivable, because they are expected to be there.

packages/edit-site/src/components/global-styles/screen-revisions/revisions-buttons.js
```diff
@@ -219,7 +220,6 @@ function RevisionsButtons( {
 								</p>
 							) : (
 								<Button
-									disabled={ areStylesEqual }
```
mirka, line 222: Due to the rendering logic, this button will never be shown when `areStylesEqual` is truthy. (⚠️ Please double-check that I'm reading this correctly!)

packages/edit-site/src/components/global-styles/screen-revisions/revisions-buttons.js
```diff
@@ -163,6 +163,7 @@ function RevisionsButtons( {
 					>
 						<Button
 							className="edit-site-global-styles-screen-revisions__revision-button"
+							__experimentalIsFocusable
```
mirka, line 166: This button can contain a lot of important information and should be perceivable even when disabled.

packages/edit-site/src/components/pagination/index.js
mirka, line 1: These pagination buttons wouldn't cause confusion when truly disabled, and removing irrelevant buttons from the tab sequence would help a keyboard user get to the nav button they want quicker.

packages/edit-site/src/components/sidebar-navigation-screen-navigation-menu/rename-modal.js
```diff
@@ -43,6 +43,7 @@ export default function RenameModal( { menuTitle, onClose, onSave } ) {
 
 						<Button
 							__next40pxDefaultSize
+							__experimentalIsFocusable
```
mirka, line 46: Should be perceivable to signal that the user input is invalid.

packages/editor/src/components/post-preview-button/index.js
```diff
@@ -168,6 +168,7 @@ export default function PostPreviewButton( {
 			className={ className || 'editor-post-preview' }
 			href={ href }
 			target={ targetId }
+			__experimentalIsFocusable
```
mirka, line 171: Should be perceivable, because the button is expected to be there.

packages/editor/src/components/post-publish-panel/index.js
```diff
@@ -93,6 +93,7 @@ export class PostPublishPanel extends Component {
 							</div>
 							<div className="editor-post-publish-panel__header-cancel-button">
 								<Button
+									__experimentalIsFocusable
```
mirka, line 96: Can cause focus loss due to a dynamic disable.

packages/list-reusable-blocks/src/components/import-form/index.js
```diff
@@ -86,6 +86,7 @@ function ImportForm( { instanceId, onUpload } ) {
 			<Button
 				type="submit"
 				isBusy={ isLoading }
+				__experimentalIsFocusable
```
mirka, line 89: Should be perceivable, and can cause focus loss.

storybook/stories/playground/with-undo-redo/index.js
mirka, line 1: Accessible disabled matches the behavior of the undo/redo buttons in the actual WP block editor.

mirka (commented):
I went ahead and looked through all of the errors myself, since I realized it would help me get a better overview of our problems.

For most of them I made a decision based on looking at the code only. Some of them, I looked into the original PR, blames, or actual app UI to gather more context. None of them are _tested_ in app. I believe this is sufficient because technically this shouldn't be introducing "regressions" — there is little damage that adding a `__experimentalIsFocusable` could do. The only potentially regressive change is https://github.com/WordPress/gutenberg/pull/62080#discussion_r1621385134, but the logic looks clear enough to me.

I wrote brief reasoning comments for each change to leave a trail in case somebody needs to `blame`.

mirka:
> My feeling now is that these linting and automating strategies are not mutually exclusive, and would complement each other.

After going through all these errors manually, I am now more convinced that we should be approaching this with a combination of automated mitigation and human decision-making (prompted by lints). Automated mitigation can prevent a baseline number of accessibility issues, but a human making context-aware design decisions for each instance can further enhance the UX for screen reader users.

packages/edit-site/src/components/global-styles/screen-revisions/revisions-buttons.js
tyxla, line 222: Confirmed that this can and should be removed because `areStylesEqual` will never be true since it's already `false` based on the ternary logic this is nested in.

packages/editor/src/components/post-preview-button/test/index.js
tyxla: Or maybe we can keep both assertions as they check for different things?

packages/block-editor/src/components/button-block-appender/index.js
tyxla: If I did the archeology correctly, it was added here: https://github.com/WordPress/gutenberg/pull/9815

It can now be seen when inserting buttons in the `Buttons` block:

Screenshot 2024-05-31 at 14 03 26

Judging by the functionality and the initial intent, my expectation is that this `disabled` was left there unintentionally and there's no reason for it to be there in the first place. I expect that the inserter appender will not be visible at all, rather than stay visible and disabled. With that in mind, I don't think it needs to be accessible if it ends up disabled.

tyxla (approved):
Thanks for taking the time to audit and fix them one by one! 👍 

I spent some time going through them all with your comments and I think they all make sense. I've left a few comments where there was doubt or uncertainty.

I think it can be shipped - do you think we should ask folks from #59518 for some feedback just in case?

packages/editor/src/components/post-preview-button/test/index.js
mirka: That makes a lot of sense, actually. Addressed in bca37a2.

packages/block-editor/src/components/button-block-appender/index.js
mirka: Thank you so much for investigating! I added a disable reason in d16d2b4.

mirka:
> I think it can be shipped - do you think we should ask folks from #59518 for some feedback just in case?

Thanks for the review, I appreciate it 🙏 

Since this touches many files, I'll go ahead and merge today. It should be easy enough to revert or tweak, whether in full or in part.

afercia:
Thanks everyone for all your effort ❤️  A linting rule seems the best approach to prevent such a11y issues and, more importantly, to improve knowledge and education.
If needed, all the cases investigated here can be tweaked but it's very good there's now a solid base to catch violations.

ellatrix:
This is a bug fix no? Looks like a bunch of buttons were made more accessible.

ellatrix:

Issue #61941, opened 2024-05-24
Synced Pattern: A possible error when a block is unmounted


Mamaduka:
I noticed this while inspecting warnings logged in debug mode - #61929.

The following e2e test for synced patterns triggers an error - `can be inserted after refresh`. It most likely happens when a block is unmounted due to page navigation.

I can't reproduce it manually because, unlike e2e tests, the browser's DevTools console doesn't keep logs from previous pages.

## Steps to reproduce
1. Enabled script debug mode for testing env. Not required but produces better logs. See the snippet below.
2. Start the `wp-env`/
3. Open pattern e2e test in UI - `npm run test:e2e -- test/e2e/specs/editor/various/patterns.spec.js --ui`
4. Run the `can be inserted after refresh` test.
5. See the error logged in the Console.

<details><summary>SCRIPT_DEBUG Patch</summary>
<p>

```diff
diff --git .wp-env.json .wp-env.json
index 20d5597e54b..57d4e1a72d8 100644
--- .wp-env.json
+++ .wp-env.json
@@ -4,6 +4,9 @@
 	"themes": [ "./test/emptytheme" ],
 	"env": {
 		"tests": {
+			"config": {
+				"SCRIPT_DEBUG": true
+			},
 			"mappings": {
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",

``` 

</p>
</details> 

## Screenshot
![CleanShot 2024-05-24 at 14 38 15](https://github.com/WordPress/gutenberg/assets/240569/63391b6f-1f2b-4560-9352-0d30a1720770)

youknowriad:
Maybe you can try reproducing by switching to "text editor" or something.

Mamaduka:
> Maybe you can try reproducing by switching to "text editor" or something.

I tried, and it didn't work :/

t-hamano:
I looked into this issue and found that the error occurs here.

https://github.com/WordPress/gutenberg/blob/62b0efd7af66c18d5717028e75cf8ff0c441c196/test/e2e/specs/editor/various/patterns.spec.js#L391

The following changes seem to resolve the issue.

```diff
diff --git a/test/e2e/specs/editor/various/patterns.spec.js b/test/e2e/specs/editor/various/patterns.spec.js
index 77f97ee300..daa962aa36 100644
--- a/test/e2e/specs/editor/various/patterns.spec.js
[...]

PR #62082, opened 2024-05-29
Fix error when reloading while saving a pattern

talldan:
## What?
See #61941

An error is currently happening when running the synced patterns e2e test. It's likely to not cause any user facing issues, but can be addressed.

## How?
On this line there's a chance `defaultView` is null-ish, so we should check for that before trying to access the `IntersectionObserver` property in the early return.

## Testing Instructions
Run the synced patterns e2e tests:
```
npm run test:e2e:debug -- test/e2e/specs/editor/various/patterns.spec.js
```

and trigger the `can be inserted after refresh` test case. The error shouldn't happen

### DIFF
--- a/packages/block-editor/src/components/block-list/use-block-props/use-scroll-into-view.js
+++ b/packages/block-editor/src/components/block-list/use-block-props/use-scroll-into-view.js
@@ -10,7 +10,7 @@ export function useScrollIntoView( { isSelected } ) {
 			if ( isSelected ) {
 				const { ownerDocument } = node;
 				const { defaultView } = ownerDocument;
-				if ( ! defaultView.IntersectionObserver ) {
+				if ( ! defaultView?.IntersectionObserver ) {
 					return;
 				}
 				const observer = new defaultView.IntersectionObserver(


packages/block-editor/src/components/block-list/use-block-props/use-scroll-into-view.js
```diff
@@ -10,7 +10,7 @@ export function useScrollIntoView( { isSelected } ) {
 			if ( isSelected ) {
 				const { ownerDocument } = node;
 				const { defaultView } = ownerDocument;
-				if ( ! defaultView.IntersectionObserver ) {
+				if ( ! defaultView?.IntersectionObserver ) {
```
ellatrix, line 13: What causes `defaultView` to be null?

ellatrix: Is this just in test or also manually reproducible?

talldan: There's some more discussion over here - https://github.com/WordPress/gutenberg/issues/61941#issuecomment-2136534664.

Not entirely sure what causes it, but my best guess is it's a side effect of the page unload, maybe the DOM elements being detached from the DOM.

Not really sure if it's something we need to fix. I just put the PR to demonstrate that the error goes away after this change for that e2e test.

PR #60453, opened 2024-04-04
Remove block renaming control from advanced inspector controls group

talldan:
## What?
Closes #60118

As proposed in that PR, removes the block renaming input from the advanced section of inspector controls.

## How?
Highlight the code, press delete key

## Testing Instructions
1. Add a block that supports renaming (paragraph, group)
2. Try renaming it in the advanced section, find out you can't
3. Try renaming it via list view, it should still work

### Before
Screenshot 2024-04-04 at 2 25 08 pm

### After
Screenshot 2024-04-04 at 2 26 03 pm

PR #62084, opened 2024-05-29
Fix: add back "Block Name" in advanced inspector

### LINKED ISSUE #62079: Add back "Block Name" text field in Advanced Inspector
In #60234 the "Block Name" Text field in the Advanced Inspector was removed. 

Having used this in a few Gutenberg Releases now, I think it's a major mistake, and we should add it back before WordPress 6.6 is released. 

Renaming is a very common action on sites with more complex page structures. However, the flow for performing this action has become much slower and more complex. 

image

fabiankaegy:
## What?

This PR reverts #60453 and therefore closes #62079

## Why?

Having used this in a few Gutenberg Releases now, I think it's a major mistake, and we should add it back before WordPress 6.6 is released.

Renaming is a very common action on sites with more complex page structures. However, the flow for performing this action has become much slower and more complex.

## How?

By reverting #60453 and making sure that the rename text field is disabled when the block has active pattern overrides.

## Testing Instructions

 - Navigate to the Advanced Inspector of any block and see the Block Name text field appear again
 - Rename a block using the contorl
 - Create a synced pattern with pattern overrides applied. 
 - See that the text field is disabled when the block has overrides enabled.

### DIFF
--- a/packages/block-editor/src/hooks/block-renaming.js
+++ b/packages/block-editor/src/hooks/block-renaming.js
@@ -3,6 +3,13 @@
  */
 import { addFilter } from '@wordpress/hooks';
 import { hasBlockSupport } from '@wordpress/blocks';
+import { __ } from '@wordpress/i18n';
+import { TextControl } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import { InspectorControls } from '../components';
 
 /**
  * Filters registered block settings, adding an `__experimentalLabel` callback if one does not already exist.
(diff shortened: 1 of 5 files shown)


jarekmorawski:
The field's placement under Advanced always felt a bit odd to me. Everything in the sidebar refers to the block's functionality, behavior, or appearance on the front-end, but the name is only used within the editor.

Would it help if we moved it to the ellipsis in the block details section at the top?

image

Alternatively, we could should a `Rename` button next to the block's name:

https://github.com/WordPress/gutenberg/assets/79307566/fcb2d1d4-d8ec-41bb-8622-957dfc2dc23d

richtabor:
> Alternatively, we could should a Rename button next to the block's name:

A similar pattern exists for renaming pages, but I'm not sure it's much more helpful than the existing "Rename" control in the block controls. 

CleanShot 2024-05-29 at 09 14 20

PR #62085, opened 2024-05-29
Editor: Unify the MediaUpload hook between post and site editors.

youknowriad:
Related #52632 

## What?

For both the post and site editors, we override the MediaUpload component with the exact same code. This is just a small PR that moves that override to the "editor" package. There should be no functional change here.


Mamaduka (approved):
Smoked tested uploads work as before ✅

PR #62086, opened 2024-05-29
Block Editor: Check for multiple block usage in the block-editor package

youknowriad:
Related #52632 

## What?

While navigating the post editor code base, I noticed that we have some code that shows a warning when a block that doesn't allow multiple usage is used more than once. 

That said, this code has nothing specific to the post editor and should be something that is bundled in the block-editor package to work in all contexts (like the site editor).

### Testing instructions

1- Create a page with a "more" block. 
2- Copy/paste the block to insert it twice
3- Notice the warning message
4- Save the page and open it in the site editor
5- The warning should still be visible.


packages/block-editor/src/components/block-list/validate-multiple-usage.js
```diff
@@ -0,0 +1,64 @@
...
+	} );
+}
+
+export function MultipleUsageWarning( {
+	originalBlockClientId,
+	name,
+	onReplace,
+} ) {
+	const { selectBlock } = useDispatch( blockEditorStore );
+	const blockType = getBlockType( name );
```
ellatrix: Should we use the store here?

packages/block-editor/src/components/block-edit/index.js
```diff
@@ -49,6 +53,7 @@ export default function BlockEdit( {
 	const layoutSupport =
 		hasBlockSupport( name, 'layout', false ) ||
 		hasBlockSupport( name, '__experimentalLayout', false );
+	const originalBlockClientId = useOriginalBlockOnlyUseOnce( clientId );
```
ellatrix: Not sure I like this, we are adding another block-editor subscription. Could you move this to the private context provider where we select all the needed data for a block?

youknowriad: We can :) just noting that we were already doing this though.

youknowriad: Let me know if I did it properly.

ellatrix (approved):
Looks like we have an e2e test so that's good :)

PR #62087, opened 2024-05-29
Document Bar: Decode HTML entities and take into account cases where there is no title

t-hamano:
## What?

This PR makes two changes to the document bar title:

- Decodes HTML entities
- If there is no title yet, displays `No Title`

## Why?

This is to match the logic in the post card panel:

https://github.com/WordPress/gutenberg/blob/3ce3bbbd3dd2bcb98bbf9a530896c4bf64f7dd63/packages/editor/src/components/post-card-panel/index.js#L87

image

## Testing Instructions

- Create a new post.
- The document bar will show "No Title".
- Include HTML entities in the post title, such as `&<>`.
- HTML entities should be decoded in the document bar title.
- You should be able to get the same results using "Pages" in the site editor

### Before

image

image

### After

image

image


Mamaduka (approved):
Thank you, @t-hamano!

t-hamano:
@Mamaduka Thanks for the review!

PR #62088, opened 2024-05-29
Block Variations: Support dot notation in `isActive` string array

### LINKED ISSUE #62068: Block Variations: Support dot notation in `isActive` string array
Block variations support an `isActive` property which can be either a function, or an array of strings (`string[]`), documented [here](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#using-isactive). _tl;dr:_ Given a block instance, that variation is used to determine which variation is currently active: If `isActive` is a function, it is given the block instance's attributes and the variation's attributes as arguments. It will return `true` if the block instance matches the variation. The `string[]` version of `isActive` is thought of as a shorthand of the above, where all items in the array are attribute names that will be compared.

As documented [here](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#caveats-to-using-isactive) and [here](https://github.com/WordPress/gutenberg/issues/41303), it's possible that multiple variations' `isActive` criteria match a given block instance, which can lead to the wrong variation being determined to be active.

There's a [potential fix](https://github.com/WordPress/gutenberg/pull/62031) that solves this problem by choosing the matching variation with the highest specificity (i.e. number of items in the `isActive` array), but that only works if all matching variations have `isActive` _arrays_ rather than functions.

One notable case where it's impossible to use an array for `isActive` is when an attribute is an object, and only one of its properties must be compared to determine if its active, for example: https://github.com/WordPress/gutenberg/blob/fe40df67ad88740d83c4c60735603d09a0ca5d55/packages/block-library/src/query/variations.js#L41-L59

If we allow dots as ["property accessors"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors) in `isActive` array items, we wouldn't need to resort to `isArray` functions in cases like the above. The example could instead be rewritten as

```js
isActive: [ 'namespace', 'query.postType' ]
```

ockham:
## What?
If a block variation's `isActive` property is a `string[]`, support "object paths" (i.e. dot ["property accessors"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors)), such as

```js
isActive: [ 'namespace', 'query.postType' ]
```

Fixes https://github.com/WordPress/gutenberg/issues/62068.

## Why?
This allows using the `string[]` form of the `isActive` property (rather than a `function`) for slightly more complex cases, such as https://github.com/WordPress/gutenberg/blob/fe40df67ad88740d83c4c60735603d09a0ca5d55/packages/block-library/src/query/variations.js#L41-L59

## How?
By using the `getValueFromObjectPath` util.

## Testing Instructions
Do some some testing with block variations in the editor.
Other than that, refer to automated tests.

### DIFF
--- a/packages/block-library/src/query/variations.js
+++ b/packages/block-library/src/query/variations.js
@@ -54,9 +54,7 @@ const variations = [
 			},
 		},
 		scope: [ 'inserter' ],
-		isActive: ( { namespace, query } ) => {
-			return namespace === 'core/posts-list' && query.postType === 'post';
-		},
+		isActive: [ 'namespace', 'query.postType' ],
 	},
 	{
 		name: 'title-date',
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -243,16 +243,30 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 			const blockType = getBlockType( state, blockName );
 			const attributeKeys = Object.keys( blockType?.attributes || {} );
 			const definedAttributes = variation.isActive.filter(
-				( attribute ) => attributeKeys.includes( attribute )
+				( attribute ) => {
+					// We support nested attribute paths, e.g. `layout.type`.
+					// In this case, we need to check if the part before the
+					// first dot is a known attribute.
+					const topLevelAttribute = attribute.split( '.' )[ 0 ];
+					return attributeKeys.includes( topLevelAttribute );
+				}
 			);
 			if ( definedAttributes.length === 0 ) {
 				return false;
 			}
-			return definedAttributes.every(
-				( attribute ) =>
-					attributes[ attribute ] ===
-					variation.attributes[ attribute ]
-			);
+			return definedAttributes.every( ( attribute ) => {
+				const attributeValue = getValueFromObjectPath(
+					attributes,
+					attribute
+				);
+				if ( attributeValue === undefined ) {
+					return false;
+				}
+				return (
+					attributeValue ===
+					getValueFromObjectPath( variation.attributes, attribute )
+				);
+			} );
 		}
 
 		return variation.isActive?.( attributes, variation.attributes );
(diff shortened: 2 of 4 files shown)


packages/blocks/src/store/selectors.js
```diff
@@ -243,16 +243,30 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
...
-				( attribute ) =>
-					attributes[ attribute ] ===
-					variation.attributes[ attribute ]
-			);
+			return definedAttributes.every( ( attribute ) => {
+				const attributeValue = getValueFromObjectPath(
+					attributes,
+					attribute
+				);
+				if ( attributeValue === undefined ) {
```
ntsekouras, line 262: Why do we need this check? Is it because of possible dot notated attribute that doesn't exist?

ntsekouras (approved):
This is a very nice enhancement, thank you! Probably it should be noted somewhere in the dev notes for 6.6.

packages/blocks/src/store/selectors.js
ockham: Yeah, exactly. While we're [verifying that the top-level attribute exists](https://github.com/WordPress/gutenberg/pull/62088/files#diff-7b33304a2d8a4d285862a54312849ae44dd6a8fc179721d444a42c83ec77fa6fR247-R250), we don't do that for any of the nested parts of the path. (I don't think that would be possible, as a `block.json` typically specifies `"type": "object"`, but not necessarily the shape of that object.)

`getValueFromObjectPath` returns `undefined` if an object path isn't found (and if no default value is provided). I couldn't think of a case where a nested attribute should be `undefined`, so I thought it we could assume that it meant that the path wasn't found 🤔

packages/block-library/src/query/variations.js
ockham, line 1: ❤️

PR #52969, opened 2023-07-26
Auto-inserting blocks: Add block inspector panel

ockham:
## What?
Follow-up to https://github.com/WordPress/gutenberg/pull/51449.

Add a new panel (tentatively named "Plugins") to the block inspector that lists blocks that are slated for auto-insertion next to that block, and provide toggles to insert (or remove) them.

Based on mockups by @jameskoster, following an idea by @mtias:

Screenshot 2023-07-19 at 17 03 44

Started during a pair-programming session with @c4rl0sbr4v0.

## Why?
While blocks are auto-inserted in a pretty straight-forward way into _unmodified_ templates, there's currently no visibility of them in modified templates. This PR seeks to provide a way to the user to easily insert them into their designated spots.

## How?

_Remaining TODO items will be tackled in follow-up PRs_

- [x] Expose `auto_insert` field via REST API.
- [x] List auto-inserted blocks for a given anchor block in a newly created block inspector panel.
- [x] Add toggles that enable/disable them. _The desired behavior is discussed in the comments below -- the most complete description is found [here](https://github.com/WordPress/gutenberg/pull/52969#issuecomment-1671056707)._
- [ ] Right-align toggles, add block icons.
- [ ] List auto-inserted blocks also when manually registered via `gutenberg_register_auto_inserted_block()` (rather than via `block.json`).
- [ ] Limit to Site Editor only
- [ ] [Better indicator what plugin is responsible for which block(s)](https://github.com/WordPress/gutenberg/pull/52969#issuecomment-1654043348) -- we're [currently](https://github.com/WordPress/gutenberg/pull/52969#issuecomment-1660296029) only using the block name's namespace.

## Testing Instructions
See video below.

## Screenshots or screencast

https://www.loom.com/share/82e50b1f30c641f9b9e470eb43ac7350?sid=687c394b-f2e5-47d9-a450-a1f2f51d8e96

PR #51162, opened 2023-06-01
Scripts: Optimize updating render paths when developing blocks

gziolo:
## What?

Follow-up for #50939.
Previously discussed in https://github.com/WordPress/gutenberg/pull/43917#discussion_r963821431.

The original issue was reported in 

> If you have `@wordpress/scripts` running with the `start` script, and add a `"render": "file:./render.php"` to your `block.json` file, the script doesn't detect it. But if you end the process and restart it, then it does.

The issue got fixed with https://github.com/WordPress/gutenberg/pull/50939, but the approach wasn't optimal in terms of performance.

## Why?

As correctly pointed out by @luisherranz in https://github.com/WordPress/gutenberg/pull/43917#discussion_r964473960 when developing the feature:

> Taking the function out of the filter was actually my suggestion to avoid reading/parsing all the block.json files per each PHP file of the project. That can slow things down for projects with multiple blocks and hundreds of PHP files (imagine something similar to WooCommerce).

While the fix works, we can improve the performance.

## How?

The proposed fix introduces a custom webpack plugin that uses the existing utility `getRenderPropPaths` to update the list of detected PHP files in the `render` field defined in `block.json` files. The idea behind it is to use the [`thisCompilation` webpack hook](https://webpack.js.org/api/compiler-hooks/#thiscompilation) that also `CopyWebpackPlugin` uses when copying files. This way, we can run the logic only once when changes to files on the disk are detected rather than whenever the `filter` callback gets executed for every PHP file living in the source folder.

## Testing Instructions

It can be tested with the testing instructions from https://github.com/WordPress/gutenberg/issues/49790.

I tested it myself by building the custom block using:

1. Open the branch from this PR.
2. Run `npx wp-create-block example-dynamic --no-wp-scripts --variant=dynamic` to create a custom dynamic block.
3. Go to the folder with the block with `cd example-dynamic`.
4. Run `../node_modules/.bin/wp-scripts build` to ensure that `src/render.php` file gets copied to the `build` folder:
Screenshot 2023-06-01 at 13 06 10

5. Remove `render` field from `src/block.json`.
6. Run `../node_modules/.bin/wp-scripts start` to begin development in the watch mode.
7. There should be no `render.php` file in the `build` folder.
8. Add back `render` field to the `src/block.json` file.
9. There should be `render.php` file in the `build` folder.

PR #43917, opened 2022-09-06
Scripts: Change webpack configuration to include `render` files in the build folder

SantosGuillamot:
## What?

Thanks to [this recent Pull Request](https://github.com/WordPress/gutenberg/pull/42430), now you can use a PHP file, defined in the `render` prop, instead of the traditional `render_callback` function. This PR aims to add that file automatically to the build folder, which isn't happening right now.

## Why?

As discussed [here](https://github.com/WordPress/block-hydration-experiments/pull/63#issuecomment-1237002023), right now it seems possible to do it manually, but it should work automatically.
 
## How?
We changed the WebPack configuration to detect and copy the `.php` file defined in the `block.json`:
* Added a new pattern in the `webpack.config.js` file to differentiate between JS and PHP files. In this second case, we are just filtering the PHP files to include only the one defined in the `render` prop, unless `process.env.WP_COPY_PHP_FILES_TO_DIST` is set to true.
* Add a function (`getRenderPropPaths`), to retrieve the file paths of all the `render` props defined in the blocks. This is used to compare in the `webpack.config.js` file.

Thanks @luisherranz for the help figuring out how this should work.

## Testing Instructions

Not sure what is the best way to test it in the Gutenberg plugin. I found this problem working on custom blocks in an external plugin (in the [BHE repo](https://github.com/WordPress/block-hydration-experiments/)), and I tested the code there. After adding this code, the `render.php` files were included in the build folder.

PR #42430, opened 2022-07-14
Blocks: Add new `render` field to `block.json`

gziolo:
## What?

A technical prototype for [Editor: Consider adding `render` property for block types](https://core.trac.wordpress.org/ticket/53148) in WordPress Trac. I will add tests and documentation only when people find this proposal useful.

Related to the proposal https://github.com/WordPress/gutenberg/pull/41289 from @ryanwelcher to add `--is-dynamic` flag to `@wordpress/create-block`.

## Why?

See comment from @spacedmonkey https://github.com/WordPress/gutenberg/pull/13693#discussion_r287337504.

> Current the render callback is a function call. How would it work as a file? Why not make it `renderTemplate` but still allow for `renderCallback`. Function calls, should be a method in a class or be namespaced, which wouldn't really work in another file.
>
> ACF blocks, has a rendercallback / render template in it's block definitions.

There is also another thread started by @spacedmonkey in https://github.com/WordPress/gutenberg/pull/13693#discussion_r288121797 which expands on this topic.

## How?

New `render` field in `block.json` file that accepts a string value.

The path to the file should be prefixed with `file:` and be relative to the `block.json` file in the build (target) folder. It's exactly the same pattern as used with scripts and styles in `block.json`.

PHP file that renders the template for the block can use 3 variables that are the same as `render_callback` takes as params:

- `$attributes` (`array`) - block attributes.
- `$content` (`string`) - block default content.
- `$block` (`WP_Block`) - block instance.

~Important note: I refactored the Archive block to use the new `render` field in `block.json` only for testing purposes. We rather won't do it as using function declarations is problematic with this approach because template files sometimes need to load several times for a given block type. When that happens, then we encounter a function redeclaration error.~ Removed from PR.

## Testing Instructions

Ensure that the temporarily refactored Archive blocks still work as before.

To test with custom blocks you need to add `render` field in the `block.json` file.

_Example:_

`src/block.json`

```json
{
    "name": "test/my-block",
    "render": "file:./render.php"
}
```

`src/render.php`

```php
<div <?php echo get_block_wrapper_attributes(); ?>>
    <?php echo $attributes['text']; ?>
</div>
```

PR #62092, opened 2024-05-29
block.json: Allow passing filename as `variations` field

ockham:
## What?
Modeled after https://core.trac.wordpress.org/changeset/54132/, which introduced the `render` field for `block.json`, as a way to point to a PHP file instead of providing a `render_callback`.

The main difference to the `variations` and `variation_callback` fields is that the `variations` field already existed prior to this PR, and it can be used to provide the static list of variations (i.e., an array). This PR makes it so that the field can be alternatively set to a string, which will be interpreted as the filename of the PHP file generating the variations.

Due to the changes to the `block.json` schema, it is recommended to view the diff [with whitespace changes hidden](https://github.com/WordPress/gutenberg/pull/62092/files?&w=1).

## Why?
See https://github.com/WordPress/wordpress-develop/pull/6668.

## How?
In the GB compat layer: By using the `block_type_metadata_settings` filter.

## Testing Instructions

Apply the following patch, and create `build/block-library/blocks/template-part/variations.php` with the content below.

<details>
<summary>
Patch
</summary>

```diff
diff --git a/packages/block-library/src/template-part/block.json b/packages/block-library/src/template-part/block.json
index 9710bdeee2e..3493946aa7a 100644
--- a/packages/block-library/src/template-part/block.json
+++ b/packages/block-library/src/template-part/block.json
@@ -29,5 +29,6 @@
 			"clientNavigation": true
 		}
 	},
+	"variations": "variations.php",
 	"editorStyle": "wp-block-template-part-editor"
 }
diff --git a/packages/block-library/src/template-part/index.php b/packages/block-library/src/template-part/index.php
index be867c4ced1..58420b999b2 100644
--- a/packages/block-library/src/template-part/index.php
+++ b/packages/block-library/src/template-part/index.php
@@ -175,115 +175,6 @@ function render_block_core_template_part( $attributes ) {
 	return "<$html_tag $wrapper_attributes>" . str_replace( ']]>', ']]&gt;', $content ) . "</$html_tag>";
 }
 
-/**
- * Returns an array of area variation objects for the template part block.
- *
- * @since 6.1.0
- *
- * @param array $instance_variations The variations for instances.
- *
- * @return array Array containing the block variation objects.
- */
-function build_template_part_block_area_variations( $instance_variations ) {
-	$variations    = array();
-	$defined_areas = get_allowed_block_template_part_areas();
-
-	foreach ( $defined_areas as $area ) {
-		if ( 'uncategorized' !== $area['area'] ) {
-			$has_instance_for_area = false;
-			foreach ( $instance_variations as $variation ) {
-				if ( $variation['attributes']['area'] === $area['area'] ) {
-					$has_instance_for_area = true;
-					break;
-				}
-			}
-
-			$scope = $has_instance_for_area ? array() : array( 'inserter' );
-
-			$variations[] = array(
-				'name'        => 'area_' . $area['area'],
-				'title'       => $area['label'],
-				'description' => $area['description'],
-				'attributes'  => array(
-					'area' => $area['area'],
-				),
-				'scope'       => $scope,
-				'icon'        => $area['icon'],
-			);
-		}
-	}
-	return $variations;
-}
-
-/**
- * Returns an array of instance variation objects for the template part block
- *
- * @since 6.1.0
- *
- * @return array Array containing the block variation objects.
- */
-function build_template_part_block_instance_variations() {
-	// Block themes are unavailable during installation.
-	if ( wp_installing() ) {
-		return array();
-	}
-
-	if ( ! current_theme_supports( 'block-templates' ) && ! current_theme_supports( 'block-template-parts' ) ) {
-		return array();
-	}
-
-	$variations     = array();
-	$template_parts = get_block_templates(
-		array(
-			'post_type' => 'wp_template_part',
-		),
-		'wp_template_part'
-	);
-
-	$defined_areas = get_allowed_block_template_part_areas();
-	$icon_by_area  = array_combine( array_column( $defined_areas, 'area' ), array_column( $defined_areas, 'icon' ) );
-
-	foreach ( $template_parts as $template_part ) {
-		$variations[] = array(
-			'name'        => 'instance_' . sanitize_title( $template_part->slug ),
-			'title'       => $template_part->title,
-			// If there's no description for the template part don't show the
-			// block description. This is a bit hacky, but prevent the fallback
-			// by using a non-breaking space so that the value of description
-			// isn't falsey.
-			'description' => $template_part->description || '&nbsp;',
-			'attributes'  => array(
-				'slug'  => $template_part->slug,
-				'theme' => $template_part->theme,
-				'area'  => $template_part->area,
-			),
-			'scope'       => array( 'inserter' ),
-			'icon'        => isset( $icon_by_area[ $template_part->area ] ) ? $icon_by_area[ $template_part->area ] : null,
-			'example'     => array(
-				'attributes' => array(
-					'slug'  => $template_part->slug,
-					'theme' => $template_part->theme,
-					'area'  => $template_part->area,
-				),
-			),
-		);
-	}
-	return $variations;
-}
-
-/**
- * Returns an array of all template part block variations.
- *
- * @since 5.9.0
- *
- * @return array Array containing the block variation objects.
- */
-function build_template_part_block_variations() {
-	$instance_variations = build_template_part_block_instance_variations();
-	$area_variations     = build_template_part_block_area_variations( $instance_variations );
-	return array_merge( $area_variations, $instance_variations );
-}
-
 /**
  * Registers the `core/template-part` block on the server.
  *
@@ -294,7 +185,6 @@ function register_block_core_template_part() {
 		__DIR__ . '/template-part',
 		array(
 			'render_callback'    => 'render_block_core_template_part',
-			'variation_callback' => 'build_template_part_block_variations',
 		)
 	);
 }
diff --git a/packages/block-library/src/template-part/variations.php b/packages/block-library/src/template-part/variations.php
new file mode 100644
index 00000000000..c47f0ffc127
--- /dev/null
+++ b/packages/block-library/src/template-part/variations.php
@@ -0,0 +1,112 @@
+<?php
+
+/**
+ * Returns an array of area variation objects for the template part block.
+ *
+ * @since 6.1.0
+ *
+ * @param array $instance_variations The variations for instances.
+ *
+ * @return array Array containing the block variation objects.
+ */
+function build_template_part_block_area_variations( $instance_variations ) {
+	$variations    = array();
+	$defined_areas = get_allowed_block_template_part_areas();
+
+	foreach ( $defined_areas as $area ) {
+		if ( 'uncategorized' !== $area['area'] ) {
+			$has_instance_for_area = false;
+			foreach ( $instance_variations as $variation ) {
+				if ( $variation['attributes']['area'] === $area['area'] ) {
+					$has_instance_for_area = true;
+					break;
+				}
+			}
+
+			$scope = $has_instance_for_area ? array() : array( 'inserter' );
+
+			$variations[] = array(
+				'name'        => 'area_' . $area['area'],
+				'title'       => $area['label'],
+				'description' => $area['description'],
+				'attributes'  => array(
+					'area' => $area['area'],
+				),
+				'scope'       => $scope,
+				'icon'        => $area['icon'],
+			);
+		}
+	}
+	return $variations;
+}
+
+/**
+ * Returns an array of instance variation objects for the template part block
+ *
+ * @since 6.1.0
+ *
+ * @return array Array containing the block variation objects.
+ */
+function build_template_part_block_instance_variations() {
+	// Block themes are unavailable during installation.
+	if ( wp_installing() ) {
+		return array();
+	}
+
+	if ( ! current_theme_supports( 'block-templates' ) && ! current_theme_supports( 'block-template-parts' ) ) {
+		return array();
+	}
+
+	$variations     = array();
+	$template_parts = get_block_templates(
+		array(
+			'post_type' => 'wp_template_part',
+		),
+		'wp_template_part'
+	);
+
+	$defined_areas = get_allowed_block_template_part_areas();
+	$icon_by_area  = array_combine( array_column( $defined_areas, 'area' ), array_column( $defined_areas, 'icon' ) );
+
+	foreach ( $template_parts as $template_part ) {
+		$variations[] = array(
+			'name'        => 'instance_' . sanitize_title( $template_part->slug ),
+			'title'       => $template_part->title,
+			// If there's no description for the template part don't show the
+			// block description. This is a bit hacky, but prevent the fallback
+			// by using a non-breaking space so that the value of description
+			// isn't falsey.
+			'description' => $template_part->description || '&nbsp;',
+			'attributes'  => array(
+				'slug'  => $template_part->slug,
+				'theme' => $template_part->theme,
+				'area'  => $template_part->area,
+			),
+			'scope'       => array( 'inserter' ),
+			'icon'        => isset( $icon_by_area[ $template_part->area ] ) ? $icon_by_area[ $template_part->area ] : null,
+			'example'     => array(
+				'attributes' => array(
+					'slug'  => $template_part->slug,
+					'theme' => $template_part->theme,
+					'area'  => $template_part->area,
+				),
+			),
+		);
+	}
+	return $variations;
+}
+
+/**
+ * Returns an array of all template part block variations.
+ *
+ * @since 5.9.0
+ *
+ * @return array Array containing the block variation objects.
+ */
+function build_template_part_block_variations() {
+	$instance_variations = build_template_part_block_instance_variations();
+	$area_variations     = build_template_part_block_area_variations( $instance_variations );
+	return array_merge( $area_variations, $instance_variations );
+}
+
+return build_template_part_block_variations();

```
</details>

<details>
<summary>
build/block-library/blocks/template-part/variations.php
</summary>

```php
<?php

/**
 * Returns an array of area variation objects for the template part block.
 *
 * @since 6.1.0
 *
 * @param array $instance_variations The variations for instances.
 *
 * @return array Array containing the block variation objects.
 */
function gutenberg_build_template_part_block_area_variations( $instance_variations ) {
	$variations    = array();
	$defined_areas = get_allowed_block_template_part_areas();

	foreach ( $defined_areas as $area ) {
		if ( 'uncategorized' !== $area['area'] ) {
			$has_instance_for_area = false;
			foreach ( $instance_variations as $variation ) {
				if ( $variation['attributes']['area'] === $area['area'] ) {
					$has_instance_for_area = true;
					break;
				}
			}

			$scope = $has_instance_for_area ? array() : array( 'inserter' );

			$variations[] = array(
				'name'        => 'area_' . $area['area'],
				'title'       => $area['label'],
				'description' => $area['description'],
				'attributes'  => array(
					'area' => $area['area'],
				),
				'scope'       => $scope,
				'icon'        => $area['icon'],
			);
		}
	}
	return $variations;
}

/**
 * Returns an array of instance variation objects for the template part block
 *
 * @since 6.1.0
 *
 * @return array Array containing the block variation objects.
 */
function gutenberg_build_template_part_block_instance_variations() {
	// Block themes are unavailable during installation.
	if ( wp_installing() ) {
		return array();
	}

	if ( ! current_theme_supports( 'block-templates' ) && ! current_theme_supports( 'block-template-parts' ) ) {
		return array();
	}

	$variations     = array();
	$template_parts = get_block_templates(
		array(
			'post_type' => 'wp_template_part',
		),
		'wp_template_part'
	);

	$defined_areas = get_allowed_block_template_part_areas();
	$icon_by_area  = array_combine( array_column( $defined_areas, 'area' ), array_column( $defined_areas, 'icon' ) );

	foreach ( $template_parts as $template_part ) {
		$variations[] = array(
			'name'        => 'instance_' . sanitize_title( $template_part->slug ),
			'title'       => $template_part->title,
			// If there's no description for the template part don't show the
			// block description. This is a bit hacky, but prevent the fallback
			// by using a non-breaking space so that the value of description
			// isn't falsey.
			'description' => $template_part->description || '&nbsp;',
			'attributes'  => array(
				'slug'  => $template_part->slug,
				'theme' => $template_part->theme,
				'area'  => $template_part->area,
			),
			'scope'       => array( 'inserter' ),
			'icon'        => isset( $icon_by_area[ $template_part->area ] ) ? $icon_by_area[ $template_part->area ] : null,
			'example'     => array(
				'attributes' => array(
					'slug'  => $template_part->slug,
					'theme' => $template_part->theme,
					'area'  => $template_part->area,
				),
			),
		);
	}
	return $variations;
}

/**
 * Returns an array of all template part block variations.
 *
 * @since 5.9.0
 *
 * @return array Array containing the block variation objects.
 */
function gutenberg_build_template_part_block_variations() {
	$instance_variations = gutenberg_build_template_part_block_instance_variations();
	$area_variations     = gutenberg_build_template_part_block_area_variations( $instance_variations );
	return array_merge( $area_variations, $instance_variations );
}

return gutenberg_build_template_part_block_variations();

```
</details>

(The latter is required since Gutenberg isn’t currently copying files specified in a `block.json`’s `render` or `variations` field to the `build/` directory. Refer to https://github.com/WordPress/gutenberg/issues/63077 for further information.)

Then, in the Site Editor, verify that instances of the block still work, and that their variations are detected correctly (e.g. a header template part is recognized as such in the block inspector).

**Specifically, open the inserter and enter "Header" into the search input field. _Both_ the Template Part block _and_ its Header variation should be shown.**

image

### DIFF
--- a/docs/reference-guides/block-api/block-metadata.md
+++ b/docs/reference-guides/block-api/block-metadata.md
@@ -427,7 +427,7 @@ See the [Example documentation](/docs/reference-guides/block-api/block-registrat
 
 ### Variations
 
--   Type: `object[]`
+-   Type: `object[]|string`
 -   Optional
 -   Localized: Yes (`title`, `description`, and `keywords` of each variation only)
 -   Property: `variations`
@@ -453,6 +453,14 @@ See the [Example documentation](/docs/reference-guides/block-api/block-registrat
 Block Variations is the API that allows a block to have similar versions of it, but all these versions share some common functionality. Each block variation is differentiated from the others by setting some initial attributes or inner blocks. Then at the time when a block is inserted these attributes and/or inner blocks are applied.
 
 _Note: In JavaScript you can provide a function for the `isActive` property, and a React element for the `icon`. In the `block.json` file both only support strings_
+
+_Note: Starting with version 6.7, it is possible to specify a PHP file that generates the list of block variations on the server side:_
+
+```json
+{
+	"variations": "file:./variations.php"
+}
+```
 
 See [the variations documentation](/docs/reference-guides/block-api/block-variations.md) for more details.
 
--- a/docs/reference-guides/block-api/block-metadata.md
+++ b/docs/reference-guides/block-api/block-metadata.md
@@ -427,7 +427,7 @@ See the [Example documentation](/docs/reference-guides/block-api/block-registrat
 
 ### Variations
 
--   Type: `object[]|string`
+-   Type: `object[]|WPDefinedPath` ([learn more](#wpdefinedpath))
 -   Optional
 -   Localized: Yes (`title`, `description`, and `keywords` of each variation only)
 -   Property: `variations`
@@ -454,12 +454,39 @@ Block Variations is the API that allows a block to have similar versions of it, 
 
 _Note: In JavaScript you can provide a function for the `isActive` property, and a React element for the `icon`. In the `block.json` file both only support strings_
 
-_Note: Starting with version 6.7, it is possible to specify a PHP file that generates the list of block variations on the server side:_
+Starting with version 6.7, it is possible to specify a PHP file in `block.json` that generates the list of block variations on the server side:
 
 ```json
-{
-	"variations": "file:./variations.php"
-}
+{ "variations": "file:./variations.php" }
+```
+
+That PHP file is expected to `return` an array that contains the block variations. For example:
+
+```php
+<?php
+
+return array(
+	array(
+		'isDefault'  => true,
+		'name'       => 'wordpress',
+		'title'      => 'WordPress',
+		'icon'       => 'wordpress',
+		'attributes' => array(
+			'service' => 'wordpress',
+		),
+		'isActive'   => array( 'service' )
+	),
+	array(
+		'name'       => 'gravatar',
+		'title'      => 'Gravatar',
+		'icon'       => 'commentAuthorAvatar',
+		'attributes' => array(
+			'service' => 'gravatar',
+		),
+		'isActive'   => array( 'service' )
+	),
+);
+
 ```
 
 See [the variations documentation](/docs/reference-guides/block-api/block-variations.md) for more details.
--- a/docs/reference-guides/block-api/block-metadata.md
+++ b/docs/reference-guides/block-api/block-metadata.md
@@ -466,6 +466,7 @@ For example:
 
 ```php
 <?php
+// Generate variations for a Social Icon kind of block.
 
 return array(
 	array(
--- a/docs/reference-guides/block-api/block-metadata.md
+++ b/docs/reference-guides/block-api/block-metadata.md
@@ -477,13 +477,17 @@ return array(
 		'isActive'   => array( 'service' )
 	),
 	array(
-		'name'       => 'gravatar',
-		'title'      => 'Gravatar',
-		'icon'       => 'commentAuthorAvatar',
-		'attributes' => array(
-			'service' => 'gravatar',
+		'name'       => 'mail',
+		'title'      => __( 'Mail' ),
+		'keywords'   => array(
+			__( 'email' ),
+			__( 'e-mail' )
 		),
-		'isActive'   => array( 'service' )
+		'icon'       => 'mail',
+		'attributes' => array(
+			'service' => 'mail',
+		),
+		'isActive'   => array( 'mail' )
 	),
 );
 
--- a/docs/reference-guides/block-api/block-metadata.md
+++ b/docs/reference-guides/block-api/block-metadata.md
@@ -460,7 +460,9 @@ Starting with version 6.7, it is possible to specify a PHP file in `block.json` 
 { "variations": "file:./variations.php" }
 ```
 
-That PHP file is expected to `return` an array that contains the block variations. For example:
+That PHP file is expected to `return` an array that contains the block variations. Strings found in the variations returned from the PHP file will not be localized automatically; instead, use the `__()` function as usual.
+
+For example:
 
 ```php
 <?php
--- a/docs/reference-guides/block-api/block-registration.md
+++ b/docs/reference-guides/block-api/block-registration.md
@@ -233,10 +233,12 @@ example: {
 
 #### variations (optional)
 
--   **Type:** `Object[]`
+-   **Type:** `Object[]|string`
 -   **Since**: `WordPress 5.9.0`
 
 Similarly to how the block's styles can be declared, a block type can define block variations that the user can pick from. The difference is that, rather than changing only the visual appearance, this field provides a way to apply initial custom attributes and inner blocks at the time when a block is inserted. See the [Block Variations API](/docs/reference-guides/block-api/block-variations.md) for more details.
+
+_Note:_ Starting with WordPress 6.7, it is possible to specify a PHP file that generates the list of block variations on the server side.
 
 #### supports (optional)
 
--- a/docs/reference-guides/block-api/block-registration.md
+++ b/docs/reference-guides/block-api/block-registration.md
@@ -233,12 +233,10 @@ example: {
 
 #### variations (optional)
 
--   **Type:** `Object[]|string`
+-   **Type:** `Object[]`
 -   **Since**: `WordPress 5.9.0`
 
 Similarly to how the block's styles can be declared, a block type can define block variations that the user can pick from. The difference is that, rather than changing only the visual appearance, this field provides a way to apply initial custom attributes and inner blocks at the time when a block is inserted. See the [Block Variations API](/docs/reference-guides/block-api/block-variations.md) for more details.
-
-_Note:_ Starting with WordPress 6.7, it is possible to specify a PHP file that generates the list of block variations on the server side.
 
 #### supports (optional)
 
--- a/lib/compat/wordpress-6.7/blocks.php
+++ b/lib/compat/wordpress-6.7/blocks.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * Temporary compatibility shims for block APIs present in Gutenberg.
+ *
+ * @package gutenberg
+ */
+
+/**
+ * Allow passing a PHP file as `variations` field for Core versions < 6.7
+ *
+ * @param array $settings Array of determined settings for registering a block type.
+ * @param array $metadata Metadata provided for registering a block type.
+ * @return array The block type settings
+ */
+function gutenberg_filter_block_type_metadata_settings_allow_variations_php_file( $settings, $metadata ) {
+	// If `variations` is a string, it's the name of a PHP file that
+	// generates the variations.
+	if ( ! empty( $settings['variations'] ) && is_string( $settings['variations'] ) ) {
+		$variations_path = wp_normalize_path(
+			realpath(
+				dirname( $metadata['file'] ) . '/' .
+				remove_block_asset_path_prefix( $metadata['variations'] )
+			)
+		);
+		if ( $variations_path ) {
+			/**
+			 * Generates the list of block variations.
+			 *
+			 * @since 6.7.0
+			 *
+			 * @return string Returns the list of block variations.
+			 */
+			$settings['variation_callback'] = static function () use ( $variations_path ) {
+				$variations = require $variations_path;
+				return $variations;
+			};
+			// The block instance's `variations` field is only allowed to be an array
+			// (of known block variations). We unset it so that the block instance will
+			// provide a getter that returns the result of the `variation_callback` instead.
+			unset( $settings['variations'] );
+		}
+	}
+	return $settings;
+}
+add_filter( 'block_type_metadata_settings', 'gutenberg_filter_block_type_metadata_settings_allow_variations_php_file', 10, 2 );
--- a/lib/compat/wordpress-6.7/blocks.php
+++ b/lib/compat/wordpress-6.7/blocks.php
@@ -19,7 +19,7 @@ function gutenberg_filter_block_type_metadata_settings_allow_variations_php_file
 		$variations_path = wp_normalize_path(
 			realpath(
 				dirname( $metadata['file'] ) . '/' .
-				remove_block_asset_path_prefix( $metadata['variations'] )
+				remove_block_asset_path_prefix( $settings['variations'] )
 			)
 		);
 		if ( $variations_path ) {
--- a/schemas/json/block.json
+++ b/schemas/json/block.json
@@ -897,91 +897,98 @@
 			]
 		},
 		"variations": {
-			"type": "array",
 			"description": "Block Variations is the API that allows a block to have similar versions of it, but all these versions share some common functionality.",
-			"items": {
-				"type": "object",
-				"required": [ "name", "title" ],
-				"additionalProperties": false,
-				"properties": {
-					"name": {
-						"type": "string",
-						"description": "The unique and machine-readable name."
-					},
-					"title": {
-						"type": "string",
-						"description": "A human-readable variation title."
-					},
-					"description": {
-						"type": "string",
-						"description": "A detailed variation description."
-					},
-					"category": {
-						"description": "A category classification, used in search interfaces to arrange block types by category.",
-						"anyOf": [
-							{
+			"oneOf": [
+				{
+					"type": "string"
+				},
+				{
+					"type": "array",
+					"items": {
+						"type": "object",
+						"required": [ "name", "title" ],
+						"additionalProperties": false,
+						"properties": {
+							"name": {
+								"type": "string",
+								"description": "The unique and machine-readable name."
+							},
+							"title": {
+								"type": "string",
+								"description": "A human-readable variation title."
+							},
+							"description": {
+								"type": "string",
+								"description": "A detailed variation description."
+							},
+							"category": {
+								"description": "A category classification, used in search interfaces to arrange block types by category.",
+								"anyOf": [
+									{
+										"type": "string"
+									},
+									{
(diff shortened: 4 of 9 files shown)


ockham:
I've added some testing instructions and noticed that isn't quite working yet. I've updated the PR description to reflect that:

> Apply the following patch [...] Then, in the Site Editor, [...]
> **Specifically, open the inserter and enter "Header" into the search input field. _Both_ the Template Part block _and_ its Header variation should be shown.**
> ⚠️ _This is currently not working; need to investigate!_
>
> image

AFAICT from the page source, the problem is that the bootstrapped block type has `variations` set to the filename (`file:./variations.php`). While we're correctly ignoring the string value on the client side, this shouldn't be happening in the first place; instead, the server should apply the logic from `gutenberg_filter_block_type_metadata_settings_allow_variations_php_file` and set the `variations` field to the variations generated by that file.

This is probably due to [an issue I've seen before](https://github.com/WordPress/gutenberg/pull/52969/files#diff-b6a5f5374aa2b01e4369049937355b47072d9123e8918523bc49b46efa501807R169) where filters aren't applied properly (?) when preloading/bootstrapping 😕

ockham:
I dug a bit deeper, and it seems like the problem on the server side is due to [`realpath`](https://www.php.net/manual/de/function.realpath.php) returning `false` when [attempting to resolve the location of the `variations.php` file](https://github.com/WordPress/gutenberg/pull/62092/files#diff-d29ffedb5c2a767cd8a9ae8f309588066eef6489619561bbcc587bdbe987eb59R19-R24).

I've checked the corresponding directory, which is the Template Part block's subfolder of the `build/` directory (`build/block-library/blocks/template-part/`), and there really isn't any `variations.php` file in that folder. Maybe that shouldn't surprise me, as we don't have a build step anywhere that would copy that file over from the `src/` directory 🤔 

It made me wonder however how we're handling `render.php` files. I checked the PR that introduced the `render` field, https://github.com/WordPress/gutenberg/pull/42430, and I noticed there isn't any build step either. Furthermore, it turns out that none of our Core blocks is using the `render` field! (Only the blocks in our e2e tests package are.)

I wonder if that means that using the `render` field in a Core block to specify the file that's supposed to render it would also break 😬 Or maybe I'm missing something. cc/ @gziolo 

Anyway, if I'm correct, then the fix would be to include a build step that copies those PHP files over to the build dir. We're [already doing that for blocks' `index.php` files](https://github.com/WordPress/gutenberg/blob/90bc142277c0eee8a1d4d17a380dcccd8e754db2/tools/webpack/blocks.js#L121-L220); what's slightly trickier here is that we'll need to determine the filename from the `variations` field (and the `render` field, respectively, if that's also broken).

_Edit:_ During that build step, we'll also need to prefix any functions inside of the `variations.php`s with `gutenberg_` (much like we do for the `index.php`s).

docs/reference-guides/block-api/block-registration.md
```diff
@@ -233,11 +233,13 @@ example: {
 
 #### variations (optional)
 
--   **Type:** `Object[]`
+-   **Type:** `Object[]|string`
 -   **Since**: `WordPress 5.9.0`
 
 Similarly to how the block's styles can be declared, a block type can define block variations that the user can pick from. The difference is that, rather than changing only the visual appearance, this field provides a way to apply initial custom attributes and inner blocks at the time when a block is inserted. See the [Block Variations API](/docs/reference-guides/block-api/block-variations.md) for more details.
 
+_Note:_ Starting with WordPress 6.7, it is possible to specify a PHP file that generates the list of block variations on the server side.
```
gziolo: This document covers block registration on the client, so that string won't load the PHP file in that context.

docs/reference-guides/block-api/block-metadata.md
```diff
@@ -454,6 +454,14 @@ Block Variations is the API that allows a block to have similar versions of it,
 
 _Note: In JavaScript you can provide a function for the `isActive` property, and a React element for the `icon`. In the `block.json` file both only support strings_
 
+_Note: Starting with version 6.7, it is possible to specify a PHP file that generates the list of block variations on the server side:_
+
+```json
+{
+	"variations": "file:./variations.php"
```
gziolo: Is there also documentation explaining how to shape this PHP file?

lib/compat/wordpress-6.7/blocks.php
```diff
@@ -0,0 +1,45 @@
...
+ * @return array The block type settings
+ */
+function gutenberg_filter_block_type_metadata_settings_allow_variations_php_file( $settings, $metadata ) {
+	// If `variations` is a string, it's the name of a PHP file that
+	// generates the variations.
+	if ( ! empty( $settings['variations'] ) && is_string( $settings['variations'] ) ) {
+		$variations_path = wp_normalize_path(
+			realpath(
+				dirname( $metadata['file'] ) . '/' .
+				remove_block_asset_path_prefix( $metadata['variations'] )
```
gziolo: One thing is puzzling here to me. The condition checks the processed `settings` array, but the path gets computed using the `metadata` loaded from the `block.json` file. Why not use settings, too?

```suggestion
				remove_block_asset_path_prefix( $settings['variations'] )
```

gziolo (commented):
> Maybe that shouldn't surprise me, as we don't have a build step anywhere that would copy that file over from the src/ directory 🤔

Yes, it isn't handled in the Gutenberg repository. The same issue most likely would exist if someone would try to use `"render": "file:./render.php"`.

> I wonder if that means that using the render field in a Core block to specify the file that's supposed to render it would also break 😬

It would require updating the webpack config which shouldn't be that much work but it wasn't necessary so far.

> During that build step, we'll also need to prefix any functions inside of the variations.phps with gutenberg_ (much like we do for the index.phps).

I think that's already happening when copying existing `index.php` files for core blocks. So that would get extended to cover the linked files through `variations` and `render` fields. I don't think it's a blocker for this PR.

At the same time, I think it would be worth adding support for variations file in `@wordpress/scripts` mirroring what we have for `render` path:

https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/scripts/config/webpack.config.js#L324

https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/scripts/config/webpack.config.js#L52-L57

https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/scripts/config/webpack.config.js#L367-L379

It probably would require making the existing custom plugin general purpose tool for discovering linked PHP files.

ockham:
@gziolo Thank you for the pointers! I dug a bit and found that https://github.com/WordPress/gutenberg/pull/43917 actually did add support for copying `render` files to the build folder, and https://github.com/WordPress/gutenberg/pull/51162 then further optimized that build step. I'll see if I can apply the same technique to the `variations` field.

Edit: To clarify, this only applies to third-party blocks, but not to Core blocks (which use a different build system).

ockham:
I've filed https://github.com/WordPress/gutenberg/issues/63077 to document that `render` files aren't copied to the `build/` directory for Core blocks.

docs/reference-guides/block-api/block-registration.md
ockham: Good point. I think I'll remove the note altogether (and revert the `Type` to just `Object[]`), as its otherwise more confusing. (There's also no documentation about the server-side only `render` field here, so it's probably fine not to mention the server-side-only option for `variations` either.)

ockham: 56f6628b583f6d4e4aebbfb6d1120f559dab8674

docs/reference-guides/block-api/block-metadata.md
ockham: Not yet! Good catch. I'll add something.

ockham: 01c8c8d718d

lib/compat/wordpress-6.7/blocks.php
ockham: That's a fair point. I copied it over from https://github.com/WordPress/wordpress-develop/pull/6668 where I'm consistently using `$metadata` to denote the "input" and `$settings` for the output, in keeping with the structure of `register_block_type_from_metadata()`.

I missed this instance of when changing `$metadata['variations']` to `$settings['variations']` in the filter.

ockham:
I've started work on including `variations` files in the build (for third-party blocks): https://github.com/WordPress/gutenberg/pull/63098

ockham:
Tests are apparently failing because of https://github.com/WordPress/gutenberg/issues/63073. I've filed https://github.com/WordPress/gutenberg/pull/63099 to fix it, we'll need to rebase this PR once that's merged.

schemas/json/block.json
```diff
@@ -897,91 +897,98 @@
...
-						"description": "A detailed variation description."
-					},
-					"category": {
-						"description": "A category classification, used in search interfaces to arrange block types by category.",
-						"anyOf": [
-							{
-								"type": "string"
+			"oneOf": [
+				{
+					"type": "string"
```
tjcafferkey: Do you think it is worth adding a `description` property here to better document the expected usage of this?

ockham: Yeah, that's a good point. I thought it might collide with the `description` a couple lines above (i.e. which applies to both `oneOf` options), but it seems to be allowed by the JSON schema spec, so I'll add it.

gziolo (approved):
> I've started work on including variations files in the build (for third-party blocks): https://github.com/WordPress/gutenberg/pull/63098

Ok, that work for me as long it's covered 👍🏻 

I don't have more comments on my side, but I see some pending feedback to address from @tjcafferkey before merging.

schemas/json/block.json
ockham: be8a9c2ba35

tjcafferkey (approved):

ockham:
Thank you very much @gziolo and @tjcafferkey!

fabiankaegy:
Hey @ockham  👋 

Thanks in advance :)

ockham:
Hey @fabiankaegy, sure thing! Please find my suggested Dev note for this enhancement [here](https://github.com/WordPress/gutenberg/issues/65784#issuecomment-2396921320).

PR #62095, opened 2024-05-29
Tests: Change how directives processing gets disabled

gziolo:
## What?

The necessary steps for the planned changes in WordPress core are as follows: https://github.com/WordPress/wordpress-develop/pull/6331.

## Why?

The existing filter gets replaced with more thigh integration with the `WP_Block` class.

## How?

New filter `interactivity_process_directives` gets introduced to allow controlling directives processing for Interactivity API.

## Testing Instructions

E2E tests should still pass after changes land in WordPress core. It will fail currently.

### DIFF
--- a/packages/e2e-tests/plugins/interactive-blocks.php
+++ b/packages/e2e-tests/plugins/interactive-blocks.php
@@ -29,11 +29,7 @@ add_action(
 			// Ensure the interactivity API is loaded.
 			wp_interactivity();
 			// But remove the server directive processing.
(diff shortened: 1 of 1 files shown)


cbravobernal (approved):
LGTM

gziolo:
WP core changes landed so we need it now: https://github.com/WordPress/wordpress-develop/commit/4b91a597b7a91d8ce099314a744fca5309dd0134.

PR #62103, opened 2024-05-29
Editor: Remove useless props from InserterSidebar component

youknowriad:
Related https://github.com/WordPress/gutenberg/issues/52632

Just a tiny non-impactful PR that removes some props being passed down to the InserterSidebar component. These props can now be selected directly in the component.

There is no functional change here.


ntsekouras (approved):
LGTM, thanks!

PR #62108, opened 2024-05-29
MediaUpload: Keep track of the dialog state

### LINKED ISSUE #62105: The Inline Image format opens two media modals in development mode
The "Inline Image" format opens two media modals. It calls the modal open method directly in the MediaUpload.render prop. It is a hack to emulate modal opening on a click, but React will call the class component methods like render [twice in development](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-double-rendering-in-development). See: #61943.

## Affected e2e tests

* `editor/various/adding-inline-tokens.spec.js` - should insert inline image.

## Steps to reproduce

* Checkout branch from #61943.
* Enable `SCRIPT_DEBUG` for non `wp-env` setups.
* Create a post.
* Add a paragraph block.
* Try adding an inline image.

Mamaduka:
## What?
Fixes #62105.
Related #61943.

PR updates `MediaUpload` to keep track of the Media Library dialog state. In rare cases, when the `open` method is called multiple times, it prevents several dialogs from opening.

## Why?
The "Inline Image" format opens the Media Library dialog by calling the `open` method during the render. It is a hack to emulate modal opening on a click, but React will call the class component methods like render [twice in development](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-double-rendering-in-development).

## Testing Instructions
1. Enable development mode for the post editor.
2. Create a post.
3. Add a Paragraph block.
4. Try adding the inline image.
5. Confirm that only one dialog is open.

<p>

```diff
diff --git .wp-env.json .wp-env.json
index 20d5597e54b..57d4e1a72d8 100644
--- .wp-env.json
+++ .wp-env.json
@@ -4,6 +4,9 @@
 	"themes": [ "./test/emptytheme" ],
 	"env": {
 		"tests": {
+			"config": {
+				"SCRIPT_DEBUG": true
+			},
 			"mappings": {
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",
diff --git packages/edit-post/src/index.js packages/edit-post/src/index.js
index 1e0b3fe7d4d..d80d3c161a6 100644
--- packages/edit-post/src/index.js
+++ packages/edit-post/src/index.js
@@ -7,7 +7,7 @@ import {
 	__experimentalRegisterExperimentalCoreBlocks,
 } from '@wordpress/block-library';
 import deprecated from '@wordpress/deprecated';
-import { createRoot } from '@wordpress/element';
+import { createRoot, StrictMode } from '@wordpress/element';
 import { dispatch, select } from '@wordpress/data';
 import { store as preferencesStore } from '@wordpress/preferences';
 import {
@@ -131,12 +131,14 @@ export function initializeEditor(
 	window.addEventListener( 'drop', ( e ) => e.preventDefault(), false );
 
 	root.render(
-		<Editor
-			settings={ settings }
-			postId={ postId }
-			postType={ postType }
-			initialEdits={ initialEdits }
-		/>
+		<StrictMode>
+			<Editor
+				settings={ settings }
+				postId={ postId }
+				postType={ postType }
+				initialEdits={ initialEdits }
+			/>
+		</StrictMode>
 	);
 
 	return root;
``` 

</p>
</details> 

Same.

### DIFF
--- a/packages/media-utils/src/components/media-upload/index.js
+++ b/packages/media-utils/src/components/media-upload/index.js
@@ -396,10 +396,11 @@ class MediaUpload extends Component {
 
 	onClose() {
 		const { onClose } = this.props;
-
 		if ( onClose ) {
 			onClose();
 		}
+
+		this.isModalOpen = false;
 	}
 
 	updateCollection() {
@@ -421,6 +422,10 @@ class MediaUpload extends Component {
 	}
 
 	openModal() {
+		if ( this.isModalOpen ) {
+			return;
+		}
+
 		const {
 			allowedTypes,
 			gallery = false,
@@ -454,6 +459,7 @@ class MediaUpload extends Component {
 		}
 		this.initializeListeners();
 		this.frame.open();
+		this.isModalOpen = true;
 	}
 
 	render() {


packages/media-utils/src/components/media-upload/index.js
```diff
@@ -396,10 +396,11 @@ class MediaUpload extends Component {
 
 	onClose() {
 		const { onClose } = this.props;
-
 		if ( onClose ) {
 			onClose();
 		}
+
+		this.isModalOpen = false;
```
Mamaduka, line 403: I'm no Media Library code expert, but I think we should call `this.frame.detach()` on close action. Otherwise, the modal leaves some "breadcrumbs" in the DOM. Core does a similar thing for [classic widgets](https://github.com/WordPress/wordpress-develop/blob/trunk/src/js/_enqueues/wp/widgets/media-audio.js#L142-L144).

## Screenshot
CleanShot 2024-05-29 at 19 37 13

ellatrix (approved):
Have you considered a mount effect in the render function?

ellatrix:
Or perhaps the component should open the modal if no `render` function is specified. 🤔

Mamaduka:
> Have you considered a mount effect in the render function?

What do you mean?

I am not sure if it's worth encouraging the hack that the "Inline Image" is using. All other React class component methods are run twice in StrictMode, so we still have to track the dialog state even if it's open on `componentDidMount`.

ellatrix:
```
render={ () => {
  useEffect( () => {
    open()
  }, [] )
} }
```

Mamaduka:
It will have the same issue. The hooks also run one extra time in development mode, so the `open()` will be called twice.

Is there a reason we want to avoid keeping the dialog state tracked in the component?

Mamaduka:
@ellatrix, what do you think about this alternative - #62135?

ellatrix:
Great ✅

Mamaduka:
Thanks for the feedback! 🙇

packages/media-utils/src/components/media-upload/index.js
joemcgill: Left a longer reply [in Slack](https://wordpress.slack.com/archives/C02SX62S6/p1717096254809339?thread_ts=1717055547.327059&cid=C02SX62S6). TL;DR I agree.

PR #62109, opened 2024-05-29
PostCardPanel: Fix ESLint error

Mamaduka:
## What?
Fixes ESLint error in the PostCardPanel component. The new rule was introduced in #62026.

## Testing Instructions
CI checks should be green.


tyxla (approved):
LGTM, thanks for the fix 🚀

t-hamano:
@Mamaduka Regarding E2E tests, that should be fixed in #62111, so would we need to force merge either PR?

PR #62110, opened 2024-05-29
Privatize an experimental prop for inserter menu

draganescu:

### DIFF
--- a/packages/block-editor/src/components/inserter/index.js
+++ b/packages/block-editor/src/components/inserter/index.js
@@ -18,9 +18,12 @@ import { plus } from '@wordpress/icons';
 /**
  * Internal dependencies
  */
-import InserterMenu from './menu';
+import { unlock } from '../../lock-unlock';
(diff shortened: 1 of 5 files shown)


draganescu:
It's unclear here what is going on. In the 1st commit I import from the private apis of the block editor package but that somehow builds fine but crashes the private imports in the browser.

In the second commit it works, but alas actually it's not InserterMenu but InserterLibrary that should be privatised. Because it is InserterLibrary which is used outside of the block editor package, not InserterMenu which introduces the prop, so the public API comes from InserterLibrary ...

But InserterLibrary is already exported as __experimental so ... does that mean anything? Should I drop __experimental now that I make a private version? Do I keep a public experimental component which is different from the private one?

draganescu:
Closed in favor of #62130

draganescu:
Closed in favor of https://github.com/WordPress/gutenberg/pull/62130

PR #62111, opened 2024-05-29
e2e: fix Site Editor Styles test

t-hamano:
## What?

This PR fixes the failing E2E tests.

## Why?

This failure first occurred after #61835 was merged, which [changed the border to box-shadow in the site editor sidebar](https://github.com/WordPress/gutenberg/pull/61835/files#diff-eafe5c0c08285ac8baec6e6005cf059bd61d4b53f70cef3a407cfd32f27373b8L131-R132). This changes the width of the editor canvas by 1px.

The failing test, on the other hand, applies padding in the global styles and asserts a calculated value. But in TT3, [this spacing value uses the clamp function](https://github.com/WordPress/wordpress-develop/blob/a53b5314eabb6afbbd2b19e4c703559f229ee9da/src/wp-content/themes/twentytwentythree/theme.json#L74), so it depends on the width of the editor canvas. Since the width of the editor canvas has changed by 1px, it seems the calculated value would change too.

## Testing Instructions

The E2E tests should pass, but Linting will fail. Linting will be fixed in #62109.

### DIFF
--- a/test/e2e/specs/site-editor/styles.spec.js
+++ b/test/e2e/specs/site-editor/styles.spec.js
@@ -45,15 +45,14 @@ test.describe( 'Styles', () => {
 			.getByRole( 'button', { name: 'Social Icons block styles' } )
 			.click();
 
-		// find the second padding control
-		const paddingControl = page
-			.locator( '[aria-label="padding"]' )
-			.nth( 1 );
-
-		// Change the padding value
-		await paddingControl.fill( '2' );
+		// Find the second padding control and change the padding value
+		await page
+			.getByRole( 'button', { name: 'Set custom size' } )
+			.nth( 1 )
+			.click();
+		await page.getByRole( 'spinbutton', { name: 'padding' } ).fill( '35' );
 
 		// Check the padding value
-		await expect( block ).toHaveCSS( 'padding-left', '35.4932px' );
+		await expect( block ).toHaveCSS( 'padding-left', '35px' );
 	} );
 } );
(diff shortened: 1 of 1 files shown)


Mamaduka (approved):
Thank you, @t-hamano!

Let's give this a try.

test/e2e/specs/site-editor/styles.spec.js
```diff
@@ -54,6 +54,6 @@ test.describe( 'Styles', () => {
 		await paddingControl.fill( '2' );
 
 		// Check the padding value
-		await expect( block ).toHaveCSS( 'padding-left', '35.4644px' );
+		await expect( block ).toHaveCSS( 'padding-left', '35.4932px' );
```
ellatrix: Maybe we should set a px value instead in the padding control for testing 😃

t-hamano: That's true, but I don't have much time left today, so I'd appreciate it if you could just merge it as is, or update the tests 🙏

SantosGuillamot: I think it could be achieved with something like this, although I am not sure if those are the best selectors:

```js
// find the second padding control
await page.getByRole("button", { name: "Set custom size" }).nth(1).click();
await page.locator('[id*="inspector-input-control"]').fill("35");

// Check the padding value
await expect(block).toHaveCSS("padding-left", "35px");
```

Happy to make a follow-up PR if we consider it necessary.

youknowriad:
Do we know why the test didn't fail in the PR?

Mamaduka:
@youknowriad, the branch on `border to box-shadow` PR was probably outdated, and the newly introduced test wasn't available there.

t-hamano:
> we should set a px value instead in the padding control for testing

I have addressed this and the tests should continue to pass.

t-hamano:
Now that I set a px value, this test should be stable. I would like to merge this PR.

ramonjd:
Thanks for fixing this @t-hamano 🍺 !

PR #62118, opened 2024-05-29
Editor: Move the InterfaceSkeleton to the editor package

youknowriad:
Related #52632 

## What?

This is one of the latest pieces of the puzzle of the unification between post and site editors. This PR moves the "InterfaceSkeleton" (the whole UI basically) of the editor to the editor package (from the edit-post package) under the EditorInterface component. To ease with review and avoid breakage and regressions, I chose not to update the site editor to use this component yet, it will require some more changes there, so for now I just focused on the post editor..

## Testing Instructions

1- There should be no functional change or UI change at all but smoke testing the post editor is necessary to check for potential unintended regressions.


packages/edit-post/src/components/back-button/index.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+function BackButton( { initialPost } ) {
+	return (
+		<BackButtonFill>
+			<motion.div
+				variants={ slideX }
+				transition={ { type: 'tween', delay: 0.8 } }
+			>
+				<FullscreenModeClose showTooltip initialPost={ initialPost } />
```
youknowriad, line 28: I'm using a Slot to inject the "W" back button to the post editor. In fact the Slot already exited before for third-party usage, I moved it to the Header component in this PR to be able to inject a different button there in the site editor (different behavior there to exit edit mode)

packages/edit-post/src/components/back-button/style.scss
```diff
@@ -65,3 +65,18 @@
 		margin-top: -($border-width);
 	}
 }
+
+/**
+ *  Show icon label overrides.
+ */
+.show-icon-labels .editor-header {
```
youknowriad, line 71: This is just copied from another CSS file. I didn't touch this CSS at all.

packages/edit-post/src/components/layout/index.js
```diff
@@ -120,7 +113,24 @@ function useEditorStyles() {
...
+		// Add a constant padding for the typewritter effect. When typing at the
+		// bottom, there needs to be room to scroll up.
+		if (
+			! isZoomedOutView &&
+			! hasMetaBoxes &&
+			renderingMode === 'post-only' &&
+			! DESIGN_POST_TYPES.includes( postType )
+		) {
+			baseStyles.push( {
+				css: 'body{padding-bottom: 40vh}',
```
youknowriad, line 130: This style was already applied in the custom "VisualEditor" component that existed in the post editor. But I removed that component (it was mostly useless).

That said, this particular style might need to move to the editor package at some point to apply to "posts" when they load in the site editor context as well.

packages/edit-post/src/components/layout/index.js
```diff
@@ -405,6 +291,25 @@ function Layout( { initialPost } ) {
...
+				// eslint-disable-next-line jsx-a11y/no-autofocus
+				autoFocus={ ! isWelcomeGuideVisible }
+			>
+				{ ! isDistractionFree && showMetaBoxes && (
+					<div className="edit-post-layout__metaboxes">
+						<MetaBoxes location="normal" />
+						<MetaBoxes location="advanced" />
+					</div>
+				) }
+			</EditorCanvas>
```
youknowriad: I wish we didn't need all these props (forceIsDirty, children, disableIframe, autoFocus...) but I think there's no way around them for now (mostly to inject metaboxes UI...)

packages/editor/src/components/text-editor/index.js
```diff
@@ -25,6 +26,14 @@ export default function TextEditor() {
 		};
 	}, [] );
 
+	const titleRef = useRef();
+	useEffect( () => {
+		if ( autoFocus ) {
+			return;
+		}
+		titleRef?.current?.focus();
+	}, [ autoFocus ] );
```
youknowriad, line 35: This actually existed before, but I removed it inadvertently in one of the previous PRs.

draganescu (approved):
Given the size of the code being moved around all I did was use the product. In my testing everything seems in place. Tests passing means 🚢 it!

packages/editor/src/components/editor-canvas/index.js
stokesman, line 1: Do excuse/ignore my bikeshedding but `EditorCanvas` seems odd to me. Is it a terrible idea to simply call it `Editor`? Other ideas: `EditorShell`, `EditorEasel`, `Editorscape`.

draganescu: > Is it a terrible idea to simply call it Editor?

Well we already have an editor which is the underlying system implementing a block editor specific for WP, upon which the post and the site editors are built.

@youknowriad why not preserve Interface and name it EditorInterface?

youknowriad: So I agree that `Editor` would be the best name but I didn't use it yet because I plan for `Editor` to be the only component one would use to render an "postType/postId" editor. 

The component here requires an `EditorProvider` to work still, so it corresponds more to the UI of the editor only. 

I think any of the ideas shared above can work: Canvas is already heavily used in the code base but I'm actually not opinionated `EditorInterface` sound good to me as well. But worth noting it's just an internal component (and may stay that way forever so the naming is not that important). I may go with `EditorInterface` for now.

The `Editor` component will come later.

PR #61602, opened 2024-05-13
Make sure to replace all instances of `:where(body)` instead of just …

tellthemachines:
…the first.

## What?

Fixes #61565 and possibly also #61582.

The non-iframed editor all instances of `body` with `.editor-styles-wrapper` in the editor styles. It also checks for `:where(body)`, but because it was using `replace` with a string instead of a regex, only the first instance was being replaced.

This PR adds a regex with the global flag, so all instances of `:where(body)` will be replaced.

## Testing Instructions

1. In the post editor, under Preferences, enable Custom Fields.
2. Reload the editor and check that base color and font styles still apply to editor content.
3. Blocks should have something like the following rule applied:

```
.editor-styles-wrapper {
    background-color: var(--wp--preset--color--base);
    color: var(--wp--preset--color--contrast);
    font-family: var(--wp--preset--font-family--body);
    font-size: var(--wp--preset--font-size--medium);
    font-style: normal;
    font-weight: 400;
    line-height: 1.55;
    --wp--style--root--padding-top: 0px;
    --wp--style--root--padding-right: 37px;
    --wp--style--root--padding-bottom: 0px;
    --wp--style--root--padding-left: 37px;
}
```

PR #62121, opened 2024-05-30
Transform Styles: Update selector so that styles work when custom fields panel is active

### LINKED ISSUE #62091: Styles missing with custom fields panel active
### Description

The nighly version causes a regression to the state described in issue #61565

1. Edit a page/post
2. From the preferences panel enable the custom fields and save

_No response_

### Environment info

WP: 6.5.3
Theme: TT4
Gutenberg: 18.6.-nightly (18.6.20240528)

Yes

Yes

andrewserong:
## What?

Fixes #62091 

This re-implements the fix from #61602.

Update the selector in `transformStyle` to match the new selector for global styles rules introduced in https://github.com/WordPress/gutenberg/pull/61638 (the rules are now prefixed with `:root `).

## Why?

When the styles were prefixed with the addition of `:root `, the transform in `transformStyles` that allows the styles to work in the non-iframed editor no longer matched against the global styles rules. This PR should restore the behaviour so that the editor looks correct in both iframed and non-iframed modes.

## How?

* Add `:root ` to the selector of the `transformStyle` logic that matches against rules intended for the body element

## Testing Instructions

1. With TT4 theme active, open the post editor
2. In editor preferences, enable Advanced > Custom fields
3. Reload the editor
4. On `trunk` note that the global styles rules are not applied in the editor. With this PR applied the styles should be restored.

## Screenshots or screencast 

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/packages/block-editor/src/utils/test/__snapshots__/transform-styles.js.snap
+++ b/packages/block-editor/src/utils/test/__snapshots__/transform-styles.js.snap
@@ -24,6 +24,12 @@ exports[`transformStyles URL rewrite should rewrite relative paths 1`] = `
 ]
 `;
 
+exports[`transformStyles error handling should handle multiple instances of \`:root :where(body)\` 1`] = `
+[
+  ".my-namespace { color: pink; } .my-namespace { color: orange; }",
+]
+`;
+
 exports[`transformStyles error handling should handle multiple instances of \`:where(body)\` 1`] = `
 [
   ".my-namespace { color: pink; } .my-namespace { color: orange; }",
--- a/packages/block-editor/src/utils/test/__snapshots__/transform-styles.js.snap
+++ b/packages/block-editor/src/utils/test/__snapshots__/transform-styles.js.snap
@@ -30,12 +30,6 @@ exports[`transformStyles error handling should handle multiple instances of \`:r
 ]
 `;
 
-exports[`transformStyles error handling should handle multiple instances of \`:where(body)\` 1`] = `
-[
-  ".my-namespace { color: pink; } .my-namespace { color: orange; }",
-]
-`;
-
 exports[`transformStyles selector wrap should ignore font-face selectors 1`] = `
 [
   "
--- a/packages/block-editor/src/utils/test/transform-styles.js
+++ b/packages/block-editor/src/utils/test/transform-styles.js
@@ -52,8 +52,8 @@ describe( 'transformStyles', () => {
 			);
 		} );
 
-		it( 'should handle multiple instances of `:where(body)`', () => {
-			const input = `:where(body) { color: pink; } :where(body) { color: orange; }`;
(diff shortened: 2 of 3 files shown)


test/e2e/specs/site-editor/styles.spec.js
```diff
@@ -54,6 +54,6 @@ test.describe( 'Styles', () => {
 		await paddingControl.fill( '2' );
 
 		// Check the padding value
-		await expect( block ).toHaveCSS( 'padding-left', '35.4644px' );
+		await expect( block ).toHaveCSS( 'padding-left', '35.4932px' );
```
andrewserong: For some reason this line was failing:

image

So I've updated this to the new expected value to see if it gets the tests passing. Separately, I wonder if this check should be simplified to ignore the fraction, as this effectively averages out to around 36px anyway 🤔

tellthemachines (commented):
Thanks for the PR! I'm thinking we might need to check for both `:root :where(body)` and `:where(body)`, because there's at least one instance of the latter (the rule that [sets the margin to 0](https://github.com/WordPress/gutenberg/blob/trunk/lib/class-wp-theme-json-gutenberg.php#L2872)) that will otherwise be ignored.

test/e2e/specs/site-editor/styles.spec.js
tellthemachines: Oh that's going to be a fun test to maintain 🤦

aaronrobertshaw (approved):
Thanks for fixing this up @andrewserong 🚀 

After a false start due to a local theme issue, this is testing well for me.

LGTM 🚢

andrewserong:
Thanks for the reviews, folks!

@tellthemachines I've updated the rule to chain the existing rule and the new one (with the new one running first). Does that look okay to do? The output in the post editor seems to look okay to me now:

image

aaronrobertshaw:
The tweak for the margin rule works for me 👍 

Screenshot 2024-05-30 at 12 06 13 PM

test/e2e/specs/site-editor/styles.spec.js
talldan: I think it would be better to use a custom value - e.g. set it to 5px, and then the expected padding-left should be 5px. I can look at updating this as part of the tests I'm working on.

tellthemachines (approved):
Thanks for updating! LGTM now ✅

test/e2e/specs/site-editor/styles.spec.js
andrewserong: Looks like it was just updated to a hard-coded value in https://github.com/WordPress/gutenberg/pull/62111 so I've reverted the change here.

Mamaduka:
Thank you, @andrewserong!

PR #62124, opened 2024-05-30
Global Styles: Prevent Typography panel title from wrapping

t-hamano:
## What?

This PR will prevent Typography panel title from wrapping in certain languages.

## Why?

The toggle button in the font library is wrapped in a `H-Stack` component. This component has `width: 100%`, so it may squash elements at the same level.

h-stack

## How?

I removed the `H-Stack` component. At the same time, I also removed the redundant `Tooltip` component.

## Testing Instructions

- Change the site language to Japanese and update the translation file.
- Open the Typography panel in the Global Styles.
- Confirm that the title does not wrap.
- When you hover over the toggle button, the tooltip appears as before.

## Screenshots or screencast 

| Locale | Before | After |
|--------|--------|--------|
| Japanese (ja) | image | image |
| 繁體中文 (zh_TW)| image | image|

### DIFF
--- a/packages/edit-site/src/components/global-styles/font-families.js
+++ b/packages/edit-site/src/components/global-styles/font-families.js
@@ -7,7 +7,6 @@ import {
 	__experimentalVStack as VStack,
 	__experimentalHStack as HStack,
 	Button,
-	Tooltip,
 } from '@wordpress/components';
 import { settings } from '@wordpress/icons';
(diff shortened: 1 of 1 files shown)


mikachan (approved):
Code looks good and tests well. Thanks @t-hamano!

t-hamano:
@mikachan Thanks for the review!

PR #57908, opened 2024-01-17
Block Styles: Extend block style variations as mechanism for achieving section styling

aaronrobertshaw:
Addresses: 
- #57537

Related:
- #57841
- https://github.com/WordPress/gutenberg/pull/56540

## What?

https://github.com/WordPress/gutenberg/pull/56540 aimed to render block style variation styles within the overall global styles stylesheet. This was problematic in that it required increasing specificity of element styles applied to block instances as well as being unable to be reliably nested.

The idea in this PR is to reduce the specificity as far as possible for block style variations and their inner element and block type styles. Then generate stylesheets for the application of these variations in the order of their application (parent > child > grandchild etc) such that the correct cascade is maintained. This is so block style variations can be applied on nested blocks.

## Why?

The ability to style entire sections of a page without having to tediously reapply the same sets of styles is greatly desired. Block style variations seemed to be the best mechanism for defining these styles.

## How?

- Leverages global styles data added to the block editor settings via [#61556](https://github.com/WordPress/gutenberg/pull/61556) 
- Allows early return from utils scoping selectors
- Style variation theme.json partials are classified as block style variations if they define a top-level `blockTypes` property. 
  - If a theme.json partial does not specify a non-empty array for the `blockTypes` key, it will be treated as a theme style variation for backwards compatibility. 
- Supports automatic registration of block style variations that are defined via:
  - Standalone theme.json partial files in `/styles` that have `"block"` within its defined scopes
  - The `styles.blocks.variations` property of a theme's primary theme.json file
  - The `styles.blocks.variations` property of a theme style variation partial theme.json
- Existing theme.json data filters have been leveraged to absorb the above shared block style variation definitions into their respective block type styles: e.g. `styles.blocks.core/group.variations`
- A `block-style-variations` block support has been added that injects a unique CSS class per block instance application of a block style variation.
- The "block style variations" block support also generates a partial theme.json stylesheet for each application of a variation (this is the key to nesting variations)
- Adds JS hook for variations and defines `useBlockProps` to generate the styles on the editor side so they update correctly upon application to blocks.
- Block style variation definitions are applied in the following order:
  - Block style registry
  - Shared theme.json partials in `/styles`
  - Primary theme.json file
  - User origin data e.g. applied theme style variation definitions
- Theme.json class has been updated to not strip out extended block style variation data e.g. inner element/block type styles or shared variation definitions.
- The theme.json class generation of block metadata was tweaked so that:
   - Block style variations within the block styles registry are included in the valid variations
[...]

PR #62125, opened 2024-05-30
Block Styles: Fix block style variation styles for blocks with complex selectors

### LINKED ISSUE #62122: Double border on buttons with outline variation applied
### Description

When adding a button and applying the outline variation, a double border is being applied. **(issue exists on front end, not in editor)** It appears that there are two instances of classes/styling being applied—one to the `<div>` and the other to the `<a>`:

```
<div class="wp-block-button is-style-outline is-style-outline--8b999049ec2131a13968b599677970ef">
<a class="wp-block-button__link wp-element-button" href="#">Learn More</a>
</div>
```

First set of styles, to the `<div>`:

```
:root :where(.is-style-outline--cb3c5e7393271d2adb5a2454ddebf7f6) {
    background: transparent none;
    border-color: currentColor;
    border-width: 1px;
    border-style: solid;
    color: currentColor;
    padding-top: calc(0.6rem - 1px);
    padding-right: calc(1rem - 1px);
    padding-bottom: calc(0.6rem - 1px);
    padding-left: calc(1rem - 1px);
}
```

Second set of styles, to the `<a>` (note some specificity overriding happening to the below):

```
:root :where(.wp-block-button.is-style-outline .wp-block-button__link) {
    background: transparent none;
    border-color: currentColor;
    border-width: 1px;
    border-style: solid;
    color: currentColor;
    padding-top: calc(0.6rem - 1px);
    padding-right: calc(1rem - 1px);
    padding-bottom: calc(0.6rem - 1px);
    padding-left: calc(1rem - 1px);
}
:root :where(.wp-element-button, .wp-block-button__link) {
    background-color: var(--wp--preset--color--contrast);
    border-radius: .33rem;
    border-color: var(--wp--preset--color--contrast);
    border-width: 0;
    color: var(--wp--preset--color--base);
    font-family: inherit;
    font-size: var(--wp--preset--font-size--small);
    font-style: normal;
    font-weight: 500;
    line-height: inherit;
    padding-top: 0.6rem;
    padding-right: 1rem;
    padding-bottom: 0.6rem;
    padding-left: 1rem;
    text-decoration: none;
}
```

1. Go to post editor.
2. Add button block, apply outline variation.
3. See issue. (note issue exists on front end, not in editor)

image

### Environment info

WordPress 6.5.3, Gutenberg Nightly (Version 18.6.20240529)

Yes

Yes

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/62122

## What?

Fixes the style generation for instanced block style variations for blocks that use complex selectors. For example:
- Button: `.wp-block-button .wp-block-button__link`
- Image: 
   - `.wp-block-image`
   - `.wp-block-image img, .wp-block-image .wp-block-image__crop-area, .wp-block-image .components-placeholder`

## Why?

Without this fix, variation styles when applied to an instance would only be applied to the block's wrapper. This means styles like the Button block's outline block style would get double borders, padding et.c

## How?

Reworks the generation of the instanced block style variation stylesheets so that the result has selectors that apply different types of styles to the block's desired elements. For example, the Button block styles go to the inner `.wp-block-button__link` element or an Image's border goes to the inner `img`

The style load order was also tweaked to ensure the block style variations styles were loaded after the block library styles. 

## Testing Instructions

1. Register some custom block style variations following [#57908](https://github.com/WordPress/gutenberg/pull/57908)'s instructions
2. Apply these in a nested fashion and ensure they still continue to work
3. Make sure you have a button block in your post or page
4. Apply the outline block style to the button
5. Make sure all the blocks display correctly in both editors and the frontend
6. Disable separate core block assets via snippet below or activate a theme which does such as [Powder](https://github.com/bgardner/powder/tree/main)
7. Confirm block library styles do not override the outline block style variation styles for Button blocks

`add_filter( 'should_load_separate_core_block_assets', '__return_false' );`

## Screenshots or screencast 

| Before | After |
|---|---|
| Screenshot 2024-05-30 at 4 25 25 PM | Screenshot 2024-05-30 at 4 25 08 PM |

### DIFF
--- a/lib/block-supports/block-style-variations.php
+++ b/lib/block-supports/block-style-variations.php
@@ -7,7 +7,7 @@
  */
 
 /**
- * Get the class name for this application of this block's variation styles.
+ * Generate class name for this application of this block's variation styles.
  *
  * @since 6.6.0
  *
@@ -16,7 +16,7 @@
  *
  * @return string The unique class name.
  */
-function gutenberg_get_block_style_variation_class_name( $block, $variation ) {
+function gutenberg_create_block_style_variation_class_name( $block, $variation ) {
 	return 'is-style-' . $variation . '--' . md5( serialize( $block ) );
 }
 
@@ -79,20 +79,27 @@ function gutenberg_render_block_style_variation_support_styles( $parsed_block ) 
 		return $parsed_block;
 	}
 
-	$class_name         = gutenberg_get_block_style_variation_class_name( $parsed_block, $variation );
+	$class_name         = gutenberg_create_block_style_variation_class_name( $parsed_block, $variation );
 	$updated_class_name = $parsed_block['attrs']['className'] . " $class_name";
 	$variation_instance = substr( $class_name, 9 );
-	$class_name         = ".$class_name";
+	$selector           = ".$class_name";
 
 	/*
-	 * To support blocks with more complex selectors, that can apply styles to
-	 * inner markup or even different inner elements depending on the feature,
-	 * the variation style data needs to be manipulated some. If the root variation
-	 * styles are moved under a variations property they will generate as desired.
+	 * Even though block style variations are effectively theme.json partials,
+	 * they can't be processed completely as though they are.
 	 *
-	 * Example blocks that require this approach include:
-	 *  - Button: `.wp-block-button .wp-block-button__link`
-	 *  - Image: `.wp-block-image` and `.wp-block-image img` for borders, shadow etc.
+	 * Block styles support custom selectors to direct specific types of styles
+	 * to inner elements. For example, borders on Image block's get applied to
+	 * the inner `img` element rather than the wrapping `figure`.
+	 *
+	 * The following relocates the "root" block style variation styles to
+	 * under an appropriate blocks property to leverage the preexisting style
+	 * generation for simple block style variations. This way they get the
+	 * custom selectors they need.
+	 *
+	 * The inner elements and block styles for the variation itself are
+	 * still included at the top level but scoped by the variation's selector
+	 * when the stylesheet is generated.
 	 */
 	$elements_data = $variation_data['elements'] ?? array();
 	$blocks_data   = $variation_data['blocks'] ?? array();
@@ -113,7 +120,7 @@ function gutenberg_render_block_style_variation_support_styles( $parsed_block ) 
 		),
 	);
 
-	// Turn off filter that excludes block nodes. They are needed here.
+	// Turn off filter that excludes block nodes. They are needed here for the variation's inner block types.
 	if ( ! is_admin() ) {
 		remove_filter( 'wp_theme_json_get_style_nodes', 'wp_filter_out_block_nodes' );
 	}
@@ -128,14 +135,14 @@ function gutenberg_render_block_style_variation_support_styles( $parsed_block ) 
 		array( 'custom' ),
 		array(
 			'skip_root_layout_styles' => true,
-			'scope'                   => $class_name,
+			'scope'                   => $selector,
 		)
 	);
 
 	// Clean up temporary block style now instance styles have been processed.
 	$styles_registry->unregister( $parsed_block['blockName'], $variation_instance );
 
-	// Restore filter that excludes block nodes. They are needed here.
+	// Restore filter that excludes block nodes.
 	if ( ! is_admin() ) {
 		add_filter( 'wp_theme_json_get_style_nodes', 'wp_filter_out_block_nodes' );
 	}
@@ -179,7 +186,7 @@ function gutenberg_render_block_style_variation_class_name( $block_content, $blo
 	 * Matches a class prefixed by `is-style`, followed by the
 	 * variation slug, then `--`, and finally a hash.
 	 *
-	 * See `gutenberg_get_block_style_variation_class_name` for class generation.
+	 * See `gutenberg_create_block_style_variation_class_name` for class generation.
 	 */
 	preg_match( '/\bis-style-(\S+?--\w+)\b/', $block['attrs']['className'], $matches );
 
@@ -211,7 +218,7 @@ function gutenberg_render_block_style_variation_class_name( $block_content, $blo
  *
  * @param array $variations Shared block style variations.
  *
- * @return array Block variations data to be merged under styles.blocks
+ * @return array Block variations data to be merged under `styles.blocks`.
  */
 function gutenberg_resolve_and_register_block_style_variations( $variations ) {
 	$variations_data = array();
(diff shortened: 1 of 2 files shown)


aaronrobertshaw:
The changes here have been included within the overall backport for the section styles feature: https://github.com/WordPress/wordpress-develop/pull/6662.

If this PR requires further updates, that backport will also need updating.

lib/block-supports/block-style-variations.php
```diff
@@ -79,31 +79,63 @@ function gutenberg_render_block_style_variation_support_styles( $parsed_block )
...
 	}
 
-	$config = array(
-		'version' => WP_Theme_JSON_Gutenberg::LATEST_SCHEMA,
-		'styles'  => $variation_data,
-	);
-
 	$class_name         = gutenberg_get_block_style_variation_class_name( $parsed_block, $variation );
 	$updated_class_name = $parsed_block['attrs']['className'] . " $class_name";
+	$variation_instance = substr( $class_name, 9 );
```
talldan: It feels unusual now that the class name is created, to immediately remove part of it here.

Maybe `gutenberg_get_block_style_variation_class_name` could be changed to `gutenberg_get_block_style_variation_instance`, and then the `is-style-` part appended on after (there could still be a `gutenberg_get_block_style_variation_class_name` function if needed for that but it accepts the variation_instance).

lib/block-supports/block-style-variations.php
```diff
@@ -79,31 +79,63 @@ function gutenberg_render_block_style_variation_support_styles( $parsed_block )
...
+	$class_name         = ".$class_name";
+
+	/*
+	 * To support blocks with more complex selectors, that can apply styles to
+	 * inner markup or even different inner elements depending on the feature,
+	 * the variation style data needs to be manipulated some. If the root variation
+	 * styles are moved under a variations property they will generate as desired.
```
talldan: I'm not quite following this TBH, might need to explain it more. 😄

talldan (commented):
It does fix the issue, but I don't understand the fix, so I'd have a hard time approving it.

I do notice there are two sets of identical styles:
Screenshot 2024-05-30 at 4 31 04 pm

If there's any way to avoid that it might be good, but it could be considered a separate change.

lib/block-supports/block-style-variations.php
aaronrobertshaw: I was trying to avoid writing an essay here. Would linking to the issue and PR help?

aaronrobertshaw: I have a better idea on how best to explain this bit of the code now. I'll update the inline comment shortly.

For the discussion on this PR though:

The crux of the issue is that blocks can define their own selectors to direct certain styles to inner parts of the blocks markup. For example, the inner `img` element for an Image block needs to get the border styles whereas the other styles might still belong on the outer `figure` wrapper for the block.

Previously, the variation's data was treated as its own "theme.json partial". The problem here is that root styles for theme.json don't have the ability to be directed to inner parts as that's what the `styles.elements` and `styles.blocks` properties do.

The fix in this PR essentially moves the variation's root styles to under the `styles.blocks` property so that the block's custom selectors can be honored.

aaronrobertshaw: Done https://github.com/WordPress/gutenberg/pull/62125/commits/2f33a3d130d8a84f4659c6f94e88abc2f9063430

lib/block-supports/block-style-variations.php
aaronrobertshaw: Done https://github.com/WordPress/gutenberg/pull/62125/commits/a4662f2256a08331ec8685e919e1b97297910a67

I'll update the backport accordingly shortly.

aaronrobertshaw:
Appreciate the thorough review @talldan 🙇 

I believe I've addressed all the feedback so far. Should be good to go for another spin now.

> If there's any way to avoid that it might be good, but it could be considered a separate change.

I think that is definitely best done via a separate follow-up. 

For the record, one set of styles is from the standard theme.json stylesheet but can't be relied upon as if variations are nested the load order needs to be different for the correct variation's styles to take precedence. That's where the second set of styles comes in as they are the styles specific to this individual application or instance of the block style variation.

The code creating the standard set of styles is actually leveraged to generate the individual instance styles. 

It is 100% something that would be good to clean up. Right now though, I think we need to land a fix for the issue affecting outline buttons etc. then I can circle back to this optimisation and clean up.

bgardner:
@aaronrobertshaw 

There seems to still be an issue with specificity and styles. My Powder theme which uses theme.json to style the outline variation (https://github.com/bgardner/powder/blob/main/theme.json#L288) is being overridden by the following:

```
:root :where(.wp-block-button .wp-block-button__link.is-style-outline), :root :where(.wp-block-button.is-style-outline>.wp-block-button__link) {
    border: 2px solid;
    padding: .667em 1.333em;
}
```

bgardner:
The button should have a border of 1px thick, but looks like this (note the lower right side styles being overridden:

image

aaronrobertshaw:
Thanks for the extra information @bgardner, it definitely helps 👍 

After downloading your theme and poking around a bit, it looks like the style load order isn't quite right. The key difference I noted while testing was that Powder disables the separate block assets via the `should_load_separate_core_block_assets` filter.

The `wp-block-library-inline-css` styles are added after the global styles and block style variation styles. When the block styles are separated however their individual stylesheets are loaded before the global styles and variation styles.

I'll work on a fix for this first thing tomorrow. It might also be better as a separate PR.

| Separate Block Assets | All Together |
|---|---|
| Screenshot 2024-05-30 at 11 39 24 PM | Screenshot 2024-05-30 at 11 29 52 PM |

_Note: The separate assets screenshot is actually TT4 but illustrates the point._

bgardner:
@aaronrobertshaw That makes sense, thanks for spelling that out. If there's an alternative (more optimal) way for me to handle things in Powder, I'm all ears.

aaronrobertshaw:
> If there's an alternative (more optimal) way for me to handle things in Powder, I'm all ears.

I don't think you should need to make any changes here. Hopefully, a bit of tweaking to the order styles load in should see this working as expected.

aaronrobertshaw:
@bgardner with fresh eyes this morning it looks like the tweak the load order is rather trivial so I've rolled it into this fix.

| Before | After |
|---|---|
| Screenshot 2024-05-31 at 9 55 29 AM | Screenshot 2024-05-31 at 9 55 09 AM |

bgardner:
@aaronrobertshaw Can confirm, per the below. LGTM!

image

talldan (approved):
Looks good now, thanks for adding the improved explanations in the code, that helped me understand the issue and fix.

aaronrobertshaw:
Appreciate all the reviews and testing. Thanks @talldan and @bgardner 🙇

Issue #62089, opened 2024-05-29
Link Control: Unable to edit text in development mode


Mamaduka:
The `LinkControl` popover steals the focus and prevents editing the text in development mode. See: #61943.

## Affected e2e tests:

* `editor/blocks/links.spec.js` - should allow editing text underneath popover when activated via mouse.
* `editor/various/block-bindings.spec.js` - should add empty button block when pressing enter.

## Steps to reproduce

* Checkout branch from #61943.
* Enable `SCRIPT_DEBUG` for non `wp-env` setups.
* Create a post.
* Add a paragraph and add a link to some selection.
* Try editing the linked text.

## Screencast

https://github.com/WordPress/gutenberg/assets/240569/d729d623-2ea8-4cad-98af-6fd862534923

Mamaduka:
My guess is that the following effect might be responsible for the bug:

https://github.com/WordPress/gutenberg/blob/7fbb0e204b2aee0d826fe2aa83c336df85fc76f8/packages/block-editor/src/components/link-control/index.js#L218-L239

cc @getdave, @jeryj

PR #62126, opened 2024-05-30
Button: Fix ESLint warning

Mamaduka:
## What?
PR updates the `useSelect` hook dependencies for the button block and fixes the ESLint warning.

Noticed while debugging #62089.

```
React Hook useSelect has a missing dependency: 'metadata?.bindings?.url'. Either include it or remove the dependency array.
```

## Testing Instructions
The functionality has e2e test coverage. CI checks should be green.

Same

### DIFF
--- a/packages/block-library/src/image/edit.js
+++ b/packages/block-library/src/image/edit.js
@@ -145,7 +145,7 @@ export function ImageEdit( {
 				scale: undefined,
 			} );
 		}
-	}, [ align ] );
+	}, [ __unstableMarkNextChangeAsNotPersistent, align, setAttributes ] );
 
 	const { getSettings } = useSelect( blockEditorStore );
 	const blockEditingMode = useBlockEditingMode();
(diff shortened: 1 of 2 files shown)


SantosGuillamot (approved):
Thanks for the PR! It looks good to me 🙂 

We should probably make the same change in [the image edit file](https://github.com/WordPress/gutenberg/blob/6437c207afefb6625bdeeb1d95e3281b63a8dc9e/packages/block-library/src/image/edit.js#L307-L332), right?

Mamaduka:
> We should probably make the same change in [the image edit file](https://github.com/WordPress/gutenberg/blob/6437c207afefb6625bdeeb1d95e3281b63a8dc9e/packages/block-library/src/image/edit.js#L307-L332), right?

Yes. I'll update the PR. Thanks for testing, @SantosGuillamot!

packages/block-library/src/image/edit.js
```diff
@@ -145,7 +145,7 @@ export function ImageEdit( {
 				scale: undefined,
 			} );
 		}
-	}, [ align ] );
+	}, [ __unstableMarkNextChangeAsNotPersistent, align, setAttributes ] );
```
Mamaduka, line 148: I decided to fix both warnings in the file. Both callbacks have stable references and won't change the effect's behavior.

SantosGuillamot: Thanks for the fixes 🙌

PR #62127, opened 2024-05-30
Blocks: remove pipe usage and dependency on compose

jsnajdr:
I've been trying to improve types for the `blocks` package -- it currently doesn't export any and yet other packages attempt to import them -- and this PR is the first related result: removes usages of `pipe`, which can be replaced by a much simpler code.


ellatrix (approved):

tyxla:
Nice one 👍

PR #60735, opened 2024-04-14
Rich text: try BITS with block binding API

ellatrix:
Depends on WordPress/wordpress-develop#6390
Test in [Playground](https://playground.wordpress.net/?blueprint-url=https://gist.github.com/dmsnell/4feeb1ad35baecb15aef6ca598ba2b6a).

## What?

bit-binding

Demo content:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"isbn"}}}}} -->
<p>original</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>ISBN: <//wp:post-meta key=isbn>.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>ISBN: <//wp:post-meta key=isbn>!</p>
<!-- /wp:paragraph -->
```

PR #62128, opened 2024-05-30
Rich text: preserve comments

### LINKED ISSUE #13318: Inline HTML comments are removed when a Block is edited
**Describe the bug**

If a Block has an inline HTML comment, the HTML comment is stripped out when the Block is modified.

Sample text:
```
This is my paragraph text with <!-- html-comment -->an HTML comment<!-- /html-comment -->
```

**To Reproduce**

1. Create a new Paragraph Block.
2. Add the sample text in HTML mode.
3. Switch to Visual mode.
4. Switch back to HTML mode and see that the HTML comment persists.
5. Switch to Visual mode and make some edit to the text.
6. Switch back to HTML mode and see that the HTML comment has disappeared.

See GIF:

htmlcommentdisappear

**Expected behavior**

My expected behavior is that the HTML comment persists when I edit the text.

In the scenario where I edit within the boundary of the HTML comment, my expected behavior is that the Block Editor would mirror the Classic Editor.

**Desktop (please complete the following information):**
 - OS: Mac OS 10.14.2
 - Browser: Firefox 64.0.2
 - Version: Gutenberg 4.8.0

ellatrix:
## What?

Extracted from #60735.
Fixes #13318.

This PR preserves HTML comments in rich text. Currently they are stripped out when HTML is parsed by rich text.

## Why?

We should leave the content intact and interfere as little as possible. While useful on its own (and fixing a bug), this is also a preparatory PR for Bits, as it will make use of comments to anchor dynamic content inline. In the editor, the comment will then be replaced through an API by either a placeholder or the data.

## How?

Rich text uses the browser's HTML parser, so it's quite easy to store this internally.

## Testing Instructions

Insert a comment in a paragraph (through the code/HTML editor) and switch back to visual. There should be a placeholder. Edit the paragraph, and the placeholder/comment should remain.

## Screenshots or screencast 

image

### DIFF
--- a/packages/block-editor/src/components/rich-text/content.scss
+++ b/packages/block-editor/src/components/rich-text/content.scss
@@ -13,10 +13,6 @@
 	&:focus {
 		// Removes outline added by the browser.
 		outline: none;
-
-		[data-rich-text-format-boundary] {
-			border-radius: $radius-small;
-		}
 	}
 }
 
@@ -32,11 +28,29 @@ figcaption.block-editor-rich-text__editable [data-rich-text-placeholder]::before
 	opacity: 0.8;
 }
 
+[data-rich-text-comment],
+[data-rich-text-script],
+[data-rich-text-format-boundary] {
+	border-radius: 2px;
+}
+
+[data-rich-text-comment],
 [data-rich-text-script] {
 	display: inline;
+	background-color: currentColor;
 
 	&::before {
-		content: "</>";
-		background: rgb(255, 255, 0);
+		filter: invert(100%);
+		color: currentColor;
 	}
+}
+
+[data-rich-text-script]::before {
+	content: "</>";
+}
+
+// Only fill in the comment content if the comment is not replaced (in the
+// future they might be by Bits).
+[data-rich-text-comment]:empty::before {
+	content: "[" attr(data-rich-text-comment) "]";
 }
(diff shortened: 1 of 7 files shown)


lysyjan (approved):

packages/block-editor/src/components/rich-text/content.scss
```diff
@@ -40,3 +36,18 @@ figcaption.block-editor-rich-text__editable [data-rich-text-placeholder]::before
 		background: rgb(255, 255, 0);
 	}
 }
+
+[data-rich-text-comment],
```
youknowriad, line 40: What is this attribute about?

packages/block-editor/src/components/rich-text/content.scss
```diff
@@ -40,3 +36,18 @@ figcaption.block-editor-rich-text__editable [data-rich-text-placeholder]::before
...
+[data-rich-text-format-boundary] {
+	border-radius: 2px;
```
youknowriad: Why not $radius-small like before?

ellatrix: Hm, I didn't see that, maybe it was changed and missed it in a rebase 🤔

packages/block-editor/src/components/rich-text/content.scss
ellatrix: It stores the data inside comments

ellatrix: And also marks a placeholder as a comment

packages/block-editor/src/components/rich-text/content.scss
ellatrix: Fixed

youknowriad (approved):

PR #58580, opened 2024-02-01
Font Library: Show 'Add fonts' button when there are no fonts installed

noisysocks:
## What?
Update the Typography panel in Global Styles to show an 'Add fonts' button when there are no fonts installed. Also switches the `typography` icon to `settings` which better matches other Global Styles panels.

## Why?
Implements the suggestions made in https://github.com/WordPress/gutenberg/issues/55179#issuecomment-1915558985 and https://github.com/WordPress/gutenberg/issues/55179#issuecomment-1921896400.

## Testing Instructions
1. Activate a theme that has fonts e.g. TT4.
1. Navigate to the site editor and open Global Styles.
1. Navigate to Typography. You should see the installed fonts listed and no 'Add fonts' button.
1. Activate a theme that has no fonts e.g. Emptytheme.
1. Navigate to the site editor and open Global Styles.
1. Navigate to Typography. You should see 'No fonts installed.' with an 'Add fonts' button underneath.

## Screenshots or screencast 

Screenshot 2024-02-02 at 10 39 47
Screenshot 2024-02-02 at 10 40 04

Issue #58403, opened 2024-01-29
Font Library: Global Styles > Typography sidebar – group related font sources


colorful-tones:
## What problem does this address?
Give site builders easier glance at activated font's source: theme, user uploaded, Google Font installed.

This is related to #55220, but this focuses on mirroring and reinforcing the grouping of font sources that #55220 mentions in the Global Styles > Typography sidebar area. Ideally, site builders would not even have to open the Font Library to inspect what fonts are activated and where they're coming from: theme, user uploaded, etc. The Font Library already has some of this UI in the modal, but not in the Global Styles > Typography sidebar.

## What is your proposed solution?
Group font sources in the sidebar under a heading and reinforce the work done in #55220 .

The screenshot below shows how the current iteration of the Font Library modal groups 'Theme Fonts' underneath user added fonts (note: there is no heading to indicate that the fonts are 'User Fonts' but would be good to do this, and it is proposed in #55220)

![Screenshot 2024-01-29 at 2 05 41 PM](https://github.com/WordPress/gutenberg/assets/405912/b24678d7-5ebc-4788-bc19-67c5f993d685)

matiasbenedetto:
I would like to know @jasmussen thoughts about this proposal.

t-hamano:
Regarding the sidebar, this will probably also be related to #55179, which discusses improvements to the sidebar UI.

Regarding the modal dialog, I think we should at least add a heading of "User Fonts" or "Custom Fonts" or "Installed Fonts" for the font group the user has installed. 

![image](https://github.com/WordPress/gutenberg/assets/54422211/9b1cc8b9-e760-416a-866a-3b20ccb9c864)

jasmussen:
> I would like to know @jasmussen thoughts about this proposal.

I think it can be explored, either in beta or for future versions; we have options there. But I wouldn't rush it — the interface has barely been used to make themes with, and before we start to split the interface apart, it'd be good to get some collect some more feedback. How would you feel about this one resting for a bit?

afercia:
I'd agree that, generally, any list of items should be preceded by a heading or other semantic labeling to clarify _what_ that list is about.

At least the modal dialog content should be improved.

RIght now, it's not clear why the theme fonts do have a visible title and other fonts don't. It's not even clear what the other fonts are and where they come from.

Actually, the 'Theme Fonts' title is not even a HTML heading. It's just text wrapped within unsemantic div and span elements.

[...]

PR #62129, opened 2024-05-30
Make the fonts management modal dialog more discoverable

### LINKED ISSUE #55179: Font Library: consider making the library more discoverable
Pulling in feedback from the [FSE Outreach Program's Final Touches call for testing](https://make.wordpress.org/test/2023/09/18/fse-program-testing-call-26-final-touches/#comment-3177):

>I clicked the “Aa” icon to open up font management just because I followed your instructions, but I couldn’t have found the font management easily without them. I love icons, but this is only visible with a label.

I've heard this from a few folks at this point and it's clear that this feature isn't easily discoverable. Perhaps we could consider it as part of iteration for design details: https://github.com/WordPress/gutenberg/issues/53483 but for now I wanted to open an issue to gather this feedback.

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/55179

## What?

Direct users feedback from the FSE Outreach Program and feedback from other contributors reported the font management modal dialog is little discoverable. https://github.com/WordPress/gutenberg/pull/58580 was only a partial improvement and only addressed the scenario when there's no fonts installed by making two changes:
- By adding an 'Add fonts' button. with visible text. This button opens the modal dialog to the 'Upload' tab.
- By replacing the `Aa` icon with the `settings` icon fot the button that opens the modal dialog to the 'Library' tab.

Screenshot of current situation:

before

I can see a few issues with this.

**When there's no fonts installed:**
- The Manage Fonts 'settings' icon button opens the modal dialog showing the Library tab panel. Of course, when there's no fonts installed, this panel is empty. There's nothing users can do here.
- As far as I can tell, this is a unique case in the editor where a 'settings' icon button opens a modal dialog. This is inconsistent and potentially unexpected for users. In fact, the 'settings' icon is generally used to switch a setting control to an alternative version or for view / display settings.
- The modal dialog is a place where users can manage, upload, install, activate, deactivate fonts. It's not about 'settings'.

When there are fonts installed:
- The 'Add fonts' button disappears. This is potentially unexpected for users. As a user, I would wonder 'where's that button that I previously clicked? How can I open the font Library?' There's only two ways to do that:
  - Click an individual font. However, that brings users to the font details 'sub view'. Users would have to click go back to get the font Library.
  - Click the settings icon button. Which is reported as little discoverable and it's the main point of this issue.

## Why?

- Opening the fonts modal dialog is little discoverable.

## How?

- Replaces the 'settings' icon button with a 'Manage fonts' button with visible text.
- The new button is consistent with the 'Add fonts' button. It's where users would expect to see it.
- It is only shown in place of the 'Add fonts' button, when there are fonts installed.

## Testing Instructions

- Start with no fonts installed, or uninstall / deactivate all fonts.
- Go to Site editor > Styles > Typography
- In the Typography panel, observe there's no 'settings' icon button any longer.
- Observe the 'Add fonts' button is unchanged.
- Add or activate some fonts and then close the modal dialog.
- Observe there's a new 'Manage fonts' button at the end of the fonts list.
- Click the 'Manage fonts' button. Observe it brings directly to the modal dialog > Library tab.

## Screenshots or screencast 

After:

after

### DIFF
--- a/packages/edit-site/src/components/global-styles/font-families.js
+++ b/packages/edit-site/src/components/global-styles/font-families.js
@@ -5,11 +5,13 @@ import { __ } from '@wordpress/i18n';
 import {
 	__experimentalItemGroup as ItemGroup,
 	__experimentalVStack as VStack,
-	__experimentalHStack as HStack,
 	Button,
 } from '@wordpress/components';
+<<<<<<< HEAD
 import { privateApis as blockEditorPrivateApis } from '@wordpress/block-editor';
 import { settings } from '@wordpress/icons';
+=======
+>>>>>>> 72a59c77a0 (Replace Manage fonts icon button with visible text button.)
 import { useContext } from '@wordpress/element';
 
 /**
@@ -51,24 +53,31 @@ function FontFamilies() {
 			) }
 
 			<VStack spacing={ 2 }>
-				<HStack justify="space-between">
-					<Subtitle level={ 3 }>{ __( 'Fonts' ) }</Subtitle>
-					<Button
-						onClick={ () => setModalTabOpen( 'installed-fonts' ) }
-						label={ __( 'Manage fonts' ) }
-						icon={ settings }
-						size="small"
-					/>
-				</HStack>
+				<Subtitle level={ 3 }>{ __( 'Fonts' ) }</Subtitle>
 				{ hasFonts ? (
-					<ItemGroup isBordered isSeparated>
-						{ customFonts.map( ( font ) => (
-							<FontFamilyItem key={ font.slug } font={ font } />
-						) ) }
-						{ themeFonts.map( ( font ) => (
-							<FontFamilyItem key={ font.slug } font={ font } />
-						) ) }
-					</ItemGroup>
+					<>
+						<ItemGroup isBordered isSeparated>
+							{ customFonts.map( ( font ) => (
+								<FontFamilyItem
+									key={ font.slug }
+									font={ font }
+								/>
+							) ) }
+							{ themeFonts.map( ( font ) => (
+								<FontFamilyItem
+									key={ font.slug }
+									font={ font }
+								/>
+							) ) }
+						</ItemGroup>
+						<Button
+							className="edit-site-global-styles-font-families__manage-fonts"
+							variant="secondary"
+							onClick={ () => setModalTabOpen( 'installed-fonts' ) }
+						>
+							{ __( 'Manage fonts' ) }
+						</Button>
+					</>
 				) : (
 					<>
 						{ __( 'No fonts installed.' ) }
--- a/packages/edit-site/src/components/global-styles/font-families.js
+++ b/packages/edit-site/src/components/global-styles/font-families.js
@@ -7,11 +7,7 @@ import {
 	__experimentalVStack as VStack,
 	Button,
 } from '@wordpress/components';
-<<<<<<< HEAD
 import { privateApis as blockEditorPrivateApis } from '@wordpress/block-editor';
-import { settings } from '@wordpress/icons';
-=======
->>>>>>> 72a59c77a0 (Replace Manage fonts icon button with visible text button.)
 import { useContext } from '@wordpress/element';
 
 /**
@@ -73,7 +69,9 @@ function FontFamilies() {
 						<Button
 							className="edit-site-global-styles-font-families__manage-fonts"
 							variant="secondary"
-							onClick={ () => setModalTabOpen( 'installed-fonts' ) }
+							onClick={ () =>
+								setModalTabOpen( 'installed-fonts' )
+							}
 						>
 							{ __( 'Manage fonts' ) }
 						</Button>
--- a/packages/edit-site/src/components/global-styles/font-families.js
+++ b/packages/edit-site/src/components/global-styles/font-families.js
@@ -69,6 +69,7 @@ function FontFamilies() {
 						<Button
 							className="edit-site-global-styles-font-families__manage-fonts"
 							variant="secondary"
+							__next40pxDefaultSize
 							onClick={ () =>
 								setModalTabOpen( 'installed-fonts' )
 							}
@@ -82,6 +83,7 @@ function FontFamilies() {
 						<Button
 							className="edit-site-global-styles-font-families__add-fonts"
 							variant="secondary"
+							__next40pxDefaultSize
 							onClick={ () => setModalTabOpen( 'upload-fonts' ) }
 						>
 							{ __( 'Add fonts' ) }
--- a/packages/edit-site/src/components/global-styles/style.scss
+++ b/packages/edit-site/src/components/global-styles/style.scss
@@ -41,7 +41,8 @@
 	color: $gray-700;
 }
 
-.edit-site-global-styles-font-families__add-fonts {
+.edit-site-global-styles-font-families__add-fonts,
+.edit-site-global-styles-font-families__manage-fonts {
 	justify-content: center;
 }
 
--- a/test/e2e/specs/site-editor/font-library.spec.js
+++ b/test/e2e/specs/site-editor/font-library.spec.js
@@ -14,15 +14,15 @@ test.describe( 'Font Library', () => {
 			await editor.canvas.locator( 'body' ).click();
 		} );
 
-		test( 'should display the "Manage Fonts" icon', async ( { page } ) => {
+		test( 'should display the "Add fonts" button', async ( { page } ) => {
 			await page.getByRole( 'button', { name: 'Styles' } ).click();
 			await page
 				.getByRole( 'button', { name: 'Typography Styles' } )
 				.click();
-			const manageFontsIcon = page.getByRole( 'button', {
-				name: 'Manage Fonts',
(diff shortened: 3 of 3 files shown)


t-hamano (commented):
Personally, I prefer this approach because I think it will make more sense in the future when the font panel is grouped into two parts: user and theme (See [this comment](https://github.com/WordPress/gutenberg/issues/58403#issuecomment-1936054980)).

@WordPress/gutenberg-design I'd like to hear your thoughts.

t-hamano (commented):
I think everything about the code is correct.

My only suggestion is whether the button size should be 32px or 40px.

| 32px | 40px |
|--------|--------|
| image | image | 

However, changing the button size may not be necessary in this PR.

@WordPress/gutenberg-design Please let me know your feedback!

jameskoster:
I could go either way. _Generally_ these full-width buttons in the Inspector are 40px, but that might be overly prominent in this case, so 32px could work.

afercia:
Worth reminding that when there's no fonts installed, the current UI already shows an 'Add font' full-width button. Its height is 36 pixels. I'd rather make both buttons 40 pixels for consistency with the new size and other buttons like the ones to set a featured imager or add a background image.

Screenshot 2024-06-21 at 10 08 30

afercia:
Screenshot of the buttons using the 'next 40 pixels height', which is more consistent with other similar full-width buttons:


similar full width buttons

[screenshot updated]

t-hamano (approved):
LGTM! I also prefer 40px for consistency with other similar buttons.

richtabor:
The "Manage fonts" button is far too prominent in this iteration, especially accounting for the entire panel's contents—not just this one component within the panel, as the screenshots are focused on. 

It's adding unnecessary complexity to the UI, introducing another type of control to the already disjointed view: 

CleanShot 2024-06-24 at 15 26 57

I'm not opposed to an empty state, perhaps not unlike the featured image/background image buttons. 

But I do propose we revert go back to the settings icon, at least until there is a step forward that is congruent with the rest of the panel (and best-case uses an existing UX pattern).

afercia:
> But I do propose we revert go back to the settings icon,

I kindly disagree. The prevalent feedback from contributors on this issue agrees that the Manage fonts button needs to be more visible. The change is consistent with other buttons used in the UI, see screenshot above with a few examples. Repeated, though anecdotal, feedback from users reported they can't find the manage fonts button as the `settings` icon doesn't seem to be the best design to open a modal dialog with a large UI to manage all fonts.
Considerations about the excessive 'prominence' of a button seem to me more personal opinions rather than objective functional analysis. I'd argue the same argumentation should then be made for all the other buttons in the editor UI that use the same styling (see screenshot above) so I'm not sure I understand why _this specific case_ seems to be more disturbing than others.
Anyways, I wouldn't be opposed to _user test_ the two options. In such cases, a some professional A/B testing with a diverse group of users is the only way to validate a design choice.

richtabor:
> The change is consistent with other buttons used in the UI, see screenshot above with a few examples. 

> I'm not sure I understand why this specific case seems to be more disturbing than others.

It's consistent with parts of the UI that are not exactly related; the use of those combination of button styles is uncommon, while the placement against the MenuGroup is not used elsewhere. 

I don't think we should be introducing many new UI conventions; the editor is already complicated. 

> Anyways, I wouldn't be opposed to user test the two options. 

I'm fine with testing alternatives, but not a solution that fundamentally is not part of the editor design language. A list of choices, with a button all the way at the bottom is not ideal by any means. 

There's also a concern of placing the button at the bottom of the list of fonts. While this may look fine in the context of a few fonts (isolated in screenshots), any sites with more fonts quickly loose out.

richtabor:
>  feedback from users reported they can't find the manage fonts button as the settings icon doesn't seem to be the best design to open a modal dialog with a large UI to manage all fonts.

An "add" icon is probably a much more connective element (in the existing placement), even though it's more than just adding fonts—the primary use case is to add (or manage existing fonts).

afercia:
> A list of choices, with a button all the way at the bottom is not ideal by any means.

I'm open to try to understand _why_, if you can bring in some objective arguments beyond personal opinions.

A recap of the main points for this change:
- There's direct feedback from the [FSE Outreach Program](https://make.wordpress.org/test/2023/09/18/fse-program-testing-call-26-final-touches/#comment-3177) that users could not find a way to open the Manage fonts modal dialog because the `Aa` icon was little understandable.
- The `Aa` icon was changed to a `settings` icon, which didn't help much. Still small, still without visible label.
- Anecdotally, I collected feedback from users in various events, including local meetups, reporting the settings icon button to open the fonts modal dialog is little discoverable.
- I'd argue the `settings` icon isn't appropriate in the first place because _in the editor design language_ it's used for different purposes and never usd to open a modal dialog. I could find 5 occurrences of `icon={ settings }` in the codebaae:
  - Three of them are used to switch a control to an alternative version:
    - Use size preset / Set custom size for the `SpacingInputControl`.
    - Use size preset / Set custom size for the `FontSizePicker`.
    - Use custom size for the X/Y position of the `ShadowInputControl`.
  - The other two occurrences are used for a more generic concept of 'settings':
    - View options for the Data views.
    - Display settings for the Query block toolbar.
- Overall, I'd argue the `settings` icon is already used inconsistently which doesn't help with building a cohesive 'editor design language'. For the first three cases, as a user, I learn that it indicates a way to switch a control to an alternative version where I can enter custom values. For the last two cases, I would expect a similar functionality but no, they indicate something entirely different. 
- The Font Library was probably the most awaited feature in WordPress 6.5. I'm not sure that hiding it behind a small icon button is ideal _from a product management perspective_ in the first place.
- Icon buttons are a problem for accessibility. It's seven years the accessibility specialists in this project are repeating that icon buttons usage should be limited to only the cases where the UI doesn't provide enough space to use visible text. I'd greatly appreciate the design team to prioritize this feedback instead of keeping dismissing it.
- The items in the list of installed fonts are clickable and do open the modal dialog but they bring users to the sub-panel to manage that specific font. To go to the 'Library' main panel, users have to click the 'Back' button which is little discoverable and counterintuitive.
- A button with visible text is the most appropriate way to solve the original issue reported by users.
- There is already at least one precedent of a full width button that opens a modal dialog: the 'Explore all patterns' button in the Inserter. As such, the 'Manage fonts' button is consistent with an already existing pattern in the editor.
- Noting that the 'Explore all patterns' button is placed in an UI that is exactly 'A list of choices, with a button all the way at the bottom' so I'm not sure I understand why it's a valid pattern in that case and it's not for the manage fonts.


> While this may look fine in the context of a few fonts (isolated in screenshots), any sites with more fonts quickly loose out.

I'd argue that if users install dozens of fonts, they're doing it wrongly in the first place. Typography should be managed by people who know what they are doing and good typography should not use more than 3-4 font faces.

Lastly,
I would greatly appreciate to see design feedback in this project take users feedback more into consideration. Being in disagreement in a large open source project is totally fine and, in my personal opinion, disagreement is an enrichment as it brings into the project diverse thoughts, knowledge, expertise, that benefit everyone.
But
when direct feedback from users gets ignored, that's far from ideal.

richtabor:
> Noting that the 'Explore all patterns' button is placed in an UI that is exactly 'A list of choices, with a button all the way at the bottom' so I'm not sure I understand why it's a valid pattern in that case and it's not for the manage fonts.

These are not related by any other means, apart from the button colors: 
 
CleanShot 2024-06-26 at 08 27 15

> I'd argue that if users install dozens of fonts, they're doing it wrongly in the first place. 

We can't simply assume the UI should work within a set of constraints that may, or may not, but "wrong."

> I would greatly appreciate to see design feedback in this project take users feedback more into consideration. Being in disagreement in a large open source project is totally fine and, in my personal opinion, disagreement is an enrichment as it brings into the project diverse thoughts, knowledge, expertise, that benefit everyone.
But
when direct feedback from users gets ignored, that's far from ideal.

You're missing purpose here. The purpose is not to make every bit of UI within the editor compete for prominence. It's to apply prominence selectively, to help people to achieve their goals. 

UX-wise, we have a list of fonts—each of which open the font library, with that font selected (with the available tabs for each). Then we have a _bigger_ button directly below that list, in a fairly prominent style applied, to something that is not intended to be prominent. Sure, the button is more prominent—but is this helpful? Is it helpful adding yet another button to this view that already has plenty of varying UI shapes across it? 

Objectively, no, this change is not worth making this view more difficult to parse. Instead of seeing "Fonts", "Elements" and "Presets", you see a big blue "Manage fonts" button in the middle of it all.

afercia:
> [the purpose] ... is to apply prominence selectively, to help people to achieve their goals.

I totally agree. Direct feedback from users reports that they can't achieve their goals. They can't find how to open the Fonts modal dialog with the Library tab showing the main view. This change tries to solve the problem by making the related control more prominent.

I don't want to repeat all the arguments already mentioned in previous comments, as I think we're looping over the same points. We're in disagreement, which is totally fine. 

I'm all for further improvements. However, I'm not sure I understand how saying 'no' _without providing an alternative solution to a valid problem_ is any better for users. I'm totally open to alternative solutions, as long as they address the reported issue in a valid, accessible way. I'd encourage you to create a new issue to explore alternative solutions.

PR #60004, opened 2024-03-19
Close inspector on pattern category select

draganescu:
## What?

Close #59963

## Why?

> You don't need the inspector open when adding patterns; it's not relevant contextually and causes the site canvas to be quite small if the Inserter + pattern categories + the Inspector are all open (especially if zoomed out mode is engaged).

## How?

Mostly copy https://github.com/WordPress/gutenberg/pull/56773

## Testing Instructions

1. Go to the site editor
2. Open the block inspector
3. Open the inserter
4. Click the patterns tab
5. Choose a category
6. Notice the block inspector is closed

Initially I thought I'd have to add it to the post editor but then I saw we auto close the inspector the moment the inserter is open there. This should be unified too I think.

### Testing Instructions for Keyboard

N/A

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/107534/fb868414-7164-4a68-9cf5-2547edc0976f

PR #61529, opened 2024-05-09
Make `BlockPopover` component public

gigitux:
## What?

> [!NOTE]  
> This PR is just a draft to get feedback. If making sense to make the `BlockPopover` component public, I would suggest marking it as an unstable API and improving the documentation.

This PR makes the `BlockPopover` component public. 

## Why?

Currently, a consumer can add the Gutenberg toolbar block with the `BlockToolbar` component. Hoverwer, the `BlockToolbar` isn't flexible, and it is complex to hide/show different buttons and text. 

Making `BlockPopover` public allows a consumer to create a custom `BlockToolbar` and customize it according to application needs.

This PR could be part of https://github.com/WordPress/gutenberg/issues/53874.

You can check https://github.com/woocommerce/woocommerce/issues/47323 and https://github.com/woocommerce/woocommerce/pull/47322 for a real-use case.

PR #62130, opened 2024-05-30
Make onPatternCategorySelection private

ntsekouras:
## What?
Follow up of: https://github.com/WordPress/gutenberg/pull/60004#discussion_r1618867922
Alternative of: https://github.com/WordPress/gutenberg/pull/62110

This PR makes the newly added `__experimentalOnPatternCategorySelection` prop private and renames it to `onPatternCategorySelection`

## Why?
We shouldn't add experimental features/props with just a prefix, because it will end up a public API in WP.

## Testing Instructions
1. Open the document sidebar and then the inserter
2. go to patterns tab and select a pattern category
3. the right sidebar should close
4. No regressions are introduced

### DIFF
--- a/packages/block-editor/src/components/inserter/library.js
+++ b/packages/block-editor/src/components/inserter/library.js
@@ -7,7 +7,7 @@ import { forwardRef } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import InserterMenu from './menu';
+import { PrivateInserterMenu } from './menu';
 import { store as blockEditorStore } from '../../store';
 
 const noop = () => {};
@@ -23,7 +23,7 @@ function InserterLibrary(
 		__experimentalInitialTab,
 		__experimentalInitialCategory,
 		__experimentalFilterValue,
-		__experimentalOnPatternCategorySelection,
+		onPatternCategorySelection,
 		onSelect = noop,
 		shouldFocusBlock = false,
 		onClose,
@@ -43,7 +43,7 @@ function InserterLibrary(
 	);
 
 	return (
-		<InserterMenu
+		<PrivateInserterMenu
 			onSelect={ onSelect }
 			rootClientId={ destinationRootClientId }
 			clientId={ clientId }
@@ -52,9 +52,7 @@ function InserterLibrary(
 			showMostUsedBlocks={ showMostUsedBlocks }
 			__experimentalInsertionIndex={ __experimentalInsertionIndex }
 			__experimentalFilterValue={ __experimentalFilterValue }
-			__experimentalOnPatternCategorySelection={
-				__experimentalOnPatternCategorySelection
-			}
+			onPatternCategorySelection={ onPatternCategorySelection }
 			__experimentalInitialTab={ __experimentalInitialTab }
 			__experimentalInitialCategory={ __experimentalInitialCategory }
 			shouldFocusBlock={ shouldFocusBlock }
@@ -64,4 +62,16 @@ function InserterLibrary(
 	);
 }
 
-export default forwardRef( InserterLibrary );
+export const PrivateInserterLibrary = forwardRef( InserterLibrary );
+
+function PublicInserterLibrary( props, ref ) {
+	return (
+		<PrivateInserterLibrary
+			{ ...props }
+			onPatternCategorySelection={ undefined }
+			ref={ ref }
+		/>
+	);
+}
+
+export default forwardRef( PublicInserterLibrary );
--- a/packages/block-editor/src/components/inserter/menu.js
+++ b/packages/block-editor/src/components/inserter/menu.js
@@ -45,7 +45,7 @@ function InserterMenu(
 		showMostUsedBlocks,
 		__experimentalFilterValue = '',
 		shouldFocusBlock = true,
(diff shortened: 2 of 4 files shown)


Mamaduka:
@ntsekouras, the difference is only the name change, correct?

ntsekouras:
>@ntsekouras, the difference is only the name change, correct?

The name change and making the two components private.

Mamaduka (approved):
Work as expected ✅

I just wish it was easier to add private props to components.

ellatrix:
As an alternative to Symbols, you could create a uuid and use that as the prop name, and lock that.

packages/block-editor/src/components/inserter/menu.js
```diff
@@ -341,4 +341,16 @@ function InserterMenu(
 	);
 }
 
-export default forwardRef( InserterMenu );
+export const PrivateInserterMenu = forwardRef( InserterMenu );
+
+function PublicInserterMenu( props, ref ) {
```
draganescu, line 346: As far as I've seen the inserter menu component is:

- used exclusively in the block editor package
- used exclusively in the inserter component in the block editor package
- not specifically exported anywhere

What do we need the private component for?

packages/block-editor/src/components/inserter/menu.js
```diff
@@ -341,4 +341,16 @@ function InserterMenu(
 	);
 }
 
-export default forwardRef( InserterMenu );
+export const PrivateInserterMenu = forwardRef( InserterMenu );
```
draganescu, line 344: As far as I've seen the inserter menu component is:

- used exclusively in the block editor package
- used exclusively in the inserter component in the block editor package
- not specifically exported anywhere

What do we need the private component for?

packages/block-editor/src/components/inserter/library.js
```diff
@@ -7,7 +7,7 @@ import { forwardRef } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import InserterMenu from './menu';
+import { PrivateInserterMenu } from './menu';
```
draganescu, line 10: I wonder did you try to lock this? If you locked it and then used lock/unlock internally would it work for you? I may have stumbled on some lock / unlock bug in [here](https://github.com/WordPress/gutenberg/pull/62110/commits/3f35a97513523d640588963806b04b0e8c5bc4aa) but I am not sure if it's not some typo on my part.

packages/block-editor/src/components/inserter/library.js
```diff
@@ -64,4 +62,16 @@ function InserterLibrary(
...
 }
 
-export default forwardRef( InserterLibrary );
+export const PrivateInserterLibrary = forwardRef( InserterLibrary );
+
+function PublicInserterLibrary( props, ref ) {
+	return (
+		<PrivateInserterLibrary
+			{ ...props }
+			onPatternCategorySelection={ undefined }
```
draganescu, line 71: this is reset to `undefined` here but to `const NOOP` in the [private inserter menu](https://github.com/WordPress/gutenberg/pull/62130/files#diff-2745181e57f35c49a91ec773fd2b1740c75f87b0f74efb857dd7a08a90b064e7R350), which makes the [optional chaining](https://github.com/WordPress/gutenberg/pull/62130/files#diff-2745181e57f35c49a91ec773fd2b1740c75f87b0f74efb857dd7a08a90b064e7R131) maybe not needed.

One more reason to remove the private inserter menu component.

draganescu (commented):
This is great @ntsekouras - thank you for stepping in as time was of essence considering the approaching beta WP. Left some questions mostly from the experience in my previous PR.

packages/editor/src/components/inserter-sidebar/index.js
```diff
@@ -3,8 +3,8 @@
  */
 import { useDispatch, useSelect } from '@wordpress/data';
 import {
-	__experimentalLibrary as Library,
```
draganescu, line 6: We kept the export right? So we export an `__experimentalLibrary` which is less public than the locked one :) Should we stabilise the experimental export?

packages/block-editor/src/components/inserter/library.js
ntsekouras: Probably yes. I was just being defensive there.

packages/block-editor/src/components/inserter/library.js
ntsekouras: I think it works, but there is no need to unlock if used in the same package.

packages/block-editor/src/components/inserter/menu.js
ntsekouras: I thought about it too, and I think it would be fine to not create a private one. Sometimes though we can decide to make some existing components public ([recent example here](https://github.com/WordPress/gutenberg/pull/61529/)). With that in mind, it would be clear that some props are meant to remain private..

packages/editor/src/components/inserter-sidebar/index.js
ntsekouras: Yeah, we could stabilize it and add a deprecation warning.

PR #62132, opened 2024-05-30
Image Block: Conditionally Render Block Control Based on Component Presence

### LINKED ISSUE #62040: Image block: when override added, toolbar shows redundant separator
### Description

After setting an override in a synced pattern for an image block, I noticed the image block with the override shows a redundant separator in the block toolbar.

1. Create a synced pattern.
2. Add an image block.
3. Under Advanced in Block Settings, enable an override for the Image block.
4. Save and publish the synced pattern.
5. Create a new page or post.
6. Add the synced pattern with the override for the image block.
7. Select the image block in the synced pattern and notice the toolbar has an extra separator. 

Screenshot 2024-05-27 at 4 38 53 PM

### Environment info

- WP 6.5.3
- GB 18.4
- TT4
- Chrome Version 123.0.6312.107 
- MacOS 14.1.1 

Yes

Yes

amitraj2203:
## What?
Fixes: #62040 

## Why?
There is a redundant separator in the block toolbar of Image block after setting an override in a synced pattern with an Image block.

## How?
Conditionally render the block control based on the presence of internal components.

## Testing Instructions
1. Create a synced pattern.
2. Add an image block.
3. Under Advanced in Block Settings, enable an override for the Image block.
4. Save and publish the synced pattern.
5. Create a new page or post.
6. Add the synced pattern with the override for the image block.
7. Select the image block in the synced pattern.

## Screenshots or screencast 
image


packages/block-library/src/image/image.js
```diff
@@ -499,43 +499,55 @@ export default function Image( {
...
-							mediaUrl={ ( image && image.source_url ) || url }
-							mediaLink={ image && image.link }
-							linkTarget={ linkTarget }
-							linkClass={ linkClass }
-							rel={ rel }
-							showLightboxSetting={ showLightboxSetting }
-							lightboxEnabled={ lightboxChecked }
-							onSetLightbox={ onSetLightbox }
-							resetLightbox={ resetLightbox }
+			{ shouldRenderOtherControls && (
```
michalczaplinski: A better name might be `showBlockControls` because they belong to the `block` group. The `other` group is below in the component tree. 

Also, for consistency, we use `show` instead of `render` verb in variable names here to indicate that something is being shown (rendered).

michalczaplinski (changes requested):
Hi @amitraj2203! Thanks for the contribution, it fixes the issue perfectly fine!

Could we group the various conditionals together like: 

```js
const showUrlInput =
  isSingleSelected &&
  ! isEditingImage &&
  ! lockHrefControls &&
  ! lockUrlControls;

const showCoverControls = isSingleSelected && canInsertCover;

const showBlockControls =
  showUrlInput || allowCrop || showCoverControls;
```

and then use them to conditionally render the `ImageURLInputUI` and `ToolbarButton`.

This way, there's only one place to add additional conditions in the future so less risk that someone forgets to update both places

packages/block-library/src/image/image.js
amitraj2203: Hi @michalczaplinski, Thank you for reviewing the PR. I have made the changes as per your suggestions and will keep this approach in mind for future work.

michalczaplinski (approved):
Great work! 👍 Thanks for the quick response.

PR #62134, opened 2024-05-30
Navigation block: Check for insert_hooked_blocks_into_rest_response i…

ockham:
Related PR: https://github.com/WordPress/wordpress-develop/pull/6677

## What?
Adding another `has_filter` check before adding the Navigation block's filter to insert hooked blocks into the REST API response for a `wp_navigation` post type.

## Why?
The logic to add hooked blocks to a `wp_navigation` post type REST responses is being moved into Core, see https://github.com/WordPress/wordpress-develop/pull/6677.

Modeled after https://github.com/WordPress/gutenberg/pull/61903.

### DIFF
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -1674,9 +1674,15 @@ function block_core_navigation_insert_hooked_blocks_into_rest_response( $respons
 $rest_prepare_wp_navigation_core_callback = 'block_core_navigation_' . 'insert_hooked_blocks_into_rest_response';
 
 /*
- * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
- * that are not present in Gutenberg's WP 6.5 compatibility layer.
+ * Do not add the `block_core_navigation_insert_hooked_blocks_into_rest_response` filter in the following cases:
+ * - If Core has added the `insert_hooked_blocks_into_rest_response` filter already (WP >= 6.6);
+ * - or if the `set_ignored_hooked_blocks_metadata` function is unavailable (which is required for the filter to work. It was introduced by WP 6.5 but is not present in Gutenberg's WP 6.5 compatibility layer);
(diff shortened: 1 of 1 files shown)


packages/block-library/src/navigation/index.php
```diff
@@ -1674,9 +1674,15 @@ function block_core_navigation_insert_hooked_blocks_into_rest_response( $respons
...
- * Injection of hooked blocks into the Navigation block relies on some functions present in WP >= 6.5
- * that are not present in Gutenberg's WP 6.5 compatibility layer.
+ * Do not add the `block_core_navigation_insert_hooked_blocks_into_rest_response` filter in the following cases:
+ * - If Core has added the `insert_hooked_blocks_into_rest_response` filter already (WP >= 6.6);
+ * - or if the `set_ignored_hooked_blocks_metadata` function is unavailable (which is required for the filter to work. It was introduced by WP 6.5 but is not present in Gutenberg's WP 6.5 compatibility layer);
+ * - or if the `$rest_prepare_wp_navigation_core_callback` filter has already been added.
  */
-if ( function_exists( 'set_ignored_hooked_blocks_metadata' ) && ! has_filter( 'rest_prepare_wp_navigation', $rest_prepare_wp_navigation_core_callback ) ) {
+if (
+	! has_filter( 'rest_prepare_wp_navigation', 'insert_hooked_blocks_into_rest_response' ) &&
```
ockham, line 1683: I think the main question for this PR is if we're happy with this tentative name (`insert_hooked_blocks_into_rest_response`).

ockham:
I'll go ahead and land this, to unblock https://github.com/WordPress/wordpress-develop/pull/6677. It's very low-risk, as it only adds another condition to determine whether or not to add the `block_core_navigation_insert_hooked_blocks_into_rest_response` filter.

It's very similar in nature to https://github.com/WordPress/gutenberg/pull/61903, and will allow us to harmonize the hooked blocks injection code (moving it out of the Navigation block and into Core, see https://core.trac.wordpress.org/ticket/60759).

PR #62021, opened 2024-05-27
Pages sidebar: adds published & scheduled items

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/59659

## What?

Adds "Published" and "Scheduled" items to the Pages sidebar.

Captura de ecrã 2024-05-27, às 12 48 48

## Why?

It's part of the design goal https://github.com/WordPress/gutenberg/issues/59659

## How?

Declares the new items in the sidebar item list for the Pages page.

## Testing Instructions

Visit the Pages page and verify that the new items are present and work as expected.

Issue #61909, opened 2024-05-23
Make the ellipsis button more visible


afercia:
### Description

Previously:
https://github.com/WordPress/gutenberg/issues/42505
https://github.com/WordPress/gutenberg/pull/42555
See also https://github.com/WordPress/gutenberg/pull/61847

The ellipsis button is barely visible for low vision users. I am a low vision user and I struggle to visually identify this button, especially when it's used 'in isolation' as in: in a spot of the UI where it is surrounded by white space and there's no other borders or shapes to draw user's attention to the button.

The three vertical dots are really thin. This is an inherent issue in the design of the icon. Also, as I tried to explain in https://github.com/WordPress/gutenberg/pull/42555#issuecomment-1831734289, the context matters. Other shapes or borders around the button make a big difference.

The ellipsis button is now widely used in the editor so that the issue previously reported for teh templates list is nwo more impactful. 

To explain how the context matters, in https://github.com/WordPress/gutenberg/pull/42555#issuecomment-1699047208 I made 4 examples. Reporting them here:

Screenshot:

<img width="1472" alt="ellipsis button" src="https://github.com/WordPress/gutenberg/assets/1682452/e563903d-8cda-4e56-97fd-dcf754653eb0">

For users with vision impairments, these 4 cases are very different. The ability to perceive the ellipsis icon also depends on the context as in: surrounding controls or borders or other shapes that help with visually perceiving the button.

Based on my (admittedly anecdotal) experience:

1. In the editor top bar: there are other buttons before the the ellipsis. This helps 'driving' my vision and helps perceiving the ellipsis dots but still isn't that great.
2. In the block toolbar: the other buttons and more importantly the visible borders greatly help. I would say this is the case where I can better perceive the ellipsis.
3. Editor fixed bar at the top: this is an old screenshot but still useful to get an idea: although there are buttons and borders, the color of the borders is a lighter gray than the one used in the block toolbar. I struggle to perceive the ellipsis.
4. The old template list, now similar to the Data Views 'table view': this is the worst case. The ellipsis button is literally placed in the middle of white space with no other controls or borders around it. Also, it is on the extreme right side, which doesn't help. I really can't see this button unless I get at a very short distance from my screen. This usage is seriously problematic for any user with vision impairment. There's now many more places in the UI where the elipsis is usd this way, 'in isolation'.

How to remediate?

[...]

PR #62136, opened 2024-05-30
Pages data view: Update view icons

jameskoster:
## What?
Follow-up to https://github.com/WordPress/gutenberg/pull/62021. Update icons for pages data views:

* Published → `published` 🆕 
* Scheduled → `scheduled` 🆕 
* Drafts → `drafts` (updated icon)
* Pending → `pending` 🆕 
* Private → `notAllowed` 🆕 

## Why?
Currently 'All pages', 'Published', and 'Scheduled' views all use the same icon, which is inconsistent when other views ('Trash' and 'Drafts') use unique icons.

| Before | After |
| --- | --- |
|  Screenshot 2024-05-31 at 15 58 24 | Screenshot 2024-05-31 at 15 56 52 |


ntsekouras (approved):
LGTM, thanks!

richtabor:
CleanShot 2024-05-30 at 08 03 02

These icons appear to be using 0.25 pixels, causing aliasing.

jasmussen:
It might be worth finding also a "scheduled post" icon, rather than just a calendar. In In principle, all of these are pages, so they would benefit from sharing DNA with the basic "page" icon. Not a blocker, but probably a followup.

richtabor:
I tweaked the draft icons in in the Figma. 

Also, what do you think about a published pages icon like this: 

CleanShot 2024-05-30 at 08 35 31


> In In principle, all of these are pages, so they would benefit from sharing DNA with the basic "page" icon. 

I agree, with the published pages icon the calendar/time icons stand out a bit much.

jameskoster:
> In principle, all of these are pages, so they would benefit from sharing DNA with the basic "page" icon.

I left a comment in https://github.com/WordPress/gutenberg/pull/62138#issuecomment-2139790728 to push back on this slightly. I think the icon should communicate the status, not the post type, as the `trash` icon does. 

That's not to say that they can't share DNA. I'm just not convinced they need to fit the 'page' footprint. 

I'll explore some options.

jameskoster:
Some of the icons might need tweaking, but I think something like this could work:

status-icons

Figma [here](https://www.figma.com/design/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?node-id=91177-10465&t=SFogyCcdBpEuYjHf-1).

richtabor:
> Some of the icons might need tweaking, but I think something like this could work:

I don't mind this.

I lean towards these draft / pending status icons: 

CleanShot 2024-05-30 at 15 06 21

carolinan:
Please do not reuse the Styles panel icon for drafts or make the pending icon look like a menu button.

afercia:
I would kindly ask to not use a circle with the horizontal ellipsis. Rationale: in https://github.com/WordPress/gutenberg/issues/61909, which originates from a long standing issue, see #42505 and #42555, I proposed to improve the 'more' ellipsis icon to make it more visible for low vision users.

One of the identified improvements options is to add a circle around the vertical allipsis, like macOS Finder does.

The new circle with the horizontal ellipsis proposed in this PR would clearly conflict with that.

Besides the presence of the circle, I would say that having horizontal allipsis and vertical ellipsis icons would be confusing in the first place. Oveerall, to me, the ellipsis doesn't seem appropriate to communicate a 'draft' or 'pending' status.

afercia (dismissed):
Please see comment https://github.com/WordPress/gutenberg/pull/62136#issuecomment-2141695434

richtabor:
> Please do not reuse the Styles panel icon for drafts or make the pending icon look like a menu button.

This is the styles icon: 

CleanShot 2024-05-31 at 07 40 30

> the pending icon look like a menu button.

I'm not following.

richtabor:
> Oveerall, to me, the ellipsis doesn't seem appropriate to communicate a 'draft' or 'pending' status.

What would you suggest for a pending status icon?

jameskoster:
For me there's enough difference here:

Screenshot 2024-05-31 at 13 03 06


* 'Filled' sides are opposite
* 'Filled' section is inset in drafts icon
* Different footprint
* Different stroke width

That said we could potentially rotate the drafts version to introduce another difference.

---

For the pending icon, off the top of my head I don't think there are any instances of a _horizontal_ ellipsis. Does that alleviate any of the concerns? Here are a couple other options to consider:


icons

afercia:
> What would you suggest for a pending status icon?

@richtabor you're asking to the wrong person 😆 If it was up to me, I wouldn't use icons at all.

Rather, I'd consider that when it becomes difficult to identify a glyph that clearly represents a feature, that's probably a sign that a glyph shouldn't be used to start with. In this specific case I'm less concerned because these aren't icon-only controls. There is visible text. The icons only 'complement' the visible text.

I see many concepts here that can hardly be represented with icons. Tha makes me think whether there's the real need to 
use icons.

What is the value added by icons when they are unclear?

A few examples of unclarity:
- The difference between scheduled and pending is hard to communicate with icons.
- Anyways, the concept of 'pending' is more related to a pending approval or pending review rather than time.
- The checkmark icon has a completely opposite meaning in some countries and cultures.
- The half-filled circle (a.k.a. 'half moon') doesn't communicate anything useful to me other than the contrast control on old televisions (this applies to Styles as well).
- 'All Posts' should not use the same icon that is already used for the Verse block (this applies to other parts of the UI as well)
- Concept like 'Recently active' are very difficult to communicate with an icon.

Overall, I'd think there's way more important things to focus on in the editor UI and I wouldn't spend much effort on a problem that can't be solved.

afercia:
Overall, in my personal opinion, the real value of icons is when they actually help users to identify a feature or a control. as such, icons should:
- Have an universal meaning for all countries and cultures.
- Clearly represent the feature / control.
- Leverage visual memory so that users are able to skip text and go straight to the identified feature / control.
- More importantly: be a _limited set of icons_. There's no great point in forcing users to 'learn' dozens and dozens of new icons. Users should not be forced to learn new icons in the first place. A good user interface should only use icons that are already commonly used on teh web.

I'd like to mention that right now [the Icons package](https://github.com/WordPress/gutenberg/blob/c9df77197c5ad8723d0a564a68a00bd48db967b9/packages/icons/src/index.js) contains more than 240 icons. Two hundred forty. Honestly, I think the editor has gone _a little_ too far with icons.

richtabor:
When used consistently, icons help create a cohesive visual language, making interfaces more user-friendly.

richtabor:
I don't mind this direction here:

CleanShot 2024-05-31 at 09 48 06

jameskoster:
> I don't mind this direction here

Updated based on this.

carolinan:
> That said we could potentially rotate the drafts version to introduce another difference.

The rotation helps for me at least, now it does not immediately make me think of the styles sidebar.

afercia:
> When used consistently, icons help create a cohesive visual language, making interfaces more user-friendly.

This is obvious. But, I would argue it only works with _limited sets_ of icons. With this PR, the `icons` package now contains 287 icons. I'm not sure forcing users to learn such a huge amount of icons helps 'making interfaces more user-friendly'.
Re: consistency, I'm not sure it is ideal throughout the UI as I see cases where the same icon is used for entirely different features. Anyways, I do realize this PR is not the right place for a broader discussion about icons.

hanneslsm:
I like the new icons, but the one for drafts also tipped me off.

How about using the last direction but for drafts the circles with the dashed border that was originally for Pending?
CleanShot 2024-06-04 at 23 26 31@2x

jasmussen (approved):
Before:

Screenshot 2024-06-18 at 14 42 21

After:

Screenshot 2024-06-18 at 14 44 33

This is a solid improvement, and the feedback in the thread looks to have been addressed.

jameskoster:
Merged this so that it is hopefully included in 6.6 🤞 

@hanneslsm I don't mind it. Could be worth gathering wider feedback in a separate PR if you have the time / inclination :)

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 88e3c937ab

PR #62138, opened 2024-05-30
Pages data view: Add Pending and Private views

jameskoster:
Follow-up to https://github.com/WordPress/gutenberg/pull/62021. I'm unsure why these weren't included in https://github.com/WordPress/gutenberg/issues/59659. I figured we can discuss in a PR.

## What?
Adds views for pending and private pages in the Pages data view.

## Why?
There are views for each of the other statuses (published, drafts, scheduled, trashed). It seems reasonable to include these too, particularly as there are such views in the wp-admin pages table.

Screenshot 2024-05-30 at 12 56 21

I think there's room to explore alternative icons in follow-ups during beta.

## Testing Instructions
1. Open the Pages data view
2. Observe the new 'Private' and 'Pending' views
3. Ensure they work as expected


jasmussen:
Question: can these be shown only if there are pages in those taxonomies? It becomes a bit much otherwise, and it seems like this could be a default behavior for any of these taxonomies, unless custom pinned. 

Separately, we should find a better icon for at least private, but probably also pending. As I suggested in the other PR, since these are page statuses, they should probably all share some DNA with the base "page" shape. 

* For private, can we show the basic page, with a lock on it?
* For pending, can we show the basic page, with a small clock on it?

jameskoster:
> As I suggested in the other PR, since these are page statuses, they should probably all share some DNA with the base "page" shape.

I'd like to push back on this a little. The icons are supposed to communicate the status, not the post type. The fact that they relate to pages should be contextually apparent. 'Combo icons' (e.g. Page + small lock) wouldn't scale very well—we'd eventually need dedicated icons for pending _posts_, pending _pages_, pending _comments_, pending _$post_type_s, and so on. Additionally it results in the icons visually blurring together and potentially failing the 'squint' test. In that sense, the draft/published icons may be the ones to revisit.

I do agree the icons need adjustment so that they're consistent. But I think that can potentially be done in a follow-up or during beta. The important thing to decide at the moment is whether or not to include these views for 6.6.

jasmussen:
The problem is, those icons were designed for different contexts, and dilute them by also applying to posts, IMO. I'm not voicing the strongest opinion in the world here, but for me, calendar is related to the _date_, not _scheduled posts or pages_. I realize that's a semantic detail, but it nevertheless feels meaningful to me.

jameskoster:
I agree with that. The status icons can share some DNA for sure, I'm just not convinced it should be based in the _page_ footprint. It could be something like small icons inside circles. Example from a very quick image search:

Screenshot 2024-05-30 at 16 29 25

My feeling is that we can adjust the icons during beta, maybe even tomorrow if we're quick. I don't think that should hold up adding this new feature though.

jameskoster:
@jasmussen how do you feel about merging this one, then addressing the icons in https://github.com/WordPress/gutenberg/pull/62136?

jasmussen (approved):

Issue #35294, opened 2021-10-01
[FR] Provide a onRender prop in ServerSideRender


pryley:
## What problem does this address?

Sometimes you need to trigger some javascript after a block is rendered with `ServerSideRender`.

Previously, the solution was to extend the ServerSideRender class. However with WordPress 5.8 this is no longer easily possible due to the changes made to allow Blocks to work outside of the post editor (i.e. sidebars and widgets).

See also: 
- https://github.com/WordPress/gutenberg/issues/17000
- https://github.com/WordPress/gutenberg/issues/32247
- https://github.com/WordPress/gutenberg/issues/7346

## What is your proposed solution?

My suggestion is to include a `onRender` prop to `ServerSideRender` and trigger it with the `useEffect` hook: 

```js
export default function ServerSideRender (props) {
    const {
        attributes,
        block,
        className,
        httpMethod = 'GET',
        onRender, // <- Here is the new "onRender" prop
        urlQueryArgs,
        EmptyResponsePlaceholder = DefaultEmptyResponsePlaceholder,
        ErrorResponsePlaceholder = DefaultErrorResponsePlaceholder,
        LoadingResponsePlaceholder = DefaultLoadingResponsePlaceholder,
    } = props;

    // ... 

    // Here is how the "onRender" prop is used
    useEffect(() => {
        if (onRender) {
            // Here we pass the response, the block, and attributes used for the response
            onRender(response, block, attributes);
        }
    }, [response]);

    // ... 

    return <RawHTML className={ className }>{ response }</RawHTML>;
}
```

And this is how you would use it in your block:

```js

const onRender = (response, block, attributes) => {
    if (!_.isEmpty(response) && !response.error) {
        console.log('the block has been rendered!');
    }
}

<ServerSideRender block="my/block" attributes={ attributes } onRender={ onRender }></ServerSideRender>
```

pryley:
@jasmussen since you worked on https://github.com/WordPress/gutenberg/issues/35027, perhaps you'd be open to looking at this since it's related to the SSR?

I'm currently using a custom ServerSideRender class which is identical to the latest commit with the loading indicator changes except that it also includes the change above...but I'd love to see this or something similar in core.

talldan:
@pryley What's the use case for such a callback?

pryley:
[...]

PR #62140, opened 2024-05-30
ServerSideRender: Fix data loading in development mode

Mamaduka:
## What?
Related #61943.

PR fixes the `ServerSideRender` component but is in development mode, where it was stuck in a loading state.

## Why?
React runs effects extra time in `StrictMode`, and the previous `isMounted` logic didn't account for this.

## How
* Set the `isMounted` ref value in the existing effect instead of relying on the initial value.
* I've also colocated loading state handlers - 0ff827771679cdec36b8191e70a9091c36896ebb

## Testing Instructions
1. Enabled development mode.
2. Open a post or page.
3. Add the RSS block and select the feed - `https://make.wordpress.org/design/`.
4. The block entries should be loaded correctly
5. Confirm that the component doesn't display Spinner when the response is immediate.
6. Confirm that on slow networks, Spinner is displayed.

<p>

```diff
diff --git .wp-env.json .wp-env.json
index 20d5597e54b..57d4e1a72d8 100644
--- .wp-env.json
+++ .wp-env.json
@@ -4,6 +4,9 @@
 	"themes": [ "./test/emptytheme" ],
 	"env": {
 		"tests": {
+			"config": {
+				"SCRIPT_DEBUG": true
+			},
 			"mappings": {
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",
diff --git packages/edit-post/src/index.js packages/edit-post/src/index.js
index 1e0b3fe7d4d..d80d3c161a6 100644
--- packages/edit-post/src/index.js
+++ packages/edit-post/src/index.js
@@ -7,7 +7,7 @@ import {
 	__experimentalRegisterExperimentalCoreBlocks,
 } from '@wordpress/block-library';
 import deprecated from '@wordpress/deprecated';
-import { createRoot } from '@wordpress/element';
+import { createRoot, StrictMode } from '@wordpress/element';
 import { dispatch, select } from '@wordpress/data';
 import { store as preferencesStore } from '@wordpress/preferences';
 import {
@@ -131,12 +131,14 @@ export function initializeEditor(
 	window.addEventListener( 'drop', ( e ) => e.preventDefault(), false );
 
 	root.render(
-		<Editor
-			settings={ settings }
-			postId={ postId }
-			postType={ postType }
-			initialEdits={ initialEdits }
-		/>
+		<StrictMode>
+			<Editor
+				settings={ settings }
+				postId={ postId }
+				postType={ postType }
+				initialEdits={ initialEdits }
+			/>
+		</StrictMode>
 	);
 
 	return root;
``` 

</p>
</details> 

Same.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/240569/271dee2d-c755-4576-a219-4ae10ffffdaf

### DIFF
--- a/packages/server-side-render/src/server-side-render.js
+++ b/packages/server-side-render/src/server-side-render.js
@@ -112,6 +112,11 @@ export default function ServerSideRender( props ) {
 
 		setIsLoading( true );
 
+		// Schedule showing the Spinner after 1 second.
+		const timeout = setTimeout( () => {
+			setShowLoader( true );
+		}, 1000 );
+
 		let sanitizedAttributes =
 			attributes &&
 			__experimentalSanitizeBlockAttributes( block, attributes );
@@ -165,6 +170,9 @@ export default function ServerSideRender( props ) {
 					fetchRequest === fetchRequestRef.current
 				) {
 					setIsLoading( false );
+					// Cancel the timeout to show the Spinner.
+					setShowLoader( false );
+					clearTimeout( timeout );
 				}
 			} ) );
 
@@ -191,21 +199,6 @@ export default function ServerSideRender( props ) {
 			debouncedFetchData();
 		}
 	} );
-
-	/**
-	 * Effect to handle showing the loading placeholder.
-	 * Show it only if there is no previous response or
-	 * the request takes more than one second.
-	 */
-	useEffect( () => {
-		if ( ! isLoading ) {
-			return;
-		}
-		const timeout = setTimeout( () => {
-			setShowLoader( true );
-		}, 1000 );
-		return () => clearTimeout( timeout );
-	}, [ isLoading ] );
 
 	const hasResponse = !! response;
 	const hasEmptyResponse = response === '';
--- a/packages/server-side-render/src/server-side-render.js
+++ b/packages/server-side-render/src/server-side-render.js
@@ -98,7 +98,7 @@ export default function ServerSideRender( props ) {
 		LoadingResponsePlaceholder = DefaultLoadingResponsePlaceholder,
 	} = props;
 
-	const isMountedRef = useRef( true );
+	const isMountedRef = useRef( false );
 	const [ showLoader, setShowLoader ] = useState( false );
 	const fetchRequestRef = useRef();
 	const [ response, setResponse ] = useState( null );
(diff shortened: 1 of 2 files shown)


packages/server-side-render/src/server-side-render.js
```diff
@@ -165,6 +170,9 @@ export default function ServerSideRender( props ) {
 					fetchRequest === fetchRequestRef.current
```
ellatrix, line 170: I guess to be completely correct, we should cancel the fetch request if the component unmounts? Then we also don't need a mount ref?

Mamaduka: I tried only to change what was necessary for the fix + minor colocation.

If you ask me, this whole component needs proper refactoring 😅

packages/server-side-render/src/server-side-render.js
```diff
@@ -165,6 +170,9 @@ export default function ServerSideRender( props ) {
...
 				) {
 					setIsLoading( false );
+					// Cancel the timeout to show the Spinner.
+					setShowLoader( false );
```
tyxla, line 174: ~Why do we need to `setShowLoader( false )` on cleanup? Feels odd to set the state on unmounting, since we won't be using that state.~
Scratch this, I was looking at the wrong place and thought this was running in an effect.

packages/server-side-render/src/server-side-render.js
```diff
@@ -98,7 +98,7 @@ export default function ServerSideRender( props ) {
 		LoadingResponsePlaceholder = DefaultLoadingResponsePlaceholder,
 	} = props;
 
-	const isMountedRef = useRef( true );
+	const isMountedRef = useRef( false );
```
tyxla, line 101: But it did mount already, didn't it? Unless I misunderstand something, this looks wrong - can you elaborate?

Mamaduka: Setting the initial value doesn't mean that the component was mounted. React can call a component (function) for other reasons without mounting it.

This is a common (anti)pattern when you want to run synchronization only after the initial mount.

```js
const didMount = useRef( false );
useEffect( () => {
   didMount.current = true;
   return () => { didMount.current = false };
}, [] );

Mamaduka: @tyxla, I think this is the third issue I fixed for a similar pattern. I explained why the original implementations were failing in development mode here: https://github.com/WordPress/gutenberg/pull/62141#issuecomment-2139551221.

tyxla: Ah, of course, got it. Wow, this component is messy 😉

Mamaduka: Yeah, it's bit of noodle soup 🍜

tyxla (approved):
This tests well for me. 👍  

As Ella suggested, there are some opportunities for cleanups and refactors, but they're not related to the PR so they can be addressed separately.

Mamaduka:
Thanks for testing and reviews, folks!

As for refactoring, @kevin940726 proposed hook API for this feature, which should give consumers more control. See https://github.com/WordPress/gutenberg/issues/35294#issuecomment-933214497.

PR #62146, opened 2024-05-30
Site Editor: Update to use the EditorInterface component from the editor package.

youknowriad:
Related #52632 
Similar to #62118

## What?

This is one of the latest pieces of the puzzle of the unification between post and site editors. This PR uses the EditorInterface from the editor package in the site editor instead of having a custom "InterfaceSkeleton" (the whole UI basically) usage. In #62118 I had updated the post editor already, this PR follows-up with the site editor changes.

## Testing Instructions

1- There should be no functional change or UI change at all but smoke testing the site editor is necessary to check for potential unintended regressions.


packages/edit-post/src/components/layout/index.js
```diff
@@ -293,6 +294,7 @@ function Layout( { initialPost } ) {
 			) }
 			<PostEditorMoreMenu />
 			<BackButton initialPost={ initialPost } />
+			<EditorSnackbars />
```
youknowriad, line 298: The snackbars need to be rendered only once in the site editor (or the post editor) so I left this responsibility out of the EditorCanvas component for now.

packages/edit-post/src/components/layout/style.scss
```diff
@@ -9,25 +9,12 @@
...
-}
-
-// Adjust the position of the notices when breadcrumbs are present
-.has-block-breadcrumbs {
-	.components-editor-notices__snackbar {
-		bottom: 40px;
-	}
+	bottom: 24px;
+	padding-left: 24px;
+	padding-right: 24px;
```
youknowriad, line 17: Simplified the position of the snackbars in the post editor (to be a fixed position like the site editor)

packages/edit-site/src/components/block-editor/style.scss
```diff
@@ -1,29 +1,3 @@
-// The button element easily inherits styles that are meant for the editor style.
-// These rules enhance the specificity to reduce that inheritance.
-// This is duplicated in visual-editor.
-.edit-site-block-editor__editor-styles-wrapper  .components-button {
```
youknowriad, line 4: This style is already in the editor package (so inherited in the site editor)

packages/edit-site/src/components/block-editor/style.scss
```diff
@@ -1,29 +1,3 @@
...
-	font-size: $default-font-size;
-	padding: 6px 12px;
-
-	&.is-tertiary,
-	&.has-icon {
-		padding: 6px;
-	}
-}
-
-.edit-site-visual-editor {
```
youknowriad, line 15: This style moved to the specific editor-canvas-container component. It's only useful there.

packages/edit-site/src/components/block-editor/use-editor-iframe-props.js
```diff
@@ -8,22 +8,17 @@ import clsx from 'clsx';
...
 /**
  * Internal dependencies
  */
 import { unlock } from '../../lock-unlock';
 import { store as editSiteStore } from '../../store';
 
-const { VisualEditor } = unlock( editorPrivateApis );
-
-function EditorCanvas( { settings } ) {
+export default function useEditorIframeProps() {
```
youknowriad, line 21: Mostly this enables the site editor to have the "click on canvas" behavior that is not present and not needed in the post editor. I wish it could be just an onClick prop to the EditorCanvas component but for now, I just left it as is and just passed down the iframeProps prop.

packages/edit-site/src/components/editor-canvas-container/index.js
```diff
@@ -45,15 +45,6 @@ function getEditorCanvasContainerTitle( view ) {
...
 }
 
-// Creates a private slot fill.
-const { createPrivateSlotFill } = unlock( componentsPrivateApis );
-const SLOT_FILL_NAME = 'EditSiteEditorCanvasContainerSlot';
-const {
-	privateKey,
-	Slot: EditorCanvasContainerSlot,
-	Fill: EditorCanvasContainerFill,
-} = createPrivateSlotFill( SLOT_FILL_NAME );
```
youknowriad, line 55: The creation and the rendering of the slot has moved to the editor package. So any editor (post or site editor) can decide to actually update the "content" are using a fill. (like we do for style book and style revisions). There are probably better ways to do implement this behavior but I didn't want to change too much here.

packages/editor/src/components/editor-canvas/index.js
```diff
@@ -47,19 +47,25 @@ const interfaceLabels = {
...
 	styles,
 	children,
 	forceIsDirty,
 	contentRef,
 	disableIframe,
 	autoFocus,
+	customSaveButton,
+	forceDisableBlockTools,
+	title,
+	iframeProps,
```
youknowriad: I wish we didn't need all these props to customize the behavior of the editor between post and site editors. But maybe we could iterate (remove some of these) in dedicated PRs.

packages/editor/src/private-apis.js
```diff
@@ -48,33 +34,19 @@ lock( privateApis, {
...
-	SavePublishPanels,
-	PostContentInformation,
-	PostLastEditedPanel,
 	ResizableEditor,
 	Sidebar,
-	TextEditor,
-	VisualEditor,
 
 	// This is a temporary private API while we're updating the site editor to use EditorProvider.
-	useAutoSwitchEditorSidebars,
```
youknowriad, line 77: These changes here just clean up the non-used private editor APIs (since now we're using the higher level components)

packages/edit-site/src/components/editor-canvas-container/index.js
ramonjd: > There are probably better ways to do implement this behavior but I didn't want to change too much here.

What do you think? Do you mean in terms of restricting the behavior to the post editor or finding a different way to swap editor content in and out?

ramonjd (commented):
I mainly did a smoke test in a web browser (not mobile app): 

✅  Site Editor: style revisions and style book functionality
✅  Site Editor: click on canvas from view mode (pages, templates, patterns)
✅  Both editors: snackbars (work and look okay in desktop/mobile viewports)
✅  Editor preferences work as expected across both editors

packages/edit-site/src/components/editor-canvas-container/index.js
youknowriad: There are a few thoughts that I have here:

 - I don't like that we're switching the canvas using a Slot, I'd prefer a children prop or something maybe.
 - I wonder if we're actually switching canvas here, I wonder if stylebook or style revisions are actually the same "editor", maybe it's the whole thing that should be swapped (so no need for Slot or anything). This is clearer for things like:
 
  - Why I have an inserter and the different block tools for stylebook and style revisions (we have a dedicated prop to remove these but the prop shouldn't exist in the first place) => in other words the header is completely different 
  - The "document" sidebar or the "block" sidebar don't make sense anymore. In these modes. We're focused on "global styles". => in other words, the left sidebar is completely different
  - There's no need for list view or inserter => right sidebar is completely different.

---
This is to say, in my mind, there are two valid paths forward:

 - Build a completely different UI (the whole thing, not just the canvas) for these modes. (Could share some very small things like save button and undo/redo)
 - Continue with the "editor" but instead of hacking things around like we're doing now, switch the "active entity" (currentPostType, and currentPostId) by passing a different "post" object to EditorProvider, and that post would be the "global styles" post. In which case, the "editor" would know that we're editing the global styles and would switch the UI accordingly.

Both of these can be valid, right now, I'm thinking the former is probably better but it's a very loosely held opinion.

ellatrix (approved):
Looks good codewise, and I see it's been smoke tested thoroughly.

t-hamano:
This PR seems to have caused the Style Book button in the navigation sidebar (not the global styles sidebar) to stop working. See #62204.

ramonjd:
> This PR seems to have caused the Style Book button in the navigation sidebar (not the global styles sidebar) to stop working. See https://github.com/WordPress/gutenberg/issues/62204.


Thanks @t-hamano 

I'll look into this. I should have caught it in testing, sorry!

packages/edit-site/src/components/editor-canvas-container/index.js
ramonjd: Thanks a lot for the thorough feedback!

> Build a completely different UI (the whole thing, not just the canvas) for these modes. (Could share some very small things like save button and undo/redo)

I could see this as an alternative, particularly if we could isolate it to the site editor. 

I guess we have access to context/edited post in the site editor [Editor](https://github.com/WordPress/gutenberg/blob/fcedde83fc5f29efb55f76058605570a0e633c60/packages/edit-site/src/components/editor/index.js) component.

Worth an experiment. 

I'll create an issue and put it on the backlog. 

Thank you!

ramonjd: The important feature, at least for global style revisions, was to have a "separate" editor view that could be edited independently of the main editor canvas.

So changes could be made to styles etc in the alt. canvas view, and when it closes, the user is returned to the main editor canvas in whatever state they left it.

PR #62149, opened 2024-05-30
CustomSelectControl: Fix `menuProps` mutation

tyxla:
## What?
This PR fixes a mutation that we're performing in the `CustomSelectControl` component and updates to use a new object instead.

## Why?
In this particular case, I'm fixing it to resolve an ESLint error that was raised by the React Compiler ESLint plugin in https://github.com/WordPress/gutenberg/pull/61788

## How?
Using a new object instead of mutating the existing one.

## Testing Instructions
Verify the v1 `CustomSelectControl` component still works.

None

## Screenshots or screencast 
None

### DIFF
--- a/packages/components/src/custom-select-control/index.js
+++ b/packages/components/src/custom-select-control/index.js
@@ -114,7 +114,7 @@ export default function CustomSelectControl( props ) {
 		return sprintf( __( 'Currently selected: %s' ), selectedItem.name );
 	}
 
-	const menuProps = getMenuProps( {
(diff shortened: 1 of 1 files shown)


tyxla:
I don't think this needs a CHANGELOG entry, but I'm happy to add one if convinced otherwise.

mirka (approved):
I have an urge to rework this a bit more drastically, but given the practical motivation for this PR and the fact that we'll be entirely replacing this with the v2 Legacy implementation soon, I think this is perfectly sufficient 👍

Issue #61135, opened 2024-04-26
Components: move border style utils to `block-editor` package


DaniGuardiola:
There are some border-related utilities marked as experimental being exported from the components package: https://github.com/WordPress/gutenberg/blob/d89e5a82fcb928819441f8371f9aca46eb8d3eee/packages/components/src/index.ts#L33-L35

Specifically:

- `hasSplitBorders`
- `isDefinedBorder`
- `isEmptyBorder`

This raised some alarms when making experimental components officially public, since they seem out of place, see comment: https://github.com/WordPress/gutenberg/pull/60837#discussion_r1570925958

These utilities are used in various different packages and places across the Gutenberg repo. Upon investigation, I believe a better home for them is probably the `block-editor` editor package. See the README: https://github.com/WordPress/gutenberg/blob/dcbccfb1ca51c560481c018a9c6f45b0ffd6aee9/packages/block-editor/README.md

This package contains a lot of style-related utilities for blocks, including the following:

- getColorClassName
- getColorObjectByAttributeValues
- getColorObjectByColorValue
- getComputedFluidTypographyValue
- getCustomValueFromPreset
- getFontSize
- getFontSizeClass
- getFontSizeObjectByValue
- getGradientSlugByValue
- getGradientValueBySlug
- getPxFromCssUnit
- getSpacingPresetCssVar
- getTypographyClassesAndStyles

cc @mirka

mirka:
Since some of this logic is also used in the component itself, the problem with moving it to `wp-block-editor` is that we'd either need to duplicate the logic, or we'd be importing functions from `wp-block-editor` into `wp-components` which is a violation of the dependency direction.

I think the order of preference should be:

1. Remove (deprecate) if possible
2. Incorporate into the component API if possible
3. Move to another package

I noticed that `isEmptyBorder` is not used in the app anymore, so that seems like a candidate for deprecation.

For the other two, have we considered passing them as metadata to the `onChange` callback? Perhaps as the second argument.

DaniGuardiola:
@mirka you're right, I hadn't realized that. Seems like I'll need to think deeper about this. Thanks for your input.

fullofcaffeine:
@DaniGuardiola and I discussed this today. Ideally, as Dani originally suggested, we'd move these experimental helpers to another package. We can't move them to `block-editor` because, as @mirka mentioned, it'd cause a circular dependency. It'd also be overkill to move only those three to a brand new package. 

[...]

Issue #59418, opened 2024-02-27
Components: Remove "Experimental" designation


mirka:
Related: [Tracking: stabilization of @wordpress/components props](https://github.com/WordPress/gutenberg/issues/65581)

## What problem does this address?

We used to designate a component as "Experimental" as a kind of backdoor to the back-compat policy, but since the policy was [made more explicit](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/#legacy-experimental-apis-plugin-only-apis-and-private-apis) and the Private API system was introduced, the "Experimental" has become meaningless. We will need to keep supporting them as regular, shipped components.

## What is your proposed solution?

1. Add non-prefixed exports for all the `__experimental` prefixed exports. (We will still need to keep the prefixed versions for back-compat.)
1. Update all documentation so they don't include `__experimental` in the import statements.
1. In Storybook, move all the "Components (Experimental)" items into the main "Components" group. (Use the Storybook redirect mechanism in #59181 to maintain permalinks.)

> [!NOTE]  
> This particular issue is only for the components themselves, not the `__experimental` props, which should be addressed separately.

## PRs:

- https://github.com/WordPress/gutenberg/pull/60913
- https://github.com/WordPress/gutenberg/pull/65586
- https://github.com/WordPress/gutenberg/pull/65475
- https://github.com/WordPress/gutenberg/pull/65469
- `Navigator`
  - [x] #60927
  - [x] #63317
  - [x] #64675
  - [x] #64786
  - [x] #64798
  - [x] #64777
  - [x] #65671
  - [x] #64613
  - [x] Remove old links from Block Editor handbook: https://github.com/WordPress/gutenberg/pull/64613#discussion_r1781424606
  - [x] https://github.com/WordPress/gutenberg/pull/65753
  - [x] https://github.com/WordPress/gutenberg/pull/65802
  - [x] Write dev note
- https://github.com/WordPress/gutenberg/pull/61014;
- #64951
- https://github.com/WordPress/gutenberg/pull/61016;
- https://github.com/WordPress/gutenberg/pull/61017;
- https://github.com/WordPress/gutenberg/pull/61018;
- https://github.com/WordPress/gutenberg/pull/61019;
- https://github.com/WordPress/gutenberg/pull/61020;
- https://github.com/WordPress/gutenberg/pull/61059;
- https://github.com/WordPress/gutenberg/pull/61060;
- https://github.com/WordPress/gutenberg/pull/61061;
[...]

PR #62155, opened 2024-05-30
Components: Stop re-exporting experimental utilities from `@wordpress/components`

fullofcaffeine:
Issue: https://github.com/WordPress/gutenberg/issues/61135
Related: https://github.com/WordPress/gutenberg/issues/59418

## What?

* Remove the re-exports for experimental utility functions from `@wordpress/components`;
* Copy the ones that are used outside wp-components near their app consumers, in a place that makes more sense to them (more info about that below).

## Why?

There are a few exports from `@wordpress/components` that smell, more specifically the ones listed in the aforementioned issue. These are utilities/helper functions that were developed as part of the `BorderBox` component but that ended up leaking to other packages in the Gutenberg app. The main problem here is that the wp-components package is not the best place to export shared utilities like this from.

This PR attempts to start solving that by first identifying where else in Gutenberg those utilities were being used, and copying them over to a tidy-up module in a package that makes more sense to them, which, as per @DaniGuardiola's initial analysis, would be `block-editor` (see the linked issue).

Later, we might DRY all these and other semantically related utilities (see the issue) into a single package that can be shared between multiple packages, including wp-components. This will be discussed and done in a follow-up.

## How?

* Copy `isDefinedBorder` and `hasSplitBorders` to a module in`block-ediotr/components/border-box-utils`;
* Copy necessary types, setup `tsconfig.json` to support the module;
* Stop exporting `isEmptyBorder` from the new module as it's not being used outside of wp-components and in this case is only used internally.
* Remove other functions that are not used outside of the wp-components context.

## Testing Instructions

* Tests should pass, plugin must compile without TS errors;
* Consumers that use the utilities should keep working without any mods.

## TODO

- [ ] Verify if there are other experimental utilities that need the same treatment

### DIFF
--- a/packages/block-editor/src/components/border-box-utils/index.ts
+++ b/packages/block-editor/src/components/border-box-utils/index.ts
@@ -1,0 +1,54 @@
+/**
+ * External dependencies
+ */
+import type { CSSProperties } from 'react';
+
+/**
+ * Internal dependencies
+ */
+import type {
+	Border,
+	AnyBorder,
+	Borders,
+	BorderProp,
+	BorderSide,
+} from './types';
+
+const sides: BorderSide[] = [ 'top', 'right', 'bottom', 'left' ];
+const borderProps: BorderProp[] = [ 'color', 'style', 'width' ];
+
+const isEmptyBorder = ( border?: Border ) => {
+	if ( ! border ) {
+		return true;
+	}
+	return ! borderProps.some( ( prop ) => border[ prop ] !== undefined );
+};
+
+export const isDefinedBorder = ( border: AnyBorder ) => {
+	// No border, no worries :)
+	if ( ! border ) {
+		return false;
+	}
+
+	// If we have individual borders per side within the border object we
+	// need to check whether any of those side borders have been set.
+	if ( hasSplitBorders( border ) ) {
+		const allSidesEmpty = sides.every( ( side ) =>
+			isEmptyBorder( ( border as Borders )[ side ] )
+		);
+
+		return ! allSidesEmpty;
+	}
+
+	// If we have a top-level border only, check if that is empty. e.g.
+	// { color: undefined, style: undefined, width: undefined }
+	// Border radius can still be set within the border object as it is
+	// handled separately.
+	return ! isEmptyBorder( border as Border );
+};
+
+export const hasSplitBorders = ( border: AnyBorder = {} ) => {
+	return Object.keys( border ).some(
+		( side ) => sides.indexOf( side as BorderSide ) !== -1
+	);
+};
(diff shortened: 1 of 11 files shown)


packages/components/src/index.native.js
```diff
@@ -35,6 +35,7 @@ export {
 
 export { default as __experimentalStyleProvider } from './style-provider';
 export { default as BaseControl } from './base-control';
+// @todo review this RN export
 export { hasSplitBorders as __experimentalHasSplitBorders } from './border-box-control/utils';
```
fullofcaffeine, line 39: I don't know much about our RN setup. If we go with the approach in this PR, should we also remove this and copy it over closer to where the native is? I'm not sure if we should also remove it from RN at this point. I did search throughout GB, and it also looks like it's not being used by RN consumers in the codebase, but I'm not sure if it could be used by other third-party RN packages.

DaniGuardiola (commented):
Looks good to me so far. Ping when ready and I'll be happy to do a deeper review. Thanks!

DaniGuardiola:
Taking over.

packages/components/src/index.native.js
ciampo: @dcalhoun @WordPress/native-mobile can you please confirm that it's safe to remove this export? Thank you! 🙏

packages/block-editor/src/components/border-box-utils/index.ts
ciampo, line 1: Would the `packages/block-editor/src/utils` be a better folder instead of the `components` folder for these utils?

packages/block-editor/src/components/index.js
```diff
@@ -175,3 +175,5 @@ export { useBlockCommands } from './use-block-commands';
  * The following rename hint component can be removed in 6.4.
  */
 export { default as ReusableBlocksRenameHint } from './inserter/reusable-block-rename-hint';
+
+export * from './border-box-utils';
```
ciampo, line 179: Would the `packages/block-editor/src/utils` be a better folder instead of the `components` folder for these utils?

ciampo:
> Taking over.

This time you beat me to it! I'll be happily help with reviews on this one.

I'll also add a label re. leaving a dev note for devs who may be experiencing breakages, recommending they import those same utils from the `block-editor` package instead  — we can also discuss later and decided if it actually needs one.

packages/components/src/index.native.js
dcalhoun: 👋🏻 Thanks for the ping. 

The mobile unit tests are passing on the CI server, so that is a good sign. 😄 I did not encountering any issues after checking out the proposed changes and "smoke testing" the editor on an iPhone simulator. 

From quickly reviewing the related code, my perception is that `__experimentalHasSplitBorders` is not utilized by the native mobile editor. So, removing this export should likely be safe.

PR #62159, opened 2024-05-31
Remove extra bracket in the site editor root padding styles.

### LINKED ISSUE #62148: Site Editor: Unwanted top margin added to the first block inside flow layout container
### Description

This must have been introduced recently. It only happens when the Gutenberg plugin is activated and seems to happen only in the site editor. 

Usually, when the layout is constrained, the top margin is removed from the first block in a container.

CleanShot 2024-05-30 at 16 03 19@2x

However, if the container has the flow layout, the top margin doesn't get removed.

CleanShot 2024-05-30 at 16 06 51@2x

1. Open the site editor
2. Add a group block
3. Add a block in the group block
4. Set the layout of the group block to flow(unconstrained).
5. See the gap between the container and the first block in it.

_No response_

### Environment info

- WP 6.5.3
- Gutenberg trunk

Yes

Yes

tellthemachines:
## What?

Fixes #62148. There was an extra culty bracket at the end of the root padding styles that was causing the rule output immediately after it to break.

## Testing Instructions

1. Add a Group block with some children in the site editor.
2. Set the layout of the group block to flow(unconstrained).
3. Check that there's no top margin on the first child block of the Group.


andrewserong (approved):
Good catch! The top margin rule is applied correctly for me in the site editor:

image

LGTM! 🚀

ramonjd (approved):
LGTM

aaronrobertshaw (approved):
Thanks for catching this!

I could replicate the original issue and confirm this PR fixes the margins.

richtabor:
Thanks for the quick follow-up. Good stuff. ✨

PR #62160, opened 2024-05-31
Use data-wp-on-async directives in core blocks when handler does not need synchronous access to event

westonruter:
This is a follow-up to #61885 for #61634 which introduced `data-wp-async-on--[event]` and implemented it for the Image block. This PR leverages it for the other core blocks where appropriate. Namely, `data-wp-on-async--[event]` can be used instead of `data-wp-on--[event]` whenever the handler does not take an Event object or if the handler doesn't need to call `preventDefault()`, `stopPropagation` or `stopImmediatePropagation()` which are the synchronous APIs on the Event object.


cbravobernal (approved):
Thanks for updating it!
Tested and works as expected.

ellatrix:

cbravobernal:
It's a performance improvement, I'll leave it to your decision @ellatrix , @vcanales  as Editor leads.

Thanks for everything and feel free to remove the label.

ellatrix:

PR #62161, opened 2024-05-31
Align both "Design" pattern list panels

richtabor:
## What?
Make these two panels for changing the design of a template or template part the same, with "Design" as the panel title.

## Why?
They do the same thing. 

## Testing Instructions

1. Open the Site Editor.
2. Select the header template part.
3. See the "Design" panel. 
4. Click on the "Template" tab from the Inspector. 
5. See the "Design" panel. 

## Screenshots or screencast 

| Before  | After |
|CleanShot 2024-05-30 at 21 11 07|CleanShot 2024-05-30 at 21 10 34|

Compared to the existing "Design" panel: 
CleanShot 2024-05-30 at 21 11 51


jasmussen (approved):
Good detail.

PR #62165, opened 2024-05-31
Fix syncing of publish date between publish and post status panel

### LINKED ISSUE #62150: Publish flow: changing date/time doesn't update in both places
In testing this PR for 6.6 https://github.com/WordPress/gutenberg/pull/62070 I noticed a few bugs including that changing the date/time under Publish doesn't reflect properly when you switch to the Status view with the more compact TimePicker.

To replicate, change date/time under Publish then open status to see the scheduled doesn't reflect the time. 

https://github.com/WordPress/gutenberg/assets/26996883/f8d1bc4e-5de6-4f0d-afee-3ba089964874

^ Note this shows two bugs I'm about to open cc @ntsekouras

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/62150

In this [PR](https://github.com/WordPress/gutenberg/pull/62070) I missed testing properly the syncing in both directions between the publish date and post status panel, when we update the publish date. This is the fix for that.

## Testing Instructions
1. In a post change the publish date through the publish panel
2. Open the post status panel and select `scheduled` status
3. The TimePicker component is rendered and should show the edited publish date

### DIFF
--- a/packages/block-editor/src/components/publish-date-time-picker/index.js
+++ b/packages/block-editor/src/components/publish-date-time-picker/index.js
@@ -17,10 +17,18 @@ export function PublishDateTimePicker(
 		onChange,
 		showPopoverHeaderActions,
 		isCompact,
+		currentDate,
 		...additionalProps
 	},
 	ref
 ) {
+	const datePickerProps = {
+		startOfWeek: getSettings().l10n.startOfWeek,
+		onChange,
+		currentDate: isCompact ? undefined : currentDate,
+		currentTime: isCompact ? currentDate : undefined,
+		...additionalProps,
+	};
 	const DatePickerComponent = isCompact ? TimePicker : DateTimePicker;
 	return (
 		<div ref={ ref } className="block-editor-publish-date-time-picker">
@@ -38,11 +46,7 @@ export function PublishDateTimePicker(
 				}
 				onClose={ onClose }
 			/>
-			<DatePickerComponent
-				startOfWeek={ getSettings().l10n.startOfWeek }
-				onChange={ onChange }
-				{ ...additionalProps }
-			/>
+			<DatePickerComponent { ...datePickerProps } />
 		</div>
 	);
 }


packages/block-editor/src/components/publish-date-time-picker/index.js
```diff
@@ -17,10 +17,18 @@ export function PublishDateTimePicker(
...
+		currentDate,
 		...additionalProps
 	},
 	ref
 ) {
+	const datePickerProps = {
+		startOfWeek: getSettings().l10n.startOfWeek,
+		onChange,
+		currentDate: isCompact ? undefined : currentDate,
+		currentTime: isCompact ? currentDate : undefined,
```
youknowriad, line 29: It's not clear to me why isCompact impacts currentData prop? 

Do we want the date to be "controlled" outside the entity in the case of the post status. What's happening exactly here?

ntsekouras: TimePicker component has [different name ](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/date-time/time/index.tsx#L96)for the `currentDate` and is `currentTime`.

Mamaduka: Naming around the `TimePicker` component is confusing.

Mamaduka (approved):
The fix works as expected ✅

packages/block-editor/src/components/publish-date-time-picker/index.js
youknowriad: ah ok then this fix makes sense :P 

Both of these components should just use value onChange to be honest.

oandregal (approved):
Changes look good to me.

I've noticed this component is also used in the post date block, which I've also tested and works as expected.

https://github.com/WordPress/gutenberg/assets/583546/caf3e877-45e9-4808-9f8b-6b30c40dd790

oandregal:
I see this is blocked by the performance job, which is failing in other PRs as well (different test suite: post-editor, site-editor, front-end-classic-theme). Tested locally with the lastest wp-env (to detect if it was a change in core) but I cannot reproduce.

ntsekouras:
I'm going to merge this, because the changes here are not related and very straightforward, as it's needed to land today for 6.6.

PR #61579, opened 2024-05-10
Site Editor Hub: Simplify

youknowriad:
## What and Why?

Currently the site editor renders a "hub" with a site logo, site title and some actions. That hub is persistent and rendered as "fixed" on top of everything.

I think the main reason it's implemented that way is to be able to animate the site logo with go in and out of view mode. That said, I personally believe this is way too complex to accommodate for a small animation.

So this PR is kind of a provocation, how simple can it be if we don't have to implement the animation of the logo. It's not necessarily meant to be merged but at least I want to raise this discussion.

Also, coincidentally, this fixes #60875

Let me know what you think.

PR #62167, opened 2024-05-31
Distraction Free: Unify the header animation

youknowriad:
Related #52632 
Extracted from #61579 
Requirement for #62146 

## What?

This PR tweaks the DFM animation just a little bit in the post editor and also moves the appearance animation for the Header of the site editor to the InterfaceSkeleton component. The idea is that we now animate the header in a single place. (It's fine to keep its children animations though).

## Testing Instructions

1- Check the animation of DFM in the post editor is good enough (the site editor is still broken like trunk)
2- Check that the header appears properly (outside DFM) when clicking the frame in the site editor.

### DIFF
--- a/packages/edit-site/src/components/editor/index.js
+++ b/packages/edit-site/src/components/editor/index.js
@@ -7,16 +7,8 @@ import clsx from 'clsx';
  * WordPress dependencies
  */
 import { useDispatch, useSelect } from '@wordpress/data';
-import {
-	Notice,
-	__unstableAnimatePresence as AnimatePresence,
-	__unstableMotion as motion,
-} from '@wordpress/components';
-import {
-	useInstanceId,
-	useViewportMatch,
-	useReducedMotion,
-} from '@wordpress/compose';
+import { Notice } from '@wordpress/components';
+import { useInstanceId, useViewportMatch } from '@wordpress/compose';
 import { store as preferencesStore } from '@wordpress/preferences';
 import {
 	BlockBreadcrumb,
@@ -80,8 +72,6 @@ const interfaceLabels = {
 	header: __( 'Editor top bar' ),
 };
 
-const ANIMATION_DURATION = 0.25;
-
 export default function Editor( { isLoading } ) {
 	const {
 		record: editedPost,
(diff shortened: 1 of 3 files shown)


packages/edit-site/src/components/editor/index.js
```diff
@@ -292,37 +281,13 @@ export default function Editor( { isLoading } ) {
 							}
 						) }
 						header={
-							<AnimatePresence initial={ false }>
-								{ canvasMode === 'edit' && (
-									<motion.div
```
youknowriad, line 297: The main purpose of this PR is to move this animation within the "InterfaceSkeleton" component because the header is also animated there. So if these two animations are in two different places, they may cause conflicts (for instance the hover animation within the skeleton might trigger while the other animation is in progress...)

jasmussen:
It's possible this is on my end, but I'm seeing an issue both in trunk, and this branch, where the top toolbar simply doesn't reappear when I move towards the top of the screen:
trunk

branch

youknowriad:
@jasmussen yes, that's the issue that the other PR solves #61579 but while solving that particular bug, we revealed another bug that was discovered on Firefox in the other PR by @draganescu 

So the current PR is actually a fix for that hidden Firefox bug but you can consider it a refactoring PR for now:

 - Check the DFM animation in post editor
 - Check the frame/header animation in the site editor with DFM disabled.

jasmussen:
Oh yes, thanks for restating that. Still caffeeinating. Here's a before:

post editor before

And after:

post editor after

They are close, except in the "after", the actual content of the toolbar animates down after a bit of a delay. Can that happen at the same time as the fade?

FWIW, I'd change both animations, so that the whole bar animates down, with no fade, rather than just the contents of the bar.

youknowriad:
I removed the delays for the child animations. It's still not 100% the same as before but feels close.

jasmussen:
Very close, animation looks good:

status

The bounce makes the contents of the toolbar briefly break the border, if you look closely. Easiest is probably to remove the bounce on the contents when they animate down.

The height of those contents also need to be reduced by 1 or 2 px, probably related to @richtabor's recent PR to unify the borders:
Screenshot 2024-05-31 at 09 29 02

youknowriad:
Ok, I removed the bouncing. I don't think I've touched the heights in this PR, so that's separate right?

draganescu (approved):
This is a welcome change helping the unification effort and reducing the complexity of the UI orchestration in DFM by bringing all animations in the same place. The result is nearly identical to what WP6.5 has so this is something to totally 🚢

jasmussen:
You're right, the half-pixel height offset is an issue in trunk as well. No blocker.

PR #62168, opened 2024-05-31
MediaUpload: Remove dialog markup on close

Mamaduka:
## What?
This is a follow-up to https://github.com/WordPress/gutenberg/pull/62108#discussion_r1619125848.

PR updates the `MediaUpload` to remove dialog markup when closed. 

## Why?
It stops leaving unused markup each time the Media Library is opened. See the related discussion on [Slack](https://wordpress.slack.com/archives/C02SX62S6/p1717055547327059) (it requires logic).

## Testing Instructions
1. Open a post or page.
2. Add an Image block.
3. Open the Media Library and close it.
4. Confirm that closing removes the Media Library markup from the DOM. Element ID starts with `__wp-uploader-`.

Same.

## Screenshots or screencast 
CleanShot 2024-05-29 at 19 37 13

### DIFF
--- a/packages/media-utils/src/components/media-upload/index.js
+++ b/packages/media-utils/src/components/media-upload/index.js
@@ -400,6 +400,8 @@ class MediaUpload extends Component {
 		if ( onClose ) {
 			onClose();
 		}
+
+		this.frame.detach();
 	}
 
 	updateCollection() {


packages/media-utils/src/components/media-upload/index.js
```diff
@@ -400,6 +400,8 @@ class MediaUpload extends Component {
 		if ( onClose ) {
 			onClose();
 		}
+
+		this.frame.detach();
```
joemcgill, line 404: This should work, but it looks like the `detach()` method calls close as well, so you might be able to bind the 'close' event directly to `detach` similar to how the `EditAttachments` frame works: https://github.com/WordPress/wordpress-develop/blob/trunk/src/js/media/views/frame/edit-attachments.js#L68

joemcgill (approved):
Approving with comments.

packages/media-utils/src/components/media-upload/index.js
Mamaduka: What would be the difference between binding `detach` directly to the event vs calling it from the binded callback?

I think they will be the same in this code base.

PR #62169, opened 2024-05-31
Inserter: show all blocks (alternative)

### LINKED ISSUE #60991: Improve Inserter experience when blocks cannot be inserted

ellatrix:
## What?

Fixes #60991
This is an alternative to #61873.

This PR changes `getInserterItems` to always return all items, but each item defines at which root client ID it can be inserted. 

Inside the inserter, when a block is selected, there's a list for blocks that can be inserted at the selection, and a list with remaining blocks. Since each of these blocks have a root client ID, we know where to show the insertion point. The index at which to insert is derived from the selection.

## Why?

See #60991.

### DIFF
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -84,16 +84,20 @@ export function BlockTypesTabPanel( {
 
 	return (
 		<>
-			{ showMostUsedBlocks && !! suggestedItems.length && (
-				<InserterPanel title={ _x( 'Most used', 'blocks' ) }>
-					<BlockTypesList
-						items={ suggestedItems }
-						onSelect={ onSelectItem }
-						onHover={ onHover }
-						label={ _x( 'Most used', 'blocks' ) }
-					/>
-				</InserterPanel>
-			) }
+			{ showMostUsedBlocks &&
+				// Only show the most used blocks if the total amount of block
+				// is larger than 1 row, otherwise it is not so useful.
+				items.length > 3 &&
+				!! suggestedItems.length && (
+					<InserterPanel title={ _x( 'Most used', 'blocks' ) }>
+						<BlockTypesList
+							items={ suggestedItems }
+							onSelect={ onSelectItem }
+							onHover={ onHover }
+							label={ _x( 'Most used', 'blocks' ) }
+						/>
+					</InserterPanel>
+				) }
 
 			{ currentlyRenderedCategories.map( ( category ) => {
 				const categoryItems = items.filter(
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -3,7 +3,9 @@
  */
 import { __, _x } from '@wordpress/i18n';
 import { useMemo, useEffect, forwardRef } from '@wordpress/element';
-import { pipe, useAsyncList } from '@wordpress/compose';
+import { useAsyncList } from '@wordpress/compose';
+import { getBlockType } from '@wordpress/blocks';
+import { useSelect } from '@wordpress/data';
 
 /**
  * Internal dependencies
@@ -14,6 +16,7 @@ import useBlockTypesState from './hooks/use-block-types-state';
 import InserterListbox from '../inserter-listbox';
 import { orderBy } from '../../utils/sorting';
 import InserterNoResults from './no-results';
+import { store as blockEditorStore } from '../../store';
 
 const getBlockNamespace = ( item ) => item.name.split( '/' )[ 0 ];
 
@@ -27,15 +30,14 @@ const MAX_SUGGESTED_ITEMS = 6;
  */
 const EMPTY_ARRAY = [];
 
-export function BlockTypesTab(
-	{ rootClientId, onInsert, onHover, showMostUsedBlocks },
-	ref
-) {
-	const [ items, categories, collections, onSelectItem ] = useBlockTypesState(
-		rootClientId,
-		onInsert
-	);
-
+export function BlockTypesTabPanel( {
+	items,
+	collections,
+	categories,
+	onSelectItem,
+	onHover,
+	showMostUsedBlocks,
+} ) {
 	const suggestedItems = useMemo( () => {
 		return orderBy( items, 'frecency', 'desc' ).slice(
 			0,
@@ -45,24 +47,6 @@ export function BlockTypesTab(
 
 	const uncategorizedItems = useMemo( () => {
 		return items.filter( ( item ) => ! item.category );
-	}, [ items ] );
-
-	const itemsPerCategory = useMemo( () => {
-		return pipe(
-			( itemList ) =>
-				itemList.filter(
-					( item ) => item.category && item.category !== 'reusable'
-				),
-			( itemList ) =>
-				itemList.reduce( ( acc, item ) => {
-					const { category } = item;
-					if ( ! acc[ category ] ) {
-						acc[ category ] = [];
-					}
-					acc[ category ].push( item );
-					return acc;
-				}, {} )
-		)( items );
 	}, [ items ] );
 
 	const itemsPerCollection = useMemo( () => {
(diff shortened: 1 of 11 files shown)


draganescu (commented):
I like the idea of this PR compared to https://github.com/WordPress/gutenberg/pull/61873 it's centralising the decision around where a block should be inserted right at the core of the process, when getting the list of blocks to be shown in the inserter.

There are some missing parts compared to the alternative though:

- the logic needs to be updated so that the insertion is always visually _after_ the currently selected block, just like normal insertion is
- for some reason the insertion point for blocks in `itemsRemaining` shows up in the site editor but not in the post editor
- this PR does not address the situation when parents cannot be used for insertion so children have to be tested 
  - when a page is edited in templated locked mode we cannot find an insertion point up the tree we need to go down the tree, in a way in which we don't specify a block by name (e.g. post-content). 
  - the original PR solves it 50% in the sense that the insertion happens in the 1st available spot, which is really not intuitive at all. Maybe we should not fix this yet?

ellatrix:
> the logic needs to be updated so that the insertion is always visually after the currently selected block, just like normal insertion is

Could you elaborate here? I think I am always inserting it after the selected block, I'm not sure what you mean :)

> when a page is edited in templated locked mode we cannot find an insertion point up the tree we need to go down the tree, in a way in which we don't specify a block by name (e.g. post-content).

Can we fall back to section root client ID instead of the absolute root `''`?

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -101,82 +85,146 @@ export function BlockTypesTab(
...
+					<>
+						<div
+							className="block-editor-inserter__panel-header"
+							style={ { display: 'block' } }
+						>
+							<h2 className="block-editor-inserter__panel-title">
+								{ __( 'Can be inserted into ' ) +
+									getBlockType( rootBlockName )?.title }
+							</h2>
+						</div>
```
ellatrix: Btw, this needs eyes from Design, I don't like the text and formatting here, it should be polished. Additionally, we need a heading in the "remaining block" section as well to make it clearer to the user.

draganescu: This is exemplified in the issue [here](https://github.com/WordPress/gutenberg/issues/60991#issuecomment-2114746204)

draganescu:
> I think I am always inserting it after the selected block, I'm not sure what you mean

Hmm, in my original testing the block got inserted _before_ the selection 🤷🏻 Now trying to record a video after latest updates it works fine ...

draganescu:
> Can we fall back to section root client ID instead of the absolute root ''?

That could work too yes!

packages/block-editor/src/components/inserter/block-types-tab.js
ellatrix: It's too simplistic imo. You can have a block selected where you can have many block across multiple categories than can be inserted, such as navigation. Imo it's better to have two sections, one with blocks that can be inserted right at the selection, and one for the remaining blocks that we'll insert in a parent.

packages/block-editor/src/components/inserter/hooks/use-insertion-point.js
```diff
@@ -154,9 +188,17 @@ function useInsertionPoint( {
...
-		( show ) => {
-			if ( show ) {
-				showInsertionPoint( destinationRootClientId, destinationIndex );
+		( item ) => {
+			if ( item?.hasOwnProperty( 'rootClientId' ) ) {
+				showInsertionPoint(
+					item.rootClientId,
+					getIndex( {
+						destinationRootClientId,
+						destinationIndex,
```
ellatrix, line 201: At some point we should remove all this `destination` stuff from and calculate it here inline, it would simplify things.

packages/block-editor/src/store/selectors.js
```diff
@@ -2039,13 +2039,52 @@ export const getInserterItems = createRegistrySelector( ( select ) =>
...
-					)
+					hasBlockSupport( blockType, 'inserter', true )
 				)
-				.map( buildBlockTypeInserterItem );
+				.map( buildBlockTypeInserterItem )
+				.reduce( ( accumulator, item ) => {
+					item.rootClientId = rootClientId;
+
+					while (
+						! canInsertBlockTypeUnmemoized(
```
ellatrix: Worth noting that, if you can select any block at the selected block, it doesn't impact the performance of this selector (and thus inserter opening) at all, because we already had this check and this loop will stop immediately. For other cases, it will loop up the tree until we find a place where it can be inserted. At some point we should investigate `getBlockEditingMode` because that's probably a very unperformant selector.

packages/block-editor/src/store/selectors.js
```diff
@@ -2039,13 +2039,52 @@ export const getInserterItems = createRegistrySelector( ( select ) =>
```
ellatrix, line 2040: `getInserterItems` is a public selector, so we might have to create a private selector and make changes to that one instead 🤔

draganescu: Why? Is this change bound to break anything? It should improve the UX across the board. The normal classic functionality remains as is.

packages/block-editor/src/store/selectors.js
```diff
@@ -2039,13 +2039,52 @@ export const getInserterItems = createRegistrySelector( ( select ) =>
...
+						} else {
+							const parentClientId = getBlockRootClientId(
+								state,
+								item.rootClientId
+							);
+							item.rootClientId = parentClientId;
+						}
+					}
+
+					// We could also add non insertable items and gray them out.
```
draganescu: That would add no value.

jasmussen:
Great work, this is a substantially better experience.

I’d recommend:

* Don’t show the subheading for the active selected block. It doesn’t feel needed in practice — the immediate rows of blocks are sorted there because they are relevant to what you have selected. Everything else gets the inserter line to show they’ll be inserted after.
* Fix the border separator between the in-block suggestions, and the after-block selections.
* Don’t show patterns in the blocks tab.

With that in place I think we can ship this.

packages/block-editor/src/store/selectors.js
ellatrix: It does if you have blocks installed or created and you don't know why they're not there. But we can talk about this after.

packages/block-editor/src/store/selectors.js
ellatrix: If you use `getInserterItems` it will now show all blocks, but with a `rootClientId` property, so the handling is different, technically a breaking change.

packages/block-editor/src/components/inserter/block-types-tab.js
```diff
@@ -27,15 +27,14 @@ const MAX_SUGGESTED_ITEMS = 6;
...
-export function BlockTypesTab(
-	{ rootClientId, onInsert, onHover, showMostUsedBlocks },
-	ref
-) {
-	const [ items, categories, collections, onSelectItem ] = useBlockTypesState(
-		rootClientId,
-		onInsert
-	);
-
+export function BlockTypesTabPanel( {
```
draganescu, line 30: This is added because we now split the blocks tab in two panels, one for blocks inserted in the current selection insertion point one that can be inserted in the nearest available insertion point relative to the current selection.

packages/block-editor/src/store/selectors.js
ellatrix: Made it a private option in [9a3f379](https://github.com/WordPress/gutenberg/pull/62169/commits/9a3f3791afca5000ee447bb1de6a52991a7eae71)

packages/block-editor/src/store/selectors.js
```diff
@@ -2037,16 +2038,70 @@ export const getInserterItems = createRegistrySelector( ( select ) =>
...
-					canIncludeBlockTypeInInserter(
-						state,
-						blockType,
-						rootClientId
-					)
+					hasBlockSupport( blockType, 'inserter', true )
 				)
 				.map( buildBlockTypeInserterItem );
 
+			if ( options[ withRootClientIdOptionKey ] ) {
```
draganescu, line 2047: I think adding a comment of how the discovery algorithm works would be great to read in a couple of months :D

draganescu (approved):
The code simplifies the approach in the alternative PR (#61873) and from my reading it only adds the least needed amount of work when needed.  By using a private option there is no public API change so we can safely land this.

This still introduces patterns in search results for blocks that hid them previously, for instance in a pattern with overrides, but they were allowed in search results in `trunk` except they were filtered out b/c of no insertion point. Now they can be inserted so they show, it is not a change of behavior.

jameskoster:
There's some duplication if you have the 'most used blocks' preference toggled on:

Screenshot 2024-05-31 at 13 18 44

jameskoster:
This is an edge case, so we might address separately but; if you're editing a page with a template that contains no Content block, then the Inserter will still turn up empty. Ideally there should be some messaging explaining what's going on. Something like:

> This template does not display content; blocks cannot be inserted. To insert blocks try switching to a different template.

richtabor:
If Most Used is rendered for the allowed blocks, could we not render the second one? Here's the latest with the Nav block: 

CleanShot 2024-05-31 at 10 13 53

ellatrix:
@richtabor Then it will also disappear if you have a shorter allowed list

ellatrix:
Let's merge since we don't have much time left before RC

richtabor:
> @richtabor Then it will also disappear if you have a shorter allowed list. 

Right, but even with a long allowedList, there probably shouldn't ever be multiple "Most Used" suggestions. 

> Let's merge since we don't have much time left before RC

Sounds good, I think tweaking most used in those scenarios with lots of blocks could be a potential follow-up fix. 

The HR styling as well (double-border).

richtabor (approved):
A solid experience win. Thanks for everyone who pushed this forward.

jsnajdr:
This PR caused a pretty big regression in the `inserterHover` metric:
```
│ inserterHover        │ '44.18 ms +3.85% -5.18%'                 │ '2.06 ms +10.68% -5.83%'   │ '2044.66%' │
```
It's both on this PR branch and on trunk after merging, so the regression is real, not some random breakdown.

The metric is calculating how long does it take to dispatch the `mouseover` and `mouseout` events. Maybe this makes the metric itself questionable, because it's measuring something that the user is not experiencing.

Mamaduka:
The `useBlockTypesState` also seems to return an unstable reference in the first `useSelect`. I'm not sure if it's related to performance regression.

## Screenshot
CleanShot 2024-06-03 at 16 58 49

youknowriad:
Looks like this impacted the "inserter hover" metric.

ellatrix:
@youknowriad You're late to the party, it's already been fixed 😛 https://github.com/WordPress/gutenberg/pull/62263#issuecomment-2146941131

youknowriad:
haha :) I wish all perf regressions were like that :P (fixed before I even notice them :P )

geriux:
Hey there 👋 We are investigating an issue on the native editor with the Social icons block, when tapping on the inline inserter, all of the blocks are shown with the social icons:

Inserter first blocks|After scrolling
-|-
<kbd></kbd>|<kbd></kbd>

When testing the Web editor with these changes, when tapping on "Browse all blocks" all blocks are available as well (although after the social icons). Is this expected?

### Web editor (Before)

https://github.com/WordPress/gutenberg/assets/4885740/67282890-2ec6-4a13-a3ed-aa6f217deb1c

### Web editor (After)

https://github.com/WordPress/gutenberg/assets/4885740/860d9329-ae67-4b4e-b445-f6b6a04b7394

Confirming if it is expected would help us move forward with a fix on the native side. Thank you!

geriux:
After going through some of the linked issues, it looks like it is indeed the desired behavior 👍 We'll work on how we sort the blocks in the inserter. Sorry for the noise!

Issue #56881, opened 2023-12-07
Disable 'auto-draft' to 'draft' status change during initial autosave


boonebgorges:
## What problem does this address?

Currently, if you begin authoring a post at post-new.php, a post object is created in the database with `post_status=auto-draft`. If you then leave the window idle for a certain interval, a request is sent to the 'autosaves' endpoint. As part of this request, the Block Editor changes the `post_status` from 'auto-draft' to 'draft'. See https://github.com/WordPress/gutenberg/blob/f6adf1e98b7e49cea4c95281ad225128b94e4e13/packages/core-data/src/actions.js#L602

This behavior feels odd. Before the block editor, it was generally safe to assume that 'auto-draft' items were items that were created without the user pressing any kind of 'save' or 'publish' button, while 'draft' items were those where a user explicitly performed a 'save' action. I'm currently working on a plugin that tries to leverage this distinction. But with the behavior described above, the distinction no longer stands.

I was able to blame the change back to https://github.com/WordPress/gutenberg/commit/4944315cac156c78313121657dff9d58ce8f4d3f, where the status change was lumped in with a number of other architectural changes to the way that core data is managed. As part of that PR, there was a brief discussion about whether this post-specific logic belongs in `core-data`: https://github.com/WordPress/gutenberg/pull/16932/files#r316549679 The assumption there is this behavior makes sense for posts, but not for other content types. I would push a bit further and ask two questions:

1. Does it, in fact, make sense for posts? As noted above, it erases a meaningful distinction between 'draft' and 'auto-draft', which could be leveraged in various ways (for instance, by ignoring 'auto-draft' items when making post queries)
2. Could we introduce a way to prevent the behavior?

## What is your proposed solution?

I don't have a specific proposal, but I would like for it to be possible to prevent the `status` from being switched to `'draft'` when Gutenberg sends its 'autosaves' request. Ideally, it'd be possible to intervene here in a general way; in my particular case, it'd be specific to a `post_type`.

For the time being I can probably work around this by intervening on the server level, perhaps as part of the `WP_REST_Autosaves_Controller` logic. But this is likely to result in some oddness in the client, for example in the way that the client `location` changes to `post.php` after the successful autosave. So, in an ideal world, it would be possible to do something in the client.

Even if this can't be addressed, I'd be grateful if someone could point to more background information on why this change in `status` was first introduced. Thanks in advance for any thoughts!

PR #62171, opened 2024-05-31
Do not auto save post status changes

ntsekouras:
## What?
Found by @Mamaduka [here](https://github.com/WordPress/gutenberg/pull/62070#pullrequestreview-2087319702):

> Unrelated note: I've noticed that a post might get accidentally Published/Scheduled if you're playing with the settings and the autosave functionality kicks in. I'm not sure if this is a new or old issue, but I would expect not to publish my posts accidentally.

With autosaves we only update very specific attributes (handled above my changes here). The exception was only for when creating a new post (initial status `auto-draft`) to make the status `draft`. Previously there was no specific `post status` panel to edit the attribute directly, so it was safe to just use the saved post status, but it seems we don't need to make that update in any other case.

## Steps to reproduce
1. Open a draft post.
2. Change status to Published or Scheduled.
3. Trigger an autosave - `window.wp.data.dispatch( 'core/editor' ).autosave()`.
4. Reload the page, and notice you'll get unsaved changes warning.
5. Notice that the post isn't a draft anymore.

## Testing Instructions
1. Everything should work as before with autosaves with the exception of being to change status

### DIFF
--- a/packages/core-data/src/actions.js
+++ b/packages/core-data/src/actions.js
@@ -598,10 +598,14 @@ export const saveEntityRecord =
 							return acc;
 						},
 						{
+							// Do not update the `status` if we have edited it when auto saving.
+							// It's very important to let the user explicitly save this change,
+							// because it can lead to unexpected results. An example would be to
+							// have a draft post and change the status to publish.
 							status:
 								data.status === 'auto-draft'
 									? 'draft'
-									: data.status,
+									: undefined,
 						}
 					);
 					updatedRecord = await __unstableFetch( {


Mamaduka:
Thanks, Nik!

This reminded me of this issue - #56881.

ntsekouras:
>This reminded me of this issue - https://github.com/WordPress/gutenberg/issues/56881.

Hm.. that's interesting and an easy change to do if we want to resolve that issue. Unfortunately I have no context at all about this logic, so I'll ping @WordPress/gutenberg-core for possible insights. I'm wondering if it had something to with the first versions of site editor that was using auto-drafts..

Having said that, I don't think this is a blocker for this PR.

jameskoster:
This is working for me 👍

Mamaduka (approved):
Definitely not a blocker. The issue can be resolved separately when we have more context around the current (auto-draft to draft) behavior.

cc @boonebgorges

PR #60367, opened 2024-04-02
Update 'Add template' screen to prefer template_name label instead of singular_name

Aljullu:
Fixes https://github.com/WordPress/gutenberg/issues/60283.

Related PR on WordPress: https://github.com/WordPress/wordpress-develop/pull/6344.

## What?

This PR makes it so the _Add template_ screen displays the `template_name` label for CPT and custom taxonomy templates, instead of relying on `singular_name`.

## Why?

This way, we allow plugins to add user-friendly names to templates related to their CPT/custom taxonomies.

## How?

This PR makes it so `template_name` takes priority over `singular_name` when possible. It also accounts for the cases of:

1. `template_name` being missing, in which case we fall back to `singular_name`, as it used to work.
2. `template_name` being duplicate, in which case the CPT/custom taxonomy slug is displayed.

## Testing Instructions

1. Add this code snippet to your site in order to create a Recipes post type and a Recipe category taxonomy (you can use the [Code Snippets](https://wordpress.org/plugins/code-snippets/) plugin if needed):

```PHP
add_action( 'init', function () {
	register_post_type(
		'recipe',
		array(
			'labels'       => array(
				'name'          => 'Recipes',
				'singular_name' => 'Recipe',
			),
			'public'       => true,
			'show_in_rest' => true,
		)
	);
	register_taxonomy(
		'recipe_cat',
		'recipe',
		array(
			'label'        => 'Categories',
			'labels'       => array(
				'name'          => 'Recipe categories',
				'singular_name' => 'Category',
			),
			'show_in_rest' => true,
		)
	);
} );
```

2. Go to Appearance > Editor > Templates > Add new and verify there are no regressions in the Add template screen: two items related to the code snippet should appear there: _Single item: Recipe_ and _Category (recipe_cat)_.

Add template screen with default labels

3. Now, modify the code snippet you created before, adding `template_name` to the post type and taxonomy:

```diff
add_action( 'init', function () {
	register_post_type(
		'recipe',
		array(
			'labels'       => array(
				'name'          => 'Recipes',
				'singular_name' => 'Recipe',
+				'template_name' => 'Single recipe',
			),
			'public'       => true,
			'show_in_rest' => true,
		)
	);
	register_taxonomy(
		'recipe_cat',
		'recipe',
		array(
			'label'        => 'Categories',
			'labels'       => array(
				'name'          => 'Recipe categories',
				'singular_name' => 'Category',
+				'template_name' => 'Recipes by category',
			),
			'show_in_rest' => true,
[...]

PR #60164, opened 2024-03-25
Text: Add text-wrap: pretty

jasmussen:
## What?

Alternative to, and followup, to #58728.

Adds `text-wrap: pretty;` to the `<Text>` component, to avoid typographic widows.

code sample showing the value

## Testing Instructions

Test using the inspector to edit the block card and try to add words to make a typographic widow happen. It shouldn't be possible:

block card

PR #58728, opened 2024-02-06
Try: text-wrap: pretty; across

jasmussen:
## What?

There's a new CSS property, `text-wrap: balance;`, which when applied to paragraphs of text makes it avoid unfortunate wrapping, specifically avoiding typographic widows. This PR applies the property very broadly — perhaps too broadly — so as to enable us to always avoid widows. The best way to demonstrate the effect is to show what happens if you intentionally edit a paragraph of text to add a typographic widow, as shown in this GIF:

text wrap before

Nothing happens, i.e. the typographic widow is permitted. The CSS property when applied, causes the wrapping to change, to avoid typographic widows, ensuring there are at least two words on the last line of text, as shown in this GIF:

text wrap after

Avoiding typographic widows is important to do editorially, but even in the best cases, there will be cases where they can appear, such as responsive contexts, or in translations. This property would prevent that from happen.

An open question is how broadly we want to apply this, however. This PR intentionally makes it super broad; should it be scoped instead to `p` and `h[n]` tags?

## Testing Instructions

Browse the editor, perhaps using the inspector try and force a typographic widow to appear, see that text re-wraps.

PR #62175, opened 2024-05-31
Align the template title to the center in the 'Add template' screen

Aljullu:
## What?

Since https://github.com/WordPress/gutenberg/pull/60367, it's possible that templates shown in the 'Add template' screen might have a long title. This PR makes sure multi-line titles appear center-aligned, keeping the same alignmend as single-line tites.

## Why?

In order to keep consistency between templates, making sure all of them look center-aligned.

## Testing Instructions

1. Make it so a long template title is rendered in the _Add template_ screen. I achieved so with this snippet:

```PHP
add_action( 'init', function () {
	register_post_type(
		'recipe',
		array(
			'public'       => true,
			'show_in_rest' => true,
			'labels'       => array(
				'name'          => 'Recipes',
				'singular_name' => 'Recipe',
				'template_name' => 'Very long recipe template name',
			),
		)
	);
} );
```
2. Go to Appearance > Editor > Templates > Add New Template.
3. Verify the _Very long recipe template name_ title is aligned to the center.

## Screenshots or screencast

Before | After
--- | ---
Screenshot showing the long title aligned to the left | Screenshot showing the long title centered

### DIFF
--- a/packages/edit-site/src/components/add-new-template/index.js
+++ b/packages/edit-site/src/components/add-new-template/index.js
@@ -128,6 +128,7 @@ function TemplateListItem( {
 					spacing={ 0 }
 				>
 					<Text
+						align="center"
 						weight={ 500 }
 						lineHeight={ 1.53846153846 } // 20px
 					>


jasmussen (approved):
We should find a way to add `text-wrap: pretty;` to this piece, so we avoid the typographic widow. But since that's an existing issue, thanks for the PR.

Aljullu:
Thanks for the review, @jasmussen!

> We should find a way to add `text-wrap: pretty;` to this piece, so we avoid the typographic widow. But since that's an existing issue, thanks for the PR.

Oh, it does in fact:

image

It's just that [`text-wrap: pretty`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap#browser_compatibility) doesn't seem to be supported in Firefox (which I used for the initial screenshot) and Safari. Do you think we should add `text-wrap: balance` as a fallback for browsers that don't support `text-wrap: pretty`? The change would be trivial and would fix this in Firefox and Safari, but wondering what are your thoughts on this as I see you worked recently in this field (https://github.com/WordPress/gutenberg/pull/58728 and https://github.com/WordPress/gutenberg/pull/60164).

(Sidenote: I'm going to merge this PR as-is and if we decide to add the `text-wrap` fallback I can do it in a separate PR)

PR #37763, opened 2022-01-06
Block API: Separate validation stage during block parsing

dmsnell:
## Description

When initially parsing and loading blocks we attempt to validate their parsed
attributes and apply auto-fixes when invalid. Previously this has taken place
inline inside of `parseRawBlock()`.

In this patch we're separating out the validation stage to aid in understanding
what actions take place during the validation and to allow `parseRawBlock()` to
avoid swapping the value of `parsedBlock` while running. With this patch each
separate stage in the parse is stored in a separate `const` variable which
enables easier debugging and comparison.

Additionaly the condition for printing warnings on block invalidation have been
updated to more clearly reflect the logic triggering them: that there were
problems during validation that were or were not resolved through the
block deprecation updating stage.

## How has this been tested?
There should be no visual or functional changes in this PR.
Passing unit tests and basic smoke tests should uncover any issues.
I have smoke-tested locally.

## Types of changes
Code refactor as preparatory PR while working on block invalidation system.
The goal of this work is to improve the handling of invalid blocks foremost by
making sure that the editor preserves markup it doesn't recognize, whereas
currently it frequently loses the content of invalid blocks when saving.

## Checklist:
- [ ] My code is tested.
- [ ] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [ ] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #62178, opened 2024-05-31
Parser: Update validateBlock to use fixedBlock

### LINKED ISSUE #62176: Parser: applyBlockValidation revalidating original block instead of fixed
When initially parsing and loading blocks we attempt to validate their parsed attributes and apply auto-fixes when invalid, before attempting to validate again after the fixes have been applied. This is done after we get a new raw block back after a round of fix attempts, which we can revalidate.

However, looking at this [section of code](https://github.com/WordPress/gutenberg/blob/trunk/packages/blocks/src/api/parser/index.js#L167-L183) it appears as though we are revalidating the original `unvalidatedBlock` instead of `fixedBlock` during the second execution of `validateBlock()`.

It looks to me like on the second run of `validateBlock()`, we should be passing `fixedBlock` as the first param instead.

tjcafferkey:
Fixes https://github.com/WordPress/gutenberg/issues/62176

## What?
Updated the second `validateBlock` execution to run against `fixedBlock` after we have attempted to apply fixes if it fails validation for the first time.

## Why?
It appears as though we are revalidating the original `unvalidatedBlock` instead of `fixedBlock` during the second execution of `validateBlock()`.

## How?
Updated the variable being passed to `validateBlock`

## Testing Instructions
1.

### DIFF
--- a/packages/blocks/src/api/parser/test/index.js
+++ b/packages/blocks/src/api/parser/test/index.js
@@ -50,6 +50,37 @@ describe( 'block parser', () => {
 	} );
 
 	describe( 'parseRawBlock', () => {
+		it( 'should apply className block validation fixes', () => {
+			registerBlockType( 'core/test-block', {
+				...defaultBlockSettings,
+				attributes: {
+					fruit: {
+						type: 'string',
+						source: 'text',
+						selector: 'div',
+					},
+				},
+				save: ( { attributes } ) => (
+					// eslint-disable-next-line react/no-unknown-property
+					<div class={ attributes.className }>
+						{ attributes.fruit }
+					</div>
+				),
+			} );
+
+			const block = parseRawBlock( {
+				blockName: 'core/test-block',
+				innerHTML: '<div class="custom-class">Bananas</div>',
+				attrs: { fruit: 'Bananas' },
+			} );
+
+			expect( block.name ).toEqual( 'core/test-block' );
+			expect( block.attributes ).toEqual( {
+				fruit: 'Bananas',
+				className: 'custom-class',
+			} );
+		} );
+
 		it( 'should create the requested block if it exists', () => {
 			registerBlockType( 'core/test-block', defaultBlockSettings );
 
--- a/packages/blocks/src/api/parser/test/index.js
+++ b/packages/blocks/src/api/parser/test/index.js
@@ -70,14 +70,15 @@ describe( 'block parser', () => {
 
 			const block = parseRawBlock( {
 				blockName: 'core/test-block',
-				innerHTML: '<div class="custom-class">Bananas</div>',
+				innerHTML:
+					'<div class="custom-class another-custom-class">Bananas</div>',
 				attrs: { fruit: 'Bananas' },
 			} );
 
 			expect( block.name ).toEqual( 'core/test-block' );
 			expect( block.attributes ).toEqual( {
 				fruit: 'Bananas',
-				className: 'custom-class',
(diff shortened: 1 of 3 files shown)


ockham:
Adding labels to mark as potential candidate for WP 6.6.

Mamaduka:
It would have been helpful to have testing instructions or testing coverage for this path, but based on [the original code](https://github.com/WordPress/gutenberg/pull/37763/files#diff-3629752a6442e4d3b18f333cd344e2d51128525fc5c921a90bcca1e1f5d7f591L203-L208), the fix looks correct.

gziolo:
> It would have been helpful to have testing instructions or testing coverage for this path, but based on [the original code](https://github.com/WordPress/gutenberg/pull/37763/files#diff-3629752a6442e4d3b18f333cd344e2d51128525fc5c921a90bcca1e1f5d7f591L203-L208), the fix looks correct.

If that was an issue with refactoring then we definitely should add a test that will prevent future regressions.

tjcafferkey:
@gziolo @Mamaduka I agree. I've added a test for the block validation fixes we are applying. Let me know if that's OK.

Mamaduka:
Thank you, @tjcafferkey!

The quick test shows that the new unit test also passes on `trunk`. Is that expected behavior? Can we showcase what this PR is actually fixing?

packages/blocks/src/api/parser/test/index.js
```diff
@@ -50,6 +50,38 @@ describe( 'block parser', () => {
...
+				attributes: {
+					fruit: {
+						type: 'string',
+						source: 'text',
+						selector: 'div',
+					},
+				},
+				save: ( { attributes } ) => (
+					// eslint-disable-next-line react/no-unknown-property
+					<div class={ attributes.className }>
```
youknowriad, line 65: If I'm not wrong there's no need to use attributes.className here as it should be done automatically no?

youknowriad:
This change is logical indeed. Makes me wonder how useful the built-in fixes are (This bug is around for some time now)

tjcafferkey:
>The quick test shows that the new unit test also passes on trunk. Is that expected behavior? Can we showcase what this PR is actually fixing?

Huh, @Mamaduka it looks like it's passing on `trunk` because `fixCustomClassname` (executed within `applyBuiltInValidationFixes`) is updating the object reference and returning it. So when it runs `validateBlock` on the second `unvalidatedBlock` it's actually been updated... and then returns `fixedBlock`.

The test would fail on `trunk` if `fixCustomClassname` created copy of `blockAttributes` to modify like I assumed it was doing when writing the test 😕

packages/blocks/src/api/parser/test/index.js
tjcafferkey: I thought so too, but apparently not.

youknowriad:
@tjcafferkey I guess that means that this PR actually doesn't do anything but clarify things. It's good anyway but at least we know that the fixes are useful and used :P

tjcafferkey:
@youknowriad ha yes. do you think I should update `fixCustomClassname` as part of this PR to not return the referenced object and instead create a new object to return?

youknowriad:
@tjcafferkey I think so, assuming there's no perf impact or anything, it's better to avoid mutating function arguments like that.

youknowriad (approved):

Mamaduka (approved):

dmsnell (commented):
A bit late, but this looks good. Thanks @tjcafferkey

PR #62183, opened 2024-05-31
Don't show non-existing and not supported attributes in block bindings panel

SantosGuillamot:
## What?
Remove the non existing and not supported attributes from the block bindings panel.

## Why?
It is a bit confusing because they won't work or do anything.

## How?
Just using the `canBindAttribute` util to decide if it should be deleted or not.

## Testing Instructions
* Add a paragraph connected to a non-existing attribute.
```
<!-- wp:paragraph {"metadata":{"bindings":{"fake":{"source":"core/post-meta","args":{"key":"text_field"}}}}} -->
<p>Default</p>
<!-- /wp:paragraph -->
```
* Check that in the block inspector there is no bindings panel.
* Add an image with some supported attributes and some non-supported attributes.
```
<!-- wp:image {"metadata":{"bindings":{"alt":{"source":"core/post-meta","args":{"key":"text_field"}},"caption":{"source":"core/post-meta","args":{"key":"text_field"}}}}} -->
<figure class="wp-block-image"><img alt=""/></figure>
<!-- /wp:image -->
```
* Check that the bindings panel only shows the alt attribute.

### DIFF
--- a/packages/block-editor/src/hooks/block-bindings.js
+++ b/packages/block-editor/src/hooks/block-bindings.js
@@ -14,10 +14,11 @@ import { useSelect } from '@wordpress/data';
 /**
  * Internal dependencies
  */
+import { canBindAttribute } from '../hooks/use-bindings-attributes';
 import { unlock } from '../lock-unlock';
 import InspectorControls from '../components/inspector-controls';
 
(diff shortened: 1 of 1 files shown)


cbravobernal (approved):
LGTM

PR #62184, opened 2024-05-31
Media & Text block: Fix nested Media & Text block media position issue with increased CSS specificity

### LINKED ISSUE #62047: Media & text block: When the media is on the right, the media position of a nested block does not work
### Description

This happens on WordPress 6.5.3 with and without Gutenberg active.
I have not tested with older versions of WordPress.

When you add a Media & Text block and set the media position to the **right** with the toolbar setting, and then insert a second media & text block in the block's content area: Then the media position option in the **second** media & text block does not work.

 -The media **starts** on the right side, but the default for the block is to position the media on the left, so the expectation is for the media to start on the left side even if it is nested.
- When using the media position option, the button looks activated, but the content area and media area does not switch places. Not in the editor or on the front.

Insert a Media & Text block.
Set the media position option in the block toolbar to the right.
Insert a second Media & Text block inside the first, and try to set its media position.

_No response_

### Environment info

_No response_

Yes

Yes

amitraj2203:
## What?
Fix: #62047 

## Why?
When we add a `Media & Text` block and set the media position to the right with the toolbar setting, and then insert a second `Media & Text` block in the block's content area: Then the media position option in the second `Media & Text` block does not work.

## How?
Adds `>` to relevant CSS selectors.

## Testing Instructions
1. Insert a Media & Text block.
2. Set the media position option in the block toolbar to the right.
3. Insert a second Media & Text block inside the first, and try to set its media position.


amitraj2203:
The same issue is also happening with the block's vertical alignment: if I change the outer block's vertical alignment to bottom then the inner block's vertical alignment doesn't work. So I have added the`>` selector to its respective class.

t-hamano (commented):
Thanks for the PR!

This PR works as expected, but we might also want to fix the issue where the layout affects nested blocks when "Stack on mobile" is enabled.

nested-stack-on-mobile

Just make the following changes:

```diff
diff --git a/packages/block-library/src/media-text/style.scss b/packages/block-library/src/media-text/style.scss
index d5da67b5e4..8a880fbab4 100644
--- a/packages/block-library/src/media-text/style.scss
+++ b/packages/block-library/src/media-text/style.scss
@@ -108,11 +108,11 @@
 @media (max-width: #{ ($break-small) }) {
        .wp-block-media-text.is-stacked-on-mobile {
                grid-template-columns: 100% !important;
-               .wp-block-media-text__media {
+               > .wp-block-media-text__media {
                        grid-column: 1;
                        grid-row: 1;
                }
-               .wp-block-media-text__content {
+               > .wp-block-media-text__content {
                        grid-column: 1;
                        grid-row: 2;
```

amitraj2203:
Thanks for reviewing the PR. I apologize for missing this 😅 
I have updated the code to address this issue now.

t-hamano (approved):
LGTM!

### Before

image

### After

image

ellatrix:

t-hamano:
This issue isn't new to WP6.6, but I think it's a great fix, so let's backport it to WP6.6.

ellatrix:

PR #62189, opened 2024-05-31
Update: Make changing order an action on the ellipsis menu.

jorgefilipecosta:
This PR implements a pending change suggested by @jameskoster at https://github.com/WordPress/gutenberg/pull/61918#issuecomment-2134863600 to make "reorder" an action on the ellipsis menu instead of an item on the page summary.

## Screenshots
Before:
Screenshot 2024-06-03 at 16 05 24

After:
Screenshot 2024-05-31 at 16 15 58

## Testing Instructions
Verified I could correctly reorder a page on the post editor, site editor, and pages list on dataviews.


jameskoster:
Thanks @jorgefilipecosta. I feel this better manages the prominence of this relatively obscure attribute. In the future it would be good to hide it altogether in favor of more intuitive experiences.

One small tweak; can we use the `small` size variant for the modal?

paaljoachim:
Hey @jorgefilipecosta 
Can you add in a before vs after screenshot please? 
(Hmmm....I am just not sure where it is today.)

jorgefilipecosta:
> Thanks @jorgefilipecosta. I feel this better manages the prominence of this relatively obscure attribute. In the future it would be good to hide it altogether in favor of more intuitive experiences.
> 
> One small tweak; can we use the `small` size variant for the modal?

Hi @jameskoster, thank you for reviewing this change, the small variant size was applied, let me know it it looks as expected.

jorgefilipecosta:
> Hey @jorgefilipecosta Can you add in a before vs after screenshot please? (Hmmm....I am just not sure where it is today.)

Hi @paaljoachim, the before image was added.

paaljoachim:
Ahhh.
Using Twenty Twenty Four and WordPress Version 6.5.3.
I see this:

Screenshot 2024-06-04 at 17 58 37



That means that a shift has happened between what I see in the screenshot that I shared and the above example Jorge shared. Moving the Reorder into the ellipsis menu seems like a good way to shorten down the long list of information seen.

jorgefilipecosta:
Hi @jameskoster, can you have another look and see if it already matches what is expected?

richtabor:
It's super obscure, and not helpful for most implementations of WordPress (any time you have a custom menu). 😅

packages/editor/src/components/page-attributes/order.js
```diff
@@ -71,74 +68,3 @@ export default function PageAttributesOrderWithChecks() {
...
-				) }
-				renderContent={ ( { onClose } ) => (
-					<div className="editor-post-order">
-						<InspectorPopoverHeader
-							title={ __( 'Order' ) }
-							onClose={ onClose }
-						/>
-						<div>
-							{ __(
-								'This attribute determines the order of pages in the Pages List block.'
```
richtabor, line 130: It's the "Page List" block.

richtabor: Let's just make this one paragraph, with the following: 

"Determines the order of pages in the Page List block. Pages with the same order value are sorted alphabetically. Negative order values are supported."

jorgefilipecosta: Hi @richtabor your feedback was applied, let me know if it is ok to merge this change.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -663,6 +664,116 @@ const renamePostAction = {
...
+		! Number.isInteger( Number( orderInput ) ) ||
+		orderInput?.trim?.() === '';
+	return (
+		<form onSubmit={ onReorder }>
+			<VStack
+				spacing="5"
+				style={ {
+					// editor styles are not loaded in dataviews so we use an inline style here.
+					maxWidth: 384 - 32 - 32 - 6 - 6,
+				} }
```
t-hamano: Is this style necessary? It seems to make the content width `372px` instead of the expected `384px ($modal-width-small)`.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -663,6 +664,116 @@ const renamePostAction = {
...
+			>
+				<div>
+					{ __(
+						'Determines the order of pages in the Page List block. Pages with the same order value are sorted alphabetically. Negative order values are supported.'
+					) }
```
t-hamano:
```suggestion
					{ __(
						'Determines the order of pages. Pages with the same order value are sorted alphabetically. Negative order values are supported.'
					) }
```

I think the order affects not only the Page List block, but also the order of pages on the dashboard.

So, how about changing it like this?

packages/editor/src/components/post-actions/actions.js
```diff
@@ -663,6 +664,116 @@ const renamePostAction = {
...
+						} }
+					>
+						{ __( 'Cancel' ) }
+					</Button>
+					<Button
+						__next40pxDefaultSize
+						variant="primary"
+						type="submit"
+						disabled={ saveIsDisabled }
+						__experimentalIsFocusable={ ! saveIsDisabled }
```
t-hamano:
```suggestion
						__experimentalIsFocusable
```

It's probably fine for this prop to always be enabled.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -663,6 +664,116 @@ const renamePostAction = {
...
+		},
+		[ postType ]
+	);
+
+	return useMemo(
+		() =>
+			supportsPageAttributes && {
+				id: 'reorder-pages',
+				label: __( 'Reorder' ),
+				RenderModal: ReorderModal,
```
t-hamano:
```suggestion
				isEligible( { status } ) {
					return status !== 'trash';
				},
				RenderModal: ReorderModal,
```

In keeping with the dashboard UI, it may be better not to display this action when their status is "trash"?

image

t-hamano (commented):
I have made some suggestions, but I would be glad to hear @richtabor's feedback again regarding [the text changes](https://github.com/WordPress/gutenberg/pull/62189#discussion_r1646032587).

packages/editor/src/components/post-actions/actions.js
jameskoster: > but also the order of pages on the dashboard.

Hmm, where do you mean? I think it's helpful to be explicit here... "The order of pages." is a bit ambiguous.

t-hamano: Sorry, my explanation was insufficient. I wanted to mention the following screen.

image

In the screenshot above, the order of "Page 2" has been changed to 1.

jameskoster: Oh yes, I see. I wonder if that's intentional because if you refresh then the `order` is no longer respected 🤔  Sorting by date (etc.) also ignores `order`.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -663,6 +664,116 @@ const renamePostAction = {
...
+			return !! postTypeObject?.supports?.[ 'page-attributes' ];
+		},
+		[ postType ]
+	);
+
+	return useMemo(
+		() =>
+			supportsPageAttributes && {
+				id: 'reorder-pages',
+				label: __( 'Reorder' ),
```
richtabor: Let's use the existing "Order" label here, as it maps to the control within the dialog as well.

packages/editor/src/components/page-attributes/order.js
richtabor: Yes, this reads better.

richtabor (commented):
I did expect "Order" to be a visible entry on the publish panel, once a value does exist for it. Otherwise there is no indicator that I have an order assigned. Was that intentional? 

Below, I have set an order value:

CleanShot 2024-07-01 at 08 48 02

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: The default sorting takes order into consideration, but if we sort by any field order is not taken into account.

jorgefilipecosta: I applied the suggestion.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Change the label to order.

jorgefilipecosta:
> I did expect "Order" to be a visible entry on the publish panel, once a value does exist for it. Otherwise there is no indicator that I have an order assigned. Was that intentional?

Hi @richtabor,
I made the order appear on the UI if there is a value set.

t-hamano:
> I did expect "Order" to be a visible entry on the publish panel, once a value does exist for it. Otherwise there is no indicator that I have an order assigned. Was that intentional?

I'm worried that this specification may cause accessibility problems.

1. Update the order number to something other than zero.
2. "Order: {number}" will be displayed in the panel.
3. Open the Order popover.
4. Change the order number back to zero.
5. At that moment the popover will disappear and lose focus.

The video below shows how with NDVA enabled, the screen reader loses focus when the order number is reset to zero on the keyboard.

https://github.com/WordPress/gutenberg/assets/54422211/e1881eda-48c2-4de0-8fed-ff3ed5f0f462

cc @richtabor

jameskoster:
Order is such an obscure attribute, and so rarely used, I don't know that we ever need to display it directly in the sidebar.

If you're familiar with the feature, you'll understand that any non-zero value = customized.

t-hamano:
> The video below shows how with NDVA enabled, the screen reader loses focus when the order number is reset to zero on the keyboard.

In my opinion, there are two possible approaches to get around this problem:

- Leave the Order action in the summary panel (meaning closing this PR 😥)
- Move the Order action to the elipsis menu and NOT display it in the summary panel

Do you have any good ideas?

jorgefilipecosta:
I guess we can go for "Move the Order action to the elipsis menu and NOT display it in the summary panel". @richtabor @WordPress/gutenberg-design would that change be ok?

jameskoster:
That approach would get my vote :)

richtabor:
> I guess we can go for "Move the Order action to the elipsis menu and NOT display it in the summary panel". @richtabor @WordPress/gutenberg-design would that change be ok?

I suppose that's the better of the two options.

jorgefilipecosta:
Thank you all for the insights I rebased the PR and applied the recommended path. I think this is ready for a last checkup and merge. cc: @jameskoster

jameskoster (approved):
It's working for me, and in terms of design this feels like an improvement on trunk given how seldom this feature is used.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -633,6 +634,120 @@ function useRenamePostAction( postType ) {
...
+						} }
+					>
+						{ __( 'Cancel' ) }
+					</Button>
+					<Button
+						__next40pxDefaultSize
+						variant="primary"
+						type="submit"
+						// eslint-disable-next-line no-restricted-syntax
+						disabled={ saveIsDisabled }
```
t-hamano:
```suggestion
						accessibleWhenDisabled
						disabled={ saveIsDisabled }
```
If you don't have a specific reason to ignore the rule, adding `accessibleWhenDisabled` might be a good idea.

packages/editor/src/components/post-actions/actions.js
t-hamano: This style causes form elements to shift in width.

| With maxWidth | Without maxWidth |
|--------|--------|
| image | image |

packages/editor/src/components/post-actions/style.scss
```diff
@@ -1,3 +1,7 @@
 .editor-action-modal {
 	z-index: z-index(".editor-action-modal");
 }
+
+.editor-action-modal__reorder-pages .components-modal__frame {
+	max-width: $modal-width-small;
+}
```
t-hamano: There is no element with this selector and it would be unnecessary.

t-hamano (approved):
I left some comments at the end but it's working fine for me too.

packages/editor/src/components/post-actions/style.scss
jorgefilipecosta: Good catch it was removed 👍

PR #62191, opened 2024-05-31
Update sidebar title + icon + site title alignment

jameskoster:
I'm unsure when this visual regression occurred, but the sidebar title has slipped out of alignment with the icons it often appears next to, and with the site title. It appears to be resting on a different plain to the icons, which is creating some tension between the `<` button and the title.

This PR addresses these details, illustrated below:

Screenshot 2024-05-31 at 17 01 01

### Testing Instructions
1. Open the Site Editor
2. Check the sidebar title and icons are centrally aligned along the X axis.
3. Check the sidebar title and site title are left aligned on the Y axis.


richtabor (approved):

t-hamano (commented):
Changing `alignment` to `center` is fine in most cases, but I understand that `flex-start` is used intentionally when the text is long.

For example, if you have a long navigation menu, it will be displayed as follows.

image

How about adjusting the top padding while maintaining `flex-start`? `padding: $grid-unit-05 * 0.5 0 0 0` seems well balanced to me.

image

jameskoster:
Oh yes, good point. Using padding is never going to be 100% perfect, but it's an improvement on trunk.

I think ideally we update these buttons to be 32px rather than the non-standard 36px, then set the line-height on the title to 32px as well. That would align everything nicely, but it's a bigger change—one to explore separately.

t-hamano (approved):
LGTM! Thanks for the update.

image

image

ellatrix:

ellatrix:

PR #58637, opened 2024-02-02
Escape as Select/Edit mode Toggle

jeryj:
In the editor, escape is a shortcut to switch to Select/Navigation mode , and then another Escape keypress resets selection to the top of the DOM. This always surprises me, and feels like a focus loss as I can't see where the focus is anymore. I like @jasmussen's recommendation [to turn Escape into a toggle between Select/Edit modes](https://github.com/WordPress/gutenberg/issues/47172#issuecomment-1457707443)

## What?

Use Escape as a toggle between Edit and Select modes.

## Why?

To have a more predictable experience. Escape doing two things (switching to select mode, then moving focus to the top of the DOM feels really off to me).

## How?

Add an isEscape check to the top of the onKeydown for the breadcrumb. This makes it operate very similarly to the Enter keypress to switch to Edit mode.

### Testing Instructions for Keyboard

1. In the block editor, press Escape to enter select mode
2. Press Escape again to switch back to edit mode

PR #62196, opened 2024-05-31
Escape moves focus to editor region when in select mode

jeryj:
An escape keypress previously [toggled the state between editing and select mode](https://github.com/WordPress/gutenberg/pull/58637/). This returns the behavior to the previous implementation of Escape clearing block focus. Now, when in select mode, escape focuses the editor region (showing the blue outline from region focus styles).

## What?

- In select mode, Escape selects the wrapping region
- Shows a blue border when the canvas is focused

## Why?

- Allow block deselection using Escape
- Show focus on canvas element

## How?

Clears block selection and focuses the region wrapper by: 
- Using the unselect shortcut (escape)
- Identify when this is happening via the navigation mode button focus (the only instance we want to capture this event) by passing a ref to the button
- If the escape key is emitted from this button, find the closest region wrapper for the block canvas by using the current ref or the iframe editor
- If we find the region, focus it
- If not, as it's not a requirement for the block editor to be wrapped in a region for third parties, then focus the editor canvas itself

### Escape to focus canvas wrapper
- Click a nested block
- Escape to enter select mode
- Escape to clear selected block and focus canvas
- repeat test using Preferences -> Show custom fields -> reload to test this works in both iframed and non-iframed versions of the editor

#### Region outline showing when you don't want it to
- Click a nested block
- Escape to enter select mode
- Click block sidebar area in bottom space
- Press Escape
- Blue outline should not show

#### Navigating regions
- Use the keyboard to navigate regions (⌃⌥N and ⌃⌥P)
- Blue outline should show when you want it and be removed as on trunk

### DIFF
--- a/packages/block-editor/src/components/block-tools/block-selection-button.js
+++ b/packages/block-editor/src/components/block-tools/block-selection-button.js
@@ -183,7 +183,7 @@ function BlockSelectionButton( { clientId, rootClientId } ) {
 		const navigateUp = ( isTab && isShift ) || isUp;
 		const navigateDown = ( isTab && ! isShift ) || isDown;
 		// Move out of current nesting level (no effect if at root level).
-		const navigateOut = isLeft || isEscape;
+		const navigateOut = isLeft;
 		// Move into next nesting level (no effect if the current block has no innerBlocks).
 		const navigateIn = isRight;
 
--- a/packages/block-editor/src/components/block-tools/block-selection-button.js
+++ b/packages/block-editor/src/components/block-tools/block-selection-button.js
@@ -164,11 +164,6 @@ function BlockSelectionButton( { clientId, rootClientId } ) {
 		const isEnter = keyCode === ENTER;
 		const isSpace = keyCode === SPACE;
 		const isShift = event.shiftKey;
-		if ( isEscape && editorMode === 'navigation' ) {
-			setNavigationMode( false );
-			event.preventDefault();
-			return;
-		}
 
 		if ( keyCode === BACKSPACE || keyCode === DELETE ) {
 			removeBlock( clientId );
@@ -188,7 +183,7 @@ function BlockSelectionButton( { clientId, rootClientId } ) {
 		const navigateUp = ( isTab && isShift ) || isUp;
 		const navigateDown = ( isTab && ! isShift ) || isDown;
 		// Move out of current nesting level (no effect if at root level).
-		const navigateOut = isLeft;
+		const navigateOut = isLeft || isEscape;
 		// Move into next nesting level (no effect if the current block has no innerBlocks).
 		const navigateIn = isRight;
 
--- a/packages/block-editor/src/components/block-tools/block-selection-button.js
+++ b/packages/block-editor/src/components/block-tools/block-selection-button.js
@@ -9,7 +9,7 @@ import clsx from 'clsx';
 import { dragHandle, trash } from '@wordpress/icons';
 import { Button, Flex, FlexItem, ToolbarButton } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
-import { useEffect, useRef } from '@wordpress/element';
+import { forwardRef, useEffect } from '@wordpress/element';
 import {
 	BACKSPACE,
 	DELETE,
@@ -48,10 +48,11 @@ import Shuffle from '../block-toolbar/shuffle';
  *
  * @param {string} props          Component props.
  * @param {string} props.clientId Client ID of block.
+ * @param {Object} ref            Reference to the component.
  *
  * @return {Component} The component to be rendered.
  */
-function BlockSelectionButton( { clientId, rootClientId } ) {
+function BlockSelectionButton( { clientId, rootClientId }, ref ) {
 	const selected = useSelect(
 		( select ) => {
 			const {
@@ -125,7 +126,6 @@ function BlockSelectionButton( { clientId, rootClientId } ) {
 		canMove,
 	} = selected;
 	const { setNavigationMode, removeBlock } = useDispatch( blockEditorStore );
-	const ref = useRef();
 
 	// Focus the breadcrumb in navigation mode.
 	useEffect( () => {
@@ -363,4 +363,4 @@ function BlockSelectionButton( { clientId, rootClientId } ) {
 	);
 }
 
-export default BlockSelectionButton;
+export default forwardRef( BlockSelectionButton );
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -156,7 +156,12 @@ export default function BlockTools( {
 			} else if ( clientIds.length === 1 ) {
 				event.preventDefault();
 				clearSelectedBlock();
-				__unstableContentRef?.current.focus();
+				// This is temporary code to see if this works with NVDA. I'm not sure the best way to find the region that contains the block editor.
+				document
+					.querySelector(
+						'[role="region"][aria-label="Editor content"]'
+					)
+					.focus();
 			}
 		} else if ( isMatch( 'core/block-editor/collapse-list-view', event ) ) {
 			// If focus is currently within a text field, such as a rich text block or other editable field,
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -172,11 +172,10 @@ export default function BlockTools( {
 
 				// The region is provivided by the editor, not the block-editor.
 				// We should send focus to the region if one is available to reuse the
-				// same interface for navigating landmarks.
-				const editorRegion = editorCanvas?.closest( '[role="region"]' );
-
-				// If no region is available, use the canvas instead.
-				const focusableWrapper = editorRegion || editorCanvas;
+				// same interface for navigating landmarks. If no region is available,
+				// use the canvas instead.
+				const focusableWrapper =
+					editorCanvas?.closest( '[role="region"]' ) ?? editorCanvas;
 
 				focusableWrapper.focus();
 			}
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -93,6 +93,8 @@ export default function BlockTools( {
 		expandBlock,
 	} = unlock( useDispatch( blockEditorStore ) );
 
+	const blockSelectionButtonRef = useRef();
+
 	function onKeyDown( event ) {
 		if ( event.defaultPrevented ) {
 			return;
@@ -153,7 +155,10 @@ export default function BlockTools( {
 				// block so that focus is directed back to the beginning of the selection.
 				// In effect, to the user this feels like deselecting the multi-selection.
 				selectBlock( clientIds[ 0 ] );
-			} else if ( clientIds.length === 1 ) {
+			} else if (
+				clientIds.length === 1 &&
+				event.target === blockSelectionButtonRef?.current
+			) {
 				event.preventDefault();
 				clearSelectedBlock();
 
@@ -204,7 +209,6 @@ export default function BlockTools( {
 			}
 		}
 	}
-
 	const blockToolbarRef = usePopoverScroll( __unstableContentRef );
 	const blockToolbarAfterRef = usePopoverScroll( __unstableContentRef );
 
@@ -235,6 +239,7 @@ export default function BlockTools( {
 
 				{ showBreadcrumb && (
 					<BlockToolbarBreadcrumb
+						ref={ blockSelectionButtonRef }
 						__unstableContentRef={ __unstableContentRef }
 						clientId={ clientId }
 					/>
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -156,12 +156,24 @@ export default function BlockTools( {
 			} else if ( clientIds.length === 1 ) {
 				event.preventDefault();
 				clearSelectedBlock();
-				// This is temporary code to see if this works with NVDA. I'm not sure the best way to find the region that contains the block editor.
-				document
-					.querySelector(
-						'[role="region"][aria-label="Editor content"]'
-					)
-					.focus();
+
+				// If we're iframed, we need to use the iframe instead of the contentRef
+				const editorCanvas =
+					__unstableContentRef.current.ownerDocument !== document
+						? document.querySelector(
+								'iframe[name="editor-canvas"]'
+						  )
+						: __unstableContentRef.current;
+
+				// The region is provivided by the editor, not the block-editor.
+				// We should send focus to the region if one is available to reuse the
+				// same interface for navigating landmarks.
+				const editorRegion = editorCanvas?.closest( '[role="region"]' );
+
+				// If no region is available, use the canvas instead.
+				const focusableWrapper = editorRegion || editorCanvas;
+
+				focusableWrapper.focus();
 			}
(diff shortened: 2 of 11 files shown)


jeryj:
@MaggieCabrera Do you have any ideas on reworking the CSS for this? The `iframe` is the right container for styling, but `iframe:has( body:focus )` styles won't apply.

blue border offset from parent and not wrapping entire canvas

This is the focus ring on the body, which will respond to `:focus` styling but isn't the right wrapper for our purpose.

alexstine:
@jeryj Why not just return false? I'm not seeing the value of Escape press going back to root level. Actually, I find this interaction created very challenging to understand.
1. When editing a block, escape goes to select mode.
2. Escape can go back to edit mode on the active block.
3. If editing a nested/inner block, escape goes to select mode.
4. Pressing escape from here goes to the root block vs. edit mode.

I'm kind of unsure why you went down this path in the first place. Deselecting a block makes total sense on escape but I think you may have slightly over complicated the select/edit mode shortcut interaction. Enter should activate, I think you would have quite a challenge finding a UI where escape could also activate.

Thanks.

alexstine:
Okay, I think I may understand a bit more. You are reverting the change.

I still don't understand why you are using escape to back all the way out though. If you press escape to switch to select mode, you could use the arrow keys to navigate. I do not think escape should focus the iFrame body ever, this would sound like focus loss to blind users using a screen reader. Not being able to see the focus ring would make the UX quite bad.
1. Edit a block.
2. Press escape to enter select mode.
3. No blocks are selected.
4. Press shift+down arrow.
5. One block is now selected.
6. Press escape to clear the selection.
7. Pressing escape again does nothing because you are already in select mode.

jeryj:
@alexstine The challenge I'm trying to solve is, where does focus go when we deselect a block? I was hoping that by using Escape to "back out" of the levels of nested blocks, it would make the path from nested block to root level clearer. And once you don't have any more blocks to select, the root level canvas was the best place I could find to clear selection without losing focus. Very open to ideas here! And thank you for reviewing this interaction!

andrewserong (commented):
Interesting exploration! I think for me, while testing it out, I'd expect the Escape key to effectively escape out of interacting within the editor canvas, rather than walk up the block hierarchy. So, when pressing Escape the second time from Edit mode to Select mode, I expected to move out of the editor canvas with no blocks selected at all.

> Deselecting a block makes total sense on escape but I think you may have slightly over complicated the select/edit mode shortcut interaction.

To me this made it feel more complicated, too. Was walking through the block parent hierarchy something that folks have been wanting to be able to do via the Escape key?

MaggieCabrera:
> @MaggieCabrera Do you have any ideas on reworking the CSS for this? The `iframe` is the right container for styling, but `iframe:has( body:focus )` styles won't apply.

> This is the focus ring on the body, which will respond to `:focus` styling but isn't the right wrapper for our purpose.

The CSS we are trying to apply here is going to be hidden by any backgrounds of elements inside, check this:

Screenshot 2024-06-03 at 12 12 54

jeryj:
@andrewserong - I'm mostly trying to see how to make escape not feel like a focus loss when you press it in select mode. This is more of a bandage so you don't go so far away from where you are if you press escape. I would love if we could press escape from select mode and:

- It deselects the block
- Leaves block focus close by in case it was an accident
- lets you know where focus currently is

I'm not sure what to do here, so I'm playing around with ideas. Any recommendations are welcome!

jeryj:
> The CSS we are trying to apply here is going to be hidden by any backgrounds of elements inside, check this:

I'm trying to apply this it to the `iframe`, not the body. I can only get the body selector to work for focus though.

jeryj:
@jasmussen @richtabor, how do you feel about this interaction? Escape while in select mode will:
- Deselect the current block
- send focus to the canvas body 
- show a blue border around the canvas

andrewserong:
This is feeling much nicer to me after the update and it going straight up to the editor canvas 👍. One question: when we back out up to the canvas, would users expect it to also switch back to edit mode, so we're kind of escaping out of the current mode of what we're doing? Essentially, if after hitting Escape the second time you then go to select a block, would you expect to be back in Edit mode again? I'm a little biased here in that I find the selection mode a bit confusing to use in general.

Keen to hear what the designers think. Thanks again for digging in here!

jasmussen:
As a quick test, this works well. Perhaps separate, but the breadcrumb for selecting the document — the only way to deselect in trunk at the moment — is setting focus on the `body`, it seems pertinent to also update that to set focus on the editor canvas instead, since that's the implication of choosing that breadcrumb. What do you think?

screenshot of the bottom left corner of the editor, showing the breadcrumb bar, with the top option emphasized

jeryj:
@andrewserong 
> Essentially, if after hitting Escape the second time you then go to select a block, would you expect to be back in Edit mode again?

I see what you mean. I tried it out, but this would be a heavier lift, as it looks like the editor gets set to navigation/select mode when you tab into it and nothing is selected. Basically, when I `setNavigationMode( false )` when clearing the selection, you press `tab` again to go to a block and it reverts back to `setNavigationMode( true )` by the editor when the first block receives focus. You can test it out by reloading a page and pressing Tab til you reach the canvas. So, I think it's best to leave this one as is.

> I'm a little biases here in that I find the selection mode a bit confusing to use in general.

Yeah, it takes a little to get the hang of, but it's especially useful to select nested blocks. It makes moving focus between blocks with a keyboard soooooo much faster. If you use it with List View open to see the structure of the blocks at the same time it will make more sense too. What made it click for me was thinking of the actions as Up/Down/In/Out instead of Up/Down/Right/Left.

@jasmussen 
> it seems pertinent to also update that to set focus on the editor canvas instead, since that's the implication of choosing that breadcrumb. What do you think?

I think that's a great idea. Let's do it on a follow-up.

jeryj:
Follow-up for template footer click fix: https://github.com/WordPress/gutenberg/pull/62290

alexstine:
@jeryj What if it followed a path like this?
1. Interact with a nested/inner block in edit mode.
2. Press escape.
3. Then we simply use the home/end keys to navigate?

Just so you can hear it, this is what it sounds like from a screen reader when we place focus on an iFrame wrapper.

> Block: Paragraph  document  editable  Testing.
> Paragraph Block. Row 1. Testing. 
> Paragraph Block. Row 1. Testing.  button
> Editor canvas  frame
> Editor canvas  document   
> You are currently in navigation mode. Navigate blocks using the Tab key and Arrow keys. Use Left and Right Arrow keys to move between nesting levels. To exit navigation mode and edit the selected block, press Enter. 
> Add title  edit  multi line  Testing   spelling error  nav item focus
> Paragraph Block. Row 1. Testing. 
> Paragraph Block. Row 1. Testing.  button

I seriously think escape should simply switch to navigation mode and then do nothing after that. Placing focus on the wrapper offers no meaningful experience for blind users. There is no other place I could think of that would make more sense.

Thanks.

jeryj:
@alexstine Yeah, that announcement is a bad experience :/ The idea is to be able to clear block selection and begin from top again. But, if we clear block selection on the canvas, then we need a place to put focus that is actually useful. 🤔

alexstine:
@jeryj What about the title field? It would be the previous focusable.

packages/block-editor/src/components/iframe/content.scss
```diff
@@ -1,5 +1,12 @@
 .block-editor-iframe__body {
 	position: relative;
+
+	// Float an outline on the top of the iframe to show focus on the canvas when pressing
+	// escape while in select mode.
+	&:focus::after {
+		@include selectedOutline();
+		position: fixed;
+	}
```
andrewserong: Is this potentially too broad? I notice that in the post editor, if I have multiple blocks selected in the canvas, then the iframe outline is also rendered, which looks unexpected:

image

andrewserong (commented):
Just gave this a re-test, and noticed a couple of subtle issues. Here's a video:

https://github.com/WordPress/gutenberg/assets/14988353/1221d0f0-d168-4a60-828c-ce82634b12e9

In the video, when selecting multiple blocks in the iframed post editor, the overall canvas outline is displayed, which looks unexpected.

Then, if I go to editor preferences and enable the custom fields, which switches off the editor iframe, pressing Escape will still clear the block selection, but there's no longer an editor iframe to be focused, so it looks like there's eventually a focus loss from pressing escape. I noticed a similar issue in https://github.com/WordPress/gutenberg/pull/62290 — I wonder if there should be a tab stop for the non-iframed post editor, too? 🤔

Please let me know if I missed something obvious in my testing, though, I'm not too familiar with the expected behaviour for the editor wrapper when it comes to focus 🙂

jeryj:
@alexstine 
> What about the title field? It would be the previous focusable.

That was my first thought too, but it would work for the post editor since the site editor doesn't have a title area. I'm still looking for a better place to put focus. Maybe even adding something or possibly a way to have a better announcement like "Editor canvas"?

@andrewserong Good catch on the multi-selection! Fixed and updated. The css selector is getting a little long, but it works :) Should work with both the iframe and div block editor now too.

jeryj:
@alexstine - I'd never done this before, but I put an aria-label on the body element and it's testing pretty well on safari with voice over. It's announcing "Block Editor, group with [number] items" which does seem useful for knowing where you are after clearing block selection. I'll let you be the judge if it's actually useful in practice though :)

alexstine:
@jeryj This does not work with NVDA on Windows with Firefox. I'm actually quite shocked this works with Voiceover. The same announcement I posted above is given. Placing focus on a wrapper and having it make sense to all users of screen readers is a non-starter. What if you simply return focus to the "Editor content" region? This might be better but I won't know until you try it.

Thanks.

jeryj:
@alexstine I've updated it to focus the editor content region. Code is sloppy. If this announcement works well for you, let me know and I'll try to get it working in a cleaner way.

andrewserong (commented):
Thanks for all the follow-ups here between this and https://github.com/WordPress/gutenberg/pull/62290. After giving this a re-test, it looks like there's now a prominent focus outline on each of the regions in the editor. Is that intentional? If so, I wonder if the scope of this PR is a little broader than the original issue 🤔. Here's a video of the regions I mean (sidebar, list view area, header, and footer of the interface):

https://github.com/WordPress/gutenberg/assets/14988353/8268715e-5e50-45c9-9720-efa6de127993

I likely won't get a chance to review this or the other PR again until later on next week. Happy to give it another spin then if it's still open 🙂

packages/block-editor/src/components/block-tools/index.js
```diff
@@ -152,6 +153,15 @@ export default function BlockTools( {
...
 				selectBlock( clientIds[ 0 ] );
+			} else if ( clientIds.length === 1 ) {
+				event.preventDefault();
+				clearSelectedBlock();
+				// This is temporary code to see if this works with NVDA. I'm not sure the best way to find the region that contains the block editor.
+				document
+					.querySelector(
+						'[role="region"][aria-label="Editor content"]'
+					)
+					.focus();
```
ellatrix: Hah, this does what I described here no? https://github.com/WordPress/gutenberg/pull/62290#discussion_r1630701078
Doesn't it fix that issue too?

ellatrix: Maybe instead of focussing right here and having to query the DOM, you could listen for block selection clearing (an effect checking for no selection) in writing flow?

jeryj: Ooooh - that could work well! That would fix any instance where block selection is cleared. It's still an issue of _where do we put focus_? Maybe we can add an extra element with an aria label and `tabindex=-1` that wraps the block list?

jeryj:
@andrewserong 

> After giving this a re-test, it looks like there's now a prominent focus outline on each of the regions in the editor. Is that intentional?

That's so odd. They all have tabindex=-1 so I thought they would only receive the focus if we sent it there with `element.focus()`. Either way, I'd like to decide _where_ we should be placing focus. Thanks again for all the reviews!

jeryj:
I've went with focusing the region, as that is already a place where focus could land when navigating the editor. I believe @ajlende and I have fixed the edge cases and have this working reliably. 🤞🏻 We'd like to do a follow-up to make it easier to send focus to a region with `useNavigateRegions` by passing a string to find the region and focus it, or at least centralize the logic used to send focus to the best place for the editor when this unselect or footer template click happens.

packages/components/src/higher-order/navigate-regions/style.scss
```diff
@@ -1,6 +1,20 @@
...
+		position: absolute;
+		top: 0;
+		left: 0;
+		right: 0;
+		bottom: 0;
+		content: "";
+		pointer-events: none;
+		outline: 4px solid $components-color-accent;
+		outline-offset: -4px;
+		z-index: z-index(".is-focusing-regions {region} :focus::after");
```
ajlende: I wonder if this should be extracted into a mixin or a `%` selector to be reused so we don't accidentally update one without the other.

packages/block-editor/src/components/block-tools/index.js
```diff
@@ -152,6 +155,30 @@ export default function BlockTools( {
...
+						  )
+						: __unstableContentRef.current;
+
+				// The region is provivided by the editor, not the block-editor.
+				// We should send focus to the region if one is available to reuse the
+				// same interface for navigating landmarks.
+				const editorRegion = editorCanvas?.closest( '[role="region"]' );
+
+				// If no region is available, use the canvas instead.
+				const focusableWrapper = editorRegion || editorCanvas;
```
ajlende:
```suggestion
				// The region is provivided by the editor, not the block-editor.
				// We should send focus to the region if one is available to reuse the
				// same interface for navigating landmarks. If no region is available,
				// use the canvas instead.
				const focusableWrapper =
					editorCanvas?.closest( '[role="region"]' ) ?? editorCanvas;
```

packages/block-editor/src/components/block-tools/block-selection-button.js
```diff
@@ -9,7 +9,7 @@ import clsx from 'clsx';
 import { dragHandle, trash } from '@wordpress/icons';
 import { Button, Flex, FlexItem, ToolbarButton } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
-import { useEffect, useRef } from '@wordpress/element';
+import { forwardRef, useEffect } from '@wordpress/element';
```
jeryj, line 12: ~~We shouldn't need to forward a ref here anymore. I think some unnecessary things made their way in during the rebase. I'll get this fixed.~~

I was wrong on this. It is necessary unless we want to check for the selection button by name, as we need to know if the target is our current navigation selection button. There may be another way to check this without passing the ref though.

jasmussen:
Thanks for working on this. The fact that this allows me to deselect blocks using a keyboard shortcut, vs. only being able to do so by clicking the top breadcrumb in the footer, is helpful.

Not a strong opinion, or blocker, happy to defer to others; it's slighty jarring that the focus style is different from any other focus style. I realize there's nuance here since it's also a region, but when I see one focus style for blocks in the canvas, and another one for the whole canvas itself, it's not clear why, and in a quick testing, the same focus style as applied to everything else seems to work well enough:

focus

jeryj:
> Not a strong opinion, or blocker, happy to defer to others; it's slighty jarring that the focus style is different from any other focus style.

@richtabor and I were discussing making the region focus styles twice as thick as the block selection outlines so they could be thicker for easier visibility on the large area, but not as thick as they are now. I'll follow-up on that after this PR is merged.

jeryj:
@jasmussen - Here's an update to the focus styles for after this PR is merged: https://github.com/WordPress/gutenberg/pull/62881

ajlende (approved):

jasmussen:
Thank you for the work here!

Issue #55023, opened 2023-10-03
CustomSelectControl: refactor with ariakit


brookewp:
## Context

Continuing with the exploration started by @ciampo in #41466, this issue will track the migration of `CustomSelectControl` to Ariakit.

Legacy: https://github.com/WordPress/gutenberg/pull/55234
New version: https://github.com/WordPress/gutenberg/pull/55790

##  Initial steps

- [x] Analyze the legacy component, learn how it works, take note of its features
- [x] Study Ariakit's Select component
- [x] Look into ARIA roles 
- [x] Consider using WordPressComponentsProps https://github.com/WordPress/gutenberg/pull/56998
- [x] Create a new, temporary component which should implement:
  - [x] `children` (and a CustomSelectControlV2.Item component) to allow its consumers to specify select options
  - [x] basic functionality (`value` / `onChange` / `defaultValue` / `label` / `size` props)
  - [x] styles that match the legacy component
  - [x] Storybook docs and examples (maybe one with post authors)
- [x] Refine README, types stories, etc #57902

##  V2 Legacy Wrapper

Completed in #57902

- [x] Analyze the "API gap" between the legacy and the new component 
- [x] Come up with a strategy on how to bridge the gap
- [x] Add support for the "monolithic" usage of the component with the options prop
- [x] Bolster the legacy CustomSelectControl unit tests https://github.com/WordPress/gutenberg/pull/56575
- [x] Use the previously bolstered unit tests to certify that the new component behaves like the legacy one
- [x] Align unit tests #62706
- [x] Fix option extra attributes types and propagation to `onChange` callback and DOM elements #62255
- [x] Investigate potential issues when using v2 legacy adapter in controlled mode, as the component doesn't seem to update correctly when a new `value` is passed? (can be reproduced by following the testing instructions in #62255 re. `FontAppearanceControl` in global styles) — fix in https://github.com/WordPress/gutenberg/pull/62733
- [x] Double check initial item selection, including the fact that `onChange` doesn't fire on first render on v1 while it does on legacy v2 (WIP PR #62308, [comment](https://github.com/WordPress/gutenberg/pull/62706#discussion_r1647734766)) — fix in https://github.com/WordPress/gutenberg/pull/62733
- [x] Double check any styling issues as flagged in [this comment](https://github.com/WordPress/gutenberg/pull/61272#issuecomment-2136367426) and following conversation  (plus check WIP in https://github.com/WordPress/gutenberg/pull/62424)
[...]

PR #61272, opened 2024-05-01
Components: Replace/old custom select control with v2 legacy adapter

fullofcaffeine:
## What?

- Replace the exported implementation for the classic `CustomSelectControl` with the legacy adapter for `CustomSelectControlV2`;
- Fix any API errors that might prevent consumers from working;
- Fix any styling issues - it should look as close to the classic one as possible, unless we decide to change the consumer UX as part of the changes here.

## Why?

The new `CustomSelectControlV2` has been recently introduced in the library and provides new accessibility features. While we don't migrate to V2, the legacy adapter provides an API adapter that should make the migration easier and more gradual. This PR is an attempt to do that, and also fix any bugs that we might find along the way.

## How?

By replacing the old 'CustomSelectControl' export with the new implementation (the legacy adapter for V2). Also, testing consumers and seeing if the new implementation works well, and fixing bugs along the way (might split into other PRs later).

## Testing Instructions

- Unit tests should pass;
- See https://github.com/WordPress/gutenberg/pull/61272#issuecomment-2118608577.

PR #62198, opened 2024-05-31
CustomSelectControlV2: Handle long strings in selected value

mirka:
Part of #55023

## What?

Correctly truncate and style long strings in the currently selected value in CustomSelectControlV2.

## Why?

Long strings were not styled correctly.

### Before

Long strings look wrong

## Testing Instructions

Compare the "With Long Labels" stories between CustomSelectControl (v1) and CustomSelectControlV2 Legacy. All `size` variants should look correct.

> [!NOTE]  
> You may notice that the `LABEL` is shifted down 1px in the v1. I looked into it and v1 is actually wrong in this case, so let's ignore. (FWIW, it will look correct in wp-admin due to a style in the `forms.css` stylesheet.)

## Screenshots or screencast 

|| v1 | v2 Default & Legacy |
|--------|--------|-------|
|Long labels|long v1|long v2 legacy|
|With hints|hint v1|hint v2 legacy|


packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -74,7 +74,7 @@ const CustomSelectButton = ( {
 			// move selection rather than open the popover
 			showOnKeyDown={ false }
 		>
-			<div>{ computedRenderSelectedValue( currentValue ) }</div>
```
mirka, line 77: `div` is unnecessary.

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -82,12 +82,12 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 		);
 
 		return (
-			<>
+			<Styled.SelectedExperimentalHintWrapper>
```
mirka, line 93: Added a wrapper div to apply text truncation styles as in v1.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -55,40 +66,41 @@ export const Select = styled( Ariakit.Select, {
...
 			},
 		};
 
 		return sizes[ size ] || sizes.default;
 	};
 
 	return css`
-		display: flex;
-		align-items: center;
-		justify-content: space-between;
```
mirka, line 80: No reason for this to be a flexbox.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -55,40 +66,41 @@ export const Select = styled( Ariakit.Select, {
...
+		display: block;
 		background-color: ${ COLORS.theme.background };
 		border: none;
+		color: ${ COLORS.theme.foreground };
 		cursor: pointer;
+		font-family: inherit;
 		font-size: ${ CONFIG.fontSize };
+		text-align: left;
```
mirka, line 95: Text align was missing.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -55,40 +66,41 @@ export const Select = styled( Ariakit.Select, {
...
 	return css`
-		display: flex;
-		align-items: center;
-		justify-content: space-between;
+		display: block;
 		background-color: ${ COLORS.theme.background };
 		border: none;
+		color: ${ COLORS.theme.foreground };
 		cursor: pointer;
+		font-family: inherit;
```
mirka, line 93: Correct `color` and `font-family` was missing.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -55,40 +66,41 @@ export const Select = styled( Ariakit.Select, {
...
 		font-size: ${ CONFIG.fontSize };
+		text-align: left;
 		width: 100%;
 
 		&[data-focus-visible] {
 			outline: none; // handled by InputBase component
 		}
 
 		${ getSize() }
+		${ ! hasCustomRenderProp && truncateStyles }
```
mirka, line 103: We shouldn't add truncate styles when the consumer passes a custom render function.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -55,40 +66,41 @@ export const Select = styled( Ariakit.Select, {
 		const sizes = {
 			compact: {
 				[ heightProperty ]: 32,
-				paddingInlineStart: space( 2 ),
-				paddingInlineEnd: space( 1 ),
+				paddingInlineStart: 8,
```
tyxla, line 69: Is there a good reason to stop using the grid space util and hardcode the padding widths like that?

packages/components/src/custom-select-control-v2/styles.ts
tyxla: Nice catch 👍 I wonder how this didn't come up earlier.

tyxla (approved):
Working well in my testing and the code looks good 👍 

Thanks for cleaning up the styles while passing through!

packages/components/src/custom-select-control-v2/styles.ts
fullofcaffeine: Curious about that, too.

fullofcaffeine (approved):
LGTM! Thanks for working on this.

I started fixing other styling issues here: https://github.com/WordPress/gutenberg/pull/62424 (it's been split from [#](https://github.com/WordPress/gutenberg/pull/61272)), so I'll rebase from this, review the current issues, and continue working there.

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Knowing past Marco, there's a chance flexbox was used to automatically trim extra white space (although it's also very possible that this is a remnant of a previous style and it's totally fine to remove it)

ciampo: Probably totally fine to remove it :D

packages/components/src/custom-select-control-v2/styles.ts
mirka: No big reason other than that I thought it was simpler to read in context, especially when I want to add the `chevronIconSize` like this. (`space()` returns a `calc()`).

PR #59616, opened 2024-03-05
Add Box Shadow support for featured image

colinduwe:
## What?
This PR adds box shadow support to the Post Featured Image Block so it has the same controls as the core Image block. 

## Why?
This PR resolves https://github.com/WordPress/gutenberg/issues/59067 and brings more parity between the Image and Post Featured Image blocks.

## How?
This PR adds block support for box shadows. The shadow style needs to be applied to the image tag inside the figure tag like the border styles. The selectors have been specified so that global styles from theme.json and elsewhere are also applied to the image rather than the figure. 

In order for this to work, the shadow.php needed to be updated to support __experimentalSkipSerialization. Without it, the box shadow style is printed to the figure.  

NOTE: This PR would potentially affect any themes that were providing a shadow style definition to the core/post-featured-image. Currently, that style is applied to the figure. With this PR it is applied to the image. This allows the box shadow to work in conjunction with a border radius.

## Testing Instructions
1. Open a post or page
2. Add a featured image
3. Add a Featured Image Block
4. Add a box shadow style. It should be visible in the editor.
5. Adjust any of the other style controls. They should all work in harmony.
6. Test the result on the front end.

### Testing Instructions for Keyboard
1. Open a post or page
2. Add a featured image
3. Add a Featured Image Block
4. Navigate to the sidebar with the keyboard
5. Navigate to the Border and & Shadow section
6. Open the kabob menu to enable the shadow with the keyboard
7. Add a box shadow style. It should be visible in the editor.
8. Adjust any of the other style controls. They should all work in harmony.
9. Test the result on the front end.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/2152870/f833d4ff-56d3-4bbe-a3de-22aeb2bae4cc
(Note that I've added a shadow to my theme.json to demonstrate that users can override global styles)

PR #62207, opened 2024-06-02
Post Featured Image: Shadows not applied to block instances

t-hamano:
Follow up #59616

## What?

This PR fixes an issue where the shadow is not applied to the Post Featured block under the following conditions:

- Apply shadow to block instance
- On the front end

## Why?

In the current implementation, CSS is generated via the style engine in the render callback function.

However, the value returned from the style engine is as follows and does not contain `css` key:

```
array(1) {
  ["declarations"]=>
  array(1) {
    ["box-shadow"]=>
    string(65) "6px 6px 0px -3px rgba(255, 255, 255, 1), 6px 6px rgba(0, 0, 0, 1)"
  }
}
```

## How?

Shadow is just a CSS value, and there is no class like `has-shadow`. In this case, I don't think there is any need to use the style engine.

## Testing Instructions

- Insert a Post Featured Image block and apply a shadow.
- The shadow will be applied on the front end.
- Bonus: In the global style, apply a shadow to the Post Featured Image block. That shadow should be overwritten by the block instance shadow.

### DIFF
--- a/packages/block-library/src/post-featured-image/index.php
+++ b/packages/block-library/src/post-featured-image/index.php
@@ -50,12 +50,9 @@ function render_block_core_post_featured_image( $attributes, $content, $block ) 
 	if ( ! empty( $attributes['scale'] ) ) {
 		$extra_styles .= "object-fit:{$attributes['scale']};";
 	}
-	if ( ! empty( $attributes['style']['shadow'] ) ) {
-		$shadow_styles = wp_style_engine_get_styles( array( 'shadow' => $attributes['style']['shadow'] ) );
 
-		if ( ! empty( $shadow_styles['css'] ) ) {
-			$extra_styles .= $shadow_styles['css'];
-		}
+	if ( ! empty( $attributes['style']['shadow'] ) ) {
+		$extra_styles .= "box-shadow:{$attributes['style']['shadow']};";
 	}
 
 	if ( ! empty( $extra_styles ) ) {


andrewserong (commented):
Thanks for looking into this — unfortunately I'm not able to reproduce the bug on `trunk`. When setting an individual featured image block to use a shadow (either one of the default shadows, or a custom shadow defined in the site editor's global styles), I can see the shadow output in the site frontend, and the `css` key is returned by the style engine.

CSS key:

image

Site frontend

| Bundled shadow | Custom shadow in global styles |
| --- | --- |
| image | image |

I'm testing with GB trunk in a test site running WP 6.5.3, and with TT4 theme. Do you have any additional testing instructions that might help us debug this?

Rather than merge this PR, I'd be keen for us to figure out in which conditions the CSS key isn't being returned by this style engine call, in case this fix masks another issue, and since I haven't been able to reproduce the problem so far.

CC: @ramonjd for visibility since this is style engine related 😅

ramonjd:
Thanks for the ping, and for the PR @t-hamano 

I can replicate with a hardcoded value, e.g., 

```php
		$shadow_styles = wp_style_engine_get_styles( array( 'shadow' => '6px 6px 0px -3px rgba(255, 255, 255, 1), 6px 6px rgba(0, 0, 0, 1)' ) );
```

> Rather than merge this PR, I'd be keen for us to figure out in which conditions the CSS key isn't being returned by this style engine call, in case this fix masks another issue, and since I haven't been able to reproduce the problem so far.

Good call.

It seems to be rather an issue with `safecss_filter_attr()` 

It doesn't pass the following test:

```php
			/*
			 * Allow CSS functions like var(), calc(), etc. by removing them from the test string.
			 * Nested functions and parentheses are also removed, so long as the parentheses are balanced.
			 */
			$css_test_string = preg_replace(
				'/\b(?:var|calc|min|max|minmax|clamp|repeat)(\((?:[^()]|(?1))*\))/',
				'',
				$css_test_string
			);

			/*
			 * Disallow CSS containing \ ( & } = or comments, except for within url(), var(), calc(), etc.
			 * which were removed from the test string above.
			 */
			$allow_css = ! preg_match( '%[\\\(&=}]|/\*%', $css_test_string );
```

See: https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/kses.php#L2606

From looking, it's because WordPress doesn't allow `rgba()`

I think WordPress should allow `rgba()` values:

- it [already does for gradients](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/kses.php#L2585).
- rgba has been widely supported for many years https://caniuse.com/?search=rgba

So I reckon, rather than hardcode a fix only for one block, let's fix it every where by adding a hook on `safecss_filter_attr_allow_css` and creating a Core patch.

ramonjd:
Ah, I think it's because it's matching `\(` in this regex check:


```php
$allow_css = ! preg_match( '%[\\\(&=}]|/\*%', $css_test_string );
```

andrewserong:
I haven't tested it, but would it work to add `rgb` and `rgba` to the list of functions a bit higher up in `'/\b(?:var|calc|min|max|minmax|clamp|repeat)(\((?:[^()]|(?1))*\))/'`?

ramonjd:
So something like this might work:

<details>

<summary>Just allowing RGB(A)</summary>

```php
if ( ! function_exists( 'gutenberg_safecss_filter_attr_allow_rgb' ) ) {
	/**
	 * Filters the check for unsafe CSS in `safecss_filter_attr` to allow `rgba()`.
	 *
	 * @param bool   $allow_css       Whether the CSS in the test string is considered safe.
	 * @param string $css_test_string The CSS string to test.
	 */
	function gutenberg_safecss_filter_attr_allow_rgb( $allow_css, $css_test_string ) {
		/*
		* Allow CSS functions like var(), calc(), etc. by removing them from the test string.
		* Nested functions and parentheses are also removed, so long as the parentheses are balanced.
		*/
		$css_test_string = preg_replace(
			// SINGLE CHANGE HERE: adding `|rgb[a]?`
			'/\b(?:var|calc|min|max|minmax|clamp|repeat|rgb[a]?)(\((?:[^()]|(?1))*\))/',
			'',
			$css_test_string
		);
		/*
		 * Disallow CSS containing \ ( & } = or comments, except for within url(), var(), calc(), etc.
		 * which were removed from the test string above.
		 */
		return ! preg_match( '%[\\\(&=}]|/\*%', $css_test_string );
	}
}

add_filter( 'safecss_filter_attr_allow_css', 'gutenberg_safecss_filter_attr_allow_rgb', 10, 2 );
```

</details>

It will allow `rgb[a]` values everywhere, e.g., `color: rgba(0, 0, 0, 0.1)` so it's pretty greedy.

Therefore we should check whether we need something stricter for `rgba` values, e.g., that they are indeed numeric.


<details>

<summary>With extra checks for RGB(A)</summary>

```php
if ( ! function_exists( 'gutenberg_safecss_filter_attr_allow_rgb' ) ) {
	/**
	 * Filters the check for unsafe CSS in `safecss_filter_attr` to allow `rgba()`.
	 *
	 * @param bool   $allow_css       Whether the CSS in the test string is considered safe.
	 * @param string $css_test_string The CSS string to test.
	 */
	function gutenberg_safecss_filter_attr_allow_rgb( $allow_css, $css_test_string ) {
		$parts     = explode( ':', $css_test_string, 2 );
		$css_value = trim( $parts[1] );
		$has_rgb   = str_contains( $css_value, 'rgba(' ) || str_contains( $css_value, 'rgb(' );
		
		// Check that rgb(a) values are digits.
		if ( $has_rgb ) {
			$regex = '/rgba?\(\s*\d{1,3},\s*\d{1,3},\s*\d{1,3}(,\s*\d(\.\d+)?)?\s*\)/';
			if ( preg_match_all( $regex, $css_value ) ) {
				$css_test_string = str_replace( $css_value, '', $css_test_string );
			}
		}

		/*
		 * The code below is taken from Core: wp-includes/kses.php.
		 * It does not need backporting.
		 */

		/*
		 * Allow CSS functions like var(), calc(), etc. by removing them from the test string.
		 * Nested functions and parentheses are also removed, so long as the parentheses are balanced.
		 */
		$css_test_string = preg_replace(
			'/\b(?:var|calc|min|max|minmax|clamp|repeat)(\((?:[^()]|(?1))*\))/',
			'',
			$css_test_string
		);

		/*
		 * Disallow CSS containing \ ( & } = or comments, except for within url(), var(), calc(), etc.
		 * which were removed from the test string above.
		 */
		$allow_css = ! preg_match( '%[\\\(&=}]|/\*%', $css_test_string );
	}
}

add_filter( 'safecss_filter_attr_allow_css', 'gutenberg_safecss_filter_attr_allow_rgb', 10, 2 );


```

</details>

ramonjd:
> I haven't tested it, but would it work to add rgb and rgba to the list of functions a bit higher up

Yeah, I was just [testing that](https://github.com/WordPress/gutenberg/pull/62207#issuecomment-2144268066)

So `'/\b(?:var|calc|min|max|minmax|clamp|repeat|rgb[a]?)(\((?:[^()]|(?1))*\))/',` works fine for both `rgb` and `rgba`.

It's probably good enough.

It'll also open up `rgb` values for all properties. I think this is fine.

But if we want to add an extra layer of security, then the rgb value can also be checked, e.g., 

```php
		// Check that rgb(a) values are digits.
		if ( $has_rgb ) {
			$regex = '/rgba?\(\s*\d{1,3},\s*\d{1,3},\s*\d{1,3}(,\s*\d(\.\d+)?)?\s*\)/';
			if ( preg_match_all( $regex, $css_value ) ) {
				$css_test_string = str_replace( $css_value, '', $css_test_string );
			}
		}

```

t-hamano:
Thanks for the detailed advice! I should have looked deeper 😅

I looked into why this issue was difficult to reproduce and found that normally the shadow preset is applied with CSS variables just like other presets:

```
<!-- wp:post-featured-image {"style":{"shadow":"var:preset|shadow|outlined"}} /-->
```

However, after updating the shadow preset or applying the shadow several times, I found that somehow it changed to a hardcoded value with `rgba` values ​​like the following (maybe I should investigate this issue more deeply):

```
<!-- wp:post-featured-image {"style":{"shadow":"6px 6px 0px rgba(0, 0, 0, 0.2)"}} /-->
```

I agree with you that a more fundamental approach is needed as you suggested.

While looking through Trac, I found a changeset that allows rgba for background colors: https://core.trac.wordpress.org/ticket/48376

In light of this approach, I am considering submitting a core patch limited to certain CSS properties. For example, the following code:

<details><summary>Patch Example</summary>

```diff
diff --git a/wp-includes/kses.php b/wp-includes/kses.php
index 8ab9afd5..99b049c6 100644
--- a/wp-includes/kses.php
+++ b/wp-includes/kses.php
@@ -2513,6 +2513,14 @@ function safecss_filter_attr( $css, $deprecated = '' ) {
                'background-image',
        );
 
+       /*
+        * CSS attributes that accept rgba data types.
+        *
+        */
+       $css_rgba_data_types = array(
+               'box-shadow'
+       );
+
        if ( empty( $allowed_attr ) ) {
                return $css;
        }
@@ -2528,6 +2536,7 @@ function safecss_filter_attr( $css, $deprecated = '' ) {
                $found           = false;
                $url_attr        = false;
                $gradient_attr   = false;
+               $rgba_attr       = false;
                $is_custom_var   = false;
 
                if ( ! str_contains( $css_item, ':' ) ) {
@@ -2546,6 +2555,7 @@ function safecss_filter_attr( $css, $deprecated = '' ) {
                                $found         = true;
                                $url_attr      = in_array( $css_selector, $css_url_data_types, true );
                                $gradient_attr = in_array( $css_selector, $css_gradient_data_types, true );
+                               $rgba_attr     = in_array( $css_selector, $css_rgba_data_types, true );
                        }
 
                        if ( $is_custom_var ) {
@@ -2588,6 +2598,14 @@ function safecss_filter_attr( $css, $deprecated = '' ) {
                        }
                }
 
+               if ( $found && $rgba_attr ) {
+                       $css_value = trim( $parts[1] );
+                       if ( preg_match( '/rgb(a)?\([^)]+\)/', $css_value ) ) {
+                               // Remove the whole `rgba` bit that was matched above from the CSS.
+                               $css_test_string = str_replace( $css_value, '', $css_test_string );
+                       }
+               }
+
                if ( $found ) {
```

</details> 

Maybe it would be better to add an equivalent filter in Gutenberg _after_ the core patch is comitted.

What do you think?

ramonjd:
Thanks @t-hamano !!

In my opinion, a Core patch would be the way to go then we can decide if a filter in Gutenberg is necessary.

I'm happy to help out, or test when it's ready. 

> In light of this approach, I am considering submitting a core patch limited to certain CSS properties. 

What do you think about allowing `rgb` values for all CSS properties, not just `box-shadow`? They can be used anywhere a color can be used.

andrewserong:
> What do you think about allowing rgb values for all CSS properties, not just box-shadow? They can be used anywhere a color can be used.

I think allowing everywhere would be the way to go. My understanding is that the safe CSS filters are about avoiding dangerous CSS values, not about validating CSS, so I think allowing `rgb()` or `rgba()` anywhere could be a simpler and pragmatic way to go.

t-hamano:
I discovered there was a ticket (opened 11 years ago!) proposing to allow `rgb(a)`: https://core.trac.wordpress.org/ticket/24157

I proposed to move the ticket forward, so I'd like to close this PR for now. If `rgb(a)` values ​​are allowed in the  core, I'll submit a PR to backport it to Gutenberg.

Thanks everyone for the advice!

ramonjd:
> I discovered there was a ticket (opened 11 years ago!) proposing to allow rgb(a): [core.trac.wordpress.org/ticket/24157](https://core.trac.wordpress.org/ticket/24157)

Now that folks have had 11 years to think about it, I think the time is right! :trollface: 

Thanks for pursuing this @t-hamano

PR #62214, opened 2024-06-03
Editor: Avoid remounts of `DocumentBar`

### LINKED ISSUE #61875: Document Bar missing the initial animation changing from page/template

stokesman:
## What?
Removes the `PostTypeSupportCheck` that is around `DocumentBar`.

## Why?
Fixes #61875

## More
`PostTypeSupportCheck` causes its children to remount as it relies on `getPostType` which is asynchronous. The change in this PR avoids the remount that interferes with animation.

## Testing Instructions
1. Open a page in the Site or Post editor
2. Select “Edit template” in the sidebar
3. Note the animation in the Document Bar

Same as above, but I can add more detail if needed.

### Currently

https://github.com/WordPress/gutenberg/assets/9000376/1717312c-4c59-4f66-997a-d69c0164e4ff

### On this branch

https://github.com/WordPress/gutenberg/assets/9000376/cc2561b6-769e-48e4-a1bf-9d623ccb63ed

### DIFF
--- a/packages/editor/src/components/document-bar/index.js
+++ b/packages/editor/src/components/document-bar/index.js
@@ -29,7 +29,6 @@ import { decodeEntities } from '@wordpress/html-entities';
 import { TEMPLATE_POST_TYPES, GLOBAL_POST_TYPES } from '../../store/constants';
 import { store as editorStore } from '../../store';
 import { unlock } from '../../lock-unlock';
-import PostTypeSupportCheck from '../post-type-support-check';
 
 const TYPE_LABELS = {
 	// translators: 1: Pattern title.
@@ -182,25 +181,20 @@ export default function DocumentBar() {
 						}
 					>
 						<BlockIcon icon={ templateIcon } />
-						<PostTypeSupportCheck supportKeys="title">
-							<Text
-								size="body"
-								as="h1"
-								aria-label={
-									TYPE_LABELS[ postType ]
-										? // eslint-disable-next-line @wordpress/valid-sprintf
-										  sprintf(
-												TYPE_LABELS[ postType ],
-												title
-										  )
-										: undefined
-								}
-							>
-								{ title
-									? decodeEntities( title )
-									: __( 'No Title' ) }
-							</Text>
-						</PostTypeSupportCheck>
+						<Text
+							size="body"
+							as="h1"
+							aria-label={
+								TYPE_LABELS[ postType ]
+									? // eslint-disable-next-line @wordpress/valid-sprintf
+									  sprintf( TYPE_LABELS[ postType ], title )
+									: undefined
+							}
+						>
+							{ title
+								? decodeEntities( title )
+								: __( 'No Title' ) }
+						</Text>
 					</motion.div>
 					<span className="editor-document-bar__shortcut">
 						{ displayShortcut.primary( 'k' ) }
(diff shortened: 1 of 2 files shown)


youknowriad:
This probably changes the header for posts without "title" support, I wonder how it looks now

youknowriad:
It looks like that. We should probably hide the icon as well no?

Screenshot 2024-06-03 at 12 41 16 PM

stokesman:
> We should probably hide the icon as well no?

Maybe. Yet perhaps a question of greater importance is whether omitting the `h1` is okay in the first place.

I noticed the document sidebar in posts without "title" support looks like this:
image

Simply removing the `PostTypeSupportCheck` in the document bar creates the same result there too and seems the better thing to do for now. So I pushed 08f4388.

Not that I think it’s ideal since there’s no distinguishing a post that can’t have title from one that lacks it. I’d like to try improving it but I think we’d need to modify `PostTypeSupportCheck` or make a hook like it to offer more nuanced usage.

youknowriad:
I don't mind that and you make a good point about the lack of H1 but would love a design check at least. @jasmussen @jameskoster

jasmussen:
> This probably changes the header for posts without "title" support, I wonder how it looks now

I guess this is the main question, correct me if I'm wrong.

Pages need an H1 so good find, important to fix. Jay is out on vacation for a few weeks, so I don't have the full context he might have, but it seems like an unintentional omission, and including the title, especially because it's in the inspector, seems fine.

Though we need to decide what it says. Currently when I add a new post, with title support but no title, it says No Title:

Screenshot 2024-06-04 at 09 05 09

But in these contexts, it says (no title):
Screenshot 2024-06-04 at 09 08 52

Screenshot 2024-06-04 at 09 09 00

Screenshot 2024-06-04 at 09 09 39

I've no strong preference here other than unifying this, which can happen separately from this, but is worth doing.

Easiest is to just change it to "(no title)" everywhere. The parenthesis subtly emphasizes that it _isn't_ the title, that it's just placeholder text, but maybe the lack of sentence case is a little weird, perhaps "(No title)" is better? In any case, "No Title" is wrong, we don't use titlecase in these contexts.

youknowriad:
@jasmussen The main question for me is whether showing "(no title)" as a title for post types that don't have titles is the way to go. It's a bit different than posts with an empty title.

jasmussen:
> The main question for me is whether showing "(no title)" as a title for post types that don't have titles is the way to go. It's a bit different than posts with an empty title.

Yes, understood. I'm hearing that "(no title)" may still suggest it _can_ hold a title? We could look at other bits of text, though I'd suggest they follow the same format, such "(title disabled)"?

I don't love that, to be clear, and I think we should, across all of WordPress, change what these say so it's consistent. But we can flex a bit to get this PR shipping: it's easy, and seems safe, to change this verbiage again on the block editor side of things, so it seems like we should pick something and go with it.

youknowriad (approved):
Ok, than for me this is good to go. We can adapt the wording later if needed.

stokesman:
Thanks Joen and Riad for weighing in and reviewing ❤️

ellatrix:

PR #61236, opened 2024-04-30
Block Bindings: Simplify the HTML replacement logic until the HTML API is ready

SantosGuillamot:
## What?
Change the block bindings HTML replacement logic to use a hardcoded list of selectors based on the block and a regex to substitute the inner content.

## Why?
The existing logic is too complex because the HTML API doesn't support CSS selectors yet and it doesn't have an official way to modify the inner HTML. This makes it difficult to understand, iterate on it and add new functionalities.

As it will be solved in the future once the HTML API is ready, I believe it is better to keep a simple hardcoded version that is easier to understand and iterate on.

## How?
* Change the selector based on the supported blocks.
* Using a regex, replace the inner content.

## Testing Instructions

0. Register a custom field in your site to test it. You can use a code snippet like this:
```
register_meta(
	'post',
	'text_custom_field',
	array(
		'show_in_rest' => true,
		'single'       => true,
		'type'         => 'string',
		'default'	   => 'https://wpmovies.dev/wp-content/uploads/2023/04/goncharov-poster-original-1-682x1024.jpeg',
	)
);
```

1. Go to a page and insert a paragraph block.
2. Go to the Code editor and connect the paragraph to the custom field by adding the metadata bindings property. It should look something like this:
```
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_custom_field"}}}}} -->
<p>Original content</p>
<!-- /wp:paragraph -->
```
3. Save the page and check that in the frontend, the text defined in the custom field is used.
4. Repeat the process for heading, and button blocks.

Issue #17221, opened 2019-08-27
Button Block: Only display buttons if the title attribute exists 


richtabor:
**Describe the bug**
Currently, a Button block without a `title` attribute still renders an empty button via the save function, leaving an empty `<a>` on the page. 

**To reproduce**
1. Add an empty Button block
2. Publish the page
3. View the published page

**Expected behavior**
A button is not rendered unless the `title` attribute is entered, preventing empty buttons from displaying if it's not intentionally configured within the editor.

**Screenshots**
![ScreenFlow](https://user-images.githubusercontent.com/1813435/63803293-721d1100-c8c9-11e9-9d19-4f223c6b0d16.gif)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Safari
 - Version: 12.1.1

**Additional context**
Using Gutenberg 6.3, although it happens regardless of whether or not the plugin is active.

bobbingwide:
Problem still occurs when you add a new button to the buttons block with Gutenberg 7.2.0

dgaik-gg:
What for example with arrow buttons where the arrow is added using a class and text is not necessary? Currently adding a space allows this button to be rendered but this is not a good way.

getsource:
Hi @dgaik-gg!

Thanks much for the note about this!
Would you mind opening [a new issue](https://github.com/WordPress/gutenberg/issues/new/choose), to open discussion about the best way to address this?

PR #62220, opened 2024-06-03
Fix applying bindings or pattern overrides to button blocks with empty text

### LINKED ISSUE #61303: Pattern Overrides: Button block left empty in original pattern and used as an override throws an error in pattern instances
### Description

If a user creates a pattern with an empty button block that has 'Allow Overrides' enabled, an error is thrown when that pattern is used and text is added to the button.

I believe this happens because the button block outputs nothing when the button block is empty:
https://github.com/WordPress/gutenberg/blob/edb7e4ea109d0155de9d6dce327b2ba72dead56b/packages/block-library/src/button/save.js#L34-L36

But block bindings still tries to update the HTML for the pattern overrides binding without checking that a button is present.

A basic fix might be to check for the presence of a button element, and do nothing if there isn't one, but this leads to a problem that no button block is output when the user specifies text in a pattern instance for an empty button. I'm not sure if we could move this check to a server side render callback to fix the problem?

The reverse is also a problem. If a button is specified in the original pattern with text, but the override removes text, an empty button is output, whereas the button block would usually output nothing.

I think the same could also happen for any other type of block binding, so this is probably not a pattern overrides specific issue.

1. Create a synced pattern with an empty button block (no text defined)
2. Allow overrides for the button block
3. Save the pattern
4. Create a new post and insert the pattern you just created
5. Enter some text into the button
6. Preview the post

Expected: No error occurs, the button is shown
Actual: An error is thrown

```
Warning: foreach() argument must be of type array|object, null given in /var/www/html/wp-content/plugins/gutenberg/lib/compat/wordpress-6.5/blocks.php on line 84 Deprecated: strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in /var/www/html/wp-content/plugins/gutenberg/lib/compat/wordpress-6.5/blocks.php on line 91 Notice: Function WP_HTML_Tag_Processor::seek was called incorrectly. Unknown bookmark name. Please see Debugging in WordPress for more information. (This message was added in version 6.2.0.) in /var/www/html/wp-includes/functions.php on line 6080 Deprecated: strcasecmp(): Passing null to parameter #1 ($string1) of type string is deprecated in /var/www/html/wp-content/plugins/gutenberg/lib/compat/wordpress-6.5/blocks.php on line 91 Notice: Function WP_HTML_Tag_Processor::seek was called incorrectly. Unknown bookmark name. Please see Debugging in WordPress for more information. (This message was added in version 6.2.0.) in /var/www/html/wp-includes/functions.php on line 6080
```

### Environment info

_No response_

Yes

Yes

talldan:
## What?
Fixes #61303

An issue for block bindings and pattern overrides is when a button block's text is left empty, the editor saves no html due to a conditional return in the block's save function.

When a binding tries dynamically to update a button block in this state, it won't be able to due to the missing content.

This PR changes the button block so that the conditional rendering is implemented dynamically on the server. This allows block bindings to dynamically change the content of the block, even when there's empty text.

## How?
Uses the tag processor to check for inner content.

The `text` of a button block is a sourced attribute, so the sever is unable to parse the attribute value, so it needs to be read from the content.

## Testing Instructions
1. Create a synced pattern with an empty button block (no text defined)
2. Allow overrides for the button block
3. Save the pattern
4. Create a new post and insert the pattern you just created
5. Enter some text into the button
6. Preview the post

Expected: The button should appear with the override text.

As a general smoke test, it's also worth checking that buttons outside of patterns continue to work as expected when they have and do not have text.

You should also be able to add a button that has text to a pattern, and then override it with no text, and it won't render in the frontend

### Editor
Screenshot 2024-06-03 at 4 26 47 pm

### Frontend
Screenshot 2024-06-03 at 4 27 58 pm

### DIFF
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -22,7 +22,6 @@ function render_block_core_button( $attributes, $content ) {
 	/*
 	 * The button block can render an `<a>` or `<button>` and also has a
 	 * `<div>` wrapper. Find the a or button tag.
-	 *
 	 */
 	$tag = null;
 	while ( $p->next_tag() ) {
@@ -32,29 +31,32 @@ function render_block_core_button( $attributes, $content ) {
 		}
 	}
 
-	// If this happens, the likelihood is there's no block content.
+	/*
+	 * If this happens, the likelihood is there's no block content,
+	 * or the block has been modified by a plugin.
+	 */
 	if ( null === $tag ) {
-		return '';
+		return $content;
 	}
 
 	// Build a string of the inner text.
 	$text = '';
 	while ( $p->next_token() ) {
-		switch ( $p->get_token_name() ) {
-			case '#text':
-				$text .= $p->get_modifiable_text();
-				break;
+		$token_name = $p->get_token_name();
+		// Stop when encountering the closing tag.
+		if ( $tag === $token_name ) {
+			break;
+		}
 
-			case 'BR':
-				$text .= '';
-				break;
+		if ( '#text' === $token_name ) {
+			$text .= $p->get_modifiable_text();
+			continue;
 		}
 	}
 
 	/*
 	 * When there's no text, render nothing for the block.
-	 * It's this way because an anchor or button element without text results
-	 * in poor accessibility.
+	 * See https://github.com/WordPress/gutenberg/issues/17221.
 	 */
 	if ( '' === trim( $text ) ) {
 		return '';
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -5,7 +5,7 @@
  * @package WordPress
  */
 
- /**
+/**
  * Renders the `core/button` block on the server,
  *
  * @since 6.6.0
@@ -16,7 +16,7 @@
  *
  * @return string The block content.
  */
-function render_block_core_button( $attributes, $content, $block ) {
+function render_block_core_button( $attributes, $content ) {
 	$p = new WP_HTML_Tag_Processor( $content );
 
 	/*
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * Server-side rendering of the `core/button` block.
+ *
+ * @package WordPress
+ */
+
+ /**
+ * Renders the `core/button` block on the server,
+ *
+ * @since 6.6.0
+ *
+ * @param array    $attributes The block attributes.
+ * @param string   $content    The block content.
+ * @param WP_Block $block      The block object.
+ *
+ * @return string The block content.
+ */
+function render_block_core_button( $attributes, $content, $block ) {
+	$p = new WP_HTML_Tag_Processor( $content );
+
+	if ( ! $p->next_tag( 'a' ) ) {
+		return '';
+	}
+
+	$text = '';
+	while ( $p->next_token() ) {
+		switch ( $p->get_token_name() ) {
+			case '#text':
+				$text .= $p->get_modifiable_text();
+				break;
+
+			case 'BR':
+				$text .= '';
+				break;
+		}
+	}
+
+	// When there's no `text` set, avoid rendering an inaccessible button.
+	if ( trim( $text ) === '' ) {
+		return '';
+	}
+
+	return $content;
+}
+
+/**
+ * Registers the `core/button` block on server.
+ *
+ * @since 6.6.0
+ */
+function register_block_core_button() {
+	register_block_type_from_metadata(
+		__DIR__ . '/button',
+		array(
+			'render_callback' => 'render_block_core_button',
+		)
+	);
+}
+add_action( 'init', 'register_block_core_button' );
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -39,26 +39,15 @@ function render_block_core_button( $attributes, $content ) {
 		return $content;
 	}
 
-	// Build a string of the inner text.
-	$text = '';
-	while ( $p->next_token() ) {
-		$token_name = $p->get_token_name();
-		// Stop when encountering the closing tag.
-		if ( $tag === $token_name ) {
-			break;
-		}
-
-		if ( '#text' === $token_name ) {
-			$text .= $p->get_modifiable_text();
-			continue;
-		}
-	}
+	// If the next token is the closing tag, the button is empty.
+	$p->next_token();
+	$is_empty = $p->get_token_name() === $tag;
 
 	/*
 	 * When there's no text, render nothing for the block.
 	 * See https://github.com/WordPress/gutenberg/issues/17221.
 	 */
-	if ( '' === trim( $text ) ) {
+	if ( $is_empty ) {
 		return '';
 	}
 
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -19,10 +19,25 @@
 function render_block_core_button( $attributes, $content, $block ) {
 	$p = new WP_HTML_Tag_Processor( $content );
 
-	if ( ! $p->next_tag( 'a' ) ) {
+	/*
+	 * The button block can render an `<a>` or `<button>` and also has a
+	 * `<div>` wrapper. Find the a or button tag.
+	 *
+	 */
+	$tag = null;
+	while ( $p->next_tag() ) {
+		$tag = $p->get_tag();
+		if ( 'A' === $tag || 'BUTTON' === $tag ) {
+			break;
+		}
+	}
+
+	// If this happens, the likelihood is there's no block content.
+	if ( null === $tag ) {
 		return '';
 	}
 
+	// Build a string of the inner text.
 	$text = '';
 	while ( $p->next_token() ) {
 		switch ( $p->get_token_name() ) {
@@ -36,8 +51,12 @@ function render_block_core_button( $attributes, $content, $block ) {
 		}
 	}
 
-	// When there's no `text` set, avoid rendering an inaccessible button.
-	if ( trim( $text ) === '' ) {
+	/*
+	 * When there's no text, render nothing for the block.
+	 * It's this way because an anchor or button element without text results
+	 * in poor accessibility.
+	 */
+	if ( '' === trim( $text ) ) {
 		return '';
 	}
 
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -40,12 +40,22 @@ function render_block_core_button( $attributes, $content ) {
 	}
 
 	// If the next token is the closing tag, the button is empty.
-	$p->next_token();
-	$is_empty = $p->get_token_name() === $tag;
+	$is_empty = true;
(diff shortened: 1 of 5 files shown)


packages/block-library/src/button/index.php
```diff
@@ -0,0 +1,79 @@
...
+		switch ( $p->get_token_name() ) {
+			case '#text':
+				$text .= $p->get_modifiable_text();
+				break;
+
+			case 'BR':
+				$text .= '';
+				break;
+		}
+	}
```
talldan: I haven't written a lot of tag processor code, so would be happy to receive feedback on this.

packages/block-library/src/button/index.php
```diff
@@ -0,0 +1,79 @@
...
+	while ( $p->next_tag() ) {
+		$tag = $p->get_tag();
+		if ( 'A' === $tag || 'BUTTON' === $tag ) {
+			break;
+		}
+	}
+
+	// If this happens, the likelihood is there's no block content.
+	if ( null === $tag ) {
+		return '';
```
dmsnell: why not render `$content` here? a viable alternative is that some other plugin already modified the button. do we want to wipe it out entirely just because there's no more `A` or `BUTTON`? e.g. maybe there's now a `<myplugin-button>` in there

packages/block-library/src/button/index.php
```diff
@@ -0,0 +1,79 @@
...
+			case 'BR':
+				$text .= '';
+				break;
+		}
+	}
+
+	/*
+	 * When there's no text, render nothing for the block.
+	 * It's this way because an anchor or button element without text results
+	 * in poor accessibility.
```
dmsnell: does this hold even when there is CSS with the `content` property set? maybe there is additional code involved somewhere else. is removing the content without any indicating the better accessible choice?

packages/block-library/src/button/index.php
dmsnell: this is great @talldan! here are some thoughts. it _looks_ like you're attempting to remove all content inside the button except for the plain text content. that is, no formatting, no spans, no inline images, etc… I don't speak to whether that's a great idea or not, but I can provide feedback for the HTML API use in doing so.

first of all, this is technically the domain of the HTML Processor, which knows HTML structure, but since that's not really ready yet this will suffice. you did the right thing using the Tag Processor.

this code, however, is going to grab only the first text nodeat the start of the `A` or `BUTTON`. Maybe we don't expect any other content, but if you want to (unreliably) limit it to the inner contents, then you can add one more step which would improve the reliability a little bit and capture more text.

```php
// Build a string of the inner text.
$text = '';
while ( $p->next_token() ) {
	switch ( $p->get_token_name() ) {
		/*
         * Neither `A` nor `BUTTON` can nest, so it doesn't matter
         * if these are opening or closing tags. Both will close
         * the open element.
         */
		case $tag:
			break;

		case '#text':
			$text .= $p->get_modifiable_text();
			continue;

		case 'BR':
			$text .= "\n";
			continue;
	}
}
```

dmsnell (commented):
This is a change I don't fully understand, so I'm probably overlooking something, but it looks like it wipes out all buttons, which seems non-optimal. Should we be checking for some condition before trashing the HTML?

Also even when we choose to, it might help to provide more insightful checks than "we didn't find an `A` or a `BUTTON`" in case plugin code has modified the markup in a way we don't expect. Perhaps we assume that if the tags are missing then someone else modified the content intentionally and we should leave it alone.

If this is for the block bindings, would it help avoid unintentional corruption if we looked for some block attribute indicating that we intend to perform Block Binding on it?

I'm just a bit nervous from this that the effect is a broad stroke that introduces normal corruption for the sake of fixing one much-rarer edge case. But like I acknowledge, I'm probably missing something major.

SantosGuillamot:
If I am not mistaken, we don't need to process the content and modify it. Right now, the modifications of the block attributes to apply the bindings are done during the rendering. This means that we can check if the text attribute exists (or if it is empty). In that case, we just return null. And if not, we just return the content. Something like this:

```php
if ( empty( $attributes["text"] ) ) {
    return null;
} else {
    return $content;
}
```

I quickly tested and it seems to work, but I didn't take a close look at the different scenarios. Do you think something like that could work?

talldan:
> I quickly tested and it seems to work, but I didn't take a close look at the different scenarios. Do you think something like that could work?

@SantosGuillamot My first hope is that the solution would be as simple as that, but I don't think it is. My understanding is that what you propose will only work when bindings are applied (`$attributes[ 'text' ]` only has a value due to the block binding). When the button block is used normally without bindings it will still need the conditional rendering to work as expected.

SantosGuillamot:
Oh, okay. I had assumed that `attributes[ "text "]` existed for normal buttons as well. Trying to understand better what we want, is this right?

* If attributes[ "text" ] exist: return content.
* If the BUTTON or A tag HAS inner text: return content.
* If the BUTTON or A tag HAS NOT inner text: return null.

If that's the case and I'm understanding it correctly, @dmsnell can help better identify the easiest way to do that check now. I can't say if the current approach is the right one or not.

If I am not mistaken, and according to [this comment](https://github.com/Wordpress/wordpress-develop/blob/trunk/src/wp-includes/html-api/class-wp-html-processor.php#L25-L30), in the future, we could potentially rely on the HTML processor to read the inner text (something like `$p->get_inner_text()`). Ideally, I imagine something like this:

```php
if ( ! empty( $attributes[ "text" ] ) ) {
    return $content;
}

while ( $p->next_tag() ) {
    $tag = $p->get_tag();
    if ( 'A' === $tag || 'BUTTON' === $tag ) {
        if ( $p->get_inner_text() ){
            return $content;
        } else {
            return null;
        }
    }
}
```

But again, I believe this is not possible right now and I am not that familiar with the HTML API, so it may be wrong.

packages/block-library/src/button/index.php
talldan: > case $tag:
>	break;

Thanks for the feedback! Should this also break out of the while loop in addition to the switch? I believe in PHP `break 2` can be used to break two levels, but I might change it from a `switch` to an `if` statement.

talldan: > it looks like you're attempting to remove all content inside the button except for the plain text content. that is, no formatting, no spans, no inline images, etc… I don't speak to whether that's a great idea or not, but I can provide feedback for the HTML API use in doing so.

The goal in this PR is for the replace the old static implementation (removed in this PR):
https://github.com/WordPress/gutenberg/blob/dbf201449e9736f672b61e422787d47659db327a/packages/block-library/src/button/save.js#L34-L36

Retaining the HTML from the save function allows for bindings to work with empty buttons.

It looks like the conditional logic introduced as a result of this issue - https://github.com/WordPress/gutenberg/issues/17221. 

It's not an accessibility reason. How many years have I incorrectly believed that?!  I'll update the comment in the code.

packages/block-library/src/button/index.php
talldan: Agreed 👍 

I've made that change.

packages/block-library/src/button/index.php
talldan: As mentioned in another comment, it originates from https://github.com/WordPress/gutenberg/issues/17221.

I've updated this comment so that it doesn't mention accessibility, instead I link straight to that issue.

packages/block-library/src/button/index.php
talldan: I did some further testing and realized there's also the case of inline images. I've now simplified the code down to just this:
```php
// If the next token is the closing tag, the button is empty.
$p->next_token();
$is_empty = $p->get_token_name() === $tag;

if ( $is_empty ) {
	return '';
}
```

Which should better match how `RichText.isEmpty` works (even a space is considered content) - https://github.com/WordPress/gutenberg/blob/trunk/packages/rich-text/src/is-empty.js.

I should've endeavoured to match that implementation first.

Anything I might be overlooking here? I'll work on adding some tests for this.

kevin940726:
From my testing in a pattern with the empty second button.

```html
<!-- wp:block {"ref":199,"content":{"2":{"text":"123"}}} /-->
```

In the editor it looks correctly:

image

However, on the frontend, the second button disappears and magically erases the first character.
image

Is there something I'm doing wrong?

talldan:
> Is there something I'm doing wrong?

I couldn't reproduce it, and I'm not sure how that could happen.

I'll do some more testing. It has been pretty challenging working on this fix as my local xdebug isn't working for block PHP. Not sure what's causing that.

talldan:
There's one caveat that I'll note for testing, this won't work for patterns already saved before you checked out this branch. Those patterns are already missing the button markup, so there's very little that can be done to fix them (beyond making the button block completely dynamic).

And as a side note, I managed to figure out what was happening with xdebug. A breakpoint needs to be set in the `build/block-library/blocks/button.php` file, it won't work in the source file. It's a bit of an inconvenience, not sure if there's a way to solve that.

packages/block-library/src/button/index.php
dmsnell: It's probably good enough, but keep in mind that there could be comments in there too.

```php
$is_empty = true;
while ( $p->next_token() && $tag !== $p->get_token_name() && $is_empty ) {
	if ( '#comment' !== $p->get_token_type() ) {
		// Anything else implies this is not empty.
		$is_empty = false;
	}
}
```

kevin940726:
> I couldn't reproduce it, and I'm not sure how that could happen.

I can only reproduce it if the override for the empty button starts with a number. e.g. `Button 2` works correctly, but `2Button 2` doesn't work for some reason. Could anyone verify this?

talldan:
> I can only reproduce it if the override for the empty button starts with a number. e.g. Button 2 works correctly, but 2Button 2 doesn't work for some reason. Could anyone verify this?

Yep, I can reproduce it now, thanks for the extra info!

talldan:
I can reproduce that in `trunk` as well using a non-empty button, so it must be a problem with the bindings code.

kevin940726:
> I can reproduce that in `trunk` as well using a non-empty button, so it must be a problem with the bindings code.

Yep. I think it's related to https://github.com/WordPress/gutenberg/issues/62347, thanks for @Mamaduka for the mention!

SantosGuillamot:
I just answered in [the mentioned issue](https://github.com/WordPress/gutenberg/issues/62347#issuecomment-2151562376):

>I believe it is caused by this refactoring of the replacement logic: https://github.com/WordPress/gutenberg/pull/61236. My fault for using regex 🤦
>
>The idea behind that pull request was to simplify the logic to support more attributes, but that didn't make it to 6.6. We should probably go back to the previous implementation and refactor the logic once the HTML API adds new methods to help with that.

kevin940726 (approved):
Works for me! I'll leave the binding code for others with more knowledge to review though.

SantosGuillamot:
I've tested it and it seems to work fine! Regarding bindings, I think it is ready. But I'll defer it to @dmsnell because most of the changes are related to how to use the HTML Tag Processor.

dmsnell (approved):
I could have added an approval before, but mainly I'm only examining the HTML API usage, and it seems fine what's happening here, so from that angle the patch is good.

ellatrix:

PR #62223, opened 2024-06-03
Try: Contextual frame bg color to avoid artifacting.

jasmussen:
## What?

When you have a theme with a dark background, the scale-up animation of the frame in the site editor has a white halo artifacting itself due to rounding math in the scale-up process:

https://github.com/WordPress/gutenberg/assets/1204802/a47b117f-e761-4317-97af-a85f12ebf735

This is due to the light gray background that is applied to the frame background peeking out.

This PR solves that by contextually applying a dark background to the framed contexts:

https://github.com/WordPress/gutenberg/assets/1204802/384cd5da-2704-4316-b8fc-53441e09ff8a

What about when you're using a light theme, you might ask? The artifacting is still addressed in that case because the outer background is dark. So it's only an issue when it's a dark background, a light frame background, and then a dark content background inside.

## Testing Instructions

Use a dark background, test the site editor, mouse over the frame to scale it up.

Note that previously the light gray background was applied specifically to address an issue with zoom out mode. Please do test that this still works.

### DIFF
--- a/packages/block-editor/src/components/block-canvas/style.scss
+++ b/packages/block-editor/src/components/block-canvas/style.scss
@@ -1,7 +1,13 @@
 iframe[name="editor-canvas"] {
+	box-sizing: border-box;
 	width: 100%;
 	height: 100%;
 	display: block;
-	background-color: $gray-300;
-	box-sizing: border-box;
+
+	// Show a dark background in "frame" mode to avoid edge artifacts.
+	// Show a gray background in all other contexts, such as zoom.
+	background-color: $gray-900;
+	.is-full-canvas & {
+		background-color: $gray-300;
+	}
 }
--- a/packages/block-editor/src/components/block-canvas/style.scss
+++ b/packages/block-editor/src/components/block-canvas/style.scss
@@ -3,11 +3,5 @@ iframe[name="editor-canvas"] {
 	width: 100%;
 	height: 100%;
 	display: block;
-
-	// Show a dark background in "frame" mode to avoid edge artifacts.
-	// Show a gray background in all other contexts, such as zoom.
-	background-color: $gray-900;
-	.is-full-canvas & {
-		background-color: $gray-300;
-	}
+	background-color: transparent;
 }
--- a/packages/edit-site/src/components/layout/style.scss
+++ b/packages/edit-site/src/components/layout/style.scss
@@ -4,6 +4,11 @@
 	color: $gray-400;
 	display: flex;
 	flex-direction: column;
+
+	// Show a dark background in "frame" mode to avoid edge artifacts.
+	&:not(.is-full-canvas) .editor-visual-editor {
+		background: $gray-900;
+	}
 }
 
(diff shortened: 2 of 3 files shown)


MaggieCabrera (approved):
Works fine with me and the Safari issue is still fixed with this solution

jasmussen:
Thanks for the review. I'll let this one sit for a bit just to give others a chance to chime in.

packages/block-editor/src/components/block-canvas/style.scss
```diff
@@ -1,7 +1,13 @@
...
 	width: 100%;
 	height: 100%;
 	display: block;
-	background-color: $gray-300;
-	box-sizing: border-box;
+
+	// Show a dark background in "frame" mode to avoid edge artifacts.
+	// Show a gray background in all other contexts, such as zoom.
+	background-color: $gray-900;
+	.is-full-canvas & {
```
youknowriad: is-full-canvas is a site editor class so this style should probably not live here. (same for the other file).

jasmussen:
Good feedback, thanks for testing. I changed the approach to leverage the fact that the class was site editor only. What do you think?

packages/block-editor/src/components/block-canvas/style.scss
```diff
@@ -1,7 +1,7 @@
 iframe[name="editor-canvas"] {
+	box-sizing: border-box;
 	width: 100%;
 	height: 100%;
 	display: block;
-	background-color: $gray-300;
```
youknowriad, line 5: So this background is not needed at all, even on the post editor?

youknowriad (commented):
This approach looks good to me as long as the removal of the background color is ok for the post editor too.

packages/block-editor/src/components/block-canvas/style.scss
MaggieCabrera: I was afraid that this would break Safari, but this is what I see:

Screenshot 2024-06-04 at 11 41 57

jasmussen: In my testing, these contexts inherit from the other class.

What do you say, good to land?

Edit: as always, happy to follow up.

jasmussen:
2024-06-05 08 06 04 local-wordpress local 40a060a04bc5

This one sat for a bit to gather feedback, but re-testing it this morning, I still can't reproduce any issues. Site editor in site view has dark canvas, all contexts in edit view inc. post editor and safari have the gray view.

So I'll merge and be on deck for any followups.

ellatrix:

jasmussen:
I'd like it backported, because it's a great fix for visually dark themes in the site editor. So I'll add the label, thanks for the reminder. It's also an entirely safe bugfix. But if it's chosen to not be backported, that also okay.

ellatrix:
Let's mark it as a bug fix then, the enhancement label confused me a bit

ellatrix:

PR #62225, opened 2024-06-03
Top toolbar: fix half a pixel artifacting of the bottom border.

jasmussen:
## What?

In distraction free mode, or with the top toolbar enabled, there's a half pixel white artifact below the toolbar which causes the bottom border to not be coherent:
broken bottom border

This PR fixes that by applying a max height, with no discernible side effects:

fixed bottom border

### DIFF
--- a/packages/components/src/toolbar/toolbar/style.scss
+++ b/packages/components/src/toolbar/toolbar/style.scss
@@ -10,7 +10,6 @@
 
 	&.is-unstyled {
 		border: none;
-		background: transparent;
 
 		& > .components-toolbar-group {
 			border-right: none;
--- a/packages/components/src/toolbar/toolbar/style.scss
+++ b/packages/components/src/toolbar/toolbar/style.scss
@@ -10,6 +10,7 @@
 
 	&.is-unstyled {
 		border: none;
+		background: transparent;
 
 		& > .components-toolbar-group {
 			border-right: none;
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 - Add `text-wrap: balance` fallback to all instances of `text-wrap: pretty` for greater cross browser compatibility. ([#62233](https://github.com/WordPress/gutenberg/pull/62233))
 
+### Bug Fixes
+
+-   `Toolbar`: Fix visual glitch in top toolbar mode ([#62225](https://github.com/WordPress/gutenberg/pull/62225)).
+
+
 ## 28.0.0 (2024-05-31)
 
 ### Breaking Changes
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -6,11 +6,6 @@
 
 - Add `text-wrap: balance` fallback to all instances of `text-wrap: pretty` for greater cross browser compatibility. ([#62233](https://github.com/WordPress/gutenberg/pull/62233))
 
-### Bug Fixes
-
--   `Toolbar`: Fix visual glitch in top toolbar mode ([#62225](https://github.com/WordPress/gutenberg/pull/62225)).
-
-
 ## 28.0.0 (2024-05-31)
 
 ### Breaking Changes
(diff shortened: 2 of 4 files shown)


t-hamano (commented):
This issue probably depends on the computer and display, but on my setup the bottom border was completely hidden.

image

Perhaps it would be more reasonable to subtract 1px from this value?

https://github.com/WordPress/gutenberg/blob/da98bdb025937d8bb0abc41a187792e75e17731d/packages/editor/src/components/collapsible-block-toolbar/style.scss#L5

jasmussen:
Thanks for the suggestion. I tried it, and it looked as if it worked so I committed it. But on closer inspection, that causes the buttons to be offset by a little bit:

suggested-fix

Still closer investigation surfaced that this is only an issue in distraction free mode, where the toolbar is positioned and animated using translate. This particular aspect makes it subject to subpixel rounding math, which is probably what's causing this glitch in the first place. Another option is removing the white background, what do you think?

For now, I've restored the previous fix, which I don't love, but does make things still line up:

previous-fix

t-hamano:
> Another option is removing the white background

This approach also seems reasonable. By adding `background-color: transparent` here, we will disable the white background only if Top Toolbar is enabled.

https://github.com/WordPress/gutenberg/blob/da98bdb025937d8bb0abc41a187792e75e17731d/packages/editor/src/components/collapsible-block-toolbar/style.scss#L8-L9

I'm sure @jeryj know a lot about toolbars, so if you have any good ideas please let me know 🙏

jeryj:
If `background-color: transparent;` when top toolbar is on addresses the issue, that seems like a great fix to me!

jasmussen:
That fix seems to work well!

Screenshot 2024-06-04 at 09 00 48

The only question remaining, is probably one for @draganescu and related to how 3rd party plugins in some cases inject buttons there, such as custom pattern insterter tools. I vaguelly recall us adding some features to the top toolbar mode that _covered those buttons_ unless manually collapsed. Would that regress if we remove the background color? Anyone recall a plugin that still does this, or was the behavior for how those buttons are forcefully added changed?

packages/components/src/toolbar/toolbar/style.scss
```diff
@@ -10,6 +10,7 @@
 
 	&.is-unstyled {
 		border: none;
+		background: transparent;
```
t-hamano: The white background doesn't originally exist in the `Toolbar` component itself. It exists in the `BlockToolbar` component, a block that extends the `Toolbar` component.

https://github.com/WordPress/gutenberg/blob/14f9a1c1bfce1f02130150bf8b9d5454cca32dc8/packages/block-editor/src/components/block-toolbar/style.scss#L66

And I think the white background only needs to be overridden when this block toolbar is rendered in the header, so I think adding it in the place suggested in [this comment](https://github.com/WordPress/gutenberg/pull/62225#issuecomment-2145021276) might be a good idea.

jasmussen: Thank you for your patience. Moved it there, and it still seems to work for me.

t-hamano (approved):
LGTM!

Nit: Since this PR did not ultimately change anything in the `components` package, the CHANGELOG entry can be removed.

draganescu (approved):
Tested this as well and the borders look perfect!

jeryj:
@jasmussen 
> I vaguelly recall us adding some features to the top toolbar mode that covered those buttons unless manually collapsed. Would that regress if we remove the background color? Anyone recall a plugin that still does this, or was the behavior for how those buttons are forcefully added changed?

The original implementation of the top toolbar was a `position: absolute` of the default block popover. Maybe that's what you're thinking of, as that would have covered up things rendered underneath of it. That got changed to rendering the toolbar directly in the header, so things will reflow natively. There shouldn't be any instance of something rendering "underneath/behind" the toolbar.

jasmussen:
Thanks for the merge and input + reviews.

ellatrix:

PR #60729, opened 2024-04-14
Change header height from 60px to 64px

t-hamano:
Fixes #58293
Part of #46734

## What?

This PR changes the block editor header height from 60px to 64px. 

## Why?

As part of the unification of UI component sizes proposed in https://github.com/WordPress/gutenberg/issues/46734. [This comment](https://github.com/WordPress/gutenberg/pull/57444#issuecomment-1871881388) also suggests changing the header height to 64px.

## How?

Changed the value of the `$header-height` CSS variable from 60px to 64px. At the same time, I made adjustments to the areas affected by this change.

## Testing Instructions

We'll probably need to do extensive testing to make sure this change doesn't cause a 4px shift. I haven't found any issues in my testing, but if you have any concerns please let me know.

> [!NOTE]
> As reported in #55568, the `npm run dev` command does not detect changes to the `base-styles` package. So you need to run `npm run dev` again after checking out the branch.

PR #57444, opened 2023-12-29
Apply custom scroll style to fixed header block toolbar

t-hamano:
## What?
This PR applies the custom scrolling style to the fixed block toolbar. I think this improves the appearance, especially on **Windows**.

> [!NOTE]
> This PR was originally to simply introduce a custom scrollbar. However, after some discussion, we have made a CSS adjustment in all editor instances to prevent the toolbar from shifting when a scrollbar appears or when a button is focused.

### Before (Chrome)

chrome-before

### After (Chrome)

chrome-after

## Why?

In the Chrome browser on Windows, the scrollbar has a _physical_ width/height. So the top toolbar buttons are pushed far up and look unnatural to me.

Additionally, this toolbar actually contains two `overflow` properties:

https://github.com/WordPress/gutenberg/blob/83e1a2c064c2233e8cd82dcccf4cd555c3497be1/packages/edit-site/src/components/header-edit-mode/style.scss#L206-L208

https://github.com/WordPress/gutenberg/blob/83e1a2c064c2233e8cd82dcccf4cd555c3497be1/packages/block-editor/src/components/block-toolbar/style.scss#L76-L77

In the case of the Site Editor, the `overfow` property value of the upper container (`.selected-block-tools-wrapper `) is `scroll` instead of `hidden`, so when element in the container overflow, the two scroll bars overlap, making it look even more unnatural.

double-scrollbar

## How?
First, I changed the `overflow `property of the upper container (`.selected-block-tools-wrapper`) from `scroll` to `hidden`. This will eliminate double scrollbars.

Additionally, to make this thick scrollbar appear more natural, I applied a `custom-scrollbars-on-hover` mixin to the inner scroll container (`.block-editor-block-toolbar`). If an element overflows, a scrollbar will appear without the need for mouseover. Personally, I feel that this is easier to understand than the behavior of disappearing the scroll bar when the mouse is out.

## Testing Instructions

- Open the Site Editor.
- Insert some block and select it. For example, image blocks are easier to test because they have a long block toolbar.
- Change the browser width in various browsers.
- For Mac OS, change the OS settings so that the scroll bar is always displayed.
- When the block toolbar is not fixed, the scrollbar should never be displayed.

## Screenshots or screencast 

Note: This shows scrollbar behavior in WindowsOS and Chrome.

### Before

https://github.com/WordPress/gutenberg/assets/54422211/a457de09-5b55-4cbd-8852-dc6352fd3e1a

### After

[...]

PR #62226, opened 2024-06-03
Try: Fix mover positioning.

jasmussen:
## What?

The mover position when Top Toolbar is enabled is current pushed upwards:

movers pushed upwards

This PR improves but doesn't entirely fix this issue, by addressing some metrics.

movers better positioned

This is a draft because https://github.com/WordPress/gutenberg/pull/60729 appears to fix this too. Let me know your thoughts.

### DIFF
--- a/packages/editor/src/components/collapsible-block-toolbar/style.scss
+++ b/packages/editor/src/components/collapsible-block-toolbar/style.scss
@@ -64,8 +64,9 @@
 		// Move up a little to prevent the toolbar shift when focus is on the vertical movers.
 		@include break-small() {
 			&:not(.is-horizontal) .block-editor-block-mover__move-button-container {
+				height: 40px;
 				position: relative;
-				top: -10px;
+				top: -5px; // Should be -4px, but that causes scrolling when focus lands on the movers, in a 60px header.
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/collapsible-block-toolbar/style.scss
```diff
@@ -64,8 +64,9 @@
 		// Move up a little to prevent the toolbar shift when focus is on the vertical movers.
 		@include break-small() {
 			&:not(.is-horizontal) .block-editor-block-mover__move-button-container {
+				height: 40px;
```
t-hamano:
```suggestion
				height: $grid-unit-50;
```

Wouldn't it be better not to have a hardcoded value?

t-hamano (approved):
I think this is a great approach. It also doesn't shift the toolbar on Windows OS, which was discussed many times in #57444.

https://github.com/WordPress/gutenberg/assets/54422211/a2ab638d-92e5-4a5d-b6d6-61d6a3139ea7

I also tested it with `top: -4px`, but the toolbar still shifted, so I think `-5px` is optimal.

https://github.com/WordPress/gutenberg/assets/54422211/416dc407-1184-4d12-b554-47db3177d6bb

jasmussen:
Thanks for the input. I'm happy to land this as we wait for https://github.com/WordPress/gutenberg/pull/60729, though let's keep this PR in mind for if we revisit that, we may be able to remove a lot of this code.

ellatrix:

ellatrix:

PR #60721, opened 2024-04-13
Pattern overrides: use block binding editing API

ellatrix:
## What?
Updates the pattern (`wp:block`) block to use the block bindings API (from #60724) in the editor.

## Why?
This fixes a specific use case where two blocks (e.g. two paragraphs) may use the same names for the pattern overrides. In this case the two blocks should be synchronized when editing, but this is broken in `trunk`.

## How?
Add new pattern overrides binding source properties for getValue and setValues

## Testing Instructions
1. Create a synced pattern with two paragraphs
2. Enable overrides for the two paragraphs and use the same block names for both
3. Save the pattern
4. Create a new post and insert the pattern
5. Try typing in one of the paragraphs

Expected: The two paragraphs can be synchronized.

PR #62229, opened 2024-06-03
Remove unused `syncDerivedUpdates` action

SantosGuillamot:
## What?
Remove `syncDerivedUpdates` private action.

## Why?

In [this pull request](https://github.com/WordPress/gutenberg/pull/60721), how pattern overrides are handled was changed and `syncDerivedUpdates` action doesn't seem to be used anymore. I assume it it better to remove it and add it back if needed.

## How?
Just removing the code.

## Testing Instructions
Tests should pass.

### DIFF
--- a/packages/block-editor/src/store/undo-ignore.js
+++ b/packages/block-editor/src/store/undo-ignore.js
@@ -1,4 +1,4 @@
 // Keep track of the blocks that should not be pushing an additional
 // undo stack when editing the entity.
-// See the implementation of `syncDerivedUpdates` and `useBlockSync`.
+// See the implementation of `useBlockSync`.
 export const undoIgnoreBlocks = new WeakSet();
--- a/packages/block-editor/src/store/undo-ignore.js
+++ b/packages/block-editor/src/store/undo-ignore.js
@@ -1,4 +1,0 @@
-// Keep track of the blocks that should not be pushing an additional
-// undo stack when editing the entity.
-// See the implementation of `useBlockSync`.
-export const undoIgnoreBlocks = new WeakSet();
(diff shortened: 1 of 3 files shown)


packages/block-editor/src/store/undo-ignore.js
```diff
@@ -1,4 +1,4 @@
 // Keep track of the blocks that should not be pushing an additional
 // undo stack when editing the entity.
-// See the implementation of `syncDerivedUpdates` and `useBlockSync`.
+// See the implementation of `useBlockSync`.
 export const undoIgnoreBlocks = new WeakSet();
```
kevin940726: I believe this variable isn't needed anymore either?

SantosGuillamot: I didn't remove it because it was being used in `useBlockSync`. But maybe we can remove [these lines](https://github.com/WordPress/gutenberg/blob/da98bdb025937d8bb0abc41a187792e75e17731d/packages/block-editor/src/components/provider/use-block-sync.js#L277-L280) as well? You know better than me 😄

kevin940726: Yep! They are not used anywhere else and were introduced in the same PR.

SantosGuillamot: Perfect! I just removed it in [this commit](https://github.com/WordPress/gutenberg/pull/62229/commits/d76e28872d72fd983b95f539e07f980774755b85). Let's see if tests pass.

kevin940726 (approved):
Thanks!

youknowriad:
@SantosGuillamot This is my friendly reminder to follow this https://github.com/WordPress/gutenberg/pull/62229#issuecomment-2144901068 :)

SantosGuillamot:
I'm really sorry for that. Totally my fault. I'll take it into account for future pull requests. And thanks for the reminder.

PR #62231, opened 2024-06-03
Update URL to uppercase.

jasmussen:
## What?

Noticed that we are using URL in lowercase in this placeholder:

lowercase url

Shouldn't that be uppercase, since it's an acronym? This PR makes it so:

uppercase URL

## Testing Instructions

Insert a link, notice the placeholder text having uppercase URL.

### DIFF
--- a/packages/block-editor/src/components/link-control/search-input.js
+++ b/packages/block-editor/src/components/link-control/search-input.js
@@ -125,7 +125,7 @@ const LinkControlSearchInput = forwardRef(
 					className={ className }
 					value={ value }
 					onChange={ onInputChange }
-					placeholder={ placeholder ?? __( 'Search or type url' ) }
+					placeholder={ placeholder ?? __( 'Search or type URL' ) }
 					__experimentalRenderSuggestions={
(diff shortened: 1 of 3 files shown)


paaljoachim:
Good one!

t-hamano (approved):
This seems reasonable to me, and from a quick look at the wordpress-develop core code, it's consistent in the upper case.

jasmussen:
Thank you for the reviews. I'll let this one sit for a little bit and let others chime in, it's been like this for a while.

paaljoachim:
I went looking for another place I thought the URL might be mentioned. Here is one.
It uses the uppercase URL.

Screenshot 2024-06-03 at 13 34 05

youknowriad:
@jasmussen This is my friendly reminder to follow this https://github.com/WordPress/gutenberg/pull/62231#issuecomment-2144923726

jasmussen:
My gosh. I've occasionally seen those automatically included, so the fact that they weren't here was my mistake. Apologies, I will do better, and thank you for the reminder.

ellatrix:

Issue #62202, opened 2024-06-01
New block created with @wordpress/scripts@28 don't show up on wordpress


originaliko:
### Description

I tried to create a new plugin / block with create-block and the block is not available in the back-end. There's no build error, php error or javascript error.

### Step-by-step reproduction instructions

1 : Create a new wordpress plugin "npx @wordpress/create-block test-plugin"
2: cd test-plugin
3 : npm start
3 : activate plugin test-plugin 
4 : edit a new page in wp-admin
5 : try to add the new block "/test-plugin"
6: nothing append 
---------
step to correct 
- edit package.json set devedepencies "@wordpress/scripts" to "^27.9.0" 
- npm install
- npm start
- refresh and edit a page
- add the new block "/test-plugin"

### Environment info

Wordpress 6.53.
@wordpress/scripts": "^28.0.0"
node  22.2

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

balazske96:
+1

However if you put the block inside a template html file with it's block name it appears on the front end.

t-hamano:
Thanks for the report. I was able to reproduce this issue.

I built the block with scripts version 27 and 28 and found a difference in the contents of `index.asset.php`. `react` has been replaced with `react-jsx-runtime`.

version 27.9:

```php
<?php return array('dependencies' => array('react', 'wp-block-editor', 'wp-blocks', 'wp-i18n'), 'version' => 'd97fdb9bbdb79e4913a4');
```

version 28.0:

```php
<?php return array('dependencies' => array('react-jsx-runtime', 'wp-block-editor', 'wp-blocks', 'wp-i18n'), 'version' => 'b0e4f78143cfbd1f56d2');
```

If I manually change this to `react`, the block can be inserted into the editor, but a critical error occurs.

![image](https://github.com/WordPress/gutenberg/assets/54422211/85b027be-abb1-4ff7-82d8-0c5785a76692)

realthemes:
With v28 scripts all my blocks in my custom block theme stop working, was there any drastic change that I had to take into account when moving from 27 to 28?

erikyo:
Same here, after updating to v28 custom blocks completely disappeared. 

```
node v20.13.1
npm 10.3.0
```

reverting to `"@wordpress/scripts": "27.8.0"` (my last know working version) fixes the issue

--- edit ---
[...]

PR #62234, opened 2024-06-03
Scripts: Pin the @wordpress/scripts version to a version supported by 6.5

youknowriad:
Related #62202

## What?

WordPress 6.6 was updated to use the new JSX transform but it's not released yet. That said, the latest version of the wp-scripts package targets 6.6 (and the Gutenberg plugin) by default and it's used to build Gutenberg itself so it was updated to use the new JSX.

For this reason, we need to pin the `@wordpress/create-block` tool to a version of wp-scripts that is supported by WordPress 6.5.

## Testing Instructions

1 : Create a new wordpress plugin "npx ../path-to-gutenberg-clone/packages/create-block"
2: cd test-plugin
3 : npm start
3 : activate plugin test-plugin
4 : edit a new page in wp-admin
5 : try to add the new block "/test-plugin"
6: The block should appear.

### DIFF
--- a/packages/create-block/lib/init-wp-scripts.js
+++ b/packages/create-block/lib/init-wp-scripts.js
@@ -16,7 +16,7 @@ module.exports = async ( { slug } ) => {
 	info(
 		'Installing `@wordpress/scripts` package. It might take a couple of minutes...'
 	);
-	await command( 'npm install @wordpress/scripts --save-dev', {
+	await command( 'npm install @wordpress/scripts@27 --save-dev', {
 		cwd,
 	} );
 


ellatrix (approved):
Can't we update `@wordpress/create-block` to work with the new transform?

youknowriad:
@ellatrix we can't because the new transform only works in WP 6.6 (which is not released yet) and also most plugins want to support both 6.6 and 6.5 at least.

ellatrix:
Right. Will the old create-block work with 6.6 (and GB)?

youknowriad:
@ellatrix yes

ellatrix:

gziolo:
I opened a follow-up to remove the limitation for the version used for `@wordpress/scripts` by increasing the minimum supported WordPress version to 6.6:

- https://github.com/WordPress/gutenberg/pull/64920

PR #62235, opened 2024-06-03
Use preview instead of publishing post in block bindings tests

SantosGuillamot:
## What?
Using the preview functionality instead of publishing a post when checking the frontend in block bindings tests.

## Why?
To avoid publishing new posts when it is not needed. As mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/60721#discussion_r1600017842), using the current approach _"won't work with subdirectory installations, and breaks running tests in MAMP."_

## How?
Using `openPreviewPage` util instead of `publishPost`. Additionally, it allows us to avoid having a custom util.

## Testing Instructions
Tests should pass.

### DIFF
--- a/test/e2e/specs/editor/various/block-bindings.spec.js
+++ b/test/e2e/specs/editor/various/block-bindings.spec.js
@@ -30,12 +30,6 @@ test.describe( 'Block bindings', () => {
 		await requestUtils.deactivatePlugin( 'gutenberg-test-block-bindings' );
 	} );
 
-	test.use( {
-		BlockBindingsUtils: async ( { editor, page, pageUtils }, use ) => {
-			await use( new BlockBindingsUtils( { editor, page, pageUtils } ) );
-		},
-	} );
-
 	test.describe( 'Template context', () => {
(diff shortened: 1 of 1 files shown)


artemiomorales (approved):
Looks good to me!

cbravobernal:
You are still creating the posts as drafts and deleting them at the end of the testing. So I'm not sure that this change will affect anything 😅

artemiomorales:
> You are still creating the posts as drafts and deleting them at the end of the testing. So I'm not sure that this change will affect anything 😅

While the functionality is still the same, it's less code, which can make it easier to understand, also less prone to breakage by not relying on utils to click the Save button and dismiss the notice. So I think it's a worthwhile change 😄

SantosGuillamot:
I agree the code looks better.

Additionally, I started working on the pull request based on [this comment](https://github.com/WordPress/gutenberg/pull/60721#discussion_r1600017842) by Ella: _"Is there no preview util? This won't work with subdirectory installations, and breaks running tests in MAMP. We should honour the WP_BASE_URL."_

I should have pointed that out in the description, sorry. Doing it now.

PR #61676, opened 2024-05-15
Editor: Avoid rendering snackbar notices twice in list data views

youknowriad:
closes #60457 

## What?

When manipulating the pages in the "list view", it can result in snackbar notices showing up twice: outside and inside the frame. This PR only shows the notices inside the frame in "edit" mode.

## Testing Instructions

1- Open the "pages" menu in the site editor
2- Rename a page from the actions menu
3- Notice that a snackbar shows up (but only once)

Issue #53198, opened 2023-07-31
Improper snackbar placement with block breadcrumbs disabled


richtabor:
Snackbars sit awkwardly high when block breadcrumbs are disabled. I would expect them to have the same space between the bottom of the notice and the viewport, with or without breadcrumbs.

<img width="538" alt="CleanShot 2023-07-31 at 16 54 28" src="https://github.com/WordPress/gutenberg/assets/1813435/f49a2b82-981d-4a62-ad95-f41b8f86a346">

richtabor:
It's not just breadcrumbs, it's also for snackbars rendering whenever there's no footer/breadcrumbs displaying.

<img width="786" alt="CleanShot 2023-08-03 at 12 03 41" src="https://github.com/WordPress/gutenberg/assets/1813435/f5b6ce37-5c2d-4d04-b9d3-697c11550a93">

PR #62237, opened 2024-06-03
Editor: Cleanup styles and classnames

youknowriad:
Related #52632 

## What?

Both site and post editor pass some classnames to the EditorInterface. The PR intends to unify these classnames as much as possible and cleanup some styles. Users of `EditorInterface` component shouldn't have to provide additional classnames for things to work/show properly.

## Testing Instructions

See inline comments.

### DIFF
--- a/packages/base-styles/_z-index.scss
+++ b/packages/base-styles/_z-index.scss
@@ -28,7 +28,6 @@ $z-layers: (
 	".block-editor-warning": 5,
 	".block-library-gallery-item__inline-menu": 20,
 	".block-editor-url-input__suggestions": 30,
-	".edit-post-layout__footer": 30,
 	".interface-interface-skeleton__header": 30,
 	".interface-interface-skeleton__content": 20,
 	".edit-widgets-header": 30,
@@ -86,13 +85,12 @@ $z-layers: (
 	// Show sidebar above wp-admin navigation bar for mobile viewports:
 	// #wpadminbar { z-index: 99999 }
 	".interface-interface-skeleton__sidebar": 100000,
-	".edit-post-layout__toggle-sidebar-panel": 100000,
 	".editor-layout__toggle-sidebar-panel": 100000,
 	".edit-widgets-sidebar": 100000,
 	".edit-post-layout .edit-post-post-publish-panel": 100001,
 	// For larger views, the wp-admin navbar dropdown should be at top of
 	// the Publish Post sidebar.
-	".edit-post-layout .edit-post-post-publish-panel {greater than small}": 99998,
+	".editor-post-publish-panel {greater than small}": 99998,
 
 	// For larger views, the wp-admin navbar dropdown should be on top of
 	// the multi-entity saving sidebar.
--- a/packages/base-styles/_z-index.scss
+++ b/packages/base-styles/_z-index.scss
@@ -87,7 +87,7 @@ $z-layers: (
 	".interface-interface-skeleton__sidebar": 100000,
 	".editor-layout__toggle-sidebar-panel": 100000,
 	".edit-widgets-sidebar": 100000,
-	".edit-post-layout .edit-post-post-publish-panel": 100001,
+	".editor-post-publish-panel": 100001,
 	// For larger views, the wp-admin navbar dropdown should be at top of
 	// the Publish Post sidebar.
 	".editor-post-publish-panel {greater than small}": 99998,
--- a/packages/block-library/src/navigation/editor.scss
+++ b/packages/block-library/src/navigation/editor.scss
@@ -468,10 +468,6 @@ $color-control-label-height: 20px;
 	top: $admin-bar-height + $header-height + $block-toolbar-height + $border-width;
 }
 
-.is-sidebar-opened .wp-block-navigation__responsive-container.is-menu-open {
-	right: $sidebar-width;
-}
-
 // When fullscreen.
 .is-fullscreen-mode {
 	.wp-block-navigation__responsive-container.is-menu-open {
--- a/packages/edit-post/src/components/layout/index.js
+++ b/packages/edit-post/src/components/layout/index.js
@@ -20,7 +20,6 @@ import {
 	privateApis as blockEditorPrivateApis,
 	store as blockEditorStore,
 } from '@wordpress/block-editor';
-import { useViewportMatch } from '@wordpress/compose';
 import { PluginArea } from '@wordpress/plugins';
 import { __, sprintf } from '@wordpress/i18n';
 import { useCallback, useMemo } from '@wordpress/element';
@@ -144,19 +143,16 @@ function useEditorStyles() {
 function Layout( { initialPost } ) {
 	useCommands();
 	useEditPostCommands();
-	const isWideViewport = useViewportMatch( 'large' );
 	const paddingAppenderRef = usePaddingAppender();
 	const shouldIframe = useShouldIframe();
 	const { createErrorNotice } = useDispatch( noticesStore );
 	const {
 		mode,
 		isFullscreenActive,
-		sidebarIsOpened,
 		hasActiveMetaboxes,
 		hasBlockSelected,
 		showIconLabels,
 		isDistractionFree,
-		showBlockBreadcrumbs,
 		showMetaBoxes,
 		hasHistory,
 		isEditingTemplate,
@@ -175,7 +171,6 @@ function Layout( { initialPost } ) {
 				!! select( blockEditorStore ).getBlockSelectionStart(),
 			showIconLabels: get( 'core', 'showIconLabels' ),
 			isDistractionFree: get( 'core', 'distractionFree' ),
-			showBlockBreadcrumbs: get( 'core', 'showBlockBreadcrumbs' ),
 			showMetaBoxes:
 				select( editorStore ).getRenderingMode() === 'post-only',
 			hasHistory: !! getEditorSettings().onNavigateToPreviousEntityRecord,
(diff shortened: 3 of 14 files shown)


packages/base-styles/_z-index.scss
```diff
@@ -86,13 +85,12 @@ $z-layers: (
 	// Show sidebar above wp-admin navigation bar for mobile viewports:
 	// #wpadminbar { z-index: 99999 }
 	".interface-interface-skeleton__sidebar": 100000,
-	".edit-post-layout__toggle-sidebar-panel": 100000,
```
youknowriad, line 89: In my test this classname was not being used.

packages/block-library/src/navigation/editor.scss
```diff
@@ -468,10 +468,6 @@ $color-control-label-height: 20px;
 	top: $admin-bar-height + $header-height + $block-toolbar-height + $border-width;
 }
 
-.is-sidebar-opened .wp-block-navigation__responsive-container.is-menu-open {
```
youknowriad, line 471: this is-sidebar-opened classname was not being added in the site editor (where the navigation block is mostly used) which means that I just removed as this style is probably not useful.

packages/edit-post/src/components/layout/index.js
```diff
@@ -201,11 +196,7 @@ function Layout( { initialPost } ) {
 	}
 
 	const className = clsx( 'edit-post-layout', 'is-mode-' + mode, {
-		'is-sidebar-opened': sidebarIsOpened,
 		'has-metaboxes': hasActiveMetaboxes,
-		'is-distraction-free': isDistractionFree && isWideViewport,
-		'has-block-breadcrumbs':
```
youknowriad, line 207: This was not used.

packages/edit-post/src/components/layout/index.js
```diff
@@ -201,11 +196,7 @@ function Layout( { initialPost } ) {
 	}
 
 	const className = clsx( 'edit-post-layout', 'is-mode-' + mode, {
-		'is-sidebar-opened': sidebarIsOpened,
 		'has-metaboxes': hasActiveMetaboxes,
-		'is-distraction-free': isDistractionFree && isWideViewport,
```
youknowriad, line 206: I moved this class within the component.

packages/editor/src/components/editor-interface/index.js
```diff
@@ -206,9 +208,7 @@ export default function EditorInterface( {
 				isRichEditingEnabled &&
 				blockEditorMode !== 'zoom-out' &&
 				mode === 'visual' && (
-					<div className="edit-post-layout__footer">
```
youknowriad, line 209: Removed this div as it was useless in my tests.

packages/editor/src/components/editor-interface/style.scss
```diff
@@ -0,0 +1,3 @@
+.editor-editor-interface .entities-saved-states__panel-header {
```
youknowriad, line 1: Move this style from and applied a style that doesn't rely on external classnames.

packages/editor/src/components/header/style.scss
```diff
@@ -190,7 +190,7 @@
 	}
 }
 
-.is-distraction-free {
+.editor-editor-interface.is-distraction-free {
```
youknowriad, line 193: This doesn't change anything really but it makes things easier to understand. Relying on global utility classnames is not great because we don't really know who adds the class, the impact...

packages/editor/src/components/post-publish-panel/style.scss
```diff
@@ -194,40 +191,34 @@
 	}
 }
 
-.edit-post-layout,
-.edit-site-editor__interface-skeleton {
```
youknowriad, line 198: It was not clear to me why these wrappers were needed, so I just removed them and the publish panel sidebars look the same for me. Am I missing something @ntsekouras

packages/interface/src/components/interface-skeleton/style.scss
```diff
@@ -170,16 +163,6 @@ html.interface-interface-skeleton__html-container {
 	@include break-medium() {
 		display: flex;
 	}
-
-	.block-editor-block-breadcrumb {
```
youknowriad, line 174: In my test the breadcrumb looked the same with and without this so I think this was useless (and misplaced anyway)

packages/edit-site/src/components/editor/index.js
```diff
@@ -234,7 +234,7 @@ export default function Editor( { isLoading } ) {
 					<SiteEditorMoreMenu />
 					<EditorInterface
 						className={ clsx(
-							'edit-site-editor__interface-skeleton',
```
carolinan, line 237: Looks like WooCommerce uses this class?
https://wpdirectory.net/search/01HZF50T58QDM6WW7MP6WHQ7Y1

packages/editor/src/components/editor-interface/index.js
carolinan: What does "it was useless" mean? https://wpdirectory.net/search/01HZF570S4FPCYK4W9343A277D

packages/interface/src/components/interface-skeleton/style.scss
carolinan: Current usage: https://wpdirectory.net/search/01HZF59YSMNJB2CYSYYC08CK8Q

packages/edit-site/src/components/editor/index.js
youknowriad: They are actually applying the class them selves, so it should be fine 

https://github.com/woocommerce/woocommerce/blob/f7ef2ca91b79d17fe3ac821723a9b97e4cc680b5/plugins/woocommerce-admin/client/customize-store/assembler-hub/editor.tsx#L78

They probably don't need to apply it at all though, it seems useless looking at their code base.

packages/interface/src/components/interface-skeleton/style.scss
youknowriad: This class hasn't been removed. It's the style that was removed.

packages/editor/src/components/editor-interface/index.js
youknowriad: It was useless within Gutenberg.

Looking at the link above, it seems all of the plugins are actually rendering this div and targeting it themselves. They're basically copy/pasting Gutenberg code in their plugins I'd say. So this change won't have any impact on them.

The only one that can be slightly impacted I think is "Virtual Classroom – Video Conferencing & Online Meeting with BigBlueButton". They're hiding the div using CSS.

packages/edit-post/src/components/layout/index.js
juanfra: We need to keep the `has-block-breadcrumbs` class to avoid a regression for [this one](https://github.com/WordPress/gutenberg/issues/53198)

packages/edit-site/src/components/layout/index.js
```diff
@@ -185,10 +180,6 @@ export default function Layout() {
 						'is-full-canvas': canvasMode === 'edit',
 						'has-fixed-toolbar': hasFixedToolbar,
 						'is-block-toolbar-visible': hasBlockSelected,
-						'has-block-breadcrumbs':
```
juanfra, line 188: We need to keep the `has-block-breadcrumbs` class to avoid a regression for [this one](https://github.com/WordPress/gutenberg/issues/53198)

youknowriad: We've discussed recently with @jameskoster that the snackbars should remain in the exact same position regardless of what the UI is showing. It's already the case in trunk and this class is not used at all in our styles.

youknowriad:
Could use an approval here to be able to follow-up with related improvements #62274

jorgefilipecosta (approved):
The code changes look good I did some tests on the post editor, site editor, widgets editor, and customizer widget blocks and things look ok in all of them.

packages/edit-site/src/components/layout/index.js
juanfra: Thank you. Is there any place I can see the discussion around the topic?

youknowriad: yes, in this PR https://github.com/WordPress/gutenberg/pull/61676

packages/interface/src/components/interface-skeleton/style.scss
andrewserong: @youknowriad it looks like this removal has caused a regression in `trunk` in the post and site editors when no block is selected:

image

As far as I can tell the `display: flex`, `height` and `align-items` were necessary here, and the left padding helped keep the breadcrumbs from bumping up against the edge of the window.

I've put up a revert PR over in https://github.com/WordPress/gutenberg/pull/62309 for just this bit of styling to get it back to what it was. We could then look at moving the CSS elsewhere in a follow-up if there's a better place for it?

packages/edit-site/src/components/layout/index.js
juanfra: Thank you!

packages/editor/src/components/post-publish-panel/style.scss
ntsekouras: I don't remember exactly, but I think it was something about the `top` property for sure and was oriented towards bottom. I tested now and it seemed fine..

ellatrix:

packages/editor/src/components/post-publish-panel/style.scss
ivan-ottinger: Hey @youknowriad & @ntsekouras 👋🏼 Turns out there's a small regression. I have created a PR, but don't have write permissions yet: https://github.com/WordPress/gutenberg/pull/62736

Feel free to take a look. Thank you. 🙂

PR #62248, opened 2024-06-03
Editor: Refine availability of rename post action

stokesman:
## What?
Omits the "rename" action on posts without "title" support.

## Why?
At present a post without "title" support has the action available but using it causes the editor to crash.

## How?
Adds a condition to omit the action when the post type lacks title support.

## Testing Instructions
If you don’t have a custom post type without title support handy there’s an e2e test added in the first commit that fails without the second commit. So from this branch with dev script running:

1. `checkout HEAD~`
2. `npm run test:e2e -- custom-post-types`
3. Observe failure because "rename" action was present
4. `git switch -`
5. Repeat step 2
6. Observe success

Or manually:
1. Open a post of a type without "title" support
2. Open the post’s "Actions" menu
3. Observe "rename" is not present.

## Screenshots or screencast
Before this PR, the crash:

https://github.com/WordPress/gutenberg/assets/9000376/337ecab0-00b0-47d0-98c0-ab9ca8336ca0

### DIFF
--- a/test/e2e/specs/editor/plugins/custom-post-types.spec.js
+++ b/test/e2e/specs/editor/plugins/custom-post-types.spec.js
@@ -59,6 +59,42 @@ test.describe( 'Test Custom Post Types', () => {
 		await expect( parentPageLocator ).toHaveValue( parentPage );
 	} );
 
+	test( 'should not be able to rename a post that is either auto-drafted or lacks title support', async ( {
+		admin,
+		editor,
+		page,
+	} ) => {
+		await admin.createNewPost( { postType: 'hierar-no-title' } );
+		await editor.canvas
+			.locator( 'role=button[name="Add default block"i]' )
+			.click();
+		await page.keyboard.type( 'hi' );
+		await editor.openDocumentSettingsSidebar();
+		await page
+			.getByRole( 'region', { name: 'Editor settings' } )
+			.getByRole( 'tab', {
+				name: 'Hierarchical No Title',
+			} )
+			.click();
+		await page.getByRole( 'button', { name: 'Actions' } ).click();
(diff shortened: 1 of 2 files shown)


test/e2e/specs/editor/plugins/custom-post-types.spec.js
```diff
@@ -59,6 +59,42 @@ test.describe( 'Test Custom Post Types', () => {
 		await expect( parentPageLocator ).toHaveValue( parentPage );
 	} );
 
+	test( 'should not be able to rename a post that is either auto-drafted or lacks title support', async ( {
```
stokesman: Adding this e2e test was the first thing that occurred to me but now I wonder if it’d be better to create unit tests for [post-actions](https://github.com/WordPress/gutenberg/tree/trunk/packages/editor/src/components/post-actions) which currently lacks any.

ntsekouras (commented):
Thanks for the PR and the fix! This is something that should be included in 6.6.

I think it's a better approach to check if a post type supports `title` [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/components/post-actions/actions.js#L1051), similar to how we check for revisions support. The reason is that `supports` is about all items of a post type, whereas `isEligible` is more useful for `per item` checks.

With the above approach, I think we don't need to also add the `auto-draft` check in `isEligible` function, correct?

stokesman:
Thanks for the 🎯 advice Nik! Checking supports on the post type is indeed the indicated remedy. I was asking myself why `isEligible` didn’t have post type available to it but it makes sense now.

ntsekouras (approved):
Looks great, thank you!

ellatrix:

PR #62249, opened 2024-06-03
Use border instead of hr for filtered block list separator

jeryj:
Follow-up to https://github.com/WordPress/gutenberg/pull/62169 as part of https://github.com/WordPress/gutenberg/issues/60991

## What?

Switches from an hr for a border line to a border bottom in the css.

## Why?

Cleanup. Have a thinner line.

## Testing Instructions

- Select a block with a filtered list of allowed blocks (such as the list block)
- Open the inserter
- Check that the divider line between the allowed blocks section and all blocks is 1px wide

### DIFF
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -207,7 +207,6 @@ export function BlockTypesTab(
 							showMostUsedBlocks={ showMostUsedBlocks }
 							className="block-editor-inserter__insertable-blocks-at-selection"
 						/>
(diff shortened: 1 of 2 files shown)


draganescu (approved):
Testing this showed the border works better than the HR. Not sure if we should separate only the insertable at selection from the rest or if every panel should be separated (e.g. most used is not separated from the insertable at _next_ insertion point - which is a list of categories). But that is outside the scope of this PR.

jeryj:
Rebasing to see if all the checks will pass.

ellatrix:

PR #62255, opened 2024-06-04
`CustomSelectControlV2`: fix handling of extra attributes passed to `options` in the legacy adapter

fullofcaffeine:
Follow-up to/split from: https://github.com/WordPress/gutenberg/pull/61272.

## What?

- Fix the object type, making sure the interface remains the same as for V1, so that consumers receive the object with the expected property/values. Read on for more details.
- Add test to V1 to prove that it supports custom properties as part of the option and that it follows the same interface/behavior as V1.

## Why

Before the changes here, the adapter built the object like this:

```jsx
(...)
const changeObject = {
	highlightedIndex: state.renderedItems.findIndex(
		( item ) => item.value === nextValue
	),
	inputValue: '',
	isOpen: state.open,
	selectedItem: {
		name: nextValue as string,
		key: nextValue as string,
	},
	type: '',
};
onChange( changeObject );
(...)
```
_Source: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/custom-select-control-v2/legacy-component/index.tsx#L46-L47_

Notice the `selectedItem` property there. There are a few things that are wrong with it.

The classic component (V1) documents the `options` as an `Array<{ key: String, name: String, style: ?{}, className: ?String, ...rest }>` ([docs](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/custom-select-control/README.md#options)). Here we find the first (but less critical) smell in the adapter, the `LegacyType` type doesn't account for `...rest`:

```jsx
type LegacyOption = {
	key: string;
	name: string;
	style?: React.CSSProperties;
	className?: string;
	__experimentalHint?: string;
};
```
_Source: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/custom-select-control-v2/types.ts#L76._

It'd be easy to solve with an additional `[key: string]: any` type property there.

The way the V2 legacy adapter is [building](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/custom-select-control-v2/legacy-component/index.tsx#L45-L48) the `selectedItem` object is wrong:

*  it's hard-coding an object with only `key` and `name`. This is wrong as V1 can have an arbitrary number of properties for an item (see below). In addition to that, 
* it's also setting both to have the same value, with is also wrong - the key and name might be different (for example, a `selectedItem` from the font-appearance select looks like this when selected from V1: `'{"key":"normal-200","name":"Extra Light","style":{"fontStyle":"normal","fontWeight":"200"}}'` and from the adapter (without the fixes here): `'{"name":"Extra Light","key":"Extra Light"}'`.

## How?

The `options` prop is the array of options passed by the consumer. It already has the `LegacyProps`, which is the right sub-type the callback param expects. So, instead of rebuilding the structure, we just pass `options`:

```diff
packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -36,16 +36,15 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
                        await Promise.resolve();
                        const state = store.getState();

+                       const option = options.find( ( item ) => item.name === nextValue );
+
                        const changeObject = {
                                highlightedIndex: state.renderedItems.findIndex(
                                        ( item ) => item.value === nextValue
                                ),
                                inputValue: '',
                                isOpen: state.open,
-                               selectedItem: {
-                                       name: nextValue as string,
-                                       key: nextValue as string,
-                               },
+                               selectedItem: option!,
                                type: '',
                        };
                        onChange( changeObject );
```

This fixes the contract with consumers by making the object passed to the callback follow the same interface as V1 (with the right structure).

There's still a problem of the selected item not being set to the current value when the element is mounted. This was already done in the parent PR and will be split in a subsequent PR.

## Testing Instructions

- Tests should pass.
- Check in Storybook
- Check in the editor 

For a quick way to test in the editor, you can apply the follow pach

<details>
<summary>Click to show the diff</summary>

```diff
diff --git a/packages/block-editor/src/components/font-appearance-control/index.js b/packages/block-editor/src/components/font-appearance-control/index.js
index 18e814ad23..b7349ab4d6 100644
--- a/packages/block-editor/src/components/font-appearance-control/index.js
+++ b/packages/block-editor/src/components/font-appearance-control/index.js
@@ -1,10 +1,17 @@
 /**
  * WordPress dependencies
  */
-import { CustomSelectControl } from '@wordpress/components';
+import { CustomSelectControl, privateApis } from '@wordpress/components';
 import { useMemo } from '@wordpress/element';
 import { __, _x, sprintf } from '@wordpress/i18n';
 
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { CustomSelectControlV2LegacyAdapter } = unlock( privateApis );
+
 const FONT_STYLES = [
 	{
 		name: _x( 'Regular', 'font style' ),
@@ -122,6 +129,10 @@ export default function FontAppearanceControl( props ) {
 							fontStyle: styleValue,
 							fontWeight: weightValue,
 						},
+						className: 'yay-custom-class-name-yolo',
+						arbitrary: 'property',
+						foo: 'bar',
+						bar: 'foooo',
 					} );
 				}
 			);
@@ -138,6 +149,10 @@ export default function FontAppearanceControl( props ) {
 				key: value,
 				name,
 				style: { fontStyle: value, fontWeight: undefined },
+				className: 'yay-custom-class-name-yolo',
+				arbitrary: 'property',
+				foo: 'bar',
+				bar: 'foooo',
 			} );
 		} );
 		return combinedOptions;
@@ -151,6 +166,10 @@ export default function FontAppearanceControl( props ) {
 				key: value,
 				name,
 				style: { fontStyle: undefined, fontWeight: value },
+				className: 'yay-custom-class-name-yolo',
+				arbitrary: 'property',
+				foo: 'bar',
+				bar: 'foooo',
 			} );
 		} );
 		return combinedOptions;
@@ -203,19 +222,36 @@ export default function FontAppearanceControl( props ) {
 		);
 	};
 
+	console.log( 'New value', currentSelection );
+
 	return (
 		hasStylesOrWeights && (
-			<CustomSelectControl
-				{ ...otherProps }
-				className="components-font-appearance-control"
-				label={ label }
-				describedBy={ getDescribedBy() }
-				options={ selectOptions }
-				value={ currentSelection }
-				onChange={ ( { selectedItem } ) =>
-					onChange( selectedItem.style )
-				}
-			/>
+			<>
+				<CustomSelectControl
+					{ ...otherProps }
+					className="components-font-appearance-control"
+					label={ label }
+					describedBy={ getDescribedBy() }
+					options={ selectOptions }
+					value={ currentSelection }
+					onChange={ ( { selectedItem } ) => {
+						console.log( { version: 'v1', selectedItem } );
+						onChange( selectedItem.style );
+					} }
+				/>
+				<CustomSelectControlV2LegacyAdapter
+					{ ...otherProps }
+					className="components-font-appearance-control"
+					label={ label }
+					describedBy={ getDescribedBy() }
+					options={ selectOptions }
+					value={ currentSelection }
+					onChange={ ( { selectedItem } ) => {
+						console.log( { version: 'v2', selectedItem } );
+						onChange( selectedItem.style );
+					} }
+				/>
+			</>
 		)
 	);
 }
diff --git a/packages/block-editor/src/components/global-styles/typography-panel.js b/packages/block-editor/src/components/global-styles/typography-panel.js
index 3106723945..0411790a78 100644
--- a/packages/block-editor/src/components/global-styles/typography-panel.js
+++ b/packages/block-editor/src/components/global-styles/typography-panel.js
@@ -431,7 +431,6 @@ export default function TypographyPanel( {
 						hasFontStyles={ hasFontStyles }
 						hasFontWeights={ hasFontWeights }
 						size="__unstable-large"
-						__nextHasNoMarginBottom
 					/>
 				</ToolsPanelItem>
 			) }

```

</details>

And then in the Site Editor, open the Global Styles sidebar, click the "Typography" button,  and then click any entry under the "Elements" group. In the new view that loads in the sidebar, notice how there are two identical "Appearance" dropdowns — the first one is implemented with the v1 component, the second one is implemented with the v2 legacy adapter.

### DIFF
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -36,16 +36,15 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 			await Promise.resolve();
 			const state = store.getState();
 
+			const option = options.find( ( item ) => item.name === nextValue );
+
 			const changeObject = {
 				highlightedIndex: state.renderedItems.findIndex(
 					( item ) => item.value === nextValue
 				),
 				inputValue: '',
 				isOpen: state.open,
-				selectedItem: {
-					name: nextValue as string,
-					key: nextValue as string,
-				},
+				selectedItem: option!,
 				type: '',
 			};
 			onChange( changeObject );
@@ -121,4 +120,15 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	);
 }
 
-export default CustomSelectControl;
+export function ClassicCustomSelectControlV2Adapter(
+	props: LegacyCustomSelectProps
+) {
+	return (
+		<CustomSelectControl
+			__experimentalShowSelectedHint={ false }
+			{ ...props }
+		/>
+	);
+}
+
+export default ClassicCustomSelectControlV2Adapter;
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -52,7 +52,7 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	} );
 
 	const children = options.map(
-		( { name, key, __experimentalHint, ...rest } ) => {
+		( { name, key, __experimentalHint, style, className } ) => {
 			const withHint = (
 				<Styled.WithHintWrapper>
 					<span>{ name }</span>
@@ -67,7 +67,8 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 					key={ key }
 					value={ name }
 					children={ __experimentalHint ? withHint : name }
-					{ ...rest }
+					style={ style }
+					className={ className }
 				/>
 			);
 		}
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -27,7 +27,11 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	// Forward props + store from v2 implementation
 	const store = Ariakit.useSelectStore( {
 		async setValue( nextValue ) {
-			if ( ! onChange ) {
+			const nextOption = options.find(
+				( item ) => item.name === nextValue
+			);
+
+			if ( ! onChange || ! nextOption ) {
 				return;
 			}
 
@@ -36,15 +40,13 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 			await Promise.resolve();
 			const state = store.getState();
 
-			const option = options.find( ( item ) => item.name === nextValue );
-
 			const changeObject = {
 				highlightedIndex: state.renderedItems.findIndex(
 					( item ) => item.value === nextValue
 				),
 				inputValue: '',
 				isOpen: state.open,
-				selectedItem: option!,
+				selectedItem: nextOption,
 				type: '',
 			};
 			onChange( changeObject );
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -14,7 +14,7 @@ import * as Styled from '../styles';
 
 function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	const {
-		__experimentalShowSelectedHint,
+		__experimentalShowSelectedHint = false,
 		__next40pxDefaultSize = false,
 		describedBy,
 		options,
@@ -123,15 +123,4 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	);
 }
 
-export function ClassicCustomSelectControlV2Adapter(
-	props: LegacyCustomSelectProps
-) {
-	return (
-		<CustomSelectControl
-			__experimentalShowSelectedHint={ false }
-			{ ...props }
-		/>
-	);
-}
-
-export default ClassicCustomSelectControlV2Adapter;
+export default CustomSelectControl;
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -574,22 +574,4 @@ describe.each( [
 			} )
 		);
 	} );
-
-	it( 'Should display the initial value passed as the selected value', async () => {
-		const initialSelectedItem = legacyProps.options[ 5 ];
-
-		const testProps = {
-			...legacyProps,
-			value: initialSelectedItem,
-		};
-
-		render( <Component { ...testProps } /> );
-
-		const currentSelectedItem = await screen.findByRole( 'combobox', {
-			expanded: false,
-		} );
-
-		// Verify the initial selected value
-		expect( currentSelectedItem ).toHaveTextContent( 'aquarela' );
-	} );
 } );
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -46,6 +46,16 @@ const legacyProps = {
 			name: 'aquamarine',
 			style: customStyles,
 		},
+		{
+			key: 'color3',
+			name: 'tomato (with custom props)',
+			className: customClassName,
+			style: customStyles,
+			// try passing a valid HTML attribute
+			'aria-label': 'test label',
+			// try adding a custom prop
+			customPropFoo: 'foo',
+		},
 	],
 };
 
@@ -289,8 +299,7 @@ describe.each( [
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
-				// TODO: key should be different — this is a known bug and will be fixed
-				selectedItem: { key: 'violets', name: 'violets' },
+				selectedItem: { key: 'flower1', name: 'violets' },
 				type: '',
 			} )
 		);
@@ -333,8 +342,7 @@ describe.each( [
 			1,
 			expect.objectContaining( {
 				selectedItem: expect.objectContaining( {
-					// TODO: key should be different — this is a known bug and will be fixed
-					key: 'violets',
+					key: 'flower1',
 					name: 'violets',
 				} ),
 			} )
@@ -347,9 +355,51 @@ describe.each( [
 			2,
 			expect.objectContaining( {
 				selectedItem: expect.objectContaining( {
-					// TODO: key should be different — this is a known bug and will be fixed
-					key: 'poppy',
+					key: 'flower3',
 					name: 'poppy',
+				} ),
+			} )
+		);
+	} );
+
+	it( 'Should accept and pass arbitrary properties to the selectedItem object in the onChange callback, but without applying them to the DOM elements apart from style and classname', async () => {
+		const onChangeMock = jest.fn();
+
+		render( <Component { ...legacyProps } onChange={ onChangeMock } /> );
+
+		const currentSelectedItem = screen.getByRole( 'combobox', {
+			expanded: false,
+		} );
+
+		await click( currentSelectedItem );
+
+		const optionWithCustomAttributes = screen.getByRole( 'option', {
+			name: 'tomato (with custom props)',
+		} );
+
+		// Assert that the option element does not have the custom attributes
+		expect( optionWithCustomAttributes ).not.toHaveAttribute(
+			'customPropFoo'
+		);
+		expect( optionWithCustomAttributes ).not.toHaveAttribute(
+			'aria-label'
+		);
+
+		await click( optionWithCustomAttributes );
+
+		// NOTE: legacy CustomSelectControl doesn't fire onChange
+		// on first render, and so at this point in time `onChangeMock`
+		// would be called only once.
+		expect( onChangeMock ).toHaveBeenCalledTimes( 2 );
+		expect( onChangeMock ).toHaveBeenCalledWith(
+			expect.objectContaining( {
+				selectedItem: expect.objectContaining( {
+					key: 'color3',
+					name: 'tomato (with custom props)',
+					className: customClassName,
+					style: customStyles,
+					'aria-label': 'test label',
+					customPropFoo: 'foo',
 				} ),
 			} )
 		);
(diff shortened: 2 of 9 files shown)


fullofcaffeine:
It looks like we'll need to support arbitrary properties (`any`) in `LegacyOptions`.  I see there's at least one instance of the `CustomSelectControl` that tries to access a property that's not one of the pre-defined/pre-typed ones in `Array<{ key: String, name: String, style: ?{}, className: ?String, ...rest }>`: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/hooks/position.js#L45. 

I'm not sure how this would work for V2 - e.g., if V2 actually supports custom properties in an option like this. However, that's not the main focus of this PR, and we can discuss this in parallel. From what I see, for V2, consumers would use this API from Ariakit -> https://ariakit.org/reference/select-item#getitem?

ciampo:
I will take over the work on this PR. I will first take a look at #61272 to get more context, then leave a review of the current status of this PR, and then start work on it

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -36,24 +36,23 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
...
 				highlightedIndex: state.renderedItems.findIndex(
 					( item ) => item.value === nextValue
 				),
 				inputValue: '',
 				isOpen: state.open,
-				selectedItem: {
-					name: nextValue as string,
-					key: nextValue as string,
-				},
+				selectedItem: option!,
```
ciampo: Instead of forcing `option` to be non-nullable (which may cause unintended bugs in the future), we may want to add a type guard / early return. I'll see if that is a possibility here

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -68,7 +67,8 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 					key={ key }
 					value={ name }
 					children={ __experimentalHint ? withHint : name }
-					{ ...rest }
+					style={ style }
+					className={ className }
```
ciampo, line 73: I will look into the forwarding of `...rest` and what issues that was causing, and I will double-check what the legacy component's behaviour was

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -121,4 +121,15 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	);
 }
 
-export default CustomSelectControl;
+export function ClassicCustomSelectControlV2Adapter(
+	props: LegacyCustomSelectProps
+) {
+	return (
+		<CustomSelectControl
+			__experimentalShowSelectedHint={ false }
```
ciampo: Not sure about the purpose of setting `__experimentalShowSelectedHint` to `false` in here, especially since it can be overridden by `props` — can we just set the default value to `false` when deconstructing props in the component's implementation?

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -53,7 +62,9 @@ const ControlledCustomSelectControl = ( {
 	onChange,
 	...restProps
 }: React.ComponentProps< typeof UncontrolledCustomSelectControl > ) => {
+	const { value: overrideValue } = restProps;
 	const [ value, setValue ] = useState( options[ 0 ] );
+	const initialValue = overrideValue ?? value;
```
ciampo: Could we achieve the same behavior by passing `props.value` to the `useState` hook ? E.g.

```
const [ value, setValue ] = useState( restProps.value ?? options[ 0 ] );
```

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -302,6 +313,7 @@ describe.each( [
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
+
```
ciampo: nit: extra space

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -328,7 +340,7 @@ describe.each( [
 		await type( 'p' );
 		await press.Enter();
 
-		expect( mockOnChange ).toHaveReturnedWith( 'poppy' );
+		expect( mockOnChange ).toHaveReturnedWith( 'flower1' );
```
ciampo, line 343: I need to look more into why this assertion has changed

packages/components/src/custom-select-control/test/index.js
```diff
@@ -48,13 +48,23 @@ const props = {
...
+const ControlledCustomSelectControl = ( {
+	options,
+	onChange: onChangeMock,
+	...restProps
+} ) => {
 	const [ value, setValue ] = useState( options[ 0 ] );
+	const onChange = ( changeObject ) => {
+		if ( onChangeMock ) {
+			onChangeMock( changeObject );
+		}
```
ciampo: Nit: prefer optional chaining

packages/components/src/custom-select-control/test/index.js
```diff
@@ -238,28 +248,56 @@ describe.each( [
 		).toHaveTextContent( 'Hint' );
 	} );
 
-	it( 'shows selected hint in list of options when added, regardless of __experimentalShowSelectedHint prop', async () => {
+	it( 'Should accept and pass arbitrary properties to the selectedItem object in the onChange callback', async () => {
```
ciampo: It looks like we changed an existing test. Should we keep the old test and add a new one instead?

packages/components/src/custom-select-control/test/index.js
```diff
@@ -238,28 +248,56 @@ describe.each( [
...
+		} );
+
+		await user.click( currentSelectedItem );
 		await user.click(
-			screen.getByRole( 'button', { name: 'Custom select' } )
+			screen.getByRole( 'option', { name: 'Custom Option' } )
 		);
 
-		expect( screen.getByRole( 'option', { name: /hint/i } ) ).toBeVisible();
+		expect( onChangeMock ).toHaveBeenCalledTimes( 1 );
```
ciampo, line 385: We should add the same assertions added for v2 checking that `customProp1` and `customProp2` are not added as props/attributes of the item

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -286,7 +297,7 @@ describe.each( [
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
-				selectedItem: { key: 'violets', name: 'violets' },
+				selectedItem: { key: 'flower1', name: 'violets' },
```
ciampo, line 302: Need to double-check if this behaviour matches v1

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: I can confirm that [the v1 `CustomSelectControl` only forwards `style` and `className` ](https://github.com/WordPress/gutenberg/blob/b76de86ffe89b437808a7a56c48021838fcb2d8a/packages/components/src/custom-select-control/index.js#L197-L209), and therefore this change seems correct.

I will make sure unit tests check this behaviour both for v1 and v2

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
ciampo: v1 doesn't behave this way, because it doesn't fire `onChange` on first render. This is something that we should investigate separately from this PR, though

packages/components/src/index.ts
```diff
@@ -64,6 +64,7 @@ export {
 } from './composite';
 export { ConfirmDialog as __experimentalConfirmDialog } from './confirm-dialog';
 export { StableCustomSelectControl as CustomSelectControl } from './custom-select-control';
+export { default as CustomSelectControlV2 } from './custom-select-control-v2';
```
ciampo: Let's make sure we're good to export the component before committing this !

packages/components/src/custom-select-control/test/index.js
ciampo: Added same test to v1

packages/components/src/custom-select-control/test/index.js
ciampo: Added back the original test

packages/components/src/custom-select-control/test/index.js
ciampo: Swapped to optional chaining and actually uniformed the code between v1 and legacy v2

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: Switched to setting default `false` value during prop destructuring

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: done

ciampo:
Update:

- I verified that v1 allows for additional attributes in the options. Those additional properties are passed to the argument of the `onChange` callback, but only `style` and `className` are applied to the actual DOM element representing the option — which was already the assumption made by Marcelo
- I added unit tests to both v1 and legacy v2 components, and I'll continue working towards "unit test parity" as much as possible, while keeping in mind that there are some known differences — for example:
  - v1 renders a `button` while legacy v2 renders a `combobox` (which we see as an improvement)
  - v1 doesn't fire `onChange` on first render, while legacy v2 does (we should investigate if we want to mimic the same behaviour in legacy v2)
  - v1 (via `downshift`) passes a `type` attribute as part of the `onChange` callback argument. Legacy v2 passes the `type` argument, but leaves it as an empty string (we may want to investigate further if this will cause any disruption)
  - there are also differences in how initial state is computed / set — Marcelo started exploration in https://github.com/WordPress/gutenberg/pull/62308, which will likely be my next area of focus after landing this PR
  - v1 stops keyboard event propagations on the listbox (context), legacy v2 doesn't for now

Next steps:
- finish taking a look at unit tests to make sure that both v1 and legacy v2 are tested against the same assertions as much as possible (#62706)
- test on Storybook / editor
- prepare the PR for next round of review
- finalise list of follow-ups and update tracking issue

ciampo:
Most unit test changes were moved to https://github.com/WordPress/gutenberg/pull/62706 and merged.

I rebased, added a few more tweaks, and updated the [tracking issue](https://github.com/WordPress/gutenberg/issues/55023) with the additional follow-up tasks that emerged while working on this PR.

**This PR is ready for a new round of review** — hopefully we're able to merge it soon enough. I also updated the PR description and added some extra testing instructions for testing in the editor.

packages/components/src/custom-select-control/stories/index.story.tsx
ciampo, line 1: Changes in this file are simply to allow the v1 Storybook to log `onChange` arguments in the "Actions" tab, thus making it easier to debug the changes in Storybook

ciampo:
Follow-up [PR fixing value setting](https://github.com/WordPress/gutenberg/pull/62733) is up and will be ready for final review as soon as this PR is merged

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -347,14 +355,56 @@ describe.each( [
...
-					// TODO: key should be different — this is a known bug and will be fixed
-					key: 'poppy',
+					key: 'flower3',
 					name: 'poppy',
 				} ),
 			} )
 		);
 	} );
 
+	it( 'Should accept and pass arbitrary properties to the selectedItem object in the onChange callback, but without applying them to the DOM elements apart from style and classname', async () => {
```
tyxla: How can we make this shorter 😅 Perhaps something like:

```suggestion
	it( "Should pass arbitrary props to onChange's selectedItem, but apply only style and className to DOM elements", async () => {
```

packages/components/src/custom-select-control/test/index.js
```diff
@@ -346,6 +356,47 @@ describe.each( [
 		);
 	} );
 
+	it( 'Should accept and pass arbitrary properties to the selectedItem object in the onChange callback, but without applying them to the DOM elements apart from style and classname', async () => {
```
tyxla: Same as above:

```suggestion
	it( 'Should pass arbitrary props to onChange's selectedItem, but apply only style and className to DOM elements', async () => {
```

packages/components/src/private-apis.ts
```diff
@@ -40,6 +41,7 @@ lock( privateApis, {
 	CompositeRowV2,
 	useCompositeStoreV2,
 	CustomSelectControl,
+	CustomSelectControlV2LegacyAdapter,
```
tyxla: Should we export it as a private API when we actually need it?

tyxla (approved):
Looking good and working well, thanks 🙌

PR #61717, opened 2024-05-16
Global Styles: Add block-level Text Alignment UI

t-hamano:
Closes #60762
Follow up #59531

## What?

This PR adds block-level Text Alignment UI in the Global Styles.
## Why?

#59531 added textAlign as block support. However, I think it would be useful to be able to change the default alignment in the GlobalStyles UI.

## How?

This UI only appears in the block-level typography panel. In other words, it will not appear on Typography > Text, Links, Headings, Captions and Buttons.

## Testing Instructions

Currently, no core blocks exist that have this support. Add support to Media & Text blocks to test this PR.

<details><summary>Details</summary>

```diff
diff --git a/packages/block-library/src/media-text/block.json b/packages/block-library/src/media-text/block.json
index 0f3519acb5..782da2e1eb 100644
--- a/packages/block-library/src/media-text/block.json
+++ b/packages/block-library/src/media-text/block.json
@@ -117,6 +117,7 @@
                        "padding": true
                },
                "typography": {
+                       "textAlign": true,
                        "fontSize": true,
                        "lineHeight": true,
                        "__experimentalFontFamily": true,
```

</details> 

Enable Emptytheme and define it in theme.json as follows. This enables textAlign support and changes the default textAlign of the Media & Text block.

> [!NOTE]
> #61182 will change the default for this support to `true`.

<details><summary>Details</summary>

```json
{
	"$schema": "../../schemas/json/theme.json",
	"version": 3,
	"settings": {
		"appearanceTools": true,
		"layout": {
			"contentSize": "840px",
			"wideSize": "1100px"
		},
		"typography": {
			"textAlign": true
		}
	},
	"styles": {
		"blocks": {
			"core/media-text": {
				"typography": {
					"textAlign": "center"
				}
			}
		}
	}
}
```

</details> 

- When you insert a Media & Text block, the text is centered by default.
- Go to the Global Styles option> blocks > Media & Text > Typograhy panel.
- The text alignment UI shows center. Changes to this button should be reflected in the block.
- Change the text alignment from the block toolbar of the inserted Media & Text block. It's supposed to override the default value of the global style.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/c71c935c-3e22-46ae-a5b1-c09abf4861a5

Issue #60763, opened 2024-04-16
Block Library: Migrate to the new textAlign block support


t-hamano:
> [!NOTE]
> ~There is currently an issue with not being able to override text alignment support at the block instance level, which is being addressed in #62260.~
> ~Please note that in order to address this issue, #62260 must be resolved first.~

Update: Now that #62260 has been merged, so we can tackle this issue.

## What problem does this address?

Over 20 core blocks have attributes and UI implemented for each block to control the `text-align` style. For example, it looks like the following.

<details><summary>block.json</summary>

```json
{
	"$schema": "https://schemas.wp.org/trunk/block.json",
	"apiVersion": 3,
	"name": "core/heading",
	"title": "Heading",
	"attributes": {
		"textAlign": {
			"type": "string"
		}
	}
}
```

</details> 

<details><summary>edit.js</summary>

```javascript
import {
	AlignmentControl,
	BlockControls,
	useBlockProps,
} from '@wordpress/block-editor';

function Edit( { attributes, setAttributes } ) {
	const { textAlign } = attributes;
	return (
		<>
			<BlockControls group="block">
				<AlignmentControl
					value={ textAlign }
					onChange={ ( nextAlign ) => {
						setAttributes( { textAlign: nextAlign } );
					} }
				/>
			</BlockControls>
			<div { ...useBlockProps() } />
		</>
	);
}
```

</details> 

These are all almost the same implementations and feel redundant.

## What is your proposed solution?

In #59531, `textAlign` is added as block support. I believe that we can gradually migrate to this block support.

Below is a list of blocks that provide attributes and UI to control text-aligin.

- [x]  core/button #73732
- [x]  core/comment-author-name #74068
- [x]  core/comment-content #74269
- [x]  core/comment-edit-link #74720
- [x]  core/comment-reply-link #74760
- [x]  core/comment-title #74945
- [x]  core/heading #74383
- [x]  core/paragraph: Note that the attribute is named "align" and not "textAlign". https://github.com/WordPress/gutenberg/pull/73111
- [x]  core/post-author: This block has been deprecated, and no action is required. #55352
- [x]  core/post-author-biography #74997
- [x]  core/post-author-name #75109
- [x]  core/post-comments-count #75321
- [x]  core/post-comments-form #75322
- [x]  core/post-comments-link #75332
- [x]  core/post-navigation-link #75557
- [x]  core/post-time-to-read #75541
- [x]  core/post-terms #75545
[...]

PR #59531, opened 2024-03-03
Try: Add new `textAlign` block support

t-hamano:
Fixes #38427
Related to #8450, #48202

## What?

This PR attempts to add a new `typography.textAlign` (left/center/right/~~justify~~) block support. I only have a basic implementation at the moment, but I'd love to hear your thoughts on whether this approach makes sense!

## Why?

The current text-align setting is implemented ad hoc in each block. In other words, the code looks like this:

```javascript
function Edit( { attributes, setAttributes } ) {
	const { textAlign } = attributes;
	const blockProps = useBlockProps( {
		className: classnames( {
			[ `has-text-align-${ textAlign }` ]: textAlign,
		} ),
	} );

	return (
		<>
			<BlockControls group="block">
				<AlignmentControl
					value={ textAlign }
					onChange={ ( nextAlign ) => {
						setAttributes( { textAlign: nextAlign } );
					} }
				/>
			</BlockControls>
			<div { ...blockProps } />
		</>
	);
}
```

Code like this exists in over 20 core blocks, so I think it makes sense to unify the block support.

I also think there are two benefits for developers:

- Ability to filter core block settings and limit alignment options
- Easily introduce alignment settings to custom blocks

## How?

The implementation itself is very similar to `align` support, with the following characteristics:

- `boolean | array` can be defined as a value.
- If the value is `true`, all `left/center/right/~~justify~~` options can be controlled.
- If the value is an `array`, only the options that match the array elements can be controlled.
- The attribute is added automatically, but the block can also have its own attribute definition.
- If the block edit mode is other than `default`, the toolbar control will not be available.

~~Additionally, if we replace this support in the core blocks in the future, the markup saved will remain the same, so there should be no need to add a deprecation.~~

~~Update: For all blocks, if text-align changes, the class name `has-text-align-{align}` is added. This PR, on the other hand, is output as an inline style. Therefore, we will need to add deprecations in every block.~~

However, I think we should note the following points:

- ~The only exception is the Paragraph block, which is defined as an attribute called `align` instead of `textAlign`. Therefore, it will be necessary to add deprecation only to this block, and adjust the alignment inheritance when transforming from/to the heading block.~
- The order of the toolbar icons on the block toolbar may be changed. For example, a Heading block has "Change level" button and "Align text" in the same "block" group. If the alignment setting was replaced with block support, "Align text", "Change level" would be rendered in that order.
[...]

PR #62260, opened 2024-06-04
Increase CSS specificity of text alignment classes

### LINKED ISSUE #62685: Theme.json > Styles > Typography > textAlign breaks text alignment options in the block toolbar
### Description

I searched but was not able to find an open issue for this.
I was trying to understand how the new textAlignment typography style and settings worked.

With Gutenberg trunk on 6.6 beta 3, if I set a default alignment to a text based block like the site title using theme.json, then I can not change the alignment using the block toolbar.

The CSS that sets the default seems to have a higher specificity and overrides the `has-text-align` CSS:

```
:root :where(.wp-block-site-title) {
    text-align: center;
}
.has-text-align-right {
    text-align: right; 
}
```

Activate Gutenberg.
Activate Twenty Twenty-Four.
Open theme.json replace the styles section with
```
	"styles": {
		"blocks": {
			"core/site-title": {
				"typography": {
					"textAlign": "center"
				}
			}
		}
	},
```
Open the Site Editor and insert a new Site Title block.
Confirm that the text in the block is centered.
Try adjusting the alignment to left or right using the option in the block toolbar.
Confirm that the alignment does not change.

_No response_

### Environment info

_No response_

Yes

Yes

### LINKED ISSUE #71313: user-overrides for P block not winning specificity over theme.json block style
### Description

As I understand it, specificity in block theme.json world is that it's core > global styles > block styles > user overrides (aka database) in the specificity war

See issue discovered while suggesting a UI enhancement : https://github.com/WordPress/gutenberg/issues/71312

Now posting as an issue since I can't change the tag on the other ticket

- define the block style for paragraphs as center in theme.json
- override a paragraph to left-align on any page or post

also detailed in enhancement outline since the two things are co-related issues

Image

Image

### Environment info

_No response_

- [x] Yes

- [x] Yes

### Please confirm which theme type you used for testing.

- [x] Block
- [ ] Classic
- [ ] Hybrid (e.g. classic with theme.json)
- [ ] Not sure

t-hamano:
Follow up https://github.com/WordPress/gutenberg/pull/59531
See this discussion: https://github.com/WordPress/gutenberg/pull/61717

## What?

This PR increases the CSS specificity of the `.has-text-align-*` selector so that text alignment support can be overridden by block instances.

## Why?

When text alignment styles are defined in the global styles or theme.json, the latest Gutenberg and WP6.6 generate styles like the following:

```css
:root :where(.wp-block-media-text) {
    text-align: center;
}
```

On the other hand, if you change text alignment on a block instance, the `.has-text-align-*` class is added. This style is part of the block library styles:

```css
.has-text-align-right {
    text-align: right;
}
```

Both selectors have the same specificity (`0-1-0`), and global styles are loaded after block library styles, so overriding styles on block instances doesn't work.

## How?

There are several possible approaches, but I added `:root` to increase the specificity of the selector by one (`0-2-0`). The reason I used `:root` is to make the format match the selector output by the global styles.

We could have added `!important`, but I didn't add it to avoid overriding corner cases like the following:

```css
// A theme developer might want to override text-alignment under certain conditions
.wp-block-media-text.is-style-test.has-text-align-center {
	text-align: left;
}
```

I also considered two other possible approaches:

### Enqueue text alignment-related selectors after the global styles

As far as I can tell, the only core inline CSS that is loaded after the global styles inline CSS is the style with the `core-block-supports` handle.

I could have moved the text alignment-related selectors there, but this style seems to be for generating dynamic selectors depending on the context, so I didn't think it was appropriate:

```
<style id='core-block-supports-inline-css'>
/**
* Core styles: block-supports
*/
.wp-container-core-group-is-layout-1 > * {
margin-block-start: 0;
margin-block-end: 0;
}
</style>
```

### Add a new selector for block support (such as `has-block-text-align-*`)

If we take this approach, we will convert the core text alignment controls to this new class when migrating to block support (See #60763).

However, some themes define additional styles that rely on the `.has-text-align-*` class, so migrating to this new class will break their styles.

Search results in WordPress core: https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20.has-text-align-&type=code

## Testing Instructions

Add textAlign support to the Media & Text block:

<details><summary>Details</summary>

```diff
diff --git a/packages/block-library/src/media-text/block.json b/packages/block-library/src/media-text/block.json
index 0f3519acb5..782da2e1eb 100644
--- a/packages/block-library/src/media-text/block.json
+++ b/packages/block-library/src/media-text/block.json
@@ -117,6 +117,7 @@
                        "padding": true
                },
                "typography": {
+                       "textAlign": true,
                        "fontSize": true,
                        "lineHeight": true,
                        "__experimentalFontFamily": true,
```

</details> 

Enable Emptytheme and change the default text alignment of the Media & Text block via theme.json:

<details><summary>Details</summary>

```json
{
	"version": 3,
	"settings": {
		"appearanceTools": true,
		"layout": {
			"contentSize": "840px",
			"wideSize": "1100px"
		}
	},
	"styles": {
		"blocks": {
			"core/media-text": {
				"typography": {
					"textAlign": "center"
				}
			}
		}
	}
}
```

</details> 

- Add a Media & Text block.
- The text should be centered by default.
- From the block toolbar, right-justify the text.
- Overrides at the block instance level should now work in both the editor and the frontend.

## Screenshots or screencast 

| Before | After |
|--------|--------|
| image| image |


ellatrix:
> Both selectors have the same specificity (0-1-0), and global styles are loaded after block library styles, so overriding styles on block instances doesn't work.

It sounds to me like block supports styles should load after global styles?

t-hamano:
> It sounds to me like block supports styles should load after global styles?

I have tried adding text alignment-related styles to `core-block-supports-inline-css` which are enqueued after the global styles via the style engine, as shown below.

I think this approach will probably work too?

```diff
diff --git a/lib/block-supports/typography.php b/lib/block-supports/typography.php
index fa2a7b81e9..e90291a74b 100644
--- a/lib/block-supports/typography.php
+++ b/lib/block-supports/typography.php
@@ -621,3 +621,20 @@ if ( function_exists( 'wp_render_typography_support' ) ) {
        remove_filter( 'render_block', 'wp_render_typography_support' );
 }
 add_filter( 'render_block', 'gutenberg_render_typography_support', 10, 2 );
+
+function gutenberg_poutput_text_align_styles() {
+       foreach ( array( 'left', 'center', 'right' ) as $alignment ) {
+               gutenberg_style_engine_get_styles(
+                       array(
+                               'typography' => array(
+                                       'textAlign' => $alignment,
+                               ),
+                       ),
+                       array(
+                               'selector' => ".has-text-align-{$alignment}",
+                               'context'  => 'block-supports',
+                       )
+               );
+       }
+}
+add_action( 'init', 'gutenberg_poutput_text_align_styles' );
diff --git a/packages/style-engine/class-wp-style-engine.php b/packages/style-engine/class-wp-style-engine.php
index 272c12705b..e519fbaf56 100644
--- a/packages/style-engine/class-wp-style-engine.php
+++ b/packages/style-engine/class-wp-style-engine.php
@@ -267,6 +267,12 @@ if ( ! class_exists( 'WP_Style_Engine' ) ) {
                                        ),
                                        'path'          => array( 'typography', 'lineHeight' ),
                                ),
+                               'textAlign'    => array(
+                                       'property_keys' => array(
+                                               'default' => 'text-align',
+                                       ),
+                                       'path'          => array( 'typography', 'textAlign' ),
+                               ),
                                'textColumns'    => array(
                                        'property_keys' => array(
                                                'default' => 'column-count',
```

</details>

core-block-supports

packages/block-library/src/common.scss
```diff
@@ -32,16 +32,17 @@
 }
 
 // Text alignments.
-.has-text-align-center {
+// Increase specificity so that styles applied via the global styles can be overridden in block instances.
+:root .has-text-align-center {
```
ellatrix, line 41: Do we not need the same for `.has-larger-font-size` etc. above?

ellatrix: Honestly, it sounds to me like these types of selectors should move _after_ theme.json styles. These are not default styles to me.

t-hamano:
In [3b093ca](https://github.com/WordPress/gutenberg/pull/62260/commits/3b093caee0b2c96f1365cb0889c51b38402fac33), I tried injecting text alignment-related styles _after_ the global styles. For example, if `text-align: right;` is applied to the Media & Text block via theme.json or global styles, the following styles are output:

```html
<style id='global-styles-inline-css'>
:root :where(.wp-block-media-text){ text-align: right; }
</style>
<style id='core-block-supports-inline-css'>
/**
 * Core styles: block-supports
 */
.has-text-align-left {
	text-align: left;
}
.has-text-align-center {
	text-align: center;
}
.has-text-align-right {
	text-align: right;
}
</style>
```

However, I couldn't figure out how to apply a similar approach on the editor side. If you have any implementation advice, I'd love to hear it 🙏

ellatrix:
While looking into this, I saw that no block currently supports text align? Is that right?

t-hamano:
> While looking into this, I saw that no block currently supports text align? Is that right?

Yes, that's right. You can find more details on what's coming in WP6.6 in [the draft dev note ](https://github.com/WordPress/gutenberg/pull/59531#issuecomment-2182754264)on this support.

ellatrix:
Is it a high priority for 6.6 to fix text alignment for the media and text block, if it's not supported by core?

t-hamano:
This issue isn't an issue unless developers explicitly opt-in to `textAlign` support for their custom blocks, or add `textAlign` support by filtering core block support.

So, while it would be ideal to fix this in WP6.6, I don't think it's a high priority.

ellatrix:
@t-hamano Are you ok with punting this? Sorry 😔

afercia:
I have a question.
I see Text alignment block support is mentioned in [the WordPress 6.6 Field guide](https://make.wordpress.org/core/2024/06/25/wordpress-6-6-field-guide/) as one of the new features that land in the release and it's detailed in [the Miscellaneous Editor changes dev note](https://make.wordpress.org/core/2024/06/24/miscellaneous-editor-changes-in-wordpress-6-6/#Added-Text-alignment-block-support).

However, I don't see any mention to the fact that this feature is, in a way, still in the works and when developers will play with it they will be very likely surprised the block toolbar alignment setting won't work as they expect.

I'd tend to think it would be important to mention this issue in the dev note. Pinging @juanmaguitar as author of the dev note.

t-hamano:
> However, I don't see any mention to the fact that this feature is, in a way, still in the works and when developers will play with it they will be very likely surprised the block toolbar alignment setting won't work as they expect.

That's certainly true. I'll add it to [this dev note](https://github.com/WordPress/gutenberg/pull/59531#issuecomment-2182754264) and ping @juanmaguitar.

juanmaguitar:
@t-hamano @afercia I have added a warning in the [Added Text alignment block support](https://make.wordpress.org/core/2024/06/24/miscellaneous-editor-changes-in-wordpress-6-6/#Added-Text-alignment-block-support) section of the [Miscellaneous Editor changes in WordPress 6.6](https://make.wordpress.org/core/2024/06/24/miscellaneous-editor-changes-in-wordpress-6-6/) DevNote, as per [this comment](https://github.com/WordPress/gutenberg/pull/59531#issuecomment-2196603971) 
Thanks for the heads up!

ellatrix:
This is only an issue for the media and text block I believe?

t-hamano:
> This is only an issue for the media and text block I believe?

This issue will likely occur with any block that supports `textAlign`.

For example, if the Heading block supported it, the same issue would occur.

heading

Just to be clear, the Media & Text block was mentioned only to test this PR, and currently no core blocks support `textAlign` by default.

carolinan:
I don't have much time to follow up today, but can add that I did not test with a block that supports text align as a **block support**. when I found the issue I was testing with the heading block. The same class name is used with blocks that have the text align **block attribute.**

t-hamano:
> The same class name is used with blocks that have the text align block attribute.

For core blocks that support `textAlign` as a _block attribute_, it is planned to migrate them to _block support_.

- #60763

carolinan:
I understand that is the plan, and I fully agree with it.  But it is not implemented right now, what is the expected test result for blocks, including third party blocks, that uses the attribute?

t-hamano:
> what is the expected test result for blocks, including third party blocks, that uses the attribute?

The UI will be displayed twice. But isn't this the same for all other block support?

For example, if you add `supports.aspectRatio` to the Image block that supports `aspectRatio` as an attribute, the Aspect Ratio UI will appear in both the Settings and Styles tabs.

Developers who want to migrate from attribute to block support will need to add explicit migrations, but I believe this applies to all block support.

carolinan:
I am confused, I meant for this PR, which I thought was intended to make the block toolbar option work again, not the conversion.

t-hamano:
> I thought was intended to make the block toolbar option work again, not the conversion.

That's right.

> This issue isn't an issue unless developers explicitly opt-in to "textAlign" support for their custom blocks, or add "textAlign" support by filtering core block support.

My explanation for this was not accurate.

For example, the Heading block currently supports text alignment as an _attribute_, not as a _block support_. If you define a default text alignment via `theme.json`, you cannot override the text alignment from a block instance (block toolbar).

```json
{
	"version": 3,
	"styles": {
		"blocks": {
			"core/heading": {
				"typography": {
					"textAlign": "center"
				}
			}
		}
	}
}
```

This PR is an attempt to address these issues as well.

However, `styles.blocks.{blockName}.typography.textAlign` itself is a new property introduced in WP6.6, so just like block support, it shouldn’t cause any problems unless developers explicitly opt in to this style.

carolinan:
As a theme developer, I do want to explicitly opt-in though 😅 because it means that default text alignments can be removed from the template and pattern code, making them much more flexible.

t-hamano:
If anyone has any good ideas for moving this forward, please let me know 🙏

Here's a summary of this PR and the latest status:

- If text alignment styles are defined via the global styles or theme.json, the following style is generated:
  ```css
  :root :where(.wp-block-media-text) {
      text-align: center;
  }
- Changing the text alignment of a block instance adds a `.has-text-align-` class. This style is part of the block library styles:
  ```css
  .has-text-align-right {
      text-align: right;
  }
  ```
- Both selectors have the same specificity ( `0-1-0` ), and the styles frm the global styles are loaded _after_ the block library styles, so we can't override the text alignment style of the block instance.
- I'm exploring an approach to solve this problem without increasing the CSS specificity of the `.has-text-align-` selector.
- I was able to inject text alignment-related styles after the global styles via the style engine on the frontend.
However, I don't know how to achieve a similar approach on the editor side (see [this comment](https://github.com/WordPress/gutenberg/pull/62260#issuecomment-2180520315)).

aaronrobertshaw:
> I was able to inject text alignment-related styles after the global styles via the style engine on the frontend.
However, I don't know how to achieve a similar approach on the editor side

At a glance, it appears in https://github.com/WordPress/gutenberg/pull/62260/commits/3b093caee0b2c96f1365cb0889c51b38402fac33 you are bundling these text alignment styles in with the others created for block supports. The equivalent of that in the editor is probably `useStyleOverride`. At least that's what each of the block supports generating per-block-instance styles do.

I'm not sure `useStyleOverride` would be the right tool for the job here though and maintaining the style load order may prove tricky. Ultimately, these styles are added via the `EditorStyles` component, which could provide an opportunity (albeit very hacky) to enqueue these text alignment styles. 

Could the block support in the editor, not only add the `has-text-align-*` class but also inline styles? This would need to happen when getting block props in the edit component and not the save function. That could potentially ensure text alignment styles in the editor, while the styles added in https://github.com/WordPress/gutenberg/pull/62260/commits/3b093caee0b2c96f1365cb0889c51b38402fac33 handle the frontend 🤷 

This is all just scraping the bottom of the barrel for ideas though. Everything else I can think of ends up requiring any block adopting the support to need deprecations which to date has been deemed undesirable.

Hopefully others have some better ideas 🤞

kasparsd:
The main issue is that `.has-text-align-*` is used as a block-level override helper (highest specificity) while being loaded as part of `library/common.scss` _before_ the theme/site-level overrides.

We have only two ways to make these nested styles work:

1. increase the specificity of CSS with every level of override (global=generic, block-level=very-specific);
2. adjust the order of CSS rules so that _later_ rules override the previous ones.

In my experience relying on the order is harder than controlling the specificity. I feel like 99fe9554bcac09f8f1453b748e37b652c5eda171 is the best route to go since it maintains backwards compatibility and provides the smallest changeset.

youknowriad:
@t-hamano Should we resurface this again. I think it's important to switch to the block support for paragraph to allow global styles support (justification...)

youknowriad:
I think the loading order is correct. block library is "core" styles that should be overridable by themes (theme.json) which should be overridable by instance block styles.

The issue is that these classnames are "user instance block styles" but defined by the block library (it can happen) but in this case, we need to ensure the specificity is higher than the theme.json styles.

Assuming https://github.com/WordPress/gutenberg/commit/99fe9554bcac09f8f1453b748e37b652c5eda171 doesn't cause any unintended consequences, I think I'd be ok with this fix.

t-hamano:
@youknowriad Sorry for the late reply.

> Assuming https://github.com/WordPress/gutenberg/commit/99fe9554bcac09f8f1453b748e37b652c5eda171 doesn't cause any unintended consequences, I think I'd be ok with this fix.

I tried this approach again. As far as I know, there is no problem with this approach.

youknowriad (approved):
I'd appreciate more testing but I think this is good. It would be great to migrate all the blocks to use the block support after this. (I have a PR for paragraph already)

PR #40718, opened 2022-04-29
getBlockTransformItems: Support single block object

Mamaduka:
## What?
PR updates `getBlockTransformItems` selector `blocks` arguments. Now it allows passing a single object or array of objects. This matches how `getBlocksByClientId` works.

## Why?
When trying to get a transformation for a single block, we often have to pass blocks argument like this - `[ getBlock( clientId ) ]`. If arguments aren't referentially equal, the `createSelector` returns the new value and re-rerenders the component. This can affect performance if left unchecked.

Discovered while working on #40696.

## How?
The blocks argument now is cast into an array inside the selector.

## Testing Instructions
Tests should be passing.

PR #62263, opened 2024-06-04
Inserter: Return the same items when the state and parameters don't change

Mamaduka:
## What?
PR updates the selector in the `useBlockTypesState` hook to return the same `items` when the state and parameters haven't changed.

This is a follow-up to https://github.com/WordPress/gutenberg/pull/62169#issuecomment-2145155218.

## Why?
The `getInserterItems` is a memoized selector, and when an argument is different on each call, it will return different (fresh) values. This behavior makes it complicated to add `array` or `object` type parameters to similar selectors. Example: #40718.

## How?
Memoize the `options` passed to the selector on the component level.

I don't like shifting this responsibility to the consumer, but I see no other way besides changing the `getInserterItems` signature.

## Testing Instructions
1. Open a post or page.
2. Open the global inserter.
3. Confirm the `useBlockTypesState` hook generates no warning.

Same.

## Screenshots or screencast 
CleanShot 2024-06-03 at 16 58 49

### DIFF
--- a/packages/block-editor/src/components/inserter/hooks/use-block-types-state.js
+++ b/packages/block-editor/src/components/inserter/hooks/use-block-types-state.js
@@ -8,7 +8,7 @@ import {
 	parse,
 } from '@wordpress/blocks';
 import { useSelect } from '@wordpress/data';
-import { useCallback } from '@wordpress/element';
+import { useCallback, useMemo } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -25,13 +25,18 @@ import { withRootClientIdOptionKey } from '../../../store/utils';
  * @return {Array} Returns the block types state. (block types, categories, collections, onSelect handler)
  */
 const useBlockTypesState = ( rootClientId, onInsert, isQuick ) => {
+	const options = useMemo(
+		() => ( { [ withRootClientIdOptionKey ]: ! isQuick } ),
+		[ isQuick ]
+	);
 	const [ items ] = useSelect(
 		( select ) => [
-			select( blockEditorStore ).getInserterItems( rootClientId, {
-				[ withRootClientIdOptionKey ]: ! isQuick,
-			} ),
+			select( blockEditorStore ).getInserterItems(
+				rootClientId,
+				options
+			),
 		],
-		[ rootClientId, isQuick ]
+		[ rootClientId, options ]
 	);
 
 	const [ categories, collections ] = useSelect( ( select ) => {
--- a/packages/block-editor/src/store/selectors.js
+++ b/packages/block-editor/src/store/selectors.js
@@ -76,6 +76,8 @@ const EMPTY_ARRAY = [];
  */
 const EMPTY_SET = new Set();
 
+const EMPTY_OBJECT = {};
+
 /**
  * Returns a block's name given its client ID, or null if no block exists with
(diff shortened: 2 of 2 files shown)


packages/block-editor/src/components/inserter/hooks/use-block-types-state.js
```diff
@@ -25,13 +25,18 @@ import { withRootClientIdOptionKey } from '../../../store/utils';
...
+	);
 	const [ items ] = useSelect(
 		( select ) => [
-			select( blockEditorStore ).getInserterItems( rootClientId, {
-				[ withRootClientIdOptionKey ]: ! isQuick,
-			} ),
+			select( blockEditorStore ).getInserterItems(
+				rootClientId,
+				options
+			),
```
ellatrix, line 37: Shouldn't we account for this in `getInserterItems`? Also the default there is an empty object on every call, so we should change that too?

Mamaduka: > Shouldn't we account for this in getInserterItems?

That would be ideal, but we need to figure out how. Functions have no control over the arguments they receive.

> Also the default there is an empty object on every call, so we should change that too?

Yes. We should either pass `undefined` or an `EMTPY_OBJECT`.

ellatrix: Yeah, this is a good fix for now. I wonder if we should just create another private `getInserterItems` selector and avoid the options altogether. Actually this would also allow us to remove all the reusable blocks logic. We're adding these blocks, just to filter them back out in the UI :smile:

Mamaduka:
Confirmed that this also fixes perf regression https://github.com/WordPress/gutenberg/actions/runs/9361910664/attempts/3#summary-25773603112

tyxla (approved):
Good one 👍 

Ideally, we'd address it inside `getInserterItems` but I agree that it can be tricky. 

Thanks @Mamaduka 🚀

ellatrix:
Do we still need the default parameter change to fix calls without options?

Mamaduka:
> Do we still need the default parameter change to fix calls without options?

Yes, and I just pushed an update for it.

packages/block-editor/src/store/selectors.js
```diff
@@ -76,6 +76,8 @@ const EMPTY_ARRAY = [];
  */
 const EMPTY_SET = new Set();
 
+const EMPTY_OBJECT = {};
```
Mamaduka, line 79: I've omitted the comment here. The other ones are a bit outdated and we should probably revise them together.

ellatrix:
This will need to be backported

packages/block-editor/src/store/selectors.js
```diff
@@ -1996,7 +1998,7 @@ const buildBlockTypeItem =
  */
 export const getInserterItems = createRegistrySelector( ( select ) =>
 	createSelector(
-		( state, rootClientId = null, options = {} ) => {
+		( state, rootClientId = null, options = EMPTY_OBJECT ) => {
```
ellatrix, line 2001: Actually I guess this doesn't matter because `createSelector` is not aware of the default, it would just not be defined. Sorry my mistake :D

ellatrix:

PR #62265, opened 2024-06-04
Documentation: Better changelogs for the JSX transform upgrade

youknowriad:
Related #62202

## What?

The new JSX transform upgrade has a side consequence that we overlooked a bit. It meant that the new build packages don't support WP 6.5 and prior anymore since they require the new JSX runtime to be present.

The current PR updates the related package changelogs to clarify that. People building for 6.5 and before should stick with their current dependencies.

### DIFF
--- a/packages/babel-preset-default/CHANGELOG.md
+++ b/packages/babel-preset-default/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+**Note** If you're using @wordpress/scripts for building JS scripts to target WordPress 6.5 or anterior, you should not upgrade to this version and continue using @wordpress/babel-preset-default@7.
+
 -   Use React's automatic runtime to transform JSX ([#61692](https://github.com/WordPress/gutenberg/pull/61692)).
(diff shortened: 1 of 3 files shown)


packages/babel-preset-default/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+**Note** If you're using @wordpress/scripts for building JS scripts to target WordPress 6.5 or anterior, you should not upgrade to this version and continue using @wordpress/babel-preset-default@7.
```
Mamaduka: I had to look it up `anterior`. Maybe we should simplify the message here.

youknowriad: Any suggestions? I guess it's easier to understand for me (coming from French)

Mamaduka: The phrase I usually see is "WP 6.5 or earlier". But let's ask native speakers 😅

youknowriad: earlier is easier indeed. I spent let's say 20 seconds on it but didn't find the obvious word :P

Mamaduka (approved):
Thank you, Riad!

PR #62267, opened 2024-06-04
Make edit site pagination buttons accessibly disabled

afercia:
Follow-up to https://github.com/WordPress/gutenberg/pull/62080
Cc @mirka 

## What?

Pagination buttons are a typical case where buttons must be still focusable also when 'disabled'. 
Looking at the code statically as per the investigation made on  https://github.com/WordPress/gutenberg/pull/62080, it may appear that they can be 'truly' disabled. However, when using and interacting with these buttons via a keyboard here's the typical scenario where a focus loss must be prevented:
- Focus is on the Next or Last button.
- Press the Entrer key as many times as necessary to reach the last page.
- When on the last page, the button gets disabled dynamically _while_ it has focus.
- As such, focus is lost.

## Why?

To prevent focus losses.

## How?

Adds the `__experimentalIsFocusable`
Removes the eslint disable comment.
Adjusts the style.

Note: the styling of a disabled-focusable button should work consistently with all button variants. For example, the Undo and Redo buttons use `opacity` while these buttons only use a color change. This should be addressed in the Button component though and there should not be 'local' styling overrides. I'll create a separate issue.

## Testing Instructions

- Go to Site Editor > Styles > Revisions.
- Meks sure to have many styles revisions so that the panel will show pagination buttons.
- Use the keyboard.
- Navigate to the pagination buttons and reach the last page (pr going backwards, reach the first page).
- Observe the buttons are still focusable and get an `aria-disabled="true"` attribute instead of a `disabled` one.

### DIFF
--- a/packages/edit-site/src/components/global-styles/screen-revisions/style.scss
+++ b/packages/edit-site/src/components/global-styles/screen-revisions/style.scss
@@ -180,7 +180,8 @@
 		margin-bottom: $grid-unit-05;
 		line-height: 1.2;
 	}
-	.components-button.is-tertiary:disabled {
+	.components-button.is-tertiary:disabled,
+	.components-button.is-tertiary[aria-disabled="true"] {
 		color: $gray-600;
 	}
 	.components-button.is-tertiary:hover {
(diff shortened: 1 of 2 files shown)


mirka (approved):
Ah ok that makes sense! Thanks for catching 🙏

ellatrix:

ellatrix:

PR #61560, opened 2024-05-10
Adjust the icons and text of the binding connected blocks

kevin940726:
## What?

Close https://github.com/WordPress/gutenberg/issues/61515 and https://github.com/WordPress/gutenberg/issues/61547.

Adjust the icons and the text descriptions of the binding connected blocks (pattern overrides, connected post meta, etc.)

## Why?

The generic connection icon is not that helpful and the block type names should be helpful.

## How?

Adjust `<BlockBindingsToolbarIndicator>`.

## Testing Instructions

- Create some patterns with overrides and some post meta connected blocks.

You can paste the markup here to create an empty post-meta connected paragraph.
```html
<!-- wp:paragraph {"metadata":{"name":"connect","bindings":{"content":{"source":"core/post-meta","args":{}}}}} -->
<p>Connected</p>
<!-- /wp:paragraph -->
```

## Screenshots or screencast 
**A pattern overrides block**
image

**Multiple pattern overrides blocks**
image

**A post-meta connected block**
image

**Multiple post-meta connected blocks**
image

PR #62268, opened 2024-06-04
Fix site editor breaking when user selects bound and non-bound blocks at the same time

SantosGuillamot:
## What?
Fix a bug that breaks the site editor when user select blocks that use bindings and blocks that don't.

If I am not mistaken, it was introduced in [this pull request](https://github.com/WordPress/gutenberg/pull/61560), so I might be missing context.

## Why?
It breaks the editor.

## How?
At [this point](https://github.com/WordPress/gutenberg/compare/fix/site-editor-breaking-with-bindings?expand=1#diff-53c7becd18a8bab09d6e368f39a58eacb5ac5dcb25a80b2826308dedeaafc570R60-R62), ensure that if there are no bindings it iterates through an empty object.

## Testing Instructions
* Go to the site editor home.
* Add a block connected to any custom field.
```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"my_custom_field"}}}}} -->
<p>Default</p>
<!-- /wp:paragraph -->
```
* Add any other block.
* Select the bound block.
* Multi select the other block using "Shift", for example.
* Check that the site editor keeps working.

## Screenshots or screencast 

**Before**

https://github.com/WordPress/gutenberg/assets/34552881/2fcb09d8-8d56-41c4-9f18-f181bf3f61fe

**After**

https://github.com/WordPress/gutenberg/assets/34552881/58890202-d9f3-47ff-91db-99e7d31d8e9a

### DIFF
--- a/packages/block-editor/src/components/block-bindings-toolbar-indicator/index.js
+++ b/packages/block-editor/src/components/block-bindings-toolbar-indicator/index.js
@@ -57,7 +57,9 @@ export default function BlockBindingsToolbarIndicator( { clientIds } ) {
 				isConnectedToPatternOverrides: getBlocksByClientId(
 					clientIds
 				).some( ( block ) =>
-					Object.values( block?.attributes.metadata?.bindings ).some(
+					Object.values(
+						block?.attributes?.metadata?.bindings || {}
+					).some(
 						( binding ) =>
 							binding.source === 'core/pattern-overrides'
 					)


artemiomorales (commented):
Works for me 👍
I'll defer approval to other folks who have more context, though.

kevin940726 (approved):
LGTM 👍 Thanks!

ellatrix:

PR #58194, opened 2024-01-24
Query Loop: Fix posts list variation detection

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/58168

Core Query Loop's variation `posts list` was missing a proper way of being detected if is active(by not using the `isActive` API). The reason for that is that the aforementioned API was introduced later.

This PR uses the proper APIs for variations and additionally removes the `__experimentalGetMatchingVariation`(that you could say it evolved into the `getActiveBlockVariation`).

## Testing Instructions
1. Existing Query Loop blocks should remain intact
2. Insert a `Posts List` block(the variation) and observe that as long as you also preserve the queried postType to be `post`, the proper title is shown in all places(list view, block card, breadcrumbs, etc..)

## Bonus testing
1. Register a block patterns with `blockTypes: ['core/query/posts-list'] and when you insert the block in the editor click `Choose`. Your pattern should be suggested.

PR #62272, opened 2024-06-04
Block Variations: Compare objects based on given properties

ockham:
## What?

If a block variation has an `isActive` property that is an array of strings (attribute names), and one of the attributes referenced therein is an object, compare that attribute based on the object properties specified by the variation, rather than by reference.

## Why?
Object attributes of a variation are currently compared by reference to the given block's attributes, meaning that the comparison will never return `true`.

For example, take the following variation of the Paragraph block:

```diff
diff --git a/packages/block-library/src/paragraph/block.json b/packages/block-library/src/paragraph/block.json
index 6f11baf838..8bd350fb1e 100644
--- a/packages/block-library/src/paragraph/block.json
+++ b/packages/block-library/src/paragraph/block.json
@@ -70,5 +70,25 @@
 		}
 	},
 	"editorStyle": "wp-block-paragraph-editor",
-	"style": "wp-block-paragraph"
+	"style": "wp-block-paragraph",
+	"variations": [
+		{
+			"name": "warning",
+			"title": "Warning",
+			"icon": "warning",
+			"attributes": {
+				"metadata": {
+					"bindings": {
+						"content": {
+							"source": "core/post-meta",
+							"args": {
+								"key": "text_field"
+							}
+						}
+					}
+				}
+			},
+			"isActive": [ "metadata.bindings.content" ]
+		}
+	]
 }
```

Prior to this PR, `isActive` would return `false` for the given variation, because it would compare the given block's `metadata.bindings.content` (nested) attribute by object reference to the variation's. 

## How?
By (recursively) iterating over the object properties specified by the variation and comparing those to the given block's. (In the example above, those properties are `source` and `args`; since `args` is an object itself, the comparison steps into it recursively and compares its `key` property to the given block's as well).

Note that this PR re-introduces the `matchesAttributes` util function (albeit in a different place) that was previously removed by #58194.

## Testing Instructions
Smoke-test block variations and verify that they still work as before.
Optionally, try adding the above variation to the Paragraph block, and verify that the variation is correctly detected (thus fixing the issue reported [in this comment](https://github.com/WordPress/gutenberg/pull/62031#issuecomment-2142102041)).

### DIFF
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -12,7 +12,7 @@ import { RichTextData } from '@wordpress/rich-text';
 /**
  * Internal dependencies
  */
-import { getValueFromObjectPath } from './utils';
+import { getValueFromObjectPath, matchesAttributes } from './utils';
 
 /** @typedef {import('../api/registration').WPBlockVariation} WPBlockVariation */
 /** @typedef {import('../api/registration').WPBlockVariationScope} WPBlockVariationScope */
@@ -276,6 +276,17 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 				);
 				if ( blockAttributeValue instanceof RichTextData ) {
 					blockAttributeValue = blockAttributeValue.toHTMLString();
+				}
+				// If the attribute value is an object, we need to compare its properties.
+				if (
+					variationAttributeValue !== null &&
+					typeof variationAttributeValue === 'object' &&
+					variationAttributeValue.constructor === Object
+				) {
+					return matchesAttributes(
+						blockAttributeValue,
+						variationAttributeValue
+					);
 				}
 				return variationAttributeValue === blockAttributeValue;
 			} );
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -277,18 +277,10 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 				if ( blockAttributeValue instanceof RichTextData ) {
 					blockAttributeValue = blockAttributeValue.toHTMLString();
 				}
-				// If the attribute value is an object, we need to compare its properties.
-				if (
-					variationAttributeValue !== null &&
-					typeof variationAttributeValue === 'object' &&
-					variationAttributeValue.constructor === Object
-				) {
-					return matchesAttributes(
-						blockAttributeValue,
-						variationAttributeValue
-					);
-				}
-				return variationAttributeValue === blockAttributeValue;
+				return matchesAttributes(
+					blockAttributeValue,
+					variationAttributeValue
+				);
 			} );
 			if ( isMatch && definedAttributesLength > maxMatchedAttributes ) {
 				match = variation;
--- a/packages/blocks/src/store/utils.js
+++ b/packages/blocks/src/store/utils.js
@@ -18,3 +18,27 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
 	} );
 	return value ?? defaultValue;
 };
+
+/**
+ * Determine whether a set of object properties matches a given object.
+ *
+ * Given an object of block attributes and an object of variation attributes,
+ * this function checks recursively whether all the variation attributes are
+ * present in the block attributes object.
+ *
+ * @param {Object} blockAttributes     The object to inspect.
+ * @param {Object} variationAttributes The object of property values to match.
+ * @return {boolean} Whether the block attributes match the variation attributes.
+ */
+export function matchesAttributes( blockAttributes, variationAttributes ) {
+	return Object.entries( variationAttributes ).every( ( [ key, value ] ) => {
+		if (
+			typeof value === 'object' &&
+			value !== null &&
+			typeof blockAttributes[ key ] === 'object'
+		) {
+			return matchesAttributes( blockAttributes[ key ], value );
+		}
+		return blockAttributes[ key ] === value;
+	} );
+}
--- a/packages/blocks/src/store/utils.js
+++ b/packages/blocks/src/store/utils.js
@@ -19,6 +19,14 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
 	return value ?? defaultValue;
 };
 
+function isObject( candidate ) {
+	return (
+		typeof candidate === 'object' &&
+		candidate.constructor === Object &&
+		candidate !== null
+	);
+}
+
 /**
  * Determine whether a set of object properties matches a given object.
  *
@@ -31,17 +39,12 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
  * @return {boolean} Whether the block attributes match the variation attributes.
  */
 export function matchesAttributes( blockAttributes, variationAttributes ) {
-	if ( ! blockAttributes ) {
-		return false;
+	if ( isObject( blockAttributes ) && isObject( variationAttributes ) ) {
+		return Object.entries( variationAttributes ).every(
+			( [ key, value ] ) =>
+				matchesAttributes( blockAttributes?.[ key ], value )
+		);
 	}
-	return Object.entries( variationAttributes ).every( ( [ key, value ] ) => {
-		if (
-			typeof value === 'object' &&
-			value !== null &&
-			typeof blockAttributes[ key ] === 'object'
-		) {
-			return matchesAttributes( blockAttributes[ key ], value );
-		}
-		return blockAttributes[ key ] === value;
-	} );
+
+	return blockAttributes === variationAttributes;
 }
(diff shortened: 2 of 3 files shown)


packages/blocks/src/store/utils.js
```diff
@@ -18,3 +18,15 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
 	} );
 	return value ?? defaultValue;
 };
+
+export function matchesAttributes( blockAttributes, variation ) {
+	return Object.entries( variation ).every( ( [ key, value ] ) => {
```
ntsekouras: I'm wondering if we have a util for this already with all the removal of lodash work.. --cc @tyxla 

Maybe `isShallowEqual`? Additionally it seem we would have the same issue with `arrays`.

packages/blocks/src/store/selectors.js
```diff
@@ -262,17 +262,28 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
...
-					attributeValue ===
-					getValueFromObjectPath( variation.attributes, attribute )
+				const blockAttributeValue = getValueFromObjectPath(
+					attributes,
+					attribute
 				);
+				// If the attribute value is an object, we need to compare its properties.
+				if (
+					variationAttributeValue !== null &&
+					typeof variationAttributeValue === 'object'
```
cbravobernal: In Interactivity, we also check that `variationAttributeValue.constructor === Object`.
Would should check that we need here too.

packages/blocks/src/store/utils.js
ockham: I actually asked @tyxla about this yesterday 😄 Turns out we used to have a util! I added a note about it to the PR description:

> Note that this PR re-introduces the `matchesAttributes` util function (albeit in a different place) that was previously removed by https://github.com/WordPress/gutenberg/pull/58194.

ockham: (`isShallowEqual` isn't quite the right fit: We only need the given block to match all the variation object attribute's properties, but it's okay if the block's object attribute has additional properties. E.g. in the example from the PR description, if the given block's `metadata.bindings.content` object has an additional `label` property, the variation should still match it.)

packages/blocks/src/store/utils.js
```diff
@@ -18,3 +18,15 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
 	} );
 	return value ?? defaultValue;
 };
+
```
ockham, line 21: Note to self: Add JSDoc.

packages/blocks/src/store/utils.js
ockham: > Additionally it seem we would have the same issue with `arrays`.

That's a fair point, but I'd rather tackle that separately. It's a bit less clear to me what the desired behavior should be for them, so I think it deserves a bit of dedicated discussion. If a variation has

```json
"attributes": {
    "arrayAttr": [ "a", "b", "c" ]
},
"isActive": [ "arrayAttr" ]
```

then which of the following blocks should we match?

1. `[ "a", "b", "c", "d" ] // Variation contains a subset`
2. `[ "b", "c", "a" ] // Variation matches array items but not array order`

packages/blocks/src/store/selectors.js
ockham: Ah, interesting! I found a few instances of `isObject` helpers across our codebase that didn't seem to include that criterion 🤔 E.g. https://github.com/WordPress/gutenberg/blob/eaf2d6550b39587dc8945bf2e20fc4290a7139fe/packages/blocks/src/api/registration.js#L127-L129

I guess it's to exclude arrays?

packages/blocks/src/store/utils.js
ockham: https://github.com/WordPress/gutenberg/pull/62272/commits/6dd85a715783050483295f2c7fe133dcb1230442

packages/blocks/src/store/selectors.js
cbravobernal: Yep. Do we need to exclude them here?

packages/blocks/src/store/utils.js
tyxla: Just to close the loop, based on the need @ockham described, I agree that it's best if we reintroduce that util.

packages/blocks/src/store/utils.js
```diff
@@ -18,3 +18,26 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
...
+ * @param {Object} variationAttributes The object of property values to match.
+ * @return {boolean} Whether the block attributes match the variation attributes.
+ */
+export function matchesAttributes( blockAttributes, variationAttributes ) {
+	return Object.entries( variationAttributes ).every( ( [ key, value ] ) => {
+		if (
+			typeof value === 'object' &&
+			typeof blockAttributes[ key ] === 'object'
+		) {
+			return matchesAttributes( blockAttributes[ key ], value );
```
tyxla: Should we add an exception for `null`? `typeof null` is `object`, but if both `value` and `blockAttributes[ key ]` are `null`, we'll call `Object.entries( null )` which will be a `TypeError`.

packages/blocks/src/store/selectors.js
ockham: Yeah, we might want to do that (also see https://github.com/WordPress/gutenberg/pull/62272#discussion_r1627318785). I'll add the check.

ockham: 1b408c75aec

packages/blocks/src/store/utils.js
ockham: Yes, good spot! I added a check in b69dc0af026 which I think should be sufficient.

packages/blocks/src/store/selectors.js
```diff
@@ -262,17 +262,29 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
...
 					return false;
 				}
-				return (
-					attributeValue ===
-					getValueFromObjectPath( variation.attributes, attribute )
+				const blockAttributeValue = getValueFromObjectPath(
+					attributes,
+					attribute
 				);
+				// If the attribute value is an object, we need to compare its properties.
```
gziolo: How far is it from `isShallowEqual` from `@wordpress/is-shallow-equal` at this point?

https://github.com/WordPress/gutenberg/blob/06470bbb92e5da50c3f4f3b5232d9d49cb021119/packages/is-shallow-equal/src/index.js#L23-L33

ockham:
@cbravobernal pointed out to me that the example in the PR desc is currently not working (thank you!)

~The reason is that `metadata` is a global attribute that's not present in the list of the block type's "own" attributes that we use to filter the attributes in `isActive`: https://github.com/WordPress/gutenberg/blob/68beef3db750b18f909aa7bdd5dfe79a54ef7fdf/packages/blocks/src/store/selectors.js#L245~

~We'll need to think of a fix for that. The "easy" fix is to look up global attributes inside this selector, but we might want to include them in other places that check for a block's attributes as well.~

~I'm inclined to tackle this in a separate PR.~

_Edit: Ignore, I just missed the `metadata` attribute_ 🤦‍♂️ 

image

ockham:
> @cbravobernal pointed out to me that the example in the PR desc is currently not working (thank you!)

I think I tracked down the error and fixed it (see the last couple of commits).

vcanales:
Checking in on PRs with the Backport to Gutenberg RC. I'll release 18.5 later today, do you think this'll make it in time?

ockham:
> Checking in on PRs with the Backport to Gutenberg RC. I'll release 18.5 later today, do you think this'll make it in time?

Depends on whether we'll get approval in time 🤞

cbravobernal (approved):
Now works as expected. Great job @ockham !

packages/blocks/src/store/utils.js
```diff
@@ -19,6 +19,14 @@ export const getValueFromObjectPath = ( object, path, defaultValue ) => {
 	return value ?? defaultValue;
 };
 
+function isObject( candidate ) {
+	return (
+		typeof candidate === 'object' &&
```
cbravobernal, line 24: We called it `item` in Interactivity. I like this nomenclature, maybe we should update to keep consistency in the codebase. Ping @DAreRodz @luisherranz

vcanales:
I just cherry-picked this PR to the release/18.5 branch to get it included in the next release: 2f2677a7fe

packages/blocks/src/store/utils.js
luisherranz: I need more context. What exactly are we referring to as "item" in the Interactivity API? What is being referred to here?

cbravobernal: https://github.com/WordPress/gutenberg/blob/81cde31199886a188fc47b80c66042c77b4dacb4/packages/interactivity/src/store.ts#L18

In Interactivity API, we check that several "items" are a JSON object in the same file (sources, state, config, proxy handlers, etc). So we call the variable to check "item".

"Candidate" seems also a good general definition, and would be ok to use the same name in all functions that are just checking that an item is an object (there are a few of them across the workspace).

luisherranz: Oh, ok. You can change that if you want 🙂

ellatrix:

PR #62273, opened 2024-06-04
Mobile Release v1.119.1

geriux:
## Description
Release 1.119.1 of the react-native-editor and Gutenberg-Mobile.


packages/block-library/src/missing/edit.native.js
```diff
@@ -48,11 +44,38 @@ export class UnsupportedBlockEdit extends Component {
...
+		const {
+			canEnableUnsupportedBlockEditor,
+			isUnsupportedBlockEditorSupported,
+		} = this.props;
+
+		return (
+			! canEnableUnsupportedBlockEditor &&
+			isUnsupportedBlockEditorSupported
+		);
+	}
```
fluiddot: I checked how we calculate if the unsupported block can be used and it seems to differ with these conditions:

https://github.com/WordPress/gutenberg/blob/fb434d8340add383e16e496a0d9cb91fa727c24c/packages/block-editor/src/components/unsupported-block-details/index.native.js#L107-L110

I haven't tested all UBE test cases yet but wanted to highlight in case we identify any issue.

fluiddot (approved):
LGTM 🎊 !

For reference, I also checked the changes between 1.119.0 and 1.119.1: https://github.com/WordPress/gutenberg/compare/rnmobile/1.119.0...rnmobile/release_1.119.1

packages/block-library/src/missing/edit.native.js
fluiddot: Oh, I've realized that `canEnableUnsupportedBlockEditor` will be false if the user needs to change the Jetpack settings. Hence, we'll only display UBE directly from the block in case there's no need to present the bottom sheet 👍.

geriux: Yup! I will add the `isEditableInUnsupportedBlockEditor` check though, to not allow blocks that don't support it and send that to the standard flow instead of the "Tap to edit" 👍

geriux: Addressed in https://github.com/WordPress/gutenberg/pull/62273/commits/6a5d15a3e829faa7d8a4966a418ee15dd3534f90

PR #62279, opened 2024-06-04
Fix: Show homepage link on frontpage instead of the slug.

### LINKED ISSUE #62065: Inspector: Polish "Link" button / dialog for static homepage
Here's the Link button / dialog for regular pages:

Screenshot 2024-05-28 at 13 24 13

Here it is for a static homepage:

Screenshot 2024-05-28 at 13 23 58

The button label shows the slug, but the dialog shows only the TLD with no edit-ability. In the case of the homepage, there's very little value in allowing users to change the slug, and it could lead to confusion.

* Should the Link row be hidden entirely for the static homepage?
* Or should it simply list the TLD as an externallink?

Screenshot 2024-05-28 at 13 28 17

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/62065
On the trunk for the homepage/frontpage of a site, we are still showing the slug as the link which is wrong. This PR fixes the issue and shows the true homepage URL for the frontpage.

cc: @jameskoster 

## Testing Instructions
I verified the true homepage URL is shown for pages marked as the frontpage.

Before:

Screenshot 2024-06-04 at 13 34 59

After:
Screenshot 2024-06-04 at 13 32 15

### DIFF
--- a/packages/editor/src/components/post-url/panel.js
+++ b/packages/editor/src/components/post-url/panel.js
@@ -6,6 +6,7 @@ import { useSelect } from '@wordpress/data';
 import { Dropdown, Button } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
 import { safeDecodeURIComponent } from '@wordpress/url';
+import { store as coreStore } from '@wordpress/core-data';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


carolinan (approved):
Works well in my test.

ellatrix:

PR #62280, opened 2024-06-04
Button Block: Allow custom width

### LINKED ISSUE #62247: Button Block: Allow custom width
Currently with the button block, I can choose several popular widths, 25%, 50%, 75%, 100%. However I want to create a pattern with 3 buttons together in a row, and need to provide a more specific pattern. Example of the pattern:

Screenshot 2024-06-03 at 9 14 45 AM

Keep the popular widths, but have an advanced option to provide a specifics widths, px, %, rem, etc.

prasadkarmalkar:
Closes: #62247
## What?

Add FontSizePicker for button block for custom width size
## Why?

To allow custom width for button block
## How?

We are using the FontSizePicker for width of the button so that we can choose the custom size as well. And instead of giving width using class we are using the style attribute.
## Screenshots or screencast 
Screenshot 2024-06-04 at 5 20 33 PM
Screenshot 2024-06-04 at 5 25 30 PM

### DIFF
--- a/packages/block-library/src/button/edit.js
+++ b/packages/block-library/src/button/edit.js
@@ -17,12 +17,11 @@ import { unlock } from '../lock-unlock';
 import { __ } from '@wordpress/i18n';
 import { useEffect, useState, useRef, useMemo } from '@wordpress/element';
 import {
-	Button,
-	ButtonGroup,
 	PanelBody,
 	TextControl,
 	ToolbarButton,
 	Popover,
+	FontSizePicker,
 } from '@wordpress/components';
 import {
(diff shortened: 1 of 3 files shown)


mirka:
Thank you for contributing! 🙌 

Please see https://github.com/WordPress/gutenberg/issues/62247#issuecomment-2274324749 for a suggestion to use a more standard mechanism for this.

skorasaurus:
thanks for your effort although allowing custom width of the button block is now already implemented in https://github.com/WordPress/gutenberg/pull/74242

Issue #62236, opened 2024-06-03
Editing bound paragraph in post editor causes "Site Updated" message to appear


artemiomorales:
### Description

In the post editor, if one modifies a block connected to post metadata via block bindings, a "Site Updated" notification displays when saving the post, which is incorrect.

First raised in https://github.com/WordPress/gutenberg/pull/61811 and a part of the [block bindings iteration](https://github.com/WordPress/gutenberg/issues/60956#issuecomment-2144790954) for WordPress 6.6.

### Step-by-step reproduction instructions

<details>
<summary>1. Register post meta by adding this snippet to your theme's functions.php</summary>

```php
add_action( 'init', 'test_block_bindings' );

function test_block_bindings() {
	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'type'              => 'string',
			'default'           => 'default text value',
		)
	);
}
```

</details>

<details>

<summary>2. In the post editor, add a paragraph block bound to the custom field using the Code Editor</summary>

```
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_field"}}}}} -->
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Press the Save button.
4. Notice that the "Site Updated" message appears, despite the fact that you made no site-wide changes.

### Screenshots, screen recording, code snippet

<img width="1238" alt="site-updated-bug" src="https://github.com/WordPress/gutenberg/assets/5360536/2aac3dce-23df-4a6e-817c-f95d45cd15e2">

### Environment info

- Gutenberg Release Candidate 18.5

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

SantosGuillamot:
The changes to trigger the save panel after editing post meta in a post have been reverted in [this pull request](https://github.com/WordPress/gutenberg/pull/63412). I guess this is not an issue anymore and we just have to ensure that whatever we build doesn't trigger this again, right?

I've opened [this new issue](https://github.com/WordPress/gutenberg/issues/63425) to gather all the context and make decisions before working on a new solution.

Should we close this one?

artemiomorales:
[...]

PR #61811, opened 2024-05-21
Add indicator for metadata changes to Save Panel when reviewing modified entities

artemiomorales:
## What?

This is a first iteration of adding metadata to the Save Panel when reviewing modified entities.

## Why?

Begins addressing https://github.com/WordPress/gutenberg/issues/61405
Related to https://github.com/WordPress/gutenberg/pull/61753
We want to indicate to the user when they are modifying the value of custom fields.

## How?

Overrides existing logic to determine whether to show the Save Panel, while also showing that metadata has changed.

## Testing Instructions

<details>
<summary>1. Register post meta by adding this snippet to your theme's functions.php</summary>

```php
add_action( 'init', 'test_block_bindings' );

function test_block_bindings() {
	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'type'              => 'string',
			'default'           => 'default text value',
		)
	);
}
```

</details>

<details>

<summary>2. Add a paragraph block bound to the custom field using the Code Editor</summary>

```
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_field"}}}}} -->
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Add a normal paragraph block and modify it.
4. Run the following code in your browser console:
```
const meta = wp.data.select('core/editor').getEditedPostAttribute('meta');
meta['text_field'] = 'testing change';
wp.data.dispatch('core/editor').editPost( meta );
```
5. Press the Save button.
6. Verify that the Save Panel appears and shows that metadata has changed.

## Screenshots or screencast 

333905890-a40e6c2c-7918-42ce-9f6b-69047fe86971

PR #62284, opened 2024-06-04
Skip 'Site Updated' message if only post meta was saved

artemiomorales:
## What?

This PR aims to avoid rendering the "Site Updated" message if only post metadata in an entity has changed, namely in relation to [editing meta via block bindings](https://github.com/WordPress/gutenberg/pull/61753).

## Why?

Addresses https://github.com/WordPress/gutenberg/issues/62236
We don't want to erroneously show the "Site Updated" message.

## How?

Right now, it adds a flag to `saveEditedEntityRecord()` if edits include post meta, and prevents the "Site Updated" message from showing if this is the only kind of edit that occurred.

## Testing Instructions

<details>

```php

	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
		)
	);
}
```

</details>

<details>

<summary>2. In the post editor, add a paragraph block bound to the custom field using the Code Editor</summary>

```
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Press the Save button.
4. Notice that the "Site Updated" message does not appear, as expected.

### DIFF
--- a/packages/core-data/src/actions.js
+++ b/packages/core-data/src/actions.js
@@ -798,7 +798,6 @@ export const saveEditedEntityRecord =
 			name,
 			recordId
 		);
-
 		const record = { [ entityIdKey ]: recordId, ...edits };
 		return await {
 			values: dispatch.saveEntityRecord( kind, name, record, options ),
--- a/packages/core-data/src/actions.js
+++ b/packages/core-data/src/actions.js
@@ -798,8 +798,12 @@ export const saveEditedEntityRecord =
 			name,
 			recordId
 		);
+
 		const record = { [ entityIdKey ]: recordId, ...edits };
-		return await dispatch.saveEntityRecord( kind, name, record, options );
+		return await {
+			values: dispatch.saveEntityRecord( kind, name, record, options ),
+			metaChange: !! edits?.meta,
+		};
 	};
 
 /**
--- a/packages/editor/src/store/private-actions.js
+++ b/packages/editor/src/store/private-actions.js
@@ -203,7 +203,7 @@ export const saveDirtyEntities =
 					registry
 						.dispatch( noticesStore )
 						.createErrorNotice( __( 'Saving failed.' ) );
-				} else {
+				} else if ( ! values.every( ( value ) => value.metaChange ) ) {
 					registry
 						.dispatch( noticesStore )
 						.createSuccessNotice( __( 'Site updated.' ), {


artemiomorales:
I'm not sure of the best way to address this bug. Is the code here on the right track?

Modifying `saveEditedEntityRecord()` as this PR currently does would require refactoring a few other places where the function is used, which I'm not sure is the right move.

Also, I'm unsure if if just checking for `edits` object for a `meta` property is a robust way of determining whether to show the "Site Updated" message or not.

I'll keep looking at it, but any thoughts appreciated!

packages/core-data/src/actions.js
```diff
@@ -799,7 +799,10 @@ export const saveEditedEntityRecord =
 			recordId
 		);
 		const record = { [ entityIdKey ]: recordId, ...edits };
-		return await dispatch.saveEntityRecord( kind, name, record, options );
+		return await {
+			values: dispatch.saveEntityRecord( kind, name, record, options ),
+			metaChange: !! edits?.meta,
+		};
```
youknowriad, line 805: I don't think we should touch core-data for this change. I think core-data shouldn't have flags like that that are specific to given entities.

packages/editor/src/store/private-actions.js
```diff
@@ -203,7 +203,7 @@ export const saveDirtyEntities =
 					registry
 						.dispatch( noticesStore )
 						.createErrorNotice( __( 'Saving failed.' ) );
-				} else {
+				} else if ( ! values.every( ( value ) => value.metaChange ) ) {
```
youknowriad, line 206: metaChange is not the only thing that should prevent the "site updated" notice. I think we should think about this notice a bit more holistically. 

Should we show "site updated" when we change just a single entity (like a page or post or a template) or something? 

Maybe we should try to be smarter here. If the "pendingRecords" is a single entity we just show `${ entityTitle } saved successfully` and if there are multiple records we should "Site updated" or something.

youknowriad:
I also don't think this is a bug or something that is specific to block bindings or 6.6, so it's probably better addressed outside the 6.6 rush.

artemiomorales:
An alternative approach might include modifying [dirtyEntityRecords](https://github.com/WordPress/gutenberg/blob/d80b5a96f297629c832a7b59ee5ffcc68d67eb92/packages/editor/src/components/entities-saved-states/hooks/use-is-dirty.js#L26-L44) and doing a more substantial overhaul inside of [saveDirtyEntities](https://github.com/WordPress/gutenberg/blob/d80b5a96f297629c832a7b59ee5ffcc68d67eb92/packages/editor/src/store/private-actions.js#L155-L220), though it seems to me it would be a fairly substantial change.

artemiomorales:
@youknowriad Great, thanks for the feedback!

> I also don't think this is a bug or something that is specific to block bindings or 6.6, so it's probably better addressed outside the 6.6 rush.

@SantosGuillamot What do you think about this?

SantosGuillamot:
I agree that we should think more holistically and not think only about post meta changes. Actually, I see it related to what I shared [here](https://github.com/WordPress/gutenberg/pull/61811#discussion_r1615603578) about triggering the "Save panel" not only for post meta changes but potentially for other things like post description, excerpt... (especially when they have their own block binding source)

If we need to work on all of that to solve that issue, I agree that moving it outside 6.6 is probably the best.

artemiomorales:
> metaChange is not the only thing that should prevent the "site updated" notice. I think we should think about this notice a bit more holistically.
>
> Should we show "site updated" when we change just a single entity (like a page or post or a template) or something?
>
> Maybe we should try to be smarter here. If the "pendingRecords" is a single entity we just show ${ entityTitle } saved successfully and if there are multiple records we should "Site updated" or something.

> I agree that we should think more holistically and not think only about post meta changes. Actually, I see it related to what I shared https://github.com/WordPress/gutenberg/pull/61811#discussion_r1615603578 about triggering the "Save panel" not only for post meta changes but potentially for other things like post description, excerpt... (especially when they have their own block binding source)

@youknowriad @SantosGuillamot I created an issue to further discuss how we could approach this: https://github.com/WordPress/gutenberg/issues/62383
Please let me know anything is unclear or if it mischaracterizes the problem.

SantosGuillamot:
Since it seems there is a [new pull request for the same purpose](https://github.com/WordPress/gutenberg/pull/63223#top), should we close this one and keep the discussion there?

artemiomorales:
Closed in favor of https://github.com/WordPress/gutenberg/pull/63223

packages/editor/src/store/private-actions.js
artemiomorales: I decided to try a simpler approach in https://github.com/WordPress/gutenberg/pull/63223, [related comment here](https://github.com/WordPress/gutenberg/pull/63223#issuecomment-2219486695), let me know what you think 🙏

PR #62241, opened 2024-06-03
Inserter: Update Openverse API URLs

Mamaduka:
## What?
PR updates media category URLs for Openverse.

## Why?
The old https://api.openverse.engineering/ began redirecting to https://api.openverse.org/ today, causing an infinite loading state in the editors and e2e test failure.

See the full [discussion on Slack](https://wordpress.slack.com/archives/C02QB2JS7/p1717419383604539) (requires login).

## Testing Instructions
1. Open a post.
2. Navigate to Inserter > Media > Openverse
3. Confirm that Openverse media items are correctly loaded and the console has no request error.
4. Repeat the same step for the Site Editor.

### Testing Instructions for Keyboard
Same.

## Screeenshot
CleanShot 2024-06-03 at 16 51 06

Issue #56343, opened 2023-11-20
Add Block Inspector button to Block Options menu


scruffian:
## What problem does this address?
This was removed (by me!) in https://github.com/WordPress/gutenberg/pull/46709.

I think this was a mistake. New users don't know about the Block Inspector, and opening it is not intuitive for them. Every time I watch a user who isn't familiar with Gutenberg I see them looking in the Block Options menu for settings and being unable to find them.

## What is your proposed solution?
If we add a link to Block Settings from the Block Options then we will provide a path for these users to discover additional block settings.

PR #56000, opened 2023-11-09
Site Editor: use EditorProvider instead of custom logic

youknowriad:
Related to #52632 
Follow up to #55970

## What?

Refer to the discussion in #52632 for the reasoning. The ultimate goal is to be able to reuse more UI components (maybe the whole editor) between edit-post and edit-site.

So this PR updates the site editor to use `EditorProvider` instead of its custom `BlockEditorProvider` usage.

The biggest impact is going to be on three things which we'll have to test properly:

 - The different template/page modes in the site editor
 - Editing navigation in the site editor

PR #62286, opened 2024-06-04
Site Editor: Remove unused code

Mamaduka:
## What?
Related #52632 
This is a follow-up to https://github.com/WordPress/gutenberg/pull/62241#discussion_r1624530969.

Remove unused code and components in the `edit-site` package. After unification, there's no need to maintain a duplicate list of inserter media categories.

## Why?
* The `block-editor/inserter-media-categories.js` hasn't been used since #56000.
* The `block-editor/block-inspector-button.js` hasn't been used since #46709.

P.S. @scruffian created an issue to restore `BlockInspectorButton` in editors—#56343. I think it should be restored as part of the `editor` package.

## Testing Instructions
1. Open the Site Editor.
2. Smoke test the Media categories.

Same.


youknowriad (approved):
Good catch 👍

PR #62289, opened 2024-06-04
Add markdown-style `em` and `strong` format support

### LINKED ISSUE #62243: Add support for "emphasis" markdown shortcuts

Right now Gutenberg has mixed support for markdown-style text shortcuts. You can use shortcuts to generate heading blocks (`#`), lists (`-`), and the horizontal rule (`---`). Using the ` (backtick) character will create an inline code tag for you. But there are a few others that could be helpful while writing.

Similar to the way Gutenberg recognizes the backtick character and converts it to code, I would like to work on extending that support to emphasis as the scope of this issue. 

The ideal result here would be that users could type a single or double asterisk `*` or underscore `_` to get `<em>` or `<strong>` tags inline.

https://daringfireball.net/projects/markdown/syntax#em

bacoords:
## What?

In the editor, if a user types `*` or `_` before and after a word/phrase, the text is then wrapped in `<em>` tags. If a user types `**` or `__` before and after a word/phrase, the text is then wrapped in `<strong>` tags. 

This follows the markdown format, similar to backticks being used for `<code>`, hash characters for headings, lists, horizontal rule, etc.

## Why?

There is inconsistent support for [Markdown-style shortcuts](https://daringfireball.net/projects/markdown/syntax#em) currently.

Closes #62243

## How?

Updates the format-library package to include support for italic and bold shortcuts.

## Testing Instructions

1. Open a post or page.
2. Insert a paragraph block.
3. Start typing and try starting/ending a phrase with `*`, `_` to see italics and `**`, `__` to see bold.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/6867360/e29c49b0-209a-4546-9239-e02ec887c1e7

### DIFF
--- a/packages/format-library/src/bold/index.js
+++ b/packages/format-library/src/bold/index.js
@@ -2,7 +2,7 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { toggleFormat } from '@wordpress/rich-text';
+import { toggleFormat, remove, applyFormat } from '@wordpress/rich-text';
 import {
 	RichTextToolbarButton,
 	RichTextShortcut,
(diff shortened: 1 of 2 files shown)


carolinan:
How do users enter actual *text and **text in their content without enforcing the formatting? 
(Not everyone will know about or will want to use the footnotes)

ellatrix:
I've considered this before and we were worries about false positives, but maybe that worry is questionable. Input rules can be undone with cmd+z, Backspace, and Escape I believe.

Mamaduka:
Yes, the input rules can be undone via CMD+Z and Backspace.

carolinan:
That assumes the users knows about it.

bacoords:
I would just note that the convention of transforming text based on special characters already exists. If a user wants to start a paragraph with a hash `#` or a hypen `-` they have this same issue - the block is transformed into a heading or list and they have to learn to use backspace or cmd-z to reverse it. The same for using backticks inline (which can only be reversed with cmd-z, not backspace, I believe). 

This would behave similar to the backtick character, in that if you have more than one in a paragraph, it will automatically transform the text in between, but you can undo it with cmd-z. Also it will not automatically transform any text that is copy/pasted or added out of order.

ryanwelcher:
This seems like a really good addition to the editing experience. I agree with @bacoords that the users are already familiar with how to undo formatting in the editor. @Mamaduka is there anything I can help with to move this forward?

PR #62290, opened 2024-06-04
Fix template footer click to focus canvas

jeryj:
## What?

Clicking the Template button in the footer breadcrumbs causes a focus loss to the body of the page. This PR sets focus to the editor canvas instead, making it similar to how the other buttons in footer breadcrumbs work.

## Why?

a11y. Focus management.

## How?

When clicking. "Template" in the footer breadcrumbs, set focus on the editor canvas.

## Testing Instructions

- Select a block
- Click the root "Template" button in the footer
- Focus should be visible on the editor canvas

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/967608/dcb0e4cb-2352-4ee0-aa40-99d04ce5af08

### DIFF
--- a/packages/block-editor/src/components/block-breadcrumb/index.js
+++ b/packages/block-editor/src/components/block-breadcrumb/index.js
@@ -12,6 +12,7 @@ import { chevronRightSmall, Icon } from '@wordpress/icons';
 import BlockTitle from '../block-title';
 import { store as blockEditorStore } from '../../store';
 import { unlock } from '../../lock-unlock';
+import { __unstableUseBlockRef as useBlockRef } from '../block-list/use-block-props/use-block-refs';
 
 /**
  * Block breadcrumb component, displaying the hierarchy of the current block selection as a breadcrumb.
@@ -37,6 +38,10 @@ function BlockBreadcrumb( { rootLabelText } ) {
 	}, [] );
 	const rootLabel = rootLabelText || __( 'Document' );
 
+	// We don't care about this specific ref, but this is a way
+	// to get a ref within the editor canvas so we can focus it later.
+	const blockRef = useBlockRef( clientId );
+
 	/*
 	 * Disable reason: The `list` ARIA role is redundant but
 	 * Safari+VoiceOver won't announce the list otherwise.
@@ -60,7 +65,11 @@ function BlockBreadcrumb( { rootLabelText } ) {
 					<Button
 						className="block-editor-block-breadcrumb__button"
 						variant="tertiary"
-						onClick={ clearSelectedBlock }
+						onClick={ () => {
+							clearSelectedBlock();
+							// Focus the editor iframe.
+							blockRef.current?.closest( 'body' )?.focus();
+						} }
 					>
 						{ rootLabel }
 					</Button>
--- a/packages/block-editor/src/components/block-breadcrumb/index.js
+++ b/packages/block-editor/src/components/block-breadcrumb/index.js
@@ -68,7 +68,9 @@ function BlockBreadcrumb( { rootLabelText } ) {
 						onClick={ () => {
 							clearSelectedBlock();
 							// Focus the editor iframe.
-							blockRef.current?.closest( 'body' )?.focus();
+							blockRef.current
+								?.closest( '[aria-label="Block canvas"]' )
+								?.focus();
 						} }
 					>
 						{ rootLabel }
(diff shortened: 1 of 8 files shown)


packages/block-editor/src/components/block-breadcrumb/index.js
```diff
@@ -60,7 +65,11 @@ function BlockBreadcrumb( { rootLabelText } ) {
 					<Button
 						className="block-editor-block-breadcrumb__button"
 						variant="tertiary"
-						onClick={ clearSelectedBlock }
+						onClick={ () => {
+							clearSelectedBlock();
+							// Focus the editor iframe.
+							blockRef.current?.closest( 'body' )?.focus();
```
jeryj: @ellatrix Is there a better way to get the editor canvas ref?

andrewserong: Would it be worth trying the `.editor-styles-wrapper` classname, since that gets attached to the `body` element?

andrewserong (commented):
It looks like attempting to focus `body` only works in the iframed editor, and not if the editor doesn't use an iframe. For example, in the post editor if custom fields are enabled in preferences.

https://github.com/WordPress/gutenberg/assets/14988353/88c896b1-a0ca-4ed5-af00-f90dbb427dff

In the above video, this PR seems to work nicely when iframed, but then when switching the custom fields on, there's still a focus loss when clicking on "Post". Is that a blocker? Now that I think of it, is there a tab stop for the canvas in the non-iframed editor 🤔

jeryj:
Thanks for the review, @andrewserong! Great catch on the non-iframed version too. I updated the focus styles to work for both versions of the editor.

andrewserong (commented):
Thanks for the update, this is feeling closer to me. I still noticed a couple of oddities in the non-iframed editor while running TT4 theme (I'm testing in Chrome on a Mac in case that's also part of it).

* When in fullscreen mode in the non-iframed editor, the outline is only visible to the top and to the right
* When not in fullscreen mode, the outline appears to be displaying inconsistently for me. Not sure if it's hidden behind other UI elements (e.g. if you scroll you can kind of see part of the outline disappearing beneath the custom fields)? I wondered if the non-iframed version might work better with absolute positioning instead of fixed 🤔

| Non-iframed fullscreen | Non-iframed not fullscreen |
| --- | --- |
| image | image |

Another thing I noticed while testing is that the outline is also visible for me when clicking on the editor canvas to deselect a block. Is that intentional? I wasn't sure if the scope of this PR was just about when clicking on the breadcrumbs in the interface footer, or if it should also apply with this interaction. It's a fairly prominent change, so just thought I'd raise it:

https://github.com/WordPress/gutenberg/assets/14988353/6300a1af-bf16-4986-a680-a98bbdd40018

jeryj:
> Another thing I noticed while testing is that the outline is also visible for me when clicking on the editor canvas to deselect a block. Is that intentional? 

It's not necessarily intentional, but the expected behavior since it's placing focus on the body element of the canvas. It focusable on trunk as well, but doesn't show you that focus is placed there.

I think these explorations and testing are really helpful. At this point, I think we should solve where focus goes on https://github.com/WordPress/gutenberg/pull/62196, as this one should get the same treatment.

packages/block-editor/src/components/block-breadcrumb/index.js
ellatrix: Not really. But maybe, when selection is cleared, focus should automatically be set to the writing flow element or something (by the writing flow component, or iframe component maybe)

ellatrix:

packages/block-editor/src/components/block-breadcrumb/index.js
ellatrix: See also https://github.com/WordPress/gutenberg/pull/62196/files#r1630727302

packages/block-editor/src/components/block-breadcrumb/index.js
```diff
@@ -60,7 +65,13 @@ function BlockBreadcrumb( { rootLabelText } ) {
 					<Button
 						className="block-editor-block-breadcrumb__button"
 						variant="tertiary"
-						onClick={ clearSelectedBlock }
+						onClick={ () => {
+							clearSelectedBlock();
+							// Focus the editor iframe.
+							blockRef.current
+								?.closest( '[aria-label="Block canvas"]' )
```
kevin940726: Any reason why we don't just create a ref for `FocusableWrapper` and pass it down via context?

jeryj: That's a good idea -- I think we're going to try to focus the editor region instead though.

scruffian (commented):
In my testing the focus isn't visible unless I use my keyboard to select the breadcrumb items. Not sure if that's deliberate?

packages/block-editor/src/components/block-breadcrumb/index.js
scruffian: I added a ref to do this. The downside is that the ref needs to be passed through props.

jeryj:
Closing in favor of https://github.com/WordPress/gutenberg/pull/62595

PR #62292, opened 2024-06-04
fix move `CONTENT_ONLY_BLOCKS` into component body to ensure the `editor.postContentBlockTypes` filter gets called whenever the values are used

fabiankaegy:
## What?

Moves the `applyFilters( 'editor.postContentBlockTypes' )` call into the function body of the React component where it is used.

## Why?

Without this change the filter only gets run once at the time the `wp-editor` JS loads in the browser. This makes it almost impossible to call hook into this filter because most scripts that work with blocks have `wp-editor` as a dependency.

## How?

Moving the constant with the `applyFilters` call into the function body of the component.


noisysocks (commented):
Makes sense. Thanks.

A bit nitpicky but can we keep the unfiltered `CONTENT_ONLY_BLOCKS` constant at the top level and then do this in the component?

```js
const contentOnlyBlocks = applyFilters( 'editor.postContentBlockTypes', CONTENT_ONLY_BLOCKS );
```

That way the capitalisation is properly indicating that the value never changes.

fabiankaegy:
@noisysocks thanks for the review :) I updated the PR as you suggested 👍

noisysocks (approved):
Groovy, thanks!

ellatrix:

PR #62293, opened 2024-06-04
Interactivity API: Increase directive `wp-each-child` priority.

cbravobernal:
## What?

Co-authored with @DAreRodz, bug found by @zaguiini
Increase `data-wp-each` priority to fix trying to process non existent elements in the dom.

## Why?

To add some context, the `data-wp-each-child` directive is intended to remove the node in which it appears. Why? Because `data-wp-each` is the one that actually renders it.

The issue is that it has the same priority level as most directives, and they are grouped into the same `Directives` component and evaluated together. Thus, even though the component returns `null`, all its hooks are evaluated, i.e., the ones that `data-wp-bind`, `data-wp-on` etc., register. Also, the `data-wp-each` server-side processing maintains the directives that appear in child nodes. 😬

That's why the element is `null` inside the `useInit` hook called in `data-wp-bind`; because it _is_. The thing is that `data-wp-bind` should not be evaluated (nor other directives).

## How?

One quick fix for this issue is to increase the `data-wp-each-child` priority, being the first one in being evaluated.

Other fix could be not to server side render the directives in the content created by the template tag, in the `data-wp-each` SSR processing. I did not measure it, but being that content already evaluated, adding the directives in the SSR should be triggering a not needed re-evaluation.

## Testing Instructions

1) Create an interactive block with `wp-each` SSR, like this one:
 ```php
<?php
$context = array('list' => ['one', 'two', 'three']);
?>
<div
	data-wp-interactive="counter-block"
    <?php echo wp_interactivity_data_wp_context($context) ; ?> 
>
    <template data-wp-each="context.list" data-wp-each-key="context.item">
        <div data-wp-bind--id="context.item">
            <span data-wp-text="context.item"></span>
        </div>
    </template>
</div>
```

2) Check that in `trunk`, an error appears on the console.
3) Apply this PR, reload, check that there are no errors.

## Screenshots or screencast 
PR not applied
Screenshot 2024-06-04 at 18 48 02

PR applied
Screenshot 2024-06-04 at 18 48 42

Leaving the PR on draft until we decide which approach is better.

### DIFF
--- a/packages/interactivity/src/directives.tsx
+++ b/packages/interactivity/src/directives.tsx
@@ -632,5 +632,5 @@ export default () => {
 		{ priority: 20 }
 	);
 
-	directive( 'each-child', () => null );
+	directive( 'each-child', () => null, { priority: 4 } );
 };
--- a/packages/interactivity/src/directives.tsx
+++ b/packages/interactivity/src/directives.tsx
@@ -632,5 +632,5 @@ export default () => {
 		{ priority: 20 }
 	);
 
-	directive( 'each-child', () => null, { priority: 4 } );
+	directive( 'each-child', () => null, { priority: 1 } );
 };
--- a/packages/e2e-tests/plugins/interactive-blocks/directive-each/render.php
+++ b/packages/e2e-tests/plugins/interactive-blocks/directive-each/render.php
@@ -235,13 +235,28 @@
 	<p data-testid="item" data-wp-each-child>delta</p>
 </div>
 
+<hr>
+
 <div
 	data-wp-interactive="directive-each"
-	data-wp-context='{ "list": [ "beta"] }'
+	data-wp-context='{ "list": [ "beta" ], "callbackRunCount": 0 }'
 	data-testid="elements with directives"
 >
 	<template data-wp-each="context.list">
-		<div data-wp-text="context.item" data-testid="item" data-wp-run="callbacks.shouldRun"></div>
+		<div
+			data-testid="item"
+			data-wp-text="context.item"
+			data-wp-priority-2-init="callbacks.updateCallbackRunCount"
+		></div>
 	</template>
-	<div data-testid="item" data-wp-each-child data-wp-run="callbacks.shouldNotRun"></div>
+	<div
+		data-wp-each-child
+		data-testid="item"
+		data-wp-text="context.item"
+		data-wp-priority-2-init="callbacks.updateCallbackRunCount"
+	></div>
+	<data
+		data-testid="callbackRunCount"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by DAreRodz** [comment]
That's a nice and easy fix. 😄

Not sure if we should add tests to this, as the directive priority levels are already tested in `directive-priorities.spec.ts`. Maybe we could add a test case in `directive-each.spec.ts`?

> Other fix could be not to server side render the directives in the content created by the template tag, in the `data-wp-each` SSR processing. I did not measure it, but being that content already evaluated, adding the directives in the SSR should be triggering a not needed re-evaluation.

Regarding this approach (I suggested it 😅), I think it would be unnecessarily complex and wouldn't solve the problem in the browser if, for some reason, the HTML is _broken_ (I mean, HTML that contains an element with `data-wp-each-child` and other directives).

File: packages/interactivity/src/directives.tsx [hunk 1]
```diff
@@ -632,5 +632,5 @@ export default () => {
 		{ priority: 20 }
 	);
 
-	directive( 'each-child', () => null );
+	directive( 'each-child', () => null, { priority: 4 } );
```
luisherranz: I think it's a good idea to increase the priority to avoid the creation of inner Directive components. 

But why use 4? Shouldn't it be the minimum possible, 1?

luisherranz (dismissed):
I don't understand how this was working. Did we have this bug from the beginning? Is this in WP 6.5?

We should at least add one test, by the way 🙂

packages/interactivity/src/directives.tsx
cbravobernal: 5 is the minimum used right now. I added 4 in case there is a future directive that would need more priority.

luisherranz: I don't think we need to give more priority to another directive that is going to process something that is just going to be deleted 😄

I was thinking more about when custom directives exist. 4 works well for now, but I think it's better to use 1 anyway.

vcanales:
Checking in on PRs with the Backport to Gutenberg RC. I'll release 18.5 later today, do you think this'll make it in time?

cbravobernal:
> Checking in on PRs with the Backport to Gutenberg RC. I'll release 18.5 later today, do you think this'll make it in time?

The e2e test is a little tricky. I'm not sure we will make it.

DAreRodz (approved):
I refactored the test a bit and ensured it only passed with the fix. LGTM!

vcanales:
I just cherry-picked this PR to the release/18.5 branch to get it included in the next release: af2e67ad28

ellatrix:

PR #62298, opened 2024-06-04
Post Date & Comment Date: Add relative date format

### LINKED ISSUE #62259: Post Date: Add relative date format
Would be good if you could set Post Date to show a relative date. This is common amongst non-block themes.

For example: _12 hours ago_, _2 days ago_, _a month ago_.

Screenshot 2024-06-04 at 14 29 56

WordPress has [`human_time_diff`](https://developer.wordpress.org/reference/functions/human_time_diff/) which can be used.

amitraj2203:
## What?
Closes: #62259 

## Why?
This PR addresses the need for the Post Date block to display the date in a relative format.

## How?
This PR adds a ~`Time Ago`~ `human-diff` date format option to display relative times (e.g., "2 days ago"), enhancing the flexibility of date formatting in the editor.

## Testing Instructions
1. Edit a Post/Page.
2. Add a Date block.
3. See the last option in `CHOOSE A FORMAT`

## Screenshots or screencast 
image

## ✍️ Dev note: New 'human-diff' Date Format Option

A new date format option has been introduced for the `Post Date` and `Comment Date` blocks, allowing users to display dates in a human-readable, relative format.

The `human-diff` format presents dates in a more intuitive way, such as `12 hours ago`, `2 days ago`, or `a month ago`. This enhancement provides a more user-friendly way to display date information, which can improve readability and engagement for readers.

To use this new format, users can select the option that displays an example of the relative time format (e.g., "2 years ago") from the `Choose a format` dropdown in the block settings panel for both the Post Date and Comment Date blocks. The feature utilizes WordPress's built-in `human_time_diff()` function to calculate and display the relative time difference.

To access this option:
1. Insert a Post Date or Comment Date block in your post or page.
2. In the block's settings panel, locate the `Choose a format` option.
3. Click on the dropdown to reveal various date format options.
4. Select the option that shows an example of the relative time format (e.g., "2 years ago").

### DIFF
--- a/packages/block-editor/src/components/date-format-picker/index.js
+++ b/packages/block-editor/src/components/date-format-picker/index.js
@@ -92,20 +92,22 @@ function NonDefaultControls( { format, onChange } ) {
 			_x( 'F j, Y', 'long date format' ),
 			/* translators: See https://www.php.net/manual/datetime.format.php */
 			_x( 'M j', 'short date format without the year' ),
-			__( 'Time Ago' ),
 		] ),
 	];
 
-	const suggestedOptions = suggestedFormats.map(
-		( suggestedFormat, index ) => ( {
+	const suggestedOptions = [
+		...suggestedFormats.map( ( suggestedFormat, index ) => ( {
 			key: `suggested-${ index }`,
-			name:
-				suggestedFormat === 'Time Ago'
-					? humanTimeDiff( EXAMPLE_DATE, new Date() )
-					: dateI18n( suggestedFormat, EXAMPLE_DATE ),
+			name: dateI18n( suggestedFormat, EXAMPLE_DATE ),
 			format: suggestedFormat,
-		} )
-	);
+		} ) ),
+		{
+			key: 'human-diff',
+			name: humanTimeDiff( EXAMPLE_DATE ),
+			format: 'human-diff',
+		},
+	];
+
 	const customOption = {
 		key: 'custom',
 		name: __( 'Custom' ),
@@ -115,7 +117,9 @@ function NonDefaultControls( { format, onChange } ) {
 	};
 
 	const [ isCustom, setIsCustom ] = useState(
-		() => !! format && ! suggestedFormats.includes( format )
+		() =>
+			!! format &&
+			! suggestedOptions.some( ( option ) => option.format === format )
 	);
 
 	return (
--- a/packages/block-editor/src/components/date-format-picker/index.js
+++ b/packages/block-editor/src/components/date-format-picker/index.js
@@ -13,11 +13,15 @@ import {
 	__experimentalVStack as VStack,
 } from '@wordpress/components';
 
-// So that we can illustrate the different formats in the dropdown properly,
-// show a date that has a day greater than 12 and a month with more than three
-// letters. Here we're using 2022-01-25 which is when WordPress 5.9 was
-// released.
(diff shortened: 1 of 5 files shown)


noisysocks:
Nice, thanks for this @amitraj2203!

I don't think it's a good idea to do this by storing `'Time Ago'` in the `format` attribute. Technically `'Time Ago'` is a valid [format string](https://www.php.net/manual/en/datetime.format.php):

```php
<?php
echo date('Time Ago'); // prints: UTC4806UTC AM22024
```

It's unlikely a user would actually want to use that format string but nonetheless I think it's best for us to not mix this data.

One idea is to add a new boolean attribute `isRelative` which when true means that the `format` is optional and the outputted date will be a relative string e.g. _2 months ago_. What do you think of that approach?

amitraj2203:
Hi @noisysocks,
Thanks for the suggestion. Yes, we can proceed with that. Just to confirm, I'll be using a ToggleControl for the new boolean attribute, correct?
For example,

https://github.com/WordPress/gutenberg/assets/77401999/225cfafd-6fb6-457c-baf9-28083a4709d0

noisysocks:
No I think we should still keep the dropdown as that's a better UI. Just changing the way that those settings are stored. Let me know if you need help.

amitraj2203:
Hi @noisysocks! 
So, if I understand correctly, when the user selects the `Time Ago` option from the dropdown, we should set the `isRelative` attribute to true and leave the `format` attribute as default or empty?

Also, the changes in the [DateFormatPicker](https://github.com/WordPress/gutenberg/pull/62298/files#diff-a81f243f8ac38c659d64de0894c81e2ea49caa131c13b3807c1ed74367c81326) component are good to go or is there anything else we need to adjust there?
I would like to change `Time Ago` to `Relative Time` as it's making more sense.

noisysocks:
Sorry @amitraj2203, I've changed my mind 😅

I think your approach of using the `format` attribute is right. Let's simply update the `format` string that denotes a relative date to be `"human-diff"`.

My change of heart comes from reading [this post](https://make.wordpress.org/core/2024/06/08/proposal-bits-as-dynamic-tokens/) by @dmsnell which uses `format="human-diff"` in a proposed Post Date _bit_ and that looks pretty good to me.

noisysocks (approved):
Thanks so much for this PR @amitraj2203.

I hope you don't mind I pushed some changes: it now uses `'human-diff'` instead of `'Time Ago'` as the `format` value that indicates a relative time, it works in the Comment Date block now as well, and it now ensures that the example date shown is a relatively recent date. I also fixed some minor internationalisation problems.

NikolaVasic12:
Hi @amitraj2203,
While solving the problem of translating  "ago", I came across your commit. Can you tell me whether this will solve the problem of translating the word "ago" and whether this change will be included in the next WP version?

Thank you

fabiankaegy:
Hey @amitraj2203  👋 

Thanks in advance :)

amitraj2203:
Hi @fabiankaegy, sorry for the delayed response. I have updated this [PR's description](https://github.com/WordPress/gutenberg/pull/62298#:~:text=%E2%9C%8D%EF%B8%8F%20Dev%20note%3A%20New%20%27human%2Ddiff%27%20Date%20Format%20Option) with the dev note. Please take a look at it and let me know if any changes need to be made.

Thanks :)

PR #62300, opened 2024-06-04
Block Inserter Category and Output Heuristics

jeryj:
Some polish for the new inserter category output:

- Only output one "most used" categories list. If there is a selected block, this goes beneath the allowed selected blocks.
- Adds current block name as the category title for its allowed blocks

- **If no block is selected**/default empty block, output all categories and blocks as it was in past versions of WordPress.
- **If a block is selected**, filter the available blocks.
  - **If more than 80% of the blocks are allowed for this block**, don't bother filtering. Output all items. This is to allow the "Most Used" blocks to be at the top without repeating it further down, and not bother filtering when most every block is allowed.
  - **If there are less than 80% of the block items available** for the selected block, output the filtered items first.
    - **If there is only one block category (ie "Text"), replace the category title with the block title.** The List Block is an example of this. This avoids situations where "Text" is followed by another "Text" category.

## What?

Improves the UX for the block inserter category output.

## Why?

Polish the UX, as there are some areas that feel off, such as:
- showing two Most Used categories. Most used seems useful when it's relative to all blocks, not a filtered list of a small number of blocks.
- Don't output multiple levels of the same heading next to each other. For example, Text followed by another Text category. 
- When there are many blocks available (like the Group block), it outputs nearly all blocks with one available beyond it. This filtering seems irrelevant at that point and feels off that there's one standalone category at the end.

## How?

Adds `selectedBlockItems` and `selectedBlockCategories` props added to the `<BlockTypesTabPanel />` component and only outputs this component once.

#### Default State

- Select the empty default block
- Open inserter
- Categories should show as normal.

#### Selected block with One Category Item

- Select a list block
- Open inserter
- First category should be "List"
- List Item should be the first block item
- All other items should show after this, including Most Used if the option is on.

#### Selected block with Multiple Categories

- Select a navigation link block
- Open inserter
- Navigation Block allowed blocks should show first, within default category titles
- All other items should show after this, including Most Used if the option is on.

#### Selected Block with almost all available blocks

- Select the group block
- Open inserter
- Categories should show as normal (no filtering of allowed blocks)

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/967608/57bcc2ae-952d-41b9-a50f-b64e0568b438

### DIFF
--- a/packages/block-editor/src/components/inserter/block-types-tab.js
+++ b/packages/block-editor/src/components/inserter/block-types-tab.js
@@ -21,6 +21,9 @@ const getBlockNamespace = ( item ) => item.name.split( '/' )[ 0 ];
 
 const MAX_SUGGESTED_ITEMS = 6;
 
+const INSERTABLE_BLOCKS_AT_SELECTION_CATEGORY =
+	'insertable-blocks-at-selection';
+
 /**
  * Shared reference to an empty array for cases where it is important to avoid
  * returning a new array reference on every invocation and rerendering the component.
@@ -121,6 +124,10 @@ export function BlockTypesTabPanel( {
 							onHover={ onHover }
 							label={ category.title }
 						/>
+						{ category.slug ===
+							INSERTABLE_BLOCKS_AT_SELECTION_CATEGORY && (
+							<div className="block-editor-inserter__category-panel-divider" />
+						) }
 					</InserterPanel>
 				);
 			} ) }
@@ -181,16 +188,16 @@ export function BlockTypesTab(
 		return <InserterNoResults />;
 	}
 
-	const currentlySelectedBlockType = getBlockType(
-		getBlockName( rootClientId )
-	);
-
 	let allCategories = [];
 
 	if ( rootClientId && categories.length ) {
+		const currentlySelectedBlockType = getBlockType(
+			getBlockName( rootClientId )
+		);
+
 		allCategories = [
 			{
-				slug: currentlySelectedBlockType.name,
+				slug: INSERTABLE_BLOCKS_AT_SELECTION_CATEGORY,
 				title: currentlySelectedBlockType.title,
 				icon: null,
 			},
@@ -211,7 +218,7 @@ export function BlockTypesTab(
 		if ( rootClientId && item.rootClientId === rootClientId ) {
 			allItems.push( {
 				...item,
-				category: currentlySelectedBlockType.name,
+				category: INSERTABLE_BLOCKS_AT_SELECTION_CATEGORY,
 			} );
 		} else {
 			allItems.push( { ...item } );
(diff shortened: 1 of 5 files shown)


jeryj:
@ellatrix Curious on your feedback for if this code is an improvement performance wise or if I'm over-thinking it.
@richtabor Do you think grouping all allowed blocks for a block within one category (block name title) is OK, or will some blocks show too many ungrouped blocks beneath them (such as the navigation block).

ellatrix (changes requested):
I feel like I talked about this a few times before: if you create a single "insertable at selection" category, all the other categories disappear. So if you open the inserter in a quote or group, they'll all be squashed into a single one.

image

ellatrix:
To me this is not a "code quality" PR, there's a change in behaviour.

ellatrix:
> Curious on your feedback for if this code is an improvement performance wise or if I'm over-thinking it.

Curious why you think `BlockTypesTabPanel` is a performance issue? It doesn't do anything unperformant?

jeryj:
> I feel like I talked about this a few times before: if you create a single "insertable at selection" category, all the other categories disappear. So if you open the inserter in a quote or group, they'll all be squashed into a single one.

Yes, I noticed that too and mentioned it above. I was thinking about redefining the heuristics so if it has a lot of blocks within it, such as the group block, then it outputs things within the normal categories. Trying to get a feel for different options.

> To me this is not a "code quality" PR, there's a change in behaviour.

I wasn't sure what to categorize it as since it's not a bug and it may not be an enhancement in its current state :) 

> Curious why you think BlockTypesTabPanel is a performance issue? It doesn't do anything unperformant?

It's doing a lot of things for the initial category that aren't necessary, like grouping into collections, outputting most used blocks. I think it could be simpler to pass one list of items and categories, so I'm trying it out. I'll probably drop close this PR, but I was curious to see how it was in practice.

jeryj:
@WordPress/gutenberg-design - Can I get feedback on the UX of the inserter category heuristics here? I think this is an improvement to the output and organization of the block inserter categories when we have block filtering relative to the selected block.

paaljoachim:
If I understand this correctly.
That Most used in the Block inserter will change depending on selected block in Canvas view. I have mixed feelings about this as for many users it can be fairly random which block is selected in Canvas view. Other times it can be helpful to select a specific block in Canvas view and have a more focused/tailored block inserter experience.

An example would be having a specific WooCommerce block selected in Canvas view and then seeing other WooCommerce blocks on the top of the Block inserter to give a signal to the user a relation of the selected block to other similar blocks in the Block inserter.

It can be very helpful to create a relation between a selected block in Canvas view <-> seeing more focused blocks in the Block inserter. Gradually over time an user can notice the relation. Then might on purpose select specific blocks in Canvas view to get related blocks in the Block inserter.

jasmussen:
Thanks for the PR. While the behavior in trunk isn't necessarily the most intuitively understandable, there's something obvious about the sorting in categories of _inserts in selected block_ and _inserts after_, but the behavior for the Group block, feels like it bridges those two gaps in an intuitive way. 

Reducing duplication, and showing towards the top the most useful and common blocks you'd likely insert contextually, that feels valid enough. But it's not clear this brings benefits to the actual user experience, especially for the group block. I could very well be missing nuance, however, the trunk behavior I haven't had a chance to absorb fully yet.

jeryj:
Closing for now, but we can open this exploration up later if needed.

PR #62301, opened 2024-06-04
Fix unintended overwrite of eslint `no-restricted-syntax`

mirka:
## What?

Fixes the `eslintrc` file so that the `no-restricted-syntax` rules are applied as intended.

## Why?

Any settings for the same rule (`no-restricted-syntax` in this case) will be completely overwritten by subsequent `overrides` settings, not merged. I had unconsciously been treating it like they would be merged 🙈  

## How?

Extracted out the restricted syntax rules to be more composable. (This is similar to what had already been done with `restrictedImports`.)

## Testing Instructions

✅ The "global" restricted syntax rules at the root level should also trigger on the following files, as originally intended:

```js
	files: [
		'packages/*/src/**/*.[tj]s?(x)',
		'storybook/stories/**/*.[tj]s?(x)',
	],
	excludedFiles: [ '**/*.native.js' ],
```

```js
	files: [ 'packages/components/src/**' ],
	excludedFiles: [
		'packages/components/src/utils/colors-values.js',
		'packages/components/src/theme/**',
	],
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -397,7 +397,6 @@ module.exports = {
 				'no-restricted-syntax': [
 					'error',
 					...restrictedSyntax,
-					...restrictedSyntaxComponents,
 					{
 						selector:
 							':matches(Literal[value=/--wp-admin-theme-/],TemplateElement[value.cooked=/--wp-admin-theme-/])',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -83,6 +83,72 @@ const restrictedImports = [
 	},
 ];
 
+const restrictedSyntax = [
+	// NOTE: We can't include the forward slash in our regex or
+	// we'll get a `SyntaxError` (Invalid regular expression: \ at end of pattern)
+	// here. That's why we use \\u002F in the regexes below.
+	{
+		selector:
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -83,6 +83,72 @@ const restrictedImports = [
 	},
 ];
 
+const restrictedSyntax = [
```
mirka, line 86: This chunk of code was just extracted. No changes.

.eslintrc.js
```diff
@@ -390,6 +396,7 @@ module.exports = {
 			rules: {
 				'no-restricted-syntax': [
 					'error',
+					...restrictedSyntax,
```
mirka, line 399: We also need to add `...restrictedSyntaxComponents` here, but since it will catch a bunch of previously uncaught violations, I'll address this in a separate PR.

packages/components/src/checkbox-control/stories/index.story.tsx
mirka, line 1: Previously uncaught violation due to the configuration error.

DaniGuardiola (approved):
lgtm!

(edit: nvm I got things mixed up, good to go)

fullofcaffeine:
LGTM!

fullofcaffeine (approved):
Looks good 👍🏻

PR #60447, opened 2024-04-03
Enable template preview in post editor for non administrators

fabiankaegy:
## What?

Closing #60316. Supersedes #58301 and follows https://github.com/WordPress/gutenberg/pull/60317 & https://github.com/WordPress/gutenberg/pull/60326

## Why?

To allow non administrators to preview templates whilst editing content in the regular post editor

## How?

Adjusting the conditions when to display templates to allow users that can only view templates to still preview them.

## Testing Instructions

1. Make sure your theme supports at least 1 additional custom template besides the default single post template.
2. Open the post editor with a user account that is an Editor or lower role
3. See that you are again able to switch between the available templates

PR #60317, opened 2024-03-29
Allow view access of template rest endpoint to anyone with the `edit_post` capability

fabiankaegy:
## What?

Part of #60316 as requested in https://github.com/WordPress/gutenberg/pull/58301#issuecomment-2012136167

Modify the template / template part rest endpoint to allow any user role with the `edit_post` capability to view entities. 

## Why?

In order to render the block template in the locked template preview inside the post editor we need to be able to fetch the contents of any block templates / template parts for any user role that can edit a post. 

## How?

Overwriting the `get_items_permissions_check` and `get_item_permissions_check` methods of the `rest_controller_class` of the `wp_template` post type to check whether the current user can `edit_posts`.

## Testing Instructions

Try making an authenticated rest request to the template rest endpoint with any user role that has the `edit_post` capability. -> The template should get returned 
Try making an unauthenticated request and still get an unauthorized error

PR #62304, opened 2024-06-04
Feature: Set editor rendering mode by post type

### LINKED ISSUE #58038: Add an option to change default renderingMode of Post Editor

TylerB24890:
## What?

**Related Issue & PR:** 
- https://github.com/WordPress/gutenberg/issues/58038
- https://github.com/WordPress/gutenberg/pull/61844

&nbsp;
**Track Ticket:** https://core.trac.wordpress.org/ticket/61811
**Backport PR:** https://github.com/WordPress/wordpress-develop/pull/7129

This PR adds a filterable `default_rendering_mode` property to the `WP_Post_Type` object allowing users to define the default rendering mode for the editor for that post type.

The `default_rendering_mode` property can be added to the arguments when registering the post type via `register_post_type()` and can be overwritten using the available filters:
- `post_type_{$post_type}_default_rendering_mode`: To target an individual specific post type.
```php
/**
 * Filters the block editor rendering for a specific post type.
 *
 * The dynamic portion of the hook name, `$post_type->name`, refers to the post type slug.
 *
 * @param string       $rendering_mode The current rendering mode set with the post type registration.
 * @param WP_Post_Type $post_type      The current post type object.
 */
$rendering_mode = apply_filters( "post_type_{$post_type->name}_default_rendering_mode", $post_type->default_rendering_mode, $post_type );
```

- `post_type_default_rendering_mode`: To target all (or multiple) post types.
```php
/**
 * Filters the block editor rendering for a post type.
 *
 * @param string       $rendering_mode  The current rendering mode set with the post type registration.
 * @param WP_Post_Type $post_type_name  The current post type object.
 */
$rendering_mode = apply_filters( 'post_type_default_rendering_mode', $post_type->default_rendering_mode, $post_type );
```

## Why?

Currently there is no way to set the default rendering mode of the block editor. You can select the mode while _in_ the block editor, but upon refreshing that mode is reset back to the default `post-only`. With this update developers have more control over the editing experience and provides a means of setting the default view for the block editor.

## How?

The [linked PR](https://github.com/WordPress/gutenberg/pull/61844) has a discussion that mentions this setting should be applied at the **post type** level, allowing for a difference editing mode per post type. This PR applies the `default_rendering_mode` property to the `WP_Post_Type` object itself and provides multiple ways of overriding or setting the default for a custom (or core) post type.

## Testing Instructions

1. Create a new `post` post type and observe the default editor UI.
2. In your `functions.php` file (or similar) use one of the available filters to set the `default_rendering_mode` property to `template-lock`:
```php
add_filter(
    'post_type_default_rendering_mode',
    function () {
        return 'template-lock';
    }
);
```
3. Refresh the `post` editor and confirm you are now seeing the Template UI instead of the default Post UI.
4. Create a new `page` post and confirm the `page` editor also loads with the Template UI.
5. Update the filter in `functions.php` to target only the `page` post type:
```php
add_filter(
    'post_type_page_default_rendering_mode',
    function () {
        return 'template-lock';
    }
);

```
6. Reload the `page` editor and confirm it still renders the Template UI.
7. Refresh the `post` editor and confirm it now renders the default Post UI.
8. Update the filter in `functions.php` to set the rendering mode for the `post` **and** `page` post types, but no others:
```php
add_filter(
    'post_type_default_rendering_mode',
    function ( $mode, $post_type ) {
        if ( 'page' === $post_type || 'post' === $post_type ) {
            return 'template-lock';
        }

        return $mode;
    },
    10,
    2
)
```
9. Register a custom post type using `register_post_type` and set the `default_rendering_mode` parameter to `template-lock`:
```php
register_post_type(
    'my_custom_type',
    [
        'show_in_rest'           => true,
        'supports'               => [ 'title', 'editor' ],
        'default_rendering_mode' => 'template-lock',
    ]
);
```
10. Visit the editor for your custom post type and confirm it loads with the Template UI.
11. Visit the **Site Editor** and select **Pages** in the left navigation.
12. Select a page and confirm it loads with the Template view (if the code from step 8 is still in place).
13. Change the filter value from step 8 to output `post-only` and refresh the Site Editor.
14. Confirm the Page now loads with the Post Content instead of the template UI.


lib/compat/wordpress-6.6/post.php
```diff
@@ -11,9 +11,28 @@
...
+ * @param array  $args      Array of post type arguments.
+ * @return array Updated array of post type arguments.
+ */
+function gutenberg_post_type_default_rendering_mode( $args ) {
+	if (
+		( isset( $args['show_in_rest'] ) && $args['show_in_rest'] ) &&
+		( isset( $args['supports'] ) && in_array( 'editor', $args['supports'], true ) ) &&
+		( ! isset( $args['rendering_mode'] ) )
+	) {
+		$args['rendering_mode'] = 'post-only';
```
fabiankaegy:
```suggestion
		$args['default_rendering_mode'] = 'post-only';
```

I would suggest renaming the new `rendering_mode` property to `default_rendering_mode`. The reason why is that users are still able to change the rendering mode in the UI. But the word `rendering_mode` implies that it is static.

lib/compat/wordpress-6.6/class-gutenberg-rest-post-types-controller-6-6.php
```diff
@@ -0,0 +1,53 @@
...
+ * @package gutenberg
+ */
+
+/**
+ * Gutenberg_REST_Post_Types_Controller_6_6 class
+ *
+ * Add Block Editor default rendering mode to the post type response
+ * to allow enabling/disabling at the post type level.
+ */
+class Gutenberg_REST_Post_Types_Controller_6_6 extends WP_REST_Post_Types_Controller {
```
fabiankaegy: Given that we are already past the Beta 1 cutoff point for WordPress 6.6 this will need to get moved to a new 6.7 compact folder.

fabiankaegy (commented):
Thanks for you work on this @TylerB24890 🚀 

This is working exactly as I hoped it would. :) I left some minor notes and will request some additional reviews :)

packages/edit-post/src/editor.js
```diff
@@ -31,14 +31,28 @@ function Editor( {
 	initialEdits,
 	...props
 } ) {
+	// Get the rendering mode for the post type.
+	const postTypeRenderingMode = useSelect(
```
youknowriad: Do we want to do the same in the site editor too? We can actually unify this in the "editor" package as well.

Worth-noting though that if we decide to unify, we'll have to make a decision about the "page" post type. Right now it's default rendering is "post-only" in the post editor but "template-locked" in the site editor.

youknowriad (commented):
This is looking very good to me. I'd appreciate broader review as well. @jameskoster @jasmussen @mcsf

lib/compat/wordpress-6.6/post.php
TylerB24890: Thanks, @fabiankaegy ! I've updated this to read `default_rendering_mode`

lib/compat/wordpress-6.6/class-gutenberg-rest-post-types-controller-6-6.php
TylerB24890: Done! Moved to `compat/wordpress-6.7`

lib/compat/wordpress-6.7/post.php
```diff
@@ -0,0 +1,26 @@
...
+ * @param array  $args      Array of post type arguments.
+ * @return array Updated array of post type arguments.
+ */
+function gutenberg_post_type_default_rendering_mode( $args ) {
+	if (
+		( isset( $args['show_in_rest'] ) && $args['show_in_rest'] ) &&
+		( isset( $args['supports'] ) && in_array( 'editor', $args['supports'], true ) ) &&
+		( ! isset( $args['default_rendering_mode'] ) )
+	) {
+		$args['default_rendering_mode'] = 'post-only';
```
ramonjd: I'm not sure where this should take place, but should Gutenberg do a check on any incoming values from the filter hooks?

That is, to make sure it's either `'post-only'` or `'template-locked'` (or any other values)?

The reason I ask is because when I set a non-valid value —

```php
add_filter(
    'post_type_default_rendering_mode',
    function () {
        return 'prisencolinensinainciusol';
    }
);

// AND/OR

add_filter(
    'page_default_rendering_mode',
    function () {
        return 'prisencolinensinainciusol';
    }
);
```

— the template view is shown even though the rendering mode is `'prisencolinensinainciusol'` in state. 

I suspect it's because there are a bunch of `renderingMode !== 'post-only` checks in Gutenberg that assume that the alternative value is always `'template-locked'`.

ramonjd (commented):
This is working great for me for what it's worth. Nice addition.

I also checked what would happen if I logged in as an Author with `template-locked` enabled - it activates the "Show template" option 👍🏻 

Just left a comment about value checking, though even if it's a valid remark, I don't think it should block the first version of this.

jasmussen:
Nice PR. Took it for a quick spin. In a very superficial test—this could use broader opinions—this seems to work as intended. If I change the default post-editor rendering mode to show templates, it does:

showing template

Otherwise it still shows the current default:

not showing template

I think of these defaults as valid user settings to set as well at some point, user settings that are reasonable to persist, so if we add this, it would be good that a future user-setting could also unset whatever a plugin or theme might do. I.e. if my theme sets my page editor to omit the template preview, but I really prefer editing with that, I should be able to override that in my user settings. Make sense?

fabiankaegy:
@jasmussen Agree that there should also be a user setting for this :) We can look into that as follow up  👍

jasmussen:
Not necessarily urgent, IMO, just something to consider for the technical review!

youknowriad:
Any thoughts on making "template-locked" default for pages for both post and site editor @jasmussen ?

jasmussen:
To make sure I read you right, you mean by default show the full template when editing pages, regardless of whether you're editing inside the site editor, or in the separate editor that's accessed from the top level Pages > Add new?

This seems something we most likely want, absolutely. As a 6.7 thing it feels reasonable to test this early.

youknowriad:
Great. Thanks for the confirmation. 

In that case, I think we should move the code here https://github.com/WordPress/gutenberg/pull/62304#discussion_r1627866620 directly to the "editor" package somehow. edit-post and edit-site shouldn't have to pass the default rendering mode anymore.

packages/edit-post/src/editor.js
```diff
@@ -31,14 +31,28 @@ function Editor( {
...
+		( select ) => {
+			const postTypeObject =
+				select( coreStore ).getPostType( initialPostType );
+			if ( postTypeObject && postTypeObject?.default_rendering_mode ) {
+				return postTypeObject.default_rendering_mode;
+			}
+
+			return 'post-only';
```
mcsf:
```suggestion
			return postTypeObject?.default_rendering_mode ?? 'post-only';
```

mcsf (commented):
Nice proposal here, I'm on board too. Quick notes:

* I agree with Riad, let's move this to the `editor` package
* Trying out `template-lock` as the default for pages in 6.7 sounds cool
* Would amending the hook name from `{$post_type->name}_default_rendering_mode` to `post_type_{$post_type->name}_default_rendering_mode` help developers infer and trace it back to its origin in the future? Otherwise I worry that certain post type names could make things confusing or make it sound like it's a third-party hook (e.g. `woo_product_default_rendering_mode`)
* Why the difference in signatures between the two hooks? I see one takes a `WP_Post_Type` instance, the other takes the type's name.

packages/edit-post/src/editor.js
TylerB24890: @mcsf 
>Why the difference in signatures between the two hooks? I see one takes a WP_Post_Type instance, the other takes the type's name.

That's a good question, and I debated that myself when implementing. My initial thought there was that `"{$post_type}_default_rendering_mode"` would be targeted to a specific post type and receive that specific post type object. Where `post_type_default_rendering_mode` would be targeted to _all_ post types, and only the name would be needed to conditionally set a value.

I can see how that could be confusing though, so we can make them consistent. Do you have a preference on which would be better to receive? 

>Would amending the hook name from {$post_type->name}_default_rendering_mode to post_type_{$post_type->name}_default_rendering_mode help developers infer and trace it back to its origin in the future? Otherwise I worry that certain post type names could make things confusing or make it sound like it's a third-party hook (e.g. woo_product_default_rendering_mode)

A good point. Will definitely make this update.

lib/compat/wordpress-6.7/post.php
TylerB24890: I think that makes sense to do, @ramonjd -- What are all the value values here? Is it just `post-only`, `template-only`, `template-lock`?

TylerB24890:
> Great. Thanks for the confirmation.
> 
> In that case, I think we should move the code here [#62304 (comment)](https://github.com/WordPress/gutenberg/pull/62304#discussion_r1627866620) directly to the "editor" package somehow. edit-post and edit-site shouldn't have to pass the default rendering mode anymore.

@youknowriad This has been moved to the `editor` package. Please take a look and let me know if this is the approach you were thinking of. Adding this here allowed me to remove the various `mode` props passed to the site-editor and post-editor, so now it's all defined in the main `editor` package. Confirmed using the filters will set the default for both the site and post editors as expected.

I **have not** applied the `template-lock` mode to the Page post type by default. If that's something we do want to do I can do so.

@mcsf 
>Why the difference in signatures between the two hooks? I see one takes a WP_Post_Type instance, the other takes the type's name.

I've consolidated these to be the same. Both filters will now pass the full `WP_Post_Type` object.

>Would amending the hook name from {$post_type->name}default_rendering_mode to post_type{$post_type->name}_default_rendering_mode help developers infer and trace it back to its origin in the future? Otherwise I worry that certain post type names could make things confusing or make it sound like it's a third-party hook (e.g. woo_product_default_rendering_mode)

This has been updated to `"post_type_${post_type->name}_default_rendering_mode"`

packages/edit-post/src/editor.js
mcsf: > and only the name would be needed to conditionally set a value.

Yeah, I tried to fill in the blanks and guessed that that could have been the motivation. :) 

> I can see how that could be confusing though, so we can make them consistent. Do you have a preference on which would be better to receive?

Honestly, this isn't a strong opinion. It's _barely_ an opinion. Consistency matters a little, but so do convenience, stability and economy. I suppose passing the `WP_Post_Type` instance would be my instinctual choice. But I'm OK with almost anything else, as long as it was deliberate!

mcsf:
> I've consolidated these to be the same. Both filters will now pass the full `WP_Post_Type` object.

Ah, jinx!

packages/editor/src/components/provider/index.js
```diff
@@ -161,8 +165,17 @@ export const ExperimentalEditorProvider = withRegistryProvider(
...
 		const mode = useSelect(
-			( select ) => select( editorStore ).getRenderingMode(),
-			[]
+			( select ) => {
+				const postTypeObject = select( coreStore ).getPostType(
+					post.type
+				);
+				return (
+					postTypeObject?.default_rendering_mode ??
+					settings.defaultRenderingMode ??
```
youknowriad: Two questions here: 

1- Should we remove the settings.defaultRenderMode from here ? 
2- Should we avoid rendering the editor until the postType object is loaded (to avoid having a jump in rendering mode)

lib/compat/wordpress-6.7/post.php
ramonjd: I see there's an [agreement](https://github.com/WordPress/gutenberg/pull/62304#pullrequestreview-2102389815) that `template-locked` should be the default, so it looks like things work out of the box, even if unintentionally 😄 

However I'd say it'd be beneficial to be intentional about which values are acceptable to avoid state pollution or any side-effects on the frontend.

> Is it just post-only, template-only, template-lock?

If these docs are accurate, then the valid values are `post-only` and `template-locked`

https://github.com/WordPress/gutenberg/blob/3bd67998aec448cfb3e663895def681d21648181/docs/reference-guides/data/data-core-editor.md#L1451-L1451

packages/editor/src/components/provider/index.js
TylerB24890: @youknowriad 

> Should we remove the settings.defaultRenderMode from here ?

I don't think it hurts to keep `settings.defaultRenderingMode` really for backwards compatibility or giving another place to define the default, but I can remove it if we feel it's unnecessary. Perhaps it's better to have a single source of truth rather than having multiple ways to define it?

> Should we avoid rendering the editor until the postType object is loaded (to avoid having a jump in rendering mode)

I think that's probably a good idea. I've added a check that `mode` isn't `falsey` with the `isReady` check, so it will return `null` until `mode` is available.

lib/compat/wordpress-6.7/post.php
TylerB24890: Updated! It seems `template-lock` and `template-locked` are both actually supported, so I've added support for both in the check.

packages/editor/src/components/provider/index.js
youknowriad: > Perhaps it's better to have a single source of truth rather than having multiple ways to define it?

I think this was not really an API third-parties were able to use easily, so I'd be for removing it and keeping things simple (if that's possible :P).

packages/editor/src/components/provider/index.js
```diff
@@ -161,18 +165,30 @@ export const ExperimentalEditorProvider = withRegistryProvider(
 		BlockEditorProviderComponent = ExperimentalBlockEditorProvider,
 		__unstableTemplate: template,
 	} ) => {
-		const { editorSettings, selection, isReady, mode } = useSelect(
+		const mode = useSelect(
```
youknowriad: Any reason for separating the two useSelect calls?

youknowriad (commented):
At the moment, the defaultRenderingMode for pages is "post-only", I guess we should make the update to "template-locked" as discussed above.

packages/editor/src/components/provider/index.js
TylerB24890: Nope! I've merged them now 🙂

packages/editor/src/components/provider/index.js
```diff
@@ -165,34 +165,25 @@ export const ExperimentalEditorProvider = withRegistryProvider(
...
+				);
 				return {
 					editorSettings: getEditorSettings(),
 					isReady: __unstableIsEditorReady(),
 					selection: getEditorSelection(),
+					mode: postTypeObject?.default_rendering_mode ?? 'post-only',
 				};
 			},
-			[]
+			[ post ]
```
youknowriad: Can we use post.type here, it seems it's the only thing we need.

TylerB24890: All set!

youknowriad (commented):
Would love some help testing this properly. (Maybe @fabiankaegy if you find some time) but this is looking good to me.

annezazu:
@WordPress/outreach for anyone who can help test and review this PR ❤️

lib/compat/wordpress-6.7/class-gutenberg-rest-post-types-controller-6-7.php
```diff
@@ -0,0 +1,60 @@
...
+		$context  = ! empty( $request['context'] ) ? $request['context'] : 'view';
+
+		// Property will only exist if the post type supports the block editor.
+		if ( 'edit' === $context && property_exists( $item, 'default_rendering_mode' ) ) {
+			/**
+			 * Filters the block editor rendering mode for a specific post type.
+			 *
+			 * The dynamic portion of the hook name, `$item->name`, refers to the post type slug.
+			 *
+			 * @since 6.6.0
```
ramonjd: Nice documenting!

Is this for the next release (6.6) or 6.7?

lib/compat/wordpress-6.7/class-gutenberg-rest-post-types-controller-6-7.php
```diff
@@ -0,0 +1,60 @@
...
+			 * @param string       $default_rendering_mode Default rendering mode for the post type.
+			 * @param WP_Post_Type $post_type              Post type object.
+			 * @return string Default rendering mode for the post type.
+			 */
+			$rendering_mode = apply_filters( "post_type_{$item->name}_default_rendering_mode", $item->default_rendering_mode, $item );
+
+			/**
+			 * Filters the block editor rendering mode for a post type.
+			 *
+			 * @since 6.6.0
```
ramonjd: Same

lib/compat/wordpress-6.7/post.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * post-only: This mode extracts the post blocks from the template and renders only those.
+ * The idea is to allow the user to edit the post/page in isolation without the wrapping template.
+ *
+ * template-locked: This mode renders both the template and the post blocks
+ * but the template blocks are locked and can't be edited. The post blocks are editable.
+ *
+ * @return array Array of available rendering modes.
+ */
+function gutenberg_rendering_modes() {
```
ramonjd: Not a blocker, just a question - I'm assuming that this will be `wp_rendering_modes` or `rendering_modes` when it eventually makes it to Core.

What about making the name be more specific, e.g., `gutenberg_post_type_rendering_modes` or even `gutenberg_get_post_type_rendering_modes`?

lib/compat/wordpress-6.7/post.php
```diff
@@ -0,0 +1,57 @@
...
+	return [
```
ramonjd: WordPress coding standards demand `array()`

Pay tribute! 😄

ramonjd (commented):
This is working well for me 🎉 

If @fabiankaegy is happy that it satisfies https://github.com/WordPress/gutenberg/issues/58038 then I suppose the next step is the accompanying Core patch?

Happy to help test that as well. Cheers!

lib/compat/wordpress-6.7/class-gutenberg-rest-post-types-controller-6-7.php
tylerb-10up: Good catch! Was supposed to be 6.7 🙂 

Updated!

lib/compat/wordpress-6.7/class-gutenberg-rest-post-types-controller-6-7.php
tylerb-10up: Updated to 6.7!

lib/compat/wordpress-6.7/post.php
tylerb-10up: That's a good question, and I assume you're correct about it being renamed to `wp_rendering_modes` in Core. 

I think it makes sense to prefix it with `post_type_` so the final function would be `wp_post_type_rendering_modes()`

I'm also not opposed to using `get_` in the prefix, if the general that's the general consensus. In the meantime I've updated it to use `gutenberg_post_type_rendering_modes()` -- happy to update again if necessary!

TylerB24890: That's a good question, and I assume you're correct about it being renamed to `wp_rendering_modes` in Core.

I think it makes sense to prefix it with post_type_ so the final function would be `wp_post_type_rendering_modes()`

I'm also not opposed to using `get_` in the prefix, if the general that's the general consensus. In the meantime I've updated it to use `gutenberg_post_type_rendering_modes()` -- happy to update again if necessary!

lib/compat/wordpress-6.7/post.php
TylerB24890: Old habits 😅 

Updated accordingly.

lib/compat/wordpress-6.7/class-gutenberg-rest-post-types-controller-6-7.php
TylerB24890: Good catch! Was supposed to be 6.7 🙂

lib/compat/wordpress-6.7/class-gutenberg-rest-post-types-controller-6-7.php
TylerB24890: Updated to 6.7!

fabiankaegy (approved):
For me this is exactly doing what I had in mind and testing perfectly! :) 

I'm not sure about the failing tests though. They don't seem to be related but I'm not sure. 

@youknowriad would love to get your eyes on that if you have a moment <3

youknowriad:
I looked at this earlier today, and it's good to ship to me as well.

I think tests are somewhat stable these days, so if restart (or rebasing) doesn't really fix them than there must be a relationship with this PR.

I think the switch to "template by default" for pages is probably impactful for e2e tests for instance.

ramonjd:
I've kicked off the tests again. 🤞🏻 🤞🏻 🤞🏻 🤞🏻 

The final thing to do is create a Core backport PR.

See: https://github.com/WordPress/gutenberg/blob/trunk/backport-changelog/readme.md

Getting a draft PR up would be enough, and then I think the CI tests will be happy and the discussion about how to integrate into Core can move there.

Thanks a lot!

fabiankaegy:
@youknowriad Thanks to the help of @Sidsector9 we finally figured out the failing tests 🚀 

We'll work on the back port PR on Friday to get that added also :) 

Let us know if there is anything else of if this is good to be merged

youknowriad:
All good for me, I think there's probably a remaining failure on the performance tests it seems. the subtlety for the performance test is that the same test should be able to run on two different environments (the PR environment and the trunk environment) to be able to compare.

TylerB24890:
Backport PR created w/ Trac ticket.

Also updated the order of the `post_type_rendering_mode` filters so the `post_type_${post_type}_*` filter will apply after the generic filter. cc @fabiankaegy

youknowriad:
Would love to see this PR move forward and land. @fabiankaegy Is that something you can help with? How can I help personally?

fabiankaegy:
@youknowriad We've been trying to move it forward but failed on how to best deal with the performance tests... 

Most of my contributions right now are focussed on the 6.7 documentation efforts. But I can work with @TylerB24890 to get the PR rebased / updated again and if you have any insights on how we can deal with the performance test that would be incredibly helpful ❤️

packages/e2e-test-utils-playwright/src/admin/create-new-post.ts
```diff
@@ -38,8 +39,12 @@ export async function createNewPost(
 		query.set( 'excerpt', excerpt );
 	}
 
+	// Navigate to the new post page.
 	await this.visitAdminPage( 'post-new.php', query.toString() );
 
+	// Set rendering mode.
+	await this.editor.setRenderingMode( 'post-only' );
```
youknowriad: Do we really want to enforce that for all post types by default in the tests?

youknowriad:
Based on the screenshots in the performance test failures, it seems like the the performance tests are expecting the page to be in "template-locked" (access the post content block) but for some reason, we're on "post-only" mode there.

packages/editor/src/components/provider/index.js
```diff
@@ -301,15 +314,15 @@ export const ExperimentalEditorProvider = withRegistryProvider(
...
+			setRenderingMode( mode );
+		}, [ mode, setRenderingMode ] );
 
 		useHideBlocksFromInserter( post.type, mode );
 
 		// Register the editor commands.
 		useCommands();
 
-		if ( ! isReady ) {
+		if ( ! isReady || ! mode ) {
```
youknowriad: One potential here is that the mode can be defined even if the "postType" didn't load yet because of the "post-only" fallback. It seems better to avoid rendering anything until "postType" is defined.

lib/compat/wordpress-6.8/post.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * template-locked: This mode renders both the template and the post blocks
+ * but the template blocks are locked and can't be edited. The post blocks are editable.
+ *
+ * @return array Array of available rendering modes.
+ */
+function gutenberg_post_type_rendering_modes() {
+	return array(
+		'post-only',
+		'template-lock',
```
youknowriad: Why is there a "template-lock" rendering mode, AFAIK, there's only two modes.

packages/editor/src/components/provider/use-block-editor-settings.js
```diff
@@ -143,7 +143,7 @@ function useBlockEditorSettings( settings, postType, postId, renderingMode ) {
 				: undefined;
 
 			function getSectionRootBlock() {
-				if ( renderingMode === 'template-locked' ) {
+				if ( renderingMode.includes( 'template-lock' ) ) {
```
youknowriad: What is this change about?

fabiankaegy: Reverted this. I think it was a sideeffect of us trying to fix the tests. But it isn't right

packages/editor/src/components/provider/index.js
fabiankaegy: Added a check to see if the post type selector has resolved. However I still see an unstyled flash of content 🤔

lib/compat/wordpress-6.8/post.php
fabiankaegy: I'm not finding any other references of this in the codebase. Removed 👍

packages/editor/src/components/provider/index.js
```diff
@@ -301,15 +323,15 @@ export const ExperimentalEditorProvider = withRegistryProvider(
 
 		// Sets the right rendering mode when loading the editor.
 		useEffect( () => {
```
youknowriad, line 328: Maybe you can try using useLayoutEffect here to see if it improves the "unstyled flashing"

fabiankaegy: Even with the layout effect, I still get this 🤔 

https://github.com/user-attachments/assets/7bb7f81a-2d22-413a-b5d6-d2fec27b997c

fabiankaegy:
I'm also still very confused what is happening with the mobile unit tests 🤔

packages/editor/src/components/provider/index.js
youknowriad: I pushed a fixed for this. The issue was elsewhere, we were using the "post" blocks if the "template" didn't load yet.

youknowriad:
For the mobile unit tests, what's happening is that `getPostType` calls don't return anything at the moment in the unit test, so the editor doesn't initialize (doesn't render the block list). The setup of these unit-tests is a bit complex so I wasn't able to fix it. We need to either mock the REST API or mock the selector call I believe. Maybe @dcalhoun could help here.

packages/editor/src/components/provider/index.js
fabiankaegy: The fix works beautifully!

fabiankaegy:
I also just noticed a bug that I probably introduced yesterday in my rebase. For the posts post type (or likely and post type that doesn't have the preview enabled by default) the template doesn't seem to get resolved at all. 

My guess is that there is some early return somewhere if the post type doesn't have the default rendering mode set to `template-locked` 

CleanShot 2024-11-15 at 11 12 17@2x

fabiankaegy:
> I also just noticed a bug that I probably introduced yesterday in my rebase. For the posts post type (or likely and post type that doesn't have the preview enabled by default) the template doesn't seem to get resolved at all.
> 
> My guess is that there is some early return somewhere if the post type doesn't have the default rendering mode set to `template-locked`
> 
> CleanShot 2024-11-15 at 11 12 17@2x

This has been fixed in https://github.com/WordPress/gutenberg/pull/62304/commits/a742e4a8c4c0f9c2f8368924967210de680df95a

dcalhoun:
> For the mobile unit tests, what's happening is that `getPostType` calls don't return anything at the moment in the unit test, so the editor doesn't initialize (doesn't render the block list). The setup of these unit-tests is a bit complex so I wasn't able to fix it. We need to either mock the REST API or mock the selector call I believe. Maybe @dcalhoun could help here.

@youknowriad this appears to be a legitimate bug in the mobile editor, not an issue in the unit test setup. The `mode` is always `undefined`, leading to the editor provider always rendering `null`. 

https://github.com/WordPress/gutenberg/blob/a742e4a8c4c0f9c2f8368924967210de680df95a/packages/editor/src/components/provider/index.js#L327-L330

I observed that, in the mobile editor, the `renderingMode` reducer never runs even though the `setRenderingMode` action is triggered in the provider's effect—the rendering mode is never set the `defaultMode` value. Do have any thoughts as to why this may be the case?

youknowriad:
@dcalhoun As you can see here, we have this code in the editor provider

```
const postTypeObject = select( coreStore ).getPostType(
					post.type
				);
```

This code performs a REST API to fetch the post type with the rendering mode. I don't know how to test the app personally but at least in the mobile unit tests, that object was always "undefined" which means, nothing was rendered.

dcalhoun:
> I don't know how to test the app personally but at least in the mobile unit tests, that object was always "undefined" which means, nothing was rendered.

@youknowriad correct, but if that object is `undefined`, the rendering mode should fallback to the default—`post-only`. Currently it is not, both in the mobile editor and its automated tests. 

https://github.com/WordPress/gutenberg/blob/a742e4a8c4c0f9c2f8368924967210de680df95a/packages/editor/src/components/provider/index.js#L197-L198

I will continue debugging, but wanted to ask if you had insight on that oddity.

youknowriad:
In case it helps, There's this condition

```
if ( ! isReady || ! mode || ! hasLoadedPostObject ) {
```

This condition means that we don't render anything if the object didn't load yet, regardless of whether there's a fallback rendering mode.

dcalhoun:
@youknowriad @TylerB24890 the mobile editor breakage and failing automated tests should be resolved by the following: 

* https://github.com/TylerB24890/gutenberg/pull/2
* https://github.com/TylerB24890/gutenberg/pull/3

The former is merged, the latter needs review and merging. 🚀

fabiankaegy:
Okay now we just have to figure out how to fix the performance tests 😅

fabiankaegy:
@youknowriad in my digging it appears that this line here when we load the sample HTML file for the large post

https://github.com/WordPress/gutenberg/blob/ec543c5939e8459680c948c27f78221f3f95d933/test/performance/fixtures/perf-utils.ts#L148

That doesn't update the blocks that are stored in the post content but rather replaces the blocks in the template itself 🤔 

I have added a workarround that switches the rendering mode to `post-only` before it calls the `replaceBlocks` dispatch function. It works as expected now but I'm not sure if there isn't a better way ti improve the tests

youknowriad:
This seems ready to go, If I'm not wrong the main impact of this PR is that pages are going to show templates in the post editor by default as well.

So just a final heads-up before merge @jasmussen 

Also can we do a rebase here to ensure that are no regressions/hidden issues with the latests updates in trunk?

fabiankaegy:
@youknowriad just updated the branch with the latest trunk. I agree it's ready 👍 Happy to merge whenever :)

packages/editor/src/components/provider/index.js
```diff
@@ -301,15 +326,15 @@ export const ExperimentalEditorProvider = withRegistryProvider(
...
-			setRenderingMode( settings.defaultRenderingMode ?? 'post-only' );
-		}, [ settings.defaultRenderingMode, setRenderingMode ] );
+			setRenderingMode( defaultMode );
+		}, [ defaultMode, setRenderingMode ] );
```
youknowriad, line 330: This change is one that I'm not entirely sure about. I think the default rendering mode should only be enforced when the "edited entity" changes. In practice though, this seems to have the same effect. 

Anyway, not a big deal for now.

test/performance/fixtures/perf-utils.ts
```diff
@@ -133,6 +133,18 @@ export class PerfUtils {
...
+			const { select } = window.wp.data;
+			return select( 'core/editor' ).getRenderingMode();
+		} );
+
+		if ( initialRenderingMode !== 'post-only' ) {
+			await this.page.evaluate( () => {
+				const { dispatch } = window.wp.data;
+				dispatch( 'core/editor' ).setRenderingMode( 'post-only' );
+			} );
+		}
```
youknowriad: I actually wonder if this should be in this function or directly in the tests that actually load pages. Because it's not clear why a "load blocks" function should change the rendering mode.

packages/edit-site/src/components/block-editor/use-site-editor-settings.js
```diff
@@ -36,9 +36,7 @@ function useNavigateToPreviousEntityRecord() {
 	return goBack;
 }
 
-export function useSpecificEditorSettings(
-	shouldUseTemplateAsDefaultRenderingMode
-) {
+export function useSpecificEditorSettings() {
```
youknowriad, line 39: Where is this function being called, should we update the caller to drop the argument.

fabiankaegy: Fixed in [557ec18](https://github.com/WordPress/gutenberg/pull/62304/commits/557ec18c66a8d5557cdeefc22ceefc0ecf062a7b)

test/performance/fixtures/perf-utils.ts
fabiankaegy: Fixed in [9068ab1](https://github.com/WordPress/gutenberg/pull/62304/commits/9068ab10786095bd2050e51839d5d5022b887616)

test/performance/specs/post-editor.spec.js
```diff
@@ -142,6 +143,7 @@ test.describe( 'Post Editor Performance', () => {
 
 		test( 'Setup the test post', async ( { admin, perfUtils, editor } ) => {
 			await admin.createNewPost();
+			await perfUtils.enterPostOnlyRenderingMode();
```
youknowriad: Why would we need this one here? posts are "post-only" by default no?

test/performance/fixtures/perf-utils.ts
```diff
@@ -97,6 +97,29 @@ export class PerfUtils {
...
+	/**
+	 * Enters the post-only rendering mode.
+	 *
+	 * This is sometimes needed when for example we want to update the contents
+	 * of the editor from a HTML file. Calling the resetBlocks method of the
+	 * core/block-editor store will replace the contents of the template if the
+	 * rendering mode is not post-only. So this should always be called before
+	 * the resetBlocks method is used.
+	 */
+	async enterPostOnlyRenderingMode() {
```
youknowriad: Should we accept the desired rendering mode as argument

fabiankaegy: Fixed in [4512fce](https://github.com/WordPress/gutenberg/pull/62304/commits/4512fce47e0f424c97c5bc4d54d0cc659ac4de9e)

test/performance/specs/post-editor.spec.js
fabiankaegy: Fixed in [4512fce](https://github.com/WordPress/gutenberg/pull/62304/commits/4512fce47e0f424c97c5bc4d54d0cc659ac4de9e)

fabiankaegy: (Good catch by the way!)

backport-changelog/6.7/7129.md
```diff
@@ -0,0 +1,3 @@
+https://github.com/WordPress/wordpress-develop/pull/7129
+
+* https://github.com/WordPress/gutenberg/pull/62304
```
youknowriad: I guess this file should be move to 6.8 folder

fabiankaegy: Fixed in [f282a47](https://github.com/WordPress/gutenberg/pull/62304/commits/f282a47a81086776bcb0e1875a8fa8e8fa6ba10c)

youknowriad:
Thanks all for your help here. This is a nice one.

costasovo:
🎉 This is great! This is exactly what we need for the email editor project. We always want to display full email content including template content.

lib/compat/wordpress-6.8/post.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+function gutenberg_post_type_default_rendering_mode( $args, $post_type ) {
+	$rendering_mode  = 'page' === $post_type ? 'template-locked' : 'post-only';
+	$rendering_modes = gutenberg_post_type_rendering_modes();
+
+	// Make sure the post type supports the block editor.
+	if (
+		wp_is_block_theme() &&
+		( isset( $args['show_in_rest'] ) && $args['show_in_rest'] ) &&
+		( isset( $args['supports'] ) && in_array( 'editor', $args['supports'], true ) )
```
jsnajdr, line 42: This is causing a crash when the `supports` arg to `register_post_type` is `false`:
```
Fatal error: Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, bool given in /var/www/html/wp-content/plugins/gutenberg/lib/compat/wordpress-6.8/post.php:42
Stack trace:
#0 /var/www/html/wp-content/plugins/gutenberg/lib/compat/wordpress-6.8/post.php(42): in_array('editor', false, true)
#1 /var/www/html/wp-includes/class-wp-hook.php(324): gutenberg_post_type_default_rendering_mode(Array, 'product_variati...')
#2 /var/www/html/wp-includes/plugin.php(205): WP_Hook->apply_filters(Array, Array)
#3 /var/www/html/wp-includes/class-wp-post-type.php(491): apply_filters('register_post_t...', Array, 'product_variati...')
#4 /var/www/html/wp-includes/class-wp-post-type.php(467): WP_Post_Type->set_props(Array)
#5 /var/www/html/wp-includes/post.php(1805): WP_Post_Type->__construct('product_variati...', Array)
#6 /var/www/html/wp-content/plugins/woocommerce/includes/class-wc-post-types.php(442): register_post_type('product_variati...', Array)
```

WooCommerce [registers its `product_variation` post type](https://github.com/woocommerce/woocommerce/blob/572470776ce9791c019d50af1ac0ab0672a10379/plugins/woocommerce/includes/class-wc-post-types.php#L450) with `supports => false`. It's a supported and documented value for `supports`, whose type is `array|false`.

jsnajdr: Fixing in #67225.

Mamaduka:
It's been a few weeks, and I still find the new default rendering mode confusing for pages. Maybe the pages should default to `template-locked` only in the Site Editor and `post-only` in the post-type Editor.

IMO, this is a significant behavior change for Pages in the post editor. Unfortunately, we'll only get real feedback after WP 6.8 is released.

**Update:** The change also breaks page editing for non-Administrator users with `edit_pages` capability, such as Editors.

P.S. There's already a report about the changes in List View behavior for page - #67857.


## Screenshot (broken page for Editors)

CleanShot 2024-12-12 at 11 35 36

fabiankaegy:
@Mamaduka I'll say that I have shipped 3 projects where we used the hacky JS approach to change the default rendering mode: 

```js
import { registerPlugin } from '@wordpress/plugins';
import { useDispatch, useSelect } from '@wordpress/data';
import { useEffect } from '@wordpress/element';
import { store as editorStore } from '@wordpress/editor';

// Define post types that show the full template locked site editor by default.
const TEMPLATE_LOCKED_POST_TYPES = ['page', 'post'];

registerPlugin('namespace-set-default-template-rendering-mode', {
	render: function Edit() {
		const { setRenderingMode } = useDispatch(editorStore);

		const [postType] = useSelect((select) => [select(editorStore).getCurrentPostType()], []);

		useEffect(() => {
			if (!TEMPLATE_LOCKED_POST_TYPES.includes(postType)) {
				return;
			}
			setRenderingMode('template-locked');
		}, [setRenderingMode, postType]);

		return null;
	},
});
```

And have had great success with it. However when we changed it, we changed it across the board for all content types so the behavior was consistent. 

I agree that it is a big change and we should definitely be careful about rolling it out. 

I hope that if we decide to roll it back though, that we keep the underlying property on the CPT so that we can override the default rendering mode much easier than we do it today since the solution today creates some unstyled flashes of content whilst loading

fabiankaegy:
@Mamaduka 

Do you have any more details about: 

> Update: The change also breaks page editing for non-Administrator users with edit_pages capability, such as Editors.

This here only changes the default? It was long before possible to preview the template even as a non admin. Sounds like there is a regression there 🤔

swissspidy:
Can we please rename default_rendering_mode to something else that makes it clearer that it‘s about the editor? Maybe prefix it with editor_ at least.

Also, have you considered using the post type supports feature for that, rather than adding a new top level prperty?

fabiankaegy:
@swissspidy all open on the name 👍 

> Also, have you considered using the post type supports feature for that, rather than adding a new top level property?

Yes but deemed it not the right choice here because even before this was merged in a block theme every single post type supports the ability to preview the template as you are editing the post. So this is not a new support, it is just enabling it by default. (Allowing you to change the default)

Since there are multiple possible options here we would need a key value pair of some kind 🤔 which post type supports are not

Mamaduka:
@fabiankaegy, I'm not saying we should revert the whole feature; I quite like it.

I think we should consider whether Pages will ship with `template-locked` as the default rendering mode. IMO, it makes sense for the site editor, but for a post-block editor, it's a big leap. Maybe we should make it `opt-in` for all post types.

> Do you have any more details about.

I don't have them now, but when the default rendering mode is set to `template-locked`, nothing is rendered in the post editor when accessing it as an Editor. We should also test other capabilities and CPTs to see what is failing.

> Also, have you considered using the post type supports feature for that, rather than adding a new top level prperty?

That would require changing the shape of the `supports` property since it is a one-dimensional array.

swissspidy:
> Since there are multiple possible options here we would need a key value pair of some kind 🤔 which post type supports are not
> That would require changing the shape of the `supports` property since it is a one-dimensional array.

So? :) That sounds feasible to implement and has come up in the past already anyway. It could be added as a config to the editor support.

Definitely worth exploring IMO. Way more future proof if you need any more configs in the future.

> all open on the name 👍

If post type support doesn‘t work, I‘d suggest editor_rendering_mode at least.

Mamaduka:
Maybe we should fold editor-specific settings into a new object. Technically speaking, `template` and `template_lock` are also block editor specific.

youknowriad:
One thing that I personally value is the consistency between site and post editors, these are not separate editors, these are just how things have evolved and it will "merge" fully at some point (single url).

That said for the choice of the default rendering mode for pages, I think it's a very hard question to answer, I've seen feedback in both ways: some folks (mostly devs) have a clear separation in their head so they prefer just the abstracted view, some users prefer to see the full page.

swissspidy:
> > Since there are multiple possible options here we would need a key value pair of some kind 🤔 which post type supports are not
> > That would require changing the shape of the `supports` property since it is a one-dimensional array.
> 
> So? :) That sounds feasible to implement and has come up in the past already anyway. It could be added as a config to the editor support.
> 
> Definitely worth exploring IMO. Way more future proof if you need any more configs in the future.
> 
> > all open on the name 👍
> 
> If post type support doesn‘t work, I‘d suggest editor_rendering_mode at least.

I take it back, this actually **is supported**!

> A feature can also be  specified as an array of arguments to provide additional information about supporting that feature.

Mamaduka:
> I take it back, this actually is supported!

@swissspidy, can you clarify? Does that mean that the post type `supports` property can have multi-dimensional arrays?

> That said for the choice of the default rendering mode for pages, I think it's a very hard question to answer, I've seen feedback in both ways: some folks (mostly devs) have a clear separation in their head so they prefer just the abstracted view, some users prefer to see the full page.

@youknowriad, I think we should ship this as an opt-in feature, at least for now.

youknowriad:
> I think we should ship this as an opt-in feature, at least for now.

The problem for me is the inconsistency with the site editor, we need to change one behavior, so it's going to change something regardless.

swissspidy:
> can you clarify? Does that mean that the post type `supports` property can have multi-dimensional arrays?

@Mamaduka Yes! Example: `array( 'my_feature', array( 'field' => 'value' ) )`.

lib/compat/wordpress-6.8/class-gutenberg-rest-post-types-controller-6-8.php
```diff
@@ -0,0 +1,61 @@
...
+		$context  = ! empty( $request['context'] ) ? $request['context'] : 'view';
+
+		// Property will only exist if the post type supports the block editor.
+		if ( 'edit' === $context && property_exists( $item, 'default_rendering_mode' ) ) {
+			/**
+			 * Filters the block editor rendering mode for a post type.
+			 *
+			 * @since 6.8.0
+			 * @param string       $default_rendering_mode Default rendering mode for the post type.
+			 * @param WP_Post_Type $post_type              Post type name.
```
swissspidy, line 33: This is actually a `WP_Post` instance here, not `WP_Post_Type`.

The variable name is also `$item`, not `$post_type`.

lib/compat/wordpress-6.8/post.php
```diff
@@ -0,0 +1,57 @@
...
+	) {
+		// Validate the supplied rendering mode.
+		if (
+			isset( $args['default_rendering_mode'] ) &&
+			in_array( $args['default_rendering_mode'], $rendering_modes, true )
+		) {
+			$rendering_mode = $args['default_rendering_mode'];
+		}
+
+		$args['default_rendering_mode'] = $rendering_mode;
```
swissspidy, line 52: Surprised that we even allow adding custom args... We should have added an allowlist.

Mamaduka:
@youknowriad, how should we handle users who can edit pages but not templates and template parts? Currently, page editing is broken for these users.

fabiankaegy:
@Mamaduka in https://github.com/WordPress/gutenberg/pull/60447 we enabled the show template option for any user role. 

This was preceded by https://github.com/WordPress/gutenberg/pull/60317 which made the templates read only for all users

Mamaduka:
@fabiankaegy, I'll debug it tomorrow. It could be an unrelated issue; I also see the `settings` endpoints error when testing with the Editor role.

We also need to test various cases to ensure everything works as expected. Here's a bug report for the case we missed, #67875.

Some flows that come to mind:

* Block theme with lower cap users.
* Block theme, but the page has a PHP template assigned to it.

annezazu:
>I think we should consider whether Pages will ship with template-locked as the default rendering mode. IMO, it makes sense for the site editor, but for a post-block editor, it's a big leap. Maybe we should make it opt-in for all post types.

FWIW I agree with @Mamaduka here. This is a **big** change for folks to see by default and I think it'll create a lot of confusion to suddenly see. WordPress.com does run the Gutenberg plugin so I'll see what I can do to get feedback earlier in the release cycle before 6.8 hits in case it helps.

t-hamano:
>I think we should consider whether Pages will ship with template-locked as the default rendering mode. IMO, it makes sense for the site editor, but for a post-block editor, it's a big leap. Maybe we should make it opt-in for all post types.

I agree. The Page is also a post type, so I don't understand why it's treated specially. I was confused when I first opened a page in the post editor.

Previously, users could only see the content area in the post editor for any post type and edit all the blocks. If templates are shown by default, it's confusing what they mean and why you can't remove blocks.

Furthermore, the only way to tell if a template is shown is by an icon in the document bar. Many people may not even know what this icon means in the first place.

Mamaduka:
@jorgefilipecosta and I somewhat branched off the discussion in #68110.

The proposed behavior will be "broken" for non-administrator users who can edit the pages, as the default rendering mode will be switched to "post-only".

packages/editor/src/components/provider/index.js
```diff
@@ -282,7 +299,15 @@ export const ExperimentalEditorProvider = withRegistryProvider(
...
 			}
-		}, [] );
+		}, [
+			createWarningNotice,
+			initialEdits,
+			settings,
+			post,
+			recovery,
+			setupEditor,
+			updatePostLock,
```
youknowriad, line 309: @fabiankaegy Is the addition of the dependencies here in relation to the current PR or was it done just to please the linter?

This is creating a bug when the post object is update, the setup happens again and initial edits are applied again even if they shouldn't be.

Mamaduka:
Created a new issue to discuss the new default rendering mode for page post type - #68684.

lib/compat/wordpress-6.8/class-gutenberg-rest-post-types-controller-6-8.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+class Gutenberg_REST_Post_Types_Controller_6_8 extends WP_REST_Post_Types_Controller {
+	/**
+	 * Add Block Editor default rendering mode setting to the response.
+	 *
+	 * @param  WP_Post_Type    $item    Post type object.
+	 * @param  WP_REST_Request $request Request object.
+	 * @return WP_REST_Response Response object.
+	 */
+	public function prepare_item_for_response( $item, $request ) {
```
jsnajdr, line 22: Not sure if this was discussed before, or if there is some hidden caveat, but to merely add a new field to an existing endpoint, you don't need to override the entire controller. There is an API for that:
```
register_rest_field( 'type', 'default_rendering_mode', ... );
```

jsnajdr: Update: now I see that this was already solved by @Mamaduka in #68549 and #67867 🚀 Nice coincidence.

Mamaduka: Thanks, @jsnajdr!

PR #61262, opened 2024-04-30
Remove additional call to `WP_Theme_JSON_Gutenberg::__construct`

kt-12:
## What?
Introduce `WP_Theme_JSON_Data_Gutenberg::get_theme_json()` to avoid the cyclic call to `WP_Theme_JSON_Gutenberg::__construct`.

Trac ticket: [Trac 61112](https://core.trac.wordpress.org/ticket/61112)

## Why?
Doing similar changes in core, we got a consistent ~2.5% improvement by reducing 8 calls to the constructor  [reference core pr](https://github.com/WordPress/wordpress-develop/pull/6271#pullrequestreview-2032238553).

## How?
`WP_Theme_JSON_Data_Gutenberg` internally has a `theme_json` private attribute which is an instance of `WP_Theme_JSON_Gutenberg`. At the moment we try to form the raw JSON data from the `WP_Theme_JSON_Gutenberg` object and then again from the Object from the raw JSON data, which leads to regression.  In this PR we use the existing `WP_Theme_JSON_Gutenberg` object available to `WP_Theme_JSON_Data_Gutenberg` instead of the cyclic way.

PR #62305, opened 2024-06-04
Migrate theme.json based on origin

ajlende:
## What?

- Updates the API for `WP_Theme_JSON_Schema_Gutenberg::migrate` to accept an origin.
- Use the new origin argument to conditionally migrate changes in theme.json v3

## Why?

See the discussion in https://github.com/WordPress/gutenberg/pull/61262/files#r1589940214

@oandregal
> Perhaps we can absorb `$config['isGlobalStylesUserThemeJSON'] = true;` as part of `WP_Theme_JSON_Data_Gutenberg`. We may want to still check that's true before returning to be extra-safe (line 553) but we wouldn't need a second transformation.

@joemcgill 
> Looks like the `$config['isGlobalStylesUserThemeJSON']` modification was made as part of [2012a36](https://github.com/WordPress/gutenberg/commit/2012a36da2a7752dceeda3dcf0af8e58985e8072) (#58409).
> 
> @ajlende is there a reason why this would still be necessary if we use the `WP_Theme_JSON_Gutenberg` object already present in the `WP_Theme_JSON_Data_Gutenberg` object?

@ajlende
> The `$config['isGlobalStylesUserThemeJSON']` is required for migrations that don't need to run for the custom origin in `WP_Theme_JSON_Schema::migrate` (called on `new WP_Theme_JSON()`). If the origin could be passed to `migrate` in some other way, then it would be fine.
> 
> https://github.com/WordPress/gutenberg/blob/cc2fcbe5e7d4384c27aa6615d2a9660f873fe572/lib/class-wp-theme-json-schema-gutenberg.php#L115-L124

### DIFF
--- a/lib/class-wp-theme-json-resolver-gutenberg.php
+++ b/lib/class-wp-theme-json-resolver-gutenberg.php
@@ -540,10 +540,8 @@ class WP_Theme_JSON_Resolver_Gutenberg {
 		}
 
 		/** This filter is documented in wp-includes/class-wp-theme-json-resolver.php */
-		$theme_json = apply_filters( 'wp_theme_json_data_user', new WP_Theme_JSON_Data_Gutenberg( $config, 'custom' ) );
-		$config     = $theme_json->get_data();
-
-		static::$user = new WP_Theme_JSON_Gutenberg( $config, 'custom' );
+		$theme_json   = apply_filters( 'wp_theme_json_data_user', new WP_Theme_JSON_Data_Gutenberg( $config, 'custom' ) );
+		static::$user = $theme_json->get_theme_json();
 
 		return static::$user;
 	}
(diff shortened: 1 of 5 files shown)


lib/class-wp-theme-json-resolver-gutenberg.php
```diff
@@ -576,9 +577,6 @@ public static function get_user_data() {
 		$theme_json = apply_filters( 'wp_theme_json_data_user', new WP_Theme_JSON_Data_Gutenberg( $config, 'custom' ) );
 		$config     = $theme_json->get_data();
 
-		// Needs to be set for schema migrations of user data.
```
oandregal, line 545: Additionally, we can add back the changes at https://github.com/WordPress/gutenberg/pull/61262:

```php
$theme_json = apply_filters( 'wp_theme_json_data_user', new WP_Theme_JSON_Data_Gutenberg( $config, 'custom' ) );
static::$user = $theme_json->get_theme_json();

return static::$user;
```

oandregal (approved):
I've left a comment to address and it seems this needs a rebase. But it's a really nice update. Thanks for the follow-up!

lib/class-wp-theme-json-resolver-gutenberg.php
oandregal: done it myself

oandregal:
This is blocked by the performance test failures, which are unrelated and are being investigated at https://github.com/WordPress/gutenberg/pull/62331

oandregal:
Backport at https://github.com/WordPress/wordpress-develop/pull/6737

PR #62306, opened 2024-06-05
Components: Fix inaccessible disabled `Button`s

mirka:
Follow-up to #62080 and #62301

## What?

Adds the restricted syntax lint rules for components to the `@wordpress/components` files, and addresses the existing violations.

## Why?

As part of an effort to maintain accessibility standards

## Testing Instructions

✅ `npm run lint:js` should pass

### DIFF
--- a/packages/components/src/autocomplete/autocompleter-ui.tsx
+++ b/packages/components/src/autocomplete/autocompleter-ui.tsx
@@ -58,6 +58,7 @@ function ListBox( {
 					id={ `components-autocomplete-item-${ instanceId }-${ option.key }` }
 					role="option"
 					aria-selected={ index === selectedIndex }
+					__experimentalIsFocusable
 					disabled={ option.isDisabled }
 					className={ clsx(
 						'components-autocomplete__result',
--- a/packages/components/src/dropdown-menu/index.tsx
+++ b/packages/components/src/dropdown-menu/index.tsx
@@ -199,6 +199,7 @@ function UnconnectedDropdownMenu( dropdownMenuProps: DropdownMenuProps ) {
 											? control.role
 											: 'menuitem'
 									}
+									__experimentalIsFocusable
 									disabled={ control.isDisabled }
 								>
 									{ control.title }
--- a/packages/components/src/range-control/index.tsx
+++ b/packages/components/src/range-control/index.tsx
@@ -326,6 +326,8 @@ function UnforwardedRangeControl(
 					<ActionRightWrapper>
 						<Button
 							className="components-range-control__reset"
+							// If the RangeControl itself is disabled, the reset button shouldn't be in the tab sequence.
+							__experimentalIsFocusable={ ! disabled }
 							disabled={ disabled || value === undefined }
 							variant="secondary"
 							size="small"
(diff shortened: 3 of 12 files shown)


packages/components/src/range-control/index.tsx
```diff
@@ -326,6 +326,8 @@ function UnforwardedRangeControl(
 					<ActionRightWrapper>
 						<Button
 							className="components-range-control__reset"
+							// If the RangeControl itself is disabled, the reset button shouldn't be in the tab sequence.
+							__experimentalIsFocusable={ ! disabled }
 							disabled={ disabled || value === undefined }
```
mirka, line 331: There is a bug in the `disabled` logic here (`value` will never be undefined), extracted as a separate issue in #63061.

packages/components/src/autocomplete/autocompleter-ui.tsx
```diff
@@ -58,6 +58,7 @@ function ListBox( {
 					id={ `components-autocomplete-item-${ instanceId }-${ option.key }` }
 					role="option"
 					aria-selected={ index === selectedIndex }
+					__experimentalIsFocusable
```
mirka, line 61: Disabled items in a popover menu should be perceivable.

packages/components/src/dropdown-menu/index.tsx
```diff
@@ -199,6 +199,7 @@ function UnconnectedDropdownMenu( dropdownMenuProps: DropdownMenuProps ) {
 											? control.role
 											: 'menuitem'
 									}
+									__experimentalIsFocusable
```
mirka, line 202: Disabled items in a dropdown menu should be perceivable.

packages/components/src/toolbar/toolbar-button/index.tsx
mirka, line 1: Disabled toolbar buttons should always be perceivable.

mirka, line 1: We'll need to fix this one after #63070 is addressed.

packages/components/src/toolbar/toolbar-button/index.tsx
```diff
@@ -85,6 +86,10 @@ function UnforwardedToolbarButton(
 				<Button
 					label={ title }
 					isPressed={ isActive }
+					// TODO: Should be focusable disabled, but adding `__experimentalIsFocusable` will trigger a
+					// focus bug when ToolbarButton is disabled via the `disabled` prop.
+					// Must address first: https://github.com/WordPress/gutenberg/issues/63070
+					// eslint-disable-next-line no-restricted-syntax
 					disabled={ isDisabled }
```
mirka, line 93: We'll need to fix this one after #63070 is addressed.

packages/components/src/toolbar/toolbar-button/index.tsx
```diff
@@ -60,6 +60,7 @@ function UnforwardedToolbarButton(
 						className
 					) }
 					isPressed={ isActive }
+					__experimentalIsFocusable
```
mirka, line 63: Disabled toolbar buttons should always be perceivable.

ciampo (approved):
LGTM 

IMO it would be better to require a `disable reason` for both adding more context for future contributors, and to increase the accountability of anyone deciding to ignore the lint rule in the future.

PR #62308, opened 2024-06-05
`CustomSelectControlV2`: Fix initial value not being set in the select for the V1 legacy adapter

fullofcaffeine:
Split from: https://github.com/WordPress/gutenberg/pull/61272.

Depends on: https://github.com/WordPress/gutenberg/pull/62255.

## What?

Fixes an issue in the legacy adapter where the initial `value` passed to the select wasn't being auto-selected, effectively causing a mismatch between the store and the rendered value.

## Why?

Consumers expect the select to have the option defined by the `value` prop selected when it mounts, but the legacy adapter was not doing that, e.g: if you selected an option, and the component was unmounted/mounted from scratch again, it would select the first option instead of the option defined by `value`. 

IMPORTANT: V2 is still not replacing the classic V1 in the GB app yet. We'll eventually replace the classic export with the V1 adapter, and the commit [here](https://github.com/WordPress/gutenberg/pull/62308/commits/77212176165031cbe946d4c6c330b6d2bc493903) will be reverted before merging (it's here just to make it easier to test the adapter in the app while reviewing this PR).

## How?

Ths PR fixes by making sure the value in the Ariakit store is selected upon mount from the `value` passed.

## Testing Instructions

* Tests should pass;
* See the screenshot below to see how to test manually.

## Screenshots or screencast

| Before the fix | After the fix |
|---------|---------|
| <video src="https://github.com/WordPress/gutenberg/assets/81248/c3447a53-720a-4763-bad1-93dff3eb63df" controls="controls" style="max-width: 730px;"></video> | <video src="https://github.com/WordPress/gutenberg/assets/81248/4baf5bb0-c4aa-4c5a-9ae6-a606ea1518c3" controls="controls" style="max-width: 730px;"></video>  |

_**NOTE**: You may notice some styling issues in the select. These were already fixed in https://github.com/WordPress/gutenberg/pull/61272 and will be split in a follow-up PR to this one._

### DIFF
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -63,7 +63,8 @@ export {
 	useCompositeState as __unstableUseCompositeState,
 } from './composite';
 export { ConfirmDialog as __experimentalConfirmDialog } from './confirm-dialog';
-export { StableCustomSelectControl as CustomSelectControl } from './custom-select-control';
+//export { StableCustomSelectControl as CustomSelectControl } from './custom-select-control';
+export { ClassicCustomSelectControlV2Adapter as CustomSelectControl } from './custom-select-control-v2/legacy-component';
 export { default as CustomSelectControlV2 } from './custom-select-control-v2';
 export { default as Dashicon } from './dashicon';
 export { default as DateTimePicker, DatePicker, TimePicker } from './date-time';
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -63,7 +63,9 @@ export {
 	useCompositeState as __unstableUseCompositeState,
 } from './composite';
 export { ConfirmDialog as __experimentalConfirmDialog } from './confirm-dialog';
-export { StableCustomSelectControl as CustomSelectControl } from './custom-select-control';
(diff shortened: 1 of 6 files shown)


packages/components/src/index.ts
```diff
@@ -63,7 +63,8 @@ export {
 	useCompositeState as __unstableUseCompositeState,
 } from './composite';
 export { ConfirmDialog as __experimentalConfirmDialog } from './confirm-dialog';
-export { StableCustomSelectControl as CustomSelectControl } from './custom-select-control';
+//export { StableCustomSelectControl as CustomSelectControl } from './custom-select-control';
+export { ClassicCustomSelectControlV2Adapter as CustomSelectControl } from './custom-select-control-v2/legacy-component';
```
fullofcaffeine, line 67: This will be reverted before merging - it's here to allow testing the V1 adapter in the GB app. We shouldn't replace the classic V1 with the legacy adapter just yet (it'll be done in a follow-up PR).

ciampo:
Closing as superseded by #62733

PR #62309, opened 2024-06-05
Revert CSS removal for interface footer breadcrumbs

andrewserong:
## What?

Reverts part of the CSS removal for the interface footer's breadcrumbs rules, removed in #62237

## Why?

This CSS added left padding and ensured the height of the footer breadcrumbs was correct when no blocks were selected in the site and post editors. Re-instating the CSS appears to get the styling working again.

There's a comment in https://github.com/WordPress/gutenberg/pull/62237/files#r1624428431 that this CSS was in the wrong place. However, for now, let's re-instate the CSS and it can be moved elsewhere in a follow-up if need be. (I.e. the padding rule is appropriate as a child of the interface footer, but perhaps the height, etc, should be in the breadcrumbs component instead?)

## How?

* Re-instate some of the removed lines from #62237

## Testing Instructions

1. In `trunk`, open the post editor and don't select any blocks. Note that the height of the footer breadcrumbs is too small.
2. With this PR applied, the height should be correct. It should also look correct if you select nested blocks.
3. Repeat in the site editor and check that the interface footer's breadcrumbs look correct

### Before

image

image

### After

image

image

### DIFF
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -163,6 +163,16 @@ html.interface-interface-skeleton__html-container {
 	@include break-medium() {
 		display: flex;
 	}
+
+	.block-editor-block-breadcrumb {
+		z-index: z-index(".edit-post-layout__footer");
+		display: flex;
+		background: $white;
+		height: $button-size-small;
+		align-items: center;
+		font-size: $default-font-size;
+		padding: 0 ($grid-unit-15 + 6px);
+	}
 }
 
 .interface-interface-skeleton__actions {
--- a/packages/base-styles/_z-index.scss
+++ b/packages/base-styles/_z-index.scss
@@ -28,6 +28,7 @@ $z-layers: (
 	".block-editor-warning": 5,
 	".block-library-gallery-item__inline-menu": 20,
 	".block-editor-url-input__suggestions": 30,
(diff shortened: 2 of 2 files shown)


ramonjd (approved):
🚀 

### Trunk

https://github.com/WordPress/gutenberg/assets/6458278/941c80e0-b260-4b3f-ada7-cedfac1005c2


### This PR

https://github.com/WordPress/gutenberg/assets/6458278/daa4959d-5a38-40ce-b44b-e72b766f10db

andrewserong:
Thanks for the quick review @ramonjd!

I'll merge this in now since it's a visual issue on `trunk`. @youknowriad made good points about this CSS potentially belonging elsewhere, but I reckon that's best suited to a follow-up if it's still important to address after this lands.

packages/interface/src/components/interface-skeleton/style.scss
```diff
@@ -163,6 +163,16 @@ html.interface-interface-skeleton__html-container {
 	@include break-medium() {
 		display: flex;
 	}
+
+	.block-editor-block-breadcrumb {
```
youknowriad, line 167: Not really clear why this is in the interface package indeed. We can move it to the breadcrumb component directly.
The other thing is that it has a lot of unnecessary stuff. Why it has a background, z-index, ...

ellatrix:

PR #61474, opened 2024-05-08
[WIP] List View: Try allowing page up and page down keys

andrewserong:
## What?

🚧 🚧 🚧 🚧 WIP: not yet ready for review 🚧 🚧 🚧 🚧 

Explore allowing page up and page down keys to move up and down the list view (and therefore the tree grid component).

## Why?

Currently page up and page down keys simply scroll the containing, which means you lose your position when navigating through the list view via keyboard. It could improve the usability when working with long list views via keyboard if we can allow the keys for page up and page down to move the focused item within the tree grid.

## How?

Add a page size to the tree grid, and move the focused row based on this when a user presses the page up or page down keys.

## To-do

* [ ] Determine the page size based on the visual height of the list view or tree grid instead of a hard-coded value
* [ ] Is this actually an improvement in terms of accessibility, or does it make it harder? I.e. what are the trade-offs here of bypassing browser defaults. The hope is that this makes the list view easier to use, but we should verify this assumption.
* [ ] Fix the list view's multi-select behaviour to support this. Currently it's unreliable.

Issue #49563, opened 2023-04-04
List View: Tweaks to improve usability, visual feedback, and keyboard behaviour


andrewserong:
The list view frequently comes up as a feature that folks both love, and also find frustrating or challenging to use. There are already a number of issues that address aspects of the list view's usability. The purpose of this issue is to track some of these issues, and provide a list of related issues that, grouped together, should hopefully help improve the overall usability of the list view.

For prioritised tasks for WordPress 6.6, see #60099

---

### Outstanding bugs to address

* [x] https://github.com/WordPress/gutenberg/issues/41613
* [x] https://github.com/WordPress/gutenberg/issues/42031
* [x] https://github.com/WordPress/gutenberg/pull/54900
* [x] https://github.com/WordPress/gutenberg/issues/56409
* [ ] https://github.com/WordPress/gutenberg/issues/56361
* [ ] https://github.com/WordPress/gutenberg/issues/56360
* [ ] https://github.com/WordPress/gutenberg/issues/60355

### Improving drag-and-drop and related visual issues

* [x] Come up with a solution for the issue where the drop indicator blends in with the color of selected blocks https://github.com/WordPress/gutenberg/issues/52917
* [x] Explore allowing dragging on a collapsed block to expand it: https://github.com/WordPress/gutenberg/issues/33684
* [ ] https://github.com/WordPress/gutenberg/issues/56539
* [ ] https://github.com/WordPress/gutenberg/issues/60242

### Improving interactions using the keyboard

* [x] Allow blocks in list view to be deselected by keyboard: https://github.com/WordPress/gutenberg/issues/48462
* [x] Allow keyboard shortcut to duplicate blocks to work in the list view
* [x] Allow CMD / CTRL + A to select all blocks
* [x] Allow CMD / CTRL + C to copy blocks
* [x] https://github.com/WordPress/gutenberg/issues/59936
* [x] Allow keyboard shortcuts for inserting a block before or after the currently focused block in the list view
* [ ] Allow Page Up / Page Down to select blocks in addition to the existing Home / End support
* [ ] When shift-selecting blocks nested within a selected parent block, it should be possible to update the selection (currently no change occurs as the parent block being selected causes all children to be treated as though they're part of the selection)

### Visual issues and tasks

* [x] https://github.com/WordPress/gutenberg/issues/46015
* [ ] https://github.com/WordPress/gutenberg/issues/53684
[...]

PR #62312, opened 2024-06-05
List View: Fix home and end key behaviour in very long lists

andrewserong:
## What?

Part of #49563 

Ensure that in very long lists, where the list view items are windowed (only rendered if they're visible) that the Home and End keys take the user to the very beginning or end of the list.

## Why?

In very long documents or templates, the windowing logic in the list view takes effect which means that only visible items in the list view are rendered. This means that when you press Home or End to go up or down the list, the logic that determines which item to focus finds the end of the _visible_ list instead of the end of the block tree. As a result, on `trunk` when you press End, you don't go to the end of the list, but instead only to the end of the current visible window.

The proposed fix is to always render the first and last items of the list, so that there are focusable items for the Home and End keys to go to. This way it's possible to immediately go to the very beginning or end of the list.

Note: Page Up and Page Down keys are not yet supported in the list view. I'm exploring that separately over in https://github.com/WordPress/gutenberg/pull/61474, though I'm unsure how much time I'll have to advance that one in the short-term.

## How?

In the list view's logic where it determines whether to render the current list view item as a "real" item, always render if it's the first or last item in the list.

## Testing Instructions

In a document with a very large number of blocks (e.g. > 100 blocks), open the list view, and then click a block in the list view twice to make sure that focus is within the list view.

Press Home and End keys on your keyboard. Notice that on `trunk` the Home and End keys don't always take you to the beginning or end of the document. With this PR applied, it should always take you to the beginning or end.

For a quick way to add a large number of blocks, I like to copy + paste text from a book in the public domain. E.g. [here's a copy of Dracula](https://www.gutenberg.org/files/345/345-h/345-h.htm).

### Before

https://github.com/WordPress/gutenberg/assets/14988353/c6d96760-0a0c-4391-a18b-15c386b06697

### After

https://github.com/WordPress/gutenberg/assets/14988353/c06afa26-267b-47e8-905d-bcd3b0083dbf

### DIFF
--- a/test/e2e/specs/editor/various/list-view.spec.js
+++ b/test/e2e/specs/editor/various/list-view.spec.js
@@ -1120,7 +1120,7 @@ test.describe( 'List View', () => {
 				'Pressing keyboard shortcut should also work when the menu is opened and focused'
 			)
 			.toMatchObject( [
-				{ name: 'core/paragraph', selected: true, focused: false },
+				{ name: 'core/paragraph', selected: true, focused: true },
 				{ name: 'core/file', selected: false, focused: false },
 			] );
 		await expect(
--- a/packages/block-editor/src/components/list-view/branch.js
+++ b/packages/block-editor/src/components/list-view/branch.js
@@ -194,10 +194,14 @@ function ListViewBranch( props ) {
 				// This prevents the entire tree from being rendered when a branch is
 				// selected, or a user selects all blocks, while still enabling scroll
 				// into view behavior when selecting a block or opening the list view.
+				// The first and last blocks of the list are always rendered, to ensure
+				// that Home and End keys work as expected.
 				const showBlock =
 					isDragged ||
 					blockInView ||
-					( isSelected && clientId === selectedClientIds[ 0 ] );
+					( isSelected && clientId === selectedClientIds[ 0 ] ) ||
+					index === 0 ||
+					index === blockCount - 1;
 				return (
 					<AsyncModeProvider key={ clientId } value={ ! isSelected }>
 						{ showBlock && (


andrewserong:
Just switching back to Draft while I look into test failures.

test/e2e/specs/editor/various/list-view.spec.js
```diff
@@ -1120,7 +1120,7 @@ test.describe( 'List View', () => {
 				'Pressing keyboard shortcut should also work when the menu is opened and focused'
 			)
 			.toMatchObject( [
-				{ name: 'core/paragraph', selected: true, focused: false },
+				{ name: 'core/paragraph', selected: true, focused: true },
```
andrewserong, line 1123: This needed to be updated because it turns out this PR appears to resolve a subtle bug:

When you open the block settings menu dropdown for the last item in the list view and then use the insert before keyboard shortcut, prior to this PR, there's a tiny moment when the last block in the list evaluates to `showBlock` being `false` while `blockInView` is being recalculated. That render then appears to mean that the focus in list view behaviour fails, and so focus isn't kept within the list view.

With this PR, we now always render the last item in the list, so in this test we now (correctly) have focus redirected to the inserted block within the list view, which is the expected behaviour.

I believe how that's all working now, but I've only done a light investigation.

andrewserong:
I've switched it back to "ready for review". I've added [an explanation](https://github.com/WordPress/gutenberg/pull/62312#discussion_r1627056234) to the updated e2e test — TL;DR, I believe this PR now fixes a bug with the insert before keyboard shortcut where sometimes the focus wouldn't be redirected back into the list view as expected.

ellatrix:

talldan (approved):
It works pretty well. For me when testing, the windowing logic is a little bit slow to update, definitely much slower than shown in the after video of the PR description:

https://github.com/WordPress/gutenberg/assets/677833/29cfcc0f-0038-4ed1-9930-cdd4d2f40336

That's with around 500 blocks (the demo post with all blocks duplicated a few times).

I wonder if there's anything that can be done to show something visual for any placeholder that's on screen. One of those grey blocky things. It could be a separate PR, no need to block this one.

andrewserong:
Thanks for reviewing @talldan! My test post only had 181 blocks in it, so I imagine the difference between our test posts is a contributing factor in the slowness there.

> I wonder if there's anything that can be done to show something visual for any placeholder that's on screen. One of those grey blocky things. It could be a separate PR, no need to block this one.

Another option I was wondering about could be to fade the list view items in one at a time when they render for real, instead of attempting to render them all at once. That way perhaps the rendering could be staggered and look a bit more natural? In any case, something to explore separately I reckon.

I'll merge this in and we can have a play with follow-up ideas separately 👍

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: a12600f46a

PR #62316, opened 2024-06-05
Performance reporter: print the stdout/stderr from the worker

jsnajdr:
Enhances the `PerformanceReporter` with `onStdOut` and `onStdErr` methods so that errors during the test run are reported even in CI where no other reporter is used.

Inspired by the `dot` reporter in Playwright: https://github.com/microsoft/playwright/blob/main/packages/playwright/src/reporters/dot.ts#L32-L42


Mamaduka (approved):
Thank you, @jsnajdr!

Issue #51373, opened 2023-06-09
Custom fields connected to Blocks


mtias:
> [!NOTE]  
> This is an integral part of the block API https://github.com/WordPress/gutenberg/issues/41236

---

At WCEU contributor day I brought up an idea to a few people on how we could connect blocks and custom fields in a compelling way. In the early days, we had explored a "meta" block attribute source that allowed block authors to create blocks consuming data from post meta. It worked alright but the behavior predated some of the latest block API affordances — like entity handling — and had the limitation of being accessible only for block developers. The other limitation was that we didn't have enough knowledge of how the final html output should render to recreate it on the server, so the PHP output was handled ad hoc on the blocks that consumed fields. This new approach can overcome both obstacles.

The way it'd work is by connecting attributes from normal static blocks (_content_ on paragraph, _url_ on image, and so on) to meta keys instead of serializing their content. So far so good. The crucial difference is that it doesn't require a new custom block with a custom attribute but leverages the built-in blocks as the primary interface. This means we can work out a UI in the block inspector that just needs to handle the connection but doesn't need to support editing nor extrinsic validation because the block itself takes care of it through its normal `setAttribute` flow. For example, updating a meta key for an image url would be as easy as dragging an image into the connected image block for a user. This also means patterns would just work out of the box transparently — a user can insert a pattern they like and map the heading block content to a field, the image source to another field, and so on. The edit UI for updating content remains the same.

To overcome the need to rebuild the html on the server we can employ the source declaration as a tag delimiter if we pair it with the new tag processor.

```json
"content": {
	"type": "string",
	"source": "html",
	"selector": "p",
	"default": "",
	"__experimentalRole": "content"
},
```

The gist of it is that the way a block declares how to query a serialized source can also function as an indication of how to reassemble that attribute properly. The implementation can thus reutilize an existing API contract instead of introducing a new syntax for tokens. (See #39831.)

## Proposal

- The `selector` functions as a token placement, so we don't need additional syntax. The tag processor can find the element and we reconstruct `innerHtml` entirely as it's just a simple composition of `<$selector>$field</$selector>`. This means we don't need the full block recipe from save because we are just replacing specific attributes.
- Since we are just dealing with block attributes, block variations can be created on demand for convenience.
- Transforming between blocks can retain the connection as long as they are valid transforms (so a paragraph using "book author"  meta can be turned into a heading using "book author" meta).
[...]

PR #62317, opened 2024-06-05
Fix showing double icons for connected blocks in pattern editor

kevin940726:
## What?

Continued from https://github.com/WordPress/gutenberg/pull/61560. Fix https://github.com/WordPress/gutenberg/pull/61560#issuecomment-2142432095.

Prevent double icons from showing when editing the synced pattern itself.

## Why?

Showing double icons is confusing to the users.

## How?

Only show the binding indicator when editing an overridable pattern as suggested in the issue.

## Testing Instructions

1. Create a synced pattern with overridable blocks (follow the testing instructions in https://github.com/WordPress/gutenberg/pull/61560)
2. Expect there to be no double icons when editing the pattern itself in the block toolbar.
3. Expect to see the binding indicator icon when editing the overridable pattern instance.

## Screenshots or screencast 
Before:
image

After:
image


packages/block-editor/src/components/block-toolbar/index.js
```diff
@@ -167,9 +175,11 @@ export function PrivateBlockToolbar( {
...
 					isLargeViewport &&
 					isDefaultEditingMode && <BlockParentSelector /> }
-				{ isUsingBindings && canBindBlock( blockName ) && (
-					<BlockBindingsIndicator clientIds={ blockClientIds } />
-				) }
+				{ isUsingBindings &&
+					hasPatternOverrides &&
+					canBindBlock( blockName ) && (
+						<BlockBindingsIndicator clientIds={ blockClientIds } />
+					) }
```
talldan: I think the `BlockBindingsIndicator` should be visible when using other types of bindings (e.g. meta) outside of the pattern block, so by adding `hasPatternOverrides` to the condition does it prevent that?

kevin940726: Good point. I just pushed [5ebaab8](https://github.com/WordPress/gutenberg/pull/62317/commits/5ebaab83bd2ed4d59368eb39790e08aa8f54d44c). It'll show double icons for post meta but I guess we need that for now?

image

talldan: Hmm, now I'm not sure what's expected for bindings. It should be one block icon, and I guess a purple one?

I'm not completely up to date on the block bindings UI changes, so would be good to get some confirmation.

cc @richtabor @SantosGuillamot

SantosGuillamot: Until now, the purple icon was just a connection icon.

Screenshot 2024-06-05 at 16 53 08

I can see a duplicated icon in `trunk` now, I guess since the linked pull request. I would personally keep the current UX showing the connection icon rather than the duplicated one, and figure out later how to keep evolving the UX based on feedback. But it'd be great to hear from @richtabor and @jasmussen .

jasmussen: I prefer the after here, honestly. We've tried out the connection icon in this iteration in trunk for a bit, but it hasn't felt like it added clarity. I think there are other things we can do, but for now, I'd omit it.

SantosGuillamot: Okay, thanks for the feedback. Let me try to clarify to ensure we are aligned. For connections that are NOT pattern overrides, like post meta, there are three options so far:

**Option 1: The connection icon (the old implementation)**

Screenshot 2024-06-05 at 16 53 08

**Option 2: The duplicated icon (the existing implementation in trunk)**

image

**Option 3: Removing the icon (this would be new)**

image

---
You refer to option 3, right?

jasmussen: Yes, but I'm worried I'm missing nuance. @richtabor's original intent, as [outlined here](https://github.com/WordPress/gutenberg/pull/61560#issuecomment-2142427551), was that _the block icon was purple_. The main issue isn't that it's purple or not—I could go either way on that—it's that the icon is duplicated, it shouldn't be.

SantosGuillamot: I agree that the duplicated icon doesn't seem right.

I'm struggling to understand what is the desired outcome, though. Let me try to break it into three different use cases

**1. Pattern overrides when used in a post**

Right now in `trunk`, I can see the purple paragraph icon and when I click on it, it triggers the message. I assume this is the expected behavior, right?

Screenshot 2024-06-06 at 10 44 05

**2. Editing a pattern override**

Right now in `trunk`, I can see the duplicated icon, which I guess it is what this pull request tries to solve.

Screenshot 2024-06-06 at 10 47 01

What is the desired behavior here? Just remove the purple icon or keep the purple icon and remove the "normal" one? If it is the second one, what should happen when users click on the icon? Should it still trigger the modal to transform the block?

**3. Using other bindings like post meta**

Right now in `trunk`, I can also see the duplicated icon.

I assume that whatever is decided for the second use case, "Editing a pattern override", should apply here as well?

jasmussen: I'll let @richtabor respond as he drove some of the inital designs. For me, if the paragraph is connected, the block icon is purple because of that, but there is only ever a single block icon. If that shows a tooltip in the transforms dialog in one case, then it can show that same tooltip in both. If that replaces the transforms dropdown that is otherwise accessible in that space, then that's the main question to answer: should I not be able to transform blocks when they are connected?

kevin940726:
I removed the backport label until we reach a design decision, and I'll add it back when it's merged.

packages/block-editor/src/components/block-toolbar/index.js
SantosGuillamot: > should I not be able to transform blocks when they are connected?

In my opinion, we should definitely allow transforming blocks when they are connected. One of the benefits of using bindings is that users can transform a connected paragraph to a connected heading the same way they transform blocks right now, while keeping the connection.

Actually, it was even mentioned in the original proposal as one of the points: [link](https://github.com/WordPress/gutenberg/issues/51373).

richtabor: I would expect the purple block icon, if connected. It denotes the value is special, coming from somewhere else. 

Perhaps the popover help copy can go below transforms, if there are transforms.

richtabor: Overrides and connections should be essentially the same.

ellatrix:

packages/block-editor/src/components/block-toolbar/index.js
SantosGuillamot: Okay. Thanks for the feedback. So, if I understand you correctly, is this the desired behavior for an initial version until we gather more feedback? (Changing the texts I'm using maybe)

**1. Pattern overrides when used in a post**

Screenshot 2024-06-06 at 10 44 05

**2. Editing a pattern override**

Screenshot 2024-06-07 at 10 35 13

**3. Using other bindings like post meta**

Screenshot 2024-06-07 at 10 34 06

jasmussen: That looks right to me, yes. The "This block is connected" message can be integrated into the transforms dropdown if need be.

SantosGuillamot: Perfect 🙂 @kevin940726 Let us know if it makes sense, if you plan to work on that, and if you need any help!

kevin940726: Thanks! I'm working on it now. Will push a commit soon!

kevin940726: I just pushed [c862489](https://github.com/WordPress/gutenberg/pull/62317/commits/c8624896c8513847662990ba3950aa2daae8d6a6). It now has the following UI:

Post meta connected block:
image

Pattern overrides connected block in focus mode:
image

Pattern overrides connected block in post:
image

LMK if I miss something!

SantosGuillamot: Regarding the UX, I believe it is much better now! Thanks for working on it. I'll take a look at the code asap.

packages/block-editor/src/components/block-bindings-toolbar-indicator/index.js
```diff
@@ -22,13 +22,10 @@ import useBlockDisplayTitle from '../block-title/use-block-display-title';
 
 export default function BlockBindingsToolbarIndicator( { clientIds } ) {
```
SantosGuillamot: If I am not mistaken, this component becomes specific to pattern overrides because it is only used when adding a pattern in a post, right? Maybe we can rename it and slightly change the implementation to reflect that?

packages/block-editor/src/components/block-switcher/index.js
```diff
@@ -118,6 +121,17 @@ function BlockSwitcherDropdownMenuContents( {
 			</p>
 		);
 	}
```
SantosGuillamot, line 123: I'm wondering if we could get `isUsingBindings` from this `useSelect`. Something similar to what is being done [here](https://github.com/WordPress/gutenberg/pull/62317/files#diff-9ac2a4466e7c0a85aaff0b40c5d7bc64c894132528ca3c879db4bebb438617dcR99-R100).

If we do this, I believe we could avoid passing `isUsingBindings` to `BlockSwitcher` and `BlockSwitcherDropdownMenuContents` components.

Additionally, I think we wouldn't need `isUsingBindings` in the block toolbar anymore ([here](https://github.com/WordPress/gutenberg/pull/62317/files#diff-9ac2a4466e7c0a85aaff0b40c5d7bc64c894132528ca3c879db4bebb438617dcR179-R183)) and we could just have something like `hasPatternOverrides`, as it is now specific to pattern overrides, right?

I am not 100% sure if it makes sense, but maybe it is worth taking a look.

kevin940726: I believe so, yes! However I don't think it makes much difference. Currently we still need `isUsingBindings` in block toolbar to apply the `is-connected` class name. We can somehow move the class name to block switcher, but in terms of performance I don't think it matters that much. Do you think it's worth exploring?

packages/block-editor/src/components/block-bindings-toolbar-indicator/index.js
kevin940726: On second thought, I don't think this even belongs in the `block-editor` package 😅. I'd imagine it'll be a bigger refactor if we figure out a way to move it to the `patterns` package though. I'd prefer to do it in a follow-up PR if necessary. WDYT?

SantosGuillamot: Moving it to the `patterns` package seems logical to me 🙂 And I guess it is okay to do it in a follow-up PR. Should we create an issue or something to ensure it is not lost?

Maybe it is still worth changing the component name to clarify it is related to patterns and not block bindings until that PR is done?

packages/block-editor/src/components/block-switcher/index.js
SantosGuillamot: I wasn't thinking about performance but about readability, but I don't have a strong opinion, so whatever you decide it's okay for me 🙂

kevin940726: I think it reads okay for me, but I'm obviously biased 😅. Happy to do some further improvements if we find something confusing later! 🙇

packages/block-editor/src/components/block-bindings-toolbar-indicator/index.js
kevin940726: Done in [72d0a8e](https://github.com/WordPress/gutenberg/pull/62317/commits/72d0a8e32cbb9086451604807281f18efdfb78b0)! I renamed some stuff, so could appreciate a final review to ensure I didn't break anything 😆.

SantosGuillamot (approved):
I've tested it, and everything seems to be working as expected. Thanks for working on it!

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: a052b4cf3e

PR #49807, opened 2023-04-13
Do not add unregistered style variations to the theme.json schema

oandregal:
Follow-up to https://github.com/WordPress/gutenberg/pull/46343

## What?

This PR makes sure unregistered block style variations are not part of the internal `theme.json` schema, so they are ignored before processing any data coming from `theme.json`.

## Why?

- The current state prevents the site editor from loading when an unregistered style variation is found in a `theme.json` dataset. 
- It also enqueues CSS rules to the front-end that don't have a selector, so they are invalid.

## How?

By making sure the style variation is not added to the allowed list, unless it's registered for that particular block.

## Testing Instructions

- Make sure your environment logs the errors to the screen by setting `WP_DEBUG` to true.
- Add the following two variations to the `theme.json` of TwentyTwentyThree:

```json
{
  "styles": {
    "blocks": {
      "core/quote": {
        "variations": {
          "plain": { "color": { "background": "indigo", "text": "gold" } },
          "unregistered": { "color": { "background": "red", "text": "blue" } },
        } 
      }
    }
  }
}
```

- Create a post that contains a quote and style it with the plain variation.

The expected result is that:

- The styles are applied to quote blocks with plain style variation in editors and front.
- The site editor loads properly.
- The front-end doesn't have a style rule such as ` { background-color: red; color: blue}` (note the lack of CSS selector).
- There is no error reported in `class-wp-theme-json-gutenberg.php`. In `trunk`, you'll see something like this:

image

Verify that upon updating the colors of the style variation via the global styles sidebar (blocks > quote > style variation plain), the user-provided colors are rendered everywhere (front & editors).

Note that by adding an unregistered variation, the core class `class-wp-theme-json.php` will still throw an error. This is expected. It'll be fixed when this PR is backported to core.

PR #62321, opened 2024-06-05
Global Styles: Prevent loss of custom variation data in API response

### LINKED ISSUE #62303: Section styles: editing a block variation in Styles doesn't reflect in the editor
While recording a demo of section styles with 6.6 beta 1, I noticed that while I could change the variation in Styles, the changes would update on the front end of the site but would revert in the editor after hitting save. Here's a video showing front end vs back end:

https://github.com/WordPress/gutenberg/assets/26996883/a68230ef-d48c-4bbe-81cb-f183b1d8abd1

Here's the demo video I was working on for good measure:

https://github.com/WordPress/gutenberg/assets/26996883/0465c244-4498-4c94-bcad-40654bfbc989

Here's the theme.json:

```
	"styles": {
		"blocks": {
			"variations": {
				"Dark": {
					"blockTypes": [
						"core/group",
						"core/columns"
					],
					"color": {
						"background": "#1e1e1e",
						"text": "#FFF"
					},
					"elements": {
						"heading": {
							"color": {
								"text": "#FFF"
							}
						}
					}
				}
			}
		}
	}
```

I am assuming this is likely because I'm _adding_ an element style to the variation (button in this case) but that doesn't explain why the Heading doesn't update. This could very well be user error but wanted to report just in case cc @aaronrobertshaw

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/62303

## What?

Ensures custom block style variation data isn't omitted from the global styles API response when saving styles in the site editor.

## Why?

When saving global styles data without this fix, any tweaks to custom block style variations are omitted from the global styles endpoint response. This missing data makes it appear as though the tweaks have been lost until the editor is reloaded.

## How?

Within the `prepare_item_for_response` method of the global styles controller, a fresh theme json instance is created. For such a request this never triggers the theme json data filters that custom block style variations rely on to register custom block styles. Unregistered block style variations get stripped out during theme json sanitization.

Rather than do all the heavy lifting of parsing theme.json partial files for the theme, applying data filters etc. this PR registers the custom block style variations found in the global styles post, directly. 

## Testing Instructions

There are some [duotone support actions](https://github.com/WordPress/gutenberg/blob/trunk/lib/block-supports/duotone.php#L21-L22) in Gutenberg that trigger all the theme json filters. So to test this fix fully, those actions need to be temporarily disabled.

1. Comment out the [duotone support actions](https://github.com/WordPress/gutenberg/blob/trunk/lib/block-supports/duotone.php#L21-L22) that masked this issue
2. Add a custom block style variation to your theme.json (see #62303 for snippet)
3. Navigate to Appearance > Editor 
4. Select a group block and apply the custom variation
5. Navigate to Styles > Blocks > Group > dark (variation name)
6. Change the background color for the variation and confirm it is reflected in the editor
7. Save the changes and confirm the background color is still correct in the editor
8. For good measure check the frontend
9. Uncomment the duotone support actions from step 1 and repeat the process. It should still work.

## Screenshots or screencast 

| Before | After |
|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/60436221/a7d59fad-29a1-40d1-84c7-9606f170aef5" /> | <video src="https://github.com/WordPress/gutenberg/assets/60436221/3ad6e58b-d05b-4048-8e35-79987ed1808c" /> |

### DIFF
--- a/lib/class-wp-rest-global-styles-controller-gutenberg.php
+++ b/lib/class-wp-rest-global-styles-controller-gutenberg.php
@@ -358,6 +358,7 @@ class WP_REST_Global_Styles_Controller_Gutenberg extends WP_REST_Controller {
 	 * @since 5.9.0
 	 * @since 6.2.0 Handling of style.css was added to WP_Theme_JSON.
 	 * @since 6.6.0 Added custom relative theme file URIs to `_links`.
+	 * @since 6.6.0 Added registration of custom block style variations to prevent them being stripped in response.
 	 *
 	 * @param WP_Post         $post    Global Styles post object.
 	 * @param WP_REST_Request $request Request object.
@@ -369,6 +370,29 @@ class WP_REST_Global_Styles_Controller_Gutenberg extends WP_REST_Controller {
 		$config                           = array();
 		$theme_json                       = null;
 		if ( $is_global_styles_user_theme_json ) {
+			if ( isset( $raw_config['styles']['blocks'] ) && is_array( $raw_config['styles']['blocks'] ) ) {
+				// Ensure custom block style variations are registered so their data isn't stripped during sanitization in WP_Theme_JSON.
+				$registry = WP_Block_Styles_Registry::get_instance();
+				foreach ( $raw_config['styles']['blocks'] as $block_name => $block_data ) {
+					if ( isset( $block_data['variations'] ) && is_array( $block_data['variations'] ) ) {
+						$registered_styles = $registry->get_registered_styles_for_block( $block_name );
+
+						foreach ( $block_data['variations'] as $variation_name => $variation_data ) {
+							// Register block style variation only if it hasn't already been registered.
+							if ( ! array_key_exists( $variation_name, $registered_styles ) ) {
+								register_block_style(
+									$block_name,
+									array(
+										'name'  => $variation_name,
+										'label' => $variation_name,
+									)
+								);
+							}
+						}
+					}
+				}
+			}
+
 			$theme_json = new WP_Theme_JSON_Gutenberg( $raw_config, 'custom' );
 			$config     = $theme_json->get_raw_data();
 		}
--- a/lib/experimental/kses.php
+++ b/lib/experimental/kses.php
@@ -31,6 +31,29 @@ function gutenberg_filter_global_styles_post( $data ) {
 	) {
 		unset( $decoded_data['isGlobalStylesUserThemeJSON'] );
 
+		// Ensure block style variations are automatically registered so they will not
+		// be omitted during theme.json sanitization.
+		if ( isset( $decoded_data['styles']['blocks'] ) && is_array( $decoded_data['styles']['blocks'] ) ) {
+			$registry = WP_Block_Styles_Registry::get_instance();
+			foreach ( $decoded_data['styles']['blocks'] as $block_name => $block_data ) {
+				if ( isset( $block_data['variations'] ) && is_array( $block_data['variations'] ) ) {
+					$registered_styles = $registry->get_registered_styles_for_block( $block_name );
+
+					foreach ( $block_data['variations'] as $variation_name => $variation_data ) {
+						// Register block style variation only if it hasn't already been registered.
+						if ( ! array_key_exists( $variation_name, $registered_styles ) ) {
+							register_block_style(
+								$block_name,
+								array(
+									'name'  => $variation_name,
+									'label' => $variation_name,
+								)
+							);
+						}
+					}
+				}
+			}
+		}
 		$data_to_encode = WP_Theme_JSON_Gutenberg::remove_insecure_properties( $decoded_data );
 
 		$data_to_encode['isGlobalStylesUserThemeJSON'] = true;
--- a/phpunit/class-wp-rest-global-styles-controller-gutenberg-test.php
+++ b/phpunit/class-wp-rest-global-styles-controller-gutenberg-test.php
@@ -514,6 +514,39 @@ class WP_REST_Global_Styles_Controller_Gutenberg_Test extends WP_Test_REST_Contr
 	}
 
 	/**
+	 * @covers WP_REST_Global_Styles_Controller_Gutenberg::update_item
+	 */
+	public function test_update_item_with_custom_block_style_variations() {
+		wp_set_current_user( self::$admin_id );
+		if ( is_multisite() ) {
+			grant_super_admin( self::$admin_id );
+		}
+		$variations = array(
+			'dark' => array(
+				'color' => array(
+					'background' => '#000000',
+					'text'       => '#ffffff',
+				),
+			),
+		);
+		$request    = new WP_REST_Request( 'PUT', '/wp/v2/global-styles/' . self::$global_styles_id );
+		$request->set_body_params(
+			array(
+				'styles' => array(
+					'blocks' => array(
+						'core/group' => array(
+							'variations' => $variations,
+						),
+					),
+				),
+			)
+		);
+		$response = rest_get_server()->dispatch( $request );
+		$data     = $response->get_data();
+		$this->assertSame( $variations, $data['styles']['blocks']['core/group']['variations'] );
+	}
+
+	/**
 	 * @doesNotPerformAssertions
 	 */
 	public function test_delete_item() {


aaronrobertshaw:
I did start on a unit test for this but got stuck with the global styles post not being updated and therefore getting unexpected/missing styles data in the test only. I'll try and sort this out tomorrow.

<details>
<summary>Snippet for unit test</summary>

```diff
diff --git a/phpunit/class-wp-rest-global-styles-controller-gutenberg-test.php b/phpunit/class-wp-rest-global-styles-controller-gutenberg-test.php
index f5b216a084..ff5b291b2f 100644
--- a/phpunit/class-wp-rest-global-styles-controller-gutenberg-test.php
+++ b/phpunit/class-wp-rest-global-styles-controller-gutenberg-test.php
@@ -513,6 +513,39 @@ class WP_REST_Global_Styles_Controller_Gutenberg_Test extends WP_Test_REST_Contr
 		$this->assertErrorResponse( 'rest_custom_css_illegal_markup', $response, 400 );
 	}
 
+	/**
+	 * @covers WP_REST_Global_Styles_Controller_Gutenberg::update_item
+	 */
+	public function test_update_item_with_custom_block_style_variations() {
+		wp_set_current_user( self::$admin_id );
+		if ( is_multisite() ) {
+			grant_super_admin( self::$admin_id );
+		}
+		$variations = array(
+			'dark' => array(
+				'color' => array(
+					'background' => '#000000',
+					'text'       => '#ffffff',
+				),
+			),
+		);
+		$request    = new WP_REST_Request( 'PUT', '/wp/v2/global-styles/' . self::$global_styles_id );
+		$request->set_body_params(
+			array(
+				'styles' => array(
+					'blocks' => array(
+						'core/group' => array(
+							'variations' => $variations,
+						),
+					),
+				),
+			)
+		);
+		$response = rest_get_server()->dispatch( $request );
+		$data     = $response->get_data();
+		$this->assertSame( $variations, $data['styles']['blocks']['core/group']['variations'] );
+	}
+
 	/**
 	 * @doesNotPerformAssertions
 	 */

```
</details>

lib/class-wp-rest-global-styles-controller-gutenberg.php
```diff
@@ -369,6 +370,29 @@ public function prepare_item_for_response( $post, $request ) { // phpcs:ignore V
...
+				// Ensure custom block style variations are registered so their data isn't stripped during sanitization in WP_Theme_JSON.
+				$registry = WP_Block_Styles_Registry::get_instance();
+				foreach ( $raw_config['styles']['blocks'] as $block_name => $block_data ) {
+					if ( isset( $block_data['variations'] ) && is_array( $block_data['variations'] ) ) {
+						$registered_styles = $registry->get_registered_styles_for_block( $block_name );
+
+						foreach ( $block_data['variations'] as $variation_name => $variation_data ) {
+							// Register block style variation only if it hasn't already been registered.
+							if ( ! array_key_exists( $variation_name, $registered_styles ) ) {
+								register_block_style(
```
ramonjd, line 383: Should this be `gutenberg_register_block_style` for backwards compat? E.g., a plugin running 6.5 or lower?

lib/class-wp-rest-global-styles-controller-gutenberg.php
```diff
@@ -369,6 +370,29 @@ public function prepare_item_for_response( $post, $request ) { // phpcs:ignore V
 		$config                           = array();
 		$theme_json                       = null;
 		if ( $is_global_styles_user_theme_json ) {
+			if ( isset( $raw_config['styles']['blocks'] ) && is_array( $raw_config['styles']['blocks'] ) ) {
```
ramonjd, line 373: Does this need to run for updates? If not we could check for `WP_REST_Server::EDITABLE === $request->get_method()` or something along those lines

ramonjd (commented):
I made some videos!

Things are working well for me. I tested using the provided snippet and modified Group and Column variations.

| Before  | After |
| <video src="https://github.com/WordPress/gutenberg/assets/6458278/a95e7189-2efe-4faf-9683-7b32a34cbbd6" /> | <video src="https://github.com/WordPress/gutenberg/assets/6458278/c89c319d-fed2-4e98-aed1-be796ae64ce3" />  |


I also tested loading and saving global styles revisions that have block variations. 

The styles are loaded as expected.

> I did start on a unit test for this but got stuck with the global styles post not being updated and therefore getting unexpected/missing styles data in the test only.

It seems to be something specific to the `variations` key. Throwing a general block style at it works.

~~The styles are stripped either before or after updating the post~~. 

The [incoming data](https://github.com/WordPress/gutenberg/blob/1380687423b646818ce22345031ce3a1e28e458f/lib/class-wp-rest-global-styles-controller-gutenberg.php#L278-L278) to `wp_update_post` looks sound in that it contains the `variation` array.

However what `get_post` [returns](https://github.com/WordPress/gutenberg/blob/1380687423b646818ce22345031ce3a1e28e458f/lib/class-wp-rest-global-styles-controller-gutenberg.php#L283-L283) is not. `$post->post_content` is `{"version":3,"isGlobalStylesUserThemeJSON":true}`

I'm not sure why yet 😄

lib/class-wp-rest-global-styles-controller-gutenberg.php
aaronrobertshaw: This call isn't passing an array of block names or needing to register style object data along with the block style so we can simply call the standard `register_block_style` function. 

It should work now and into the future without a refactor once `gutenberg_register_block_style` eventually gets deprecated.

lib/class-wp-rest-global-styles-controller-gutenberg.php
aaronrobertshaw: I initially looked to do something similar however if this doesn't run then the variations data is stripped if a consumer is calling `get_item`.

aaronrobertshaw:
Thanks for digging into the test issue @ramonjd 

> The [incoming data](https://github.com/WordPress/gutenberg/blob/1380687423b646818ce22345031ce3a1e28e458f/lib/class-wp-rest-global-styles-controller-gutenberg.php#L278-L278) to wp_update_post looks sound in that it contains the variation array.
> 
> However what get_post [returns](https://github.com/WordPress/gutenberg/blob/1380687423b646818ce22345031ce3a1e28e458f/lib/class-wp-rest-global-styles-controller-gutenberg.php#L283-L283) is not. $post->post_content is {"version":3,"isGlobalStylesUserThemeJSON":true}

This is what I was seeing as well. I also tried creating an entirely separate post and testing the results of `get_item` but again there the variations data was missing. I can't work out what's going on just yet.

Do you think this fix is worth landing and following up on the test separately? 

Maybe not as I can't see the backport for the fix being accepted without the test. My concern was more towards preventing this from blocking people testing the section styles feature in the 6.6 beta.

lib/class-wp-rest-global-styles-controller-gutenberg.php
ramonjd: Ah, so if we don't allow the code to run for updates then it'll affect `get_item()`?

I meant whether it should only run for GET request (and not updates or anything else), so something like this is probably more appropriate: `WP_REST_Server::READABLE === $request->get_method()`

I tested by calling an item — `await wp.apiFetch( { path: '/wp/v2/global-styles/11' } );` — and could still see my variations. Very likely I'm not grokking it.

ramonjd: Never mind, I see it doesn't work in the site editor. That's weird 🤔

andrewserong:
I don't have anything helpful to add, unfortunately, but just confirming that in the test environment when logging out the `$post` in the following line, I'm seeing what you're both seeing (the variations data was there prior to calling `wp_update_post`, but isn't in the response from `get_post( $request['id'] )`).

https://github.com/WordPress/gutenberg/blob/1380687423b646818ce22345031ce3a1e28e458f/lib/class-wp-rest-global-styles-controller-gutenberg.php#L283

It's very confusing, I wonder if there's a hook running that's doing an additional round of sanitization? 🤔
I've run out of time for the week to keep testing, but happy to help out next week if this is still open!

ramonjd:
> This is what I was seeing as well. I also tried creating an entirely separate post and testing the results of get_item but again there the variations data was missing. I can't work out what's going on just yet.

Very interesting.

Fetching this endpoint — `await wp.apiFetch( { path: '/wp/v2/global-styles/11' } );` — calls the controller's method `get_item` directly. In the response I'm seeing the variations with the correct values after I've updated theme in the Site Editor.


Screenshot 2024-06-07 at 2 14 42 PM


Same goes for the POST request when updating them.

So why doesn't the test do this 😆 

Must be missing something.

> Do you think this fix is worth landing and following up on the test separately?

I agree it would be good to work out why the tests fail first,  in case it's hiding a bigger issue.

aaronrobertshaw:
It looks like the variations are being filtered due to the content save filters added in [Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/lib/experimental/kses.php#L54-L57) and [Core](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/kses.php#L2226).

My hunch is these filters get called before we have other block style variations registered via the theme.json data filters. So the variations are sanitized by the WP_Theme_JSON class. Basically, the same issue this PR was trying to solve for the update endpoint.

ramonjd:
> It looks like the variations are being filtered due to the content save filters added in [Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/lib/experimental/kses.php?rgh-link-date=2024-06-07T04%3A43%3A26Z#L54-L57) and [Core](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/kses.php?rgh-link-date=2024-06-07T04%3A43%3A26Z#L2226).

Great sleuthing. Another piece of the puzzle falls into place... 😄

aaronrobertshaw:
I've updated the kses content save filter and the test is passing. From a quick test of the editor it appears the original change in the global styles controller is still required.

This all seems to point to the sanitization of variations no longer being useful. Long term, I believe the plan is to eventually allow the creation of block style variations via the Global Styles UI. At that point we'll probably need to remove the sanitization of variations too. 

I'll leave this PR open for the moment in case we do need a quick fix for the WP6.6 beta. In the meantime, I'll work through what's required to remove the validation of variations without reintroducing the issues https://github.com/WordPress/gutenberg/pull/49807 intended to solve.

aaronrobertshaw:
Closing in favour of https://github.com/WordPress/gutenberg/pull/62405

PR #62322, opened 2024-06-05
Data: Add error handle to the 'registry.batch' method

### LINKED ISSUE #62311: Data: Stores become unresponsive when batched actions trigger an error
### Description

Possible solution: change from `catch` statement to `finally` statement in this line https://github.com/WordPress/gutenberg/blob/trunk/packages/core-data/src/resolvers.js#L298

Execute `useEntityRecords('root', 'site')` inside the edit function of a block.

_No response_

### Environment info

WordPress 6.6 beta 1
Twenty-Twenty Four
Chrome
Desktop with Windows 11

Yes

Yes

Mamaduka:
## What?
Resolves: #62311.

PR adds an error handler to the `registry.batch` method and avoids pausing all store emitters during the error.

## Testing Instructions
1. Open a post or page.
2. Run selector via Console - `wp.data.select( 'core' ).getEntityRecords( 'root', 'site' );`.
3. Confirm that the editor is still responsive.

Same.

### DIFF
--- a/packages/data/src/registry.js
+++ b/packages/data/src/registry.js
@@ -322,9 +322,21 @@ export function createRegistry( storeConfigs = {}, parent = null ) {
 
 		emitter.pause();
 		Object.values( stores ).forEach( ( store ) => store.emitter.pause() );
-		callback();
-		emitter.resume();
-		Object.values( stores ).forEach( ( store ) => store.emitter.resume() );
+		try {
+			callback();
+			emitter.resume();
+			Object.values( stores ).forEach( ( store ) =>
+				store.emitter.resume()
+			);
+		} catch ( error ) {
+			emitter.resume();
+			Object.values( stores ).forEach( ( store ) =>
+				store.emitter.resume()
+			);
+
+			// Re-throw the error after resuming the work for debugging.
+			throw error;
+		}
 	}
 
 	let registry = {
--- a/packages/data/src/registry.js
+++ b/packages/data/src/registry.js
@@ -324,18 +324,11 @@ export function createRegistry( storeConfigs = {}, parent = null ) {
 		Object.values( stores ).forEach( ( store ) => store.emitter.pause() );
 		try {
 			callback();
+		} finally {
 			emitter.resume();
 			Object.values( stores ).forEach( ( store ) =>
 				store.emitter.resume()
 			);
-		} catch ( error ) {
-			emitter.resume();
-			Object.values( stores ).forEach( ( store ) =>
-				store.emitter.resume()
-			);
-
-			// Re-throw the error after resuming the work for debugging.
-			throw error;
 		}
 	}
 
--- a/packages/data/src/test/registry.js
+++ b/packages/data/src/test/registry.js
@@ -755,6 +755,28 @@ describe( 'createRegistry', () => {
 			} );
 			expect( listener ).toHaveBeenCalledTimes( 1 );
 		} );
+
+		it( 'should handle errors', () => {
+			const store = registry.registerStore( 'myAwesomeReducer', {
+				reducer: ( state = 0 ) => state + 1,
+			} );
+			const listener = jest.fn();
+			const error = new Error( 'Whoops' );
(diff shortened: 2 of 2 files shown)


packages/data/src/registry.js
```diff
@@ -322,9 +322,21 @@ export function createRegistry( storeConfigs = {}, parent = null ) {
...
+			);
+		} catch ( error ) {
+			emitter.resume();
+			Object.values( stores ).forEach( ( store ) =>
+				store.emitter.resume()
+			);
+
+			// Re-throw the error after resuming the work for debugging.
+			throw error;
+		}
```
Mamaduka: I guess, I can do  `try...finally` here and avoid re-throwing the error.

jsnajdr, line 332: Why not `try { callback(); } finally { emitter.resume(); ... }`? That does the same thing, including throwing the error.

Mamaduka: Yup. I came to  the same conclusion - https://github.com/WordPress/gutenberg/pull/62322#discussion_r1627303033.

jsnajdr (approved):
:shipit:

tyxla (approved):
Nice, thanks for tightening it up @Mamaduka  👍 

Could this also use a unit test?

Mamaduka:
> Could this also use a unit test?

That would be great. I tried to add unit tests, but it's been a while since I've worked on store tests and properly emulated the reducer's error. If you have any pointers, I would be happy to take it from there.

packages/data/src/registry.js
```diff
@@ -322,9 +322,14 @@ export function createRegistry( storeConfigs = {}, parent = null ) {
 
 		emitter.pause();
 		Object.values( stores ).forEach( ( store ) => store.emitter.pause() );
-		callback();
-		emitter.resume();
-		Object.values( stores ).forEach( ( store ) => store.emitter.resume() );
+		try {
```
tyxla, line 325: Should we also wrap with `try/catch/finally` the `callback()` that we run if we're already batching above (L319)?

tyxla:
> That would be great. I tried to add unit tests, but it's been a while since I've worked on store tests and properly emulated the reducer's error. If you have any pointers, I would be happy to take it from there.

Could be something straightforward, like simulating an error by an unallowed operation inside the reducer. For example, 

```
const store = registry.registerStore( 'myAwesomeReducer', {
	reducer: ( state = Object.seal( {} ) ) => {
		state.test = 5;
		return state;
	},
} );

registry.batch( () => {
	store.dispatch( { type: 'dummy' } );
} );
```

This should simulate a `TypeError` inside the reducer since you can't add extra properties to a sealed object.

tyxla:
On second thought, I'm not sure the `try` you're adding will catch that error, because the error won't be thrown in that same callback function call, but rather by Redux's reducer function...

packages/data/src/registry.js
Mamaduka: L319 returns early and doesn't touch emitters. Bubbling the errors should be okay there.

Mamaduka:
> because the error won't be thrown in that same callback function call, but rather by Redux's reducer function

You put it better in words. I think I had the same problem when trying initially.

packages/data/src/test/registry.js
```diff
@@ -755,6 +755,28 @@ describe( 'createRegistry', () => {
...
+					throw error;
+				} );
+			} ).toThrow( error );
+			expect( listener ).not.toHaveBeenCalled();
+
+			registry.batch( () => {
+				store.dispatch( { type: 'dummy' } );
+				store.dispatch( { type: 'dummy' } );
+			} );
+			expect( listener ).toHaveBeenCalledTimes( 1 );
```
Mamaduka, line 778: @tyxla, what do you think about this?

tyxla: Looks good to me - thanks for adding a test!

ellatrix:

PR #61357, opened 2024-05-03
Update: Implement the new discussion panel design.

jorgefilipecosta:
Implements the new discussion panel design proposed at: https://github.com/WordPress/gutenberg/issues/59689#issuecomment-2090763388

cc: @jameskoster 

## Screenshots

Screenshot 2024-05-03 at 13 48 35

## Testing Instructions

Tested the discussion panel options and verified they work as expected and are correctly saved.

PR #62323, opened 2024-06-05
Editor: Make revisions more prominent

### LINKED ISSUE #62041: Unified publish flow: return revisions to more prominence
In testing 6.6, I noticed that revisions is now a bit more hidden and doesn't include the number of revisions available after this PR was merged: https://github.com/WordPress/gutenberg/pull/61867

**Before:**

https://github.com/WordPress/gutenberg/assets/26996883/d22a1a4d-620f-45f7-835c-bbe6d6771d2e

**After:**

https://github.com/WordPress/gutenberg/assets/26996883/90336b83-1a34-41bc-97f2-45cbe531795b

In talking with enterprise folks, I imagine this might cause some panic and confusion. Revisions are a key part of the foundational experience of what WordPress offers and, when many folks are working on a piece, having easy access is important. Opening this issue partially to see if feedback comes up within the 6.6 cycle and to raise an initial flag.

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/62041

There is some feedback in the issue that the revisions link and revisions number should be a bit more prominent, besides the post action in the ellipsis menu. This PR adds this info in post summary like suggested in the below [comment](https://github.com/WordPress/gutenberg/pull/62323#issuecomment-2149329070).

## Testing Instructions
1. Observe the revisions info in the post summary panel.

## Screenshots or screencast 

Screenshot 2024-06-06 at 9 49 28 AM

### DIFF
--- a/packages/editor/src/components/post-last-revision/index.js
+++ b/packages/editor/src/components/post-last-revision/index.js
@@ -11,19 +11,11 @@ import { addQueryArgs } from '@wordpress/url';
  * Internal dependencies
  */
 import PostLastRevisionCheck from './check';
+import PostPanelRow from '../post-panel-row';
 import { store as editorStore } from '../../store';
 
-/**
- * Renders the component for displaying the last revision of a post.
- *
- * @return {Component} The component to be rendered.
- */
-function PostLastRevision() {
-	return <PrivatePostLastRevision isLink={ false } />;
-}
-
-export function PrivatePostLastRevision( { isLink } ) {
-	const { lastRevisionId, revisionsCount } = useSelect( ( select ) => {
+function usePostLastRevisionInfo() {
+	return useSelect( ( select ) => {
 		const { getCurrentPostLastRevisionId, getCurrentPostRevisionsCount } =
 			select( editorStore );
 		return {
@@ -31,6 +23,15 @@ export function PrivatePostLastRevision( { isLink } ) {
 			revisionsCount: getCurrentPostRevisionsCount(),
 		};
 	}, [] );
+}
+
+/**
+ * Renders the component for displaying the last revision of a post.
+ *
+ * @return {Component} The component to be rendered.
+ */
+function PostLastRevision() {
+	const { lastRevisionId, revisionsCount } = usePostLastRevisionInfo();
 
 	return (
 		<PostLastRevisionCheck>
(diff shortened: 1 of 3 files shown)


jasmussen:
Jay is out on a break, but we can ask @richtabor for a quick take too.

I'm wondering, could it just be the "Last edited 6 days ago" that shows the revisions? Alternately, I'd lean this way:

Screenshot 2024-06-05 at 11 30 55

I.e. another row.

annezazu:
Agreed on showing it as an additional row. That feels like the proper weighting and like a better design choice.

aaronjorbin:
I like @jasmussen's design a lot more than my quick idea (not surprising to me).  I did some quick testing in the playground using chrome and this is looking good.  I see that the revisions update upon a save and that they appear after there is a second revision. Clicking them takes them to the revisions screen. 

I think that overall this reduces the prominence of revisions vs wp6.5 but not at the expense of making them hard to find or making the number of revisions only available on the revisions screen. Thank you!

creativecoder (commented):
This looks good and is also working well for me.

Should the dropdown menu link be removed, since revisions are now displayed in the post summary list?

image

Mamaduka:
> Should the dropdown menu link be removed, since revisions are now displayed in the post summary list?

I've a same question.

ntsekouras:
>Should the dropdown menu link be removed, since revisions are now displayed in the post summary list?

Personally I think it's fine to have it in both places, but that's something designers could know better. From the technical aspect, this action is also used in Data Views, so we would need to check the APIs for something like context, we've been discussing.

packages/editor/src/components/post-last-revision/index.js
```diff
@@ -11,22 +11,27 @@ import { addQueryArgs } from '@wordpress/url';
...
 		};
 	}, [] );
+}
+
+/**
+ * Renders the component for displaying the last revision of a post.
+ *
+ * @return {Component} The component to be rendered.
+ */
+function PostLastRevision() {
```
Mamaduka, line 33: I'm looking at some other panel refactorings; they didn't keep the original versions. Example: #61357.

Should we do the same here?

Mamaduka (approved):
Codewise changes look good. The action can be removed separately based on design feedback.

jasmussen:
I'd remove the link from the menu.

packages/editor/src/components/post-last-revision/index.js
ntsekouras: I don't think it's the same case because here we're not changing the entire panel (`PostLastRevisionPanel`) that is exported. So consumers of the above panel, should have the 'old' `PostLastRevision` component.

Mamaduka: If the policy is to keep the old components intact, then we might need to revisit some of the refactoring, like #61357. Because `PostDiscussionPanel` has a new structure.

ntsekouras: I don't think there is a clear policy about that, but I think the changes should not be disruptive in the sense that they would make sense in other designs too. An example would be if we change a control and the exporting component is just that. In this case IMO it's fine to not keep the old version. In this PR for me it made more sense to keep the old one as it matches better when used inside `PostLastRevisionPanel` with the panel body and we actually don't use `PostLastRevisionPanel` anymore in our codebase.

Can you share the components you're concerned with the discussions panel? Also cc @jorgefilipecosta for visibility.

Mamaduka: My concern is that the discussion panel consumers expect it to render `PanelBody > PanelRow`. Now, they'll get a different design.

I personally don't mind either path for changes as long as they're consistent.

ntsekouras: I know they get a different design and might be a bit different, but in my mind they still get the whole functionality (the whole panel). If I had to update the panel here for revisions personally I'd follow the same approach with the discussions panel. This is subjective though..

hrkhal:
This is much better, cheers for actioning.

ntsekouras:
Let's land this and explore how to remove the action in the editor context separately in actions API.

ellatrix:

PR #62324, opened 2024-06-05
Interactivity API template create block: Removed warning for generated README from template

juanmaguitar:
## What?
Removes the following **Warning** from the generated README when using `create-block-interactive-template` template for `create-block` tool

> This block requires Gutenberg 16.2 or superior to work. The Interactivity API is, at the moment, not part of WordPress Core as it is still very experimental, and very likely to change.

## Why?
Because the API has evolved since that message was set and that warning is not needed anymore


cbravobernal (approved):
LGTM

PR #56364, opened 2023-11-21
Show visual label for Categories block in dropdown mode

carolinan:
## What?

Updated January 3 2024.

Right now the Categories block in dropdown mode does not show a visual label.
This pull request makes the label text customizable and visible by default if the block's dropdown option is enabled.

Closes https://github.com/WordPress/gutenberg/issues/11003
## Why?

A label is required because the select list is a form element.
The label is always present in the markup, but before the PR, it was not editable.

So even if the user kept the label visually hidden, they were unable to edit the label based on context.
There was no way to differentiate the blocks if multiple of them were placed together.

There was no visible label in the editor, and no option to show the label on the front.

## How?

Adds two new block attributes to the categories block: label and showLabel.
Replaces the hard coded label with richText so that users can change the text.
Adds a toggle button in the block sidebar that toggles the visibility of the label. This option is only available when  dropdown option is enabled.

## Testing Instructions

1. Create a new post or page
2. Insert a categories list block.
3. In the block settings sidebar, enable the option "Display as dropdown".
4. Confirm that there is a visible text above the dropdown. The text should say "Categories" and should have a lighter color, to indicate that it is a placeholder text.
5. Place the cursor anywhere in the text, and type tour custom label. 
6. Select part of the text and make it bold via the option in the toolbar.
7. Go back to the block settings sidebar and open the Styles tab. Change the font size. Confirm that the size of the label changes.
8. Duplicate the block.
9. Next, turn off the "Show label" toggle option in the sidebar. You should now have one block with a visible label, and one with a hidden label.
10. Save and view the front of the website. Confirm that both copies of the block are correct. -View the source of the page to confirm that the visually hidden label is present on the page, with the correct text.

### Testing Instructions for Keyboard
I tested this with VoiceOver on Chrome, mac.

1. Create a new post or page
2. Insert a categories list block.
3. Navigate to the block settings sidebar. Navigate past the tab selection until you reach the Settings tab-panel.
The first option in the settings tab-panel is the "Display as dropdown". This is a tickbox. Pressing space or enter should change the state from unticked to ticked.

4. Navigate back to the block in the editor canvas. 
When the block is selected, the block name is announced.
Press the down arrow key. The screen-reader should announce that you are on a label element and that the text can be edited. It should read the current label text, and that the inserter is at the end of the text.
Editing the text should read out the individual letters.

5. Save, and continue testing the block on the front.
Navigate to the menu-popup in the content. Confirm that the text you entered in the block's label is announced when the select list is focused.

[...]

PR #55219, opened 2023-10-10
Use all the settings origins for a block that consumes paths with merge.

matiasbenedetto:
## What?
Use all the settings origins for a block that consumes paths with merge.

## Why?
Because we are not displaying all the available settings if there are more than one origin in use.

## How?
By concatenating all the origins

## Testing Instructions
- Add content to a setting using more than one origin. ('theme' and 'custom', for example).
- Check that the component consuming that setting displays the content from all the origins.

Please take a look at the screencast below to see how you can test this using the editor UI.

## Screenshots or screencast 

This change is not exclusive to font families and is not specific about the font library, I'm just showing the usage of the font library because it's easy to test the change using the UI because the installed fonts are in a different key than the theme ones.

https://github.com/WordPress/gutenberg/assets/1310626/ec2ccaa0-ad02-4d0b-ab02-5e7e19674622

Fixes https://github.com/WordPress/gutenberg/issues/55011

Co-authored-by: Jason Crist <146530+pbking@users.noreply.github.com>

Issue #55185, opened 2023-10-10
Custom colors in color palette don't increment on non-English sites


properlypurple:
### Description

When using the color palette under global styles to add custom colors, every new color should show up like `Color 1`, `Color 2`, and so on. When on a site with a different language, the numbers in custom colors don't increment.

### Step-by-step reproduction instructions

1. On a fresh site, set the site language to French or German.
2. If using the Gutenberg plugin, install the language pack from Dashboard > Updates
3. Start with a block theme(2023, Spectra, etc), and go to `Appearance > Editor`
4. Go `Styles > Colors > Palette`. Click the button to add a new custom color. In French, this would be under `Personnalisé`.
5. Add a few more colors, and notice that the names are all `Couleur 1` by default.
6. Close the editor, and switch the site's language to English.
7. Repeat steps 3-5, and notice that the color names auto-increment `Color 1`, `Color 2`, and so on.

### English 

https://github.com/WordPress/gutenberg/assets/40858144/c4f17b2d-ce08-44a0-8a89-7dfbcc5052e7

### French

https://github.com/WordPress/gutenberg/assets/40858144/b82ab7e1-f703-48ca-b7ad-c53135cff4fc

I've also tested this with German, Spanish, and Brazilian Portuguese, with the same results.

### Environment info

- WP 6.3.1, with or without the Gutenberg (16.7.1) plugin

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

properlypurple:
It looks like this is related to how the translation is handled here

https://github.com/WordPress/gutenberg/blob/27964574bbf4828a415555a456baeaf61fbdffb0/packages/components/src/palette-edit/index.tsx#L105

t-hamano:
Thanks for the report.

Some issues have been reported for non-Latin languages with the global style custom color palette.

- #49711
- #39210

It may be difficult to solve all of these problems at once, but I believe that #54332 will be the first step towards solving them.

t-hamano:
This issue has been fixed in #52212. The screencast below shows that palette names are incremented correctly even on a French site.


https://github.com/WordPress/gutenberg/assets/54422211/9268adce-08db-4af9-b590-a11bcab86495

PR #62328, opened 2024-06-05
Font size presets UI

### LINKED ISSUE #61987: Add UI to Global styles to allow modification of font size presets

matiasbenedetto:
## What?
Adds UI to Global styles to allow modification of font size presets.
Fixes: https://github.com/WordPress/gutenberg/issues/61987

## Why?
To allow users to edit the font size presets using the editor.

## How?
By implementing the UI to handling creation, update and removal of font size presets.

## Screencasts
See the comments below to find the screencasts.


jasmussen:
Feel free to let @WordPress/gutenberg-design know when this PR is ready to review! Since the main issue is labelled as needing design, however, I took a quick look. But I'm unsure where to look. Would you be able to add screenshots or testing instructions to the PR or issue?

matiasbenedetto:
@WordPress/gutenberg-design here's a screencast featuring some work about this. Could you please take a look?

Not all of this has already been pushed to this branch.

[Screencast from 10-06-24 13:06:38.webm](https://github.com/WordPress/gutenberg/assets/1310626/232ac015-b3d7-4b85-935c-6eba289ced5d)

matiasbenedetto:
Adding a new font size preset:

[Screencast from 10-06-24 16:43:14.webm](https://github.com/WordPress/gutenberg/assets/1310626/cb806838-1b75-41e8-bcce-a44ce1e6ca8e)

matiasbenedetto:
Removing font size preset:

[Screencast from 10-06-24 16:46:22.webm](https://github.com/WordPress/gutenberg/assets/1310626/ac745ea9-b03f-4ed7-9da9-a560afc24cb2)

matiasbenedetto:
:information_source: I pushed all the changes to this branch so it can be tested.

jasmussen:
Nice work, this is not that far off! I'll give this a more thorough review when I get a moment, for now I've signal boosted this a bit for feedback.

Two things that do stand out to me in this part:

Screenshot 2024-06-11 at 10 12 46

The "Reset" button next to the font size slider, and the Fluid yes/no. For the former, can we just remove the reset button? You can always backspace the value to clear it, or remove the font preset. For the latter, outside of better ideas, perhaps just a toggle that says "Fluid typography" rather than a yes/no control might work.

beafialho:
This is exciting to see! It will greatly impact my work and, I'm sure, of all those creating/playing with themes as well.

It would be neat if the added custom font size presets could be automatically organized by size values from **Smallest > Largest** so that we don't see a `0.8rem` after a `3.9rem`. Would that be possible?

matiasbenedetto:
> It would be neat if the added custom font size presets could be automatically organized by size values from Smallest > Largest so that we don't see a 0.8rem after a 3.9rem. Would that be possible?

That's a nice suggestion, thanks. There's only one technical/design difficulty about that: the mix of different units. Would it be okay to order the sizes primarily by unit and secondarily by quantity? 

If we have a list of multiple font sizes with different units, it could look like this:

- 12px
- 20px
- 1.5 rem
- 3 rem
- 2 vw
- 4 vw

Would this type of order be OK @beafialho @jasmussen?

Another possibility is to calculate the relative units with a coefficient that allows us to make an approximative scale of the real size of that font for a 'typical' desktop screen.


In this example we would order by the 'result' column. The order is slightly different. It tries to replicate what would be the size of the fonts in a 'typical' (1920px wide) screen.


Font size | Unit | Pixel coefficient | Result
-- | -- | -- | --
12 | px | 1 | 12
20 | px | 1 | 20
1.5 | rem | 16 | 24
2 | vw | 19.2 | 38.4
3 | rem | 16 | 48
4 | vw | 19.2 | 76.8

beafialho:
> Would this type of order be OK @beafialho @jasmussen?

That would be ok IMO, as long as the custom ones within the same unit are ordered by sizes 👍

packages/edit-site/src/components/global-styles/font-sizes.js
```diff
@@ -0,0 +1,83 @@
...
+import { NavigationButtonAsItem } from './navigation-button';
+
+function getAvailableIndex( sizes ) {
+	// Filter the slug field that ends with custom-{number}.
+	const indexes = sizes.map( ( size ) => {
+		const match = size.slug.match( /custom-(\d+)$/ );
+		return match ? parseInt( match[ 1 ], 10 ) : 0;
+	} );
+	return Math.max( ...indexes ) + 1;
+}
```
t-hamano: We should be able to use the existing [`getNewIndexFromPresets()` function](https://github.com/WordPress/gutenberg/blob/19307c0b0b93b04a58fc25b7d22d3bcf40fd1a94/packages/edit-site/src/components/global-styles/utils.js#L24).

packages/edit-site/src/components/global-styles/font-sizes.js
```diff
@@ -0,0 +1,83 @@
...
+			<HStack justify="space-between">
+				<Subtitle level={ 3 }>{ __( 'Font Sizes' ) }</Subtitle>
+				<Tooltip text={ __( 'Manage font sizes' ) }>
+					<Button
+						aria-label={ __( 'Add font size ' ) }
+						icon={ plus }
+						size="small"
+						onClick={ handleAddFontSize }
+					/>
+				</Tooltip>
```
t-hamano: Why is the tooltip text and `aria-label` different?

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,257 @@
...
+									size="small"
+									icon={ moreVertical }
+									label={ __( 'Menu' ) }
+								/>
+							}
+						>
+							<DropdownMenuItem onClick={ handleRemoveFontSize }>
+								<DropdownMenuItemLabel>
+									{ __( 'Remove font size' ) }
+								</DropdownMenuItemLabel>
```
t-hamano: image

Like the shadow presets, "Delete" might be good. Also, how about moving the rename function here.

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,257 @@
...
+							</DropdownMenuItem>
+						</DropdownMenu>
+					</Spacer>
+				</FlexItem>
+			</HStack>
+
+			<Spacer marginX={ 4 }>
+				<FontSizePreview fontSize={ fontSize } />
+			</Spacer>
```
t-hamano: Is it possible to layout the components below with `VStack` without using a `Spacer` component? The margins of adjacent `Spacer` components will interfere with each other, making it difficult to see what margins are actually being applied.

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,257 @@
...
+function FontSize() {
+	const {
+		params: { slug },
+		goBack,
+	} = useNavigator();
+	const [ fontSizes, setFontSizes ] = useGlobalSetting(
+		'typography.fontSizes'
+	);
+	// Get the font sizes from the theme or use the default ones.
+	const sizes = fontSizes.theme ?? fontSizes.default;
```
t-hamano: Do we need to consider the case when `defaultFontSizes` is `false`?

t-hamano (commented):
Thanks for the great work!

My concern is that the `FontSizePicker` component has been significantly updated and a new component called `SizeControl` has also been added.

The `SizeControl` component is marked as private, but `FontSizePicker` uses it internally. Since the `FontSizePicker` component is public, it may indirectly become public even if the `SizeControl` component is private.

Do you intend for the `SizeControl` component to be used outside of the Global Styles?

If not, it might be better to extend [the `FontSizePicker` component in the `block-editor` package](https://github.com/WordPress/gutenberg/blob/add/font-size-presets-ui/packages/block-editor/src/components/font-sizes/font-size-picker.js) rather than making changes to the `components` package.

I think we will need feedback from the @WordPress/gutenberg-components team if we move forward with the current approach.

packages/edit-site/src/components/global-styles/font-sizes.js
matiasbenedetto: done here: [`8465ece` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/8465ece2b6a0d330737821f189e41e31ddd86715)

packages/edit-site/src/components/global-styles/font-sizes.js
matiasbenedetto: updated here: [`a3a97d9` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/a3a97d9c420cbf27779915d22e36fa8eb0b8f7ae)

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: updated here: [`0b6c490` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/0b6c490712fd1ded4e9a641a869051b5e2c1e42d)

matiasbenedetto:
> Do you intend for the SizeControl component to be used outside of the Global Styles?

> If not, it might be better to extend [the FontSizePicker component in the block-editor package](https://github.com/WordPress/gutenberg/blob/add/font-size-presets-ui/packages/block-editor/src/components/font-sizes/font-size-picker.js?rgh-link-date=2024-06-11T15%3A22%3A57Z) rather than making changes to the components package.

As in this PR, I think the current component couldn't be achieved by extending the FontSizeControl component.  The new SizeControl component is only a subset of the FontSizePicker extracted to be re-used independently.

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: done here: [`239b781` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/239b781256ffe4600b0a799e57cf209ff8a2cc2f)

packages/components/src/size-control/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+	default as UnitControl,
+	parseQuantityAndUnitFromRawValue,
+	useCustomUnits,
+} from '../unit-control';
+
+import { Spacer } from '../spacer';
+
+export const DEFAULT_UNITS = [ 'px', 'em', 'rem', 'vw', 'vh' ];
+
+function SizeControl( props: SizeControlProps ) {
```
tyxla: @WordPress/gutenberg-components how do we feel about the new `SizeControl` component being introduced as a separate component (a private API to start with)?

markhowellsmead:
- A visible reset button is hugely advantageous for most users. Simply entering the field and manually removing the value requires more interaction, and is less intuitive than a common button pattern.
- As in many places in the editor, the addition of more extensive help text (or links to documentation) would be hugely useful for many users. For example, an explanation of what fluid sizing is, in layperson terms.

jasmussen:
Agreed on the advantage, but I'd rather improve the existing UI for resetting values across, than make exceptions to the core componentry in a local PR like this. Changes to the core componentry should be separate PRs.

matiasbenedetto:
Based on @jasmussen 's feedback, I changed the toggle component and removed the reset buttons:

[Screencast from 19-06-24 14:33:14.webm](https://github.com/WordPress/gutenberg/assets/1310626/66f55d98-3e4a-4943-8938-4ad5de439f78)

matiasbenedetto:
Based on @beafialho 's feedback I implemented the coefficient ordering for font sizes described in this comment: https://github.com/WordPress/gutenberg/pull/62328#issuecomment-2160459582

image

matiasbenedetto:
> As in many places in the editor, the addition of more extensive help text (or links to documentation) would be hugely useful for many users. For example, an explanation of what fluid sizing is, in layperson terms.

@markhowellsmead, you are welcome to suggest texts/hints to improve the current UI.

markhowellsmead:
> @markhowellsmead, you are welcome to suggest texts/hints to improve the current UI.

Where would I do that? Here?

packages/block-editor/README.md
```diff
@@ -388,7 +388,7 @@ Undocumented declaration.
 
 _Related_
 
--   <https://github.com/WordPress/gutenberg/blob/HEAD/packages/block-editor/src/components/font-sizes/README.md>
+-   <https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/font-sizes#readme>
```
t-hamano: What is the intent behind this change? Maybe the original link doesn't work because the correct link extension is in uppercase.

https://github.com/WordPress/gutenberg/blob/HEAD/packages/block-editor/src/components/font-sizes/README.MD

packages/edit-site/src/components/global-styles/font-sizes.js
```diff
@@ -0,0 +1,116 @@
...
+			<HStack justify="space-between">
+				<Subtitle level={ 3 }>{ __( 'Font Sizes' ) }</Subtitle>
+				<Tooltip text={ __( 'Add font size' ) }>
+					<Button
+						aria-label={ __( 'Add font size' ) }
+						icon={ plus }
+						size="small"
+						onClick={ handleAddFontSize }
+					/>
+				</Tooltip>
```
t-hamano:
```suggestion
				<Button
					label={ __( 'Add font size' ) }
					icon={ plus }
					size="small"
					onClick={ handleAddFontSize }
				/>
```

I think this is simpler.

packages/components/src/size-control/index.tsx
```diff
@@ -0,0 +1,158 @@
+/**
```
t-hamano, line 1: If we were to add a new `SizeControl`, I think we also need to add Storybook, unit tests, and readme.

However, as [this comment](https://github.com/WordPress/gutenberg/pull/62328#discussion_r1642884569) suggests, we may need to have a bit more discussion beforehand about whether we should add this component.

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+								) }
+								checked={ isCustomFluid }
+								onChange={ handleCustomFluidValues }
+							/>
+						) }
+
+						{ isCustomFluid && (
+							<>
+								<SizeControl
+									label={ __( 'Minimun' ) }
```
t-hamano: `Minimum` or `Min` ?

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+									value={ fontSize.fluid?.min }
+									withSlider
+									onChange={ handleMinChange }
+									hasUnits={ hasUnits }
+									withReset={ false }
+								/>
+								<SizeControl
+									label={ __( 'Maximun' ) }
+									value={ fontSize.fluid?.max }
```
t-hamano: `Maximum` or `Max` ?

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+		const newFontSizes = sizes.map( ( size ) => {
+			if ( size.slug === slug ) {
+				return { ...size, [ key ]: value }; // Create a new object with updated key
+			}
+			return size;
+		} );
+
+		setFontSizes( {
+			...fontSizes,
+			theme: newFontSizes,
```
t-hamano: Is the origin a theme? I think we need to change the origin where the settings are saved to match the origin the user edited.

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+		updateFontSize( 'size', newValue );
+	};
+
+	const handleFluidChange = ( value ) => {
+		updateFontSize( 'fluid', value );
+	};
+
+	const handleCustomFluidValues = ( value ) => {
+		if ( value ) {
+			// If custom values are use init the values with the current ones.
```
t-hamano:
```suggestion
			// If custom values are used, init the values with the current ones.
```

Is this grammatically correct?

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+	// Memoized initial value of fontSize
+	const initialFontSize = initialFontSizeRef.current;
+
+	const handleNameChange = ( value ) => {
+		updateFontSize( 'name', value );
+	};
+
+	const handleFontSizeChange = ( value ) => {
+		// If the user is resetting the value, use the initial value.
+		const newValue = value ?? initialFontSize;
```
t-hamano: Since the reset button has been removed, is this condition no longer necessary?

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+
+	const handleMinChange = ( value ) => {
+		// If the user is resetting the value, use the initial value.
+		const newValue = value ?? fontSize.size;
+		updateFontSize( 'fluid', { ...fontSize.fluid, min: newValue } );
+	};
+
+	const handleMaxChange = ( value ) => {
+		// If the user is resetting the value, use the initial value.
+		const newValue = value ?? fontSize.size;
```
t-hamano: I don't think `fontSize.size` is the initial value for `fluid`. As with [this comment](https://github.com/WordPress/gutenberg/pull/62328#discussion_r1646892497), is this process unnecessary?

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+	};
+
+	const handleRemoveFontSize = () => {
+		// Navigate to the font sizes list.
+		goBack();
+
+		const newFontSizes = sizes.filter( ( size ) => size.slug !== slug );
+		setFontSizes( {
+			...fontSizes,
+			theme: newFontSizes,
```
t-hamano: Similar to [this comment](https://github.com/WordPress/gutenberg/pull/62328#discussion_r1646887376).

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+						/>
+
+						<SizeControl
+							label={ __( 'Size' ) }
+							value={ fontSize.size }
+							withSlider
+							onChange={ handleFontSizeChange }
+							hasUnits={ hasUnits }
+							withReset={ false }
+						/>
```
t-hamano: How about adding `__nextHasNoMarginBottom` and `__next40pxDefaultSize`? The same goes for min/max fields.

packages/edit-site/src/components/global-styles/style.scss
```diff
@@ -23,6 +23,22 @@
 	overflow: hidden;
 }
 
+.edit-site-typography-panel__full-width-control {
+	grid-column: 1 / -1;
+	max-width: 100%;
+}
```
t-hamano: Perhaps this style is not being applied anywhere.

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+
+						<ToggleControl
+							label={ __( 'Fluid typography' ) }
+							checked={ isFluid }
+							onChange={ handleFluidChange }
```
t-hamano:
```suggestion
							onChange={ handleFluidChange }
							__nextHasNoMarginBottom
```

packages/edit-site/src/components/global-styles/font-size.js
```diff
@@ -0,0 +1,239 @@
...
+						/>
+
+						{ isFluid && (
+							<ToggleControl
+								label={ __( 'Custom fluid values' ) }
+								help={ __(
+									'Set custom min and max values for the fluid font size.'
+								) }
+								checked={ isCustomFluid }
+								onChange={ handleCustomFluidValues }
```
t-hamano:
```suggestion
								onChange={ handleCustomFluidValues }
								__nextHasNoMarginBottom
```

t-hamano (commented):
Thanks for the update. I would like to make a few suggestions and would be grateful for your feedback.

---

It might be nice to have a confirmation dialog when removing a font, similar to the custom shadow palette.

shadow-delete-confirm

---

> Based on @beafialho 's feedback I implemented the coefficient ordering for font sizes described in this comment: #62328 (comment)

Personally, I find this automatic sorting a bit unnatural - there is no automatic sorting for custom colors or shadows, and when you add a new font size, it is not added to the end of the presets.

new-font-size

---

Regarding renaming, how about moving it to the ellipsis menu like we did with custom shadows?

rename-shadow

Otherwise, changing the preset name would be immediately reflected in the title and description, which would feel unnatural. Additionally, it would be nice to be able to disallow empty values ​​for preset names.

https://github.com/WordPress/gutenberg/assets/54422211/42b17918-b44e-4fee-8ac4-cdc16d5da509

---

We can rename or delete the default/theme presets, which is unexpected for me.

With color and shadow presets, we can't rename or delete the default presets.

packages/block-editor/README.md
matiasbenedetto: the intention is to fix the broken link. The new link goes to the same file.

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: @jasmussen I'd appreciate your input about the wording of this label.

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: Let's continue discussing this in the other thread https://github.com/WordPress/gutenberg/pull/62328#discussion_r1646882626. I'm marking this as resolved.

packages/block-editor/README.md
t-hamano: If so, how about changing the extension of the readme file to lower case?

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: Let's continue discussing this in that thread https://github.com/WordPress/gutenberg/pull/62328#discussion_r1646887376. Having multiple threads for the same thing make things harder to read. Marking this as resolved.

packages/edit-site/src/components/global-styles/style.scss
matiasbenedetto: Thanks for pointing that out. This was added accidentally when I merged the latest trunk branch changes. Removed here: [`73617a8` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/73617a8c5ae2822195628f09794ff43e4b7564c7)

packages/edit-site/src/components/global-styles/font-size.js
jasmussen: Assuming you're referring to this part:

Screenshot 2024-06-20 at 12 32 49

Let's go with the full labels, "Minimum" and "Maximum", mostly because in other places of the UI, like "Minimum height" for the Cover block, we also use the full word. The help text reads okay as is, to me, but if need be that can be more verbose too.

Do note, though, that there's a typo at present. They say "Minimu**n**", and "Maximu**n**", replace the n with m.

Separate to this PR, we should look at updating the default font size control to be the correct 40px hight and use the new unit style. The Image block looks correct, note the unit selector and input height, as well as inner padding:

Screenshot 2024-06-20 at 12 35 10

Then compare to the controls here:

Screenshot 2024-06-20 at 12 32 29

Not specifically your task, but we want to update all of these instances so they have the proper 40px defaults.

matiasbenedetto: Yes, we were referring to that part. Ok, let's stick to 'Minimum' and 'Maximum'. Thanks for noticing the misspelled words. Fixed them here: [`0c8ecde` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/0c8ecdeb071f2167e1f8e73d1800205b1157a8ff)

matiasbenedetto:
> We can rename or delete the default/theme presets, which is unexpected for me.
> With color and shadow presets, we can't rename or delete the default presets.

These comments seem to suggest that the font size presets should work as the color palettes or shadow presets. I have 2 thoughts about this:
1. This is not how font size presets work in the editor right now. If you have font sizes defined in more than one 'key' (example: `default` and `theme`), only the theme ones are listed on the UI for the user to select. The default and the theme sizes are not combined in one list. Only the theme ones are listed. This is opinionated, but each type of setting has its own logic regarding this.
@jasmussen @youknowriad I'd appreciate your input about this.

 2. I don't think that having many 'groups' of font size presets, like in color palettes, adds much value for users. @WordPress/block-themers input welcome.
 
 > Is the origin a theme? I think we need to change the origin where the settings are saved to match the origin the user edited.
 
 If we keep the current behavior, saving all the sizes in 'theme' or 'custom' should be ok, as the other alternative is unnecessary and more complex.

packages/block-editor/README.md
matiasbenedetto: Oh, I see; yes, renaming the file is the best option. Done that on this PR: https://github.com/WordPress/gutenberg/pull/62697, as it is not strictly related to the changes introduced in this PR. Thanks for pointing that out.

youknowriad:
@t-hamano @matiasbenedetto it seems to me that whatever we do for colors, we should follow for fonts presents. 

If I'm not wrong for colors, we allow editing theme and custom palettes (but not default I think) and we show all of them when picking a color for a given style.

matiasbenedetto:
> If I'm not wrong for colors, we allow editing theme and custom palettes (but not default I think) and we show all of them when picking a color for a given style.

Yes, that's true, but it is not how the UI currently works for font sizes. I remember that there was at least one attempt to make all the presets available to the user (default, theme, custom), but that was considered a bug. I remember discussing this in this PR (https://github.com/WordPress/gutenberg/pull/55219), but there could be other threads, too. 

@youknowriad @jasmussen @WordPress/block-themers , do you think replicating the color palette model adds value for users? What would the font picker look like with the size presets coming from all the origins?

youknowriad:
I'm not strongly opinionated on the UI to be honest, so I'll let the other chime-in.

jasmussen:
> it seems to me that whatever we do for colors, we should follow for fonts presents.

I believe @richtabor and @scruffian are working on exactly this. 

Sidenote, @WordPress/gutenberg-design is a great ping for feedback.

matiasbenedetto:
> Where would I do that? Here?

@markhowellsmead yep, here is a good place. You can suggest improvements in this pull request about this particular addition to the editor

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: I removed all the reset related logic here: [`cc582db` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/cc582dbed6b82f512c542ec391fa1ae22de337b3)

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: I removed all the reset related logic here: [`cc582db` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/cc582dbed6b82f512c542ec391fa1ae22de337b3)

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: added here: [`55d4b40` (#62328)](https://github.com/WordPress/gutenberg/pull/62328/commits/55d4b40942c2d23772b71631967d797a1ff340f1)

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: added

packages/edit-site/src/components/global-styles/font-size.js
matiasbenedetto: added

matiasbenedetto:
> It might be nice to have a confirmation dialog when removing a font, similar to the custom shadow palette.

> Regarding renaming, how about moving it to the ellipsis menu like we did with custom shadows?

Based on @t-hamano 's feedback I added the rename and delete dialogs:

[Screencast from 21-06-24 16:32:47.webm](https://github.com/WordPress/gutenberg/assets/1310626/7ddda676-65e8-415e-bd38-760e7eb79551)

annezazu:
@WordPress/block-themers feedback is needed here from a block theme perspective 👀

packages/components/src/size-control/index.tsx
mirka: As @jasmussen [said](https://github.com/WordPress/gutenberg/pull/62328#issuecomment-2178531035), a big change in core components like this should be discussed in a separate issue to figure out the best approach. And the "why" of the change is not immediately clear to me from this PR — we may be able to find other approaches once the "why" is more clear.

ciampo: IMO, `SizeControl` shouldn't live in the `@wordpress/components` package, since it feels too tailored to a specific need of the editor.

We could instead discuss the need for a more general purpose `UnitControl` + `RangeControl` component, which was already initially discussed in https://github.com/WordPress/gutenberg/issues/40507 and then abandoned for lack of priority.

matiasbenedetto: Thanks everyone for your input.

> And the "why" of the change is not immediately clear to me from this PR — we may be able to find other approaches once the "why" is more clear.

@mirka I submitted the component in a new PR: https://github.com/WordPress/gutenberg/pull/62866

> IMO, SizeControl shouldn't live in the @wordpress/components package, since it feels too tailored to a specific need of the editor.

@ciampo What does it mean 'tailored for a specific need of the editor' in this context?
I don't see how SizeControl is more tailored 'for a specific need of the editor' than the FontSizePicker. Could you explain, please?

ciampo: > I don't see how SizeControl is more tailored 'for a specific need of the editor' than the FontSizePicker. Could you explain, please?

That is true, and in fact I also think that `FontSizePicker` shouldn't be in the `@wordpress/components` package either for the same reason — but it is there and we can't easily remove it.

t-hamano: Looking at the code, it seems to me that the reason the SizeControl component was added is to make several sets of UI common.

If that is the only purpose, couldn't the SizeControl component be placed as an internal component in the global styles directory rather than in the component package? In other words, `packages/edit-site/src/components/global-styles/font-sizes/size-control.js`.

In the future, if something like the `SliderControl` described in #40507 is added to the components package, we can easily replace the internal component with the official one.

ciampo: @t-hamano that would be my suggestion too — @mirka , what you think?

matiasbenedetto: > the reason the SizeControl component was added is to make several sets of UI common. If that is the only purpose, couldn't the SizeControl component be placed as an internal component in the global styles directory rather than in the component package?

Isn't that the reason behind all the components? I don't get why certain things should be components and certain not. The criteria doesn't seem to be clear at least for me.

matiasbenedetto: More context: I was planning to use the proposed component also in the implementation of the spacing presets UI: https://github.com/WordPress/gutenberg/issues/61986

t-hamano: I think the criteria for adding new components to the components package are defined [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/CONTRIBUTING.md).

The components package is used by many developers. The reason I hesitate to add new components to a component package is that once a new component is exposed, it cannot be easily removed or modified, and requires maintenance work.

I feel that the versatility of the `SizeControl` component has not been fully evaluated so far. I think we can implement it as an internal component and move forward with this PR until the discussion progresses.

mirka: I'm not enthusiastic about duplicating all the SizeControl logic/styles and making them live in wp-edit-site though. I wouldn't say it's a trivial component.

The possibilities I see are:

- Rework the `disableCustomFontSizes` prop in FontSizePicker so that it allows for custom font sizes only (which I believe would be the equivalent of the proposed SizeControl). For example we could deprecate `disableCustomFontSizes` and make a new enum prop with three options (e.g. `"custom" | "predefined" | "both"`).
- Extract to a component as proposed in #62866. For context, we do have some [precedent](https://github.com/WordPress/gutenberg/issues/52734#issuecomment-2038252660) in DateTimePicker, and I have seen a somewhat similar proposal for SizeControl already (#61180). We're definitely seeing devs needing more modularity like this. In this case, I think I would prefer exporting it as a subcomponent of FontSizePicker (e.g. `FontSizePicker.CustomInput`) rather than a completely separate component.

matiasbenedetto: > I think the criteria for adding new components to the components package are defined [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/CONTRIBUTING.md?rgh-link-date=2024-06-26T14%3A38%3A55Z).

Yes, thanks, but I had already read that before I commented. Still, I don't see how SizeControl differs from other components available in the package and why it should not exist in the components package.

----

Anyway, I am not available to continue pursuing the component path anymore. It isn't the main objective of this PR to add a new component but to provide a solution for users wanting to manage [font size](https://github.com/WordPress/gutenberg/issues/61987) and [spacing](https://github.com/WordPress/gutenberg/issues/61986) presets. Creating the component was a side task motivated by the desire to avoid repeating code from the font size picker, but that's taking much more time than I anticipated and blocking the original objective. I'll add the UI for this inside `packages/edit-site/src/components/global-styles/size-control/` and maintain the code as short as possible to make the preset management work. If there's a need for a new component, it can be added later as a follow-up. I think that will help this move forward.

carolinan:
I did find it confusing that when I click on the plus icon to add a size, a size is added automatically but unlike the colors I can't edit them directly.

As a developer, I would like to know if I can customize the slug when I use the UI to add sizes and then export the theme,
because custom-X is not so descriptive.

In a separate issue @jasmussen suggested that when a control is conditional, like the settings that are only available when "fluid typography" is toggled on, the controls should always be visible but disabled by default, and enabled when the condition applies.
The control should also be intended, to indicate that it is dependent on the parent.
You can read more about the reasoning here: https://github.com/WordPress/gutenberg/pull/56364#issuecomment-1985286961

jasmussen:
> like the settings that are only available when "fluid typography" is toggled on, the controls should always be visible but disabled by default, and enabled when the condition applies.

There can be nuance here. 

* In the case of the dropdown label, the label being always visible but grayed, can implicitly explain why there's no "show label" toggle unless the "display as dropdown" label is toggled first: they are connected.
* In the case of fluid typography, those min/max values are entirely sub-sets of the main fluid typograhy toggle, you wouldn't look for them otherwise.

So there's an argument that both patterns can be judiciuosly applied based on works best in the given context. If we _have_ to extract a basic rule I'd actually lean on _hiding_ child options, including reverting my own statement from your PR, but then be sure to always indent child-options from the left, so there's a visible indication of hierarchy.

Any preference?

carolinan:
My preference would be consistency and to not have to go back and forward because it creates extra unnecessary work.

matiasbenedetto:
I would like to receive your input about the expected behavior of the font size presets:

Font size presets in the font picker and colors palette are working differently. The colors admit several 'origins' (default, theme, custom) and all of them are available to be selected by the user. Meanwhile, fonts sizes are working differently, 'theme' presets (if they are defined) overrides 'default' font sizes.  'default' presets can not be selected or seen by the user if 'theme' size presets are available.

What path should we follow here?

carolinan:
I did find it confusing, but the consistency remark was about the visibility and positioning of the conditional controls.
I know we can not always get everything right the first time, or second, or third, but it creates so much extra work when a decision is made and then reverted.

So about the conditional controls, that are only visible depending on the parent option, it seems the common ground is to indent them. In this case, there is more than one condition, and I am not sure if it would "look good" or if translations would fit in this space if the last nested control was indented twice.

carolinan:
I don't think there is precedence for this, but perhaps when the new size is added automatically, it could have a bit of a background color fade or similar to indicate that it is the newest item? I did not find it so easy to see which size was the one I had just added.

matiasbenedetto:
> As a developer, I would like to know if I can customize the slug when I use the UI to add sizes and then export the theme,
because custom-X is not so descriptive.

@carolinan is this done for any other setting/preset? Updating the slug seems risky because all the elements where the preset was applied will point to a preset that no longer exist.

carolinan:
As a developer I would like WP to update the slug everywhere ;)
As a developer I also would not like to start working on that because of the performance issues....

I don't want to put this on you, because you have already made progress here and probably have thought about this in detail,
but if the the size was not actually added until I choose a name and save, then the slug could already match the chosen name.

t-hamano:
> As a developer, I would like to know if I can customize the slug when I use the UI to add sizes and then export the theme,
because custom-X is not so descriptive.

Previously, color palettes had slugs automatically generated based on the label you entered. However, if non-alphabetical characters were used in the label, the slug would not be generated correctly or would be duplicated.

To solve this, a sequential number was adopted, independent of the label. See the discussion in issues below.

- https://github.com/WordPress/gutenberg/issues/55185
- https://github.com/WordPress/gutenberg/issues/51230

This rule is consistent across all of the custom color palettes, custom shadow presets, and custom font size presets in this PR.

The question of whether to allow editing of the slug, or generate it by some rule based on the label, might be better discussed in a new issue covering colors, shadows, and font sizes as a whole, rather than in this PR.

matiasbenedetto:
> In the case of the dropdown label, the label being always visible but grayed, can implicitly explain why there's no "show label" toggle unless the "display as dropdown" label is toggled first: they are connected.
> In the case of fluid typography, those min/max values are entirely sub-sets of the main fluid typograhy toggle, you wouldn't look for them otherwise.
> So there's an argument that both patterns can be judiciuosly applied based on works best in the given context. If we have to extract a basic rule I'd actually lean on hiding child options, including reverting my own statement from your PR, but then be sure to always indent child-options from the left, so there's a visible indication of hierarchy.
> Any preference?

I tried both with code. They look like this.
The one with the left padding looks untidy for me.
The one with disabled component looks better but it doesn't hide the unnecessary complexity when the options are not available.
 (side note: the button of the unit controller should not be blue when it is disabled it seems[ like a bug of that component](https://github.com/WordPress/gutenberg/issues/62969)).

| With disabled component | With left indent | 
| --- | --- |
| Screenshot from 2024-06-28 13-49-48 | Screenshot from 2024-06-28 13-40-17 |

packages/components/src/size-control/index.tsx
ciampo: I think that this suggestion from @mirka hits the sweet spot between reducing the amount of new code introduced, future maintenance need and speed of iteration:

> Rework the `disableCustomFontSizes` prop in `FontSizePicker` so that it allows for custom font sizes only (which I believe would be the equivalent of the proposed SizeControl). For example we could deprecate disableCustomFontSizes and make a new enum prop with three options (e.g. "custom" | "predefined" | "both").

I'll see if I can spin up a PR with the necessary changes, so that this PR can use directly `FontSizePicker` — would that be ok @matiasbenedetto @mirka @t-hamano ?

matiasbenedetto: I think that should be a follow-up independent of this task. The component has already consumed too much time for this implementation and it should not be a blocker. Apart from that, I'm not sure modifying the FontSizePicker component is the right thing to do and it certainly doesn't work for the other similar implementation: https://github.com/WordPress/gutenberg/issues/61986

carolinan:
> I tried both with code. They look like this.

Consistency is important for me because it helps me work more efficiently, both when coding the controls and using them, but yeah, I want to say that I don't _like_ this result either.
I know "like" and "feel" are very subjective but this indentation probably works best when there are multiple options that are below the controls.

carolinan:
> The question of whether to allow editing of the slug, or generate it by some rule based on the label, might be better discussed in a new issue covering colors, shadows, and font sizes as a whole, rather than in this PR.

Yes, thank you for the background info!

luminuu:
Seeing this here, what is the difference between "Fluid typography" and "Custom fluid values"?

matiasbenedetto:
> what is the difference between "Fluid typography" and "Custom fluid values"?

If fluid typography is enabled the default values are used for min and max. If the user want's to set manually the min and max, it should enable 'Custom fluid values'.

Reference from this make post: https://make.wordpress.org/core/2022/10/03/fluid-font-sizes-in-wordpress-6-1/

> For presets, individual font sizes can be defined by their minimum and maximum values, which determine the smallest and largest values of a font size. Moreover, individual preset font sizes can opt out by setting settings.typography.fontSizes[].fluid to false.

> In the absence of explicit minimum and maximum values, the editor will scale the existing size value up and down to calculate smallest and largest values automatically using computed lower and upper font size values based on the single value.

jasmussen:
Thanks for trying this. It really seems like the right solution is what you had, to hide the added controls until the toggle is flipped. There might not be a perfect solution that applies consistently for each of these cases.

luminuu:
> If fluid typography is enabled the default values are used for min and max. If the user want's to set manually the min and max, it should enable 'Custom fluid values'.
> 
> Reference from this make post: https://make.wordpress.org/core/2022/10/03/fluid-font-sizes-in-wordpress-6-1/

Thanks! I wonder if it might be helpful to have a description for the `Fluid typography` label too. I'm thinking more of the users who are not aware what fluid typography means.

jasmussen:
> Font size presets in the font picker and colors palette are working differently. The colors admit several 'origins' (default, theme, custom) and all of them are available to be selected by the user. Meanwhile, fonts sizes are working differently, 'theme' presets (if they are defined) overrides 'default' font sizes. 'default' presets can not be selected or seen by the user if 'theme' size presets are available.
> What path should we follow here?

Just making sure I understand this question right: you're asking if it's okay that theme font size presets _overrides and replaces_ those provided by other means (default, theme, custom)?

I feel like I'm still missing a bit of nuance, so I'll cast a bit of a broader net by pinging also @WordPress/gutenberg-design on this question. IMO the two aren't directly comparable, since in the case of colors there need to be fallbacks based on the named slugs. This is less the case with fonts, where they are often applied in a more sweeping style across blocks. 

Not sure if this is helping, but I'd tend to think that whatever you see in that font size presets panel, is going to be the only set of font sizes you can choose from in the typography panels of your site. If they are not customized, that means they come from the theme. If they are customized after the fact, they now come from the user.

packages/components/src/size-control/index.tsx
mirka: @matiasbenedetto What are the requirements for #61986 that cannot be addressed by modifying FontSizePicker? Is there some logic in FontSizePicker that makes it unsuitable as a more generic sizing control, or is it a terminology issue?

matiasbenedetto:
> I wonder if it might be helpful to have a description for the Fluid typography label too. I'm thinking more of the users who are not aware what fluid typography means.

:+1: Added in https://github.com/WordPress/gutenberg/pull/63057

matiasbenedetto:
Thanks everyone for your feedback.

As there was no consensus about adding a new component to the components library I'm closing this PR in favor of https://github.com/WordPress/gutenberg/pull/63057. Please continue providing feedback there. I think I implemented all the suggestion sreceived here, please review and comment in https://github.com/WordPress/gutenberg/pull/63057

packages/components/src/size-control/index.tsx
matiasbenedetto: At a first glance I see 3 issues, that make me think that adding a 'mode' prop to determine the component to determine what it renders isn't the best approach:
- FontSizePicker control only handle a small subset of css units.
- FontSizePicker naming.
- Adding this complexity doesn't seem to add much value to the FontSizePicker component and degrades the developer experience and it's more prone to errors.

Still, I haven't started working on the implementation of https://github.com/WordPress/gutenberg/issues/61986 yet. So I may find other reasons on the way.

mirka: Ok, we'll leave it to the feature folks to figure out the best abstraction. Thanks for elaborating.

I'll point out that we already do have a handful of SizeControl-like instances on the app side ([SpacingInputControl](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/spacing-sizes-control/input-controls/spacing-input-control.js), [BorderRadiusControl](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/border-radius-control/index.js), [GridLayoutMinimumWidthControl](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/layouts/grid.js#L207)) where people built their own UnitControl + slider-only RangeControl combo and did their own orchestration logic. I'm sure they could all be refactored to a certain degree, but I'm not immediately sure if it's worth the consolidation effort.

PR #62270, opened 2024-06-04
Replace exec with spawn in performance runner script

WunderBart:
Related slack convo: [link](https://wordpress.slack.com/archives/C02QB2JS7/p1717483146113199?thread_ts=1717143284.055259&cid=C02QB2JS7)

## What?
Performance tests runner seems to be [swallowing the tests errors](https://github.com/WordPress/gutenberg/actions/runs/9361910664/job/25769690941), printing a cryptic exception:

```sh
Error: Command failed: npm run test:performance -- site-editor

    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at ChildProcess.exithandler (node:child_process:422:12)
    at ChildProcess.emit (node:events:519:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5) {
  code: 1,
  killed: false,
  signal: null,
  cmd: 'npm run test:performance -- site-editor'
}
```

Let's see if passing the outputs the recommended way makes the errors appear as expected.

PR #61450, opened 2024-05-07
Performance Tests: Improve collection and reporting

swissspidy:
## What?

* Collect `Server-Timing` information in performance tests
* Print test results as GitHub Actions summaries for easier overview

See #33645
Fixes #61411

To-do:

* Ideally we'd prepare & reset the environment like core does (disable cron, disable external HTTP requests, flush object cache). But I don't know where to start 🤔  The setup is quite different from core's.

PR #62331, opened 2024-06-05
E2E Utils: Add retry mechanism to the REST API discovery

### LINKED ISSUE #60366: Perf Tests: Update reporter to include test failures
The performance test reporter omits Playwright errors ([example](https://github.com/WordPress/gutenberg/actions/runs/8450329618)), which makes it harder to determine the actual reason for the failure.

Let's update [`PerformanceReporter`](https://github.com/WordPress/gutenberg/blob/trunk/test/performance/config/performance-reporter.ts) to include Playwright-generated errors.

cc @kevin940726, @WunderBart

### LINKED ISSUE #61627: Playwright e2e tests are configured to start wp-env but don't wait for REST API to be advertised
## Description

Playwright is [configured to start wp-env](https://github.com/WordPress/gutenberg/blob/78e53bb99f7e0bf77eb2e0c4576358c83f3b1d72/packages/scripts/config/playwright.config.js#L47-L50) if not already running, but when it does the REST API likely isn't ready in time for the `globalSetup` function to use it. To get around that, the GitHub workflow [starts wp-env ahead of the tests](https://github.com/WordPress/gutenberg/blob/78e53bb99f7e0bf77eb2e0c4576358c83f3b1d72/.github/workflows/end2end-test.yml#L44-L52) in the hope that it's started by the time the `globalSetup` function tries to use it.

In the gutenberg repo, make sure `wp-env` is not already running, then run the e2e tests:

```bash
npm run test:e2e
```

## Expected behaviour

Playwright should wait for all required services to be available before running tests, or should not be configured to start wp-env at all, to clarify that this needs to be done ahead of time.

## Actual behaviour

Playwright starts wp-env, waits for the WordPress port to be open, and then attempts to discover the location of the REST API. The REST API isn't immediately advertised via the `link` header, so discovery fails and the tests are not run.

```
> gutenberg@18.3.0 test:e2e
> wp-scripts test-playwright --config test/e2e/playwright.config.ts

Error: Failed to discover REST API endpoint.
 Link header: undefined

   at ../../../packages/e2e-test-utils-playwright/src/request-utils/rest.ts:32

  30 |
  31 | 	if ( ! restLink ) {
> 32 | 		throw new Error( `Failed to discover REST API endpoint.
     | 		      ^
  33 |  Link header: ${ links }` );
  34 | 	}
  35 |

    at getAPIRootURL (.../packages/e2e-test-utils-playwright/src/request-utils/rest.ts:32:9)
    at RequestUtils.setupRest (.../packages/e2e-test-utils-playwright/src/request-utils/rest.ts:42:29)
    at globalSetup (.../test/e2e/config/global-setup.ts:26:2)
```

WunderBart:
Fixes https://github.com/WordPress/gutenberg/issues/61627
Fixes https://github.com/WordPress/gutenberg/issues/60366

Related thread: https://wordpress.slack.com/archives/C02QB2JS7/p1717143284055259

## What?
Add a retry mechanism to the REST API discovery so that it doesn't fail when the env is not ready yet.

## Testing Instructions
- All tests should pass.
- Perf tests on `trunk` have been failing, so I've triggered a workflow using params from the failing job with this branch as a test runner. [It should pass](https://github.com/WordPress/gutenberg/actions/runs/9383258293/job/25836466597).

### DIFF
--- a/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
+++ b/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
@@ -45,7 +45,7 @@ async function getAPIRootURL( request: APIRequestContext ) {
 				timeout: 60_000, // 1 minute.
 			}
 		)
-		.not.toBeNull();
+		.not.toBeFalsy();
 
 	const [ , rootURL ] = restLink as RegExpMatchArray;
 
--- a/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
+++ b/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
@@ -3,6 +3,7 @@
  */
 import * as fs from 'fs/promises';
 import { dirname } from 'path';
+import { expect } from '@playwright/test';
 import type { APIRequestContext } from '@playwright/test';
 
 /**
@@ -39,10 +40,30 @@ async function getAPIRootURL( request: APIRequestContext ) {
 }
 
 async function setupRest( this: RequestUtils ): Promise< StorageState > {
-	const [ nonce, rootURL ] = await Promise.all( [
-		this.login(),
-		getAPIRootURL( this.request ),
-	] );
+	let nonce = '';
+	let rootURL = '';
+
+	await expect
+		.poll(
+			async () => {
+				try {
+					[ nonce, rootURL ] = await Promise.all( [
+						this.login(),
+						getAPIRootURL( this.request ),
+					] );
+				} catch ( error ) {
+					// Prints the error if the timeout is reached.
+					return error;
+				}
+
+				return nonce && rootURL ? true : false;
+			},
+			{
+				message: 'Failed to setup REST API.',
+				timeout: 60_000, // 1 minute.
+			}
+		)
+		.toBe( true );
 
 	const { cookies } = await this.request.storageState();
 
--- a/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
+++ b/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
@@ -3,7 +3,6 @@
  */
 import * as fs from 'fs/promises';
 import { dirname } from 'path';
-import { expect } from '@playwright/test';
 import type { APIRequestContext } from '@playwright/test';
 
 /**
@@ -23,31 +22,18 @@ function splitRequestsToChunks( requests: BatchRequest[], chunkSize: number ) {
 }
 
 async function getAPIRootURL( request: APIRequestContext ) {
-	let restLink: unknown = null;
+	// Discover the API root url using link header.
+	// See https://developer.wordpress.org/rest-api/using-the-rest-api/discovery/#link-header
+	const response = await request.head( WP_BASE_URL );
+	const links = response.headers().link;
+	const restLink = links?.match( /<([^>]+)>; rel="https:\/\/api\.w\.org\/"/ );
 
-	// Retry until the REST API root URL is discovered.
-	// See https://github.com/WordPress/gutenberg/issues/61627
-	await expect
-		.poll(
-			async () => {
-				// Discover the API root url using link header.
-				// See https://developer.wordpress.org/rest-api/using-the-rest-api/discovery/#link-header
-				const response = await request.head( WP_BASE_URL );
-				const links = response.headers().link;
-				restLink = links?.match(
-					/<([^>]+)>; rel="https:\/\/api\.w\.org\/"/
-				);
+	if ( ! restLink ) {
+		throw new Error( `Failed to discover REST API endpoint.
+ Link header: ${ links }` );
+	}
 
-				return restLink;
-			},
-			{
-				message: 'Failed to discover REST API endpoint.',
-				timeout: 60_000, // 1 minute.
-			}
-		)
-		.not.toBeFalsy();
-
-	const [ , rootURL ] = restLink as RegExpMatchArray;
+	const [ , rootURL ] = restLink;
 
 	return rootURL;
 }
--- a/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
+++ b/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
@@ -42,7 +42,7 @@ async function getAPIRootURL( request: APIRequestContext ) {
 			},
 			{
 				message: 'Failed to discover REST API endpoint.',
-				timeout: 60_000, // 1 minute.
+				timeout: 120_000, // 2 minutes.
 			}
 		)
 		.not.toBeNull();
--- a/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
+++ b/packages/e2e-test-utils-playwright/src/request-utils/rest.ts
@@ -3,6 +3,7 @@
  */
 import * as fs from 'fs/promises';
 import { dirname } from 'path';
+import { expect } from '@playwright/test';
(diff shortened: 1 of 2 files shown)


packages/e2e-test-utils-playwright/src/request-utils/rest.ts
```diff
@@ -22,18 +23,31 @@ function splitRequestsToChunks( requests: BatchRequest[], chunkSize: number ) {
...
-	const links = response.headers().link;
-	const restLink = links?.match( /<([^>]+)>; rel="https:\/\/api\.w\.org\/"/ );
-
-	if ( ! restLink ) {
-		throw new Error( `Failed to discover REST API endpoint.
- Link header: ${ links }` );
-	}
-
-	const [ , rootURL ] = restLink;
+	let restLink: unknown = null;
```
kevin940726: Nit: Let's make it `RegExpMatchArray | null`.
```suggestion
	let restLink: RegExpMatchArray | null = null;
```

packages/e2e-test-utils-playwright/src/request-utils/rest.ts
```diff
@@ -22,18 +23,31 @@ function splitRequestsToChunks( requests: BatchRequest[], chunkSize: number ) {
...
+				);
+
+				return restLink;
+			},
+			{
+				message: 'Failed to discover REST API endpoint.',
+				timeout: 60_000, // 1 minute.
+			}
+		)
+		.not.toBeFalsy();
```
kevin940726: Nit: I think `.not.toBeNull()` will be enough here.

```suggestion
		.not.toBeNull();
```

kevin940726 (approved):
Thank you! A few nitpicks but it's good to go IMO!

packages/e2e-test-utils-playwright/src/request-utils/rest.ts
WunderBart: The TS gymnastics here was that it needs to be casted as `unknown` first so that it can be casted as `RegExpMatchArray` later. TS does not work with expect assertions as it does with `if`s so if we cast it as you suggested, we'll need an additional `if ( ! restLink )` after the `expect.poll`, which is obsolete as the expect will throw if the `restLink` is `null`. Does that make sense or am I missing something?

swissspidy (approved):

packages/e2e-test-utils-playwright/src/request-utils/rest.ts
```diff
@@ -39,10 +40,30 @@ async function getAPIRootURL( request: APIRequestContext ) {
...
+				} catch ( error ) {
+					// Prints the error if the timeout is reached.
+					return error;
+				}
+
+				return nonce && rootURL ? true : false;
+			},
+			{
+				message: 'Failed to setup REST API.',
+				timeout: 60_000, // 1 minute.
```
kevin940726, line 65: Nit: Do we have a rough number when the REST API will be available? A minute seems like a long wait if it returns nothing. Is 5 seconds or 10 seconds enough?

packages/e2e-test-utils-playwright/src/request-utils/rest.ts
```diff
@@ -39,10 +40,30 @@ async function getAPIRootURL( request: APIRequestContext ) {
...
 
 async function setupRest( this: RequestUtils ): Promise< StorageState > {
-	const [ nonce, rootURL ] = await Promise.all( [
-		this.login(),
-		getAPIRootURL( this.request ),
-	] );
+	let nonce = '';
+	let rootURL = '';
+
+	await expect
```
kevin940726, line 48: A comment above explaining why will be great!

packages/e2e-test-utils-playwright/src/request-utils/rest.ts
WunderBart: Good question, and I have no idea TBH. Testing this thing takes ages, and it only fails in CI. Having said that, since this solution seems to be working across the board, I can lower the timeout value and do a couple more reruns. We can also address it in a follow-up PR since the failing perf tests keep blocking folks. What do you think?

test/performance/playwright.config.ts
```diff
@@ -14,9 +14,7 @@ process.env.ASSETS_PATH = path.join( __dirname, 'assets' );
 
 const config = defineConfig( {
 	...baseConfig,
-	reporter: process.env.CI
-		? './config/performance-reporter.ts'
-		: [ [ 'list' ], [ './config/performance-reporter.ts' ] ],
+	reporter: [ [ 'list' ], [ './config/performance-reporter.ts' ] ],
```
WunderBart, line 17: I think we should merge it with this change as it now prints the errors correctly, which would allow us to close https://github.com/WordPress/gutenberg/issues/60366. What do you think @Mamaduka @kevin940726?

Mamaduka: Sounds good. The performance test results are now attached to the summary so they won't be lost in the noise (#61450).

WunderBart: Ah, enabling the list reporter in this case actually means that only the error will be printed, not the tests themselves. It's because we're still using promisified `exec` and don't resolve with the `stdout` data. So, nothing in the output will change unless an error is thrown! 😄 

An example error looks like this now:

```
▶ Running tests
    > Using: WordPress v6.5
    > Suite: front-end-block-theme (round 1 of 1)
        > Branch: 9aec5fe38aa66962d1b0dc3c73bc22c7af08e3c5
            > Starting environment
            > Running tests

> gutenberg@18.5.0-rc.1 test:performance
> wp-scripts test-playwright --config test/performance/playwright.config.ts front-end-block-theme

Error: expect(received).not.toThrow()

Matcher error: received value must be a function

Received has value: undefined

Call Log:
- Timeout 60000ms exceeded while waiting on the predicate

   at ../../../packages/e2e-test-utils-playwright/src/request-utils/rest.ts:46

  44 | 	let rootURL = '';
  45 |
> 46 | 	await expect
     | 	^
  [47](https://github.com/WordPress/gutenberg/actions/runs/9385403378/job/25843486775#step:8:48) | 		.poll(
  48 | 			async () => {
  49 | 				[ nonce, rootURL ] = await Promise.all( [

    at RequestUtils.setupRest (/tmp/wp-performance-tests/tests/packages/e2e-test-utils-playwright/src/request-utils/rest.ts:46:2)
    at globalSetup (/tmp/wp-performance-tests/tests/test/performance/config/global-setup.ts:26:2)



Error: Command failed: npm run test:performance -- front-end-block-theme

    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at ChildProcess.exithandler (node:child_process:422:12)
    at ChildProcess.emit (node:events:[51](https://github.com/WordPress/gutenberg/actions/runs/9385403378/job/25843486775#step:8:52)9:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5) {
  code: 1,
  killed: false,
  signal: null,
  cmd: 'npm run test:performance -- front-end-block-theme'
```

Before it was printing just the `genericNodeError` part, and now all the necessary info is there.

Mamaduka: Good to know 😅

WunderBart: Pinging @jsnajdr, since it also means we don't need to [refactor to `spawn`](https://github.com/WordPress/gutenberg/pull/62270)! 😄

PR #62334, opened 2024-06-05
[Mobile] - Fix Inserter items list

geriux:
**Related PRs:**
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/6911

## What?
This PR restores the Inserter functionality on mobile.

## Why?
After recent changes introduced in https://github.com/WordPress/gutenberg/pull/62169 where now all blocks are shown in the inserter, we needed to change this behavior back in mobile since we don't have any categorization at the moment, so it was confusing to show blocks like Social Link or Button after the Core blocks.

## How?
By setting the `isQuick` param to `true` for the `useBlockTypesState` selector.

### Test case 1 - Social links block

- Open the editor
- Add a Social links block
- **Tap** on the inline appender
- Expect the inserter to open with only social link blocks

### Test case 2 - Buttons block

- Open the editor
- Add a Buttons block
- Tap on the first button
- Open the inserter using the Add block button in the toolbar
- Expect the inserter to open with only the Button block

### Test case 3 - Inner blocks

Smoke test inner blocks to double check other blocks can be added, for example in Group, Cover, and Columns blocks.

N/A

## Screenshots or screencast 

Before|After
-|-
<video src="https://github.com/WordPress/gutenberg/assets/4885740/063e9556-9ceb-4472-b91b-080c82744128" width=250 />|<video src="https://github.com/WordPress/gutenberg/assets/4885740/f9658bce-c2f6-4bb9-9e3b-e2b81d247392" width=250 />


twstokes (approved):
LGTM @geriux! I tested on an iOS sim. 🚀

PR #62274, opened 2024-06-04
Editor: Introduce the Editor component and use it in the site editor

youknowriad:
Related #52632 

## What?

This PR is the culmination of all the previous work done to unify post and site editors. It introduces a high-level `Editor` component in the editor package that have the following API:

```
<Editor postType={ postType } postId={ id } settings={ settings } />
```

This component renders a full functional editor (equivalent to what we see in post and site editors).

Note that there are still some extra props that we should ideally try to mitigate/remove over time. But for now these are necessary to account for the differences and history of post/site editor packages.

In the current PR, I'm only updating the site editor package to use that new component to keep the scope of the PR contained and help with testing/reviews.

## Testing Instructions

E2e tests should probably cover everything here but smoke testing editing various entities in the site editor would be helpful.

PR #62339, opened 2024-06-05
Editor: Use the Editor component in the post editor.

youknowriad:
Related #52632 
Similar to #62274 but for the post editor.

## What?

This PR updates the post editor to use the `Editor` component to render a full editor with UI and context, similar to how we do it in the site editor.

Note that there are still some extra props that we should ideally try to mitigate/remove over time. But for now these are necessary to account for the differences and history of post/site editor packages.

## Testing Instructions

E2e tests should probably cover everything here but smoke testing editing various entities in the site editor would be helpful.


packages/edit-post/src/store/private-selectors.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * WordPress dependencies
+ */
+import { createRegistrySelector } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
+import { store as editorStore } from '@wordpress/editor';
+
+export const getEditedPostTemplateId = createRegistrySelector(
```
youknowriad, line 8: This is the same logic that was in `getEditedPostTemplate` but I moved it to this new selector because we don't really need the full object.

ramonjd (approved):
Here's what I tested:

✅  Post editor loads find with all expected functionality
✅  Displaying/swapping/editing post/page templates works as expected
✅  General editing/adding blocks/styling/saving

Also tested:

✅  All the of the above in a Classic theme
✅  All of the above using a custom post type
✅  All of the above using an Author account
✅  All of the above for pages set as home/posts pages


> E2e tests should probably cover everything here but smoke testing editing various entities in the site editor would be helpful.

I assume `site editor` should be `post-editor`, but I smoke tested site editor and editing pages as well. 👍🏻

Issue #61653, opened 2024-05-14
Widgets page and Customizer: The inserter should not use ARIA tabs semantics


afercia:
### Description

In the Widgets page, the Inserter doesn't need tabs because it is only supposed to show the 'Blocks' panel.

https://github.com/WordPress/gutenberg/pull/61510 hid the entire panel header via CSS. The panel header contains the wrapper with `role="tablist"` that contains the 3 `role="tab"` panels. They're now hidden with `display: none` which makes them not perceivable by assistive technology. However, the panel with blocks still use a `role="tabpanel"`, which is not associated with any 'tab'. As such, the ARIA tabs pattern is semantically broken.

Instead of using CSS, the inserter in the Widgets page should render a dirrerent UI that doesn't use the ARIA tabs pattern in the first palce, as it was before the latest changes to reunify the editor.

Cc @youknowriad @MaggieCabrera @scruffian 

### Step-by-step reproduction instructions

- Activate Twenty Twenty-One.
- Go to Appearance > Widgets.
- Open the main inserter.
- Observe the main inserter doesn't have tabs (they are hidden via CSS together with the whole panel header).
- Inspect the source via your browser dev tools.
- Obaserve the panel element with class `block-editor-inserter__tabpanel` still has a `role="tabpanel"` attribute.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

afercia:
I updated this issue title to include the Customizer.

After latest changes, the 'inserter' panel in Customize > Widgets shows 3 tabs, including Patterns and Media. It is expected to only show the Blocks panel.

Also, after https://github.com/WordPress/gutenberg/pull/61421 the panel now has two 'Close' buttons. Screenshot:

On latest trunk:

![trunk](https://github.com/WordPress/gutenberg/assets/1682452/70b011fe-6048-468f-a041-eb31ad2684da)

On WordPress 6.5:

![wp 6 5](https://github.com/WordPress/gutenberg/assets/1682452/b27e4dce-34b6-47de-b6ea-816006bf8fe6)

As mentioned above, the unnecessary tabs can't be hidden with CSS. Instead, the panel should render a UI that doesn't use ARIA tabs.

Rishit30G:
Thanks for sharing the issue, 

> Also, after https://github.com/WordPress/gutenberg/pull/61421 the panel now has two 'Close' buttons. 

[...]

PR #61553, opened 2024-05-10
Add header title to list view sidebar

jeryj:
Alternative to https://github.com/WordPress/gutenberg/pull/61514.
Fixes https://github.com/WordPress/gutenberg/issues/59013.

This code is not ready, but should give an idea of what the experience would be like. It is only implemented for the list view and inserter sidebars. The Styles and Settings sidebars have the old style so you can compare the behaviors more easily. Before merging a change like this, we would need to address all the sidebars.

## What?

Fixes accessibility issues where focus order must match visual order, and also not be between the tabs and tab panel.

## Why?

Accessibility.

## How?

Moves Close button to before the tabs and add header title.

## Testing Instructions

- Open inserter, test close button
- Open List view, test close button

## Screenshots or screencast 
| Before | After |
| - | - |
| image | image |

PR #62343, opened 2024-06-05
Sidebar: Add a shared component for the inserter and list view

scruffian:
## What?
Creates a new `TabbedPanel` component that is used by both the inserter and the list view. 

## Why?
This will ensure visual and behaviour consistency between these two panels. In time we should also update the block inspector and global styles to use the component too.

## How?
- Create a new `TabbedPanel` component
- Refactor shared code to use the new component

## Testing Instructions
0. Open the Site Editor
1. Open the block inserter and check that the tabs behave the same way as in trunk
2. Open the list view and check that the tabs behave the same way as in trunk

Also check that the keyboard interactions are the same as in trunk.

Note: There is ongoing discussion about the best way to handle keyboard interactions with these panels. This PR aims to consolidate these panels to use the same code and approach. When we update the approach to keyboard interactions then we'll only need to do it in one place for all panels.

## Screenshots or screencast 

|  Before | After  |
|---|---|
| Screenshot 2024-06-17 at 12 39 27| Screenshot 2024-07-02 at 13 14 18 |

| Screenshot 2024-06-17 at 12 41 30 |  Screenshot 2024-07-02 at 13 14 30 |

The visual differences are:
 - The blue border under the tabs are now the same, which means the one in the inserter is now slightly wider
 - The tabs in the list view are very slightly wider


afercia:
I'd be totally in favor of a shared component to reduce duplication and have more consistency.

Worth noting the Inserter should not use an ARIA tabs interface when it shows only one kind of object. That's the case in the Widgets page and in the Customizer-widgets panel which only show Blocks. Could this component render an alternative ui with no ARIA tabs semantics when there's only one kind of objects? The implementation that as in place before the recent editor reunification effort did use an alternative UI for this case but that went lost.
See https://github.com/WordPress/gutenberg/issues/61653

DaniGuardiola:
Inner styles should not be overriden for `@wordpress/components`, in general. If we want a version of tabs that stretches to fill its container, that should be requested and implemented (if agreed upon) in the component itself, as a variant. Style overrides are very fragile and make maintenance of the underlying component pretty painful.

Separately, there are some valid arguments against these full-width styles, like https://github.com/WordPress/gutenberg/pull/61974#issuecomment-2149719264

While I think a component like this that unifies styles is a good addition (without having looked too deep into the details), I think the conversation in #61974 should be resolved first before committing to one or the other. If full-width is the decision, then there should be work to implement those styles in `Tabs` before merging this. cc @mirka @ciampo

scruffian:
We are already overriding these styles in the individual components, so this doesn't break that convention any more than it's already been broken, it just centralizes that code.

DaniGuardiola:
@scruffian I think that's reasonable. Please note that the first example in the description has an intentional design detail where the tabs indicator overlaps the border below it, as can be seen in the "before" image. If we unify all of these (and the ones you mention for the future), we should also make a unified decision regarding that detail. cc @WordPress/gutenberg-design

jasmussen:
Good catch, @DaniGuardiola, that is definitely intentional, and is about ensuring legibility of the active state, so that's an important detail to retain.

packages/block-editor/src/components/tabbed-sidebar/index.js
```diff
@@ -0,0 +1,71 @@
...
+				selectOnMove={ false }
+				defaultTabId={ defaultTabId }
+				onSelect={ onSelect }
+				selectedTabId={ selectedTab }
+			>
+				<div className="block-editor-tabbed-sidebar__tablist-and-close-button">
+					<Button
+						className="block-editor-tabbed-sidebar__close-button"
+						icon={ closeSmall }
+						label={ __( 'Close block inserter' ) }
```
jeryj: This will need to be dynamic, as the list view close button tooltip also says "Close block inserter"

packages/block-editor/src/components/index.js
```diff
@@ -162,6 +162,7 @@ export { default as __experimentalPublishDateTimePicker } from './publish-date-t
 export { default as __experimentalInspectorPopoverHeader } from './inspector-popover-header';
 export { default as BlockPopover } from './block-popover';
 export { useBlockEditingMode } from './block-editing-mode';
+export { default as TabbedSidebar } from './tabbed-sidebar';
```
jeryj: This is a helpful refactor to work towards unified sidebars, so I don't want to block this progress. That said, I think it will be important to not combine the content of the sidebar with the interaction of the sidebar wrapper. I see the eventual goal as having a shared sidebar component (or another name that does not reference the visual position of the component) with a close button. The content within it can be whatever it needs to be. 

This will be possible if we move to a header title for each sidebar, as proposed in https://github.com/WordPress/gutenberg/pull/61553

packages/block-editor/README.md
```diff
@@ -804,6 +804,10 @@ _Related_
 
 -   <https://github.com/WordPress/gutenberg/blob/HEAD/packages/data/README.md#registerStore>
 
+### TabbedSidebar
```
jeryj: Should this be a public component?

packages/block-editor/src/components/tabbed-sidebar/README.md
```diff
@@ -0,0 +1,76 @@
...
+				name: 'slug-2',
+				title: _x( 'Title 2', 'context' ),
+				panel: <PanelContents />,
+			},
+		] }
+		onClose={ onClickCloseButton }
+		onSelect={ onSelectTab }
+		defaultTabId="slug-1"
+		ref={ tabsRef }
+	/>
```
jeryj, line 39: To elaborate on the above, I think this should eventually look more like:

```
<StickyPanel title="Block Inserter" onClose={ onClickCloseButton }>
  <Tabs>
    ...
   </Tabs>
</StickyPanel>
```
Name tbd 😬 😅

jeryj (commented):
Love all the red of removed code :) 

With the zoom-out pattern experiment enabled, the patterns list is not positioned correctly. 
Patterns sidebar shown to the right of a 300px wide gap of white space

DaniGuardiola:
Just wanna say two things:

- Great work, thank you for doing this!
- This is a great opportunity to make sure this component is rock-solid in terms of a11y. Let's please ensure that we're very confident about this aspect before merging. This includes any nuance about ARIA roles and interactivity (like @afercia mentions in this comment: https://github.com/WordPress/gutenberg/pull/62343#issuecomment-2178084836), and thorough testing, as well as a sufficient strategy for integration with the rest of the app in a way that doesn't break a11y. While this work doesn't necessarily need to involve me, I'm happy to engage if it helps.

jeryj:
@DaniGuardiola - to improve the accessibility, we'd need design agreement for [adding a header with title and close button to all the sidebars](https://github.com/WordPress/gutenberg/pull/61553). I would love if this could get momentum. It has agreement and support from folks on the WordPress Accessibility team.

DaniGuardiola:
@jeryj love that! This PR is also a great chance to add that as a follow-up, and get it going in all sidebars.

youknowriad:
Based on the discussion on the linked PR, it seems a consensus is going to be hard to reach for that design. Can we instead focus on the unification part in this PR while we continue iterating on the design separately.

DaniGuardiola:
I agree with @youknowriad, let's leave the visual design discussion separate from this PR.

Let's still make sure this component is accessible to all users, please :)

tyxla:
> Can we instead focus on the unification part in this PR while we continue iterating on the design separately.

> Let's still make sure this component is accessible to all users, please :)

While I think we should ensure the component is accessible, I agree with @youknowriad and believe any accessibility improvements could be made separately as they aren't necessarily part of the work to unify the existing components in question.

packages/block-editor/src/components/tabbed-sidebar/README.md
DaniGuardiola: Agree on this. We should try to keep things composable to avoid ["YAP" syndrome](https://kyleshevlin.com/quit-your-yapping/).

jeryj:
Agreed :) I think any changes to the styles should be a follow-up. 

After this is merged, can this tabbed sidebar component be applied to the Settings panel as well?

scruffian:
> @scruffian I think that's reasonable. Please note that the first example in the description has an intentional design detail where the tabs indicator overlaps the border below it, as can be seen in the "before" image. If we unify all of these (and the ones you mention for the future), we should also make a unified decision regarding that detail. cc

Thanks. I have moved the indicator so it overlaps the border and updated the screenshots.

scruffian:
I have updated this based on the feedback above. I have made it a locked component to give us time to iterate on it.

> After this is merged, can this tabbed sidebar component be applied to the Settings panel as well?

Definitely

jeryj (approved):
I think this is better than what we have since it consolidates the code more and can also be applied to the Settings sidebar. However, I do have concerns over how the Tabs are tied to a close button that controls the sidebar outside of it. It feels like two unrelated things are being forced together. But, pragmatically, this is better than what we have and the component is private so we should be able to refactor it when the time comes.

scruffian:
I'm going to merge this so that we can continue to align the various panels to use this component and ensure that improvments are made to all at once. I am very happy for others to jump in and make improvments - for example making the component more composable etc.

Issue #55638, opened 2023-10-26
Post Date block available date formats should be extendable


fabiankaegy:
## What problem does this address?

The Post Date block exposes six format options for users to choose from. Sadly, these six options are hardcoded in the Gutenberg codebase and cannot be replaced with custom options. 

Many publishers have custom style guides for how they are outputting dates. And while the block does use the site setting by default, the available options cannot be altered.

It is also possible to manually insert a custom date format in the block. But that has to be done manually every single time for every block where the setting should be used which is not a great solution for large editorial teams. 

## What is your proposed solution?

Make it possible for extenders to alter the available data formats exposed in the format picker. They are currently hardcoded here: 

https://github.com/WordPress/gutenberg/blob/8aaac1ac800ea5e4f5ac20f903c21d1c54477dd7/packages/block-editor/src/components/date-format-picker/index.js#L79-L89

youknowriad:
Could these be theme.json settings (just like color presets or things like that)?

fabiankaegy:
@youknowriad I think so yes!

PR #62345, opened 2024-06-05
Post Date: Extendable Suggested Formats

TylerB24890:
## What?

This PR adds a `date.formatOptions` property to the `__experimentalFeatures` object in the editor settings. The `formatOptions` value is then used as the suggested date formats in the `core/post-date` block.

**NOTE:** If `__experimentalFeatures` is not the correct key this kind of setting should live under please let me know. This might be better off living in the date settings from `@wordpress/date` 

## Why?

**Related Issue:** https://github.com/WordPress/gutenberg/issues/55638

This PR makes the hard-coded suggested date formats extendable so extenders can choose which date formats are allowed for selecting. If a unique date format is required, extenders can add it to this setting and have it selectable instead of requiring editors to use the Custom Format option every time.

## How?

By applying the `date.formatOptions` to the Block Editor Settings, extenders can use the `block_editor_settings_all` filter to change the suggested date formats available to editors.

## Testing Instructions

1. Create a new Post with the block editor.
2. Add the `core/post-date` block
3. Note the available formats in the block sidebar.
4. In `functions.php` (or similar) add the following:
```php
add_filter(
    'block_editor_settings_all',
    function ( $settings ) {
        $settings['__experimentalFeatures']['date']['formatOptions'] = [
            'M j',
            'F j, Y'
        ];

        return $settings;
    }
);
```
5. Refresh the block editor & confirm the `core/post-date` block Format Options are reduced to only the two defined (plus the "Custom" option).
    - If you used the same Date formats defined above, you should see the following format options available:
        - `Jan 25`
        - `January 25, 2022`


packages/block-editor/src/components/date-format-picker/index.js
```diff
@@ -76,24 +82,9 @@ function NonDefaultControls( { format, onChange } ) {
...
-			_x( 'M j, Y', 'medium date format' ),
-			/* translators: See https://www.php.net/manual/datetime.format.php */
-			_x( 'M j, Y g:i A', 'medium date format with time' ),
-			/* translators: See https://www.php.net/manual/datetime.format.php */
-			_x( 'F j, Y', 'long date format' ),
-			/* translators: See https://www.php.net/manual/datetime.format.php */
-			_x( 'M j', 'short date format without the year' ),
-		] ),
-	];
+	const editorSettings = select( blockEditorStore ).getSettings();
```
fabiankaegy: @TylerB24890 can you refactor this to use `useSelect` instead? There are some architectural differences and the usage of select itself is discouraged

TylerB24890: All set.

Issue #61348, opened 2024-05-03
getEntityRecords returns empty list for custom endpoint


MatzeKitt:
### Description

I have a custom API endpoint to request the user roles of a user, which returns basically this:

```php
$user = wp_get_current_user();
return rest_ensure_response( $user->roles );
```

This endpoint is accessible via `/namespace/v1/roles` and works fine.

Then, I use this endpoint in a `PluginSidebar` like so:

```jsx
const SidebarContent = compose( [
	withDispatch( ( dispatch ) => {
		// allow using a custom REST API
		dispatch( 'core' ).addEntities( [
			{
				name: 'user_roles',
				kind: 'namespace',
				baseURL: '/namespace/v1/roles',
			},
		] );
		
		return null;
	} ),
] )( ( props ) => {
	const data = useSelect( ( select ) => {
		return {
			// request the REST API
			roles: select( 'core' ).getEntityRecords( 'namespace', 'user_roles', { per_page: 100 } ),
		}
	} );
	console.log({data});
	
	return null;
} )

registerPlugin(
	'my-sidebar',
	<div>
		<SidebarContent />
	</div>
);
```

 The `data.roles` is just an empty array despite the API endpoint returns the correct data as array (e.g. `["subscriber", "administrator"]`), as also seen in the network tab in the developer tools.

### Step-by-step reproduction instructions

1. Create custom API endpoint
2. Create plugin sidebar from above
3. Check the output of `data.roles`, which is empty

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

Mamaduka:
I haven't started debugging this yet, but I would guess that the returned response "shape" is the cause, which is usually an array of objects with a unique key.

The `settings` endpoint might be similar to yours; it exposes a single entity for a site and doesn't support "plural" selectors.

```js
// Returns null.
wp.data.select( 'core' ).getEntityRecords( 'root', 'site' );

// Returns settings object.
wp.data.select( 'core' ).getEntityRecord( 'root', 'site' );
```

Can you test if `wp.data.select( 'core' ).getEntityRecord` returns values properly?

MatzeKitt:
Using `getEntityRecord`, I just get a single role returned, even though the user has multiple roles.

Changing the API output to an array of objects and use `getEntityRecords` again also returns an empty array:

```json
[{
    "role": "editor"
}, {
    "role": "administrator"
}]

```

[...]

PR #62346, opened 2024-06-05
Core Data: Fix errors when the entities list doesn't contain config key

### LINKED ISSUE #62311: Data: Stores become unresponsive when batched actions trigger an error

Mamaduka:
## What?
Closes #62311.

PR fixes errors when data returned by the endpoint for `getEntityRecords` doesn't contain config or default `key` property.

## Testing Instructions
1. Open a post or page.
2. Run selector via Console - `wp.data.select( 'core' ).getEntityRecords( 'root', 'site' );.`
3. Check the errors via - `wp.data.select( 'core' ).getResolutionError( 'getEntityRecords', [ 'root', 'site' ] );`.
4. You can also set a logpoint [here](https://github.com/WordPress/gutenberg/blob/0011ca615e4d5424e7220fc363a443bf62fe0852/packages/core-data/src/resolvers.js#L299). 

Same.

### DIFF
--- a/packages/core-data/src/queried-data/reducer.js
+++ b/packages/core-data/src/queried-data/reducer.js
@@ -110,7 +110,11 @@ export function items( state = {}, action ) {
 				[ context ]: {
 					...state[ context ],
 					...action.items.reduce( ( accumulator, value ) => {
-						const itemId = value[ key ];
+						const itemId = value?.[ key ];
+						if ( ! itemId ) {
+							return accumulator;
+						}
+
 						accumulator[ itemId ] = conservativeMapItem(
 							state?.[ context ]?.[ itemId ],
 							value
@@ -164,7 +168,10 @@ export function itemIsComplete( state = {}, action ) {
 				[ context ]: {
 					...state[ context ],
 					...action.items.reduce( ( result, item ) => {
-						const itemId = item[ key ];
+						const itemId = item?.[ key ];
+						if ( ! itemId ) {
+							return result;
+						}
 
 						// Defer to completeness if already assigned. Technically the
 						// data may be outdated if receiving items for a field subset.
@@ -232,7 +239,7 @@ const receiveQueries = compose( [
 	return {
 		itemIds: getMergedItemIds(
 			state?.itemIds || [],
-			action.items.map( ( item ) => item[ key ] ),
+			action.items.flatMap( ( item ) => item?.[ key ] ?? [] ),
 			page,
 			perPage
 		),
--- a/packages/core-data/src/queried-data/reducer.js
+++ b/packages/core-data/src/queried-data/reducer.js
@@ -111,9 +111,6 @@ export function items( state = {}, action ) {
 					...state[ context ],
 					...action.items.reduce( ( accumulator, value ) => {
 						const itemId = value?.[ key ];
-						if ( ! itemId ) {
-							return accumulator;
-						}
 
 						accumulator[ itemId ] = conservativeMapItem(
 							state?.[ context ]?.[ itemId ],
@@ -169,9 +166,6 @@ export function itemIsComplete( state = {}, action ) {
 					...state[ context ],
 					...action.items.reduce( ( result, item ) => {
 						const itemId = item?.[ key ];
-						if ( ! itemId ) {
-							return result;
-						}
 
 						// Defer to completeness if already assigned. Technically the
 						// data may be outdated if receiving items for a field subset.
--- a/packages/core-data/src/queried-data/reducer.js
+++ b/packages/core-data/src/queried-data/reducer.js
@@ -233,7 +233,7 @@ const receiveQueries = compose( [
 	return {
 		itemIds: getMergedItemIds(
 			state?.itemIds || [],
-			action.items.flatMap( ( item ) => item?.[ key ] ?? [] ),
+			action.items.map( ( item ) => item?.[ key ] ).filter( Boolean ),
 			page,
 			perPage
 		),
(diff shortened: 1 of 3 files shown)


Mamaduka:
I think I broke a few things, I will investigate them tomorrow.

packages/core-data/src/queried-data/reducer.js
```diff
@@ -110,7 +110,8 @@ export function items( state = {}, action ) {
 				[ context ]: {
 					...state[ context ],
 					...action.items.reduce( ( accumulator, value ) => {
-						const itemId = value[ key ];
+						const itemId = value?.[ key ];
```
Mamaduka, line 113: It turns out that `undefined` is a "valid" item ID, and site settings are keyed using it.

## Screenshot
CleanShot 2024-06-06 at 00 13 33

packages/core-data/src/queried-data/reducer.js
```diff
@@ -232,7 +233,7 @@ const receiveQueries = compose( [
 	return {
 		itemIds: getMergedItemIds(
 			state?.itemIds || [],
-			action.items.map( ( item ) => item[ key ] ),
+			action.items.flatMap( ( item ) => item?.[ key ] ?? [] ),
```
tyxla: Won't `action.items` be an array of objects? If that's the case, it feels odd if `item?.[ key ]` falls back to an empty array.

tyxla (commented):
While the proposed changes look mostly safe, I wasn't able to reproduce the original error, so I can't confirm that this PR resolves it. I've been following the testing instructions from this PR on `trunk`, but no error appears. Any idea what I might be missing?

Mamaduka:
@tyxla, if you log errors [here](https://github.com/WordPress/gutenberg/blob/0011ca615e4d5424e7220fc363a443bf62fe0852/packages/core-data/src/resolvers.js#L299) on trunk and call the selector from testing instructions you should be able to reproduce the error.

packages/core-data/src/queried-data/reducer.js
Mamaduka: This line extracts the `nextItemIds` array. It's usually an array of `key` items. When every key is `undefined` we want to pass an empty array.

This `flatMap` pattern is from [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#mapped_array_contains_undefined). It's the same as `items.map( ( item ) => item?.[ key ] ).filter( ( itemId ) => itemId !== undefined );`

Should I use explicit iteration?

tyxla: Thanks for confirming. I think this is too idiomatic for most tastes though - I know the pattern and still didn't think of it when I first saw it. I think when one sees `.flatMap()` they expect this works with nested arrays, and this isn't the case here, so it could create false expectations. The fact that this area of the code isn't properly typed doesn't make it better. So yes, if I could choose, I'd prefer being more explicit here.

tyxla:
> @tyxla, if you log errors [here](https://github.com/WordPress/gutenberg/blob/0011ca615e4d5424e7220fc363a443bf62fe0852/packages/core-data/src/resolvers.js#L299) on trunk and call the selector from testing instructions you should be able to reproduce the error.

Correct, when I add logging there, I see the error logged, but I still don't understand in what circumstances it will actually be manifested as an error - it's supposed to be caught there, no?

Mamaduka:
The error can occur when a response for `getEntityRecords` doesn't return a correctly shaped object. The requirement should be a list of objects containing a primary key (defined in config, defaulting to `id`). Related discussion: #61348.

The example site entity used here returns a single object with properties for the `/settings` endpoint, which triggers an error in reducers.

packages/core-data/src/queried-data/reducer.js
```diff
@@ -232,7 +233,9 @@ const receiveQueries = compose( [
 	return {
 		itemIds: getMergedItemIds(
 			state?.itemIds || [],
-			action.items.map( ( item ) => item[ key ] ),
+			action.items
+				.map( ( item ) => item?.[ key ] )
+				.filter( ( itemId ) => itemId ),
```
tyxla: `filter( Boolean )` is yet another idiomatic way to do the same thing:

```suggestion
				.filter( Boolean ),
```

tyxla (approved):
Changes look safe and good to go 👍 

Thanks @Mamaduka 🚀

packages/core-data/src/queried-data/reducer.js
Mamaduka: Yeah, that's a nice shorthand. I went back and forth with filter's predicate, then decided to keep the current version.

tyxla: Up to you, it's not a big deal. Thanks 👍

Mamaduka: I had to rebase and included the update.

ellatrix:
Worth backporting?

Mamaduka:
This is probably not a new bug, but it's worth backporting alongside the #62322.

ellatrix:

PR #62350, opened 2024-06-05
Add `html` to reset styles for the `.editor-styles-wrapper` container

### LINKED ISSUE #57176: Regression: iframed post editor can show some admin styles with classic themes

sabernhardt:
Fixes: #57176

(similar to option in #61959, but only increasing the specificity for one selector instead of the whole group)

## What?
Raise the specificity of only the `:where(.editor-styles-wrapper)` selector in `reset.scss` so that the CSS reset overrides styles in [wp-admin/css/common.css](https://github.com/WordPress/wordpress-develop/blob/96de28cc29774aeb4577b0d314e6b1724c1804ae/src/wp-admin/css/common.css#L223-L230).

## Why?
The admin CSS bleed affects the iframed post editor in 'classic' themes that do not specify `font-size`, `line-height` and/or `background-color` for the `body`. In addition to the extra discrepancies between editor and front end, the smaller text is more difficult to read when editing post content, and the background reduces color contrast.

## How?
1. Add `html` to the `:where(.editor-styles-wrapper)` ruleset (in the compiled stylesheet).
2. Remove inaccurate comment that says the reset is unnecessary in an iframe. Enqueuing the admin styles makes the reset necessary.

## Testing Instructions
1. Activate a 'classic' theme such as Twenty Thirteen or Twenty Fourteen (those two currently have the wrong font size and background color).
2. Create a new post and add blocks including a Paragraph and/or List.
3. Check whether the post editor is using an iframe by using the [browser inspector](https://blog.hubspot.com/website/how-to-inspect) on text in a Paragraph block. If it is **not** inside an iframe, you may need to save the post and then deactivate plugins or go to Preferences and remove a panel such as Custom Fields.
4. In the browser inspector, look for styles on the `body` element coming from `common.css` (or `load-styles.php` if `SCRIPT_DEBUG` is `false`). With the PR applied, the `background`, `color`, `font-family`, `font-size` and `line-height` properties should have a line through them (the `min-width` is not reset, but that seems to be fine).
admin body styles crossed out, except for min-width


packages/block-library/src/reset.scss
```diff
@@ -8,19 +8,18 @@
...
 
-	// Many themes with white backgrounds load editor styles but fail to also provide
-	// an explicit white background color, assuming a white editing canvas.
-	// So to match browser defaults, we provide a white default here as well.
-	background: #fff;
+		// Many themes with white backgrounds load editor styles but fail to also provide
+		// an explicit white background color, assuming a white editing canvas.
+		// So to match browser defaults, we provide a white default here as well.
+		background: #fff;
+	}
```
ellatrix, line 22: The admin styles seem to specifically target the `body`. Could we move these particular styles out of the `here(.editor-styles-wrapper)` block and under `body`, maybe with a link to which styles are being reset?

Also when did this break? Where you able to pinpoint the specific commit or release?

ellatrix: Cc @aaronrobertshaw in case it rings a bell

sabernhardt: The selector needs at least `0-0-1` specificity to override `body`, which it had before [removing `html` in 46752](https://github.com/WordPress/gutenberg/pull/46752/files#diff-bda8e417e9cfc8724167f4c82aab1d0d205245c51360a8afa4bba99ba078e6dbR10).

I made #61959 to restore that selector exactly as it was for the whole set, and Aaron already suggested to [isolate the change](https://github.com/WordPress/gutenberg/pull/61959#discussion_r1618026389) to only the first selector of the stylesheet.

ellatrix: In that case, we should probably change the selector to:

```css
div:where(.editor-styles-wrapper),
body:where(.editor-styles-wrapper)
```

So that it overrides the admin styles for both an iframed and non-iframe editor.

sabernhardt: `html :where(.editor-styles-wrapper)` applied to either element in WordPress 6.3, and I just made a quick edit _directly_ in `wp-includes\css\dist\block-library\reset.css` with 6.6 beta to confirm that it still covers both elements.

iframe:

iframe editor with corrected background and font size

non-iframe:

non-framed editor with element inspector open

(the theme for these screenshots is Twenty Thirteen)

ellatrix: Why not just restore it to https://github.com/WordPress/gutenberg/pull/46752/files#diff-bda8e417e9cfc8724167f4c82aab1d0d205245c51360a8afa4bba99ba078e6dbR10?

sabernhardt: Restoring is what I tried first :) #61959

ellatrix: Yes, so we only want it to affect the top level styles. I guess I meant, why not take out that section and do it in a separate block? I find that a bit more readable than `html &`.

ellatrix (approved):

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 9f0380fcba

PR #56934, opened 2023-12-11
[RNMobile] Add OfflineStatus component

derekblank:
## What?
Adds a mobile component to display a visual indicator when working offline within the mobile editor. (Note: the functionality to subscribe to network status is handled in https://github.com/WordPress/gutenberg/pull/56861.)

See related discussions in a [previous attempt](https://github.com/WordPress/gutenberg/pull/56848).

## Why?
Overall, to improve the user's experience when uploading media. 

Addresses:
* https://github.com/wordpress-mobile/gutenberg-mobile/issues/6405

## How?
Adds OfflineStatus component and associated icon/style files. 

## Testing Instructions
The component is included in block-list behind a dev flag. To test this PR, build the app locally and turn on Airplane Mode while using the editor.

<details><summary>Test Diff</summary>

```diff
diff --git a/packages/block-editor/src/components/block-list/index.native.js b/packages/block-editor/src/components/block-list/index.native.js
index 810e23e4c1..19540eddb4 100644
--- a/packages/block-editor/src/components/block-list/index.native.js
+++ b/packages/block-editor/src/components/block-list/index.native.js
@@ -28,6 +28,7 @@ import {
 	DEFAULT_BLOCK_LIST_CONTEXT,
 } from './block-list-context';
 import { BlockDraggableWrapper } from '../block-draggable';
+import OfflineStatus from '../offline-status';
 import { useEditorWrapperStyles } from '../../hooks/use-editor-wrapper-styles';
 import { store as blockEditorStore } from '../../store';
 
@@ -235,6 +236,7 @@ export default function BlockList( {
 			onLayout={ onLayout }
 			testID="block-list-wrapper"
 		>
+			<OfflineStatus />
 			{ isRootList ? (
 				<BlockListProvider
 					value={ {
```

</details>

## Screenshots or screencast 
Light | Dark
-|-
 |

PR #62353, opened 2024-06-06
Icons: Fix React warning error for offline icon

t-hamano:
Related to #56934

## What?

This PR fixes the react warning error that appears when using the offline icon.

```
Warning: Invalid DOM property `fill-rule`. Did you mean `fillRule`?
Warning: Invalid DOM property `clip-rule`. Did you mean `clipRule`?
```
icon

## How?

This can be solved by changing the attribute name to `fillRule`, `clipRule`, but since the icon did not change even without this attribute, I deleted it.

## Testing Instructions

- Make sure you don't get any browser warning errors when using code like the one below.
  ```javascript
  import { useBlockProps } from '@wordpress/block-editor';
  import { Icon, offline } from '@wordpress/icons';

  export default function Edit() {
  	const blockProps = useBlockProps();
  	return (
  		<div { ...blockProps }>
  			<Icon icon={ offline } />
  		</div>
  	);
  }
  ```
- The icon should remain the same in the mobile app.


dcalhoun (approved):
The icon successfully rendered on iOS and Android in my testing. Thanks!

t-hamano:
Thanks for the review!

PR #62354, opened 2024-06-06
List view: show context menu for content-only blocks in posts

### LINKED ISSUE #62352: List view: empty settings menus for non-editable items instead of a lock icon
### Description

Block settings and other actions can be accessed via an ellipsis menu for items in the list view.

When a block is locked there are no actions/settings, however an empty drop down menu still appears (in [BlockSettingsMenu](https://github.com/WordPress/gutenberg/blob/578f7d019635576f0a1ddf8abef9362327da2af9/packages/block-editor/src/components/list-view/block.js#L609-L609)).

Previously there was a lock icon. See: https://github.com/WordPress/gutenberg/blob/578f7d019635576f0a1ddf8abef9362327da2af9/packages/block-editor/src/components/list-view/block-select-button.js#L64-L64

1. Go to any page or post in the post editor
2. In the Post/Page settings on the right, activate "Show template"
3. Open the list view and click on the ellipsis menu of non-editable items, e.g., template parts, content or title blocks
4. Observe that the menu is empty

https://github.com/WordPress/gutenberg/assets/6458278/f2d87ab3-fb05-4132-9ec5-6127333889c4

Screenshot 2024-06-06 at 2 23 16 PM

### Environment info

Latest Gutenberg trunk on WP 6.6-beta1-58338

Also on 6.6-beta1-58338 with Gutenberg deactivated

Yes

Yes

ramonjd:
## What?
Maybe resolves https://github.com/WordPress/gutenberg/issues/62352

Ensures the Edit template context menu is shown for all post types in the post editor/site editor pages.

Display a not-very-pretty dialogue box where a user cannot edit a template.

## Why?

Since https://github.com/WordPress/gutenberg/pull/61127, a context menu is displayed for list view items that a locked, e.g., template parts and blocks

Previously the context menu was only shown for pages, and there was no check if the user can edit template. 

## How?

Rather than check for a post type, just look for a `templateId`.

A `templateId` will exists for post types that have a template and not for templates, patterns, template parts themselves.

## Testing Instructions

As admin, open up a post in the post editor. In the post settings, click on "Show template"

Screenshot 2024-06-06 at 3 53 47 PM

Open the list view and click on the ellipsis menu for template-related items, e.g., Title, Content and any template parts like a Header.

https://github.com/WordPress/gutenberg/assets/6458278/c204597d-a4fa-453a-9399-74386179fcef

Ensure that you can see the edit context menu above. 

Template parts and other page/post template blocks such as post title, content and featured images should have the "Edit template" context menu.

Locked items in the page/post content will get the "Unlock" context menu.

Check that all the above applies for pages in the post editor and the site editor.

Now log in as an Author, create a new post. In the post settings, click on "Show template".

Open the list view and click on the ellipsis menu for template-related items.

You should see a message that explains that you can't edit the block.

Screenshot 2024-06-07 at 4 17 51 PM

### DIFF
--- a/packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
+++ b/packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
@@ -11,7 +11,6 @@ import { store as coreStore } from '@wordpress/core-data';
 import { __experimentalText as Text, MenuItem } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
-import { lockSmall as lock } from '@wordpress/icons';
 
 /**
  * Internal dependencies
@@ -80,32 +79,20 @@ function ContentOnlySettingsMenuItems( { clientId, onClose } ) {
 		);
 	}
 
-	// The post has a parent template part, but the user does not have the capability to edit template parts.
+	const isPattern = entity.type === 'wp_block';
+	let helpText = isPattern
+		? __(
+				'Edit the pattern to move, delete, or make further changes to this block.'
+		  )
+		: __(
+				'Edit the template to move, delete, or make further changes to this block.'
+		  );
+
 	if ( ! canEditTemplates ) {
-		return (
-			<>
-				<BlockSettingsMenuFirstItem>
-					<MenuItem
-						icon={ lock }
-						iconPosition="left"
-						role="none"
-						disabled
-					>
-						{ __( 'Locked' ) }
-					</MenuItem>
-				</BlockSettingsMenuFirstItem>
-				<Text
-					variant="muted"
-					as="p"
-					className="editor-content-only-settings-menu__description"
-				>
-					{ __( 'Sorry, you cannot edit this block.' ) }
-				</Text>
-			</>
+		helpText = __(
+			'Only users with permissions to edit the template can move or delete this block'
 		);
 	}
-
-	const isPattern = entity.type === 'wp_block';
 
 	return (
 		<>
@@ -117,6 +104,7 @@ function ContentOnlySettingsMenuItems( { clientId, onClose } ) {
 							postType: entity.type,
 						} );
 					} }
+					disabled={ ! canEditTemplates }
 				>
 					{ isPattern ? __( 'Edit pattern' ) : __( 'Edit template' ) }
 				</MenuItem>
@@ -126,13 +114,7 @@ function ContentOnlySettingsMenuItems( { clientId, onClose } ) {
 				as="p"
 				className="editor-content-only-settings-menu__description"
 			>
-				{ isPattern
-					? __(
-							'Edit the pattern to move, delete, or make further changes to this block.'
-					  )
-					: __(
-							'Edit the template to move, delete, or make further changes to this block.'
-					  ) }
+				{ helpText }
 			</Text>
 		</>
 	);
--- a/packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
+++ b/packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
@@ -20,62 +20,68 @@ import { store as editorStore } from '../../store';
 import { unlock } from '../../lock-unlock';
 
 function ContentOnlySettingsMenuItems( { clientId, onClose } ) {
-	const { entity, onNavigateToEntityRecord, canEditTemplateParts } =
-		useSelect(
-			( select ) => {
-				const {
-					getBlockEditingMode,
-					getBlockParentsByBlockName,
-					getSettings,
-					getBlockAttributes,
-				} = select( blockEditorStore );
-				const contentOnly =
-					getBlockEditingMode( clientId ) === 'contentOnly';
-				const _canEditTemplateParts = select( coreStore ).canUser(
-					'create',
-					'templates'
-				);
-				if ( ! contentOnly ) {
-					return {
-						canEditTemplateParts: _canEditTemplateParts,
-					};
-				}
-				const patternParent = getBlockParentsByBlockName(
-					clientId,
-					'core/block',
-					true
-				)[ 0 ];
+	const { entity, onNavigateToEntityRecord, canEditTemplates } = useSelect(
+		( select ) => {
+			const {
+				getBlockEditingMode,
+				getBlockParentsByBlockName,
+				getSettings,
+				getBlockAttributes,
+			} = select( blockEditorStore );
+			const contentOnly =
+				getBlockEditingMode( clientId ) === 'contentOnly';
+			if ( ! contentOnly ) {
+				return {};
+			}
+			const patternParent = getBlockParentsByBlockName(
+				clientId,
+				'core/block',
+				true
+			)[ 0 ];
 
-				let record;
-				if ( patternParent ) {
+			let record;
+			if ( patternParent ) {
+				record = select( coreStore ).getEntityRecord(
+					'postType',
+					'wp_block',
+					getBlockAttributes( patternParent ).ref
+				);
+			} else {
+				const { getCurrentTemplateId } = select( editorStore );
+				const templateId = getCurrentTemplateId();
+				if ( templateId ) {
 					record = select( coreStore ).getEntityRecord(
 						'postType',
-						'wp_block',
-						getBlockAttributes( patternParent ).ref
+						'wp_template',
+						templateId
 					);
-				} else {
-					const { getCurrentTemplateId } = select( editorStore );
-					const templateId = getCurrentTemplateId();
-					if ( templateId ) {
-						record = select( coreStore ).getEntityRecord(
-							'postType',
-							'wp_template',
-							templateId
-						);
-					}
 				}
-				return {
-					canEditTemplateParts: _canEditTemplateParts,
-					entity: record,
-					onNavigateToEntityRecord:
-						getSettings().onNavigateToEntityRecord,
-				};
-			},
-			[ clientId ]
+			}
+			const _canEditTemplates = select( coreStore ).canUser(
+				'create',
+				'templates'
+			);
+			return {
+				canEditTemplates: _canEditTemplates,
+				entity: record,
+				onNavigateToEntityRecord:
+					getSettings().onNavigateToEntityRecord,
+			};
+		},
+		[ clientId ]
+	);
+
+	if ( ! entity ) {
+		return (
+			<TemplateLockContentOnlyMenuItems
+				clientId={ clientId }
+				onClose={ onClose }
+			/>
 		);
+	}
 
 	// The post has a parent template part, but the user does not have the capability to edit template parts.
-	if ( ! canEditTemplateParts && entity ) {
+	if ( ! canEditTemplates ) {
 		return (
 			<>
 				<BlockSettingsMenuFirstItem>
@@ -96,15 +102,6 @@ function ContentOnlySettingsMenuItems( { clientId, onClose } ) {
 					{ __( 'Sorry, you cannot edit this block.' ) }
 				</Text>
 			</>
-		);
-	}
-
-	if ( ! entity ) {
-		return (
-			<TemplateLockContentOnlyMenuItems
-				clientId={ clientId }
-				onClose={ onClose }
-			/>
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
```diff
@@ -19,53 +20,81 @@ import { store as editorStore } from '../../store';
...
+					canEditTemplateParts: _canEditTemplateParts,
+					entity: record,
+					onNavigateToEntityRecord:
+						getSettings().onNavigateToEntityRecord,
+				};
+			},
+			[ clientId ]
+		);
+
+	if ( ! canEditTemplateParts ) {
```
ramonjd: I also looked into just showing a lock icon like Gutenberg did before.

I can't do a `canUser` check in the [list item component](https://github.com/WordPress/gutenberg/blob/ffb3231caae60239629e995362cc74c5ed5b0612/packages/block-editor/src/components/list-view/block-select-button.js#L161-L161) as it's from the core store and can't be imported into the block editor package.

Open to suggestions!

andrewserong:
Thanks for the quick work here! I know this is a WIP so you've probably already seen, but just in case, it looks like the `Locked` state is appearing in the other dropdowns (I think due to the early return in the `useSelect` which means that `canEditTemplateParts` will be falsy):

image

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
```diff
@@ -19,53 +20,81 @@ import { store as editorStore } from '../../store';
...
 						'postType',
-						'wp_template',
-						templateId
+						'wp_block',
+						getBlockAttributes( patternParent ).ref
 					);
+				} else {
+					const { getCurrentTemplateId } = select( editorStore );
+					const templateId = getCurrentTemplateId();
+					if ( templateId && _canEditTemplateParts ) {
```
ramonjd: Probably don't need `&& _canEditTemplateParts` here given the `if ( ! canEditTemplateParts ) {` below.

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
```diff
@@ -19,53 +20,81 @@ import { store as editorStore } from '../../store';
...
+				};
+			},
+			[ clientId ]
+		);
+
+	if ( ! canEditTemplateParts ) {
+		return (
+			<>
+				<BlockSettingsMenuFirstItem>
+					<MenuItem
```
ramonjd: So this was hacked together and needs review.

There are two items — MenuItem and Text — because there is a line separating the two that persists.

Ideally, the dropdown wouldn't be required at all and we'd replace the ellipsis menu with a lock icon.

Ran out of time to investigate that today.

ramonjd (commented):
Pinging @kevin940726 to check the change to the `templateId` condition.

Thanks!

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
```diff
@@ -19,53 +20,84 @@ import { store as editorStore } from '../../store';
...
+			( select ) => {
+				const {
+					getBlockEditingMode,
+					getBlockParentsByBlockName,
+					getSettings,
+					getBlockAttributes,
+				} = select( blockEditorStore );
+				const contentOnly =
+					getBlockEditingMode( clientId ) === 'contentOnly';
+				const _canEditTemplateParts = select( coreStore ).canUser(
```
talldan:
```suggestion
				const _canEditTemplates = select( coreStore ).canUser(
```

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
```diff
@@ -19,53 +20,84 @@ import { store as editorStore } from '../../store';
...
 				);
-			} else {
-				const { getCurrentPostType, getCurrentTemplateId } =
-					select( editorStore );
-				const currentPostType = getCurrentPostType();
-				const templateId = getCurrentTemplateId();
-				if ( currentPostType === 'page' && templateId ) {
+				if ( ! contentOnly ) {
+					return {
+						canEditTemplateParts: _canEditTemplateParts,
```
talldan: I don't think this is needed in the early return (see the next comment).

ramonjd:
> Thanks for the quick work here! I know this is a WIP so you've probably already seen, but just in case, it looks like the Locked state is appearing in the other dropdowns (I think due to the early return in the useSelect which means that canEditTemplateParts will be falsy):

Yep, thanks @andrewserong 

I saw that and was struggling with how to best satisfy all the conditions, so I just committed what I had locally in [4dd5b63](https://github.com/WordPress/gutenberg/pull/62354/commits/4dd5b63678d5f0726076f2b19da81aeb4d385415) and am hoping that the rest comes out in testing. 😄 

I'll carry on with this tomorrow.

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
```diff
@@ -19,53 +20,84 @@ import { store as editorStore } from '../../store';
...
+					entity: record,
+					onNavigateToEntityRecord:
+						getSettings().onNavigateToEntityRecord,
+				};
+			},
+			[ clientId ]
+		);
+
+	// The post has a parent template part, but the user does not have the capability to edit template parts.
+	if ( ! canEditTemplateParts && entity ) {
```
talldan: I think you could move this `if` statement to after the one that follows (`if ( ! entity )`) to achieve the same thing.

ramonjd: This needs better testing, I'm worried about unknown-to-me scenarios.

It should only show where the post has a template and the user can't edit said template.

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
ramonjd: Ah yeah, good call!

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
ramonjd: Yasss!

packages/editor/src/components/block-settings-menu/content-only-settings-menu.js
talldan: I don't think the lock icon is desired, as users can still edit the content of blocks like Post Title, even with low permissions. So this is on the right track, but the copy needs to be adjusted.

I think some text like "Only users with permissions to edit the template can move or delete this block" is more in the spirit of https://github.com/WordPress/gutenberg/pull/61127. Perhaps with an aria-disabled 'Edit Template' button.

@jameskoster may have some suggestions.

TLDR for you - the question is around what to show in list view here when the user doesn't have permissions to edit the template:
Screenshot 2024-06-06 at 3 03 25 pm

ramonjd: > I don't think the lock icon is desired, as users can still edit the content of blocks like Post Title, even with low permissions. So this is on the right track, but the copy needs to be adjusted.
> 
> I think some text like "Only users with permissions to edit the template can move or delete this block" is more in the spirit of https://github.com/WordPress/gutenberg/pull/61127.

I like this. I've updated the PR so that we're using the existing component, but disabling the button and adding the suggested text for now, until further design guidance 😄

kevin940726:
On trunk, it shows `Unlock` for `contentOnly` blocks:

image

But in this PR it shows `Edit template`, which is wrong.

image

(By contentOnly blocks, I mean the one inside `templateLock: contentOnly`, like for the below markup:)
```html
<!-- wp:group {"templateLock":"contentOnly","layout":{"type":"constrained"}} -->
<p>content locked</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```

I think this is why I added the `page` check. I guess neither one works 😅. I remember it bringing me headaches when I first implemented it 😅. I wonder if there's a better condition we can use here 🤔.

ramonjd:
> I wonder if there's a better condition we can use here

I wonder that too!  

If you have any ideas... otherwise I can take a brief look today, but very happy if you wanna push a commit here too 😄

ramonjd:
> On trunk, it shows Unlock for contentOnly blocks:

Except when it's a page 😄 

Screenshot 2024-06-07 at 3 38 47 PM

I guess there should be some condition that also takes into account the post editor.

This makes me think the `getCurrentPostType()` check should be completely removed, as this PR does.

But I'm not sure what the right condition would entail. `getRenderingMode()` doesn't really help as there is a mixture of template-level blocks and page/post contentOnly blocks. 🤔 

Maybe we need to resort to checking the block name?

https://github.com/WordPress/gutenberg/blob/4886740311dc30004385b0cfa999f52232c782a5/packages/editor/src/components/provider/disable-non-page-content-blocks.js#L9-L14

ramonjd:
> Maybe we need to resort to checking the block name?

I've done that in c50b7ba to test it out

ellatrix:

ramonjd:
> Should this be backported for WP 6.6? If so, please add the Backport to WP Beta/RC label.

👍🏻 Not sure if we've found the right solution yet, but I'll add it optimistically. Thanks!

talldan:
> I think this is why I added the page check. I guess neither one works 😅. I remember it bringing me headaches when I first implemented it 😅. I wonder if there's a better condition we can use here 🤔.

I haven't looked too deeply at the problem, but there might be some inspiration in the block inspector and the way it uses the `getContentLockingParent` selector - https://github.com/WordPress/gutenberg/blob/b0c9f4be156c505eb66f96b425543381d0d3078f/packages/block-editor/src/components/block-inspector/index.js#L97

ramonjd:
> but there might be some inspiration in the block inspector and the way it uses the getContentLockingParent selector 

Oh terrific, thanks @talldan I'll run a test on it.

ramonjd:
>  there might be some inspiration in the block inspector and the way it uses the getContentLockingParent selecto

This appears to work as expected, and actually corresponds with the [selector](https://github.com/WordPress/gutenberg/blob/77a02a4251a45170780e28e026f5c4a88673fc1b/packages/editor/src/components/block-settings-menu/content-only-settings-menu.js#L133-L133) in the TemplateLockContentOnlyMenuItems component. 👍🏻  

The selector checks the ancestor blocks to see if they're a pattern `core/block` or `contentOnly` locked.

https://github.com/WordPress/gutenberg/blob/9218b6ea3a62536c3083e6a7a7f7574e8f374ffb/packages/block-editor/src/store/private-selectors.js#L474-L486


The only thing I'm not certain about is what should happen after a user has unlocked the block. The message says:

`"Temporarily unlock the parent block to edit, delete or make further changes to this block."`

Is it supposed to be temporary? I'm testing in the post editor, and once I unlock it, it stays that way 😄 

<details>

<summary>Example locked content</summary>

```html
<!-- wp:group {"templateLock":"contentOnly","layout":{"type":"constrained"}} -->
<div class="wp-block-group"><!-- wp:image {"id":27,"sizeSlug":"full","linkDestination":"none"} -->
<figure class="wp-block-image size-full"><img src="http://localhost:8888/wp-content/uploads/2024/06/white-bellied-swallow-rotated.jpg" alt="" class="wp-image-27"/></figure>
<!-- /wp:image -->

<!-- wp:cover {"overlayColor":"accent","isUserOverlayColor":true,"isDark":false,"layout":{"type":"constrained"}} -->
<div class="wp-block-cover is-light"><span aria-hidden="true" class="wp-block-cover__background has-accent-background-color has-background-dim-100 has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:paragraph {"align":"center","placeholder":"Write title…","fontSize":"large"} -->
<p class="has-text-align-center has-large-font-size">content locked cover</p>
<!-- /wp:paragraph --></div></div>
<!-- /wp:cover -->

<!-- wp:paragraph -->
<p>content locked</p>
<!-- /wp:paragraph -->

<!-- wp:group {"layout":{"type":"constrained"}} -->
<p>content locked nested</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->

```

</details>

kevin940726:
> Is it supposed to be temporary? I'm testing in the post editor, and once I unlock it, it stays that way 😄

It seems to be working as expected on my end. It should select the parent block and show the "Done" button on the toolbar. It's not very noticeable, though. There's a separate issue https://github.com/WordPress/gutenberg/issues/61554 tracking that. Is this what you mean here?

ramonjd:
> It seems to be working as expected on my end. It should select the parent block and show the "Done" button on the toolbar. It's not very noticeable, though. There's a separate issue https://github.com/WordPress/gutenberg/issues/61554 tracking that. Is this what you mean here?


Thanks for checking @kevin940726  - yep that's exactly what I meant. So I just didn't see the "Done" button 🙈 

Ignore me then.

kevin940726 (approved):
Seems to work as expected! Thanks! ❤️

ellatrix:

Issue #62347, opened 2024-06-05
Pattern Overrides: Initial number of number set  (or $ + 2 numbers) disappear from edited Paragraph & Heading Blocks


EVERGIB:
### Description

While using Synced Pattern Overrides on a pattern to display prices on a Paragraph Block for a restaurant menu, I noticed that the "$" and next two numbers of the price would not display on the website front end after editing from the original. Additionally, if I used only numbers for the price and no "$", I noticed that the first number of the set of numbers does not display on the front end (looks normal in the editor). The same issue happens with the "Headings" block when part of a Pattern Override.

### Step-by-step reproduction instructions

1. Create Group Block
2. Add Heading Block and Paragraph Block to this group
3. Create Synced Pattern from this group and Enable Overrides for the Headings and Paragraph blocks
4. Add an instance of the synced pattern to your page
5. edit either the heading block or paragraph block with three or more numbers, or a $ followed by three or more numbers 
6. view the page on the front-end - note that the initial number or $+2 numbers has disappeared

### Screenshots, screen recording, code snippet

Images show the Pattern Override after editing the price - and the result on the website front end.
<img width="1223" alt="Pattern-Override_Backend-Editing" src="https://github.com/WordPress/gutenberg/assets/20259087/4672c6fe-0990-4de7-8d0f-1f3fd8f93f27">
<img width="845" alt="Pattern-Override_FrontEnd-Result" src="https://github.com/WordPress/gutenberg/assets/20259087/fcc994db-6cbc-44dd-8708-589111af958a">

### Environment
- WordPress: 6.6-beta1
- PHP: 8.1.23
- Server: nginx/1.16.0
- Database: mysqli (Server: 8.0.16 / Client: mysqlnd 8.1.23)
- Browser: Chrome 125.0.0.0 (macOS)
- Theme: Twenty Twenty-Four 1.1
- MU-Plugins: None activated
- Plugins:
  * WordPress Beta Tester 3.5.5

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

Mamaduka:
cc @talldan, @kevin940726

talldan:
Thanks for the ping @Mamaduka, we just discovered it here too - https://github.com/WordPress/gutenberg/pull/62220#issuecomment-2151322298

SantosGuillamot:
[...]

PR #62355, opened 2024-06-06
Revert changes to bindings replacement logic to not use regex

SantosGuillamot:
## What?
Revert the changes made in [this pull request](https://github.com/WordPress/gutenberg/pull/61236/) to avoid using regex that can cause potential bugs like [this one](https://github.com/WordPress/gutenberg/issues/62347#top).

## Why?
The idea of that pull request was to simplify the logic to support more attributes, but, in the end, there is no plan to support more attributes for 6.6. As it is causing other problems, it seems it is better to revert it.

## How?
Just reverting the changes made.

## Testing Instructions
Follow the testing instructions of https://github.com/WordPress/gutenberg/issues/62347#top and check that the issue is solved.

### DIFF
--- a/backport-changelog/6.6/6744.md
+++ b/backport-changelog/6.6/6744.md
@@ -1,0 +1,3 @@
+https://github.com/WordPress/wordpress-develop/pull/6744
+
+* https://github.com/WordPress/gutenberg/pull/62355
--- a/lib/compat/wordpress-6.5/blocks.php
+++ b/lib/compat/wordpress-6.5/blocks.php
@@ -65,27 +65,68 @@ function gutenberg_block_bindings_replace_html( $block_content, $block_name, str
 	switch ( $block_type->attributes[ $attribute_name ]['source'] ) {
 		case 'html':
 		case 'rich-text':
-			// Hardcode the selectors and processing until the HTML API is able to read CSS selectors and replace inner HTML.
-			// TODO: Use the HTML API instead.
-			if ( 'core/paragraph' === $block_name && 'content' === $attribute_name ) {
-				$selector = 'p';
-			}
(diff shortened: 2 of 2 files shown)


kevin940726 (approved):
Thanks for the quick PR!

talldan:
> but that didn't make it to 6.6 and it is causing other problems, so it seems it is better to revert it.

I see it did make it and there's a core PR https://github.com/WordPress/wordpress-develop/pull/6744.

Testing with both that and this PR checked, the revert works, but with only this one the block bindings code from core still causes the issue. I guess block bindings must be running twice when the gutenberg plugin is active.

talldan (approved):
Works for me too 🎉

SantosGuillamot:
> I see it did make it and there's a core PR https://github.com/WordPress/wordpress-develop/pull/6744.

Oh sorry, I meant that the support of more attributes didn't make it in 6.6. I'll change the description for clarity.

> Testing with both that and this PR checked, the revert works, but with only this one the block bindings code from core still causes the issue.

Do you mean that with WordPress `trunk` and this branch you still see the issue? I will double-check, but I think it wasn't the case for me.

talldan:
> Do you mean that with WordPress trunk and this branch you still see the issue? I will double-check, but I think it wasn't the case for me.

Yep, when debugging I could see that `WP_Block->replace_html` is being called on the content before `gutenberg_block_bindings_replace_html`.

SantosGuillamot:
You're totally right. Anyway, with the backport in core I believe we should be fine.

Apart from that, I think we don't want to run `gutenberg_block_bindings_replace_html` if core is doing it. It should just act for older versions of WordPress. But I can work on that in another pull request.

PR #58991, opened 2024-02-14
Enqueue block custom CSS only when block renders on the page.

tellthemachines:
## What?

Follow up from https://github.com/WordPress/wordpress-develop/pull/5892.

Global styles per-block custom CSS should respect the `wp_should_load_separate_core_block_assets` flag and only be enqueued when the block renders on the page if separate assets are enabled.

Ideally this work would be done core-first, but given we're currently in Beta it will have to wait a while, so it's worth trying out here in the meantime.

## Testing Instructions

1. Test with a block theme that adds custom CSS to a few of its blocks, such as TT4, or add custom CSS to blocks in global styles. Also add some base custom CSS (in the global styles sidebar, open the "More" dropdown to show Additional CSS if it's not visible in the UI).
2. In the front end, load up a page that contains blocks with custom CSS and one that doesn't. Check that custom CSS is loaded only when blocks are on the page (custom CSS styles can be found inside `<style id='global-styles-inline-css'>`)
3. Add `add_filter( 'should_load_separate_core_block_assets', '__return_false', 11 );` to the theme's `functions.php` and check that custom CSS for all blocks always loads in the front end.

PR #47396, opened 2023-01-25
Move custom css to its own inline style generation method and combine with customizer CSS

glendaviesnz:
## What?
Moves global styles custom CSS to its own methods so its position in the output can be modified independently to the other global styles

## Why?
This allows the global styles custom CSS to be loaded after any Customizer custom CSS that may have been added in hybrid themes, otherwise [this CSS has higher specificity](https://github.com/WordPress/gutenberg/issues/30142#issuecomment-1372950974).

## How?
Extract the loading of custom CSS into its own method, separate to the other global styles, and include any customizer CSS first in this output.

## Testing Instructions

- Add some custom CSS in global styles, including some block level custom CSS
- Load in front end and make sure styles appear at the bottom of the `global-styles-inline-css` style element
- In a hybrid theme like BlocksBase also add some custom CSS in Customizer and make sure this CSS loads before the Global Style custom CSS in the `global-styles-inline-css` style element

## Screenshots or screencast 

Before:
Screenshot 2023-02-01 at 10 34 49 AM

After:

Screenshot 2023-02-01 at 10 32 45 AM

PR #62357, opened 2024-06-06
Update custom CSS handling to be consistent with block global styles.

tellthemachines:
## What?

Addresses [several](https://github.com/WordPress/gutenberg/pull/58991#discussion_r1507923422) [concerns](https://github.com/WordPress/gutenberg/pull/58991#discussion_r1507927224) on the PR that added conditional loading of block custom CSS. 

* Custom CSS for blocks should be loaded only when the block is present on the page when `wp_should_load_separate_core_block_assets` is true;
* Global styles custom CSS should be handled by the same logic as other global styles;
* Any custom CSS coming from the Customizer should be loaded before the global styles custom CSS, as has been the case since #47396.
* Functions previously used for custom CSS handling and no longer needed have been deprecated.
   * Deprecation notice for functions in core is tagged with WP 6.7
   * For Gutenberg-only functions it's tagged with next plugin version.
* Because base and block-specific custom CSS are now generated in different places, I replaced the test for the now-deprecated `get_custom_css` function with two different tests.

## Testing Instructions

1. Test with custom CSS added at base level and at block level, in theme.json or through the global styles UI
2. Check that custom CSS for blocks is loaded only when the block is present on the page when `wp_should_load_separate_core_block_assets` is true (which it is by default in most block themes);
3. Add `add_filter( 'should_load_separate_core_block_assets', '__return_false', 11 );` to the theme's functions.php and check that custom CSS for all blocks always loads in the front end.

### DIFF
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -1409,6 +1409,7 @@ class WP_Theme_JSON_Gutenberg {
 	 * @return string The global styles custom CSS.
 	 */
 	public function get_custom_css() {
+		_deprecated_function( __METHOD__, '6.7.0', 'get_stylesheet' );
 		$block_custom_css = '';
 		$block_nodes      = $this->get_block_custom_css_nodes();
 		foreach ( $block_nodes as $node ) {
@@ -1427,6 +1428,7 @@ class WP_Theme_JSON_Gutenberg {
 	 * @return string The global styles base custom CSS.
 	 */
 	public function get_base_custom_css() {
+		_deprecated_function( __METHOD__, 'Gutenberg 18.6.0', 'get_stylesheet' );
 		return isset( $this->theme_json['styles']['css'] ) ? $this->theme_json['styles']['css'] : '';
 	}
 
@@ -1438,6 +1440,7 @@ class WP_Theme_JSON_Gutenberg {
 	 * @return array The block nodes.
 	 */
 	public function get_block_custom_css_nodes() {
+		_deprecated_function( __METHOD__, 'Gutenberg 18.6.0', 'get_block_nodes' );
 		$block_nodes = array();
 
 		// Add the global styles block CSS.
@@ -1470,6 +1473,7 @@ class WP_Theme_JSON_Gutenberg {
 	 * @return string The global styles custom CSS for the block.
 	 */
 	public function get_block_custom_css( $css, $selector ) {
+		_deprecated_function( __METHOD__, 'Gutenberg 18.6.0', 'get_styles_for_block' );
 		return $this->process_blocks_custom_css( $css, $selector );
 	}
 
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -1355,6 +1355,12 @@ class WP_Theme_JSON_Gutenberg {
 			$stylesheet .= $this->get_preset_classes( $setting_nodes, $origins );
 		}
 
+		// Load the custom CSS last so it has the highest specificity.
+		if ( in_array( 'custom-css', $types, true ) ) {
+			// Add the global styles root CSS.
+			$stylesheet .= _wp_array_get( $this->theme_json, array( 'styles', 'css' ) );
+		}
+
 		return $stylesheet;
 	}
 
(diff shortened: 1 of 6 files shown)


lib/class-wp-theme-json-gutenberg.php
```diff
@@ -2855,6 +2866,11 @@ static function ( $pseudo_selector ) use ( $selector ) {
 			$block_rules .= static::to_ruleset( ":root :where($style_variation_selector)", $individual_style_variation_declarations );
 		}
 
+		// 7. Generate and append any custom CSS rules.
+		if ( isset( $node['css'] ) ) {
+			$block_rules .= $this->process_blocks_custom_css( $node['css'], $selector );
```
ramonjd, line 2869: I think this line allows adding custom CSS declarations to the root selector in theme.json

```json
	"styles": {
		"css": "background-color:purple;"
	},
```

I can see the background color on the frontend but not in the site editor

Whereas in trunk only this works:

```json
	"styles": {
		"css": "body{background-color:purple;}"
	},
```

Should Gutenberg allow both? Or is it a regression?

ramonjd:
Other than the question [above](https://github.com/WordPress/gutenberg/pull/62357#discussion_r1634126541), this PR tests well for me.

I tested in site editor and via theme.json 👍🏻

lib/class-wp-theme-json-gutenberg.php
tellthemachines: Hmm that's a good question. This is how it was before #47396, but given that PR is over a year old it might be best to keep things as they were more recently. I'd expect to have to write complete rules in the top level custom CSS control and not necessarily for loose properties to be attached to the body element.

tellthemachines: Oooooh this is interesting; I think we've detected a bug in the existing behaviour. Testing in trunk, the loose properties are output outside any rule, so they don't apply because they're invalid, both in editor and front end:

Screenshot 2024-06-11 at 3 43 26 PM

Screenshot 2024-06-11 at 3 47 01 PM

Whereas in this branch, those properties are nested under the root selector in the front end, though they're still invalid output in the editor. I think ideally we should fix things so they aren't output at all. I'll look into it!

tellthemachines: Ok I fixed things so that this branch behaves like trunk (no valid CSS output if only properties are added to the top level custom CSS control) but the loose properties are still being output (just like they are on trunk).

The problem with this is that any rule output straight after a loose property won't work. I can repro this in core trunk, and it's also a problem with the customizer additional CSS (which may not manifest in classic themes because additional CSS is enqueued after everything else, but on hybrid themes we're adding it before global styles custom CSS, so it can also break stuff)

We could maybe just check that there aren't any loose properties at the end of the custom CSS string? Checking the whole string is valid CSS might be overkill as site admins should be responsible for the validity of their CSS. We could also Do Nothing 😄 and leave things as they are.

If we do want to do anything though, it's probably better as a separate PR.

ramonjd: > If we do want to do anything though, it's probably better as a separate PR.

For sure. Thanks for looking into it. 

> We could maybe just check that there aren't any loose properties at the end of the custom CSS string? 

Or maybe some light-weight parsing where the styles are added?

https://github.com/WordPress/gutenberg/blob/01a0e429483e85245112219bb66f12a65063014b/lib/class-wp-theme-json-gutenberg.php#L1361-L1361

ramonjd (approved):
Tested again. LGTM

Styles loaded only when the block is present on the page 👍🏻 

Otherwise behaves as it does on trunk.

Thanks!

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -1421,6 +1428,7 @@ public function get_custom_css() {
 	 * @return string The global styles base custom CSS.
 	 */
 	public function get_base_custom_css() {
```
ramonjd, line 1430: So these never made it to Core, right? Should we remove the `@since` annotations and maybe add a "don't backport note"?

Doesn't have to be here, just wondering if it'll help later.

tellthemachines: Oh good point, we should remove the `@since` on these!

PR #62360, opened 2024-06-06
Fix increasingly big canvas in the post editor when editing patterns

talldan:
## What?
There seems to have been a bug introduced lately to editing patterns in the post editor. The editor canvas has lots of empty space at the bottom that grows increasingly bigger as more content is added.

This PR fixes it.

## How?
I tracked it down to a `padding-bottom: 40vh` that's added in the `useEditorStyles` hook. It shouldn't be added for patterns, but it looks like the React hook is missing the `postType` dependency.

Then a second issue is that some code needs to be added to remove the styles when already added.

Thirdly, I added an extra check to make sure multiple padding rules aren't being added.

## Testing Instructions
1. Insert a synced pattern in the post editor
2. Click 'Edit original' on the toolbar
3. Confirm the canvas should be sized to the content
4. Try adding more blocks to the pattern and confirm the size stays consistent

### Before
Screenshot 2024-06-06 at 5 16 56 pm

### After
Screenshot 2024-06-06 at 5 16 27 pm

### DIFF
--- a/packages/edit-post/src/components/layout/index.js
+++ b/packages/edit-post/src/components/layout/index.js
@@ -137,6 +137,7 @@ function useEditorStyles() {
 		editorSettings.disableLayoutStyles,
 		editorSettings.styles,
 		hasThemeStyleSupport,
+		postType,
 	] );
 }
 
--- a/packages/edit-post/src/components/layout/index.js
+++ b/packages/edit-post/src/components/layout/index.js
@@ -59,6 +59,10 @@ const DESIGN_POST_TYPES = [
 	'wp_navigation',
 ];
 
+const TYPEWRITER_STYLE = {
+	css: 'body{padding-bottom: 40vh}',
+};
+
 function useEditorStyles() {
 	const {
 		hasThemeStyleSupport,
@@ -114,11 +118,11 @@ function useEditorStyles() {
 			} );
 		}
 
-		const baseStyles = hasThemeStyles
+		let baseStyles = hasThemeStyles
 			? editorSettings.styles ?? []
 			: defaultEditorStyles;
 
-		// Add a constant padding for the typewritter effect. When typing at the
+		// Add a constant padding for the typewriter effect. When typing at the
 		// bottom, there needs to be room to scroll up.
 		if (
 			! isZoomedOutView &&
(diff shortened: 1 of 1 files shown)


packages/edit-post/src/components/layout/index.js
```diff
@@ -114,21 +118,24 @@ function useEditorStyles() {
...
+			! baseStyles.includes( TYPEWRITER_STYLE )
 		) {
-			baseStyles.push( {
-				css: 'body{padding-bottom: 40vh}',
-			} );
+			baseStyles.push( TYPEWRITER_STYLE );
+		} else if ( baseStyles.includes( TYPEWRITER_STYLE ) ) {
+			baseStyles = baseStyles.filter(
+				( style ) => style !== TYPEWRITER_STYLE
+			);
```
youknowriad: So first, I think this fix is fine and we can ship it as is if you like. 

But I have a couple thoughts:

Removing the style shouldn't be needed. I think the issue is that we're providing the value of the setting by reading the exact same settings. I think this hook shouldn't be using the "editorSettings" from the editor package. Instead it should be using the "settings" prop of the Layout component.  

The second thing is that I think that this code should probably move within the editor package somewhere. Because I believe this padding should probably be applied to the editor every time "posts/pages" are rendered in post-only mode regardless of whether it's the post editor or site editor or some other page.

---
Both of these comments can be addressed separately though.

youknowriad (approved):

packages/edit-post/src/components/layout/index.js
talldan: Thanks for the speedy review.

I can follow up on that, though it'll be next week as it's end of day for me now and I'll only be back working on Monday.

ellatrix: Right, why does the `TYPEWRITER_STYLE` need to be removed if this is the only place where it's adding it to the `baseStyles`?

talldan: It looks like it's due to `editorSettings.styles` being mutated, so it should be an easy fix to prevent that. I've pushed some commits to the PR.

talldan: I've gone ahead and merged the PR, but happy to revisit in a follow-up if there's any further feedback.

ellatrix:

PR #62366, opened 2024-06-06
Move the template part menu items to the editor package

youknowriad:
Related #52632 

## What?

In the site editor, when editing templates, you can create template parts and detach template parts from the block settings menu. This PR brings that code to the editor package, that way the same capabilities are also available when editing templates in the post editor.

## Testing Instructions

You can go the post editor, click "edit template" from the sidebar, and then you can select a template part like "header" and you'll see a "Detach" link.


packages/editor/src/components/provider/index.js
```diff
@@ -301,6 +302,7 @@ export const ExperimentalEditorProvider = withRegistryProvider(
 							{ ! settings.__unstableIsPreviewMode && (
 								<>
 									<PatternsMenuItems />
+									<TemplatePartMenuItems />
```
ntsekouras, line 305: Nit: In trunk we would render this above `PatternsMenuItems`

ntsekouras (approved):
LGTM, thanks!

packages/editor/src/components/provider/index.js
youknowriad: Oh right! Was this intentional? I feel "patterns" are more common than template parts, so maybe this order is better?

PR #62372, opened 2024-06-06
Fix: Omit default parameters from pages, template parts, and patterns

### LINKED ISSUE #60709: DataViews: Routing related issues.
### Description

Sometimes some url arguments linger when navigating between data views related pages, potentially causing bugs in some pages.

## Bug 1

 1- Go to pages: The url should be `path=%2Fpage`
 2- Click "drafts": The url becomes something like `path=%2Fpage&activeView=drafts&isCustom=false` 
 
 The first issue here is that I think "isCustom=false" shouldn't be present in the url because "false" is the default value. If an optional argument equals its default value, it should ideally be dropped from the url (nicer urls)

 3- Go back to "all pages" The url becomes something like `path=%2Fpage&activeView=all&isCustom=false`

Notice that 1 and 3 refer to the same page with two different URLs, this is not great, 3 should just use the same url as 1 (omitting default values from the url).

## Bug 2

This one is more important than the previous bug:

 1- Go to pages: The url should be `path=%2Fpage`
 2- Click "drafts": The url becomes something like `path=%2Fpage&activeView=drafts&isCustom=false` 
 3- Click the "back" button on the sidebar to go back to the root level of the site editor: The url becomes `activeView=drafts&isCustom=false`

Note the lingering activeView and isCustom arguments that should ideally be removed at this stage (we should end up with no argument in the url (it's the root of the site editor)

 4- Click "templates", you'll end up in a broken state in the templates page (empty list) because of the lingering arguments in the url.

## Bug 3

- Go to Patterns.
- Select the "All Template part" category and then one of the Template Parts.
- Go to the editor.
- Go back to the patterns page.
- Note how the section selected is the "All patterns" though it should be "All template parts".

_No response_

### Environment info

_No response_

Yes

Yes

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60709

Fixes the first bug @youknowriad identified.

## Testing Instructions
1- Go to pages: The url should be `wp-admin/site-editor.php?postType=page`
2- Click "drafts": The url becomes something like `/wp-admin/site-editor.php?postType=page&activeView=drafts`.
3- Go back to "all pages" Verify the url is still `wp-admin/site-editor.php?postType=page`.

Repeat these steps for template parts and patterns verify that when switching views and going back to all view the URL is still the same

### DIFF
--- a/packages/edit-site/src/components/sidebar-dataviews/dataview-item.js
+++ b/packages/edit-site/src/components/sidebar-dataviews/dataview-item.js
@@ -35,11 +35,15 @@ export default function DataViewItem( {
 	const iconToUse =
 		icon || VIEW_LAYOUTS.find( ( v ) => v.type === type ).icon;
 
+	let activeView = isCustom ? customViewId : slug;
+	if ( activeView === 'all' ) {
+		activeView = undefined;
+	}
 	const linkInfo = useLink( {
 		postType,
 		layout,
-		activeView: isCustom ? customViewId : slug,
-		isCustom: isCustom ? 'true' : 'false',
+		activeView,
+		isCustom: isCustom ? 'true' : undefined,
 	} );
 	return (
 		<HStack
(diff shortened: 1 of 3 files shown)


youknowriad (approved):
Works great. Thanks for the fix.

ellatrix:

PR #32457, opened 2021-06-04
Introduce Block type level lock control

senadir:
Parts of https://github.com/WordPress/gutenberg/issues/29864

## Description
This PR introduces parts of #29864 mainly, it adds:
- Attribute level support for locking a block from being moved/removed.
- Hides the block moving arrows and delete button if a block is locked.

The PR also adds checks at the datastore level for both moving and removing blocks.

## API definition
When defining a block, the value is set at the attributes level.

```
attributes: {
	lock: {
		type: 'object',
		default: {
			move: true,
			remove: true,	
		},
	},
},
```
This would lock the block ability to be moved or removed.

## How has this been tested?
- Modify any block attributes to include moving or removing.
- modify any innerBlocks templateLock to be `insert`, `all`, or `false` to see the all the possible combinations.
- Try to move or remove the block, it should respect your choices.

This table shows the intersect of conditions between `templateLock` and `lock`

For removing:
| lock.remove\templateLock | `"all"`      | `"insert"`   | `false`      |
|--------------------------|--------------|--------------|--------------|
| `true`                   | can't Remove | can't Remove | can't Remove |
| `false`                  | can Remove   | can Remove   | can Remove   |
| `undefined`              | can't Remove | can't Remove | can Remove   |

For moving:
| lock.move\templateLock | `"all"`    | `"insert"` | `false`    |
|------------------------|------------|------------|------------|
| `true`                 | can't Move | can't Move | can't Move |
| `false`                | can Move   | can Move   | can Move   |
| `undefined`            | can't Move | can Move   | can Move   |

## Screenshots 

In this example. the parent block is locked to `all`, the first block "Contact information" has locking disabled for move and remove, which means it can be moved or removed.
https://user-images.githubusercontent.com/6165348/120816789-fc87af80-c548-11eb-821e-227083a9badf.mov

## Use case
There are several valid uses cases outlined in #29864
Another use case that we're building for is having a Checkout Block with different blocks that act as fundamental steps, we don't want people to delete or move those steps since they're fundamental and their order is also important, but we want to allow people to select them, access settings, and insert blocks in between them.

PR #62374, opened 2024-06-06
Patterns: Fix the 'canRemove' condition for the 'Detach' action

### LINKED ISSUE #62368: Detaching patterns not available when editing pages in the Site Editor
### Description

Similar to https://github.com/WordPress/gutenberg/issues/62367 (but this occurs regardless of nested block level). The "Detach" control is missing when selecting a synced pattern while editing a page in the Site Editor. 

1. Open the Site Editor.
2. Select "Pages".
3. Select "Add new page".
3. Open the Inserter. 
4. Insert a synced pattern (denoted by a purple icon). 
5. Open the block toolbar options menu.
6. See missing "Detach" control. 

https://github.com/WordPress/gutenberg/assets/1813435/f41a8c32-1806-4b27-842d-d298ec58e290

### Environment info

WordPress trunk
Gutenberg trunk

Yes

Yes

Mamaduka:
## What?
Fixes #62368.

PR fixes the `canRemoveBlock` selector call in the `PatternsManageButton` component so that it correctly displays the "Detach" menu item.

## Why?
The components should check if removal action is available relative to the root block. The global root (`''`) is locked for pages, but the content is editable.

## Testing Instructions
1. Open the Site Editor.
2. Select "Pages".
3. Select "Add new page".
4. Open the Inserter.
5. Insert a synced pattern (denoted by a purple icon).
6. Open the block toolbar options menu.
7. See the "Detach" menu item.

Same.

### DIFF
--- a/packages/patterns/src/components/patterns-manage-button.js
+++ b/packages/patterns/src/components/patterns-manage-button.js
@@ -18,13 +18,14 @@ import { unlock } from '../lock-unlock';
 function PatternsManageButton( { clientId } ) {
 	const { canRemove, isVisible, managePatternsUrl } = useSelect(
 		( select ) => {
-			const { getBlock, canRemoveBlock, getBlockCount } =
+			const { getBlock, canRemoveBlock, getBlockRootClientId } =
 				select( blockEditorStore );
 			const { canUser } = select( coreStore );
 			const reusableBlock = getBlock( clientId );
+			const rootClientId = getBlockRootClientId( clientId );
 
 			return {
-				canRemove: canRemoveBlock( clientId ),
+				canRemove: canRemoveBlock( clientId, rootClientId ),
 				isVisible:
 					!! reusableBlock &&
 					isReusableBlock( reusableBlock ) &&
@@ -33,7 +34,6 @@ function PatternsManageButton( { clientId } ) {
 						'blocks',
 						reusableBlock.attributes.ref
 					),
-				innerBlockCount: getBlockCount( clientId ),
 				// The site editor and templates both check whether the user
 				// has edit_theme_options capabilities. We can leverage that here
 				// and omit the manage patterns link if the user can't access it.


ellatrix:
Could we add an e2e test for this?

Mamaduka:
> Could we add an e2e test for this?

It's probably not worth the effort; this shouldn't regress. The same bug was in `ReusableBlocksManageButton`, which got copied over.

The check is correctly done in `BlockActions`.

https://github.com/WordPress/gutenberg/blob/524eca4ca1219942e54bd2cdb37178769da19a8b/packages/block-editor/src/components/block-actions/index.js#L50

t-hamano (approved):
LGTM!

I found a similar issue with the Classic block, so I submitted #62430.

packages/patterns/src/components/patterns-manage-button.js
```diff
@@ -18,13 +18,14 @@ import { unlock } from '../lock-unlock';
...
-			const { getBlock, canRemoveBlock, getBlockCount } =
+			const { getBlock, canRemoveBlock, getBlockRootClientId } =
 				select( blockEditorStore );
 			const { canUser } = select( coreStore );
 			const reusableBlock = getBlock( clientId );
+			const rootClientId = getBlockRootClientId( clientId );
 
 			return {
-				canRemove: canRemoveBlock( clientId ),
+				canRemove: canRemoveBlock( clientId, rootClientId ),
```
ellatrix, line 28: Why does `canRemoveBlock` require a `rootClientID`, while it can derive that itself from the `clientId`? Can't we just fix the function and fix all cases at once?

Commented here too https://github.com/WordPress/gutenberg/pull/62430/files#r1632567240

ntsekouras: That's a very interesting observation Ella. `rootClientId` seems to be needed for every case.

Mamaduka: I can't recall; it's probably part of the original API design.

I can follow up on the selector change, but I'm unsure if similar enhancements can be included in release betas/RCs.

ellatrix: @Mamaduka It's a bug fix: what happens when you call it with `rootClientId` that is not actually the root? It doesn't make sense to keep it, it eliminates bug across the board for users that  call `canRemoveBlock` without it or wrongly by accident. So it's fine to include.

t-hamano: This API was added in [#32457](https://github.com/WordPress/gutenberg/pull/32457/files#diff-aa92c48095c35f17feb834dd919baf52274a73c7b1180ff5acce7b6087b126cfR1314). However, even reading the discussion, there is no mention of why `rootClientID` is required as a parameter, or whether there are cases where `rootClientID` is not required.

Many other selectors accept a `rootClientID` as an option, so this API may have been designed to be consistent with them.

PR #60284, opened 2024-03-28
Data views: Add ability to display fields as a badge in grid layout

jameskoster:
## What?
This PR does a few things which can be boiled down to three main changes:

1. Allow data view consumers to specify that certain fields should appear as a label-less 'badge' in grid layout.
2. Display 'badge' fields together above all others.
3. Update the Patterns page so that the 'sync status' field displays as a badge, and style accordingly.

## Why?
Follow-up to https://github.com/WordPress/gutenberg/pull/60083, which introduced the ability to arrange fields in a 'column' layout and forced the display of field labels.

This change is for fields where the label is not necessary, IE it can be interpreted from the value alone. 

Of the current field set across all data views, 'sync status' in patterns is an appropriate consumer: There is no need for the verbose: "Sync status: Synced" when "Synced" alone will do.

As data views expand to other areas (e.g. Media Library: https://github.com/WordPress/gutenberg/issues/55238) this affordance will become more useful. Media is a good example because media files have many fields that do not require a label (size, file name, type, dimensions, etc.). This PR will make it possible to implement the following mockup:

Export

## How?
1. Adds `displayAsBadgeFields` parameter to `GridItem`, similar to `displayAsColumnFields`
2. Detect and display all 'badge fields' in a new `HStack` _above_ all other fields.
3. Include `field.id` as a classname on all field outputs.
4. Add basic styling for badges.
5. Update `defaultConfigPerViewType` in the Patterns page to specific that `sync-status` is a badge field.
6. Update `sync-status` field render to include class name based on value
7. Apply purple accents to synced pattern badges

Two other minor tweaks:

1. Sync status field name: "Sync Status" -> "Sync status".
2. In pages data view, "Date" field name is now more explicit: "Publish date".

## Testing Instructions
1. Open the patterns page.
2. Toggle on the 'Sync status' field.
3. Observe the new display.
4. Check that all other data views and layouts work as before.

### Before

Screenshot 2024-03-28 at 12 53 00

### After

Screenshot 2024-03-28 at 12 53 10

For more advanced testing you can temporarily swap fields in the Pages data view to display as badges by opening `packages/edit-site/src/components/sidebar-dataviews/default-views.js` and updating `DEFAULT_CONFIG_PER_VIEW_TYPE` like so:

```
export const DEFAULT_CONFIG_PER_VIEW_TYPE = {
	[ LAYOUT_TABLE ]: {
		primaryField: 'title',
	},
	[ LAYOUT_GRID ]: {
[...]

PR #55935, opened 2023-11-07
Add context for translators to any unclear usage of "synced"

talldan:
## What?
Fixes #55783

As mentioned in that issue, it's unclear if the word "synced" is singular or plural and this can result in incorrect translations.

## How?
Add a translator comment explaining each usage of the word 'synced'.

PR #62375, opened 2024-06-06
i18n: Patterns: Disambiguate singular & plural uses of 'Synced' & 'Unsynced'

mcsf:
## What?
Add or amend i18n context to labels "Synced" and "Unsynced" to disambiguate strings where the adjective qualifies a singular noun ("a [synced] pattern") from those qualifying a plural noun ("the [synced] patterns").

## Why?
English spoils developers by being so lightly inflected. :)

In many languages, adjectives have case inflections for number. In my usual locale (pt-PT):
* o padrão **sincronizado** (_the synced pattern_)
* os padrões **sincronizados** (_the synced patterns_)

Without clear context attached to short strings like `'Synced'`, translators have no way to know which case to use, nor to encode different translations (one for singular and one for plural). For instance, in my locale, I spotted the sync status badge for patterns "mis-numbered":

sync-state-number-mismatch

## How?
Consistently provide i18n context to the strings "Synced" and "Unsynced" whenever they appear in UI to refer to patterns' sync status. This context (via `_x()`) should read:
* `pattern (singular)` when the string is used in the context of a single pattern (e.g. a a status badge in a Data Views grid listing; a sync option when creating a new pattern)
* `patterns` (preexisting context) when the string is used in the context of a group of patterns (e.g. a select control to filter between "All" / "Synced" / "Unsynced")

Previously, strings either had no context or had one that would lead translators to assume the wrong number.

## Context

- #55935: Fix addressing a similar report during the 6.4 cycle, which saw the first addition of i18n context
- #60284: Introduction of badge fields in Data Views

### DIFF
--- a/packages/reusable-blocks/src/components/reusable-blocks-menu-items/index.js
+++ b/packages/reusable-blocks/src/components/reusable-blocks-menu-items/index.js
@@ -9,6 +9,10 @@ import { BlockSettingsMenuControls } from '@wordpress/block-editor';
 import ReusableBlockConvertButton from './reusable-block-convert-button';
 import ReusableBlocksManageButton from './reusable-blocks-manage-button';
 
+/*
+ * FIXME: CAN WE GET RID OF THIS COMPONENT? AS FAR AS I CAN SEE, IT'S EXPORTED
+ * BUT NEVER USED, ONLY MENTIONED IN DOCUMENTATION.
+ */
 export default function ReusableBlocksMenuItems( { rootClientId } ) {
 	return (
 		<BlockSettingsMenuControls>
--- a/packages/reusable-blocks/src/components/reusable-blocks-menu-items/index.js
+++ b/packages/reusable-blocks/src/components/reusable-blocks-menu-items/index.js
@@ -9,10 +9,6 @@ import { BlockSettingsMenuControls } from '@wordpress/block-editor';
 import ReusableBlockConvertButton from './reusable-block-convert-button';
 import ReusableBlocksManageButton from './reusable-blocks-manage-button';
 
-/*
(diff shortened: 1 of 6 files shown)


packages/reusable-blocks/src/components/reusable-blocks-menu-items/index.js
```diff
@@ -9,6 +9,10 @@ import { BlockSettingsMenuControls } from '@wordpress/block-editor';
 import ReusableBlockConvertButton from './reusable-block-convert-button';
 import ReusableBlocksManageButton from './reusable-blocks-manage-button';
 
+/*
+ * FIXME: CAN WE GET RID OF THIS COMPONENT? AS FAR AS I CAN SEE, IT'S EXPORTED
+ * BUT NEVER USED, ONLY MENTIONED IN DOCUMENTATION.
+ */
 export default function ReusableBlocksMenuItems( { rootClientId } ) {
```
youknowriad: It seems like the whole "reusable-blocks" package can be deprecated. Might be good to do that in its own PR.

mcsf: > It seems like the whole "reusable-blocks" package can be deprecated. Might be good to do that in its own PR.

Yeah, I added this as a separate commit to easily revert — I just didn't want to forget

mcsf:
Also cc @oandregal for visibility, since you worked a lot on the filters:

One thing I considered for a moment was to generalise this problem at the field level. Specifically, we could require that the possible states of an enumeration include a singular and a plural label. But I've dismissed it for now: in practice, until we have more use cases, it seems like defaulting to singular is enough. WDYT?

ellatrix:
Should this be backported? If so, please add the `Backport to WP Beta/RC` label. Thanks!

youknowriad (approved):
LGTM 👍

ellatrix:

PR #62378, opened 2024-06-06
Fix: DataViews: Active page is not highlighted properly in list view.

### LINKED ISSUE #60708: DataViews: Active page is not highlighted properly in list view.
### Description

When you're navigating the the page lists (list view).
you can select pages and they become highlighted,
you can click the "preview" to enter the editor and you can go back to the previous page (two steps through the details page)
But when you're back to the list view, the right page is actually rendered in the preview but it's not highlighted properly in the list.

It seems like we need a "controlled" mode for the selection in the DataViews component (passing the selection as a prop)

A quick way to trigger this is to open a url like the following:  http://localhost:8888/wp-admin/site-editor.php?postId=2&path=%2Fpage

Screenshot 2024-04-12 at 2 08 36 PM

Sample page is active on the preview but not on the list view.

### Environment info

_No response_

Yes

Yes

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60708.

This PR adds a "controlled" mode for the selection in the DataViews component (passing the selection as a prop) as suggested by @youknowriad.
Adds logic to make the current link in the URL the selection. For this to work we needed some logic to avoid race conditions e.g.: we should only include the selection if the items are loaded otherwise dataviews code automatically removes the selection.
We also had some cases where the selection id were integers instead of strings and we missed a usage getItemId.

## Testing Instructions
Go to http://localhost:7888/site-wp-dev/wp-admin/site-editor.php?postType=page&postId=1246 (1246 needs to be a valid page id)
Verify the active page is highlighted properly.

### DIFF
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -38,10 +38,12 @@ interface DataViewsProps< Item extends AnyItem > {
 		totalPages: number;
 	};
 	supportedLayouts: string[];
+	selection?: string[];
+	setSelection?: ( selection: string[] ) => void;
 	onSelectionChange?: ( items: Item[] ) => void;
 }
 
-const defaultGetItemId = ( item: AnyItem ) => item.id;
+const defaultGetItemId = ( item: AnyItem ) => item.id + '';
 const defaultOnSelectionChange = () => {};
 
 function useSomeItemHasAPossibleBulkAction< Item extends AnyItem >(
@@ -72,9 +74,22 @@ export default function DataViews< Item extends AnyItem >( {
 	isLoading = false,
 	paginationInfo,
 	supportedLayouts,
+	selection: selectionProperty,
+	setSelection: setSelectionProperty,
 	onSelectionChange = defaultOnSelectionChange,
 }: DataViewsProps< Item > ) {
-	const [ selection, setSelection ] = useState< string[] >( [] );
+	const [ selectionState, setSelectionState ] = useState< string[] >( [] );
+	let selection, setSelection;
+	if (
+		selectionProperty !== undefined &&
+		setSelectionProperty !== undefined
+	) {
+		selection = selectionProperty;
+		setSelection = setSelectionProperty;
+	} else {
+		selection = selectionState;
+		setSelection = setSelectionState;
+	}
 	const [ openedFilter, setOpenedFilter ] = useState< string | null >( null );
 
 	useEffect( () => {
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -43,7 +43,7 @@ interface DataViewsProps< Item extends AnyItem > {
 	onSelectionChange?: ( items: Item[] ) => void;
 }
 
-const defaultGetItemId = ( item: AnyItem ) => item.id + '';
+const defaultGetItemId = ( item: AnyItem ) => item.id;
 const defaultOnSelectionChange = () => {};
 
 function useSomeItemHasAPossibleBulkAction< Item extends AnyItem >(
--- a/packages/edit-site/src/components/page-pages/index.js
+++ b/packages/edit-site/src/components/page-pages/index.js
@@ -206,6 +206,18 @@ export default function PagePages() {
 	const [ view, setView ] = useView( postType );
 	const history = useHistory();
 
+	const [ selection, setSelection ] = useState( [] );
+
+	const {
+		params: { postId },
+	} = useLocation();
+	const [ postIdToSelect, setPostIdToSelect ] = useState();
+	useEffect( () => {
+		if ( postId ) {
+			setPostIdToSelect( postId );
+		}
+	}, [ postId ] );
+
 	const onSelectionChange = useCallback(
 		( items ) => {
 			const { params } = history.getLocationWithParams();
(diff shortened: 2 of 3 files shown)


ellatrix:
Should this be backported? If so, please add the `Backport to WP Beta/RC` label. Thanks!

packages/edit-site/src/components/page-pages/index.js
```diff
@@ -206,6 +206,18 @@ export default function PagePages() {
 	const [ view, setView ] = useView( postType );
 	const history = useHistory();
 
+	const [ selection, setSelection ] = useState( [] );
+
+	const {
+		params: { postId },
+	} = useLocation();
+	const [ postIdToSelect, setPostIdToSelect ] = useState();
```
youknowriad: Can we use postId as default value here to avoid initial re-rendering?

packages/dataviews/src/dataviews.tsx
```diff
@@ -38,10 +38,12 @@ interface DataViewsProps< Item extends AnyItem > {
 		totalPages: number;
 	};
 	supportedLayouts: string[];
+	selection?: string[];
+	setSelection?: ( selection: string[] ) => void;
 	onSelectionChange?: ( items: Item[] ) => void;
 }
 
-const defaultGetItemId = ( item: AnyItem ) => item.id;
+const defaultGetItemId = ( item: AnyItem ) => item.id + '';
```
youknowriad: Is this needed?

jorgefilipecosta: Yes it is needed, it is making sure the ID returned is a string, in some cases, it was an integer that made the comparisons fail. The selections have a type of string[] and previously many times it was an integer array.

packages/edit-site/src/components/page-pages/index.js
jorgefilipecosta: Yes, good catch the code was updated 👍

packages/dataviews/src/dataviews.tsx
youknowriad: our types, say it's always a string. So if we were actually using typescript in the consuming side we'd see the error. So maybe instead of changing the default getting we could provide `( item: AnyItem ) => item.id.toString()` as a prop for the case where the id is integer no?

packages/edit-site/src/components/page-pages/index.js
```diff
@@ -266,6 +278,20 @@ export default function PagePages() {
 		totalPages,
 	} = useEntityRecords( 'postType', postType, queryArgs );
 
+	useEffect( () => {
```
youknowriad: Isn't this a duplicate of the effect above? Also the code feels a bit fragile to me to be honest. Do we expect all dataviews to have this same logic. It's not really straightforward. Is there a way to improve this?

jorgefilipecosta: It is not the same effect as above. Above, when the page gets a postId in the URL parameter, we set that ID as something to be part of the selection. We cannot select it immediately because the items may still be loading.
In this effect, if things are loaded and an item should be selected, we update the selection.
It looks fragile but seems to work well. Previously, I tried other ideas, but they ended up not working all the time or being even more complex. I am not sure how we can simplify this further. What I can do is put the effects and the state inside a hook that we can reuse in other places if needed (and that way, at least the logic is joined in a specific place).

packages/dataviews/src/dataviews.tsx
jorgefilipecosta: Hi @youknowriad I applied your suggestion.

packages/edit-site/src/components/page-pages/index.js
jorgefilipecosta: I made this change to extract the logic in a specific hook.

carolinan:
_Not a code review_

To test this issue, make sure that you are not testing the background color change which happens when you manually click on a page in the list: You need to test the **currently** active page, so you need to refresh the browser.

**Before:** The list has no visible indicator that a page is selected.
**After:** There is a visible background color change for the active page in the list:

The page list view (data view), with the active page highlighted by a background color.



While there is a light visible background color change, I am unable to detect any indicator that a page is the active page when I am using a screen reader. Is there an open issue for this problem, or are there specific testing instructions for screen readers?

The background color alone is not enough to indicate that the page is the active page, the color has too low contrast compared to the default color.

carolinan:
I did a quick test and I am seeing the same issue for the templates (missing indicator of the active template when the list layout is used), and this is not solved by this PR. Is there a separate issue for that?

oandregal (approved):
The templates screen also uses the list view and faces the same issue — it'd be nice to have a follow-up for that as well.

jorgefilipecosta:
Thank you for the reviews @carolinan and @oandregal.

I will follow up with something similar for the templates.

> While there is a light visible background color change, I am unable to detect any indicator that a page is the active page when I am using a screen reader. Is there an open issue for this problem, or are there specific testing instructions for screen readers?

We are showing aria-pressed="true" do you think we should include other additional aria information? If so I'm happy to iterate on this.

carolinan:
> > While there is a light visible background color change, I am unable to detect any indicator that a page is the active page when I am using a screen reader. Is there an open issue for this problem, or are there specific testing instructions for screen readers?
> 
> We are showing aria-pressed="true" do you think we should include other additional aria information? If so I'm happy to iterate on this.

The element that includes the information about the page in the list layout is not a button and does not have `role="button"`, where is `aria-pressed="true"` used? I do not see it in the markup.
The element that has the `is-selected` class is an `<li>` with the `role="row"`, and the background color is set on a div that is inside that.

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: e3097e22c9

Issue #49501, opened 2023-03-31
Site Logo: Dark transparent site logos are invisible


jasmussen:
When you upload an image to a Site Logo block the first time, a square crop of that image is used also as your Site Icon. This is likely going to be somewhat common for white sites with black text logo types, such as this:

<img width="486" alt="Screenshot 2023-03-31 at 09 24 17" src="https://user-images.githubusercontent.com/1204802/229053220-3b5fe001-a844-4fc9-b8d9-6bfa62f655bc.png">

In that situation, the site icon becomes barely visible:

<img width="846" alt="Screenshot 2023-03-31 at 09 26 20" src="https://user-images.githubusercontent.com/1204802/229053332-f6ace5de-bc6d-4ca4-86a9-0be50b505a85.png">

Let's ensure that the site logo always has the site background applied, so that the color the logo has been designed to rely on is applied for contrast:
<img width="406" alt="Screenshot 2023-05-02 at 08 42 01" src="https://user-images.githubusercontent.com/1204802/235597494-f2d53a31-d2a7-4e58-a705-f1e55f400006.png">

Can just apply the background color to the `img` directly.

courtneyr-dev:
Possibly related? https://wordpress.org/support/topic/gutenberg-dosent-work/

jasmussen:
Seems likely, yes. One idea, add a circular dot behind any image, so there's always going to be something visible at least.

richtabor:
>  It would be easy to add a white or light gray background, and that should do it, right?

What if we used the site's background color as the fallback? Probably would be better with the header template part's background even—though I'm not sure how possible that would be.

jasmussen:
> What if we used the site's background color as the fallback? Probably would be better with the header template part's background even—though I'm not sure how possible that would be.

That seems like it would be a great solution.

mtias:
+1 for using the background of the site, in the same way the global styles card is created.

talldan:
Something that might be considered in conjunction with this is the contrast of the favicon when a browser is in dark mode. This was my test of using a logo like the one in the issue description (1st and 3rd tabs are my WordPress site, and yes I do tend to have too many tabs open):
![Screen Shot 2023-05-03 at 4 06 33 pm](https://user-images.githubusercontent.com/677833/235862633-026d3de5-13e4-4fad-9b46-0740be281208.png)

I wonder if there are any good options for helping users avoid that, like a preview of an example browser tab in the block inspector.

[...]

PR #62386, opened 2024-06-06
Edit site: Restore animations on site hub and canvas

stokesman:
## What
Restores animations and another detail or two that were lost in #61579.

## Why
The previous transition was a good bit nicer. The current will lend to an impression that the app is shoddy.

## More 
The idea with #61579 was to simplify. It certainly had a nice ratio of deletions to additions but while some concerns may have been simplified others were complicated. Moreover, its deletions must not have been enabled just because the hub is no longer persistent and fixed above everything. I say that because this PR has net zero additions while restoring the previous transition.

The main caveat with this is that Distraction Free mode has a visual difference. The hub persists there too. However, the mode works. To me, this makes a better trade-off and it can be completely fixed too. I don’t want this PR getting any more unwieldy for review so I’ll put such changes in another.

## How?

- Puts Site icon button in front of the editor header
- Removes back button in the header
- Restores hover and transition on Site icon button (with CSS instead of framer-motion)
- Removes view transitions

The rest of these are more independent and could probably each be extracted as their own PRs in case any of them prove debatable.
- 338a600bb8e05e550d14781276474dfb1303372c
  - Restores border-radius on site icon (when it’s an image)
  - Tries improving the focus ring of the site icon
- 21910a4b41953f12d715f47dcf59a813772e7053 Respects reduced motion of canvas border-radius transition and tweaks the duration otherwise.
- b9654ef3861cd88eb8ab365997d36100254e887b Fixes a disparate bit of header styling in Distraction Free mode.
- e1d6734bbd006ac3234719460e3e161ddacd86fa Fixes a flash of background color and unintended initial transition of the site icon.

## Testing Instructions
1. Open the site editor and observe animations while switching between view/edit.

### On this branch (0.2x speed)

https://github.com/WordPress/gutenberg/assets/9000376/f522971e-d95d-4855-82fa-a12582e67512

### On trunk (0.2x speed)

https://github.com/WordPress/gutenberg/assets/9000376/198ba7eb-e19d-40d5-bee5-ef594ed0a00a

### DIFF
--- a/packages/edit-site/src/components/block-editor/style.scss
+++ b/packages/edit-site/src/components/block-editor/style.scss
@@ -1,15 +1,4 @@
-/* stylelint-disable -- Disable reason: View Transitions not supported properly by stylelint. */
-::view-transition-old(frame),
-::view-transition-new(frame) {
-  animation-duration: 0;
-}
-/* stylelint-enable */
-
 .edit-site-visual-editor__editor-canvas {
-	/* stylelint-disable -- Disable reason: View Transitions not supported properly by stylelint. */
-	view-transition-name: frame;
-	/* stylelint-enable */
-
 	&.is-focused {
 		outline: calc(2 * var(--wp-admin-border-width-focus)) solid var(--wp-admin-theme-color);
 		outline-offset: calc(-2 * var(--wp-admin-border-width-focus));
--- a/packages/edit-site/src/components/editor/index.js
+++ b/packages/edit-site/src/components/editor/index.js
@@ -7,7 +7,6 @@ import clsx from 'clsx';
  * WordPress dependencies
  */
 import { useDispatch, useSelect } from '@wordpress/data';
-import { Button } from '@wordpress/components';
 import { useInstanceId } from '@wordpress/compose';
 import {
 	EditorKeyboardShortcutsRegister,
@@ -41,11 +40,10 @@ import {
 } from '../editor-canvas-container';
 import SaveButton from '../save-button';
 import SiteEditorMoreMenu from '../more-menu';
-import SiteIcon from '../site-icon';
 import useEditorIframeProps from '../block-editor/use-editor-iframe-props';
 import useEditorTitle from './use-editor-title';
 
-const { Editor, BackButton } = unlock( editorPrivateApis );
+const { Editor } = unlock( editorPrivateApis );
 const { useHistory } = unlock( routerPrivateApis );
 const { BlockKeyboardShortcuts } = unlock( blockLibraryPrivateApis );
 
@@ -122,7 +120,6 @@ export default function EditSiteEditor( { isLoading } ) {
 		],
 		[ settings.styles, canvasMode, currentPostIsTrashed ]
 	);
-	const { setCanvasMode } = unlock( useDispatch( editSiteStore ) );
 	const { createSuccessNotice } = useDispatch( noticesStore );
 	const history = useHistory();
 	const onActionPerformed = useCallback(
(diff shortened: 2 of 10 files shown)


youknowriad:
For me as it stands, this PR is a step in the wrong direction. I understand that it restores a better animation but it compromises something that I feel is fundamental when we start thinking about the site editor shell.

This introduces a strong dependency between the site hub that is rendered in the site editor shell and the frame. It assumes that there's an "editor" that is present in the site editor.

While it's the case now, there might be pages where there's no block editor. Like the "media library" for instance (which is going to reuse the site editor layout/shell). 

So IMO, there are two paths forward:

 - Keep things as in trunk and seek ways to improve the animation, I think we can improve it to be honest, the only issue I saw is the "W" logo padding (I think with some time, we can solve that). Did you see other issues in the animation?

 - Make the "site hub" a route area (like header, sidebar,...). I don't like that personally because most of the time the site hub is the same.

---

I noticed that the site hub is now always visible in distraction free mode in this PR.

stokesman:
> This introduces a strong dependency between the site hub that is rendered in the site editor shell and the frame. It assumes that there's an "editor" that is present in the site editor.

I think to be more precise it’s a strong dependency on the canvas mode. If there’s not an "editor" present, there’s not an "edit" mode and the Site Hub won’t change to its correspondent "mode". There are various other bits of the layout component that also depend on the canvas mode. Pardon me if I miss your point. I do wish to understand.

> Did you see other issues in the animation?

The biggest thing for me is the tearing of the top of the frame when going back to view mode. Here’s a frame capture from the video in the description:
image
There is also ghosting of elements within the frame and its border-radius snaps (though the latter should be an easy fix). 

> Keep things as in trunk and seek ways to improve the animation

I've worked at a few approaches to improving the animation in ways that preserve the separate site hub and back button. It’s certainly improvable but none are quite this good. I may have one more idea to try on that. Anyway, I'm not convinced the separate buttons have any real advantage. They still act as if they’re a single button and are more complicated than just a single "moded" button. Not that I feel the single one is perfect. Seems like maybe there is a design change that could help resolve this.

> I noticed that the site hub is now always visible in distraction free mode in this PR.

Thanks for testing. That’s right and I mentioned that in the description. The mode works though and that can also be resolved I just didn’t want this PR having more changes.

jasmussen:
Thanks for sweating the details!

Between trunk and this PR, I think both have room for improvement. Technical discussions aside—though I think those are important to address—there's something about the animation orchestration that's still not quite right.

First off it'll help to think of the default site logo, the WordPress logo, as being the same as an actual site logo. In the latter case, we need to solve #49501 so site logos that are transparent have a background.

Once that's done, we can think of both the W and the site logo, as simply icons on a billboard. For the WordPress logo, the billboard is a dark color, similar to the default WordPress background. For a site icon, it's a site-background-colored billboard. In both cases, they stack like this:

stacked billboards

I find when orchestrating animation it often helps to think of each element as a physical card on a table. When it animates, it has to come from somewhere, have a stacking order, physical properties, etc. That places the site icon as the top level item, above both site and edit views. And the billboard behavior will help explain what happens when we move into the editor. For me that puts the ideal animation somewhere between trunk and this PR.

youknowriad:
> The biggest thing for me is the tearing of the top of the frame when going back to view mode. Here’s a frame capture from the video in the description:

Interesting (looking at the screenshot). That's not supposed to happen even in trunk and I'm not sure that's related to the site hub at all. I'm on WCEU these days so I might not have the focus time to dive deeper her in the next coming days, but let's try to work on these details, and find a path forward for these animations... I'll probably have better availabilities next week.

stokesman:
> Thanks for sweating the details!

😃 I guess details are relative. I didn’t consider the broken bits of the animation in trunk as details. I guess they are but less so than some timing/orchestration concerns. I feel it’s a bit of a euphemism to have labeled this an "enhancement".

I’d agree there is room for improvement. I may be able to deduce what that might mean for you having looked at the timing differences with trunk and this branch. Mainly I see the site icon waits before scaling up until the canvas frame has gotten into fullscreen placement. I suspect that came about due to a technical rather than design concern in #61579 but if it’s considered better this branch can certainly follow suit.

----

> Interesting (looking at the screenshot). That's not supposed to happen even in trunk…

Well for a bit I’d begun to think I had some funny version of Chrome or something but I've since tested other versions and computers. Now it can easily be tested anywhere cause it’s in the [WP nightly](https://playground.wordpress.net/?php=8.0&wp=nightly&storage=none).

jasmussen:
> I guess details are relative. I didn’t consider the broken bits of the animation in trunk as details. 

Just to clarify, I consider details to be the individual pieces that make up the whole. The whole is just a collection of details, and if any one of them is off, the whole thing feels off. So I mean it as the sincerest form of a compliment, that you care about the details.

Issue #62371, opened 2024-06-06
DataViews for Posts, Categories and Tags


youknowriad:
Similarly to the pages, patterns and templates pages in the site editor, we should explore how we can leverage the same DataViews UIs, frame, editor, navigation... to power the Posts section of WP-Admin as part of the Admin redesign project.

There are a lot of challenges here:

 - Reusing the "pages" dataviews without duplicating all the code.
 - Reusing the site editor shell, navigation menus and routing logic
 - Generating forms for categories and tags creation and modification without necessarily creating custom code #59745 
 - Extensibility #61084 
 - Backward compatibility
 - Transition 

There's a lot to cover and a lot of uncertainties, for these reasons, I'm proposing that we start a new experimental page in the Gutenberg plugin, where we explore a new page that could serve as a playground to explore these ideas and address all the blockers. 

## Task list:

 - [x] Bootstrap a site-editor like shell #62409 
 - [x] Add the "posts" dataviews page. (@ntsekouras)
 - [ ] Add the "categories" dataviews page.
 - [ ] Add the "tags" dataviews page.
 - [ ] Leverage DataViews forms to implement "Add new category" and "Add new tag" (and edit forms as well)

fabiankaegy:
Thank you for creating this and starting this work :) 

I'm really excited about this one! As you mention there are many challenges / open questions. The main things for me are all focussed around what the rollout of the "Site Editor Frame" will mean to a lot of the other areas of the WordPress Admin. 

Will this experimental posts screen display the old traditional WP left sidebar? Will it feature the admin bar? Admin notifications etc? Or will it replace the entire screen with the site editor like experience? 

So essentially which of these three approaches are we going to take with this? 

| Table Only | Content Area Only | Entire Screen |
|--------|--------|--------|
| ![CleanShot 2024-06-06 at 19 36 48@2x](https://github.com/WordPress/gutenberg/assets/20684594/def56396-99d1-43c6-95f1-efd4f9247fae) | ![CleanShot 2024-06-06 at 19 36 48@2x copy](https://github.com/WordPress/gutenberg/assets/20684594/2c4688ab-a913-4c96-9b2a-9ec4ce5c4063) | ![CleanShot 2024-06-06 at 19 36 48@2x copy 2](https://github.com/WordPress/gutenberg/assets/20684594/b77c58e0-96eb-4141-a461-0ce719ba66c4) |

youknowriad:
[...]

PR #62389, opened 2024-06-06
Site Editor: Remove editor specific classes from shell wrapper.

youknowriad:
Related to #62371 

## What?

As we start exploring building a site editor like experience for posts and media libraries... we need to ensure the "Layout" or "Shell" of the site editor is not opinionated about its content/pages... 

With that in mind, this PR removes some class names from that wrapper (most of these classnames are applied appropriately to the editor wrapper or elsewhere) which I found to be useless.

## Testing Instructions

1- Check the site editor.
2- Check the navigation block overlay in the site editor.

### DIFF
--- a/packages/block-library/src/navigation/editor.scss
+++ b/packages/block-library/src/navigation/editor.scss
@@ -457,12 +457,6 @@ $color-control-label-height: 20px;
 	}
 }
 
-.has-fixed-toolbar .wp-block-navigation__responsive-container.is-menu-open {
-	@include break-medium() {
-		top: $admin-bar-height + $header-height + $block-toolbar-height + $border-width;
-	}
-}
-
 .is-mobile-preview .wp-block-navigation__responsive-container.is-menu-open,
 .is-tablet-preview .wp-block-navigation__responsive-container.is-menu-open {
 	top: $admin-bar-height + $header-height + $block-toolbar-height + $border-width;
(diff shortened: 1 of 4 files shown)


packages/block-library/src/navigation/editor.scss
```diff
@@ -457,12 +457,6 @@ $color-control-label-height: 20px;
 	}
 }
 
-.has-fixed-toolbar .wp-block-navigation__responsive-container.is-menu-open {
```
youknowriad, line 460: I tested the navigation block, it's the only place where this has-fixed-toolbar was used but I didn't find that these styles had any meaningful impact.

jasmussen: It's been a while, but I believe that these styles are probably related to when the overlay menu is open. It may be fine to remove if this still tests well, regardless there should be an incoming overlay-to-template-part PR that would most likely make it redundant.

youknowriad: Yes, I tried this flow, and didn't notice any difference with or without the style.

ntsekouras (approved):
LGTM as I couldn't spot any differences from trunk. As a follow up we should probably consider removing some of these classes like `is-distraction-free` and conditionally render the components that are affected by such logic. Currently we do some conditional rendering and some css hiding..

PR #62392, opened 2024-06-07
Update get-started-with-create-block.md

pietrodeferrari:
Added a little note for those who can't figure out why the block isn't showing inside the Editor with a link to the issue and temporary fix.

I myself got stuck here for a while and found some comments around the web asking for a solution.

## What?
Showing a note with a solution

## Why?
Some users get stuck and can't find the block they just created using @wordpress/scripts@28

## How?
The temporary solution is showing a note that explains the problem

## Testing Instructions
1. Use npx @wordpress/create-block@latest todo-list inside plugins folder
2. A folder for a block called "todo-list" is created
3. Activate the block from the Plugins section inside WP dashboard
4. Go inside the Editor
5. Search the block
6. Block isn't showing

### DIFF
--- a/docs/getting-started/devenv/get-started-with-create-block.md
+++ b/docs/getting-started/devenv/get-started-with-create-block.md
@@ -25,6 +25,10 @@ The `slug` provided (`todo-list`) defines the folder name for the scaffolded plu
 
 Navigate to the Plugins page of our local WordPress installation and activate the "Todo List" plugin. The example block will then be available in the Editor. 
 
+<div class="callout callout-info">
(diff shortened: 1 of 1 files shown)


gziolo:
@ryanwelcher or @ndiego, do you think this addition would still be valuable? The context probably would need to be better outlined so it's clear that for WordPress 6.6 and higher everything should work as intended with every version of WP Scripts.

ndiego:
> @ryanwelcher or @ndiego, do you think this addition would still be valuable? The context probably would need to be better outlined so it's clear that for WordPress 6.6 and higher everything should work as intended with every version of WP Scripts.

Given that 6.7 was released a month ago, I personally don't think this message is needed.

gziolo:
@pietrodeferrari, thank you for opening this pull request. It should no longer be a prominent issue as discussed above.

Issue #58133, opened 2024-01-23
Unify and fix accessibility of the various Pagination components


afercia:
### Description

The editor uses a few user interfaces where pagination buttons are in place. For example:

- dataviews
- patterns explorer 
- global styles revisions
- possibly other places I may have missed 

While theoretically these pagination are very similar and should work in a very similar way, they are all different and use different components.

From an accessibility perspective, there are a few issues. Most notably:

- Buttons get dynamically disabled while they are focused, thus triggering a focus loss. See https://github.com/WordPress/gutenberg/issues/56547
- The buttons don't show a tooltip so that their accessible name isn't visually exposed.

I'd tend to think these pagination tools should _work and look_ the same way throughout the whole user interface.
It's also a good opportunity to fix the most relevant accessibility issuses and abstract some code.

Screenshots:

Dataviews pagination:

![01 pagination dataviews](https://github.com/WordPress/gutenberg/assets/1682452/45b1738e-2a5b-4c83-a2be-eb8dfc1f993e)

Patterns explorer pagination:

![02 pagination patterns explorer](https://github.com/WordPress/gutenberg/assets/1682452/d3932eeb-5bc0-4630-a83a-1673ffe72eda)

Global styles revisions pagination:

![03 revisions pagination](https://github.com/WordPress/gutenberg/assets/1682452/f9797f35-ec71-41d5-8a2d-5ebbf2cb2455)

NEW: Pages pagination: note: as far as I can tell this is the only case where the buttons do not trigger a focus loss because they use `arira-disabled` instead of `disabled`. Tooltips work as well. The implementation of the current page indicator is arguable though as it's a mix of translatable strings and labeling / value of the `<select>` element.

![new pages pagination](https://github.com/WordPress/gutenberg/assets/1682452/574de3cb-ca8e-4cb3-bec7-a320e1967deb)

### Step-by-step reproduction instructions

- Test one of these paginations for example the global styles one.
- Go to Site editor > Styles > Revisions.
- Use the keyboard and Tab to the 'Last page' button.
- Observe the button doesn't show a tooltip.
- Press Enter.
- Observe the 'Last page' button is now not focusable because it has a `disabled` HTML attribute.
- Observe focus is lost: press Tab again and observe the tab order starts from scratch from the document root.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

mikachan:
[...]

PR #62395, opened 2024-06-07
Global styles: display tooltips for pagination buttons on styles revision

### LINKED ISSUE #62283: Global Styles: Pagination buttons must have a tooltip
### Description

Any icon button must visually expose the button accessible name via a tooltip. Worth mentioning tooltips are already a compromise for accessibility, as visible text is always preferred.

The Global styles pagination buttons are an interesting case, as the implementation breaks the logic of the Button component in the first place. To me, this is a sign that the component can be misused. Cc @mirka  

The Button component logic is, shortly:
- When the Button has a text children, it doesn't show a tooltip (unless there's a keyboard shortcut that needs to be shown in the tooltip).
- When the Button only shows an icon, the `lebel` prop is used for the tooltip.

Turns out the Global styles pagination buttons contain text children, but [they're unicode characters for the chevrons](https://github.com/WordPress/gutenberg/blob/a9fd33c99b69ef6ea747fed96569ad837632aedb/packages/edit-site/src/components/pagination/index.js#L46-L54): `«, ‹, ›, »`.

As such, even by changing the buttons `aria-label` to `label`, the tooltip is not rendered because the Button component _thinks_ there is visible text.

Unfortunately, the 'visible' text is just glyphs that are visually equivalent to icons. There is no accessible name exposed.

This is clearly a misuse of the component but it could happen in other scenarios as well, then the 'text' isn't meaningful text that exposes the control's accessible name.

In this specific case, a quick fix would be replacing the `«, ‹, ›, »` with actual icons by the means of the `icon` prop.

- Go to Site editor > Styles > Revisions.
- Make sure there's enough revisions to show the pagination buttons ad the end of the revisions list.
- Hoer or focus the pagination buttons.
- Observe no tooltips appear.

_No response_

### Environment info

_No response_

Yes

Yes

ramonjd:
## What?
Maybe resolves https://github.com/WordPress/gutenberg/issues/62283

Adds tooltips to pagination buttons for Global Styles revisions.

## Why?

To expose the button name via tooltip.

## How?

Using the `showTooltip` prop on the Button component.

## Testing Instructions

- Go to Site editor > Styles
- Create at least 11 styles revisions by changes the styles and saving. This ensure the styles revisions will paginate.
- Open styles revisions.
- Hover or focus the pagination buttons.
- Observe that  tooltips appear.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/6458278/8925ec82-b3b7-42ba-a7e2-69f95f7b9367

### DIFF
--- a/packages/edit-site/src/components/global-styles/screen-revisions/style.scss
+++ b/packages/edit-site/src/components/global-styles/screen-revisions/style.scss
@@ -173,19 +173,14 @@
 		// Safari from rendering the page / total text.
 		will-change: opacity;
 	}
+
 	.components-button.is-tertiary {
-		font-size: 28px;
-		font-weight: 200;
 		color: $gray-900;
-		margin-bottom: $grid-unit-05;
-		line-height: 1.2;
 	}
+
 	.components-button.is-tertiary:disabled,
 	.components-button.is-tertiary[aria-disabled="true"] {
 		color: $gray-600;
-	}
-	.components-button.is-tertiary:hover {
-		background: transparent;
 	}
 }
 
--- a/packages/edit-site/src/components/pagination/index.js
+++ b/packages/edit-site/src/components/pagination/index.js
@@ -12,6 +12,7 @@ import {
 	Button,
 } from '@wordpress/components';
 import { __, _x, _n, sprintf } from '@wordpress/i18n';
+import { previous, chevronLeft, chevronRight, next } from '@wordpress/icons';
 
 export default function Pagination( {
 	currentPage,
@@ -49,20 +50,16 @@ export default function Pagination( {
 					__experimentalIsFocusable
 					disabled={ disabled || currentPage === 1 }
 					label={ __( 'First page' ) }
-					showTooltip
-				>
-					«
-				</Button>
+					icon={ previous }
+				/>
 				<Button
 					variant={ buttonVariant }
 					onClick={ () => changePage( currentPage - 1 ) }
 					__experimentalIsFocusable
 					disabled={ disabled || currentPage === 1 }
 					label={ __( 'Previous page' ) }
-					showTooltip
-				>
-					‹
-				</Button>
+					icon={ chevronLeft }
+				/>
 			</HStack>
 			<Text variant="muted">
 				{ sprintf(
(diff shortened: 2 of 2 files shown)


afercia (approved):
@ramonjd thanks for your PR

I pushed a commit to remove the no longer necessary `aria-label` props, as the new `label` props are equivalent. LGTM thanks!

afercia:
Will try to use the existing icons `previous, chevronLeft, chevronRight, next`.

afercia:
In the last two commits I replaced the unicode characters with SVG icons and cleaned up a bit the CSS. Buttons have the same 32 by 32 pixels size. The alignment of the icons within the buttons has slightly improved, as they now use more standard CSS / props. Screenshot before and after:

before and after

t-hamano (commented):
I think this is ideal from an accessibility and code standpoint, but since it's a design change it might be a good idea to get some design feedback.

| Before | After |
|--------|--------|
| image | image | 

If these icons feel a bit large, do we need `previousSmall` and `nextSmall` icons as well as `chevronLeftSmall` and `chevronRightSmall`? cc @WordPress/gutenberg-design 

I think this PR is okay to ship as is, but in the future, we need to apply the same approach and design to other paginations.

### Patterns Dialog

image

### Font Library: Install Fonts

image

t-hamano:
> If these icons feel a bit large, do we need `previousSmall` and `nextSmall` icons as well as `chevronLeftSmall` and `chevronRightSmall`?

Ah, I was wrong, if the icon size is not suitable, we can simply adjust it with the `iconSize` prop.

jasmussen:
At the moment, data views use the regular size chevron, as a 24x24px icon inside a 32x32 button:

Screenshot 2024-06-10 at 08 52 30

Ideally button sizes here are the same. If we run out of space for this particular interface, we can consider removing the first-page/last-page interface, or do something else.

afercia:
> I think this PR is okay to ship as is, but in the future, we need to apply the same approach and design to other paginations.

A while ago I created an issue to unify the various Pagination components, see https://github.com/WordPress/gutenberg/issues/58133. Other improvements should be addressed there and I would love to see more focus on that issue. This is a functional change, as this pagination is fundamentally broken from an accessibility perspective. Thanks everyone for feedback and review.

ramonjd:
Thanks for merging this!

ellatrix:

Issue #39964, opened 2022-04-01
Improve the search for tags when a substring doesn't appear in the first 20 results.


costdev:
## What problem does this address?
See Trac ticket: [55511](https://core.trac.wordpress.org/ticket/55511).

rodrigocarlos:
I want to share an idea that might be the solution. Maybe you could put in the link search window, tabs that would filter by type of post (posts, tags, pages, categories, ...). So by default it can have the **All** tab and the search result will be for all post types (as it is currently). But if we want the result to be just a tag, then just click on the **Tags** tab and only tags related to the research we did will appear. If we want the result to show only categories, then just click on the **Categories** tab and it will only show categories related to the search term.

For example, I want to find all tags that have the word "netflix", so I just type the word "netflix" in the search bar and then click on the **Tags** tab, and the result will show only tags (ex.: /tag/netflix, /tag/netflix-movies, /tag/netflix-series, ...)

Here's an example of what these tabs could look like: https://prnt.sc/jgQUfLYBKek6

Thank you!

Mamaduka:
Still relevant.

sakirsensoy:
Issue was opened exactly 2 years ago but the problem still persists :/

noisysocks:
Closing as a dupe of https://github.com/WordPress/gutenberg/issues/56478 as it has a better description.

A PR https://github.com/WordPress/gutenberg/pull/62397 is open which will fix this.

PR #62397, opened 2024-06-07
Update fetchLinkSuggestions to sort results by relevancy

### LINKED ISSUE #56478: Link suggestions in editor “never” shows taxonomies
### Description

- The link suggestions tool does 3 separate requests (`posts`, `taxonomies`, `media`)
- it then just shows the first x results

**Problem**

When there are too many `posts `containing a result → `taxonomies `(or `media`) are never shown as an option, even if more relevant (i.e. have exact title match)

1. Create many posts with a specific keyword (don't know the threshold, maybe 25)
2. Create a term with teh same keyword

Try to link the term via the search → not possible, as you can't select it

https://github.com/WordPress/gutenberg/assets/504835/a5806101-e847-4f34-bae9-851a4a6dc85f

### Environment info

_No response_

Yes

Yes

Here is an export from a plain wordpress

* Many "duplicated" posts
* Keyword "suggestions" as category and in title of every post -> not possible to link the category
* Keyword "unique" -> only used as category

[basic.WordPress.2023-11-24.zip](https://github.com/WordPress/gutenberg/files/13456671/basic.WordPress.2023-11-24.zip)

noisysocks:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/56478.

Updates `fetchLinkSuggestions`, which is used by `LinkControl`, which is used by the popover that appears when you insert a link, to sort its results by relevancy to the search query that the user typed in.

## Why?

When inserting a link, the user can search for posts, pages, tags, categories, post formats, and media. The search is done via the `/wp/v2/search` REST API endpoint.

Unfortunately WordPress can only search one database table at a time. For example it can't search `wp_posts` and `wp_terms` using one SQL query. The REST API does not hide this limitation and forces you to call`/wp/v2/search` with just one `type` param.

To get around this limitation, `fetchLinkSuggestions` currently makes 4 requests (posts, taxonomies, post formats, media) at once using a `Promise.all`, and then concatenates the results together. There is a max of 20 results per request for a combined max of 80 results. The first 20 results of the combined results are shown to the user.

There's a problem with this approach. Say the user is searching for a category named "Travel Tips" and that there are 20 pages or posts containing the words "travel" or "tips". The "Travel Tips" category will never be shown because it is crowded out by the 20 posts that appear in the combined results before any single category appears.

## How?

My fix for this issue is to sort the combined results before we take the first 20 and show them to the user.

~I'm sorting the results by how similar the title is to the search query that the user provided. This is done using [cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity). We treat the search query as one document and the result title as another document. Then, we build term frequency map vectors for each document. Finally, we calculate the relevancy score by taking the cosine similarity of the two vectors.~

This PR now sorts the results by scoring each result where the score is the number of tokens in the title that are also in the search query, divided by the total number of tokens in the title. This gives us a score between 0 and 1, where 1 is a perfect match. This achieves good enough results but is much simpler to understand than the previous approach described above.

### Alternative approaches

Ideally this would all be handled at the database level using full text search. I don't think we can assume that every WordPress installation's MySQL database has full text search enabled, however.

We could look at doing logic similar to what's in this PR at the server level. I'm not sure if we should, though. The REST API is arguably correct to not encapsulate this limitation. Not all clients will want to order and combine results the same way.

Doing this in the client doesn't prevent us from moving the logic to the server in the future, so it's a good place to start.

~I tried simple keyword matching as an alternative algorithm for ranking results. This is where you award 1 point per keyword that the title and search term have in common. It didn't work as well for me in testing, though, because cosine similarity will give smaller titles that are similar to the query an edge over long titles that contain lots of irrelevant words in addition to the query.~

### House keeping

This PR also contains some house keeping while I'm in this part of the codebase:

- ~Rename `__experimentalFetchLinkSuggestions` to `fetchLinkSuggestions`.~
- Rewrite `fetchLinkSuggestions` in TypeScript. (It was already partially JSDoc typed.)

## Testing Instructions
You really don't notice this bug unless you're testing with real data and have more than 20 posts, tags, etc.

If you don't have any real data, you use WP CLI to create some realistic-enough posts, pages, tags, and categories. Here's some test data that I had ChatGPT spit out:

<details><summary>Test data</summary>
<p>

```
# If you're using wp-env, this alias is helpful.
alias wp='npx wp-env run cli wp'

wp post create --post_type=post --post_title="Exploring the Streets of Paris" --post_status=publish
wp post create --post_type=post --post_title="A Weekend in Rome: What to See and Do" --post_status=publish
wp post create --post_type=post --post_title="Hidden Beaches of Thailand" --post_status=publish
wp post create --post_type=post --post_title="Hiking Trails in the Swiss Alps" --post_status=publish
wp post create --post_type=post --post_title="Cultural Delights of Tokyo" --post_status=publish
wp post create --post_type=post --post_title="Road Trip Through the Australian Outback" --post_status=publish
wp post create --post_type=post --post_title="Discovering Ancient Ruins in Greece" --post_status=publish
wp post create --post_type=post --post_title="Foodie Adventures in Mexico City" --post_status=publish
wp post create --post_type=post --post_title="Safari Experiences in Kenya" --post_status=publish
wp post create --post_type=post --post_title="Island Hopping in the Philippines" --post_status=publish
wp post create --post_type=post --post_title="Exploring the Markets of Marrakech" --post_status=publish
wp post create --post_type=post --post_title="Cityscape Views from New York" --post_status=publish
wp post create --post_type=post --post_title="The Nightlife of Berlin" --post_status=publish
wp post create --post_type=post --post_title="Adventure Sports in New Zealand" --post_status=publish
wp post create --post_type=post --post_title="Luxury Travel in Dubai" --post_status=publish
wp post create --post_type=post --post_title="Historical Sites in Egypt" --post_status=publish
wp post create --post_type=post --post_title="Wine Tasting in Napa Valley" --post_status=publish
wp post create --post_type=post --post_title="Cruising the Caribbean" --post_status=publish
wp post create --post_type=post --post_title="Exploring National Parks in the USA" --post_status=publish
wp post create --post_type=post --post_title="City Guide to Barcelona" --post_status=publish

wp post create --post_type=page --post_title="Top Travel Destinations 2024" --post_status=publish
wp post create --post_type=page --post_title="Ultimate Packing Guide for Travelers" --post_status=publish
wp post create --post_type=page --post_title="How to Plan a Budget Trip" --post_status=publish
wp post create --post_type=page --post_title="Travel Safety Tips for Solo Travelers" --post_status=publish
wp post create --post_type=page --post_title="Family Travel: Best Destinations" --post_status=publish
wp post create --post_type=page --post_title="Romantic Getaways Around the World" --post_status=publish
wp post create --post_type=page --post_title="Top Cultural Festivals to Attend" --post_status=publish
wp post create --post_type=page --post_title="Best Road Trips in the USA" --post_status=publish
wp post create --post_type=page --post_title="Luxury Hotels and Resorts" --post_status=publish
wp post create --post_type=page --post_title="Backpacking Tips for Beginners" --post_status=publish
wp post create --post_type=page --post_title="How to Travel Sustainably" --post_status=publish
wp post create --post_type=page --post_title="Best Cruise Lines for 2024" --post_status=publish
wp post create --post_type=page --post_title="Guide to Adventure Travel" --post_status=publish
wp post create --post_type=page --post_title="Exploring Local Cuisines" --post_status=publish
wp post create --post_type=page --post_title="Top Travel Apps You Need" --post_status=publish
wp post create --post_type=page --post_title="Best Beaches in the World" --post_status=publish
wp post create --post_type=page --post_title="Traveling with Pets: What You Need to Know" --post_status=publish
wp post create --post_type=page --post_title="City Guides: Where to Go and What to See" --post_status=publish
wp post create --post_type=page --post_title="Travel Insurance: Do You Need It?" --post_status=publish
wp post create --post_type=page --post_title="How to Travel with Kids" --post_status=publish

wp term create category "European Adventures" --description="Explore the best of Europe."
wp term create category "Asian Escapades" --description="Discover the wonders of Asia."
wp term create category "African Safaris" --description="Experience the wildlife of Africa."
wp term create category "American Road Trips" --description="Best road trips in the USA."
wp term create category "Oceania Discoveries" --description="Explore Australia and New Zealand."
wp term create category "South American Journeys" --description="Adventure through South America."
wp term create category "City Guides" --description="Top cities to visit around the world."
wp term create category "Beach Holidays" --description="Best beach destinations."
wp term create category "Cultural Experiences" --description="Immerse in different cultures."
wp term create category "Historical Travels" --description="Visit historical sites."
wp term create category "Luxury Travels" --description="Travel in luxury."
wp term create category "Budget Travels" --description="Travel on a budget."
wp term create category "Family Vacations" --description="Best destinations for families."
wp term create category "Romantic Getaways" --description="Perfect destinations for couples."
wp term create category "Adventure Travel" --description="For the adventurous souls."
wp term create category "Food and Travel" --description="Explore culinary delights."
wp term create category "Sustainable Travel" --description="Eco-friendly travel tips."
wp term create category "Solo Travel" --description="Tips for traveling alone."
wp term create category "Travel Tips" --description="General travel advice."
wp term create category "Cruise Holidays" --description="Best cruises to take."

wp term create post_tag "Travel Tips" --description="Essential tips for travelers."
wp term create post_tag "Adventure" --description="For the thrill-seekers."
wp term create post_tag "Beaches" --description="Best beaches around the world."
wp term create post_tag "Cultural" --description="Cultural experiences and festivals."
wp term create post_tag "Historical" --description="Visit historical sites and monuments."
wp term create post_tag "Luxury" --description="Luxury travel experiences."
wp term create post_tag "Budget" --description="How to travel on a budget."
wp term create post_tag "Family" --description="Best destinations for families."
wp term create post_tag "Romantic" --description="Romantic getaways for couples."
wp term create post_tag "Solo" --description="Tips for solo travelers."
wp term create post_tag "Foodie" --description="Explore local cuisines."
wp term create post_tag "Nature" --description="Nature and wildlife experiences."
wp term create post_tag "Urban" --description="City travel guides."
wp term create post_tag "Road Trip" --description="Best road trips."
wp term create post_tag "Cruise" --description="Cruise holidays."
wp term create post_tag "Hiking" --description="Best hiking trails."
wp term create post_tag "Beach" --description="Top beach destinations."
wp term create post_tag "Mountain" --description="Mountain adventures."
wp term create post_tag "Island" --description="Island hopping experiences."
wp term create post_tag "Festival" --description="Top festivals to attend."
```

</p>
</details> 

Now:

1. Edit a template or post.
2. Insert a link. A good place to test this is in the Navigation block.
3. Search for something that's not a post, e.g. a tag or a category.

## Screenshots or screencast 

Before:

https://github.com/WordPress/gutenberg/assets/612155/ad2fb81c-27ef-4d10-8545-d53ff73128db

After:

https://github.com/WordPress/gutenberg/assets/612155/3e26b2aa-831d-458d-9974-42592a30d7a0


ellatrix:
Is this really a bug? Should this be backported for WP 6.6? If so, please add the `Backport to WP Beta/RC` label. Thanks!

packages/core-data/src/fetch/index.js
```diff
@@ -8,7 +8,7 @@ import { camelCase } from 'change-case';
  */
 import apiFetch from '@wordpress/api-fetch';
 
-export { default as __experimentalFetchLinkSuggestions } from './__experimental-fetch-link-suggestions';
+export { default as fetchLinkSuggestions } from './fetch-link-suggestions';
```
ellatrix: Do we really want to make this public?

noisysocks: It's been public since WP 5.8, so I'm removing `__experimental` to reflect that it's stable.

noisysocks:
> Is this really a bug? Should this be backported for WP 6.6? If so, please add the `Backport to WP Beta/RC` label. Thanks!

I think so. The existing experience is broken. I don't think it needs to be backported to 6.6.

packages/core-data/src/fetch/index.js
ellatrix: I don't think I agree that, because something has been experimental for a long time, it should be stable.

noisysocks: I think we should strive to eventually stabilise, deprecate, or privatise all of our `__experimental` APIs. This one has existed for 3 years without much change, could be of some use to a third party, and is available in core, so I think stabilise is the best of those options.

To flip the question: why _not_ stabilise this?

I don't really care, it's not essential at all to this PR, just trying to tidy the codebase incrementally 🙂

ntsekouras: I think it's fine to stabilize for all the reasons @noisysocks mentions. It's already public and is used in core.

noisysocks:
Legit test failures due to TypeScript errors. I also need to update all the existing calls to __experimentalFetchLinkSuggestions. Tuesday's problem! Have a nice weekend.

**edit:** done.

packages/core-data/src/fetch/index.js
ellatrix: Where? https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20__experimentalFetchLinkSuggestions&type=code

ntsekouras: As part of Gutenberg.

noisysocks: You can check by disabling Gutenberg and running `wp.coreData.__experimentalFetchLinkSuggestions` in DevTools.

ellatrix: Yes, I know it's there, but @ntsekouras said _used_ in core, which I can't find. Yes it is used internally in GB, for which we have private APIs.

skorasaurus:
Thanks for this; As I understand, this would also close https://github.com/WordPress/gutenberg/issues/39964

packages/core-data/src/fetch/index.js
noisysocks: Right. No, core doesn't use it, but it's available for plugins and sites to use.

Do you think `fetchLinkSuggestions` should be a private API?

To me it seems like a useful API for third parties that want to hook up `@wordpress/block-editor` in a WordPress environment, so even if we discount the "it already has been public for some time" argument, I'd lean towards stabilising it.

noisysocks:
Rebased this and fixed the tests. It's ready for review.

> I tried simple keyword matching as an alternative algorithm for ranking results. This is where you award 1 point per keyword that the title and search term have in common. It didn't work as well for me in testing, though, because cosine similarity will give smaller titles that are similar to the query an edge over long titles that contain lots of irrelevant words in addition to the query.

The simpler approach achieves _good enough_ results so I am happy to switch to it if cosine similarity is too difficult to understand or more than we need. Let me know, I won't be offended.

packages/core-data/src/fetch/fetch-link-suggestions.ts
```diff
@@ -0,0 +1,356 @@
...
+	editorSettings: EditorSettings = {}
+): Promise< SearchResult[] > {
+	if (
+		searchOptions.isInitialSuggestions &&
+		searchOptions.initialSuggestionsSearchOptions
+	) {
+		searchOptions = {
+			...searchOptions,
+			...searchOptions.initialSuggestionsSearchOptions,
+		};
```
andrewserong: Apologies if this is obvious, but why does `searchOptions` need to be mutated and given its own value back again? If it's needed is it worth adding a reason why?

packages/core-data/src/fetch/fetch-link-suggestions.ts
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/**
+ * Returns a map of term frequencies.
+ *
+ * For example "the cat in the hat" becomes `{ the: 2, cat: 1, in: 1, hat: 1 }`.
+ *
+ * The map can be interpreted as a vector in a multi-dimensional space, where
+ * each dimension corresponds to a unique word, and the value of the dimension
+ * is the frequency of that word.
```
andrewserong: Not a blocker: but in this case it's a flat array of terms, so my understanding in looking at the code is that it's a single dimension of words and their respective quantities, rather than a multi-dimensional space. Or to put this comment differently: is there a simpler way to describe this? Or is it key that we flag this is a kind of vector so that the description for `getCosineSimilarity` makes sense?

packages/core-data/src/fetch/fetch-link-suggestions.ts
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * For example, `"I'm having a ball"` becomes `[ "im", "having", "a", "ball" ]`.
+ *
+ * @param text
+ */
+function tokenize( text: string ) {
+	return text
+		.toLowerCase()
+		.split( /\s+/ )
+		.map( ( word ) => word.replace( /\W/g, '' ) )
```
ramonjd: This strips out non-word characters including non-latin and accent/diacritical marks, e.g., `お寿司が欲しいです` or the `ö` in `Ich möchte Sushi`

The results seem to be accurate enough —

Screenshot 2024-06-13 at 2 52 38 PM

— but I was just wondering if it was getting the benefit of all the bag of words/cosine similarity work you've done!

ramonjd (commented):
I've been testing this with lots and lots of taxonomy terms, posts and pages in English, German and Japanese.

It's hard to see the benefit of these changes at first — trunk behaves mostly the same until you have 100s of pages/posts with similar keywords.

Here's me looking for "Paris"

https://github.com/WordPress/gutenberg/assets/6458278/23dee8ef-d40d-4887-a20e-5f2331b6011a

Great stuff - overall I think it's a big improvement to have tags/cats surfaced this way.

ramonjd:
> The simpler approach achieves good enough results so I am happy to switch to it if cosine similarity is too difficult to understand or more than we need. Let me know, I won't be offended.

Chat-GPT and "ELI5" set me straight 😄 

But it is very clever, so I paused at whether this will be accessible to folks who want to iterate on the feature. 

Is the "simpler" approach less code? Does it perform as well? Easier to read? 

If the answer is "yes" to these questions I'd probably consider using the simpler approach, or at least stating a convincing reason to go with cosine similarity, e.g., it produces much better results more of the time. 

What do you think?

packages/core-data/src/fetch/fetch-link-suggestions.ts
```diff
@@ -0,0 +1,356 @@
...
+ * - A cosine similarity of 0 indicates that the two vectors are orthogonal (no similarity).
+ * - A cosine similarity of -1 indicates that the two vectors are diametrically opposed.
+ *
+ * In the context of text similarity, a higher cosine similarity indicates that the documents share
+ * more common terms and are therefore more similar in content.
+ *
+ * @param vector1
+ * @param vector2
+ */
+function getCosineSimilarity(
```
andrewserong: Is is worth adding unit tests for this in particular, or is it enough that it gets called as part of the test for `sortResults`?

andrewserong (commented):
Great work here, and thanks for the detailed explanations for how this works, and for the test wp cli commands 👍

> It didn't work as well for me in testing, though, because cosine similarity will give smaller titles that are similar to the query an edge over long titles that contain lots of irrelevant words in addition to the query.

IMO I think the advantage of smaller titles that more directly match the query could be worth it, especially if you wind up having a simple category like `Travel` where you'd expect it to be at the top. With this PR applied, it's working very nicely for me:

image

Whereas on `trunk` I get pages first of all, and categories and tags are way down at the bottom of the list.

| Trunk top of list | Trunk bottom of list |
| --- | --- |
| image | image |

Just left a few questions, but overall I think this is a big improvement, and I also like the idea of stabilising the API for it. The function has been around for a long time and it seems general purpose enough to be useful in situations outside of Gutenberg to me (i.e. I could imagine a plugin in an admin area wanting a quick way of fetching link suggestions).

Would it be worth getting a second opinion / see if anyone objects to the more complex cosine similarity approach?

noisysocks:
> But it is very clever, so I paused at whether this will be accessible to folks who want to iterate on the feature.
>
> Is the "simpler" approach less code? Does it perform as well? Easier to read?
>
> If the answer is "yes" to these questions I'd probably consider using the simpler approach, or at least stating a convincing reason to go with cosine similarity, e.g., it produces much better results more of the time.

This has been on my mind as well and I like how you've put the question.

I spent some time just now evaluating the simpler approach of having `sortResults` use keyword matching. Here's what it looks like:

<details><summary>Code</summary>
<p>

```ts
/**
 * Sort search results by relevance to the given query.
 *
 * Sorting is necessary as we're querying multiple endpoints and merging the results. For example
 * a taxonomy title might be more relevant than a post title, but by default taxonomy results will
 * be ordered after all the (potentially irrelevant) post results.
 *
 * We use simple keyword matching to determine relevance. The more keywords in the search query
 * that are found in the title, the higher the relevance score.
 *
 * @param results
 * @param search
 */
export function sortResults( results: SearchResult[], search: string ) {
	const searchTokens = tokenize( search );

	const scores = {};
	for ( const result of results ) {
		const titleTokens = tokenize( result.title );
		scores[ result.id ] = titleTokens.reduce( ( score, token ) => {
			return searchTokens.includes( token ) ? score + 1 : score;
		}, 0 );
	}

	return results.sort( ( a, b ) => scores[ b.id ] - scores[ a.id ] );
}

/**
 * Turns text into an array of tokens.
 *
 * For example, `"I'm having a ball"` becomes `[ "im", "having", "a", "ball" ]`.
 *
 * @param text
 */
function tokenize( text: string ) {
	return text
		.toLowerCase()
		.split( /\s+/ )
		.map( ( word ) => word.replace( /\W/g, '' ) )
		.filter( ( word ) => !! word );
}
```

</p>
</details>

It's about 20 fewer lines of code (ignoring comments) but the big difference is that we don't need `getCosineSimilarity` which is obviously quite complicated to understand.

But, in testing it just now, it _really_ doesn't perform well. Keyword matching doesn't take the number of non-matching keywords into account. If I'm searching "vegetarian travel" then the title "Top tips for vegetarian travel" and the title "Vegetarian travel" both have scores of 2 even though the latter is exactly what the user entered.

This becomes a big problem when you search for a very common keyword. For example, with the real world data I'm testing against (a travel blog with 290 posts, 72 categories, and 2532 tags), if you search "europe" you are not able to see the category titled "Europe" because there are more than 20 posts that contain the word "Europe" somewhere in the title.

| Keyword matching | Cosine similarity |
|--------|--------|
| Screenshot 2024-06-13 at 15 23 26 | Screenshot 2024-06-13 at 15 22 31 |

We could look at adjusting the keyword matching algorithm to _subtract_ from the score when there is a non-matching keyword (for example: +1 when keyword matches, -0.1 if not) but it's a bit fraught because you then have to decide how many non-matching keywords is equivalent to one matching keyword.

So, all in all, I think the complexity is worth it.

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: Good catch! I've made `tokenize` unicode aware.

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: Removed the mutation 👍 

Thinking about it though, `isInitialSuggestions` is a pretty strange API design. It should be the callers responsibility... I'll look at cleaning this up since we're stabilising.

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: Yeah so it **is** a dictionary that maps words to their respective quantities. This dictionary can _also_ be viewed as a multi-dimensional vector where each dimension is a unique word and the value of that dimension is the quantity.

So `'brown brown fox'` becomes `{ brown: 2, fox: 1 }` which can be thought of as the 2d vector of `(2, 1)`.

Pretending that the dictionary is a vector is the magic trick that lets us compare how similar two dictionaries are by calculating the angle between the two vectors. (But instead of angle we use _cosine_ of the angle because that's easier to do using vector operations: dot product, magnitude.)

I'll try to update the comments to explain better 😀

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: Yeah, why not. I can add simple unit tests for all the helper functions.

ramonjd:
> So, all in all, I think the complexity is worth it.

Thanks for going down that rabbit hole for the spectators 😄  A win for users.

packages/core-data/src/fetch/fetch-link-suggestions.ts
andrewserong: Ah, thanks! That helps clarify it for me 👍

noisysocks:
I tried a scoring approach where we simply divide the number of tokens in the title that are also in the search query, divided by the total number of tokens in the title. This means shorter titles receive higher scores, fixing the problem I noted in https://github.com/WordPress/gutenberg/pull/62397#issuecomment-2164496585.

It seems to work well enough in my testing and is much simpler to understand. I'd appreciate if you can test it with various search terms, etc. again though 🙂

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: I added more tests.

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: Trying a simpler scoring function so maybe don't need this.

noisysocks:
The more I look at the API of `fetchLinkSuggestions` the less I like it 😅 so I'm going to keep it experimental in this PR and come back to stabilising it / cleaning it up in a follow-up PR.

packages/core-data/src/fetch/index.js
noisysocks: Ended up deciding to keep this experimental for now anyway. https://github.com/WordPress/gutenberg/pull/62397#issuecomment-2167265679

andrewserong:
This is still testing nicely for me after the latest change 👍

> [Keep experimental for now](https://github.com/WordPress/gutenberg/pull/62397/commits/a163f5b1d984e4ad3613c3f8ea9a984364cce9ce)

Is this since we're still iterating on the logic? Sounds reasonable to defer stabilising it for now.

packages/core-data/src/fetch/fetch-link-suggestions.ts
noisysocks: Later's problem https://github.com/WordPress/gutenberg/pull/62397#issuecomment-2167265679

noisysocks:
Updated PR description. This is altogether a much simpler PR now 😅

> Is this since we're still iterating on the logic? Sounds reasonable to defer stabilising it for now.

No but that's not a bad reason either.

andrewserong (approved):
This is still testing nicely for me, and I like that the logic is simpler, easier to read and to maintain. Not for now, but another potential future enhancement could be to look at partial matches with tokens / fuzzy search, too, as I need to type the whole word "travel" before the travel category gets to the top of the list:

"trave"

image

"travel"

image

This is already a big improvement, though, so just jotting this down as a thought, nothing to worry about for now.

LGTM! 🚀

ramonjd (approved):
Ran through similar tests as before. 500+ database entities with different languages.
Tags and categories are surfaced as expected. Snappy results.

https://github.com/WordPress/gutenberg/assets/6458278/c0be18cc-fecb-431d-9bc0-f322eb9a71b8


Very nice. 🚢

noisysocks:
Thanks for bearing 🐻 with me while I ventured unnecessarily deep into rabbit 🐰 holes!

PR #62398, opened 2024-06-07
Build JS module only in development mode

kevin940726:
## What?

Part of https://github.com/WordPress/gutenberg/discussions/36142.

When running `npm run dev`, only build the modern module JS and skip building commonJS code.

## Why?

I don't see much value in building commonJS code in development mode anymore. It seems like every tooling we use only leverages the module code anyway. We still want to keep the commonJS code for production build for now for backward-compatibility.

This should theoretically speed up development build and each incremental builds by 2X.

## How?

Add `NODE_ENV=development` and skip the commonJS code in build.

## Testing Instructions

1. `npm run dev` should still work as expected and only output the `build-module` folder for each package.
2. `npm run build` should still build the commonJS code and output `build` folder for each package.

N/A

## Screenshots or screencast 
N/A


gziolo (approved):
That makes perfect sense ❤️

Unrelated to that, I wonder for how long we should keep CJS version before transitioning to ESM in all packages.

PR #62399, opened 2024-06-07
Docs: Explicitly mention new behavior coming in WP 6.6 for block variations

gziolo:
## What?

Adds mentions about the behavior change coming in WP 6.6 for Block Variations.

## Why?

WordPress 6.6 is still not released, so it will be helpful to clarify the differences between WP major versions.

## How?

Updates to the documentation.

Code styling changes come from the Prettier integration, which I have always enabled in my IDE.

### DIFF
--- a/docs/reference-guides/block-api/block-variations.md
+++ b/docs/reference-guides/block-api/block-variations.md
@@ -1,11 +1,10 @@
 # Variations
 
-The Block Variations API  allows you to define multiple versions (variations) of a block. A block variation differs from the original block by a set of initial attributes or inner blocks. When you insert the block variation into the Editor, these attributes and/or inner blocks are applied.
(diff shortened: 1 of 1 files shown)


fabiankaegy (approved):
Thanks for these updates :)

ockham (approved):
Thank you!

ockham:
I added the "Backport to WP Beta/RC" label since I think we'd like these changes in 6.6, as that's what determines what will be at https://developer.wordpress.org/ IIUC.

gziolo:
> I added the "Backport to WP Beta/RC" label since I think we'd like these changes in 6.6, as that's what determines what will be at https://developer.wordpress.org/ IIUC.

Historically, we have always been using `trunk` to generate developer docs for the block editor. So, we can verify that in 15-20 minutes. Link:

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#using-isactive

gziolo:
The changes are already synced in the block editor handbook. Removed the label.

PR #62400, opened 2024-06-07
Interactivity API: Convert types of generator functions to async functions

luisherranz:
## What?

This pull request updates the store return types to have their generators converted into async functions.

This is still a work in progress, as I would like to find a way to properly test these types before merging this change.

## Why?

Even though developers have to use generators to create asynchronous functions in the stores, the store proxy returns an async function, not a generator. So when those actions are used as async functions, TypeScript still considers them as generators.

The async function is created on the fly by the store proxy: 

https://github.com/WordPress/gutenberg/blob/0d1943cf84e60ad4646333ed68eca7cc7b879009/packages/interactivity/src/store.ts#L108-L109

## How?

With a utility that checks if the type corresponds to a generator and if so, changes it to an async function recursively.


packages/interactivity/src/store.ts
```diff
@@ -207,6 +207,24 @@ interface StoreOptions {
 	lock?: boolean | string;
 }
 
+// Utility type to check if a function is a generator function.
+type IsGenerator< T > = T extends (
+	...args: infer A
+) => Generator< infer R, any, any >
+	? ( ...args: A ) => Promise< R >
+	: never;
```
sirreal: This doesn't look quite right, the generators should _yield_ promises, but I think the async function this transforms to should resolve to the returned value. Probably something more like this:

```suggestion
) => Generator< any, infer R, any >
	? ( ...args: A ) => Promise< R >
	: never;
```

The first type argument could be `Promise< any >`, but that's probablye too strict. We await the yielded values and you can await values of any type (although it only makes sense for promises).

The name `IsGenerator` is a bit strange because this actually does the conversion.

luisherranz: Indeed, thank you 🙂

Fixed in https://github.com/WordPress/gutenberg/pull/62400/commits/5c73a529452d0b8c82ec0dc1827681e5a198ead2.

luisherranz:
In the end, I ended up adding a system to test types in Gutenberg, including a npm script and a GitHub Action.

This is how it works:

<div>
    <a href="https://www.loom.com/share/8394c04f3273422cbaf967db0fb42923">
      
    </a>
  </div>

https://www.loom.com/share/8394c04f3273422cbaf967db0fb42923

---

The tests I have added to verify that the types of the Store and the Interactivity API work correctly are still very basic, but I will add more tests to cover all possibilities before merging this pull request.

sirreal:
Rather than rolling our own type testing, [let's evaluate the options listed here](https://github.com/microsoft/DefinitelyTyped-tools/tree/main/packages/dtslint#just-looking-for-a-type-testing-tool):

> If you are just looking for a TypeScript type testing tool, use:
> 
> - [tsd](https://github.com/tsdjs/tsd)
> - [eslint-plugin-expect-type](https://github.com/JoshuaKGoldberg/eslint-plugin-expect-type)
> - [tstyche](https://github.com/tstyche/tstyche)

luisherranz:
Oh, nice. Can any of those be integrated with the existing Jest setup?

sirreal:
I'm not directly familiar with any of them, we'd need to evaluate which is best.

It looks like at least one integrates with eslint and uses special comments, so that might be something to add directly to source files and errors would appear in linting. Another seems to extend jest to check types so that would be part of unit tests.

mirka:
> Rather than rolling our own type testing, [let's evaluate the options listed here](https://github.com/microsoft/DefinitelyTyped-tools/tree/main/packages/dtslint?rgh-link-date=2024-07-02T09%3A55%3A51Z#just-looking-for-a-type-testing-tool):

For what it's worth, in the components package we've simply been [appending](https://github.com/search?q=repo%3AWordPress%2Fgutenberg+%22static+typ%22+path%3A%22%2Ftest%2F%22&type=code) a "static type checks" section to our unit tests, which has been good enough for our needs. We only feel the need for type tests when there are complex conditionals involved, so not often.

luisherranz:
We'll try that. Thanks @MIRKA!

luisherranz:
@mirka, I have added the type tests, including a failing test, to a unit test file, but I don't see the error being reported in any way. Neither by Jest when you run the tests, nor when you do `npm run dev` in the build process or later in the watch. 

This is the file: https://github.com/WordPress/gutenberg/blob/14cc9c4b70d8d6939b4d499c0238268d4b604827/packages/interactivity/src/test/store.ts

What is your workflow to see that a test is failing? Or do I have to add that unit test file somewhere?

mirka:
> @mirka, I have added the type tests, including a failing test, to a unit test file, but I don't see the error being reported in any way. Neither by Jest when you run the tests, nor when you do `npm run dev` in the build process or later in the watch.

It looks like you're set up correctly. The type tests are static, so problems will be caught by the type checker. On CI, that runs in "Static Analysis". So if your passing test case starts to fail, or if the `expect-error` test case stops erroring, TS can flag it both locally (e.g. in IDE, `lint-staged`) and on CI. Is that not what you're seeing?

luisherranz:
> Is that not what you're seeing?

That's not what I'm seeing, indeed. I have included a test that should fail [here](https://github.com/WordPress/gutenberg/blob/433cd28795c6d1e528105cf00c7946af0edbb7e3/packages/interactivity/src/test/store.ts#L35-L37), but as you can see in the GitHub Actions, the Static Analysis has passed correctly. It also didn't give me any warnings when committing (`lint-staged`).

- https://github.com/WordPress/gutenberg/actions/runs/10385526502/job/28754870771

luisherranz:
I think we also need to convert the input types so that people can define their actions using asynchronous functions, and these, internally, are compatible with generators. In other words, this not only has to work with inferred types, but also with explicit types.

Users should be able to define store types like this:

```ts
type MyStore = {
  actions: {
    myAsyncAction: () => Promise<void>;
  };
};

// This should be fine.
const { actions } = store<MyStore>('...', {
  actions: {
    *myAsyncAction() {
      // ...
    },
  },
});

// This should also be fine.
await actions.myAsyncAction();
```

luisherranz:
I just realized that I made some modifications from `github.dev` where the linter doesn't run, nor does the fork of Prettier, and I broke the Static Analysis test, but that has nothing to do with the previous issue, sorry. I'll fix it now from my local environment.

mirka:
> That's not what I'm seeing, indeed. I have included a test that should fail [here](https://github.com/WordPress/gutenberg/blob/433cd28795c6d1e528105cf00c7946af0edbb7e3/packages/interactivity/src/test/store.ts#L35-L37)

You're right, I can confirm that removing a `@ts-expect-error` from your tests don't fail the `tsc` build. Doing the same thing in a type test in the `@wordpress/components` _will_ break the build. So, some kind of tsconfig issue? (I'm testing with `npx tsc --build` by the way.)

luisherranz:
Replaced by #64577

I have moved the changes of TypeScript and the tests to that pull request, although they are not failing there either.

cc: @mirka

PR #62402, opened 2024-06-07
Editor: Deprecate PostSwitchToDraftButton

ntsekouras:
## What?
PostSwitchToDraftButton is no longer used in our codebase, so let's deprecate it.


Mamaduka (approved):
🚢

PR #62405, opened 2024-06-07
Global Styles: Do not sanitize block style variations based on registration status

### LINKED ISSUE #62303: Section styles: editing a block variation in Styles doesn't reflect in the editor

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/62303

## What?

Only sanitize the properties of block style variations within theme.json instead of also omitting any that are not currently registered.

## Why?

With the introduction of the Section Styles feature that leverages extended block style variations, there are a number of methods through which block style variations can be registered. This primarily happens via the theme.json data filters in the `WP_Theme_JSON_Resolver` class.

Unfortunately, there are several places where global styles data is processed (such as the global styles REST API endpoint) that do not apply these filters and instantiate the `WP_Theme_JSON` class directly. The result then is some block style variations are not registered and are incorrectly stripped during the sanitization of the theme.json data.

## How?

This PR:
- Removes the sanitization of unregistered block style variations
- Maintains the sanitization of style properties within all block style variations
- Prevents any unregistered block style variations from being included in a block types node data during style generation
    - This avoids issues where a variation does not have a defined selector and invalid CSS is generated

## Testing Instructions

There are some [duotone support actions](https://github.com/WordPress/gutenberg/blob/trunk/lib/block-supports/duotone.php#L21-L22) in Gutenberg that trigger all the theme json filters. So to test this fix fully, those actions need to be temporarily disabled.

1. Comment out the [duotone support actions](https://github.com/WordPress/gutenberg/blob/trunk/lib/block-supports/duotone.php#L21-L22) that masked this issue
2. Add a custom block style variation to your theme.json (see https://github.com/WordPress/gutenberg/issues/62303 for snippet)
3. Navigate to Appearance > Editor
4. Select a group block and apply the custom variation
5. Navigate to Styles > Blocks > Group > dark (variation name)
6. Change the background color for the variation and confirm it is reflected in the editor
7. Save the changes and confirm the background color is still correct in the editor
8. For good measure check the frontend
9. Uncomment the duotone support actions from step 1 and repeat the process. It should still work.

## Screenshots or screencast 

| Before | After |
|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/60436221/a7d59fad-29a1-40d1-84c7-9606f170aef5" /> | <video src="https://github.com/WordPress/gutenberg/assets/60436221/3ad6e58b-d05b-4048-8e35-79987ed1808c" /> |

### DIFF
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -745,8 +745,7 @@ class WP_Theme_JSON_Gutenberg {
 		$registry            = WP_Block_Type_Registry::get_instance();
 		$valid_block_names   = array_keys( $registry->get_all_registered() );
 		$valid_element_names = array_keys( static::ELEMENTS );
-		$valid_variations    = static::get_valid_block_style_variations();
-		$this->theme_json    = static::sanitize( $this->theme_json, $valid_block_names, $valid_element_names, $valid_variations );
+		$this->theme_json    = static::sanitize( $this->theme_json, $valid_block_names, $valid_element_names );
 		$this->theme_json    = static::maybe_opt_in_into_settings( $this->theme_json );
 
 		// Internally, presets are keyed by origin.
@@ -847,10 +846,9 @@ class WP_Theme_JSON_Gutenberg {
 	 * @param array $input               Structure to sanitize.
 	 * @param array $valid_block_names   List of valid block names.
 	 * @param array $valid_element_names List of valid element names.
-	 * @param array $valid_variations    List of valid variations per block.
 	 * @return array The sanitized output.
 	 */
-	protected static function sanitize( $input, $valid_block_names, $valid_element_names, $valid_variations ) {
+	protected static function sanitize( $input, $valid_block_names, $valid_element_names ) {
 
 		$output = array();
 
@@ -927,13 +925,9 @@ class WP_Theme_JSON_Gutenberg {
 			$style_variation_names = array();
 			if (
 				! empty( $input['styles']['blocks'][ $block ]['variations'] ) &&
-				is_array( $input['styles']['blocks'][ $block ]['variations'] ) &&
-				isset( $valid_variations[ $block ] )
+				is_array( $input['styles']['blocks'][ $block ]['variations'] )
 			) {
-				$style_variation_names = array_intersect(
-					array_keys( $input['styles']['blocks'][ $block ]['variations'] ),
-					$valid_variations[ $block ]
-				);
+				$style_variation_names = array_keys( $input['styles']['blocks'][ $block ]['variations'] );
 			}
 
 			$schema_styles_variations = array();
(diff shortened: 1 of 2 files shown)


phpunit/class-wp-theme-json-test.php
```diff
@@ -1463,12 +1463,14 @@ public function test_get_stylesheet_generates_fluid_typography_values() {
...
-			$theme_json->get_stylesheet()
+			$stylesheet
```
aaronrobertshaw, line 1473: It turned out the deregistration of the test block before generating the stylesheet was causing this test to fail on this branch. Not 100% sure why yet but I'll look into that on Monday.

oandregal:
Hi Aaron, after your braindump (thank you!) I took a look at this issue and realized there is an alternative way to fix it: we can register the block style variations defined by the theme on `init` instead of using `wp_theme_json_data_*` filters.

The `init` hook is what themes had to use until now in their `functions.php`. Unless I'm missing some use case, this is a safe place to register the block style variations as we can always count on them being registered. I think it's best if we do that rather than trying to patch specific paths, sanitization, etc. It's also simpler to reason about.

The proposal is at https://github.com/WordPress/wordpress-develop/pull/6756 I had some issues in my environment and ended up preparing it for core.

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -2632,10 +2626,14 @@ private static function get_block_nodes( $theme_json, $selectors = array() ) {
 			$variation_selectors = array();
 			if ( isset( $node['variations'] ) ) {
 				foreach ( $node['variations'] as $variation => $node ) {
-					$variation_selectors[] = array(
-						'path'     => array( 'styles', 'blocks', $name, 'variations', $variation ),
-						'selector' => $selectors[ $name ]['styleVariations'][ $variation ],
-					);
+					// Only add registered variations to this node's variations to
+					// avoid CSS generated without a selector.
+					if ( isset( $selectors[ $name ]['styleVariations'][ $variation ] ) ) {
```
oandregal, line 2631: So, what this PR does is:

- do not remove the **unregistered** variations when we build the theme.json in memory
- but make sure we only process the **registered** ones when we use them

This is the kind of thing that I mentioned: we are not removing sanitization but moving it **everywhere**. Each time we have to work with them, we need to make sure to use the ones that are registered.

aaronrobertshaw:
Closing this PR in favour of moving the block style registration to `init` as [detailed above](https://github.com/WordPress/gutenberg/pull/62405#issuecomment-2155131064).

PR #62406, opened 2024-06-07
Posts Dashboard: Add a new experimental empty page

youknowriad:
Related #62371 

## What?

This PR doesn't do much, it adds a blank page under the "Gutenberg" submenu when you enable the "posts dashboard" experiment. The idea is that it gives us a blank canvas to explore the DataViews for posts like outlined in #62371 

It does make some choices:

I'm planning to reuse components and code from the site editor, so I had two choices:

 - Create a new package for this page (edit-posts or something) and extract the common code from edit-site package.
 - Add a new specific initialization function to the edit-site package.

I think ultimately, we'll probably end up with the former, but I don't want to create early abstractions that we'll be throwing later, so I added a new experimental function that only works in the Gutenberg plugin (tree-shaked from WordPress Core) and left it in the edit-site package for now.

## Testing Instructions

1- Enable the "Posts Dashboard Experiment"
2- You'll see a "posts" page appear under the "Gutenberg" menu. 
3- The page is empty for now, it just shows: "Welcome to Posts"

### DIFF
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Bootstraps the new posts dashboard page.
+ *
+ * @package Gutenberg
+ */
+
+add_action( 'admin_menu', 'gutenberg_replace_posts_dashboard' );
+
+/**
+ * Renders the new posts dashboard page.
+ */
+function gutenberg_posts_dashboard() {
+	wp_register_style(
+		'wp-gutenberg-posts-dashboard',
+		gutenberg_url( 'build/edit-site/posts.css', __FILE__ ),
+		array(),
+	);
+	wp_enqueue_style( 'wp-gutenberg-posts-dashboard' );
+	wp_add_inline_script( 'wp-edit-site', 'window.wp.editSite.initializePostsDashboard( "gutenberg-posts-dashboard" );', 'after' );
+	wp_enqueue_script( 'wp-edit-site' );
+
+	echo '<div id="gutenberg-posts-dashboard"></div>';
+}
+
+/**
+ * Replaces the default posts menu item with the new posts dashboard.
+ */
+function gutenberg_replace_posts_dashboard() {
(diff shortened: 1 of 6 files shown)


fabiankaegy (approved):
Works as advertised  👍 Excited to see this work kick off :) 

(Would it make sense to have a new label for this work? (`Admin Redesign` or something similar?))

youknowriad:
I'd be up for a label but not "Admin Redesign", I'd do more "Posts Dashboard" or something. "Admin Redesign" would be too broad to me. Calling @annezazu the expert of labels just in case :)

youknowriad:
Ok I added a label for now, feel free to edit/rename it...

annezazu:
At first glance, I'm not sure why we need a separate label outside of using Data Views, unless we want individual labels for each spot data views is applied to. I'd want to name this label: "Applying Data Views" or "Expanding Data Views" if anything as it seems to also entail applying data views to posts, tags, and categories. Maybe even something more practical like "Data Views outside Site Editor". The current label description for "data views" talks about how the work will apply outside of the site editor and that leads me back to wanting to stick, for now, with just the Data Views label overall until the work progresses more. How does that sound?

youknowriad:
For me at least, DataViews is more about the component itself, the framework that powers all these pages while this one is more about the new WPAdmin pages that will contain both the posts, categories and tags dataviews but also their forms/editors. 

That said, I don't mind much whatever you all think is best, works for me.

annezazu:
Got it! I updated it to:

>[Feature] Posts/Tags/Categories Screen

With the following as the description:

>Bringing Data Views to Posts, Categories and Tags alongside their own Forms/Editors.

This matches bringing [[Feature] Widgets Screen](https://github.com/WordPress/gutenberg/labels/%5BFeature%5D%20Widgets%20Screen) which brought the block-based screen that replaced widgets.php and feels like an apt comparison for consistency rather than the word dashboard which is a bit loaded.

If the title feels too long, I think we can go with just `[Feature] Posts Screen`.

Issue #62396, opened 2024-06-07
Pattern Overrides not rendering on Front End


jamiemarsland:
### Description

Editing Pattern Overrides isn't working for me in WordPress 6.6 beta 1

### Step-by-step reproduction instructions

1. Create Pattern
2. Enable Overrides
3. Insert Pattern into Page
4. Edit Override content
5. Publish Page 
6. Override content isn't displaying properly

### Screenshots, screen recording, code snippet

https://vimeo.com/954664238/63cf83c1ed?share=copy

### Environment info

WordPress Beta 6.6 (1) 
Tested with Gutenberg plugin installed and also tested without Gutenberg plugin installed

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

SantosGuillamot:
Thanks for reporting! If I am not mistaken, it should be fixed after [these changes](https://github.com/WordPress/gutenberg/pull/62355). And this is [the backport in WordPress](https://github.com/WordPress/wordpress-develop/pull/6744).

talldan:
It looks exactly like the same bug as https://github.com/WordPress/gutenberg/issues/62347, so let's close this now that a fix has been merged. If it's not resolved for beta 2, we can reopen the issue.

Thanks for reporting it, @jamiemarsland.

PR #62410, opened 2024-06-07
Add e2e test to ensure block bindings work well with symbols and numbers

SantosGuillamot:
## What?
Add an e2e case to cover use cases like [this one](https://github.com/WordPress/gutenberg/issues/62396#top), where symbols and numbers couldn't be used at the start of a binding value.

## Why?
To ensure we don't break things again in the future.

## How?
I just added a e2e test for a paragraph connected to a custom field. I thought about replicating the same for pattern overrides, but I felt I was just replicating the test. Happy to add it if considered necessary.

## Testing Instructions
Tests should pass once [this code](https://github.com/WordPress/wordpress-develop/pull/6744) in WordPress is updated.

### DIFF
--- a/test/e2e/specs/editor/various/block-bindings.spec.js
+++ b/test/e2e/specs/editor/various/block-bindings.spec.js
@@ -1975,6 +1975,41 @@ test.describe( 'Block bindings', () => {
 				).toHaveText( 'new value' );
 			} );
 
+			// Related issue: https://github.com/WordPress/gutenberg/issues/62347
+			test( 'should be possible to use symbols and numbers as the custom field value', async ( {
+				editor,
+			} ) => {
+				await editor.insertBlock( {
+					name: 'core/paragraph',
+					attributes: {
+						anchor: 'paragraph-binding',
+						content: 'paragraph default content',
+						metadata: {
+							bindings: {
+								content: {
+									source: 'core/post-meta',
+									args: { key: 'text_custom_field' },
+								},
+							},
+						},
+					},
+				} );
+				const paragraphBlock = editor.canvas.getByRole( 'document', {
+					name: 'Block: Paragraph',
+				} );
+
+				await expect( paragraphBlock ).toHaveAttribute(
+					'contenteditable',
+					'true'
+				);
+				await paragraphBlock.fill( '$10.00' );
+				// Check the value of the custom field is being updated by visiting the frontend.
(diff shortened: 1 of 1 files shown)


gziolo:
> Tests should pass.

Will it pass without the fix for https://github.com/WordPress/gutenberg/issues/62396? 😄

SantosGuillamot:
> Will it pass without the fix for https://github.com/WordPress/gutenberg/issues/62396?

I think it should pass once the [core code](https://github.com/WordPress/wordpress-develop/pull/6744) is merged 😄 I tested it locally and it does.

I should have mentioned it in the issue. Adding it now.

gziolo:
> > Will it pass without the fix for #62396?
> 
> I think it should pass once the [core code](https://github.com/WordPress/wordpress-develop/pull/6744) is merged 😄 I tested it locally and it does.
> 
> I should have mentioned it in the issue. Adding it now.

Screenshot 2024-06-07 at 13 37 47

That's expected for now. Everything going according to the plan 💯

gziolo (approved):

test/e2e/specs/editor/various/block-bindings.spec.js
```diff
@@ -1975,6 +1975,41 @@ test.describe( 'Block bindings', () => {
...
+				} );
+				const paragraphBlock = editor.canvas.getByRole( 'document', {
+					name: 'Block: Paragraph',
+				} );
+
+				await expect( paragraphBlock ).toHaveAttribute(
+					'contenteditable',
+					'true'
+				);
+				await paragraphBlock.fill( '$10.00' );
```
gziolo, line 2005: I was wondering how it gets saved, but it looks like autosave triggers when opening a preview 👍🏻

SantosGuillamot:
Now that the [core code](https://github.com/WordPress/wordpress-develop/pull/6744) has been merged, the test is passing as expected. Merging it now.

EDIT: I didn't add the props properly. Sorry @gziolo ! Is there a way to fix that?

cbravobernal:
> Now that the [core code](https://github.com/WordPress/wordpress-develop/pull/6744) has been merged, the test is passing as expected. Merging it now.
> 
> EDIT: I didn't add the props properly. Sorry @gziolo ! Is there a way to fix that?

Is better to just ignore it 😆

PR #62411, opened 2024-06-07
Update disabled checkbox cursor style

### LINKED ISSUE #62349: Data views: hide checkboxes in grid layout when unable to select
When viewing items in data views in the Site Editor, checkboxes are shown even if you can't actually select anything. Every time I see this, I still think I can check the box so either the greyed out look needs to shift, the checkboxes need to be removed if you can't select an item, or something else I can't think of right now :) Connecting the dots, it looks like there's feedback to have checkboxes always be visible in certain layouts https://github.com/WordPress/gutenberg/issues/59175

Here's a video showing what I mean: 

https://github.com/WordPress/gutenberg/assets/26996883/1f0f8955-95be-4291-a9b2-79b2d23c8272

amitraj2203:
## What?
Fixes: #62349

## Why?
In the Site Editor's data views, checkboxes remain visible even when items are unselectable, causing confusion and suggesting interaction.

## How?
Changing the cursor to `not allowed` for unselectable items in the Site Editor improves clarity and usability.

## Testing Instructions

1. Open site editor.
2. Select Templates.
3. Try to select the unselectable templates.

## Screenshots or screencast 
https://github.com/WordPress/gutenberg/assets/77401999/206806b7-0242-4c43-b1df-270970e14b94

### DIFF
--- a/packages/base-styles/_mixins.scss
+++ b/packages/base-styles/_mixins.scss
@@ -301,7 +301,7 @@
 	&:disabled {
 		background: $gray-100;
 		border-color: $gray-300;
-		cursor: default;
+		cursor: not-allowed;
 
 		// Override style inherited from wp-admin. Required to avoid degraded appearance on different backgrounds.
 		opacity: 1;


jorgefilipecosta (commented):
The change looks good code-wise, but I'm going to cc: @WordPress/gutenberg-design so designers can have a look.

jameskoster:
I think this is reasonable. Visually there may not be enough contrast between enabled/disabled checkboxes, and the cursor helps mouse users. The question I have is whether the cursor should be applied holistically across _all_ disabled elements.

PR #62412, opened 2024-06-07
Add Button support for iconPosition top and showDescription

### LINKED ISSUE #62373: Button: add support for iconPosition top and showDescription

I've been exploring potential solutions for https://github.com/WordPress/gutenberg/issues/60206 and realized there's many cases where the editor uses custom components or ad-hoc CSS to modify or replace the `Button` base component.

One notable example of this is for the Group, Columns, and Query block placeholders. These placeholders design and accessibility would benefit from two new features of the Button component:
- The ability to make a button show the Icon at the top and the text below the icon. Currently, there's only support for `iconPosition?: 'left' | 'right';`
- The ability to show the Button description. Currently, when passing a description to the `describedBy` prop, a visually hidden element is rendered after the button and properly referenced with aria-describedby/id attributes. The description is visually hidden but in some scenarios it would be beneficial to make it visible.

Cc @mirka 

After a quick look at the Button component, seems to me that:
- Support for `iconPosition="top"` would be pretty trivial. The DOM order would be the same as `left`. It would likely need only a CSS class and some styling.
- A new prop `showDescription` seems pretty trivial as well. When true, it should make the button description use a paragraph insteaed of the `VisuallyHidden` component.

I'd appreciate any feedback and thoughts.

Use cases examples:

For better accessibility in the block placeholders that show variations, I'd like to change the design to make the variation title and description visible. Example screenshot:

Screenshot 2024-06-06 at 16 09 55

To evaluate:
Other editor components e.g. PreferenceToggleMenuItem > MenuItem already show both a label and a description. They do it wrongly, by putting both the label and description as content of the button. As such, the accessible name of the button is a very long string that is a barrier for speech recognition software users ang generally not a good practice. The description should be separated from teh label and likely the visible description should be supported by the base Button component instead of reimplementing ad-hoc solutions. Screenshot:

Screenshot 2024-06-06 at 16 39 50

Finally, in the future the editor may just have the need to have buttons that show icon and label vertically stacked. Example screenshot:

Screenshot 2024-06-06 at 16 15 10

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/62373

## What?

- Extends the `iconPosition` prop to support a new `top` value.
- Adds a `showDescription` prop to make the Button accessible description visible.

## Why?

As detailed in https://github.com/WordPress/gutenberg/issues/62373 there are scenarios where a Button would benefit from a couple new features:
- Show the icon on toop and the label below the icon, stacked vertically.
- Show the visually hidden accessible description. The current implementation incorrectly assumes the description has to always be visually hidden. Instead, there are cass where the description has to be visible.

Both features would allow to avoid 'ad-hoc' implementations and overrides in the editor components, which also add maintenance cost, inconsistencies, and are prone to bugs and reduced accessibility.

## How?

See above

## Testing Instructions

- Build and run the Storybook `npm run storybook:dev`.
- In the Storybook, select the Button component.
- Set an icon, and set the `iconPosition` prop to 'top.
- Observe the button renders with the icon at the top and the label text below the icon. Example screenshot:

Screenshot 2024-06-07 at 15 08 29

- Set a description text for the `describedBy` prop.
- Set the new `showDescription` prop to `true`.
- Observe the accessible description is visible below the button.
- Note: I opted to style the description with the same default styling of the Inputbase description. Example screenshots:

Screenshot 2024-06-07 at 14 17 44

Screenshot 2024-06-07 at 14 40 05

- Play with all the other props and button variants to check bad paths and potential breakages.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Enhancements
 
+-   `Button`: Add Button support for iconPosition top and showDescription. ([#62412](https://github.com/WordPress/gutenberg/pull/62412))
 -   `PaletteEdit`: use `Item` internally instead of custom styles ([#66164](https://github.com/WordPress/gutenberg/pull/66164)).
 
 ## 28.10.0 (2024-10-16)
--- a/packages/components/src/button/README.md
+++ b/packages/components/src/button/README.md
@@ -164,7 +164,7 @@ If provided, renders an [Icon](/packages/components/src/icon/README.md) componen
 
 -   Required: No
 
-#### `iconPosition`: `'left' | 'right'`
+#### `iconPosition`: `'left' | 'right' | 'top'`
 
 If provided with `icon`, sets the position of icon relative to the `text`. Available options are `left|right`.
 
--- a/packages/components/src/button/README.md
+++ b/packages/components/src/button/README.md
@@ -140,9 +140,16 @@ An optional additional class name to apply to the rendered button.
 
 #### `description`: `string`
 
-An accessible description for the button.
+A visually hidden accessible description for the button.
 
 -   Required: No
+
+#### `showDescription`: `boolean`
+
+Whether to show the accessible description.
+
+-   Required: No
+-   Default: `false`
 
 #### `disabled`: `boolean`
 
--- a/packages/components/src/button/index.tsx
+++ b/packages/components/src/button/index.tsx
@@ -161,6 +161,7 @@ export function UnforwardedButton(
 		'is-destructive': isDestructive,
 		'has-text': !! icon && ( hasChildren || text ),
 		'has-icon': !! icon,
+		'has-icon-position-top': iconPosition === 'top',
 	} );
 
 	const trulyDisabled = disabled && ! accessibleWhenDisabled;
@@ -224,7 +225,7 @@ export function UnforwardedButton(
 
 	const elementChildren = (
 		<>
-			{ icon && iconPosition === 'left' && (
+			{ icon && ( iconPosition === 'left' || iconPosition === 'top' ) && (
 				<Icon icon={ icon } size={ iconSize } />
 			) }
 			{ text && <>{ text }</> }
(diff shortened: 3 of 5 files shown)


afercia:
Note: while working on this PR I noticed a with icon buttons and the buttons flex layout. By default the SVG icons are flex items and they can shrink. Normally, this isn't a problem. But, when a button shows both the icon and the text _and_ the button width is constrained in some way, the icon shrinks and gets smaller. The width constrained may come, for example, from a container element or from some width set to a custom button.

I don't think icons should ever shrink so I added `flex-shrink: 0` to the icons. Aniamted GIF to illustrate the issue:

buttons icon shrink

ciampo:
I second @mirka 's feedback shared [here](https://github.com/WordPress/gutenberg/issues/62373#issuecomment-2154534317). it seems that looking for a more generic and modular solution like in https://github.com/WordPress/gutenberg/issues/62373#issuecomment-2408177320 could be a better way forward.

More in general, I don't think `Button` needs more props — if anything, it needs fewer props, and consumers should lean more heavily into composing it with other components (tooltip, icon, etc); in other words, use more browser/platform APIs and less abstractions.

I'll let Lena answer directly next week, though.

afercia:
> I second @mirka 's feedback shared [here](https://github.com/WordPress/gutenberg/issues/62373#issuecomment-2154534317). it seems that looking for a more generic and modular solution like in [#62373 (comment)](https://github.com/WordPress/gutenberg/issues/62373#issuecomment-2408177320) could be a better way forward.

I wouldn't object to better ways to make the Button component support more layouts. However, it's not just about styling and layouts.

As [I mentioned on the issue](https://github.com/WordPress/gutenberg/issues/62373#issuecomment-2503240793), there must be a way to make the visually hidden description visible. At the moment, the component assumes the description must always be visually hidden, which is a a very limiting assumption.

Also as mentioned by @draganescu:

> I too fully support buttons with vertical icons, the separation of label and description, design API that removes the need to make divs with button roles and sensible ways to always enable text (labels instead of icons, and visible descriptions). 

While the 'vertical icons' thing falls down to providing styling and layout flexibility by other means, the description is a different matter. Either there should be a prop to make the description visible or maybe it should be a composable sub-component? Any thoughts welcome.

PR #60217, opened 2024-03-26
Fix group block placeholder buttons color contrast ratio and improve consistency

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/60206

## What?

WIP

Issue #60206, opened 2024-03-26
Block placeholder buttons have insufficient color contrast


afercia:
### Description

When adding a Group block, the block placeholder shows 4 buttons to choose a variation:
- Group
- Row
- Stack
- Grid

![Screenshot 2024-03-26 at 10 34 05](https://github.com/WordPress/gutenberg/assets/1682452/d66b5cb2-2088-495f-b9f7-fca809525bbf)

Visually, these four buttons contain only an icon. They don't have borders. As such, the icon must meet the color contrast ratio requirements for text.

The color of the icons is `#cccccc` against a white background color, which results in a contrast ratio of only 1.61:1. 

Worth reminding there are two color contrast requirements:
- Text or images of text: at least 4.51:1. There are exceptions for large text. For normal icon buttons the editor by default uses `#1e1e1e`. The reasons is that icon buttons contain only an icon. As such, the icon acts as the 'text' amd must meet the requirement for text.
- Non-text Contrast: the visual presentation of ueer interface components must have a contrast ratio of at least 3:1. This applies to non-text, for example button borders.

In this specific case the icons are pretty big: 44 by 32 pixels. I'm undecided which contrast requirement these icons should meet and I'd appreciate some thoughts. Cc @joedolson 

Anyways, the current ratio of 1.61:1 is too low.

## Overall design

Looking at the broader picture, I'm not sure the design of the placeholder buttons is ideal in the first place. It's also inconsistent with other blocks. Screenshot to compare with the Columns block:

![Screenshot 2024-03-26 at 10 34 05 2](https://github.com/WordPress/gutenberg/assets/1682452/bb560e9e-4335-44cf-8b74-0e8ba67eaf56)

- The Columns buttons use a blue color for both the icons and borders. Historically, WordPress always used blue for intereactive controls.
- The Columns buttons do have borders.
- The Columns buttons are bigger, which helps targeting the button.
- More importantly, the The Columns buttons have some visually associated text. Unfortunately, the text doesn't match the accessible name, which is only revealed via a tooltip.
- Anyways, the visible text meets the contrast ratio requirement, as well as the blue used for the buttons.

Instead:
- The Group buttons use a too light gray.
- They aren't blue. As such they don't follow the pattern used by WordPress since ages.
- They don't have borders.
- They are smaller.
- They don't have visible text.

Overall, I'd like to propose to make the Group buttons (and other similar cases if any) look like the Columns one.

### Step-by-step reproduction instructions

[...]

PR #59275, opened 2024-02-22
Unify placeholders

jasmussen:
## What?

Update and unify placeholder styles. This makes the Columns block have visually the same setup as Group does. Also updates Image, Video, Audio and other placeholders, emphasizing compactness so that they scale better to narrow contexts.

placeholders latest appearance

<details><summary>Previous iteration of this description</summary>

WIP exploration to explore unifying the Placeholder component. 

Before:

placeholders-before

After:

placeholders-after

</details>

### Testing Instructions for Keyboard

Test blocks that have setup states; group, columns, query loop, image, etc.

PR #62416, opened 2024-06-07
Placeholders: Fix contrast.

jasmussen:
## What?

The icons in the placeholder don't meet 3:1 UI contrast:

before

With this PR, they do:

after

## Testing Instructions

Insert group, columns, or other placeholders. The variation icons should have sufficient contrast.

### DIFF
--- a/packages/block-editor/src/components/block-variation-picker/content.scss
+++ b/packages/block-editor/src/components/block-variation-picker/content.scss
@@ -13,7 +13,7 @@
 	font-size: $helptext-font-size;
 
 	svg {
-		fill: $gray-400 !important;
+		fill: $gray-600 !important;
 	}
 
 	.components-button {


t-hamano:
@afercia I think this PR will fix one of the issues pointed out in [this comment](https://github.com/WordPress/gutenberg/pull/59275#issuecomment-2151977289). I would appreciate it if you could review this PR.

vcanales:
A reminder that Beta 2 is tomorrow, so we will be cherry-picking merged PRs labeled with "Backport to WP Beta/RC" today.

afercia:
Thanks for this PR. Although there's more into play with these 'variations picker' placeholders as detailed in https://github.com/WordPress/gutenberg/issues/60206 and https://github.com/WordPress/gutenberg/pull/60217, I think this PR solves the color contrast issue.

Worth nothing the Group variations buttons don't show any text, which is less than ideal for a11y and it's inconsistent with other blocks. In this case I _think_ we should consider the icons as 'large text' so that in this specific case the `$gray-600: #949494` meets the 3:1 minimum contrast for large text, see [the WCAG 1.4.3 Contrast (Minimum) criterium](https://www.w3.org/TR/WCAG22/#contrast-minimum). @joedolson any thoughts?

Current state in this PR: Group, Columns, Query blocks:

Screenshot 2024-06-10 at 10 54 51

afercia (approved):
LGTM

jasmussen:
Appreciate the review. @vcanales this should be ready to backport.

ellatrix:

PR #62424, opened 2024-06-07
`CustomSelectControlV2`: Fix styling issues

fullofcaffeine:
Split from: https://github.com/WordPress/gutenberg/pull/61272.
Depends on: https://github.com/WordPress/gutenberg/pull/62308.

## What?

Fix styling issues found while testing it via the legacy adapter. See each commit for a description of each fix.

## Why?

We're preparing V2 for shipping, and these visual issues need to be fixed before we replace the V1 implementation with the adapter. This is part of a series of PRs with other fixes/tweaks to the new control (see the top of the description).

## How?

By tweaking the CSS rules for the various elements that make up the component.

## Testing Instructions

* Check all the stories in storybook - for V2 and the adapter. Make sure they look alright;

* Then tests with consumers. Here are a few consumers that can be used to test it in practice:

1. `wp-admin/site-editor.php?canvas=edit`: Add a `date` block and uncheck the "Default format' checkbox in the settings in order for the select to show;
2. In any editor, select a paragraph, and under "Typography" add the "Appearance" select;
3. Add a `spacer` block, and in the "Height" settings, there's a custom select for it as well.
4. Then there's the storybook stories, too.

There's another one under https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/hooks/position.js#L290, but I couldn't find exactly where this feature is yet.

Other slight visual changes might also be made, and the review's goal is to discuss whether it's okay to keep them or if we should fix them (as in making it exactly/closer to V1).

## TODO

- [ ] Rebase from https://github.com/WordPress/gutenberg/pull/62198, fix any conflicts and review issues;
- [ ] Address other styling issues listed [here](https://github.com/WordPress/gutenberg/pull/61272#issuecomment-2136367426) that are still an issue after the rebase.

### DIFF
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -92,7 +92,7 @@ function _CustomSelect(
 	} = props;
 
 	return (
-		<>
+		<Styled.SelectWrapper>
 			{ hideLabelFromVision ? ( // TODO: Replace with BaseControl
 				<VisuallyHidden as="label">{ label }</VisuallyHidden>
 			) : (
@@ -116,7 +116,7 @@ function _CustomSelect(
 					</CustomSelectContext.Provider>
 				</Styled.SelectPopover>
 			</InputBase>
-		</>
+		</Styled.SelectWrapper>
 	);
 }
 
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -14,6 +14,10 @@ import type { CustomSelectButtonSize } from './types';
 
 const ITEM_PADDING = space( 2 );
 
+export const SelectWrapper = styled.div`
+	display: block;
+`;
+
 export const WithHintWrapper = styled.div`
 	display: flex;
 	justify-content: space-between;
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -96,6 +96,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	border-radius: 2px;
 	background: ${ COLORS.theme.background };
 	border: 1px solid ${ COLORS.theme.foreground };
+	z-index: 9999; // Ensure the popover is on top
 
 	&[data-focus-visible] {
 		outline: none; // outline will be on the trigger, rather than the popover
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -104,6 +104,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	position: absolute !important;
 	max-height: 400px;
 	overflow: auto;
+	min-width: 100%;
 
 	&[data-focus-visible] {
 		outline: none; // outline will be on the trigger, rather than the popover
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -117,6 +117,8 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	padding: ${ ITEM_PADDING };
 	font-size: ${ CONFIG.fontSize };
 	line-height: 2.15rem; // TODO: Remove this in default but keep for back-compat in legacy
+	user-select: none;
+
 	&[data-active-item] {
 		background-color: ${ COLORS.theme.gray[ 300 ] };
 	}
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -97,6 +97,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	background: ${ COLORS.theme.background };
 	border: 1px solid ${ COLORS.theme.foreground };
 	z-index: 9999; // Ensure the popover is on top
+	position: absolute !important;
 
 	&[data-focus-visible] {
 		outline: none; // outline will be on the trigger, rather than the popover
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -98,6 +98,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	border: 1px solid ${ COLORS.theme.foreground };
 	z-index: 9999; // Ensure the popover is on top
 	position: absolute !important;
+	max-height: 400px;
 
 	&[data-focus-visible] {
 		outline: none; // outline will be on the trigger, rather than the popover
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -99,6 +99,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	z-index: 9999; // Ensure the popover is on top
 	position: absolute !important;
 	max-height: 400px;
+	overflow: auto;
 
 	&[data-focus-visible] {
 		outline: none; // outline will be on the trigger, rather than the popover
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -22,6 +22,7 @@ export const WithHintWrapper = styled.div`
 	display: flex;
 	justify-content: space-between;
 	flex: 1;
+	flex-wrap: wrap;
 `;
 
 export const SelectedExperimentalHintItem = styled.span`


packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -96,6 +96,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	border-radius: 2px;
 	background: ${ COLORS.theme.background };
 	border: 1px solid ${ COLORS.theme.foreground };
+	z-index: 9999; // Ensure the popover is on top
```
fullofcaffeine, line 104: | Before | After |
|---------|---------|
|  Screenshot 2024-06-07 at 12 07 05 p m |  Screenshot 2024-06-07 at 12 04 22 p m |

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -96,6 +96,8 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
...
+	position: absolute !important;
```
fullofcaffeine, line 105: | Before | After |
|---------|---------|
| Screenshot 2024-06-07 at 12 04 22 p m  | Screenshot 2024-06-07 at 12 13 22 p m |

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -96,6 +96,9 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
...
+	max-height: 400px;
```
fullofcaffeine, line 106: | Before | After |
|---------|---------|
|Screenshot 2024-06-07 at 12 13 22 p m  | Screenshot 2024-06-07 at 12 17 41 p m |

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -99,6 +99,7 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	z-index: 9999; // Ensure the popover is on top
 	position: absolute !important;
 	max-height: 400px;
+	overflow: auto;
```
fullofcaffeine, line 107: This works in synergy with https://github.com/WordPress/gutenberg/pull/62424/commits/a1b9b3e9b18e93c5e24ecb40a7f9e6bbda7f296d (see [#](https://github.com/WordPress/gutenberg/pull/62424#discussion_r1631550848)), and allows the user to scroll up/down when hovering over the popover and the size of items is > the `max-height` of the container.

packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -92,7 +92,7 @@ function _CustomSelect(
 	} = props;
 
 	return (
-		<>
+		<Styled.SelectWrapper>
```
fullofcaffeine, line 95: | Before | After |
|---------|---------|
|  Screenshot 2024-06-07 at 12 30 31 p m |  Screenshot 2024-06-07 at 12 30 51 p m |


See the full changeset here: https://github.com/WordPress/gutenberg/pull/62424/commits/8142ef7c7a5b2e192529c94b9fcbe7787176a20a.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -109,6 +117,8 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	padding: ${ ITEM_PADDING };
 	font-size: ${ CONFIG.fontSize };
 	line-height: 2.15rem; // TODO: Remove this in default but keep for back-compat in legacy
+	user-select: none;
```
fullofcaffeine, line 122: | Before | After |
|---------|---------|
| Screen Recording on 2024-06-07 at 01-45-24 | Screen Recording on 2024-06-07 at 01-46-27 |

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -96,6 +100,11 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	border-radius: 2px;
 	background: ${ COLORS.theme.background };
 	border: 1px solid ${ COLORS.theme.foreground };
+	z-index: 9999; // Ensure the popover is on top
+	position: absolute !important;
+	max-height: 400px;
+	overflow: auto;
+	min-width: 100%;
```
fullofcaffeine, line 108: | Before | After |
|---------|---------|
|  Screenshot 2024-06-07 at 12 56 33 p m | Screenshot 2024-06-07 at 12 55 16 p m |

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -14,10 +14,15 @@ import type { CustomSelectButtonSize } from './types';
...
 
+export const SelectWrapper = styled.div`
+	display: block;
+`;
+
 export const WithHintWrapper = styled.div`
 	display: flex;
 	justify-content: space-between;
 	flex: 1;
+	flex-wrap: wrap;
```
fullofcaffeine, line 25: | Before | After |
|---------|---------|
|  Screenshot 2024-06-07 at 12 49 47 p m|  Screenshot 2024-06-07 at 1 15 14 p m |



This is not evident in the storybook story, but happens when the width is constrained, and the hint is long enough.

ciampo:
I'm going to take over the work on this PR.

I will review this PR and see:
- if the issues that were flagged are still issues
- if the proposed fixes work as expected


I will then take over authoring, either in this PR or in a new one.

packages/components/src/custom-select-control-v2/custom-select.tsx
ciampo: The label's margin seems to work fine when the component is [rendered in isolation ](https://wordpress.github.io/gutenberg/?path=/story/components-customselectcontrol-v2-legacy--default).

After further investigation, the issue in the editor is caused by the fact that `CustomSelectControl` doesn't render a wrapper root element (instead, it renders a `React.Fragment`). The lack of such root DOM element causes `ToolsPanelItem` styles to affect the label.

Replacing `<>` with any DOM element should fix the bug — and since the v1 `CustomSelectControl` uses a `div`, we could simply use a `div` here too.

Additionally, we should think about what's the best `display` value that we should adopt for the `V2` (non-legacy) component: should it be `block` or `inline-block` ?

packages/components/src/custom-select-control-v2/styles.ts
ciampo: I don't personally see this as an issue that needs fixing — neither Ariakit, nor the V1 version of the component disallowed user selection on the individual option items.

packages/components/src/custom-select-control-v2/styles.ts
ciampo: I can't replicate the popover width issue on Storybook — plus, it seems like [we're already passing the `sameWidth` prop to the underlying Ariakit implementation ](https://github.com/WordPress/gutenberg/blob/6fa347111e9714141c0a38711b518d9ef2485485/packages/components/src/custom-select-control-v2/custom-select.tsx#L113)

packages/components/src/custom-select-control-v2/styles.ts
ciampo: While this can be also replicated in the V1 component, I agree that it's something that could be improved and I think that your suggestion makes sense. We can tackle it in a separate issue and update V1 at the same time to keep consistency, in case we think we should address it.

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Since the original `V1` implementation used the pre-determinded `z-index` for popovers, we should at least use the same `z-index` value — here's what we did in `DropdownMenuV2`:

https://github.com/WordPress/gutenberg/blob/c94f863b5df5034524e86490427864779b12754e/packages/components/src/dropdown-menu-v2/styles.ts#L76-L78

packages/components/src/custom-select-control-v2/styles.ts
ciampo: I'm not sure about the `position: absolute` here, since Ariakit should already set it automatically

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Setting the max-height to `400px` is in line with what the V1 does, but we should do it in way suggested by Ariakit, using the `--popover-available-height` CSS variable so that the popover will be even shorter if there isn't available screen estate:

```CSS
max-height: min(var(--popover-available-height, 400px), 400px);
overflow: auto;
overscroll-behavior: contain;
```

packages/components/src/custom-select-control-v2/custom-select.tsx
ciampo: Opened https://github.com/WordPress/gutenberg/pull/62803 to work on this specific fix separately

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Opened https://github.com/WordPress/gutenberg/pull/62821 to work on the select popover fixes separately

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Opened https://github.com/WordPress/gutenberg/pull/62821 to work on the select popover fixes separately

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Opened https://github.com/WordPress/gutenberg/pull/62821 to work on the select popover fixes separately

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Opened https://github.com/WordPress/gutenberg/pull/62848 to tackle this separately

packages/components/src/custom-select-control-v2/styles.ts
ciampo: On second thoughts, I double checked what other libraries do and there seems to be alignment in adding `user-select: none`. I will add it to https://github.com/WordPress/gutenberg/pull/62825

ciampo:
We can close this issue, since all fixes proposed have been iterated on / moved to separate PRs

PR #62428, opened 2024-06-08
List: fix pasting

ellatrix:
## What?

Currently pasting a list or paragraphs in a list doesn't work.

~~I thought we had tests for this, so needs some investigation.~~ Apparently not, so I added two e2e tests.

Probably broken since #54543.

## Why?

When pasting paragraphs in a list, it should convert to list items.

### DIFF
--- a/packages/block-editor/src/components/writing-flow/use-clipboard-handler.js
+++ b/packages/block-editor/src/components/writing-flow/use-clipboard-handler.js
@@ -215,6 +215,11 @@ export default function useClipboardHandler() {
 					if ( canInsertBlockType( block.name, rootClientId ) ) {
 						newBlocks.push( block );
 					} else {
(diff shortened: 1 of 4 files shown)


ntsekouras:
I guess the way to test this is by having raw HTML in the clipboard, right? By doing that I noticed a difference between pasting the same content into the editor in paragraph and in list. The content I tested also had a `br` inside the `p` tag and when inserting in paragraph created one paragraph, in `list` it created two list items. Is this wanted?

ellatrix:
> The content I tested also had a br inside the p tag and when inserting in paragraph created one paragraph, in list it created two list items. Is this wanted?

Yes

ntsekouras (approved):
Let's try this, thanks!

ellatrix:

PR #62429, opened 2024-06-08
Unify DataViews Header Title & Subtitle

t-hamano:
## What?

This PR achieves the following two things in the header of the Patterns page:

#### Before

image

#### After

image

#### Before

image

#### After

image

## Testing Instructions

- Visit the Patterns page
- Check all template parts, all patterns, template part category, and pattern category layouts.


packages/edit-site/src/components/page-patterns/style.scss
```diff
@@ -98,6 +98,11 @@
 
 		.edit-site-patterns__title {
 			min-height: $grid-unit-50;
+
+			.components-heading {
+				flex-grow: 1;
+				flex-basis: 0;
+			}
```
t-hamano: Without this style, the add button would be pushed out.

image

jasmussen:
Good one, thank you for sweating these details. Question, how does the `truncate` property work? In casual superficial testing using the inspector, this isn't working for me:

Screenshot 2024-06-10 at 08 28 38

Is that because truncation is handled differently, or did I do something wrong in my testing?

t-hamano:
@jasmussen Thanks for the testing.

This PR only works on the Patterns page. Because in the Pages and the Templates views, the DataView title is very short and cannot be changed by the user.

jasmussen:
I'm still doing something wrong, because I see the following in the patterns page:
 
2024-06-11 10 34 42 local-wordpress local 7ea38257248f

I appreciate the PR happy to give it a green check. But my main pushback remains: is there a way we can code this so it applies to all headings? The less custom CSS and custom classes we can get away with, the better. It might never be relevant on Page and Template sections, but in future tags and categories sections (#55083) it might again be.

t-hamano:
> I'm still doing something wrong, because I see the following in the patterns page:

Sorry, I was wrong. It seems it wasn't working properly if the text contained spaces. Now it should work for any text.

image

image

Additionally, I've applied the same styles to the Templates and the Pages. To test this, you'll need to change the text in your browser developer tools.

image

image

In the future, we may be able to standardize headers across Patterns, Pages, and Templates to reduce the amount of CSS.

jasmussen (approved):
Thank you for your work!

screenshot of confirmed elided heading

PR #62430, opened 2024-06-08
Classic Block: Display "Convert to blocks" button even if it is inside a Content block

t-hamano:
I discovered this while reviewing #62374

## What?
This PR fixes the issue where the "Convert to blocks" button is not displayed when a Classic block is inside a Content block.

### Before

image

### After

image

## Why?

I think the cause is the same as described in #62374. We need to check the root block to see if the button can be converted, i.e. if the block can be deleted.

## Testing Instructions

- Open the page editor.
- Switch to the code editor and enter just text.
- Switch to visual mode. The text should be automatically converted to a Classic block.
- Enable "Show template" in the sidebar.
- The "Convert to blocks" button should appear in the block toolbar.

### DIFF
--- a/packages/block-library/src/freeform/edit.js
+++ b/packages/block-library/src/freeform/edit.js
@@ -40,7 +40,12 @@ function isTmceEmpty( editor ) {
 export default function FreeformEdit( props ) {
 	const { clientId } = props;
 	const canRemove = useSelect(
-		( select ) => select( blockEditorStore ).canRemoveBlock( clientId ),
+		( select ) => {
+			const { canRemoveBlock, getBlockRootClientId } =
+				select( blockEditorStore );
+			const rootClientId = getBlockRootClientId( clientId );
+			return canRemoveBlock( clientId, rootClientId );
+		},
 		[ clientId ]
 	);
 	const [ isIframed, setIsIframed ] = useState( false );


packages/block-library/src/freeform/edit.js
```diff
@@ -40,7 +40,12 @@ function isTmceEmpty( editor ) {
 export default function FreeformEdit( props ) {
 	const { clientId } = props;
 	const canRemove = useSelect(
-		( select ) => select( blockEditorStore ).canRemoveBlock( clientId ),
+		( select ) => {
+			const { canRemoveBlock, getBlockRootClientId } =
+				select( blockEditorStore );
+			const rootClientId = getBlockRootClientId( clientId );
+			return canRemoveBlock( clientId, rootClientId );
```
ellatrix, line 47: Why doesn't `canRemoveBlock` get the rootClientId automatically?

ellatrix: Commented on the other PR too

Mamaduka (approved):

t-hamano (commented):
There is ongoing discussion in [this comment](https://github.com/WordPress/gutenberg/pull/62374#discussion_r1632571173) about the `canRemoveBlock` spec itself, and I would like to pause merging this PR until we have a clearer direction in that discussion.

PR #62435, opened 2024-06-09
Edit site: WAAPIfy canvas moving animation

stokesman:
## What?
Refactors canvas moving animation so ~~it doesn’t run on the main thread~~ its timing is less impacted by jank.

## Why?
- Performance.
- One fewer dependency.

## How?
Computes the animation keyframes before starting it and passes them to [`animate`](https://developer.mozilla.org/en-US/docs/Web/API/Element/animate).

## More
The idea and the (adapted) implementation are borrowed from Svelte. This generates the keyframes needed for an animation beforehand and that enables `Element`’s `animate` method to handle the rest ~~(off the main thread 🎉)~~. That’s in contrast to the current implementation that on every animation frame computes the keyframe and applies it to the element imperatively.

Update: I’ve learned that because the animation is changing the element’s `width` and `height` the animation can still stutter if the main thread gets busy. However, the duration remains close to what it’s supposed to be unlike the current implementation where the duration gets stretched out. So it’s still a slight improvement for the worst case scenarios.

## Testing Instructions
1. Open the site editor
2. Navigate around and/or enter/exit view and edit modes.
3. Verify the animation feels the same.

### DIFF
--- a/packages/edit-site/src/components/layout/animation.js
+++ b/packages/edit-site/src/components/layout/animation.js
@@ -1,125 +1,206 @@
 /**
- * External dependencies
- */
-import { Controller, easings } from '@react-spring/web';
-
-/**
  * WordPress dependencies
  */
 import { useLayoutEffect, useMemo, useRef } from '@wordpress/element';
 
-function getAbsolutePosition( element ) {
-	return {
-		top: element.offsetTop,
-		left: element.offsetLeft,
-	};
-}
-
 const ANIMATION_DURATION = 400;
 
 /**
- * Hook used to compute the styles required to move a div into a new position.
+ * Animates an element’s placement.
  *
- * The way this animation works is the following:
- *  - It first renders the element as if there was no animation.
- *  - It takes a snapshot of the position of the block to use it
- *    as a destination point for the animation.
- *  - It restores the element to the previous position using a CSS transform
- *  - It uses the "resetAnimation" flag to reset the animation
- *    from the beginning in order to animate to the new destination point.
+ * It works on the FLIP principle:
+ * [First, Last, Invert, Play](https://aerotwist.com/blog/flip-your-animations/)
  *
  * @param {Object} $1                          Options
- * @param {*}      $1.triggerAnimationOnChange Variable used to trigger the animation if it changes.
+ * @param {*}      $1.triggerAnimationOnChange Variable whose changes trigger the animation.
  */
 function useMovingAnimation( { triggerAnimationOnChange } ) {
 	const ref = useRef();
 
-	// Whenever the trigger changes, we need to take a snapshot of the current
-	// position of the block to use it as a destination point for the animation.
-	const { previous, prevRect } = useMemo(
-		() => ( {
-			previous: ref.current && getAbsolutePosition( ref.current ),
-			prevRect: ref.current && ref.current.getBoundingClientRect(),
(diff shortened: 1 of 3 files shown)


stokesman:
🎰 
image
💰

ellatrix:
Are there any libraries around `element.animate`? It does make it a bit harder to understand what's going on, and thus harder to maintain. Btw, there's also a [block moving animation](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/use-moving-animation/index.js) that could be swapped, maybe in a separate PR though.

stokesman:
> Are there any libraries around `element.animate`?

[motion](https://github.com/motiondivision/motionone) and [@okikio/animate](https://github.com/okikio/native/tree/beta/packages/animate). Motion One is quite small and it’d probably be worth a try for this. I'm not sure that it supports custom easing like the animation currently uses. I suppose that may not be a hard requirement though. If not, the implementation here might also be simplified to such a degree that maintenance isn’t a concern.

ellatrix:
I think @youknowriad moved it from Motion to React Spring, but I could be recalling it incorrectly.

stokesman:
> but I could be recalling it incorrectly

I believe you're correct. Just that it was moved from `framer-motion` and `motion` is a different package (same author).

packages/edit-site/src/components/layout/animation.js
```diff
@@ -1,125 +1,207 @@
...
+}
+
+/**
+ * Animates an element, according to the provided configuration.
+ * @param {Element}                  element
+ * @param {AnimationConfig}          config
+ * @param {number}                   target   The target progression — `1` is end, `0` is beginning.
+ * @param {(() => void) | undefined} callback
+ */
+const animate = ( element, config, target = 1, callback ) => {
```
youknowriad, line 106: Is this some kind of generic utility you wrote on your own? Do we expect this to be reused elsewhere (block animation maybe as I just copied that one and adapted it to create this file)

youknowriad:
Are we able to notice perf improvements or less lags or something with this PR?

youknowriad:
In my testing, the trunk animation feels better (less jumpy). For this one for instance I can see the edge of the frame move left than right in some occasions.

packages/edit-site/src/components/layout/animation.js
stokesman: I pared down [Svelte’s internal `animate`](https://github.com/sveltejs/svelte/blob/84ad208f2e986fa549683a12e7f999c3d31b127f/packages/svelte/src/internal/client/dom/elements/transitions.js#L268). It would serve for reuse, the block animation could pass a different config (one that operates only on `transfrom` which should make that one run on the GPU). I’d thought we’d want to try that if this worked well in this instance.

stokesman:
> Are we able to notice perf improvements or less lags or something with this PR?

Ideally there’s no difference in best-case scenarios where both animations aren’t impacted by main thread slowness.

> In my testing, the trunk animation feels better (less jumpy).

I think that’s possible. This way of running the animation keeps the duration close to specified, so if it stutters it’s a bigger stutter because the motion is compensated to keep the duration accurate.

> I can see the edge of the frame move left than right in some occasions.

I didn’t notice that but I’m curious and will try to repro.

Thanks for testing!

PR #60151, opened 2024-03-25
Background image: move controls into a popover

ramonjd:
## What?

This PR:

1. moves the background controls into a flyout panel
2. consolidates properties in the tool panel beneath the "background image" property, which means that resetting an image will reset all associated properties

See: 
- https://github.com/WordPress/gutenberg/pull/57005#issuecomment-1866076523
- https://github.com/WordPress/gutenberg/pull/60100#issuecomment-2178866354

Doing this with a view to consolidate colors and fill. See: https://github.com/WordPress/gutenberg/pull/60100#issuecomment-2160054475

Also take into account:

- https://github.com/WordPress/gutenberg/pull/59454#issuecomment-2020150910
- https://github.com/WordPress/gutenberg/pull/59454#issuecomment-2021680599

### TODO

- [x] Refactor main components in packages/block-editor/src/components/global-styles/background-panel.js
- [x] Condense and consolidate controls in the popover
- [x] Mobile view - ensure it works as expected. 
- [x] Test keyboard interaction
- [ ] ~Wire in tools panel reset for individual properties? E.g., position and repeat?~

## Why?

To win back space in the sidebar controls.

## How?

Moving background image controls into a flyout popover. This includes, where an image exists, the media select control.

To get the Tools Panel controls to work, I've created an invisible `ToolsPanelItem` components for each control set - this registers reset callbacks and so on in the Tools Panel.

## Testing Instructions

The flyout appears in the Global styles (Layout > Background Image) and at the block level for Group, Quote, Post Content blocks. For both:

1. Check that you add/upload/reset background images from the sidebar panel.
3. When a background image is selected, the sidebar control activates the flyout.
4. The controls the flyout should update the background image properties as expected
5. Ensure the Tools Panel controls works, that is, everything should open, close and reset.
6. In mobile view (< 782px) check that the flyout appears in the viewport width. 
7. Resetting the image should also reset size, position, and repeat.

### Testing Instructions for Keyboard

Check:

1. You can tab through the sidebar controls, and add a background image
2. . Where a background image exists, check you can tab to the background image control and hit Enter to open and focus the new flyout.
3. . Tab through the flyout and change the background image properties

## Screenshots or screencast 

| Global styles  | Blocks |
| ------------- | ------------- |
| Screenshot 2024-06-27 at 11 47 29 AM  | Screenshot 2024-06-27 at 11 48 01 AM |

Resetting background properties:

[...]

PR #62438, opened 2024-06-10
Uniform Focal point labels

### LINKED ISSUE #62408: Uniform Focal point labels
Since version 6.5, it is possible to define the focus point of the background image for a Group block.
But the label of this option is "Position", and not "Focal point" as for the Banner block.
As this is the same function (defining the focus point), it would be desirable for the labels of this option to be identical for both blocks.
image
image

As Position can refer to the location of the background image within the Group block (e.g., right, left, center), the Focal point label is much more appropriate.
Please change the "Position" label to "Focal point" for the blocks: Group, Row and Stack.

akasunil:
## What?
Fixes https://github.com/WordPress/gutenberg/issues/62408
 
## Testing Instructions
- Add group block
- Open style setting
- Add background image
- look for focal point control
 
## Screenshots or screencast 
image


t-hamano (commented):
Thanks for the PR!

In my feeling, the term "Focal Point" itself seems to be a bit confusing. The background panel controls four things: `background-image`, `background-position`, `background-repeat`, and `background-size`, but since the `FocalPointPicker` component controls background-**position**, it feels easier to understand as it is now.

Therefore, I would lean towards changing the label "Focal Point" to something that better matches the context.

Currently, the label "Focal Point" is used in the Cover block and Media & Text block. In these two cases, I would prefer to change the label as follows:

### Media & Text Block

`Focal point` to `Media position`

media-text

### Cover Block

`Focal point` to `Image position`

cover

cc @WordPress/gutenberg-design @ramonjd Please let us know your thoughts.

jasmussen:
I could go either way on this one. On the one hand, I appreciate how focal point seems to work well for the visual below with the drag handle. On the other hand, I'm leaning towards position, since top/left position values are shown right below.

jameskoster:
'Position' feels more universally recognisable to me. 

The prefix '**Media** position' is mostly useful in the Media & Text block since 'Position' would be an ambiguous option in the tools panel. But I don't know that we need that prefix elsewhere.

ramonjd:
Thanks for the PR. I agree the labels should be the same, or similar. 

I think "Position" is valid and universal label for instances of the control. I'd argue even terser and more straight forward than "focal point".

It also relates directly to the CSS property it updates (in the case of the cover block `object-position`) - though the tool does give the impression of setting the image origin.

richtabor:
Why not update the background image control to use the existing focal point term? Has there been feedback that "Focal point" is not clear? It's certainly clearer than "Position", in my opinion.

akasunil:
Thank you for the input, everyone. I appreciate the different perspectives on this. I agree that consistency across labels is important for user understanding and ease of use.

"Focal Point" is visually descriptive, it seems to cause some confusion about what it controls. "Position" might simplify user understanding, especially for those who are familiar with CSS properties ( as @ramonjd mentioned ). I agree with @t-hamano suggestions.

let me know your thoughts.

richtabor:
I don't prefer "Position" for these two reasons:

1. Position is used elsewhere in the UI for setting CSS position, like absolute positioning. 

2. Position is less clear than focal point. The ideal behind "focal point" is that you're deciding where the focus is on this image. The "Position" label is indicative of you deciding the position of the focal point. Why abstract further from the actual intent?

t-hamano:
I noticed that the background image UI is a popover in the latest Gutenberg. In this case, which is better, "Position" or "Focal point"?

image

ramonjd:
> I noticed that the background image UI is a popover in the latest Gutenberg. In this case, which is better, "Position" or "Focal point"?

I think it should be consistent, whether one or the other. As this PR stands, changing it to Focal point is 👍🏻 

Just needs a rebase after https://github.com/WordPress/gutenberg/pull/60151 as mentioned [above](https://github.com/WordPress/gutenberg/pull/62438#issuecomment-2210535297)

t-hamano (approved):
Ok, let's ship this PR 👍

PR #62443, opened 2024-06-10
Add `context` in post actions API

ntsekouras:
## What?

Related: https://github.com/WordPress/gutenberg/pull/62323 - in the aspect of not showing the `revisions` action in `edit` mode.

This PR proposes the addition of a new `context` prop in post actions API. Post actions in their majority are meant to be used in any context like in data views (lists) or in editors as part of the ellipsis menu in inspector controls. There are some exceptions though and a couple examples are:
1. In editors we want to give more prominence to the revisions actions and is currently also shown as part of the post summary. For this reason we want to avoid the extraneous action in the ellipsis menu (related linked PR above)
2. In lists of entities (currently only in site editor) we have the `editPost` action which doesn't make sense in the editors, as we already editing. For that reason this actions lived in edit-site package and was separated from all the other actions that live in `editor` package (for now).

My suggestion for the new `context` property values is  a separation between `list` and whatever view/layout might use them in the future...  Due to the unification of editors I don't think it makes sense to consider `edit-post|edit-site` etc.. and also experimented with `edit/view` context but the lines might blur when we have `list` views with editing allowed. It's a private API, so we can always revisit.

A consumer of actions in order to get them from where ever (right now with `usePostActions` hook) can provide a specific context if needs to, but the absence of it means fetch all (default). Additionally actions can provide a context where they are allowed and is separate from `isEligible`, as it's meant for checks regarding a specific item. 

I'd welcome any feedback and suggestions about the API.

## Testing Instructions
1. Everything should work as before, including the `edit post` action
2. View revisions action is only available in site editor lists

### DIFF
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -354,8 +354,14 @@ export default function DataviewsPatterns() {
 		return filterSortAndPaginate( patterns, viewWithoutFilters, fields );
 	}, [ patterns, view, fields, type ] );
 
-	const templatePartActions = usePostActions( TEMPLATE_PART_POST_TYPE );
-	const patternActions = usePostActions( PATTERN_TYPES.user );
+	const templatePartActions = usePostActions( {
+		postType: TEMPLATE_PART_POST_TYPE,
+		context: 'list',
+	} );
+	const patternActions = usePostActions( {
+		postType: PATTERN_TYPES.user,
+		context: 'list',
+	} );
 	const editAction = useEditPostAction();
 
 	const actions = useMemo( () => {
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -583,6 +583,7 @@ const viewPostAction = {
 
 const postRevisionsAction = {
 	id: 'view-post-revisions',
+	context: 'list',
 	label( items ) {
 		const revisionsCount =
 			items[ 0 ]._links?.[ 'version-history' ]?.[ 0 ]?.count ?? 0;
@@ -1029,7 +1030,7 @@ export const duplicateTemplatePartAction = {
 	},
 };
 
-export function usePostActions( postType, onActionPerformed ) {
+export function usePostActions( { postType, onActionPerformed, context } ) {
 	const { defaultActions, postTypeObject } = useSelect(
 		( select ) => {
 			const { getPostType } = select( coreStore );
@@ -1057,7 +1058,7 @@ export function usePostActions( postType, onActionPerformed ) {
 			return [];
 		}
 
-		const actions = [
+		let actions = [
 			postTypeObject?.viewable && viewPostAction,
 			supportsRevisions && postRevisionsAction,
 			globalThis.IS_GUTENBERG_PLUGIN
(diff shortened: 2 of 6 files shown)


packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -354,16 +353,21 @@ export default function DataviewsPatterns() {
...
+	} );
+	const patternActions = usePostActions( {
+		postType: PATTERN_TYPES.user,
+		context: 'view',
+	} );
 
 	const actions = useMemo( () => {
 		if ( type === TEMPLATE_PART_POST_TYPE ) {
-			return [ editAction, ...templatePartActions ].filter( Boolean );
+			return templatePartActions.filter( Boolean );
```
mcsf: Is filtering still doing anything useful? I'd be surprised if `usePostActions` were returning empty values.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -612,6 +615,36 @@ const postRevisionsAction = {
...
+			},
+			callback( items ) {
+				const post = items[ 0 ];
+				history?.push( {
+					postId: post.id,
+					postType: post.type,
+					canvas: 'edit',
+				} );
+			},
+			context: 'view',
```
mcsf: Nit: place `context` right after `id` for prominence.

mcsf (commented):
Quick thoughts:

* I think introducing `context` makes sense
* But I worry about overloading WordPress's preexisting `edit|view` dichotomy with something that is close but not quite like it
* I wouldn't dismiss `context: 'list'` too quickly: there may be something worthwhile there

packages/editor/src/components/post-actions/actions.js
```diff
@@ -612,6 +615,36 @@ const postRevisionsAction = {
 			onActionPerformed( posts );
 		}
 	},
+	context: 'view',
+};
+
+export const useEditPostAction = () => {
```
youknowriad: In my mental model, the "editor" package (later probably core-data or a dedicated core-dataviews package) is about fields and actions that can work in any place.

This action has a direct dependency towards the site editor routes, so I believe it shouldn't be defined in the editor package and would prefer if it stays in the site editor (where the route is defined)

ntsekouras: Sounds good, yes. I wasn't too fond of doing that either.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -583,6 +583,7 @@ const viewPostAction = {
 
 const postRevisionsAction = {
 	id: 'view-post-revisions',
+	context: 'list',
```
jorgefilipecosta, line 586: I know we only support one context right now, but I expect this will grow and we may endup with more contexts. Would it make sense for context to be array with all the supported contexts?

ntsekouras: Nothing is public right now, so I think we can add it when we'll need it.

jorgefilipecosta (approved):
I think we can merge this solution it seems very likely we will need a context for the actions, and we can iterate in the future how that context looks like.

ntsekouras:
This PR was [cherry picked ](https://github.com/WordPress/gutenberg/pull/62443) for 6.6.

Issue #62441, opened 2024-06-10
Inserter: Tab-focus diverges from other tab-bars


jasmussen:
### Description

If you load the site editor or post editor fresh, then open the inserter, the focus-visible ring is shown on the "Blocks" tab. This is different from similar panels, like the list view, block inspector, document inspector, where this focus-ring is only visible if you used keyboard interactions to open the panel. Shown in this GIF:

![tab focus](https://github.com/WordPress/gutenberg/assets/1204802/7f7891ab-51cf-4b4c-a1a3-dda82ab83697)

The behavior across tab-bars should be consistent. Specifically, the focus-visible property added to the tab-bar should continue to behave as intended, with the ring visible when using the keyboard, and not when clicking.

### Step-by-step reproduction instructions

1. Load one of the editors fresh
2. Click to open the inserter and notice the focus-ring
3. Compare to other panels, like the block inspector.

### Environment info

Nightly WordPress, trunk Gutenberg.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

scruffian:
The reason for having the focus indicator is to show a user where the focus is once the panel opens. In the other panels the focus remains on the button that opened the panel, rather than moving into the panel. If we are going to move the focus then we need a way to show the user where focus has moved to. The only exception to this might be if we could determine that a user has opened the panel using a mouse rather than the keyboard - in that case we could consider moving focus and showing the focus ring, only if the user has opened the panel using the keyboard.

An additional challenge is that it's difficult to tell the browser to focus and element but not focus-visible it. Firefox supports a focusVisible option for focus, but none of the other browsers have implemented it yet.

jasmussen:
This was fixed!

PR #62444, opened 2024-06-10
Inserter: Don't focus the inserter when opening

scruffian:
## What?
This is to address: https://github.com/WordPress/gutenberg/issues/62441

## Why?
In other sidebar panels we don't automatically focus the panel, we only do it when a user interacts with the panel.

## How?
Removes the code that autofocusses the inserter.

## Testing Instructions
1. Reload the site editor
2. Open the inserter
3. Confirm that focus remains on the toggle inserter button

### DIFF
--- a/packages/block-editor/src/components/inserter/menu.js
+++ b/packages/block-editor/src/components/inserter/menu.js
@@ -12,7 +12,6 @@ import {
 	useCallback,
 	useMemo,
 	useRef,
-	useLayoutEffect,
 } from '@wordpress/element';
 import { VisuallyHidden, SearchControl, Popover } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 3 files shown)


jeryj:
This is tricky. The ideal interaction is to not move focus when opening the panel and have the panel be the next tab stop. In our case, the panel is many tab stops from the button that opens it, so we move focus. 

To get consistency between panels, the other panels _should_ autofocus the panel they open unless they are the next tab stop.

jasmussen:
Want to clarify that #62441 is not about handling focus, it's about the fact the focus-visible which is applied to the tab-bar component is, in this case, ignored. I'll update the issue to clarify that bit.

scruffian:
Closing in favour of https://github.com/WordPress/gutenberg/pull/62513

PR #62448, opened 2024-06-10
DataViews: Fix unnecessary horizontal scrollbar in list layout

t-hamano:
## What?

This PR fixes the horizontal scrollbar that occurs when the DataViews is in the list layout.

Note: The screenshot below is from Windows OS. On MacOS, you can reproduce this issue by changing the scrollbar setting to always visible.

dataviews

## Why?

I think the problem is probably due to the fact that the visually invisible button sticks out from the content when the item is not focused or selected.

image

## How?

There are probably a few different approaches we could consider, but defining `right:0` prevents the button from overflowing the parent element and solves the problem.

## Testing Instructions

- If you are testing on MacOS, change the scrollbar settings to always show.
- In the Templates or the Pages screen, change the data view layout to List.

### DIFF
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -436,6 +436,7 @@
 			.components-button {
 				opacity: 0;
 				position: fixed;
+				right: 0;
 			}
 		}
 


stokesman (approved):
Nice. This one has been bugging me for a bit. Here’s trunk on macOS:
image

And here with this PR:
image

The style change seems an apt fix to me. 🚀

t-hamano:
@stokesman Thanks for the review!

Note: Since this PR fixes a list layout issue that first shipped in WP6.6, I'm adding the backport label.

ellatrix:

PR #62450, opened 2024-06-10
Fix: Add `network-active` to valid options in `PluginStatus` Type definition

fabiankaegy:
## What?

Adds `network-active` to valid options in `PluginStatus` type definition

## Why?

It is a valid option that the Rest API returns: https://github.com/WordPress/wordpress-develop/blob/bca94a5fe45f9414f0d11a16372895a4817b1716/src/wp-includes/rest-api/endpoints/class-wp-rest-plugins-controller.php#L61

## How?

Adding the string to the type

### DIFF
--- a/packages/core-data/src/entity-types/plugin.ts
+++ b/packages/core-data/src/entity-types/plugin.ts
@@ -73,7 +73,7 @@ declare module './base-entity-records' {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Mamaduka (approved):
Good catch, @fabiankaegy!

PR #62459, opened 2024-06-11
Use wordmarks dangit (GitHub, JavaScript)

peterwilsoncc:
## What?

Replaces Github with GitHub in various comments
Replaces Javascript  with JavaScript in various comments

WP-Dev ticket: WP#[60699](https://core.trac.wordpress.org/ticket/60699) (6.6 docs improvements)
WP-Dev PR: https://github.com/WordPress/wordpress-develop/pull/6785

## Why?

If Wordpress and by extension GutenBerg is going to worry about word-marks, it should do so consistently

## How?

Manual review after a search. 

No code  changes as I didn't want to break anything that's case sensitive.

## Testing Instructions

Please review the PR and make sure I haven't changed the name of any symbols.


fabiankaegy (approved):
Good catch 👍

juanmaguitar:
It would be great if we could add some automation to these spell-checking tasks.
Here's what seems to be a good approach: https://seankilleen.com/2021/01/adding-spell-checking-to-my-blogs-build-process-with-github-actions-and-cspell/

dmsnell:
> If Wordpress and by exte…

🤔

lib/compat/wordpress-6.5/html-api/class-gutenberg-html-tag-processor-6-5.php
```diff
@@ -298,8 +298,8 @@
  *
  * The special elements are:
  *  - `SCRIPT` whose contents are treated as raw plaintext but supports a legacy
- *    style of including Javascript inside of HTML comments to avoid accidentally
- *    closing the SCRIPT from inside a Javascript string. E.g. `console.log( '</script>' )`.
+ *    style of including JavaScript inside of HTML comments to avoid accidentally
+ *    closing the SCRIPT from inside a JavaScript string. E.g. `console.log( '</script>' )`.
```
dmsnell, line 302: This is backported code, and will disappear once WordPress 6.7 is shipped. It may not be worth introducing the divergence from Core and these "built assets." In other words, this is akin to modifying the contents of dependencies inside `node_modules`. Just wanted to point this out, as it has and can cause confusion when the code is different than in Core for the given version, beyond the `Gutenberg_` prefix naming.

peterwilsoncc: @dmsnell I never know which order I should update compat code. I've pushed the equivalent branch to WP-Dev with the goal of getting it in to WP 6.6 to update the developer docs on wp.org. See https://github.com/WordPress/wordpress-develop/pull/6785

peterwilsoncc:
As I didn't change any symbols to avoid the risk of introducing bugs, I've change the type of this to developer docs.

As many of the typos fixed in this ticket are reflected on developer.wordpress.org, I'll mark it or backport to the 6.6 branch. cc @ellatrix

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 838f409f0b

Issue #60842, opened 2024-04-18
Improvements to `class WP_Theme_JSON` and its tests


tellthemachines:
## What problem does this address?

Currently, any change made to styles generated by `class WP_Theme_JSON` cause multiple tests from its test suite to fail, mostly due to diffs in CSS strings being checked for equality.

Adjustments to CSS output, whether properties or rule specificity, are very common, and test strings are incredibly painful to update when they inevitably fail.

The biggest culprits are the 30 or so tests that check the output of `get_stylesheet`. I think the main reason there are so many tests focusing on that function is that it brings together the output of a bunch of protected functions such as `get_layout_styles`, `get_block_classes` and `get_css_variables`, among others, and it's being used essentially to test the output of those functions.

## What is your proposed solution?

How can this be improved?

I don't love the idea of tests comparing CSS strings, because the outcome of the test relies too much on implementation details (the exact shape of the CSS string, which the `get_stylesheet` function is not even responsible for). A possible alternative would be to check for the presence of one or another property or classname inside the string, instead of trying to match the whole string.

In general, unit tests are useful to check that the behaviour of a function remains unchanged after an edit or refactor. If we're testing an exact string of CSS, then _all_ the changes we make to these functions will alter the output, because our styles are continually evolving. So what's the point of these unit tests? They're never catching any actual bugs. If we can abstract what we're checking for a bit - say we check that the function outputs a valid CSS string, but not its exact shape - maybe the tests become more useful while not being such a maintenance burden.

I'd love to hear about other potential approaches here!

Other things we can do to mitigate the impact of these tests on developer experience:

* Test output of functions called by `get_stylesheet` independently.
* Use a minimal set of properties for each test, so the test strings are smaller and easier to update when they inevitably change.

Further improvements for `class WP_Theme_JSON` itself:

* Look into removing the custom CSS-specific functions in favour of using the general global styles ones, as per [this comment](https://github.com/WordPress/gutenberg/pull/58991#discussion_r1507927224).
* There's a `get_svg_filters` function that isn't being used anywhere in either core or gutenberg, should we deprecate it? A quick check in https://www.wpdirectory.net/ doesn't show any use either.

ajlende:
> There's a `get_svg_filters` function that isn't being used anywhere in either core or gutenberg, should we deprecate it?

[...]

PR #46496, opened 2022-12-13
Block.json: Refactor and stabilize selectors API

aaronrobertshaw:
Related:
- https://github.com/WordPress/gutenberg/issues/45194
- https://github.com/WordPress/gutenberg/pull/46579

## What?

Refactors the block.json selectors API.

- Moves the `__experimentalSelector` props within `supports` into their own `selectors` config.
- The block's primary `__experiementalSelector` prop is merged into the new `selectors` config via the `root` prop.
- The selectors API is stabilized as part of this refactor
- Adds greater flexibility allowing individual styles within block support features to be applied to arbitrary selectors
- Allows fallback to appropriate "root" selectors if the individual style isn't set in the selectors config
- Allows entire block support feature to have a single selector applied via a simple string value

## Why?

The experimental selectors have been around for a while now and are beginning to be relied upon more. There's also an increasing need for greater flexibility in setting selectors for individual block support styles.

For more info and history see: https://github.com/WordPress/gutenberg/issues/45194

#### https://github.com/WordPress/gutenberg/pull/46496/commits/d43c09913e02268a25b2918ca7ddbb41de9216e5

Updates Gutenberg’s block registration to polyfill the selectors prop until it is included in the server-provided definitions. Temporary unit test added for the polyfill.

#### https://github.com/WordPress/gutenberg/pull/46496/commits/43257babd9526a91fa2594a41e64b5d148077082

Adds theme.json unit tests for stylesheet generation using blocks leveraging the new selectors API. The old feature selectors test is maintained and could be deprecated as blocks migrate to the stabilised API.

Due to the updated block metadata, a test block needs to be registered during the phpunit bootstrap so that it is present before WordPress runs and the theme.json block metadata is cached.  
  
A follow-up is currently looking at if we can invalidate the metadata cache and move the test block registrations to their respective tests.

#### https://github.com/WordPress/gutenberg/pull/46496/commits/9d073cf07e412b70014ad41151dcdfe2557591c6

Adds the `selectors` to the block.json schema. The available shape for the selectors API almost matches that of the block supports.

An important difference is that it will accept a `root` selectors which replace the old `__experimentalSelector` for the block as a whole as well as act as generic selector for each feature e.g. color, typography etc. 

#### https://github.com/WordPress/gutenberg/pull/46496/commits/b28fd4a8779bc2579cb9e2823411569cd7a102ac

Updates the Image block.json to use the new selectors API instead of `__experimentalBorder.__experimentalSelector`.

[...]

PR #46486, opened 2022-12-13
Global Styles: Fix Reset to defaults action by moving fills to be within context provider

andrewserong:
## What?

Fixes: https://github.com/WordPress/gutenberg/issues/46484

Moves `<SidebarComplementaryAreaFills />` which contains the global styles sidebar to be within the global styles provider.

Kudos: @ramonjd for the idea!

## Why?

Without being inside the provider, the `Reset to defaults` action does not work.

## How?

* Move `<SidebarComplementaryAreaFills />` to be within the global styles provider.

## Testing Instructions

* Follow the issue over in #46484 — without this PR, it isn't possible to Reset to defaults within global styles
* With this PR applied, you should be able to go to global styles > ellipsis menu > Reset to defaults to reset back to theme defaults
* Double check that otherwise, global styles (and the site editor) are working correctly

## Screenshots or screencast 

| Before | After |
| --- | --- |
|  |  |

PR #62463, opened 2024-06-11
Improve maintainability of theme json class tests.

tellthemachines:
## What?

Partially addresses #60842.

This PR syncs https://github.com/WordPress/wordpress-develop/pull/6734 to Gutenberg. I also adjusted a couple tests that don't exist in core yet to not output base layout styles.

## Testing Instructions

1. Run the PHP tests and check that everything passes.


phpunit/class-wp-theme-json-test.php
```diff
@@ -1099,12 +1044,12 @@ public function test_get_stylesheet_with_deprecated_feature_level_selectors() {
 			)
 		);
```
tellthemachines, line 1030: If I'm reading the comments above this test and the next one correctly, we should be able to update them to use core blocks now?

aaronrobertshaw: That's my understanding. We should be able to update these tests to core blocks and remove the filter registering the test block from the phpunit bootstrap.

I can look a little deeper on this tomorrow if that helps?

tellthemachines: Thanks for confirming, I'll give it a go and see if it works!

tellthemachines: Does the "deprecated" in the test name refer to defining selectors with `__experimentalSelector` inside the `supports` array (like Calendar or Button do) as opposed to outside (like Image does)?

andrewserong: I believe so, it looks like this test was renamed from `test_get_stylesheet_with_block_support_feature_level_selectors` in https://github.com/WordPress/gutenberg/pull/46496/ which stabilised that feature and moved it outside of `supports`.

aaronrobertshaw: The selectors API feature got merged by a good samaritan before it was actually ready so following threads like #46486 will be tricky as the individual PRs only contain part of the "story".

To answer the question, yes, the deprecated `__experimentalSelector` properties are what the test was about as we have to maintain backwards compatibility for them unfortunately.

tellthemachines: Thanks folks, all done! I trimmed the properties output to match what each block supports, and that also makes the strings a more manageable size.

phpunit/class-wp-theme-json-test.php
```diff
@@ -4801,36 +4689,17 @@ public function test_get_top_level_background_image_styles() {
...
-		$this->assertSame( $expected_styles, $theme_json->get_stylesheet(), 'Styles returned from "::get_stylesheet()" with top-level background styles type does not match expectations' );
+		$body_node = array(
+			'path'     => array( 'styles' ),
+			'selector' => 'body',
+		);
+
+		$expected_styles = "html{min-height: calc(100% - var(--wp-admin--admin-bar--height, 0px));}:root :where(body){background-image: url('http://example.org/image.png');background-position: center center;background-repeat: no-repeat;background-size: contain;}";
+		$this->assertSame( $expected_styles, $theme_json->get_styles_for_block( $body_node ), 'Styles returned from "::get_stylesheet()" with top-level background styles type does not match expectations' );
```
andrewserong:
```suggestion
		$this->assertSame( $expected_styles, $theme_json->get_styles_for_block( $body_node ), 'Styles returned from "::get_styles_for_block()" with top-level background styles type does not match expectations' );
```

phpunit/class-wp-theme-json-test.php
```diff
@@ -4846,8 +4715,8 @@ public function test_get_top_level_background_image_styles() {
...
-		$this->assertSame( $expected_styles, $theme_json->get_stylesheet(), 'Styles returned from "::get_stylesheet()" with top-level background image as string type does not match expectations' );
+		$expected_styles = "html{min-height: calc(100% - var(--wp-admin--admin-bar--height, 0px));}:root :where(body){background-image: url('http://example.org/image.png');background-position: center center;background-repeat: no-repeat;background-size: contain;}";
+		$this->assertSame( $expected_styles, $theme_json->get_styles_for_block( $body_node ), 'Styles returned from "::get_stylesheet()" with top-level background image as string type does not match expectations' );
```
andrewserong:
```suggestion
		$this->assertSame( $expected_styles, $theme_json->get_styles_for_block( $body_node ), 'Styles returned from "::get_styles_for_block()" with top-level background image as string type does not match expectations' );
```

andrewserong (approved):
Looking good — very cool that real core blocks could be used for both cases of the deprecated and proper selectors, too, that helps simplify things a fair bit.

Just left a couple of tiny nit comments in the strings that are output on test failures, but other than that this LGTM! 🚀

phpunit/class-wp-theme-json-test.php
tellthemachines: Oh, well spotted, thanks!

PR #62469, opened 2024-06-11
Global Styles: Only use single property variations as color/type presets

scruffian:
## What?
In this PR we change the behaviour of color and typography presets so that we use only variations that contain nothing but that specific property. 

## Why?
This reduces the impact of the change to introduce color and typography presets as it means that only variations that target a specific property are used.

## How?
Filter out all variations unless they contain only a specific variation.

## Testing Instructions
1. Using TT4
2. Open the site editor
3. Open Global Styles
4. Open Colors
5. Open Palette
6. Notice that you now only see two presets

## Screenshots or screencast 
Before:
Screenshot 2024-06-11 at 09 52 49

After:
Screenshot 2024-06-11 at 09 52 20


juanmaguitar:
> Notice that you now only see two presets

@scruffian, according to the specification...

> color and typography presets so that we use only variations that contain nothing but that specific property.

...shouldn't there be just one preset corresponding to the only color-only variation in TT4 (Onyx)? At least this is the behavior I would expect

scruffian:
> ...shouldn't there be just one preset corresponding to the only color-only variation in TT4 (Onyx)? At least this is the behavior I would expect

The other one is the default, so you can switch back to it.

juanmaguitar:
> The other one is the default, so you can switch back to it.

That was my guess, yes. I understand the need to reset to the default colors. However, I find this mechanism 'not immediately easy to grasp' in its current form. Are there other approaches (from a UI perspective) to 'un-apply' the preset that could be used here?

I think this whole feature would be a bit easier to understand if there was only as many presets available as the ones generated from color-only variations (or typography-only variations). 

But maybe this is just me. I'll let others chime in and share their feedback. 
cc: @ndiego @justintadlock

ndiego:
I think this approach is fine. It just looks strange in TT4 due to the way the style variations were originally built. Ideally, TT4 would have more dedicated color-only style variations to demonstrate this feature, or the Onyx style variation could be updated with an additional superfluous setting to disable the color-only palette generation.

scruffian:
cc @jasmussen for a design perspective

richtabor:
I think it's fine that the first one is the default. 

I'd rather update TT4 with more colors than opt the theme out of this completely. It's a symptom of how many theme variations are built (to be color only).

ndiego:
> I'd rather update TT4 with more colors than opt the theme out of this completely.

It would be a nice marketing highlight as well. "With WordPress 6.6, the Twenty Twenty-Four theme has been updated to include X new color palettes..."

colorful-tones:
I tested this with the Frost theme, and it works
Screenshot 2024-06-11 at 11 30 52 AM
 smoothly.

jasmussen (approved):

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -51,14 +51,12 @@ export function removePropertyFromObject( object, property ) {
  * A convenience wrapper for `useThemeStyleVariationsByProperty()` that fetches the current theme style variations,
```
ajlende: Since `useThemeStyleVariationsByProperty` is deleted, this comment and the file names should probably be updated.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -71,27 +69,26 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
 		};
 	}, [] );
 	const { user: baseVariation } = useContext( GlobalStylesContext );
+	const clondedBaseVariation = cloneDeep( baseVariation );
```
ajlende: Typo.

```suggestion
	const clonedBaseVariation = cloneDeep( baseVariation );
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -71,27 +69,26 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
 
-	const variations = useMemo( () => {
+	const variationsWithSinglePropertyAndBase = variationsFromTheme
+		.filter( ( variation ) => {
+			return isVariationWithSingleProperty( variation, property );
+		} )
+		.map( ( variation ) => {
+			return mergeBaseAndUserConfigs( clondedBaseVariation, variation );
+		} );
```
ajlende: Move this inside the `useMemo`. As it is now, the `useMemo` is just pure overhead because this is creating a new array every time.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
ajlende: Actually, I don't think there's any reason to do a deep merge here. `mergeBaseAndUserConfigs` uses [`deepmerge`](https://www.npmjs.com/package/deepmerge#mergex-y-options) which doesn't mutate the original object.

scruffian:
I found a bug in this - the default variation doesn't contain any other user settings, so when I select it, everything is reset, not just the colors.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
scruffian: Done!

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
scruffian: Since I fixed a bug in [ba440bd](https://github.com/WordPress/gutenberg/pull/62469/commits/ba440bde6baac3e35960239be7f4522c2e3921fc), I think we do need the clone again..

scruffian:
@ajlende thanks for the review, I think I fixed all the issues :)

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -72,26 +70,26 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
+		const variationsWithSinglePropertyAndBase = variationsFromTheme
+			.filter( ( variation ) => {
+				return isVariationWithSingleProperty( variation, property );
+			} )
+			.map( ( variation ) => {
+				return mergeBaseAndUserConfigs( baseVariation, variation );
+			} );
+
+		return [ userVariation, ...variationsWithSinglePropertyAndBase ];
+	}, [ baseVariation, variationsFromTheme ] );
```
ajlende:
```suggestion
	}, [ property, baseVariation, variationsFromTheme ] );
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -72,26 +70,26 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
+	return useMemo( () => {
+		const clonedBaseVariation = cloneDeep( baseVariation );
 
-	return useThemeStyleVariationsByProperty( {
-		variations,
-		property,
-		filter,
-		baseVariation: removePropertyFromObject(
-			cloneDeep( baseVariation ),
+		// Get user variation and remove the color settings.
```
ajlende:
```suggestion
		// Get user variation and remove the settings for the given property.
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -72,26 +70,26 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
-	} );
+		);
+		userVariation.title = __( 'Default' );
+
+		const variationsWithSinglePropertyAndBase = variationsFromTheme
+			.filter( ( variation ) => {
+				return isVariationWithSingleProperty( variation, property );
+			} )
+			.map( ( variation ) => {
+				return mergeBaseAndUserConfigs( baseVariation, variation );
```
ajlende: Should this be using the `clonedBaseVariation`? If `removePropertyFromObject` mutates the argument, then the [old code below](https://github.com/WordPress/gutenberg/pull/62469/files#diff-0ee3f0040b18ec7b1a93b7d799f26d84216f6b6a513a9a7337f1cc66c6d987ccL184) is merging the one without the property.

scruffian: Yes you're right, great catch. We want to apply only the typography styles from the variation, not include the ones from the base. I updated the PR and renamed a few variables too for clarity.

ajlende (approved):
LGTM 👍

ellatrix:
Is this 6.6 related? Should it be backported?

ndiego:
> Is this 6.6 related? Should it be backported?

Yes, I believe it should be. Can you confirm @richtabor and @scruffian?

scruffian:
Yes sorry it should be

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: dbd9830690

PR #60715, opened 2024-04-12
More consistent root padding

richtabor:
Closes https://github.com/WordPress/gutenberg/issues/60308.

I'm using [this gist](https://gist.github.com/richtabor/90df9a709815302fc1b322c6c68f3dc9) as a basis for testing. The content should have the same layout behavior throughout WordPress.

### Why

Patterns today, [even in Twenty Twenty Four](https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/cta-rsvp.php#L11), have to hard-code left and right padding values into theme patterns, so that patterns render properly with left and right spacing anywhere they were used. 

People creating patterns on their sites are required to do the same, even though it's not clear. In some instances when creating a pattern the outer padding is expressed as expected, but when you view that pattern in a focus view, it's not.

Here's why that's not ideal:

1. **There are current inconsistencies with when, and where, global padding gets applied**, including targeting direct blocks. These can be simplified, resolving the "missing" outer padding where expected, without changing the end result of a site (much, if any). For example, patterns and template parts in the Site Editor don't show outer padding, but they're rendered within templates and on the front-end. 
2. **The Global Styles > Layout > Padding control for assigning left and right outer padding across the site is no longer connected to any patterns inserted on the site, now —or ever.** If a user wants to change the site padding, they have to change it on each individual pattern inserted on the site or fork and modify all patterns with a different value. The only exception is if the theme designer manually places the outer padding variable within each pattern, which seems like an unnecessary burden that we can alleviate—and even then, we can't assume users will know how to do this when they create their own patterns. 
3. **A pattern I insert from one theme, like Twenty Twenty Four, will always have the hard-coded pattern values associated with it.** It will almost never look like it can belong placed on a page with another theme active, if that second theme uses any different value for padding (which is highly likely). It's fine if a pattern explicitly opts to include padding (opting out of global padding, just as if it opts out of a default font size), but out-of-the-box, outer padding should be applied by default where expected.  
4. The same as 3, but for patterns in the Patterns Directory. **If we can build patterns without having to scope the left and right outer padding values, then each pattern becomes much more viable across the design library of WordPress** itself.

### Testing Instructions

1. Open the Site Editor.
2. Select the header template part's inner group block. 
3. Remove any padding left/right from the group within. 
4. See global padding is applied.
5. Select the header template part.
6. Select "Edit" to view the header in a focus mode.
7. See global padding is still applied. 
8. Select the global styles icon in the editor header to open Global Styles. 
9. Select the "Layout" menu item.
[...]

PR #62470, opened 2024-06-11
fix: multiple :not() classes

chrisbellboy:
See #60715 

## What?
Fixes multiple `:not()` classes

## Why?
The CSS is currently not rendering due to invalid `:not()` implementation.

## How?
By chaining the multiple `:not()` separately.


chrisbellboy:
Closing as TIL that :not() does [allow multiple classes](https://www.w3.org/TR/selectors-4/#negation) in a single selector since CSS Selectors Level 4 🤓

Issue #62287, opened 2024-06-04
Pattern overrides: Image captions appear to not be handled correctly


afercia:
### Description

In a joined testing session with @carolinan and @SergeyBiryukov we wanted to get familiar with and test the Pattern overrides feature unblocked in https://github.com/WordPress/gutenberg/pull/62011 and planned to be released in WordPress 6.6.

Our testing was limited to the Image and Paragraph blocks in a pattern so it was a very partial testing with only _two_ blocks. We noticed a few unexpected behaviors especially with the Image block. We are concerned _there may be more_ unexpected behaviors with other blocks at the point we shared the feeling this feature appears to not be fully ready to be released.

I'd like to point out we're not very familiar with the implementation and we tested this feature as 'advanced users'. Users expectations are important though and the general perception we got is that it doesn't feel polished enough to be 'ready'.

Personally, I would argue that content that disappears in an unexpected way or that is not replaced as users would expect doesn't contribute to perceive this feature as reliable and doesn't help users feel confident enough to use it.

Teh reproduction steps are a little complex, I'll try to simplify to only two very evident cases of unexpected behavior.

### Step-by-step reproduction instructions

Scenario 1:
- Add an image block picking an image from the Media Library that does have an alt attribute and does _not_ have a caption.
- Add a paragraph block.
- Select the two blocks and click Create a pattern from the block toolbar ellipsis button.
- Give a pattern a name, keep it synced, and save.
- Select the Pattern and choose 'Edit original'.
- In the pattern editor, select the Image block.
- In the settings panel, expand 'Advanced' and click 'Enable overrides'.
- In the modal dialog that oopens, giv ethe override a name and click 'Enable'.
- Click 'Save' at the top right of the editor.
- Go back to the post editor.
- Select the overridable image in the Pattern.
- Replace the image with one from the Media Library that does have a caption.
- Save.
- Observe the alt attribute has been replaced with the one of the new image.
- Observe the image block does show a caption.
- Refresh the page.
- Observe the caption disappeared.
- Expected: the caption of the new image to persist.

Scenario 2:
- Repeat the steps above in a new post and creating a new pattern but this time start with an image that does have a caption.
- At the end, replace the image with one from the Media Librady that does _not_ have a caption.
- Observe the caption of the old image persists.
- Expected: the caption to disappear, as the new image doesn't have one.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

[...]

PR #60694, opened 2024-04-12
Add `__default` binding for pattern overrides

kevin940726:
## What?

Part of https://github.com/WordPress/gutenberg/issues/59819, close https://github.com/WordPress/gutenberg/issues/58601. Alternative to https://github.com/WordPress/gutenberg/pull/59956.

> When creating a pattern and toggling 'Allow instance overrides' for a block, the block receives some attribute data that looks like this:
>
> ```
> "bindings":{"url":{"source":"core/pattern-overrides"},"title":{"source":"core/pattern-overrides"},"alt":{"source":"core/pattern-overrides"}}}
> ```
> This is the block bindings declaration, that declares that image block url, title and alt can be overriden within patterns.
>
> A problem arises when the block bindings feature supports more attributes in the future. Because this data is saved in a post type, it isn't easy to update after the fact.

## Why?

See https://github.com/WordPress/gutenberg/issues/58601.

## How?

Add a special `__default` binding for pattern overrides as suggested in https://github.com/WordPress/gutenberg/pull/60066#issuecomment-2033935584.

`__default` is a way to communicate that we want to connect all supported attributes to a source (pattern overrides.) So that when we add support for new attributes the pattern will be automatically connected.

## Testing Instructions

1. Create a synced pattern and add some supported blocks.
2. Assign names and allow overrides for those blocks.
3. Insert the pattern in a post.
4. The overridable blocks should be editable.
5. The edited blocks should reflect on frontend.

## Screenshots or screencast 
N/A

PR #62471, opened 2024-06-11
Block Bindings / Pattern Overrides: Prevent normal attribute updates when a __default binding exists

talldan:
## What?
See #62287, and specifically the comment here - https://github.com/WordPress/gutenberg/issues/62287#issuecomment-2159921605.

When overriding an image block, the image caption set in the media library can show and then unexpectedly disappear after a save and a refresh.

At the moment, the caption field is unsupported as a block binding (or pattern overrides) attribute, so it shouldn't be changeable from a pattern instance. This may change in the future, but that's a separate investigation.

## Why?
The issue happens due to the way block bindings continues to update the unbound attributes for a block. When adding an image to an image block, there are a range of attributes that are set by the block (`url`, `id`, `alt`, `caption` etc), and only some of these are supported by block bindings.

The ones that are supported get set as pattern overrides correctly. The ones that aren't still cause the normal attributes of the block to be updated. This is why the caption can appear. After a refresh, the caption disappears because the blocks in a pattern instance aren't saved anywhere.

The block bindings behavior makes sense for most binding sources. For example, if you only bind an image src to post meta, it's still possible to edit the other attributes. But for patterns, that's not true, users can't edit the unbound attributes.

It's worth noting that it's not just caption, the image block also sets some other attributes that might cause unexpected bugs, so even if caption is supported by block bindings in the future, a change like this is still required.

## How?
The PR addresses it by leveraging the `__default` binding added in https://github.com/WordPress/gutenberg/pull/60694. Essentially it considers the `__default` binding as covering all block attributes, but only the supported ones have active bindings.

## Testing Instructions
1. Add an empty image to a post
2. Select it and create a pattern using the block options menu on the toolbar
3. In the resulting dialog, name the pattern, keep it as a synced pattern, and submit
4. The pattern should be created in the post and selected. Select the 'Edit original' button from the block toolbar to edit the pattern.
5. Select the image
6. From the block inspector's advanced panel, select the Enable Overrides option
7. In the dialog, give the block a name (i.e. 'Image') and submit
8. Save the pattern
9. Go back to the post using the back button in the top bar
10. Select the image block, and select the 'Media Library' upload option
11. Select an image from the media library that has a caption set, or upload one and set a caption and choose it for the block

In this branch: the caption doesn't appear on the image
In trunk: the caption appears on the image but disappears after a refresh

### Before
https://github.com/WordPress/gutenberg/assets/677833/8de70340-aa66-4312-82fe-4e3800d91f12

### After
https://github.com/WordPress/gutenberg/assets/677833/650a0997-957e-43f0-8b67-229ed1619982

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -97,8 +97,9 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 			unlock( select( blocksStore ) ).getAllBlockBindingsSources()
 		);
 		const { name, clientId, context } = props;
-		const hasDefaultBinding =
-			props.attributes.metadata?.bindings?.[ DEFAULT_ATTRIBUTE ];
+		const hasPatternOverridesDefaultBinding =
+			props.attributes.metadata?.bindings?.[ DEFAULT_ATTRIBUTE ]
+				?.source === 'core/pattern-overrides';
 		const bindings = useMemo(
 			() =>
 				replacePatternOverrideDefaultBindings(
@@ -215,8 +216,11 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 						}
 					}
 
+					// Only apply normal attribute updates to blocks
+					// that have partial bindings. Currently this is
+					// only skipped for pattern overrides sources.
 					if (
-						! hasDefaultBinding &&
+						! hasPatternOverridesDefaultBinding &&
 						Object.keys( keptAttributes ).length
 					) {
 						setAttributes( keptAttributes );
@@ -231,7 +235,7 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 				context,
 				setAttributes,
 				sources,
-				hasDefaultBinding,
+				hasPatternOverridesDefaultBinding,
 			]
 		);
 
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/hooks/use-bindings-attributes.js
```diff
@@ -97,6 +97,8 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 			unlock( select( blocksStore ) ).getAllBlockBindingsSources()
 		);
 		const { name, clientId, context } = props;
+		const hasDefaultBinding =
+			props.attributes.metadata?.bindings?.[ DEFAULT_ATTRIBUTE ];
```
SantosGuillamot: Instead of just checking if it has the `__default` attribute, maybe we can check that it is `core/pattern-overrides` as we are doing [here](https://github.com/WordPress/gutenberg/blob/95999a97132560b801e9e18d28d33a87ff313e4e/packages/block-editor/src/hooks/use-bindings-attributes.js#L48-L50)? If we do that, we might want to change the name of the variable.

If we decide to support `__default` in any source in the future, I am not sure if skipping `setAttributes` for any attribute would be okay for all of them. It seems specific to pattern overrides to me.

Additionally, adding the `__default` property pointing to any other source, currently not supported, would skip `setAttributes` as well.

talldan: Sounds good. I've made that change 👍 

Ultimately I'm not sure if this PR is the right fix, but it's a small change, and I can't think of many other options.

SantosGuillamot (approved):
LGTM! We can always revisit the implementation to decide if this is the right approach. Should we add the backport label.

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 79d03f39b2

PR #62475, opened 2024-06-11
Update React API reference links in @wordpress/element reference-guides

amitraj2203:
## What?
Changes React APIs links to point to the updated one.

## Why?
The old links go to `https://legacy.reactjs.org` and shows `This content is out of date` notice on the site.

Visit https://developer.wordpress.org/block-editor/reference-guides/packages/packages-element/ and try to open any API reference link under _Related_ section. For example: [React lazy](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-element/#lazy) or [React memo](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-element/#memo)


t-hamano (commented):
Thanks for the PR!

This API reference is automatically generated from JSDoc, so we cannot update the Markdown directly.

Could you update the JSDoc of [this file](https://github.com/WordPress/gutenberg/blob/trunk/packages/element/src/react.js), run `npm run docs:build`, and then commit and push it?

amitraj2203:
@t-hamano Thanks for reviewing the PR. I have addressed the feedback now.

t-hamano (approved):
LGTM 👍

PR #53026, opened 2023-07-27
Image: Fix layout shift when lightbox is opened and closed

cbravobernal:
## What?

Currently, opening and closing the lightbox causes the website layout to shift, which can cause the zoom animation to appear broken, particularly on mobile. It is also possible to scroll the webpage while the lightbox is open by directly manipulating the scrollbar or using a mobile device, which is not in line with web standards.

This PR improves the lightbox implementation by removing the layout shift and updating the scroll handling so that the lightbox behaves reliably whenever a scroll is performed.

## Why?

We want to offer a polished lightbox experience, and the current experience appears broken in some cases. 

## How?
This PR removes `overflow: hidden` from the styles and updates the callback for the scroll handling.

Note: Expected behavior is that scrolling will be disabled on both desktop and mobile.

## Testing Instructions

1. Insert an image in a post and add enough content to ensure that the scrollbar will be visible.
2. Enable the lightbox using the block settings.
4. Publish and view the post on desktop.
5. Click on the image to open and close the lightbox — make sure that the scrollbar remains intact and that there is no layout shift.
6. Attempt scrolling while the lightbox is open using the mousewheel, `up` and `down` arrow keys, the `PageUp` and `PageDown` keys, and by directly clicking and dragging the scrollbar — verify that the lightbox stays open and doesn't scroll when performing any of those actions.
7. On mobile devices, test that scrolling is disabled and that the lightbox remains open when attempting to scroll using touch input. Test also that pinch to zoom while the lightbox is open works reasonably well.

#### Before the Fix

https://github.com/WordPress/gutenberg/assets/5360536/3789a30b-5df3-42d3-85aa-c4369f61bd77

#### After the Fix

https://github.com/WordPress/gutenberg/assets/5360536/3682b764-17a0-4964-9d9f-2f9cdc4f777f

PR #62477, opened 2024-06-11
Image lightbox: fix jitter scrolling by adopting css based overflow fix

madhusudhand:
## What?

Image lightbox currently uses javascript based scroll position reset on scroll event. This causes slight jitter effect in the UI.

lightbox-scroll

## How?

This fix removes the javascript based handlers. Instead it sets `overflow: hidden` to the `body` element when lightbox is invoked and restores its `original overflow state` when lightbox is closed.

## Testing Instructions

1. Enable "Expand on click" option for the image block in global styles.
2. Add an image to a page or post.
3. Click on the image to expand.
4. Try scrolling with and without this fix. Observe that scroll jitter is gone.
5. Also, test in a touch device.

### DIFF
--- a/packages/block-library/src/image/style.scss
+++ b/packages/block-library/src/image/style.scss
@@ -207,7 +207,7 @@
 
 	.lightbox-image-container {
 		position: absolute;
-		overflow: hidden;
+		// overflow: hidden;
 		top: 50%;
 		left: 50%;
 		transform-origin: top left;
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -258,12 +258,8 @@ function block_core_image_print_lightbox_overlay() {
 			data-wp-class--show-closing-animation="state.showClosingAnimation"
 			data-wp-watch="callbacks.setOverlayFocus"
 			data-wp-on--keydown="actions.handleKeydown"
-			data-wp-on-async--touchstart="actions.handleTouchStart"
-			data-wp-on--touchmove="actions.handleTouchMove"
-			data-wp-on-async--touchend="actions.handleTouchEnd"
 			data-wp-on-async--click="actions.hideLightbox"
 			data-wp-on-async-window--resize="callbacks.setOverlayStyles"
-			data-wp-on-async-window--scroll="actions.handleScroll"
 			tabindex="-1"
 			>
 				<button type="button" aria-label="$close_button_label" style="fill: $close_button_color" class="close-button">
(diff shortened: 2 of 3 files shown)


packages/block-library/src/image/style.scss
```diff
@@ -207,7 +207,7 @@
 
 	.lightbox-image-container {
 		position: absolute;
-		overflow: hidden;
+		// overflow: hidden;
```
madhusudhand, line 210: I believe, this is no longer required and additionally, it preserves effects such as shadow in the lightbox. But not quite sure if that should be the intended behavior.

| Before | After |
| -- | -- |
| image | image |

artemiomorales:
@madhusudhand Thanks for the PR! We actually had an implementation similar to this just using CSS in https://github.com/WordPress/gutenberg/pull/53026, and it was revised to prevent inconsistent behavior on mobile and layout shifts.

I tested this PR and noticed that the experience sometimes breaks on mobile (tested using Safari on iPhone). This is a web usability issue, because it can cause users to lose context of where they are on the page:

https://github.com/WordPress/gutenberg/assets/5360536/58c625cb-d531-45b9-a046-5209fe11f41f

Also, the layout shifts using this approach, which is particularly noticeable at smaller resolutions:

https://github.com/WordPress/gutenberg/assets/5360536/f53131a0-7114-4f33-ab18-2a18ee23bce0

https://github.com/WordPress/gutenberg/assets/5360536/125e69f7-8f78-4e7d-8316-c8cfb21fe046

The JavaScript approach and jitter isn't ideal; it would be great if we could have a CSS solution that behaves consistently on mobile and avoids jittering and layout shifts.

Would also be great to hear if anyone has other thoughts or ideas on how to address this 🙏

mikachan (commented):
> Would also be great to hear if anyone has other thoughts or ideas on how to address this 🙏

I'm adding some thoughts as I've attempted to fix similar issues in the past, but most solutions have trade-offs.

1. In addition to setting `overflow: hidden` on `body`, we could try setting `position: fixed` on `body` as well. Unfortunately, this also resets the scroll position, but perhaps we could scroll to the previous position with JS. (There's some examples of this [here](https://www.jayfreestone.com/writing/locking-body-scroll-ios/).)
2. We could consider using an additional package, like [body-scroll-lock](https://www.npmjs.com/package/body-scroll-lock). But I'm hesitate to suggest adding something like this if we can fix it using either just CSS or a combination of CSS and JS.

madhusudhand:
Closing it as the fix breaking the experience as mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/62477#issuecomment-2161218945).

PR #61606, opened 2024-05-13
List View: Add a special case for shortcuts coming from modals

Mamaduka:
## What?
An alternative to #61583

Avoid handling single-key block deletion shortcuts when events come from the modals while preserving the main shortcut for the action. This matches the canvas behavior.

## Why?
See https://github.com/WordPress/gutenberg/pull/61583#pullrequestreview-2051637643.

## Testing Instructions
1. Open a post or page.
2. Add a group block.
3. Open the List View.
4. Rename the group block via List View options.
5. Confirm you can delete the new block name.
6. Move focus to the modal buttons.
7. Press <kbd>Delete</kbd> confirm that block isn't deleted.

### Testing Instructions for Keyboard
Same.
## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/240569/f1817493-7ee3-447f-ba19-e880cbea14a8

PR #62479, opened 2024-06-11
List View: Respect default shortcuts in modals

### LINKED ISSUE #62478: [WP6.6] Select all shortcut doesn't work in modal opened from the List View
### Description

When a modal is opened from the List View, the shortcut to select all text (`Ctrl + a`, `command + a`) does not work in the text field inside the modal.

Instead of the text being selected, it seems to select a block of the List View.

This issue does not occur in WP6.5.

Note: I have confirmed that this problem occurs on Windows OS, but I would appreciate it if someone could test it on MacOS as well.

- Insert some blocks.
- Open the List View and select one block.
- Open the "Rename" or "Create pattern" menu.
- Execute `Ctrl + a`, (`command + a`) in the text field.

https://github.com/WordPress/gutenberg/assets/54422211/72b4d9fa-6d9f-47b0-a103-11e5364ab2cf

### Environment info

_No response_

Yes

Yes

t-hamano:
Similar to #61606
Fixes #62478

## What?

This PR will respect the default behavior when keyboard events occur in a modal.

This fixes the issue where we couldn't select all the text in modals with `Ctrl + a`, (`command + a`).

## Why?

This is because the event matches the shortcut for selecting all blocks at this point:

https://github.com/WordPress/gutenberg/blob/2d34e2aaa56e52c379b01bed2df2344ba01c9bb4/packages/block-editor/src/components/list-view/block.js#L290

## How?

I initially considered adding special handling for this shortcut. But it seemed more logical to me to just respect the default shortcuts for all events that occur inside a modal.

## Testing Instructions

- Insert some blocks.
- Open the List View and select one block.
- Open the "Rename" or "Create pattern" menu.
- Execute `Ctrl + a`, (`command + a`) in the text field.
- All of the text should be selected.
- Close the modal and select a block from the list view.
- Execute `Ctrl + a`, (`command + a`) in the text field.
- All of the blocks should be selected.
- Bonus: Follow the steps in #61606 to make sure the Delete and BackSpace keys still work.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/b1095b42-7510-41c8-ac66-33a7d729a826

### DIFF
--- a/packages/block-editor/src/components/list-view/block.js
+++ b/packages/block-editor/src/components/list-view/block.js
@@ -181,6 +181,12 @@ function ListViewBlock( {
 			return;
 		}
 
+		// Do not handle events if it comes from modals;
+		// retain the default behavior for these keys.
+		if ( event.target.closest( '[role=dialog]' ) ) {
+			return;
+		}
+
 		const isDeleteKey = [ BACKSPACE, DELETE ].includes( event.keyCode );
 
 		// If multiple blocks are selected, deselect all blocks when the user
@@ -196,12 +202,6 @@ function ListViewBlock( {
 			isDeleteKey ||
 			isMatch( 'core/block-editor/remove', event )
 		) {
-			// Do not handle single-key block deletion shortcuts when events come from modals;
-			// retain the default behavior for these keys.
-			if ( isDeleteKey && event.target.closest( '[role=dialog]' ) ) {
-				return;
-			}
-
 			const {
 				blocksToUpdate: blocksToDelete,
 				firstBlockClientId,


mcsf:
Thanks for the fix, @t-hamano! This makes sense to me, but I'd like @Mamaduka's opinion, in case we're missing anything.

colorful-tones:
I tested this in Gutenberg Playground PR, and it solves the issue. 👍

Mamaduka (approved):
Thank you, @t-hamano!

Switching to early return makes sense. Currently, I can’t think of a case when block shortcuts might need to be propagated from modals.

P.S. I think there’s an e2e test failure, not sure if it’s related.

andrewserong (approved):
+1 an early return sounds good here, and this is testing well for me with each of the list view keyboard shortcuts still working nicely when focus is within the list view, and with default behaviour working within the Rename modal 👍

t-hamano:
Thank you everyone for your reviews!

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 30648f5c45

Issue #62281, opened 2024-06-04
Button component: link buttons should not use aria-disabled


afercia:
### Description

Discovered while checking https://github.com/WordPress/gutenberg/pull/62080

The Button component accepts a `href` prop which is supposed to be an URL. When set, the Button renders an `<a>` element: a link with a proper href attribute.

However, when the Button component receives also the `disabled` and `;__experimentalIsFocusable` props both set to `true`, the rendered link gets an `aria-disabled="true"` attribute.

That's not correct for two reasons:
- It's invalid HTML: aria-disabled-true is invalid on an `<a>` element.
- Conceptually incorrect: links can't be 'disabled'. Either they are links or not.

Weirdly enough, when setting `__experimentalIsFocusable` to `false` and keeping the `href` prop, the Button is rendered as a disabled `<button>` element.

Rendered invalid HTML example:

```
<a href="https://wordpress.org" aria-disabled="true" class="components-button">Code is poetry</a>
```

It is worth mentioning that links can't 'really' be disabled. Their real nature is to _not_ be disabled. The defautl action can be prevented via JavaScript to make a click event do nothing but that's less than ideal. It's against the nature of a link.That's the reason why an `aria-disabled="true"` or `disabled` attribute on `<a>` elements are invalid HTML.

### Step-by-step reproduction instructions

- Go to https://wordpress.github.io/gutenberg/?path=/story/components-button--default&args=disabled:!true;__experimentalIsFocusable:!true
- Make sure both the `__experimentalIsFocusable` and `disabled` props are set to `true`.
- Set the `href` prop to `https://wordpress.org`
- Inspect the button rendered at the top of the page.
- Oberve it renders and `<a>` element with an `aria-disabled="true"` attribute.

### Screenshots, screen recording, code snippet

aria-disabled="true" invalid HTML

![Screenshot 2024-06-04 at 15 02 56](https://github.com/WordPress/gutenberg/assets/1682452/51807f9f-5c98-4424-9b11-d3b7db2db681)

disabled invalid HTML

![Screenshot 2024-06-04 at 15 18 44](https://github.com/WordPress/gutenberg/assets/1682452/e7cbf6ff-822e-48c1-be06-260dec4ff6bc)

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

afercia:
[...]

PR #19344, opened 2019-12-27
Try G2

youknowriad:
closes #18195 
closes #18069 
closes #18223 
closes #16617 
closes #19624
closes #7646 
closes #19018
closes #12260 
closes #15489 
closes #20086 
closes #4287 
closes #9522 

This branch should be used to serve as a base branch for all updates related to this issue #18667 that can't land individually.

 For now, it contains:

 - [x] Movers moved to the toolbar #18685 
 - [x] Change the block toolbar borders to black.
 - [x] Simplify hover/focus and radius styles for Buttons.
 - [x] Remove is-selected styles and replace with focus styles.
 - [x] white placeholders with dotted background.
 - [x] Block Switcher with an arrow on the corder of block toggle.
 - [x] New Dropdown Menu styles.
 - [x] No arrow indicator for dropdowns. 
 - [ ] Update the icons used.

PR #62480, opened 2024-06-11
Button: Improve the aria-disabled focus style

### LINKED ISSUE #62278: Button component: aria-disabled focusable button should use consistent focus style
### Description

Discovered while working on https://github.com/WordPress/gutenberg/pull/62267

Buttons can be made `aria-disabled` to still be focusable and this pattern has been recently made more solid in https://github.com/WordPress/gutenberg/pull/62080

When `aria-disabled` buttons receive focus, their focus style should be consistent and this should be managed in the Button component itself. Instead, I observed that sometimes some `aria-disabled` buttons use a custom focus style.

An important note for accessibility is that:
- Disabled controls don't have to meet the color contrast requirement.
- However, the entire purpose of making `aria-disabled` buttons still focusable is to make them discoverable and perceivable when they receive focus. This should apply also to the focus style so that I'd argue the focus style should meet the contrast requirement. Feedback and thoughts welcome Cc @joedolson 

The current situation of the focus style for `aria-disabled` buttons is pretty inconsistent and should be improved.

1
The Default button uses [some opacity](https://github.com/WordPress/gutenberg/blob/e57eed571e8aaa7d65394bdcf861b103dee9843f/packages/components/src/button/style.scss#L246-L250) so that both the text and the focus outline are lighter:

default

2
The Primary button only makes the text lighter. The outline (a box-shadow actually) is different:

primary

3
The Secondary button only makes the text color lighter. Anyways, the text color is different from the one of the Default button (which uses opacity):

secondary

4
Same for the Tertiary button:

tertiary

- Go to the Storybook https://wordpress.github.io/gutenberg/?path=/docs/components-button--docs
- Test the 4 Button variants: Default, Primary, Secondary, and Tertiary.
- Set both the `__experimentalIsFocusable` and `disabled` props to `true`.
- Move focus to the Button preview at the top of the page.
- Observe the focus style inconsistencies.

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/62278

## What?

The focus style for `aria-disabled` buttons is inconsistent across the button variants. Also, for some variants, it doesn't meet the WCAG contrast guidelines. For screenshots of the inconsistencies and insufficient color contrast of the focus style, please see the issue https://github.com/WordPress/gutenberg/issues/62278.

## Why?

For better accessibility, while the button itself may have some styling to visually indicate it's 'disabled', the focus style needs to be fully visible. The focus outline is a positional marker to allow a user to identify their current position in the page, and that needs to meet contrast guidelines.

## How?

- Removes any CSS `opacity` property for the button element styling. The only opacity value is now used for the SVG icon, when disabled.
- The disabled / aria-disabled color is now always `$gray-600: #949494` except for the `is-destructive` button.

### There are two visible changes:
- Default button, when disabled: it now uses `$gray-600 #949494` instead of opacity so it's slightly darker but it's consistent with the gray used for other variants.
- The `is-link` button, when disabled, is now gray instead of blue. I'm not too concerned about this change because, first of all, links can't be _really_ disabled. They should not have a disabled styling to start with. This is under discussion in https://github.com/WordPress/gutenberg/issues/62281. Anyways, I don't think it should stay blue (although dimmed) when 'disabled' because, historically, in WordPress the color blue indicates an _enabled_ interactive control.

## Testing Instructions

- Switch to this branch and run the Storybook: `npm run storybook:dev`
- The Storybook should automatically launch your browser and run at http://localhost:50240/
- To compare with the current styles, you can use the online version of the Storybook at 
https://wordpress.github.io/gutenberg/
- Test all the Button variants:
  - Default 
  - Primary
  - Secondary 
  - Tertiary 
  - Link 
  - Is Destructive 
  - Icon 
- First, test all the Button variants in their default state and observe there are no visual changes. Test the hover, focus, and active states.
- Then, test with only the `disabled` prop set to `true`.
    - Observe there are no visual changes exceot the ones mentioned above.
    - Test the hover and active states: observe the styling doesn't change when interacting with the button.
- Then, test with both the `disabled` and `__experimentalIsFocusable` props set to `true`.
    - Focus the buttons and observe the focus outline is always a solid color and has no opacity any longer.
    - Test the hover, focus, and active states: observe the styling doesn't change when interacting with the buttons, except for the focus outline.

## Screenshots or screencast 

Example screenshots to illustrate:

Before:

before

After: also arira-disabled buttons have a fully visible, consistent, focus outline:

after


packages/components/src/button/style.scss
```diff
@@ -23,16 +23,10 @@
 	}
 
 	&[aria-expanded="true"],
-	&:hover {
+	&:hover:not(:disabled, [aria-disabled="true"]) {
 		color: $components-color-accent;
 	}
 
-	// Unset some hovers, instead of adding :not specificity.
```
ciampo, line 37: Reading this comment, there could be some unintended consequences across the editor in terms of rule specificity with these changes

packages/components/src/button/style.scss
```diff
@@ -237,6 +234,11 @@
 		&:focus {
 			border-radius: $radius-block-ui;
 		}
+
+		&:disabled,
+		&[aria-disabled="true"] {
+			color: $gray-600;
```
ciampo, line 247: If we wanted to retain the original design, we could also use [`color-mix()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) to add an exact amount of opacity to the default text color

packages/components/src/button/style.scss
```diff
@@ -246,7 +248,11 @@
 	&:disabled,
 	&[aria-disabled="true"] {
 		cursor: default;
-		opacity: 0.3;
+		color: $gray-600;
+
+		svg {
+			opacity: 0.66;
+		}
```
ciampo: This update from `0.3` to `0.6` opacity for SVG in disabled options should also probably be reviewed by the design team

ciampo (commented):
Thank you for working on this 🙏 Probably good to get feedback from @WordPress/gutenberg-design

packages/components/src/button/style.scss
ciampo: Also, looking at other parts in the repo, `$gray-700` could be a better value here, should we not go for the `color-mix()` and opacity route.

But I'll let design folks comment with better context

jameskoster: It's not entirely clear to me why the default and `secondary` variant use inconsistent color methods (opacity vs specific value). The opacity method seems to pre-date the specific value, so it's probably fine to change this up.

packages/components/src/button/style.scss
jameskoster: Any reason for the svg not to use the same color as the button text? That will eliminate any inconsistencies between text + icon when both are visible.

packages/components/src/button/style.scss
joedolson: The CSS that was replaced by this was `&:not(:disabled):not([aria-disabled="true"]):not(.is-secondary):not(.is-primary):not(.is-tertiary):not(.is-link):hover`; I suspect that this was just intended for CSS simplification. It's always possible there's some unexpected consequences, but I don't think it's a significant concern.

joedolson: Reference: https://github.com/WordPress/gutenberg/pull/19344/files#diff-0688ad494526b3b921c4444dc8bc6f63efc4ad74c7c1123b061490cfc57f293a

packages/components/src/button/style.scss
joedolson: `$gray-700` would raise the color contrast to be more visible than I think it should be for a disabled control; we still need to pretty clearly convey that this control is disabled.

ciampo: If I understand correctly, then, you suggest using solid color? Which one would be the correct one, the 600 or 700 gray variant? (or any other ?)

packages/components/src/button/style.scss
joedolson: I assume that this is intended to shift the SVG to have a similar rendering to the text color? This change isn't reflected in the screenshot shared in the PR; I think that this is too dark, and results in insufficient difference between active and disabled buttons. 

image

Image shows the undo/redo buttons; the redo button has the opacity of .66 applied, and the difference is barely noticeable.

joedolson (dismissed):
I think that the SVG opacity change is too dark, but otherwise I think this is good.

packages/components/src/button/style.scss
joedolson: I don't have a preference on the opacity vs solid color question; but for a solid color, I'd say 600.

ciampo:
I can also confirm (via the [manual visual regression tests](https://github.com/WordPress/gutenberg/blob/fd058dff348903a1a2545f31e4c997ce7a5d8faa/test/storybook-playwright/README.md)) that, with the current changes in this PR, the differences with respect to `trunk` are:

| Scenario | Before (trunk) | After (this PR) |
|---|---|---|
| `link` variant, `disabled` | Screenshot 2024-07-19 at 18 48 32 | Screenshot 2024-07-19 at 18 48 12 |
| default (no variant), `disabled` | Screenshot 2024-07-19 at 18 48 28 | Screenshot 2024-07-19 at 18 48 15 |

packages/components/src/button/style.scss
afercia: One of the reasons why I would prefer solid colors over colors with opacity or alpha transparency is that measuring the color contrast ratio is way easier with solid colors. @jameskoster I understand some blocks e.g. the cover text need opacity/alpha but for user interface controls colors I'd prefer to always use solid colors. Any reason, from a design perspective, to not do that?

packages/components/src/button/style.scss
afercia: >  the redo button has the opacity of .66 applied, and the difference is barely noticeable.

@joedolson I'm not sure I'm seeing the same thing. Will comment on the PR

afercia:
Realted to https://github.com/WordPress/gutenberg/pull/62480/files#r1684570141 I'm not sure we're seeing the same color while testing this PR. In the screenshot below:
- Left: current trunk
- Right: this PR

The change to the SVG opacity which is now `0.66` on the gray `#949494` is barely noticeable. The SVG looks almost the same compared to the previous one that was inheriting from the button the color `var(--wp-components-color-foreground, #1e1e1e)` dn an opacity of `0..3`.

Rather, the _text_ color is different and I'd agree is too dark as pointed out. I'll look into it.

before after

afercia:
In the last commits I'm introducing a new color `$gray-500: #b7b7b7;`. In my testing this is the equivalent solid color of the previous one that was the result of `#1e1e1e` with `opacity: .3;`. A solid color is simpler, more understandable, and can be changed globally if necessary.

Screenshots where I added buttons text to better illustrate:

Before:

before


After:

after


I'm using the new gray for all the Button variants, including `link` and for the 'destructive' style. I'd think it's not correct for these buttons to still be blue or red when disabled. Gray is now the default color for all disabled/aria-disabled buttons.
Note: worth reminding that the `Button` component with `variant="link` renders a `<button>` element when disabled.

Screenshot:

Disabled:

Screenshot 2024-07-23 at 13 43 11

Enabled:

Screenshot 2024-07-23 at 13 43 19

packages/components/src/button/style.scss
jameskoster: A solid color seems more intentional, but might not work as well when the button appears on a non-white background. However opacity isn't perfect in that regard either, so changing this is probably okay. gray-600 seems a good fit.

jameskoster:
A new variable should probably be discussed separately. Any reason not to use gray-600 for now?

afercia:
> A new variable should probably be discussed separately. Any reason not to use gray-600 for now?

I wanted to use the exact, same color that is used on trunk now. The new `#b7b7b7` matches the current color given by `#1e1e1e` plus `opacity: .3`. As such, it's not a _new_ color. It is already in use. What problem introduces a new variable? Its intended usage is clearly documented in the `_colors.scss` file.

> Any reason not to use gray-600 for now?

I thought it is too dark. See screenshot, the second 'Redo' is too dark to be clearly distinguished from the enabled 'Undo':

Screenshot 2024-07-23 at 16 27 01

jameskoster:
> What problem introduces a new variable?

No problem. Indeed it's quite strange that the 500 value is missing from the scale. It just feels like a detail to discuss on it's own. That said it is easy to adjust later if required, so I won't object if Marco and Lena are happy :)

packages/components/src/button/style.scss
mirka: As we start adopting more structured theming systems (not necessarily the [Theme component](https://wordpress.github.io/gutenberg/?path=/docs/components-theme--docs), but any system that would allow arbitrary background colors), it would definitely be easier to programatically ensure sufficient contrast if we used solid colors. So if contrast is ever a concern, I would recommend avoiding `opacity`.

afercia: Thanks @mirka, that's a very good point I'd support it totally.

ciampo:
> > What problem introduces a new variable?
> 
> No problem. Indeed it's quite strange that the 500 value is missing from the scale. It just feels like a detail to discuss on it's own. That said it is easy to adjust later if required, so I won't object if Marco and Lena are happy :)

A hypothetical gray 500 variant would be part of the base styles and the general design system (not just the components package). So, as far as there are clear design guidelines on when to use it, and any contrast requirements are met, I don't see any technical issues with it.

afercia:
Thanks everyone for your feedback and thoughts. I no objections, I'd appreciate a final review by testing all the Button variants 🙏🏽

ciampo:
I believe @afercia is currently AFK, and therefore I'm happy to take over.

@joedolson I think that your blocking feedback has been addressed, but wanted to double-check.

@jameskoster :

- are these changes still ok, given the conversation happening in https://github.com/WordPress/gutenberg/issues/63856 ?
- are you ok with the introduction of `gray-500`, or does it warrant further / separate discussion?

I'd also ask for folks to do one last round of smoke testing around the editor. I'll also do one more round of visual regression on the `Button` component.

joedolson:
Yep. I'm good.

mirka:
I did a visual snapshot test of the Button variant matrix, and there is a regression in the `isPressed` && `disabled` case (became solid black, doesn't look disabled).

| Before | After |
|--------|-------|
|Button variant matrix, before|Button variant matrix, after|

(I will be adding a few more to the Button matrix, since in trunk it's currently missing the `disabled` versions of `isDestructive` and `isPressed` #64120)

packages/base-styles/_colors.scss
```diff
@@ -8,12 +8,13 @@
 $black: #000;			// Use only when you truly need pure black. For UI, use $gray-900.
 $gray-900: #1e1e1e;
 $gray-800: #2f2f2f;
-$gray-700: #757575;		// Meets 4.6:1 text contrast against white.
-$gray-600: #949494;		// Meets 3:1 UI or large text contrast against white.
+$gray-700: #757575;	// Meets 4.6:1 text contrast against white.
+$gray-600: #949494;	// Meets 3:1 UI or large text contrast against white.
+$gray-500: #b7b7b7;	// Used for disabled controls text and icons.
```
mirka: For what it's worth, in wp-components we have generally been using `$gray-600` for disabled text.

https://github.com/WordPress/gutenberg/blob/060d4ca57ab8441255db34e308c21a640adad2a6/packages/components/src/utils/colors-values.js#L64

Should we update all the other components to use `gray-500`? If so I will do this in a follow-up.

ciampo: Yup, that's what I've been trying to understand as well (cc @WordPress/gutenberg-design )

ciampo:
@mirka I pushed an update which should fix the regression that you noticed on pressed + disabled buttons.

Let's wait on final feedback from design folks on the overall changes (and on using a new shade of `$gray-500`) before final review + merging.

packages/base-styles/_colors.scss
jameskoster: I suggested to use `600` already, but it was [suggested](https://github.com/WordPress/gutenberg/pull/62480#issuecomment-2245427416) that it might be too dark.

If there's an existing precedent for using `600` for disabled text then I'd prefer to do that here too, and explore the new variable separately.

ciampo: Sounds reasonable. Let's take one step at a time

ciampo:
I reverted disabled text to use `$gray-600`, as discussed in https://github.com/WordPress/gutenberg/pull/62480#discussion_r1697529306. I will open a follow-up issue where we can discuss that aspect separately, since it involves the whole Gutenberg design system and not just the `Button` component.

This is how 	all `Button` variants look like with the current changes in the PR:

Screenshot 2024-07-31 at 16 16 44

The focus outline is still fully visible (ie. not transparent) even for disabled (but keyboard accessible) buttons.



I'd be in favour of merging this PR as-is, as it represents an improvement over what's on `trunk`, and therefore I'd like to get one final review (and hopefully an approval).

I will also be AFK for a week starting from tomorrow, so please feel free to merge this PR if it gets approved.

mirka (approved):
I checked the lastest button matrix and it looks correct. No issues popping up in the editor smoke testing either.

mirka:
I'm going to go ahead and merge to unblock, since it seems like @afercia is on vacation 😄

PR #62482, opened 2024-06-11
Fix: Unquoted file argument in declaration check script

sirreal:
## What?
Quote the file argument used in the script that checks compiled types.

## Why?

If a path includes spaces, it's interpreted as multiple arguments causing the script to fail.

[This was mentioned in Core slack](https://wordpress.slack.com/archives/C02QB2JS7/p1718086430650019). @WunderBart noticed the problematic paths seemed to include spaces.

## How?

Quote the file argument.

## Testing Instructions

`npm run build:package-types` should continue to work as expected.


up1512001:
@sirreal Thanks for your fix, I have checked it locally and now `npm run dev` is working as expected.

WunderBart (approved):
🚀

Issue #62208, opened 2024-06-02
`.wp-element-caption` class incorrectly applied to Quote block `<cite>` element in Editor


justintadlock:
### Description

Since the addition of the toggle button to show/hide the Quote citation in the editor in Gutenberg 18.1, it has applied the `.wp-element-caption` class to the `<cite>` element: https://github.com/WordPress/gutenberg/pull/59073

It creates inconsistent styling between the editor and front end because of the addition of the class, which means that it picks up caption styling (it should only have cite styling in both places)

### Step-by-step reproduction instructions

1. Add a Quote block with a citation.
2. Test on WordPress/Gutenberg pre-18.1.
3. Test on Gutenberg 18.1 or newer and see the addition of the `.wp-element-caption` class on the editor side.

### Screenshots, screen recording, code snippet

**Editor:**

![quote-cite-editor](https://github.com/WordPress/gutenberg/assets/1816309/33cd64e8-3cd7-466e-91d2-1f35cd991eb7)

**Front end:**

![quote-cite-front-end](https://github.com/WordPress/gutenberg/assets/1816309/d0656cd9-563f-48cc-bed1-554ac4ea674f)

### Environment info

- WordPress trunk
- Gutenberg 18.5 RC1

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

aatanasovdev:
For the `<cite>` element, the Quote block relies on a utility component called [Caption](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/utils/caption.js), which is used by other blocks such as Video, Gallery, etc. It was prepared in [this pull request](https://github.com/WordPress/gutenberg/pull/56606). 

The front end doesn't have the `.wp-element-caption` because the save method doesn't use `__experimentalGetElementClassName( 'caption' )`  as it is done for the other blocks.

Based on that, we have the following approaches:
1. If we need to reuse the existing Caption utility component, it is better if we use the `.wp-element-caption` class in the Editor and on the front end.
2. We create a new component for the `<cite>` element.

The second option will be appropriate if we don't consider the citation as a caption and then reuse the new component for other `<cite>` related places in the future. Otherwise, the first option makes sense for better consistency.

justintadlock:
> If we need to reuse the existing Caption utility component, it is better if we use the `.wp-element-caption` class in the Editor and on the front end.

[...]

PR #62485, opened 2024-06-11
Caption utility component: Allow the main CSS Class Name to be excluded from the markup.

aatanasovdev:
## What?

Extending the Caption utility component by allowing the main CSS Class Name to be excluded from the HTML Markup using a prop called `excludeElementClassName`. Also, the Quote block was modified to use the prop.

## Why?

This update is related to this issue: [62208](https://github.com/WordPress/gutenberg/issues/62208)

The new prop allows us to reuse the existing Caption component in places where its main CSS Class Name (`.wp-element-caption`) isn't needed.

## How?

It addresses the issue by adding a prop to the Caption component, which excludes the main CSS Class Name when needed.

## Testing Instructions

1. Insert a Quote block with a citation.
2. Inspect the HTML markup of the block.
3. The `wp-element-caption` CSS Class Name shouldn't exist in the `<cite>` element.
4. Insert a Gallery block with some images and add captions to them.
5. Inspect the HTML markup of an image caption.
6. The `wp-element-caption` CSS Class Name should exist in the `<figcaption>` element.

### DIFF
--- a/packages/block-library/src/utils/caption.js
+++ b/packages/block-library/src/utils/caption.js
@@ -32,6 +32,7 @@ export function Caption( {
 	placeholder = __( 'Add caption' ),
 	label = __( 'Caption text' ),
 	showToolbarButton = true,
+	excludeMainElementClassName,
 	className,
 	readOnly,
 	tagName = 'figcaption',
@@ -70,6 +71,7 @@ export function Caption( {
 		},
 		[ isCaptionEmpty ]
 	);
+
 	return (
 		<>
 			{ showToolbarButton && (
@@ -96,7 +98,9 @@ export function Caption( {
 						tagName={ tagName }
 						className={ clsx(
 							className,
-							__experimentalGetElementClassName( 'caption' )
+							excludeMainElementClassName
+								? ''
+								: __experimentalGetElementClassName( 'caption' )
 						) }
 						ref={ ref }
 						aria-label={ label }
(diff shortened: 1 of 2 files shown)


aatanasovdev:
Regarding the failed tests:
 - **Enforce labels on Pull Request:** I missed adding the necessary labels: `[Block] Quote` and `[Type] Bug`
 - **React Native E2E Tests (Android) / Unit Tests / JavaScript (Node.js 22) 2/4** Reviewing the details, it appears the errors aren't related to the PR code itself, and I'm not sure if any actions should be done for these two in this PR.

packages/block-library/src/utils/caption.js
```diff
@@ -32,6 +32,7 @@ export function Caption( {
 	placeholder = __( 'Add caption' ),
 	label = __( 'Caption text' ),
 	showToolbarButton = true,
+	excludeMainElementClassName,
```
talldan: I think the naming of the prop could be simplified:
```suggestion
	excludeElementClassName,
```

talldan (approved):
Thanks for working on this change!

The code looks good to me, but I think the prop naming could be simplified a little.

It doesn't really affect the functionality though, so I'll approve it.

aatanasovdev:
@talldan, thank you for approving the change and suggesting a simplified prop name.

I've pushed a new commit with the simplified name.

talldan:
Thanks @aatanasovdev for your contribution.

There's a unit test failure related to the `caniuse` dependency being outdated, but this is happening in `trunk` too, so I'll go ahead and merge the PR and look to fix those failures separately (fixed in https://github.com/WordPress/gutenberg/pull/62611).

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: b6178b74f4

PR #62488, opened 2024-06-11
Fix: Post_type template is not used when creating a page in site editor.

### LINKED ISSUE #60710: WP_Post_Type->template is not used in site editor
### Description

The initial state of a new post type [is defined](https://core.trac.wordpress.org/browser/tags/6.5/src/wp-includes/class-wp-post-type.php#L287) in `WP_Post_Type->template`. This initial state is acknowledged by the block editor when creating a new post/page/etc but not in the site editor. 

1. Register a [post type block template](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-templates/#api) for a post type that appears in the site editor. We'll use`page` (but you could use `wp_block`). Place this in a `functions.php` or plugin file:

```php
function myplugin_register_template() {
    $post_type_object = get_post_type_object( 'page' );
    $post_type_object->template = array(
        array( 'core/image' ),
    );
}
add_action( 'init', 'myplugin_register_template' );
```
2. Start a new page from the regular dashboard, which loads the block editor, and you'll see the initial block content (an image block) in the post content.
3. Start a new page from inside the site editor - you'll see that there is no initial block content.

_No response_

### Environment info

- WordPress 6.5.2
- Gutenberg 18.1.0

Yes

Yes

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60710

The site editor now allows to create a new page but if the page post type has a template assigned to it the template is ignored.
This PR fixes the issue.

To fix the issue, the Rest API has been extended for the types endpoint. Two new fields were added template and template_lock.

## Testing Instructions
I added the following test code to scrip-loader.php.
```
function myplugin_register_template() {
    $post_type_object = get_post_type_object( 'page' );
    $post_type_object->template = array(
        array( 'core/image' ),
    );
}
add_action( 'init', 'myplugin_register_template' );
```
I created a new page on the site editor and verified the template was used as expected.

### DIFF
--- a/lib/rest-api.php
+++ b/lib/rest-api.php
@@ -30,3 +30,51 @@ if ( ! function_exists( 'gutenberg_register_edit_site_export_controller_endpoint
 }
 
 add_action( 'rest_api_init', 'gutenberg_register_edit_site_export_controller_endpoints' );
+
+if ( ! function_exists( 'gutenberg_register_wp_rest_post_types_controller_fields' ) ) {
+	/**
+	 * Adds `template` and `template_lock` fields to WP_REST_Post_Types_Controller class.
+	 */
+	function gutenberg_register_wp_rest_post_types_controller_fields() {
+		register_rest_field(
+			'type',
+			'template',
+			array(
+				'get_callback' => function ( $item ) {
+					$post_type = get_post_type_object( $item['slug'] );
+					if ( ! empty( $post_type ) && ! empty( $post_type->template ) ) {
+						return $post_type->template;
+					}
+					return null;
+				},
+				'schema'       => array(
+					'type'        => 'array',
+					'description' => __( 'The template associated with the post type.', 'gutenberg' ),
+					'readonly'    => true,
+					'context'     => array( 'view', 'edit', 'embed' ),
+				),
+			)
+		);
+		register_rest_field(
+			'type',
+			'template_lock',
+			array(
+				'get_callback' => function ( $item ) {
+					$post_type = get_post_type_object( $item['slug'] );
+					if ( ! empty( $post_type ) && ! empty( $post_type->template_lock ) ) {
+						return $post_type->template_lock;
+					}
+					return null;
+				},
+				'schema'       => array(
+					'type'        => 'string',
+					'emum'        => array( 'all', 'insert', 'contentOnly' ),
+					'description' => __( 'The template_lock specified for the post type.', 'gutenberg' ),
+					'readonly'    => true,
+					'context'     => array( 'view', 'edit', 'embed' ),
+				),
+			)
+		);
+	}
+}
+add_action( 'rest_api_init', 'gutenberg_register_wp_rest_post_types_controller_fields' );
--- a/lib/rest-api.php
+++ b/lib/rest-api.php
@@ -68,7 +68,7 @@ if ( ! function_exists( 'gutenberg_register_wp_rest_post_types_controller_fields
 				},
 				'schema'       => array(
 					'type'        => 'string',
-					'emum'        => array( 'all', 'insert', 'contentOnly' ),
+					'enum'        => array( 'all', 'insert', 'contentOnly' ),
 					'description' => __( 'The template_lock specified for the post type.', 'gutenberg' ),
 					'readonly'    => true,
 					'context'     => array( 'view', 'edit', 'embed' ),
--- a/lib/rest-api.php
+++ b/lib/rest-api.php
@@ -45,7 +45,6 @@ if ( ! function_exists( 'gutenberg_register_wp_rest_post_types_controller_fields
 					if ( ! empty( $post_type ) && ! empty( $post_type->template ) ) {
 						return $post_type->template;
 					}
-					return null;
 				},
 				'schema'       => array(
 					'type'        => 'array',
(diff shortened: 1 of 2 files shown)


lib/rest-api.php
```diff
@@ -30,3 +30,51 @@ function gutenberg_register_edit_site_export_controller_endpoints() {
...
+				'get_callback' => function ( $item ) {
+					$post_type = get_post_type_object( $item['slug'] );
+					if ( ! empty( $post_type ) && ! empty( $post_type->template_lock ) ) {
+						return $post_type->template_lock;
+					}
+					return null;
+				},
+				'schema'       => array(
+					'type'        => 'string',
+					'emum'        => array( 'all', 'insert', 'contentOnly' ),
```
oandregal:
```suggestion
					'enum'        => array( 'all', 'insert', 'contentOnly' ),
```

oandregal: Also: it seems `template_lock` can be a string or a boolean and `false` is a valid value, according to [this](https://github.com/WordPress/wordpress-develop/blob/5c7f94a6aa979dc70ff935e9006957fe0f262932/src/wp-includes/class-wp-post-type.php#L302), see also [this](https://github.com/WordPress/wordpress-develop/blob/5c7f94a6aa979dc70ff935e9006957fe0f262932/src/wp-includes/blocks/blocks-json.php#L3228).

lib/rest-api.php
```diff
@@ -30,3 +30,51 @@ function gutenberg_register_edit_site_export_controller_endpoints() {
...
+		register_rest_field(
+			'type',
+			'template_lock',
+			array(
+				'get_callback' => function ( $item ) {
+					$post_type = get_post_type_object( $item['slug'] );
+					if ( ! empty( $post_type ) && ! empty( $post_type->template_lock ) ) {
+						return $post_type->template_lock;
+					}
+					return null;
```
oandregal: According to [this other logic](https://github.com/WordPress/wordpress-develop/blob/5c7f94a6aa979dc70ff935e9006957fe0f262932/src/wp-admin/edit-form-blocks.php#L231), it should return `false`.

oandregal (approved):
I've left a comment about `template_lock` valid values we need to address. Other than that, this looks good and it's working so I'm pre-approving to speed up the merge process given the time-constrains (beta period for 6.6).

lib/rest-api.php
jorgefilipecosta: Hi @oandregal, setting a value of false is the same as not having template_lock, I guess we can ignore the false values, and return null/not return when locking is not set either because there is no locking referred at all or because it is false. I don't see any downside to that and this would allow us to use an enum as a type with only array( 'all', 'insert', 'contentOnly' ). We can not really have a type that is enum or boolean. What do you think?

oandregal: What about nested content? Is there a difference between `false` or `null` in that scenario? I'm thinking about the locking lookup: if I remember this correctly, if a block (`wp_navigation` or `wp_block` types, for example) doesn't have a lock, we look it up above in the parents. I don't remember the details of "does not have a lock": was that checking for `false` or for `null`/non existence (or both)?

Other than double checking that scenario, the current approach makes sense to me.

jorgefilipecosta: The nested locking is something at the block level, this template lock is only used now on edit-post, and the template locks of CPTs can not be nested. The locking of templates on CPTs is always top level there is never a need to overite with false as may happen on the block level. So I think we are safe not returning the false value at all.

oandregal: Ah, that sounds good. Thanks for the braindump.

lib/rest-api.php
```diff
@@ -30,3 +30,49 @@ function gutenberg_register_edit_site_export_controller_endpoints() {
...
+					'enum'        => array( 'all', 'insert', 'contentOnly' ),
+					'description' => __( 'The template_lock specified for the post type.', 'gutenberg' ),
+					'readonly'    => true,
+					'context'     => array( 'view', 'edit', 'embed' ),
+				),
+			)
+		);
+	}
+}
+add_action( 'rest_api_init', 'gutenberg_register_wp_rest_post_types_controller_fields' );
```
ellatrix, line 78: Should we be worried about changes to the REST API this late in the release cycle? This PR will need a backport PR asap.

jorgefilipecosta: We are only adding two fields that will for sure be required, so I don't think we have back-compatibility problems or unexpected issues. I'm preparing a backport PR, but I don't have a strong option. we can opt for not include this fix in WP 6.6 if you think it is safer.

lib/rest-api.php
```diff
@@ -30,3 +30,49 @@ function gutenberg_register_edit_site_export_controller_endpoints() {
...
+			array(
+				'get_callback' => function ( $item ) {
+					$post_type = get_post_type_object( $item['slug'] );
+					if ( ! empty( $post_type ) && ! empty( $post_type->template ) ) {
+						return $post_type->template;
+					}
+				},
+				'schema'       => array(
+					'type'        => 'array',
+					'description' => __( 'The template associated with the post type.', 'gutenberg' ),
```
mcsf, line 51: Should the phrasing reflect that the template is optional and refers to _block_ templates as opposed to a `wp_template`? For example:

```diff
-The template associated with the post type.
+The block template associated with the post type, if it exists.
```

ellatrix:
Manually resolved this in 5afddc857ab4cbdbc6e53a6f292accc2ca2fc475

PR #57841, opened 2024-01-15
Try reducing specificity of global styles block selector.

tellthemachines:
## What?

Tries to address specificity issues in https://github.com/WordPress/gutenberg/pull/56540#issuecomment-1891118802 by wrapping the block global styles root selector, global element selectors, global feature selectors and global layout selectors in `:where()`. 

For this to work it was also necessary to lower the specificity of some block-library styles, especially ones from the optionally enqueued "theme" files.

This should get some pretty thorough testing with themes, all sorts of blocks with global styles and variations thereof. 

## Testing Instructions

1. Test global styles on all blocks and their style variations and check they still work as expected. 
2. Test global and local layout styles and check that blocks with global margins set don't override the layout margins.
3. Test block themes with theme support for `wp-block-styles` (only early block themes such as TT2 have this) and check global styles override block-library styles correctly.

PR #50905, opened 2023-05-24
Performance tests: Make theme versions consistent cross-env

WunderBart:
## What?

Make performance testing consistent cross-env theme-wise.

_This is an alternative approach to https://github.com/WordPress/gutenberg/pull/50868 (See https://github.com/WordPress/gutenberg/pull/50868#issuecomment-1558989101)_

## Why?

[For consistency sake](https://github.com/WordPress/gutenberg/pull/47037/files#r1087561588), we've been using [fixed versions](https://github.com/WordPress/gutenberg/blob/trunk/bin/plugin/commands/performance.js#L340-L341) of `twentytwentyone` and `twentytwentythree` themes in our performance tests. The way they are currently being activated, though, is applied only to the `performance.js` runner (which we use in CI). The same tests, when run directly via `npm run test:performance <suite-name>`, will not use the same themes because they're not available via the [root wp-env instance](https://github.com/WordPress/gutenberg/blob/trunk/.wp-env.json) – they will use the latest versions instead.

## How?

As with other test themes, we can upload the target versions to the `gutenberg-test-themes` folder and utilize them via test env folder mapping in the wp-env config.

_For more details see @oandregal's comment: https://github.com/WordPress/gutenberg/pull/50905#discussion_r1205181738_

## Testing Instructions

- CI perf tests should pass,
- The following should pass locally:
   - `npm run test:performance front-end-block-theme.test.js`
   - `npm run test:performance front-end-classic-theme.test.js`
   - `npm run test:performance site-editor.test.js`
   - `npm run test:performance post-editor.test.js`

- The following suites should be using the following themes:
   - `front-end-block-theme.test.js` → `twentytwentythree@v1.0`
   - `front-end-classic-theme.test.js` → `twentytwentyone@v1.7`
   - `site-editor.test.js` → `emptytheme`
   - `post-editor.test.js` → `emptytheme`

PR #62492, opened 2024-06-11
remove: unnessary theme installation at docker setup

### LINKED ISSUE #62491: Double time TT4, TT3 and TT1 theme is given in appearance
### Description

I have set up the Gutenberg repo locally and I am running it using Docker. There are multiple instances of TT4, TT3, and TT1 installed with different versions, including version specified in `.wp-env.json` and the latest.

- clone Gutenberg repo
- cd to gutenberg folder
- run `npm install` first
- start docker desktop and run `npm run wp-env start`
- now run `npm run dev`

Screenshot 2024-06-11 at 23 12 07

### Environment info

_No response_

Yes

Yes

up1512001:
## What?

This PR removes TT4, TT3 and TT1 theme installation when running `npm run wp-env start` as we are already getting all default themes from WordPress.

## Why?
https://github.com/WordPress/gutenberg/issues/62491

## How?
remove unnecessary theme installations from `.wp-env.json` file.

## Screenshot after modifying `.wp-env.json`

Screenshot 2024-06-11 at 23 36 24

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -8,10 +8,7 @@
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",
 				"wp-content/plugins/gutenberg-test-plugins": "./packages/e2e-tests/plugins",
-				"wp-content/themes/gutenberg-test-themes": "./test/gutenberg-test-themes",
-				"wp-content/themes/gutenberg-test-themes/twentytwentyone": "https://downloads.wordpress.org/theme/twentytwentyone.2.1.zip",
-				"wp-content/themes/gutenberg-test-themes/twentytwentythree": "https://downloads.wordpress.org/theme/twentytwentythree.1.3.zip",
-				"wp-content/themes/gutenberg-test-themes/twentytwentyfour": "https://downloads.wordpress.org/theme/twentytwentyfour.1.0.zip"
+				"wp-content/themes/gutenberg-test-themes": "./test/gutenberg-test-themes"
 			}
 		}
 	}


.wp-env.json
```diff
@@ -8,10 +8,7 @@
 				"wp-content/plugins/gutenberg": ".",
 				"wp-content/mu-plugins": "./packages/e2e-tests/mu-plugins",
 				"wp-content/plugins/gutenberg-test-plugins": "./packages/e2e-tests/plugins",
-				"wp-content/themes/gutenberg-test-themes": "./test/gutenberg-test-themes",
-				"wp-content/themes/gutenberg-test-themes/twentytwentyone": "https://downloads.wordpress.org/theme/twentytwentyone.2.1.zip",
-				"wp-content/themes/gutenberg-test-themes/twentytwentythree": "https://downloads.wordpress.org/theme/twentytwentythree.1.3.zip",
-				"wp-content/themes/gutenberg-test-themes/twentytwentyfour": "https://downloads.wordpress.org/theme/twentytwentyfour.1.0.zip"
```
youknowriad, line 18: I think these themes are actually used in performance tests no?

youknowriad: I think if we remove these, the performance tests in CI might still run but locally, you'll have issues running these. cc @oandregal

up1512001: @youknowriad but when I see that under the test folder all three theme folders are empty even before this modification also 🤔

vcanales: They're being downloaded in the test environment, and they're used here:

https://github.com/WordPress/gutenberg/blob/57b8b613d0507fd2c1ed0b441db09f76a43d579f/bin/plugin/commands/performance.js#L422-L430

vcanales:
Regarding the original issue, https://github.com/WordPress/gutenberg/issues/62491, are you going to the site on port `8888` (development) or `8889` (test)? I usually do my local manual testing on the development site, since `8889` is used by automation. I can reproduce your issue, but I also expect the test site to have its own quirks.

```sh
➜  gutenberg git:(trunk) npm run wp-env start -- --xdebug=profile,trace,debug

> gutenberg@18.5.0 wp-env
> wp-env start --xdebug=profile,trace,debug

WordPress development site started at http://localhost:8888 # I use this one for development
WordPress test site started at http://localhost:8889
```

Perhaps we're lacking a bit of cleanup on the test site, though.

up1512001:
I check on the `8889` port. @vcanales

up1512001:
Still, one strange thing why the TT4, TT3 and TT1 folders are empty even after running the performance test?
cc: @vcanales @youknowriad 
Screenshot 2024-06-11 at 23 57 31

.wp-env.json
up1512001: @youknowriad I run a performance test locally and it's working fine. 
`npm run test:performance`

t-hamano: This was added in #50905, but the purpose seems to be to keep theme versions consistent across performance tests.

talldan:
> I check on the 8889 port.

It's recommended to use 8888 for anything you're doing manually, 8889 is for running tests. Those automated tests expect a clean environment, so you should avoid changing stuff in the WordPress instance there. When you run the tests, they'll also do some clean up like deleting posts, sometimes patterns, or even resetting styles, which is something you might not want to happen.

.wp-env.json
up1512001: I guess we should update the performance test to run on the latest version to avoid duplicate theme installation.

up1512001:
I guess we should update the performance test to run on the latest version to avoid duplicate theme installation.

SirLouen:
Yes, I agree with @up1512001 this is a faulty behaviour (lucky that he found this out of chance playing around in `:8889`. 

Basically, these are being installed twice in the testing container

```
# ls
emptytheme             index.php     twentyfifteen   twentynineteen   twentysixteen  twentythirteen  twentytwenty      twentytwentyfour  twentytwentythree
gutenberg-test-themes  twentyeleven  twentyfourteen  twentyseventeen  twentyten      twentytwelve    twentytwentyfive  twentytwentyone   twentytwentytwo
root@30d0550a4305:/var/www/html/wp-content/themes# ls gutenberg-test-themes/
block-bindings  darktheme  emptyhybrid  style-variations  twentytwentyfour  twentytwentyone  twentytwentythree
```

Also with this removed it seems that all tests are passing @talldan. 

Tests should not be failing as these themes are being installed on the default image
Also it doesnt make any sense having to maintain those versions. 

Can you elaborate a bit more on your concern?

talldan:
> Also with this removed it seems that all tests are passing @talldan.

> Tests should not be failing as these themes are being installed on the default image
> Also it doesnt make any sense having to maintain those versions.

> Can you elaborate a bit more on your concern?

I don't think it was me that shared the concern - I just mentioned that port 8888 is what we use for local development, while 8889 is what we use for automated tests. (in relation to this comment - https://github.com/WordPress/gutenberg/pull/62492#issuecomment-2161365678)

The concerns were shared in this thread by others - https://github.com/WordPress/gutenberg/pull/62492#discussion_r1635298642

Those seem valid to me. The themes used in the test environments are fixed to a specific version number to keep the environment stable - it ensures upstream changes to themes don't cause the tests to fail. I think it's mostly a concern for performance tests where a specific baseline is used to compare performance.

Maybe when those locked version themes are added, any already installed themes of the same name should be removed?

SirLouen:
> I don't think it was me that shared the concern - I just mentioned that port 8888 is what we use for local development, while 8889 is what we use for automated tests. (in relation to this comment - [#62492 (comment)](https://github.com/WordPress/gutenberg/pull/62492#issuecomment-2161365678))
> 
> The concerns were shared in this thread by others - [#62492 (comment)](https://github.com/WordPress/gutenberg/pull/62492#discussion_r1635298642)

@talldan @vcanales in fact this is not completely right. Performance tests actually create their own env file and download and install their own versions regardless of the ones specified here. 

As @t-hamano commented, this was added in #50905 for performance testing but in #57841 they noted that versions were causing it to fail.

> I guess we should update the performance test to run on the latest version to avoid duplicate theme installation.

If certain specific versions are required for this to work (which doesn't make any sense to have fixed versions because theoretically these tests should be always done against latest versions), then they could simply be moved into the default directory. Not only to avoid duplicate theme duplication, but to test against the reality. Tests are meant to protect against current threats.

Technically, the performance part is not required any more. Unless I'm wrong, it doesn't make any sense because `writeFileSync` doesn't write on top of current `.wp-env.json` and since tests are being thrown against with the code as-is, it looks redundant. 

I've added #70754 just to check the performance testing suite on GHA. I've tested locally and it's passing 100%.

PS: Note that when #50905 was sent, `wp-env` did not include by default all these themes. This is why they had to be added on top of the current installation.

PR #62497, opened 2024-06-12
Quote blocks: add background image and minimum height support

ramonjd:
## What?

This PR adds support for background images for the Pullquote and Quote blocks. 

For flexibility to display varying image heights, it also opts in to minimum height for these blocks, but not as a default control.

## Why?

The blockquote element has indentation — an inherit visual property that serves to differentiate it from surrounding text. 

The ability to add background images extends visual differentiation, and allows users to set unique design patterns. Example:

Screenshot 2024-06-12 at 2 36 32 PM

As for adding minimum height support, it ensures that users can adequately space quote blocks to accommodate taller images.

## How?

## Testing Instructions

Add a Quote and Pullquote block to a post.

Select background images for both, and play around with minimum height and other styles according to your delight.

Here's some HTML to get you started:

```html
<!-- wp:quote -->
<blockquote class="wp-block-quote"><!-- wp:paragraph -->
<p>"I love mountains..."</p>
<!-- /wp:paragraph --></blockquote>
<!-- /wp:quote -->

<!-- wp:pullquote -->
<figure class="wp-block-pullquote"><blockquote><p>"And forests!"</p><cite>Me</cite></blockquote></figure>
<!-- /wp:pullquote -->
```

### DIFF
--- a/packages/block-library/src/pullquote/block.json
+++ b/packages/block-library/src/pullquote/block.json
@@ -26,6 +26,13 @@
 	"supports": {
 		"anchor": true,
 		"align": [ "left", "right", "wide", "full" ],
+		"background": {
+			"backgroundImage": true,
+			"backgroundSize": true,
+			"__experimentalDefaultControls": {
+				"backgroundImage": true
+			}
+		},
 		"color": {
 			"gradients": true,
 			"background": true,
@@ -33,6 +40,12 @@
 			"__experimentalDefaultControls": {
 				"background": true,
 				"text": true
+			}
+		},
+		"dimensions": {
+			"minHeight": true,
+			"__experimentalDefaultControls": {
+				"minHeight": false
 			}
 		},
 		"spacing": {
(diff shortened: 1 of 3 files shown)


packages/block-library/src/pullquote/block.json
```diff
@@ -26,6 +26,13 @@
 	"supports": {
 		"anchor": true,
 		"align": [ "left", "right", "wide", "full" ],
+		"background": {
+			"backgroundImage": true,
+			"backgroundSize": true,
+			"__experimentalDefaultControls": {
+				"backgroundImage": true
```
ramonjd, line 33: Should we collapse the background image tools panel by default? 

I'm tending towards "yes" - there are quite a few controls here.

ramonjd: Ignore this. It's pretty compact

tellthemachines (approved):
Nice! This is a useful enhancement and works well in testing ✅

packages/block-library/src/pullquote/block.json
tellthemachines: Yeah it doesn't bother me! Also helps with discoverability.

ramonjd:
Thanks for the lightning fast reviews 🙇🏻

PR #62498, opened 2024-06-12
Verse block: add background image and minimum height support

ramonjd:
## What?

This PR adds support for background images for the Verse block. 

For flexibility to display varying image heights, it also opts in to minimum height, but not as a default control.

## Why?

The Verse block contains a pre element that formats text as it would be presented in the browser. Top notch for poetry!

The ability to add background images extends visual differentiation, and allows users to set unique design patterns. 

Example:

Screenshot 2024-06-12 at 3 02 22 PM

As for adding minimum height support, it ensures that users can adequately space Verse blocks to accommodate taller images.

## How?

## Testing Instructions

Add a Verse block to a post.

Select a background image, and play around with minimum height and other styles according to your fancy.

Here's some HTML to get you started:

```html
<!-- wp:verse -->
<pre class="wp-block-verse">When the letter arrived,<br>     and a day so contrived.<br><br>the first word of it, scrawled and thin<br>     breath drawing in<br><br>I read it, and read it<br>      I read it again.<br><br>With the sun in the west,<br>I stubbed out the candle<br>    And with the beef casserole,<br>    threw it all out with the rest.     <br></pre>
<!-- /wp:verse -->
```

### DIFF
--- a/packages/block-library/src/verse/block.json
+++ b/packages/block-library/src/verse/block.json
@@ -21,12 +21,25 @@
 	},
 	"supports": {
 		"anchor": true,
+		"background": {
+			"backgroundImage": true,
+			"backgroundSize": true,
+			"__experimentalDefaultControls": {
+				"backgroundImage": true
+			}
+		},
 		"color": {
 			"gradients": true,
 			"link": true,
 			"__experimentalDefaultControls": {
 				"background": true,
 				"text": true
+			}
+		},
+		"dimensions": {
+			"minHeight": true,
+			"__experimentalDefaultControls": {
+				"minHeight": false
 			}
 		},
 		"typography": {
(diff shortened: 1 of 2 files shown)


packages/block-library/src/verse/block.json
```diff
@@ -21,6 +21,13 @@
 	},
 	"supports": {
 		"anchor": true,
+		"background": {
+			"backgroundImage": true,
+			"backgroundSize": true,
+			"__experimentalDefaultControls": {
+				"backgroundImage": true
```
ramonjd, line 28: Should we collapse the background image tools panel by default? 

I'm tending towards "yes" to avoid clutter.

ramonjd: Ignore this. It's pretty compact

Screenshot 2024-06-12 at 3 15 09 PM

tellthemachines (approved):
Loving these tiny enhancement PRs ❤️ 

Also, min height is a great addition to Verse even without background image! 

Screenshot 2024-06-12 at 3 33 32 PM
Screenshot 2024-06-12 at 3 33 41 PM

PR #62504, opened 2024-06-12
Add entity type for individual Term objects

fabiankaegy:
## What?

Adds a new `Term` type to the available entity types exported from the `core-data` package.

## Why?

' getEntityRecord' is commonly used to deal with individual terms outside of any taxonomy. Currently, we don't expose any types for it.

## How?

Adding the type to the correct folder and exporting it.

### DIFF
--- a/packages/core-data/src/entity-types/term.ts
+++ b/packages/core-data/src/entity-types/term.ts
@@ -1,0 +1,53 @@
+/**
+ * Internal dependencies
+ */
+import type { Context, ContextualField, OmitNevers } from './helpers';
+
+import type { BaseEntityRecords as _BaseEntityRecords } from './base-entity-records';
+
+declare module './base-entity-records' {
+	export namespace BaseEntityRecords {
+		export interface Term< C extends Context > {
+			/**
+			 * A human-readable description of the term.
+			 */
+			description: ContextualField< string, 'view' | 'edit', C >;
+			/**
+			 * The title for the term.
+			 */
+			name: string;
+			/**
+			 * An alphanumeric identifier for the term.
+			 */
+			slug: string;
+			/**
+			 * The number of objects with the term.
+			 */
+			count: number;
+			/**
+			 * The ID for the term.
+			 */
+			id: number;
+			/**
+			 * The link for the term.
+			 */
+			link: string;
+			/**
+			 * Meta fields for the term.
+			 */
+			meta: { [ key: string ]: unknown };
+			/**
+			 * The parent term ID.
+			 */
+			parent: number;
+			/**
+			 * The term taxonomy slug.
+			 */
+			taxonomy: string;
+		}
+	}
+}
+
+export type Term< C extends Context = 'edit' > = OmitNevers<
+	_BaseEntityRecords.Term< C >
+>;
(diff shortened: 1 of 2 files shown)


Mamaduka:
Unfortunately, I'm not the best person to review TS changes.

cc @tyxla, @jsnajdr

packages/core-data/src/entity-types/term.ts
```diff
@@ -0,0 +1,53 @@
...
+			 */
+			id: number;
+			/**
+			 * The link for the term.
+			 */
+			link: string;
+			/**
+			 * Meta fields for the term.
+			 */
+			meta: { [ key: string ]: unknown };
```
tyxla, line 38: Isn't meta essentially an associative array of meta fields? And because of the database schema, they'll always end up strings?

Also it supports `view` and `edit` contexts, right?

```suggestion
			meta: ContextualField<
				Record< string, string >,
				'view' | 'edit',
				C
			>;
```

packages/core-data/src/entity-types/term.ts
```diff
@@ -0,0 +1,53 @@
...
+	export namespace BaseEntityRecords {
+		export interface Term< C extends Context > {
+			/**
+			 * A human-readable description of the term.
+			 */
+			description: ContextualField< string, 'view' | 'edit', C >;
+			/**
+			 * The title for the term.
+			 */
+			name: string;
```
tyxla, line 18: According to the [REST API docs](https://developer.wordpress.org/rest-api/reference/categories/#schema-name), this can be available in multiple contexts:

```suggestion
			name: ContextualField< string, 'view' | 'embed' | 'edit', C >;
```

packages/core-data/src/entity-types/term.ts
```diff
@@ -0,0 +1,53 @@
...
+			/**
+			 * An alphanumeric identifier for the term.
+			 */
+			slug: string;
```
tyxla, line 22: Same as above:

```suggestion
			slug: ContextualField< string, 'view' | 'embed' | 'edit', C >;
```

packages/core-data/src/entity-types/term.ts
```diff
@@ -0,0 +1,53 @@
...
+			/**
+			 * A human-readable description of the term.
+			 */
+			description: ContextualField< string, 'view' | 'edit', C >;
+			/**
+			 * The title for the term.
+			 */
+			name: string;
+			/**
+			 * An alphanumeric identifier for the term.
```
tyxla, line 20: I've always found the "alphanumeric" description ambiguous, considering that terms can also contain dashes.

packages/core-data/src/entity-types/term.ts
```diff
@@ -0,0 +1,53 @@
...
+			 */
+			slug: string;
+			/**
+			 * The number of objects with the term.
+			 */
+			count: number;
```
tyxla, line 26: I'd suggest auditing all fields for context. `count` seems to be available for `view` and `edit` and not for `embed`.

tyxla (commented):
This looks pretty good, thanks @fabiankaegy 🙌 

The primary thing that we should check is whether we're correctly providing the context of each field. Or perhaps you're intentionally skipping them? 

Other than that, I think this is good to go!

PR #62506, opened 2024-06-12
Support registering block style variations from the default `theme.json`

oandregal:
## What?

This PR adds support for registering block style variations from the default `theme.json`.

## Why?

We should aim to support any feature from all origins. `default` was missed, so we only supported the `theme` and `user` origins.

Note that the `blocks` origin is a bit different in that we only support styles specific to the block (no support for settings, etc.). This PR doesn't change that.

## How?

- Register the variations defined in the `default` origin in `gutenberg_register_block_style_variations_from_theme`.
- Hook into the `wp_theme_json_data_default` filter with the `gutenberg_resolve_block_style_variations_from_theme_json_default` callback that takes take of providing the styles for the variations defined by the `default` origin.

Note I've also renamed a couple of functions for clarity about its purpose.

## Testing Instructions

Using the TwentyTwentyFour theme, create the following setup:

- The `default` origin. Update `lib/theme.json` by adding it the following in `styles.blocks`:

```json
                       "variations": {
                               "FromGutenberg": {
                                       "blockTypes": [
                                               "core/group",
                                               "core/columns"
                                       ],
                                       "color": {
                                               "background": "red"
                                       }
                               }
                       },
 ```

- The `theme` origin. Update the TT4 `theme.json` file by adding it the following in `styles.blocks`:

```json
                       "variations": {
                               "FromTheme": {
                                       "blockTypes": [
                                               "core/group",
                                               "core/columns"
                                       ],
                                       "color": {
                                               "background": "blue"
                                       }
                               }
                       },
```

- The `user` origin. Update the TT4 `styles/ember.json` file by adding it the following in `styles.blocks`:

```json
                       "variations": {
                               "FromUser": {
                                       "blockTypes": [
                                               "core/group",
                                               "core/columns"
                                       ],
                                       "color": {
                                               "background": "yellow"
                                       }
                               }
                       },
```

- Partials. Create a new file at `styles/partials.json` with the following content:

```json
{
        "version": 2,
        "title": "FromPartial",
        "blockTypes": [ "core/group", "core/columns", "core/media-text" ],
        "styles": {
                "color": {
                        "background": "green"
                }
        }
}
```

Now, go to the site or post editors and verify that you have the following block style variations in the group block: `FromGutenberg`, `FromTheme`, `FromPartial` and that all work as expected:

Captura de ecrã 2024-06-12, às 11 40 45

You can also test that everything works as before:

- Go to the site editor, apply the `Ember` theme style variation. Verify that you now have the `FromUser` block style variation as well.
- Modifications to any of the block style variations done by the user are respected.

### DIFF
--- a/lib/block-supports/block-style-variations.php
+++ b/lib/block-supports/block-style-variations.php
@@ -299,7 +299,7 @@ function gutenberg_merge_block_style_variations_data( $variations_data, $theme_j
  *
  * @return WP_Theme_JSON_Data_Gutenberg
  */
-function gutenberg_resolve_block_style_variations_from_theme_style_variation( $theme_json ) {
+function gutenberg_resolve_block_style_variations_from_theme_json_user( $theme_json ) {
 	$theme_json_data   = $theme_json->get_data();
 	$shared_variations = $theme_json_data['styles']['blocks']['variations'] ?? array();
 	$variations_data   = gutenberg_resolve_block_style_variations( $shared_variations );
@@ -334,7 +334,25 @@ function gutenberg_resolve_block_style_variations_from_theme_json_partials( $the
  *
  * @return WP_Theme_JSON_Data_Gutenberg
  */
-function gutenberg_resolve_block_style_variations_from_primary_theme_json( $theme_json ) {
+function gutenberg_resolve_block_style_variations_from_theme_json_default( $theme_json ) {
+	$theme_json_data        = $theme_json->get_data();
+	$block_style_variations = $theme_json_data['styles']['blocks']['variations'] ?? array();
+	$variations_data        = gutenberg_resolve_block_style_variations( $block_style_variations );
+
+	return gutenberg_merge_block_style_variations_data( $variations_data, $theme_json, 'default' );
+}
+
+/**
+ * Merges shared block style variations registered within the
+ * `styles.blocks.variations` property of the primary theme.json file.
+ *
+ * @since 6.6.0
+ *
+ * @param WP_Theme_JSON_Data_Gutenberg $theme_json Current theme.json data.
+ *
+ * @return WP_Theme_JSON_Data_Gutenberg
+ */
+function gutenberg_resolve_block_style_variations_from_theme_json_theme( $theme_json ) {
 	$theme_json_data        = $theme_json->get_data();
 	$block_style_variations = $theme_json_data['styles']['blocks']['variations'] ?? array();
 	$variations_data        = gutenberg_resolve_block_style_variations( $block_style_variations );
@@ -394,8 +412,8 @@ if ( function_exists( 'wp_enqueue_block_style_variation_styles' ) ) {
 	remove_action( 'wp_enqueue_scripts', 'wp_enqueue_block_style_variation_styles' );
 }
 
-if ( function_exists( 'wp_resolve_block_style_variations_from_primary_theme_json' ) ) {
-	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_primary_theme_json' );
+if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_default' ) ) {
+	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_theme_json_default' );
 }
 if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_partials' ) ) {
 	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_theme_json_partials' );
@@ -403,8 +421,8 @@ if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_partial
 if ( function_exists( 'wp_resolve_block_style_variations_from_styles_registry' ) ) {
 	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_styles_registry' );
 }
-if ( function_exists( 'wp_resolve_block_style_variations_from_theme_style_variation' ) ) {
-	remove_filter( 'wp_theme_json_data_user', 'wp_resolve_block_style_variations_from_theme_style_variation' );
+if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_user' ) ) {
+	remove_filter( 'wp_theme_json_data_user', 'wp_resolve_block_style_variations_from_theme_json_user' );
 }
 
 // Add Gutenberg filters and action.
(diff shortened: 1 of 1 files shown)


oandregal:
This happened:

Captura de ecrã 2024-06-12, às 11 43 59

I don't know why github removed a code owner review request. I'm adding it back.

lib/block-supports/block-style-variations.php
```diff
@@ -490,25 +510,9 @@ function gutenberg_register_block_style_variations_from_theme() {
...
-	$variations_theme = $theme_json_theme->get_data()['styles']['blocks']['variations'] ?? array();
-	gutenberg_register_block_style_variations_from_theme_json_data( $variations_theme );
-
-	// User data linked for this theme.
-	$theme_json_user = WP_Theme_JSON_Resolver_Gutenberg::get_user_data();
-	$variations_user = $theme_json_user->get_data()['styles']['blocks']['variations'] ?? array();
-	gutenberg_register_block_style_variations_from_theme_json_data( $variations_user );
+	// theme.json (default, theme, user).
+	$variations = gutenberg_get_global_styles( array( 'blocks', 'variations' ) );
+	gutenberg_register_block_style_variations_from_theme_json_data( $variations );
```
oandregal, line 515: Besides supporting any feature from all origins, this is another advantage of supporting `default`: code simplification & performance. The merging already takes care of a lot of things, so we can work like there was a single `theme.json`.

I've noticed `wp/gutenberg_get_global_styles` is not cached, unlike the other `wp/gutenberg_get_*` methods. This should be addressed separately.

aaronrobertshaw:
Early on in the life of section styles there was some work done toward allowing origins to remove a block style variation. For example, whether a theme style variation could flag a shared variation definition for deregistration perhaps like:

```json
{
    "styles": {
        "blocks": {
            "variations": {
                "dark": false
            }
        }
    }
}
```

It was suggested that could be a future enhancement to the section styles feature. Do you foresee any issues only processing the merged data?

lib/block-supports/block-style-variations.php
```diff
@@ -394,17 +412,17 @@ function gutenberg_enqueue_block_style_variation_styles() {
 	remove_action( 'wp_enqueue_scripts', 'wp_enqueue_block_style_variation_styles' );
 }
 
-if ( function_exists( 'wp_resolve_block_style_variations_from_primary_theme_json' ) ) {
-	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_primary_theme_json' );
+if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_default' ) ) {
+	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_theme_json_default' );
```
aaronrobertshaw, line 416: I think there's a copy paste error here, it should be `wp_theme_json_data_default` right?

lib/block-supports/block-style-variations.php
```diff
@@ -334,7 +334,25 @@ function gutenberg_resolve_block_style_variations_from_theme_json_partials( $the
...
+	$theme_json_data        = $theme_json->get_data();
+	$block_style_variations = $theme_json_data['styles']['blocks']['variations'] ?? array();
+	$variations_data        = gutenberg_resolve_block_style_variations( $block_style_variations );
+
+	return gutenberg_merge_block_style_variations_data( $variations_data, $theme_json, 'default' );
+}
+
+/**
+ * Merges shared block style variations registered within the
+ * `styles.blocks.variations` property of the primary theme.json file.
```
aaronrobertshaw, line 347: If we needed to rename the functions for clarity, perhaps the reference to "primary theme.json file" here also needs updating?

lib/block-supports/block-style-variations.php
```diff
@@ -299,7 +299,7 @@ function gutenberg_merge_block_style_variations_data( $variations_data, $theme_j
  *
  * @return WP_Theme_JSON_Data_Gutenberg
  */
-function gutenberg_resolve_block_style_variations_from_theme_style_variation( $theme_json ) {
+function gutenberg_resolve_block_style_variations_from_theme_json_user( $theme_json ) {
```
aaronrobertshaw, line 302: Do we need to tweak the docblock to reflect this is more than just a theme style variation as well i.e. it also includes user definitions?

lib/block-supports/block-style-variations.php
```diff
@@ -394,17 +412,17 @@ function gutenberg_enqueue_block_style_variation_styles() {
...
 if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_partials' ) ) {
 	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_theme_json_partials' );
 }
 if ( function_exists( 'wp_resolve_block_style_variations_from_styles_registry' ) ) {
 	remove_filter( 'wp_theme_json_data_theme', 'wp_resolve_block_style_variations_from_styles_registry' );
 }
-if ( function_exists( 'wp_resolve_block_style_variations_from_theme_style_variation' ) ) {
-	remove_filter( 'wp_theme_json_data_user', 'wp_resolve_block_style_variations_from_theme_style_variation' );
+if ( function_exists( 'wp_resolve_block_style_variations_from_theme_json_user' ) ) {
+	remove_filter( 'wp_theme_json_data_user', 'wp_resolve_block_style_variations_from_theme_json_user' );
```
aaronrobertshaw, line 425: I think we're now missing a removal of a core filter. With this rename, we remove 4 core filters but we have 5, at least we will have 5 once the `default` origin change is backported.

aaronrobertshaw (commented):
Thanks for putting this together @oandregal 👍 

I've only really had a chance to read through the code so far but I've left a few minor suggestions. From what I can see there's only one real issue that needs fixing i.e. a missing removal of a core filter (once this is backported).

I'll give this a more thorough test shortly but a quick smoke test looks promising.

aaronrobertshaw:
This is tests as advertised for me, nice work ✨ 

I did encounter one issue that I believe is unrelated to this PR that we might be able to follow-up on separately. 

The issue I found relates to a block style variation persisting as an option in the block inspector's styles panel when switching back to the default theme style variation option or resetting styles.

To replicate:
1. Set up variationls as per this PR's test instructions
2. In the site editor switch to the ember theme style variation. You'll get the fromUser variation as an option
3. Switch back to the default theme style variation i.e. Styles > Browse Styles > Default
4. Select a group block and view its inspector styles panel, the fromUser variation will still be present
5. Navigate to the group block types styles and fromUser will correctly not appear in the configurable variations: Styles > Blocks > Group

There's a difference in how the available styles are retrieved. The global styles [variations panel](https://github.com/WordPress/gutenberg/blob/1e829be3c7245ba5eb7a39974f47179eecfdd04c/packages/edit-site/src/components/global-styles/variations/variations-panel.js#L40) retrieves any from the blocks store via `getBlockStyles` then filters by the style's source (`block` or included in variation styles).  The block inspectors panel retrieves styles only via the `getBlockStyles` selector within [`useStylesForBlocks`](https://github.com/WordPress/gutenberg/blob/1e829be3c7245ba5eb7a39974f47179eecfdd04c/packages/block-editor/src/components/block-styles/use-styles-for-block.js#L54) and does not filter them. 

I expect the lack of filtering in the block inspector panel is due to that being used in the post editor where until a couple of weeks ago it didn't have any access to global styles data. When I get a chance, I'll take a run at a quick fix, to filter by the global styles variations using the new global styles data in the editor settings.

oandregal:
I talked to Aaron by DM about direction for this and it sounds that we shouldn't do this — at least in the 6.6 cycle. There's conflicting opinions on whether the `default` origin should support this feature. Let's punt for later (closing, it can be reopened when required).

PR #62507, opened 2024-06-12
Fix: Comments and Pingbacks get cleared out when a pattern is edited.

### LINKED ISSUE #61827: Comments and Pingbacks get cleared out when a pattern is edited (within the editor) and a metabox is shown
### Description

When editing a post, if we also edit a pattern from within the post, the comment_status and pingbacks get turned off.

This happens only (from my tests) when a metabox is added / showing on the editor (example: the custom fields metabox or the Yoast SEO one).

1. Edit or create a post (Gutenberg)
2. Make sure you have Custom Fields showing on the editor (hamburguer menu, preferences, advanced , custom fields)
3. If there's no pattern on that post yet, create one (that syncs across posts)
4. Make sure the posts and pingbacks are turned on.
5. Edit the pattern and save it
6. Head back to the Post editor. If you refresh the page here you'll notice that the comments and pingbacks are turned off. Same happens if instead of refreshing you just "Save Draft" or Update Post. This means somehting is happening in the backend and changing these fields without updating the UI (only when we get new data from the server, do those fields show as off).

Screen Capture on 2024-05-21 at 16-37-38

### Environment info

This can be replicated with a fresh WordPress installation or WordPress playground without plugins installed.

This can't be replicated in older versions (before 6.5 - I think), seems the issue is present only when we edit the pattern on the post editor itself.

Yes

Yes

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/61827
cc: @mlaetitia

Right now, if meta boxes and custom fields are shown when we save a pattern in a post the comments and pingbacks are disabled. This PR fixes the issue.

The issue happens because when save is pressed and metaboxes/custom fields are enabled we need to trigger a post request saving them and for backwards compatibility comments and pingbacks settings are added in that form.
We were retrieving the comments and pingbacks settings from the current post in the editor. That is wrong the current post may be a pattern that has comments and pingbacks disabled. Instead of using the current post in the editor, we need to retrieve the settings from the post from which meta boxes are being saved, normally the original post opened in the editor.

## Testing Instructions
Edit or create a post (Gutenberg)
Make sure you have Custom Fields showing on the editor (hamburguer menu, preferences, advanced , custom fields)
If there's no pattern on that post yet, create one (that syncs across posts)
Make sure the posts and pingbacks are turned on.
Edit the pattern and save it
Head back to the Post editor. If you refresh the page here you'll notice that the comments and pingbacks are turned on ( on trunk they got disabled).

### DIFF
--- a/packages/edit-post/src/store/actions.js
+++ b/packages/edit-post/src/store/actions.js
@@ -9,6 +9,7 @@ import {
 } from '@wordpress/editor';
 import deprecated from '@wordpress/deprecated';
 import { addFilter } from '@wordpress/hooks';
+import { store as coreStore } from '@wordpress/core-data';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


carolinan (approved):
I was able to reproduce the issue, and the PR solves it. The correct comment and ping status is updated.

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 5906113de4

PR #62508, opened 2024-06-12
Ensure that openref is defined before accessing to .current

gigitux:
## What?

This PR improves a check to ensure that JS errors aren't triggered. You can see more details at: https://github.com/woocommerce/woocommerce/issues/48308

## Why?

By default, InsertionPointOpenRef is defined as an empty context in Gutenberg. You can find the context definition [here](https://github.com/WordPress/gutenberg/blob/c06914890b23b26542a616487fde6a6b6a09df8b/packages/block-editor/src/components/block-tools/index.js).

When Gutenberg is used as a framework, the BlockTools component ([that defined the context](https://github.com/WordPress/gutenberg/blob/c06914890b23b26542a616487fde6a6b6a09df8b/packages/block-editor/src/components/block-tools/index.js/#L192-L193)) may not be utilized in some projects (e.g., Customize Your Store project). In such cases, JavaScript will throw an error at the following line: [use-in-between-inserter.js#L43](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/block-list/use-in-between-inserter.js#L43).

To prevent this error, we need to update the check for InsertionPointOpenRef to handle scenarios where the context isn't defined. This PR will ensure that the context is defined, avoiding runtime errors.

### DIFF
--- a/packages/block-editor/src/components/block-list/use-in-between-inserter.js
+++ b/packages/block-editor/src/components/block-list/use-in-between-inserter.js
@@ -40,6 +40,8 @@ export function useInBetweenInserter() {
 			}
 
 			function onMouseMove( event ) {
+				// openRef is the reference to the insertion point between blocks.
+				// If the reference is not set or the insertion point is already open, return.
 				if ( openRef === undefined || openRef.current ) {
 					return;
 				}
--- a/packages/block-editor/src/components/block-list/use-in-between-inserter.js
+++ b/packages/block-editor/src/components/block-list/use-in-between-inserter.js
@@ -40,7 +40,7 @@ export function useInBetweenInserter() {
 			}
 
 			function onMouseMove( event ) {
-				if ( openRef.current ) {
+				if ( openRef === undefined || openRef.current ) {
 					return;
 				}
 


gigitux:
If this PR makes sense, it would be great if it could be backported to WordPress 6.6 🙇

packages/block-editor/src/components/block-list/use-in-between-inserter.js
```diff
@@ -40,7 +40,7 @@ export function useInBetweenInserter() {
 			}
 
 			function onMouseMove( event ) {
-				if ( openRef.current ) {
+				if ( openRef === undefined || openRef.current ) {
```
cbravobernal, line 45: I think this would work too.
```suggestion
				if ( openRef?.current ) {
```

gigitux: It was my first commit :D 

However, with this code, the function will not return at line 44, but will continue with the logic.

BTW, I don't have a strong opinion! What do you think?

vcanales: These are different checks:

1. The proposed logic in the PR is bail on `openRef === undefined` OR if `openRef.current` is truthy, which I'm not sure is correct.
2. @cbravobernal's suggestion bails on openRef truthy, but `openRef === undefined` keeps going.

@gigitux Could you explain why the handler should bail on `openRef === undefined`?

cbravobernal: > It was my first commit :D
> 
> However, with this code, the function will not return at line 44, but will continue with the logic.
> 
> BTW, I don't have a strong opinion! What do you think?

Agh, we made the same mistake then 😆 

Return means not showing the in-between inserter. openRef seems to be a reference of where the next block should be insterted.

So, if the reference of the next position to be insterted does not exist (is undefined), the indicator should be hidden.

PD: I cannot get `openRef.current` to return true while testing on a post editor.

gigitux: @vcanales Good point! Thanks for clarifying this! My first intuition was the right one. Sometimes overthinking isn't a good idea 😅

vcanales: Oh now I understand, thanks for the explanation @cbravobernal
@gigitux I think `openRef === undefined || openRef.current` is the correct check, since we don't want to show the inserter when the next ref is undefined.

vcanales (commented):
Could I ask you to also add a comment to the handler explaining its flow a little bit?

@cbravobernal's comment explains part of why we're returning early, but I still find it hard to grasp at a glance by just looking at the code

> Return means not showing the in-between inserter. openRef seems to be a reference of where the next block should be insterted.
>
> So, if the reference of the next position to be insterted does not exist (is undefined), the indicator should be hidden.

packages/block-editor/src/components/block-list/use-in-between-inserter.js
cbravobernal: This is from where I understood that returning is hiding the inserter btw:
https://github.com/WordPress/gutenberg/blob/2d34e2aaa56e52c379b01bed2df2344ba01c9bb4/packages/block-editor/src/components/block-list/use-in-between-inserter.js#L132

gigitux:
Some clarifications: it looks like this handler is used to render the inserter between blocks.

This is the default behavior:


https://github.com/WordPress/gutenberg/assets/4463174/d5b54b6e-a78d-4c6d-8235-553bd44ad3a3

This is the behavior if I define the openRef as true. The inserter between blocks isn't rendered:


https://github.com/WordPress/gutenberg/assets/4463174/52f8ad42-3a5a-4f57-96b1-e2ff148f3409

Now, this hook works only if the component is inside the `InsertionPointOpenRef.Provider` hierarchy. For this reason, I think that it doesn't make sense to continue with the flow in case `openRef` is undefined. 


I added a comment with this commit 2ea0e3e2d7d7d0f13107e6083d3386443fec0fa1, but I'm open to feedback.


Thanks for the review and help! 🙇

vcanales (approved):
Looks good. Thanks for illustrating and documenting too!

cbravobernal (approved):
Thanks for the clarification @gigitux !

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 49cc18e017

PR #62509, opened 2024-06-12
Update cover block min height control to use HeightControl 

### LINKED ISSUE #56020: Update cover block min height control to use HeightControl
We should make the "Min. height" control for the core/cover block match the implementation within Dimensions controls (the core/group block for example). 

The Cover block is using a `UnitControl`, but it should use a `HeightControl`, so that it looks and feels the same. We can probably also reduce some of the custom height calculation logic in the Cover block so that the function is consistent across blocks. 

| Current  | Expected |
|CleanShot 2023-11-09 at 17 43 56|CleanShot 2023-11-09 at 17 41 39|

akasunil:
## What?

Fixes: https://github.com/WordPress/gutenberg/issues/56020

This PR when applied,
Replace `UnitControl` component with `HeightControl` in cover block
 
## Screenshots or screencast 
image


akasunil:
@richtabor Should we replace height control label "Minimum height of cover" with "Minimum height"?

richtabor:
> @richtabor Should we replace height control label "Minimum height of cover" with "Minimum height"?

Yes, thanks!

akasunil:
Hi @richtabor, Could you please review this PR?

jasmussen (approved):
This could use a code review, but this definitely seems to be a better solution than https://github.com/WordPress/gutenberg/pull/62742 which was just a half-step. 

@WordPress/gutenberg-design what do you think?

jameskoster:
Seems good to me. 

I'm curious about `HeightControl` though. I suspect it has applications beyond only height... should it have a more generic name?

richtabor:
> Seems good to me.
> 
> I'm curious about `HeightControl` though. I suspect it has applications beyond only height... should it have a more generic name?

This is the same component used in the dimensions panel ([source](https://github.com/WordPress/gutenberg/blob/e618232ef535f733c242d2326c7061f3df7be3f2/packages/block-editor/src/components/global-styles/dimensions-panel.js#L673)).

richtabor:
| Before  | After |
|CleanShot 2024-06-25 at 09 00 43|CleanShot 2024-06-25 at 09 00 05|

jameskoster:
It was more a meta comment; I could see that component (unit control + range) being used for other numerical settings. E.g. width, margin, padding, etc. In fact Padding already uses a custom version of this pattern:

Screenshot 2024-06-25 at 14 28 04

It would be good to tidy that up (not in this PR, obviously) :)

akasunil:
We could rename it to something like `RangeInputControl`, a self explanatory.

richtabor:
> We could rename it to something like RangeInputControl, a self explanatory.

Maybe, let's consider that for a follow-up though.

test/e2e/specs/editor/blocks/cover.spec.js
```diff
@@ -156,11 +156,11 @@ test.describe( 'Cover', () => {
...
 			.click();
 
-		// Ensure there the default value for the minimum height of cover is undefined.
+		// Ensure there the default value for the Minimum height is undefined.
 		const defaultHeightValue = await coverBlockEditorSettings
-			.getByLabel( 'Minimum height of cover' )
+			.getByRole( 'slider', { name: 'Minimum height' } )
 			.inputValue();
-		expect( defaultHeightValue ).toBeFalsy();
+		expect( defaultHeightValue ).toEqual( '0' );
```
t-hamano: This test is trying to verify that the minimum height is "undefined", but it changes the expected result.

I think the actual value being checked is a field for entering a number, not a slider.

Can't the following code achieve the same thing without changing the expected result?

```javascript
const defaultHeightValue = await coverBlockEditorSettings
	.getByRole( 'spinbutton', { name: 'Minimum height' } )
	.inputValue();
expect( defaultHeightValue ).toBeFalsy();
```

packages/block-library/src/cover/test/edit.js
```diff
@@ -371,13 +371,12 @@ describe( 'Cover block', () => {
...
-				await userEvent.clear(
-					screen.getByLabelText( 'Minimum height of cover' )
-				);
-				await userEvent.type(
-					screen.getByLabelText( 'Minimum height of cover' ),
-					'300'
-				);
+
+				const [ heightControl ] =
+					screen.getAllByLabelText( /Minimum height/ );
```
t-hamano:
```suggestion
				const heightControl = screen.getByRole( 'spinbutton', {
					name: 'Minimum height',
				} );
```

It is better to [use accessible selectors](https://github.com/WordPress/gutenberg/blob/fix/issue-56020/docs/contributors/code/e2e/README.md#use-accessible-selectors).

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -42,46 +39,29 @@ function CoverHeightInput( {
...
-		availableUnits: availableUnits || [ 'px', 'em', 'rem', 'vw', 'vh' ],
-		defaultValues: { px: 430, '%': 20, em: 20, rem: 20, vw: 20, vh: 50 },
-	} );
-
 	const handleOnChange = ( unprocessedValue ) => {
-		const inputValue =
-			unprocessedValue !== ''
-				? parseFloat( unprocessedValue )
-				: undefined;
+		// eslint-disable-next-line @wordpress/no-unused-vars-before-return
```
t-hamano: I'd like to avoid disabling ESLint errors if possible. Can I use logic like the following to deal with this?

```javascript
const parsedQuantityAndUnit =
	parseQuantityAndUnitFromRawValue( unprocessedValue );
const currentValue = parsedQuantityAndUnit[ 0 ];

if ( isNaN( currentValue ) && currentValue !== undefined ) {
	return;
}

const currentUnit = parsedQuantityAndUnit[ 1 ];

onUnitChange( currentUnit );
onChange( currentValue );
```

This is similar to [this part](https://github.com/WordPress/gutenberg/blob/7586f2f0904f64ab64d63d4d1853981a530c9f34/packages/components/src/unit-control/test/index.tsx#L41-L49).

t-hamano (commented):
Thanks for the PR! Overall it looks good.

We might be able to migrate to the block support (`supports.dimensions.minHeight`) in the future, but we'd need to add [Block deprecation](https://github.com/WordPress/gutenberg/blob/trunk/docs/reference-guides/block-api/block-deprecation.md), which would be a big change, so for now I think this approach is fine.

t-hamano:
Sorry for the late response.

I thought we could ship this PR, but there was one thing that bothered me.

That is, when I move the slider and it reaches zero, the height of the Cover block suddenly becomes higher.

https://github.com/WordPress/gutenberg/assets/54422211/519e430f-3146-43d1-9d86-017b99591514

This is based on the following two facts:

- If minHeight is zero, the style `min-height` is not generated ([source](https://github.com/WordPress/gutenberg/blob/4044d674a630d0618cd1460c9a348e6e17fac0af/packages/block-library/src/cover/save.js#L64-L66))
- This block has `min-height: 430px` by default ([Source](https://github.com/WordPress/gutenberg/blob/4044d674a630d0618cd1460c9a348e6e17fac0af/packages/block-library/src/cover/style.scss#L3))

In other words, the moment the slider reaches zero, `min-height: 430px` is applied.

To avoid this, we need to remove `undefined` [here](https://github.com/WordPress/gutenberg/blob/4044d674a630d0618cd1460c9a348e6e17fac0af/packages/block-library/src/cover/save.js#L65) so that `min-height: 0` is applied.

Changing this means that the output markup will change, so we need to add [Block Deprecation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/).

If we are going to add a block deprecation anyway, I think it might be better to migrate to block support now.

Specifically, the following measures are required:

- Remove the `minHeight` and `minHeihgtUnit` attributes from `block.json`, and remove the related code from `edit.js` and `save.js`.
- Add `dimensions.minHeight` support.
- Add Block Deprecation so that the `minHeight` and `minHeihgtUnit` attributes are migrated to `minHeihgt` as block support. In other words, the following migration is expected:

```html
  // From...
  <!-- wp:cover {"minHeight":0,"minHeightUnit":"px","layout":{"type":"constrained"}} -->
  <div class="wp-block-cover">
  	<span aria-hidden="true" class="wp-block-cover__background has-contrast-2-background-color has-background-dim-100 has-background-dim"></span>
  	<div class="wp-block-cover__inner-container"></div>
  </div>
  <!-- /wp:cover -->

  // To...
  <!-- wp:cover {"style":{"dimensions":{"minHeight":"0px"}},"layout":{"type":"constrained"}} -->
  <div class="wp-block-cover" style="min-height:0px">
  	<span aria-hidden="true" class="wp-block-cover__background has-contrast-2-background-color has-background-dim-100 has-background-dim"></span>
  	<div class="wp-block-cover__inner-container"></div>
  </div>
  <!-- /wp:cover -->
  ```

- Add fixtures to ensure the migrations is run

Would you like to try this approach?

t-hamano:
@akasunil 

> If we are going to add a block deprecation anyway, I think it might be better to migrate to block support now.
> Specifically, the following measures are required:

Sorry, I found this suggestion inappropriate.

`minHeight` as block support is `false` [by default](https://github.com/WordPress/gutenberg/blob/26a767a9a5552e774863d5d5651977be47738150/schemas/json/theme.json#L318). This means that this UI will disappear unintentionally in classic themes that don't have a `theme.json`, or in themes that don't have `dimensions.minHeight` enabled via `theme.json`.

After all, I think it's difficult to migrate to block support at this point.

Therefore, the additional approach I propose is as follows. Sorry for the confusion 🙇‍♂️

1. In `save.js`, we would probably make changes like the following to ensure that a zero `minHeight` is output as the inline style:
  ```javascript
  const minHeight =
  	minHeightProp !== undefined && minHeightUnit
  		? `${ minHeightProp }${ minHeightUnit }`
  		: undefined;
  
  const style = {
  	minHeight,
  };
  ```
2. Because the [`min` prop is defined](https://github.com/WordPress/gutenberg/blob/26a767a9a5552e774863d5d5651977be47738150/packages/block-library/src/cover/edit/inspector-controls.js#L79), we cannot set it to `0` via the UI, but we can set it to `0` via the code editor. To prevent this edge case from breaking, we add a block deprecation. In other words, the following migration is expected:
  ```html
  <!-- From... -->
  <!-- wp:cover {"minHeight":0,"minHeightUnit":"px","layout":{"type":"constrained"}} -->
  <div class="wp-block-cover">
  	<span aria-hidden="true" class="wp-block-cover__background has-background-dim-100 has-background-dim"></span>
  	<div class="wp-block-cover__inner-container">
  		<!-- wp:paragraph -->
  		<p></p>
  		<!-- /wp:paragraph -->
  	</div>
  </div>
  <!-- /wp:cover -->

  <!-- To... -->
  <!-- wp:cover {"minHeight":0,"minHeightUnit":"px","layout":{"type":"constrained"}} -->
  <div class="wp-block-cover" style="min-height:0px">
  	<span aria-hidden="true" class="wp-block-cover__background has-background-dim-100 has-background-dim"></span>
  	<div class="wp-block-cover__inner-container">
  		<!-- wp:paragraph -->
  		<p></p>
  		<!-- /wp:paragraph -->
  	</div>
  </div>
  <!-- /wp:cover -->
  ```
3. Add fixtures to ensure the migrations is run

t-hamano:
@aaronrobertshaw @andrewserong 

As we move forward with this PR, I'd appreciate some feedback on whether this approach in [this comment](https://github.com/WordPress/gutenberg/pull/62509#issuecomment-2205522700) makes sense 🙇‍♂️

The history of this PR can be summarized as follows:

- The original purpose of this PR was to replace the `UnitControl` component with the `HeightControl` component to improve consistency of the Min Height UI.
- The `HeightControl` component allows a zero value. Because the zero value does not output an inline style, I found an issue with the block height suddenly changing, as mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/62509#issuecomment-2203278584).
- To prevent this, I think we need to update the `save.js` function so that `min-height: 0` is output as an inline style.
- Since we have updated the `save.js` function, we need to add the block deprecation.
- I thought that if we were going to add block deprecation anyway, it would be better to switch to block support instead. However, because this support is disabled by default, it is not currently possible to migrate to block support in order to allow themes that do not have this support enabled to continue using this UI.

t-hamano (changes requested):
I'm blocking this PR for now as we need to determine the ideal approach.

richtabor:
Why not just set the minimum value of the RangeControl to 50, which is what it is effectively in trunk? That would be 1:1 to supporting today's experience with the new component, right? If you try to set `0` in trunk, it jumps to `50`.

akasunil:
> Why not just set the minimum value of the RangeControl to 50

We could do that, what do you say @t-hamano?

t-hamano:
> Why not just set the minimum value of the RangeControl to 50

This approach might also be possible.

However, the `HeightControl` component [does not support the `min` prop.](https://github.com/WordPress/gutenberg/blob/4044d674a630d0618cd1460c9a348e6e17fac0af/packages/block-editor/src/components/height-control/index.js#L66-L68) Therefore, we will need to enforce the minimum value in `handleOnChange`. Also, note that the minimum value of `50` is only enforced when the unit is `px`.

richtabor:
Is there any way forward on this effort, or should we close this?

aaronrobertshaw:
> Is there any way forward on this effort, or should we close this?

@t-hamano and @akasunil are probably best placed to make the call however my read of the discussion so far is that this PR just needs updating to the settled on [approach](https://github.com/WordPress/gutenberg/pull/62509#issuecomment-2205522700) (including the [50px min tweak?](https://github.com/WordPress/gutenberg/pull/62509#issuecomment-2226671651)).

> However, the HeightControl component [does not support the min prop.](https://github.com/WordPress/gutenberg/blob/4044d674a630d0618cd1460c9a348e6e17fac0af/packages/block-editor/src/components/height-control/index.js#L66-L68)

That component is only in the block-editor package so might have more freedom to be updated. I may be misremembering but at one stage I thought we had some controls that allowed minimum values defined per unit e.g. px vs em etc. A similar thing could be accepted in that new prop.

> However, because this support is disabled by default, it is not currently possible to migrate to block support in order to allow themes that do not have this support enabled to continue using this UI.

This might not be a solution but the theme.json data is filterable. The min height block support could also be enabled by default for the Cover block type within the default theme.json file. So by that I mean, it stays as `false` by default overall, it just gets enabled by default for the Cover block only.

That might be worth exploring if we want to consolidate around block supports and limit deprecations.

akasunil:
> I may be misremembering but at one stage I thought we had some controls that allowed minimum values defined per unit e.g. px vs em etc. A similar thing could be accepted in that new prop.

That was the `UnitControl` currently [implemented in cover](https://github.com/WordPress/gutenberg/pull/62509/files#diff-4af4f90d0daaf9ff4f7c1836fe266febc31fac60488d8084b942cb7f62cb2a7bL50-L53) block. We can update heightControl component to support min value. 

In trunk: 
- If we set min-height to 0 value with PX unit, it fall back to 50.
- If we set min-height to 0 value to any other unit, It stay 0 in control, But doesn't output min-height value, it fall back to min-height from styles.

in this PR:
- If we set min-height to 0 value with PX unit or any other unit, It stay 0 in control, But doesn't output min-height value, it fall back to min-height from styles.

Difference here i could find is only min value for `heightControl` with PX unit. If we implement this minimum value support in `heightControl` component, i don't think we need any deprecation here. I know there still the issue of sudden change in cover block height when value reach to zero. but that also already exists in trunk currently. For which follow-up PR with deprecation or new issue for discussion can be created.

> This might not be a solution but the theme.json data is filterable. The min height block support could also be enabled by default for the Cover block type within the default theme.json file. So by that I mean, it stays as `false` by default overall, it just gets enabled by default for the Cover block only.
> 
> That might be worth exploring if we want to consolidate around block supports and limit deprecations.

This is something can be explored but, it would required PHP change for filter which required back-porting, we will also need deprecation for this too.

PR #62512, opened 2024-06-12
mess with focus

scruffian:
This is an attempt to address https://github.com/WordPress/gutenberg/issues/62441

My observation is that sometings focussing a tab with a script triggers focus visible and other times it doesn't. I thought there could be several reasons for this:

1. The tab component gets remounted every time the inserter opens
2. The focus happens on a request animation frame

This PR tries to change those things to see if it makes a difference but it doesn't.


scruffian:
Fix in https://github.com/WordPress/gutenberg/pull/62513

Issue #60619, opened 2024-04-10
Come up with a way to extract pattern-override code from the block-editor package


kevin940726:
## What problem does this address?

Originally reported in https://github.com/WordPress/gutenberg/pull/60234#discussion_r1547201296.

Pattern Overrides is a WordPress-only feature. It would be nice to extract the logic and code from the block-editor package to maybe the `editor` package instead.

## What is your proposed solution?

I haven't figured out an elegant way yet but I have some ideas.

Since that we probably shouldn't have multiple menu items with the same label, maybe we can use the label as a key to override the existing ones. For instance, the block-editor package would implement the `Rename` menu item normally without the overrides warning, and we'll give it a `rename` key.

When we try to add a new menu item with the same `rename` key in the `editor` package, then it would simply override the existing ones. This will allow us to implement the warning message there, without leaking WordPress-specific logic to the block-editor package.

kevin940726:
Another potential improvement is the pattern overrides toolbar indicator reported here: https://github.com/WordPress/gutenberg/pull/62317#discussion_r1631024218

talldan:
Superceded by https://github.com/WordPress/gutenberg/issues/65486 (it proposes doing the opposite, and considering pattern overrides as a block editor first feature).

PR #62514, opened 2024-06-12
Extract the pattern overrides toolbar indicator from the `block-editor` package

kevin940726:
## What?

Part of https://github.com/WordPress/gutenberg/issues/60619. Address the issue discussed in https://github.com/WordPress/gutenberg/pull/62317#discussion_r1631024218.

Extract out pattern overrides related code from the `block-editor` package to the `patterns` package, specifically the toolbar indicator.

## Why?

Pattern Overrides is a WP-only feature so ideally it doesn't belong in the `block-editor` package.

## How?

Create a `PatternOverridesBlockControls` component and conditionally render it using the pattern-overrides hook.

## Testing Instructions

1. Create a pattern with overrides and insert it into a post.
2. Select an overridable block and expect the toolbar indicator to show (see below image).
3. Select multiple overridable blocks and expect the toolbar indicator to show (see below image).
4. Selecting blocks inside the pattern focus mode should show a different indicator (see below image).

## Screenshots or screencast 

Select a single pattern overrides block:
image

Select multiple pattern overrides blocks:
image

Select a block inside the pattern focus mode:
image


packages/block-editor/src/components/pattern-overrides-toolbar-indicator/index.js
kevin940726, line 1: Moved to `packages/patterns/src/components/pattern-overrides-block-controls.js`.

packages/block-editor/src/components/pattern-overrides-toolbar-indicator/style.scss
kevin940726, line 1: Moved to `packages/patterns/src/components/style.scss`.

packages/editor/src/hooks/pattern-overrides.js
```diff
@@ -43,6 +44,7 @@ const withPatternOverrideControls = createHigherOrderComponent(
 				{ props.isSelected && isSupportedBlock && (
 					<ControlsWithStoreSubscription { ...props } />
 				) }
+				{ isSupportedBlock && <PatternOverridesBlockControls /> }
```
talldan, line 47: It's a shame there's no way to add a condition like `( props.isSelected || props.isMultiselected )` here

talldan (approved):
Code looks good and it tests well, thanks for addressing this.

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 8d8f6818b4

packages/patterns/src/components/pattern-overrides-block-controls.js
```diff
@@ -0,0 +1,155 @@
...
+			);
+			const _hasParentPattern = selectedClientIds.every(
+				( clientId ) =>
+					getBlockParentsByBlockName( clientId, 'core/block', true )
+						.length > 0
+			);
+			return {
+				clientIds: selectedClientIds,
+				hasPatternOverrides: _hasPatternOverrides,
+				hasParentPattern: _hasParentPattern,
```
Mamaduka, line 144: I am unsure if this affects anything substantial, but these values will be `true` when `selectedClientIds` is an empty array. Do to how `Array.every` works in JS.

packages/editor/src/hooks/pattern-overrides.js
Mamaduka: We should probably add a comment there to avoid accidental regression.

PR #62521, opened 2024-06-12
Fix: Resetting template part causes notification saying it's been delete

### LINKED ISSUE #62458: Resetting template part causes snackbar notification saying it's been deleted
Using 6.6 beta 1, if you make a change to a theme provided template part and then reset those changes from data views, you will see a snackbar notice that says the template part has been deleted rather than reset:

https://github.com/WordPress/gutenberg/assets/26996883/6852a728-0d3e-4b2b-b0d4-347ec1fe422b

I can replicate this with 6.6 beta 1 with and without GB 18.5 installed.

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/62458.
cc: @annezazu 
Addresses an issue where for template parts when a revert happens we say they were deleted.

## Testing Instructions
With 2024 theme installed apply some changes to the header template part.
Go to the patterns section template part do do a reset.
Verify we say the template part was reset not deleted as before.

### DIFF
--- a/packages/editor/src/store/private-actions.js
+++ b/packages/editor/src/store/private-actions.js
@@ -370,6 +370,13 @@ export const revertTemplate =
 export const removeTemplates =
 	( items ) =>
 	async ( { registry } ) => {
+		const isResetting = items.every(
+			( item ) =>
+				!! item &&
+				( item.has_theme_file ||
+					( item.templatePart && item.templatePart.has_theme_file ) )
+		);
+
 		const promiseResult = await Promise.allSettled(
 			items.map( ( item ) => {
 				return registry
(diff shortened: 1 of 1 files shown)


annezazu:
Worked great! Tested for both resetting a template part and template for good measure. Video of template part: 

https://github.com/WordPress/gutenberg/assets/26996883/80753dba-9498-4242-a2ae-1358f6d4898f

packages/editor/src/store/private-actions.js
```diff
@@ -370,6 +370,14 @@ export const revertTemplate =
 export const removeTemplates =
 	( items ) =>
 	async ( { registry } ) => {
+		const isResetting = items.every(
+			( item ) =>
+				!! item &&
+				( item.has_theme_file ||
+					( item.templatePart && item.templatePart.has_theme_file ) )
+		);
+		console.log( { items } );
```
carolinan: I don't think the console.log is needed?

carolinan: Ah, I see that's what causing the static analysis error.

carolinan (approved):
The message uses the correct text when the theme template part is reset.
The PR can be merged when the tests pass.

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 0595c91481

PR #62522, opened 2024-06-12
Improve performance of `compute_style_properties` method.

spacedmonkey:
Improve logic in compute_style_properties to improve performance. This commit fixes a number of performance issues including.

- Expensive repeated call to `str_starts_with`.
- Return easy if not array.
- Repeated call to `strlen`.
- Move call to `get_property_value` where it is needed.
- Move call to implode to after empty check.

### Before
Screenshot 2024-06-12 at 22 52 46

### After
Screenshot 2024-06-12 at 22 55 05

### DIFF
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -2262,7 +2262,7 @@ class WP_Theme_JSON_Gutenberg {
 		}
 		$declarations             = array();
 		$root_variable_duplicates = array();
-		$string_length            = strlen( '--wp--style--root--' );
+		$root_style_length        = strlen( '--wp--style--root--' );
 
 		foreach ( $properties as $css_property => $value_path ) {
 			if ( ! is_array( $value_path ) ) {
@@ -2283,7 +2283,7 @@ class WP_Theme_JSON_Gutenberg {
 			}
 
 			if ( $is_root_style && $use_root_padding ) {
-				$root_variable_duplicates[] = substr( $css_property, $string_length );
+				$root_variable_duplicates[] = substr( $css_property, $root_style_length );
 			}
 
 			// Processes background styles.
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -2253,43 +2253,37 @@ class WP_Theme_JSON_Gutenberg {
 	 * @return array  Returns the modified $declarations.
 	 */
 	protected static function compute_style_properties( $styles, $settings = array(), $properties = null, $theme_json = null, $selector = null, $use_root_padding = null ) {
+		if ( empty( $styles ) ) {
+			return array();
+		}
+
 		if ( null === $properties ) {
 			$properties = static::PROPERTIES_METADATA;
 		}
-
-		$declarations = array();
-		if ( empty( $styles ) ) {
-			return $declarations;
-		}
-
+		$declarations             = array();
 		$root_variable_duplicates = array();
+		$string_length            = strlen( '--wp--style--root--' );
 
 		foreach ( $properties as $css_property => $value_path ) {
-			$value = static::get_property_value( $styles, $value_path, $theme_json );
-
-			if ( str_starts_with( $css_property, '--wp--style--root--' ) && ( static::ROOT_BLOCK_SELECTOR !== $selector || ! $use_root_padding ) ) {
+			if ( ! is_array( $value_path ) ) {
 				continue;
 			}
+
+			$is_root_style = str_starts_with( $css_property, '--wp--style--root--' );
+			if (  $is_root_style && ( static::ROOT_BLOCK_SELECTOR !== $selector || ! $use_root_padding ) ) {
+				continue;
+			}
+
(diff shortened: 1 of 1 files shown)


lib/class-wp-theme-json-gutenberg.php
```diff
@@ -2253,43 +2253,37 @@ protected static function flatten_tree( $tree, $prefix = '', $token = '--' ) {
...
-			return $declarations;
+			return array();
 		}
 
+		if ( null === $properties ) {
+			$properties = static::PROPERTIES_METADATA;
+		}
+		$declarations             = array();
 		$root_variable_duplicates = array();
+		$string_length            = strlen( '--wp--style--root--' );
```
joemcgill: This variable could be more descriptive. Would need to update usages below.

```suggestion
		$root_style_length        = strlen( '--wp--style--root--' );
```

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -2304,6 +2298,16 @@ protected static function compute_style_properties( $styles, $settings = array()
 				continue;
 			}
 
+			// Look up protected properties, keyed by value path.
+			// Skip protected properties that are explicitly set to `null`.
+			$path_string = implode( '.', $value_path );
```
joemcgill, line 2303: Do we need to retain the `is_array()` check here?

joemcgill (commented):
This looks like a good refactor to me. Left a couple of comments inline.

lib/class-wp-theme-json-gutenberg.php
spacedmonkey: This is moved to line 2268.

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -2253,43 +2253,37 @@ protected static function flatten_tree( $tree, $prefix = '', $token = '--' ) {
...
-			if ( is_array( $value_path ) ) {
-				$path_string = implode( '.', $value_path );
-				if (
-					isset( static::PROTECTED_PROPERTIES[ $path_string ] ) &&
-					_wp_array_get( $settings, static::PROTECTED_PROPERTIES[ $path_string ], null ) === null
-				) {
-					continue;
-				}
+			if ( $is_root_style && $use_root_padding ) {
+				$root_variable_duplicates[] = substr( $css_property, $string_length );
```
spacedmonkey:
```suggestion
				$root_variable_duplicates[] = substr( $css_property, $root_style_length );
```

lib/class-wp-theme-json-gutenberg.php
joemcgill: Oh, I see. I don't know enough about this function, but is it possible that `$value_path` is sometimes a string instead of an array? If so, previously those would have still been processed where now they would not because of the early `continue` at line :2268.

tellthemachines: > is it possible that $value_path is sometimes a string instead of an array?

I don't believe so. At least, `compute_style_properties` is never passed any value for `$properties` inside this class. Also, my testing shows that if `get_property_value` were passed a string instead of an array it would silently fail with no global styles being output, which would happen if $value_path were a string.

jonnynews: These line also just guesses that value_path is an array. 

```
if ( 'background' === $value_path[0] && isset( $styles['background'] ) ) {
```

If for some reason `$value_path` was a string, it would get the first character of that string. This seems weird and wrong to me.

I think ensuring this code is a type safe is a good thing, but if we want to be safe, then I can refactor and moving the logic down.

ellatrix:
The `Backport to WP Beta/RC` label can be removed here since this will need to be backported manually, not through packages. Once a backport PR is made, add a [changelog](https://github.com/WordPress/gutenberg/tree/trunk/backport-changelog) entry for 6.6 and we'll keep track of it that way.

spacedmonkey:
> The `Backport to WP Beta/RC` label can be removed here since this will need to be backported manually, not through packages. Once a backport PR is made, add a [changelog](https://github.com/WordPress/gutenberg/tree/trunk/backport-changelog) entry for 6.6 and we'll keep track of it that way.

That is not how that works. All PRs that need to be manaully ported over, have that label assigned and are added the backport issue. Has this been changed? Where is that documented?

ellatrix:
Sure, it's fine to leave the label. Most important is that there's a core PR.

spacedmonkey:
Core PR https://github.com/WordPress/wordpress-develop/pull/6707. Will update and commit once this PR is merged.

lib/class-wp-theme-json-gutenberg.php
joemcgill: With this question resolved, I think this is good to commit.

joemcgill (approved):

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 1116c38719

ellatrix:
Why did you merge this with failing tests? A backport changelog entry should have been added for this PR

PR #62523, opened 2024-06-12
Alternative: Escape in select mode moves focus to canvas wrapper

jeryj:
Alternative to https://github.com/WordPress/gutenberg/pull/62196. Builds off of work in https://github.com/WordPress/gutenberg/pull/62290.

An escape keypress previously [toggled the state between editing and select mode](https://github.com/WordPress/gutenberg/pull/58637/). This returns the behavior to the previous implementation of Escape clearing block focus. Now escape when in select mode focuses a Block canvas labelled wrapper element within the body element of the canvas.

## What?

- In select mode, Escape selects the wrapping "Block canvas"
- Shows a blue border when the canvas is focused

## Why?

- Allow block deselection using Escape
- Show focus on canvas element

## Testing Instructions

- Click a nested block
- Escape to enter select mode
- Escape to clear selected block and focus canvas

### DIFF
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -81,6 +81,7 @@ export default function BlockTools( {
 	} = useShowBlockTools();
 
 	const {
+		clearSelectedBlock,
 		duplicateBlocks,
 		removeBlocks,
 		replaceBlocks,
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/components/block-tools/index.js
```diff
@@ -152,6 +153,12 @@ export default function BlockTools( {
 				// block so that focus is directed back to the beginning of the selection.
 				// In effect, to the user this feels like deselecting the multi-selection.
 				selectBlock( clientIds[ 0 ] );
+			} else if ( clientIds.length === 1 ) {
+				event.preventDefault();
+				clearSelectedBlock();
+				__unstableContentRef?.current
+					?.querySelector( '[aria-label="Block canvas"]' )
```
ellatrix, line 160: What is `[aria-label="Block canvas"]`? I can't find such element in the DOM. What about focusing `__unstableContentRef`?

jeryj: ~~I think this was a bug due to rebasing. It should be sending focus to the region, as that seems to be the best one for screen reader announcements.~~

Edit: I'm thinking of the [other PR](https://github.com/WordPress/gutenberg/pull/62196). This was an alternative. I'll close this one.

jeryj:
Closing in favor of https://github.com/WordPress/gutenberg/pull/62196 to focus the editor region.

Issue #51460, opened 2023-06-13
Command centre button: missing focus style and incorrect labeling


afercia:
### Description

In the Site Editor, the button that opens the 'command centre' has a couple accessibility issues:

- It misses s focus style.
- The labelling doesn't tell anything about the button action. Actually, the button opens the command centre modal dialog. There's no text / aria-label or other labelling to communicate what this button does, not even visually. The actual labelling is determined by the button's content, which is, for example: `Editing template: Home ⌘K`. This labeling is less than ideal because it communicates the current state of the UI instead of the button's action. Instead, any interactive control must communicate what it does. The state should be communicated separately.

Minor:
This whole chunk of HTML is invalid HTML:
- A button element can't contain div elements.
- A button element can't contain headings h1, h2, etc.
- Worth reminding that also span elements can't contain headings h1, h2, etc.

Screenshot: this is how the accessible name of this button is announced by screen readers:

<img width="884" alt="Screenshot 2023-06-13 at 15 19 00" src="https://github.com/WordPress/gutenberg/assets/1682452/fd469f04-7072-4c71-8d9f-487a5a55d548">

Also worth noting that the visible text and the actual accessible name mismatch: this makes impossible to speech recognition software users issuing a vocal command. See WCAG 2.5.3 [Label in Name](https://www.w3.org/TR/WCAG21/#label-in-name). the best practice is to make sure that the actual accessible name matches, or at least _starts_, with the text that is visible to users.

### Step-by-step reproduction instructions

- Go to the Site editor.
- Use the Tab key to navigate the controls in the top bar.
- Observ ethe button that opens the command centre doesn't have any focus style.
- Use a screen reader or just inspect the source to check what the accessible name of the button is.
- Observe the accessible name is, for example: `Editing template: Home ⌘K`.
- Observe the button name doesn't tell anything about the button action.
- Observe that also visually there's nothing to communicate this button will open the command centre. I don;'t think the visible `⌘K` is a clear indication of what this button does.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

afercia:
Note that this issue is in part connected with the existing issue about the main H1 in the Site editor and about what the H1 should communicate to users. To start with, seems to me placing the main H1 within this button is very far from ideal, as the button and the H1 should have very different purposes.

afercia:
Additionally, go to:
[...]

PR #62527, opened 2024-06-13
Try revamped Document Bar

stokesman:
## What?
An attempt to fix/improve the Document Bar.

## Why?
- Its markup should be valid HTML
- The command palette button’s focus and hover states shouldn’t appear to include the back button.
- The command palette button should have an appropriate accessible label and indicate that it has a popup.
- Its responsive display and truncation of text should work better.

These points are all part of #51460, though I'm not sure this completely resolves that issue.

## How?
- Restructures
   - for valid markup (moves `<h1>` and `<div>` out of the `<button>`)
   - to keep the document title from being in the accessible text of a button that launches the command palette
- Restyles
  - to respond better at narrower sizes
  - so the command palette button doesn’t overlap the back button
This change lead to the substantial design difference in this PR. Currently the background and its hover state are on the root element and each button needs its own. Putting the same background on the buttons to preserve the design isn’t enough. Their border-radius would have to be removed on the sides the buttons meet. Then during transitions the buttons fading and overlapping ruin the effect*
- Puts an accessible label related to the button action on the command palette button and gives it an `aria-haspopup` attribute
- Tries an exit animation for the changes between template/document
- Tries a layout animation on the container (this isn’t ideal or even on focus—just an attempt to soften the layout jump that was somewhat recently introduced)

*Perhaps this can be retried as it was before revising the animation and now it possibly could work.

## Testing Instructions
1. Open a post/page in the editor 
2. Interact with the document bar
3. Edit the template
4. Repeat step 2

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/9000376/d51a0771-1986-4d9a-8b91-41630df95b05

### DIFF
--- a/packages/editor/src/components/collapsible-block-toolbar/style.scss
+++ b/packages/editor/src/components/collapsible-block-toolbar/style.scss
@@ -78,5 +78,5 @@
 }
 
 .editor-collapsible-block-toolbar__toggle {
-	margin-left: 2px; // Allow focus ring to be fully visible
+	align-self: center;
 }
--- a/packages/editor/src/components/document-bar/index.js
+++ b/packages/editor/src/components/document-bar/index.js
@@ -198,6 +198,8 @@ export default function DocumentBar() {
 						onClick={ () => openCommandCenter() }
 						size="compact"
 						variant="tertiary"
+						label={ __( 'Open the command palette' ) }
+						aria-haspopup="true"
 					>
 						<span className="editor-document-bar__shortcut">
 							{ displayShortcut.primary( 'k' ) }
--- a/packages/editor/src/components/document-bar/index.js
+++ b/packages/editor/src/components/document-bar/index.js
@@ -19,7 +19,7 @@ import { chevronLeftSmall, chevronRightSmall } from '@wordpress/icons';
 import { displayShortcut } from '@wordpress/keycodes';
 import { store as coreStore } from '@wordpress/core-data';
 import { store as commandsStore } from '@wordpress/commands';
-import { useRef, useEffect } from '@wordpress/element';
+import { useRef } from '@wordpress/element';
 import { useReducedMotion } from '@wordpress/compose';
 import { decodeEntities } from '@wordpress/html-entities';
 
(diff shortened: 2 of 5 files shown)


afercia:
@stokesman thanks for working on this ❤️ 
At a first glance it's certainly a good improvement in terms of markup and focus style.

A few issues to address though.

The heading shouldn't be a H1. There's already a visually hidden H1 in the page. As such, there are now two H1. That's not OK for semantics and accessibility. While this issue may be not entirely new, this is a good opportunity to fix it. It should be changed to a H2. Screenshot:

Screenshot 2024-05-31 at 10 02 22 2

When editing some objects, e.g. a template, the heading gets an `aria-label` attribute that prepends the text `editing {object type}:` That's not ideal because the visible text and the actual accessible name mismatch. If the `editing {object type}:` part is considered important information, I'd recommend to just make it visible. I don't see a good reason to visually hide it. I'd remove hte aria-label and use its value as the heading text.

I wouldn't recommend to overlay the button and heading, then using `pointer-events: none` to have a larger tartet size and be able to click the whole thing. Also, the visual order and the DOM order mismatch. 
As a user, I see the post title. I don't expect the post title to be clickable to open the command palette (even if under the hood I'm actually clicking the button). I would recommend to change the design and _make things look what they actually are_:
- The Back button should look like an interactive control users can click.
- The post title heading should not look like an interactive button. It's not an interactive control. It's a heading.
- The command palette should look like an interactive control users can click.

Roughly, something like in this screenshot (focus set on both buttons to better illustrate):
Screenshot 2024-06-18 at 14 32 07

Lastly, as also noted elsewhere, the Document bar should not use the `verse` icon because that's already used for the... Verse block. Same applies to the Post settings panel card, but that's another issue. To build a cohesive visual language, icons should not be used for featurees that are entirely different. Screenshot:
Screenshot 2024-06-18 at 14 12 41
If it was up to me, I'd even remove the icon. However, if an icon needs to be used to represent the edited object type, I'd like to remind the design team WordPress uses another icon for the Post object since a few years... 
Screenshot 2024-06-18 at 14 46 06

afercia (changes requested):
Please see previous comment.

PR #62528, opened 2024-06-13
Site Editor: Fix modal incorrect size for edit template modal

### LINKED ISSUE #62524: Site Editor: modal incorrect size for edit template modal
Using beta 2 for 6.6 without GB installed. When clicking around the broader template area while editing a page in the Site Editor, I noticed that the modal that pops up doesn't match the usual modal size. Below is what it currently looks like:

Screenshot 2024-06-12 at 2 06 38 PM

Below is what it should look like:

Screenshot 2024-06-12 at 2 10 14 PM

While not a "bug" per say, I think this is worth rectifying as a low code change but an important detail to get right for a smooth site editing experience.

akasunil:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/62524
Changed modal width to medium size. 

## Screenshots or screencast 
image

### DIFF
--- a/packages/components/src/modal/style.scss
+++ b/packages/components/src/modal/style.scss
@@ -158,6 +158,7 @@
 	&.hide-header {
 		margin-top: 0;
 		padding-top: $grid-unit-40;
+		max-width: $modal-width-medium;
 	}
 
 	&.is-scrollable:focus-visible {


t-hamano:
Hi @akasunil,

Thanks for addressing this issue.

You can change the size of the Modal with [the `size` prop](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/modal/README.md#size-small--medium--large--fill).

Also, it looks like #62532 is making changes to a wider range of components, so it might be better to move #62532 forward.

akasunil:
Covered by PR https://github.com/WordPress/gutenberg/pull/62532

PR #62529, opened 2024-06-13
Global Styles: Skip registration of variation styles when unsupported

aaronrobertshaw:
## What?

Prevents going through all the work of searching for theme.json partials, parsing them and the core theme.json etc when the theme doesn't have its own theme.json.

## Why?

Performance and simplicity.

## How?

Make a check to `wp_theme_has_theme_json()` and return early if applicable.

## Testing Instructions

1. Add a block style variation to a theme via any source e.g. through its theme.json and the snippet below:

```json
                       "variations": {
                               "FromTheme": {
                                       "blockTypes": [
                                               "core/group",
                                               "core/columns"
                                       ],
                                       "color": {
                                               "background": "blue"
                                       }
                               }
                       },
```

2. Confirm this variation still shows and works appropriately
3. Switch theme to a classic theme and confirm no variations are shown
4. Update a hybrid theme to have a block style variation within its theme.json via snippet below. 

```json
{
	"version": 2,
	"styles": {
		"blocks": {
			"variations": {
				"FromTheme": {
					"blockTypes": [ "core/group", "core/columns" ],
					"color": {
						"background": "blue"
					}
				}
			}
		}
	}
}

```

5. Confirm the variation shows

### DIFF
--- a/lib/block-supports/block-style-variations.php
+++ b/lib/block-supports/block-style-variations.php
@@ -486,6 +486,13 @@ function gutenberg_register_block_style_variations_from_theme_json_data( $variat
  * @access private
  */
 function gutenberg_register_block_style_variations_from_theme() {
+	$has_partials_directory = is_dir( get_stylesheet_directory() . '/styles' ) || is_dir( get_template_directory() . '/styles' );
+
+	// Skip any registration of styles if no theme.json or variation partials.
+	if ( ! wp_theme_has_theme_json() && ! $has_partials_directory ) {
+		return;
+	}
+
 	// Partials from `/styles`.
 	$variations_partials = WP_Theme_JSON_Resolver_Gutenberg::get_style_variations( 'block' );
 	gutenberg_register_block_style_variations_from_theme_json_data( $variations_partials );
--- a/lib/block-supports/block-style-variations.php
+++ b/lib/block-supports/block-style-variations.php
@@ -486,9 +486,19 @@ function gutenberg_register_block_style_variations_from_theme_json_data( $variat
  * @access private
  */
 function gutenberg_register_block_style_variations_from_theme() {
+	/*
+	 * Skip any registration of styles if no theme.json or variation partials are present.
+	 *
+	 * Given the possibility of hybrid themes, this check can't rely on if the theme
+	 * is a block theme or not. Instead:
+	 *   - If there is a primary theme.json, continue.
+	 *   - If there is a partials directory, continue.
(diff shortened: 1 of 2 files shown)


oandregal (approved):

aaronrobertshaw:
I'm holding off on merging this one for the moment. 

The solution would work nicely for block themes but also impacts hybrid themes. I'll take another run at this tomorrow.

lib/block-supports/block-style-variations.php
```diff
@@ -486,6 +486,13 @@ function gutenberg_register_block_style_variations_from_theme_json_data( $variat
  * @access private
  */
 function gutenberg_register_block_style_variations_from_theme() {
+	$has_partials_directory = is_dir( get_stylesheet_directory() . '/styles' ) || is_dir( get_template_directory() . '/styles' );
+
+	// Skip any registration of styles if no theme.json or variation partials.
+	if ( ! wp_theme_has_theme_json() && ! $has_partials_directory ) {
+		return;
+	}
+
```
andrewserong: Is this meant to be an `||` instead of an `&&` for the return?

Also, a tiny optimisation, but if it's meant to be an `||` we could do two separate checks, so that we return if `! wp_theme_has_theme_json()` and then return if `! $has_partials_directory`. That way there's no need to check if a directory exists if we know there's no theme JSON?

aaronrobertshaw: Good question 👍 

The `||` was deliberate as a block theme may or may not have a theme.json file. Hybrid themes also complicate the situation. 

I've pushed a more detailed explanatory inline comment for this early return.

```
	/*
	 * Skip any registration of styles if no theme.json or variation partials are present.
	 *
	 * Given the possibility of hybrid themes, this check can't rely on if the theme
	 * is a block theme or not. Instead:
	 *   - If there is a primary theme.json, continue.
	 *   - If there is a partials directory, continue.
	 *   - The only variations to be registered from the global styles user origin,
	 *     are those that have been copied in from the selected theme style variation.
	 *     For a theme style variation to be selected it would have to have a partial
	 *     theme.json file covered by the previous check.
	 */
```

andrewserong: Looks good, thanks for adding the explanatory comment!

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: f0327a0673

PR #62530, opened 2024-06-13
Fix "insert before/after" not showing for blocks in site editor

### LINKED ISSUE #62367: "Add before" and "Add after" missing when editing pages in the Site Editor
### Description

Looks like something with content locking is reducing the editing capabilities when using the Site Editor. This occurs with blocks nested directly within the "Content" block.  

I'd expect all standard editing controls available.  

1. Open the Site Editor.
2. Select "Pages".
3. Select "Add new page".
4. Add content to the page.
5. Open the block toolbar options menu. 
6. See missing "Add before" and "Add after" controls. 

https://github.com/WordPress/gutenberg/assets/1813435/01c6dd8c-4e46-4f16-9b7e-daeceff4a393

### Environment info

- WordPress trunk
- Gutenberg trunk

Yes

Yes

kevin940726:
## What?

Fix https://github.com/WordPress/gutenberg/issues/62367.

Fix "Insert before" & "Insert after" not showing in the site editor when editing pages.

## Why?

The bug was introduced in https://github.com/WordPress/gutenberg/pull/61127#discussion_r1629851012. The condition was not enough because it's checking the root client id (parent block).

## How?

Move the check to the block settings menu and check the block itself.

## Testing Instructions

1. Follow the instructions in https://github.com/WordPress/gutenberg/issues/62367 and expect to see "Insert before/after".
2. Expect "Insert before/after" not to show when selecting binding blocks inside pattern overrides.

## Screenshots or screencast 
image

### DIFF
--- a/packages/block-editor/src/components/block-actions/index.js
+++ b/packages/block-editor/src/components/block-actions/index.js
@@ -31,12 +31,10 @@ export default function BlockActions( {
 				getDirectInsertBlock,
 				canMoveBlocks,
 				canRemoveBlocks,
-				getBlockEditingMode,
 			} = select( blockEditorStore );
 
 			const blocks = getBlocksByClientId( clientIds );
 			const rootClientId = getBlockRootClientId( clientIds[ 0 ] );
-			const rootBlockEditingMode = getBlockEditingMode( rootClientId );
 			const canInsertDefaultBlock = canInsertBlockType(
 				getDefaultBlockName(),
 				rootClientId
@@ -48,9 +46,7 @@ export default function BlockActions( {
 			return {
 				canMove: canMoveBlocks( clientIds, rootClientId ),
 				canRemove: canRemoveBlocks( clientIds, rootClientId ),
-				canInsertBlock:
-					( canInsertDefaultBlock || !! directInsertBlock ) &&
-					rootBlockEditingMode === 'default',
+				canInsertBlock: canInsertDefaultBlock || !! directInsertBlock,
 				canCopyStyles: blocks.every( ( block ) => {
 					return (
 						!! block &&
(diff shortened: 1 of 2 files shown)


talldan:
> Add check for checking if it's inside a pattern parent. It's a bit unfortunate that it seems like the only way for now. 🤔

I feel like there must be a way. I think it's very difficult to understand the logic, so here's a quick breakdown of what I know it to be:
- `templateLock === 'contentOnly'` - this only 'locks' the inner blocks and not the parent. I imagine it can still be overridden for some inner blocks by setting the `blockEditingMode`.
- `blockEditingMode === 'contentOnly'` - this does the opposite and only 'locks' the parent block and not its inner blocks which by default still use a `default` block editing mode, but can be set to different block editing modes explicitly. 

The two shouldn't be confused, though I know it's very confusing (I left this feedback on the PR that implemented it, but it was ignored).

Editable blocks (the block itself) within a pattern will have a `blockEditingMode` of `contentOnly`. In a page the inner blocks of 'Content' will have a block editing mode of `default`, so there's a definite distinction.

kevin940726:
> Editable blocks (the block itself) within a pattern will have a `blockEditingMode` of `contentOnly`. In a page the inner blocks of 'Content' will have a block editing mode of `default`, so there's a definite distinction.

Yeah, I think you're right! The original condition failed because it's checking the root client id 🤦. I think I fixed it now in [ce14a1c](https://github.com/WordPress/gutenberg/pull/62530/commits/ce14a1c0ee71ad9d349f38781415cb65ecacdd7c).

It's indeed very confusing 😅.

noisysocks:
This looks right to me. Sorry it's confusing 😔 To straighten it out we probably need to deprecate both and move to an API design that doesn't rely on encapsulating a bunch of different behaviours into one string.

Testing now.

noisysocks (approved):
I tested the steps in the linked issue and checked that the movers don't appear when selecting an editable block in a synced pattern with overrides 👍

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: fe9fe31dd2

PR #62531, opened 2024-06-13
Editor: Render editPost slots only in the post editor (same for site editor)

youknowriad:
Related #52632 

## What?

It seems that some plugins are using the old editPost and editSite slots in the same script files and might also be using these script files in site editor even if they're meant only for post editor (or vice versa). While obviously there are better solutions for these plugin authors to avoid loading the slots and scripts were not needed, this PR tries to mitigate the slots unification effects on this kind of scripts.

 - EditPost slots only work on edit-post.
 - EditSite slots only work on edit-site.

### DIFF
--- a/packages/edit-post/src/deprecated.js
+++ b/packages/edit-post/src/deprecated.js
@@ -12,6 +12,7 @@ import {
 	PluginSidebar as EditorPluginSidebar,
 	PluginSidebarMoreMenuItem as EditorPluginSidebarMoreMenuItem,
 } from '@wordpress/editor';
+import { getPath } from '@wordpress/url';
 import deprecated from '@wordpress/deprecated';
 
 /**
@@ -19,6 +20,10 @@ import deprecated from '@wordpress/deprecated';
  */
 import { unlock } from './lock-unlock';
 const { PluginPostExcerpt } = unlock( editorPrivateApis );
+
+const isSiteEditor = getPath( window.location.href )?.includes(
+	'site-editor.php'
+);
 
 const deprecateSlot = ( name ) => {
 	deprecated( `wp.editPost.${ name }`, {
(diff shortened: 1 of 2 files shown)


jeherve:
> It seems that some plugins are using the old editPost and editSite slots in the same script files and might also be using these script files in site editor even if they're meant only for post editor (or vice versa).

The Jetpack plugin does just that, and I can confirm that this patch solves the issue ; slots that are meant for the post editor are no longer displayed in the site editor.

> While obviously there are better solutions for these plugin authors to avoid loading the slots and scripts were not needed

I'd love to see a dev note or a post on the developer blog that would go into more details about the different solutions that will be available to us starting in WordPress 6.6.

youknowriad:
> I'd love to see a dev note or a post on the developer blog that would go into more details about the different solutions that will be available to us starting in WordPress 6.6.

Yes, definitely on my todo list.

ndiego:
I just tested with [Stackable](https://wordpress.org/plugins/stackable-ultimate-gutenberg-blocks/) and [Block Visibility](https://wordpress.org/plugins/block-visibility/).

| This PR | WordPress 6.6 Beta 2 |
|-|-|
|image|image|

Block Visibility currently uses:

```
import { PluginMoreMenuItem } from '@wordpress/edit-post';
```

With this PR, the menu item does not display in the Site Editor. I think this approach makes sense. As soon as I update it to: 

 ```
import { PluginMoreMenuItem } from '@wordpress/editor';
```

it does display as expected. 

Overall, I think this is a great PR since it retains the current experience for extenders. Everyone should migrate to `@wordpress/editor` and I think the deprecation message helps emphasize this. 

image

The last piece of the puzzle will be to document how you can use `@wordpress/editor` but add conditions to restrict where your content is displayed.

ellatrix (approved):

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 79c00ca887

PR #62536, opened 2024-06-13
Use stable reference for `getEntityActions` action

ntsekouras:
## What?
Small follow up of: https://github.com/WordPress/gutenberg/pull/62052#discussion_r1637800466

>Fixes the "The 'useSelect' hook returns different values when called with the same state and parameters. This can lead to unnecessary rerenders. " warning.

### Testing instructions
Everything should work as before regarding the post actions

### DIFF
--- a/packages/editor/src/dataviews/store/private-selectors.ts
+++ b/packages/editor/src/dataviews/store/private-selectors.ts
@@ -1,8 +1,15 @@
 /**
+ * WordPress dependencies
+ */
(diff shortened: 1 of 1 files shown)


youknowriad (approved):

ellatrix:
Does this need backporting?

ntsekouras:
>Does this need backporting?

@ellatrix no, it doesn't.

PR #62537, opened 2024-06-13
Template inspector: Small visual adjustments.

jasmussen:
## What?

A small amount of spacing changes to the post/page/template inspector summary panel.

## Why?

The added spacing avoids some wrapping and improves legibility.

## How?

* The left column in the summary panel is 8% wider avoiding wrapping of, among other things, "Posts per page"
* Spacing above and below summary items is now identical, so "Areas" doesn't look so tightly coupled.
* If a template has a purple icon denoting synced/used in multiple places, now the text is also purple.

Page inspector before and after:

| Before | After |
|--------|--------|
|  |  | 

Template inspector ebefore and after:

| Before | After |
|--------|--------|
| template before | template after | 

## Testing Instructions

Test the main document inspector in post, page, site editors, across pages and templates, and posts for good measure.

### DIFF
--- a/packages/editor/src/components/post-card-panel/style.scss
+++ b/packages/editor/src/components/post-card-panel/style.scss
@@ -35,6 +35,10 @@
 
 .editor-post-card-panel__icon.is-sync {
 	fill: var(--wp-block-synced-color);
+
+	& + h2 {
+		color: var(--wp-block-synced-color);
+	}
 }
 
 .editor-post-card-panel__title-badge {
--- a/packages/editor/src/components/post-card-panel/style.scss
+++ b/packages/editor/src/components/post-card-panel/style.scss
@@ -36,7 +36,7 @@
 .editor-post-card-panel__icon.is-sync {
 	fill: var(--wp-block-synced-color);
 
-	& + h2 {
+	& + .editor-post-card-panel__title {
 		color: var(--wp-block-synced-color);
 	}
 }
--- a/packages/editor/src/components/post-panel-row/style.scss
+++ b/packages/editor/src/components/post-panel-row/style.scss
@@ -6,7 +6,7 @@
 }
 
 .editor-post-panel__row-label {
-	width: 30%;
+	width: 38%;
 	flex-shrink: 0;
 	min-height: $grid-unit-40;
 	display: flex;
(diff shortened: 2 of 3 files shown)


packages/editor/src/components/post-card-panel/style.scss
```diff
@@ -35,6 +35,10 @@
 
 .editor-post-card-panel__icon.is-sync {
 	fill: var(--wp-block-synced-color);
+
+	& + h2 {
+		color: var(--wp-block-synced-color);
+	}
```
t-hamano:
```suggestion
	& + .editor-post-card-panel__title {
		color: var(--wp-block-synced-color);
	}
```

This is just my preference, so we may not need to change it.

t-hamano (commented):
I think this is a great improvement.

Whether the text wraps will depend on the locale, but I was particularly bothered by the narrow label width in Japanese. Since it's impractical to prevent all text wrapping in all locales, I think 38% is a reasonable value.

| Before | After |
|--------|--------|
| before | after |
| image | image | 

> If a template has a purple icon denoting synced/used in multiple places, now the text is also purple.

This PR also changes the text color of synced patterns and template parts, but that's to be expected, right?

image

image

jasmussen:
> This PR also changes the text color of synced patterns and template parts, but that's to be expected, right?

Yes, this is at least my take on it. We're being a bit inconsistent about it at the moment, showing all-purple in some places:

site title

But not in others:
block toolbar

pattern inserter

There's just something more scannable when the title plus icon is the same color, than if it's only the icon. But if you prefer, I'm happy to extract that change to a separate PR, so we can land the other improvements?

Speaking of 38%—is that wrapping actually better for japanese? Or would 40% be better? Is there a sweet-spot balance here? We want to be careful with the balance because in most cases, the content in the right column is the content that needs space. But if 40% is better than 38, happy to try it.

t-hamano (approved):
> Speaking of 38%—is that wrapping actually better for japanese? Or would 40% be better?

In CJK languages, there are no spaces between words like in Latin languages. And there is no mechanism in CSS yet to distinguish between words in CJK languages ​​and wrap them appropriately (I think that's why `text-wrap:pretty` doesn't work in CJK languages).

To be precise, there is [text-wrap: auto-phrase](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#auto-phrase), but it is experimental and not supported by Firefox or Safari.

Therefore, I believe that wrapping text mid-word is common and acceptable in CJK languages.

If we want to completely prevent text wrapping in Japanese , we need a larger width.

Post: Requires `46%` width

post

Blog home template: Requires `48%` width

blog_home


These widths will feel very wide in other languages, so I think it's fine to leave them at `38%`.

So I'm happy to give this PR a green check.

jasmussen:
Thank you for the check. I'll merge when they pass, but then also be here for any followups re: purple text or not.

PR #62542, opened 2024-06-13
Save Panel: Remove connections icon and fix padding

artemiomorales:
## What?

This PR revises the styles of the indicator regarding post meta in the save panel, namely removing the connections icon and fixing the padding.

## Why?

While the connections icon is explanatory when post meta has been modified via Block Bindings, it is possible to modify post meta by other means and it is not accurate to show the icon in all circumstances.

## How?

This PR removes the connections icon and modifies the styles to fix the padding.

## Testing Instructions

<details>

```php

	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
		)
	);
}
```

</details>

<details>

<summary>2. Add a paragraph block bound to the custom field using the Code Editor</summary>

```
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Press the Save button.
4. Verify that the Save Panel appears with the updated styles.

### Before
save-panel_before

### After
save-panel_after

### DIFF
--- a/packages/editor/src/components/entities-saved-states/style.scss
+++ b/packages/editor/src/components/entities-saved-states/style.scss
@@ -32,7 +32,7 @@
 }
 
 .entities-saved-states__post-meta {
-	margin-left: $grid-unit-30;
+	margin-left: 28px;
 	align-items: center;
 }
 
--- a/packages/editor/src/components/entities-saved-states/style.scss
+++ b/packages/editor/src/components/entities-saved-states/style.scss
@@ -30,16 +30,3 @@
 		margin-bottom: $grid-unit-05;
 	}
 }
-
-.entities-saved-states__post-meta {
-	margin-left: 28px;
-	align-items: center;
-}
-
-.entities-saved-states__connections-icon {
-	flex-grow: 0;
-}
-
-.entities-saved-states__bindings-text {
-	flex-grow: 1;
-}
--- a/packages/editor/src/components/entities-saved-states/entity-record-item.js
+++ b/packages/editor/src/components/entities-saved-states/entity-record-item.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { CheckboxControl, Flex, PanelRow } from '@wordpress/components';
+import { CheckboxControl, PanelRow } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { useSelect } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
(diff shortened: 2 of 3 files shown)


artemiomorales:
I've hardcoded the `margin-left` for the post meta indicator. To make it line up with the checkbox field above it, ideally we would use something like the following:

```margin-left: calc(var(--checkbox-input-size) + var(--checkbox-input-margin));```

However. `calc()` doesn't seem to work for this use case. Does anyone have any suggestions, or should we just proceed with the hardcoded value for now?

File: packages/editor/src/components/entities-saved-states/style.scss [hunk 1]
```diff
@@ -32,7 +32,7 @@
 }
 
 .entities-saved-states__post-meta {
-	margin-left: $grid-unit-30;
+	margin-left: 28px;
```
cbravobernal: Can we use an unit instead of hard-coding a margin?

cbravobernal:
> or should we just proceed with the hardcoded value for now?

Let's look for an alternative :-)

youknowriad:
Thanks for following-up with this, I really appreciate it. 

Can we use the same styling for "post meta" as the style we use to detail global styles changes

Screenshot 2024-06-14 at 7 56 59 AM

The other thing that is probably best addressed for later is whether we should show the full list of modified properties included the list of meta fields. Something like:

title, content, post meta ( my_meta_field, my_other_meta_field... )

cc @jasmussen @jameskoster

jasmussen:
The original motivation was to separate out, with checkboxes, [individual attributes](https://www.figma.com/design/W2SO7yJG09O65q7JjfXZ4m/Block-connections?node-id=602-8369&t=ncsdv0XlFylQ9rmE-1). In absence of that, I would agree with Riad that we should just use the regular/default gray-text bulleted layout.

artemiomorales:
> Can we use the same styling for "post meta" as the style we use to detail global styles changes

@youknowriad @cbravobernal Done! This is what it looks like now: 

Screenshot 2024-06-14 at 10 30 13 AM

packages/editor/src/components/entities-saved-states/style.scss
artemiomorales: @cbravobernal In light of the [design changes](https://github.com/WordPress/gutenberg/pull/62542#issuecomment-2167528721), I now just use the preexisting `entities-saved-states__changes` class for styling 👍

cbravobernal:
I'm not getting those styles. Am I missing something?

Tested on Firefox.

Screenshot 2024-06-14 at 12 40 11

artemiomorales:
> I'm not getting those styles. Am I missing something?
>
> Tested on Firefox.

@cbravobernal I see it on Firefox without issue 🤔
Maybe try clearing cache, rebuilding, etc.?

SantosGuillamot (approved):
I've tested it in Chrome and Safari, and I can see it working properly.

Screenshot 2024-06-17 at 08 44 20

Apart from that, it seems it is using the same markup as [the Global Styles changes](https://github.com/WordPress/gutenberg/blob/7aadf9a419e7eec092389dc9c47dfb332d5da4b4/packages/editor/src/components/entities-saved-states/entity-type-list.js#L58-L62), so I assume it is correct.

I'll wait to merge the pull request until @cbravobernal can confirm if the issue he was having is solved or is still present.

cbravobernal:
Got it working. Not sure about what was preventing the style from view.

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: ecc7107617

PR #62545, opened 2024-06-13
fix rawHandling, pasteHandling for mixed content with blocks and classic

saulyz:
## What?

The rawHandler and the pasteHandler had a naive check for the block existence. The test was looking for string "
When we convert legacy content to blocks sometimes whole article or some parts silently fail. Clicking "convert to blocks" does nothing - everything stays the same. In the _Editor view_ one could not distinguish a problem content and would not get any error.

## How?

Not changing the overall logic or the process, instead refining the check further checking if parsing has resulted with the same 'core/freeform' block. If content parsing results with the same and single freeform, let's continue the content conversion (previously process would stop).

## Testing Instructions

Adding test files to PR that illustrate the issue.


saulyz:
@ellatrix need your eyes here

ellatrix (approved):
I'm surprised this was never caught earlier.

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: e08c1582a2

PR #62548, opened 2024-06-13
Reorders colors in color palette with body and text colors first

### LINKED ISSUE #62117: Try ordering colors in color variations with body and text colors first
Currently color palette colors are ordered exactly as depicted in a theme variation's JSON file. The challenge here is that it's difficult to parse which palette would be considered a "dark" palette. 

For example, in the visual below, the last color palette sets the background of the site to dark green, but it's unclear that it does so, as the variation has more than five colors (the current max colors visible in a palette) and the color used as the background color is not visible in the palette. 

A proposal could be to use the color assigned as the body background color first, the color assigned as the body text color second, and then the remaining colors in order up to five. 

The result would be what's depicted on the right:

CleanShot 2024-05-29 at 13 47 19

Perhaps not 100% ideal, but there is more clarity. Thoughts?

Related https://github.com/WordPress/gutenberg/issues/61215

amitraj2203:
## What?
Closes: #62117 

## Why?
Currently color palette colors are ordered exactly as depicted in a theme variation's JSON file. So it doesn't shows which palette would be considered a `dark` palette.

## How?
Fetches the background and text colors and replaces them from the palette, positioning them at the first and second indexes, respectively.

### DIFF
--- a/packages/edit-site/src/components/global-styles/preset-colors.js
+++ b/packages/edit-site/src/components/global-styles/preset-colors.js
@@ -1,11 +1,42 @@
 /**
+ * WordPress dependencies
+ */
+import { privateApis as blockEditorPrivateApis } from '@wordpress/block-editor';
+
+/**
  * Internal dependencies
  */
 import { useStylesPreviewColors } from './hooks';
+import { unlock } from '../../lock-unlock';
 
 export default function PresetColors() {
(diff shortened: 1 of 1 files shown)


amitraj2203:
Hi @richtabor, I have created a PR this [issue](https://github.com/WordPress/gutenberg/issues/62117). Please let me know if it requires any changes.

Thanks

PR #62549, opened 2024-06-13
Remove html encoded entities from post slug when cleaning

### LINKED ISSUE #62543: The `&` symbol becomes `amp` in the draft URL
### Description

When you type `Ampersand & And` as the title of an Post and save the Post as a draft, you will see that the URL becomes `/ampersand-amp-and/`. If you publish the Post, the URL becomes `/ampersand-and/`.

I was originally going to write this issue asking if it was necessary to add `amp` to the URL, because from an SEO point of view it makes no sense and just unnecessarily lengthens the URL. However, during initial testing, when publishing the Post, the URL became `/ampersand-and/`.

Now, I wonder if this may cause confusion to other people. If this is a bug or if it makes any sense for `amp` to appear in the URL, even if only in drafts. Also, it is not the same with `%`, `$`, `#`, `@`, or others. For example, if you write `Ampersand & And % One # Two $ Three @ Four` the URL of the draft will be `/ampersand-amp-and-one-two-three-four/`

1. Go to Settings > Permalinks and set `Post name` as a permalink structure.
2. Create a Post and type `Ampersand & And` as the Post title.
3. Save the Post as a Draft.
4. In the editor sidebar click on the URL to view the generated URL. It is this: `/ampersand-amp-and/`
5. Publish the Post.
6. Chceck URL of the Post. It is this: `/ampersand-and/`

ampersand-and

### Environment info

_No response_

Yes

Yes

tharsheblows:
Fixes #62543

## What?

This removes the html encoded entity strings for & and < from the slug in `cleanForSlug()`. These are only two entities which seem to cause an issue.

## Why?

>When you type Ampersand & And as the title of an Post and save the Post as a draft, you will see that the URL becomes /ampersand-amp-and/. If you publish the Post, the URL becomes /ampersand-and/.

**Step-by-step reproduction instructions**

- Go to Settings > Permalinks and set Post name as a permalink structure.
- Create a Post and type Ampersand & And as the Post title.
- Save the Post as a Draft.
- In the editor sidebar click on the URL to view the generated URL. It is this: /ampersand-amp-and/
- Publish the Post.
- Chceck URL of the Post. It is this: /ampersand-and/

## How?

Replaces `&amp;` and `&lt;` with an empty string in `cleanForSlug()`. Also adds tests.

## Testing Instructions

- create a post with the title `And & &&& &amp; &&& &lt;1 039898':!#*() &&& <>.“‘”’" hey` (note the string `&amp;` and string `&lt;`)
- check the post slug shown in the Link field in the sidebar, it should be  `and-amp-lt1-039898-hey`
- publish the post and look at the post name bit in the slug on the front end, it should be the same `and-amp-lt1-039898-hey`

### DIFF
--- a/packages/url/src/clean-for-slug.js
+++ b/packages/url/src/clean-for-slug.js
@@ -24,7 +24,7 @@ export function cleanForSlug( string ) {
 		return '';
 	}
 	return (
-		removeAccents( string )
+		removeAccents( string.replace( /&amp;/g, '' ).replace( /&lt;/g, '' ) )
 			// Convert each group of whitespace, periods, and forward slashes to a hyphen.
 			.replace( /[\s\./]+/g, '-' )
 			// Remove anything that's not a letter, number, underscore or hyphen.
--- a/packages/url/src/test/index.js
+++ b/packages/url/src/test/index.js
@@ -1117,6 +1117,24 @@ describe( 'cleanForSlug', () => {
 		expect( cleanForSlug( 'the long - cat' ) ).toBe( 'the-long-cat' );
 		expect( cleanForSlug( 'the----long---cat' ) ).toBe( 'the-long-cat' );
 	} );
+
+	it( 'Should remove html entities for ampersands', () => {
+		expect(
+			cleanForSlug( 'the long cat &amp; a dog &amp;&amp; fish' )
+		).toBe( 'the-long-cat-a-dog-fish' );
+		expect( cleanForSlug( 'the long cat &amp;amp; dog' ) ).toBe(
+			'the-long-cat-amp-dog'
+		);
+	} );
+
+	it( 'Should remove html entities for less thans', () => {
+		expect( cleanForSlug( '&lt;red is not blue>' ) ).toBe(
+			'red-is-not-blue'
+		);
+		expect( cleanForSlug( '&amp;lt; red and blue are purple' ) ).toBe(
+			'lt-red-and-blue-are-purple'
+		);
+	} );
 } );
 
 describe( 'normalizePath', () => {


tharsheblows:
Adding a comment to update the props as I've linked my account now (thanks to whoever set up that bot with instructions!)

packages/url/src/clean-for-slug.js
```diff
@@ -24,7 +24,7 @@ export function cleanForSlug( string ) {
 		return '';
 	}
 	return (
-		removeAccents( string )
+		removeAccents( string.replace( /&amp;/g, '' ).replace( /&lt;/g, '' ) )
```
tyxla, line 27: What about any other commonly used HTML entities, like `&gt;`, `&nbsp;`, `&apos;`, `&quot;` and others (not an exhaustive list by any means)?

packages/url/src/test/index.js
```diff
@@ -1117,6 +1117,24 @@ describe( 'cleanForSlug', () => {
...
+	it( 'Should remove html entities for ampersands', () => {
+		expect(
+			cleanForSlug( 'the long cat &amp; a dog &amp;&amp; fish' )
+		).toBe( 'the-long-cat-a-dog-fish' );
+		expect( cleanForSlug( 'the long cat &amp;amp; dog' ) ).toBe(
+			'the-long-cat-amp-dog'
+		);
+	} );
+
+	it( 'Should remove html entities for less thans', () => {
```
tyxla, line 1130: Likely not applicable after addressing my other feedback, but this could be a good case for [`it.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) as that could simplify the test code and remove redundancy.

packages/url/src/clean-for-slug.js
tyxla, line 27: The `cleanForSlug` function is intended to be a mirror of [`sanitize_title()`](https://developer.wordpress.org/reference/functions/sanitize_title/), which doesn't account for HTML entities. This is also confirmed by the `cleanForSlug` documentation, which specifically claims it [doesn't touch any HTML entities](https://github.com/WordPress/gutenberg/blob/a3360a8f228a6eea778c130c327c0374f56d6352/packages/url/src/clean-for-slug.js#L15-L16). With that in mind, perhaps there is a better place to address this change?

tyxla (commented):
Thanks for the contribution, @tharsheblows!

I have left a few initial thoughts I had while reviewing the code. Let me know what you think!

packages/url/src/clean-for-slug.js
tharsheblows: @tyxla Yes, I agree with you, it should handle all entities. I wavered between keeping it limited to what you would get from the WP REST API response for the title (only those entities need to be handled) and all entities, wasn't quite sure which approach to take!

packages/url/src/test/index.js
tharsheblows: Nice! I'll add in after the discussion below, thank you!

packages/url/src/clean-for-slug.js
tharsheblows: Yes, I wondered why it wasn't a mirror – there is a core default filter added to `sanitize_title()` which runs the `$title` through [`sanitize_title_with_dashes()`](https://developer.wordpress.org/reference/functions/sanitize_title_with_dashes/#source) and the expectation when doing PHP development is that `sanitize_title( 'the dog & the cat' )` will return `the-dog-the-cat`. 

It's that core filter which handles the heavy lifting of sanitization and removal of the html entities among other things, this is the reason the string returned by `cleanForSlug()` doesn't match the actual slug returned by `sanitize_title()`.

People might be using this and changing the returned value could horribly mess their urls up – what about keeping back compat and introducing a new function `sanitizeForSlug()` which mirrors `sanitize_title()` with the default filter added (ie add in all the sanitization done by `sanitize_title_with_dashes()`) and then using `sanitizeForSlug()` when displaying the slug. 

Does that approach sound ok?

tyxla: > there is a core default filter added to sanitize_title() which runs the $title through [sanitize_title_with_dashes()](https://developer.wordpress.org/reference/functions/sanitize_title_with_dashes/#source) and the expectation when doing PHP development is that sanitize_title( 'the dog & the cat' ) will return the-dog-the-cat.

This makes sense, thanks for pointing it out 👍 

> People might be using this and changing the returned value could horribly mess their urls up – what about keeping back compat and introducing a new function sanitizeForSlug() which mirrors sanitize_title() with the default filter added (ie add in all the sanitization done by sanitize_title_with_dashes()) and then using sanitizeForSlug() when displaying the slug.

> Does that approach sound ok?

A separate utility function sounds good to me 👍 We'll definitely want to keep `cleanForSlug` as it is, since it intentionally doesn't handle entities.

tyxla: I see. In light of the recent conversation to stay compatible with the `sanitize_title_with_dashes` filter, let's introduce a utility function compatible with `sanitize_title` filtered by `sanitize_title_with_dashes` which converts some entities (spaces and dashes) to hyphens, and obliterates the rest of the HTML entities.

Mamaduka:
Thanks for the initial proposal, @tharsheblows!

I'm going to close this in favor of #70078.

PR #62550, opened 2024-06-13
Section styles: consolidate variation name

oandregal:
## What?

This PR enables theme.json partials to provide a different slug than the generated kebab-case to follow what any other registration mechanism (`block.json`, `register_style_variation`) does.

## Why?

For theme.json partials to work like any other registration mechanism, and to enable them to override variations with names that don't follow the kebab-case convention.

The existing mechanisms to register block style variations require a `name` and a `label`, see [docs](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-styles/). The `name` is used to generate the associated class (e.g.: `is-style-NAME`) and we don't do any processing of it — all the following are correct:

| Variation Name | Generated class |
| --- | --- |
| `dark` | `is-style-dark` |
| `Dark` | `is-style-Dark` |
| `PitchDark` | `is-style-PitchDark` |
| `pitch-dark` | `is-style-pitch-dark` |

**Before** this PR, the `title` from the theme.json partials was preprocessed to generate a kebab-cased variation name & class:

| Partial's Title | Variation Name | Generated class |
| --- | --- | --- |
| `dark` | `dark` | `is-style-dark` |
| `Dark` | `dark` | `is-style-dark` |
| `PitchDark` | `pitch-dark` | `is-style-pitch-dark` |
| `pitch-dark` | `pitch-dark` | `is-style-pitch-dark` |

**After** this PR, the theme.json partials can contain an optional `slug` (alternative: `name`?) key to be used in registering the variation and generating its class. This enables the theme.json partials to work with any variation name or class that wasn't kebab-case:

| Partial's Title | Partial's slug | Variation Name | Generated class |
| --- | --- | --- | --- |
| `dark` | - | `dark` | `is-style-dark` |
| `Dark` | `Dark` | `Dark` | `is-style-Dark` |
| `PitchDark` | `PitchDark` | `PitchDark` | `is-style-PitchDark` |
| `pitch-dark` | - | `pitch-dark` | `is-style-pitch-dark` |

Note the `slug` is optional and most of the cases will just let `title` be kebab-cased.

## How?

Allow `slug` to be provided in theme.json partials:

- Use `slug` if present, otherwise kebab-case the `title` a602aae7a669a2be1dae4ad4ad65fb115146224c
- Update schemas c0cca639858f0db2ea7d1784beca62c693ad6d4d

## Testing Instructions

Register a variation with a non-kebab-cased name. For example, paste the following in the `functions.php` of TT4:

```php
		register_block_style(
			'core/group',
			array(
				'name'         => 'MyVariation',
				'label'        => __( 'My variation', 'twentytwentyfour' ),
				'inline_style' => '
				.is-style-MyVariation {
					background-color: red;
				}',
			)
		);
```

- Go to the editor and verify there is a variation for the group block called "My variation" and that applying it would set the background to red.
- Create the file `styles/partial.json` with the following contents:

```json
{
    "version": 2,
    "title": "My variation",
    "styles": {
            "color": {
                    "background": "blue",
                    "text": "white"
            }
    }
}
```

- Go to the editor. You'll see two variations registered. 

Captura de ecrã 2024-06-13, às 23 15 33

- Now, update the `partial.json` to include a `slug` that matches the variation name provided by `register_block_style` call:

```json
{
    "version": 2,
    "title": "My variation",
    "slug": "MyVariation",
    "styles": {
            "color": {
                    "background": "blue",
                    "text": "white"
            }
    }
}
```

- Go to the editor and verify that there is only one variation again:

Captura de ecrã 2024-06-13, às 23 12 48

- Confirm that the partial's style gets applied correctly as it takes precedence over the PHP block style registration.

### DIFF
--- a/lib/block-supports/block-style-variations.php
+++ b/lib/block-supports/block-style-variations.php
@@ -455,7 +455,7 @@ function gutenberg_register_block_style_variations_from_theme_json_data( $variat
 		 * Block style variations read in via standalone theme.json partials
 		 * need to have their name set to the kebab case version of their title.
 		 */
-		$variation_name  = $have_named_variations ? $key : _wp_to_kebab_case( $variation['title'] );
+		$variation_name  = $have_named_variations ? $key : ( $variation['slug'] ?? _wp_to_kebab_case( $variation['title'] ) );
 		$variation_label = $variation['title'] ?? $variation_name;
 
 		foreach ( $supported_blocks as $block_type ) {
--- a/lib/block-supports/block-style-variations.php
+++ b/lib/block-supports/block-style-variations.php
@@ -244,7 +244,7 @@ function gutenberg_resolve_block_style_variations( $variations ) {
 		 * Block style variations read in via standalone theme.json partials
 		 * need to have their name set to the kebab case version of their title.
 		 */
-		$variation_name = $have_named_variations ? $key : _wp_to_kebab_case( $variation['title'] );
+		$variation_name = $have_named_variations ? $key : ( $variation['slug'] ?? _wp_to_kebab_case( $variation['title'] ) );
 
 		foreach ( $supported_blocks as $block_type ) {
 			// Add block style variation data under current block type.
--- a/backport-changelog/6.6/6824.md
+++ b/backport-changelog/6.6/6824.md
@@ -1,0 +1,3 @@
+https://github.com/WordPress/wordpress-develop/pull/6824
+
+* https://github.com/WordPress/gutenberg/pull/62550
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -357,6 +357,7 @@ class WP_Theme_JSON_Gutenberg {
 		'styles',
 		'templateParts',
 		'title',
+		'slug',
 		'version',
 	);
 
--- a/phpunit/block-supports/block-style-variations-test.php
+++ b/phpunit/block-supports/block-style-variations-test.php
@@ -94,6 +94,22 @@ class WP_Block_Supports_Block_Style_Variations_Test extends WP_UnitTestCase {
 			),
 		);
 
+		/*
+		 * This style is to be deliberately overwritten by the theme.json partial
+		 * See `phpunit/data/themedir1/block-theme/styles/block-style-variation-with-slug.json`.
+		 */
+		register_block_style(
+			'core/group',
+			array(
+				'name'       => 'WithSlug',
+				'style_data' => array(
+					'color' => array(
+						'background' => 'whitesmoke',
+						'text'       => 'black',
+					),
+				),
+			)
+		);
 		register_block_style(
 			'core/group',
 			array(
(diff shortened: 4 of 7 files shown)


lib/block-supports/block-style-variations.php
```diff
@@ -455,7 +455,7 @@ function gutenberg_register_block_style_variations_from_theme_json_data( $variat
 		 * Block style variations read in via standalone theme.json partials
 		 * need to have their name set to the kebab case version of their title.
 		 */
-		$variation_name  = $have_named_variations ? $key : _wp_to_kebab_case( $variation['title'] );
+		$variation_name  = $have_named_variations ? $key : ( $variation['slug'] ?? _wp_to_kebab_case( $variation['title'] ) );
```
oandregal, line 458: For most scenarios, just kebab-casing the `title` will be enough and that's a great devexp we may want to keep. The `slug` (could also use `name`) is a way to make it work for any scenario.

aaronrobertshaw:
> There's an issue with styles generation: when the slug is provided the styles are not generated.

The resolution of block style variations from partials needed to be tweaked in the same way the registration was so their key is consistent. I pushed a tweak in https://github.com/WordPress/gutenberg/pull/62550/commits/ebf216269210b27d65d4272ed144a12f71bf20f3

I'll give this a more thorough review shortly. Then create a backport PR, update the backport changelog etc.

aaronrobertshaw (approved):
Thanks for putting this together @oandregal 👍 

This tests well for me resolving both the "duplicate" style option and variation merging issues.

I've created a [backport](https://github.com/WordPress/wordpress-develop/pull/6824) for the PHP changes and added a backport changelog entry to this PR.

aaronrobertshaw:
Before we merge this one, we might need to update a couple of tests for the block style variations support and theme.json resolver. I'll sort those out on https://github.com/WordPress/wordpress-develop/pull/6824 and then bring those back across to this PR.

aaronrobertshaw:
The unit test updates are pretty basic but cover:
- That a partial using a slug overrides one supplied directly through `register_block_style` (duplicate UI option issue)
- That a partial supplied block style variation is registered and resolved by its `slug` (styles not generating correctly if it relied on title)

sirreal:
It looks like tests have been failing on trunk since this merged and these changes look suspect. @aaronrobertshaw @oandregal have you noticed the test failures? Can you confirm whether they're related or not?

```
1) Gutenberg_REST_Block_Editor_Settings_Controller_Test::test_get_items
Undefined array key "WithSlug"
```

[Here's a recent example run.](https://github.com/WordPress/gutenberg/actions/runs/9513702454/job/26224543289)

aaronrobertshaw:
The error does appear related, we'll look into it 👀

aaronrobertshaw:
The changes to the tests have been reverted in https://github.com/WordPress/gutenberg/pull/62579 while we investigate the issue via https://github.com/WordPress/gutenberg/pull/62587

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 7cfc6c5265

PR #62553, opened 2024-06-13
Update api-reference.md

ptesei:
add escape to the wp_interactivity_data_wp_context  see ticket https://core.trac.wordpress.org/ticket/61430


Mamaduka:
Thanks for contributing, @ptesei!

It looks like there's no consensus on the linked ticket. I'm going to close this PR since it's not actionable at the moment.

PR #62554, opened 2024-06-13
Fix: Unable to remove custom colors on safari.

### LINKED ISSUE #62261: Global Styles: Can't remove a custom color from palette in Safari
### Description

1. Use Safari.

### Environment info

_No response_

Yes

Yes

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/62261

To be honest I don't know the root cause of the issue. It seems to happen only on Safari. I tried many many fixes and this combination of changes seem to work:
- Increases the z-index of the remove button to be higher than sidebar.
- Don't pass an onClick event to the parent if we are already editing that color.
- Stop the event propagation on the remove button.

cc: @noisysocks and @richtabor in case you can confirm if the issue is also fixed for you.

## Testing Instructions
Use Safari.
Go to Appearance → Editor → Colors → Palette.

### DIFF
--- a/packages/base-styles/_z-index.scss
+++ b/packages/base-styles/_z-index.scss
@@ -92,6 +92,8 @@ $z-layers: (
 	// For larger views, the wp-admin navbar dropdown should be at top of
 	// the Publish Post sidebar.
 	".editor-post-publish-panel {greater than small}": 99998,
+	// above the sidebar
+	".components-palette-edit__remove-button": 100010,
 
 	// For larger views, the wp-admin navbar dropdown should be on top of
 	// the multi-entity saving sidebar.
--- a/packages/components/src/palette-edit/index.tsx
+++ b/packages/components/src/palette-edit/index.tsx
@@ -201,7 +201,6 @@ function Option< T extends Color | Gradient >( {
 		<PaletteItem
 			className={ isEditing ? 'is-selected' : undefined }
 			as={ isEditing ? 'div' : 'button' }
-			onClick={ isEditing ? undefined : onStartEditing }
 			aria-label={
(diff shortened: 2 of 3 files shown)


noisysocks:
This doesn't seem to fix it for me. I confirmed that the button has a z-index.

https://github.com/WordPress/gutenberg/assets/612155/45ae2fbe-5d5d-4641-b611-7c8fe73e72ac

aatanasovdev:
It seems the issue is related to the [onStopEditing](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/palette-edit/index.tsx#L329-L333) method. It is triggered first, and because it [clears the currently edited element](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/palette-edit/index.tsx#L331), the Remove color method isn't triggered (it relies on `isEditing`, which is false if `editingElement` is null).

That means, the `onStopEditing` shouldn't be triggered when the [RemoveButton](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/palette-edit/index.tsx#L249-L254) is clicked.

On the other side, the `onStopEditing` depends on the experimental [useFocusOutside](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts) hook. 

Based on my review, it looks like the hook should be optimized so it doesn't trigger when a child component with an `onClick` event is clicked.

jorgefilipecosta:
Thank you for the tests @noisysocks and thank you for the debugging @aatanasovdev. I'm exploring a different fix https://github.com/WordPress/gutenberg/pull/62628.

PR #61522, opened 2024-05-09
Convert warning package to TypeScript

jpstevens:
## What?
This PR converts the `warning` package to TypeScript.

## Why?
Ensures package is fully type checked.

## How?
- converted files to `.ts`
- added explicit return types for functions in `index` and `utils` files

## Testing Instructions
`npm run test:unit -- packages/warning` tests pass
`npm run build:package-types` returns a zero exit code

PR #62557, opened 2024-06-13
Convert warning package to TS

up1512001:
## What?
This PR converts the warning package to TypeScript.

## Why?

## How?
- converted files to .ts
- added explicit return types for functions in index and utils files.

## Testing Instructions
- `npm run test:unit -- packages/warning` tests pass

## reference of this PR

Closes https://github.com/WordPress/gutenberg/pull/61522

### DIFF
--- a/packages/warning/src/index.ts
+++ b/packages/warning/src/index.ts
@@ -3,15 +3,12 @@
  */
 import { logged } from './utils';
 
-function isDev() {
-	// eslint-disable-next-line @wordpress/wp-global-usage
-	return globalThis.SCRIPT_DEBUG === true;
+function isDev(): boolean {
+	return typeof SCRIPT_DEBUG !== 'undefined' && SCRIPT_DEBUG === true;
 }
 
 /**
  * Shows a warning with `message` if environment is not `production`.
- *
- * @param {string} message Message to show in the warning.
  *
  * @example
  * ```js
@@ -25,7 +22,7 @@ function isDev() {
  * }
  * ```
  */
-export default function warning( message ) {
+export default function warning( message: string ): void {
 	if ( ! isDev() ) {
 		return;
 	}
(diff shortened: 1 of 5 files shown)


packages/warning/src/index.ts
```diff
@@ -3,15 +3,19 @@
...
-function isDev() {
-	// eslint-disable-next-line @wordpress/wp-global-usage
-	return globalThis.SCRIPT_DEBUG === true;
+function isDev(): boolean {
+	return (
+		// eslint-disable-next-line @wordpress/wp-global-usage
+		typeof globalThis.SCRIPT_DEBUG !== 'undefined' &&
+		// eslint-disable-next-line @wordpress/wp-global-usage
+		globalThis.SCRIPT_DEBUG === true
+	);
```
sirreal: The implementation should not change here.

packages/warning/src/test/index.ts
```diff
@@ -4,9 +4,7 @@
 import warning from '..';
 import { logged } from '../utils';
 
-// eslint-disable-next-line eslint-comments/disable-enable-pair
```
sirreal: This file doesn't seem to require any changes, can we revert them?

sirreal (changes requested):
Thanks for picking up this work. I've left some feedback, let's try to keep these changes as minimal as possible and ensure we're not changing any implementation when adding types unless it's justified.

packages/warning/src/test/index.ts
up1512001: @sirreal change the isDev implementation back to original and revert the test file. Please review it now.

sirreal:
This looks good, will you add a changelog entry please? Then I think we can merge.

https://github.com/WordPress/gutenberg/pull/62569 has an example we can follow for this package.

up1512001:
@sirreal added a changelog entry.

sirreal (approved):
Thanks!

PR #59985, opened 2024-03-19
Extensibility: support the PluginDocumentSettingPanel slot in the site editor

youknowriad:
Related #52632 

## What?

This PR is going to set a precent about how we merge extensibility APIs between post and site editors, so we need to review and test properly.

The idea is that all the slots that exit in either post or site editors can be shared through the `editor` package. That way, a plugin registering a fill can work both in the post editor and site editor (assuming the plugin loads its JS in both pages).

## How?

- Moves the PluginDocumentSettingPanel from `@wordpress/edit-post` to. `@wordpress/editor`
- Adds the slot to the site editor as well.

## Testing Instructions

You should be able to paste something like the following in both post and site editors and have the panel show up in the sidebar of pages.

```js
var el = React.createElement;
var __ = wp.i18n.__;
var registerPlugin = wp.plugins.registerPlugin;
var PluginDocumentSettingPanel = wp.editor.PluginDocumentSettingPanel;

function MyDocumentSettingPlugin() {
	return el(
		PluginDocumentSettingPanel,
		{
			className: 'my-document-setting-plugin',
			title: 'My Panel',
			name: 'my-panel',
		},
		__( 'My Document Setting Panel' )
	);
}

registerPlugin( 'my-document-setting-plugin', {
	render: MyDocumentSettingPlugin,
} );
```

PR #62558, opened 2024-06-13
Fixed incorrect package reference

aaronware:
## What?

Fixed an incorrect package reference @wordpress/editor vs @wordpress/edit-post in order to load the component

## Why?

Help with users when they are copying and pasting examples

## How?

It fixes it

## Testing Instructions

Copy and paste fix to test


aaronware:
Perhaps there aren't so many tests needed for a documentation edit?

Screenshot by Dropbox Capture

t-hamano (commented):
Thanks for the PR!

This slot used to be in the `@wordpress/edit-post` package, but has been moved to the `@wordpress/editor` package.

See #59985 for details.

I believe the current code example is correct, so let me close this PR.

PR #62560, opened 2024-06-13
Fixing minor syntax in DataView example code

aaronware:
## What?

Fixing a syntax error

## Why?

Helping users copy/paste code examples

## How?

Adding a comma to an array of objects


t-hamano (approved):
Nice catch 👍

PR #62199, opened 2024-05-31
Fix UI order for theme.json spacing sizes

ajlende:
## What?

Solves the ordering issue for spacing sizing presets in v2 theme.json files that don't follow the `'10'`, `'20'`, etc. convention for slugs.

~The labels in the range slider remain numbers as, I believe, that was the intention of the feedback that #44247 fixed on the server instead of in the UI. It was a bug if the slider didn't show numbers.~

The labels are kept as t-shirt sizes when using the slider now.

## Why?

Fixes #62194

> With Gutenberg 18.5 RC1 active, numbers are now being used in areas such as padding, and the mapping is out of order:
> 
> 1 = var(--wp--preset--spacing--large)
> 2 = var(--wp--preset--spacing--medium)
> 3 = var(--wp--preset--spacing--small)
> 4 = var(--wp--preset--spacing--x-large)
> 5 = var(--wp--preset--spacing--x-small)

## How?

It works by only sorting presets when they need to be merged.

See https://github.com/WordPress/gutenberg/issues/62194#issuecomment-2142933855 for details, but to summarize the constraints of the problem:

- Any solution must to work for both v2 and v3 as there is no distinction in the code (v2 shaped theme.json data automatically get converted to the v3 shape before anything can be done with the data).
- Sorting of multi-origin presets must be done in the UI as they are separated out on the server and cannot be merged until they're loaded in the JS.
- Renaming of multi-origin presets must be done in the UI for similar reasons.
- v2 themes with `spacingSizes` slugs that do not conform to the number convention must use the ordering defined in the theme.json file.

V2 theme.json files will _always_ only have one `spacingSizes` array, either the one defined in theme.json the one generated (in order) from `spacingSizes` values.

V2 theme.json files will _always_ only have one origin of `spacingSizes` defined.

V3 theme.json files allow both `spacingScale` and `spacingSizes` to be defined at the same time and the results merged into one array. When this is done on the server, we can sort the merged values. These merged values become `spacingSizes` for their origin in the frontend.

V3 theme.json files allow `default`, `theme`, and `custom` origins for `spacingSizes` that are separated until they reach the frontend. Only sort them in the frontend if more than one origin is defined.

## Testing Instructions

Add this in a `"version": 2` theme.json file.

```json
"spacing": {
	"spacingSizes": [
		{
			"name": "xSmall",
			"size": "clamp(10px, 2vw, 20px)",
			"slug": "x-small"
		},
		{
			"name": "Small",
[...]

PR #62567, opened 2024-06-14
Sort spacing sizes when all slugs begin numerically

### LINKED ISSUE #62520: theme.json schema change for spacingSizes pattern incorrect
With this change to `spacingSizes` and `theme.json` v3, the schema was changed to specifically add a numeric string pattern instead of any string for the `slug` property: https://github.com/WordPress/gutenberg/pull/61842

If I use a non-numeric string for `slug`, this gave no warning in the past:

```json
"settings": {
	"spacing": {
		"spacingSizes": [
			{
				"name": "Base",
				"size": "clamp(1.06rem, 0.37cqi + 0.95rem, 1.19rem)",
				"slug": "base"
			}
		]
	}
}
```

Now, VS Code or other schema checkers will show a warning that the `slug` doesn't match the pattern.

I am concerned that theme authors who were previously using non-numeric strings in the past will think they must migrate/upgrade their spacing sizes in some way, which would not be something we'd want to ask theme authors to do.

Not to mention that the Theme Handbook has shown non-numeric slugs as valid examples since last year: https://developer.wordpress.org/themes/global-settings-and-styles/settings/spacing/

I think the new `description` is fine:

> Unique identifier for the space size preset. Will be sorted numerically. For best cross theme compatibility these should be in the form '10','20','30','40','50','60', etc. with '50' representing the 'Medium' size step.

However, I would remove the `pattern` from the schema because non-numeric strings can and still do work in v3.

CC: @ajlende

ajlende:
## What?

Changes the sorting heuristic for `spacingSizes` from having more than one origin to all slugs from all origins beginning numerically.

Removes the numeric requirement for slugs in the schema.

## Why?

Fixes #62520

## How?

> The heuristics for sorting that I added in https://github.com/WordPress/gutenberg/pull/62199 were focused on figuring out v2 vs v3 schemas. If we do away with heuristics and simply sort when all slugs start with a number, the sorting becomes an opt-in feature by the theme using numeric slugs. The numeric slugs become a bonus feature rather than a requirement for correct operation.
>
> We'll still want to update the UI to have sections, but I don't think that's critical to fix until we have UI for adding user origin spacing presets. At that point, if a user adds a new preset that isn't numeric, then that will effectively opt-out of the merging and sorting of presets even if the theme is using numeric slugs.

https://github.com/WordPress/gutenberg/issues/62520#issuecomment-2167205559

## Testing Instructions

Try numeric and non-numeric slugs with and without defaults enabled to see if sorting happens when you would expect.

### DIFF
--- a/packages/block-editor/src/components/spacing-sizes-control/hooks/use-spacing-sizes.js
+++ b/packages/block-editor/src/components/spacing-sizes-control/hooks/use-spacing-sizes.js
@@ -44,15 +44,8 @@ export default function useSpacingSizes() {
 			...defaultSizes,
 		];
 
-		// Only sort if more than one origin has presets defined in order to
-		// preserve order for themes that don't include default presets and
-		// want a custom order.
-		if (
-			( customSizes.length && 1 ) +
-				( themeSizes.length && 1 ) +
-				( defaultSizes.length && 1 ) >
-			1
-		) {
+		// Using numeric slugs opts-in to sorting by slug.
+		if ( sizes.every( ( { slug } ) => /^[0-9]/.test( slug ) ) ) {
 			sizes.sort( ( a, b ) => compare( a.slug, b.slug ) );
 		}
 
--- a/schemas/json/theme.json
+++ b/schemas/json/theme.json
@@ -444,9 +444,8 @@
 										"type": "string"
 									},
 									"slug": {
-										"description": "Unique identifier for the space size preset. Will be sorted numerically. For best cross theme compatibility these should be in the form '10','20','30','40','50','60', etc. with '50' representing the 'Medium' size step.",
-										"type": "string",
-										"pattern": "^[0-9].*$"
+										"description": "Unique identifier for the space size preset. For best cross theme compatibility these should be in the form '10','20','30','40','50','60', etc. with '50' representing the 'Medium' size step. If all slugs begin with a number they will be merged with default and user slugs and sorted numerically.",
+										"type": "string"
 									},
 									"size": {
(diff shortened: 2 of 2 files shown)


packages/block-editor/src/components/spacing-sizes-control/hooks/use-spacing-sizes.js
```diff
@@ -44,15 +44,8 @@ export default function useSpacingSizes() {
...
-		// preserve order for themes that don't include default presets and
-		// want a custom order.
-		if (
-			( customSizes.length && 1 ) +
-				( themeSizes.length && 1 ) +
-				( defaultSizes.length && 1 ) >
-			1
-		) {
+		// Using numeric slugs opts-in to sorting by slug.
+		if ( sizes.every( ( { slug } ) => /^[0-9]/.test( slug ) ) ) {
```
ajlende, line 48: I have this just checking that it _starts with_ a number, allowing slugs like `50-medium` to be sorted, but maybe it would make more sense to only sort slugs that are completely numeric.

scruffian: I think this is good

scruffian (approved):
Look good, and works as expected in my testing.

justintadlock (approved):
Changes look good to me. Sorting works, and I think it's OK to continue sorting when the slug begins with the number.

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 8ffc66ffd7

Issue #62559, opened 2024-06-13
[Template Parts] PHP Warning thrown when `get_block_file_template()` returns `null`.


hellofromtonya:
### Description

The function `get_block_file_template()` can return `null` when a template part does not exist. When this happens, a PHP Warning is thrown in `render_block_core_template_part()` when attempting to use `$block_template` (assigned from `get_block_file_template()`) as an object to get the `content` property. 

>Warning: Attempt to read property "content" on null in .../wp-includes/blocks/template-part.php on line 73

A check is needed before using the variable as object.

This issue was first raised by @carolinan [here](https://github.com/WordPress/gutenberg/pull/54354#issuecomment-1892044231) as part of an investigation to produce a null to non-nullable deprecation notice in the same function for the same scenario of a missing template part.

It was raised in testing of this Trac ticket (to confirm PR #54354 resolved the deprecation notice) https://core.trac.wordpress.org/ticket/59318#comment:15.

### Step-by-step reproduction instructions

Setup:
1. Use PHP 8.1 or higher.
2. Use WP 6.3.4 when attempting to reproduce the reported issue.
3. Use WP 6.6 Beta 2 or newer to test if the bug is fixed in Core.

Testing Steps:
1. Open wp-content/themes/twentytwentyfour/templates/home.html file in your favorite code editor.
2. In the footer template-part, change the slug to something that does not exist:
	
3. `$is_debug` needs to be false. 3 ways to accomplish this:
   a. Set `WP_DEBUG` to false in `wp-config.php`.
   b. Set `WP_DEBUG_DISPLAY` to false in `wp-config.php`.
   c. Temporarily modify the [following line of code](https://core.trac.wordpress.org/browser/tags/6.3.4/src/wp-includes/blocks/template-part.php#L110) in wp-includes/blocks/template-part.php to hardcode it to false. This approach lets you keep WP_DEBUG and WP_DEBUG_DISPLAY set to true to further see the deprecation notice in the front-end.
```php   
$is_debug = false; //WP_DEBUG && WP_DEBUG_DISPLAY;
```
4. Open the home page in the front. 🐞 A Deprecation notice is thrown (viewable in the server logs, debug.log, and/or on the screen in your browser).

### Screenshots, screen recording, code snippet

Test Report and details are found here in Core Trac https://core.trac.wordpress.org/ticket/59318#comment:15.

### Environment info

- WordPress 6.6 Beta 2 or newer
- PHP 8.1 or newer

[...]

Issue #59318, opened 2024-02-23
Richtext : "multiline=p" property is not working in latest WP version


md-kushang-tailor:
### Issue:
I'm facing issues with the Richtext component When I'm used `"multiline=p"` property in latest WP version - 6.4.x.

### The scenario I've:
I've one custom GB slider block that includes some Richtext for contents, Gallery for images and links as multi-level functionality.

**Multi-level functionality:**
- _On click of Plus Button,_ The slider will increase by one with all the empty/placeholder contents & components.
- _On click of Minus/Remove Button_, The current active slider will remove completely.

After updating WP to the latest version the `"multiline=p"` property not working properly. I don't know why this is happened. But, I've many posts to depends on the `"multiline=p"` property, Because using this property, The contents of the slider's automatically converted all the copied heading tags & List items into different paragraph tags. But, Now in latest WP version, When I added copied contents, It's throw the content like `[object Object]`.

I have a below code, After WordPress 6.3 > when I published WordPress 6.4.x and copy any heading like text from doc file to WordPress it gives below error.
- **Video:** https://share.cleanshot.com/YCXhp7K4t5Y5SFXLLvXT
- **Code:** https://share.cleanshot.com/VSKrc2p1H5wpcdLcZ5J9

In other hand, I've some document files ready for this slider's content so I just copied all the contents from that document file and added here in this block.

Is there any way like it should not add the above content when multiline enabled?

**Note:** Above only happens when multiline enabled in rich text in WordPress latest version **( 6.4.x )**.

As it seems like multiline prop is removed after WordPress 6.4, and it suggests alternative of inner blocks. While I have serverside render functionality with custom slider so my rich text is repeating in loop. And I assume that this will not work with inner blocks in loop with serverside render.

**My block example:** https://share.cleanshot.com/wRh8VvjdFgSpDKXYwXrb

Please suggest any other solution if any can work for above situation.

Thanks!

### Custom Gallery slider Block
I have serverside render functionality with custom slider so my rich text is repeating in loop.

### The Code example:
```
<RichText
	tagName="p"
	placeholder={__('Heading', 'test')}
	value={heading}
	onChange={heading => setAttributes({ heading })}
	multiline="p"
/>
```

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

Mamaduka:
[...]

PR #62568, opened 2024-06-14
Fix PHP Warning thrown when get_block_file_template() returns null

Rajinsharwar:
## What?
This is a refresh PR for the core trac ticket: [#59318](https://core.trac.wordpress.org/ticket/59318), and for the GH issue #62559 

## Why?
To fix the deprecation notice passing null to non-nullable when referring to missing template parts.

## How?
N/A

## Testing Instructions
1. Open wp-content/themes/twentytwentyfour/templates/home.html file in your favorite code editor.
2. In the footer template-part, change the slug to something that does not exist:
	
3. $is_debug needs to be false. Two ways to accomplish this:
    a. Set WP_DEBUG to false in wp-config.php.
    b . Set WP_DEBUG_DISPLAY to false in wp-config.php`.
    c. Temporarily modify the [following line of code](https://core.trac.wordpress.org/browser/tags/6.3.4/src/wp-includes/blocks/template-part.php#L110) in wp-includes/blocks/template-part.php to hardcode it to false. This approach lets you keep WP_DEBUG and WP_DEBUG_DISPLAY set to true to further see the deprecation notice in the front-end.
$is_debug = false; //WP_DEBUG && WP_DEBUG_DISPLAY;	
4. Open the home page in the front. 🐞 A Deprecation notice is thrown (viewable in the server logs, debug.log, and/or on the screen in your browser).

Props @hellofromtonya 

## Screenshots or screencast 
N/A

### DIFF
--- a/packages/block-library/src/template-part/index.php
+++ b/packages/block-library/src/template-part/index.php
@@ -114,7 +114,11 @@ function render_block_core_template_part( $attributes ) {
 	$is_debug = WP_DEBUG && WP_DEBUG_DISPLAY;
 
 	if ( is_null( $content ) ) {
-		if ( $is_debug && isset( $attributes['slug'] ) ) {
+		if ( $is_debug ) {
+			if ( ! isset( $attributes['slug'] ) ) {
+				// If there is no slug this is a placeholder and we dont want to return any message.
+				return '';
+			}
 			return sprintf(
 				/* translators: %s: Template part slug. */
 				__( 'Template part has been deleted or is unavailable: %s' ),


packages/block-library/src/template-part/index.php
```diff
@@ -114,7 +114,11 @@ function render_block_core_template_part( $attributes ) {
 	$is_debug = WP_DEBUG && WP_DEBUG_DISPLAY;
 
 	if ( is_null( $content ) ) {
-		if ( $is_debug && isset( $attributes['slug'] ) ) {
+		if ( $is_debug ) {
+			if ( ! isset( $attributes['slug'] ) ) {
+				// If there is no slug this is a placeholder and we dont want to return any message.
+				return '';
+			}
```
cbravobernal, line 121: I guess this check should be done outside of `$is_debug`. And only send the error message in debug mode.

youknowriad:
I think this has been solved already. Thanks for your contribution.

PR #61541, opened 2024-05-09
Convert blob package to TypeScript

jpstevens:
## What?
This PR converts the `blob` package to TypeScript.

## Why?
Ensures package is fully type checked.

## How?
- converted files to `.ts`
- explicitly typed variables and functions in `index` file
- added `@ts-ignore` statements to tests that call functions in ways that are _not_ type-compliant

## Testing Instructions
`npm run test:unit -- packages/blob` tests pass
`npm run build:package-types` returns a zero exit code

PR #62569, opened 2024-06-14
Convert blob package to TS

up1512001:
## What?
This PR converts the `blob` package to TypeScript.

## Why?

## How?
- converted files to .ts
- explicitly typed variables and functions in index file
- added @ts-ignore statements to tests that call functions in ways that are not type-compliant

## Testing Instructions
- `npm run test:unit -- packages/blob` tests pass

### reference PR

Closes https://github.com/WordPress/gutenberg/pull/61541

### DIFF
--- a/packages/blob/src/index.ts
+++ b/packages/blob/src/index.ts
@@ -1,13 +1,11 @@
-type BlobPart = BufferSource | Blob | string;
-
 const cache: Record< string, File > = {};
 
 /**
  * Create a blob URL from a file.
  *
- * @param {File} file The file to create a blob URL for.
+ * @param file The file to create a blob URL for.
  *
- * @return {string} The blob URL.
+ * @return The blob URL.
  */
 export function createBlobURL( file: File ): string {
 	const url = window.URL.createObjectURL( file );
@@ -22,9 +20,9 @@ export function createBlobURL( file: File ): string {
  * `createBlobURL` and not removed by `revokeBlobURL`, otherwise it will return
  * `undefined`.
  *
- * @param {string} url The blob URL.
+ * @param url The blob URL.
  *
- * @return {File|undefined} The file for the blob URL.
+ * @return The file for the blob URL.
  */
 export function getBlobByURL( url: string ): File | undefined {
 	return cache[ url ];
@@ -35,9 +33,9 @@ export function getBlobByURL( url: string ): File | undefined {
  * `createBlobURL` and not removed by `revokeBlobURL`, otherwise it will return
  * `undefined`.
  *
- * @param {string} url The blob URL.
+ * @param url The blob URL.
  *
- * @return {string|undefined} The blob type.
+ * @return The blob type.
  */
 export function getBlobTypeByURL( url: string ): string | undefined {
 	return getBlobByURL( url )?.type.split( '/' )[ 0 ]; // 0: media type , 1: file extension eg ( type: 'image/jpeg' ).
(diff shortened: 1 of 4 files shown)


up1512001:
Added `type BlobPart = BufferSource | Blob | string;` as eslint is giving error of undefined type 
`The type 'BlobPart' is undefined.eslint`[jsdoc/no-undefined-types](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/no-undefined-types.md#repos-sticky-header)

packages/blob/src/index.ts
```diff
@@ -1,7 +1,6 @@
-/**
- * @type {Record<string, File|undefined>}
- */
-const cache = {};
+type BlobPart = BufferSource | Blob | string;
+
```
sirreal: We shouldn't need this, it's part of the DOM types.

```suggestion
```

packages/blob/src/index.ts
```diff
@@ -94,7 +93,11 @@ export function isBlobURL( url ) {
  * @param {BlobPart} content     File content (BufferSource | Blob | string).
```
sirreal: Eslint complained here about the param type, but this has been converted to TypeScript so we need to remove all the types from JSDoc.

sirreal (changes requested):
Thanks for picking up this work. All the types need to be removed from JSDoc because they're moving into TypeScript annotations.

packages/blob/src/index.ts
up1512001: @sirreal please review now I have removed the param type.

packages/blob/src/test/index.ts
```diff
@@ -13,6 +13,8 @@ describe( 'isBlobURL', () => {
 	} );
 
 	it( 'returns false if the url is not defined', () => {
+		// calling isBlobURL without a URL is not type compliant, so ignore it
+		// @ts-ignore
```
sirreal: Can we use `@ts-expect-error` everywhere that we _know_ the types are wrong (and expect it)?

```suggestion
		// @ts-expect-error: This is not a valid call according to types.
```

Ideally we can target only the violating line:

```ts
expect(
// @ts-expect-error
isBlobURL()
).toBe( … )
```

packages/blob/src/test/index.ts
```diff
@@ -23,6 +25,8 @@ describe( 'getBlobTypeByURL', () => {
 	} );
 
 	it( 'returns undefined if the url is not defined', () => {
+		// calling getBlobTypeByURL without a URL is not type compliant, so ignore it
+		// @ts-ignore
 		expect( getBlobTypeByURL() ).toBe( undefined );
```
sirreal: We can also use a better undefined assertion here since we're touching this file:

```suggestion
		// calling getBlobTypeByURL without a URL is not type compliant, so ignore it
		// @ts-ignore
		expect( getBlobTypeByURL() ).toBeUndefined();
```

packages/blob/src/test/index.ts
```diff
@@ -36,17 +42,18 @@ describe( 'downloadBlob', () => {
...
 	jest.spyOn( document.body, 'removeChild' );
 	beforeEach( () => {
 		// Can't seem to spy on these static methods. They are `undefined`.
 		// Possibly overwritten: https://github.com/WordPress/gutenberg/blob/trunk/packages/jest-preset-default/scripts/setup-globals.js#L5
-		window.URL = {
-			createObjectURL,
-			revokeObjectURL,
-		};
+		window.URL.createObjectURL = createObjectURL;
+		window.URL.revokeObjectURL = revokeObjectURL;
```
sirreal: This is a change in behavior. Before, the object was replaced with a different object. Later, the original object was restored.

In this change, some properties of the object are changed, mutating the original object. That means that the restoration does not actually do anything.

sirreal (approved):
Thanks!

up1512001:
I will note these points for the next TS conversion PRs.
Thanks @sirreal

PR #62572, opened 2024-06-14
Autoformat e2e-tests package JavaScript

sirreal:
## What?

- Remove prettier ignore for e2e-tests
- Autoformat repository

## Why?

There's plenty of JavaScript in the e2e-tests package, I'm not sure why it was included in our prettier ignore. Let's format it.

## How?
Remove the ignore rule and apply autoformatting.

## Testing Instructions
CI passes.

PR #54458, opened 2023-09-14
Mobile Release v1.104.0

fluiddot:
## Description
Release 1.104.0 of the react-native-editor and Gutenberg-Mobile.

For more information about this release and testing instructions, please see the related Gutenberg-Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/6204

Issue #53653, opened 2023-08-14
Font Library: How to add lib-font package (or equivalent).


matiasbenedetto:
## What ?
We need to add the [lib-font package](https://github.com/Pomax/lib-font) (or a better equivalent) to read font assets metadata in the Font Library.

## Current situation: 
We are adding the files library files in the codebase, adding minor changes:

See these how we are adding the library files on [Font Library Frontend PR](https://github.com/WordPress/gutenberg/pull/53884):
https://github.com/WordPress/gutenberg/tree/d01a98d03724c90fb08a47a2504608e52c78b800/packages/edit-site/lib

## Context
In Create Block Theme plugin, we use the same library to read the font assets metadata (font name, font weight, font style, etc) without changing the library source code. To avoid compiling errors, we needed to add [custom Webpack config options](https://github.com/WordPress/create-block-theme/blob/trunk/webpack.config.js).

@youknowriad suggested that we should avoid changing the Webpack configuration file for Gutenberg. Because of that, we are using the bundling approach, changing a few things in the library's source code (see the comments in the  [Fonts Library frontend PR](https://github.com/WordPress/gutenberg/tree/34dd73eaa26a461897413f60fea6cb065e6c27ae/packages/edit-site/lib) ).

## More Context
Added a question in the lib-font repo about the Webpack bundling: https://github.com/Pomax/lib-font/issues/135

## Alternatives
It is not a requirement to use [Lib-font](https://github.com/Pomax/lib-font). We can try another alternative.
The pros of this lib:
- It is used on Create Block Theme plugin, and it is working OK
- The most lightweight alternative I found.

The con seems to be this issue.

## Tracking issue
https://github.com/WordPress/gutenberg/issues/52698

mikachan:
Do we have to bundle the package in Gutenberg? I'm wondering if we could call the lib-font functionality from a CDN like unpkg.com or jsDelivr, e.g. from https://www.unpkg.com/lib-font@2.4.0/lib-font.js / https://cdn.jsdelivr.net/npm/lib-font@2.4.0/lib-font.js.

jeffikus:
As far as I'm aware, MIT license is compatible with the GPL, but I'm trying to find out more about this.

youknowriad:
Judging by the response here https://github.com/Pomax/lib-font/issues/135#issuecomment-1593316978

[...]

PR #62573, opened 2024-06-14
Edit site: Do not autoformat edit-site/lib

sirreal:
## What?

Disable and revert autoformatting for `edit-site/lib` vendored files. These appear to be compiled external files that should not be auto-formatted.

## Why?

In #62572 I noticed some suspicious files being autoformatted so investigated. It seems that these files were autoformatted in https://github.com/WordPress/gutenberg/pull/54458 ([see this comment](https://github.com/WordPress/gutenberg/pull/54458#pullrequestreview-1628654375)) so I reverted that autoformatting as well. According to git, that was the last change to the `edit-site/lib` directory so I restored the directory to it's state prior to that commit with `git checkout b06240a7ce93caf3d1c410267634b3d3d6d5a514~ -- packages/edit-site/lib`.

[This also adds a git attribute to indicate that the directory contains vendored code](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md#vendored-code). This should improve GitHub statistics for the project.

## Testing Instructions

`npm run format` does not format anything in `packages/edit-site/lib` (on trunk it would).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,6 @@
 *.jpg binary
 *.png binary
 *.ttf binary
+
+# Directories with vendored code.
+packages/edit-site/lib/** linguist-vendored
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@ packages/block-serialization-spec-parser/parser.js
 packages/react-native-editor/bundle
 packages/url/src/test/fixtures
 vendor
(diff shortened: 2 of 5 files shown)


sirreal:
Related: https://github.com/WordPress/gutenberg/issues/53653

youknowriad (approved):

Issue #62574, opened 2024-06-14
Author Not Visible in Post Editor on Multisite Network When Super Admin Creates Post


rohitmathur-7:
### Description

This issue occurs on a multisite network, So on a sub-site when a super admin creates a post and the super admin does not have their user registered in the particular sub-site then on the post editor we are not able to see the author's name.

### Step-by-step reproduction instructions

- Login to your WordPress site using a Super admin user.
- Go on a sub-site where the Super admin user you are using does not have its user registered.
- Open a new post or page.
- Check the author in the right panel of the editor.

### Screenshots, screen recording, code snippet

https://github.com/WordPress/gutenberg/assets/62138457/1d11d59e-bb1e-414c-9883-8328ffb6b05a

<img width="295" alt="post-author" src="https://github.com/WordPress/gutenberg/assets/62138457/0c785917-b30a-4b32-b69e-25d56eea8704">

### Environment info

WordPress Version: 6.5.4
Gutenberg Version: 18.5.0
Theme: Twenty Twenty-Four

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

t-hamano:
Hi @rohitmathur-7,

Thanks for the report.

I think the underlying issue is that we can't get user data for a super admin via the REST API if the super admin doesn't exist as a user in the subsite.

So I proposed a change in the core: https://core.trac.wordpress.org/ticket/61444

rohitmathur-7:
Hi @t-hamano ,
I also thought of this approach only first,
But is there a particular reason why we don't get user data for super admins via the REST API in the first place?

t-hamano:
> But is there a particular reason why we don't get user data for super admins via the REST API in the first place?

I don't know if this is intentional or a bug, please see [the core ticket](https://core.trac.wordpress.org/ticket/61444) for more details.

t-hamano:
Sorry for the late reply.

#64165 now shows `(No author)` as a fallback when the post author does not exist. This also applies if there is no super admin for that subsite.

Ideally, the author dropdown should include super admin even if there are no super admin for that subsite. However, as this comment says, there is an issue with the REST API where super admin cannot be retrieved.

Therefore, I think some kind of workaround on the core side will be needed to fundamentally solve this issue.

PR #62575, opened 2024-06-14
Fix author visibility issue in the editor.

rohitmathur-7:
## What?
Fixed the Author Visibility issue by displaying the alphabetically first author's name when the post/page is created by a super admin that does not have their user registered in the sub-site.
Fixes the issue: https://github.com/WordPress/gutenberg/issues/62574

## Why?
There is a UI issue if we are not able to see any author's name. It fixes that.

## How?
When a super admin adds a new post/page and its user is not registered on the sub-site, then instead of showing nothing for the author I have added the name of the alphabetically first author's name. 

## Testing Instructions
- Login to your WordPress site using a Super admin user.
- Go on a sub-site where the Super admin user you are using does not have its user registered.
- Open a new post or page.
- Check the author in the right panel of the editor.

## Screenshots or screencast 
author-fix
Screenshot 2024-06-14 at 4 16 26 PM

### DIFF
--- a/packages/editor/src/components/post-author/hook.js
+++ b/packages/editor/src/components/post-author/hook.js
@@ -59,5 +59,7 @@ export function useAuthorsQuery( search ) {
 		return fetchedAuthors;
 	}, [ authors, postAuthor ] );
 
-	return { authorId, authorOptions, postAuthor };
+	const firstAuthor = authors && authors.length > 0 ? authors[ 0 ] : '';
+
+	return { firstAuthor, authorId, authorOptions, postAuthor };
 }
--- a/packages/editor/src/components/post-author/panel.js
+++ b/packages/editor/src/components/post-author/panel.js
@@ -15,8 +15,8 @@ import PostPanelRow from '../post-panel-row';
 import { useAuthorsQuery } from './hook';
 
 function PostAuthorToggle( { isOpen, onClick } ) {
-	const { postAuthor } = useAuthorsQuery();
-	const authorName = postAuthor?.name || '';
+	const { firstAuthor, postAuthor } = useAuthorsQuery();
+	const authorName = postAuthor?.name || firstAuthor.name;
 	return (
 		<Button
 			size="compact"


t-hamano (commented):
Thanks for the PR!

> then instead of showing nothing for the author I have added the name of the alphabetically first author's name.

I think this might be confusing because it shows a user who isn't the actual user who created the post (super admin).

I'd like to suggest an alternative approach. See [the comment](https://github.com/WordPress/gutenberg/issues/62574#issuecomment-2168010414) here for more details.

rohitmathur-7:
> Thanks for the PR!
> 
> > then instead of showing nothing for the author I have added the name of the alphabetically first author's name.
> 
> I think this might be confusing because it shows a user who isn't the actual user who created the post (super admin).
> 
> I'd like to suggest an alternative approach. See [the comment](https://github.com/WordPress/gutenberg/issues/62574#issuecomment-2168010414) here for more details.

Sorry I just saw the comment on the PR now, and I added the name of the alphabetically first author's name because in the previous versions if a super admin created a post then it showed alphabetically first author's name as the author, so to maintain consistency I did the same.

t-hamano:
> because in the previous versions if a super admin created a post then it showed alphabetically first author's name as the author

Oh, I didn't know that. Do you know in which version of WordPress this issue first occurred?

rohitmathur-7:
> > because in the previous versions if a super admin created a post then it showed alphabetically first author's name as the author
> 
> Oh, I didn't know that. Do you know in which version of WordPress this issue first occurred?

No I don't have idea in which WordPress version this first occurred.

t-hamano:
Sorry for the late reply.

To fundamentally solve this problem, I think some kind of work needs to be done on the core side. See [this comment](https://github.com/WordPress/gutenberg/issues/62574#issuecomment-2285317506) for details.

Mamaduka:
Closing as not applicable. Displaying the first author instead of the actual post author is misleading.

As noted in discussion this requires resolution on the REST API side.

PR #62576, opened 2024-06-14
[Mobile] Support prefix transforms in mobile

geriux:
**Related PRs:**
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/6932

## What?
This PR adds prefix support in the native editor.

## Why?
To allow users to quickly create lists, headings, or quote blocks with different prefixes.

## How?
By implementing the Web editor's functionality into the native editor.

Updates the `inputRule` function with parts of the [web's code](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/rich-text/event-listeners/input-rules.js#L48-L92).

Adds a new check within the `onChangeFromAztec` callback within the RichText component to check for input rules.

Modifies `getBlockTransforms` to allow `prefix` types in mobile.

## Testing Instructions

> [!NOTE]  
> Please use the following builds to test:
> - [iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/23357)
> - [Android](https://github.com/wordpress-mobile/WordPress-Android/pull/20982)

- Open the editor
- Using a Paragraph or Heading block
- Try the following prefixes: `#`, `##`, `###`,  `-`, `>`, `1.`
- **Expect** the block to be transformed into a `Heading` (Different levels), `Quote`, `List` (Numbered, standard)
- **NOTE:** Block transforms hide the keyboard when they're created and it's shown again once the block is rendered, this needs to be investigated in the future to find how to improve that.

N/A

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/4885740/3220c3c5-b874-4108-9e03-8aa4a0c42fb8


twstokes (approved):
Awesome! Testing worked great in iOS and Android @geriux. I didn't try with a keyboard on iPad, but presume that would feel even more natural. 🚀

packages/block-editor/src/components/rich-text/event-listeners/input-rules.js
```diff
@@ -14,7 +14,7 @@ import {
 	START_OF_SELECTED_AREA,
 } from '../../../utils/selection';
 
-function findSelection( blocks ) {
+export function findSelection( blocks ) {
```
geriux, line 17: Hey @ellatrix 👋 I was wondering if there would be any concerns about exporting this so the native editor can leverage its functionality?

PR #62579, opened 2024-06-14
Revert test data for `WithSlug` variation

oandregal:
## What

Removes test data, so it doesn't trigger [this error](https://github.com/WordPress/gutenberg/actions/runs/9513702454/job/26224543289) in CI tests.

See [slack conversation](https://wordpress.slack.com/archives/C02QB2JS7/p1718361269834599) in the core-editor channel.

## Why

To unblock further PRs from landing. We'll continue investigating this and will add it back.

## How

Undoes the changes to the test and removes the corresponding data. We don't know the source of the issue yet.


aaronrobertshaw (approved):
LGTM!

✅ Reverted tests and partial match https://github.com/WordPress/gutenberg/pull/62550
✅ Tests pass locally again after checking out this PR

Thanks for putting this PR up to unblock others 🙇

PR #61607, opened 2024-05-13
chore: convert escape-html package to typescript

jpstevens:
## What?

This PR converts the `escape-html` package to TypeScript.

## Why?

Ensures package is fully type checked.

## How?

- converted files to `.ts`
- explicitly type functions and incorrectly-implied variables, based on `@type`, `@param` and `@return` comments

## Testing Instructions

- [x] `npm run test:unit -- packages/escape-html` tests pass
- [ ] `npm run build:package-types` returns a zero exit code

PR #62586, opened 2024-06-14
convert escape-html package to TS

up1512001:
## What?
This PR converts the escape-html package to TypeScript.

## Why?

## How?
converted files to .ts
explicitly type functions and incorrectly-implied variables, based on @type, @param and @return comments

## Testing Instructions
- `npm run test:unit -- packages/escape-html` tests pass

### reference PR
Closes https://github.com/WordPress/gutenberg/pull/61607

### DIFF
--- a/packages/escape-html/CHANGELOG.md
+++ b/packages/escape-html/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


sirreal (approved):
Thanks!

PR #62587, opened 2024-06-14
Revert "Revert test data for `WithSlug` variation (#62579)"

oandregal:
Reverts https://github.com/WordPress/gutenberg/pull/62579
Follow-up to https://github.com/WordPress/gutenberg/pull/62550

This PR brings back one test we had to put in quarantine to unblock other PRs from lading.

### DIFF
--- a/phpunit/block-supports/block-style-variations-test.php
+++ b/phpunit/block-supports/block-style-variations-test.php
@@ -94,6 +94,22 @@ class WP_Block_Supports_Block_Style_Variations_Test extends WP_UnitTestCase {
 			),
 		);
 
+		/*
+		 * This style is to be deliberately overwritten by the theme.json partial
+		 * See `phpunit/data/themedir1/block-theme/styles/block-style-variation-with-slug.json`.
+		 */
+		register_block_style(
+			'core/group',
+			array(
+				'name'       => 'WithSlug',
+				'style_data' => array(
+					'color' => array(
+						'background' => 'whitesmoke',
+						'text'       => 'black',
+					),
+				),
+			)
+		);
 		register_block_style(
 			'core/group',
 			array(
@@ -106,6 +122,12 @@ class WP_Block_Supports_Block_Style_Variations_Test extends WP_UnitTestCase {
 		$group_styles = $theme_json['styles']['blocks']['core/group'] ?? array();
 		$expected     = array(
 			'variations' => array(
+				'WithSlug'                => array(
+					'color' => array(
+						'background' => 'aliceblue',
+						'text'       => 'midnightblue',
+					),
+				),
 				'my-variation'            => $variation_styles_data,
 
 				/*
@@ -129,6 +151,7 @@ class WP_Block_Supports_Block_Style_Variations_Test extends WP_UnitTestCase {
 		);
 
 		unregister_block_style( 'core/group', 'my-variation' );
+		unregister_block_style( 'core/group', 'WithSlug' );
 
 		$this->assertSameSetsWithIndex( $group_styles, $expected );
 	}
--- a/phpunit/block-supports/block-style-variations-test.php
+++ b/phpunit/block-supports/block-style-variations-test.php
@@ -10,6 +10,41 @@
 
 class WP_Block_Supports_Block_Style_Variations_Test extends WP_UnitTestCase {
 	/**
+	 * Administrator ID.
+	 *
+	 * @var int
+	 */
+	protected static $administrator_id;
+
+	/**
+	 * WP_Theme_JSON_Resolver_Gutenberg::$blocks_cache property.
+	 *
+	 * @var ReflectionProperty
+	 */
+	private static $property_blocks_cache;
+
+	/**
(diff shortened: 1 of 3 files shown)


aaronrobertshaw:
I've had another session attempting to ascertain why this test causes the seemingly unrelated test to fail but haven't had any success yet. I'll need to revisit this later this week as there are a few other fixes that need to land in time for WP6.6 beta 3.

tellthemachines:
I'm not entirely sure what's happening here, but all these tests seem to be using the same `block-theme` so perhaps the problem can be avoided by creating a new theme specifically to test the block style variations on.

aaronrobertshaw:
> I'm not entirely sure what's happening here, but all these tests seem to be using the same block-theme so perhaps the problem can be avoided by creating a new theme specifically to test the block style variations on.

That's definitely possible. I'd like to understand why it's causing the failure in a seemingly unrelated test to make sure I'm not missing some deeper issue.

It's also worth noting that the `block-theme` already has a couple of block style variations in it. There's something about this use case. It's also odd that the test passes fine in core.

andrewserong (commented):
This might be a stretch, but just thinking out loud while testing locally:

The resolver tests include some more detailed cache resets after they're run to ensure that resolver caching is cleared for subsequent tests: https://github.com/WordPress/gutenberg/blob/23b285162fd913718090b211948634dbcc2396ce/phpunit/class-wp-theme-json-resolver-test.php#L67

However, the test in this PR is for `WP_Block_Supports_Block_Style_Variations_Test`, but the test in question uses the resolver class. Do we need to copy over any of the resolver's cleanup logic to ensure the resolver's cache doesn't have any stale values in its `static` properties, after the block style variations tests are concluded?

I'm mostly wondering this because if I just test the block editor settings class locally, all tests pass for me:

```
npm run test:unit:php:base -- --filter Gutenberg_REST_Block_Editor_Settings_Controller_Test
```

However I get the test failure as in CI when I run all the tests:

```
npm run test:unit:php:base
```

So it feels like there are stale values happening _somewhere_ here 🤔

oandregal:
> Do we need to copy over any of the resolver's cleanup logic to ensure the resolver's cache doesn't have any stale values in its static properties, after the block style variations tests are concluded?

I gave this a shot and it works — locally. Pushed to see what CI thinks of it.

Quickly looking at what those setup/teardown functions do, it makes sense they are related: clearing the resolver caches make them to be fresh for the next test. For example, block metadata, would no longer hold any `WithSlug` variation previously registered. Not super familiar with them, so I'd appreciate if anyone gave them a second look and tweak as necessary.

andrewserong (approved):
Nice, glad that approach worked! Resetting the cache in this way sounds like the right way to go to me, and it matches what was in the resolver tests. If we run into any issues further down the track, I suppose we could look at also attempting to reset the `'theme'` and `'user'` properties, but for now, since this gets the tests passing again, I think this looks good to go.

LGTM 🚀

ramonjd:
I tested with and without b879472 and the caching trick works. Good one @andrewserong !

I'll go ahead and merge this.

> I suppose we could look at also attempting to reset the 'theme' and 'user' properties, but for now, since this gets the tests passing again

Was `WP_Theme_JSON_Resolver_Gutenberg::clean_cached_data()` not an option? It resets core, user, theme and more.

https://github.com/WordPress/gutenberg/blob/b879472d9f8b2820a9dce060ff1a78fbce5290e1/lib/class-wp-theme-json-resolver-gutenberg.php#L676-L676

ramonjd:
> Was WP_Theme_JSON_Resolver_Gutenberg::clean_cached_data() not an option? It resets core, user, theme and more.

Seems to work locally (running in `tear_down()` between tests) - I can create a new PR to test it in CI

andrewserong:
> Was WP_Theme_JSON_Resolver_Gutenberg::clean_cached_data() not an option? It resets core, user, theme and more.

Gosh, that looks like an obvious method to try now!

Feel free to merge this as-is and we can try that as a follow-up. (Or we can push that change and try it now, whichever you prefer).

ramonjd:
Oh, I just saw that `tear_down()` has always called `_gutenberg_clean_theme_json_caches()`, which calls `WP_Theme_JSON_Resolver_Gutenberg::clean_cached_data()` 🤔 so maybe I'll try in another PR, perhaps in teardown class or something seeing as this one works for now.

Cheers!

ramonjd:
> WP_Theme_JSON_Resolver_Gutenberg::clean_cached_data() 🤔 so maybe I'll try in another PR, perhaps in teardown class or something seeing as this one works for now.

Well, this is fun. That works fine. 

But now I've removed the caching changes completely in https://github.com/WordPress/gutenberg/pull/62637 and the tests pass on CI and locally. 🤷🏻

andrewserong:
> But now I've removed the caching changes completely in https://github.com/WordPress/gutenberg/pull/62637 and the tests pass on CI and locally. 🤷🏻

🤯 that's very confusing 🤔

PR #23174, opened 2020-06-15
Editor: Extend PostTrashCheck with canUser() check for delete action

ocean90:
## Description
Fixes #23144.

This is a version which doesn't require a new `action-delete` action on the resource.

## How has this been tested?
* Create a user with only `read` and `edit_posts` capabilities
* Create a new post and notice that there's no longer a trash link.

For testing you can use WP-CLI to create a user "author" with the role "author_only_edit" which can only create/edit draft posts.

```
npx wp-env run cli wp role create author_only_edit AuthorOnlyEdit
npx wp-env run cli wp cap add author_only_edit read edit_posts
npx wp-env run cli wp user create author author@example.com
npx wp-env run cli wp user set-role author@example.com author_only_edit
```

## Types of changes
Bug fix

## Checklist:
- [ ] My code is tested.
- [x] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR.

Issue #23144, opened 2020-06-14
Don't show "Move to Trash" link if current user cannot delete posts


ocean90:
**Describe the bug**
Noticed this while working on #23119. The [current component](https://github.com/WordPress/gutenberg/tree/4857ad58c1241b3d63d21a6880c989b85746c3dc/packages/editor/src/components/post-trash) has no checks for current user's capabilities. But there are three that need to be checked: `delete_posts`, `delete_others_posts`, and `delete_published_posts`.

**To reproduce**
* Create a user with only `read` and `edit_posts` capabilities
* Create a new post and click the trash link.
* You'll get the "Sorry, you are not allowed to delete this post." error notice.

**Expected behavior**
If the user has no access to delete posts the link should not be displayed.

**Screenshots**
![image](https://user-images.githubusercontent.com/617637/84599718-72949180-ae74-11ea-9284-4d9a45147ce6.png)

**Additional context**
This probably needs a patch for core to extend the checks in [`WP_REST_Posts_Controller::get_available_actions()`](https://github.com/WordPress/wordpress-develop/blob/9c0fd945582622506312aafbd8079a5b933218a5/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1984-L1986) too.

ocean90:
Ticket for adding an `action-delete` action: https://core.trac.wordpress.org/ticket/50388

ocean90:
On a second thought, we might be able to use the existing `select( 'core' ).canUser()` for the check.

PR #62589, opened 2024-06-14
Fix: the trash post action doesn't take into account user capabilities.

jorgefilipecosta:
Right now we show the trash post action all the time even if the user does not have the correct capabilities.
We essentially regressed on https://github.com/WordPress/gutenberg/issues/23144.
This PR adds equivalent checks to the ones added at https://github.com/WordPress/gutenberg/pull/23174.

It is a bit hacky because we are calling a selector inside isEligible and to make sure things rerender we make the memorization depend on the result of getCachedResolvers.

It is also not ideal in the sense we make a REST API request per post to make sure the user has the permissions. In a list with 50 posts, it is not a very good solution.

I considered not using canUser, and instead retrieving the capabilities of the post and the user: delete_others_posts, delete_post, delete_posts, delete_private_posts, delete_published_posts,delete_published_posts. Implement the verification on the client e.g.: check if the post is published or private, and verify if the user capabilities are there. These would only depend on postType and user so we would not require a request per post. However, it is incompatible with some ways developers change capabilities (e.g.: the sample below would not work with this approach).

So I opted for the more reliable approach that we were already using.

This should be part of 6.6 as showing the UI to trash posts when the user has no capability for that is a regression.

cc: @youknowriad, @ntsekouras as people inside the actions work.

<b>Reviewers note:</b>
Hide whitespace: https://github.com/WordPress/gutenberg/pull/62589/files?w=1

The only changes are the inclusion of useRegistry and useSelect and their use inside isEligible, excluding these changes the action was kept exactly as before.

## Testing Instructions
I pasted the following code snipped on load.php:
```
function remove_delete_post_capability( $caps, $cap, $user_id, $args ) {
    if ( 'delete_post' == $cap ) {
            $caps[] = 'do_not_allow';
        }
    }

    return $caps;
}

add_filter( 'map_meta_cap', 'remove_delete_post_capability', 10, 4 );
```

I verified the option to trash pages was not available on Dataviews pages list and site editor.
I opened a post on the post editor and verified that the action to trash the post was unavailable.

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -309,12 +309,12 @@ const trashPostAction = {
 
 function useTrashPostAction( postType ) {
 	const registry = useRegistry();
-	const { resource, canUserResolvers } = useSelect(
+	const { resource, cachedCanUserResolvers } = useSelect(
 		( select ) => {
 			const { getPostType, getCachedResolvers } = select( coreStore );
 			return {
 				resource: getPostType( postType )?.rest_base || '',
-				canUserResolvers: getCachedResolvers().canUser,
+				cachedCanUserResolvers: getCachedResolvers().canUser,
 			};
 		},
 		[ postType ]
@@ -331,8 +331,10 @@ function useTrashPostAction( postType ) {
 				);
 			},
 		} ),
+		// We are making this use memo depend on cachedCanUserResolvers as a way to make the component using this hook re-render
+		// when user capabilities are resolved. This makes sure the isEligible function is re-evaluated.
 		// eslint-disable-next-line react-hooks/exhaustive-deps
-		[ registry, resource, canUserResolvers ]
+		[ registry, resource, cachedCanUserResolvers ]
 	);
 }
 
@@ -1143,9 +1145,7 @@ export function usePostActions( { postType, onActionPerformed, context } ) {
 		isPattern,
 		postTypeObject?.viewable,
 		duplicatePostAction,
-		permanentlyDeletePostAction,
 		trashPostActionForPostType,
-		restorePostAction,
 		onActionPerformed,
 		isLoaded,
 		supportsRevisions,
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -147,6 +147,166 @@ const deletePostAction = {
 	},
 };
 
+const trashPostAction = {
+	id: 'move-to-trash',
+	label: __( 'Move to Trash' ),
+	isPrimary: true,
+	icon: trash,
+	isEligible( item ) {
+		return ! [ 'auto-draft', 'trash' ].includes( item.status );
+	},
+	supportsBulk: true,
+	hideModalHeader: true,
+	RenderModal: ( { items, closeModal, onActionPerformed } ) => {
+		const [ isBusy, setIsBusy ] = useState( false );
+		const { createSuccessNotice, createErrorNotice } =
+			useDispatch( noticesStore );
+		const { deleteEntityRecord } = useDispatch( coreStore );
+		return (
+			<VStack spacing="5">
+				<Text>
+					{ items.length === 1
+						? sprintf(
+								// translators: %s: The item's title.
+								__(
+									'Are you sure you want to move to trash "%s"?'
+								),
+								getItemTitle( items[ 0 ] )
+						  )
+						: sprintf(
+								// translators: %d: The number of items (2 or more).
+								_n(
+									'Are you sure you want to move to trash %d item?',
+									'Are you sure you want to move to trash %d items?',
+									items.length
+								),
+								items.length
+						  ) }
+				</Text>
+				<HStack justify="right">
+					<Button
+						variant="tertiary"
+						onClick={ closeModal }
+						disabled={ isBusy }
+						__experimentalIsFocusable
+					>
+						{ __( 'Cancel' ) }
+					</Button>
+					<Button
+						variant="primary"
+						onClick={ async () => {
+							setIsBusy( true );
+							const promiseResult = await Promise.allSettled(
+								items.map( ( item ) =>
+									deleteEntityRecord(
+										'postType',
+										item.type,
+										item.id,
+										{},
+										{ throwOnError: true }
+									)
+								)
+							);
+							// If all the promises were fulfilled with success.
+							if (
+								promiseResult.every(
+									( { status } ) => status === 'fulfilled'
+								)
+							) {
+								let successMessage;
+								if ( promiseResult.length === 1 ) {
+									successMessage = sprintf(
+										/* translators: The item's title. */
+										__( '"%s" moved to trash.' ),
+										getItemTitle( items[ 0 ] )
+									);
+								} else if ( items[ 0 ].type === 'page' ) {
+									successMessage = sprintf(
+										/* translators: The number of items. */
+										__( '%s items moved to trash.' ),
+										items.length
+									);
+								} else {
+									successMessage = sprintf(
+										/* translators: The number of posts. */
+										__( '%s items move to trash.' ),
+										items.length
+									);
+								}
+								createSuccessNotice( successMessage, {
+									type: 'snackbar',
+									id: 'move-to-trash-action',
+								} );
+							} else {
+								// If there was at least one failure.
+								let errorMessage;
+								// If we were trying to delete a single item.
+								if ( promiseResult.length === 1 ) {
+									if ( promiseResult[ 0 ].reason?.message ) {
+										errorMessage =
+											promiseResult[ 0 ].reason.message;
+									} else {
+										errorMessage = __(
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -147,165 +147,207 @@ const deletePostAction = {
...
-								promiseResult.every(
-									( { status } ) => status === 'fulfilled'
-								)
-							) {
-								let successMessage;
-								if ( promiseResult.length === 1 ) {
-									successMessage = sprintf(
-										/* translators: The item's title. */
-										__( '"%s" moved to trash.' ),
+function useTrashPostAction( postType ) {
```
youknowriad, line 310: I don't like where this is going. I understand why this was done but I like to think actions should be global objects that don't require a hook to define them. This is going in the opposite direction of all my extensibility PRs.

jorgefilipecosta: Hi @youknowriad, this seems important to address on WP 6.6, we will be showing the trash button on the post editor on posts a user has no permission to trash, which is a considerable regression. As of right now there is no reliable way to fix the issue without using a hook, as soon as our APIs allow it and we can fix the issue without using a hook I'm happy to iterate and remove the hook here.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -1020,6 +1062,7 @@ export function usePostActions( { postType, onActionPerformed, context } ) {
 	);
 
 	const duplicatePostAction = useDuplicatePostAction( postType );
+	const trashPostAction = useTrashPostAction( postType );
```
ellatrix: Why don't we check dynamically add the action here instead of trying to check for the capabilities in `isEligible`?

packages/editor/src/components/post-actions/actions.js
youknowriad: Indeed 6.6 is close, so I'm totally fine with this fix for 6.6 but we should definitely look at better APIs for actions after that.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Hi @ellatrix, because a user may be able to trash a post and not another one. The permissions depend on the specific post. For example, a user maybe have permissions to delete only drafs, or permission to delete drafs authored by that user or permissions to delete private posts but not public ones, etc... We need to check for the specific post in isEligible.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -1113,6 +1143,9 @@ export function usePostActions( { postType, onActionPerformed, context } ) {
 		isPattern,
 		postTypeObject?.viewable,
 		duplicatePostAction,
+		permanentlyDeletePostAction,
+		trashPostActionForPostType,
+		restorePostAction,
```
ellatrix: @jorgefilipecosta Any reason for these extra unneeded dependencies?

packages/editor/src/components/post-actions/actions.js
```diff
@@ -307,6 +307,35 @@ const trashPostAction = {
...
+				return (
+					trashPostAction.isEligible( item ) &&
+					registry
+						.select( coreStore )
+						.canUser( 'delete', resource, item.id )
+				);
+			},
+		} ),
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+		[ registry, resource, canUserResolvers ]
```
ellatrix: Why do we need `canUserResolvers` here?

ellatrix:
I minimised the diff, so it's easier to review and less likely to have cherry-pick conflicts.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -1048,7 +1078,7 @@ export function usePostActions( { postType, onActionPerformed, context } ) {
 			isTemplateOrTemplatePart ? resetTemplateAction : restorePostAction,
 			isTemplateOrTemplatePart || isPattern
 				? deletePostAction
```
ellatrix, line 1082: Don't we need the same for `deletePostAction`?

packages/editor/src/components/post-actions/actions.js
```diff
@@ -161,186 +321,13 @@ function useTrashPostAction( postType ) {
 	);
 	return useMemo(
 		() => ( {
-			id: 'move-to-trash',
-			label: __( 'Move to Trash' ),
-			isPrimary: true,
-			icon: trash,
+			...trashPostAction,
 			isEligible( item ) {
```
youknowriad, line 325: The problem is that the `isEligible` call here won't be re-called when the value of `canUser( 'delete', resource, item.id )` changes no?

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: It was probably an issue during a rebase I think we can remove them.

packages/editor/src/components/post-actions/actions.js
ellatrix: Yeah, maybe we need something like `useEligible` somewhere. Or maybe this object should define which capabilities to check.

youknowriad: Related discussion https://github.com/WordPress/gutenberg/pull/62505#discussion_r1646480478

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: When we do registry.select( coreStore ).canUser( 'delete', resource, item.id ), on the first call it will return false, then the selector starts returning true but nothing is causing a rerender, so the trash action would not be available until a rerender happens e.g: the user closes and reopen the sidebar. We make the memo depend on canUserResolvers as a way to cause a rerender when the resolver finishes. Of course, this is a temporary hacky fix, our actions API needs to offer a long-term solution as we are discussing in https://github.com/WordPress/gutenberg/pull/62647#discussion_r1644336613.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Yes, we need the same for some other actions, my plan was to define the fix here and after we agree on a fix I will replicate it to the other cases where it makes sense.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: It will because we make the memo depend on canUserResolvers (details in https://github.com/WordPress/gutenberg/pull/62589#discussion_r1647524094) so when the resolver finishes a rerender happens and isEligible will be recalled.

ellatrix: 🤨 seems fine for now but not ideal

youknowriad: It's very hard to follow. I'm not sure I understand what `canUserResolvers` is?

packages/editor/src/components/post-actions/actions.js
ellatrix: Could you add an inline comment?

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: canUserResolvers is the result getCachedResolvers().canUser. So basically when there is a change in the caching of canUser resolvers we will rerender making isEligible recalled. I will add some inline comments, explaining what we discussed.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Hi @ellatrix I added an inline comment as suggested 👍

packages/editor/src/components/post-actions/actions.js
ellatrix: Is permanently delete a separate capability? I'm still seeing this:

image

ellatrix: I tested with the testing instructions, and it seems to be the same capability: I can also not delete it permanently in trunk.

ellatrix (approved):
Not ideal, but we're not exposing any public APIs.

packages/editor/src/components/post-actions/actions.js
ntsekouras: Also why keeping the action as object just to be used inside the `hook` action? Anyway since it's for 6.6, but it becomes clear that an action needs a way to have access to the store in more places than the callback and we should work towards that direction.

ellatrix: @ntsekouras I did that to minimise the diff, it was huge and unclear what changed + less likely to have cherry pick conflicts, and if it does, easier to resolve manually (I don't know what has changed in these files recently)

ellatrix: And it's useful for reusability https://github.com/WordPress/gutenberg/pull/62754#discussion_r1649305615

ellatrix:
Handled in #62823.

PR #62592, opened 2024-06-14
Fix: Check Capability to create pages on DataViews add new page button

jorgefilipecosta:
We display the "Add New Page" option for all users, causing errors for those without page creation capabilities.
This PR fixes the issue we now check if the user is able to create pages when rendering the button.

## Testing Instructions
I pasted this test code on load.php to remove the capability to create pages:
```
// Allow only administrators to create pages by default.
add_filter(
	'register_page_post_type_args',
	function ( $args ) {

		return $args;
	},
);
```
I went to `/wp-admin/site-editor.php?postType=page` and verified that "Add New Page" button was not available.

### DIFF
--- a/packages/edit-site/src/components/page-pages/index.js
+++ b/packages/edit-site/src/components/page-pages/index.js
@@ -277,15 +277,19 @@ export default function PagePages() {
 		[ totalItems, totalPages ]
 	);
 
-	const { frontPageId, postsPageId, addNewLabel } = useSelect( ( select ) => {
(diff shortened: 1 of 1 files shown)


carolinan (approved):
The PR works as described.
I confirmed that a user who has access to the Site Editor, but do not have access to creating pages, are not able to create new pages from Appearance > Editor > Pages. The "Add new" button is not present.

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 3b10fb71ee

PR #62593, opened 2024-06-14
Updates the reference links in @wordpress/compose documentation.

vipul0425:
## What?
Updates the reference link in the @wordpress/compose documentation.

## Why?

The old reference link for hooks goes to the `https://legacy.reactjs.org/docs/hooks-intro.html`, so updated this to the new documentation link and for Higher Order Components goes to `https://facebook.github.io/react/docs/higher-order-components.html` which is giving not found error. So updated this with the legacy link as new documentation doesn't have specific page for it.

### DIFF
--- a/docs/contributors/code/coding-guidelines.md
+++ b/docs/contributors/code/coding-guidelines.md
@@ -758,7 +758,7 @@ When documenting an example, use the markdown <code>\`\`\`</code> code block to 
 
 ### Documenting React components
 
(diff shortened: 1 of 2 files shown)


shail-mehta:
@vipul0425 

There is another instance of the URL https://reactjs.org/docs/hooks-intro.html in `docs/contributors/code/coding-guidelines.md at line 761`. Could you please update the URL in that file as well?

juanmaguitar (approved):

juanmaguitar:
LGTM! ✅

PR #62595, opened 2024-06-14
Focus Editor Region from Template Footer Click

jeryj:
## What?

Clicking the Template button in the footer breadcrumbs causes a focus loss to the body of the page. This PR sets focus to the editor region instead, making it similar to how the other buttons in footer breadcrumbs work.

## Why?

a11y. Focus management.

## How?

When clicking "Template" in the footer breadcrumbs, set focus on the editor canvas region through a new utility, getEditorRegion. 

## Testing Instructions

- Select a block
- Click the root "Template" button in the footer
- Focus should _not_ be visible on the canvas as it uses focus-visible (only shows focus if activated via moving to it via keyboard), but focus should be there. 
- Tab to check focus is at the top of the editor.

- Select a block
- Tab many times to the root "Template" button in the footer
- Focus should be visible on the canvas

### DIFF
--- a/packages/block-editor/src/utils/get-editor-region.js
+++ b/packages/block-editor/src/utils/get-editor-region.js
@@ -1,0 +1,32 @@
+/**
+ * Gets the editor region for a given editor canvas element or
+ * returns the passed element if no region is found
+ *
+ * @param { Object } editor
+ * @return { Object } The editor region or given editor element
+ */
+export default function getEditorRegion( editor ) {
+	if ( ! editor ) {
+		return null;
+	}
+
+	// If there are multiple editors, we need to find the iframe that contains our contentRef to make sure
+	// we're focusing the region that contains this editor.
+	const editorCanvas =
+		document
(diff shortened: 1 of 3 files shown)


alexstine (approved):
@jeryj Thanks, this is testing well.

ajlende (approved):

packages/block-editor/src/utils/get-editor-region.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Gets the editor region for a given editor canvas element or
+ * returns the passed element if no region is found
+ *
+ * @param { Object } editor The editor canvas element.
+ * @return { Object } The editor region or given editor element
+ */
+export default function getEditorRegion( editor ) {
+	if ( ! editor ) {
+		return null;
```
ellatrix, line 10: Note that this could be returning `null`, but you're calling focus() unconditionally.

jeryj: Good catch! Fixed here: https://github.com/WordPress/gutenberg/pull/62980

PR #62598, opened 2024-06-15
Correct documentation for WP_Block_Parser_Frame class

dilipbheda:
See https://core.trac.wordpress.org/ticket/60423 Core ticket for more details.

### DIFF
--- a/packages/block-serialization-default-parser/class-wp-block-parser-frame.php
+++ b/packages/block-serialization-default-parser/class-wp-block-parser-frame.php
@@ -65,8 +65,9 @@ class WP_Block_Parser_Frame {
 	 * @param WP_Block_Parser_Block $block              Full or partial block.
 	 * @param int                   $token_start        Byte offset into document for start of parse token.
 	 * @param int                   $token_length       Byte length of entire parse token string.
-	 * @param int|null              $prev_offset        Byte offset into document for after parse token ends. Default null.
-	 * @param int|null              $leading_html_start Byte offset into document where leading HTML before token starts. Default null.
+	 * @param int|null              $prev_offset        Optional. Byte offset into document for after parse token ends. Default null.
+	 * @param int|null              $leading_html_start Optional. Byte offset into document where leading HTML before token starts.
+	 *                                                  Default null.
 	 */
 	public function __construct( $block, $token_start, $token_length, $prev_offset = null, $leading_html_start = null ) {
 		$this->block              = $block;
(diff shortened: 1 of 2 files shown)


packages/block-serialization-default-parser/class-wp-block-parser-frame.php
```diff
@@ -65,8 +65,8 @@ class WP_Block_Parser_Frame {
 	 * @param WP_Block_Parser_Block $block              Full or partial block.
 	 * @param int                   $token_start        Byte offset into document for start of parse token.
 	 * @param int                   $token_length       Byte length of entire parse token string.
-	 * @param int                   $prev_offset        Byte offset into document for after parse token ends.
-	 * @param int                   $leading_html_start Byte offset into document where leading HTML before token starts.
+	 * @param int|null              $prev_offset        Byte offset into document for after parse token ends. Default null.
+	 * @param int|null              $leading_html_start Byte offset into document where leading HTML before token starts. Default null.
```
dmsnell: If we're here making this change it might be helpful to also include an `Optional. ` prefix, as this pattern is predominant in other code.

Could you add that to the start of each of the parameter descriptions?

dmsnell (commented):
Thanks for the patch! I left a note because I think it would be good to add the additional `Optional.` note in the parameter descriptions, but otherwise this can be quick and easy. 😄

packages/block-serialization-default-parser/class-wp-block-parser-frame.php
dilipbheda: @dmsnell Thank you for the review, I've added the `Optional.` prefix descriptions in the latest commit.

dmsnell (approved):
Looks like you'll need to rebase this or run the tests again, but I think they aren't related to your change.

Once they pass this looks ready to go. Thanks again!

dilipbheda:
> Looks like you'll need to rebase this or run the tests again, but I think they aren't related to your change.
> 
> Once they pass this looks ready to go. Thanks again!

@dmsnell Now all tests are passed, I'm not sure about the `Verify Core Backport Changlog / Check CHANGELOG diff` checklist. Should I open a new PR for WordPress/develop a repo with the same changes?

dmsnell:
@dilipbheda if you click on the "Details" link for that failing test it gives you instructions for creating a PR or a change to ensure that this propagates upward into Core. You might search for a similar ticket with "Backport" in the name. See what you can do; if we need help we can reach out to someone more familiar with that process.

dilipbheda:
@dmsnell  Now passed `Verify Core Backport Changlog / Check CHANGELOG diff` checklist as well.

PR #62605, opened 2024-06-16
Query Loop block: Clarify explanation around query loop variation example

artemiomorales:
## What?

This PR clarifies that specifying the layout of a query loop block variation is needed in order for the variation to function properly.

## Why?

Currently, the documentation is not clear that the `Customize your variation layout` step is needed — and if this is not done, the variation defaults to allowing the user to pick a layout in the UI, which overwrites the variation's attributes.

## How?

It adds numbers to the headings under `Extending the block with variations` and also adds a clarifying sentence to ensure users follow the next step.

## Testing Instructions

N/A

N/A

### DIFF
--- a/docs/how-to-guides/block-tutorial/extending-the-query-loop-block.md
+++ b/docs/how-to-guides/block-tutorial/extending-the-query-loop-block.md
@@ -14,7 +14,7 @@ With the block variations API you can provide the default settings that make the
 
 Let's go on a journey, for example, of setting up a variation for a plugin which registers a `book` [custom post type](https://developer.wordpress.org/plugins/post-types/).
 
-### Offer sensible defaults
+### 1. Offer sensible defaults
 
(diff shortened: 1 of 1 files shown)


artemiomorales:
I'm not sure that this PR will completely clarify the explanation, though I believe it will help. Any thoughts appreciated! 🙏

juanmaguitar:
Thanks for the PR @artemiomorales! 
The changes look good and clarify the prerequisites for a customized query loop to work properly a bit more.

To provide even more clarity over these "requisites" I would also add just before...

```
Let's go on a journey, for example, of setting up a variation for a plugin ...
```

Something along the following lines:

```
In order to have a Query Loop variation properly working we'll need to 
- register the block variation for the `core/query` block with some default values
- define a layout for the block variation
- use the `namespace` attribute in the `isActive` block variation property
```

To provide an overview of what's explained next in more detail to ensure the block variation works as expected.

juanmaguitar (approved):

PR #62609, opened 2024-06-17
Github PHP changes workflow: remove it

ramonjd:
Removing the PHP changes workflow

## Why?

The PHP detections action checks for PHP changes and adds a comment to ask whether a backport is necessary. 

Example:

Screenshot 2024-06-17 at 11 28 08 AM

I think it's just noise now that the check [backport changelog workflow](https://github.com/WordPress/gutenberg/blob/ce54632acbc9da0e5615dd392d37d5270bc5b077/.github/workflows/check-backport-changelog.yml) is in place.

Also to reduce clutter.


noisysocks:
Looks good to me unless anyone in @WordPress/gutenberg-core has any objections 👍

tellthemachines (approved):
This makes sense, thanks for the cleanup!

ellatrix:
No objections, if you think the failing test is enough. I was wondering if the backport log test should leave an actual Github comment instead, though there should be a clear error message if you open the test results.

Issue #62133, opened 2024-05-30
Pattern overrides: Button and Image blocks sharing the same name have incompatible attributes


talldan:
### Description

When pattern overrides are enabled for a block, the block name becomes the key that attribute data is stored under in instances of the pattern block. For example, a button block called "CTA" would store it's `url` attribute value in a pattern block's content like so:
```js
content: { "CTA": { "url": "https://wordpress.org" } }
```

Another feature of pattern overrides is that two blocks can share the same name, so two paragraphs, or a paragraph and a heading can have their text content synced, with the data being stored in the same object.

A problem occurs when a user gives a button block and an image block the same name. The image block has the following attributes:
- `url` - used for the `<img>` tag's `src` attribute
- `href` - used for the optional `<a>` tag's `href` attribute

While the button block has:
- `url` - used for the optional `<a>` tag's `href` attribute

For image and button `url` means two different things. If an image and a button share the same override name, it can result in a broken image src when the user sets the button block's link. It's unfortunate that the button block's attribute isn't called `href`.

Another factor is that the image block's `href` isn't currently a supported attribute for pattern overrides / block bindings.

### Step-by-step reproduction instructions

1. Add an image and a button to a post
2. Select both and create a pattern
3. Enable overrides for both the image and the button, and use the same block name.
4. Save the pattern and go back to the post
5. Set the button block's link to something that's not an image url (e.g. `https://wordpress.org`)

Expected: The link for the image is set to the same thing given the two blocks are named the same
Actual: The image block has a broken image

### Screenshots, screen recording, code snippet

![Screenshot 2024-05-30 at 5 05 34 pm](https://github.com/WordPress/gutenberg/assets/677833/46b1a10d-30cc-40c7-8530-4d8bc08aba8a)

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

talldan:
@youknowriad Not sure if you recall, but we discussed this issue in the past (though I can't remember which PR we discussed it on now).

I've put together a PR that renames the button block attributes to see exactly what it would entail - https://github.com/WordPress/gutenberg/pull/62612.

It ends up being quite a big change. 😬 

I'm not sure if there's anything else we can do to mitigate the issue before 6.6..

An alternative could be to prevent users from using the same name for button and image blocks. I can look into that. What do you think?

youknowriad:
[...]

PR #62612, opened 2024-06-17
Rename button block attributes from url and text to href and content

talldan:
## What?
This is an attempt to solve https://github.com/WordPress/gutenberg/issues/62133 by renaming the button block's attributes to be consistent with other blocks.

Mostly an exploration to see what it would entail (it turns out a lot of code changes, but it didn't take too long).

## Why?
With block bindings and pattern overrides, the block's attributes are forming more of a schema. As described in the issue, with pattern overrides, attribute values can be shared across blocks with different types, and the values are accessed using the attribute name. When the attribute names are inconsistent for the content types they represent (like the `url` attribute name representing the `src` attribute in the image block but also the `href` attribute for a button block), it causes unexpected issues (the image `src` ending up as the button block's `href` or vice versa).

## How?
Rename the attributes to make these attribute schemas consistent across block types.

## Testing Instructions
TBD


talldan:
One issue is that the block bindings code in the gutenberg plugin doesn't execute when the core class is present:
https://github.com/WordPress/gutenberg/blob/c6e7ddb770e04919586247da34fcda86c007de73/lib/compat/wordpress-6.5/blocks.php#L50

So that means the e2e tests for block bindings that are failing in this PR won't pass unless this change is backported.

Feels like those e2e tests should be moved to wordpress core instead if they're testing core functionality.

talldan:
Closing based on feedback in https://github.com/WordPress/gutenberg/issues/62133.

PR #62617, opened 2024-06-17
Performance: run block variation hook only for matches

ellatrix:
## What?

See #57908.

Adds an `isMatch` function so we can avoid running the block variation hook when a blocks has class names (which is common), but there are no variation class names. Additionally, avoid the backtracking regex.

## Why?

Performance.

## How?

With an `isMatch` function.

## Testing Instructions

See things to test in #57908, everything should work correctly.

### DIFF
--- a/packages/block-editor/src/hooks/block-style-variation.js
+++ b/packages/block-editor/src/hooks/block-style-variation.js
@@ -17,6 +17,23 @@ import { useStyleOverride } from './utils';
 import { store as blockEditorStore } from '../store';
 import { globalStylesDataKey } from '../store/private-keys';
 
+const VARIATION_PREFIX = 'is-style-';
+
+function getVariationMatches( className ) {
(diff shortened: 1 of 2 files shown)


ellatrix:
I have an additional performance concern here: https://github.com/WordPress/gutenberg/pull/57908#discussion_r1642602448

aaronrobertshaw (approved):
Nice work! Thanks for putting this together @ellatrix 🙇 

✅ Approach looks good
✅ Variations are applied correctly in post/site editors
✅ Variations are saved and applied on the frontend still too


https://github.com/WordPress/gutenberg/assets/60436221/6a5a7bc4-a22d-42f6-bfda-65e7885942d6

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 2bb7aa6a77

PR #62620, opened 2024-06-17
Fix: Check create capability on duplicate post action.

jorgefilipecosta:
The action for duplicating a post does not currently check if the user has the necessary capabilities to create a post. We talked about potentially adding a specific capability for duplicating posts in the future, but for now, we should at least use the capabilities that are already available.
## Testing Instructions
I disabled the ability to create pages with:
```
// Allow only administrators to create pages by default.
add_filter(
	'register_page_post_type_args',
	function ( $args ) {

		return $args;
	},
);
```
I verified the duplicate button was rendered for posts but not for pages.

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -722,132 +722,162 @@ const renamePostAction = {
 	},
 };
 
-const duplicatePostAction = {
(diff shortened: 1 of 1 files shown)


ellatrix:
Does this need backporting?

PR #36839, opened 2021-11-24
[Docs]: Add docs for `schema` in block API

dmsnell:
## Description

While the `schema` property is mentioned in the block API documentation
for `raw` transforms it's not explained.

In this patch we're adding a short sub-section to expand on the purpose
and implementation of the `schema` in a way that can hopefully give
developers a stronger starting point for when they need to implement
new raw transformations.

## How has this been tested?

This is a documentation-only change and does not touch the code.
Please review the new explanations and provide feedback on the
wording and example and whether or not you find the structure of
inlining the subsection into the existing block API document to be
a helpful move, or whether you think it belongs elsewhere.

## Types of changes
This change adds an explanation to existing block API documentation.

## Checklist:
- [ ] My code is tested.
- [ ] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #62623, opened 2024-06-17
Docs: Update an html usage with pure markdown.

jorgefilipecosta:
There is no reason to use h4 instead of pure markdown. As an advantage of the change the link to the Schemas and Content Models section will work properly when reading the document on a text editor like VS code.

### DIFF
--- a/docs/reference-guides/block-api/block-transforms.md
+++ b/docs/reference-guides/block-api/block-transforms.md
@@ -221,7 +221,7 @@ transforms: {
 }
 ```
 
-<h4 id="schemas-and-content-models">Schemas and Content Models</h4>
+#### Schemas and Content Models
 
 When pasting content it's possible to define a [content model](https://html.spec.whatwg.org/multipage/dom.html#content-models) that will be used to validate and process pasted content. It's often the case that HTML pasted into the editor will contain a mixture of elements that _should_ transfer as well as elements that _shouldn't_. For example, consider pasting `<span class="time">12:04 pm</span>` into the editor. We want to copy `12:04 pm` and omit the `<span>` and its `class` attribute because those won't carry the same meaning or structure as they originally did from where they were copied.
 


docs/reference-guides/block-api/block-transforms.md
```diff
@@ -221,7 +221,7 @@ transforms: {
 }
 ```
 
-<h4 id="schemas-and-content-models">Schemas and Content Models</h4>
+#### Schemas and Content Models
```
ellatrix, line 224: Are the anchors auto-generated? Looks like this was added by @dmsnell in #36839.

jorgefilipecosta: I think they were not auto-generated.

dmsnell: I may have done this to create a fixed id for the header, as not all Markdown flavors support doing so.

Note that this change likely breaks the link above in line 197

dmsnell:
> the link to the Schemas and Content Models section will work properly when reading the document on a text editor like VS code.

seems like a bug in whatever Markdown preview you are using. what is broken right now in VS Code?

docs/reference-guides/block-api/block-transforms.md
jorgefilipecosta: The link on line 197 still seems to work well. If we follow this link  https://github.com/WordPress/gutenberg/blob/docs/replace-and-html-usage-with-pure-markdown/docs/reference-guides/block-api/block-transforms.md#schemas-and-content-models we still go to the right section.

jorgefilipecosta:
> > the link to the Schemas and Content Models section will work properly when reading the document on a text editor like VS code.
> 
> seems like a bug in whatever Markdown preview you are using. what is broken right now in VS Code?

Opening the file docs/reference-guides/block-api/block-transforms.md on the normal code/text editor view going to line 197, clicking above #schemas-and-content-models while pressing cmd does not follow the link without this change, with this change it follows the link.

docs/reference-guides/block-api/block-transforms.md
dmsnell: very good. we don't have any way to prevent someone else breaking that when they change the text of the heading? if anything, that is why I manually added the HTML as a safeguard - because right now, the fact that the link works is accidental in Markdown at best. When someone undoubtedly comes in and changes the text of the heading, we want that link to remain.

dmsnell:
> Opening the file docs/reference-guides/block-api/block-transforms.md on the normal code/text editor view going to line 197, clicking above #schemas-and-content-models while pressing cmd does not follow the link without this change, with this change it follows the link.

Thanks for clarifying. That definitely sounds like a bug in the VS Code plugin. I don't have strong opinions here, but I do see this change as making the association between the header and the link pointing to it as less reliable, more prone to change.

Can you attempt to see if you can add a fixed id through Markdown alone? I'd rather not make our documentation riskier just because VS Code's Markdown handler doesn't happen to parse the HTML ids _at this moment in time_. We could file a bug in VS Code and point out that their system is broken. We're surely not the only ones doing this.

jorgefilipecosta:
Hi @dmsnell, thank you for looking into this. Supposedly in markdown one can also add an ID with {#schemas-and-content-models}, but this did not work in VS code either so I'm going to close this PR as it is related to a VS code bug.

PR #57120, opened 2023-12-15
Added additional explanations to attributes and supports sections

juanmaguitar:
This PR adds some additional context for the [Data Storage in the Block with attributes](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-json/#data-storage-in-the-block-with-attributes) and [Enable UI settings panels for the block with supports](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-json/#enable-ui-settings-panels-for-the-block-with-supports) sections and removes related pages under [How to guides > blocks](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/) as I consider the content in them is properly covered with the pages under ["Fundamentals of Block Development" ](https://developer.wordpress.org/block-editor/getting-started/fundamentals/)

The pages removed are:
- [Introducing Attributes and Editable Fields](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/introducing-attributes-and-editable-fields/) - `attributes` are already introduced in [Data Storage in the Block with attributes](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-json/#data-storage-in-the-block-with-attributes) and an example of an editable text component is already covered in the [coming-soon tutorial ](https://github.com/WordPress/gutenberg/pull/56931)
- [Block Supports](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/block-supports-in-static-blocks/) - this property is properly introduced in [Enable UI settings panels for the block with supports](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-json/#enable-ui-settings-panels-for-the-block-with-supports)
- [Block Supports in dynamic blocks](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/block-supports-in-dynamic-blocks/) - the content of this page is properly covered in  [Enable UI settings panels for the block with supports](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-json/#enable-ui-settings-panels-for-the-block-with-supports)  and [The block wrapper](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-wrapper/)
- [Generate Blocks with WP-CLI](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/generate-blocks-with-wp-cli/) - this page is no longer relevant as stated in its own content

@ndiego for the following pages a redirection to [block.json ](https://developer.wordpress.org/block-editor/getting-started/fundamentals/block-json/) would be needed:
[...]

PR #62624, opened 2024-06-17
Fix: Remove inexistent link from the documentation.

jorgefilipecosta:
The section /docs/how-to-guides/block-tutorial/introducing-attributes-and-editable-fields.md was removed, and this tutorial is not available anymore. This PR returns a link to the section that is invalid and does not exist.

### DIFF
--- a/docs/reference-guides/block-api/block-edit-save.md
+++ b/docs/reference-guides/block-api/block-edit-save.md
@@ -188,7 +188,7 @@ When saving your block, you want to save the attributes in the same format speci
 
 ## Examples
 
-Here are a couple examples of using attributes, edit, and save all together. For a full working example, see the [Introducing Attributes and Editable Fields](/docs/how-to-guides/block-tutorial/introducing-attributes-and-editable-fields.md) section of the Block Tutorial.
+Here are a couple examples of using attributes, edit, and save all together.
 
 ### Saving Attributes to Child Elements
 


docs/reference-guides/block-api/block-edit-save.md
```diff
@@ -188,7 +188,7 @@ When saving your block, you want to save the attributes in the same format speci
 
 ## Examples
 
-Here are a couple examples of using attributes, edit, and save all together. For a full working example, see the [Introducing Attributes and Editable Fields](/docs/how-to-guides/block-tutorial/introducing-attributes-and-editable-fields.md) section of the Block Tutorial.
```
ellatrix, line 191: What happened to it?

jorgefilipecosta: That section was removed here https://github.com/WordPress/gutenberg/pull/57120 but we missed the removal of this link. The idea according to that PR was that Building your first block tutorial would be a good replacement for that section. But in this context of a block API reference linking to the build your first block tutorial is not very useful.

ellatrix (approved):

docs/reference-guides/block-api/block-edit-save.md
ellatrix: Commented on that PR, it a bit weird to remove these pages and break links on the web.

PR #62626, opened 2024-06-17
env: Ignore `\$schema` key in environment config parsing

### LINKED ISSUE #62619: wp-env: $schema property is not allowed
#36276 added a JSON schema for wp-env, but if the `$schema` property appears in wp-env configuration, it throws an error:

```
Invalid /Users/jonsurrell/a8c/wordpress-develop/trunk/.wp-env.json: "$schema" is not a configuration option.
```

It should allow that property, or possibly even better use the schema to parse configuration.

_Originally posted by @sirreal in https://github.com/WordPress/gutenberg/issues/36276#issuecomment-2173085546_

t-hamano:
Fixes #62619

## What?

This PR fixes an exception error when a schema key is defined in the env config file.

## Why?

This key is only used for validation and should be excluded from configuration options.

## Testing Instructions

- Add the `$schema` key to the `.wp-env.json` file.
- Run `npm run wp-env start`.
- No errors should occur.

### DIFF
--- a/packages/env/CHANGELOG.md
+++ b/packages/env/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancement
+
+-   Ignore `$schema` key in environment config parsing ([#62626](https://github.com/WordPress/gutenberg/pull/62626)).
+
 ## 10.1.0 (2024-06-15)
 
 ## 10.0.0 (2024-05-31)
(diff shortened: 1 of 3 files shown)


sirreal (approved):
Thanks! Please add a changelog entry before landing ("Enhancements").

ellatrix:
Manually resolved this in 4bffbc7661bb1ab71d043b0c5bd613c120503bee

PR #62628, opened 2024-06-17
Fix: Unable to remove custom colors on safari. (Try 2)

### LINKED ISSUE #62261: Global Styles: Can't remove a custom color from palette in Safari

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/62261
Supersedes: https://github.com/WordPress/gutenberg/pull/62554

The cause of the issue seems to be that on safari while clicking the remove button onStopEditing is also triggered, to overcome this issue we add reference and set it to true when clicking is started.

cc: @noisysocks, @richtabor, @aatanasovdev  in case you can confirm if the issue is also fixed for you.

## Testing Instructions
Use Safari.
Go to Appearance → Editor → Colors → Palette.

### DIFF
--- a/packages/compose/src/hooks/use-focus-outside/index.ts
+++ b/packages/compose/src/hooks/use-focus-outside/index.ts
@@ -32,6 +32,9 @@ type FocusNormalizedButton =
 function isFocusNormalizedButton(
 	eventTarget: EventTarget
 ): eventTarget is FocusNormalizedButton {
+	if ( eventTarget instanceof window.SVGElement ) {
+		return !! eventTarget.closest( 'button' );
+	}
 	if ( ! ( eventTarget instanceof window.HTMLElement ) ) {
 		return false;
 	}


ellatrix:
Does this need the backport label? I saw the original PR had the label.

aatanasovdev:
It seems the logic needs to be added to the hook itself, not the components where it is used. Otherwise, this will lead to duplicated code (when the hook is used in other components, and they experience the same issue).

After reviewing the issue further, it looks like the root of the problem is that when the Remove button is clicked, the target is an SVG element (the button icon), and this isn't considered by the [isFocusNormalizedButton](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts#L32C10-L32C33) function.

I'd like to propose modifying the [useFocusOutside](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts) hook and proceeding with any of these solutions:
 1. Check if the clicked element is a child element of the element that uses the hook. If true, the callback function isn't executed.
 2. Check if the parent element of the clicked element matches these [cases](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts#L39-L40). The same as the first solution, the callback function isn't executed.

jorgefilipecosta:
> It seems the logic needs to be added to the hook itself, not the components where it is used. Otherwise, this will lead to duplicated code (when the hook is used in other components, and they experience the same issue).
> 
> After reviewing the issue further, it looks like the root of the problem is that when the Remove button is clicked, the target is an SVG element (the button icon), and this isn't considered by the [isFocusNormalizedButton](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts?rgh-link-date=2024-06-18T08%3A43%3A55Z#L32C10-L32C33) function.
> 
> I'd like to propose modifying the [useFocusOutside](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts?rgh-link-date=2024-06-18T08%3A43%3A55Z) hook and proceeding with any of these solutions:
> 
> 1. Check if the clicked element is a child element of the element that uses the hook. If true, the callback function isn't executed.
> 2. Check if the parent element of the clicked element matches these [cases](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.ts?rgh-link-date=2024-06-18T08%3A43%3A55Z#L39-L40). The same as the first solution, the callback function isn't executed.

Thank you for the insights @aatanasovdev I tried to do something on isFocusNormalizedButton before but my breakpoints were not hitting and it was because I missed that SVG is not an instance of window.HTMLElement. Also, the event itself may be on the path node. I added a check for SVG inside the button on isFocusNormalizedButton it seems to fix the issue.

jorgefilipecosta:
> Does this need the backport label? I saw the original PR had the label.

Good catch the backport label was added.

packages/compose/src/hooks/use-focus-outside/index.ts
```diff
@@ -32,6 +32,9 @@ type FocusNormalizedButton =
 function isFocusNormalizedButton(
 	eventTarget: EventTarget
 ): eventTarget is FocusNormalizedButton {
+	if ( eventTarget instanceof window.SVGElement ) {
```
ellatrix, line 35: So the SVG is capturing the click instead of the button?

ellatrix: Why is the SVG the target? If there's no other fix, can we generalise this to normalise for any element within a button? Also we should avoid using globals and use `ownerDocument.defaultView` intstead.

ellatrix:
I found an alternative solution in case it helps: #62698

jorgefilipecosta:
Closed in favor of https://github.com/WordPress/gutenberg/pull/62698.

PR #62632, opened 2024-06-17
Editor: fix scrollbar doubling

### LINKED ISSUE #62319: Double scrollbar issue when block is selected
### Description

With the first WordPress 6.6 beta 1, I found a strange scrollbar issue when a block is selected and you scroll on the page. When the block disappears from the viewport, a second scrollbar is displayed.

1. Select a block
2. scroll in the viewport with the block selected
3. when the block disappear from the viewport area a second scrollbar will appear

https://github.com/WordPress/gutenberg/assets/492145/78abd516-411e-4dc0-86dc-c4edd3a2f41b

### Environment info

- WordPress 6.6 beta 1

Yes

Yes

stokesman:
## What?
An improvement of overflow handling in the block canvas.

## Why?
Fixes #62319 (double scrollbars when selected block is out of view).

## How?
Makes the block canvas contained by the height of its parent so the parent won’t overflow. 

### Fixed scrollbar doubling
1. In an editor with content that overflows
2. Scroll down and select a block that will be out of view unless scrolled
3. Scroll up to conceal the selected block’s toolbar
4. Note that no additional scrollbars appear

The general instructions should be applicable.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/9000376/a4756171-c3c0-4819-bc7a-21b56fca5da6

### DIFF
--- a/packages/block-editor/src/components/block-canvas/index.js
+++ b/packages/block-editor/src/components/block-canvas/index.js
@@ -58,7 +58,12 @@ export function ExperimentalBlockCanvas( {
 	return (
 		<BlockTools
 			__unstableContentRef={ localRef }
-			style={ { height, display: 'flex' } }
+			style={ {
+				height,
+				display: 'flex',
+				overflow: 'hidden',
+				position: 'relative',
+			} }
 		>
 			<Iframe
 				{ ...iframeProps }
--- a/packages/block-editor/src/components/iframe/content.scss
+++ b/packages/block-editor/src/components/iframe/content.scss
@@ -5,7 +5,11 @@
 .block-editor-iframe__container {
 	width: 100%;
 	height: 100%;
-	overflow-x: hidden;
+	overflow: auto;
+
+	&.is-zoomed-out {
+		overflow-x: hidden;
+	}
 }
 
 .block-editor-iframe__scale-container {
@@ -13,7 +17,7 @@
 	height: 100%;
 }
 
-.block-editor-iframe__scale-container.is-zoomed-out {
+.is-zoomed-out.block-editor-iframe__container .block-editor-iframe__scale-container {
 	$container-width: var(--wp-block-editor-iframe-zoom-out-container-width, 100vw);
 	$prev-container-width: var(--wp-block-editor-iframe-zoom-out-prev-container-width, 100vw);
 	width: $prev-container-width;
--- a/packages/block-editor/src/components/iframe/content.scss
+++ b/packages/block-editor/src/components/iframe/content.scss
@@ -5,11 +5,7 @@
 .block-editor-iframe__container {
 	width: 100%;
 	height: 100%;
-	overflow: auto;
-
-	&.is-zoomed-out {
-		overflow-x: hidden;
-	}
+	overflow-x: hidden;
 }
 
 .block-editor-iframe__scale-container {
@@ -17,7 +13,7 @@
 	height: 100%;
 }
 
-.is-zoomed-out.block-editor-iframe__container .block-editor-iframe__scale-container {
+.block-editor-iframe__scale-container.is-zoomed-out {
 	$container-width: var(--wp-block-editor-iframe-zoom-out-container-width, 100vw);
 	$prev-container-width: var(--wp-block-editor-iframe-zoom-out-prev-container-width, 100vw);
 	width: $prev-container-width;
--- a/packages/block-editor/src/components/iframe/index.js
+++ b/packages/block-editor/src/components/iframe/index.js
@@ -447,16 +447,13 @@ function Iframe( {
 	);
 
 	return (
-		<div
-			className={ clsx(
-				'block-editor-iframe__container',
-				isZoomedOut && 'is-zoomed-out'
-			) }
-			ref={ windowResizeRef }
-		>
+		<div className="block-editor-iframe__container" ref={ windowResizeRef }>
 			{ containerResizeListener }
 			<div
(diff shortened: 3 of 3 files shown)


ellatrix:
There's a lot of changes going on here which makes me nervous to include it in 6.6. What is the minimal fix without improving anything else?

stokesman:
I dropped the changes that improved anything else. There are alternatives approaches that could be more minimal.

Here’s one that I didn’t prefer because there’s more coupling (some of it cross-package) when ideally none of these components have to concern themselves with each other. Also, it uses `:has()` which I wasn’t sure meets the policy for browser compatibility.
```diff
diff --git a/packages/edit-post/src/components/layout/style.scss b/packages/edit-post/src/components/layout/style.scss
index 2c5a035cad..5fdaceaa00 100644
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -1,8 +1,3 @@
-.editor-visual-editor {
-	flex: 1 0 auto;
-	height: auto;
-}
-
 .edit-post-layout__metaboxes {
 	flex-shrink: 0;
 	clear: both;
diff --git a/packages/editor/src/components/editor-interface/style.scss b/packages/editor/src/components/editor-interface/style.scss
index e7e0a93af2..c2d6777d21 100644
--- a/packages/editor/src/components/editor-interface/style.scss
+++ b/packages/editor/src/components/editor-interface/style.scss
@@ -1,3 +1,7 @@
 .editor-editor-interface .entities-saved-states__panel-header {
 	height: $header-height + $border-width;
 }
+
+.editor-editor-interface:has(.editor-visual-editor.is-iframed) .interface-interface-skeleton__content {
+	overflow: hidden;
+}


```

Here’s one I like better. It’s encapsulated and two birds with one stone because it also fixes #62639 but would probably create the same caution for 6.6 that some folks had about #62681. `contain` is a bit more recent but I think it works for browser compatibility policy.
```diff
diff --git a/packages/editor/src/components/visual-editor/style.scss b/packages/editor/src/components/visual-editor/style.scss
index 597769ff7f..42f4fe2d7d 100644
--- a/packages/editor/src/components/visual-editor/style.scss
+++ b/packages/editor/src/components/visual-editor/style.scss
@@ -3,6 +3,8 @@
 	height: 100%;
 	display: block;
 	background-color: $gray-300;
+	// Make this element the context for positioning and stacking (z-index) of children.
+	contain: layout;
 
 	// Centralize the editor horizontally (flex-direction is column).
 	align-items: center;

```

ellatrix:
Would love some opinion from CSS experts (maybe @jasmussen or if you know someone else 😃)

stokesman:
Okay, I realized there was one more thing this was still doing besides fixing the double scrollbars and pushed a commit to revert that. Now it’s about as minimal as any of the alternative solutions I mentioned. What this was doing besides was allowing horizontal overflow when not zoomed out. That was so that if a device preview overflows horizontally it’s not clipped and can be scrolled. It seems like an oversight when zoom out was added but it’s probably a minor concern.

I removed the zoom out label since now the changes don’t touch anything related.

stokesman:
The one thing I originally included in this PR that I think might qualify to include is fixing the triple scrollbars that appear in the post editor in device previews. I mentioned it in the issue https://github.com/WordPress/gutenberg/issues/62319#issuecomment-2166539672. It is not present 6.5 so also a regression we’d ideally fix for 6.6. It can be done simply:
```diff
diff --git a/packages/edit-post/src/components/layout/style.scss b/packages/edit-post/src/components/layout/style.scss
index 2c5a035cad..5fdaceaa00 100644
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -1,8 +1,3 @@
-.editor-visual-editor {
-	flex: 1 0 auto;
-	height: auto;
-}
-
 .edit-post-layout__metaboxes {
 	flex-shrink: 0;
 	clear: both;

```
I don’t really see it as a separate issue but I'll put it in separate PR if that’s what others think is best.

jasmussen:
Thanks for the ping. Based on the most recent changes there's not a lot of CSS, so it might be safer now. In a quick surface test this seems to work as intended, though I had a hard time replicating the double scrollbar in trunk myself. 

It's a hard one to test deeply, while these are indeed sensitive rules, it's also hard to foresee every side effect. It seems fine  given its current small footprint.

ellatrix:
I also can't reproduce the original issue. 🤔

ellatrix:
Which browser are you using?

stokesman:
> I had a hard time replicating the double scrollbar

Maybe it depends on having OS preferences so that scrollbars always show. I would think it doesn’t but maybe it’s hard to tell that there are nested areas of overflow without.

I'm using Chrome/macOS.

t-hamano:
In my environment (Windows OS, Chrome), I can consistently reproduce this issue by following these steps. If you're on a Mac, you may need to change your OS settings to always show the scroll bars.

- Activate TT4 theme.
- Go to the site editor.
- Open the Blog Home template.
- Disable the Top Toolbar.
- Select the third section ("An array of resources" section).
- Scroll up the content.
- Repeat the same steps for the mobile view.

I have confirmed that this issue is resolved in this PR.

https://github.com/WordPress/gutenberg/assets/54422211/7f8ea643-6836-49db-b2a9-766f20c5c172

stokesman:
#62894 resolved this nicely so I'll close.

PR #62633, opened 2024-06-17
Fix: Check ability to create patterns on the add new pattern modal.

jorgefilipecosta:
A user may have permission to edit patterns and should see the patterns section but may not have permission to create new patterns. 
If the user has no permission to create patterns we currently show the option to create and import anyway but when using these features there is a crash.

This PR adds a check for the capability to create patterns, and also a check for the capability to create template parts. The check for the capability to create template parts would not be strictly necessary because currently the capability to create template parts just depends on edit_theme_options, and if edit_theme_options is not available the site editor does not even load. But let us add the check anyway for template parts in case things change or some capability management plugin does something we are not expecting.

## Testing Instructions
I pasted the following test code that removes the ability to create patterns (while still allowing to edit them).
```
add_filter(
	'register_wp_block_post_type_args',
	function ( $args ) {

		return $args;
	},
);
```
I went to wp-admin/site-editor.php?postType=wp_block verified patterns loaded well and I could edit them, but the ability to create new patterns was not available.

### DIFF
--- a/packages/edit-site/src/components/add-new-pattern/index.js
+++ b/packages/edit-site/src/components/add-new-pattern/index.js
@@ -36,21 +36,29 @@ export default function AddNewPattern() {
 	const [ showPatternModal, setShowPatternModal ] = useState( false );
 	const [ showTemplatePartModal, setShowTemplatePartModal ] =
 		useState( false );
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
 	const { createPatternFromFile } = unlock( useDispatch( patternsStore ) );
 	const { createSuccessNotice, createErrorNotice } =
 		useDispatch( noticesStore );
 	const patternUploadInputRef = useRef();
-	const { isBlockBasedTheme, addNewPatternLabel, addNewTemplatePartLabel } =
-		useSelect( ( select ) => {
-			const { getCurrentTheme, getPostType } = select( coreStore );
-			return {
-				isBlockBasedTheme: getCurrentTheme()?.is_block_theme,
-				addNewPatternLabel: getPostType( PATTERN_TYPES.user )?.labels
-					?.add_new_item,
-				addNewTemplatePartLabel: getPostType( TEMPLATE_PART_POST_TYPE )
-					?.labels?.add_new_item,
-			};
-		}, [] );
+	const {
+		isBlockBasedTheme,
+		addNewPatternLabel,
+		addNewTemplatePartLabel,
+		canCreatePattern,
+		canCreateTemplatePart,
+	} = useSelect( ( select ) => {
+		const { getCurrentTheme, getPostType, canUser } = select( coreStore );
+		return {
+			isBlockBasedTheme: getCurrentTheme()?.is_block_theme,
+			addNewPatternLabel: getPostType( PATTERN_TYPES.user )?.labels
+				?.add_new_item,
+			addNewTemplatePartLabel: getPostType( TEMPLATE_PART_POST_TYPE )
+				?.labels?.add_new_item,
+			canCreatePattern: canUser( 'create', 'blocks' ),
+			canCreateTemplatePart: canUser( 'create', 'template-parts' ),
+		};
+	}, [] );
 
 	function handleCreatePattern( { pattern } ) {
 		setShowPatternModal( false );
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/components/add-new-pattern/index.js
```diff
@@ -36,21 +36,29 @@ export default function AddNewPattern() {
...
+		canCreateTemplatePart,
+	} = useSelect( ( select ) => {
+		const { getCurrentTheme, getPostType, canUser } = select( coreStore );
+		return {
+			isBlockBasedTheme: getCurrentTheme()?.is_block_theme,
+			addNewPatternLabel: getPostType( PATTERN_TYPES.user )?.labels
+				?.add_new_item,
+			addNewTemplatePartLabel: getPostType( TEMPLATE_PART_POST_TYPE )
+				?.labels?.add_new_item,
+			canCreatePattern: canUser( 'create', 'blocks' ),
```
carolinan, line 58: It can be confusing that 'blocks' here represents **patterns**. I wonder if this needs a comment.

carolinan:
This PR does hide the "Add new pattern" button, but the Duplicate option is still available from the action menu, is there a separate PR for that?

packages/edit-site/src/components/add-new-pattern/index.js
ellatrix: Is `blocks` referring to the CPT?

packages/edit-site/src/components/add-new-pattern/index.js
```diff
@@ -36,21 +36,29 @@ export default function AddNewPattern() {
 	const [ showPatternModal, setShowPatternModal ] = useState( false );
 	const [ showTemplatePartModal, setShowTemplatePartModal ] =
 		useState( false );
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
 	const { createPatternFromFile } = unlock( useDispatch( patternsStore ) );
```
ellatrix, line 40: Another trick is to use `useRegistry`, call dispatch and unlock in the action callback.

packages/edit-site/src/components/add-new-pattern/index.js
```diff
@@ -114,8 +128,16 @@ export default function AddNewPattern() {
...
-					label={ addNewPatternLabel }
+					text={
+						canCreatePattern
+							? addNewPatternLabel
+							: addNewTemplatePartLabel
+					}
+					label={
+						canCreatePattern
+							? addNewPatternLabel
+							: addNewTemplatePartLabel
```
ellatrix: Why are we switching here now? What if you can do neither?

packages/edit-site/src/components/add-new-pattern/index.js
jorgefilipecosta: Yes, blocks refer to wp_blocks CPT, one needs to be able to create a post there to create a pattern otherwise things fail.

packages/edit-site/src/components/add-new-pattern/index.js
jorgefilipecosta: It felt strange to have a menu saying Add New Pattern, while adding a template part was the option. If the user can not add template parts and can not add patterns, the condition if ( controls.length === 0 ) { will be true and we will not render any menu.

ellatrix: There's also a check for `addNewPatternLabel` which is maybe a bit strange now.

jorgefilipecosta: Hi @ellatrix, I reverted this change and now we just focus on checking permissions before rendering controls.

jorgefilipecosta:
> This PR does hide the "Add new pattern" button, but the Duplicate option is still available from the action menu, is there a separate PR for that?

Hi @carolinan, thank you for the review. I plan on fixing the duplicate pattern action in a separate PR following the same approach we ended up deciding on for the trash post-action that is being discussed at https://github.com/WordPress/gutenberg/pull/62633.

packages/edit-site/src/components/add-new-pattern/index.js
carolinan: This is why I think a comment about it would help.

ellatrix (approved):
This is what I'm seeing 👍 

image

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: d4930245cb

PR #62635, opened 2024-06-17
Update Private API opt-in string for WP 6.6.

peterwilsoncc:
## What?

Update the opt-in string for Private APIs for use in WordPress 6.6

## Why?

The opt-in string is intended to be rotated on a regular basis to prevent themes and plugins from opting in to private and unstable APIs.

## How?

Search and replace of the previous string. 

## Testing Instructions
1. Identify code making use of the Private APis within Gutenberg
2. Ensure the code continues to work with the string rotated.
3. Search for the old string `I know using unstable features means my theme or plugin will inevitably break in the next version of WordPress.` and ensure it only appears in the history section of the private api's readme file.

N/A 

## Screenshots or screencast 
N/A

### DIFF
--- a/docs/contributors/code/coding-guidelines.md
+++ b/docs/contributors/code/coding-guidelines.md
@@ -162,7 +162,7 @@ do so by opting-in to `@wordpress/private-apis`:
 import { __dangerousOptInToUnstableAPIsOnlyForCoreModules } from '@wordpress/private-apis';
 export const { lock, unlock } =
 	__dangerousOptInToUnstableAPIsOnlyForCoreModules(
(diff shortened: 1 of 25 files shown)


talldan (approved):
Smoke tested the PR, and everything's working well.

Also searched for the old string, and only found it in the README where it's expected to be.

Looks like the only thing failing is the (optional) changelog check for the components package. Not sure if it's worth adding something for private APIs, I'd be happy to merge without an entry.

Thanks for handling this @peterwilsoncc.

ellatrix:
I just cherry-picked this PR to the wp/6.6-beta-3 branch to get it included in the next release: 6ac4b9824d

Issue #52032, opened 2023-06-28
Top Toolbar: Refine responsiveness and apply across mobile


jasmussen:
The top toolbar mode option puts the block toolbar in the top center of the main toolbar. 

As the screen grows smaller, we can refine this responsiveness as shown here:

<img width="1472" alt="Responsive scheme" src="https://github.com/WordPress/gutenberg/assets/1204802/6731c689-b9cc-4459-8dfb-25cd382e2063">

* At 960px, the document title contracts, and the rightmost tools are hidden as well, when a block is selected.
* At 480, the document title contracts further, and hides the shortcut hint.
* On mobile and embedded, the document title is entirely hidden.

Related: #51711.

jasmussen:
Closing this one for now, as separate inititatives have caused drift with this one and are being tracked elsewhere.

PR #53913, opened 2023-08-24
Patterns: Disable the preview option button when editing

t-hamano:
Fixes: #53862
Follow-up on: #52427

## What?

This PR ~~removes~~ disables the header device preview button in the pattern edit mode.

patterns

## Why?

The pattern edit page should not need this button, as the canvas is resizable by handles, as are the template parts and navigation.

## How?

I have discovered that there is a `FOCUSABLE_ENTITIES` constant that defines whether the current entity is focusable or not.
I moved `constant.js` to the proper location to reuse this constant. This will prevent regression if editable entities are updated in the future.

## Testing Instructions

- Create an unsynced pattern.
- Enter edit mode.
- Make sure that the preview button is ~~not displayed~~ disabled in the header.

PR #58979, opened 2024-02-13
Refactor: useBlockTools hook

jeryj:
Closes https://github.com/WordPress/gutenberg/issues/57288 alternative to https://github.com/WordPress/gutenberg/pull/57291 and https://github.com/WordPress/gutenberg/pull/58947.

There are times where a block is selected but no tools are available. If we don't have any tools, we don't want to show a toolbar. This combines the logic for showing the block toolbar that was sprinkled throughout the codebase:

- block-toolbar/index.js checked if it should render based on if a toolbar was enabled for the block and if there were block controls in the slots
- block-tools/index.js determined if the popover should render based on cascading rules if the inserter popover or block select mode popover were showing (and other rules)
- use-can-block-toolbar-be-focused made an approximation of if the block toolbar would be showing or not, but fell short in some areas such as when a block toolbar should have been rendered, but there were no block tools available.

## What?

Unify logic for which block tools will be showing so there is a source of truth for what will be showing (block toolbar popover, fixed block toolbar, block inserter, block breadcrumb).

## Why?

Fixes https://github.com/WordPress/gutenberg/issues/57288.

## How?

- Adds a `useHasBlockToolbar` hook that has the logic for if the block-toolbar will render or not.
- Adds a `useBlockTools` hook that privately exports which block tools will be showing (block toolbar popover, fixed block toolbar, block inserter, block breadcrumb).
- Checks if any block tools are available before adding the `is-collapsed` class to the central toolbar area
- Removes selected block tools wrapper from DOM if no block tools available. I'm not sure if this was intentional to leave it in the DOM before, or if it was because there wasn't a reliable way of checking for it.

## Testing Instructions

Toolbar should show/hide in the same way as before.

### Testing Instructions for Keyboard

- Go to Site Editor
- Turn on Top Toolbar
- Edit Navigation Template Part
- Click navigation item
- Press Up Arrow until the block toolbar disappears
- Central command area should be present
- Central command area should not have `is-collapsed` class
- Alt + F10 should move focus to the document tools

PR #56921, opened 2023-12-09
Editor: Unify the preview dropdown between post and site editors

youknowriad:
Related #52632 

## What?

This PR continues the work on the great unification between post and site editors. This PR unifies the DevicePreview or PreviewDropdown component. It also brings some consistency meaning that now:

 - When you're editing a page, either it's in the site or post editors, it offers you a way to preview the page in the frontend.
 - When you're editing a template it shows the "view site" link instead.

## Testing instructions

 - Check the device preview dropdown in both post and site editors.

PR #62636, opened 2024-06-18
Editor: Improve Header layout

stokesman:
## What?
An improvement to center the Document Bar and prevent layout shifts at > medium breakpoints when switching between page and template.

## Why?
- It seems like the Document Bar is intended to be centered yet it’s not (especially noticeable when editing templates).
- The layout shift while switching between a page and its template spoils the animation a bit.

## How?
Uses grid instead of flex because that was the only way I knew to resiliently center the Document Bar. There is a bit more to it than that but I think it may be better to cover that with some review comments.

## More
The layout shift used to not happen but now in template editing the "View Page" button hides and causes it. Maybe the button shouldn’t hide and instead merely disable https://github.com/WordPress/gutenberg/pull/53913#issuecomment-1697346079 but it still seems like a good idea to make the layout more stable (and truly centered).

## Testing Instructions
1. Open a page in the site or post editor
2. Note the Document Bar is centered in the header
3. Edit the template and note that the Document Bar stays centered
4. Resize to various widths and verify that the responsiveness acceptable throughout

### Layout on trunk editing a page
near-centered-document-bar

### Layout on trunk editing a template
offset-document-bar-for-template

### Layout on this branch editing a page
centered-document-bar

Capture of editing a template on the branch omitted because there’s no layout shift.

### Trunk switching between page and template
https://github.com/WordPress/gutenberg/assets/9000376/46c6f410-6bac-475b-a1c3-2eadf997f056

### This branch switching between page and template
https://github.com/WordPress/gutenberg/assets/9000376/84a8b0d1-4828-4a1e-925b-f77697fe6444

### DIFF
--- a/packages/editor/src/components/collapsible-block-toolbar/index.js
+++ b/packages/editor/src/components/collapsible-block-toolbar/index.js
@@ -9,20 +9,12 @@ import clsx from 'clsx';
 import {
 	BlockToolbar,
 	store as blockEditorStore,
-	privateApis as blockEditorPrivateApis,
 } from '@wordpress/block-editor';
 import { useEffect } from '@wordpress/element';
 import { Button, Popover } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { next, previous } from '@wordpress/icons';
 import { useSelect } from '@wordpress/data';
-
-/**
- * Internal dependencies
- */
-import { unlock } from '../../lock-unlock';
-
-const { useHasBlockToolbar } = unlock( blockEditorPrivateApis );
 
 function CollapsableBlockToolbar( { isCollapsed, onToggle } ) {
 	const { blockSelectionStart } = useSelect( ( select ) => {
@@ -31,7 +23,6 @@ function CollapsableBlockToolbar( { isCollapsed, onToggle } ) {
 				select( blockEditorStore ).getBlockSelectionStart(),
 		};
 	}, [] );
-	const hasBlockToolbar = useHasBlockToolbar();
 
 	const hasBlockSelection = !! blockSelectionStart;
 
@@ -41,10 +32,6 @@ function CollapsableBlockToolbar( { isCollapsed, onToggle } ) {
 			onToggle( false );
 		}
 	}, [ blockSelectionStart, onToggle ] );
-
-	if ( ! hasBlockToolbar ) {
-		return null;
-	}
 
 	return (
 		<>
--- a/packages/editor/src/components/collapsible-block-toolbar/index.js
+++ b/packages/editor/src/components/collapsible-block-toolbar/index.js
@@ -16,7 +16,7 @@ import { __ } from '@wordpress/i18n';
 import { next, previous } from '@wordpress/icons';
 import { useSelect } from '@wordpress/data';
 
-function CollapsableBlockToolbar( { isCollapsed, onToggle } ) {
+export default function CollapsibleBlockToolbar( { isCollapsed, onToggle } ) {
 	const { blockSelectionStart } = useSelect( ( select ) => {
 		return {
 			blockSelectionStart:
@@ -60,5 +60,3 @@ function CollapsableBlockToolbar( { isCollapsed, onToggle } ) {
 		</>
 	);
 }
-
-export default CollapsableBlockToolbar;
--- a/packages/editor/src/components/collapsible-block-toolbar/index.js
+++ b/packages/editor/src/components/collapsible-block-toolbar/index.js
@@ -9,12 +9,20 @@ import clsx from 'clsx';
 import {
 	BlockToolbar,
 	store as blockEditorStore,
+	privateApis as blockEditorPrivateApis,
 } from '@wordpress/block-editor';
 import { useEffect } from '@wordpress/element';
 import { Button, Popover } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { next, previous } from '@wordpress/icons';
 import { useSelect } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { useHasBlockToolbar } = unlock( blockEditorPrivateApis );
 
 export default function CollapsibleBlockToolbar( { isCollapsed, onToggle } ) {
 	const { blockSelectionStart } = useSelect( ( select ) => {
@@ -23,6 +31,7 @@ export default function CollapsibleBlockToolbar( { isCollapsed, onToggle } ) {
 				select( blockEditorStore ).getBlockSelectionStart(),
 		};
 	}, [] );
+	const hasBlockToolbar = useHasBlockToolbar();
 
 	const hasBlockSelection = !! blockSelectionStart;
 
@@ -32,6 +41,10 @@ export default function CollapsibleBlockToolbar( { isCollapsed, onToggle } ) {
 			onToggle( false );
 		}
 	}, [ blockSelectionStart, onToggle ] );
+
+	if ( ! hasBlockToolbar ) {
+		return null;
+	}
 
 	return (
 		<>
--- a/packages/editor/src/components/collapsible-block-toolbar/style.scss
+++ b/packages/editor/src/components/collapsible-block-toolbar/style.scss
@@ -74,9 +74,10 @@
 
 	&.is-collapsed {
 		display: none;
+
+		~ .editor-collapsible-block-toolbar__toggle {
+			margin-left: $grid-unit-10;
+		}
 	}
 }
 
-.editor-collapsible-block-toolbar__toggle {
-	margin-left: 2px; // Allow focus ring to be fully visible
-}
--- a/packages/editor/src/components/collapsible-block-toolbar/style.scss
+++ b/packages/editor/src/components/collapsible-block-toolbar/style.scss
@@ -74,10 +74,6 @@
 
 	&.is-collapsed {
 		display: none;
-
-		~ .editor-collapsible-block-toolbar__toggle {
-			margin-left: $grid-unit-10;
-		}
 	}
 }
 
--- a/packages/editor/src/components/document-bar/style.scss
+++ b/packages/editor/src/components/document-bar/style.scss
@@ -10,7 +10,7 @@
 	min-width: 0;
 	background: $gray-100;
 	border-radius: $grid-unit-05;
-	width: min(100%, 416px);
+	width: min(100%, 450px);
 
 	&:hover {
 		background-color: $gray-200;
@@ -24,10 +24,6 @@
 		&:hover {
 			background: $gray-200;
 		}
-	}
-
-	@include break-large() {
-		width: min(100%, 450px);
 	}
 }
 
--- a/packages/editor/src/components/document-tools/style.scss
+++ b/packages/editor/src/components/document-tools/style.scss
@@ -65,12 +65,12 @@
 .editor-document-tools__left {
 	display: inline-flex;
 	align-items: center;
-	padding-left: $grid-unit-20;
 	gap: $grid-unit-10;
 
-	// Some plugins add buttons here despite best practices.
-	// Push them a bit rightwards to fit the top toolbar.
-	margin-right: $grid-unit-10;
+	// Some plugins add buttons here despite best practices — accommodate them with a gap.
+	&:not(:last-child) {
+		margin-inline-end: $grid-unit-10;
+	}
 }
 
 .editor-document-tools .editor-document-tools__left > .editor-document-tools__inserter-toggle.has-icon {
--- a/packages/editor/src/components/header/back-button.js
+++ b/packages/editor/src/components/header/back-button.js
@@ -9,16 +9,16 @@ import {
 // Keeping an old name for backward compatibility.
 const slotName = '__experimentalMainDashboardButton';
 
+export const useHasBackButton = () => {
+	const fills = useSlotFills( slotName );
+	return Boolean( fills && fills.length );
+};
+
 const { Fill, Slot } = createSlotFill( slotName );
 
 const BackButton = Fill;
-const BackButtonSlot = ( { children } ) => {
+const BackButtonSlot = () => {
 	const fills = useSlotFills( slotName );
-	const hasFills = Boolean( fills && fills.length );
-
-	if ( ! hasFills ) {
-		return children;
-	}
 
 	return (
 		<Slot
--- a/packages/editor/src/components/header/index.js
+++ b/packages/editor/src/components/header/index.js
@@ -20,7 +20,7 @@ import {
 /**
  * Internal dependencies
  */
-import BackButton from './back-button';
+import BackButton, { useHasBackButton } from './back-button';
 import CollapsibleBlockToolbar from '../collapsible-block-toolbar';
 import DocumentBar from '../document-bar';
 import DocumentTools from '../document-tools';
@@ -101,21 +101,25 @@ function Header( {
 		) : null;
 
 	const showDocumentBar = ! blockToolbar || isBlockToolsCollapsed;
+	const hasBackButton = useHasBackButton();
 
 	// The edit-post-header classname is only kept for backward compatibilty
 	// as some plugins might be relying on its presence.
 	return (
 		<div
 			className={ clsx( 'editor-header edit-post-header', {
+				'has-back-button': hasBackButton,
 				'has-center': isMobileViewport && showDocumentBar,
 			} ) }
 		>
-			<motion.div
-				variants={ backButtonVariations }
-				transition={ { type: 'tween' } }
-			>
-				<BackButton.Slot />
-			</motion.div>
+			{ hasBackButton && (
+				<motion.div
+					variants={ backButtonVariations }
+					transition={ { type: 'tween' } }
+				>
+					<BackButton.Slot />
+				</motion.div>
+			) }
 			<motion.div
 				variants={ toolbarVariations }
 				className="editor-header__toolbar"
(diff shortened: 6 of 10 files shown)


packages/editor/src/components/collapsible-block-toolbar/index.js
```diff
@@ -9,29 +9,20 @@ import clsx from 'clsx';
...
 
-/**
- * Internal dependencies
- */
-import { unlock } from '../../lock-unlock';
-
-const { useHasBlockToolbar } = unlock( blockEditorPrivateApis );
-
-function CollapsableBlockToolbar( { isCollapsed, onToggle } ) {
+export default function CollapsibleBlockToolbar( { isCollapsed, onToggle } ) {
```
stokesman, line 27: Typo correction, the filename is correct but the component itself was typo’d.

packages/editor/src/components/collapsible-block-toolbar/index.js
```diff
@@ -9,29 +9,20 @@ import clsx from 'clsx';
...
 	const { blockSelectionStart } = useSelect( ( select ) => {
 		return {
 			blockSelectionStart:
 				select( blockEditorStore ).getBlockSelectionStart(),
 		};
 	}, [] );
-	const hasBlockToolbar = useHasBlockToolbar();
```
stokesman: Removing this in favor of using it in the consuming component allows for better conditional rendering. Before the consumer can’t easily know if the there’s nothing rendered. The impetus is to be able to show the Document Bar when the block toolbar would be empty.

packages/editor/src/components/document-tools/style.scss
```diff
@@ -65,12 +65,12 @@
...
-	padding-left: $grid-unit-20;
 	gap: $grid-unit-10;
 
-	// Some plugins add buttons here despite best practices.
-	// Push them a bit rightwards to fit the top toolbar.
-	margin-right: $grid-unit-10;
+	// Some plugins add buttons here despite best practices — accommodate them with a gap.
+	&:not(:last-child) {
+		margin-inline-end: $grid-unit-10;
+	}
```
stokesman, line 73: I thought conditionally adding this "gap" was better than universally doing so. It saves a bit of space when plugins aren’t adding buttons there (or "Top toolbar" is not on as it also serves there). Note this change would leave no space before the collapsible block toolbar’s toggle button so that is instead handled by header styles.

packages/editor/src/components/document-bar/style.scss
```diff
@@ -25,10 +25,6 @@
 			background: $gray-200;
 		}
 	}
-
-	@include break-large() {
-		width: min(100%, 450px);
-	}
```
stokesman, line 31: I'm not sure just why the width needed a distinct maximum at the large breakpoint. I moved the `450px` up to the base style on L13 and with the overall changes the Document Bar smoothly narrows according to available space.

packages/editor/src/components/collapsible-block-toolbar/style.scss
```diff
@@ -74,9 +74,10 @@
 
 	&.is-collapsed {
 		display: none;
+
+		~ .editor-collapsible-block-toolbar__toggle {
+			margin-left: $grid-unit-10;
+		}
```
stokesman: EDIT: moved this to the header styles instead as it’s not actually a concern of this component.

Adds a "gap" equal to that of the document tools toolbar. This was previously down with a margin on that toolbar that I changed it because it didn’t always need to be there. There’s more info in a comment on that change.

packages/editor/src/components/collapsible-block-toolbar/style.scss
```diff
@@ -74,9 +74,10 @@
...
 	}
 }
 
-.editor-collapsible-block-toolbar__toggle {
-	margin-left: 2px; // Allow focus ring to be fully visible
-}
```
stokesman, line 82: The parent element’s style changes now ensure this isn’t necessary.

stokesman (commented):
I’ve added some explanatory review comments.

jasmussen:
Nice one, another very important detail to get right. Appreciate the PR. At a glance, works okay. I know this is something both @richtabor and @jameskoster have tinkered with, so here's a CC.

Before:

before

Note how the title is visible on mobile, but not in a functional way. And note how it is flush with the save button as well, on the tablet breakpoint.

After:

after

Note how both the mentioned issues are addressed, the title is hidden on mobile, and the overall responsiveness is more fluid. So how do you see or change the document title on mobile? Through the document inspector as part of the new unified summary panel:

document inspector with title highlighted and the titles ellipsis menu open to show the rename option

At this point, I would say this more or less fixes #52032. If there are a couple of small bits from that issue you can absorb, feel free to, though!

stokesman:
Here’s something that may need addressed at some point. The site icon in the header currently is rendered into a slot. Conceivably, there could be an app that chooses not to fill the slot yet end up with a space for it anyway because the styles in this branch don’t account for that. It seems a moot point for now because these components are private APIs and all apps are filling the slot. Still, in trunk it happens that the space is not reserved in such cases:

This branch without a fill for the slot:
image

Trunk without a fill for the slot:
image

jameskoster:
This is looking good, though I'm a little anxious about hiding the title on mobile since it removes access to the Command Palette. I've seen comments elsewhere about revealing the undo/redo buttons on mobile too. I wonder if we could hide _more_, and surface actions through the command palette instead.

commands

---

Probably unrelated, but potentially worth fixing here; I see a 'Preview' button on mobile, which is duplicative of the external icon button:

Screenshot 2024-06-20 at 16 04 47

stokesman:
> I'm a little anxious about hiding the title on mobile

I wanted to keep it around at mobile. There’s enough room for it to be functional (providing we resolve that odd "Preview" button). I wanted to hide at some point to help preserve just how narrow the header can currently go but I didn’t see another breakpoint at < mobile that could be used for that though. Perhaps it doesn’t have to hide and can just squish out of existence like it currently does.

> I wonder if we could hide _more_, and surface actions through the command palette instead.

I think it’s a worthwhile exploration though perhaps not so much for this PR.

> Probably unrelated, but potentially worth fixing here; I see a 'Preview' button on mobile, which is duplicative of the external icon button:

Thanks for pointing that out. I’ve been seeing it too and wondered why Joen’s screen recording didn’t have it. Then while testing most recently I noticed it was absent during one session but seems like a refresh later it was back. I haven’t yet looked into it. It does seem unrelated but I'll try and understand what’s happening.

----

> Here’s something that may need addressed at some point. The site icon in the header currently is rendered into a slot

https://github.com/WordPress/gutenberg/pull/62636#issuecomment-2179369654

After making that comment I realized this was busted in the post editor when the editor isn’t fullscreen (also at less than 782px where there is no fullscreen) as it hides the Site icon. It’s effectively the same issue I referred to in the comment although the slot is filled and its button is hidden by CSS. I suppose I could have written some style overrides only for the post editor but I decided it’s probably neater overall to address it in the header itself. So I added 07dd535a57adb2ec82f72a8b2334d3d12db08ae2 and 132045e3c2968eca518f0028b70a130b572a6cec to do so. The post editor won’t fill the slot unless fullscreen and the header is aware of whether the slot is filled or not.

stokesman:
> Probably unrelated, but potentially worth fixing here; I see a 'Preview' button on mobile, which is duplicative of the external icon button

I looked into this some and that button is only present when editing a post type that can be previewed (i.e. pages and not template parts, patterns or others). Then it’s hidden at < small breakpoints. As best I can tell this came about in #56921 [diff of relevant addition](https://github.com/WordPress/gutenberg/pull/56921/files#diff-ef1669926cfbac743ceeabe7570da47639b3bb5e62775ef9107c8074aded21e3R254-R259) and seems like a mistake. Shouldn’t the button be hidden for smaller viewports instead of shown? I've pushed a commit 41ba1770250662eb2d31b36265cefa8d03f30bdc that makes it so.

It’s an aside but the Preview button is not totally duplicative of the View button because the former will show the post as edited and the latter will show the post as saved. I wonder if it’d be an improvement to combine the controls into a menu (for the post types it’s applicable to anyway).

More aside; here are some other details I noticed that seem like they warrant fixes but are not related enough to this PR:
image

- The Preview button’s text baseline isn’t quite aligned to that of the Document Bar or Save button.
- Both the Preview and View buttons have a focus `color` that isn’t shared by any other buttons I tested.


> I'm a little anxious about hiding the title on mobile

I've gone ahead and tried keeping it around for about 100px less than mobile 30d5b4af3fcb8989898bfb31cffc7063ff1d8b71. So it’s now closer to how it is in trunk. Here’s the max-width it’s still shown at when editing a page:
image

And when editing a template:
image

If we wanted, a separate breakpoint could be used to keep the Document Bar visible even narrower for templates but I'm not sure it’s worth it.

jameskoster:
Ah, I guess it's because the Preview action is bundled in the device preview toggles on desktop. Obviously these aren't applicable on smaller screens, hence they get replaced with the 'Preview' text button. The whole viewing/previewing menu needs some attention – obviously that's a separate endeavour.

> Here’s the max-width it’s still shown at when editing a page

I think that looks okay, and leaves the door open to something like https://github.com/WordPress/gutenberg/pull/62636#issuecomment-2180960878 down the road.

stokesman:
> Ah, I guess it's because the Preview action is bundled in the device preview toggles on desktop

Right, thank you for reminding me. So the minor issue there is is that currently the separate Preview button isn’t shown at the same breakpoint that the Device Preview dropdown is hidden. I've reverted the change I made regarding that and because that affects the space available for the Document Bar (while editing a page) revised the custom breakpoint used to hide it. It’s now at 403px instead of 382px. Here are the latest page/template captures for the max-width it’s still present.

Editing a page (the Document Bar is still present but totally obscured):
image

Editing a template:
image

jameskoster:
Obviously the mobile experience is not perfect, but the enhancements here are a step in the right direction compared to trunk.

I think this one just needs a code review.

richtabor:
@t-hamano @Mamaduka — if one of you could give this a good look-over, it'll be good to go. Thanks!

Mamaduka:
@stokesman, do you mind rebasing this on top of the latest trunk?

I don't like some metric changes (`type` and `focus`) in the current [performance test reports](https://github.com/WordPress/gutenberg/actions/runs/9667132009?pr=62636). The `useHasBlockToolbar` hook has been known to [affect those](https://github.com/WordPress/gutenberg/pull/58979#issuecomment-1970770057), and I would like to confirm it.

cc @jeryj

packages/editor/src/components/collapsible-block-toolbar/index.js
jeryj: I agree with you on the structure. That's how I originally did it as well, but it ended up having a performance impact due to the underlying 'useHasAnyBlockControls' that checks every slot in the toolbar to see if the toolbar should show or not.

By moving the check here, it only runs when we want to show the top toolbar. If I remember right, this performance impact is only significant while on the default floating toolbar, so this was an appropriate fix. I should have left a code comment about it.

packages/editor/src/components/collapsible-block-toolbar/index.js
jeryj: Nice catch!

stokesman:
> I don't like some metric changes (`type` and `focus`)

I appreciate you pointing that out. Looks like [this time around](https://github.com/WordPress/gutenberg/actions/runs/9780322604?pr=62636) `type` came out fine but not some of the other `type*`s and `focus` is still slower. Absolutely, the differences in milliseconds seem negligible but I get how we don’t want it creeping up. I’ll look into a way to avoid it and otherwise I'm okay with omitting the change to the conditional rendering for this PR if considered too impactful on performance.

stokesman:
I've now reverted the conditional rendering changes to see how that’d affect the metrics. Here’s [the run after](https://github.com/WordPress/gutenberg/actions/runs/9788015461?pr=62636). It helped with `focus` but oddly the `type` and `inserterSearch` ones didn’t go well. I hardly know what to make of it.

Anyway, I think it’s fine to leave those out and focus the PR on the the layout improvements. CSS could be used to achieve what the conditional rendering changes were doing and I’ll look to try that separately.

packages/editor/src/components/collapsible-block-toolbar/index.js
stokesman: Makes sense. Thanks for your comment. I guess the use cases for a way to check whether a block actually has toolbar controls are few so the effort to find a way of doing it that’s more performant has so far been spared. I like that you had some ideas https://github.com/WordPress/gutenberg/pull/58979#issuecomment-1971681017

> * Adding an `isBlockToolbarAvailable` state to the block editor store

It seems something like that would be good though I don’t know how it’d work.

Anyway, I've reverted this to have the PR focused on the layout improvement.

packages/editor/src/components/header/back-button.js
```diff
@@ -9,16 +9,16 @@ import {
...
 
 const BackButton = Fill;
-const BackButtonSlot = ( { children } ) => {
+const BackButtonSlot = () => {
 	const fills = useSlotFills( slotName );
-	const hasFills = Boolean( fills && fills.length );
-
-	if ( ! hasFills ) {
-		return children;
-	}
```
Mamaduka, line 21: @stokesman Is this change required? If yes, can you elaborate more on "why"? Thanks.

packages/editor/src/components/header/index.js
```diff
@@ -82,21 +83,29 @@ function Header( {
...
+			} ) }
+		>
+			{ hasBackButton && (
+				<motion.div
+					variants={ backButtonVariations }
+					transition={ { type: 'tween' } }
+				>
+					<BackButton.Slot />
+				</motion.div>
+			) }
```
Mamaduka: Will there be a case when a "back button" is missing? Seems like a problematic path for fullscreen users.

Cc @youknowriad (since you worked on this unification recently).

youknowriad: I don't understand the question. Would you mind clarifying?

packages/editor/src/components/header/index.js
```diff
@@ -82,21 +83,29 @@ function Header( {
...
-			<motion.div
-				variants={ backButtonVariations }
-				transition={ { type: 'tween' } }
-			>
-				<BackButton.Slot />
-			</motion.div>
+		<div
+			className={ clsx( 'editor-header edit-post-header', {
+				'has-back-button': hasBackButton,
+				'has-center': hasCenter,
```
youknowriad: Do we really need this kind of classes? Can't we figure out a unified layout that works without requiring these?

packages/editor/src/components/header/index.js
youknowriad: I think the "back button" is always present. It can be overridden by third-parties though.

Mamaduka: > I think the "back button" is always present. It can be overridden by third-parties though.

Exactly. The header should always render a "back button", either the default or one provided by third-parties. This condition and the `BackButton` slot-fill changes seem unnecessary.

stokesman: > Will there be a case when a "back button" is missing?

In the Post editor the back button is missing when not fullscreen (and at < 783px even with fullscreen preference on). Currently it’s done via CSS. The header layout needs to account for that and rather then add overriding styles in the edit-post package, I thought these changes would be cleaner overall and support the conceivable use case for a third-party to not render a back button (without them having to add overriding styles to avoid empty space). I’d noted this first in https://github.com/WordPress/gutenberg/pull/62636#issuecomment-2179369654 and later mentioned why I opted for these changes https://github.com/WordPress/gutenberg/pull/62636#issuecomment-2181058765 (in the latter part, after the `hr`).

packages/editor/src/components/header/back-button.js
stokesman: I didn’t have to remove this but it seemed redundant if we let the `Header` decide whether to render the slot. "Why" is hopefully clarified in my reply to your other comment on the header.

packages/editor/src/components/header/index.js
stokesman: Potentially more advanced and probably less readable selectors in the CSS could be used. I have another design change I’d like to propose for the header that should eliminate the need for `has-center` and simplify some CSS but it’s too distinct for this PR.

I’ll take another look anyway. I don’t love the classes either.

stokesman: Okay, done in 574b8bc4c529e17c3c70d10cd1543f671756c37c and 884c534862dfb78a21f20f235bf49c9a38a79d75. I still don’t love using `:has()` but perhaps I’ll be able to get rid of it with the aforementioned change I’ll propose in a later PR.

packages/editor/src/components/header/back-button.js
Mamaduka: Thanks for the explanation, @stokesman! I guess it's okay to change the component's behavior; it's a private component, so there are no requirements for BC.

Mamaduka (approved):
I think this is good to land.

stokesman:
Let’s 🚢. Thanks everyone who tested and reviewed.

packages/editor/src/components/header/style.scss
```diff
@@ -15,19 +25,32 @@
...
-	overflow: hidden;
-	// Leave enough room for the focus ring to show.
-	padding: 2px 0;
+	// Make narrowing to less than content width possible. Block toolbar will hide overflow and allow scrolling on fixed toolbar.
+	min-width: 0;
 	align-items: center;
-	// Allow focus ring to be fully visible on furthest right button.
-	@include break-medium() {
-		padding-right: var(--wp-admin-border-width-focus);
+	// Clip the box while leaving room for focus rings.
```
ellatrix, line 50: Why was this added? It's hard to understand from this comment, it doesn't mention why there's a clip at all

stokesman: Without it, on very narrow screens the sibling (`.editor-header__settings`) overlaps:
image

Here’s with the clip:
image

PR #62610, opened 2024-06-17
Global Styles: Add style variation partial cache

aaronrobertshaw:
Related:
- https://github.com/WordPress/gutenberg/pull/62529

## What?

Adds caching to processed theme.json partials for style variations.

## Why?

With registration of block style variations moved to a separate `init` action, the processing of `/styles` directories, including parsing and translating variation partials, was happening multiple times.

## How?

Processes all variation partials and stores them in a cache keyed by the theme, then locale.

## Testing Instructions

- Unit tests should be passing
- Smoke test display and application of theme style variations in the site editor
- Test section styles are still registered correctly and applying them works
    - Detailed instructions on setting up custom block style variations for testing section styles can be found in [#57908](https://github.com/WordPress/gutenberg/pull/57908)

PR #62638, opened 2024-06-18
Theme JSON resolver: read theme.json files from the styles/ folder only once

ramonjd:
## What?
Cache reading theme.json files from styles/ folder.

Syncs the following Core patch to Gutenberg:

- https://github.com/WordPress/wordpress-develop/pull/6843

## Why?

From https://github.com/WordPress/wordpress-develop/pull/6843:

> I've realized that, unlike the theme.json files defined by core & theme, we don't cache these — hence we end up reading them a few times from the filesystem.

## How?

From https://github.com/WordPress/wordpress-develop/pull/6843:

> By using the existing read_json_file function instead of using wp_json_file_decode directly.

## Testing Instructions

From https://github.com/WordPress/wordpress-develop/pull/6843:

1. Verify the theme style variations are working as expected.

Go to "Site Editor > Styles" and apply one of them. Verify the changes are reflected in the frontend. 

Do the same from the global styles sidebar in the site editor.

2. Verify the block style variations defined via theme.json are working as expected.

Create a `partial.json` file within the `styles/` folder with the following contents:

```json
{
        "version": 2,
        "title": "Partial",
        "styles": {
                "color": {
                        "background": "aliceblue"
                }
        }
}
```

Go to any editor, add a group block, and verify there is a "Partial" style variation (Block Settings > Styles). Apply the variation and save the changes. Verify the contents are the expected (background color is aliceblue) — also in the frontend.

### DIFF
--- a/lib/class-wp-theme-json-resolver-gutenberg.php
+++ b/lib/class-wp-theme-json-resolver-gutenberg.php
@@ -761,7 +761,7 @@ class WP_Theme_JSON_Resolver_Gutenberg {
 		}
 		ksort( $variation_files );
 		foreach ( $variation_files as $path => $file ) {
-			$decoded_file = wp_json_file_decode( $path, array( 'associative' => true ) );
+			$decoded_file = self::read_json_file( $path );
 			if ( is_array( $decoded_file ) && static::style_variation_has_scope( $decoded_file, $scope ) ) {
 				$translated = static::translate( $decoded_file, wp_get_theme()->get( 'TextDomain' ) );
 				$variation  = ( new WP_Theme_JSON_Gutenberg( $translated ) )->get_raw_data();


aaronrobertshaw (approved):
LGTM 👍 

Tests as advertised.

✅ Theme style variations are available as before
✅ Block style variations are also correct
✅ Works with child theme definitions

| Theme Style Variations | Block Style Variations |
|---|---|
| Screenshot 2024-06-18 at 2 32 42 PM | Screenshot 2024-06-18 at 2 32 51 PM |

ramonjd:
From https://github.com/WordPress/wordpress-develop/pull/6843#pullrequestreview-2123879525

> I wonder if we can also avoid the potential of processing the same decoded files multiple times as well?

I briefly played around with caching decoded, processed variation files in a local class property. 

I took "processed" to mean:

- run through `static::translate`, then
- using the output to generate a new WP_Theme_JSON_Gutenberg object

Pretty naive implementation, by using the variation file path as a key, similar to `theme_json_file_cache`. 

Seems to work with _very_ limited testing, but probably deserves its own PR with test cases. Also, the question of whether to extend caching to other theme data. Caching is hard. 😄

aaronrobertshaw:
> I briefly played around with caching decoded, processed variation files in a local class property.

Sounds similar to the original PR: https://github.com/WordPress/gutenberg/pull/62610

I'm sure we can narrow down a solution. This is a good first step though.

PR #62642, opened 2024-06-18
Fix e2e tests in dev mode

kevin940726:
## What?

Originally reported in [Slack](https://wordpress.slack.com/archives/C02QB2JS7/p1718367320344229). Since #62398, `npm run dev` no longer produces commonJS code, which breaks running e2e tests in dev mode. This PR fixes that with a duct-tape for now.

## Why?

It's convenient to run e2e tests locally when in development mode.

## How?

It seems like we are only using `@wordpress/keycodes` for now in the `e2e-test-utils-playwright` package. We only have to inline some of the code to fix this issue. Note that this might not be the final solution though, but seems the easiest for now.

We should probably look into other cases as well.

## Testing Instructions

1. Run `npm run dev`.
2. Run `npm run test:e2e -- --ui`
3. Expect the e2e test UI to start up

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54051,9 +54051,6 @@
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
-				"@wordpress/api-fetch": "file:../api-fetch",
-				"@wordpress/keycodes": "file:../keycodes",
-				"@wordpress/url": "file:../url",
 				"change-case": "^4.1.2",
 				"form-data": "^4.0.0",
 				"get-port": "^5.1.1",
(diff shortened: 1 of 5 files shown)


gziolo:
It looks like a good intermediate solution. In the long run, we will have to switch packages to ESM as the default type and define CJS only as a fallback using the combination of [`type`](https://nodejs.org/api/packages.html#type) and [`exports`](https://nodejs.org/api/packages.html#exports).

`npm install` is still necessary to make the CI pass. The rest looks good.

gziolo (approved):

PR #62641, opened 2024-06-18
WP 6.6 Beta 3 🍒

ellatrix:
## What?

Backporting all PRs with the "Backport to WP Beta/RC" label.

https://make.wordpress.org/core/handbook/about/release-cycle/block-editor-release-process-for-major-releases/#cherry-picking-prs

PR #62578, opened 2024-06-14
Global Styles: Don't apply the background and text colors to typography previews

scruffian:
## What?
Fix #61217. Alternative to #61898. 

Removes all style rules from color and typography variation previews, except the for the property that is in the preset.

## Why?
The typography presets only change typography styles so it makes sense to keep these preview clear of any other settings.

## How?
Pass the property that we are previewing to the variation component as a prop. If present we use this prop to filter out only this style so it can be used in the preview.

## Testing Instructions
1. Open the site editor
2. Open the Styles section in preview mode
3. Check that the typography previews don't have background colors
4. Open Global Styles > Typography
5. Check that the typography previews don't have background colors

### Testing Instructions for Keyboard
This is a visual change.

## Screenshots or screencast 
Screenshot 2024-06-14 at 12 42 51
Screenshot 2024-06-14 at 12 42 40

PR #62643, opened 2024-06-18
Fix cherry-pick script, add missing --repo arg

ellatrix:
## What?

When cherry-picking for Beta 2 and Beta 3, the script failed to comment and remove labels from PRs with the following error:

```
Error: GraphQL: Could not resolve to a PullRequest with the number of 62529. (repository.pullRequest)
```

Additionally, the script removes and then re-adds the label. See [example](https://github.com/WordPress/gutenberg/pull/62578#issuecomment-2175397346). So I deleted the line that adds the label back, I'm not sure why it's there.

## How?

It turns out that the `--repo` arg is missing.

## Testing Instructions

It was tested in #62641.

### DIFF
--- a/bin/cherry-pick.mjs
+++ b/bin/cherry-pick.mjs
@@ -380,15 +380,34 @@ function reportSummaryNextSteps( successes, failures ) {
 function GHcommentAndRemoveLabel( pr ) {
 	const { number, cherryPickHash } = pr;
 	const comment = prComment( cherryPickHash );
+	const repo = 'WordPress/gutenberg';
 	try {
-		cli( 'gh', [ 'pr', 'comment', number, '--body', comment ] );
-		cli( 'gh', [ 'pr', 'edit', number, '--remove-label', LABEL ] );
+		cli( 'gh', [
+			'pr',
+			'comment',
+			number,
+			'--repo',
+			repo,
+			'--body',
+			comment,
+		] );
+		cli( 'gh', [
+			'pr',
+			'edit',
+			number,
+			'--repo',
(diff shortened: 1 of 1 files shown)


bin/cherry-pick.mjs
```diff
@@ -380,15 +380,17 @@ function reportSummaryNextSteps( successes, failures ) {
 function GHcommentAndRemoveLabel( pr ) {
 	const { number, cherryPickHash } = pr;
 	const comment = prComment( cherryPickHash );
+	const repo = [ '--repo', 'WordPress/gutenberg' ];
```
youknowriad: Does it fail because you have multiple origins or something?

ellatrix: I have no idea why it's failing for me without explicitly passing it. Could be multiple remotes, yes. Or a different `gh` version? In any case, seems good to pass explicitly?

youknowriad (approved):

bin/cherry-pick.mjs
ellatrix: @youknowriad Sorry, I'm completely unfamiliar with `gh`. Looks like it's using a security remote for me, but the `origin` remote is set to this one... 🤷‍♀️ I don't know why it's using that and how to change it.

ellatrix: Anyway, we've hardcoded it in other places, so updated those too: https://github.com/WordPress/gutenberg/pull/62643/commits/555040c933d37c3e398f27421b1c6bd6ca3c3a19

PR #62647, opened 2024-06-18
DataViews: Register the permanentlyDelete action like any third-party action.

youknowriad:
Related #61084 

## What?

The current PR explore the possibility to use the API to register one action: "permanently delete". Check the inline comments for more details about the decisions made...

## Testing Instructions

1- Open the "trash" pages dataviews.
2- You should be able to see the "permanently delete" action in the actions dropdown menu


packages/editor/src/components/post-actions/actions.js
```diff
@@ -1007,6 +918,7 @@ export const duplicateTemplatePartAction = {
 };
 
 export function usePostActions( { postType, onActionPerformed, context } ) {
+	useDefaultActions();
```
youknowriad, line 921: I think this is probably not the right place to use this hook. 

Because ideally, this only run ones and third-parties should be able to unregister the core actions even before they are actually rendered. So I'm not sure what the best path forward here:

 - Move this call to the top level component in both post and site editors? Not ideal to have to duplicate this call. 
 - Avoid using a hook and just call this top level, when loading the module. 
 - Use a "filter" instead of suggesting to third-parties to use the unregister functions?

@gziolo maybe you have better ideas here since you worked on similar things for blocks registrations..

packages/editor/src/dataviews/actions/delete-permanently.ts
```diff
@@ -0,0 +1,120 @@
...
+					);
+				}
+				// If we were trying to permanently delete multiple posts
+			} else {
+				const errorMessages = new Set();
+				const failedPromises = promiseResult.filter(
+					( { status } ) => status === 'rejected'
+				);
+				for ( const failedPromise of failedPromises ) {
+					const result = failedPromise as CoreDataError;
```
youknowriad, line 86: This type cast is not ideal, but I'm not sure what I need to change to avoid it. It comes from CoreData, so it might be challenging to type that properly. cc @sirreal @jsnajdr

packages/editor/src/dataviews/store/private-selectors.ts
```diff
@@ -1,15 +1,22 @@
...
-const EMPTY_ARRAY: Action< any >[] = [];
-
-export function getEntityActions( state: State, kind: string, name: string ) {
-	return state.actions[ kind ]?.[ name ] ?? EMPTY_ARRAY;
-}
+export const getEntityActions = createSelector(
+	( state: State, kind: string, name: string ) => {
+		return [
+			...( state.actions[ kind ]?.[ name ] ?? [] ),
+			...( state.actions[ kind ]?.[ '*' ] ?? [] ),
```
youknowriad, line 15: I implemented the `*` to register actions for all post types.

packages/editor/src/dataviews/types.ts
```diff
@@ -0,0 +1,17 @@
...
+type PostStatus =
+	| 'published'
+	| 'draft'
+	| 'pending'
+	| 'private'
+	| 'future'
+	| 'auto-draft'
+	| 'trash';
+
+export interface Post extends AnyItem {
```
youknowriad, line 15: Ideally these "Model" types should be in core-data but it's fine for now here. I think ultimately the whole "dataviews" folder here will move to "core-data".

packages/editor/src/dataviews/types.ts
```diff
@@ -0,0 +1,17 @@
...
+	status?: PostStatus;
```
youknowriad: @mcsf @jsnajdr  @sirreal This is not ideal. I was forced to make the status optional but in reality it's mandatory in "Post" objects. The reason I was forced to do that is that otherwise `Action<Post>` would give me an error saying `status` is not always present in `AnyItem`. I think this highlights something wrong in our "Action" types but I'm not really sure how to solve that.

packages/editor/src/dataviews/actions/delete-permanently.ts
```diff
@@ -0,0 +1,120 @@
...
+import {
+	TEMPLATE_POST_TYPE,
+	TEMPLATE_PART_POST_TYPE,
+} from '../../store/constants';
+
+type CoreDataError = {
+	reason: { message?: string };
+};
+
+const permanentlyDeletePostAction: Action< Post > = {
```
jorgefilipecosta, line 23: Hi @youknowriad, in order to know if the user has the right permissions to delete or trash a post, we need to call some selectors. Currently, we have regressions and are showing trash action even when we should not. I'm fixing that for Trash action at https://github.com/WordPress/gutenberg/pull/62589/ (which we should include for 6.6). We also need to do the same to delete permanently but we could not use a selector to check permissions in the API I guess we need to fix that at the API level.

jorgefilipecosta: Also, something where we would check permissions to delete before registering the action would not work, the user may be able to delete some items (e.g.: the items they created or draft items) but not other items.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: I think the "Move this call to the top level component in both post and site editors?" is the better approach. Could we put on some top-level editor component used by both edit site and edit post so we avoid the duplication?

packages/editor/src/dataviews/actions/delete-permanently.ts
youknowriad: The "callback" can receive the "registry" object, we might want to do the same for the isElligible.

packages/editor/src/components/post-actions/actions.js
youknowriad: The site editor is not just an "Editor", it's the equivalent of the "whole admin". The editor can be unmounted there.

packages/editor/src/dataviews/actions/delete-permanently.ts
jorgefilipecosta: Hi @youknowriad, I think simply passing the registry to isElligible is not a solution and will not work as expected. In the beginning, while checks are being done canUser will return false, and if isEligible is false then canUser may say true but isEligible is not re-executed because there is nothing that causes a React rerender. This is not a theoretical issue it was happening on https://github.com/WordPress/gutenberg/pull/62589 and that was the reason I used the hacky solution of getCachedResolvers around useMemo to trigger the rerender.


I see two possible ways forward:
- Support hook-based actions on the API.
- Make is eligible something async and then it is up to the component users of isEligible to make sure the component rerender after isEligible finishes the computation.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Good point in that case I would still opt for "Move this call to the top level component in both post and site editors?" and would duplicate the call on edit post and edit site, it is just one line of code.

packages/editor/src/dataviews/actions/delete-permanently.ts
youknowriad: What if we just call isEligible within useSelect and pass "select". I'm not sure if I like it but it's a solution as well.

packages/editor/src/components/post-actions/actions.js
gziolo: Regarding block registration, we put some similar code in all top-level components of the usual apps: editor, post editor, widgets editor, etc.

Site editor:

https://github.com/WordPress/gutenberg/blob/1050efa5bb2d9b620469a91eab15b7bd53de2422/packages/edit-site/src/index.js#L37-L41

Post editor:

https://github.com/WordPress/gutenberg/blob/1050efa5bb2d9b620469a91eab15b7bd53de2422/packages/edit-post/src/index.js#L73-L86

There are some legacy implications, so it might be more complex than necessary there. However, the biggest challenge with filters is that they might get applied before or after the top-level component gets mounted, or even before or after the individual store action implementing a filter gets executed. In effect, we decided in the past to account for it. The comment in the code might better illustrate that:

https://github.com/WordPress/gutenberg/blob/1050efa5bb2d9b620469a91eab15b7bd53de2422/packages/blocks/src/store/actions.js#L33-L46

As with everything, it all comes with defining the overall strategy knowing that 3rd party depending on the defined script dependencies might load before or after the code that defines the default data view action handlers. The biggest challenge with unregistering the action handlers is also the fact that you need to ensure that the script gets executed after it gets registered, which is sometimes problematic when some other code has the opposite requirement. I'm mostly illustrating the considerations that we have been witnessing in the bug reports over time with the extensibility model we iterated upon for block registration. I'm happy to discuss the implications of the approach you prefer to take here and help to think about potential edge cases.

packages/editor/src/dataviews/actions/delete-permanently.ts
jsnajdr: The problem is that `failedPromises` is an array of `PromiseFulfilledResult | PromiseRejectedResult`. The `filter` function ensures that it's an array of only `PromiseRejectedResult`s, but TypeScript doesn't make that inference. Maybe TS 5.5 will do better.

And of course TS also can't do the more complicated inference that:
- if `result.every(r => r.status === 'fulfilled') === false`
- and `result.length === 1`
- then necessarily `result[0].status === 'rejected'`.

Only `PromiseRejectedResult` has the `reason` field, so you need to cast `failedPromise` manually before accessing it. `PromiseSettledResult` has a `value` field.

Another thing that's good to know that TypeScript has full types only for resolved promises. Errors are not typed. `Promise<T>` means that the resolved value is of type `T`, but doesn't say anything about what is the error value when rejected. That's always `any`.

Instead of casting `failedPromise` to `CoreDataError` you could also cast it to `PromiseRejectedResult`, but then `failedPromise.reason` is `any`. With `CoreDataError` you have a better type.

packages/editor/src/dataviews/actions/delete-permanently.ts
jorgefilipecosta: > What if we just call isEligible within useSelect and pass "select". I'm not sure if I like it but it's a solution as well.

It is also a possible solution, it is going to be called a lot e.g: in a list with 50 times we will have 50 useSelects calling is eligible running on each store change. That may be ok. Another downside is that it is not very flexible it forces people to use WordPress data e.g.: I can not check if something is eligible on my own API without using resolvers etc.

packages/editor/src/dataviews/actions/delete-permanently.ts
jsnajdr: I confirmed that TypeScript 5.5.0 is going to be much smarter about this code. In [this example](https://www.typescriptlang.org/play/?target=99&ts=5.5.0-beta&ssl=1&ssc=1&pln=7&pc=2#code/IYZwngdgxgBAZgV2gFwJYHsL3egPAFQD4AKABwCd0BbEALgAVKrUQBTAwgbQF0BKGAN4AoGDCiYQyGOVYgYAXhjAA7sFRTG1FqwB0wADb6Ayq2TJ9rACZkmIXgG4RYiVLhr9CGXMVedcVPrIrOTE5AqE0jqSwMgI3vKKAOQyAFasUEGWiQ5OcOhhxOIQkvDunqww6HClAeV2gk6iRSDoFjr66ADmxG61MjoyoJg5ogC+QuNAA):
```js
async function foo<T>(proms:Promise<T>[]) {
  const res = await Promise.allSettled(proms);
  const failures = res.filter(r => r.status === 'rejected');
  for (const failure of failures) {
    console.log(failure.reason);
  }
}
```
the new 5.5.0 version automatically detects that `failures` is a `PromiseRejectedResult[]` and that it's OK to access the `.reason` field. The current 5.4.5 will flag `.reason` as type error because it thinks `failures` is still the more general `PromiseSettledResult<T>[]`.

I'm looking forward to the new version 🙂

packages/editor/src/dataviews/types.ts
mcsf: I still don't understand the `AnyItem` / `Item extends AnyItem` abstraction, which means I'm probably not qualified to help here.

(Until then, my intuition is still that we are making extensive concessions in our type logic to appease the machine at the expense of simple, more maintainable types.)

mcsf: That said, from what I know of TS, I'm not sure `Action` is necessarily the culprit, but maybe `Post`.

```ts
interface Post extends AnyItem {
	status?: PostStatus;
}

type AnyItem = Record< string, any >;
```

Extending an interface A into an interface B means something straightforward: you declare a new interface which inherits the properties of its parent. But what does it mean to "extend" a custom record type (`AnyItem`) into an _interface_? The semantics of an interface are stricter than that of a record.

mcsf: > Extending an interface A into an interface B means something straightforward: you declare a new interface which inherits the properties of its parent. But what does it mean to "extend" a custom record type (`AnyItem`) into an _interface_? The semantics of an interface are stricter than that of a record.

After playing a bit in the TS Playground, I withdraw this suspicion. :)

jsnajdr: `B extends A` means that `B` is a subtype of `A`. That means that a value of type `B` can be safely used everywhere `A` is required.

Extending classes in classical OOP is one example of subtypes. If type `A` has a `a: string` field, and `B` extends that with `b: string` field, then `B` has both `a` and `b` fields. And `B` can be used in place of `A` because it has a compatible `a` field.

Another example of a subtype is when `'a' | 'b'` is a subtype of `string`. Of course, if a `string` is expected somewhere, surely you can use `'a'` or `'b'` there. But it seems paradoxical because `'a' | 'b'` is hardly an extension of `string`. Quite vice versa, the `'a' | 'b'` type is much smaller than `string`!

But then you need to realize that the set of all possible `B`s is also much smaller than the set of `A`s. Because `B` is a subset of `A`: it's a set of records that also have the `b` field in addition to `a`.

I think that the `extends` keyword is a bit confusing. It comes from classic OOP languages where classes and subclasses are the only usage of subtypes. But it feels counterintuitive when applied to other types.

mcsf: Thanks for taking the time to explain things. :) I did know these things, but it's always useful to check our assumptions, and I agree with you that "extends" was an interesting choice of word. I conjectured a bit around `Post` but my suspicion was premature.

jsnajdr: @youknowriad It gets fixed for me when `ActionButton` extends from `ActionBase<Item>` instead of `ActionBase<AnyItem>`. It was probably a typo because the `ActionModal` does the extension correctly.

jsnajdr: In the error message for `permanentlyDeletePostAction`, this part is the most important:
```
Types of property 'isEligible' are incompatible.
      Type '({ status, type }: Post) => boolean' is not assignable to type '(item: AnyItem) => boolean'.
```
The `isEligible` method is on the `ActionBase` type, and if the type is `ActionBase<AnyItem>`, the type of `isEligible` is:
```ts
( item: AnyItem ) => boolean
```
Because `Action<Post>` extends from `ActionBase<AnyItem>`, the `isEligible` method is expected to accept any `item` that has type `AnyItem`.

But you're assigning to `permanentlyDeletePostAction` an object whose `isEligible` is of a narrower type:
```ts
isEligible( { status, type }: Post )
```
This function doesn't accept an argument of type `AnyItem`. Because on `AnyItem` the `status` field can be anything, it can be `undefined` or anything else. But your `isEligible` expects it to be `PostStatus` and nothing else.

This gets fixed when `Action<Post>` extends from `ActionBase<Post>` because then the type of the `isEligible` `item` parameter is also `Post`.

I'm surprised that TypeScript was happy with the optional `status?: PostStatus`. Because that's still incompatible with `AnyItem` because `AnyItem` can have `status` value of type `number`, which is not compatible with `status?: PostStatus`.

jsnajdr: > I'm surprised that TypeScript was happy with the optional `status?: PostStatus`.

[This example](https://www.typescriptlang.org/play/?target=99&noImplicitOverride=true&ts=5.4.5#code/C4TwDgpgBAggdiAksCBbKBeKAlCBjAewCcATAHgGdgiBLOAcwBooBDBAPgG4AobulIgDMWeaAAUCVKBAAeKOCQqwEyNFADe3KFCotgAVwoAuHdTr0eAX178IQkdABCLChDIAVdhq1QaFAKIANjT0NABGgRAAFDQoqEbuAJRGYQQEkWxWvIRwUmEuECbOrmTwSHFeWJrafkEh4ZEmMXFGElSJmF6xaAB0ugYUfcAsRMAUAOqxABZRAOQQwUazidzW3PmuPbXBoRHR6v2GJnD6gYGWiZxAA) behaves in a way that I don't understand:
```ts
type AnyItem = Record<string, any>;

interface Post extends AnyItem {
  status: string;
}

interface Base<T> {
  isEligible(item: T): boolean;
}

const base: Base<AnyItem> = {
  isEligible: (item: Post) => item.status.startsWith('eli:')
}

base.isEligible({status: null});
```
In this program the types of `isEligible` don't match, it will crash at runtime (accessing `.startsWith` of `null`) and yet TypeScript doesn't report any type error. @sirreal Do you have any idea what's going on?

sirreal: Here's what I think is going on. It seem that a few things together create problems with the example.

- `AnyItem` is kind of a dangerous type. It says that it's an object where every string property has the value of type `any`. `any` is dangerous! If we don't just extend this type but actually use the type (like in `const base: Base<AnyItem>`), then `any` sneaks into the type system and wreaks havoc.
- ~`AnyItem` can be assigned to `Post`. This allows use to `Post` where we expect `AnyItem` (in the isEligible argument type).~ **This is wrong. See my next comment.**
- A type annotation `Post` is included on the argument `item` in `isEligible( item: Post )`. Because `AnyItem` is assignable to `Post`, this is allowed. If we'd used e.g. `string`, we'd see an error like this:
  > Types of parameters 'item' and 'item' are incompatible.
  >   Type 'AnyItem' is not assignable to type 'string'.

[I'd make a few changes:](https://www.typescriptlang.org/play/?target=99&noImplicitOverride=true&ts=5.4.5#code/C4TwDgpgBAggdiAksCBbKBeKAlCBjAewCcATAHgGdgiBLOAcwBooBXOAazgIHc4A+ANwAoIXRREAZgEM80RMjRQA3kKhQqU4CwoAuddTr1hagNoAPPVVoMAunracecIQF8RYiJJnQACgSpQEGYocCQUsAgK6CpqGlq6+tZGru5w4tKyUABCUhQQZAAqgcEQoeHwSCiofMqqUDQUAKIANjT0NABGzRAAFDRVOgUAlDodBATdUnDCbkKEcAEduRB6OXlkFVE1WDH1Ta3tXStQfVVDmDX9aAB0cdq3wFJEwBQA6v0AFj0A5BCtOt8hiklnlrg0Wm1Ot0eko7gk4Cxms0XENhHMCAtgFAQRA-FRVssyHjgNtampwQcocdTmhzhhLlUHpp7hpnm9Pj8-jQAUC3DjiWD9pCjjC4ToEUiUQIgA)

- `AnyItem` uses `Record<string, unknown>`. This prevents `any` types from sneaking in.
- The argument type annotation on `item` is removed from `base`. We'll now see a type error on `item.status` if using `Base<AnyItem>`.
- If we expect errors on `base.isEligible({ status: null })` then base should be typed `Base<Post>` instead of `Base<AnyItem>`.

```ts
type AnyItem = Record<string, unknown>;

interface IItem {
  status: string;
  [x: string]: unknown
}

interface Post extends AnyItem {
  status: string;
}

interface Base<T extends AnyItem> {
  isEligible(item:T):boolean;
}

const base: Base<AnyItem> = {
  isEligible: (item) => item.status.startsWith('eli:') // TYPE ERROR: 'item.status' if of type 'unknown'
}
base.isEligible({status: null});

const basePost: Base<Post> = {
  isEligible: (item) => item.status.startsWith('eli:')
}
basePost.isEligible({status:null}); // TYPE ERROR: 'null' is not assignable to type 'string'
```

sirreal: I was reflecting further and trying to produce a more succinct, direct, and clear example and discovered something fascinating.


I'm able to fix your example with one small change to the `Base` interface:

```diff
 interface Base<T> {
-  isEligible(item: T): boolean;
+  isEligible: (item: T) => boolean;
 }
```

This seems to work because the function type becomes an atomic type, so when we try to assign `(item: Post) => boolean` to `(item: AnyItem) => boolean`, it shows a type error. Before, it was checking compatibility of the argument type. I can't really explain this or why that apparently superficial change would significantly change the behavior of the type system, I think it may be a bug.

<details>

<summary>Some tricks I tried to get a minimal reproducible example</summary>

I tried using a function type alias and function interface. They are slightly different but both give a type error:

```ts
type AnyItem = Record<string, any>;

interface Post extends AnyItem { status: string; }

type F<T> = (item: T) => boolean;
interface IF<T> { (item: T): boolean; }

interface Base<T> {
  isEligible: F<T>
  isEligible2: IF<T>
}

const base: Base<AnyItem> = {
  isEligible: (item: Post) => item.status.startsWith('eli:'),
/*
Type '(item: Post) => boolean' is not assignable to type 'F<AnyItem>'.
  Types of parameters 'item' and 'item' are incompatible.
    Property 'status' is missing in type 'AnyItem' but required in type 'Post'.
*/ 
  isEligible2: (item: Post) => item.status.startsWith('eli:'),
/*
Type '(item: Post) => boolean' is not assignable to type 'IF<AnyItem>'.
  Types of parameters 'item' and 'item' are incompatible.
    Type 'AnyItem' is not assignable to type 'Post'.
*/
}
```

</details>

---

This statement from my previous comment is incorrect:

> `AnyItem` can be assigned to `Post`. This allows use to `Post where we expect `AnyItem` (in the `isEligible` argument type).

It's easy enough to see that's not true:

```ts
declare var a: AnyItem;
const b: Post = a; // type error
```

What confused me was that as soon as we change this section in your example we get a type error:

```ts
const base: Base<AnyItem> = {
  isEligible: (item: string) => item.status.startsWith('eli:')
  // Type '(item: string) => any' is not assignable to type '(item: AnyItem) => boolean'.
  //  Types of parameters 'item' and 'item' are incompatible.
  //    Type 'AnyItem' is not assignable to type 'string'
}
```

If the error is that `AnyItem` is not assignable to `string`, when there's no error with `Post` we'd assume that `AnyItem` is assignable to `Post`, but we've just seen that isn't the case…

jsnajdr: > I still don't understand the AnyItem / Item extends AnyItem abstraction

I think we could remove the `AnyItem` type completely. And whenever there is a `Foo< Item extends AnyItem >` type parameter, make it just `Foo< Item >`, without any constraints.

Because what do the generic types (`Action<Item>`, `ViewProps<Item>`) really do with the `Item` type? They put it into arrays and callback parameters. You can do that with any type. The types like `Action` can be completely generic.

Constraining the `Item` type to objects, and forbidding strings for example, is nice, but not very valuable IMO.

youknowriad: I'm going to apply @jsnajdr's fix and we can consider removing AnyItem separately.

sirreal: I agree, I think that's a good idea.

packages/editor/src/dataviews/actions/delete-permanently.ts
sirreal: TypeScript 5.5 has been released so we can see about upgrading. It does bring some nice features.

packages/editor/src/dataviews/types.ts
sirreal: There's a lint rule to prevent using the weaker syntax [mentioned in this comment](https://github.com/WordPress/gutenberg/pull/62647#discussion_r1647980168), I'll proposed adding that rule in https://github.com/WordPress/gutenberg/pull/62718.

jsnajdr: I tried to remove `AnyItem` locally and it fails at three places:
1. `DataViews<Item>` has a `getItemId` prop which is a function of type `Item => string`. The prop is optional and defaults to `defaultGetItemId` which is like `item => item.id`. It works only when `Item` has a `id: string` field. Not for generic `Item`. To fix this, we must encode that `getItemId` is optional only when `Item` is a subtype of `{ id: string }`, otherwise you must supply a custom implementation.
2. Similar problem with defaults is in `normalizeFields`. There is a line:
   ```js
   const getValue = field.getValue || ( ( { item } ) => item[ field.id ] );
   ```
   Again, the default works only for specific subtypes where `Item` is a record that has a field that matches the `field.id` value. It's a type error to create a `Field<Item>` where `getValue` is fully optional.
3. In `ViewList<Item>` there is a line:
   ```js
   const selectedItem = data.findLast( ( item ) => selection.includes( item.id ) );
   ```
   Looking at `item.id` is not generic enough. The `getItemId` function should be used here.

Other than that, removing `AnyItem` works very well. And it turns out that the `Record<string,any>` construct was hiding some errors, because `any` tends to turn off typechecking. Now it's revealed that the `Field<Item>.id` should match the `Item`, and that it's not always safe to access `item.id`.

I pushed the branch, the commit id is a5cf5a4b1ca, you can have a look.

Maybe in the end we'll end up not removing `AnyItem`, but merely transforming it into `{ id: string }` base type, making `id` mandatory for all `Item`s.

youknowriad: >  but merely transforming it into { id: string } base type, making id mandatory for all Items.

Is it possible somehow to say items should either have `id` string  or provide a getItemId prop.

jsnajdr: I tried out this and it works:
```tsx
type ItemWithId = { id: string };

type DataViewsProps< Item > = {
  /* ... */
} & ( Item extends ItemWithId
  ? { getItemId?: ( item: Item ) => string }
  : { getItemId: ( item: Item ) => string } );

const defaultGetItemId = ( item: ItemWithId ) => item.id;

jsnajdr: I managed to fix all three type errors I mentioned above, here's the draft PR: #62856.

youknowriad:
So this PR is actually blocked. I can't rebase this PR unless we find a solution for this first. https://github.com/WordPress/gutenberg/pull/62647#discussion_r1644145645

youknowriad:
Going to close this PR for now, let's get back to this when we have a good way to access stores in `isElligible` callbacks.

PR #62648, opened 2024-06-18
Fix: a11y text for site editor

### LINKED ISSUE #62644: Site title: link text and accessible name mismatch
### Description

Follow up to https://github.com/WordPress/gutenberg/pull/61258

The Site Title in the Site editor navigation panel is now a link to the front end. So far so good. However, there's a couple issues to fix:

1. The link visible text and the link accessible name mismatch. This is not ok for a11y and prevents some users from understanding and using the link. Sighted screen reader users, speech recognition software usrers will be impacted.
  - The visible text is the `blogname` value, for example: "My awesome site".
  - The accessible name is determined by the link `aria-label="View site (opens in a new tab)"` which is an hardcoded string.
2. The icon (actually a `↗` unicode character via a CSS pseudo element) mismatches other icons used for the same 'view site' functionality. I'm not sure this helps building a cohesive visual language. Icons usage should be consistent, as also mentioned in the original PR https://github.com/WordPress/gutenberg/pull/61258#pullrequestreview-2033031447

- Go to the site editor.
- With the navigation panel open, observe the site title link has an `aria-label="View site (opens in a new tab)"` that mismatches the visible text.
- Hover or focus the link
- Observe the icon that appears is different from the icon used for other 'view site' links.

Screenshot 2024-06-18 at 09 44 41

### Environment info

_No response_

Yes

Yes

up1512001:
## What?
- updated aria label text as site text 

## Why?
Fixes https://github.com/WordPress/gutenberg/issues/62644

## How?
- using site title for aria label `label={ decodeEntities( siteTitle ) }`

## Screenshots or screencast 
Screenshot 2024-06-18 at 15 42 47

https://github.com/WordPress/gutenberg/assets/75293077/c356aeeb-50c7-4efc-bfcd-2e6b18b7bd06

### DIFF
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -88,7 +88,7 @@ const SiteHub = memo(
 								<VisuallyHidden as="span">
 									{
 										/* translators: accessibility text */
-										__( '(opens in a new tab)' )
+										__( 'View site (opens in a new tab)' )
 									}
 								</VisuallyHidden>
 							</Button>
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -88,7 +88,7 @@ const SiteHub = memo(
 								<VisuallyHidden as="span">
 									{
 										/* translators: accessibility text */
-										__( 'View site (opens in a new tab)' )
+										__( '(opens in a new tab)' )
 									}
 								</VisuallyHidden>
 							</Button>
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -7,7 +7,11 @@ import clsx from 'clsx';
  * WordPress dependencies
  */
 import { useSelect, useDispatch } from '@wordpress/data';
-import { Button, __experimentalHStack as HStack } from '@wordpress/components';
+import {
+	Button,
+	__experimentalHStack as HStack,
+	VisuallyHidden,
+} from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { store as coreStore } from '@wordpress/core-data';
 import { decodeEntities } from '@wordpress/html-entities';
@@ -78,9 +82,15 @@ const SiteHub = memo(
 								href={ homeUrl }
 								target="_blank"
 								icon={ external }
-								label={ decodeEntities( siteTitle ) }
+								iconPosition="right"
 							>
 								{ decodeEntities( siteTitle ) }
+								<VisuallyHidden as="span">
+									{
+										/* translators: accessibility text */
+										__( '(opens in a new tab)' )
+									}
+								</VisuallyHidden>
 							</Button>
 						</div>
 						<HStack
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -77,7 +77,7 @@ const SiteHub = memo(
 								variant="link"
 								href={ homeUrl }
 								target="_blank"
-								label={ __( 'View site (opens in a new tab)' ) }
+								label={ decodeEntities( siteTitle ) }
 							>
 								{ decodeEntities( siteTitle ) }
 							</Button>
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -12,7 +12,7 @@ import { __ } from '@wordpress/i18n';
 import { store as coreStore } from '@wordpress/core-data';
 import { decodeEntities } from '@wordpress/html-entities';
 import { memo, forwardRef } from '@wordpress/element';
-import { search } from '@wordpress/icons';
+import { search, external } from '@wordpress/icons';
 import { store as commandsStore } from '@wordpress/commands';
 import { displayShortcut } from '@wordpress/keycodes';
 import { filterURLForDisplay } from '@wordpress/url';
@@ -77,6 +77,7 @@ const SiteHub = memo(
 								variant="link"
 								href={ homeUrl }
 								target="_blank"
+								icon={ external }
 								label={ decodeEntities( siteTitle ) }
 							>
 								{ decodeEntities( siteTitle ) }
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -16,7 +16,7 @@ import { __ } from '@wordpress/i18n';
 import { store as coreStore } from '@wordpress/core-data';
 import { decodeEntities } from '@wordpress/html-entities';
 import { memo, forwardRef } from '@wordpress/element';
-import { search, external } from '@wordpress/icons';
+import { search } from '@wordpress/icons';
 import { store as commandsStore } from '@wordpress/commands';
 import { displayShortcut } from '@wordpress/keycodes';
 import { filterURLForDisplay } from '@wordpress/url';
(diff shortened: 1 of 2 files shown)


youknowriad:
I think @richtabor worked on this recently, so this might be a design regression a bit.

packages/edit-site/src/components/site-hub/index.js
```diff
@@ -77,7 +77,8 @@ const SiteHub = memo(
 								variant="link"
 								href={ homeUrl }
 								target="_blank"
-								label={ __( 'View site (opens in a new tab)' ) }
+								icon={ external }
+								label={ decodeEntities( siteTitle ) }
```
afercia: I'd suggest to entirely remove the label and use a `VisuallyHidden` text after the siteTitle to provide the info `(opens in a new tab)`. The visible text ant the actual accessible name willl be different but that's OK accorging to [WCAG Label in Name](https://www.w3.org/TR/WCAG22/#label-in-name).
See an example of this pattern at https://github.com/WordPress/gutenberg/blob/a1623c0216a58ce900cd0b56957789cec1606b68/packages/editor/src/components/preview-dropdown/index.js#L106-L111

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -33,6 +32,9 @@
 	text-decoration: none;
 	text-overflow: ellipsis;
 	white-space: nowrap;
+	display: flex;
+	flex-direction: row-reverse;
```
afercia: The Button component supports a prop and value `iconPosition="right"` to swap the icon position. Please use that instead of custom CSS. Worth noting:
- The button component is already `inline-flex` and already uses `align-items: center;` so the CSS could be simplified. However...
- The removed `display: block;` is necessary and needs to be restored otherwise `text-overflow: ellipsis;` will not work. Screenshot to illustrate the missing ellipsis:

Screenshot 2024-06-19 at 09 23 19

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -55,16 +57,20 @@
...
+	svg {
 		opacity: 0;
-		position: absolute;
-		right: 0;
-		transition: opacity 0.1s linear;
+		transition: opacity 0.1s ease-in-out;
 		@include reduce-motion("transition");
 	}
 
+	&:hover {
```
afercia: The icon should be revealed also on `:focus`.

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -55,16 +57,20 @@
...
+		svg {
+			opacity: 1;
+			transition: opacity 0.1s ease-in-out;
+			@include reduce-motion("transition");
```
afercia: Not sure repeating the transition and reduce-motion mixin here is necessary. I think these two lines should be removed, unless I'm missing something.

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -55,16 +57,20 @@
...
+		}
+	}
+
 	&:hover::after,
 	&:focus::after,
 	&:active::after {
```
afercia: this entire rule should be probably removed

afercia (changes requested):
Please see inline comments.

up1512001:
@afercia addressed your feedback please review the PR.

packages/edit-site/src/components/site-hub/index.js
```diff
@@ -77,9 +81,16 @@ const SiteHub = memo(
...
 								target="_blank"
-								label={ __( 'View site (opens in a new tab)' ) }
+								icon={ external }
+								iconPosition="right"
 							>
 								{ decodeEntities( siteTitle ) }
+								<VisuallyHidden as="span">
+									{
+										/* translators: accessibility text */
+										__( 'View site (opens in a new tab)' )
```
afercia:
```suggestion
										__( '(opens in a new tab)' )
```

For consistency with other occurrences I'd remove the 'View site' part.

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -36,8 +36,17 @@
 
 	&:hover,
 	&:focus,
-	&:active {
-		color: $gray-200;
+	&:active,
+	&:focus-visible {
+		svg {
+			opacity: 1;
+			transition: opacity 0.1s ease-in-out;
```
afercia: I'd suggest to remove the transition here and add it to the normal state of the svg.

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -36,8 +36,17 @@
...
+		}
+	}
+
+	svg {
+		opacity: 0;
+		vertical-align: middle;
```
afercia:
```suggestion
		vertical-align: middle;
		transition: opacity 0.1s ease-in-out;
		@include reduce-motion("transition");
```

packages/edit-site/src/components/site-hub/style.scss
```diff
@@ -36,8 +36,17 @@
 
 	&:hover,
 	&:focus,
-	&:active {
-		color: $gray-200;
```
afercia: I think this color needs to be kept at least for the :active state otherwise when clicking the link, the text will become black on a black background. Screenshot:
Screenshot 2024-06-24 at 11 34 43

afercia: Also, the button padding should be adjusted to keep the current spacing as much as possible. Screenshot with comparison of current trunk and this PR:

Screenshot 2024-06-24 at 11 34 55

up1512001:
@afercia updated as per your feedback. please check.

richtabor (changes requested):
> I think @richtabor worked on this recently, so this might be a design regression a bit.

Yes, let's resolve the a11y feedback apart from icon changes. 

The [ExternalLink component has been updated](https://github.com/WordPress/gutenberg/blob/d97e3f7a25ac28eebe7f0221860ab3c451835982/packages/components/src/external-link/index.tsx#L55-L79) to leverage the ↗ unicode character. This is intended to share the same DNA as that component, as it is also a text only external link.

up1512001:
@richtabor @afercia I have reverted style-related changes and only kept accessibility related changes.

afercia:
@up1512001 thank you. Looks good to me 👍🏽 
I will creat a follow-up issue for the  `↗` CSS pseudo element.

afercia (approved):
LGTM

up1512001:
@afercia I thought we should use the wordpress icon for consistency instead of the `after` pseudo element so whenever the icon is updated it will reflect to everywhere.
Let me know your thoughts.

afercia:
+1 when @richtabor approves.

afercia:
> @afercia I thought we should use the wordpress icon for consistency instead of the `after` pseudo element so whenever the icon is updated it will reflect to everywhere. 

@up1512001 personally I don't like the design of this component and I don't understand the need to replace the `external` icon with the North East Arrow unicode character. If it was up to me, I would change it but hte prevalent opinion from design is to keep the arrow so yes, fair enough 🙂

richtabor:
> I don't understand the need to replace the external icon with the North East Arrow unicode character. 

It's much more palatable when placed directly alongside text. The external icon was too big and disorienting when in those placements. Making it smaller made it less communicative/complex. 

I'm fine if we do the unicode symbol as an icon, resulting in no change visually.

afercia:
> Making it smaller made it less communicative/complex.

Making it smaller made it also less readable. See https://github.com/WordPress/gutenberg/issues/62832 for a more in depth analysis.

up1512001:
@richtabor Do you have any suggestions for this PR?
cc: @afercia

up1512001:
Hi @richtabor any feedback for this PR?

richtabor (approved):

Mamaduka (approved):

PR #62659, opened 2024-06-18
Convert to blocks: upgrade and remove spans if possible

ellatrix:
## What?

The idea behind "convert to blocks" (compared to past handling) is to preserve content as much as possible. However, at the moment, we are preserving it a bit too much.

* Spans with a normal font weight should be removed.
* Spans with a bold font weight should be converted to `strong`.
* Spins with an italic font style should be converted to `em`.
* `b` should be converted to `strong`.
* `i` should be converted to `em`.
* And so on.
* And remaining spans without any attributes should be removed. If there are remaining attributes, they should be preserved. (Key difference with paste.)

I've separated out the anchor reducer to only run on paste (because it modifies content).

## Why?

We have core formats for these, which is a better experience.
Spans that do nothing leads to a bad user experience.

## Testing Instructions

I've added an integration test.

Paste this in the code editor.

```
<span style="font-weight: 400">test</span>
```

Switch to visual and click "convert to blocks". There should be no spans.

### DIFF
--- a/packages/blocks/src/api/raw-handling/anchor-reducer.js
+++ b/packages/blocks/src/api/raw-handling/anchor-reducer.js
@@ -1,0 +1,25 @@
+export default function anchorReducer( node ) {
+	if ( node.nodeName === 'A' ) {
+		// In jsdom-jscore, 'node.target' can be null.
+		// TODO: Explore fixing this by patching jsdom-jscore.
+		if ( node.target && node.target.toLowerCase() === '_blank' ) {
+			node.rel = 'noreferrer noopener';
+		} else {
+			node.removeAttribute( 'target' );
+			node.removeAttribute( 'rel' );
+		}
+
+		// Saves anchor elements name attribute as id
+		if ( node.name && ! node.id ) {
+			node.id = node.name;
+		}
+
+		// Keeps id only if there is an internal link pointing to it
+		if (
+			node.id &&
+			! node.ownerDocument.querySelector( `[href="#${ node.id }"]` )
+		) {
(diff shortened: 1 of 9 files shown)


saulyz (approved):
tests on the Editor with various content elements pass

`...<li style="list-style-type: none;">dirty item</li>...`

`<span style="font-weight: 400">Some content</span>`

saulyz (approved):

saulyz (approved):

packages/blocks/src/api/raw-handling/anchor-reducer.js
```diff
@@ -0,0 +1,25 @@
+export default function anchorReducer( node ) {
```
ellatrix, line 1: This is just separated out from https://github.com/WordPress/gutenberg/pull/62659/files#diff-56e921b5ac7d8092c4085c3311a2611048fbd44a11f49de8e307bd796a4cb671

saulyz:
I've found that list items having spans are not processed well.
`<ul><li><span style="font-weight: 400">First list item</span></li><li><span style="font-weight: 700">Second list item</span></li></ul>` 

The conversion results to
`<!-- wp:list --><ul class="wp-block-list"><!-- wp:list-item --><li><span style="font-weight: 400">First list item</span></li><!-- /wp:list-item --><!-- wp:list-item --><li><span style="font-weight: 700">Second list item</span></li><!-- /wp:list-item --></ul><!-- /wp:list -->`

To handle these cases I've created custom`core/list` raw transform which extends standard raw transfom. But this is too complicated way to deal with.

PR #62662, opened 2024-06-18
Try: Extend site editor

tyxla:
## What?
This PR is a rough experiment that demonstrates how we can extend the site editor via plugins.

## Why?
At some point, plugins may want to extend the site editor - add their own route and their own settings page.

## How?
* We're making the site editor routes extensible.
* We're introducing an extensible slot for site editor sidebar menu items
* We're registering a custom route and a custom menu item (ideally, this should all be in a plugin).

This is still WIP, we still have to iron out a bunch of things. This should ideally have a better API, baked into the `registerPlugin` API for example - I'll try things out. We also need to add support for legacy plugin pages, which is its own can of worms.

## Testing Instructions
TBD

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/8436925/bf2e29ab-1a71-4799-b6cb-7f84510ac96c

### DIFF
--- a/packages/edit-site/src/components/layout/router.js
+++ b/packages/edit-site/src/components/layout/router.js
@@ -1,9 +1,13 @@
 /**
  * WordPress dependencies
  */
+import { applyFilters, addFilter } from '@wordpress/hooks';
 import { privateApis as routerPrivateApis } from '@wordpress/router';
+import { cog } from '@wordpress/icons';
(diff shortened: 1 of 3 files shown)


scruffian:
Can we close this now?

tyxla:
Ancient. Thanks for the nudge @scruffian.

PR #62663, opened 2024-06-18
Interactivity API docs: Add wp-async directives doc

cbravobernal:
## What?

Add documentation for `data-wp-async` directives.

Completes #61885

PR #62665, opened 2024-06-18
Rename yieldToMain to splitTask and export from @wordpress/interactivity

westonruter:
Follow up to https://github.com/WordPress/gutenberg/pull/61885

I just realized that the new `yieldToMain` function isn't currently exported by `@wordpress/interactivity`. This should be exported because it allows for directives to reuse this logic to break up their own long tasks when they can't make use of the `wp-async` directive (or even if they do and still are doing a lot of work). 

For example, consider this synchronous `wp-on` directive action which calls `event.preventDefault()` but then does work.

```js
import { store, splitTask } from '@wordpress/interactivity';

store( 'foo', {
	actions: {
		showSomething: function* ( event ) => {
			event.preventDefault();
			showLoadingSpinner();
			yield splitTask();
			const data = fetchData();
			processData( data );
			yield splitTask();
			renderData( data );
		}
} );
```

(Granted: A developer could just re-implement `yieldToMain` on their own.)

If approved and backported, this will need to be mentioned in the doc updates in https://github.com/WordPress/gutenberg/pull/62663, in particular under [Async Actions](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/api-reference/#:~:text=%7D\)%3B-,Async%20actions,-Async%20actions%20should).

### DIFF
--- a/docs/reference-guides/interactivity-api/api-reference.md
+++ b/docs/reference-guides/interactivity-api/api-reference.md
@@ -813,7 +813,8 @@ const { state } = store("myPlugin", {
 As mentioned above with [`wp-on`](#wp-on), [`wp-on-window`](#wp-on-window), and [`wp-on-document`](#wp-on-document), an async action should be used whenever the `async` versions of the aforementioned directives cannot be used due to the action requiring synchronous access to the `event` object. Synchronous access is reqired whenever the action needs to call `event.preventDefault()`, `event.stopPropagation()`, or `event.stopImmediatePropagation()`. To ensure that the action code does not contribute to a long task, you may manually yield to the main thread after calling the synchronous event API. For example:
 
 ```js
-function toMainThread() {
+// Note: In WordPress 6.6 this splitTask function is exported by @wordpress/interactivity.
+function splitTask() {
   return new Promise(resolve => {
     setTimeout(resolve, 0);
   });
@@ -823,7 +824,7 @@ store("myPlugin", {
   actions: {
     handleClick: function* (event) {
       event.preventDefault();
-      yield toMainThread();
+      yield splitTask();
       doTheWork();
     },
   },
--- a/packages/interactivity/src/directives.tsx
+++ b/packages/interactivity/src/directives.tsx
@@ -11,13 +11,7 @@ import { deepSignal, peek, type DeepSignal } from 'deepsignal';
 /**
  * Internal dependencies
  */
-import {
-	useWatch,
-	useInit,
-	kebabToCamelCase,
-	warn,
-	yieldToMain,
-} from './utils';
+import { useWatch, useInit, kebabToCamelCase, warn, splitTask } from './utils';
 import type { DirectiveEntry } from './hooks';
 import { directive, getScope, getEvaluate } from './hooks';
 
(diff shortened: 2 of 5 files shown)


westonruter:
Humm. In a generator context, the `yieldToMain()` name doesn't look so good:

```js
yield yieldToMain();
```

westonruter:
For now I've manually copied the `yieldToMain()` function into the Async Actions docs section in #62663.

luisherranz (approved):
Looks good to me.

> Humm. In a generator context, the `yieldToMain()` name doesn't look so good:
>
> ```js
> yield yieldToMain();
> ```

I don't think it's that bad, but feel free to rename 🙂

---

By the way, shouldn't we also update this section of the documentation? Or do you want to make those changes in another pull request?

https://github.com/WordPress/gutenberg/blob/d86ecdb1f13a683df371fc5fbb3cf84af9d1ccae/docs/reference-guides/interactivity-api/api-reference.md#L813-L831

cbravobernal:
Docs have been merged with `toMainThread`, maybe we can rename it to that.

luisherranz:
> Docs have been merged with toMainThread, maybe we can rename it to that.

Up to you, folks 🙂

westonruter:
Or simpler, what about just `mainThread`? Usage:

```js
await mainThread();
yield mainThread();
```

I like this because the preposition "to" doesn't work with "await" in English (one does not just "await to something") although "to" is used with "yield" here. By removing the preposition entirely, then this awkwardness is removed. Also, by omitting a verb from the function name itself it implies that the function doesn't actually do anything on its own. Indeed, it must be used with `await` or `yield` to have any purpose.

westonruter:
I checked with @tunetheweb and he suggested `splitTask()` instead.

docs/reference-guides/interactivity-api/api-reference.md
```diff
@@ -813,7 +813,8 @@ const { state } = store("myPlugin", {
 As mentioned above with [`wp-on`](#wp-on), [`wp-on-window`](#wp-on-window), and [`wp-on-document`](#wp-on-document), an async action should be used whenever the `async` versions of the aforementioned directives cannot be used due to the action requiring synchronous access to the `event` object. Synchronous access is reqired whenever the action needs to call `event.preventDefault()`, `event.stopPropagation()`, or `event.stopImmediatePropagation()`. To ensure that the action code does not contribute to a long task, you may manually yield to the main thread after calling the synchronous event API. For example:
 
 ```js
-function toMainThread() {
+// Note: In WordPress 6.6 this splitTask function is exported by @wordpress/interactivity.
```
westonruter, line 816: This note assumes this change will be backported.

westonruter:
@cbravobernal I'll let you merge if you are satisfied.

docs/reference-guides/interactivity-api/api-reference.md
westonruter: > By the way, shouldn't we also update this section of the documentation? Or do you want to make those changes in another pull request?

@luisherranz This comment in the docs here I believe updates the section accordingly.

cbravobernal (approved):

docs/reference-guides/interactivity-api/api-reference.md
luisherranz: Fantastic, thank you.

DAreRodz:
Nice addition. 😄 Not sure, but maybe we need to add an entry log for the `splitTask()` function, right?

luisherranz:
Oh, yes please!

westonruter:
@DAreRodz @luisherranz Like https://github.com/WordPress/gutenberg/pull/62805?

luisherranz:
Merged. Thanks Weston!

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: f37faf727c

Issue #62216, opened 2024-06-03
Editor Canvas Container: refactor and isolate to the site editor


ramonjd:
## Background

The Editor Canvas Container is a slot for the Site Editor that was intended to act as an "alternative" editor canvas to render blocks and styles independently of the main canvas container. 

It was first added in https://github.com/WordPress/gutenberg/pull/45960 and refined in https://github.com/WordPress/gutenberg/pull/49973

The important feature, at least for global style revisions, was to have a "separate" editor view that could be edited independently of the main editor canvas. So changes could be made to styles and so on in the alternative canvas view. When it closes, the user is returned to the main editor canvas in whatever state they left it.

The slot is currently used by:

- The Style Book
- Global Styles revisions

## What is your proposed solution?

Refactor the way Gutenberg displays these alternative slots.

Options could be:

- move away from slots, and rather use editor context or children props
- create a new editor UI for such alternative editor views, e.g., removing unnecessary controls and views such as the list view, inserter and so on.

See feedback:

- https://github.com/WordPress/gutenberg/pull/62146/files#r1620915555

There are a lot of unknowns. An approach needs to be confirmed.

## What problem does this address?

Since https://github.com/WordPress/gutenberg/pull/62146, the Site Editor uses the editor interface from the editor package, which means that the post editor could theoretically fill the Editor Canvas Container is a slot. Currently the only use case for the slot is to house alternative Site Editor views.

Also the way the alternative views are rendered and tracked by is cumbersome and not scalable. [Example](https://github.com/WordPress/gutenberg/blob/da98bdb025937d8bb0abc41a187792e75e17731d/packages/edit-site/src/components/global-styles/ui.js#L250-L250).

## Beyond

The potential is there for Global Styles could be "promoted" to be a top-level entity. 

So much like the editor can load a template or a page, it could "load" the site's merged global styles entity with the style book, in conjunction with the global style panel and controls, being the "interface".

## Related
- https://github.com/WordPress/gutenberg/issues/53431
- https://github.com/WordPress/gutenberg/issues/53483
- https://github.com/WordPress/gutenberg/issues/41119

ramonjd:
I did a little splashing around the shallow waters of the editor, while thinking about possible approaches. 

Here are some notes. 

They may or may not have import in relation to the main issue 😄 

### Loose requirements

- Create a canvas "view" to display global styles-specific features, e.g., style book, revisions. 
    - The view:
        - loads a given block list preview (not editable)
[...]

PR #62669, opened 2024-06-19
Display the DocumentBar for Style Book and Style Revisions

### LINKED ISSUE #51667: Add DocumentActions to Style Revisions header
Let's add the DocumentActions component to the header when style revisions are viewed. 

- [ ] Add DocumentActions to the header.
- [x] Update the "Global styles revisions" text to "Style revisions". 

CleanShot 2023-06-19 at 17 53 22

Related issue for Style Book: https://github.com/WordPress/gutenberg/issues/54514

### LINKED ISSUE #54514: Add DocumentActions to Style Book
As discussed in https://github.com/WordPress/gutenberg/issues/51667#issuecomment-1598801110, lets add the `DocumentActions` to the Style Book editor header. 

### Before: 
CleanShot 2023-09-15 at 15 45 39

### After: 
CleanShot 2023-09-15 at 15 46 03

Related issue for Style Revisions: https://github.com/WordPress/gutenberg/issues/51667

creativecoder:
## What?

When viewing Style Book or Style Revisions in the site editor, displays the DocumentBar in header rather than just the title.

Fixes https://github.com/WordPress/gutenberg/issues/51667 and https://github.com/WordPress/gutenberg/issues/54514

## Why?

Makes the header more consistent. Opening the Style Book or Style Revisions overlay now only changes the title and icon, rather than removing the DocumentBar component and replacing it with a string.

## How?

Given the [plans for refactoring and/or removing the EditorCanvasContainer component](https://github.com/WordPress/gutenberg/issues/62216), this PR attempts to make the minimal changes needed to solve the issue

- Adds `title` and `icon` props to `DocumentBar` so it can display them when an overlay is visible
- Updates the `Header` component to pass `title` and `icon` props to `DocumentBar`, rather than displaying the `title` prop in place of `DocumentBar`
- Temporarily removes the command context in the site editor when Style Book and/or Style Revisions are open, clearing the suggested commands that are specific to editing the current entity (e.g. when editing a pattern or a page)

## Testing Instructions

- Go to the site editor and open the global styles sidebar
- Click on the Style Book and Style Revisions icons at the top of the sidebar
- See that the title and icon in the header change when opening those overlays
- See that you can click on the DocumentBar and open the command menu when the overlays are open

No relevant changes

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/1699996/7083e36a-a726-471c-a464-fe4fb5673936


packages/edit-site/src/components/editor/index.js
```diff
@@ -173,6 +170,17 @@ export default function EditSiteEditor( { isLoading } ) {
...
+	// Replace the title and icon displayed in the DocumentBar when there's an overlay visible.
+	const { setEditorContentOverlayTitleAndIcon } = unlock(
+		useDispatch( editorStore )
+	);
+	const { title, icon } = getEditorCanvasContainerTitleAndIcon(
+		editorCanvasContainerView
+	);
+	useEffect( () => {
+		setEditorContentOverlayTitleAndIcon( title, icon );
+	}, [ title, icon, setEditorContentOverlayTitleAndIcon ] );
```
youknowriad: This basically replaces the "title" prop with an action API. Do you think it's a better API? Why?

If we're to keep this, I guess we should remove the "title" prop from the Editor component (and its children)

creativecoder: @youknowriad Thanks for taking a look

> This basically replaces the "title" prop with an action API. Do you think it's a better API? Why?

I did it this way because I was unsure of the backcompat restraints on the `@wordpress/editor` components, and this way keeps all API changes private.

Alternatively we could make the following modifications

- Add an `icon` prop to `Editor` (from `@wordpress/editor`) so that `EditSiteEditor` can use it to override the icon (in addition to the title override it's using now) 
- Add title and icon props to the `DocumentBar` component to override the title and icon inferred from the currently open document or template, so that we can display the title and icon overrides from `EditSiteEditor`  inside the DocumentBar when an overlay is visible (rather than just replacing the component with a string, like we do now)

Given the [plans to refactor EditorCanvasContainer](https://github.com/WordPress/gutenberg/issues/62216) for Style Book and Style Revisions I'm hesitant to add new props to `@wordpress/editor` components just for the sake of the site editor. On the other hand, those implementing custom editors might appreciate the title/icon override props, so it could make sense to keep them around. What do you think?

youknowriad: To be honest, neither solution satisfy me completely :P 

 - I like the simplicity of the props but don't like the fact that they need to be passed down to the components.
 - I don't like the complexity of the actions/selectors but they don't need to be passed down to components.

Worth nothing that the components we're talking about are also private AFAIK. So maybe for now, I have a small preference towards just adding an icon prop.

creativecoder: > To be honest, neither solution satisfy me completely ... So maybe for now, I have a small preference towards just adding an icon prop.

I feel similarly! If nothing else, using props is a little less code overall. I'll rework the PR to use props and we can see how that feels.

creativecoder: By the way, `DocumentBar` is exported from `@wordpress/editor`, so adding props to the component does change the public API of the package.

packages/edit-site/src/components/editor/index.js
```diff
@@ -199,11 +200,8 @@ export default function EditSiteEditor( { isLoading } ) {
 						_isPreviewingTheme && <SaveButton size="compact" />
 					}
 					forceDisableBlockTools={ ! hasDefaultEditorCanvasView }
-					title={
-						! hasDefaultEditorCanvasView
-							? getEditorCanvasContainerTitle( editorCanvasView )
-							: undefined
-					}
+					title={ title }
+					icon={ icon }
```
creativecoder, line 204: Using the `! hasDefaultEditorCanvasView` check here causes the `DocumentBar` title/icon to flash back and forth between the document and overlay titles when activating the Style Book in addition to Style Revisions and vice versa.

Since the default title and icon are empty strings and the return value of `getEditorCanvasContainerTitleAndIcon` is directly tied the canvas container state, the check isn't really needed.

packages/editor/src/components/commands/index.js
```diff
@@ -299,9 +300,10 @@ function useEditedEntityContextualCommands() {
 		};
 	}, [] );
 	const { openModal } = useDispatch( interfaceStore );
+	const hasEditorContentOverlay = useHasEditorContentOverlay();
 	const commands = [];
 
-	if ( postType === PATTERN_POST_TYPE ) {
+	if ( ! hasEditorContentOverlay && postType === PATTERN_POST_TYPE ) {
```
ramonjd: When editing a page, there's an Edit Template command as well:

Screenshot 2024-06-20 at 12 51 00 PM

Do you think that should also be hidden?

ramonjd (commented):
This is a great enhancement and makes the header more consistent, thanks! 

I tested switching between overlays and the heading is accurate at each step:

https://github.com/WordPress/gutenberg/assets/6458278/e13bb615-9f6d-4785-a01e-d0eb339b965a

Left a question about hiding other controls.

packages/editor/src/components/commands/index.js
youknowriad: I'm not sure I understand the change here to be honest?

ramonjd: Yeah, might need a comment. 

When editing a pattern, and then activating the style book, this condition ensures the default pattern commands aren't shown by default when the command centre is opened:

Screenshot 2024-06-20 at 4 39 53 PM

youknowriad: Is this the right fix? Maybe when going into "style book" and "style revision", we should be changing the "context" instead. (See `useSetCommandContext`) Again, I think this is just a side effect of us not actually changing the active "entity" in the EditorProvider to the global styles entity, but that is definitely a bigger change.

creativecoder: Good points! I've updated to have the `edit-site` package remove the command context temporarily when Style Book and/or Style Revisions are open, clearing the suggested commands that are specific to editing the current entity (for both patterns and page): https://github.com/WordPress/gutenberg/pull/62669/files#diff-caaab4735e981969da0fb0108be1c6a34e2f039940615a695fa05e061dac42d9R40

> Again, I think this is just a side effect of us not actually changing the active "entity" in the EditorProvider to the global styles entity, but that is definitely a bigger change.

Agreed, I think this is another example supporting how opening a global styles entity would make more sense. But hopefully this workaround is acceptable for now, until we've made that much bigger change.

packages/edit-site/src/hooks/commands/use-set-command-context.js
```diff
@@ -33,5 +37,13 @@ export default function useSetCommandContext() {
 	if ( hasBlockSelected ) {
 		commandContext = 'block-selection-edit';
 	}
+	if ( hasEditorCanvasContainer ) {
+		/*
+		 * The editor canvas overlay will likely be deprecated in the future, so for now we clear the command context
+		 * to remove the suggested commands that may not make sense with Style Book or Style Revisions open.
+		 * See https://github.com/WordPress/gutenberg/issues/62216.
```
ramonjd, line 44: So, in the future we'd have something like `commandContext = getCurrentSiteEditorEntity()` or something?

Anyhoo, I'm going to try to shuffle some things around and start looking at that issue after 6.6. When the time comes, it'd be great to your feedback on the possible approaches 🙇🏻

ramonjd (approved):
LGTM

Thanks again.

Behold, my GIF!

packages/edit-site/src/hooks/commands/use-set-command-context.js
creativecoder: > So, in the future we'd have something like commandContext = getCurrentSiteEditorEntity() or something?

I don't think a new command context would be needed. Presumably if opening Style Book was changing the currently edited entity to a `wp_global_styles` post, then the check here could just be removed. The command context would go back to being `entity-edit` with the `wp_global_styles` post type returned by `getCurrentPostType` from the editor store, which should cover any checks needed for adding suggested commands.

For example, in the case of the [suggested pattern commands I was manually checking before](https://github.com/WordPress/gutenberg/blob/2b87be1903e4d509037d5bf2c47c2ef07e9526c2/packages/editor/src/components/commands/index.js#L306), the post type is no longer `wp_pattern`, so those would be hidden when Style Book is open without any additional changes.

creativecoder: > When the time comes, it'd be great to your feedback on the possible approaches 🙇🏻

Yes, for sure! Putting together this PR was a great crash course on understanding the current relationship between the `editor` and `edit-site` packages--I'd be glad to put that knowledge to more use.

creativecoder:
I'm not sure what's going on with CI: [phpunit tests look green](https://github.com/WordPress/gutenberg/actions/runs/9654465123/job/26628753216?pr=62669) but it's not coming through here in the PR.

ramonjd:
@creativecoder Tests have been failing frequently lately. If you're confident the fails are not related — which has been the case in the last week for many committed PRs — then I think it should be fine. Especially if this PR isn't to be backported to the 6.6 RC.

creativecoder:
Thanks @ramonjd ! In this case everything was passing, but the test status wasn't coming through to the PR. Looks like it's working again now. I'll run everything again, just to be sure, then I'll get this merged.

Mamaduka:
Thank you, @creativecoder!

creativecoder:
Leaving myself some notes, for when I come back to any related issues:

The DocumentBar component in the header, used to show the icon/title and trigger the command center modal, lives in a component hierarchy like this

- `EditorSiteEditor` (`@wordpress/edit-site` package, renders `Editor` from `@wordpress/editor`)
  - `EditorCanvasContainer` (uses `EditorContentSlotFill.Fill`, passes children into the fill)
  - `StyleBook` (wraps UI in `EditorCanvasContainer` to display Style Book)
  - `Revisions` (wraps UI in `EditorCanvasContainer` to display Style Revisions)
- `Editor` (`@wordpress/editor` package, private component)
  - `EditorInterface` (provides `EditorContentSlotFill.Slot`)
    - `Header` (`forceDisableBlockTools` prop to partially disable toolbar in the header)
      - `DocumentBar`

The `EditorCanvasContainer` is an `@wordpress/edit-site` component used as a fill for the `EditorContentSlotFill`, a private slotfill provided by `@wordpress/editor`. The `EditorCanvasContainer` creates an overlay that fills the main content area when Style Revisions and/or Style Book are open.

DocumentBar usually shows the title and icon from the entity currently being edited (post, template, etc), and now has `title` and `icon` props that override those values. The `title` and `icon` props and are passed down through the component hierarchy and populated by `EditorSiteEditor` when the overlay is visible.

The `forceDisableBlockTools` prop is also passed down by `EditorSiteEditor` to `Header` to partially disable the toolbar when the overlay is visible.

In the future (see https://github.com/WordPress/gutenberg/issues/62216), Style Book and Style Revisions will likely open the current `wp_global_styles` post and its revisions as the entity being edited, so the title/icon override props shouldn't be needed for these views, as they will be inferred like they are now for other entities. We would also need an alternative signal for `forceDisableBlockTools` when the overlay is removed.

PR #62670, opened 2024-06-19
Allow children of alignfull flow layouts to have root padding.

### LINKED ISSUE #62597: Global padding: ".wp-block-block" class references in CSS selectors force padding to 0 in nested Group blocks
### Description

WordPress 6.6 and Gutenberg plugin in latest version have global padding CSS selectors including references to a `.wp-block-block` class, in `.has-global-padding :where(.has-global-padding:not(.wp-block-block, .alignfull, .alignwide))` for example.

I don't think `.wp-block-block` is a class used either in the editor or in the front-end. Wouldn't that be a mistake? It seems to me that it should refer to the `.wp-block-group` class instead.

As a result of what appears to be an error, nested "Group" blocks have their internal margins reset to 0.

I detail the procedure to reproduce the bug below.

1. Create a page on a WP with 6.6 Beta 2 and Twenty Twenty-Four block theme.
2. Add a Group in full-width with no "Inner blocks use content width" option.
3. Add another Group inside the first one with "Inner blocks use content width" option.
4. Add paragraph or content in it.
5. Save and view the page, you'll see the inline style applied to the inner group with wrong selector. Left and right padding values are reset to 0 when they shouldn't be. It's the global padding values that apply and should not be reset.

_No response_

### Environment info

- WordPress 6.6 Beta 2
- Gutenberg latest

Yes

Yes

tellthemachines:
## What?

Fixes #62597.

When a container with inner content width is nested inside another container that is both full width and flow type layout (without content width), the inner container should still receive root padding. 

## Testing Instructions

<details>
<summary>Test markup</summary>

```
<!-- wp:group {"align":"full","style":{"border":{"width":"3px","style":"dashed"}},"borderColor":"accent","layout":{"type":"default"}} -->
<div class="wp-block-group alignfull has-border-color has-accent-border-color" style="border-style:dashed;border-width:3px"><!-- wp:paragraph -->
<p>group block full width no content width</p>
<!-- /wp:paragraph -->

<!-- wp:group {"style":{"border":{"style":"dashed","width":"2px"}},"borderColor":"accent-2","layout":{"type":"constrained"}} -->
<div class="wp-block-group has-border-color has-accent-2-border-color" style="border-style:dashed;border-width:2px"><!-- wp:paragraph -->
<p>group block with content width</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```
</details>

1. Test with a block theme that has `useRootPaddingAwareAlignments` enabled in its theme.json settings.
2. Add the markup to a post, save and publish.
3. Check that both in editor and front end, the inner Group block has root padding applied.

## Screenshots or screencast 

Before:

Screenshot 2024-06-19 at 2 28 56 PM

After:

Screenshot 2024-06-19 at 2 29 45 PM


tellthemachines:
I've tested this with all the combinations of markup I could think of, but would be great to have a confidence check that this is viable as a fix before going ahead with the trac ticket and core PR rigmarole 😅

aaronrobertshaw (commented):
This is testing pretty well for me @tellthemachines 👍 

I really appreciate the test block markup, screenshots etc. Made it a lot easier to make sure I was testing the right things as I'm no layout expert 😅 

✅ Could replicate the original issue
✅ Site editor correctly displayed padding and was consistent with frontend
✅ Post editor applied padding and was consistent with frontend
✅ Padding applied consistently across TT4, TT3, Assembler, and Powder themes
✅ Didn't spot any differences in themes without `useRootPaddingAwareAlignments` 
✅ Multiple levels of nested group blocks with and without content width work as expected

I think this is ready for a couple of unit test updates and the backport.  It looks like we only have coverage of what generated for these styles in the PHP theme.json tests. The `useGlobalStylesOutput` hook tests passed.

Screenshot 2024-06-19 at 3 28 58 PM

tellthemachines:
Thanks for the review @aaronrobertshaw ! Tests are updated and sync PR is up: https://github.com/WordPress/wordpress-develop/pull/6854

aaronrobertshaw:
> Tests are updated and sync PR is up

🚀 Appreciate the quick turnaround, the PHP unit tests are now passing 

If you're not a fan of updating the [use global styles hook's tests](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/global-styles/test/use-global-styles-output.js) because it's just testing style generation, perhaps a better option would be to add an e2e like [those added](https://github.com/WordPress/gutenberg/pull/61638/files#diff-5a56cf40ab8147fc3913bfdd085bdfced46d8a89f79777ede550772704f356f6) in the [specificity changes PR](https://github.com/WordPress/gutenberg/pull/61638), to guard against regressions here?

tellthemachines:
> perhaps a better option would be to add an e2e like [those added](https://github.com/WordPress/gutenberg/pull/61638/files#diff-5a56cf40ab8147fc3913bfdd085bdfced46d8a89f79777ede550772704f356f6) in the [specificity changes PR](https://github.com/WordPress/gutenberg/pull/61638), to guard against regressions here?

These styles are pretty unstable - they've been updated many times since initial implementation - so I'm reluctant to add more tests for them as it means more maintenance overhead for little benefit. 

A large part of the problem is there has never been agreement on what precisely `useRootPaddingAwareAlignments` should do, so it's unclear what we should be testing for. The bug reports that come in for these styles usually result from specific combinations of blocks and attributes that we'd almost certainly not catch by setting up some e2e with a bunch of test markup covering the common cases (such as those described in the PHP comments). Maybe if at some point we go for one or two release cycles without making a change to these styles, we can consider them stable enough to invest in tests 😅 but for now I think it's more trouble than it's worth.

andrewserong (approved):
Nice work, this is testing well for me, too! It's a tiny bit unfortunate that we need to add more complexity to the selectors, but it's a good issue to solve 👍

> Maybe if at some point we go for one or two release cycles without making a change to these styles, we can consider them stable enough to invest in tests 😅 but for now I think it's more trouble than it's worth.

In principle, I agree with the idea of trying to ensure we've got good test coverage of the styling rules we're outputting. In practice for this feature I agree it's likely more trouble than it's worth as we'd just be string matching for now.

Arguably, there could be a case to be made to ensure that what's output from `use-global-styles-output.js` matches the PHP version to ensure no-one makes any typos, but at this stage, I think it'd be good to treat how we reliably test the root padding styles as a separate exploration from this fix. The important thing IMO is that multiple folks have taken the PR for a manual spin to test with block markup in a variety of situations and themes.

LGTM! 🚀 

I've added the `Backport to WP Beta/RC` label since there's a JS change in here. Please update the labels if this isn't right, though 🙂

tellthemachines:
> Arguably, there could be a case to be made to ensure that what's output from `use-global-styles-output.js` matches the PHP version to ensure no-one makes any typos

Ideally there wouldn't even be this duplication in the code! But yes it's something to look at later 😅

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 510684e007

PR #62671, opened 2024-06-19
Update link to API section in block-editor README.md

amitraj2203:
## What?
Change the link to the correct one.

## Why?
The Block Editor package reference docs have a link for the API section which doesn't work, probably because of the capitalized link.
Visit [here](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-block-editor/#api:~:text=(See%20the%20API%20for%20the%20list%20of%20all%20the%20available%20components).) and try to click on the API link.


jorgefilipecosta (approved):
LGTM 👍

Issue #57813, opened 2024-01-12
Reconsider the usage of the WP logo / Site Icon in the editor UI


afercia:
### Description

This issue doesn't aim to focus on the behavior and functionality of the WP logo / Site icon link and button used in the editors. Rather, it aims to only focus on the visual metrics and meaning of the images used for those functionalities.

Both the Post editor and the Site editor use the WP logo (or the Site icon if set) for the link to exit the editor / button to toggle the navigation panel. There's a few visual inconsistencies across the two editors. Alos, I'm not sure the _meaning_ of the images is the most appropriate for a link and button that basically mean 'go back to the classic admin' or 'toggle the navigation panel'.

Visual inconsistencies contribute to make the user interface be perceived as unpolished and confusing. The visual meaning of the images doesn't really represent the underlying functionality. Both things are confusing for users and add cognitive load. As such, I'm marking this issue with the Accessibility label.

- The images default sizes are different between the two editors. I'm not sure what is the reasoning behind this but there doesn't seem to be a good reason to introduces such an inconsistency.
- When setting a custom site icon, that will be used in place of the WP logo. Depending on the image aspect ratio, the visual difference is even more noticeable.
- The transition on hover is different. In the Post editor, the image scales _up_. Instead, in the Site editor the image scales _down_. I'm not sure whether this is intentional but if a transition effect is really necessary, I'd say it should be consistent. Personally, I'm not sure the transition adss any value for users and I'd rather remove it entirely.
- The meaning of hte images doesn't seem to be the most appropriate for the functionality.
  - The WP logo: Historically, the WP logo has been used in the admin menu for the link to the About page (and its related dropdown menu). While it's probably the less used link in the admin interface, it's very prominent. It's there in the left top corner of the user interface: all users are familiar with it and they learned it means 'WordPress About page' or links to wp.org pages. Instead, in the editor the WP is supposed to have another meaning, which is unexpeced and confusing for users. To me, the WP logo certainly doesn't mean 'go back'.
  - The Site icon: as a used, when I see the Site icon I mentally associate it to the front end. To me, it means 'go to the front end'. I'm not sure I understand what the added value is for users by showingt the Site icon in the editor user interface, for controls that have a different meaning than 'go to the front end'.

## Visual inconsistencies details

Post editor:
- Image default size: 36 by 36 pixels.
- If the image is larger, it is constrained to 36 by 36 pixels anyways.
[...]

PR #62675, opened 2024-06-19
fix: wp icon focus issue

### LINKED ISSUE #62672: Site editor: WP logo focus style regression
### Description

The focus style of the WP logo in the Site editor is now barely visible and not standard.

This is the focus style on WordPress 6.5:

wp 6 5

This is the focus style on Gutenberg trunk:

trunk

The WP logo uses a custom implementation of the focus style. I _think_ the regression comes from the fact the `transform scale` CSS is now applied to the link element itself while previously it was applied to the inner div element.

- Go to the Site editor.
- Tab to the WP logo.
- Observe the focus style is barely visible and not standard.

_No response_

### Environment info

_No response_

Yes

Yes

up1512001:
Fixes #62672

## What?
- fix the focus issue with the wp icon in the editor.

## Why?
#62672

## How?
- moved the focus from before to the `a` tag itself.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/75293077/3e2c9f4e-43f3-4dda-921e-d85be2120c0a


bgardner:
Confirming this works as expected!

talldan:
@up1512001 Thanks for another contribution! Something that's helpful when creating PRs is to type 'Fixes #62672' in the PR description, that links the PR to the issue it solves, sets it in progress and will automatically close the issue when the PR's merged:
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

I've gone ahead and added it to this PR's description.

talldan:
It does look better than `trunk`:
Screenshot 2024-06-24 at 3 19 09 pm

But the focus style in this PR still looks too thin compared to a normal focus style:
Screenshot 2024-06-24 at 3 18 12 pm

I think because of the css `scale` that's applied to the anchor element as mentioned in the issue.

Maybe the `scale` implemented here:
https://github.com/WordPress/gutenberg/blob/58932edaccdf7148f1fae42dddfec4099e007e87/packages/edit-site/src/components/site-hub/index.js#L66

Could be implemented on a div around the `<SiteIcon>` instead? Or applied tothe `SiteIcon` itself. I imagine there's also a `transition` style somewhere that would have to be updated too.

I've added @youknowriad as a reviewer, as the change seems to have come from #61579.

afercia:
> It does look better than trunk:

Yes it deoes look better but i'm not sure it fully solves the issue. Ideally, the focus style should be the same blue outline used for all buttons. 

I believe this issue comes from https://github.com/WordPress/gutenberg/pull/61579 Cc @youknowriad 
While I understand the need for simplification on that PR, the CSS 'scale' should not be applied to the button itself because it impacts also the focus style. That said, I would eveb be in favor of entirely removing the animation because I'm not sure it adds great value. Also, as mentioned in that PR the `view-transition` CSS API is only fully supported in Chrome and I'm not sure WordPress should use unsupported features not even as progressive ehnancement.

youknowriad:
Can we try to check that this doesn't impact the focus styles ... on the full screen editor. These styles might be shared between both and the expectations different. So I just want to make sure we don't regress elsewhere.

up1512001:
I have updated the shadow and it's looking like this please let me know if it's ok.
cc: @youknowriad @afercia 

https://github.com/WordPress/gutenberg/assets/75293077/4140417c-ccb7-454b-ad3f-fdf3570734fd

youknowriad:
Screenshot 2024-06-25 at 10 39 01 AM

The focus styles in the full editor looks broken to me. (See top left)

afercia:
It is worth reminding that the whole mechanism of the WP logo / site icon is being discussed in https://github.com/WordPress/gutenberg/issues/57813.

To me, the current implementation especially in the Site Editor is unnecessarily complicated and it's mainly done this way only to add an animation that doesn't add great value. Also, as mentioned in https://github.com/WordPress/gutenberg/issues/57813, when a site icon is set, this link is used to 'preview' the site icon which, in my opinion, is just confusing. Given the complexity of this implementation, even a simple thing like making sure there is correct focus style becomes complicated. I'd rather see a drastic simplification:
- remove the animaiton
- remove the site icon preview

up1512001:
@afercia I guess then we can close this PR as https://github.com/WordPress/gutenberg/issues/57813 already has PR which is merged.

up1512001:
> The focus styles in the full editor looks broken to me. (See top left)

@youknowriad I have added Fix for this can you please verify it?

youknowriad:
Maybe I'm not testing properly but I think the focus style is meant to be like that in the editor

Screenshot 2024-07-15 at 10 20 21 AM

In my testing, the blue outline is not visible at the moment in editor mode.

up1512001:
@youknowriad This is how it's looking now.


https://github.com/user-attachments/assets/797624e1-568c-482f-a655-fc58868b3d22

youknowriad (approved):
LGTM 

I'd appreciate a second review. cc @jameskoster @jasmussen

jasmussen:
Seems to fix the regression 👍 👍

ellatrix:
@talldan is this really important to land in a minor release?

talldan:
@ellatrix I agree it is minor, but also seems low risk and affects quite a prominent part of the site editor.

I don't really mind, feel free to remove the label.

ellatrix:
It's fine to leave it there's high confidence.

ellatrix:
Manually backported in 9c341965cccd929e893cb4b0fb7a71476e45321b

PR #62676, opened 2024-06-19
Site Editor: Improve the frame animation

youknowriad:
Alternative to #62386

## What?

@stokesman raised that the changes in #61579 introduce a regression in the frame animation where during the frame animation, sometimes the header appears separate from the frame.

It looks like it happens only on browsers that support view transitions. My understand of the issue is that we have some things that are using view transitions but others framer motion animation which caused the conflict.

## How?

This PR solves the issue by actually removing the view transition name from the "frame" which was not really needed AFAIK

## Testing Instructions

1- Open the site editor
2- Navigate to "edit" mode 
3- Keep your eyes focused on the header area during the animation and you shouldn't feel like the header got "disconnected" from the frame (hard to describe) :P

### DIFF
--- a/packages/edit-site/src/components/block-editor/style.scss
+++ b/packages/edit-site/src/components/block-editor/style.scss
@@ -1,15 +1,4 @@
-/* stylelint-disable -- Disable reason: View Transitions not supported properly by stylelint. */
-::view-transition-old(frame),
-::view-transition-new(frame) {
-  animation-duration: 0;
-}
-/* stylelint-enable */
-
 .edit-site-visual-editor__editor-canvas {
-	/* stylelint-disable -- Disable reason: View Transitions not supported properly by stylelint. */
-	view-transition-name: frame;
-	/* stylelint-enable */
-
 	&.is-focused {
 		outline: calc(2 * var(--wp-admin-border-width-focus)) solid var(--wp-admin-theme-color);
 		outline-offset: calc(-2 * var(--wp-admin-border-width-focus));
--- a/packages/edit-site/src/style.scss
+++ b/packages/edit-site/src/style.scss
@@ -36,6 +36,19 @@
 @import "./components/pagination/style.scss";
 @import "./components/global-styles/variations/style.scss";
 
+/* stylelint-disable -- Disable reason: View Transitions not supported properly by stylelint. */
+::view-transition-image-pair(root) {
+	isolation: auto;
+}
+
+::view-transition-old(root),
+::view-transition-new(root) {
+	animation: none;
+	mix-blend-mode: normal;
+	display: block;
+}
+/* stylelint-enable */
+
 body.js #wpadminbar {
 	display: none;
 }


jasmussen (approved):
This is very curious, but as best I can tell, this works for me. I'd love for @stokesman to also take a look, but this fixes the glitch that can blink-and-you'll-miss-it appear when going back from the editor.

stokesman (approved):
Here’s what I see:

https://github.com/WordPress/gutenberg/assets/9000376/18f07a51-e4c0-4a14-a77e-58f189c5bb90



This improves what it’s supposed to and what I think was the most obvious issue with the animation so ✅. It also seems to fix the transition of the canvas border-radius (which I guess was still happening but somehow being obscured and appearing to snap).

Still, I don’t want us to overlook that this increases the number of elements that crossfade during the transition and while its not too unsightly, it’s a not a intentional choice for the animation either.

Other than that I'm noticing the delay for the site icon is when entering edit mode is too short to prevent it from moving a bit off the left edge and back again with the canvas. It can be seen in the second half of the video above. It’s probably quite performance dependent on the canvas moving animation. If that lags it probably increases the amount of delay required to keep the site icon from moving horizontally.

And the last thing I’ll point out is the—yes, blink and you’ll miss it—editor header layout shift at the start of the transition to view mode. The back button instantly disappears so the inserter and document tools shift left. This makes the inserter appear to come out from behind the site icon. Ideally, they shouldn’t shift and start from where they were. Here’s a still from the video above to help show what I'm referring to:
image

I don’t see this as qualifying as an alternative to #62386 but by all means merge 👍

youknowriad:
>  this increases the number of elements that crossfade during the transition

Ah right, this is probably the main reason for why I added that code in the first place. I did remember something about cross fade but when testing I didn't notice it. I'll take a deeper look. @jasmussen do you think the cross fade is a bigger issue that what this PR is solving?

youknowriad:
> Other than that I'm noticing the delay for the site icon is when entering edit mode is too short to prevent it from moving a bit off the left edge and back again with the canvas. It can be seen in the second half of the video above. It’s probably quite performance dependent on the canvas moving animation. If that lags it probably increases the amount of delay required to keep the site icon from moving horizontally.

Yes, that delay is only there because the main frame animation is not using view transitions, when I experimented with using view transitions for it, the delay was not needed. The problem is I didn't want to use view transitions for it yet because of the browser support. While the lack of the logo animation in Safari and Firefox for now is acceptable, I'm not sure the lack of frame animation is.

youknowriad:
I think I found a way to disable the cross-fade entirely.

For the header shift, I'm not sure yet how to solve it.

stokesman:
> I think I found a way to disable the cross-fade entirely.

Nice! That really cleans up the transitions.

> For the header shift, I'm not sure yet how to solve it.

It just occurred to me, the changes in #62636 happen to fix that. I’d certainly not say the shift is a blocker here.

youknowriad:
Ahh forgot to apply the props properly. Sorry folks.

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: 4cf1c8c585

PR #62677, opened 2024-06-19
Limit pattern shuffling to theme and user patterns only

### LINKED ISSUE #60135: Limit pattern shuffling to theme and user patterns only
While exploring [shuffling](https://github.com/WordPress/gutenberg/issues/45552), we're making strides towards shuffling to relevant patterns, including https://github.com/WordPress/gutenberg/pull/60088 and https://github.com/WordPress/gutenberg/pull/60074. 

Next, I'd like to try shuffling only using the theme and user patterns, not core patterns. 

I do think that we can open it back up to core patterns eventually, when they're more suitable for the theme to style. But for now, omitting those will help solidify shuffling with less variables.

MaggieCabrera:
## What?

Closes https://github.com/WordPress/gutenberg/issues/60135

This limits shuffling to only patterns added by the theme or the user

## Why?

Core patterns or pattern directory patterns may not fit the design of the theme. By excluding them we make the shuffling more consistent with the theme's design

## How?

By checking the source of the pattern. The condition will exclude all patterns from the pattern directory (testing on TT4 I could see patterns with the source `pattern-directory/core` and `pattern-directory/featured`), all core patterns (such as the social icons one) and keep `pattern-directory/theme` patterns (TT4 includes two of those)

## Testing Instructions

- With TT4 installed, open the pattern inserter
- Select Featured > Hero
- With the new pattern selected on canvas, click shuffle. It shouldn't show core patterns that are not theme or user patterns

### DIFF
--- a/packages/block-editor/src/components/block-toolbar/shuffle.js
+++ b/packages/block-editor/src/components/block-toolbar/shuffle.js
@@ -56,7 +56,8 @@ export default function Shuffle( { clientId, as = Container } ) {
 		return patterns.filter( ( pattern ) => {
 			const isCorePattern =
 				pattern.source === 'core' ||
-				( pattern.source?.startsWith( 'pattern-directory' ) && pattern.source !== 'pattern-directory/theme' );
+				( pattern.source?.startsWith( 'pattern-directory' ) &&
+					pattern.source !== 'pattern-directory/theme' );
 			return (
 				// Check if the pattern has only one top level block,
 				// otherwise we may shuffle to pattern that will not allow to continue shuffling.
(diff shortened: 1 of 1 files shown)


packages/block-editor/src/components/block-toolbar/shuffle.js
```diff
@@ -54,10 +54,17 @@ export default function Shuffle( { clientId, as = Container } ) {
 			return EMPTY_ARRAY;
 		}
 		return patterns.filter( ( pattern ) => {
+			const isCorePattern =
+				pattern.source === 'core' ||
+				( pattern.source !== 'pattern-directory/theme' &&
+					pattern.source?.startsWith( 'pattern-directory' ) ===
+						true );
```
scruffian, line 61:
```suggestion
					pattern.source?.startsWith( 'pattern-directory' ) );
```

scruffian:
```suggestion
				( pattern.source?.startsWith( 'pattern-directory' ) && pattern.source !== 'pattern-directory/theme' );
```

scruffian: I find this easier to read personally

scruffian (approved):
Works as described. I'm not sure if this is a better experience though IMO :)

PR #62681, opened 2024-06-20
Editor: make VisualEditor a stacking context

### LINKED ISSUE #62639: Block toolbar removes bottom shadow of header
### Description

Reported in this comment: https://github.com/WordPress/gutenberg/pull/61835#issuecomment-2174482262

With any block selected, if you scroll the canvas up and part of the block toolbar overlaps with the header, the header shadow disappears.

I used `git bisect` to investigate the issue, and it looks like this issue first appeared in #61835.

- Disable the Top Toolbar.
- Click any block.
- Scroll down the canvas.
- When the block is visible in the canvas, the block toolbar will be docked below the header.
- Scroll the canvas some more.
- As the block becomes invisible, the block toolbar will overlap the header.
- The shadow below the overlap should disappear.

https://github.com/WordPress/gutenberg/assets/54422211/03fa7533-1023-4150-8206-f3246d759825

### Environment info

_No response_

Yes

Yes

stokesman:
## What?
A change to `VisualEditor` so it contains z-indexes of children within it (as a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context))

## Why?
To fix #62639, in which it’s noted that the block toolbar is not actually underneath the editor header and so the header’s shadow disappears behind it.

## How?
Add `isolation: isolate` to the component’s root element.

## Testing Instructions
1. Disable the Top Toolbar.
2. Click any block.
3. Scroll down the canvas.
4. When the block is visible in the canvas, the block toolbar will be docked below the header.
5. Scroll the canvas some more.
6. As the block becomes invisible, the block toolbar will overlap the header.
7. The header shadow should shade the block toolbar.

### DIFF
--- a/packages/editor/src/components/visual-editor/style.scss
+++ b/packages/editor/src/components/visual-editor/style.scss
@@ -3,6 +3,8 @@
 	height: 100%;
 	display: block;
 	background-color: $gray-300;
+	// Make this a stacking context to contain the z-index of children elements.
+	isolation: isolate;
 
 	// Centralize the editor horizontally (flex-direction is column).
 	align-items: center;


t-hamano (commented):
This PR is working as expected. Also, as far as I tested, there are no other issues caused by this PR.

### Before

image

### After

image

However, the next release this PR will be backported is RC version, and to be on the safe side, I would like to ask for a second review by people who are knowledgeable in this area 🙏

@youknowriad @draganescu @jeryj

packages/editor/src/components/visual-editor/style.scss
```diff
@@ -3,6 +3,8 @@
 	height: 100%;
 	display: block;
 	background-color: $gray-300;
+	// Make this a stacking context to contain the z-index of children elements.
+	isolation: isolate;
```
youknowriad, line 7: 😬 I just learned about this

youknowriad:
Seems ok to me but I wonder what impact this change can have. (other z-index related impact)

t-hamano:
Personally, I think this issue is less serious. Would it be better to test it only on Gutenberg for now and not backport it to WP 6.6?

packages/editor/src/components/visual-editor/style.scss
ellatrix: Same 😅

stokesman:
> Seems ok to me but I wonder what impact this change can have. (other z-index related impact)

I’m all for making haste slowly. The property could have been on the `__content` or `__body` elements of the interface skeleton but putting it deeper in the tree seemed more conservative. This should only affect z-indexes within. Here are the ones I've found and tested:
- block toolbar
- in-between inserter
- spacing visualizer
- grid visualizer

They are all probably expected to be no higher than the visual editor itself and seem to work as expected. Any popover launched from those is out in the fallback container as are any other popovers I've found in the interface so can’t be affected by this change.

packages/editor/src/components/visual-editor/style.scss
stokesman: > 😬 I just learned about this

And yet you [used the property here](https://github.com/WordPress/gutenberg/pull/62676/commits/2212dbfefaff02c937156cd09db904676c512e49#diff-5065a00f747fc9cdddcc2220e19e8e56df99e52f99cc25e1a4e51a21daf0897aR41) 😆

youknowriad: I know I know, that's where I saw it first. I copied it form the View Transition Guide (by Google) to reset cross-shadow transitions :P

t-hamano (approved):

t-hamano:
@stokesman I think it would be better to merge this PR without the backport label just to be safe. What do you think?

stokesman:
It’s fine by me. Is there a chance to still fix the issue in 6.6? If not by backporting this later then perhaps a PR which fixes it by changing z-index? I know the issue isn’t critical but it’d be nice have it done.

t-hamano:
Pinging @ellatrix @vcanales  - I would be happy to decide whether to backport this PR to WP6.6.

ellatrix:
Sure it's fine, but needs to be merged asap, we're cherrypicking rn

t-hamano:
It seemed like GitHub Actions had stopped, so I rebased it to get it running again.

t-hamano:
@stokesman Maybe this PR caused an issue with popovers. Is there a way to fix this?

Please see #62998

PR #62683, opened 2024-06-20
Performance: block style variations: only calc styles for present inner blocks

ellatrix:
## What?

Currently we generate styles for _all_ block types when a block has a variation. This PR attempts to limit generating styles only for blocks types that are nested within the block.

## Why?

Performance.

## Testing Instructions

See things to test in https://github.com/WordPress/gutenberg/pull/57908, everything should work correctly.

### DIFF
--- a/packages/block-editor/src/hooks/block-style-variation.js
+++ b/packages/block-editor/src/hooks/block-style-variation.js
@@ -118,17 +118,42 @@ function useBlockProps( { name, className, clientId } ) {
 		clientId
 	);
 
+	const descendantBlockNames = useSelect(
+		( select ) => {
+			// Avoid a subscription if there is no variation.
+			if ( ! variation ) {
+				return;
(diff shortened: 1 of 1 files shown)


aaronrobertshaw:
Closing this as stale. 

The code in question moved on to `packages/block-editor/src/hooks/block-style-variation.jsx` and `getBlockSelectors` has a different signature now. It might need a closer look rather than a rebase.

For anyone else picking the idea up, the saving seems to be a little smaller than it looks because `toStyles` only generates CSS for what is in the config tree and that is one block with one variation. The full block type list only costs us building the selector map. The proposed `useSelect` runs on every block editor store update and walks the block's descendants each time, so it trades a one off cost at mount for a repeated one while typing.

Maybe we could cache `getBlockSelectors( getBlockTypes() )` against registered block types and styles instead?

PR #62684, opened 2024-06-20
Performance: reduce block support hook calls further with isMatch

ellatrix:
## What?

Currently we only filter block support hook calls when it doesn't match the top level attribute name. The problem is that these attributes have evolved to nested `style` properties, which causes these all of these hooks to match as soon as a block has a style attribute, even if the styles are not relevant.

## Why?

Performance.

Site editor load patterns: -12.44%, -6.2%, 2.9%, -4.67%
Navigate: -2.47%, -3.39%, -10.28%, -10.87%

## How?

There's already an `isMatch` function so block support hooks can fine-tune which blocks should match.

### DIFF
--- a/packages/block-editor/src/hooks/background.js
+++ b/packages/block-editor/src/hooks/background.js
@@ -79,14 +79,7 @@ export function setBackgroundStyleDefaults( backgroundStyle ) {
 	return backgroundStylesWithDefaults;
 }
 
-function useBlockProps( { name, style } ) {
-	if (
-		! hasBackgroundSupport( name ) ||
-		! style?.background?.backgroundImage
-	) {
-		return;
-	}
-
+function useBlockProps( { style } ) {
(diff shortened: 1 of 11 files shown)


vcanales:
> Performance.
> 
> Site editor load patterns: -12.44%, -6.2%, 2.9%, -4.67%
Navigate: -2.47%, -3.39%, -10.28%, -10.87%

I'd appreciate guidance on how to verify these tests, both to check the changes here and to catch performance issues elsewhere as well

ellatrix:
@vcanales You can check the test re-runs

jorgefilipecosta (approved):
The code changes look good to me and make sense, I did some smoke testing around block styling and things seemed to work well.

Mamaduka:
@ellatrix, let's rebase this, and it probably should be good to be merged.

ellatrix:
Yes, would be good. I just didn't see a performance improvement so I paused it 🤔

PR #62695, opened 2024-06-20
Project Management Automation: Update action to run on node20

sirreal:
## What?

Use node20, not the very old, deprecated node12.

## Why?

node12 is outdated, unsupported, and deprecated.

Other PRs will have warnings, [see an example.](https://github.com/WordPress/gutenberg/actions/runs/9596257051)

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20 …
> The following actions uses node12 which is deprecated and will be forced to run on node16 …

## How?

Update the action.

## Testing Instructions

[See it run on CI](https://github.com/WordPress/gutenberg/actions/runs/9596333361). I'm not sure why there's still a node12 warning, maybe that will go away after this is landed.


sirreal:
#62851 did this.

PR #62698, opened 2024-06-20
Fix color remove button in Safari

### LINKED ISSUE #62261: Global Styles: Can't remove a custom color from palette in Safari

ellatrix:
## What?

This is an alternative PR to #62628. The problem is that the SVG element is focusable in Safari, which triggers the focus-outside hook (which normally ignores button clicks), so the onClick handler for the button is never called.

## Why?

Fixes https://github.com/WordPress/gutenberg/issues/62261.

## How?

The SVG is already made non focusable, but Safari seems to disobey that attribute. ~~However, `disabled` works.~~ While this worked for me, it didn't seem to work for @jorgefilipecosta. However, `pointer-events: 'none'` works for both of us. Maybe we should keep `disabled`, not sure.

## Testing Instructions

Use Safari.
Go to Global Styles→ Colors → Palette.
Click the three dots and then "Show details".


ellatrix:
FYI @afercia if you have more context.

jorgefilipecosta (approved):
Worked very well on my tests, thank you for looking into this issue @ellatrix 👍

afercia:
I have no idea why this specific icon button seems to behave differently from other buttons in Safari. Are we sure it's the SVG?

I can only report that `focusable: false` was originally introduced to solve a specific issue with Internet Explorer 11, where the SVGs where really focusable when placed inside a focuable control like a button. Tha required to presss the Tab key _twice_ to navigate through all buttons. But that was only specific to IE11.

ellatrix:
> I have no idea why this specific icon button seems to behave differently from other buttons in Safari. Are we sure it's the SVG?

It's because, while this button is open, there's also a popover that uses the `useFocusOutside` hook. This hook will detect clicks outside the popover to close. Because the SVG is the target, the popover will first close, which also removes the button.

afercia:
@ellatrix I replaced the `RemoveButton` with a normal `Button` and it seems to work for me in Safari. Is there any known issue with styled components and Safari? If that works, I think a more solid solution would be to use a standard Button and avoid styled components in the first place.

ellatrix:
Does it work consistently? @jorgefilipecosta Mentioned it can be inconsistent.

afercia:
> Does it work consistently? @jorgefilipecosta Mentioned it can be inconsistent.

I re-tested it and no, unfortunately it doesn't. 

To me, it's a matter of timing rather than focus or the SVG. It appears some browsers fire the click event before the button gets removed, while in Safari the button is removed first. I'd argue the root problem is in the design of this UI in the first place.
I'd like to mention it will need to be refactored anyways because it can't be used with a keyboard. For examxple, there is no way to remove a color when using only the keyboard. I will create a separete issue.

ellatrix:
@afercia Completely agree, after testing, it looks like this whole things is completely inaccessible. Once I open the popover, I can't even close it. It looks like this will need a larger refactor to get to work properly, and it feels too late just a few days before RC. Additionally, I wasn't aware that this problem (both the Safari issue and accessibility issue) has existed for a long time.

ellatrix:
I'm looking into a complete fix now

ellatrix:
@afercia Made a fix here: #62753. Does that fix some of the accessibility issues as well?

PR #62699, opened 2024-06-20
Zoom Out: Try click to disengage

scruffian:
## What?
This is just an experiment to see how it feels to use click to disengage zoom out.

The way it works is that it listens to clicks on a block. If you click the same block more than once in succession then zoom out mode is disengaged.

### DIFF
--- a/packages/block-editor/src/components/block-list/use-block-props/index.js
+++ b/packages/block-editor/src/components/block-list/use-block-props/index.js
@@ -11,6 +11,7 @@ import { __, sprintf } from '@wordpress/i18n';
 import { __unstableGetBlockProps as getBlockProps } from '@wordpress/blocks';
 import { useMergeRefs, useDisabled } from '@wordpress/compose';
 import warning from '@wordpress/warning';
+import { useDispatch } from '@wordpress/data';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


jasmussen:
Good instincts. I think the biggest headache right now is exactly that it's a bit far from both the cursor, and where focus is, how to close/exit the inserter in zoom mode. Here's trunk:

before

Note here that at one point I'm pressing Esc with the intent to close the inserter. That does appear to toggle off zoom out mode, but it doesn't close my patterns, or the inserter.

Here's this branch:

after

This branch does the same, except as the description notes, click twice on a block and it disengages zoom out mode. 

The problem in both cases is that the inserter remains open. Perhaps that's me personally, but that feels like the biggest headache to solve; can the inserter fully close when I press Esc? Clicking twice can do the same, that seems reasonable. My instinct also goes towards clicking anywhere in the gray gutter should close the inserter—the action I'm doing there is one of intentionally setting focus there by clicking. The main risk here is the vertical toolbar we want to show in this mode, we wouldn't want you to accidentally close the inserter, and thus zoom out mode, if you miss one of those buttons.

annezazu:
Did a quick test on this and found some oddities, namely trying to double click to get out and clicking in the template editing area, getting a notice. Along with this, when double clicking to get out, it's a bit odd where the Inserter with the patterns over top of it covers the editing experience:

https://github.com/WordPress/gutenberg/assets/26996883/b5a836e8-b7cc-4f96-938f-29477a3a5e0b

getdave:
@scruffian We solved in https://github.com/WordPress/gutenberg/pull/64573 right? Can we close?

scruffian:
I think we can close, but also exploring something similar in https://github.com/WordPress/gutenberg/pull/65286.

PR #62700, opened 2024-06-20
Keep an inserted pattern within view when added to the page

### LINKED ISSUE #59962: Keep an inserted pattern within view when added to the page
Currently, selecting a pattern from the Inserter results in the pattern added to the bottom of the page, but does not give you any real indication that it has been successfully added—other than a snackbar notification. Each pattern is already focused when inserted, but if it's not one of the first patterns on a page, you wouldn't see where it's applied/focused.

I propose that when adding patterns to a page, the pattern should be within view. 

This way: 
- I don't have to scroll down to find the pattern added to my page.
- I have immediate confidence that my selection occurred (rather than relying solely on a snackbar). 
- I can quickly take the next steps, whether that means to move it, [shuffle it to another pattern](https://github.com/WordPress/gutenberg/pull/59251), or delete it. 

### Current

https://github.com/WordPress/gutenberg/assets/1813435/e0680f4b-c3a2-4923-8e65-79a9106de86e

akasunil:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/59962

## Why?

Currently, when we add a pattern via the Inserter, the pattern is added to the bottom of the page and cannot be seen. This PR resolves the issue and makes the pattern visible on insert, allowing the user to see the newly added pattern.
 
## Testing Instructions
- Go do editor
- Insert pattern
- Observe that your pattern in view
 
## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/61308756/ed13acb2-c4a5-41c8-aea6-0352e2b86fc2


richtabor:
@akasunil is this still relevant? I'm seeing this behavior in trunk already:

https://github.com/user-attachments/assets/23e609c1-6434-41aa-bbb5-7943d30c93cc

akasunil:
@richtabor Yes, it works fine now in latest trunk. I'll close the PR.

akasunil:
Closing this out as issue is already fixed in latest trunk.

PR #62705, opened 2024-06-20
Posts list powered by DataViews

ntsekouras:
## What?
Part of: https://github.com/WordPress/gutenberg/issues/62371

This PR explores the addition of Data Views for the new experimental `posts` list under `Gutenberg->posts` menu.
This is just a starting point, but wanted to have it on GH to start some discussions.

What this PR does for now is the start of making an abstraction that could be used for any post type, but we need to do it at least for `posts` and `pages` for now.

Maybe the notable things for now to discuss are:
1. When we'll choose to edit an item, should it load the `post` or the `site` editor? I've started by having as a goal to load the site editor. This has nuances though. Do we want to have for example global styles in a the editor of the post? Additionally the editor adds some extra settings during initialization (`initializeEditor`) that are needed here too. How can we abstract this to be reused what's needed.
2. Design wise, we're also aiming towards having the `categories, tags`, etc.. in the sidebar where we would load the lists for the taxonomies. How does this affect the current `views` we have(`drafts, published, etc..`)? Should we consider changing placement of these views into the frame, which might also simplify some shared information that were wanted for views. An example is the number of items per view, which is not possible right now.

### Notes
1. We would need a similar API with `usePostActions` for fields. Now we are reusing the exact same code for the `posts` list and the existing `pages` list in site editor. So testing the `pages` list that works as before is also part of this PR to avoid regressions.

### DIFF
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -24,6 +24,19 @@ function gutenberg_posts_dashboard() {
 }
 
 /**
+ * Redirects to the new posts dashboard page and adds the postType query arg.
+ * TODO: there should be a better way to do this..
+ */
+function gutenberg_add_post_type_arg() {
+    global $pagenow;
+    if ( $pagenow == 'admin.php' && isset( $_GET['page'] ) && $_GET['page'] == 'gutenberg-posts-dashboard' && empty( $_GET['postType'] ) ) {
+        wp_redirect( admin_url( '/admin.php?page=gutenberg-posts-dashboard&postType=post' ) );
+        exit;
+    }
+}
+add_action( 'admin_init', 'gutenberg_add_post_type_arg' );
+
+/**
  * Replaces the default posts menu item with the new posts dashboard.
  */
 function gutenberg_replace_posts_dashboard() {
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -7,6 +7,14 @@
 
 add_action( 'admin_menu', 'gutenberg_replace_posts_dashboard' );
 
+// Default to is-fullscreen-mode to avoid jumps in the UI.
+add_filter(
+	'admin_body_class',
+	static function ( $classes ) {
+		return "$classes is-fullscreen-mode";
+	}
+);
+
 /**
  * Renders the new posts dashboard page.
  */
@@ -15,13 +23,13 @@ function gutenberg_posts_dashboard() {
 	// $current_screen = get_current_screen();
 	// $current_screen->is_block_editor( true );
 
-	// Default to is-fullscreen-mode to avoid jumps in the UI.
-	// add_filter(
-	// 	'admin_body_class',
-	// 	static function ( $classes ) {
-	// 		return "$classes is-fullscreen-mode";
-	// 	}
-	// );
+
+
+	// $indexed_template_types = array();
+	// foreach ( get_default_block_template_types() as $slug => $template_type ) {
+	// 	$template_type['slug']    = (string) $slug;
+	// 	$indexed_template_types[] = $template_type;
+	// }
 
 	$block_editor_context = new WP_Block_Editor_Context( array( 'name' => 'core/edit-site' ) );
 	$custom_settings      = array(
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -77,7 +77,6 @@ function gutenberg_posts_dashboard() {
 		)
 	);
 	wp_enqueue_script( 'wp-edit-site' );
-	// wp_enqueue_style( 'wp-edit-site' );
 
 	wp_enqueue_media();
 
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -11,14 +11,67 @@ add_action( 'admin_menu', 'gutenberg_replace_posts_dashboard' );
  * Renders the new posts dashboard page.
  */
 function gutenberg_posts_dashboard() {
+	// Flag that we're loading the block editor.
+	// $current_screen = get_current_screen();
+	// $current_screen->is_block_editor( true );
+
+	// Default to is-fullscreen-mode to avoid jumps in the UI.
+	// add_filter(
+	// 	'admin_body_class',
+	// 	static function ( $classes ) {
+	// 		return "$classes is-fullscreen-mode";
+	// 	}
+	// );
+
+	$block_editor_context = new WP_Block_Editor_Context( array( 'name' => 'core/edit-site' ) );
+	$custom_settings      = array(
+		'siteUrl'                   => site_url(),
+		// 'postsPerPage'              => get_option( 'posts_per_page' ),
+		'styles'                    => get_block_editor_theme_styles(),
+		// 'defaultTemplateTypes'      => $indexed_template_types,
+		// 'defaultTemplatePartAreas'  => get_allowed_block_template_part_areas(),
+		'supportsLayout'            => wp_theme_has_theme_json(),
+		// 'supportsTemplatePartsMode' => ! wp_is_block_theme() && current_theme_supports( 'block-template-parts' ),
+	);
+	$editor_settings = get_block_editor_settings( $custom_settings, $block_editor_context );
+
+	$active_global_styles_id = WP_Theme_JSON_Resolver::get_user_global_styles_post_id();
+	$active_theme            = get_stylesheet();
+
+	// Preload server-registered block schemas.
+	wp_add_inline_script(
+		'wp-blocks',
+		'wp.blocks.unstable__bootstrapServerSideBlockDefinitions(' . wp_json_encode( get_block_editor_server_block_settings() ) . ');'
+	);
+
+	wp_add_inline_script(
+		'wp-blocks',
+		sprintf( 'wp.blocks.setCategories( %s );', wp_json_encode( isset( $editor_settings['blockCategories'] ) ? $editor_settings['blockCategories'] : array() ) ),
+		'after'
+	);
+
+	/** This action is documented in wp-admin/edit-form-blocks.php */
+	do_action( 'enqueue_block_editor_assets' );
 	wp_register_style(
 		'wp-gutenberg-posts-dashboard',
 		gutenberg_url( 'build/edit-site/posts.css', __FILE__ ),
-		array( 'wp-components', 'wp-commands' )
+		array( 'wp-components', 'wp-commands', 'wp-edit-site' )
 	);
 	wp_enqueue_style( 'wp-gutenberg-posts-dashboard' );
-	wp_add_inline_script( 'wp-edit-site', 'window.wp.editSite.initializePostsDashboard( "gutenberg-posts-dashboard" );', 'after' );
+	// wp_add_inline_script( 'wp-edit-site', 'window.wp.editSite.initializePostsDashboard( "gutenberg-posts-dashboard" );', 'after' );
+	wp_add_inline_script(
+		'wp-edit-site',
+		sprintf(
+			'wp.domReady( function() {
+				wp.editSite.initializePostsDashboard( "gutenberg-posts-dashboard", %s );
+			} );',
+			wp_json_encode( $editor_settings )
+		)
+	);
 	wp_enqueue_script( 'wp-edit-site' );
+	// wp_enqueue_style( 'wp-edit-site' );
+
+	wp_enqueue_media();
 
 	echo '<div id="gutenberg-posts-dashboard"></div>';
 }
@@ -29,7 +82,7 @@ function gutenberg_posts_dashboard() {
  */
 function gutenberg_add_post_type_arg() {
     global $pagenow;
-    if ( $pagenow == 'admin.php' && isset( $_GET['page'] ) && $_GET['page'] == 'gutenberg-posts-dashboard' && empty( $_GET['postType'] ) ) {
+    if ( 'admin.php' === $pagenow && isset( $_GET['page'] ) && 'gutenberg-posts-dashboard' === $_GET['page'] && empty( $_GET['postType'] ) ) {
         wp_redirect( admin_url( '/admin.php?page=gutenberg-posts-dashboard&postType=post' ) );
         exit;
     }
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -25,6 +25,7 @@ function gutenberg_posts_dashboard() {
 		'styles'         => get_block_editor_theme_styles(),
 		'supportsLayout' => wp_theme_has_theme_json(),
 	);
+
 	$editor_settings         = get_block_editor_settings( $custom_settings, $block_editor_context );
 	$active_global_styles_id = WP_Theme_JSON_Resolver::get_user_global_styles_post_id();
 	$active_theme            = get_stylesheet();
@@ -70,11 +71,11 @@ function gutenberg_posts_dashboard() {
  * Redirects to the new posts dashboard page and adds the postType query arg.
  */
 function gutenberg_add_post_type_arg() {
-    global $pagenow;
-    if ( 'admin.php' === $pagenow && isset( $_GET['page'] ) && 'gutenberg-posts-dashboard' === $_GET['page'] && empty( $_GET['postType'] ) ) {
-        wp_redirect( admin_url( '/admin.php?page=gutenberg-posts-dashboard&postType=post' ) );
-        exit;
-    }
+	global $pagenow;
+	if ( 'admin.php' === $pagenow && isset( $_GET['page'] ) && 'gutenberg-posts-dashboard' === $_GET['page'] && empty( $_GET['postType'] ) ) {
+		wp_redirect( admin_url( '/admin.php?page=gutenberg-posts-dashboard&postType=post' ) );
+		exit;
+	}
 }
 add_action( 'admin_init', 'gutenberg_add_post_type_arg' );
 
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -19,43 +19,29 @@ add_filter(
  * Renders the new posts dashboard page.
  */
 function gutenberg_posts_dashboard() {
-	// Flag that we're loading the block editor.
-	// $current_screen = get_current_screen();
-	// $current_screen->is_block_editor( true );
-
-
-
-	// $indexed_template_types = array();
-	// foreach ( get_default_block_template_types() as $slug => $template_type ) {
-	// 	$template_type['slug']    = (string) $slug;
-	// 	$indexed_template_types[] = $template_type;
-	// }
-
 	$block_editor_context = new WP_Block_Editor_Context( array( 'name' => 'core/edit-site' ) );
 	$custom_settings      = array(
-		'siteUrl'                   => site_url(),
-		// 'postsPerPage'              => get_option( 'posts_per_page' ),
-		'styles'                    => get_block_editor_theme_styles(),
-		// 'defaultTemplateTypes'      => $indexed_template_types,
-		// 'defaultTemplatePartAreas'  => get_allowed_block_template_part_areas(),
-		'supportsLayout'            => wp_theme_has_theme_json(),
-		// 'supportsTemplatePartsMode' => ! wp_is_block_theme() && current_theme_supports( 'block-template-parts' ),
+		'siteUrl'        => site_url(),
+		'styles'         => get_block_editor_theme_styles(),
+		'supportsLayout' => wp_theme_has_theme_json(),
 	);
-	$editor_settings = get_block_editor_settings( $custom_settings, $block_editor_context );
-
+	$editor_settings         = get_block_editor_settings( $custom_settings, $block_editor_context );
 	$active_global_styles_id = WP_Theme_JSON_Resolver::get_user_global_styles_post_id();
 	$active_theme            = get_stylesheet();
+
+	$preload_paths = array(
+		array( '/wp/v2/media', 'OPTIONS' ),
+		'/wp/v2/types?context=view',
+		'/wp/v2/global-styles/' . $active_global_styles_id . '?context=edit',
+		'/wp/v2/global-styles/' . $active_global_styles_id,
+		'/wp/v2/global-styles/themes/' . $active_theme,
+	);
+	block_editor_rest_api_preload( $preload_paths, $block_editor_context );
 
 	// Preload server-registered block schemas.
 	wp_add_inline_script(
 		'wp-blocks',
 		'wp.blocks.unstable__bootstrapServerSideBlockDefinitions(' . wp_json_encode( get_block_editor_server_block_settings() ) . ');'
-	);
-
-	wp_add_inline_script(
-		'wp-blocks',
-		sprintf( 'wp.blocks.setCategories( %s );', wp_json_encode( isset( $editor_settings['blockCategories'] ) ? $editor_settings['blockCategories'] : array() ) ),
-		'after'
 	);
 
 	/** This action is documented in wp-admin/edit-form-blocks.php */
@@ -66,7 +52,6 @@ function gutenberg_posts_dashboard() {
 		array( 'wp-components', 'wp-commands', 'wp-edit-site' )
 	);
 	wp_enqueue_style( 'wp-gutenberg-posts-dashboard' );
-	// wp_add_inline_script( 'wp-edit-site', 'window.wp.editSite.initializePostsDashboard( "gutenberg-posts-dashboard" );', 'after' );
 	wp_add_inline_script(
 		'wp-edit-site',
 		sprintf(
@@ -77,15 +62,12 @@ function gutenberg_posts_dashboard() {
 		)
 	);
 	wp_enqueue_script( 'wp-edit-site' );
-
 	wp_enqueue_media();
-
 	echo '<div id="gutenberg-posts-dashboard"></div>';
 }
 
 /**
  * Redirects to the new posts dashboard page and adds the postType query arg.
- * TODO: there should be a better way to do this..
  */
 function gutenberg_add_post_type_arg() {
     global $pagenow;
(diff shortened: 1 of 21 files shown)


packages/edit-site/src/components/posts-app/index.js
```diff
@@ -10,29 +10,27 @@ import { privateApis as routerPrivateApis } from '@wordpress/router';
...
-	areas: {
-		sidebar: 'Empty Sidebar',
-		content: <Page>Welcome to Posts</Page>,
-		preview: undefined,
-		mobile: <Page>Welcome to Posts</Page>,
-	},
-};
+function PostsLayout() {
+	// This ensures the edited entity id and type are initialized properly.
+	useInitEditedEntityFromURL();
```
youknowriad, line 23: How hard would it be to move away from this hook and instead of having postId postType and context in the site editor store, just pass it down as props to the `Editor` component. (For the site editor, not the posts page, we could keep syncing these to the store for backwards compatibility)

packages/edit-site/src/components/page-pages/style.scss
```diff
@@ -1,3 +1,6 @@
+// TODO: we need to examine if these styles could be the default styles
+// for every post type and/or how to have different styles for different post types.
+// Example is `posts` and `pages`.
```
youknowriad: The classnames seems wrong now right?

packages/edit-site/src/components/layout/router.js
```diff
@@ -92,15 +92,15 @@ export default function useLayoutAreas() {
...
+				content: <PostsList postType={ postType } />,
 				preview: ( isListLayout || canvas === 'edit' ) && (
 					<Editor isLoading={ isSiteEditorLoading } />
 				),
 				mobile:
 					canvas === 'edit' ? (
 						<Editor isLoading={ isSiteEditorLoading } />
 					) : (
-						<PagePages />
+						<PostsList postType={ postType } />
```
youknowriad, line 99: This is the kind of change I was hoping to see :)

packages/edit-site/src/components/posts-app/router.js
```diff
@@ -0,0 +1,82 @@
...
+	const { postType, layout, canvas } = params;
+	const labels = useSelect(
+		( select ) => {
+			return select( coreStore ).getPostType( postType )?.labels;
+		},
+		[ postType ]
+	);
+
+	// Posts list.
+	if ( [ 'page', 'post' ].includes( postType ) ) {
```
youknowriad: Why page?

packages/edit-site/src/components/posts-app/router.js
```diff
@@ -0,0 +1,82 @@
...
+import DataViewsSidebarContent from '../sidebar-dataviews';
+import PostsList from '../posts-app/posts-list';
+
+const { useLocation } = unlock( routerPrivateApis );
+
+export default function useLayoutAreas() {
+	const isSiteEditorLoading = useIsSiteEditorLoading();
+	const { params = {} } = useLocation();
+	const { postType, layout, canvas } = params;
+	const labels = useSelect(
```
youknowriad, line 23: I would prefer if we avoid useSelect and complex logic in the "router". Ideally this is actually just a global object (when we'll be able to define "paths" as part of the config). So I'd actually move these calls to dedicated components that I render in areas.

packages/edit-site/src/components/posts-app/router.js
```diff
@@ -0,0 +1,82 @@
...
+import Editor from '../editor';
+import SidebarNavigationScreen from '../sidebar-navigation-screen';
+import SidebarNavigationScreenMain from '../sidebar-navigation-screen-main';
+import DataViewsSidebarContent from '../sidebar-dataviews';
+import PostsList from '../posts-app/posts-list';
+
+const { useLocation } = unlock( routerPrivateApis );
+
+export default function useLayoutAreas() {
+	const isSiteEditorLoading = useIsSiteEditorLoading();
```
youknowriad: Already existing in the other router, but maybe we can just move this to the "Editor" component now?

youknowriad (commented):
Great start

packages/edit-site/src/components/posts-app/router.js
ntsekouras: Just a lot of temp code to get first feedback. These should be generic eventually and we could even restrict this for `posts` as a first step..

packages/edit-site/src/components/posts-app/router.js
ntsekouras: Yeah, I have this in mind for things that end up being reused, to be moved elsewhere.. Just temp code for first feedback.

youknowriad:
> When we'll choose to edit an item, should it load the post or the site editor? I've started by having as a goal to load the site editor. This has nuances though. Do we want to have for example global styles in a the editor of the post? Additionally the editor adds some extra settings during initialization (initializeEditor) that are needed here too. How can we abstract this to be reused what's needed.

Good question, I'm thinking we use the same "component", it doesn't matter if we call it post or site editor. That component will probably be the site editor one for a start because we're building this stuff in the site editor package today. The Global Styles question is a good one, I don't think we need to get everything right in the first PR. I think it's fine to keep it as is for now.

> Design wise, we're also aiming towards having the categories, tags, etc.. in the sidebar where we would load the lists for the taxonomies. How does this affect the current views we have(drafts, published, etc..)? Should we consider changing placement of these views into the frame, which might also simplify some shared information that were wanted for views. An example is the number of items per view, which is not possible right now.

Good question as well, maybe "posts" is a top level menu item and you gave to go into it first (like pages in the site editor) but it raises the question of what is the "front page" of this experience. Also a design question mainly that can be addressed in a follow-up.

> We would need a similar API with usePostActions for fields. Now we are reusing the exact same code for the posts list and the existing pages list in site editor. So testing the pages list that works as before is also part of this PR to avoid regressions.

Definitely 💯

jasmussen:
> Good question as well, maybe "posts" is a top level menu item and you gave to go into it first (like pages in the site editor) but it raises the question of what is the "front page" of this experience. Also a design question mainly that can be addressed in a follow-up.

That's the key bit. It can live here fo now:

Screenshot 2024-06-21 at 09 56 13

But in order to solve the structural questions it does need to move to the top level.

It's a good start, but I am seeing some errors, for awareness:
2024-06-21 09 54 55 local-wordpress local f52f2822480a

packages/edit-site/src/components/posts-app/posts-list.js
```diff
@@ -293,13 +292,18 @@ export default function PagePages() {
...
 	const {
-		records: pages,
+		records,
 		isResolving: isLoadingPages,
 		totalItems,
 		totalPages,
 	} = useEntityRecords( 'postType', postType, queryArgs );
 
-	usePostIdLinkInSelection( selection, setSelection, isLoadingPages, pages );
+	usePostIdLinkInSelection(
```
jorgefilipecosta: This can now be removed.

packages/edit-site/src/components/posts-app/posts-list.js
```diff
@@ -312,20 +316,26 @@ export default function PagePages() {
...
 			const { getEntityRecord, getPostType, canUser } =
 				select( coreStore );
 			const siteSettings = getEntityRecord( 'root', 'site' );
 			return {
 				frontPageId: siteSettings?.page_on_front,
 				postsPageId: siteSettings?.page_for_posts,
-				addNewLabel: getPostType( 'page' )?.labels?.add_new_item,
-				canCreatePage: canUser( 'create', 'pages' ),
+				labels: getPostType( postType )?.labels,
+				// TODO: check what is the proper way to make this work for any post type..
```
jorgefilipecosta: To get the resource (posts, pages etc) we can do this:
```
const postTypeObject = getPostType( postType );
const resource = postTypeObject?.rest_base || '';
```

packages/edit-site/src/components/sidebar-dataviews/default-views.js
```diff
@@ -54,103 +58,114 @@ const DEFAULT_PAGE_BASE = {
...
-			view: {
-				...DEFAULT_PAGE_BASE,
-				filters: [
-					{
-						field: 'status',
-						operator: OPERATOR_IS_ANY,
-						value: 'future',
+				},
+				{
+					title: __( 'Scheduled' ),
```
jorgefilipecosta, line 94: We are repeating the status view definition all the time I guess we can do something like:

```
const createStatusView = (title, slug, icon, status) => ({
  title,
  slug,
  icon,
  view: {
    ...DEFAULT_POST_BASE,
    filters: [
      {
        field: 'status',
        operator: OPERATOR_IS_ANY,
        value: status,
      },
    ],
  },
});

// Then use it like:
createStatusView(__('Published'), 'published', published, 'publish'),
createStatusView(__('Drafts'), 'drafts', drafts, 'draft'),
// ... and so on
```

annezazu:
I was trying to test this today and got an error:

Screenshot 2024-06-26 at 1 36 45 PM

>TypeError: k.styles is not iterable
    at https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/edit-site/index.min.js?ver=29fd8339093b1150d2ff:72:3791
    at Object.useMemo (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:66553)
    at t.useMemo (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react.min.js?ver=18:2:6087)
    at cg (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/edit-site/index.min.js?ver=29fd8339093b1150d2ff:72:3779)
    at hu (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:60255)
    at xi (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:119196)
    at bs (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:108448)
    at vs (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:108376)
    at gs (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:108239)
    at as (https://playground.wordpress.net/scope:0.5644724289443985/wp-content/plugins/gutenberg/build/vendors/react-dom.min.js?ver=18:2:105074)

Minor too but this should be "Add post" rather than "Add page":
Screenshot 2024-06-26 at 1 36 52 PM

ntsekouras:
>I was trying to test this today and got an error:

@annezazu it's being tracked and in progress. I only hid your comment because I'm expecting too many of them in this PR and had big screenshots 😄 .

packages/edit-site/src/posts.js
```diff
@@ -11,16 +30,67 @@ import PostsApp from './components/posts-app';
...
+	// This is needed because edit-site doesn't initialize with EditorProvider,
+	// which internally uses updateEditorSettings as well.
+	dispatch( editorStore ).updateEditorSettings( {
+		defaultTemplateTypes: settings.defaultTemplateTypes,
+		defaultTemplatePartAreas: settings.defaultTemplatePartAreas,
+	} );
+
+	// Prevent the default browser action for files dropped outside of dropzones.
+	window.addEventListener( 'dragover', ( e ) => e.preventDefault(), false );
+	window.addEventListener( 'drop', ( e ) => e.preventDefault(), false );
```
youknowriad, line 81: Feels like too much boilerplate. It's ok for now but might be good to take a look at this at some point to see where these things should be moved (or removed)

packages/edit-site/src/posts.js
```diff
@@ -11,16 +30,67 @@ import PostsApp from './components/posts-app';
 
 /**
  * Initializes the "Posts Dashboard"
- * @param {string} id DOM element id.
+ * @param {string} id       ID of the root element to render the screen in.
+ * @param {Object} settings Editor settings.
  */
-export function initializePostsDashboard( id ) {
+export function initializePostsDashboard( id, settings ) {
```
youknowriad, line 34: Can we avoid using the settings from the server? I would love if we can fetch all the necessary settings using REST API instead.

packages/edit-site/src/posts.js
ntsekouras: Agreed. I'm just looking at site editor load(PHP + JS) and will need to investigate what are the things we need.

packages/edit-site/src/posts.js
ntsekouras: I'll explore this, yes. I'm wondering though why we have it like this in site editor for now.. Maybe the endpoint is missing some settings 🤔 ?

youknowriad: There's a lot of things that get through that settings object both in post and site editors and yeah initially the endpoints didn't have most of the settings. I'm fine with you did here for now. But let's track this. I think we should look at each setting separately and see how to avoid it.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
ntsekouras: I don't think we gain much for now to do this change. Maybe see how to best refactor this code when we work more on `views`.

packages/edit-site/src/components/posts-app/posts-list.js
ntsekouras: Thanks Jorge!

ntsekouras:
@jorgefilipecosta can you help me figure out why some actions like the `trashPostAction` don't appear for `posts` list? I need to check more the actions and it seems there are some differences between `post and page`, but I wasn't able to figure it out yet.

Additionally I observed that when we don't have any bulk actions and therefore no `checkbox`, at least the `table` view renders weird, because we have attached styles to that checkbox. We should probably finds a more resilient way to have good default styles. --cc @jameskoster 

Finally a last thing that maybe Jay can help is that I couldn't find why in `posts` list and `grid` view, the title is not aligned vertically..

jorgefilipecosta:
Sure @ntsekouras, I'm having a look into the actions issue.

jameskoster:
> Additionally I observed that when we don't have any bulk actions and therefore no checkbox, at least the table view renders weird, because we have attached styles to that checkbox. We should probably finds a more resilient way to have good default styles

I pushed a fix for that one.

jameskoster:
> Finally a last thing that maybe Jay can help is that I couldn't find why in posts list and grid view, the title is not aligned vertically..

It's because `dataviews-view-grid__primary-field` has a different `display` property in the posts list. It should be `flex` but here it is `block`. 

Edit: Pushed a fix for this one too.

lib/experimental/posts/load.php
```diff
@@ -7,22 +7,95 @@
 
 add_action( 'admin_menu', 'gutenberg_replace_posts_dashboard' );
 
+// Default to is-fullscreen-mode to avoid jumps in the UI.
+add_filter(
+	'admin_body_class',
+	static function ( $classes ) {
```
jorgefilipecosta, line 13: Normally anonymous functions are not recommended in core, because they don't allow a third-party plugin to use remove_filter.

lib/experimental/posts/load.php
```diff
@@ -7,22 +7,95 @@
...
+		return "$classes is-fullscreen-mode";
+	}
+);
+
 /**
  * Renders the new posts dashboard page.
  */
 function gutenberg_posts_dashboard() {
+	// Flag that we're loading the block editor.
+	// $current_screen = get_current_screen();
```
jorgefilipecosta: Are these code comments still required?

ntsekouras: Still need to do at least a first good pass of needed settings. I'll update soon.

lib/experimental/posts/load.php
ntsekouras: That's what we do for site editor in core right now. I'll have to check if I can add the filter only in this new page though.. I didn't observe something weird so far though..

lib/experimental/posts/load.php
```diff
@@ -7,22 +7,95 @@
...
+	);
 	wp_enqueue_script( 'wp-edit-site' );
 
+	wp_enqueue_media();
+
 	echo '<div id="gutenberg-posts-dashboard"></div>';
 }
 
+/**
+ * Redirects to the new posts dashboard page and adds the postType query arg.
```
jorgefilipecosta, line 71: Instead of doing a redirect to include the post type. Could we assume a default post-type? e.g: if empty( $_GET['postType'] ) we assume it has the value post on our code? 
The current post list seems to work like that it does not have a postType parameter in case of post.

packages/dataviews/src/dataviews.tsx
```diff
@@ -51,22 +54,6 @@ const defaultGetItemId = ( item: ItemWithId ) => item.id;
 
 const defaultOnSelectionChange = () => {};
 
-function useSomeItemHasAPossibleBulkAction< Item >(
```
jorgefilipecosta, line 54: The change does not seem related. Probably it was a rebase issue.

jorgefilipecosta (commented):
The back button is not working:
Screenshot 2024-06-28 at 10 55 11

lib/experimental/posts/load.php
jorgefilipecosta: Maybe the current one for site editor, passed during a code review but as far as I know anonymous functions should be avoided, they work well the issue is if a plugin wants to remove the filter the plugin can not do it.

packages/edit-site/src/components/sync-state-with-url/use-init-edited-entity-from-url.js
```diff
@@ -213,7 +213,8 @@ function useResolveEditedEntityAndContext( { postId, postType } ) {
 		if ( postType && postId && authorizedPostTypes.includes( postType ) ) {
 			return { postType, postId };
 		}
-
+		// TODO: for post types lists we should probably not render the front page, but maybe a placeholder
```
jorgefilipecosta, line 216: I guess for other post types, we should show the list page for that post type instead of the front page. For example, for posts we could show the blog front page.

packages/edit-site/src/posts.js
```diff
@@ -1,24 +1,96 @@
...
 	if ( ! globalThis.IS_GUTENBERG_PLUGIN ) {
 		return;
 	}
 	const target = document.getElementById( id );
 	const root = createRoot( target );
 
+	dispatch( blocksStore ).reapplyBlockTypeFilters();
```
jorgefilipecosta, line 41: This code seems to be a duplication of packages/edit-site/src/index.js, could we create some function used in both places to avoid the duplication?
I fear in the future someone may easily change code in a place forgetting about the other place.

youknowriad: I think duplication is fine sometimes, as it reminds us that this code is unnecessary boilerplate, it's not clear why we need to call it here in the first place.

jorgefilipecosta: I guess if we are duplicating at least we should include comments in both files saying they are duplication of each other. Duplication ends up leading to divergence and we may end up with the double of the code to maintain.
I think code like block registration etc may be needed because the preview of the post depends on the blocks being registered and the editor being properly initialized.

jorgefilipecosta (commented):
Things are working mostly well there are only some minor todo items pending. I think we should add a posts item in the site editor. It is very easy to run into a different site editor menu that then does not allow to go back to posts.

For example,open a post, edit the template of that post and press the back button. It is strange that a posts menu item is not there. Also, the items there like pages styles etc don't work at all.

That makes me wonder if instead of following this path, where we have a different initialization initializePostsDashboard, We could not just add a Posts tab in the site editor, and use the normal initializeEditor( id, settings ) as initialization given that we have a postType=post URL param the posts section would open right away. If we don't want to expose the posts section generally yet we could have a flag that just makes the section visible if the URL parameter postType=post was set when the component mounted.

youknowriad:
@jorgefilipecosta I agree it's easier if this was developed as part of the site editor but that's not the UX experience we're targetting at this point. The idea is to replace the "posts" menu in wp-admin top level as otherwise the site editor will lose its "design" meaning. 

The fact that it's in a separate experimental page is temporary until we reach parity with the posts section.

packages/dataviews/src/dataviews.tsx
ntsekouras: It's unrelated but intentional. We had this duplicate function in the same package with no reason. I could remove from this PR, but I think it's harmless and super straightforward.

lib/experimental/posts/load.php
ntsekouras: We need to have the param in the URL for the editor to resolve the current entities etc..

jorgefilipecosta:
> @jorgefilipecosta I agree it's easier if this was developed as part of the site editor but that's not the UX experience we're targetting at this point. The idea is to replace the "posts" menu in wp-admin top level as otherwise the site editor will lose its "design" meaning.
> 
> The fact that it's in a separate experimental page is temporary until we reach parity with the posts section.

Hi @youknowriad, I understand that but I'm not convinced we need a different initialization. We could for example use exactly the same one, with a setting/flag to hide the other sections besides posts. Posts would be a normal section. So if the flag was not there the sidebar would render normal sections patterns, templates etc but not posts, with the flag there it would only include the posts section for now, given that we have a ?postType=posts we would automatically load with the post section open by default.


I guess we also need to decide what happens when the user goes to edit a template or pattern and presses the back button. Should we show the normal site editor sections if yes we need to make them work well, if not I guess we should add some logic on the back to on that case go back to the posts sidebar.

packages/dataviews/src/dataviews.tsx
jorgefilipecosta: Got it feel free to leave it.

lib/experimental/posts/load.php
jorgefilipecosta: Maybe we could have a small js code on loading adding postType=post with history.push({postType:'post'}) if a postType is not set.

youknowriad:
> Hi @youknowriad, I understand that but I'm not convinced we need a different initialization. We could for example use exactly the same one, with a setting/flag to hide the other sections besides posts. Posts would be a normal section. So if the flag was not there the sidebar would render normal sections patterns, templates etc but not posts, with the flag there it would only include the posts section for now, given that we have a ?postType=posts we would automatically load with the post section open by default.

I again agree with you that it's easiest to do this. But for me, doing a separately initialization is better because it actually force us to think about the initialization code base and remove it / simplify it basically. IMO, it shouldn't be necessary most of the time. It's either in the wrong place, not using the right API, not collocated in the right component... and using a separate initialization function force us to look at this gaps that we have and fix them.

> I guess we also need to decide what happens when the user goes to edit a template or pattern and presses the back button. Should we show the normal site editor sections if yes we need to make them work well, if not I guess we should add some logic on the back to on that case go back to the posts sidebar.

That's a good point and one we need to solve regardless of how we initialize this. I think there's probably a short term and a long term reply to this:

 - short term version is keep the behavior we currently have in the "post editor". "W" button always returns to the posts list.
 - long term version: do not touch the "W" button but navigating to a template or template part actually navigates to the site editor's template and template part. This can only work if everything is in a single SPA.

packages/edit-site/src/components/block-editor/use-site-editor-settings.js
```diff
@@ -136,7 +136,10 @@ export function useSpecificEditorSettings() {
 				templateSlug: _record.slug,
 				canvasMode: getCanvasMode(),
 				settings: getSettings(),
-				postWithTemplate: _context?.postId,
+				postWithTemplate:
```
youknowriad: "posts" actually have template, so I would change the value of this flag personally.

 - So either we should rename to something about `shouldUseTemplateAsDefaultRenderingMode` if it's only used for that config.
 - Move the "post type" check to later (when setting the `defaultRenderingMode` variable.

packages/edit-site/src/components/editor/index.js
```diff
@@ -217,9 +218,20 @@ export default function EditSiteEditor() {
...
 										className="edit-site-layout__view-mode-toggle"
-										onClick={ () =>
-											setCanvasMode( 'view' )
-										}
+										onClick={ () => {
+											setCanvasMode( 'view' );
+											// TODO: this is a temporary solution to navigate to the posts list if we are
+											// come here throught `posts list` and are in focus mode editing a template or template part.
+											if (
+												isPostsList &&
```
youknowriad, line 226: I think that's fine but we should be careful about not using this variable in a lot of places.

packages/edit-site/src/components/block-editor/use-site-editor-settings.js
ntsekouras: Makes sense to rename and was only used for the rendering mode selection, so I renamed to `shouldUseTemplateAsDefaultRenderingMode.

youknowriad:
One thing that is unclear to me is what to do in the "home page" of the posts section. Right now it's showing the site's home page like the site editor but I feel that's not idea. 

That said I think we can merge this PR now. There's a lot of design questions that we need to follow-up with.

youknowriad (approved):

ntsekouras:
>One thing that is unclear to me is what to do in the "home page" of the posts section. Right now it's showing the site's home page like the site editor but I feel that's not idea.

For me the ideal thing is to show a placeholder like many other similar apps/views do which could be something along `Please select something to preview` and/or `nothing is selected`.

jorgefilipecosta:
> One thing that is unclear to me is what to do in the "home page" of the posts section. Right now it's showing the site's home page like the site editor but I feel that's not idea.
> 
> That said I think we can merge this PR now. There's a lot of design questions that we need to follow-up with.

Hi @youknowriad, I think if there is a blog front page it should show it.

ellatrix:
Curious what happened here that affected the navigate metric so positively. Also the metric is more unstable now.

ntsekouras:
>Curious what happened here that affected the navigate metric so positively. Also the metric is more unstable now.

This PR loads the post list under a new admin page and shouldn't affect any existing tests. 🤔

ellatrix:
I'm talking about the pages list, that one was touched in this PR, no?

PR #62706, opened 2024-06-20
CustomSelectControl: align v1 and legacy v2 unit tests

ciampo:
## What?

While working on #62255 and, more in general, on refining the new version of `CustomSelectControl`, one of the main tasks is to make sure that the new version of the component offers an easy way to replace the old implementation with the new implementation pain-free — and that is what the legacy v2 version is for. It implements the same API surface as the v1 and it aims at behaving as closely as possible as the v1, despite its internal implementation is completely different.

This PR aims at aligning further the unit tests between v1 and legacy v2 as much as possible, as a way to gain more confidence when making future fixes and changes to the legacy v2 version.

## Why?

See above.

## How?

Making sure that the unit tests for v1 and legacy v2 overlap as much as possible

## Testing Instructions

- Make sure both suites of unit tests pass
- Review code changes (commit by commit may be easier)

No runtime changes

### DIFF
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -289,6 +289,7 @@ describe.each( [
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
+				// TODO: key should be different — this is a known bug and will be fixed
 				selectedItem: { key: 'violets', name: 'violets' },
 				type: '',
 			} )
@@ -332,7 +333,8 @@ describe.each( [
 			1,
 			expect.objectContaining( {
 				selectedItem: expect.objectContaining( {
-					key: 'flower1',
+					// TODO: key should be different — this is a known bug and will be fixed
+					key: 'violets',
 					name: 'violets',
 				} ),
 			} )
@@ -345,7 +347,8 @@ describe.each( [
 			2,
 			expect.objectContaining( {
 				selectedItem: expect.objectContaining( {
-					key: 'flower3',
+					// TODO: key should be different — this is a known bug and will be fixed
+					key: 'poppy',
 					name: 'poppy',
 				} ),
 			} )
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -14,7 +14,11 @@ import { useState } from '@wordpress/element';
  */
 import UncontrolledCustomSelectControl from '..';
 
-const customClass = 'amber-skies';
+const customClassName = 'amber-skies';
+const customStyles = {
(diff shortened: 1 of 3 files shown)


packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -281,11 +282,14 @@ describe.each( [
...
 		);
 
+		// NOTE: legacy CustomSelectControl doesn't fire onChange
+		// at this point in time.
 		expect( mockOnChange ).toHaveBeenNthCalledWith(
 			1,
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
+				// TODO: key should be different — this is a known bug and will be fixed
```
tyxla, line 292: 👍

packages/components/src/custom-select-control/test/index.js
```diff
@@ -262,6 +271,81 @@ describe.each( [
...
+			/>
+		);
+
+		await user.click(
+			screen.getByRole( 'button', {
+				expanded: false,
+			} )
+		);
+
+		// DIFFERENCE WITH V2: NOT CALLED
```
tyxla, line 292: If this is something that we need to update, let's add a TODO comment to remember to address it

packages/components/src/custom-select-control/test/index.js
```diff
@@ -262,6 +271,81 @@ describe.each( [
...
+
+		render(
+			<Component
+				{ ...props }
+				value={ props.options[ 0 ] }
+				onChange={ mockOnChange }
+			/>
+		);
+
+		await user.click(
```
tyxla, line 286: Just to confirm, using `@testing-library/user-event` for v1 vs using `@ariakit/test` for legacy v2 is an expected difference, right?

tyxla (approved):
Looks great. I appreciate the clarity this brings when looking at the 2 distinct test suites 🙌

packages/components/src/custom-select-control/test/index.js
ciampo: It is — we had to use `@ariakit/test` for ariakit-based components as (if I recall) `@testing-library/user-event` doesn't work well with how Ariakit performs DOM updates

packages/components/src/custom-select-control/test/index.js
ciampo: I'm not sure about how much we can do about this, since:
- we don't want to change the v1 behaviour (which doesn't call `onChange` on initial render)
- the v2 behaviour comes directly from `Ariakit` — I'm not sure how easy it would be to suppress _only_ the first call without introducing unwanted side effects.

I will add an item to the [tracking issue](https://github.com/WordPress/gutenberg/issues/55023)

ciampo: Good news! This will be fixed in https://github.com/WordPress/gutenberg/pull/62733

PR #62709, opened 2024-06-20
Global Styles: Include spacing rules as part of typography variations

### LINKED ISSUE #61214: Allow spacing in typography variations
As a follow-up to https://github.com/WordPress/gutenberg/pull/60220, I propose we allow for `spacing` to be allowed within typography variations. 

This would allow typography variations to be much more refined, as the space between paragraphs (`spacing.blockGap`) may be modified by a typography variation, as well as adjustments for elements like buttons—which may need more padding with an accompanying larger font size. 

### Visual

Here's a visual of the current typography variations, which are filtered from theme style variations: 

CleanShot 2024-04-29 at 15 16 56

scruffian:
## What?
Includes spacing as part of typography presets. Fixes #61214.

## Why?

Spacing is an important part of typography styles, so we should treat these rules as part of the typography variation.

## How?

Several functions have been expanded to take an array of properties rather than a single property:
- `useCurrentMergeThemeStyleVariationsWithUserConfig`
- `isVariationWithSingleProperty` -> `isVariationWithProperties`
- `filterObjectByProperty` -> `filterObjectByProperties`
- `removePropertyFromObject` -> `removePropertiesFromObject`

## Testing Instructions
1. Using TT4 add a new style variation that contains only typography and spacing rules. [Here's an example](https://gist.github.com/scruffian/3dab906ee04d3c1c689bb08d23198f84).
2. Open the Site Editor
3. Go to Global Styles > Typography
4. You should see a new typography variation under "Presets"

## Screenshots or screencast 
Screenshot 2024-06-21 at 10 50 45

### DIFF
--- a/packages/edit-site/src/components/global-styles/hooks.js
+++ b/packages/edit-site/src/components/global-styles/hooks.js
@@ -114,7 +114,7 @@ export function useSupportedStyles( name, element ) {
 
 export function useColorVariations() {
 	const colorVariations = useCurrentMergeThemeStyleVariationsWithUserConfig( {
-		property: 'color',
+		properties: [ 'color' ],
 	} );
 	/*
 	 * Filter out variations with no settings or styles.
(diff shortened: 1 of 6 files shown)


scruffian:
There was some concern about the approach here - does it make sense to make this exception? Can we find/create other tools within typography that gives us the same thing?

scruffian:
I'm not convinced this is a good idea.

PR #40625, opened 2022-04-26
[WIP] Account for positioned elements when positioning Block Toolbar

getdave:
## What?

This PR is an experiment which seeks the element which

## Why?

In https://github.com/WordPress/gutenberg/issues/40382 we learnt that some blocks have content with extends beyond the boundary of the block's root element. 

As the block's toolbar is positioned immediately below the bottom boundary of the block, this can, on occasion, obstruct the ability to interact with crucial elements that are underneath the block toolbar.

This is most prominent on the Navigation block, where the toolbar will often be positioned on top of flyout submenus.

Closes https://github.com/WordPress/gutenberg/issues/40382

## How?

This PR explores one of the possible approaches outlined in https://github.com/WordPress/gutenberg/issues/36977#issuecomment-992513795, namely:

> Keep the conditional bottom positioning, but manage to include any popover/dropdown for children blocks in the calculation of where the bottom position should be so we aren't obscuring anything like this.

Currently the implementation 

- looks for the root block element and gets it's bottom offset from the top of the viewport (using [`Element.getBoundingClientRect().bottom`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)
- iterates through all its child elements comparing their bottom offset against the stored root bottom offset
- any elements that extend _beyond_ the bounds of the root element's bottom offset value are considered to be "outside" the boundary of the block. 
- If found, we pick the one which extends the furtherest and use that as the anchor reference for the `<Popover>` component. 
- The toolbar is then affixed to the bottom of the extruding element rather than the bottom of the root block's element. This means the toolbar doesn't obstruct the content.

The implementation also watches for DOM mutations and will retrigger the toolbar positioning in case of changes (for example elements which are shown/hidden based on classnames .etc).

### Caveats/Concerns

My main concern is inefficiency of the algorithm:

- we currently recusively loop through all child elements. I was hoping to only recurse if the parent proved to extend beyond the root boundary but it turns out that won't work as child elements can extend even if their parent's do not.
- We use `MutationObserver` and I have a hunch that this could be computationally expensive.
- `getComputedStyle` is used which [can force layout reflows](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle) 
- the conditions for determining whether an element is currently "visable" are arbitrary

I also think this needs much more stress testing. What other 3rd party blocks have instances whether the block's content extends beyond its layout bounds?

## Testing Instructions

1. Create Nav block with several levels of nesting using the submenu block.
[...]

PR #62711, opened 2024-06-21
Position BlockToolbar below all of the selected block's descendants

### LINKED ISSUE #40382: Navigation Block: Toolbar overlaps with submenu links when opening sidebar or using smaller viewport
### Description

Navigation toolbar overlaps on submenu links when we open sidebar settings or on smaller viewport. This makes editing or addting submneu links difficult.

1. Add navigation block
2. Add page link
3. Add submenu links
4. Open  setting sidebar or test on smaller viewport

Uploading Error Sub menu.mp4…

### Environment info

WordPress - 5.9.3 on local
Gutenberg - 13.0.0
Default Theme
No other plugin or theme installed.
Browser - Chrome

Yes

Yes

noisysocks:
## What, why, how
Fixes https://github.com/WordPress/gutenberg/issues/40382.

Currently `BlockToolbar` positions itself around a block's bounds using `element.getBoundingClientRect()`.

The problem with this is that it doesn't take into account nested elements that overflow the block. For example, Submenu blocks within a Navigation block will overflow the Navigation block when the user hovers over a menu item. This causes the toolbar to be positioned on top of the Submenu which results in a very unusable experience. 

This PR is basically a refresh of @getdave's approach in https://github.com/WordPress/gutenberg/pull/40625, though I've tried to make things a bit more generic.

Instead of using `element.getBoundingClientRect()` we recurse through `element`'s descendants and compute the block's bounds by taking a union of the root rect and all descendent rects. Hidden elements are ignored. This results in a rect that corresponds to what the user _sees_ of the block.

This logic is applied in two places: 1) `BlockPopover` where the `Popover` components `anchor` is computed; and 2) `useBlockToolbarPopover` which decides whether or not to set `flip` and `shift` on the `Popover` depending on whether the block is too close to the header.

## Testing Instructions
1. Install and activate _Emptytheme_, or use the Create Block Theme plugin to create a new empty theme.
2. Go to Appearance → Editor and edit the default template.
3. Insert a Navigation block into the first position in the header template part.
4. Edit the Navigation block and add a submenu to one of the navigation items.
5. Try to edit the submenu. The block toolbar should not obscure your cursor.

Also need to test that the block popover appears correctly for other blocks in other instances.

## Screenshots or screencast 

Before:

https://github.com/WordPress/gutenberg/assets/612155/ffdbeb04-e7c7-42ec-bbb5-ec80af1962a3

After:

https://github.com/WordPress/gutenberg/assets/612155/a21339a4-d4e4-4c49-843c-6e6cdb302a8d

### DIFF
--- a/packages/block-editor/src/components/block-tools/block-toolbar-popover.js
+++ b/packages/block-editor/src/components/block-tools/block-toolbar-popover.js
@@ -49,7 +49,7 @@ export default function BlockToolbarPopover( {
 
 	const popoverProps = useBlockToolbarPopoverProps( {
 		contentElement: __unstableContentRef?.current,
-		clientId,
+		clientId: capturingClientId || clientId,
 	} );
 
 	return (
--- a/packages/block-editor/src/components/block-tools/block-toolbar-popover.js
+++ b/packages/block-editor/src/components/block-tools/block-toolbar-popover.js
@@ -47,15 +47,19 @@ export default function BlockToolbarPopover( {
 		isToolbarForcedRef.current = false;
 	} );
 
+	// If the block has a parent with __experimentalCaptureToolbars enabled,
+	// the toolbar should be positioned over the topmost capturing parent.
+	const clientIdToPositionOver = capturingClientId || clientId;
+
 	const popoverProps = useBlockToolbarPopoverProps( {
 		contentElement: __unstableContentRef?.current,
-		clientId: capturingClientId || clientId,
+		clientId: clientIdToPositionOver,
 	} );
 
 	return (
 		! isTyping && (
 			<BlockPopover
-				clientId={ capturingClientId || clientId }
+				clientId={ clientIdToPositionOver }
 				bottomClientId={ lastClientId }
 				className={ clsx( 'block-editor-block-list__block-popover', {
 					'is-insertion-point-visible': isInsertionPointVisible,
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -95,7 +95,12 @@ function rectIntersect( rect1, rect2 ) {
  * @return {boolean} Whether the element is visible.
  */
 function isElementVisible( element ) {
-	const style = window.getComputedStyle( element );
+	const viewport = element.ownerDocument.defaultView;
+	if ( ! viewport ) {
+		return false;
+	}
+
+	const style = viewport.getComputedStyle( element );
 	if (
 		style.display === 'none' ||
 		style.visibility === 'hidden' ||
@@ -140,6 +145,11 @@ function isElementVisible( element ) {
  * @return {DOMRect} Bounding client rect.
  */
 export function getVisibleBoundingRect( element ) {
+	const viewport = element.ownerDocument.defaultView;
+	if ( ! viewport ) {
+		return new window.DOMRect();
+	}
+
 	let bounds = element.getBoundingClientRect();
 
 	const stack = [ element ];
@@ -158,8 +168,8 @@ export function getVisibleBoundingRect( element ) {
 	const viewportRect = new window.DOMRect(
 		0,
 		0,
-		window.innerWidth,
-		window.innerHeight
+		viewport.innerWidth,
+		viewport.innerHeight
 	);
 	return rectIntersect( bounds, viewportRect );
 }
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -57,3 +57,101 @@ export function getBlockClientId( node ) {
 
 	return blockNode.id.slice( 'block-'.length );
 }
+
+/**
+ * Returns the union of two DOMRect objects.
+ *
+ * @param {DOMRect} rect1 First rectangle.
+ * @param {DOMRect} rect2 Second rectangle.
+ * @return {DOMRect} Union of the two rectangles.
+ */
+export function rectUnion( rect1, rect2 ) {
+	const left = Math.min( rect1.left, rect2.left );
+	const top = Math.min( rect1.top, rect2.top );
+	const right = Math.max( rect1.right, rect2.right );
+	const bottom = Math.max( rect1.bottom, rect2.bottom );
+	return new window.DOMRect( left, top, right - left, bottom - top );
+}
+
+/**
+ * Returns the intersection of two DOMRect objects.
+ *
+ * @param {DOMRect} rect1 First rectangle.
+ * @param {DOMRect} rect2 Second rectangle.
+ * @return {DOMRect} Intersection of the two rectangles.
+ */
+function rectIntersect( rect1, rect2 ) {
+	const left = Math.max( rect1.left, rect2.left );
+	const top = Math.max( rect1.top, rect2.top );
+	const right = Math.min( rect1.right, rect2.right );
+	const bottom = Math.min( rect1.bottom, rect2.bottom );
+	return new window.DOMRect( left, top, right - left, bottom - top );
+}
+
+/**
+ * Returns whether an element is visible.
+ *
+ * @param {Element} element Element.
+ * @return {boolean} Whether the element is visible.
+ */
+function isElementVisible( element ) {
+	const style = window.getComputedStyle( element );
+	if (
+		style.display === 'none' ||
+		style.visibility === 'hidden' ||
+		style.opacity === '0'
+	) {
+		return false;
+	}
+
+	const bounds = element.getBoundingClientRect();
+	return (
+		bounds.width > 0 &&
+		bounds.height > 0 &&
+		bounds.right >= 0 &&
+		bounds.bottom >= 0 &&
+		bounds.left <= window.innerWidth &&
+		bounds.top <= window.innerHeight
+	);
+}
+
+/**
+ * Returns the rect of the element that is visible in the viewport.
+ *
+ * Visible nested elements, including elements that overflow the parent, are
+ * taken into account. The returned rect is clipped to the viewport.
+ *
+ * This function is useful for calculating the visible area of a block that
+ * contains nested elements that overflow the block, e.g. the Navigation block,
+ * which can contain overflowing Submenu blocks.
+ *
+ * The returned rect is suitable for passing to the Popover component to
+ * position the popover relative to the visible area of the block.
+ *
+ * @param {Element} element Element.
+ * @return {DOMRect} Bounding client rect.
+ */
+export function getVisibleBoundingRect( element ) {
+	let bounds = element.getBoundingClientRect();
+
+	const stack = [ element ];
+	let currentElement;
+
+	while ( ( currentElement = stack.pop() ) ) {
+		for ( const child of currentElement.children ) {
+			if ( isElementVisible( child ) ) {
+				const childBounds = child.getBoundingClientRect();
+				bounds = rectUnion( bounds, childBounds );
+				stack.push( child );
+			}
+		}
+	}
+
+	const viewportRect = new window.DOMRect(
+		0,
+		0,
+		window.innerWidth,
+		window.innerHeight
+	);
+	return rectIntersect( bounds, viewportRect );
+}
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -105,14 +105,22 @@ function isElementVisible( element ) {
 	}
 
 	const bounds = element.getBoundingClientRect();
-	return (
-		bounds.width > 0 &&
-		bounds.height > 0 &&
-		bounds.right >= 0 &&
-		bounds.bottom >= 0 &&
-		bounds.left <= window.innerWidth &&
-		bounds.top <= window.innerHeight
-	);
+
+	if ( bounds.width === 0 || bounds.height === 0 ) {
+		return false;
+	}
+
+	// wp-components visually hides elements by setting their width and
+	// height to 1px and positioning them off-screen.
+	if (
+		bounds.width === 1 &&
+		bounds.height === 1 &&
+		( bounds.left < 0 || bounds.top < 0 )
+	) {
+		return false;
+	}
+
+	return true;
 }
 
 /**
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -74,21 +74,6 @@ export function rectUnion( rect1, rect2 ) {
 }
 
 /**
- * Returns the intersection of two DOMRect objects.
- *
- * @param {DOMRect} rect1 First rectangle.
- * @param {DOMRect} rect2 Second rectangle.
- * @return {DOMRect} Intersection of the two rectangles.
- */
-function rectIntersect( rect1, rect2 ) {
-	const left = Math.max( rect1.left, rect2.left );
-	const top = Math.max( rect1.top, rect2.top );
-	const right = Math.min( rect1.right, rect2.right );
-	const bottom = Math.min( rect1.bottom, rect2.bottom );
-	return new window.DOMRect( left, top, right - left, bottom - top );
-}
-
-/**
  * Returns whether an element is visible.
  *
  * @param {Element} element Element.
@@ -129,22 +114,22 @@ function isElementVisible( element ) {
 }
 
 /**
- * Returns the rect of the element that is visible in the viewport.
+ * Returns the rect of the element including all visible nested elements.
  *
  * Visible nested elements, including elements that overflow the parent, are
- * taken into account. The returned rect is clipped to the viewport.
+ * taken into account.
  *
  * This function is useful for calculating the visible area of a block that
  * contains nested elements that overflow the block, e.g. the Navigation block,
  * which can contain overflowing Submenu blocks.
  *
- * The returned rect is suitable for passing to the Popover component to
- * position the popover relative to the visible area of the block.
+ * The returned rect represents the full extent of the element and its visible
+ * children, which may extend beyond the viewport.
  *
  * @param {Element} element Element.
- * @return {DOMRect} Bounding client rect.
+ * @return {DOMRect} Bounding client rect of the element and its visible children.
  */
-export function getVisibleBoundingRect( element ) {
+export function getVisibleElementBounds( element ) {
 	const viewport = element.ownerDocument.defaultView;
 	if ( ! viewport ) {
 		return new window.DOMRect();
@@ -165,11 +150,5 @@ export function getVisibleBoundingRect( element ) {
 		}
 	}
 
-	const viewportRect = new window.DOMRect(
-		0,
-		0,
-		viewport.innerWidth,
-		viewport.innerHeight
-	);
-	return rectIntersect( bounds, viewportRect );
+	return bounds;
 }
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -85,27 +85,21 @@ function isElementVisible( element ) {
 		return false;
 	}
 
+	// Check for <VisuallyHidden> component.
+	if ( element.classList.contains( 'components-visually-hidden' ) ) {
+		return false;
+	}
+
+	const bounds = element.getBoundingClientRect();
+	if ( bounds.width === 0 || bounds.height === 0 ) {
+		return false;
+	}
+
 	const style = viewport.getComputedStyle( element );
 	if (
 		style.display === 'none' ||
 		style.visibility === 'hidden' ||
 		style.opacity === '0'
-	) {
-		return false;
-	}
-
-	const bounds = element.getBoundingClientRect();
-
-	if ( bounds.width === 0 || bounds.height === 0 ) {
-		return false;
-	}
-
-	// wp-components visually hides elements by setting their width and
-	// height to 1px and positioning them off-screen.
-	if (
-		bounds.width === 1 &&
-		bounds.height === 1 &&
-		( bounds.left < 0 || bounds.top < 0 )
 	) {
 		return false;
 	}
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -112,16 +112,11 @@ function isElementVisible( element ) {
 		return false;
 	}
 
-	const style = viewport.getComputedStyle( element );
-	if (
-		style.display === 'none' ||
-		style.visibility === 'hidden' ||
-		style.opacity === '0'
-	) {
-		return false;
-	}
-
-	return true;
+	return element.checkVisibility( {
+		opacityProperty: true,
+		contentVisibilityAuto: true,
+		visibilityProperty: true,
+	} );
 }
 
 /**
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -92,7 +92,7 @@ export function rectUnion( rect1, rect2, containerRect ) {
 	 * horizontal limits of the container in which an element is supposed to be "visible".
 	 * For example, if an element is positioned -10px to the left of the window x value (0),
 	 * this function discounts the negative overhang because it's not visible and
-	 * therefore to be counted in the visible calculations.
+	 * therefore not to be counted in the visibility calculations.
 	 * Top and bottom values are not accounted for to accommodate vertical scroll.
 	 */
 	if ( containerRect ) {
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -58,13 +58,6 @@ export function getBlockClientId( node ) {
 	return blockNode.id.slice( 'block-'.length );
 }
 
-/*
- * Cache for rectUnion() return values.
- * When popovers are present on the canvas,
- * this calculation function is called
- * multiple times with the same values.
- */
-const rectUnionCache = new Map();
 /**
  * Calculates the union of two rectangles, and optionally constrains this union within a containerRect's
  * left and right values.
@@ -76,12 +69,6 @@ const rectUnionCache = new Map();
  * @return {DOMRect} Union of the two rectangles.
  */
 export function rectUnion( rect1, rect2, containerRect ) {
-	const cacheKey = JSON.stringify( { rect1, rect2, containerRect } );
-
-	if ( rectUnionCache.has( cacheKey ) ) {
-		return rectUnionCache.get( cacheKey );
-	}
-
 	let left = Math.min( rect1.left, rect2.left );
 	let right = Math.max( rect1.right, rect2.right );
 	const bottom = Math.max( rect1.bottom, rect2.bottom );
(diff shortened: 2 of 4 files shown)


noisysocks:
Changing anything to do with the popover usually carries a pretty high risk of unexpected positioning regressions so I wouldn't backport this to 6.6.

packages/block-editor/src/components/block-tools/block-toolbar-popover.js
```diff
@@ -49,7 +49,7 @@ export default function BlockToolbarPopover( {
 
 	const popoverProps = useBlockToolbarPopoverProps( {
 		contentElement: __unstableContentRef?.current,
-		clientId,
+		clientId: capturingClientId || clientId,
```
tellthemachines: Might be good to leave a comment about what `capturingClientId` actually is because it's not super clear even from reading the rest of the file 😅

tellthemachines (commented):
This is looking pretty good!

Before:
Screenshot 2024-06-21 at 3 03 41 PM

After:
Screenshot 2024-06-21 at 3 18 09 PM


One thing that feels a bit unexpected is the toolbar doesn't disappear when scrolling down the page:
Screenshot 2024-06-21 at 3 19 01 PM

I can reproduce it with several blocks in this PR, so something might need tweaking here.

noisysocks:
> One thing that feels a bit unexpected is the toolbar doesn't disappear when scrolling down the page:

Good catch! My `isElementVisible` function was too aggressive and filtered out children that weren't on the viewport because of scrolling. Should be fixed now.

tellthemachines:
Hmm, I'm still seeing the toolbar moving along the page when scrolling:


https://github.com/WordPress/gutenberg/assets/8096000/d613a703-58b8-4be0-b5b1-9cede4beee22

noisysocks:
Forgot to ping @jasmussen 🙂 this all came from his suggestion in https://github.com/WordPress/gutenberg/issues/40382#issuecomment-2149007134.

Oops, thanks @tellthemachines. I think I misread your initial comment and fixed something different. Will look into this later.

jasmussen:
Thanks for working on this! 

At a glance this seems a solid improvement, with no side-effects I can see. Can you think of any potential side-effects/downsides?

packages/block-editor/src/components/block-tools/block-toolbar-popover.js
```diff
@@ -47,15 +47,19 @@ export default function BlockToolbarPopover( {
...
 	} );
 
+	// If the block has a parent with __experimentalCaptureToolbars enabled,
+	// the toolbar should be positioned over the topmost capturing parent.
+	const clientIdToPositionOver = capturingClientId || clientId;
+
 	const popoverProps = useBlockToolbarPopoverProps( {
 		contentElement: __unstableContentRef?.current,
-		clientId,
+		clientId: clientIdToPositionOver,
```
ramonjd, line 56: > Hmm, I'm still seeing the toolbar moving along the page when scrolling:

I tried to force `shift: false`, and it does have an effect for some blocks, but for some reason the header and its contents are still sticky in this branch:


https://github.com/user-attachments/assets/43408c76-b867-481b-b8d8-7751b39786a2

On trunk, all toolbars scroll out of view with the anchored element.

noisysocks:
I've rebased this and (I think) fixed the bug @tellthemachines noticed. @ramonjd reckon you could test? ❤️

ramonjd:
Scroll issue looks to be resolved. Good one!

https://github.com/user-attachments/assets/ca0e4a28-087e-4aac-a031-256124c8e312

On thing I noticed is that the toolbar align itself to the right - from what I can tell it's the inline `translateX` value of the `block-editor-block-list__block-popover` element.

https://github.com/user-attachments/assets/dfbafd9d-1e59-4bd0-8f18-14c7aff0af4a

It appears to affect `dir="rtl"` locales differently in that the toolbar for nav items are positioned to the absolute left (in trunk they're underneath the block)


| Trunk | This PR |
|--------|--------|
| Screenshot 2024-08-19 at 6 51 14 PM | Screenshot 2024-08-19 at 6 51 29 PM |

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,98 @@ export function getBlockClientId( node ) {
...
+/**
+ * Returns the union of two DOMRect objects.
+ *
+ * @param {DOMRect} rect1 First rectangle.
+ * @param {DOMRect} rect2 Second rectangle.
+ * @return {DOMRect} Union of the two rectangles.
+ */
+export function rectUnion( rect1, rect2 ) {
+	const left = Math.min( rect1.left, rect2.left );
+	const top = Math.min( rect1.top, rect2.top );
```
ramonjd: > On thing I noticed is that the toolbar align itself to the right - from what I can tell it's the inline translateX value of the block-editor-block-list__block-popover element.

Could be due to negative values being misinterpreted by the Popover component.

If I pass `0` in the case of negative values — `left = left < 0 ? 0 : left` — things look okay 🤔 

https://github.com/user-attachments/assets/203cadab-40be-44e7-9a26-ac7859114c5b

I'm not 100% on this. Smells like a Popover thing.

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,98 @@ export function getBlockClientId( node ) {
...
+ * @param {Element} element Element.
+ * @return {boolean} Whether the element is visible.
+ */
+function isElementVisible( element ) {
+	const viewport = element.ownerDocument.defaultView;
+	if ( ! viewport ) {
+		return false;
+	}
+
+	const style = viewport.getComputedStyle( element );
```
ramonjd: Can thisand `element.getBoundingClientRect()` below be destructured? 

I don't think it matters at all, it's readable as it is.

ramonjd (approved):
This PR is definitely an improvement on the current status. Good stuff.

The positioning would be a good follow up if it can't be done neatly here. I think it's expected, and easier to access the controls, when the toolbar positions itself near the selected block.

packages/block-editor/src/utils/dom.js
noisysocks: Hmm. Can you share the markup you have in the template?

ramonjd: Nothing out of the ordinary I think. Here's the template:

```html
<!-- wp:template-part {"slug":"header","theme":"emptytheme","tagName":"header"} /-->

<!-- wp:query {"queryId":1,"query":{"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","sticky":"","perPage":10}} -->
<div class="wp-block-query"><!-- wp:post-template -->
<!-- wp:post-title {"isLink":true} /-->

<!-- wp:post-excerpt /-->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```

And the header template part:

```html
<!-- wp:navigation {"ref":6} /-->

<!-- wp:site-title /-->

<!-- wp:site-tagline /-->
```

Does that help? 😄

noisysocks: Hmm there is probably some hidden element that makes `getVisibleElementBounds` return a bounds that's larger than it should be, but I'm having a hard time reproducing. Could you export your theme? Or come up with steps to reproduce that begin with installing Emptytheme?

packages/block-editor/src/utils/dom.js
noisysocks: It can't because they return fancy DOM objects that aren't vanilla JS objects.

packages/block-editor/src/utils/dom.js
noisysocks: Try the latest commit I just pushed @ramonjd. I updated our check for hidden elements to explicitly check for `<VisuallyHidden>` which I think is causing problems because it hides its content using `clip` which is hard to detect.

ramonjd: >  Could you export your theme? Or come up with steps to reproduce that begin with installing Emptytheme?

Hmm, it's still happening for me on Firefox, Chrome and Safari. 

It's just empty theme on a clean install. 

Head to the site editor and add a navigation block as the first element in the header.

https://github.com/user-attachments/assets/d5ddc1f0-c0e0-48b7-8a81-cb82d87d2089

aaronrobertshaw: Following Ramon's steps above I can also replicate the misaligned toolbar in Chrome, Firefox, Safari etc.


https://github.com/user-attachments/assets/ce39d2c0-bec8-4cf2-a059-34dace295446

ramonjd: Thanks @aaronrobertshaw 

If I get time, I might play around with disallowing negative values. I'm not super confident about that approach but hacking never hurt anyone, except maybe some major corporations, millions of consumers and certain small democracies.

aaronrobertshaw (commented):
I've taken this for a quick spin and in general it's testing pretty well for me, with the exception of the misaligned toolbar already [under discussion](https://github.com/WordPress/gutenberg/pull/62711#discussion_r1722964020).

I also did some smoke testing of other components based off the `BlockPopover` such as spacing visualizers and resizable cover popover. Didn't spot any issues using this PR, other than gremlins in the visualizers that already exist on trunk.

packages/block-editor/src/utils/dom.js
ramonjd: I'm not sure where the calculations are happening, but a `console.trace` confirms that that the `-1` value is running through these new functions

```
{left: -1, top: 0, width: 1781, height: 63}
rectUnion	@	dom.js:73
getVisibleElementBounds	@	dom.js:142
getBoundingClientRect	@	index.js:94
```


But after looking in the wrong place for ages, from what I can tell, it's because the selected submenu has a left value of `-1px` in its editor styles

https://github.com/WordPress/gutenberg/blob/ae2d23d890614d5d538b3fe25b9a12163d198687/packages/block-library/src/navigation-submenu/editor.scss#L20-L20

https://github.com/WordPress/gutenberg/blob/ae2d23d890614d5d538b3fe25b9a12163d198687/packages/block-library/src/navigation/style.scss#L197-L197

Remove these, and it's kosher.

https://github.com/user-attachments/assets/4f653b7a-0e01-476c-a481-3dd0ac208016

So, I suppose there are two choices as far as I can tell:

1. Remove the -1px rules on the submenu (the quickest fix). I tested [the bug it was trying to fix](https://github.com/WordPress/gutenberg/pull/35716) and I can't reproduce with the values removed, but it doesn't address the primary cause.

2. "Normalize" negative values in `rectUnion` (e.g.,round up to zero in the case of `x`) to calculate the **[visible element bounds](https://github.com/WordPress/gutenberg/blob/ae2d23d890614d5d538b3fe25b9a12163d198687/packages/block-editor/src/utils/dom.js#L142-L142)**. On the surface this makes the most sense since we're only interested in what's visible in the viewport.

ramonjd: > "Normalize" negative values in rectUnion (e.g.,round up to zero in the case of x) to calculate the visible element bounds. On the surface this makes the most sense since we're only interested in what's visible in the viewport.

<details>

<summary>This is what I was thinking</summary>

```diff
diff --git a/packages/block-editor/src/utils/dom.js b/packages/block-editor/src/utils/dom.js
index eef322959a..2504c7824b 100644
--- a/packages/block-editor/src/utils/dom.js
+++ b/packages/block-editor/src/utils/dom.js
@@ -65,12 +65,25 @@ export function getBlockClientId( node ) {
  * @param {DOMRect} rect2 Second rectangle.
  * @return {DOMRect} Union of the two rectangles.
  */
-export function rectUnion( rect1, rect2 ) {
-	const left = Math.min( rect1.left, rect2.left );
-	const top = Math.min( rect1.top, rect2.top );
-	const right = Math.max( rect1.right, rect2.right );
-	const bottom = Math.max( rect1.bottom, rect2.bottom );
-	return new window.DOMRect( left, top, right - left, bottom - top );
+export function rectUnion( rect1, rect2, viewportRect ) {
+	let left = Math.min( rect1.left, rect2.left );
+	let top = Math.min( rect1.top, rect2.top );
+	let right = Math.max( rect1.right, rect2.right );
+	let bottom = Math.max( rect1.bottom, rect2.bottom );
+
+	if ( viewportRect ) {
+		left = Math.max( left, viewportRect.left );
+		top = Math.max( top, viewportRect.top );
+		right = Math.min( right, viewportRect.right );
+		bottom = Math.min( bottom, viewportRect.bottom );
+	}
+
+	return new window.DOMRect(
+		left,
+		top,
+		right - left,
+		bottom - top
+	);
 }
 
 /**
@@ -130,7 +143,12 @@ export function getVisibleElementBounds( element ) {
 	}
 
 	let bounds = element.getBoundingClientRect();
-
+	const viewportRect = {
+		top: 0,
+		right: viewport.innerWidth,
+		bottom: viewport.innerHeight,
+		left: 0,
+	};
 	const stack = [ element ];
 	let currentElement;
 
@@ -138,7 +156,7 @@ export function getVisibleElementBounds( element ) {
 		for ( const child of currentElement.children ) {
 			if ( isElementVisible( child ) ) {
 				const childBounds = child.getBoundingClientRect();
-				bounds = rectUnion( bounds, childBounds );
+				bounds = rectUnion( bounds, childBounds, viewportRect );
 				stack.push( child );
 			}
 		}

```

</details>

Thoughts?

aaronrobertshaw: > Remove the -1px rules on the submenu

Without looking, I assume this is just to prevent the appearance of a double border along the window edge. That sounds like something we'd visually want to keep.

> "Normalize" negative values in rectUnion

In the absence of the first option. This seems like the winner by default 🤷 

> Thoughts?

Looks viable to me. There's a lot of `Math.min/max` going on, would a quick inline comment help our future selves grok why the viewport rect is used etc?

ramonjd: > would a quick inline comment help our future selves grok why the viewport rect is used etc?

Definitely. Thanks for the sanity check. I'll commit this change. Appears to work as intended.

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,132 @@ export function getBlockClientId( node ) {
...
 	return blockNode.id.slice( 'block-'.length );
 }
+
+/*
+ * Cache for rectUnion() return values.
+ * When popovers are present on the canvas,
+ * this calculation function is called
+ * multiple times with the same values.
+ */
+const rectUnionCache = new Map();
```
ramonjd: This is necessary (for now) because, if not cached, `rectUnion()` is called a few times every second.

On a page left untouched with a popover open, the call count can potentially reach `n`.

This cache reduces the times the internal logic of `rectUnion` runs dramatically.

Most of the functions up the chain that use this function are written with `useCallback` and `useMemo`, but it  indicates that the Popover component is updating all these values using `requestAnimationFrame`.

Presumably indirectly, via `useFloating` and/or `autoUpdate` from `floating-ui/@floating-ui/react-dom`

https://github.com/WordPress/gutenberg/blob/d712323418d5c0a41983aea18060330d9773fbf9/packages/components/src/popover/index.tsx#L291-L291

ramonjd:
I've added a commit that trims the return left/right values of `rectUnion` if they fall beyond the left and right bounds of a given container, in our case the window.

This logic applies only in `getVisibleElementBounds` because presumably, we only interested in the "visible" bounds.

Anyway, it works it seems. 

Without it, the popover might not position itself correctly if the selected element's right or left values are offscreen. See https://github.com/WordPress/gutenberg/pull/62711#issuecomment-2296034009

### LTR

https://github.com/user-attachments/assets/7d6d9381-f211-4591-bd76-33901ad5580f


https://github.com/user-attachments/assets/693d1bc4-d097-416e-8d27-693748b4a951


### RTL



https://github.com/user-attachments/assets/c6ab5da7-d111-4c46-a2ce-3f92e57b1743

https://github.com/user-attachments/assets/73e7cff2-4dca-4786-b23d-f78c0c723f6e

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,132 @@ export function getBlockClientId( node ) {
...
+	return result;
+}
+
+/**
+ * Returns whether an element is visible.
+ *
+ * @param {Element} element Element.
+ * @return {boolean} Whether the element is visible.
+ */
+function isElementVisible( element ) {
```
kevin940726, line 99: TIL that there's a newly available [`checkVisibility`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) API that we can use here.

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,132 @@ export function getBlockClientId( node ) {
...
+		0,
+		viewport.innerWidth,
+		viewport.innerHeight
+	);
+
+	const stack = [ element ];
+	let currentElement;
+
+	while ( ( currentElement = stack.pop() ) ) {
+		for ( const child of currentElement.children ) {
```
kevin940726, line 156: This means we have to iterate through every child element which could be a bit costly? I don't know if we can cut off some operations here 🤔.

An immediate thought would be asking the reference element (nav link) to declaratively specify the overflow element (submenu). We could potentially use a `data-popover-target` attribute to find the overflow elements and only perform this extra check if it's specified. I think this would also prevent any unexpected side effects and maintain backward compatibility. The downside is that sometimes it's not better to declare an extra attribute.

Another rough idea is to hook into the `floating-ui`'s [middleware](https://floating-ui.com/docs/middleware) system. Instead of traversing through every descendant, We only check for those that overflow by leveraging [`document.elementsFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementsFromPoint). An untested pseudo code looks something like:

```ts
const overflowDescendants: Middleware = {
  name: 'overflowDescendants',
  fn({
    x,
    y,
    elements: { reference },
    rects: { reference: refRect, floating: floatingRect },
  }) {
    // Get the four vertices of the floating element (block toolbar).
    const vertices = [
      [x, y],
      [x + floatingRect.width, y],
      [x, y + floatingRect.height],
      [x + floatingRect.width, y + floatingRect.height],
    ] as const;
    // Get all the elements at those points.
    const elements = vertices.flatMap((vertice) =>
      document.elementsFromPoint(...vertice)
    );
    // Filter out those who are not the descendants of the reference element (anchor block).
    const descendents = elements.filter(
      (element) =>
        reference !== element && (reference as HTMLElement).contains(element)
    );
    // Bail out if no elements are found.
    if (!descendents.length) {
      return {};
    }

    let minX = refRect.x;
    let maxX = refRect.x + refRect.width;
    let minY = refRect.y;
    let maxY = refRect.y + refRect.height;

    // Calculate the new union bounding rect.
    descendents.forEach((descendent) => {
      const rect = descendent.getBoundingClientRect();
      minX = Math.min(minX, rect.x);
      maxX = Math.max(maxX, rect.right);
      minY = Math.min(minY, rect.y);
      maxY = Math.max(maxY, rect.bottom);
    });

    return {
      reset: {
        rects: {
          floating: floatingRect,
          reference: {
            x: minX,
            y: minY,
            width: maxX - minX,
            height: maxY - minY,
          },
        },
      },
    };
  },
};
```

I'm not even sure if this is more performant, though 😅. We probably need some testing if this is even worth trying. This can be done in a separate explorative PR if needed too.

packages/block-editor/src/utils/dom.js
kevin940726: One possible problem would be that this cache is never cleared. So a long-running page with multiple popovers might cause memory leaks.

What if we use memoization rather than a global cache? [`memize`](https://github.com/aduth/memize) might be helpful here. We can also create and clear the memoization for each instance. This should also help us avoid the need to generate a cache key string using `JSON.stringify`.

kevin940726 (commented):
This certainly is an interesting problem 😅. I left some suggestions but they are not blocking, just throwing out some potential ideas.

ramonjd:
> One possible problem would be that this cache is never cleared. So a long-running page with multiple popovers might cause memory leaks.
> What if we use memoization rather than a global cache?

Very good point. I'll look into memoization. 👍🏻

packages/block-editor/src/utils/dom.js
ramonjd: Oh nice!

So we could replace this stuff:

```js
		style.display === 'none' ||
		style.visibility === 'hidden' ||
		style.opacity === '0'
```

With something like this?

```js
element.checkVisibility({
    opacityProperty: true,
    contentVisibilityAuto: true,
   visibilityProperty: true,
 })
```

packages/block-editor/src/utils/dom.js
ramonjd: Thanks for looking into it! Looks like a follow up to me 😄

packages/block-editor/src/utils/dom.js
ramonjd: Great idea, thanks for the feedback

packages/block-editor/src/utils/dom.js
kevin940726: Yep I think so! I think it also checks for the bounding box too but I'm not sure.

packages/block-editor/src/utils/dom.js
ramonjd: I tried a few rounds with memoizing the function, and it made things a little janky when scrolling the page, even with a `maxSize` to limit the cache size.

So I've taken out the optimization for now. It wasn't in response to any perceived performance issue, only a defensive tactic.

packages/block-editor/src/utils/dom.js
ramonjd: I tested this, comparing the long-hand way with `checkVisibility` and the return values match 100% of the time.

I'll update.

packages/block-editor/src/utils/dom.js
kevin940726: Is it possible that you're using the `DOMRect` as the input parameter which cannot be memoized? I think the simplest way is to just flatten the input parameters and use primitives like so:

```js
rectUnion(leftX, leftY, leftWidth, leftHeight, rightX, rightY, rightWidth, rightHeight)
```

It gets a little verbose but then they're all primitives which can be memoized accordingly.

That said, we don't need to do this at all if this is not the bottleneck until proven otherwise! 👍

ramonjd: >  we don't need to do this at all if this is not the bottleneck until proven otherwise

yeah, there's no demonstrated bottleneck, I was just surprised that this function was running 100,000s of times in a short period of time

aaronrobertshaw (approved):
I've retested this one and it is working as advertised. I can no longer replicate the misaligned toolbar.

Screenshot 2024-08-23 at 2 59 22 PM

I'm happy to give this an approval now but it would be great if @kevin940726 can give the code changes a final ✅

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,110 @@ export function getBlockClientId( node ) {
...
+	if ( ! viewport ) {
+		return false;
+	}
+
+	// Check for <VisuallyHidden> component.
+	if ( element.classList.contains( 'components-visually-hidden' ) ) {
+		return false;
+	}
+
+	const bounds = element.getBoundingClientRect();
```
kevin940726, line 110: Do we still need this if we're using `checkVisibility`?

kevin940726 (approved):
LGTM 👍. Thanks! 💯

packages/block-editor/src/utils/dom.js
ramonjd: Good question.

I believe so because this tests for elements that would otherwise be visible if not for the `0` dimensions.

E.g., for `<div id="test" style="width:0;height:0;">Test</div>`, `checkVisibility( ...options )` returns `true`.

https://drafts.csswg.org/cssom-view-1/#dom-element-checkvisibility

packages/block-editor/src/utils/dom.js
noisysocks: 🤯 where has this been my whole life

packages/block-editor/src/utils/dom.js
```diff
@@ -57,3 +57,110 @@ export function getBlockClientId( node ) {
...
+	 * therefore not to be counted in the visibility calculations.
+	 * Top and bottom values are not accounted for to accommodate vertical scroll.
+	 */
+	if ( containerRect ) {
+		left = Math.max( left, containerRect.left );
+		right = Math.min( right, containerRect.right );
+	}
+
+	return new window.DOMRect( left, top, right - left, bottom - top );
+}
```
noisysocks, line 91: The approach makes sense but I don't really think this belongs in `rectUnion`. In a previous commit we had `rectIntersect` which we could bring back.

packages/block-editor/src/utils/dom.js
noisysocks: Caching comes with its own performance concerns 😀 So long as `isElementVisible` is fast it should be okay. I ordered the checks in order of most performant to least performant.

noisysocks:
Thanks for landing this @ramonjd! ❤️

packages/block-editor/src/utils/dom.js
noisysocks: https://github.com/WordPress/gutenberg/pull/64855

annezazu:
Absolutely love to see this. Huge usability improvement :)

packages/block-editor/src/utils/dom.js
youknowriad: I'm not sure this is supported in all the browsers that we support. I keep getting errors when using Safari.

getdave: Same problem here

https://caniuse.com/?search=checkVisibility

kevin940726: Based on the [browserslist config](https://github.com/WordPress/gutenberg/blob/trunk/packages/browserslist-config/index.js) we support, we support the [last 2 versions of Safari](https://browsersl.ist/#q=%3E+1%25%0Alast+1+Android+versions%0Alast+1+ChromeAndroid+versions%0Alast+2+Chrome+versions%0Alast+2+Firefox+versions%0Alast+2+Safari+versions%0Alast+2+iOS+versions%0Alast+2+Edge+versions%0Alast+2+Opera+versions) (17.5, 17.4). Both versions support `checkVisibility` according to [caniuse](https://caniuse.com/mdn-api_element_checkvisibility).

Maybe the config we use is too loose if this is unexpected?

youknowriad: Safari only updates with MacOS updates for me. (stable MacOS) and I'm still on 17.3 for this reason. I think we might want to update our Safari browser list config to 3 or 4 versions because of this.

andrewserong: Good catch — yeah, I think we should handle calling this optionally. Even if we officially support the latest two Safari versions, we shouldn't be in a situation where we're breaking the editor in older Safari versions. Here's what I'm seeing when I select a block in the post editor in Safari 16.4.1:

image

I've put up a quick fix PR here: https://github.com/WordPress/gutenberg/pull/65069

t-hamano:
Due to this PR, it seems that when a block is scrollable, the block toolbar position moves with it. Could you help me find out the cause of this?

- https://github.com/WordPress/gutenberg/issues/66112

Issue #6868, opened 2018-05-21
Meta tags and categories are not displayed in the Gutenberg editor. How to fix?!


shamhaner:
Meta tags and categories are not displayed in the Gutenberg editor. How to fix?!

danielbachhuber:
Hi @shamhaner,

I'm not sure I follow. Could you share more details about your issue, including a screenshot if possible, your Gutenberg version, and any debugging steps you've tried?

youknowriad:
Are you talking about the default tags/categories or a custom taxonomy?
By default, the tags/categories taxonomies have a `show_in_rest` set to `true`. Make sure you didn't change this and if it's a custom taxonomy, make sure it's set to true as well.

shamhaner:
Taxonomy on the site works. In my case, it is not possible to specify a category and a tag in the new record; they are not displayed in the menu of the gutenberg editor. The classic editor has no problems with this, the categories and shortcuts are where they are assigned, everything works correctly and without problems. Please tell me in which file I should check the specified parameters in the themes of the Wordpress or the gutenberg plugin.

shamhaner:
It's decided. Plugin settings "iThemes Security" => "WordPress Tweaks" => expose "Default Access"  (Access to REST API data is left as default. Information including published posts, user details, and media library entries is available for public access.) => "Save Settings"

Issue #62686, opened 2024-06-20
Global Styles and Theme.json: Next Steps


youknowriad:
The Global Styles architecture and features is growing significantly and the complexity of both the implementation and the mental model is increasing due to the organic nature of the iteration process. It is a good time though to take a step back, take a look at the main issues we're having and how to address them while simplifying the mental model.

## Currently supported features:

 1. Themes provide a theme.json (styles an settings) file that contain what I'm going to call a **style variation**. It defines colors, typography, palettes ... for the active theme.
 2. The default "style variation" can be edited in the UI and saved in the wp_global_styles CPT in a single entry (per theme).
 3. Themes can also provide alternative "style variations" for their themes, by creating new files under the `styles` folder like `styles/red.json` or `styles/blue.json` 
 4. These variations are not editable in the UI but instead, you can actually copy their content and apply it to your theme in the the default CPT entry for the theme.
 5. In 6.2, we introduced the possibility to edit "block style variations" in the UI. So for instance a user can edit the "outline" style variation of a button (apply colors, typography...) and save this in the CPT entry under `styles.blocks.core/button.variations.outline`. In a way we're using the same theme.json structure and same global styles UI to edit a **style variation** that applies only to a given block.
 6. In 6.6, we're introducing a way to register these style variations that apply to one or multiple block types using files in `styles` folder and by using the top level `blockTypes` property to choose which block type they apply to. The structure is very similar to any style variation (default theme.json, red.json or blue.json)
 7. 6.6 also offers a way to register the exact same block specific style variations as nested keys (instead of separate files) in theme.json file (or red.json, blue.json) under `blocks.styles.variations`. The advantage here is that theme authors can define these variations within a parent theme style variation `red.json` file, so when the user picks the "red" style variation for his theme in the UI, it also changes the styles of the available style variations for the blocks. 
 8. The block specific style variations registered using the method in 6 or 7 are editable in the UI just like the default block styles variations (see **5**) and the changes are saved similarly. 

These are great features, but it's easy to see how the multiple ways to write and save these variations can start to become confusing for users, in addition to the complex mental model, here are some issues that were raised before related to this model and some that are more hypothetic issues and features that we'd want to support:

 - The issue that comes often, is that people lose their modification of the default style variations (theme.json) when they apply an alternative theme style variation (red.json) in the UI.
 - It is not possible to edit the alternative theme style variations in the UI and keep the changes, to go back and forth between the different variations. Any change is saved as the default variation.
[...]

PR #62465, opened 2024-06-11
Global Styles: Prevent duplicate block style variations CSS

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/62457

## What?

As the new section styles feature requires styles generated per application of block style variations so they can be nested, there's superfluous CSS generated in the frontend theme.json stylesheet. This PR removes that.

## Why?

Less CSS.

## How?

Adds option in theme json to opt out of style generation for block style variations. A similar option is added to the `useGlobalStylesOutput` hook in the editor.

## Testing Instructions

1. Register a custom block style variation e.g. through a theme.json partial: `/styles/partial.json`

```
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 2,
	"title": "Custom",
	"slug": "custom",
	"blockTypes": [ "core/group" ],
	"styles": {
		"color": {
			"background": "aliceblue"
		}
	}
}
```

2. In both editors add a group block and apply the `Custom` block style variation.
3. Inspect the group block and confirm there is only a single `is-style-custom-*` css class matched
4. Save the page or post and confirm the same on the frontend

Note: Core blocks with variations can also be used for testing, e.g. Button or Image. These blocks those also have other styles from the block library enqueued that would match. The removal of redundant CSS is clearer when testing with a custom variation.

| Before | After |
|---|---|
| Screenshot 2024-06-14 at 7 27 54 PM | Screenshot 2024-06-14 at 7 29 16 PM |

PR #62712, opened 2024-06-21
Section Styles: improve performance and conceptual consistency

aaronrobertshaw:
Related: 
- https://github.com/WordPress/wordpress-develop/pull/6868
- https://github.com/WordPress/wordpress-develop/pull/6873

## What?

Remains unchanged: the block style variations (JSON files within `styles/` **with** a `blockTypes` key).

Changes to the primary theme.json and theme style variations (JSON files within `styles/` **without** a `blockTypes`) key :

1. Move shared variations from `styles.blocks.variations` to `styles.variations`.
2. Remove `blockTypes` from `styles.variations`.
3. Do not register shared variations from these files.

Other changes:

- Instead of using the `wp_theme_json_data_*` filters to merge the data, uses the WP_Theme_JSON_Resolver and WP_Theme_JSON classes, as per [#6868](https://github.com/WordPress/wordpress-develop/pull/6868).

## Why?

- More conceptual consistency in overrides:
  - block-level variations (`styles.blocks.blockType.variations`) are higher priority than top-level variations (`styles.variations`)
  - from lower to higher priority: block registry < block style variation files < theme.json < user changes (including theme style variations) 
- Improved performance.

## Testing Instructions

1. Disable these duotone filters so they don't register block style variations by calling the theme.json resolver which doesn't happen in core. See this [comment](https://github.com/WordPress/gutenberg/pull/62712#issuecomment-2185787727) for issue found during testing.
```diff
diff --git a/lib/block-supports/duotone.php b/lib/block-supports/duotone.php
index 98791fe05c..df8aa39414 100644
--- a/lib/block-supports/duotone.php
+++ b/lib/block-supports/duotone.php
@@ -18,8 +18,8 @@ if ( class_exists( 'WP_Duotone' ) ) {
 	remove_action( 'wp_loaded', array( 'WP_Duotone', 'set_global_styles_presets' ) );
 	remove_action( 'wp_loaded', array( 'WP_Duotone', 'set_global_style_block_names' ) );
 }
-add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_styles_presets' ), 10 );
-add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_style_block_names' ), 10 );
+// add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_styles_presets' ), 10 );
+// add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_style_block_names' ), 10 );
 
 // Add classnames to blocks using duotone support.
 if ( function_exists( 'wp_render_duotone_support' ) ) {

```
2. Defines block style variations via the different sources below:
   - `register_block_style`
   - Theme.json partial file under `/styles` directory including `blockTypes`
   - Within a theme style variation under `styles.variations` 
       - Define at least one that is registered via `register_block_style` or a theme.json partial as well
       - Define a variation that hasn't been registered
3. Confirm that only registered variations appear for selection (those defined via `register_block_style`, theme.json partials, or in core block types `styles` property such as Button's Outline style)
4. Try out all the variations and ensure they get applied correctly in the editor and frontend
5. Update the variation definitions to define inner element and block type styles as well
6. Check the inner element and block type styles are applied correctly for variations
4. Test applying variations on blocks in a nested fashion
5. Update variation styles via Styles > Blocks > Block Type > Variation and confirm they apply correctly everywhere

<details>
<summary><strong>`register_block_style` Example</strong></summary>

```php
gutenberg_register_block_style(
	array( 'core/group', 'core/columns' ),
	array(
		'name'       => 'section-b',
		'label'      => __( 'Section B' ),
		'style_data' => array(
			'color'    => array(
				'background' => '#4f6f52',
				'text'       => '#d2e3c8',
			),
			'blocks'   => array(
				'core/group' => array(
					'color' => array(
						'background' => '#739072',
						'text'       => '#e3eedd',
					),
				),
			),
			'elements' => array(
				'link' => array(
					'color'  => array(
						'text' => '#ead196',
					),
					':hover' => array(
						'color' => array(
							'text' => '#ebd9b4',
						),
					),
				),
			),
		),
	)
);

```
</details>

<details>
<summary><strong>Theme.json Partial Example</strong></summary>

```json
{
	"version": 3,
	"title": "Section A",
	"slug": "section-a",
	"styles": {
		"color": {
			"background": "slategrey",
			"text": "snow"
		},
		"blocks": {
			"core/group": {
				"color": {
					"background": "darkslategrey",
					"text": "whitesmoke"
				}
			}
		}
	}
}
```
</details>

<details>
<summary><strong>Theme Style Variation Example (`styles.variations`)</strong></summary>

```json
{

	"styles": {
		"variations": {
			"section-a": {
				"color": {
					"background": "var(--wp--preset--color--theme-2)",
					"text": "var(--wp--preset--color--theme-5)"
				},
				"blocks": {
					"core/group": {
						"color": {
							"background": "var(--wp--preset--color--theme-4)",
							"text": "var(--wp--preset--color--theme-2)"
						}
					}
				}
			}
		}
	}
}
```
</details>

## Demo

https://github.com/WordPress/gutenberg/assets/60436221/5dafdd5e-e67e-4b8d-a680-4f71de469608


ramonjd:
I had to time box testing today, but can carry on later. 

 I went through the steps outlined in https://github.com/WordPress/gutenberg/pull/57908, taking into account the move to `styles.variations` and that I can't register shared variations in theme.json or theme variations.

### Block style variation JSON files under `/styles` 

These are files with `blockTypes` at the top level. 

<details>

<summary>Example</summary>

```json
{
	"version": 3,
	"title": "White fang",
	"blockTypes": [
		"core/group",
		"core/image",
		"core/columns"
	],
	"styles": {
		"color": {
			"background": "yellow"
		},
		"border": {
			"radius": "1111px",
			"width": "10px",
			"color": "red",
			"style": "solid"
		},
		"blocks": {
			"core/paragraph": {
				"border": {
					"width": "0"
				},
				"color": {
					"background": "aliceblue",
					"text": "purple"
				}
			}
		}
	}
}


```

</details>

I can see the variation buttons in the side bar and the styles are applied as expected in the editor and the frontend.

Nested block styles overwrite upper level styles as expected with the correct classnames (no duplicate classnames).

### Editing block style variations in the Site Editor

When editing a block style in Global styles, the variations are saved in the database under `styles.blocks[blockName].variations[variationName]`.

When editing the block styles, the editor updates in real time and the frontend reflects the new styles. 

Updating the block style for one block does not affect other blocks that may share the original style variation.

### Registering block style variations programmatically

Works as expected via `gutenberg_register_block_style`, similar to the findings above.

### Global styles revisions/style book

To confirm.

I'm not sure if this works yet, or if I'm doing something wrong. When switching to the Style Book or browsing past global styles revisions I'm not seeing any variation styles, even if changes to variations are saved in the CPT revisions.

aaronrobertshaw:
> Global styles revisions/style book

The preview of the styles doesn't seem to be working at first glance. Once I hit apply they seem to. I'll need to dig into that one.

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -790,6 +791,69 @@ public function __construct( $theme_json = array( 'version' => WP_Theme_JSON_Gut
...
+	 * {
+	 *   "styles": {
+	 *     "blocks": {
+	 *       "variations": {
+	 *         "blockTypes": [ "core/paragraph", "core/group" ],
+	 *         "color": { "background": "backgroundColor" }
+	 *       }
+	 *     }
+	 *   }
+	 * }
```
talldan: I noticed it doesn't seem to match what the function does (there's a `blocks` key and the variation name is missing). Here's an attempt at a fix:
```suggestion
	 * {
	 *   "styles": {
	 *     "variations": {
	 *       "light": {
	 *         "blockTypes": [ "core/paragraph", "core/group" ],
	 *         "color": { "background": "backgroundColor" }
	 *       }
	 *     }
	 *   }
	 * }
```

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -790,6 +791,69 @@ public function __construct( $theme_json = array( 'version' => WP_Theme_JSON_Gut
...
+	 * {
+	 *   "styles": {
+	 *     "blocks": {
+	 *       "variations": {
+	 *         "core/paragraph": { "color": { "background": "backgroundColor" } },
+	 *         "core/group": { "color": { "background": "backgroundColor" } }
+	 *       }
+	 *     }
+	 *   }
+	 * }
```
talldan: Similar here, my attempted fix:
```suggestion
	 * {
	 *   "styles": {
	 *     "blocks": {
	 *       "core/paragraph": {
	 *         "variations": {
	 *           "light": { "color": { "background": "backgroundColor" } }
	 *         }
	 *       },
	 *       "core/group": {
	 *         "variations": { 
	 *           "light": { "color": { "background": "backgroundColor" } }
	 *         }
	 *       }
	 *     }
	 *   }
	 * }
```

lib/class-wp-theme-json-gutenberg.php
aaronrobertshaw: Ah thanks, I'd only just spotted that copy and paste from https://github.com/WordPress/wordpress-develop/pull/6868. I've tweaked it and can double check its accuracy as this settles.

aaronrobertshaw:
Replication steps for the Global Styles revisions issue:

1. Edit some block style variations that have been applied to the page or choose a theme style variation that redefines the block style variations
2. Save the changes
3. Navigate to Global Styles > Revisions
4. Select a prior revision that should have block style variations styles
5. Note that the variation styles aren't applied in the editor
6. Click apply and note the variations are now applied


https://github.com/WordPress/gutenberg/assets/60436221/c6416315-7faa-4216-b5b2-027f86ddf771

oandregal:
Prepared backport at https://github.com/WordPress/wordpress-develop/pull/6873 to gauge performance impact.

lib/class-wp-theme-json-resolver-gutenberg.php
```diff
@@ -859,4 +868,79 @@ public static function resolve_theme_file_uris( $theme_json ) {
...
+	/**
+	 * Adds variations sourced from block style variations files to the supplied theme.json data.
+	 *
+	 * @since 6.6.0
+	 *
+	 * @param array $data   Array following the theme.json specification.
+	 * @param array $variations Shared block style variations.
+	 * @return array Theme json data including shared block style variation definitions.
+	 */
+	private static function inject_variations_from_block_style_variation_files( $data, $variations ) {
```
oandregal, line 884: There's a subtle bug in this method:

- The block `core/heading` has a variation called 'Mine'.
- A partial defines a variation 'Mine' for the blocks `core/paragraph` and `core/group`.
- With this code, the styles the partial defined for `core/paragraph` and `core/group` will also apply to `core/heading`.

We'd need to 1) register the variations defined in the partial at the block level (`styles.blocks.blockType.variations`) but 2) make sure the top-level variation of the `theme.json` still override them. The method `inject_variations_from_block_styles_registry` can serve as an example.

richtabor:
I spun up [a PR for Assembler](https://github.com/Automattic/themes/pull/7891) (a theme I'm using to test this extensively) to align with this proposal.

richtabor:
One change I noted is that the generated class is exclusively defined by the `title` from the style variation partial, whereas previously the slug was defined by the file name. Is that intentional? 

For example, this results in a class of `is-style-style-1`:

CleanShot 2024-06-21 at 16 24 05

While previously, it would have resulted in `is-style-section-1`.

richtabor:
I suppose I could set slug independently as well.

lib/class-wp-theme-json-resolver-gutenberg.php
aaronrobertshaw: So the bug was caused by the shift to omit the `blockTypes` prop?

aaronrobertshaw:
> One change I noted is that the generated class is exclusively defined by the title from the style variation partial, whereas previously the slug was defined by the file name. Is that intentional?

As far as I can tell this has changed at all.

In both `trunk` and this branch, the variation's class name was based off the title for variation partial theme.json files.

I also checked prior to each of the following PR commits
- https://github.com/WordPress/gutenberg/pull/62550
- https://github.com/WordPress/gutenberg/pull/62461

I went all the way back to the introduction of the feature in:
- https://github.com/WordPress/gutenberg/pull/57908

It behaved the same at each point for me. With no `slug` value in the partial and the title differing from the file name, the name the variation was registered under was the kebab-cased version of its title. At each point tested, it is the variation name used for the CSS class. 

Is it possible this was a misconception rather than an actual regression?

lib/class-wp-theme-json-gutenberg.php
```diff
@@ -790,6 +791,78 @@ public function __construct( $theme_json = array( 'version' => WP_Theme_JSON_Gut
...
+	 *
+	 * @param array $theme_json A structure that follows the theme.json schema.
+	 * @return array Theme json data with shared variation definitions unwrapped under appropriate block types.
+	 */
+	private static function unwrap_shared_block_style_variations( $theme_json ) {
+		if ( ! isset( $theme_json['styles']['variations'] ) ) {
+			return $theme_json;
+		}
+
+		$registered_styles = WP_Block_Styles_Registry::get_instance()->get_all_registered();
```
aaronrobertshaw: It doesn't appear that core block type variations (e.g. Button outline) get registered in the block styles registry.

We might be able to loop registered block types and check their `styles` property similar to what happens when determining valid variations for sanitization etc.

aaronrobertshaw: Perhaps a better option would be to move the call to `unwrap_shared_block_style_variations` to after the collection of `$valid_variations`, then pass those as a new param. They should then contain both the core block styles as well as any registered in the `WP_Block_Styles_Registry`.

What do you think?

aaronrobertshaw: In the interests of saving time, I've pushed a commit that I think does the job: https://github.com/WordPress/gutenberg/pull/62712/commits/bc758fce265e813bdd5b2c47a3f1cd8ae3147544.

aaronrobertshaw:
> One change I noted is that the generated class is exclusively defined by the title from the style variation partial, whereas previously the slug was defined by the file name. Is that intentional?

I think I've got an explanation for this one now.

Since the section styles feature was introduced, if a theme.json partial had a `title` property the kebab-cased version of that was used as the variation name and in turn formed the basis of the CSS class name. However, if a theme.json partial omitted the `title` (and the `slug` after that was introduced), the file name is used. 

This [fallback to the filename](https://github.com/WordPress/gutenberg/blob/trunk/lib/class-wp-theme-json-resolver-gutenberg.php#L781) is still in place, so I'm confident there's been no regression in this regard.

richtabor:
> Is it possible this was a misconception rather than an actual regression?

In the current version of the Assembler theme, where I'm testing this, I have "Style 1" visible labels, while using "section-1" formatting for the action variations. I don't think this is a regression, but a difference in how it works currently in trunk (whether that's intended or not). If a combination of using `slug` and `title` gets the same result, it's all good.

aaronrobertshaw:
>  I don't think this is a regression, but a difference in how it works currently in trunk

That's the catch though, for me, it works the same as on `trunk`. 

When the theme.json partials are parsed, if it doesn't have a `title` property it gets the base of the filename set to the property. You can see that code [here](https://github.com/WordPress/gutenberg/blob/trunk/lib/class-wp-theme-json-resolver-gutenberg.php#L781) which hasn't changed since https://github.com/WordPress/gutenberg/pull/46554 (about a year).

The only recent change around this functionality was introducing support for the `slug` property, which is optional and would take precedence over the `title`, if `slug` is present. That occurred in https://github.com/WordPress/gutenberg/pull/62550.

If there is a difference I haven't been able to replicate it, nor see it in code.

> If a combination of using slug and title gets the same result, it's all good.

👍 That's the main thing I guess. If you're happy, I'm happy.

ramonjd:
> The preview of the styles doesn't seem to be working at first glance. Once I hit apply they seem to. I'll need to dig into that one.

Passing `variationStyles: true` to `toStyles` seems to fix it

```js
		const globalStyles = toStyles(
			updatedConfig,
			blockSelectors,
			hasBlockGapSupport,
			hasFallbackGapSupport,
			disableLayoutStyles,
			options?.disableRootPadding,
			{
				variationStyles: true,
			}
		);
```

https://github.com/WordPress/gutenberg/blob/bc758fce265e813bdd5b2c47a3f1cd8ae3147544/packages/block-editor/src/components/global-styles/use-global-styles-output.js#L1366-L1373

The only question I have is whether `variationStyles` needs to always be `true`? I'm not sure, due to https://github.com/WordPress/gutenberg/pull/62465, which introduced `variationStyles`

I have a PR with the change over here:

aaronrobertshaw:
> The only question I have is whether variationStyles needs to always be true? I'm not sure, due to https://github.com/WordPress/gutenberg/pull/62465, which introduced variationStyles

There might be a catch with simply passing the `variationStyles` option at this stage. I'd expect it to generate generic `is-style-*` styles but not ones specific to an individual application of a block style variation. 

Passing that option should appear to work for simple uses of variations but would likely fall short when variations are applied in a nested fashion. For example, a light section that contains a dark section, then another light section within the dark one.

I'll take a look at https://github.com/WordPress/gutenberg/pull/62768 later today and double check what I've said. 

Thanks for putting together that fix so quickly 🚀 

_Update: Unfortunately, I could confirm my hunch was correct. More details here: https://github.com/WordPress/gutenberg/pull/62768#pullrequestreview-2134072392_

aaronrobertshaw:
The JS unit test failures here are related to the changes in this PR.

The intention was to omit `styles.variations` from the theme.json schema to reinforce the recommendation to use theme.json partials as much as possible given the potential for `styles.variations` to be deprecated soon.

If we stick to that plan we'll need to update the `theme-schema` test to ignore the test theme that covers `styles.variations`. It's that, or, `styles.variations` is kept in the schema for now.

What do people think?

lib/class-wp-theme-json-resolver-gutenberg.php
oandregal: No, the partial still has the `blockTypes` but this method doesn't use it. It just stores the variations found in partials in `styles.variations` instead of using the `blockTypes` to assign them to the specific blocks — that's the bug.

aaronrobertshaw: This function was updated to use `blockTypes` over the weekend in https://github.com/WordPress/gutenberg/pull/62712/commits/368de7c6b934ff6b3a97613b48c0dda319f141a7.

aaronrobertshaw: I also should have been clearer in my question earlier. I meant it more that, previously the shared definitions still had the `blockTypes` prop which was used in the unwrapping of those definitions into the individual block types' variations. When the unwrapping was updated to unwrap shared definitions based on what was registered in the registry instead of `blockTypes`, this function was missed.

aaronrobertshaw: There was also an issue that block styles contained in a block type's `styles` property, e.g. Button's Outline, weren't in the registry so shared definitions for them weren't unwrapped correctly.

I believe this has all been fix in recent days, so I'll resolve this thread pending further review.

andrewserong (commented):
I haven't had a close look at the code, but this is testing very nicely for me! The testing steps all worked for me, and here's a quick set of notes from my testing:

✅ theme.json partials are registered automatically and available in post and site editors, and output correctly
✅ Block styles nested within a partial are working correctly
✅ Nesting multiple blocks each using variations is working correctly
✅ Variations defined in the root `theme.json` file's `style.variations` are not registered automatically
✅ Variations defined in the root `theme.json` file that have been registered separately elsewhere have their styles applied correctly in the site editor and site frontend
✅ Updating block style variations in the site editor works correctly, and the updates are applied where expected

> It's that, or, styles.variations is kept in the schema for now.
> What do people think?

Keeping it in the schema as you've done in https://github.com/WordPress/gutenberg/pull/62712/commits/e90b542e70f1d4eece51de36e4e34931a9aa027c sounds good to me 👍

andrewserong:
Just had a quick look at the failing e2e tests, and they appear unrelated to this PR, as they're also failing over in a completely unrelated PR: https://github.com/WordPress/gutenberg/pull/62770.

aaronrobertshaw:
I've [removed](https://github.com/WordPress/gutenberg/pull/62712/commits/af5b1804863d368d08351ce192f06f25531b4f01) `title` and `blockTypes` from the shared definitions in the theme.json schema (i.e. `styles.variations`) given we now require prior registration of variations.

aaronrobertshaw:
Thanks for the testing @andrewserong 👍 

I've given this one another round of testing and I didn't notice any issues either. The extra unit tests around merging data give a nice bump in confidence as well.

A rough demo video, done after testing, has been added to the PR description for good measure.

aaronrobertshaw:
While testing the backport for these changes I noticed that the theme style variation based definitions of block style variations are not being applied. I believe this is due to a couple of [duotone filters](https://github.com/WordPress/gutenberg/blob/trunk/lib/block-supports/duotone.php#L21-L22) in Gutenberg which trigger registration of block style variations from theme.json partials before the variations endpoint in the global styles controller is executed.

I've updated the test instructions to include toggling these duotone filters off.

```diff
diff --git a/lib/block-supports/duotone.php b/lib/block-supports/duotone.php
index 98791fe05c..df8aa39414 100644
--- a/lib/block-supports/duotone.php
+++ b/lib/block-supports/duotone.php
@@ -18,8 +18,8 @@ if ( class_exists( 'WP_Duotone' ) ) {
 	remove_action( 'wp_loaded', array( 'WP_Duotone', 'set_global_styles_presets' ) );
 	remove_action( 'wp_loaded', array( 'WP_Duotone', 'set_global_style_block_names' ) );
 }
-add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_styles_presets' ), 10 );
-add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_style_block_names' ), 10 );
+// add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_styles_presets' ), 10 );
+// add_action( 'wp_loaded', array( 'WP_Duotone_Gutenberg', 'set_global_style_block_names' ), 10 );
 
 // Add classnames to blocks using duotone support.
 if ( function_exists( 'wp_render_duotone_support' ) ) {

```

aaronrobertshaw:
The fix for the issue noted above is to ensure the block style variations from partials were registered before the theme style variations are retrieved in the rest api endpoint.

I've tested and confirmed this fix is working while the duotone filters are disabled.


https://github.com/WordPress/gutenberg/assets/60436221/4ad9df23-eb9f-4f9d-9c1e-c2acf9f40e49

oandregal (approved):

oandregal:
I gave this some testing: tried things randomly, but also the core flows (layers override each other, users can update variations, etc.).

The direction is future-proof according to https://github.com/WordPress/gutenberg/issues/62686 and this approach removes any concerns about backwards compatibility. It also improves performance as per the [core PR's tests](https://github.com/WordPress/wordpress-develop/actions/runs/9641627253/job/26587676384?pr=6873).

oandregal:
There's a couple of playground failures. They are unrelated to this PR and happen in `trunk`, people are investigating. In the interest of merging this on time before WordPress 6.6 RC1, I'm going to force merge this PR when every other test is green.

oandregal:
Corresponding core PR has been merged as well https://github.com/WordPress/wordpress-develop/pull/6873 I was told we no longer need the label, as we are now only using it for tracking PRs that need package updates (we use the backport-changelog to cherry-pick the changes to the corresponding gutenberg branch).

PR #62716, opened 2024-06-21
Release automation: try cherry-picking automation

ellatrix:
## What?

The idea is to have a designated label `Backport to WP 6.6 Beta/RC` and once a PR against trunk is merged with that label, it automatically cherry-picks that PR into the right release branch, removes the label, and comments on the PR. It should also work if the label is applied after merge. If there is a conflict, it will comment on the PR to make a PR against the `wp/*` manually.

## Why?

Cherry-picking can be tedious, and it would be good to cherry-pick as soon as PRs are merged into trunk. That way conflicts can be detected and addressed early.

## How?

See the workflow for details in logic.

To see it in action, this is the latest version: https://github.com/WordPress/gutenberg/pull/62735#issuecomment-2182500147

Also, jobs will wait for the last one to finish so the checkout are always fresh.

image

## Testing Instructions

I will attempt to create two branches that are based on this one:
* [`wp/0.0`](https://github.com/WordPress/gutenberg/tree/wp/0.0), to test if PRs get cherry-picked.
* [`try/cherry-pick-automation-trunk`](https://github.com/WordPress/gutenberg/tree/try/cherry-pick-automation-trunk) to simulate trunk.

Then I will create a few PRs against `try/cherry-pick-automation-trunk` to test if they get cherry-picked:
* ✅ One applying the label before merge: #62731
* ✅ One applying the label after merge: https://github.com/WordPress/gutenberg/pull/62735#issuecomment-2182500147
* ✅ One with a conflict. It works: https://github.com/WordPress/gutenberg/pull/62724#issuecomment-2182323507

### DIFF
--- a/.github/workflows/cherry-pick-wp-release.yml
+++ b/.github/workflows/cherry-pick-wp-release.yml
@@ -11,30 +11,43 @@ jobs:
         runs-on: ubuntu-latest
         if: github.event.pull_request.merged == true
         steps:
+            - name: Determine if label should trigger cherry-pick
+              id: label-check
+              uses: actions/github-script@v7
+              with:
+                  script: |
+                      const labels = context.payload.pull_request.labels.map(label => label.name);
+                      console.log(`Labels: ${labels}`);
+                      const regex = /^Backport to WP ([0-9]+\.[0-9]+) Beta\/RC$/;
+                      let matched = false;
+                      for (const label of labels) {
+                        const match = label.match(regex);
+                        if (match) {
+                          const version = match[1];
+                          console.log(`Matched label: ${label}`);
+                          console.log(`Extracted version: ${version}`);
+                          core.exportVariable('cherry_pick', 'true');
+                          core.exportVariable('version', version);
+                          matched = true;
+                          break;
+                        }
+                      }
+                      if (!matched) {
+                        core.exportVariable('cherry_pick', 'false');
+                      }
+
             - name: Checkout repository
+              if: env.cherry_pick == 'true'
               uses: actions/checkout@v2
               with:
+                  token: ${{ secrets.GUTENBERG_TOKEN }}
                   fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


vcanales (approved):
LGTM. I like the requirement of creating PRs to the `wp/*` branch in order to merge changes that had caused conflicts.

desrosj:
This is really neat! Thanks for this, @ellatrix.

I did make one small adjustment to the syntax for the versions of third-party actions to use in #62836. When possible we should use [full commit SHAs as an extra level of hardening to be safe](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). The `actions/checkout` was also a few versions behind, but Dependabot took care of that for us.

ellatrix:
Sounds good, thanks!

gziolo:
This is awesome! One question, who decides about adding `Backport to WP 6.6 Beta/RC` label? On the technical level, the commit can always get reverted from the release branch if the release squad disagrees, so maybe it is not an issue at all.

gziolo:
It could also get extended to backporting to the Gutenberg plugin release branches when there is a bug fix to land during RC or after the final release. At that point, we could delete the old [bin/cherry-pick.mjs](https://github.com/WordPress/gutenberg/blob/trunk/bin/cherry-pick.mjs) script crafted by @adamziel long time ago 😀

ellatrix:
Yes, anyone who can add labels can add it. Th Editor Tech Leads should double check the commits before publishing packages, revert if they disagree with something. They would have to do this previously as well, but before cherry-picking. Tbh, I don't think it's a big deal, we should put trust in people and correct/ping if needed.

gziolo:
> Tbh, I don't think it's a big deal, we should put trust in people and correct/ping if needed.

Agreed. I guess it was the experience in the WP 6.6 release cycle so it's best to repeat it in the future 👍🏻

PR #62718, opened 2024-06-21
Eslint-plugin: Add method-signature-style TypeScript lint rule

sirreal:
## What?

Add the [method-signature-style](https://typescript-eslint.io/rules/method-signature-style) to the recommended TypeScript rule set.

## Why?

This subtle change is not obvious but can dramatically weaken (or strengthen) the type system.

```ts
type Rec = Record<string, unknown>;

interface Post extends Rec { status: string; }

interface Base<T> {
  isEligibleMethod(item: T): boolean;
  isEligibleFunc: (item: T) => boolean;
}

const base: Base<Rec> = {
  // Using the "method" type syntax above weakens the type checking.
  // This is not a type error!!!
  isEligibleMethod: (item: Post) => item.status.startsWith('foo'),

  // This is a type error as expected.
  isEligibleFunc: (item: Post) => item.status.startsWith('foo'),
}

// With method, we could trigger runtime type errors that seem obviously wrong:
base.isEligibleMethod({}); // Cannot read properties of undefined (reading 'startsWith')
```

See https://typescript-eslint.io/rules/method-signature-style

This was discovered in https://github.com/WordPress/gutenberg/pull/62647#discussion_r1647980168 with the help of @jsnajdr.

## How?

Add and enable the rule in the recommended TypeScript rule set.

## Testing Instructions

CI passes.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -36,6 +36,7 @@
 -   `CustomSelectControlV2`: fix trigger button font size. ([#63131](https://github.com/WordPress/gutenberg/pull/63131))
 -   Extract `TimeInput` component from `TimePicker` ([#60613](https://github.com/WordPress/gutenberg/pull/60613)).
 -   `TimeInput`: Add `label` prop ([#63106](https://github.com/WordPress/gutenberg/pull/63106)).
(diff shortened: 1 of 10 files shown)


youknowriad (approved):
👍 Really weird though :P

PR #62725, opened 2024-06-21
[Developer Documentation] fix link url

michakrapp:
## What?
Fix a link in the create your first block tutorial.

## Why?
Link is pointing to the wrong page.

## How?
Change the link url href.

## Testing Instructions

1. Open the page
2. Click the link
3. See that you are on the correct page (Get started with create-block)

Same as testing with the keyboard


juanmaguitar (approved):

juanmaguitar:
Good catch! Thanks for the fix 👍

PR #62730, opened 2024-06-21
Button block: Fix fatal error on WordPress 6.4

cbravobernal:
## What?

For at least one month, Gutenberg should be compatible with WordPress 6.4

A block bindings [commit](https://github.com/WordPress/gutenberg/commit/9c3dabc595101bd285723a4fac266c132fbd489c) added a render PHP for the button block. And is using a `WP_HTML_Tag_Processor` method that doesn't exist in 6.4.

6.5 compatibility Tag processor contains it.

## Screenshots or screencast 

Without the fix:
Screenshot 2024-06-21 at 11 58 59

With the fix:
Screenshot 2024-06-21 at 11 59 24

### DIFF
--- a/lib/compat/wordpress-6.4/html-api/class-gutenberg-html-tag-processor-6-4.php
+++ b/lib/compat/wordpress-6.4/html-api/class-gutenberg-html-tag-processor-6-4.php
@@ -2447,5 +2447,4 @@ class Gutenberg_HTML_Tag_Processor_6_4 {
 
 		return true;
 	}
-
 }
--- a/lib/compat/wordpress-6.4/html-api/class-gutenberg-html-tag-processor-6-4.php
+++ b/lib/compat/wordpress-6.4/html-api/class-gutenberg-html-tag-processor-6-4.php
@@ -2447,4 +2447,5 @@ class Gutenberg_HTML_Tag_Processor_6_4 {
 
 		return true;
 	}
+
 }
--- a/packages/block-library/src/button/index.php
+++ b/packages/block-library/src/button/index.php
@@ -18,6 +18,7 @@
  */
 function render_block_core_button( $attributes, $content ) {
 	$p = new Gutenberg_HTML_Tag_Processor_6_5( $content );
+
 	/*
 	 * The button block can render an `<a>` or `<button>` and also has a
 	 * `<div>` wrapper. Find the a or button tag.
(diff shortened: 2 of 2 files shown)


gziolo:
If there was a fatal error, then it means that no site can currently run Gutenberg with WordPress 6.4, so we should simply bump the minimum required version to WP 6.5.

cbravobernal:
> If there was a fatal error, then it means that no site can currently run Gutenberg with WordPress 6.4, so we should simply bump the minimum required version to WP 6.5.

It has been 2 weeks since that error and nobody complaint, so I guess that bumping the version is a quick win.

SantosGuillamot:
It "only" happens when there is a button on the page, but I assume that's a pretty common scenario. Bumping the minimum required version seems logical to me.

Apart from that, with the current implementation, if I have an empty button, it won't remove it in 6.4, right? Which is still a downgrade from the implementation previous to that commit.

Anyway, for me, it's fine to add this check and remove it once 6.6 is out and 6.4 is not supported anymore. Whatever you decide.

cbravobernal:
> Apart from that, with the current implementation, if I have an empty button, it won't remove it in 6.4, right? Which is still a downgrade from the implementation previous to that commit.

Yes, as you mention, an empty button will appear.

Screenshot 2024-06-21 at 14 54 22

But still, it would be similar to what the user has in their editor screen, having parity between them.

Screenshot 2024-06-21 at 14 57 45

Has the "removing the button if empty" approach been discussed with design?

SantosGuillamot:
> Has the "removing the button if empty" approach been discussed with design?

I think that logic was added as a result of this issue: [link](https://github.com/WordPress/gutenberg/issues/17221).

dmsnell:
I would assume that if we rely on that code we would want to or need to backport the Core changes into Gutenberg. That's probably my fault for not keeping up with that.

We would copy the Tag Processor class into Gutenberg and call it `Gutenberg_HTML_Tag_Processor_6_5`. We could then rely on its methods.

On the other hand we're free to wait until we're _two versions ahead_ of the new methods so we don't have to do any backfilling. Either way: patience, or porting, should be fine.

SantosGuillamot:
It seems this has been solved in [this other pull request](https://github.com/WordPress/gutenberg/pull/63082), right?

talldan:
Yes, sorry, I didn't realize this one existed.

cbravobernal:
Let's close this one. @dmsnell also updated all HTML functions compat here:
https://github.com/WordPress/gutenberg/pull/63089

PR #62733, opened 2024-06-21
CustomSelectControl V2: fix setting initial value and reacting to external controlled updates

ciampo:
## What?

Supersedes #62308
Depends on #62255

Fixes how the `CustomSelectControlV2` legacy adapter behaves around its initial value (and consequent firing of the `onChange` callback)

## Why?

- Align the behaviour of v2 legacy adapter as closely as possible to the v1
- Fixing a bug where the initial value was not always correctly set and external controlled updates were not correctly applied

## How?

Pass explicit `value` and `defaultValue` to the `useStore` call in the legacy adapter. This has two consequences:
- by setting `value` explicitly, we fix the controlled updates bug
- by setting `defaultValue` explicitly to the first option, we don't rely on Ariakit's internal automatic selection and thus we avoid firing the `onChange` callback on initial render

I edited existing unit tests and added a few more in both the v1 and the v2 legacy adapter to cover those changes and ensure that the two components behave the same.

## Testing Instructions

- Read and code review the changes in the unit tests, making sure that they reflect the correct expected behavior
- Make sure that the unit tests pass
- Undo this PR's changes to `packages/components/src/custom-select-control-v2/legacy-component/index.tsx`
- Run unit tests, check that they fail

Additionally, check testing instructions from #62308

### DIFF
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -292,33 +292,21 @@ describe.each( [
 			} )
 		);
 
-		// NOTE: legacy CustomSelectControl doesn't fire onChange
-		// at this point in time.
-		expect( mockOnChange ).toHaveBeenNthCalledWith(
-			1,
-			expect.objectContaining( {
-				inputValue: '',
-				isOpen: false,
-				selectedItem: { key: 'flower1', name: 'violets' },
-				type: '',
-			} )
-		);
-
 		await click(
 			screen.getByRole( 'option', {
 				name: 'aquamarine',
 			} )
 		);
 
-		expect( mockOnChange ).toHaveBeenNthCalledWith(
-			2,
+		expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+		expect( mockOnChange ).toHaveBeenLastCalledWith(
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
 				selectedItem: expect.objectContaining( {
 					name: 'aquamarine',
 				} ),
-				type: '',
+				type: expect.any( String ),
 			} )
 		);
 	} );
@@ -336,23 +324,11 @@ describe.each( [
 			} )
 		).toHaveFocus();
 
-		// NOTE: legacy CustomSelectControl doesn't fire onChange
-		// at this point in time.
-		expect( mockOnChange ).toHaveBeenNthCalledWith(
-			1,
-			expect.objectContaining( {
-				selectedItem: expect.objectContaining( {
-					key: 'flower1',
-					name: 'violets',
-				} ),
-			} )
-		);
-
 		await type( 'p' );
 		await press.Enter();
 
-		expect( mockOnChange ).toHaveBeenNthCalledWith(
-			2,
+		expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+		expect( mockOnChange ).toHaveBeenLastCalledWith(
 			expect.objectContaining( {
 				selectedItem: expect.objectContaining( {
 					key: 'flower3',
@@ -387,10 +363,7 @@ describe.each( [
 
 		await click( optionWithCustomAttributes );
 
-		// NOTE: legacy CustomSelectControl doesn't fire onChange
-		// on first render, and so at this point in time `onChangeMock`
-		// would be called only once.
-		expect( onChangeMock ).toHaveBeenCalledTimes( 2 );
+		expect( onChangeMock ).toHaveBeenCalledTimes( 1 );
 		expect( onChangeMock ).toHaveBeenCalledWith(
 			expect.objectContaining( {
 				selectedItem: expect.objectContaining( {
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -83,11 +83,48 @@ const ControlledCustomSelectControl = ( {
 	);
 };
 
+it( 'Should apply external controlled updates', async () => {
+	const mockOnChange = jest.fn();
+	const { rerender } = render(
+		<UncontrolledCustomSelectControl
+			{ ...legacyProps }
+			value={ legacyProps.options[ 0 ] }
+			onChange={ mockOnChange }
+		/>
+	);
+
+	const currentSelectedItem = screen.getByRole( 'combobox', {
+		expanded: false,
+	} );
+
+	expect( currentSelectedItem ).toHaveTextContent(
+		legacyProps.options[ 0 ].name
+	);
+
+	expect( mockOnChange ).not.toHaveBeenCalled();
+
+	rerender(
+		<UncontrolledCustomSelectControl
+			{ ...legacyProps }
+			value={ legacyProps.options[ 1 ] }
+		/>
+	);
+
+	expect( currentSelectedItem ).toHaveTextContent(
+		legacyProps.options[ 1 ].name
+	);
+
+	// Necessary to wait for onChange to potentially fire
+	await sleep();
+
+	expect( mockOnChange ).not.toHaveBeenCalled();
+} );
+
 describe.each( [
 	[ 'Uncontrolled', UncontrolledCustomSelectControl ],
 	[ 'Controlled', ControlledCustomSelectControl ],
 ] )( 'CustomSelectControl (%s)', ( ...modeAndComponent ) => {
-	const [ mode, Component ] = modeAndComponent;
+	const [ , Component ] = modeAndComponent;
 
 	it( 'Should select the first option when no explicit initial value is passed without firing onChange', async () => {
 		const mockOnChange = jest.fn();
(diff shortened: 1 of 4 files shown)


packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -64,7 +64,7 @@ const ControlledCustomSelectControl = ( {
 	onChange: onChangeProp,
 	...restProps
 }: React.ComponentProps< typeof UncontrolledCustomSelectControl > ) => {
-	const [ value, setValue ] = useState( options[ 0 ] );
+	const [ value, setValue ] = useState( restProps.value ?? options[ 0 ] );
```
ciampo, line 67: This change allows the `Controlled` version of the component to use the `value` prop as the initial value (the same change was applied to the v1 component)

packages/components/src/custom-select-control/test/index.js
```diff
@@ -285,46 +337,29 @@ describe.each( [
...
 		const mockOnChange = jest.fn();
 
-		render(
-			<Component
-				{ ...props }
-				value={ props.options[ 0 ] }
-				onChange={ mockOnChange }
-			/>
-		);
+		render( <Component { ...props } onChange={ mockOnChange } /> );
```
ciampo, line 348: Align the test with the legacy v2 adapter (which doesn't set the `value` prop)

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -87,7 +87,73 @@ describe.each( [
...
+				onChange={ mockOnChange }
+				value={ legacyProps.options[ 3 ] }
+			/>
+		);
+
+		expect(
+			screen.getByRole( 'combobox', {
+				expanded: false,
+			} )
+		).toHaveTextContent( 'amber' );
```
tyxla: I find it inconsistent that when populating `value` we pass `legacyProps.options[ 3 ]`, but here, we directly reference `amber`. Should this also be:

```suggestion
		).toHaveTextContent( legacyProps.options[3].name );
```

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -87,7 +87,73 @@ describe.each( [
...
+
+	it( 'Should select the first option when no explicit initial value is passed without firing onChange', async () => {
+		const mockOnChange = jest.fn();
+		render( <Component { ...legacyProps } onChange={ mockOnChange } /> );
+
+		expect(
+			screen.getByRole( 'combobox', {
+				expanded: false,
+			} )
+		).toHaveTextContent( 'violets' );
```
tyxla: Just as my comment below - this might make sense to be:

```suggestion
		).toHaveTextContent( legacyProps.options[ 0 ].name );
```

We're also doing that in some of the other tests already, so it makes sense for consistency reasons.

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -87,7 +87,73 @@ describe.each( [
...
+
+		// Necessary to wait for onChange to potentially fire
+		await sleep();
+
+		expect( mockOnChange ).not.toHaveBeenCalled();
+	} );
+
+	// Using the "Uncontrolled" version of the component
+	// in order to apply controlled logic directly in the test
+	if ( mode === 'Uncontrolled' ) {
```
tyxla: Is there a good reason to keep it as a conditional test? Unless there's a solid benefit, I'd likely prefer moving the test out of this `describe.each()` group.

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -87,7 +87,73 @@ describe.each( [
...
+			expect( currentSelectedItem ).toHaveTextContent( 'violets' );
+
+			rerender(
+				<Component
+					{ ...legacyProps }
+					value={ legacyProps.options[ 1 ] }
+				/>
+			);
+
+			expect( currentSelectedItem ).toHaveTextContent( 'crimson clover' );
```
tyxla: Does it also make sense to verify that `onChange` was not called? Essentially like we do in the above tests:

```suggestion
			const mockOnChange = jest.fn();
			const { rerender } = render(
				<Component
					{ ...legacyProps }
					onChange={ mockOnChange }
					value={ legacyProps.options[ 0 ] }
				/>
			);

			const currentSelectedItem = screen.getByRole( 'combobox', {
				expanded: false,
			} );

			expect( currentSelectedItem ).toHaveTextContent( 'violets' );

			rerender(
				<Component
					{ ...legacyProps }
					onChange={ mockOnChange }
					value={ legacyProps.options[ 1 ] }
				/>
			);

			expect( currentSelectedItem ).toHaveTextContent( 'crimson clover' );

			await sleep();

			expect( mockOnChange ).not.toHaveBeenCalled();
```

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -292,33 +358,21 @@ describe.each( [
...
+		expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+		expect( mockOnChange ).toHaveBeenLastCalledWith(
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
 				selectedItem: expect.objectContaining( {
 					name: 'aquamarine',
 				} ),
-				type: '',
+				type: expect.any( String ),
```
tyxla, line 384: `type` will literally always be an empty string, though?

packages/components/src/custom-select-control/test/index.js
```diff
@@ -82,10 +82,62 @@ const ControlledCustomSelectControl = ( {
...
+
+	it( 'Should select the first option when no explicit initial value is passed without firing onChange', () => {
+		const mockOnChange = jest.fn();
+		render( <Component { ...props } onChange={ mockOnChange } /> );
+
+		expect(
+			screen.getByRole( 'button', {
+				expanded: false,
+			} )
+		).toHaveTextContent( 'violets' );
```
tyxla: Same as above in terms of whether to use a string literal or reference the test fixtures like we do in some of the tests:

```suggestion
		).toHaveTextContent( props.options[ 0 ].name );
```

packages/components/src/custom-select-control/test/index.js
```diff
@@ -82,10 +82,62 @@ const ControlledCustomSelectControl = ( {
...
+
+		expect( mockOnChange ).not.toHaveBeenCalled();
```
tyxla, line 128: No delay / sleep necessary?

This applies to the rest of the places where we're asserting whether `onChange` callback was called.

packages/components/src/custom-select-control/test/index.js
```diff
@@ -82,10 +82,62 @@ const ControlledCustomSelectControl = ( {
...
+				onChange={ mockOnChange }
+				value={ props.options[ 3 ] }
+			/>
+		);
+
+		expect(
+			screen.getByRole( 'button', {
+				expanded: false,
+			} )
+		).toHaveTextContent( 'amber' );
```
tyxla: Same as above:

```suggestion
		).toHaveTextContent( props.options[ 3 ].name );
```

packages/components/src/custom-select-control/test/index.js
```diff
@@ -82,10 +82,62 @@ const ControlledCustomSelectControl = ( {
...
+
+		expect( mockOnChange ).not.toHaveBeenCalled();
+	} );
+
+	// Using the "Uncontrolled" version of the component
+	// in order to apply controlled logic directly in the test
+	if ( mode === 'Uncontrolled' ) {
```
tyxla: Same as above in terms of moving this out of the `describe.each()`

packages/components/src/custom-select-control/test/index.js
```diff
@@ -82,10 +82,62 @@ const ControlledCustomSelectControl = ( {
...
+		it( 'Should apply external controlled updates', async () => {
+			const { rerender } = render(
+				<Component { ...props } value={ props.options[ 0 ] } />
+			);
+
+			const currentSelectedItem = screen.getByRole( 'button', {
+				expanded: false,
+			} );
+
+			expect( currentSelectedItem ).toHaveTextContent( 'violets' );
```
tyxla:
```suggestion
			expect( currentSelectedItem ).toHaveTextContent( props.options[ 0 ].name );
```

packages/components/src/custom-select-control/test/index.js
```diff
@@ -82,10 +82,62 @@ const ControlledCustomSelectControl = ( {
...
+
+			rerender( <Component { ...props } value={ props.options[ 1 ] } /> );
+
+			expect( currentSelectedItem ).toHaveTextContent( 'crimson clover' );
```
tyxla:
```suggestion
			expect( currentSelectedItem ).toHaveTextContent( props.options[ 1 ].name );
```

packages/components/src/custom-select-control/test/index.js
```diff
@@ -285,46 +337,29 @@ describe.each( [
...
+		expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+		expect( mockOnChange ).toHaveBeenLastCalledWith(
 			expect.objectContaining( {
 				inputValue: '',
 				isOpen: false,
 				selectedItem: expect.objectContaining( {
 					name: 'aquamarine',
 				} ),
-				type: '__item_click__',
+				type: expect.any( String ),
```
tyxla, line 370: To make sure I understand this change: it's for compatibility so the both versions of the tests would have the same expectation against the different implementations?

tyxla (approved):
This looks and works well for me, also by following the test instructions in #62308! 👍 

Thanks for adding some tests 🚀 

I've left some minor suggestions, but nothing seems to be blocking, so take them as optional ones.

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
ciampo: It will, but since we don't really document it in the v1, I thought that tests should be generic about what value to expect — as far as the attribute exists and it's a string. This also allows for the v1 and legacy v2 unit tests to be aligned on this particular assertion

packages/components/src/custom-select-control/test/index.js
ciampo: Yup! See reply in https://github.com/WordPress/gutenberg/pull/62733#discussion_r1649059793

packages/components/src/custom-select-control/test/index.js
ciampo: In my tests, it was not necessary for the v1 version, while it was unfortunately necessary for the ariakit-based version

ciampo:
All feedback addressed (either with code changes or replies). Will merge when CI checks pass. Thank you @tyxla for the review!

PR #62555, opened 2024-06-13
Interactivity: Fix `withScope` to restore scope on thrown exception inside async functions

DAreRodz:
## What?

Fixes `withScope()` to maintain the Interactivity scope on captured exceptions inside async functions.

Related to https://github.com/WordPress/gutenberg/pull/59708

## Why?

This was previously fixed inside the store proxy handlers but not inside `withScope()`, which contains the same logic.

## How?

Done so by 1) fixing the code inside `withScope()`, then 2) removing the duplicated code, and 3) using `withScope()` inside the store proxy handlers.

## Testing Instructions

Tested by e2e tests (inside `test/e2e/specs/interactivity/async-actions.spec.ts`).

PR #62734, opened 2024-06-21
Interactivity API: Refactor internal proxy and signals system

### LINKED ISSUE #63651: Interactivity API: Inability to use plain arrays in context
### Description

Reported by @otakupahp in https://github.com/WordPress/gutenberg/issues/60219#issuecomment-2233003274

---

Before 6.6 I had an array with other arrays:
```
array(
    array( 'id' => 1, 'item' => 'item 1 name' ),
    array( 'id' => 1, 'item' => 'item 2 name' ),
    array( 'id' => 1, 'item' => 'item 3 name' ),
);
```

That create this JSON: 
```
[
    {"id":1, "item":"item 1 name"},
    {"id":2, "item":"item 2 name"},
    {"id":3, "item":"item 3 name"},
]
```

The Interactive API created a context object that contains the array without problems.

Now, it just renders a string instead of an array object

To fix the error, I have to modify my array into this:

```
array(
    'data' => array(
        array( 'id' => 1, 'item' => 'item 1 name' ),
        array( 'id' => 1, 'item' => 'item 2 name' ),
        array( 'id' => 1, 'item' => 'item 3 name' ),
    )
);
```

Which created this JSON:
```
{
    "data": [
        {"id":1, "item":"item 1 name"},
        {"id":2, "item":"item 2 name"},
        {"id":3, "item":"item 3 name"},
    ]
}
```

This is indeed parsed as an object.

Unknown yet, @otakupahp needs to provide more details.

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

### LINKED ISSUE #63864: iAPI: Merge of state and context after client-side navigation
### Description

There's no way to distinguish the global state belonging to blocks that are inside a router region from the global state belonging to blocks that are not inside a router region. Therefore, we can't use the strategy of replacing the client state with the state coming from the server during client-side navigations.

1. Create a block that uses global state. 
2. Place that block outside of a router region. 
3. Navigate to a page where the global state is different from the initial page. 
4. The state of the block was updated, which it shouldn't have, since it's outside a router region.

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

DAreRodz:
## What?

The main goal of this refactoring is to have more control when defining state/context properties. The idea is to keep a unique [computed](https://github.com/preactjs/signals?tab=readme-ov-file#computedfn) for each of these properties *per scope*, which tracks changes even when its value is replaced with a getter. This way, properties can contain an initial value and be replaced with a getter later on or vice-versa, ensuring that directives update accordingly in this case. This is especially relevant for stores that are loaded asynchronously, like the `core/router` store from the `@wordpress/interactivity-router` module.

Also, we can take this opportunity to remove an external dependency ([deepsignal](https://github.com/luisherranz/deepsignal/)), keeping the code specific to the Interactivity API and making it more maintainable.

### Tasks

- [x]  Outline code structure
- [x]  Add tests to the state proxy handlers
- [x]  Ensure computeds work without scope
- [x]  Ensure objects maintain references between different namespaces
- [x] Add tests to the store proxy handlers
- [x] Add tests to `peek()`
- [x]  Review naming
- [x]  Review types and add TSDocs

### Next
I think these tasks should be addressed in a subsequent PR:
-  Use maps to store the context objects by namespace
-  Refactor context and related tests

## Testing Instructions

No testing instructions yet.


packages/interactivity/src/proxies/registry.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const supported = new Set( [ Object, Array ] );
+
+export const createProxy = < T extends object >(
+	namespace: string,
+	obj: T,
+	handlers: ProxyHandler< T >
+): T => {
+	if ( ! shouldProxy( obj ) ) {
+		throw Error( 'This object can be proxified.' );
```
michalczaplinski: Shouldn't this say `"can't be proxified"`? 🙂

packages/interactivity/src/proxies/registry.ts
```diff
@@ -0,0 +1,34 @@
...
+export const getProxy = < T extends object >( obj: T ): T =>
+	objToProxy.get( obj ) as T;
+
+export const getProxyNs = ( proxy: object ): string => proxyToNs.get( proxy )!;
+
+export const shouldProxy = ( val: any ): val is Object | Array< unknown > => {
+	if ( typeof val !== 'object' || val === null ) {
+		return false;
+	}
+	return ! ignore.has( val ) && supported.has( val.constructor );
```
michalczaplinski: Why do we need a separate `ignore` WeakSet here? Can't we re-use `proxyToNS` for this purpose instead and check if `proxyToNS.has(val)` instead?

packages/interactivity/src/directives.tsx
```diff
@@ -127,21 +130,18 @@ const proxifyContext = ( current: object, inherited: object = {} ): object => {
 /**
  * Recursively update values within a deepSignal object.
  *
- * @param target A deepSignal instance.
- * @param source Object with properties to update in `target`.
+ * @param proxy  A deepSignal instance.
```
michalczaplinski: We should probably remove references to `deepSignal` in the docstring.

packages/interactivity/src/proxies/store.ts
```diff
@@ -0,0 +1,56 @@
...
+	},
+};
+
+export const proxifyStore = < T extends object >(
+	namespace: string,
+	obj: T,
+	isNotRoot = false
+): T => {
+	const proxy = createProxy( namespace, obj, storeHandlers );
+	if ( proxy && ! isNotRoot ) {
```
michalczaplinski: Could we call the parameter `isRoot` and then change the `&& ! isNotRoot` to `&& isRoot` ?

packages/interactivity/src/proxies/state.ts
```diff
@@ -0,0 +1,200 @@
...
+const stateHandlers: ProxyHandler< object > = {
+	get( target: object, key: string, receiver: object ): any {
+		const desc = descriptor( target, key );
+
+		/*
+		 * If peeking, the property comes from the Object prototype, or the key
+		 * is a well-known symbol, then it should not be processed.
+		 */
+		if (
+			peeking ||
```
michalczaplinski, line 100: I might be misunderstanding something, but is this not a potential race condition? 

Can it happen that `peeking === true` because some key `A` is being peeked at, but at the very same time, the `get()` handler is invoked for some **different** key `B`?

michalczaplinski (commented):
I left a couple of comments on the technical side of things and pushed a minor commit too.

packages/interactivity/src/proxies/state.ts
luisherranz: Those operations are synchronous and, if they are nested, they should all peek. But feel free to add a test to verify that's the case 🙂

packages/interactivity/src/proxies/registry.ts
DAreRodz: Of course. 😂

packages/interactivity/src/proxies/registry.ts
DAreRodz: Yeah, we could reuse `proxyToNS`. I was considering changing the `ignore` prop, as no other entities but proxies are being "ignored" here.

packages/interactivity/src/directives.tsx
DAreRodz: Sure thing.

packages/interactivity/src/proxies/store.ts
DAreRodz: Actually, that was my first approach. I just changed it because I felt the default case would be creating a root object. I can change the value to be `true` by default, though. 🙂

packages/interactivity/src/proxies/state.ts
DAreRodz: Yup, it works like Luis mentioned here. ☝️

packages/interactivity/src/proxies/registry.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/1997db5db92672203bfd73e908322679071be0e6

packages/interactivity/src/proxies/registry.ts
DAreRodz: Done in https://github.com/WordPress/gutenberg/pull/62734/commits/1997db5db92672203bfd73e908322679071be0e6

packages/interactivity/src/proxies/store.ts
gziolo: Let's avoid the variable names with negations, as it's hard to reason about `! isNotRoot` in contrast to `isRoot`.

luisherranz:
@darerodz I have linked [this issue](https://github.com/WordPress/gutenberg/issues/63651) to be closed when this pull request is merged with a test that makes sure that plain arrays cannot be used at the root of the context.

packages/interactivity/src/proxies/registry.ts
michalczaplinski: Thanks! For very large stores that should save us a bit of memory! 🙂

packages/interactivity/src/proxies/state.ts
michalczaplinski: Gotcha, thanks guys!

packages/interactivity/src/proxies/index.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Internal dependencies
+ */
+export { proxifyState, peek } from './state';
+export { proxifyStore } from './store';
```
DAreRodz, line 5: Not sure about the naming here. I feel "proxify" is comprehensible, but maybe the appropriate word is "proxy", used as a verb.

What do you think? 🤔

packages/interactivity/src/init.ts
```diff
@@ -29,6 +30,10 @@ export const initialVdom = new WeakMap< Element, ComponentChild[] >();
 
 // Initialize the router with the initial DOM.
 export const init = async () => {
+	// Parse and populate the initial state and config.
+	const data = parseInitialData();
+	populateInitialData( data );
+
```
DAreRodz: I had to move the state initialization here to prevent issues with circular references between modules. ☝️

packages/interactivity/src/utils.ts
```diff
@@ -145,18 +145,26 @@ export function withScope( func: ( ...args: unknown[] ) => unknown ) {
...
+				} finally {
+					resetScope();
+					resetNamespace();
 				}
+
 				if ( it.done ) {
 					break;
 				}
 			}
+
```
DAreRodz, line 161: I introduced here the changes from https://github.com/WordPress/gutenberg/pull/62555.

packages/interactivity/src/proxies/store.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/5c17dd76651fb88620a4c48b4eb4ad45a6f11f5f.

DAreRodz:
I've marked the PR as ready for review.

packages/interactivity/src/store.ts
```diff
@@ -23,7 +12,10 @@ const deepMerge = ( target: any, source: any ) => {
 		for ( const key in source ) {
 			const getter = Object.getOwnPropertyDescriptor( source, key )?.get;
 			if ( typeof getter === 'function' ) {
-				Object.defineProperty( target, key, { get: getter } );
+				Object.defineProperty( target, key, {
+					get: getter,
+					configurable: true,
```
DAreRodz: This is required in case someone wants to overwrite a getter with a value or update the getter.

packages/interactivity/src/init.ts
luisherranz: Is that a real issue or just a Webpack warning?

packages/interactivity/src/store.ts
luisherranz: Interesting. Do we have a use case that justifies this possibility?

DAreRodz:
@sirreal has faced an issue in this PR with the initial state merge. After taking a look, I've found these two problems we should address:

1. I moved the state initialization inside the `init()` function to avoid circular dependency problems ([here](https://github.com/WordPress/gutenberg/pull/62734/files#r1689895609)). When that code is reached, blocks have already registered their stores, so there are getters inside the state, and now `deepMerge()` reads them from `target` without doing any checks. This was not happening before, so we hadn't noticed the issue. To fix this, we could move the code to its original place and solve the circular dependency issue instead.
2. I've changed how getters are defined so now they can be overwritten ([here](https://github.com/WordPress/gutenberg/pull/62734/files#r1691315133)). This is going to fail with the current `navigate` implementation, which makes the state from the server take precedence, overwriting getters with the values from the initial state. I haven't tried it yet, but I'm sure this happens. Not sure how this part should work, though. Maybe I can revert this change until we fix the [merge problems during `navigate`](https://github.com/WordPress/gutenberg/issues/63864).

cc: @luisherranz

packages/interactivity/src/proxies/signals.ts
```diff
@@ -0,0 +1,142 @@
+/**
+ * External dependencies
+ */
+import {
+	computed,
+	signal,
+	batch,
+	type Signal,
+	type ReadonlySignal,
+} from '@preact/signals-core';
```
luisherranz: Could we import everything from `@preact/signals` to avoid mixing packages?

```suggestion
} from '@preact/signals';
```

packages/interactivity/package.json
```diff
@@ -28,7 +28,7 @@
 	"types": "build-types",
 	"dependencies": {
 		"@preact/signals": "^1.2.2",
-		"deepsignal": "^1.4.0",
+		"@preact/signals-core": "^1.4.0",
```
luisherranz: Same question here.

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
+/* eslint-disable @typescript-eslint/no-shadow */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
+/**
+ * External dependencies
+ */
+import { effect } from '@preact/signals-core';
```
luisherranz: And here.

packages/interactivity/src/proxies/index.ts
luisherranz: I think proxify is fine 🙂

packages/interactivity/src/store.ts
```diff
@@ -280,13 +148,12 @@ export function store(
...
 		}
 		const rawStore = {
-			state: deepSignal( isObject( state ) ? state : {} ),
+			state: proxifyState( namespace, isObject( state ) ? state : {} ),
 			...block,
 		};
-		const proxiedStore = new Proxy( rawStore, handlers );
+		const proxiedStore = proxifyStore( namespace, rawStore );
 		rawStores.set( namespace, rawStore );
 		stores.set( namespace, proxiedStore );
```
luisherranz: `proxyfiedStore`? 🙂

```suggestion
		const proxifiedStore = proxifyStore( namespace, rawStore );
		rawStores.set( namespace, rawStore );
		stores.set( namespace, proxifiedStore );
```

luisherranz:
1. Two things:

- Why don't we fix the fact that `deepMerge` calls the getters? That feels wrong to start with.
- We should make a final decision on how stores are merged. My current take is that merging should work this way:
  - Initial server state
  - Call `#1` to `store` - replaces existing values
  - Call `#2` to `store` - replaces existing values
  - Call `#X…` to `store` - replaces existing values
  - New server state - doesn't replace existing values
  
  If that's the final decision, then the first thing that needs to happen is the initial server state.
2. If we decide that subsequent calls (`#2`, `#X…`) to `store` should replace existing values, then we need to make the getters replaceable. So let's revert the server merge before merging this PR and that should solve the issue, shouldn't it?

luisherranz:
Let's also see if we can get rid of the dependency on `DOMContentLoaded`, which has been causing some issues.

packages/interactivity/src/store.ts
```diff
@@ -1,20 +1,9 @@
...
-	resetScope,
-	getNamespace,
-	setNamespace,
-	resetNamespace,
-} from './hooks';
+import { proxifyState, proxifyStore } from './proxies';
+import { getNamespace } from './hooks';
+
 const isObject = ( item: unknown ): item is Record< string, unknown > =>
 	Boolean( item && typeof item === 'object' && item.constructor === Object );
```
luisherranz: What is the difference between this function and the one called `isPlainObject` in the `utils` file?

packages/interactivity/src/hooks.tsx
```diff
@@ -139,8 +139,15 @@ const namespaceStack: string[] = [];
...
  */
-export const getContext = < T extends object >( namespace?: string ): T =>
-	getScope()?.context[ namespace || getNamespace() ];
+export const getContext = < T extends object >( namespace?: string ): T => {
+	const scope = getScope();
+	if ( ! scope ) {
+		throw Error(
+			'Cannot call `getContext()` outside getters and actions used by directives.'
+		);
+	}
```
luisherranz: It’s not just getters and actions, so we should find a more generalistic way to say this and add possible solution instructions. Apart from that, can we show the error only in development? 

Something like this (I'm not sure about the correct syntax):

```suggestion
	if ( globalThis.SCRIPT_DEBUG && ! scope ) {
		throw Error(
			'Cannot call `getContext()` when there is no scope. If you are using an async function, please consider using a generator instead. If you are using some sort of async callbacks, like `setTimeout`, please wrap the callback with `withScope(callback)`.'
		);
	}
```

We should do the same in the `getElement` function.

To be honest, I'm not sure when we should throw errors in production and when we shouldn't. I think we haven't been very consistent in that regard. Perhaps we should do a general review of all these errors to have a more consistent policy.

packages/interactivity/src/directives.tsx
```diff
@@ -280,9 +286,10 @@ export default () => {
 							`The value of data-wp-context in "${ namespace }" store must be a valid stringified JSON object.`
 						);
 					}
-					updateSignals( currentValue.current, {
-						[ namespace ]: deepClone( value ),
-					} );
+					updateContext(
+						currentValue.current[ namespace ],
+						deepClone( value ) as object
```
luisherranz, line 292: Do we need to keep doing a deep clone of the object? If I'm not mistaken, objects are no longer merged on each context, so there is only a unique object for each context. Is that correct?

packages/interactivity/src/proxies/registry.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Gets the namespace associated with the given proxy.
+ *
+ * Proxies have a namespace assigned upon creation. See {@link createProxy}.
+ *
+ * @param proxy Proxy.
+ * @return Namespace.
+ */
+export const getProxyNs = ( proxy: object ): string => proxyToNs.get( proxy )!;
```
luisherranz: These two are not very consistent; maybe they should be called `getProxyFromObject` and `getNamespaceFromProxy` in case we need more of these in the future?

packages/interactivity/src/proxies/state.ts
```diff
@@ -0,0 +1,238 @@
...
+			return Reflect.get( target, key, receiver );
+		}
+
+		// At this point, the property should be reactive.
+		const desc = Object.getOwnPropertyDescriptor( target, key );
+		const prop = getPropSignal( receiver, key, desc );
+
+		if ( peeking ) {
+			return prop.getComputed().peek();
+		}
```
luisherranz: How do we get here if we have a previous conditional with `peeking` as well?

packages/interactivity/src/proxies/state.ts
```diff
@@ -0,0 +1,238 @@
...
+		const isNew = ! ( key in target );
+		const result = Reflect.defineProperty( target, key, desc );
+
+		if ( result ) {
+			const receiver = getProxy( target );
+			const prop = getPropSignal( receiver, key );
+			const { get, value } = desc;
+			if ( get ) {
+				prop.setGetter( desc.get! );
+			} else {
```
luisherranz, line 160: Does that work?

```suggestion
				prop.setGetter( get );
```

packages/interactivity/src/proxies/state.ts
```diff
@@ -0,0 +1,238 @@
...
+			}
+
+			if ( isNew && objToIterable.has( target ) ) {
+				objToIterable.get( target )!.value++;
+			}
+
+			if ( Array.isArray( target ) ) {
+				const length = getPropSignal( receiver, 'length' );
+				length.setValue( target.length );
+			}
```
luisherranz: We don't need to update to `length` if no one is subscribed.

Something like this?

```suggestion
			if ( Array.isArray( target ) && getPropSignal( receiver, 'length' ) ) {
				getPropSignal( receiver, 'length' ).setValue( target.length );
			}
```

packages/interactivity/src/proxies/state.ts
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		return result;
+	},
+
+	ownKeys( target: object ): ( string | symbol )[] {
+		if ( ! objToIterable.has( target ) ) {
+			objToIterable.set( target, signal( 0 ) );
+		}
+		( objToIterable as any )._ = objToIterable.get( target )!.value;
```
luisherranz, line 211: Let's add a comment, just in case someone wants to delete this line. Also, we should check in the minified bundle that it is indeed being preserved.

```suggestion
		// This subscribes to the signal while preventing the minifier from deleting this line in production. 
		( objToIterable as any )._ = objToIterable.get( target )!.value;
```

packages/interactivity/src/proxies/signals.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Identifier for property computeds not associated to any scope.
+ */
+const NO_SCOPE = Symbol();
+
+/**
+ * Structure that manages reactivity for a property in a state object. It uses
+ * signals to keep track of property value or getter modifications.
+ */
+export class PropSignal {
```
luisherranz, line 29: I wonder if we should change this name. Have you thought of any other options?

packages/interactivity/src/proxies/signals.ts
```diff
@@ -0,0 +1,142 @@
...
+	 * @param param0.get   New getter.
+	 * @param param0.value New value.
+	 */
+	private update( { get, value }: { get?: () => any; value?: unknown } ) {
+		if ( ! this.valueSignal ) {
+			this.valueSignal = signal( value );
+			this.getterSignal = signal( get );
+		} else if (
+			value !== this.valueSignal.peek() ||
+			get !== this.getterSignal?.peek()
```
luisherranz: Better? I suppose that `getterSignal` will always exist if we have reached this `if`, right?

```suggestion
			get !== this.getterSignal!.peek()
```

luisherranz (commented):
I haven't looked at the tests yet, but the code seems phenomenal to me. Very clear, very well organized, and easy to understand. Great job! 👏

I've left some comments and questions here.

packages/interactivity/src/proxies/test/store-proxy.ts
```diff
@@ -0,0 +1,125 @@
...
+import {
+	setScope,
+	resetScope,
+	setNamespace,
+	resetNamespace,
+	getContext,
+} from '../../hooks';
+
+describe( 'interactivity api - store proxy', () => {
+	describe( 'get', () => {
```
luisherranz: Maybe like this?

```suggestion
describe( 'interactivity api', () => {
	describe( 'store proxy', () => {
```

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+	setNamespace( ns );
+	try {
+		return callback();
+	} finally {
+		resetNamespace();
+		resetScope();
+	}
+};
+
+describe( 'interactivity api - state proxy', () => {
```
luisherranz: Same here?


```suggestion
describe( 'interactivity api', () => {
	describe( 'state proxy', () => {
```

And we should also change it in the tests of the `utils` file.

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+
+			const scope = { context: { test: { value: 'from context' } } };
+			try {
+				setScope( scope as any );
+				setNamespace( 'test' );
+				expect( state.y ).toBe( 'from context' );
+			} finally {
+				resetNamespace();
+				resetScope();
+			}
```
luisherranz: Shouldn't namespace be interally set?

The test pass with this:

```suggestion
			try {
				setScope( scope as any );
				expect( state.y ).toBe( 'from context' );
			} finally {
				resetScope();
			}
```

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+
+			const scope = { context: { test: { value: 1 } } };
+			try {
+				setScope( scope as any );
+				setNamespace( 'test' );
+				expect( state.sumContextValue( 2 ) ).toBe( 3 );
+			} finally {
+				resetNamespace();
+				resetScope();
+			}
```
luisherranz: Same here.

```suggestion
			try {
				setScope( scope as any );
				expect( state.sumContextValue( 2 ) ).toBe( 3 );
			} finally {
				resetScope();
			}
```

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+			expect( state.array.length ).toBe( 3 );
+			state.array.splice( 1, 2 );
+			expect( state.array.length ).toBe( 1 );
+		} );
+
+		it( 'should update when mutations happen', () => {
+			expect( state.a ).toBe( 1 );
+			state.a = 11;
+			expect( state.a ).toBe( 11 );
+		} );
```
luisherranz: What is the difference between this and 'should update like plain objects/arrays'?

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+
+		it( 'should support setting getters on the fly', () => {
```
luisherranz: Maybe we should add a test similar to this one, but where `state.double` does exist (it's not `undefined`).

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+			const raw1 = { obj: {} };
+			const raw2 = { obj: {} };
+			const state1 = proxifyState( 'test-1', raw1 );
+			const state2 = proxifyState( 'test-2', raw2 );
+			state2.obj = state1.obj;
+			expect( state2.obj ).toBe( state1.obj );
+			expect( raw2.obj ).toBe( state1.obj );
+		} );
+
+		it( 'should use its namespace by default inside setters', () => {
```
luisherranz: Why specifically within setters (and not within getters, for example)?

packages/interactivity/src/proxies/test/state-proxy.ts
```diff
@@ -0,0 +1,1187 @@
...
+		} );
+
+		it( 'should return the same proxy if initialized more than once', () => {
+			const raw = {};
+			const state1 = proxifyState( 'test', raw );
+			const state2 = proxifyState( 'test', raw );
+			expect( state1 ).toBe( state2 );
+		} );
+
+		it( 'should return the same proxy when trying to re-proxify a state object', () => {
```
luisherranz: This?

```suggestion
		it( 'should throw when trying to re-proxify a state object', () => {
```

luisherranz (commented):
I took a look at the tests, very good job! 

I have left other comments.

gziolo:
Nice to see a slight decrease in the bundle size:

image

DAreRodz:
> Let's also see if we can get rid of the dependency on `DOMContentLoaded`, which has been causing some issues.

Done in https://github.com/WordPress/gutenberg/pull/62734/commits/ebbe873ea602932492ed12a278e46a45e28dcdd1. As the `init()` function uses `splitTask()` internally, the hydration of any interactive region happens asynchronously after all modules have run, i.e., right after `DOMContentLoaded`.

packages/interactivity/src/init.ts
DAreRodz: It was a real issue. IIRC, the `proxies/state.ts` module was `undefined` while trying to import `proxifyState` inside `store.ts`. The circular dependency was like this:

```
proxies/state > hooks > store > proxies/state
```

I fixed it  in https://github.com/WordPress/gutenberg/pull/62734/commits/e1694a6da4c1b43abe0cbe9f48ea26428b5068a1 by moving the scope and namespace-related logic to their respective modules.

packages/interactivity/src/proxies/index.ts
michalczaplinski: I also think `proxify` sounds all right 🙂

packages/interactivity/src/proxies/signals.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/9fb783cfe5fca6ebb3ab2b939253a90f34eb4bff.

packages/interactivity/package.json
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/9fb783cfe5fca6ebb3ab2b939253a90f34eb4bff.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/9fb783cfe5fca6ebb3ab2b939253a90f34eb4bff.

packages/interactivity/src/store.ts
DAreRodz: No difference at all. 😄 Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/2f376d9f300bc2ed119209324117cf470309ffa1.

packages/interactivity/src/hooks.tsx
DAreRodz: I'll include your suggestion. We can do the general review later. 🙂

luisherranz: Thanks, I have included it in the [Tracking Issue](https://github.com/WordPress/gutenberg/issues/63232):  "Review all errors and make them more consistent"

packages/interactivity/src/directives.tsx
DAreRodz: Uhm, if I remember correctly, I added that because the parsed JSON defined in the `data-wp-context` directive―i.e., the object returned by `toVdom()` as a prop―was being modified and that caused some sort of trouble during client-side navigation.

Maybe that won't be needed once we fix https://github.com/WordPress/gutenberg/issues/63864. 🤔

packages/interactivity/src/proxies/state.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/e8718e9145ba5ea105a9e1d49f565a0504abc38c.

packages/interactivity/src/proxies/registry.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/af00c39f536bc58975c439c853c790e610ed845c.

packages/interactivity/src/proxies/state.ts
DAreRodz: I guess you mean the line above the `else` right? Yup, it works fine.

DAreRodz: Updated in https://github.com/WordPress/gutenberg/pull/62734/commits/7a3cf86076d6052b4cf416789af6f26d453cdb7f.

packages/interactivity/src/proxies/state.ts
DAreRodz: Well, `getPropSignal` automatically creates a `PropSignal` instance, but I haven't implemented a way to know whether that `PropSignal` has been accessed. 😅 

I can think of two options:

1. Implement a `hasPropSignal` function in `proxies/state.ts` that tells whether a `PropSignal` exists for the given property. It should be as simple as this:
   ```js
   const hasPropSignal = ( proxy, key ) => proxyToProps.get( proxy )?.has( key );
   ```
2. Implement a `hasBeenAccessed` flag inside `PropSignal` instances that become true once `getComputed()` is called.

---

PS: Note that a `PropSignal`'s internal signals are not instantiated until either `getComputed()`, `setValue()`, or `setGetter()` are called.

packages/interactivity/src/proxies/signals.ts
DAreRodz: Naming is tricky. 😅

I thought of other options, but this one was the most appropriate (in my opinion). I'm open to suggestions, though! 😄

packages/interactivity/src/proxies/signals.ts
DAreRodz: Updated in https://github.com/WordPress/gutenberg/pull/62734/commits/91ae5e81b16a59558e757ae4c16e6f2bde69eb12.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: They look the same. I'll remove this one.

BTW, those tests came from `luisherranz/deepsignal` (see [this](https://github.com/luisherranz/deepsignal/blob/8fb50d677615b368eae90a4e8b9e4570fa459f62/packages/deepsignal/core/test/index.test.tsx#L183-L190) and [this](https://github.com/luisherranz/deepsignal/blob/8fb50d677615b368eae90a4e8b9e4570fa459f62/packages/deepsignal/core/test/index.test.tsx#L223-L227)), so maybe it should be fixed there as well? 🤔

packages/interactivity/src/proxies/test/store-proxy.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/75d088908bd5e8782df88b61215151ea1e43de4f.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/75d088908bd5e8782df88b61215151ea1e43de4f.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/8a6589aba4a69b7f339a327ffa3543f2b3ea8ef0

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/8a6589aba4a69b7f339a327ffa3543f2b3ea8ef0

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/c7f13ad847fd634b7c13e72b04a594f2e975f8a4.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Done in https://github.com/WordPress/gutenberg/pull/62734/commits/378bc766f6ad83b7faf52a2e3aa89c035abbe43e. I've also added another test to check that the effects are re-executed.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: I've added a test case for that in https://github.com/WordPress/gutenberg/pull/62734/commits/8a00d34d945fa32342a808f44da538af2ee7e0a2.

packages/interactivity/src/proxies/test/state-proxy.ts
DAreRodz: Nice catch. 😄 Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/417a0c591c2ce0b57bbc0fc83fa2c5529404d9fa.

packages/interactivity/src/proxies/state.ts
DAreRodz: I used the first option (doing the check directly in the conditional). Fixed in https://github.com/WordPress/gutenberg/pull/62734/commits/3eea77152f4cb4163c5b5502f0ae525eea697c70.

DAreRodz:
Luis, I've addressed all the actionable feedback you shared. Feel free to take a look at the PR again. 🙂

packages/interactivity/src/directives.tsx
luisherranz: > the object returned by toVdom() as a prop―was being modified

Hmm... do you know by who? `toVdom` creates new objects on each run, doesn't it? https://github.com/WordPress/gutenberg/blob/644264cbcee25b34dd00c6ee8ba65097462e0ef0/packages/interactivity/src/vdom.ts#L105

packages/interactivity/src/proxies/state.ts
luisherranz: Yes, sorry.

packages/interactivity/src/proxies/state.ts
luisherranz: Perfect, thank you.

packages/interactivity/src/proxies/signals.ts
luisherranz: It's internal, so let's leave it like this for now. We can change it in the future if we come up with something better 🙂

packages/interactivity/src/proxies/test/state-proxy.ts
luisherranz: > BTW, those tests came from `luisherranz/deepsignal` (see [this](https://github.com/luisherranz/deepsignal/blob/8fb50d677615b368eae90a4e8b9e4570fa459f62/packages/deepsignal/core/test/index.test.tsx#L183-L190) and [this](https://github.com/luisherranz/deepsignal/blob/8fb50d677615b368eae90a4e8b9e4570fa459f62/packages/deepsignal/core/test/index.test.tsx#L223-L227)), so maybe it should be fixed there as well? 🤔

I'll fix that, thank you!

packages/interactivity/src/store.ts
luisherranz: The use case is when there are multiple calls to the `store` function for the same namespace.

luisherranz (approved):
Let's merge this so it can be included in the next release.

I'll add more tests in a new PR to try to ensure we haven't missed anything.

PR #62739, opened 2024-06-21
PR template: add before/after table

ellatrix:
## What?

Adds a Before/After table to the PR template in the screenshot section.

## Why?

* I'm a bit tired of having to write out the table structure all the time.
* People may not be aware that it's possible to use tables to display screenshots side-by-side.

## How?

Adds a Markdown table skeleton.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,3 +20,7 @@ https://github.com/WordPress/gutenberg/blob/trunk/CONTRIBUTING.md -->
 <!-- How can you test the changes by using the keyboard only? Please note, this is required for PRs that change the user interface (UI). This ensures the PR can be tested for any possible accessibility regressions. -->
 
 ## Screenshots or screencast <!-- if applicable -->
+
+|Before|After|
+|-|-|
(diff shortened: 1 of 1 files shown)


t-hamano (commented):
If we go ahead with this PR, I feel a comment description may be necessary. One example is as follows. What do you think?

```
## Screenshots or screencast
<!-- If you would like to upload screenshots, please paste them in the table cells below, showing before and after this PR was applied. -->

|Before|After|
|-|-|
||| 
```

ellatrix:
Or:

```
|Before|After|
|-|-|
|<!-- Before screenshot here -->|<!-- After screenshot here -->|
```

t-hamano:
Personally, I think it would be easier to understand if there were both comments suggested by you and mine.

t-hamano (approved):
I've updated this PR according to [this comment](https://github.com/WordPress/gutenberg/pull/62739#issuecomment-2183305132), and as I believe this PR still makes sense, I would like to approve it.

ellatrix:
Thanks @t-hamano, I've added a comment, also indicating that it's merely a suggestion and not a requirement when adding screenshots.

PR #62741, opened 2024-06-21
Global Styles: Allow block variations to be included in color and typography presets.

scruffian:
## What?
Block variations require two extra properties - `blockTypes` and `title`. In trunk this means they aren't treated as "pure" color or typography variations, because they have these extra properties.

## Why?

These properties are configuration rather than style rules, so we should ignore them when determining if a style variation is only targetting colors / typography.

## How?

Takes a lot of the work from https://github.com/WordPress/gutenberg/pull/62709 but uses it to make an exception for the properties.

## Testing Instructions
1. Switch to the assembler theme (https://github.com/Automattic/themes/tree/trunk/assembler)
2. Open the Site Editor > Styles
3. Confirm that as well as many style variations you also see color presets (this is not the case in trunk

## Screenshots or screencast 
Screenshot 2024-06-21 at 13 27 17

### DIFF
--- a/packages/edit-site/src/components/global-styles/hooks.js
+++ b/packages/edit-site/src/components/global-styles/hooks.js
@@ -114,7 +114,7 @@ export function useSupportedStyles( name, element ) {
 
 export function useColorVariations() {
 	const colorVariations = useCurrentMergeThemeStyleVariationsWithUserConfig( {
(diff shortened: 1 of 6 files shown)


scruffian:
This was fixed another way

Issue #46734, opened 2022-12-22
UI Components, settle on 40px and 32px "small" sizes.


jasmussen:
UI components across the editor (input fields, buttons, etc.) are currently a range of sizes, between 32px, 36px, and 40px:

| <img width="271" alt="Screenshot 2022-12-22 at 11 21 06" src="https://user-images.githubusercontent.com/1204802/209113168-a6a99e37-411b-42e0-b0aa-3082be32788f.png">  |  <img width="351" alt="Screenshot 2022-12-22 at 11 20 59" src="https://user-images.githubusercontent.com/1204802/209113343-dad05050-3c81-450b-83e8-c9b8f3f3de85.png"> |
| --- | --- |
| The inserter button is 32x32, the other buttons are 36x36  | Buttons here are 36px, including the toggles |

| <img width="335" alt="Screenshot 2022-12-22 at 11 24 03" src="https://user-images.githubusercontent.com/1204802/209113777-92b5d64c-c5d2-406a-a6a2-6a7509116407.png"> | <img width="276" alt="Screenshot 2022-12-22 at 11 21 12" src="https://user-images.githubusercontent.com/1204802/209113738-0974341d-5fa0-4db4-8a41-dba9fb22943d.png">  |
| --- | --- |
| Toggle buttons in the block toolbar are 32x32  | Inspector controls are standardizing on 40px |

There's an opportunity here to standardize and make consistent the controls, so there's less guesswork as to which dimensions controls should have. We could do this:

* All controls (inputs, segmented controls, sliders, dropdowns, and buttons) are 40px by default.
* All toggles (block toolbar button, toggles in the inspector such as quick transforms, the active state inside a segmented control) are 32x32px.
* The Button component has a 32px small version, to be used when the button is used in context of toggles, such as the top toolbar.
* The 36px size is retired entirely.

Here's a grid of metrics to illustrate how this could work:

<img width="1247" alt="grid" src="https://user-images.githubusercontent.com/1204802/209149927-4279c3e3-f085-42f1-aeed-dfed42627453.png">

Both 32 and 40px are base8 friendly, which would simplify the grid system. And with a single default (40px for controls, 32 for toggles inside), there would be no guesswork as to which size to use. The 32px button size would be the exception to use, as noted, in the top toolbar for buttons.

mirka:
I'm very happy to see this! We'll start working out the technical logistics in #46741.

jasmussen:
I wonder if we can close this one as fixed?

[...]

PR #62742, opened 2024-06-21
Cover: Update min-height control.

jasmussen:
## What?

The Cover block has a min-height control using legacy input size, and a label, "Minimum height of cover", redundant since it's only ever showin inside the inspector for the cover block:

before

This PR updates the input to be the new default size, and larger unit selector, and changes the label to read "Minimum height":

after

For comparison, here's the image width/height controls:

image

## Testing Instructions

Insert a cover block, test the min-height inspector control.

### DIFF
--- a/packages/block-library/src/cover/edit/inspector-controls.js
+++ b/packages/block-library/src/cover/edit/inspector-controls.js
@@ -73,15 +73,16 @@ function CoverHeightInput( {
 
 	return (
 		<UnitControl
-			label={ __( 'Minimum height of cover' ) }
+			label={ __( 'Minimum height' ) }
 			id={ inputId }
 			isResetValueOnUnitChange
 			min={ min }
 			onChange={ handleOnChange }
 			onUnitChange={ onUnitChange }
-			__unstableInputWidth="80px"
+			__unstableInputWidth="calc(50% - 4px)"
 			units={ units }
 			value={ computedValue }
+			size="__unstable-large"
 		/>
 	);
 }
--- a/packages/block-library/src/cover/edit/inspector-controls.js
+++ b/packages/block-library/src/cover/edit/inspector-controls.js
@@ -79,7 +79,6 @@ function CoverHeightInput( {
 			min={ min }
 			onChange={ handleOnChange }
 			onUnitChange={ onUnitChange }
-			__unstableInputWidth="calc(50% - 4px)"
 			units={ units }
 			value={ computedValue }
 			size="__unstable-large"
@@ -300,6 +299,7 @@ export default function CoverInspectorControls( {
 			) }
 			<InspectorControls group="dimensions">
 				<ToolsPanelItem
+					className="single-column"
 					hasValue={ () => !! minHeight }
 					label={ __( 'Minimum height' ) }
 					onDeselect={ () =>
--- a/packages/block-library/src/cover/test/edit.js
+++ b/packages/block-library/src/cover/test/edit.js
@@ -372,10 +372,10 @@ describe( 'Cover block', () => {
 					} )
 				);
 				await userEvent.clear(
-					screen.getByLabelText( 'Minimum height of cover' )
+					screen.getByLabelText( 'Minimum height' )
 				);
 				await userEvent.type(
-					screen.getByLabelText( 'Minimum height of cover' ),
+					screen.getByLabelText( 'Minimum height' ),
 					'300'
 				);
 
--- a/test/e2e/specs/editor/blocks/cover.spec.js
+++ b/test/e2e/specs/editor/blocks/cover.spec.js
@@ -158,7 +158,7 @@ test.describe( 'Cover', () => {
 
 		// Ensure there the default value for the minimum height of cover is undefined.
 		const defaultHeightValue = await coverBlockEditorSettings
-			.getByLabel( 'Minimum height of cover' )
+			.getByLabel( 'Minimum height' )
 			.inputValue();
 		expect( defaultHeightValue ).toBeFalsy();
 


packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -73,15 +73,16 @@ function CoverHeightInput( {
...
 		<UnitControl
-			label={ __( 'Minimum height of cover' ) }
+			label={ __( 'Minimum height' ) }
 			id={ inputId }
 			isResetValueOnUnitChange
 			min={ min }
 			onChange={ handleOnChange }
 			onUnitChange={ onUnitChange }
-			__unstableInputWidth="80px"
+			__unstableInputWidth="calc(50% - 4px)"
```
jasmussen: Those 4px make up half the 8px gap that sits between width/height in the image block. There's a better way to do this, but is it this PR? I can also just type in 120px.

jameskoster: I think that 8px may be incorrect 🙈 

According to https://github.com/WordPress/gutenberg/issues/46734 the gap between columns should be 16px. This is the case in the background size inputs:

Screenshot 2024-06-21 at 16 15 38

Ideally we avoid a hard-coded value, just in case the inspector width changes in the future.

t-hamano: Probably, there is a difference in the gap values ​​between the two, so in the future it will be necessary to unify the gap rules.

`ToolsPanel` has a gap value of `8px` ([Source](https://github.com/WordPress/gutenberg/blob/0384b0ee3f57fbf77038e786f08c254a0e1b27ea/packages/components/src/tools-panel/tools-panel/component.tsx#L33)):

tools-panel

On the other hand, the `FocalPointPicker` component has a gap value of `16px` ([Source](https://github.com/WordPress/gutenberg/blob/0384b0ee3f57fbf77038e786f08c254a0e1b27ea/packages/components/src/focal-point-picker/controls.tsx#L57)):

focal-point-picker

Which one should we unify it to? I would lean towards standardizing on `8px`.

By the way, width/height in the image block, `grid-column: span 1` is applied:

https://github.com/WordPress/gutenberg/blob/0384b0ee3f57fbf77038e786f08c254a0e1b27ea/packages/block-editor/src/components/dimensions-tool/width-height-tool.js#L15-L17

Therefore, I think the following approach is also possible:

<details><summary>Details</summary>

```diff
diff --git a/packages/block-library/src/cover/edit/inspector-controls.js b/packages/block-library/src/cover/edit/inspector-controls.js
index 39dd425c62..65c32fd526 100644
--- a/packages/block-library/src/cover/edit/inspector-controls.js
+++ b/packages/block-library/src/cover/edit/inspector-controls.js
@@ -79,7 +79,6 @@ function CoverHeightInput( {
                        min={ min }
                        onChange={ handleOnChange }
                        onUnitChange={ onUnitChange }
-                       __unstableInputWidth="calc(50% - 4px)"
                        units={ units }
                        value={ computedValue }
                        size="__unstable-large"
@@ -314,6 +313,7 @@ export default function CoverInspectorControls( {
                                        } ) }
                                        isShownByDefault
                                        panelId={ clientId }
+                                       style={ { gridColumn: 'span 1' } }
                                >
                                        <CoverHeightInput
                                                value={
```

</details>

jasmussen: I think the intended value is 16px, and there's a discrepancy here. I'm happy to take a stab at trying to unify it here. 

Screenshot 2024-06-24 at 09 47 08

I have a feeling this might be best done separately, so what do you think I should set as a value for this one in the mean time?

t-hamano: I noticed that there is a style for a single column of the dimension panel:

https://github.com/WordPress/gutenberg/blob/688ff13706638ce0174fff396375468cadc63920/packages/block-editor/src/hooks/dimensions.scss

Since the min-height control is also part of the dimension panel, we should be able to make the following changes to this PR:

```diff
diff --git a/packages/block-library/src/cover/edit/inspector-controls.js b/packages/block-library/src/cover/edit/inspector-controls.js
index 39dd425c62..80110d5158 100644
--- a/packages/block-library/src/cover/edit/inspector-controls.js
+++ b/packages/block-library/src/cover/edit/inspector-controls.js
@@ -79,7 +79,6 @@ function CoverHeightInput( {
                        min={ min }
                        onChange={ handleOnChange }
                        onUnitChange={ onUnitChange }
-                       __unstableInputWidth="calc(50% - 4px)"
                        units={ units }
                        value={ computedValue }
                        size="__unstable-large"
@@ -300,6 +299,7 @@ export default function CoverInspectorControls( {
                        ) }
                        <InspectorControls group="dimensions">
                                <ToolsPanelItem
+                                       className="single-column"
                                        hasValue={ () => !! minHeight }
                                        label={ __( 'Minimum height' ) }
                                        onDeselect={ () =>
```

The width value ​​is not hard-coded, so when the `ToolsPanel` gap changes from `8px` to `16px`, we don't need to update the min-height control's style.

jasmussen: Nice, that works!
Screenshot 2024-06-24 at 10 51 51

You'll see that it doesn't line up with the border controls below, but as you surfaced with width/height and left/top, that's a separate thing to fix!

t-hamano (commented):
LGTM!

I think E2E tests will fail, but that's another issue in the core, see the link below.

https://github.com/WordPress/wordpress-develop/pull/6875#issuecomment-2185694119
https://wordpress.slack.com/archives/C02QB2JS7/p1719209877579299

I'm not sure whether to force merge this PR or not, but let's wait a bit.

t-hamano (approved):

jasmussen:
Thank you for the review and the context. I think the ship has sailed for this to be a backport, though, so there's no rush on force merging.

akasunil:
@jasmussen I had created [PR](https://github.com/WordPress/gutenberg/pull/62509) to implement range slider for height control of cover block. Is this still relevant? please let me know.

jasmussen:
@akasunil Sorry, I missed that.

I think that's still relevant, yes. I think if you can include the title change from this PR, perhaps I can close mine in favor of yours?

akasunil:
Yes, I have updated [title](https://github.com/WordPress/gutenberg/pull/62509/files#diff-4af4f90d0daaf9ff4f7c1836fe266febc31fac60488d8084b942cb7f62cb2a7bR62) too. Screenshot is old one, ill update it.
Thanks @jasmussen

jasmussen:
Closing this one in favor of #62509.

PR #62747, opened 2024-06-21
Pattern overrides: disallow override for image with caption/href

### LINKED ISSUE #62287: Pattern overrides: Image captions appear to not be handled correctly

ellatrix:
## What?

Fixes #62287 (the remainder of it, the other part was fixed by #62471)

When creating pattern overrides, disallow creating an override for an image with either a caption or href/link.

## Why?

Because it's not supported by block bindings right now.

## Testing Instructions

* Select an image with a caption (non empty), click the three dots and create a pattern.
* Edit original, then select the image again.
* Open the inspector, click Advanced, make sure that there is no override button.
* You could also repeat these steps for an image without a caption, and make sure the button is there.

### DIFF
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -426,6 +426,9 @@ export default function Image( {
 		</InspectorControls>
 	);
 
+	const arePatternOverridesEnabled =
+		metadata?.bindings?.__default?.source === 'core/pattern-overrides';
+
 	const {
 		lockUrlControls = false,
 		lockHrefControls = false,
@@ -470,7 +473,7 @@ export default function Image( {
 				lockHrefControls:
 					// Disable editing the link of the URL if the image is inside a pattern instance.
 					// This is a temporary solution until we support overriding the link on the frontend.
-					hasParentPattern,
+					hasParentPattern || arePatternOverridesEnabled,
 				lockCaption:
 					// Disable editing the caption if the image is inside a pattern instance.
 					// This is a temporary solution until we support overriding the caption on the frontend.
@@ -964,16 +967,20 @@ export default function Image( {
 		<>
 			{ controls }
 			{ img }
-
-			<Caption
-				attributes={ attributes }
-				setAttributes={ setAttributes }
-				isSelected={ isSingleSelected }
-				insertBlocksAfter={ insertBlocksAfter }
-				label={ __( 'Image caption text' ) }
-				showToolbarButton={ isSingleSelected && hasNonContentControls }
-				readOnly={ lockCaption }
-			/>
+			{ /* Don't add caption component and controls in images with pattern overrides */ }
+			{ ! arePatternOverridesEnabled && (
+				<Caption
+					attributes={ attributes }
+					setAttributes={ setAttributes }
+					isSelected={ isSingleSelected }
+					insertBlocksAfter={ insertBlocksAfter }
+					label={ __( 'Image caption text' ) }
+					showToolbarButton={
+						isSingleSelected && hasNonContentControls
+					}
+					readOnly={ lockCaption }
+				/>
+			) }
 		</>
 	);
 }
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -967,6 +967,7 @@ export default function Image( {
 		<>
 			{ controls }
 			{ img }
+
 			<Caption
 				attributes={ attributes }
 				setAttributes={ setAttributes }
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -967,20 +967,19 @@ export default function Image( {
 		<>
 			{ controls }
 			{ img }
-			{ /* Don't add caption component and controls in images with pattern overrides */ }
-			{ ! arePatternOverridesEnabled && (
-				<Caption
-					attributes={ attributes }
-					setAttributes={ setAttributes }
-					isSelected={ isSingleSelected }
-					insertBlocksAfter={ insertBlocksAfter }
-					label={ __( 'Image caption text' ) }
-					showToolbarButton={
-						isSingleSelected && hasNonContentControls
-					}
-					readOnly={ lockCaption }
-				/>
-			) }
+			<Caption
+				attributes={ attributes }
+				setAttributes={ setAttributes }
+				isSelected={ isSingleSelected }
+				insertBlocksAfter={ insertBlocksAfter }
+				label={ __( 'Image caption text' ) }
+				showToolbarButton={
+					isSingleSelected &&
+					hasNonContentControls &&
+					! arePatternOverridesEnabled
+				}
+				readOnly={ lockCaption }
+			/>
 		</>
 	);
 }
--- a/packages/patterns/src/components/pattern-overrides-controls.js
+++ b/packages/patterns/src/components/pattern-overrides-controls.js
@@ -101,10 +101,6 @@ function PatternOverridesControls( {
 						variant="secondary"
 						aria-haspopup="dialog"
 						onClick={ () => {
-							if ( hasUnsupportedImageAttributes ) {
-								return;
-							}
-
 							if ( allowOverrides ) {
 								setShowDisallowOverridesModal( true );
 							} else {
--- a/packages/patterns/src/components/pattern-overrides-controls.js
+++ b/packages/patterns/src/components/pattern-overrides-controls.js
@@ -31,7 +31,11 @@ function addBindings( bindings ) {
 	};
 }
 
-function PatternOverridesControls( { attributes, setAttributes } ) {
+function PatternOverridesControls( {
+	attributes,
+	setAttributes,
+	name: blockName,
+} ) {
 	const controlId = useId();
 	const [ showAllowOverridesModal, setShowAllowOverridesModal ] =
 		useState( false );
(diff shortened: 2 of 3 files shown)


SantosGuillamot:
For me, not showing the "Enable Overrides" button when an image has a link or a caption feels a bit confusing. Especially because users won't know why in some images the button exists and not in others. Maybe we should add a message explaining it?

Anyway, I believe @talldan will have better context than me.

ellatrix:
Sure, we can could disable the button and add a warning maybe.

SantosGuillamot:
Another possibility could be to "disallow caption for images with overrides" instead of "disallow overrides for images with caption/href".

If I am not mistaken, it should be possible doing something like this:

In the [image save](https://github.com/WordPress/gutenberg/blob/f4b229a036a3ac0031e130f2197c6714052fa018/packages/block-library/src/image/save.js#L87-L93), check `metadata.bindings.__default !== 'core/pattern-overrides`:

```js
{
  !RichText.isEmpty(caption) &&
    metadata?.bindings?.__default?.source !== "core/pattern-overrides" && (
      <RichText.Content
        className={__experimentalGetElementClassName("caption")}
        tagName="figcaption"
        value={caption}
      />
    );
}

``` 

In the [image edit](https://github.com/WordPress/gutenberg/blob/f4b229a036a3ac0031e130f2197c6714052fa018/packages/block-library/src/image/image.js#L969-L981), do the same check:

```js
{
  metadata?.bindings?.__default?.source !== "core/pattern-overrides" && (
    <Caption
      attributes={attributes}
      setAttributes={setAttributes}
      isSelected={isSingleSelected}
      insertBlocksAfter={insertBlocksAfter}
      label={__("Image caption text")}
      showToolbarButton={
        isSingleSelected &&
        (hasNonContentControls || isContentOnlyMode) &&
        !hideCaptionControls
      }
      readOnly={lockCaption}
    />
  );
}
```

talldan:
> Another possibility could be to "disallow caption for images with overrides" instead of "disallow overrides for images with caption/href".

I prefer this option. 👍

ellatrix:
> Another possibility could be to "disallow caption for images with overrides" instead of "disallow overrides for images with caption/href".

I'm not sure I understand this. So you'd _remove_ the caption of an image once you press the override button? That seems unexpected too?

SantosGuillamot:
> So you'd remove the caption of an image once you press the override button? That seems unexpected too?

Yes, you are right that seems unexpected too. I was just sharing it to know more thoughts. For me, it's fine to follow this approach if disable the "Enable Overrides" button and add a message asking the user to remove caption and links.

I started a [draft pull request](https://github.com/WordPress/gutenberg/pull/62775/) to explore the other approach, but I must say this one seems safer because it doesn't interfere with the content already created by the users.

talldan:
With this PR, couldn't the user first enable overrides and then add the caption? I don't think this approach can completely solve the full issue, and so I wonder whether it's worth it. There are probably a lot of small ways to workaround it that lead to the user still encountering problems.

FWIW, I'm personally not against keeping how it works in trunk. I think users would pretty quickly discover that captions can't be overridden and not use image captions with overrides.

Could also consider a simpler option of adding help text mentioning 'Overrides currently do not support caption text and links in the image block' next to the Enable Overrides button (for the image block only), and then letting the user decide what to do.

ellatrix:
I don't think you can?

image

talldan:
> I don't think you can?

When creating/editing the pattern you can 'Enable Overrides' then add a caption to the image:
Screenshot 2024-06-24 at 4 34 25 pm

So that's one way of achieving what this PR tries to prevent.

SantosGuillamot:
Maybe we need something like [this](https://github.com/WordPress/gutenberg/pull/62775/files#diff-a14b61c874095649b236b78413e805b4541277b24eb71ce350a17d9f2b6396b1R473-R480)?

> FWIW, I'm personally not against keeping how it works in trunk. I think users would pretty quickly discover that captions can't be overridden and not use image captions with overrides.

To me, it is tricky and a bit confusing because it can create weird user experiences. Preventing pattern creators from enabling overrides for images with captions or links seems safer.

Personally, I would add two things to this pull request:
* Instead of hiding the "Enable Overrides" button, I would disable it and change [the help text](https://github.com/WordPress/gutenberg/blob/e2cbac62a91bda35317b3e10595dc2a2a34c43a0/packages/patterns/src/components/pattern-overrides-controls.js#L81) to say caption and links need to be removed.
* Hide the caption and link controls from images with pattern overrides enabled.

With that, I think the user experience would be much better and we would avoid unexpected results.

talldan:
@SantosGuillamot Ok, lets try that. Shall we push straight to this PR? If I handle the first item you mention, are you ok to push the changes for the second?

SantosGuillamot:
> Shall we push straight to this PR? If I handle the first item you mention, are you ok to push the changes for the second?

As long as it is fine for @ellatrix , I can help push this forward however it is needed 🙂

ellatrix:
Feel free, of course :)

talldan:
I've pushed a commit with the change mentioned:
Screenshot 2024-06-24 at 5 13 01 pm

packages/patterns/src/components/pattern-overrides-controls.js
```diff
@@ -71,28 +75,44 @@ function PatternOverridesControls( { attributes, setAttributes } ) {
...
-					) }
+					help={ helpText }
 				>
 					<Button
 						__next40pxDefaultSize
 						className="pattern-overrides-control__allow-overrides-button"
 						variant="secondary"
 						aria-haspopup="dialog"
 						onClick={ () => {
+							if ( hasUnsupportedImageAttributes ) {
```
ellatrix: Is the onClick handler called after disabling it?

ellatrix:
> When creating/editing the pattern you can 'Enable Overrides' then add a caption to the image: Screenshot 2024-06-24 at 4 34 25 pm
> 
> So that's one way of achieving what this PR tries to prevent.

Oh, that one. Yes, we should disable it when an image block is being overridden.

packages/patterns/src/components/pattern-overrides-controls.js
talldan: No, looks like the button component will replace the event handler with one that stops clicks. I've pushed another commit to remove this if statement

SantosGuillamot:
I've pushed [another commit](https://github.com/WordPress/gutenberg/pull/62747/commits/c8e46f4235d5ba7bf3f271cb0d0c1c865439d05c) to hide the link and caption controls. The caption logic is slightly different from the href controls because when an image without overrides is used in a pattern instance, we still want to show the caption as "read-only". For `href`, in both cases we want just to hide the controls.

talldan:
It works well for me in testing:
- An image with a caption or link has a disabled 'Enable Overrides' button with the message
- For an image with overrides enabled the block in the original pattern has no 'Captions' or 'Link' buttons on the toolbar
- Pattern instances seem to be unaffected by the change.
- Image block without overrides still have the same functionality

talldan:
I've rebased the branch to see if it gets the tests passing

talldan (approved):
LGTM, though I worked on it, so any other approvals also appreciated 👍

SantosGuillamot (approved):
From my tests, it seems to be working fine as well. But I was also involved in the code changes 😄

packages/block-library/src/image/image.js
```diff
@@ -964,16 +967,20 @@ export default function Image( {
...
-				setAttributes={ setAttributes }
-				isSelected={ isSingleSelected }
-				insertBlocksAfter={ insertBlocksAfter }
-				label={ __( 'Image caption text' ) }
-				showToolbarButton={ isSingleSelected && hasNonContentControls }
-				readOnly={ lockCaption }
-			/>
+			{ /* Don't add caption component and controls in images with pattern overrides */ }
+			{ ! arePatternOverridesEnabled && (
+				<Caption
```
ellatrix: Why do we need this? Can't we just disable the caption button?

SantosGuillamot: I just assumed we didn't need the `Caption` component. But I guess we could just disable the button. Maybe changing [this property](https://github.com/WordPress/gutenberg/pull/62747/files#diff-a14b61c874095649b236b78413e805b4541277b24eb71ce350a17d9f2b6396b1R978-R980)?

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: aae9d92bfe

PR #62750, opened 2024-06-21
 Add unit support for image block height and width

### LINKED ISSUE #55059: Add unit support for image block height and width
In an effort to consolidate aspect ratio across blocks, let's add the unit control functionality that the Featured Image block supports for height and width, to the Image block.  Note that aspect ratio should be maintained, regardless of units. 

The image block only supports `px` currently:

CleanShot 2023-10-04 at 14 36 08

As seen in the featured image block: 

CleanShot 2023-10-04 at 14 34 20

akasunil:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/55059

## Why?

Width/height control have only `PX` unit enabled in Image block settings.
  
## Testing Instructions

- Go to editor
- Add image block
- go to block setting, and look for height/width control
- Control have other dimension units

## Screenshots or screencast 
image

### DIFF
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -355,14 +355,17 @@ export default function Image( {
 		);
 	}
 
+	const [ lightboxSetting, availableUnits ] = useSettings(
+		'lightbox',
+		'spacing.units'
+	);
+
 	// TODO: Can allow more units after figuring out how they should interact
 	// with the ResizableBox and ImageEditor components. Calculations later on
 	// for those components are currently assuming px units.
 	const dimensionsUnitsOptions = useCustomUnits( {
-		availableUnits: [ 'px' ],
+		availableUnits: availableUnits || [ 'px', '%', 'vw', 'em', 'rem' ],
 	} );
-
-	const [ lightboxSetting ] = useSettings( 'lightbox' );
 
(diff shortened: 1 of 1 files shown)


t-hamano (commented):
Thanks for the PR!

The point of this problem is, as [this comment](https://github.com/WordPress/gutenberg/blob/4f72dba4ac2118a033e0fda8a8ca958c5448e1a6/packages/block-library/src/image/image.js#L363-L365) indicates, that the `ResizableBox` and `ImageEdit` components assume `px` units.

Therefore, this PR works on the frontend but not in the editor.

I don't have any good ideas at the moment, but we need to solve this problem first.

https://github.com/WordPress/gutenberg/assets/54422211/a6b6696c-f279-4865-b391-517da562f51e

akasunil:
Thank you @t-hamano for insight.

I see, that's why it just had a PX unit till now, and because the featured image block isn't resizable, it has another unit in the width/height controls.

akasunil:
Cover block also has same functionality where min-height unit changed to PX unit whenever `ResizableBox` is used.
 
[Edit-Post-“Hello-World-”-‹-gutenberg-—-WordPress.webm](https://github.com/user-attachments/assets/9aa3e269-f229-437a-8e82-b75e170acbc4)

PR #62753, opened 2024-06-21
Accessibility: Fix custom color palette

### LINKED ISSUE #62728: Styles: the edit Palette UI is not operable with a keyboard
### Description

Splitting this out from https://github.com/WordPress/gutenberg/pull/62698

I couldn't find a way to _add_ or _remove_ a custom color in the edit palette UI when using only the keyboard. 
This is a serious violation as _any_ user interface must be operable with the keyboard.

I guess this issue is not limited to the Edit palette UI and can be reproduced wherever the color picker is used with no explicit buttons to Cancel and Save.

- Go to the Site Editor > Styles > Colors > Palette.
- Use the keyboard.

First not accessible flow:
- Move focus to the plus icon button 'Add color' and press Enter or Spacebar.
- The color picker opens.
- Observe that although the focus style is not always visible, all controls within the color picker aroperable with a keyboard.
- But...
- Once a color is set, there is no way to save or cancel and close the color picker with the keyboard.
- Observe there's no Button to Save.
- Press Escape: observe that nothing happens. Expected: the color value to be reset and the color picker to close.
- Use the mouse to click 'Done'.

Second not accessible flow:
- Move focus to the circular button of the new color and press Enter or Spacebar.
- The color picker opens.
- Change the values of the color usign the keyboard.
- Press Escape.
- Observe that now the color picker does close and the changes are saved.
- However, this is not the expected interaction: the Escape key should never, ever 'save'. Instead, it should always be used to cancel the current operation and reset any value.

Thir not accessible flow:
- Once a custom color is set, mofe focus to the ellipsis button 'Color options', press Enter, choose 'Show details' and press Enter.
- Move focus to the custom color item and press Enter or Spacebar.
- The color picker opens.
- Obserrve that whatever you do you can't exit the color picker with the keyboard.
- Observe there is no way to move focus to the `lineSolid` icon button to remove the color.
- Observe the remoce color button is only rendered in this 'editing' state of the UI, when keyboard users can't exit the color picker.

For clarity, this is the UI to be tested:

Screenshot 2024-06-21 at 11 03 28

### Environment info

_No response_

Yes

Yes

ellatrix:
## What?

Makes the custom color palette options more accessible. Fixes #62728, at least partly.

## Why?

The current setup is very overcomplicated and buggy. There's a wrapper button which is used to "edit a color", and only on click shows the input and remove button, but instantly puts focus in an unescapable popover.

## How?

1. Always should the input and remove button.
2. Make the color indicator a button, which opens the popover.
3. Define an `onClose` function for the popover.

## Testing Instructions

Go to Site Editor → Styles → Colors → Palette.
Click the three dots and then "Show details".
All fields should be focusable for each button.
When editing the color, it should be possible to escape the popover.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 -   `DropZone`: rewrite animation without depending on framer-motion. ([#62044](https://github.com/WordPress/gutenberg/pull/62044))
+-   `__experimentalPaletteEdit`: improve the accessibility. ([#62753](https://github.com/WordPress/gutenberg/pull/62753))
 
 ### Internal
 
--- a/packages/components/src/palette-edit/index.tsx
+++ b/packages/components/src/palette-edit/index.tsx
@@ -15,10 +15,7 @@ import {
 } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
 import { lineSolid, moreVertical, plus } from '@wordpress/icons';
-import {
-	__experimentalUseFocusOutside as useFocusOutside,
-	useDebounce,
-} from '@wordpress/compose';
+import { useDebounce } from '@wordpress/compose';
 
 /**
  * Internal dependencies
@@ -174,16 +171,13 @@ function Option< T extends Color | Gradient >( {
 	canOnlyChangeValues,
 	element,
 	onChange,
-	isEditing,
-	onStartEditing,
 	onRemove,
-	onStopEditing,
 	popoverProps: receivedPopoverProps,
 	slugPrefix,
 	isGradient,
 }: OptionProps< T > ) {
-	const focusOutsideProps = useFocusOutside( onStopEditing );
 	const value = isGradient ? element.gradient : element.color;
+	const [ isEditingColor, setIsEditingColor ] = useState( false );
 
 	// Use internal state instead of a ref to make sure that the component
 	// re-renders when the popover's anchor updates.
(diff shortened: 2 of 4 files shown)


joedolson:
Can I ask that we provide testing information that uses the names of things as they're exposed in the editing environment? There's nothing literally called "Global Styles" in the editor, so it's difficult for outside users to make use of those instructions to locate the item to be tested.

I believe this is referring to the color options in the Stylebook, but please confirm.

ellatrix:
It's called "Styles". I think it used to be called Global Styles and people still refer to it by that name

image

joedolson:
This is definitely a significant improvement.

I'd still very much like to see two changes:

1) Have a 'cancel' option in the color popover, so that it's possible to change a color, then return to the previous color without needing to have the hex value memorized. Right now, if you change a color, but don't have the previous value recorded, there's no way to return to the previous color. 

2) Have the `esc` key trigger the cancel behavior, and have a button available to save the changes. I think that a click outside of the popover should be treated as a completion action; but escape should be treated as a cancellation.

afercia:
Thanks for looking into this @ellatrix 
It does solve some of the issues discovered on https://github.com/WordPress/gutenberg/pull/62698 and then reported on https://github.com/WordPress/gutenberg/issues/62728

I'd agree with @joedolson that the color picker would need Cancel and Save buttons. Not sure if that can be addressed in this PR or separately, as it appears to be a broader issue with the color picker.

I'd suggest two further improvements to be made, if possible, in this PR:

For consistnecy, the focus style of the new button to edit custom colors should be, ideally, a circular outline. Screenshot of the two different focus styles of the color buttons:

Screenshot 2024-06-24 at 09 44 35

When removing a custom color by pressing the minus icon button 'Remove color', focus should be managed to avoid a focus loss. I'd think the only reasonable place to move focus to would be the plus icon button 'Add color'.

afercia (changes requested):
Please see last comment.

ellatrix:
I can check again later today. I'd like this PR to be included in RC1, so it should be ready today, but I have a few other things to work through first.

ellatrix:
To me even the focus outline might be need a larger change (needing to pull in and extracting other components), so I'm not sure if it should block this improvement tbh... I went for the quick fix of wrapping the color in a Button.

ellatrix:
I could try to prevent the focus loss on remove though 👍

afercia:
I'd agree it would be good to have these improvements in the RC. Further fixes can be done as follow-up in https://github.com/WordPress/gutenberg/issues/62728

ellatrix:
@afercia I fixed the focus loss.

ellatrix:
Let's do this PR instead of #62698.

afercia:
> I fixed the focus loss.

@ellatrix thanks. I'm not sure the 'Done' button is the right choice though. When there's only one custom color, removing it will also make the 'Done' button disappear so in this case focus is lost anyways.

ellatrix:
@afercia Where do you suggest it to move?

afercia:
> I'd think the only reasonable place to move focus to would be the plus icon button 'Add color'.

Although not ideal, I think the 'Add color' button is the only one that is guaranteed to be always rendered?

ellatrix:
I'm not sure I have time left to fix this, I need to release the editor packages asap. If not, this will slip to 6.7

afercia:
I can try to have a look at it.
Oh, I see also the Add color button is rendered conditionally on `! canOnlyChangeValues` :/

afercia:
After https://github.com/WordPress/gutenberg/pull/62753/commits/262a4867232d70e690e5162b42b8d8565e2a5cd6 focus is now moved back to the 'Add color' button when removing a color. When users `canOnlyChangeValues`, the focus loss doesn't occur in the first place because there's no 'Remove' button.

This UI would need more work as there are other focus losses and room for accessibility improvements but I'd think this PR is an improvement that would be nice to have in the 6.6 release.

ellatrix:
Great, thanks @afercia! Could you approve the PR? I'll merge it and cherry pick it onto the release branch

afercia (approved):

afercia:
Done, thanks 👍🏽

ellatrix:
Manually backported this in https://github.com/WordPress/gutenberg/pull/62824

PR #62754, opened 2024-06-21
Fix: Include permission management on permanently delete, rename, and restore.

jorgefilipecosta:
Expands what was done in https://github.com/WordPress/gutenberg/pull/62589 to other three actions where it makes sense: Permanently delete, rename, and restore post action.

## Testing Instructions
I moved a page to the trash.
I added the following code to a PHP file:
```
function remove_delete_post_capability( $caps, $cap, $user_id, $args ) {
    if ( 'delete_post' == $cap ) {
            $caps[] = 'do_not_allow';
        }
    }

    return $caps;
}

add_filter( 'map_meta_cap', 'remove_delete_post_capability', 10, 4 );
```
I went to the trash page views and verified the permanently delete page action is unavailable.
I went to drafts view and verified the trash post action is still not available.

I added the following code to a PHP file:
```
function remove_edit_post_capability( $caps, $cap, $user_id, $args ) {
    if ( 'edit_post' === $cap ) {
            $caps[] = 'do_not_allow';
        }
    }

    return $caps;
}

add_filter( 'map_meta_cap', 'remove_edit_post_capability', 10, 4 );
```

I went to the drafts pages list and verified the ability to rename a page was not shown.
I went to the trash pages list and verified the ability to restore pages was not present.

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -307,7 +307,7 @@ const trashPostAction = {
 	},
 };
 
-function useTrashPostAction( postType ) {
+function useCanUserEligibilityCheckPostType( postType, capability, action ) {
 	const registry = useRegistry();
 	const { resource, cachedCanUserResolvers } = useSelect(
 		( select ) => {
@@ -321,20 +321,28 @@ function useTrashPostAction( postType ) {
 	);
 	return useMemo(
 		() => ( {
-			...trashPostAction,
+			...action,
 			isEligible( item ) {
 				return (
-					trashPostAction.isEligible( item ) &&
+					action.isEligible( item ) &&
 					registry
 						.select( coreStore )
-						.canUser( 'delete', resource, item.id )
+						.canUser( capability, resource, item.id )
 				);
 			},
 		} ),
 		// We are making this use memo depend on cachedCanUserResolvers as a way to make the component using this hook re-render
 		// when user capabilities are resolved. This makes sure the isEligible function is re-evaluated.
 		// eslint-disable-next-line react-hooks/exhaustive-deps
-		[ registry, resource, cachedCanUserResolvers ]
+		[ action, registry, capability, resource, cachedCanUserResolvers ]
+	);
+}
+
+function useTrashPostAction( postType ) {
+	return useCanUserEligibilityCheckPostType(
+		postType,
+		'delete',
+		trashPostAction
 	);
 }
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -535,6 +551,14 @@ const restorePostAction = {
 	},
 };
 
+function useRestorePostAction( postType ) {
+	return useCanUserEligibilityCheckPostType(
+		postType,
+		'update',
+		restorePostAction
```
ellatrix, line 545: Well, turns out minimising the diff was useful for reusability too 😉

packages/editor/src/components/post-actions/actions.js
```diff
@@ -307,7 +307,7 @@ const trashPostAction = {
 	},
 };
 
-function useTrashPostAction( postType ) {
+function useCanUserEligibilityCheckPostType( postType, capability, action ) {
 	const registry = useRegistry();
 	const { resource, cachedCanUserResolvers } = useSelect(
 		( select ) => {
```
ellatrix: Could we avoid creating this subscription many times, and just do it once?

ellatrix:
When I remove the edit cap, I'm still seeing the Edit button:

image

jorgefilipecosta:
> When I remove the edit cap, I'm still seeing the Edit button:
> 
> image

Hi @ellatrix,

I think that's ok, the edit means open in the site editor, but in my tests the site editor does permission checks and for example, if you open a page there but don't have permission to edit you will not be able to change the page content. I guess if you have permission to change the template you may change the template but not the content.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Yes for sure the diff change made things easier to reuse.

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Hi @ellatrix, I applied a change and now this PR does not add any store subscription.

ellatrix (approved):

packages/editor/src/components/post-actions/actions.js
```diff
@@ -307,34 +307,29 @@ const trashPostAction = {
...
-			...trashPostAction,
+			...action,
 			isEligible( item ) {
 				return (
-					trashPostAction.isEligible( item ) &&
+					action.isEligible( item ) &&
 					registry
 						.select( coreStore )
-						.canUser( 'delete', resource, item.id )
+						.canUser( capability, resource, item.id )
```
ellatrix, line 320: I guess it doesn't really happen in the case, but the problem here is that `isEligible` will not be called if the user capability changes, right? So maybe all of these need to be precalculated/subscribed to at some point so the buttons re-render.

ellatrix:
Handled in #62823.

Issue #56310, opened 2023-11-20
Multiple image blocks set to open in light box should connect to create cohesive carousel.


jeffgolenski:
## What problem does this address?

Currently, if I have several full sized images within a post set to open in a lightbox, it's very tedious to view the next image. Each image in a post must be clicked or tapped, then closed in order for viewer to get to the next one. It's not a fluid experience for the viewer.

If I have multiple images in a post set to open in the lightbox, a viewer should be able to tap on one to open the lightbox, and then be able to fuildly navigate to each image within the lightbox, like a carousel. 

## What is your proposed solution?

* Desktop: User should be able to use left or right arrow keys to navigate through all the images in the same lightbox
* Mobile: User should be able to swipe left or right to easily browse all images in the lightbox

Example: In my post I have many images set to open in the lightbox. If I tap on the first image, the lightbox opens, but to view the others I have to close it and tap the other images to re-open the lightbox. 

![IMG_2492](https://github.com/WordPress/gutenberg/assets/214813/59d49c9a-bead-46dc-b7df-643264f180a8)
![IMG_2493](https://github.com/WordPress/gutenberg/assets/214813/82d34359-e6f5-4816-b7a8-f4b205d731d5)

cc @mtias @jasmussen

jeffgolenski:
Video recording https://github.com/WordPress/gutenberg/assets/214813/e9dae2fa-8130-4c6e-b01c-4a34242d7385

jasmussen:
Good issue, would definitely agree that if we are going to have lightbox enabled in galleries, it should either be a nice gallery there as well, chevrons left and right plus arrow key navigation. Or, just be disabled. I could swear there was already an issue for this, but I didn't find one linked from #51132. CC: @michalczaplinski

michalczaplinski:
I don't recall there being an issue for this yet, so thank you @jeffgolenski!

michalczaplinski:
I've linked this issue in [Tracking issue: Image Lightbox #51132](https://github.com/WordPress/gutenberg/issues/51132)

bradhogan:
+1 on this - very common in all other platforms (webflow, sqs, etc.) would love to see this added soon.  Thanks all.

jacobgkau:
[...]

Issue #37652, opened 2021-12-29
Lightbox option for gallery block


matthiasballreich:
It would be very nice to have the Lightbox option for the gallery block. I now i can click on every single image inside the gallery block and enable there the lightbox option, but if i have many images, this ist not the best way. It would be much better to have the lightbox option for the whole gallery block.

lwangdu:
If there is lightbox option for the gallery block, that would be wonderful.  So we don't have to install another plugin, for that reason. Thank you.

paaljoachim:
That is a very good and helpful idea!
@glendaviesnz Glen what are your thoughts around having a lightbox in the Gallery block?

glendaviesnz:
It has been discussed, but not sure what priority currently is, have added this issue to the gallery improvements project.

paaljoachim:
@glendaviesnz Glen. Can you add a link to the [Gallery Improvements project](https://github.com/WordPress/gutenberg/projects/56)? Thanks.

masteradhoc:
+1 for this enhancement

masteradhoc:
@paaljoachim @glendaviesnz is there any chance this gets a higher priority soon?

paaljoachim:
I will go ahead and add the "Needs Design" label to it.

rfischmann:
Now that WordPress 6.4 is out with the native lightbox option for individual images, it doesn't make any sense that it won't work for galleries as well, right?

masteradhoc:
@paaljoachim is there really a design needed for this?

maybe @artemiomorales or @t-hamano are able to give us an insight how much can be reused from the image component?

paaljoachim:
Hey @masteradhoc Brian. There is no need for a new design as the Lightbox option has been added to the Image block. I look forward to seeing this also in the Gallery block.

glendaviesnz:
The Gallery block is essentially just a collection of Image blocks, so I think the main work here would be:

1. Implement a way to apply the lightbox on setting to all the child images of the gallery. The gallery already does this with the linkto and resolution settings, so shouldn't be too much work
2. Implement a carousel. I think this would be a basic expectation of users as you don't really want to be opening and closing every image in the gallery, you most likely want to page through them once the lightbox is open

rfischmann:
>     2. Implement a carousel. I think this would be a basic expectation of users as you don't really want to be opening and closing every image in the gallery, you most likely want to scroll through them once the lightbox is open

[...]

PR #35191, opened 2021-09-28
Fix native BlockAlignmentControl

dcalhoun:
## Related PRs

* https://github.com/wordpress-mobile/gutenberg-mobile/pull/4042

## Description
Fixes #35173. Relates to #34710. A recent web-centric change introduced usage of `MenuGroup` and `MenuItem`, both of which are unsupported in the native editor currently. This updates the conditional within `BlockAlignmentControl` to reinstate the previous UI for the native platform.

### Native Alignment

1. Add a Cover block within the editor. 
2. Press the Alignment icon within the block toolbar. 
3. Change the selected alignment from None to another option.
4. **Expected:** Verify the correct alignment is persisted for the block. 

### Web Alignment
Complete Testing Instructions found in https://github.com/WordPress/gutenberg/pull/34710. 

## Screenshots 
n/a

## Types of changes
Bug fix

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #62762, opened 2024-06-22
Move gallery link controls to the block toolbar

### LINKED ISSUE #55022: Move gallery link controls to the block toolbar
Related to https://github.com/WordPress/gutenberg/issues/55020, I propose moving the gallery block's "Link to" control from the inspector to a more familiar, and consistent, link control in the block toolbar. This would consolidate the varying linking controls found in image and galleries today, while also making way for applying "Expand on click" to all images within a gallery block. 

A follow-up to this would be to include "Expand on click" at the gallery parent level here, just like single image blocks will have in https://github.com/WordPress/gutenberg/issues/55020. 

### Before
CleanShot 2023-10-03 at 16 25 30

### Proposed
Lightbox for galleries, i4

akasunil:
## What?

Fixes: https://github.com/WordPress/gutenberg/issues/55022

## Why?

Required to move gallery block's "Link to" control from the inspector to a more familiar, and consistent, link control in the block toolbar.
 
## Screenshots or screencast 

[Edit-Post-“Hello-world-”-‹-gutenberg-—-WordPress (1).webm](https://github.com/user-attachments/assets/2e967984-1a09-4990-a0ec-5c37e27889b0)


t-hamano (commented):
Thanks for the PR!

I did a quick test and found some layout issues:

- When the Gallery block is inside a Group block, the parent selector button is small in size.
- The link buttons seem to be slightly wider on both sides.

image

I think the component to use here is `ToolbarDropdownMenu`.
If you search the `packages/block-libraty` directory for f`ToolbarDropdownMenu`, you will find several implementation examples.

Additionally, the text of the following messages has become longer and could be made clearer.

trunk:

image

This PR:

image|

akasunil:
I appreciate your feedback.

> the text of the following messages has become longer and could be made clearer.

I updated the texts to reflect the suggestions made here on [issue](https://github.com/WordPress/gutenberg/issues/55022)

akasunil:
Link icon issues are fixed now
image

t-hamano (commented):
Thanks for the update! It's working as expected.

As a further improvement, I'm wondering if we can keep the notification text as suggested in [this comment](https://github.com/WordPress/gutenberg/pull/62762#pullrequestreview-2178003931).

Here are the changes I propose. What do you think?

```diff
diff --git a/packages/block-library/src/gallery/edit.js b/packages/block-library/src/gallery/edit.js
index 1332c8d21b..7fda27ecdc 100644
--- a/packages/block-library/src/gallery/edit.js
+++ b/packages/block-library/src/gallery/edit.js
@@ -70,16 +70,19 @@ const linkOptions = [
                icon: customLink,
                label: __( 'Link images to attachment pages' ),
                value: LINK_DESTINATION_ATTACHMENT,
+               noticeText: __( 'Attachment Pages' ),
        },
        {
                icon: imageIcon,
                label: __( 'Link images to media files' ),
                value: LINK_DESTINATION_MEDIA,
+               noticeText: __( 'Media Files' ),
        },
        {
                icon: linkOff,
                label: _x( 'None', 'Media item link option' ),
                value: LINK_DESTINATION_NONE,
+               noticeText: __( 'None' ),
        },
 ];
 const ALLOWED_MEDIA_TYPES = [ 'image' ];
@@ -377,7 +380,7 @@ function GalleryEdit( props ) {
                        sprintf(
                                /* translators: %s: image size settings */
                                __( 'All gallery image links updated to: %s' ),
-                               linkToText.label
+                               linkToText.noticeText
                        ),
                        {
```

</details>

t-hamano:
Regarding the unit test failure, it seems that the block editor itself crashes on mobile, as [this action shows](https://github.com/WordPress/gutenberg/actions/runs/10019462053/job/27696050361?pr=62762).

We might be adding some code that is not compatible with mobile.

Unfortunately, I'm a Windows user and can't build the mobile app.

@geriux @dcalhoun, If you know why the unit tests are failing, I would appreciate it if you could let me know 🙇‍♂️ This PR moves the link control of the Gallery block from the block sidebar to the block toolbar.

akasunil:
> Here are the changes I propose. What do you think?

It good suggestion, we could do that. ill update the PR.

t-hamano (approved):
LGTM! 

For unit test failures, let's wait for help from the mobile team 😅

paaljoachim:
I might have misread or there might be something I am not understanding...

I notice that in Gutenberg plugin version 18.8.0

We have the following.
Screenshot 2024-07-24 at 14 51 15

Link to controls have already been moved to the Sidebar settings.

Next up would be to add Expand on click to the Link to drop down.

This overlaps with what I mentioned here in point number 1
https://github.com/WordPress/gutenberg/pull/62906#issuecomment-2245014838

akasunil:
@paaljoachim Do you mean that the `Expand on click` option that appears when we click the link icon in each gallery image should be included to the gallery link control?

image

paaljoachim:
@akasunil Yes

You got it. As the user would recognize what is seen in the Image block link Expand on click could then also be used in the Link to used in the Gallery block even though it is a bit hidden away.

As one would link to Expand on click. Somewhat similar to how it is done in the Image block.

Image block.
Screenshot 2024-07-24 at 15 38 43

Current Gallery block.
Screenshot 2024-07-24 at 15 38 55
Here it would be natural to add in the Expand on click option as seen in the Image block link area.

akasunil:
We could do that. My recommendation was to address it in a different PR. However, we currently have a test case that is failing this PR, and we are awaiting comment from the mobile team. I will address the suggestion you made.

paaljoachim:
@akasunil Yes. It seems like a good idea to open a new PR addressing the things I mentioned. Then just adding a comment here to the other PR.

akasunil:
In fact, we may need to wait until this PR is merged before opening a new one due to the dependacy of this PR, as doing so will require more time to resolve conflicts following the merger. 

We can incorporate your suggested change into the current PR since it pertains to a specific capability that this PR addresses.

akasunil:
Here is how it works

[Edit-Post-“Hello-world-”-‹-gutenberg-—-WordPress (1).webm](https://github.com/user-attachments/assets/2e967984-1a09-4990-a0ec-5c37e27889b0)

paaljoachim:
Hi @akasunil 

**I appologize I see that I was not clear in the above comment.**


It looks like you added the link to the Gallery block toolbar.

Let's look at both toolbars.
The Gallery block has this toolbar.
Screenshot 2024-07-24 at 18 52 16

The Gallery block has minor information in the toolbar and uses the sidebar settings more to control the Gallery.

The Image block has this toolbar.
Screenshot 2024-07-24 at 18 52 30

The Image block has a more focused toolbar approach.

Lets just make a minor adjustment right now and use the LINK To drop down feature in the right sidebar Gallery block settings to add the Expand on click. Adding adjustments to the toolbar is a bigger and more complex UI adjustment and is something that will be need to discussed in a broader discussion.

**Add the option only into the LINK To dropdown:**
Not into the Gallery block toolbar. Thank you. Again I appologzie for the misunderstanding.

Screenshot 2024-07-24 at 19 01 22

Lets keep to smaller adjustments. Keeping to this adjustment in relation to this comment from @glendaviesnz 
https://github.com/WordPress/gutenberg/issues/37652#issuecomment-1819727874

akasunil:
Actually, as stated in the PR description, the purpose of this PR is to relocate the `Link To` control from the sidebar to the toolbar as suggested in the issue https://github.com/WordPress/gutenberg/issues/55022. And i agree with that change, as image block have the `Link To` control in toolbar too. We should follow the same pattern for gallery block. 

If the same setting is in a different place for two different blocks, it will be confusing from the user point of view.

paaljoachim:
Ahhh gotcha. I am going full circle here....Ops....my bad....

Doing a summary. The LINK To drop down controls is being removed and instead a link is being added to the Gallery block toolbar adding consistency with the Image block link.

Screenshot 2024-07-24 at 19 35 43

paaljoachim:
@WordPress/gutenberg-design 
@jameskoster and @richtabor 

Any comments regarding the change of removing the Gallery block LINK TO dropdown in the sidebar settings and instead adding the same Image block toolbar link feature to the Gallery block toolbar?

t-hamano:
In my opinion, this PR is about moving existing functionality in the block sidebar to the block toolbar, and should avoid adding new functionality.

Looking at [this comment](https://github.com/WordPress/gutenberg/pull/62906#issuecomment-2245014838), it seems that a new option called "Expand on click" needs to be added to the Gallery block, and when enabled, it is expected that a UI will be displayed to control whether to show a chevron as an additional option.

If that is the case, would it be better not to move the link option to the block toolbar in the first place?

Either way, it may be better to hold off on this PR until the specs for #62906 are solidified.

akasunil:
> this PR is about moving existing functionality in the block sidebar to the block toolbar, and should avoid adding new functionality.

I acknowledge the concern and I have expressed my opinion regarding this on [comment](https://github.com/WordPress/gutenberg/pull/62762#issuecomment-2248367348)


> it seems that a new option called "Expand on click" needs to be added to the Gallery block, and when enabled, it is expected that a UI will be displayed to control whether to show a chevron as an additional option.

Since the "Expand on click" option is already present on the picture under the gallery block, it should be included there as well, since the gallery block already has all other `Link To` options identical to those on the image block. Coming to the slider capability in lightbox, i think that is something that needs discussion that i agree. But that is separate enhancement i believe.

By adding `Expand on click` option on gallery block, We are enabling the functionality that is already available on image block. Users can then enable the `Expand on click` option from any individual image block ( as it already available in trunk ) or can enable `Expand on click` from gallery block ( as this PR implement ), Both methods lead to the same outcome. Thus, I don't anticipate any problems.

t-hamano:
> By adding `Expand on click` option on gallery block, We are enabling the functionality that is already available on image block.

I understand this, of course.

However, we can't simply add the "Expand on click" option to the Gallery block, because we need to check whether the theme supports lightbox. And the current Image block has complex logic to decide whether to show this setting or not. Try searching for "lightbox" in the [packages/block-library/src/image/image.js](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/image.js) file.

Perhaps the Gallery block needs equivalent logic, and I think it will be necessary to thoroughly verify it in a separate PR to make it happen.

paaljoachim:
It really comes back to what I mentioned. It is a lot easier making smaller changes such as keeping the LINK To dropdown in the sidebar settings **for now** and just adding Expand into it. 

Adding something new even though it is consistent with the Image block is a lot more complex and really should be done at a later time.

I am circling back to this again. Simpler steps are easier to make compared to more complex steps. Breaking down complex steps into smaller simpler steps does help.

1. Keep the LINK To Gallery option in the sidebar and adding Expand into it. 
2. Discovering scenarios as @t-hamano Aki outlines that will impact the Expand option in the sidebar and addressing those. Actually this should likely be done at the same time as number 1. 
When these various concerns are adressed then...
3. One might be ready to make the next step as in moving the LINK To Gallery sidebar option to the Gallery Block toolbar. Again moving an option from the sidebar to the toolbar is a bigger step and one needs to figure out the smaller steps to get to it. That is also what I have been working on earlier adressing.

We are going to need some perspective on this. @glendaviesnz could you take a look?

luisherranz:
Linking here a relevant conversation about the "Expand on click" option in galleries and pinging @joen.

- https://github.com/WordPress/gutenberg/issues/56310#issuecomment-2206080002

akasunil:
> And the current Image block has complex logic to decide whether to show this setting or not. Try searching for "lightbox" in the [packages/block-library/src/image/image.js](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/image.js) file.

Thank you for clarifying @t-hamano, I now understand your concern. In that case, should i revert the lightbox code?

paaljoachim:
I would suggest creating a new PR that only addresses the smallest step possible. Adding the "Expand on click" to the LINK To dropdown in the Gallery block sidebar settings. Addressing issues that @t-hamano mentioned. As there seems to be a few scenarios that will need to be addressed. Also going through the discussions at https://github.com/WordPress/gutenberg/issues/56310 which might impact the Expand option for the Gallery block.
Let's see what Aki says as well.

madhusudhand (commented):
This works well in my testing.

Only Issue observed was the link icon is not set to active unlike the image, when a value is set.

image

Expectation: 

image

t-hamano:
@madhusudhand I am hesitant to add the new "Expand on clock" option to the Gallery block in this PR because there are a lot of things that need to be taken into consideration. See [this comment](https://github.com/WordPress/gutenberg/pull/62762#issuecomment-2249532142).

madhusudhand:
@akasunil I tested the new "Expand on click" at gallery against [Gallery: Add lightbox support](https://github.com/WordPress/gutenberg/pull/62906) and it work perfectly fine. 

But it is best to bring the lightbox related changes into a new PR and keep the current PR specific to movement of the toolbar option, by doing so, it doesn't have to wait for #62906

madhusudhand:
> @madhusudhand I am hesitant to add the new "Expand on clock" option to the Gallery block in this PR because there are a lot of things that need to be taken into consideration. See https://github.com/WordPress/gutenberg/pull/62762#issuecomment-2249532142.

I agree.

akasunil:
Thank you @madhusudhand, I'll revert the change for `Expand on click` option, and create separate PR.

packages/block-library/src/gallery/edit.js
```diff
@@ -594,6 +615,43 @@ function GalleryEdit( props ) {
...
+					icon={ linkIcon }
+					label={ __( 'Link To' ) }
+				>
+					{ ( { onClose } ) => (
+						<MenuGroup className="blocks-gallery__link-to-control">
+							{ linkOptions.map( ( linkItem ) => {
+								const isOptionSelected =
+									linkTo === linkItem.value;
+								return (
+									<MenuItem
```
dcalhoun: > @geriux @dcalhoun, If you know why the unit tests are failing, I would appreciate it if you could let me know 🙇‍♂️ This PR moves the link control of the Gallery block from the block sidebar to the block toolbar.

First, thank you for attempting to resolve the mobile unit test failures and for the ping. Much appreciated! 🙇🏻 

Both `MenuGroup` and `MenuItem` are unavailable for the native mobile editor currently. This results in the crash and unit test failure regarding the `Dropdown` component attempting to render/return an `undefined` value. 

We encountered this situation before and addressed it in https://github.com/WordPress/gutenberg/pull/35191 by bifurcating the code along the web and native platform line. It is not an ideal solution — see https://github.com/WordPress/gutenberg/pull/35191#issuecomment-930245860 — but, unfortunately, the lack of support for `MenuGroup` and `MenuItem` remains the current state. 

If we take a similar approach here, we might retain the previous `SelectControl` for native platforms. I opened https://github.com/WordPress/gutenberg/pull/63953 implementing this option, if we'd like to use this approach to move the proposed web changes forward quickly. 

Unfortunately, I do not believe the native mobile team will prioritize `MenuGroup` or `MenuItem` components in the near future.

artemiomorales:
Hi all, after taking a look at this discussion and related ones in https://github.com/WordPress/gutenberg/pull/62906, I [posted a summary](https://github.com/WordPress/gutenberg/issues/56310#issuecomment-2252075254) with some potentially helpful context that can perhaps offer ideas on next steps and/or cases to consider 🙏

Thanks @akasunil for taking a look into this!

packages/block-library/src/gallery/edit.js
akasunil: @dcalhoun appreciate your response. Thank you for the details on the issue and PR too.

akasunil: @t-hamano is it worth doing this way?

madhusudhand:
@akasunil

> Only Issue observed was the link icon is not set to active unlike the image, when a value is set.

+one minor tooltip difference. Image calls it `Link` and Gallery says `Link To`. I think both can just be called `Link`

@jasmussen Can you suggest on the tooltip?

akasunil:
> +one minor tooltip difference. Image calls it `Link` and Gallery says `Link To`. I think both can just be called `Link`

That's good suggestion.

packages/block-library/src/gallery/edit.js
```diff
@@ -594,6 +617,45 @@ function GalleryEdit( props ) {
...
 				</PanelBody>
 			</InspectorControls>
+			{ Platform.isWeb ? (
+				<BlockControls group="block">
+					<ToolbarDropdownMenu
+						icon={ linkIcon }
+						label={ __( 'Link To' ) }
+					>
+						{ ( { onClose } ) => (
+							<MenuGroup className="blocks-gallery__link-to-control">
```
t-hamano:
```suggestion
							<MenuGroup>
```

Nit: This class name doesn't seem to be used anywhere.

t-hamano (approved):
@dcalhoun @akasunil

I appreciate your efforts in resolving the problem 🙇‍♂️

@artemiomorales @madhusudhand 

I think this PR is ready to ship, is it ok to merge this PR before moving forward with #56310?

artemiomorales (commented):
> I think this PR is ready to ship, is it ok to merge this PR before moving forward with https://github.com/WordPress/gutenberg/issues/56310?

@t-hamano Yup! I think merging this PR first or in parallel makes sense.

However, for me the issue [mentioned here](https://github.com/WordPress/gutenberg/pull/62762#pullrequestreview-2199083871) is still present, namely that the link icon in the gallery toolbar is not highlighted when a selection is made.

When an option for the gallery link is selected, the icon should appear highlighted, the same as happens with the icon in image toolbar. Here's a video demonstrating the contrast:

https://github.com/user-attachments/assets/cb67c930-e8af-465b-8e4e-77704a9dbfd4

t-hamano:
> the link icon in the gallery toolbar is not highlighted when a selection is made.

From my understanding, this is expected. Because this dropdown is just a tool to change the `linkDestination` attribute of Image blocks in bulk.

Indeed, the Gallery block itself has the `linkTo` attribute, but it is only used to maintain the selection state of the select box (or the dropdown menu in this PR) and is not used anywhere else.

For example, change the "Link To" value in the Gallery block to "Link images to media files". Then the user unlinks all the images in it. It seems strange that the link icon is highlighted in the block toolbar of the Gallery block, even though all the images don't have links or lightboxes.

akasunil:
Shall we merge this PR?

t-hamano:
@akasunil 

Before merging this PR, let's all agree and confirm that the issue mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/62762#pullrequestreview-2202884580) is indeed the problem that this PR should solve or not.

akasunil:
> +one minor tooltip difference. Image calls it `Link` and Gallery says `Link To`. I think both can just be called `Link`

@t-hamano can you share your thoughts on this?

t-hamano:
> +one minor tooltip difference. Image calls it "Link" and Gallery says "Link To". I think both can just be called "Link"

This suggestion makes sense and I think it's OK to make this change in this PR.

madhusudhand (approved):
Changes looks good and Let's merge this PR. 🚀 

@akasunil A followup PR can be created [for this issue](https://github.com/WordPress/gutenberg/pull/62762#pullrequestreview-2199083871) (highlighting link icon) for design feedback and approval, based on which the change can be included or ignored.

akasunil:
> A followup PR can be created [for this issue](https://github.com/WordPress/gutenberg/pull/62762#pullrequestreview-2199083871) (highlighting link icon) for design feedback and approval, based on which the change can be included or ignored.

I have addressed this issue on PR https://github.com/WordPress/gutenberg/pull/64014

jeffpaul:
@randomburner and @RCNeil mind sharing your WordPress.org username so I can ensure its included in the 6.7 credits listing?

RCNeil:
@jeffpaul it is @rcneil as well.  Thanks!

PR #62767, opened 2024-06-22
Social Icons: Add examples for Social Link variations.

ndiego:
## What?
The variations for the Social Link block do not have the `example` property set, which causes the block inserter to display the message `No preview available`. This PR fixes that. 

## How?
The following code to adds example content to each variation.  I've set `#` for the `url` property since unlinked icons have opacity applied. 

```
if ( ! variation.example ) {
	variation.example = {
		attributes: {
			service: variation.name,
			url: '#',
		},
	};
}
```

## Testing Instructions
1. Add a Social Icons block to a post
2. Click to add an icon and then click "Browse all" and the block inserter will open on the left
3. Hover over the available icons, and you will now see the icon preview

## Screenshots or screencast 

| Before | After | 
|-|-|
|image|image|

The preview is not the best, but it's better than displaying no preview at all.

### DIFF
--- a/packages/block-library/src/social-link/variations.js
+++ b/packages/block-library/src/social-link/variations.js
@@ -341,16 +341,22 @@ const variations = [
 ];
 
 /**
- * Add `isActive` function to all `social link` variations, if not defined.
- * `isActive` function is used to find a variation match from a created
- *  Block by providing its attributes.
+ * Add `isActive` and `example` properties to all Social Link variations, if not defined.
  */
 variations.forEach( ( variation ) => {
-	if ( variation.isActive ) {
-		return;
+	if ( ! variation.isActive ) {
+		variation.isActive = ( blockAttributes, variationAttributes ) =>
+			blockAttributes.service === variationAttributes.service;
 	}
-	variation.isActive = ( blockAttributes, variationAttributes ) =>
-		blockAttributes.service === variationAttributes.service;
+
+	if ( ! variation.example ) {
+		variation.example = {
+			attributes: {
+				service: variation.name,
+				url: '#',
+			},
+		};
+	}
 } );
 
 export default variations;


packages/block-library/src/social-link/variations.js
```diff
@@ -341,16 +341,22 @@ const variations = [
...
+			blockAttributes.service === variationAttributes.service;
+	}
+
+	if ( ! variation.example ) {
+		variation.example = {
+			attributes: {
+				service: variation.name,
+				url: '#',
+			},
+		};
```
talldan, line 358: I think adding a `viewportWidth` in the `example` object would be good as the preview is quite small and scaled down.

A value of `0` results in no scaling which is an option. `150` results in the icon nicely centered but zoomed in and showing some pixelation.

talldan (commented):
It might be ok to ship this as it is, but the trouble is the preview isn't truly showing what the social icon will look like. For example, the WordPress icon is supposed to look like this:
Screenshot 2024-06-24 at 11 15 37 am

But the preview is showing this:
Screenshot 2024-06-24 at 11 16 05 am

This might be why there's no example 😄 

I think this is a side effect of how the block has been implemented. The example is showing a social link icon without the social links parent, but the social links block is what provides all the styles at the moment. I have no idea why it was implemented like this, my feeling is that each block should be responsible for providing its own base styles.

A fix would be to refactor the css of the blocks, though lots of care would have to be taken to make sure nothing is broken along the way.

t-hamano (commented):
Another concern is that it generates invalid HTML in the preview area. The `li` elements need to be placed as children of `ul` or `ol` element, but the parent of the `li` element is a `div`.

social-link

If the block itself could know in what context it was being rendered, it might be possible to, for example, wrap it in a Social Links block or remove the `li` element, but as far as I know there is no way to do this 🤔

talldan:
That's a good observation. Maybe supporting a `name` property for the `example` could be one way to solve the issue for blocks that are very coupled to a parent, it would allow the example to render a different block type:
```js
example: {
    name: 'core/social-links',
    innerBlocks: [
        {
            name: 'core/social-link',
            attributes: { service: variation.name, url: '#' },
        },
    ],
}
```

t-hamano:
> Maybe supporting a `name` property for the `example` could be one way to solve the issue for blocks that are very coupled to a parent, it would allow the example to render a different block type

I think this is a good idea. We might be able to ship this PR before that approach is implemented, but I'm hesitant about whether to tolerate invalid markup.

talldan:
> I think this is a good idea. We might be able to ship this PR before that approach is implemented, but I'm hesitant about whether to tolerate invalid markup.

One drawback I can think of is that if a user selects a different social links style (e.g. 'Huge' size, with 'Pill Shaped' block style), then the preview would show the incorrect style, and it might be quite hard to fix that with the way the social links block is hard coded. 

Having said that, it already seems to be a problem with other blocks (e.g. if you trying adding a paragraph inside a quote, the preview shows the paragraph with the wrong style).

edit: It could be an option like `showAsChildOfCurrentParentBlock: true` (naming could be better 😄  )

ndiego:
Well, at the very least, I am glad this PR started a good discussion around block previews. 😉

The "No preview available" message feels like something is broken, especially when it appears on Core blocks, and this situation with child blocks is relatively common. I have a block that uses a similar approach. 

I think adding the `name` could be very interesting, especially if you could also specify attributes for the parent.

```
example: {
    name: 'core/social-links',
    innerBlocks: [
        {
            name: 'core/social-link',
            attributes: { service: variation.name, url: '#' },
        },
    ],
}
```

> Having said that, it already seems to be a problem with other blocks (e.g. if you trying adding a paragraph inside a quote, the preview shows the paragraph with the wrong style).

I am not sure the example needs to represent the current state of the block when inserted in a specific spot or styled in a specific way by the parent. That sounds like it would get complicated very quickly. As long as it displays a realistic default state for the block, it's a good baseline IMO.

PR #62768, opened 2024-06-23
Global styles revisions: ensure that user-defined variation styles CSS is generated

ramonjd:
## What?

Hi.

Global styles revisions should output block style variations CSS if the revision contains variations styles. This PR does that.

### TODO

- [x] E2E tests

## How? 

By rendering the editor styles after the block list has rendered, at which the variation overrides are in the state and, therefore, accessible.

Then, using the current revision history config , grabbing the blocks' current style overrides, regenerating them using the current revision config, and re-registering them via `useStyleOverride`.

## Why?

~~A bit of code quality as it makes room for future, additional properties, but also~~ to fix a bug where the CSS for user-edited style variations in the global styles CPT/revisions aren't generated in the global styles revisions preview.

To generate style variation overrides, the style overrides CSS is generated using the global styles config in the `useBlockProps` hook.

However, the revision history contains multiple versions of the global styles config that need to be swapped in and out dynamically. 

## Testing Instructions

1. Register a custom block style variation, under `/styles` e.g., `/styles/white-fang.json`

```json
{
	"version": 3,
	"title": "White fang",
	"blockTypes": [
		"core/group",
		"core/image",
		"core/columns"
	],
	"styles": {
		"color": {
			"background": "yellow"
		},
		"border": {
			"radius": "1111px",
			"width": "10px",
			"color": "red",
			"style": "solid"
		},
		"blocks": {
			"core/paragraph": {
				"border": {
					"width": "0"
				},
				"color": {
					"background": "aliceblue",
					"text": "purple"
				}
			}
		}
	}
}
```

3. In the Site Editor, add a Group block and apply the `White fang` block style variation.
4. Then open the global styles panel for Group blocks and make a few edits to the `White fang` block style variation, saving each time.
5. Open the Styles revisions panel and preview your revisions. The changes to your style variation should be reflected for each revision are reflected on your Group block.
6. Check that there are no regressions to https://github.com/WordPress/gutenberg/pull/62465 (no duplicate CSS declarations for style variations in the editor)
7. Also please check that the variations appear correctly in the post editor and frontend.

Thank you!

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/6458278/393d7a92-9625-4676-bda5-123bbffeeafe


packages/edit-site/src/components/revisions/index.js
```diff
@@ -56,7 +56,9 @@ function Revisions( { userConfig, blocks } ) {
 		[ originalSettings ]
 	);
 
-	const [ globalStyles ] = useGlobalStylesOutputWithConfig( mergedConfig );
+	const [ globalStyles ] = useGlobalStylesOutputWithConfig( mergedConfig, {
```
ramonjd: I'm guessing the Style Book should also render these variations styles:

https://github.com/WordPress/gutenberg/blob/7f7128eeec0f19ffa86621cb993bd139ef163946/packages/edit-site/src/components/style-book/index.js#L230-L230

I tried it quickly, but I'm not yet sure it's enough - the preview doesn't render the style variation CSS, or update it when editing the global block styles.

Could be missing something. I'll look next week. Might be a different PR.

packages/block-editor/src/components/global-styles/use-global-styles-output.js
```diff
@@ -1369,7 +1376,10 @@ export function useGlobalStylesOutputWithConfig(
 			hasBlockGapSupport,
 			hasFallbackGapSupport,
 			disableLayoutStyles,
-			disableRootPadding
+			options?.disableRootPadding,
+			{
```
ramonjd: Note to self: I think `toStyles` is also ripe for a small refactor. Moving the following arguments

```
	hasBlockGapSupport,
	hasFallbackGapSupport,
	disableLayoutStyles = false,
	disableRootPadding = false,
```

into `styleOptions`

aaronrobertshaw (commented):
Thanks for getting the ball rolling on this one @ramonjd 💪 

While this does appear to solve the issue for simple applications of block style variations, I can confirm there is still an [issue](https://github.com/WordPress/gutenberg/pull/62712#issuecomment-2184405313) when it comes to nested variations that define inner block type or element styles.

Here's a rough demo video showing incorrect styles when variations define styles for inner group blocks.

https://github.com/WordPress/gutenberg/assets/60436221/de0d465d-43a4-420e-be28-515f00a3ec94



With the current approach in this PR the styles get generated once only which causes the last defined variation style to target the inner block to take precedence incorrectly.

This nesting use case was the driving force behind applying styles in a per-application or instance way. To solve it here we'll probably need something similar to the `block-style-variation.js` hook.

aaronrobertshaw:
In case it helps, here's some sample block markup and a few very simple section styles that can be used to replicate the issue. They aren't exactly what was in the video in my last comment but close enough.

<details>
<summary>Example block markup</summary>

```html
<!-- wp:group {"className":"is-style-section-c","style":{"spacing":{"padding":{"right":"var:preset|spacing|30","left":"var:preset|spacing|30","top":"var:preset|spacing|30","bottom":"var:preset|spacing|30"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group is-style-section-c" style="padding-top:var(--wp--preset--spacing--30);padding-right:var(--wp--preset--spacing--30);padding-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--30)"><!-- wp:paragraph -->
<p>Parent</p>
<!-- /wp:paragraph -->

<!-- wp:group {"className":"is-style-section-b","style":{"spacing":{"padding":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30","left":"var:preset|spacing|30","right":"var:preset|spacing|30"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group is-style-section-b" style="padding-top:var(--wp--preset--spacing--30);padding-right:var(--wp--preset--spacing--30);padding-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--30)"><!-- wp:paragraph -->
<p>Group</p>
<!-- /wp:paragraph -->

<!-- wp:group {"className":"is-style-section-a","style":{"spacing":{"padding":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30","left":"var:preset|spacing|30","right":"var:preset|spacing|30"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group is-style-section-a" style="padding-top:var(--wp--preset--spacing--30);padding-right:var(--wp--preset--spacing--30);padding-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--30)"><!-- wp:paragraph -->
<p>Child</p>
<!-- /wp:paragraph -->

<!-- wp:group {"className":"is-style-default","style":{"spacing":{"padding":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30","left":"var:preset|spacing|30","right":"var:preset|spacing|30"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group is-style-default" style="padding-top:var(--wp--preset--spacing--30);padding-right:var(--wp--preset--spacing--30);padding-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--30)"><!-- wp:paragraph -->
<p>Grandchild</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
<!-- /wp:group --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```
</details>

<details>
<summary>Section style partial theme.json files</summary>

```json
{
	"version": 3,
	"title": "Section A",
	"slug": "section-a",
	"styles": {
		"color": {
			"background": "slategrey",
			"text": "snow"
		},
		"blocks": {
			"core/group": {
				"color": {
					"background": "darkslategrey",
					"text": "whitesmoke"
				}
			}
		}
	}
}
```


```json
{
	"version": 3,
	"title": "Section B",
	"slug": "section-b",
	"styles": {
		"color": {
			"background": "teal",
			"text": "seashell"
		},
		"blocks": {
			"core/group": {
				"color": {
					"background": "mediumturquoise",
					"text": "white"
				}
			}
		}
	}
}
```


```json
{
	"version": 3,
	"title": "Section C",
	"slug": "section-c",
	"styles": {
		"color": {
			"background": "darkslateblue",
			"text": "aliceblue"
		},
		"blocks": {
			"core/group": {
				"color": {
					"background": "slateblue",
					"text": "linen"
				}
			}
		}
	}
}
```

</details>

ramonjd:
Thanks for explaining all that. I haven't delved too deeply yet, but it looks tricky. 

> This nesting use case was the driving force behind applying styles in a per-application or instance way. To solve it here we'll probably need something similar to the block-style-variation.js hook.

Interesting. So I see the editor is generating styles under "variation classname + clientId" selectors to target nested elements. And it's happening in the [useBlocks hook](https://github.com/WordPress/gutenberg/blob/7f7128eeec0f19ffa86621cb993bd139ef163946/packages/block-editor/src/hooks/block-style-variation.js#L108-L108). 

[Revisions prints a block list](https://github.com/WordPress/gutenberg/blob/7f7128eeec0f19ffa86621cb993bd139ef163946/packages/edit-site/src/components/revisions/index.js#L88-L88), but template level blocks, e.g.,  post-content block. `useBlockProps` doesn't run when switching between revision styles, and, even if it did, we'd have to make sure it uses the styles config merged with the currently-selected revision.

I still have a superficial overview, so I don't completely understand what needs to be done yet, but how right am I in assuming that we'd have to:

1. Somehow check each block to tell whether it's a variation from it's classname. (Maybe there's a hook?)
2. Rewrite the style object to include a `${ variation }-${ clientId }` for each eligible block, à la [useBlockStyleVariation](https://github.com/WordPress/gutenberg/blob/7f7128eeec0f19ffa86621cb993bd139ef163946/packages/block-editor/src/hooks/block-style-variation.js#L62-L62)
3. Generate the styles using the rewritten style object

And all that every time a user switches between revisions.

For point one, I'll check if there's a hook. I don't like the sound of iterating through a tree of blocks and checking for variations.

aaronrobertshaw:
>  how right am I in assuming that we'd have to:

That aligns with my initial impressions once this issue had been uncovered.

> 1. Somehow check each block to tell whether it's a variation from it's classname. (Maybe there's a hook?)

There was a hook specific to this smaller task however there was a recent refactor that improved the `useBlockProps` hook to only run when a variation class matched the class name rather than when a block had a class name. Perhaps that could be reused but I might have another idea.

> 2. Rewrite the style object to include a ${ variation }-${ clientId } for each eligible block, à la [useBlockStyleVariation](https://github.com/WordPress/gutenberg/blob/7f7128eeec0f19ffa86621cb993bd139ef163946/packages/block-editor/src/hooks/block-style-variation.js#L62-L62)
> 3. Generate the styles using the rewritten style object

Something like this will be needed.

>  I don't like the sound of iterating through a tree of blocks and checking for variations.

There might be a shortcut we can take to determine the blocks that have variations applied. 

The styles in the editor are generated and applied as [style overrides](https://github.com/WordPress/gutenberg/blob/7f7128eeec0f19ffa86621cb993bd139ef163946/packages/block-editor/src/hooks/utils.js#L138). To achieve nested section styling we needed to ensure these overrides were stored with the blocks `clientId` so the overrides could be sorted in order of the blocks on the page.

My understanding (which is very limited) of the revisions UI is that the existing blocks on the page you are viewing are passed to a new iframed editor. Will those blocks' clientIds change? If not, we could leverage the style overrides to know which blocks need the section styles generated rather than iterating the block tree.

ramonjd:
> My understanding (which is very limited) of the revisions UI is that the existing blocks on the page you are viewing are passed to a new iframed editor. Will those blocks' clientIds change? 

No, the block list is static once open 👍🏻 

> To achieve nested section styling we needed to ensure these overrides were stored with the blocks clientId so the overrides could be sorted in order of the blocks on the page.
> we could leverage the style overrides to know which blocks need the section styles generated rather than iterating the block tree.

Ah, I didn't realize the style overrides were stored and we could access them. That opens another door, thanks!

ramonjd:
> My understanding (which is very limited) of the revisions UI is that the existing blocks on the page you are viewing are passed to a new iframed editor. Will those blocks' clientIds change?
> 
> > No, the block list is static once open 👍🏻

I think understand the question better now: yes the client ids change because the block list is rendered again in the revisions view. Between applying style revisions, the client ids don't change again.

So this, I think, throws doubts on using the `getStyleOverrides`

aaronrobertshaw:
> So this, I think, throws doubts on using the getStyleOverrides

That's a shame 🤔 

If we set the variation class on the blocks that have variations applied before passing them to revisions perhaps that gives us something to rely on. 

Without having had a chance to dig into the code yet, it seems to me like we need a few things:
1. A reliable means to target blocks with styles e.g. setting the variation class rather than rely on clientId in revisions
2. A way to know which variation is applied to which blocks. This could be extracted from the class when they are set/collected above
3. Recreate the variation styles with each revision's merged style data using the class and variation name from above

packages/edit-site/src/components/revisions/index.js
```diff
@@ -88,6 +87,7 @@ function Revisions( { userConfig, blocks } ) {
 						settings={ settings }
 					>
 						<BlockList renderAppender={ false } />
+						<EditorStyles styles={ editorStyles } />
```
ramonjd, line 95: So now, the overrides store reflects the current state as the block list has been rendered and `useBlockProps` run.

However, now I need to merge the revisions style data with existing variations+clientId rules.

Working on that now...

packages/block-editor/src/components/editor-styles/index.js
```diff
@@ -67,15 +67,18 @@ function useDarkThemeBodyClassName( styles, scope ) {
 	);
 }
 
-function EditorStyles( { styles, scope } ) {
-	const overrides = useSelect(
-		( select ) => unlock( select( blockEditorStore ) ).getStyleOverrides(),
-		[]
+function EditorStyles( { styles, scope, overrides } ) {
```
ramonjd: Converted to draft while I refactor and question life choices.

This is the main thrust of things: passing overrides to EditorStyles, which can override the overrides.  🙃 

I'm overriding them all. Maybe better to merge them?

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -59,6 +59,72 @@ function getVariationNameFromClass( className, registeredStyles = [] ) {
 	return null;
 }
 
+export function useGetBlockVariationStylesWithConfig(
```
ramonjd: This is a first draft, but the basic logic is to grab the current overrides and update their CSS values with any incoming config. In our case, the user config from a styles revisions.

Ignore the naming. Maybe `useUpdateBlockVariationOverridesWithConfig`

I'm also looking for opportunities to reuse some of the logic within the block style variation hook, but that's a later thing.

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -157,6 +223,7 @@ function useBlockProps( { name, className, clientId } ) {
 		id: `variation-${ clientId }`,
 		css: variationStyles,
 		__unstableType: 'variation',
+		variation,
```
ramonjd, line 281: This is new. Required to identify the variation that the override targets.

Otherwise, we could do it the long way, that is, parsing the styles as the hook does. I think.

packages/edit-site/src/components/revisions/index.js
```diff
@@ -25,13 +25,29 @@ const {
...
+const cloneObject = ( obj ) => JSON.parse( JSON.stringify( obj ) );
+function RevisionStyles( { styles, config } ) {
+	const variationOverrides = useSelect(
+		( select ) => unlock( select( blockEditorStore ) ).getStyleOverrides(),
+		[ config ]
+	);
+
+	const mergedOverrides = useGetBlockVariationStylesWithConfig(
+		config,
+		cloneObject( variationOverrides )
```
ramonjd: Otherwise we're mutating the overrides object it seems. If we don't clone, we can just run the hook and not pass the result to Editor Styles.

This seems like a bug in the store, because mutating shouldn't be possible? That is, the result from useSelect should always be a unique reference. 

I don't know yet. Too hungry. I'll check.

packages/edit-site/src/components/revisions/index.js
```diff
@@ -88,6 +103,10 @@ function Revisions( { userConfig, blocks } ) {
 						settings={ settings }
 					>
 						<BlockList renderAppender={ false } />
+						<RevisionStyles
```
ramonjd: At this point the overrides from the useBlockProps hooks are in the store.

packages/block-editor/src/components/editor-styles/index.js
ramonjd: I'm only passing EditorStyles the overrides that need to override the overrides 😄 

And now, merging them with the store overrides.

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -59,6 +59,85 @@ function getVariationNameFromClass( className, registeredStyles = [] ) {
...
+ * based on an incoming theme config. If a matching style is found in the config,
+ * a new override is created and returned.
+ *
+ * @param {Object}               config             A global styles object, containing settings and styles.
+ * @param {Array<Array, Object>} variationOverrides An array of existing block variation overrides.
+ * @return {Array} An array of new block variation overrides.
+ */
+export function useUpdateBlockVariationOverridesWithConfig(
+	config,
+	variationOverrides = []
```
ramonjd: There are probably opportunities to abstract this and share some love with `useBlockStyleVariation`. Perhaps return an override object instead. I'll look into it later - first on the agenda is to see if this PR flies at all.

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -59,6 +60,85 @@ function getVariationNameFromClass( className, registeredStyles = [] ) {
...
 
+/**
+ * This hook is used to generate new block style variation overrides
+ * based on an incoming theme config. If a matching style is found in the config,
+ * a new override is created and returned.
+ *
+ * @param {Object} config A global styles object, containing settings and styles.
+ * @return {Array} An array of new block variation overrides.
+ */
+export function useUpdateBlockVariationOverridesWithConfig( config ) {
```
ramonjd: Actually, what I'm thinking would be better than all this is for `useBlockStyleVariation()` to have access to a mergedConfig that contains the current revision. 

https://github.com/WordPress/gutenberg/blob/8ba1ae20c98258bfc9ff13b3a19cee7879025334/packages/block-editor/src/hooks/block-style-variation.js#L142-L153

Maybe it could be return value from `useContext( GlobalStylesContext );`, e.g., 

```js
const { merged: mergedConfig, currentRevisionConfig } = useContext( GlobalStylesContext );
```

I'm just riffing here. Not certain if it'd work.

However, I think this scenario should be taken into account for:

- https://github.com/WordPress/gutenberg/issues/62216

One of the suggestions is to have the site editor recognise global styles as a valid entity, as it does for templates, pages and so on. Global styles could not only be the current mergedConfig, but a revision config. 

If I get time I'll start looking at that.

ellatrix:
I'm having a hard time understanding what the problem is. Is it the order of rendering the style overrides? This _could_ also be fixed by passing the clientId, and then rendering the styles in the right block order?

ramonjd:
> I'm having a hard time understanding what the problem is. Is it the order of rendering the style overrides? 

Good question. Sorry, the PR is young and dynamic and I haven't updated the description properly.

No, it's the content of the style overrides. 

Initially, the overrides are set using the global styles config in the `useBlockProps` hook.

However, the revision history contains multiple versions of the global styles config that need to be swapped in and out dynamically. 

> This could also be fixed by passing the clientId, and then rendering the styles in the right block order?

Could you elaborate? Passing the clientId to where? I'm interested to see if this helps.

packages/block-editor/src/components/editor-styles/index.js
```diff
@@ -95,7 +113,7 @@ function EditorStyles( { styles, scope } ) {
 				.map( ( style ) => style.assets )
 				.join( '' ),
 		];
-	}, [ styles, overrides, scope ] );
+	}, [ styles, styleOverrides, scope ] );
```
aaronrobertshaw: Won't we still need `overrides` still in the dependencies array?

packages/block-editor/src/hooks/block-style-variation.js
aaronrobertshaw: Both options sound worth exploring post 6.6>

```suggestion
export function useUpdateBlockStyleVariationOverridesWithConfig( config ) {
```

I know the name is getting pretty long but I think it helps to be clear this is to do with block style variations, not block variations.

packages/block-editor/src/hooks/block-style-variation.js
aaronrobertshaw: The tiny cost of adding it here certainly looks better than having to re-parse class names.

packages/block-editor/src/hooks/index.js
```diff
@@ -88,3 +88,4 @@ export { getTypographyClassesAndStyles } from './use-typography-props';
 export { getGapCSSValue } from './gap';
 export { useCachedTruthy } from './use-cached-truthy';
 export { useZoomOut } from './use-zoom-out';
+export { useUpdateBlockVariationOverridesWithConfig } from './block-style-variation';
```
aaronrobertshaw:
```suggestion
export { useUpdateBlockStyleVariationOverridesWithConfig } from './block-style-variation';
```

If adopting the earlier suggestion to keep the delineation between block style variations and block variations clear, we'll need this and the other references below updated too.

packages/block-editor/src/private-apis.js
```diff
@@ -20,7 +20,11 @@ import { cleanEmptyObject, useStyleOverride } from './hooks/utils';
 import BlockQuickNavigation from './components/block-quick-navigation';
 import { LayoutStyle } from './components/block-list/layout';
 import { BlockRemovalWarningModal } from './components/block-removal-warning-modal';
-import { useLayoutClasses, useLayoutStyles } from './hooks';
+import {
+	useLayoutClasses,
+	useLayoutStyles,
+	useUpdateBlockVariationOverridesWithConfig,
```
aaronrobertshaw:
```suggestion
	useUpdateBlockStyleVariationOverridesWithConfig,
```

packages/block-editor/src/private-apis.js
```diff
@@ -88,4 +92,5 @@ lock( privateApis, {
 	PrivatePublishDateTimePicker,
 	useSpacingSizes,
 	useBlockDisplayTitle,
+	useUpdateBlockVariationOverridesWithConfig,
```
aaronrobertshaw:
```suggestion
	useUpdateBlockStyleVariationOverridesWithConfig,
```

packages/edit-site/src/components/revisions/index.js
```diff
@@ -25,13 +25,21 @@ const {
 	ExperimentalBlockEditorProvider,
 	GlobalStylesContext,
 	useGlobalStylesOutputWithConfig,
+	useUpdateBlockVariationOverridesWithConfig,
```
aaronrobertshaw:
```suggestion
	useUpdateBlockStyleVariationOverridesWithConfig,
```

packages/edit-site/src/components/revisions/index.js
```diff
@@ -25,13 +25,21 @@ const {
...
 } = unlock( blockEditorPrivateApis );
 const { mergeBaseAndUserConfigs } = unlock( editorPrivateApis );
 
 function isObjectEmpty( object ) {
 	return ! object || Object.keys( object ).length === 0;
 }
 
+function RevisionStyles( { styles, config } ) {
+	const mergedOverrides =
+		useUpdateBlockVariationOverridesWithConfig( config );
```
aaronrobertshaw:
```suggestion
		useUpdateBlockStyleVariationOverridesWithConfig( config );
```

aaronrobertshaw (commented):
Really appreciate the time and effort that's gone into this fix @ramonjd 🙇 

I've only had a chance to take a quick look at the code so far to gain a little context around the approach. I've left a few tiny nits as suggestions but I'll start testing in earnest shortly.

packages/block-editor/src/components/editor-styles/index.js
ramonjd: Oh yeah, missed that. Ta.

Updated.

aaronrobertshaw (approved):
This is testing really well for me, nice work ✨ 

✅ Smoke test of existing section styles functionality didn't reveal any issues
✅ Revisions containing block style variations display correctly
✅ Applying revisions still works and correct variation styles are in effect
✅ Tweaks to custom block style variations work as expected
✅ Revisions containing style variation changes are reflected in post editor and frontend after applying


https://github.com/WordPress/gutenberg/assets/60436221/03fc3605-74b5-40a0-9fb7-9512bb490129

I'm happy to give this an approval now in the interests in getting this into the 6.6 RC at the earliest opportunity. My understanding is anything else raised so far can be explored as a follow-up, is that correct?

ramonjd:
Thanks a lot for the help on this PR @aaronrobertshaw 

>  My understanding anything else raised so far can be explored as a follow-up, is that correct?

That's my understanding too. As far as the change in this PR goes, it doesn't introduce any public APIs and any refactor can be done later to achieve the same results. 

The only thing I'd be keeping an eye on is side effects, but since we're not touching the overrides store itself I can't find any. 😅

aaronrobertshaw:
> As far as the change in this PR goes, it doesn't introduce any public APIs and any refactor can be done later to achieve the same results.

Agreed.

> The only thing I'd be keeping an eye on is side effects, but since we're not touching the overrides store itself I can't find any.

FWIW I didn't spot any either.

ellatrix:
> Could you elaborate? Passing the clientId to where? I'm interested to see if this helps.

Well, this was for making sure the order of the style overrides matches the order of the blocks. For the I would pass the block clientId to the style override, and then when rendering the overrides, I'd check the flattened block order and render the styles in that order.

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -59,6 +60,85 @@ function getVariationNameFromClass( className, registeredStyles = [] ) {
...
+ *
+ * @param {Object} config A global styles object, containing settings and styles.
+ * @return {Array} An array of new block variation overrides.
+ */
+export function useUpdateBlockStyleVariationOverridesWithConfig( config ) {
+	const overrides = useSelect(
+		( select ) => unlock( select( blockEditorStore ) ).getStyleOverrides(),
+		[]
+	);
+	const { getBlockStyles } = useSelect( blocksStore );
```
ellatrix: There are not called within a callback, so they should be within the subscription callback?

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -59,6 +60,85 @@ function getVariationNameFromClass( className, registeredStyles = [] ) {
...
 
+/**
+ * This hook is used to generate new block style variation overrides
+ * based on an incoming theme config. If a matching style is found in the config,
+ * a new override is created and returned.
+ *
+ * @param {Object} config A global styles object, containing settings and styles.
+ * @return {Array} An array of new block variation overrides.
+ */
+export function useUpdateBlockStyleVariationOverridesWithConfig( config ) {
```
ellatrix: I don't really understand any of this. What about these style overrides makes them revision-specific?

packages/block-editor/src/components/editor-styles/index.js
```diff
@@ -67,17 +67,35 @@ function useDarkThemeBodyClassName( styles, scope ) {
...
+function getStyleOverrideById( overrides, id ) {
+	if ( ! overrides?.length || ! id ) {
+		return [];
+	}
+
+	return overrides.find( ( [ _id ] ) => _id === id ) || [];
+}
+
+function EditorStyles( { styles, scope, overrides } ) {
+	const styleOverrides = useSelect(
```
ellatrix: If I'm understanding this correctly, this was already selected if you implement `overrides`. So we don't need to select it here again if `overrides` is defined, and we don't need to merge both?

ellatrix: What I'm seeing is that `useUpdateBlockStyleVariationOverridesWithConfig` uses `unlock( select( blockEditorStore ) ).getStyleOverrides()` from the store, then merges it with something else, then again we get `unlock( select( blockEditorStore ) ).getStyleOverrides()` and they are merged. So there's one merge too much here?

packages/edit-site/src/components/revisions/index.js
ellatrix: But `<EditorStyles>` normally listens for changes to the store? Why doesn't that work here?

ellatrix:
Are there e2e tests for this?

aaronrobertshaw:
> Well, this was for making sure the order of the style overrides matches the order of the blocks.

I believe the issue this PR addresses is a little different.

To preview each global styles revision, the current blocks are passed through to an [iframed block list](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-site/src/components/revisions/index.js#L72) along with that revision's [version of global styles config](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-site/src/components/revisions/index.js#L61).

This leads to two issues:
- The passed blocks get different clientIds and no longer match the style overrides applying block style variations
- When switching between revisions, the style overrides for variations need to be updated to reflect that revision's global styles config

I hope that helps clarify the problem this PR attempts to solve. I'm sure @ramonjd would be able to more eloquently explain it and the proposed solution though.

> Are there e2e tests for this?

Not yet but would be good to have.

packages/block-editor/src/components/editor-styles/index.js
ramonjd: `useUpdateBlockStyleVariationOverridesWithConfig` fetches the overrides only as a reference to check if a block has a block style variation applied. 

If so, and a matching path in the currently-selected revision user config exists, then it will return a new array containing a set of overrides specific to the current revision. 

The merge here in EditorStyles is a merge of CSS only, basically, "hey, I have these other overrides I'd like to output."

packages/block-editor/src/hooks/block-style-variation.js
ramonjd: Hmm, yeah good point.  

~~I was having a bit of trouble with that. I need to grab the block name from an array of clientIds. ~~

~~I'll try mapping the `overrides` array within the useSelect callback.~~

Can't do that, need the clientId 🤦🏻

packages/edit-site/src/components/revisions/index.js
ramonjd: From memory, it's not the order, but the fact that it's rendered as a child of the block editor provider. 

I didn't look too deeply into it after it started working.

packages/block-editor/src/hooks/block-style-variation.js
ramonjd: The style overrides in the store are generated using the current user global config.

We have to be able to recreate/create style variation overrides using any config, such as a global styles revision CPT.


With this PR I found the following constraints: 

- The revision preview recreates the block list, which creates new clientIds
- `useStyleOverride()` runs for these ideas, but it uses the current user global styles config
- The quickest way to tell if a style variation override existed for the blocks rendered in the revision preview was to check the overrides store.
- I wanted to call `useStyleOverride()` again to overwrite the overrides in the store using the revision config, but I had to do so by looping over a collection

It was a all bit of a rush job. 

That it's causing some confusion informs me that I should pursue [the other idea](https://github.com/WordPress/gutenberg/pull/62768#discussion_r1650735516) of setting a "currentRevision" config or something that can be fetched in [useBlockStyleVariation](https://github.com/WordPress/gutenberg/blob/fab458fd04491c8930a4cbdd8aab9765d5221428/packages/block-editor/src/hooks/block-style-variation.js#L142-L142)

ramonjd (commented):
Thanks again for taking the time to review this PR, folks.

As mentioned, the fact that confusion exists surrounding this change tells me:

- I need to explain the problems we're trying to solve more clearly (admittedly it took me a while to get my head around style variations in the last few days!), and 
- I need to look at  [my other idea](https://github.com/WordPress/gutenberg/pull/62768#discussion_r1650735516) of setting a "currentRevision" config or something that can be fetched in [useBlockStyleVariation](https://github.com/WordPress/gutenberg/blob/fab458fd04491c8930a4cbdd8aab9765d5221428/packages/block-editor/src/hooks/block-style-variation.js#L142-L142). This would, if it works like I think it would, mean far less code, but potentially require a "currentRevision" record somewhere in the core data store. 

> Are there e2e tests for this?

I'll work on that tomorrow.

packages/block-editor/src/hooks/block-style-variation.js
ramonjd: Oh sorry, I was referring to `getBlockName` below, not `getBlockStyles` 🤦🏻

 I'll move it up into the callback. Thanks!

test/e2e/specs/site-editor/block-style-variations.spec.js
```diff
@@ -0,0 +1,303 @@
...
+ */
+const { test, expect } = require( '@wordpress/e2e-test-utils-playwright' );
+const TEST_PAGE_TITLE = 'Test Page for Block Style Variations';
+test.use( {
+	siteEditorBlockStyleVariations: async ( { page, editor }, use ) => {
+		await use( new SiteEditorBlockStyleVariations( { page, editor } ) );
+	},
+} );
+
+test.describe( 'Block Style Variations', () => {
```
ramonjd, line 12: This is a bit of a monster. It checks:

- Applying block style variations in the site editor to blocks and nested blocks
- Checking that those styles are applied correctly
- Updating block style variations in global styles, and checking that those styles are applied correctly.
- Checking that the revisions preview matches the editor canvas

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -69,6 +70,9 @@ function OverrideStyles( { override } ) {
  * based on an incoming theme config. If a matching style is found in the config,
  * a new override is created and returned. The overrides can be used in conjunction with
  * useStyleOverride to apply the new styles to the editor.
+ * NOTE: This component is required to load global styles revisions config. Style variation overrides are
+ * generated using the canvas's current global styles (see useBlockStyleVariation()). This component is,
+ * however, due to be refactored and therefore it's use elsewhere is not recommended.
```
aaronrobertshaw: Do you think this could be communicated through the component's name? 

I'm only wondering if `Experimental` gets ignored more than something like "unstable". It's hard to see another case where this would be reused before the the proposed updates to the global styles architecture though.

Feel free to ignore this comment.

test/e2e/specs/site-editor/block-style-variations.spec.js
```diff
@@ -0,0 +1,303 @@
...
+	let stylesPostId;
+	test.beforeAll( async ( { requestUtils } ) => {
+		await Promise.all( [
+			requestUtils.activateTheme(
+				'gutenberg-test-themes/style-variations'
+			),
+			requestUtils.deleteAllPages(),
```
aaronrobertshaw: Is this needed if we are deleting all pages in the `beforeEach` as well?

test/e2e/specs/site-editor/block-style-variations.spec.js
```diff
@@ -0,0 +1,303 @@
...
+			.locator( '[data-type="core/group"]' )
+			.nth( 2 );
+
+		// Apply a block style to the parent Group block.
+		await editor.selectBlocks( firstGroup );
+		await editor.openDocumentSettingsSidebar();
+		await page.getByRole( 'tab', { name: 'Styles' } ).click();
+		await page
+			.getByRole( 'button', { name: 'Block Style Variation A' } )
+			.click( { force: true } );
```
aaronrobertshaw: Was there a particular reason we needed the `force: true` here?

test/e2e/specs/site-editor/block-style-variations.spec.js
```diff
@@ -0,0 +1,303 @@
...
+			.getByRole( 'button', { name: 'Block Style Variation B' } )
+			.click( { force: true } );
+
+		// Check nested child styles.
+		await expect( secondGroup ).toHaveCSS( 'border-style', 'dashed' );
+		await expect( secondGroup ).toHaveCSS( 'border-width', '2px' );
+
+		// Check nested grandchild group block inherited styles.
+		await expect( thirdGroup ).toHaveCSS( 'border-style', 'groove' );
+		await expect( thirdGroup ).toHaveCSS( 'border-width', '22px' );
```
aaronrobertshaw: I know this test is a beast already but I was wondering if we should also apply the first variation to the grandchild here at the end?

Something to ensure that earlier defined style variations don't have their styles overridden by a later definition's inner block type styles.

test/e2e/specs/site-editor/block-style-variations.spec.js
```diff
@@ -0,0 +1,303 @@
...
+	} );
+
+	test( 'update block style variations in global styles and check revisions match styles', async ( {
+		editor,
+		page,
+		siteEditorBlockStyleVariations,
+	} ) => {
```
aaronrobertshaw, line 102: Should this test also cover that the variation styles update when switching revisions? 

Knowing how this works, I appreciate that the first revision's styles wouldn't match the expectations if they weren't being processed correctly. It still might be handy to ensure that switching to a non-current revision updates appropriately. What do you think?

aaronrobertshaw (commented):
Nice work @ramonjd 👍 

I like that we're not touching the public EditorStyles component now. Bonus points for the e2e test as well!

Another round of testing the latest approach shows everything still working as expected.

✅ Existing section styles functionality is working
✅ Revisions containing block style variations display correctly
✅ Applying revisions still works and correct variation styles are in effect
✅ Tweaks to custom block style variations work as expected
✅ Revisions containing style variation changes are reflected in post editor and frontend after applying

I left a few minor questions about the e2e as inline comments but all in all this is looking pretty good to me. ✨

packages/block-editor/src/hooks/block-style-variation.js
ramonjd: I'm open to that. I see `__unstable` used a lot for components. 

Something like `__unstableBlockStyleVariationOverridesWithConfig` might be a better warning.

aaronrobertshaw: > It's hard to see another case where this would be reused before the the proposed updates to the global styles architecture though

I might have spoken too soon. It looks like the previews for patterns etc might also need some updates to reflect section styles.

test/e2e/specs/site-editor/block-style-variations.spec.js
ramonjd: Looks good to remove. 

Copy 🍝

test/e2e/specs/site-editor/block-style-variations.spec.js
ramonjd: Only because it was on my clip board 😄 Removing...

test/e2e/specs/site-editor/block-style-variations.spec.js
ramonjd: I was going to, but mechanism is the same. The config is passed to the new hook via the same channel. 

Won't hurt to check though 👍🏻

test/e2e/specs/site-editor/block-style-variations.spec.js
ramonjd: Just checking I understand - so the test is to apply a block style variation (A) to the nested grandchild block to ensure that A's styles overwrite any inherited styles from parent blocks.

I'll add it and you can tell me if I got it right 😄

packages/block-editor/src/hooks/block-style-variation.js
ramonjd: > It looks like the previews for patterns etc might also need some updates to reflect section styles.


Oh, good spotting! Let me know if I can help there too.

aaronrobertshaw: Thanks, after a bit of digging it appears to be a slightly different issue to this PR. 

Here the variation style overrides were "lost" entirely after the client ids change and need to be regenerated. This other issue now appears to be a matter of global styles being passed to the preview overriding the style overrides for variations in the parent editor (outside the preview's iframe).

**TL;DR**
I don't think the other issue will be able to reuse much if any of the fix here so shouldn't hold this PR up at all.

test/e2e/specs/site-editor/block-style-variations.spec.js
aaronrobertshaw: Yep, I believe we're on the same page.

If only a simple stylesheet was being generated and there was a single set of CSS for each variation, including inner block type styles, for example:

```css
:root :where(.is-style-variation-a .wp-block-group) {...}
:root :where(.is-style-variation-b .wp-block-group) {...}
```

Then checking that variation B gets applied when selecting it on the child covers whether that variation then overcomes the inner block type styles for variation A applied to the parent. So far, so good 👍 

Given the same specificity in the example above, the variation B inner block type styles would override the selection of variation A on the grandchild due to being loaded later. 

Adding this extra test for the application of variation A again on the grandchild will cover that we're not relying on the load order of the CSS. In other words, it ensures we're creating the per-instance styles appropriately so that the most immediate variation's styles are in effect.

I hope that makes it clearer 🤞

aaronrobertshaw (approved):
Thanks for the hard work and patience here @ramonjd 👍 

I've given this PR another test and it continues to work nicely. 

I'm happy to give this an approval again but if you do feel it warrants a second opinion after my false start, that's fine too 😁

ramonjd:
Cherry pick PR  to `wp/6.6` here:

- https://github.com/WordPress/gutenberg/pull/62905

ellatrix:
This was manually cherry-picked in #62905.

PR #62769, opened 2024-06-23
Site Editor: don't render empty body tag 

ramonjd:
## What?

Fix trailing curly brace, and only render margin collapsing CSS on the body tag in canvas view mode.

## Why?

To prevent `body{}}` from being rendered in the Site Editor in "edit" mode. 

Screenshot 2024-06-23 at 11 34 23 AM

## How?
Only building the CSS string when the Site Editor canvas mode is "view". 

I think that preserves the original intention of https://github.com/WordPress/gutenberg/pull/62146/files#diff-53f39da7a5f1f996f5359a2dd7b2747fb3d60a494523dd73dba946fab8e82592R144

## Testing Instructions

Head over to the Site Editor. View the source of the Editor frame. Inside the body tag you should see the body rule

Screenshot 2024-06-23 at 11 52 53 AM

Now enter edit mode. 

The body tag should not be there.

### DIFF
--- a/packages/edit-site/src/components/editor/index.js
+++ b/packages/edit-site/src/components/editor/index.js
@@ -110,14 +110,12 @@ export default function EditSiteEditor( { isLoading } ) {
 			{
 				// Forming a "block formatting context" to prevent margin collapsing.
 				// @see https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Block_formatting_context
-
-				css: `body{${
+				css:
 					canvasMode === 'view'
-						? `min-height: 100vh; ${
+						? `body{min-height: 100vh; ${
 								currentPostIsTrashed ? '' : 'cursor: pointer;'
-						  }`
-						: ''
-				}}}`,
+						  }}`
+						: undefined,
 			},
 		],
 		[ settings.styles, canvasMode, currentPostIsTrashed ]


packages/edit-site/src/components/editor/index.js
```diff
@@ -110,14 +110,12 @@ export default function EditSiteEditor( { isLoading } ) {
 			{
 				// Forming a "block formatting context" to prevent margin collapsing.
 				// @see https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Block_formatting_context
-
-				css: `body{${
+				css:
 					canvasMode === 'view'
-						? `min-height: 100vh; ${
+						? `body{min-height: 100vh; ${
 								currentPostIsTrashed ? '' : 'cursor: pointer;'
```
ellatrix, line 116: I wonder why this UI related CSS is not outside the iframe, while the iframe can itself can be set to `pointer-events:none`.

ellatrix (approved):

packages/edit-site/src/components/editor/index.js
ramonjd: 🤔 I'm not sure. It was a [legacy style pulled over in the unification PR](https://github.com/WordPress/gutenberg/pull/62146/files#diff-0475617e1471501c29fbc5c42e2490532cb18793b2c8d1d146fd7bb220ad3542L78)

Looking at the component hierarchy, I'm assuming it could be passed down to `VisualEditor` to `BlockCanvas` via `iframeProps` from the edit-site package.

PR #61212, opened 2024-04-29
Use output buffer and HTML tag processor to inject directives on BODY tag for full-page client-side navigation

westonruter:
## What?

This implements a resolution for this task from #60951: 

> * Explore how to modify the body tag without the hack: [link](https://github.com/WordPress/gutenberg/pull/59707#discussion_r1559952673).

## Why?

As [discussed](https://github.com/WordPress/gutenberg/pull/59707#discussion_r1559952673), the code is currently injecting an additional `<body>` tag after the BODY has already been opened, like so:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <h1>
      Hello World!
    </h1>
    <body data-wp-context="{}" data-wp-interactive="core/experimental">
  </body>
</html>
```

Surprisingly (to me at least), this actually [works](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Chtml%20lang%3D%22en%22%3E%0A%20%20%3Chead%3E%0A%20%20%20%20%3Cmeta%20charset%3D%22utf-8%22%3E%0A%20%20%3C%2Fhead%3E%0A%20%20%3Cbody%3E%0A%20%20%20%20%3Ch1%3E%0A%20%20%20%20%20%20Hello%20World!%0A%20%20%20%20%3C%2Fh1%3E%0A%20%20%20%20%3Cbody%20data-wp-context%3D%22%7B%7D%22%20data-wp-interactive%3D%22core%2Fexperimental%22%3E%0A%20%20%3C%2Fbody%3E%0A%3C%2Fhtml%3E):

image

Maybe I shouldn't be surprised given HTML's loose parsing rules. But I can imagine other plugins that try to parse the HTML to, for example, apply various optimizations would get might confused when encountering multiple BODY tags.

The attributes could rather get injected on the existing BODY tag instead with the HTML Tag Processor. I see this is not currently possible since the BODY tag is hard-coded in [`template-canvas.php`](https://github.com/WordPress/wordpress-develop/blob/f525e665b6ea6e88289d82cad5fc1dfac57db109/src/wp-includes/template-canvas.php#L20). This brings up the question again of output buffering for the entire template (Core-43258). It's something I've been working on in the context of the Performance team's [Optimization Detective](https://wordpress.org/plugins/optimization-detective/) plugin which output buffers the entire template and then uses HTML Tag Processor to do optimizations.

## How?

[...]

Issue #60280, opened 2024-03-28
wp:post-content - templateLock to work inside page templates


SamTyurenkov:
## What problem does this address?
Often Wordpress sites include a number of unique pages - homepage, policy, about, shop and some other with custom blocks on them. 
And developers provide those pages as templates in block based themes, e.g. `templates/page-about.html`

```
<!-- wp:template-part {"lock":{"move":true,"remove":true},"slug":"header","area":"header","tagName":"header"} /-->

	<!-- wp:mytheme/breadcrumbs /-->

	<!-- wp:mytheme/container -->
		<!-- wp:post-title /-->
	<!-- /wp:mytheme/container -->

	<!-- wp:post-content -->
		<!-- wp:mytheme/news-filtered /-->
                <!-- wp:mytheme/my-other-block /-->
	<!-- /wp:post-content -->

<!-- wp:template-part {"lock":{"move":true,"remove":true},"slug":"footer","area":"footer","tagName":"footer"} /-->
```

Note how wp:post-content might be meant to use with specific blocks, for that specific template

## What is your proposed solution?

I would expect something like that:
- ability to add templateLock to wp:post-content and specify inner blocks.
- in the editor, a notice to the user that he can reset template to the specified
- on new page creation, if page has empty content and a page template is selected from sidebar - apply desired blocks on template enable

```
	<!-- wp:post-content {"templateLock":"all","lock":{"move":true,"remove":true}} -->
		<!-- wp:mytheme/news-filtered /-->
                <!-- wp:mytheme/my-other-block /-->
	<!-- /wp:post-content -->
	```

juanpredev13:
I need this too

PR #62770, opened 2024-06-23
Only apply output buffer filter when in final phase

westonruter:
As [suggested](https://github.com/WordPress/gutenberg/pull/61212#issuecomment-2150662611) by @nextend in a follow-up to https://github.com/WordPress/gutenberg/pull/61212:

> We use output buffering in our 900k+ plugin and based on my experience I suggest the following changes:
> 
> ```php
> <?php
> function _gutenberg_buffer_template_output( string $passthrough ): string {
> 	ob_start(
> 		static function ( string $output, ?int $phase): string {
> 
> 			if ($phase & PHP_OUTPUT_HANDLER_FINAL) {
> 				/**
> 			 	* Filters the template output buffer prior to sending to the client.
> 			 	*
> 			 	* @param string $output Output buffer.
> 				 * @return string Filtered output buffer.
> 				 */
> 				return (string) apply_filters( 'gutenberg_template_output_buffer', $output );
> 			}
> 
> 			return $output;
> 		}
> 	);
> 	return $passthrough;
> }
> ```
> 
> As I remember these are related when `ob_clean` and `ob_flush` called during the output buffer is open. It skips the processing of the output when its dropped.

### DIFF
--- a/lib/experimental/full-page-client-side-navigation.php
+++ b/lib/experimental/full-page-client-side-navigation.php
@@ -86,14 +86,17 @@ add_filter( 'gutenberg_template_output_buffer', '_gutenberg_add_client_side_navi
  */
 function _gutenberg_buffer_template_output( string $passthrough ): string {
 	ob_start(
-		static function ( string $output ): string {
-			/**
-			 * Filters the template output buffer prior to sending to the client.
-			 *
-			 * @param string $output Output buffer.
-			 * @return string Filtered output buffer.
-			 */
-			return (string) apply_filters( 'gutenberg_template_output_buffer', $output );
+		static function ( string $output, ?int $phase ): string {
+			if ( $phase & PHP_OUTPUT_HANDLER_FINAL ) {
+				/**
+				 * Filters the template output buffer prior to sending to the client.
+				 *
+				 * @param string $output Output buffer.
+				 * @return string Filtered output buffer.
+				 */
+				$output = (string) apply_filters( 'gutenberg_template_output_buffer', $output );
+			}
+			return $output;
 		}
 	);
 	return $passthrough;


westonruter:
@nextend I'm having a hard time triggering a scenario in which `PHP_OUTPUT_HANDLER_FINAL` is not passed. I'm marking this as a draft so this doesn't get merged prior to validating the need for it.

westonruter:
I tried adding a test case in https://github.com/WordPress/performance/pull/1317 but I'm not having success.

nextend:
Here is an example that can trigger the `$phase` argument to not be `PHP_OUTPUT_HANDLER_FINAL`:

[https://bitbucket.org/rolandsoos/workspace/snippets/aq7bnz
](https://bitbucket.org/rolandsoos/workspace/snippets/aq7bnz
)
The content of the first `echo` will be dropped by the `ob_clean()` call. The current implementation in #60280 would apply the filter for this dropped message.

westonruter:
@nextend Thanks for providing that test case. Sorry for the delay in responding. To be honest I've been having a hard time wrapping my brain around this. I've iterated on the corresponding PR in the WordPress/performance repo and I'm getting closer but I'm still perplexed why it isn't behaving as expected. Could you take a look? https://github.com/WordPress/performance/pull/1317/files

nextend:
@westonruter I think the missing piece is to call `ob_end_flush();` before calling `ob_get_clean();`.

On line 91, a new output buffer is created. This output buffer is necessary as it allows you to access the output in the same way the PHP runtime does when outputting the default output buffer.

In the original code, the following happened:
- System [1] output buffer <- This will be sent to the browser.
  - Line 91 -> [2] output buffer
    - `od_buffer_output` -> [3] output buffer
    - This part is tricky: Calling `ob_get_contents()` at this point does not trigger the `ob callbac`k, as there might be more text to add to the buffer after calling `ob_get_contents`. Example: https://gist.github.com/nextend/ac5c90de0e2d628fba02355819998a34
    - If you check the source code of PHP, `ob_get_clean` has two parts: first, it calls `ob_get_contents`, and then it discards the output buffer. This is where a common misunderstanding occurs. In the first example, I showed why `ob_get_contents` does not trigger the `ob callback`, and since it is used inside `ob_get_clean`, there will be no `ob callback` call. Reference: https://github.com/php/php-src/blob/48916cec3383e5ac83bd9486d5445f847a36b922/main/output.c#L1408

In the new code:
- System [1] output buffer <- This will be sent to the browser.
  - Line 91 -> [2] output buffer
    - `od_buffer_output` -> [3] output buffer
    - `ob_end_flush` -> Output buffer [3] is closed -> `ob callback`  gets called -> Result of the callback is flushed into [2] buffer
  - Now we can access the real output with `ob_get_clean`, which gives us the content of the [2] buffer (and then discards the output).

westonruter:
@nextend Thank you. I iterated on it a bit more with your change applied and I think it's in a good shape: https://github.com/WordPress/performance/pull/1317

PR #62771, opened 2024-06-23
Fix registerFormatType type

mikeybinns:
## What?

Fixing the type of the `registerFormatType` function which is currently incorrect due to JSDoc types.

## Why?

Currently this function throws a type error if you don't specify the name parameter in the settings parameter, which is incorrect because the name is provided as the first argument.

## How?

I have converted the `registerFormatType` file, the Redux store files and all related test files to TypeScript.
I converted the WPFormat JSDoc type to a TypeScript type and moved it into the `./types` file.
I updated the name to be `RichTextFormatFull` instead of `WPFormat`. This should really be `RichTextFormat` however I didn't want to conflict with the JSDoc type of the same name so I've used this temporary name for now, and I've exported it as `RichTextFormat` as the old type wasn't exported.
I also added some missing / undocumented properties to the `RichTextFormatFull` and changed the `interactive` prop to optional.
I have avoided changing the base JS as much as possible to make sure this is only a types improvement, however there's a few minor places where I had to change some JS things, however all tests should still pass, and they are only minor changes which shouldn't affect functionality.

## Testing Instructions

Note: these changes are all covered by existing automated tests, but here's how to test manually.
1. Create a new TypeScript file, import `registerFormatType` from the public package version.
2. Register a new format using valid settings, leaving name out of the settings argument.
3. Note the Error Typescript is now throwing.
4. Now import the `registerFormatType` from this PR change and note how the type error is now fixed, and allows a couple new properties such as `attributes`.

## Screenshots or screencast 
image


mikeybinns:
I've investigated those failed tests, but they _seem_ to be unrelated to my PR, though it's possible I'm missing something. If it's something I can fix, I'd appreciate some advice on how to proceed.

mikeybinns:
Argh! sorry, was just trying to update my branch to the latest trunk version and it's messed up the PR, I will sort this or recreate the PR if needed.

mikeybinns:
Okay, after carefully considering some different options, I think while it's possible I could fix the branch in Git, I think GitHub has now subscribed a lot of people to this PR automatically, so to avoid annoying those people with updates to a PR they didn't subscribe to, I'm going to close this PR and open a new one. Sorry again.

PR #62772, opened 2024-06-24
Update react.js regarding useEffect old url

anamwp:
## What?
Update auto generated url of useEffect to new site url

## Why?
When user reading doc regarding useEffect it reference to the old url.

## How?
I update old url to new url manually

## Testing Instructions
Open this url https://developer.wordpress.org/block-editor/reference-guides/packages/packages-element/#useeffect
Click on the related link. It should be https://react.dev/reference/react/useEffect

### DIFF
--- a/packages/element/src/react.js
+++ b/packages/element/src/react.js
@@ -170,7 +170,7 @@ export { useDebugValue };
 export { useDeferredValue };
 
 /**
- * @see https://reactjs.org/docs/hooks-reference.html#useeffect
+ * @see https://react.dev/reference/react/useEffect
  */
 export { useEffect };
 


t-hamano (commented):
Thanks for the PR!

We also need to update the README for this package, but [this section](https://github.com/WordPress/gutenberg/blob/trunk/packages/element/README.md#useeffect) is auto-generated from JSDoc.

Could you run `npm run docs:api-ref`, commit, and push?

Mamaduka:
Thanks for your contribution, @anamwp!

It looks like the URL has been updated in #63494. I'm going to close this PR.

PR #23677, opened 2020-07-03
Image editing: move zoom control to toolbar dropdown

ellatrix:
## Description

See #22569.

This PR moves the zoom range control to the block toolbar as a dropdown.

* The zoom control doesn't belong in the block itself.
* Worth noting that this slider is a back-up for zooming the image. You can also "scroll" or pinch the image to zoom (and move!) the image.
* There was talk to put the slider in the toolbar itself, instead of a drop down.
  * Horizontal arrow keys will clash, so it would need a special way to enter and leave the control with the keyboard.
  * Not sure if it's worth overloading the toolbar with a controls that's meant as a back-up.
  * Tbh, sometimes I wonder if we need this control at all. You might be able to zoom, but you still can't move the image with keyboard only. 🙂

## Screenshots 

Screenshot 2020-07-03 at 15 04 42

## Checklist:
- [ ] My code is tested.
- [ ] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR.

PR #62776, opened 2024-06-24
Image Block: fix Aspect Ratio button position

t-hamano:
## What?

This PR fixes the position of the Aspect Ratio button in the block toolbar of the image block. This issue only occurs when the Top Toolbar is enabled.

aspect-ratio

## Why?

This issue appears to have first appeared in WP6.5. It was probably caused by a styling change to the header block toolbar starting with #57444.

## How?

Removed the styles specific to this button. These styles were added about 4 years ago in #23677. Maybe the styles don't fit the current block toolbar.

## Testing Instructions

- Insert an image block and apply an image.
- Click the Crop button.
- Whether the Top Toolbar is enabled or not. All button positions should be correct.

### DIFF
--- a/packages/block-editor/src/components/image-editor/aspect-ratio-dropdown.js
+++ b/packages/block-editor/src/components/image-editor/aspect-ratio-dropdown.js
@@ -78,7 +78,6 @@ export default function AspectRatioDropdown( { toggleProps } ) {
 			label={ __( 'Aspect Ratio' ) }
 			popoverProps={ POPOVER_PROPS }
 			toggleProps={ toggleProps }
-			className="wp-block-image__aspect-ratio"
 		>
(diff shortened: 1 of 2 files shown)


jasmussen (approved):
Nice:

Screenshot 2024-06-24 at 10 12 39
Screenshot 2024-06-24 at 10 12 46

Let's backport this, and thank you.

t-hamano:
WP 6.6 RC1 is tomorrow and it seems there are still other tests failing 😅 This issue isn't the first time it's happened with WP6.6 so it might be best not to backport it.

jasmussen:
The code is safe, though, and it's definitely a bugfix. If it can't be backported, so be it, but it would be a pity.

ellatrix:
This looks fine to get it 👍

PR #61025, opened 2024-04-24
Grid interactivity: Allow blocks to be positioned in manual mode using drag and drop

noisysocks:
This PR adds the ability to explicitly position a grid item by dragging and dropping it to a cell.

In order to allow more precise manipulation of the grid, when switching to manual mode all grid items are fixed in place. They can then be moved around with drag and drop, with the block movers (in any direction) or by changing the values of column and row in the Dimensions sidebar.

Positioning a grid item is done by setting the style.layout.columnStart and style.layout.rowStart attributes. These were added in https://github.com/WordPress/gutenberg/pull/59483.

The drag and drop mechanism works by having a DropZone in each of the GridVisualizer's cells. There's some trickery here to work around the fact thatGridVisualizer needs pointer-events: none so that you can interact with the blocks that are underneath the visualiser. Pointer events are disabled on the cells until the user begins to drag a block, at which point we enable pointer events so that the drop zones work as expected.

Steps for testing:

1. In Gutenberg > Experiments, enable "Grid interactivity"
2. Add a Grid block to a post or template and add a few items inside it.
3. Set the Grid to Manual mode; adjust Columns and Rows as desired. 
4. Try dragging and dropping items from one cell to another.

https://github.com/WordPress/gutenberg/assets/8096000/99ed66aa-cd58-401d-91f1-b2c45cb1edce

Note that if a block is too big for a particular cell and there are not enough empty adjacent cells to fit it, the drop zone won't appear.

Also note that in this configuration it's possible to expand a grid item in a way that causes it to overlap with adjacent blocks (because all blocks in this mode have fixed column and row starts). This doesn't feel ideal.

PR #60941, opened 2024-04-22
try adding max columns to grid

SaxonF:
## What?
Adds a maximum column control to grid when in auto mode. 

## Why?
If we move forward with all grid items being manually positioned when in manual mode, we'll be missing a way to create an adaptive grid with a set number of columns.

Issue #57478, opened 2024-01-02
Grid Layout Improvements


tellthemachines:
Part of [Design Tools overview](https://github.com/WordPress/gutenberg/issues/33447) and [Layout improvements](https://github.com/WordPress/gutenberg/issues/57571).

The Grid layout type [shipped in WP 6.3](https://make.wordpress.org/core/2023/07/14/layout-updates-in-the-editor-for-wordpress-6-3/) and in [WP 6.6](https://make.wordpress.org/core/2024/06/18/grid-layout-type/) a new Grid variation for the Group block was stabilised, along with support for grid children spanning multiple columns and rows, and on-canvas resize handles. 

## What's this currently look like?

Under Gutenberg > Experiements, enable "Grid interactivity".

When a Grid block is set to Manual mode (in the block sidebar under Layout), it's [now possible](https://github.com/WordPress/gutenberg/pull/61025) to drag and drop blocks into specific grid cells, as well as moving them up, down, left and right with the movers:

https://github.com/WordPress/gutenberg/assets/8096000/afe0fee0-375d-499e-8513-115a0192f843

### Needed before we remove experimental flag

- [x] https://github.com/WordPress/gutenberg/pull/62777
- [x] https://github.com/WordPress/gutenberg/pull/63391
- [x] https://github.com/WordPress/gutenberg/pull/63464
- [x] https://github.com/WordPress/gutenberg/pull/63395
- [x] https://github.com/WordPress/gutenberg/pull/63333
- [x] https://github.com/WordPress/gutenberg/pull/63394
- [x] https://github.com/WordPress/gutenberg/issues/63275
- [x] Fix miscellaneous bugs reported in https://github.com/WordPress/gutenberg/issues/62928
- [x] https://github.com/WordPress/gutenberg/pull/63108
- [x] https://github.com/WordPress/gutenberg/issues/62518
- [x] https://github.com/WordPress/gutenberg/pull/62894
- [x] #63973
- [x] #63976
- [ ] https://github.com/WordPress/gutenberg/issues/64289
- [ ] Grid guides don't always update on changing column count when grid only has text content.
- [ ] Min width control bugs: using the slider reverts the units to `rem`; switching between units causes layout shift.

### Needed before merge to Core

- [x] https://github.com/WordPress/gutenberg/issues/63279
- [x] #63979
- [ ] #63974
- [x] https://github.com/WordPress/gutenberg/issues/63965
[...]

PR #62777, opened 2024-06-24
Try using `manualPlacement` attribute to set manual grid mode and allow responsive behaviour in both modes.

tellthemachines:
## What?

Follow-up to #61025; addresses part of #57478; alternative to #60941.

* In #61025 we're changing the distinction between Manual and Auto modes to be about placement: in Auto mode, grid children occupy the first available grid cell, while in Manual all grid children can be placed in any grid cell.
* To provide the same functionality as Manual mode did previously, this PR allows setting either/both `minimumColumnWidth` and `columnCount` in Auto mode. When setting both, the grid becomes responsive with a maximum column count.
* This PR also adds `minimumColumnWidth` as an option in Manual mode, which enables the grid to become responsive in that mode.

## Testing Instructions

3. In Auto mode, try setting"Columns" and resize the grid (it should max at the set column number but resize down); 
4. Try setting "Minimum column width" to 0 and resize the grid (it should maintain same number of columns);
5. In Manual mode, try setting "Minimum column width" and check that the grid resizes.

## Screenshots or screencast 

Auto mode:

https://github.com/WordPress/gutenberg/assets/8096000/606a4e80-8a81-4041-ad66-446a84af50da

Manual mode:

https://github.com/WordPress/gutenberg/assets/8096000/b4bbdd0e-c6e7-4cc3-b94d-e1eb16e0e322


tellthemachines:
I think this is in a good place functionality-wise, so marking ready for review!

jasmussen:
Thanks for keeping at this. I tried the test steps and things seem to work as intended, though I'm curious what's actually new. The min-column width was already in place, no? Forgive me if I'm missing something, CSS grid is still pretty complex for me.

Sidenote, did the grid-outline-indicators change to become dotted? I'd personally prefer still they are solid lines rather than dots; the dots become busy due to their more complex silhouette.

packages/block-editor/src/layouts/grid.js
```diff
@@ -264,18 +284,22 @@ function GridLayoutColumnsAndRowsControl( {
...
-									columnCount: validValue,
+									columnCount:
+										window.__experimentalEnableGridInteractivity
+											? parseInt( value, 10 ) || null
+											: validValue,
 								} );
 							} }
 							value={ columnCount }
-							min={ 1 }
+							min={ 0 }
```
andrewserong, line 307: Just curious, but why are the min values changes to `0`?

packages/block-editor/src/layouts/grid.js
```diff
@@ -322,21 +347,37 @@ function GridLayoutTypeControl( { layout, onChange } ) {
 	const [ tempColumnCount, setTempColumnCount ] = useState(
 		columnCount || 3
 	);
+	const [ tempRowCount, setTempRowCount ] = useState( rowCount || 2 );
```
andrewserong: Why does row count receive a default of 2? Will this feature still support Grid layouts that shouldn't specify a number of rows — e.g. the post template block?

I noticed that with the post template block, when I toggle between auto and manual the height jumps so I wondered if that's because of the default row value 🤔

https://github.com/WordPress/gutenberg/assets/14988353/14b12e3f-5707-4e5d-b9da-7bf5b1025191

tellthemachines:
> I'm curious what's actually new. The min-column width was already in place, no?

Min column width is already in place but only for Auto grids. What this PR changes is adding it to Manual grids too, so they can become responsive.
The other change is adding the colums control to Auto mode, so that responsive grids can have a max number of columns.

This only makes sense after #61025, which added custom placement to Manual mode. Essentially, the difference between Auto and Manual now is how the grid children are placed:

* In Auto, children are auto-placed on the grid in consecutive cells;
* In Manual, children are explicitly placed in their cells with `grid-column` and `grid-row` attributes, which is why we can drag and drop them into a specific cell.

I hope this makes it a bit clearer! I also did a demo of the responsive behaviour in both modes in the [recent hallway hangout ](https://www.youtube.com/watch?v=3LJWLOjOn5o).

> Sidenote, did the grid-outline-indicators change to become dotted?

I can see this on trunk when a grid child is selected, but if the grid block itself is selected the indicator becomes a solid line. Let me see what might have borked it! The good news is it's appearing correctly in 6.6 😅

andrewserong (commented):
Very cool, I particularly like the responsive behaviour in manual mode from the inclusion of min column width + column count 👍

Just left a couple of tiny questions, apologies if they were already answered. The main issue I ran into while testing was with manually changing the row count. Sometimes when editing the field manually, the field didn't update immediately or was slow to respond. when I quickly clicked on the number spinner, I then ran into a maximum update depth exceeded error: 
image

I'm not too sure what wound up causing it, but here's the block markup I had when the error occurred in case it helps:

<details>

<summary>Block markup</summary>

```html
<!-- wp:group {"layout":{"type":"grid","columnCount":12,"rowCount":1,"manualPlacement":true,"minimumColumnWidth":"200px"}} -->
<div class="wp-block-group"><!-- wp:paragraph {"style":{"layout":{"columnStart":1,"rowStart":1}}} -->
<p>A paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":2,"rowStart":1}}} -->
<p>Another paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":3,"rowStart":1}}} -->
<p>Another paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":1,"rowStart":2}}} -->
<p>Another paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":2,"rowStart":2}}} -->
<p>Another paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":3,"rowStart":2}}} -->
<p>Another paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":1,"rowStart":3}}} -->
<p>Another paragraph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":2,"rowStart":3}}} -->
<p>Another one</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":3,"rowStart":3}}} -->
<p>Another paragarph</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"layout":{"columnStart":1,"rowStart":4}}} -->
<p></p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->

<!-- wp:paragraph -->
<p></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p></p>
<!-- /wp:paragraph -->
```

</details>

Happy to do further testing next week if you need any more info!

packages/block-editor/src/layouts/grid.js
tellthemachines: We need to be able to unset this value so we can have auto-reflowing columns with no max column number. Given that Layout doesn't use a toolspanel there's no convenient Reset option available, and I didn't want to introduce random new UI here 😅

packages/block-editor/src/layouts/grid.js
tellthemachines: Oh well spotted! There shouldn't be a minimum Row number at all.

tellthemachines:
Thanks for giving this a test run!

> when I quickly clicked on the number spinner, I then ran into a maximum update depth exceeded error:

This should be fixed now. It was due to the effect inside `useGridLayoutSync` being re-run every time the layout object changed 😅

packages/block-editor/src/layouts/grid.js
tellthemachines: Fixed!

andrewserong:
Thanks for the updates! Not sure if it was introduced in this PR, but I notice with the experiment switched off that the range control in Manual mode isn't aligned correctly to the column count:

| Auto (correct) | Manual (not centered) |
| --- | --- |
| image | image |

Also, with the experiment off and selecting a Post Template block, the help text isn't quite right. In manual mode, we're setting the number of columns but in this context manual placement doesn't apply, but the help text says "Grid items can be manually placed in any position on the grid."

tellthemachines:
> Also, with the experiment off and selecting a Post Template block, the help text isn't quite right. In manual mode, we're setting the number of columns but in this context manual placement doesn't apply, but the help text says "Grid items can be manually placed in any position on the grid."

Hmm, that help text was added in #61025 and inadvertently exposed in non-experimental mode; I'll put it behind the flag again!

tellthemachines:
> with the experiment switched off that the range control in Manual mode isn't aligned correctly to the column count

This should also be fixed now!

ramonjd:
Hi! I'm a bit of a grid newbie, so I think I'm doing something wrong.

Is there a difference between a null value and 0px for min column width? I'm not sure what I'm supposed be seeing when I set it to 0px using the slider vs deleting the value.

Most of the video is me fumbling about, but at around 18 seconds I toggle between 0px and empty value and see different column counts. The video is on manual mode, but I get the same for Auto mode.

https://github.com/WordPress/gutenberg/assets/6458278/158687ea-1ae5-4fc1-8916-53dc3a5c4c51

Thanks!

tellthemachines:
> at around 18 seconds I toggle between 0px and empty value and see different column counts.

Thanks for testing! that definitely shouldn't be happening, looking into it now!

tellthemachines:
Ok, minimum column width behaviour should be fixed now!

noisysocks:
There's a few bugs when you try to set a manual number of rows.

I'm not seeing row lines appear in the grid visualiser when I increase the number of rows:

https://github.com/WordPress/gutenberg/assets/612155/7ceece15-01f0-4da8-a569-367c3497fa65

The number of rows is reset to 0 (or 1, if there's inner blocks) when I re-select the grid block:

https://github.com/WordPress/gutenberg/assets/612155/e24f7fac-03c1-4e9f-9215-4f1f71041424

packages/block-editor/src/components/grid/use-grid-layout-sync.js
```diff
@@ -11,31 +11,22 @@ import { store as blockEditorStore } from '../../store';
...
-		},
-		[ gridClientId ]
-	);
-
-	const { getBlockAttributes } = useSelect( blockEditorStore );
+	const { getBlockOrder, getBlockAttributes } = useSelect( blockEditorStore );
 	const { updateBlockAttributes, __unstableMarkNextChangeAsNotPersistent } =
 		useDispatch( blockEditorStore );
 
+	const blockOrder = getBlockOrder( gridClientId );
```
noisysocks: Why are we now calling `getBlockOrder` outside of the `useSelect`?  If we want it to be a dep of the `useEffect` then it needs to be called inside the `useSelect` so that it updates when the store is updated.

packages/block-editor/src/components/grid/use-grid-layout-sync.js
```diff
@@ -11,31 +11,22 @@ import { store as blockEditorStore } from '../../store';
...
-
-		const { columnCount, rowCount = 2 } = gridLayout;
-		const isManualGrid = !! columnCount;
+		const { columnCount, rowCount, manualPlacement } = gridLayout;
+		const isManualGrid = !! manualPlacement;
 
 		if ( isManualGrid ) {
 			const rects = [];
 			let cellsTaken = 0;
+			let lowestRowEnd = 0;
```
noisysocks: This is a somewhat confusing name because "lowest row end" could mean the bottom-most row or the smallest "row end".

packages/block-editor/src/components/grid/use-grid-layout-sync.js
```diff
@@ -104,6 +88,15 @@ export function useGridLayoutSync( { clientId: gridClientId } ) {
 						},
 					},
 				};
+				lowestRowEnd = Math.max( lowestRowEnd, rowStart + rowSpan - 1 );
+			}
+			if ( rowCount !== lowestRowEnd ) {
+				updates[ gridClientId ] = {
+					layout: {
+						...gridLayout,
+						rowCount: minimumNeededRows,
```
noisysocks: Is this supposed to be `lowestRowEnd`?

noisysocks (commented):
Nice. I really like that this clarifies that Auto / Manual are for placement and nothing else.

Should the `manualPlacement` attribute begin with an `is` prefix since it's a boolean? I can't remember if we do that for attributes or not.

I don't really like the UX around resetting column and row values. We might want to bring in `ToolsPanel`. Also the mix of input controls and range sliders looks cluttered to me. Something to iterate on in follow-ups.

packages/block-editor/src/components/grid/use-grid-layout-sync.js
tellthemachines: It doesn't seem to make a difference - on trunk, if I add a new block to a grid by pressing enter on a Paragraph block already inside it, the effect doesn't re-run. That might be due to this hook rendering inside the grid component. Ideally we get it to update whenever a grid child is added or updated, but I think that'll require bigger changes.

noisysocks: I think you're right. We'll need to move `<GridLayoutSync>` out of `toolBarControls` and into a component that's not just rendered when the grid block is selected.

Regardless, I don't think this PR should be moving `getBlockOrder` outside of the `useSelect` 😀 If it were working on trunk I'd expect this change to break it.

tellthemachines:
Thanks for reviewing @noisysocks!

> I'm not seeing row lines appear in the grid visualiser when I increase the number of rows:

This was a side-effect of addressing the feedback about uniform row height in #62928. Because I changed the default row size to `auto`, empty rows collapse. I've now compromised on `minmax(8px, auto)` so there's always a min height.

Empty rows should now also be preserved.

packages/block-editor/src/components/grid/use-grid-layout-sync.js
tellthemachines: `bottomestRow`? 😄

tellthemachines:
> Should the `manualPlacement` attribute begin with an `is` prefix since it's a boolean? I can't remember if we do that for attributes or not.

I'm not sure we have that many boolean attributes. The approach tends to be to deduce state from whatever attribute values are present, not to use explicit booleans. But in this case we need the boolean, so if `isManualPlacement` makes it clearer then let's change it!

packages/block-editor/src/components/grid/use-grid-layout-sync.js
tellthemachines: I've reverted this for now and will move the hook in the PR that separates the visualizer out from the grid toolbar (working branch [here](https://github.com/WordPress/gutenberg/compare/try/inserter-in-grid?expand=1), still very early stages)

noisysocks:
Looks like our naming is a mixed bag. Sometimes we use `is*` e.g. `isLink` but mostly we use a verb e.g. `showComments`. So... I don't know. Up to you 😅

https://github.com/search?q=repo%3AWordPress%2Fgutenberg%20%22type%22%3A%20%22boolean%22&type=code

In testing, I am no longer seeing those two issues 👍 but now when I switch from Auto to Manual all of the grid items are put into the first cell:

https://github.com/WordPress/gutenberg/assets/612155/54c1bdb6-79a3-4caa-a210-2ad430f99ab8

tellthemachines:
oooh that's what `minimumNeededRows` was doing 😅

tellthemachines:
Ok I think it's working properly now:

* switching from Manual to Auto and back to Manual preserves empty rows and leaves blocks positioned sequentially
* switching from an Auto to Manual for the first time (no rows explicitly defined) positions blocks sequentially and adds the minimum needed amount of rows

packages/block-editor/src/layouts/grid.js
```diff
@@ -294,10 +335,10 @@ function GridLayoutColumnsAndRowsControl( {
 								onChange={ ( value ) =>
 									onChange( {
 										...layout,
-										columnCount: value,
+										columnCount: parseInt( value, 10 ),
```
noisysocks: `RangeControl` provides `number | undefined` to `onChange` so this `parseInt` is unnecessary.

noisysocks (approved):
It works! Good job 👍

packages/block-editor/src/layouts/grid.js
tellthemachines: Oh I meant to do that to the NumberControl above 🤦

Issue #62752, opened 2024-06-21
Design tools exploration: Contextual color controls and new panels for Background, and Elements


jameskoster:
Extracted from discussion in https://github.com/WordPress/gutenberg/pull/60100#issuecomment-2180553281 and https://github.com/WordPress/gutenberg/pull/60151.

In container blocks like Group, the Color panel provides options to apply colors to child elements such as buttons, links, and headings.

<img width="937" alt="color" src="https://github.com/WordPress/gutenberg/assets/846565/a55293a6-3784-4fe5-9c15-52f9ede411e8">

While this feature is useful, it has limitations – for example, it doesn't allow for adjusting the _dimensions_ of buttons within the Group.

Another issue is the disorganized background options: the background color control is in the Color panel, while the background image control is in a separate Background Image panel.

### Contextual color controls
<img width="937" alt="panels" src="https://github.com/WordPress/gutenberg/assets/846565/8ee16faf-b0c9-4d29-9ea5-3b8cb25bee37">

I suggest moving text and background controls to more appropriate panels. The text color control could be relocated to the Typography panel, and the background color control to a new Background panel, alongside the background image control.

A dedicated Background panel would allow users to add multiple background images and reorder them as needed.

### Child elements

<img width="937" alt="elements" src="https://github.com/WordPress/gutenberg/assets/846565/1fa303fd-d0f4-480b-a3a3-5640cf6ae2e3">

To enhance functionality, a new 'Elements' panel could enable users to fully style child elements within a container. This approach is similar to styling blocks in Global styles, which has been quite effective.

---

Additionally, this reorganization could simplify the 'Overlay' option in the Cover block, combining background images and transparent background colors into a single control that can be applied to any block, not just the Cover block.

<img width="937" alt="overlay" src="https://github.com/WordPress/gutenberg/assets/846565/cfec1cac-52d2-4e91-ac74-c4c379d3e7dd">

cc @WordPress/gutenberg-design for thoughts.

richtabor:
Interesting exploration. Would background controls open via a popover?

jameskoster:
Precisely. There's a PR for that over here: https://github.com/WordPress/gutenberg/pull/60151

PR #62779, opened 2024-06-24
Flex dimensions: Rename "Fill" to "Grow".

jasmussen:
## What?

Flex tools have a dimensions control that lets you choose between Fit, Fill and Fixed:

before

This PR renames "Fill" to "Grow":

after

## Why?

* As background image tools and color controls expand and possibly consolidate (see also #62752), the word "Fill" is increasingly going to refer to background-fills, rather than the behavior of filling out space.
* "Grow" both matches some of the flex properties it works with, but it also intuitively suggests its behavior in a way that's arguably less clear with "Fill". 

Here's a menu-design use case, as an example:

menu test

In that GIF, a horizontal navigation menu featureing a site title, a menu, and a button is built. In this, the site title is set to "fill", so it pushes the navigation menu and the button to the right most side of the header, accomplishing the desire design. But it would easily be confused in this case, to suggest that "fill" should fill the available space, and not also imply pushing the other content rightwards. Whereas "Grow" does suggest filling the available space, but also implies pushing other contents aside.

## Testing Instructions

Insert a row block with a couple of child blocks. View the inspetor for a child block, and see "Grow" instead of "Fill".

### DIFF
--- a/packages/block-editor/src/components/child-layout-control/index.js
+++ b/packages/block-editor/src/components/child-layout-control/index.js
@@ -12,7 +12,7 @@ import {
 	Flex,
 	FlexItem,
 } from '@wordpress/components';
-import { __ } from '@wordpress/i18n';
+import { __, _x } from '@wordpress/i18n';
 import { useEffect } from '@wordpress/element';
 import { useSelect, useDispatch } from '@wordpress/data';
 
@@ -142,17 +142,26 @@ function FlexControls( {
 				<ToggleGroupControlOption
 					key="fit"
 					value="fit"
-					label={ __( 'Fit' ) }
+					label={ _x(
+						'Fit',
+						'Intrinsic block width in flex layout'
+					) }
 				/>
 				<ToggleGroupControlOption
 					key="fill"
 					value="fill"
-					label={ __( 'Grow' ) }
+					label={ _x(
+						'Grow',
+						'Block with expanding width in flex layout'
+					) }
 				/>
 				<ToggleGroupControlOption
 					key="fixed"
 					value="fixed"
-					label={ __( 'Fixed' ) }
+					label={ _x(
+						'Fixed',
+						'Block with fixed width in flex layout'
+					) }
 				/>
 			</ToggleGroupControl>
 			{ selfStretch === 'fixed' && (
--- a/packages/block-editor/src/components/child-layout-control/index.js
+++ b/packages/block-editor/src/components/child-layout-control/index.js
@@ -147,7 +147,7 @@ function FlexControls( {
 				<ToggleGroupControlOption
 					key="fill"
 					value="fill"
-					label={ __( 'Fill' ) }
+					label={ __( 'Grow' ) }
 				/>
 				<ToggleGroupControlOption
 					key="fixed"


fditrapani:
Nice! I think the label "Grow" works well here @jasmussen. 

Do we also need to update the preview text that shows under the button. I'm not set up to run this yet so I can't tell if the copy `Stretch to fill available space.`  is going to be updated or not. Perhaps it would make sense to say `Grow to fill the available space` to match the UI update better. 

While we're at this. What do you think of updating "Fit" as well? The term fit isn't very clear. When I read it, it makes me think fit within the space. Taking inspiration from Figma and calling this Hug might work better. The corresponding descriptor in this case could read "Hug content".

t-hamano (commented):
I don't have a strong opinion on the label, but translators who see it may not know how to translate it into their own language. This word, at least in Japanese, is translated into various words depending on the context. The word "Grow" is more abstract than the word "Fill".

Therefore, these options may need context. For example:

```jsx
<ToggleGroupControlOption
	key="fit"
	value="fit"
	label={ _x( 'Fit', 'Block width in flex layout' ) }
/>
<ToggleGroupControlOption
	key="fill"
	value="fill"
	label={ _x( 'Grow', 'Block width in flex layout' ) }
/>
<ToggleGroupControlOption
	key="fixed"
	value="fixed"
	label={ _x( 'Fixed', 'Block width in flex layout' ) }
/>
```

I am also concerned that the mismatch between the `label` and `value` might cause some problems. cc @tellthemachines

tellthemachines:
> I am also concerned that the mismatch between the `label` and `value` might cause some problems. cc @tellthemachines

Label and value are independent so this shouldn't cause any problem.

dionysous:
Regarding the label: In German, the term "Grow" typically describes a process rather than a static state. For a flex item, it might cause confusion as it implies an ongoing action rather than describing the end state of how it should look in the layout. "Fill" more intuitively conveys the idea of occupying available space and represents the end state more clearly.

jasmussen:
Good feedback, keep it coming and feel free to also suggest help text improvements.

> Regarding the label: In German, the term "Grow" typically describes a process rather than a static state. 

Is there a different word for "Grow" in German that would work better here? In Danish there are a couple that could work, and if we can provide sufficient context for translators, perhaps a good term could be found? Perhaps even the existing one? Mainly, this feels like one of those aspects where the term deserves care for each language.

dionysous:
The difference in meaning is this: "Wachsen" (grow) implies a natural, organic process of increasing in size over time, similar to how a plant grows. "Erweitern" (extend), on the other hand, suggests a deliberate action to increase in size or capacity, which is more appropriate for a flex item that adjusts to fill available space. Providing sufficient context for translators would be good, yes.

t-hamano:
I asked the chat GPT for help 😄

https://chatgpt.com/share/8db0b22f-b875-4c32-8fa5-ed1b61850bd0

They suggested the following English words, what do you think?

1. **Expand**: The element expands to fill available space.
2. **Stretch**: The element stretches to occupy the remaining space.
3. **Grow**: The element grows to fill the extra space.
4. **Flexible**: The element is flexible and adapts to the container's available space.
5. **Adaptable**: The element adapts to take up any additional space.
6. **Fill**: The element fills the leftover space.
7. **Extend**: The element extends to occupy the remaining area.

jasmussen:
Thank you all for chiming in.

I've grown _quite_ fond of "Grow" for english, and I'm wondering, would it be able to have that term, but provide clear enough instructions for translators that for example German might translate it to "Erweitern"?

dominique-vias:
I interpret _Grow_ more as _Hug_, as in it grows as the elements it contains increase, and it happens incrementally.

jameskoster:
I suspect this is highly subjective, so a bit tricky to find definitive consensus. 

'Fill' is slightly more intuitive to me, but that's probably personal bias due to years of Figma use:
 
Screenshot 2024-07-03 at 09 13 23

I appreciate that 'Grow' aligns more with the underlying css property... the arguments for change are solid. On that basis renaming seems fine, so long as there are no strong objections, and hints are provided for translators.

t-hamano:
> I appreciate that 'Grow' aligns more with the underlying css property... the arguments for change are solid. On that basis renaming seems fine, so long as there are no strong objections, and hints are provided for translators.

I see, so should we decide to go with "Grow"?

To ship this PR, I think we need to resolve the conflicts and add translation context.

jasmussen:
> I see, so should we decide to go with "Grow"?

There's been enough user feedback that "fill" is confusing, so yes, I think it's worth trying "grow" and see if people find it more intuitive. 

> To ship this PR, I think we need to resolve the conflicts and add translation context.

I'll take a stab in a minute! Thank you.

jasmussen:
Tried to add very specific translation contexts. Thank you for the code samples. Let me know if this works for you.

t-hamano (approved):
LGTM! However, since English is not my native language, I would appreciate it if someone could give us feedback on whether the context strings are appropriate 😅

fditrapani (approved):
Looks good to me too!

jasmussen:
Cool, let's try this! And I'm here for followups.

PR #62786, opened 2024-06-24
Fix: List view renders an empty menu when no actions are eligible.

jorgefilipecosta:
If no actions are eligible on the list view we render an empty menu. We are checking for actions?.length while we should check for eligibleActions?.length. This PR fixes the issue.

## Testing Instructions
I pasted the following code snippet:
```
function remove_edit_delete_post_capability( $caps, $cap, $user_id, $args ) {
    if ( 'edit_post' === $cap || 'delete_post' === $cap) {
            $caps[] = 'do_not_allow';
        }
    }

    return $caps;
}
add_filter( 'map_meta_cap', 'remove_edit_delete_post_capability', 10, 4 );
```

I went to the trash pages section `wp-admin/site-editor.php?postType=page&layout=list&activeView=trash`
I verified the actions menu is not available. On trunk it is and renders an empty menu:
Screenshot 2024-06-21 at 17 25 03

### DIFF
--- a/packages/dataviews/src/view-list.tsx
+++ b/packages/dataviews/src/view-list.tsx
@@ -184,7 +184,7 @@ function ListItem< Item extends AnyItem >( {
 						</HStack>
 					</CompositeItem>
 				</div>
-				{ actions?.length > 0 && (
+				{ eligibleActions?.length > 0 && (
 					<HStack
 						spacing={ 1 }
 						justify="flex-end"


ellatrix (approved):

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: bc3d7c93c2

PR #62787, opened 2024-06-24
Fix: DataViews: Active template is not highlighted properly in list view

jorgefilipecosta:
Follow up to https://github.com/WordPress/gutenberg/pull/62378.
Applies the same approach we had to highlight the active page to the templates.

Move the already existing hook usePostIdLinkInSelection to a utils file with a comment it is now used in PagePages and PageTemplates.
Uses the hook in the templates list.

## Testing Instructions
Go to wp-admin/site-editor.php?postType=wp_template&layout=list&postId=twentytwentyfour%2F%2Fhome (twentytwentyfour%2F%2Fhome needs to be a valid template id)
Verify the active template is highlighted properly (on trunk it is not).

### DIFF
--- a/packages/edit-site/src/components/page-pages/utils.js
+++ b/packages/edit-site/src/components/page-pages/utils.js
@@ -1,0 +1,44 @@
+/**
+ * WordPress dependencies
+ */
+import { useState, useEffect } from '@wordpress/element';
+import { privateApis as routerPrivateApis } from '@wordpress/router';
+
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { useLocation } = unlock( routerPrivateApis );
+
+// Used on PagePages and PageTemplates.
+export function usePostIdLinkInSelection(
+	selection,
+	setSelection,
+	isLoadingItems,
+	items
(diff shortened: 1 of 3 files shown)


packages/edit-site/src/components/page-pages/utils.js
```diff
@@ -0,0 +1,44 @@
...
+) {
+	const {
+		params: { postId },
+	} = useLocation();
+	const [ postIdToSelect, setPostIdToSelect ] = useState( postId );
+	useEffect( () => {
+		if ( postId ) {
+			setPostIdToSelect( postId );
+		}
+	}, [ postId ] );
```
ellatrix, line 29: I'm not quite following why this extra state is needed. Could you clarify it?

jorgefilipecosta: Hi @ellatrix, when we have a postId on the URL parameter we should include it in the selection. But we can not include it in the selection right way we can only include it in the selection after the data is loaded. So basically what we are doing is marking the postIdToSelect in state as a "Todo" and when items are loaded if there is a selection on "Todo" we change the selection.

PR #62788, opened 2024-06-24
Align naming modals

jameskoster:
Screenshot 2024-06-24 at 13 25 27

## What?
Apply consistent design and behavior to naming modals in the site editor. Specifically:

* Utilise one of the `size` options for `Modal` (most commonly `small`).
* Auto-focus the name input when the modal is opened.

## Why?
Consistency is good. Some modals (e.g. renaming a block) already exhibit these features.

## Testing Instructions
In the following flows, renaming should invoke a modal with a specific `size`, and the name input should be auto-focussed:

- Rename pattern from command palette
- Rename navigation menu from dark sidebar
- Rename entity from data view 🗒️ 
- Rename entity from Inspector 🗒️ 
- Rename pattern category
- Create new template from the Inspector
- Create Pattern from data view

🗒️  In these scenarios all action modals will receive the given props. There may be a better way to do this, but for now it could be okay given renaming is the only action that invokes a modal.


jasmussen (approved):
Nice cleanup. 👍 👍, probably just needs a code review.

ntsekouras (approved):
Looks good, thanks!

t-hamano:
It seems that when the modal mounts, the focus goes to the first element. This caused the E2E test where the Tab key is pressed to move focus to the Name field to no longer pass.

I fixed it so that it doesn't depend on the number of Tab key presses.

richtabor:
I'm getting this for the block rename modal. This should match the others, correct? 

CleanShot 2024-06-25 at 09 05 14

jameskoster:
I was mostly focussed on the modal size and auto-focussing the input in this PR. But yes, there's probably a case that the input format should be the same across.

I've made a note to circle back!

PR #62789, opened 2024-06-24
Docs/iapi warnings new directives wp 6 6

juanmaguitar:
## What?
Added proper callouts snippets at the [API Reference](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/api-reference) of the Interactivity API so they appear properly in the Block Editor Handbook

## Why?
To improve readability

## How?
By using proper callouts as documented [here](https://developer.wordpress.org/block-editor/contributors/documentation/#callout-notices)


cbravobernal (approved):
LGTM

PR #61430, opened 2024-05-07
Create Block: Match specified engines with Gutenberg and Core

Mamaduka:
## What?
This is a follow-up to #60962.

PR updates engines specified in the `package.json` file to match [Gutenberg](https://github.com/WordPress/gutenberg/blob/e92b45adb5b10dd7d7eb9162d908f7493b94ab4d/package.json#L17-L20) and [WP Core](https://github.com/WordPress/wordpress-develop/blob/b0308b8f5299362c8a12e45e83d6a26cb01d7325/package.json#L9-L12).
CI checks should be green.

## Why?
The `npm install` might fail when using node version managers due to a version mismatch.

### Testing Instructions for Keyboard
Same.

PR #62791, opened 2024-06-24
Create Block: Add missing changelog entries

gziolo:
## What?

Follow-up for #61430 and #61430.

Raised by @fabiankaegy in https://github.com/WordPress/gutenberg/pull/61430#issuecomment-2172596305.

## Why?

> The way this was released means that any tool that isn't yet on node 20 broke when NPM updated the dependency if it was installed with the default ^ version selector.

## How?

Add  missing changelog entries.


fabiankaegy (approved):
Thanks for adding the notes here :)

PR #62794, opened 2024-06-24
Cache theme and core data

kt-12:
https://core.trac.wordpress.org/ticket/59600

## What?
Caching persistently inside `get_theme_data` and `get_core_data` is estimated to provide a performance benefit of at least ~2%. Block data from `theme` and `core` are relatively static, so they can be persistently cached for a shorter period.

## Why?
For faster load time.

## How?
By caching and skipping constructor calls to `WP_Theme_JSON | WP_Theme_JSON_Gutenberg`, which is costly.

### DIFF
--- a/lib/class-wp-theme-json-resolver-gutenberg.php
+++ b/lib/class-wp-theme-json-resolver-gutenberg.php
@@ -158,12 +158,30 @@ class WP_Theme_JSON_Resolver_Gutenberg {
 	 * @return WP_Theme_JSON_Gutenberg Entity that holds core data.
 	 */
 	public static function get_core_data() {
+		$can_use_cached = ! wp_is_development_mode( 'theme' );
+		if ( $can_use_cached ) {
+			$cache_key = 'core_data';
+			if ( null === static::$core ) {
+				$cache_value = static::get_cache_data( $cache_key );
+				if ( $cache_value ) {
+					// Defined in this function below.
+					$theme_json   = apply_filters( 'wp_theme_json_data_default', $cache_value );
+					static::$core = $theme_json->get_theme_json();
+				}
+			}
+		}
+
 		if ( null !== static::$core && static::has_same_registered_blocks( 'core' ) ) {
 			return static::$core;
 		}
 
 		$config = static::read_json_file( __DIR__ . '/theme.json' );
 		$config = static::translate( $config );
+
+		$theme_json = new WP_Theme_JSON_Data_Gutenberg( $config, 'default' );
+		if ( $can_use_cached ) {
+			static::set_cache_data( $cache_key, $theme_json );
+		}
 
 		/**
 		 * Filters the default data provided by WordPress for global styles & settings.
@@ -172,7 +190,7 @@ class WP_Theme_JSON_Resolver_Gutenberg {
 		 *
 		 * @param WP_Theme_JSON_Data_Gutenberg Class to access and update the underlying data.
 		 */
-		$theme_json   = apply_filters( 'wp_theme_json_data_default', new WP_Theme_JSON_Data_Gutenberg( $config, 'default' ) );
+		$theme_json   = apply_filters( 'wp_theme_json_data_default', $theme_json );
 		static::$core = $theme_json->get_theme_json();
 
 		return static::$core;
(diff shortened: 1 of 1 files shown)


lib/class-wp-theme-json-resolver-gutenberg.php
```diff
@@ -951,4 +1015,55 @@ private static function inject_variations_from_block_styles_registry( $data ) {
...
+	 * @return bool True if the value was set, false otherwise.
+	 */
+	private static function set_cache_data( $cache_key, $value ) {
+		// This should be inexpensive as DB only happenes once per pageload.
+		$cache = get_site_transient( 'wp_theme_json_resolver_cache' );
+		if ( null === $cache ) {
+			$cache = array();
+		}
+		$cache[ $cache_key ] = $value;
+		return set_site_transient( 'wp_theme_json_resolver_cache', $cache, 10 * MINUTE_IN_SECONDS );
```
joemcgill, line 1051: Since this transient is storing a combination of core data, theme data, and registered bocks, I think the will need to be site specific and use `set|get_transient()` rather than network wide (e.g., `set|get_site_transient()`) functions.

We should resolve the open questions about transients usage in [this comment](https://core.trac.wordpress.org/ticket/59595#comment:61) before using a similar approach here.

🔢 Applies throughout.

lib/class-wp-theme-json-resolver-gutenberg.php
```diff
@@ -158,21 +158,39 @@ protected static function translate( $theme_json, $domain = 'default' ) {
...
 	 */
 	public static function get_core_data() {
+		$can_use_cached = ! wp_is_development_mode( 'theme' );
+		if ( $can_use_cached ) {
+			$cache_key = 'core_data';
+			if ( null === static::$core ) {
+				$cache_value = static::get_cache_data( $cache_key );
+				if ( $cache_value ) {
+					// Defined in this function below.
+					$theme_json   = apply_filters( 'wp_theme_json_data_default', $cache_value );
```
joemcgill, line 168: Any time we're passing theme.json data through a filter, or pulling from cache, I think we need to make sure it's been validated using the current `WP_Theme_JSON` schema expected by the application, similar to what we did in https://github.com/WordPress/wordpress-develop/commit/6ae489ccaeb1e426d496183c2c1c997295f05d42, which hasn't been backported to this class yet.

In fact, it would probably be safer to cache the raw data (e.g. `WP_Theme_JSON_Data_Gutenberg::get_data()` )to ensure that the first time we're repopulating the data from cache, that we're passing it through the `WP_Theme_JSON_Data_Gutenberg()` constructor. That might negate the performance benefits by avoiding the initial `WP_Theme_JSON_Gutenberg` construction, but would good to check.

joemcgill (changes requested):
I like the idea here, but I worry about the memory consumption since we end up duplicating the theme.json data in the object cache (when pulled from the transient) and also in static methods of this object.

I think this could be cleaner by making the existing static properties an implementation detail of a generic cache getter and replacing the checks for the static properties with checks for a static value. Here's a pseudocode example:

```php
// In get_core_data().
public static function get_core_data() {
  $cached = get_cached_data( 'core' );
  
  if ( null !== $cached && static::has_same_registerd_blocks( 'core' ) ) {
    return $cached;
  }
  
  // Do non-cached logic here.
}

// Getting cached data.
private static function get_cached_data( $key ) {
  $can_use_cached = ! wp_is_development_mode( 'theme' );

  // Return cache from the transient if possible.
  if ( $can_use_cached ) {
    $cached = get_transient( 'wp_theme_json_resolver_cache' );
    return $cached[ $key ] ?? false;
  }

  // Return false for an empty value.
  return self::${ $key } ?: false;
}

lib/class-wp-theme-json-resolver-gutenberg.php
kt-12: We can try caching `get_data`, but it will negate all our effort as most of the performance regression came from `WP_Theme_JSON_Data_Gutenberg` constructor call.

For the other part, should we wait for the backporting, or should I copy past that here? I was avoiding it here because merging might become complicated. I have handled the merge for core PR.

joemcgill: > most of the performance regression came from WP_Theme_JSON_Data_Gutenberg constructor call

I thought that might be the case.

Let's handle backporting https://github.com/WordPress/wordpress-develop/commit/6ae489ccaeb1e426d496183c2c1c997295f05d42 separately. I'll get a PR open.

PR #62795, opened 2024-06-24
Fix lock modal dialog accessibility and semantics

### LINKED ISSUE #47248: Fix the Lock Block modal dialog group labelling and semantics
### Description

When locking a block, a modal dialog opens. So far, so good.
There are a few accessibility issues to fix in the modal dialog content though.

1
The checkboxes within the modal are grouped in a `div` element with `role=group` and `aria-labelledby`. Turns out the `aria-labelledby` ID reference value points to the first checkbox label text. As a result, the actual name of the group is 'Lock all':

https://github.com/WordPress/gutenberg/blob/37bb7c59f032afd034c136db18456388bb45c1fe/packages/block-editor/src/components/block-lock/modal.js#L111-L121

When tabbing into the group, screen readers will announce the group name 'Lock all", which is wrong. The complete announcement would be something like: `Lock all, unticked, thickbox, Lock all, group`:

Screenshot 2023-01-18 at 15 06 02

Instead, the group name should clarify what all the three checkboxes are about. That's something along the lines of 'Choose what to restrict or lock' (better wording needed).

Worth also reminding that a `<fieldset>` element with a visible `<legend>` element is always preferable to the ARIA pattern with `role=group`. One more thing to remind is that the labelling of a fieidset / group should be short: it's the group name. It's not the place for long descriptions or instructions.

2
The first checkbox `<label>` element contains a `<span>` element, not sure why. Please remove it. If an ID is needed, that should be set on the `<label>` element itself but it's likely that to fix the previous point an `aria-label` should be used in place of `aria-labelledby` so the ID won't be necessary any longer.

3
Checkboxes nesting. 
The first checkbox 'Lock all' is logically a 'first-level' option that contains tho logical 'sub-options'. However, only the second and third checkboxes are list items in a single-level unordered list. There's no great value in using a list this way. If the goal is to convey the logical nesting of the options, the 'Lock all' checkbox should be within a list and the other two checkboxes should be in a nested list.
Also, worth reminding that Safari doesn't expose to screen readers the list role when the list is styled. VoiceOver doesn't announce the list as a list, while other screen readers do. To cover this Safari + VoiceOver inconsistency, each `ul` element should have a (technically redundant) `role=list` attribute.

- Edit a post.
- Select a paragraph block and choose 'Lock' from its toolbar > Options menu.
- A 'Lock paragraph' modal dialog opens.
- Use the Chrome dev tools > accessibility tan to inspect the div element with role=group.
- Observe the div accessible name is 'Lock all'.
- Inspect the first checkbox.
- Observe its associated `label` element contains an unnecessary `span` element.
- Inspect the other two checkboxes.
- Observe they're within an unordered list, while the first checkbox isn't.

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/47248

## What?

Labeling and semantics of the block lock modal dialog can be improved.

## Why?

- Native HTML elements should be preferred instead of using ARIA roles and attributes. See: [First rule of ARIA use](https://www.w3.org/TR/using-aria/#rule1).
- The labeling of the checkboxes _group_ was incorrectly 'Lock all'.
- The checkboxes visual layout communicates a nesting hierarchy that should be communicated also semantically.

## How?

- Uses fieldset and legend elements instead of an ARIA role=group that was incorrectly labeled pointing at the label of the first checkbox.
- Makes all the checkboxes a nested list so that the semantics matches the visual nesting.
- Removes a no longer necessary span.
- Adjusts the CSS.

## Testing Instructions

- To see all the options in the lock modal dialog, add a Navigation block.
- Select the Navigation block and from its toolbar choose Options > Lock.
- The Lock modal dialog opens.
- Inspect the source and observe the form now contains a `<fieldset>` element with a `<legend>` element that gives the fieldset a name.
- Observe the visible instructions text is the actual legend.
- Observe the checkboxes are now placed in a nested unordered list.
- Observe there are no relevant visual changes. Only the top and bottom spacing of the legend text is now slightly taller and it's inline with the design grid system (16 pixels) while previously it was 1em (13 pixels) inherited from the WordPress `common.css`.

## Screenshots or screencast 

Screenshots before  and after:

before and after

### DIFF
--- a/packages/block-editor/src/components/block-lock/modal.js
+++ b/packages/block-editor/src/components/block-lock/modal.js
@@ -103,77 +103,101 @@ export default function BlockLockModal( { clientId, onClose } ) {
 							'Choose specific attributes to restrict or lock all available options.'
 						) }
 					</legend>
-					<CheckboxControl
-						__nextHasNoMarginBottom
-						className="block-editor-block-lock-modal__options-all"
-						label={ __( 'Lock all' ) }
-						checked={ isAllChecked }
-						indeterminate={ isMixed }
-						onChange={ ( newValue ) =>
-							setLock( {
-								move: newValue,
-								remove: newValue,
-								...( allowsEditLocking
-									? { edit: newValue }
-									: {} ),
-							} )
-						}
-					/>
-					<ul className="block-editor-block-lock-modal__checklist">
-						{ allowsEditLocking && (
-							<li className="block-editor-block-lock-modal__checklist-item">
-								<CheckboxControl
-									__nextHasNoMarginBottom
-									label={ __( 'Restrict editing' ) }
-									checked={ !! lock.edit }
-									onChange={ ( edit ) =>
-										setLock( ( prevLock ) => ( {
-											...prevLock,
-											edit,
-										} ) )
-									}
-								/>
-								<Icon
-									className="block-editor-block-lock-modal__lock-icon"
-									icon={ lock.edit ? lockIcon : unlockIcon }
-								/>
-							</li>
-						) }
-						<li className="block-editor-block-lock-modal__checklist-item">
+					{ /*
+					 * Disable reason: The `list` ARIA role is redundant but
+					 * Safari+VoiceOver won't announce the list otherwise.
+					 */
+					/* eslint-disable jsx-a11y/no-redundant-roles */ }
+					<ul
+						role="list"
+						className="block-editor-block-lock-modal__checklist"
+					>
+						<li>
 							<CheckboxControl
 								__nextHasNoMarginBottom
-								label={ __( 'Disable movement' ) }
-								checked={ lock.move }
-								onChange={ ( move ) =>
-									setLock( ( prevLock ) => ( {
-										...prevLock,
-										move,
-									} ) )
+								className="block-editor-block-lock-modal__options-all"
+								label={ __( 'Lock all' ) }
+								checked={ isAllChecked }
+								indeterminate={ isMixed }
+								onChange={ ( newValue ) =>
+									setLock( {
+										move: newValue,
+										remove: newValue,
+										...( allowsEditLocking
+											? { edit: newValue }
+											: {} ),
+									} )
 								}
 							/>
-							<Icon
-								className="block-editor-block-lock-modal__lock-icon"
-								icon={ lock.move ? lockIcon : unlockIcon }
-							/>
-						</li>
-						<li className="block-editor-block-lock-modal__checklist-item">
-							<CheckboxControl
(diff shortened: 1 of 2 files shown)


alexstine (commented):
Accessibility testing well but needs code review. GitHub's new and improved diff experience for screen reader users sucks worse than an Atlassian product 😞. Hopefully someone will be around to check this eventually.

t-hamano (approved):
I think the code is absolutely fine. NVDA's speech has changed to the following:

### When the Lock modal is opened

Before:

```
Lock Navigation
heading    level 1  Lock Navigation
button    Close
Choose specific attributes to restrict or lock all available options.

grouping    check box  not checked    Lock all
list  with 3 items  check box  not checked    Restrict editing
check box  not checked    Disable movement
check box  not checked    Prevent removal  out of list
check box  not checked    Apply to all blocks inside  out of grouping
button    Cancel
button    Apply
```

After:

```
Lock Navigation
heading    level 1  Lock Navigation
button    Close
grouping    Choose specific attributes to restrict or lock all available options.

list  with 1 item  check box  not checked    Lock all  list  with 3 items  check box  not checked    Restrict editing
check box  not checked    Disable movement
check box  not checked    Prevent removal  out of list  out of list
check box  not checked    Apply to all blocks inside  out of grouping
button    Cancel
button    Apply
```

### When the form elements are focused in order

Before:

```
Lock all  grouping  Lock all  check box  not checked  
list  with 3 items  Restrict editing  check box  not checked  
Disable movement  check box  not checked  
Prevent removal  check box  not checked  
Apply to all blocks inside  check box  not checked 
```

After:

```
Choose specific attributes to restrict or lock all available options.  grouping  list  with 1 item  Lock all  check box  not checked  
list  with 3 items  Restrict editing  check box  not checked  
Disable movement  check box  not checked  
Prevent removal  check box  not checked  
Apply to all blocks inside  check box  not checked  
```

afercia:
Thanks all for your feedback and review.

Ideally, the legend text should be a little more concise. The current one is too long:
`Choose specific attributes to restrict or lock all available options.`

It's also unclear what the difference between attributes and options is. Actually, I don't think it really matters for users. If anyone can think of a shorter, simpler, text to name the group I'd gladly update it.

t-hamano:
> Ideally, the legend text should be a little more concise. The current one is too long:

I agree with this, but I can't think of a good copy right now 😅 Perhaps we can address this in a follow-up.

afercia:
I will split the wording into a separate issue, as I think the options naming could be improved as well.

PR #62796, opened 2024-06-24
DataViews: remove selection correcting, set initial state

ellatrix:
## What?

Just trying this as an alternative to both #62787 and #62792.

If DataViews doesn't try to correct the selection when items don't exist or are removed (I think that should be the concern of the implementor), then we can also simply set an initial state.

## Why?

Reduces the amount of state setting in effects, which reduces the probability of infinite loops.

### DIFF
--- a/packages/edit-site/src/components/page-pages/index.js
+++ b/packages/edit-site/src/components/page-pages/index.js
@@ -201,6 +201,10 @@ function FeaturedImage( { item, viewType } ) {
 	);
 }
 
+function getItemId( item ) {
+	return item.id.toString();
+}
+
 export default function PagePages() {
 	const postType = 'page';
 	const [ view, setView ] = useView( postType );
@@ -464,9 +468,27 @@ export default function PagePages() {
 		[ authors, view.type, frontPageId, postsPageId ]
 	);
 
+	const onActionPerformed = useCallback(
+		( actionId, items ) => {
+			if (
+				actionId === 'move-to-trash' ||
+				actionId === 'permanently-delete'
+			) {
+				setSelection(
+					selection.filter(
+						( id ) =>
+							! items.some( ( item ) => getItemId( item ) === id )
+					)
+				);
+			}
+		},
+		[ selection, setSelection ]
+	);
+
 	const postTypeActions = usePostActions( {
 		postType: 'page',
 		context: 'list',
+		onActionPerformed,
 	} );
 	const editAction = useEditPostAction();
 	const actions = useMemo(
@@ -538,7 +560,7 @@ export default function PagePages() {
 				selection={ selection }
 				setSelection={ setSelection }
 				onSelectionChange={ onSelectionChange }
-				getItemId={ ( item ) => item.id.toString() }
+				getItemId={ getItemId }
 			/>
 		</Page>
 	);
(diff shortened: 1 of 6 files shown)


packages/edit-site/src/components/page-pages/index.js
```diff
@@ -464,9 +468,27 @@ export default function PagePages() {
...
+	const onActionPerformed = useCallback(
+		( actionId, items ) => {
+			if (
+				actionId === 'move-to-trash' ||
+				actionId === 'permanently-delete'
+			) {
+				setSelection(
+					selection.filter(
+						( id ) =>
+							! items.some( ( item ) => getItemId( item ) === id )
```
ellatrix: Why not set it to []?

jorgefilipecosta: I may have multiple items selected some could be deleted some could not. The delete may only apply to some items I should not clear the others. For example on pages I may select a filter on status for trash and drafts, I select a trashed item and draft I use the delete action. The items that was already in the trash should not be unselected nothing changed on that item.

jorgefilipecosta (approved):
Nice improvements, thank you for working on this! In my tests, I did not find any regression and this is a good simplification to the selection management.

We probably need a better long-term solution for selection management, when a destructive action happens, but I guess the current solution is ok for now.

packages/edit-site/src/components/page-pages/index.js
```diff
@@ -495,9 +468,27 @@ export default function PagePages() {
...
 	);
 
+	const onActionPerformed = useCallback(
+		( actionId, items ) => {
+			if (
+				actionId === 'move-to-trash' ||
+				actionId === 'permanently-delete'
+			) {
+				setSelection(
+					selection.filter(
```
ellatrix: I personally like the pattern of a callback more: `( selection ) => selection.filter`

jorgefilipecosta: Yes, it works better on this case and removes the selection dependency on the callback, I will update.

packages/edit-site/src/components/page-pages/index.js
```diff
@@ -495,9 +468,27 @@ export default function PagePages() {
...
+			) {
+				setSelection( ( _selection ) =>
+					_selection.filter(
+						( id ) =>
+							! items.some( ( item ) => getItemId( item ) === id )
+					)
+				);
+			}
+		},
+		[ setSelection ]
```
ellatrix: This is dependency is not needed

packages/edit-site/src/components/page-pages/index.js
```diff
@@ -274,6 +274,16 @@ export default function PagePages() {
 		totalPages,
 	} = useEntityRecords( 'postType', postType, queryArgs );
 
+	useEffect( () => {
+		if ( pages && ! pages.find( ( page ) => page.id === +postId ) ) {
```
jorgefilipecosta: I guess we should use getItemId here.

ellatrix:
> We probably need a better long-term solution for selection management, when a destructive action happens, but I guess the current solution is ok for now.

Fixed in 00737ac136318a0d4d0f72e37f14d4b6af03db79

PR #62798, opened 2024-06-24
Global Styles: Add block variations to the isactive style variation calculation

### LINKED ISSUE #62758: Global Styles: Active variation not shown when using block variations
If style variations contain block variations then the UI doesn't recognise which style variation is active, for example, using the Assembler theme, with the blueberry variation active, but not shown as active in the UI
Screenshot 2024-06-21 at 18 33 11

We need to update `areGlobalStyleConfigsEqual` so that it is aware of block variations and can correctly compare the current user settings with a style variation.

scruffian:
## What?
Fixes https://github.com/WordPress/gutenberg/issues/62758.

This processes style variation previews in the same way that we process variations when they are committed so that they are formatted the same way and can be detected as the same.

## Why?
This is necessary so that style variations are marked as active:
Screenshot 2024-06-24 at 15 46 20

## How?
Process the style variation with `useResolvedBlockStyleVariationsConfig` before comparing it to the saved user variation.

There are some other possibilities for how to solve this:
1. We could move the `useResolvedBlockStyleVariationsConfig` call to inside `areGlobalStyleConfigsEqual` so that other uses of this function benefit from the change.
2. We could move `useResolvedBlockStyleVariationsConfig` up the call stack so that the variation is processed before this point.

## Testing Instructions
1. Switch to a theme that contains a style variation which has block styles (I'm using Assembler)
2. Select a style variation
3. Check that the selected variation is highlighted as the active variation.

## Screenshots or screencast 
Screenshot 2024-06-24 at 15 46 20

### DIFF
--- a/packages/edit-site/src/components/global-styles/variations/variation.js
+++ b/packages/edit-site/src/components/global-styles/variations/variation.js
@@ -18,7 +18,8 @@ import { privateApis as editorPrivateApis } from '@wordpress/editor';
 import { filterObjectByProperty } from '../../../hooks/use-theme-style-variations/use-theme-style-variations-by-property';
 import { unlock } from '../../../lock-unlock';
 
-const { mergeBaseAndUserConfigs } = unlock( editorPrivateApis );
+const { mergeBaseAndUserConfigs, useResolvedBlockStyleVariationsConfig } =
+	unlock( editorPrivateApis );
 const { GlobalStylesContext, areGlobalStyleConfigsEqual } = unlock(
 	blockEditorPrivateApis
 );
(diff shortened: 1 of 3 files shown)


aaronrobertshaw (commented):
Thanks for the ping to review @scruffian 👍 

This PR highlighted some code that was missed in the recent efforts to switch shared variations from `styles.blocks.variations` to `styles.variations`. 

Those changes landed yesterday in #62712. The result is that the `useResolvedBlockStyleVariationsConfig` has no effect and isn't required anymore.  I've created a PR to remove that hook entirely in https://github.com/WordPress/gutenberg/pull/62808.

On `trunk` I can't replicate the original [issue](https://github.com/WordPress/gutenberg/issues/62758) so I think we might be able to close both the issue and this PR.

https://github.com/WordPress/gutenberg/assets/60436221/93a8ee9d-08a0-45e7-ae72-ffeb9248cd44

PR #55198, opened 2023-10-10
Focus submenu button when clicked

SantosGuillamot:
## What?
This pull request adds some JavaScript logic to ensure focus the submenu button when it is clicked. It solves this issue: https://github.com/WordPress/gutenberg/issues/54991

It also adds a fix for closing the submenu overlay on a click outside the submenu on iOS.

## Why?
It seems that Safari doesn't place the focus on the button element but on a parent `div`, which causes some issues with the directives assuming the focus is on the button. We can see the difference between Chrome and Safari:

**Chrome**

https://github.com/WordPress/gutenberg/assets/34552881/021bbbda-fb39-4b26-af8b-6dd4a504790b

**Safari**

https://github.com/WordPress/gutenberg/assets/34552881/2017fb96-f4d6-47f4-b433-568b473b3f7f

## How?
I added some JavaScript logic that, when the submenu button is clicked, it focuses it. It'd be great if someone from the accessibility team could double-check if this approach is correct or if it will break something I am unaware of. cc: @afercia, @alexstine , @jeryj , @joedolson  

## Testing Instructions
1. Add a navigation block with an item with submenu.
2. In the block settings, enable the "Show arrow" setting.
3. Open the site in Safari.
4. Click on the arrow to open the submenu.
5. Check that the `activeElement` is the arrow button.
6. Click outside of the submenu and check it is closed.
7. Open the submenu again by clicking the arrow.
8. Press the ESC key and check that the submenu is closed.

On iOS:
1. Click on the arrow to open the submenu.
2. Click on the arrow to close the submenu.
3. Click on the arrow to open the submenu.
4. Click on the body to close the submenu.

PR #62800, opened 2024-06-24
Navigation block: Fix submenu not opening on macOS Safari

luisherranz:
## What?

This pull request fixes an issue in Safari Desktop (macOS) where submenus do not open upon clicking due to a change in https://github.com/WordPress/gutenberg/pull/62160.

## Why?

Right now (Gutenberg 18.6), the submenus that have the "Open on Click" option do not open on Safari Desktop.

## How?

For now, I have simply reverted the change that causes the submenus to stop opening. I haven't investigated exactly why it happens yet.

It took us a while to figure out how to make the focus logic work properly in Safari since it functions completely differently than in Chrome and Firefox. That logic is also not as clean and straightforward as we would like:

- https://github.com/WordPress/gutenberg/pull/55198

I think for now we could leave this event handler as asynchronous to fix this bug and take a closer look at the Safari focus logic in a subsequent PR.

## Testing Instructions

- Add a menu with a submenu
- Activate the option "Open on click"
- Use Safari macOS
- Test that the submenu opens on click
- Test that the rest of features/accessibility work as usual

### DIFF
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -828,7 +828,7 @@ function block_core_navigation_add_directives_to_submenu( $tags, $block_attribut
 		$tags->set_attribute( 'data-wp-interactive', 'core/navigation' );
 		$tags->set_attribute( 'data-wp-context', '{ "submenuOpenedBy": { "click": false, "hover": false, "focus": false }, "type": "submenu" }' );
 		$tags->set_attribute( 'data-wp-watch', 'callbacks.initMenu' );
-		$tags->set_attribute( 'data-wp-on-async--focusout', 'actions.handleMenuFocusout' );
+		$tags->set_attribute( 'data-wp-on--focusout', 'actions.handleMenuFocusout' );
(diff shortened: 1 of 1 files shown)


luisherranz:
I think we need to add this fix to WP 6.6 and maybe even release a Gutenberg 18.6.1 version, don't we?

cc: @cbravobernal, @gziolo, @westonruter

westonruter (approved):
Aside: It would be great if the navigation block could start to leverage the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) since it is newly available in all browsers which would enable the menus to work even with JavaScript turned off entirely (cf. https://core.trac.wordpress.org/ticket/50182#comment:9)

ellatrix:
I just cherry-picked this PR to the wp/6.6-rc-1 branch to get it included in the next release: f5ee0ddcee

luisherranz:
@cbravobernal just released GB 18.6.1 with this fix:

- https://github.com/WordPress/gutenberg/releases/tag/v18.6.1

Issue #62462, opened 2024-06-11
Update JSON schemas to use draft-07


ajlende:
We've been following the [SchemaStore best practices](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#best-practices) for the development of our JSON schemas. They recommend a schema version based on what is widely supported in editors and IDEs. Previously this was draft-04, so all of our schemas currently use that version. However the new recommendation is draft-07 which comes with some quality of life improvements that can improve the readability and maintainability of our JSON schemas. Namely:

1. The [`$comment`](https://json-schema.org/draft-07/json-schema-core#rfc.section.9) keyword to explain reasoning behind schema decisions inline.
2. The [`propertyNames`](https://json-schema.org/draft-07/json-schema-validation#rfc.section.6.5.8) keyword to improve merging of two or more object subschemas while limiting `additionalProperties`.

Additionally, SchemaStore has an unofficial "strict mode" draft-07 meta schema `https://json.schemastore.org/metaschema-draft-07-unofficial-strict.json` that requires `type`, `title`, and `description` properties which we should be using and provides links to [understanding-json-schema](https://json-schema.org/understanding-json-schema/index.html) for each hint/check to help new JSON Schema contributors.

- There are some [breaking changes from draft-04 to draft-06](https://json-schema.org/draft-06/json-schema-release-notes#backwards-incompatible-changes). You can migrate schemas from draft-04 to draft-07 using [ajv-cli](https://github.com/ajv-validator/ajv-cli).
- The pattern of `*PropertiesComplete` can be updated to [use `propertyNames`](https://json-schema.org/draft-06/json-schema-release-notes#q:-what-happened-to-all-the-discussions-around-re-using-schemas-with) instead of lists of empty objects.
- The `enum` properties with single values can be updated to `const` where it would make sense to use `const`.

DivyanshVinayak23:
Hi, can I work on this?

ajlende:
> Hi, can I work on this?

@DivyanshVinayak23 Absolutely!

Check out the [contributing guidelines](https://github.com/WordPress/gutenberg/blob/trunk/CONTRIBUTING.md), specifically the [developer section](https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/README.md), for general information on how to contribute code.

The `#core-editor` channel on [Make WordPress Slack](https://make.wordpress.org/chat) (registration required) is also a good place to ask questions if you have any.

[...]

PR #62802, opened 2024-06-24
update schema version

scruffian:
## What?

Fix for #62462


lib/class-wp-rest-global-styles-controller-gutenberg.php
```diff
@@ -530,7 +530,7 @@ public function get_item_schema() {
 		}
 
 		$schema = array(
-			'$schema'    => 'http://json-schema.org/draft-04/schema#',
```
ajlende, line 533: I don't think we can update the controller schemas. The [code for validating REST API schemas](https://github.com/WordPress/wordpress-develop/blob/e8fed338f8f27363cb24f0ea2792cfa3b311b612/src/wp-includes/rest-api.php#L2071) only supports draft-04.

Let's just stick with the code in `schemas`.

PR #62803, opened 2024-06-24
CustomSelectControlV2: Add root element wrapper

ciampo:
## What?

Extracted from #62424

Add a root element wrapper to `CustomSelectControlV2`

## Why?

- To fix layout bugs where outer CSS rules "leak" to the internal elements of `CustomSelectControl` (eg. if the parent has `display: flex`, which is what happens when using `CustomSelectControl` as a direct child of `ToolsPanelItem`)
- For legacy reasons, in order to have a  place where to apply the `components-custom-select-control` className

## How?

Adding a root element wrapper (ie a `div`) instead of a React Fragment, wrapping label + button under one element.

### In the editor

To test in the editor, apply this patch

<details>

<summary>Click to expand</summary>

```diff
diff --git a/packages/block-editor/src/components/font-appearance-control/index.js b/packages/block-editor/src/components/font-appearance-control/index.js
index 18e814ad23..1340396c86 100644
--- a/packages/block-editor/src/components/font-appearance-control/index.js
+++ b/packages/block-editor/src/components/font-appearance-control/index.js
@@ -1,10 +1,20 @@
 /**
  * WordPress dependencies
  */
-import { CustomSelectControl } from '@wordpress/components';
+import {
+	CustomSelectControl,
+	privateApis as componentsPrivateApis,
+} from '@wordpress/components';
 import { useMemo } from '@wordpress/element';
 import { __, _x, sprintf } from '@wordpress/i18n';
 
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { CustomSelectControlV2Legacy } = unlock( componentsPrivateApis );
+
 const FONT_STYLES = [
 	{
 		name: _x( 'Regular', 'font style' ),
@@ -205,7 +215,7 @@ export default function FontAppearanceControl( props ) {
 
 	return (
 		hasStylesOrWeights && (
-			<CustomSelectControl
+			<CustomSelectControlV2Legacy
 				{ ...otherProps }
 				className="components-font-appearance-control"
 				label={ label }
diff --git a/packages/components/src/private-apis.ts b/packages/components/src/private-apis.ts
index f55373664e..9a26b7c7e3 100644
--- a/packages/components/src/private-apis.ts
+++ b/packages/components/src/private-apis.ts
@@ -14,6 +14,7 @@ import {
 	useCompositeStore as useCompositeStoreV2,
 } from './composite/v2';
 import { default as CustomSelectControl } from './custom-select-control';
+import { default as CustomSelectControlV2Legacy } from './custom-select-control-v2/legacy-component';
 import { positionToPlacement as __experimentalPopoverLegacyPositionToPlacement } from './popover/utils';
 import { createPrivateSlotFill } from './slot-fill';
 import {
@@ -40,6 +41,7 @@ lock( privateApis, {
 	CompositeRowV2,
 	useCompositeStoreV2,
 	CustomSelectControl,
+	CustomSelectControlV2Legacy,
 	__experimentalPopoverLegacyPositionToPlacement,
 	createPrivateSlotFill,
 	ComponentsContext,


```
</details>

2. Load the post editor
3. Select the block, and in the block toolbar, add the "Appearance" controls in the typography section
4. Notice how the label now has bottom margin

|---|---|
| Screenshot 2024-06-25 at 10 54 05 | Screenshot 2024-06-25 at 10 54 45 |

### DIFF
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -92,7 +92,7 @@ function _CustomSelect(
 	} = props;
 
 	return (
-		<>
+		<div>
 			{ hideLabelFromVision ? ( // TODO: Replace with BaseControl
 				<VisuallyHidden as="label">{ label }</VisuallyHidden>
 			) : (
@@ -116,7 +116,7 @@ function _CustomSelect(
 					</CustomSelectContext.Provider>
 				</Styled.SelectPopover>
 			</InputBase>
-		</>
+		</div>
 	);
 }
 
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -88,11 +88,13 @@ function _CustomSelect(
 		label,
 		size,
 		store,
+		className,
 		...restProps
 	} = props;
 
 	return (
-		<div>
+		// Where should `restProps` be forwarded to?
+		<div className={ className }>
 			{ hideLabelFromVision ? ( // TODO: Replace with BaseControl
 				<VisuallyHidden as="label">{ label }</VisuallyHidden>
 			) : (
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -3,6 +3,7 @@
  */
 // eslint-disable-next-line no-restricted-imports
 import * as Ariakit from '@ariakit/react';
+import clsx from 'clsx';
 
 /**
  * Internal dependencies
@@ -21,6 +22,7 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 		onChange,
 		size = 'default',
 		value,
+		className: classNameProp,
 		...restProps
 	} = props;
 
@@ -122,6 +124,10 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 			}
 			size={ translatedSize }
 			store={ store }
+			className={ clsx(
+				'components-custom-select-control',
+				classNameProp
+			) }
 			{ ...restProps }
 		>
 			{ children }
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Internal
+
+-   `CustomSelectControlV2`: add root element wrapper. ([#62803](https://github.com/WordPress/gutenberg/pull/62803))
+
 ## 28.2.0 (2024-06-26)
 
 ### Enhancements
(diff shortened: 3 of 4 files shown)


packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -88,11 +88,13 @@ function _CustomSelect(
 		label,
 		size,
 		store,
+		className,
 		...restProps
 	} = props;
 
 	return (
-		<>
+		// Where should `restProps` be forwarded to?
```
ciampo, line 96: This is something unrelated to this PR, but that we should definitely discuss.

Currently, `...restProps` are assumed to be intended for `CustomSelectButton`, but I'd argue that folks would expect that most attributes would be assigned to the root wrapper (which will likely become `BaseControl`)

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -21,6 +22,7 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 		onChange,
 		size = 'default',
 		value,
+		className: classNameProp,
```
mirka, line 25: Why the rename?

packages/components/src/custom-select-control-v2/custom-select.tsx
mirka: I think it should be the `CustomSelectButton`. Is there anything other than `className` that you'd want to pass to the root wrapper? I'm thinking about refs and event listeners, which would pass more naturally onto the button.

mirka (approved):
Looks good 🚀

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: Because I would otherwise get a `'className' is already declared in the upper scope` linting error later in the file where `options` is mapped into `<CustomSelectItem />`s

packages/components/src/custom-select-control-v2/custom-select.tsx
ciampo: I'm worried that if we start adding exceptions like `className`, devEx will suffer because it won't be clear where each prop will be forwarded to. Also, if a consumer of the component wanted to set a `className` on the trigger button, how would they do it?


Another though: if we don't pass _any_ props to the root wrapper element, do you see that as a limitation for consumers? For example,:
- let's say that as a consumer I want to hide `CustomSelectControl` — setting `display: none` would only hide the button, but keep the label around
- let's say I'm rendering `CustomSelectControlV2` inside a flexbox parent, and I'd like to set `flex-shring` on it. Using `className` or `style` won't cut it, since those styles should be applied to the root wrapper. The only way to achieve that will be to add an external wrapper component to `CustomSelectControlV2`.

I think that this is a broader conversation that we need to have as we export components that internally compose smaller components.

For example, in `DropdownMenuV2` we solved this by exporting a separate `Trigger` subcomponent, and exposing a `trigger` render prop. Which means that props by default can be passed to the root element, while trigger-specific props are passed directly to the `Trigger` subcomponent.

Curious to hear @DaniGuardiola 's thoughts on this one too.

ciampo:
I'm going to merge once CI cheks pass — although we should continue the conversation in https://github.com/WordPress/gutenberg/pull/62803#discussion_r1652321562 and work on it via a follow-up as necessary.

PR #62804, opened 2024-06-24
Fix: extra `wp-block-navigation` getting on `ul` tag

### LINKED ISSUE #62690: Navigation Block: block wrapper class name applied to both outer element and its child element
### Description

If you inspect the markup of a navigation block with the overlay turned off (makes it easier to reproduce), the `wp-block-navigation` class name is applied twice, once to the outer `<nav>` and once to its child `<ul>`:
```html
<nav class="wp-block-navigation is-layout-flex wp-block-navigation-is-layout-flex" aria-label="Navigation 2">
<ul class="wp-block-navigation__container  wp-block-navigation">
```

Blocks should only have the wrapper classname on the outermost element.

The bug seems to be caused by `get_block_wrapper_attributes` being used to compute the attributes of the `<ul>` element:
https://github.com/WordPress/gutenberg/blob/49070907e6d0bdc19e47628e12b0754cd7ddcb6b/packages/block-library/src/navigation/index.php#L177-L182

A concern is that changing this might cause some issues, it'd require a dev note as it may break some styles, but I think it needs to be fixed to ensure themes don't end up with unsual style rules like `ul.wp-block-navigation`.

1. Add a naviation block
2. Turn off the overlay
3. Preview the frontend
4. Inspect the markup

_No response_

### Environment info

Reproduced on latest trunk core/gutenberg

Yes

Yes

up1512001:
## What?
- created function similar to `get_block_wrapper_attributes` which support extra parameter to exclude some class.

## Why?
Fixes https://github.com/WordPress/gutenberg/issues/62690

## How?
- as `get_block_wrapper_attributes` enters default block class, which causes issues by having the block class on the inner container if it's used.

## Testing Instructions
- create navigation menu
- check `ul` tag 
- now there will not be `wp-block-navigation` class which should only be present on main navigation wrapper

## Screenshots or screencast 
Screenshot 2024-06-25 at 00 13 53

### DIFF
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -162,7 +162,7 @@ class WP_Navigation_Block_Renderer {
 	/**
 	 * Returns the inner container attributes.
 	 *
-	 * @since 6.6.0
+	 * @since 6.7.0
 	 *
 	 * @param array $extra_attributes The extra attributes to add.
 	 * @param string $exclude_class The class to exclude.
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -160,6 +160,67 @@ class WP_Navigation_Block_Renderer {
 	}
 
 	/**
+	 * Returns the inner container attributes.
+	 *
+	 * @since 6.6.0
+	 *
+	 * @param array $extra_attributes The extra attributes to add.
+	 * @param string $exclude_class The class to exclude.
+	 * @return string Returns the inner container attributes.
+	 */
+	private static function get_inner_container_attributes( $extra_attributes = array(), $exclude_class = '' ) {
+		$new_attributes = WP_Block_Supports::get_instance()->apply_block_supports();
+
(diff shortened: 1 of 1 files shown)


up1512001:
@talldan can you please review this PR?

talldan:
@up1512001 I'm not really sure I know the PHP code for the navigation block well enough or how the block is supposed to function in terms of that wrapper. I've added a few other reviewers who may be able to help.

My instinct is that such a close copy of `get_block_wrapper_attributes` wouldn't be needed, but perhaps a simplified one that handles merging the classes and styles and outputting them in a similar format.

scruffian:
I wonder if a simple regex is better? https://github.com/WordPress/gutenberg/pull/63801

draganescu:
Howdy and thanks for the work here 🙇🏻 

While ingenious the duplication of the `get_block_wrapper_attributes` is impossible to maintain, hence a no go. We either need to augment that function in core with some param to selectively get output or to update the navigation block itself to handle the problem.

I think this approach can be closed. It's unclear if https://github.com/WordPress/gutenberg/pull/63801 is a better approach.

PR #62806, opened 2024-06-24
Fix: Don't send meta on the duplicate action just because of empty foonotes.

jorgefilipecosta:
Even if a post has no footnotes its meta-object is {footnotes: ''}. So on duplicate if the post has a meta object equal to {footnotes: ''} there is no need to pass the meta object on the API request at all. This PR does that.

This change addresses an issue/bug where if the user has no edit_post capability (but still has the create_post capability so can duplicate a post) we throw an error message on duplicate saying the user is unable to change the footnotes meta property, even if the page has no footnotes.

## Testing Instructions
I pasted the following PHP code to remove the edit_post capability:
```
function remove_edit_delete_post_capability( $caps, $cap, $user_id, $args ) {
    if ( 'edit_post' === $cap || 'delete_post' === $cap) {
            $caps[] = 'do_not_allow';
        }
    }

    return $caps;
}
add_filter( 'map_meta_cap4', 'remove_edit_delete_post_capability', 10, 4 );
```

I went to the pages list `wp-admin/site-editor.php?postType=page` and duplicated a page. I verified there were no errors (on trunk there is an error related to footnotes).

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -765,7 +765,6 @@ const useDuplicatePostAction = ( postType ) => {
 									? item.content
 									: item.content.raw,
 							excerpt: item.excerpt.raw,
-							meta: item.meta,
 							parent: item.parent,
 							password: item.password,
 							template: item.template,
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -765,7 +765,6 @@ const useDuplicatePostAction = ( postType ) => {
 									? item.content
 									: item.content.raw,
 							excerpt: item.excerpt.raw,
-							meta: item.meta,
```
ellatrix: Why is the footnotes meta of the post to be duplicated empty?

jorgefilipecosta:
Rebased onto latest `trunk`. The original change touched `packages/editor/src/components/post-actions/actions.js`, but the `useDuplicatePostAction` hook has since been extracted to `packages/fields/src/actions/duplicate-post.tsx` and converted to TypeScript. The same conditional-meta fix has been ported to the new location: `meta` is removed from the initial `newItemObject` literal, then assigned only when `item.meta` is non-empty and contains either a real `footnotes` value or other (non-footnotes) keys. The `newItemObject` was retyped as `Record<string, unknown>` so the conditional assignment compiles cleanly alongside the existing dynamic `assignableProperties` writes.

PR #62815, opened 2024-06-25
WIP: Block bindings - add an API to determine whether a binding source is active and use it for pattern overrides

talldan:
## What?
Proposes adding an `isActive` API for binding sources. At the moment this is added as a private API while it can be assessed.

## Why?
Most of the time the pattern overrides block bindings don't need to be processed, the block bindings only become active when there's a parent pattern block. 

When editing the original pattern, there are bindings present in the block markup that are processed but shouldn't need to be. As an example, the `getValue` callback requires an early return:
https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/editor/src/bindings/pattern-overrides.js#L17-L19

And the `setValues` callback requires some code to allow editing source patterns (setting attributes for the blocks in the patterns:
https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/editor/src/bindings/pattern-overrides.js#L48-L66

Block bindings already has code to do this where it falls back to calling `setAttributes`, so this PR removes that code:
https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/block-editor/src/hooks/use-bindings-attributes.js#L234

Both of the above bits of code require some work by block bindings to get to that point, all of which isn't needed if the code can return earlier.

## How?
Allows a new `isActive` function to be added to binding sources. In this PR it's set to private.

For pattern overrides, this function returns `true` whenever the `pattern/overrides` block context is present. When it returns `false` it indicates the block isn't within a pattern. In the block binding code, the source is filtered out, and it allows pattern override bindings to be skipped.

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -1,7 +1,10 @@
 /**
  * WordPress dependencies
  */
-import { store as blocksStore } from '@wordpress/blocks';
+import {
+	store as blocksStore,
+	privateApis as blocksPrivateApis,
+} from '@wordpress/blocks';
 import { createHigherOrderComponent } from '@wordpress/compose';
 import { useRegistry, useSelect } from '@wordpress/data';
 import { useCallback, useMemo } from '@wordpress/element';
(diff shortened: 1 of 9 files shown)


gziolo:
This PR is fairly outdated at the moment. @talldan, do you think it still would be valuable to bring it to the finish line? I'm not sure I fully understand the intention, but once I learn more, I'll be happy to help with the review and testing.

talldan:
> This PR is fairly outdated at the moment. @talldan, do you think it still would be valuable to bring it to the finish line? I'm not sure I fully understand the intention, but once I learn more, I'll be happy to help with the review and testing.

I forget now, I think I was trying to find a way to disable the pattern overrides bindings when editing the original source pattern. In this case the binding shouldn't be active, we want the block to display its own attribute value.

From a user perspective this already works, but it's achieved by processing binding and returning the block's own data, so we do a lot of unnecessary work.

My idea is to disable the binding when in this context. Maybe unregistering the binding could also be an option.

I probably won't explore it further so will close the PR.

PR #62817, opened 2024-06-25
Docs/iAPI: Recommend kebab-case in data-wp-class

luisherranz:
## What?

Add a notice to recommend users to use kebab-case for the class names when using `data-wp-class`.

I've also removed a couple of sentences that referred to the arguments we were passing to the callbacks of `wp-class` and `wp-style`, which were eliminated before WordPress 6.5.

## Why?

Because HTML is not case-sensitive in attribute names and using camelCase can cause problems and confusion.

## How?

I've simply added an explanation and an example in the `wp-class` section.


cbravobernal (approved):
LGTM

PR #62825, opened 2024-06-25
CustomSelectControlV2: fix item styles

ciampo:
## What?

Part of #55023

Fix items (options) styles of `CustomSelectControlV2` component

## Why?

- To fix visual bugs
- To match more closely look of V1 component
- To add quality of life improvements

## How?

- Update V2 items in order to match the same line height of the V1 version
- Add `disabled` styles
- Added `user-select: none` to trigger button and option items
- Added `scroll-margin` to improve scroll experience when selecting items that are out of view in the popover

## Testing Instructions

- Compare V1 and V2 (legacy) of `CustomSelectControl`, make sure that the items in the select popover now render with the same line height (28px).
- Try to select with the cursor the text of the items in the select popover — V2 should prevent text selection

| V1 | V2 before (trunk) | V2 after (this PR) |
|---|---|---|
| Screenshot 2024-06-25 at 13 13 47 | Screenshot 2024-06-25 at 13 14 27 | Screenshot 2024-06-25 at 13 23 21 |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Internal
 
 -   `CustomSelectControlV2`: prevent keyboard event propagation in legacy wrapper. ([#62907](https://github.com/WordPress/gutenberg/pull/62907))
+-   `CustomSelectControlV2`: fix item styles ([#62825](https://github.com/WordPress/gutenberg/pull/62825))
 -   `CustomSelectControlV2`: add root element wrapper. ([#62803](https://github.com/WordPress/gutenberg/pull/62803))
 -   `CustomSelectControlV2`: fix popover styles. ([#62821](https://github.com/WordPress/gutenberg/pull/62821))
 -   `CustomSelectControlV2`: fix trigger text alignment in RTL languages ([#62869](https://github.com/WordPress/gutenberg/pull/62869)).
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -133,7 +133,9 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	justify-content: space-between;
 	padding: ${ ITEM_PADDING };
 	font-size: ${ CONFIG.fontSize };
-	line-height: 2.15rem; // TODO: Remove this in default but keep for back-compat in legacy
+	// Legacy line height is 28px.
+	// TODO: reassess for non-legacy v2
+	line-height: calc( 28px / ${ CONFIG.fontSize } );
 	&[data-active-item] {
 		background-color: ${ COLORS.theme.gray[ 300 ] };
 	}
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -137,6 +137,8 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	// TODO: reassess for non-legacy v2
 	line-height: calc( 28px / ${ CONFIG.fontSize } );
 
+	user-select: none;
+
 	&[aria-disabled='true'] {
 		opacity: 0.5;
 	}
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -136,6 +136,11 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	// Legacy line height is 28px.
 	// TODO: reassess for non-legacy v2
 	line-height: calc( 28px / ${ CONFIG.fontSize } );
+
+	&[aria-disabled='true'] {
+		opacity: 0.5;
+	}
+
 	&[data-active-item] {
 		background-color: ${ COLORS.theme.gray[ 300 ] };
 	}
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -93,6 +93,7 @@ export const Select = styled( Ariakit.Select, {
 		font-family: inherit;
 		font-size: ${ CONFIG.fontSize };
 		text-align: start;
+		user-select: none;
 		width: 100%;
 
 		&[data-focus-visible] {
@@ -136,7 +137,7 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	// Legacy line height is 28px.
 	// TODO: reassess for non-legacy v2
 	line-height: calc( 28px / ${ CONFIG.fontSize } );
-
+	scroll-margin: ${ space( 1 ) };
 	user-select: none;
 
 	&[aria-disabled='true'] {
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -140,7 +140,7 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	user-select: none;
 
 	&[aria-disabled='true'] {
-		opacity: 0.5;
+		cursor: not-allowed;
 	}
 
 	&[data-active-item] {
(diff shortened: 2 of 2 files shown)


packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -108,7 +108,14 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	justify-content: space-between;
 	padding: ${ ITEM_PADDING };
 	font-size: ${ CONFIG.fontSize };
-	line-height: 2.15rem; // TODO: Remove this in default but keep for back-compat in legacy
```
ciampo, line 136: To match 28px, the line height value should have been `2.15` (without the `rem` unit). I felt like a more explicit `calc` syntax would help to understand the reason for an arbitrary number.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -133,7 +134,16 @@ export const SelectItem = styled( Ariakit.SelectItem )`
...
+	// Legacy line height is 28px.
+	// TODO: reassess for non-legacy v2
+	line-height: calc( 28px / ${ CONFIG.fontSize } );
```
mirka: Apparently this `calc` [doesn't work](https://developer.mozilla.org/en-US/docs/Web/CSS/calc#rules_and_best_practices_while_using_calc) 😕 ("For `/`, the right operand must be unitless".)

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -133,7 +134,16 @@ export const SelectItem = styled( Ariakit.SelectItem )`
...
+	scroll-margin: ${ space( 1 ) };
+	user-select: none;
+
+	&[aria-disabled='true'] {
+		opacity: 0.5;
+	}
```
mirka: I actually omitted the disabled styles on purpose, because the item styles are fully customizable and the appropriate disabled style will vary. For example, adding a `disabled` to the "Purple" item in the [Default story](https://wordpress.github.io/gutenberg/?path=/story/components-customselectcontrol-v2-default--default) doesn't result in an understandable visual.

https://github.com/WordPress/gutenberg/blob/9a331f13e8e1c59b9defb27a0bae38265de38578/packages/components/src/custom-select-control-v2/types.ts#L196-L202

What do you think?

packages/components/CHANGELOG.md
```diff
@@ -22,6 +23,7 @@
 -   `CustomSelectControlV2`: fix handling of extra option attributes in the `onChange` callbacks and when forwarding them to the option DOM elements. ([#62255](https://github.com/WordPress/gutenberg/pull/62255))
 -   `CustomSelectControlV2`: fix setting initial value and reacting to external controlled updates. ([#62733](https://github.com/WordPress/gutenberg/pull/62733))
 -   `CustomSelectControlV2`: Handle long strings in selected value ([#62198](https://github.com/WordPress/gutenberg/pull/62198)).
+-   `CustomSelectControlV2`: fix item styles ([#62825](https://github.com/WordPress/gutenberg/pull/62825))
```
tyxla: We should remove this in favor of the top entry.

packages/components/src/custom-select-control-v2/styles.ts
tyxla: It's curious where the 2.15 came from originally...

packages/components/src/custom-select-control-v2/styles.ts
tyxla: Maybe if we wrap `CONFIG.fontSize()` in a `stringToNumber()`? I appreciate the clarity that the explicit calc adds.

packages/components/src/custom-select-control-v2/styles.ts
ciampo: It's the ratio of the line height (28px) to the font size (13px)

Also, since `28 / 13 = 2.15384615`, another consequence of using `2.15` is that the resulting line height wouldn't be exactly 28px (as per design), but slightly less because of the lower decimal precision.

In short: I prefer to let the browser (or sass) do the calculation, and if not possible, to state the `line-height` directly.

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Sometimes less is more — we can also just remove the `${ CONFIG.fontSize }` part, and simply state `28px` — pushed in https://github.com/WordPress/gutenberg/pull/62825/commits/9a5d26fa9b60ff173e0ec61f509857027ca565da

packages/components/src/custom-select-control-v2/styles.ts
ciampo: You raise a good point — although I also think that the component should offer some good defaults.

Testing a middle ground in https://github.com/WordPress/gutenberg/pull/62825/commits/37ee79c6bb721b8fe7f3dade34603b56de7ce459 , using `cursor: not-allowed` as a more un-opinionated way to apply default disabled item styles.

Let me know if you still think this is too much

packages/components/CHANGELOG.md
ciampo: Woops! Rebase artifact. Thank you for flagging

packages/components/src/custom-select-control-v2/styles.ts
tyxla: > Also, since 28 / 13 = 2.15384615, another consequence of using 2.15 is that the resulting line height wouldn't be exactly 28px (as per design), but slightly less because of the lower decimal precision.

Which also makes me question if 28/13 was the intent. But yes, having a clear calc there makes things much clearer.

packages/components/src/custom-select-control-v2/styles.ts
tyxla: But these 2 aren't the same, are they:

```
font-size: 13px;
line-height: 2.15384615
```

and:

```
font-size: 13px
line-height: 28px
```

because if someone overrides only the `font-size`, the `line-height` will behave differently - considering the new font-size in the first example, and ignoring it in the second one.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -133,7 +134,15 @@ export const SelectItem = styled( Ariakit.SelectItem )`
 	justify-content: space-between;
 	padding: ${ ITEM_PADDING };
 	font-size: ${ CONFIG.fontSize };
-	line-height: 2.15rem; // TODO: Remove this in default but keep for back-compat in legacy
+	// TODO: reassess line-height for non-legacy v2
+	line-height: 28px;
```
tyxla, line 138: I'd prefer `2.15384615` and adding a comment about how it's calculated, because that way we'll keep the flexibility/relativity in case someone modifies only the `font-size`.

tyxla (approved):
Left a minor suggestion, but other than that, this looks good to me. Thanks! :shipit:

packages/components/src/custom-select-control-v2/styles.ts
ciampo: To adhere more closely to legacy behavior, It's actually better to keep the hardcoded `28px` as V1 does, exactly for the reason you state. For example, in the default Storybook example of `CustomSelectControl` each select item has a custom `font-size`, but the line-height is supposed to stay the same instead of scaling up/down.

packages/components/src/custom-select-control-v2/styles.ts
ciampo: See https://github.com/WordPress/gutenberg/pull/62825#discussion_r1661279512

packages/components/src/custom-select-control-v2/styles.ts
tyxla: > the line-height is supposed to stay the same instead of scaling up/down.

Interesting - is this always the case? What if I have font-size 56px, won't it be ugly with line-height of 28px?

ciampo: I guess we can review that when assessing what `line-height` to adopt for the V2 — keeping it as hardcoded `28px` is probably the safest choice while swapping the V1 implementation

PR #62826, opened 2024-06-25
Cherry pick workflow: improve message after conflict

ellatrix:
## What?

When there's a cherry-pick conflict and the job aborts, let's leave a more helpful comment. See #62716.

### DIFF
--- a/.github/workflows/cherry-pick-wp-release.yml
+++ b/.github/workflows/cherry-pick-wp-release.yml
@@ -142,7 +142,11 @@ jobs:
                         git checkout -b my-branch
                         # Cherry-pick the commit.
                         git cherry-pick ${commitSha}
+                        # Check which files have conflicts.
+                        git status
                         # Resolve the conflict...
+                        # Add the resolved files to the staging area.
+                        git status
                         git add .
                         git cherry-pick --continue
                         # Push the branch to the repository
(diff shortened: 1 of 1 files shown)


.github/workflows/cherry-pick-wp-release.yml
```diff
@@ -122,12 +123,32 @@ jobs:
...
+
+                        \`\`\`
+                        # Checkout the ${targetBranch} branch instead of trunk.
+                        git checkout ${targetBranch}
+                        # Create a new branch for your PR.
+                        git checkout -b my-branch
+                        # Cherry-pick the commit.
+                        git cherry-pick ${commitSha}
+                        # Resolve the conflict...
+                        git add .
```
desrosj, line 150: Should we mention the contributor should make sure they don't have any unintended file changes? `git add .` will stage all files, so any that are not explicitly `.gitignore`d will be added.

desrosj (approved):
Just had a clarifying question, but the changes make sense to me!

.github/workflows/cherry-pick-wp-release.yml
ellatrix: Sure, can add a line suggesting to check that with `git status` 👍

PR #62827, opened 2024-06-25
Global Styles: Simplify code to fetch color and typography variation

scruffian:
## What?
Small refactor to simplify this code and share more code between the two variations.

## Why?
Less code means less maintenance and fewer bugs.

## How?
Remove `useColorVariations` and `useTypographyVariations` and call `useCurrentMergeThemeStyleVariationsWithUserConfig` directly instead.

## Testing Instructions
0. Switch to a theme that uses color and typography variations (I'm using a modified TT4 with this variation: https://gist.github.com/scruffian/3dab906ee04d3c1c689bb08d23198f84)
1. Open the site editor
2. Open Global Styles > Typography
3. Check you can still see the list of Typography variations
4. Open Global Styles > Colors > Palette
5. Check you can still see the list of Color variations

## Screenshots or screencast 

Screenshot 2024-06-25 at 13 11 47
Screenshot 2024-06-25 at 13 12 04

### DIFF
--- a/packages/edit-site/src/components/global-styles/variations/variations-color.js
+++ b/packages/edit-site/src/components/global-styles/variations/variations-color.js
@@ -10,12 +10,14 @@ import {
  * Internal dependencies
  */
 import StylesPreviewColors from '../preview-colors';
-import { useColorVariations } from '../hooks';
+import { useCurrentMergeThemeStyleVariationsWithUserConfig } from '../../../hooks/use-theme-style-variations/use-theme-style-variations-by-property';
 import Subtitle from '../subtitle';
 import Variation from './variation';
 
 export default function ColorVariations( { title, gap = 2 } ) {
-	const colorVariations = useColorVariations();
+	const propertiesToFilter = [ 'color' ];
+	const colorVariations =
+		useCurrentMergeThemeStyleVariationsWithUserConfig( propertiesToFilter );
 
 	// Return null if there is only one variation (the default).
 	if ( colorVariations?.length <= 1 ) {
@@ -31,7 +33,7 @@ export default function ColorVariations( { title, gap = 2 } ) {
 						key={ index }
 						variation={ variation }
 						isPill
-						properties={ [ 'color' ] }
+						properties={ propertiesToFilter }
 						showTooltip
 					>
 						{ () => <StylesPreviewColors /> }
--- a/packages/edit-site/src/components/global-styles/variations/variations-typography.js
+++ b/packages/edit-site/src/components/global-styles/variations/variations-typography.js
@@ -9,13 +9,17 @@ import {
 /**
  * Internal dependencies
  */
+
 import StylesPreviewTypography from '../preview-typography';
-import { useTypographyVariations } from '../hooks';
+import { useCurrentMergeThemeStyleVariationsWithUserConfig } from '../../../hooks/use-theme-style-variations/use-theme-style-variations-by-property';
 import Variation from './variation';
 import Subtitle from '../subtitle';
 
 export default function TypographyVariations( { title, gap = 2 } ) {
-	const typographyVariations = useTypographyVariations();
+	const propertiesToFilter = 'typography';
+	const typographyVariations =
+		useCurrentMergeThemeStyleVariationsWithUserConfig( propertiesToFilter );
+
 	// Return null if there is only one variation (the default).
 	if ( typographyVariations?.length <= 1 ) {
 		return null;
@@ -34,7 +38,7 @@ export default function TypographyVariations( { title, gap = 2 } ) {
 						<Variation
 							key={ index }
 							variation={ variation }
-							properties={ [ 'typography' ] }
+							properties={ propertiesToFilter }
 							showTooltip
 						>
 							{ () => (
--- a/packages/edit-site/src/components/global-styles/variations/variations-typography.js
+++ b/packages/edit-site/src/components/global-styles/variations/variations-typography.js
@@ -16,7 +16,7 @@ import Variation from './variation';
 import Subtitle from '../subtitle';
 
 export default function TypographyVariations( { title, gap = 2 } ) {
-	const propertiesToFilter = 'typography';
+	const propertiesToFilter = [ 'typography' ];
 	const typographyVariations =
 		useCurrentMergeThemeStyleVariationsWithUserConfig( propertiesToFilter );
 
--- a/packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
+++ b/packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
@@ -50,6 +50,23 @@ export function removePropertiesFromObject( object, properties ) {
 }
 
 /**
+ * Checks whether a style variation is empty.
+ *
+ * @param {Object} variation          A style variation object.
+ * @param {string} variation.title    The title of the variation.
+ * @param {Object} variation.settings The settings of the variation.
+ * @param {Object} variation.styles   The styles of the variation.
+ * @return {boolean} Whether the variation is empty.
+ */
+function isEmptyStyleVariation( { title, settings, styles } ) {
+	return (
+		title === __( 'Default' ) || // Always preseve the default variation.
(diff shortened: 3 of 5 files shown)


packages/edit-site/src/components/global-styles/variations/variations-typography.js
```diff
@@ -10,14 +10,16 @@ import {
...
-import { useTypographyVariations } from '../hooks';
+import { useCurrentMergeThemeStyleVariationsWithUserConfig } from '../../../hooks/use-theme-style-variations/use-theme-style-variations-by-property';
 import TypographyExample from '../typography-example';
 import PreviewIframe from '../preview-iframe';
 import Variation from './variation';
 import Subtitle from '../subtitle';
 
 export default function TypographyVariations( { title, gap = 2 } ) {
-	const typographyVariations = useTypographyVariations();
+	const propertyToFilter = 'color';
```
scruffian:
```suggestion
	const propertyToFilter = 'typography';
```

packages/edit-site/src/components/global-styles/variations/variations-color.js
```diff
@@ -31,7 +33,7 @@ export default function ColorVariations( { title, gap = 2 } ) {
 						key={ index }
 						variation={ variation }
 						isPill
-						property="color"
+						property={ propertyToFilter }
```
jeryj: Will these changes work [alongside the changes in this PR](https://github.com/WordPress/gutenberg/pull/62847/files#diff-db2b09d114135d6ef1148111439fedad393f380c16be6161a836ec235b3f8caeR34)?

packages/edit-site/src/components/global-styles/variations/variations-typography.js
```diff
@@ -38,7 +40,7 @@ export default function TypographyVariations( { title, gap = 2 } ) {
 						<Variation
 							key={ index }
 							variation={ variation }
-							property="typography"
+							property={ propertyToFilter }
```
jeryj: Also here.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -96,6 +110,15 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
 	}, [ property, userVariation, variationsFromTheme ] );
+
+	/*
+	 * Filter out variations with no settings or styles.
+	 */
+	return variationsByProperty?.length
+		? variationsByProperty.filter( ( variation ) =>
+				isEmptyStyleVariation( variation )
+		  )
+		: [];
```
jeryj: Why not put this within the `useMemo`?

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -47,17 +47,31 @@ export function removePropertyFromObject( object, property ) {
...
+ *
+ * @param {Object} variation          A style variation object.
+ * @param {string} variation.title    The title of the variation.
+ * @param {Object} variation.settings The settings of the variation.
+ * @param {Object} variation.styles   The styles of the variation.
+ * @return {boolean} Whether the variation is empty.
+ */
+function isEmptyStyleVariation( { title, settings, styles } ) {
+	return (
+		title === __( 'Default' ) || // Always preseve the default variation.
```
jeryj:
```suggestion
		title === __( 'Default' ) || // Always preserve the default variation.
```

jeryj: Why is default always an empty style variation?

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -47,17 +47,31 @@ export function removePropertyFromObject( object, property ) {
...
+		Object.keys( settings ).length > 0 ||
+		Object.keys( styles ).length > 0
```
jeryj, line 65: I'm not familiar with how `theme.json` settings get applied. From looking at the  structure, I'm wondering if this should be an AND:

`( Object.keys( settings ).length > 0 && Object.keys( styles ).length > 0 )`

It looks like it could be a valid theme style while having a setting or style? Or does it need both?

jeryj (commented):
Asked a lot of questions because I'm not very familiar with this code :) It's testing fine for me though!

packages/edit-site/src/components/global-styles/variations/variations-color.js
scruffian: Yes, if that one is merged we'll get a conflict here. I think there's more we'll be able to simplify once the two PRs are combined.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
scruffian: Default removes all changes to the user global styles so that it reverts back to whatever theme.json is setting.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
scruffian: Ah the problem here is actually the name of the function. It is returning true if the variation is NOT empty!

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
scruffian: Good idea, done in a68c2cebde.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -47,17 +47,31 @@ export function removePropertyFromObject( object, property ) {
...
+ * @param {Object} variation.styles   The styles of the variation.
+ * @return {boolean} Whether the variation is empty.
+ */
+function isEmptyStyleVariation( { title, settings, styles } ) {
+	return (
+		title !== __( 'Default' ) && // Always preserve the default variation.
+		Object.keys( settings ).length === 0 &&
+		Object.keys( styles ).length === 0
+	);
+}
```
jeryj: Do you want to rename instead of changing the function then? I think it would be simpler as: `hasThemeVariation`

Then you can leave the previous `||` lines as before so we're not making a functional change here too.

```suggestion
function hasThemeVariation( { title, settings, styles } ) {
	return (
		title !== __( 'Default' ) || // Always preserve the default variation.
		Object.keys( settings ).length > 0 ||
		Object.keys( styles ).length > 0
	);
}
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -91,10 +105,19 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
 			...variationsWithSinglePropertyAndBase,
 		];
+
+		/*
+		 * Filter out variations with no settings or styles.
+		 */
+		return variationsByProperty?.length
+			? variationsByProperty.filter(
+					( variation ) => ! isEmptyStyleVariation( variation )
+			  )
```
jeryj:
```suggestion
			? variationsByProperty.filter( hasThemeVariation )
```

ramonjd:
I think `<SidebarNavigationItemGlobalStyles />` needs to be refactored too? 

For me the browser crashes when visiting Site Editor > Styles


https://github.com/WordPress/gutenberg/blob/refactor/type-and-color-variations/packages/edit-site/src/components/sidebar-navigation-screen-global-styles/index.js#L29-L32

scruffian:
Thanks @ramonjd. I have fixed this and also rebased it, so it should be working again...

ramonjd (approved):
LGTM

🚢

PR #62828, opened 2024-06-25
Pattern overrides: Fix aspect ratio not working in image with overrides

### LINKED ISSUE #62814: Pattern Overrides: Unable to set some attributes like Image block Aspect Ratio when editing the original pattern
### Description

When editing the original pattern and the block has overrides active, some attributes like 'aspect ratio' don't work.

This is a side effect of https://github.com/WordPress/gutenberg/pull/62471, which disabled regular attribute updates for any block with a `__default` pattern override binding. The intention of that PR was to only affect pattern instances to stop situations where edits could be applied to 'locked' parts of blocks. However, it also affects the block when editing the original pattern.

I think it may be worth reverting that PR, even though reverting will cause two new issues:
- The original issue that #62471 tried to solve, captions could incorrectly be set via the media library for overridden blocks in pattern instances, but the value doesn't persist when saving
- For editing the original pattern, #62747 tried to solve an issue where users can display a caption for an image block with overrides, even though captions aren't supported. #62471 disallows setting the caption, but also disallows setting any other unbound attributes too.

1. Create a pattern
2. Add an image block
3. Enable bindings for the image block
4. Upload an image to the block
5. Try adjusting Aspect Ratio, observe it doesn't work.

_No response_

### Environment info

_No response_

Yes

Yes

SantosGuillamot:
## What?
Fixes https://github.com/WordPress/gutenberg/issues/62814.

In order to do so, it includes the changes from [this pull request](https://github.com/WordPress/gutenberg/pull/62861),  where block context is used to detect presence of parent pattern for overrides.

## Why?
As reported in the mentioned issue, the aspect ratio, and other attributes, don't work as expected when editing an image with pattern overrides.

## How?
Ensuring that we only skip setting non-bound attributes when users are actually overriding a pattern, and not when they are editing the original one. To do so, it check if the block has a parent pattern with the code of [this other pull request](https://github.com/WordPress/gutenberg/pull/62861).

## Testing Instructions
There are three main issues to consider:

* **Aspect ratio works as expected**.
1. Create a pattern
2. Add an image block
3. Enable bindings for the image block
4. Upload an image to the block
5. Try adjusting Aspect Ratio, observe it works.

**Before**

https://github.com/WordPress/gutenberg/assets/34552881/7f935bb6-7f19-4cba-ae1b-d3d511779c73

**After**

https://github.com/WordPress/gutenberg/assets/34552881/1e675341-a4a4-4a2e-a2c1-fdfcf41c73e7

* **Ensure caption is not added while replacing the original image in a post**

1. Add an image block by picking an image from the Media Library that does not have a caption.
2. Click create a synced pattern from the block toolbar.
3. Select the Pattern and choose 'Edit original'.
4. In the pattern editor, select the Image block and "Enable overrides" in the Advanced section.
5. Go back to the post editor.
6. Replace the image with one from the Media Library that does have a caption.
7. Check that the new image does not have a caption.

* **Ensure caption is not added when replacing the image in the original pattern**.

1. Add an image block by picking an image from the Media Library that does not have a caption.
2. Click create a synced pattern from the block toolbar.
3. Select the Pattern and choose 'Edit original'.
4. In the pattern editor, select the Image block and "Enable overrides" in the Advanced section.
6. Replace the image with one from the Media Library that does have a caption.
7. Check that the new image does not have a caption.

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -176,6 +176,20 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 					}
 
 					const keptAttributes = { ...nextAttributes };
+
+					// Don't update caption and href in pattern overrides until they are supported.
+					if ( hasPatternOverridesDefaultBinding ) {
+						delete keptAttributes?.caption;
+						delete keptAttributes?.href;
+					}
+
+					// Don't run bindings logic when a block is using pattern overrides but
+					// the user is editing the original pattern.
+					if ( ! hasParentPattern ) {
+						setAttributes( keptAttributes );
+						return;
+					}
+
 					const updatesBySource = new Map();
 
 					// Loop only over the updated attributes to avoid modifying the bound ones that haven't changed.
@@ -232,20 +246,13 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 						}
 					}
 
+					// Only apply normal attribute updates to blocks
+					// that have partial bindings. Currently this is
+					// only skipped for pattern overrides sources.
 					if (
-						// Don't update non-connected attributes if the block is using pattern overrides
-						// and the editing is happening while using the pattern (not editing the original).
-						! (
-							hasPatternOverridesDefaultBinding &&
-							hasParentPattern
-						) &&
+						! hasPatternOverridesDefaultBinding &&
 						Object.keys( keptAttributes ).length
 					) {
-						// Don't update caption and href until they are supported.
-						if ( hasPatternOverridesDefaultBinding ) {
-							delete keptAttributes?.caption;
-							delete keptAttributes?.href;
-						}
 						setAttributes( keptAttributes );
 					}
 				} );
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -179,7 +179,10 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 
 					// Don't run bindings logic when a block is using pattern overrides but
 					// the user is editing the original pattern.
-					if ( ! hasParentPattern ) {
+					if (
+						! hasParentPattern &&
+						hasPatternOverridesDefaultBinding
+					) {
 						// Don't update caption and href in pattern overrides until they are supported.
 						delete keptAttributes?.caption;
 						delete keptAttributes?.href;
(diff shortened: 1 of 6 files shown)


packages/block-editor/src/hooks/use-bindings-attributes.js
```diff
@@ -93,10 +94,25 @@ export function canBindAttribute( blockName, attributeName ) {
...
+					select( blocksStore )
+				);
+				const { getBlockParentsByBlockName } = unlock(
+					select( blockEditorStore )
+				);
+
+				return {
+					sources: getAllBlockBindingsSources(),
+					hasParentPattern:
+						getBlockParentsByBlockName( clientId, 'core/block' )
```
ellatrix: This adds a block editor store subscription for all bindable block, no? Any early returns we can set here to avoid it?

talldan: I also noticed this lookup for a parent pattern block has been added in a few places, another example here and there are more for the UI parts of pattern overrides - https://github.com/WordPress/gutenberg/blob/7bca2fadddde7b2b2f62823b8a4b81378f117412/packages/editor/src/bindings/pattern-overrides.js#L16-L20

We could look into using block context as a way of checking for a parent pattern, though it'd mean every (supported) block needs to subscribe to that context.

There isn't an obvious context attribute that the pattern block could provide. There's only `ref` (which may be removed in the future when the block uses a `slug`), and `content` (which can be `undefined` so might not reliable, unless it's changed so a default value is provided).

talldan: I made a PR based on this that shows how that would work - https://github.com/WordPress/gutenberg/pull/62861.

SantosGuillamot: I answered you there, but I think using context makes sense. Thanks for working on that! 🙂

cbravobernal:
I can reproduce the bug in trunk and can confirm that this PR fixes it.

packages/block-library/src/image/image.js
```diff
@@ -508,7 +504,12 @@ export default function Image( {
 					: __( 'Connected to dynamic data' ),
 			};
 		},
-		[ clientId, isSingleSelected, metadata?.bindings ]
+		[
+			arePatternOverridesEnabled,
```
cbravobernal, line 508:
```suggestion
			patternOverridesEnabled,
```
We can remove the verb from the variable in follow-up PRs.

cbravobernal (approved):
LGTM and works as expected.

talldan (approved):
Thanks @SantosGuillamot, this is working well for me, and it's a nice minimal change.

PR #62829, opened 2024-06-25
Global Styles: Make a shared component for typography and color preview

scruffian:
## What?
Use the same approach to display typography variations that we use for color variations.

## Why?
If we use a similar approach to display these presets it will be easier to update them to have shared components. This will be possible once https://github.com/WordPress/gutenberg/pull/62827 is merged.

## How?
Creates a new `StylesPreviewTypography` component similar to `StylesPreviewColors`.

## Testing Instructions

0. Switch to a theme that uses color and typography variations (I'm using a modified TT4 with this variation: https://gist.github.com/scruffian/3dab906ee04d3c1c689bb08d23198f84)
1. Open the site editor
2. Open Global Styles > Typography
3. Check you can still see the list of Typography variations

### DIFF
--- a/packages/edit-site/src/components/global-styles/variations/variations-typography.js
+++ b/packages/edit-site/src/components/global-styles/variations/variations-typography.js
@@ -29,24 +29,22 @@ export default function TypographyVariations( { title, gap = 2 } ) {
 				gap={ gap }
 				className="edit-site-global-styles-style-variations-container"
 			>
-				{ typographyVariations &&
-					typographyVariations.length &&
-					typographyVariations.map( ( variation, index ) => {
-						return (
-							<Variation
-								key={ index }
-								variation={ variation }
-								property="typography"
-								showTooltip
-							>
-								{ () => (
-									<StylesPreviewTypography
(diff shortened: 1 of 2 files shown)


jeryj:
There's a [stray `a` in your json test file](https://gist.github.com/scruffian/3dab906ee04d3c1c689bb08d23198f84#file-gistfile1-txt-L88) :)

MaggieCabrera (approved):
The code looks good to me and it's working as intended

packages/edit-site/src/components/global-styles/variations/variations-typography.js
```diff
@@ -34,38 +31,22 @@ export default function TypographyVariations( { title, gap = 2 } ) {
 			>
 				{ typographyVariations &&
 					typographyVariations.length &&
```
MaggieCabrera: this can go too, right? we are already checking this before and escaping early

scruffian: Thanks! Done in 1b5d0c341e

PR #62835, opened 2024-06-25
Docs/iAPI: Fix wrong code snippet in data-wp-run example

luisherranz:
## What?

Fix a wrong code snippet in the `data-wp-run` example.

Discovered by @fabiankaegy and reported in Slack: https://wordpress.slack.com/archives/C071CRKGKUP/p1719289120450429

## Why?

The `data-wp-run` callbacks run in the body of the Preact component, and `ref` is basically a `useRef` hook, which means that `ref` is `null` on the first render.

All this means that, in contrary to how it works in `data-wp-watch` and `data-wp-init`, which are basically fancy versions of `useEffect`, you need to access `ref` inside a `useEffect`-like hook (`useEffect`, `useInit`, `useWatch`…) when you use `data-wp-run`, exactly like you would use it in (P)React:

```js
const Comp = () => {
  const ref = useRef();
  console.log(ref.current); // null on first render
  useEffect(() => {
    console.log(ref.current); // it's fine here
  });
  return <div ref={ref}>...</div>;
};
```

## How?

In this specific example, I moved the `getElement()` call to the `useEffect` hook of the `useInView` hook, so `ref` is not `null` anymore.

### DIFF
--- a/docs/reference-guides/interactivity-api/api-reference.md
+++ b/docs/reference-guides/interactivity-api/api-reference.md
@@ -535,13 +535,14 @@ import { store, useState, useEffect } from '@wordpress/interactivity';
 
 // Unlike `data-wp-init` and `data-wp-watch`, you can use any hooks inside
 // `data-wp-run` callbacks.
-const useInView = ( ref ) => {
+const useInView = () => {
   const [ inView, setInView ] = useState( false );
   useEffect( () => {
+    const { ref } = getElement();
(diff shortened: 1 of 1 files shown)


luisherranz:
I'm going to add a paragraph to explain how `ref` works within `data-wp-run` as well.

fabiankaegy (approved):
Thanks for the updates here :)

PR #62846, opened 2024-06-25
DataViews: simplify selection setting

ellatrix:
## What?

* Changes the _internal_ `onSelectionChange` function to simply be the React state set function, and set selection via IDs rather than items.
* This allows us to internally set selection with a callback.
* Simplify some of the internal state setting to use a callback function.
* Fix types.

## Why?

Simplicity.

### DIFF
--- a/packages/dataviews/src/bulk-actions.tsx
+++ b/packages/dataviews/src/bulk-actions.tsx
@@ -15,6 +15,7 @@ import { useRegistry } from '@wordpress/data';
  */
 import { unlock } from './lock-unlock';
 import type { Action, ActionModal } from './types';
+import type { SetSelection } from './private-types';
 
 const {
 	DropdownMenuV2: DropdownMenu,
@@ -46,7 +47,7 @@ interface BulkActionsProps< Item > {
 	data: Item[];
 	actions: Action< Item >[];
 	selection: string[];
-	onSelectionChange: ( selection: Item[] ) => void;
+	onSelectionChange: SetSelection;
 	getItemId: ( item: Item ) => string;
 }
 
@@ -248,7 +249,11 @@ export default function BulkActions< Item >( {
 						disabled={ areAllSelected }
 						hideOnClick={ false }
 						onClick={ () => {
-							onSelectionChange( selectableItems );
+							onSelectionChange(
+								selectableItems.map( ( item ) =>
+									getItemId( item )
+								)
+							);
 						} }
 						suffix={ numberSelectableItems }
 					>
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -7,7 +7,7 @@ import type { ComponentType } from 'react';
  * WordPress dependencies
  */
 import { __experimentalHStack as HStack } from '@wordpress/components';
-import { useMemo, useState, useCallback } from '@wordpress/element';
+import { useMemo, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -22,6 +22,7 @@ import BulkActions from './bulk-actions';
 import { normalizeFields } from './normalize-fields';
 import BulkActionsToolbar from './bulk-actions-toolbar';
 import type { Action, Field, View, ViewBaseProps } from './types';
+import type { SetSelection, SelectionOrUpdater } from './private-types';
 
 type ItemWithId = { id: string };
 
@@ -40,7 +41,7 @@ type DataViewsProps< Item > = {
 	};
 	supportedLayouts: string[];
 	selection?: string[];
-	setSelection?: ( selection: string[] ) => void;
+	setSelection?: SetSelection;
 	onSelectionChange?: ( items: Item[] ) => void;
 } & ( Item extends ItemWithId
 	? { getItemId?: ( item: Item ) => string }
@@ -83,26 +84,22 @@ export default function DataViews< Item >( {
 	onSelectionChange = defaultOnSelectionChange,
 }: DataViewsProps< Item > ) {
 	const [ selectionState, setSelectionState ] = useState< string[] >( [] );
-	let selection, setSelection;
-	if (
-		selectionProperty !== undefined &&
-		setSelectionProperty !== undefined
-	) {
-		selection = selectionProperty;
-		setSelection = setSelectionProperty;
-	} else {
-		selection = selectionState;
-		setSelection = setSelectionState;
-	}
+	const isUncontrolled =
+		selectionProperty === undefined || setSelectionProperty === undefined;
+	const selection = isUncontrolled ? selectionState : selectionProperty;
+	const setSelection = isUncontrolled
+		? setSelectionState
+		: setSelectionProperty;
 	const [ openedFilter, setOpenedFilter ] = useState< string | null >( null );
 
-	const onSetSelection = useCallback(
-		( items: Item[] ) => {
-			setSelection( items.map( ( item ) => getItemId( item ) ) );
-			onSelectionChange( items );
-		},
-		[ setSelection, getItemId, onSelectionChange ]
-	);
+	function setSelectionWithChange( value: SelectionOrUpdater ) {
+		const newValue =
+			typeof value === 'function' ? value( selection ) : value;
+		onSelectionChange(
+			data.filter( ( item ) => newValue.includes( getItemId( item ) ) )
+		);
+		return setSelection( value );
+	}
 
 	const ViewComponent = VIEW_LAYOUTS.find( ( v ) => v.type === view.type )
 		?.component as ComponentType< ViewBaseProps< Item > >;
@@ -149,7 +146,7 @@ export default function DataViews< Item >( {
 						<BulkActions
 							actions={ actions }
 							data={ data }
-							onSelectionChange={ onSetSelection }
+							onSelectionChange={ setSelectionWithChange }
 							selection={ _selection }
 							getItemId={ getItemId }
 						/>
@@ -168,7 +165,7 @@ export default function DataViews< Item >( {
 				getItemId={ getItemId }
 				isLoading={ isLoading }
 				onChangeView={ onChangeView }
-				onSelectionChange={ onSetSelection }
+				onSelectionChange={ setSelectionWithChange }
 				selection={ _selection }
 				setOpenedFilter={ setOpenedFilter }
 				view={ view }
@@ -184,7 +181,7 @@ export default function DataViews< Item >( {
 						data={ data }
 						actions={ actions }
 						selection={ _selection }
-						onSelectionChange={ onSetSelection }
+						onSelectionChange={ setSelectionWithChange }
 						getItemId={ getItemId }
 					/>
 				) }
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -3,6 +3,11 @@
  */
 import type { ReactElement, ReactNode } from 'react';
 
+/**
+ * Internal dependencies
+ */
+import type { SetSelection } from './private-types';
+
 export type SortDirection = 'asc' | 'desc';
 
 /**
@@ -383,7 +388,7 @@ export interface ViewBaseProps< Item > {
 	getItemId: ( item: Item ) => string;
 	isLoading?: boolean;
 	onChangeView( view: View ): void;
-	onSelectionChange: ( items: Item[] ) => void;
+	onSelectionChange: SetSelection;
 	selection: string[];
 	setOpenedFilter: ( fieldId: string ) => void;
 	view: View;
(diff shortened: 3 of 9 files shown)


packages/dataviews/src/types.ts
```diff
@@ -1,7 +1,9 @@
 /**
  * External dependencies
  */
-import type { ReactElement, ReactNode } from 'react';
+import type { ReactElement, ReactNode, Dispatch, SetStateAction } from 'react';
+
+export type setSelection = Dispatch< SetStateAction< string[] > >;
```
youknowriad: What does "dispatch" do?

--
The rest of the comment here is not blocking, mostly discussion and typescript learnings / guidelines discussion.

I think there's a balance to be found between the types that make sense as exported or not. For smaller types like this, I think duplication may be fine (avoid indirection). But I don't know if folks more familiar with typescript think. cc @sirreal @jsnajdr 

Also curious what @mcsf thinks here.

packages/dataviews/src/view-table.tsx
```diff
@@ -297,7 +300,6 @@ function TableRow< Item extends AnyItem >( {
 	selection,
 	getItemId,
 	onSelectionChange,
-	data,
```
youknowriad, line 299: I'm guessing this needs to be removed from some time `TableRowProps` type no?

packages/dataviews/src/types.ts
ellatrix: I can revert 2e2f38c4764242cbcb739edfc083901eea616e3f if you like

packages/dataviews/src/view-grid.tsx
```diff
@@ -40,7 +46,6 @@ interface GridItemProps< Item extends AnyItem > {
 
 function GridItem< Item extends AnyItem >( {
 	selection,
-	data,
```
youknowriad, line 43: Should this get removed from `GridItemProps` type.

ellatrix: Hm, why did TS not complain?

packages/dataviews/src/types.ts
youknowriad: Oh I see what Dispatch does, it's useful I'd say :) 

Curious about the guidelines on exposing this kind of low level types.

packages/dataviews/src/view-grid.tsx
youknowriad: I'm guessing because we're deconstructing the props here but nothing says in the code that "data" is not present in the props.

packages/dataviews/src/view-list.tsx
```diff
@@ -335,10 +335,8 @@ export default function ViewList< Item extends AnyItem >(
 			)
 	);
 
-	const onSelect = useCallback(
-		( item: Item ) => onSelectionChange( [ item ] ),
-		[ onSelectionChange ]
-	);
+	const onSelect = ( item: Item ) =>
+		onSelectionChange( [ getItemId( item ) ] );
```
ellatrix, line 337: Note that `useCallback` is not needed here, it's just used in a React event callback down the line.

packages/dataviews/src/types.ts
ellatrix: Yeah, I forgot I changed a public type too. I can revert it to `( selection: string[] | ( ( prevState: string[] ) => string[] ) ) => void` to unblock if you prefer.

packages/dataviews/src/dataviews.tsx
```diff
@@ -92,13 +100,17 @@ export default function DataViews< Item extends AnyItem >( {
...
-	const onSetSelection = useCallback(
-		( items: Item[] ) => {
-			setSelection( items.map( ( item ) => getItemId( item ) ) );
-			onSelectionChange( items );
-		},
-		[ setSelection, getItemId, onSelectionChange ]
-	);
+	const prevSelection = usePrevious( selection );
+
+	useEffect( () => {
```
jorgefilipecosta: Instead of an effect, what do you think of keeping exactly the other changes but instead of having an effect we create a callback with exactly the same signature as setSelection so all the other changes are valid, but that also calls the onSelectionChange inside and not only the setSelection? It would remove the need for the effect.

ellatrix: It's a bit less clean. Previously it was using a callback. If we move this all in a callback, it will pretty much re-render all the time based on these dependencies. I guess I can check if that callback was really needed in the first place.

Why do you not like the effect? Cause it could be triggered from outside selection changes? I think that would be expected?

packages/dataviews/src/view-grid.tsx
sirreal: It looks like `AnyItem` is `Record<string, any>`, so any string key is associated with a value of type `any`. `Item` extends `AnyItem`, so `Item` must also have `string` -> `any` mapping. This weakens the type system, it would be preferable to have a better type to extend or remove the `AnyItem` type somehow.

packages/dataviews/src/types.ts
jsnajdr: My feedback is that it just took me non-trivial effort to figure out what this `Dispatch<SetStateAction>` actually is. Better to write it plainly as @ellatrix proposes. Also make the first letter of the type name uppercase, `SetSelection` 🙂

packages/dataviews/src/dataviews.tsx
jorgefilipecosta: > Why do you not like the effect? Cause it could be triggered from outside selection changes? I think that would be expected?

Yah, the main difference with using an effect is that even if the consumer is the one doing the selection change the consumer will still get a callback on the onSelectionChange that may be unexpected. I'm not sure what is the good practice on these cases.

ellatrix: @jorgefilipecosta I changed it

packages/dataviews/src/types.ts
sirreal: Convention is typically to capitalize types, it might be good to change the type name `setSelection` to `SetSelection`.

This type seems to be used repeatedly and could be error prone to write out over and over, it seems like there's value in a reusable type.

As far as exporting the type or not, I'm not sure how important the questions is. The truth is that any types that are part of the public interface are effectively public types because they will be visible externally.

This type _is_ directly exported, so I guess that may be the question:

https://github.com/WordPress/gutenberg/blob/1ee2e3c10f60212a9e1e1c27d62d9000401653aa/packages/dataviews/src/index.ts#L4

There, I would tend to be cautious. Is this something that is immediately useful for third parties? If not, then we might still export a single type and use it throughout the repository, but not export it directly. Again, ultimately the type will be visible and therefore accessible externally, the question is really about how much of the public API or how visible it should be. It's not the same to access it in laborious ways or just import the type.

It looks like this type is changing, but it may be strictly wider? All values that satisfied the old type also satisfy the new type? In that case I don't think it would qualify as a breaking change.

One thing that might be interesting around types is proposed by @luisherranz in https://github.com/WordPress/gutenberg/pull/62400. That PR proposes adding automated tests specifically for types. That could be extremely helpful for determining the types of changes in packages because any change that would make code using a previous versions types _invalid_ would be a breaking change.

ellatrix: Thanks for the feedback. I capitalised it, made the type private, and used a simpler type.

packages/dataviews/src/view-grid.tsx
jsnajdr: @sirreal You're right, `AnyItem` is currently hiding several type errors, because it pollutes the types with `any`, and also it turns out to be redundant, can be removed completely. We discussed it already in https://github.com/WordPress/gutenberg/pull/62647#discussion_r1648576427 and here's my draft PR (on top of #62647) that removes `AnyItem` and fixes the three type error that the removal has revealed: #62856.

packages/dataviews/src/types.ts
luisherranz: > One thing that might be interesting around types is proposed by @luisherranz in https://github.com/WordPress/gutenberg/pull/62400.

Thank you for mentioning this, @sirreal.

Indeed, I want to propose a way to test types in Gutenberg. This PR shows how it could be done using `tsc`.

jorgefilipecosta:
Hi @ellatrix, I think there is a bug, on the page and templates table view I'm not able to select items by clicking on their checkbox.

packages/dataviews/src/dataviews.tsx
```diff
@@ -79,26 +80,22 @@ export default function DataViews< Item extends AnyItem >( {
...
-		selectionProperty !== undefined &&
-		setSelectionProperty !== undefined
-	) {
-		selection = selectionProperty;
-		setSelection = setSelectionProperty;
-	} else {
-		selection = selectionState;
-		setSelection = setSelectionState;
-	}
+	const selection =
```
oandregal: This is not quite the same logic as before: the `selection` could end up being `selectionState` while `setSelection` being `setSelectionProperty`.

packages/dataviews/src/bulk-actions.tsx
```diff
@@ -248,7 +249,11 @@ export default function BulkActions< Item extends AnyItem >( {
 						disabled={ areAllSelected }
 						hideOnClick={ false }
 						onClick={ () => {
-							onSelectionChange( selectableItems );
+							onSelectionChange(
```
oandregal, line 252: Before, this used an `Item[]`, now it uses `string[]`. Is this problematic?

oandregal: Consumers are expecting an `Item[]` ([pages](https://github.com/WordPress/gutenberg/blob/b86375d38920c3938b3ff6ecd6a1aa24af5ab7a8/packages/edit-site/src/components/page-pages/index.js#L227), [templates](https://github.com/WordPress/gutenberg/blob/b86375d38920c3938b3ff6ecd6a1aa24af5ab7a8/packages/edit-site/src/components/page-templates/index.js#L240)), so they need to be updated.

oandregal: What should `onSelectionChange` offer to consumers, ids (`string[]`) or items (`Item[]`)? 

I'm inclined to work with `Item[]`.

It enables consumers to immediately act on any needs they have instead of having to look the item up based on id. For example, upon click, the consumer may want to display extra info about the item in a panel outside of dataviews — if they have the item they already have they info, otherwise, they'd have to look it up based on id, first.

packages/dataviews/src/view-table.tsx
```diff
@@ -337,27 +338,11 @@ function TableRow< Item extends AnyItem >( {
...
-							data.filter( ( _item ) => {
-								const itemId = getItemId?.( _item );
-								return (
-									itemId !== id &&
-									selection.includes( itemId )
-								);
-							} )
-						);
-					}
+					onSelectionChange( ( state ) =>
```
oandregal: Can all these calls that use the `updater` be rewritten as?

```js
onSelectionChange(
    selection.includes(id)
        ? selection.filter((itemId) => id !== itemId)
        : [...selection, id]
);
```

I haven't checked them all but it seems that where `onSelectionChange` is available, `selection` also is. If we can, it makes the updater redundant, and the API would simpler without it.

packages/dataviews/src/bulk-actions.tsx
ellatrix: This is not the public API

packages/dataviews/src/dataviews.tsx
ellatrix: Fixed

packages/dataviews/src/view-table.tsx
ellatrix: > the API would simpler without it

I haven't changed the API at all, these are just React setState calls. Updater functions are nice because, if you're in memoized callbacks, you don't depend on the state, though I've removed all of those. It's also surfaced an issue: if you click on the checkbox, we're calling setState twice. This worked before because it would set state twice with the same value, but with the the updater, the second call is undoing the last call.

packages/dataviews/src/bulk-actions.tsx
ellatrix: `onSelectionChange` still offers `Item[]` to consumers, I didn't change that public API at all. This is just setting internal state, which was already storing only the IDs. Instead of passing Items around internally it now just passes IDs around.

ellatrix:
@jorgefilipecosta The bug is fixed, the cause was this: https://github.com/WordPress/gutenberg/pull/62846#discussion_r1655262562

Imo that should be fixed, we are calling setState twice, which is working in trunk because we don't use an updater function. But I think it's outside the scope of this PR so I changed it back to calling setState with the value.

packages/dataviews/src/bulk-actions.tsx
oandregal: Oh, I had just glanced at the code, sorry. Gave it a second/deeper look now and I see the changes 👍

oandregal (approved):

packages/dataviews/src/view-table.tsx
oandregal: I see your point, thanks for elaborating 👍

jorgefilipecosta: Interesting finding regarding the duplicate setState call, I will have a look into that and see if I can fix it.

PR #62847, opened 2024-06-25
Global Styles: Allow variations to be filtered by multiple properties

scruffian:
## What?
This updates several functions to allow presets to incorporate several properties.

## Why?
This was originally proppsed as part of https://github.com/WordPress/gutenberg/pull/62709, and was also part of https://github.com/WordPress/gutenberg/pull/62741. It's not clear what direction we want to go with this, but this code maintains the same functionality while adding flexibility by allowing you to provide an array to these functions rather than a string.

## How?
Update several functions to accept an array rather than a string.

## Testing Instructions
0. Use a theme with style variations (like TT4)
1. Open the Site Editor
2. Open Global Styles
3. Check that you can still see color variations and style variations and that there are no typography variations.

### DIFF
--- a/packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
+++ b/packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
@@ -18,14 +18,14 @@ const { GlobalStylesContext } = unlock( blockEditorPrivateApis );
 const { mergeBaseAndUserConfigs } = unlock( editorPrivateApis );
 
 /**
- * Removes all instances of a property from an object.
+ * Removes all instances of properties from an object.
  *
- * @param {Object} object   The object to remove the property from.
- * @param {string} property The property to remove.
+ * @param {Object} object     The object to remove the properties from.
+ * @param {string} properties The properties to remove.
  * @return {Object} The modified object.
  */
-export function removePropertyFromObject( object, property ) {
-	if ( ! property || typeof property !== 'string' ) {
+export function removePropertiesFromObject( object, properties ) {
+	if ( ! properties || typeof properties !== 'object' ) {
 		return object;
 	}
 
@@ -38,25 +38,25 @@ export function removePropertyFromObject( object, property ) {
 	}
 
 	for ( const key in object ) {
-		if ( key === property ) {
+		if ( properties.includes( key ) ) {
 			delete object[ key ];
 		} else if ( typeof object[ key ] === 'object' ) {
-			removePropertyFromObject( object[ key ], property );
+			removePropertiesFromObject( object[ key ], properties );
 		}
 	}
 	return object;
 }
 
 /**
- * Fetches the current theme style variations that contain only the specified property
+ * Fetches the current theme style variations that contain only the specified properties
  * and merges them with the user config.
  *
- * @param {Object} props          Object of hook args.
- * @param {string} props.property The property to filter by.
+ * @param {Object} props            Object of hook args.
+ * @param {string} props.properties The properties to filter by.
  * @return {Object[]|*} The merged object.
  */
 export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
-	property,
+	properties,
 } ) {
 	const { variationsFromTheme } = useSelect( ( select ) => {
 		const _variationsFromTheme =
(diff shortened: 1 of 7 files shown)


packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -74,51 +74,54 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
 			} );
 
 		return [
-			userVariationWithoutProperty,
-			...variationsWithSinglePropertyAndBase,
+			userVariationWithoutProperties,
+			...variationsWithPropertiesAndBase,
 		];
-	}, [ property, userVariation, variationsFromTheme ] );
+	}, [ properties, userVariation, variationsFromTheme ] );
```
youknowriad: I'm not familiar with this code at all but there's something that is probably not right here. If you're using properties as a dependency for a useMemo call, I'm guessing that you expect the same object instance to be returned here if the properties don't change (maybe for some performance reasons), but here it's almost never the case since properties is passed as an inline array, so properties will always trigger a recompute and a new instance returned by useMemo.

scruffian: Ah yes good point. Would it work to spread the properties in the dependency array, like this?

```suggestion
	}, [ ...properties, userVariation, variationsFromTheme ] );
```

youknowriad: That wouldn't work either, because the size of the properties array might change and I believe it's not allowed in these react hooks to have a variable size.

scruffian: What if we convert the array to a string?

youknowriad: It would work I think.

scruffian: Done in [ea0be03](https://github.com/WordPress/gutenberg/pull/62847/commits/ea0be0360b6e1372d22a1cf84aba58800be12e39)

youknowriad:
I'm not familiar enough with this part of the code to do a full review here. I'll let others chime in.

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -38,25 +38,25 @@ export function removePropertyFromObject( object, property ) {
...
 
 /**
- * Fetches the current theme style variations that contain only the specified property
+ * Fetches the current theme style variations that contain only the specified properties
  * and merges them with the user config.
  *
- * @param {Object} props          Object of hook args.
- * @param {string} props.property The property to filter by.
+ * @param {Object} props            Object of hook args.
+ * @param {string} props.properties The properties to filter by.
```
ramonjd: Should this type be `string[]`?

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -18,14 +18,14 @@ const { GlobalStylesContext } = unlock( blockEditorPrivateApis );
 const { mergeBaseAndUserConfigs } = unlock( editorPrivateApis );
 
 /**
- * Removes all instances of a property from an object.
+ * Removes all instances of properties from an object.
  *
- * @param {Object} object   The object to remove the property from.
- * @param {string} property The property to remove.
+ * @param {Object} object     The object to remove the properties from.
+ * @param {string} properties The properties to remove.
```
ramonjd: Same here, is it `string[]`?

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -18,14 +18,14 @@ const { GlobalStylesContext } = unlock( blockEditorPrivateApis );
...
  * @return {Object} The modified object.
  */
-export function removePropertyFromObject( object, property ) {
-	if ( ! property || typeof property !== 'string' ) {
+export function removePropertiesFromObject( object, properties ) {
+	if ( ! properties || typeof properties !== 'object' ) {
```
ramonjd:
```suggestion
	if ( ! properties?.length ) {
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -38,25 +38,25 @@ export function removePropertyFromObject( object, property ) {
...
  *
- * @param {Object} props          Object of hook args.
- * @param {string} props.property The property to filter by.
+ * @param {Object} props            Object of hook args.
+ * @param {string} props.properties The properties to filter by.
  * @return {Object[]|*} The merged object.
  */
 export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
-	property,
+	properties,
```
ramonjd: Would it be useful to define defaults here to prevent misuse and toString coercion on undefined values? 

```suggestion
	properties = [],
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -74,51 +74,54 @@ export function useCurrentMergeThemeStyleVariationsWithUserConfig( {
...
  * maintain the original object tree structure.
- * The function is recursive, so it will perform a deep search for the given property.
- * E.g., the function will return `{ a: { b: { c: { test: 1 } } } }` if the property is `test`.
+ * The function is recursive, so it will perform a deep search for the given properties.
+ * E.g., the function will return `{ a: { b: { c: { test: 1 } } } }` if the properties are  `[ 'test' ]`.
  *
- * @param {Object} object   The object to filter
- * @param {Object} property The property to filter by
+ * @param {Object} object     The object to filter
+ * @param {Array}  properties The properties to filter by
```
ramonjd:
```suggestion
 * @param {string[]}  properties The properties to filter by
```

Or `Array<string>` - they're the same, but some folks have a preference. 🤷🏻

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -128,23 +131,23 @@ export const filterObjectByProperty = ( object, property ) => {
...
- * Compares a style variation to the same variation filtered by a single property.
- * Returns true if the variation contains only the property specified.
+ * Compares a style variation to the same variation filtered by the specified properties.
+ * Returns true if the variation contains only the properties specified.
  *
- * @param {Object} variation The variation to compare.
- * @param {string} property  The property to compare.
- * @return {boolean} Whether the variation contains only a single property.
+ * @param {Object} variation  The variation to compare.
+ * @param {string} properties The properties to compare.
```
ramonjd:
```suggestion
 * @param {string[]} properties The properties to compare.
```

packages/edit-site/src/hooks/use-theme-style-variations/use-theme-style-variations-by-property.js
```diff
@@ -128,23 +131,23 @@ export const filterObjectByProperty = ( object, property ) => {
...
+export function isVariationWithProperties( variation, properties ) {
+	const variationWithProperties = filterObjectByProperties(
 		cloneDeep( variation ),
-		property
+		properties
 	);
 
 	return (
-		JSON.stringify( variationWithProperty?.styles ) ===
+		JSON.stringify( variationWithProperties?.styles ) ===
```
ramonjd: Optional, but also available is `areGlobalStyleConfigsEqual`

```js
const { areGlobalStyleConfigsEqual } = unlock(
	blockEditorPrivateApis
);
...
return areGlobalStyleConfigsEqual( variationWithProperties, variation);
```

ramonjd (commented):
I couldn't see any regressions in the editor - between trunk and this PR the style variations in TT4 are the same in both the Site Editor sidebar (view mode) and the global styles inspector. 👍🏻 

The change makes sense to me  - extracting several properties at a time.

Looks like the unit tests need to be updated.

scruffian:
Thanks for the review! I have updated the tests, hopefully this is good to go :)

ramonjd (approved):
Thanks for the updates @scruffian 🙇🏻 

Tests well for me

PR #62848, opened 2024-06-25
CustomSelectControl V2: allow wrapping item hint to new line

ciampo:
## What?

Part of #55023

Improve how inidivual item hints are rendered for the `CustomSelectControl` V2 legacy adapter, by allowing the hint to wrap below the item's content, instead of being forced on the same line.

## Why?

Even if the previous layout is how the V1 component behaves, it does not produce good-looking UI when the select popover is narrow. And it often overridden in the editor via custom style overrides anyway, which is bad.

These changes basically bring to the component the overrides that were applied in the editor.

## How?

Using flexbox to allow the hint to break the line and render below the item's content.

### Storybook

1. Apply the following patch

<details>

<summary>Click to show diff</summary>

```diff
diff --git a/packages/components/src/custom-select-control/stories/index.story.tsx b/packages/components/src/custom-select-control/stories/index.story.tsx
index 8ff9a023c5..4636fd34d3 100644
--- a/packages/components/src/custom-select-control/stories/index.story.tsx
+++ b/packages/components/src/custom-select-control/stories/index.story.tsx
@@ -111,7 +111,8 @@ WithHints.args = {
 		{
 			key: 'medium',
 			name: 'Medium',
-			__experimentalHint: '250x250',
+			__experimentalHint:
+				'A long hint, which ultimately could have been just "250x250"',
 		},
 		{
 			key: 'large',

```
</details>

2. Load the "With Hints" storybook example for the V2 legacy adapter
3. Resize the window and check the new hint wrapping behaviour
4. Compare it with the same storybook example for the V1 component

| V1 | V2 legacy adapter - before (trunk) | V2 legacy adapter - after (this PR) |
|---|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/2a350164-8e92-479a-86bd-939a38add268" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/96e06334-3e2d-47e6-886c-359f45a2f981" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/59b58335-ad85-47dd-915d-f7d8f3cd6631" /> |

### Editor

First, apply the following patch 

<details>

<summary>Click to show diff</summary>

```diff
diff --git a/packages/block-editor/src/components/date-format-picker/index.js b/packages/block-editor/src/components/date-format-picker/index.js
index 19ec0bf8c2..f71cdd5e48 100644
--- a/packages/block-editor/src/components/date-format-picker/index.js
+++ b/packages/block-editor/src/components/date-format-picker/index.js
@@ -8,11 +8,18 @@ import {
 	TextControl,
 	ExternalLink,
 	VisuallyHidden,
-	CustomSelectControl,
 	ToggleControl,
 	__experimentalVStack as VStack,
+	privateApis as componentsPrivateApis,
 } from '@wordpress/components';
 
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { CustomSelectControlV2Legacy } = unlock( componentsPrivateApis );
+
 // So that we illustrate the different formats in the dropdown properly, show a date that is
 // somwhat recent, has a day greater than 12, and a month with more than three letters.
 const exampleDate = new Date();
@@ -128,7 +135,7 @@ function NonDefaultControls( { format, onChange } ) {
 
 	return (
 		<VStack>
-			<CustomSelectControl
+			<CustomSelectControlV2Legacy
 				label={ __( 'Choose a format' ) }
 				options={ [ ...suggestedOptions, customOption ] }
 				value={
diff --git a/packages/block-editor/src/components/date-format-picker/style.scss b/packages/block-editor/src/components/date-format-picker/style.scss
index d1ad52408d..748e43bb8d 100644
--- a/packages/block-editor/src/components/date-format-picker/style.scss
+++ b/packages/block-editor/src/components/date-format-picker/style.scss
@@ -4,13 +4,4 @@
 
 .block-editor-date-format-picker__custom-format-select-control__custom-option {
 	border-top: 1px solid $gray-300;
-
-	&.has-hint {
-		grid-template-columns: auto 30px;
-	}
-
-	.components-custom-select-control__item-hint {
-		grid-row: 2;
-		text-align: left;
-	}
 }
diff --git a/packages/block-editor/src/hooks/position.js b/packages/block-editor/src/hooks/position.js
index a5d4a0cb43..b184afca46 100644
--- a/packages/block-editor/src/hooks/position.js
+++ b/packages/block-editor/src/hooks/position.js
@@ -26,7 +26,9 @@ import { cleanEmptyObject, useStyleOverride } from './utils';
 import { unlock } from '../lock-unlock';
 import { store as blockEditorStore } from '../store';
 
-const { CustomSelectControl } = unlock( componentsPrivateApis );
+const { CustomSelectControl, CustomSelectControlV2Legacy } = unlock(
+	componentsPrivateApis
+);
 
 const POSITION_SUPPORT_KEY = 'position';
 
@@ -287,7 +289,7 @@ export function PositionPanelPure( {
 						__nextHasNoMarginBottom
 						help={ stickyHelpText }
 					>
-						<CustomSelectControl
+						<CustomSelectControlV2Legacy
 							__next40pxDefaultSize
 							className="block-editor-hooks__position-selection__select-control"
 							label={ __( 'Position' ) }
diff --git a/packages/block-editor/src/hooks/position.scss b/packages/block-editor/src/hooks/position.scss
index b3bd6b1b9e..552ee8cf7c 100644
--- a/packages/block-editor/src/hooks/position.scss
+++ b/packages/block-editor/src/hooks/position.scss
@@ -3,16 +3,3 @@
 		display: none;
 	}
 }
-
-.block-editor-hooks__position-selection__select-control__option {
-	&.has-hint {
-		grid-template-columns: auto 30px;
-		line-height: $default-line-height;
-		margin-bottom: 0;
-	}
-
-	.components-custom-select-control__item-hint {
-		grid-row: 2;
-		text-align: left;
-	}
-}
diff --git a/packages/components/src/private-apis.ts b/packages/components/src/private-apis.ts
index f55373664e..9a26b7c7e3 100644
--- a/packages/components/src/private-apis.ts
+++ b/packages/components/src/private-apis.ts
@@ -14,6 +14,7 @@ import {
 	useCompositeStore as useCompositeStoreV2,
 } from './composite/v2';
 import { default as CustomSelectControl } from './custom-select-control';
+import { default as CustomSelectControlV2Legacy } from './custom-select-control-v2/legacy-component';
 import { positionToPlacement as __experimentalPopoverLegacyPositionToPlacement } from './popover/utils';
 import { createPrivateSlotFill } from './slot-fill';
 import {
@@ -40,6 +41,7 @@ lock( privateApis, {
 	CompositeRowV2,
 	useCompositeStoreV2,
 	CustomSelectControl,
+	CustomSelectControlV2Legacy,
 	__experimentalPopoverLegacyPositionToPlacement,
 	createPrivateSlotFill,
 	ComponentsContext,


```
</details>

#### DateFormat picker

1. In the post editor, add a "Posts List" block
2. Pick a template which features the post data (most do)
3. Select the post date block, and in the sidebar interact with the date format (you may have to toggle off the "Default format" toggle)
4. Scroll the list of date format options all the way to the bottom, where the "Custom" option is 
5. Make sure that the layout matches what's currently on `trunk`

| V1 | V2 legacy adapter - before (trunk) | V2 legacy adapter - after (this PR) |
|---|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/0f44a6f2-ab77-472b-8624-58feb64c31e9" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/788ccbc6-86c3-4e92-972e-99ed15a0b1bd" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/7228d681-d724-431a-8461-48cc95acfc8b" /> |

#### Position picker

1. In the post editor, add a "Group" block
2. Select the group block, and in the sidebar expand the "Position" panel
3. Open the position select, and notice the Sticky option 
5. Make sure that the layout matches what's currently on `trunk`

| V1 | V2 legacy adapter - before (trunk) | V2 legacy adapter - after (this PR) |
|---|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/a12d7b89-7e7d-4973-ad5f-be87705bfcd9" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/8e580767-c426-414b-a462-53af2652a012" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/84403aba-8ac9-4c99-8a78-7cd468e24d2b" /> |

#### Known bugs

There are several known layout bugs related to `CustomSelectControlV2` — most of them are addressed in separate PRs (see #55023). When reviewing this PR, please focus only on the layout of option item hints.

### DIFF
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -78,7 +78,14 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 					value={ name }
 					children={ __experimentalHint ? withHint : name }
 					style={ style }
-					className={ className }
+					className={ clsx(
+						// Legacy classnames
+						'components-custom-select-control__item',
+						className,
+						{
+							'has-hint': !! __experimentalHint,
+						}
+					) }
 				/>
 			);
 		}
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -65,9 +65,7 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 		( { name, key, __experimentalHint, style, className } ) => {
 			const withHint = (
 				<Styled.WithHintItemWrapper>
-					<Styled.WithHintItemContent>
-						{ name }
-					</Styled.WithHintItemContent>
+					<span>{ name }</span>
 					<Styled.WithHintItemHint className="components-custom-select-control__item-hint">
 						{ __experimentalHint }
 					</Styled.WithHintItemHint>
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -64,12 +64,12 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	const children = options.map(
 		( { name, key, __experimentalHint, style, className } ) => {
 			const withHint = (
-				<Styled.WithHintWrapper>
-					<span>{ name }</span>
-					<Styled.ExperimentalHintItem className="components-custom-select-control__item-hint">
+				<Styled.ItemHintWrapper>
+					<Styled.ItemHintContent>{ name }</Styled.ItemHintContent>
+					<Styled.ItemHint className="components-custom-select-control__item-hint">
 						{ __experimentalHint }
-					</Styled.ExperimentalHintItem>
-				</Styled.WithHintWrapper>
+					</Styled.ItemHint>
+				</Styled.ItemHintWrapper>
 			);
 
 			return (
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -64,12 +64,14 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	const children = options.map(
 		( { name, key, __experimentalHint, style, className } ) => {
 			const withHint = (
-				<Styled.ItemHintWrapper>
-					<Styled.ItemHintContent>{ name }</Styled.ItemHintContent>
-					<Styled.ItemHint className="components-custom-select-control__item-hint">
+				<Styled.WithHintItemWrapper>
+					<Styled.WithHintItemContent>
+						{ name }
+					</Styled.WithHintItemContent>
+					<Styled.WithHintItemHint className="components-custom-select-control__item-hint">
 						{ __experimentalHint }
-					</Styled.ItemHint>
-				</Styled.ItemHintWrapper>
+					</Styled.WithHintItemHint>
+				</Styled.WithHintItemWrapper>
 			);
 
 			return (
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -66,7 +66,10 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 			const withHint = (
 				<Styled.WithHintItemWrapper>
 					<span>{ name }</span>
-					<Styled.WithHintItemHint className="components-custom-select-control__item-hint">
+					<Styled.WithHintItemHint
+					// TODO: Legacy classname. Add V1 styles are removed from the codebase
+					// className="components-custom-select-control__item-hint"
+					>
 						{ __experimentalHint }
 					</Styled.WithHintItemHint>
 				</Styled.WithHintItemWrapper>
@@ -79,12 +82,13 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 					children={ __experimentalHint ? withHint : name }
 					style={ style }
 					className={ clsx(
-						// Legacy classnames
-						'components-custom-select-control__item',
-						className,
-						{
-							'has-hint': !! __experimentalHint,
-						}
+						// TODO: Legacy classname. Add V1 styles are removed from the codebase
+						// 'components-custom-select-control__item',
+						className
+						// TODO: Legacy classname. Add V1 styles are removed from the codebase
+						// {
+						// 	'has-hint': __experimentalHint,
+						// }
 					) }
 				/>
 			);
@@ -101,7 +105,10 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 		return (
 			<Styled.SelectedExperimentalHintWrapper>
 				{ currentValue }
-				<Styled.SelectedExperimentalHintItem className="components-custom-select-control__hint">
+				<Styled.SelectedExperimentalHintItem
+				// TODO: Legacy classname. Add V1 styles are removed from the codebase
+				// className="components-custom-select-control__hint"
+				>
 					{ currentHint?.__experimentalHint }
 				</Styled.SelectedExperimentalHintItem>
 			</Styled.SelectedExperimentalHintWrapper>
@@ -132,7 +139,8 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 			size={ translatedSize }
 			store={ store }
 			className={ clsx(
-				'components-custom-select-control',
+				// TODO: Legacy classname. Add V1 styles are removed from the codebase
+				// 'components-custom-select-control',
 				classNameProp
 			) }
 			isLegacy
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -21,10 +21,11 @@ const truncateStyles = css`
 	white-space: nowrap;
 `;
 
-export const WithHintWrapper = styled.div`
+export const ItemHintWrapper = styled.div`
 	display: flex;
 	justify-content: space-between;
 	flex: 1;
+	flex-wrap: wrap;
 `;
 
 export const SelectedExperimentalHintWrapper = styled.div`
@@ -36,7 +37,11 @@ export const SelectedExperimentalHintItem = styled.span`
 	margin-inline-start: ${ space( 2 ) };
 `;
 
-export const ExperimentalHintItem = styled.span`
+export const ItemHintContent = styled.span`
+	padding-inline-end: ${ space( 4 ) };
+`;
+
+export const ItemHint = styled.span`
 	color: ${ COLORS.theme.gray[ 600 ] };
 	text-align: right;
 	margin-inline-end: ${ space( 1 ) };
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -44,7 +44,7 @@ export const ItemHintContent = styled.span`
 export const ItemHint = styled.span`
 	color: ${ COLORS.theme.gray[ 600 ] };
 	text-align: right;
-	margin-inline-end: ${ space( 1 ) };
+	padding-inline-end: ${ space( 1 ) };
 `;
 
 export const SelectLabel = styled( Ariakit.SelectLabel )`
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -168,5 +168,6 @@ export const ItemHint = styled.span`
 		text-align: initial;
 		line-height: ${ CONFIG.fontLineHeightBase };
 		padding-inline-end: ${ space( 1 ) };
+		margin-block: ${ space( 1 ) };
 	}
 `;
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -139,14 +139,6 @@ const truncateStyles = css`
 	white-space: nowrap;
 `;
 
-export const ItemHintWrapper = styled.div`
-	display: flex;
-	justify-content: space-between;
-	align-items: center;
-	flex-wrap: wrap;
-	flex: 1;
-`;
-
 export const SelectedExperimentalHintWrapper = styled.div`
 	${ truncateStyles }
 `;
@@ -156,11 +148,19 @@ export const SelectedExperimentalHintItem = styled.span`
 	margin-inline-start: ${ space( 2 ) };
 `;
 
-export const ItemHintContent = styled.span`
+export const WithHintItemWrapper = styled.div`
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	flex-wrap: wrap;
+	flex: 1;
+`;
+
+export const WithHintItemContent = styled.span`
 	padding-inline-end: ${ space( 4 ) };
 `;
 
-export const ItemHint = styled.span`
+export const WithHintItemHint = styled.span`
 	// Necessary to override V1 styles passed via the legacy
 	// '.components-custom-select-control__item-hint' class
 	${ SelectItem } && {
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -18,7 +18,7 @@ const ITEM_PADDING = space( 2 );
 export const SelectLabel = styled( Ariakit.SelectLabel )`
 	font-size: 11px;
 	font-weight: 500;
-	line-height: 1.4;
+	line-height: ${ CONFIG.fontLineHeightBase };
 	text-transform: uppercase;
 	margin-bottom: ${ space( 2 ) };
 `;
@@ -166,7 +166,7 @@ export const ItemHint = styled.span`
 	${ SelectItem } && {
 		color: ${ COLORS.theme.gray[ 600 ] };
 		text-align: initial;
-		line-height: 1.4;
+		line-height: ${ CONFIG.fontLineHeightBase };
 		padding-inline-end: ${ space( 1 ) };
 	}
 `;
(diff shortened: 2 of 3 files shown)


packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -36,10 +37,14 @@ export const SelectedExperimentalHintItem = styled.span`
...
-export const ExperimentalHintItem = styled.span`
+export const ItemHintContent = styled.span`
+	padding-inline-end: ${ space( 4 ) };
+`;
+
+export const ItemHint = styled.span`
 	color: ${ COLORS.theme.gray[ 600 ] };
 	text-align: right;
-	margin-inline-end: ${ space( 1 ) };
+	padding-inline-end: ${ space( 1 ) };
```
ciampo, line 47: Using padding here to better match legacy behaviour. Also, since in the editor the V1 styles are loaded, if we use the legacy `components-custom-select-control__item` we will get those styles applied to the V2 as well 😅 Without changing `margin-inline-end` to `padding-inline-end`, the hint would get both a margin and a padding applied

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -62,12 +63,12 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
...
-					<span>{ name }</span>
-					<Styled.ExperimentalHintItem className="components-custom-select-control__item-hint">
+				<Styled.ItemHintWrapper>
+					<Styled.ItemHintContent>{ name }</Styled.ItemHintContent>
+					<Styled.ItemHint className="components-custom-select-control__item-hint">
 						{ __experimentalHint }
-					</Styled.ExperimentalHintItem>
-				</Styled.WithHintWrapper>
+					</Styled.ItemHint>
+				</Styled.ItemHintWrapper>
```
ciampo: Notable changes:

- Converted the `span` that wraps `{ name }` to a styled component to add some `padding-inline-end`, ensuring a bit of spacing between the item content and the hint
- Renamed the remaining styled components to have a more coherent name structure

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -76,7 +77,12 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 					value={ name }
 					children={ __experimentalHint ? withHint : name }
 					style={ style }
-					className={ className }
+					className={ clsx(
+						// Legacy classnames
+						'components-custom-select-control__item',
+						!! __experimentalHint && 'has-hint',
```
ciampo: Consumers of the V1 component expect these classnames and use them to style specifically the `hint` part (try searching for `.has-hint` in the codebase).

I'll see if I can bring in more changes to the hint so that these overrides are not necessary in the editor, but it's probably sensible to leave those classnames there, at least for the legacy adapter.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -15,33 +15,6 @@ import type { CustomSelectButtonSize } from './types';
...
-	color: ${ COLORS.theme.gray[ 600 ] };
-	margin-inline-start: ${ space( 2 ) };
-`;
-
-export const ExperimentalHintItem = styled.span`
-	color: ${ COLORS.theme.gray[ 600 ] };
-	text-align: right;
-	margin-inline-end: ${ space( 1 ) };
-`;
-
```
ciampo, line 44: Moved to the bottom of the file in order to use `${ SelectItem }` in the CSS selector (see inline code comment)

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -132,3 +110,41 @@ export const SelectedItemCheck = styled( Ariakit.SelectItemCheck )`
...
+	overflow: hidden;
+	text-overflow: ellipsis;
+	white-space: nowrap;
+`;
+
+export const ItemHintWrapper = styled.div`
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	flex-wrap: wrap;
```
ciampo: Added these two rules to allow hint to wrap, and to preserve vertical centering even after adding [a smaller `line-height` for the hint](https://github.com/WordPress/gutenberg/pull/62848/files#r1654492389)

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -132,3 +110,41 @@ export const SelectedItemCheck = styled( Ariakit.SelectItemCheck )`
...
+`;
+
+export const ItemHint = styled.span`
+	// Necessary to override V1 styles passed via the legacy
+	// '.components-custom-select-control__item-hint' class
+	${ SelectItem } && {
+		color: ${ COLORS.theme.gray[ 600 ] };
+		text-align: initial;
+		line-height: 1.4;
+		padding-inline-end: ${ space( 1 ) };
```
ciampo: Using padding here to better match legacy behavior. Also, the V1 styles are loaded in the editor, if we use the legacy `components-custom-select-control__item` we will get those styles applied to the V2 as well 😅 Without changing `margin-inline-end` to `padding-inline-end`, the hint would get both a margin and a padding applied

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -132,3 +110,41 @@ export const SelectedItemCheck = styled( Ariakit.SelectItemCheck )`
...
+export const ItemHintContent = styled.span`
+	padding-inline-end: ${ space( 4 ) };
+`;
+
+export const ItemHint = styled.span`
+	// Necessary to override V1 styles passed via the legacy
+	// '.components-custom-select-control__item-hint' class
+	${ SelectItem } && {
+		color: ${ COLORS.theme.gray[ 600 ] };
+		text-align: initial;
```
ciampo: Since we use flexbox to push the hint to the right, `text-align` isn't necessary _per se_, but since V1 styles are "leaking" to this implementation too, we need to reset the value to `initial`. This allows the hint text to be correctly aligned both when it's on the same line as the item's content, and when it wraps to its own independent line.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -132,3 +110,41 @@ export const SelectedItemCheck = styled( Ariakit.SelectItemCheck )`
...
+		line-height: ${ CONFIG.fontLineHeightBase };
```
ciampo: The hint needs to use the default `line-height` value (and not the one of the option item), since it could render multi-line text. When rendered on the same line as the option item text, vertical alignment is still [guaranteed by flexbox](https://github.com/WordPress/gutenberg/pull/62848/files#r1654481023).

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -132,3 +110,42 @@ export const SelectedItemCheck = styled( Ariakit.SelectItemCheck )`
...
+		padding-inline-end: ${ space( 1 ) };
+		margin-block: ${ space( 1 ) };
```
ciampo: Adding a bit of block margin improves the vertical spacing when the hint takes its own separate line. But the amount of margin is not enough to increase the row height when the hint is on the same line as the option item content. Applying both top and bottom margins ensures good vertical centering too.

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
mirka: The name `ItemHintContent` feels wrong to me, since it wraps the item content and not the hint content.

Also, did we consider adding a `gap` instead of padding?

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -128,14 +101,19 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
...
-	display: flex;
-	align-items: center;
-	justify-content: space-between;
-	padding: ${ ITEM_PADDING };
-	font-size: ${ CONFIG.fontSize };
-	line-height: 2.15rem; // TODO: Remove this in default but keep for back-compat in legacy
-	&[data-active-item] {
-		background-color: ${ COLORS.theme.gray[ 300 ] };
+	// Necessary to override V1 styles passed via the legacy
+	// '.components-custom-select-control__item' class
```
mirka: Do you mean [these styles](https://github.com/WordPress/gutenberg/blob/0ea6751e73c4f19d48b51f4f96cc73e7edee56f5/packages/components/src/custom-select-control/style.scss#L44-L75)? I guess we can remove the `&&` hack right after we replace the v1 with v2 legacy, but also we might just want to remove the [v1 stylesheet import](https://github.com/WordPress/gutenberg/blob/0ea6751e73c4f19d48b51f4f96cc73e7edee56f5/packages/components/src/style.scss#L22) while we test v2 legacy styling because there are probably other weird interference as well.

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: > The name ItemHintContent feels wrong to me, since it wraps the item content and not the hint content.

I wasn't happy with it either but I've been struggling with good naming — [pushed a new attempt](https://github.com/WordPress/gutenberg/pull/62848/commits/4c012ec84527cc19140b8cdb37f913d2ddc9170a), hopefully things are more clear!

> Also, did we consider adding a gap instead of padding?

I initially considered using `gap`, but then realised that I needed different gaps based on whether the hint was rendering on the same line, or wrapping on the line line — and therefore I opted for padding and [margin](https://github.com/WordPress/gutenberg/pull/62848#discussion_r1654628543).

But after your question, I looked into it again, and discovered and `row-gap` and `column-gap` apply to flexbox too, and so I [refactored](https://github.com/WordPress/gutenberg/pull/62848/commits/21950ec00a39df89e61b59c03b11eb78c81c126c) accordingly — thank you for the suggestion!

packages/components/src/custom-select-control-v2/styles.ts
ciampo: We're not adding `block-margin` anymore aftering using the `row-gap` and `column-gap` props (see discussion [here](https://github.com/WordPress/gutenberg/pull/62848#discussion_r1660962088))

packages/components/src/custom-select-control-v2/styles.ts
ciampo: > Do you mean [these styles](https://github.com/WordPress/gutenberg/blob/0ea6751e73c4f19d48b51f4f96cc73e7edee56f5/packages/components/src/custom-select-control/style.scss#L44-L75)?

Yup!

> I guess we can remove the && hack right after we replace the v1 with v2 legacy

That's what I had in mind as well — I've added an extra item in the tracking issue so that we don't forget

> we might just want to remove the [v1 stylesheet import](https://github.com/WordPress/gutenberg/blob/0ea6751e73c4f19d48b51f4f96cc73e7edee56f5/packages/components/src/style.scss#L22) while we test v2 legacy styling because there are probably other weird interference as well.

That's a good point. Let's schedule some testing with/without importing V1 styles once all current style fixes are merged — added that to the tracking issues too.

mirka: Another way I just thought of is to comment out the legacy classnames we're passing in `legacy-component/index.tsx` until we're ready to remove the v1 implementation and styles.

Your call, but that might be simpler than cleaning up the small hacks we're adding in this PR.

mirka (approved):
Looks good and works well, though [this](https://github.com/WordPress/gutenberg/pull/62848#discussion_r1661496245) might be an alternative to merging in some of the temporary hacks.

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -64,12 +64,12 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	const children = options.map(
 		( { name, key, __experimentalHint, style, className } ) => {
 			const withHint = (
-				<Styled.WithHintWrapper>
+				<Styled.WithHintItemWrapper>
 					<span>{ name }</span>
-					<Styled.ExperimentalHintItem className="components-custom-select-control__item-hint">
+					<Styled.WithHintItemHint className="components-custom-select-control__item-hint">
```
tyxla: `WithHintItemHint` sounds a bit odd. Does the hint item also need the `WithHint` prefix? Can't it just be `ItemHint` or something?

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -78,7 +78,14 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 					value={ name }
 					children={ __experimentalHint ? withHint : name }
 					style={ style }
-					className={ className }
+					className={ clsx(
+						// Legacy classnames
+						'components-custom-select-control__item',
+						className,
+						{
+							'has-hint': !! __experimentalHint,
```
tyxla: Nit: should be fine without the type casting, right?

tyxla (approved):
Looking good and testing well for me 👍

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Clever! I went ahead and followed your suggestion — seems to work well!

When gradually swapping to legacy V2 we'll probably have to enable them (at least while testing) so that overrides like [this one](https://github.com/WordPress/gutenberg/blob/9764ee77d87ec000bf478cba9c5dc60709c97127/packages/block-editor/src/hooks/position.scss#L2-L4) can take effect — unless we ignore those overrides and have the component behave "vanilla".

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: It's a mouthful, I agree — although I personally like the clarity that the `WithHint` "namespace" brings when understanding the different styled components.

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: Should be!  I removed the `!!` notation (even if for now that code [will stay commented out](https://github.com/WordPress/gutenberg/pull/62848#discussion_r1662682481))

PR #62850, opened 2024-06-25
CustomSelectControlV2: tweak item inline padding based on size

ciampo:
## What?

Part of #55023 

Match the `CustomSelectControl` V2 item inline padding with the dropdown trigger (button)

## Why?

For better visual polish, and to match the V1 version of the component

## How?

- Forward the `size` prop from the root component to the item component via internal context
- Use the `size` to apply the same horizontal padding that is applied to the select trigger button

In the process, I also refactored some common inline padding values to a separate variable

## Testing Instructions

- Open the `CustomSelectControlV2` Storybook example
- open the select dropdown and observe the horizontal spacing on the trigger button and each option item
- tweak the `size` prop (and, optionally, the `__next40pxDefaultSize` prop)
- make sure that the spacing around all items is well aligned

## Screenshots or screencast 

| V1 | V2 before (trunk) | V2 after (this PR) |
|---|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/c3baafa6-c570-4ce9-8f77-e7eff98fac32" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/9da0ed87-3189-4446-bcb3-0661817a5bc7" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/7679b6b6-43ad-430b-be23-a066a34c7592" /> |

### DIFF
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -107,6 +107,8 @@ function _CustomSelect(
 			[ isLegacy ]
 		);
 
+	const contextValue = useMemo( () => ( { store, size } ), [ store, size ] );
+
 	return (
 		// Where should `restProps` be forwarded to?
 		<div className={ className }>
@@ -134,7 +136,7 @@ function _CustomSelect(
 					slide={ false }
 					onKeyDown={ onSelectPopoverKeyDown }
 				>
-					<CustomSelectContext.Provider value={ { store } }>
+					<CustomSelectContext.Provider value={ contextValue }>
 						{ children }
 					</CustomSelectContext.Provider>
 				</Styled.SelectPopover>
--- a/packages/components/src/custom-select-control-v2/styles.ts
+++ b/packages/components/src/custom-select-control-v2/styles.ts
@@ -19,6 +19,54 @@ const INLINE_PADDING = {
 	default: 16, // space(4)
 };
 
+const getSelectSize = (
+	size: NonNullable< CustomSelectButtonSize[ 'size' ] >,
+	heightProperty: 'minHeight' | 'height'
+) => {
+	const sizes = {
+		compact: {
+			[ heightProperty ]: 32,
+			paddingInlineStart: INLINE_PADDING.compact,
+			paddingInlineEnd: INLINE_PADDING.compact + chevronIconSize,
+		},
+		default: {
+			[ heightProperty ]: 40,
+			paddingInlineStart: INLINE_PADDING.default,
+			paddingInlineEnd: INLINE_PADDING.default + chevronIconSize,
+		},
+		small: {
+			[ heightProperty ]: 24,
+			paddingInlineStart: INLINE_PADDING.small,
+			paddingInlineEnd: INLINE_PADDING.small + chevronIconSize,
+		},
+	};
+
+	return sizes[ size ] || sizes.default;
+};
+
+const getSelectItemSize = (
+	size: NonNullable< CustomSelectButtonSize[ 'size' ] >
+) => {
+	// Used to visually align the checkmark with the chevron
+	const checkmarkCorrection = 6;
+	const sizes = {
+		compact: {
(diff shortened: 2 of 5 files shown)


packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -111,7 +111,7 @@ function _CustomSelect(
 					store={ store }
 				/>
 				<Styled.SelectPopover gutter={ 12 } store={ store } sameWidth>
-					<CustomSelectContext.Provider value={ { store } }>
+					<CustomSelectContext.Provider value={ { store, size } }>
```
ciampo, line 114: TODO: should probably memoize this object to avoid passing a new instance every time

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -127,21 +131,50 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
...
+				paddingInlineEnd: INLINE_PADDING.default - checkmarkCorrection,
+			},
+			small: {
+				paddingInlineStart: INLINE_PADDING.small,
+				paddingInlineEnd: INLINE_PADDING.small - checkmarkCorrection,
+			},
+		};
+
+		return sizes[ size ] || sizes.default;
+	};
```
tyxla: Should this be declared outside of `SelectItem`? Feels like we're needlessly redeclaring it on every re-render

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -127,21 +131,50 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
...
+
+		${ getSize() }
+	`;
+} );
 
 export const SelectedItemCheck = styled( Ariakit.SelectItemCheck )`
 	display: flex;
 	align-items: center;
-	margin-inline-start: ${ ITEM_PADDING };
+	margin-inline-start: ${ space( 2 ) };
```
tyxla, line 164: It's not immediately clear why we're using `space( 2 )` here but we use the hardcoded `INLINE_PADDING` values in other areas.

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -13,7 +13,11 @@ import { space } from '../utils/space';
 import { chevronIconSize } from '../select-control/styles/select-control-styles';
 import type { CustomSelectButtonSize } from './types';
 
-const ITEM_PADDING = space( 2 );
+const INLINE_PADDING = {
+	compact: 8, // space(2)
```
tyxla, line 17: I wonder if a better way would be to just export `GRID_BASE` and do a `compact: GRID_BASE * 2` calculation here?

tyxla (approved):
Nice work on getting proper consistency with v1 👍 So cool to see the v2 getting closer and closer!

Left some minor feedback, feel free to treat everything as optional and non-blocking.

🚀

packages/components/src/custom-select-control-v2/styles.ts
ciampo: I stopped using the `ITEM_PADDING` variable because, IMO, the space between the item's contents and the checkmark is not tied to the item padding. Using that variable was misleading (especially as, in this PR, we add variable padding based on the `size` prop). It just so happened that the margin before the check icon is the same as the padding for the `small`/`compact` variant of the component, but that's a coincidence

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Done in https://github.com/WordPress/gutenberg/pull/62850/commits/c49a4f306df9f547e855dabc7d22c6b24ebcbfbb

packages/components/src/custom-select-control-v2/styles.ts
ciampo: I see the reason behind your point, at the same time it feels this would warrant a bigger, separate conversation about the grid system and related utilities. I'd be inclined to keep it as-is and refactor (if necessary) separately.

PR #62853, opened 2024-06-26
Generic name for previous version of WP tests.

peterwilsoncc:
## What?

Modifies the job name for running the PHP Unit tests against the previous version of WordPress to be version agnostic.

Instead of being named `PHP x.x (WP x.x.x)...` they are named `PHP x.x (WP previous major version)...`

## Why?

Each time a new version of WordPress is released, the list of required actions in this repo's settings become out of date and need to be updated. This causes expected jobs to fail to report on pull requests, stalling the potential for merging.

> [!Warning]
>
> After merge, this will require updating the list of expected jobs in the settings. 
> 
> The goal is that this will be the last time this is required.

## How?

Replaces the WordPress matrix version number with `previous major version` and determines what that is when setting the `WP_ENV_CORE` environment variable.

Debug information is added to the docker containers to run `wp core version` and log the version being tested to the logs.

## Testing Instructions
1. Review the Unit tests job (Click Checks > Unit Tests)
2. Ensure the jobs title PHP Unit tests previous major version are running the previous major version
3. Once a previous major version of WP receives a version bump, confirm that the version is bumped in the unit tests.

N/A

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -240,6 +240,7 @@ jobs:
                   npm run wp-env run tests-wordpress php -i
                   npm run wp-env run tests-wordpress /var/www/html/wp-content/plugins/gutenberg/vendor/bin/phpunit -- --version
                   npm run wp-env run tests-wordpress locale -a
+                  npm run wp-env run tests-cli wp core version
 
             - name: Running single site unit tests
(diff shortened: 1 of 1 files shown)


talldan (approved):
Thanks for addressing this, the changes look good to me. The version can still be checked in the log output. I'll approve/merge this now and update the required checks.

PR #62856, opened 2024-06-26
DataViews: remove the AnyItem type

jsnajdr:
Removes the `AnyItem` type from dataviews, as it turns out that `Item` can be completely generic without any constraints.

Fixes several type errors that the removal revealed. Correct types for the `getItemId` and `getValue` getters that can be optional only for some `Item` subtypes because the default getter is not 100% generic. Also uses the `getItemId` function instead of non-generic `item.id` in `ViewList`.

Inspired by this discussion: https://github.com/WordPress/gutenberg/pull/62647#discussion_r1648576427

My plan is that we can finish and merge this PR after #62647 is landed.

### DIFF
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -21,9 +21,11 @@ import { VIEW_LAYOUTS } from './layouts';
 import BulkActions from './bulk-actions';
 import { normalizeFields } from './normalize-fields';
 import BulkActionsToolbar from './bulk-actions-toolbar';
-import type { Action, AnyItem, Field, View, ViewBaseProps } from './types';
+import type { Action, Field, View, ViewBaseProps } from './types';
 
-interface DataViewsProps< Item extends AnyItem > {
+type ItemWithId = { id: string };
+
+type DataViewsProps< Item > = {
 	view: View;
 	onChangeView: ( view: View ) => void;
 	fields: Field< Item >[];
@@ -31,7 +33,6 @@ interface DataViewsProps< Item extends AnyItem > {
 	searchLabel?: string;
 	actions?: Action< Item >[];
 	data: Item[];
-	getItemId?: ( item: Item ) => string;
 	isLoading?: boolean;
 	paginationInfo: {
 		totalItems: number;
@@ -41,12 +42,15 @@ interface DataViewsProps< Item extends AnyItem > {
 	selection?: string[];
 	setSelection?: ( selection: string[] ) => void;
 	onSelectionChange?: ( items: Item[] ) => void;
-}
+} & ( Item extends ItemWithId
+	? { getItemId?: ( item: Item ) => string }
+	: { getItemId: ( item: Item ) => string } );
 
-const defaultGetItemId = ( item: AnyItem ) => item.id;
+const defaultGetItemId = ( item: ItemWithId ) => item.id;
+
 const defaultOnSelectionChange = () => {};
 
-function useSomeItemHasAPossibleBulkAction< Item extends AnyItem >(
+function useSomeItemHasAPossibleBulkAction< Item >(
 	actions: Action< Item >[],
 	data: Item[]
 ) {
@@ -62,7 +66,7 @@ function useSomeItemHasAPossibleBulkAction< Item extends AnyItem >(
 	}, [ actions, data ] );
 }
 
-export default function DataViews< Item extends AnyItem >( {
+export default function DataViews< Item >( {
 	view,
 	onChangeView,
 	fields,
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -37,12 +37,12 @@ export type Operator =
 	| 'isAll'
 	| 'isNotAll';
 
-export type AnyItem = Record< string, any >;
+export type ItemRecord = Record< string, unknown >;
 
 /**
  * A dataview field for a specific property of a data type.
  */
-export interface Field< Item extends AnyItem > {
+export type Field< Item > = {
 	/**
 	 * The unique identifier of the field.
 	 */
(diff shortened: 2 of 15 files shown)


packages/dataviews/src/view-list.tsx
```diff
@@ -318,7 +316,7 @@ export default function ViewList< Item extends AnyItem >(
 	} = props;
 	const baseId = useInstanceId( ViewList, 'view-list' );
 	const selectedItem = data?.findLast( ( item ) =>
-		selection.includes( item.id )
+		selection.includes( getItemId( item ) )
```
ellatrix, line 319: So we are changing `selection` state to not be normalised ID?

packages/dataviews/src/dataviews.tsx
```diff
@@ -21,30 +21,34 @@ import { VIEW_LAYOUTS } from './layouts';
 import BulkActions from './bulk-actions';
 import { normalizeFields } from './normalize-fields';
 import BulkActionsToolbar from './bulk-actions-toolbar';
-import type { Action, AnyItem, Field, View, ViewBaseProps } from './types';
+import type { Action, Field, View, ViewBaseProps } from './types';
 
-interface DataViewsProps< Item extends AnyItem > {
+type ItemWithId = { id: string };
+
+type DataViewsProps< Item > = {
```
ellatrix, line 28: I'm not really familiar with any of this. Where is `Item` defined? Below in `Item extends ItemWithId` right? I thought an `Item` was an item in the data array? Those don't have a `getItemId` function?

ellatrix: I guess I'm not getting this part

```ts
( Item extends ItemWithId
 	? { getItemId?: ( item: Item ) => string }
 	: { getItemId: ( item: Item ) => string } )
```

jsnajdr: `Item` is a type parameter, it's completely generic. `DataViews` can work with any data type whatsoever, you can have a dataview of `number`s. As long as you provide functions to extract IDs and field values from the item.

The `Item extends ItemWithId` part says that if the `Item` is an object that has an `id` field, the `getItemId` prop is optional, because we have a suitable default for it.

But if `Item` is not compatible with `ItemWithId`, then `getItemId` is mandatory. Because the default (`item => item.id`) is not going to work with such a type.

packages/dataviews/src/dataviews.tsx
```diff
@@ -21,30 +21,34 @@ import { VIEW_LAYOUTS } from './layouts';
...
-	getItemId?: ( item: Item ) => string;
 	isLoading?: boolean;
 	paginationInfo: {
 		totalItems: number;
 		totalPages: number;
 	};
 	supportedLayouts: string[];
 	onSelectionChange?: ( items: Item[] ) => void;
-}
+} & ( Item extends ItemWithId
```
youknowriad, line 45: TIL conditions wow :P

packages/dataviews/src/view-list.tsx
youknowriad: I guess that's a bug fix.

packages/dataviews/src/dataviews.tsx
jsnajdr: Yes, with this the type system is Turing complete and you can build any program with it 😄

youknowriad:
@jsnajdr https://github.com/WordPress/gutenberg/pull/62647 This PR is blocked because of this https://github.com/WordPress/gutenberg/pull/62505#discussion_r1646480478 We need to find a good API/way to access/check permissions in dataviews while allowing global registration of actions... (Any suggestions here are welcome)

Maybe, we can try to land the current PR without 62647

packages/dataviews/src/view-list.tsx
jsnajdr: Yes. Another little suspicious thing is the optional chaining in `data?.findLast`. The types say that `data` is mandatory. If that's really true, the optional chaining is not needed. But that's a very minor issue compared with other problems.

jsnajdr:
OK I can try to rebase it onto `trunk`.

jsnajdr:
OK now this PR is independent from #62647 and can be merged directly into trunk.

packages/dataviews/src/types.ts
```diff
@@ -103,17 +97,34 @@ export interface Field< Item extends AnyItem > {
...
+			 * Defaults to `item[ field.id ]`.
+			 */
+			getValue: ( args: { item: Item } ) => any;
+	  } );
+
+export type NormalizedField< Item > = Field< Item > & {
+	header: string;
+	getValue: ( args: { item: Item } ) => any;
+	render: ( args: { item: Item } ) => ReactNode;
+};
```
jsnajdr, line 120: I had to do this because the `Required<Pick>` construct doesn't work with the conditional type where `getValue` is defined twice.

youknowriad: I think this is actually nicer and easier to understand

youknowriad (approved):

packages/dataviews/src/types.ts
```diff
@@ -103,17 +97,34 @@ export interface Field< Item extends AnyItem > {
...
+			/**
+			 * Callback used to retrieve the value of the field from the item.
+			 * Defaults to `item[ field.id ]`.
+			 */
+			getValue?: ( args: { item: Item } ) => any;
+	  }
+	: {
+			/**
+			 * Callback used to retrieve the value of the field from the item.
+			 * Defaults to `item[ field.id ]`.
```
sirreal, line 111: Is the description of the default correct here? It looks like the default is `item => item.id`. Reading this documentation, it looks like it's configurable by a field option?

https://github.com/WordPress/gutenberg/blob/89ee50d5953afc5e9651e7aca2e5afa92915326b/packages/dataviews/src/dataviews.tsx#L49

sirreal (approved):
Nice simplification here, thank you. [I did leave one question](https://github.com/WordPress/gutenberg/pull/62856/files#r1654618946) I'd like to understand before this is merged.

packages/dataviews/src/types.ts
sirreal: Oh, never mind. I see I was confusing the id with the value 🤦

jsnajdr: Yes, there is `item.id` and also `field.id` 🙂 I wonder if it would be possible to typecheck that the `Item` really has the field read by the `Field` type. Something like:
```js
type Field<Item extends ItemWithField<FieldName>, FieldName> = {
  id: FieldName;
  getValue: ( item: Item ) => unknown;
}
```
which would fail if the `Item` and `FieldName` types don't match.

PR #62857, opened 2024-06-26
translation should depend on number of items

iworks:
## What?

The translation string should be depend on number of items.

## How?

I added only a string for single item.


Mamaduka (approved):
Thanks for catching this, @iworks!

Mamaduka:
@ellatrix, @vcanales, I'm flagging this for the next RC. This is a new string, and a fix for it can probably go even after "string freeze."

ellatrix:
We're in [hard string freeze](https://make.wordpress.org/polyglots/handbook/glossary/#hard-freeze) though. I can ask about it.

ellatrix:
Backported with #63038

PR #62861, opened 2024-06-26
Use block context to detect presence of parent pattern for overrides

talldan:
## What?
Based on the branch from #62828, tries using block context to detect a parent pattern as an alternative to using selectors and traversing up the block tree.

## Why?
Improved performance.

This also more closely matches the server implementation.

## How?
- Adds a `providesContext` to the pattern block's block.json, providing the `content` attribute
- Adds a default value for the `content` attribute so that it's always truthy (the code can consistently check for a parent pattern using the attribute.
- Adds `usesContext` for the content attribute to all supported block binding blocks
- Replaces the code that traverses the block hierarchy for a simpler `context` check.

## Testing Instructions
Follow the testing instructions in #62828

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -11,7 +11,6 @@ import { addFilter } from '@wordpress/hooks';
  * Internal dependencies
  */
 import { unlock } from '../lock-unlock';
-import { store as blockEditorStore } from '../store';
 
 /** @typedef {import('@wordpress/compose').WPHigherOrderComponent} WPHigherOrderComponent */
 /** @typedef {import('@wordpress/blocks').WPBlockSettings} WPBlockSettings */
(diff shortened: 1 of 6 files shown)


ellatrix (approved):
I guess context makes sense.

packages/block-editor/src/hooks/use-bindings-attributes.js
```diff
@@ -292,6 +280,8 @@ function shimAttributeSource( settings, name ) {
 	return {
 		...settings,
 		edit: withBlockBindingSupport( settings.edit ),
+		// TODO - move this to be located with pattern overrides code.
+		usesContext: [ 'patternOverridesContent', ...settings?.usesContext ],
```
SantosGuillamot: Should we move this to the `usesContext` property when registering the pattern overrides source? https://github.com/WordPress/gutenberg/blob/320dcb9633b2c2fcb5dd6ae99f517ce04e265e9e/lib/compat/wordpress-6.6/block-bindings/pattern-overrides.php#L57

SantosGuillamot:
I agree that using the context makes sense. Thanks for working on that! 🙂 

I can see we have similar logic in the image edit file: [link](https://github.com/WordPress/gutenberg/blob/320dcb9633b2c2fcb5dd6ae99f517ce04e265e9e/packages/block-library/src/image/image.js#L454-L455). Should we use context there as well?

packages/block-editor/src/hooks/use-bindings-attributes.js
talldan: Will this also register `usesContext` for client side code?

Considering both `pattern/overrides` and `patternOverridesContent` are both sharing the same attribute via context, I guess they should be named the same thing. I'm not enamoured with using `pattern/overrides` naming  in js code where the variable is often accessed via dot notation or destructuring.

I think changing it would require a backport.

talldan: > Will this also register usesContext for client side code?

If not, I think it could still be moved to this file if we change it over to the `registerBlockType` hook similar to the one in the `use-binding-attributes.js` file:
https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/hooks/pattern-overrides.js#L110

SantosGuillamot: > Will this also register usesContext for client side code?

Yes, it should work client-side. We might make some changes to how it is handled, but it should work.

I'm not familiar with how that context is used, but it'd be great to use just one property if possible.

> I think changing it would require a backport.

Yes, I think it would need one. But it should be just one line.

talldan: I decided to change it to `pattern/overrides` on the client as well, so that it's consistent and we won't need a backport.

I think I've addressed all the feedback.

SantosGuillamot (approved):
Using block context makes sense to me as well. After a round of testing, everything seems to be working as expected, and automated tests are passing.

PR #62813, opened 2024-06-25
Show publish panel only in edit mode

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/62702

>In the editor, you can use the tab key to focus on the "Open publish panel" button that appears in the bottom right. Clicking this button will show the Publish panel.

This PR fixes that by only rendering this when we're in edit mode, which is the default for post editor.

## Testing Instructions
1. In site editor using the `tab` key you shouldn't be able to reproduce this:

https://github.com/WordPress/gutenberg/assets/54422211/a2352eb5-ba1e-4699-b1f0-d7e297eeb542

2. Using `tab` key in post editor should work as before and show the `open publish panel` button.

Issue #62694, opened 2024-06-20
Unify publishing flow between post and site editor - technical refinements


ntsekouras:
This issue can be considered as the second part of: https://github.com/WordPress/gutenberg/issues/59878 in order to properly fully unify the publishing flow between post and site editors.

Currently the publishing flow seems and is similar in terms of UI, but we still need to have some follow ups to consolidate some functionality in a better way.

The extra use cases in site editor include:

1. Fix duplicate saving panels. This is related to the next item and we need to look into how to best consolidate save panels and button in editor package.
2. Activating/previewing a theme (change of labels, callbacks, etc..). See `customSaveButton` in Editor component.
3. The save button in site editor's sidebar (view mode) renders the save panel in a Modal
4. Some specific logic for how to handle the save when making changes to global styles page
5. Some cleanup is needed and let's refactor a bit where we can, while doing these follow ups ([ex](https://github.com/WordPress/gutenberg/pull/60077#discussion_r1538696144))

Let's track the refinement of the flow and UI in a separate issue to keep the technical improvement separate.

youknowriad:
Anything missing here. Should we close it or are there some quick wins to do still.

ntsekouras:
>Should we close it or are there some quick wins to do still.

No, let's keep it open and I'll see to it soon.

ntsekouras:
Closing this for now, since it's been a while and there were many iterations in code. We should re-evaluate and see what's needed to simplify/improve.

PR #62863, opened 2024-06-26
Editor: Fix duplicate save panels

### LINKED ISSUE #62702: [WP6.6] Publish Panel: Unexpected Behavior in the Site Editor
### Description

Originally reported on Slack: https://wordpress.slack.com/archives/C02QB2JS7/p1718858165581179

In the editor, you can use the tab key to focus on the "Open publish panel" button that appears in the bottom right. Clicking this button will show the Publish panel.

This is expected behavior in the post editor, but may cause unexpected behavior in the site editor.

This issue does not occur in WP6.5.

- Access the site editor.
- Click the tab key repeatedly to move focus to the "Open puglish panel" button.
- Click that button.
- The Publish panel appears.
- When you hover the mouse over the editor canvas, the Publish panel moves inside the canvas.

The following issue occurs whether the canvas is in view or edit mode.

- Open any template.
- Click the tab key repeatedly to move focus to the "Open puglish panel" button.
- Click that button.
- The Publish panel appears.
- URL is undefined.
- Copy button doesn't work.
- The 'View Template' button doesn't work.
- Clicking the 'Add New Template' button shows an error.

https://github.com/WordPress/gutenberg/assets/54422211/a2352eb5-ba1e-4699-b1f0-d7e297eeb542

### Environment info

_No response_

Yes

Yes

ntsekouras:
## What?
Part of: https://github.com/WordPress/gutenberg/issues/62694
Related: https://github.com/WordPress/gutenberg/pull/62813
Also fixes: https://github.com/WordPress/gutenberg/issues/62702

This PR does a couple of things:
1. In site editor we would render two `save panels` and you could `tab` through both of them. Here I render only one and for now I followed the same approach we take for the `customSaveButton`. We pass the custom save panel only when we are previewing a theme and always render the site editor's save panel in sidebar and not in its layout component. This eventually should be handled in the `editor` package probably without needing to pass these extra custom save button and panel.

https://github.com/WordPress/gutenberg/assets/16275880/85df6563-d83f-4781-aa91-38a9030a927d

 

2. Currently the default save panel would be to `open publish panel` and here I changed it to `open save panel` which makes more sense in most cases, since an entity couldn't be `publishable` at all. This fixes [this comment](https://github.com/WordPress/gutenberg/issues/62702#issuecomment-2188441995) too.

https://github.com/WordPress/gutenberg/assets/16275880/bd8b477b-7f12-4d4a-87d2-e8d68e3e3a80

3. Added some extra checks to properly disable the default `open save panel` button.

## Testing Instructions
1. No regression for this PR: https://github.com/WordPress/gutenberg/pull/62813. When in `view` mode in site editor we shouldn't be able to navigate to a save panel using tab.
2. No rendering of two save panels. This can be tested in site editor.
3. The default save panel is `open save panel` and is disabled when needs to.
5. Previewing and activating themes in site editor works as before with any combination of other changes or not.
6. In site editor the `save button` in the left sidebar should work exactly as before and render the save changes modal and save properly

### DIFF
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -37,7 +37,6 @@ import { store as editSiteStore } from '../../store';
 import SiteHub from '../site-hub';
 import ResizableFrame from '../resizable-frame';
 import { unlock } from '../../lock-unlock';
-import SavePanel from '../save-panel';
 import KeyboardShortcutsRegister from '../keyboard-shortcuts/register';
 import KeyboardShortcutsGlobal from '../keyboard-shortcuts/global';
 import { useIsSiteEditorLoading } from './hooks';
@@ -236,8 +235,6 @@ export default function Layout( { route } ) {
 						</div>
 					) }
 				</div>
-
-				<SavePanel />
 			</div>
 		</>
 	);
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -43,6 +43,7 @@ import { useIsSiteEditorLoading } from './hooks';
 import useMovingAnimation from './animation';
 import SidebarContent from '../sidebar';
 import SaveHub from '../save-hub';
+import SavePanel from '../save-panel';
 import useSyncCanvasModeWithURL from '../sync-state-with-url/use-sync-canvas-mode-with-url';
 
 const { useCommands } = unlock( coreCommandsPrivateApis );
@@ -162,6 +163,7 @@ export default function Layout( { route } ) {
 											{ areas.sidebar }
 										</SidebarContent>
 										<SaveHub />
+										<SavePanel />
 									</motion.div>
 								) }
 							</AnimatePresence>
--- a/packages/editor/src/components/editor-interface/index.js
+++ b/packages/editor/src/components/editor-interface/index.js
@@ -55,6 +55,7 @@ export default function EditorInterface( {
 	disableIframe,
 	autoFocus,
 	customSaveButton,
+	customSavePanel,
 	forceDisableBlockTools,
 	title,
 	icon,
@@ -214,7 +215,7 @@ export default function EditorInterface( {
 				)
 			}
 			actions={
-				! isPreviewMode ? (
+				customSavePanel || (
 					<SavePublishPanels
 						closeEntitiesSavedStates={ closeEntitiesSavedStates }
 						isEntitiesSavedStatesOpen={
@@ -225,7 +226,7 @@ export default function EditorInterface( {
 						}
 						forceIsDirtyPublishPanel={ forceIsDirty }
 					/>
-				) : undefined
+				)
 			}
 			shortcuts={ {
 				previous: previousShortcut,
--- a/packages/editor/src/components/editor-interface/index.js
+++ b/packages/editor/src/components/editor-interface/index.js
@@ -215,18 +215,22 @@ export default function EditorInterface( {
 				)
 			}
 			actions={
-				customSavePanel || (
-					<SavePublishPanels
-						closeEntitiesSavedStates={ closeEntitiesSavedStates }
-						isEntitiesSavedStatesOpen={
-							entitiesSavedStatesCallback
-						}
-						setEntitiesSavedStatesCallback={
-							setEntitiesSavedStatesCallback
-						}
-						forceIsDirtyPublishPanel={ forceIsDirty }
-					/>
-				)
+				! isPreviewMode
+					? customSavePanel || (
+							<SavePublishPanels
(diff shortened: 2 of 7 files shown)


test/e2e/specs/editor/various/switch-to-draft.spec.js
```diff
@@ -57,7 +57,9 @@ test.describe( 'Clicking "Switch to draft" on a published/scheduled post/page',
 							.getByRole( 'button', { name: 'Close Settings' } )
 							.click();
 					}
-					await page.getByRole( 'button', { name: 'Save' } ).click();
+					await page
+						.getByRole( 'button', { name: 'Save', exact: true } )
+						.click();
```
Mamaduka: Suggestion (non-blocking): We can also add scope to the locator. Example #62454.

Mamaduka:
@ntsekouras, follow-ing e2e test is still failing:

```
[chromium] › site-editor/multi-entity-saving.spec.js:29:2 › Site Editor - Multi-entity save flow › save flow should work as expected
```

ntsekouras:
>@ntsekouras, follow-ing e2e test is still failing:

Yeah, I updated it.

packages/editor/src/components/editor-interface/index.js
```diff
@@ -214,7 +215,7 @@ export default function EditorInterface( {
 				)
 			}
 			actions={
-				! isPreviewMode ? (
+				customSavePanel || (
```
youknowriad: I think we should keep isPreviewMode check here. I mean if you don't have customSavePanel and you're in preview mode, we shouldn't render the save panel.

packages/edit-site/src/components/layout/index.js
```diff
@@ -236,8 +235,6 @@ export default function Layout( { route } ) {
 						</div>
 					) }
 				</div>
-
-				<SavePanel />
```
youknowriad, line 240: This PR probably removes the save panel from pages that don't render an editor, I think that's probably a bad idea. We should be able to "save" from these pages as well.

ntsekouras:
I had a chat with @youknowriad in private about some of the challenges this PR and the approach here, regarding his previous two comments. We agreed to preserve the `isPreviewMode` in `EditorInterface` that is effective for all editors (previously I had moved this check in site editor's editor) and render site editor's save panel in sidebar instead of its layout component.

packages/edit-site/src/components/layout/index.js
ntsekouras: see: https://github.com/WordPress/gutenberg/pull/62863#issuecomment-2201990675

packages/editor/src/components/editor-interface/index.js
ntsekouras: see: https://github.com/WordPress/gutenberg/pull/62863#issuecomment-2201990675

packages/editor/src/components/editor-interface/index.js
```diff
@@ -214,18 +215,22 @@ export default function EditorInterface( {
...
-							entitiesSavedStatesCallback
-						}
-						setEntitiesSavedStatesCallback={
-							setEntitiesSavedStatesCallback
-						}
-						forceIsDirtyPublishPanel={ forceIsDirty }
-					/>
-				) : undefined
+				! isPreviewMode
+					? customSavePanel || (
```
youknowriad, line 219: I'm guessing these two conditions can be merged using a `&&`

test/e2e/specs/site-editor/multi-entity-saving.spec.js
```diff
@@ -41,7 +41,7 @@ test.describe( 'Site Editor - Multi-entity save flow', () => {
 		).toBeEnabled();
 		await expect(
 			page
-				.getByRole( 'region', { name: 'Save panel' } )
+				.getByRole( 'region', { name: 'Editor publish' } )
```
youknowriad, line 44: Why is this change needed? The region is for saving not just publishing no?

youknowriad (approved):
Works well in my testing.

test/e2e/specs/site-editor/multi-entity-saving.spec.js
ntsekouras: Yeah, but that's what we [call it ](https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/components/editor-interface/index.js#L43)for a while.. Should we change it in a follow up?

packages/editor/src/components/editor-interface/index.js
ntsekouras: They could, but it would be like ` !x && a || b` and because of the `||` I think it's a bit clearer this way..

PR #62864, opened 2024-06-26
Update new release issue template to remove core editor chat item

talldan:
## What?
Removes the item "Attend `#core-editor` meeting (14:00UTC)" from the release issue. The core editor meeting no longer happens, so this step is no longer needed


Mamaduka (approved):
Thanks, @talldan!

PR #62866, opened 2024-06-26
Add SizeControl component

matiasbenedetto:
## What?
Provide a new UI component to define a size.

This PR extracts a part of the FontSizePicker controller and makes it a new independent controller that can be used inside the FontSizePicker controller and as a stand-alone controller.

## Why?
The new SizeControl provides an agnostic UI to define a size that could be used to define not only font sizes but spacings and probably other settings.  

Most of the SizeControl component UI and logic already exist inside the FontSizePicker control, making it impossible to reuse them outside of it. 

This component could be used in the implementation of these two interfaces:
- https://github.com/WordPress/gutenberg/issues/61987
- https://github.com/WordPress/gutenberg/issues/61986

## How?
Extracting the UI and logic from FontSizePicker and making it a stand-alone component, simplifying the logic where possible.

## Testing Instructions
- Run unit tests
- Use the storybook stories of the SizeControl component.

## Screenshots or screencast 
image

See screencasts of the component in use in https://github.com/WordPress/gutenberg/pull/62328

### DIFF
--- a/packages/components/src/font-size-picker/index.tsx
+++ b/packages/components/src/font-size-picker/index.tsx
@@ -13,14 +13,6 @@ import { useState, useMemo, forwardRef } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { Button } from '../button';
-import RangeControl from '../range-control';
-import { Flex, FlexItem } from '../flex';
-import {
-	default as UnitControl,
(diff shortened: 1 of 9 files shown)


mirka:
Related (?): #61180

PR #62869, opened 2024-06-26
CustomSelectControl V2: fix trigger text alignment in RTL languages

ciampo:
## What?

Part of #55023 

Allow the text on the `CustomSelectControlV2` trigger button to follow RTL text alignment.

## Why?

Following user preferences when using the component in a RTL environment

## How?

Using logical properties instead of physical (`text-align: start` instead of `text-align: left`)

## Testing Instructions

- Load storybook examples for `CustomSelectControlV2`
- Via the Storybook top toolbar controls, change the text direction from LTR to RTL
- Observe how the select trigger button label aligns correctly to the right (while on `trunk` is would stay on the left)

## Screenshots or screencast 

|---|---|
| Screenshot 2024-06-26 at 13 29 50 | Screenshot 2024-06-26 at 13 28 46 |


mirka (approved):
Good catch, thanks 🙏

PR #61468, opened 2024-05-08
Edit post: Do not consider sidebars mutually exclusive

ntsekouras:
## What?
This PR removes the handling in post editor that considers the sidebars mutually exclusive. This is now consistent with site editor.

## Testing Instructions
1. In post editor open the instector controls sidebar and the insterter
2. They should both remain opened

PR #62871, opened 2024-06-26
Performance tests: fix for 6.5

ellatrix:
## What?

Makes the locator compatible with both 6.5 and 6.6.

## Why?

This was changed in #61468, but the role for this in 6.5 was not a tab, but a button.

### DIFF
--- a/test/performance/specs/post-editor.spec.js
+++ b/test/performance/specs/post-editor.spec.js
@@ -668,7 +668,9 @@ test.describe( 'Post Editor Performance', () => {
 
 				const startTime = performance.now();
 
+				// This is the WP v6.5 and older locator.
 				const oldLocator = page.getByRole( 'button', { name: 'Test' } );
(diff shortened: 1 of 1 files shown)


test/performance/specs/post-editor.spec.js
```diff
@@ -668,7 +668,10 @@ test.describe( 'Post Editor Performance', () => {
 
 				const startTime = performance.now();
 
-				await page.getByRole( 'tab', { name: 'Test' } ).click();
+				const oldLocator = page.getByRole( 'button', { name: 'Test' } );
+				const newLocator = page.getByRole( 'tab', { name: 'Test' } );
+
+				await oldLocator.or( newLocator ).click();
```
Mamaduka: Let's add an inline comment. It helps with confidence when we might want to remove the old locator.

Mamaduka (approved):

Issue #60360, opened 2024-04-02
ConfirmDialog: Replace 'OK' with more descriptive label


afercia:
### Description

See also https://github.com/WordPress/gutenberg/issues/60027

The `ConfirmDialog` component was introduced in https://github.com/WordPress/gutenberg/pull/34153 because of two main purposes:

Quoting from the related issues, see https://github.com/WordPress/gutenberg/issues/16583 and  https://github.com/WordPress/gutenberg/issues/33937

- The native `window.confirm()` isn't an ideal experience as we're unable to customise the OK button to have a more descriptive label.
- Major browsers - starting with Chrome - will soon restrict the alert/confirm/prompt functions from being used from within cross-origin iframes.

Regarding the first point, it appears some instances of the `ConfirmDialog` just use the default `OK` button text, thus defeating one of the purposes it was introduced for.

I'd tend to think re-enforcing the in-progress action by using a more meaningul button text would make the user experience better. Also, there's some inconsistency throughout the UI. I'd tend to think all `ConfirmDialog` should either use 'OK' oar a more specific text.

Some exampke screenshots:

![Screenshot 2024-04-02 at 11 49 38](https://github.com/WordPress/gutenberg/assets/1682452/8e552b4a-2e01-486c-8333-423e82a09ffc)

![Screenshot 2024-04-02 at 12 12 08](https://github.com/WordPress/gutenberg/assets/1682452/564687e1-8da9-42af-8a24-41a3b9173d3d)

![Screenshot 2024-04-02 at 11 46 16](https://github.com/WordPress/gutenberg/assets/1682452/1f9d95f4-0c53-47b6-8682-ecaf0dfac655)

![Screenshot 2024-04-02 at 11 50 52](https://github.com/WordPress/gutenberg/assets/1682452/7569b4d8-31e5-4a6f-9b67-254d71f0bd4a)

![Screenshot 2024-04-02 at 12 13 20](https://github.com/WordPress/gutenberg/assets/1682452/1d8d5653-b224-4e08-aa07-dcb3eac776cb)

![Screenshot 2024-04-02 at 12 19 38](https://github.com/WordPress/gutenberg/assets/1682452/4f58cbb8-028f-44a6-a8ca-8fc7933da6c5)

![Screenshot 2024-04-02 at 12 21 20 2](https://github.com/WordPress/gutenberg/assets/1682452/890a0ac3-1c75-4d44-8dbf-890c3911f07b)

### Step-by-step reproduction instructions

[...]

Issue #16583, opened 2019-07-15
Consider replacing usages of Window.confirm() with Modal


noisysocks:
We're currently using [`Window.confirm()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) in four places. This isn't an ideal experience as we're unable to customise the OK button to have a more descriptive label.

Moving these instances to use a `Modal` component might provide a better experience. Here's an example of a `Modal` being used this way, from @kjellr's mockup in https://github.com/WordPress/gutenberg/issues/16315#issuecomment-507331101:

<img width="464" alt="Screen Shot 2019-07-15 at 16 03 13" src="https://user-images.githubusercontent.com/612155/61197808-27c94180-a71a-11e9-947a-e7f2b14798fc.png">

---

### 1. Switching a published post to draft

Current:

<img width="458" alt="Screen Shot 2019-07-15 at 15 57 17" src="https://user-images.githubusercontent.com/612155/61197709-bc7f6f80-a719-11e9-8089-256826dece35.png">

Proposed:

```
Switch to Draft

Are you sure you want to unpublish this post?

[ Cancel ] [ Unpublish ]
```

### 2. Switching a published post to private

Current:

<img width="458" alt="Screen Shot 2019-07-15 at 15 57 42" src="https://user-images.githubusercontent.com/612155/61197711-bc7f6f80-a719-11e9-97e8-d2e2e397be9d.png">

Proposed:

```
Privately Publish

Are you sure you want to privately publish this post now?

[ Cancel ] [ Privately Publish ]
```

### 3. Deleting a reusable block 

Current:

<img width="460" alt="Screen Shot 2019-07-15 at 15 58 12" src="https://user-images.githubusercontent.com/612155/61197713-bc7f6f80-a719-11e9-8bd6-0dddfc21bf56.png">

Proposed:

```
Delete Reusable Block

Are you sure you want to delete this Reusable Block? 

It will be permanently removed from all posts and pages that use it.

[ Cancel ] [ Delete ]
```

### 4. Pressing _Reset the template_ when template doesn't match post content

Current:

<img width="462" alt="Screen Shot 2019-07-15 at 15 59 21" src="https://user-images.githubusercontent.com/612155/61197715-bd180600-a719-11e9-8f09-083eba5c5d82.png">

Proposed:

```
Reset Template

Resetting the template may result in loss of content, do you want to continue?

[ Cancel ] [ Reset Template ]
```

kjellr:
[...]

PR #62874, opened 2024-06-26
Align rename modals

jameskoster:
Follow-up to #62788.

The vast majority of rename modals appear in the following format:

Screenshot 2024-06-26 at 12 08 04

* Succinct title
* Visible "Name" label
* "Save" button
* 40px buttons

This PR brings a couple of outliers into alignment with this convention.

### Navigation menus

| Before | After |
| --- | --- |
| Screenshot 2024-06-26 at 12 10 18 | Screenshot 2024-06-26 at 14 04 27 |

Note the visible label. 

### Renaming blocks

| Before | After |
| --- | --- |
| Screenshot 2024-06-26 at 12 09 19 | Screenshot 2024-06-26 at 14 06 10 |
| Screenshot 2024-06-26 at 12 09 30 | Screenshot 2024-06-27 at 11 11 58 |

### Renaming custom views
| Before | After |
| --- | --- |
| Screenshot 2024-06-26 at 12 58 28 |  Screenshot 2024-06-26 at 14 14 39 |

* Changed modal heading
* Button size = 40px
* Changed button label

## Testing Instructions
1. Rename a navigation menu
2. Observe the modal matches the 'After' screenshot above.
3. Rename a block without overrides
4. Observe the modal matches the 'After' screenshot above.
5. Rename a block with overrides
6. Observe the modal matches the 'After' screenshot above (the one with the notice).
7. Enable the custom data views experiment and rename a custom view
8. Observe the modal matches the 'After' screenshot above.

## Additional alignment considerations

* Is the 'Name' label worth displaying above the input, or should it be visually hidden? I can see arguments both ways.
* Should the primary button read "Save" or "Rename". In most confirmation dialogs the primary button repeats the action, so I lean towards the latter, but keen to hear thoughts.


jasmussen:
Nice work.

The only one I don't agree with is the notice change, this one:

> In this case the warning is moved out of the help text into a Notice which seems appropriate given the potential consequences.

I can see sense in using the basic "Name" label for the input field, but I'd keep the help text below, and keep it as plain body text. The consequences are important to understand, but it doesn't need further elevation since you're already in a modal context.

This is also mostly about that particular notice style being generally out of place in the block editor, it doesn't share any DNA with the snackbar, for instance.

t-hamano (commented):
I submitted a commit to fix the e2e tests.

Regarding the notice message, I also feel it's a bit too noticeable.

jameskoster:
Good feedback, I'll change the notice 👍 

How do y'all feel about these questions?

> * Is the 'Name' label worth displaying above the input, or should it be visually hidden? I can see arguments both ways.
> * Should the primary button read "Save" or "Rename". In most confirmation dialogs the primary button repeats the action, so I lean towards the latter, but keen to hear thoughts.

jasmussen:
> Is the 'Name' label worth displaying above the input, or should it be visually hidden? I can see arguments both ways.

No strong opinion.

> Should the primary button read "Save" or "Rename". In most confirmation dialogs the primary button repeats the action, so I lean towards the latter, but keen to hear thoughts.

Save read well to me. It's active language just like Rename, but benefits from the reduced duplication, while still being the action taken.

t-hamano:
> Good feedback, I'll change the notice 👍
> 
> How do y'all feel about these questions?
> 
> > * Is the 'Name' label worth displaying above the input, or should it be visually hidden? I can see arguments both ways.
> > * Should the primary button read "Save" or "Rename". In most confirmation dialogs the primary button repeats the action, so I lean towards the latter, but keen to hear thoughts.

I think accessibility feedback is needed on this.

I remember that @afercia submitted some issues on this point to improve the consistency/accessibility of modals/dialogs. He might have some advice.

afercia:
Generally, visible labels are always better especially with complex forms with several input fields. 
In this case, the form is very simple and there's no ambiguity as the context is clearly about changing the name. However, a visible label doesn't harm and ideally it should be the default pattern for all inputs unless there's a very, very, good reason to make it visually hidden.

Re: the buttons text, I provided some previous feedback but that was specifically about the `ConfirmDialog` component as one of the purposes of its introduction was to be able to provide a more descriptive label than 'Ok'. See https://github.com/WordPress/gutenberg/issues/60360. The `ConfirmDialog` doesn't have a title so it makes sense to use the action for the button label. Instead, these modal dialogs do have a title: 'Rename'. I don't have a strong opinion but probably avoiding duplication is best?

jameskoster:
I lean towards avoiding duplication too. But I am also slightly bothered by the inconsistency here:

Screenshot 2024-06-27 at 15 51 46

Screenshot 2024-06-27 at 15 51 32

t-hamano:
> I lean towards avoiding duplication too. But I am also slightly bothered by the inconsistency here:

These inconsistencies will need to be fixed in the future.

Considering avoiding the "OK" label and avoiding duplicate labels, my options for this PR are "Save", "Update", and "Apply".

Of these, I prefer "Save", which means we'll ship this PR as is.

afercia:
> I lean towards avoiding duplication too. But I am also slightly bothered by the inconsistency here:

Eheh yes that's a good case. I'd love to see some established guidelines added to the documentation of these components. Maybe it is worth creating a follow-up issue?.

For reference, regarding the `ConfirmDialog` component, the desire to have a more descriptive label for the 'OK' button was outlined in the original PR https://github.com/WordPress/gutenberg/issues/16583. That's mentioned in the readme file under the [Best practices](https://github.com/WordPress/gutenberg/blob/82598f94a50d5b34df56b40db770c86f50989b1d/packages/components/src/confirm-dialog/README.md#best-practices) section. It could be improved though.
In the same way, maybe worth considering adding guide lines for the buttons in the Modal readme as well?

jameskoster:
I'm going to flip-flop on this and lean towards repetition 🙊 for two main reasons:

1. Creates consistency (in terms of button labels) with confirmdialog instances/guidance.
2. Functions as a clear convention to follow for actions that don't have an obvious confirmatory action label. 
 
I could be swayed back again, but I'm curious... if not "Duplicate", what would the primary button label be in the `duplicate` modal? "Save", "Apply", "Update" don't quite work for me. The repetition could arguably be seen as useful here because it removes any ambiguity. 

In any case, it's a broader topic, so I'm happy to ship this PR as is, and discuss the labels further in a separate issue/pr.

t-hamano (approved):
> In any case, it's a broader topic, so I'm happy to ship this PR as is, and discuss the labels further in a separate issue/pr.

I agree with this. This PR only includes one button text change, and by unifying everything to "Save", we have improved consistency.

I think the other issues can be addressed by following up.

PR #62876, opened 2024-06-26
DataViews: make `view.hiddenFields` optional

oandregal:
## What?

This PR makes the `view.hiddenFields` config optional.

## Why?

It should not be required to provide an empty `hiddenFields` array it the consumer doesn't have any hidden fields.

## How?

- Update the `hiddenFields` type to optional.
- Make sure all views consider the optional case when they derive the `visibleFields`.
- Remove it from the `DEFAULT_VIEW` config.

## Testing Instructions

Visit "Site Editor > Patterns" and verify everything works as expected.

Before this PR, if the `hiddenFields` property was removed from here (https://github.com/WordPress/gutenberg/pull/62876/commits/468c294f6668d8835b4af652b9a0cfc8c376e3f6) the Patterns page would not load.

### DIFF
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -216,7 +216,7 @@ interface ViewBase {
 	/**
 	 * The hidden fields.
 	 */
-	hiddenFields: string[];
+	hiddenFields?: string[];
 }
 
 export interface ViewTable extends ViewBase {
--- a/packages/dataviews/src/view-table.tsx
+++ b/packages/dataviews/src/view-table.tsx
@@ -223,9 +223,9 @@ const _HeaderMenu = forwardRef( function HeaderMenu< Item extends AnyItem >(
 							onHide( field );
 							onChangeView( {
 								...view,
-								hiddenFields: view.hiddenFields?.concat(
-									field.id
-								),
+								hiddenFields: (
+									view.hiddenFields ?? []
+								).concat( field.id ),
 							} );
 						} }
 					>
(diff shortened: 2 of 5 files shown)


packages/dataviews/src/types.ts
```diff
@@ -216,7 +216,7 @@ interface ViewBase {
 	/**
 	 * The hidden fields.
 	 */
-	hiddenFields: string[];
+	hiddenFields?: string[];
```
youknowriad, line 219: I've been reading our types lately and this one actually stood out to me :P 

I think it makes more sense to have "visibleFields" rather than "hiddenFields" as we move forward. I'm pretty sure most entities will have a lot more fields that what we currently have and an allow list would be better.

Omitting it could have the same effect as today when you omit hiddenFields.

Not a concern for this PR though, just a thought that I had.

packages/dataviews/src/view-table.tsx
```diff
@@ -223,7 +223,7 @@ const _HeaderMenu = forwardRef( function HeaderMenu< Item extends AnyItem >(
 							onHide( field );
 							onChangeView( {
 								...view,
-								hiddenFields: view.hiddenFields.concat(
+								hiddenFields: view.hiddenFields?.concat(
```
youknowriad: This feels wrong? Shouldn't it be `(view.hiddenFields ?? []).concat` ?

oandregal: good catch! 03ef404f206c274e6ea4e36b2aa7808e68963820

packages/dataviews/src/types.ts
oandregal: Yeah, that's a good point: post types have a lot of metadata and if we aim to share the field definition between a view and a form (e.g.: details panel), a lot of the fields would be hidden in the view.

youknowriad (approved):
LGTM 👍

Issue #62020, opened 2024-05-27
Tabs with longer strings may trigger horizontal scrolling of the panels they live in


afercia:
### Description

See related https://github.com/WordPress/gutenberg/issues/62018

As mentioned in See related https://github.com/WordPress/gutenberg/issues/62018, limited width space and horizontal tabs aren't good friends.

When the 'Show button text labels' preference is enabled and WordPress is set to languages other than English, some tabs may use strings that are way longer than the English ones.

The tabs use a CSS flex layout that doesn't allow flex wrap.
The text inside the tabs can wrap to multiple lines but that triggers one more problem, see https://github.com/WordPress/gutenberg/issues/62018

Horizontal scrolling is a problem for usability, readability, and accessibility as some content may be cut-off and not fully readable. Additionally, scrolling in a horizontal direction ro reveal content is a problem for keyboard users.
In general, horizontal scrolling should be prevented. Also, the editor doesn't have (yet) a 'scrollable' component that is fully accessible.

Animated GIF to illustrate the horizontal scrolling with 3 tabs in the Inspector panel and WordPress language set to German:

![hscrolling german](https://github.com/WordPress/gutenberg/assets/1682452/8b5deb7a-62fc-44fe-816c-1840e2ad5794)

The same problem can be reproduced with Dutch and Italian, and I guess with more languages:

![dutch](https://github.com/WordPress/gutenberg/assets/1682452/5ea046e0-1830-4d3e-a105-26ac74662913)

![italian](https://github.com/WordPress/gutenberg/assets/1682452/ca7c6baa-78f9-4874-a532-11e4c88dbbe7)

### Step-by-step reproduction instructions

- Switch WordPress language to German, Dutch, or Italian.
- Go to the post editor and add a navigation block.
- Note: the navigation block i just an example. This problem occurs wherever there are two or three tabs in a narrow panel and the tabs text contaisn long, unbreakable, words..
- Go to Options > Preferences > Accessibility, and enable the 'Show button text labels' preference.
- With the navigation block selected, observe the block inspector panel.
- Move your mouse horizontally over the Inspector panel and observe the panel scrolls horizontally.
- Alternatively, tab through the tabs.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

afercia:
This requires some design feedback first Cc @WordPress/gutenberg-design 

I can think of two options:
[...]

PR #62879, opened 2024-06-26
Tabs: add vertical indicator animation

DaniGuardiola:
## What?

Added an animated vertical tabs indicator (related: https://github.com/WordPress/gutenberg/pull/60560#issuecomment-2126452564). I also removed the style overrides from one instance (block editor preferences dialog) where it was a complete match. Tested other overrides and they look correct to me.

## How?

Using the existing animation system.

## Testing Instructions

Check out the Storybook story for vertical tabs, or instances of vertical tabs in Gutenberg.

Focus it and use up/down arrow keys.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/8862967/ffbe765c-6e3f-4399-b889-cd7c2a7cd993

Preferences dialog - before:

https://github.com/WordPress/gutenberg/assets/8862967/b55740fa-fe6e-4c4b-9226-acb4e293febe

Preferences dialog - after:

https://github.com/WordPress/gutenberg/assets/8862967/cf07b184-29da-4d92-b721-d9a6130b3ac5

With `prefers-reduced-motion`:

https://github.com/WordPress/gutenberg/assets/8862967/1f9a2f52-daa4-43e2-8c6e-7da22c4dcafa

### DIFF
--- a/packages/components/src/tabs/styles.ts
+++ b/packages/components/src/tabs/styles.ts
@@ -36,22 +36,20 @@ export const TabListWrapper = styled.div`
 		outline-offset: -1px;
 	}
 	&:not( [aria-orientation='vertical'] )::after {
-		left: var( --indicator-left );
 		bottom: 0;
+		left: var( --indicator-left );
 		width: var( --indicator-width );
 		height: 0;
 		border-bottom: var( --wp-admin-border-width-focus ) solid
 			${ COLORS.theme.accent };
 	}
 	&[aria-orientation='vertical']::after {
-		/* Temporarily hidden, context: https://github.com/WordPress/gutenberg/pull/60560#issuecomment-2126670072 */
-		opacity: 0;
-
-		right: 0;
+		z-index: -1;
+		left: 0;
+		width: 100%;
 		top: var( --indicator-top );
 		height: var( --indicator-height );
-		border-right: var( --wp-admin-border-width-focus ) solid
-			${ COLORS.theme.accent };
+		background-color: ${ COLORS.theme.gray[ 100 ] };
 	}
 `;
 
--- a/packages/preferences/src/components/preferences-modal-tabs/style.scss
+++ b/packages/preferences/src/components/preferences-modal-tabs/style.scss
@@ -6,31 +6,6 @@ $vertical-tabs-width: 160px;
 	// Aligns button text instead of button box.
 	left: $grid-unit-20;
 	width: $vertical-tabs-width;
-
-	&::after {
-		content: none !important;
-	}
-}
-
-.preferences__tabs-tab {
-	border-radius: $radius-block-ui;
-	font-weight: 400;
-
-	&[aria-selected="true"] {
-		background: $gray-100;
-		box-shadow: none;
-		font-weight: 500;
-	}
-
-	&[role="tab"]:focus:not(:disabled) {
-		box-shadow: inset 0 0 0 var(--wp-admin-border-width-focus) var(--wp-admin-theme-color);
-		// Windows high contrast mode.
-		outline: 2px solid transparent;
-	}
-
-	&:focus-visible::before {
-		content: none;
-	}
 }
 
 .preferences__tabs-tabpanel {


jasmussen (approved):
Impressive amount of red code changes considering what an improvement this is. Works great:

tabs

I checked and validated that this animation respects the operating system set "reduce motion", it does, good.

I don't know how much more of a code review this needs, it works well and appears safe. Nice work!

tyxla:
Nice to see this one!

jasmussen:
@DaniGuardiola Thanks again for this PR. 

It seems like there was an unintended side-effect, however! This Inserter > Patterns list of tabs is apparently also using the tablist component, meaning the gray backdrop is now present behind the light blue style it's meant to have. 

What would be a good way to address this? Should the custom styles for the inserter-usage of TabList be refactored to set a light blue background on the backdrop, instead of on the button item itself? LMK, happy to help.

DaniGuardiola:
Thanks for the heads up @jasmussen. In line with the best practices we're trying to achieve regarding the components package (mostly aimed at reducing custom style overrides that cause breakages and high maintenance costs down the line), I would suggest considering the following in order of preference:

- Whether this UI can use the styles built into the Tabs, and do away with the custom ones.
- If not, whether it makes sense to add what is currently custom styled as a variant to Tabs. This should be thought through and compromised on, as we can't just add a variant for each new design. Future designs should stick to the existing variants moving forward as much as possible, and be mindful that any new addition to the base styles has a high cost.
- If that doesn't work either, it suggests that this is a highly custom, one-off UI that could be addressed with Ariakit directly with custom styles. This comes with a few trade-offs, so ideally we'd settle on one of the above options.

If none of these three options work, we can always address this as a regression and tweak the custom styles as you propose, though I think it's a good chance to address it at a more fundamental level. Of course, that doesn't merit blocking a fix so happy to take the simpler route in the short term. Let me know what you think from a design standpoint :)

jasmussen:
I think there might be a regression to fix for the immediate color disparity, the gray + blue overlay is poor.

However your point is well made. I wonder, would the light-blue style work for the preferences modal @WordPress/gutenberg-design ?

ciampo:
As Dani points out, these regressions are actually a good opportunity to see if we can make our UI more coherent across Gutenberg, avoiding overrides and sticking as much as possible with the default component's look!

jameskoster:
I see no reason for the active tab styles to differ between the Preferences modal and the Patterns Inserter, or any other (vertical) tabs for that matter. I'd love to see more coherence across. The light blue works for me.

Thinking about https://github.com/WordPress/gutenberg/issues/62020#issuecomment-2182740050 I could even go a step further and suggest that the orientation shouldn't affect the styling, but that's a separate discussion.

jasmussen:
Excellent conversation. 

We want the gray backdrop fixed for 6.6 as it's a really poor regression (my bad!). So I wonder: what's the best way to get that fixed? I assume that actually fixing this at the component level is _not_ going to be possible. Another option is to revert the animated backdrop for the 6.6 branch, and fix this for the 6.7 branch. Any thoughts?

ciampo:
If the patterns inserter is the only exception, we could also add some temporary, targeted style overrides that disable the gray backdrop only for that part of the UI (to be cherry-picked for 6.6) and in the meantime agree + implement a universal backdrop style that works for all tabs across the editor (to be merged in `trunk` and included in the next releases)

jasmussen:
Sounds fine if we believe that is okay to backport. We're pretty late in the RC phase, there are special rules. Perhaps @Mamaduka might know?

Mamaduka:
If the bug was introduced during the WP 6.6 release cycle, it could be fixed in RCs, but I don't think there are any more scheduled RC releases. [The target for the WordPress 6.6 release is next Tuesday, July 16, 2024.](https://wordpress.org/news/2024/07/wordpress-6-6-rc3/).

But we can target the next minor release.


I'm unsure if it's related, but vertical tabs animations can be a bit "junky" when navigating pattern categories.

## Screencast

https://github.com/WordPress/gutenberg/assets/240569/dca387b6-3897-4ddf-8c08-5f2610f46181

ciampo:
> vertical tabs animations can be a bit "junky" when navigating pattern categories.

I noticed that too, and my hunch is that the jankiness is caused by the pattern previews being quite intensive to render (and not to the `Tabs` component animation per se)

ciampo:
Disabled the indicator animation in the patterns inserted in this PR #63352

Let's start thinking about the next steps too since ideally we shouldn't be keeping that custom style override around:

- let's agree on indicator styles + animation that can be used universally across the editor
- let's apply those styles to the patterns inserter, instead of using custom indicator styles

DaniGuardiola:
@Mamaduka depending on the technical details of this, we might be able to solve that with a react transition in the right place. Either in the Tabs component, as a generic solution (ideal, but a bit unlikely) or in this specific instance (a bit more likely to work).

DaniGuardiola:
A couple of updates, a PR I'm working on (https://github.com/WordPress/gutenberg/pull/64371) fixes:

- The vertical indicator to be light blue everywhere - https://github.com/WordPress/gutenberg/pull/62879#issuecomment-2218239017 cc @jameskoster @jasmussen @WordPress/gutenberg-design 
- The animation in patterns/media tabs which now works without stuttering (not sure why, but it does work now 😅) - https://github.com/WordPress/gutenberg/pull/62879#issuecomment-2219744969 @Mamaduka @ciampo

PR #62880, opened 2024-06-26
Block bindings: UI for connecting bindings

### LINKED ISSUE #62944: Block Bindings: Add ability to add/remove bindings in attributes panel

Currently, the attributes panel indicated when attributes are connected via block bindings like so:

Screenshot 2024-06-27 at 13 27 39

However, connecting the bindings in this way requires copying and pasting JSON into the code editor — it is not currently possible to add or remove bindings from within the panel UI.

We should include buttons to add and remove bindings within the panel similar to the following:

attributes-panel

cbravobernal:
## What?

Closes #62944

This PR adds the capability of connecting attributes with custom fields via block bindings. Previously, the only way to do it was by updating the post content via the code editor.

The PR uses the `Tool Item` component to show and enable different attributes per block, and `Dropdown components` to select the `post meta` that will be attached.

The block bindings code has been moved to the `editor` package, as it will fetch post meta data, that should not be a dependency in the `block-editor` library. While this feature is per block, it uses general information from the editor.

There is a list of follow ups to this work, but I don't want the PR to get bigger and harder to review:

 - Move `getters` and `setters` to an `useFunction`. It will need to use some hooks for accesing the block store, so moving to an utils folder is not enough.
 - Extract an API to add different sources. Still needs planning, but extenders needs to add their own sources to the list. ACF or Pods are good examples.
 - Avoid custom CSS as much as possible. Tool Item component may have a helper attribute so we avoid adding plain HTMl. Adding an option for items to remove their margins, without having to tweak borders.  Pinging @WordPress/gutenberg-components for feedback about this. 
 - Make this option work on templates (site editor). Right now is only working with real content. Creating layouts with custom fields is a great feature for block bindings.
 - Improve e2e testing. Adding happy paths for removing attributes and adding attributes in `Button`, `Heading` and `Image` blocks.
 - Prevent bindings panel from moving by creating a custom group/slot for it.
 - Review image URL attribute binding. It seems that is not removing correctly from the markup.
 - Change `openDocumentSettingsSidebar` to only open the settings sidebar, and create an `openDocumentStylesSidebar` option. I had to update some tests that were assuming that, if there are no settings for paragraph, the style editor should open by default.
 - Address design feedback.
 - Removing the width CSS to the minimum makes icons get smaller.

## How to test

Register post meta sources.
```
register_meta(
		'post',
		'pokemon_image',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'default'           => 'https://www.pokemon.com/static-assets/content-assets/cms2/img/pokedex/full/006.png',
		)
	);
```

Use the UI to link them to paragraphs, buttons, images and headings.

https://github.com/user-attachments/assets/dfe31ad7-6ade-426d-9f1c-7a31fb5fe5ff

## Screenshot
Screenshot 2024-07-18 at 16 22 11


sc0ttkclark:
I'm worried about long custom field names which can be typical with many people who have large sites with many custom fields. How might those display and how flexible would the UI be for that?

sc0ttkclark:
Oh also is the plan for this UI to eventually accept other registered custom sources where "Custom Fields" is an example of the heading of the source? If so, I think this would work well for other sources in general.

cbravobernal:
> I'm worried about long custom field names which can be typical with many people who have large sites with many custom fields. How might those display and how flexible would the UI be for that?

I'm truncating the content, but happy to try / discover different approaches. I've seen the option of two rows, being the value below the title and with a different color.

Captura de pantalla 2024-07-15 a las 9 16 39
Captura de pantalla 2024-07-15 a las 9 16 27

cbravobernal:
> Oh also is the plan for this UI to eventually accept other registered custom sources where "Custom Fields" is an example of the heading of the source? If so, I think this would work well for other sources in general.

We didn't think yet about how to group different sources, but the idea is to eventually accept other registered custom sources as you mention. Would this UI be enough for them (with a search field)?

gziolo:
@jarekmorawski, @jasmussen, what do you think about the progress on the UI? It's looking close to what you shared in https://github.com/WordPress/gutenberg/discussions/39831#discussioncomment-9895203. In particular:

https://github.com/WordPress/gutenberg/assets/79307566/dbf0927c-95fa-4dbd-bf10-5f76c5c9d1b5

jasmussen:
I'm heading AFK for some summer AFK soon, so won't be able to follow this that closely. If Jarek has the bandwidth to follow this work, I trust him to have to have good instincts, I know he has WooCommerce experience too.

One quick glance:

Screenshot 2024-07-15 at 11 02 47

If a menu opens from the block toolbar, it should have the dark border and not the shadow menu style, just like all other block toolbar dropdowns. I thought this was automatic (CC: @WordPress/gutenberg-components), but otherwise I think it might involve an `isAlternate` prop. I also personally preferred the _plugged_ connection icon to the lightning bolt, though think this UI could survive without an icon at all, since the text is so verbose and there are so many other icons. 

That's a general thought too, start with as little as possible, get a feel for it, then add on top of this only when those things become necessary.

jarekmorawski:
Thanks, folks! I'll do a few more iterations sometime this week and keep you updated on the progress.

tyxla:
> If a menu opens from the block toolbar, it should have the dark border and not the shadow menu style, just like all other block toolbar dropdowns. I thought this was automatic (CC: @WordPress/gutenberg-components), but otherwise I think it might involve an `isAlternate` prop.

This is true AFAIK, the only thing I'd add is that `isAlternate` is deprecated nowadays - you should use `variant="toolbar"` instead.

packages/editor/src/hooks/block-bindings.scss
ciampo, line 1: `.components-*` classnames should be treated as private APIs and should not be used outside of the `@wordpress/components` package, and more in general style overrides to components from `@wordpress/components` should be avoided as much as possible, since they cause our UIs to be incoherent.

packages/editor/src/hooks/block-bindings.js
```diff
@@ -0,0 +1,303 @@
...
+					offset: 259,
+				},
+		  }
+		: {};
+};
+
+function BlockBindingsPanelDropdown( { postMeta, addConnection, attribute } ) {
+	return (
+		<DropdownContentWrapper paddingSize="small">
+			<MenuGroup label={ __( 'Custom Fields' ) }>
```
ciampo: Have you considered using [`DropdownMenu`](https://wordpress.github.io/gutenberg/?path=/docs/components-dropdownmenu--docs) instead? It should also render with the correct borders and shadow variants automatically when used inside a `Toolbar`

ciampo (commented):
> Avoid custom CSS as much as possible. Tool Item component may have a helper attribute so we avoid adding plain HTML. Adding an option for items to remove their margins, without having to tweak borders. Pinging @WordPress/gutenberg-components for feedback about this.

I'd personally need to look more into it to understand fully the request. Pinging also @aaronrobertshaw who originally authored the component.

packages/editor/src/hooks/block-bindings.js
cbravobernal: I think I had issues with it, as it was in https://github.com/WordPress/gutenberg/pull/62880/commits/9445effed690b4ced4b162a70f6d2e2cd16510f4#diff-f9c2cc67b161615938447e7ad71c47ce49a06d58639477d3b9b7551a9fb1c497L96

I'll re-check

aaronrobertshaw:
Thanks for the ping 👍 

> Avoid custom CSS as much as possible. Tool Item component may have a helper attribute so we avoid adding plain HTML. Adding an option for items to remove their margins, without having to tweak borders. Pinging @WordPress/gutenberg-components for feedback about this.

I'm not 100% clear on what is being proposed here. A dedicated issue for any `ToolsPanelItem` updates might be a good path forward.

After a quick glance at the code in this PR though, it appears that the block bindings UI is wishing to replicate the faux item group styling of the color panel. 

The color panel was meant to be a special case but if this is something that will be more common moving forward, perhaps extending `ToolsPanelItem` to have an item group like variant would help avoid [duplicated CSS](https://github.com/WordPress/gutenberg/pull/62880/files#diff-5aea9adc7c3a5c6c20419d686cb52120af415e1cf4daf53cf07a12ebc5dd94c4R24-R54).

As @ciampo [noted](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1678104246) we should also avoid using the component classes directly. This is why the [original CSS](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/colors-gradients/style.scss#L57-L91) that was copied, used a custom class `.block-editor-tools-panel-color-gradient-settings__item`, which you can see [added here](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/colors-gradients/dropdown.js#L47).

artemiomorales (commented):
I did some initial testing and it looks like the truncating wasn't working properly for me in the popover, causing it to span the width of my browser window due to a long field:

big-popover

I'll also note that, when clicking on the paragraph, it felt jarring for the sidebar to default to block settings rather than styles. That may just be because I'm accustomed to the previous behavior:

settings-note

Also, it wasn't immediately apparent to me that I should click on the attribute to finish creating the binding after adding it, and that only the text itself was clickable, not the whole box. Perhaps we can consider adding a stronger call to action or prompt to make this more intuitive?

cta-note

Overall though it was very cool to see this all working 🙌 Will be taking a closer look at the code to see if I can offer additional feedback there.

packages/editor/src/hooks/block-bindings.js
```diff
@@ -0,0 +1,303 @@
...
+					? undefined
+					: newMetadata,
+		} );
+	};
+	const postMeta = useSelect( ( select ) => {
+		return select( editorStore ).getEditedPostAttribute( 'meta' );
+	}, [] );
+	if ( postMeta === undefined || bindableAttributes.length === 0 ) {
+		return null;
+	}
```
SantosGuillamot: Taking a closer look at the code now, I believe it is worth trying moving this to the source definition and just read the sources from here. Basically, I'm suggesting something like this:

```js
	const fieldsList = {};
	Object.values( registeredSources ).forEach(
		( { getFieldsList, label } ) => {
			if ( getFieldsList ) {
				fieldsList[ label ] = getFieldsList( {
					registry,
					context,
				} );
			}
		}
	);

	if ( Object.keys( fieldsList ).length === 0 ) {
		return null;
	}
```

And then, define the `getFieldsList` in each source. For example, in [post meta](https://github.com/WordPress/gutenberg/blob/9a5fcaf9e3dfe43fc3f96a4227c5eac054d9e9a4/packages/editor/src/bindings/post-meta.js), we would add a new property like this:

```js
	getFieldsList( { registry, context } ) {
                 // TODO: Filter footnotes.
		return registry
			.select( coreDataStore )
			.getEditedEntityRecord(
				'postType',
				context?.postType,
				context?.postId
			).meta;
	},
```
This way:
* We could avoid moving the hooks to the editor package. This would be the main reason.
* The API is still private so we can iterate on that.
* It'd be a first step to allow other sources to hook into this list. If we want another source to appear there, they would just need to add that property to the source registration.

Let me know what you think 🙂

cbravobernal:
> I did some initial testing and it looks like the truncating wasn't working properly for me in the popover, causing it to span the width of my browser window due to a long field:
> 

It seems that last CSS removal is forcing the width to be as much as possible. I'll set a fixed width so It can be truncated.

packages/editor/src/hooks/block-bindings.js
cbravobernal: That way you are actually allowing custom sources to list their own fields. Being post meta one of them. Pretty smart!

cbravobernal:
> Also, it wasn't immediately apparent to me that I should click on the attribute to finish creating the binding after adding it, and that only the text itself was clickable, not the whole box.

This was due to the last change removing width from the buttons too. Will fix that.

packages/editor/src/hooks/block-bindings.js
cbravobernal: Can a DropdownMenu v1 have a ToolPanel Item as a trigger or only an icon?

ciampo: > Can a DropdownMenu v1 have a ToolPanel Item as a trigger

Not sure I understand what you mean — `ToolsPanelItem` is a sort of container and should be able to host any control. What result are you trying to achieve?

The trigger is a button, as you can see from the [default Storybook example](https://wordpress.github.io/gutenberg/?path=/docs/components-dropdownmenu--docs). You can add `children` to the trigger button via the `toggleProps.children` prop ([example](https://wordpress.github.io/gutenberg/?path=/story/components-dropdownmenu--default&args=toggleProps.children:Any+react+children+here)). 

If you're rendering the dropdown menu in a `Toolbar`, you may want to use `ToolbarDropdownMenu ` instead (you can see it in action in the [`Toolbar` default Storybook example](https://wordpress.github.io/gutenberg/?path=/docs/components-toolbar--docs))

Storybook examples are the best way to familiarize with a component and its APIs.

ciampo: (I initially misinterpreted the question, but I now updated the initial reply, hopefully it addresses your questions better!)

jarekmorawski:
Here's another iteration of the bindings/connection management UX. I did my best to realign with the core componentry and make Connections feel like a fully native feature. I removed the drag handle from the block toolbar, replaced it with a Connections button (cc @mtias), and redesigned the attribute list.

The quick (3:43) video below explains the design in detail. Let me know what you think!

https://github.com/user-attachments/assets/e6a8bd6f-c82c-4a0b-b5c0-83e9ef2905e3

cc @WordPress/gutenberg-design

richtabor:
@jarekmorawski I still don't think connections should be available as controls in the block toolbar. Only in the Inspector. It's a technical tool, and should be considered as such - not front and center in the toolbar.

jarekmorawski:
The strongest use case for keeping them in toolbar is Bits and how users must be able to insert them inline just like links. We currently show the Link button in the toolbar even when no text is selected so I was thinking Bits could work in a similar way (given they're virtually content-level connections). Related: https://github.com/WordPress/gutenberg/discussions/39831#discussioncomment-9895203

If that's not a consideration, I wouldn't mind removing Connections from the toolbar. We need to indicate in the canvas that a block is connected. In such a state, most block settings aren't editable, so I wonder if we'd show an icon on hover or something.

https://github.com/user-attachments/assets/c5aaf86e-902a-4838-8164-6d160ed8ba90

jarekmorawski:
Alternatively, we'd revisit one of the earlier explorations where the connected blocks had a purple color and a dot. We'd then show the connection information on hover without changing the on-click behavior.

image

paaljoachim:
Thinking out loud...

When a block is unique. I would say that this is a very special and unique block.
What visual hints should be created to show that this is not a regular block but an unique one?
Should it contain another color, different icons in the toolbar, hover etc etc....?
Is there a procedure that exists today in Gutenberg that can be also used here? Or perhaps a standard procedure should be made that can also be applied to other unique blocks when these in the future show up.

richtabor:
> The strongest use case for keeping them in toolbar is Bits and how users must be able to insert them inline just like links. 

I still think they'd be better suited as part of the individual controls, or in the Inspector. At least to start. Perhaps we can explore more later, but I'm hesitant that this UI in the toolbar will make  supported blocks more confusing.

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -1,43 +1,122 @@
...
+	return ! isMobile
+		? {
+				popoverProps: {
+					placement: 'left-start',
+					// For non-mobile, inner sidebar width (248px) - button width (24px) - border (1px) + padding (16px) + spacing (20px)
+					offset: 259,
+				},
+		  }
+		: {};
+};
```
SantosGuillamot, line 52: I can see the `useToolsPanelDropdownMenuProps` is already defined the same way in a couple of places: [Global Styles utils](https://github.com/WordPress/gutenberg/blob/5243817d40e62f064200f53c5090361a9542808f/packages/block-editor/src/components/global-styles/utils.js#L144-L155) and [Block Library utils](https://github.com/WordPress/gutenberg/blob/5243817d40e62f064200f53c5090361a9542808f/packages/block-library/src/utils/hooks.js#L92-L103).

After taking a quick look, it seems `dropdownMenuProps` is initialized using this util. I'm wondering if this should be part of the component, maybe the default or an option. Or have one unique util that can be used everywhere.

It doesn't make too much sense to me to replicate this code in all the places.

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -1,43 +1,122 @@
...
+	attribute,
+	filteredBindings,
+} ) {
+	return (
+		<Item { ...toggleProps }>
+			<VStack spacing={ 0 }>
+				<Truncate>{ attribute }</Truncate>
+				{ !! filteredBindings[ attribute ] && (
+					<Truncate className="block-editor-bindings__item-explanation">
+						{ filteredBindings[ attribute ].args.key }
```
SantosGuillamot: This breaks when the `key` argument is not part of the source. This usually happens when users connect the block to an external source. It should probably try to get the key and, if it doesn't exist, use the source name or label instead.

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -48,43 +127,155 @@ export const BlockBindingsPanel = ( { name, metadata } ) => {
...
+			<ToolsPanel
+				label={ __( 'Attributes' ) }
+				resetAll={ () => {
+					removeAllConnections();
+				} }
+				dropdownMenuProps={ dropdownMenuProps }
+				className="block-editor-bindings__panel"
+				hasInnerWrapper
+				__experimentalFirstVisibleItemClass="first"
+				__experimentalLastVisibleItemClass="last"
```
SantosGuillamot: What is the purpose of these props? Are we doing something different for the first and last items?

packages/block-editor/src/hooks/block-bindings.scss
```diff
@@ -1,3 +1,28 @@
...
+			width: 100%;
+			&.is-open {
+				background-color: $gray-100;
+			}
+		}
+	}
+}
+.block-editor-bindings__popover {
+	width: 360px;
+}
```
SantosGuillamot: Would it make sense to include a property in the popover component that handles this?

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,116 @@
...
+							</VStack>
+						</Item>
+					) ) }
+				</ItemGroup>
+			) ) }
+		</DropdownContentWrapper>
+	);
+}
+
+function BlockBindingsAttribute( {
```
SantosGuillamot: This name feels a bit confusing to me. What about something like `BlockBindingsAttributeConnection` or something similar?

SantosGuillamot (commented):
I did a first review of the pull request focused on the UI and the components used. There are still a few things that need to be fixed before merging, but it is shaping up nicely.

I will continue reviewing the code and testing, but these are my comments so far.

## Proposed changes commited
I made a few commits with some proposals trying to simplify the code a little bit, feel free to revert them.

### Use `ItemGroup` inside `ToolsPanel`

Done in [this commit](https://github.com/WordPress/gutenberg/pull/62880/commits/100bc16eabb182c30dbe68771e49257f042b17b2).

I made this change because it seems the `ItemGroup` component comes with some of the CSS we were implementing manually and it removes the use of a custom `div` wrapper.

Apart from that, I'm using the `VStack` component to create the same effect as previously.

### Use the `Text` component instead of `<p>`
Done in [this commit](https://github.com/WordPress/gutenberg/pull/62880/commits/2f5133ca63d6539463e8dc91687a375e518a8d1d).

Same reasoning as the first point, I'm trying to use core components instead of a custom `<p>` element. It also provides some options to avoid adding a few lines of CSS.

### Use `getBlockBindingsSources` function instead of using the store
Done in [this commit](https://github.com/WordPress/gutenberg/pull/62880/commits/5243817d40e62f064200f53c5090361a9542808f).

Recently, the `getBlockBindingsSources` function was added, so I'm just using it instead of calling the store.

## Other proposed changes
Apart from that, I identified a few things that might need a discussion/fix.

### Show panel in templates, even if not editable
I understand that creating/editing the connections in templates might require a follow-up because it is tricky. However, I believe we should still show the existing connections created through the Code Editor. It is something working in `trunk` and it would be a downgrade.

### Fix panel for sources not using the `key` argument.

Started a thread [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1687839390). We can keep the discussion there.

Basically, right now the panel breaks when a source is not using the `key` argument, which is common for external sources. We should fix that.

### Should we abstract `useToolsPanelDropdownMenuProps`?

Started a thread [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1687837965). We can keep the discussion there.

That util is replicated in different packages and it seems it is always used in the dropdown component. We might want to make it part of the component itself or have a global util.

### Should the popover component accept a width property?

Started a thread [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1687844367). We can keep the discussion there.

Apart from that, I left a few small comments more.

## UI workflows
These are a few UI workflows that might require further discussion as well.

### Show which custom field is connected

Right now, when you open the popover of an attribute that is already connected, it isn't easy to notice which custom field is bound.

Screenshot 2024-07-23 at 13 18 15

We should probably add some visual indicator for that. I believe in the past we explored the possibility of adding a check icon, but maybe we can start simpler with some color adjustments.

### Should we close the popover when selecting a custom field?

_This could be done in a follow-up._

One thing I was wondering is if the popover should be closed when a custom field is selected, which is currently happening. I checked other styles using a similar popover and it is not close on selection, but they all have a way to know which value is selected. Maybe fixing that is enough?

### Should we add a "clear" button to remove the connection?

_This could be done in a follow-up._

I've seen that other styles like duotone have a "clear" button to reset the selection. Should we add something similar?

Screenshot 2024-07-23 at 13 21 45

### Should we add a prompt or helper message when an attribute is not connected?

Right now, when an attribute is added to the panel but it isn't connected yet, it doesn't show anything. It feels to me it could benefit from having a prompt or a helper text.

Screenshot 2024-07-23 at 13 24 16

### Revisit the labels and helpers in the different scenarios
Now that we are adding the possibility of creating the connections, I think we might want to rethink the labels and helpers we are using.

For example, when there are not attributes connected, the helper still says _"Attributes connected to various sources"_. Is this correct? Is the "attributes" label even clear?

Screenshot 2024-07-23 at 13 26 22

packages/block-editor/src/hooks/block-bindings.js
cbravobernal: At the beginning, they were used to fit the borders, but with just CSS is enough. We can delete them.

packages/block-editor/src/hooks/block-bindings.scss
cbravobernal: I think it would. But is more a components / design decision.

packages/block-editor/src/hooks/block-bindings.js
cbravobernal: BlockBindingItem maybe? 

I find your option quite verbose. It is true that is each block attribute that can be connected but... 🤔 

It's not a strong opinion, feel free to update it.

packages/block-editor/src/hooks/block-bindings.js
cbravobernal: I agree with that, but, at the same time, we would need to allow this option to be extensible and editable for external developers.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: What about `AttributeBinding` or `BlockAttributeBinding`? In the end, this component takes care of showing the connection of a specific attribute, right?

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I've removed them in [this commit](https://github.com/WordPress/gutenberg/pull/62880/commits/68611fc200212ccda9428ce9b254bb8eabdf4e1f).

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I modified the logic in this comment to support the mentioned use cases: [link](https://github.com/WordPress/gutenberg/pull/62880/commits/5600242c51e29a73180f0da50d1ac70428dc603c).

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: Pinging @WordPress/gutenberg-components because they will probably have more context if this makes sense or feasible.

SantosGuillamot:
In the [latest commit](https://github.com/WordPress/gutenberg/pull/62880/commits/b9eff6a9998d301bebc8ac2b2fff43fe6157ff16), I experimented with using `DropdownMenuV2` experimental component because it could help with some aspects:
* Less custom CSS is needed.
* It provides functionalities like adding a checkbox or hide on click by default.

Screenshot 2024-07-23 at 18 41 40

As mentioned, this component is experimental, so we need to discuss if the benefits are enough reason to use it.

tyxla:
> In the [latest commit](https://github.com/WordPress/gutenberg/pull/62880/commits/b9eff6a9998d301bebc8ac2b2fff43fe6157ff16), I experimented with using `DropdownMenuV2` experimental component because it could help with some aspects:

cc @ciampo - it could be a good one to keep an eye on since `DropdownMenuV2` is still actively being developed and subject to rapid changes.

packages/block-editor/src/hooks/block-bindings.js
ciampo: Thank you for the ping!

Those popover settings (placement and offset) are related to the layout of the block editor and the block  inspector sidebar. The `@wordpress/components` package needs to provide UI components to more consumers than just the block editor, and therefore it should not be aware of those details otherwise it would become tightly coupled with the editor.

I agree though that it would be best to store those settings in a common place.

Since those settings are related to the sidebar, I would suggest storing them in a package that is a common dependency of the block editor, but not a generic package (like components, data, compose, gets). 

Maybe in the `@wordpress/interface` package (where complementary area, the component used for the sidebar, [is defined](https://github.com/WordPress/gutenberg/blob/61659a34f7af204c20860b207a04963439ae6842/packages/interface/src/components/complementary-area/index.js#L45)), or even in the `@wordpress/editor` package itself?

ciampo:
As @tyxla said, please be aware that `DropdownMenuV2` is still actively being worked on and its APIs / behavior / styles may change in the upcoming weeks.

Also, if I understand correctly the work done in this PR, I'm not sure that using checkbox menu items is the correct choice, since checkboxes semantically indicate a list of options that can be all independently checked or unchecked. I would go for either radio menu items (only one option can be selected at a single time, and a selected option can't be unchecked), or for standard menu items (clicking an item performs an action).

SantosGuillamot:
> please be aware that DropdownMenuV2 is still actively being worked on and its APIs / behavior / styles may change in the upcoming weeks.

It seems using it would help with some use cases like the radio items or automatically truncate items. It reduces the custom CSS needed. I guess the only option is using other components and replicate the missing functionalities ourselves, but doesn't seem ideal. For example, is it possible right now to get the "radio" experience with other components without adding too much custom code?

> I would go for either radio menu items (only one option can be selected at a single time, and a selected option can't be unchecked), or for standard menu items (clicking an item performs an action).

If we decide to keep using `DropdownMenuV2` I can move it to radio menu items. Is it possible to change the icon to have the same one as the checkbox?

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,164 @@
...
+					<DropdownMenuGroup key={ label }>
+						{ Object.keys( fieldsList ).length > 1 && (
+							<Text
+								className="block-editor-bindings__source-label"
+								upperCase
+								variant="muted"
+							>
+								{ label }
+							</Text>
+						) }
```
ciampo: We currently don't offer dropdown menu group labels because we initially weren't sure we would want folks to use them, but it looks like you may need them here.

cc @WordPress/gutenberg-design we should probably decide if we want to allow for group labels in the component (in that case I can prepare an official dropdown menu component for that), or if that's a design pattern that should be avoided

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,164 @@
...
+							<DropdownMenuCheckboxItem
+								key={ key }
+								onClick={ () =>
+									addConnection( key, attribute )
+								}
+								name={ attribute + '-binding' }
+								value={ key }
+								checked={ key === currentKey }
+							>
+								<DropdownMenuItemLabel numberOfLines={ 1 }>
```
ciampo: Even if it currently works, please do not use the `numberOfLines` prop. It's currently not part of `DropdownMenuItemLabel` and `DropdownMenuItemHelpText` public APIs and it could lead to unexpected behaviour / regressions in case we update the component in the future.

You won't get TypeScript autocompletion at the moment because of `lock/unlock` not supporting it, but you can have a look at the public API of `DropdownMenuV2` and its subcomponents in the [Storybook docs](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs)

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,164 @@
...
+				<Button
+					__next40pxDefaultSize
+					disabled={ ! binding }
+					accessibleWhenDisabled
+					onClick={ () => removeConnection( attribute ) }
+					variant="tertiary"
+				>
+					{ __( 'Clear' ) }
+				</Button>
+			</ButtonGroup>
```
ciampo: This is not semantically valid markup — from the [spec](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/), children of `role="menu"` should have role `menu` (for nested menus), `group`, `menuitem`, `menuitemcheckbox`, or `menuitemradio`.

If you want to clear the selection, I recommend using a `DropdownMenuItem`.

Also, I believe that, when using `ToolsPanel` like in this case, clearing/resetting the value of a control usually happens via the ellipses menu in the top-right of the panel. Controls rendered in `ToolsPanel` usually don't provide their own reset UI.

packages/block-editor/src/hooks/block-bindings.js
jameskoster: If a genuine need exists then I think it's fine to add. What I'm most concerned about is potential misuse leading to a proliferation of labels in menus. For instance a menu containing a single group probably shouldn't use a label/heading. Is there any way to build that constraint into the component?

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,164 @@
...
+					<DropdownMenuItemLabel numberOfLines={ 1 }>
+						{ attribute }
+					</DropdownMenuItemLabel>
+					{ !! binding && (
+						<DropdownMenuItemHelpText
+							numberOfLines={ 1 }
+							className="block-editor-bindings__item-explanation"
+						>
+							{ args?.key || sourceProps?.label || sourceName }
+						</DropdownMenuItemHelpText>
```
ciampo: While this seems to work ok, please do not use `DropdownMenuItem*` subcomponents inside other components (like `Item`). DropdownMenu subcomponents are meant to be used together with other DropdownMenu subcomponents.

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,164 @@
...
+	fieldsList,
+	addConnection,
+	removeConnection,
+	attribute,
+	binding,
+} ) {
+	const currentKey = binding?.args?.key;
+	return (
+		<>
+			{ Object.entries( fieldsList ).map( ( [ label, fields ] ) => (
```
ciampo: This `map` is currently generating a React warning about each child in a list needing to have its own unique `key`

ciampo:
@SantosGuillamot It's probably ok if you use `DropdownMenuV2` for now, but please stick to the documented APIs ([example](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1689891761)), and don't add style overrides or any other custom behavior ([example](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1689906728)) because it would make it much harder for us to iterate on the component.

> If we decide to keep using DropdownMenuV2 I can move it to radio menu items. Is it possible to change the icon to have the same one as the checkbox?

Not as of today. The component is designed to use two separate icons for checkbox and radio items to help the user understand the type of UI they're interacting with (in the same way `input[type="checkbox"]` and `input[type="radio"]` look different in the browser). Let's ping @WordPress/gutenberg-design in case they have thoughts on this specific aspect, though.

jameskoster:
> Is it possible to change the icon to have the same one as the checkbox?

I wouldn't advise using a check icon in a radio item for the same reasons Marco already mentioned.

packages/block-editor/src/hooks/block-bindings.js
ciampo: We could try to run a check at runtime, collect that sort of info, and output a warning to the console in case (only in dev/test mode). But I'd be doing that as a last resort, only if actually see folks abusing of groups and group labels.

jameskoster: Adding guidelines in storybook could also help, if possible.

SantosGuillamot:
Thanks a lot for all the feedback. Really appreciated ☺️ I'll go through the comments, address the requested changes, and check if something would be missing or we can still use the current components.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: It would be nice to have it for this use case. Is it a problem the way it is created right now? Can we use this until an official component is created?

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: > Even if it currently works, please do not use the numberOfLines prop.

Is it possible to truncate on the first line? I couldn't find another way of doing that. Is it something you consider adding in the future?

It could be done using something like [this](https://github.com/WordPress/gutenberg/pull/62880/files#diff-bff98e322c4269c50aa13e4b251beb9d53d2d0c3d4066406f735d2d49bed83a1R107-R114) and not using `DropdownMenuItemLabel`, but doesn't seem ideal.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I can remove the clear button for now. I was just experimenting how it could look like.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: Okay. I guess we need to use `Truncate` and `VStack` for this use case then.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: By the way, the "Clear" button is added in other places where the `ToolsPanel` is used. Some examples could be the paragraph colors or the image duotone.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: Okay, thanks for the context! We can figure out the best place for this in a follow-up pull request I believe.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I changed it to use `VStack` and `Truncate` instead: [link](https://github.com/WordPress/gutenberg/pull/62880/files#diff-bff98e322c4269c50aa13e4b251beb9d53d2d0c3d4066406f735d2d49bed83a1R104-R116). Hope that's okay.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I modified it to have just a wrapper component with a key that should be unique. I believe it should solve the problem, but I can't confirm because I wasn't able to reproduce it.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I removed it. We can consider adding it properly in a follow-up pull request.

gziolo:
@ciampo, what is the recommended resource to learn in depth about best practices for composing existing UI components to build accessible interfaces? I had similar dillemas like in this PR when working on the custom block.

packages/block-editor/src/hooks/block-bindings.js
artemiomorales: How about `AttributeBinder`? This is the component that actually creates the binding, so I think a name that clearly signals this would be helpful for anyone looking at this file.

Alternative names: `Binder`, `BindingController`.

packages/block-editor/src/hooks/block-bindings.js
ciampo: It should be ok to use a custom implementation for now, but probably best to add `aria-hidden="true"`

packages/block-editor/src/hooks/block-bindings.js
ciampo: > Is it possible to truncate on the first line? I couldn't find another way of doing that.

It is not at the moment

> Is it something you consider adding in the future?

Possibly! I guess we want to find the sweet spot between giving consumers of the components flexibility, while also ensuring a certain level of "art direction" on how the component behaves.

@WordPress/gutenberg-design do you have any guidelines on text truncation for both label and help text in the dropdown menu?

packages/block-editor/src/hooks/block-bindings.js
ciampo: > By the way, the "Clear" button is added in other places where the ToolsPanel is used. Some examples could be the paragraph colors or the image duotone.

I wasn't aware of that — I always assumed that `ToolsPanel` would take care of resetting controls via the ellipsis menu, instead of having the individual controls provide a "clear" button, but I may be out of date here (cc @aaronrobertshaw ).

In case you wanted to add a "Clear" button, my advice of using a separator + a standard dropdown menu item still stands.

packages/block-editor/src/hooks/block-bindings.js
ciampo: It should be fine, although I haven't tested those changes on my machine

packages/block-editor/src/hooks/block-bindings.js
ciampo: Thank you! The way I reproduced it was to open the dropdown menu. The browser console would show an error for me.

Also, looking at the code, I see that you're rendering the separator as a child of the menu group. The separator should be rendered outside of the menu group (see [Storybook example](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs))

ciampo:
> @ciampo, what is the recommended resource to learn in depth about best practices for composing existing UI components to build accessible interfaces? I had similar dillemas like in this PR when working on the custom block.

I guess there are multiple sources, I'll try to list a few sources / best practices:

- for the accessibility part, the [WAI ARIA APG for patterns](https://www.w3.org/WAI/ARIA/apg/patterns/) is a great place to learn about semantics and specs for the most common UI patterns. For `DropdownMenuV2` specifically, the relevant spec is [the `menu` one](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)
- I also quite like Heydon Pickering's work like [inclusive components](https://inclusive-components.design/).
- for our components, Storybook is a great place to learn about how to use a given component and their public API (ie. props) ([example for `DropdownMenuV2`](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs))
- especially for this new "wave" of ariakit-based components, the [ariakit docs](https://ariakit.org/guide/getting-started) can be also useful in learning a few core concepts
- @DaniGuardiola has also a few articles on his blog about component API design and composition (like [this one](https://dio.la/article/the-everything-bagel-of-components) and [this one](https://dio.la/article/the-open-closed-component-part-1))

Hopefully, this can be a good start, but please do ask more questions if you have more!

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,191 @@
...
+							upperCase
+							variant="muted"
+						>
+							{ label }
+						</Text>
+					) }
+					{ Object.entries( fields ).map( ( [ key, value ] ) => (
+						<DropdownMenuRadioItem
+							key={ key }
+							onClick={ () => addConnection( key, attribute ) }
```
ciampo: Nit: When working with DropdownMenuRadioItems, using `onChange` is preferred with respect to `onClick`  as it relates to the radio item's value changing (and not the specific user interaction)

```suggestion
							onChange={ (newKey) => addConnection( newKey, attribute ) }
```

But using `onClick` as we do currently should also work.

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,191 @@
...
+							name={ attribute + '-binding' }
+							value={ key }
+							checked={ key === currentKey }
+						>
+							<DropdownMenuItemLabel numberOfLines={ 1 }>
+								{ key }
+							</DropdownMenuItemLabel>
+							<DropdownMenuItemHelpText numberOfLines={ 1 }>
+								{ value }
+							</DropdownMenuItemHelpText>
```
ciampo: Similar feedback as in https://github.com/WordPress/gutenberg/pull/62880#discussion_r1689891761 re. `numberOfLines`

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,191 @@
...
+				<Item key={ attribute }>
+					<BlockBindingsAttribute
+						attribute={ attribute }
+						binding={ binding }
+					/>
+				</Item>
+			) ) }
+		</>
+	);
+}
```
artemiomorales, line 138: Should we add a visual indicator to let users know when the panel is read only? Otherwise the panel may just appear broken in those cases. I imagine this would be especially useful when a user doesn't have sufficient permissions to modify the bindings.

We could just add a gray background to the box. What do you think?

This PR is already big, so we could also handle that in a separate PR.

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,191 @@
...
+		<VStack>
+			<Truncate>{ attribute }</Truncate>
+			{ !! binding && (
+				<Text
+					variant="muted"
+					className="block-editor-bindings__item-explanation"
+				>
+					<Truncate>
+						{ args?.key || sourceProps?.label || sourceName }
+					</Truncate>
```
artemiomorales, line 118: When we extract an API to add different sources, we should maybe consider how to let sources intentionally define a value here other than `key`, that way they could provide additional context if needed.

packages/block-editor/src/hooks/block-bindings.js
artemiomorales: Done 👍

packages/block-editor/src/hooks/block-bindings.js
artemiomorales: It looks like the `DropdownMenuItemLabel` gets truncated at one line even when `numberOfLines` is undefined, so at least as far as this PR is concerned, it seems like truncation is only an issue with the `DropdownMenuItemHelpText`:

truncation

packages/block-editor/src/hooks/block-bindings.js
artemiomorales: I added `aria-hidden` to the group heading 👍

artemiomorales (commented):
I finally had a chance to take a look at the code and raised a couple of items to consider below, most likely in separate PRs as follow ups. In summary:

- What should the style of the read only panel be?
- Should we allow sources to define a value other than `key` to show in the attributes panel?

I see there is a pending discussion on how we should [truncate text](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1690539170), which would also affect [this comment](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1690591076).

Most of the other discussions on this PR can likely be resolved, though I'll leave that to @SantosGuillamot or others who have been participating in them.

Once we address the truncation and failing tests, I think this PR will good to go as a first iteration 🙌

artemiomorales:
I just pushed a commit to address the failing tests.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: The main issue I see with truncating to two lines in this use case is that it creates items with different widths depending on the size of the value, which feels a bit unusual.

packages/block-editor/src/hooks/block-bindings.scss
```diff
@@ -1,3 +1,14 @@
-.components-panel__block-bindings-panel .components-item__block-bindings-source {
-	color: $gray-700;
+div.block-editor-bindings__panel {
+	grid-template-columns: auto;
```
SantosGuillamot, line 2: For this use case, it seems we need to use just 1 column in the [`ToolsPanel` component](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-toolspanel--docs), but it seems it always uses 2 columns. @WordPress/gutenberg-components did you consider adding a new property to define the number of grid columns? This is something the [Grid component](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-grid--docs) is doing, for example.

Apart from that, I had to add more specificity to override those styles.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I agree the UI for read only needs to be revisited, and I also agree it can be done in a follow-up.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: Yes, we need to explore how other sources will hook into this. I don't know yet if using a key should be enforced or not and how much flexibility we should provide in this specific UI. They can always create their own UI for their use case if it defers from the Custom Fields one.

We can discuss that in a follow-up before opening the APIs.

gziolo: In my opinion, `key` alone makes sense only as long we allow only `Post Meta` here. We should default to the name of the source, and show more details on subsequent action like hover. The alternative might be to offer an optional way for sources to define how the label is structured. There is something like that for block types which never got out of an experiment, example:

https://github.com/WordPress/gutenberg/blob/3dfdfef3eb5cf35abf67bdce9013cc8cef5927b7/packages/block-library/src/heading/index.js#L30-L56

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I was trying to remove the `numberOfLines` property and it seems there is a problem when I have a long string without spaces (it could be a URL, for example).

**With `numberOfLines` set to 1** 

Screenshot 2024-07-25 at 11 57 56

**Without `numberOfLines`**

Screenshot 2024-07-25 at 11 55 35

---
Would be okay to keep it and figure out how to solve the problem in a follow-up pull request?

tyxla: We spotted this during testing yesterday, and @ciampo filed #63916, which should solve that exact problem.

gziolo: It looks like it got solved as you typed. That was a quick one! :)

tyxla: Yep, it was just merged - feel free to rebase and try it out!

gziolo:
> > @ciampo, what is the recommended resource to learn in depth about best practices for composing existing UI components to build accessible interfaces? I had similar dillemas like in this PR when working on the custom block.
>
> I guess there are multiple sources, I'll try to list a few sources / best practices:

I was thinking specifically about `@wordpress/components`. So far I was using:
- [Strorybook](https://wordpress.github.io/gutenberg/?path=/docs/components-introduction--docs) - which has great guidelines embedded for how to contribute to the components library
- [@wordpress/components package reference](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-components/) is another resource that has the individual components documented

What I couldn't locate was the guide on how to assemble existing components and what to do with these marked as experimental. In general, how to think about the system holistically. I can't find the communication on Make Core related to the vision set in the past, but that article from WP Tavern seems relevant: https://wptavern.com/g2-components-a-from-scratch-reimagining-of-wordpress-components. So, to clarify my question is, if I want to contribute to WordPress core, Gutenberg or extend it, how do I consume these components so I optimize for flexibility and avoid as much as possible writing custom CSS. I'm happy to open an issue, if such resource doesn't exist today, as I needed something like that to accelerate the process of transforming Figma designs into React components.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: Great! I just rebased and the problem is not there anymore. Thank you! 🙂 I removed the `numberOfLines` prop until we figure out if it should be accepted as a parameter.

SantosGuillamot:
After making a few changes and addressing the existing feedback, I believe this is ready. Although I would appreciate another review as I made a bunch of modifications to the code.

This is a quick summary:

## Initial prototype
Here you have a demo of the workflows:

https://github.com/user-attachments/assets/a366cf1e-035f-48ed-b70c-addb78fdd520

We have to keep in mind that this is the first iteration of the UI, we will keep improving it in future pull requests. I believe it is much better than the current experience, use the Code Editor, so as long as the code is okay and we aren't breaking anything, it should be fine.

Apart from that, another important aspects:
* It doesn't show in pattern overrides.
* It doesn't show in blocks not supported by block bindings.
* It shows a "read-only" mode in scenarios where user can't connect the custom fields. For example, in templates until that's supported in a follow-up PR.

Screenshot 2024-07-25 at 13 18 06

* It keeps other metadata properties like "name".

## UI concerns
@WordPress/gutenberg-design It'd be great to have a final review from your side to ensure this is good enough as an initial prototype, aside from the follow-ups mentioned below and other improvements. There are a couple of aspects that would need special attention:

* **Is it okay to show the settings by default when clicking a paragraph or heading?**

With this implementation, we are adding a tools panel for every paragraph or heading to let users create bindings. That means that the "Settings" tab is always filled, showing by default. It seems that in all blocks that's the order of the tabs.

Previously, for normal paragraphs, it showed the "Styles" tab.

| Before | After |
|----------|----------|
| Image 1 | Image 2 |

* **Should we change the titles and description before merging?**

I am not convinced about the used titles or descriptions in the panel. Should they be revisited before merging or addressed in a follow-up?

## Follow-ups
I'm trying to gather the different mentioned follow-ups:

### Keep iterating on the UI
We will need to keep iterating on the UI to add improvements suggested in the [UI issue](https://github.com/WordPress/gutenberg/issues/62944). One good example could be adding a search bar, or if we should add a clear button as discussed [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1689900649).

### Allow connecting custom fields in templates
Right now, templates have a "read-only" mode because there is no post ID to retrieve the list of available custom fields. We need to explore how to handle this use case. Creating block bindings in templates is an important use case.

### Improve UI of "read-only" panel
Even if we make it work in templates, there would be cases where the "read-only" panel could be needed. For example, when the user doesn't have permissions. As suggested [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1690691944), it could help clarifying that it is in "read-only" mode. 

### Move functions to remove/add bindings to global utils

The idea is to convert [these functions](https://github.com/WordPress/gutenberg/pull/62880/files#diff-bff98e322c4269c50aa13e4b251beb9d53d2d0c3d4066406f735d2d49bed83a1R215-R261) into utils so they can be used outside these components. This would help other sources, like pattern overrides or external sources, to create their own UI. Additionally, we wouldn't need to pass them as props as we are doing right now.

A similar use case could be the useGradient hook: [link](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/gradients/use-gradient.js#L57).

### Revisit the APIs to let extenders hook into this UI
Before making the APIs public, we need to polish the workflows to ensure they work for the different use cases. It will require some discussions as [this one](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1690701965).

### Should we let users/developer decide which fields should be connectable?
In some cases, there are some custom fields that are only used for internal purposes. Right now, we are showing all the fields that are exposed in the REST API. Is that enough or should we reduce it even more? This is a discussion we need to have at some point.

### Abstract the `useToolsPanelDropdownMenuProps` util
As discussed [here](https://github.com/WordPress/gutenberg/pull/62880/files#r1687837965), that util has been created in multiple places with exactly the same code, and it seems it is the way to go for using the ToolsPanel in the editor. We should consider having a unique util and import it from the different places.

### Sync the edit values
It'd be great to ensure that if I modify a custom field value in the store, it also gets updated in the popover triggered by the panel created. If I am not mistaken, this is not happening right now.

### Explore exposing a label for meta fields and block attributes
Right now, the keys seem too technical, but there are no "user-friendly" alternatives. It'd be great to explore the possibility of letting developers define this while creating a custom field or a block attribute.

---
@cbravobernal I wasn't able to reproduce the following points. Could you confirm if they have been solved after the latest changes or they need a follow-up?

* Prevent bindings panel from moving by creating a custom group/slot for it.
* Review image URL attribute binding. It seems that it is not removed correctly from the markup.

## Follow-ups related to components
@WordPress/gutenberg-components This pull request triggered a bunch of questions related to `@wordpress/components` used. We can moved them to independent issues if needed.

### Should the `DropdownMenuGroup` accept a label?
As discussed [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1689884904), we are adding labels manually for this component. Would it make sense to integrate it?

### Should the `DropdownMenuItemHelpText` component accept `numberOfLines`?
As discussed [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1689891761), it is not possible to define the desired numberOfLines where the help text should truncate.

### Should `ToolsPanel` component accept columns prop?
As shared [here](https://github.com/WordPress/gutenberg/pull/62880#discussion_r1691111191), in this use case we wanted to use just one grid column instead of the default 2, and it created some issues. Should this be supported by the ToolsPanel component directly?

### Should `ToolsPanel` component accept a help prop?
If I am not mistaken, other components like BaseControl have a property to automatically add a help text to the panel. In this pull request, we added it manually [here](https://github.com/WordPress/gutenberg/pull/62880/files#diff-bff98e322c4269c50aa13e4b251beb9d53d2d0c3d4066406f735d2d49bed83a1R320-R322). Should this be a prop of the `ToolsPanel` or adding it manually is enough? Developers could handle it differently.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I would say it is not exactly the component that creates the binding. It just shows the attribute and where it is connected to. It is reused in "read-only" mode as well. I would say the one controlling the creation is the Dropdown.

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I moved it outside of the dropdown menu in [this commit](https://github.com/WordPress/gutenberg/pull/62880/commits/d14d50feb969794728c7bc6244c3af33e18b3a06).

packages/block-editor/src/hooks/block-bindings.js
```diff
@@ -2,42 +2,199 @@
...
+								<DropdownMenuItemLabel>
+									{ key }
+								</DropdownMenuItemLabel>
+								<DropdownMenuItemHelpText>
+									{ value }
+								</DropdownMenuItemHelpText>
+							</DropdownMenuRadioItem>
+						) ) }
+					</DropdownMenuGroup>
+					{ i !== Object.keys( fieldsList ).length - 1 && (
```
gziolo, line 95: This separator always was problematic as it rarely should be rendered after the last element. I dream of the future when there is a prop on the menu to automatically injects the separator when there are groups defined 😄

packages/block-editor/src/hooks/block-bindings.js
gziolo: > Maybe in the @wordpress/interface package (where complementary area, the component used for the sidebar, [is defined](https://github.com/WordPress/gutenberg/blob/61659a34f7af204c20860b207a04963439ae6842/packages/interface/src/components/complementary-area/index.js#L45)), or even in the @wordpress/editor package itself?

To my knowledge, there is intent to sunset `@wordpress/interface` in favor of `@wordpress/editor`.

cbravobernal:
> @cbravobernal I wasn't able to reproduce the following points. Could you confirm if they have been solved after the latest changes or they need a follow-up?
> Prevent bindings panel from moving by creating a custom group/slot for it.
> Review image URL attribute binding. It seems that it is not removed correctly from the markup.

 - Prevent bindings panel from moving by creating a custom group/slot for it.
 
 This one is fixed by keeping the hooks inside `block-editor` package function.

- Review image URL attribute binding. It seems that it is not removed correctly from the markup.

This one keeps failing. The steps are, upload an image, set a custom field for url, then replace with a different image from the library. Happy to review with you in a sync call.


https://github.com/user-attachments/assets/1fa20916-c20c-47de-bbd5-d266e83c2e7d

SantosGuillamot:
> This one keeps failing. The steps are, upload an image, set a custom field for url, then replace with a different image from the library. Happy to review with you in a sync call.

Thanks a lot for the video 🙂 I am not able to reproduce it in my local environment. Maybe it is related to some image properties? Could you test if it also happens in `trunk` or if it is related to this PR?

https://github.com/user-attachments/assets/93d80054-9648-40d8-86e4-88ced56fbcd7

artemiomorales:
> Thanks a lot for the video 🙂 I am not able to reproduce it in my local environment. Maybe it is related to some image properties? Could you test if it also happens in trunk or if it is related to this PR?

I was able to reproduce Carlos's issue just once. After that, it started working as expected.

I've tried many different combinations of going back and forth between custom fields, changing images, etc., but I'm unable to reproduce it again.

artemiomorales (approved):
I think this PR is in a good state for a first iteration — I'll go ahead and approve. We can continue discussing and keeping in sync with the components as they evolve and block bindings development progresses.

I haven't been able to reproduce Carlos's issue with the image again. Let's keep an eye out for it as we continue iterating.

gziolo:
I wanted to test keyboard navigation to ensure that the flow for connecting sources is accessible. I pinned a Playground instance with the Gutenberg plugin installed, and I ran into an edge case where I had no sources exposed which essentially impacts every new WP instance:


https://github.com/user-attachments/assets/d5d1f525-0a06-47d2-a03c-88d987ad569e

When I move focus to a different block, the `content` with no connection is no longer listed. This edge case needs to be better secured before landing this PR.

By the way, keyboard navigation works correctly in this limited scenario. Let's make sure to test how keyboard navigation works with multiple attributes connected and multiple Post Meta options registered. At the moment, it only shows Post Meta options even when some custom bindin sources are registered, right?

paaljoachim:
I hope it is alright that I ask some basic questions.

I thought Attributes would be an advanced feature so that it might reside inside the Advanced panel. Or perhaps a new Connection/Attributes panel just above the Advanced panel. Placing it directly into a panel that can be opened which also gives more space for additional controls/options that will be needed.
As it seems a lot more complex compared to anything that exists today and is something that only the intermediate to advanced users would likely be using. 

Is there a connection/attributes kind of Group block or another option in the Group block that would be the parent of inner blocks that it will impact? Or is this just meant for some blocks?
I do likely also have other basic questions that probably is answered some place. If these take up too much space in this PR then someone can instead just DM me on slack. Thanks.

cbravobernal:
> I hope it is alright that I ask some basic questions.
> 

Of course is alright, maybe it would better to have them in a discussion rather than in a Pull Request that will be eventually merged. But no question is basic, no question is not important.

> I thought Attributes would be an advanced feature so that it might reside inside the Advanced panel. Or perhaps a new Connection/Attributes panel just above the Advanced panel. Placing it directly into a panel that can be opened which also gives more space for additional controls/options that will be needed. As it seems a lot more complex compared to anything that exists today and is something that only the intermediate to advanced users would likely be using.

My opinion is that it should have its own panel. As in a future, it can be a tool to create entire web layouts. I agree is an advanced feature. Ping @jarekmorawski  for opinion. I think we can move the panel without issues after some feedback.

> Is there a connection/attributes kind of Group block or another option in the Group block that would be the parent of inner blocks that it will impact? Or is this just meant for some blocks?

Attributes available for connections are listed here:
https://github.com/WordPress/gutenberg/blob/d14d50feb969794728c7bc6244c3af33e18b3a06/packages/block-editor/src/hooks/use-bindings-attributes.js#L19-L31
```
const BLOCK_BINDINGS_ALLOWED_BLOCKS = {
	'core/paragraph': [ 'content' ],
	'core/heading': [ 'content' ],
	'core/image': [ 'id', 'url', 'title', 'alt' ],
	'core/button': [ 'url', 'text', 'linkTarget', 'rel' ],
};
```
We need to create documentation to share them, and also an API to add your own block-attributes relation. More attribute/blocks compatibility will be added in a future. There are a few PRs to enable them.

> I do likely also have other basic questions that probably is answered some place. If these take up too much space in this PR then someone can instead just DM me on slack. Thanks.

As mentioned before, we can create a [GitHub discussion](https://github.com/WordPress/gutenberg/discussions). And I'll try to answer all of them. Then we can use that information for documentation. Let's avoid private discussions.

SantosGuillamot:
> I ran into an edge case where I had no sources exposed which essentially impacts every new WP instance

I believe the issue you are facing is not because there are no sources (by default `core/post-meta` is registered) but because you have no meta fields. The first case should be handled by the existing code, and I just pushed a fix that should make it work when no meta fields are registered: [link](https://github.com/WordPress/gutenberg/pull/62880/commits/2dff252c23fb4f6b204d7d950523d69e3be7d0b0).

> When I move focus to a different block, the content with no connection is no longer listed. This edge case needs to be better secured before landing this PR.

I am not sure if this is the expected behavior of ToolsPanel. It seems to be the case for other styles using it. For example, if I go to paragraph styles and I click on the three dots of Color or Typography, I add an option, and I move to a different block, it is not there anymore. I understand it like this: "You started creating a connection, you didn't finish it, so it is not created. It doesn't show until you create it again".

Having said that, I don't have a strong opinion on what should be the behavior, but I believe it is not specific to this pull request.

> Let's make sure to test how keyboard navigation works with multiple attributes connected and multiple Post Meta options registered.

It seems the keyboard navigation of the dropdown is not working as expected. Although I believe it is related to the [DropDownMenu component](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs) we are using, because it doesn't seem to work in the docs examples either.

Steps to reproduce: Go to the [component docs](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs) -> Navigate to the "Open Menu" button with the keyboard -> Press "Enter" -> Try to navigate through the items.

@ciampo Is this a known issue? Should I submit one, or do you have any clue how to solve it?

EDIT: Maybe this is expected and users can just navigate with the arrows? 🤔 It seems to be the case in [Ariakit Menu component](https://ariakit.org/components/menu).

> At the moment, it only shows Post Meta options even when some custom binding sources are registered, right?

Yes. The idea is to explore how to support other sources in follow-up pull requests.

SantosGuillamot:
It seems the last remaining decision here is:

### Should we move the UI to the Advanced panel?

This was suggested in [this comment](https://github.com/WordPress/gutenberg/pull/62880#issuecomment-2254243829) and in this [WordPress Slack conversation](https://wordpress.slack.com/archives/C015GUFFC00/p1721919513250229?thread_ts=1721918654.275249&cid=C015GUFFC00).

Basically, the suggestion comes from the fact that it seems an advanced feature.

Apart from that, moving it to the advanced panel wouldn't change the default tab of blocks like paragraphs and headings. As explained as part of [this comment](https://github.com/WordPress/gutenberg/pull/62880#issuecomment-2250184606), with the current implementation, when you create a new paragraph it shows the "Settings" tab instead of the "Styles" tab because we are adding the "Attributes" panel.

Whatever we decide, the position of the panel can always be changed later. It seems moving it to the Advanced section would be "safer" and would give us time to decide how to handle the edge cases.

On the other hand, I believe there is no precedent for using the `ToolsPanel` inside the Advanced section, which could need some modifications.

**How it looks outside the Advanced panel (current implementation)**

Screenshot 2024-07-29 at 09 38 53

**How it looks inside the Advanced panel by default**

Screenshot 2024-07-29 at 09 38 19

**How it looks inside the Advanced panel with custom CSS**
Screenshot 2024-07-29 at 09 37 57

Any thoughts?

jarekmorawski:
I'm not sure it's a good idea. If we want to minimize the footprint of this section, we can remove the help text and rename it to `Connections` (`Attributes` feels like a technical term; in the future bindings may support more than just attributes). Users who know what Connections are will know it's there; everyone else won't mind.

image

My argument against putting it in `Advanced` is that it may become difficult to find even for more savvy users.

The section is collapsed by default and contains features that are rarely needed for daily use. It's also quite consistent across different blocks so if people aren't used to looking there while editing bindings-enabled blocks, they may never discover the feature.

Lastly, the Advanced section is always placed at the bottom of the settings panel, which may cause information hierarchy issues. Applied connections will affect the fields and data _above_ which may feel odd and backfire at some point.

For instance, a user wanting to insert the post title in the Image block's alt text may start looking for such a setting in the Settings section which would be placed high above Connections tucked away under Advanced. Going all the way down there only to be forced to scroll back up to see the result may throw people off.

tyxla:
> It seems the keyboard navigation of the dropdown is not working as expected. Although I believe it is related to the [DropDownMenu component](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs) we are using, because it doesn't seem to work in the docs examples either.
> 
> Steps to reproduce: Go to the [component docs](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs) -> Navigate to the "Open Menu" button with the keyboard -> Press "Enter" -> Try to navigate through the items.
> 
> @ciampo Is this a known issue? Should I submit one, or do you have any clue how to solve it?
> 
> EDIT: Maybe this is expected and users can just navigate with the arrows? 🤔 It seems to be the case in [Ariakit Menu component](https://ariakit.org/components/menu).

Can you please elaborate on what doesn't work as expected? In my testing, keyboard navigation works as expected.

packages/block-editor/src/hooks/block-bindings.js
ciampo: Sorry, was AFK over the past days — the truncation for the label should be solved.

Apart from that, the item label and the item help text components intentionally truncate text to 1 line (for the label) and 2 lines (for the help text). It's a design decision, as we didn't want the content of a menu item to become very long — user should try to have short labels and help text.

If we think that such truncation is too limiting, please leave feedback in a separate issue and we'll discuss it there

packages/block-editor/src/hooks/block-bindings.js
ciampo: > We could just add a gray background to the box. What do you think?
> it can be done in a follow-up.

Agreed. Also, since this part of the UI is using `ItemGroup` and `Item`, it's a good opportunity to codify what a disabled `Item` looks like with the design folks, and add support for it directly to the component.

Please ping @WordPress/gutenberg-components when you start working on it!

gziolo:
> Can you please elaborate on what doesn't work as expected? In my testing, keyboard navigation works as expected

It looks like it would help to see the steps used on both ends to cross-check the expected behavior.

packages/block-editor/src/hooks/block-bindings.scss
ciampo: We were not as involved in the `ToolsPanel` component as on other components, and therefore we may lack some context on how the component layout works — but, looking at [the component's docs](https://github.com/WordPress/gutenberg/blob/391aef63dc6020610d58065dc42d94761e280e06/packages/components/src/tools-panel/tools-panel/README.md#toolspanel-layout), it seems like those decisions were taken deliberately.

My guess is that `ToolsPanel` was thought for a very specific case, and at the time it was not considered to allow more flexibility (perhaps on purpose, to ensure more coherency across the block editor's UI).

My advice is to follow the recommendations of the docs, and at most open a separate issue where we can discuss if and how to make `ToolsPanel`'s layout more flexible (by allowing a custom number of columns? by allowing each item to specify a col span ?)

gziolo:
> > When I move focus to a different block, the content with no connection is no longer listed. This edge case needs to be better secured before landing this PR.
>
> I am not sure if this is the expected behavior of ToolsPanel. It seems to be the case for other styles using it. For example, if I go to paragraph styles and I click on the three dots of Color or Typography, I add an option, and I move to a different block, it is not there anymore. I understand it like this: "You started creating a connection, you didn't finish it, so it is not created. It doesn't show until you create it again".
>
> Having said that, I don't have a strong opinion on what should be the behavior, but I believe it is not specific to this pull request.

I can confirm in my testing that it's working in a similar fashion to other panels. One thing that is important to highlight here is that it's at least two-step process for creating the connection so we should iterate over it later to make the feedback loop better. For example in the following state:

Screenshot 2024-07-29 at 14 57 55

We could provide a visual indicator (warning icon with a hint on hover) and accessible label on the button explaining that the attribute isn't fully connected. An alternative could be some sort of confirmation dialog that asks the user if they want to stop connecting the source with this attribute, assuming the focus management isn't too complex to achieve that. I'm just throwing some ideas to spark the discussion.

> I'm not sure it's a good idea. If we want to minimize the footprint of this section, we can remove the help text and rename it to Connections (Attributes feels like a technical term; in the future bindings may support more than just attributes). Users who know what Connections are will know it's there; everyone else won't mind.

That could be enough to start with. It probably should also conditionally check if there are any custom post meta fields (anything to connect) and conditionally show the panel. In the future, that would need to scan all sources to check if there is anything to connect with.

Overall, I think we are close to landing this work based on the type of feedback we are currently discussing 🎉 Let's clarify whether the UX is accessible from the start, too. Otherwise, we can go through a round of enhancements in smaller targeted PRs so everyone can share feedback freely without being afraid to add further noise as I start to feel with my feedback now 😅

gziolo (approved):
50f6f1e](https://github.com/WordPress/gutenberg/pull/62880/commits/50f6f1ef174fb15ba7aaf288e620faf1cf664718) - with that it’s behind the experiment. Let’s land it and decide separately when to expose it to the users. In my opinion, it’s going to be really soon, even before the next Gutenberg RC1 planned for next week. However, if this is the confidence necessary then I’m happy with it. It could be that e2e tests will need some updating to account for the experiment UI.

SantosGuillamot:
> Can you please elaborate on what doesn't work as expected? In my testing, keyboard navigation works as expected.

@tyxla I was trying to navigate the menu items with "SHIFT" and it seems I need to use the arrows instead. As said in the edit, maybe this is expected because it is how Ariakit Menu component works.

> One thing that is important to highlight here is that it's at least two-step process for creating the connection so we should iterate over it later to make the feedback loop better.

I agree this could be improved. I suggested something similar in [this comment](https://github.com/WordPress/gutenberg/pull/62880#pullrequestreview-2193617739): _"Should we add a prompt or helper message when an attribute is not connected?"_.

> It probably should also conditionally check if there are any custom post meta fields (anything to connect) and conditionally show the panel. In the future, that would need to scan all sources to check if there is anything to connect with.

That should already happen in [this logic](https://github.com/WordPress/gutenberg/blob/50f6f1ef174fb15ba7aaf288e620faf1cf664718/packages/block-editor/src/hooks/block-bindings.js#L270-L309). When there are no custom fields to connect to, it will show only the panel when there are bindings created.

> Overall, I think we are close to landing this work based on the type of feedback we are currently discussing 🎉

As it seems there are some concerns, I moved the UI under an experimental flag. I feel landing the prototype will help moving forward the follow-up discussions, fixes, and pull requests.

packages/block-editor/src/hooks/block-bindings.scss
SantosGuillamot: So, instead of changing the wrapper to have just one column, we need to change the style of each item to fit both columns?

For reference, other examples where the "one-column" layout is used:

**Colors**
Screenshot 2024-07-30 at 08 45 45

**Dimensions**
Screenshot 2024-07-30 at 08 46 17

**Filters**
Screenshot 2024-07-30 at 08 46 02

**Border & Shadow**
Screenshot 2024-07-30 at 08 46 34

tyxla:
> > Can you please elaborate on what doesn't work as expected? In my testing, keyboard navigation works as expected.
> 
> @tyxla I was trying to navigate the menu items with "SHIFT" and it seems I need to use the arrows instead. As said in the edit, maybe this is expected because it is how Ariakit Menu component works.

Thanks for confirming, this is exactly how it's supposed to work - with arrows, and you can also use SPACE and ENTER for selection and navigating between levels (in the nested version).

ciampo:
>  it is how Ariakit Menu component works.

To add to what Marin said, that is how `menu` widgets should work (regardless of the underlying implementation), as per the [WAI ARIA spec](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)

SantosGuillamot:
As it is now wrapped under an experimental flag, I went ahead and merged the pull request. I'll add the follow-ups to the tracking issue and we can discuss items separately. Thanks everyone for all the feedback and the great work 👏

packages/block-editor/src/hooks/block-bindings.js
SantosGuillamot: I have created [this follow-up pull request](https://github.com/WordPress/gutenberg/pull/64105) to use only one util and keep discussing it.

fabiankaegy:
Hey all 👋 

Is is correct that the bindings UI only shows the core meta binding sources as options to select from? Custom sources registered via the PHP `register_block_bindings_source` are not displaying for me 🤔 

CC: @SantosGuillamot

SantosGuillamot:
> Is is correct that the bindings UI only shows the core meta binding sources as options to select from?

Yes, that's correct. The UI is specific for post meta at this moment. We explored the possibility of extending the UI to other sources, but we didn't consider the APIs stable enough to open them for 6.7. It will probably be one of the focus for 6.8, but we need to ensure whatever is built is solid enough to avoid potential issues in the future.

fabiankaegy:
Thanks for confirming :) That is another detail that I think will be really important to include in the Dev Note for the block bindings feature :)

SantosGuillamot:
I started exploring the possibility of providing the APIs to extend the UI in [this pull request](https://github.com/WordPress/gutenberg/pull/65994). I plan to work on another one on top of it to provide a "default" UI for sources only registered in the server.

PR #62881, opened 2024-06-26
Use Base Focus Styles for Region Focus

jeryj:
## What?

Make the outline style for region and block selection outline have a consistent color and weight ratio.

## Why?

To make the editor UX more consistent.

## How?

- refactors the selected block outline mixin to be globally available
- uses the selected block focus mixin for both the selected blocks and region focus.
- uses a $widthRatio to be able to pass a number to the outline mixin to get a nicely stepped width

#### Deselecting Blocks
- Select block
- a thin blue outline should be visible around the block selection.
- Escape to enter navigation mode
- Escape again to move focus to the region. 
- A blue outline around the editor canvas should be visible.
- The blue outline should be twice as thick as the selected block outline.

#### Region Navigation
- Use the region navigation shortcuts (Previous: ctrl+option+p and Next: ctrl+option+n)  
- A blue outline around the region should be visible.
- The blue outline should be twice as thick as the selected block outline.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/967608/f1af4a9e-3d44-4468-ac0c-5c1ad9811aab

### DIFF
--- a/packages/base-styles/_mixins.scss
+++ b/packages/base-styles/_mixins.scss
@@ -597,3 +597,21 @@
 		}
 	}
 }
+
+@mixin selected-block-outline($widthRatio: 1) {
+	outline-color: var(--wp-admin-theme-color);
+	outline-style: solid;
+	outline-width: calc(#{$widthRatio} * (var(--wp-admin-border-width-focus) / var(--wp-block-editor-iframe-zoom-out-scale, 1)));
+	outline-offset: calc(#{$widthRatio} * ((-1 * var(--wp-admin-border-width-focus) ) / var(--wp-block-editor-iframe-zoom-out-scale, 1)));
+}
+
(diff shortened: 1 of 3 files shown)


scruffian (approved):
LGTM. It might be good to have a design review too as the borders do change slightly:

Before:
Screenshot 2024-07-05 at 15 54 45

After: 
Screenshot 2024-07-05 at 15 55 35

jasmussen:
Can work. Still not sure this outline needs to be thicker than any othe focus style, all of which are 1.5px. Is it not confusing that this one is suddenly thicker? What's special about this one? Not a strong opinion, but asking specifically since it's becoming a mixin, suggesting application elsewhere.

scruffian:
It is possible to have both a block selected and a region focussed. Giving them a slightly different visual treatment feels useful in helping users to distinguish the difference between those two concepts.

PR #62883, opened 2024-06-26
Remove link to polyfill.io

westonruter:
The polyfill.io domain has been hijacked in a supply chain attack. While we don't seem to be pulling anything from the domain, we are linking to it. This link should be removed. 

See:

* https://thehackernews.com/2024/06/over-110000-websites-affected-by.html
* https://www.theregister.com/2024/06/25/polyfillio_china_crisis/
* https://www.bleepingcomputer.com/news/security/polyfillio-javascript-supply-chain-attack-impacts-over-100k-sites/


ockham (approved):
Thank you!

ellatrix (approved):
Could we link to the npm package instead?

ellatrix:
Whoops, sorry missed the previous approval.

Looks like it's coming from here:

https://github.com/WordPress/wordpress-develop/blob/36e6d98a228729c2993b3c7921d9a31efca55d0a/tools/webpack/packages.js#L88

https://www.npmjs.com/package/polyfill-library

PR #62890, opened 2024-06-26
Fix Interactivity API Reference minor typos

jffng:
## What?
Fixes a couple of import typos I encountered while using the docs.


fabiankaegy (approved):
Great catches :)

PR #62891, opened 2024-06-26
Adds Aspect ratio control on Image blocks in Grids

### LINKED ISSUE #62889: Grid: Aspect ratio controls are disabled on Image blocks in Grids
### Description

If you add an Image block to a Grid, the Aspect Ratio and Width/Height controls are disabled. The Width/Height makes sense since the actual width and height of the image will be dictated by the grid layout, but the Aspect Ratio should still be enabled, in my opinion. Perhaps this is intentional, but it feels like a bug. 

I ran into a situation where I had a number of images that all had different aspect ratios. I wanted to set them all to"Square" and then arrange them in the new Grid block. After inserting the images, I realized the Aspect Ratio control no longer appeared in the block inspector. However, if I added an Image outside of the Grid and set the aspect ratio to "Square", I could then move it into the Grid, and the aspect ratio would be preserved. The video below provides a walkthrough.

1. Use WordPress 6.6 or Gutenberg trunk
2. Add a Grid block
3. Add a few images with different aspect ratios to the Grid
4. Notice that the Aspect Ratio control is **not** visible
5. Add an Image block outside of the Grid, notice that the Aspect Ratio control **is** visible, and set it to "Square"
6. Now move that Image block inside of the Grid and notice that the aspect ratio is preserved

https://github.com/WordPress/gutenberg/assets/4832319/cbee7dc4-6854-48bc-9b00-0be80867498e

### Environment info

- WordPress 6.6
- Gutenberg trunk

Yes

Yes

amitraj2203:
## What?
Fixes: #62889 

## Why?
Currently, when adding an Image block within a Grid layout, the Aspect Ratio control is disabled. This can limit user's ability to set consistent visual proportions across images within the grid. Enabling the Aspect Ratio control allows users to maintain uniformity in image display.

## How?
Conditionally shows Aspect ratio control on grid layouts.

## Testing Instructions
1. Add a Grid block
2. Add few images with different aspect ratios to the Grid
3. Notice that the Aspect Ratio control is **visible** for the images

## Screenshots or screencast 
https://github.com/WordPress/gutenberg/assets/77401999/e6d7c76e-ca18-4744-aba1-44f4b1b91a52


tellthemachines (commented):
Thanks for taking this on @amitraj2203! 

I think ideally we'd find a solution that doesn't involve `DimensionsTool` having to know about layout. Perhaps instead of passing it `parentLayoutType` directly, we can introduce an array type prop with a list of tools to render. It could default to all tools, e.g. `[ 'aspectRatio', 'widthHeight', 'scale']`, but if we pass it ['aspectRatio'] it only renders the aspect ratio tool. Then we can leave deciding which tools to render based on parent layout type to the image block.

amitraj2203:
@tellthemachines Updated it as per your suggestion.

tellthemachines (commented):
Thanks for iterating on this! The code is looking good now, there's only one small issue remaining. Currently, setting aspect ratio on images in a Grid block results in the images becoming distorted:
Screenshot 2024-06-28 at 10 57 40 AM

This is due to the `object-fit` property not being set. In Image blocks outside Grid, [`cover` is being used](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/image.js#L401) as a default value for the `scale` attribute, which translates to the `object-fit` CSS property. Inside a grid, because we're not exposing the scale tool, I think the best solution is to set the attribute directly in `aspectRatioControl`, like so:

```
setAttributes( {
	aspectRatio: newAspectRatio,
	scale: 'cover',
} );
```
Apologies for not spotting this in my earlier review!

tellthemachines:
Oh, I just noticed something else in testing: the resizable box handles on the Image block are now appearing beneath the grid resizers:

Screenshot 2024-06-28 at 11 09 29 AM

We'll need to disable them, but I'm not 100% sure how. Looking into it now.

Update: changing the condition on [this line](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/image.js#L851) to `! isResizable || parentLayoutType === 'grid'` should stop the resizable box from rendering.

amitraj2203:
Thanks for reviewing it thoroughly. I have addressed the changes.

tellthemachines (approved):
Thanks for the update @amitraj2203! This is working well for me now.

PR #62894, opened 2024-06-27
Fix extra scrollbar when a popover extends past the viewport.

tellthemachines:
## What?

When an active popover inside `interface-skeleton__content` extends vertically past the viewport, an extra scrollbar appears on trunk. This can be reproduced with the following steps:

* Make sure scrollbars are enabled at all times in your computer settings;
* Add a Grid block to a post or template, and add enough content inside it that it becomes scrollable;
* Make sure either the Grid or one of its children are selected, and scroll to a point where the Grid is partially hidden beyond the bottom edge of the viewport;
* See the extra vertical scrollbar appear.

The overflow rule being changed here has been around for a long time so I want to make sure this change doesn't break anything else 😅 
My testing shows everything working correctly: both editor canvas and sidebar are still scrollable and look as expected. I think the overflow may have been set to auto to unset its default `visible`, because when setting it to `visible` things do look a bit funny. If that's the extent of what the overflow is doing, this change should be safe enough.

Before:

Screenshot 2024-06-27 at 11 24 40 AM

After:

Screenshot 2024-06-27 at 11 33 02 AM

### DIFF
--- a/packages/editor/src/components/visual-editor/style.scss
+++ b/packages/editor/src/components/visual-editor/style.scss
@@ -19,7 +19,7 @@
 
 	// In the iframed canvas this keeps extra scrollbars from appearing (when block toolbars overflow). In the
 	// legacy (non-iframed) canvas, overflow must not be hidden in order to maintain support for sticky positioning.
-	.is-iframed {
+	&.is-iframed {
 		overflow: hidden;
 	}
 
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -82,9 +82,10 @@ html.interface-interface-skeleton__html-container {
 	display: flex;
 	flex-direction: column;
 
-	// Hide the Y overflow to avoid an extra scrollbar when an active popover
-	// extends vertically beyond the viewport.
-	overflow-y: auto;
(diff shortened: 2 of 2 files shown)


talldan (commented):
Mostly tests well for me, but I noticed there are some issues scrolling in code editor mode and in the widget editor

talldan (approved):
Tests well for me. It fixes the grid block issue, the earlier issues is resolved, and I'm still seeing scrollbars working in places like focus mode, widget editor, code editor mode, style book, the various post editor modes.

t-hamano:
It seems that #62632 was also attempting to solve this issue. Instead of closing #62632, would it be better to backport this PR to WP6.6?

cc @ellatrix @stokesman

talldan:
@t-hamano I think #62632 might be a different bug. I've spotted a few different ones, also another PR here - https://github.com/WordPress/gutenberg/pull/62909.

I agree this one should be backported.

edit: Actually, maybe it is the same bug, it's hard to say!

stokesman:
This is a nice fix and does resolve #62319 in my testing. Great finding that ineffective selector as with another fix like #62632 it would have merely become cruft.

ellatrix:
Nice!

PR #62900, opened 2024-06-27
Cherry pick automation: fix for forks

ellatrix:
## What?

This workflow was introduced in #62716, and it generally works well. The problem is that when a PR is merged from a fork, the workflow runs in the context of that fork and lacks the token.

The idea of this PR is to move automation from PRs to pushes. I'll first test this on a `test/trunk` branch.

## Why?

Because it doesn't currently work for PRs from forks, such as #62857.

## Testing Instructions

It works:

* I created a `test/trunk` branch with this new workflow.
* I created a PR for it with the label `Backport to WP 0.0 Beta/RC`, merged it, and it was correctly cherry-picked: 
  * Workflow run: https://github.com/WordPress/gutenberg/actions/runs/9691420804/job/26742868834
  * Comment: https://github.com/WordPress/gutenberg/pull/62903#issuecomment-2193863431
  
Labelling after a merge still works too, but worth noting that labelling after will still not work on forks because this workflow must run on the PR:
* Workflow run: https://github.com/WordPress/gutenberg/actions/runs/9692818737/job/26746955238
* Comment: https://github.com/WordPress/gutenberg/pull/62910#issuecomment-2194041795

### DIFF
--- a/.github/workflows/cherry-pick-wp-release.yml
+++ b/.github/workflows/cherry-pick-wp-release.yml
@@ -28,6 +28,7 @@ jobs:
                   script: |
                       const commit_sha = context.payload.pull_request ? context.payload.pull_request.merge_commit_sha : context.sha;
                       console.log(`Commit SHA: ${commit_sha}`);
+                      core.exportVariable('commit_sha', commit_sha);
                       const prs = await github.rest.repos.listPullRequestsAssociatedWithCommit({
                         owner: context.repo.owner,
                         repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


vcanales (commented):
> Labelling after a merge still works too, but worth noting that labelling after will still not work on forks because this workflow must run on the PR

If this means that we'd have to cherry-pick those manually I think it's fine, but could we add some indication that some PRs are "missing"?

My concern is that the presence of automation in part of the process might _relax_ those in charge of the release if we don't state explicitly that they still have to check the cherry picks.

ellatrix:
> If this means that we'd have to cherry-pick those manually I think it's fine, but could we add some indication that some PRs are "missing"?

We anyway still have to check the label (which would be kept). There might be other job runs that failed for some other reason. It happens. At some point I'll update the handbook and I will mention that it's still required to check the label manually.

vcanales (approved):
LGTM :)

PR #62464, opened 2024-06-11
Block style variation: rename hook

ramonjd:
## What?
Just renaming the function.

## Testing Instructions
I smoke tested  by adding a block style variation

```json
{
	"$schema": "../../schemas/json/theme.json",
	"version": 3,
	"settings": {
		"appearanceTools": true
	},
	"title": "My Variation",
	"blockTypes": [ "core/group", "core/columns", "core/media-text" ],
	"styles": {
		"color": {
			"background": "green",
			"text": "red"
		}
	}
}
```

And applied to a Group block in the editor. Save and make sure all is well.

PR #62905, opened 2024-06-27
Manual cherry pick of #62768 to wp/6.6

ramonjd:
## What?

A manual cherry pick of the following PRs to `wp/6.6`

- https://github.com/WordPress/gutenberg/pull/62464

## Why?

There was a tiny conflict because the hook name change in #62464 was not cherry-picked into `wp/6.6`

## Testing Instructions
Smoke test according to test instructions in https://github.com/WordPress/gutenberg/pull/62768

### DIFF
--- a/packages/block-editor/src/hooks/block-style-variation.js
+++ b/packages/block-editor/src/hooks/block-style-variation.js
@@ -16,6 +16,7 @@ import {
 import { useStyleOverride } from './utils';
 import { store as blockEditorStore } from '../store';
 import { globalStylesDataKey } from '../store/private-keys';
+import { unlock } from '../lock-unlock';
 
 const VARIATION_PREFIX = 'is-style-';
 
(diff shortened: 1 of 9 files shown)


aaronrobertshaw (approved):
Thanks for putting this together @ramonjd 👍 

✅ PR tests as advertised in https://github.com/WordPress/gutenberg/pull/62768
✅ Code changes match the combined updates of both PRs


https://github.com/WordPress/gutenberg/assets/60436221/91d231b9-fd78-433c-a820-b8d99a071c2b

andrewserong (approved):
+1, this is testing well for me, too, and the code change looks good visually (includes both PRs). 🚀

packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -59,7 +60,127 @@ function getVariationNameFromClass( className, registeredStyles = [] ) {
...
+ */
+export function __unstableBlockStyleVariationOverridesWithConfig( { config } ) {
+	const { getBlockStyles, overrides } = useSelect(
+		( select ) => ( {
+			getBlockStyles: select( blocksStore ).getBlockStyles,
+			overrides: unlock( select( blockEditorStore ) ).getStyleOverrides(),
+		} ),
+		[]
+	);
+	const { getBlockName } = useSelect( blockEditorStore );
```
ellatrix, line 87: So both `getBlockStyles` and `getBlockName` can't be moved into subscriptions? The catch here is that if any of these change, the overrides won't update.

ramonjd: The blocks/block clientIds won't change between revisions.

We could map over the results of `getStyleOverrides()` in the `useSelect` callback and inject the block name into the overrides array items. 

So something like:

```js
	const { overrides, getBlockStyles } = useSelect(
		( select ) => {
			const { getBlockName, getStyleOverrides } = unlock( select( blockEditorStore ) ) ;
			const _overrides = getStyleOverrides();
			return {
				overrides: _overrides?.length ?
					_overrides.map( ( [ id, override ] ) =>
						[ id, { ...override, blockName: getBlockName( override?.clientId ) } ]
					) : [],
				getBlockStyles: select( blocksStore ).getBlockStyles,
			}
		},
		[]
	);
```
What do you reckon?

I think there's a flakey E2E test I'll have to take care of too as a follow up.

ramonjd: I'll do the above in a follow up in trunk today or tomorrow first thing 👍🏻 with a view to merge back into wp/6.6

ramonjd: Follow up https://github.com/WordPress/gutenberg/pull/62915

ellatrix: Commented there. Also I think it's only really valuable if `getBlockStyles` can be changed too.

packages/edit-site/src/components/revisions/index.js
```diff
@@ -25,6 +25,7 @@ const {
 	ExperimentalBlockEditorProvider,
 	GlobalStylesContext,
 	useGlobalStylesOutputWithConfig,
+	__unstableBlockStyleVariationOverridesWithConfig,
```
ellatrix, line 28: Why is this `__unstable`? It's already private, right?

ramonjd: There was a [bit of discussion](https://github.com/WordPress/gutenberg/pull/62768#discussion_r1654254618) around it. 

To answer the "why", to advertise without ambiguity that, despite being private, it's unstable.

Similar to how [private block editor functions](https://github.com/WordPress/gutenberg/blob/55f3a98159a754cee5162a05852881fa168134a7/packages/block-editor/src/private-apis.js#L67-L67) have the `Experimental` or `Private` prefixes 🤷🏻 


I don't think it does any harm, but could easily be communicated in comments. I can rename it ~next week~ if you feel strongly about it.

PR #30557, opened 2021-04-07
Stop keypresses being caught by other elements when they happen in a CustomSelectControl

opr:
## Description

In the `<ul>` output by `CustomSelectControl` I have added a function to handle `onKeyDown` - this stops the event from propagating, but still calls the `onKeyDown` handler (which is derived ultimately from `downshift`). 

This PR fixes #30204.

## How has this been tested?

In the linked issue (#30204) the reproducible steps can also be used to test:

1. Open the storybook for this project: `npm run storybook:dev`
2. Go to the `CustomSelectControl > Default` component, click it and start typing `s` to get to the entry `Small`
3. Notice the storybook sidebar no longer closes due to the `s` keypress being cancelled.

## Types of changes

Bug fix

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards.  **Unsure, I had to add eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions but I don't see any other way of capturing this event**
- [x] I've tested my changes with keyboard and screen readers.

PR #62907, opened 2024-06-27
CustomSelectControl V2: prevent keyboard event propagation in legacy wrapper

ciampo:
## What?

Part of #55023 

Prevent keyboard events propagation from the select popover in `CustomSelectControlV2` legacy wrapper

## Why?

To align the behavior of the V2 legacy wrapper to the V1 version of the component. Here's more [context](https://github.com/WordPress/gutenberg/pull/30557) to why the V1 stops event propagation.

> [!NOTE]  
> I personally think that this behaviour is non-standard and shouldn't be retained for the V2 (non-legacy) version of the component.

## How?

- Introduced a way for the internal V2 implementation to know whether its consumer is "legacy" or not
- when in "legacy" mode, stopped `keydown` event propagation on the select dropdown
- restored legacy V2 unit tests that make sure that the V2 legacy wrapper behaves like the V1 component

> [!Tip]  
> The "legacy" flag introduced in this PR can be further explored to trigger more differences in behaviour and styles as we further define specs of the V2 (non-legacy) version.

## Testing Instructions

- Make sure that the un-skipped unit test for the V2 legacy wrapper passes
- In Storybook:
  1. Load the default Storybook example for the V2 legacy wrapper
  2. Open the select dropdown and press the "S" key — the "Small" element should be highlighted, but the Storybook sidebar should not toggle its opened/closed state
  3. The same should be true when testing the V1 version of the component
  4. Do the same on the `trunk` version of Storybook for the V2 legacy wrapper, and notice how the sidebar toggles between opened and closed

### DIFF
--- a/packages/components/src/custom-select-control-v2/types.ts
+++ b/packages/components/src/custom-select-control-v2/types.ts
@@ -68,6 +68,10 @@ export type _CustomSelectProps = CustomSelectButtonProps & {
 	 * Accessible label for the control.
 	 */
 	label: string;
+	/**
+	 * True if the consumer is emulating the legacy component behavior and look
+	 */
+	isLegacy?: boolean;
 };
 
 export type CustomSelectProps = _CustomSelectProps &
--- a/packages/components/src/custom-select-control-v2/types.ts
+++ b/packages/components/src/custom-select-control-v2/types.ts
@@ -75,6 +75,8 @@ export type _CustomSelectProps = CustomSelectButtonProps & {
 };
 
 export type CustomSelectProps = _CustomSelectProps &
+	// TODO: is it necessary to pass CustomSelectButtonProps, since
+	// they are already part of _CustomSelectProps?
 	CustomSelectButtonProps &
 	CustomSelectSize;
 
--- a/packages/components/src/custom-select-control-v2/types.ts
+++ b/packages/components/src/custom-select-control-v2/types.ts
@@ -49,6 +49,15 @@ export type CustomSelectButtonProps = {
 	value?: string | string[];
 };
 
+// Props only exposed on the internal implementation
+export type _CustomSelectInternalProps = {
+	/**
+	 * True if the consumer is emulating the legacy component behavior and look
+	 */
+	isLegacy?: boolean;
+};
+
+// Props that are exposed in exported components
 export type _CustomSelectProps = CustomSelectButtonProps & {
 	/**
 	 * Additional className added to the root wrapper element.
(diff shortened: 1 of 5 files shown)


packages/components/src/custom-select-control-v2/types.ts
```diff
@@ -70,9 +79,7 @@ export type _CustomSelectProps = CustomSelectButtonProps & {
 	label: string;
 };
 
-export type CustomSelectProps = _CustomSelectProps &
-	CustomSelectButtonProps &
```
ciampo, line 74: Removed `CustomSelectButtonProps` since it's already part of `_CustomSelectProps`

packages/components/src/custom-select-control-v2/types.ts
```diff
@@ -49,6 +49,15 @@ export type CustomSelectButtonProps = {
 	value?: string | string[];
 };
 
+// Props only exposed on the internal implementation
+export type _CustomSelectInternalProps = {
+	/**
+	 * True if the consumer is emulating the legacy component behavior and look
+	 */
+	isLegacy?: boolean;
+};
```
ciampo, line 58: I needed a way to allow the internal implementation to expose a `isLegacy` flag, without forwarding it to the public API of the V2 component. Not sure if we think there's a better way to do that? @mirka

tyxla: Can't think of a better way, TBH. 

What a bummer that we have to do it with hacks like this. It's an indication that the legacy adapter approach has probably too many downsides and won't be a recommended one in the future.

tyxla (approved):
Looks good and tests well 👍 

Maybe it's just another sign that the adapter approach wasn't the best idea. Something to learn from for future refactors.

packages/components/src/custom-select-control-v2/types.ts
ciampo: @mirka — just making sure that you see this change, since you refactored these types last

packages/components/src/custom-select-control-v2/types.ts
ciampo: The legacy adapted approach definitely has its pros and cons.

It may not be worth keeping a common underlying implementation for V1 and V2 if we then have to abstract it and complicate it with extra props and exceptions. Let's keep it as-is for now and look back at it once V2 legacy wrapper has shipped and we look at V2

packages/components/src/custom-select-control-v2/types.ts
mirka: Good catch, thanks 🙏

PR #62912, opened 2024-06-27
Fix variable name in color directive example for Interactivity API

airman5573:
## What?
This PR corrects a variable name in the color directive example code for the Interactivity API.

## Why?
The current example code contains a minor error where 'color' is used instead of 'colors' in the destructured parameter. This could lead to confusion for developers trying to implement the color directive. By fixing this, I ensure that the documentation accurately reflects the correct usage of the API.

## How?
The implementation involves a simple change in the example code:
- Changed `{ directives: { color } }` to `{ directives: { color: colors } }` in the destructured parameter of the directive function.
- Path is `packages/interactivity/src/hooks.tsx` line 234.

### DIFF
--- a/packages/interactivity/src/hooks.tsx
+++ b/packages/interactivity/src/hooks.tsx
@@ -231,7 +231,7 @@ const directivePriorities: Record< string, number > = {};
  * ```js
  * directive(
  *   'color', // Name without prefix and suffix.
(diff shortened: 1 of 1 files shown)


luisherranz (approved):
Good catch!

PR #62913, opened 2024-06-27
DataViews: Register the deletePost action like any third-party action.

youknowriad:
Related #61084 
Similar to #62647 

## What?

The current PR explore the possibility to use the API to register one action: "delete post". 

## Testing Instructions

1- Open the patterns dataviews and try removing a user created pattern.
2- You should be able to see the "delete" action in the actions dropdown menu


youknowriad:
The other similar PR https://github.com/WordPress/gutenberg/pull/62647 got blocked because we need access to the store and we need isElligible to be reactive (which is hard to solve). 

So I'm trying another action that doesn't have these requirements in this PR. But the current PR is also blocked for two reasons:

 - I'm importing JS files in TS files and would like to ignore the JS files in the typing build but it doesn't seem to work so far.
 - I noticed that the template parts dataviews don't pass the "template part" objects as "items", instead they nest them inside `item.templatePart` which is a no-go for me and I'm trying to understand why we're doing this.

packages/editor/src/dataviews/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface TemplateOrTemplatePart extends BasePost {
+	type: 'template' | 'template-part';
+	source: string;
+	has_theme_file: boolean;
+}
+
+export interface OtherPost extends BasePost {
+	type: Exclude< string, ReservedPostTypes >;
+}
```
sirreal: I think we can simplify this section, the following might work:

```suggestion
export interface Post {
	status?: PostStatus;
	title: string | { rendered: string };
	type: string;
}

export interface TemplateOrTemplatePart extends Post {
	type: 'template' | 'template-part';
	source: string;
	has_theme_file: boolean;
}
```

The main reason is that `Exclude< string, ReservedPostTypes >` can't really do anything to narrow the type, it's just `string`. The type "any string except these two literals" is too broad and just becomes `string`. Any `OtherPost` would also satisfy the `TemplateOrTemplatePart` type.

A type guard would be suitable to narrow the type in application code:

```tsx
function isTemplateOrTemplatePartPost( p: Post ): p is TemplateOrTemplatePart {
    return p.type === 'template' || p.type === 'template-part';
}

declare var p: Post;
if ( isTemplateOrTemplatePartPost(p)) {
    p.has_theme_file; // Now `p` is of type `TemplateOrTemplatePart`
}
```

youknowriad:
This should be ready for review.

packages/edit-post/src/index.js
```diff
@@ -91,6 +93,7 @@ export function initializeEditor(
 			enableFSEBlocks: settings.__unstableEnableFullSiteEditingBlocks,
 		} );
 	}
+	registerDefaultActions();
```
youknowriad, line 96: I couldn't find a better place to call this. I kind of maps what we do for blocks.

packages/editor/src/dataviews/types.ts
```diff
@@ -0,0 +1,21 @@
...
+	title: string | { rendered: string };
+	type: string;
+}
+export interface TemplateOrTemplatePart extends BasePost {
+	type: 'template' | 'template-part';
+	source: string;
+	has_theme_file: boolean;
+}
+
+export type Post = TemplateOrTemplatePart | BasePost;
```
youknowriad, line 21: I like a lot that we could use the "actions" refactoring to slowly augment our "entities" types. In the long run, this will help a lot when build UI for entities.

patches/postcss-urlrebase+1.3.0.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/node_modules/postcss-urlrebase/index.d.ts b/node_modules/postcss-urlrebase/index.d.ts
```
jorgefilipecosta, line 1: Why is this change required?

jorgefilipecosta (approved):
Everything worked well on the tests I did 👍

patches/postcss-urlrebase+1.3.0.patch
youknowriad: Something about fixing the typescript types. But @sirreal would know more precisely.

sirreal: A browser package (editor, I believe) uses postcss in the browser.

This dependency expected to work in a node environment (most common for postcss stuff) and use the URL type imported from node. This is not available in the environment. However, it's fine to use the URL global, which is what this change does. Otherwise, the package would error when it tried to import a non-existent (node.js) type.

sirreal: https://github.com/strarsis/postcss-urlrebase/pull/2

sirreal: Add package patch note here: https://github.com/WordPress/gutenberg/pull/63015

packages/editor/src/dataviews/actions/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+	}
+	// In patterns list page we map the templates parts to a different object
+	// than the one returned from the endpoint. This is why we need to check for
+	// two props whether is custom or has a theme file.
+	return (
+		[ template.source, template.source ].includes(
+			TEMPLATE_ORIGINS.custom
+		) &&
+		! template.has_theme_file &&
+		! template?.has_theme_file
```
ntsekouras, line 48: This is the same check now.. https://github.com/WordPress/gutenberg/pull/63021

PR #62915, opened 2024-06-27
Block styles variations E2E: wait for Save button before editing global styles

ramonjd:
## What and how?

Follow up to:

Tweak the E2E test to wait for the Save button to be visible between global styles saves. 🤞🏻 it'll make it less flakey. 

Should be backported to wp/6.6 branch because the original tests are also there.

## Testing Instructions

E2E CI tests should pass.

### DIFF
--- a/packages/block-editor/src/hooks/block-style-variation.js
+++ b/packages/block-editor/src/hooks/block-style-variation.js
@@ -19,6 +19,7 @@ import { globalStylesDataKey } from '../store/private-keys';
 import { unlock } from '../lock-unlock';
 
 const VARIATION_PREFIX = 'is-style-';
+const EMPTY_ARRAY = [];
 
 function getVariationMatches( className ) {
 	if ( ! className ) {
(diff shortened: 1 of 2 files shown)


packages/block-editor/src/hooks/block-style-variation.js
```diff
@@ -77,13 +78,25 @@ function OverrideStyles( { override } ) {
...
-	const { getBlockStyles, overrides } = useSelect(
-		( select ) => ( {
+	const { overrides, getBlockStyles } = useSelect( ( select ) => {
+		const { getBlockName, getStyleOverrides } = unlock(
+			select( blockEditorStore )
+		);
+		const _overrides = getStyleOverrides();
+		return {
+			overrides: _overrides?.length
+				? _overrides.map( ( [ id, override ] ) => [
```
ellatrix: We can't return a new array in a selector unfortunately, it will invalidate every selector result and trigger a re-render for every store change.

ramonjd: Ah true. I knew it was too good to be true. I'll revert and leave this as an E2E update.

test/e2e/specs/site-editor/block-style-variations.spec.js
```diff
@@ -149,12 +149,14 @@ test.describe( 'Block Style Variations', () => {
 				},
 			},
 		} );
-		// The save button has been re-enabled.
+
+		// Wait for the save button to be re-enabled.
```
andrewserong, line 153: No blockers here, just making sure I understand the change here:

This change is now checking that that `Save` is visible, rather than enabled. From looking at the editor manually, while saving, the button says "Saving" rather than "Save", so `.toBeVisible()` is a good check to use, as it'll only pass if the button says "Save" rather than "Saving". Also, once the revision has been saved the button is in a disabled state, so `toBeVisible` is a more reliable check than `toBeEnabled`.

Assuming that's all how it's meant to be, that sounds great to me!

andrewserong (approved):
Looks good to me! 🚀

test/e2e/specs/site-editor/block-style-variations.spec.js
ramonjd: Exactly what you said, thanks for confirming. 

Much better expressed than what my caveman brain was thinking: "Wait for save. Save there. Good save."

PR #62919, opened 2024-06-27
CustomSelectControl V2: keep legacy arrow down behavior only for legacy wrapper

ciampo:
## What?

Tracked in #55023 

Change the result of pressing the arrow down key when focussing the trigger button on V2 `CustomSelectControl` (non-legacy). While the V1 and the V2 legacy wrapper select previous/next option item (with the select popover hidden), the V2 (non legacy) component opens the select popover instead.

## Why?

The V1 behavior felt non-standard, but we see value in retaining the original behavior in the V2 legacy wrapper.

Having a separate V2 allows us to switch to a better, more standard default.

## How?

- Do some type gymnastic to expose the `showOnKeyDown` prop to internal users of the `Button`
- Use the `isLegacy` prop to affect the `showOnKeyDown` behavior
- Add unit tests

## Testing Instructions

- Make sure that unit tests pass
- In Storybook, load the examples for the V1, V2 (legacy wrapper) and V2 (default) versions of `CustomSelectControl`. Focus the trigger button and press the arrow down key. V1 and V2 (legacy wrappers) should keep the select popover closed and change item selection, while V2 (default) should open the select popover.

### DIFF
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -12,6 +12,7 @@ import * as Styled from './styles';
 import type {
 	CustomSelectContext as CustomSelectContextType,
 	CustomSelectStore,
+	CustomSelectButtonInternalProps,
 	CustomSelectButtonProps,
 	CustomSelectButtonSize,
 	_CustomSelectInternalProps,
@@ -52,7 +53,10 @@ const CustomSelectButton = ( {
 	...restProps
 }: Omit<
 	WordPressComponentProps<
-		CustomSelectButtonProps & CustomSelectButtonSize & CustomSelectStore,
+		CustomSelectButtonInternalProps &
+			CustomSelectButtonProps &
+			CustomSelectButtonSize &
+			CustomSelectStore,
 		'button',
 		false
 	>,
@@ -71,9 +75,6 @@ const CustomSelectButton = ( {
 			size={ size }
 			hasCustomRenderProp={ !! renderSelectedValue }
 			store={ store }
-			// to match legacy behavior where using arrow keys
-			// move selection rather than open the popover
-			showOnKeyDown={ false }
 		>
 			{ computedRenderSelectedValue( currentValue ) }
 		</Styled.Select>
@@ -128,6 +129,8 @@ function _CustomSelect(
 					{ ...restProps }
 					size={ size }
 					store={ store }
+					// Match legacy behavior (move selection rather than open the popover)
+					showOnKeyDown={ ! isLegacy }
 				/>
 				<Styled.SelectPopover
 					gutter={ 12 }
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -1,4 +1,10 @@
 /**
+ * External dependencies
+ */
+// eslint-disable-next-line no-restricted-imports
+import type * as Ariakit from '@ariakit/react';
+
+/**
  * WordPress dependencies
  */
 import { createContext, useCallback, useMemo } from '@wordpress/element';
@@ -12,13 +18,11 @@ import * as Styled from './styles';
 import type {
 	CustomSelectContext as CustomSelectContextType,
 	CustomSelectStore,
-	CustomSelectButtonInternalProps,
 	CustomSelectButtonProps,
 	CustomSelectButtonSize,
 	_CustomSelectInternalProps,
 	_CustomSelectProps,
 } from './types';
-import type { WordPressComponentProps } from '../context';
 import InputBase from '../input-control/input-base';
 import SelectControlChevronDown from '../select-control/chevron-down';
 
@@ -52,14 +56,10 @@ const CustomSelectButton = ( {
 	store,
 	...restProps
 }: Omit<
-	WordPressComponentProps<
-		CustomSelectButtonInternalProps &
-			CustomSelectButtonProps &
-			CustomSelectButtonSize &
-			CustomSelectStore,
-		'button',
-		false
-	>,
+	React.ComponentProps< typeof Ariakit.Select > &
+		CustomSelectButtonProps &
+		CustomSelectButtonSize &
+		CustomSelectStore,
 	'onChange'
 > ) => {
 	const { value: currentValue } = store.useState();
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -573,7 +573,11 @@ describe.each( [
 
 			await press.ArrowDown();
 			await press.ArrowDown();
-			expect( screen.queryByRole( 'listbox' ) ).not.toBeInTheDocument();
+			expect(
+				screen.queryByRole( 'listbox', {
+					name: legacyProps.label,
+				} )
+			).not.toBeInTheDocument();
 
 			expect( currentSelectedItem ).toHaveTextContent(
 				legacyProps.options[ 2 ].name
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -557,6 +557,28 @@ describe.each( [
 			expect( currentSelectedItem ).toHaveTextContent( 'amber' );
 		} );
 
+		it( 'Can change selection with a focused input and closed dropdown while pressing arrow keys', async () => {
+			render( <Component { ...legacyProps } /> );
+
+			const currentSelectedItem = screen.getByRole( 'combobox', {
+				expanded: false,
+			} );
+
+			await press.Tab();
+			expect( currentSelectedItem ).toHaveFocus();
+			expect( currentSelectedItem ).toHaveTextContent(
+				legacyProps.options[ 0 ].name
+			);
+
+			await press.ArrowDown();
+			await press.ArrowDown();
+			expect( screen.queryByRole( 'listbox' ) ).not.toBeInTheDocument();
+
+			expect( currentSelectedItem ).toHaveTextContent(
+				legacyProps.options[ 2 ].name
+			);
+		} );
+
 		it( 'Should have correct aria-selected value for selections', async () => {
 			render( <Component { ...legacyProps } /> );
 
--- a/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
@@ -204,14 +204,14 @@ describe.each( [
 		await press.Enter();
 		expect(
 			screen.getByRole( 'listbox', {
-				name: 'label!',
+				name: legacyProps.label,
 			} )
 		).toBeVisible();
 
 		await press.Escape();
 		expect(
 			screen.queryByRole( 'listbox', {
-				name: 'label!',
+				name: legacyProps.label,
 			} )
 		).not.toBeInTheDocument();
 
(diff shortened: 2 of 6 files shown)


ciampo:
Hey @t-hamano , if I'm not mistaken you use a Windows machine regularly. I have found this comment in the V1 component:

https://github.com/WordPress/gutenberg/blob/b76de86ffe89b437808a7a56c48021838fcb2d8a/packages/components/src/custom-select-control/index.js#L26-L29

Are you aware of this behavior? Would it be ok if, in the V2 (non legacy) version of the component, we relied on the arrow down key to open the select popover?

Thank you 🙏

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -558,6 +558,28 @@ describe.each( [
...
+
+			await press.Tab();
+			expect( currentSelectedItem ).toHaveFocus();
+			expect( currentSelectedItem ).toHaveTextContent(
+				legacyProps.options[ 0 ].name
+			);
+
+			await press.ArrowDown();
+			await press.ArrowDown();
+			expect( screen.queryByRole( 'listbox' ) ).not.toBeInTheDocument();
```
tyxla: Should we add some precision to the query, like `{ name: 'label!' }` or `{ name: legacyProps.label }` for example?

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
```diff
@@ -558,6 +558,28 @@ describe.each( [
...
 		} );
 
+		it( 'Can change selection with a focused input and closed dropdown while pressing arrow keys', async () => {
+			render( <Component { ...legacyProps } /> );
+
+			const currentSelectedItem = screen.getByRole( 'combobox', {
+				expanded: false,
+			} );
+
+			await press.Tab();
```
tyxla, line 568: Historically, those `press.Tab()` needed some `await sleep();` to prevent flakiness. Should we keep those just in case, for consistency at least, if not for anything else? Then if we decide to remove them, we'd remove them all in a single sweep.

packages/components/src/custom-select-control-v2/test/index.tsx
```diff
@@ -416,4 +416,23 @@ describe.each( [
...
+		} );
+
+		await press.Tab();
+		expect( currentSelectedItem ).toHaveFocus();
+		expect( currentSelectedItem ).toHaveTextContent( items[ 0 ].value );
+
+		await press.ArrowDown();
+		expect(
+			screen.getByRole( 'listbox', {
+				name: 'label!',
```
tyxla: Nit: should we use:

```suggestion
				name: defaultProps.label,
```

like we do for other text fixtures in tests?

packages/components/src/custom-select-control-v2/test/index.tsx
```diff
@@ -416,4 +416,23 @@ describe.each( [
...
 	} );
+
+	it( 'Should open the select popover when focussing the trigger button and pressing arrow down', async () => {
+		render( <Component { ...defaultProps } /> );
+
+		const currentSelectedItem = screen.getByRole( 'combobox', {
+			expanded: false,
+		} );
+
+		await press.Tab();
```
tyxla, line 428: Same as above with regards to including a preceding `await sleep();`.

packages/components/src/custom-select-control/test/index.js
```diff
@@ -542,6 +542,30 @@ describe.each( [
...
+
+			await user.tab();
+			expect( currentSelectedItem ).toHaveFocus();
+			expect( currentSelectedItem ).toHaveTextContent(
+				props.options[ 0 ].name
+			);
+
+			await user.keyboard( '{arrowdown}' );
+			await user.keyboard( '{arrowdown}' );
+			expect( screen.queryByRole( 'listbox' ) ).not.toBeInTheDocument();
```
tyxla: Same as above with regards to precision (adding a `name` to the query).

tyxla (commented):
Looks good overall 👍 

We still have to merge https://github.com/WordPress/gutenberg/pull/62907 and rebase, but left some feedback in the meantime.

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
ciampo: Done in https://github.com/WordPress/gutenberg/pull/62919/commits/5f0c596bece0c1f17ecfa3a1c216565ced6aafa9

packages/components/src/custom-select-control/test/index.js
ciampo: https://github.com/WordPress/gutenberg/pull/62919/commits/5f0c596bece0c1f17ecfa3a1c216565ced6aafa9

packages/components/src/custom-select-control-v2/legacy-component/test/index.tsx
ciampo: Yup, done in https://github.com/WordPress/gutenberg/pull/62919/commits/5b1bdaf1fde5eab85bb3aa6b6d7119ccd24ab85a

packages/components/src/custom-select-control-v2/test/index.tsx
ciampo: https://github.com/WordPress/gutenberg/pull/62919/commits/5b1bdaf1fde5eab85bb3aa6b6d7119ccd24ab85a

packages/components/src/custom-select-control-v2/test/index.tsx
ciampo: Done in this specific instance and throughout the three suites of tests https://github.com/WordPress/gutenberg/pull/62919/commits/c2f7c5917ee7e4b78f6a628ae9993cdce60295f6

packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -52,7 +53,10 @@ const CustomSelectButton = ( {
...
 }: Omit<
 	WordPressComponentProps<
-		CustomSelectButtonProps & CustomSelectButtonSize & CustomSelectStore,
+		CustomSelectButtonInternalProps &
+			CustomSelectButtonProps &
+			CustomSelectButtonSize &
+			CustomSelectStore,
 		'button',
 		false
 	>,
```
mirka: Somewhat tangential, but maybe this should've composed from `Ariakit.Select` props rather than `<button>` props. Then we wouldn't need to add `CustomSelectButtonInternalProps` definitions for whatever other Ariakit props we want to pass. (No strong opinion, but probably cleaner in the long run.)

```diff
diff --git a/packages/components/src/custom-select-control-v2/custom-select.tsx b/packages/components/src/custom-select-control-v2/custom-select.tsx
index 144ccda7af..0520e392b9 100644
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -1,3 +1,9 @@
+/**
+ * External dependencies
+ */
+// eslint-disable-next-line no-restricted-imports
+import type * as Ariakit from '@ariakit/react';
+
 /**
  * WordPress dependencies
  */
@@ -12,13 +18,11 @@ import * as Styled from './styles';
 import type {
 	CustomSelectContext as CustomSelectContextType,
 	CustomSelectStore,
-	CustomSelectButtonInternalProps,
 	CustomSelectButtonProps,
 	CustomSelectButtonSize,
 	_CustomSelectInternalProps,
 	_CustomSelectProps,
 } from './types';
-import type { WordPressComponentProps } from '../context';
 import InputBase from '../input-control/input-base';
 import SelectControlChevronDown from '../select-control/chevron-down';
 
@@ -52,14 +56,10 @@ const CustomSelectButton = ( {
 	store,
 	...restProps
 }: Omit<
-	WordPressComponentProps<
-		CustomSelectButtonInternalProps &
-			CustomSelectButtonProps &
-			CustomSelectButtonSize &
-			CustomSelectStore,
-		'button',
-		false
-	>,
+	React.ComponentProps< typeof Ariakit.Select > &
+		CustomSelectButtonProps &
+		CustomSelectButtonSize &
+		CustomSelectStore,
 	'onChange'
 > ) => {
 	const { value: currentValue } = store.useState();

```

t-hamano:
> Are you aware of this behavior? Would it be ok if, in the V2 (non legacy) version of the component, we relied on the arrow down key to open the select popover?

This is the first time I've seen this code, but it looks like it's intended to simulate the behavior of the native select element in Windows OS.

On Windows OS, the native select elements behave as follows:

- If the dropdown is not open, the up and down arrow keys change the option but do not open the dropdown.
- The `enter` key opens the dropdown.
- If the dropdown is open, the up and down arrow keys change the option.
- The `Escape` key closes the dropdown.

https://github.com/WordPress/gutenberg/assets/54422211/805af1d1-e541-4540-adcd-da6549d3a31d

This behavior is the same for `CustomSelectControl`, `CustomSelectControl v2` (legacy), and `CustomSelectControl v2` (non-legacy).

On the other hand, in this PR, the up and down arrow keys always open the dropdown.

https://github.com/WordPress/gutenberg/assets/54422211/50547f36-a1d0-4cec-a586-168f8bc626e1

I'm concerned that behavior that differs from the Windows OS standard may be a bit confusing for keyboard users. I don't really understand the history of this PR being submitted, so sorry if I'm off the mark 🙏

ciampo:
Thank you, @t-hamano ! Always appreciate your help :D 

> On Windows OS, the native select elements behave as follows:

It seems that there isn't a standard behavior across OSs — on MacOS, pressing arrow down on a closed `select` element opens the dropdown, while on Windows it changes option without opening the dropdown.

Looking into the [WAI-ARIA APG](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/#kbd_label), it looks like the recommended behavior is to open the dropdown when pressing arrow down:

| Key | Function |
|---|---|
| Down&nbsp;Arrow | Opens the listbox if it is not already displayed without moving focus or changing selection. DOM focus remains on the combobox. |

> I'm concerned that behavior that differs from the Windows OS standard may be a bit confusing for keyboard users.

That is a valid concern, but unfortunately, one that we can't easily solve across all OSs unless we try to sniff the OS (which I'd like to avoid).

In this scenario, I'd suggest we follow the WAI-ARIA APG — that would mean that, in `CustomSelectControl` V2, pressing arrow down would open the dropdown.

packages/components/src/custom-select-control-v2/custom-select.tsx
ciampo: Good suggestion — applied in https://github.com/WordPress/gutenberg/pull/62919/commits/3ed4327d98855f41bed5a19000d9dcd4aa9c37b6

t-hamano:
> In this scenario, I'd suggest we follow the WAI-ARIA APG — that would mean that, in "CustomSelectControl" V2, pressing arrow down would open the dropdown.

I see, if that’s the case then I think it’s okay 👍

ciampo:
Hey folks, I think this PR should be ready for a final round of review, if we agree on the approach taken

packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -1,3 +1,9 @@
+/**
+ * External dependencies
+ */
+// eslint-disable-next-line no-restricted-imports
```
tyxla, line 4: 🤔 Shouldn't we allow Ariakit imports within the components package? Sounds like it could be an exception to the ESLint rule.

tyxla (approved):
Working as expected and the code looks good 👍 🚀 

From what I understand, @t-hamano agrees with following the WAI-ARIA recommendations, so I believe this is good to ship 🚢

packages/components/src/custom-select-control-v2/custom-select.tsx
ciampo: https://github.com/WordPress/gutenberg/pull/63123

PR #62924, opened 2024-06-27
Add justification to block toolbar in addition to sidebar

### LINKED ISSUE #60838: Content Block: add justification to block toolbar in addition to sidebar
**Suggestion:** 
Add justification options to the block toolbar in addition to the settings sidebar.

**Why:** 
- Because justification settings for other blocks (like text blocks) are in the toolbar. I overlooked this one many times, because I expected it to be in the toolbar, not the sidebar.
- And: there's lots of room for it.

image

shreya0204:
## What?

Added justification options to the block toolbar for "type constrained" blocks.

Fixes: #60838 

## Why?

The justification settings for other blocks, such as text blocks, are found in the toolbar. Many users overlook the settings when they are only in the sidebar, expecting them to be in the toolbar. Adding the justification options to the toolbar improves consistency and user experience.

## How?

Added Justification `toolBarControls` of the constrained layout type blocks.

## Testing Instructions

- Add any "type constrained" block, such as Group or Content.
- Check the toolbar to see if the justification option is available there.
- Verify that the justification options are functional and behave as expected.

## Screenshots or screencast 

Screenshot 2024-06-27 at 5 54 32 PM
Screenshot 2024-06-27 at 5 54 58 PM


jameskoster:
I was looking for this option just yesterday, so I'm very happy to see this PR! In terms of design everything looks good to me, and the option seems to be working as advertised 👍

packages/block-editor/src/layouts/constrained.js
```diff
@@ -278,3 +294,72 @@ export default {
...
+					<ToggleGroupControlOptionIcon
+						key={ value }
+						value={ value }
+						icon={ icon }
+						label={ label }
+					/>
+				);
+			} ) }
+		</ToggleGroupControl>
+	);
```
noisysocks: I believe this is dead code as `isToolbar` will always be `true`. Can you simplify `DefaultLayoutJustifyContentControl` and remove the `isToolbar` prop?

jasmussen:
Thanks for the PR!

packages/block-editor/src/layouts/constrained.js
shreya0204: Thank you for pointing this out @noisysocks . I have removed the `isToolbar` props and simplified the component.

t-hamano (commented):
Thanks for the PR!

I think we need to check that `allowJustification` is enabled in addition to `allowSwitching`.

For example, suppose we make the following change to the Group block to disable `allowJustification`:

```diff
diff --git a/packages/block-library/src/group/block.json b/packages/block-library/src/group/block.json
index db7d09c55d..ef578cff97 100644
--- a/packages/block-library/src/group/block.json
+++ b/packages/block-library/src/group/block.json
@@ -86,7 +86,8 @@
                        }
                },
                "layout": {
-                       "allowSizingOnChildren": true
+                       "allowSizingOnChildren": true,
+                       "allowJustification": false
                },
                "interactivity": {
                        "clientNavigation": true
```

In this case, even though the justification control will not be displayed in the sidebar, the UI will unintentionally be displayed in the toolbar:

allow-justification

I think we probably need code like this:

```javascript
const { allowSwitching, allowJustification = true } =
	layoutBlockSupport;
if ( allowSwitching || ! allowJustification ) {
	return null;
}
```

shreya0204:
Thank you for the feedback @t-hamano 

I've updated the implementation to include checks for both `allowSwitching` and `allowJustification` as you suggested.

Please let me know if this adjustment aligns with the expected functionality, or if there are any other modifications or improvements you would recommend.

t-hamano:
Thanks for the update!

I feel this PR is okay to ship because it improves UI consistency, but I am a bit skeptical about whether this UI is really necessary in the Group variations (constrained layout) in the first place. Maybe there is another approach to remove this UI from the block sidebar.

What do you think, @tellthemachines?

tellthemachines:
> I am a bit skeptical about whether this UI is really necessary in the Group variations (constrained layout) in the first place.

Given that flex blocks also have the justification controls in both sidebar and block toolbar, I think it's fine to have them in both places here. If that did change, I'd expect it to change across all the layout types.

packages/block-editor/src/layouts/constrained.js
```diff
@@ -146,8 +147,25 @@ export default {
...
-		return null;
+	toolBarControls: function DefaultLayoutToolbarControls( {
+		layout = {},
+		onChange,
+		layoutBlockSupport,
+	} ) {
+		const { allowSwitching, allowJustification = true } =
+			layoutBlockSupport;
+
+		if ( allowSwitching || ! allowJustification ) {
```
tellthemachines: Any reason we're checking for `allowSwitching` here? It's pretty much a legacy attribute, because the controls it exposes when enabled haven't been maintained since they were added. No harm in leaving it here though.

t-hamano: My guess is that the pull request was created based on the code here:

https://github.com/WordPress/gutenberg/blob/16420c4ec6c6868a10d08e46b08612cd726e776b/packages/block-editor/src/layouts/flex.js#L98-L100

@shreya0204 Can you refactor it as follows?

```javascript
const { allowJustification = true } = layoutBlockSupport;
if ( ! allowJustification ) {
	return null;
}
```

shreya0204: @t-hamano Certainly! I've refactored the code as requested. Let me know if it needs any further adjustments. Thanks!

t-hamano (approved):
LGTM 🚀

Issue #54305, opened 2023-09-08
@wordpress/scripts: Improve TypeScript support in `lint-js`


Chrico:
## What problem does this address?

Right now the `@wordpress/scripts` package provides `lint-js` (eslint), `lint-style` (stylelint), `lint-md-docs` (markdownlint) and `lint-pkg-json` (npmPkgJsonLint). 

While tools like `lint-style` via `stylelint` are able to find syntax errors - as an example:

**index.scss**
```scss
foo []
```

```shell
$ yarn wp-scripts lint-style
yarn run v1.22.19

resources/scss/index.scss
 1:1  ✖  Unknown word  CssSyntaxError

1 problem (1 error, 0 warnings)

```

the `lint-js` does not actually check typescript code to detect type errors. As an example:

**index.ts**
```ts
class Foo {
	protected bar: string;
	public constructor( protected value: string ) {
		this.bar = 'bar';
	}

	public getValue = (): string => {
		return this.value;
	};
}
new Foo( 1 );
```

```shell
$ yarn wp-scripts lint-js
yarn run v1.22.19
Done in 2.10s.
```

Running `tsc` with `--noEmit` will detect an error:

```shell
 tsc --noEmit --skipLibCheck resources/ts/*

resources/ts/index.ts:14:10 - error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.

14 new Foo( 1 );
            ~

Found 1 error in resources/ts/index.ts:14
```

---

## What is your proposed solution?

Introduce a new `lint-ts` script in `@wordpress/scripts` which does lint Typescript code. To do that we can make use of the `tsc` with `--noEmit` to do only the type checking part and not the compilation.

See also: https://www.typescriptlang.org/tsconfig#noEmit

Chrico:
Just wanted to add some "findings". It seems like the `@wordpress/eslint-plugin` mentions support for Typescript

> Finally, this ruleset also includes an optional integration with the [@typescript-eslint/eslint-plugin](https://github.com/typescript-eslint/typescript-eslint) package that enables ESLint to support [TypeScript](https://www.typescriptlang.org/) language. You can activate it by installing the [typescript](https://www.npmjs.com/package/typescript) package separately with:

See here: [gutenberg/packages/eslint-plugin#usage](https://github.com/WordPress/gutenberg/tree/76b0165002690569cc693fefc6b3e94e7e09204a/packages/eslint-plugin#usage)

[...]

PR #27143, opened 2020-11-20
Scripts: Add TypeScript support to linting command

sirreal:
## Description

With some TypeScript in Gutenberg, .ts and .tsx (TypeScript) files should undergo the similar formatting and linting to `.js` files.

## How has this been tested?

Change a .ts file, stage, and commit it.
Work on `@wordpress/react-i18n` and see that lint issues are reported correctly.

## Types of changes

Breaking change: Add lint setup and rules for TypeScript files.

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR.

PR #62925, opened 2024-06-27
Scripts: Ensure that typescript-eslint checks for unused vars

Chrico:
## What?

`'no-unused-vars': 'off'` in eslint-plugin/configs/recommanded.js will also disable the Typescript eslint when running `wp-scripts lint-js`.

## Why?
eslint/no-unused-vars got disabled when introducing support for TypeScript with #27143 (2-3 years ago) when Gutenberg was moving to Typescript. By disabling this feature, the Typescript eslint will not be executed. We actively enable "no-unused-vars" and "@typescript-eslint/no-unused-vars" and additionally setting "ignoreRestSiblings" to "true", which is actively used in Gutenberg.

See more information in: #54305

## Testing Instructions
Create a new file: 

**index.ts**

```ts
class Foo {
	protected bar: string;
	public constructor( protected value: string ) {
		this.bar = 'bar';
	}

	public getValue = (): string => {
		return this.value;
	};
}
new Foo( 1 );
```

with **package.json**:

```json
{
    "name": "tsc-eslint",
    "version": "1.0.0",
    "license": "GPL-2.0-or-later",
    "devDependencies": {
        "@wordpress/scripts": "^25.0.0",
        "typescript": "^4.9.4"
    }
}
```

and execute on cli:

```shell
$ wp-scripts lint-js *.{ts,tsx}
```

**before** this fix the Typescript eslint did not catch any type errors:

```shell
Done in 1.83s.
```

**after**  this fix the Typescript eslint will match errors:

```shell
/{snip}/index.ts
  5:32  error  'value' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```


gziolo:
I see that changes work as CI reports many linting errors that would have to be fixed:

Screenshot 2024-06-28 at 08 27 25

The question is whether both rules should be enabled at the same time, as they often report the same issues. I don't know the details but at least for the Gutenberg repository `no-unused-var` is reporting violations way often.

Chrico:
@gziolo 
The `@typescript-eslint/no-unused-vars` extends the `eslint/no-unused-vars`. I guess we need to change it to following:

```js
{
   'no-unused-vars': 'off',
   '@typescript-eslint/no-unused-vars': [ 'error', { ignoreRestSiblings: true } ]
}
```

At least, this is the recommended way in the docs https://typescript-eslint.io/rules/no-unused-vars/#how-to-use ;-)

gziolo:
There are still 24 errors reported, way better but it probably reports unused variables that need to be put in the function callbacks:

Screenshot 2024-06-28 at 11 06 46

Chrico:
Should we ( I ?) also fix the problems which now occur or is this a "follow-up" Issue/PR? :-) Probably, depending on how we want to continue, some code might needs to be touched and/or we need to add some `// eslint-disable-next-line @typescript-eslint/no-unused-vars` ?!

gziolo:
It needs to be evaluated before landing whether these issues are correctly highlighted and the should get fixed. Alternatively, maybe there are some config options that could relax these checks to pass the validation.

Chrico:
The new reported errors look really like a quick win. Maybe just needs some jep/nope from you (?). :-)

Quickly went through all items:

----

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/autocomplete/index.tsx#L16
--> Could be removed, it is not used in that file.

----

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/border-box-control/border-box-control-visualizer/component.tsx#L4
--> Could be removed, it is not used in that file.

---

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/date-time/date-time/index.tsx#L10
--> Could be removed, it is not used in that file.

----

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/font-size-picker/utils.ts#L4
--> Could be removed, it is not used in that file.

---

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/palette-edit/test/index.tsx#L19

`_` is unused, probably could be rewritten into:

```js
[ ...Array( input.value.length ) ].forEach( async () => await press.Backspace() )
```

----

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/private-apis.ts#L4
--> Could be removed, it is not used in that file.

----

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/theme/stories/index.story.tsx#L64
--> disable for that line?

---

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/components/src/toggle-group-control/toggle-group-control/component.tsx#L12
--> Could be removed, it is not used in that file.

----

https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/compose/src/higher-order/test/compose.ts#L33
https://github.com/WordPress/gutenberg/blob/v18.6.1/packages/compose/src/higher-order/test/pipe.ts#L33

--> Seems like the case `f` is really not used in the following `expect()`. So: Remove that line or add to the `expect()`?

----

gziolo:
Maybe you could commit to all the proposed changes to help decide.

Chrico:
Seems like, every time something is fixed, something new appears. The `wp-scripts lint-js` does not show all errors at once (probably due the hugh amount of warnings). This might can take a bit longer, but probably also needs some help to not break things ;-)

gziolo:
So far so good. 12 errors fixed, 12 more to go.

Chrico:
Here we go @gziolo :) ☕

packages/eslint-plugin/configs/recommended.js
```diff
@@ -55,8 +55,15 @@ if ( isPackageInstalled( 'typescript' ) ) {
 				// Don't require redundant JSDoc types in TypeScript files.
 				'jsdoc/require-param-type': 'off',
 				'jsdoc/require-returns-type': 'off',
-				// Handled by TS itself.
+				// Setting "no-unused-vars" to "off" will also disable
+				// the typescript type checker. So we need to explicitly enable it,
```
sirreal: This comment doesn't look accurate, eslint and the TypeScript compiler are completely separate. Nothing configured here should affect the compiler.

There _is_ overlap between some tsconfig, notably [`noUnusedLocals`](https://www.typescriptlang.org/tsconfig/#noUnusedLocals) and [`noUnusedParameters `](https://www.typescriptlang.org/tsconfig/#noUnusedParameters):

https://github.com/WordPress/gutenberg/blob/d1987d09632968b59ac52e1d259dd16591ac7dda/tsconfig.base.json#L21-L22

Those tsconfig settings are why this would be disabled.

I think it would be better to replace this entire comment with something like:

> Use eslint for unused variable and parameter detection. This overlaps with TypeScript `noUnusedLocals` and `noUnusedParameters` settings. TypeScript may only run on a subset of files. Prefer eslint which is more likely to run on the entire codebase.

---

And since we're turning on this rule, let's also remove those "unused" settings I linked above from our tsconfig in this PR. There's no reason to report the same thing multiple times.

test/performance/config/performance-reporter.ts
```diff
@@ -118,7 +118,7 @@ export function curateResults(
 	return Object.fromEntries(
 		Object.entries( output )
 			// Reduce the output to contain taken metrics only.
-			.filter( ( [ _, value ] ) => value !== undefined )
+			.filter( ( item ) => item[ 1 ] !== undefined )
```
sirreal: Let's use a more minimal change here, we can remove the unused `_` binding:

```suggestion
			.filter( ( [ , value ] ) => value !== undefined )
```

packages/components/src/palette-edit/test/index.tsx
```diff
@@ -16,9 +16,9 @@ async function clearInput( input: HTMLInputElement ) {
 	await click( input );
 
 	// Press backspace as many times as the input's current value
-	for ( const _ of Array( input.value.length ) ) {
-		await press.Backspace();
-	}
+	[ ...Array( input.value.length ) ].forEach(
+		async () => await press.Backspace()
+	);
```
sirreal: Let's avoid this change and instead configure the rule to allow `_`-prefixed variables.

We don't want to be rewriting valid code to appease the linter. Please revert this change.

packages/eslint-plugin/configs/recommended.js
```diff
@@ -55,8 +55,15 @@ if ( isPackageInstalled( 'typescript' ) ) {
...
+				// but want to ignore "unused rest siblings".
+				// @issue https://github.com/WordPress/gutenberg/issues/54305
 				'no-unused-vars': 'off',
+				'@typescript-eslint/no-unused-vars': [
+					'error',
+					{ ignoreRestSiblings: true },
```
sirreal, line 65: I flagged a valid use case in another comment. Let's allow `_`-prefixed unused vars. There appears to be configuration for vars, destructuring, args, etc. Let's start with vars, I think that should be sufficient to allow our valid code case:

```json
{
"varsIgnorePattern": "^_",
}
```

sirreal (changes requested):
Thanks for working on this, good spot! I've left a few suggestions I'd like to see changed before this moves ahead.

[Will you also add a **Breaking Changes** changelog entry to the eslint-plugin package](https://github.com/WordPress/gutenberg/blob/trunk/packages/README.md#maintaining-changelogs)?

packages/components/src/palette-edit/test/index.tsx
Chrico: The problem is not, that the `_` variables are not allowed, the problem is that the `_` is not used in the `for`-block. So that would mean, that I need to add a `eslint-disable` here. Are you agreeing with that?

packages/eslint-plugin/configs/recommended.js
Chrico: I don't think, that this pattern will make anything easier. I just created a simple test file with some cases where `_` can be present in a variable, argument or import:

```ts
/**
 * WordPress dependencies
 */
import {
	__experimentalAlignmentMatrixControl,
	ToggleControl as _,

const myFunc = ( _, _bar ): void => {
	console.log( 'huhu' );
};

const myFunc2 = (): void => {
	for ( const _ of Array( 5 ) ) {
		console.log('huhu');
	}
};
```

and I've added following to the configuration:

```ts
{
  // snip
'no-unused-vars': 'off',
'@typescript-eslint/no-unused-vars': [
	'error',
	{ ignoreRestSiblings: true, varsIgnorePattern: "^_", argsIgnorePattern: "^_" },
],
}
```

The `@typescript-eslint` will completely ignore everything starting with `_` from being "unused".

packages/eslint-plugin/configs/recommended.js
Chrico: updated here: https://github.com/WordPress/gutenberg/pull/62925/commits/94ed5a3a1ded56e0054c524119d5a99e7ed1263e and here: https://github.com/WordPress/gutenberg/pull/62925/commits/92944da70459db66ffc74136d4814e2c10fdb5fc

packages/eslint-plugin/configs/recommended.js
sirreal: You're absolutely right. How about a smarter regex that's aware of localization functions and experimental names? Then we can use `_` —a common convention— but disallow these known values that shouldn't be valid ignores. We should cover `__unstable` as well:

`^_(?!_$|n$|x$|nx$|_experimental|_unstable)`

sirreal:
[Let's also remove the `noUnused*` config from all the package tsconfigs where it's redundant now](https://github.com/search?q=repo%3AWordPress%2Fgutenberg%20noUnused%20language%3A%22JSON%20with%20Comments%22&type=code).

packages/eslint-plugin/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- `@typescript-eslint/no-unused-vars` is now turned on with level "error" when running `wp-scripts lint-js`.
```
sirreal:
```suggestion
- Add [`@typescript-eslint/no-unused-vars` rule](https://typescript-eslint.io/rules/no-unused-vars) to the recommended TypeScript ruleset ([#62925](https://github.com/WordPress/gutenberg/pull/62925)).
```

packages/eslint-plugin/configs/recommended.js
Chrico: We can do that, but by adding this, we still face the problem, that unused vars (which match that regex) are not detected. So by doing:

```ts
import { __experimentalAlignmentMatrixControl } from '@wordpress/components';
```

and not use it, it will be fine. The whole point for this propsal is to enable "unused-vars"-detection and actively use it. 👀 

Is it just about the rewritten code here https://github.com/WordPress/gutenberg/pull/62925#discussion_r1660803695 ? :-) We could also do following as a "rewrite" to replace for `for-of`-loop:

```ts
for( let i = 0; i < input.value.length, i++) {
	await press.Backspace()
}
```

sirreal: Fair enough. Since we're regularly using values that start with `_`, it's not a good ignore prefix.

I remain strongly opposed to rewriting perfectly valid and correct code in order to appease the linter. In the valid cases that are flagged here, let's just add an eslint disable comment, that's what they're for 🙂

packages/components/src/palette-edit/test/index.tsx
sirreal: Let's revert this and add the appropriate disable comment.

sirreal (commented):
Thanks for the thoughtful discussion. I'm ready to approve and land this pending two changes:
- Revert and eslint-disable this valid case: https://github.com/WordPress/gutenberg/pull/62925#discussion_r1665802945
- Remove other tsconfig `noUnused*`: https://github.com/WordPress/gutenberg/pull/62925#issuecomment-2200677236

packages/components/src/palette-edit/test/index.tsx
Chrico: As discussed, reverted the change here https://github.com/WordPress/gutenberg/pull/62925/commits/2b21b06932ab2f53944da3fed7b7df9eed0504b5 and added "eslint-disable".

Chrico:
> * Revert and eslint-disable this valid case: [Scripts: Ensure that typescript-eslint checks for unused vars #62925 (comment)](https://github.com/WordPress/gutenberg/pull/62925#discussion_r1665802945)

Change was done here: https://github.com/WordPress/gutenberg/pull/62925/commits/2b21b06932ab2f53944da3fed7b7df9eed0504b5

> * Remove other tsconfig `noUnused*`: [Scripts: Ensure that typescript-eslint checks for unused vars #62925 (comment)](https://github.com/WordPress/gutenberg/pull/62925#issuecomment-2200677236)

Updates were done here: https://github.com/WordPress/gutenberg/pull/62925/commits/93da83f82eb7fa5b25f80f7b0782cc4f0a877032

sirreal (approved):
Thanks, nice work on this one @Chrico.

gziolo:
Thank you both for taking it to the finish line 🥇

PR #57540, opened 2024-01-04
Media & Text: Replace the deprecated __experimentalImageSizeControl with ResolutionTool

carolinan:
## What?

ImageSizeControl (__experimentalImageSizeControl) is used in the Media & text block. It was marked as deprecated in [#51545](https://github.com/WordPress/gutenberg/pull/51545/files#diff-5a9a2234d89e4080e031dbc85aafab7b248af84799f65710ad15ca7ca03bd5f1R34-R38).

This PR replaces __experimentalImageSizeControl with the ResolutionTool.

Partial for https://github.com/WordPress/gutenberg/issues/52223

## Why?

Deprecated controls should not be used.

## How?

Adds a new tools panel and moves the existing block sidebar options to this panel.
Replaces __experimentalImageSizeControl with the ResolutionTool.

### Testing Instructions for Keyboard

Deactivate Gutenberg.
Add two new media and text blocks with images. One with default options, one with the "Crop image to fill entire column" option enabled and a non default resolution.
Save.

Activate Gutenberg with the PR applied. Refresh the page and confirm that:
-  There are no block validation errors.
- The options use a tools panel.
- All options, including the resolution, works.
- The reset and the "reset all" works.

Save and view the front of the website. Confirm that the blocks display correctly, and the correct image size is loaded.

## Screenshots or screencast 
Before:
Media & text block with the settings panel opened, before the PR

After:
Media & text block with the settings panel opened, with the PR, using the tools item panels

Media & text block with the options panel opened.

Issue #54858, opened 2023-09-27
Image lightbox: configure lightbox settings in `theme.json` at the top-level


oandregal:
### Description

The lightbox feature can be controlled via `theme.json` with two settings: `allowEditing` to hide/show the UI panel) and `enabled` (to control whether it is enabled or not).

While configuring the value of these settings at the block-level works as expected, doing the same at the top-level does not work.

### Step-by-step reproduction instructions

- Open the post editor, add one image, and publish the post.

**BLOCK LEVEL. Test the UI can be disabled and the feature enabled (works as expected):**

- Edit the `theme.json` of the active theme such as:

```json
{
  "settings": {
    "blocks": {
      "core/image": {
        "lightbox": {
          "allowEditing": false,
          "enabled": true
        }
      }
    }
  }
}
```

The expected behaviour is that:

- the lightbox feature is not visible in the post editor or in the site editor
- the lightbox feature is enabled in the front (the image expands upon clicking)

**TOP-LEVEL. Test the UI can be disabled at the feature enabled (does not work):**

- Edit the `theme.json` of the active theme such as:

```json
{
  "settings": {
    "lightbox": {
      "allowEditing": false,
      "enabled": true
    }
  }
}
```

The expected behaviour is that:

- the lightbox feature is not visible in the post editor or in the site editor
- the lightbox feature is enabled in the front (the image expands upon clicking)

### Screenshots, screen recording, code snippet

This is the UI for the lightbox feature, when is visible:

| Post editor (block inspector of the image block) | Site editor (global styles sidebar > blocks > image) |
| --- | --- |
| <img width="279" alt="Captura de ecrã 2023-09-27, às 12 30 07" src="https://github.com/WordPress/gutenberg/assets/583546/b311f785-ca84-4474-8d6c-85ebab9d79d6"> | <img width="279" alt="Captura de ecrã 2023-09-27, às 12 31 58" src="https://github.com/WordPress/gutenberg/assets/583546/0441b137-dac3-48c9-86c8-8fe38c862cdc"> |

### Environment info

WordPress 6.4 beta1, Gutenberg trunk (16.7.0-rc.2), TwentyTwentyThree theme. I could replicate the issue with and without Gutenberg enabled.

oandregal:
cc @artemiomorales @michalczaplinski

artemiomorales:
Duplicate of https://github.com/WordPress/gutenberg/issues/54544

michalczaplinski:
[...]

Issue #52507, opened 2023-07-11
Option to choose image size in Cover and Image block


pedjas:
## What problem does this address?

When Featured image is chosen for Post, full size image is linked.

When Cover or Image block is inserted in post or template and set to use Featured image, full size image is used. This happens even if actual view of image is much smaller.

For Image block there is option to additionally choose image size used so user must be aware that his site unnecessary loads full size images just shows them small and use this option to change that.

For cover block there is not option to choose image size. Full image size is used each time.

This is huge waste of resources on site. The first rule of using images is to reduce their pixel resolution to match actual size that should be displayed, not to load large image and reduce display size using CSS.

## What is your proposed solution?

User interface should warns users on this and help them easily notice discrepancy in physical and displayed image size and fix that.

For Cover images there should be option for user to choose size of image.

hanneslsm:
I agree - I also have proposed that even for regular images the correct image size should be used automatically. Somehow same as for the cover block, as you described.
This is part of the Phase 3: https://make.wordpress.org/core/2023/07/07/media-library/
Since Phase 3 just has started I don't want you to have high hopes to see this implemented very fast. However, it's on the list.

pedjas:
> I agree - I also have proposed that even for regular images the correct image size should be used automatically. Somehow same as for the cover block, as you described. This is part of the Phase 3: https://make.wordpress.org/core/2023/07/07/media-library/ Since Phase 3 just has started I don't want you to have high hopes to see this implemented very fast. However, it's on the list.

This is great!

Luke-SF:
I don't see the option for image block to change from "large" to "full". What code enables that in the theme/Wordpress?

annezazu:
Thanks for opening this. Good news! I'm closing this out both because it is a duplicate of https://github.com/WordPress/gutenberg/issues/19223 and https://github.com/WordPress/gutenberg/issues/58819 and because this option was added for the Cover block in this PR: https://github.com/WordPress/gutenberg/pull/62926

PR #24795, opened 2020-08-25
Add Image Size dropdown for Media & Text block

david-szabo97:
Closes: https://github.com/WordPress/gutenberg/issues/24688

## Description
Add an Image Size dropdown for Media & Text block. We can change the size of the image for the Image block, but we didn't have the option to do the same for Media & Text block. Which caused the Media & Text block to always render the full-size image.

**NOTE:** Default value for Image Size is **Full**. I think Large would be better in general, but that would break existing posts.

## How has this been tested?
1. Open Editor
2. Add Media & Text Block
3. Select an **image** from the media library
4. Make sure Media & Text Block is in focus
5. You should see an Image Size dropdown in the right sidebar

## Screenshots
image

## Types of changes
New feature (non-breaking change which adds functionality) 

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR.

PR #62926, opened 2024-06-27
Cover Block: Add Image Resolution options 

### LINKED ISSUE #58819: Cover Block: Missing Image Resolution options
Some client sites have very large, high resolutions images uploaded, and would be very useful if when using the Cover block, users had the option to specify which resolution to use, similar to how the Media & Text or Image blocks currently work.

This will significantly reduce page load if the appropriate image size is used.

This is issue might also affect other blocks making use of media elements, but the cover block is the only one that's currently relevant at the moment.

Implement the [`ImageSizeControl`](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/image-size-control) or [`ResolutionTool`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/resolution-tool/index.js) component within the block just like the Media & Text or Image block currently do.

akasunil:
## What?

Fixes: https://github.com/WordPress/gutenberg/issues/58819

## Why?

Resolution option were missing from cover block for background images

## How?

This PR add `ResolutionTool` to cover block, when cover has background image

## Testing Instructions

- Add cover block
- Set image as background in cover block
- Go to block setting in sidebar, Look for resolution control.

## Screenshots or screencast 

[cover-block-image-resolution-control.webm](https://github.com/user-attachments/assets/1703631b-428c-481d-b8a0-866d340e5ef2)


akasunil:
@t-hamano could use review here.

richtabor:
If we do this, it should go in the "Settings" panel, like the Image block does.

akasunil:
@richtabor If we have to transfer `Resolution Dropdown` to the block setting panel, we may need to refactor setting panel with the `ToolsPanel` component.

I see you've raised an issue (https://github.com/WordPress/gutenberg/issues/64703) to provide background support to the cover block as a popover, like we do with the group block. If we're going to incorporate it in the cover block, I believe we should also include a resolution option in the background support. What do you think?

t-hamano:
> If we have to transfer "Resolution Dropdown" to the block setting panel, we may need to refactor setting panel with the "ToolsPanel" component.

This is correct. I also think that the "Resolution" should be in the Settings panel, but since the `ResolutionTools` component is composed of `ToolsPanelItem`, it must be placed inside the `ToolsPanel` component. The current Settings panel does not use the `ToolsPanel` component, so the entire settings panel will need to be refactored.

> an issue (https://github.com/WordPress/gutenberg/issues/64703) to provide background support to the cover block as a popover, like we do with the group block.

This is something to think about carefully, because `background` support is **disabled by default**.

For example, imagine a theme that doesn't have `apperanceTools` enabled. The Cover block has background functionality by default, but if we migrate background functionality to block support, those themes will no longer have background UI. This means forcing theme developers to add `settings.apperanceTools` or `settings.background`.

Perhaps a solution would be to enable background support by default _only_ for the Cover block, but this would create another problem: it would override the root-level background support. A similar issue was reported in [this comment](https://github.com/WordPress/gutenberg/issues/54858#issuecomment-1737845464).

This is similar to how the `minHeight` attribute of the Cover block and the `aspectRatio` attribute of the Image block cannot be easily migrated to block support.

Either way, I think the best thing to do at this point is to refactor the Settings panel to use `ToolsPanel` component.

#57540 might help.

akasunil:
Thank you @t-hamano, I had already started with implementation. But, after seeing the issue that @richtabor had created, I got concerned that I might have to close this PR. :smile:

Now that you have explained it, I'll continue with development. Thank you for reference PR.

akasunil:
@t-hamano could use your review here.

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -160,72 +195,139 @@ export default function CoverInspectorControls( {
...
 							<>
-								<ToggleControl
-									__nextHasNoMarginBottom
+								<ToolsPanelItem
 									label={ __( 'Fixed background' ) }
-									checked={ hasParallax }
-									onChange={ toggleParallax }
-								/>
+									isShownByDefault
+									hasValue={ () => !! hasParallax }
```
t-hamano: `hasParallax` is a boolean so I don't think typecasting is necessary.

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -160,72 +195,139 @@ export default function CoverInspectorControls( {
...
 	return (
 		<>
 			<InspectorControls>
 				{ !! url && (
-					<PanelBody title={ __( 'Settings' ) }>
+					<ToolsPanel
+						label={ __( 'Settings' ) }
+						resetAll={ () => {
+							setAttributes( {
+								hasParallax: undefined,
```
t-hamano:
```suggestion
								hasParallax: false,
```
If we do not explicitly specify a default value, a warning error will occur:

image

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -160,72 +195,139 @@ export default function CoverInspectorControls( {
...
+								<ToolsPanelItem
 									label={ __( 'Fixed background' ) }
-									checked={ hasParallax }
-									onChange={ toggleParallax }
-								/>
+									isShownByDefault
+									hasValue={ () => !! hasParallax }
+									onDeselect={ () =>
+										setAttributes( {
+											hasParallax: undefined,
```
t-hamano:
```suggestion
											hasParallax: false,
```

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -160,72 +195,139 @@ export default function CoverInspectorControls( {
...
 
-								<ToggleControl
-									__nextHasNoMarginBottom
+								<ToolsPanelItem
 									label={ __( 'Repeated background' ) }
-									checked={ isRepeated }
-									onChange={ toggleIsRepeated }
-								/>
+									isShownByDefault
+									hasValue={ () => !! isRepeated }
```
t-hamano:
```suggestion
									hasValue={ () => isRepeated }
```

`isRepeated` is a boolean so I don't think typecasting is necessary.

t-hamano (commented):
Sorry for the late review. I left some feedback but overall it works fine 👍

| Before | After |
|--------|--------|
| before| after | 

Also, I noticed while reviewing this PR that it is necessary to consider when multiple blocks are selected. At the very least, the alt text should be hidden when multiple blocks are selected. ([Example of the Image block](https://github.com/WordPress/gutenberg/blob/1664022139daa11353788a3ef59686fa7f58b262/packages/block-library/src/image/image.js#L742-L783)) This can be addressed in a follow-up.

t-hamano (commented):
Thanks for the update! Everything works as expected, but I think we need to address one last issue. Imagine the following scenario:

- In the trunk branch, insert a Cover block and apply a background image to it.
- Check out this PR
- Resolution option is set to "Thumbnail"

What we expect here is that the Resolution option should be "Full size", just like when we insert a new cover block and apply a background image to it in this PR.

Perhaps we need a fallback default value, like with the Media and Text block:

https://github.com/WordPress/gutenberg/blob/9ed79ba48d954395580617bcda9e3e09cf6c8752/packages/block-library/src/media-text/edit.js#L160

akasunil:
> Perhaps we need a fallback default value, like with the Media and Text block:

I was puzzled why this piece of code in the mediaText block was there; I couldn't find a use case for it. Now that I know, I'll do the same with cover block.

t-hamano:
@akasunil Thanks for the update!

I thought this PR was ready to ship, but after some more thought, I came to the conclusion that **a block deprecation** is needed.

Typically, when a block with an image is saved, a `size-{slug}` classname is included in the markup. This PR does the same, adding a class named `size-{slug}` to the `img` or `div` element in the Cover block.

This slug works as expected, but it doesn't take into account the Cover block already in the content. This means that we will need to either reselect full size from the Resolution option or reselect the image in order to update the markup to include the `size-full` class name.

To address this, I think we need to add a block deprecation and add `size-full` to the markup of existing Cover blocks.

Have you ever implemented a block deprecation?

Another question I have about this is whether or not we should include the `size-{slug}` class in the markup when the background image is not an `img` element, i.e. the `background-style` of a div element:

https://github.com/WordPress/gutenberg/blob/18c3e9396737c8e24e96ddbca543e30561dc1d96/packages/block-library/src/cover/save.js#L133-L147

I may not fully understand the block deprecation, so let me send a ping to @aaronrobertshaw and @andrewserong 🙇‍♂️

aaronrobertshaw (commented):
Thanks for the ping 👍 

I'm still trying to understand the full intent behind some of the changes here so let me know if I've misunderstood something.

**`size-{slug}` classes**

What's the reasoning behind adding these? 

- From what I can tell, these classes aren't used in relation to the display of an image of the requested size.
- The only references I could find were in old default themes (e.g. < TwentyFifteen)
- Could the sudden inclusion of these semi-generic classes break older themes?
- Are these classes only being added for "parity" with the Image block?
    - If so, I don't think that's achieved given the Image block applies the class to the `figure` wrapper and this PR applies it to the Cover block's inner `img`.
- Do we actually need the `size-{slug}` classes?
- These classes aren't added to the block in the editor. It should be consistent with the frontend


**Deprecations**

If the decision is that the `size-{slug}` classes are required in the markup, another call needs to be made on whether to enforce the "default" size class i.e. should blocks without a value in `sizeSlug` also get the `size-full` class.

If we enforce a class for the default image size, then yes a deprecation will be needed. If we don't enforce the default size class, then technically we might be able to avoid a deprecation as they are only required when an old set of attributes would generate invalid markup or when those old attributes need to be migrated at all.

Having said that, It is worth considering if avoiding a deprecation could create a "gotcha" down the line when someone else comes to update this complicated Cover block without knowing about the `sizeSlug` changes here. There is a performance cost adding deprecations so there are pros and cons.

**Managing PR's scope**

A final thought worth sharing is, perhaps we could split the refactor of the Cover controls to use `ToolsPanel` into a separate PR? 

It would help better document the history of the block, make debugging in future easier etc. This is probably even more important if the scope of the PR expands due to needing to define a deprecation.

What do you all think?

andrewserong:
Nice work here, it's testing well for me so far. Also a +1 for what Aaron mentioned, I'm curious about the need for the additional classnames. Sometimes when we have a need for additional classnames or deprecations I think it's worth taking a moment to see if there's another approach — the Cover block already has 13 deprecations that we need to maintain indefinitely so there's a bit of complexity here to wrangle. Also, with the HTML tag processor available, is there an opportunity to dynamically inject classnames when blocks are rendered, if that would avoid the need for a deprecation?

I also see the feature isn't available when using featured images, where we'd need a server-side approach, too (should we wish to implement that in the future).

> A final thought worth sharing is, perhaps we could split the refactor of the Cover controls to use ToolsPanel into a separate PR?

That sounds like a good idea to me, the more complex the Cover block gets, the harder I find it to review any changes to the block as it can be tricky to untangle if something isn't working correctly.

Another thought: this adds the feature as an ad hoc control as part of the Cover block only. Now that background image support is being used more widely across other blocks (Group, Post Content, Quote, etc), will it feel odd if there's control over the background image resolution in the Cover block but not those other blocks? This isn't an argument for not proceeding with this PR, but I mostly wanted to flag that if we go with an ad hoc control, we'll likely still need to rebuild it in a unified way in order to support other blocks, so just thought I'd mention it in case it affects how we're designing this feature.

akasunil:
> A final thought worth sharing is, perhaps we could split the refactor of the Cover controls to use `ToolsPanel` into a separate PR?

I agree with you, ill create separate PR for refactor of the Cover controls

akasunil:
> * Do we actually need the `size-{slug}` classes?

I'm not entirely sure about this, i was following https://github.com/WordPress/gutenberg/pull/24795 PR. Classes doesn't make any difference on front-end though.

Also, I have split the PR for refactor of setting pane in cover block.

t-hamano:
@aaronrobertshaw @andrewserong Thank you for the feedback!

> Do we actually need the "size-{slug}" classes?

In my experience, if any media was used in the block, the class names `wp-image-{id}`, `size-{slug}` were applied to the image element. I followed that convention, but this class may no longer be necessary 😅

[Here](https://github.com/WordPress/wordpress-develop/blob/bf09fe506620678fb82c3b872309edda0ed8ce61/src/wp-content/themes/twentyten/style.css#L800-L804) is one purpose of this class that I could find. It seems to be used to apply styles only to images uploaded via WordPress.

If this class is not necessary for the Cover block, we should not need to add the deprecation.

> we could split the refactor of the Cover controls to use "ToolsPanel" into a separate PR?

Yes, we did not expect the refactoring initially, but we needed to use `ToolsPanel` to achieve our purpose, but I should have proposed a split PR at an earlier stage.

> Now that background image support is being used more widely across other blocks (Group, Post Content, Quote, etc), will it feel odd if there's control over the background image resolution in the Cover block but not those other blocks?

I think this is a very good question. If I compare the background image support with the ad-hoc control of the Cover block, the only thing that background image support is missing is the resolution option, and the functionality is almost the same. And considering the performance of the website, background image support should also have a resolution option.

One concern is that the cover block supports the featured image, and if the featured image is used, background image support itself needs to be disabled. As far as I can tell, there is no way to dynamically control whether block support is enabled/disabled based on a specific block's attributes.

Ultimately, is the best solution the approach of refactoring using a `TooltPanel` and adding ad-hoc resolution control?

akasunil:
Could use your review on https://github.com/WordPress/gutenberg/pull/65432 
cc: @aaronrobertshaw @t-hamano

packages/block-library/src/cover/edit/index.js
```diff
@@ -199,6 +202,22 @@ function CoverEdit( {
 			averageBackgroundColor
 		);
 
+		if ( backgroundType === IMAGE_BACKGROUND_TYPE && mediaAttributes.id ) {
+			const { imageDefaultSize } = getSettings();
+
+			// Try to use the previous selected image size if its available
+			// otherwise try the default image size or fallback full size.
```
aaronrobertshaw:
```suggestion
			// Try to use the previous selected image size if it's available
			// otherwise try the default image size or fallback to full size.
```
Nit: Typos

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -219,6 +256,7 @@ export default function CoverInspectorControls( {
 								</ToolsPanelItem>
 							</>
 						) }
+
```
aaronrobertshaw:
```suggestion
```

Nit: Accidental whitespace change after rebase?

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -284,6 +322,13 @@ export default function CoverInspectorControls( {
 								/>
 							</ToolsPanelItem>
 						) }
+						{ ! useFeaturedImage && !! imageSizeOptions.length && (
+							<ResolutionTool
+								value={ sizeSlug }
+								onChange={ updateImage }
+								options={ imageSizeOptions }
+							/>
```
aaronrobertshaw: The reset behaviour for this control seems pretty odd. 

I think it might be missing the `defaultValue` prop otherwise it shows as having been modified before the user touches it. It's also used when resetting the control individually. 

If the default value prop isn't provided it will fall back to the first option in the available sizes which in my testing was thumbnail. That is different than how this panel's `resetAll` is resetting the `sizeSlug`.

aaronrobertshaw (commented):
Thanks for your patience here @akasunil 🙏 

> The current Settings panel does not use the ToolsPanel component, so the entire settings panel is refactored.

When you get a chance could you update the PR description to reflect the current state of the PR?

In the meantime, I've finally had a chance to give this another look after https://github.com/WordPress/gutenberg/pull/65432 has been merged and this PR rebased.

It mostly works well except for some issues around resetting the new resolution control. I've left an inline comment on a thread to follow to fix it. 


https://github.com/user-attachments/assets/e27b234a-814f-4ef2-b25d-9dbf828a7390

packages/block-library/src/cover/edit/inspector-controls.js
akasunil: It should be working fine now, i have set default value to "Full Size"

[Edit-Post-“Test”-‹-gutenberg-—-WordPress.webm](https://github.com/user-attachments/assets/00110e65-fd8c-46e8-b70f-f775c18a5183)

akasunil:
Thank you @aaronrobertshaw for quick review.

> When you get a chance could you update the PR description to reflect the current state of the PR?

I have updated the description.

packages/block-library/src/cover/edit/inspector-controls.js
```diff
@@ -112,7 +117,38 @@ export default function CoverInspectorControls( {
 		overlayColor,
 	} = currentSettings;
 
+	const sizeSlug = attributes.sizeSlug || DEFAULT_MEDIA_SIZE_SLUG;
+
 	const { gradientValue, setGradient } = __experimentalUseGradient();
+	const { getSettings } = useSelect( blockEditorStore );
+
+	const imageSizes = getSettings().imageSizes;
```
aaronrobertshaw: It appears as though there are a large number of places in the codebase where the presence of `imageSizes` in the settings is assumed. This could still be a touch more robust in the face of future changes if we fell back to an empty array if it isn't available. Up to you if you'd like to tweak it before landing this.

aaronrobertshaw (approved):
Thanks for sticking with this one @akasunil 💪 

The resetting behaviour now works for me after the latest updates. I think this is good to go 🚢 



https://github.com/user-attachments/assets/de3478fc-7281-4279-bbf7-4031c9a756ec

packages/block-library/src/cover/edit/inspector-controls.js
akasunil: If `imageSizes` is set to empty, `ResolutionTool` wont be visible. We could set fallback value for this, but if user have set imageSizes setting to empty array, should we fallback it to default value again?

Also, We could create hook to set default value if imageSizes setting is empty.

aaronrobertshaw: If it is null or undefined, the filter and map call below would throw an error.

akasunil: I see.

akasunil: I have added optional chaining operator to avoid errors. Thanks for the feedback.

packages/block-library/src/cover/edit/index.js
```diff
@@ -199,6 +202,22 @@ function CoverEdit( {
...
+			const { imageDefaultSize } = getSettings();
+
+			// Try to use the previous selected image size if it's available
+			// otherwise try the default image size or fallback to full size.
+			if ( sizeSlug && newMedia?.sizes?.[ sizeSlug ] ) {
+				mediaAttributes.sizeSlug = sizeSlug;
+				mediaAttributes.url = newMedia?.sizes?.[ sizeSlug ]?.url;
+			} else if ( newMedia?.sizes?.[ imageDefaultSize ] ) {
+				mediaAttributes.sizeSlug = imageDefaultSize;
+				mediaAttributes.url = newMedia?.sizes?.[ sizeSlug ]?.url;
```
Mamaduka, line 215: It looks like we missed a typo here, which caused #66765 regression. Here's the fix #66782.

I would appreciate the testing/reviews.

aaronrobertshaw: ✅ Done. 

Thanks for catching that @Mamaduka, looks like I'd gone blind for a bit there 🙈

huubl:
I'm curious why the resolution setting isn’t available when a feature image is used. Are there specific reasons for this limitation?

ajgagnon:
Yes, we really need this for featured images. It seems to have been missed and is using the full size.

aaronrobertshaw:
> Are there specific reasons for this limitation?

Featured images work differently, so they likely need specific handling. My understanding is that this was outside the scope of this PR. 

With the complexities encountered here, it makes sense to refine that and ensure the approach is solid before expanding the scope to featured images as well. 

> Yes, we really need this for featured images. It seems to have been missed and is using the full size.

If there isn't an issue already dedicated to this, it would be great if you could create one. Having a dedicated issue creates awareness and helps any contributor that is interested to jump in and propose fixes and enhancements.

ajgagnon:
It looks like this is the dedicated issue:

https://github.com/WordPress/gutenberg/issues/52507

huubl:
> > Are there specific reasons for this limitation?
> 
> Featured images work differently, so they likely need specific handling. My understanding is that this was outside the scope of this PR.
> 
> With the complexities encountered here, it makes sense to refine that and ensure the approach is solid before expanding the scope to featured images as well.
> 
> > Yes, we really need this for featured images. It seems to have been missed and is using the full size.
> 
> If there isn't an issue already dedicated to this, it would be great if you could create one. Having a dedicated issue creates awareness and helps any contributor that is interested to jump in and propose fixes and enhancements.

Hi @aaronrobertshaw,  

Maybe I'm missing something, but isn't it just a matter of passing the `sizeSlug` to the `get_the_post_thumbnail()` and `get_the_post_thumbnail_url()` functions in [the index.php file](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/cover/index.php)?

aaronrobertshaw:
Thanks for the question @huubl 👍 

> Maybe I'm missing something, but isn't it just a matter of passing the sizeSlug to the get_the_post_thumbnail() and get_the_post_thumbnail_url() functions in [the index.php file](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/cover/index.php)?

Unfortunately, there's more to it than that. Your suggestion only covers the rendering of the featured image for the block on the frontend. It misses updating the editor. 

The edit component for the block will need to be updated to make sure it is using the correct featured image media url for the selected image size. The featured image can also be updated outside of the block itself so there'd be a filter handling that url that will need tweaking further.

With the benefit of having landed the handling of image resolution options for normal images already, it should now be a smaller step to supporting featured images as well. In other words, the existing code for uploaded images can serve as a reference for anyone that would like to take a run at extending the image resolution support.

This enhancement isn't something I have the bandwidth for at the moment but I'd be very happy to help out with reviews if you'd like to put a PR together?

huubl:
Thanks for the detailed explanation, @aaronrobertshaw! That clears things up. I see @akasunil has made a [pull request](https://github.com/WordPress/gutenberg/pull/67273) which is already merged. That's awesome!

PR #62927, opened 2024-06-27
Patterns: Avoid mapping template parts objects to patterns

youknowriad:
Part of https://github.com/WordPress/gutenberg/issues/55083
Related https://github.com/WordPress/gutenberg/issues/59659
Unblocks #62913 

## What?

When working on a the dataviews extensibility actions. I noticed that some of the post type actions receive objects that are not actual post types, they are temporary objects that were created for the purpose of rendering template parts in the patterns page. We were mapping template parts to weird objects that resemble patterns.

For me this is not something we should be doing because:

 - We'll be creating temporary objects whose type is very unclear.
 - the actions are supposed to be generic and used both in dataviews and also in the editor. Which means that sometimes they get the normal templatePart object. I think there's a high chance that some of the actions are currently broken in the editor in trunk because of this.

## How?

This PR just removes that mapping and tries to update all the places where that mapping was necessary. The code is a bit hard to follow, so there's a chance that I missed some places. Some help testing everything is needed.

Unrelated: I was not able to test the template parts actions menu in the editor because I was not able to open template parts in the editor (I'll investigate this separately)

**Note** I noticed that there's also mapping happening between registered patterns (theme patterns) and saved patterns (user patterns). I think for these as well, we should be removing the mapping.

## Testing Instructions

1- Use the different filters, fields and actions in the template parts dataviews.

### DIFF
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -31,6 +31,7 @@ import { useEntityRecords } from '@wordpress/core-data';
 import { privateApis as editorPrivateApis } from '@wordpress/editor';
 import { privateApis as routerPrivateApis } from '@wordpress/router';
 import { parse } from '@wordpress/blocks';
+import { decodeEntities } from '@wordpress/html-entities';
 
 /**
  * Internal dependencies
@@ -54,6 +55,7 @@ import PatternsHeader from './header';
 import { useLink } from '../routes/link';
 import { useAddedBy } from '../page-templates/hooks';
 import { useEditPostAction } from '../dataviews-actions';
+import { defaultGetTitle } from './search-items';
 
 const { ExperimentalBlockEditorProvider, useGlobalStyle } = unlock(
 	blockEditorPrivateApis
@@ -201,8 +203,7 @@ function Title( { item } ) {
 		postId: isUserPattern || isTemplatePart ? item.id : item.name,
 		canvas: 'edit',
 	} );
-	const title =
-		typeof item.title === 'string' ? item.title : item.title.rendered;
+	const title = decodeEntities( defaultGetTitle( item ) );
 	return (
 		<HStack alignment="center" justify="flex-start" spacing={ 2 }>
 			<Flex
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -120,7 +120,7 @@ function Preview( { item, viewType } ) {
 	const [ backgroundColor ] = useGlobalStyle( 'color.background' );
 	const { onClick } = useLink( {
 		postType: item.type,
-		postId: isUserPattern ? item.id : item.name,
+		postId: isUserPattern || isTemplatePart ? item.id : item.name,
 		canvas: 'edit',
 	} );
 	const blocks = useMemo( () => {
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -30,6 +30,7 @@ import { usePrevious } from '@wordpress/compose';
 import { useEntityRecords } from '@wordpress/core-data';
 import { privateApis as editorPrivateApis } from '@wordpress/editor';
 import { privateApis as routerPrivateApis } from '@wordpress/router';
+import { parse } from '@wordpress/blocks';
 
 /**
  * Internal dependencies
@@ -116,14 +117,21 @@ function Preview( { item, viewType } ) {
 	const descriptionId = useId();
 	const isUserPattern = item.type === PATTERN_TYPES.user;
 	const isTemplatePart = item.type === TEMPLATE_PART_POST_TYPE;
-	const isEmpty = ! item.blocks?.length;
-
 	const [ backgroundColor ] = useGlobalStyle( 'color.background' );
 	const { onClick } = useLink( {
 		postType: item.type,
 		postId: isUserPattern ? item.id : item.name,
 		canvas: 'edit',
 	} );
+	const blocks = useMemo( () => {
+		return (
+			item.blocks ??
+			parse( item.content.raw, {
+				__unstableSkipMigrationLogs: true,
+			} )
+		);
+	}, [ item?.content?.raw, item.blocks ] );
+	const isEmpty = ! blocks?.length;
 
 	return (
 		<div
@@ -140,7 +148,7 @@ function Preview( { item, viewType } ) {
 				{ ! isEmpty && (
 					<Async>
 						<BlockPreview
-							blocks={ item.blocks }
+							blocks={ blocks }
 							viewportWidth={ item.viewportWidth }
 						/>
 					</Async>
@@ -187,11 +195,14 @@ function Author( { item, viewType } ) {
 
 function Title( { item } ) {
 	const isUserPattern = item.type === PATTERN_TYPES.user;
+	const isTemplatePart = item.type === TEMPLATE_PART_POST_TYPE;
 	const { onClick } = useLink( {
 		postType: item.type,
-		postId: isUserPattern ? item.id : item.name,
+		postId: isUserPattern || isTemplatePart ? item.id : item.name,
 		canvas: 'edit',
 	} );
+	const title =
+		typeof item.title === 'string' ? item.title : item.title.rendered;
 	return (
 		<HStack alignment="center" justify="flex-start" spacing={ 2 }>
 			<Flex
@@ -201,7 +212,7 @@ function Title( { item } ) {
 				className="edit-site-patterns__pattern-title"
 			>
 				{ item.type === PATTERN_TYPES.theme ? (
-					item.title
+					title
 				) : (
 					<Button
 						variant="link"
@@ -210,7 +221,7 @@ function Title( { item } ) {
 						// See https://github.com/WordPress/gutenberg/pull/51898#discussion_r1243399243.
 						tabIndex="-1"
 					>
-						{ item.title || item.name }
+						{ title || item.name }
 					</Button>
 				) }
 			</Flex>
@@ -321,7 +332,7 @@ export default function DataviewsPatterns() {
 			_fields.push( {
 				header: __( 'Author' ),
 				id: 'author',
-				getValue: ( { item } ) => item.templatePart.author_text,
+				getValue: ( { item } ) => item.author_text,
 				render: ( { item } ) => {
 					return <Author viewType={ view.type } item={ item } />;
 				},
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -418,7 +418,7 @@ export default function DataviewsPatterns() {
 					fields={ fields }
 					actions={ actions }
 					data={ data || EMPTY_ARRAY }
-					getItemId={ ( item ) => item.name }
+					getItemId={ ( item ) => item.name ?? item.id }
 					isLoading={ isResolving }
 					view={ view }
 					onChangeView={ onChangeView }
--- a/packages/edit-site/src/components/page-patterns/search-items.js
+++ b/packages/edit-site/src/components/page-patterns/search-items.js
@@ -20,11 +20,14 @@ import {
 	PATTERN_DEFAULT_CATEGORY,
 	PATTERN_USER_CATEGORY,
 	PATTERN_TYPES,
+	TEMPLATE_PART_POST_TYPE,
 } from '../../utils/constants';
 
 // Default search helpers.
-const defaultGetName = ( item ) => item.name || '';
-const defaultGetTitle = ( item ) => item.title;
+const defaultGetName = ( item ) =>
+	item.type !== TEMPLATE_PART_POST_TYPE ? item.name || '' : '';
+const defaultGetTitle = ( item ) =>
+	typeof item.title === 'string' ? item.title : item.title.rendered;
 const defaultGetDescription = ( item ) => item.description || '';
 const defaultGetKeywords = ( item ) => item.keywords || [];
 const defaultHasCategory = () => false;
--- a/packages/edit-site/src/components/page-patterns/search-items.js
+++ b/packages/edit-site/src/components/page-patterns/search-items.js
@@ -26,7 +26,7 @@ import {
 // Default search helpers.
 const defaultGetName = ( item ) =>
 	item.type !== TEMPLATE_PART_POST_TYPE ? item.name || '' : '';
-const defaultGetTitle = ( item ) =>
+export const defaultGetTitle = ( item ) =>
 	typeof item.title === 'string' ? item.title : item.title.rendered;
 const defaultGetDescription = ( item ) => item.description || '';
 const defaultGetKeywords = ( item ) => item.keywords || [];
--- a/packages/edit-site/src/components/page-patterns/use-patterns.js
+++ b/packages/edit-site/src/components/page-patterns/use-patterns.js
@@ -5,7 +5,6 @@ import { parse } from '@wordpress/blocks';
 import { useSelect, createSelector } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
 import { store as editorStore } from '@wordpress/editor';
-import { decodeEntities } from '@wordpress/html-entities';
 
 /**
  * Internal dependencies
@@ -16,7 +15,6 @@ import {
 	PATTERN_TYPES,
 	PATTERN_SYNC_TYPES,
 	TEMPLATE_PART_POST_TYPE,
-	TEMPLATE_ORIGINS,
 	TEMPLATE_PART_AREA_DEFAULT_CATEGORY,
 } from '../../utils/constants';
 import { unlock } from '../../lock-unlock';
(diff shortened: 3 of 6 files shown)


packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -187,11 +195,14 @@ function Author( { item, viewType } ) {
...
 	const isUserPattern = item.type === PATTERN_TYPES.user;
+	const isTemplatePart = item.type === TEMPLATE_PART_POST_TYPE;
 	const { onClick } = useLink( {
 		postType: item.type,
-		postId: isUserPattern ? item.id : item.name,
+		postId: isUserPattern || isTemplatePart ? item.id : item.name,
 		canvas: 'edit',
 	} );
+	const title =
+		typeof item.title === 'string' ? item.title : item.title.rendered;
```
ellatrix: Should this also use `defaultGetTitle`?

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -116,14 +117,21 @@ function Preview( { item, viewType } ) {
...
 		canvas: 'edit',
 	} );
+	const blocks = useMemo( () => {
+		return (
+			item.blocks ??
+			parse( item.content.raw, {
+				__unstableSkipMigrationLogs: true,
+			} )
+		);
+	}, [ item?.content?.raw, item.blocks ] );
```
ellatrix, line 135: Nit: in these cases I personally like to set a `const hasBlocks = !! item.blocks` to avoid triggering the callback. This could be a micro-optimisation though, not sure if this matters much in practice.

packages/edit-site/src/components/page-patterns/index.js
ellatrix: Also, do we still need `decodeEntities` here?

packages/edit-site/src/components/page-patterns/use-patterns.js
```diff
@@ -25,38 +23,16 @@ import { store as editSiteStore } from '../../store';
...
-	categories: [ templatePart.area ],
-	description: templatePart.description || '',
-	isCustom: templatePart.source === TEMPLATE_ORIGINS.custom,
-	keywords: templatePart.keywords || [],
-	id: createTemplatePartId( templatePart.theme, templatePart.slug ),
-	name: createTemplatePartId( templatePart.theme, templatePart.slug ),
-	title: decodeEntities( templatePart.title.rendered ),
-	type: templatePart.type,
-	_links: templatePart._links,
-	templatePart,
```
ellatrix, line 44: Huh 🤔

packages/editor/src/components/post-actions/actions.js
```diff
@@ -52,17 +53,18 @@ function isTemplateRemovable( template ) {
 	// than the one returned from the endpoint. This is why we need to check for
 	// two props whether is custom or has a theme file.
 	return (
-		[ template.source, template.templatePart?.source ].includes(
-			TEMPLATE_ORIGINS.custom
-		) &&
+		[ template?.source ].includes( TEMPLATE_ORIGINS.custom ) &&
```
ellatrix: We no longer need the `includes` here, right?

packages/editor/src/store/private-actions.js
```diff
@@ -371,10 +371,7 @@ export const removeTemplates =
 	( items ) =>
 	async ( { registry } ) => {
 		const isResetting = items.every(
-			( item ) =>
-				!! item &&
-				( item.has_theme_file ||
-					( item.templatePart && item.templatePart.has_theme_file ) )
+			( item ) => !! item && item.has_theme_file
```
ellatrix: Nit: can also just be `item?.has_theme_file`

ellatrix (commented):
I'm not sure what to test specifically. Do we have e2e test coverage for template parts/patterns?

I did get a warning `react-jsx-runtime.development.js:87 Warning: Each child in a list should have a unique "key" prop.` when loading the "all template parts" page.

packages/edit-site/src/components/page-patterns/index.js
youknowriad: defaultGetTitle is in another package, I don't want to export a small util like that.

youknowriad: actually, no it's in the same package but I'm still choosing to not using for now, cause it's "search" specific for now and this would break that.

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -210,7 +222,7 @@ function Title( { item } ) {
 						// See https://github.com/WordPress/gutenberg/pull/51898#discussion_r1243399243.
 						tabIndex="-1"
 					>
-						{ item.title || item.name }
+						{ title || item.name }
```
ntsekouras, line 225: Should we `decodeEntities` here to include `item.name`?

packages/editor/src/components/post-actions/actions.js
```diff
@@ -1031,22 +1026,32 @@ export const duplicateTemplatePartAction = {
...
+			);
+		}, [ item?.content?.raw, item.blocks ] );
 		const { createSuccessNotice } = useDispatch( noticesStore );
 		function onTemplatePartSuccess() {
 			createSuccessNotice(
 				sprintf(
 					// translators: %s: The new template part's title e.g. 'Call to action (copy)'.
 					__( '"%s" duplicated.' ),
-					item.title
+					typeof item.title === 'string'
```
ntsekouras: Nit: `getItemTitle` here..

ntsekouras (commented):
This LGTM and seems to work well! The test failures seem related though..

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -116,14 +119,21 @@ function Preview( { item, viewType } ) {
...
 	const [ backgroundColor ] = useGlobalStyle( 'color.background' );
 	const { onClick } = useLink( {
 		postType: item.type,
-		postId: isUserPattern ? item.id : item.name,
+		postId: isUserPattern || isTemplatePart ? item.id : item.name,
 		canvas: 'edit',
 	} );
+	const blocks = useMemo( () => {
+		return (
+			item.blocks ??
```
oandregal, line 130: Being more intentional about providing blocks to preview sounds good, but I'm not sure _when_ the `item.blocks` could be empty. Have you run into this or have a use case I test?

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -187,11 +197,13 @@ function Author( { item, viewType } ) {
 
 function Title( { item } ) {
 	const isUserPattern = item.type === PATTERN_TYPES.user;
+	const isTemplatePart = item.type === TEMPLATE_PART_POST_TYPE;
 	const { onClick } = useLink( {
 		postType: item.type,
-		postId: isUserPattern ? item.id : item.name,
+		postId: isUserPattern || isTemplatePart ? item.id : item.name,
```
oandregal, line 203: I've looked into this and this is what I see:

|                    | id                                     | name                                           | title                                             |
| ------------------ | -------------------------------------- | ---------------------------------------------- | ------------------------------------------------- |
| Template part      | `twentytwentyfour//footer`             | `twentytwentyfour//footer`                     | Footer                                            |
| User template part | `twentytwentyfour//user-template-part` | `twentytwentyfour//user-template-part`         | user template part                                |
| Pattern            | -                                      | `twentytwentyfour/text-title-left-image-right` | Title text and button on left with image on right |
| User pattern       | 1836                                   | `my-pattern`                                   | My pattern                                        |

So, it seems that:

- `id` or `name` can be used as the `postId` for template parts as they are the same (trying to understand if this is a bug in current `trunk`, but it isn't).
- This link is only provided for items that are not patterns by the theme: the `postType` could be simplified to always use `item.id`, as it's always present for the type of items we attach links to.
- The proposed change is also fine as it is.

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -187,11 +197,13 @@ function Author( { item, viewType } ) {
...
 		canvas: 'edit',
 	} );
+	const title = decodeEntities( defaultGetTitle( item ) );
```
oandregal, line 206: Nice catch 👍 

Before:

https://github.com/WordPress/gutenberg/assets/583546/3bf73a06-2202-46c0-ad6f-6a465020db07

After:

https://github.com/WordPress/gutenberg/assets/583546/960d32d1-e7da-4384-a64c-0837f52b48b4

packages/edit-site/src/components/page-patterns/index.js
oandregal: As far as I've tested, the `name` gets those entities removed, so we don't need to do that.

For example, creating a pattern with title `andr&eacute;` creates the following:

- id: 1838
- name: `andr`
- title: `andr&eacute;`

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -406,7 +418,7 @@ export default function DataviewsPatterns() {
 					fields={ fields }
 					actions={ actions }
 					data={ data || EMPTY_ARRAY }
-					getItemId={ ( item ) => item.name }
+					getItemId={ ( item ) => item.name ?? item.id }
```
oandregal, line 421: Based on https://github.com/WordPress/gutenberg/pull/62927/files#r1658553606 there's always a `item.name`.

packages/edit-site/src/components/page-patterns/search-items.js
```diff
@@ -20,11 +20,14 @@ import {
...
 	PATTERN_USER_CATEGORY,
 	PATTERN_TYPES,
+	TEMPLATE_PART_POST_TYPE,
 } from '../../utils/constants';
 
 // Default search helpers.
-const defaultGetName = ( item ) => item.name || '';
-const defaultGetTitle = ( item ) => item.title;
+const defaultGetName = ( item ) =>
+	item.type !== TEMPLATE_PART_POST_TYPE ? item.name || '' : '';
```
oandregal, line 28: [Reference about names](https://github.com/WordPress/gutenberg/pull/62927/files#r1658553606).

It seems the `name` is used to search items by rank here. This change avoids targeting the `name` of a template part while still using it for searching patterns.

Given the name is the slug, it sounds like it shouldn't be used to search items. Though, I reckon I'm not familiar with how we do this operation in other parts of the app and can be convinced one way or the other. What I find odd is the inconsistency of using it for searching patterns but not parts.

packages/edit-site/src/components/page-patterns/use-patterns.js
```diff
@@ -298,11 +274,7 @@ export const usePatterns = (
 	return useSelect(
 		( select ) => {
 			if ( postType === TEMPLATE_PART_POST_TYPE ) {
-				return selectTemplatePartsAsPatterns(
-					select,
-					categoryId,
-					search
-				);
+				return selectTemplateParts( select, categoryId, search );
```
oandregal, line 277: I've only found two uses of the `usePatterns` hook:

- One is in the page-patterns screen, also adapted in this PR.
- The other is in the [sidebar](https://github.com/WordPress/gutenberg/blob/c66631de63dc4cc9ce279ec5fd8e873004a1ed0e/packages/edit-site/src/components/sidebar-navigation-screen-patterns/use-pattern-categories.js#L28), that doesn't need any adaption as it's only used to retrieve data from patterns but not from template parts post types (what is being modified in this PR).

packages/edit-site/src/components/page-patterns/index.js
youknowriad: That's not entirely correct. Template parts don't have "name". You can try this locally to check `wp.data.select('core').getEntityRecords( 'postType', 'wp_template_part' );` 

The lack of fallback to id was also what was triggering the React key warning raised by Ella.

packages/edit-site/src/components/page-patterns/search-items.js
youknowriad: The name doesn't exist for template parts, in trunk it uses the id which doesn't make sense. For patterns names are more meaningful in a sense.

packages/edit-site/src/components/page-patterns/index.js
youknowriad: item.blocks doesn't exist in template parts by default, it only exists if you get the edited template part objects after it was modified in the editor (and not saved)

packages/edit-site/src/components/page-patterns/use-patterns.js
oandregal: I presume that, back in the days, the `usePatterns` hook was created as a way to avoid if conditions in the page (rules of hooks) and also to provide a stable structure to consumers (sidebar, page) independently from where the data came from. It seems to me the consumers still do a lot of data wrangling depending on the type, so not sure how effective this has been for the second goal.

It'd be good to rename it to `usePatternsOrParts` of something similar for clarity at some point: not a requirement for this PR.

packages/edit-site/src/components/page-patterns/index.js
oandregal: I was checking that now. This was something that we did before on `templatePartToPattern`. 👍

packages/edit-site/src/components/page-patterns/index.js
oandregal: Not sure what other processing we do and how the data differs in other parts of the codebase (still reviewing parts of the PR and catching up), but the `item` processed by DataViews does have that data (compiled the table above using a console.log within the `Title` component to).

I've also tested that links from the title worked fine in `trunk` for template parts (open the editor) — if they didn't have the `name`, it would have failed and would be something to cherry-pick for 6.6 I presume.

oandregal: Oh, I see what happened: I extracted the data in `trunk`, where we do the `templatePartToPattern` thing that is removed in this PR. 👍

packages/edit-site/src/components/page-patterns/search-items.js
oandregal: Yeah, I saw the other thread, the data I extracted was based on what this PR actually removes (`templateToPatterns`), so this is good 👍

packages/edit-site/src/components/page-patterns/index.js
oandregal: There's not. See thread for reference.

packages/edit-site/src/components/page-patterns/index.js
youknowriad: Yes, that's actually the purpose of this PR. We should be using the default objects and not create local formats.

packages/edit-site/src/components/page-patterns/use-patterns.js
```diff
@@ -25,38 +23,16 @@ import { store as editSiteStore } from '../../store';
 
 const EMPTY_PATTERN_LIST = [];
 
-const createTemplatePartId = ( theme, slug ) =>
-	theme && slug ? theme + '//' + slug : null;
-
-const templatePartToPattern = ( templatePart ) => ( {
-	blocks: parse( templatePart.content.raw, {
```
oandregal, line 32: Reviewed how this (or whether) is being used in the consumer code that uses parts (page-patterns). Everything has been migrated:


- blocks: adapted consumer code
- categories: unused
- description: no need to do anything
- isCustom: unused
- keywords: unused
- id: parts already have an id
- name: removed & adapted the consumer code
- title: removed & adapted the consumer code
- type: as it is
- `_links`: as it is

packages/edit-site/src/components/page-patterns/index.js
oandregal: The only thing I still need to review is `packages/editor/src/components/post-actions/actions.js`. I need to be somewhere else now, will review that one later. Changes are looking good so far.

packages/editor/src/components/post-actions/actions.js
oandregal: What Nik suggested would do the job. Came here to share that in `templatePartToPattern` the `title` passed through `decodedEntities` https://github.com/WordPress/gutenberg/pull/62927/files#diff-0f49c90dc39786ecdf45f374247e433f763d295d80bdaecda3bab11357e4d7fdL41

packages/edit-site/src/components/page-patterns/use-patterns.js
oandregal: I see actions were also a consumer of this (through dataviews's item). I've only found one case that requires updating (`title` is not using `decodedEntities` https://github.com/WordPress/gutenberg/pull/62927/files#r1658485082 ) but the rest were already ported (`isCustom`, etc.).

oandregal (commented):
Code-wise, this is a good move. The `usePatterns` hook and the idea of providing a stable structure for the page-patterns across patterns & parts wasn't working: it leaked to actions & the page-patterns still had to do a lot of data wrangling depending on the post type.


I've done some testing: both the Patterns dataviews page & actions within the editor.

Issues:

- Parts: duplicate action doesn't work properly (it works in `trunk`).
  - in dataviews: shows weird title (object, not the title string)
  - in editor: it triggers an error (probably related)

Captura de ecrã 2024-06-28, às 16 45 21

- User patterns: the title for the duplicate action needs passing through `decodeEntities` (creating a pattern whose name is `andr&eacute;` should help testing this: compare duplicate action vs rename action to see the difference)

Edit:
Captura de ecrã 2024-06-28, às 16 47 08


Duplicate (dataviews):

Captura de ecrã 2024-06-28, às 16 47 13

Duplicate (editor):

Captura de ecrã 2024-06-28, às 16 50 36

packages/edit-site/src/components/page-patterns/index.js
oandregal: done the review

youknowriad:
Thanks for the thorough testing @oandregal, we definitely need more e2e test. Probably one per action. I'm going to try to add some as I refactor and type the actions.

youknowriad:
We should also look at removing `convertPatternPostToItem` mapping if someone is available to work on it, please go ahead :)

ntsekouras (approved):
Thank you!

PR #62934, opened 2024-06-27
ApiFetch: Remove i18n package dependency

sirreal:
## What?

Remove `api-fetch` package's dependency on `i18n`.

## Why?

The primary reason to remove this is to unblock work on an api-fetch module: https://core.trac.wordpress.org/ticket/60647

This package is for making api calls and there's not a strong motivation to localize most strings. None of the localized texts are end-user-facing. This dependency was only used to localize error messages targeted at developers. Other packages, for example hooks, do not localize these types of messages:

https://github.com/WordPress/gutenberg/blob/fe67010143b20d23349051e80110aa1424e79999/packages/hooks/src/validateHookName.js#L19

Every dependency may increase the amount of JavaScript required on the client, so removing a dependency like this may improve the performance of sites using `apiFetch`.

## How?

Remove the i18n dependency and use static strings instead of localized ones.

## Testing Instructions
CI passes.

### DIFF
--- a/packages/api-fetch/CHANGELOG.md
+++ b/packages/api-fetch/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
+
+-   Remove dependency on `@wordpress/i18n` ([#tbd](tbd)).
+
 ## 7.2.0 (2024-06-26)
 
 ## 7.1.0 (2024-06-15)
--- a/packages/api-fetch/CHANGELOG.md
+++ b/packages/api-fetch/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Enhancements
 
--   Remove dependency on `@wordpress/i18n` ([#tbd](tbd)).
+-   Remove dependency on `@wordpress/i18n` ([#62934](https://github.com/WordPress/gutenberg/pull/62934)).
 
 ## 7.2.0 (2024-06-26)
 
--- a/packages/api-fetch/package.json
+++ b/packages/api-fetch/package.json
@@ -29,7 +29,6 @@
 	"types": "build-types",
 	"dependencies": {
 		"@babel/runtime": "^7.16.0",
-		"@wordpress/i18n": "file:../i18n",
 		"@wordpress/url": "file:../url"
 	},
 	"publishConfig": {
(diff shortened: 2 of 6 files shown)


ellatrix:
> This package is for making api calls and there's not a strong motivation to localize most strings. None of the localized texts are end-user-facing. This dependency was only used to localize error messages targeted at developers. Other packages, for example hooks, do not localize these types of messages:

Why not only return a code then, if they are not user facing? These codes could be documented, and "translated" into readable text by the implementor of api-fetch.

packages/api-fetch/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
```
sirreal, line 5: This is probably a breaking change.

```suggestion
### Breaking Changes
```

sirreal:
> None of the localized texts are end-user-facing.

This was a bad assumption on my part. These do look like strings that are intended to be printed to the end users:

- You are probably offline.
- Media upload failed. If this is a photo or a large image, please scale it down and try again.
- An unknown error occurred.

It seems very likely these are printed in toast messages somewhere.

ellatrix:
@sirreal Yes, but my point remains: if you really want to remove translations from this package, you _could_ leave the code only, and then handle the code in the consuming package?

ellatrix:
(And maybe just leave an untranslated string for back compat, not sure)

sirreal:
Moving from a translated string to an untranslated string is a regression. There's a tradeoff that has some benefits and I believe the tradeoff is justified. 

Moving from a translated string to a code is a bigger regression with little benefit. The error messages are mostly guesses as to what happened, but removing them entirely doesn't seem to add additional benefit over providing an untranslated string.

You have given me an idea that may allow us to avoid the regression entirely while unblocking work on an api-fetch module. We can move the error messages to their own file and only include them in the apiFetch script form.

---

We can still consider this PR and removing the dependency, but may not block the modules work.

gziolo:
The messages are quite simple so you could move them to the config file and expose the translated form from the server using the new way to share data with the client. It will have the expected impact on the frontend regarding the bundle size. In the editor it won’t have any negative impact other than a bit indirect way to handle these translations.

The description of PR needs to get updated to reflect the fact that these messages are user facing in the editor.

sirreal:
I like that idea @gziolo. I'll close this PR because we have better alternatives.

PR #62936, opened 2024-06-27
CustomSelectControlV2: expose legacy wrapper through private APIs

ciampo:
## What?

Part of #55023

Make the `CustomSelectControl` V2 legacy wrapper available via the component package's private APIs

## Why?

So that we can start experimenting with using the new V2 legacy wrapper instead of V1 in the editor, as we prepare for swapping V1 implementation with the V2 legacy wrapper implementation

## How?

By exporting the component via `lock`ed APIs

## Testing Instructions

Make sure that it is possible to import the component via the `unlock` function from other packages.

### DIFF
--- a/packages/components/src/private-apis.ts
+++ b/packages/components/src/private-apis.ts
@@ -14,6 +14,7 @@ import {
 	useCompositeStore as useCompositeStoreV2,
 } from './composite/v2';
 import { default as CustomSelectControl } from './custom-select-control';
+import { default as CustomSelectControlV2LegacyWrapper } from './custom-select-control-v2/legacy-component';
 import { positionToPlacement as __experimentalPopoverLegacyPositionToPlacement } from './popover/utils';
 import { createPrivateSlotFill } from './slot-fill';
 import {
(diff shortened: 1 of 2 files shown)


packages/components/src/private-apis.ts
```diff
@@ -39,6 +40,7 @@ lock( privateApis, {
 	CompositeItemV2,
 	CompositeRowV2,
 	useCompositeStoreV2,
+	CustomSelectControlV2LegacyWrapper,
```
tyxla: Naming nit: is this really a wrapper though? Since the legacy component isn't really a wrapper as part of a composable component group, but rather a monolithic component, I'd prefer naming this `CustomSelectControlV2Legacy` or `CustomSelectControlV2LegacyAdapter` since `Wrapper` implies a wrapper component in the same way the `Tabs` parent component wraps the rest of the subcomponents. 

No strong feelings about it, but I feel like others might potentially be confused by that naming.

tyxla (approved):
LGTM 👍 

Just a minor thought on the naming choice.

packages/components/src/private-apis.ts
ciampo: Good point, pushed an update

PR #62940, opened 2024-06-27
Post editor: Fix triple scrollbars in device previews

### LINKED ISSUE #62911: Extra scrollbar appears while editing patterns/template parts whenever there are notices
### Description

When editing patterns any notices result in a second scrollbar appearing.

The bug happens because the notices push the editor canvas area off-screen, which triggers a scrollbar.

I'm personally unsure of the right fix. For the normal post editing experience, the editor canvas shrinks when there's a notice, so perhaps the focus mode should do the same?

1. Edit a pattern (post or site editor)
2. Add enough content so that a scrollbar appears
3. Run this JavaScript snippet in your browser console:
```js
wp.data.dispatch( 'core/notices' ).createInfoNotice( 'Notice' );
```

Observe two scrollbars

Screenshot 2024-06-27 at 4 21 31 pm

### Environment info

_No response_

Yes

Yes

stokesman:
## What?
Fixes #62911

A fix for a regression in the Post editor where device previews create three scrolling areas.

## Why?
In WP 6.5 device previews in the Post editor don’t have three scrolling areas.

## How?
Revises a selector to more selectively apply some styles that cause the issue. The styles appear to only be helpful when metaboxes are present.

## Testing Instructions
1. Open the Post editor
2. Select either "Tablet" or "Mobile" preview
3. If the Device preview is shorter than your browser’s viewport, shorten the browser window until it does not contain the full height of the device preview.
4. Observe there are only two scrollbars

### Before
image

### After
image

## More about the issue
As best I can tell this came about from introducing the "Zoom out" experiment. While that is still an experiment it did introduce markup and style changes that are present even when not opting into the experiment. Testing [WP nightly in playground](https://playground.wordpress.net/?php=8.0&wp=nightly&storage=none) I can reproduce the issue:
image

### DIFF
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -1,8 +1,3 @@
-.editor-visual-editor {
-	flex: 1 0 auto;
-	height: auto;
-}
-
 .edit-post-layout__metaboxes {
 	flex-shrink: 0;
 	clear: both;
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -1,3 +1,8 @@
+.edit-post-layout.has-metaboxes .edit-post-visual-editor {
+	flex: 1 0 auto;
+	height: auto;
+}
+
 .edit-post-layout__metaboxes {
 	flex-shrink: 0;
 	clear: both;


talldan (approved):
Thanks for the fix. In my testing it fixes both #62911 and #62940, and it doesn't seem to cause any other issues that I could see.

talldan:
I've also made another small PR that only shows the second device preview scrollbar when really needed - https://github.com/WordPress/gutenberg/pull/62952

t-hamano:
I came across this PR while investigating the issue reported in #63110.

This PR may have caused two issues for the post editor that is not iframed and does not have meta boxes.

- There is not enough space at the bottom of the content
- The background color disappears halfway through

"The post editor that is not iframed and does not have meta boxes" refers to when a block version 3 or below is registered without custom fields enabled.

In this case, the `.has-metaboxes` class is not applied to the editor. As a result, the selector updated by this PR is no longer applied, so the height of `.edit-post-visual-editor` becomes `100%` instead of `auto`.

For more information on this issue, see #63110.

PR #62942, opened 2024-06-27
Tabs: split animation logic into multiple separate composable utilities.

DaniGuardiola:
## What?

I organized and documented these.

## Why?

In a follow-up, I'm gonna move them to a shared file where they'll be used in other components (first one will be `ToggleGroupControl`.

## How?

Make each utility do a single thing really well, document everything, ensure nice minimal APIs and types.

## Testing Instructions

Tabs should still work properly, can check the storybook stories or places with tabs in Gutenberg.

n/a

## Screenshots or screencast 

n/a

### DIFF
--- a/packages/components/src/tabs/tablist.tsx
+++ b/packages/components/src/tabs/tablist.tsx
@@ -38,11 +38,15 @@ type AnyFunction = ( ...args: any ) => any;
  * Creates a stable callback function that has access to the latest state and
  * can be used within event handlers and effect callbacks. Throws when used in
  * the render phase.
+ *
  * @example
+ *
+ * ```tsx
  * function Component(props) {
  *   const onClick = useEvent(props.onClick);
  *   React.useEffect(() => {}, [onClick]);
  * }
+ * ```
  */
 function useEvent< T extends AnyFunction >( callback?: T ) {
 	const ref = useRef< AnyFunction | undefined >( () => {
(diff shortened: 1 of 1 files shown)


packages/components/src/tabs/tablist.tsx
```diff
@@ -25,57 +26,121 @@ import { TabListWrapper } from './styles';
...
+ *   const onClick = useEvent(props.onClick);
+ *   React.useEffect(() => {}, [onClick]);
+ * }
+ * ```
+ */
+function useEvent< T extends AnyFunction >( callback?: T ) {
+	const ref = useRef< AnyFunction | undefined >( () => {
+		throw new Error( 'Cannot call an event handler while rendering.' );
 	} );
+	useInsertionEffect( () => {
```
ciampo, line 55: I guess we're using `useInsertionEffect` because it runs event before `useLayoutEffect` — probably good to document it?

ciampo (approved):
Good work! Let's just leave a CHANGELOG entry (under the `Internal` section) before merging

DaniGuardiola:
@ciampo thx for the review, I'll add a changelog entry for the whole thing in the last PR, since this is really step 1 of 3 of the same task (refactor ToggleGroupControl to use the same animation logic as Tabs) 👍

packages/components/src/tabs/tablist.tsx
DaniGuardiola: (discussed in our sync call)

PR #62945, opened 2024-06-27
Fix typo to be preposition, not verb, in some package comments and documentations

roygbyte:
## What?
Fix typos

## Why?
For excellence!

## How?
Using ed(1), the standard text editor. 

## Testing Instructions
Look at the words and see if they make sense.


t-hamano (approved):
Nice catch 👍

PR #41001, opened 2022-05-11
Support a new API for useResizeObserver

youknowriad:
I'll detail more as I make progress.

PR #39458, opened 2022-03-15
Eslint: Exempt TS files from `jsdoc/require-param`

mirka:
## What?

Exempts TypeScript files from the [`jsdoc/require-param`](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/.README/rules/require-param.md) rule.

## Why?

TypeScript files generally have their type information in code rather than in JSDoc comments. The `jsdoc/require-param` rule requires all params to be documented in the JSDoc, which would be redundant in these cases.

For example, `jsdoc/require-param` would consider this an error, because `props` is not documented in the comment:

```ts
/**
 * This is Foo
 */
export function Foo( props: Props ) { ... }
```

## Testing Instructions

- Simple doc comments like the above should not trigger eslint errors in TS files.
- The behavior remains the same for `.js` files.

PR #62946, opened 2024-06-27
Tabs: move animation-related utilities into separate utils file.

DaniGuardiola:
Note: #62942 needs to be merged first

## What?

Move animation-related utilities from Tabs into separate utils file.

## Why?

For future use in other components. Related: #62942


.eslintrc.js
```diff
@@ -379,7 +379,7 @@ module.exports = {
 		{
 			files: [
 				'**/@(storybook|stories)/*',
-				'packages/components/src/**/*.tsx',
```
ciampo: Nitty, but it would be great to move this to a separate PR to keep the scope of this one tight

packages/components/src/utils/react.ts
ciampo, line 1: Looking at the current structure of the `utils` folder, I'm not sure about moving these 3 utils in the same file, and calling it `react` — maybe to adhere more to the current conventions, we could split each utility function to its separate file, and move them under `packages/components/src/utils/hooks/*` ?

packages/components/src/utils/react.ts
```diff
@@ -0,0 +1,244 @@
...
+ * @example
+ *
+ * ```tsx
+ * function Component(props) {
+ *   const onClick = useEvent(props.onClick);
+ *   React.useEffect(() => {}, [onClick]);
+ * }
+ * ```
+ */
+export function useEvent< T extends AnyFunction >( callback?: T ) {
```
ciampo: cc @youknowriad for now we're keeping this hook internal to components package. But if you think it would be useful, we could also move it to `@wordpress/compose`

ciampo (approved):
Just left a couple of comments, but overall looking good (let's also add a CHANGELOG entry, could be using the same entry as https://github.com/WordPress/gutenberg/pull/62942)

packages/components/src/utils/react.ts
youknowriad: Definitely something that looks useful. I thought at some point that React was working on providing it no?

packages/components/src/utils/react.ts
```diff
@@ -0,0 +1,244 @@
...
+ * const [ targetElement, setTargetElement ] = useState< HTMLElement | null >();
+ *
+ * useResizeObserver( targetElement, ( element ) => {
+ *   console.log( 'Element resized:', element );
+ * } );
+ *
+ * <div ref={ setTargetElement } />;
+ * ```
+ */
+export function useResizeObserver(
```
youknowriad: Why do we have this hook here and useResizeObserver in compose as well?

packages/components/src/utils/react.ts
ciampo: They were, but from what I understand the release of the hook has been put on hold. We could add it anyway, and if it ever gets officially released in react, we can move to the official one

packages/components/src/utils/react.ts
ciampo: I'll let @DaniGuardiola reply with more details if necessary, but from earlier conversations he took a look at the `useResizeObserver` hook in `@wordpress/compose` and found it overly complex and difficult to compose with other smaller utilities (like in this PR).

Now that resize observers are well supported by browsers, we could look at simplifying the `@wordpress/compose` version with something similar to the above implementation

youknowriad: I understand that it's easier to duplicate stuff and add new internal utils but It would be good to avoid recreating the same things over and over. A small effort can bring benefits to the whole project. In fact I had started improving it in the past but lost interest on it. https://github.com/WordPress/gutenberg/pull/41001

.eslintrc.js
mirka: It definitely should be separate, since the omission of `ts` files was deliberate! We got some strong pushback on this point (#39458).

DaniGuardiola: I've only skimmed that issue, but is the main point about docgen? Because if that's the case I'd be happy to fix whatever needs to be fixed. This rule and its autofix almost caused my utilities to be wrongly named and documented, as it adds blank `@param` declarations, so the result would have been arguably worse. In any case, I'm just gonna disable it for this file since these utils are internal, and leave that matter for another day.

Thanks for the context!

packages/components/src/utils/react.ts
DaniGuardiola: With that approach, we'll end up with about 8 new files, spread between hooks and other places.

I personally dislike that very much because the split is very arbitrary. The utility being a hook is not something we should care about. If I have multiple utilities that are related to each other (e.g. think sizing and positioning utilities), what good does it do to have them spread around the directory? Organizing by whether something is a hook seems comparable to organizing based on whether a utility returns a number or a string. And I also think splitting each little function into a separate file makes everything hard to find, navigate and parse.

This structure that I chose is largely inspired in Ariakit's organization, which has saved me a lot of time by having a flat directory of simple categories that contain collections of related utilities. Look how neat and easy to work with this is: https://github.com/ariakit/ariakit/tree/main/packages/ariakit-core/src/utils

I think we should strive for something like this to reduce maintenance overhead and potential duplication. I definitely don't feel compelled to look for existing utilities looking at the current folder, where I have to go subdirectory by subdirectory, opening one file for each utility, to see if something does what I need.

Anyway, if you think this is enough of a change to grant separate discussion and consensus, I'm happy to re-organize this as requested and leave it for another day. Let me know and I'll do it, but know I'm gonna be pushing for a simplification of the utilities directory in a follow-up discussion :P

packages/components/src/utils/react.ts
DaniGuardiola: The hook called `useResizeObserver` has a misleading name. It does use a `ResizeObserver` under the hood, but has a hyper-specific API focused on a narrow use-case (getting the viewport-relative size of a parent element). Neither the API nor the sizing computation worked in my case. The existing hook should probably be called `useParentElementBoundingClientRect`, and even then it would not describe all of the nuance fully because it's a chunky hook full of undocumented logic.

My hook here is completely different and does a single thing: takes an element and a callback, and executes the callback on resize by using a ResizeObserver. Nothing more. All of the nuance is also clearly documented. As you can see further down this file, it is nicely composed with other utilities (like one to obtain the sizing and position in the appropriate context here) to achieve what was needed for this task.

So to be clear: there isn't any prior code that does what this set of utilities do, so there's no duplication. In order to effectively avoid duplication in a future-proof manner, utilities must be carefully crafted to do a single thing with a minimal API, so that they can be composed with other utilities. Unfortunately, that's not the case of the pre-existing `useResizeObserver` utility.

DaniGuardiola: I wrote more specifics here: https://github.com/WordPress/gutenberg/pull/60560#discussion_r1610163431

DaniGuardiola: If you're interested, I'd be happy to pair on bringing atomic-sized and composable utilities for resize observer related stuff to the shared utils package. My hook here for instance is a good primitive to build things upon, since it doesn't get simpler than that. We can easily build more specific variants from it, like the one present in this file that tracks the overflow-parent-relative rect. It should also be quite easy to add more ergonomic APIs to work with refs more directly. Just ping me if you're interested :)

youknowriad: It is less about how the hook works and more about having one way to do something in the whole repo and be consistent and avoid silos (components doing things one way, the rest doing something else), I think it's an important principle that we should stick to. 

Now on the hook, both hooks have the same purpose, observe the changes of size of an element. If you check the PR that I linked, it actually offers a very similar API to the hook in components, so it's possible to migrate (offer both while allowing migration).

Also, on my opinion, some changes are not worth it. I'm not saying it's the case here, I don't know but sometimes it's ok to stick for something that is a bit older and maybe not ideal of the shiniest approach requires big refactoring without important gains (either performance or else).

DaniGuardiola: I appreciate the points you're making.

That said, I suggest you look at the links I dropped. As I explained, the existing hook does a very different thing to what I do, in a very different way. It wouldn't have worked for my use case. Also, I would argue that the fact that a utility abstracts a web primitive shouldn't block the use of said primitive in other places where it'd be useful but the abstraction just doesn't work, like here.

I understand that there's a history and a context to all of this, don't think I'm ignoring this fact - I still think it's productive to recognize the limitations of it and understand why it's not ideal, design-wise. Independently of what the pragmatic approach is here, I'm arguing that writing utilities in the style I've described (and used in this PR) is a more effective way to reduce duplication and future-proof them. For example, you could very easily construct the older hook using mine, but it'd be outright impossible the other way around. I think composition is king when it comes to these things.

I think there's a lot to learn from the ["do one thing and do it well" principle of "Unix philosophy"](https://en.wikipedia.org/wiki/Unix_philosophy#Do_One_Thing_and_Do_It_Well).

DaniGuardiola: Upon closer inspection, I do believe my hook could probably be renamed to something like `useOnElementSizeUpdate`, since the point of it is to just be notified of size updates and the resize observer is merely an implementation detail. In fact, the [ResizeObserverEntry](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry) objects are ignored by it (they are useless for this use-case), and there's a special behavior by default that fires the handler initially with the purpose of keeping things up to date always.

Maybe this clarifies why the other hook is not a good fit here better?

I also took a closer look at your PR and I think it's much closer to what I would call a minimal and composable API, love it! Definitely much better than the absolutely positioned child approach, IMO.

DaniGuardiola: Alright so, since I've been getting a bit of pushback on this, I've stopped to do a bit of a reset and look at this from the outside, and I've realized a few things. I think I have been explaining myself poorly, and grown a bit frustrated which has led me to missing some important things. Let me elaborate:

- I think my main mistake has been that I haven't explained the initial purpose of this logic. Initially, it didn't even have a `ResizeObserver`. In most cases, the observer won't even fire, ever! I assume this'll be surprising because I have failed to explain myself properly. Let me recap:
  - I needed to track which tab is active, and then obtain its position and size. That was easy enough to achieve by wiring a bit of Ariakit state with some logic to compute the size (relative to the offset parent, long story, yadda yadda).
  - However, in some rare cases, a tab would change its size in a re-render (e.g. in the block editor, left sidebar, the first tab initially has a label of "Document", which then gets instantly replaced by "Template", or "Post", or "Page".
  - Since the size computation was already done before that re-render, the indicator would remain with a size that was computed with "Document", which does not match the new tab dimensions with the new label.
  - To solve for this, I needed a way to be notified when this happens to re-compute size and position. For that, a `ResizeObserver` (at least the part where it fires a callback when there's a resize) is the perfect tool.
  - Since I was using it, someone suggested that I used the pre-existing hook. However, for the reasons stated (and a few more), it just doesn't fit the use-case, and it wouldn't have any benefit over just using the primitive directly, as I do.
- Another mistake that I've realized now is that my hook here is not named appropriately. It should probably be called something like `useTrackElementRectUpdates` instead. The main use-case, as explained, is to be notified when the target element is swapped (e.g. from tab 1 to tab 2) to obtain an up-to-date size and position, **and** also get those updates when the same target element alters those properties (e.g. "Document" to "Post" label). It leaves the implementation of actually computing any values to the user, which allows for the neat composition I achieve below in this file. Again, `ResizeObserver` is purely an implementation detail, and it could be very easily swapped by something else in the future if a better browser API were to come out.
- I also think I overlooked what the old hook is doing. This was also a big mistake, as I didn't realize that it was much more pegged to the `ResizeObserver` API than I thought. I have now looked at it and, after learning more about [ResizeObserverEntry](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry) objects (which I have never had a use-case for), I realize that I've been obtusely arguing that the hook has too much custom logic, which is false. I apologize for this. I still have some opinions on the API (as you do, for example the change in your PR is much more aligned with what I would describe as a better API for this).

All in all, I've had to do some self-reflection here and I apologize for being obtuse, and for overlooking important details. I let some mild frustration cloud my judgment. My bad!

DaniGuardiola: TL;DR: sorry for being a dummy dum dum and thank you for engaging with me :)

With this new perspective, I'd be happy to re-open your PR and help get it to land @youknowriad

youknowriad: Thanks for the discussion here. Just want to say that this discussion doesn't have to be a blocker for the current PR. I'm just trying to share my mental model and some guiding principles.

ciampo: Thank you Dani for sharing in detail your train of thoughts, it helps a lot! I personally empathise with Riad's sentiment — consistency is important in a project like Gutenberg. And if we could improve the `useResizeObserver` hook from `@wordpress/compose`, the whole project (and its third-party consumers) would benefit from it.

As for the next steps, we could do the following:

1. It would great to pick up Riad's work from https://github.com/WordPress/gutenberg/pull/41001, and see if we can provide a new version of the hook based on modern Web technologies that satisfies the requirements of its consumers in the editor, providing a good, non-breaking migration strategy
2. In parallel, we can continue moving utils to a separate folder, private to the components package. For the "local" `useResizeObserver` utility discussed in this PR, we can add a TODO comment and remove it in favor of the `@wordpress/compose` version when we're done with the rewrite (see above point).

What do y'all think?

packages/components/src/utils/react.ts
ciampo: > we'll end up with about 8 new files

We could get away with 3 new files:

- one for `useEvent` & related types (which should be moved to `@wordpress/compose` as a follow up, [as discussed](https://github.com/WordPress/gutenberg/pull/62946/files#r1659004661))
- one for `useOnValueUpdate` & related types
- one for `useTrackElementOffsetRect` and the remaining hooks

> Anyway, if you think this is enough of a change to grant separate discussion and consensus, I'm happy to re-organize this as requested and leave it for another day. Let me know and I'll do it, but know I'm gonna be pushing for a simplification of the utilities directory in a follow-up discussion :P

I also agree that the `utils` folder is untidy and could do with better organization! But it's good to keep PR scope as small as possible.

Happy to move the conversation to a new issue and agree on next steps

packages/components/src/utils/react.ts
DaniGuardiola: @ciampo agreed on 1! for 2, unfortunately the hook will still not work for the reasons I stated in my first bullet point. I need an event to fire when the target element is first observed, both initially and subsequently, and I also don't need the overhead of the resize observer entries.

I think I should just rename the hook so that it can't be confused as an adapter for ResizeObserver (which it isn't), for which we already have one. It's a utility that does something, and happens to use the `ResizeObserver` browser API under the hood, as an implementation detail. Hopefully this makes sense.

ciampo: Sounds good to me, at least for the immediate action items

As we work on refactoring the `useResizeObserver` shared utility from `@wordpress/compose` we could also evaluate adding new features to it, like firing when the target element is first observed.

DaniGuardiola: I updated the hook's name to `useTrackElementRectUpdates` for clarity, and updated some of the docs. In the process of doing that, I discovered something interesting to consider in future ResizeObserver-related work: https://stackoverflow.com/a/60026394

.eslintrc.js
DaniGuardiola: Reverted change.

packages/components/src/utils/react.ts
DaniGuardiola: Reorganized utils into 3 files!

packages/components/src/utils/react.ts
DaniGuardiola: Took note of this! Resolving.

DaniGuardiola:
I addressed all comments, please re-review @ciampo @WordPress/gutenberg-components :)

packages/components/src/utils/element-rect.ts
ciampo, line 1: Should this file be renamed `use-track-element-offset-rec.ts` to match how other files are named, and moved to the `hooks` directory?

ciampo (approved):
LGTM 🚀  

Can be merged once open feedback is addressed.

---

As for the next steps, here's a recap of what I gathered from all conversations in this PR:

- move `useEvent` to the `@wordpress/compose` package
- resume work on https://github.com/WordPress/gutenberg/pull/41001, and see if we can provide a new version of the hook based on modern Web technologies that satisfies the requirements of its consumers in the editor, providing a good, non-breaking migration strategy
- [optional] review the structure and organisation of the `packages/components/src/utils/` folder
- refactor `ToggleGroupControl` to use the newly introduced utils to animate its indicator without using `framer-motion`

packages/components/src/utils/element-rect.ts
DaniGuardiola: It contains multiple utilities related to tracking and obtaining element rects, and they are a mix of hooks and other things, so it wouldn't make sense to do so without completely dislocating all of these utilities. It's why I raised the point that the utils folder just doesn't have a good structure.

DaniGuardiola:
@ciampo yep, got all of that on my backlog already 👍

packages/components/src/utils/react.ts
DaniGuardiola: A few updates here:

- https://github.com/WordPress/gutenberg/pull/64820
- https://github.com/WordPress/gutenberg/pull/64926
- https://github.com/WordPress/gutenberg/pull/64943

packages/components/src/utils/react.ts
DaniGuardiola: Update: https://github.com/WordPress/gutenberg/pull/64943

PR #60897, opened 2024-04-19
Tooltip: remove unecessary delay in tests except where needed

DaniGuardiola:
## What?

There was a need to add a delay after every Tooltip test. After an Ariakit update, it's not necessary anymore for every test. See discussion: https://github.com/WordPress/gutenberg/pull/57345#discussion_r1435167187

## Why?

Faster tests.

## How?

Removed delay after every test and only added where necessary (after one specific test).

## Testing Instructions

Run Tooltip unit tests.

### Testing Instructions for Keyboard

N/A

## Screenshots or screencast 

N/A

PR #54225, opened 2023-09-06
Migrate `Composite` component from `reakit` to `ariakit`

andrewhayward:
## What?

As per #53278, we need to move away from `reakit` to use `ariakit` instead. This PR tackles part of that, by starting the migration of the `Composite` component.

## Why?

> [Reakit](https://reakit.io/) does not explicitly support React 18, so it is causing peer dependency warnings.
> Reakit has already been succeeded by [Ariakit](https://ariakit.org/), so it will not receive any more updates.

`Composite` is currently a simple wrapper around `reakit`, re-exporting its various component parts. It needs to move away from this and export `ariakit` instead.

## How?

**This is the first step in migrating `Composite` away from `reakit`:**

 * Refactors usages of `reakit` to `ariakit` in `@wordpress/components`, but without modifying the version exported by the `@wordpress/components` package
 * Temporarily exports the `ariakit` version of `Composite` via the private APIs

**It does not tackle the following, which will be follow-up tickets:**

 * Iteratively refactor all usages of the `reakit` composite components across Gutenberg
 * Remove the `reakit` implementation, and use the `ariakit`-based version instead, removing the `ariakit` private API export

> [!NOTE]
> Even though `Composite` is [currently exported from the components package](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/index.ts#L59-L65) as `__unstable`, it is being used publicly, so we will have to proceed with caution!

## Testing Instructions

Despite the API changes, both [`AlignmentMatrixControl`](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-alignmentmatrixcontrol--docs) and [`CircularOptionPicker`](https://wordpress.github.io/gutenberg/?path=/docs/components-circularoptionpicker--docs) should continue to work as now.

PR #62947, opened 2024-06-27
Components: upgrade Ariakit to latest

DaniGuardiola:
Outdated PR, see https://github.com/WordPress/gutenberg/pull/64066

---

Moved over from personal repo, see original PR: #60992

Went through all breaking changes in all relevant versions. Only 0.4.0 had breaking changes (it's a "pseudo-major", so it makes sense). Here's a summary of those changes, TL;DR it doesn't seem like we need to do anything about them, and we just need to test enough to be confident in the upgrade.

| Change                                                                                                   | Notes                                                                                      |
| -------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ |
| [Improved animation support](https://ariakit.org/changelog#improved-animation-support)                   | n/a                                                                                        |
| [ComboboxList is no longer focusable](https://ariakit.org/changelog#comboboxlist-is-no-longer-focusable) | Only instance is `packages/dataviews/src/search-widget.tsx`, unaffected. |
| [Composite widgets with grid role](https://ariakit.org/changelog#composite-widgets-with-grid-role) | Inspected all instances, unaffected. |
| [Radio types have improved](https://ariakit.org/changelog#radio-types-have-improved) | Inspected all instances, unaffected. |
| [Public data attributes have now boolean values](https://ariakit.org/changelog#public-data-attributes-have-now-boolean-values) | Inspected all instances, unaffected. |
| [Removed deprecated features](https://ariakit.org/changelog#removed-deprecated-features) | All mentioned breakages would cause type errors. |

There is another change not considered "breaking" by Ariakit, but it deserves consideration:

Previously, Ariakit added `aria-describedBy` to tooltip anchors. Now, it doesn't, and users are responsible for making anchors accessible, see: https://ariakit.org/components/tooltip#tooltip-anchors-must-have-accessible-names

---

Remaining work:

- [x] Replace `render` in problematic tests with `@ariakit/test` version.
- [x] Upgrade `@ariakit/test` to latest version.
- [x] Progressively upgrade until latest version.
- [ ] Fix remaining failing tests (if any).
- [ ] Contribute `render` changes to upstream `@ariakit/test` (returning the output of the wrapped `ReactTestingLibrary.render` call, e.g. `rerender`, etc).
- [ ] Remove local temporary fork of `@ariakit/test`.
- [x] Remove dataviews type hack (see: https://github.com/WordPress/gutenberg/pull/62947/files#diff-66986a72a5a66aa4d65ca24ca5dc95d4b836836aa86630a10b9685b7a2fe12c5R261).
- [ ] Smoke test.
- [ ] Ensure that the animation change doesn't affect other components other than custom-select-control, custom-select-control-v2 and dropdown-menu-v2.
- [ ] Document testing procedure and context around `render` in appropriate place.
- [ ] Discuss idea: enforce `@ariakit/test`'s `render` in all components package tests? (e.g. `noRestrictedImports`)
- [ ] Discuss: "aria-describedBy" in tooltip anchors (see above)

---

Update 1 - Initial state of tests as of v0.3.13 (79fa2887735afe2dc8a30a85284242ccf74e182d):

image

After turning a couple of `getByRole` into `await findByRole` (179f3746d2ea603a565e38ab85ba653194a69578):

image

(sometimes it fails on 16 tests, not sure which one is the flaky one yet)

---

Update 2 - Still on v0.3.13. After merging trunk in, there are new failing tests (aebd13240b14944bb5da95053d07ae8d9b83b3dd):

image

Additionally, there's a new type error after unifying the version in `dataviews`:

image

This is due to this unification being essentially a downgrade, since the prop value is from a recent version:

image

For now, I'm doing a quick fix in order to keep upgrading version by version:

image

---

Update 3 - Still on v0.3.13. Using `render` from Ariakit in `custom-select-control/test/index.tsx` prevents `act` warnings, fixing one test (0c64d637a4e49e26d62c5f00ce0ce4d43b3e8ccf).

Before:

image

After:

image

The remaining test is related to this re-render issue, for which we're waiting on Ariakit to signal if a fix can be provided on their side: https://github.com/ariakit/ariakit/issues/3939#issuecomment-2251231903 

Current global state of tests:

image

(no change probably due to the flaky one)

---

Update 4 - Still on v0.3.13. Using `render` from Ariakit in `tabs/test/index.tsx` prevents `act` warnings, fixing seven tests  (3be0bc20dfda457277586c3f480a87739b3cd815).

Unfortunately, this also causes a few new errors due to the [rerender issue](https://github.com/ariakit/ariakit/issues/3939#issuecomment-2251231903), but those should be fixed once that's addressed at a fundamental level.

~There are also some remaining tests that weren't failing due to act warnings and still fail after this change. They will need to be reviewed individually to figure out what's going on with them.~ Apparently, not true. It was all fixed by proper rerendering. See next update.

Before:

image

After:

image

Current global state of tests:

image

Higher amount (+2 including flaky, which didn't fail this time around) due to reasons stated above.

---

Update 5 - Implemented `rerender` method (bac40268878fb9ae4e0b876920dc5f488b015087), and it works!! All tests in Tabs are fixed by this after updating to this API (11a24927f8547b23c09c01477c646db9013e9b58).

To achieve this, I temporarily copied the Ariakit `render` method into the codebase (68afa3c22f7636e8f5ef9c652aab98832799f43c), and modified to provide a way to re-render with the same safeguards as the original `render` (e.g. preventing `act` warnings). It is used like this:

image

After updating the Tabs tests, all of them now pass! 🎉 

image

Current global state of tests:

image

As you can see, with Tabs tests fixed, the number is WAY down.

---

Update 6 - Reverted manual fixes from update 1 (affecting) two files, and switch to the Ariakit `render` method instead (f8ada306c54480ec88a1864ff41008448e981028 and 375c5ca86470156fb36dab87acb8d386261e1a7e).

As there's now a fundamental fix, it's better to consistently use it instead of relying in manual fixes.

---

Update 7 - After updating remaining tests with the Ariakit `render` utility, all tests are now passing! (f4f5cb98a26b2b1e3bd71eb88d616a4338f92063 and 38383d19af9fa12048e8c863e06bacd80a031396)

image

---

Update 8 - The rest of the 0.3.x minors pass all tests, but tests start failing again after upgrading to 0.4.0 (7e7cb32e66e5913461dd01a0ff5082cb4353de57).

image

Oh well! Got more work cut out for me.

Full output: https://pastebin.com/raw/rDPQ1XFs

Note: `@ariakit/test` was upgraded to latest version, didn't break anything.

---

Update 9 - I found the root causes of all test failures and left TODO comments on each relevant place, see this diff: https://github.com/WordPress/gutenberg/pull/62947/files/4c022ed8d80fdce62795e3c0895fd3dfcbc08421..114badda5f68770118090baaf032c1a19b4cf84a

Here's a full list of causes, see linked diff above for details:

- Composite (legacy): outdated assertion was looking for a `''` value in a data attribute, but the right value is now `'true'`.
- CustomSelectControl (v1 & v2) and DropdownMenu v2: animation changes make the popovers stay open for longer after closing, making tests that depend on it being closed fail.
- ToggleGroupControl: snapshot outdated because of the data attribute value change.
- Tooltip: they don't support acting as description to their anchors anymore, making some test queries fail.

These findings cover ALL failing tests cases.

---

Update 10 - On 0.4.0 with all tests passing! (6679cc6db85541b90eded2afd3dfab947224443a)

image

---

Update 11 - Upgraded to 0.4.6 version by version, and all tests pass in each of them (6ed775dd3d04a682e91b9da68074e779de8dc7f0)

However, when upgrading to 0.4.7 there are some tests that now fail, see full output: https://pastebin.com/raw/0BAqt1th

image

Separately, the amount of deletions in the `package.json` seems extremely sus, so I'm gonna look into that as I might have messed something up along the way (I needed to do some dirty hacks to get npm to actually upgrade in some cases).

image

image

---

Update 12 - The package-lock file was indeed messed up in some way. After cleaning up and testing a variety of versions of Ariakit, it seems like those failures happen even down to the 0.4.0 versions. Will need to investigate more, but at least the mess has been cleaned up now and we're on the latest versionof all ariakit packages (c32b9a0d653e7dc9427a840736eea6ef41719217).

Still 5 test failures to go (2 of them seem flaky). Updated test output: https://pastebin.com/raw/fDZjTc4k


DaniGuardiola:
Since there are no evident breaking changes that affect us on changelog inspection, but we have many failing tests, I'm going to take another strategy. I will go version by version and addressing any errors step by step.

DaniGuardiola:
Also ping @WordPress/gutenberg-components if you can take a look at failing tests in CI and see if anything rings a bell or if you have any ideas. Since you're more familiar with the code, you might be able to save me some research in some cases.

ciampo:
Happy to pair on this one whenever you've got time to work on it

tyxla:
> Also ping @WordPress/gutenberg-components if you can take a look at failing tests in CI and see if anything rings a bell or if you have any ideas. Since you're more familiar with the code, you might be able to save me some research in some cases.

For some of the tests, it might be as easy as replacing the `@testing-library/user-event` library with `@ariakit/test`. For example, for the `should populate tooltip with description content for buttons with visible labels (buttons with children)` test, it works for me if I change:

```
await user.tab();
```

to:

```
import { press } from '@ariakit/test';
// ...
await press.Tab();
```

Note that historically when tabbing, we needed to do `await sleep();` before tabbing to quick-solve some flaky testing behavior. Hopefully after the update this is no longer the case and we can remove all those `await sleep();` instances.

ciampo:
> For some of the tests, it might be as easy as replacing the @testing-library/user-event library with @ariakit/test

That has definitely proven true for other ariakt-based components like `Tabs` and `DropdownMenuV2` (cc @diegohaz  )

DaniGuardiola:
I have some unpushed fixes like that, let me push soon and I'll let y'all know so we can figure out any remaining cases.

ciampo:
Flagging related conversation in https://github.com/WordPress/gutenberg/pull/63433#issuecomment-2222973211

ciampo:
After a pairing session with @DaniGuardiola , we determined that the test failures caused by the `act(...)` warnings are caused at the ariakit level. We opened https://github.com/ariakit/ariakit/issues/3939

diegohaz:
Quoting my [comment](https://github.com/WordPress/gutenberg/pull/60992#issuecomment-2080093840) from the other PR to ensure it doesn't get lost in the conversation:

> Looks like most of the failing tests are due to a change on `data-*` attributes: [Public data attributes have now boolean values](https://ariakit.org/changelog#public-data-attributes-have-now-boolean-values)
> 
> Also, tooltips no longer assign `aria-labelledby`/`aria-describedby` attributes: [ariakit/ariakit#3318](https://github.com/ariakit/ariakit/pull/3318)
> 
> Other failures appear to prompt `act` warnings, potentially indicating an internal change in sync/async stuff. Typically, replacing `getBy` with `findBy`, or `waitFor` a change on the screen before making an assertion resolves the problem. Alternatively, you can use `@ariakit/test` events, which may also be beneficial and prevent these warnings from recurring since it doesn't depend on `act`. More context: [#54225 (comment)](https://github.com/WordPress/gutenberg/pull/54225#issuecomment-1711949971)

ciampo:
Also quoting this other [suggestion](https://github.com/ariakit/ariakit/issues/3939#issuecomment-2225472231) from Diego:

> With RTL, you will need to `waitFor` something in the DOM immediately after calling `render`.
> 
> In Ariakit, we use our own `render` utility from `@ariakit/test/react`, which is asynchronous and accounts for those timers.

tyxla:
> Also quoting this other [suggestion](https://github.com/ariakit/ariakit/issues/3939#issuecomment-2225472231) from Diego:
> 
> > With RTL, you will need to `waitFor` something in the DOM immediately after calling `render`.
> > In Ariakit, we use our own `render` utility from `@ariakit/test/react`, which is asynchronous and accounts for those timers.

Have we tried this already, and whether it has an effect on the `act` failures?

Also, it looks like the branch is quite behind and needs to be rebased. Some of the failures are in the v1 `CustomSelectControl` tests, which are no longer in `trunk`.

DaniGuardiola:
@tyxla unless I missed something, the branch was synced with trunk yesterday.

`waitFor` works in some instances; in others, it's a bit unclear what we should be waiting for that makes sense.

I'm also trying our `render` locally and it seems to work in a few instances I tested, with a few remaining things to look into (like re-renders).

ciampo:
> in others, it's a bit unclear what we should be waiting for that makes sense.

out of experience, we'd usually wait for some `data-` or `aria-` attribute added by ariakit that would signify that the internal state was initialised and applied to the DOM.

tyxla:
> > in others, it's a bit unclear what we should be waiting for that makes sense.
> 
> out of experience, we'd usually wait for some `data-` or `aria-` attribute added by ariakit that would signify that the internal state was initialised and applied to the DOM.

This. Or historically it's been `tabindex`. Here's some prior art:

https://github.com/WordPress/gutenberg/blob/decaae66b0e4ca173b1bbb1ee7a059819074a918/packages/components/src/alignment-matrix-control/test/index.tsx#L20-L29

ciampo:
Also adding a link to [this convo](https://github.com/WordPress/gutenberg/pull/60897#discussion_r1572248641) that I found while reviewing old notifications — we should be able to remove that timer in the tooltip tests after updating to latest ariakit.

DaniGuardiola:
@ciampo yep, already on my backlog as a follow-up to this, thanks!

image

DaniGuardiola:
After investigation, it does NOT seem like the ariakit `render` utility supports rerenders properly: https://github.com/ariakit/ariakit/issues/3939#issuecomment-2251231903

Waiting on Diego to know if supporting re-renders is a possibility, since that'd help us fix tests where we depend on it.

DaniGuardiola:
Update 5 (see description) is a breakthrough! With re-render support in my temporary fork of Ariakit's `render`, all tests so far have been fixed :)

Next up: apply the re-render API to other tests to see how far we can get with it.

DaniGuardiola:
Gonna use the local fork for now, but I'm aiming to contribute this to upstream Ariakit so we don't depend on our local fork, see https://github.com/ariakit/ariakit/issues/3939#issuecomment-2251499387

DaniGuardiola:
image

DaniGuardiola:
See last update (update 8) - 0.4.0 version makes a bunch of tests fail.

tyxla:
> See last update (update 8) - 0.4.0 version makes a bunch of tests fail.

At a first glance, it seems like something changed with the ariakit test helpers like `click`, `press`, etc., where changes are not immediately reflected and our tests are not prepared for that change. A fix can be straightforward by providing the test some wait time where necessary.

For example, I've pushed 49a959f55fd468fe29315b6833d9f02093f4301a, which should resolve the `CustomSelectControl` test failures. Similar fixes should resolve many, if not most, of the failures.

DaniGuardiola:
See update 9, I found the root causes to all test failures on v0.4.0 🎉 

Edit: fixed now and all tests pass, see update 10.

mirka:
> All mentioned breakages would cause type errors.

Unfortunately a large portion of the Gutenberg app is not type checked 😅 (`@wordpress/block-editor`, `@wordpress/edit-site`, etc)

DaniGuardiola:
@mirka good point. Fortunately, I think most of these features were not exposed by components in the first place (child render function, `as` prop, etc). With the quoted comment I was referring to the components package itself, which afaik is typed for the most part, at least in the parts where we interact with Ariakit.

That said, I will do another exhaustive pass to see if there's something missing in regards to this.

packages/components/src/tooltip/test/index.tsx
```diff
@@ -65,36 +65,6 @@ describe( 'Tooltip', () => {
...
-			).toBeInTheDocument();
-
-			// Hover outside of the anchor, tooltip should hide
-			await hoverOutside();
-			await waitExpectTooltipToHide();
-			expect(
-				screen.queryByRole( 'button', { description: 'tooltip text' } )
-			).not.toBeInTheDocument();
-		} );
-
```
ciampo, line 97: Is this test going to be restored before merging this PR?

packages/components/src/tooltip/test/index.tsx
```diff
@@ -493,11 +463,6 @@ describe( 'Tooltip', () => {
 			expect(
 				screen.queryByRole( 'tooltip', { name: 'Inner tooltip' } )
 			).not.toBeInTheDocument();
-			expect(
-				screen.getByRole( 'button', {
-					description: 'Outer tooltip',
-				} )
-			).toBeVisible();
```
ciampo, line 500: Why was this assertion deleted?

packages/components/src/composite/legacy/test/index.tsx
```diff
@@ -47,7 +47,7 @@ async function renderAndValidate( ...args: Parameters< typeof render > ) {
 		const activeButton = queryByAttribute(
 			'data-active-item',
 			view.baseElement,
-			''
+			'true'
```
ciampo, line 50: With the new render function, could we actually remove the `renderAndValidate` utility in this file?

ciampo (commented):
Great work so far, @DaniGuardiola ! I left some comments, hopefully I didn't miss some context.

That custom render function looks scary, hopefully we'll manage to move that code either into `@ariakit/test` or `@testing-library` 🤞

packages/components/src/tooltip/test/index.tsx
DaniGuardiola: No, unless we want to re-implement this functionality (I don't think we should), see https://github.com/WordPress/gutenberg/pull/62947/files/4c022ed8d80fdce62795e3c0895fd3dfcbc08421..114badda5f68770118090baaf032c1a19b4cf84a#diff-eb9fcb45579ca92e96f172d835623c0682387b4ec0ff0e7ed5009b958b7c2c5c

packages/components/src/tooltip/test/index.tsx
DaniGuardiola: Same reason as above, see https://github.com/WordPress/gutenberg/pull/62947/files/4c022ed8d80fdce62795e3c0895fd3dfcbc08421..114badda5f68770118090baaf032c1a19b4cf84a#diff-eb9fcb45579ca92e96f172d835623c0682387b4ec0ff0e7ed5009b958b7c2c5c

packages/components/src/composite/legacy/test/index.tsx
DaniGuardiola: Potentially!

DaniGuardiola:
Thanks @ciampo! I'm moving to https://github.com/WordPress/gutenberg/pull/64066 because it was better to start clean due to the annoying npm issue.

@WordPress/gutenberg-components this PR can still be useful to reference old information I left here, but I tried to make the new PR "complete enough" in terms of context needed to understand what's going on. I also ensured that CI unit tests ran completely for the versions with failures, so that we can check those logs if necessary and for the record. Plus the version-by-version commit history should help in future `git bisect`s.

See you there!

PR #62948, opened 2024-06-28
Ensure disable overrides button is active for image blocks with captions or links

talldan:
## What?
In #62747, the 'Enable Overrides' button was disabled for image blocks with captions or links within patterns, as they're not supported:
Screenshot 2024-06-28 at 12 37 09 pm

The code for this change also incorrectly applies the same treatment to the 'Disable overrides' version of the button when such a block already has overrides (a user would've been able to add these in the gutenberg plugin or WP6.6 beta before that PR was merged):
Screenshot 2024-06-28 at 12 39 08 pm

These users should still be able to remove overrides.

## How?
Makes a small adjustment to the conditions for the button to also check the `hasOverrides` variable.

## Testing Instructions
1. Create a synced pattern
2. Switch to the code editor mode and add the following markup:
```html
<!-- wp:image {"aspectRatio":"16/9","scale":"cover","sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large"><img src="https://pd.w.org/2023/07/44364b18862589f06.53436652.jpg" alt="" style="aspect-ratio:16/9;object-fit:cover"/><figcaption class="wp-element-caption">Test</figcaption></figure>
<!-- /wp:image -->
<!-- wp:image {"aspectRatio":"16/9","scale":"cover","sizeSlug":"large","linkDestination":"none","metadata":{"bindings":{"__default":{"source":"core/pattern-overrides"}},"name":"image"}} -->
<figure class="wp-block-image size-large"><img src="https://pd.w.org/2023/07/44364b18862589f06.53436652.jpg" alt="" style="aspect-ratio:16/9;object-fit:cover"/><figcaption class="wp-element-caption">Test</figcaption></figure>
<!-- /wp:image -->
```
3. Select the first image block. It doesn't have overrides enabled, but because it has captions it should show this button in the advanced inspector tools:
Screenshot 2024-06-28 at 12 53 40 pm
4. Select the first image block. It *does* have overrides enabled and even though it has captions it allows disabling overrides in the advanced inspector tools:
Screenshot 2024-06-28 at 12 55 08 pm

### DIFF
--- a/packages/patterns/src/components/pattern-overrides-controls.js
+++ b/packages/patterns/src/components/pattern-overrides-controls.js
@@ -44,7 +44,7 @@ function PatternOverridesControls( {
 
 	const hasName = !! attributes.metadata?.name;
 	const defaultBindings = attributes.metadata?.bindings?.__default;
-	const allowOverrides =
+	const hasOverrides =
 		hasName && defaultBindings?.source === PATTERN_OVERRIDES_BINDING_SOURCE;
 	const isConnectedToOtherSources =
 		defaultBindings?.source &&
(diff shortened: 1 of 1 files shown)


talldan:
I considered whether this one should be backported. I think it should. Even though only plugin/beta users will be able to reproduce, plugin users might decide to deactivate the plugin, and I think core should still be able to handle this case.

kevin940726 (approved):
Tested and worked as expected! 💯

PR #62949, opened 2024-06-28
List: maintain nested list on parent item removal

ellatrix:
## What?

Best reviewed [without whitespace changes](https://github.com/WordPress/gutenberg/pull/62949/files?diff=split&w=1).

Currently, when removing a parent item, the nested list items are appended at the bottom of the list instead of preserved at the position.

## Why?

Note that right now, in trunk, not only does the indentation disappear, the indented list item is also moved to the bottom! If you have a long nested list, it feels like your content _disappears_.

|Before|After|
|-|-|
|list-before|list-after|

## Testing Instructions

See GIF. Create this setup:

* 1
    *  
    * 2
* 3

And press Backspace twice from the empty list item.


youknowriad (approved):
LGTM 👍

PR #62951, opened 2024-06-28
Site Editor: Fix template parts 'Reset' action

### LINKED ISSUE #62604: Missing "Reset" from template part Inspector actions
You can reset templates from the vertical ellipsis, but not template parts. 

I'd expect to be able to "Reset" template parts, as I can templates. 

As seen on templates: 

CleanShot 2024-06-16 at 16 07 33

As seen on template parts: 

CleanShot 2024-06-16 at 16 01 18

Mamaduka:
## What?
Fixes #62604.

PR fixes the "is eligible" condition for template parts and makes the "Reset" action work properly.

## Why?
The action was no longer available for template part entities.

## Testing Instructions
1. Open the Site Editor.
2. Make changes to the template parts provided by the theme.
3. Confirm you can reset template parts when editing one.
4. Confirm you can reset one or more template parts in data views.
5. Confirm that the reset action isn't available for parts created by the users.

Same.

## Screenshots or screencast 
**Editor sidebar**
CleanShot 2024-06-28 at 09 19 35

**DataViews**
CleanShot 2024-06-28 at 09 19 18


packages/editor/src/store/private-actions.js
```diff
@@ -241,7 +243,12 @@ export const revertTemplate =
...
-		if ( ! isTemplateRevertable( template ) ) {
+
+		const isRevertable =
+			template.type === TEMPLATE_PART_POST_TYPE
+				? isTemplatePartRevertable
+				: isTemplateRevertable;
+		if ( ! isRevertable( template ) ) {
 			registry
 				.dispatch( noticesStore )
 				.createErrorNotice( __( 'This template is not revertable.' ), {
```
Mamaduka: The message needs to reflect the `type`, but the fix needs to be backported into RC, and we're currently in the "hard string-freeze" phase.

I'll handle string updates separately.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -896,47 +884,39 @@ const resetTemplateAction = {
...
-		);
+		const { revertTemplate } = unlock( useDispatch( editorStore ) );
 		const { saveEditedEntityRecord } = useDispatch( coreStore );
 		const { createSuccessNotice, createErrorNotice } =
 			useDispatch( noticesStore );
 		const onConfirm = async () => {
 			try {
-				if ( items[ 0 ].type === TEMPLATE_PART_POST_TYPE ) {
-					await removeTemplates( items );
-				} else {
```
Mamaduka, line 829: @jorgefilipecosta, this special condition was introduced in #61612, but I couldn't find any details on why it was necessary.

Calling `removeTemplates` when resetting a template part from the Site Editor causes a missing entity error.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -59,11 +60,6 @@ function isTemplateRemovable( template ) {
 		! template.templatePart?.has_theme_file
 	);
 }
-const canDeleteOrReset = ( item ) => {
-	const isTemplatePart = item.type === TEMPLATE_PART_POST_TYPE;
-	const isUserPattern = item.type === PATTERN_TYPES.user;
-	return isUserPattern || ( isTemplatePart && item.isCustom );
-};
```
Mamaduka: I've removed this check.

## Why?

The "Reset" action is only available for the `wp_template` and `wp_template_part` entities.

richtabor (approved):
Works as expected; just need to clear the conflict. Thanks!

Mamaduka:
@youknowriad, I think your #63017 PR could supersede this unless there's another 6.6 RC; in that case, this probably should go in first.

Additionally, some of the notes from this PR could also apply there.

cc @ellatrix

ellatrix:
Looks solid to me, this is good to include. Would be good to have an extra check from @jorgefilipecosta

packages/editor/src/store/utils/is-template-part-revertable.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+import { TEMPLATE_ORIGINS } from '../constants';
+
+/**
+ * Check if a template part is revertable to its original theme-provided template file.
+ *
+ * @param {Object} templatePart The template part entity to check.
+ * @return {boolean} Whether the template part is revertable.
+ */
+export default function isTemplatePartRevertable( templatePart ) {
```
youknowriad: There's already a function called isTemplateRevertable which I believe is the exact same function (we can reuse it by naming the argument templateOrTemplatePart

youknowriad:
I'm happy to update my PR after this one lands. I don't think I changed the logic much in my PR

packages/editor/src/store/utils/is-template-part-revertable.js
Mamaduka: Updated in [ab9e7b1](https://github.com/WordPress/gutenberg/pull/62951/commits/ab9e7b145e84a743fbabc93f98c40494b5687856).

packages/editor/src/components/post-actions/actions.js
jorgefilipecosta: Hi @Mamaduka,

In https://github.com/WordPress/gutenberg/pull/61612 I implemented things this way to replicate the logic that already existed before for template part reset.

Previously when a template part reset happened the template part post of the user was deleted it was always the case so I kept the behavior. 
With this change instead of deleting we keep the user template part and change it to be equal to what the template set. This may cause issues during updates etc.
Previously @youknowriad even said that maybe even for templates during a reset we should just delete the user template too. I'm not sure if going in the direction of keeping the user entity also for template parts is a good idea.

jorgefilipecosta: Update: In my tests, it seems the template part is still deleted as expected so things look good. I'm checking how it happens given that revertTemplate at least previously did not apply a delete 🤔

packages/editor/src/store/private-actions.js
```diff
@@ -241,6 +241,7 @@ export const revertTemplate =
 	async ( { registry } ) => {
 		const noticeId = 'edit-site-template-reverted';
 		registry.dispatch( noticesStore ).removeNotice( noticeId );
+
```
ellatrix: Can be good to clean this up, in case it helps to avoid cherry-pick conflicts.

packages/editor/src/components/post-actions/actions.js
Mamaduka: Checking the code from `wp/6.5` branch. It looks like that the editor sidebar was using the [`revertTemplate`](https://github.com/WordPress/gutenberg/blob/39b57b0c63d91c6dc1b1b9998c441f22b4cdfb61/packages/edit-site/src/components/sidebar-edit-mode/template-panel/template-actions.js#L44) while "DataViews" was using the [`removeTemplates`](https://github.com/WordPress/gutenberg/blob/7ebb55ca27e7bdf6275e73ff75de608158127533/packages/edit-site/src/components/page-patterns/dataviews-pattern-actions.js#L215) action.

The problem with using `removeTemplates` while viewing the same entity in canvas is that it also deletes the entity from the store and requires page refresh to load it.

Here's what happens when I restore old behavior:

https://github.com/WordPress/gutenberg/assets/240569/01db00d1-8f68-48b1-9c91-5f812c5ba5ba

jorgefilipecosta: On revert templates when we pass source: 'theme', to edit an entity it is in fact an instruction to delete the entity from the database. I will add a comment there to make the code easier to read in the future, I was not aware of that.

But yah we should apply this simplification as we can use the use behavior on both.

jorgefilipecosta (approved):
Works well and the changes look good, thank you for fixing this 👍

packages/editor/src/components/post-actions/actions.js
Mamaduka: Thank you, @jorgefilipecosta!

Mamaduka:
@ellatrix, here's the manual backport PR - #63084.

PR #62952, opened 2024-06-28
Ensure device previews extra scrollbar only appears when needed

talldan:
## What?
A small improvement to go alongside #62940.

I noticed even though that PR removes the third scrollbar, there are still often two even when the device fully fits in the browser window, and scrolling the second one doesn't do anything useful.

## Why?
The issue seems to be with the `.block-editor-iframe__scale-container`, which always extends beyond the browser viewport when Mobile/Tablet view is active. 

## How?
I'm no css expert, but adding `display: flex` is an easy way to make it fit in the screen, with no obvious repercussions.

## Testing Instructions
1. Test both the post and the site editor with long post style content
2. Activate the mobile/tablet views ensuring your browser viewport is bigger than the views
3. Check that there's only one scrollbar alongside the content
4. Now reduce the size of your browser window to be smaller than the device view
5. A second scrollbar should appear allowing you to view the full device.

Also do a general smoke test of things like:
- Editing patterns
- Style book
- Code editor mode
- The Widget editor

## Screenshots or screencast 
Screenshot 2024-06-28 at 3 38 47 pm

### DIFF
--- a/packages/block-editor/src/components/iframe/content.scss
+++ b/packages/block-editor/src/components/iframe/content.scss
@@ -11,6 +11,7 @@
 .block-editor-iframe__scale-container {
 	width: 100%;
 	height: 100%;
+	display: flex;
 }
 
 .block-editor-iframe__scale-container.is-zoomed-out {


stokesman:
> there are still often two even when the device fully fits in the browser window

I’m glad you noticed too. I tested this, it does what it says on the tin and I would count it as an improvement.

Yet another thing that I was considering to be part of the issue is that the device previews put a margin around the iframe that presumably should be visible on all sides (as it is in WP 6.5). With this branch the bottom margin isn’t included in the overflow.

Here’s 6.5:

https://github.com/WordPress/gutenberg/assets/9000376/56225958-13f3-4d85-8018-85176ce07795

Here’s this branch:

https://github.com/WordPress/gutenberg/assets/9000376/d51ccc2b-0341-4cfc-9daa-ad3fae3a2d3d

So this isn’t quite there if we’d like to match 6.5. I already had a branch for that but so far it’s considerably less simple.

---- 

Beyond that, there are some design considerations that perhaps should come into play but they may lead to a intentional departure from how device previews shipped in 6.5 and therefore maybe have to wait because, as I understand it, this late in the 6.6 release cycle it’s not supposed to be done. Yet I'll expound a bit on that anyway.

First off, that bottom margin doesn’t seem very valuable. It’s also inconsistent with how "focus" mode does a similar thing but doesn’t apply it to the bottom (and does it with `padding` instead and on a different element). So we could make the choice to remove the bottom margin. Ideally, we might also try to unify the how the "margin" is applied in both focus and device preview modes.

Then another consideration is that horizontal overflow is actually prevented in device previews by resizing the iframe horizontally. I wonder if it wouldn’t be a better to do the same vertically. That way a bottom margin would always be visible and there’d be no reason for a second scrollbar. I noticed that this branch can be made that way with just one more line of CSS. I’m including a demo to illustrate despite that it may be outside of what we can consider for this branch.

https://github.com/WordPress/gutenberg/assets/9000376/fc41ab11-acad-4e3b-b0b8-8092dba26ab0

stokesman (approved):
Despite how I pointed out this doesn’t quite restore how device previews were in 6.5 it’s still a very simple fix for part of it and I think it’d be worthwhile. It didn’t cause any unintended effects that I could find. I tested zoom out mode and it seems unaffected.

I did tinker more on restoring the device previews just as they were in 6.5 and #62989 is the most minimal way I found. I left it drafted because I'm calling it a night and probably should add a bit more detail to explain some some changes.

talldan:
Thanks @stokesman. Lets get this one in for 6.6, and then follow up with a more considered fix for 6.7. 👍

stokesman:
> Yet another thing that I was considering to be part of the issue is that the device previews put a margin around the iframe that presumably should be visible on all sides (as it is in WP 6.5)

> Lets get this one in for 6.6, and then follow up with a more considered fix for 6.7. 👍

Just wanted to note that after #63724, the margin around device previews is fully scrollable as it was in 6.5 🎉

PR #56875, opened 2023-12-07
Block editor: hooks: manage BlockListBlock filters in one place

ellatrix:
## What?

_[Best checked without whitespace](https://github.com/WordPress/gutenberg/pull/56875/files?diff=split&w=1)_

Similar to #56862, but for the BlockListBlock filters.
Related to https://github.com/WordPress/gutenberg/discussions/51005.

I've shaped the internal API around "block props". These filters never add more than wrapper props to the block, or in the case of duotone and layout, add editor styles and a class name.

To do: add the rest of them :)

## Why?

Reduces code duplication.
Allows us to manage the filter and do performance optimisations in one place.

PR #55288, opened 2023-10-12
useBlockPreview: Try outputting EditorStyles to ensure local style overrides are rendered

andrewserong:
## What?

Fixes: #54956

Explores an alternative to #55241 that instead of syncing style overrides, tries to output `EditorStyles` instead. This PR fixes an issue where Duotone styles set within a Post Template block are not rendered in non-selected instances of a Query Loop within the post or site editors. It also fixes an issue where Duotone styles in Safari were not updating for these previews, either.

## Why?

Prior to this PR, for things that use `useBlockPreview`, the styles that are generated within that preview were not output anywhere. This is because `ExperimentalBlockEditorProvider` uses its own registry, so is designed to have values passed down to it, but not to propagate values outside of itself. So any calls to `setStyleOverride` would update state within the local block editor provider, and not be captured by the parent's `EditorStyles`.

The fix in this PR is to output an `EditorStyles` component within the preview, and ensure that the local-to-the-preview block editor state only contains local styles, so we avoid and duplicates between the preview and the parent block editor environment.

There is a small drawback that the `style` and `svg` tags in this PR are rendered as a direct child of the post template block's `li` element, however based on feedback on this PR and #55241, it seems that this drawback is not as bad as the difficulty of accurately syncing between `overridesStyles` states, which was explored in #55241.

This PR is currently my preferred approach to try to fix this bug.

## How?

* Try using `EditorStyles` to output styles for the `useBlockPreview`'s local styles instead of syncing as in #55241.
* In the `useBlockPreview`'s settings that get passed down to `ExperimentalBlockEditorProvider`, strip it of `styles` so that any styles applied in the parent environment are not duplicated in the child preview's editor styles. This ensures that only styles generated within that preview are output via `EditorStyles`.
* In the Duotone hook, add some logic to repaint the block that uses Duotone styles, if the current session is running Safari. While we usually like to avoid browser sniffing, this ensures that the slight performance hit is only encountered in browsers where the repaint hack is needed.

## Testing Instructions

1. Add a couple of posts with a featured image set.
2. In another post, page, or template, add a Query Loop / Post Template block containing featured images and at least 2 items per page (you can use the test markup below).
3. In the editor, prior to this PR, only the selected item within the Post Template will render Duotone styles.
4. With this PR applied, each item in the Post Template loop should render Duotone styles.
5. If you go to adjust the Duotone styles for any of the blocks in the loop, it should update in real-time (with a slight delay as styles are propagated).
[...]

PR #62953, opened 2024-06-28
Duotone: fix code typo

ellatrix:
## What?

Better late than never, I saw the comment by @jsnajdr https://github.com/WordPress/gutenberg/pull/56875#discussion_r1611399988

I mistakenly passed the instance ID as the client ID, which broke the fix from https://github.com/WordPress/gutenberg/pull/55288 by @andrewserong.

## Why?

Restores the fix from #55288.

## Testing Instructions

See #55288.

### DIFF
--- a/packages/block-editor/src/hooks/duotone.js
+++ b/packages/block-editor/src/hooks/duotone.js
@@ -314,7 +314,7 @@ function useDuotoneStyles( {
 	}, [ isValidFilter, blockElement ] );
 }
 
-function useBlockProps( { name, style } ) {
+function useBlockProps( { clientId, name, style } ) {
 	const id = useInstanceId( useBlockProps );
 	const selector = useMemo( () => {
 		const blockType = getBlockType( name );
@@ -362,7 +362,7 @@ function useBlockProps( { name, style } ) {
 	const shouldRender = selector && attribute;
 
 	useDuotoneStyles( {
-		clientId: id,
+		clientId,
 		id: filterClass,
 		selector,
 		attribute,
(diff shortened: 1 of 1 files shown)


jsnajdr (approved):
Thank you for following up on this :tada:

andrewserong (commented):
Thanks for the ping!

This fixes the typo, but I think there might still be a bug in `useDuotoneStyles`. The fix in  https://github.com/WordPress/gutenberg/pull/55288 had `colors` in the dependency array for the `useEffect` that updates `blockElement` to force a re-render in Safari.

It looks like right now we only have `isValidFilter` and `blockElement` in that dependency array, so if we go to update the Duotone colors in Safari, that `useEffect` isn't called while updating, so the old bug where sometimes Safari doesn't rerender is present:

https://github.com/WordPress/gutenberg/assets/14988353/15f8746a-3df5-493a-a49b-5231399e3091

Here's the current dependency array:

https://github.com/WordPress/gutenberg/blob/59d0cf66caaac60c6156c579b1019c2d8fd7112f/packages/block-editor/src/hooks/duotone.js#L314

If I add `colors` to it locally, then retesting in Safari with the rest of this PR, the Duotone colors update reliably for me:

https://github.com/WordPress/gutenberg/assets/14988353/28c8a42c-5a81-4478-ad79-036e1c286d89

What do you think?

ellatrix:
@andrewserong Thanks! The strange thing is that it works the first few time, and then it stop working. I did as you suggested and now it works consistently.

andrewserong:
Thanks Ella! 🙇

PR #62956, opened 2024-06-28
Block Bindings: Replace `useSelect` hook that calls no selectors with `useMemo`

talldan:
## What?
Small PR that updates what I think was originally a typo, replacing `useSelect` with `useMemo`

## Why?
I couldn't see any reason to use `useSelect`. Either the selectors were refactored away, or it was intended to be a `useMemo`.

## Testing Instructions
There's no impact, apart from maybe some very minor performance gains?

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -109,7 +109,7 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 				),
 			[ props.attributes.metadata?.bindings, name ]
 		);
-		const boundAttributes = useSelect( () => {
+		const boundAttributes = useMemo( () => {
 			if ( ! bindings ) {
 				return;
 			}


talldan:
Tests are failing, so that indicates this breaks things. 🤔 

edit: though only one test.

talldan:
It looks like `useSelect` is needed, as it ensures `boundAttributes` is updated whenever attribute value changes occur. The `source.getValue` call may `registry.select` to get an attribute value, but that won't happen if it's `useMemo`

SantosGuillamot:
I was just gonna share something similar 🙂 Thanks a lot for exploring this option! Maybe it is worth adding a comment in the code as well?

talldan:
> I was just gonna share something similar 🙂 Thanks a lot for exploring this option! Maybe it is worth adding a comment in the code as well?

Yep, I'll make a small PR that does that 👍 

Done in #63005

Issue #62679, opened 2024-06-19
Core block styles are overriding theme.json styles in WordPress 6.6


ndiego:
### Description

If you have applied styles to specific elements in `theme.json`, perhaps blocks as well, there is a chance that they will be overridden by default block styles. I have not had a chance to test this issue widely, but I discovered (with the help of @bgardner) that the element styles in `theme.json` for Image block captions are no longer respected in 6.6. 

Take a look at the screenshots below. You will see that the specificity changes in https://github.com/WordPress/gutenberg/pull/61638 seem to have swapped the style order. This causes the block style to override `theme.json`. In my opinion, `theme.json` should always "beat" default block styles. 

### Step-by-step reproduction instructions

1. Use the Twenty Twenty-Four theme and WordPress 6.6 Beta 3+
2. Add an image with a caption to the page
3. In `theme.json`, add the following to the caption element

```
"spacing": {
	"margin": {
		"top": "10px",
		"bottom": "0"
	}
}
```

4. View the image on the front end and see that the block styles override for margin override those set in the theme.json

### Screenshots, screen recording, code snippet

| WordPress 6.5 | WordPress 6.6 |
|-|-|
|<img width="715" alt="image" src="https://github.com/WordPress/gutenberg/assets/4832319/09dd1741-f695-4a0b-a77b-52f963981b23">|<img width="715" alt="image" src="https://github.com/WordPress/gutenberg/assets/4832319/fa8a4d01-fc07-4dcf-93c8-7ee9c06a0c03">|

### Environment info

- WordPress 6.6 Beta 3+
- Gutenberg 18.5

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

ndiego:
cc @ellatrix @aaronrobertshaw 🙏

bgardner:
I can concur this is happening, as the example @ndiego used came from work I was doing earlier today. I have experienced other similar cases of core block styles overriding theme.json as well, so it's not limited to what he mentioned above.

bgardner:
Another example of this can be seen with css styles I have applied to Post Comments Form block in theme.json (https://github.com/bgardner/powder/blob/main/theme.json#L369) which now require the use of !important to ensure the border-color is being applied properly. For the sake of demo, here's a screenshot with the override happening after I have omitted the !Important in the inspector. The textarea should be #0a0a0a (contrast color) but it's now #949494.

[...]

PR #62960, opened 2024-06-28
Styles specificity: allow comment form input overrides

ellatrix:
## What?

Similar to #63049.

See https://github.com/WordPress/gutenberg/issues/62679#issuecomment-2180421157

Apparently, global styles allows for overrides with completely custom CSS, which means comment form input and textarea CSS should be possible to override. The problem of course is that the global styles specificity is forced at 0-1-0, while the default styles are higher. This PR lowers them.

Knowing that basically any CSS property can be modified with global styles through custom CSS makes me wonder if we should reduce _all_ block styles to 0-1-0.

## Why?

Fixes remaining issue from #62679.

## How?

Lower specificity to 0-1-0.

## Testing Instructions

Ensure this theme rule works:

```
:root :where(.wp-block-post-comments-form textarea, .wp-block-post-comments-form input) {
    border: 1px solid red;
}
```

image

### DIFF
--- a/packages/block-library/src/post-comments-form/style.scss
+++ b/packages/block-library/src/post-comments-form/style.scss
@@ -1,3 +1,18 @@
+// Allow these default styles to be overridden by global styles.
+:where(.wp-block-post-comments-form) {
+	textarea,
+	input:not([type="submit"]) {
+		border: 1px solid $gray-600;
+		font-size: 1em;
+		font-family: inherit;
+	}
+
+	textarea,
+	input:not([type="submit"]):not([type="checkbox"]) {
+		padding: calc(0.667em + 2px); // The extra 2px is added to match outline buttons.
+	}
+}
+
 .wp-block-post-comments-form {
 	// This block has customizable padding, border-box makes that more predictable.
 	box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


packages/block-library/src/post-comments-form/style.scss
```diff
@@ -1,3 +1,18 @@
...
+:where(.wp-block-post-comments-form) {
+	textarea,
+	input:not([type="submit"]) {
+		border: 1px solid $gray-600;
+		font-size: 1em;
+		font-family: inherit;
+	}
+
+	textarea,
+	input:not([type="submit"]):not([type="checkbox"]) {
```
tellthemachines: These chained `:not()` will make the specificity of this rule be 0-2-1. If the goal is to level all the rules at 0-1-0, it might be better to wrap these nested selectors in `:where()` instead of the outer selector.

tellthemachines:
Thanks for the PR! Could you add some steps to test so we can make sure we're all looking at the same problem? 😅

talldan (approved):
Fixes the immediate issue at hand. To test I updated my theme.json to have the following:
```json
"core/post-comments-form": {
  "css": "& textarea, input{border-radius:.33rem; border: 5px solid red;}"
},
```

In trunk I'm seeing the input not receive the border style. In this PR it is.

> Knowing that basically any CSS property can be modified with global styles through custom CSS makes me wonder if we should reduce all block styles to 0-1-0.

Yep, I wonder about this too.

PR #62965, opened 2024-06-28
Format Library: Clean up 'Highlight' format components

Mamaduka:
## What?
PR makes the following updates to the "Highlight" format components:

* Fix React ESLint warnings. These are always helpful, especially after we land #61788.
* Removes unnecessary callback memoization. None of the components require memoized callbacks.

## Why?
It's just a routine maintenance.

## Testing Instructions
1. Open a post or page.
2. Insert a paragraph block.
3. Add some text.
4. Confirm text highlighting works as before.

The feature also has test coverage.

Same.

## Screenshots or screencast 
CleanShot 2024-06-28 at 13 51 12

### DIFF
--- a/packages/format-library/src/text-color/index.js
+++ b/packages/format-library/src/text-color/index.js
@@ -2,7 +2,7 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useCallback, useMemo, useState } from '@wordpress/element';
+import { useMemo, useState } from '@wordpress/element';
 import { RichTextToolbarButton, useSettings } from '@wordpress/block-editor';
 import {
 	Icon,
@@ -66,21 +66,13 @@ function TextColorEdit( {
 		'color.palette'
 	);
 	const [ isAddingColor, setIsAddingColor ] = useState( false );
-	const enableIsAddingColor = useCallback(
-		() => setIsAddingColor( true ),
-		[ setIsAddingColor ]
-	);
-	const disableIsAddingColor = useCallback(
-		() => setIsAddingColor( false ),
-		[ setIsAddingColor ]
-	);
 	const colorIndicatorStyle = useMemo(
 		() =>
 			fillComputedColors(
 				contentRef.current,
 				getActiveColors( value, name, colors )
 			),
-		[ value, colors ]
+		[ contentRef, value, colors ]
 	);
 
 	const hasColorsToChoose = colors.length || ! allowCustomControl;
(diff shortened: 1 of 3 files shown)


packages/format-library/src/text-color/inline.js
```diff
@@ -145,7 +137,11 @@ function ColorPicker( { name, property, value, onChange } ) {
 	return (
 		<ColorPalette
 			value={ activeColors[ property ] }
-			onChange={ onColorChange }
+			onChange={ ( color ) => {
+				onChange(
+					setColors( value, name, colors, { [ property ]: color } )
+				);
+			} }
```
Mamaduka, line 144: I checked other usages of the `ColorPalette` component; memoization isn't required.

ellatrix (approved):
Nice clean up

packages/format-library/src/text-color/index.js
```diff
@@ -66,21 +66,13 @@ function TextColorEdit( {
...
-		[ setIsAddingColor ]
-	);
 	const colorIndicatorStyle = useMemo(
 		() =>
 			fillComputedColors(
 				contentRef.current,
 				getActiveColors( value, name, colors )
 			),
-		[ value, colors ]
+		[ contentRef, value, colors ]
```
Mamaduka, line 75: Note: The `contentRef` is a ref object, but the ESLint plugin can't know this because it's passed as props. tl;dr; It's a false positive that we must fix, but it won't regress memo behavior.

packages/format-library/src/text-color/inline.js
dcalhoun: For clarity and my own edification, would you please elaborate on what you mean by "required?" 

My interpretation is that you mean that memoizing this callback is unnecessary as `ColorPalette` itself it not memoized. Is that accurate?

Thanks! 🙇🏻

packages/format-library/src/text-color/index.native.js
dcalhoun, line 1: I quickly tested the these changes on an iPhone simulator and Android emulator. I did not note any issues. Mentioning @geriux for awareness, just in case you have any concerns regarding these changes. 🙇🏻

packages/format-library/src/text-color/inline.js
Mamaduka: >My interpretation is that you mean that memoizing this callback is unnecessary as ColorPalette itself it not memoized. Is that accurate?

Correct and the `onChange` prop isn't a dependency for any side-effect ( `useEffect` ).

Issue #30029, opened 2021-03-19
FSE: Most site building blocks need previews and placeholders


jameskoster:
- [x] Site Logo - #35588
- [x] Site Tagline
- [x] Site Title - #35590
- [x] Login/out - https://github.com/WordPress/gutenberg/pull/62937
- [x] Post Title - https://github.com/WordPress/gutenberg/pull/62955
- [x] Post Content - https://github.com/WordPress/gutenberg/pull/62968
- [x] Post Author - https://github.com/WordPress/gutenberg/pull/62978
- [ ] Post Comments Count 
- [x] Post Date - https://github.com/WordPress/gutenberg/pull/63368
- [x] Post Excerpt - https://github.com/WordPress/gutenberg/pull/63010
- [x] Post Featured Image - https://github.com/WordPress/gutenberg/pull/63011
- [x] Post Tags - https://github.com/WordPress/gutenberg/pull/63369
- [ ] Term Description 
- [x] Posts List - https://github.com/WordPress/gutenberg/pull/63286
- [ ] Archive Title 
- [ ] Post Categories 
- [ ] Next post 
- [ ] Previous post 

[Here's a figma file](https://www.figma.com/file/xMD1TfdiAl6nmS9opIlUya/Site-building-block-placeholders-previews?node-id=1%3A18) where we can work on this.

jameskoster:
Some initial ideas for simplest text-based blocks:

| Block | Placeholder |
| --- | --- |
| Site Tagline | The site description  |
| Site Title | The site title  |
| Login/out | Log in |
| Post Title | The post title |
| Post Content | The content of a post, page, or any other content-type. It may include a variety of different blocks like paragraphs, lists, image galleries, and more. |
| Post Author | The post author  |
| ~Post Comment~ |  ~The content of a comment.~ |
| Post Comments Count | 1 |
| Post Date | [Gutenberg release date?] |
| Post Excerpt | The excerpt of a post, page, or any other content-type. |
| Post Tags | <a href="#">Tag</a> |
| Term Description | The description of a tag, category, or other custom taxonomy. |
| Archive Title | The title of an archive like a category, tag, author, or date |
| Post Categories | <a href="#">Category</a> |
| Next post | <a href="#">Next post</a> |
| Previous post | <a href="#">Previous post</a> |

carolinan:
I am wondering about this `Login/out | Log in` because in the editor we are always logged in, should it not be "Log out"?

[...]

PR #62968, opened 2024-06-28
Post content block: Add example of the block

MaggieCabrera:
## What?

Add a basic example to the `core/post-content` block.

## Why?

## How?

## Testing Instructions

Open the block inserter and search for the content block, hover over it and you'll see the example. This block is only available on the site editor.

## Screenshots or screencast 
Screenshot 2024-06-28 at 12 10 59


jameskoster (approved):
Working for me :)

PR #42000, opened 2022-06-28
UnitControl: Update unit dropdown design

mirka:
Part of #41973

## What?

Updates the unit dropdown to the new designs for the 40px size variant.

## Why?

To get our components closer to the mockups.

## How?

Add conditional styles for the dropdown depending on the size variant.

## Testing Instructions

1. `npm run storybook:dev`
2. See the stories for UnitControl.

### Known issues

- #42650

## Screenshots or screencast 

https://user-images.githubusercontent.com/555336/181043192-35466198-5e32-47b7-970a-ab559d47a1f6.mp4

### When the unit isn't a dropdown (i.e. single unit case)

Dropdown text has no accent color when the dropdown is inactive

PR #62970, opened 2024-06-28
Fix UnitControl select disabled state colors

### LINKED ISSUE #62969: UnitControl component keeps accent color when is disabled
### Description

UnitControl component keeps accent color when is disabled

1. Add a UnitControl component to the editor and pass `disabled` prop as true.
2. Look at the rendered styles (the unit select component still renders the accent color instead of the greyed disabled one).

[Screencast from 28-06-24 14:41:30.webm](https://github.com/WordPress/gutenberg/assets/1310626/27eaba66-3a5e-4566-b102-70c333760c0f)

### Environment info

- WordPress 6.5.4
- Gutenberg trunk

Yes

Yes

matiasbenedetto:
## What?
Fix UnitControl select disabled state colors

## Why?
Fixes: https://github.com/WordPress/gutenberg/issues/62969

## How?
Render the accent colors when the component is not disabled.

## Testing Instructions
See https://github.com/WordPress/gutenberg/issues/62969

### Before
[Screencast from 28-06-24 14:41:30.webm](https://github.com/WordPress/gutenberg/assets/1310626/52bae4ca-4e9a-450d-816f-0b5f8f141ac6)

### After
[Screencast from 28-06-24 14:38:27.webm](https://github.com/WordPress/gutenberg/assets/1310626/0568089b-2910-4ae0-bc4f-dac28ef3b692)

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-- `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
+### Bug Fixes
+
+-   `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
 
 ### Internal
 
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
+
 ### Internal
 
 -   `CustomSelectControlV2`: add root element wrapper. ([#62803](https://github.com/WordPress/gutenberg/pull/62803))
--- a/packages/components/src/unit-control/styles/unit-control-styles.ts
+++ b/packages/components/src/unit-control/styles/unit-control-styles.ts
@@ -114,7 +114,6 @@ const unitSelectSizes = ( { selectSize = 'default' }: SelectProps ) => {
 			align-items: center;
 
 			&:where( :not( :disabled ) ):hover {
-				color: ${ COLORS.ui.borderFocus };
 				box-shadow: inset 0 0 0
 					${ CONFIG.borderWidth + ' ' + COLORS.ui.borderFocus };
 				outline: ${ CONFIG.borderWidth } solid transparent; // For High Contrast Mode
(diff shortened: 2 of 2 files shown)


packages/components/src/unit-control/styles/unit-control-styles.ts
```diff
@@ -106,7 +113,7 @@ const unitSelectSizes = ( { selectSize = 'default' }: SelectProps ) => {
 			justify-content: center;
 			align-items: center;
 
-			&:hover {
+			&:not( :disabled ):hover {
 				color: ${ COLORS.ui.borderFocus };
```
mirka: Looks like this line is redundant? Because we already get the `color` from `baseUnitLabelStyles`.

packages/components/src/unit-control/styles/unit-control-styles.ts
```diff
@@ -106,7 +113,7 @@ const unitSelectSizes = ( { selectSize = 'default' }: SelectProps ) => {
 			justify-content: center;
 			align-items: center;
 
-			&:hover {
+			&:not( :disabled ):hover {
```
mirka:
```suggestion
			&:where( :not( :disabled ) ):hover {
```

I think we need to reduce the specificity here, to prevent the hover style from overriding the focus style when the select is already focused when hovered. (By the way I confirmed that the difference between the hover and focus style is [intentional](https://github.com/WordPress/gutenberg/pull/42000#issuecomment-1168598352).)

packages/components/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
```
mirka: Can we put this under a "Bug Fixes" heading please?
```suggestion
### Bug Fixes

-   `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
```

mirka (approved):
Great catch, thank you!

packages/components/src/unit-control/styles/unit-control-styles.ts
matiasbenedetto: thanks for the review. I didn't get what you mean in this comment.

mirka: It looks like `color: ${ COLORS.ui.borderFocus }` is unnecessary, because the correct color is already set here:

https://github.com/WordPress/gutenberg/blob/f9d406c5e98b4b4164418a4d19f6905a821c658e/packages/components/src/unit-control/styles/unit-control-styles.ts#L68-L70

It should be ok to remove, unless I'm missing something.

matiasbenedetto: Ohh I see. Yes, removed: [`9cd4cec` (#62970)](https://github.com/WordPress/gutenberg/pull/62970/commits/9cd4cecc7df454f18e01b33686d06c67dfff8c13)

Issue #62885, opened 2024-06-26
Set Default CSS Transitions via Theme.json


bhubbard:
## What problem does this address?
CSS Transitions can be common on sites, for hover effects on buttons, links, etc. Having basic support for this in theme.json can lead to developers using them for blocks and patterns, while helping keeping them consistent for the site.

## What is your proposed solution?
I would like to see some common default css transitions set via theme.json, with the ability to set custom, just like fonts and colors.

Example:
```
// Transition Durations
$transition-duration-fast: 0.2s;
$transition-duration-medium: 0.4s;
$transition-duration-slow: 0.6s;

// Transition Timings
$transition-timing-ease-in: ease-in;
$transition-timing-ease-out: ease-out;
$transition-timing-ease-in-out: ease-in-out;
$transition-timing-linear: linear;

// Transition Delays
$transition-delay-none: 0s;
$transition-delay-short: 0.1s;
$transition-delay-medium: 0.2s;
$transition-delay-long: 0.3s;
```

t-hamano:
How about defining CSS custom properties via `settings.custom`?

```json
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 3,
	"settings": {
		"custom": {
			"transition": {
				"duration": {
					"fast": "0.2s",
					"medium": "0.4s",
					"slow": "0.6s"
				},
				"timing": {
					"ease-in": "ease-in",
					"ease-out": "ease-out",
					"ease-in-out": "ease-in-out",
					"linear": "linear"
				},
				"delay": {
					"none": "0s",
					"short": "0.1s",
					"medium": "0.2s",
					"long": "0.3s"
				}
			}
		}
	}
}
```

These definitions will be output in both the editor and the front-end as follows.

```css
:root{
    --wp--custom--transition--duration--fast: 0.2s;
    --wp--custom--transition--duration--medium: 0.4s;
    --wp--custom--transition--duration--slow: 0.6s;
    --wp--custom--transition--timing--ease-in: ease-in;
    --wp--custom--transition--timing--ease-out: ease-out;
    --wp--custom--transition--timing--ease-in-out: ease-in-out;
    --wp--custom--transition--timing--linear: linear;
    --wp--custom--transition--delay--none: 0s;
    --wp--custom--transition--delay--short: 0.1s;
    --wp--custom--transition--delay--medium: 0.2s;
    --wp--custom--transition--delay--long: 0.3s;
}
```

They can be used in patterns and blocks.

PR #62972, opened 2024-06-28
Add Custom Transition Settings to theme.json

geekofshire:
### Summary
This PR adds custom settings for CSS transitions in the `theme.json` file. The new settings include transition durations, timing functions, and delays, each with detailed descriptions, types, and default values.

### Changes
- Added `transition` object under `settings.custom` in `theme.json`.
- Defined properties for `duration`, `timing`, and `delay` with metadata.
- Enhanced documentation within `theme.json` for better understanding and maintenance.

### Impact
These changes allow themes to define default CSS transitions, providing a consistent and customizable user experience across blocks and patterns.

Related Issue - #62885


t-hamano (commented):
Thanks for the PR!

You probably submitted this PR after seeing [my comment](https://github.com/WordPress/gutenberg/issues/62885#issuecomment-2192777270).

However, the code I proposed is not intended to be incorporated as the default into WordPress core/Gutenberg, but merely defined an example of adding custom properties as needed to each theme.

Since the CSS properties we want to add vary depending on the theme, it would be difficult to incorporate hard-coded values ​​as defaults.

Therefore, I will close this PR, but please feel free to comment if there is anything I have missed.

Issue #62315, opened 2024-06-05
Backports for 6.6 Beta/RC


ellatrix:
This issue supersedes #61226. See [Back-merging code to WordPress Core](https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/back-merging-to-wp-core.md).

I'm starting a new issue because I'd like to try a new approach based on the backport log. Note that, since we started the backport log before Beta, this will also include backports from before Beta.

PRs have title/status previews, so it's easy to have an overview of overall status (no need for a checklist).

- https://github.com/WordPress/wordpress-develop/pull/6279
  * https://github.com/WordPress/gutenberg/pull/60063
- https://github.com/WordPress/wordpress-develop/pull/6522
  * https://github.com/WordPress/gutenberg/pull/60106
  * https://github.com/WordPress/gutenberg/pull/60228
  * https://github.com/WordPress/gutenberg/pull/61638
- https://github.com/WordPress/wordpress-develop/pull/6559
  * https://github.com/WordPress/gutenberg/pull/60349
  * https://github.com/WordPress/gutenberg/pull/60464
  * https://github.com/WordPress/gutenberg/pull/60491
  * https://github.com/WordPress/gutenberg/pull/61757
- https://github.com/WordPress/wordpress-develop/pull/6567
  * https://github.com/WordPress/gutenberg/pull/47271
  * https://github.com/WordPress/gutenberg/pull/61774
- https://github.com/WordPress/wordpress-develop/pull/6590
  * https://github.com/WordPress/gutenberg/pull/59531
  * https://github.com/WordPress/gutenberg/pull/61182
  * https://github.com/WordPress/gutenberg/pull/61717
- https://github.com/WordPress/wordpress-develop/pull/6605
  * https://github.com/WordPress/gutenberg/pull/61851
- https://github.com/WordPress/wordpress-develop/pull/6616
  * https://github.com/WordPress/gutenberg/pull/58409
  * https://github.com/WordPress/gutenberg/pull/61328
  * https://github.com/WordPress/gutenberg/pull/61842
  * https://github.com/WordPress/gutenberg/pull/62199
  * https://github.com/WordPress/gutenberg/pull/62252
- https://github.com/WordPress/wordpress-develop/pull/6656
  * https://github.com/WordPress/gutenberg/pull/60715
- https://github.com/WordPress/wordpress-develop/pull/6662
  * https://github.com/WordPress/gutenberg/pull/57908
[...]

PR #62973, opened 2024-06-28
Automatically sync backport changelog to issue

ellatrix:
## What?

Creates a new workflow to automatically sync the changelog to the right issue.

## Why?

Reduce repetitive manual labor.

## How?

When a change is pushed to trunk the workflow will:

* Find the last created issue with the label "🤖 Sync Backport Changelog".
* Derive the version number from the title.
* Generate the backport list from that.
* Check the existing issue content for delimiters:
  * If there are none, append the list with delimiters at the end.
  * If there are, replace the contents. Only update the issue if the contents are different.

When a new release cycle is started, someone can just create a new issue with the label "🤖 Sync Backport Changelog" and mention 6.7 (for example) in the title. On the next push to trunk the issue will be updated.

The script will only sync the last created issue, previous ones will be ignored and should probably be closed.

### Future work

We could add a PR workflow that also adds unmerged backport PRs to the issue in the future. Worth noting that even the old cherry-pick script never took any unmerged PRs into account.

## Testing Instructions

You can see it in action here: https://github.com/WordPress/gutenberg/issues/62315


.github/workflows/check-backport-changelog.yml
```diff
@@ -18,22 +18,15 @@ on:
...
         steps:
-            - name: 'Get PR commit count'
-              run: echo "PR_COMMIT_COUNT=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> "${GITHUB_ENV}"
-            - name: Checkout code
-              uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+            - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
               with:
                   ref: ${{ github.event.pull_request.head.ref }}
                   repository: ${{ github.event.pull_request.head.repo.full_name }}
-                  fetch-depth: ${{ env.PR_COMMIT_COUNT }}
```
ellatrix, line 31: We don't need commit history for this test, so cleaning up a bit.

.github/workflows/sync-backport-changelog.yml
```diff
@@ -0,0 +1,70 @@
...
+                      if (!versionMatch) {
+                        console.log('Could not find a version number in the latest issue title.');
+                        return;
+                      }
+
+                      const version = versionMatch[1];
+                      console.log(`Latest version: ${version}`);
+
+                      const { execSync } = require('child_process');
+                      const processedChangelog = execSync(`awk '/./ {print ($0 ~ /^[-*]/ ? "  " : "- ") $0}' backport-changelog/${version}/*.md`).toString().trim();
```
vcanales, line 49: Could we check the last commit for changes in this directory before doing anything else? Seems like it might save us some API requests.
Another alternative could be conditioning this workflow on a check similar to the one `check-backport-changelog` uses.

ellatrix: What API requests would it save? I'm already checking if the content is different.

vcanales: The very first one 

https://github.com/WordPress/gutenberg/pull/62973/files#diff-b61aaec5beeb30ae49637bf34830218b191db2c3fb71db321153b7cf6a93226dR18-R25

What I'm getting at is that I think that not starting the workflow if it's not necessary is a good idea

ellatrix: Added an early abort 8bb22a2b43fc439fcdb67b2151ffb80fb4a3ba02

vcanales (approved):
Looks good!

ellatrix:
😔 I forgot to add props, I'm so sorry

ellatrix:
@vcanales What I don't like now, after the last change, is that it will only add the changes to a newly created issue after another backport was made.

PR #62978, opened 2024-06-28
Post Author blocks: Add example and preview

MaggieCabrera:
## What?

Add a basic example to the `core/post-author`, `core/post-author-biography` and `core/post-author-name` blocks.

## Why?

## How?

I define a viewport size for the example and add some text for the byline. Please let me know if we should change the text here

Open the block inserter and search for the author block, hover over all 3 of the blocks and you'll see the examples

## Screenshots or screencast 

Screen Capture on 2024-06-28 at 16-00-00

### DIFF
--- a/packages/block-library/src/post-author/block.json
+++ b/packages/block-library/src/post-author/block.json
@@ -34,13 +34,6 @@
 		}
 	},
 	"usesContext": [ "postType", "postId", "queryId" ],
-	"example": {
-		"viewportWidth": 350,
-		"attributes": {
-			"showBio": true,
-			"byline": "Posted by"
-		}
-	},
 	"supports": {
 		"html": false,
 		"spacing": {
--- a/packages/block-library/src/post-author/block.json
+++ b/packages/block-library/src/post-author/block.json
@@ -34,6 +34,13 @@
 		}
 	},
 	"usesContext": [ "postType", "postId", "queryId" ],
+	"example": {
+		"viewportWidth": 350,
+		"attributes": {
+			"showBio": true,
+			"byline": "Posted by"
+		}
+	},
 	"supports": {
 		"html": false,
 		"spacing": {
--- a/packages/block-library/src/post-author-biography/block.json
+++ b/packages/block-library/src/post-author-biography/block.json
@@ -12,6 +12,9 @@
 		}
 	},
 	"usesContext": [ "postType", "postId" ],
+	"example": {
+		"viewportWidth": 350
+	},
 	"supports": {
 		"spacing": {
 			"margin": true,
--- a/packages/block-library/src/post-author-name/block.json
+++ b/packages/block-library/src/post-author-name/block.json
@@ -20,6 +20,9 @@
 		}
 	},
 	"usesContext": [ "postType", "postId" ],
+	"example": {
+		"viewportWidth": 350
+	},
 	"supports": {
 		"html": false,
 		"spacing": {
--- a/packages/block-library/src/post-author/index.js
+++ b/packages/block-library/src/post-author/index.js
@@ -1,6 +1,7 @@
 /**
  * WordPress dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { postAuthor as icon } from '@wordpress/icons';
 
 /**
@@ -15,6 +16,13 @@ export { metadata, name };
 
 export const settings = {
 	icon,
+	example: {
+		viewportWidth: 350,
+		attributes: {
+			showBio: true,
+			byline: __( 'Posted by' ),
+		},
+	},
 	edit,
 };
 


artemiomorales (approved):
Tested this and it looks good to me.

# Before

Previously no preview was available:

https://github.com/WordPress/gutenberg/assets/5360536/b325242b-0cd2-4649-a900-617e3e9363ba

# After

When the author details _are not_ available in the context, it shows fallback values:

https://github.com/WordPress/gutenberg/assets/5360536/23437ee6-828b-457e-85db-75ecb5e9cd9f

If the details _are_ available, it shows them accordingly:

https://github.com/WordPress/gutenberg/assets/5360536/ccf71b4b-5643-4d3a-9d4b-7069d2d98abf

carolinan:
In the Site Editor, where the real data can not be shown, it is perhaps a bit small still. I personally struggle to see the text "Author Name" when Twenty Twenty-Four is active, because its blurry. The theme sets the default text size to small.

site-editor-styles

The examples are strangely positioned differently in different previews, but that is unrelated to this PR.
They also have different background colors:
Comparison between the inserter preview and styles sidebar preview

MaggieCabrera:
> In the Site Editor, where the real data can not be shown, it is perhaps a bit small still. I personally struggle to see the text "Author Name" when Twenty Twenty-Four is active, because its blurry. The theme sets the default text size to small.
> 
> site-editor-styles
> 
> The examples are strangely positioned differently in different previews, but that is unrelated to this PR. They also have different background colors: Comparison between the inserter preview and styles sidebar preview

@jameskoster do you think in this case we should force a different font size? It's a common block to have a smaller one

jameskoster:
That's a tricky one. It's not a strong feeling, but I'd be inclined not to add custom sizing on a per-block basis. 

It feels reasonably important to me that previews are accurate relative to one another. IE the text in a preview for a block with a `small` font size setting should look smaller than the text in the preview of a block with `medium` font size. Otherwise it can lead to confusion.

I'd welcome feedback from others in @WordPress/gutenberg-design.

MaggieCabrera:
> That's a tricky one. It's not a strong feeling, but I'd be inclined not to add custom sizing on a per-block basis.
> 
> It feels reasonably important to me that previews are accurate relative to one another. IE the text in a preview for a block with a `small` font size setting should look smaller than the text in the preview of a block with `medium` font size. Otherwise it can lead to confusion.
> 
> I'd welcome feedback from others in @WordPress/gutenberg-design.

By selecting a specific viewport size we are already skewing the relative size of the content. I've maintained the same size for the PRs I've opened but I've seen cases that change it (like the site title). I'm just mentioning this here so we try and keep it consistent.

packages/block-library/src/post-author/block.json
```diff
@@ -34,6 +34,13 @@
 		}
 	},
 	"usesContext": [ "postType", "postId", "queryId" ],
+	"example": {
+		"viewportWidth": 350,
+		"attributes": {
+			"showBio": true,
+			"byline": "Posted by"
```
t-hamano: This sentence probably won't be localized, so I think it would be better to register the example via JavaScript, [like the  Paragraph block](https://github.com/WordPress/gutenberg/blob/872f8b37ddb51b3b7f2d16da1138f97d0887df3d/packages/block-library/src/paragraph/index.js#L23-L29).

MaggieCabrera: good call!

t-hamano (approved):
Sorry for the late re-review.

I think it's OK to ship this PR.

I think the problem here is that even though the block preview area in the block inserter and the block preview area in the global styles are almost the same size, the block position and scale are different.

The block preview in the block inserter is generated here:

https://github.com/WordPress/gutenberg/blob/3c8bb803ab8b06208dd74c5bb78d4c397fa13d39/packages/block-editor/src/components/inserter/preview-panel.js#L39-L45

On the other hand, the block preview in the global styles is generated here:

https://github.com/WordPress/gutenberg/blob/3c8bb803ab8b06208dd74c5bb78d4c397fa13d39/packages/edit-site/src/components/global-styles/block-preview-panel.js#L48-L62

As you can see, the `minHeight` prop and the `additionalStyles` prop are different.

These props shouldn't affect the scale or position of the block itself, but somehow these props seem to affect the scale and position.

Therefore, I think we need to investigate the specs of the `BlockPreview` component itself and fix the differences between the two views.

MaggieCabrera:
> Sorry for the late re-review.
> 
> I think it's OK to ship this PR.
> 
> I think the problem here is that even though the block preview area in the block inserter and the block preview area in the global styles are almost the same size, the block position and scale are different.
> 
> The block preview in the block inserter is generated here:
> 
> 
> On the other hand, the block preview in the global styles is generated here:
> 
> 
> As you can see, the `minHeight` prop and the `additionalStyles` prop are different.
> 
> These props shouldn't affect the scale or position of the block itself, but somehow these props seem to affect the scale and position.
> 
> Therefore, I think we need to investigate the specs of the `BlockPreview` component itself and fix the differences between the two views.

Thank you for this! we are spread out over multiple PRs on this topic so I'm sorry that you might have missed my efforts to unify this [here](https://github.com/WordPress/gutenberg/pull/63177). I'd be glad for you to have a look at it if you have the chance!

Issue #62938, opened 2024-06-27
Block Bindings: Notify users of granular post meta edits in save panel


artemiomorales:
## What problem does this address?

When saving a post with a modified block binding, the save panel currently flags that the Post Meta has changed with the following:

![Screenshot 2024-06-27 at 13 30 44](https://github.com/WordPress/gutenberg/assets/34552881/b00962d4-c0d6-47f2-95af-737284213368)

However, this doesn't surface the individual changes that have occurred to the post meta, nor does it allow users to opt out of those edits in a piecemeal way.

## What is your proposed solution?

We should instead revise the save panel so that it shows the following experience:

<img width="1638" alt="332440083-302bf4e4-6133-4c60-9d75-3c037402f5af" src="https://github.com/WordPress/gutenberg/assets/5360536/9024f6a4-713f-4263-9e1f-55f93aa0f7c2">

We can start with just showing what changes have been made, then later add the checkboxes.

Issue #62329, opened 2024-06-05
Add indicator for modified post meta in pre-publish flow


artemiomorales:
## What problem does this address?

Follow up to https://github.com/WordPress/gutenberg/pull/61811

While we are now indicating to users when metadata has changed to published posts, we do not yet do so when a post is still in pre-published state.

We should determine what this UX should be and any revisions that would be needed to accommodate this in the pre-publish flow.

## What is your proposed solution?

One idea: We can consider integrating information about the post meta in the pre-publish check as another collapsed panel next to suggestions related to tags and categories, as observed in [this comment](https://github.com/WordPress/gutenberg/pull/61811#issuecomment-2137438378).

<img width="1289" alt="334820578-ad6c39a2-e222-4540-8feb-fb8d5cba5311" src="https://github.com/WordPress/gutenberg/assets/5360536/cb8c1c3e-d76f-4b12-add2-f322dc0a051e">

We should also consider surfacing this information whenever a draft is saved.

artemiomorales:
If we were to show the meta changes in the pre-publish flow as a panel, what would we put in it? This is what two of the other panels look like expanded:

<img width="1173" alt="Screenshot 2024-07-10 at 5 24 45 PM" src="https://github.com/WordPress/gutenberg/assets/5360536/5ca85c27-f3b4-4380-9b48-b93f8d8345f3">

We could just include the modified meta shown in the following Posts panel from https://github.com/WordPress/gutenberg/pull/62982:

<img width="279" alt="344239912-c89418c1-285b-4521-93fe-1135ec912ea6" src="https://github.com/WordPress/gutenberg/assets/5360536/76e3419c-1d51-4025-b421-28f63c28d537">

I don't believe the "Posts" heading would work in this case. Perhaps it should say "Meta"? Other aspects of the panel would also need to change, including:
- Removing "There is X site change(s) waiting to be saved"
- Removing "Post Meta"
- Removing the post name, since we'll be in the post editor and displaying the name would be redundant

Maybe it could look something like this:

<img width="278" alt="post-meta-design2" src="https://github.com/WordPress/gutenberg/assets/5360536/e88dd4ba-e673-4cbf-8cf8-216ccab27799">

jasmussen:
[...]

PR #62982, opened 2024-06-28
Block Bindings: Output granular post meta changes in a list inside save panel

artemiomorales:
## What?

This PR granularly outputs post metadata changes in the save panel.

## Why?

Begins addressing https://github.com/WordPress/gutenberg/issues/62938
We want to give users better visibility on what is being saved when post metadata is modified.

## How?

It renames the existing `hasPostMetaChanges` to `getPostMetaChangs` and compares the original post object's `meta` property to the modified one.

## Testing Instructions

<details>
<summary>1. Register some post meta fields by adding this snippet to your theme's functions.php</summary>

```php

	register_meta(
		'post',
		'text_field_1',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'sanitize_callback' => 'wp_strip_all_tags',
			'default'           => 'bindings value 1',
		)
	);

	register_meta(	
		'post',
		'text_field_2',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'sanitize_callback' => 'wp_strip_all_tags',
			'default'           => 'bindings value 2',
		)
	);

	register_meta(
		'post',
		'text_field_3',
		array(
			'show_in_rest'      => true,
			'single'            => true,
			'sanitize_callback' => 'wp_strip_all_tags',
			'default'           => 'bindings value 3',
		)
	);
}
```

</details>

<details>

<summary>2. In published post, add paragraph blocks bound to the custom fields using the Code Editor</summary>

```
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_field_1"}}}}} -->
<p>Bound Paragraph 1 more editing test</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_field_2"}}}}} -->
<p>Bound Paragraph 2 test</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"core/post-meta","args":{"key":"text_field_3"}}}}} -->
<p>Bound paragraph 2</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Override the paragraph contents by typing into the bound paragraphs.

4. Press Save and see that an indicator for the meta changes is shown in the save panel.

Repeat the same for the heading, image, and button blocks.

## Screenshots or screencast 
Screenshot 2024-06-28 at 6 02 53 PM

### DIFF
--- a/packages/editor/src/components/post-publish-button/index.js
+++ b/packages/editor/src/components/post-publish-button/index.js
@@ -61,8 +61,7 @@ export class PostPublishButton extends Component {
 			// TODO: Explore how to manage `getPostMetaChanges` and pre-publish workflow properly.
 			if (
 				( hasNonPostEntityChanges ||
-					( Object.keys( postMetaChanges ).length > 0 &&
-						isPublished ) ) &&
+					( postMetaChanges.length > 0 && isPublished ) ) &&
 				setEntitiesSavedStatesCallback
 			) {
 				// The modal for multiple entity saving will open,
--- a/packages/editor/src/components/post-publish-button/index.js
+++ b/packages/editor/src/components/post-publish-button/index.js
@@ -48,7 +48,7 @@ export class PostPublishButton extends Component {
 		return ( ...args ) => {
 			const {
 				hasNonPostEntityChanges,
-				hasPostMetaChanges,
+				postMetaChanges,
 				setEntitiesSavedStatesCallback,
 				isPublished,
 			} = this.props;
@@ -58,10 +58,11 @@ export class PostPublishButton extends Component {
 			// We also need to check that the `setEntitiesSavedStatesCallback`
 			// prop was passed. See https://github.com/WordPress/gutenberg/pull/37383
 			//
-			// TODO: Explore how to manage `hasPostMetaChanges` and pre-publish workflow properly.
+			// TODO: Explore how to manage `getPostMetaChanges` and pre-publish workflow properly.
 			if (
 				( hasNonPostEntityChanges ||
-					( hasPostMetaChanges && isPublished ) ) &&
+					( Object.keys( postMetaChanges ).length > 0 &&
+						isPublished ) ) &&
 				setEntitiesSavedStatesCallback
 			) {
 				// The modal for multiple entity saving will open,
@@ -226,7 +227,7 @@ export default compose( [
 			isSavingNonPostEntityChanges,
 			getEditedPostAttribute,
 			getPostEdits,
-			hasPostMetaChanges,
+			getPostMetaChanges,
 		} = unlock( select( editorStore ) );
 		return {
 			isSaving: isSavingPost(),
@@ -244,7 +245,7 @@ export default compose( [
 			postStatus: getEditedPostAttribute( 'status' ),
 			postStatusHasChanged: getPostEdits()?.status,
 			hasNonPostEntityChanges: hasNonPostEntityChanges(),
-			hasPostMetaChanges: hasPostMetaChanges(),
+			postMetaChanges: getPostMetaChanges(),
 			isSavingNonPostEntityChanges: isSavingNonPostEntityChanges(),
 		};
 	} ),
--- a/packages/editor/src/components/post-publish-button/index.js
+++ b/packages/editor/src/components/post-publish-button/index.js
@@ -61,7 +61,9 @@ export class PostPublishButton extends Component {
 			// TODO: Explore how to manage `getPostMetaChanges` and pre-publish workflow properly.
 			if (
 				( hasNonPostEntityChanges ||
-					( postMetaChanges.length > 0 && isPublished ) ) &&
+					( postMetaChanges &&
+						postMetaChanges.length > 0 &&
+						isPublished ) ) &&
 				setEntitiesSavedStatesCallback
 			) {
 				// The modal for multiple entity saving will open,
--- a/packages/editor/src/store/private-selectors.js
+++ b/packages/editor/src/store/private-selectors.js
@@ -23,6 +23,7 @@ import {
 import { TEMPLATE_PART_POST_TYPE } from './constants';
 import { getFilteredTemplatePartBlocks } from './utils/get-filtered-template-parts';
 import { getEntityActions as _getEntityActions } from '../dataviews/store/private-selectors';
+import { deepCompare } from '../../../block-editor/src/components/global-styles/get-global-styles-changes';
 
 const EMPTY_INSERTION_POINT = {
 	rootClientId: undefined,
@@ -165,21 +166,10 @@ export const getPostMetaChanges = createRegistrySelector(
 		);
 
 		if ( ! original?.meta || ! edits?.meta ) {
-			return {};
+			return [];
 		}
 
-		const keys = Object.keys( original.meta );
-		const differences = {};
-		for ( const key of keys ) {
-			if ( key === 'footnotes' ) {
-				continue;
-			}
-			if ( original.meta[ key ] !== edits.meta[ key ] ) {
-				differences[ key ] = edits.meta[ key ];
-			}
-		}
-
-		return differences;
+		return deepCompare( edits.meta, original.meta );
 	}
 );
 
--- a/packages/editor/src/store/private-selectors.js
+++ b/packages/editor/src/store/private-selectors.js
@@ -1,9 +1,4 @@
 /**
- * External dependencies
- */
-import fastDeepEqual from 'fast-deep-equal';
-
-/**
  * WordPress dependencies
  */
 import { store as blockEditorStore } from '@wordpress/block-editor';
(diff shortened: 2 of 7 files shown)


artemiomorales:
This is still a work in progress, though I could use an opinion on the following:

- Should we do a deep comparison of the original post metadata object and the edits object instead of a shallow one, and if so, is there an existing function in the codebase we can use?
- ~~The comparison assumes that the original object and the edits one will have the same set of keys — not sure if that's a safe assumption to make (I imagine not? Though if some keys exist in the edits object and not the original, something must be off)~~ (no longer applicable)

I'll continue exploring this other point:
- Is a simple comparison between the objects good enough to handle the checkboxes proposed in https://github.com/WordPress/gutenberg/issues/62938, and for these other issues as well?
  - [Editing bound paragraph in post editor causes "Site Updated" message to appear](https://github.com/WordPress/gutenberg/issues/62236)
  - [Add indicator for modified post meta in pre-publish flow](https://github.com/WordPress/gutenberg/issues/62329).

cbravobernal:
>     is there an existing function in the codebase we can use?
> 

This may work: https://github.com/WordPress/gutenberg/blob/15da2aae42d059e220face536bec5bfe4c3735e6/packages/block-editor/src/components/global-styles/get-global-styles-changes.js#L71

For the rest of the questions, I still need to re-read everything 😅

artemiomorales:
>
> This may work:
> https://github.com/WordPress/gutenberg/blob/15da2aae42d059e220face536bec5bfe4c3735e6/packages/block-editor/src/components/global-styles/get-global-styles-changes.js#L71
>
>For the rest of the questions, I still need to re-read everything 😅

@cbravobernal Great, thanks! I switched to use the deep comparison, which seems like the right move and resolves my questions regarding the objects. Still need to explore if this approach would work for other the other mentioned issues.

SantosGuillamot:
> Should we do a deep comparison of the original post metadata object and the edits object instead of a shallow one, and if so, is there an existing function in the codebase we can use?

I am not sure we should export a function that lives inside a component that is specific to global styles, which seems unrelated to this. The [description of that function](https://github.com/WordPress/gutenberg/blob/a9b7067515baace75a45e76f7a933ed57554b101/packages/block-editor/src/components/global-styles/get-global-styles-changes.js#L65) says _"A deep comparison of two objects, optimized for comparing global styles."_ I'm wondering if we should create our own function for compare or create a global util for this.

Doesn't `fastDeepEqual` do a deep comparison? The name suggests it does and it is widely used across Gutenberg packages.

EDIT: Oh, we want to compare and get the diff, not just check equal, right?

> The comparison assumes that the original object and the edits one will have the same set of keys — not sure if that's a safe assumption to make.

I assume it is possible to have an `undefined` value in the original object because the custom field hasn't been defined for that post, and populate it later after creating a binding. It'd be great to explore and support that use case if needed.

cbravobernal:
>  I'm wondering if we should create our own function for compare or create a global util for this.

Agree with that. It's weird we didn't need it until then. But usually there could be an utils file per package. If it will be widely used in the project, then a `fastDeepEqual` approach.

artemiomorales:
> Doesn't fastDeepEqual do a deep comparison? The name suggests it does and it is widely used across Gutenberg packages.
>
> EDIT: Oh, we want to compare and get the diff, not just check equal, right?

Right, the diff is why fastDeepEqual isn't good enough for this use case.

> I assume it is possible to have an undefined value in the original object because the custom field hasn't been defined for that post, and populate it later after creating a binding. It'd be great to explore and support that use case if needed.

Ok will look at this 👍

> Agree with that. It's weird we didn't need it until then. But usually there could be an utils file per package. If it will be widely used in the project, then a fastDeepEqual approach.

Ok, I can create a deep equal comparison for the `editor` package.

packages/editor/src/store/private-selectors.js
```diff
@@ -34,6 +29,49 @@ const EMPTY_INSERTION_POINT = {
...
+			changedObject[ key ],
+			originalObject[ key ],
+			path
+		);
+		if ( changedPath ) {
+			diffs = diffs.concat( changedPath );
+		}
+	}
+	return diffs;
+}
```
artemiomorales, line 74: How does this deep equality comparison look? The one mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/62982#issuecomment-2197288989) works well, so I essentially copied it, just changing the appearance of the diff slightly. I'm not sure in what other scenarios metadata might change or how, and if this comparison would suit those scenarios.

packages/is-shallow-equal/src/objects.js
```diff
@@ -31,7 +57,7 @@ export default function isShallowEqualObjects( a, b ) {
 			//
 			// Example: isShallowEqualObjects( { a: undefined }, { b: 5 } )
 			( aValue === undefined && ! b.hasOwnProperty( key ) ) ||
-			aValue !== b[ key ]
+			! isEqual( aValue, b[ key ] )
```
artemiomorales, line 60: In using `getPostMetaChanges()`, I was receiving a warning: `The 'use select' hook returns different values  when called with the same state and parameters. This can lead to unnecessary re-renders.`

Here's a screenshot of the values being compared along with the warning:

Screenshot 2024-07-04 at 5 46 45 AM

This check is called in the following lines:

https://github.com/WordPress/gutenberg/blob/7dda8fdc90d1e5b44211143e4ed963f3d0a9f9cd/packages/data/src/components/use-select/index.js#L161-L166

Essentially, a new array is returned each time from `getPostMetaChanges()`, causing the warning even though the array values are equal. What's the best way to handle this?

I'm not sure the equality check was written with this scenario in mind, so I revised it to remove the warning and aid with discussion. Would appreciate any pointers or thoughts 🙏

artemiomorales (commented):
I think this PR is ready for feedback. Would appreciate any thoughts especially on the below!

artemiomorales:
> - Is a simple comparison between the objects good enough to handle the checkboxes proposed in https://github.com/WordPress/gutenberg/issues/62938, and for these other issues as well?
https://github.com/WordPress/gutenberg/issues/62236
https://github.com/WordPress/gutenberg/issues/62329.

After taking a deeper look, I believe these issues are all best handled separately, as their needs and discussions, while related, will be different.

packages/editor/src/store/private-selectors.js
artemiomorales: Also, since we're only using the deep comparison in this file, I decided to just keep it local rather than creating a utils file for now. How does that sound?

packages/is-shallow-equal/src/objects.js
SantosGuillamot: I don't feel fully comfortable changing the conditional of the warning. I would like to understand better why the existing implementation is triggering that warning. I've been doing some checks with global styles, which is a really similar use case to what we are trying to achieve and this is what I found:

**In `trunk`**

In the Site Editor:
* when I load a specific post, it doesn't trigger any warning.
* when I modify a global style and click save, it doesn't trigger any warning.

**In this branch without the changes in `is-shallow-equal` package**
* when I load any post, even if it doesn't have any block connected, it triggers the warning.
* when I modify a global style and click save, it DOES trigger the warning.
* when I modify post meta and click save, it DOES trigger the warning.

That makes me think that something might be wrong with this implementation. That's why I would like to understand better the root cause. Why is it working in `trunk` for global styles, which is a really similar use case, and not in this case?

packages/editor/src/store/private-selectors.js
SantosGuillamot: The only thing that changes is the way we show nested properties, right? ([This line](https://github.com/WordPress/gutenberg/pull/62982/files#diff-a1cdb9f3f4ce98ee4ac08627daff3c4cb003766468d4bb3780fe0be22a0ad87eR48))

For me, both cases are really similar so maybe it makes sense to keep consistency between them?

Additionally, I think it's fine to keep it local for now, but I would open a follow-up pull request to move `deepCompare` to a util to foster the discussion. It seems that at least global styles and post meta could benefit from this. I assume it should be "format agnostic" and it can be formatted after the deepCompare return for each specific use case.

packages/editor/src/components/post-publish-button/index.js
```diff
@@ -58,10 +58,12 @@ export class PostPublishButton extends Component {
 			// We also need to check that the `setEntitiesSavedStatesCallback`
 			// prop was passed. See https://github.com/WordPress/gutenberg/pull/37383
 			//
-			// TODO: Explore how to manage `hasPostMetaChanges` and pre-publish workflow properly.
+			// TODO: Explore how to manage `getPostMetaChanges` and pre-publish workflow properly.
 			if (
 				( hasNonPostEntityChanges ||
-					( hasPostMetaChanges && isPublished ) ) &&
+					( postMetaChanges &&
+						postMetaChanges.length > 0 &&
```
SantosGuillamot, line 65: Is not enough to use `postMetaChanges?.length` here?

packages/editor/src/store/private-selectors.js
```diff
@@ -34,6 +29,49 @@ const EMPTY_INSERTION_POINT = {
...
+
+/**
+ * A deep comparison of two objects, optimized for comparing metadata changes.
+ * @param {Object} changedObject  The changed object to compare.
+ * @param {Object} originalObject The original object to compare against.
+ * @param {string} parentPath     A key/value pair object of block names and their rendered titles.
+ * @return {string[]}             An array of paths whose values have changed.
+ */
+function deepCompare( changedObject, originalObject, parentPath = '' ) {
+	// We have two non-object values to compare.
```
ntsekouras, line 42: This doesn't seem to be the place to have such a util. It has nothing to do with the store.

packages/is-shallow-equal/src/objects.js
ntsekouras: This shouldn't change and the warning tries to protect us from unnecessary rerenders. Probably what we need here is have a selector that gets the data(returns stable reference) and do the comparison in the consumer components.

We could keep the `hasPostMetaChanges` selector probably and only calculate the diff if it has changes.

artemiomorales:
I'm closing this for now; explorations can continue if or when discussion on the UX regarding post meta changes continues.

PR #39303, opened 2022-03-09
`UnitControl`: Revamp support for changing unit by typing

stokesman:
## What?
Revamps the type-to-change-units feature in `UnitControl` and removes it's dependency on the `isPressEnterToChange` prop being `true`. ~~Also removes `BoxControl`’s defaulting of `isPressEnterToChange` to `true` in order for it to send changes instantaneously (as opposed to now how it waits for either <kbd>ENTER</kbd> key press or a `blur` event)~~.

## Why?
~~Betters UX consistency in regard to inputs that produce changes instantly. Unlike all other text inputs in Gutenberg, `BoxControl`-based ones do not produce changes upon input. It seems the only reason for this difference is to enable the type-to-change-units feature which requires it.~~ With this PR, the feature works even when the mode of these controls set to instantaneous (`isPressEnterToChange={ false }`)

The feature itself may also be found to work better but testing will tell. It's new implementation is certainly less complex and better from a maintenance perspective.

Related:
* Issue with consistency noted https://github.com/WordPress/gutenberg/issues/30222#issue-840656311
* We've also recently merged #39109 that seems to have been motivated by the urge to alleviate this.

## How?
* Removal of the logic that currently supports the feature.
* Addition of simpler logic that merely moves focus to the unit select when a key that matches the first character of a unit. 
* ~~Change of the default `isPressEnterToChange` of `BoxControl` to `false`.~~

## Testing Instructions
1. Play with a `UnitControl` in Storybook or Gutenberg.
2. Verify that the typing a unit from the text input changes the units.

## Screenshots or screencast
https://user-images.githubusercontent.com/9000376/157374684-111cead1-3c3f-4021-9a60-7285d4a3c878.mp4

PR #62988, opened 2024-06-29
UnitControl: Fix an issue where keyboard shortcuts unintentionally shift focus on Windows OS

### LINKED ISSUE #62967: [Input] - Impossible to paste a value in all input fields with CTRL +V
### Description

It's impossible to paste a value in all input fields with CTRL +V, it seems to trigger a TAB

1. Go to admin right panel 
2. Copy a value CTRL + C
3. Try to paste this value in another field with CTRL + V

[353c2e5b-36ab-4848-a1d2-e4aa6ffd3995.webm](https://github.com/WordPress/gutenberg/assets/62699215/26edecf8-8b5f-424e-b12e-f5559c581af0)

### Environment info

- WP 6.5.5
- Every themes
- Windows 10
- Chrome 126.0.0.0, Firefox 127.0, Edge 126.0.2592.68

Yes

Yes

t-hamano:
Follow-up #39303
Fixes #62967

## What?

This PR fixes an issue where the focus unintentionally moves to the unit when executing the paste shortcut (`Ctrl+V`) in an input field on Windows OS.

## Why?
Only `event.metaKey` is checked to give focus to the unit, which may work on MacOS, but on Windows OS the `Ctrl` key is `event.ctrlKey`.

## Testing Instructions

The following are test procedures for Windows OS, but they should continue to work as expected on MacOS.

- Run the Storybook.
- Access `localhost:50240/?path=/docs/components-experimental-unitcontrol--docs`
- Copy the numbers beforehand.
- Run `Ctrl+V` in the input field.
- The copied numbers should be pasted.
- Press only the `p`, `v` or `e` key.
- The focus should move to the unit, and the unit should change according to the key pressed.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/536a7482-90e5-4ebd-98ab-83557b403626


packages/components/src/unit-control/index.tsx
```diff
@@ -172,6 +172,7 @@ function UnforwardedUnitControl(
 			// matches the first character of a unit.
 			if (
 				! event.metaKey &&
+				! event.ctrlKey &&
```
tyxla, line 175: Let's also update the comment above (to reference the `ctrl` key in addition to the `meta` key).

tyxla (approved):
LGTM, thanks @t-hamano 🙌 

I'd only update the inline comment above your change before shipping.

Otherwise, ready to 🚀

packages/components/src/unit-control/index.tsx
t-hamano: Thanks for the review! Updated by b90cc5a

PR #61159, opened 2024-04-26
Data views: Unbox items in grid layout

jameskoster:
Recently-added field display options ([badge](https://github.com/WordPress/gutenberg/pull/60284), [row / column](https://github.com/WordPress/gutenberg/pull/60083)) and other tweaks like [decreasing checkbox size](https://github.com/WordPress/gutenberg/pull/60475) improve information density in grid layout, thereby alleviating the need to visually stress the grid structure. 

In this PR the 'box' around each grid item is removed giving the overall layout more room to breath.

### Before
Screenshot 2024-04-26 at 15 56 09

### After
Screenshot 2024-04-26 at 15 55 40

### Testing instructions
1. Open a data view such as Patterns or Pages in the site editor.
2. Switch to grid layout
3. Observe the updated grid item display
4. Toggle fields on and off and ensure everything appears as expected

PR #62991, opened 2024-06-29
DataViews: Restore preview focus outline in grid layout

### LINKED ISSUE #62958: Data Views Grid layout: The focusable preview has no visible focus style
### Description

In the grid layout, items are presented with a preview (or with a grey empty area) which is also a button.
If the item is editable, activating the button opens the editor.
If the item is not editable, the button is disabled.

When you hover over an editable item, the cursor changes to the hand, but it is the only visual indicator that the item preview is hovered.
When you use the keyboard to navigate between focusable elements on the page, there is no visual indicator that the item preview is focused.

This means that is is not possible to see where the current focus is.
For example, if you move focus to the "View options" option and press the tab key, the focus is on the first item preview, but it is not visible, and pressing tab again moves focus to the checkbox next to the item name.

Activate a block theme.
Open Editor > Termplates.
Make sure the Grid layout is active.
On this screen, place the cursor in the search field.
Use the tab key repeatedly to move forwards.
Notice that when you reach the list of templates, the preview has no visible focus style.

Repeat the test for pages and patterns.

_No response_

### Environment info

_No response_

Yes

Yes

t-hamano:
Fixes #62958

## What?

This PR fixes an issue where the focus outline was not displayed on the preview button in the grid layout.

## Why?

This issue is probably related to #61159 where we unboxed grid items. The box-shadow is invisible because `overflow: hidden` is applied to the preview button wrapper `.dataviews-view-grid__media`.

## How?

I removed `overhidden:flow` and adjusted the rounded corners. The preview button itself already has `overflow: hidden`, so there should be no problem with the current unboxed style.

## Testing Instructions

- Switch to a grid layout in Patterns, Templates, or Pages view.
- Use the keyboard to focus on the Preview button.
- You should see a focus outline.
- Test different variations, such as items with less content and items with background color, to make sure they all work correctly.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/a7afc919-fd01-4d78-8135-8d39a74ea034

### DIFF
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -340,7 +340,6 @@
 		aspect-ratio: 1/1;
 		background-color: $gray-100;
 		border-radius: $grid-unit-05;
-		overflow: hidden;
 		position: relative;
 
 		img {
--- a/packages/edit-site/src/components/page-patterns/style.scss
+++ b/packages/edit-site/src/components/page-patterns/style.scss
@@ -35,13 +35,6 @@
 		display: flex;
 		flex-direction: column;
 		height: 100%;
-		border-radius: 3px 3px 0 0;
-
-		&.is-viewtype-grid {
-			.block-editor-block-preview__container {
-				border-radius: 3px 3px 0 0;
-			}
-		}
 
 		&.is-viewtype-table {
 			width: 96px;
@@ -62,7 +55,7 @@
 			cursor: pointer;
 			overflow: hidden;
 			height: 100%;
-			border-radius: 3px 3px 0 0;
+			border-radius: $grid-unit-05;
 
 			&:focus-visible {
 				box-shadow: 0 0 0 var(--wp-admin-border-width-focus) var(--wp-admin-theme-color);
--- a/packages/edit-site/src/components/page-templates/style.scss
+++ b/packages/edit-site/src/components/page-templates/style.scss
@@ -2,7 +2,6 @@
 	display: flex;
 	flex-direction: column;
 	height: 100%;
-	border-radius: 3px 3px 0 0;
 
 	.page-templates-preview-field__button {
 		box-shadow: none;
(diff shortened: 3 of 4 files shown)


t-hamano:
I would like to add a backport label to this PR because I consider it a regression that occurred in WP6.6.

carolinan:
Thank you. I tested on Chrome on Windows 11
On the template previews, the bottom right and left corners have no border radius.

packages/edit-site/src/components/page-pages/style.scss
```diff
@@ -43,7 +43,7 @@
 	overflow: hidden;
 	height: 100%;
 	width: 100%;
-	border-radius: 3px 3px 0 0;
+	border-radius: 3px;
```
jameskoster: Should we make this 4px?

That would be consistent with other UI's. I don't think we use 3px elsewhere. It would also fix this small discrepancy in the corners, where the background bleeds through between the image and the focus ring:

Screenshot 2024-07-01 at 11 22 39

jameskoster:
Good catch :)

t-hamano:
Thank you for your reviews.

The size should now be unified to 4px, and all four corners should be rounded on all screens.

By the way, the style around Dataview has changed significantly in the trunk, so this PR resolves the conflict. To backport this PR to WP6.6, cherry-picking will not work, and we will need to manually create a commit with the changes for WP6.6 😅

jasmussen (approved):
Before:

before

after

carolinan (approved):

t-hamano:
> By the way, the style around Dataview has changed significantly in the trunk, so this PR resolves the conflict. To backport this PR to WP6.6, cherry-picking will not work, and we will need to manually create a commit with the changes for WP6.6

There were no problems - the automatic cherry-pick committed the changes correctly and I verified that the wp/6.6 branch showed the focus outline we expected.

PR #60689, opened 2024-04-12
Patterns: remove "Manage all parts" page & link

oandregal:
Part https://github.com/WordPress/gutenberg/issues/59659 and https://github.com/WordPress/gutenberg/issues/55083

## What?

- Removes the link to "Manage all template parts" in patterns.
- The `/wp_template_part/all` URL is still accessible to hybrid themes (classic themes that declare support for `block-template-part`).
- The `/patterns` URL is still accessible unofficially to classic themes (with or without support for `block-template-part`).

| | Before | After |
| --- | --- | --- |
| **Block themes**. The "Manage all parts" link is no longer accessible. | Captura de ecrã 2024-04-22, às 12 47 32 | Captura de ecrã 2024-04-22, às 12 55 08 |
| **Hybrid themes**. They still can access the `/wp_template_part/all` URL, as before | Captura de ecrã 2024-04-22, às 12 47 49 | Captura de ecrã 2024-04-22, às 13 10 32 |
| **Hybrid themes**. They still can access the unofficially supported `/patterns` URL, as before | Captura de ecrã 2024-04-22, às 12 48 15 | Captura de ecrã 2024-04-22, às 13 10 39 |
| **Classic themes**. They still can access the unofficially supported `/patterns` URL, as before | Captura de ecrã 2024-04-22, às 12 48 33 | Captura de ecrã 2024-04-22, às 13 00 00 |

## Why?

- That's the target design we want, as per https://github.com/WordPress/gutenberg/issues/59659
- Over small PRs, all functionality have been migrated to the existing Parts pages in Patterns, so this is just removing the link and the code.

## How?

> The `/wp_template_part/all` URL still exists, because hybrid themes require it.

This PR updates the components that are used for that route. It now loads the code from Patterns (sidebar & content components). The reason for this is that in a future step hybrid themes will also have access to the Patterns from the site editor, so we'll load the same components. 

Changes:

- Router: `packages/edit-site/src/components/layout/router.js`
  - Updates the components used. From `PageTemplatesTemplateParts` to `PagePatterns` and from `SidebarNavigationScreenTemplatesBrowse` to `SidebarNavigationScreenPatterns`.
- Patterns
  - Sidebar: `packages/edit-site/src/components/sidebar-navigation-screen-patterns/index.js`. Configures the component depending on whether we're loading the `/wp_template_part/all` URL (display only Template parts) or `/patterns` (display everything, Patterns & Parts).
[...]

Issue #62878, opened 2024-06-26
Navigational dead end on smaller viewports


jameskoster:
On smaller viewports, once you navigate into a data view in the Site Editor (e.g. Pages or Patterns), you end up stuck in that view. There's no way to return to the main navigation, or (in the case of Patterns & Templates) open the Editor.

https://github.com/WordPress/gutenberg/assets/846565/352ff578-54ea-4d67-8b8e-d3fb9c401352

Related: https://github.com/WordPress/gutenberg/issues/60847.

youknowriad:
I think there's two issues here:

 - Not having the ability to return to previous pages: I think this is arguable because you can always click back in your browser but I agree that it's inconsistent with desktop and we might want to bring a navigation bar/breadcrumb on top.


 - The fact that you can't navigate to the editor is the most concerning one and would love to have it fixed for 6.6. So I'm temptatively adding this issue to 6.6 for this issue.

jameskoster:
> I think this is arguable because you can always click back in your browser

True, but that will not work well in the following common flow: `Open site editor → Pages → Select page → Edit → Click (W) Button → Click browser back button`.

In this case the back button would return you to the Editor.

t-hamano:
Perhaps it's related to #62106 as well?

ramonjd:
> Not having the ability to return to previous pages: I think this is arguable because you can always click back in your browser but I agree that it's inconsistent with desktop and we might want to bring a navigation bar/breadcrumb on top.

You can use the browser's back button and it will take you back, but only back to where you came from. 😄 

 So if you click into Templates, and then to the Editor (you'll need https://github.com/WordPress/gutenberg/pull/63002), and then back to Templates, clicking the browser's back button will take you back to the Editor. 


https://github.com/WordPress/gutenberg/assets/6458278/d611fcc0-f951-4010-8da4-adad628359b4



It would useful to be able to navigation back to the Design menu directly. 

I didn't have a lot of time, but I hacked together a very quick workaround for this. 

https://github.com/WordPress/gutenberg/assets/6458278/2f7bead0-d722-4b0d-980d-d657f2a67f2a

> The fact that you can't navigate to the editor is the most concerning one and would love to have it fixed for 6.6. So I'm temptatively adding this issue to 6.6 for this issue.

Agree, this one is the blocker. Looking at a fix over in https://github.com/WordPress/gutenberg/pull/63002

oandregal:
[...]

PR #63002, opened 2024-07-01
Site Editor: patterns and templates cannot be edited from sidebar mobile view

### LINKED ISSUE #62649: Site Editor: Unable to select specific template on mobile
### Description

When navigating to the Site Editor using a mobile-sized viewport, it is not possible to tap templates (please see the screen recording below).

The issue can be reproduced with versions higher than `16.4.0`.

1. Use a mobile-sized viewport.
2. In wp-admin, open the navigation, and open the site editor (Appearance → Editor).
3. Select "Templates"
4. Attempt to select a template.

Here is the screen recording of the issue:

https://github.com/WordPress/gutenberg/assets/25105483/bf62463c-70d6-458d-a5e1-107b5ddfc806

### Environment info

Higher than `16.4.0`, e.g. `16.5.0` or `18.6.0-rc.1`

Yes

Yes

### LINKED ISSUE #62932: Site Editor: "Edit Template" links unresponsive on mobile browsers (iPhone & Android)
### Description

Using the plugin (18.6.1) ellipsis buttons are unresponsive .

This happens with various mobile browsers but only on iPhone. Android works fine.

1. Go to Appearance > Editor on an iPhone
2. Try clicking on any of the ellipsis buttons ⋮ and notice how they do not work.
3. On the Templates section, the drop-down menu does show, but the Edit button does nothing.

### Environment info

Gutenberg 18.6.1

Yes

Yes

ramonjd:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/62932
Fixes https://github.com/WordPress/gutenberg/issues/62649

Specifically, this PR addressing the following:

1. Clicking on Edit from the Site Editor sidebar in "view" mode does not redirect to the editor.
2. Clicking on the WP logo once in edit mode crashes the browser

cc @oandregal to make sure everything from https://github.com/WordPress/gutenberg/pull/60689 still works

## Why?

So that clicking Edit on Templates and Patterns in the data views redirects the browser to the block editor in edit mode.

Also, to avoid crashing the browser.

## How?
In the mobile area prop in the Site Editor router, check for `canvas === 'edit'` and render the Editor component.

Skip using `document.startViewTransition` in mobile view when switching canvas modes.

## Testing Instructions

1. Fire up this branch and head to the Site Editor
3. Ensure your browser is narrower than `782px` wide to trigger the mobile view
4. Open the Templates sidebar panel
5. Click on a template or click on "Edit" in the ellipsis menu
6. The page should redirect to the editor
7. Click on the WP logo to return to "view" mode. The browser shouldn't crash. This primarily affected Chrome browsers for me, but it be related to the fact that `document.startViewTransition` doesn't have [full support outside Chrome](https://developer.mozilla.org/en-US/docs/Web/API/Document/startViewTransition).
8. Repeat the above for custom patterns and template parts

### Before

https://github.com/WordPress/gutenberg/assets/6458278/4d612d5c-ca98-467c-87ef-b923afd77c03

### After

https://github.com/WordPress/gutenberg/assets/6458278/a86453be-2a20-4795-abef-510904ef592f

### DIFF
--- a/packages/edit-site/src/store/private-actions.js
+++ b/packages/edit-site/src/store/private-actions.js
@@ -21,7 +21,7 @@ export const setCanvasMode =
 				registry.dispatch( editorStore ).setDeviceType( 'Desktop' );
 				registry
 					.dispatch( blockEditorStore )
-					.__unstableSetEditorMode( mode );
+					.__unstableSetEditorMode( 'edit' );
 				const isPublishSidebarOpened = registry
 					.select( editorStore )
 					.isPublishSidebarOpened();
--- a/packages/edit-site/src/store/private-actions.js
+++ b/packages/edit-site/src/store/private-actions.js
@@ -13,15 +13,15 @@ import { store as editorStore } from '@wordpress/editor';
 export const setCanvasMode =
 	( mode ) =>
 	( { registry, dispatch } ) => {
+		const isMediumOrBigger =
+			window.matchMedia( '(min-width: 782px)' ).matches;
 		const switchCanvasMode = () => {
 			registry.batch( () => {
-				const isMediumOrBigger =
-					window.matchMedia( '(min-width: 782px)' ).matches;
 				registry.dispatch( blockEditorStore ).clearSelectedBlock();
 				registry.dispatch( editorStore ).setDeviceType( 'Desktop' );
 				registry
 					.dispatch( blockEditorStore )
-					.__unstableSetEditorMode( 'edit' );
+					.__unstableSetEditorMode( mode );
 				const isPublishSidebarOpened = registry
 					.select( editorStore )
 					.isPublishSidebarOpened();
@@ -59,7 +59,11 @@ export const setCanvasMode =
 			} );
 		};
 
-		if ( ! document.startViewTransition ) {
+		/*
+		 * Skip transition in mobile, otherwise it crashes the browser.
+		 * See: https://github.com/WordPress/gutenberg/pull/63002.
+		 */
+		if ( ! isMediumOrBigger || ! document.startViewTransition ) {
 			switchCanvasMode();
 		} else {
 			document.documentElement.classList.add(
--- a/packages/edit-site/src/components/layout/router.js
+++ b/packages/edit-site/src/components/layout/router.js
@@ -75,6 +75,7 @@ function useRedirectOldPaths() {
 export default function useLayoutAreas() {
 	const { params } = useLocation();
 	const { postType, postId, path, layout, isCustom, canvas } = params;
+	const hasEditCanvasMode = canvas === 'edit';
 	useRedirectOldPaths();
 
 	// Page list
@@ -91,13 +92,12 @@ export default function useLayoutAreas() {
 					/>
 				),
 				content: <PostsList postType={ postType } />,
-				preview: ( isListLayout || canvas === 'edit' ) && <Editor />,
-				mobile:
-					canvas === 'edit' ? (
-						<Editor />
-					) : (
-						<PostsList postType={ postType } />
-					),
+				preview: ( isListLayout || hasEditCanvasMode ) && <Editor />,
+				mobile: hasEditCanvasMode ? (
+					<Editor />
+				) : (
+					<PostsList postType={ postType } />
+				),
 			},
 			widths: {
 				content: isListLayout ? 380 : undefined,
@@ -115,8 +115,8 @@ export default function useLayoutAreas() {
 					<SidebarNavigationScreenTemplatesBrowse backPath={ {} } />
 				),
 				content: <PageTemplates />,
-				preview: ( isListLayout || canvas === 'edit' ) && <Editor />,
-				mobile: <PageTemplates />,
+				preview: ( isListLayout || hasEditCanvasMode ) && <Editor />,
+				mobile: hasEditCanvasMode ? <Editor /> : <PageTemplates />,
 			},
 			widths: {
 				content: isListLayout ? 380 : undefined,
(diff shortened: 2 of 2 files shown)


andrewserong (commented):
Thanks for looking into this! In general the links going deeper into page templates is working for me 👍

One issue I ran into, not sure if it's related, is that the W icon at the top left when attempting to go back from the edit view winds up crashing for me, with no error in the console:

https://github.com/WordPress/gutenberg/assets/14988353/92c67e99-a39b-4c36-8684-c3ceeb44582d

This doesn't happen for me when navigating up the hierarchy from editing a template. Only while editing a pattern. Can anyone reproduce that? Just wondering if it's something weird with my environment.

Not _quite_ related to this bug, but one thing I found confusing at first with the mobile version of the dataviews screens is that there's no W icon, so I had to rely on browser back buttons to navigate up the hierarchy. This is challenging to do when we add in the ability to go to the edit view on mobile, as the back button might take us "back" to the editor if we've come from there, so it can be difficult to navigate up to the root of the site editor on mobile if we're going into the edit view.

Here's a quick demo:

https://github.com/WordPress/gutenberg/assets/14988353/1850d809-d354-42a3-9e46-1bacc14fdfae

1. Go to Templates screen
2. Tap on a template to go to the edit view
3. Tap on the W icon to go up a level
4. There's no way to get out via tapping here, so use the browser back button
5. We wind up back in the editor again

andrewserong:
> Repeat the above for custom patterns and template parts

Just checking if it's only me, but I wasn't able to find template parts from the mobile view. When I go to Patterns, I can only see filters for the synced or not synced patterns, and there doesn't appear to be links for the template parts like there is on Desktop 🤔 

image

ramonjd:
Thanks for testing @andrewserong 🙇🏻 

So it sounds like the entire flow is broken for mobile. A bit of a shame, I was hoping for a quick win. I should know better 🤣 

I'll dig a little more deeply this week to suss out where things fit together and see what's fixable.

🍺

t-hamano:
> One issue I ran into, not sure if it's related, is that the W icon at the top left when attempting to go back from the edit view winds up crashing for me, with no error in the console:

I was able to reproduce this issue. The easiest way is to go to the Pages screen, open any editor canvas, and click the W logo.

- This issue can be reproduced on any small viewport, regardless of whether it is a mobile device or not.
- Does not reproduce in WP6.6 RC1. Reproduces in Gutenberg trunk.

I think there are still many issues with the mobile viewport. The main issue is probably #62878.

It's a difficult question as to what extent this needs to be fixed in WP6.6 😅

ramonjd:
> It's a difficult question as to what extent this needs to be fixed in WP6.6 

Thanks for the thoughts @t-hamano 

After thinking about it for a nanosecond, I'll put this out there. For 6.6, users should at least be able to:

- edit templates et al. in narrow viewport width 
- navigate back to the "view" sidebar

So I'll investigate the crash. With those things fixed hopefully we've restored base functionality. 

Any UX cluster bombs we can fix as follow ups.

ramonjd:
> Just checking if it's only me, but I wasn't able to find template parts from the mobile view. When I go to Patterns, I can only see filters for the synced or not synced patterns, and there doesn't appear to be links for the template parts like there is on Desktop

Forgot to reply to this, sorry.

I don't see any filters for template parts on Desktop either. 

Screenshot 2024-07-02 at 11 12 43 AM

It looks like Template Parts are accessible also from the sidebar view, but in mobile view, this sidebar is not accessible so there's no way for users to list them. 😅

ramonjd:
> So I'll investigate the crash. With those things fixed hopefully we've restored base functionality.

I think I've traced the crash to the use of `document.startViewTransition` in [setCanvasMode](https://github.com/WordPress/gutenberg/blob/1a8c9485ca364ed9c7b9a35b544332613be18254/packages/edit-site/src/store/private-actions.js#L13-L13). 

Only firing that in desktop mode seems to have fixed the crashing for me, however I haven't located the root cause. Chrome supports `document.startViewTransition` - I expect it short circuits rendering somehow as the entire HTML document is kaputt. Not sure.

ramonjd:
> I found confusing at first with the mobile version of the dataviews screens is that there's no W icon, so I had to rely on browser back buttons to navigate up the hierarchy. 

There's an issue for this: https://github.com/WordPress/gutenberg/issues/62878 

I hacked together something to address it in the short term, and have commented over there. 👍🏻

packages/edit-site/src/store/private-actions.js
```diff
@@ -13,15 +13,15 @@ import { store as editorStore } from '@wordpress/editor';
...
 		const switchCanvasMode = () => {
 			registry.batch( () => {
-				const isMediumOrBigger =
-					window.matchMedia( '(min-width: 782px)' ).matches;
 				registry.dispatch( blockEditorStore ).clearSelectedBlock();
 				registry.dispatch( editorStore ).setDeviceType( 'Desktop' );
 				registry
 					.dispatch( blockEditorStore )
-					.__unstableSetEditorMode( 'edit' );
+					.__unstableSetEditorMode( mode );
```
youknowriad: I'm not sure this change is correct. There's no "view" mode in the block editor mode.

ramonjd: Thanks for catching that. That was a hangover from debugging 🤦🏻 

I'll update quickly

youknowriad (approved):
LGTM 👍

ramonjd:
Thanks for the reviews, folks!

ramonjd:
Cherry-pick and merge to the wp/6.6 branch here:

- https://github.com/WordPress/gutenberg/pull/63055

oandregal:
I've prepared https://github.com/WordPress/gutenberg/pull/63060 as a potential quick change for 6.6.

PR #63010, opened 2024-07-01
Excerpt block: add example of the block

MaggieCabrera:
## What?

Add n example to the `core/post-excerpt` block.

## Why?

## How?

By defining a viewport width size for the example

Create a post and add some content to it.
Open the block inserter and search for the excerpt block, hover over it and you'll see the example

## Screenshots or screencast 
Screenshot 2024-07-01 at 12 05 02


jameskoster (approved):
Working for me.

PR #63012, opened 2024-07-01
Upgrade TypeScript to 5.5

sirreal:
## What?

- Upgrade TypeScript to 5.5
- Dedupes packages
- Add a jsonc langauge hint for tsconfig files.

## Why?

TypeScript 5.5 brings a number of nice changes, upgrade it:
https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/

https://github.com/WordPress/gutenberg/pull/63605 is a follow-up to leverage some of the new TS 5.5 features.

Deduping packages is a good idea in general. Npm is supposed to try to dedupe, but we seem to wind up with duplicates anyways:

https://github.com/WordPress/gutenberg/blob/4730e122c6d608cd2601fc0132e19628654f07ae/.npmrc#L4

The jsonc language hint will allow better rendering of tsconfig files in GitHub UI for things like comments in JSON (allowed in tsconfig.json files).

## Testing Instructions

CI passes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,3 +20,6 @@ changelog.txt linguist-language=Markdown
 
 # Flag docs directory as documentation for GitHub stats.
 docs/** linguist-documentation
+
+# TSConfig files use jsonc.
+tsconfig*.json linguist-language=jsonc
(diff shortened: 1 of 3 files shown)


jsnajdr (approved):
The lockfile dedupe is quite radical, many unrelated changes, even package upgrades. But why not, we should do such updates regularly.

Issue #35907, opened 2021-10-25
Gallery Refactor: Restore data-id attribute for image blocks


celloexpressions:
### Description

Block gallery image markup has included a `data-id` attribute on each gallery image `<img>` tag since WordPress 5.0. This attribute includes the image attachment post object's id for images hosted locally in the media library. It allows plugins and themes to access other structured data related to the image through the core `attachment` post type. This can be used in both PHP and JS to pre- or post-process the post content markup.

The data-id attribute is currently missing from the refactored gallery block (as of Gutenberg 11.7.1). Now that all gallery images are image blocks, this attribute should be provided for all locally-hosted images in the content. Adding the attribute to more images is much less likely to cause breakage than removing it from existing markup and on new gallery posts; in fact, it would probably resolve issues with current extender usage of this markup. In my usage, removing the attribute breaks several themes/plugin/sites that I maintain.

This was previously fixed here: https://github.com/WordPress/gutenberg/pull/30710, but appears to have been reverted. I can't find any discussion of why it was reverted. This is the only remaining blocking issue that I see for including the gallery refactor in core.

A possible backup plan would be to instead fix support for the core `get_post_galleries()` function that has been broken since WordPress 5.0. That would allow some programmatic relationship between images in the content and their associated attachment post objects. However, that would still be a break in backwards compatibility and require theme/plugin changes.

### Step-by-step reproduction instructions

1. Create an image gallery in WordPress 5.8 with images uploaded to the media library. 
2. Observe that the published gallery markup includes a `data-id` attribute on the gallery `<img>` tags.
3. Enable the gallery refactor experiment with Gutenberg 11.7 and create another image gallery with images uploaded to the media library.
4. Observe that the published gallery markup does not include the `data-id` attribute on the gallery `<img>` tags.
5. An example of the resulting breakage can be seen with the Arbutus theme. Using the `gallery` post format, the front page excerpt view should display a partial gallery with the first several images from in the post content. This doesn't work with the refactored gallery block - a blank bar displays instead.

### Screenshots, screen recording, code snippet

This was previously fixed with the code here: https://github.com/WordPress/gutenberg/commit/a2253ec518f3769b280c7c7c27e8ea1d0f265a3f

### Environment info

Testing with WordPress 5.9-alpha and Gutenberg 11.7.1 with the gallery block refactor experiment enabled.

### Please confirm that you have searched existing issues in the repo.

Yes

[...]

PR #63013, opened 2024-07-01
Image block: Ensure extenders that rely on media ids in block html are supported by block bindings

### LINKED ISSUE #62886: Synced Pattern Overrides: Image metadata like data-permalink, data-image-title, keeps original image data
### Description

When you create a synced pattern and include an image override, the original image's metadata will be shown in the HTML even when you've chosen a new image in a pattern's instance. So if the image was called 'my-image-placeholder.jpg", that is the data that will be shown in the code, even if the displayed image is "bird-graphic.jpg".

For example:
`Figure 5. A histogram plot showing distribution of missing death data according to age group`

P.S. I might open another issue if I have time, but so I don't forget, the same thing happens if you had enabled a link in the image after creating the override. The best example being the "attachment" link which opens the image in a lightbox. It also shows the placeholder image in the lightbox instead of the current instance's image.

Thanks!

1. Create a synced pattern that contains an image block.
2. Enable overrides on the image block.
3. Place the synced pattern on a post or page.
4. Replace the image with another image.
5. Load the page/post and view the HTML code.

Screenshot 2024-06-26 at 11 45 07 AM

### Environment info

WordPress 6.5.5, Gutenberg 18.6.0

Yes

Yes

talldan:
Fixes #62886

## What?
The image block outputs the image id in its markup in a couple of places—a `data-id` attribute (which only appears for images in galleries) and a `wp-image-123` classname.

When pattern overrides (and to some extend, other types of block binding) are used, these ids can be incorrect in the front-end. The html will still show the id from the original pattern and not for the overridden image. 

## Why?
In a couple of issues, it's mentioned how these are used by extenders to add additional features to the image block:
- [Gallery Refactor: Restore data-id attribute for image blocks](https://github.com/WordPress/gutenberg/issues/35907#top)
- [Synced Pattern Overrides: Image metadata like data-permalink, data-image-title, keeps original image data](https://github.com/WordPress/gutenberg/issues/62886#top)

## How?
Dynamically update the markup of the image block to use the correct ids.

For the classname, a new bit of code is added that detects a mismatch for the id, and if so updates the classname.

For the data attribute, the defacto `id` attribute is used instead of the injected `data-id` attribute, since it'll have the correct value.

## Testing Instructions
1. Create a new post
2. Add a gallery that contains an image from your media library.
3. Take note of the image id (go to the code editor and look for the image block's `id` attribute)
4. Back in the visual editor, select the gallery and create a synced pattern from it (Block toolbar > Options > Create pattern)
5. Select 'Edit original' from the pattern block's toolbar to edit the pattern
6. Select the image block in the gallery, and from the Advanced inspector tools, select 'Enable overrides'. 
7. In the dialog, give the override any name and enable it.
8. Save the pattern and select the 'Back' button from the editor top bar to go back to the post originally being edited
9. Select the image again, and replace the image with a different one from your media library
10. Take a note of the new image id (go to the code editor and find the id in the `wp:block` `content` attribute)
11. Preview the post
12. Inspect the image, it should have a `data-id` attribtue and `wp-image-` class that reflect the id you noted in step 10.

### DIFF
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -28,33 +28,44 @@ function render_block_core_image( $attributes, $content, $block ) {
 		return '';
 	}
 
-	if ( isset( $attributes['id'] ) ) {
-		$id = $attributes['id'];
+	$has_id_binding = isset( $attributes['metadata']['bindings']['id'] ) && isset( $attributes['id'] );
 
-		// Use block context to detect whether the block is nested in a gallery.
-		// As the context names are very generic, check multiple array keys to
-		// reduce the chance a non-gallery parent is providing the same context.
-		$is_gallery_child = array_key_exists( 'allowResize', $block->context )
-			&& array_key_exists( 'imageCrop', $block->context )
-			&& array_key_exists( 'fixedHeight', $block->context );
-
-		// Adds the data-id="$id" attribute to the img element to provide backwards
-		// compatibility for the Gallery Block, which now wraps Image Blocks within
-		// innerBlocks. The data-id attribute is added in a core/gallery
-		// `render_block_data` hook.
-		if ( $is_gallery_child ) {
-			$p->set_attribute( 'data-id', $id );
-		}
-
+	// Ensure the `wp-image-id` classname on the image block supports block bindings.
+	if ( $has_id_binding ) {
 		// If there's a mismatch with the 'wp-image-' class and the actual id, the id was
 		// probably overridden by block bindings. Update it to the correct value.
 		// See https://github.com/WordPress/gutenberg/issues/62886 for why this is needed.
+		$id = $attributes['id'];
 		$image_classnames  = $p->get_attribute( 'class' );
-		$expected_id_class = "wp-image-$id";
-		if ( is_string( $image_classnames ) && ! str_contains( $image_classnames, $expected_id_class ) ) {
-			$image_classnames = preg_replace( '/wp-image-(\d+)/', "wp-image-$id", $image_classnames );
+		$class_with_binding_value = "wp-image-$id";
+		if ( is_string( $image_classnames ) && ! str_contains( $image_classnames, $class_with_binding_value ) ) {
+			$image_classnames = preg_replace( '/wp-image-(\d+)/', $class_with_binding_value, $image_classnames );
 			$p->set_attribute( 'class', $image_classnames );
 		}
+	}
+
+	// For backwards compatibility, the data-id html attribute is only
+	// set for image blocks nested in a gallery.
+	// `data-id` is an attribute set by the gallery block using a filter,
+	// so checkings it's set serves this purpose.
+	if ( isset( $attributes['data-id'] ) ) {
+		$data_id;
+
+		if ( $has_id_binding ) {
+			// If there is a binding for the id, use the `id` attribute.
+			// This ensures the `data-id` html attribute contains the correct
+			// block binding value since the `data-id` block attribute does
+			// not support bindings.
+			$data_id = $attributes['id'];
+		} else {
+			// If there are no bindings for the id, for backward compatibility
+			// use the `data-id` attribute added by the gallery block.
+			// This attribute may be filtered by third parties to change what's
+			// stored in the data-id html attribute.
+			$data_id = $attributes['data-id'];
+		}
+
+		$p->set_attribute( 'data-id', $data_id );
 	}
 
 	$link_destination  = isset( $attributes['linkDestination'] ) ? $attributes['linkDestination'] : 'none';
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -39,7 +39,7 @@ function render_block_core_image( $attributes, $content, $block ) {
 		// If there's a mismatch with the 'wp-image-' class and the actual id, the id was
 		// probably overridden by block bindings. Update it to the correct value.
 		// See https://github.com/WordPress/gutenberg/issues/62886 for why this is needed.
-		$image_classes = $p->get_attribute( 'class' );
+		$image_classes     = $p->get_attribute( 'class' );
 		$expected_id_class = "wp-image-$id";
 		if ( is_string( $image_classes ) && ! str_contains( $image_classes, $expected_id_class ) ) {
 			$image_classes = preg_replace( '/wp-image-(\d+)/', "wp-image-$id", $image_classes );
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -35,8 +35,8 @@ function render_block_core_image( $attributes, $content, $block ) {
 		// If there's a mismatch with the 'wp-image-' class and the actual id, the id was
 		// probably overridden by block bindings. Update it to the correct value.
 		// See https://github.com/WordPress/gutenberg/issues/62886 for why this is needed.
-		$id = $attributes['id'];
-		$image_classnames  = $p->get_attribute( 'class' );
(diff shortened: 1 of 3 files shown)


packages/block-library/src/image/index.php
```diff
@@ -28,12 +28,23 @@ function render_block_core_image( $attributes, $content, $block ) {
...
 
-	if ( isset( $attributes['data-id'] ) ) {
+	if ( isset( $attributes['id'] ) ) {
 		// Adds the data-id="$id" attribute to the img element to provide backwards
 		// compatibility for the Gallery Block, which now wraps Image Blocks within
 		// innerBlocks. The data-id attribute is added in a core/gallery
 		// `render_block_data` hook.
-		$p->set_attribute( 'data-id', $attributes['data-id'] );
+		$id = $attributes['id'];
+		$p->set_attribute( 'data-id', $id );
```
talldan: The side effect of this is that the image block will now always have `data-id`, whereas in trunk it's only added when the image block is in a gallery. Perhaps there's another way to detect when the image block is nested that doesn't rely on the extra block attribute.

May also need to add some back compat for the possibility that third parties filter the `data-id` block attribute. 🤔

packages/block-library/src/image/index.php
```diff
@@ -28,12 +28,33 @@ function render_block_core_image( $attributes, $content, $block ) {
 		return '';
 	}
 
-	if ( isset( $attributes['data-id'] ) ) {
```
talldan, line 31: The one thing I'm nervous about in this PR is removing the use of the temporary `data-id` block attribute, as plugins could be filtering it to a different value. Though I don't know why they would do that.

Unfortunately it's very difficult to check whether that might be the case, and removing it is needed for block bindings to work as expected.

I do think it's a bit of tech debt—particularly the way it causes the block's `id` is being stored in two attributes and the camel casing of the attribute name, which is non-standard. I can add a dev note, but keen to get the thoughts of others.

cbravobernal:
It was not working on my side cause I had Gutenberg disabled 🤦‍♂️

packages/block-library/src/image/index.php
cbravobernal: It is a breaking change 😢 . But filtering a temporary value is also not a really good development choice, right?

SantosGuillamot:
I've been discussing it with @cbravobernal to see if we could keep the backward compatibility.

I've made [this commit](ad46c2d9bd325eec3d8d44113dcbce3933982de0) with an alternative approach to keep the code as close as possible to the current implementation. It avoids removing the `data-id` filter to ensure it remains backward-compatible.

Feel free to revert it if you don't consider correct, @talldan . You might have more context than me.

Apart from that, this seems related to a bigger related topic to block bindings. We should reconsider whether the block attributes should be modified during the block parsing and not only during the block rendering to support this kind of filters. But I am not sure how that could work or the implications, so it can be discussed in another place.

packages/block-library/src/image/index.php
SantosGuillamot: As explained in [this comment](https://github.com/WordPress/gutenberg/pull/63013#issuecomment-2202771482), I made a commit with a potential alternative approach that doesn't require removing the temporary `data-id` block attribute. It seems to work and it seems closer to the existing implementation.

cbravobernal (approved):
Works as expected, and the solution implies less caveats. I also agree that @talldan has more context to do the final review. I leave the approve, as that way nor the PR creator or @SantosGuillamot has to do that.

packages/block-library/src/image/index.php
talldan: Thanks for pushing a commit. I did consider something similar to that, and I think it's not a bad choice, but still has some drawbacks.

With the new changes, the image block's `id`  attribute (`$rendered_id`) is still being used for the `data-id` html attribute, whereas in trunk it's the injected `data-id`, so I don't think it completely addresses the back compat concerns. One use case for filtering `data-id` would be to modify the html attribute to something else, and that would no longer work. My conclusion is that we can't really support `data-id` (the block attribute) properly, and that's why I lean towards removing it in favor of the regular `id` (if it's going to break, then it's a good opportunity to clean up the code as much as possible).

The changes also seem to couple the `wp-image-` classname to the `data-id` attribute (the whole logic is in the `isset( $attributes['data-id'] )` if statement), which wasn't the case before. `data-id` is only present for image blocks in galleries, so this would result in the classname only being replaced for gallery images and not standalone images. A small change would be needed to fix that.

talldan:
Thanks for adding the commit @SantosGuillamot and for reviewing @cbravobernal.

I decided to revert it due to what was mentioned in this comment - https://github.com/WordPress/gutenberg/pull/63013#discussion_r1663439826. The quick breakdown is that it didn't restore full backwards compatibility because the `data-id` block attribute isn't used the same way it was before. Weighing things up my feeling it that it might be better to remove the technical debt.

Perhaps it might be better to break this PR up into two parts - one for the `wp-image` classname, and another for the `data-id` that we can keep experimenting on (or even punt to 6.6.x or 6.7).

talldan:
I started a slack thread in the core-editor channel to ask for more input on this - https://wordpress.slack.com/archives/C02QB2JS7/p1719990975720999.

SantosGuillamot:
You're right that it doesn't fully address the backwards compatibility issue. There are three main aspects of the current status of the pull request that makes me a bit uncomfortable:
* Removing the `data-id` attribute as mentioned.
* The way it checks if it is inside a gallery block: [link](https://github.com/WordPress/gutenberg/pull/63013/files#diff-1e4aa29ea183efda9c795271dc1acf6e8049355db104d0dca5b0ea5caf7f619cR37-R39). I assume it works, but it feels a bit weird.
* Using regex to replace the class name: [link](https://github.com/WordPress/gutenberg/pull/63013/files#diff-1e4aa29ea183efda9c795271dc1acf6e8049355db104d0dca5b0ea5caf7f619cR54-R57). Although I am not sure how to do it differently.

Another option I was considering is to keep the existing solution in `trunk` and just modify the `data-id` and the class name when the `id` is bound. Something like this:

```php
	if ( isset( $attributes['data-id'] ) ) {
		$id = $attributes['data-id'];
		// The id attribute could change during rendering if it is connected.
		if ( ! empty( $attributes['metadata']['bindings']['id'] ) ) {
			$id = $attributes['id'];
		}
		$p->set_attribute( 'data-id', $id );
	}

	// Replace class name
	if ( isset( $attributes['id'] ) && ! empty( $attributes['metadata']['bindings']['id'] ) ) {
		$image_classnames = $p->get_attribute( 'class' );
		preg_match( '/wp-image-(\d+)/', $image_classnames, $matches );
		if ( ! empty( $matches ) ) {
			$wp_image_class_id = $matches[1];
		}

		if ( $wp_image_class_id !== $attributes['id'] ) {
			$p->remove_class( 'wp-image-' . $wp_image_class_id );
			$p->add_class( 'wp-image-' . $attributes['id'] );
		}
	}
```

Although I assume it is not ideal either.

talldan:
> Another option I was considering is to keep the existing solution in trunk and just modify the data-id and the class name when the id is bound.

That's a pretty good option. It keeps `data-id` working the same whenever there's no binding. I think the code would also have to check for the `__default` binding as I don't think the individual pattern override bindings will have been added yet by the processing code.

edit - I'll work on these changes tomorrow

kevin940726 (approved):
Tested and code looks good to me 👍. I didn't follow the entire convo above though, so not sure if it requires anything else?

talldan:
I did a final test with jetpack, and this fixes the issue, will go ahead and merge now.

Thanks everyone for the help finding a solution. ❤️

PR #63017, opened 2024-07-01
DataViews: Register the reset template and template part action like any third-party action

youknowriad:
Related #61084 
Similar to #62913 

## What?

The current PR explore the possibility to use the API to register one action: "reset template and template part". 

## Testing Instructions

1- Open a theme template
2- Edit and save edit.
2- You should be able to see the "reset" action in the actions dropdown menu

### DIFF
--- a/docs/reference-guides/data/data-core.md
+++ b/docs/reference-guides/data/data-core.md
@@ -790,7 +790,7 @@ _Parameters_
 -   _kind_ `string`: Kind of the entity.
 -   _name_ `string`: Name of the entity.
 -   _recordId_ `Object`: ID of the record.
--   _options_ `Object`: Saving options.
+-   _options_ `Object=`: Saving options.
 
 ### saveEntityRecord
 
--- a/packages/editor/src/dataviews/types.ts
+++ b/packages/editor/src/dataviews/types.ts
@@ -11,11 +11,15 @@ export interface BasePost {
 	status?: PostStatus;
 	title: string | { rendered: string };
 	type: string;
+	id: string | number;
 }
 export interface TemplateOrTemplatePart extends BasePost {
-	type: 'template' | 'template-part';
+	type: 'wp_template' | 'wp_template_part';
 	source: string;
 	has_theme_file: boolean;
 }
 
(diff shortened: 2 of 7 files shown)


youknowriad:
I would love an e2e test per action. I'll be looking into that.

youknowriad:
Ok, so looks like resetting already has e2e tests, so this should be ready.

youknowriad:
I rebased this based on your changes @Mamaduka I hope I didn't miss anything

docs/reference-guides/data/data-core.md
```diff
@@ -790,7 +790,7 @@ _Parameters_
 -   _kind_ `string`: Kind of the entity.
 -   _name_ `string`: Name of the entity.
 -   _recordId_ `Object`: ID of the record.
--   _options_ `Object`: Saving options.
+-   _options_ `Object=`: Saving options.
```
ntsekouras, line 793: Hm.. what's the `=` here?

packages/editor/src/dataviews/types.ts
```diff
@@ -11,11 +11,16 @@ export interface BasePost {
 	status?: PostStatus;
 	title: string | { rendered: string };
 	type: string;
+	id: string | number;
 }
 export interface TemplateOrTemplatePart extends BasePost {
-	type: 'template' | 'template-part';
+	type: 'wp_template' | 'wp_template_part';
```
ntsekouras, line 17: So, this was a TS bug? I'm curious why in our code base it didn't have any errors? 🤔

ntsekouras (approved):
Tests well, thank you!

docs/reference-guides/data/data-core.md
youknowriad: optional in typescript jsdocs

packages/editor/src/dataviews/types.ts
youknowriad: yes, but since these types are very new, they didn't impact much.

Mamaduka:
Thank you, @youknowriad!

PR #58207, opened 2024-01-24
Move Posts Per Page, Offset, and Pages controls from the block toolbar into Inspector Controls

ryanwelcher:
## What?
This PR moves the Posts Per Page, Offset, and Max Pages to Show controls into the Inspector controls side bar and wraps them in the new ToolsPanel component.

## Why?
Having controls in multiple locations (toolbar and sidebar) is bad UX and makes these controls easy to miss. I wrote a whole plugin around adding these in 😳 .

## Testing Instructions
1. Insert a new instance of the Query Loop block
2. Confirm that there is a a new Display section in the sidebar.
3. Confirm that Posts Per Page is displayed by default.
4. Confirm that this section does not appear when the block is set to inherit the query

## Screenshots or screencast 

screenshot

PR #63020, opened 2024-07-01
Make usage of the settings icon more consistent

### LINKED ISSUE #62873: Make usage of the `settings` icon more consistent and meaningful
### Description

Splitting this out from https://github.com/WordPress/gutenberg/pull/62129 where an interesting conversation about cohesive design language in the editor UI took place.

I believe cohesive language should apply also to icons _and their usage_ in the editor.

Some icons represent very familiar logical concepts for users. For example, a plus icon `+` immediately associates with the concepts of 'Add', 'Create new', 'Insert', or 'Increase'. 

Other icons need to be 'learned' in the context of the application and UI they are used into. This implies some initial cognitive load and learning curve. When users get familiar with the _meaning_ and _purpose_ of that icon, the icon helps identifying a specific feature in the user interface. As long as the _usage_ of that icon is consistent.

Noting that, from an accessibility perspective, proliferation of icons is a problem and it would be preferable to have icons accompanied with visible text labels, but that's another issue.

Consider the following icon:

Screenshot 2024-06-26 at 13 32 19

Its name is `settings`. Visually, it's two sliders made of a track and a handle (or 'knob'). Regardless of the icon name `settings`, I would say the logical association isn't immediate but can be learned. It's about values that can be 'adjusted' and 'set to' so that, in a broader sense, this icon represent the concept of 'settings'.

## How this icon is actually used in the editor

At the time of writing this issue I can find 5 occurrences of `icon={ settings }` in the codebase.

In the first three cases, this icon isn't used to access 'settings'. Actually, it is used to switch a control to an alternative version where users can set custom values.

1
`SpacingInputControl` component.
By clicking the icon button the range control shows an additional input field where users can enter a custom value. There are many instances of this component for the various margin, padding, block spacing values etc.

SpacingInputControl

2
`FontSizePicker` component.
Same: clicking the icon button switches the control from preset values to an alternative version where users can enter a custom value.

FontSizePicker

3
`ShadowInputControl` control in the `ShadowPopover` component.
Same: clicking the icon button switches the range slider to an input field where users can enter a custom value.

ShadowInputControl in ShadowPopover

> [!IMPORTANT]
> So far, as a user I learned that this icon button purpose is to 'switch' a control to an alternative version.

As a user, it took me a while to 'learn' what this icon button is about. Now that I know what it is about, I would expect this icon to _always be used consistently_ and always have the same purpose.

Instead, this is not the case.

4
The fourth occurrence of this icon usage is for the DataViews 'View options` button. Screenshot:

DataViews View actions

It opens a dropdown menu with a multitude of options for the data views layout, filtering, items to show, items per page. In this context, this icon button meaning and purpose are entirely different.

> [!IMPORTANT]
> As a user, I previously learned this icon means 'switch this control to an alternative version'. But now, suddenly, it means something else. 

IMHO, this inconsistent usage doesn't help in building a cohesive design language. Also, as a user, the cognitive load is now doubled because I have to remember that this icon has different meanings in different contexts.

5
The fifth occurrence of this icon is in the Query loop block toolbar. Screenshot:

Query block Display settings

It gives access to a popover where users can set the items per page, the offset, and the max pages to show for the Query.

Again, the icon meaning and purpose in this context is entirely different and inconsistent. As a user, I have now to remember that this icon means something entirely different in three different contexts.

## Conclusion
The 'settings' icon is a good example of how the actual icons _usage_ can be problematic when it's inconsistent. Ideally, icons that need to be 'learned' should have only _one, specific, consistent_ usage in an user interface. There is no point in forcing users to _learn_ what an icon means when the icon meaning changes depending on the context.

Overall, I'd like to see a more strict, cohesive usage of icons throughout the editor, especially for icons that users have to 'learn'. I'm not sure the current situation helps much with building a good user experience. 

To me, the cases 4) and 5) are less than ideal. Those controls should be changed to use another icon. The `settings` icon should be reserved to the 'switch' button for the controls alternative version.

The editor is already complicated. I'd think design should strive to _simplify wherever possible_ and have UI conventions that are consistent to reduce cognitive load as much as possible.

N/A
Go through the various UIs illustrated in this issue description.
Observe the inconsistent icon usage.

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/62873

## What?

This PR only aims to make the usage of the `settings` icon more consistent in the editor UI. This icon should only be used to switch a control to an alternative version, see for example the various margin, padding, spacing controls.

Instead, it was used inconsistently in at least 3 other cases:
- The icon button to open the Fonts Library, now replaced with a button with visible text after https://github.com/WordPress/gutenberg/pull/62129
- The 'View Options' icon button in the data views.
- The 'Display settings' icon button in the Query Loop block toolbar.

Screenshot of the icon, for reference:

Screenshot 2024-07-01 at 13 46 12

Work on this PR surfaced other issues. Most notably, the Query loop block had the 'Display settings' placed in the toolbar, while it makes more sense to have them in the Inspector together with other similar settings.
Other findings will be tracked in new issues.

## Why?

Icons should uniquely identify specific features or, at least, _similar_ features.
There is already an icon for the concept of 'settings': the `cog` icon is commonly used for this purpose. It makes sense to reuse popular conventions to make the editor UI more intuitive and less disorienting.

## How?

- Moves the Query Loop block 'Display settings' to the inspector.
- Simplifies the  'Items per page', 'Offset', and 'Max pages to show' settings by removing the `BaseControl`s that were wrapping the `NumberControl`s. Not sure why these wrapping `BaseControl`s were used in teh first place but they seem unnecessary. Also, the description of the 'Max pages to show' setting was incorrectly set to the wrapping `BaseControl` instead of the actual `NumberControl`.
- Renames 'Max page to show' (singular 'page') to 'Max pages to show' (plural 'pages')
- Changes the dataviews 'View Options` icon to use the `cog` icon.

### Old testing instructions

- Select a Query Loop block.
- Make sure 'Inherit query from template' in the Inspector is disabled.
- Observe they Query Loop block toolbar doesn't show the 'Display settings' button any longer.
- Observe the 'Items per page', 'Offset', and 'Max pages to show' settings are now in the Inspector.
- Test all the three settings and observe they work as expected.
- Note: the 'Max pages to show' doesn't update the preview in the editor, not even on trunk. I'll create a separate issue for this.
- Observe the 'Max pages to show' description is correctly associated with the number field via `aria-describedby` and ID attributes.
- Go to the Site Editor > Pages
- Observe the 'View Options' icon button now uses the `cog` icon.

### New testing instructions
- Go to the Site Editor > Pages
- Observe the 'View Options' icon button now uses the `cog` icon.
- Select a Query Loop block.
- Make sure 'Inherit query from template' in the Inspector is disabled.
- Click  'Display settings' in the Query Loop block toolbar.
- Observe the UI within the popover doesn't have relevant visual changes.
- Observe the description of the 'Max pages to show' number field is correctly associated to the input with aria-describedby.

## Screenshots or screencast 

The Query loop block settings moved to the Inspector:

Screenshot 2024-07-01 at 13 50 21

The new cog icon in the data views:

Screenshot 2024-07-01 at 12 43 25

### DIFF
--- a/test/e2e/specs/editor/various/is-typing.spec.js
+++ b/test/e2e/specs/editor/various/is-typing.spec.js
@@ -36,4 +36,36 @@ test.describe( 'isTyping', () => {
 		// Toolbar Popover is hidden again
 		await expect( blockToolbarPopover ).toBeHidden();
 	} );
+
+	test( 'should not close the dropdown when typing in it', async ( {
+		editor,
+		page,
+	} ) => {
+		// Add a block with a dropdown in the toolbar that contains an input.
+		await editor.insertBlock( { name: 'core/query' } );
+
+		await editor.canvas
+			.getByRole( 'document', { name: 'Block: Query Loop' } )
+			.getByRole( 'button', { name: 'Start blank' } )
+			.click();
+
+		await editor.canvas
+			.getByRole( 'button', { name: 'Title & Date' } )
+			.click();
+
+		// Moving the mouse shows the toolbar.
+		await editor.showBlockToolbar();
+		// Open the dropdown.
+		const displaySettings = page.getByRole( 'button', {
+			name: 'Display settings',
+		} );
+		await displaySettings.click();
+		const itemsPerPageInput = page.getByLabel( 'Items per Page' );
+		// Make sure we're where we think we are
+		await expect( itemsPerPageInput ).toBeFocused();
+		// Type inside the dropdown's input
+		await page.keyboard.type( '00' );
+		// The input should still be visible.
+		await expect( itemsPerPageInput ).toBeVisible();
+	} );
 } );
--- a/test/e2e/specs/editor/various/is-typing.spec.js
+++ b/test/e2e/specs/editor/various/is-typing.spec.js
@@ -36,36 +36,4 @@ test.describe( 'isTyping', () => {
 		// Toolbar Popover is hidden again
 		await expect( blockToolbarPopover ).toBeHidden();
 	} );
-
-	test( 'should not close the dropdown when typing in it', async ( {
-		editor,
-		page,
-	} ) => {
-		// Add a block with a dropdown in the toolbar that contains an input.
-		await editor.insertBlock( { name: 'core/query' } );
-
-		await editor.canvas
-			.getByRole( 'document', { name: 'Block: Query Loop' } )
-			.getByRole( 'button', { name: 'Start blank' } )
-			.click();
-
-		await editor.canvas
-			.getByRole( 'button', { name: 'Title & Date' } )
-			.click();
-
-		// Moving the mouse shows the toolbar.
-		await editor.showBlockToolbar();
-		// Open the dropdown.
-		const displaySettings = page.getByRole( 'button', {
-			name: 'Display settings',
-		} );
-		await displaySettings.click();
-		const itemsPerPageInput = page.getByLabel( 'Items per Page' );
-		// Make sure we're where we think we are
-		await expect( itemsPerPageInput ).toBeFocused();
-		// Type inside the dropdown's input
-		await page.keyboard.type( '00' );
-		// The input should still be visible.
-		await expect( itemsPerPageInput ).toBeVisible();
-	} );
 } );
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -10,6 +10,7 @@ import {
 	Notice,
 	__experimentalToolsPanel as ToolsPanel,
 	__experimentalToolsPanelItem as ToolsPanelItem,
+	__experimentalNumberControl as NumberControl,
 } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { privateApis as blockEditorPrivateApis } from '@wordpress/block-editor';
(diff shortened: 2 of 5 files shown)


test/e2e/specs/editor/various/is-typing.spec.js
```diff
@@ -36,36 +36,4 @@ test.describe( 'isTyping', () => {
 		// Toolbar Popover is hidden again
 		await expect( blockToolbarPopover ).toBeHidden();
 	} );
-
-	test( 'should not close the dropdown when typing in it', async ( {
```
ntsekouras: @ellatrix do you think we should update this test to something like typing in the `link` dialog or something instead of removing it?

ntsekouras (commented):
Thanks for the PR! Code wise this looks good. It would be good to have some design input though, especially about the moved controls in inspector controls.  --cc @WordPress/gutenberg-design 

Additionally, if the [change](https://github.com/WordPress/gutenberg/pull/63020/files#diff-35311dc3d34ca7575b2aabdc330138890fcd42f86ee21d23f8405ff5dc9e486cR317)  of icon in DataViews requires more discussion for some reason, we can always split this PR and land the Query Loop changes first.

t-hamano (commented):
> Moves the Query Loop block 'Display settings' to the inspector.

Refactoring the Query Loop block controls feels like a secondary effort to avoid using icons.

#58207 has already been submitted for refactoring the Query Loop block controls, so can we just discuss the DataViews icon in this PR?

jameskoster:
> can we just discuss the DataViews icon in this PR?

+1.

I actually have it on my list to update the view options icon in data views (also to split the layout controls into a separate menu, but that's another story). The cog is what I had in mind, so this works well for me 👍

annezazu:
>https://github.com/WordPress/gutenberg/pull/58207 has already been submitted for refactoring the Query Loop block controls, so can we just discuss the DataViews icon in this PR?

+1 to this too. I'd rather see that previous PR picked up again to ensure it aligns design wise.

afercia:
Thanks for pointing me at https://github.com/WordPress/gutenberg/pull/58207

One of th problems in this project is PRs without an associated issue (which is _required_ accordingly to the contributign guidelines) don't help contributors to follow all the fast-moving parts in this project. 
@youknowriad https://github.com/WordPress/gutenberg/pull/58207 is actually one more PR without assoicated issue, following our conversation at WCEU.

On the specifics od the design proposed in https://github.com/WordPress/gutenberg/pull/58207 I will comment on that PR but I'm not sure I like hte idea of burying down into an ellispis menu such important controls.

afercia:
> Additionally, if the [change](https://github.com/WordPress/gutenberg/pull/63020/files#diff-35311dc3d34ca7575b2aabdc330138890fcd42f86ee21d23f8405ff5dc9e486cR317) of icon in DataViews requires more discussion for some reason [...]

I perceived [this comment on the issue](https://github.com/WordPress/gutenberg/issues/62873#issuecomment-2199854377) as a green light to use the `cog` icon.

priethor:
> One of th problems in this project is PRs without an associated issue (which is required accordingly to the contributign guidelines) don't help contributors to follow all the fast-moving parts in this project.

I agree for the most part, but there are some exceptional cases where a simple PR following up on another one shouldn't need an extra task; that would actually prevent the project from moving fast and create noise. I would be happy to update the contributing guidelines to reflect this point.

priethor:
> I would be happy to update the contributing guidelines to reflect this point.

Actually, this is not required, as the guidelines specify that only `Non-trivial pull requests should be preceded by a related issue.`

afercia:
> > I would be happy to update the contributing guidelines to reflect this point.
> 
> Actually, this is not required, as the guidelines specify that only `Non-trivial pull requests should be preceded by a related issue.`

I know. That's why I stated that creating an issue _is required accordingly to the contributing guidelines of this project_ (for non-trivial changes). A while ago I also researched the history of the project to read the conversation and arguments made when the guidelines were introduced in the documentation and I'd totally agree with those argumentations.

> there are some exceptional cases where a simple PR following up on another one shouldn't need an extra task

Sure, in that case the PR could just have a short notation 'Follow up to' with a link to the relevant issue.

afercia:
I reverted the changes to the Query loop block and fixed the typo `Max page to show` -> `Max pages to show`.

I'll try to help https://github.com/WordPress/gutenberg/pull/58207 move on.

test/e2e/specs/editor/various/is-typing.spec.js
ntsekouras: This test should be kept for sure now, that you have reverted the Query Loop changes..

ntsekouras (approved):
We need to add back the removed test and this looks good to land. Thanks!

afercia:
Oh yes thanks. I also forgot to remove the unnecessary `BaseControl` wrapping the `NumberControl` that prevent aria-describedby to work as expected.

afercia:
Added back the missing test and removed the redundant `BaseControl` components. The only minor visual change is the slightly smaller bottom margin in teh popover after the description. Screenshot before and after:

before after last

I will update the testing instructions.

t-hamano (approved):

Issue #59745, opened 2024-03-11
DataForm component


youknowriad:
This document keeps track of all the work and tasks to build a powerful DataForm component capable of rendering all kind of forms and panels used to edit data. Document inspector in the editor, quick edit panels, settings forms...

DataForm is based on the same Fields abstraction that powers the [DataViews component](#55083) and support grouping fields, layouts, label positioning, conditional rendering, validation.

<img width="544" height="489" alt="Image" src="https://github.com/user-attachments/assets/85d4ad54-f1ac-4e67-a848-aff6b277b3f3" />

## Tasks

- [x] https://github.com/WordPress/gutenberg/pull/63032
- [x] https://github.com/WordPress/gutenberg/pull/63600
- [x] https://github.com/WordPress/gutenberg/pull/63840
- [x] https://github.com/WordPress/gutenberg/pull/63895
- [x] https://github.com/WordPress/gutenberg/pull/63983
- [x] https://github.com/WordPress/gutenberg/pull/64164
- [x] https://github.com/WordPress/gutenberg/pull/64168
- [x] https://github.com/WordPress/gutenberg/issues/64519
- [x] https://github.com/WordPress/gutenberg/issues/66338
- [x] Layout flexibility https://github.com/WordPress/gutenberg/pull/66531

## Next Steps

- [x] Implement Form Validation tools
- [ ] Strong conditional rendering
- [x] Grouping fields in cards, tabs, field sets (needs design)
- [ ] https://github.com/WordPress/gutenberg/issues/65685

## Low priority

- [ ] https://github.com/WordPress/gutenberg/issues/66616
- [ ] https://github.com/WordPress/gutenberg/issues/66336
- [ ] https://github.com/WordPress/gutenberg/issues/66332

joshuatf:
@youknowriad @oandregal Out of curiosity, will the DataForm ever allow control over layout within a form (columns, tabs, sections, etc) or would those be comprised of multiple DataForms?

youknowriad:
@joshuatf My current thinking is that "columns" and "tabs" are something you should be able to control on the `form` object. It's not clear to me at this point if a single "form type" can absorb everything or if we'd need several "form layouts" or "form types" like we have for data views, but I think we should be able do say, I want to group these fields in a tab...

jameskoster:
Noting this is closely related to https://github.com/WordPress/gutenberg/issues/63624.

oandregal:
Added a new section in the issue for "Task/Log".

[...]

Issue #55101, opened 2023-10-05
Data views: quick-editing


jameskoster:
### Table and Grid layout
In these layouts, we can use the same UI as the full-screen editor to toggle the visibility of the Inspector panel:

https://github.com/user-attachments/assets/66cfe71c-7906-4614-8e3e-5bd3defbd8eb

### List layout
In List layout it feels important that the preview remains visible. Instead of exposing a dedicated panel, the Inspector contents are loaded into the content frame via drilldown pattern:

https://github.com/user-attachments/assets/81516f30-1160-4683-b9ad-db9a744367f8

---

This feature would be equivalent to the quick edit experience in wp-admin:

<img width="1258" alt="Screenshot 2024-04-24 at 15 50 25" src="https://github.com/WordPress/gutenberg/assets/846565/49c863a4-2e07-4f88-af34-c28298643496">

<details><summary>Original issue</summary>Inline editing has considerable overlap with quick-actions (https://github.com/WordPress/gutenberg/issues/55096). For example most quick-actions for pages (https://github.com/WordPress/gutenberg/issues/55628) could also be accessed by clicking on values in the table directly. So for the initial implementation we might reuse where possible. Hopefully this mockup captures that:

<img src="https://user-images.githubusercontent.com/846565/279426851-3d473c16-8dbd-40a8-917f-56a8ee768fa5.png">

Incidentally the "Author" menu in the above concept should be virtually the same as the one that has already been implemented for filtering, so hopefully some legwork is already done 🤞 :

<img width="520" alt="Screenshot 2023-11-17 at 14 00 21" src="https://github.com/WordPress/gutenberg/assets/846565/d5b87906-772f-4c76-a8d7-4cfc847e2b52">

One exception we might make straight away is boolean values, e.g. whether comments are enabled or not. In this case the cell can probably contain a Toggle.</details>

jameskoster:
Noting that the work in https://github.com/WordPress/gutenberg/issues/59689 can unlock using the Inspector as a bulk/quick edit interface:

![Bulk edit](https://github.com/WordPress/gutenberg/assets/846565/58e754e2-8ac1-4ca2-b913-d34b93679ae3)

seifeldinio:
I've explored this solution for quick-editing within the inspector panel, here's the prototype for it:

https://github.com/user-attachments/assets/fc1b52a3-35e7-44db-931d-5d8fc8601754


Mockups:
### Single Page Selection

[...]

PR #63032, opened 2024-07-01
DataForm: implement first prototype using duplicate page action

oandregal:

## What?

First step in implementing the `DataForm` API. It's only used by the duplicate action from pages, which is only available in the plugin.

Captura de ecrã 2024-07-01, às 18 35 34

## Why?

We want to start exploring this API for several aspects of our flows. The first step is in replacing the "Duplicate item" action for pages because:

- That action is only available in the plugin, providing a nice place to hide this
- After merging, there's an organic path for expanding this API:
  - Rename actions for page and custom patterns. It's the same as this one.
  - Duplicate action for parts. It requires a new component (list for area).
  - Duplicate action for patterns. It requires adding two new components (boolean for synced, labels for categories).
  - etc.

| Page | Template part  | Pattern |
| ----- | --------------  | -------- |
| Captura de ecrã 2024-07-01, às 18 35 34 | 240701-duplicate-action-parts | 240701-duplicate-action-patterns |

## How?

- Creates a `DataForm` component that temporarily lives within the `@wordpress/dataviews` until we find a better place.
- Refactors the duplicate action for pages to use the new component.

## Testing Instructions

- Go to "Site Editor > Pages". In any page, execute the "Duplicate action" and verify it works.
- Edit a page using any editor (Post or Site). Execute the "Duplicate action" and verify it works.

## Follow-ups

- Extract field definition to a common place (registry?) that can be accessed by the actions & the pages components. https://github.com/WordPress/gutenberg/pull/63849
- Expand DataForm to other actions (see example above).
- Storybook story. https://github.com/WordPress/gutenberg/pull/63840


packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,93 @@
...
+	form: {
+		closeForm: () => void;
+		isBusy: boolean;
+		onSubmitLabel?: string;
+		onCancelLabel?: string;
+	};
+};
+
+export default function DataForm< Item >( {
+	item: data,
```
oandregal: According to https://github.com/WordPress/gutenberg/issues/55101#issuecomment-2058754635 there're use cases when `DataForm` could take an array of items as input.

oandregal:
This is what I got today. There're a few TODOs over the code that I'm going to look at in the coming days. Tomorrow, I'll focus on clarifying the last bit to have a generic `DataForm` (the internal `onChange` method), and then address the other details.

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+	// TODO: make it work for all fields.
+	const onChange = useCallback( ( title: string ) => {
+		setItem( { ...item, title } );
+	}, [] );
+
+	const components = useMemo(
+		() =>
+			fields.map( ( field ) => {
+				if ( field.type === 'text' ) {
```
oandregal: Alternatively, instead of providing a `type` the field could offer an `edit` view to provide its own control. I found this to be an advance use case, so I decided to start with the `type`.

This is also a reason why starting with the actions is a good way to scope down the PR. The form rendered here needs to provide an input component, which we don't have in the Fields API yet — the `render` is only used in DataViews for now. When we get there, we need to consider whether the `render` can be shared between `DataViews` and `DataForms`. We already have situations where each type of view in DataViews needs different things, so it's not unthinkable that DataForms would need its own. I thought it'd be better to explore that problem space in a separate PR.

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Internal dependencies
+ */
+import type { NormalizedField } from './types';
+
+type DataFormProps< Item > = {
+	item: Item;
+	onUpdateItem: ( item: Item ) => void;
+	fields: NormalizedField< Item >[]; // TODO: use Field. Normalize them first.
+	form: {
```
fabiankaegy: Would `metadata` be a better name for this? 

`form` seems to imply this is contains the actual form fields.

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,94 @@
...
+} from '@wordpress/components';
+import { useCallback, useMemo, useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import type { NormalizedField } from './types';
+
+type DataFormProps< Item > = {
+	item: Item;
```
fabiankaegy: In the context of a `form` `item` seems odd 🤔 

Would `values` or something maybe make more sense conceptually?

packages/dataviews/src/dataform.tsx
oandregal: As per https://github.com/WordPress/gutenberg/issues/59745, I think we want to call it form, like we did in DataViews (e.g.: `<DataViews view={view} />`, `<DataForm form={form} />`).

packages/dataviews/src/dataform.tsx
oandregal: Updated it to `data` at https://github.com/WordPress/gutenberg/pull/63032/commits/1328850a167e48d2c19e69f22df39e652a27d778 to be inline with `<DataViews data={data} />`.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -39,6 +39,17 @@ import { CreateTemplatePartModalContents } from '../create-template-part-modal';
 const { PATTERN_TYPES, CreatePatternModalContents, useDuplicatePatternProps } =
 	unlock( patternsPrivateApis );
 
+// TODO: this should be shared with other components (page-pages).
```
oandregal, line 42: Consolidation can be done in a follow-up PR, but I wanted to gather early thoughts: do we create a registry for fields per postType? where do we extract the Field information for now? Actions are defined in the `editor` package while the `edit-site` package defines the DataViews.

youknowriad: Yeah, I think right now the registry should be in the. `editor/src/dataviews/fields` folder like we have `editor/src/dataviews/actions` folder if we're talking about the default WP entities fields.

My thinking is that I'll probably move `editor/src/dataviews` to either a dedicated package `@wordpress/core-dataviews` or just move it into `@wordpress/core-data` because it's actually lower level than the editor and can be used outside the editor.

packages/dataviews/src/dataform.tsx
youknowriad: Actually for me the `form` object as I imagined it is less about these event handlers `onChange`, `onSubmit`...  and more about the "layout" of the form, which fields are visible, how they are groups...

Seeing, what's currently in the form object, It raised the question about whether all the submit button and form submission handling should actually be part of `DataForm` or not.

I'm leaning towards no for the moment, for instance in the "details panel" case, there's no submit button. So I see `DataForms` as something you actually can use within a form but not the whole form. I see it as a group of `*Control` components basically. If we simplify, I see it as a component that has a value (data here) and a single `onChange` handler, all the rest is left to the wrapper component (submit...)

oandregal: I iterated on this and moved back the submit/cancel actions to the consumer.

I agree `form` is about the layout. If we wanted to absorb the form actions (submit/cancel) plus the corresponding labels, a better API could be:

```js
<DataForm
  data={ data }
  fields={fields}
  form={ form }
  onUpdate={ onUpdate }
  /* Absorb submit/cancel */
  labels={ labels }
  onSubmit={ onSubmit }
  onCancel={ onCancel }
  />
```

For an use case like the details panel, the interaction model suggested `onUpdate` would be triggered for every change. Though for modals like in this PR's use case, I went back and forth about triggering the update for every change vs only doing it upon submitting. The former seems a simpler approach, I didn't like we had to absorb the `event` handler for forms.

oandregal:
I want to look at improving a couple of TypeScript types (see TODOs), but other than that it's working and API ready. What this PR introduces:

```js
const form = {
	visibleFields: [ ... ],
};

<DataForm
  data={ data }
  fields={ fields }
  form={ form }
  onUpdate={ onUpdate }
  />
```

This PR assumes the `DataForm` receives a single item, but there's an use case ([bulk editing](https://github.com/WordPress/gutenberg/issues/55101#issuecomment-2058754635)) where it may receive multiple: this is best iterated when that's implemented.

In follow-up PRs we can expand the usage of `DataForm` to:

- Rename actions for page and custom patterns. It's the same as this one.
- Duplicate action for parts. It requires a new type/component (list for area).
- Duplicate action for patterns. It requires adding two new components (boolean for synced, labels for categories).

packages/dataviews/src/dataform.tsx
youknowriad: Makes me wonder whether we should always pass a single item though. Even though we're batch editing a list of items, we're still showing a single value per field. I wonder if there's a use-case where we show the multiple values somehow. Anyway, not that important for now.

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,83 @@
...
+	const [ value, setValue, debouncedValue ] = useDebouncedInput(
+		field.getValue( { item: data } )
+	);
+
+	useEffect( () => {
+		onUpdate( ( prevItem: any ) => ( {
+			...prevItem,
+			[ id ]: debouncedValue,
+		} ) );
+	}, [ debouncedValue ] );
```
youknowriad: Curious about your thoughts on debouncing. Why did you choose to debounce the onChange calls within the form?

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,83 @@
...
+			normalizeFields(
+				fields.filter(
+					( { id } ) => !! form.visibleFields?.includes( id )
+				)
+			),
+		[ fields, form.visibleFields ]
+	);
+
+	return visibleFields.map( ( field ) => {
+		if ( field.type === 'text' ) {
```
youknowriad: It would be cool to model a "registry" (even if closed to start with) of field types.

```
const text = {
   name: 'text',
   edit: DataFormTextControl
}
```

This can materialize in follow-ups as we start expanding.

youknowriad (commented):
One follow-up could be to add a storybook story.

packages/dataviews/src/dataform.tsx
oandregal: My rationale was to avoid calling the consumer's `onUpdate` for every character: this way the consumer `onUpdate` is only called after the user has finished (or after a few characters). Does that make sense?

youknowriad: I wonder how generic is that. Wouldn't we have forms that require immediate feedback. I'm wondering if something like this should be part of the form itself or the consumer just debounces the onChange callback itself? Or should it be a per field config.

oandregal: That's a valid point. I haven't seen any use case that does that, which doesn't mean it doesn't exist. Removed the debouncing at https://github.com/WordPress/gutenberg/pull/63032/commits/2e04938631822e1abec151e49924dfa2ab5ef502

We can iterate on this in follow-ups should be needed.

packages/dataviews/src/dataform.tsx
oandregal: Pushed changes to use a registry-like structure at https://github.com/WordPress/gutenberg/pull/63032/commits/a29032179b200b28ef698b1a25989ff79117422b

oandregal:
There are [mobile test failures](https://github.com/WordPress/gutenberg/actions/runs/9773465094/job/26979690720?pr=63032) that don't seem related to this PR:

> FAIL packages/block-library/src/code/test/edit.native.js
>  ● Test suite failed to run
>
>    TypeError: Cannot use 'in' operator to search for 'WebkitAnimation' in undefined
>
>      3 |  */
>      4 | // eslint-disable-next-line no-restricted-imports
>      5 | import * as Ariakit from '@ariakit/react';

oandregal:
All feedback was addressed, the PR is working, and it's a good first step in terms of API & direction, in my view. Thoughts on landing this @youknowriad @jorgefilipecosta @ntsekouras?

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * Internal dependencies
+ */
+import type { Form, Field, NormalizedField } from './types';
+import { normalizeFields } from './normalize-fields';
+
+type DataFormProps< Item > = {
+	data: Item;
+	fields: Field< Item >[];
+	form: Form;
+	onUpdate: Dispatch< SetStateAction< Item > >;
```
youknowriad: In DataViews, we have `onChangeView` and `onSelectionChange` and here we have `onUpdate`

I think I'd love for these to be:

 - onChangeView
 - onChangeSelection
 - onChange (no suffix here to match Controls and because it's slightly different in terms of semantic, we're changing the data)

I'm also wondering whether `data` should be just `value` like the Controls component but that's more arguable.

packages/dataviews/src/dataform.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+type DataFormControlProps< Item > = {
+	data: Item;
+	field: NormalizedField< Item >;
+	onUpdate: Dispatch< SetStateAction< Item > >;
+};
+
+function DataFormTextControl< Item >( {
```
youknowriad, line 32: Would be cool to have a "types" folder with a file per type maybe at some point.

youknowriad (commented):
I didn't test properly yet but definitely good to land for me. Great start.

packages/dataviews/src/dataform.tsx
oandregal: Renamed `onUpdate` to `onChange` at https://github.com/WordPress/gutenberg/pull/63032/commits/374702023d68c87eee292d4134af19b622728938

`data` is in line with `DataViews` and also at some point we should work with multiple items, so I lean towards keeping it.

oandregal: https://github.com/WordPress/gutenberg/pull/63087 renames `onSelectionChange` to `onChangeSelection` for dataviews.

oandregal:
This is currently blocked by [mobile test failures](https://github.com/WordPress/gutenberg/actions/runs/9777079855/job/26992055758?pr=63032) that I don't see related to this PR:

> FAIL packages/block-library/src/code/test/edit.native.js
>  ● Test suite failed to run
>
>    TypeError: Cannot use 'in' operator to search for 'WebkitAnimation' in undefined
>
>      3 |  */
>      4 | // eslint-disable-next-line no-restricted-imports
>      5 | import * as Ariakit from '@ariakit/react';

Asked for help in the mobile slack channel ([thread](https://wordpress.slack.com/archives/C02RQC4LY/p1720006970831539)) to clarify what is happening.

oandregal:
I've pushed https://github.com/WordPress/gutenberg/pull/63032/commits/969ab6cf3d9f890b7d11b22f141d1dde7951f346 as per @geriux suggestion in the mobile slack channel, which fixes the issue and unblocks this PR from landing.

For full disclosure and future reference, this is what we know about what happened:

- The react native editor uses some parts of the `@wordpress/editor` package (it was importing the `usePostActions` hooks).
- When this kind of errors have happened in the past, it was due to not having a full DOM implementation in some of the packages (see, for example, this [bug](https://github.com/facebook/react/issues/25920) with `jsdom-jscore-rn`).
- This PR adds `@wordpress/dataviews` as a new dependency to the `@wordpress/editor` package.
- Apparently, the error was triggered by some Ariakit import (a transitive dependency of dataviews via `@wordpress/component`). However, components is already a dependency of the editor and the `TextControl` component used by `DataForms` is also already used by the editor, so it may be a red herring. Or perhaps the parts of the editor used by native do not depend on components. To be investigated separately.

youknowriad (approved):

youknowriad:
Should we add something about the new component to the changelog file?

oandregal:
> Should we add something about the new component to the changelog file?

To the dataviews' changelog? I'm not sure. If this is supposed to become its own standalone package, I wouldn't add anything. I'd add it if this is a more permanent place where `DataForm` would live.

youknowriad:
I'm not sure this one should be in a separate package. I think DataViews and DataForms will ultimately share a lot of things (field types ...) so I see this as a single package personally to render data in different shapes.

oandregal:
Pushed a changelog entry at https://github.com/WordPress/gutenberg/pull/63032/commits/8e0661525997ec9d03b4a8c4ea0d3dd450059bed

packages/dataviews/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Added a new `DataForm` component to render controls from a given configuration (fields, form), and data.
```
youknowriad, line 7: Maybe add a `### New Features` heading?

oandregal: https://github.com/WordPress/gutenberg/pull/63032/commits/c188d112e6506d5fb6b5482acba31fd4e5053ace

oandregal:
> Extract field definition to a common place (registry?) that can be accessed by the actions & the pages components.

I gave this a shot at https://github.com/WordPress/gutenberg/pull/63849

oandregal:
> Storybook story.

Riad started working on this at https://github.com/WordPress/gutenberg/pull/63840

PR #63034, opened 2024-07-01
Docgen: Fix function param for const function expression

sirreal:
## What?

Fix an issue where docgen does not get the parameter types of a function expression.
Add test for the issue.

## Why?

It failed with named exports like this:

```ts
/** @param p */
const func = function ( p: string ): string { return p; }
```

This was an issue in #63012 where docgen failed on data-controls types.

Normally this would either be an exported function or a variable assigned to an arrow expression. Apparently this case was uncovered.

## How?

Fix docgen to correctly handle this.

## Testing Instructions
CI passes.

### DIFF
--- a/packages/docgen/CHANGELOG.md
+++ b/packages/docgen/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


cbravobernal (approved):
LGTM

PR #34522, opened 2021-09-03
Block Styles: Show style preview when hovered or focused 

ramonjd:
Resolves https://github.com/WordPress/gutenberg/issues/33933

## Description

Hark! This PR attempts the following based on the explorations in https://github.com/WordPress/gutenberg/issues/33933:

1. Updates the styles switcher in the block inspector to use buttons.
2. Displays previews in a dedicated Slot when hovering over styles buttons in the block inspector. 
3. Updates the style switcher in the block toolbar to a menu with check icons.
4. Hides the default styles picker dropdown _unless_ a non-default style has been previously saved. This is because the selecting default styles will one day be moved to Global Styles. See [comments below](https://github.com/WordPress/gutenberg/pull/34522#issuecomment-929008566).

## Testing

Before switching to this branch, head to the Block Editor and insert a block that has block style variations. The Image Block under the TT1 theme is a satisfactory choice.

But, if you get time, also check other blocks, e.g., Image, Buttons, Social Links, Table and Quote blocks, in several base themes too. <sup>Thank you!</sup>

Select anything other than "Default" as the default style in the dropdown. For example, select "Rounded".

Save the post and checkout this branch.

Refresh the post page and select the Image Block you have just inserted. 

In the Styles Panel in the block inspector, the block previews should appear when you hover or focus over the block style variation buttons in the inspector.

https://user-images.githubusercontent.com/6458278/140431090-c9fe3ba9-85d3-47c8-bbd8-a2f0d5bb736b.mp4

The preview should hide itself when the mouse (or focus) leaves the corresponding button. It should be positioned in the top right corner. Try with a mouse and tabbing keys for extra points! 💯 

Clicking or hitting `SPACE` or `ENTER` keys should activate the style for the selected block, and make the preview... disappear! 🪄  

Ensure that the previews do not show in narrow viewports.

Screen Shot 2021-09-21 at 12 00 26 pm

Switch styles using the "Change block type or style" menu in the block toolbar.

Screen Shot 2021-09-20 at 2 05 10 pm

The styles should change as you'd expect. The selected style should be marked with a check icon.

Inserting subsequent Image Blocks should display your chosen default style.

Using the default style picker dropdown, selecting either the "Default" or "Not set" values should set the default style and cause the dropdown picker to hide itself.

Now when Inserting subsequent Image Blocks, you won't see the default style picker.

Using your console skillz add some extra text to one of the style buttons. If the text overflows, you should see an ellipsis:

| Narrow width  | Wide width |
[...]

PR #63039, opened 2024-07-02
Block Styles: Remove preview from inspector panel

### LINKED ISSUE #63035: Section styles: improve (or remove) preview
Currently, the previews for block style variations don't offer much value when applied to styling sections of blocks with the new feature in 6.6. If anything, the preview looks broken and doesn't match all of the changes one can expect to see:

Screenshot 2024-07-01 at 11 34 44 AM

Here's a video:

https://github.com/WordPress/gutenberg/assets/26996883/53f6acc1-7f6a-408f-ae71-dc97a81d3186

As a result, I think we should either remove these previews for Groups & Columns for now or consider a way to improve the preview offered cc @aaronrobertshaw

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/63035

## What?

Removes preview popover for block styles in the block inspector sidebar.

## Why?

A block style's preview is generated from the selected block type's example content. This results in some less-than-helpful previews that are also disconnected from the selected block. 

This PR trials removing these previews entirely given their limited value and the ease at which a user can toggle between the available block styles to see them in action and in the real context of their page.

## How?

- Remove the preview popover component
- Remove BlockStyles component's on hover handling

## Testing Instructions

1. Add a block that has block styles defined for it e.g. Button block, or a new section style
2. Select the block and open the Styles tab
3. Hover over the available styles and confirm no preview is shown
4. Ensure that selecting block styles still works and they are reflected in the editor and saved correctly

## Screenshots or screencast 

| Before | After |
|---|---|
| Screenshot 2024-07-02 at 4 34 05 PM | Screenshot 2024-07-02 at 4 29 07 PM |


aaronrobertshaw:
I've left this PR in draft status for the moment until I can confirm whether it makes sense to still remove the previews from the native editor too.

aaronrobertshaw:
The native block styles UI still uses the inline preview approach and wasn't updated to the web's popover preview in https://github.com/WordPress/gutenberg/pull/34522. I think it makes sense to address the native UI in a dedicated follow-up.

Screenshot 2024-07-02 at 5 25 22 PM

ramonjd (approved):
LGTM - if in doubt, chuck it out. 

## Before

https://github.com/WordPress/gutenberg/assets/6458278/2b6f5719-57f9-490f-a889-744d7cf72ba5


## After

https://github.com/WordPress/gutenberg/assets/6458278/97936cdc-867b-4bb6-9fd6-45ae59141896

mtias:
I think we shouldn't remove the preview but instead improve the contents of the group example source—add a nicer layout with a heading, text, maybe an image or button, that better represents groups as layout units.

richtabor:
Thanks for this. Let's also try an alternate approach, to see if we can get a bit closer to the request. 

>  If anything, the preview looks broken and doesn't match all of the changes one can expect to see:

Perhaps something as simple as updating the group block's `example` would do the trick. In the future it'd be awesome if the actual block from the content was previewed, but for the interim, this may do well enough.

aaronrobertshaw:
Thanks for the reviews and feedback 👍 

I'll spin up a fresh PR to improve the existing group block example. 

> In the future it'd be awesome if the actual block from the content was previewed, but for the interim, this may do well enough.

It's pretty straightforward to get the actual block into the preview. The catch is constraining its size, layout changes etc. If the preview is scaled, we might run into issues where the value of the preview is diminished because it's all scaled too small.

Screenshot 2024-07-04 at 11 48 27 AM

Here's a quick demo after hacking around a little:

https://github.com/WordPress/gutenberg/assets/60436221/fddbb1fd-a6bc-46a6-a186-461f7fa77b1c

aaronrobertshaw:
I have an [alternate PR](https://github.com/WordPress/gutenberg/pull/63114) (draft only at this stage) which updates the Group block example to include a heading, paragraph, image, and button.

Previously, all the Group block variations (Row, Stack, and Grid) share the same example. In the new PR, I've kept the old example content for those variations. 

If have any suggestions on exactly what content each variation should display, please drop a comment over on https://github.com/WordPress/gutenberg/pull/63114 and I'll tweak it?

aaronrobertshaw:
Closing this in favour of https://github.com/WordPress/gutenberg/pull/63114

richtabor:
> If the preview is scaled, we might run into issues where the value of the preview is diminished because it's all scaled too small.

Yea, we'd want to leverage viewportSize to maintain a "desktop" view for the preview. As well as allowing for a max height of the preview (though perhaps it's scrollable vertically).

PR #63042, opened 2024-07-02
DataViews: Removing mapping of user patterns to temporary object

youknowriad:
Similar to #62927 but for patterns
Related https://github.com/WordPress/gutenberg/issues/59659

## What?

When working on a the dataviews extensibility actions. I noticed that some of the post type actions receive objects that are not actual post types, they are temporary objects that were created for the purpose of rendering template parts in the patterns page. We were mapping user patterns to weird objects that resemble theme patterns.

For me this is not something we should be doing because:

 - We'll be creating temporary objects whose type is very unclear.
 - the actions are supposed to be generic and used both in dataviews and also in the editor. Which means that sometimes they get the normal pattern object. 

## How?

This PR just removes that mapping and tries to update all the places where that mapping was necessary. The code is a bit hard to follow, so there's a chance that I missed some places. Some help testing everything is needed.

## Testing Instructions

1- Use the different filters, fields and actions in the patterns dataviews (and pattern editor).

### DIFF
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -117,6 +117,7 @@ function PreviewWrapper( { item, onClick, ariaDescribedBy, children } ) {
 
 function Preview( { item, viewType } ) {
 	const descriptionId = useId();
+	const description = item.description || item?.excerpt?.raw;
 	const isUserPattern = item.type === PATTERN_TYPES.user;
 	const isTemplatePart = item.type === TEMPLATE_PART_POST_TYPE;
 	const [ backgroundColor ] = useGlobalStyle( 'color.background' );
@@ -143,7 +144,7 @@ function Preview( { item, viewType } ) {
 			<PreviewWrapper
 				item={ item }
 				onClick={ onClick }
-				ariaDescribedBy={ item.description ? descriptionId : undefined }
+				ariaDescribedBy={ !! description ? descriptionId : undefined }
 			>
 				{ isEmpty && isTemplatePart && __( 'Empty template part' ) }
 				{ isEmpty && ! isTemplatePart && __( 'Empty pattern' ) }
@@ -156,9 +157,9 @@ function Preview( { item, viewType } ) {
 					</Async>
 				) }
 			</PreviewWrapper>
-			{ item.description && (
+			{ !! description && (
 				<div hidden id={ descriptionId }>
-					{ item.description }
+					{ description }
 				</div>
 			) }
 		</div>
@@ -222,7 +223,7 @@ function Title( { item } ) {
 						// See https://github.com/WordPress/gutenberg/pull/51898#discussion_r1243399243.
 						tabIndex="-1"
 					>
-						{ title || item.name }
+						{ title }
 					</Button>
 				) }
 			</Flex>
@@ -300,6 +301,8 @@ export default function DataviewsPatterns() {
 				header: __( 'Sync status' ),
 				id: 'sync-status',
 				render: ( { item } ) => {
+					const syncStatus =
+						item.wp_pattern_sync_status || PATTERN_SYNC_TYPES.full;
 					// User patterns can have their sync statuses checked directly.
 					// Non-user patterns are all unsynced for the time being.
 					return (
@@ -309,8 +312,7 @@ export default function DataviewsPatterns() {
 							{
 								(
 									SYNC_FILTERS.find(
-										( { value } ) =>
-											value === item.syncStatus
+										( { value } ) => value === syncStatus
 									) ||
 									SYNC_FILTERS.find(
 										( { value } ) =>
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -302,23 +302,19 @@ export default function DataviewsPatterns() {
 				id: 'sync-status',
 				render: ( { item } ) => {
 					const syncStatus =
-						item.wp_pattern_sync_status || PATTERN_SYNC_TYPES.full;
+						'wp_pattern_sync_status' in item
+							? item.wp_pattern_sync_status ||
+							  PATTERN_SYNC_TYPES.full
+							: PATTERN_SYNC_TYPES.unsynced;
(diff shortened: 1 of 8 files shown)


youknowriad:
I noticed that "uncategorized" patterns page is broken (it shows user pattern with categories) but I think it's broken on trunk too.

ntsekouras:
>I noticed that "uncategorized" patterns page is broken (it shows user pattern with categories) but I think it's broken on trunk too.

~~It seems it's in trunk too this way. I'm wondering if it was introduced when the other mapping was removed..~~

--edit:

I don't see this in trunk...

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -117,6 +117,7 @@ function PreviewWrapper( { item, onClick, ariaDescribedBy, children } ) {
 
 function Preview( { item, viewType } ) {
 	const descriptionId = useId();
+	const description = item.description || item?.excerpt?.raw;
```
ntsekouras, line 120: I know this was from before, but it surprized me to see `raw` and not `rendered` here..

youknowriad: Same :) I think there's some consistency work to be done on the REST API.

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -300,6 +301,8 @@ export default function DataviewsPatterns() {
 				header: __( 'Sync status' ),
 				id: 'sync-status',
 				render: ( { item } ) => {
+					const syncStatus =
+						item.wp_pattern_sync_status || PATTERN_SYNC_TYPES.full;
```
ntsekouras: It seems we need to check if it's a user pattern, [similarly](https://github.com/WordPress/gutenberg/pull/63042/files#diff-0f49c90dc39786ecdf45f374247e433f763d295d80bdaecda3bab11357e4d7fdR147) with the search.

youknowriad: If you see a few lines above, this field is only added for "user" patterns.

ntsekouras: Will check now, but the rendering of the label is wrong for patterns..

ntsekouras: I pushed a [fix](https://github.com/WordPress/gutenberg/pull/63042/commits/0d6d391f43d3830c5464127a10d1d7aaa15df703) for this, but feel free to revert/update if you have a better solution..

packages/edit-site/src/components/page-patterns/use-patterns.js
```diff
@@ -128,8 +128,11 @@ const selectPatterns = createSelector(
 			patterns: themePatterns,
 			isResolving: isResolvingThemePatterns,
 		} = selectThemePatterns( select );
-		const { patterns: userPatterns, isResolving: isResolvingUserPatterns } =
-			selectUserPatterns( select );
+		const {
+			patterns: userPatterns,
+			isResolving: isResolvingUserPatterns,
+			categories: userPatternCategories,
```
ntsekouras, line 134: This is not needed or used.

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -302,23 +302,19 @@ export default function DataviewsPatterns() {
 				id: 'sync-status',
 				render: ( { item } ) => {
 					const syncStatus =
-						item.wp_pattern_sync_status || PATTERN_SYNC_TYPES.full;
+						'wp_pattern_sync_status' in item
```
youknowriad, line 305: How can this condition be false if "type" is always user? (line 299)

packages/edit-site/src/components/page-patterns/use-patterns.js
youknowriad: It's used in the `hasCategory` callback down here.

youknowriad:
I fixed the uncategorized patterns issue.

packages/edit-site/src/components/page-patterns/use-patterns.js
ntsekouras: This is quite hard to follow.. okay.

ntsekouras:
>I fixed the uncategorized patterns issue.

Thanks. I had  a similar fix but hadn't pushed yet because I see one more problem. If we have created custom pattern categories, the sidebar doesn't render them. You can test that by creating a new pattern with a new category that didn't exist before.

youknowriad:
Good catch, it's fixed in the latest commit.

packages/edit-site/src/components/page-patterns/index.js
ntsekouras: Probably it's when we merge the `theme` patterns with the user ones? You can see the issue without this change in categories like `about`, etc..

youknowriad:
This should be ready right?

ntsekouras (approved):
I think this is good to land. Thank you Riad! Even though we tested a lot,  let's just keep an eye for possible bugs from here, since the code is quite complex..

youknowriad:
This is the kind of code that could benefit from being typescripted :P

packages/editor/src/components/post-actions/export-pattern-action.js
```diff
@@ -23,9 +24,9 @@ function getJsonFromItem( item ) {
 	return JSON.stringify(
 		{
 			__file: item.type,
-			title: item.title || item.name,
-			content: item.patternPost.content.raw,
-			syncStatus: item.patternPost.wp_pattern_sync_status,
+			title: getItemTitle( item ),
+			content: item.content.raw,
```
t-hamano, line 28: In the DataViews, pattern content is indeed present in `item.content.raw`. But when exported from the sidebar, the content is present in `item.content`, so there seems to be no content `field` in the exported JSON file.

However, pattern export from the sidebar doesn't seem to have worked before. See [this comment](https://github.com/WordPress/gutenberg/issues/63217#issuecomment-2212320713) for more details.

ramonjd: I have a small PR here:

- https://github.com/WordPress/gutenberg/pull/63227

PR #63043, opened 2024-07-02
Data views list layout: Re-organise actions display

jameskoster:
## What?
This PR attempts to address two issues in list layout.

The first is the awkward alignment of the title when all fields are toggled off:

Screenshot 2024-07-01 at 16 37 41

The second is the reflow that occurs when hovering/selecting a row that contains a field with a very long value, or just many fields:

https://github.com/WordPress/gutenberg/assets/846565/da6f4cf0-0aab-4044-b237-32d2a19b3b2a

## How?
Currently the title + fields are in one column, and the actions are in another. When the actions column appears, the title + fields column gets squished to accommodate it causing wrapping.

In this PR I've re-organised the markup so that the title and actions are in a row, with fields beneath. When the actions are revealed the title is squished, but truncated to avoid reflow.

To accomplish this I had to increase the height of the title slightly to align with the action icons. This also increased the height of the title + fields container in a two-row configuration by 4px, so the thumbnail was increased to match. This structure is better described in an image:

guides

The result is that regardless of how many fields are visible (or how long the field values are), alignment is maintained:

| Title + 1 row of fields | Title + several rows of fields | Title alone |
| --- | --- | --- |
| Screenshot 2024-07-02 at 10 39 26 | Screenshot 2024-07-02 at 10 44 09 | Screenshot 2024-07-02 at 10 45 02 |

## Testing Instructions
1. Navigate to a dataview like pages and switch to list layout.
2. Toggle fields on / off and ensure alignment is maintained, and that the two issues are fixed.
3. Ensure that keyboard navigation works as expected.


jasmussen:
This seems overall a great step forward. The only one that gives me pause is this one:



In this one, the actions on the right don't feel well-placed. 

Is there a way we can keep those right-most actions vertically centered according to the thumbnail? The description would need to not go to the right-most edge, but that seems an improvement regardless.

jameskoster:
I couldn't think of a way that doesn't result in other issues. Leaving the actions in their own column will result in the 'jump' issue described above.

I tried absolutely positioning the actions, but that causes them to overlap long titles and fields which seems undesirable to me.

We also need to consider that in the future there might be data view consumers that do not employ a thumbnail in List layout. That's why aligning to the _title_ makes a bit more sense to me.

A different avenue to explore might be tweaking the metrics a bit, to make this placement feel a bit more intentional. Perhaps increasing the spacing:

Screenshot 2024-07-02 at 16 38 35

Figma [here](https://www.figma.com/design/P3n31QJoIR84tGHearPQlc/Template-management?node-id=1097-6541&t=xPPpleBZBoWi6XeR-1) if you'd like to try some options.

jameskoster:
> Perhaps increasing the spacing

One argument for this approach is that it enables us to include icons in field values, e.g.

Screenshot 2024-07-02 at 19 20 44

oandregal:
Thanks for considering the keyboard navigation in your test instructions :) There're some [e2e tests](https://github.com/WordPress/gutenberg/blob/trunk/test/e2e/specs/site-editor/dataviews-list-layout-keyboard.spec.js) for that but it's still nice to do a manual one in case there's something not covered. I've also done it myself and it's working properly.

The changes improve the experience, and it's working as expected. I can take a deeper look from a code point of view when design direction is confirmed. I'll be AFK soon, so don't wait for my approval if I'm not around — my major concern here is that keyboard navigation still works fine.

jameskoster:
The author field will also soon display the users avatar (https://github.com/WordPress/gutenberg/pull/63142), which further indicates we should tweak the layout slightly. I'll work on an update soon.

jameskoster:
Connecting another dot; the status field may also be updated to include the icon: https://github.com/WordPress/gutenberg/issues/63199.

jameskoster:
Updated the metrics a bit:

Screenshot 2024-07-05 at 16 34 21

Here's how things look when support for icons in field values is added:

Screenshot 2024-07-05 at 16 35 29

This one is ready for a fresh review.

t-hamano:
Keyboard navigation still works correctly as before, and I don't think it would be difficult to resolve the e2e test failure.

I think the accessibility issue that needs to be resolved here is applying element roles properly.

### trunk

Elements in a row are divided into grid cells according to their role.

```
- grid
  -  row
    - gridcell
      - Button: Preview image, title, author, etc.
    - gridcell
      - button: Edit button
    - gridcell
      - button: Actions button
```

### This PR

There is only one grid cell in a row, and grid cells are nested.

```
- grid
  -  row
    - gridcell
      - button 
        - Preview image, title, author, etc.
        - gridcell
          - button: Edit button
        - gridcell
          - button: Actions button
```

Is there any way to improve the layout while still maintaining the original roles and hierarchy?

jameskoster:
Thank you for pointing that out. It demonstrates another issue on trunk that this PR fixes; clicking around the action buttons will not select the row, despite the hover style suggesting it would:

https://github.com/WordPress/gutenberg/assets/846565/05e3b283-d858-4d1a-9272-8a7b44910e76

That said, I'm not sure it's possible to retain the layout and avoid incorrect nesting of roles. This could suggest that absolutely positioning the actions is the way to go after all.

I'll probably spin up a different PR to try that.

jameskoster:
Closing due to #63299 being merged.

PR #63044, opened 2024-07-02
File block: Use native feature detection for pdf support

### LINKED ISSUE #58891: PDF embedding/preview is greedily blocked on Android Firefox/Chrome even though it should work
### Description

In an Android 13 device with Firefox and Chrome browsers, it is possible to have inline embed PDF be visible. But the present implementation based on discussions in https://github.com/WordPress/gutenberg/issues/46382 https://github.com/WordPress/gutenberg/issues/34825 etc makes it that if the User Agent includes "Mobi" or "Android" the pdf embed is hidden totally.[1]

[1]: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/file/utils/index.js#L7

This greedily blocks even browsers/devices that do support PDF

1. Go to Gutenberg editor
2. Insert a file block with preview turned on
3. Go to a mobile device
4. Do not see the preview

_No response_

### Environment info

Android 13 with wordpress.com blog

Yes

Yes

asdofindia:
Fixes #58891

Relies on https://developer.mozilla.org/en-US/docs/Web/API/Navigator/pdfViewerEnabled

## What?
This PR adds a reliable way to detect pdf viewer support.

## Why?
It fixes #58891 and deterministically looks for pdf viewer support.

## How?
We check whether the browser signals that pdf viewer is available and if so, immediately assumes that the browser supports pdf viewing (without relying on user agent detection). It falls back to the old behavior if this pdfViewerEnabled property is not available in navigator.

## Testing Instructions
1. Open a post or page
2. Insert a file block and choose a PDF
3. Save the post.
4. View from a mobile device

Same as above.

### DIFF
--- a/packages/block-library/src/file/utils/index.js
+++ b/packages/block-library/src/file/utils/index.js
@@ -5,6 +5,11 @@
  * @return {boolean} Whether or not the browser supports inline PDFs.
  */
 export const browserSupportsPdfs = () => {
+	// Use native feature detection if available.
+	if ( window.navigator.pdfViewerEnabled ) {
+		return true;
+	}
+
 	// Most mobile devices include "Mobi" in their UA.
 	if ( window.navigator.userAgent.indexOf( 'Mobi' ) > -1 ) {
 		return false;


packages/block-library/src/file/utils/index.js
```diff
@@ -5,6 +5,11 @@
  * @return {boolean} Whether or not the browser supports inline PDFs.
  */
 export const browserSupportsPdfs = () => {
+	// Use native feature detection if available
+	if ( window.navigator.pdfViewerEnabled ) {
+		return true;
+	}
+
```
SirLouen: There is an ESLint error in this line (some extra spaces you should remove)

SirLouen (dismissed):

SirLouen:
## Test Report
### Description
✅ This report validates whether the indicated patch works as expected.

Patch tested: https://patch-diff.githubusercontent.com/raw/WordPress/gutenberg/pull/63044.diff

### Environment
- WordPress: 6.9-alpha-60523
- PHP: 8.2.29
- Server: Apache/2.4.62 (Debian)
- Database: mysqli (Server: 11.8.2-MariaDB-ubu2404 / Client: mysqlnd 8.2.29)
- Browser: Firefox 141.0 (Mobile)
- OS: Android
- Theme: Twenty Seventeen 3.9
- MU Plugins: None activated
- Plugins:
  * Gutenberg 21.3.0
  * Test Reports 1.2.0

### Actual Results
1.  ✅ Issue resolved with patch.

### Additional Notes
- I've reviewed it and this patch looks good to me. [As I commented here](https://github.com/WordPress/gutenberg/issues/58891#issuecomment-3141576565), this is not a bug but an enhancement, because removing PDF support for mobiles was expected by default. 

- This support has been there in Firefox for a while, but it's been rolled now in Chrome Mobile browsers (my Chrome Mobile doesn't support this yet, but Firefox Mobiles does). 

cc @Mamaduka 

### Supplemental Artifacts
Before the patch
image

After the patch:
image

packages/block-library/src/file/utils/index.js
asdofindia: Thank you! I should have run it locally the first time around itself. Have fixed the same as can be seen in https://github.com/WordPress/gutenberg/actions/runs/16665804520/job/47171729775?pr=63044

asdofindia:
I've force pushed the whitespace change for fixing lint. The unit tests seem to fail for reasons outside this PR. Do let me know if anything else needs to be done.

Mamaduka:
Yes, it's an unrelated problem. This issue should be fixed once #71009 is merged, but will require another rebase.

Update: Merge the fix. Rebasing will resolve the problem now.

packages/block-library/src/file/utils/index.js
```diff
@@ -5,6 +5,11 @@
  * @return {boolean} Whether or not the browser supports inline PDFs.
  */
 export const browserSupportsPdfs = () => {
+	// Use native feature detection if available
```
SirLouen: Add a full-stop by the end of the comment

asdofindia: Added.

Mamaduka (approved):
Thanks, @asdofindia!

SirLouen (approved):

PR #63045, opened 2024-07-02
Components: Split stories and test files into separate projects

sirreal:
## What?

I did some TypeScript analysis and found that most TypeScript compilation time is spent on the components package. Some very costly compilation was in stories.

## Why?

I wanted to try to split out stories (and tests) for a few reasons:
- The use an environment that is not suitable for application code with jest, node, and other globals.
- There's no reason to emit anything (including declarations) for stories and tests. We can typecheck them, but nothing consumes the types so we can save the emit.

## How?

Split into a separate project.

## Testing Instructions
CI passes.
Look at some stories in an editor with TypeScript integration. Does everything seem OK?

### DIFF
--- a/packages/components/tsconfig.json
+++ b/packages/components/tsconfig.json
@@ -33,7 +33,7 @@
 		"src/**/*.native.js",
 		"src/**/react-native-*",
 
-		// Stories and tests are included in another project dedicated project.
+		// Stories and tests are included in another project.
 		"src/**/stories",
 		"src/**/test"
 	]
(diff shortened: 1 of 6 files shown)


packages/components/tsconfig.json
```diff
@@ -31,13 +26,15 @@
...
+	"include": [ "src" ],
 	"exclude": [
 		"src/**/*.android.js",
 		"src/**/*.ios.js",
 		"src/**/*.native.js",
 		"src/**/react-native-*",
-		"src/**/stories/**/*.js", // only exclude js files, tsx files should be checked
-		"src/**/test/**/*.js" // only exclude js files, ts{x} files should be checked
+
+		// Stories and tests are included in another project dedicated project.
```
sirreal:
```suggestion
		// Stories and tests are included in another project.
```

PR #63046, opened 2024-07-02
DataViews: Register the export pattern action like any third-party action

youknowriad:
Related #61084 
Similar to #62913 

## What?

The current PR explore the possibility to use the API to register one action: "export pattern". 

## Testing Instructions

1- Open the patterns dataviews
2- Click "My patterns"
3- On the actions menu per pattern, click "Export as JSON"

### DIFF
--- a/packages/editor/src/dataviews/actions/export-pattern.tsx
+++ b/packages/editor/src/dataviews/actions/export-pattern.tsx
@@ -34,12 +34,6 @@ const exportPattern: Action< Pattern > = {
 	id: 'export-pattern',
 	label: __( 'Export as JSON' ),
 	supportsBulk: true,
-	isEligible: ( item ) => {
-		if ( ! item.type ) {
-			return false;
-		}
-		return item.type === 'wp_block';
-	},
 	callback: async ( items ) => {
 		if ( items.length === 1 ) {
 			return downloadBlob(
--- a/packages/editor/src/dataviews/actions/export-pattern.tsx
+++ b/packages/editor/src/dataviews/actions/export-pattern.tsx
@@ -9,18 +9,15 @@ import { downloadZip } from 'client-zip';
  */
 import { downloadBlob } from '@wordpress/blob';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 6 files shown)


packages/editor/src/dataviews/actions/export-pattern.tsx
```diff
@@ -33,15 +30,15 @@ function getJsonFromItem( item ) {
...
+const exportPattern: Action< Pattern > = {
 	id: 'export-pattern',
 	label: __( 'Export as JSON' ),
 	supportsBulk: true,
 	isEligible: ( item ) => {
 		if ( ! item.type ) {
 			return false;
 		}
-		return item.type === PATTERN_TYPES.user;
+		return item.type === 'wp_block';
```
ntsekouras: We still need this because of the mapping of user patterns? When your other PR lands, it should be enough to just register it with the `wp_block` type, correct?

youknowriad: Right it seems we might not need this.

packages/editor/src/dataviews/store/reducer.ts
```diff
@@ -20,18 +20,26 @@ function actions( state: ActionState = {}, action: ReduxAction ) {
 			return {
 				...state,
 				[ action.kind ]: {
+					...state[ action.kind ],
 					[ action.name ]: [
-						...( state[ action.kind ]?.[ action.name ] ?? [] ),
+						...(
+							state[ action.kind ]?.[ action.name ] ?? []
+						).filter(
+							( _action ) => _action.id !== action.config.id
```
ntsekouras, line 27: Why do we need this filtering? Is it just defensive programming for trying to register the same action twice or was it needed actually for this PR?

ntsekouras (approved):
Besides my small question about the filtering, this looks good. Thanks!

packages/editor/src/dataviews/store/reducer.ts
youknowriad: Yes, It's to override the previous action with the same id.

youknowriad:
I can't merge this PR until the parent PR is reviewed #63042

PR #63049, opened 2024-07-02
Reduce specifity of social link icon specific colors

talldan:
## What?
Reduces the specificity of the social links block background and foreground colors to 0,1,0.

## Why?
This allows colors defined in the theme.json block `css` property to override the default block css.

## How?
Wraps the selector in `:where` which results in the generated css having selectors like `:where(.wp-block-social-links:not(.is-style-logos-only)) .wp-social-link-behance` compared to the previous output of `.wp-block-social-links:not(.is-style-logos-only) .wp-social-link-behance`.

## Testing Instructions
1. Add something like the following snippet to your theme.json in the `styles.blocks` object:
```json
"core/social-links": {
	"css": " .wp-block-social-link {background-color: silver; color: red;}"
},
```
2. Add a social links block and some individual icons, check that the background is silver and the color is red.
3. Remove that code from the theme.json and check that the social links block works normally with no changes.

## Screenshots or screencast 
In trunk (no silver background or red icon):
Screenshot 2024-07-02 at 6 36 04 pm

In this PR (silver background and red icon):
Screenshot 2024-07-02 at 6 35 39 pm

### DIFF
--- a/packages/block-library/src/social-links/style.scss
+++ b/packages/block-library/src/social-links/style.scss
@@ -107,7 +107,7 @@
 }
 
 // Provide colors for a range of icons.
-.wp-block-social-links:not(.is-style-logos-only) {
+:where(.wp-block-social-links:not(.is-style-logos-only)) {
 	// Generic items such as mail, feed, etc.
 	@import "../social-link/socials-with-bg.scss";
 }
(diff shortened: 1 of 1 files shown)


bgardner:
@talldan Using Playground, I just tested to see if the removal of !Important leaves the social icons styled as expected.

Confirmed, and works! Possible we get this into RC2?

bgardner:
Here’s a video of me removing the 3 instances of !important in CSS generated by theme.json:

https://github.com/WordPress/gutenberg/assets/486261/bf47bb11-5e5d-462e-826a-ac650f09a74a

ellatrix (approved):

PR #63052, opened 2024-07-02
Fix Incorrect URL basename logic in EmbedPreview

### LINKED ISSUE #60248: Embeds: Incorrect URL basename logic in `EmbedPreview`
### Description

The `EmbedPreview` component can display a message such as `Embedded content from %s can't be previewed in the editor.`

It uses the embedded URL's hostname for that, but does so poorly, causing the message to be quite weird and inconsistent.

Two suggestions:

1. Pass provider name instead (e.g. Twitter, YouTube)
2. Use something like `new URL(..).origin` rather than `splice()` etc. to make the URL fallback more consistent

1. Insert embed block(s) for various providers
2. Set `previewable` attribute to `false`
3. See the message

Screenshot 2024-03-27 at 11 18 24

### Environment info

- Gutenberg 18

Yes

Yes

dhananjaykuber:
Fixes #60248

## What?
This PR improves the URL parsing in the EmbedPreview component to display a more consistent and user-friendly source for embedded content.

## Why?
The current implementation of the EmbedPreview component displays inconsistent messages about the source of embedded content. This PR addresses this issue by providing a more reliable method to extract the hostname from URLs.

## How?
1. We replaced the complex URL parsing logic with a simpler approach using the URL API:
```const embedSourceUrl = new URL(url).hostname;```
2. This new embedSourceUrl is used in  iframe title and error message

## Testing Instructions
1. Open a post or page.
2. Insert embed block(s) for various providers
3. Set previewable attribute to false
4. See the consistent message

#### Before
Screenshot 2024-07-02 at 2 46 33 PM

#### After
Screenshot 2024-07-02 at 4 17 12 PM

### DIFF
--- a/packages/block-library/src/embed/embed-preview.js
+++ b/packages/block-library/src/embed/embed-preview.js
@@ -73,14 +73,11 @@ class EmbedPreview extends Component {
 		const { interactive } = this.state;
 
 		const html = 'photo' === type ? getPhotoHtml( preview ) : preview.html;
-		const parsedHost = new URL( url ).host.split( '.' );
-		const parsedHostBaseUrl = parsedHost
-			.splice( parsedHost.length - 2, parsedHost.length - 1 )
-			.join( '.' );
+		const embedSourceUrl = new URL( url ).hostname;
 		const iframeTitle = sprintf(
 			// translators: %s: host providing embed content e.g: www.youtube.com
 			__( 'Embedded content from %s' ),
-			parsedHostBaseUrl
+			embedSourceUrl
 		);
 		const sandboxClassnames = clsx(
 			type,
@@ -136,7 +133,7 @@ class EmbedPreview extends Component {
 								__(
 									"Embedded content from %s can't be previewed in the editor."
 								),
-								parsedHostBaseUrl
+								embedSourceUrl
 							) }
 						</p>
 					</Placeholder>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/block-library/src/embed/embed-preview.js [hunk 1]
```diff
@@ -73,14 +73,11 @@ class EmbedPreview extends Component {
 		const { interactive } = this.state;
 
 		const html = 'photo' === type ? getPhotoHtml( preview ) : preview.html;
-		const parsedHost = new URL( url ).host.split( '.' );
-		const parsedHostBaseUrl = parsedHost
-			.splice( parsedHost.length - 2, parsedHost.length - 1 )
-			.join( '.' );
+		const embedSourceUrl = new URL( url ).hostname;
```
Mamaduka: I think we can use the [`getAuthority( url )`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-url/#getauthority) method from the URL package here.

You also need to update the e2e test for embedding to reflect the fixed basename.

CleanShot 2024-07-03 at 09 33 46

dhananjaykuber: Hi @Mamaduka,
I have used [getAuthority( url )](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-url/#getauthority) method and updated e2e test case. Please review it once.

Thank you.

Mamaduka (approved):
Thank you, @dhananjaykuber!

The fix works as expected ✅

PR #63056, opened 2024-07-02
Build: Enable TypeScript skipDefaultLibCheck

sirreal:
## What?

Enabling this configuration should improve TypeScript build performance with few or no downsides.

## Why?

This skips _checking_ types that are bundled with TypeScript. We should
be able to trust these types, so they can be used without checking them.

Many packages use the DOM types and this appeared as a hot spot when
analyzing the TypeScript build.

Adding this configuration should improve TypeScript build performance.

### Benchmark

Run `./node_modules/.bin/tsc --build --force`, which is a build without caches:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `branch` | 80.844 ± 2.019 | 78.665 | 83.717 | 1.00 |
| `trunk` | 91.972 ± 1.819 | 88.670 | 94.230 | 1.14 ± 0.04 |

That's a ~10s build improvement on my machine.

## Testing Instructions
CI passes.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -14,6 +14,8 @@
 		"emitDeclarationOnly": true,
 		"isolatedModules": true,
 
+		"skipDefaultLibCheck": true,
+
 		/* Strict Type-Checking Options */
 		"strict": true,
 


tsconfig.base.json
```diff
@@ -14,6 +14,8 @@
 		"emitDeclarationOnly": true,
 		"isolatedModules": true,
 
+		"skipDefaultLibCheck": true,
```
tyxla, line 17: TS doc [recommends](https://www.typescriptlang.org/tsconfig/#skipDefaultLibCheck) that we use `skipLibCheck` instead. Any good reason to prefer `skipDefaultLibCheck`?

tyxla (commented):
Nice improvement 👍 Just one little question before we 🚀

tsconfig.base.json
sirreal: I noticed that, I'm not sure why they recommend that.

`skipLibCheck` ensures that the libraries we depend on typecheck. I understand that `skipDefaultLibCheck` only skips checking bundle libs (like DOM, ES2020, etc.). These libs are bundled with TypeScript and they should be the highest quality.

With other libs that come from other node_modules, it seems best to check their types especially because if we don't check these files, the errors can appear in other projects that have `skipLibCheck` disabled. We've seen that packages become stale or have conflicting types that we need to address, I dealt with a lot of this recently in https://github.com/WordPress/gutenberg/pull/61486 and related PRs.

tyxla: Makes sense to me. Seems like TS docs need some love!

tyxla (approved):
I'm seeing a ~8s and ~10s improvement respectively on 2 runs 👍 Let's go for it 🚀

PR #63057, opened 2024-07-02
Adding Font size presets UI

### LINKED ISSUE #61987: Add UI to Global styles to allow modification of font size presets

matiasbenedetto:
## What?
Adds UI to Global styles to allow modification of font size presets.
Fixes: https://github.com/WordPress/gutenberg/issues/61987

This implementation started here https://github.com/WordPress/gutenberg/pull/62328 but the approach used changed so I'm continuing that work in this new PR. 

## Why?
To allow users to edit the font size presets using the editor.

## How?
By implementing the UI to handling creation, update and removal of font size presets.

## Testing Instructions
Watch the screencast to see how the UI works and try to use the feature on your testing env.

## Screenshots or screencast 
[Screencast from 02-07-24 17:45:39.webm](https://github.com/WordPress/gutenberg/assets/1310626/a952f941-b8a0-4e48-aa57-8f1c5a4b0716)

## :memo: Reminder:
Before merging add the props from the first PR:
```
Co-authored-by: matiasbenedetto <mmaattiiaass@git.wordpress.org>
Co-authored-by: t-hamano <wildworks@git.wordpress.org>
Co-authored-by: tyxla <tyxla@git.wordpress.org>
Co-authored-by: jasmussen <joen@git.wordpress.org>
Co-authored-by: mirka <0mirka00@git.wordpress.org>
Co-authored-by: ciampo <mciampini@git.wordpress.org>
Co-authored-by: beafialho <beafialho@git.wordpress.org>
Co-authored-by: markhowellsmead <markhowellsmead@git.wordpress.org>
Co-authored-by: youknowriad <youknowriad@git.wordpress.org>
Co-authored-by: annezazu <annezazu@git.wordpress.org>
Co-authored-by: carolinan <poena@git.wordpress.org>
Co-authored-by: luminuu <luminuu@git.wordpress.org>
```


BenjaminZekavica:
It looks amazing :) Good job @matiasbenedetto 👍 One question: Couldn't this component be used for margins and paddings too, so they can be adjusted centrally?

matiasbenedetto:
> It looks amazing :) Good job @matiasbenedetto 👍 One question: Couldn't this component be used for margins and paddings too, so they can be adjusted centrally?

@BenjaminZekavica do you mean implementing this kind of UI to manage the spacing presets?
If got your question right, the  answer is yes! There's an issue created about that (https://github.com/WordPress/gutenberg/issues/61986). I plan to start working on that implementation meanwhile this PR is reviewed.

jasmussen:
Took this one for a spin. It's close!

font presets

I'm not sure if all of the following feedback has to be addressed in this PR, probably much of it are existing issues. But we should circle back to fix these, they are starting to compound. One issue: 

Screenshot 2024-07-03 at 12 52 45

This is using the "ItemGroup" pattern. Those should always be 40px tall, like buttons. This one is ~44px.

CC: @richtabor as I know you've worked on an adjacent palette drilldown, in case there's some DNA we can share. If the heading should match from ItemGroup to drilldown title, it should say "Font size presets" (note, no number of presets, and fix the typo that says "Font size**s** presets". Depending on what Rich says, though, we might want to change it to "Edit size presets".

Similarly, these itemgroup items are ~32px, and should also be 40px:

Screenshot 2024-07-03 at 12 55 21

A few more things: 

* Instead of "Manage font size presets", you can type out "Create and edit the presets used for font sizes across the site."—remember the period at the end.
* Can we use $gray-700 text color for the font sizes in the right side of the itemgroup
* All of these ItemGroup items drill down a level, so they should have the > chevron on the right.

There's still some curious focus style issue for itemgroup items:
Screenshot 2024-07-03 at 13 01 30
Notice here the double border at the bottom. That appears if I click the item, then drag out. Happens to all ItemGroup items.

When I open the modal to rename a preset, focus is on the dialog itself:

Screenshot 2024-07-03 at 13 03 00

Can you set focus on the input field? Then it will match similar rename dialogs elsewhere in the UI:

Screenshot 2024-07-03 at 13 03 17

This option could use an "Are you sure" confirm:

Screenshot 2024-07-03 at 13 04 30

For the individual font sizes:

Screenshot 2024-07-03 at 13 05 26

* Can you change the help text for the Fluid Typography help text. Something that explains what it is, perhaps: "Scale the font size dynamically to fit the screen or viewport." There's no need to imply default values, that is implied by the "Custom" toggle that appears when you enable it.

A possible followup to consider; if you set a min-value that's larger than the max value, nothing happens, and the max value is used. Perhaps we can be smarter? Or maybe it's fine?

With that, I think we can land this one.

matiasbenedetto:
@jasmussen Thanks for the review! I implemented all your suggestions. I left out only 2 items:

> There's still some curious focus style issue for itemgroup items.

This seems related to the styles of that component. This PR doesn't add or modify the styles of that component so I consider this item out of the scope of this PR.

> A possible followup to consider; if you set a min-value that's larger than the max value, nothing happens, and the max value is used. Perhaps we can be smarter? Or maybe it's fine?

Make the controls smarter seems good for a follow-up. I think it should not  block this implementation.

jasmussen:
This is getting close. Here's a status:

status

A question:

Screenshot 2024-07-04 at 15 01 57

What happens to the base "Size" when you add custom fluid values? The preview appears to still render from the base size. This may become a rabbit hole, but I wonder if the base size panel should be grayed and say "Fluid" in italic" inside the disabled field, when custom values are added? Kind of like how Image dimensions say "auto"? Would appreciate a gut-check by @WordPress/gutenberg-design on this one.

These items are now 49px tall:
 
Screenshot 2024-07-04 at 15 05 24

As I noted in another PR, every ItemGroup should be exactly 40px tall, unless we're looking at edgecases where the text wraps. I also noted that this may be a separate issue. The padding left and right is also larger than in other ItemGroup Items.  But in case you added a "large" prop here, best remove it at least.

A way to think of it is that we are creating a typography counterpart to this palette screen:

Screenshot 2024-07-04 at 15 06 56

The last piece is this one, and I'd also love additional input on this:
Screenshot 2024-07-04 at 15 02 13

IMO it should say either "Font size presets" (without a number), or it should say "Edit size presets" (also without a number). The former matches the drilldown page title, the latter is more active language. 

If there's broad support for including the number, then it should be "6 font size presets" (lowercase f), but let's hear from @WordPress/gutenberg-design.

These all seem small, so nice work, this is very close!

Agreed many of the feedback items can be separate.

matiasbenedetto:
> As I noted in another PR, every ItemGroup should be exactly 40px tall, unless we're looking at edgecases where the text wraps. I also noted that this may be a separate issue. The padding left and right is also larger than in other ItemGroup Items. But in case you added a "large" prop here, best remove it at least.

I removed the large prop, but the tall of the ItemGroup component seems out of the scope of this PR because it wasn't changed in any way in this implementation and I think it should not block it.

matiasbenedetto:
> What happens to the base "Size" when you add custom fluid values? The preview appears to still render from the base size. This may become a rabbit hole, but I wonder if the base size panel should be grayed and say "Fluid" in italic" inside the disabled field, when custom values are added? Kind of like how Image dimensions say "auto"?

It seems like the base size is ignore when a min a max is set so it would good to add some kind of visual hint about that.

packages/edit-site/src/components/global-styles/ui.js
```diff
@@ -231,7 +233,12 @@ function GlobalStylesBlockLink() {
 		if ( newPath !== currentPath ) {
 			navigator.goTo( newPath, { skipFocus: true } );
 		}
-	}, [ selectedBlockClientId, selectedBlockName, blockHasGlobalStyles ] );
+	}, [
+		selectedBlockClientId,
+		selectedBlockName,
+		blockHasGlobalStyles,
+		navigator,
```
mikachan: This new dependency on `navigator` is causing the `push-to-global-styles` test to fail. The test expects to need to click the "Headings" button in the Styles panel, but the Styles panel path is now updated whenever the selected block changes, which means the Headings panel is opened automatically when a Heading block is selected.

This saves a click, but I'm not sure what the intended behaviour is. Either:
1. The Styles panel path should update whenever the selected block changes - saves on clicks, but is different to the current behaviour.
2. The Styles panel path does not change regardless of the selected block, and it will only change with direct interaction (i.e. navigation) within the Styles panel.

If it's number 1 we should update the test to remove these additional navigation clicks; if it's number 2 then I think we should remove the dependency on `navigator` here.

jasmussen:
> It seems like the base size is ignore when a min a max is set so it would good to add some kind of visual hint about that.

Also an additional nuance, it appears that the font size in the _preview_ takes from the base size, not the min/max values. So beyond any additional ideas coming in, can we:

* Make the preview always have the font size of the min-max values if set?
* Gray out the base size field if min/max values are set?

t-hamano:
Thanks for your continued efforts.

Some of these were mentioned in the previous PR, but after some quick testing I noticed the following:

**1. The font size value is not fully displayed even though the label is short.**

| Actual | What I expect |
|--------|--------|
| actual | expected | 

**2. Default/Theme palettes can be renamed and edited.**

No such specification exists for color or shadow palettes. I think a condition [like this](https://github.com/WordPress/gutenberg/blob/679122ec8d835a266ce2ad5e6ba23a6d5a7fec15/packages/edit-site/src/components/global-styles/shadows-edit-panel.js#L164-L166) for a shadow palette is necessary.

packages/edit-site/src/components/global-styles/font-sizes/confirm-delete-font-size-dialog.js
```diff
@@ -0,0 +1,49 @@
...
+	isOpen,
+	toggleOpen,
+	handleRemoveFontSize,
+} ) {
+	const navigator = useNavigator();
+
+	const handleConfirm = async () => {
+		toggleOpen();
+		handleRemoveFontSize( fontSize );
+		navigator.goTo( '/typography/font-sizes/' );
```
t-hamano: We don't need `navitator.goTo()` because we have `goBack()` [here](https://github.com/WordPress/gutenberg/pull/63057/files#diff-88536b41b2c498208a524ef21e82ff4796a9bd768fafa50c82ef0bbe79d67f21R111).

packages/edit-site/src/components/global-styles/font-sizes/font-size.js
```diff
@@ -0,0 +1,248 @@
...
+							marginTop={ 3 }
+							marginBottom={ 0 }
+							paddingX={ 4 }
+						>
+							<DropdownMenu
+								trigger={
+									<Button
+										size="small"
+										icon={ moreVertical }
+										label={ __( 'Menu' ) }
```
t-hamano:
```suggestion
										label={ __( 'Font size options' ) }
```

Improved accessibility. There might be a more appropriate label.

packages/edit-site/src/components/global-styles/font-sizes/font-sizes.js
```diff
@@ -0,0 +1,251 @@
...
+								onClick={ handleAddFontSize }
+							/>
+						) }
+						{ !! handleResetFontSizes && (
+							<DropdownMenu
+								trigger={
+									<Button
+										size="small"
+										icon={ moreVertical }
+										label={ __( 'Menu' ) }
```
t-hamano:
```suggestion
										label={ __( 'Font size presets options' ) }
```

packages/edit-site/src/components/global-styles/font-sizes/font-size.js
```diff
@@ -0,0 +1,248 @@
...
+				handleRemoveFontSize={ handleRemoveFontSize }
+			/>
+
+			{ isRenameDialogOpen && (
+				<RenameFontSizeDialog
+					fontSize={ fontSize }
+					isOpen={ isRenameDialogOpen }
+					toggleOpen={ toggleRenameDialog }
+					handleRename={ handleNameChange }
+				/>
```
t-hamano:
```suggestion
				<RenameFontSizeDialog
					fontSize={ fontSize }
					toggleOpen={ toggleRenameDialog }
					handleRename={ handleNameChange }
				/>
```

The `RenameFontSizeDialog` component does not receive an `isOpen` prop.

packages/edit-site/src/components/global-styles/font-sizes/rename-font-size-dialog.js
```diff
@@ -0,0 +1,65 @@
...
+						<Button variant="tertiary" onClick={ toggleOpen }>
+							{ __( 'Cancel' ) }
+						</Button>
+					</FlexItem>
+					<FlexItem>
+						<Button variant="primary" type="submit">
+							{ __( 'Save' ) }
+						</Button>
+					</FlexItem>
+				</Flex>
```
t-hamano:
```suggestion
				<VStack spacing="3">
					<InputControl
						__next40pxDefaultSize
						autoComplete="off"
						value={ newName }
						onChange={ setNewName }
						label={ __( 'Name' ) }
						placeholder={ __( 'Font size preset name' ) }
					/>
					<HStack justify="right">
						<Button
							__next40pxDefaultSize
							variant="tertiary"
							onClick={ toggleOpen }
						>
							{ __( 'Cancel' ) }
						</Button>
						<Button
							__next40pxDefaultSize
							variant="primary"
							type="submit"
						>
							{ __( 'Save' ) }
						</Button>
					</HStack>
				</VStack>
```

Make it more consistent with other rename UIs

| Before | After |
|--------|--------|
| image| image |

packages/edit-site/src/components/global-styles/size-control/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+	const [ valueQuantity, valueUnit = 'px' ] =
+		parseQuantityAndUnitFromRawValue( value, units );
+
+	const isValueUnitRelative =
+		!! valueUnit && [ 'em', 'rem', 'vw', 'vh' ].includes( valueUnit );
+
+	// Receives the new value from the UnitControl component as a string containing the value and unit.
+	const handleUnitControlChange = ( newValue ) => {
+		onChange?.( newValue );
```
t-hamano: Isn't the `?` unnecessary?

matiasbenedetto:
> Default/Theme palettes can be renamed and edited.

@t-hamano thanks for the review. I removed the ability to 'remove' and 'rename' font size presets in [`908da98` (#63057)](https://github.com/WordPress/gutenberg/pull/63057/commits/908da98ecf76a06208ac72b5237643c114d2c71d). I think your coment was around that. The 'edit' ability (change the value of the preset) is still there because the equivalent action is possible with both colors and shadows.

matiasbenedetto:
> Also an additional nuance, it appears that the font size in the preview takes from the base size, not the min/max values. So beyond any additional ideas coming in, can we:

> Make the preview always have the font size of the min-max values if set?


@jasmussen, yes! added here: [`fd86a59` (#63057)](https://github.com/WordPress/gutenberg/pull/63057/commits/fd86a59f86777c767cfbed2096e8c9f242a6a934). In this way we use the real formula used by the preset.

> Gray out the base size field if min/max values are set?

Yes, done here:[`fb1eec7` (#63057)](https://github.com/WordPress/gutenberg/pull/63057/commits/fb1eec78f4233bd74fd5b6ff5c5c588c1bce9461)

packages/edit-site/src/components/global-styles/ui.js
t-hamano: I suspect this change was made to prevent ESLint warnings, but it shouldn't be related to the implementation of font size presets. For now, I think it's better to revert this change to prevent unintended issues from occurring.

This also applies to [this line](https://github.com/WordPress/gutenberg/pull/63057/files#diff-23a41dabd87707ed160ff1803dd675b95532d1ef65b870fe635d8603371fe539R292)

packages/edit-site/src/components/global-styles/font-sizes/font-size.js
```diff
@@ -0,0 +1,250 @@
...
+					<ScreenHeader
+						title={ fontSize.name }
+						description={ sprintf(
+							/* translators: %s: font size preset name. */
+							__( 'Manage the font size %s.' ),
+							fontSize.name
+						) }
+						onBack={ () => goTo( '/typography/font-sizes/' ) }
+					/>
+					{ origin !== 'default' && (
```
t-hamano:
```suggestion
					{ origin === 'custom' && (
```

The current conditional statement allows theme fonts to be deleted and renamed.

If we want to match the specifications of color and shadow presets, shouldn't only **custom** presets be allowed to be deleted and renamed?

packages/edit-site/src/components/global-styles/size-control/index.js
```diff
@@ -0,0 +1,88 @@
...
+	};
+
+	// Receives the new value from the RangeControl component as a number.
+	const handleRangeControlChange = ( newValue ) => {
+		onChange?.( newValue + valueUnit );
+	};
+
+	return (
+		<BaseControl { ...baseControlProps }>
+			<Flex className="components-size-control__custom-size-control">
```
t-hamano: Since this component doesn't belong to `@wordpress/components` package yet, I don't think a class name is necessary.

The same goes for [this line](https://github.com/WordPress/gutenberg/pull/63057/files#diff-0d1577d32484a20cb4c1e1bd1a8910775dc23961fd34e0ab449416786dc825c2R69).

packages/edit-site/src/components/global-styles/size-control/index.js
```diff
@@ -0,0 +1,88 @@
...
+				<FlexItem isBlock>
+					<UnitControl
+						__next40pxDefaultSize
+						__nextHasNoMarginBottom
+						label={ __( 'Custom' ) }
+						labelPosition="top"
+						hideLabelFromVision
```
t-hamano:
```suggestion
						hideLabelFromVision
```

Since `hideLabelFromVision` is enabled, `labelPosition` is not necessary.

t-hamano (commented):
Thanks for the update! From what I can see, the code base is nearly complete.

t-hamano:
> The font size value is not fully displayed even though the label is short.

Would this be possible to solve the issue by making the following changes?

```diff
diff --git a/packages/edit-site/src/components/global-styles/font-sizes/font-sizes.js b/packages/edit-site/src/components/global-styles/font-sizes/font-sizes.js
index f0f9f7307b..a95bc32cd8 100644
--- a/packages/edit-site/src/components/global-styles/font-sizes/font-sizes.js
+++ b/packages/edit-site/src/components/global-styles/font-sizes/font-sizes.js
@@ -10,6 +10,7 @@ import {
        __experimentalItemGroup as ItemGroup,
        __experimentalVStack as VStack,
        __experimentalHStack as HStack,
+       FlexBlock,
        FlexItem,
        Button,
 } from '@wordpress/components';
@@ -89,7 +90,9 @@ function FontSizeGroup( {
                                                                        <Button
                                                                                size="small"
                                                                                icon={ moreVertical }
-                                                                               label={ __( 'Font size presets options' ) }
+                                                                               label={ __(
+                                                                                       'Font size presets options'
+                                                                               ) }
                                                                        />
                                                                }
                                                        >
@@ -113,9 +116,9 @@ function FontSizeGroup( {
                                                                path={ `/typography/font-sizes/${ origin }/${ size.slug }` }
                                                        >
                                                                <HStack direction="row">
-                                                                       <FlexItem className="edit-site-font-size__item">
+                                                                       <FlexBlock className="edit-site-font-size__item">
                                                                                { size.name }
-                                                                       </FlexItem>
+                                                                       </FlexBlock>
                                                                        <FlexItem>
                                                                                <HStack justify="flex-end">
                                                                                        <FlexItem className="edit-site-font-size__item edit-site-font-size__item-value">
diff --git a/packages/edit-site/src/components/global-styles/style.scss b/packages/edit-site/src/components/global-styles/style.scss
index d8f1c0b7c8..8fa66b6d1b 100644
--- a/packages/edit-site/src/components/global-styles/style.scss
+++ b/packages/edit-site/src/components/global-styles/style.scss
@@ -31,7 +31,6 @@
 }
 
 .edit-site-font-size__item-value {
-       max-width: 50% !important;
        color: $gray-700;
 }
```

</details> 

| Before | After |
|--------|--------|
| image| image|

packages/edit-site/src/components/global-styles/font-sizes/font-size.js
matiasbenedetto: Yes, done

packages/edit-site/src/components/global-styles/size-control/index.js
matiasbenedetto: :+1: removed here: [`d94d75b` (#63057)](https://github.com/WordPress/gutenberg/pull/63057/commits/d94d75bb6480b3aa333da8683923768485452564)

matiasbenedetto:
> Would this be possible to solve the issue by making the following changes?

Good, yes that fixes the problem. @t-hamano Feel free to commit directly next time.

packages/edit-site/src/components/global-styles/ui.js
matiasbenedetto: I reverted the `navigator` dependency. The linter is not complaining and the tests are passing, so I think it's ok this way. Thanks for the research about this!

packages/edit-site/src/components/global-styles/font-sizes/confirm-delete-font-size-dialog.js
matiasbenedetto: Actually the navigation needed wasn't needed. I removed that here: [`82c1b9a` (#63057)](https://github.com/WordPress/gutenberg/pull/63057/commits/82c1b9aa56737ef8831ce5bb843babfccac200fa)

packages/edit-site/src/components/global-styles/size-control/index.js
matiasbenedetto: it's doesn't hurt but still I removed it.

matiasbenedetto:
Thanks for the reviews. I think I addressed all the feedback received.

mikachan (approved):
Thanks for all the work on this, @matiasbenedetto 👏 

This is testing great for me:

- Can see all font size presets from the current theme
- Can add and delete custom presets
- The font sizing controls all work as I expect
- Font size presets are successfully saved in theme.json via CBT plugin (and therefore must be saved successfully in global styles), and appear under 'Theme' rather than 'Custom' after saving to the theme

I also looked through the previous feedback and I can't see anything that wasn't addressed 👍 

This looks good to me, although it might be good to get one more 👀 from @WordPress/gutenberg-design before bringing this in.

jasmussen:
Working well for me in a quick test, and well addressed past feedback. Thanks for working on this!

Issue #61051, opened 2024-04-24
Proposal: Keep the Inserter open


richtabor:
I wanted to explore if the Inserter behaved like other sidebars, where it remains open unless explicitly closed. This way the Inserter does not disappear when you're engaging with blocks after inserting them. 

As part of this exploration, it became clear the expectation is that there's a close button, like on every other sidebar (List View, Inspector, Styles, etc). To align with those other panels, that means placing the close button at the end of the tabs that depict the contents below. 

In doing so, the search field would need to be moved below the tabs, where each tab (Blocks, Patterns, and Media) would now scope search results—rather than returning all types of results. For example, when the Blocks tab is active, you are searching for blocks, and blocks render as search results.  

I would also expect a search term would persist, if an alternate tab is selected. For example, if I search for "gallery" I would see the "Gallery" block while the "Block" tab is active, and any gallery patterns while the "Gallery" tab is active. 

### Visual 

https://github.com/WordPress/gutenberg/assets/1813435/95d29750-6415-4870-a1a4-b4615e3b0115

| Before  | After |
| ------------- | ------------- |
|![CleanShot 2024-04-24 at 15 57 12](https://github.com/WordPress/gutenberg/assets/1813435/15a32b02-dd5e-4584-a4e3-be3383713565)|![CleanShot 2024-04-24 at 15 57 25](https://github.com/WordPress/gutenberg/assets/1813435/a8523284-c0ad-42bc-be7d-35a25f2b66f2)|

- You'll note a few other UI adjustments to generally clean up the inserter as well. If we proceed down this route, those are separate concerns we can make individual PRs for. 
- I am not proposing changing any default states of either the Blocks, Patterns, or Media inserter views—only search results. 

### Prototype

I have a [Figma prototype](https://www.figma.com/proto/oEm1ck3f5gfGYyOR1Fs71T/Gutenberg-Ideas?page-id=2246%3A2419&type=design&node-id=2254-29653&viewport=-3694%2C-1273%2C1&t=IN2KfLr7cUNn8EOj-1&scaling=min-zoom&starting-point-node-id=2254%3A29653&mode=design) to demonstrate this.

fabiankaegy:
I'm in favor of the idea of keeping the inserted open. I think that would be great especially for working with patterns. 

The only think about your proposed change I'm not sure about is moving the search bar below the tabs. 

[...]

PR #63059, opened 2024-07-02
Inserter: Remove the dialog behaviour

scruffian:
## What?
This removes the dialog behaviour from the inserter sidebar panel. This was originally proposed in #61051 but was removed so that it didn't go into WordPress 6.6.

## Why?
Consistency. All of the other panels behave like this.

## How?
Remove the useDialog hook.

## Testing Instructions
0. Open the Site Editor
1. Open the inserter
2. Click on the canvas
3. Check that the inserter remains open.

### DIFF
--- a/packages/editor/src/components/inserter-sidebar/index.js
+++ b/packages/editor/src/components/inserter-sidebar/index.js
@@ -6,10 +6,7 @@ import {
 	store as blockEditorStore,
 	privateApis as blockEditorPrivateApis,
 } from '@wordpress/block-editor';
-import {
-	useViewportMatch,
-	__experimentalUseDialog as useDialog,
-} from '@wordpress/compose';
+import { useViewportMatch } from '@wordpress/compose';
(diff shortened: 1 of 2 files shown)


richtabor (approved):
LGTM 

https://github.com/WordPress/gutenberg/assets/1813435/235ebd5b-6229-455e-a02c-294f632db2d2

afercia:
For history:
This PR introduced a regression in the behavior of the Patterns explorer modal dialog reported at https://github.com/WordPress/gutenberg/issues/68728
- Previously, the Patterns explorer modal dialog when inserting a pattern, because the Inserter panel closed in the first place.
- Now, the modal dialog stays open introducing two problems:
  - The Snackbar notices appears behind the modal dialog.
  - More importantly, focus still gets moved to the newly inserted pattern, that is: outside of the modal dialog thus breaking the expected keyboard interaction.

Issue #60847, opened 2024-04-18
Site editor responsiveness


SaxonF:
## What problem does this address?
As we [expand data views](https://github.com/WordPress/gutenberg/issues/55083) in a way that requires more horizontal space we need to more strongly consider how to respond to various screen sizes. 

## What is your proposed solution?
**Sidebar**
Reduce sidebar to min width of 240px. If a particular page or plugin requires more space, it should be able to dynamically adjust the size. We are working towards removing detail views (e.g. page detail) from the sidebar and once we do that the need for extra width will mostly go away.

![image](https://github.com/WordPress/gutenberg/assets/1072756/7411e1a3-34b9-4b89-97ad-08a66b1c09ac)

**At 1440**
There is enough horizontal space to contain the sidebar, the content frame in a minimised state (e.g. [list views](https://github.com/WordPress/gutenberg/issues/59659#issuecomment-2029944054) and the preview frame

![image](https://github.com/WordPress/gutenberg/assets/1072756/7f5fe4fd-0802-4f50-8c84-f7cc8cf6a5f9)

**At 1024**

Tables start to scroll horizontally if need more space. Horizontal overflow when both content frame and preview frame are visible.

![image](https://github.com/WordPress/gutenberg/assets/1072756/1f44806d-c87c-4777-aa7d-342929d17455)

https://github.com/WordPress/gutenberg/assets/1072756/43baea59-d278-4182-9de5-a348d78ec1c1

**At 768**

Sidebar is condensed with the header of the sidebar visible above the two frames. Clicking site icon (or more menu icon) reveals fixed sidebar floating above. We horizontal scroll when content frame and preview frame are visible at the same time.

![image](https://github.com/WordPress/gutenberg/assets/1072756/9b541c95-1fae-48e3-9f9e-4e2bf1a4faec)

https://github.com/WordPress/gutenberg/assets/1072756/7ee442a0-6cc3-4830-9405-50e7064a7129

**At 480**
Much the same as 768 except preview frame is now hidden. List view types take a user directly to the editor on clicking (or whatever page is the primary action for other post types).

![image](https://github.com/WordPress/gutenberg/assets/1072756/1056c11e-cfe7-48f0-98ef-6a2074a6e581)

PR #63060, opened 2024-07-02
Make SiteHub available in mobile viewports

oandregal:
Related to https://github.com/WordPress/gutenberg/issues/60847 and https://github.com/WordPress/gutenberg/issues/62878

## What?

This makes the SiteHub always available for mobile viewports, for the Pages, Patterns, and Templates pages. It mimics what we had in 6.5.

| In `trunk` | This PR |
| --- | --- |
| Captura de ecrã 2024-07-02, às 17 03 35 | Captura de ecrã 2024-07-02, às 17 02 53 |

## Why?

This brings back what we had in 6.5 for the Patterns page. While the overall direction is https://github.com/WordPress/gutenberg/issues/60847 we can me this quick change for improved navigation.

## How?

Render the `SiteHub` component on mobile viewports.

## Testing Instructions

Navigate to Pages, Patterns, and Templates and verify the SiteHub is present.

### DIFF
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -174,10 +174,12 @@ export default function Layout( { route } ) {
 
 					{ isMobileViewport && areas.mobile && (
 						<div className="edit-site-layout__mobile">
-							<SiteHub
-								ref={ toggleRef }
-								isTransparent={ isResizableFrameOversized }
-							/>
+							{ canvasMode !== 'edit' && (
+								<SiteHub
+									ref={ toggleRef }
+									isTransparent={ isResizableFrameOversized }
+								/>
+							) }
 							{ areas.mobile }
 						</div>
 					) }
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -175,10 +175,14 @@ export default function Layout( { route } ) {
 					{ isMobileViewport && areas.mobile && (
 						<div className="edit-site-layout__mobile">
 							{ canvasMode !== 'edit' && (
-								<SiteHub
-									ref={ toggleRef }
-									isTransparent={ isResizableFrameOversized }
-								/>
+								<SidebarContent routeKey={ routeKey }>
+									<SiteHub
+										ref={ toggleRef }
+										isTransparent={
+											isResizableFrameOversized
+										}
+									/>
+								</SidebarContent>
 							) }
 							{ areas.mobile }
 						</div>
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -174,6 +174,10 @@ export default function Layout( { route } ) {
 
 					{ isMobileViewport && areas.mobile && (
 						<div className="edit-site-layout__mobile">
+							<SiteHub
+								ref={ toggleRef }
+								isTransparent={ isResizableFrameOversized }
+							/>
 							{ areas.mobile }
 						</div>
 					) }
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -153,13 +153,13 @@ export default function Layout( { route } ) {
 										} }
 										className="edit-site-layout__sidebar"
 									>
-										<SiteHub
-											ref={ toggleRef }
-											isTransparent={
-												isResizableFrameOversized
-											}
-										/>
 										<SidebarContent routeKey={ routeKey }>
+											<SiteHub
+												ref={ toggleRef }
+												isTransparent={
+													isResizableFrameOversized
+												}
+											/>
 											{ areas.sidebar }
 										</SidebarContent>
 										<SaveHub />
--- a/packages/edit-site/src/components/site-hub/index.js
+++ b/packages/edit-site/src/components/site-hub/index.js
@@ -8,14 +8,16 @@ import clsx from 'clsx';
  */
 import { useSelect, useDispatch } from '@wordpress/data';
 import { Button, __experimentalHStack as HStack } from '@wordpress/components';
+import { useViewportMatch } from '@wordpress/compose';
 import { __ } from '@wordpress/i18n';
 import { store as coreStore } from '@wordpress/core-data';
 import { decodeEntities } from '@wordpress/html-entities';
-import { memo, forwardRef } from '@wordpress/element';
+import { memo, forwardRef, useContext } from '@wordpress/element';
 import { search } from '@wordpress/icons';
 import { store as commandsStore } from '@wordpress/commands';
 import { displayShortcut } from '@wordpress/keycodes';
 import { filterURLForDisplay } from '@wordpress/url';
+import { privateApis as routerPrivateApis } from '@wordpress/router';
 
 /**
  * Internal dependencies
@@ -23,9 +25,15 @@ import { filterURLForDisplay } from '@wordpress/url';
 import { store as editSiteStore } from '../../store';
 import SiteIcon from '../site-icon';
 import { unlock } from '../../lock-unlock';
+const { useHistory } = unlock( routerPrivateApis );
+import { SidebarNavigationContext } from '../sidebar';
 
 const SiteHub = memo(
 	forwardRef( ( { isTransparent }, ref ) => {
+		const isMobileViewport = useViewportMatch( 'medium', '<' );
+		const history = useHistory();
+		const { navigate } = useContext( SidebarNavigationContext );
+
 		const { dashboardLink, homeUrl, siteTitle } = useSelect( ( select ) => {
 			const { getSettings } = unlock( select( editSiteStore ) );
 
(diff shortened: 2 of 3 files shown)


oandregal:
Note how the Patterns page behaved in WordPress 6.5. This PR essentially brings back the same behavior:

https://github.com/WordPress/gutenberg/assets/583546/ef4326c2-77d8-4251-b4e3-47a58e4ae632

oandregal:
**Behavior in `trunk`:**

https://github.com/WordPress/gutenberg/assets/583546/e33f8465-6663-4b58-9d8b-b82e9846cb09

**Behavior in this PR:**

https://github.com/WordPress/gutenberg/assets/583546/d3b2705b-0811-454c-8823-f991ec16414c

youknowriad:
I think the main issue here, is that clicking the site hub shouldn't go to the "dashboard" unless you're in the root level, it should instead navigate back.

youknowriad:
Alternatively, we could have an extra back button in the hub for mobile?

jameskoster:
Would it be unfeasible for the site icon to behave differently on mobile? 

* At site editor root level it would take you to the Dashboard
* At any other level it works like the 'back' button

That would be more closely aligned to the proposal in https://github.com/WordPress/gutenberg/issues/60847.

annezazu:
Adding backport label as ideally these fixes could be included in 6.6. If that's incorrect, feel free to remove.

ramonjd:
Having the search/command centre available is great, thanks!

> clicking the site hub shouldn't go to the "dashboard" unless you're in the root level, it should instead navigate back.

Not sure how to display this, but I agree it would be helpful to be able to return to the main menu.

In mobile view, could the site title's arrow icon be displayed at all times (and not just on hover/focus) to indicate that it'll open a new tab?

It's a bit of a mystery what the various items do unless you click around.

andrewserong:
> Would it be unfeasible for the site icon to behave differently on mobile?

I like that idea, too. In order to implement it, would we need to move the sidebar navigation context further up the hierarchy so that it can be accessed from the `SiteHub`? It seems that the site hub will need to know whether or not it's at the root level to determine whether to go "back" / up a level of the hierarchy, or go to the dashboard if we're at the root of the site editor navigation. This is the context provider I mean:

https://github.com/WordPress/gutenberg/blob/4866c3b3d8d331904fe5b4bc45737077d9322c92/packages/edit-site/src/components/sidebar/index.js#L85

That said, IMO the current state of this PR is an improvement over trunk in that it now makes it possible to back out again from the template screen, so I wouldn't be opposed to the idea of landing this PR and then we could see if we could add more smarts to the W icon button in follow-ups if need be?

packages/edit-site/src/components/layout/index.js
```diff
@@ -175,10 +175,14 @@ export default function Layout( { route } ) {
 					{ isMobileViewport && areas.mobile && (
 						<div className="edit-site-layout__mobile">
 							{ canvasMode !== 'edit' && (
-								<SiteHub
-									ref={ toggleRef }
-									isTransparent={ isResizableFrameOversized }
-								/>
+								<SidebarContent routeKey={ routeKey }>
```
youknowriad, line 178: @jsnajdr if you have any thoughts here.

oandregal:
> Would it be unfeasible for the site icon to behave differently on mobile?
> 
> - At site editor root level it would take you to the Dashboard
> - At any other level it works like the 'back' button

I've pushed some changes to make it work like this.

https://github.com/WordPress/gutenberg/assets/583546/dc43ff17-3bba-4cbc-9253-c505022e79b9

packages/edit-site/src/components/site-hub/index.js
```diff
@@ -8,24 +8,35 @@ import clsx from 'clsx';
...
 import { unlock } from '../../lock-unlock';
+const { useHistory } = unlock( routerPrivateApis );
+import { SidebarNavigationContext } from '../sidebar';
 
 const SiteHub = memo(
 	forwardRef( ( { isTransparent }, ref ) => {
+		const isMobileViewport = useViewportMatch( 'medium', '<' );
+		const history = useHistory();
+		const { navigate } = useContext( SidebarNavigationContext );
+		const isRoot =
```
oandregal, line 36: Is there a better way to detect we are in the root?

This relies on the `search` being empty, which it always is at the root level, as far as I've tested. If there's some pollution (a `postType=X` remanent in the URL, for example) this wouldn't work (but also: we should clean that).

The [sidebar](https://github.com/WordPress/gutenberg/blob/db284b1a17eb78e19d1eb00d93518756c8ce07b8/packages/edit-site/src/components/sidebar-navigation-screen-main/index.js#L36) uses a explicit `isRoot` property. Given this is a temporary thing for 6.6 and will be improved by the plans/direction at https://github.com/WordPress/gutenberg/issues/60847 I didn't want to modify the public API of `SiteHub`.

jameskoster:
That's working well @oandregal.

I appreciate I am likely asking for the moon on a stick here, but how about restricting this behavior to when the content frame is visible (when you drill in to Pages, Templates, or Patterns)?

On the other pages (Styles and Navigation) the `<` button is visible, so the W button could retain it's normal function.

Probably fine to work on that separately if it's not trivial. 

I'm to get more thoughts on the behavior here from @WordPress/gutenberg-design.

packages/edit-site/src/components/layout/index.js
```diff
@@ -153,13 +153,13 @@ export default function Layout( { route } ) {
...
 										className="edit-site-layout__sidebar"
 									>
-										<SiteHub
-											ref={ toggleRef }
-											isTransparent={
-												isResizableFrameOversized
-											}
-										/>
 										<SidebarContent routeKey={ routeKey }>
+											<SiteHub
```
andrewserong, line 157: Unfortunately it looks like moving the `SiteHub` to be a child of `SidebarContent` means that the side bar content's focusing behaviour now matches against the buttons within the site hub, which causes one of the e2e tests to fail (`'Can use keyboard to navigate the site editor'`).

From testing manually it seems that what's happening is that if you navigate to the Pages button from the root and hit Enter, when you're taken to the pages screen, the W icon is focused instead of the back button. I.e. here's where focus now lands:

image

Here's the function that handles the focusing: https://github.com/WordPress/gutenberg/blob/4866c3b3d8d331904fe5b4bc45737077d9322c92/packages/edit-site/src/components/sidebar/index.js#L22

I'm wondering if we need to either have a way for that focusing to skip the site hub and just look at the sidebar area, or if we need a way to give the site hub access to the navigation context without the other wrapping logic?

packages/edit-site/src/components/layout/index.js
andrewserong: ~Another side effect of wrapping in `SidebarContent` is that the site hub receives extra padding, so there's a jump horizontally when we dig into the lower levels of the hierarchy:~

https://github.com/WordPress/gutenberg/assets/14988353/97c95039-8f78-4822-b28b-0581332699fa

Edit: After re-checking out this branch I can't reproduce the above 🤔

packages/edit-site/src/components/layout/index.js
andrewserong: I've pushed a commit in https://github.com/WordPress/gutenberg/pull/63060/commits/4401f6037f19563d60fe50d3b1e21add43a106e3 that hopefully addresses the focus issue 🤞. It's a little hacky but let's see if it gets the test passing.

andrewserong:
This is testing pretty well for me, too @oandregal. I've pushed a small (but potentially hacky) fix in https://github.com/WordPress/gutenberg/pull/63060/commits/4401f6037f19563d60fe50d3b1e21add43a106e3 to try to get the focusing behaviour in the sidebar to skip the site hub if present. 🤞 that gets the keyboard behaviour and tests passing again. If it feels too hacky, though, do feel free to revert that commit!

I also explored a slightly more complex alternative over in https://github.com/WordPress/gutenberg/pull/63109, which looked at lifting up where we place the sidebar navigation context, but that wound up causing issues with the back/forward animation and the site hub, so I've closed that one out.

aaronrobertshaw:
Add me to the list of people this is testing well for too 🙂 

@andrewserong's tweak does appear to have sorted out the failing keyboard navigation e2e and works with manual testing. My only question is, do we need to set focus to the W icon button on the pages that don't have the back button?


https://github.com/WordPress/gutenberg/assets/60436221/932b9f9c-fdf0-4b05-bec3-af0f78700a57

oandregal:
I've prepared an alternative PR that I like more at https://github.com/WordPress/gutenberg/pull/63118

It ticks all the boxes (this one doesn't):

1. At site editor root level it takes you to the Dashboard.
2. Within the Styles and Navigation pages it takes you to the Dashboard.
3. Within the Pages, Patterns, and Template pages it works like the 'back' button.

And it doesn't suffer from the technical issues raised: [route leaked to SiteHub for behavior control](https://github.com/WordPress/gutenberg/pull/63060#discussion_r1664024756) and [component leaked to Sidebar for focus control](https://github.com/WordPress/gutenberg/pull/63060#discussion_r1664957185).

andrewserong:
Closing this out now that #63118 has landed. Thanks everyone!

PR #63066, opened 2024-07-02
Fix Pinterest URLs not embedding when not .com

### LINKED ISSUE #59400: Pinterest URLs not embedding when not .com (.es, .it...)
## Description

Pinterest URLs that don't come from the .com domain won't be embedded.
Pinterest uses a different domain for different countries and people in those countries might find it difficult to use this block.

For example, these URLs won't get embedded:
https://www.pinterest.es/pin/320670435980314659/
https://www.pinterest.it/pin/308004062030905959/

But changing the `.es` and `.it` for `.com`will make it work:
https://www.pinterest.com/pin/320670435980314659/
https://www.pinterest.com/pin/308004062030905959/

## Expected behaviour

Any URL coming from the existing interest domains in each country should work.
Having a list of them should be enough to deal with this.

## Actual behaviour

URLs won't embed if not coming from the .com domain.

## Screenshots or screen recording (optional)

image

## WordPress information
- WordPress version:  6.4.3
- Gutenberg version: 
- Are all plugins except Gutenberg deactivated?  Yes
- Are you using a default theme. Yes

dhananjaykuber:
Fixes [#59400](https://github.com/WordPress/gutenberg/issues/59400)

## What?
Fixes an issue where Pinterest URLs from non-.com domains weren't embedding correctly.

## Why?
Pinterest uses different country-specific domains, causing embed failures. This PR ensures all Pinterest URLs embed correctly by converting non-.com domains to .com.

## How?
Introduced replaceDomain function to convert any domain extension to '.com' in URLs. Updated Embed block to use this function for Pinterest URLs.

## Testing Instructions

1. Open a post or page in the Gutenberg editor.
2. Insert an Embed block with a Pinterest URL from a non-.com domain (e.g., .es, .it).
3. Ensure the URL converts to .com and the embed preview displays correctly.

## Screenshots or screencast  
Screenshot 2024-07-02 at 10 52 35 PM
Screenshot 2024-07-02 at 10 52 46 PM

### DIFF
--- a/packages/block-library/src/embed/edit.js
+++ b/packages/block-library/src/embed/edit.js
@@ -8,6 +8,7 @@ import {
 	fallback,
 	getEmbedInfoByProvider,
 	getMergedAttributesWithPreview,
+	replaceDomain,
 } from './util';
 import EmbedControls from './embed-controls';
 import { embedContentIcon } from './icons';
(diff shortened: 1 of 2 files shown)


Mamaduka:
Thanks for contributing, @dhananjaykuber!

I've already commented on the issue (https://github.com/WordPress/gutenberg/issues/59400#issuecomment-2205145445). The bug needs to be resolved in WP Core. The Embed block does not require an update/fix.

Let's close the issue and PR here and move the discussion to the [core Trac](https://core.trac.wordpress.org/).

Mamaduka:
Closing in favor of https://core.trac.wordpress.org/ticket/61558.

PR #63071, opened 2024-07-03
Section Styles: Prevent flash of variation styles in post editor

### LINKED ISSUE #62668: Outline button variation slow-load styles in TT4
### Description

Likely a bi-product of work done in https://github.com/WordPress/gutenberg/issues/62451, there seems to be a lag during the styling in the editor (on page load) for buttons with the outline variation. (Original issue was fixed, now the lag exists)

1. Go to a post or page and add the buttons block.
2. Create a button, apply Outline variation.
3. See bug.

https://github.com/WordPress/gutenberg/assets/486261/ba34769f-fbc6-4efb-ab4d-aaaed3b23cfd

### Environment info

WordPress 6.6 beta3 (Gutenberg NOT active)
Issue goes away with Gutenberg 18.5 active.

Yes

No

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/62668

## What?

Preloads theme and global styles data from REST API to avoid flash of variation styles in the post editor.

## Why?

The flash of styling looks pretty rough and negatively impacts the experience of using the post editor when there are block style variations applied to blocks in the post.

## How?

- Adds a `block_editor_rest_api_preload_paths` filter to preload everything required for global styles data in the block editor
- Brings in tweak from [#61095](https://github.com/WordPress/gutenberg/pull/61095) to avoid issue with preloading request using `url` option.

## Testing Instructions

1. Active a theme that defines styles for the button outline block style variation e.g. TT4
2. In the post editor add a button, select the Outline style, and save.
3. Reload the post editor and ensure there is no flash of styling (i.e. from a solid color to the outline variant)
4. In the site editor navigate to Styles > Blocks > Button > Outline
5. Apply custom styles for the Outline block style variation e.g. change the border color, background etc.
6. Save the Global Styles changes
7. Switch back to the post editor, loading your earlier post with the button outline block
8. When the editor loads there should be no flash of styling from the default theme outline styles to the custom styles assigned via Global Styles

## Screenshots or screencast 

| Before | After |
|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/60436221/9fb45b16-2544-4c51-9797-36ae7d3bdaa1" /> | <video src="https://github.com/WordPress/gutenberg/assets/60436221/36d94b27-47fe-4132-b19e-ef4a76c416e4" /> |

### DIFF
--- a/lib/compat/wordpress-6.6/rest-api.php
+++ b/lib/compat/wordpress-6.6/rest-api.php
@@ -157,3 +157,20 @@ if ( ! function_exists( 'gutenberg_register_wp_rest_themes_template_directory_ur
 	}
 }
 add_action( 'rest_api_init', 'gutenberg_register_wp_rest_themes_template_directory_uri_field' );
+
+/**
+ * Preload theme and global styles paths to avoid flash of variation styles in post editor.
+ *
+ * @param array                   $paths REST API paths to preload.
+ * @param WP_Block_Editor_Context $context Current block editor context.
+ * @return array Filtered preload paths.
+ */
+function gutenberg_block_editor_preload_paths_6_6( $paths, $context ) {
+	if ( 'core/edit-post' === $context->name ) {
+		$paths[] = '/wp/v2/global-styles/themes/' . get_stylesheet();
+		$paths[] = '/wp/v2/themes?context=edit&status=active';
+		$paths[] = '/wp/v2/global-styles/' . WP_Theme_JSON_Resolver::get_user_global_styles_post_id() . '?context=edit';
+	}
+	return $paths;
+}
+add_filter( 'block_editor_rest_api_preload_paths', 'gutenberg_block_editor_preload_paths_6_6', 10, 2 );
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -543,13 +543,12 @@ export const __experimentalGetCurrentGlobalStylesId =
 		const globalStylesURL =
 			activeThemes?.[ 0 ]?._links?.[ 'wp:user-global-styles' ]?.[ 0 ]
 				?.href;
-		if ( globalStylesURL ) {
-			const globalStylesObject = await apiFetch( {
-				url: globalStylesURL,
-			} );
-			dispatch.__experimentalReceiveCurrentGlobalStylesId(
-				globalStylesObject.id
-			);
+		if ( ! globalStylesURL ) {
+			return;
+		}
+		const id = +globalStylesURL.split( '/' ).pop();
+		if ( id ) {
+			dispatch.__experimentalReceiveCurrentGlobalStylesId( id );
 		}
 	};
 
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -547,8 +547,10 @@ export const __experimentalGetCurrentGlobalStylesId =
 			return;
 		}
 
-		const url = new URL( globalStylesURL );
-		const id = +url.pathname.split( '/' ).pop();
+		// Regex matches the ID at the end of a URL or immediately before
+		// the query string.
+		const matches = globalStylesURL.match( /\/(\d+)(?:\?|$)/ );
+		const id = matches ? Number( matches[ 1 ] ) : null;
 
 		if ( id ) {
 			dispatch.__experimentalReceiveCurrentGlobalStylesId( id );
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -546,7 +546,12 @@ export const __experimentalGetCurrentGlobalStylesId =
 		if ( ! globalStylesURL ) {
 			return;
 		}
-		const id = +globalStylesURL.split( '/' ).pop();
+
+		// Regex matches the ID at the end of a URL or immediately before
+		// the query string.
+		const matches = globalStylesURL.match( /\/(\d+)(?:\?|$)/ );
+		const id = matches ? Number( matches[ 1 ] ) : null;
+
 		if ( id ) {
 			dispatch.__experimentalReceiveCurrentGlobalStylesId( id );
 		}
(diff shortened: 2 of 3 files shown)


packages/core-data/src/resolvers.js
```diff
@@ -543,13 +543,12 @@ export const __experimentalGetCurrentGlobalStylesId =
...
-			const globalStylesObject = await apiFetch( {
-				url: globalStylesURL,
-			} );
-			dispatch.__experimentalReceiveCurrentGlobalStylesId(
-				globalStylesObject.id
-			);
+		if ( ! globalStylesURL ) {
+			return;
+		}
+		const id = +globalStylesURL.split( '/' ).pop();
```
talldan: I'm a little nervous about the `split` and `pop` as it assumes the href from the link will have the same format (for example, it'll not have any query strings). That's not something that this code has any control over, and If it ever had a query string the `id` would end up as `NaN`. Maybe a fairly loose regex would be safer? I guess it needs to take into account that the URL might have some differences on different WordPress installations (I think the base of the rest API is configurable).

Other than that, the PR looks good to me, the added preload paths bring the post editor more in line with the site editor, which would be a good thing to do with the editor unification work.

talldan: I guess we should also follow up with getting https://github.com/WordPress/gutenberg/pull/61095 merged, so replicating whatever changes are made to this code in that PR.

aaronrobertshaw: Thanks 👍 

I've updated this to use a very basic regex to extract the id in https://github.com/WordPress/gutenberg/pull/63071/commits/d96ff045eea11f5577284fe4fe07d9d1827f09c1.

The regex isn't 100% future-proof either, however, so long as the ID is the last segment of the URL before a possible query string, it should be good enough. Are you happy with that?

talldan (approved):
LGTM, thanks for addressing the feedback.

lib/compat/wordpress-6.6/rest-api.php
```diff
@@ -157,3 +157,20 @@ function gutenberg_register_wp_rest_themes_template_directory_uri_field() {
...
+
+/**
+ * Preload theme and global styles paths to avoid flash of variation styles in post editor.
+ *
+ * @param array                   $paths REST API paths to preload.
+ * @param WP_Block_Editor_Context $context Current block editor context.
+ * @return array Filtered preload paths.
+ */
+function gutenberg_block_editor_preload_paths_6_6( $paths, $context ) {
+	if ( 'core/edit-post' === $context->name ) {
```
ellatrix, line 169: Why only post editor?

packages/core-data/src/resolvers.js
```diff
@@ -543,13 +543,17 @@ export const __experimentalGetCurrentGlobalStylesId =
...
-			dispatch.__experimentalReceiveCurrentGlobalStylesId(
-				globalStylesObject.id
-			);
+		if ( ! globalStylesURL ) {
+			return;
+		}
+
+		// Regex matches the ID at the end of a URL or immediately before
+		// the query string.
+		const matches = globalStylesURL.match( /\/(\d+)(?:\?|$)/ );
```
ellatrix, line 552: We could use [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) to at least remove the query?

lib/compat/wordpress-6.6/rest-api.php
talldan: The site editor already has the paths.

ellatrix: Really? Not sure why I was adding it in #61095 then

talldan: Here's the same paths in core for the site editor:
https://github.com/WordPress/wordpress-develop/blob/7159243c090e429a7d2a1fd2a9509e323f67a39d/src/wp-admin/site-editor.php#L97-L100

And your PR removes the one on line 99. (which this PR isn't adding for the post editor)

aaronrobertshaw: As @talldan noted above #61095 was slightly different. Ultimately, this PR brings the two editors closer together regarding preloading global styles paths.

packages/core-data/src/resolvers.js
aaronrobertshaw: I almost reached for `URL` but ended up with the more familiar regex. What's the rationale behind switching to `URL`?

I'm happy to make the change tomorrow, just looking at how I could make better choices in the future.

ellatrix: It just feels more like the right tool for it. It's just a nit, feel free to ignore.

bgardner:
This testing extremely well for me, and with some pretty hefty exploration of Section Styles. (Meaning my testing goes beyond just the Outline button.) LGTM for sure, with the hope it can get into RC3!!!

packages/core-data/src/resolvers.js
aaronrobertshaw: Sounds good. I've updated this via https://github.com/WordPress/gutenberg/pull/63071/commits/3f0f7b54b5b2e334ec0f115955de02a5e5a92d22

aaronrobertshaw:
It looks like the switch to `URL` has caused a related test failure. While it manually tests well, reverting it locally has the tests passing again. I'll investigate further shortly.

aaronrobertshaw:
After a brief look into the e2e failures, I'm not sure we can assume a standard URL to leverage `URL()` in extracting the ID value. Sometimes it's getting a global styles URL in the format of `rest_route=/wp/v2/global-styles/...`.

Given the suggestion was [noted](https://github.com/WordPress/gutenberg/pull/63071#discussion_r1664224056) as a "nit", I've reverted it for now in the interests of ensuring this fix lands for the next 6.6 RC.

aaronrobertshaw:
There is a backport for the PHP changes here available in https://github.com/WordPress/wordpress-develop/pull/6959

aaronrobertshaw:
Cherry-pick and merge to the wp/6.6 branch here:

- https://github.com/WordPress/gutenberg/pull/63111

Issue #63001, opened 2024-07-01
Stabilize Typography Block Supports Keys


andrewserong:
Part of:

* #43241 and
* #43242 

This issues tracks progress towards stabilizing the typography block support. Note that in order to provide backwards compatibility for block plugins out in the wild, fallback support for the existing `__experimental` prefixes should be provided.

Stabilize the following typography block supports:

* `__experimentalFontFamily` → `fontFamily`
* `__experimentalTextDecoration` → `textDecoration`
* `__experimentalFontStyle` → `fontStyle`
* `__experimentalFontWeight` → `fontWeight`
* `__experimentalLetterSpacing` → `letterSpacing`
* `__experimentalTextTransform` → `textTransform`
* ~`__experimentalWritingMode` → `writingMode`~

## Tasks

* [x] Update PHP block support code in `lib/block-supports/typography.php` to use the non `__experimental` prefix, falling back to the `__experimental` prefix if available.
* [x] Update the JS implementations of the typography block supports to point to the non `__experimental` prefix, falling back to the `__experimental` prefix if available.
* [ ] Update the `block.json` schema to support the non `__experimental` prefix
* [ ] Update the core blocks' `block.json` files to use the non `__experimental` prefixes
* [ ] [Update the supports documentation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#typography) to include the new non prefixed versions.

In my opinion, in order to provide the widest possible support for block plugins, I think it could be valuable to provide fallback support for the `__experimental` in the long-term. In other words, when implementing the fallback ideas above, let's do so in a way that's easy to maintain.

carolinan:
@scruffian I am unsure of if writingMode should be stabilized without the text orientation changes? (#60030)

scruffian:
> @scruffian I am unsure of if writingMode should be stabilized without the text orientation changes? (https://github.com/WordPress/gutenberg/pull/60030)

I agree

andrewserong:
Oh, thanks for flagging! I'm working on a draft PR over in https://github.com/WordPress/gutenberg/pull/63401 — I'll omit `writingMode` from the stabilisation for now 👍

ndiego:
This is a fantastic initiative @andrewserong . While we are at it, can we also tackle `__experimentalBorder` and `__experimentalDefaultControls`? I am happy to open an issue for these, but didn't want to jump the gun if you were already planning on doing so. Just let me know either way.

andrewserong:
[...]

PR #63072, opened 2024-07-03
[WIP] Support both experimental and non-experimental typography supports in JS

andrewserong:
## What?

🚧 🚧 🚧 🚧 Early exploration into stabilizing the Typography block supports, not yet ready for testing 🚧 🚧 🚧 🚧 

Part of #63001

Allow support for the following typography block supports in both JS and PHP logic:

* `__experimentalFontFamily` → `fontFamily`
* `__experimentalTextDecoration` → `textDecoration`
* `__experimentalFontStyle` → `fontStyle`
* `__experimentalFontWeight` → `fontWeight`
* `__experimentalLetterSpacing` → `letterSpacing`
* `__experimentalTextTransform` → `textTransform`
* `__experimentalWritingMode` → `writingMode`

## Why?

To enable blocks to opt in to typography supports without needing to use the `__experimental` prefix. At the same time, we need to support any block plugins out in the wild that still use the experimental prefix, so support will likely need to be maintained in perpetuity.

## How?

TBC

## To-do

* [ ] Update functions / logic to ensure that `usePasteStyles` works as expected
* [ ] Update PHP code to use non experimental keys, falling back to the experimental ones
* [ ] Update this PR description with testing instructions
* [ ] Add or update tests

## Testing Instructions

TBC

### DIFF
--- a/packages/blocks/src/store/private-selectors.js
+++ b/packages/blocks/src/store/private-selectors.js
@@ -8,7 +8,7 @@ import { createSelector } from '@wordpress/data';
  */
 import { getBlockType } from './selectors';
 import { getValueFromObjectPath } from './utils';
-import { __EXPERIMENTAL_STYLE_PROPERTY as STYLE_PROPERTY } from '../api/constants';
+import { __EXPERIMENTAL_STYLE_PROPERTY as RAW_STYLE_PROPERTY } from '../api/constants';
 
 const ROOT_BLOCK_SUPPORTS = [
 	'background',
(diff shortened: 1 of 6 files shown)


packages/blocks/src/store/private-selectors.js
```diff
@@ -32,6 +32,43 @@ const ROOT_BLOCK_SUPPORTS = [
...
+	__experimentalTextDecoration: 'textDecoration',
+	__experimentalFontStyle: 'fontStyle',
+	__experimentalFontWeight: 'fontWeight',
+	__experimentalLetterSpacing: 'letterSpacing',
+	__experimentalTextTransform: 'textTransform',
+	__experimentalWritingMode: 'writingMode',
+};
+
+// Stabilize the style property support keys by mapping experimental supports to stable supports.
+const STYLE_PROPERTY = Object.keys( RAW_STYLE_PROPERTY ).reduce(
```
andrewserong, line 46: Note: I've added this here instead of updating the `STYLE_PROPERTY` directly in `api/constants` as unfortunately the constants are exported publicly. If we think it's safe to change, I can update that file directly instead of this part of the change. Here's the original file:

https://github.com/WordPress/gutenberg/blob/02c6a8fa2f8996f1781d4df0e899c73a6de34c5f/packages/blocks/src/api/constants.js#L17-L272

packages/blocks/src/store/private-selectors.js
```diff
@@ -93,6 +130,36 @@ function filterElementBlockSupports( blockSupports, name, element ) {
 	} );
 }
 
+function getStableBlockSupports( blockType ) {
```
andrewserong, line 133: TODO: Add a comment explaining what this does (treats experimental typography keys as non-experimental)

andrewserong:
Update: while this technically works, it's a bit fiddly. @aaronrobertshaw had the good suggestion to try updating `registerBlockType` and the PHP equivalent to migrate the keys from the experimental to the "real" keys.

Specifically, the place to do the processing is likely in `processBlockType` which gets called by `addUnprocessedBlockType`, which is called by `registerBlockType` here: 

https://github.com/WordPress/gutenberg/blob/406f4b7158fa26e7b79cd8aa1caf26213549d421/packages/blocks/src/api/registration.js#L252

https://github.com/WordPress/gutenberg/blob/ad3135529ff7d589de1c25a4e2b39486c3272924/packages/blocks/src/store/process-block-type.js#L77

This work is a _little_ on the backburner for me, so I won't get around to it immediately, but my next step will be try that idea out (likely in a separate PR) so we can explore the different options (hopefully by sometime next week).

andrewserong:
Just chatted with @aaronrobertshaw some more about the ideas here, and for a potential approach on a neater way to handle this. Ideas to try next:

* On the JS side, move the migration logic to `processBlockType` as in the previous comment
* On the PHP side, in the Gutenberg plugin, use the `register_block_type_args` filter to migrate experimental typography support to the stable properties
* In a core PR, propose adding that migration directly to either the `WP_Block_Type_Registry` or the `WP_Block_Type` classes, so that a filter isn't required
* In the JS and PHP code for the block supports, update the hard-coded strings to remove the `__experimental` prefix as the migration will already be handled by the block registration code

andrewserong:
Just closing this out now as I've made a start on an alternative implementation over in #63401, which seems promising. I've only done the JS changes for now, but I'll continue on with the PHP logic next week.

Thanks again for the tips, Aaron! 🙇

Issue #29693, opened 2021-03-09
Allow for persistent unique block attributes


stacimc:
## What problem does this address?
Blocks cannot use attributes to store information that is unique to the block instance but also persisted across page loads. Here's an illustrative example:

I'm working on the Jetpack `Pay with PayPal` block. It contains a `productId` attribute which uniquely identifies the actual product record referenced by the block. When a new block is inserted this attribute will be undefined; we hit the API to create a new product, and store the returned id in this attribute.

When the block is duplicated, I'd like to create a _new_ product record with identical fields (title/price/etc). But because the `productId` is cloned along with all the other attributes, both blocks will point to the same resource. It's possible to keep track of the block's unique `clientId` and reset the product when this value changes — but while this detects block duplication, it also changes every time the post loads, and I only want to reset the product on duplication.

## What is your proposed solution?
One option would be to add support for `unique` attributes, which do not get copied during block duplication. Another option would be to allow blocks to listen for/handle the duplication event.

Tropicalista:
I have similar problem.

Here's my issue: https://github.com/WordPress/gutenberg/issues/29694

I'm creating a form builder. I need a way to store reference to form id in database.  When a block is made reusable it stores  the reference, but if a user convert to normal block I need a way to detach the old reference in DB and create a new one. 

I've just discovered that there's a new `__experimentalIsEditingReusableBlock` that probably can suite for my use case.

ntsekouras:
I've been thinking about this and a possible solution I had in mind to try out involves these: 

1.  Set the unique attribute value on load only if it doesn't exist and mark this change as persistent or not depending on your needs (add undo level). Something like this: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/query/edit/index.js#L94
2. Use the `__experimentalRole` attribute to something like `id` or `unique` etc.., and then use the existing `__experimentalGetBlockAttributesNamesByRole` util to get them.
3. Handle them in the appropriate places, like in the duplicate function and where is needed.

All the above are using experimental APIs though which were introduced here: https://github.com/WordPress/gutenberg/pull/30469. I guess this could be okay though as the new functionality should also be experimental for start.

talldan:
This could also be relevant for the widget editors, which store an internal `widgetId` as an attribute. (cc @kevin940726 @noisysocks).

gziolo:
[...]

Issue #39223, opened 2022-03-04
Unsaved changes should consider image uploads in progress


adamsilverstein:
### Description

When users are uploading media and navigate away, uploads and image display break.

Gutenberg displays an "Unsaved Changes" [warning](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-widgets/src/components/layout/unsaved-changes-warning.js) if users try to navigate away from the editor with unsaved changes (triggered by `window.beforeunload`). The check for unsaved changes misses images that are still being uploaded allowing users to navigate away while WordPress is still uploading or processing an image. 

In progress uploads may never complete regeneration if PHP is able to generate all sub sizes in one go, because our [retry mechanism](https://github.com/WordPress/gutenberg/blob/1df72fc1b3d628b2bc7b13bed74f6b25ab44f82b/packages/api-fetch/src/middlewares/media-upload.js#L51-L62) will not fire. Regardless, Gutenberg will continue marking the image as "in progress" with a spinner when the upload doesn't complete.  

### Step-by-step reproduction instructions

1. Create a new post and publish it.
2. drag in one or more very large images that will take several seconds to be processed
3. before image processing completes, hit update to update the post
4. click view post to navigate away
5. the image may or may show, the srcset will be incomplete (missing sizes)
6. edit the post again, note the images are stuck in an uploading state.

### Screen recording 

https://share.getcloudapp.com/8LuDGjnD

### Screenshots
![image](https://user-images.githubusercontent.com/2676022/168336423-075e02ab-18d2-447d-8d07-8e495483ad57.png)
![image](https://user-images.githubusercontent.com/2676022/168336471-bf99e1dc-9da5-4631-912e-e8192f68129f.png)
![image](https://user-images.githubusercontent.com/2676022/168336516-3fbcd388-af1a-40a2-8dac-0da575362237.png)

### Environment info

- WordPress trunk
- tested with and without Gutenberg plugin (built at trunk)

### Please confirm that you have searched existing issues in the repo.

I did, but I missed this issue which is a duplicate and is now closed - https://github.com/WordPress/gutenberg/issues/16030

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

adamsilverstein:
👋🏼 @ellatrix - I got something working in https://github.com/WordPress/gutenberg/pull/41120 but it is a bit hacky because I ran into a couple of problems. If you can take a look, I would appreciate any tips!

[...]

Issue #55823, opened 2023-11-02
Duplicating Query Block can result in duplicate Query IDs


peterwilsoncc:
### Description

Duplicating the Query Block can result in duplicate query IDs resulting in unexpected enhanced pagination behaviour.

### Step-by-step reproduction instructions

1. Install WP Trunk ( https://github.com/WordPress/wordpress-develop/commit/64bfb23202af5a023d2918bc73ee82421c6bbb63 )
2. Do not enabled Gutenberg
3. Generate some posts via wp cli, `wp post generate`
4. Go to the site editor
5. Edit the home page
6. Remove all group blocks except for the group containing Query block (this is to make it easier to see the effect while testing)
7. Select the query block
8. Ensure "Force page reload" is turned off
9. Add a pagination block if required
10. Select the group block above the query block (ie, the group containing the heading "Watch, Read, Listen")
11. Duplicate the group
12. In the duplicated group, modify the query so it doesn't inherit from the main query
13. Save the template
14. Visit the front end of the site
15. Click next page on either query block
16. Observe the page changes on both blocks

### Environment info

- WordPress Trunk ( https://github.com/WordPress/wordpress-develop/commit/64bfb23202af5a023d2918bc73ee82421c6bbb63 )
- Gutenberg Plugin: Deactivated

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

Mamaduka:
This is a known issue with internal attributes - #29693. Here're two ongoing explorations: #34750 and #38643.

gziolo:
> This is a known issue with internal attributes - #29693. Here're two ongoing explorations: #34750 and #38643.

There is a detailed overview of the current limitations of block attributes provided by @noisysocks in https://github.com/WordPress/gutenberg/discussions/49466. There are multiple ways we could address it, but all the options should have working prototypes so we would need to finally make a decision what is the best way path forward and incorporate them with the Query block to fix the issue.

ryanwelcher:
Wanted to bump this as I just received a report on my Advanced Query Loop plugin for this - https://github.com/ryanwelcher/advanced-query-loop/issues/42

archiveofwork:
I was about to post about this same issue

skorasaurus:
This still occurs however I've worked around this by manually changing the queryid of one of the loops in the code editor.

albanyacademy:
[...]

PR #49456, opened 2023-03-30
Use Maps to pass "constructor args" to blocks from a transform

noisysocks:
## What & why

This is an alternative approach to https://github.com/WordPress/gutenberg/pull/49391. 

Instead of adjusting the block API, we can take advantage of the fact that `transforms` and `edit` exist within the same package (`block-library`) and send "constructor args" to `edit` via a `Map` defined at the module level.

```js
const blobURL = createBlobURL( file );
const block = createBlock( 'core/file' );
blobURLs.set( block.clientId, blobURL );
```

With this approach we can fix a longstanding issue we've had in Gutenberg which is that we use attributes to store "internal" state in certain blocks. For example an undo level is created when dragging-and-dropping an image or file into the editor because the blob URL is stored in the `url` attribute while the file is uploaded. I've included a fix for the File block (61ad7075048e97243228f81b0b7f603b9eed016f) to illustrate how `params` helps.

We can also allow Menu widgets wrapped in a Legacy Widget block to be converted to a Navigation block. This is done by having the transform provide the Navigation block with the menu to convert via a `Map`. See https://github.com/WordPress/gutenberg/issues/47285.

PR #63076, opened 2024-07-03
Block API: Introduce "local" attributes and use it for the image block

youknowriad:
Related #39223 
Related discussion #49466.

## What?

For some blocks (mostly media blocks), we need a way to create the blocks from transforms using temporary files. After the temporary image is inserted, we upload the image to the media library and upon success update the block with the Right URL.

So far, we were doing this using "blob urls" (local urls) in the URL attribute for images. The problem with this approach is that if you're fast enough and save your post while the image is still being uploaded, you might end up with "temporary urls" in the saved post. This should never be the case.

One potential solution that has been considered is to block "saving" when an upload is in progress. I think that's not ideal and doesn't address the root cause where anyone can actually trigger the saving programmatically.

The root cause is that we're actually putting local state (temporary urls) in the block attributes that are meant to be saved because we need immediate feedback (block insertion) when drag and dropping blocks. So a solution I'm proposing in this PR is to introduce the concept of "temporary block attributes":

 - These are attributes that don't get saved into the database. When you call `serialize` for a block, they are actually ignored and do not persist in the HTML. 
 - In this PR, I've updated the image block to use this attribute instead of the previous blob urls.

**Note:** There's probably some undo/related logic that need updating but I'm thinking that most of the time, when we actually set this attribute, we'll also be setting another non temporary attribute, so in most cases, there's no impact on undo/redo.

**Note 2:** I know that this idea has been discussed previously on multiple occasions, feel free to update the PR descriptions and link into any relevant issue or discussion.

## Testing Instructions

1- Throttle your network speed (that way uploading images will take a long time)
2- Drag and drop an image file into the post editor.
3- Save the post immediately.
4- Open the post in a new tab and notice that the inserted image is actually empty.
6- Notice that the image block has been updated properly with the right URL.

### DIFF
--- a/packages/block-library/src/image/block.json
+++ b/packages/block-library/src/image/block.json
@@ -9,7 +9,7 @@
 	"keywords": [ "img", "photo", "picture" ],
 	"textdomain": "default",
 	"attributes": {
-		"temporaryUrl": {
+		"blob": {
 			"type": "string",
 			"isTemporary": true
 		},
--- a/packages/block-library/src/image/block.json
+++ b/packages/block-library/src/image/block.json
@@ -11,7 +11,7 @@
 	"attributes": {
 		"blob": {
 			"type": "string",
-			"isTemporary": true
+			"__experimentalRole": "local"
 		},
 		"url": {
 			"type": "string",
--- a/packages/block-library/src/image/block.json
+++ b/packages/block-library/src/image/block.json
@@ -9,6 +9,10 @@
 	"keywords": [ "img", "photo", "picture" ],
 	"textdomain": "default",
 	"attributes": {
+		"temporaryUrl": {
+			"type": "string",
+			"isTemporary": true
+		},
 		"url": {
 			"type": "string",
 			"source": "attribute",
--- a/packages/block-library/src/utils/hooks.js
+++ b/packages/block-library/src/utils/hooks.js
@@ -48,17 +48,16 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 			return;
 		}
 
-		if ( ! latestArgs.current.temporaryUrl ) {
+		if ( ! latestArgs.current.url ) {
 			return;
 		}
 
-		const file = getBlobByURL( latestArgs.current.temporaryUrl );
+		const file = getBlobByURL( latestArgs.current.url );
 		if ( ! file ) {
 			return;
 		}
 
-		const { temporaryUrl, allowedTypes, onChange, onError } =
-			latestArgs.current;
+		const { url, allowedTypes, onChange, onError } = latestArgs.current;
 		const { mediaUpload } = getSettings();
 
 		hasUploadStarted.current = true;
@@ -71,12 +70,12 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 					return;
 				}
 
-				revokeBlobURL( temporaryUrl );
+				revokeBlobURL( url );
 				onChange( media );
 				hasUploadStarted.current = false;
 			},
 			onError: ( message ) => {
-				revokeBlobURL( temporaryUrl );
+				revokeBlobURL( url );
 				onError( message );
 				hasUploadStarted.current = false;
 			},
--- a/packages/block-library/src/utils/hooks.js
+++ b/packages/block-library/src/utils/hooks.js
@@ -47,8 +47,10 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 		if ( hasUploadStarted.current ) {
 			return;
 		}
-
-		if ( ! latestArgs.current.url ) {
+		if (
+			! latestArgs.current.url ||
+			! isBlobURL( latestArgs.current.url )
+		) {
 			return;
 		}
 
--- a/packages/block-library/src/utils/hooks.js
+++ b/packages/block-library/src/utils/hooks.js
@@ -48,19 +48,17 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 			return;
 		}
 
-		if (
-			! latestArgs.current.url ||
-			! isBlobURL( latestArgs.current.url )
-		) {
+		if ( ! latestArgs.current.temporaryUrl ) {
 			return;
 		}
 
-		const file = getBlobByURL( latestArgs.current.url );
+		const file = getBlobByURL( latestArgs.current.temporaryUrl );
 		if ( ! file ) {
 			return;
 		}
 
-		const { url, allowedTypes, onChange, onError } = latestArgs.current;
+		const { temporaryUrl, allowedTypes, onChange, onError } =
+			latestArgs.current;
 		const { mediaUpload } = getSettings();
 
 		hasUploadStarted.current = true;
@@ -73,12 +71,12 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 					return;
 				}
 
-				revokeBlobURL( url );
+				revokeBlobURL( temporaryUrl );
 				onChange( media );
 				hasUploadStarted.current = false;
 			},
 			onError: ( message ) => {
-				revokeBlobURL( url );
+				revokeBlobURL( temporaryUrl );
 				onError( message );
 				hasUploadStarted.current = false;
 			},
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -388,7 +388,7 @@ Insert an image to make a visual statement. ([Source](https://github.com/WordPre
 -	**Name:** core/image
 -	**Category:** media
 -	**Supports:** align (center, full, left, right, wide), anchor, color (~~background~~, ~~text~~), filter (duotone), interactivity, shadow ()
--	**Attributes:** alt, aspectRatio, caption, height, href, id, lightbox, linkClass, linkDestination, linkTarget, rel, scale, sizeSlug, title, url, width
+-	**Attributes:** alt, aspectRatio, caption, height, href, id, lightbox, linkClass, linkDestination, linkTarget, rel, scale, sizeSlug, temporaryUrl, title, url, width
 
 ## Latest Comments
 
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -388,7 +388,7 @@ Insert an image to make a visual statement. ([Source](https://github.com/WordPre
 -	**Name:** core/image
 -	**Category:** media
 -	**Supports:** align (center, full, left, right, wide), anchor, color (~~background~~, ~~text~~), filter (duotone), interactivity, shadow ()
--	**Attributes:** alt, aspectRatio, caption, height, href, id, lightbox, linkClass, linkDestination, linkTarget, rel, scale, sizeSlug, temporaryUrl, title, url, width
+-	**Attributes:** alt, aspectRatio, blob, caption, height, href, id, lightbox, linkClass, linkDestination, linkTarget, rel, scale, sizeSlug, title, url, width
 
 ## Latest Comments
 
--- a/packages/block-library/src/gallery/edit.js
+++ b/packages/block-library/src/gallery/edit.js
@@ -249,9 +249,9 @@ function GalleryEdit( props ) {
 		const imageArray = newFileUploads
 			? Array.from( selectedImages ).map( ( file ) => {
 					if ( ! file.url ) {
-						return pickRelevantMediaFiles( {
-							url: createBlobURL( file ),
-						} );
+						return {
+							temporaryUrl: createBlobURL( file ),
+						};
 					}
 
 					return file;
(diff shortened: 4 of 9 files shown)


packages/block-library/src/image/block.json
```diff
@@ -9,6 +9,10 @@
 	"keywords": [ "img", "photo", "picture" ],
 	"textdomain": "default",
 	"attributes": {
+		"temporaryUrl": {
+			"type": "string",
+			"isTemporary": true
```
youknowriad: Is this the right way to flag this kind of attributes?

packages/block-library/src/image/block.json
```diff
@@ -9,6 +9,10 @@
 	"keywords": [ "img", "photo", "picture" ],
 	"textdomain": "default",
 	"attributes": {
+		"temporaryUrl": {
```
mtias: maybe call this one `blob`?

packages/block-library/src/image/block.json
mtias: is there a way to represent this more like "don't persist"? Historically we have tried to use state for attributes that are ephemeral. Representing these ephemeral attributes at the block object level seems slightly wrong.

youknowriad: > is there a way to represent this more like "don't persist"?

This is the way to represent "don't persist". 

> Historically we have tried to use state for attributes that are ephemeral. Representing these ephemeral attributes at the block object level seems slightly wrong.

The problem here is that it's not possible to use local state, and that's why the issue persisted for a long time. When you drag and drop a file, you want to create the block immediately (immediate feedback) and creating blocks means setting attributes.

I went back and forth between representing this as an attribute that has a special behavior (don't persist) and a new property on the block object `localState` or something that is basically just like attributes but treated separately.

Ultimately, I think making it an attribute is better because:

 - The impact of introducing a new property on the block object is way too big and the benefits are small
 - We may have other kind of attributes in the future that have different behaviors (for instance attributes that impact undo/redo in a certain way...). In the past we discussed with @noisysocks and @gziolo several kind of other attributes, maybe they recall.

ellatrix:
When did we start doing this? The last thing I remember was that we wait to set attributes until the image is uploaded until we set the URL attribute, and the blob was just a local state. That's how the undo trap was fixed too. I wonder what changed where we put blob URLs in the attributes?

ellatrix:
Super dumb idea:

What if we saved images when the post is saved, and replace URLs server side as images are uploaded and URLs resolved? This would also fix any sort of undo issues.

Mamaduka:
> I wonder what changed where we put blob URLs in the attributes?

The file type block transformations. Currently, it can't be avoided there.

youknowriad:
> When did we start doing this? The last thing I remember was that we wait to set attributes until the image is uploaded until we set the URL attribute, and the blob was just a local state. That's how the undo trap was fixed too. I wonder what changed where we put blob URLs in the attributes?

To be honest, that's very old, almost from the start (when we implemented drag and drop files)

youknowriad:
> What if we saved images when the post is saved, and replace URLs server side as images are uploaded and URLs resolved? This would also fix any sort of undo issues.

Not dumb at all but I think it's way too complex to implement. It would be good if we had a way to "create entities" with temporary ids or things like that and refer to them in other places of our data (things like categories, tags, images, even patterns, ...) but it's a huge and very unclear undertaking.

packages/block-library/src/image/block.json
talldan: You've probably already considered it, but another option is to still use the same `url` attribute and have a way to mark it as not persisting. Maybe even a different version of `setAttributes` function.

I guess ultimately that may still end up being more complicated (both for us and block implementers) than a separate attribute and might not solve the other use cases you mentioned, though it could also work alongside the idea of permanently temporary attributes.

mtias: > and have a way to mark it as not persisting.

Yeah, this was my thinking. Also agreed it might be more complicated than a separate attribute, but it seems less elegant overall.

What about `isLocal` instead of `isTemporary`?

ellatrix:
With this PR, there's a problem on undo: the image goes back to a loading state. This is something that was fixed and works correctly in trunk: the images goes back to the placeholder state on undo, or reverts to the previous image.

Maybe we need to mark any setting of this temporary attribute as non persistent?

image

ntsekouras:
Should we consider the `__experimentalRole` instead of `isTemporary`?

packages/block-library/src/utils/hooks.js
```diff
@@ -48,10 +48,7 @@ export function useUploadMediaFromBlobURL( args = {} ) {
 			return;
 		}
 
-		if (
-			! latestArgs.current.url ||
-			! isBlobURL( latestArgs.current.url )
```
Mamaduka, line 53: Let's restore the `isBlobURL` check. I think it's a good safeguard.

youknowriad:
> Maybe we need to mark any setting of this temporary attribute as non persistent?

That's possible but won't solve the issue here because the issue is that the block is created right away with a non persistent attribute. We want to keep the block there but empty it (like in trunk). I applied the same fix that we have in trunk (using local state)

packages/block-library/src/utils/hooks.js
youknowriad: I'm not sure. If you set "temporaryUrl" you want to explicitly trigger this hook otherwise, why did you use temporaryUrl in the first place?

youknowriad:
> Should we consider the __experimentalRole instead of isTemporary?

I think that's a good point. so `__experimentalRole="local"` or something instead of a separate `isLocal` flag. Any thoughts @mtias ?

packages/block-library/src/utils/hooks.js
Mamaduka: Worse things have happened :D 

The hook expects `args.url` to be a `blob` value. I think an extra check makes sense.

mtias:
Could work.

packages/block-library/src/utils/hooks.js
youknowriad: This is actually a shared hook so you're right about keeping the check. I restored it.

noisysocks:
A simpler solution for _just_ the image block issue is https://github.com/WordPress/gutenberg/pull/49456/commits/37d6855f9a040ce2f623741301941e11d57e1551. Basically we store the blob in a `@wordpress/block-library` module variable.

This PR is the same approach that https://github.com/WordPress/gutenberg/pull/34750 took so it's worth reading the discussion there.

In revisiting it again, I do like this approach and I'm keen to just try it out instead of discussing it again ad nauseam.

You'll need to add a check for `__experimentalRole === 'local'` to the code that handles copying a block so that, if you duplicate an image block while it is in its temporary blob state, you get a new image block showing the placeholder.

I like `local` or `internal` as a name better than `temporary` as attributes like this can be useful for non-temporary things e.g. storing `widgetId` in the widget block or storing `productId` in a Pay with PayPal block. See https://github.com/WordPress/gutenberg/discussions/49466.

That brings me to the real question I want us to consider here:

Sometimes we want an attribute that **is not** serialised to HTML and **is not** copied when you duplicate a block. For example: `blob`, `widgetId`. Other times we want an attribute that **is** serialised to HTML but **is not** copied when you duplicate a block. For example: `productId`.

This PR handles the former but not the latter. Do you think we should, in addition to `__experimentalRole = 'local'` to handle the former, add another `__experimentalRole` to handle the latter? Then we can close https://github.com/WordPress/gutenberg/issues/29693.

youknowriad:
> Sometimes we want an attribute that is not serialised to HTML and is not copied when you duplicate a block. For example: blob, widgetId. Other times we want an attribute that is serialised to HTML but is not copied when you duplicate a block. For example: productId.

> This PR handles the former but not the latter. Do you think we should, in addition to __experimentalRole = 'local' to handle the former, add another __experimentalRole to handle the latter? Then we can close https://github.com/WordPress/gutenberg/issues/29693.

I'm not sure, why would you not want to duplicate a product block with a productID ? I'm wondering if there's something we're missing on the UX side here and that duplication is actually fine here. Do you have more use-cases in mind. 

Overall though, you I think it should be fine to have a `shouldDuplicate = false` or something but to be honest, I have a lot of doubt here. Because can be created in different ways anyway (outside duplication), like programmatically and in that case, it's not possible to handle duplication, so this leave me to think that maybe a better UX here is to actually show something in the block itself that it's a duplicate block, like we do for blocks that should only be used once.

youknowriad:
I'm actually even hesitant to prevent these local attributes from being duplicated. Sure, in our current case it will result in two image blocks with the same images but uploaded twice but isn't that better than an empty image and an uploaded image from a user's perspective.

youknowriad:
For the widgetId and as implemented today, I think it can be seen as a "local" attribute but ultimately, I also wonder if our widgets editor is the right UI for these and I would love if in the future (I know it's not priority) we merge it with template parts as widget areas are template parts are actually the exact same thing, so I can see the site editor's approach to editing template parts used for widget editor too (we could even use the `@wordpress/editor` package in that case). Which makes me think that we shouldn't be optimizing for that use-case for now.

ntsekouras (approved):
I think it's okay to go with this approach and I was in favor of this way back with the [other PR](https://github.com/WordPress/gutenberg/pull/34750). 

Let's 🚢

noisysocks:
> I'm not sure, why would you not want to duplicate a product block with a productID ? I'm wondering if there's something we're missing on the UX side here and that duplication is actually fine here. Do you have more use-cases in mind.

From https://github.com/WordPress/gutenberg/issues/29693:

> I'm working on the Jetpack Pay with PayPal block. It contains a productId attribute which uniquely identifies the actual product record referenced by the block. When a new block is inserted this attribute will be undefined; we hit the API to create a new product, and store the returned id in this attribute.
>
> When the block is duplicated, I'd like to create a new product record with identical fields (title/price/etc). But because the productId is cloned along with all the other attributes, both blocks will point to the same resource. It's possible to keep track of the block's unique clientId and reset the product when this value changes — but while this detects block duplication, it also changes every time the post loads, and I only want to reset the product on duplication.

noisysocks:
OK, but if we don't care about the `productId` use case or the `widgetId` use case then why not go with an approach that doesn't require a new API?

Either something like https://github.com/WordPress/gutenberg/commit/37d6855f9a040ce2f623741301941e11d57e1551 which stores the blob in a module variable.

Or set the `blob` attribute when calling `createBlock()` but don't add `blob` to `block.json`. The block editor currently ignores attributes that aren't in `block.json` when serialising a block. In other words this PR should work fine without the changes to `packages/block-library/image/block.json` and `packages/blocks`.

Mamaduka:
Here's yet another duplication-related issue, #55823.

youknowriad:
@noisysocks I'm not disregarding the duplication issue but I think it's less important and unclear to me still. It feels like a separate concern to me at this point than local attributes.

For the solution you propose, it doesn't work, because we want to be able to create blocks with these temporary attributes from outside the block itself (block transforms...) and `createBlock` function calls sanitize the attributes and remove the unnecessary ones. I think there's value in being explicit about the things that you can pass when creating a block.

noisysocks (approved):
I don't want to make perfect the enemy of good, let's try it. Thanks for listening 😀

gziolo:
I updated the title to use the `local` keyword to reflect the implementation better.

As a follow-up, it would finally rename the field `__experimentalRole` to `role` and document it so folks can use it confidently in 3rd party projects. It seems like a step in the right direction to help resolve several similar issues outlined by @noisysocks in https://github.com/WordPress/gutenberg/discussions/49466 and highlighted in this PR.

adamsilverstein:
> One potential solution that has been considered is to block "saving" when an upload is in progress. I think that's not ideal and doesn't address the root cause where anyone can actually trigger the saving programmatically.

I still feel the UI should reflect the incomplete state of uploading media. If the media shows a spinner and the image is being uploaded, _users should be prevented from publishing a post or from updating a published post_. 

If they do so, the front end of the post will reflect a "broken" state where media is only partially uploaded. Even if revisiting the editor will automatically complete the upload, we should provide a better user experience by not letting users inadvertently publish broken content.

See https://github.com/WordPress/gutenberg/issues/39223

youknowriad:
@adamsilverstein yes, I didn't close that issue because I also think there's value in letting the user know that there are things in progress. I think both things are necessary though. We need the data to be correct, which this PR does, and we need the UI to be controllable by blocks somehow which we don't have a good API for yet.

talldan:
> users should be prevented from publishing a post or from updating a published post.

Just to put another option out there, rather than preventing, the save button could still be clickable, but the save request could wait for the image upload to be complete before initiating. This way it could be almost imperceptible for users, except for saving taking a bit longer while the two requests complete.

fabiankaegy:
Hey @youknowriad  👋

Would you be able to help write a Dev Note for this feature? We are tracking all the dev notes for 6.7 them in this tracking issue: https://github.com/WordPress/gutenberg/issues/65784

We are hoping to have all drafts compiled by October 13th so we have some time to review before RC1

youknowriad:
We should include this one with the "stabilize role dev note" but yeah I can write a section for it.

fabiankaegy:
Good call :) Will group them in the tracking issue  👍

PR #56507, opened 2023-11-24
Input Field Block: Use `useblockProps` hook in save function

t-hamano:
Fixes #56458
Follow up #44214

## What?

This PR adds the missing wrapper div and `blockProps` in the save function of the Input Filed (Form Input) block. This allows block support styles such as margins, borders, etc. to work on the front end.

## Why?
Without this hook, block support styles will not be applied. Also, the markup in the editor and frontend doesn't match.

## How?
At the same time as adding a hook to the save function, I also added a block deprecation.

## Testing Instructions

- Switch to trunk branch.
- Paste the markup below into your code editor and save your post.
  ```html
  <!-- wp:form -->
  <form class="wp-block-form" enctype="text/plain">
  
  	<!-- wp:form-input -->
  	<label class="wp-block-form-input__label">
  		<span class="wp-block-form-input__label-content">Label</span>
  		<input class="wp-block-form-input__input" type="text" name="label" aria-required="false"/></label>
  	<!-- /wp:form-input -->
  
  	<!-- wp:form-submit-button -->
  	<div class="wp-block-form-submit-button">
  		<!-- wp:buttons -->
  		<div class="wp-block-buttons">
  			<!-- wp:button {"tagName":"button","type":"submit"} -->
  			<div class="wp-block-button">
  				<button type="submit" class="wp-block-button__link wp-element-button">Submit</button>
  			</div>
  			<!-- /wp:button -->
  		</div>
  		<!-- /wp:buttons -->
  	</div>
  	<!-- /wp:form-submit-button -->
  </form>
  <!-- /wp:form -->
  ```
- Switch to this PR branch.
- Reload the browser.
- You should see "**Updated Block: core/form-input**" displayed in your browser console.
- Add margin and border to this block and save the post.
- The block should not break even if you reload your browser.
- In the front end, the block support style should be applied correctly.

PR #55755, opened 2023-11-01
Properly apply styles and classes to the experimental form block

aristath:
## What?
Properly applies styles and classes to the experimental form block. In https://github.com/WordPress/gutenberg/pull/44214#issuecomment-1786948891 it shows that styles don't get properly applied to the `core/form` block. This PR fixes that issue.

## Why?
Because it's a bug - or rather something that wasn't properly implemented in the initial PR introducing the experiment.

## How?
* ~Removes the default `style` and `classes` attributes from the `<form>` element, and then adds them using `get_block_wrapper_attributes()`.~

---

Updated by @t-hamano:

- Remove the hard-coded `className` prop in the save function. This means that styles, classes, etc. from useblockProps will be respected, and block support will work properly.
- Add a block deprecation process so as not to break existing blocks in the contents. This deprecation is unusual and intentionally leaves the block support empty. See [this comment](https://github.com/WordPress/gutenberg/pull/55755/files#diff-f5e02f945f8e2ade1188bf021549961746ef35a7b7d1ae6e0c7a43d3b7dce201R13-R23) for the reason.

---

## Testing Instructions
* Enable the forms-blocks experiment, create a post and add a form in there.
* Change the background colors, padding, margin, typography... whatever you want.
* Save the post and visit it on the front (in the editor it was working properly, but not when viewing the post)
* Confirm that your styles were properly applied.

---

Added by @t-hamano:

### Testing for Block Deprecation

- Check out the trunk branch.
- Insert a Form block in the Post Editor. Apply a background color or change the font size.
- Save the post.
- Check out this branch.
- Reload the browser in the Post Editor.
- The block should not break, and your browser console logs that the migration process has been done correctly.
---

### Testing Instructions for Keyboard
Not applicable

PR #63081, opened 2024-07-03
Add stylelint rule to prevent usage of flex-direction reverse values

### LINKED ISSUE #63048: Add stylelint rule to prevent usage of the `*-reverse` CSS flex direction values
### Description

SimilarlySimilarly to https://github.com/WordPress/gutenberg/issues/61241 / https://github.com/WordPress/gutenberg/pull/61243 the CSS flex `row-reverse` and `column-reverse` values for `flex-direction` introduce a mismatch between visual/reading and DOM order.

For more details about why this is a problem for accessibility and the very limited cases where these values can be used, please see https://github.com/WordPress/gutenberg/issues/61241.

To prevent future usages of these values I'd like to porpose to add a new stylelint rule.

Currently, there are 5 occurrences of `-reverse;` in the `*.scss` files in the codebase:

```
packages/block-editor/src/hooks/block-hooks.scss:
  7: 		flex-direction: row-reverse;

packages/block-editor/src/hooks/use-editor-wrapper-styles.native.scss:
  8: 	flex-direction: column-reverse;

packages/block-library/src/form-input/style.scss:
  27: 		flex-direction: row-reverse;

packages/block-library/src/media-text/style.native.scss:
  10: 	flex-direction: row-reverse;
  17: 		flex-direction: column-reverse;
```

All these occurrences should be double checked and if they do alter visual/DOM order, the implementation should be changed.

- Run a search in the codebase for the string `-reverse;`.

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63048

## What?

The `flex-direction` CSS property values `row-reverse` and `column-reverse` should not be used when they make visual order and DOM order mismatch in a way that impacts meaning and interaction. That is, in _almost all cases_ they should not be used. See the issue for more details.

## Why?

For keyboard users and assistive technology users, the visual order and DOM order must match.

## How?

- Adds a `declaration-property-value-allowed-list` rule for `flex-direction` allowing only the following values:
  - "row"
  - "column"
  - "inherit"
  - "initial"
  - "unset"
- The rule will trigger an error with a custom message.
- Refactors the implementation for the experimental `core/form-input` block. The checkbox label is no longer swapped via CSS. There's now a condition to render the checkbox and its label in the expected order. This required to modify the edit and save functions of the block. I'm not too concerned about it as this is still flagged as a Gutenberg experiment.
- Adds 7 disable comments for the rule as follows:

**Valid cases:**
- 2 occurrences in `packages/block-library/src/navigation/style.scss`
- 1 occurrence in `packages/block-library/src/page-list/style.scss`
These flex-direction properties use a CSS variable `--navigation-layout-direction` that would be flagged as an error. To my understanding, the variable value is either 'row', 'column' or 'initial' so it's OK.

**Cases that need refactoring. I will split them in separate issues:**
- 1 occurrence in `packages/block-editor/src/hooks/block-hooks.scss`
- 1 occurrence in `packages/block-editor/src/hooks/use-editor-wrapper-styles.native.scss`
- 2 occurrences in `packages/block-library/src/media-text/style.native.scss`

## Testing Instructions

- Run the CSS linter in your terminal: `npm run lint:css`.
- Observe there are no errors reported.
- Uncomment one of the disable comments added in this PR.
- Run the CSS linter again.
- Observe the reported error in your console.
- In the WP admin, go to Gutenberg > Experiments, and enable the `Form and input blocks` experiment.
- Add a form block to a post.
- Select the form and add a checkbox input.
- Observe the checkbox input comes first and its label after.
- Inspect the source and observe the DOM order matches the visual order.
- Save and go to the front end.
- Observe the checkbox and its label visual order matches the DOM order.


afercia:
@WordPress/gutenberg-core any chance you have time for a review? 🙏🏽

aristath (approved):
LGTM 👍

.stylelintrc.json
```diff
@@ -4,6 +4,20 @@
...
 		"at-rule-no-unknown": null,
 		"comment-empty-line-before": null,
+		"declaration-property-value-allowed-list": [
+			{
+				"flex-direction": [
+					"row",
+					"column",
+					"inherit",
+					"initial",
+					"unset"
```
youknowriad: Is there a way to allow variables as values?

tyxla: We should be able to do it with regex. I think it might be easier if we just list all the possible reverse values and with a negative regex (untested):



```suggestion
					"/-reverse$/",
```

That will also allow using variables as @youknowriad raised above.

.stylelintrc.json
```diff
@@ -4,6 +4,20 @@
...
+				]
+			},
+			{
+				"message": "Avoid the flex-direction reverse values. For accessibility reasons, visual, reading, and DOM order must match. Only use the reverse values when they do not affect reading order, meaning, and interaction,"
```
tyxla: Nit: Did we mean to add a dot instead of a comma?

```suggestion
				"message": "Avoid the flex-direction reverse values. For accessibility reasons, visual, reading, and DOM order must match. Only use the reverse values when they do not affect reading order, meaning, and interaction."
```

.stylelintrc.json
```diff
@@ -18,7 +32,7 @@
 		"property-disallowed-list": [
 			[ "order" ],
 			{
-				"message": "Avoid the order property. For accessibility reasons, visual, reading, and DOM order must match. Only use the order property when it does not affect reading order, meaning, and interaction"
+				"message": "Avoid the order property. For accessibility reasons, visual, reading, and DOM order must match. Only use the order property when it does not affect reading order, meaning, and interaction,"
```
tyxla:
```suggestion
				"message": "Avoid the order property. For accessibility reasons, visual, reading, and DOM order must match. Only use the order property when it does not affect reading order, meaning, and interaction."
```

t-hamano:
> Refactors the implementation for the experimental `core/form-input` block. The checkbox label is no longer swapped via CSS. There's now a condition to render the checkbox and its label in the expected order. This required to modify the edit and save functions of the block. I'm not too concerned about it as this is still flagged as a Gutenberg experiment.

I think it needs to be clarified and documented/agreed upon whether it's okay not to add block deprecation to experimental blocks.

We are attempting to add a block deprecation to the Form block in #55755. See the discussion starting in [this comment](https://github.com/WordPress/gutenberg/pull/55755#issuecomment-1937766583) for more details.

Personally, I think it's out of scope to update the `form-input` block's save function in this PR, regardless of whether a deprecation should be added or not.

For now, I think it's better to just ignore `flex-direction: row-reverse`, like everywhere else.

richtabor:
> That is, in almost all cases they should not be used. See the issue for more details.

What cases are valid?

afercia:
> What cases are valid?

Only the ones where reordering doesn't affect 'meaning and operability'.

WCAG 2.2

1.3.2 Meaningful Sequence (Level A) https://www.w3.org/TR/WCAG22/#meaningful-sequence
When the sequence in which content is presented affects its meaning, a correct reading sequence can be programmatically determined.

2.4.3 Focus Order (Level A) https://www.w3.org/TR/WCAG22/#focus-order
If a Web page can be navigated sequentially and the navigation sequences affect meaning or operation, focusable components receive focus in an order that preserves meaning and operability.

CSS Flexible Box Layout Module Level 1
5.4.1 Reordering and Accessibility
https://www.w3.org/TR/css-flexbox-1/#order-accessibility

.stylelintrc.json
afercia: @tyxla sure but in order to do that we should use `declaration-property-value-disallowed-list` which is already used. It's not clear to me how to add multiple disallowed properties with custom messages under the same Stylelint rule. Any clue?

mirka: > add multiple disallowed properties with custom messages under the same Stylelint rule

This is where we need to first convert the config file from JSON to JS, as well as update the Stylelint dependency to a newer version that supports custom message functions for this rule (#63090).

afercia: +1 to converting it. @mirka would you advise to wait for the conversion to happen or maybe proceed with this PR as is, and then iterate when https://github.com/WordPress/gutenberg/issues/63090 will be solved?

tyxla: > @tyxla sure but in order to do that we should use `declaration-property-value-disallowed-list` which is already used. It's not clear to me how to add multiple disallowed properties with custom messages under the same Stylelint rule. Any clue?

Good point, but I think you should be able to do it with `declaration-property-value-allowed-list`, just need to invert the regex to match anything BUT `row-reverse` and `column-reverse`, for example:

```
"/^(?!(row|column)-reverse).*$/"
```

Haven't tested it, so let me know if that works for you.

tyxla: Otherwise, converting to the JS format makes sense, although it doesn't look like a high-priority task.

afercia:
> just need to invert the regex to match anything BUT `row-reverse` and `column-reverse`, for example:
> 
> ```
> "/^(?!(row|column)-reverse).*$/"
> ```

Thanks @tyxla the regex seems to work nicely. Pushed in the last commit. To test:
- Remove one of the disable comments e.g. [here](https://github.com/WordPress/gutenberg/blob/00f59d5bd3fed5b1ac15bf47aa469bb61f2fcd93/packages/block-editor/src/hooks/block-hooks.scss#L7-L8).
- Run `npm run lint:css`.
- Observe the `*-reverse` value is caught.

Playing devil's advocate: is there any way to catch a variable when the variable resolves to a `*-reverse` value? For example, what if [this variable](https://github.com/WordPress/gutenberg/blob/00f59d5bd3fed5b1ac15bf47aa469bb61f2fcd93/packages/block-library/src/navigation/style.scss#L121) was `row-reverse` or `column-reverse`?

tyxla:
> Thanks @tyxla the regex seems to work nicely. Pushed in the last commit. 

Glad to hear!

> Playing devil's advocate: is there any way to catch a variable when the variable resolves to a *-reverse value? For example, what if [this variable](https://github.com/WordPress/gutenberg/blob/00f59d5bd3fed5b1ac15bf47aa469bb61f2fcd93/packages/block-library/src/navigation/style.scss#L121) was row-reverse or column-reverse?

Not really - stylelint, like most linters, performs static code analysis and has no way to deal with what's inside the dynamic values. This shouldn't be something to worry about, though; a reverse flex-direction in a variable sounds like a particularly rare scenario we'll likely never see.

afercia:
> we'll likely never see.

Hopefully 😆 
Thanks again for the regex. I will wait for a final review and the tests to pass before merging.

t-hamano:
Merging this PR will implicitly add a rule that says "**We don't have to add block deprecation for experimental blocks**", but is that okay?

The reason I'm so particular about this is that, as far as I know, whenever the save function of a form-related block was updated, a block deprecation was added without exception.

And there is still an open PR that has been trying to add a block deprecation for a long time.

- #55755

Personally, I propose to revert the update of the form block's save function from this PR and simply add an ignore style rule.

afercia:
Thanks @t-hamano. I see a `v1` deprecation for the `InputFieldBlock` was already added in https://github.com/WordPress/gutenberg/pull/56507 so it makes sense to me to add a `v2`. 

Reverting the update of the form block's save function wouldn't be great because it needs to be fixed anyways, either in this PR or a new one. Open to other thoughts Cc @aristath @carolinan

t-hamano:
I think it's OK to ship this PR as long as it includes the appropriate block deprecation.

Personally, I'd recommend addressing the block deprecation in a follow-up PR, since the purpose of this PR is to add a stylelint rule.

afercia:
Thanks @t-hamano. In the next weeks I'll be away and won't have time to look into this.
Anyone willing to take care of this PR and a follow-up is very welcome, please do feel free to make it yours.

t-hamano:
@afercia 

I'm happy to take over this PR. In order to get this PR ready to ship sooner, would it be ok to revert the changes in the `form-input/` directory and add `stylelint-disable` instead? I'd also like to address the changes to the Form Input block, including the block deprecation, in a follow-up.

afercia:
@t-hamano as long as it gets fixed soon in a follow-up, no objections from me.

t-hamano:
I have reverted the changes to `form-input/*`. I have also applied [this suggestion](https://github.com/WordPress/gutenberg/pull/63081#discussion_r1684290388).

Can this PR be shipped?

ciampo (approved):
LGTM

Noting that these rules won't be able to catch styles written via CSS in JS.

PR #63083, opened 2024-07-03
Global Styles: Don't display browse styles unless the theme has full styles available

scruffian:
## What?
This removes the "Browse Styles" option from Global Styles when there are no full theme styles available.

## Why?
Without this bug fix the browse styles area still appears but is empty.

## How?
Create a new hook that gets all of the theme variations and then filters out the color and typography only variations.

## Testing Instructions
0. Remove all full style variations from your theme (for example in TT4 remove everything except Onyx)
1. Open the Site Editor
2. Open Global Styles
3. Check that the "Browse Styles" option isn't visible.

## Screenshots or screencast 
Screenshot 2024-07-03 at 12 23 57

### DIFF
--- a/packages/edit-site/src/components/global-styles/screen-root.js
+++ b/packages/edit-site/src/components/global-styles/screen-root.js
@@ -25,6 +25,7 @@ import { IconWithCurrentColor } from './icon-with-current-color';
 import { NavigationButtonAsItem } from './navigation-button';
 import RootMenu from './root-menu';
 import PreviewStyles from './preview-styles';
+import { useThemeStyles } from '../../hooks';
 import { unlock } from '../../lock-unlock';
 
 const { useGlobalStyle } = unlock( blockEditorPrivateApis );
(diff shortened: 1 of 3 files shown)


richtabor:
We should also resolve it here: 
CleanShot 2024-07-03 at 12 08 47

scruffian:
Good catch. Fixed in [fa7a474](https://github.com/WordPress/gutenberg/pull/63083/commits/fa7a474eb45fbc02d00718fd45e0f0d1a972db5a)

richtabor:
With the addition of https://github.com/WordPress/gutenberg/pull/63173, I wonder if this is still necessary. 

Perhaps just removing the default style variation within will be fine instead. 

CleanShot 2024-07-11 at 15 08 20

bgardner:
I agree with @richtabor, per his scresnshot, because as a result of this PR, the entire Styles panel goes away:

image

scruffian:
Replaced by https://github.com/WordPress/gutenberg/pull/63555

PR #63088, opened 2024-07-03
Focal Point Picker: Set grid overlay state when dragging

Mamaduka:
## What?
PR updates the `FocalPointPicker` to set the `showGridOverlay` state via `onDragStart` and `onDragEnd` callbacks.

## Why?
The component has clear events that trigger point updates; we can set related states during these events and avoid effect synchronization.

## Testing Instructions
1. Open a post or page.
2. Insert a Cover block and set an image.
3. Try updating the image focal point.
4. The grid overlay should be displayed as before.

Same.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/240569/8a3c5c05-06f3-44cc-bd00-8ea44d078001

### DIFF
--- a/packages/components/src/focal-point-picker/index.tsx
+++ b/packages/components/src/focal-point-picker/index.tsx
@@ -27,15 +27,12 @@ import {
 	MediaContainer,
 } from './styles/focal-point-picker-style';
 import { INITIAL_BOUNDS } from './utils';
-import { useUpdateEffect } from '../utils/hooks';
 import type { WordPressComponentProps } from '../context/wordpress-component';
 import type {
 	FocalPoint as FocalPointType,
 	FocalPointPickerProps,
 } from './types';
 import type { KeyboardEventHandler } from 'react';
-
-const GRID_OVERLAY_TIMEOUT = 600;
 
 /**
  * Focal Point Picker is a component which creates a UI for identifying the most important visual point of an image.
@@ -116,6 +113,7 @@ export function FocalPointPicker( {
 			}
 
 			onDragStart?.( value, event );
+			setShowGridOverlay( true );
 			setPoint( value );
 		},
 		onDragMove: ( event ) => {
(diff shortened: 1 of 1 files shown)


packages/components/src/focal-point-picker/index.tsx
```diff
@@ -237,15 +236,6 @@ export function FocalPointPicker( {
 	const instanceId = useInstanceId( FocalPointPicker );
 	const id = `inspector-focal-point-picker-control-${ instanceId }`;
 
-	useUpdateEffect( () => {
-		setShowGridOverlay( true );
-		const timeout = window.setTimeout( () => {
-			setShowGridOverlay( false );
-		}, GRID_OVERLAY_TIMEOUT );
```
Mamaduka, line 244: I've omitted this logic from the update. I'm happy to restore it, but I've not noticed a big difference.

stokesman: In general I'm a fan of avoiding effects. The difference this makes is that the grid overlay is not shown when using the inputs to change the focal point.

Possibly there’s a way to rework how the `Grid` visibility changes so that it show/hides without using state and avoid the effect but I'm not sure.

I don’t have much of an opinion on whether it’s okay to omit showing the grid for changes from the inputs.

ciampo:
As @stokesman also mentioned, the grid overlay currently shows any time the value of the focal point changes — that can happen in a few ways:

- user interactions:
  - user clicking on the canvas
  - user dragging the "dot"
  - user changing x / y values in the number inputs
- any other external controlled updates by changing the `value` prop

Currently, in this PR the grid shows only when dragging the dot. Personally, I feel like we should continue to show the grid overlay on all user interactions (including clicks and number input interactions).

Also, out of curiosity — is there a particular reason why we need to rework this logic? As much as I don't like even synchronisation, the current implementation seemed to work well.

Mamaduka:
I totally missed that grid was showing on any other interaction 😅

Got idea of the idea of this refactoring after I fixed a bug in this area.

I agree that with described terms effect sync makes sense.

Thanks for the feedback 🙇

PR #63089, opened 2024-07-03
Compat: Update HTML API with changes from 6.6

dmsnell:
## What?

Ports code from WordPress 6.6 into `lib/compat/wordpress-6.6`.
Ports code from WordPress 6.5 into `lib/compat/wordpress-6.5`.
Ports code from WordPress 6.4 into `lib/compat/wordpress-6.4`.

Because of the breaking change in 6.5 of the Span and Text Replacement, the 6.6 functions still refer to `Gutenberg_HTML_Span_6_5` and `Gutenberg_HTML_Text_Replacement_6_5`.

## Why?

Ensures that code relying on functionality in the HTML API will work on Gutenberg sites running on older WordPress versions.

## How?

Copies files from Core in the `6.6` branch and replaces `WP_` classes that have updates in 6.6 with their `Gutenberg_` counterparts.

## Testing Instructions

Let the tests run. Since nothing calls these classes yet the only thing that should matter is whether a site loads with the plugin installed.

### DIFF
--- a/lib/compat/wordpress-6.4/html-api/class-wp-html-processor.php
+++ b/lib/compat/wordpress-6.4/html-api/class-wp-html-processor.php
@@ -101,16 +101,12 @@ if ( ! class_exists( 'WP_HTML_Processor' ) ) {
 	 *
 	 * The following list specifies the HTML tags that _are_ supported:
 	 *
-	 *  - Containers: ADDRESS, BLOCKQUOTE, DETAILS, DIALOG, DIV, FOOTER, HEADER, MAIN, MENU, SPAN, SUMMARY.
-	 *  - Form elements: BUTTON, FIELDSET, SEARCH.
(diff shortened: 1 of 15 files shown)


lib/compat/wordpress-6.4/html-api/class-wp-html-processor.php
```diff
@@ -658,50 +636,32 @@ private function step_in_body() {
...
 				case '-FIGURE':
-				case '-FOOTER':
-				case '-HEADER':
-				case '-HGROUP':
-				case '-MAIN':
-				case '-MENU':
-				case '-NAV':
-				case '-SEARCH':
-				case '-SECTION':
-				case '-SUMMARY':
```
dmsnell, line 688: I apparently added code from the 6.5 branch when I initially added this file.

ockham (approved):
LGTM! I diffed all files against their counterparts from the respective Core branches, and the diffs looked fine 👍

Issue #63009, opened 2024-07-01
WP 6.6 RC1: Custom blocks stopped working


leoloso:
### Description

In my plugin, I have a CPT that loads custom blocks. After upgrading to v6.6 RC1, all the custom blocks stopped working:

![Screenshot 2024-07-01 at 16-20-13 Add New GraphQL Schema Configuration ‹ Gato GraphQL — WordPress](https://github.com/WordPress/gutenberg/assets/1981996/eb1cd8c2-5a51-4d2a-9cca-9c8ab2867fe7)

The compiled blocks are targeting WordPress 5.4, they work all the way down there. If possible, I'd expect them to still work on WP 6.6.

Error in the JS console:

```
Uncaught TypeError: L.a is undefined
    <anonymous> https://gatographql-pro.lndo.site/wp-content/plugins/gatographql/blocks/schema-config-composable-directives/build/index.js?ver=7bc47881f5458ab13f7b8b29b9ee3745:1
    a https://gatographql-pro.lndo.site/wp-content/plugins/gatographql/blocks/schema-config-composable-directives/build/index.js?ver=7bc47881f5458ab13f7b8b29b9ee3745:1
    n https://gatographql-pro.lndo.site/wp-content/plugins/gatographql/blocks/schema-config-composable-directives/build/index.js?ver=7bc47881f5458ab13f7b8b29b9ee3745:1
    <anonymous> https://gatographql-pro.lndo.site/wp-content/plugins/gatographql/blocks/schema-config-composable-directives/build/index.js?ver=7bc47881f5458ab13f7b8b29b9ee3745:1
    <anonymous> https://gatographql-pro.lndo.site/wp-content/plugins/gatographql/blocks/schema-config-composable-directives/build/index.js?ver=7bc47881f5458ab13f7b8b29b9ee3745:1
```

I did `npm start` to get a friendlier error msg:

```
Uncaught TypeError: _babel_runtime_regenerator__WEBPACK_IMPORTED_MODULE_0___default.a is undefined
    js resolvers.js:60
    Webpack 14
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        js
        __webpack_require__
        checkDeferredModules
        <anonymous>
        <anonymous>
[...]

PR #63091, opened 2024-07-03
Dependency extraction: Map to `regenerator-runtime` instead of `wp-polyfill`

swissspidy:
## What?

This is a follow-up to https://core.trac.wordpress.org/ticket/60962 in WP 6.6, where `regenerator-runtime` was removed as a dependency from `wp-polyfill`. Now it needs to be added explicitly if still needed.

## Why?

Via #63009 I realized that `@WordPress/dependency-extraction-webpack-plugin` still points to `wp-polyfill` though when using regenerator (which shouldn't happen when targeting newer browsers anyway).

## How?

Externalizes regenerator to `regenerator-runtime` directly instead of `wp-polyfill`

## Testing Instructions

If using generators in a script and targeting older browsers such as IE 11, `regenerator-runtime` would be added as a dependency instead of `wpo-polyfill`.

N/A

## Screenshots or screencast 

N/A

### DIFF
--- a/packages/dependency-extraction-webpack-plugin/lib/util.js
+++ b/packages/dependency-extraction-webpack-plugin/lib/util.js
@@ -117,7 +117,7 @@ function defaultRequestToExternalModule( request ) {
 function defaultRequestToHandle( request ) {
 	switch ( request ) {
 		case '@babel/runtime/regenerator':
-			return 'wp-polyfill';
+			return 'regenerator-runtime';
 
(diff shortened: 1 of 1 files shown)


youknowriad:
This changes make sense to me. Should we backport this one into 6.6?

youknowriad (approved):

swissspidy:
Ya let‘s do that so its published earlier 👍

PR #63093, opened 2024-07-03
Fix: Removed shuffle button when only 1 pattern is present

### LINKED ISSUE #63069: Block Pattern: Shuffle button appears in toolbar even when no patterns to shuffle to in WP 6.6 RC 2
### Description

When you insert an unsynced block pattern that has a category, it can still show the shuffle button, even if there are no other patterns that he button will load.

1. Create a simple unsynced pattern with a unique category; make sure the contents of the group are wrapped in a single group block.
2. Using the `+` menu in the top left, insert the pattern (I ran into [a separate issue](https://github.com/WordPress/gutenberg/issues/63068) where the shuffle button wouldn't appear when the pattern was inserted with `/`).
3. Note the shuffle button appears, but when you click it, nothing seems to happen.

https://github.com/WordPress/gutenberg/assets/177561/f877d9b3-db4b-4603-9542-19fb3c741882

### Environment info

- WordPress 6.6 RC 2 (Gutenberg plugin not installed)
- Twenty Twenty Four theme

Yes

Yes

Sourav61:
Fixes https://github.com/WordPress/gutenberg/issues/63069

## What?
Removes the shuffle button when only 1 pattern is present in the same category

## Why?
Currently, if we have a single pattern having a single category then we are seeing the shuffle button in the toolbar which shouldn't be there as there is no other pattern to go to.

## How?
I have updated the condition to display the shuffle button only if we have 2 or more unsynced patterns of the same category.

## Testing Instructions

1. Create a simple unsynced pattern with a unique category, make sure the contents of the group are wrapped in a single group block.
2. Using the + menu in the top left, insert the pattern.
3. The shuffle button would not appear.

### DIFF
--- a/packages/block-editor/src/components/block-toolbar/shuffle.js
+++ b/packages/block-editor/src/components/block-toolbar/shuffle.js
@@ -66,7 +66,8 @@ export default function Shuffle( { clientId, as = Container } ) {
 			);
 		} );
 	}, [ categories, patterns ] );
-	if ( sameCategoryPatternsWithSingleWrapper.length === 0 ) {
+
+	if ( sameCategoryPatternsWithSingleWrapper.length < 2 ) {
 		return null;
 	}
 


talldan (approved):
Thanks for the contribution @Sourav61, this fix is testing well 👍

noisysocks:
Tested this against `wp/6.6`, looks good 👍

ellatrix:
Manually backported in 07d5d60e1b1ba2519cba9e1f53ad4d7b9960d282

Issue #63077, opened 2024-07-03
block.json: `render` files not copied to `build/` folder for Core blocks


ockham:
### Description

While working on https://github.com/WordPress/gutenberg/pull/62092, I [discovered](https://github.com/WordPress/gutenberg/pull/62092#issuecomment-2189100515) that if a Core block's `block.json` file contains a `render` field that specifies a PHP file to render the block, that file is not copied to the `build/` folder. This means that if a Core block were to use the `render` field, it would break.

The same is not a problem for third-party blocks, as they are handled by code in `wp-scripts` (added by https://github.com/WordPress/gutenberg/pull/43917 and amended by https://github.com/WordPress/gutenberg/pull/51162).

By contrast, Core blocks are built using a special Webpack config in; while there is [code to copy their `index.php` files to the `build/` folder](https://github.com/WordPress/gutenberg/blob/90bc142277c0eee8a1d4d17a380dcccd8e754db2/tools/webpack/blocks.js#L121-L220), there's none to handle `render` files.

This is currently reflected by the fact that no Core blocks are using the `render` field. However, it is somewhat confusing that they cannot use a documented API that is available for third-party blocks.

### Step-by-step reproduction instructions

The following patch moves the Template Part block's render callback from its `index.php` to a newly created `render.php`, and adds a `render` field to `block.json` that points to that file.

Apply the patch and rebuild Gutenberg. Note that Template Part blocks no longer work on the frontend (they appear blank), and that the `render.php` file is missing from `build/block-library/blocks/template-part/`.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

ockham:
The same problem will also apply to the `variations` field, if we extend it to allow filenames (i.e. `variations.php`), see https://github.com/WordPress/gutenberg/pull/62092.

PR #63098, opened 2024-07-03
Scripts: Include variations paths in build

ockham:
## What?
If a `block.json`'s `variations` field is set to a (PHP) file name, include that file in the build, i.e. prefix its functions with `gutenberg_` and copy the resulting file to the `build/block-library/blocks/` directory.

Note that this applies to third-party blocks, but not Core blocks. Making it work for Core blocks is a separate issue: #63077.

## Why?
As of #62092, it is possible to set the `variations` field in a block's `block.json` file to the name of a PHP file that generates the block's variations. 

## How?
By extending the existing logic introduced for the `render` field (by #43917 and #51162) to encompass `variations`.

## Testing Instructions

- On this branch, apply the patch below to create a new subdirectory that contains a basic example block that has a `block.json` file with a `variations` field that points to a PHP file.
- `cd example-variations/`
- Run `../node_modules/.bin/wp-scripts build`.
- Verify that `src/variations.php` has been copied to the `build/` folder.
- Set the `variations` field in `src/block.json` to a variations array: `"variations": [ { "name": "warning" } ]`.
- Run `../node_modules/.bin/wp-scripts start` to begin development in watch mode.
- There should be _no_ `variations.php` file in the `build/` folder.
- Set the `variations` field in `src/block.json` back to `"variations": "file:./variations.php"`.
- The `variations.php` file should re-appear in the `build/` folder.

<details>
<summary>
Patch
</summary>

```diff
diff --git a/example-variations/package.json b/example-variations/package.json
new file mode 100644
index 00000000000..5a39f05d9ab
--- /dev/null
+++ b/example-variations/package.json
@@ -0,0 +1,8 @@
+{
+	"name": "example-variations",
+	"version": "0.1.0",
+	"description": "Example block scaffolded with Create Block tool.",
+	"author": "The WordPress Contributors",
+	"license": "GPL-2.0-or-later",
+	"scripts": {}
+}
diff --git a/example-variations/src/block.json b/example-variations/src/block.json
new file mode 100644
index 00000000000..200e3fa493f
--- /dev/null
+++ b/example-variations/src/block.json
@@ -0,0 +1,5 @@
+{
+	"name": "example/variations",
+	"title": "Block with Variations",
+	"variations": "file:./variations.php"
+}
diff --git a/example-variations/src/variations.php b/example-variations/src/variations.php
new file mode 100644
index 00000000000..bed66d95441
--- /dev/null
+++ b/example-variations/src/variations.php
@@ -0,0 +1,10 @@
+<?php
+
+return array(
+	array(
+		'name'        => 'warning',
+		'title'       => 'warning',
+		'description' => 'Shows warning.',
+		'keywords'    => array( 'warning' ),
+	),
+);
```
</details


ockham:
To make this easier to test, we might want to consider fixing https://github.com/WordPress/gutenberg/issues/63077 first. This will allow us to switch an existing Core block that currently uses `variation_callback` to setting its `block.json`'s `variations` field to a PHP filename instead.

Edit: It's a bit of a chicken-egg problem. Generally, unit-testing this is hard. Alternatively, we've considered using `bin/test-create-block.sh` as an alternative. 
 
That script uses existing templates from the `@wordpress/create-block` package (e.g. `example-static-es5` and `example-static`). We would need to add a new template for a block that set the `variations` field to a PHP file (or modify an existing template).  It's probably also not worth changing the example blocks that ship with that package, so we'll just have to test this PR manually.

packages/scripts/config/webpack.config.js
```diff
@@ -50,23 +50,26 @@ const hasExperimentalModulesFlag = getAsBooleanFromENV(
...
+	static paths;
 
 	apply( compiler ) {
 		const pluginName = this.constructor.name;
 
 		compiler.hooks.thisCompilation.tap( pluginName, () => {
-			this.constructor.renderPaths = getRenderPropPaths();
+			this.constructor.paths = getPhpFilePaths( [
+				'render',
+				'variations',
```
ockham: Maybe obtain these from constructor argument?

packages/scripts/config/webpack.config.js
```diff
@@ -321,7 +324,7 @@ const scriptConfig = {
 				cleanStaleWebpackAssets: false,
 			} ),
 
-		new RenderPathsPlugin(),
+		new PhpFilePathsPlugin(),
```
ockham:
```suggestion
		new PhpFilePathsPlugin( [ 'render', 'variations' ] ),
```

packages/scripts/config/webpack.config.js
ockham: No, let's follow the example of the functions in `block-json.js` and keep these hard-wired.

packages/scripts/utils/config.js
```diff
@@ -354,6 +354,16 @@ function getWebpackEntryPoints( buildType ) {
  * @return {Array}  The list of all the `render` prop paths included in `block.json` files.
  */
 function getRenderPropPaths() {
+	return getPhpFilePaths( [ 'render' ] );
+}
```
SantosGuillamot, line 358: I assume we can't remove this function, right? Are these utils public? It seems `getRenderPropPaths` is not used anywhere internally.

packages/scripts/utils/config.js
```diff
@@ -367,20 +377,29 @@ function getRenderPropPaths() {
 
 	const srcDirectory = fromProjectRoot( getWordPressSrcDirectory() + sep );
 
-	const renderPaths = blockMetadataFiles.map( ( blockMetadataFile ) => {
-		const { render } = JSON.parse( readFileSync( blockMetadataFile ) );
-		if ( render && render.startsWith( 'file:' ) ) {
+	return blockMetadataFiles.flatMap( ( blockMetadataFile ) => {
```
SantosGuillamot, line 380: Asking from ignorance, do we need to use `flatMap` instead of `map`? We are always dealing with strings after [this check](https://github.com/WordPress/gutenberg/pull/63098/files#diff-db38fb177aab60f25e5ed4bfc8b5ce6d371b2e17d96a465a04d89dab56aa8736R386), right? I don't know the implications it could have.

packages/scripts/config/webpack.config.js
```diff
@@ -321,7 +343,7 @@ const scriptConfig = {
 				cleanStaleWebpackAssets: false,
 			} ),
 
-		new RenderPathsPlugin(),
+		new PhpFilePathsPlugin( { props: [ 'render', 'variations' ] } ),
```
SantosGuillamot, line 346: Do we expect to have something apart from `props` in the future? I'm asking because I was wondering if we should just pass an array with the properties, but maybe having the object with `props` is clear. I don't have an opinion, just sharing random thoughts 😄

SantosGuillamot (approved):
I've been testing it and everything seems to work as expected 🙂 I left a few comments with the goal of learning more than considering them issues.

I've been testing it in an external plugin pointing to this version of `wp-scripts` and this is what I tested:
* Using arrays in `variations` keep working as expected.
* Using `variations` to point to a PHP file works as expected. I can see it include in the `build` folder and variations appear in the editor inserter.
* They work both with `build` and `start`.
* The `render` file keeps working as expected.
* I can include both a `render` file and a `variations` file and they work as expected.
* If I point a different property apart from `render` or `variation` to a PHP file, it isn't built.
* Other non-PHP files like `view` or CSS work as expected.

packages/scripts/config/webpack.config.js
ockham: Yeah, it's kind of an ugly signature 😅 I tried to stick with the format that I've seen in [Webpack's docs on writing a plugin](https://webpack.js.org/contribute/writing-a-plugin/), where the plugin constructor accepts an `options` object (i.e. with named properties).

On a more pragmatic note, I actually have a WIP PR (#63311) that would another argument, so it might be a good idea to stick to this format (even if it's a bit ugly 😉).

packages/scripts/utils/config.js
ockham: Previously, we were dealing with one loop: We iterated over all `block.json` files, and from each of them, we extracted the `render` file (if there was one; otherwise, we set the path to `false`; we actually had to remove those values at the end, which is what the `renderPaths.filter( ( renderPath ) => renderPath );` was for).

But now, there's a second, nested loop: In each `block.json` file, we iterate over all props that could contain PHP filenames (currently `render` and `variations`), and if they're present, we keep track of their values. So at the `block.json` file level, we get an array that's either empty (if neither a `render` nor a `variations` field are found), or contains one element (if one of those fields are present), or two (if both are present).

This means that we end up with an array of arrays, which we then need to flatten. This is what the `flatMap` is for. (A welcome side effect is that we no longer need the `filter`, as this is now handled by flattening the empty arrays that we're getting if neither field is present in a `block.json` file).

packages/scripts/utils/config.js
ockham: Good question. The function isn't exported from the `@wordpress/scripts` package, but it seems like we have some precedent of recommending importing functions from a subdirectory of that package, e.g. [here](https://github.com/WordPress/gutenberg/blob/8112ec2e82b6968f8699e672ae1a4b22897c58e8/packages/dependency-extraction-webpack-plugin/README.md#webpack):

> ```js
> const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
> ```

This is the reason why I didn't remove it, however slim chances are that someone actually does this with `getRenderPropPaths`.

However, your comment reminded me of the [`@wordpress/deprecated` package](https://github.com/WordPress/gutenberg/blob/8112ec2e82b6968f8699e672ae1a4b22897c58e8/packages/deprecated/README.md). Maybe I'll use that to annotate the function as deprecated, which should pave the way for removing it at a later time 🤔

packages/scripts/utils/config.js
SantosGuillamot: Thanks a lot for taking the time to provide the context 🙂

packages/scripts/config/webpack.config.js
SantosGuillamot: Great! That makes sense to me 🙂

packages/scripts/utils/config.js
SantosGuillamot: Okay 🙂 I agree probably nobody is using `getRenderPropPaths` but the "compatibility" code is not complex either.

ockham: Ah, it's a bit of a chicken-egg situation 😬 Seems like we can't `require( '@wordpress/deprecated' )` in `@wordpress/scripts`, as the latter package is needed to build the former 😅

ockham: I might just keep it as-is for now (to ensure I'm not breaking any obscure third-party thing that's relying on this).

gziolo: You should be able to remove `getRenderPropPaths`. We don't offer support for internal utils. I also don't expect this one would ever be necessary for 3rd party projects.

Overall, the idea always was that developers use mostly the scripts, but we started supporting also the default webpack config as it's where people tend to apply the changes the most often.

aicuaboom:
👾

luisherranz:
This is just a bit of a random thought but, is adding a field for each type of PHP file that might be needed the best path forward, instead of adding a more solution towards including any type of PHP file?

```json
{
  "serverIncludes": [
    "file:./variations.php",
    "file:./components.php",
    "file:./functions.php",
    "..."
  ]
}
```

I'm not sure how well adding individual fields _per file type_ is going to scale if there are similar needs in the future.

Again, just a random thought so happy to hear your opinions 🙂

ockham:
> This is just a bit of a random thought but, is adding a field for each type of PHP file that might be needed the best path forward, instead of adding a more solution towards including any type of PHP file?
> 
> ```json
> {
>   "serverIncludes": [
>     "file:./variations.php",
>     "file:./components.php",
>     "file:./functions.php",
>     "..."
>   ]
> }
> ```
> 
> I'm not sure how well adding individual fields _per file type_ is going to scale if there are similar needs in the future.
> 
> Again, just a random thought so happy to hear your opinions 🙂

Ah, TBH, that didn't really occur to me. I'm not sure we need (or should expect?) that many other server-side includes; plus it could be a bit tricky to have them depend on one another (e.g. `components.php` might require `functions.php`).

The most compelling argument IMO is that for both `render` and `variations`, we had `render_callback` and `variation_callback` already. `render` arguably represents the server-side interface of a dynamic block, and has a well-defined "signature" (the file -- much like the callback -- [has access to](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-metadata/#render) `$attributes`, `$content`, and `$block` arguments, and is expected to `echo` the markup for the dynamic block).

In a similar vein, there's a defined requirement for `variations` (and `variation_callback`) to `return` an array of block variations (with a defined schema).

AFAICT, the same cannot be said of more generic functions or components (as they aren't directly interfacing with the block). So for the time being, I don't see the need to change the established fields we're using for `render` and `variations`. We might however revisit if we find other use cases 😊

packages/scripts/utils/config.js
ockham: Alright, I'll file a quick PR! Thank you for weighing in 😄

ockham: https://github.com/WordPress/gutenberg/pull/63661

luisherranz:
Makes sense. Thanks for the extra context, @ockham 🙂

PR #63100, opened 2024-07-03
Group block: Simplify variations' isActive fields

ockham:
## What?
Simplify active variation detection for the Group block.

## Why?
To simplify code and showcase recent enhancements to active block variation detection.
It _might_ also help with fixing https://github.com/WordPress/gutenberg/issues/41303.

## How?
By leveraging changes to active variation detection from https://github.com/WordPress/gutenberg/pull/62088 and https://github.com/WordPress/gutenberg/pull/62031, which allow dot notation for nested attributes (e.g. `layout.type`), and picks the variation with the highest specificity if multiple variations match the given `isActive` criteria.

## Testing Instructions
Insert multiple Group blocks, choosing a different variation each time. Save your post and reload the editor. Verify that the block inspector reflects the correct variation for each block.

unnamed

### DIFF
--- a/packages/block-library/src/group/variations.js
+++ b/packages/block-library/src/group/variations.js
@@ -53,11 +53,6 @@ const variations = [
 		attributes: { layout: { type: 'constrained' } },
 		isDefault: true,
 		scope: [ 'block', 'inserter', 'transform' ],
-		isActive: ( blockAttributes ) =>
-			! blockAttributes.layout ||
-			! blockAttributes.layout?.type ||
-			blockAttributes.layout?.type === 'default' ||
-			blockAttributes.layout?.type === 'constrained',
 		icon: group,
 	},
 	{
@@ -66,10 +61,7 @@ const variations = [
 		description: __( 'Arrange blocks horizontally.' ),
 		attributes: { layout: { type: 'flex', flexWrap: 'nowrap' } },
 		scope: [ 'block', 'inserter', 'transform' ],
-		isActive: ( blockAttributes ) =>
-			blockAttributes.layout?.type === 'flex' &&
-			( ! blockAttributes.layout?.orientation ||
-				blockAttributes.layout?.orientation === 'horizontal' ),
+		isActive: [ 'layout.type' ],
 		icon: row,
 		example,
 	},
@@ -79,9 +71,7 @@ const variations = [
 		description: __( 'Arrange blocks vertically.' ),
 		attributes: { layout: { type: 'flex', orientation: 'vertical' } },
 		scope: [ 'block', 'inserter', 'transform' ],
-		isActive: ( blockAttributes ) =>
-			blockAttributes.layout?.type === 'flex' &&
-			blockAttributes.layout?.orientation === 'vertical',
+		isActive: [ 'layout.type', 'layout.orientation' ],
 		icon: stack,
 		example,
 	},
@@ -91,8 +81,7 @@ const variations = [
 		description: __( 'Arrange blocks in a grid.' ),
 		attributes: { layout: { type: 'grid' } },
 		scope: [ 'block', 'inserter', 'transform' ],
-		isActive: ( blockAttributes ) =>
-			blockAttributes.layout?.type === 'grid',
+		isActive: [ 'layout.type' ],
 		icon: grid,
 		example,
 	},


Mamaduka:
@ockham, do you mind rebasing this on top of the latest trunk?

ockham:
> @ockham, do you mind rebasing this on top of the latest trunk?

Done!

Mamaduka (commented):
Thanks, @ockham! Everything works as before ✅

michalczaplinski:
I'm seeing a minor issue but I'm not 100% sure if I'm using the feature correctly. This is my variation registration code in my custom block:

```js
// `index.js` of my custom block

import { registerBlockVariation } from "@wordpress/blocks";
import { __ } from "@wordpress/i18n";
import domReady from "@wordpress/dom-ready";

domReady(() => {
	registerBlockVariation("core/group", {
		name: "wpcomsp/marquee-content",
		title: "Marquee Content",
		icon: "editor-table",
		description: __("A marquee content block."),
		scope: ["block", "inserter"],
		attributes: {
			isMarquee: true,
			layout: { type: "flex", flexWrap: "nowrap" },
			className: "wpcomsp-marquee-content",
			allowedBlocks: [
				"core/paragraph",
				"core/heading",
				"core/image",
				"core/buttons",
			],
		},
		isActive: ["layout.type", "isMarquee"],
	});
});
```

I can see the `Marquee Content` block in the inserter, but upon inserting I get the Row block instead. 

Curiously, when I **increase the specificity** by adding an additional item in the `isActive` array like `"layout.flexWrap"`, it seems to work correctly for the Marquee Content block:

```js
  isActive: ["layout.type", "layout.flexWrap", "isMarquee"],
```

**But**, trying to insert a Row block ends up inserting a Marquee Content instead.

ockham:
> I can see the `Marquee Content` block in the inserter, but upon inserting I get the Row block instead.

Oh, that's very weird, and somewhat concerning -- the block variation inserted being different from the one selected in the inserter seems like there's a very weird disconnect happening.

I'm curious, what happens if you try using your code with GB `trunk`?

Mamaduka:
I did more testing and noticed that normal Group variation (`layout.type === 'constrained'`) is no longer marked as active in the sidebar, while it displays correct labels.

ockham:
> I did more testing and noticed that normal Group variation (`layout.type === 'constrained'`) is no longer marked as active in the sidebar, while it displays correct labels.

Ah, good spot! I opted against setting `isActive` to `[ 'layout.type' ]` as the original criterion also allowed for another value (`default`), as well as for the nested attribute to be missing altogether. I was expecting that `isDefault` would be used to determine the fallback if no variation matches, but that's apparently not the case.

I'm thinking to implement the latter behavior (potentially in a separate PR).

ockham:
> I'm thinking to implement the latter behavior (potentially in a separate PR).

https://github.com/WordPress/gutenberg/pull/63858

michalczaplinski (dismissed):
This makes sense!

- If there **is** a match for the `isActive` field, we return that match. 
- If there **isn't** a match for the `isActive` fiend, we return the variation with the `isDefault` flag.

I don't see any flaws in that logic 🙂. I would have also expected the `isDefault` to work this way.

In a follow-up, it would be nice to log a warning if multiple blocks accidentally have the `isDefault` flag. With the current implementation, the first variation in the array would be returned, which is reasonable behavior.

michalczaplinski (commented):
Sorry, I have to take back my approval as this feedback was meant for your other PR! the good news is that I'm approving your other PR 😅 .

packages/block-library/src/group/variations.js
```diff
@@ -66,10 +61,7 @@ const variations = [
 		description: __( 'Arrange blocks horizontally.' ),
 		attributes: { layout: { type: 'flex', flexWrap: 'nowrap' } },
 		scope: [ 'block', 'inserter', 'transform' ],
-		isActive: ( blockAttributes ) =>
-			blockAttributes.layout?.type === 'flex' &&
-			( ! blockAttributes.layout?.orientation ||
-				blockAttributes.layout?.orientation === 'horizontal' ),
+		isActive: [ 'layout.type' ],
```
Mamaduka, line 64: Won't this conflict with the Grid variation, since now that we have "same" active condition?

ockham: It shouldn't, since the active variation detection logic will compare the _value_ of this attribute (as specified by the variation's `attributes` definition). For the Row variation, `layout.type` is `flex` (see two lines above); for the Grid variation, it's [`grid`](https://github.com/WordPress/gutenberg/pull/63100/files#diff-1076d9cb26ab701d1cbb5d6a12a12eae41f28269a67b200890703bd48ef7df2aR82).

(For the Column variation, where `layout.type` is `flex`, i.e. the same as for the Row, we supply an additional criterion -- `layout.orientation`, which needs to be `vertical` -- to increase specificity and thus avoid collisions.)

ockham:
@Mamaduka Think this is good to land?

Mamaduka (approved):
Thanks, @ockham!

I just re-tested the Group variations and everything works as before ✅

Issue #61099, opened 2024-04-25
Components: Document our deprecation procedures


mirka:
## What problem does this address?

In the past, there has been a lot of ambiguity in our back compat policy. We've been working to bring more clarity around these issues, through efforts like [Private APIs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-private-apis/) and removing `__experimental` prefixes (#59418).

Here, I'd like to increase clarity around component deprecations. The goal of this is twofold:

- Minimize uncertainty and disruption for **extenders**, by setting clearer expectations.
- Simplify the decision-making process for **maintainers**, by having unambiguous guidelines.

## What is your proposed solution?

A "deprecation", at the very basic level, means that we don't encourage new usages of the component.

There are two reasons we may feel this way:

- **Reduce maintenance cost**. Some components are not used enough to justify their continued maintenance. In general, anything that is marked as deprecated will no longer receive enhancements or bug fixes.
- **Encourage use of a better alternative**. There may be a new, improved version in the component library. Or, it's now easier to use standard HTML elements and CSS, rather than a component that abstracted or polyfilled it.

If there is no big harm in keeping existing usages, we will be as unobtrusive as possible about the deprecation ("passive deprecation"). On the other hand, if there are strong enough reasons to discourage continued usage, we will also log a deprecation warning in the console and write a Dev Note ("soft deprecation"). Additionally, if the bundle size or maintenance cost justifies the eventual removal of the code, the deprecation warning will include an end version ("hard deprecation").

If circumstances change, a component may later progress into a stronger level of deprecation.

#### Passive deprecation

- Mark with TypeScript [`@deprecate`](https://www.typescriptlang.org/play/?#example/jsdoc-deprecated)
- Update docs (readme, JSDoc, Storybook)

#### Soft deprecation

- Mark with TypeScript `@deprecate`
- Update docs (readme, JSDoc, Storybook)
- [Dev note](https://make.wordpress.org/core/handbook/tutorials/writing-developer-notes/)
- [`deprecated()`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-deprecated/#default) warning with no end version

#### Hard deprecation

- Mark with TypeScript `@deprecate`
- Update docs (readme, JSDoc, Storybook)
- Dev note
- `deprecated()` warning with end version
- Delete from package when end version arrives

```mermaid
flowchart TD
  1{{"`Do we encourage _new_ usages of this component?`"}}
[...]

PR #20008, opened 2020-02-03
Convert core toolbar buttons into ToolbarButton

diegohaz:
## Context

This PR is part of #18619, whose main goal is to implement [roving tabindex](https://www.w3.org/TR/wai-aria-practices/#kbd_roving_tabindex) on the `@wordpress/components`' `Toolbar` component and to use it on the block toolbars so they become a single tab stop as recommended by the [WAI-ARIA Toolbar Pattern](https://www.w3.org/TR/wai-aria-practices/#toolbar). Related issues are #15331 and #3383.

GIF showing the navigation through the Gutenberg UI using the tab key

This PR improves [`NavigableToolbar`](https://github.com/WordPress/gutenberg/blob/cbcb6195cea11c4dd3469ecfcb54721af873a3a4/packages/block-editor/src/components/navigable-toolbar/index.js) and converts the core toolbar buttons (those buttons that appear on all blocks: movers, block switcher and more options) into [`ToolbarButton`](https://github.com/WordPress/gutenberg/blob/cbcb6195cea11c4dd3469ecfcb54721af873a3a4/packages/components/src/toolbar-button/index.js) or [`ToolbarItem`](https://github.com/WordPress/gutenberg/blob/cbcb6195cea11c4dd3469ecfcb54721af873a3a4/packages/components/src/toolbar-item/index.js).

*`ToolbarItem` is the generic headless component that is used by `ToolbarButton` underneath and can be used to make other components, such as [`DropdownMenu`](https://github.com/WordPress/gutenberg/blob/fd5816e2fed2f241c4eeaea99d94b53af14293db/packages/components/src/dropdown-menu/index.js), aware of the context state provided by the parent [`Toolbar`](https://github.com/WordPress/gutenberg/blob/cbcb6195cea11c4dd3469ecfcb54721af873a3a4/packages/components/src/toolbar/toolbar-container.js) component, which enables the roving tabindex method.*

This also converts a few other buttons so the block toolbar on rich text blocks and other basic blocks already works with roving tabindex.

## `Toolbar` vs. `ToolbarGroup`

> Or `.components-toolbar` vs. `.components-toolbar-group`

Most of the changes on this PR are related to this. On #18534, `Toolbar` was renamed to [`ToolbarGroup`](https://github.com/WordPress/gutenberg/blob/e5dac46fcc0fb44d59d47182a28348a056ab431b/packages/components/src/toolbar-group/index.js), but the old class name (`components-toolbar`) was kept as there's a lot of code relying on it.

[...]

PR #63101, opened 2024-07-03
ToolbarButton: Deprecate `isDisabled` prop and merge with `disabled`

### LINKED ISSUE #63070: ToolbarButton: Accepts both `disabled` and `isDisabled` props with different behavior
ToolbarButton is intended to be disabled with the designated `isDisabled` prop, judging from the code and TypeScript types. 

However, it also accepts the standard `disabled` prop because it allows standard `<button>` props via `restProps`. To complicate things further, the `disabled` prop in these `restProps` are first passed onto `Ariakit.ToolbarItem` before eventually being passed down to `Button`, which opts out the toolbar item from the roving tabindex set regardless of whether the underlying `Button` allows focusable disabled.

So to allow focusable disabled, you'll need to pass a combination of either `disabled` & `accessibleWhenDisabled` (Ariakit API which is undocumented in our docs), or `isDisabled` & `__experimentalIsFocusable`.

I'm not immediately sure how we'd want to address this, but things should be less confusing. We might want to tweak the logic so that passing `disabled` will result in the same behavior as passing `isDisabled`.

---

Related: #57701

mirka:
Follow-up to https://github.com/WordPress/gutenberg/pull/62306#discussion_r1663277829
Fixes #63070

## What?

Deprecates the `isDisabled` prop on `ToolbarButton` and merges it into `disabled`.

## Why?

ToolbarButton accepts both `isDisabled` and `disabled` props to disable the button, and can result in slightly different behavior between the two, especially when wanting to achieve a focusable disabled button (see details in #63070).

Current API usage in our codebase shows that most people reach for the `disabled` prop. I could also find no meaningful significance for the `isDisabled` prop to be there ([original PR where it was introduced](https://github.com/WordPress/gutenberg/pull/20008#discussion_r421248963)).

## How?

[Passive deprecate](https://github.com/WordPress/gutenberg/issues/61099) `isDisabled` and alias it to `disabled`.

## Testing Instructions

Play around with the [Storybook code](https://github.com/WordPress/gutenberg/blob/566b82a13633cb72006d1900931ca46a8e0d7937/packages/components/src/toolbar/stories/index.story.tsx#L118-L121) for the `ToolbarButton`s in the `Toolbar` story. Passing `isDisabled` should result in the same behavior as passing `disabled`, especially in combination with accessible disabled props like `__experimentalIsFocusable` or `accessibleWhenDisabled`.


packages/components/src/toolbar/toolbar-button/index.tsx
```diff
@@ -19,51 +19,60 @@ import ToolbarButtonContainer from './toolbar-button-container';
...
 						event: ReactMouseEvent<
 							HTMLButtonElement & HTMLAnchorElement,
 							MouseEvent
 						>
 					) => {
 						event.stopPropagation();
-						// TODO: Possible bug; maybe use onClick instead of props.onClick.
-						if ( props.onClick ) {
-							props.onClick( event );
+						// TODO: Possible bug; maybe use onClick instead of restProps.onClick.
```
ciampo, line 63: Unrelated, but what is this comment about? 🤔

ciampo (approved):
LGTM.

One thing that I noticed while testing is that, if I pass `disabled` and `__experimentalIsFocusable` to `ToolbarButton`, the disabled button is still skipped when using arrow keys, but is not skipped when using the Tab key.

packages/components/src/toolbar/toolbar-button/index.tsx
mirka: The `onClick` function written here will be overridden by the consumer's `onClick` function passed through with the `restProps` 🙃

mirka:
> One thing that I noticed while testing is that, if I pass `disabled` and `__experimentalIsFocusable` to `ToolbarButton`, the disabled button is still skipped when using arrow keys, but is not skipped when using the Tab key.

Good catch 😳 I confirmed this also happens in trunk, but I think it will be normalized by #63102.

PR #63102, opened 2024-07-03
ToolbarButton: Always keep focusable when disabled

### LINKED ISSUE #57701: ToolbarButton: aria-disabled prop is also given the disabled attribute at the same time
### Description

For `Button` component, the `aria-disabled` prop is used instead of the `disabled` prop to make the button inactive but still focusable.

For example, the Undo/Redo button in the Post Editor or Site Editor.

image

https://github.com/WordPress/gutenberg/blob/5380b6c549668b94e38ae5619aa725f57ef7eb0f/packages/editor/src/components/editor-history/undo.js#L23-L36

On the other hand, `ToolbarButton` is used for the Undo/Redo button in the Widget Editor.

https://github.com/WordPress/gutenberg/blob/5380b6c549668b94e38ae5619aa725f57ef7eb0f/packages/edit-widgets/src/components/header/undo-redo/undo.js#L18-L27

It's mentioned that the `ToolbarButton` component accepts the same API as the `Button` component, but it seems like the `aria-disabled` prop also opts into the `disabled` prop at the same time.

toolbar-button

As a result, as mentioned in #57647 and #57678, it is not possible to focus the Undo/Redo button in the Widget Editor when it is not enabled.

This can be solved by refactoring the Undo/Redo button into the `Button` component, as attempted in #57677, but if the current behavior of `ToobarButton` is not what we intended, I think it will need to be fixed.

- Activate classic theme (Twenty Twenty-One).
- Open the Wdiget Editor.
- Use the tab key or arrow keys to move the focus within the header.
- You should not be able to focus on the Undo/Redo button.

Note: In the screencast below, you can also see that the focus outline is cut off, but this issue should also be fixed separately.

https://github.com/WordPress/gutenberg/assets/54422211/bb192ba9-63ed-4d08-a9e4-5ce235fb9f77

### Environment info

_No response_

Yes

Yes

mirka:
Fixes #57701

## What?

Keeps disabled `ToolbarButton`s focusable in all instances.

## Why?

Toolbar/menu items are one of the UI elements that are [generally recommended](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) to keep focusable when disabled.

## Testing Instructions

Play around with the [Storybook code](https://github.com/WordPress/gutenberg/blob/566b82a13633cb72006d1900931ca46a8e0d7937/packages/components/src/toolbar/stories/index.story.tsx#L118-L121) for the `ToolbarButton`s in the `Toolbar` story. Passing `disabled` should keep the button focusable.

### DIFF
--- a/packages/components/src/toolbar/toolbar-button/index.tsx
+++ b/packages/components/src/toolbar/toolbar-button/index.tsx
@@ -115,7 +115,6 @@ function UnforwardedToolbarButton(
  * ```jsx
  * import { Toolbar, ToolbarButton } from '@wordpress/components';
  * import { edit } from '@wordpress/icons';
-  import { __ } from '../../../../i18n/build-types/create-i18n';
  *
  * function MyToolbar() {
  *   return (
--- a/packages/components/src/toolbar/toolbar-button/index.tsx
+++ b/packages/components/src/toolbar/toolbar-button/index.tsx
@@ -32,7 +32,7 @@ function useDeprecatedProps( {
 function UnforwardedToolbarButton(
 	props: Omit<
 		WordPressComponentProps< ToolbarButtonProps, typeof Button, false >,
-		'__experimentalIsFocusable' // ToolbarButton will always be focusable.
+		'__experimentalIsFocusable' // ToolbarButton will always be focusable even when disabled.
 	> &
 		ToolbarButtonDeprecatedProps,
 	ref: ForwardedRef< any >
--- a/packages/components/src/toolbar/toolbar-button/index.tsx
+++ b/packages/components/src/toolbar/toolbar-button/index.tsx
@@ -16,7 +16,7 @@ import Button from '../../button';
 import ToolbarItem from '../toolbar-item';
 import ToolbarContext from '../toolbar-context';
 import ToolbarButtonContainer from './toolbar-button-container';
-import type { ToolbarButtonProps } from './types';
+import type { ToolbarButtonDeprecatedProps, ToolbarButtonProps } from './types';
 import type { WordPressComponentProps } from '../../context';
 
 function useDeprecatedProps( {
@@ -30,7 +30,11 @@ function useDeprecatedProps( {
 }
 
 function UnforwardedToolbarButton(
-	props: WordPressComponentProps< ToolbarButtonProps, typeof Button, false >,
+	props: Omit<
+		WordPressComponentProps< ToolbarButtonProps, typeof Button, false >,
+		'__experimentalIsFocusable' // ToolbarButton will always be focusable.
+	> &
+		ToolbarButtonDeprecatedProps,
 	ref: ForwardedRef< any >
 ) {
 	const {
@@ -111,6 +115,7 @@ function UnforwardedToolbarButton(
  * ```jsx
  * import { Toolbar, ToolbarButton } from '@wordpress/components';
  * import { edit } from '@wordpress/icons';
+  import { __ } from '../../../../i18n/build-types/create-i18n';
  *
  * function MyToolbar() {
  *   return (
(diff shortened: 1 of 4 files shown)


packages/components/src/toolbar/toolbar-item/index.tsx
```diff
@@ -50,6 +50,7 @@ function ToolbarItem(
 
 	return (
 		<Ariakit.ToolbarItem
+			accessibleWhenDisabled
```
mirka, line 53: I don't foresee a need for it, but technically a consumer can override this by passing `accessibleWhenDisabled={ false }`. This is currently undocumented on our end, but we could document it if it ever becomes a need.

packages/components/src/toolbar/toolbar-button/types.ts
```diff
@@ -37,6 +37,16 @@ export type ToolbarButtonProps = {
 	title?: string;
 };
 
+export type ToolbarButtonDeprecatedProps = {
+	/**
+	 * Whether to keep the button focusable when disabled.
+	 *
+	 * @deprecated ToolbarButton will always be focusable.
```
ciampo:
```suggestion
	 * @deprecated ToolbarButton will always be focusable even when disabled.
```

packages/components/src/toolbar/toolbar-button/index.tsx
```diff
@@ -30,7 +30,11 @@ function useDeprecatedProps( {
 }
 
 function UnforwardedToolbarButton(
-	props: WordPressComponentProps< ToolbarButtonProps, typeof Button, false >,
+	props: Omit<
+		WordPressComponentProps< ToolbarButtonProps, typeof Button, false >,
+		'__experimentalIsFocusable' // ToolbarButton will always be focusable.
```
ciampo:
```suggestion
		'__experimentalIsFocusable' // ToolbarButton will always be focusable even when disabled.
```

ciampo (approved):
🚀

tyxla (approved):
LGTM 👍 

Noting that in https://github.com/WordPress/gutenberg/commit/406ef77ae09b16060fbba9e86840c6d53501393c which I added to #63107 I'm suggesting to use the new stable prop, but this won't be necessary if this PR lands first - instead we'll have to remove the experimental prop usage.

mirka:
> Noting that in [406ef77](https://github.com/WordPress/gutenberg/commit/406ef77ae09b16060fbba9e86840c6d53501393c) which I added to #63107 I'm suggesting to use the new stable prop, but this won't be necessary if this PR lands first - instead we'll have to remove the experimental prop usage.

Good catch, thanks! I guess I'll do a cleanup & normalization pass in a separate PR. I've seen some `aria-disabled` and `accessibleWhenDisabled` usage here and there that should be cleaned up too.

talldan:
I noticed that since this PR was merged there seem to be some styling changes in the block toolbar.

Before:
Screenshot 2024-07-09 at 5 43 49 pm

After:
Screenshot 2024-07-09 at 4 51 32 pm


It's possible (very likely) the block toolbar styles are implemented incorrectly, so it might not be the fault of this PR, but thought I'd mention it.

mirka:
@talldan Is there a special repro condition for that? This is what I'm seeing at current trunk 8b470e0d74f880bba7c289239e8f92b0a4ec06f5

## Default
Block toolbar

## All locked
Block toolbar for all locked block

talldan:
Oh, sorry, for some reason I didn't include any 🤦 

If you edit any page in the site editor, and try selecting a template part block, the toolbar shows what it does in my comment.

mirka:
Ah ok, got it. That color is easy to fix, but when we take consistency and accessibility into consideration, I'm not sure if/how we should fix it 🤔 For example, it is in fact visibly grayed out in the "All locked" case I posted.

I do think the button needs to remain focusable, but if so we need to visually show that it's disabled. Maybe something like this?

https://github.com/WordPress/gutenberg/assets/555336/fb8c9e81-0031-4801-8ad8-4b20f0a7ee11

But all in all, I'm leaning toward predictability/consistency, which is to visually gray out as usual when a button is disabled. It's potentially confusing to even a sighted mouse user if an enabled-looking button is unresponsive. What do you think?

PR #63105, opened 2024-07-03
Implement a content panel on the Template Part block inspector.

jorgefilipecosta:
Follow up to a comment by @jasmussen at https://github.com/WordPress/gutenberg/pull/62034#pullrequestreview-2125177005. It implements a Content Panel on the template part block inspector similar to the one on the document tab.

cc: @jasmussen, @jameskoster 

## Screenshot
Screenshot 2024-07-03 at 18 47 26

## Testing Instructions
- Open a template like "All archives" on the site editor.
- Using the list view, select the header block.
- Go to the block inspector and verify a content panel is rendered there.

### DIFF
--- a/packages/block-library/src/template-part/edit/content-panel.js
+++ b/packages/block-library/src/template-part/edit/content-panel.js
@@ -1,0 +1,55 @@
+/**
+ * WordPress dependencies
+ */
+import { useSelect } from '@wordpress/data';
+import { useMemo } from '@wordpress/element';
+import { store as blocksStore } from '@wordpress/blocks';
+import {
+	store as blockEditorStore,
+	privateApis as blockEditorPrivateApis,
+} from '@wordpress/block-editor';
+import { PanelBody } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { BlockQuickNavigation } = unlock( blockEditorPrivateApis );
+
+export default function TemplatePartContentPanel( { clientId } ) {
+	const { descendentBlockIds, descendentBlockNames, blockTypes } = useSelect(
+		( select ) => {
+			const { getClientIdsOfDescendants, getBlockNamesByClientId } =
+				select( blockEditorStore );
+			const { getBlockTypes } = select( blocksStore );
+			const _descendentBlockIds = getClientIdsOfDescendants( clientId );
+			return {
+				descendentBlockIds: _descendentBlockIds,
+				descendentBlockNames:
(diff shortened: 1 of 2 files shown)


jasmussen:
Thanks for following up on this one. I could see this moving forward, though it is increasingly surfacing a question we need to answer at some point. Which is, what happens when you select an item in the content panel? Does it drill-down? Does it switch tabs? Both? The main reason this is worth thinking about is that in this view, the active tab is the _block_ tab. But in the focus mode view, the active tab is the _template part_ tab. So between both drilldowns and tab-switching, if we do those, it starts conflicting. Ultimately the goal of either would be to provide light-weight styling options for each piece, as accessed from this panel. Notably in contenOnly locked patterns, as well as in synced patterns with overrides. The answer may be that, despite by instincts otherwise, maybe this wasn't the right instinct after all; and maybe instead of showing the contents, we can show an "Edit template part" button? Sorry if this is a detour, Jorge, and let's hear from Jay as well.

jameskoster:
For now I'm not sure we should expose the Content panel when a template part is selected in the context of a parent document (template). The granularity of edit capabilities we want to offer in this context isn't super clear. In fact, over time we've moved towards _less_ granular edit capabilities, funnelling folks into the focus mode for more detailed editing.

So on that note, I think it would make sense to expose this panel when you're editing the template part in isolation. Perhaps we start there?

jasmussen:
It should already be exposed in isolation, if I'm not wrong. It was my suggestion to also show it here, but as noted, now that it's here, I'm also not sure. Apologies for the misdirect, Jorge.

jorgefilipecosta:
Rebased onto current `trunk`.

The branch was very stale (32k+ commits ahead due to an old merge of `master` into `add-packages`). The PR effectively contained a single meaningful commit (`Implement a content panel on the Template Part block inspector.`), so the rebase was done by cherry-picking that commit onto the latest `trunk`.

Conflict resolved in `packages/block-library/src/template-part/edit/index.js`: trunk had moved `TemplatesList` into `<InspectorControls group="settings">`. The PR's new `<ContentPanel />` was placed inside that same settings-group `InspectorControls`, alongside `TemplatesList`.

No other conflicts. Force-pushed with `--force-with-lease`.

jasmussen:
What are you looking for as far as feedback, to land this?

jameskoster:
Noting that I'm already seeing the Content panel when editing a template part:

Screenshot 2026-06-03 at 09 57 19

My feedback from [before](https://github.com/WordPress/gutenberg/pull/63105#issuecomment-2208667102) still stands. I don't know that we should show content when selecting a template part block since it's not directly editable.

PR #60613, opened 2024-04-09
DateTime: Create TimeInput component and integrate into TimePicker

bogiii:
## What?
* Introduces a simple `TimeInput` component
* Integrates the `TimeInput` into `TimePicker` component

## Why?

Before these changes, there was no option to use the simple time input component.

| Before | After |
|--------|--------|
| Screenshot 2024-04-11 at 12 11 18 | Screenshot 2024-04-11 at 12 11 29 | 

## How?

Here are more details: https://github.com/WordPress/gutenberg/issues/52734

## Testing Instructions

* Go to TimeInput component: `http://localhost:50240/?path=/docs/components-timeinput--docs`
* Play with applying different combination of arguments

--

* Go to TimePicker component: `http://localhost:50240/?path=/docs/components-timepicker--docs`
* Check if everything work on same way as before this changes

Unit tests: Good old unit tests found a bug, which is resolved. ✅ 

## Screenshots or screencast 

Screen Capture on 2024-04-11 at 12-42-03

---

## TimePicker: Time input can be used in isolation

The time input of the `TimePicker` component can now be used in isolation as `<TimePicker.TimeInput />`. In this case, the `value` will be passed as an object in 24-hour format (`{ hours: number, minutes: number }`).

Issue #52734, opened 2023-07-18
TimePicker components should have option to disable date field


Tropicalista:
At the moment there is no component to display a time picker that only show time.

The timepicker component has no optio to disable the date field. So it's mostly a light version of datetimepicker.

t-hamano:
This suggestion makes sense to me. My feeling is that `TimePicker` should not inherently have the date control.

The current implementation of `DateTimePicker`, `DatePicker`, `TimePicker` and the specifications I expect are as follows:

### Actual

| Component | Has date input? | Has calendar? | Has time input? |
|--------|--------|--------|--------|
| `DateTimePicker` | ✅ | ✅ | ✅ |
| `DatePicker` |  | ✅ |  |
| `TimePicker` | ✅ |  | ✅ | 

### Expected

| Component | Has date input? | Has calendar? | Has time input? |
|--------|--------|--------|--------|
| `DateTimePicker` | ✅ | ✅ | ✅ |
| `DatePicker` | ✅ | ✅ |  |
| `TimePicker` |  |  | ✅ |

mirka:
I think the complexity here is that `TimePicker` is built for manipulating full `Date` values like `2023-08-17T08:41:00`. So it isn't just a matter of visually hiding the inputs for the date part, unless you're ok with passing half-fake `Date` values to `currentTime` and getting a half-fake value back from the `onChange`.

bogiii:
I absolutely agree with the proposed expected table specification from [the comment](https://github.com/WordPress/gutenberg/issues/52734#issuecomment-1642150244) above.

> I think the complexity here is that TimePicker is built for manipulating full Date values like 2023-08-17T08:41:00
> Unless you're ok with passing half-fake Date values to currentTime and getting a half-fake value

Changing a format is not an option since it sounds like a breaking change.

So, if we want to support hiding the date part, the only option is to keep the same component configuration with the full format and provide an additional boolean property.

Here is the draft PR: https://github.com/WordPress/gutenberg/pull/60034

@mirka, what do you think?

bogiii:
I've investigated the other component libraries and watched how they return the time picker value. Most of them return it in the full-date format, with the current date if it's not set, which is also our case.

From my perspective, **it isn't a half-fake date string** and a blocker for further component improvements.

If this still doesn't sound like a possible solution, I would recommend creating a new separate Time picker without a date part. I can open a feature request, and then we can shape the details.
For example:
[...]

Issue #51740, opened 2023-06-21
Make sure interactive controls are always labeled


afercia:
### Description

Any interactive control (inputs, textarea, select, checkboxes, radio buttons, buttons, etc.) should always be labeled. This is a basic accessibility requirement as interactive controls need to inform users what they're about.

Turns out that many (all?) the base components for interactive controls do not enforce any check for some proper labeling nor hey make their `lbel` prop a required prop. This opens the door to developers misuse, as it is possible to:
- either pass an empty label prop
- or omit the label prop entirely

In both cases, the control will be unlabeled. Worth reminding that this is not only about the _visual_ text associated with a control (which usually is a `<label>` element). It's is about the actual accessible name of a control: in case of missing or empty associated `<label>` element, missing aria-label / aria-labelledby or any other labeling mechanism, the control accessible name is empty. 

Being open to misuse, my concern is that these components may potentially contribute to lower the accessibility level of the entire internet, given WordPress is used on millions and millions of web sites. 

For no reason, ever, an interactive control should be unlabeled. All these components should enforce proper labeling.

As en example of developers misuse, here's a screenshot from a widely used WordPress plugin for e-commerce:

<img width="1257" alt="gutenberg allows for empty and incorrect labeling woo wizard" src="https://github.com/WordPress/gutenberg/assets/1682452/70f15844-2f56-440d-ad92-bd9dbbbbbd47">

- A CheckboxControl control is passed an empty `label` prop, or the prop is omitted entirely.
- The CheckboxControl still renders a `<label>` element, but it's empty.
- The checkbox is unlabeled.
- Developers added some visible text close to the checkbox, but this text is only a paragraph. Visually, it looks like a label but it's not. 

This is very far from ideal. Gutenberg should not allow for unlabeled controls. We can't rely on developers awareness and knowledge. Proper labeling must be enforced by the components themselves. Right now, passing an empty `label` prop or omitting it entirely doesn't even trigger a warning or error.

Cc @ciampo @mirka 

### Step-by-step reproduction instructions

- Add some interactive controls somewhere, omitting the `label` prop:

```
<CheckboxControl />
<TextControl autoComplete="off" value=""} />
```

- Observe a checkbox and an input field are rendered with no errors or warning.
- Observe the two controls are unlabeled.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

afercia:
[...]

PR #63106, opened 2024-07-03
TimeInput: Add `label` prop

mirka:
Part of #52734
Follow-up to #60613

## What?

Adds a `label` prop to `TimeInput`.

## Why?

Since some people will be using this as an alternative to `TimePicker`, we need a way to add a visual label as well. (Technically, it needs to be a fieldset legend rather than a `label`.)

## Testing Instructions

See Storybook for TimeInput to test behavior.

(Code changes are easier to review with whitespace hidden)

### DIFF
--- a/packages/components/src/date-time/types.ts
+++ b/packages/components/src/date-time/types.ts
@@ -62,6 +62,11 @@ export type TimeInputProps = {
 	minutesProps?: React.ComponentProps< typeof MinutesInput >;
 
 	/**
+	 * The label for the time input.
+	 */
+	label?: string;
+
+	/**
 	 * The function is called when a new time has been selected.
 	 * Passing hours and minutes as an object properties.
 	 */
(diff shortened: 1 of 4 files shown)


tyxla (approved):
Tested with and without a label - works as expected, thanks! 🚀

packages/components/src/date-time/types.ts
```diff
@@ -61,6 +61,11 @@ export type TimeInputProps = {
 	 */
 	minutesProps?: React.ComponentProps< typeof MinutesInput >;
 
+	/**
+	 * The label for the time input.
+	 */
+	label?: string;
```
ciampo, line 67: Should we consider making the `label` required by default, so that when the input is used standalone consumers are nudged towards labelling it? We could disable the rule when compositing is in `DateTimePicker`.

mirka: I don't think it's a huge issue for this one specifically, because the hours and minutes inputs are both labeled.

In the general case though (re: #51740), I'm starting to think that an eslint rule for our repo might be a good way to nudge devs without compromising the "truthfulness" of our TS types.

PR #62282, opened 2024-06-04
Button: Stabilize `__experimentalIsFocusable` prop

mirka:
## What?

Stabilizes the `__experimentalIsFocusable` prop on the Button component into a `accessibleWhenDisabled` prop.

### TODO

- [x] Fix eslintrc cascade around `no-restricted-syntax` rule (#62306)
- [x] Migrate existing usage in repo (#63107)
- [x] Adjust eslint rule for new prop (#63107)

## Why?

We are tightening our standards (#62080) around disabled Button accessibility.

The `accessibleWhenDisabled` naming and pattern aligns with what Ariakit is doing, which gives some confidence that this is sustainable and scalable across a component library.

## How?

- Improve documentation to encourage usage and developer education.
- [Passive deprecate](https://github.com/WordPress/gutenberg/issues/61099) the `__experimentalIsFocusable` prop.

## Testing Instructions

- Readme and Storybook props table descriptions should look good.
- Static checks on CI (i.e. the linter) should pass after #63107 is merged into this branch.

---

## ✍️ Dev Note

The `Button` component had a `__experimentalIsFocusable` prop to keep the button focusable when disabled. This prop is now stabilized and renamed to `accessibleWhenDisabled`. (Existing usages of `__experimentalIsFocusable` will continue to work indefinitely.)

In most cases, it is recommended to set this to `true`. Disabling a control without maintaining focusability can cause accessibility issues, by hiding their presence from screen reader users, or by preventing focus from returning to a trigger element.

Learn more about the [focusability of disabled controls](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) in the WAI-ARIA Authoring Practices Guide.

PR #63107, opened 2024-07-03
Replace `__experimentalIsFocusable` with `accessibleWhenDisabled`

mirka:
Stacked on #62282

## What?

Replaces `__experimentalIsFocusable` with `accessibleWhenDisabled` on all existing `Button`s.

## Why?

For #62282 (extracted to this PR for easier review)

## How?

Codemod + manual inspection of the code changes.

## Testing Instructions

✅ Linter passes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -143,9 +143,9 @@ const restrictedSyntax = [
 const restrictedSyntaxComponents = [
 	{
 		selector:
-			'JSXOpeningElement[name.name="Button"]:not(:has(JSXAttribute[name.name="__experimentalIsFocusable"])) JSXAttribute[name.name="disabled"]',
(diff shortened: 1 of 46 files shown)


tyxla (approved):
Looks good, thanks 👍 

In 406ef77ae09b16060fbba9e86840c6d53501393c we handle a few `ToolbarButton` instances, because why not? These were the last usages of `__experimentalIsFocusable` in production code, so it only made sense to cover them as well IMHO. Feel free to drop the commit if you disagree or if you were planning to address that separately. 

On a side note, with the ESLint rule, should we also handle `ToolbarButton` and any other button components we know about that also pass `restProps` down to `Button`? This can totally be done in a separate PR of course.

tyxla:
Just saw #63102, but it seems like instead of changing the prop to the new one we'll now need to remove it? Feel free to remove my commit and just remove the props in another one if #63102 ends up landing earlier.

PR #61383, opened 2024-05-06
Use `grid-auto-rows` to ensure min height of new grid rows.

tellthemachines:
## What?

Fixes an issue that is becoming more visible with the work on #60986: when a grid item spans several rows, but there aren't otherwise enough items in the grid to occupy that number of rows, the auto rows created to support the multiple-row item are collapsed to 0 height. This looks quite odd, and is unlikely to be what's expected when resizing an item to take up more vertical space:
Screenshot 2024-05-06 at 12 22 26 PM

This PR adds a default value to `grid-auto-rows`, so that auto-created rows always have the same minimum height as the other grid items:

Screenshot 2024-05-06 at 12 24 29 PM

## Testing Instructions

To test, add the following markup to a post:

<details>
<summary>test markup</summary>

```
<!-- wp:group {"layout":{"type":"grid"}} -->
<div class="wp-block-group"><!-- wp:paragraph {"style":{"color":{"background":"#91ded8"},"layout":{"columnSpan":"2"}}} -->
<p class="has-background" style="background-color:#91ded8">grid item</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"color":{"background":"#91ded8"},"layout":{"rowSpan":"3"}}} -->
<p class="has-background" style="background-color:#91ded8">grid item</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"color":{"background":"#91ded8"}}} -->
<p class="has-background" style="background-color:#91ded8">grid item</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"color":{"background":"#91ded8"}}} -->
<p class="has-background" style="background-color:#91ded8">grid item</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"color":{"background":"#91ded8"}}} -->
<p class="has-background" style="background-color:#91ded8">grid item</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"style":{"color":{"background":"#91ded8"}}} -->
<p class="has-background" style="background-color:#91ded8">grid item</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```
</details>

Then set the alignment of the grid to full, and resize the window so that all grid items fit on the same column.

PR #63108, opened 2024-07-04
Try inserting blocks directly in empty grid cells

### LINKED ISSUE #62518: Grid Block: clicking on empty grid item doesn't add a new block
### Description

When editing a Grid, I expected to be able to click on the outline of the empty grid item to add a block to it. It feels inconsistent with the similar UI found in the Columns Block

1. Add a Grid Block
2. Type something in the first block
3. Try clicking on the outline of the grid item next to it
4. Nothing will happen

https://github.com/WordPress/gutenberg/assets/14153300/99513017-8c3a-4cd9-91b8-19c9fc8ad6fc

### Environment info

- WordPress 6.6 Beta 2, Gutenberg 18.5

Yes

Yes

tellthemachines:
## What?

Fixes #62518.
Part of #57478.

Behind the experiment flag, in manual mode:

* Makes the grid visualizer always visible even when grid isn't selected, so blocks can be dragged into cells from outside the grid;
* Adds a block appender to empty grid cells so new blocks can be added in any position.

## Testing Instructions

1. Under Gutenberg > Experiments, enable "Grid interactivity";
2. Add a Grid block to a post or template and set it to Manual mode;
3. Adjust columns and rows and add a few blocks inside the grid, but leave some cells empty;
4. Try dragging a block from outside the grid into an empty cell;
5. Try clicking the block appender in an empty cell and add a new block.

### DIFF
--- a/packages/block-editor/src/components/button-block-appender/index.js
+++ b/packages/block-editor/src/components/button-block-appender/index.js
@@ -18,7 +18,7 @@ import deprecated from '@wordpress/deprecated';
 import Inserter from '../inserter';
 
 function ButtonBlockAppender(
-	{ rootClientId, className, onFocus, tabIndex, blockAttributes },
+	{ rootClientId, className, onFocus, tabIndex, onSelectCallback },
 	ref
 ) {
 	return (
@@ -26,7 +26,7 @@ function ButtonBlockAppender(
 			position="bottom center"
 			rootClientId={ rootClientId }
 			__experimentalIsQuick
-			blockAttributes={ blockAttributes }
+			onSelectCallback={ onSelectCallback }
 			renderToggle={ ( {
 				onToggle,
 				disabled,
--- a/packages/block-editor/src/components/button-block-appender/index.js
+++ b/packages/block-editor/src/components/button-block-appender/index.js
@@ -18,7 +18,7 @@ import deprecated from '@wordpress/deprecated';
 import Inserter from '../inserter';
 
 function ButtonBlockAppender(
-	{ rootClientId, className, onFocus, tabIndex, onSelectCallback },
+	{ rootClientId, className, onFocus, tabIndex, onSelect },
 	ref
 ) {
 	return (
@@ -26,7 +26,7 @@ function ButtonBlockAppender(
 			position="bottom center"
 			rootClientId={ rootClientId }
 			__experimentalIsQuick
-			onSelectCallback={ onSelectCallback }
+			onSelectOrClose={ onSelect }
 			renderToggle={ ( {
 				onToggle,
 				disabled,
--- a/packages/block-editor/src/components/button-block-appender/index.js
+++ b/packages/block-editor/src/components/button-block-appender/index.js
@@ -18,7 +18,7 @@ import deprecated from '@wordpress/deprecated';
 import Inserter from '../inserter';
 
 function ButtonBlockAppender(
-	{ rootClientId, className, onFocus, tabIndex },
+	{ rootClientId, className, onFocus, tabIndex, blockAttributes },
 	ref
 ) {
 	return (
@@ -26,6 +26,7 @@ function ButtonBlockAppender(
 			position="bottom center"
 			rootClientId={ rootClientId }
 			__experimentalIsQuick
+			blockAttributes={ blockAttributes }
 			renderToggle={ ( {
 				onToggle,
 				disabled,
--- a/packages/block-editor/src/components/grid/grid-visualizer.js
+++ b/packages/block-editor/src/components/grid/grid-visualizer.js
@@ -15,9 +15,10 @@ import { __experimentalUseDropZone as useDropZone } from '@wordpress/compose';
  */
 import { __unstableUseBlockElement as useBlockElement } from '../block-list/use-block-props/use-block-refs';
 import BlockPopoverCover from '../block-popover/cover';
-import { range, GridRect, getGridInfo } from './utils';
+import { range, GridRect, getGridInfo, getComputedCSS } from './utils';
 import { store as blockEditorStore } from '../../store';
 import { useGetNumberOfBlocksBeforeCell } from './use-get-number-of-blocks-before-cell';
+import ButtonBlockAppender from '../button-block-appender';
 
 export function GridVisualizer( { clientId, contentRef, parentLayout } ) {
 	const isDistractionFree = useSelect(
@@ -43,6 +44,28 @@ export function GridVisualizer( { clientId, contentRef, parentLayout } ) {
 		/>
 	);
 }
+
+const checkIfCellOccupied = ( gridElement, column, row ) => {
+	const cell = Array.from( gridElement.children ).find( ( child ) => {
+		const [ columnStart, columnSpan ] =
+			getComputedCSS( child, 'grid-column' ).match( /\d+/g ) || [];
+		const [ rowStart, rowSpan ] =
+			getComputedCSS( child, 'grid-row' ).match( /\d+/g ) || [];
+		const columnEnd = columnSpan
+			? parseInt( columnStart, 10 ) + parseInt( columnSpan, 10 ) - 1
+			: columnStart;
+		const rowEnd = rowSpan
+			? parseInt( rowStart, 10 ) + parseInt( rowSpan, 10 ) - 1
+			: rowStart;
+		return (
+			column >= columnStart &&
+			column <= columnEnd &&
+			row >= rowStart &&
+			row <= rowEnd
+		);
+	} );
+	return cell !== undefined;
+};
 
 const GridVisualizerGrid = forwardRef(
 	( { clientId, gridElement, isManualGrid }, ref ) => {
@@ -99,25 +122,44 @@ const GridVisualizerGrid = forwardRef(
 					{ isManualGrid
 						? range( 1, gridInfo.numRows ).map( ( row ) =>
 								range( 1, gridInfo.numColumns ).map(
-									( column ) => (
-										<GridVisualizerCell
-											key={ `${ row }-${ column }` }
-											color={ gridInfo.currentColor }
-										>
-											<GridVisualizerDropZone
-												column={ column }
-												row={ row }
-												gridClientId={ clientId }
-												gridInfo={ gridInfo }
-												highlightedRect={
-													highlightedRect
-												}
-												setHighlightedRect={
-													setHighlightedRect
-												}
-											/>
-										</GridVisualizerCell>
-									)
+									( column ) => {
+										const isCellOccupied =
+											checkIfCellOccupied(
+												gridElement,
+												column,
+												row
+											);
+										return (
+											<GridVisualizerCell
+												key={ `${ row }-${ column }` }
+												color={ gridInfo.currentColor }
+											>
+												<GridVisualizerDropZone
+													column={ column }
+													row={ row }
+													gridClientId={ clientId }
+													gridInfo={ gridInfo }
+													highlightedRect={
+														highlightedRect
+													}
+													setHighlightedRect={
+														setHighlightedRect
+													}
+												/>
+												{ ! isCellOccupied && (
+													<ButtonBlockAppender
+														rootClientId={
+															clientId
+														}
+														className="grid-visualizer-appender"
+														style={ {
+															color: gridInfo.currentColor,
+														} }
+													/>
+												) }
+											</GridVisualizerCell>
+										);
+									}
 								)
 						  )
 						: Array.from(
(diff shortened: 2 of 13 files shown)


noisysocks:
Heck yeah, great start!

1) Things look pretty gnarly when the grid is empty:

   screenshot_2024-07-04_at_16 43 27

   Maybe try darkening the empty cell on hover to indicate that it's clickable and only showing the + icon in the middle of the cell on hover when the cell is big enough.

2) I'm not sure about _always_ showing the grid visualiser. I think we should copy the Columns block where if the Grid is not selected then it looks how it will on the frontend. All the UI affordances only appear when the block is selected. How about we show the visualiser when the grid block is selected, when a child block is selected, or when the user is dragging a block (`select( blockEditorStore ).isDraggingBlocks()`)?

3) I'm no longer able to drag and drop blocks to position in manual mode.

tellthemachines:
Thanks for testing @noisysocks ! I've updated the PR so the grid visualizer only appears either when the block (or child) is selected or a block is being dragged. I've also adjusted block appender styles, but haven't got this quite right yet:

> Maybe try darkening the empty cell on hover to indicate that it's clickable and only showing the + icon in the middle of the cell on hover when the cell is big enough.

I can't seem to get rid of the plus button 😅

tellthemachines:
Marking this ready for review as I think it's in a pretty good place now!

packages/block-editor/src/hooks/grid-visualizer.js
```diff
@@ -0,0 +1,67 @@
...
+			<GridLayoutSync clientId={ clientId } />
+		</>
+	);
+}
+
+const addGridVisualizerToBlockEdit = createHigherOrderComponent(
+	( BlockEdit ) => ( props ) => {
+		if (
+			! props.attributes?.layout ||
+			props.attributes.layout?.type !== 'grid'
```
noisysocks: Can be simplified to `props.attributes.layout?.type !== 'grid'`.

packages/block-editor/src/components/inserter/menu.js
```diff
@@ -40,6 +40,7 @@ function InserterMenu(
 		clientId,
 		isAppender,
 		__experimentalInsertionIndex,
+		__experimentalOnSelect,
 		onSelect,
```
noisysocks: This component has two `onSelect` props. It should only need one.

packages/block-editor/src/components/inserter/index.js
```diff
@@ -150,6 +150,7 @@ class PrivateInserter extends Component {
 			prioritizePatterns,
 			onSelectOrClose,
 			selectBlockOnInsert,
+			onSelectCallback,
```
noisysocks: This component has `onSelectOrClose` and `onSelectCallback`. It should only need one.

packages/block-editor/src/components/grid/grid-visualizer.js
```diff
@@ -221,21 +318,58 @@ function GridVisualizerDropZone( {
 			__unstableMarkNextChangeAsNotPersistent();
 			moveBlocksToPosition(
 				[ srcClientId ],
-				gridClientId,
+				getBlockRootClientId( srcClientId ),
```
noisysocks, line 275: What's this change for?

packages/block-editor/src/components/grid/grid-visualizer.js
```diff
@@ -99,25 +133,84 @@ const GridVisualizerGrid = forwardRef(
...
+																	block.clientId,
+																],
+																clientId,
+																clientId,
+																getNumberOfBlocksBeforeCell(
+																	column,
+																	row
+																)
+															);
+														} }
```
noisysocks: I'd move this callback logic into `GridVisualizerAppender` so that it's more consistent with `GridVisualizerDropZone` (which does all of that logic in the component) and so that this code isn't so heavily indented.

packages/block-editor/src/components/grid/grid-visualizer.js
```diff
@@ -44,6 +45,28 @@ export function GridVisualizer( { clientId, contentRef, parentLayout } ) {
 	);
 }
 
+const checkIfCellOccupied = ( gridElement, column, row ) => {
+	const cell = Array.from( gridElement.children ).find( ( child ) => {
```
noisysocks: Bit worried about performance here. We're iterating through the grid's children once per cell. So for a 16x16 grid with blocks in every cell that's `16*16*16*16` calls to `getComputedCSS` on every render.

An alternative approach might be to calculate `rects` similar to `use-grid-layout-sync.js:50` and then you can tell if a cell is occupied by looking for a rect where `rect.contains( row, column )`.

I might try that if you don't mind me pushing to this branch.

packages/block-editor/src/components/button-block-appender/index.js
```diff
@@ -18,14 +18,15 @@ import deprecated from '@wordpress/deprecated';
...
-	{ rootClientId, className, onFocus, tabIndex },
+	{ rootClientId, className, onFocus, tabIndex, onSelectCallback },
 	ref
 ) {
 	return (
 		<Inserter
 			position="bottom center"
 			rootClientId={ rootClientId }
 			__experimentalIsQuick
+			onSelectCallback={ onSelectCallback }
```
noisysocks: Is the `Callback` suffix necessary? The `on` prefix already hints that it's a function so I'd simply call this `onSelect`.

noisysocks:
This is cool. Makes the flow surrounding Manual mode so much more powerful now that you can start with an empty manual grid and then build up from there. Here's how everything looks:

https://github.com/WordPress/gutenberg/assets/612155/930bfe2d-5703-4e31-8e8e-9ad55fbaba9a

We need to work on some of the design details (colours, etc.) with @jasmussen. I'm happy to merge this behind the experimental flag in a somewhat ugly state and then do a round of polish later.

The minimum cell height being 8px is annoying. Collapsed cells are difficult to click on. I'm not certain how we should fix that. Do we just make collapsed cells render with a 16px minimum size? The problem is then that it wouldn't be WYSIWYG.

I'm seeing one annoying bug where the appender appears on top of a newly inserted paragraph block:

https://github.com/WordPress/gutenberg/assets/612155/492362e2-8554-40a8-a5a0-d4d67ec9f891

packages/block-editor/src/components/grid/grid-visualizer.js
tellthemachines: Sure, go ahead!

tellthemachines: I think the bug you're seeing with appender still appearing when a paragraph is added must happening in this function, because the `gridElement` refreshes with the new paragraph and its correct col and row attributes, but `isCellOccupied` still returns false for that cell.

noisysocks: How's this look? 9a626e3c6f

Not very scientific but it feels a bit snappier to me clicking around.

tellthemachines: Hmmm, I'm seeing `The 'useSelect' hook returns different values when called with the same state and parameters. This can lead to unnecessary rerenders. ` in the console when selecting the grid.

tellthemachines: But the issue with the appender still rendering is fixed! And the code looks much neater ☺️

noisysocks: Oops. Fixed.

packages/block-editor/src/components/grid/grid-visualizer.js
tellthemachines: We need to allow blocks from outside the grid (so with a different rootClientId) to be dropped inside the grid too.

tellthemachines:
> The minimum cell height being 8px is annoying. Collapsed cells are difficult to click on. I'm not certain how we should fix that. Do we just make collapsed cells render with a 16px minimum size? The problem is then that it wouldn't be WYSIWYG

Currently the 8px minimum height is WYSIWYG because I added it to the front end too 😁 . I wouldn't object to making it _slightly_ bigger. It seems weird to deliberately add empty rows to a grid and then not expect anything to change visually. Might be good to get feedback on this one though, there was some pushback on adding row height in #61383.

I think I've addressed all the remaining feedback now!

jasmussen:
Thanks for including the summary + videos, Robert, makes it substantially easier to review.

> Makes the flow surrounding Manual mode so much more powerful now that you can start with an empty manual grid and then build up from there.
> The minimum cell height being 8px is annoying. Collapsed cells are difficult to click on. I'm not certain how we should fix that. Do we just make collapsed cells render with a 16px minimum size? The problem is then that it wouldn't be WYSIWYG.

My first instinct in seeing this is that the slivers of plus buttons are tool small to be meaningfully clickable, especially if you've limited dexterity. 

I realize we discussed adding a min-height of 1em (same as a default paragraph height) to those rows in the past, and it was suggested (if I'm not misremembering) that we should add as little CSS as possible. Yet those two challenges come to a head here, as without a taller min-height, clicking that plus is just not going to be a good experience.

Should we add editor-only CSS? How much would this break the WYSIWYG?

Some other observations:

Screenshot 2024-07-08 at 10 52 09

As-is, we can't add these mover controls to the block toolbar:

* The block toolbar becomes too wide. Even if it scrolls horizontally, it's going to be problematic, again, considering folks with limited dexterity. Consider smaller context, including tablets and mobile. 
* It's also a different visual style than the existing movers, which were designed to be very compact for the same reason.
* Beyond other ideas that can be explored, I'd consider adding them to the block toolbar ellipsis menu, and then adding shortcuts. 

Speaking of shortcuts, the existing shortcuts for moving up and down are these:
Screenshot 2024-07-08 at 11 00 23

We don't show those in the tooltips:

Screenshot 2024-07-08 at 10 59 46

We also don't show these in the ellipsis menu, there's just a "Move to" item:
Screenshot 2024-07-08 at 11 02 01

Finally, these options fall apart in text-labels only mode. There's currently no way to horizontally scroll the block toolbar, and even if there was, it would be challenging for people with motor control disabilities:
2024-07-08 11 15 48 local-wordpress local 6607c61c8eb0

It suggests there's an opportunity to consolidate and improve these flows. Supposedly the [DropdownMenuV2](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs) supports nested menu items, this would allow us to do something like this quick test:

New movers

In this, I've also explored more ergonomic keyboard shortcuts. At the moment, those particular shortcuts are used by my browser, but it seems fine in the name of simplicity to override those in the contet of the editor.

None of that I would expect done in this PR. But I would avoid adding the mover controls to the block toolbar until they can be added in a way that is considerate of the accessibility problems they currently introduce around legibility and motor control. 

Is there a way to leverage the existing "Move to" option in the ellipsis menu for this?

noisysocks:
Thanks for the feedback @jasmussen. 

Agree let's try giving the rows a 1em minimum height to match the Paragraph block. I think I'd prefer that this only exists in the editor as an affordance for making the grid visualiser usable. We should be as light as possible on what frontend CSS there is for grids.

Regarding the block movers, definitely didn't intend on shipping with the current ugly implementation 😅 In my mind I was hoping we could do a version of the movers where we combine the horizontal and vertical movers into one control. Something like this:

Mover

(Excuse the mockup, I'm bad at Figma.) Is that a bad idea?

I don't really like the idea of burying them behind a More → Move dropdown as it's quite a central interaction to using a manual grid and our drag and drop implementation might not be foolproof in some cases.

I DO really like the idea of having better keyboard shortcuts. I'm always trying to move blocks around using <kbd>Cmd</kbd>+<kbd>Option</kbd>+<kbd>↑</kbd> etc. because that's what works in Bear, Notion, Craft, etc.

tellthemachines:
> I don't really like the idea of burying them behind a More → Move dropdown as it's quite a central interaction to using a manual grid and our drag and drop implementation might not be foolproof in some cases.

Yeah I tend to agree, block movers in the manual grid are important to have handy. Could the toolbar wrap if it gets too long? That might be clunky, but it's still a better experience than not being able to access the the whole thing at all.

➕ to keyboard shortcuts!

noisysocks (approved):
16px/1rem still feels pretty small as a hit area to me, but I don't know what else to do. Let's ruminate on it.

And yes still a lot of UI details to work out, especially wrt to the movers. I believe it's already noted in https://github.com/WordPress/gutenberg/issues/57478.

Still, this is good progress on an experimental feature. Onwards! 🙂

jasmussen:
> Could the toolbar wrap if it gets too long? That might be clunky, but it's still a better experience than not being able to access the the whole thing at all.

I think of the block toolbar as a forcing function. It is designed to be only the most important tools at your disposal, in the canvas, and anything else in the inspector which is designed to be the more expansive place to interact with blocks. So I would suggest, no, the block toolbar should never wrap. 

But this forcing function also suggests we still have work to do, to improve the block toolbar to better serve this purpose, and to be better across mobile and text-labels-only modes. Specifically for text-labels only mode, it suggests we can use more active and concise language. "Select parent" instead of "Select parent block: Grid". "Move up", instead of "Move block up" (wouldn't that need to be "Move block upwards" regardless?), "Filter", instead of "Apply duotone filter", "Add text" instead of "Add text over image", etc.

PR #63109, opened 2024-07-04
Make SiteHub available in mobile viewports: Try moving sidebar navigation context up higher

andrewserong:
## What?

Split out from #63060, this PR attempts to move the `SidebarNavigationContext` provider up to the layout in the site editor.

Note: this PR targets `try/sitehub-mobile` instead of `trunk` to (hopefully) make it easier to see what's going on.

## Why?

So that we don't need to wrap the site hub in the `SidebarContent` component which was causing issues with which buttons are focused, and added unnecessary padding.

I've created this as a separate PR just to see if the idea works, and to make it easier to throw away than committing to #63060.

## How?

* Remove no longer needed `<SidebarContent>` wrapping for the site hub
* Move the `SidebarNavigationContext` provider up to the layout so that the context can be shared with the site hub, without requiring extra wrapping

## Testing Instructions

Same as for #63060

Note: when looking at this PR, you might want to look at it with whitespace removed so that the diff is easier to read: https://github.com/WordPress/gutenberg/pull/63109/files?diff=unified&w=1

### DIFF
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -41,7 +41,7 @@ import KeyboardShortcutsRegister from '../keyboard-shortcuts/register';
 import KeyboardShortcutsGlobal from '../keyboard-shortcuts/global';
 import { useIsSiteEditorLoading } from './hooks';
 import useMovingAnimation from './animation';
-import SidebarContent from '../sidebar';
+import SidebarContent, { SidebarNavigationContext } from '../sidebar';
 import SaveHub from '../save-hub';
 import SavePanel from '../save-panel';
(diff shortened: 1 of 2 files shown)


andrewserong:
Update: I was hoping this might help fix things, but it looks like the `SiteHub` flickers in and out a bit during the animation:

https://github.com/WordPress/gutenberg/assets/14988353/4c375282-bf8f-43a8-bd2a-bac56344b667

andrewserong:
Closing this out, as I think I've found a simpler fix for the focus issue over in https://github.com/WordPress/gutenberg/pull/63060/commits/4401f6037f19563d60fe50d3b1e21add43a106e3

PR #63112, opened 2024-07-04
Grid layout: Use numbers instead of strings for layout attributes

noisysocks:
There are several numeric grid layout attributes. Some are stable and were shipped in WP 6.6: `columnCount`, `columnSpan`, and `rowSpan`. Some are new and behind the _Grid interactivity_ experimental flag: `rowCount`, `columnStart`, and `rowStart`.

In WP 6.6 the attributes are consistently a string even though they always contain a number, e.g. `'3'`.

In the plugin, and especially with the experimental flag enabled, the attributes alternate between being a string e.g. `'3'` and being a number e.g. `3` because we now frequently perform arithmetic on these values.

The inconsistency leads to a terrible developer experience. It's difficult to write bug-free code. You never know what data type these values are, you frequently run into `NaN` errors, and our code ends up littered with `parseInt` calls.

I propose that we consistently store these attributes as numbers.

This PR:

- Updates the grid layout code to ensure that we're always giving `setAttributes( ... )` a number.
- Adds a check that will throw an error if a non-number ends up in block attribute state.
- Removes unnecessary `parseInt` calls now that we can be sure we're working with numbers.
- Migrates existing `columnCount`, `columnSpan`, and `rowSpan` attributes that are out there in the wild from string to number. (These were the ones that shipped in WP 6.6.)

To test, do a complete run-through of the grid functionality (flag turned on and off) locally using `nom run dev` (so that the assertions will run) and ensure that there are no crashes. Also check that grid markup that was saved with the plugin turned off continues to work with the plugin turned back on.

### DIFF
--- a/packages/block-editor/src/layouts/grid.js
+++ b/packages/block-editor/src/layouts/grid.js
@@ -125,6 +125,23 @@ export default {
 			rowCount = null,
 		} = layout;
 
+		// Check that the grid layout attributes are of the correct type, so that we don't accidentally
+		// write code that stores a string attribute instead of a number.
+		if ( process.env.NODE_ENV === 'development' ) {
+			if (
+				minimumColumnWidth &&
+				typeof minimumColumnWidth !== 'string'
+			) {
+				throw new Error( 'minimumColumnWidth must be a string' );
+			}
+			if ( columnCount && typeof columnCount !== 'number' ) {
+				throw new Error( 'columnCount must be a number' );
+			}
+			if ( rowCount && typeof rowCount !== 'number' ) {
+				throw new Error( 'rowCount must be a number' );
+			}
+		}
+
 		// If a block's block.json skips serialization for spacing or spacing.blockGap,
 		// don't apply the user-defined value to the styles.
 		const blockGapValue =
@@ -290,18 +307,32 @@ function GridLayoutColumnsAndRowsControl( {
 						<NumberControl
 							size="__unstable-large"
 							onChange={ ( value ) => {
-								/**
-								 * If the input is cleared, avoid switching
-								 * back to "Auto" by setting a value of "1".
-								 */
-								const validValue = value !== '' ? value : '1';
-								onChange( {
-									...layout,
-									columnCount:
-										window.__experimentalEnableGridInteractivity
-											? parseInt( value, 10 ) || null
-											: parseInt( validValue, 10 ),
-								} );
+								if (
+									window.__experimentalEnableGridInteractivity
+								) {
+									// Allow unsetting the column count when in auto mode.
+									const defaultColumnCount = isManualPlacement
+										? 1
+										: null; // TODO: Why not undefined?
+									const newColumnCount =
+										value === ''
+											? defaultColumnCount
+											: parseInt( value, 10 );
+									onChange( {
+										...layout,
+										columnCount: newColumnCount,
+									} );
+								} else {
+									// Don't allow unsetting the column count.
+									const newColumnCount =
+										value === ''
+											? 1
+											: parseInt( value, 10 );
+									onChange( {
+										...layout,
+										columnCount: newColumnCount,
+									} );
+								}
 							} }
 							value={ columnCount }
 							min={ 0 }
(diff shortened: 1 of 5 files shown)


packages/blocks/src/api/parser/convert-legacy-block.js
```diff
@@ -77,6 +77,35 @@ export function convertLegacyBlockNameAndAttributes( name, attributes ) {
 		newAttributes.legacy = true;
 	}
 
+	// Grid layout attributes were stored as strings in WP 6.6. Convert them to numbers.
```
noisysocks: Apparently `columnCount` shipped in 6.3. Need to check that though.

```suggestion
	// Grid layout attributes were stored as strings in WP 6.3-6.6. Convert them to numbers.
```

packages/block-editor/src/layouts/grid.js
```diff
@@ -290,18 +307,32 @@ function GridLayoutColumnsAndRowsControl( {
...
-											? parseInt( value, 10 ) || null
-											: parseInt( validValue, 10 ),
-								} );
+								if (
+									window.__experimentalEnableGridInteractivity
+								) {
+									// Allow unsetting the column count when in auto mode.
+									const defaultColumnCount = isManualPlacement
+										? 1
+										: null; // TODO: Why not undefined?
```
noisysocks: Need to look into this. Shouldn't we remove the attribute instead of making it `null`?

tellthemachines: Yeah it should be unsettable here, we're no longer using columnCount to distinguish Auto from Manual mode so it shouldn't matter if it exists or not.

packages/blocks/src/api/parser/convert-legacy-block.js
tellthemachines: It did! See [here](https://make.wordpress.org/core/2023/07/14/layout-updates-in-the-editor-for-wordpress-6-3/) for details. The manual mode (using `columnCount`) was added to Post Template too, so it was definitely in use from that point onwards. The good news is in TT4 markup (see [here](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentytwentyfour/patterns/posts-images-only-3-col.php#L20) for instance) it's always a number.

packages/blocks/src/api/parser/convert-legacy-block.js
```diff
@@ -77,6 +77,35 @@ export function convertLegacyBlockNameAndAttributes( name, attributes ) {
...
+			...newAttributes.layout,
+			columnCount: parseInt( attributes.layout.columnCount, 10 ),
+		};
+	}
+	if ( typeof attributes.style?.layout?.columnSpan === 'string' ) {
+		newAttributes.style = {
+			...newAttributes.style,
+			layout: {
+				...newAttributes.style.layout,
+				columnSpan: parseInt( attributes.style.layout.columnSpan, 10 ),
```
tellthemachines: Should we check the string isn't empty here? Both `columnSpan` and `rowSpan` can be empty strings if the input has been cleared. It shouldn't make much difference in practice but it does pop up a console warning.

tellthemachines (commented):
This is generally working well, and the code looks much nicer! The only issue I'm seeing is with the range slider on the column control defaulting to mid-range when the number input is cleared:

Screenshot 2024-07-04 at 4 54 07 PM

The practical difference is that by removing the `parseInt()` from the value we pass the range control, it's now receiving `null` when the field is cleared, instead of `NaN` which it seems to deal with better. That seems more like a bug with RangeControl, but in any case we should make sure to pass it either `NaN` or `0` if `columnCount` is null 😅

packages/blocks/src/api/parser/convert-legacy-block.js
noisysocks: Ah good point. I'll add a `isNaN` check in case it's an empty string or not a number.

packages/block-editor/src/layouts/grid.js
noisysocks: Yep makes sense. I'll make it remove the attribute instead of setting it to null, though.

noisysocks:
How's that @tellthemachines? I fixed your feedback and made it so that we're consistently unsetting the attributes when you clear the InputControl instead of setting it to an empty string or null. Definitely worth testing again, this stuff is fraught.

tellthemachines (approved):
This is working well for me now!

PR #63114, opened 2024-07-04
Group: Update block example

### LINKED ISSUE #63035: Section styles: improve (or remove) preview

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/63035

## What?

Updates the inner blocks for the default Group example block while keeping the existing example for the Row, Stack, and Grid block variations.

Alternative to https://github.com/WordPress/gutenberg/pull/63039.

## Why?

The current example doesn't add a lot of value, especially when a user is attempting to preview block style variations. See #63035

## How?

- Move existing example to the block variation definitions
- Update the default block example

## Testing Instructions
1. Activate a theme that defines block style variations for group blocks such as Assembler
2. In the site editor, select a group block and open the Styles tab in the selector.
3. Hover over the Block Styles buttons to trigger the previews using the example block
4. Ensure these previews show the new example
5. Open the main block inserter via the plus icon in the top left
6. Filter the blocks by "Group"
7. Confirm the standard Group block displays the new example
8. Check that each Group block variation (Row, Stack, and Grid) all show the original example.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/60436221/099edc24-1182-4044-a038-b74694ccfc92

## Possible Follow-ups

- Update the block inspector's preview to reflect the current block variation's example

### DIFF
--- a/packages/block-library/src/group/index.js
+++ b/packages/block-library/src/group/index.js
@@ -22,53 +22,49 @@ export { metadata, name };
 export const settings = {
 	icon,
 	example: {
+		attributes: {
+			layout: {
+				type: 'constrained',
+				justifyContent: 'center',
+			},
+			style: {
+				spacing: {
+					padding: {
+						top: '1em',
+						right: '1em',
+						bottom: '1em',
+						left: '1em',
+					},
+				},
+			},
+		},
 		innerBlocks: [
 			{
-				name: 'core/paragraph',
+				name: 'core/heading',
 				attributes: {
-					customTextColor: '#cf2e2e',
-					fontSize: 'large',
-					content: __( 'One.' ),
+					content: __( 'La Mancha' ),
+					textAlign: 'center',
+				},
+			},
+			{
+				name: 'core/image',
+				attributes: {
+					url: 'https://s.w.org/images/core/5.3/Glacial_lakes%2C_Bhutan.jpg',
 				},
 			},
 			{
 				name: 'core/paragraph',
 				attributes: {
-					customTextColor: '#ff6900',
-					fontSize: 'large',
-					content: __( 'Two.' ),
+					align: 'center',
+					content: __(
+						'In a village of La Mancha, the name of which I have no desire to call to mind, there lived not long since one of those gentlemen that keep a lance in the lance-rack, an old buckler, a lean hack, and a greyhound for coursing.'
+					),
 				},
 			},
 			{
-				name: 'core/paragraph',
+				name: 'core/button',
 				attributes: {
-					customTextColor: '#fcb900',
-					fontSize: 'large',
-					content: __( 'Three.' ),
-				},
-			},
-			{
-				name: 'core/paragraph',
-				attributes: {
-					customTextColor: '#00d084',
-					fontSize: 'large',
-					content: __( 'Four.' ),
-				},
-			},
-			{
-				name: 'core/paragraph',
-				attributes: {
-					customTextColor: '#0693e3',
-					fontSize: 'large',
-					content: __( 'Five.' ),
-				},
-			},
-			{
-				name: 'core/paragraph',
-				attributes: {
-					customTextColor: '#9b51e0',
-					fontSize: 'large',
-					content: __( 'Six.' ),
+					text: __( 'Read more' ),
 				},
 			},
 		],
--- a/packages/block-library/src/group/index.js
+++ b/packages/block-library/src/group/index.js
@@ -30,10 +30,10 @@ export const settings = {
 			style: {
 				spacing: {
 					padding: {
-						top: '1em',
-						right: '1em',
-						bottom: '1em',
-						left: '1em',
+						top: '4em',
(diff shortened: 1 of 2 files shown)


packages/block-library/src/group/index.js
```diff
@@ -22,53 +22,49 @@ export { metadata, name };
...
 				},
 			},
 			{
-				name: 'core/paragraph',
+				name: 'core/image',
 				attributes: {
-					customTextColor: '#00d084',
-					fontSize: 'large',
-					content: __( 'Four.' ),
+					url: 'https://s.w.org/images/core/5.3/Glacial_lakes%2C_Bhutan.jpg',
```
ramonjd: I don't know if it matters, but maybe we could marry the text with the image?

> Lake Bhutan is a high-altitude lake in Bhutan, known for its pristine beauty, clear waters, and serene surroundings, attracting nature lovers and trekkers.

Or, there are some free use images related to Don Quixote over at https://www.oldbookillustrations.com/search/?q=don+quixote

packages/block-library/src/group/index.js
```diff
@@ -22,53 +22,49 @@ export { metadata, name };
...
-			{
-				name: 'core/paragraph',
-				attributes: {
-					customTextColor: '#cf2e2e',
-					fontSize: 'large',
-					content: __( 'One.' ),
-				},
+		attributes: {
+			layout: {
+				type: 'constrained',
```
ramonjd, line 27: Simple and looking pretty good. 

https://github.com/WordPress/gutenberg/assets/6458278/b2bb9810-6e36-407c-a734-764201088ad0

ramonjd (commented):
Just a smoke test to see how it looks. Seems to behave pretty well.

Other ideas might be to use a background image, or add a border. I'll leave it to the designers' eyes 😄

packages/block-library/src/group/index.js
aaronrobertshaw: > maybe we could marry the text with the image?

100%. This was my intention and why I left this in draft status 🙂

aaronrobertshaw: It would be nice if all the blocks' example content had some level of consistency or theme. It's pretty random at the moment 😅

richtabor:
>  Possibly update the block inspector's preview to reflect the current block variation's example

I think this would be a nice follow-up.

richtabor:
I pushed some small tweaks: 

CleanShot 2024-07-07 at 12 03 24

packages/block-library/src/group/variations.js
```diff
@@ -4,6 +4,59 @@
 import { __, _x } from '@wordpress/i18n';
 import { group, row, stack, grid } from '@wordpress/icons';
 
+const example = {
```
richtabor, line 7: Is this example being used at all? 

I have a row block here and it's using the current index.js example. Stack and grid are the same. 

CleanShot 2024-07-07 at 12 07 17

richtabor (commented):
Left one question. Otherwise this is nice. Follow-up may be to define examples for the variations; although if we could represent the selected block as the example here, that may be a better result.

richtabor:
@aaronrobertshaw is anything blocking this? seems like a nice quick update.

packages/block-library/src/group/variations.js
aaronrobertshaw: It's used for previewing the different block variations in the block inserter.

Steps 5 - 8 in the test instructions cover this aspect. To save you having to fire up the PR locally you can see the original example with the main block inserter from the 30s mark on in the description's demo video.

aaronrobertshaw:
> is anything blocking this? seems like a nice quick update.

Only the fact I was waiting on decisions and then the WP 6.6 release caused this to drop down the priority list.

If you are happy with the current state of the example in this PR, we can easily merge and iterate further via follow-ups. Someone else may need to pick up the follow-ups as I'll be AFK for a couple of weeks in the near future.

I've rebased this PR and flagged it ready for review.

packages/block-library/src/group/variations.js
aaronrobertshaw: This might be easier still 🙂 

https://github.com/user-attachments/assets/2c056d26-9196-4b6a-b6b1-f4c8124ce7d3

ramonjd (approved):
I think it's a good improvement. Let's iterate!

Screenshot 2024-07-30 at 10 38 08 AM

PR #63117, opened 2024-07-04
Block Bindings: Improve the way block bindings sources are registered

SantosGuillamot:
## What?
This pull request changes the way block bindings sources are registered to follow the same pattern as `blockType`, `blockVariation`, `blockCollection`, etc, and much more similar to how it is handled in the server.

## Why?
The current way of registering them doesn't feel right. The registration currently lives outside of a hook or a component in the editor package: [link](https://github.com/WordPress/gutenberg/blob/45a7fba610962b68d81e15e1d8b32d55567607b5/packages/editor/src/bindings/index.js#L1). And it uses directly actions and selectors defined in the store.

## How?
Trying to follow a similar approach to how blocks are registered, I did:
* Create new `registerBlockBindingsSource`, `unregisterBlockBindingsSource`, `getBlockBindingsSource`, `getBlockBindingsSources` functions similar to the existing ones for blockTypes exported by `@wordpress/blocks`. There, I added some validation similar to the one we have in the server: [link](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-block-bindings-registry.php#L98-L183).
* Create a new `registerCoreBlockBindingsSources` function to register the current core sources, post meta and pattern overrides. This function is similar to `registerCoreBlocks`. It is placed in `@wordpress/editor` because some core sources, like post meta, might need to access the editor store.
* Call the `registerCoreBlockBindingsSources` from `edit-site` and `edit-post` as suggested by Riad [here](https://github.com/WordPress/gutenberg/pull/63117#discussion_r1665357520).

## Follow-ups
* Add proper docs.
* Polish the APIs before making them public.

## Testing Instructions
Automated tests should pass.


packages/editor/src/components/provider/index.js
```diff
@@ -223,6 +225,9 @@ export const ExperimentalEditorProvider = withRegistryProvider(
 			setRenderingMode,
 		} = unlock( useDispatch( editorStore ) );
 		const { createWarningNotice } = useDispatch( noticesStore );
+		const { registerBlockBindingsSource } = unlock(
+			useDispatch( blocksStore )
+		);
```
youknowriad: There's a subtlety here. The "blocks" store is a global store. The same store is used between all `EditorProvider`  instances while the `EditorProvider` can also create multiple code/editor stores. (depending on a prop).

So the question to be asked here, what if two editor providers are loaded at the same time, do we want to call `registerBlockBindingsSource` twice. It's kind of the exact same reason why we don't register blocks here (in the EditorProvider) and instead we register them in `edit-site/src/index` and `edit-post/src/index`

So I'm wondering if we should just do like the block registration instead. All of these things (bindings, blocks, format library) are probably better registered in the same way.

SantosGuillamot: That would make sense. I'll try to move it to the `edit-site` and `edit-post` packages. Something I'm wondering is how external developers would hook into that once the APIs are public, but I can take a look at that as well.

Thanks a lot for the guidance!

SantosGuillamot: I've changed the approach trying to simulate something similar to how I understood block registration works. I've also updated the description of the pull request to reflect that. Basically, what I did:

Trying to follow a similar approach to how blocks are registered, I did:

* Create a new `registerBlockBindingsSource` function exported by `@wordpress/blocks`. There, I added some validation similar to the one we have in the server: [link](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-block-bindings-registry.php#L98-L183).
* Create a new `registerCoreBlockBindingsSources` function to register the current core sources, post meta and pattern overrides. This function uses the previous one and it is placed in `@wordpress/editor` because some sources, like post meta, might need to access the editor store.
* Call the `registerCoreBlockBindingsSources` from `edit-site` and `edit-post` as suggested.

The code should be simplified once the APIs are public and we should add more functions like `unregisterBlockBindingsSource`, but I believe that could be handled in another pull request.

Please, let me know if that's not what you had in mind.

artemiomorales:
This approach looks clean and makes sense to me — it feels good and is less cognitive overhead seeing `registerCoreBlockBindingsSources()` alongside the other initialization logic in `edit-site` and `edit-post` packages. Will continue following the discussion 🙂

SantosGuillamot:
I made a bunch of changes to make it more similar to how block types, variations, or collections are handled and I added a first set of unit tests similar to the ones we have in the server.

As these APIs are still private, I'm marking this pull request as ready for review to start gathering more feedback.

There are other pull requests that deal with the bindings editor APIs that may be affected by the outcome of this one.

artemiomorales (approved):
The changes continue looking good to me. I'll drop an approval but am also looking forward to any feedback that folks with more experience may have 🙏

packages/blocks/src/api/registration.js
```diff
@@ -758,3 +758,178 @@ export const registerBlockVariation = ( blockName, variation ) => {
...
+		getPlaceholder,
+		canUserEditValue,
+	} = source;
+
+	// Check if the source is already registered.
+	const existingSource = unlock(
+		select( blocksStore )
+	).getBlockBindingsSource( name );
+	if ( existingSource ) {
+		console.error(
```
cbravobernal, line 808: Should this be translatable and dev only? Can be done in a follow-up.

SantosGuillamot: Maybe? 🤔 I checked the rest of the `console.error` and `console.warn` Gutenberg has and it seems they are never translated. Maybe it is a good discussion to have globally? By the way, what do you mean with dev only?

gziolo: The most optimal would be:

```js
import { warning } from `@wordpress/warning`; 
warning( 'This is the message...' );
```

It's a hint for a developer, but the code continues to work after making the best informed decision.

SantosGuillamot: I can make that change 🙂 

I just followed what is being done in the `registerBlockType` function: [link](https://github.com/WordPress/gutenberg/blob/e6e7bb05206a94e56cd9f27405a54ae99faf7758/packages/blocks/src/api/registration.js#L227-L242).

cbravobernal (approved):
The console warnings are a nitpick.

packages/blocks/src/api/registration.js
SantosGuillamot: I am not sure when we should throw a warning and when an error. In both cases, the editor won't break but the bindings won't work as expected.

**Error**
Screenshot 2024-07-16 at 11 59 55


**Warning**
Screenshot 2024-07-16 at 11 59 10

I'll create a follow-up pull request to properly discuss that.

SantosGuillamot:
Merging this now. Happy to explore other paths if we receive feedback after merging. Good news is that the APIs are still private.

packages/blocks/src/api/registration.js
cbravobernal: That's a good question. Interactivity API uses warnings, whether the interactivity worked or not.

gziolo: It would be great to unify. `@wordpress/warning` was introduced not that long ago, so historically, people would use `console.error` or `console.warn` based on their personal preferences. I would simplify it and use `warning` from `@wordpress/warning` as it doesn't add too much value on production anyway. The other benefit is that `@wordpress/warning` uses hooks internally so 3rd party plugin can intercept these warning. If I recall correctly @johnbillion integrated that into [Query Monitor plugin](https://wordpress.org/plugins/query-monitor/) that has 200k active installs.

packages/blocks/src/api/registration.js
```diff
@@ -758,3 +758,178 @@ export const registerBlockVariation = ( blockName, variation ) => {
...
 };
+
+/**
+ * Registers a new block bindings source with an object defining its
+ * behavior. Once registered, the source is available to be connected
+ * to the supported block attributes.
+ *
+ * @param {Object}   source                    Properties of the source to be registered.
+ * @param {string}   source.name               The unique and machine-readable name.
+ * @param {string}   source.label              Human-readable label.
```
gziolo, line 769: Label should be optional as it can be provided from the server.

packages/blocks/src/api/registration.js
```diff
@@ -758,3 +758,178 @@ export const registerBlockVariation = ( blockName, variation ) => {
...
+ * ```js
+ * import { _x } from '@wordpress/i18n';
+ * import { registerBlockBindingsSource } from '@wordpress/blocks'
+ *
+ * registerBlockBindingsSource( {
+ *     name: 'plugin/my-custom-source',
+ *     label: _x( 'My Custom Source', 'block bindings source' ),
+ *     getValue: () => 'Value to place in the block attribute',
+ *     setValue: () => updateMyCustomValue(),
+ *     setValues: () => updateMyCustomValuesInBatch(),
```
gziolo, line 786: Have you decided on the final API? There was discussion about deciding about using with voices in favor of `getValues` and `setValues`.

packages/blocks/src/api/registration.js
SantosGuillamot: I made it optional in the pull request to bootstrap the source from the server: [link](https://github.com/WordPress/gutenberg/pull/63470/files#diff-30ce126b90014d659302087765876898f16562219f796796ed407fdb346c8e4dR769). Until then, I think label should be required, right?

packages/blocks/src/api/registration.js
SantosGuillamot: I feel confident about using just `getValues` and `setValues`. I was just rebasing the pull request to adapt the registration there: [link](https://github.com/WordPress/gutenberg/pull/63185/files#diff-30ce126b90014d659302087765876898f16562219f796796ed407fdb346c8e4d). My plan is to make the changes needed, leave a comment sharing my opinion, and open it for review.

packages/blocks/src/api/registration.js
SantosGuillamot: That makes sense. I will create a pull request to change all the `console.error` and `console.warn` in the registration file to use `@wordpress/warning`.

packages/blocks/src/store/reducer.js
```diff
@@ -372,18 +372,22 @@ export function collections( state = {}, action ) {
...
+			return {
+				...state,
+				[ action.name ]: {
+					label: action.label,
+					getValue: action.getValue,
+					setValue: action.setValue,
+					setValues: action.setValues,
+					getPlaceholder: action.getPlaceholder,
+					canUserEditValue:
+						action.canUserEditValue || ( () => false ),
```
gziolo, line 386: What's the reasoning for providing a default handler only in this place? `getValue`, `setValue`, `setValues` ang `getPlaceholder` are all also optional so they will be undefined. Should the default handler for `canUserEditValue` be moved to the selector instead?

packages/blocks/src/api/registration.js
gziolo: That works if you have a follow-up lined up.

packages/blocks/src/api/registration.js
gziolo: It wasn't very clear what happened to the open discussions that didn't reach any conclusions. When landing changes, it would be helpful for all observers to explain the action plan for next steps in case they would like to continue following the work.

gziolo: I think that's the challenge when documenting the current API that isn't considered a final one.

gziolo:
It's nice to see how the API is shaping up. There is still some follow-up work necessary, but for me, it's very close to what we can see in WordPress 6.7. Great progress.

packages/blocks/src/api/registration.js
SantosGuillamot: Follow-up created here: https://github.com/WordPress/gutenberg/pull/63610. Although I encountered some issues.

packages/blocks/src/api/registration.js
SantosGuillamot: Yes, there is a follow-up already working. I just need to rebase and it should be ready.

packages/blocks/src/api/registration.js
SantosGuillamot: I kept them open to leave time for feedback. I wanted to post a new comment with what I consider the conclusion.

packages/blocks/src/store/reducer.js
SantosGuillamot: For sources that aren't registered or don't provide that property, we want to lock the editing when a paragraph (or other block) is connected to it. 

I didn't think about moving it to the selector, I can give that a try. I guess it would require changing both `getAllBlockBindingsSources` and `getBlockBindingsSource`: [link](https://github.com/WordPress/gutenberg/blob/9e73c1158bb8eb79d6415743a7519fad2d2ffef1/packages/blocks/src/store/private-selectors.js#L197-L211).

SantosGuillamot: I created a follow-up pull request for this as well: [link](https://github.com/WordPress/gutenberg/pull/63628). I didn't move the logic to the selector, but I think I prefer that approach better. Let me know what you think 🙂

packages/blocks/src/api/registration.js
SantosGuillamot: I have just posted [this comment](https://github.com/WordPress/gutenberg/pull/63185#issuecomment-2231143154) and marked it as ready for review.

PR #63123, opened 2024-07-04
Allow ariakit and framer motion imports in the components package

ciampo:
## What?

Tweak the ESLint configuration to allow, only for the components package, importing from `ariakit` and `framer-motion`, as initially suggested in https://github.com/WordPress/gutenberg/pull/62919#discussion_r1665489822

## Why?

For better DevEx. The reason they are disabled in the first place is that they should be only consumer by the `@wordpress/components` package, so this ESlint config change just puts that advice in practive.

## How?

- Added a rule override in the ESLint config
- Removed unnecessary `eslint-disable` comments for `ariakit` and `framer-motion` imports in the components package

## Testing Instructions

- Make sure that the project builds without ESLint errors
- Make sure that there are no leftover unnecessary `eslint-disable` comments left in the components package
- Make sure that `ariakit` and `framer-motion` imports still generate an ESLint error when imported outside of the components package

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -425,6 +425,9 @@ module.exports = {
 			rules: {
 				'no-restricted-imports': [
 					'error',
+					// The `ariakit` and `framer-motion` APIs are meant to be consumed via
+					// the `@wordpress/components` package, hence why importing those
+					// dependencies should be allowed in the components package.
 					{
 						paths: restrictedImports.filter(
 							( { name } ) =>
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -421,6 +421,23 @@ module.exports = {
 			extends: [ 'plugin:ssr-friendly/recommended' ],
 		},
 		{
+			files: [ 'packages/components/src/**' ],
+			rules: {
+				'no-restricted-imports': [
+					'error',
+					{
+						paths: restrictedImports.filter(
+							( { name } ) =>
+								! [
+									'@ariakit/react',
+									'framer-motion',
+								].includes( name )
+						),
+					},
+				],
+			},
+		},
+		{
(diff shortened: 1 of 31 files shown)


.eslintrc.js
```diff
@@ -420,6 +420,23 @@ module.exports = {
...
+							( { name } ) =>
+								! [
+									'@ariakit/react',
+									'framer-motion',
+								].includes( name )
+						),
+					},
+				],
+			},
+		},
```
tyxla: It may be useful to add a comment that we're re-exporting them in the package so we're allowing usage inside the package. Right now it's not so obvious without looking deeper into why they were forbidden in the first place.

tyxla (approved):
Thanks @ciampo, looks good 🚀 

It's a bit of a bummer that the ESLint config gets more and more complicated, but the change makes sense and I appreciate not having to ignore each and every import in the components package.

.eslintrc.js
ciampo: Sure, I'll add a comment

DaniGuardiola:
This makes sense, though I agree that the config is getting complicated. We could simplify it by creating a package-specific config file that extends the root one, and then apply the overrides there.

tyxla:
> This makes sense, though I agree that the config is getting complicated. We could simplify it by creating a package-specific config file that extends the root one, and then apply the overrides there.

Not sure we wanna go that way though. If we want to upgrade to ESLint v9+, we might want to inspect migrating to a [flat config](https://eslint.org/blog/2023/10/flat-config-rollout-plans/) instead. They're planning to [remove `.eslintrc` support in v10](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#eslintrc-removed-in-eslint-v10.0.0) FWIW.

PR #33701, opened 2021-07-26
Components: promote `ItemGroup`

ciampo:
## Description

- Promote the `ItemGroup` and `Item` components from the `packages/components/src/ui` folder to the `packages/components/src` folder
- Refactor the folder structure, so that each component has its own folder
- Fix a bug with the CSS pseudo-elements selectors, where the `[first|last]-of-type` selectors where causing an issue when mixing and matching different HTML element types (e.g. when one item has `isAction={true}` and the rest of the elements don't). To fix this bug, every `Item` is now wrapped in a `div` component to make sure that all children of `ItemGroup` are of the same type
- Add README and JSDocs comments to types
- Add a few basic unit tests
- Export component from the `@wordpress/components` package

## How has this been tested?

- All tests pass
- Run storybook locally, and make sure that the components behaves in the same way as it does currently on [production storybook](https://wordpress.github.io/gutenberg/?path=/story/components-experimental-itemgroup--default)

## Screenshots 

Screenshot 2021-07-26 at 20 31 27

## Types of changes

Refactor

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #63125, opened 2024-07-04
Fix item font family item height in the sidebar

matiasbenedetto:
## What?
Fix item font family item height in the sidebar

## Why?
To match the other components height

## How?
Setting the large prop in the component

## Screenshots or screencast 

| Before | After |
| --- | --- |
| Screenshot from 2024-07-04 16-01-39 | Screenshot from 2024-07-04 15-58-19 |


mikachan (commented):
I'd like to defer to the design team for final sign-off, but this looks like a good visual improvement to me 👍

jasmussen:
This PR makes the itemgroup items closer to 40px than 32, so that's good:

Screenshot 2024-07-04 at 14 48 04

A few things, but they may be more on the component side, so CC @tyxla:

* No ItemGroup items should ever be any size _other_ than 40px. We don't have a 32px component at all, so I'm curious why we need to add the "large" prop here.
* ItemGroup items should be _exactly_ 40px, not 40.5 as is the case here, I suspect it's different from platform to platform based on font height subpixels. It can be a min-height, potentially, to allow edge-cases where the text needs to wrap or otherwise.

But overall, the intent of this PR is right, I'm just not sure this is the right fix on the code level. Seems like this should be fixed in the component itself?

tyxla (approved):
Code-wise, this looks good 👍 The variety we're seeing in the total height is not related to this PR.

tyxla:
If we need to always have the same item height, we might have to play with the padding math right there:

https://github.com/WordPress/gutenberg/blob/c94f863b5df5034524e86490427864779b12754e/packages/components/src/item-group/styles.ts#L95-L117

If we're aiming at a particular item height, it might actually be more straightforward to not rely on padding at all, but rather to apply item height directly and center the contents vertically. 


cc @WordPress/gutenberg-components

ciampo:
> Code-wise, this looks good 👍 The variety we're seeing in the total height is not related to this PR.

Agreed!


> No ItemGroup items should ever be any size other than 40px. We don't have a 32px component at all, so I'm curious why we need to add the "large" prop here.
> ItemGroup items should be exactly 40px, not 40.5 as is the case here, I suspect it's different from platform to platform based on font height subpixels. It can be a min-height, potentially, to allow edge-cases where the text needs to wrap or otherwise.
>
> If we're aiming at a particular item height, it might actually be more straightforward to not rely on padding at all, but rather to apply item height directly and center the contents vertically.

It looks like the `size` prop has been around since the introduction of the component in the repo (porting from G2) in #33701.

As we're looking at [stabilizing `ItemGroup`](https://github.com/WordPress/gutenberg/pull/61060), this is a great time to apply the `40px` default height, potentially deprecate the `size` prop, and understand how we want the `Item` component to behave in terms of its `height`:
 
- should it be considered a _minimum height_, which would allow its children to push the `Item` height to be taller if necessary?
- or should it be a _fixed height_, meaning that children that don't fit on one line would need to find alternative solutions (ie. text truncation, alternative designs...)

tyxla:
If the `size` prop isn't widely used, we can definitely deprecate. Then, the min-height option @ciampo proposed seems safe enough to me as a good next step.

t-hamano:
@matiasbenedetto @mikachan 

Can you resolve the conflicts and merge this PR? In trunk, fonts are grouped by source.

image

By the way, there is an issue proposing to revert this grouping: #63505

mikachan:
> Can you resolve the conflicts and merge this PR?

Done! I'll enable auto-merge.

t-hamano:
@mikachan 

Sorry, I missed it, but don't we need to change the height of the custom fonts? Here it is:

https://github.com/WordPress/gutenberg/pull/63125/files#diff-bcf63753fcc766be009c30f4fa40f011bcfdb4170a90fe419ab69370df7cf479R90

mikachan:
> Sorry, I missed it, but don't we need to change the height of the custom fonts? Here it is:

Thanks @t-hamano! I missed it too, I think I managed to push the change just before the auto-merge happened 👀

PR #63126, opened 2024-07-04
Interactivity API: add tests for the getElement() attributes

luisherranz:
## What?

It is still unclear how the attribute object returned by `getElement()` works. We should reconsider how it functions:
- Does it return the props of the virtual DOM or the attributes of the DOM?
- Can the attributes be modified?
- Are attributes reactive?
- If so, what happens when they are modified in the DOM (as opposed to via `data-wp-bind`)?

## Why?

We've only considered the main use case (read `data` attributes on getters) so far, and it'd be good to think about all its implications to finalize this API.

## How?

I've only added a couple of e2e tests so far. We can add more and change the implementation as we decide how to move forward.

### DIFF
--- a/packages/e2e-tests/plugins/interactive-blocks/get-element/block.json
+++ b/packages/e2e-tests/plugins/interactive-blocks/get-element/block.json
@@ -1,0 +1,15 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 2,
+	"name": "test/get-element",
+	"title": "E2E Interactivity tests - getElement",
+	"category": "text",
+	"icon": "heart",
+	"description": "",
+	"supports": {
+		"interactivity": true
+	},
+	"textdomain": "e2e-interactivity",
+	"viewScriptModule": "file:./view.js",
+	"render": "file:./render.php"
+}
--- a/packages/e2e-tests/plugins/interactive-blocks/get-element/render.php
+++ b/packages/e2e-tests/plugins/interactive-blocks/get-element/render.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * HTML for testing the `store` function.
+ *
+ * @package gutenberg-test-interactive-blocks
+ */
+?>
+
+<div data-wp-interactive="test/get-element">
+	<div
+		data-testid="read from attributes"
+		data-some-value="Initial value"
+		data-wp-bind--data-some-value="state.dataSomeValue"
+		data-wp-text="state.someValue"
+	></div>
+	<button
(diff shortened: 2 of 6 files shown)


michalczaplinski:
I gave it some thought, and I don't have any definite answers but rather just some loose thoughts. I'm still getting re-acquainted with the internals of the Interactivity API after a long time away from it, so there's a chance I got sth wrong :)

### `attributes` naming is currently confusing. 

We are not currently returning actual attributes from `getElement()` but instead the props of the vDOM. So, we have a few options:
- Add a new property like `props` and "deprecate" the `attributes` (but we cannot remove them anymore).
- Update the interface of `attributes` to return element attributes. This brings me to the next point:

### Returning actual `attributes` ?

If we wanted to return the attributes **of the HTML** , we'd have to figure out a way for the runtime to wait until the DOM node is mounted. It's the same reason why `ref` initially returns `null` - it's because it has not been attached to the actual DOM element yet. 

So, to access the actual HTML attributes, folks can already use `ref.attributes` (subject to the following limitation):

> #### ref
> ref is the reference to the DOM element as an [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement). It is equivalent to useRef in Preact or React, so it can be null when ref has not been attached to the actual DOM element yet, i.e., when it is being hydrated or mounted.

Therefore, changing the `someValue` getter to something like the following _should_ pass the tests in this PR:

```js
get someValue() {
	const { ref } = getElement();
	return ref.dataset.someValue;
}
```

However, it hits the same problem of the `ref` not being attached to the DOM yet.

luisherranz:
I have been reviewing this with @darerodz carefully and we think the current logic is fine since it only contains the attributes of the original element. In that case, it seems correct to keep calling it attributes and not props.

Now then, since these properties are present in the Virtual DOM, they are actually the attributes that the Interactivity API is aware of. Initially, this might seem incorrect because if someone modifies the DOM imperatively, they could become unsynchronized. But after careful consideration, we believe it's safer this way because it will force people not to make imperative modifications to the DOM but to use the Interactivity API (`data-wp-bind`) to modify the DOM and have that reflected in the attributes that `getElement` returns.

Similarly, we believe that making the attributes reactive is not a good idea since it's a client-side feature and could cause problems with server-side rendering. Thus, we think the best approach is to keep them read-only and to ensure that the modification of the DOM is done with a part of the state linked to the `data-wp-bind` directives to ensure that server-side rendering is always correct and to avoid exposing many ways to do the same thing, which would be confusing.

In summary, there isn't much more to do in this pull request except to add some tests to confirm the current behavior.

Issue #58012, opened 2024-01-19
Data views table layout: Enable fields to be visually combined with title


jameskoster:
In the Templates data view the spacing between title and description can become awkward and difficult to read:

<img width="1202" alt="description" src="https://github.com/WordPress/gutenberg/assets/846565/a3824d66-8077-464b-99e2-2a0d115ce07b">

Similarly, in the Pages data view the featured images feels a bit isolated from the title, and has no material benefit living a dedicated column:

<img width="677" alt="fimg" src="https://github.com/WordPress/gutenberg/assets/846565/200abb37-e0e9-45b4-9043-995724b4017b">

In both cases, it could be good to visually combine these fields into a single cell. E.g.:

<img width="921" alt="Screenshot 2024-05-23 at 11 09 41" src="https://github.com/WordPress/gutenberg/assets/846565/1723ddee-3138-41ab-b5e9-e4d17465a8d1">

jameskoster:
Noting that this problem needs more design. 

I'm a bit concerned that disabling/enabling fields based on layout opens a can of worms that we're not ready to tackle just yet. Instead I wonder if such a field can be made to work in all layouts. If not, another approach may be required here.

getdave:
@jameskoster Has there been any progress on this Issue that you are aware of? Do you still see this as _required_ for WP 6.5?

My understanding is that the primary task is to combine the `title` and `description` fields into a new `Summary` field. 

Just a note that we'd need to be careful about a11y when implementing this.

jameskoster:
No not _required_, mostly because the existing table display matches the new one. 

The path forward here isn't entirely clear to me. 

The visuals are improved (quite significantly) by combining certain fields in Table layout, and a 'Summary' field is one way to go about it. However, such a field makes much less sense in Grid + List layouts. It also enables users to display the summary _and all of its constituent fields_ in a single table, so it's a bit of a flaky solution.

The original suggestion was to literally combine certain fields in a single table cell as required, rather than adding the 'Summary'. I don't recall the exact reason, but @youknowriad wasn't a fan.

youknowriad:
Combining fields is fine but it's more complex. These UIs (dataviews) are generated, so combining fields mean we need to define precisely how fields can be combined, what UI to offer the user to combine fields... Is it something to allow for multiple layouts or specific to table... I think the complexity is not worth it for now at least.

colorful-tones:
[...]

Issue #57596, opened 2024-01-05
Templates: Hide description field when switching to List layout


jameskoster:
Since template descriptions can sometimes be quite long, they are not handled beautifully in data views list layout:

<img width="752" alt="Screenshot 2024-04-16 at 11 50 42" src="https://github.com/WordPress/gutenberg/assets/846565/e84db1b7-1edd-4d34-8db5-1f093c25680c">

When switching from the default layout (Grid) to List, it might be good to auto-hide the description field.

<details><summary>Original issue</summary>In the template management data view – list layout – the description isn't particularly useful for two main reasons:

1. Most descriptions are too long to be fully visible
2. Some templates do not have descriptions at all, resulting in an unhelpful `–` which takes up a lot of vertical space

<img width="1079" alt="Screenshot 2024-01-05 at 14 22 53" src="https://github.com/WordPress/gutenberg/assets/846565/c5b99b4e-b726-4909-a4c0-2617b7022d9b">

Removing the description would reduce a lot of noise and save space:

<img width="997" alt="Screenshot 2024-01-05 at 14 24 33" src="https://github.com/WordPress/gutenberg/assets/846565/0a529508-a776-4b12-a80b-2f3688428821"></details>

carolinan:
I don't think removing them is the best solution. We know that users are still struggling with concepts of index, home, and front page. The descriptions definitely have value for these users.

I am not able to see where this vertical layout is used? How do I access it? Did you mean to only hide it on narrow screens?

jameskoster:
> I am not able to see where this vertical layout is used?

Enable the "New admin views" experiment in Gutenberg then navigate to Appearance > Editor > Templates > Manage all templates. From there change the layout to "List" in the view options. 

Since this view is optimised for quickly browsing records in a more visual manner, I don't know that the description is required _in the list_. It can be accessed easily once it's possible to reach the details panel for each record (see https://github.com/WordPress/gutenberg/pull/57523).

jameskoster:
I think this is much less of an issue since Grid was made the default layout for templates, but it would be good to auto-hide the description when switching from grid to list layout.

jasmussen:
If you can invoke the normalized details panel / inspector in this view, then you'd be able to see the description there.

jameskoster:
Yup. You could also manually expose the description in view options if you really need to see it in this layout.

Souptik2001:
Hi @jameskoster @jasmussen ,

[...]

PR #63127, opened 2024-07-04
DataViews: Replace hiddenFields config with fields property instead

youknowriad:
Related #55083 
First step in #58012 

## What?

This PR replaces the `hiddenFields` property in the DataViews component with a `fields` property. And there are several reasons for this:

 - When we'll open third-party fields registration in dataviews, we'd want to explicitly list the visible fields for a view rather than hiding some. We'll have a lot to hide and a few to show.
 - `fields` (as opposed to `visibleFields` or `hiddenFields`) can allow us to do a lot more than show/hide fields. It opens the door to:
     - Reordering fields (not implemented here as it would require some UI explorations: how to change the order in the dropdown)
     - Combining fields like we want to do in #62975 
     - Absorb view specific field configuration (things like "badge fields", "media fields"...). I didn't do this in this PR but would be open if someone refactors these existing properties into the the fields property `fields: [ { field: 'date', renderAs: 'badge' } ]`

**Note**

 - For convenience (and for a partial backwards compatibility support), this PR supports showing all fields when omitting the `view.fields` property.

## Testing Instructions

Check that the different dataviews (patterns, pages...) work as expected in different layouts (grid, table, list), that you can hide/show fields in the view dropdown or in the table headers.

### DIFF
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -78,9 +78,7 @@ const DEFAULT_VIEW = {
 		field: 'title',
 		direction: 'asc',
 	},
-	// All fields are visible by default, so it's
-	// better to keep track of the hidden ones.
-	hiddenFields: [ 'preview' ],
+	fields: [ 'title', 'description', 'author' ],
 	layout: defaultConfigPerViewType[ LAYOUT_GRID ],
 	filters: [],
 };
--- a/packages/dataviews/CHANGELOG.md
+++ b/packages/dataviews/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+
+### Breaking Changes
+
+- Replace the `hiddenFields` property in the view prop of `DataViews` with a `fields` property that accepts an array of visible fields instead.
+
 ### New features
 
 - Added a new `DataForm` component to render controls from a given configuration (fields, form), and data.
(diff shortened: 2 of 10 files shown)


youknowriad:
The typescript types were very helpful in this refactoring :)

packages/edit-site/src/components/page-templates/index.js
```diff
@@ -78,9 +78,7 @@ const DEFAULT_VIEW = {
 		field: 'title',
 		direction: 'asc',
 	},
-	// All fields are visible by default, so it's
-	// better to keep track of the hidden ones.
-	hiddenFields: [ 'preview' ],
```
ellatrix, line 83: Question: what if you hide the preview field, but it's required in another layout?

youknowriad: It's still shown in the other layout (there's special treatment for media and primary fields). That said, we've been discussing with @jameskoster and @oandregal and we'll be following-up with a PR where the visible fields and fields rendering config resets when you switch layout. We found that it makes little sense to keep these when you switch layouts (as opposed to filters...)

youknowriad: Related https://github.com/WordPress/gutenberg/issues/57596#issuecomment-2134807657

ellatrix (approved):
I'm convinced and it's a much clearer API. Tests well.

PR #63129, opened 2024-07-04
ProgressBar: Fix indeterminate RTL support

tyxla:
## What?
Fix the RTL support for the indeterminate `ProgressBar` component.

## Why?
Currently on RTL, the indeterminate progress bar animates from left to right.

## How?
By parametrizing the animation by direction and taking it into account.

## Testing Instructions
* Open the `ProgressBar` component in storybook
* Verify indeterminate mode works well with LTR and RTL
* Change the `value` and verify determinate mode also works well in LTR and RTL.

None

## Screenshots or screencast 
Before:

https://github.com/WordPress/gutenberg/assets/8436925/be357996-f77d-4461-8b53-91ec9cd85475

After:

https://github.com/WordPress/gutenberg/assets/8436925/8c73bdab-5034-4b83-a04a-3e40b05cc2e2

### DIFF
--- a/packages/components/src/progress-bar/styles.ts
+++ b/packages/components/src/progress-bar/styles.ts
@@ -5,18 +5,27 @@ import styled from '@emotion/styled';
 import { css, keyframes } from '@emotion/react';
 
 /**
+ * WordPress dependencies
+ */
+import { isRTL } from '@wordpress/i18n';
+
+/**
  * Internal dependencies
  */
 import { COLORS, CONFIG } from '../utils';
 
-const animateProgressBar = keyframes( {
-	'0%': {
-		left: '-50%',
-	},
-	'100%': {
-		left: '100%',
-	},
-} );
+function animateProgressBar( isRtl = false ) {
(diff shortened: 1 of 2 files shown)


mirka (approved):
I never thought of this 😱 Good one!

packages/components/src/progress-bar/styles.ts
```diff
@@ -67,7 +76,7 @@ export const Indicator = styled.div< {
 					animationDuration: '1.5s',
 					animationTimingFunction: 'ease-in-out',
 					animationIterationCount: 'infinite',
-					animationName: animateProgressBar,
+					animationName: animateProgressBar( isRTL() ),
```
ciampo, line 79: I don't think I've ever thought of using the `isRTL()` check directly in the styled components, I always  used it in the render function and passed it as a prop. Interesting!

tyxla: It felt simpler to me, rather than having to introduce yet another prop. I'm happy to consider changing it if you feel it can be improved.

PR #63130, opened 2024-07-04
Normalize focusable disabled ToolbarButton usage

mirka:
Follow-up to #63102

## What?

Cleans up the code that was previously used to keep disabled `ToolbarButton`s focusable.

## Why?

ToolbarButtons will now always remain focusable when disabled (#63102).

## Testing Instructions

Only the change in `packages/customize-widgets/src/components/header/index.js` is somewhat unpredictable. To test this, activate the Twenty Twenty-One theme and go to the [Widgets editor](http://localhost:8888/wp-admin/widgets.php). The disabled undo/redo buttons in the toolbar should behave as expected.


tyxla (approved):
LGTM, `ToolbarButton`s no longer need the special treatment! 👍 

Thank you for the cleanup 🚀

PR #63131, opened 2024-07-04
CustomSelectControl V2 legacy adapter: fix trigger button font size

ciampo:
## What?

Part of #55023

Update the font size of the `CustomSelectControl` V2 legacy adapter to use the same underlying logic as V1

## Why?

V2 legacy adapter should match V1 styles

## How?

The [V1 implementation](https://github.com/WordPress/gutenberg/blob/b76de86ffe89b437808a7a56c48021838fcb2d8a/packages/components/src/custom-select-control/index.js#L163-L178) re-uses some [`SelectControl` styles](https://github.com/WordPress/gutenberg/blob/0221522f253e094b277a1485b7a2d186cb172632/packages/components/src/select-control/styles/select-control-styles.ts#L142), which in turn use a [utility from `InputControl` ](https://github.com/WordPress/gutenberg/blob/1ec58db138ca5febf4c54de83a8dde32f45b77de/packages/components/src/input-control/styles/input-control-styles.tsx) to determine the font size.

I've updated `CustomSelectControlV2` to use the same utility.

## Testing Instructions

In Storybook, open `CustomSelectControlV2` legacy adapter examples:

- Make sure that the font size is `16px` regardless of the `size` prop for screens narrower than `600px`
- Make sure that, for screens wider than `600px`, the font size is `11px` when the `size` is `small`, and `13px` otherwise
- In general, make sure that the behavior is aligned with the V1 version of the component

> [!NOTE]  
> During my testing, I realised that `CustomSelectControl` V1 has a bug for which the font size doesn't change as it should according to the size for screens wider than `600px` — for some reason, the `size` doesn't get correctly passed to the `fontSizeStyles` utility, and as a consequence the default font size is always applied (`13px`). You can verify the buggy behaviour by setting `size="small"` on `CustomSelectControl` V1 and observe how the font size is still `13px` instead of `11px`.
> I don't think we should spend time fixing it, since the V1 implementation is soon going away.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -28,6 +28,7 @@
 -   `CustomSelectControlV2`: allow wrapping item hint to new line ([#62848](https://github.com/WordPress/gutenberg/pull/62848)).
 -   `CustomSelectControlV2`: fix select popover content overflow. ([#62844](https://github.com/WordPress/gutenberg/pull/62844))
 -   `CustomSelectControlV2`: keep legacy arrow down behavior only for legacy wrapper. ([#62919](https://github.com/WordPress/gutenberg/pull/62919))
+-   `CustomSelectControlV2`: fix trigger button font size. ([#63131](https://github.com/WordPress/gutenberg/pull/63131))
 -   Extract `TimeInput` component from `TimePicker` ([#60613](https://github.com/WordPress/gutenberg/pull/60613)).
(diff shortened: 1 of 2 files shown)


tyxla (approved):
Confirmed to always be 16px on mobile layouts, just like on the v1:

Small size:
Screenshot 2024-07-04 at 16 55 36

Default size:
Screenshot 2024-07-04 at 16 55 47

(Unstable) Large size:
Screenshot 2024-07-04 at 16 55 59

Font size is 16px for all: ✅ 
Screenshot 2024-07-04 at 16 56 19

PR #63133, opened 2024-07-04
Fix accessibility of the hooked blocks toggles

### LINKED ISSUE #63094: Refactor the hooked blocks plugins UI to not use flex-direction row-reverse
### Description

See 
https://github.com/WordPress/gutenberg/pull/52969
https://github.com/WordPress/gutenberg/pull/54029

In https://github.com/WordPress/gutenberg/pull/63081 a new Stylelint custom rule is added to prevent the usage of the `flex-direction` CSS property values `row-reverse` and `column-reverse` values.

For more details about why this is a problem for accessibility and the very limited cases where these values can be used, please see https://github.com/WordPress/gutenberg/issues/63048 and https://github.com/WordPress/gutenberg/issues/61241

The UI for the hooked blocks toggles uses [row-reverse](https://github.com/WordPress/gutenberg/blob/1dcc812a6ae2a825573350f67a0a6dda8f3e0387/packages/block-editor/src/hooks/block-hooks.scss#L7). By doing so, it alters the meaning and interaction of the toggle labels and the toggles themselves. This is not OK for accessibility.

Also importantly: base component should not be altered with local implementations or CSS overrides.
The base component are designed, built, and tested with accessibility in mind. Or, at least, they aim to be as accessible as possible. Instead, it appeaers the implementation of this UI in https://github.com/WordPress/gutenberg/pull/52969 and https://github.com/WordPress/gutenberg/pull/54029 didn't take accessibility into consideration enough, neither the PRs had any accessibility label.

Local ad-hoc implementations and overrides also introduce UI inconsistencies that don't help providing an overall cohesive and consistent user experiences. In most of hte cases they just add cognitive load. Not to mention maintenances cost in the long term.

This UI needs to be refactored to not use `row-reverse`. To my understanding, it was used only to allow the addition of icons representing the plugins in the UI. I'd argue those icons add a little value. They're not necessary and they break the toggles in terms of expected design, consistency, usability, and accessibility.

From a semantics and accessibility perspective, the toggles are, under the hood, inputs of type `checkbox`. The expectation is to have the input on the left and the label on the right. Additionally, the unnecessary spacing between the label on the left and the toggle on the right doesn't help users who use screen magnifiers. 

References:
Anyone interested in understanding why, historically, labels for input fields are placed _before_ the field and labels for checkboxes and radio buttons are placed _after_, may want to have a read at https://www.w3.org/WAI/WCAG22/Techniques/general/G162.html

- To make this UI appear in the editor it's necessary to add some hooked blocks first.
- Quoting part of the instructions from https://github.com/WordPress/wordpress-develop/pull/6358:
- Add the following code to your themes functions.php:

```
function register_loginout_hooked_block( $hooked_blocks, $position, $anchor_block, $context ) {
	if ( $anchor_block === 'core/post-content' && $position === 'after' ) {
		$hooked_blocks[] = 'core/loginout';
	}

	return $hooked_blocks;
}
add_filter( 'hooked_block_types', 'register_loginout_hooked_block', 10, 4 );

function register_pagelist_hooked_block( $hooked_blocks, $position, $anchor_block, $context ) {
	if ( $anchor_block === 'core/post-content' && $position === 'before' ) {
		$hooked_blocks[] = 'core/page-list';
	}

	return $hooked_blocks;
}
add_filter( 'hooked_block_types', 'register_pagelist_hooked_block', 10, 4 );
```

- Go to the Site Editor > Templates > Single Posts template.
- Observe that:
  - The page list block is inserted before the post content.
  - The login/out block is inserted after the post content.
- Select the post content block. 
- Observe the toggles to show/hide the hooked blocks in the Inserter > Block tab > Plugins section.
- Observe the toggles are on the right and their labels on the left.
- Expected: the toggles to be on the left and their labels to immediately follow to the right.
- Inspect the sources and observe the toggle position is swapped with `flex-direction: row-reverse;` set on their container.

The highlighted toggles in the screenshot below are the ones with their order altered with `flex-direction: row-reverse;`.
Notice the design inconsistency with the default toggle above in the panel.

Screenshot 2024-07-03 at 11 45 15

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63094

## What?

The hooked block toggles aren't accessible, as the visual order and DOM order mismatch.
Additionally, base component should not be overridden with local, ad-hoc custom implementations.

## Why?

VIsual order and DOM order must match, when the order affects meaning, oerability and interaction.
Base component should not be overridden: they are designed and tested to provide as much aaccessibility as possible and consistent design.

## How?

- Changes the toggles to use a plain string for their labels instead of an icon plus a string;.
- Removes the `flex-direction: row-reverse` CSS property that was used to swap the toggle and label position.

## Testing Instructions

- Follow the instructios in the issue to add as much hooked blocks you like.
- Select the anchor block (ie. the block the hocked blocks are hooked to, in our examplee it's the post content in the Single Posts Template).
- Observe the 'Plugins' section in the INspector.
- Observe the toggles and their lables have now the standard layout with the toggle on the left and the lable immediately on the right.
- Observe the toggle do not use icons any longer.

## Screenshots or screencast 

Before and after:

before and after

### DIFF
--- a/packages/block-editor/src/hooks/block-hooks.js
+++ b/packages/block-editor/src/hooks/block-hooks.js
@@ -3,18 +3,14 @@
  */
 import { __ } from '@wordpress/i18n';
 import { Fragment, useMemo } from '@wordpress/element';
-import {
-	__experimentalHStack as HStack,
-	PanelBody,
-	ToggleControl,
-} from '@wordpress/components';
+import { PanelBody, ToggleControl } from '@wordpress/components';
 import { createBlock, store as blocksStore } from '@wordpress/blocks';
 import { useDispatch, useSelect } from '@wordpress/data';
 
 /**
  * Internal dependencies
  */
-import { BlockIcon, InspectorControls } from '../components';
+import { InspectorControls } from '../components';
 import { store as blockEditorStore } from '../store';
 
 const EMPTY_OBJECT = {};
@@ -204,14 +200,7 @@ function BlockHooksControlPure( {
 										__nextHasNoMarginBottom
 										checked={ checked }
 										key={ block.title }
-										label={
-											<HStack justify="flex-start">
-												<BlockIcon
-													icon={ block.icon }
-												/>
-												<span>{ block.title }</span>
-											</HStack>
-										}
+										label={ block.title }
 										onChange={ () => {
 											if ( ! checked ) {
 												// Create and insert block.
--- a/packages/block-editor/src/hooks/block-hooks.scss
+++ b/packages/block-editor/src/hooks/block-hooks.scss
@@ -1,14 +1,5 @@
 .block-editor-hooks__block-hooks {
 	/**
-	 * Since we're displaying the block icon alongside the block name,
-	 * we need to right-align the toggle.
-	 */
-	.components-toggle-control .components-h-stack {
-		/* stylelint-disable-next-line declaration-property-value-allowed-list -- This should be refactored to not use the row-reverse value. */
-		flex-direction: row-reverse;
-	}
-
-	/**
 	 * Un-reverse the flex direction for the toggle's label.
 	 */
 	.components-toggle-control .components-h-stack .components-h-stack {


joedolson:
I'm not sure I agree with the removal of icons here; I think that a) icons don't need to be unique if they are supplemental to visible text, and b) icons can be useful references in supporting people with low literacy in the UI language or with cognitive impairments.

The argument that the magnifying glass + the text "search" is going to imply that this will perform a search any more than the word "search" does by itself seems specious; the context should provide sufficient clarity there. 

The change of order is good; but I'm not totally bought in to the argument that the icons need to be removed to make this happen.

afercia:
> I'm not sure I agree with the removal of icons here; I think that a) icons don't need to be unique if they are supplemental to visible text, and b) icons can be useful references in supporting people with low literacy in the UI language or with cognitive impairments.

@joedolson I see a point in keeping the icons. Some UIs, for example menus, certainly benefit from using both icons and visible text labels. However, this specific UI is a little different. The toggles need to be on the left. At that point, I'm not sure there's any good spot where the icons can be placed into. The visuals of toggle + icon + visible label wouldn't great, as illustrated in the screenshot below:

Screenshot 2024-08-27 at 12 59 30

afercia:
In absence of a new design I'm inclined to merge this PR. @WordPress/gutenberg-design If you have a chance to provide a new design soon that would be welcomed, thanks. The key part is:
- Toggles on the left.
- Labels on the right.

joedolson (approved):
This works for me.

Issue #42630, opened 2022-07-22
Components: label multiple / composed inputs correctly


ciampo:
## What problem does this address?

_For more context: this topic was first pointed out in https://github.com/WordPress/gutenberg/pull/42118#issuecomment-1175921129 and https://github.com/WordPress/gutenberg/issues/42095, and recently discussed in https://github.com/WordPress/gutenberg/pull/42351#pullrequestreview-1039376506_

There are a few instances in the `@wordpress/components` package where **multiple input elements are combined into higher-level components without appropriate labelling.**

As established in previous conversations (see links above) and in [w3c's official documentation](https://www.w3.org/WAI/tutorials/forms/grouping/), the most correct way would be to **associate related inputs by wrapping them in a `fieldset` elements, while keeping individual `label`s for each single `input`**.

The challenge in doing so is to avoid nested `fieldset`s when composing components which already internally group multiple `input`s. For example:
- `BorderControl` is rendered by `BorderBoxControl` (as discussed in https://github.com/WordPress/gutenberg/pull/42351#pullrequestreview-1039376506)
- `UnitControl` is rendered by `RangeControl` and many other higher-level controls
- ... and many others

## What is your proposed solution?

We should come up with a solution that can be applied systematically across all `*Control` components.

ciampo:
One idea that I though we could explore, is to:

- make sure all components in need of grouping multiple inputs are wrapped with the same component — could be the existing `BaseControl`, or a new `ControlGroup` (name TBD) component
- add functionality to `BaseControl` (or the new component) to optionally render a `fieldset` around the control
- leveraging the context system, `BaseControl` (or the new component) could ready its context to look for a `disallowFieldset` (name TBD) prop that would disallow rendering the `fieldset`
- therefore, each `BaseControl` (or the new component) that is supposed to render a `fieldset` would set the `disallowFieldset` flag via context, which will be consumed by all nested `BaseControl` components


Also, an alternative approach, according to the official [W3C's guidelines](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria), would be to use a `div role=group` and the `aria-labelledby` attribute. Given that [we established](https://github.com/WordPress/gutenberg/pull/42351#issuecomment-1187245438) that it's not the best approach, would it still be a good fallback for when we can't render a `fieldset` ?

Thoughts?

ciampo:
[...]

PR #63137, opened 2024-07-04
CustomSelectControl V2: fix labelling with a visually hidden label

ciampo:
## What?

Part of #55023

Requires #63169 

Fix a bug in the V2 implementation where the component wouldn't be labelled property to assistive technology when the label is hidden visually via the `hideLabelFromVision` prop.

## Why?

Making sure that controls are properly labelled is essential to ensure a good experience for all users.

## How?

I refactored the code and made it so that the internal implementation of `CustomSelectControlV2` always relies on the `Ariakit.SelectLabel` component even when the `hideLabelFromVision` is `true`, which takes care of setting the correct attributes necessary for the control labeling. I was able to do so by leveraging the `render` prop ([see docs](https://ariakit.org/guide/composition#composing-with-custom-components)).

## Testing Instructions

- Make sure that the newly added unit tests pass (and that the newly added tests for the V2 version don't pass on `trunk`)
- Load Storybook examples, play with the `hideLabelFromVision` prop:
  - Make sure that the component looks and behaves like on `trunk`
  - Inspect the DOM and check that the button with `role="combobox"` is correctly labeled by the hidden `label`

### DIFF
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -113,13 +113,17 @@ function _CustomSelect(
 	return (
 		// Where should `restProps` be forwarded to?
 		<div className={ className }>
-			{ hideLabelFromVision ? ( // TODO: Replace with BaseControl
-				<VisuallyHidden as="label">{ label }</VisuallyHidden>
-			) : (
-				<Styled.SelectLabel store={ store }>
-					{ label }
-				</Styled.SelectLabel>
-			) }
+			<Styled.SelectLabel
+				store={ store }
+				render={
+					hideLabelFromVision ? (
+						// @ts-expect-error `children` are passed via the render prop
+						<VisuallyHidden as="label" />
+					) : undefined
+				}
+			>
+				{ label }
+			</Styled.SelectLabel>
 			<InputBase
 				__next40pxDefaultSize
 				size={ size }
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -1,8 +1,7 @@
 /**
  * External dependencies
  */
-// eslint-disable-next-line no-restricted-imports
-import type * as Ariakit from '@ariakit/react';
+import * as Ariakit from '@ariakit/react';
 
 /**
  * WordPress dependencies
@@ -25,6 +24,7 @@ import type {
 } from './types';
 import InputBase from '../input-control/input-base';
 import SelectControlChevronDown from '../select-control/chevron-down';
+import BaseControl from '../base-control';
 
 export const CustomSelectContext =
 	createContext< CustomSelectContextType >( undefined );
@@ -113,17 +113,20 @@ function _CustomSelect(
 	return (
 		// Where should `restProps` be forwarded to?
 		<div className={ className }>
-			<Styled.SelectLabel
+			<Ariakit.SelectLabel
 				store={ store }
 				render={
 					hideLabelFromVision ? (
 						// @ts-expect-error `children` are passed via the render prop
-						<VisuallyHidden as="label" />
-					) : undefined
+						<VisuallyHidden />
+					) : (
+						// @ts-expect-error `children` are passed via the render prop
+						<BaseControl.VisualLabel as="div" />
+					)
 				}
 			>
 				{ label }
-			</Styled.SelectLabel>
+			</Ariakit.SelectLabel>
 			<InputBase
 				__next40pxDefaultSize
 				size={ size }
--- a/packages/components/src/custom-select-control-v2/test/index.tsx
+++ b/packages/components/src/custom-select-control-v2/test/index.tsx
@@ -436,4 +436,14 @@ describe.each( [
 			} )
 		).toBeVisible();
 	} );
+
+	it( 'Should label the component correctly even when the label is not visible', () => {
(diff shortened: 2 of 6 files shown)


mirka (commented):
I feel like the component should be wrapped in a `BaseControl` (for consistency with other controls), or at the very least use `BaseControl.VisualLabel` instead of `Styled.SelectLabel`. Are there some considerations I'm missing?

ciampo:
> Are there some considerations I'm missing?

Not really, my initial fix focused on using the components that were already being used (and that made me miss the `TODO` inline comment). I'll look into using as much of `BaseControl` as possible

ciampo:
@mirka since I don't think it would be possible to use `<BaseControl />` while also using `Ariakit.SelectLabel`, I opted for your second-best strategy of keeping most code as-is and swapping `Styled.SelectLabel` with `BaseControl.VisualLabel`.

I also had to allow `BaseControl.VisualLabel` to forward ref in order for it work correctly with `Ariakit.SelectLabel` (since ariakit [needs to grab a reference to the label element](https://github.com/ariakit/ariakit/blob/main/packages/ariakit-react-core/src/select/select-label.tsx#L62)).

The only visible difference is that the label now renders a tiny gap above itself, due to the fact that `BaseControl.VisualLabel` has `display: inline-block`, which doesn't trim the whitespace. That is an inconsistency across 	`BaseControl`-based components, though:
- components using "pure" `BaseControl` show the extra white space above label
- components using only `BaseControl.VisualLabel` may or may not show the white space based on a bunch of factors. The most common one is whether the label wrapper element is a flex or grid parent (which trims whitespace automatically).

packages/components/src/base-control/stories/index.story.tsx
```diff
@@ -55,7 +55,6 @@ WithHelpText.args = {
  * otherwise use if the `label` prop was passed.
  */
 export const WithVisualLabel: StoryFn< typeof BaseControl > = ( props ) => {
-	// @ts-expect-error - Unclear how to fix, see also https://github.com/WordPress/gutenberg/pull/39468#discussion_r827150516
```
ciampo: Well, it looks like the way to fix it was to allow ref forwarding 🤷‍♂️

packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -113,13 +113,20 @@ function _CustomSelect(
...
-			) }
+			<Ariakit.SelectLabel
+				store={ store }
+				render={
+					hideLabelFromVision ? (
+						// @ts-expect-error `children` are passed via the render prop
+						<VisuallyHidden />
+					) : (
+						// @ts-expect-error `children` are passed via the render prop
+						<BaseControl.VisualLabel as="div" />
```
ciampo, line 124: Using the `label` HTML element is not correct in this case, as also confirmed by the [ariakit docs](https://github.com/ariakit/ariakit/blob/main/packages/ariakit-react-core/src/select/select-label.tsx#L19-L22):

> Since it's not a native select element, we can't use the native label element.

ciampo:
I extracted the `BaseControl.VisualLabel` changes to a separate PR: #63169

Once merged, I'll rebase and remove those same changes from this PR.

mirka:
> * components using "pure" `BaseControl` show the extra white space above label
> * components using only `BaseControl.VisualLabel` may or may not show the white space based on a bunch of factors. The most common one is whether the label wrapper element is a flex or grid parent (which trims whitespace automatically).

Furthermore, this visual difference won't manifest in a wp-admin context because `common.css` has `body { line-height: 1.4em; }`, normalizing the line-height of `.components-base-control__field`. Do you think it's worth adding an explicit line-height to `components-base-control__field`? 🤔 Could be risky.

packages/components/src/custom-select-control-v2/custom-select.tsx
mirka: Interesting! I played around and confirmed that although a native `label` will work as expected in Chrome, it behaves differently in Firefox and Safari. Clicking on the label not only focuses but clicks the "select button", expanding the dropdown. So yeah, I guess it does need special handling 🫤

ciampo:
> Do you think it's worth adding an explicit line-height to components-base-control__field? 🤔 Could be risky.

I agree that it would be risky — let's leave as-is. In the long term, I'd like to re-think `BaseControl` to enable more flexibility for which `<Label />` component to render to better composition with other components, and to handle out of the box [composed inputs](https://github.com/WordPress/gutenberg/issues/42630).

I merged https://github.com/WordPress/gutenberg/pull/63169 and rebased this PR. It should be ready for a last round of review

packages/components/src/custom-select-control-v2/custom-select.tsx
ciampo: Yup, that's part of what `Ariakit.SelectLabel` does under the hood

packages/components/src/custom-select-control-v2/test/index.tsx
```diff
@@ -436,4 +436,14 @@ describe.each( [
 			} )
 		).toBeVisible();
 	} );
+
+	it( 'Should label the component correctly even when the label is not visible', () => {
```
tyxla, line 440: Thanks for adding a test 🙌

tyxla (commented):
Went through a round of testing on this and it tested well according to the test instructions. 

I have nothing to add on the code, IMHO it's good to go, but I'll let @mirka approve it since she had some feedback before.

mirka (approved):
Perfect, thank you!

PR #63139, opened 2024-07-04
Block position controls: use V2 legacy adapter instead of V1 `CustomSelectControl`

ciampo:
## What?

Part of #55023
Requires #63131, #63137  and #63230 to be merged first

Replace the V1 `CustomSelectControl` with the V2 legacy adapter in the block position control

## Why?

The goal is to ultimately replace the legacy V1 implementation entirely with the V2 legacy adapter. We're performing the migration gradually, and fixing any bugs / gaps as we go.

## How?

 - Import the V2 legacy adapter instead of the V1 implementation
 - Remove unnecessary style overrides and class names

## Testing Instructions

1. Add a block that supports the position hook (ie. the group block)
2. Select the newly added block, and in the inspector sidebar expand the "Position" panel
3. Interact with the dropdown that allows to pick different positions. Make sure that everything looks and behaves like on `trunk` 

> [!NOTE]
> There are some small style differences on the items with hint (see screenshots below):
> - The checkmark on trunk was vertically top aligned, while in this PR it is vertically centered with the whole item (including the hint)
> - There is more vertical (block) spacing before and after the hint text, which comes from a few styles tweaks done in the V2 to allow for hint wrapping directly in the component (instead that via style overrides). The block margin could be removed, but there's a risk that the white space under the hint would be too short compared to the space above the main item text. And since we don't have an easy way to know whether the hint wraps on a new line or not, we can't just add more bottom margin, as it would look bad when the hint renders on the same line as the item contents

|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/f83e035a-53ba-4503-bb8b-4f83d2464250" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/744340af-7871-4390-82e4-e6b892f23160" /> |
| Screenshot 2024-07-05 at 11 32 04 | Screenshot 2024-07-05 at 11 31 37 |


packages/block-editor/src/hooks/position.scss
```diff
@@ -1,18 +0,0 @@
...
-		grid-template-columns: auto 30px;
-		line-height: $default-line-height;
-		margin-bottom: 0;
-	}
-
-	.components-custom-select-control__item-hint {
-		grid-row: 2;
-		text-align: left;
-	}
-}
```
ciampo, line 18: These overrides shouldn't be necessary anymore because the V2 legacy adapter already allows the item hint to render on a separate line when there isn't enough space and sets the correct line height

packages/block-editor/src/hooks/position.scss
```diff
@@ -1,18 +0,0 @@
-.block-editor-hooks__position-selection__select-control {
-	.components-custom-select-control__hint {
-		display: none;
-	}
-}
```
ciampo, line 5: These overrides shouldn't  be necessary if we don't set the `__experimentalShowSelectedHint` prop on `CustomSelectControl`

packages/block-editor/src/hooks/position.js
ciampo, line 1: Removing the `.block-editor-hooks__position-selection*` class names, since style overrides don't seem to be necessary anymore (and I couldn't find any other references to these class names in the codebase)

ciampo:
Hey @andrewserong , tagging you in this PR since you seem to have worked on this part of the editor last. You may be able to review and validate some of my assumptions, especially around removing style overrides (https://github.com/WordPress/gutenberg/pull/63139#discussion_r1665930464, https://github.com/WordPress/gutenberg/pull/63139#discussion_r1665929884)

ciampo:
Probably better to get some feedback from @WordPress/gutenberg-design regarding the small style differences for items with a hint (see the "Note" in the `Testing instructions` section of the PR description).

For context, we're trying to remove custom style overrides as part of a refactor of `CustomSelectControl`. The new version of the component implements a similar "hint wrapping" behavior without the need for style overrides, but there are a couple of small differences. We'd need to understand if we're ok with those differences, or if we feel like we'd like to make tweaks to the component itself.

jameskoster:
I appreciate all of this will be separate, but I wonder if we should consider aligning some of the details here with DropdownMenuV2:

* Place selection indicator on the left
* Use the dot icon for single selection
* Use the checkmark for multi-selection
* Spacing
* Hover/disabled styles

---

I go back and forth on whether the icon should be aligned to the main label or not, keen to hear input from other @WordPress/gutenberg-design folks on that.

The larger gap isn't a deal-breaker for me. A 24px line-height (rather than 28px) – or a 24px min-height – would be more universally aligned though. We use that technique in a few places to ensure text aligns with icons/small buttons.

jasmussen:
It would be a good task to separate out for any design system work, once that can happen, to simply design and specify metrics for every single component, and how they latch together. Inner paddings, font sizes, gap between, etc. The closest we have at the moment, is [this](https://www.figma.com/design/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?node-id=53440-15441&t=frep2bjDUIyv3Mpa-1), where we can compare metrics, and as far as I can tell, this looks right enough. 

> I go back and forth on whether the icon should be aligned to the main label or not, keen to hear input from other @WordPress/gutenberg-design folks on that.

Yeah, that's a tricky one. My instinct would be to align with the label, and at the same time I just gave feedback on data views that went in the other direction. 

At the moment, we have this:

Screenshot 2024-07-05 at 13 47 52

That's not great, tbh. 

I wonder if it's possible to align the icon to _the first two lines of text_? I.e. it will be centered vertically if there's 1 or 2 lines of text, but it will have the same placement from 2 and upwards. Not sure if that'd be better or worse, but seems like it should be possible with some flex magic?

ciampo:
Thank you both for chiming in! Here is some additional context.

We are currently in the process of swapping the current `CustomSelectControl` implementation with a new one that aims at mimicking as much as possible the old implementation — and this is what we're doing in this PR. **So, specifically for this PR, the feedback I'm looking for is whether these small differences are acceptable.**

Separately, we're going to be looking at creating a separate `V2` version of `CustomSelectControl` (like we're doing for `DropdownMenuV2`) — and I think that's going to be the perfect moment for us to re-discuss the whole design of the component and align it with the rest of the design system. Eventually the V1 will be deprecated and won't be used anymore in the editor.

> Use the dot icon for single selection
> Use the checkmark for multi-selection

I know that I just said that I don't think that discussing these design aspects is in the scope of this PR, but I'd like share my mental model for types of selection

- mutually exclusive selection (radio semantics)
- non-mutually exclusive selection, either single or multiple (checkbox semantics)

In my mental model, the dot is associated to radio semantics, and the checkmark to checkbox semantics

jameskoster:
I don't love the centrally aligned icon, but I can live with it until a broader redesign of the component.

ciampo:
> I don't love the centrally aligned icon, but I can live with it until a broader redesign of the component.

That should be easy enough to tweak quickly — I'll give it a try next week.

ciampo:
@jameskoster checkmark alignment changes extracted in https://github.com/WordPress/gutenberg/pull/63230 — once merged, I will rebase this PR to include them

ciampo:
> checkmark alignment changes extracted in #63230 — once merged, I will rebase this PR to include them

Rebased to include those changes. Here's what the block position control looks like when selecting an item with a long hint:

Screenshot 2024-07-08 at 13 34 30

Let me know if there's any additional feedback — otherwise, we're probably good for a final review round.

mirka (approved):
Looks good 🚀

packages/block-editor/src/hooks/position.scss
```diff
@@ -1,18 +0,0 @@
-.block-editor-hooks__position-selection__select-control {
```
tyxla, line 1: So cool to see all the custom styling cleaned up 👏

packages/block-editor/src/hooks/position.js
```diff
@@ -283,13 +279,11 @@ export function PositionPanelPure( {
 			options.length > 1 ? (
 				<InspectorControls group="position">
 					<BaseControl
-						className="block-editor-hooks__position-selection"
```
tyxla, line 286: Any chance someone was relying on these custom classes?

tyxla (approved):
Tests well and looks good! 🚀

packages/block-editor/src/hooks/position.js
ciampo: I couldn't find direct usages in Gutenberg, and I also did a quick scan on WP Directory without finding any matches.

andrewserong (commented):
I'm coming in very late here, but thanks for the ping and for tidying this one up! It's looking good to me on `trunk`.

> So cool to see all the custom styling cleaned up 👏

+1 so cool to see 😀

PR #63140, opened 2024-07-04
ToolbarButton: Fix docs for `accessibleWhenDisabled`

mirka:
## What?

Fix the Storybook props table for `ToolbarButton` so it shows the correct default for the `accessibleWhenDisabled` prop.

## Why?

The `__experimentalIsFocusable` prop was stabilized and renamed to `accessibleWhenDisabled` (#62282) in the `Button` component where this component inherits its props from. The default for the prop in Button is `false`, whereas here it is functionally equivalent to `true`.

## Testing Instructions

The Storybook props table for the `ToolbarButton` component (subcomponent of `Toolbar`) should show the correct default for `accessibleWhenDisabled`.

## Screenshots or screencast 

Storybook props table showing the correct default


DaniGuardiola (approved):
good stuff

tyxla (approved):
👍

PR #63141, opened 2024-07-04
Block Hooks: Optimize selectors in the control component

Mamaduka:
## What?
A micro-optimization for the `BlockHooksControlPure` component removes extra block editor store subscription.

## Why?
Each micro-optimization might improve the performance.

## How?
Use static selector getter and select a value in the `insertBlockIntoDesignatedLocation` callback.

## Testing Instructions
1. Register a hooked block type. See the snippet below.
2. Go to the Site Editor > Templates > Single Posts template.
3. Observer that the Page List block is inserted after the post content block.
4. You can toggle settings in the controls.

### Snippet
```php
add_filter( 'hooked_block_types', function ( $hooked_blocks, $position, $anchor_block ) {
	if ( $anchor_block === 'core/post-content' && $position === 'after' ) {
		$hooked_blocks[] = 'core/page-list';
	}

	return $hooked_blocks;
}, 10, 3 );
```

Same.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/240569/757b2df1-eea2-43f0-ab35-74e3413eb355

### DIFF
--- a/packages/block-editor/src/hooks/block-hooks.js
+++ b/packages/block-editor/src/hooks/block-hooks.js
@@ -43,25 +43,12 @@ function BlockHooksControlPure( {
 		[ blockTypes, name, ignoredHookedBlocks ]
 	);
 
-	const { blockIndex, rootClientId, innerBlocksLength } = useSelect(
-		( select ) => {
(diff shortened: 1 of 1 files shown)


ellatrix (approved):
Nice

Issue #63128, opened 2024-07-04
Tracking issue: Further advancing data views


jameskoster:
### View options
- [x] #63195
- [x] https://github.com/WordPress/gutenberg/issues/57669
- [x] https://github.com/WordPress/gutenberg/issues/63414
- [x] https://github.com/WordPress/gutenberg/issues/63872
- [x] https://github.com/WordPress/gutenberg/issues/65600

### Bulk actions / editing
- [x] #63193
- [ ] Quick editing https://github.com/WordPress/gutenberg/issues/55101
- [x] https://github.com/WordPress/gutenberg/issues/63150

### Fields
- [x] #63199
- [ ] https://github.com/WordPress/gutenberg/issues/58012
- [x] Pages: Include avatar in Author field (@jorgefilipecosta)
- [ ] https://github.com/WordPress/gutenberg/issues/57596

### Filtering
- [x] https://github.com/WordPress/gutenberg/issues/60696 (@jorgefilipecosta)
- [x] https://github.com/WordPress/gutenberg/issues/60468
- [x] https://github.com/WordPress/gutenberg/issues/60472
- [x] https://github.com/WordPress/gutenberg/issues/62057
- [ ] https://github.com/WordPress/gutenberg/issues/59696
- [x] https://github.com/WordPress/gutenberg/issues/63152

### General polish
- [x] List / Table layout – selected item stroke should be tinted blue. (@jorgefilipecosta)
- [ ] https://github.com/WordPress/gutenberg/issues/56199
- [ ] https://github.com/WordPress/gutenberg/pull/62028
- [x] https://github.com/WordPress/gutenberg/issues/61245
- [x] Pages: Trash view should default to table layout (@jorgefilipecosta)
- [x] Posts: Trash view should default to table layout
- [x] https://github.com/WordPress/gutenberg/issues/63675
- [x] Use 'raquo' icons for pagination. ([PR](https://github.com/WordPress/gutenberg/pull/63594))
- [ ] Sticky actions column in table layout https://github.com/WordPress/gutenberg/pull/65086
- [ ] https://github.com/WordPress/gutenberg/issues/62748
- [x] https://github.com/WordPress/gutenberg/pull/63299
- [x] #63302
- [ ] https://github.com/WordPress/gutenberg/issues/60847
- [ ] https://github.com/WordPress/gutenberg/issues/59832
- [x] https://github.com/WordPress/gutenberg/issues/63153
[...]

PR #63142, opened 2024-07-04
Updated: Pages: Include avatar in Author field.

jorgefilipecosta:
This PR includes the author avatar on the author field.

Part of: https://github.com/WordPress/gutenberg/issues/63128

cc: @jameskoster 

## Screenshots
Screenshot 2024-07-04 at 18 30 58

## Testing Instructions
Assign an email with a gravatar image to one of the other of a page.
Load the pages list and verify the avatar is correctly shown.


jameskoster:
Nice, this is working well on Grid and Table layouts. Can we also include it in the List layout? 🤞

ellatrix:
What if I want to turn off the avatars? I've wondered a few times how we could have entries that are composed of multiple fields/values. Another example of this is date and time (I may want to remove the time). Maybe we need a sub-field concept? Or a format concept? For example I could configure a date format, and for authors you could add options for `avatar`, `username`, `first-last`, etc. etc. I also saw @youknowriad [mention](https://github.com/WordPress/gutenberg/pull/63127) a `renderAs` property that is a bit similar.

jameskoster:
That's a valid feature, but potentially something that will get covered in https://github.com/WordPress/gutenberg/issues/58012. I don't know if it needs to hold up this PR?

packages/edit-site/src/components/page-templates/author-field.js
```diff
@@ -0,0 +1,82 @@
...
+		<HStack alignment="left" spacing={ 1 }>
+			{ withIcon && imageUrl && (
+				<div
+					className={ clsx( 'page-templates-author-field__avatar', {
+						'is-loaded': isImageLoaded,
+					} ) }
+				>
+					<img
+						onLoad={ () => setIsImageLoaded( true ) }
+						alt=""
```
carolinan: Why is the alt text empty?

packages/edit-site/src/components/posts-app/posts-list.js
```diff
@@ -395,6 +396,11 @@ export default function PostsList( { postType } ) {
 						value: id,
 						label: name,
 					} ) ) || [],
+				render: ( { item } ) => {
+					return (
```
ntsekouras, line 451: I don't think we should reuse this render field function. Posts list is meant to be reused by every post type eventually, have an API to get the fields by `postType`, but I see templates a bit different than the 'normal' post types. Maybe I'm wrong, but it wouldn't hurt to duplicate just some code for now.

ntsekouras (commented):
I think this use case will be handled when we implement the combination of fields, but it's fine to move forwards with this one for now.

packages/edit-site/src/components/page-templates/author-field.js
jorgefilipecosta: Hi @carolinan good question, it was already empty before in this PR I'm just moving code from the previous AuthorField component https://github.com/WordPress/gutenberg/pull/63142/files#diff-4bc1dbbf462da7ed7ab2b76d40ee01a0a2906482b12bea67383ecbb5697176acL119. But I think it should have an alt text at least something simple as "Author Text" is better than nothing, I made that update.

packages/edit-site/src/components/posts-app/posts-list.js
jorgefilipecosta: Hi @ntsekouras, I applied your suggestion and did not reuse the field from the templates.

jorgefilipecosta:
> I think this use case will be handled when we implement the combination of fields, but it's fine to move forwards with this one for now.

Yes, when we have field combination I can do a small refactor to use that, but I guess for now we can merge this one.

packages/edit-site/src/components/posts-app/posts-list.js
```diff
@@ -207,6 +217,46 @@ function getItemId( item ) {
...
+			const user = getUser( item.author );
+			return {
+				icon: authorIcon,
+				imageUrl: user?.avatar_urls?.[ 48 ],
+				text: user?.name,
+			};
+		},
+		[ item ]
+	);
+	const withIcon = viewType !== LAYOUT_LIST;
```
ntsekouras: Nit: let's rename to `withAvatar` for more clarity. Additionally and also nit is to combine the checks like ` {withAvatar && ( imageUrl ? ... : ) }

ntsekouras (approved):
Left a couple of nits, but this looks good. Thank you!

packages/edit-site/src/components/posts-app/posts-list.js
```diff
@@ -1,7 +1,16 @@
+/**
```
youknowriad, line 1: Unrelated with this PR but this file should be moved to its own components/post-list folder or something, it's not specific to "posts app"

jameskoster:
@jorgefilipecosta it looks like the List layout was missed here. I only see the avatar in the Table or Grid layout.

jameskoster:
@jorgefilipecosta Also, I appreciate this is separate, but in the Templates table the author avatar appears in the Table layout, but not the List layout. It would be good to align across.

jorgefilipecosta:
Hi @jameskoster with the iterations we did there are avatars, on list, grid, and table on templates and pages. So everything you pointed out should be fixed. Let me know if there is anything related remaining.

PR #63144, opened 2024-07-04
Zoom Out: Replace deprecated selector

Mamaduka:
## What?
PR replaces the deprecated `editor.getSelectedBlock` selector in the `ZoomOutModeInserters` component.

## Why?
The core shouldn't use deprecated selectors.

## How?
The component just needs a boolean if there's a block selection; we can use `getSelectionStart().clientId` for it.

## Testing Instructions
1. Enabled "zoomed out view when selecting a pattern category in the main inserter"  from Gutenberg > Experiments.
2. Go to Site Editor > any template.
3. Open the Inserter
4. Select a Pattern category.
5. Zoom out mode shouldn't enabled.
6. Select a section.
7. In-between inserters should be visible as before.

Same.

## Screenshots or screencast 
CleanShot 2024-07-04 at 22 22 20

CleanShot 2024-07-04 at 22 45 06

### DIFF
--- a/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
+++ b/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
@@ -21,9 +21,10 @@ function ZoomOutModeInserters() {
 		sectionRootClientId,
 		insertionPoint,
 		setInserterIsOpened,
-		selectedSection,
+		hasSelection,
 	} = useSelect( ( select ) => {
-		const { getSettings, getBlockOrder } = select( blockEditorStore );
+		const { getSettings, getBlockOrder, getSelectionStart } =
+			select( blockEditorStore );
 		const { sectionRootClientId: root } = unlock( getSettings() );
 		// To do: move ZoomOutModeInserters to core/editor.
 		// Or we perhaps we should move the insertion point state to the
@@ -33,7 +34,7 @@ function ZoomOutModeInserters() {
 		// eslint-disable-next-line @wordpress/data-no-store-string-literals
 		const editor = select( 'core/editor' );
 		return {
-			selectedSection: editor.getSelectedBlock(),
+			hasSelection: !! getSelectionStart().clientId,
 			blockOrder: getBlockOrder( root ),
 			insertionPoint: unlock( editor ).getInsertionPoint(),
 			sectionRootClientId: root,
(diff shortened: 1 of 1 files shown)


packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
```diff
@@ -33,7 +34,7 @@ function ZoomOutModeInserters() {
 		// eslint-disable-next-line @wordpress/data-no-store-string-literals
 		const editor = select( 'core/editor' );
 		return {
-			selectedSection: editor.getSelectedBlock(),
+			hasSelection: !! getSelectionStart().clientId,
```
ellatrix, line 37: Why not switch to the correct store? Maybe it could use `getSelectedBlockClientId` instead? `getSelectionStart` will also work for multi selection, not sure if that's relevant here.

ellatrix: Actually, it looks like it is relevant

ellatrix (approved):

packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
Mamaduka: I've seen `hasSelection: !! getSelectionStart().clientId` pattern in a few places, so I also decided to use it here.

I don't have a strong preference for this and am happy to update.

Mamaduka: > Actually, it looks like it is relevant

What do you mean?

ellatrix: I mean: the zoom out inserters should probably not be displayed if there's multi-selection either. Probably a bug rn.

Mamaduka: Should I switch the `getSelectedBlockClientId` or `getSelectedBlock` selector?

ellatrix: I mean, that would be the minimal change, without changing functionality. But I think in this case it was also meant to catch multi-selection, so I would like it as you did.

PR #63145, opened 2024-07-04
TimeInput: Expose as subcomponent of TimePicker

### LINKED ISSUE #52734: TimePicker components should have option to disable date field

mirka:
Follow-up to #60613
Closes #52734

## What?

Expose `TimeInput` as a subcomponent of `TimePicker`. It can now be used (publicly) as `TimePicker.TimeInput`.

## Why?

So we don't clutter the global namespace when exposing modular parts of a component.

## Testing Instructions

- The Storybook for `TimePicker` should be clear about the use of `TimePicker.TimeInput`.
- `TimePicker.TimeInput` should show a JSDoc in IntelliSense:

    IntelliSense for TimePicker.TimeInput

## Screenshots or screencast 

TimePicker.TimeInput in Storybook

---

cc @bogiii


packages/components/src/date-time/stories/time.story.tsx
```diff
@@ -49,3 +51,16 @@ const Template: StoryFn< typeof TimePicker > = ( {
 };
 
 export const Default: StoryFn< typeof TimePicker > = Template.bind( {} );
+
+const TimeInputTemplate: StoryFn< typeof TimePicker.TimeInput > = ( args ) => {
+	return <TimePicker.TimeInput { ...args } />;
```
tyxla, line 56: Felt a bit odd to me that the `currentTime` doesn't seem to work for the `TimeInput` story, while it does for the default `TimePicker` one:


https://github.com/WordPress/gutenberg/assets/8436925/29a73d96-958a-4c28-a8b2-35b66118b0f9

packages/components/src/date-time/stories/time.story.tsx
```diff
@@ -16,6 +16,8 @@ import TimePicker from '../time';
 const meta: Meta< typeof TimePicker > = {
 	title: 'Components/TimePicker',
 	component: TimePicker,
+	// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
```
tyxla, line 19: 👍

packages/components/src/date-time/stories/time.story.tsx
tyxla, line 56: Another thing I spotted in the latest Chrome - there is a subtle visual difference between the colon of the `TimePicker` and in the solo `TimeInput`:

`TimePicker`:
Screenshot 2024-07-05 at 13 35 33

`TimeInput`:
Screenshot 2024-07-05 at 13 35 39

You'll spot that for the `TimeInput` in isolation, the colon appears a bit bigger.

packages/components/src/date-time/time/index.tsx
```diff
@@ -237,4 +237,29 @@ export function TimePicker( {
...
+ * 			value={ time }
+ * 			onChange={ setTime }
+ * 			label="Time"
+ * 		/>
+ * 	);
+ * };
+ * ```
+ */
+TimePicker.TimeInput = TimeInput;
+Object.assign( TimePicker.TimeInput, { displayName: 'TimePicker.TimeInput' } );
```
tyxla, line 278: We're now exposing `TimePicker.TimeInput` as a public subcomponent, right? In that case, shouldn't we be documenting it in the README? This reminds me we also don't document `TimePicker` and `DatePicker` at this time - do you know if this is intentional?

tyxla (commented):
Looking good overall. 

Glad `subcomponents` support [won't be dropped anytime soon](https://github.com/storybookjs/storybook/issues/20782#issuecomment-1892178858).

I've left a few questions, nothing critical.

packages/components/src/date-time/stories/time.story.tsx
mirka: You're right, I should've hidden the controls for that story (1bb3f2f).

mirka: Good catch 😳 Fixed in d94947d

packages/components/src/date-time/time/index.tsx
mirka: Correct. I'll be adding a readme in the next PR. I guess there is a bit of inconsistency for subcomponent readmes, but I think TimeInput warrants a separate readme because the props are distinct.

> do you know if this is intentional?

It doesn't seem like an oversight, because the DateTimePicker readme does mention that DatePicker and TimePicker can be used individually.

tyxla (approved):
LGTM and tests well 👍 

One thing I've spotted: the order of components in Storybook is still a bit off:

Screenshot 2024-07-08 at 14 09 35

Could be fixed in a follow-up.

ciampo:
I was going to create an issue where we discuss which naming convention we want to adopt across the components package for "component families" / "compound components" / subcomponents, for better consistency (especially as we plan on having more compound components gong forward).

Do you think it would be beneficial to agree on that naming convention before exposing a new component which may potentially require an export name change soon?

mirka:
> the order of components in Storybook is still a bit off

@tyxla Is your expectation that these be alphabetical and not grouped together?

mirka:
> Do you think it would be beneficial to agree on that naming convention before exposing a new component which may potentially require an export name change soon?

@ciampo I didn't have any alternative namings for this in mind, but I'm curious what your proposals are. I can wait for your issue to be posted 👍

tyxla:
> > the order of components in Storybook is still a bit off
> 
> @tyxla Is your expectation that these be alphabetical and not grouped together?

Can be either alphabetical or grouped together, but right now they're none of that, and that was unexpected.

ciampo:
> > Do you think it would be beneficial to agree on that naming convention before exposing a new component which may potentially require an export name change soon?
> 
> @ciampo I didn't have any alternative namings for this in mind, but I'm curious what your proposals are. I can wait for your issue to be posted 👍

There it is https://github.com/WordPress/gutenberg/issues/63242

mirka:
I'm going to merge now since we've agreed on export conventions (#63242).

PR #63147, opened 2024-07-04
Query and Search blocks: support for Instant Search

### LINKED ISSUE #63053: Query and Search blocks: support for Instant Search

Now that we have the Interactivity API in Core, it's time to start experimenting with enhanced user experiences. One of these would be the ability to refresh the content of a Query block without refreshing the page while the user is typing a search.

I believe the simplest way to start with the Instant Search would be to convert the search into an instant search when the Search block is inside a Query block that has the `Forge Page Reload` option disabled.

When the `Force Page Reload` option is disabled, the `enhancedPagination` attribute is `true` and is passed via block context. Therefore, the Search block can subscribe to the `enhancedPagination` context. If it is `true`, it can add the necessary directives using the `HTML_Tag_Processor` and enqueue de `view` file.

r-chrzan:
## What?
Initial implementation for Instant Search using the Search and Query Loop blocks. Added as a new experiment on the Gutenberg Experiments page.

Related to [#63053](https://github.com/WordPress/gutenberg/issues/63053)

https://github.com/user-attachments/assets/7ee8ddc7-472c-4575-a27e-af1972278c6c

### How does it work?

1. Enable the `Instant Search and Query Block` Gutenberg experiment.
2. Insert a Query block anywhere on your site. 
3. Make sure to disable `Force Page Reload` i.e. use "enhanced pagination" in that Query block.
4. Insert a Search block anywhere inside of that Query block.

When the Search block is inside of the Query block using "enhanced pagination", it automatically gets "updated" to an Instant Search block. 

Any search input in the Instant Search block gets passed as the `?instant-search-<queryId>=<search-term>` query search param in the URL.

### Multiple Query + Search blocks

It's possible to have multiple Instant Search blocks in a page/post/template. In such a case, ensuring their functionality is isolated is essential. For this to work, the `queryId` is part of the param syntax:
- `instant-search-<queryId>=<search-term>`

### Search button

The search block can contain a search button (in some configurations). However, when using the Instant Search functionality, the button is redundant because the list of posts updates as you type. For this reason, **in the editor**, when the Search block is placed inside the Query Loop block with enhanced pagination ON, the Block Controls related to the Search button are removed. The displayed name of the block (via `label`) is changed to `Instant Search (Search)`.

**On the frontend**, the Search button is also always removed for each (Instant) Search block.

https://github.com/user-attachments/assets/f75ba0a5-2672-4d36-839b-253f25fa4de8

### Pagination

The instant search functionality **respects** the pagination of the Query block, which uses the `query-<queryId>-page` syntax.

1. Every time the search is updated, the pagination of its respective query is reset back to page 1. 
2. The pagination numbers and next/previous, are updated to reflect the number of results in the search.
3. Clearing the search also resets the pagination back to page 1.

### Limitations

- Only the `Custom` query types are supported in the current experiment. The `Default` query types are currently not supported but will be in the future version. This is prototyped already in https://github.com/WordPress/gutenberg/pull/67289.

## Alternatives

- https://github.com/WordPress/gutenberg/pull/67013 - using the `render_block_context` filter.
- https://github.com/WordPress/gutenberg/pull/67181 - using the `query_loop_block_query_vars` filter.

## Further work

This is an initial prototype and the following features is **intentionally** not yet implemented:
- Support for the `Default` queries, including multiple `Default` queries on a single page/post/template. Those have been prototyped already in https://github.com/WordPress/gutenberg/pull/67289.
- Explore an alternative mechanism to convert the Search block into a Instant Search. Currently, this is done by nesting the Search inside of Query Loop with enhanced pagination ON, but a different, more explicit mechanism might be appropriate.


luisherranz:
Great work!! 👏👏

Let's see what's next 🙂

1. Modifying the query in inherited queries.

   I don't know if you've noticed, but the `query_loop_block_query_vars` filter only works when the "Inherit query from template" option is disabled.

   Screenshot 2024-07-08 at 16 33 20

   We should check if the query is inherited, and if so, apply the modification to the general query so it works in both cases.

2. The search parameter in inherited queries.

   It's interesting that we can't use the search parameter `?s=` because, in that case, the page that loads is the Search template.

   I'm going to think about it a bit, but it is true that it might not be possible to simply use `?s=`. If we can't use it, then we need to think carefully about what the most suitable parameter is. Maybe `?is=` for instant search?

3. The search parameter in non-inherited queries.

   When the "Inherit query from template" option is disabled, the query pagination uses a custom search parameter instead of the regular `?paged=2`/`/page/2` of the inherited queries: `?query-X-page=P` where `X` is the query id (`$block->context['queryId']`).

   In that case, we should use a similar parameter: `?query-X-search=test`.

4. Pagination when searching.

   You commented [in Slack](https://wordpress.slack.com/archives/C071CRKGKUP/p1720213560128309?thread_ts=1719917306.005549&cid=C071CRKGKUP):

   > sometimes, when we are on e.g. page 3 and typing something, i see prev link button without results

   That's what happens when the new results don't have as many pages as the page you were on just before you started the search.

   I think we should try resetting the page number every time a new search is performed. In other words, if the user starts a search on page 3, we should remove the search param from the URL before navigating.

   - User is in `/page/3` or `?query-X-page=3`.
   - User types new characters.
   - We add the search parameter, but remove the pagination: `/?is=test`.

   This is probably going to present problems, as it might not be that easy to remove the pagination from the URL in the client for certain permalinks, but I think it's worth a try and we'll see what happens. If that's the case, one option to investigate would be to generate the URL of the first page on the server and send it to the client via the `state`.

5. Search button closing itself during navigation.

   You commented [in Slack](https://wordpress.slack.com/archives/C071CRKGKUP/p1720213560128309?thread_ts=1719917306.005549&cid=C071CRKGKUP):

   > it doesn’t work very well when search is just a button. sometimes toggle closes while typing

   That's because the context has a property called `isSearchInputVisible` which is set to false on the server. So when a navigation occurs, it gets reset to false.

   The way to solve this is to transform that property into an initial property and have a client-only property so that the server property is only checked on the first page load and then only the client-only property is considered.

   I made a [commit to implement this pattern](https://github.com/WordPress/gutenberg/pull/63147/commits/0159a4a7273e0b4c185e856ad461838583a398ac) by adding this type of getter:

   ```js
   get isSearchInputVisible() {
     const ctx = getContext();
     if ( typeof ctx.isSearchInputVisible === 'undefined' ) {
       return ctx.isSearchInputInitiallyVisible;
     }
     return ctx.isSearchInputVisible;
   },
   ```

gziolo:
@r-chrzan, I see that @michalczaplinski opened a PR https://github.com/WordPress/gutenberg/pull/65950 where he continued with addressing the feedback shared. If you plan to resume work, you should discuss how to proceed with collaboration moving on.

r-chrzan:
> @r-chrzan, I see that @michalczaplinski opened a PR #65950 where he continued with addressing the feedback shared. If you plan to resume work, you should discuss how to proceed with collaboration moving on.

Thanks @gziolo I spoke with @michalczaplinski last week. We will continue on my fork.

packages/block-library/src/search/view.js
```diff
@@ -66,6 +81,41 @@ const { actions } = store(
 					actions.closeSearchInput();
 				}
 			},
+			*updateSearch() {
+				const { ref } = getElement();
+				const { value } = ref;
+
+				// Don't navigate if the search didn't really change.
+				if ( value === state.search ) {
```
michalczaplinski: @r-chrzan I've removed the checks for `name === 's'` here because I didn't quite see their purpose. Were they required for anything?

packages/block-library/src/search/index.php
```diff
@@ -90,6 +94,11 @@ function render_block_core_search( $attributes ) {
 			$input->set_attribute( 'aria-hidden', 'true' );
 			$input->set_attribute( 'tabindex', '-1' );
 		}
+		// Instant search is only available when using the enhanced pagination.
+		if ( $enhanced_pagination ) {
+			$input->set_attribute( 'data-wp-bind--value', 'state.search' );
+			$input->set_attribute( 'data-wp-on--input', 'actions.updateSearch' );
```
sirreal: How about on-async here?
```suggestion
			$input->set_attribute( 'data-wp-on-async--input', 'actions.updateSearch' );
```

packages/block-library/src/search/index.php
```diff
@@ -165,20 +174,32 @@ function render_block_core_search( $attributes ) {
...
+				'isSearchInputInitiallyVisible' => $open_by_default,
+				'inputId'                       => $input_id,
+				'ariaLabelExpanded'             => $aria_label_expanded,
+				'ariaLabelCollapsed'            => $aria_label_collapsed,
 			)
 		);
-		$form_directives      = '
-		 data-wp-interactive="core/search"'
-		. $form_context .
+		$form_directives     .= $form_context .
```
sirreal: Out of scope:

An observation that the mix of managing HTML via HTML API, string concatenation, and interpolation here is interesting. It would be nice to stick to one approach in the future, although I know the HTML API doesn't support everything yet.

packages/block-library/src/search/view.js
```diff
@@ -3,7 +3,11 @@
  */
 import { store, getContext, getElement } from '@wordpress/interactivity';
 
-const { actions } = store(
+const isEmpty = ( obj ) =>
+	[ Object, Array ].includes( ( obj || {} ).constructor ) &&
+	! Object.entries( obj || {} ).length;
```
sirreal: I'm surprised about adding `isEmpty` here… is it necessary? It's used to check the input value which should be a string, isn't it?

packages/block-library/src/search/view.js
```diff
@@ -66,6 +81,41 @@ const { actions } = store(
...
+
+				// Debounce the search by 300ms to prevent multiple navigations.
+				// We can do this by yielding a promise that resolves after 300ms and
+				// then bailing out if the search has changed.
+				yield new Promise( ( resolve ) => setTimeout( resolve, 300 ) );
+				if ( value !== state.search ) {
+					return;
+				}
+
+				const url = new URL( window.location );
```
sirreal: Minor: the URL constructor takes a string or URL object. This will work by coercing `window.location` to a string, but passing `window.location.href` correctly passes the string:

```suggestion
				const url = new URL( window.location.href );
```

packages/block-library/src/search/view.js
```diff
@@ -66,6 +81,41 @@ const { actions } = store(
...
+				if ( value === state.search ) {
+					return;
+				}
+
+				state.search = value;
+
+				// Debounce the search by 300ms to prevent multiple navigations.
+				// We can do this by yielding a promise that resolves after 300ms and
+				// then bailing out if the search has changed.
+				yield new Promise( ( resolve ) => setTimeout( resolve, 300 ) );
```
sirreal: I have concerns about the way this is gated. I think we want only the last case to proceed, with previous ones being cancelled.

I _think_ we only want one of these to occur on the page with the most recent one taking precedence.

One proposal:
- Wrap this yield in try/catch/finally.
- The catch branch just returns (it means this invocation was superseded)
- Store a rejection callback in module state (as a variable in the module scope)
- The finally removes the rejection callback from module state (sets it to `null`)
- Upon entering this function, the rejection callback is called if set. This cancels previous executions.


---

There's a similar example (although not identical - it's using abort controllers) [here in the HTML API Debugger plugin](https://github.com/sirreal/html-api-debugger/blob/a68ce9e1fa0a3148c9ef2cc8192948478bf949d7/html-api-debugger/view.js#L299-L327).

packages/block-library/src/search/view.js
```diff
@@ -66,6 +81,41 @@ const { actions } = store(
...
+					url.searchParams.delete( 'instant-search' );
+				}
+
+				const { actions: routerActions } = yield import(
+					'@wordpress/interactivity-router'
+				);
+
+				routerActions.navigate(
+					`${ window.location.pathname }${ url.search }`
+				);
```
sirreal: I don't think we should need to concatenate the URL string like this, but just use the url that's been constructed:

```suggestion
				routerActions.navigate( url.href );
```

sirreal (commented):
I've left some feedback focused mostly on the JavaScript implementation. I haven't tested yet.

packages/block-library/src/search/index.php
michalczaplinski: yeah, that makes sense of course! 🙂

michalczaplinski: done in 9a4e78f48985973d79128119c626c06947a836b1

packages/block-library/src/search/index.php
michalczaplinski: Agreed. Also somewhat out of scope: I would **love** a native templating
solution that uses HTML API under the hood 🙂.

packages/block-library/src/search/view.js
michalczaplinski: I think we can remove it. Did it have any purpose beyond what @sirreal shared,
@r-chrzan ?

packages/block-library/src/search/view.js
michalczaplinski: Oh, I see what you mean. You're saying that it's possible that a new keystroke could arrive AFTER we had already checked if `value !== state.search` but before `updateSearch()` finishes execution. In that case, the "previous" search would NOT be cancelled even though it's stale.

For the first version, a simple debounce should be enough (I think). It's just a UX optimization after all. But I will take a look at your example to see if we can add it!

packages/block-library/src/search/view.js
michalczaplinski: done in aaeee09cbdaba188bd2c5aa926f0a883bc7dbe56

packages/block-library/src/search/view.js
michalczaplinski: @cbravobernal might wanna chime in because we looked at it in our pair programming 🙂

packages/block-library/src/search/view.js
michalczaplinski: done in 818bc96a30ab3d944d0c6514032c6df9dbcaed6a

cbravobernal:
Watching the demo, I've seen that is showing only one result when searching "Hello", but the pagination is still there with 5 pages. I guess we are not updating the pagination links, is that right?

packages/block-library/src/search/view.js
cbravobernal: Although seems more complicated, the 'include abort' method seems much better. That way we save a bunch of requests, right?

michalczaplinski: The function will _virtually always_ bail / abort if there is a new keystroke. That's because **after** `yield new Promise( ( resolve ) => setTimeout( resolve, 300 ) )`  we check if the state had changed in the meantime:

https://github.com/WordPress/gutenberg/blob/818bc96a30ab3d944d0c6514032c6df9dbcaed6a/packages/block-library/src/search/view.js#L99-L101

Because state is updated **synchronously** in the Interactivity API, this should virtually always result in the correct debouncing.

---

Now, what I was referring to [above](https://github.com/WordPress/gutenberg/pull/63147/files#r1803359838) (and why I said virtually always_) is this: There is a tiny chance that the new keystroke will arrive within the few miliseconds that **this** part of the function is executing:

https://github.com/WordPress/gutenberg/blob/818bc96a30ab3d944d0c6514032c6df9dbcaed6a/packages/block-library/src/search/view.js#L102-L115

which would be **after** we check if the state has updated but before the function had finished. But I don't see that as a huge problem, and I like the simplicity of the current solution 🙂 .

sirreal: I disagree that the way this is implemented now is simpler. Concurrency is very complex and this is simplified by limiting concurrency, not by allowing it and attempting to control concurrent conditions.

Relying on pre- and post- conditions to manage concurrent operations is error prone. I'd much rather only allow 1 operation and worry about superseding and limiting concurrency than try to allow concurrent operations and reason about the way things can change concurrently. Reasoning about how things should happen in 300ms if certain things have changed or not is very difficult to do.

I'll share a suggestion for changes here shortly.

packages/block-library/src/search/view.js
sirreal: I'm pretty certain this method can go.

```suggestion
```

packages/block-library/src/search/view.js
```diff
@@ -66,6 +81,39 @@ const { actions } = store(
...
+					return;
+				}
+
+				const url = new URL( window.location.href );
+				if ( ! isEmpty( value ) ) {
+					state.search = value;
+					url.searchParams.set( 'instant-search', value );
+				} else {
+					url.searchParams.delete( 'instant-search' );
+				}
```
sirreal: I'm pretty sure we want to clear `paged` param whenever navigate is called. It doesn't make sense to replace the search results and start at another page that may or may not even exist.

Instead of `isEmpty`, a falsy check is done on the `value` string.

```suggestion
				const url = new URL( window.location.href );
				url.searchParams.delete( 'paged' );
				if ( value ) {
					url.searchParams.set( 'instant-search', value );
				} else {
					url.searchParams.delete( 'instant-search' );
				}
```

packages/block-library/src/search/view.js
sirreal: This is the full implementation I'd recommend here for this method:

```diff
diff --git before/packages/block-library/src/search/view.js after/packages/block-library/src/search/view.js
index 9bdc827506..66e9fea674 100644
--- before/packages/block-library/src/search/view.js
+++ after/packages/block-library/src/search/view.js
@@ -3,9 +3,8 @@
  */
 import { store, getContext, getElement } from '@wordpress/interactivity';
 
-const isEmpty = ( obj ) =>
-	[ Object, Array ].includes( ( obj || {} ).constructor ) &&
-	! Object.entries( obj || {} ).length;
+/** @type {( () => void ) | null} */
+let supersedePreviousSearch = null;
 
 const { state, actions } = store(
 	'core/search',
@@ -81,9 +80,9 @@ const { state, actions } = store(
 					actions.closeSearchInput();
 				}
 			},
-			*updateSearch() {
-				const { ref } = getElement();
-				const { value } = ref;
+
+			*updateSearch( e ) {
+				const { value } = e.target;
 
 				// Don't navigate if the search didn't really change.
 				if ( value === state.search ) {
@@ -93,16 +92,22 @@ const { state, actions } = store(
 				state.search = value;
 
 				// Debounce the search by 300ms to prevent multiple navigations.
-				// We can do this by yielding a promise that resolves after 300ms and
-				// then bailing out if the search has changed.
-				yield new Promise( ( resolve ) => setTimeout( resolve, 300 ) );
-				if ( value !== state.search ) {
+				supersedePreviousSearch?.();
+				const { promise, resolve, reject } = Promise.withResolvers();
+				const timeout = setTimeout( resolve, 300 );
+				supersedePreviousSearch = () => {
+					clearTimeout( timeout );
+					reject();
+				};
+				try {
+					yield promise;
+				} catch {
 					return;
 				}
 
 				const url = new URL( window.location.href );
-				if ( ! isEmpty( value ) ) {
-					state.search = value;
+				url.searchParams.delete( 'paged' );
+				if ( value ) {
 					url.searchParams.set( 'instant-search', value );
 				} else {
 					url.searchParams.delete( 'instant-search' );
```

michalczaplinski:
> Watching the demo, I've seen that is showing only one result when searching "Hello", but the pagination is still there with 5 pages. I guess we are not updating the pagination links, is that right?

@cbravobernal That's right. I haven't looked into it yet and plan to add it in follow-up PRs.

packages/block-library/src/search/view.js
r-chrzan: @michalczaplinski it was related to previous implementation where we've had query loop variation

gziolo:
Some notes from testing. There can be multiple Query blocks on a single page.

Example URL 1: http://localhost:8888/?query-1-page=2

Screenshot 2024-10-22 at 13 59 49

Example URL 2: http://localhost:8888/?paged=2&query-1-page=1

Screenshot 2024-10-22 at 14 01 10

Every Query block can be controlled independently using URL params to change the page that shows the items. 

That was summarized in this comment https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2214460127:

> When the "Inherit query from template" option is disabled, the query pagination uses a custom search parameter instead of the regular ?paged=2//page/2 of the inherited queries: ?query-X-page=P where X is the query id ($block->context['queryId']).

It's now `Default` (inherits fro the template) vs `Custom` query type:

Screenshot 2024-10-22 at 14 04 18

The rest still applies. The same approach should get replicated for handling search strings. When the user puts `?s=test` in the URL, then the search page loads, so it will work out of the box with `Default` query. When the Query is configured to `Custom` then it can be controlled with the combination of `query-X-page` and `query-X-s` to follow the same pattern allowing controlling all Query blocks on the page at the same time.

packages/block-library/src/search/view.js
michalczaplinski: > Relying on pre- and post- conditions to manage concurrent operations is error prone.

I agree that it's generally true. And especially true when dealing with distributed systems, etc. But in this specific case, I find it much harder to understand and reason about this:


```js
let supersedePreviousSearch = null;

...

supersedePreviousSearch?.();
const { promise, resolve, reject } = Promise.withResolvers();
const timeout = setTimeout( resolve, 300 );
supersedePreviousSearch = () => {
  clearTimeout( timeout );
  reject();
};
try {
  yield promise;
} catch {
  return;
}
```

than this:

```js
yield new Promise( ( resolve ) => setTimeout( resolve, 300 ) );
if ( value !== state.search ) {
  return;
}
```

I don't see how it could be argued that the latter is harder to reason about. I'm happy to be proven wrong, though if there's something I'm missing 🙂 And I'm happy to refactor this action in the future if the `Search & Query` includes other async operations or grows significantly in complexity in the future.

sirreal: There's a tradeoff between the simplicity of these lines and their correctness. I'd like like to take a slightly more complex and slightly more correct approach. It may be that we disagree on where it's best to go in this tradeoff. It's unlikely to have a big impact because the problems have potential to manifest in edge cases and/or in the future.

My point is that the execution is not properly limited. It is not debounced. Execution still continues, possibly multiple times, instead of being cancelled when superseded.

Here's a quick demonstration.

<details><summary>Apply this patch</summary>

```diff
diff --git before/packages/interactivity-router/src/index.ts after/packages/interactivity-router/src/index.ts
index 8a46d4ce35..f62ffbdaec 100644
--- before/packages/interactivity-router/src/index.ts
+++ after/packages/interactivity-router/src/index.ts
@@ -261,6 +261,11 @@ export const { state, actions } = store< Store >( 'core/router', {
 		 * @return  Promise that resolves once the navigation is completed or aborted.
 		 */
 		*navigate( href: string, options: NavigateOptions = {} ) {
+			console.log(
+				'navigate called with %d times, now: %s',
+				++calls,
+				href
+			);
 			const { clientNavigationDisabled } = getConfig();
 			if ( clientNavigationDisabled ) {
 				yield forcePageReload( href );
@@ -455,3 +460,4 @@ if ( globalThis.IS_GUTENBERG_PLUGIN ) {
 		);
 	}
 }
+let calls = 0;
```

</details>

Then open the search and quckly type a letter and delete it: `a`<kbd>DEL</kbd>`a`<kbd>DEL</kbd>. If you continue doing that the issue becomes more apparent.

What is the desired result? Nothing should happen until 300ms after the final keypress.

What actually happens? Depending on how quickly the inputs happened, multiple promises will queue up, 300ms will pass, and function execution will continue for all of them. Some of them will hit the equality check and return, others will pass the equality check and continue. **At this point, we must consider the implications of multiple executions of the rest of the function in rapid succession or concurrently.**

This is where I would much rather limit concurrency where it appears and not worry about it further.

If you've heard me out and disagree, go ahead and make a different tradeoff. I won't block this PR about this detail.

michalczaplinski: I gave it some further thought, and I concede that you're completely right. The implementation is simply incorrect when relying on the post-condition. Thank you for standing your ground and challenging my assumptions 🙂.

Another way to demonstrate why the behavior is buggy is to increase the timeout to something like 3000ms. Then, it's easier to observe that the function calls are not debounced properly.

I'll implement your recommendation!

sirreal:
If this is adding new block view script modules, some changes to the block-library package json will be required.

There's documentation added in https://github.com/WordPress/gutenberg/pull/66428 that hasn't landed on trunk yet, please take a look.

lib/experimental/editor-settings.php
```diff
@@ -37,6 +37,10 @@ function gutenberg_enable_experiments() {
 	if ( $gutenberg_experiments && array_key_exists( 'gutenberg-media-processing', $gutenberg_experiments ) ) {
 		wp_add_inline_script( 'wp-block-editor', 'window.__experimentalMediaProcessing = true', 'before' );
 	}
+
+	if ( $gutenberg_experiments && array_key_exists( 'gutenberg-search-query-block', $gutenberg_experiments ) ) {
+		wp_add_inline_script( 'wp-block-editor', 'window.__experimentalSearchQueryBlock = true', 'before' );
+	}
```
gziolo: `window.__experimentalSearchQueryBlock` is never used in the code.

lib/experiments-page.php
```diff
@@ -187,6 +187,18 @@ function gutenberg_initialize_experiments_settings() {
 		)
 	);
 
+	add_settings_field(
+		'gutenberg-search-query-block',
+		__( 'Instant Search and Query Block', 'gutenberg' ),
```
gziolo, line 192: What's the reasoning behind adding an experiment for this feature? In practice, it's just a new behavior on the Search block when placed inside the Query block.

packages/block-library/src/search/index.php
```diff
@@ -48,6 +48,8 @@ function render_block_core_search( $attributes ) {
 	// This variable is a constant and its value is always false at this moment.
 	// It is defined this way because some values depend on it, in case it changes in the future.
 	$open_by_default = false;
+	// Check if the block is using the enhanced pagination.
+	$enhanced_pagination = isset( $block->context['enhancedPagination'] ) && $block->context['enhancedPagination'];
```
gziolo, line 57: Note that this logic isn't guarded with the experiment check like in other places. In effect, this functionality automatically actives when the block is places inside Query block that has instant navigation enabled.

gziolo (dismissed):
I was testing the use cases outlined in my previous comment https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2429104611. 

I see that there are two issues still present. The proposed URL structure is incompatible with how WordPress works currently on the search results page. In particular, it is impossible to use two different URL params to control the same feature: `s` and `instant-search`. I tested it by modyfing the Search Results template and moving the Search block inside Query block.

Screenshot 2024-10-31 at 14 46 12
Screenshot 2024-10-31 at 14 46 38

Example URL:

https://playground.wordpress.net/scope:ambitious-noisy-town/?s=hello&instant-search=none

There is also a change in the behavior of the Search block with the button when the instant search is enabled. Without the feature, the user needs to click the button to trigger the search. With the feature, it gets triggered automatically on every keystroke. It's something to discuss further as it's a substantial change in the user experience.

lib/experimental/editor-settings.php
michalczaplinski: That's true. Good catch 👍

michalczaplinski: Updated in 2bb439e5e145fe81e954f0ca072db6d285067790

lib/experiments-page.php
michalczaplinski: I reasoned that the feature should remain an experiment while we polish it. 

I don't want the current PR to stay up for weeks or months. Instead, I'd prefer to merge an experimental version of the block, even if it's incomplete, and add the remaining functionality in follow-up PRs.

packages/block-library/src/search/index.php
michalczaplinski: Good catch as well! 👍 

Fixed in 1edf04482a631616d6ed630139ee66beb365334d

michalczaplinski:
Thanks for the review @gziolo!

You're right on both counts:

1. Instant search does not yet work correctly in the Search template. I should have mentioned it in the PR description (updated now; see "Limitations"). If we can leave the block as experimental, I'd prefer to add it in the following PR to keep the current PR smaller.
2. The search button is indeed somewhat redundant because the results update as you type. I'm not sure what the correct behavior should be. Let's discuss below.

---

The PR is ready for another round of review, **note the updated PR description.** 

Here are some of the open questions that I'd like to discuss.

### Search query URL param syntax

1. Which syntax to use for the URL search param for the `Default` query? Should we use `instant-search=` (used currently in this PR) or `is=` or something else? I slightly prefer `?instant-search=` because it's more explicit.

2. Which syntax to use for the URL search param for the `Custom` query?

   - `?instant-search-123=test`. This is the **current syntax** in this PR. It's consistent with the `Default` query syntax (`?instant-search=`).
   - `?query-123-search=test`. This syntax was suggested [here](https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2214460127) and is more consistent with `query-123-page`.
   - `?query-123-instant-search=test`. More verbose but consistent with both `query-123-page` and `Default` query syntax (`?instant-search=`).
   - `?query-123-s=test`. This is not consistent with the `Default` query, but IS consistent with the `?s=` search param commonly used for Search templates in WordPress.

3. Pagination for `Default` queries uses the `/page/2/` format when using pretty permalinks. Should it be removed from the URL? Here we have two options:

   - _**Remove it from the URL**_. Downside is possibly breaking links as we have to parse the URL with a regex and remove the `/page/X/` part.
   - _**Keep it in the URL**_ and use the `?paged` URL param. This works fine in my tests and redirects the request to the correct page. E.g. `/page/2/?paged=3` will redirect to `/page/3/`. The downside is that it requires a redirect so is slightly slower.

### Search submission behavior

1. How should submitting the search query (using the `Search` button of the Search block or pressing Enter) work now?

   Currently, it takes you to the Search template for the given search query, but that's probably not what we want since the Instant Search already provides the result.

2. Should the Search block inform the user **in the editor** that when it's placed inside of the Query block, it becomes an "instant search"?

Those two questions are related in my mind. If we can transparently inform the user in the editor that the Search block **_becomes an Instant Search block_**, then this Instant Search block should also have the correct behavior on the frontend that is **different** from the default Search block. The specific mechanism for doing this is not yet clear. Maybe we should be more explicit and add a toggle in the block settings that makes the Search block an Instant Search block?

Regardless of the solution in the editor, the Instant Search block _should_ (on the frontend):

1. Never show the `Submit` button. It's not necessary and redundant.
2. Never submit the search query when pressing Enter.

### Technical questions

1. The current implementation runs a **new Query** for many of the blocks when Instant Search is enabled. It seems fine, but I want to ensure it's not causing performance issues. Examples: [here](https://github.com/r-chrzan/gutenberg/blob/1edf04482a631616d6ed630139ee66beb365334d/packages/block-library/src/query-no-results/index.php#L53-L54), [here](https://github.com/r-chrzan/gutenberg/blob/1edf04482a631616d6ed630139ee66beb365334d/packages/block-library/src/query-pagination-numbers/index.php#L70-L71), [here](https://github.com/r-chrzan/gutenberg/blob/1edf04482a631616d6ed630139ee66beb365334d/packages/block-library/src/query-pagination-next/index.php#L68-L77) or [here](https://github.com/r-chrzan/gutenberg/blob/1edf04482a631616d6ed630139ee66beb365334d/packages/block-library/src/post-template/index.php#L94-L95).

2. I also have some lingering doubts about temporarily [substituting and then restoring](https://github.com/r-chrzan/gutenberg/blob/1edf04482a631616d6ed630139ee66beb365334d/packages/block-library/src/query-pagination-next/index.php#L68-L77) the global `$wp_query` in `query-pagination-next.php`.


cc @DAreRodz @gziolo @sirreal @r-chrzan

packages/block-library/src/query-pagination-next/index.php
```diff
@@ -51,10 +57,40 @@ function render_block_core_query_pagination_next( $attributes, $content, $block
...
+			// Store the original global $wp_query
+			$temp_query = $wp_query;
+
+			// Temporarily replace global $wp_query with a new query that includes the
+			// search query because get_next_posts_link() uses global $wp_query.
+			$wp_query = new WP_Query( $args );
+			$content  = get_next_posts_link( $label );
+
+			// Restore the original global $wp_query
+			$wp_query = $temp_query;
```
michalczaplinski: Are there any issues with replacing the global `$wp_query` and restoring it after calling `get_next_posts_link()` ?

gziolo:
> Limitations
> 
> ⚠️ Multiple Instant Search blocks using the Default query on in the same template are currently not supported.
> ⚠️ The Instant Search block does not yet work correctly when placed in the Search template. See https://github.com/WordPress/gutenberg/pull/63147#pullrequestreview-2408057726.

That’s fair. I’m fine with adding enhancement in steps. What’s the reasoning for adding the global handling for the default query in this PR if it isn’t expected to be functioning in the current shape?

> The current implementation runs a new Query for many of the blocks when Instant Search is enabled. It seems fine, but I want to ensure it's not causing performance issues.

It’s definitely something that needs to be further investigated. What’s the reason the same query needs to be recomputed for every child block? Can it be shared? What prevents from using a WP filter to modify the vars passed to the existing query in Gutenberg?

https://github.com/WordPress/wordpress-develop/blob/25d71f18fa9784f8a6d0e94518e7f04e2c17ecac/src/wp-includes/blocks.php#L2617

> Pagination for Default queries uses the /page/2/ format when using pretty permalinks

Is there a way to pass a canonical URL without the pagination applied to use it instead without all the complexity explained?

sirreal:
> Pagination for Default queries uses the /page/2/ format when using pretty permalinks. Should it be removed from the URL? Here we have two options:
> 
> - Remove it from the URL. Downside is possibly breaking links as we have to parse the URL with a regex and remove the /page/X/ part.
> - Keep it in the URL and use the ?paged URL param. This works fine in my tests and redirects the request to the correct page. E.g. /page/2/?paged=3 will redirect to /page/3/. The downside is that it requires a redirect so is slightly slower.

Is the information for how a URL should be constructed for a given site available somehow?

michalczaplinski:
> That’s fair. I’m fine with adding enhancement in steps. What’s the reasoning for adding the global handling for the default query in this PR if it isn’t expected to be functioning in the current shape?

That's a good point. I'm happy to handle the default queries in another PR if that makes is easier to review. I will split the current PR up.

> What’s the reason the same query needs to be recomputed for every child block? Can it be shared? What prevents from using a WP filter to modify the vars passed to the existing query in Gutenberg?

The reason is that the `query_loop_block_query_vars` filter does not run for the global queries (inherited from template). See [this explanation](https://wordpress.org/support/topic/query_loop_block_query_vars-doesnt-run-when-query-is-inherited-from-template/).

In the follow-up PR (handling the global/inherited query) I can try to use the `pre_get_posts` filter instead. That should be the appropriate filter for filtering the global/inherited query.

> Is there a way to pass a canonical URL without the pagination applied to use it instead without all the complexity explained?

I'm not aware of one. I took a look, and there is a [`wp_canonical_url` filter](https://developer.wordpress.org/reference/functions/wp_get_canonical_url/), but that only works for posts and will not work for archives, for example.

In general, I don't think it is possible to do it **reliably** server-side in PHP. A site could also, e.g., sit behind an Nginx proxy, which could be serving on a different URL and a WordPress function would have no way of knowing it.

michalczaplinski:
> Is the information for how a URL should be constructed for a given site available somehow?

@sirreal I think that the closest resource would be https://developer.wordpress.org/themes/basics/template-hierarchy/

The final URL depends on both the [permalink settings](https://wordpress.org/documentation/article/settings-permalinks-screen/) and the Template hierarchy. Is that what you had in mind?

sirreal:
Sorry that was unclear, I was wondering **if the site can be queried to determine** whether a search param should be used or some other form.

For this initial version the redirect seems alright and it can be improved in the future. It may be able to observe the redirect and adjust behavior accordingly.

michalczaplinski:
> Sorry that was unclear, I was wondering if the site can be queried to determine whether a search param should be used or some other form.

We could get the value of `get_option( 'permalink_structure' );` and pass it to the router's store on the server I guess. If a site is using "ugly permalinks" the value will be an empty string ([reference](https://wordpress.org/documentation/article/customize-permalinks/))

But this would not help us much. We'd still have to update the URL on the client side. And for this, we need to parse the URL on the client side to e.g update `mysite.lol/page/2#hash?and=query` to `mysite.lol/page/3#hash?and=query`.

gziolo:
There is still a lot of code repetition for every core block that can be nested inside the Query block. In the case of this PR, nearly the same changes have to be applied to 5 existing core blocks. Unfortunately, the same logic won't be available to custom blocks. I explored some alternatives, and I was surprised at how powerful it could be to use a filter, for example:

```php
function block_core_query_add_url_filtering( $context ) {
	if ( empty( $context['queryId'] ) ) {
		return $context;
	}
	$search_key  = 'query-' . $context['queryId'] . '-s';
	if ( ! isset( $_GET[ $search_key ] ) ) {
		return $context;
	}

	$context['query']['search'] = sanitize_text_field( $_GET[ $search_key ] );

	return $context;
}
add_filter( 'render_block_context', 'block_core_query_add_url_filtering' );
```

It's a quick proof of concept. This would execute too many times, as it would be applicable to every block nested inside the Query, but it's quite powerful. It probably could be improved by checking the parent block. It basically enabled filtering by a search term provided in the URL targeting a specific custom query. Example URL:

- http://localhost:8888/page/2?query-31-page=1&query-31-s=hey

In my testing, it turned out the URL structure for custom queries doesn't follow the settings for permalinks. The example I shared uses numeric permalinks for the default query, but not for custom queries. 

The open question is what would be the ultimate fix in WordPress core. The challenge is that the context is exposed based on the `query` attribute on the Query block, which is serialized in the database, so we would have to find a way to change the context based on the params passed in the URL matching the current query id.

michalczaplinski:
That's interesting, thanks for exploring it @gziolo.

I think that this approach might work. I'm going to try it now - we can definitely make it only run when the an ancestor block is Query Loop and has enhanced pagination enabled, etc.

> In my testing, it turned out the URL structure for custom queries doesn't follow the settings for permalinks. The example I shared uses numeric permalinks for the default query, but not for custom queries.

Yup, that's true. Only the default queries use the permalink settings, the custom queries use the `query-N-page=X` param.

> The open question is what would be the ultimate fix in WordPress core. The challenge is that the context is exposed based on the query attribute on the Query block, which is serialized in the database, so we would have to find a way to change the context based on the params passed in the URL matching the current query id.

So this means that the latest query (with search parameter) always gets saved in the database? And the problem is that a request might NOT include any (instant) search param, but the child blocks (query-pagination, query-numbers, etc.) could not distinguish between `context['query']['search']` that was added by the `render_block_context ` filter and the one that might come from the database, right?

Regardless, I'll give this approach a try 🙂

michalczaplinski:
> The open question is what would be the ultimate fix in WordPress core. The challenge is that the context is exposed based on the query attribute on the Query block, which is serialized in the database, so we would have to find a way to change the context based on the params passed in the URL matching the current query id.

I need some clarification on how this is a problem. 🤔 

We're only overriding the context at runtime when a request is being made. So the `"query"` context including the new `"search"` parameter should never be saved in the database. Or were you thinking about the case when there **is EXPLICITLY** a `search` parameter saved in the database, which would then be overridden by the param from the URL?

michalczaplinski:
I've started to explore the alternative in https://github.com/WordPress/gutenberg/pull/67013. We can discuss there 🙂 .

michalczaplinski:
> We're only overriding the context at runtime when a request is being made. So the "query" context including the new "search" parameter should never be saved in the database. Or were you thinking about the case when there is EXPLICITLY a search parameter saved in the database, which would then be overridden by the param from the URL?

This case is now handled via baa61883d349b01fcce68a1026374c40df747654. 

The remaining doubt I have is how to handle the URL. Should we (on the fronted) modify the URL to include a query param like `instant-search-11=<search-term>`? Currently, the URL does not change if we have a search term in the block's attributes:


https://github.com/user-attachments/assets/ca37cb49-b345-4c23-ad77-2384392a2a9d



I cherry picked baa61883d349b01fcce68a1026374c40df747654 to https://github.com/WordPress/gutenberg/pull/67013 as well.

gziolo:
> The open question is what would be the ultimate fix in WordPress core. The challenge is that the context is exposed based on the query attribute on the Query block, which is serialized in the database, so we would have to find a way to change the context based on the params passed in the URL matching the current query id.

I said that the search term is configurable, and the default value can be set in the editor through UI:

Screenshot 2024-11-15 at 09 23 00

In effect, the default value is always stored in the database and it doesn't have to be an empty string. Now, the functionality that is being explored has to offer a way to override that `search` value set in the block context by the Query block when rendering the block. Currently, the context for child block is equal to the value serialized in the database. Overriding on the frontend the context based on the URL param detected seems like the most intuitive approach as otherwise, every block has to check the URL to figure out if it should use that value or the one provided by theme/editor's user.

michalczaplinski:
> I said that the search term is configurable, and the default value can be set in the editor through UI:

Right, that's correct. I think we're talking about the same thing 🙂. I just did not realize that one can update the search term **in the UI** in addition to simply editing the markup (I did the latter in this [example](https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2477662458)). 

The effect is the same either way: the `query.search` attribute gets updated. 

It should now work as of https://github.com/WordPress/gutenberg/commit/baa61883d349b01fcce68a1026374c40df747654. Whenever the user updates the search term (via UI or just directly editing the markup), the search term will appear in the Search input on the frontend upon loading the page.

What I was considering here:

> The remaining doubt I have is how to handle the URL. Should we (on the fronted) modify the URL to include a query param like instant-search-11=<search-term>?

is something different. I wondered: What to do when loading a page for the first time, including an Instant Search block with a custom search term. In the video in the [previous comment](https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2477662458) I show that the word `Surfing` appears in the Search input, but the URL does NOT reflect that. I had thought that on initial load, we could (on the front-end in JS) update the URL to include the search term. 

However, I gave it some more thought, and it's probably not a good idea and would generate more problems than solve, I think. We should not just update the links in JS after the page loads based on the page's contents.

gziolo:
I discussed with @ntsekouras some options for the overall strategy of making the Query block filterable on the front end through URL params on WordPress Slack (https://make.wordpress.org/chat/):
https://wordpress.slack.com/archives/C02QB2JS7/p1731660635692969

packages/block-library/src/search/view.js
luisherranz: Please, even though WordPress' frontend support is unclear, always check the support of all the browser features we add. Support for `Promise.withResolvers` is still [a bit low (87%)](https://caniuse.com/mdn-javascript_builtins_promise_withresolvers) compared to the rest of the features we use, which have at least >93% support.

We can start a conversation here about what percentage of browsers WordPress should support; however, this API is not critical for the block's functionality, so I think it's simpler to look for an alternative implementation that is better supported.

packages/block-library/src/image/block.json
```diff
@@ -4,7 +4,14 @@
...
 	"category": "media",
-	"usesContext": [ "allowResize", "imageCrop", "fixedHeight", "postId", "postType", "queryId" ],
+	"usesContext": [
+		"allowResize",
+		"imageCrop",
+		"fixedHeight",
+		"postId",
+		"postType",
+		"queryId"
+	],
```
luisherranz: This seems unrelated to this pull request.

packages/block-library/src/post-content/block.json
```diff
@@ -69,4 +69,4 @@
 	},
 	"style": "wp-block-post-content",
 	"editorStyle": "wp-block-post-content-editor"
-}
\ No newline at end of file
+}
```
luisherranz: The same here.

If these changes are because there's some kind of linter error, perhaps this can be solved by merging `trunk`?

packages/block-library/src/query-pagination-next/index.php
```diff
@@ -54,7 +64,17 @@ function render_block_core_query_pagination_next( $attributes, $content, $block
...
+		// Add check for instant search experiment and search query
+		if ( $enhanced_pagination && $instant_search_enabled && ! empty( $search_query_direct ) ) {
+			$args         = array_merge(
+				build_query_vars_from_query_block( $block, $page ),
+				array( 's' => $search_query_direct )
+			);
+			$custom_query = new WP_Query( $args );
+		} else {
+			$custom_query = new WP_Query( build_query_vars_from_query_block( $block, $page ) );
+		}
```
luisherranz, line 76: If the Pagination Next block needs this change, I understand that the Pagination Previous block will also need it, right?

packages/block-library/src/search/view.js
sirreal: `Promise.withResolvers` is not essential and should be straightforward to replace 👍

luisherranz:
I was trying to test the UI, but I'm getting a `Maximum update depth exceeded` error in the `<SearchEdit>` component in the Editor.

packages/block-library/src/query-no-results/index.php
```diff
@@ -33,7 +47,13 @@ function render_block_core_query_no_results( $attributes, $content, $block ) {
...
+
+		// Add search parameter if instant search is enabled and search query exists
+		if ( $enhanced_pagination && $instant_search_enabled && ! empty( $search_query_direct ) ) {
+			$query_args['s'] = $search_query_direct;
+		}
+
+		$query = new WP_Query( $query_args );
 	}
 
 	if ( $query->post_count > 0 ) {
```
luisherranz, line 59: We can include the contents of this block but hide them when there are no posts. That way, the CSS and JS of the inner blocks will be correctly loaded and the blocks will work when the user navigates to a "no results" page in the client.

We can remove this workaround if we finally improve how client-side navigation works before WP 6.8.

packages/block-library/src/search/view.js
```diff
@@ -66,6 +80,58 @@ const { actions } = store(
...
+				// Don't navigate if the search didn't really change.
+				if ( value === ctx.search ) {
+					return;
+				}
+
+				ctx.search = value;
+
+				// Debounce the search by 300ms to prevent multiple navigations.
+				supersedePreviousSearch?.();
+				const { promise, resolve, reject } = Promise.withResolvers();
```
luisherranz: Support for `Promise.withResolvers` is still a bit low (87%) when compared with the rest of the things we use so far in the frontend (>93%). As this is not critical in any sense, let's find viable alternative that doesn't lower down our browser support.

packages/block-library/src/query-no-results/index.php
```diff
@@ -33,7 +47,13 @@ function render_block_core_query_no_results( $attributes, $content, $block ) {
...
 	} else {
 		$query_args = build_query_vars_from_query_block( $block, $page );
-		$query      = new WP_Query( $query_args );
+
+		// Add search parameter if instant search is enabled and search query exists
+		if ( $enhanced_pagination && $instant_search_enabled && ! empty( $search_query_direct ) ) {
+			$query_args['s'] = $search_query_direct;
+		}
+
+		$query = new WP_Query( $query_args );
```
luisherranz, line 56: Why do we need to create a new Query (and do so multiple times) instead of updating the current one?

luisherranz (changes requested):
I haven't tested the UI yet (due to [this problem](https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2504255612)), but here are a few suggestions/questions 🙂

luisherranz:
> 1. Pagination for `Default` queries uses the `/page/2/` format when using pretty permalinks. Should it be removed from the URL? Here we have two options:
>    
>    - _**Remove it from the URL**_. Downside is possibly breaking links as we have to parse the URL with a regex and remove the `/page/X/` part.
>    - _**Keep it in the URL**_ and use the `?paged` URL param. This works fine in my tests and redirects the request to the correct page. E.g. `/page/2/?paged=3` will redirect to `/page/3/`. The downside is that it requires a redirect so is slightly slower.

Regarding this, why don't we pass the root/canonical link from the server (using iAPI's `state`/`config`, for example) and use that link instead? I haven't looked at it, but I suppose it won't be difficult to obtain on the server.

michalczaplinski:
> I was trying to test the UI, but I'm getting a Maximum update depth exceeded error in the <SearchEdit> component in the Editor.

Oh, apologies about this. It was the very last commit that broke it. I've reverted it now (in 3dcf20ff0d95f547d7e22e68f4c2000fb5a7ccb0).

packages/block-library/src/image/block.json
michalczaplinski: Indeed, but I left it in as the file seemed to be formatted incorrectly. I've removed this change now.

packages/block-library/src/post-content/block.json
michalczaplinski: Same as in https://github.com/WordPress/gutenberg/pull/63147#discussion_r1863909846

I originally left it in as the file seemed to be formatted incorrectly. I've removed this change now.

packages/block-library/src/query-no-results/index.php
michalczaplinski: I'm not sure I understand what you mean. 

We're not adding (another) new Query here. The code of the `Query No Results` block was already creating a `new WP_Query()` (line 36) - we're just passing the search parameter to it.

michalczaplinski: You'll notice that in other blocks modified in this PR ([`post template`](https://github.com/WordPress/gutenberg/pull/63147/files/1b7b4bff8f1c528180652cf5d4eef8efd169fd4b#diff-4657bc4ddd88ab4cf8b48ad35b2fcf177edd40040c5976e2a3f9b89b52dd2bc7R91), [`query-pagination-next`](https://github.com/WordPress/gutenberg/pull/63147/files/1b7b4bff8f1c528180652cf5d4eef8efd169fd4b#diff-d58528784de1473c7d03da21f7c2cd742e8b36eefe09a252bb39d384ed892b39R75), or [`query-pagination-numbers`](https://github.com/WordPress/gutenberg/pull/63147/files/1b7b4bff8f1c528180652cf5d4eef8efd169fd4b#diff-14fbad12ea000a511336bf91b560cb395b654f881060942f2b4495f0bb16770bR66)) we're also not creating any additional queries beyond what had existed before. In `query-pagination-next` and `query-pagination-numbers` at most one query will run - which one depends on whether Instant Search is enabled.

packages/block-library/src/query-pagination-next/index.php
michalczaplinski: That's what I had had thought initially as well but actually it won't 🙂 

The `Query Pagination, Next` block has to create a new query to determine if it's on the last page of the results (if it is, then the block shouldn't show anything). We have to pass the search parameter to this new query. 

In `Query Pagination Previous`, no such new query is created, so it just works out of the box!

packages/block-library/src/search/view.js
michalczaplinski: Good catch. I'll update this 👍

packages/block-library/src/query-no-results/index.php
michalczaplinski: Oh, nice. That's a good idea 👍 

Can we add this in a follow-up PR though? I feel that it's not essential for the first version of the experimental block.

packages/block-library/src/query-pagination-next/index.php
michalczaplinski: I just noticed an edge case: 

If you navigate to a page with `?instant-search` param AND a pagination parameter like `http://gutenberg.local/?instant-search-18=qwponbfv&query-18-page=2`, then the `Pagination Previous` block shows up although it should not.

I'll hold on with fixing this for now though until we make a decision on whether we should go ahead with this PR or the approach in https://github.com/WordPress/gutenberg/pull/67181 or https://github.com/WordPress/gutenberg/pull/67013.

michalczaplinski:
> Regarding this, why don't we pass the root/canonical link from the server (using iAPI's state/config, for example) and use that link instead? I haven't looked at it, but I suppose it won't be difficult to obtain on the server.

For this we _could_ use [`wp_get_canonical_url`](https://developer.wordpress.org/reference/functions/wp_get_canonical_url/) but I have doubts about whether this is a reliable method. We cannot know for sure that the URL returned `wp_get_canonical_url` is going to the same as the URL the client sees. The WP server can sit behind a proxy like Cloudflare that can modify the hostname, pathname or params.

luisherranz:
> The WP server can sit behind a proxy like Cloudflare that can modify the hostname, pathname or params.

Are you sure? I don't think a WordPress server can function correctly without being configured with its final hostname. I haven't seen any WordPress installation that works well if the path or parameters are changed either.

It seems like [`wp_get_canonical_url` returns the page](https://developer.wordpress.org/reference/functions/wp_get_canonical_url/), so we might have better luck simply using [`get_permalink`](https://developer.wordpress.org/reference/functions/get_permalink/).

Let's try it, and if people report issues when using proxys, we will look for a solution.

packages/block-library/src/query-no-results/index.php
luisherranz: Oh, [I see](https://github.com/WordPress/gutenberg/blob/b867a5fe7c1133e906682eff84b53b6621dea559/packages/block-library/src/query-pagination-next/index.php#L57C43-L57C76).

michalczaplinski:
> Are you sure? I don't think a WordPress server can function correctly without being configured with its final hostname. I haven't seen any WordPress installation that works well if the path or parameters are changed either.

I'm not 100% sure 🙂 I see that it's possible to have a dynamic `WP_SITEURL`: 
- https://developer.wordpress.org/advanced-administration/wordpress/wp-config/#wp-siteurl
- https://stackoverflow.com/questions/26014426/how-to-set-dynamic-home-and-siteurl-in-wordpress

It might still work though. I'll try the approach using `get_permalink()`

packages/block-library/src/search/view.js
michalczaplinski: Done in 7b326cee38c577db3c6123895d6ddb8f25089c26

michalczaplinski:
I believe that the current feedback's been addressed 🙂. Thank you all! I will close this PR in favour of https://github.com/WordPress/gutenberg/pull/67181 as agreed in https://github.com/WordPress/gutenberg/issues/63053#issuecomment-2518480452.

As for the following issues:

- Use `get_permalink()` to get the canonical link value and use it when updating the URL for `Default` queries ([comment](https://github.com/WordPress/gutenberg/pull/63147#issuecomment-2504335923))
  - I'm working on this in [#67289](https://github.com/WordPress/gutenberg/pull/67289). The approach seems promising.
- Include contents of the `Query No Results` block but hide them with CSS ([comment](https://github.com/WordPress/gutenberg/pull/63147#discussion_r1860951647))
  - Happy to add this in a follow-up PR 👍 
- There's a bug where the `Previous` button does not work correctly ([comment](https://github.com/WordPress/gutenberg/pull/63147#discussion_r1863935064))
  - I'll fix this in [#67181](https://github.com/WordPress/gutenberg/pull/67181) where work on Instant Search continues.

luisherranz:
Thank you, Michał, and thanks also to @r-chrzan for starting this work 🙂

packages/block-library/src/query-pagination-next/index.php
michalczaplinski: This appears to be a general issue with the Query Loop block as discussed with @ntsekouras. I've opened an issue in https://github.com/WordPress/gutenberg/issues/67748

packages/block-library/src/query-no-results/index.php
luisherranz: Sure.

PR #56942, opened 2023-12-11
DataViews: add new field `types` and `formats` to modify the default render

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/55083

## What?

This PR is a proof of concept that experiments with replacing the `render` function of fields by `types` and `formats`.

Examples:

- A `date` field that uses the default render:

Captura de ecrã 2023-12-11, às 13 27 43

- A `text` field that needs links and a custom empty state:

Captura de ecrã 2023-12-11, às 13 29 47

## Why?

As the API currently stands, the fields' `render` function needs to account in which view layout the field would be rendered (see implementations for [templates](https://github.com/WordPress/gutenberg/blob/update/replace-render/packages/edit-site/src/components/page-templates/dataviews-templates.js#L153) and [pages](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-site/src/components/page-pages/index.js#L196)).

This approach demands more from the consumer. They need to account for accessibility (labels, empty cells should use `–`, etc.), keyboard navigation in different views (do not provide links in list view, for example), or remember to use the same styling in all the pages that use the same view, for example.

Some examples of the issues this causes are:

- https://github.com/WordPress/gutenberg/issues/56320
- https://github.com/WordPress/gutenberg/pull/56746#issuecomment-1841133118
> We should also set [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) on each item pointing to the primaryField, and possibly [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) pointing to the rest of the content.

- Different styles for page & template pages: _aka_ how dataview can provide a coherent experience when we have dozens of consumers?

| Templates | Pages |
| --- | --- |
| Captura de ecrã 2023-12-11, às 12 36 41 | Captura de ecrã 2023-12-11, às 12 36 33 |

- But also there's the question of how we'd allow a user to create a field with an API such as `render`.

## How?

- It introduces additional field types (`text`, `date`) that provide a default render for fields.
- It adds the concept of `formats`, which are modifiers of the default render provided by the type. Examples are:
  - Fields can declare a `link` format. It's up to the view if it uses it or not (e.g.: can be supported in `table` but not in `list`).
  - Fields can declare a different empty state from the default render.
[...]

PR #63148, opened 2024-07-04
DataViews: Update view fields rendering config

youknowriad:
Related to #55083 
Follow-up to #63127 

## What?

This PR does two things:

 - Removes the `layout` property from the `view` object. In favor of a `defaultFields` prop that can be passed optionally to the DataViews component. The defaultFields will tell which fields are visible by default on each layout (when switching layouts) 
 - Moves all the "badge", "media", "primary" configs into the `view.fields` config instead of being separate. So the view fields are defined like `[{ render: 'media', field: 'preview' }, "author", "description" ]`. The idea is that we'll be using this later to combine fields as well.

It's a significant change to the API, so it's a good idea to test most dataviews:

**Note** One known regression here is that I removed the "column/row" distinction in the grid layout. Personally, I think it was wrong to have this in the first place. I'd argue that we could potentially hide the "field labels" in this layout like we do for "list". 

**Note2** I believe badge fields are only supported in grid views right now, I think this can be easily extended to all layouts if we want.

## Testing Instructions

1- Test all layouts.
2- Test toggling field visibility for all layouts.
3- Test custom views.

### DIFF
--- a/packages/dataviews/src/view-actions.tsx
+++ b/packages/dataviews/src/view-actions.tsx
@@ -177,7 +177,7 @@ function FieldsVisibilityMenu< Item >( {
 		fields
 	);
 	const mediaFieldId = fieldRenderConfigs.find(
-		( fieldRender ) => fieldRender.render === 'media'
+		( fieldRender ) => fieldRender.format === 'media'
 	)?.field;
 	const hidableFields = fields.filter(
 		( field ) => field.enableHiding !== false && field.id !== mediaFieldId
--- a/packages/dataviews/src/view-actions.tsx
+++ b/packages/dataviews/src/view-actions.tsx
@@ -20,7 +20,8 @@ import { cog } from '@wordpress/icons';
 import { unlock } from './lock-unlock';
 import { SORTING_DIRECTIONS, sortLabels } from './constants';
 import { VIEW_LAYOUTS } from './layouts';
-import type { NormalizedField, View } from './types';
+import type { FieldRenderConfig, NormalizedField, View } from './types';
+import { normalizeFieldRenderConfigs } from './normalize-field-render-configs';
 
 const {
 	DropdownMenuV2: DropdownMenu,
@@ -35,6 +36,7 @@ interface ViewTypeMenuProps {
 	view: View;
 	onChangeView: ( view: View ) => void;
 	supportedLayouts?: string[];
+	defaultFields?: Record< string, FieldRenderConfig[] >;
 }
 
 interface PageSizeMenuProps {
@@ -59,12 +61,14 @@ interface ViewActionsProps< Item > {
 	view: View;
 	onChangeView: ( view: View ) => void;
 	supportedLayouts?: string[];
+	defaultFields?: Record< string, FieldRenderConfig[] >;
 }
 
 function ViewTypeMenu( {
 	view,
 	onChangeView,
 	supportedLayouts,
+	defaultFields,
 }: ViewTypeMenuProps ) {
 	let _availableViews = VIEW_LAYOUTS;
 	if ( supportedLayouts ) {
@@ -106,7 +110,9 @@ function ViewTypeMenu( {
 									return onChangeView( {
 										...view,
 										type: e.target.value,
-										layout: {},
+										fields:
+											defaultFields?.[ e.target.value ] ??
+											undefined,
 									} );
 							}
 							throw new Error( 'Invalid dataview' );
@@ -166,11 +172,16 @@ function FieldsVisibilityMenu< Item >( {
 	onChangeView,
 	fields,
 }: FieldsVisibilityMenuProps< Item > ) {
-	const hidableFields = fields.filter(
-		( field ) =>
-			field.enableHiding !== false && field.id !== view.layout.mediaField
+	const fieldRenderConfigs = normalizeFieldRenderConfigs(
+		view.fields,
+		fields
 	);
-	const viewFields = view.fields || fields.map( ( field ) => field.id );
+	const mediaFieldId = fieldRenderConfigs.find(
+		( fieldRender ) => fieldRender.render === 'media'
+	)?.field;
+	const hidableFields = fields.filter(
+		( field ) => field.enableHiding !== false && field.id !== mediaFieldId
+	);
 	if ( ! hidableFields?.length ) {
 		return null;
 	}
@@ -185,19 +196,23 @@ function FieldsVisibilityMenu< Item >( {
 			}
 		>
 			{ hidableFields?.map( ( field ) => {
+				const isVisible = !! fieldRenderConfigs.find(
+					( f ) => f.field === field.id
+				);
 				return (
 					<DropdownMenuCheckboxItem
 						key={ field.id }
 						value={ field.id }
-						checked={ viewFields.includes( field.id ) }
+						checked={ isVisible }
 						onChange={ () => {
 							onChangeView( {
 								...view,
-								fields: viewFields.includes( field.id )
-									? viewFields.filter(
-											( id ) => id !== field.id
+								fields: isVisible
+									? fieldRenderConfigs.filter(
+											( fieldRender ) =>
+												fieldRender.field !== field.id
 									  )
-									: [ ...viewFields, field.id ],
+									: [ ...fieldRenderConfigs, field.id ],
 							} );
 						} }
 					>
@@ -304,6 +319,7 @@ function _ViewActions< Item >( {
 	view,
 	onChangeView,
 	supportedLayouts,
+	defaultFields,
 }: ViewActionsProps< Item > ) {
 	return (
 		<DropdownMenu
@@ -320,6 +336,7 @@ function _ViewActions< Item >( {
 					view={ view }
 					onChangeView={ onChangeView }
 					supportedLayouts={ supportedLayouts }
+					defaultFields={ defaultFields }
 				/>
 				<SortMenu
 					fields={ fields }
--- a/packages/dataviews/src/view-actions.tsx
+++ b/packages/dataviews/src/view-actions.tsx
@@ -48,6 +48,7 @@ interface FieldsVisibilityMenuProps< Item > {
 	view: View;
 	onChangeView: ( view: View ) => void;
 	fields: NormalizedField< Item >[];
+	defaultFields?: Record< string, FieldRenderConfig[] >;
 }
 
 interface SortMenuProps< Item > {
@@ -171,6 +172,7 @@ function FieldsVisibilityMenu< Item >( {
 	view,
 	onChangeView,
 	fields,
+	defaultFields,
 }: FieldsVisibilityMenuProps< Item > ) {
 	const fieldRenderConfigs = normalizeFieldRenderConfigs(
 		view.fields,
@@ -185,6 +187,10 @@ function FieldsVisibilityMenu< Item >( {
 	if ( ! hidableFields?.length ) {
 		return null;
 	}
+	const normalizedDefaultFields = normalizeFieldRenderConfigs(
+		defaultFields?.[ view.type ],
+		[]
+	);
 	return (
 		<DropdownMenu
 			trigger={
@@ -205,6 +211,9 @@ function FieldsVisibilityMenu< Item >( {
 						value={ field.id }
 						checked={ isVisible }
 						onChange={ () => {
+							const fieldFormat = normalizedDefaultFields.find(
+								( f ) => f.field === field.id
+							);
 							onChangeView( {
 								...view,
 								fields: isVisible
@@ -212,7 +221,10 @@ function FieldsVisibilityMenu< Item >( {
 											( fieldRender ) =>
 												fieldRender.field !== field.id
 									  )
-									: [ ...fieldRenderConfigs, field.id ],
+									: [
+											...fieldRenderConfigs,
+											fieldFormat ?? field.id,
+									  ],
 							} );
 						} }
 					>
@@ -347,6 +359,7 @@ function _ViewActions< Item >( {
 					fields={ fields }
 					view={ view }
 					onChangeView={ onChangeView }
+					defaultFields={ defaultFields }
 				/>
 				<PageSizeMenu view={ view } onChangeView={ onChangeView } />
 			</DropdownMenuGroup>
--- a/packages/dataviews/src/view-list.tsx
+++ b/packages/dataviews/src/view-list.tsx
@@ -54,7 +54,7 @@ interface ListViewItemProps< Item > {
 	onSelect: ( item: Item ) => void;
 	primaryField?: NormalizedField< Item >;
 	store: CompositeStore;
-	defaultFields: NormalizedFieldRenderConfig[];
+	otherFields: NormalizedFieldRenderConfig[];
 }
 
 const {
@@ -75,7 +75,7 @@ function ListItem< Item >( {
 	onSelect,
 	primaryField,
 	store,
-	defaultFields,
+	otherFields,
 }: ListViewItemProps< Item > ) {
 	const registry = useRegistry();
 	const itemRef = useRef< HTMLElement >( null );
@@ -174,7 +174,7 @@ function ListItem< Item >( {
 									id={ descriptionId }
 								>
 									{ fields.map( ( field ) => {
-										const isVisible = !! defaultFields.find(
+										const isVisible = !! otherFields.find(
 											( fr ) => field.id === fr.field
 										);
 										if ( ! isVisible ) {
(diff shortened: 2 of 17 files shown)


packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -64,25 +58,27 @@ const { usePostActions } = unlock( editorPrivateApis );
...
-	},
-	[ LAYOUT_GRID ]: {
-		mediaField: 'preview',
-		primaryField: 'title',
-		badgeFields: [ 'sync-status' ],
-	},
+const defaultFields = {
+	[ LAYOUT_TABLE ]: [
+		'preview',
+		{ render: 'primary', field: 'title' },
```
oandregal:
```js
{
  fields: [
    { field: 'title', render: 'primary' }
  ]
}
```

Can we use `id` instead of `field`? It seems a bit inconsistent (different from the field config) and redundant (each item in the list is a field):

```js
{
  fields: [
    { id: 'title', render: 'primary' }
  ]
}
```

packages/edit-site/src/components/sidebar-dataviews/default-views.js
```diff
@@ -26,18 +26,24 @@ import {
...
-		mediaField: 'featured-image',
-		primaryField: 'title',
-	},
-	[ LAYOUT_LIST ]: {
-		primaryField: 'title',
-		mediaField: 'featured-image',
-	},
+export const defaultFields = {
+	[ LAYOUT_TABLE ]: [
+		{ render: 'primary', field: 'title' },
```
oandregal: What I like about absorbing `primary`, `media`, etc. as render functions is that they become "UI slots". It's like saying: render this field in the _slot_ of the primary field.

They had other meanings (cannot be hidden, etc.), we need to consider whether those are well covered and still clear.

packages/dataviews/src/view-actions.tsx
```diff
@@ -106,7 +110,9 @@ function ViewTypeMenu( {
 									return onChangeView( {
 										...view,
 										type: e.target.value,
-										layout: {},
+										fields:
+											defaultFields?.[ e.target.value ] ??
```
oandregal, line 115: This ties `view.fields` to `view.type`. However, when considering bundled or custom views, I see that it could be useful to have a few different `table` views that use different fields (same `view.type` but different `view.fields`), for example.

This is why I see this as a consumer decision, not dataview's. I'd remove the `defaultFields` from DataViews API and keep the `onChange` callback we have in consumers to augment the information based on the `type` change.

packages/edit-site/src/components/page-patterns/index.js
youknowriad: We can but I believe we use field to refer to fields elsewhere in the view object like in filters, order and we should be consistent with all the places in the view object

packages/dataviews/src/view-actions.tsx
youknowriad: I can see both ways but the current approach allows a clear cut (resets fields on each change) when you switch layouts which is a better default and if someone wants to keep or reuse fields, it's possible.

A small advantage too is that we don't have to repeat the same code in all consumers 

I'm open to change but I have a small preference as is. Would you be ok with trying this for some time ?

packages/dataviews/src/view-table.tsx
```diff
@@ -459,16 +468,10 @@ function ViewTable< Item >( {
...
-			[ view.layout.mediaField ].includes( field.id )
-	);
 	const hasData = !! data?.length;
-
-	const primaryField = fields.find(
-		( field ) => field.id === view.layout.primaryField
+	const fieldRenderConfigs = normalizeFieldRenderConfigs(
+		view.fields,
+		fields
 	);
```
oandregal, line 475:
```suggestion
	const fieldsToRender = getFieldsToRender(
		view.fields,
		fields
	);
```

How would you like this proposal? I find it makes the code self-explanatory from here on.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
oandregal: Having looked a bit more into the details, I think using `render: primary` is conflating concepts. Note that the fields still have to provide its own render function, even if they are the primary field. To me, we should only use `render: 'anything'` if that makes the field.render function redundant. For example, for combining fields, when we have types that provide a default render function, etc.

So, what about doing this instead:

```js
[
  { id: 'title', slot: 'primary' },
]
```

This has a couple of advantages:

- Conceptually, the purpose is explicit and it's also clear how the `render` function works (the field still has to provide one).
- It also simplifies the code: we can lift up the `normalizeFieldRenderConfigs`/`getFieldsToRender` to the DataViews component (shared across view types), and only work with `fields` internally — removing any reference to fieldsToRender. For example:

```js
// In dataviews.jsx

/*
 * - fields to render doesn't contain any field not to render (visible fields)
 * - at this point, it has also resolved any "preset render" (e.g.: render: 'my-custom-render', or type: 'date', etc.) into a function we can call.
 * - it preserves the slot: primary, media, etc.
 */
const fieldsToRender = getFieldsToRender( fields, view.fields );
 
// Later on, in view-table.jsx

<TableRow
  fields={ fieldsToRender }
  // Doesn't need the extra fieldsToRender prop
  />
```

packages/dataviews/src/view-actions.tsx
oandregal: Sure, not a blocker for me. I see how it can be a nice devexp touch.

The way I understand this prop is as a mechanism to augment or provide the "reset state" for the fields prop — any other view prop has a "natural" reset state (empty). Perhaps the main concern I have is that the name (`defaultFields`) doesn't really convey that concept.

Throwing some names in case some stick as self-explanatory:

- `onViewTypeReset`: this is the clearest I could come up with. It should also allow for providing a function besides the fields object. There's the question of whether other view props should be reset-configurable as well. I don't see a strong need for this, though perhaps `view.perPage` or future `view.density` could benefit from this as well. Ok to start with `fields` only.
- Other options: `viewTypeResetFieldConfig`, `viewTypeResetFields`, `fieldsToResetOnViewTypeChange`, `fieldsPerViewType`.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
youknowriad: The `render` is not a slot for me and should't override the `field.render` for me at all. Here's how I see things:

 - The `render` is a render config for a field. Potentially it can container more configuration that are passed to the field render function later but it's not mandatory.
 - The field.render function is a function that is meant to render the base rendering of the field (without any decoration like making it primary, making it a badge...)
 - There can be `render` types that render multiple fields later (vertical, horizontal...) and these don't really represent slots.

I know that as things are implemented today, they can seem like "slots" (especially for media and badges). I agree that this abstraction is not perfect but I believe the slot abstraction won't cover the horizonal/vertical use-cases while the current one will.

I think there's room for improvement though and I would like to see how we should evolve this once we get a better feeling of it.

packages/dataviews/src/view-actions.tsx
youknowriad: I agree that there's potential for adding similar reset behaviors to other properties. At this point, the prop becomes `defaultViews` or something like that.

packages/dataviews/src/view-list.tsx
```diff
@@ -318,20 +336,21 @@ export default function ViewList< Item >( props: ViewListProps< Item > ) {
...
-			)
+	const mediaFieldId = fieldRenderConfigs.find(
+		( fieldRender ) => fieldRender.render === 'media'
+	)?.field;
+	const primaryFieldId = fieldRenderConfigs.find(
+		( fieldRender ) => fieldRender.render === 'primary'
+	)?.field;
+	const mediaField = fields.find( ( f ) => f.id === mediaFieldId );
+	const primaryField = fields.find( ( f ) => f.id === primaryFieldId );
+	const defaultFields = fieldRenderConfigs.filter(
```
oandregal: `otherFields` would self-explain itself.

packages/dataviews/src/view-table.tsx
youknowriad: For me `fieldsToRender` and `view.fields` are actually just the same thing but in order to avoid manipulating multiple supported shapes, we're normalizing it with this function and adding default values in case view.fields is empty. 

I don't mind the change of name if it was the only place where this was used but we also have `normalizeFields` function which does the exact same thing but for `fields`.

Also I don't like the `fieldsToRender` variable because these are not fields, they are "rendering configurations for fields".

packages/dataviews/src/view-list.tsx
```diff
@@ -396,13 +415,14 @@ export default function ViewList< Item >( props: ViewListProps< Item > ) {
...
 						actions={ actions }
+						fields={ fields }
 						item={ item }
 						isSelected={ item === selectedItem }
 						onSelect={ onSelect }
 						mediaField={ mediaField }
 						primaryField={ primaryField }
 						store={ store }
-						visibleFields={ visibleFields }
+						defaultFields={ defaultFields }
```
oandregal: We could remove this prop and only operate with `fields={ otherFields }`, see [this comment](https://github.com/WordPress/gutenberg/pull/63148/files#r1666512418).

youknowriad:
> Note One known regression here is that I removed the "column/row" distinction in the grid layout. Personally, I think it was wrong to have this in the first place. I'd argue that we could potentially hide the "field labels" in this layout like we do for "list".

@jameskoster I would love an opinion on this.

jameskoster:
Primarily, the thing to avoid is this:

Screenshot 2024-07-05 at 12 41 56

The template description is unnecessarily tall and difficult to read due to it being squished by the label.

Hiding the label is one way to go about it. But we should consider that in some cases multiple fields could end up with similar values. With hidden labels it can be tricky to interpret one from the other. E.g. alt text, caption, and description in media items:

Screenshot 2024-07-05 at 12 45 17

In this case there should still be a way for users to determine the field label. 

To be honest I think there could be room for both options 🙊 

* For template description we can simply hide the label
* For alt text, caption, description we employ the stacked layout
* For short/obvious fields we use the badge layout

packages/edit-site/src/components/sidebar-dataviews/default-views.js
youknowriad: Thinking more about this. I think if we take a look at the "list" and "grid" views, the media and primary fields there are clearly slots. The "badge" is more of a rendering config and "primary" in "tables" is kind of a rendering config too. When we'll introduce "horizontal" and "vertical" combination, these will be rendering configs for sure and not slots.

It's clear that there are two concepts that are overlapping each other and that the perfect solution is not clear enough. I'd like to go with a unique concept for now (rendering config) even if it's bended a little bit by the view components for now and see how far we can go with that.

youknowriad:
@jameskoster Fair, do you see the "columns" and "rows" as something that can mix and match or do you see us rendering first columns and then rows (or the opposite like the screenshot).

jameskoster:
Mixing and matching can make it more difficult to visually scan. This is not ideal:

Screenshot 2024-07-05 at 13 01 58

So for now I'd lean towards grouping.

However, I don't know that we can categorically say that columns should always appear before rows or vice versa. Same width badges to be honest.

Could the order be configurable? 

If not I'd lean towards: Badges, Rows, Columns.

youknowriad:
Ok, so in the latest commit, I restored the support for "column" fields. The problem though is that the rendering config is "lost" when you hide a field, so if you toggle "description" off and on, it becomes a regular "row" field.

I think this can be solved by separating "visibleFields" from the "fields" definition, something we wanted to avoid with @oandregal but we can explore it as a follow-up WDYT?

packages/edit-site/src/components/page-patterns/index.js
oandregal: I see, that's a fair point 👍

packages/edit-site/src/components/sidebar-dataviews/default-views.js
oandregal: I agree with the direction, we just need more iterations to flesh out all the details.

I'd perhaps suggest renaming this prop to `renderAs: primary`, `renderConfig: primary`, etc. (a different name than `render` to prevent confusion with the `field.render` function).

packages/dataviews/src/view-table.tsx
oandregal: I see your point. I am coming from a different angle ([see](https://github.com/WordPress/gutenberg/pull/63148#discussion_r1666512418)). It's ok to ship this as it is until we fully clarify the role of the renderConfig/renderAs property.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
oandregal: I also agree we have working with two concepts here: "slots" and "decorations/formats" to how a field is rendered. 

For awareness, I had [this PR](https://github.com/WordPress/gutenberg/pull/56942) a while where I explore the idea of "formats" to make the field.render function declarative. Perhaps we can revisit this idea with what we know today. Specially because the field render functions are coupled with the view types: see [this example](https://github.com/WordPress/gutenberg/blob/df2793df94c71a9c1f1f8312b86a73b9150760c5/packages/edit-site/src/components/posts-app/posts-list.js#L339) where we decorate the field with a link depending on the view type — that's another example that could be absorbed by "decorations/formats".

jameskoster:
That's a bit clunky, but fine to address in a follow up, assuming you're confident a solution can be found.

oandregal:
> Ok, so in the latest commit, I restored the support for "column" fields. The problem though is that the rendering config is "lost" when you hide a field, so if you toggle "description" off and on, it becomes a regular "row" field.

As I see it, this means the `FieldsVisibilityMenu` should account for the `defaultFields` info when toggling on/off fields, like the `ViewTypeMenu` does.

packages/dataviews/src/view-actions.tsx
oandregal: Taking into account the conversations in the other threads, I think we should start this exploration with a better name than `defaultFields`: `fieldsRenderConfig` would work for me.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
youknowriad: I actually like the "format" name better than "render". The best name for me would be "type" actually, it's a type of rendering but it can be confusing with field types, so I'm probably switching to format for now. WDYT

youknowriad:
> As I see it, this means the FieldsVisibilityMenu should account for the defaultFields info when toggling on/off fields, like the ViewTypeMenu does.

Yes, but one of the goals of "defaultFields" was to also replace the visible fields per layout and not just define their formats. 🤔.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
oandregal: Format(s) sounds good to me as well 👍

oandregal: But then, if we go with formats, we should use it internally as well (e.g.: `fieldFormats = normalizeFieldFormats`, etc.) 🙈

youknowriad: yes, although if we think about what "view.fields" today does (in this PR), it serves both as field formats but also as "visible fields". A lot of the tension is coming from this in all the places. Maybe we can just embrace "fieldFormats" in this PR and just continue iterating/polishing as we add the other formats, solve the reordering and the loss of formats...

youknowriad: If you think about it, `view.fields` will represent the "template" of an item in the list view.

packages/dataviews/src/view-actions.tsx
youknowriad: but it's actually the default "fields" property of view. `view.fields`

packages/edit-site/src/components/sidebar-dataviews/default-views.js
youknowriad: Thinking more, I think `template` is the ideal world here and ideally its format is different between view types.

For tables it's just an array of fields, for list it's a `{ primary, media, otherFields }` for grid it's `{ primary, media, badgeFields, columnFields, rowFields }`. 

I think that's probably the ideal (different template shape per layout) but the fact that we toggle field visibility in the UI per field rather than as a template is what is making things hard.

That said, I'm going to stop bike-shedding right now and commit to the "format" we came up with for now so be able to follow-up with combination. I think we can revisit the template idea later.

youknowriad:
@jameskoster I applied @oandregal's suggestion to fix the issue. I'm still not entirely satisfied with what we have but I think we can run with this.

I'd appreciate some smoke testing.

youknowriad:
I'll also try to explore an alternative proposal that represent a smaller departure from what we have in a separate PR (I'm second guessing myself :P )

youknowriad:
Going to close this in favor of #63236

A refactoring like this one can be done later once we gain more clarity.

PR #63154, opened 2024-07-04
Remove third party domain

peterwilsoncc:
## What?
Replaces references to `mysite.com` with `example.com`

Fixes https://core.trac.wordpress.org/ticket/61570

## Why?
This comes out of discussion in the polyglots slack channel to avoid references to potentially live third party sites.

As WordPress has no control over the sites, this could be problematic if the content of the demonstration URL was either malware or content unsuitable for people of all ages.

## How?

`mysite.com` is replaced with `example.org`

Initially I used the homeURL of the site but @ramonjd and I were worried it could become too long for the copy space so we went with the reserved domain instead. If needs be, it can be switched to home URL later.

## Testing Instructions
1. Create a new page
2. Wait for the templates to load, close it immediately 
3. Click the Parent: None link
4. The inline documentation should show the example.org rather than mysite.com

## Screenshots or screencast 

Screenshot 2024-07-05 at 9 41 44 AM

### DIFF
--- a/docs/how-to-guides/metabox.md
+++ b/docs/how-to-guides/metabox.md
@@ -243,7 +243,7 @@ When the meta box area is saving, we display an updating overlay, to prevent use
 
 An example save url would look like:
 
-`mysite.com/wp-admin/post.php?post=1&action=edit&meta-box-loader=1`
+`example.com/wp-admin/post.php?post=1&action=edit&meta-box-loader=1`
 
 This url is automatically passed into React via a `_wpMetaBoxUrl` global variable.
 
--- a/docs/how-to-guides/metabox.md
+++ b/docs/how-to-guides/metabox.md
@@ -243,7 +243,7 @@ When the meta box area is saving, we display an updating overlay, to prevent use
 
 An example save url would look like:
 
-`example.com/wp-admin/post.php?post=1&action=edit&meta-box-loader=1`
+`example.org/wp-admin/post.php?post=1&action=edit&meta-box-loader=1`
 
 This url is automatically passed into React via a `_wpMetaBoxUrl` global variable.
 
--- a/packages/editor/src/components/page-attributes/parent.js
+++ b/packages/editor/src/components/page-attributes/parent.js
@@ -261,7 +261,9 @@ export function ParentRow() {
 								__(
 									"Child pages inherit characteristics from their parent, such as URL structure. For instance, if 'Web Design' is a child of 'Services,' its URL would be %1$s/services/web-design."
 								),
-								homeUrl.replace( /^https?:\/\//, '' )
+								homeUrl
+									.replace( /^https?:\/\//, '' )
+									.replace( /\/$/, '' )
 							) }
 							<p>
 								{ __(
--- a/packages/editor/src/components/page-attributes/parent.js
+++ b/packages/editor/src/components/page-attributes/parent.js
@@ -250,7 +250,8 @@ export function ParentRow() {
 						/>
 						<div>
 							{ __(
-								"Child pages inherit characteristics from their parent, such as URL structure. For instance, if 'Web Design' is a child of 'Services,' its URL would be mysite.com/services/web-design."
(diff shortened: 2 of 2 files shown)


packages/editor/src/components/page-attributes/parent.js
```diff
@@ -249,8 +256,12 @@ export function ParentRow() {
 							onClose={ onClose }
 						/>
 						<div>
-							{ __(
-								"Child pages inherit characteristics from their parent, such as URL structure. For instance, if 'Web Design' is a child of 'Services,' its URL would be mysite.com/services/web-design."
+							{ sprintf(
+								/* translators: %1$s The home URL of the WordPress installation without the scheme. */
+								__(
+									"Child pages inherit characteristics from their parent, such as URL structure. For instance, if 'Web Design' is a child of 'Services,' its URL would be %1$s/services/web-design."
```
peterwilsoncc: What's correct in American English: `'Services,'` or `'Services',`?

ramonjd: Normally American English puts punctuation inside speech quotes as far as I know. 

However, in this example, we're citing a UI label. There's no comma in the label - adding one would change its meaning. 

I think it's therefore acceptable to use `'Services',`.

docs/how-to-guides/metabox.md
```diff
@@ -243,7 +243,7 @@ When the meta box area is saving, we display an updating overlay, to prevent use
 
 An example save url would look like:
 
-`mysite.com/wp-admin/post.php?post=1&action=edit&meta-box-loader=1`
+`example.com/wp-admin/post.php?post=1&action=edit&meta-box-loader=1`
```
ramonjd: Very weakly-held opinion - would `example.org` reflect more the non-profit nature of the WordPress org?

packages/editor/src/components/page-attributes/parent.js
```diff
@@ -217,6 +217,13 @@ function PostParentToggle( { isOpen, onClick } ) {
 }
 
 export function ParentRow() {
+	const { homeUrl } = useSelect( ( select ) => {
+		const { getUnstableBase } = select( coreStore );
+		return {
+			homeUrl: getUnstableBase()?.home,
+		};
+	}, [] );
```
ramonjd: Adding the current site host is a nice touch.

The only reservation I'd have is for sites with very long domain names. It could push out the UI. In that case something like `example.com|org` could be safer? What do you reckon?

100% optional.

```suggestion
	const { homeUrl } = useSelect( ( select ) => ( {
		homeUrl: select( coreStore ).getUnstableBase()?.home
	} ), [] );
```

ramonjd: Probably not a huge risk:


Screenshot 2024-07-05 at 10 14 46 AM

ramonjd (approved):
LGTM

packages/editor/src/components/page-attributes/parent.js
peterwilsoncc: Yeah, I'm not sure what is better and most clear for the end user: accuracy or brevity. 

For wordpress.com sub-domains it can be quite long, the same will be true for the various make.wordpress.org blogs.

| Screenshot 2024-07-05 at 10 22 31 AM | Screenshot 2024-07-05 at 10 25 50 AM |

ramonjd: Maybe just use `example.com|org`? 

Less code, `0.00001` faster (no `useSelect`) and folks can always do a follow up.

peterwilsoncc: Cool, I've pushed that. I'll give others the chance to weigh in before merging.

hellofromtonya:
@ellatrix was this change included in today's package release for WP 6.6 RC3 package update?

ellatrix:
(Commented in Slack, yes)

PR #63157, opened 2024-07-05
Tooltip Component: Add custom class name support

aliaghdam:
## What?
The Tooltip component does not supports custom css class name.

## Why?
I added the custom class name support that that for many reasons can be used for design customization of tooltip.

## How?
I added the `className` prop and also added test to make sure it works properly.

## Screenshots or screencast 

Unlinked contributors: aliaghdam.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Unreleased
 
-
 ### Enhancements
 
 -   `Tooltip`: Add support for `className` prop ([#63157](https://github.com/WordPress/gutenberg/pull/63157)).
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,10 +1,11 @@
 <!-- Learn how to maintain this file at https://github.com/WordPress/gutenberg/tree/HEAD/packages#maintaining-changelogs. -->
 
 ## Unreleased
--   `Tooltip`: Add support for `className` prop ([#63157](https://github.com/WordPress/gutenberg/pull/63157)).
+
 
 ### Enhancements
 
+-   `Tooltip`: Add support for `className` prop ([#63157](https://github.com/WordPress/gutenberg/pull/63157)).
 -   `Toolbar`: Add support for `vertical` orientation ([#60123](https://github.com/WordPress/gutenberg/pull/60123)).
 
 ### Bug Fixes
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,6 +1,7 @@
 <!-- Learn how to maintain this file at https://github.com/WordPress/gutenberg/tree/HEAD/packages#maintaining-changelogs. -->
 
 ## Unreleased
+-   `Tooltip`: Add support for `className` prop ([#63157](https://github.com/WordPress/gutenberg/pull/63157)).
 
 ### Enhancements
 
--- a/packages/components/src/tooltip/index.tsx
+++ b/packages/components/src/tooltip/index.tsx
@@ -50,6 +50,7 @@ function UnforwardedTooltip(
 		position,
 		shortcut,
 		text,
+		className,
 
 		...restProps
 	} = props;
@@ -112,7 +113,7 @@ function UnforwardedTooltip(
 			{ isOnlyChild && ( text || shortcut ) && (
 				<Ariakit.Tooltip
 					{ ...restProps }
-					className="components-tooltip"
+					className={ className === undefined ? "components-tooltip" : "components-tooltip " + className }
 					unmountOnHide
 					gutter={ 4 }
 					id={ describedById }
--- a/packages/components/src/tooltip/index.tsx
+++ b/packages/components/src/tooltip/index.tsx
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import * as Ariakit from '@ariakit/react';
+import clsx from 'clsx';
 
 /**
  * WordPress dependencies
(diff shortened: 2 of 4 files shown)


packages/components/src/tooltip/index.tsx
```diff
@@ -112,7 +113,7 @@ function UnforwardedTooltip(
 			{ isOnlyChild && ( text || shortcut ) && (
 				<Ariakit.Tooltip
 					{ ...restProps }
-					className="components-tooltip"
+					className={ className === undefined ? "components-tooltip" : "components-tooltip " + className }
```
tyxla: This can be done with `clsx`:

```suggestion
					className={ clsx( 'components-tooltip', className ) }
```

You'll need to add an import at the top:

```
import clsx from 'clsx';
```

packages/components/src/tooltip/types.ts
```diff
@@ -58,6 +58,12 @@ export type TooltipProps = {
 	 * The text shown in the tooltip when anchor element is focused or hovered.
 	 */
 	text?: string;
+	/**
+	 * The custom class name for the tooltip.
+	 * The default class for tooltip is `components-tooltip` and always will be added.
+	 * The custom class name will be appended to the default class.
```
tyxla: Having a default class and appending a custom one to it is pretty regular behavior, so probably the description could be shortened to something like:

```suggestion
	 * Custom class name for the tooltip, in addition to the default `components-tooltip`.
```

packages/components/src/tooltip/test/index.tsx
```diff
@@ -105,6 +105,33 @@ describe( 'Tooltip', () => {
 				screen.getByRole( 'button', { name: 'Anchor' } )
 			).not.toHaveAttribute( 'data-foo' );
 		} );
+
+		it( 'should add custom class to the tooltip', async () => {
+			render( <Tooltip { ...props } className='foo' /> );
```
tyxla: Nit: With string literals as component props, we usually use double quotes:

```suggestion
			render( <Tooltip { ...props } className="foo" /> );
```

packages/components/src/tooltip/test/index.tsx
```diff
@@ -105,6 +105,33 @@ describe( 'Tooltip', () => {
...
+			);
+
+			// Check custom class
+			await waitFor( () =>
+				expect(
+					screen.getByRole( 'tooltip', {
+						name: 'tooltip text',
+					} )
+				).toHaveClass('foo')
+			);
```
tyxla: This could likely be cleaned up a bit, no need to do 2 separate `await waitFor` and 2 duplicate queries:

```suggestion
			// Check default and custom classnames
			await waitFor( () =>
				expect(
					screen.getByRole( 'tooltip', {
						name: 'tooltip text',
					} )
				).toHaveClass( 'components-tooltip', 'foo' )
			);
```

Also, we might want to keep the code style in consideration, there were some spacing issues that were causing the static analysis CI check to fail.

packages/components/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 <!-- Learn how to maintain this file at https://github.com/WordPress/gutenberg/tree/HEAD/packages#maintaining-changelogs. -->
 
 ## Unreleased
+-   `Tooltip`: Add support for `className` prop ([#63157](https://github.com/WordPress/gutenberg/pull/63157)).
```
tyxla: I think we should move this under the "Enhancements" section.

packages/components/src/tooltip/index.tsx
```diff
@@ -50,6 +50,7 @@ function UnforwardedTooltip(
 		position,
 		shortcut,
 		text,
+		className,
```
tyxla: Nit: The pre-existing props are alphabetically sorted, so perhaps we should move this one to the top to keep that order.

packages/components/src/tooltip/test/index.tsx
```diff
@@ -105,6 +105,33 @@ describe( 'Tooltip', () => {
 				screen.getByRole( 'button', { name: 'Anchor' } )
 			).not.toHaveAttribute( 'data-foo' );
 		} );
+
+		it( 'should add custom class to the tooltip', async () => {
```
tyxla: Should the test name reflect that we're testing both the original default class, and the additional custom one?

tyxla (commented):
Thanks for your contribution, @aliaghdam 🙌 

I've left some suggestions and questions for your consideration, let me know what you think.

packages/components/src/tooltip/types.ts
```diff
@@ -16,6 +16,10 @@ export type TooltipProps = {
 	 * **Note**: Accepts only one child element.
 	 */
 	children: React.ReactElement;
+	/**
+	 * Custom class name for the tooltip, in addition to the default `components-tooltip`.
```
ciampo: Internal class names are not part of a components' public APIs and we discourage their usage — let's not mention `.components-tooltip`

```suggestion
	 * Custom class name for the tooltip.
```

packages/components/CHANGELOG.md
```diff
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
+
```
tyxla: We can remove this empty line now:
```suggestion
```

tyxla (commented):
Thanks for addressing all the feedback @aliaghdam 🙌 

The PR should be good to go once @ciampo's feedback is addressed.

ciampo (approved):
LGTM  🚀 

Thank you for your work!

PR #63160, opened 2024-07-05
Resizing in Auto mode shouldn't add `columnStart` and `rowStart` values

tellthemachines:
## What?

Fixes a regression introduced in #62777.

When the grid interactivity experiment is enabled, it's possible to add a maximum number of columns to a Grid block in Auto mode. This should not cause grid children to get `columnStart` and `rowStart` values when they're resized with the resizing handles. `columnStart` and `rowStart` should only be set when the grid is in Manual mode.

Update: also removes dropzones from Auto grids with `columnCount` set.

## Testing Instructions

3. In Auto mode, set a non-zero number of "Columns" in the sidebar;
4. Resize one of the grid children with the resize handles and check that in its sidebar, under "Dimensions", the "Column" and "Row" fields aren't visible.

Before:

Screenshot 2024-07-05 at 11 24 13 AM

After:

Screenshot 2024-07-05 at 11 24 28 AM

### DIFF
--- a/packages/block-editor/src/components/block-mover/index.js
+++ b/packages/block-editor/src/components/block-mover/index.js
@@ -62,7 +62,7 @@ function BlockMover( {
 				orientation: getBlockListSettings( _rootClientId )?.orientation,
 				isManualGrid:
 					layout.type === 'grid' &&
-					!! layout.columnCount &&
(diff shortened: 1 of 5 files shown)


noisysocks (approved):
<3

PR #63162, opened 2024-07-05
Remove dotted border from grid dropzone

tellthemachines:
## What?

Fixes regression introduced in #61025 as per [this comment](https://github.com/WordPress/gutenberg/pull/62777#issuecomment-2194002144).

Removes dotted border from grid cell dropzones.

## Testing Instructions

1.In Gutenberg > Experiments enable the Grid experiment;
3. Switch the Grid to Manual mode in the block sidebar under "Layout";
4. Select one of the child blocks of the Grid;
5. Check that grid cells are highlighted with a solid line and not a dotted one.

Before:

Screenshot 2024-07-05 at 1 32 59 PM

After:

Screenshot 2024-07-05 at 1 33 54 PM

### DIFF
--- a/packages/block-editor/src/components/grid/style.scss
+++ b/packages/block-editor/src/components/grid/style.scss
@@ -27,7 +27,6 @@
 
 .block-editor-grid-visualizer__drop-zone {
 	background: rgba($gray-400, 0.1);
-	border: $border-width dotted $gray-300;
 	width: 100%;
 	height: 100%;
 


noisysocks:
Why do the cells change colour slightly when I change selection between the grid and the grid item? Or am I seeing things?

https://github.com/WordPress/gutenberg/assets/612155/b44fa466-45b6-4c70-b4b4-55a34726e37c

noisysocks (approved):
Fixes the issue 👍

tellthemachines:
> Why do the cells change colour slightly when I change selection between the grid and the grid item?

When grid items are selected the dropzones appear, but when it's just the parent grid selected they don't.

jasmussen:
Nice one. Thanks for following up.

I'd love to also see a 2px rounded radius on the cells, if that's possible.

PR #63166, opened 2024-07-05
Constrain `is-fullscreen-mode` admin body class to posts list

ntsekouras:
## What?
Fixes a bug introduced here: https://github.com/WordPress/gutenberg/pull/62705

This PR constrains the `is-fullscreen-mode` admin body class to posts list.

### DIFF
--- a/lib/experimental/posts/load.php
+++ b/lib/experimental/posts/load.php
@@ -7,13 +7,15 @@
 
 add_action( 'admin_menu', 'gutenberg_replace_posts_dashboard' );
 
-// Default to is-fullscreen-mode to avoid jumps in the UI.
-add_filter(
-	'admin_body_class',
-	static function ( $classes ) {
-		return "$classes is-fullscreen-mode";
-	}
-);
+if ( isset( $_GET['page'] ) && 'gutenberg-posts-dashboard' === $_GET['page'] ) {
+	// Default to is-fullscreen-mode to avoid jumps in the UI.
+	add_filter(
+		'admin_body_class',
+		static function ( $classes ) {
+			return "$classes is-fullscreen-mode";
+		}
+	);
+}
 
 /**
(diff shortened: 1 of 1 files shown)


lib/experimental/posts/load.php
```diff
@@ -7,13 +7,15 @@
...
 add_action( 'admin_menu', 'gutenberg_replace_posts_dashboard' );
 
-// Default to is-fullscreen-mode to avoid jumps in the UI.
-add_filter(
-	'admin_body_class',
-	static function ( $classes ) {
-		return "$classes is-fullscreen-mode";
-	}
-);
+if ( isset( $_GET['page'] ) && 'gutenberg-posts-dashboard' === $_GET['page'] ) {
```
nerrad, line 10: What confidence level is there that this route name will remain the same if/when this is ported to WordPress (given the `gutenberg` namespace used here). Should page references like this be returned from some sort of central map so that the name can be changed in one place when ported?

For example:

```
const PAGE_ROUTE_MAP = {
     POSTS_DASHBOARD: 'gutenberg-posts-dashboard',
     /** ...mapping for other known routes... **/
};
```

nerrad (approved):
Code wise looks okay (with the caveat in my comment). I haven't tested so I'm assuming you've confirmed this still addresses the original reason the code was added.

lib/experimental/posts/load.php
ntsekouras: This is just the v1 of posts lists that a GB experiment and we don't know when it's going to be added in core and will have many updates over time. Eventually it will have a specific admin page, where we won't have to do the redirection and these param checks, similar to the site editor page.

nerrad: Makes sense. Thanks for the fix!

Issue #63151, opened 2024-07-04
Data Views: Changing items per page to 100 results in poor loading experience


annezazu:
If you set Items per page to 100 then scroll down, you’ll be met with a blank screen as items seem to load. You’ll also see the bulk edit counting up here again to 100, matching the following issue https://github.com/WordPress/gutenberg/issues/63150

https://github.com/WordPress/gutenberg/assets/26996883/17575cc3-bacd-47b7-9a4c-6d4a66f0304c

This is using WP 6.5.5 & GB 18.7.1 & Chrome on MacOS. You can replicate this [using this link](https://href.li/?https://playground.wordpress.net/#%7B%20%22landingPage%22:%20%22/wp-admin/site-editor.php%22,%20%22preferredVersions%22:%20%7B%20%22php%22:%20%227.4%22,%20%22wp%22:%20%226.5.5%22%20%7D,%20%22phpExtensionBundles%22:%20%5B%20%22kitchen-sink%22%20%5D,%20%22features%22:%20%7B%20%22networking%22:%20true%20%7D,%20%22steps%22:%20%5B%20%7B%20%22step%22:%20%22login%22,%20%22username%22:%20%22admin%22,%20%22password%22:%20%22password%22%20%7D,%20%7B%20%22step%22:%20%22importWordPressFiles%22,%20%22wordPressFilesZip%22:%20%7B%20%22resource%22:%20%22url%22,%20%22url%22:%20%22https://block-museum.com/wp-content/uploads/2024/07/wordpress-playground-200-pages.zip%22%20%7D%20%7D,%20%7B%20%22step%22:%20%22installPlugin%22,%20%22pluginZipFile%22:%20%7B%20%22resource%22:%20%22wordpress.org/plugins%22,%20%22slug%22:%20%22gutenberg%22%20%7D,%20%22options%22:%20%7B%20%22activate%22:%20true%20%7D%20%7D%20%5D%20%7D) (might make you login) which will spin up a playground instance of the site I found this issue on.

Mamaduka:
The performance should be improved with #64504, but it probably won't fully resolve the issue.

annezazu:
I can no longer replicate this when I test with Gutenberg 19.7! Closing out as a result.

PR #63170, opened 2024-07-05
DataViews: add performance test for pages

ellatrix:
## What?

See #63151.

Adds a performance test for the pages data view (per page = 20, the default). These are just simple pages. Maybe at some point we can add featured images.

## Why?

It's easier to get a sense of progress when we're measuring.

## How?

See the test's code.

## Testing Instructions

There should be an extra entry for `loadPages` is the performance test.

### DIFF
--- a/test/performance/specs/site-editor.spec.js
+++ b/test/performance/specs/site-editor.spec.js
@@ -29,6 +29,7 @@ const results = {
 	listViewOpen: [],
 	navigate: [],
 	loadPatterns: [],
+	loadPages: [],
 };
 
 test.describe( 'Site Editor Performance', () => {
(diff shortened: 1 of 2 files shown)


test/performance/specs/site-editor.spec.js
```diff
@@ -384,6 +385,63 @@ test.describe( 'Site Editor Performance', () => {
...
+
+			const samples = 10;
+			for ( let i = 1; i <= samples; i++ ) {
+				// We want to start from a fresh state each time, without
+				// queries or patterns already cached.
+				await admin.visitSiteEditor();
+
+				const startTime = performance.now();
+
+				await page.getByRole( 'button', { name: 'Pages' } ).click();
```
youknowriad: I wonder if we should be using the "list" view. I think maybe using the "table" view would be better and we could set the url explicitly in the `admin.visitSiteEditor` call as it's less fragile to changes we might do to the root page of "pages".

test/performance/specs/site-editor.spec.js
```diff
@@ -384,6 +385,63 @@ test.describe( 'Site Editor Performance', () => {
...
+
+				await Promise.all(
```
youknowriad, line 434: What are we waiting for exactly here, can we add a clarifying comment?

test/performance/specs/site-editor.spec.js
youknowriad: list is also kind of tested in the "navigate" metric.

test/performance/specs/site-editor.spec.js
ellatrix: Added one, we are waiting for all pages to load.

test/performance/specs/site-editor.spec.js
ellatrix: Sure I changed it to start with the trash table view then. The reason for this is that we don't want to measure a complete page load.

youknowriad (approved):

test/performance/specs/site-editor.spec.js
ellatrix: > list is also kind of tested in the "navigate" metric.

Actually, I just checked this test, and it doesn't test loading the list view, it only tests selecting item from the list after it has loaded.

PR #63171, opened 2024-07-05
DateFormatPicker: use CustomSelectControl V2 legacy adapter

ciampo:
## What?

Part of #55023

Replace the V1 `CustomSelectControl` with the V2 legacy adapter in the date format picker control

## Why?

The goal is to ultimately replace the legacy V1 implementation entirely with the V2 legacy adapter. We're performing the migration gradually, and fixing any bugs / gaps as we go.

## How?

 - Import the V2 legacy adapter instead of the V1 implementation

## Testing Instructions

1. Insert query loop and pick the standard template
2. Select the post date in the newly added block
3. In the block sidebar, turn off the "Default format" toggle
4. Play with the "Choose a format" dropdown, make sure it looks and behaves like on `trunk`

_Note: on trunk, each option item in the dropdown has a an extra `6px` bottom margin that is the result of a style leak (see the second set of screenshots below). I assume that's not intended. The leak does not apply when using the new version of the component since its option items are not rendered using the `li` HTML element._

## Screenshots or screencast 

|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/01e62d62-cc04-4f37-9d56-fcdeeb1ec3d2" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/2f19a0fc-728c-4876-a525-454877a1d3b6" /> |
| Screenshot 2024-07-05 at 10 46 55 | Screenshot 2024-07-05 at 10 47 09 |

### DIFF
--- a/packages/block-editor/src/components/date-format-picker/index.js
+++ b/packages/block-editor/src/components/date-format-picker/index.js
@@ -8,10 +8,19 @@ import {
 	TextControl,
 	ExternalLink,
 	VisuallyHidden,
-	CustomSelectControl,
 	ToggleControl,
 	__experimentalVStack as VStack,
+	privateApis as componentsPrivateApis,
 } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import { unlock } from '../../lock-unlock';
+
+const { CustomSelectControlV2Legacy: CustomSelectControl } = unlock(
+	componentsPrivateApis
+);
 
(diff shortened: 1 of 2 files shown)


tyxla (approved):
LGTM 👍 

FWIW the v2 legacy version without the extra item bottom margin looks better than the v1, which was a bit too spread out.

The v1 also keeps closing after a couple of tab changes while comparing to the v2 in a neighbor browser tab. The v2 does not do it, which I find convenient. 

🚀

PR #63173, opened 2024-07-05
Global Styles: Add colors and typography to the browse styles section

scruffian:
## What?
Adds the color and typography presets to the browse styles panel.

## Why?
Since this is what we do in the left hand navigation, this adds consistency and means that its easier to find these settings.

## How?
Create a new shared component `SidebarNavigationScreenGlobalStylesContent` and use it in both the left and right Global Styles settings.

## Testing Instructions
1. Switch to a theme that has both color and typography presets
2. Open the Site Editor
3. Open Global Styles > Browse Styles
4. Confirm that you see both style variations, color preset and typography presets
5. Check that the dark, left Styles sidebar remains unchanged.

## Screenshots or screencast  
Screenshot 2024-07-05 at 09 45 54

### DIFF
--- a/packages/edit-site/src/components/global-styles/screen-style-variations.js
+++ b/packages/edit-site/src/components/global-styles/screen-style-variations.js
@@ -9,7 +9,7 @@ import { useZoomOut } from '@wordpress/block-editor';
  * Internal dependencies
  */
 import ScreenHeader from './header';
-import StyleVariationsContainer from './style-variations-container';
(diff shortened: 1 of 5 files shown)


richtabor (approved):
Makes sense. Pulls this panel 1:1 with the Styles panel on the dark material. Nice.

bgardner:
This is great @scruffian, just tested with Playground and looks amazing:

image

PR #46727, opened 2022-12-22
Inline preview: center align block in global styles inline preview

madhusudhand:
## What?

Blocks such as Button, Heading are top left aligned in the inline preview. This PR will keep the blocks center aligned for the better user experience.

| Before | After |
| --- | --- |
| image | image |
| image | image |

PR #63177, opened 2024-07-05
Apply same styles to block previews on inserter and Global Styles

### LINKED ISSUE #30029: FSE: Most site building blocks need previews and placeholders

MaggieCabrera:
## What?

This tries to apply the same styles to the Block Preview that lives in Global Styles to the one in the inserter, and tries to fix [some minor visual issues](https://github.com/WordPress/gutenberg/issues/30029#issuecomment-2209382539) for both in the meantime.

## Why?

The previews for block look very different in the GS sidebar and the inserter. Different CSS is applied to both and GS wasn't even scaling with the viewport width selected on the blocks settings. This aims to align the two together for a more consistent experience.

## How?

This tries to bring the changes from https://github.com/WordPress/gutenberg/pull/46727 to the inserter previews and polish any issues that arise for certain blocks because of it.

## Testing Instructions

- On a post or the site editor (the two experiences are different because the post editor has context for some of the blocks) open the block inserter.
- Scroll and hover over the blocks to see their previews. They shouldn't look broken. They should look centered.
- Go to Global Styles and select Blocks. Open them to check their previews. They should look like the inserter ones, with the caveat that the height of the preview here is fixed and some blocks will require scrolling. Nothing should look broken here either

## Screenshots or screencast 

**Before:**

Inserter |  GS
--- | --- 
Screenshot 2024-07-08 at 16 50 54 | Screenshot 2024-07-08 at 16 50 35
Screenshot 2024-07-08 at 16 51 00 | Screenshot 2024-07-08 at 16 50 45

**After:**

Inserter |  GS
--- | --- 
Screenshot 2024-07-08 at 16 43 03 |  Screenshot 2024-07-08 at 16 44 04
Screenshot 2024-07-08 at 16 43 37 | Screenshot 2024-07-08 at 16 43 50

### DIFF
--- a/packages/block-editor/src/components/inserter/preview-panel.js
+++ b/packages/block-editor/src/components/inserter/preview-panel.js
@@ -57,6 +57,7 @@ function InserterPreviewPanel( { item } ) {
 											min-height:${ Math.round( minHeight ) }px;
 											display:flex;align-items:center;justify-content:center; 
 										}
+										.is-root-container { width: 100%; }
 									`,
 								},
 							] }
--- a/packages/block-editor/src/components/inserter/preview-panel.js
+++ b/packages/block-editor/src/components/inserter/preview-panel.js
@@ -32,6 +32,13 @@ function InserterPreviewPanel( { item } ) {
 	}, [ name, example, initialAttributes ] );
 	// Same as height of BlockPreviewPanel.
 	const previewHeight = 144;
+	const sidebarWidth = 280;
+	const viewportWidth = example?.viewportWidth ?? 500;
+	const scale = sidebarWidth / viewportWidth;
+	const minHeight =
+		scale !== 0 && scale < 1 && previewHeight
+			? previewHeight / scale
+			: previewHeight;
 
 	return (
 		<div className="block-editor-inserter__preview-container">
@@ -40,13 +47,14 @@ function InserterPreviewPanel( { item } ) {
 					<div className="block-editor-inserter__preview-content">
 						<BlockPreview
 							blocks={ blocks }
-							viewportWidth={ example?.viewportWidth ?? 500 }
+							viewportWidth={ viewportWidth }
 							minHeight={ previewHeight }
 							additionalStyles={ [
 								{
 									css: `
 										body { 
 											padding: 24px;
+											min-height:${ Math.round( minHeight ) }px;
 											display:flex;align-items:center;justify-content:center; 
 										}
 									`,
--- a/packages/block-editor/src/components/inserter/preview-panel.js
+++ b/packages/block-editor/src/components/inserter/preview-panel.js
@@ -55,7 +55,7 @@ function InserterPreviewPanel( { item } ) {
 										body { 
 											padding: 24px;
 											min-height:${ Math.round( minHeight ) }px;
-											display:flex;align-items:center;justify-content:center; 
+											display:flex;align-items:center;
 										}
 										.is-root-container { width: 100%; }
 									`,
--- a/packages/block-editor/src/components/inserter/preview-panel.js
+++ b/packages/block-editor/src/components/inserter/preview-panel.js
@@ -49,9 +49,11 @@ function InserterPreviewPanel( { item } ) {
 							blocks={ blocks }
 							viewportWidth={ viewportWidth }
 							minHeight={ previewHeight }
-							additionalStyles={ [
-								{
-									css: `
+							additionalStyles={
+								//We want this CSS to be in sync with the one in BlockPreviewPanel.
+								[
+									{
+										css: `
 										body { 
 											padding: 24px;
 											min-height:${ Math.round( minHeight ) }px;
@@ -60,8 +62,9 @@ function InserterPreviewPanel( { item } ) {
 										}
 										.is-root-container { width: 100%; }
 									`,
-								},
-							] }
+									},
+								]
+							}
 						/>
 					</div>
 				) : (
--- a/packages/block-editor/src/components/inserter/preview-panel.js
+++ b/packages/block-editor/src/components/inserter/preview-panel.js
@@ -55,7 +55,8 @@ function InserterPreviewPanel( { item } ) {
 										body { 
 											padding: 24px;
 											min-height:${ Math.round( minHeight ) }px;
-											display:flex;align-items:center;
+											display:flex;
+											align-items:center;
 										}
 										.is-root-container { width: 100%; }
 									`,
--- a/packages/block-editor/src/components/inserter/preview-panel.js
+++ b/packages/block-editor/src/components/inserter/preview-panel.js
@@ -40,7 +40,7 @@ function InserterPreviewPanel( { item } ) {
 					<div className="block-editor-inserter__preview-content">
 						<BlockPreview
 							blocks={ blocks }
-							viewportWidth={ example?.viewportWidth ?? 350 }
+							viewportWidth={ example?.viewportWidth ?? 500 }
 							minHeight={ previewHeight }
 							additionalStyles={ [
 								{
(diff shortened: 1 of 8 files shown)


MaggieCabrera:
I've found an issue with the grid block (I don't think it happens to any other core blocks):

Screenshot 2024-07-08 at 16 31 27

This is caused by the flexbox rules we are applying to the body to center it. Because the grid block example doesn't have a width defined, it will shrink to center horizontally since there is no content pushing for a minimum width.

I think about this and I can only think of two solutions that maintain the centering using flexbox:

- Force the width of the example via attributes when it's failing
- Add an extra rule for the class `.is-layout-grid` forcing a min width

Both are similar solutions and I'm not particularly inclined by one or the other.

I can't think of a way to center the examples that won't incur in this problem or others like this one, but maybe I need to let it sit and a 💡 will come up

The reason this wasn't a problem before is because the centering was only being applied for the previews on global styles, and in there the group block preview only shows the normal version of group, not of grid.

MaggieCabrera:
@jasmussen @jameskoster  this one is mostly there, I need to keep testing more blocks and settle on a solution for the grid block (and add screenshots of the testing to the PR description) but if you have any insights that would change the direction of this PR they would be welcome

jasmussen:
This one looks great to me, substantially better than trunk in virtually all cases. There's some animation happening that's also  in trunk, but for some reason it stood out to me more here:

animation

animation2

No action item necessarily, except perhaps turning off any transitions if you find them while you're here.

The main point is the grid, which definitely looks more like a grid in trunk:

Screenshot 2024-07-09 at 10 45 43

You mention this:

> This is caused by the flexbox rules we are applying to the body to center it. Because the grid block example doesn't have a width defined, it will shrink to center horizontally since there is no content pushing for a minimum width.

> * Force the width of the example via attributes when it's failing
> * Add an extra rule for the class .is-layout-grid forcing a min width

Is there any reason why blocks inside the preview should _not_ be full-wide? If no, then it seems reasonable to apply a blanket 100% width to all blocks inside the previews.

Can you explain in more detail the extra rule?

packages/block-editor/src/components/inserter/preview-panel.js
```diff
@@ -38,9 +47,18 @@ function InserterPreviewPanel( { item } ) {
...
+							minHeight={ previewHeight }
 							additionalStyles={ [
-								{ css: 'body { padding: 24px; }' },
+								{
+									css: `
+										body { 
+											padding: 24px;
+											min-height:${ Math.round( minHeight ) }px;
+											display:flex;align-items:center;justify-content:center; 
+										}
```
MaggieCabrera, line 62:
```suggestion
										}
										.is-layout-grid { width: 100vw; }
```

@jasmussen from my tests we would need something like this (100% wouldn't work because there's another wrapper div that is already the wrong size)

MaggieCabrera:
> Is there any reason why blocks inside the preview should _not_ be full-wide? If no, then it seems reasonable to apply a blanket 100% width to all blocks inside the previews.
> 

Thanks for the input @jasmussen !

The main reason I see is cases like the button block, where if we apply the blanket width: 100% we would end up with the button aligned left. Same with blocks like site logo or anything that has a short text (login/out, site title, etc)

jasmussen:
Gotcha. It seems what you've found here is something you repro by putting a grid inside a row block:

Screenshot 2024-07-09 at 11 20 50

Which suggests two paths foward, unless I'm missing nuance:

1. Apply some additional CSS in the grid block, which would fix both cases
2. Avoid the flex behavior on the parent in the preview

I don't know that I like either of those paths forward. Perhaps 2 if possible. But at the same time, flex is the main missing ingredient to make all these previews better, yes?

How do we generally feel about adding bespoke CSS for individual previews? Is this something plugin blocks can do? If yes, seems fine to do here.

MaggieCabrera:
@jasmussen I found out that the flexbox issue was causing problems with the separator block too (it would resolve to width 0 and be invisible). I changed tactics and did the following:

- Set the root container that contains the block list in the preview to be width: 100%. This voids the horizontal alignment but fixes the issue flexbox was causing
- I centered as many blocks as I could with their current supports (some need a wrapper to be centered, that is not something the current block previews support and I think it probably should stay that way.

I think this solves the breakage and the tradeoffs are minimal (some blocks are left aligned but still look ok). It also doesn't require any actions from third party blocks.

The problem you pointed out that happens when combining flexbox and grid I couldn't figure out an easy way to solve, since that is how CSS interacts in the wild, and seemed a bit out of scope here but still important to have in mind

jasmussen:
Seems fine with me. I think this one could use a broader code review than I can provide, though.

MaggieCabrera:
> Seems fine with me. I think this one could use a broader code review than I can provide, though.

I'll go fish for technical reviews

packages/block-editor/src/components/inserter/preview-panel.js
```diff
@@ -38,9 +47,19 @@ function InserterPreviewPanel( { item } ) {
...
+							viewportWidth={ viewportWidth }
+							minHeight={ previewHeight }
 							additionalStyles={ [
-								{ css: 'body { padding: 24px; }' },
+								{
+									css: `
+										body { 
+											padding: 24px;
+											min-height:${ Math.round( minHeight ) }px;
+											display:flex;align-items:center;
```
scruffian:
```suggestion
											display:flex;
											align-items:center;
```
for readability

scruffian (approved):
This is looking great. Well done 👏

packages/edit-site/src/components/global-styles/block-preview-panel.js
```diff
@@ -53,9 +54,11 @@ const BlockPreviewPanel = ( { name, variation = '' } ) => {
 						{
 							css: `
 								body{
-									min-height:${ previewHeight }px;
-									display:flex;align-items:center;justify-content:center;
+									padding: 24px;
+									min-height:100%;
+									display:flex;align-items:center;
```
scruffian:
```suggestion
									display:flex;
									align-items:center;
```

packages/edit-site/src/components/global-styles/block-preview-panel.js
```diff
@@ -53,9 +54,12 @@ const BlockPreviewPanel = ( { name, variation = '' } ) => {
 						{
 							css: `
 								body{
-									min-height:${ previewHeight }px;
-									display:flex;align-items:center;justify-content:center;
+									padding: 24px;
```
scruffian, line 59: We could add a comment here about keeping this CSS in sync with the lines above.

PR #33654, opened 2021-07-23
[RNMobile][Embed block] Block settings

ceyhun:
## Description

If embed block variation and site's theme supports it, a settings icon is shown inside the block and a bottom sheet opens when pressing it, which shows a toggle for “Resize for smaller devices.”

Gutenberg Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3753

## How has this been tested?

Which embed variations support this setting can be checked from the `responsive` attribute in `variations.js`:

Supported:
https://github.com/WordPress/gutenberg/blob/3e77c785d114e3a4cad7d34b5df66a0fbea6f1f9/packages/block-library/src/embed/variations.js#L44

Unsupported:
https://github.com/WordPress/gutenberg/blob/3e77c785d114e3a4cad7d34b5df66a0fbea6f1f9/packages/block-library/src/embed/variations.js#L338

#### For supported case:
1. Add embed block
2. Enter a supported embed variation URL (e.g. https://www.youtube.com/watch?v=8qpkAdp1K0M)
3. Press settings button
4. Turn toggle off
5. Update post and preview
6. Close preview and turn toggle on
7. Update post and preview again
 
#### For unsupported case:
~A WP.com free plan site may not support `responsive-embeds`, so it can be used to test where the setting doesn't appear.~ This is actually not related to the site plan, but only related to which theme the site is using and if that theme supports `responsive-embeds`. For example, `Mayland` theme supports responsive-embeds, but `Baskerville 2` doesn’t.

Another way to test unsupported case is with an unsupported embed variation like `Amazon Kindle` embed.

## Known issues

Facebook and Instagram embed variations have `block` in their `scope` as they are deprecated:

https://github.com/WordPress/gutenberg/blob/aa22cd7714bbbd63b95dd9cfab0a894cba6d5ec2/packages/block-library/src/embed/variations.js#L66

This results in `responsive` attribute not being passed into the embed block and as a result it's assumed `false`.

## Screenshots 

iOS | Android
-- | -- 
|

Toggle Off | Toggle On
-- | -- 
|

## Types of changes

New feature (non-breaking change which adds functionality)

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
[...]

PR #63183, opened 2024-07-05
[Mobile] - Remove `themes` from supported endpoints

geriux:
**Related PRs:**
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/6967

## What?
Due to recent changes, the editor now fetches the `/themes` endpoint when it opens, this is causing an issue on Android devices.

**Note:** The theme endpoint was added in https://github.com/WordPress/gutenberg/pull/33654 to be able to show the `Resize for smaller devices` block option for the Embed block but I haven't been able to see this option with a testing site I have, same on the Web editor.

## Why?
This is currently affecting self-hosted sites preventing them from using the editor which freezes after attempting to call this endpoint.

The native host apps already fetch the themes data and sends the information the editor needs. It appears there are some issues with how the native editor makes the GET request through the bridge and the host app.

## How?
By removing `themes` from the `SUPPORTED_ENDPOINTS` list for now.

> [!NOTE]  
> Use the following build to test:
> - [Android](https://github.com/wordpress-mobile/WordPress-Android/pull/21032#issuecomment-2210636299)

- Install the app (fresh installation)
- Log in using a self-hosted site without Jetpack and Gutenberg plugin enabled
- Open the editor
- **Expect** to be able to edit/create content
- **Expect** to see the default theme colors in the Color palettes (Text/Background color)

## Screenshots or screencast 

Before|After
-|-
<video src="https://github.com/WordPress/gutenberg/assets/4885740/0065fabf-ed4c-4c62-8e4e-1002935a6587" width=250 />|<video src="https://github.com/WordPress/gutenberg/assets/4885740/655d083c-035a-4917-a838-8bb7b82aa919" width=250 />

### DIFF
--- a/packages/react-native-editor/src/api-fetch-setup.js
+++ b/packages/react-native-editor/src/api-fetch-setup.js
@@ -7,9 +7,13 @@ import { applyFilters } from '@wordpress/hooks';
 
 const SUPPORTED_METHODS = [ 'GET', 'POST' ];
 // Please add only wp.org API paths here!
-// Don't add /themes endpoint due to an issue on Android
-// with the wp.org API implementation for React Native.
 const SUPPORTED_ENDPOINTS = {
+	// Temporarily disabling themes endpoint calls within the editor.
+	// Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/21034
+	// The editor's GET requests to the themes endpoint are not functioning as expected.
+	// This is likely due to the method used for performing GET requests within the host Android app.
+	// TODO: Investigate and resolve the issue with GET requests from the editor.
+	// Until then, themes endpoint calls are disabled to prevent unexpected behavior.
 	GET: [
 		/wp\/v2\/(media|categories|blocks)\/?\d*?.*/i,
 		/wp\/v2\/search\?.*/i,
--- a/packages/react-native-editor/src/api-fetch-setup.js
+++ b/packages/react-native-editor/src/api-fetch-setup.js
@@ -7,9 +7,11 @@ import { applyFilters } from '@wordpress/hooks';
 
 const SUPPORTED_METHODS = [ 'GET', 'POST' ];
 // Please add only wp.org API paths here!
+// Don't add /themes endpoint due to an issue on Android
+// with the wp.org API implementation for React Native.
 const SUPPORTED_ENDPOINTS = {
 	GET: [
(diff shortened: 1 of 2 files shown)


packages/react-native-editor/src/api-fetch-setup.js
```diff
@@ -7,9 +7,11 @@ import { applyFilters } from '@wordpress/hooks';
 
 const SUPPORTED_METHODS = [ 'GET', 'POST' ];
 // Please add only wp.org API paths here!
+// Don't add /themes endpoint due to an issue on Android
+// with the wp.org API implementation for React Native.
```
dcalhoun: It may be worth publishing a comment on https://github.com/wordpress-mobile/WordPress-Android/issues/21005 capturing the identified issue and linking to it within this code comment. Alternative, inlining an explanation in the code comment would work well too. WDYT?

dcalhoun (approved):
Thank you for addressing this issue! 🙇🏻 

I tested these changes on a physical Samsung Galaxy S20 and iPhone 15 simulator. The test plan succeeded and I did not encounter any issues related to theme colors — both with and without an activated Gutenberg plugin. 

> * **Expect** to see the default theme colors in the Color palettes (Text/Background color)

To avoid any confusion, I interpreted this to be mean _default colors_ packaged with WordPress core. I.e., the colors are not from/associated with the active theme. With that understanding, the expectation was met. 

After activating the Gutenberg plugin, the _theme's color_ were expectedly then present and available in the block editor. 

💡 We might consider adding a change log entry.

geriux:
> To avoid any confusion, I interpreted this to be mean default colors packaged with WordPress core. I.e., the colors are not from/associated with the active theme. With that understanding, the expectation was met.

That's it, the expected colors are the default ones from core 👍 

> 💡 We might consider adding a change log entry.

I'll add it now 🚀 

Thanks for the review and testing!

packages/react-native-editor/src/api-fetch-setup.js
geriux: Good idea! I'll update it 👍

geriux: Updated in https://github.com/WordPress/gutenberg/pull/63183/commits/7a5d33c555f1eef1add380a40d4f75eef885b1dd

PR #63184, opened 2024-07-05
Inspector: Use the TabbedSidebar component

scruffian:
## What?
Implements the new TabbedSidebar component in the Inspector Controls

## Why?
For consistency it's good to use the same component for all our tabbed sidebars. `TabbedSidebars` is used for the List View and the Inserter, so this adds it to the Inspector as well.

## How?
Refactoring to use `TabbedSidebar`.

## Testing Instructions
1. Open the Site Editor
2. Open the inspector
3. Check that the behaviour is the same as trunk

## Screenshots or screencast 
| Before    | After |
| -------- | ------- |
| Screenshot 2024-07-05 at 12 57 32  | Screenshot 2024-07-05 at 12 56 36 |


packages/block-editor/src/components/inserter/menu.js
```diff
@@ -316,6 +316,7 @@ function InserterMenu(
 		>
 			<div className="block-editor-inserter__main-area">
 				<TabbedSidebar
+					className="block-editor__tabbed-sidebar"
```
scruffian, line 319: We need this so that we can set a width for only the inserter a the list view.

packages/block-editor/src/components/tabbed-sidebar/index.js
```diff
@@ -16,54 +21,65 @@ import { unlock } from '../../lock-unlock';
...
-			>
-				<div className="block-editor-tabbed-sidebar__tablist-and-close-button">
-					<Button
-						className="block-editor-tabbed-sidebar__close-button"
-						icon={ closeSmall }
-						label={ closeButtonLabel }
-						onClick={ () => onClose() }
-						size="small"
-					/>
+		<Tabs.Context.Provider value={ tabsContextValue }>
```
scruffian: We probably shouldn't output this unless we have a value. Maybe we don't actually need it at all?

packages/editor/src/components/sidebar/index.js
```diff
@@ -28,7 +29,6 @@ import PluginSidebar from '../plugin-sidebar';
 import PostSummary from './post-summary';
 import PostTaxonomiesPanel from '../post-taxonomies/panel';
 import PostTransformPanel from '../post-transform-panel';
-import SidebarHeader from './header';
```
scruffian, line 31: We can delete this file

packages/editor/src/components/sidebar/index.js
```diff
@@ -176,18 +223,16 @@ const Sidebar = ( { extraPanels, onActionPerformed } ) => {
 	);
 
 	return (
-		<Tabs
-			selectedTabId={ tabName }
-			onSelect={ onTabSelect }
-			selectOnMove={ false }
-		>
+		<Tabs selectOnMove={ false }>
```
scruffian: I think we can remove this component entirely.

packages/interface/src/components/complementary-area/index.js
```diff
@@ -183,6 +183,7 @@ function ComplementaryArea( {
 	title,
 	toggleShortcut,
 	isActiveByDefault,
+	hideHeader = false,
```
scruffian: I don't like adding a prop for this, but I don't have another idea.

youknowriad (commented):
It feels like more code to use a shared component rather than avoid it? Isn't that an issue?

isn't the `hideHeader` the same thing as passing an empty `header`?

Just wondering if we're actually building the right abstraction basically or if we're bending something to use an abstraction.

packages/editor/src/components/sidebar/index.js
scruffian: Done in [2d23a17](https://github.com/WordPress/gutenberg/pull/63184/commits/2d23a17d1eb3391a0b2c508ef9bb4a15c7c4bbcc)

scruffian:
> It feels like more code to use a shared component rather than avoid it? Isn't that an issue?

IMO The reason for the abstraction is less about minimizing code and more about having consistent UX. At the moment each of these panels looks and behaves differently, and even when we have made efforts to get them in sync they have moved apart again. The hope is that by using the same component we'll have the same UX for all panels.

scruffian:
Incidentally now I have removed the `SidebarHeader` component this PR does now remove more code than it adds:

`+129 −143`

scruffian:
> isn't the hideHeader the same thing as passing an empty header?

No, if you don't pass a header then the `ComplementaryArea` component has a default.

packages/interface/src/components/complementary-area/index.js
```diff
@@ -295,47 +295,51 @@ function ComplementaryArea( {
...
-									}
-									isPressed={ isPinned }
-									aria-expanded={ isPinned }
-									size="compact"
-								/>
-							) }
-						</>
-					) }
-				</ComplementaryAreaHeader>
+				{ header !== false && (
```
scruffian, line 298: This is the only change here - if header is explicitly set to false then we don't output it. This might break other things.

packages/editor/src/components/sidebar/index.js
```diff
@@ -107,22 +103,43 @@ const SidebarContent = ( {
...
+						title: __( 'Block' ),
+						panel: <BlockInspector />,
+					},
+				] }
+				onClose={ () => disableComplementaryArea( 'core' ) }
+				onSelect={ onSelect }
+				selectedTab={ tabName }
+				defaultTabId={ sidebars.document }
+				ref={ tabListRef }
+				closeButtonLabel={ __( 'Close Settings' ) }
```
jeryj, line 141: The previous label included the shortcut was `Close Settings. ⇧⌘,`

packages/editor/src/components/sidebar/index.js
```diff
@@ -107,22 +103,43 @@ const SidebarContent = ( {
 			toggleShortcut={ keyboardShortcut }
 			icon={ isRTL() ? drawerLeft : drawerRight }
 			isActiveByDefault={ SIDEBAR_ACTIVE_BY_DEFAULT }
+			header={ false }
```
jeryj, line 106: If `header` is `false`, do we still need the `headerClassName`?

packages/block-editor/src/components/inserter/menu.js
jeryj: This ends up compiling to `block-editor-tabbed-sidebar block-editor__tabbed-sidebar` which feels like a workaround. How about one of these options:

- Give it a name that describes the 350 width, like: `block-editor-tabbed-sidebar--wide`
- Name it as the inserter and list view, i.e. `block-editor-inserter__tabbed-inserter`
- Use the className above it to nest the CSS.

packages/editor/src/components/list-view-sidebar/style.scss
```diff
@@ -49,3 +49,8 @@
 	}
 }
 
+.block-editor__tabbed-sidebar {
+	@include break-medium() {
+		width: 350px;
+	}
+}
```
jeryj, line 56: This style also defines the inserter sidebar's width, so it feels like it's blurring the lines of the inserter and list view.

jeryj (commented):
Partial review :) I'll do more tomorrow.

afercia:
Regarding the X Close button order, pleae see my comment at https://github.com/WordPress/gutenberg/issues/59013#issuecomment-2220142840

packages/editor/src/components/sidebar/index.js
jeryj: I don't think we need the closeLabel here either.

packages/editor/src/components/sidebar/index.js
```diff
@@ -176,20 +194,16 @@ const Sidebar = ( { extraPanels, onActionPerformed } ) => {
 	);
 
 	return (
-		<Tabs
+		<SidebarContent
+			tabName={ tabName }
```
jeryj, line 198: I think the tab-related data here should get moved within the SidebarContent rather than passed as props. I think it was done here previously because the Tabs needed be wrapping the SidebarContent. Now that we have a different structure, I think we could move it within.

packages/interface/src/components/complementary-area/index.js
jeryj: I don't see a better way to do this either.

jeryj:
The current mobile view has tabs that are quite wide. On `trunk` they are much smaller.

Mobile view of tabbed sidebar showing wide tabs

PR #63185, opened 2024-07-05
Block Bindings: Unify `getValue`/`getValues` and `setValue`/`setValues` APIs

SantosGuillamot:
## What?
Initially discussed [here](https://github.com/WordPress/gutenberg/pull/60721#issuecomment-2111810386).

I'm opening this pull request to discuss some of the editor APIs that should be exposed by block bindings and how to make them more coherent. Right now, we have:

* `getValue`: To get the value of each connected attribute one by one. 
* `setValue`: To update the value of each connected attribute one by one.
* `setValues`: To update the values of all the attributes connected to a specific source in batch.

In [the pull request to use block bindings in pattern overrides](https://github.com/WordPress/gutenberg/pull/60721), `setValues` function was introduced because updating in batch was needed for this particular use case.

This is a valid use case not only for pattern overrides, but also for other sources that might want to call an API only once to update all the attributes connected to it. And this not only happens for updating the attributes (`setValues`), but to get the value as well (`getValues`). For example, in post meta we are using `getValue` and `setValue`, which means we are calling the REST API for each attribute connected to post meta.

I can think of two main options:

### ✅ (Selected option) Option 1: Have ONLY `getValues` and `setValues`.
_Covered in this pull request._

Basically, this would require source developers to update the values in batch. They will receive an object with all of the block attributes that are bound to that specific source, and it's up to them to decide what to do.

The developer experience might be slightly more complicated, and maybe it is a bit harder conceptually.

You can see how each source would use the different possibilities in the alternative pull request:
* Post meta `getValue` vs `getValues`: [link](https://github.com/WordPress/gutenberg/blob/5c3221f46bfaed90bd79c5c050a8d8870a0a6bf8/packages/editor/src/bindings/post-meta.js#L20-L44).
* Post meta `setValue` vs `setValues`: [link](https://github.com/WordPress/gutenberg/blob/5c3221f46bfaed90bd79c5c050a8d8870a0a6bf8/packages/editor/src/bindings/post-meta.js#L47-L66).
* Pattern overrides `getValue` vs`getValues`: [link](https://github.com/WordPress/gutenberg/blob/5c3221f46bfaed90bd79c5c050a8d8870a0a6bf8/packages/editor/src/bindings/pattern-overrides.js#L12-L54).

You can check the differences in [pattern overrides](https://github.com/WordPress/gutenberg/compare/try/block-bindings-unify-get-set-calls-apis?expand=1#diff-deffea7af4722e1274cca18247b89c007e041054cc2fa315c68a0bb1425d9eb8R12-R30) and [post meta](https://github.com/WordPress/gutenberg/compare/try/block-bindings-unify-get-set-calls-apis?expand=1#diff-cf4447b4134dbf24fa2819703999f0131449fb79a0611bd544c5e05fd80aa36bL18-R32) functions.

### 🔴 (Discarded option) Option 2: Have `getValue`, `getValuesInBatch`, `setValue`, and `setValuesInBatch`.
_Covered in [this other pull request](https://github.com/WordPress/gutenberg/pull/63186)_

Another option would be to allow both options and let developers decide which one to use. This will give more flexibility, but at the same time, it implies having two functions for a really similar purpose, which could also be confusing. Additionally, it complicates the core logic, and it will probably require more maintenance.

---

**I would love to know more opinions on which approach you would take or other alternatives I didn't consider.**

## Why?
As an effort to polish the existing block bindings APIs in order to make them public, I believe it makes sense to revisit the this part and make it coherent.

## Testing Instructions

Automated tests should pass. Block bindings and pattern overrides should keep working the same way.

### DIFF
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -129,7 +129,7 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 				const { source: sourceName, args: sourceArgs } = binding;
 				const source = sources[ sourceName ];
 				if (
-					( ! source?.getValue && ! source?.getValuesInBatch ) ||
+					! source?.getValues ||
 					! canBindAttribute( name, attributeName )
 				) {
 					continue;
@@ -146,39 +146,17 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 			if ( bindingsBySource.size ) {
 				for ( const [ source, sourceBindings ] of bindingsBySource ) {
 					// Get values in batch if the source supports it.
-					if ( source.getValuesInBatch ) {
-						const values = source.getValuesInBatch( {
-							registry,
-							context,
-							clientId,
-							sourceBindings,
-						} );
-						for ( const [ attributeName, value ] of Object.entries(
-							values
-						) ) {
-							attributes[ attributeName ] = value;
-						}
-					} else {
-						for ( const [
-							attributeName,
-							{ args },
-						] of Object.entries( sourceBindings ) ) {
-							attributes[ attributeName ] = source.getValue( {
-								registry,
-								context,
-								clientId,
-								attributeName,
-								args,
-							} );
-						}
-					}
-
-					// Add placeholders when value is undefined.
-					// TODO: Revisit this part.
-					for ( const [ attributeName, { args } ] of Object.entries(
-						sourceBindings
+					const values = source.getValues( {
+						registry,
+						context,
+						clientId,
+						sourceBindings,
+					} );
+					for ( const [ attributeName, value ] of Object.entries(
+						values
 					) ) {
-						if ( attributes[ attributeName ] === undefined ) {
+						// Use placeholder when value is undefined.
+						if ( value === undefined ) {
 							if ( attributeName === 'url' ) {
 								attributes[ attributeName ] = null;
 							} else {
@@ -188,9 +166,12 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 										context,
 										clientId,
 										attributeName,
-										args,
+										args: sourceBindings[ attributeName ]
+											.args,
 									} );
 							}
+						} else {
+							attributes[ attributeName ] = value;
 						}
 					}
 				}
@@ -225,10 +206,7 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 
 						const binding = bindings[ attributeName ];
 						const source = sources[ binding?.source ];
-						if (
-							! source?.setValue &&
-							! source?.setValuesInBatch
-						) {
+						if ( ! source?.setValues ) {
 							continue;
 						}
 						bindingsBySource.set( source, {
@@ -246,29 +224,12 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 							source,
 							sourceBindings,
 						] of bindingsBySource ) {
-							// Set values in batch if the source supports it.
-							if ( source.setValuesInBatch ) {
-								source.setValuesInBatch( {
-									registry,
-									context,
-									clientId,
-									sourceBindings,
-								} );
-							} else {
-								for ( const [
-									attributeName,
-									{ args, newValue: value },
-								] of Object.entries( sourceBindings ) ) {
-									source.setValue( {
-										registry,
-										context,
-										clientId,
-										attributeName,
-										args,
-										value,
-									} );
-								}
-							}
+							source.setValues( {
+								registry,
+								context,
+								clientId,
+								sourceBindings,
+							} );
 						}
 					}
 
--- a/packages/block-editor/src/hooks/use-bindings-attributes.js
+++ b/packages/block-editor/src/hooks/use-bindings-attributes.js
@@ -121,33 +121,77 @@ export const withBlockBindingSupport = createHigherOrderComponent(
 
 			const attributes = {};
 
-			for ( const [ attributeName, boundAttribute ] of Object.entries(
+			const bindingsBySource = new Map();
+
+			for ( const [ attributeName, binding ] of Object.entries(
 				bindings
 			) ) {
-				const source = sources[ boundAttribute.source ];
+				const { source: sourceName, args: sourceArgs } = binding;
+				const source = sources[ sourceName ];
 				if (
-					! source?.getValue ||
+					( ! source?.getValue && ! source?.getValuesInBatch ) ||
 					! canBindAttribute( name, attributeName )
 				) {
 					continue;
 				}
 
-				const args = {
-					registry,
-					context,
-					clientId,
-					attributeName,
-					args: boundAttribute.args,
-				};
+				bindingsBySource.set( source, {
+					...bindingsBySource.get( source ),
+					[ attributeName ]: {
+						args: sourceArgs,
+					},
+				} );
+			}
 
-				attributes[ attributeName ] = source.getValue( args );
-
-				if ( attributes[ attributeName ] === undefined ) {
-					if ( attributeName === 'url' ) {
-						attributes[ attributeName ] = null;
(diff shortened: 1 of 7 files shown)


gziolo:
Can you confirm the server-side counterpart for completeness? I believe it is `getValueCallback` which takes three params:
- `args` - arguments saved with the binding
- `blockInstance` - the `WP_Block` instance for the currently parsed block
- `attributeName` - the name of the currently edited attribute

It would be valuable to discuss all of these APIs together. I'm also thinking how much this should mirror the known patterns we have for updating and reading block attributes as props passed to React components and through the block editor store.

packages/block-editor/src/hooks/use-bindings-attributes.js
```diff
@@ -121,33 +121,58 @@ export const withBlockBindingSupport = createHigherOrderComponent(
...
-							source.getPlaceholder?.( args );
+			if ( bindingsBySource.size ) {
+				for ( const [ source, sourceBindings ] of bindingsBySource ) {
+					// Get values in batch if the source supports it.
+					const values = source.getValues( {
+						registry,
+						context,
+						clientId,
+						sourceBindings,
+					} );
```
artemiomorales, line 154: I find the `sourceBindings` name here unclear, especially because we'd be expecting folks to destructure and use it when registering their sources. Maybe we could do `bindingsObject` or `bindingsCollection` instead, something along those lines? I know naming things is hard.

The same goes for other places in this PR where `sourceBindings` is used.

artemiomorales (commented):
I think Option 1 is better. It's slightly more code to write, and will perhaps be harder to configure as a result, but conceptually it makes sense to me for the source to receive the binding data and perform the necessary logic.

Below is an initial thought. I'll keep looking at this PR as discussion evolves and it or the alternative PR becomes ready for review 😄

cbravobernal:
My opinion is that less functions, less maintenance, less learning is always better. A get/set values with a batch of just one item would do the same work than a get/set values for several ones.

packages/block-editor/src/hooks/use-bindings-attributes.js
SantosGuillamot: What about using just `bindings`? Well, not sure. I'm fine with whatever you consider better 🙂

SantosGuillamot:
Let's go with **Option 1: Have ONLY getValues and setValues**. Based on the feedback, it seems the preferred option and I agree it makes more sense. We can always add individual `getValue/setValue` in the future.

After taking a look at how attributes are handled, there are two similar use cases:
* `attributes` and `setAttributes` props from the block.
* `updateBlockAttributes` and `getBlockAttributes` as functions. `getBlockAttribute` (singular) exists, but it seems it is only used internally: [link](https://github.com/WordPress/gutenberg/blob/b6ad4cdfed7beb65a22e759f3be9ce040ab19c9c/packages/blocks/src/api/parser/get-block-attributes.js#L108).

It's true that those use cases are slightly different because the implementation is managed by Gutenberg, and in this case we will ask external developers to define `getValues` and `setValues`, but I believe it makes sense to keep consistency.

With the implementation of this pull request we would have:

### getValues - [Link](https://github.com/WordPress/gutenberg/pull/63185/files#diff-cf4447b4134dbf24fa2819703999f0131449fb79a0611bd544c5e05fd80aa36bR18-R33)
**Params - [Link](https://github.com/WordPress/gutenberg/pull/63185/files#diff-99d79bdba99fe4517413b0fd9503eef924e439c03d44504c1ad0606b727cebb6R150-R153)**
It receives:
* registry
* clientId
* context
* bindings: this is an object with all the block attributes that are connected to that specific source and the arguments used. Something like this:

```js
{ 
   url: {
        args: {
            key: 'url_field'
        }
    },
    title: {
        args: {
            key: 'text_field'
        }
    }
}
```
Normally, we would have `name` and `args`, but `name` doesn't make sense here. We could get rid of the `args` level, but I'm not sure if we will have more properties in the future. For example, I'm thinking if we want to pass the `currentValue` at some point, similar to the value we are passing in `setValues`: [link](https://github.com/WordPress/gutenberg/pull/63185/files#diff-99d79bdba99fe4517413b0fd9503eef924e439c03d44504c1ad0606b727cebb6R215-R216).

**Return**
It must return an object with the attributes and the bindings values. Something like:

```js
{
    url: 'https://wp.org',
    title: 'Custom field value',
}
```

### setValues - [Link](https://github.com/WordPress/gutenberg/pull/63185/files#diff-cf4447b4134dbf24fa2819703999f0131449fb79a0611bd544c5e05fd80aa36bR34-R44)

**Params - [Link](https://github.com/WordPress/gutenberg/pull/63185/files#diff-99d79bdba99fe4517413b0fd9503eef924e439c03d44504c1ad0606b727cebb6R228-R231)**

It receives the same params as `getValues` with a small difference. The `bindings` contain the `newValue` pass to the `setAttributes` function.

```js
{ 
   url: {
        args: {
            key: 'url_field'
        },
       newValue: 'https://wp.org/2'
    },
    title: {
        args: {
            key: 'text_field'
        },
       newValue: 'update custom field value'
    }
}
```

**Return**
It doesn't return anything. It just do whatever it wants. In the case of post meta, update the post entity.

---

> Can you confirm the server-side counterpart for completeness?

Regarding the server-side counterpart, you're right we only have `get_value_callback` right now with `source_args`, block instance, and attribute names as params: [link](https://github.com/WordPress/wordpress-develop/blob/cf8e9ce375564f92efccfa02935c463b91a47cb4/src/wp-includes/block-bindings.php#L79-L90).

I'd like to explore having a `get_values_callback` in the server as well, I remember people asking for getting values in batch in the server as well. Although I believe it could be explored in a follow-up pull request? If we want to remove the existing `get_value_callback` we will need some kind of deprecation.

---
Apart from that, I plan to work on improving the code docs and the types to ensure it is clear how to use these functions before making them public.

packages/block-editor/src/hooks/use-bindings-attributes.js
artemiomorales: Just `bindings` would be great! But it's already being used in the upper scope on line 104.

I think it's probably worth renaming the value in the upper scope to something like `bindingsData`, that way we can use `bindings` in the public-facing API.

SantosGuillamot: I've changed it in [this commit](https://github.com/WordPress/gutenberg/pull/63185/commits/e244e702ef11e6a1b2b1c40b813a5f71ddd97414). I finally used `blockBindings` for the top variable because they are actually the bindings for the block that is being processed.

Let me know what you think. It's fairly simple to change it to a preferred name.

artemiomorales (approved):

packages/block-editor/src/hooks/use-bindings-attributes.js
artemiomorales: Looks good to me 🙌

packages/block-editor/src/hooks/use-bindings-attributes.js
```diff
@@ -115,110 +115,117 @@ export const withBlockBindingSupport = createHigherOrderComponent(
...
-			if ( ! bindings ) {
+			if ( ! blockBindings ) {
 				return;
 			}
 
 			const attributes = {};
 
-			for ( const [ attributeName, boundAttribute ] of Object.entries(
-				bindings
+			const bindingsBySource = new Map();
```
cbravobernal: Should this be `blockBindingsBySource` to keep consistency?

cbravobernal:
We may need to take a look at how to make this API public without asking the developer to pass `registry` and `clientId`.

Good work @SantosGuillamot ! Feel free to merge when you feel is ok.

packages/block-editor/src/hooks/use-bindings-attributes.js
SantosGuillamot: I have just changed this 🙂

PR #63187, opened 2024-07-05
DataViews's filterSortAndPaginate utility: support sorting by number

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/55083

## What?

Adds support for sorting by number to the `filterSortAndPaginate` DataViews utility.

## Why?

DataViews shouldn't be limited to sort by strings.

## How?

Implement sorting by number.

## Testing Instructions

Verify tests pass.

### DIFF
--- a/packages/dataviews/src/filter-and-sort-data-view.ts
+++ b/packages/dataviews/src/filter-and-sort-data-view.ts
@@ -142,6 +142,16 @@ export function filterSortAndPaginate< Item >(
 			filteredData.sort( ( a, b ) => {
 				const valueA = fieldToSort.getValue( { item: a } ) ?? '';
 				const valueB = fieldToSort.getValue( { item: b } ) ?? '';
+
+				if (
+					typeof valueA === 'number' &&
+					typeof valueB === 'number'
+				) {
+					return view.sort?.direction === 'asc'
+						? valueA - valueB
+						: valueB - valueA;
+				}
+
 				return view.sort?.direction === 'asc'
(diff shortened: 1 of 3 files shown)


oandregal:
Sorting is another case where having the `field.type` would help.

packages/dataviews/src/filter-and-sort-data-view.ts
```diff
@@ -61,7 +61,7 @@ export function filterSortAndPaginate< Item >(
 		} );
 	}
 
-	if ( view.filters.length > 0 ) {
+	if ( view.filters?.length > 0 ) {
```
oandregal, line 64: We shouldn't require filters presence to be able to process data.

ntsekouras (approved):
Looks good, thanks for also updating the storybook.

packages/dataviews/src/filter-and-sort-data-view.ts
```diff
@@ -142,6 +142,16 @@ export function filterSortAndPaginate< Item >(
 			filteredData.sort( ( a, b ) => {
 				const valueA = fieldToSort.getValue( { item: a } ) ?? '';
 				const valueB = fieldToSort.getValue( { item: b } ) ?? '';
+
+				if (
+					typeof valueA === 'number' &&
+					typeof valueB === 'number'
```
youknowriad, line 148: This ideally should come from the "field" type and not by guessing the value type. Any other words, the "field types" should define a sorting function. But this is more down the road. This is ok for now.

Issue #63146, opened 2024-07-04
Component Optimization: Pattern overrides hook


cbravobernal:
## What problem does this address?

Right now, the code inside `ControlsWithStoreSubscription` function is executed on each block change (including writing a post, each keystroke or hovering a block) regardless the `Override block` is present or not (it appears only in patterns as far as I know)
https://github.com/WordPress/gutenberg/blob/e39ae77a9ea59c3f41ee63f1f1f1de31209519d0/packages/editor/src/hooks/pattern-overrides.js#L95

https://github.com/WordPress/gutenberg/assets/37012961/84a885da-b8c8-4a6d-9bf1-24d2233c0777

## How to test

- Add a console log here: https://github.com/WordPress/gutenberg/blob/e39ae77a9ea59c3f41ee63f1f1f1de31209519d0/packages/editor/src/hooks/pattern-overrides.js#L95
- Write a post with the browser console opened, check that the log appears tons of times.

ping @Mamaduka

Mamaduka:
Generally, I think it's okay if a component rerenders like this unless it's costly to render and blocks the UI, which doesn't seem to be the case here.

As @ellatrix suggested in the Slack thread, we can optimize this using `React.memo` and only pass the required props. The props needed for the Control and its child components are `clientId`, `name`, `metadata` and `setAttributes`.

One issue with memoization is that it's easy to break. Let's say in the future someone updates the code and passes the whole `attributes` object as props. It will set us back to where we started with + a small cost of component memoization. But we'll have the same issue in the `block-editor` package, so maybe that's okay.

I created a draft PR just in case - #63189.

@kevin940726, @talldan, what do you think?

kevin940726:
Are the re-renders costly here? Are we fixing a performance bottleneck or doing premature optimizations? Could this be solved if we leverage the React Compiler in the future?

> we can optimize this using `React.memo` and only pass the required props. 

As pointed out, memoing also comes with a cost. We can audit it to see the trade-offs and whether it's worth it.

I'm not opposed to improving it in any way, but I think we should be careful of optimizing based on our assumptions without data support. I appreciate the draft PR explorations a lot! Thanks! ❤️

cbravobernal:
The component has not been moved. Re-renders are not costly so we don't need any development here.
Thanks @Mamaduka for your contribution!

PR #63189, opened 2024-07-05
Pattern Overrides: Memoize controls component

Mamaduka:
## What?
See #63146.

### DIFF
--- a/packages/editor/src/hooks/pattern-overrides.js
+++ b/packages/editor/src/hooks/pattern-overrides.js
@@ -6,6 +6,7 @@ import { privateApis as patternsPrivateApis } from '@wordpress/patterns';
 import { createHigherOrderComponent } from '@wordpress/compose';
 import { useBlockEditingMode } from '@wordpress/block-editor';
 import { useSelect } from '@wordpress/data';
+import { memo } from '@wordpress/element';
 import { store as blocksStore } from '@wordpress/blocks';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/editor/src/hooks/pattern-overrides.js
```diff
@@ -41,9 +104,14 @@ const withPatternOverrideControls = createHigherOrderComponent(
 
 		return (
 			<>
-				<BlockEdit { ...props } />
+				<BlockEdit key="edit" { ...props } />
 				{ props.isSelected && isSupportedBlock && (
-					<ControlsWithStoreSubscription { ...props } />
+					<MemoizedControlsWithStoreSubscription
```
ellatrix, line 109: Can't we use `createBlockEditFilter` like the other hooks?

Mamaduka: This package is different, but I can copy the same logic here. This is still draft PR :)

ellatrix: Oh, maybe we could add a private API for now?

Mamaduka: This is only the `editor.BlockEdit` filter in the editor package. Does it make sense to abstract it now?

Mamaduka:
I missed a couple of updates in the child components. This is the reason for the e2e test failures.

I also noticed that the `PatternOverridesControls` component uses other attributes when the block is an Image.

https://github.com/WordPress/gutenberg/blob/6034e900bcb52ca6de20e298d2180977eeedcad5/packages/patterns/src/components/pattern-overrides-controls.js#L78-L80

packages/editor/src/hooks/pattern-overrides.js
ellatrix: I think it does at some point anyway, we need to start thinking about the block supports API.

Mamaduka: In my case, to actually start building it 😅 Hopefully, I can share something before 6.7 betas.

cbravobernal: Block bindings panels should use it too, in case it helps.

Mamaduka: @cbravobernal, which component exactly are we talking about?

packages/editor/src/hooks/pattern-overrides.js
```diff
@@ -41,9 +104,19 @@ const withPatternOverrideControls = createHigherOrderComponent(
...
+						clientId={ props.clientId }
+						name={ props.name }
+						metadata={ props.attributes.metadata }
+						setAttributes={ props.setAttributes }
+						hasUnsupportedAttributes={
+							props.name === 'core/image' &&
+							( !! props.attributes.caption?.length ||
+								!! props.attributes.href?.length )
+						}
```
Mamaduka, line 118: I had to lift this condition to avoid passing the whole attribute object.

packages/editor/src/hooks/pattern-overrides.js
cbravobernal: I'm still working on it, and will move it from `block-editor` to `editor` package. But is this one:
https://github.com/WordPress/gutenberg/pull/62880

Mamaduka:
@cbravobernal, the UI implementation from #62880 will stay in the `block-editor` packages. Is that correct?

In that case, as @kevin940726 noted, there's probably no need for optimization here - https://github.com/WordPress/gutenberg/issues/63146#issuecomment-2242353890.

cbravobernal:
Correct, we can close this draft PR and its issue.

PR #60466, opened 2024-04-04
Site Editor: create router adapter for sidebar

jsnajdr:
First working version of a consolidated router for Site Editor sidebar.

The `NavigatorProvider` in `Sidebar` no longer has its own in-memory router synchronized with the browser history with `useSyncPathWithURL`.

Instead, we create a "router adapter" around the native browser history, and pass it as a `router` prop to `Sidebar` and in turn to `NavigatorProvider`. The `router` value is identical to the context value that is then passed down in `NavigatorContext`.

This router adapter currently duplicates a lot of code from the `NavigatorProvider`. The reducer to register screens, the code for matching paths and finding parent paths. After we add all the missing features to `goTo` and `goToParent` (focus management etc.), there will be more duplication. We'll have to solve that somehow.

Currently focus management (focusing after navigation and after going back) is not implemented, I'll work on that further.

The sole source of truth for the current location and its history is the global browser history. The router adapter doesn't store its own history state. It merely maps the global location params to the `Sidebar` URL scheme. And then maps it in the opposite direction in methods like `goTo`, which maps the path and then acts on the browser history directly.

PR #63194, opened 2024-07-05
Site Editor Sidebar: Hide horizontal scrollbar when navigating

### LINKED ISSUE #63192: Site Editor Sidebar: horizontal scrollbar appears when navigating
### Description

When accessing any screen from the site editor top page (Design), a scroll bar briefly appears below the sidebar.

This issue is most noticeable on Windows OS and Chrome browsers, and this issue is new to WP6.6.

This issue should also be reproducible on MacOS, where you need to change the scrollbar settings to always be visible.

https://github.com/WordPress/gutenberg/assets/54422211/d368630b-a66f-477e-a2a6-960a191ec59b

### Environment info

_No response_

Yes

Yes

t-hamano:
Fixes #63192

## What?
This PR hides the horizontal scrollbar that appears when navigating in the site editor sidebar.

scroll-bar

## How?
To be honest, I didn't know the root cause. Maybe it's related to animations when navigating.

The only solution I found was to apply `overflow-x: hidden;` to the sidebar content.

## Testing Instructions

- If you're using MacOS, make sure the scrollbar is always visible.
- Navigate to any screen in the site editor.

### Before

https://github.com/WordPress/gutenberg/assets/54422211/ab46d7c2-c3d3-4e23-b558-5117d00c42ab

### After

https://github.com/WordPress/gutenberg/assets/54422211/371044ac-3974-4d24-a170-14386c28e1e6

### DIFF
--- a/packages/edit-site/src/components/sidebar/style.scss
+++ b/packages/edit-site/src/components/sidebar/style.scss
@@ -1,7 +1,12 @@
 .edit-site-sidebar__content {
 	flex-grow: 1;
-	overflow-x: hidden;
 	overflow-y: auto;
+	// Prevents horizontal overflow while animating screen transitions
+	overflow-x: hidden;
+	// Mark this section of the DOM as isolated, providing performance benefits
+	// by limiting calculations of layout, style and paint to a DOM subtree rather
+	// than the entire page.
+	contain: content;
 }
 
 @keyframes local--slide-from-right {
--- a/packages/edit-site/src/components/sidebar/style.scss
+++ b/packages/edit-site/src/components/sidebar/style.scss
@@ -1,5 +1,6 @@
 .edit-site-sidebar__content {
 	flex-grow: 1;
+	overflow-x: hidden;
 	overflow-y: auto;
 }
 


t-hamano:
This issue is very noticeable on Windows OS and Chrome. Hopefully it will be backported to WP6.6 RC3.

jsnajdr (approved):
Nice catch! It's a regression I introduced in #60466. Before that, the `overflow-x` style would come from Emotion:

Screenshot 2024-07-08 at 13 13 34

These Emotion styles are defined as `styles.navigatorProviderWrapper` here: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/navigator/styles.ts#L6-L14

But after my PR the `NavigatorProvider` is no longer used, so these CSS properties are missing.

To fully fix the regression, you should also add the `contain: content` property.

t-hamano:
Thanks for the review!

> To fully fix the regression, you should also add the `contain: content` property.

I added `contain: content`. Also, because I felt that the intent of each CSS was hard to understand at a glance, I also ported the comments.

jsnajdr:
Thanks! Now the fix should be complete.

PR #63197, opened 2024-07-05
Fix unlabeled Remove shadow buttons.

### LINKED ISSUE #63196: Remove shadow buttons are unlabeled and don't show a tooltip
### Description

The 'Remove shadow' icon buttons in the Global Styles are unlabeled. 

More accurately, propr to the button [are passed via an object](https://github.com/WordPress/gutenberg/blob/6034e900bcb52ca6de20e298d2180977eeedcad5/packages/edit-site/src/components/global-styles/shadows-edit-panel.js#L368-L376), where an `ariaLabel` and a `tooltip` keys are set. 
However, the `Button` component either accepts a `label` prop or  an `aria-label` attribute.
Not sure why the `tooltip` prop passed this way doesn't work btut, regardless, it's unnecessary if the label prop is set correctly.

The rendered markup is invalid as the key names are rendered as HTML attributes:
- the button element has an `arialabel` attribute, it should be `aria-label`
- the button element has an `tooltip` attribute

Screenshot of the button markup in the dev tools inspector and the accessibility tab selected to show the button doesn't have an accessible name:

Screenshot 2024-07-05 at 16 07 26

I'm not sure why those props are passed via an object in the first place, instead of being passed individually to the Button component. As far as I can tell there's no need to set a variable for that object in the first place.

I have to note this is one more occurrence of an unlabeled control in the editor. There have been dozens of cases because contributors to this project keep 'doing it wrong' and the base component are open to misuesd.

As pointed out in several other issues and conversations, unlabeled controls are not acceptable. The only way to make sure this doesn't happen again is by refactoring the base components and make the `label` prop required. See https://github.com/WordPress/gutenberg/issues/51740

- Go to the Site Editor > Styles > Shadows > click one of the default shadows > click the plus icon button 'Add shadow'.
- A new shadow item will appear in the list below, with a minus icon button. Screenshot:

Screenshot 2024-07-05 at 16 05 19

- Hover or focus the minus icon button.
- Observe no tooltip appears.
- Inspect the source and observe the button has an `arialabel` attribute instead of `aria-label`. 
- Observe there's also a `tooltip` attribute.

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
FIxes https://github.com/WordPress/gutenberg/issues/63196

## What?

The 'Remove shadow' buttons are unlabeled.
They don't show a tooltip as well.

## Why?

All interactive controls _must_ be labeled.
Icon buttons _must_ visually expose their accessible name via a tooltiip.

## How?

Fixes the `label` prop. 
Removes incorrect `tooltip` prop.

## Testing Instructions

- Repeat the test instructions from https://github.com/WordPress/gutenberg/issues/63196
- Hover or focus the Remove shadow button and observe it does show a tooltip.
- Inspect the source and observe the button markup does not have an `arialabel` or `tooltip` HTML attributes.
- Instead, observe the button markup does have an `aria-label` attribute.

### DIFF
--- a/packages/edit-site/src/components/global-styles/shadows-edit-panel.js
+++ b/packages/edit-site/src/components/global-styles/shadows-edit-panel.js
@@ -371,8 +371,7 @@ function ShadowItem( { shadow, onChange, canRemove, onRemove } ) {
 						'edit-site-global-styles__shadow-editor__remove-button',
 						{ 'is-open': isOpen }
 					),
-					ariaLabel: __( 'Remove shadow' ),
-					tooltip: __( 'Remove shadow' ),
+					label: __( 'Remove shadow' ),
 				};
 
 				return (


t-hamano (approved):
LGTM!

carolinan (approved):

talldan:
I've added the label for this to be backported for 6.6, as I believe this was introduced in 6.6, and is currently an issue in the RC.

PR #63198, opened 2024-07-05
RangeControl: Fix RTL support for custom marks

tyxla:
## What?
Fix the RTL support for `RangeControl` with custom marks.

Spotted as I was playing with RTL in Storybook while working on #63129.

## Why?
Currently on RTL, the custom marks are off for RTL browsers.

## How?
We're essentially applying some RTL considerations to the custom mark styles 

## Testing Instructions
* Open the `RangeControl` component in storybook
* Try all stories with marks and verify they work well on both LTR and RTL.

None

## Screenshots or screencast 
The "With Decimal Step And Marks" story on RTL:

Before:

Screenshot 2024-07-05 at 17 38 11

After:

Screenshot 2024-07-05 at 17 37 55


mirka (approved):
🚀

Issue #63193, opened 2024-07-05
Move bulk actions menu to the Footer, consolidate with floating toolbar and total items display


jameskoster:
<img width="2656" alt="bulk-edit" src="https://github.com/WordPress/gutenberg/assets/846565/bd76cf8b-7e5d-46dd-b29b-ddae1050dbee">

The mockup above consolidates the bulk action UX in a single UI. In the page footer it's possible to select/deselect all visible items and perform bulk actions like trashing.

jorgefilipecosta:
Hi @WordPress/gutenberg-design, is this issue ready to be developed following the mockup above?

jameskoster:
@jorgefilipecosta barring any minor tweaks, I think so yes :)

jorgefilipecosta:
Hi @jameskoster, what is the plan regarding bulk actions like export as json, or permanently delete that currently don't have icons and are executed on "Bulk edit" menu? Should we add icons to them and show them in the footer like the others?

jameskoster:
Yes, to begin with we can try adding icon buttons for all available actions.

PR #63203, opened 2024-07-05
Move filter UI into a toggle-able panel to improve experience on narrow viewports/containers

### LINKED ISSUE #60696: Move filter UI into a toggle-able panel to improve experience on narrow viewports/containers
Filter panel

Instead of permanently displaying filter controls and selections, we can place them in a toggle-able panel. This will allow users to selectively tidy up the interface which will prove particularly useful on narrow viewports (mobile) and containers (list layout).

We can be smart about when to show/hide the panel:

* Hidden by default across all layouts when no filters are applied
* Visible on load when the container is larger than X (Table / Grid layout), _and_ filters are applied
* Hidden on load when the container is smaller than X (List layout), even if filters are applied

<details><summary>Original issue</summary>In https://github.com/WordPress/gutenberg/issues/59696 the display of individual chips can be finessed. 

But we still need to figure out the display of many filters in narrow containers / viewports. It's not a huge problem right now because the existing data views only have <=2 filters. 

However, things like Posts or Media (https://github.com/WordPress/gutenberg/issues/55238) can have many filter options, and this experience isn't ideal:

Screenshot 2024-04-12 at 10 12 11

Let's explore how to better display many chips in the filter UI.</details>

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60696

Adds a button to toggle the filter visibility. The button also shows the number of filters applied on the corner.

## Screenshot
Screenshot 2024-07-05 at 17 54 22
Screenshot 2024-07-05 at 17 54 12

## Testing Instructions
I verified a button to toggle the filter visibility was available and worked as expected.
I verified that the count shown on the button matched the number of filters enabled and if there are no filters, the count is not shown.

### DIFF
--- a/packages/dataviews/src/components/dataviews/index.tsx
+++ b/packages/dataviews/src/components/dataviews/index.tsx
@@ -15,7 +15,11 @@ import { useMemo, useState } from '@wordpress/element';
 import { default as DataViewsBulkActions } from '../dataviews-bulk-actions';
 import DataViewsBulkActionsToolbar from '../dataviews-bulk-actions-toolbar';
 import DataViewsContext from '../dataviews-context';
-import DataViewsFilters from '../dataviews-filters';
+import {
+	default as DataViewsFilters,
+	useFilters,
+	FilterVisibilityToggle,
+} from '../dataviews-filters';
 import DataViewsLayout from '../dataviews-layout';
 import DataviewsPagination from '../dataviews-pagination';
 import DataViewsSearch from '../dataviews-search';
@@ -69,6 +73,8 @@ export default function DataViews< Item >( {
 }: DataViewsProps< Item > ) {
 	const [ selectionState, setSelectionState ] = useState< string[] >( [] );
 	const [ density, setDensity ] = useState< number >( 0 );
+	const [ isShowingFilter, setIsShowingFilter ] =
+		useState< boolean >( false );
 	const isUncontrolled =
 		selectionProperty === undefined || onChangeSelection === undefined;
 	const selection = isUncontrolled ? selectionState : selectionProperty;
@@ -90,6 +96,7 @@ export default function DataViews< Item >( {
 		);
 	}, [ selection, data, getItemId ] );
 
+	const filters = useFilters( _fields, view );
 	return (
 		<DataViewsContext.Provider
 			value={ {
@@ -113,14 +120,18 @@ export default function DataViews< Item >( {
 					alignment="top"
 					justify="start"
 					className="dataviews__view-actions"
+					spacing={ 1 }
 				>
-					<HStack
-						justify="start"
-						className="dataviews-filters__container"
-						wrap
-					>
+					<HStack justify="start" wrap>
 						{ search && <DataViewsSearch label={ searchLabel } /> }
-						<DataViewsFilters />
+						<FilterVisibilityToggle
+							filters={ filters }
+							view={ view }
+							onChangeView={ onChangeView }
+							setOpenedFilter={ setOpenedFilter }
+							setIsShowingFilter={ setIsShowingFilter }
+							isShowingFilter={ isShowingFilter }
+						/>
 					</HStack>
 					{ view.type === LAYOUT_GRID && (
 						<DensityPicker
@@ -140,6 +151,7 @@ export default function DataViews< Item >( {
 						{ header }
 					</HStack>
 				</HStack>
+				{ isShowingFilter && <DataViewsFilters /> }
 				<DataViewsLayout />
 				<DataviewsPagination />
 				<DataViewsBulkActionsToolbar />
--- a/packages/dataviews/src/components/dataviews-filters/add-filter.tsx
+++ b/packages/dataviews/src/components/dataviews-filters/add-filter.tsx
@@ -32,29 +32,18 @@ interface AddFilterProps {
 	setOpenedFilter: ( filter: string | null ) => void;
 }
 
-function AddFilter(
-	{ filters, view, onChangeView, setOpenedFilter }: AddFilterProps,
-	ref: Ref< HTMLButtonElement >
-) {
-	if ( ! filters.length || filters.every( ( { isPrimary } ) => isPrimary ) ) {
-		return null;
-	}
+export function AddFilterDropdownMenu( {
+	filters,
+	view,
+	onChangeView,
+	setOpenedFilter,
+	trigger,
+}: AddFilterProps & {
+	trigger: React.ReactNode;
+} ) {
 	const inactiveFilters = filters.filter( ( filter ) => ! filter.isVisible );
 	return (
-		<DropdownMenu
-			trigger={
-				<Button
-					accessibleWhenDisabled
-					size="compact"
-					className="dataviews-filters__button"
-					variant="tertiary"
-					disabled={ ! inactiveFilters.length }
-					ref={ ref }
-				>
-					{ __( 'Add filter' ) }
-				</Button>
-			}
-		>
+		<DropdownMenu trigger={ trigger }>
 			{ inactiveFilters.map( ( filter ) => {
 				return (
 					<DropdownMenuItem
@@ -82,6 +71,33 @@ function AddFilter(
 				);
 			} ) }
 		</DropdownMenu>
+	);
+}
+
+function AddFilter(
+	{ filters, view, onChangeView, setOpenedFilter }: AddFilterProps,
+	ref: Ref< HTMLButtonElement >
+) {
+	if ( ! filters.length || filters.every( ( { isPrimary } ) => isPrimary ) ) {
+		return null;
+	}
+	const inactiveFilters = filters.filter( ( filter ) => ! filter.isVisible );
+	return (
+		<AddFilterDropdownMenu
+			trigger={
+				<Button
+					accessibleWhenDisabled
+					size="compact"
+					className="dataviews-filters-button"
+					variant="tertiary"
+					disabled={ ! inactiveFilters.length }
+					ref={ ref }
+				>
+					{ __( 'Add filter' ) }
+				</Button>
+			}
+			{ ...{ filters, view, onChangeView, setOpenedFilter } }
+		/>
 	);
 }
 
--- a/packages/dataviews/src/components/dataviews-filters/index.tsx
+++ b/packages/dataviews/src/components/dataviews-filters/index.tsx
@@ -1,65 +1,150 @@
 /**
  * WordPress dependencies
  */
-import { memo, useContext, useRef } from '@wordpress/element';
-import { __experimentalHStack as HStack } from '@wordpress/components';
+import {
+	memo,
+	useContext,
+	useRef,
+	useMemo,
+	useCallback,
+} from '@wordpress/element';
+import { __experimentalHStack as HStack, Button } from '@wordpress/components';
+import { funnel } from '@wordpress/icons';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
 import FilterSummary from './filter-summary';
-import AddFilter from './add-filter';
+import { default as AddFilter, AddFilterDropdownMenu } from './add-filter';
 import ResetFilters from './reset-filters';
 import DataViewsContext from '../dataviews-context';
 import { sanitizeOperators } from '../../utils';
 import { ALL_OPERATORS, OPERATOR_IS, OPERATOR_IS_NOT } from '../../constants';
-import type { NormalizedFilter } from '../../types';
+import type { NormalizedFilter, NormalizedField, View } from '../../types';
+
+export function useFilters( fields: NormalizedField< any >[], view: View ) {
+	return useMemo( () => {
+		const filters: NormalizedFilter[] = [];
+		fields.forEach( ( field ) => {
+			if ( ! field.elements?.length ) {
+				return;
+			}
+
+			const operators = sanitizeOperators( field );
+			if ( operators.length === 0 ) {
+				return;
+			}
+
+			const isPrimary = !! field.filterBy?.isPrimary;
+			filters.push( {
+				field: field.id,
+				name: field.label,
+				elements: field.elements,
+				singleSelection: operators.some( ( op ) =>
+					[ OPERATOR_IS, OPERATOR_IS_NOT ].includes( op )
+				),
+				operators,
+				isVisible:
+					isPrimary ||
+					!! view.filters?.some(
+						( f ) =>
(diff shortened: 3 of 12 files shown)


jasmussen:
The button does suggest a drop-down. I also believe we had a dropdown design for filters somewhere, though there were likely complexity reasons for this. IMO this seems okay to me, as it improves the situation, but I'll defer to Jay or others!

jameskoster:
Thanks @jorgefilipecosta, this looks like a good start.

One quick change would be to ensure the panel is closed by default, rather than open. We might also try opening by default _when filters are applied_, but I don't feel as strongly about that part.

Given the visibility is now toggleable I think we can contain the filter UI within a dedicated row beneath `.dataviews-filters__view-actions` (we should probably rename that class too :D).

That structural change would decrease the likelihood of wrapping on narrow containers. In other words this:

Screenshot 2024-07-12 at 12 13 25


Instead of this:

Screenshot 2024-07-12 at 12 05 11

jorgefilipecosta:
Hi @jameskoster, the changes you suggested were implemented.

ntsekouras:
It feels a bit weird that we have to click the toggle to then be presented with the `add filter` button, no? This is in the cases where no filter is active.

Additionally should the `text|global` search be taken into account the active filters badge?

packages/dataviews/src/dataviews.tsx
```diff
@@ -150,7 +147,33 @@ export default function DataViews< Item >( {
...
+						size="compact"
+						icon={ funnel }
+						label={ __( 'Toggle filter display' ) }
+						onClick={ () =>
+							setIsShowingFilter( ! isShowingFilter )
+						}
+						isPressed={ isShowingFilter }
+						aria-expanded={ isShowingFilter }
+					/>
+					{ !! view.filters?.length && view.filters?.length > 0 && (
```
ntsekouras: Nit: `!! view.filters?.length` should be enough.

ntsekouras:
There is one more nuance with `openedFilter`. This was added initially to in order to open the filter when we add it for the first time. With these changes, if we add a filter and then toggle the filter visibility, it still opens the last inserted filter.

jameskoster:
> It feels a bit weird that we have to click the toggle to then be presented with the add filter button, no? This is in the cases where no filter is active.

Yes, I think that's a good point. When no filters are set the filter button could expose the 'Add filter' menu instead. Here's the flow:

add-filter

Once filters are added it behaves like a toggle for the panel.

jorgefilipecosta:
> There is one more nuance with `openedFilter`. This was added initially to in order to open the filter when we add it for the first time. With these changes, if we add a filter and then toggle the filter visibility, it still opens the last inserted filter.

Hi @ntsekouras, I think this issue is fixed.

> Yes, I think that's a good point. When no filters are set the filter button could expose the 'Add filter' menu instead. Here's the flow:

Hi @jameskoster, the solution you suggested was implemented, let me know if it looks as expected.

(the end to end tests are not updated yet, but if we agree on this solution I will do the update on this PR)

packages/dataviews/src/dataviews.tsx
```diff
@@ -111,7 +119,7 @@ export default function DataViews< Item >( {
 			>
 				<HStack
 					justify="start"
-					className="dataviews-filters__container"
+					className="dataviews-search__container"
```
ntsekouras: I guess we won't need the `hstack` now.. We need to verify though if it has visual side effects..

packages/dataviews/src/filters.tsx
```diff
@@ -22,6 +24,120 @@ interface FiltersProps< Item > {
...
+			<AddFilterDropdownMenu
+				filters={ filters }
+				view={ view }
+				onChangeView={ onChangeViewWithFilterVisibility }
+				setOpenedFilter={ setOpenedFilter }
+				trigger={
+					<Button
+						size="compact"
+						icon={ funnel }
+						label={ __( 'Toggle filter display' ) }
```
ntsekouras: Should this label be `add filter` in this case?

ntsekouras:
This tests well and code wise looks good. Let's see the design input and see to update the tests afterwards.

packages/dataviews/src/filters.tsx
jorgefilipecosta: Yes it should, it was updated 👍

packages/dataviews/src/dataviews.tsx
jorgefilipecosta: Yes, we can remove hstack but I needed to add a small, a small CSS rule to make the search still expand.
The search input also got a little bit bigger but I think we have space so it may be ok to default to the normal input control width (240px). Going to cc: @WordPress/gutenberg-design to check if we can do this small dimension change.

jameskoster:
This is working really well!

A couple of tiny visual tweaks:

* The gap between the Layout and Settings icon buttons is 4px. But the gap between those buttons and the Filter button is 8px. Ideally that gap should be 4px too. 
* The toggle button is spilling out of its container breaking alignment with the primary action button above. `flex-shrink: 0` on `.dataviews-filters-toggle__container` seems to fix.
* On List layout we can remove the `max-width` applied to the search container so that it fills the available space.

jorgefilipecosta:
This PR passed by a big rebase, but it seems now everything is working again.

jorgefilipecosta:
Hi @jameskoster, thank you for the reviews, the first two tweaks you suggested were applied. The third one is being done on this PR https://github.com/WordPress/gutenberg/pull/63755 as it is a change to a behavior already on the trunk.

jameskoster:
Thanks for adding those tweaks :) I'm noticing a small bug; on the Pages table, adding one filter seems to add all filter options:

https://github.com/user-attachments/assets/b1fcf10b-598c-4bed-a230-849eb93b0c49

jorgefilipecosta:
> I'm noticing a small bug; on the Pages table, adding one filter seems to add all filter options:

Hi @jameskoster, thank you for catching this it was an issue introduced during a rebase and is now fixed.

jameskoster (commented):
I pushed a couple of small style tweaks, I think this is ready to go now, but let's see if some more feedback comes in :)

jorgefilipecosta:
Perfect thank you for reviewing this @jameskoster, I will update the end-to-end tests in this PR to take into account this new UI.

youknowriad:
Probably not specific to this branch, but why are we losing filters when switching the layout type (testing on pages dataviews)

youknowriad:
Screenshot 2024-07-23 at 10 03 36 AM

I feel like there's something wrong here personally. Too much buttons and controls.
Also, the quick edit should probably be last.

youknowriad:
I also think it's more logical for the filters toggle to be closer to the "search" input (left of the bar rather than right)

jorgefilipecosta:
> Probably not specific to this branch, but why are we losing filters when switching the layout type (testing on pages dataviews)

Good catch @youknowriad, that's not an issue of this branch. I will try to fix this in a separate PR.

jorgefilipecosta:
So should we try to reorder the items, making bulk edit the last one and the filters the first one (closer to the search)? Would that work @jameskoster, @youknowriad?

On https://github.com/WordPress/gutenberg/pull/63852 I am also moving the item size picker reducing the number of buttons/controls at the top.

jameskoster:
Yeah let's try moving the filter button closer to the search input.

Worth noting that the 'Bulk edit' button will also be moved (https://github.com/WordPress/gutenberg/issues/63193) which will help the the over-crowding.

jorgefilipecosta:
I made the filters button the first one as suggested.

youknowriad:
How do you feel about the button behaving sometimes as a dropdown menu (initial filter) and sometimes as a toggle when a filter is already applied. Should we be consistent and make it always a toggle?

youknowriad:
Also should we move it right after the search input (left area)?

jorgefilipecosta:
> How do you feel about the button behaving sometimes as a dropdown menu (initial filter) and sometimes as a toggle when a filter is already applied. Should we be consistent and make it always a toggle?

Hi @youknowriad, being always a toggle was the initial solution I proposed, but following feedback from @ntsekouras that it felt weird to need to show the filters and then press add filter https://github.com/WordPress/gutenberg/pull/63203#issuecomment-2230922691 we implemented this behavior. I think the behavior works well and is not uncommon, some applications use something similar.

youknowriad:
> Hi @youknowriad, being always a toggle was the initial solution I proposed, but following feedback from @ntsekouras that it felt weird to need to show the filters and then press add filter https://github.com/WordPress/gutenberg/pull/63203#issuecomment-2230922691 we implemented this behavior. I think the behavior works well and is not uncommon, some applications use something similar.

Ok, I guess we can try and see.

jameskoster:
@jorgefilipecosta I reckon we could move it right next to the search input with an 8px gap between:

Screenshot 2024-07-24 at 13 31 09

jorgefilipecosta:
Hi @jameskoster the filter toggle/add button is now at the side of the search as suggested.

jameskoster:
I pushed a tiny style tweak to the count. 

There's only one small issue I see outstanding: The filter panel visibility persists as you navigate between different views. IE if I open the panel on one view, it stays open when I navigate to a different view. Ideally each view treats the open/close state dependently. Would that be possible?

jorgefilipecosta:
Hi @jameskoster thank you for the style enhancement.

> There's only one small issue I see outstanding: The filter panel visibility persists as you navigate between different views. IE if I open the panel on one view, it stays open when I navigate to a different view. Ideally each view treats the open/close state dependently. Would that be possible?

I made an update, and now the state should not persist across the views.

packages/dataviews/src/components/dataviews/index.tsx
```diff
@@ -49,6 +49,8 @@ type DataViewsProps< Item > = {
 	selection?: string[];
 	onChangeSelection?: ( items: string[] ) => void;
 	header?: ReactNode;
+	isShowingFilter?: boolean;
+	setIsShowingFilter?: React.Dispatch< React.SetStateAction< boolean > >;
```
youknowriad: I would love if avoid these props please. I don't think we should be adding random props like that to the DataViews component.

Why do we need them?

jorgefilipecosta: Hi @youknowriad, we added these props to reset the isShowingFilter state when the active view changes (e.g: from drafts, to scheduled). This allows us to control or reset the state from outside of the component. If these props aren't passed, the component will use its own state, but having them allows outside control when necessary.

youknowriad: I see. I think it's not ideal in terms of Api for the DataViews component. Why control this state and not other states like dropdown states... I think this doesn't make for a great API in DataViews.

Two options from me:

 - Can we accept that this remains an internal state.
 - If it's about resetting the state, we can force remount with a separate `key` on DataViews if needed.

jorgefilipecosta: > If it's about resetting the state, we can force remount with a separate key on DataViews if needed.

I have thought of this option but I think it will probably also have undesired effects. I can give it a try and see how it goes.

jorgefilipecosta:
Hi @youknowriad, with https://github.com/WordPress/gutenberg/pull/63889 merged and an additional small fix on search (included in this PR) the key-based solution to reset the filter showing state seems to be working. This PR is ready for review.

youknowriad (approved):
This is working well for me. What do you think @jameskoster

jorgefilipecosta:
@jameskoster I merged this one to avoid potential complex rebases as this PR has a considerable size. But if there is anything we should improve just let me know and I will gladly work on a follow-up.

oandregal:
@jameskoster @jorgefilipecosta what are your thoughts regarding filters that declare to be primary?

- Before this PR, primary filters were always visible and not listed in the "Add filter" component.
- After this PR, they still are not listed in the "Add filter" component but they are hidden until you click on "Add filter".

There're some inconsistencies with regards to the +1 number as well. Note how it's not set in Patterns (sync is a "primary" filter) but it's set in Pages (all secondary filters) before the user picks a value:

https://github.com/user-attachments/assets/7470a704-3c2b-4a47-9320-fa28aac3ff70

So, I wonder if the concept of "primary filters" still make sense. If so, I believe we should make them visible as before. If we don't want to do that, I think we could just remove the `filter.isPrimary` API because the only thing it does now is to hide the filters from the "Add filter" component.

oandregal:
Also note that the "Add filter" button is hidden if all filters are primary, and that creates weird interactions. Compare Patterns (only 1 filter that is primary) vs Templates (only 1 filter but it is NOT primary). It's not clear why the "Add filter" button would be visible but disabled in the later but hidden in the former:

https://github.com/user-attachments/assets/99b6632a-ac60-4bdb-a981-1d2c52b87272

jameskoster:
> There're some inconsistencies with regards to the +1 number as well

We should fix that 👍 The question is whether to show the count when a filter is added, or when a value is selected. I can see arguments for both. What do you think?

The intent in the [original issue](https://github.com/WordPress/gutenberg/issues/60696) was for the panel to be open automatically if any filters are added. So if a view had primary filters the panel would be open. Looks like we missed tracking that as a follow-up.

> It's not clear why the "Add filter" button would be visible but disabled in the later but hidden in the former

I suspect this is a quirk relating the the a11y requirements. Most views supply non-primary filters that users can add as required. When all filters are added, the button remains accessible to help folks understand no more filters can be added, and to avoid confusion around the button vanishing from the UI.  Because there's only a single filter on the Patterns page, and it's primary, users cannot ever add a new filter here. Consequently it's okay to hide the button entirely.

oandregal:
Follow-up for having the filter toggle opened on load if there are primary filters https://github.com/WordPress/gutenberg/pull/64651

oandregal:
There's another follow-up: what to do with the filter toggle when there are no filters?  Report: https://github.com/WordPress/gutenberg/issues/64631

There's this PR that hides the toggle https://github.com/WordPress/gutenberg/pull/64640

Alternatively, we could show it as "disabled". However, I find the disabled state is useful when there's opportunities for the component to be active. In this case, given there are no filters, there're no opportunities for it to become active — hence my preference to just hide it. In the interest of velocity, I'll approve and go ahead with that PR. Let me know if you'd like to do something different.

jameskoster:
That seems reasonable to me 👍

oandregal:
There's also a couple of details about keyboard interaction / how the "add filter" button should work:

- https://github.com/WordPress/gutenberg/issues/64689
- https://github.com/WordPress/gutenberg/issues/64697

PR #63209, opened 2024-07-05
TimePicker: Fix time zone overflow

mirka:
## What?

Fixes a style regression introduced in #60613 where the time zone in the `TimePicker` was overflowing a bit.

This was especially noticeable in the "Date" block:

Time zone overflowing in popover

## How?

Remove the `width: 100%` style in the container element, by adding a `expanded={ false }` prop to `HStack`.

## Testing Instructions

Compare the result in the Storybook for `TimePicker`. Or, in the editor, the post publish popover or the "Date" block.

## Screenshots or screencast 

| Before | After |
|--------|-------|
|Time zone is overflowing|Time zone is in correct position|


tyxla (approved):
Looks good. Thanks for the fix 🚀

Issue #60528, opened 2024-04-06
Font Library post-WP 6.5 Issues


creativecoder:
This is a collection of open Font Library bugs and enhancements to be considered for WP 6.6, grouped by category, with bugs at the top of each category:

### Accessibility

- [ ] #58319 (bug)
- [ ] #55179

### Collections and installing fonts

- [x] #59408
- [x] #58011
- [ ] #58989
- [x] #58277
- [x] #54797
- [ ] #54779
- [ ] #60914
- [x] #63418
- [ ] #62875
- [ ] [Font Library: Replace error message "A font face matching those settings already exists" with notice (trac#60716)](https://core.trac.wordpress.org/ticket/60716)
- [ ] [Font Library: add terms and conditions (trac#60593)](https://core.trac.wordpress.org/ticket/60593)

### Managing installed fonts

- [x] #54401 (bug)
- [ ] #54601 (bug)
- [ ] #60392 (bug)
- [x] #59704
- [x] #58403
- [x] #55220
- [ ] #58276
- [ ] #55176
- [ ] #62977
- [ ] [Font Library: Add an ability to replace font (trac#60717)](https://core.trac.wordpress.org/ticket/60717)
- [ ] https://github.com/WordPress/gutenberg/issues/51151

### Font files

- [ ] #58375 (bug)
- [x] #59417
- [ ] #59102
- [ ] #58229

### Font post types

- [ ] #59486 (bug)
- [ ] #59353 (bug)
- [ ] #59137 (bug)
- [ ] #58670

### REST API

- [ ] **[Install several formats of the same font one by one is not possible (trac#60654)](https://core.trac.wordpress.org/ticket/60654)**

### Google Fonts collection

- [ ] #54671
- [ ] #58521

### All collections

- [ ] #58774
- [x] #58695
- [ ] #59094
- [ ] https://github.com/WordPress/gutenberg/pull/59412

## Font Names and Previews

- [ ] #60188 (bug)
- [ ] https://github.com/WordPress/google-fonts-to-wordpress-collection/issues/23
- [ ] #58411
- [ ] #59976
- [x] https://github.com/WordPress/gutenberg/pull/59569
- [ ] https://github.com/WordPress/gutenberg/issues/62701

## System Fonts

- [ ] #57295
- [ ] #54186

## Relation to Global Styles and Themes

- [ ] #59974 (bug)
- [ ] #55175
- [ ] #54985
- [ ] #54222

## Related Typography Issues

[...]

Issue #57720, opened 2024-01-10
SelectControl: Add borderless prop


jameskoster:
As noted in https://github.com/WordPress/gutenberg/pull/57670, a couple of UI concepts in the data views work call for a SelectControl with reduced visual prominence. Two examples are; operator selection in filter configuration, and the page selector in pagination:

<img width="991" alt="Screenshot 2024-07-08 at 14 03 57" src="https://github.com/WordPress/gutenberg/assets/846565/36670246-6f1b-46e0-b512-8988c229d8bd">

<img width="795" alt="Screenshot 2024-07-08 at 14 03 44" src="https://github.com/WordPress/gutenberg/assets/846565/5b28bb8e-11cf-404e-b72b-acfe3fae651c">

We can get halfway there by using the `small` size, but it would be good to have a built-in way to hide the border as well.

ciampo:
It looks like that border is inherited from `InputControl`. Is having borderless controls a need that we should consider for the wider group of controls in the package? (cc @mirka )

Not my area of expertise, but have we considered the potential accessibility implications of a borderless dropdown, especially for visually impaired users? (cc @andrewhayward )

jameskoster:
> Is having borderless controls a need that we should consider for the wider group of controls in the package?

It seems _unlikely_ there would be a need for other controls to omit the border, largely because they lose their identity with out one. A borderless SelectControl can work (in theory) because most of the time the chevron will be present and this helps indicate the purpose of the element. 

Hopefully appropriate labelling, hover and focus styles can handle the accessibility requirements, but let's see what Andrew says.

ciampo:
@mirka in terms of APIs, what do you think about adding a `variant` prop with type `undefined | 'unstyled'` ?

mirka:
> @mirka in terms of APIs, what do you think about adding a `variant` prop with type `undefined | 'unstyled'` ?

It is a possibility, but we should first assess whether the borderless style makes sense with all the available sizes/paddings, labeling options, `multiple`, prefix/suffix, etc. If not (which I predict will be the case here), it would be a lot simpler to make a thin facade component with a limited subset of props.

jameskoster:
The use cases that have arisen essentially call for a select with a smaller footprint and less visual prominence. So the likelihood of the options you mention being required is minimal. If that points to a 'facade component' then I think it could work.

ciampo:
> it would be a lot simpler to make a thin facade component with a limited subset of props.

[...]

Issue #55650, opened 2023-10-26
Components: Add `Pagination` component


t-hamano:
## What problem does this address?

The current Gutenberg project uses pagination in several places.

These paginations are built uniquely in each location, but I imagine they could be implemented as a common component. In other words, it is a component similar to [the_posts_pagination()](https://developer.wordpress.org/reference/functions/the_posts_pagination/) function.

## What is your proposed solution?

If we mimic the output of `the_posts_pagination()`, it will probably look something like this:

![buttons-pagination](https://github.com/WordPress/gutenberg/assets/54422211/db9d5dbf-35bb-4ec9-9914-bbd50b62ccec)

Apart from this, we may also need a type that allows you to enter the page number in a text box:

![of-pagination](https://github.com/WordPress/gutenberg/assets/54422211/80f111d9-0065-4a41-a8ec-6725f474cc05)

I'm imagining the following usage, for example:

```javascript
import { Pagination } from '@wordpress/components';

function MyComponent() {
	const [ currentPage, setCurrentPage ] = useState( 1 );
	return (
		<div className="my-pagination">
			<Pagination
				total="10"
				current={ currentPage }
				midSize="3"
				prevText="Previous page"
				nextText="Next page"
				type="list"
				hasPageInput
				isSmall
				onChange={ ( newPage ) => setCurrentPage( newPage ) }
			/>
		</div>
	);
}
```

## Screenshots

Below is the pagination that I was able to find. In addition to this, pagination may also be implemented for revisions, as seen in #53621.

### Patterns Inserter

![patterns-inserter](https://github.com/WordPress/gutenberg/assets/54422211/3fe9746f-dab4-450c-b4ba-7c07ae429f55)

### Pattern Page

![patterns](https://github.com/WordPress/gutenberg/assets/54422211/e9d6741f-06e6-4d35-bdde-3c47261a3c30)

### Data View

![data-view](https://github.com/WordPress/gutenberg/assets/54422211/e174a7b5-aedf-46dc-a8d8-27f75ce43689)

youknowriad:
Closing as the root issue here is addressed by DataViews: the admin views that each had bespoke pagination (patterns, pages, templates…) now run on the DataViews framework, which provides the shared pagination control this proposed.

PR #63210, opened 2024-07-06
Font Library Modal: enhance pagination appearance

### LINKED ISSUE #59408: Align font library pagination appearance with data views
Here's the pagination appearance in the font library:

Ideally this should match the one found in data views:

Screenshot 2024-02-27 at 15 08 39

t-hamano:
Fixes #59408
Part of #60528

## What?

This PR improves the appearance of pagination in the Font Library modal, making it more consistent with the pagination UI in the Data View and Revisions panel.

## How?

- The layout is constructed in much the same way as other similar pagination.
- Changed the display position of the tooltip from bottom to top.
- Also includes minor CSS class name refactoring.

## Testing Instructions

In the font library, make sure pagination still works as before.

### Before

image

### After

image

### DIFF
--- a/packages/edit-site/src/components/global-styles/font-library-modal/font-collection.js
+++ b/packages/edit-site/src/components/global-styles/font-library-modal/font-collection.js
@@ -28,7 +28,13 @@ import {
 } from '@wordpress/components';
 import { debounce } from '@wordpress/compose';
 import { sprintf, __, _x } from '@wordpress/i18n';
-import { moreVertical, chevronLeft } from '@wordpress/icons';
+import {
+	moreVertical,
+	chevronLeft,
+	chevronRight,
+	previous,
+	next,
+} from '@wordpress/icons';
 
 /**
  * Internal dependencies
@@ -429,7 +435,7 @@ function FontCollection( { slug } ) {
 					{ selectedFont && (
 						<Flex
 							justify="flex-end"
-							className="font-library-modal__tabpanel-layout__footer"
+							className="font-library-modal__footer"
 						>
 							<Button
 								variant="primary"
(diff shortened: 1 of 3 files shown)


jasmussen:
This feels like an improvement, so 👍 👍. Though for the pagination on the list view work going on, pagination is still evolving, with some changes around next/previous pages, the dropdown for selecting a specific page, etc. I wonder if it would be worth it to turn that whole piece into a single Pagination component, so we don't have to maintain several different styles of this?

t-hamano:
> I wonder if it would be worth it to turn that whole piece into a single Pagination component, so we don't have to maintain several different styles of this?

This is certainly worthwhile. I suggested this in #55650, but I'm still not sure what UI works best and covers all use cases.

jasmussen:
I think @jameskoster may have worked on some bits recently.

jameskoster:
For data views it has been suggested not to include first/last icon buttons. They add noise, consume space (problematic on mobile), and when there are enough pages to make use of them (quite rare) it's generally easier to use the dedicated page selection UI.

Another suggestion has been to use the double-arrow for next/previous, to reduce overlap between these buttons and other chevrons.

This is the design:

Screenshot 2024-07-08 at 14 02 18

But it's blocked by https://github.com/WordPress/gutenberg/issues/57720

t-hamano:
> For data views it has been suggested not to include first/last icon buttons.

When I created this PR, I wondered whether to apply this approach to the font library modal as well.

On the other hand, the Revisions panel has First Page and Last Page buttons:

image

In addition, there are a huge number of Google fonts, so some users may want to go back to the first page with one click. Also, the first page and last page buttons were originally in the font library modal.

I'm wondering to what extent these UIs should be unified 😅

jasmussen:
I thought about the first and last page buttons too, in the name of the huge amount of pages. However it's not clear to me that those are hugely valuable when you have this many pages, arguably improved _search_ would be better. There use of the double chevron resonates since we really have many triangle chevrons in the UI already. Not a strong opinion.

t-hamano:
I removed the first page/last page buttons.

image

But the icons are not double chevrons because all paginations use single chevrons for previous/next buttons.

Is it ok to change it to double chevrons in this PR? Or should we unify everything to double chevrons, including other paginations, in a follow-up PR?

jasmussen:
> Is it ok to change it to double chevrons in this PR? Or should we unify everything to double chevrons, including other paginations, in a follow-up PR?

Seems fine to me, yes. Regardless, there's work to be done across pagination, even for list views.

jasmussen (approved):
Seems okay as an interim step, but more work can be done separately.

PR #63211, opened 2024-07-06
Font Library: Group fonts by source

### LINKED ISSUE #58403: Font Library: Global Styles > Typography sidebar – group related font sources

t-hamano:
Closes #58403
Part of #60528

## What?

This PR groups fonts by source (`theme`, `custom`) in the Global Styles.

## Why?

- To help users understand which source a font comes from without having to open the font library.
- To improve consistency with other presets (Color, Shadow).

## How?

The thing I struggled with the most was what label to apply to the source that came from the `custom` source. There are three possible labels: `Installed Fonts`, `User Fonts`, and `Custom Fonts`, but I chose "**User Font**".

At the same time, to improve consistency between the sidebar and the Font Library modal, I've updated the source order to `theme`, `custom`. In the Font Library modal, this means that the order will change from **Custom, Theme** to **Theme, Custom**.

## Testing Instructions

- Access the font library.
- Confirm that the UI is displayed properly in the following four scenarios:
  - Theme fonts and user fonts
  - Only theme fonts
  - Only user fonts
  - No fonts

## Screenshots or screencast 

| Context | Image |
|--------|--------|
| Theme fonts and user fonts | theme-and-user-fonts |
| Only theme fonts | theme-fonts |
| Only user fonts | user-fonts |
| No fonts | no-fonts |


packages/edit-site/src/components/global-styles/font-families.js
```diff
@@ -48,48 +49,51 @@ function FontFamilies() {
...
-						>
-							{ __( 'Add fonts' ) }
-						</Button>
-					</>
+					</VStack>
+				) }
+				{ ! hasFonts && (
+					<VStack>
+						<Subtitle level={ 3 }>{ __( 'Fonts' ) }</Subtitle>
+						<Text>{ __( 'No fonts installed.' ) }</Text>
```
afercia: Nitpicking: By default, the Text component renders a `<span>` element. I'd rather use a paragraph `as="p"`. Even though in practice it won't make a perceivable difference, it would illustrate a good best practice: use semantic elements when appropriate.

afercia (commented):
Grouping the fonts makes sense to me and I think this is an improvement.
Only two questions:

Wouldn't be better to list the User Fonts first? I'm assuming that if I install some fonts that's because I'm going to 'prefer' them over the theme ones and use them more often. In a way, the Theme Fonts would have less relevance because I installed 'my fonts' to use. As such, I would expect 'my fonts' to be listed first. But yes, it's an assumption, in a way.

Re: naming: how about 'Your Fonts' instead of 'User Fonts'?

packages/edit-site/src/components/global-styles/font-families.js
t-hamano: Fixed by d6d7f1a

t-hamano:
Thanks for the review!

> Wouldn't be better to list the User Fonts first?

I don't have a strong opinion on this one, and I think there are reasonable reasons for either ordering.

- **Theme, Custom**: Custom presets are always placed last among color and shadow presets, so font families follow that rule too.
- **Custom, Theme**: This sorting is consistent with the current font library modal, and is a good idea if many users use custom fonts more frequently than theme fonts.

>Re: naming: how about 'Your Fonts' instead of 'User Fonts'?

I feel like this label makes it unclear which _origin_ the font belongs to, because in the bigger picture, theme fonts might also be considered "Your fonts".

I think we need to use one of the words "Installed", "User", or "Custom" to indicate that the font does not belong to the theme font.

If we follow a color preset or shadow preset, "Custom Fonts" might be a good choice.

afercia:
> I feel like this label makes it unclear which origin the font belongs to, because in the bigger picture, theme fonts might also be considered "Your fonts".

Thanks for your feedback, I don't have strong opinions as well. However, I'd tend to think the whole concept of 'source' is a little unfriendly for users.

packages/edit-site/src/components/global-styles/font-families.js
```diff
@@ -48,48 +49,51 @@ function FontFamilies() {
...
-						>
-							{ __( 'Add fonts' ) }
-						</Button>
-					</>
+					</VStack>
+				) }
+				{ ! hasFonts && (
+					<VStack>
+						<Subtitle level={ 3 }>{ __( 'Fonts' ) }</Subtitle>
+						<Text as="p">{ __( 'No fonts installed.' ) }</Text>
```
matiasbenedetto, line 84: If there are fonts installed but none of them are active the text, 'No fonts installed' is technically incorrect. Should we replaced this by 'No fonts activated' or add a conditional to render 'activated'/'installed' depending on the availability of base fonts?

matiasbenedetto (commented):
This seems to be working fine. 
Should we maintain the same wording as in other parts of the editor and use 'Custom' instead of 'User'? I haven't seem the term 'User' applied to the different origins of the styles across the editor.

Apart from that the word 'fonts' in the origin subtitle (i.e.:'Theme fonts') could be redundant and I haven't seem it used in other parts of the editor for similar settings. Colors and Shadows use just 'Default', 'Theme' and 'Custom'. Probably is good to stick to that to be more consistent with the rest of the UI.

packages/edit-site/src/components/global-styles/font-families.js
afercia: Good point. I'd agree that's incorrect but it already happens on trunk. It's not related to this PR.
I'd consider to open a separate issue. With fonts installed but no one activated, the wording should be different. Also, the 'Add fonts' button would be misleading and should be instead the 'Manage fonts' one.

t-hamano:
>Should we maintain the same wording as in other parts of the editor and use 'Custom' instead of 'User'?

Updated.

If we find the term "Custom" confusing, we may need to consider it for all presets: color presets, shadow presets, and font size presets (Font size presets UI is currently under development in #63057).

### Global Styles Sidebar

image

### Font Library Modal

image

packages/edit-site/src/components/global-styles/font-families.js
afercia: I will split this to a new issue so that this PR can move on

matiasbenedetto (approved):
This looks good to me. I think it's ready to go.

Still, I would like this feedback to be addressed before merging:

> Apart from that the word 'fonts' in the origin subtitle (i.e.:'Theme fonts') could be redundant and I haven't seem it used in other parts of the editor for similar settings. Colors and Shadows use just 'Default', 'Theme' and 'Custom'. Probably is good to stick to that to be more consistent with the rest of the UI.

PR #63213, opened 2024-07-06
Patterns Page: Hide preview column by default

### LINKED ISSUE #60704: Patterns Page: Hide preview column by default

#60337 introduced the table layout on Pattern page. Currently, the pattern preview is displayed and cannot be hidden.

image

I suggest the following changes:

- Hide previews by default in the table layout.
- Can display a preview from the view options menu.

These two match the specifications of the current template page and template parts page.

image

akasunil:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/60704

## Why?

Pattern preview is displayed by default on patterns list page for table view, where it should be hidden by default as we have implemented for templates page.
 
## Screenshots or screencast 
[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress.webm](https://github.com/WordPress/gutenberg/assets/61308756/af4d08ab-f0e4-4417-b345-5bb9f2878500)

### DIFF
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -82,6 +82,7 @@ const DEFAULT_VIEW = {
 	layout: {
 		...defaultConfigPerViewType[ LAYOUT_GRID ],
 	},
+	fields: [ 'title', 'sync-status' ],
 	filters: [],
 };
 
@@ -285,7 +286,6 @@ export default function DataviewsPatterns() {
 					<Preview item={ item } viewType={ view.type } />
 				),
 				enableSorting: false,
-				enableHiding: false,
 				width: '1%',
 			},
 			{


t-hamano (commented):
Thanks for the review!

From a code perspective everything seems correct.

@oandregal @jameskoster I'm OK with shipping this PR, what do you think?

Since the preview isn't shown by default in table view for Templates and Pages, I think this PR makes sense.

jameskoster:
Aligning with the templates table makes sense 👍 

I see Table layout as generally more of a 'management' view, optimised for bulk actions and so on. So it seem reasonable to hide the preview by default as it's not really helpful for those management tasks, particularly given its impact on performance and information density.

t-hamano:
@jameskoster Thanks for the feedback! Now let's ship this PR.

t-hamano (approved):

PR #63215, opened 2024-07-06
Font Appearance Control: Refactor font appearance fallbacks

### LINKED ISSUE #63479: Pattern Inserter: Multiple undo histories are created
### Description

When I insert a pattern, multiple undo histories are created, so I cannot delete a pattern unless I press the Undo button several times after inserting it. The amount of history created depends on the pattern inserted.

I looked back through the commit history and it looks like this issue first appeared in #61915.

After inserting a pattern, click the Undo button.

### Before #61915

https://github.com/user-attachments/assets/b78407f5-1986-4701-a63b-b4c00f5228d6

### After #61915

https://github.com/user-attachments/assets/4852705c-baca-4d7c-9b93-0d74b1aee909

### Environment info

_No response_

Yes

Yes

mikachan:
## What?

Refactors the logic around the font weight and style fallbacks applied when a new font family is selected. Follow-up to https://github.com/WordPress/gutenberg/pull/61915.

## Why?

Addresses this comment: https://github.com/WordPress/gutenberg/pull/61915#discussion_r1666394116.

## How?

Adds a new `findNearestStyleAndWeight` function that finds the nearest available weight and style to apply when the font family is changed, regardless of whether the font family is selected using global styles or dynamically via a new block. Also, as the logic now sits in a function, it is easier to add automated tests.

## Testing Instructions

Ensure there are a variety of font faces installed in multiple font families, including different font weights, styles and at least one system font.

1. Open Global Styles > Typography
2. Change the active font by using the Font dropdown menu to a font that has different weights/styles
3. Ensure that the nearest available appearance options are selected automatically, e.g. the closest available font weight and the same font style where available

Try running the following in the console:

```
var block = wp.blocks.createBlock('core/paragraph', {
  content: 'Hello',
  style: { typography: { fontWeight: '300' } } }
);
wp.data.dispatch('core/block-editor').insertBlock( block )
```

The nearest available font weight and style should be automatically applied.


packages/block-editor/src/components/global-styles/typography-panel.js
```diff
@@ -257,33 +256,16 @@ export default function TypographyPanel( {
...
 			} );
-		}
-
-		// Reset font appearance if the font family does not have the selected font style
-		if ( ! hasFontStyle ) {
+		} else {
+			// Reset font appearance if there are no available styles or weights
 			resetFontAppearance();
 		}
 	}, [ fontFamily ] );
```
talldan: There are some missing dependencies, which results in a linter warning.

I think it'd be good to make sure that's fixed or the rule is disabled for this line (with a comment mentioning why).

To fix it, it looks like `setFontAppearance` and `resetFontAppearance` would have to be wrapped in `useCallback`.

packages/block-editor/src/components/global-styles/typography-panel.js
```diff
@@ -257,33 +256,16 @@ export default function TypographyPanel( {
...
-				fontWeight,
+				fontStyle: nearestFontStyle,
+				fontWeight: nearestFontWeight,
 			} );
-		}
-
-		// Reset font appearance if the font family does not have the selected font style
-		if ( ! hasFontStyle ) {
+		} else {
+			// Reset font appearance if there are no available styles or weights
```
talldan: Nit: the coding standards mention comments should end in a period, though it's not enforced by the linter, so not very important (I wouldn't worry too much about fixing ones not added in this PR).

talldan (approved):
The changes look good, the code is much neater, while also fixing that problem. Thanks for following up so quickly.

The one possible issue I'm seeing is that when adding a block programmatically using the snippet, the appearance dropdown still shows 'default'.

I guess that leads to a philosophical question of whether an absence of a `fontStyle` means some kind of default, which is tricky as the theme might have some custom css setting the font style to something else entirely. 🤔 

It might be ok to leave it as it is, I'm not too familiar with the feature or the code, so it might be good to gather opinions from those more in the know.

mikachan:
Thanks for the quick review, @talldan!

> The one possible issue I'm seeing is that when adding a block programmatically using the snippet, the appearance dropdown still shows 'default'.

I think this should only happen if there are no available styles/weights from the active font family for the settings passed in this snippet. This should be unlikely if a system font is active, as this should have all of the available styles and weights. Do you still see this if a system font is active before you run this snippet?

Here's an example of running the snippet with just a weight setting of 300:

https://github.com/WordPress/gutenberg/assets/1645628/153a7b6d-d688-4bea-82cd-6ccb8ce61ca5

packages/block-editor/src/components/global-styles/typography-panel.js
mikachan: I tried adding all of the dependencies but it looks like this causes an infinite loop, and I think it's OK to rely on only the `fontFamily` changing here as this is the main prop we want to watch in this case.

I've disabled the rule for this line with a comment explaining why: https://github.com/WordPress/gutenberg/pull/63215/commits/37963050aebcb6b0da3b42cf3139c5795421732d

I left the `useCallback` functions in for `setFontAppearance` and `resetFontAppearance` as they seem to work the same either way, and I'm guessing this might have some small performance benefits.

packages/block-editor/src/components/global-styles/typography-panel.js
mikachan: Thanks! I've added periods to the end of the comments added in this PR.

mikachan:
I've added some e2e tests to help ensure the Appearance dropdown is updated correctly, based on the snippet that adds a new block with appearance settings: https://github.com/WordPress/gutenberg/pull/63215/commits/7e027b4588e48715fd2d7475800d32b63396e1b1.

I've also split the "finding the nearest font style" logic out into its own function: https://github.com/WordPress/gutenberg/pull/63215/commits/72199ccaa807f0f3a138e1bcb68a45b70b67a9a9. There should be no difference in functionality, I just thought this was better as it's easier to read and more closely matches the font weight logic.

packages/block-editor/src/components/global-styles/typography-utils.js
```diff
@@ -185,3 +185,106 @@ export function findNearestFontWeight(
...
 	return nearestFontWeight;
 }
+
+/**
+ * Returns the nearest font style based on the new font style.
+ * Defaults to an empty string if the new font style is not valid or available.
+ *
+ * @param {Array}  availableFontStyles Array of available font weights.
+ * @param {string} newFontStyleValue   New font style value.
+ * @return {string} Nearest font style.
```
creativecoder: It looks like this function could also return `undefined` if none of the conditionals are triggered, is that right? It so, let's update the return type here.

creativecoder (commented):
This seems to be functioning as expected for all of the test cases discussed in https://github.com/WordPress/gutenberg/pull/61915.

And I can see how programmatic block insertion now has the specified weight and style appearance.

https://github.com/WordPress/gutenberg/assets/1699996/957b0d77-6d5b-487c-89cc-c814c8e8fb4c

packages/block-editor/src/components/global-styles/typography-utils.js
mikachan: Oh good spot! My intention was that it would default to an empty string if none of the conditionals were triggered. I've updated the function here: https://github.com/WordPress/gutenberg/pull/63215/commits/b407be1654b1ce6851b5a72e155180c39f27e9dd, I think this should cover all branches.

mikachan:
Thanks for the reviews @talldan and @creativecoder! I'd appreciate one more approval before bringing this in as there have been a couple of changes since the last approval. I think this is ready now though, thank you for the help 🙇

noisysocks:
Thanks for adding 18e1bfebf9b50cae3bb9fc81d1bf3814681b7922. I checked and running the effect only when the value changes (not when the component mounts) indeed fixes the issue I was noticing yesterday in https://github.com/WordPress/gutenberg/pull/63333 ❤️

t-hamano:
I reported an issue in #63479 where inserting a pattern would create multiple undo histories.

This PR seems to solve that issue as well.

creativecoder (commented):
My results testing this PR today:

I manually changed a theme.json `fontStyle` for one of the fonts to `oblique` for testing.

1. ✅ Switch from font with Bold Italic to Black Italic and back
  - Both trunk and PR: Appearance setting switches to nearest font weight 
2. ✅ Switch from font with Oblique Regular or Bold settings to one without Oblique
  - trunk: Appearance setting switches to Default
  - PR: Appearance setting switches to nearest font weight with italic
3. ❓Insert block from console with unavailable font weight and style
  - trunk: Appearance setting is Default, font does not display the specified appearance
  - PR: Appearance setting is Default, but font does display the nearest weight and style

Number 3 is different than [my earlier test](https://github.com/WordPress/gutenberg/pull/63215#pullrequestreview-2164376463), so seems like a regression?

Another question I have: should we _not_ add the faux italic appearance if the font already has an oblique style (it seems redundant)? But we could handle this in a follow-up, if needed.

mikachan:
> Number 3 is different than https://github.com/WordPress/gutenberg/pull/63215#pullrequestreview-2164376463, so seems like a regression?

Thanks @creativecoder. Yes, I think it is a regression. I think the logic for checking if the font appearance settings have changed should sit in `setFontAppearance` rather than the new `useEffect`. Does this work better after the last commit - https://github.com/WordPress/gutenberg/pull/63215/commits/8288dd40a8eecf0be54ef06f2919540bd3687420?

> Another question I have: should we not add the faux italic appearance if the font already has an oblique style (it seems redundant)? But we could handle this in a follow-up, if needed.

Good point, this sounds like a good follow-up!

talldan (approved):
This is still working well for me with the latest changes, though I haven't tested as many aspects as @creativecoder, so it might be good to also check for Grant's approval.

creativecoder (approved):
> Does this work better after the last commit - https://github.com/WordPress/gutenberg/commit/8288dd40a8eecf0be54ef06f2919540bd3687420?

Yes, all of [my test cases](https://github.com/WordPress/gutenberg/pull/63215#pullrequestreview-2175861915) are now working 😄.

mikachan:
Thanks so much for all the reviews and testing! I'll bring this in 🎉

andrewserong:
I just cherry-picked this PR to the release/18.8 branch to get it included in the next release: 72df623a5d6

PR #60739, opened 2024-04-15
WIP Experiment: combine gradient and background image into `background-image`

ramonjd:
Initial test of an experiment to combine`color.gradient` and `background.backgroundImage` into one CSS property `background-image`

I'm testing merging values from two leafs in the theme.json styles tree:

1. `color.gradient`
2. `background.backgroundImage`

The merge also needs to occur where one of those two values is defined by global styles.

It's all a bit messy, which makes me think that a neater approach might be to focus on the `background` block supports and add gradient support there.

## TODO

- [x] Think about ordering: what if users what to have a transparent png overlaying a gradient background? Can be part of an alternative [theme.json type](https://github.com/WordPress/gutenberg/pull/60739#issuecomment-2081893441), e.g., an array of values.
- [ ] To merge the CSS in the editor on the clientside, the block needs to know if it has a global style value. Need a standard way to do this, in another PR. Possibly in useBlockProps? 
- [ ] The same point above applies for the background backend block supports.

## Why?

- https://github.com/WordPress/gutenberg/issues/60401
- https://github.com/WordPress/gutenberg/issues/32787

## Testing Instructions

<details>

<summary>Test theme.json</summary>

```json

{
	"$schema": "../../schemas/json/theme.json",
	"version": 3,
	"settings": {
		"appearanceTools": true,
		"color": {
			"defaultGradients": true,
			"gradients": [
				{
					"slug": "fiery-cavern-of-doom",
					"gradient": "linear-gradient(180deg, rgba(100,201,231,0.26936712184873945) 0%, rgba(212,183,64,0.5746892507002801) 46%, rgba(255,0,0,0.6335127801120448) 100%)",
					"name": "Fiery cavern of doom"
				}
			]
		}
	},
	"styles": {
		"color": {
			"gradient": "var(--wp--preset--gradient--fiery-cavern-of-doom)",
			"text": "#ffffff"
		},
		"background": {
			"backgroundImage": {
				"url": "https://images.pexels.com/photos/105819/pexels-photo-105819.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr="
			},
			"backgroundSize": "cover",
			"backgroundAttachment": "fixed"
		},
		"blocks": {
			"core/quote": {
				"color": {
					"gradient": "linear-gradient(180deg, rgba(158,231,100,0.26936712184873945) 0%, rgba(64,212,151,0.5746892507002801) 46%, rgba(0,168,255,0.6335127801120448) 100%)"
				},
				"background": {
[...]

Issue #60401, opened 2024-04-03
Setting background-image overrides background gradient


susurruses:
### Description

When editing a block that supports both a Background Gradient and Background Image, these are set as inline styles in such a way that the image overrides the gradient. For example, here's the style attribute for a group block with both a gradient and image:

```style="background:linear-gradient(265deg,#D8613C 0%,#F9F9F9 100%);background-image:url('https://example.com/wp-content/uploads/2024/04/image.png');background-size:cover;"```

Ideally, this should instead output something like the following:

```style="background-image:url('https://example.com/wp-content/uploads/2024/04/image.png'),linear-gradient(265deg,#D8613C 0%,#F9F9F9 100%);background-size:cover;"```
(Note the switch to `background-image` from `background`, and comma separation between the two `background-image`s.)

The use case for having both a background-image and background gradient is when a partially-transparent image is shown above a gradient, such as putting a pattern or other graphic on top of a gradient. (In theory there could also be use cases of wanting a translucent gradient on top of an image, but that would likely be better served by a Cover block.)

There is already a ticket for switching `background` to `background-image` ([32787](https://github.com/WordPress/gutenberg/issues/32787)), but that alone won't solve the issue here of needing a single `background-image` property to contain both the gradient and image.

A work-around is to nest group blocks, setting the gradient on one and the image on the other, but this is clunky and the presence of both background and background-image in the interface implies that both can be added, with no error shown for trying to add both.

### Step-by-step reproduction instructions

1. Create a Group block
2. Set a background gradient for the group block
3. Set a background image (e.g., you could try one of these [transparent background images](https://www.transparenttextures.com/))
4. Save, view the page, and view the source to see a `background` property for the gradient and `background-image` for the image.

### Environment info

WordPress 6.5 with TT4 theme

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

**Comment by ramonjd** [comment]
This could be done while addressing https://github.com/WordPress/gutenberg/issues/32787 I think

Ideally Gutenberg would use `background-image` for both. And, where there are two, create something like:

```css
background-image: linear-gradient(rgba(0, 0, 255, 0.5), rgba(255, 255, 0, 0.5)),
                  url("../../uploads/dragon.png");
```

susurruses:
[...]

PR #60264, opened 2024-03-28
Global styles: background UI control labels

ramonjd:
Part of:

- https://github.com/WordPress/gutenberg/issues/54336

Follow up to:

- https://github.com/WordPress/gutenberg/pull/59454

## What?

This PR changes the heading labels for background block supports tools panel and global styles.

### Post editor

- "Background" > "Background image" for the tools panel

### Site editor

- "Background" > "Background image" for the styles panel
- "Background" > "Image" for the tools panel, to reduce repetition between the style panel header.

It also applies the following label logic for the media upload component: 

- If an image has been selected, then `truncatedImageTitle || truncatedImageFilename`
- No image: `"Select or upload an image"`

See context: https://github.com/WordPress/gutenberg/pull/59454#issuecomment-2022531023

> [!TIP]
> This PR also harmonizes the tool panel items between block and site editors except for global styles, where the background size panel displays by default for top-level styles, but only when there's an image available.

## Why?

To reduce confusion between background color. Since the only available functionality relates to background image, then I think it's okay to be precise.

## Testing Instructions

I'm using 2024 to test.

1. In the post editor, insert a Group block. 
2. Open the block settings sidebar and select the styles tab.
3. Confirm that the tools panel title is `"Background image"` and the status of media upload component: `"Select or upload an image"`
4. Upload and image and check that it uses either the title or the file name of the image. 
5. Reset the values. The upload component should revert to  `"Select or upload an image"`
6. Now repeat the above steps in the site editor, checking the the tools panel header is "Image"
7. Finally check that the background size controls are not activated by default, except for top-level global files where it's displayed by default, but only when there's an image available:

### Post editor
Screenshot 2024-04-15 at 10 01 11 am

Screenshot 2024-03-28 at 11 08 08 am

### Site editor
Screenshot 2024-04-15 at 9 58 56 am

Screenshot 2024-03-28 at 11 08 42 am

Screenshot 2024-03-28 at 11 08 19 am

PR #63216, opened 2024-07-06
Global Styles: Improve organisation of site background color and image

### LINKED ISSUE #63096: Improve organisation of site background options
Extracted from https://github.com/WordPress/gutenberg/issues/62752 as a smaller step forwards.

bg

In global styles background options are a bit dis-oraganised and in the case of background image, tricky to find. This can be addressed by adding a 'Background' section which collates both options.

The background color option can continue to live in the Color section as it does today.

### Icon svg:

```
<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M11.53 4.47a.75.75 0 1 0-1.06 1.06l8 8a.75.75 0 1 0 1.06-1.06l-8-8Zm5 1a.75.75 0 1 0-1.06 1.06l2 2a.75.75 0 1 0 1.06-1.06l-2-2Zm-11.06 10a.75.75 0 0 1 1.06 0l2 2a.75.75 0 1 1-1.06 1.06l-2-2a.75.75 0 0 1 0-1.06Zm.06-5a.75.75 0 0 0-1.06 1.06l8 8a.75.75 0 1 0 1.06-1.06l-8-8Zm-.06-3a.75.75 0 0 1 1.06 0l10 10a.75.75 0 1 1-1.06 1.06l-10-10a.75.75 0 0 1 0-1.06Zm3.06-2a.75.75 0 0 0-1.06 1.06l10 10a.75.75 0 1 0 1.06-1.06l-10-10Z" /></svg>
```

amitraj2203:
## What?
Closes: #63096 

## Why?
To improve the organisation for background color and images by creating a new Background Panel in global styles.

## How?
This PR adds a `Background` section which collates both options.

## Testing Instructions
1. Open a Site Editor.
2. Go to global styles.
3. See the Background Panel.

## Screenshots or screencast 
https://github.com/WordPress/gutenberg/assets/77401999/bc5e6c41-1028-4157-a136-4d2b97c5afe1


jasmussen:
Nice! This looks really good:

branch

Potentially we could merge this as an iteration, pending a rebase to fix the conflicts. 

Further improvements would be for these two pieces, which should probably be separate PRs we can consider:

Screenshot 2024-07-31 at 13 05 24

Screenshot 2024-07-31 at 13 05 31

The centered "Add background image" looks a little off, we might want to left align that text and add an upload icon in the spot on the left. 

And the background image inside the ItemGroup once uploaded should have a rounded-rectangle shape, not the circle "swatch" shape.

amitraj2203:
@jasmussen Thanks for the review. I have rebased the branch.

richtabor:
@jasmussen what do you think about an empty state just like colors? 

CleanShot 2024-07-31 at 16 08 52

richtabor:
These should have the same exact UI as the items in the Color panel:

CleanShot 2024-07-31 at 16 10 18

The color panel: 


CleanShot 2024-07-31 at 16 10 45

richtabor:
Could use the same treatment as the Colors panel "Elements" perhaps (with no border divider too):

CleanShot 2024-07-31 at 16 14 15

jasmussen:
> what do you think about an empty state just like colors?

Love it.

amitraj2203:
Thanks @richtabor & @jasmussen  for your review. I have addressed the border styles on focus issue and also added the upload icon for background image and changed its shape. Can you please review it again?

https://github.com/user-attachments/assets/2ee097f5-1a86-4470-906e-970838933925

jasmussen:
Nice work, getting very close now:

styles

The border radius on the square here should be 2px:

Screenshot 2024-08-02 at 08 39 38

Also there's some border changes happening which means if you insert a group, the background control is missing a top border:

Screenshot 2024-08-02 at 08 40 13

Since this is touching some ItemGroup related pieces, it might be good with a code check from @WordPress/gutenberg-components too.

But functionally, this is 90% there. Thank you for the PR!

amitraj2203:
@jasmussen Thanks for the review. I have addressed it now.

jasmussen:
Behavior-wise, design-wise, this is solid now, nice work, thank you!

Code-wise this needs a review. There's a bit more CSS than I expected, that looks specialized mainly to this context. I wonder if our components group (already pinged) has a better approach we can take here, or not. Otherwise, this is good to go from the UI perspective.

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -188,6 +198,13 @@ function InspectorImagePreviewItem( {
 							} }
 						/>
 					</span>
+				) : (
+					<span
+						className="block-editor-global-styles-background-panel__inspector-image-indicator-wrapper"
+						aria-hidden
+					>
+						<span className="component-color-indicator" />
```
mirka: We should never use CSS classes from other components because it breaks component encapsulation. I would recommend a more simpler way using the actual `ColorIndicator` component, for example:

```diff
diff --git a/packages/block-editor/src/components/global-styles/background-panel.js b/packages/block-editor/src/components/global-styles/background-panel.js
index ffc23383b8..3189f19df2 100644
--- a/packages/block-editor/src/components/global-styles/background-panel.js
+++ b/packages/block-editor/src/components/global-styles/background-panel.js
@@ -186,26 +186,16 @@ function InspectorImagePreviewItem( {
 				as="span"
 				className="block-editor-global-styles-background-panel__inspector-preview-inner"
 			>
-				{ imgUrl ? (
-					<span
-						className="block-editor-global-styles-background-panel__inspector-image-indicator-wrapper"
-						aria-hidden
-					>
-						<span
-							className="block-editor-global-styles-background-panel__inspector-image-indicator"
-							style={ {
-								backgroundImage: `url(${ imgUrl })`,
-							} }
-						/>
-					</span>
-				) : (
-					<span
-						className="block-editor-global-styles-background-panel__inspector-image-indicator-wrapper"
-						aria-hidden
-					>
-						<span className="component-color-indicator" />
-					</span>
-				) }
+				<ColorIndicator
+					aria-hidden
+					style={ {
+						'--image-url': imgUrl ? `url(${ imgUrl })` : undefined,
+					} }
+					className={ clsx(
+						'block-editor-global-styles-background-panel__inspector-image-indicator',
+						{ 'has-image': !! imgUrl }
+					) }
+				/>
 				<FlexItem as="span" style={ imgUrl ? {} : { flexGrow: 1 } }>
 					<Truncate
 						numberOfLines={ 1 }
diff --git a/packages/block-editor/src/components/global-styles/style.scss b/packages/block-editor/src/components/global-styles/style.scss
index 98a913b686..0df4f3c016 100644
--- a/packages/block-editor/src/components/global-styles/style.scss
+++ b/packages/block-editor/src/components/global-styles/style.scss
@@ -198,39 +198,14 @@
 	text-align: start;
 }
 
-.block-editor-global-styles-background-panel__inspector-preview-inner {
-	.block-editor-global-styles-background-panel__inspector-image-indicator-wrapper {
-		width: 20px;
-		height: 20px;
-		min-width: auto;
-
-		.component-color-indicator {
-			border-radius: $radius-medium;
-		}
-	}
-}
 
 .block-editor-global-styles-background-panel__inspector-image-indicator {
 	background-size: cover;
-	border-radius: $radius-medium;
-	width: 20px;
-	height: 20px;
-	display: block;
-	position: relative;
-}
+	border-radius: $radius-small;
 
-.block-editor-global-styles-background-panel__inspector-image-indicator::after {
-	content: "";
-	position: absolute;
-	top: -1px;
-	left: -1px;
-	bottom: -1px;
-	right: -1px;
-	border-radius: $radius-medium;
-	box-shadow: inset 0 0 0 $border-width rgba(0, 0, 0, 0.2);
-	// Show a thin outline in Windows high contrast mode, otherwise the button is invisible.
-	border: 1px solid transparent;
-	box-sizing: inherit;
+	&.has-image {
+		background-image: var(--image-url);
+	}
 }
 
 .block-editor-global-styles-background-panel__dropdown-content-wrapper {

```

packages/block-editor/src/components/global-styles/style.scss
```diff
@@ -73,18 +73,67 @@
...
+			border-top: 1px solid $gray-300;
+			border-radius: $radius-block-ui;
+		}
+
+		> div,
+		> div > div,
+		> div > button,
+		> div > div > button {
+			border-radius: inherit;
+		}
```
mirka: This seems like a code smell to me. It looks like we're not using [`ItemGroup`](https://wordpress.github.io/gutenberg/?path=/story/components-experimental-itemgroup--with-border) correctly. `<ItemGroup isBordered isSeparated>` is supposed to handle these border concerns. It should be wrapping both the "Image" and "Color" items, but right now it is only wrapping the "Image" item.

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -720,44 +939,84 @@ export default function BackgroundPanel( {
...
+								disableCustomGradients:
+									! areCustomGradientsEnabled,
+							} }
+							panelId={ panelId }
+						/>
+					) }
+				</>
 			</div>
 
 			{ /* Dummy ToolsPanel items, so we can control what's in the dropdown popover */ }
```
mirka: I think this "Dummy ToolsPanel" hack is causing the item order in the dropdown to be out of sync with the actual items. Not immediately sure why we need to hack it like this — ideally it shouldn't be a hack, but we could at least move it up the file so the order is correct.

ToolsPanel dropdown out of order

mirka (commented):
I left some notes around component usage. The ItemGroup one is not necessarily blocking for this PR, as I believe it may require a substantial refactor. I do recommend cleaning it up though, as it looks quite fragile and hacky.

I'll tag @ramonjd for a better technical review as he may have insights into the feasibility/priority of a refactor.

packages/block-editor/src/components/global-styles/style.scss
```diff
@@ -73,18 +73,67 @@
...
 	grid-column: 1 / -1;
 
+	& > div {
+		// Border styles.
+		border-left: 1px solid $gray-300;
+		border-right: 1px solid $gray-300;
+		border-bottom: 1px solid $gray-300;
+
+		&:not(:only-child):nth-child(1 of &) {
+			border-top-left-radius: $radius-block-ui;
```
tyxla: Might want to use `$radius-small` instead - we're aiming to deprecate `$radius-block-ui`, see #64007

packages/edit-site/src/components/global-styles/root-menu.js
```diff
@@ -73,6 +75,15 @@ function RootMenu() {
 						{ __( 'Layout' ) }
 					</NavigationButtonAsItem>
 				) }
+				{ hasBackgroundPanel && (
```
mtias: Let's place this one at the top instead of bottom

mtias:
Looking good!

mtias:
One thing that slightly weird is that the reset dropdown menu shows Color first and Image second, but the UI shows it inverted.

image

packages/block-editor/src/components/global-styles/background-panel.js
mtias: I was reporting the same thing :)

packages/edit-site/src/components/global-styles/root-menu.js
mtias: Moved it in [7f4951d](https://github.com/WordPress/gutenberg/pull/63216/commits/7f4951d474b52a0774532ee43a0d396f92352013)

packages/block-editor/src/components/global-styles/style.scss
mtias: Yeah, we should reconcile this

mtias:
@jasmussen @jameskoster we should also not call this tool panel "Elements" since these are not elements. Not sure what to use, since "Background" would be redundant with the entire panel name.

packages/icons/src/library/background.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * WordPress dependencies
+ */
+import { Path, SVG } from '@wordpress/primitives';
+
+const background = (
+	<SVG width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg">
+		<Path
+			fill-rule="evenodd"
```
ramonjd: React wants these to be `fillRule` and `clipRule`

packages/block-editor/src/components/global-styles/background-panel.js
ramonjd: I [added dummy tools panel item](https://github.com/WordPress/gutenberg/pull/60151#issuecomment-2024403013) so that the toolspanel ellipsis menu would play nicely, at least semi-nicely with the popover.

It was unclear whether we wanted to preserve the ToolPanel functionality of being able to select "default" controls, and also toggle non-default block support controls.

You can see this in action with many of the Typography controls.

https://github.com/user-attachments/assets/de174b82-66ce-48b4-840f-94d931d336db

Tools panel items need to be rendered inside a ToolsPanel component in order to be registered. Because the controls are in the popover, toggling these non-default controls didn't work while the popover was closed.

So essentially it was a bit of a hack, but one that would leave open the possibility of allowing toggling. It  was introduced before I [moved the controls into the popover](https://github.com/WordPress/gutenberg/pull/60151).

In my mind was the [ongoing work](https://github.com/WordPress/gutenberg/pull/60739) to refactor `color.gradient` to share the `background-image` CSS property, and possibility of applying position/size/repeat values to gradient and not just images.

**However...**

Now that you've brought it up, I reckon we could reconsider all this and let the background controls all live as one control group, with no sub-style toggling. 

Especially since they're now in a popover and space isn't such an issue, but also I think it would be annoying to have to toggle.

Whenever the gradient/image consolidation takes place, we can revisit the controls. 

This PR is big enough, I'll throw up a smaller one 😄

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -720,44 +939,84 @@ export default function BackgroundPanel( {
...
+						</BackgroundControlsPanel>
+					) : (
+						<BackgroundImageControls
+							onChange={ onChange }
+							style={ value }
+							inheritedValue={ inheritedValue }
+							themeFileURIs={ themeFileURIs }
+						/>
+					) }
+					{ showBackgroundPanel && path === '/background' && (
```
ramonjd: Given this change, for now, is targeting top-level styles in the site editor only, I think it'd be good to extract the color-related styles into a separate component.

The background panel is still used for block supports and block global styles, and it would make this file easier to work with in my opinion.

Also, there's a lot of refactoring going on right now (e.g., https://github.com/WordPress/gutenberg/pull/64192) - it might save us a bit of rebase disco 😄

ramonjd (commented):
Thanks for picking this up 🎉 

It's a great start. 

To add to the feedback, I think we ought to split the component given that it targets a single context. It will make follow ups easier to reason about (and rebase) I think.

Not only that, he component it under going a lot of change right now and is a bit of a monster already. 

Not really a huge deal, but as a follow up to this PR we might want to rename some of the CSS classes to make them more generic avoid confusion, e.g.,  the "image" in `block-editor-global-styles-background-panel__image-tools-panel-item`

Additionally, the `block-editor-global-styles-background-panel__inspector-media-replace-container` element used to only wrap the image components and uses the `is-open` class to control the selected state of the media replace container. Maybe that could be isolated to the controls as well.





----

_Not a blocker, but for disclosure, I'll add the following:_

For me, this redesign exposes more clearly how gradient does not play well with the other children 😄 

In the coming weeks I'll be trying to reconcile background gradient and background image values, which, ideally, should be merged under the `background-image` CSS property.

Right now `color.gradient` is output as the value of `background`, which [overwrites](https://github.com/WordPress/gutenberg/issues/60401) other styles and is generally a real pain.

I expect the controls will have to be tweaked to accommodate this later, e.g., sharing position/repeat/size.

packages/block-editor/src/components/global-styles/background-panel.js
ramonjd: PR here: https://github.com/WordPress/gutenberg/pull/64242

jameskoster:
> we should also not call this tool panel "Elements" since these are not elements. Not sure what to use, since "Background" would be redundant with the entire panel name.

Does this panel need a title? It seems it's only really there to facilitate the Reset action via ToolsPanel. This could potentially move to the popover instead, similar to the Clear action in the color picker:

Screenshot 2024-08-05 at 10 46 47


If it is necessary, maybe "Fill" could work.

jasmussen:
> we should also not call this tool panel "Elements" since these are not elements. Not sure what to use, since "Background" would be redundant with the entire panel name.

The panel in question:

Screenshot 2024-08-05 at 12 40 44

And the precedence:

Screenshot 2024-08-05 at 12 39 24

Since the behavior here is strictly related to the ToolsPanel component, it should have a header, including in the flyout. I agree "Elements" is a bit confusing. Some other ideas:

* Properties
* Options
* Settings
* Tools
* Attributes
* Controls
* Details

Perhaps simply "Tools"?

jasmussen:
Separate to this PR, but noticing. There's a typographic widow in this drilldown description:

Screenshot 2024-08-05 at 12 41 22

I thought that was handled on the Text components level, but perhaps that component isn't used here? Should it be? (Again, separate PR to not balloon this otherwise wonderful PR!)

jasmussen:
Very nice work @amitraj2203, thank you!

packages/block-editor/src/components/global-styles/background-panel.js
amitraj2203: Thanks for the feedback, I have addressed it

packages/block-editor/src/components/global-styles/style.scss
amitraj2203: Thanks for the feedback, I have addressed it

packages/block-editor/src/components/global-styles/style.scss
amitraj2203: Hi @mirka @mtias,

I have made the changes to wrap both the "Image" and "Color" items with ItemGroup. However, after resolving the recent merge conflicts, it has become confusing. If you can help here, I would appreciate it.

ramonjd:
Sorry, I pushed a couple of bug fixes to the background image panel, which means this PR will require a rebase. 🙏🏻  Let me know if you run into trouble.

amitraj2203:
Thank You @ramonjd @jasmussen @mirka.
I've resolved the conflicts now. I apologize for the delay, it was a busy week. Could you please review the changes at your convenience?

ramonjd:
Thanks for the ping! 

Things are working as I'd expect in the editor. There are a couple of minor points that might need some feedback from the design folks.

1. Previously, there was no preview icon placeholder. Now, there is a rectangular-like shape and its dimensions don't match the image preview icon. The discussion to remove the placeholder icon was on [this PR](https://github.com/WordPress/gutenberg/pull/60264#issuecomment-2024868647). I guess it was re-added for balance with the color control below, but it affects all controls (global styles and block level).
2. There's a white line at the bottom of the active state (when the popover panel is open).


### Trunk

| Empty | With image |
|--------|--------|
| Screenshot 2024-08-19 at 11 31 46 AM | Screenshot 2024-08-19 at 11 31 54 AM | 



### This PR

| Empty | With image |
|--------|--------|
| Screenshot 2024-08-19 at 11 29 14 AM | Screenshot 2024-08-19 at 11 30 39 AM | 



Also, as [mentioned](https://github.com/WordPress/gutenberg/pull/63216#discussion_r1703467718), I think it would be worth looking for abstraction opportunities, given that the component now forks the UI based on context, e.g., abstracting out the ColorPanelDropdown from `background-panel.js`. 

This could be done in a follow up though as it might bloat this PR.

jasmussen:
> There's a white line at the bottom of the active state (when the popover panel is open).

I think this is an existing issue with the ItemGroup, so not something that need be fixed in this PR.

jasmussen:
Would appreciate dev help and reviews, this would be a nice one to land! And thanks for contributing!

packages/block-editor/src/components/global-styles/style.scss
```diff
@@ -158,6 +171,16 @@
 	cursor: pointer;
 	background: transparent;
 	border: none;
+
+
```
tyxla: Three consecutive empty new lines are not recommended, we usually [go with two](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/#structure):

> Add two blank lines between sections and one blank line between blocks in a section.

packages/block-editor/src/components/global-styles/style.scss
```diff
@@ -147,6 +152,14 @@
...
 }
 
+.block-editor-global-styles-background-panel__color-tools-panel-item {
+
+	button.components-button {
+
+		padding: 0;
+	}
+}
+
```
tyxla: Leading/trailing new lines inside a section are not redundant:

```suggestion
.block-editor-global-styles-background-panel__color-tools-panel-item {
	button.components-button {
		padding: 0;
	}
}

```

packages/edit-site/src/components/global-styles/root-menu.js
```diff
@@ -33,10 +34,20 @@ function RootMenu() {
 	const hasShadowPanel = true; // useHasShadowPanel( settings );
 	const hasDimensionsPanel = useHasDimensionsPanel( settings );
 	const hasLayoutPanel = hasDimensionsPanel;
+	const hasBackgroundPanel = true; // useHasBackgroundPanel( settings );
```
tyxla: Can you elaborate on why we are hardcoding this?

tyxla (commented):
cc @WordPress/gutenberg-components since this overrides some component styles and it's worth auditing the usage.

packages/block-editor/src/components/global-styles/style.scss
```diff
@@ -147,6 +152,14 @@
 	}
 }
 
+.block-editor-global-styles-background-panel__color-tools-panel-item {
+
+	button.components-button {
```
ciampo: While style overrides are 

Instead of using `button.components-button` (which is meant to be a private API of the components package), could we assign a custom classname to the `button` itself and use that to apply style overrides?

Also, do we need to use a `Button` component, if we're going to override its styles? Should we consider just use a vanilla `button` instead?

---

(These comments apply to all instances of style overrides for `Button` in this file)

packages/block-editor/src/components/global-styles/style.scss
```diff
@@ -83,13 +83,18 @@
 	}
 
 	.block-editor-global-styles-background-panel__image-tools-panel-item {
+		padding: 0;
 		flex-grow: 1;
 		border: 0;
 		.components-dropdown {
```
ciampo, line 89: I know it was not introduced by this PR, but — as also mentioned about `.components-button` — `.components-*` classnames re not meant to be used outside of the components package. We can use custom classnames instead.

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -845,14 +1051,57 @@ export default function BackgroundPanel( {
...
-								setIsDropDownOpen( false );
 								resetBackground();
 							} }
-							onRemoveImage={ () => setIsDropDownOpen( false ) }
 						/>
 					) }
 				</ToolsPanelItem>
-			</div>
+				{ shouldShowBackgroundColorControls &&
+					path === '/background' && (
```
ciampo: Excuses in case I'm missing the context, but why do we need to check the `path` ?

Even if we needed to, this inline check in the middle of the JSX output could be easily overlooked. If we need this check, I would probably extract it as a prop of the `Screen` itself

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -845,14 +1051,57 @@ export default function BackgroundPanel( {
...
+									},
+									hasGradientColors && {
+										key: 'gradient',
+										label: __( 'Gradient' ),
+										inheritedValue: gradient,
+										setValue: setGradient,
+										userValue: userGradient,
+										isGradient: true,
+									},
+								].filter( Boolean ) }
```
ciampo: Potentially premature optimisation, but as of now the `tabs` array is being re-creater every time the component re-renders.

ciampo (commented):
I agree with @mirka that the background panel (and probably the rest of the global style screens) are pushing the boundaries of how a few components are supposed to be used (`ItemGroup`, `Dropdown`, `ToolsPanel`), showing that those components have some limitations that could be improved.

Apart from that (and some inline comments), I personally find the dropdown that opens when clicking "Add background image" a bit awkward, since most dropdowns / popovers would open to the left of the sidebar

Screenshot 2024-08-19 at 18 38 20

packages/block-editor/src/components/global-styles/background-panel.js
amitraj2203: The BackgroundColorControl is also visible in block-level global styles. For example, if we remove this check, it will also be visible in the Group block's global styles.
image

Can you please tell me exactly where you want this to be moved as a prop?

packages/block-editor/src/components/global-styles/style.scss
amitraj2203: Should I make the change in this PR only, or a separate PR would be good for it?

packages/edit-site/src/components/global-styles/root-menu.js
amitraj2203: I am not sure exactly why, but the value of `backgroundImage` and `backgroundSize` is false here 🤔 .

packages/block-editor/src/components/global-styles/style.scss
ciampo: Happy for it to be a separate, follow-up PR.

The main changes that would be needed:
- understand why we need to set a `36px` height on `.components-dropdown`. Ideally, we should use the default button height. If not, let's understand why, and at most use the `renderToggle` prop to pass differently sized button;
- avoid overriding color and focus styles of `.components-button`, and only use the component props to tweak those aspects (mainly the `variant` prop)
- passing a custom `className` to the components instances, and then using that custom classname in the styles file;

packages/edit-site/src/components/global-styles/root-menu.js
ciampo: That can be investigated separately, I guess? Let's make sure we revert this change before merging

packages/block-editor/src/components/global-styles/background-panel.js
ciampo: This component shouldn't really know about global styles routing, even more so that it's defined in a different package.

Instead, it could expose a boolean prop (something like `showBackgroundControl` or any other name that makes sense). Is showing those controls an exception? Then the default value of the boolean prop should be `false`. Otherwise, it should be `true`.



Let's assume it's `false` by defatul — this means that we should set it to `true` only where we need to show it, ie. in the [background screen](https://github.com/WordPress/gutenberg/blob/85db58f2de6e6e353c4cad8b0c3a5e46e610409a/packages/edit-site/src/components/global-styles/screen-background.js) via the [`<BackgroundPanel />` component](https://github.com/WordPress/gutenberg/blob/d669eb661c69af31816ead9359c5986cffcea301/packages/edit-site/src/components/global-styles/background-panel.js) (and not sure where else).

amitraj2203: I have tried to implement it, could you please check again?

ciampo: Yup, you did follow my advice above.

However, I'd like to delegate review & advice on this aspect to someone with more experience with this part of the codebase, since I get the feeling that what I suggested is not optimal either.

In general, I get the feeling that, if we want the background panel to be rendered in a few places (global styles root, single block screen in global styles, and potentially anywhere else in the sidebar), we should generalize its logic a bit more — or delegate it to the parent react component.

For example, I wonder if we should make tweaks to the [`useHasBackgroundPanel` hook](https://github.com/WordPress/gutenberg/blob/7b8110a652532bf3df08c11ca73861fc34db0e09/packages/block-editor/src/components/global-styles/background-panel.js#L91-L93) (or create a new hook), since:

- it only seems to care about global settings, even if it looks like this panel needs to be used in block level screens too
- it only references the background image, while we need to also check for the background color (the main object of this conversation)


cc @ramonjd since it looks like he's done a bunch of work in this area of the codebase

packages/edit-site/src/components/global-styles/root-menu.js
ciampo: Related: https://github.com/WordPress/gutenberg/pull/63216#discussion_r1740569717

Basically, I wonder if `useHasBackgroundPanel` should be updated and/or the logic of this panel refactored to include:

- the fact that there are controls for both the background image and the background color, which may depend on separate settings
- the fact that this panel can be used both for global site settings, and for block-level settings

packages/edit-site/src/components/global-styles/root-menu.js
```diff
@@ -33,10 +35,20 @@ function RootMenu() {
 	const hasShadowPanel = true; // useHasShadowPanel( settings );
 	const hasDimensionsPanel = useHasDimensionsPanel( settings );
 	const hasLayoutPanel = hasDimensionsPanel;
+	const hasBackgroundPanel = useHasBackgroundPanel( settings );
```
ramonjd, line 38: The background panel doesn't show at all for me at the moment. I think we should be using `rawSettings` here, not the return value of `useSettingsForBlockElement()`

```suggestion
	const hasBackgroundPanel = useHasBackgroundPanel( rawSettings );
```

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -722,6 +869,75 @@ export default function BackgroundPanel( {
...
+			( { gradient } ) => gradient === gradientValue
+		);
+		return gradientObject
+			? 'var:preset|gradient|' + gradientObject.slug
+			: gradientValue;
+	};
+
+	const decodeValue = ( rawValue ) =>
+		getValueFromVariable( { settings }, '', rawValue );
+	const shouldShowBackgroundColorControls = useHasBackgroundPanel( settings );
```
ramonjd, line 904: Hmm, I think there's an unfortunate naming clash. Wouldn't we be more interested in the function of the same same from the color panel?

https://github.com/WordPress/gutenberg/blob/9cd986ca8e7331cfb8302a85d1070a5bacf2992c/packages/block-editor/src/components/global-styles/color-panel.js#L100-L100

I think one or both of the functions names could be changes, e.g. `useHasBackgroundImagePanel` and `useHasBackgroundColorPanel`

I'll get a quick PR up for that 👍🏻 

In the mean time we could import the right function and rename it, e.g., `import { useHasBackgroundPanel as useHasBackgroundColorPanel } from './color-panel';`

packages/block-editor/src/components/global-styles/background-panel.js
ramonjd: Thanks for helping out here @ciampo 

> In general, I get the feeling that, if we want the background panel to be rendered in a few places (global styles root, single block screen in global styles, and potentially anywhere else in the sidebar), we should generalize its logic a bit more — or delegate it to the parent react component.

The latter is definitely something worth to think about. 

The proposal is to only add the background color picker to the background panel on for root global styles. 

I suspect it'd be easier to test, and easier to use if `<BackgroundPanel />` were generalized and the image/color/etc components abstracted and rendered as children of the former.

I'm not suggested a complete refactor of this PR, but it does need more testing. 


> it only seems to care about global settings, even if it looks like this panel needs to be used in block level screens too

I think that's consistent with other `use*` functions in sibling style panels. So, if theme.json turns these settings off, they're off everywhere. Unless I'm missing something 🤔 

> it only references the background image, while we need to also check for the background color (the main object of this conversation)

There's also a function of the same name `useHasBackgroundPanel` in the color panel that we should be using. https://github.com/WordPress/gutenberg/blob/9cd986ca8e7331cfb8302a85d1070a5bacf2992c/packages/block-editor/src/components/global-styles/color-panel.js#L100-L100

I'm going to get a PR up to fix the naming clash, but yeah, in order to determine whether to show the background panel at all, it should take into account all the properties.

It might be easier to reason about if things were abstracted.

packages/block-editor/src/components/global-styles/background-panel.js
```diff
@@ -697,6 +703,146 @@ function BackgroundToolsPanel( {
...
+				className="block-editor-panel-color-gradient-settings__color-name"
+				title={ label }
+			>
+				{ label }
+			</FlexItem>
+		</HStack>
+	</Item>
+);
+
+function ColorPanelDropdown( {
```
ramonjd, line 751: I've mentioned it before, but I think this file is getting too large and probably needs some composition, e.g., 

```jsx
<BackgroundPanel>
      <BackgroundImageControls />
      <BackgroundColorControls />
</BackgroundPanel>
```

ramonjd (commented):
Thanks for your continued efforts on this @amitraj2203 

It's not a straight forward task, so I appreciate your perseverance. 🙇🏻 

I want these types of changes to move forward - I agree 100% that grouping controls that affect the background makes sense. 

I'm just not sold that we're doing things in the right order, that is, reorganizing the UX on top of a styles system that doesn't support it.

For example, it's really confusing that I can't see my gradient background color behind the image:

Screenshot 2024-09-03 at 10 42 04 AM

It's because we haven't yet fixed the `background`/ `background-image` clash. 

See the item under **Setting background-image overrides background gradient** in the [background follow up tasks](https://github.com/WordPress/gutenberg/issues/54336).

In a perfect world, this styles clash would have been fixed for 6.7, but time/resources were constraints. 

With it fixed, the new UI proposed here would be the perfect complement. 

I think we should concentrate on addressing the gradient/background-image CSS issue — which, in my opinion, is a fairly large bug and completely unintuitive to both developers and users — before we expose its limitation to users more obviously in the UX. 

That's just my 2c. 😄 

Also, on a more trivial point, if we're going with the square placeholder icon, I still think we should iron out the differences between its dimensions and the uploaded image's icon: https://github.com/WordPress/gutenberg/pull/63216#issuecomment-2295516133

packages/block-editor/src/components/global-styles/background-panel.js
ramonjd: Renaming PR: https://github.com/WordPress/gutenberg/pull/64993

jasmussen:
> I think we should concentrate on addressing the gradient/background-image CSS issue — which, in my opinion, is a fairly large bug and completely unintuitive to both developers and users — before we expose its limitation to users more obviously in the UX.

Well, both. The disorganization of controls is certainly an issue, but it's already been an issue across Cover, Group, and the new unified background tools. This is a large undertaking, and one of the most important aspects of _this PR_, or perhaps more specifically, of #63096, is to move the background tools out of "Layout" and into the global styles inspector as a top level item. There are some gotchas with this, but this organization is an important step that ideally we find a path towards.

packages/block-editor/src/components/global-styles/background-panel.js
ciampo: > I think that's consistent with other `use*` functions in sibling style panels. So, if theme.json turns these settings off, they're off everywhere. Unless I'm missing something 🤔

That makes sense — and shows how little experience I have with this part of the codebase 😓  

> The proposal is to only add the background color picker to the background panel on for root global styles.
> 
> I suspect it'd be easier to test, and easier to use if `<BackgroundPanel />` were generalized and the image/color/etc components abstracted and rendered as children of the former.
> 
> It might be easier to reason about if things were abstracted.

I trust your instinct on what the best solution should be.

But I definitely get the impression that the current implementation needs some re-architecting to make sure it can work well regardless of where it's rendered — root global styles screen, block-level global styles screen, or anywhere else in the inspector sidebar

ramonjd:
> The disorganization of controls is certainly an issue, but it's already been an issue across Cover, Group, and the new unified background tools. 

Thanks @jasmussen Definitely not disputing the need to reorganize the controls - I only wanted to communicate that the way the underlying `background` styles are expressed in CSS is buggy and fixing it should share priority. Admittedly, given time constraints a fix for that won't make it for 6.7 - I had planned it in, but won't get to it unfortunately. 

> is to move the background tools out of "Layout" and into the global styles inspector as a top level item

My vote would be to only move the background image tool as a first iteration - it might make integrating color later easier to implement and test. 🤷🏻

jasmussen:
> My vote would be to only move the background image tool as a first iteration - it might make integrating color later easier to implement and test. 🤷🏻

Is this something you can help with? Thanks for the clarification!

ramonjd:
> Is this something you can help with? Thanks for the clarification!

I don't have much development bandwidth for 6.7 unfortunately, but I think @amitraj2203's work so far has been great to illustrate how the controls can work (thank you @amitraj2203).

In my view, we could split this PR into the things it's trying to achieve:

1. At a minimum, isolate the work in this PR that moves the background image to the background navigation item for top-level global styles. This was the original state before it was moved into Layout. 
2. Once that's done, [rework the component](https://github.com/WordPress/gutenberg/pull/63216#discussion_r1741314097) so that controls can be better managed. This will be particularly beneficial to manage conditional rendering, for example, in the case where a theme supports background colors but not background images in the theme.json settings.
3. Create and integrate the background color component for this section. Most of the work has been done in this PR for that. 		 


@amitraj2203 let me know if this makes sense to you - before 6.7 Beta 1 I should get some time to help out with 1 and 2.

ramonjd:
Following up on this comment: https://github.com/WordPress/gutenberg/pull/63216#issuecomment-2330295914

I've done points 1 and 2 in this PR:

- https://github.com/WordPress/gutenberg/pull/65304

If that's okay, I think it will make merging background color and image controls (from this PR) a little less complex, at least it'll be fewer lines of code 🤞🏻 

Thank you!

mtias:
Good call @ramonjd

amitraj2203:
> Create and integrate the background color component for this section. Most of the work has been done in this PR for that.

@ramonjd I have Raised a [PR](https://github.com/WordPress/gutenberg/pull/66736) for the above point.

ramonjd:
Closing this as it's being taken care of in:

- https://github.com/WordPress/gutenberg/pull/65304
- https://github.com/WordPress/gutenberg/pull/66736

PR #63220, opened 2024-07-07
Block Editor: Remove CSS hack for Internet Explorer 11

t-hamano:
Part of #58034

This PR is the final one to remove fallback styles for IE11 for stylesheets.

## What?
Remove the CSS code for IE11 that exists in the `block-editor` package. The styles that will be removed are all related to the block toolbar.

## Why?
Support for IE11 has been [officially removed in WordPress 5.8](https://make.wordpress.org/core/2021/04/22/ie-11-support-phase-out-plan/). And since the Gutenberg plugin no longer supports WordPress 5.8, we can remove the fallback code for IE11.

## Testing Instructions

The layout of the block toolbar should remain the same.

### DIFF
--- a/packages/block-editor/src/components/block-toolbar/style.scss
+++ b/packages/block-editor/src/components/block-toolbar/style.scss
@@ -24,8 +24,6 @@
 	.components-toolbar-group,
 	.components-toolbar {
 		background: none;
-		// IE11 has thick paddings without this.
-		line-height: 0;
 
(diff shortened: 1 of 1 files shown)


jasmussen (approved):
What's a good way to test this?

t-hamano:
@jasmussen Thanks for the review!

> What's a good way to test this?

I think it's enough to just make sure that the block toolbar is displayed in the correct layout as before 👍

PR #63222, opened 2024-07-07
Fix background height and padding in non-iframe editor canvas

### LINKED ISSUE #63110: Post Editor: Background issue in non-iframe editor
### Description

Note: This issue first occurs in WP6.6.

If the post editor is NOT an iframe, e.g. contains a version 2 block, two issues with the background will arise:

- There is not enough space at the bottom of the content
- The background color disappears halfway through

## 1. There is not enough space at the bottom of the content

There should be [`40vh` of space reserved](https://github.com/WordPress/gutenberg/blob/e916313652cdfa069e5d7fb5aca0dd26fdea0141/packages/edit-post/src/components/layout/index.js#L141), but there is no space at all, which means you cannot click the space below the last block to insert a new block.

### WP6.5

Below the last paragraph, there is a padding of 40vh, which you can click to insert a new block.

image

### WP6.6 or the Latest Gutenberg

It looks like there is a little bit of space below the paragraph, but this is the browser default margin, not the editor canvas padding, so clicking here will do nothing.

image 

### WP6.5

The background color fills the entire canvas.

image

### WP6.6 or the Latest Gutenberg

When scrolling down with a lot of content, the background color disappears.

image

- Enable the TT1 theme, which has a more visible background color.
- To make the post editor not an iframe, do one of the following:
  - Activate the Jetpack plugin. This plugin contains version 1 blocks.
  - Change the apiVerson of one of the core blocks to 2 via the following hook:
    ```php
    function example_filter_metadata_registration( $settings, $metadata ) {
    	if ( 'core/paragraph' === $metadata['name'] ) {
    		$settings['api_version'] = 2;
    	}
    	return $settings;
    };
    add_filter( 'block_type_metadata_settings', 'example_filter_metadata_registration', 10, 2 );
    ```
- Access the post editor and enter more content.
- Scroll to the very bottom of the content.

_No response_

### Environment info

_No response_

Yes

Yes

t-hamano:
Fixes #63110
Related to #62940

## What?

This PR fixes two issues with the background when **the post editor is not an iframe but does not have a meta box**:

### There is not enough space at the bottom of the content

issue_1

### The background color disappears halfway through

issue_2

## Why?

In #62940, the post editor was styled as follows to fix the triple scroll issue:

```css
// From
.editor-visual-editor {
	flex: 1 0 auto;
	height: auto;
}

// To
.edit-post-layout.has-metaboxes .edit-post-visual-editor {
	flex: 1 0 auto;
	height: auto;
}
```

This style works in the non-iframe editor where meta boxes (custom fields) are enabled. However, the post editor will not become an iframe if there is a v2 or lower block present, even if it does not have a meta box. In this case, the `.has-metaboxes` class does not exist, so the editor defaults to `height:100%`, which seems to cause the issue mentioned above.

## How?

I used `.is-iframed` instead of `.has-metaboxes` class to accurately determine if the editor is an iframe or not.

## Testing Instructions

- Enable the TT1 theme, which has a more visible background color.
- Activate the Jetpack plugin. This plugin won't add a meta box to the post editor, but because the plugin includes version 1 blocks, the post editor will no longer be an iframe.
- Access the post editor and enter more content.
- Scroll to the very bottom of the content.
- The background color should extend to the bottom of the canvas and have sufficient bottom padding.

Also, there have been some scrollbar fixes in this area in the past, so please make sure we have not run into regressions or unwanted scrollbars in the following cases:

- Iframe editor
- Mobile or tablet preview
- Pattern editor
- Add lots of notices (`wp.data.dispatch( 'core/notices' ).createInfoNotice( 'Notice' );`)

**Note:** However, there is one issue that this PR doesn't solve: when you open the Pattern Editor in a non-iframe editor, the background gets cut off.

pattern-editor

I believe the root cause of this issue is based on the fact that when the post editor is not an iframe, the pattern editor is not an iframe either. I believe the pattern editor should be an iframe regardless of whether the post editor is an iframe or not, just like the tablet and mobile views. However, since this change is impactful, I thought it should be addressed in a follow-up.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/cb45b418-eb85-4f45-aca3-38b94bbff88f

### DIFF
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -1,4 +1,4 @@
-.edit-post-layout.has-metaboxes .edit-post-visual-editor {
+.edit-post-visual-editor:not(.is-iframed) {
 	flex: 1 0 auto;
 	height: auto;
 }


t-hamano:
This issue first appeared in WP6.6 and I expect it will affect many sites, so hopefully it will be backported to WP6.6.

talldan:
Not sure what I'm doing wrong, but even with jetpack installed or one of the core blocks manually set to v2, the editor is still iframed.
Screenshot 2024-07-08 at 12 11 25 pm

Maybe I'll try hacking the code to force a non-iframed canvas, though this has me concerned that there's another bug.

stokesman:
> Not sure what I'm doing wrong

I think the theme also has to be a non-block theme?

t-hamano:
Sorry, I missed it, but this problem may not be reproducible with block themes.

Because if the Gutenberg plugin is enabled and it is a block theme, the `useShouldIframe` hook is always `true`:

https://github.com/WordPress/gutenberg/blob/6d3282e0ba3fb56bb4e55cb797eb5f0d5dae8795/packages/edit-post/src/components/layout/use-should-iframe.js#L36

You should be able to reproduce the issue with Jetpack enabled in the following environments:

- WP6.6 (without Gutenberg plugin) and block/classic theme
- Gutenberg plugin enabled and classic theme

talldan (approved):
Thanks for fixing this @t-hamano. In my testing it solves the issue and doesn't seem to regress any other behaviors.

Agree, this is an important fix for 6.6. 👍

t-hamano:
Thank you everyone for your reviews!

enricobattocchi:
Hey @t-hamano , it looks like this has introduced a problem with tablet/mobile mode when a metabox with a lot of content is present.
If you e.g. have Yoast SEO (or ACF and you add field group with 10 or more fields), switching to tablet or mobile mode will make the metabox occupy the whole height, hiding the content completely: the only way to access the content then is to collapse the metabox.
Before this change, switching to tablet/mobile made _the content_ occupy the whole height, and you had to scroll to reveal the metaboxes.
You can see the bug in action in this video: https://www.loom.com/share/4982c8fa41024d4c85fc54da860e758a

t-hamano:
@enricobattocchi Thanks for the report.

Yes, I have confirmed that this PR does indeed cause issues in mobile/tablet views. I have summarized the cause in #63708.

I would like to explore a more ideal solution.

PR #63223, opened 2024-07-07
Block Bindings: Remove "Site Updated" notification for post entity changes

artemiomorales:
## What?

Currently, the block editor erroneously shows a `Site Updated` notification when modifying a block connected to post metadata via block bindings as reported in https://github.com/WordPress/gutenberg/issues/62236. This PR fixes that.

## Why?

We don't want to show users misleading or incorrect information.

## How?

It introduces a check to see if the current entity changes are of the kind `postType`, and it only displays the `Site Updated` message if they are not.

## Testing Instructions

<details>

```php

	register_meta(
		'post',
		'text_field',
		array(
			'show_in_rest'      => true,
			'single'            => true,
		)
	);
}
```

</details>

<details>

<summary>2. In the post editor, add a paragraph block bound to the custom field using the Code Editor</summary>

```
<p>Paragraph content</p>
<!-- /wp:paragraph -->
```

</details>
 
3. Press the Save button.
4. See that the `Site Updated` message _does not_ appear, as expected.

Please also test other scenarios in the site editor to ensure the `Site Updated` message continues to show as normal when expected.

 ## Screenshots or screencast

### Before

site-updated-bug

### After

Screenshot 2024-07-07 at 6 42 55 PM

### DIFF
--- a/packages/editor/src/store/private-actions.js
+++ b/packages/editor/src/store/private-actions.js
@@ -153,6 +153,7 @@ export const saveDirtyEntities =
 		close?.( entitiesToSave );
 		const siteItemsToSave = [];
 		const pendingSavedRecords = [];
+		let showPostUpdatedMessage = false;
 		let showSiteUpdatedMessage = false;
 		entitiesToSave.forEach( ( { kind, name, key, property } ) => {
 			if ( 'root' === kind && 'site' === name ) {
@@ -182,7 +183,9 @@ export const saveDirtyEntities =
 				// This flag allows us to avoid showing the 'site updated' message
 				// in scenarios related to entity changes when the message shouldn't be
 				// triggered, like when modifying post meta via block bindings.
-				if ( kind !== 'postType' ) {
+				if ( kind === 'postType' ) {
+					showPostUpdatedMessage = true;
+				} else {
 					showSiteUpdatedMessage = true;
 				}
 			}
@@ -213,19 +216,30 @@ export const saveDirtyEntities =
 					registry
 						.dispatch( noticesStore )
 						.createErrorNotice( __( 'Saving failed.' ) );
-				} else if ( showSiteUpdatedMessage ) {
-					registry
-						.dispatch( noticesStore )
-						.createSuccessNotice( __( 'Site updated.' ), {
-							type: 'snackbar',
-							id: saveNoticeId,
-							actions: [
-								{
-									label: __( 'View site' ),
-									url: homeUrl,
-								},
-							],
-						} );
+				} else {
+					if ( showSiteUpdatedMessage ) {
+						registry
+							.dispatch( noticesStore )
+							.createSuccessNotice( __( 'Site updated.' ), {
+								type: 'snackbar',
+								id: saveNoticeId,
+								actions: [
+									{
+										label: __( 'View site' ),
+										url: homeUrl,
+									},
+								],
+							} );
+					}
+
+					if ( showPostUpdatedMessage ) {
+						registry
+							.dispatch( noticesStore )
+							.createSuccessNotice( __( 'Post updated.' ), {
+								type: 'snackbar',
+								id: saveNoticeId,
+							} );
+					}
 				}
 			} )
 			.catch( ( error ) =>
--- a/packages/editor/src/store/private-actions.js
+++ b/packages/editor/src/store/private-actions.js
@@ -153,9 +153,11 @@ export const saveDirtyEntities =
 		close?.( entitiesToSave );
 		const siteItemsToSave = [];
 		const pendingSavedRecords = [];
+		let showSiteUpdatedMessage = false;
 		entitiesToSave.forEach( ( { kind, name, key, property } ) => {
 			if ( 'root' === kind && 'site' === name ) {
 				siteItemsToSave.push( property );
+				showSiteUpdatedMessage = true;
 			} else {
 				if (
 					PUBLISH_ON_SAVE_ENTITIES.some(
@@ -176,6 +178,13 @@ export const saveDirtyEntities =
 						.dispatch( coreStore )
 						.saveEditedEntityRecord( kind, name, key )
 				);
+
+				// This flag allows us to avoid showing the 'site updated' message
+				// in scenarios related to entity changes when the message shouldn't be
+				// triggered, like when modifying post meta via block bindings.
+				if ( kind !== 'postType' ) {
+					showSiteUpdatedMessage = true;
+				}
 			}
 		} );
 		if ( siteItemsToSave.length ) {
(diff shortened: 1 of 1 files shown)


packages/editor/src/store/private-actions.js
```diff
@@ -176,6 +178,13 @@ export const saveDirtyEntities =
 						.dispatch( coreStore )
 						.saveEditedEntityRecord( kind, name, key )
 				);
+
+				// This flag allows us to avoid showing the 'site updated' message
+				// in scenarios related to entity changes when the message shouldn't be
+				// triggered, like when modifying post meta via block bindings.
+				if ( kind !== 'postType' ) {
+					showSiteUpdatedMessage = true;
+				}
```
artemiomorales: Previously, the logic in this code assumed that an entity change must be a site change; this just introduces a flag to circumvent the `Site Updated` message if the entity change was of kind `postType`.

I think it works, but I’m not sure this is the most robust solution.

artemiomorales (commented):
As noted in the comment below, I'm not sure if this fix works in all scenarios or is the best solution. Let me know what you think 🙏

SantosGuillamot:
Before reviewing the code, I have a conceptual question: In which scenarios do we want to show the "Site updated" message, and which others should we show "Post updated" or other things?

For example, if I'm on a page with a query loop and modify the custom fields of other posts, which message should appear?

SantosGuillamot:
I just came across this comment from Riad: [link](https://github.com/WordPress/gutenberg/pull/62284#discussion_r1626167036).

> Should we show "site updated" when we change just a single entity (like a page or post or a template) or something?
> 
> Maybe we should try to be smarter here. If the "pendingRecords" is a single entity we just show ${ entityTitle } saved successfully and if there are multiple records we should "Site updated" or something.

artemiomorales:
> Before reviewing the code, I have a conceptual question: In which scenarios do we want to show the "Site updated" message, and which others should we show "Post updated" or other things?
>
> For example, if I'm on a page with a query loop and modify the custom fields of other posts, which message should appear?

I think if you're in the site editor and update a post inside a query loop, then it should say "Post Updated."
If you also update some site settings, then it should say both "Post Updated" and "Site Updated."

I just pushed a commit as a test that does just that, but now the "Post Updated" message appears twice in the post editor.

Screenshot 2024-07-09 at 11 18 09 PM

artemiomorales:
> I just came across this comment from Riad: https://github.com/WordPress/gutenberg/pull/62284#discussion_r1626167036.
>
>> Should we show "site updated" when we change just a single entity (like a page or post or a template) or something?
>> 
>> Maybe we should try to be smarter here. If the "pendingRecords" is a single entity we just show ${ entityTitle } saved successfully and if there are multiple records we should "Site updated" or something.

@youknowriad Do you think we should still take this approach? It looks like the "Template Updated" message is handled somewhere else. Maybe checking explicitly for 'postType' as in this PR at the moment is good enough? 

I'll explore doing the smarter option too, but wanted to get your thoughts on this.

youknowriad:
Thanks for taking another stab at this, I like that that it takes a more generic approach but I noticed a few things: We basically have two different save flows that result in different logic for the notices:

 - Single entity save flow uses the `savePost` action which uses `getNotificationArgumentsForSaveSuccess` util to get the right notification. This seems to work fairly well. 
 - Multi-entity save flow uses the `saveDirtyEntity` action which doesn't use `getNotificationArgumentsForSaveSuccess` util and is kind of broken right now in general. 

1- I think ideally the "save process" notification handling (maybe even the save action itself but this is out of scope) should be the same. It would be good to use the exact same util for both. Whether that means making `getNotificationArgumentsForSaveSuccess` accept multiple entities or having a wrapper around that function, I don't know and don't care much.

2- It's still unclear what the message should be for me in the case of multiple entities. What happens when I save just the "site" entity, what happens when I save "template" and "post", what happens when I save "template and template part", what happens when I save "site" and "post"... I think some help from designers mapping out the different cases would help us implement the right solution. cc @jameskoster @jasmussen 

3- While testing this PR, I noticed random things like when saving "template" and "site", I end up with two notices : "Template updated" and "post updated", why?

jameskoster:
For single-saves, my take is that what we have works reasonably well:

* Post / Pages: "$post_type updated [View]"
* Template: "Template updated"
* Pattern/Template part: "$post_type updated"

I wonder if we should expand that to cover site-specific data, e.g:

* "Site title updated"
* "Site logo updated"

Or whether a singular "Site updated" is adequate. No strong opinion there.

For multi-entity, to differentiate from the 'site' item, perhaps a more generic "All changes saved" could work?

artemiomorales:
> 1- I think ideally the "save process" notification handling (maybe even the save action itself but this is out of scope) should be the same. It would be good to use the exact same util for both. Whether that means making getNotificationArgumentsForSaveSuccess accept multiple entities or having a wrapper around that function, I don't know and don't care much.
 
 Ok, thanks for the guidance! I can try refactoring the `saveDirtyEntity` action to use `getNotificationArgumentsForSaveSuccess`.

> 2- It's still unclear what the message should be for me in the case of multiple entities. What happens when I save just the "site" entity, what happens when I save "template" and "post", what happens when I save "template and template part", what happens when I save "site" and "post"... I think some help from designers mapping out the different cases would help us implement the right solution.

> For multi-entity, to differentiate from the 'site' item, perhaps a more generic "All changes saved" could work?

I can change the multi-entity save notification to be "All changes saved."

> For single-saves, my take is that what we have works reasonably well:
>
> - Post / Pages: "$post_type updated [View]"
> - Template: "Template updated"
> - Pattern/Template part: "$post_type updated"
>
> I wonder if we should expand that to cover site-specific data, e.g:
> - "Site title updated"
> - "Site logo updated"
>
> Or whether a singular "Site updated" is adequate. No strong opinion there.

My initial reaction is to keep the first 3 (post/pages, template, and pattern/template part) as specified, and just use "Site updated" for other changes. I'll aim to get something working so we can take a look.

SantosGuillamot:
Sharing same comment as in the issue:

The changes to trigger the save panel after editing post meta in a post have been reverted in [this pull request](https://github.com/WordPress/gutenberg/pull/63412). I guess this is not an issue anymore and we just have to ensure that whatever we build doesn't trigger this again.

I've opened [this new issue](https://github.com/WordPress/gutenberg/issues/63425) to gather all the context and make decisions before working on a new solution.

artemiomorales:
I'm going to close this for now, and we can reopen this or explore further whenever we discussions continue around notifications for post entity changes. Here's a [related comment](https://github.com/WordPress/gutenberg/issues/65683#issuecomment-2387638879) regarding similar ideas around post entity changes.

PR #62673, opened 2024-06-19
Move dep to dev deps.

DaniGuardiola:
## What?

Accidentally forgot to apply @jsnajdr's suggestion here before merging: https://github.com/WordPress/gutenberg/pull/62476#discussion_r1635779467

## How?

Move the dep to dev deps in the `package.json` file.

PR #63224, opened 2024-07-08
Quality: Remove postcss-local-keyframes from dependencies

t-hamano:
Related to #62673

## What?

I found out that `postcss-local-keyframes` exists in both `dependencies` and `devDependencies`, so I deleted the one that exists in `dependencies`.

## How?

- Manually delete `postcss-local-keyframes` entry that exists in `dependencies` group
- Run `npm install`

## Testing Instructions

All CIs should pass.


jsnajdr (approved):
Nice fix :shipit:

PR #63226, opened 2024-07-08
Only hide drop indicator when grid has `isManualPlacement` set.

tellthemachines:
## What?

Follow-up to #62777 and #63160.

Fixes the logic in innerBlocks that hides the drop indicator when the grid is in Manual mode to use `isManualPlacement` instead of `columnCount`.

## Testing Instructions

3. In Auto mode, set a non-zero number of "Columns" in the sidebar;
4. Drag one of the grid children around and verify drop indicators show within the grid.


noisysocks (approved):
There are so many of these 🫥

tellthemachines:
> There are so many of these 🫥

I think this is the last one 😅 but yeah there might be a better way to do this.

Issue #63217, opened 2024-07-07
Issue with exporting pattern in the sidebar


t-hamano:
### Description

![image](https://github.com/WordPress/gutenberg/assets/54422211/0a91f449-ef39-4707-87db-d302529fef11)

In WordPress 6.6 RC2 and the latest Gutenberg, I encountered different issues when exporting patterns from the ellipsis menu in the sidebar.

_Note_: In WP6.5, there is no ellipsis menu in the sidebar, so this issue does not occur in the first place.

![image](https://github.com/WordPress/gutenberg/assets/54422211/4c952d75-3619-4724-88e8-e099a032b6b1)

## WP6.6 RC2

The following error is logged in the browser console and nothing happens:

<details><summary>Log</summary>

```
editor.js?ver=1b1858d2bbca98769f87:24917 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'content')
    at getJsonFromItem (editor.js?ver=1b1858d2bbca98769f87:24917:31)
    at callback (editor.js?ver=1b1858d2bbca98769f87:24933:118)
    at Object.callback (editor.js?ver=1b1858d2bbca98769f87:25836:15)
    at onClick (editor.js?ver=1b1858d2bbca98769f87:26032:31)
    at Object.current (components.js?ver=54eab95952fe254d64e9:72265:38)
    at components.js?ver=54eab95952fe254d64e9:2651:53
    at HTMLUnknownElement.callCallback (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
    at invokeGuardedCallback (react-dom.development.js:4277:31)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4291:25)
getJsonFromItem @ editor.js?ver=1b1858d2bbca98769f87:24917
callback @ editor.js?ver=1b1858d2bbca98769f87:24933
callback @ editor.js?ver=1b1858d2bbca98769f87:25836
onClick @ editor.js?ver=1b1858d2bbca98769f87:26032
(anonymous) @ components.js?ver=54eab95952fe254d64e9:72265
(anonymous) @ components.js?ver=54eab95952fe254d64e9:2651
callCallback @ react-dom.development.js:4164
invokeGuardedCallbackDev @ react-dom.development.js:4213
invokeGuardedCallback @ react-dom.development.js:4277
invokeGuardedCallbackAndCatchFirstError @ react-dom.development.js:4291
[...]

PR #63228, opened 2024-07-08
Fix: Restrict export pattern action to user patterns

ntsekouras:
## What?
According to this issue: https://github.com/WordPress/gutenberg/issues/61787 the `export pattern` action is not meant to be for theme patterns. With a recent refactoring of that action this regressed and we ended up with showing this action in theme patterns, which also caused other issues of not exporting properly(#63217). 

I think the solution is to add back the check about the `wp_block` (theme patterns have `type===pattern`). This makes it a bit confusing though with our API for registering actions, which right now is `registerEntityAction( 'postType', 'wp_block', exportPattern );`.

We should look into the `type===pattern` and where/why is needed.. 

## Testing Instructions
1. Exporting/importing patterns should work properly in patterns list (there is [an open issue](https://github.com/WordPress/gutenberg/issues/63217) for the post summary actions)
2. Export action should not be available for theme patterns

### DIFF
--- a/packages/editor/src/dataviews/actions/export-pattern.tsx
+++ b/packages/editor/src/dataviews/actions/export-pattern.tsx
@@ -34,9 +34,7 @@ const exportPattern: Action< Pattern > = {
 	id: 'export-pattern',
 	label: __( 'Export as JSON' ),
 	supportsBulk: true,
-	isEligible: ( item ) => {
-		return item.type === 'wp_block';
-	},
+	isEligible: ( item ) => item.type === 'wp_block',
 	callback: async ( items ) => {
 		if ( items.length === 1 ) {
 			return downloadBlob(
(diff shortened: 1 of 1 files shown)


packages/editor/src/dataviews/actions/export-pattern.tsx
```diff
@@ -34,6 +34,9 @@ const exportPattern: Action< Pattern > = {
 	id: 'export-pattern',
 	label: __( 'Export as JSON' ),
 	supportsBulk: true,
+	isEligible: ( item ) => {
+		return item.type === 'wp_block';
+	},
```
ramonjd: Optional

```suggestion
	isEligible: ( item ) => item.type === 'wp_block',
```

ramonjd (approved):
LGTM

I can export custom patterns but not theme patterns 👍🏻 

Screenshot 2024-07-08 at 5 55 47 PM



Screenshot 2024-07-08 at 5 55 52 PM

t-hamano (approved):
LGTM!

Not a blocker, but I think it might be better to use `PATTERN_TYPES.user` instead of a hardcoded string.

However, it seems necessary to write `@ts-ignore` to avoid type errors in typescript file, like here:

https://github.com/WordPress/gutenberg/blob/879cdc541c3509ca32f1b7819dde221cb480fdc6/packages/editor/src/dataviews/actions/delete-post.tsx#L14-L15

PR #63229, opened 2024-07-08
CustomSelectControlV2: collapse checkmark space when unchecked

ciampo:
## What?

Part of #55023

Extracted from #63179

Allow checkmark wrapper to collapse when not shown

## Why?

So that the item's content can go full width in most scenarios, like it does already in the legacy V1 implementation

## How?

By setting the `font-size` to `0` when necessary.

## Testing Instructions

- In Storybook, visit the `CustomSelectControlV2` legacy "with long labels" example
- Open the select popover, and resize the page until the text needs to wrap on two or more lines.
- Notice how the text goes all the way to the end of the popover (since the checkmark is collapsed)
- Select that item, and notice how the checkmark appears, causing the text to take less space
- Make sure this is the same behavior of V1
- Visit the "With Hints" example, and make sure that the checkmark wrapper never collapses, even for items that are not checked

## Screenshots or screencast 

| Legacy V1 | V2 adapter - before (trunk) | V2 adapted - after (this PR) |
|---|---|---|
| Screenshot 2024-07-08 at 10 16 03 | Screenshot 2024-07-08 at 10 15 39 | Screenshot 2024-07-08 at 10 19 59 |
| Screenshot 2024-07-08 at 10 16 13 | Screenshot 2024-07-08 at 10 15 28 | Screenshot 2024-07-08 at 10 20 15 |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -36,6 +36,7 @@
 -   `CustomSelectControlV2`: keep legacy arrow down behavior only for legacy wrapper. ([#62919](https://github.com/WordPress/gutenberg/pull/62919))
 -   `CustomSelectControlV2`: fix trigger button font size. ([#63131](https://github.com/WordPress/gutenberg/pull/63131))
 -   `CustomSelectControlV2`: fix labelling with a visually hidden label. ([#63137](https://github.com/WordPress/gutenberg/pull/63137))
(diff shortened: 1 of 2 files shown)


tyxla (approved):
LGTM and tests well:

Screenshot 2024-07-08 at 12 42 09
Screenshot 2024-07-08 at 12 42 29

PR #63231, opened 2024-07-08
Fix: block selection mode in block lists with disabled blocks

### LINKED ISSUE #62935: Synced Patterns: Non-overridable content is erroneously editable in Navigation Mode when using keyboard
### Description

Actual: When inserting a synced pattern into a post, if one navigates using a keyboard in Navigation Mode, one is able to press Enter on non-overridable blocks and change their content.

Expected: Non-overridable blocks in should not be editable.

Note that changes to these non-overridable blocks do not persist when a post is saved; even so, it's a confusing experience for anyone navigating with a keyboard.

1. Create a synced pattern and add some content (headings, paragraphs, etc.)
2. Enable overrides on some elements, but not others
3. Save the pattern and insert it into a post
4. Tab through pattern using a keyboard
5. On a non-overridable block, press `Enter` and see that the content can be edited

https://github.com/WordPress/gutenberg/assets/5360536/2b7602d4-e41a-4d9a-b5ec-274b093b9651

### Environment info

WP 6.5, Gutenberg 18.7.0-rc.1

Yes

Yes

talldan:
## What?
Fixes #62935

For a few releases, the block editor has had the concept of blocks that cannot be edited or even selected via the block editing mode feature (`blockEditingMode === 'disabled').

Generally these blocks can only be edited by editing a different document (e.g. for example when editing a page, the blocks that are part of the template need to be edited by opening the template, they can't be edited from the page). The idea is that this simplifies editing, as users are only interacting with one document at a time.

There's a bug in trunk that the editor's select mode still allows selecting and editing blocks from other documents. They should be bypassed via select mode. This is confusing because the edits made are not something that can be saved.

## How?
Refactors select mode navigation to navigate the block hierarchy returned by `getEnabledClientIdsTree`. This existing memoized selector returns the same block hierarchy as shown by list view, and ignores disabled blocks.

This completely changes the way the navigation worked, so has required some refactoring. There were also some code quality issues in the file, so I've attempted to solve them, but it might be worth moving them into a separate PR to make the diff easier to understand.

TBC - I'm thinking of a good way to explain how this can be tested 🤔

### DIFF
--- a/packages/block-editor/src/components/block-tools/block-selection-button.js
+++ b/packages/block-editor/src/components/block-tools/block-selection-button.js
@@ -138,12 +138,6 @@ function BlockSelectionButton( { clientId, rootClientId }, ref ) {
 
 		const selectedBlockClientId = getSelectedBlockClientId();
 		const selectionEndClientId = getMultiSelectedBlocksEndClientId();
-		const selectionBeforeEndClientId = getPreviousBlockClientId(
-			selectionEndClientId || selectedBlockClientId
-		);
-		const selectionAfterEndClientId = getNextBlockClientId(
-			selectionEndClientId || selectedBlockClientId
-		);
 
 		const navigateUp = ( isTab && isShift ) || isUp;
 		const navigateDown = ( isTab && ! isShift ) || isDown;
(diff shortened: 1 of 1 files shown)


talldan:
Navigation mode doesn't exist any more so I'll close this.

PR #63235, opened 2024-07-08
Core Commands: fix Pages command link

t-hamano:
Related to #61741

## What?

I noticed that the core command "Pages" wasn't linking to the correct pages, so I fixed that.

## Why?

The parameter was changed [here](https://github.com/WordPress/gutenberg/pull/61741/files#diff-ebd9886af023e11457166051cbc482c19a82e428a3db1797da2e3ba8eb846afeR317), but it looks like a typo to me.

## Testing Instructions

- Launch the command palette in the site editor.
- Select "Pages"
- You should see the Page screen.
- It should work the same way when you run the command palette in the post editor.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/4118ebfa-382e-4362-be93-36d242b90b47

### DIFF
--- a/packages/core-commands/src/site-editor-navigation-commands.js
+++ b/packages/core-commands/src/site-editor-navigation-commands.js
@@ -315,7 +315,7 @@ function useSiteEditorBasicNavigationCommands() {
 				icon: page,
 				callback: ( { close } ) => {
 					const args = {
-						post_type: 'page',
+						postType: 'page',
 					};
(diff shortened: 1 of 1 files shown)


t-hamano:
#61741 first shipped in WP6.6, which means this issue first appeared in WP6.6, so I'm adding the backport label to this PR.

youknowriad (approved):
Good catch, Thanks for the fix.

Issue #57669, opened 2024-01-09
DataViews option persistence


jameskoster:
In DataViews there are various configurable options that you'd expect to persist across sessions. For example changing the number of items per page, or the fields that are displayed. Currently this is not the case.

There is also a question to answer about whether view options should be tied to layout, or the overall view. For instance if I'm viewing All Pages and change the 'items per page' option in Table layout, should that also apply in Grid layout?

bgardner:
Huge fan of this, particularly for Layouts (Pages and Templates) which seem to have inconsistent defaults anyway (Grid and List, respectively). Full disclosure: I prefer Table view on both, so having persistence once I set to these views would be amazing.

bacoords:
+1 and this is making me wonder if we need a full "Visual State Persistence" tracking issue as [this keeps coming up](https://github.com/WordPress/gutenberg/issues/41544#issuecomment-2289485629) across the user interface (CC @annezazu)? 

The site editor should consider including at least some ability to keep some of these visual views in some sort of state or local storage for each individual screen so that everything doesn't always get reset every time you click into a single post object to edit it.

kubiqsk:
I believe that everyone will be happy with a table layout as a default option everywhere,
because this is crazy resources expensive to render all the block in all the templates once you click on Templates in sidepanel in Site Editor.
I'm trying to find any solution for this, because people are getting really angry about freezing browsers...

oandregal:
Added three more features (user choices) that would benefit from being persisted:

- #64678
- #64679
- #65600

jameskoster:
I updated the OP slightly. I think there's a bigger question to answer about what the persistence is attached to; the different layouts within the view, or the view itself.

I lean slightly towards the layout, as they're optimised for different workflows. For example in Table layout it probably makes sense for many fields to be visible. But Grid layout is more visual; in most cases it's desirable to show only one or two fields. As a practical example in the shipping software, template descriptions are useful in Table layout, but [in List layout they're kind of overwhelming](https://github.com/WordPress/gutenberg/issues/57596):

![Image](https://github.com/user-attachments/assets/b8a11848-e1f5-4dbe-90df-14aa583b7b50)


@oandregal perhaps you could clarify how things work currently? @youknowriad I suppose this could be related to https://github.com/WordPress/gutenberg/issues/67391.

youknowriad:
[...]

PR #63236, opened 2024-07-08
DataViews: Support combined fields

youknowriad:
Related to #55083 
Follow-up to #63127 

## What?

This PR implements combining fields in table views. It allows combining one or more fields horizontally or vertically in table views. To do so, you have to define a "combinedField" in `view.layout.combinedFields` For instance:

```
{
	id: 'template',
	header: __( 'Template' ),
	children: [ 'title', 'description' ],
	direction: 'vertical',
}
```

and then you can update the `fields` property to include the `template` field. 

This doesn't work entirely well in the fields visibility menu.

I was considering an alternative approach that was way bigger (See #63148) but now I'm thinking the refactoring is not worth doing yet as there's still a lot of unknowns. So this PR implements the combination with the minimal changes.

**Testing instructions**

 - You can check the "table" layout in the templates dataviews.

### DIFF
--- a/packages/dataviews/src/layouts.ts
+++ b/packages/dataviews/src/layouts.ts
@@ -16,6 +16,7 @@ import ViewTable from './view-table';
 import ViewGrid from './view-grid';
 import ViewList from './view-list';
 import { LAYOUT_GRID, LAYOUT_LIST, LAYOUT_TABLE } from './constants';
+import type { View } from './types';
 
 export const VIEW_LAYOUTS = [
 	{
@@ -37,3 +38,29 @@ export const VIEW_LAYOUTS = [
 		icon: isRTL() ? formatListBulletsRTL : formatListBullets,
 	},
 ];
+
+export function getMandatoryFields( view: View ): string[] {
+	if ( view.type === 'table' ) {
+		return [ view.layout?.primaryField ]
+			.concat(
+				view.layout?.combinedFields?.flatMap(
+					( field ) => field.children
+				) ?? []
+			)
+			.filter( ( item ): item is string => !! item );
+	}
+
+	if ( view.type === 'grid' ) {
+		return [ view.layout?.primaryField, view.layout?.mediaField ].filter(
+			( item ): item is string => !! item
+		);
+	}
+
+	if ( view.type === 'list' ) {
+		return [ view.layout?.primaryField, view.layout?.mediaField ].filter(
+			( item ): item is string => !! item
+		);
+	}
+
+	return [];
+}
--- a/packages/dataviews/src/stories/fixtures.js
+++ b/packages/dataviews/src/stories/fixtures.js
@@ -167,20 +167,17 @@ export const fields = [
 				<img src={ item.image } alt="" style={ { width: '100%' } } />
 			);
 		},
-		width: 50,
 		enableSorting: false,
 	},
 	{
 		header: 'Title',
 		id: 'title',
-		maxWidth: 400,
 		enableHiding: false,
 		enableGlobalSearch: true,
 	},
 	{
 		header: 'Type',
 		id: 'type',
-		maxWidth: 400,
 		enableHiding: false,
 		elements: [
 			{ value: 'Not a planet', label: 'Not a planet' },
@@ -197,7 +194,6 @@ export const fields = [
 	{
 		header: 'Description',
 		id: 'description',
-		maxWidth: 200,
 		enableSorting: false,
 		enableGlobalSearch: true,
 	},
--- a/packages/dataviews/src/stories/index.story.js
+++ b/packages/dataviews/src/stories/index.story.js
@@ -39,6 +39,20 @@ Default.args = {
 		[ LAYOUT_TABLE ]: {
 			layout: {
 				primaryField: 'title',
+				styles: {
+					image: {
+						width: 50,
+					},
+					title: {
+						maxWidth: 400,
+					},
+					type: {
+						maxWidth: 400,
+					},
+					description: {
+						maxWidth: 200,
+					},
+				},
 			},
 		},
 		[ LAYOUT_GRID ]: {
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -211,6 +211,10 @@
 				}
 			}
 		}
+
+		.components-v-stack > .dataviews-view-table__cell-content-wrapper:not(:first-child) {
+			min-height: 0;
+		}
 	}
 	.dataviews-view-table-header-button {
 		padding: $grid-unit-05 $grid-unit-10;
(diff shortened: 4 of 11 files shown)


jameskoster:
If we're going to force the description to be visible in table layout – as this PR does – then the 'Description' toggle in view options feels superfluous. I think the toggle should either be removed (only for table layout), or connected with the field inside the "Template" column. The latter would obviously be more flexible, but probably trickier to implement?

One minor detail; if the description doesn't live in it's own column, then we no longer need to show the `—` when no description is present.

youknowriad:
@jameskoster That's a good point indeed, and I think it might take a few PRs to get everything right here. For start, I'm now hiding all fields that are part of the "combined fields".

jameskoster:
Functionality-wise this works for me, for now :)

In terms of styling I have a small request; 

* For VStack arrangements, the gap can be 0.
* For HStack; 12px.

I wonder if we should restrict the width of the description somehow, legibility on widescreens is a bit awkward:

Screenshot 2024-07-09 at 11 47 40

Do you think it would be best to do that on the whole column, or just the description field?

youknowriad:
On "widths" configurations. We do have "widths" defined on fields but for now I'm ignoring them in the PR because I think this was not the right implementation. These widths don't make sense in other layouts, so they should mostly be layout specific configurations. 

The work I'm doing here #63287 will allow us to define these "layout + field" specific rendering configs more easily.

jameskoster (approved):

youknowriad:
@jameskoster now that the other PR landed, I rebased this one and I restored the ability to set "widths" for columns. Let me know if it works as you expect. In my testing, setting min/max widths didn't really any meaningful impact but I'm not sure if that's very different than trunk.

jameskoster:
@youknowriad The preview field seems to have moved to the end of the row:

Screenshot 2024-07-10 at 11 17 14

---

The column width doesn't seem to have any affect, I suppose that's just the nature of tables. Maybe the description field should include a max-width. Probably fine to look into that separately.

---

One other observation is that the space between title + description is different based on the number of lines:

Screenshot 2024-07-10 at 11 21 04

It's caused by the `.dataviews-view-table__cell-content-wrapper` class. The styles associated with that class need only apply to the first text-based field in the VStack (it's to ensure alignment across the row). 

Something like this would fix:

```
.dataviews-view-table tbody .components-v-stack .dataviews-view-table__cell-content-wrapper + .dataviews-view-table__cell-content-wrapper {
min-height: 0;
}
```

Though we probably need something more specific than `.components-v-stack`.

youknowriad:
> The preview field seems to have moved to the end of the row

I was wondering when you'll discover that :P. It's not just the end, depending on the order of when you trigger the visibility of the fields, they may appear in different orders. The reality here is that there's no way around that aside providing a way to reorder fields in the UI. We can either decide to ship this PR as is or pause and implement field reordering first in trunk and come back to this PR afterwards.

> The column width doesn't seem to have any affect, I suppose that's just the nature of tables. Maybe the description field should include a max-width. Probably fine to look into that separately.

I found that setting "width" actually is better than max/min widths but yeah it's the same as trunk, so we should explore it separately.

> It's caused by the .dataviews-view-table__cell-content-wrapper class. The styles associated with that class need only apply to the first text-based field in the VStack (it's to ensure alignment across the row).

Can you explain the purpose of these styles so I can try to think about the right fix.

jameskoster:
> I was wondering when you'll discover that :P

Sneaky 🥷 So there's no way to define a default field order without also building re-ordering? If that's the case then we may need to prioritise re-ordering. It seems a bit unexpected for the Preview to appear at the end of the row. Interested to hear @jasmussen's thoughts on that though.

> Can you explain the purpose of these styles so I can try to think about the right fix.

`.dataviews-view-table__cell-content-wrapper` and the applied `min-height` ensures alignment along the x axis:

Screenshot 2024-07-10 at 15 58 10

Additionally, when the row contains a field that is taller than the min-height, everything keeps alignment at the top of the row:

Screenshot 2024-07-10 at 15 58 22

youknowriad:
> Sneaky 🥷 So there's no way to define a default field order without also building re-ordering? If that's the case then we may need to prioritise re-ordering. It seems a bit unexpected for the Preview to appear at the end of the row. Interested to hear @jasmussen's thoughts on that though.

I can try some temporary solutions but nothing that I like. I do think we need to prioritize reordering.

youknowriad:
> I can try some temporary solutions but nothing that I like. I do think we need to prioritize reordering.

The main issue is that we can use the "fields" definition as the default order, but it doesn't work when you inject "virtual fields" like "template" in this PR. So we need to either consider the "template" as last or first item when toggling visibility.

youknowriad:
> It's caused by the .dataviews-view-table__cell-content-wrapper class. The styles associated with that class need only apply to the first text-based field in the VStack (it's to ensure alignment across the row).

This is not something that can be codified and I'm not entirely sure how to solve this issue to be honest.

youknowriad:
And for the reordering issue, I'd rather solve it first as trying to fit something here would be too hacky. So what do you think we should do?

youknowriad:
I pushed a small fix for the line height, I think it's not a great fix as these rules are very fragile and the don't scale/compose well when the tables get complex but I can live with it for now.

jameskoster:
If re-ordering isn't a huge lift then I suppose it makes sense to prioritise. The design should be relatively simple:

Screenshot 2024-07-10 at 16 36 47

> these rules are very fragile and the don't scale/compose well when the tables get complex but I can live with it for now.

I agree, but I'm hesitant to admit defeat and resort to `vertical-align: middle` because that presents other issues:

Screenshot 2024-07-10 at 16 45 47

Tables... 😓

youknowriad:
Ok, so what do you think about shipping this PR as is and I can work on reordering as a follow-up. The designs look easy enough to implement.

jameskoster:
Sounds like a plan 👍 

I suspect we may also need to prioritise https://github.com/WordPress/gutenberg/issues/57669. It would be annoying to have to toggle visibility _and_ order frequently. Or will the re-ordering function make it possible to set a default order too?

youknowriad:
> Or will the re-ordering function make it possible to set a default order too? 

 Default order when you first load the layout but not when you enable fields. I think that would be too much.

packages/dataviews/src/layouts.ts
```diff
@@ -37,3 +38,29 @@ export const VIEW_LAYOUTS = [
 		icon: isRTL() ? formatListBulletsRTL : formatListBullets,
 	},
 ];
+
+export function getMandatoryFields( view: View ): string[] {
+	if ( view.type === 'table' ) {
+		return [ view.layout?.primaryField ]
+			.concat(
+				view.layout?.combinedFields?.flatMap(
```
ntsekouras, line 46: Why should we consider combined fields as mandatory? 🤔

youknowriad: Because we don't have the right UI and components to actually disable them. If you manage to make them work without issues, you can remove this condition :)

PR #63238, opened 2024-07-08
Video Block: Do not persist blob urls and fix undo

youknowriad:
Related #39223 
Similar to #63076

## What?

In #63076, we introduce the concept of local attributes to allow setting attributes that won't be persisted. It's useful during file uploads for instance where we want immediate user feedback but without impacting the saved post.

In the previous PR, we implemented it for image blocks and this PR does the same for the video block. This PR also fixes an undo bug in the video block.

## Testing Instructions

1- Throttle your network speed (that way uploading images will take a long time)
2- Drag and drop a video file into the post editor.
3- Save the post immediately.
4- Open the post in a new tab and notice that the inserted image is actually empty.
6- Notice that the video block has been updated properly with the right URL.
You can also try undoing uploads and it should results in empty video blocks.


ellatrix (approved):
Tests well

Issue #61761, opened 2024-05-17
Global styles panel header buttons text overlaps when 'Show button text labels' is enabled


afercia:
### Description

When the 'Show button text labels' preference is enabled, the text of the buttons in the Styles panel header overlaps, making thte text barely readable.

Screenshot from trunk:

![trunk](https://github.com/WordPress/gutenberg/assets/1682452/24bfe907-2b17-4ccc-9c34-71caaddcb28b)

This appears to be a regression from WordPress 6.5, where the text is readable. Screenshot:

![wp 6 5](https://github.com/WordPress/gutenberg/assets/1682452/29d92cef-1720-474f-8123-44e0842c7a55)

I _think_ the recent addition of the `is-compact` CSS class and styling to the buttons hasn't been tested for 'Show button text labels'. Regardless, the buttons text should never overlap.

Worth noting the horizontal space in the panel header is very limited. Placing there a title, 3 action buttons and a close button doesn't seem ideal in the first place.

On the other hand, the 'Show button text labels' preference seems to be often neglected and missed during the design and testing phase. I'd love to see this feature being treated as a first-class citizen in this project at the point that I'd think anydesign for new feature should provide design and mockups for 'Show button text labels' as well.

Note: I created a new GitHub label `[Feature] Show button text labels`, similarly to other used preferences like `[Feature] Distraction Free` to better track issues related to it. Thre have been several issues and PRs to remediate to breakages of buttons with this prefrence enabled that are all scattered around and difficult to search for.

### Step-by-step reproduction instructions

- Go to the Site editor > Options > Preferences > Accessibility, and enable the 'Show button text labels' preference.
- Open the Styles panel.
- Observe the buttons text overlaps.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

jasmussen:
What design feedback are you looking for?

afercia:
@jasmussen as I see it:
- The root problem is that all those button shouldn't be placed there in the first place.
- I'd appreciate the design team to provide design and mockups also for the 'Show button text labels' preference, when relevant, as this is an integral part of the interface.

Too many buttons in a panel header that shoud likely only contain the title and the X close button:

![Screenshot 2024-05-21 at 10 04 39](https://github.com/WordPress/gutenberg/assets/1682452/95d8a169-ea17-465c-8b0e-e2e2fa5358fc)

[...]

PR #63243, opened 2024-07-08
Fix Global styles panel header buttons text overlap for 'Show button text labels'

afercia:
Temporary fix for https://github.com/WordPress/gutenberg/issues/61761

A more solid solution should be implemented by changing the design and not placing buttons in the panels header in the first place. This PR overrides the Button component CSS, which isn't ideal but it is worth noting the 'Show button text labels' feature itself overrides the base component CSS. An alternative solution is being explored in https://github.com/WordPress/gutenberg/issues/61763

Additionally, makes the size of the Close button consistent with the size of the other buttons.

## What?

The buttons text overlaps.

## Why?

Overlapping text isn't readable.

## How?

Makes the button width be `auto` when 'Show button text labels' is enabled.
Makes the Close button use the `compact` size so that it is consistent with the other buttons.

## Testing Instructions

- Go to the Site Editor > Styles
- In Options > Preferences > Accessibility, enable the 'Show button text labels' preference.
- Alternatively, issue this command in your browser dev tools console: `window.wp.data.dispatch( 'core/preferences' ).set( 'core', 'showIconLabels', true )`
- Observe the text of the buttons in the Styles panel header don't overlap any longer.
- Test with the admin language set to languages that may provide longer strings e.g. German, Spanish, Italian.
- Disable 'Show button text labels' from the preferences modal or alternatively issue this command in the console: `window.wp.data.dispatch( 'core/preferences' ).set( 'core', 'showIconLabels', false )`
- Observe the size of the X close button is 32 by 32 pixels.
- Install a plugin that adds its own settings panel e.g. Yoast SEO.
- Go to the Post Editor.
- Open the Plugin sidbar from the Options dropdown menu.
- Observe the Plugin panel Close button size is 32 by 32 pixels.

## Screenshots or screencast 

Overlapping text - Before: English and italian

Screenshot 2024-07-08 at 15 06 08

After:

english

Close button size before (all buttons focused to better illustrate):

inconsistent size

After:

Screenshot 2024-07-08 at 14 46 46

### DIFF
--- a/packages/edit-site/src/components/global-styles-sidebar/index.js
+++ b/packages/edit-site/src/components/global-styles-sidebar/index.js
@@ -137,6 +137,7 @@ export default function GlobalStylesSidebar() {
 					</FlexItem>
 					<Flex
 						justify="flex-end"
+						gap={ 1 }
 						className="edit-site-global-styles-sidebar__header-actions"
 					>
 						<FlexItem>
--- a/packages/edit-site/src/components/global-styles-sidebar/index.js
+++ b/packages/edit-site/src/components/global-styles-sidebar/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { FlexItem, FlexBlock, Flex, Button } from '@wordpress/components';
+import { FlexItem, Flex, Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { styles, seen, backup } from '@wordpress/icons';
 import { useSelect, useDispatch } from '@wordpress/data';
@@ -130,38 +130,45 @@ export default function GlobalStylesSidebar() {
 					className="edit-site-global-styles-sidebar__header"
 					gap={ 1 }
 				>
-					<FlexBlock style={ { minWidth: 'min-content' } }>
+					<FlexItem>
 						<h2 className="edit-site-global-styles-sidebar__header-title">
 							{ __( 'Styles' ) }
 						</h2>
-					</FlexBlock>
-					<FlexItem>
-						<Button
-							icon={ seen }
-							label={ __( 'Style Book' ) }
-							isPressed={
-								isStyleBookOpened || isRevisionsStyleBookOpened
-							}
-							accessibleWhenDisabled
-							disabled={ shouldClearCanvasContainerView }
-							onClick={ toggleStyleBook }
-							size="compact"
-						/>
 					</FlexItem>
-					<FlexItem>
-						<Button
-							label={ __( 'Revisions' ) }
-							icon={ backup }
-							onClick={ toggleRevisions }
-							accessibleWhenDisabled
-							disabled={ ! hasRevisions }
-							isPressed={
-								isRevisionsOpened || isRevisionsStyleBookOpened
-							}
-							size="compact"
-						/>
-					</FlexItem>
-					<GlobalStylesMenuSlot />
+					<Flex
+						justify="flex-end"
+						className="edit-site-global-styles-sidebar__header-actions"
+					>
+						<FlexItem>
+							<Button
+								icon={ seen }
+								label={ __( 'Style Book' ) }
+								isPressed={
+									isStyleBookOpened ||
+									isRevisionsStyleBookOpened
+								}
+								accessibleWhenDisabled
+								disabled={ shouldClearCanvasContainerView }
+								onClick={ toggleStyleBook }
+								size="compact"
+							/>
+						</FlexItem>
+						<FlexItem>
+							<Button
+								label={ __( 'Revisions' ) }
+								icon={ backup }
+								onClick={ toggleRevisions }
+								accessibleWhenDisabled
+								disabled={ ! hasRevisions }
+								isPressed={
+									isRevisionsOpened ||
+									isRevisionsStyleBookOpened
+								}
+								size="compact"
+							/>
+						</FlexItem>
+						<GlobalStylesMenuSlot />
+					</Flex>
 				</Flex>
 			}
 		>
--- a/packages/edit-site/src/components/global-styles-sidebar/style.scss
+++ b/packages/edit-site/src/components/global-styles-sidebar/style.scss
@@ -31,6 +31,10 @@
 	margin: 0;
 }
 
+.edit-site-global-styles-sidebar .edit-site-global-styles-sidebar__header-actions {
+	flex: 1;
+}
+
 .edit-site-global-styles-sidebar .components-navigation__menu-title-heading {
 	font-size: $default-font-size * 1.2;
 	font-weight: 500;
--- a/packages/edit-site/src/components/global-styles-sidebar/style.scss
+++ b/packages/edit-site/src/components/global-styles-sidebar/style.scss
@@ -75,6 +75,8 @@
 .show-icon-labels {
 	.edit-site-global-styles-sidebar__header {
 		.components-button.has-icon {
+			width: auto;
+
 			// Hide the button icons when labels are set to display...
 			svg {
 				display: none;
--- a/packages/interface/src/components/complementary-area-header/style.scss
+++ b/packages/interface/src/components/complementary-area-header/style.scss
@@ -26,11 +26,6 @@
 
 	.components-button.has-icon {
 		display: none;
-		margin-left: auto;
-
-		~ .components-button {
-			margin-left: 0;
-		}
 
 		@include break-medium() {
 			display: flex;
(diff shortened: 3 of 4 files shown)


afercia:
@WordPress/gutenberg-core I'd appreciate a review, when you have a chance.

t-hamano (commented):
It probably works fine for Latin languages, but at least for Japanese it seems to cause overflow.

In both trunk and this PR I can see extreme text wrapping.

trunk:

image

This PR:

image

One approach might be to apply `white-space: nowrap` to the header title (`.edit-site-global-styles-sidebar__header-title`):

image

Regarding the size of the close button, I personally lean towards keeping it at the current size (24px), since the size of the close button on the main inserter button is 24px:

image

However, I would be grateful for any design feedback on this point.

afercia:
@t-hamano thanks for testing in Japanese. I'd love to see all designers and developers test often with other languages in their daily workflow.

There's really no ideal solution to this layout issue other than changing a design that is problematic in the first place. The panel header can't contain all that content and the length of the content is unpredictable. The original design must be reconsidered. See https://github.com/WordPress/gutenberg/issues/63251

What we can do now is to try the best compromise for a temporary improvement. In the last commit I've refactored a bit the flexbox implementation of this specific panel header.

Is this wrapping acceptable in Japanese @t-hamano?
Screenshots with various languages:

Screenshot 2024-10-18 at 15 16 34

As I mentioned earlier, this can't be perfect for all cases but I think it's an improvement compared to trunk.

t-hamano:
> Is this wrapping acceptable in Japanese @t-hamano?

Yes. In Japanese, there are no spaces between words, and it is quite common for words to wrap mid-word depending on the width of the container.

So I think this is not unnatural and is acceptable (and probably Chinese as well).

afercia:
Thanks @t-hamano.
As mentioned, this is only a temporary fix. The design needs to be changed anyways, including the styling of the buttons. From an accessibility perspective, they are not distinguishable from normal text. While this can be acceptable in the top bar or the block toolbar, where the context and additional shapes clarofy those are buttons, in this case these buttons look like normal text:

a11y

A final review and approval would be appreciated. Thanks.

jameskoster:
The revisions button invokes a drilldown, which feels strange given the access point (panel header toggle button). I wouldn't object to moving this to the panel footer, similar to the Styles sidebar:

Screenshot 2024-10-18 at 12 49 31

This would create some extra space for the other buttons which might help?

t-hamano:
Those two margin-lefts should no longer be necessary:

https://github.com/WordPress/gutenberg/blob/245176e0afa5b0f235aa1cad55f7db376f598bc4/packages/interface/src/components/complementary-area-header/style.scss#L29-L33

Also, I noticed that the button is not sized correctly in the block sidebar:

image

Additionally, I remember that the center of this close button might be expected to be aligned with the center of the Options button:

| trunk | This PR |
|--------|--------|
| trunk |  pr | 

Therefore, I lean toward leaving the button size at 24px.

afercia:
> The revisions button invokes a drilldown, which feels strange given the access point (panel header toggle button). I wouldn't object to moving this to the panel footer,

@jameskoster interesting point. To me all those buttons shouldn't be in the Styles panel header in the first place. For example, also the ellipsis button is arguably consistent. I'd expect an ellipsis button to always be _the last button_ in a set of controls. Rearranging these buttons is out of the scope of this PR though. We agreed in past conversations to avoid scope creep so I'd encourage you to create a new issue.

I adjusted the Close button size to 'small', as requested. However, I disagree with the recent change that made these buttons smaller. The target size of any control should be as large as possible, I'm not sure that intentionally (and unnecessarily) reducing the target size is a good strategy to provide a bettter user experience. I will create a new issue.

aaronrobertshaw:
> I'd expect an ellipsis button to always be the last button in a set of controls.

There's also a separate expectation around where a close button is positioned. I definitely agree these are in conflict with each other and a factor in why the UI here feels a bit off.

If I had to pick one to take precedence, I'd think the expectation around the close button would be strongest. So I can understand the past reasoning that lead us here.

> Rearranging these buttons is out of the scope of this PR though. We agreed in past conversations to avoid scope creep so I'd encourage you to create a new issue.

A different perspective here (well, mine specifically 🙂) is that the scope of the PR was to fix the overlapping headings. The proposed relocation of the revisions button/options is a potential part of a solution to that. Sometimes, we can do both things i.e. be cautious of true scope creep and remain flexible. 

Correct me if I'm wrong but the overlapping isn't a new bug? So perhaps we can afford to take a moment to hone in on a better solution?

afercia:
> If I had to pick one to take precedence,

Thanks @aaronrobertshaw for bringing in additional considerations. I'm not sure trying to establish which one of these buttons has precedence to be relocated as last would solve the actual issue. There are arguments for both cases. To me, relocating would cure only the symptom. The root cause is that the additional buttons should not be placed in the panel header in the first place.

The panel header should conly contain:
- Title
- X close button

Any other content should be moved to a 'sub-header' as proposed on https://github.com/WordPress/gutenberg/issues/63251

> Correct me if I'm wrong but the overlapping isn't a new bug?

Correct, it's a long standing issue. A previous PR tried to improve things but it's still not ideal. I'd like to unblock this PR to solve an issue that has been reported on May, that's a few months ago already and it's a bit too much time to fix such a small issue. Of course, this is a temporary improvement. I'd encourage further discussion on the overall design that, in my opinion, should be changed to address the root cause.

aaronrobertshaw:
> I'm not sure trying to establish which one of these buttons has precedence to be relocated as last would solve the actual issue

My comment wasn't intended as a solution, just a counter point to the expectation of the ellipsis button being last. It was also aimed at understanding the problem space and the context behind the current situation.

> The root cause is that the additional buttons should not be placed in the panel header in the first place.

True.

We are where we are though so relocating a button to the bottom of the panel is a good first step towards an ideal scenario. Whether that is moving all buttons outside the header or into a sub-header as proposed. 

> it's a bit too much time to fix such a small issue

Understood. Sometimes things need time though.

> I'd encourage further discussion on the overall design that, in my opinion, should be changed to address the root cause.

I agree that a holistic solution is best left for a separate discussion. 

My two cents on the current approach though is that while it is an improvement on the overlapping labels, is it that much extra to move the button to the bottom of the panel for a much bigger payoff?

This branch could still result in overlapping labels, removing a button helps mitigate the chance of that.

Screenshot 2024-10-22 at 11 45 51 am

ntsekouras (approved):
Thanks for the PR! I get @jameskoster and @aaronrobertshaw points about the possibility of moving the revisions button to the bottom of the panel, but I also think this can be handled in a follow up. The main reason is that this PR improves the current situation and the moving of revisions can trigger way more discussions, as it did with post revisions.

People might want more prominence and discuss position etc..

I'll give a tentative approval and suggest to create a new issue about moving the revisions in a follow up PR. What do you say @jameskoster and @aaronrobertshaw?

aaronrobertshaw:
> the moving of revisions can trigger way more discussions, as it did with post revisions.
> ... What do you say

Fair enough, if that was the experience of moving the post revisions 👍

afercia:
Thanks everyone. I'll create the new issue to consider relocating the revisions toggle if it hasn't been created yet.

For the future: the only way to prevent this kind of layout issues is that the design team must provide design also for the 'Show button text labels' preference. New UIs should not be merged without accounting for 'Show button text labels'. 
I'd love to see this point added to the design guidelines and workflow. Thanks.
Cc @WordPress/gutenberg-design

jasmussen:
That's a note for all contributors to the project, it's not useful to single out one group of contributors.

afercia:
> it's not useful to single out one group of contributors.

@jasmussen can you please add more context and clarify? Thanks.

mtias:
It also seems the "show button text labels" is being applied too broadly to areas where it was not intended to be, like utility actions in panel headers and inspector tabs.

afercia:
> It also seems the "show button text labels" is being applied too broadly to areas where it was not intended to be, like utility actions in panel headers and inspector tabs.

I kindly disagre. Rather, the  "show button text labels" preference should be appliet to _all_ icon buttons and any design should take it into account.

mtias:
Yeah, I don't think we'll agree on that :) Expanding the application of this preference beyond toolbars is a stretch and would need to be done thoughtfully for each context. Cases like this one cannot be trivially turned into text actions:

image

afercia:
> Yeah, I don't think we'll agree on that :) 

Indeed, we don't agree. I'm aiming to make the UI as accessible as possible. I can only note that it seems that's not the focus for everyone in this project.

> Cases like this one cannot be trivially turned into text actions:

I would argue that cases like this one (that example is from the data views) are just a less than ideal design that I'd call a 'wrong' design, not to blame anyone. And it's it wrong because it didn't take into consideration the preference since the beginning.

I do realize that organizine the complexity of the editor UI is challenging but hiding content, removing or truncating visible text and the like. isn't a good strategy to make the UI _usable for everyone_.

PR #63245, opened 2024-07-08
Try: Inspector inner shadow border fix.

jasmussen:
## What?

There's a double-border at the footer of the inspector:

double border

Perhaps not double, but stacked, at least. This is related to changing some of the borders to be shadow based. This PR changes the direction of the shadow, so this stacked shadow disappears:

single border

**Note:** From reading the code, the 1px `y` position value seemed like it was intentionally added. So I'd love a good look at what side-effects zeroing that out might have. I couldn't find any, but I could be missing a beat.

## Testing Instructions

Test the site and post editors at various breakpoints, with and without the inspector.

### DIFF
--- a/packages/interface/src/components/interface-skeleton/style.scss
+++ b/packages/interface/src/components/interface-skeleton/style.scss
@@ -132,7 +132,7 @@ html.interface-interface-skeleton__html-container {
 	right: 0;
 
 	@include break-medium() {
-		box-shadow: $border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
+		box-shadow: $border-width 0 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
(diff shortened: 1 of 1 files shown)


t-hamano:
Can we also improve the block inserter sidebar shadow?

### Before

```css
.interface-interface-skeleton__secondary-sidebar {
	@include break-medium() {
		box-shadow: $border-width 0 0 0 rgba($color: #000, $alpha: 0.133);
	}
}
```

image

### After


```css
.interface-interface-skeleton__secondary-sidebar {
	@include break-medium() {
		box-shadow: $border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133);
	}
}
```

image

jasmussen:
Nice one, works well:

Screenshot 2024-07-09 at 12 54 07

t-hamano (approved):
LGTM!

I've re-read the discussion in #61835 and tested various patterns, and I believe the y-offset is unnecessary and should not cause any side effects.

PR #63248, opened 2024-07-08
CustomSelectControl V2 legacy adapter: stabilize experimental props

ciampo:
## What?

Part of #55023

Stabilize the `__experimentalShowSelectedHint`  and `options[]. __experimentalHint` props of `CustomSelectControl` V2 legacy adapter by providing an equivalent, un-prefixed version: `showSelectedHint` and `options[].hint`

## Why?

As we're about to swap the current V1 implementation with a new implementation based on `ariakit` (ie. the V2 legacy adapter), it's also a good time to stabilize those props in preparation for the swap.

Once the swap happens, we will also be able to remove [the private `CustomSelectControl` export](https://github.com/WordPress/gutenberg/blob/938a51387ff6c1b7f8c7a4ad4deb745afeff84d5/packages/components/src/private-apis.ts#L39).

## How?

By soft-deprecating those props. Existing consumers using those props won't experience any breakages or warnings, but the deprecated props will show as such in the docs.

## Testing Instructions

- Code-review the code snippet in charge of the soft deprecation
- Smoke test V1 and V2 legacy adapter Storybook examples — especially the "with hint" one. Make sure that the individual option hints show as expected, and play around with the `showSelectedHint` prop.
- Make sure that, in the V2 legacy adapter Storybook controls, the `__experimentalShowSelectedHint` is not shows anymore
- Smoke test component's usages in the editor, especially in the font picker

## ✍️ Dev note

See dev note in https://github.com/WordPress/gutenberg/pull/63258

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -19,6 +19,7 @@
 -   `ToolbarButton`: Always keep focusable when disabled ([#63102](https://github.com/WordPress/gutenberg/pull/63102)).
 -   `ProgressBar`: Fix indeterminate RTL support. ([#63129](https://github.com/WordPress/gutenberg/pull/63129)).
 -   `RangeControl`: Fix RTL support for custom marks ([#63198](https://github.com/WordPress/gutenberg/pull/63198)).
+-   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
 -   `TimePicker`: Add `dateOrder` prop ([#62481](https://github.com/WordPress/gutenberg/pull/62481)).
 
 ### Bug Fixes
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 -   `CustomSelectControlV2`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
 
+### Enhancements
+
+-   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
+
 ## 28.3.0 (2024-07-10)
 
 ### Enhancements
@@ -19,7 +23,6 @@
 -   `ToolbarButton`: Always keep focusable when disabled ([#63102](https://github.com/WordPress/gutenberg/pull/63102)).
 -   `ProgressBar`: Fix indeterminate RTL support. ([#63129](https://github.com/WordPress/gutenberg/pull/63129)).
 -   `RangeControl`: Fix RTL support for custom marks ([#63198](https://github.com/WordPress/gutenberg/pull/63198)).
--   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
 -   `TimePicker`: Add `dateOrder` prop ([#62481](https://github.com/WordPress/gutenberg/pull/62481)).
 
 ### Bug Fixes
--- a/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
+++ b/packages/components/src/custom-select-control-v2/legacy-component/index.tsx
@@ -9,7 +9,7 @@ import clsx from 'clsx';
  */
 import _CustomSelect from '../custom-select';
 import CustomSelectItem from '../item';
-import type { LegacyCustomSelectProps } from '../types';
+import type { LegacyCustomSelectProps, LegacyOption } from '../types';
 import * as Styled from '../styles';
 
 function useDeprecatedProps( {
@@ -26,6 +26,14 @@ function useDeprecatedProps( {
 				: showSelectedHint,
 	};
 }
+const computeOptionDeprecatedProps = ( {
+	__experimentalHint,
+	hint,
+	...rest
+}: LegacyOption ) => ( {
+	...rest,
+	hint: typeof __experimentalHint !== 'undefined' ? __experimentalHint : hint,
+} );
 
 function CustomSelectControl( props: LegacyCustomSelectProps ) {
 	const {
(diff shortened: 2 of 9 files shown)


packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -60,56 +85,56 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
 		defaultValue: options[ 0 ]?.name,
 	} );
 
-	const children = options.map(
```
ciampo, line 63: The following block of code is just about changing from  mostly `__experimentalHint` to `hint`, the rest is whitespace-related changes (indentation)

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -60,56 +85,56 @@ function CustomSelectControl( props: LegacyCustomSelectProps ) {
...
 		return (
 			<Styled.SelectedExperimentalHintWrapper>
 				{ currentValue }
-				<Styled.SelectedExperimentalHintItem
-				// TODO: Legacy classname. Add V1 styles are removed from the codebase
-				// className="components-custom-select-control__hint"
-				>
-					{ currentHint?.__experimentalHint }
-				</Styled.SelectedExperimentalHintItem>
+				{ selectedOptionHint && (
```
ciampo, line 130: Previously, we were rendering this component even where there wasn't a hint to show

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -12,18 +12,48 @@ import CustomSelectItem from '../item';
...
+			const hint =
+				typeof o.__experimentalHint !== 'undefined'
+					? o.__experimentalHint
+					: o.hint;
+
+			if ( hint ) {
+				toReturn.hint = hint;
+			}
+
+			return toReturn;
```
ciampo: Adding the new `hint` property to each option only if necessary, because otherwise this could break consumers of the component that rely on object identity (eg. [`FontSizePicker`](https://github.com/WordPress/gutenberg/blob/9c110ce524ec56062e536bf5f3eea2a11ffdbfb6/packages/components/src/font-size-picker/font-size-picker-select.tsx#L88))

packages/components/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 -   `ToolbarButton`: Always keep focusable when disabled ([#63102](https://github.com/WordPress/gutenberg/pull/63102)).
 -   `ProgressBar`: Fix indeterminate RTL support. ([#63129](https://github.com/WordPress/gutenberg/pull/63129)).
 -   `RangeControl`: Fix RTL support for custom marks ([#63198](https://github.com/WordPress/gutenberg/pull/63198)).
+-   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
```
mirka: Careful with merging this, we just had a release 👀

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -12,18 +12,48 @@ import CustomSelectItem from '../item';
...
+		} ),
+		showSelectedHint:
+			typeof __experimentalShowSelectedHint !== 'undefined'
+				? __experimentalShowSelectedHint
+				: showSelectedHint,
+	};
```
mirka: This prioritizes the `__experimental` over the stable prop. Would it be better to do it like this, so the stable prop is prioritized?

```js
function useDeprecatedProps( {
  __experimentalShowSelectedHint,
  ...otherProps
} ) {
  return {
    showSelectedHint: __experimentalShowSelectedHint,
    ...otherProps
  }
}
```

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
mirka: Hmm, that's actually a breaking change for third-party consumers that could also be relying on object identity.

I'm thinking, how about we leave the options array alone, and do the `__experimental` adapting at [read time](https://github.com/WordPress/gutenberg/blob/28e9e414e7a4c6e9c383a353cd7b2177fe04135c/packages/components/src/custom-select-control-v2/legacy-component/index.tsx#L63)?

packages/components/CHANGELOG.md
ciampo: Thank you for flagging — moved back to the "Unreleased" section

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: Good point, done!

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -12,18 +12,41 @@ import CustomSelectItem from '../item';
...
+// on object identity (see https://github.com/WordPress/gutenberg/pull/63248/files/175907e16310fcd3ef13fcb2964e138a69134760#r1668842238)
+function applyOptionDeprecations( {
+	__experimentalHint,
+	...rest
+}: LegacyCustomSelectProps[ 'options' ][ number ] ) {
+	return {
+		hint: __experimentalHint,
+		...rest,
+	};
+}
```
ciampo, line 36: @mirka Relatively to https://github.com/WordPress/gutenberg/pull/63248#discussion_r1672213131, I applied your suggestion of adapting the `options` array at runtime where necessary

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
```diff
@@ -12,18 +12,41 @@ import CustomSelectItem from '../item';
...
+}: LegacyCustomSelectProps ) {
+	return {
+		showSelectedHint: __experimentalShowSelectedHint,
+		...otherProps,
+	};
+}
+
+// The removal of `__experimentalHint` in favour of `hint` doesn't happen in
+// the `useDeprecatedProps` hook in order not to break consumers that rely
+// on object identity (see https://github.com/WordPress/gutenberg/pull/63248/files/175907e16310fcd3ef13fcb2964e138a69134760#r1668842238)
```
tyxla: That link won't work, did you mean:

```suggestion
// on object identity (see https://github.com/WordPress/gutenberg/pull/63248#discussion_r1672213131)
```

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
tyxla: Good catch!

tyxla (commented):
The v2 legacy stories still use the v1 ones as a base:

https://github.com/WordPress/gutenberg/blob/48422a4f2180dc5a760ed66db5804691fe762948/packages/components/src/custom-select-control-v2/stories/legacy.story.tsx#L15

So as a consequence, the options still use the `__experimentalHint` prop:

Screenshot 2024-07-10 at 17 43 47

ciampo:
> The v2 legacy stories still use the v1 ones as a base:
> [...]
> So as a consequence, the options still use the `__experimentalHint` prop:

That should be ok to leave as-is for now — the Storybook examples will be updated when we remove the downshift-based implementation (see [proposed change](https://github.com/WordPress/gutenberg/pull/63258/files#diff-3487c6aeed8de6e9aece7d62cf3c6ce4def19dc02dc6304c20e2e4f3c2922726)). It shouldn't otherwise impact the component negatively, since it should be support the older version of the prop for backwards compat.

packages/components/src/custom-select-control-v2/legacy-component/index.tsx
ciampo: Done!

tyxla:
> > The v2 legacy stories still use the v1 ones as a base:
> > [...]
> > So as a consequence, the options still use the `__experimentalHint` prop:
> 
> That should be ok to leave as-is for now — the Storybook examples will be updated when we remove the downshift-based implementation (see [proposed change](https://github.com/WordPress/gutenberg/pull/63258/files#diff-3487c6aeed8de6e9aece7d62cf3c6ce4def19dc02dc6304c20e2e4f3c2922726)). It shouldn't otherwise impact the component negatively, since it should be support the older version of the prop for backwards compat.

But these stories test the v2, so given that we already deprecated the experimental prop, I'd expect we override the options to use the stable prop. Not a deal breaker, but can be confusing right now IMHO.

tyxla (approved):
LGTM 👍 

The deprecated option prop usage in the story is not a blocker IMHO, up to you if you want to address it or not before 🚢

mirka (approved):
🚀

PR #63258, opened 2024-07-08
CustomSelectControl: switch to ariakit-based implementation

ciampo:
## What?

Part of #55023 

Replace the legacy, `downshift`-based implementation of `CustomSelectControl`, with the new `ariakit`-based implementation. The public APIs of the component shouldn't be affected.

## Why?

For a few reasons:

- remove `downshift` as a dependency of the project
- better accessibility of the component
- better alignment with other ariakit-based components
- a starting point for the next version of the component

See #55023 for more info.

## How?

  - Removed V1 implementation, replaced with "V2 legacy adapter)
  - Removed downshift as dependency
  - Removed private export of V1
  - Removed V2 legacy adapter
  - Added back legacy classnames to the new implementation
  - Updated and aligned README and JSDocs

> [!NOTE]
> For now, the new `ariakit` V1 imports base components and styles from the V2. As we further the development of the V2 version, we may have to "duplicate" this code so that the V1 implementation becomes standalone, and the V2 can diverge without worrying about repercussions on V1.

## Testing Instructions

_I recommend reviewing this PR commit-by-commit to minimize the noise_

- Make sure project builds and all tests pass
- Make sure Storybook examples are associated with the right folder path, and that controls work as expected
- Smoke test both in Storybook (CustomSelectControl, FontSizePicker) and in the editor
- Make sure there are no mentions of `downshift` or `CustomSelectControlV2` outside of the source folders

## ✍️ Dev note

The `CustomSelectControl` component has been completely rewritten using `ariakit`. Notably, this refactor enables the removal of the `downshift` library as a dependency of the Gutenberg repository, causing a `~7.75 kB` decrease for the `@wordpress/components` package bundle size.

While there are no breaking changes to the component public APIs, the refactor also allowed us to make a few improvements to the component:

- the trigger button has the `combobox` role instead of `button`, in line with the WAI-ARIA specs;
- the `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props have been unprefixed and stabilized;
- the select popover has a new entrance animation when opened, in line with other components;
- while the select popover's `max-height` hasn't changed, the popover's height adapts to the available space, without causing extra overflow on the page;
- item hints can now wrap to a separate row from the item's content when there isn't sufficient space to render on the same line;
- item internal padding now matches the trigger button padding for better visual consistency.


packages/components/src/custom-select-control/README.md
```diff
@@ -63,81 +63,77 @@ function MyControlledCustomSelectControl() {
...
+Optional classname for the component.
 
--   Type: `String`
 -   Required: No
 
-#### hideLabelFromVision
+#### `hideLabelFromVision`: `boolean`
 
-Used to visually hide the label. It will always be visible to screen readers.
+Hide the label visually, while keeping available to assistive technology
```
mirka:
```suggestion
Hide the label visually, while keeping available to assistive technology.
```

packages/components/src/custom-select-control/README.md
```diff
@@ -63,81 +63,77 @@ function MyControlledCustomSelectControl() {
...
 -   Required: Yes
 
-#### onChange
+#### `size`: `'default' | 'small' | '\_\_unstable-large'`
 
-Function called with the control's internal state changes. The `selectedItem` property contains the next selected item.
+The size of the control.
 
--   Type: `Function`
+-   Default: `false`
```
mirka:
```suggestion
-   Default: `"default"`
```

packages/components/src/custom-select-control/types.ts
```diff
@@ -0,0 +1,111 @@
+/**
+ * External dependencies
+ */
+import type { FocusEventHandler, MouseEventHandler } from 'react';
+
+/**
+ * The legacy object structure for the options array.
```
mirka:
```suggestion
 * The object structure for the options array.
```

There are a couple other mentions of "legacy" in the description comments in this file.

packages/components/src/custom-select-control/test/index.tsx
mirka, line 1: Should we also rename the `legacyProps` → `props` in this file?

mirka (approved):
Looks good! This folder structure feels correct to me too 👍

packages/components/src/custom-select-control/test/index.tsx
ciampo: Done!

packages/components/src/custom-select-control/types.ts
ciampo: Good catch, done

ciampo:
Thank you for the feedback. I'll have a final round of smoking tests and merge tomorrow

andrewserong (commented):
Hi @ciampo and @mirka, I just noticed that the font appearance control is sitting at a different position vertically than all the other controls in the Typography panel now. Could that have been caused by this PR? It seems a possible candidate, but I haven't done a detailed look into other recent changes to this component. Here's a quick screenshot of how it's looking in `trunk`:

image

Note that the Appearance label sits lower than Line Height along with its select box.

I've opened an issue over in https://github.com/WordPress/gutenberg/issues/63886.

PR #63266, opened 2024-07-08
SelectControl: Fix disabled styles

mirka:
## What?

Fixes `disabled` styles for `SelectControl` (caret color and cursor style).

## Testing Instructions

See the Storybook for `SelectControl`.

## Screenshots or screencast 

| Before | After |
|--------|-------|
|Wrong disabled styles|Correct disabled styles|


ciampo (approved):
🚀 

Looks like some snapshots need to be updated, but everything else LGTM

PR #45420, opened 2022-10-31
hide insertion point when moving out of the canvas

youknowriad:
fixes #45418

## What?

It's very tricky to address all the situations where we have to hide the insertion point. This one for instance is about leaving  the canvas in Chrome.

## Testing Instructions

1- Hover between two blocks to show the insertion point while the sidebar is open
2- Move the mouse directly into the sidebar
3- The insertion point should hide itself.

This PR address one case, it's not clear whether it's the only case or not.

PR #63270, opened 2024-07-08
In-between inserter: allow scrolling over indicator

### LINKED ISSUE #63260: In-between Inserter: can't scroll over it
### Description

This is a very frustrating bug, and it was once fixed. Any time you try to scroll and the pointer is accidentally between any block, scrolling is blocked. I run into this over and over again. Note that you can't even scroll before the inserter shows up!

Try to scroll with the pointer over the in-between inserter.

image

_No response_

### Environment info

_No response_

Yes

Yes

ellatrix:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/63260.

## Why?

Because it's super annoying. :) 

## How?

Looks like pointer-events: all was re-applied to early on the entire indicator instead of just on the button.

## Testing Instructions

Hover over the space in between blocks and scroll.

Make sure that #45420 does not regress.

### DIFF
--- a/packages/block-editor/src/components/block-list/use-in-between-inserter.js
+++ b/packages/block-editor/src/components/block-list/use-in-between-inserter.js
@@ -161,7 +161,7 @@ export function useInBetweenInserter() {
 
 				if ( event.type === 'click' ) {
 					hideInsertionPoint();
-					selectBlock( clientId );
+					selectBlock( clientId, -1 );
 					return;
 				}
 
--- a/packages/block-editor/src/components/block-list/use-in-between-inserter.js
+++ b/packages/block-editor/src/components/block-list/use-in-between-inserter.js
@@ -30,7 +30,7 @@ export function useInBetweenInserter() {
 		getBlockEditingMode,
 		getBlockName,
 	} = useSelect( blockEditorStore );
-	const { showInsertionPoint, hideInsertionPoint } =
+	const { showInsertionPoint, hideInsertionPoint, selectBlock } =
 		useDispatch( blockEditorStore );
 
 	return useRefEffect(
@@ -159,15 +159,23 @@ export function useInBetweenInserter() {
 					return;
 				}
 
+				if ( event.type === 'click' ) {
+					hideInsertionPoint();
+					selectBlock( clientId );
+					return;
+				}
+
 				showInsertionPoint( rootClientId, index, {
 					__unstableWithInserter: true,
 				} );
 			}
 
 			node.addEventListener( 'mousemove', onMouseMove );
+			node.addEventListener( 'click', onMouseMove );
 
 			return () => {
 				node.removeEventListener( 'mousemove', onMouseMove );
+				node.removeEventListener( 'click', onMouseMove );
 			};
 		},
 		[
--- a/packages/block-editor/src/components/block-popover/style.scss
+++ b/packages/block-editor/src/components/block-popover/style.scss
@@ -30,18 +30,10 @@
 	// drop zone.
 	pointer-events: none;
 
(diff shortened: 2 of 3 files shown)


packages/block-editor/src/components/block-list/use-in-between-inserter.js
```diff
@@ -159,15 +159,23 @@ export function useInBetweenInserter() {
...
 					__unstableWithInserter: true,
 				} );
 			}
 
 			node.addEventListener( 'mousemove', onMouseMove );
+			node.addEventListener( 'click', onMouseMove );
 
 			return () => {
 				node.removeEventListener( 'mousemove', onMouseMove );
+				node.removeEventListener( 'click', onMouseMove );
```
ellatrix, line 178: This onClick handler was previously in the popover, but we can handled it here since it's now click-through.

packages/block-editor/src/components/block-popover/style.scss
```diff
@@ -30,18 +30,10 @@
 	// drop zone.
 	pointer-events: none;
 
-	* {
-		pointer-events: none;
```
ellatrix, line 34: `pointer-events` is inherited.

packages/block-editor/src/components/block-tools/insertion-point.js
```diff
@@ -81,36 +79,9 @@ function InbetweenInsertionPointPopover( {
...
 
 	const disableMotion = useReducedMotion();
 
-	function onClick( event ) {
-		if (
-			event.target === ref.current &&
-			nextClientId &&
-			getBlockEditingMode( nextClientId ) !== 'disabled'
-		) {
-			selectBlock( nextClientId, -1 );
```
ellatrix, line 94: As noted above, this has moved to the canvas instead.

packages/block-editor/src/components/block-tools/insertion-point.js
```diff
@@ -81,36 +79,9 @@ function InbetweenInsertionPointPopover( {
...
-		}
-	}
-
-	function maybeHideInserterPoint( event ) {
-		// Only hide the inserter if it's triggered on the wrapper,
-		// and the inserter is not open.
-		if ( event.target === ref.current && ! openRef.current ) {
-			hideInsertionPoint();
```
ellatrix, line 102: These things seem to be workarounds to the fact that the indicator caught clicks (and didn't trigger the `mousemove` event listener in the canvas, so it had to be duplicated here.

ellatrix:
It's getting a bit too complex to consider for 6.6.

talldan:
> It's getting a bit too complex to consider for 6.6.

Is it possible to reduce it down to just the CSS changes for 6.6 or are all the JS changes required too?

I tested and it works, but I'd like a bit more time to understand the JS changes.

ellatrix:
@talldan No, not possible unfortunately because by setting pointer events to none, those event handlers will basically never be called.

talldan:
@ellatrix It'd be good to revisit this one and finally get it fixed. If you want to rebase the PR I'd be happy to give it a test and 👍 if everything's still working.

im3dabasia:
> Is it possible to reduce it down to just the CSS changes for 6.6 or are all the JS changes required too?

I have a only CSS change. I'm not sure if it is correct or not. Can I raise a PR for you to check, @talldan?

t-hamano:
I think #63260 is already resolved by #68075, so I would like to close this PR. Please reopen if I am wrong.

By the way, note that another problem is being introduced by #68075. See https://github.com/WordPress/gutenberg/issues/70594 for more details.

PR #63253, opened 2024-07-08
Add local version of wp-env schema to .wp-env.json

t-hamano:
Similar to #61312

## What?

This PR adds a reference to the local schema in the wp-env config file used by the Gutenberg project.

## Why?

- This will prevent incorrect fields from being defined in future updates to `.wp-env.json`.
- As mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/36276#issuecomment-2214222158), there are some fields missing in the current schema. When fixing these, it will be easier to test using the `.wp-env.json` in the Gutenberg project.

## Testing Instructions

- Check out this PR.
- Open the `.wp-env.json` file in a code editor.
- The code editor should help you auto-complete fields and warn you about invalid fields and values.

PR #63281, opened 2024-07-09
.wp-env.json schema: Fix schema and add unit tests

t-hamano:
Follow-up #63253
Implement the suggestion in [this comment](https://github.com/WordPress/gutenberg/pull/63253#pullrequestreview-2163673709)

## What?
This PR adds a unit test to validate the wp-env meta schema and the `.wp-env.json` present in the project.

At the same time, it fixes errors in the metadata detected by the tests.

## Why?

To ensure that we provide a correct meta-schema for developers.

## How?

The implementation is similar to the unit tests for other schemas.

## Testing Instructions

- Modify the [`.wp-env.json` file](https://github.com/WordPress/gutenberg/blob/trunk/.wp-env.json) and add the invalid value or property.
- `npm run test:unit test/integration/wp-env-schema.test.js `
- The `validates schema for ".wp-env.json"` test should fail.

⚠️ Note: One strange thing is that when I run this unit test _locally_, I get the following error:

```
["strict mode: missing type \"object\" for keyword \"properties\" at \"#/allOf/1\" (strictTypes)"],["strict mode: missing type \"object\" for keyword \"propertyNames\" at \"#/allOf/2\" (strictTypes)"]
```

Please let me know if this problem occurs in your environment.


test/integration/wp-env-schema.test.js
```diff
@@ -0,0 +1,43 @@
...
+import Ajv from 'ajv';
+import glob from 'fast-glob';
+
+/**
+ * Internal dependencies
+ */
+import wpEnvSchema from '../../schemas/json/wp-env.json';
+
+describe( '.wp-env.json schema', () => {
+	const jsonFiles = glob.sync( [ '.wp-env.json' ], { onlyFiles: true } );
```
t-hamano: I don't think any new `.wp-env.json` files will be added in the future, but I am concerned that there may be a better pattern.

ajlende: I'd remove the `jsonFiles` array entirely if there's only one.

- Import the `.wp-env.json` file just like the schema as an internal dependency.
- Remove the `'found .wp-env.json files'` test.
- Replace the `test.each` with a singular test.
- Replace the `require( filepath )` with the imported variable.

ajlende (approved):
> ⚠️ Note: One strange thing is that when I run this unit test locally, I get the following error:

I wasn't seeing any errors when running locally.

So this looks good to me 👍

PR #63282, opened 2024-07-09
File block: Do not persist blob urls and fix undo

youknowriad:
Related #39223 
Similar to #63076 and #63238 

## What?

In #63076, we introduced the concept of local attributes to allow setting attributes that won't be persisted. It's useful during file uploads for instance where we want immediate user feedback but without impacting the saved post.

In the previous PR, we implemented it for image, audio and video blocks and this PR does the same for the file block. This PR also fixes an undo bug in the file block.

## Testing Instructions

1- Throttle your network speed (that way uploading file will take a long time)
2- Drag and drop a pdf file into the post editor.
3- Save the post immediately.
4- Open the post in a new tab and notice that the inserted file block is actually empty.
6- Notice that the file block has been updated properly with the right URL.
You can also try undoing uploads and it should results in empty file blocks.

### DIFF
--- a/packages/block-library/src/file/edit.js
+++ b/packages/block-library/src/file/edit.js
@@ -23,7 +23,7 @@ import {
 	store as blockEditorStore,
 	__experimentalGetElementClassName,
 } from '@wordpress/block-editor';
-import { useEffect } from '@wordpress/element';
+import { useEffect, useState } from '@wordpress/element';
 import { useCopyToClipboard } from '@wordpress/compose';
 import { __, _x } from '@wordpress/i18n';
 import { file as icon } from '@wordpress/icons';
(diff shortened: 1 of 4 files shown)


ellatrix (approved):
Nice. How many of these are there? 😅

packages/block-library/src/file/edit.js
```diff
@@ -120,7 +127,9 @@ function FileEdit( { attributes, isSelected, setAttributes, clientId } ) {
 			displayPreview: isPdf ? true : undefined,
 			previewHeight: isPdf ? 600 : undefined,
 			fileId: `wp-block-file--media-${ clientId }`,
+			blob: undefined,
 		} );
+		setTemporaryURL();
```
ellatrix, line 132: I wonder if it might make sense to have a helper util for this

youknowriad:
This is the last one :)

packages/block-library/src/file/edit.js
youknowriad: I think there's a lot of duplication between how these media blocks are setup, a lot could be share but there are also some things that are to explain: some use "url" others "href" and others "src" for the uploaded files

Mamaduka:
The Cover and Media & Text block might need an audit as well. I've not worked on these blocks in a while, but anything using  `mediaUpload` + storing URLs in the content will have similar issues.

PR #63286, opened 2024-07-09
Add example for query block and posts list

MaggieCabrera:
## What?

This adds an example for `core/query` and the variation posts list in the inserter

## Why?

## How?

By including the example attributes and settings in the block's index.js

## Testing Instructions

Open the block inserter and search for the query or posts list block, hover over it and you'll see the example. 

## Screenshots or screencast 

Screenshot 2024-07-09 at 10 27 51

### DIFF
--- a/packages/block-library/src/query/index.js
+++ b/packages/block-library/src/query/index.js
@@ -19,6 +19,46 @@ export { metadata, name };
 export const settings = {
 	icon,
 	edit,
+	example: {
+		viewportWidth: 650,
+		attributes: {
+			namespace: 'core/posts-list',
+			query: {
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I went with the layout I felt showed more clearly what this does. I'm not sure if we can have a different example for query and posts lists, but I'm thinking the answer is no (without some heavy modifications), since they are both the same block. Please advice in terms of the design/appearance of this.

jasmussen (approved):
Yep.

Before:
Screenshot 2024-07-09 at 11 43 27

After:
Screenshot 2024-07-09 at 11 43 07

PR #63287, opened 2024-07-09
DataViews: Replace supportedLayouts prop with defaultLayouts prop instead 

youknowriad:
Related #55083 

## What?

Each time we change the "layout" of the DataViews component, we often need to change the fields configuration or tweak the "layout" property of the "view" object... Basically what we're doing here is that we're setting a default value for the "view" object (keeping filters, search...) 

Also, on the other hand, when using the DataViews component, we define the "supportedLayouts" prop.

For me, this suggests that these two things are the same and only thing. This PR combines these props in a unique `defaultViews` prop. This is going to be more important with the introduction of field combination as the differences between view types are growing over time.

## Testing Instructions

This is an API change, but there should be no functional change.

### DIFF
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -24,7 +24,13 @@ import {
 } from './bulk-actions';
 import { normalizeFields } from './normalize-fields';
 import BulkActionsToolbar from './bulk-actions-toolbar';
-import type { Action, Field, View, ViewBaseProps } from './types';
+import type {
+	Action,
+	Field,
+	View,
+	ViewBaseProps,
+	SupportedLayouts,
+} from './types';
 import type { SetSelection, SelectionOrUpdater } from './private-types';
 
 type ItemWithId = { id: string };
@@ -42,7 +48,7 @@ type DataViewsProps< Item > = {
 		totalItems: number;
 		totalPages: number;
 	};
-	supportedLayouts: string[];
+	defaultLayouts: SupportedLayouts;
 	selection?: string[];
 	setSelection?: SetSelection;
 	onSelectionChange?: ( items: Item[] ) => void;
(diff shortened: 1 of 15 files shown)


packages/dataviews/src/dataviews.tsx
```diff
@@ -42,7 +48,7 @@ type DataViewsProps< Item > = {
 		totalItems: number;
 		totalPages: number;
 	};
-	supportedLayouts: string[];
+	defaultLayouts: SupportedLayouts;
```
ntsekouras, line 51: Not from here, but probably better to move all types to the respective file.

packages/dataviews/src/filter-and-sort-data-view.ts
```diff
@@ -61,7 +61,7 @@ export function filterSortAndPaginate< Item >(
 		} );
 	}
 
-	if ( view.filters?.length > 0 ) {
+	if ( view.filters && view.filters?.length > 0 ) {
```
ntsekouras, line 64: What does this change?

packages/dataviews/src/dataviews.tsx
youknowriad: which respective file?

packages/dataviews/src/filter-and-sort-data-view.ts
youknowriad: I made view.filters optional that way folks don't have to define it if they don't have filters.

ntsekouras (approved):
Looks good and tests well. Thank you!

packages/dataviews/src/filter-and-sort-data-view.ts
ntsekouras: Yeah, I meant why the previous `view.filters?.length` wasn't enough..

packages/dataviews/src/dataviews.tsx
ntsekouras: `types.ts`

youknowriad: I prefer for the `types.ts` to stay mostly about the structural types of the package and the temporary types that are specific to components... I prefer to keep them next to the components directly. These types don't have any value as shared types.

Issue #61210, opened 2024-04-29
Typewriter Experience doesn't work when Custom Fields are shown


jsit:
### Description

When the Custom Fields panel is shown, the Typewriter Experience / Typewriter Scrolling (#16460) doesn't work.

### Step-by-step reproduction instructions

1. In Editor Preferences on the Edit page, enable "Custom fields"
2. Save pref and reload the page
3. See that cursor is able to reach the bottom of the editor window, and can't be scrolled past

### Environment info

- WordPress 6.5.2
- Gutenberg version...unsure? Whatever is bundled with WP 6.5.2

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

jhimross:
@jsit I am not sure if there's an issue here. I tried replicating the issue by following the instructions you provided above.
I attempted to replicate the issue using WP 6.5.2, the Twenty Twenty-Four theme, and no other plugins installed.

I enabled the Custom Fields, and this is what I see. Watch: https://share.zight.com/JruE7ePK

No issues were experienced. The typewriter experience still works as expected and can be scrolled.

Could you please provide more details to replicate the issue, or if possible, provide a screenshot or screen recording?

jsit:
@jhimross Hm, I don't know what we are doing differently. This screen recording was taken on a WP 6.5.3 install, with the Twenty Twenty-Four theme, and no active plugins:

https://www.dropbox.com/scl/fi/usy2nhrdube8109brmsxk/Screen-Recording-2024-05-10-at-12.51.53-PM.mov?rlkey=6d5cwhzdwd91gem1k6r6y4z1y&e=1&st=eoaois5o&dl=0

This is in Safari, but I see the same issue in Chrome.

jhimross:
Tried replicating the issue again using WP 6.5.3, Twenty Twenty Four theme, and no active plugins. But I still can't replicate it using different browsers.

This is what's happening when using:
Chrome and Edge in Windows:- https://share.zight.com/2Nu8rnWm
Using Safari 17.3 :- https://share.zight.com/mXuWmwvX

jsit:
What you are showing in the first video is what I’m describing. When you scroll down with no Custom Fields, there is a lot of space between the final block and the bottom of the input area; you can be in the final block and be typing in the “middle” of the window. You can’t do this if Custom Fields are shown.

jsit:
[...]

PR #59242, opened 2024-02-21
Move legacy meta boxes to toggle

ellatrix:
## What?

(would also fix #61210)

image

## Why?

Legacy meta boxes prevent the editor content from loading in an iframe. We cannot have two scrollable areas in the content area (and both the iframe and meta box area would be scrollable).

See #34105 for more background.

### Testing Instructions for Keyboard

Go to Options > Preferences > General > Advanced and toggle custom fields.

PR #63288, opened 2024-07-09
Post editor: increase specificity bottom padding

### LINKED ISSUE #63278: Post Editor: 40vh padding applied to the bottom is overwritten with zero
### Description

In WP6.6, when `useRootPaddingAwareAlignments` is `false` in `theme.json`, the expected `40vh` padding below the content in the post editor is overwritten with zero.

This makes it difficult to click below the content and insert a new block.

I think the issue is related to the CSS selector that applies the zero padding changing from `body` to `:root :where(body)`, which has a higher specificity.

### ✅ useRootPaddingAwareAlignments: true

image

### ✅ useRootPaddingAwareAlignments: false

image

### ✅ useRootPaddingAwareAlignments: true

image

### ❌ useRootPaddingAwareAlignments: false

override-padding

- Define the theme.json for your block theme as follows:
  ```json
  {
  	"version": 2,
  	"settings": {
  		"layout": {
  			"contentSize": "620px",
  			"wideSize": "1280px"
  		}
  	}
  }
  ```
- Open the post editor and insert some content.
- Scroll to the very bottom of the content.
- There will be no padding below the content.

## Actual

actual

## Expected

expected

### Environment info

WordPress 6.6 RC2 and latest Gutenberg plugin

Yes

Yes

ellatrix:
## What?

Increases the specificity of the body bottom padding to override any padding set by global styles.

## Why?

Fixes #63278.

## How?

Set to 0-1-0. Alternatively we could use `!important`, but I guess we better avoid that if we know the specificity needed.

## Testing Instructions

See #63278.

### DIFF
--- a/packages/edit-post/src/components/layout/index.js
+++ b/packages/edit-post/src/components/layout/index.js
@@ -138,7 +138,9 @@ function useEditorStyles() {
 			return [
 				...baseStyles,
 				{
-					css: 'body{padding-bottom: 40vh}',
+					// Should override global styles padding, so ensure 0-1-0
+					// specificity.
+					css: ':root :where(body){padding-bottom: 40vh}',
 				},
 			];
 		}


t-hamano (approved):
Thanks for the fix!  It appears to work correctly in all scenarios.

---

- Classic Theme, Post Editor is iframed:

image

- Classic Theme, Post Editor is not iframed:

image

- Block Theme, Post Editor is iframed, `useRootPaddingAwareAlignments` is `true`:

image

- Block Theme, Post Editor is iframed, `useRootPaddingAwareAlignments` is `false`:

image

---

The following two points cannot be tested in the usual way on the Gutenberg plugin. This is because [if the Gutenberg plugin is enabled and the theme is a block, the editor will always be iframed](https://github.com/WordPress/gutenberg/blob/f8aa83a9eea998cce30b7f4a1e1c5f9d2feafe9d/packages/edit-post/src/components/layout/use-should-iframe.js#L35-L39).

I temporarily removed the `|| ( isGutenbergPlugin && isBlockBasedTheme )` condition and tested a block theme, non-iframe editor.

- Block Theme, Post Editor is not iframed, `useRootPaddingAwareAlignments` is `true`:

image

- Block Theme, Post Editor is not iframed, `useRootPaddingAwareAlignments` is `false`:

image

---

Another issue I noticed while testing this PR is that when the editor is not iframed because a meta box is present, the bottom padding is not applied:

image


On the other hand, when the meta box is not present but a v2 or lower block is registered and the editor is not iframed, the bottom padding is applied.

However, I found that this issue also occurs in WP6.5, so I don't think it's a blocker for this PR.

ellatrix:
Yes, I think this was reported here #61210. I think that was once done so they are not pushed under the fold. Ideally we fix this by doing #59242 in 6.7.

PR #59039, opened 2024-02-14
Interactivity API: Only add proxies to plain objects inside the store

cbravobernal:
## What?

Refactor the isObject function to only accept plain objects as stores.

## Why?

Before this PR, the code was only checking with `typeof`. In JavaScript, almost everything is an Object. We want to make sure that only plain objects are proxified. (Proxify means to add a getter that handles the namespace of the block/plugin and the Scope).

PR #63290, opened 2024-07-09
Image block: Remove unnecessary variables on expand on click implementation

luisherranz:
## What?

I have deleted some unnecessary variables from the implementation of the expand-on-click feature of the image block.

## Why?

Because at the time this implementation was done, there was a bug in the Interactivity API: when an element was added to the context or the state, the internal variables of that element tried to become reactive and failed.

That bug was fixed in [this pull request](https://github.com/WordPress/gutenberg/pull/59039), so the variables in this implementation that stored the reference of the button and the reference of the image are no longer needed.

## How?

Simply deleting those variables and letting everything be in the context/state.

## Testing Instructions

1. Add an image to a post.
2. Enable the "Expand on click" option.
3. Open the post on the frontend and check that everything works correctly.

### DIFF
--- a/packages/block-library/src/image/view.js
+++ b/packages/block-library/src/image/view.js
@@ -19,20 +19,6 @@ let isTouching = false;
  */
 let lastTouchTime = 0;
 
-/**
- * Stores the image reference of the currently opened lightbox.
- *
- * @type {HTMLElement}
- */
-let imageRef;
-
-/**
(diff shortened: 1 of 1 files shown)


madhusudhand (approved):
Thanks for refactoring. Verified the fix and it works as before without any issues.
🚢

cbravobernal (approved):
Nice, thanks @luisherranz !

PR #46335, opened 2022-12-06
Navigation: Add a 'open list view' button

scruffian:
## What?
This is a replacement for https://github.com/WordPress/gutenberg/pull/45853.
This updates the list view button on the navigation block in the site editor to open the inspector controls, and also adds a new button to the navigation block in the post editor to open the inspector controls.

## Why?
This is a bit controversial - on the one hand we shouldn't need a button to simply open the inspector controls. On the other hand, if users have the navigation block selected but they don't have the inspector controls open, this could be a useful prompt to open them.

## How?
Adds a plugin to the edit-post package and updates the current list view button in edit-site.

## Testing Instructions
1. Open a new post
2. Add a navigation block
3. Add some items
4. Notice that there is "Open navigation list view" button in the Block Toolbar:
Screenshot 2022-12-06 at 16 55 15
5. Click the button and confirm it opens the inspector controls for the Navigation block:
Screenshot 2022-12-06 at 16 56 50
6. Do the same test in the site editor

### Testing Instructions for Keyboard
1. Open a new post
2. Add a navigation block
3. Add some items
4. Tab to the block toolbar
5. Use the arrow keys to move tab focus in the block toolbar
6. Select the navigation block
7. Move the focus to the "open navigation list view" button
8. Select this button
9. Confirm that the inspector controls open
10. Do the same test in the site editor

## Screenshots or screencast 

https://user-images.githubusercontent.com/275961/205975435-f9e98cf4-3f18-4b73-82d7-826959d5015e.mp4

PR #39290, opened 2022-03-08
Navigation: add dedicated sidebar for managing layout of navigation menus

gwwar:
## What?

This PR adds a new navigation sidebar in the site editor. This sidebar replaces the list view modal, and allows one to see the contents of navigation menus even if they aren't in the current template.

https://user-images.githubusercontent.com/444434/158396446-53c04c19-5f7b-4748-8f49-1867db28ad14.mov

## Why?

Part of explorations for https://github.com/WordPress/gutenberg/issues/36667.

This PR pulls out a more pragmatic piece of https://github.com/WordPress/gutenberg/pull/38600 out to focus specifically on a Navigation sidebar.

~We may find that it's still a little early to display a new top level panel while we iterate on UX options, so let me know if folks would like to add a feature flag.~ @getdave: I suggest we ship and iterate based on feedback.

## How?

A couple of key points in this PR:
- A new selector `__experimentalGetGlobalBlocksByName` allows us to display the correct current menu when the panel is open.
- Using a `BlockEditorProvider` gives us the editing and save functionality we mostly expect, however **there are side-effects that occur in the rendering of BlockList**. This means that we need to think through how to provide information like allowed block information (which blocks may be inserted into another), when BlockList is not rendered (as is the case for ListView). I've added a temporary workaround by dispatching this information manually. Folks should discuss if this temporary patch is worthwhile, or if we should think of a better solution with some refactoring of block list. Note that the current solution is fragile and will drift, but it does fix dragging and has a relatively small surface area.
- A new slotfill in blocktoolbar to decouple the navigation block from having to know about a site-editor component.
- We only had interest in adding this panel to the site editor, so this PR effectively removes the list view modal in the post-editor without an alternative. @getdave: I think that's ok since we already have the standard List View available in the Post Editor which does the exact same thing.
- ~List View has gained new properties to allow all levels to be collapsed by default.~ (now extracted into https://github.com/WordPress/gutenberg/pull/39486)

## Todo

- [x] Collapse nested levels by default
- [x] Improve "mover" arrows placement for long menu items.
- [ ] ~Disable `Show more settings` option  under Block Actions - doesn't work unless Nav block is selected in editor canvas.~ Out of scope for this PR.
- [x] Extract all changes to `ListView` to a separate PR - now done in:
    -  https://github.com/WordPress/gutenberg/pull/39486.
    - https://github.com/WordPress/gutenberg/pull/39494/
- [x] Extract any other changes that are making this PR overly "noisy" (too many changes):
[...]

PR #63291, opened 2024-07-09
Pattern overrides: Ensure "Reset" button always shows as last item and with border

### LINKED ISSUE #62310: Pattern overrides: The image block toolbar has no separator before the Reset option
### Description

Blocks inside synced patterns that have override enabled have a reset option in the block toolbar.
The heading, paragraph and button block all have a separator in the toolbar between the block options and the reset (they all use the rich text options), but it is missing from the image block.

Create a synced pattern with an image block inside.
Edit the original pattern, select the image, open the Advanced panel in the settings sidebar and enable overrides. Save,
Go back to where you had placed the pattern.
Select the image and view the toolbar.

image-block-toolbar-reset

### Environment info

Gutenberg trunk
WP 6.6-beta1-58338, with the beta tester plugin.

Yes

No

SantosGuillamot:
## What?
Fixes https://github.com/WordPress/gutenberg/issues/62310

It ensures that the "Reset" button added by pattern overrides always shows as the last item and with the appropriate borders.

## Why?
To keep consistency across the different blocks.

## How?
Using the `__unstableBlockToolbarLastItem` component. Is it okay to use this component?

## Testing Instructions
1. Create a synced pattern with an image block inside.
2. Edit the original pattern, select the image, open the Advanced panel in the settings sidebar and enable overrides. Save,
3. Go back to where you had placed the pattern.
4. Select the image and view the toolbar.

**Before: The "Reset" button was in the same group as "Replace, Alt, Title".**

Screenshot 2024-07-09 at 12 42 34

**After: The "Reset" button has its own group.**

Screenshot 2024-07-09 at 12 37 03

### DIFF
--- a/packages/patterns/src/components/reset-overrides-control.js
+++ b/packages/patterns/src/components/reset-overrides-control.js
@@ -3,7 +3,7 @@
  */
 import {
 	store as blockEditorStore,
-	BlockControls,
+	__unstableBlockToolbarLastItem as BlockToolbarLastItem,
 } from '@wordpress/block-editor';
 import { ToolbarButton, ToolbarGroup } from '@wordpress/components';
 import { useRegistry, useSelect } from '@wordpress/data';
@@ -79,12 +79,12 @@ export default function ResetOverridesControl( props ) {
 	}
 
 	return (
-		<BlockControls group="other">
+		<BlockToolbarLastItem>
 			<ToolbarGroup>
 				<ToolbarButton onClick={ onClick } disabled={ ! isOverriden }>
 					{ __( 'Reset' ) }
 				</ToolbarButton>
 			</ToolbarGroup>
-		</BlockControls>
+		</BlockToolbarLastItem>
 	);
 }


packages/patterns/src/components/reset-overrides-control.js
```diff
@@ -3,7 +3,7 @@
  */
 import {
 	store as blockEditorStore,
-	BlockControls,
+	__unstableBlockToolbarLastItem as BlockToolbarLastItem,
```
ellatrix, line 6: It looks like we're not currently using that for anything. 🤔

ellatrix: It was introduced in #39290.

ellatrix: Maybe @youknowriad has an opinion here (I think you added the grouping once upon a time). But I'm not against this approach.

SantosGuillamot: You're right it is not used anymore 🤷 It seems it was also used in [this pull request](https://github.com/WordPress/gutenberg/pull/46335), and the initial implementation was "removed" in [this other one](https://github.com/WordPress/gutenberg/pull/46335).

talldan (approved):
Thanks for fixing this @SantosGuillamot.

I don't know much about `BlockToolbarLastItem`, but looks like [it's used by plugins](https://www.wpdirectory.net/search/01J2DE1XWVZ2YZVBVTHKZGKGP4), so I don't think it can be removed now.

It might be a good one to follow up and stabilize it, maybe switching it to a regular `BlockControls.Slot`. I'd be happy to look at that.

ellatrix:
But if it's used by plugins, it's not guaranteed to be last? 🙈 
I guess that's still fine though.
Btw, this will have to be in 6.6.1

youknowriad:
Actually, I don't understand this PR at all. We do control the pattern block, why use a hook to inject `withPatternOverrideControls` instead of just moving this code into the block itself?

youknowriad:
Oh I guess the button is not added to the "pattern block" but to the children. Ok Maybe this is fine :)

youknowriad:
Although I do wonder about the UX, it seems to me that a global "Reset" on the pattern block is probably better for the user than sprinkled reset button on all the children.

SantosGuillamot:
> Although I do wonder about the UX, it seems to me that a global "Reset" on the pattern block is probably better for the user than sprinkled reset button on all the children.

If I am not mistaken, there is a global "Reset" button that removes all the changes applied to the pattern and another "Reset" button for each block with overrides enabled to be able to remove the changes to that specific block.

SantosGuillamot:
I'm merging this pull request, but I'm happy to explore other paths if considered necessary.

ellatrix:
✅ Looks good to me to include in the next minor.

SantosGuillamot:
I agree it should be safe to include. Although it's also true that it doesn't seem critical in case it doesn't make it into the fast release.

PR #61644, opened 2024-05-14
Post Actions: Pass entity permission to the 'isEligible' method

Mamaduka:
## What?
🚧 PoC for introducing entity permission to the "Post Actions"

## Testing Instructions
1. Create a test use with a Contributor role.
2. Remove `delete_posts` capability - `wp cap remove 'contributor' 'delete_posts'`
3. Create a post and save it as a draft.
4. The big red "Move to Trash" button shouldn't be visible
5. The "Move to Trash" action shouldn't be visible.

### Testing Instructions for Keyboard
Same.

## Screenshots or screencast 
CleanShot 2024-05-14 at 12 49 27

Issue #61637, opened 2024-05-14
Expose the Site Editor Patterns page to all users


t-hamano:
Further improve #52150

## What problem does this address?

#52150 aims to expose the Site Editor Patterns page for every "theme".

However, currently the Site Editor is only accessible to admin users. For non-admin users, links related to the Pattern page link to the classic post list (`edit.php?post_type=wp_block`).

## What is your proposed solution?

We should be able to remove this restriction and allow all users to access the Site Editor Patterns page.

To achieve this, I think we need to implement at least the following points:

- Even if the current theme is a block theme, non-admin users are not allowed to access anything other than the Patterns page.
- Check capabilities to restrict actions such as creating, duplicating, deleting, and editing patterns. The "wp_block" post type has [complex capabilities](https://github.com/WordPress/wordpress-develop/blob/ff2c933afa7e432084b3e1d2540a1f6c2bc54eb7/src/wp-includes/post.php#L315-L327).
- Implement redirects in case the post list page is bookmarked or the post list page is hard-coded by third-party product developers.

t-hamano:
There may be a better approach, but the `capabilities` seem to be available through the API. Works with all roles except Subscriber role.

```javascript
const currentUserId = wp.data.select('core').getCurrentUser().id
wp.data.select('core').getUser(currentUserId)?.capabilities
```

![image](https://github.com/WordPress/gutenberg/assets/54422211/aac1040a-163c-48f8-b1d6-fc5bd859466e)

youknowriad:
Throughout our code base we generally use `canUser` "core" selector to check capabilities. 

@jsnajdr Just noting that this issue might mean introducing "protected routes" in the site editor routing.

jsnajdr:
That should place any special demands on the router: just register only these routes that correspond to the `canUser` capabilities, e.g., `canUser( 'edit_theme_options' )`, and limit what is shown in the top-level sidebar.

We'll have to preload the user REST data, if they aren't being preloaded already.

Security concerns (user can't see certain data) should be fully addressed by the REST API -- the responses shouldn't contain any data that the non-admin user is not supposed to see.

Calypso, the WordPress.com dashboard SPA, has been doing all this (preloading the user, conditional routes and sidebars) since forever, it's not a big deal.

t-hamano:
I hope to address this issue at least a little bit for WP6.7.

[...]

PR #12378, opened 2018-11-28
Reusable blocks: Improve UX for non-privileged users

noisysocks:
### What this is

Fixes #12338.

Improves the UX of creating, editing, and deleting a reusable block when logged in as an author or contributor by disabling the _Add to Reusable Blocks_, _Edit_, and _Remove from Reusable Blocks_ buttons when necessary.

This is accomplished under the hood by introducing the `canUser()` selector to `core-data` which allows callers to query whether the REST API supports performing a given action on a given resource, e.g. one can query whether the logged in user can create posts by running `wp.data.select( 'core' ).canUser( 'create', 'posts' )`.

The existing `hasUploadPermissions()` selector is changed to use `canUser( 'create', 'media' )` under the hood.

### How it looks

_Add to Reusable Blocks_ button is visible for admins and editors:

admin - add

_Add to Reusable Blocks_ button is hidden for contributors:

contributor - add

_Edit_ button is enabled and _Remove from Reusable Blocks_ is visible for admins and editors:

admin - edit remove

_Edit_ button is disabled and _Remove from Reusable Blocks_ is hidden for authors:

author - edit remove

### How to test

1. Try to create a reusable block as a contributor. The _Add to Reusable Blocks_ button should be disabled.
1. Try to updating another user's reusable block as an author. The _Edit_ button should be disabled.
1. Try to deleting another user's reusable block as an author or contributor. The _Remove from Reusable Blocks_ button should be disabled.

PR #63292, opened 2024-07-09
Core Data: Introduce entity-aware permission selector

### LINKED ISSUE #43751: Core Data selector canUser does not handle entity records
### Description

The `@wordpress/core-data` module provides a selector `canUser( action, resource, id )` that can interrogate whether a user has permission to perform the given CRUD action for the given resource and optionally a specific record.

For example, to check whether the user can update a `page` with the id of `5`, you can perform the following check.

```js
select( 'core' ).canUser( 'update', 'pages', 5 )
```

Unfortuantely, this method only supports resources that are in the `wp/v2` namespace. Additionally, it requires you to know the final REST API path. Typically, however, only an entity `kind` and `name` are known.

There currently exists a `canUserEntityRecord` selector, but it is only a wrapper for `canUser` and does not Post Type entity records. Additionally, it only supports Post Types that have the `wp/v2` namespace which is not a requirement since WP 5.9.

https://github.com/WordPress/gutenberg/blob/1d778aa5e7506390c4d1a89974ec69088026855a/packages/core-data/src/selectors.ts#L996-L1009

I think `canUserEntityRecord` should be adapted to actually perform the permission handling logic utilizing the `baseURL` property of the entity config. Then `canUser` would be deprecated.

1. Register a custom post type with a custom namespace.

```php
register_post_type( 'custom-ns', [
	'public'         => true,
	'show_in_rest'   => true,
	'rest_namespace' => 'my/namespace',
	'supports'       => [ 'editor', 'title', 'custom-fields' ],
] );
```

2. Call the `canUser` selector via the browser console.

```js
wp.data.select('core').canUser('create', 'custom-ns');
```

_No response_

### Environment info

_No response_

Yes

Yes

Mamaduka:
## What?
Resolves #43751.

PR introduces a new entity-aware user permission selector for the `core-data` store. I've also replaced the usage of `canUserEditEntityRecord` in the code base.

## Why?
* `canUser` - only supports resources in the `wp/v2` namespace. Additionally, you must know the final REST API path. Typically, however, only an entity's kind and name are known.
* `canUserEntityRecord` - is limited to only the edit action check. It was a wrapper for `canUser` and only supported the `wp/v2` namespace, which hasn't been required since WP 5.9.

The new selector also opens up the possibility to leverage ["target hints"](https://github.com/WordPress/gutenberg/pull/61644#discussion_r1600465798) and bulk fetch permissions data for DataViews.

## Notes
* I'll make the selector private. It is only public for testing fast testing purposes.
* The name isn't final, and I'm open to suggestions.
* The selector is very similar to `canUser`.
* The state uses the same storage as `canUser` - `cacheKey = isAllowed`.

### Basic Testing
```
wp.data.select( 'core' ).hasPermission( 'update', 'root', 'site' );
wp.data.select( 'core' ).hasPermission( 'delete', 'root', 'site' );
```

### Testing canUserEntityRecord changes
1. Open a post.
2. Add Title and Excerpt blocks.
3. Confirm you can edit them as before.

Same.

### DIFF
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -519,16 +519,7 @@ export const hasPermission =
 export const canUserEditEntityRecord =
 	( kind, name, recordId ) =>
 	async ( { dispatch } ) => {
-		const configs = await dispatch( getOrLoadEntitiesConfig( kind, name ) );
-		const entityConfig = configs.find(
-			( config ) => config.name === name && config.kind === kind
-		);
-		if ( ! entityConfig ) {
-			return;
-		}
-
-		const resource = entityConfig.__unstable_rest_base;
-		await dispatch( canUser( 'update', resource, recordId ) );
+		await dispatch( hasPermission( 'update', kind, name, recordId ) );
 	};
 
 /**
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -425,6 +425,90 @@ export const canUser =
 	};
 
 /**
+ * Checks whether the current user can perform the given action on the entity record.
+ *
+ * @param {string}        action Action to check. One of: 'create', 'read', 'update', 'delete'.
+ * @param {string}        kind   Entity kind.
+ * @param {string}        name   Entity name
+ * @param {number|string} key    Optional record's key.
+ */
+export const hasPermission =
+	( action, kind, name, key ) =>
+	async ( { dispatch, registry } ) => {
+		const configs = await dispatch( getOrLoadEntitiesConfig( kind, name ) );
+		const entityConfig = configs.find(
+			( config ) => config.name === name && config.kind === kind
+		);
+		if ( ! entityConfig ) {
+			return;
+		}
+
+		const { hasStartedResolution } = registry.select( STORE_NAME );
+		const supportedActions = [ 'create', 'read', 'update', 'delete' ];
+
+		if ( ! supportedActions.includes( action ) ) {
+			throw new Error( `'${ action }' is not a valid action.` );
+		}
+
+		// Prevent resolving the same resource twice.
+		for ( const relatedAction of supportedActions ) {
+			if ( relatedAction === action ) {
+				continue;
+			}
+			const isAlreadyResolving = hasStartedResolution( 'hasPermission', [
+				relatedAction,
+				kind,
+				name,
+				key,
+			] );
+			if ( isAlreadyResolving ) {
+				return;
+			}
+		}
+
+		let response;
+		try {
+			response = await apiFetch( {
+				path: entityConfig.baseURL + ( key ? '/' + key : '' ),
+				method: 'OPTIONS',
+				parse: false,
+			} );
+		} catch ( error ) {
+			// Do nothing if our OPTIONS request comes back with an API error (4xx or
+			// 5xx). The previously determined isAllowed value will remain in the store.
+			return;
+		}
+
+		// Optional chaining operator is used here because the API requests don't
+		// return the expected result in the native version. Instead, API requests
+		// only return the result, without including response properties like the headers.
+		const allowHeader = response.headers?.get( 'allow' );
+		const allowedMethods = allowHeader?.allow || allowHeader || '';
+
+		const permissions = {};
+		const methods = {
+			create: 'POST',
+			read: 'GET',
+			update: 'PUT',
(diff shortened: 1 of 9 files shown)


packages/core-data/src/resolvers.js
```diff
@@ -443,8 +443,83 @@ export const canUserEditEntityRecord =
...
+		let response;
+		try {
+			response = await apiFetch( {
+				path: entityConfig.baseURL + ( key ? '/' + key : '' ),
+				method: 'OPTIONS',
+				parse: false,
+			} );
+		} catch ( error ) {
+			// Do nothing if our OPTIONS request comes back with an API error (4xx or
+			// 5xx). The previously determined isAllowed value will remain in the store.
```
TimothyBJacobs, line 478: I don't think this is right. It should indicate that the user has no permissions for the record I believe.

Mamaduka: I found this comment in the original PR: https://github.com/WordPress/gutenberg/pull/12378/files#r243442316.  This is similar to what we do in other resolvers - not touching the store when the request fails.

jsnajdr:
Could this functionality be integrated into `canUser`? The selectors are mostly the same. The syntax could be something like `canUser( 'update', { type: 'entity', kind, name }, 5 )`. Or `canUser( 'update', { type: 'entity', kind, name, id } )`. Then you have two parameters, asking if the user can do operation `param#1` on object `param#2`. Very elegant 🙂 

There's also a request that `canUser` should be able to check the user capabilities. Discussed here: https://github.com/WordPress/gutenberg/issues/61637#issuecomment-2185960467 That's also a very reasonable request.

Mamaduka:
Avoiding yet another public selector deprecation is probably preferable. We can change `resource => resourceOrEntity` and support both signatures:

```js
// new
canUser( 'update', { 'postType', 'wp_block', 5 } );

// legacy
canUser( 'update, 'blocks', 5 );
```

> There's also a request that canUser should be able to check the user capabilities. Discussed here: https://github.com/WordPress/gutenberg/issues/61637#issuecomment-2185960467 That's also a very reasonable request.

I'll try to leave a comment in the discussion, but I don't see it as a blocker for this PR, more like something we can support in the future.

youknowriad:
I would be cautious about objects in selector arguments, there might be impacts on memoization of selectors and things like that. Personally I don't like that pattern much other than for well defined objects like "query".

jsnajdr:
> I would be cautious about objects in selector arguments

Then maybe we can invent some syntax like `entity:postType/wp_block/5`.

@mamaduka says about the user capabilities:
> I'll try to leave a comment in the discussion, but I don't see it as a blocker for this PR

My idea is to keep some space for the possibility that we use `canUser` to check a user capability:
```js
canUser( 'activate_plugins', 'on:this:site' )
```
How would we specify the "object", instead of the `on:this:site` placeholder?

Mamaduka:
> I would be cautious about objects in selector arguments, there might be impacts on memoization of selectors and things like that. Personally I don't like that pattern much other than for well defined objects like "query".

While I generally agree, how likely is it that we would have to memoize a selector that returns `boolean|undefined`?

youknowriad:
> While I generally agree, how likely is it that we would have to memoize a selector that returns boolean|undefined?

Good point, maybe not :). I was just raising it for consideration.

- I think my preference would be a separate selector like this PR proposes but again, it's unclear how that handles the potential future use-case raised by @jsnajdr 
- I don't like inventing formats either

So yeah, if the object argument addresses all the needs without downsides, that's fine by me too.

tyxla (commented):
Thanks for working on this one!

Haven't had a chance to review the code, but using "permission" seems off to me. In the WordPress core, **users** can have a **role** and each role determines their **capabilities**. With that in mind, I prefer **capability** to **permission**. To me, **permission** indicates file/folder permissions on the server, rather than having access to read data or perform certain types of updates. An with that in mind, I'd expect this new API to be consistent with the terms and tech slang that WordPress has already established.

Mamaduka:
@tyxla, it looks like we'll reuse the existing `canUser` selector and add entity support. I'll create a separate PR to update it.

Let's get more feedback for more granular capability controls in #61637. While related, I don't think it's a blocker for entity-aware selectors.

Mamaduka:
The alternative PR is ready for review - https://github.com/WordPress/gutenberg/pull/63322.

Mamaduka:
It was superseded by #63322.

Issue #3144, opened 2017-10-25
Custom statuses are not supported


hlashbrooke:
## Issue Overview
On a site with custom post statuses (on a CPT in this case), the custom statuses are not available anywhere in the Gutenberg UI, so they are are effectively unusable without switching to the classic editor.

## Steps to Reproduce
1. Create custom statuses for a CPT
2. Notice their absence in Gutenberg
3. :sob: 

Using:
- Gutenberg 1.5.1
- WordPress 4.8.2
- Chrome 62.0.3202.62 on Kubuntu 16.04 (although I doubt that's relevant here)

## Expected Behaviour
The custom statuses should be available in Gutenberg.

## Current Behavior
No custom statuses in Gutenberg.

ntwb:
Related: https://core.trac.wordpress.org/ticket/12706 _"Custom post status bugs"_

gziolo:
Related #2457.

danielbachhuber:
Also related https://github.com/WP-API/WP-API/issues/1820

danielbachhuber:
I've [documented this issue](https://github.com/danielbachhuber/gutenberg-migration-guide/blob/master/feature-custom-post-statuses.md) in the migration guide.

danielbachhuber:
Moving this to `WP 5.1 Onwards`. We'll need to decide upon the formal API for custom post statuses before Gutenberg can support this.

skorasaurus:
Would also love to see this implemented; lack of custom statuses in Gutenberg is a dealbreaker for our organization and many other organizations, I imagine.

We are working towards using custom statuses to label and organized content so it can then be improved and edited by multiple parties (the original writers, copy editors, someone else to review it for accessibility, correct any image formatting, SEO, etc) before it is finally published publicly. 

I'm interested to hear how other sites who are using pressbooks, publishpress, Editflow, and Oasis-Workflow, are dealing with this.  

(Written as a CPL employee)
(2022 Update: we ended up having to go around this by the Revisions plugin in WordPress)

greatislander:
@skorasaurus 👋 I'm from Pressbooks. We talked about this a bit in the core REST API chat last week. I'm going to do a bit of digging into the history around custom statuses and propose a path for formal support in core, which my help move this issue forward.

mikelking:
Curious has anyone made any progress on this in the last 6 months?

benlk:
[...]

PR #63293, opened 2024-07-09
Enable `save draft` button for posts with custom post status

### LINKED ISSUE #63161: Save As button not displayed with custom post_status
### Description

The Save As button (button.editor-post-save-draft) is not displayed if the post is saved with a custom non-published status.  As a result, if third party code uses register_post_status() to define a custom post status, posts of that status cannot be updated without publishing. 

This issue does not exist in Gutenberg 17.7 and earlier.

1. In PHP, register a custom post status and insert a sample post with that status: 
```php
function sample_post_for_save_as_issue() {
  register_post_status('testing', ['label' => 'Testing', 'internal' => false, 'protected' => true]); 
  
  if (!is_user_logged_in()) {return;}
    
  $query = new WP_Query(['name' => 'status_test_345234']);
  
  if (!$query->have_posts()) {
    wp_insert_post(['post_name' => 'status_test_345234', 'post_title' => 'Status Test', 'post_status' => 'testing']);
  }
}

add_action('init', 'sample_post_for_save_as_issue');
```

2. Open the sample post "Status Test" in Gutenberg.

3. Key in some post content.

### Expected behavior: 

1. Gutenberg displays the Save Draft button. 

2. Clicking the Save Draft button causes post content changes to be saved, while retaining the custom post status.

3. The Posts screen still lists "Sample Post" and the status count caption "Testing" is still visible next to All / Mine / Published / Drafts.

Note that this is not expected to be a finished use case, just a functional Save As button which third party code can build on.  It works as such with Gutenberg 17.7 and WordPress 6.5.

**Sample Post saved with custom post status (indicated by status count caption though not in table row):**

image

.
**Save Draft button displayed in Gutenberg 17.7:**

image

.
**Save Draft button missing in Gutenberg 18.7:**

image

### Environment info

- WordPress 6.6 rc2
- Gutenberg 18.7.1
- Twenty Twenty-Three theme
- Firefox 127.0.2
- Windows 10 on desktop

Yes

Yes

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/63161

Before the update of the publishing flow, we would render the `save draft` button for a post with custom post type, even if it wasn't really a post with `draft` status. This enabled use cases to make changes in such posts without having to publish first.

This PR adds an extra check when to show this button, to see if the post status is a custom one.

## Testing Instructions
From the issue: 

1.  In PHP, register a custom post status and insert a sample post with that status: 
```php
function sample_post_for_save_as_issue() {
  register_post_status('testing', ['label' => 'Testing', 'internal' => false, 'protected' => true]); 
  
  if (!is_user_logged_in()) {return;}
    
  $query = new WP_Query(['name' => 'status_test_345234']);
  
  if (!$query->have_posts()) {
    wp_insert_post(['post_name' => 'status_test_345234', 'post_title' => 'Status Test', 'post_status' => 'testing']);
  }
}

add_action('init', 'sample_post_for_save_as_issue');
```
2. Open the sample post "Status Test" in Gutenberg.
3. Make and save changes using the `save draft` button.
4. Check existing flows for posts without custom statuses that there is no regression.

### Notes
Custom post statuses are not handled well right now and this is tracked in [this issue](https://github.com/WordPress/gutenberg/issues/3144).

### DIFF
--- a/packages/editor/src/components/post-saved-state/index.js
+++ b/packages/editor/src/components/post-saved-state/index.js
@@ -21,6 +21,7 @@ import { store as preferencesStore } from '@wordpress/preferences';
 /**
  * Internal dependencies
  */
+import { STATUS_OPTIONS } from '../../components/post-status';
 import { store as editorStore } from '../../store';
 
 /**
@@ -104,10 +105,19 @@ export default function PostSavedState( { forceIsDirty } ) {
 		return null;
 	}
 
+	// We shouldn't render the button if the post has not one of the following statuses: pending, draft, auto-draft.
+	// The reason for this is that this button handles the `save as pending` and `save draft` actions.
+	// An exception for this is when the post has a custom status and there should be a way to save changes without
+	// having to publish. This should be handled better in the future when custom statuses have better support.
+	// @see https://github.com/WordPress/gutenberg/issues/3144.
+	const isIneligibleStatus =
+		! [ 'pending', 'draft', 'auto-draft' ].includes( postStatus ) &&
+		STATUS_OPTIONS.map( ( { value } ) => value ).includes( postStatus );
+
 	if (
 		isPublished ||
 		isScheduled ||
-		! [ 'pending', 'draft', 'auto-draft' ].includes( postStatus ) ||
+		isIneligibleStatus ||
 		( postStatusHasChanged &&
 			[ 'pending', 'draft' ].includes( postStatus ) )
 	) {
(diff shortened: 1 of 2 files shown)


Mamaduka (approved):
Thanks, Nik!

The fix works as expected ✅ 

Question (non-blocking): Should we fallback to the `status` slug when its label isn't defined in the `PostStatus` component?

## Screenshot

CleanShot 2024-07-09 at 16 15 59

ntsekouras:
>Question (non-blocking): Should we fallback to the status slug when its label isn't defined in the PostStatus component?

Maybe. I think we should see the custom post statuses issue a bit more and handle this properly. For now and 6.6 let's get this to solve the primary issue for saving.

Mamaduka:
Sounds good!

PR #63296, opened 2024-07-09
Add: Permission checks to avoid 403 errors on non admin roles.

### LINKED ISSUE #60987: Editor: Fix 403 forbidden error for users with low capabilities
Currently, the editor generates two 403 (forbidden) errors for users with low capabilities. Both requests are made to the `/settings` endpoint, which requires `manage_options` capabilities; the capability is only available for administrators by default.

## Sources
* `PostContentInformation` [component](https://github.com/WordPress/gutenberg/blob/314beb747eea61f0941bacca085ab92dda9013da/packages/editor/src/components/post-content-information/index.js#L29).
* `getEditedPostTemplate` [selector](https://github.com/WordPress/gutenberg/blob/73a17b0d3dc9e08aaeb84bf03c9398d61a327a8c/packages/edit-post/src/store/selectors.js#L566).

## Proposed solution
* Re-evaluate each case and apply the fix.
* Add an e2e test to avoid similar regressions in the future.

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60987

Adds permission checks for site settings and global styles retrieval on the editor package to avoid 403 HTTP request errors currently happening when a non-admin role like an editor loads the post editor.

With a user whose role is editor, I created a new page and verified on the browser console that there are no 403 HTTP requests (on trunk there are).

### DIFF
--- a/packages/editor/src/components/global-styles-provider/index.js
+++ b/packages/editor/src/components/global-styles-provider/index.js
@@ -41,9 +41,14 @@ function useGlobalStylesUserConfig() {
 			} = select( coreStore );
 			const _globalStylesId =
 				select( coreStore ).__experimentalGetCurrentGlobalStylesId();
+
+			// Doing canUser( 'read', 'global_styles' ) returns false even for users with the capability.
+			// See: https://github.com/WordPress/gutenberg/issues/63438
+			// So we need to check the user capabilities directly.
 			const userId = getCurrentUser()?.id;
 			const canEditThemeOptions =
 				userId && getUser( userId )?.capabilities?.edit_theme_options;
+
 			const record =
 				_globalStylesId && canEditThemeOptions
 					? getEditedEntityRecord(
@@ -139,6 +144,10 @@ function useGlobalStylesBaseConfig() {
 			getUser,
 			__experimentalGetCurrentThemeBaseGlobalStyles,
 		} = select( coreStore );
+
+		// Doing canUser( 'read', 'global_styles' ) returns false even for users with the capability.
+		// See: https://github.com/WordPress/gutenberg/issues/63438
+		// So we need to check the user capabilities directly.
 		const userId = getCurrentUser()?.id;
 		const canEditThemeOptions =
 			userId && getUser( userId )?.capabilities?.edit_theme_options;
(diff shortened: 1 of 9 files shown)


packages/editor/src/components/global-styles-provider/index.js
```diff
@@ -33,17 +33,25 @@ export function mergeBaseAndUserConfigs( base, user ) {
...
 			const _globalStylesId =
 				select( coreStore ).__experimentalGetCurrentGlobalStylesId();
-			const record = _globalStylesId
-				? getEditedEntityRecord(
-						'root',
-						'globalStyles',
-						_globalStylesId
-				  )
-				: undefined;
+			const userId = getCurrentUser()?.id;
```
jorgefilipecosta, line 48: Ideally, we would do canUser( 'read', 'global_styles' ) but this is not working even for admins so it is probably another bug we should check.

Mamaduka: Let's add inline comments here. It's probably worth opening an issue for REST API changes.

packages/editor/src/components/global-styles-provider/index.js
```diff
@@ -126,9 +134,19 @@ function useGlobalStylesUserConfig() {
...
-			coreStore
-		).__experimentalGetCurrentThemeBaseGlobalStyles();
+		const {
+			getCurrentUser,
+			getUser,
+			__experimentalGetCurrentThemeBaseGlobalStyles,
+		} = select( coreStore );
+		const userId = getCurrentUser()?.id;
+		const canEditThemeOptions =
+			userId && getUser( userId )?.capabilities?.edit_theme_options;
```
Mamaduka, line 153: Same here: Let's add a comment regarding `getUser( userId )?.capabilities?.edit_theme_options` usage.

packages/editor/src/components/global-styles-provider/index.js
jorgefilipecosta: Hi @Mamaduka, thank you for the review. I added comments and created an issue as suggested.

Mamaduka:
Thanks, @jorgefilipecosta!

It looks like we missed a spot in `useAllowSwitchingTemplates`.

Additionally, it might be a good idea to start using the new syntax for `canUser`, which supports all entities (#63322). I'm working on a general update in #63415.

jorgefilipecosta:
Hi @Mamaduka,

> Thanks, @jorgefilipecosta!
> 
> It looks like we missed a spot in `useAllowSwitchingTemplates`.

Nice catch the check was added there.
 
> Additionally, it might be a good idea to start using the new syntax for `canUser`, which supports all entities (#63322). I'm working on a general update in #63415.

I updated to the new syntax.

Mamaduka (approved):
Thanks, @jorgefilipecosta!

Tested with the author role and couldn't spot any 403 errors.

PR #63297, opened 2024-07-09
Tabs: fix "With tab icons" Storybook example

### LINKED ISSUE #63271: Tabs: "With Tab Icons" story has duplicated focus styles

The `Tabs` ["With Tab Icons And Tooltips" story](https://wordpress.github.io/gutenberg/?path=/story/components-experimental-tabs--with-tab-icons-and-tooltips) has duplicated focus styles.

Icon tabs with duplicate focus styles

Ideally, we switch the documented implementation to something that doesn't result in double focus indicators. We shouldn't add overrides to avoid this double focus, if possible.

ciampo:
## What?

Fixes #63271

Tweaks the "With tab icons" Storybook example for the `Tabs` component so that tab styles are correct

## Why?

Previously, the example was using the `Button` component, which would being in its own styles, overriding the `Tabs.Tab` styles.

## How?

By switching from `Button` to using the individual `Tooltip` and `Icon` components, as suggested by @mirka in https://github.com/WordPress/gutenberg/issues/63271#issuecomment-2217429525

## Testing Instructions

- Load the "With tab icons" Storybook example for the `Tabs` component
- Play around with the component, using the keyboard to select different tabs
- Make sure that focus styles look correct, that tab selection works as expected, and the tooltip shows as expected while hovering or focussing a tab

## Screenshots or screencast 

|---|---|
| Screenshot 2024-07-09 at 14 43 08 | Screenshot 2024-07-09 at 14 41 17 |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -23,6 +23,7 @@
 -   Fix inaccessibly disabled `Button`s ([#62306](https://github.com/WordPress/gutenberg/pull/62306)).
 -   `TimePicker`: Fix time zone overflow ([#63209](https://github.com/WordPress/gutenberg/pull/63209)).
 -   `Tabs`: Fix text-align when text wraps in vertical mode ([#63272](https://github.com/WordPress/gutenberg/pull/63272)).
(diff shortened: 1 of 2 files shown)


mirka (approved):
Works as expected, thanks!

PR #63298, opened 2024-07-09
Inserter: fix 'append-after' keyboard a11y in both modes

### LINKED ISSUE #61391: Block list appender no longer reachable with the keyboard from the settings panel
### Description

Reported by @alexstine [on Slack](https://wordpress.slack.com/archives/C02RP4X03/p1714750136659839).

It appears an important keyboard interaction affordance changed from WordPress 6.5 to Gutenberg latest trunk 18.3.0-rc.1, I'm assuming the change is not intentional thus marking this issue as regression.

Previously, on WordPress 6.5, with the post editor in 'Select' mode (also known as 'navigation mode'), pressing Shift + Tab from the Settings panel brought users to the 'block list appender' at the bottom of the blocks list, which provided the ability to reach:
- the 'Add block' plus icon button 
- the 'Add default block'

This used to be one of the easiest ways for keyboard users and screen reader users to reach the bottom of the post content and add blocks there.

On latest Gutenberg trunk this doesn't work any longer. Pressing Shift + Tab from the Settings panel brings users directly to the last block, skipping entirely the 'Add block' and 'Add default block'.

These kind of changes that impact keyboard interaction are extremely impactful for keyboard users. Suddenly, a known affordance and user flow to add block disappears, making adding blocks a way longer and tedious process.

When fixed, it would be nice to add e2e tests to prevent such unexpected changes in the future.

When developing, _manually testing the editor with the keyboard would be gretly appreciated_ to prevent such keyboard usability regressions.

See attached GIFs to better illustrate.

Expected keyboard interaction on WordPress 6.5:

Changed interaction on Gutenberg trunk:

gb trunk

- Test on WordPress 6.5.
- Edit a post that contains at least one block.
- Switch the editor to 'Select mode' from the top toolbar > Tools dropdown menu.
- Press Tab as much as necessary to reach the Settings panel (open it if necessary).
- From the Settings Panel, press Shift + Tab to navigate backwards.
- Observe it is possible to reach the block list appender at the bottom of the post, so that focus goes to the 'Add block' plus icon button and then to the 'Add default block'.
- Test on latest Gutenberg trunk.
- Repeat the steps above.
- Observe the 'Add block' plus icon button and the 'Add default block' are not reachable. Observe they're not even rendered when using this keyboard flow.

_No response_

### Environment info

_No response_

Yes

Yes

ellatrix:
## What?

Fixes #61391.

Adds a tabbable inserter:

* In edit mode, after the selected block.
* In navigation/select mode, at the end of the canvas.

## Why?

To make it easier to add a block after the selected block.

## How?

The approach here is to leverage the "in-between" inserter. Instead of being accessible with pointer devices only, make it also keyboard accessible.

## Testing Instructions

* In edit mode, select a block, then press tab. Focus should mode the inserter. Press tab again and it should move to the sidebar. Reverse tab should also work correctly.
* In navigation/select mode (press Esc), Tab to the last block. The next tab should focus the inserter, the next the sidebar. Reverse tab should also work correctly.

See above, this is a keyboard only fix.

## Screenshots or screencast 

insert-after

### DIFF
--- a/packages/block-editor/src/components/block-tools/index.js
+++ b/packages/block-editor/src/components/block-tools/index.js
@@ -202,12 +202,6 @@ export default function BlockTools( {
 		// eslint-disable-next-line jsx-a11y/no-static-element-interactions
 		<div { ...props } onKeyDown={ onKeyDown }>
 			<InsertionPointOpenRef.Provider value={ useRef( false ) }>
-				{ ! isTyping && (
-					<InsertionPoint
-						__unstableContentRef={ __unstableContentRef }
-					/>
-				) }
-
 				{ showEmptyBlockSideInserter && (
 					<EmptyBlockInserter
 						__unstableContentRef={ __unstableContentRef }
@@ -251,6 +245,11 @@ export default function BlockTools( {
 					name="__unstable-block-tools-after"
 					ref={ blockToolbarAfterRef }
 				/>
+				{ ! isTyping && (
+					<InsertionPoint
+						__unstableContentRef={ __unstableContentRef }
+					/>
+				) }
 				{ window.__experimentalEnableZoomedOutPatternsTab &&
 					isZoomOutMode && (
 						<ZoomOutModeInserters
--- a/packages/block-editor/src/components/block-popover/style.scss
+++ b/packages/block-editor/src/components/block-popover/style.scss
@@ -47,10 +47,12 @@
 
 .block-editor-block-list__insertion-point {
 	opacity: 0;
+	pointer-events: none;
 
 	&:focus-within,
 	.is-visible {
 		opacity: 1;
+		pointer-events: all;
 	}
 }
 
--- a/packages/block-editor/src/components/block-popover/style.scss
+++ b/packages/block-editor/src/components/block-popover/style.scss
@@ -45,6 +45,14 @@
 	}
 }
 
+.block-editor-block-list__insertion-point {
+	opacity: 0;
+
+	&:focus-within,
+	.is-visible {
+		opacity: 1;
+	}
+}
 
 .components-popover.block-editor-block-popover__drop-zone {
 	// Disable pointer events for dragging and dropping.
--- a/packages/block-editor/src/components/block-tools/insertion-point.js
+++ b/packages/block-editor/src/components/block-tools/insertion-point.js
@@ -26,6 +26,7 @@ function InbetweenInsertionPointPopover( {
 	__unstableContentRef,
 	operation = 'insert',
 	nearestSide = 'right',
+	isVisible,
 } ) {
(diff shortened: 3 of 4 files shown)


packages/block-editor/src/components/block-tools/index.js
```diff
@@ -251,6 +245,11 @@ export default function BlockTools( {
 					name="__unstable-block-tools-after"
 					ref={ blockToolbarAfterRef }
 				/>
+				{ ! isTyping && (
+					<InsertionPoint
+						__unstableContentRef={ __unstableContentRef }
+					/>
+				) }
```
ellatrix, line 252: Note that the order previously didn't matter here, it was available only to pointer devices.
Also note that the zoom-out inserters are actually right below this one now.

alexstine (commented):
@ellatrix Thanks for the PR. Accessibility is much better here and glad that this inserter is now reachable with the keyboard. I gave this a good test. Lots of E2E failures, hopefully unrelated to this PR.

ellatrix:
@alexstine Thanks for testing. I'll fix the e2e tests, it probably needs some updated snapshots etc.

afercia:
@ellatrix thanks for this PR. I haven't looked at the code, just quickly tested the interaction. The inserter is added and reachable with the keyboard. I noticed something to address though. 

When I move the last block up, the new inserter seems 'attached' to it. When tabbing, it does appear after what previously was the last block, which is now in a new position though. Animated GIF to illustrate:


moving

ellatrix:
> When I move the last block up, the new inserter seems 'attached' to it. When tabbing, it does appear after what previously was the last block, which is now in a new position though. Animated GIF to illustrate:

Could you elaborate on what's unexpected? I checked the GIF a few times, but I didn't see anything strange. The insertion point moves with the selected in edit mode, yes. It's similar to the block toolbar, which is also visible for the currently selected block.

afercia:
> The insertion point moves with the selected in edit mode, yes

To me, this is unexpected. The whole point of this inserter is to provide keyboard users with the ability to _add a block at the end of the canvas_. If the inserter moves, they can't add a block at the end easily.

ellatrix:
> To me, this is unexpected. The whole point of this inserter is to provide keyboard users with the ability to add a block at the end of the canvas. If the inserter moves, they can't add a block at the end easily.

They can add one at the end of the canvas through navigation mode, as it was the case before.

* Worth noting that edit mode currently has _no_ inserter at all similar to this one.
* Also worth noting that "end of canvas" doesn't always make sense. In the site editor or a locked post template, it's quite rare to append a block at the very end of the canvas.

One idea that I'd like to try in a follow up PR is allow moving the inserter with arrow keys. So once you tab into it, you could navigate it up and down, and maybe even left to move it one block up. Maybe we can also make the home/end keys work or something to move it down the inner blocks list.

PR #63299, opened 2024-07-09
Dataviews List: Update item layout

jameskoster:
Alternative to https://github.com/WordPress/gutenberg/pull/63043.

## What?
This PR addresses three issues in the List layout appearance / behavior.

### Issue 1
When all fields are toggled off, the alignment is broken:

### Issue 2

Awkward reflow occurs when hovering/selecting a row that contains a field with a very long value, or just many fields:

https://github.com/WordPress/gutenberg/assets/846565/da6f4cf0-0aab-4044-b237-32d2a19b3b2a

### Issue 3

Clicking around the action buttons will not select the row, despite the hover style suggesting it would:

https://github.com/WordPress/gutenberg/assets/846565/05e3b283-d858-4d1a-9272-8a7b44910e76

## How?
* Action buttons are absolutely positioned to avoid reflow when they appear/disappear. This also fixes issue 3.
* When the fields container is empty different styling is applied.
* Updated layout metrics to align the action buttons with the title.

Here's the result:

https://github.com/WordPress/gutenberg/assets/846565/7c9a1c08-f0cf-4310-a3fe-b02bfa9857ee

## Testing Instructions
1. Navigate to a dataview like pages and switch to list layout.
2. Toggle fields on / off and ensure alignment is maintained, and that the three issues are fixed.
3. Ensure that keyboard navigation works as expected.

### DIFF
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -448,7 +448,7 @@
 			.dataviews-view-list__item-actions {
 				background: #f8f8f8;
 				padding-left: $grid-unit-10;
-				padding-right: $grid-unit-30;
+				margin-right: $grid-unit-30;
 				box-shadow: -12px 0 8px 0 #f8f8f8;
 
 				.components-button {
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -421,10 +421,24 @@
 		}
 
 		.dataviews-view-list__item-actions {
+			position: absolute;
+			top: $grid-unit-20;
+			right: 0;
+
+
+			> div {
+				height: $button-size-small;
+			}
+
 			.components-button {
 				opacity: 0;
-				position: fixed;
-				right: 0;
+			}
+		}
+
+		&:has(.dataviews-view-list__fields:empty) {
+			.dataviews-view-list__item-actions {
+				top: 50%;
+				transform: translateY(-50%);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


packages/dataviews/src/style.scss
```diff
@@ -421,27 +421,47 @@
...
 
 		&.is-selected,
 		&.is-hovered,
 		&:focus-within {
 			.dataviews-view-list__item-actions {
-				padding-right: $grid-unit-30;
+				background: #f8f8f8;
+				padding-left: $grid-unit-10;
+				margin-right: $grid-unit-30;
+				box-shadow: -12px 0 8px 0 #f8f8f8;
```
jasmussen, line 452: Can we substitute a few variables here?

packages/dataviews/src/style.scss
```diff
@@ -421,27 +421,47 @@
...
 					position: static;
 				}
 			}
+		}
 
-			.dataviews-view-list__item {
-				padding-right: 0;
+		&.is-selected {
+			.dataviews-view-list__item-actions {
+				background-color: rgb(247 248 255);
```
jasmussen, line 463: Same here.

jasmussen (approved):
Works well.

packages/dataviews/src/style.scss
jameskoster: We have an outstanding todo item for this gray. It will mean expanding the existing scale a bit, so needs a dedicated PR. 

I'm not sure variables really necessary for shadows.

packages/dataviews/src/style.scss
jameskoster: This is a 'solid' version of `rgba(var(--wp-admin-theme-color--rgb), 0.04)`, which should probably be a variable too given it's increasing use. I'd lean towards tackling this together, separately.

jameskoster:
I'm going to merge this since it fixes some troublesome bugs. We can circle back to explore the variables separately.

PR #34839, opened 2021-09-15
Fix navigation editor undo button being active when editor loads

talldan:
## Description
Fixes #34802

The navigation editor is an entity driven editor, like the post or widget editor. But the entity that drives it is never saved, it's only a convenient in-memory entity. It also didn't have any configuration (edit: it actually presently uses the `postTypes` entity, which is a bit of a hack), which caused the undo feature in the nav editor to work differently to other editors (more details in point 2 here - https://github.com/WordPress/gutenberg/issues/34802#issuecomment-919839931).

In this PR I've made it so that the entity config is created at initialisation of the editor. As it's only a convenience entity, I don't think it makes sense to add it to the `entities.js` file. This seems like a decent alternative.

It also changes the kind and type so that the navigation post is more like a `postType` entity.

To avoid this entity making REST API calls I've added an `__experimentalNoFetch` property to it.

## How has this been tested?
1. Create a menu with menu items and save it
2. Reload the editor

Expected: The `undo` button should be disabled.

## Types of changes

Bug fix (non-breaking change which fixes an issue) 

## Checklist:
- [ ] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #63303, opened 2024-07-09
Core Data: Remove entity config '__experimentalNoFetch' flag checks

Mamaduka:
## What?
PR removes `entityConfig.__experimentalNoFetch` checks from resolvers and selectors.

## Why?
It was introduced initially to handle a particular case in the now-removed Navigation editor (#34839). I don't see a reason for core entities to use this flag.

## Testing Instructions
All CI checks should be green.

### DIFF
--- a/packages/core-data/src/actions.js
+++ b/packages/core-data/src/actions.js
@@ -292,7 +292,7 @@ export const deleteEntityRecord =
 		);
 		let error;
 		let deletedRecord = false;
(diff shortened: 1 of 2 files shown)


youknowriad (approved):
Looking through wpdirectory I think we're fine here but it's always painful to go through the results of plugins that "bundle" packages.

Mamaduka:
I also searched for wpdirectory and GitHub repos — no usage besides Gutenberg forks.

PR #63311, opened 2024-07-09
Build: Include Core blocks' `render` and `variations` files

ockham:
## What?
If a Core block's `block.json` contains a `render` and/or `variations` field that points to a PHP file name, copy those files to the build directory.

## Why?
It's an attempt at fixing https://github.com/WordPress/gutenberg/issues/63077.

## How?
By reusing logic from the `@wordpress/scripts` package.

## Testing Instructions

- On this branch, apply the patch below. (It will move the Template Part block's `render` method from `index.php` to a newly created `render.php`, and add a `render` field to `block.json` that points to that `render.php` file.)
- Run `npm run start`.
- Verify that `packages/block-library/src/template-part/render.php` has been copied to `build/block-library/blocks/template-part/render.php`.
- Make sure you're using a block theme such as TT4.
- In your web browser, verify that all instances of the Template Part block -- e.g. the header and footer -- render correctly.

<details>
<summary>
Patch
</summary>

```diff
diff --git a/packages/block-library/src/template-part/block.json b/packages/block-library/src/template-part/block.json
index 9710bdeee2e..7c3e28f52c0 100644
--- a/packages/block-library/src/template-part/block.json
+++ b/packages/block-library/src/template-part/block.json
@@ -29,5 +29,6 @@
 			"clientNavigation": true
 		}
 	},
-	"editorStyle": "wp-block-template-part-editor"
+	"editorStyle": "wp-block-template-part-editor",
+	"render": "file:./render.php"
 }
diff --git a/packages/block-library/src/template-part/index.php b/packages/block-library/src/template-part/index.php
index be867c4ced1..a9aef23f3da 100644
--- a/packages/block-library/src/template-part/index.php
+++ b/packages/block-library/src/template-part/index.php
@@ -5,176 +5,6 @@
  * @package WordPress
  */
 
-/**
- * Renders the `core/template-part` block on the server.
- *
- * @since 5.9.0
- *
- * @global WP_Embed $wp_embed WordPress Embed object.
- *
- * @param array $attributes The block attributes.
- *
- * @return string The render.
- */
-function render_block_core_template_part( $attributes ) {
-	static $seen_ids = array();
-
-	$template_part_id = null;
-	$content          = null;
-	$area             = WP_TEMPLATE_PART_AREA_UNCATEGORIZED;
-	$theme            = isset( $attributes['theme'] ) ? $attributes['theme'] : get_stylesheet();
-
-	if ( isset( $attributes['slug'] ) && get_stylesheet() === $theme ) {
-		$template_part_id    = $theme . '//' . $attributes['slug'];
-		$template_part_query = new WP_Query(
-			array(
-				'post_type'           => 'wp_template_part',
-				'post_status'         => 'publish',
-				'post_name__in'       => array( $attributes['slug'] ),
-				'tax_query'           => array(
-					array(
-						'taxonomy' => 'wp_theme',
-						'field'    => 'name',
-						'terms'    => $theme,
-					),
-				),
-				'posts_per_page'      => 1,
-				'no_found_rows'       => true,
-				'lazy_load_term_meta' => false, // Do not lazy load term meta, as template parts only have one term.
-			)
-		);
-		$template_part_post  = $template_part_query->have_posts() ? $template_part_query->next_post() : null;
-		if ( $template_part_post ) {
-			// A published post might already exist if this template part was customized elsewhere
-			// or if it's part of a customized template.
-			$block_template = _build_block_template_result_from_post( $template_part_post );
-			$content        = $block_template->content;
-			if ( isset( $block_template->area ) ) {
-				$area = $block_template->area;
-			}
-			/**
-			 * Fires when a block template part is loaded from a template post stored in the database.
-			 *
-			 * @since 5.9.0
-			 *
-			 * @param string  $template_part_id   The requested template part namespaced to the theme.
-			 * @param array   $attributes         The block attributes.
-			 * @param WP_Post $template_part_post The template part post object.
-			 * @param string  $content            The template part content.
-			 */
-			do_action( 'render_block_core_template_part_post', $template_part_id, $attributes, $template_part_post, $content );
-		} else {
-			$template_part_file_path = '';
-			// Else, if the template part was provided by the active theme,
-			// render the corresponding file content.
-			if ( 0 === validate_file( $attributes['slug'] ) ) {
-				$block_template = get_block_file_template( $template_part_id, 'wp_template_part' );
-
-				$content = $block_template->content;
-				if ( isset( $block_template->area ) ) {
-					$area = $block_template->area;
-				}
-
-				// Needed for the `render_block_core_template_part_file` and `render_block_core_template_part_none` actions below.
-				$block_template_file = _get_block_template_file( 'wp_template_part', $attributes['slug'] );
-				if ( $block_template_file ) {
-					$template_part_file_path = $block_template_file['path'];
-				}
-			}
-
-			if ( '' !== $content && null !== $content ) {
-				/**
-				 * Fires when a block template part is loaded from a template part in the theme.
-				 *
-				 * @since 5.9.0
-				 *
-				 * @param string $template_part_id        The requested template part namespaced to the theme.
-				 * @param array  $attributes              The block attributes.
-				 * @param string $template_part_file_path Absolute path to the template path.
-				 * @param string $content                 The template part content.
-				 */
-				do_action( 'render_block_core_template_part_file', $template_part_id, $attributes, $template_part_file_path, $content );
-			} else {
-				/**
-				 * Fires when a requested block template part does not exist in the database nor in the theme.
-				 *
-				 * @since 5.9.0
-				 *
-				 * @param string $template_part_id        The requested template part namespaced to the theme.
-				 * @param array  $attributes              The block attributes.
-				 * @param string $template_part_file_path Absolute path to the not found template path.
-				 */
-				do_action( 'render_block_core_template_part_none', $template_part_id, $attributes, $template_part_file_path );
-			}
-		}
-	}
-
-	// WP_DEBUG_DISPLAY must only be honored when WP_DEBUG. This precedent
-	// is set in `wp_debug_mode()`.
-	$is_debug = WP_DEBUG && WP_DEBUG_DISPLAY;
-
-	if ( is_null( $content ) ) {
-		if ( $is_debug && isset( $attributes['slug'] ) ) {
-			return sprintf(
-				/* translators: %s: Template part slug. */
-				__( 'Template part has been deleted or is unavailable: %s' ),
-				$attributes['slug']
-			);
-		}
-
-		return '';
-	}
-
-	if ( isset( $seen_ids[ $template_part_id ] ) ) {
-		return $is_debug ?
-			// translators: Visible only in the front end, this warning takes the place of a faulty block.
-			__( '[block rendering halted]' ) :
-			'';
-	}
-
-	// Look up area definition.
-	$area_definition = null;
-	$defined_areas   = get_allowed_block_template_part_areas();
-	foreach ( $defined_areas as $defined_area ) {
-		if ( $defined_area['area'] === $area ) {
-			$area_definition = $defined_area;
-			break;
-		}
-	}
-
-	// If $area is not allowed, set it back to the uncategorized default.
-	if ( ! $area_definition ) {
-		$area = WP_TEMPLATE_PART_AREA_UNCATEGORIZED;
-	}
-
-	// Run through the actions that are typically taken on the_content.
-	$content                       = shortcode_unautop( $content );
-	$content                       = do_shortcode( $content );
-	$seen_ids[ $template_part_id ] = true;
-	$content                       = do_blocks( $content );
-	unset( $seen_ids[ $template_part_id ] );
-	$content = wptexturize( $content );
-	$content = convert_smilies( $content );
-	$content = wp_filter_content_tags( $content, "template_part_{$area}" );
-
-	// Handle embeds for block template parts.
-	global $wp_embed;
-	$content = $wp_embed->autoembed( $content );
-
-	if ( empty( $attributes['tagName'] ) || tag_escape( $attributes['tagName'] ) !== $attributes['tagName'] ) {
-		$area_tag = 'div';
-		if ( $area_definition && isset( $area_definition['area_tag'] ) ) {
-			$area_tag = $area_definition['area_tag'];
-		}
-		$html_tag = $area_tag;
-	} else {
-		$html_tag = esc_attr( $attributes['tagName'] );
-	}
-	$wrapper_attributes = get_block_wrapper_attributes();
-
-	return "<$html_tag $wrapper_attributes>" . str_replace( ']]>', ']]&gt;', $content ) . "</$html_tag>";
-}
-
 /**
  * Returns an array of area variation objects for the template part block.
  *
@@ -293,7 +123,7 @@ function register_block_core_template_part() {
 	register_block_type_from_metadata(
 		__DIR__ . '/template-part',
 		array(
-			'render_callback'    => 'render_block_core_template_part',
+			//'render_callback'    => 'render_block_core_template_part',
 			'variation_callback' => 'build_template_part_block_variations',
 		)
 	);
diff --git a/packages/block-library/src/template-part/render.php b/packages/block-library/src/template-part/render.php
new file mode 100644
index 00000000000..f22a06d954c
--- /dev/null
+++ b/packages/block-library/src/template-part/render.php
@@ -0,0 +1,171 @@
+<?php
+
+/**
+ * Renders the `core/template-part` block on the server.
+ *
+ * @since 5.9.0
+ *
+ * @global WP_Embed $wp_embed WordPress Embed object.
+ *
+ * @param array $attributes The block attributes.
+ *
+ * @return string The render.
+ */
+//function render_block_core_template_part( $attributes ) {
+	static $seen_ids = array();
+
+	$template_part_id = null;
+	$content          = null;
+	$area             = WP_TEMPLATE_PART_AREA_UNCATEGORIZED;
+	$theme            = isset( $attributes['theme'] ) ? $attributes['theme'] : get_stylesheet();
+
+	if ( isset( $attributes['slug'] ) && get_stylesheet() === $theme ) {
+		$template_part_id    = $theme . '//' . $attributes['slug'];
+		$template_part_query = new WP_Query(
+			array(
+				'post_type'           => 'wp_template_part',
+				'post_status'         => 'publish',
+				'post_name__in'       => array( $attributes['slug'] ),
+				'tax_query'           => array(
+					array(
+						'taxonomy' => 'wp_theme',
+						'field'    => 'name',
+						'terms'    => $theme,
+					),
+				),
+				'posts_per_page'      => 1,
+				'no_found_rows'       => true,
+				'lazy_load_term_meta' => false, // Do not lazy load term meta, as template parts only have one term.
+			)
+		);
+		$template_part_post  = $template_part_query->have_posts() ? $template_part_query->next_post() : null;
+		if ( $template_part_post ) {
+			// A published post might already exist if this template part was customized elsewhere
+			// or if it's part of a customized template.
+			$block_template = _build_block_template_result_from_post( $template_part_post );
+			$content        = $block_template->content;
+			if ( isset( $block_template->area ) ) {
+				$area = $block_template->area;
+			}
+			/**
+			 * Fires when a block template part is loaded from a template post stored in the database.
+			 *
+			 * @since 5.9.0
+			 *
+			 * @param string  $template_part_id   The requested template part namespaced to the theme.
+			 * @param array   $attributes         The block attributes.
+			 * @param WP_Post $template_part_post The template part post object.
+			 * @param string  $content            The template part content.
+			 */
+			do_action( 'render_block_core_template_part_post', $template_part_id, $attributes, $template_part_post, $content );
+		} else {
+			$template_part_file_path = '';
+			// Else, if the template part was provided by the active theme,
+			// render the corresponding file content.
+			if ( 0 === validate_file( $attributes['slug'] ) ) {
+				$block_template = get_block_file_template( $template_part_id, 'wp_template_part' );
+
+				$content = $block_template->content;
+				if ( isset( $block_template->area ) ) {
+					$area = $block_template->area;
+				}
+
+				// Needed for the `render_block_core_template_part_file` and `render_block_core_template_part_none` actions below.
+				$block_template_file = _get_block_template_file( 'wp_template_part', $attributes['slug'] );
+				if ( $block_template_file ) {
+					$template_part_file_path = $block_template_file['path'];
+				}
+			}
+
+			if ( '' !== $content && null !== $content ) {
+				/**
+				 * Fires when a block template part is loaded from a template part in the theme.
+				 *
+				 * @since 5.9.0
+				 *
+				 * @param string $template_part_id        The requested template part namespaced to the theme.
+				 * @param array  $attributes              The block attributes.
+				 * @param string $template_part_file_path Absolute path to the template path.
+				 * @param string $content                 The template part content.
+				 */
+				do_action( 'render_block_core_template_part_file', $template_part_id, $attributes, $template_part_file_path, $content );
+			} else {
+				/**
+				 * Fires when a requested block template part does not exist in the database nor in the theme.
+				 *
+				 * @since 5.9.0
+				 *
+				 * @param string $template_part_id        The requested template part namespaced to the theme.
+				 * @param array  $attributes              The block attributes.
+				 * @param string $template_part_file_path Absolute path to the not found template path.
+				 */
+				do_action( 'render_block_core_template_part_none', $template_part_id, $attributes, $template_part_file_path );
+			}
+		}
+	}
+
+	// WP_DEBUG_DISPLAY must only be honored when WP_DEBUG. This precedent
+	// is set in `wp_debug_mode()`.
+	$is_debug = WP_DEBUG && WP_DEBUG_DISPLAY;
+
+	if ( is_null( $content ) ) {
+		if ( $is_debug && isset( $attributes['slug'] ) ) {
+			return sprintf(
+				/* translators: %s: Template part slug. */
+				__( 'Template part has been deleted or is unavailable: %s' ),
+				$attributes['slug']
+			);
+		}
+
+		return '';
+	}
+
+	if ( isset( $seen_ids[ $template_part_id ] ) ) {
+		return $is_debug ?
+			// translators: Visible only in the front end, this warning takes the place of a faulty block.
+			__( '[block rendering halted]' ) :
+			'';
+	}
+
+	// Look up area definition.
+	$area_definition = null;
+	$defined_areas   = get_allowed_block_template_part_areas();
+	foreach ( $defined_areas as $defined_area ) {
+		if ( $defined_area['area'] === $area ) {
+			$area_definition = $defined_area;
+			break;
+		}
+	}
+
+	// If $area is not allowed, set it back to the uncategorized default.
+	if ( ! $area_definition ) {
+		$area = WP_TEMPLATE_PART_AREA_UNCATEGORIZED;
+	}
+
+	// Run through the actions that are typically taken on the_content.
+	$content                       = shortcode_unautop( $content );
+	$content                       = do_shortcode( $content );
+	$seen_ids[ $template_part_id ] = true;
+	$content                       = do_blocks( $content );
+	unset( $seen_ids[ $template_part_id ] );
+	$content = wptexturize( $content );
+	$content = convert_smilies( $content );
+	$content = wp_filter_content_tags( $content, "template_part_{$area}" );
+
+	// Handle embeds for block template parts.
+	global $wp_embed;
+	$content = $wp_embed->autoembed( $content );
+
+	if ( empty( $attributes['tagName'] ) || tag_escape( $attributes['tagName'] ) !== $attributes['tagName'] ) {
+		$area_tag = 'div';
+		if ( $area_definition && isset( $area_definition['area_tag'] ) ) {
+			$area_tag = $area_definition['area_tag'];
+		}
+		$html_tag = $area_tag;
+	} else {
+		$html_tag = esc_attr( $attributes['tagName'] );
+	}
+	$wrapper_attributes = get_block_wrapper_attributes();
+
+	echo "<$html_tag $wrapper_attributes>" . str_replace( ']]>', ']]&gt;', $content ) . "</$html_tag>";
+//}
```
</details>

### DIFF
--- a/packages/scripts/CHANGELOG.md
+++ b/packages/scripts/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ### Enhancements
 
-- Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
+-   Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
+-   Add automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
 
 ### Bug Fixes
 
--- a/packages/scripts/CHANGELOG.md
+++ b/packages/scripts/CHANGELOG.md
@@ -10,8 +10,7 @@
 
 ### Enhancements
 
--   Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
--   Add automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
+- Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
 
 ### Bug Fixes
 
--- a/packages/scripts/CHANGELOG.md
+++ b/packages/scripts/CHANGELOG.md
@@ -11,7 +11,7 @@
 ### Enhancements
 
 -   Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
--   Adds automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
+-   Add automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
 
 ### Bug Fixes
 
--- a/packages/scripts/CHANGELOG.md
+++ b/packages/scripts/CHANGELOG.md
@@ -11,7 +11,7 @@
 ### Enhancements
 
 -   Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
--   Add automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
+-   Adds automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
 
 ### Bug Fixes
 
--- a/packages/scripts/utils/php-file-paths-plugin.js
+++ b/packages/scripts/utils/php-file-paths-plugin.js
@@ -6,7 +6,7 @@ const { validate } = require( 'schema-utils' );
 /**
  * Internal dependencies
  */
-const { getPhpFilePaths } = require( '../utils' );
+const { getPhpFilePaths } = require( './config' );
 
 const phpFilePathsPluginSchema = {
 	type: 'object',
--- a/packages/scripts/utils/php-file-paths-plugin.js
+++ b/packages/scripts/utils/php-file-paths-plugin.js
@@ -11,6 +11,9 @@ const { getPhpFilePaths } = require( '../utils' );
 const phpFilePathsPluginSchema = {
 	type: 'object',
 	properties: {
+		context: {
+			type: 'string',
+		},
 		props: {
 			type: 'array',
 			items: {
@@ -46,7 +49,10 @@ class PhpFilePathsPlugin {
 		const pluginName = this.constructor.name;
 
 		compiler.hooks.thisCompilation.tap( pluginName, () => {
-			this.constructor.paths = getPhpFilePaths( this.options.props );
+			this.constructor.paths = getPhpFilePaths(
+				this.options.context,
+				this.options.props
+			);
 		} );
 	}
 }
--- a/packages/scripts/utils/php-file-paths-plugin.js
+++ b/packages/scripts/utils/php-file-paths-plugin.js
@@ -1,0 +1,54 @@
+/**
+ * External dependencies
+ */
+const { validate } = require( 'schema-utils' );
+
+/**
+ * Internal dependencies
+ */
+const { getPhpFilePaths } = require( '../utils' );
+
+const phpFilePathsPluginSchema = {
+	type: 'object',
+	properties: {
+		props: {
+			type: 'array',
+			items: {
+				type: 'string',
+			},
+		},
+	},
+};
+
+/**
+ * The plugin recomputes PHP file paths once on each compilation. It is necessary to avoid repeating processing
+ * when filtering every discovered PHP file in the source folder. This is the most performant way to ensure that
+ * changes in `block.json` files are picked up in watch mode.
+ */
+class PhpFilePathsPlugin {
+	/**
+	 * PHP file paths from `render` and `variations` props found in `block.json` files.
+	 *
+	 * @type {string[]}
+	 */
+	static paths;
+
+	constructor( options = {} ) {
+		validate( phpFilePathsPluginSchema, options, {
+			name: 'PHP File Paths Plugin',
+			baseDataPath: 'options',
+		} );
+
+		this.options = options;
+	}
+
+	apply( compiler ) {
+		const pluginName = this.constructor.name;
+
+		compiler.hooks.thisCompilation.tap( pluginName, () => {
+			this.constructor.paths = getPhpFilePaths( this.options.props );
+		} );
+	}
+}
+
+module.exports = { PhpFilePathsPlugin };
--- a/tools/webpack/blocks.js
+++ b/tools/webpack/blocks.js
@@ -4,11 +4,13 @@
 const CopyWebpackPlugin = require( 'copy-webpack-plugin' );
 const { join, sep } = require( 'path' );
 const fastGlob = require( 'fast-glob' );
+const { realpathSync } = require( 'fs' );
 
 /**
  * WordPress dependencies
  */
 const DependencyExtractionWebpackPlugin = require( '@wordpress/dependency-extraction-webpack-plugin' );
+const { getRenderPropPaths } = require( '@wordpress/scripts/utils' );
 
 /**
  * Internal dependencies
@@ -77,6 +79,28 @@ const createEntrypoints = () => {
 	}, {} );
 };
 
+/**
+ * The plugin recomputes the render paths once on each compilation. It is necessary to avoid repeating processing
+ * when filtering every discovered PHP file in the source folder. This is the most performant way to ensure that
+ * changes in `block.json` files are picked up in watch mode.
+ */
+class RenderPathsPlugin {
+	/**
+	 * Paths with the `render` props included in `block.json` files.
+	 *
+	 * @type {string[]}
+	 */
+	static renderPaths;
+
+	apply( compiler ) {
+		const pluginName = this.constructor.name;
+
+		compiler.hooks.thisCompilation.tap( pluginName, () => {
+			this.constructor.renderPaths = getRenderPropPaths();
+		} );
+	}
+}
+
 module.exports = [
 	{
 		...baseConfig,
@@ -90,6 +114,7 @@ module.exports = [
 		plugins: [
 			...plugins,
 			new DependencyExtractionWebpackPlugin( { injectPolyfill: false } ),
+			new RenderPathsPlugin(),
 			new CopyWebpackPlugin( {
 				patterns: [].concat(
 					[
@@ -127,17 +152,32 @@ module.exports = [
 							'build/widgets/blocks/',
 					} ).flatMap( ( [ from, to ] ) => [
 						{
-							from: `${ from }/**/index.php`,
+							from: `${ from }/**/*.php`,
 							to( { absoluteFilename } ) {
-								const [ , dirname ] = absoluteFilename.match(
-									new RegExp(
-										`([\\w-]+)${ escapeRegExp(
-											sep
-										) }index\\.php$`
+								const [ , dirname, basename ] =
+									absoluteFilename.match(
+										new RegExp(
+											`([\\w-]+)${ escapeRegExp(
+												sep
+											) }([\\w-]+)\\.php$`
+										)
+									);
+
+								if ( basename === 'index' ) {
+									return join( to, `${ dirname }.php` );
+								}
+								return join( to, dirname, `${ basename }.php` );
+							},
+							filter: ( filepath ) => {
+								return (
+									filepath.endsWith( sep + 'index.php' ) ||
+									RenderPathsPlugin.renderPaths.includes(
+										realpathSync( filepath ).replace(
+											/\\/g,
+											'/'
+										)
 									)
 								);
-
-								return join( to, `${ dirname }.php` );
 							},
 							transform: ( content ) => {
 								const prefix = 'gutenberg_';
--- a/tools/webpack/blocks.js
+++ b/tools/webpack/blocks.js
@@ -152,7 +152,7 @@ module.exports = [
 							filter: ( filepath ) => {
 								return (
 									filepath.endsWith( sep + 'index.php' ) ||
-									PhpFilePathsPlugin.renderPaths.includes(
+									PhpFilePathsPlugin.paths.includes(
 										realpathSync( filepath ).replace(
 											/\\/g,
 											'/'
--- a/tools/webpack/blocks.js
+++ b/tools/webpack/blocks.js
@@ -94,7 +94,7 @@ module.exports = [
 			new DependencyExtractionWebpackPlugin( { injectPolyfill: false } ),
 			new PhpFilePathsPlugin( {
 				context: './packages/block-library/src/',
-				props: [ 'render' ],
+				props: [ 'render', 'variations' ],
 			} ),
 			new CopyWebpackPlugin( {
 				patterns: [].concat(
--- a/tools/webpack/blocks.js
+++ b/tools/webpack/blocks.js
@@ -92,7 +92,10 @@ module.exports = [
 		plugins: [
 			...plugins,
 			new DependencyExtractionWebpackPlugin( { injectPolyfill: false } ),
-			new PhpFilePathsPlugin( { props: [ 'render' ] } ),
+			new PhpFilePathsPlugin( {
+				context: './packages/block-library/src/',
+				props: [ 'render' ],
+			} ),
 			new CopyWebpackPlugin( {
 				patterns: [].concat(
 					[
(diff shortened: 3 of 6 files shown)


packages/scripts/utils/php-file-paths-plugin.js
```diff
@@ -0,0 +1,54 @@
...
+ * when filtering every discovered PHP file in the source folder. This is the most performant way to ensure that
+ * changes in `block.json` files are picked up in watch mode.
+ */
+class PhpFilePathsPlugin {
+	/**
+	 * PHP file paths from `render` and `variations` props found in `block.json` files.
+	 *
+	 * @type {string[]}
+	 */
+	static paths;
```
ockham, line 37: Probably need to change this so there won't be any collisions between class instances.

tools/webpack/blocks.js
```diff
@@ -90,6 +92,10 @@ module.exports = [
 		plugins: [
 			...plugins,
 			new DependencyExtractionWebpackPlugin( { injectPolyfill: false } ),
+			new PhpFilePathsPlugin( {
+				context: './packages/block-library/src/',
```
ockham, line 96: Should probably add the other paths used as `from` arg to the `CopyWebpackPlugin` instance as well.

gziolo:
What's missing to have it ready for review?

packages/scripts/utils/php-file-paths-plugin.js
ockham: Maybe I'm being overly cautious here. My concern is basically that if these paths are stored as a `static` class variable, they might get polluted. However, I'm not sure if the latter is realistic. While I can think of `npm run dev` running in parallel with another job that builds some 3rd-party package, that might not be enough to pollute class-level vars.

(OTOH, making it non-static -- in the context of a Webpack plugin -- seems to be non-trivial.)

ockham:
> What's missing to have it ready for review?

Really just the two things I've left notes for. Maybe you can weigh in on the [first one](https://github.com/WordPress/gutenberg/pull/63311#discussion_r1679183548); I can work on the [second one](https://github.com/WordPress/gutenberg/pull/63311#discussion_r1679212408).

packages/scripts/utils/php-file-paths-plugin.js
gziolo: It worked exactly the same before and no issues were reported. As far as I understand it, the webpack config is executed only once when you run `wp-scripts start`. It's when the plugin get instatiated, all subsequent changes to paths happen through th webpack lifecycle events.

tools/webpack/blocks.js
gziolo: It's an internal plugin that will never be used outside so I don't think we need to spend more time on it if it works as is.

gziolo: Ah, I see what you did. It's also imported to use within the Gutenberg plugin. Hmm, that complicates it a little bit. In general, `@wordpress/scripts` should only offer scripts and default configs. We discourage using internal utils and extending configs as they are subject to change.

packages/scripts/utils/php-file-paths-plugin.js
ockham: > It worked exactly the same before and no issues were reported. 

It wasn't used for Core blocks -- i.e. Gutenberg code itself -- before, so my concern was vaguely that e.g. running the GB build in watch mode might collide with running a wp-scripts command to build a third-party block.

> As far as I understand it, the webpack config is executed only once when you run `wp-scripts start`. It's when the plugin get instatiated, all subsequent changes to paths happen through th webpack lifecycle events.

Yeah, I guess we should be fine 🤔

tools/webpack/blocks.js
ockham: What I meant with my comment is that the `CopyWebpackPlugin` initialization includes other directories that might contain blocks (not just `packages/block-library/src`, but also `packages/edit-widgets/src/blocks` and `packages/widgets/src/blocks`):

https://github.com/WordPress/gutenberg/blob/18164f0fcd09a282157f02bba7fd737012cc44fd/tools/webpack/blocks.js#L121-L128

AFAICS, it's just three blocks total across those two directories, but in theory, they _could_ at some point include a `render` or `variations` field in one of their `block.json`s, which is why I said we might want to include them in `PhpFilePathsPlugin` as well.

On the downside, this would make the logic in `PhpFilePathsPlugin` even more complicated 😕 So maybe it's okay to omit them.

---

> Ah, I see what you did. It's also imported to use within the Gutenberg plugin. Hmm, that complicates it a little bit. In general, `@wordpress/scripts` should only offer scripts and default configs. We discourage using internal utils and extending configs as they are subject to change.

Hmm, are you saying we shouldn't altogether be doing what we're doing in this PR? AFAICS, we have some instances where we're doing "deep" imports from `@wordpress/scripts` in other parts of GB, e.g. [this](https://github.com/WordPress/gutenberg/blob/18164f0fcd09a282157f02bba7fd737012cc44fd/test/e2e/playwright.config.ts#L11). (I thought deep imports might be fine, especially if the relevant symbols aren't exported publicly at package level?)

gziolo: Deep imports work perfectly fine. It’s manageable inside the Gutenberg monorepo as breaking changes have a high chance of bubbling up during the development. However it should be discouraged for usage through npm. In effect, we should never support requests to make it backward compatible on that level.

gziolo: > AFAICS, it's just three blocks total across those two directories, but in theory, they could at some point include a render or variations field in one of their block.jsons, which is why I said we might want to include them in PhpFilePathsPlugin as well.

Got it! Initially, I didn’t think about it because I assumed these changes apply exclusively to the `@wordpress/scripts` package. I personally would be in favor of landing changes in the current shape and try a follow up if you feel strongly about covering blocks located in different packages.

ockham: No, I'd be more than happy to land this without further ado. If anything, I was trying to gauge if this has the potential of breaking things, but I guess that chance is pretty minimal.

I'll set the PR as ready for review.

gziolo: I'll try to give it a round of testing this week before approving it.

gziolo:
I can confirm that both `render` and `variations` fields defined in the `block.json` get correctly copied when running `npm run dev` (including edits to `block.json` while in the watch mode) and `npm run build` in the Gutenberg plugin:

Screenshot 2024-09-06 at 18 47 33

gziolo (approved):
It works as expected also for `@wordpress/scripts` package. Excellent job with this one 👏🏻

gziolo:
Looking at the `CHANGELOG.md` file that I updated to add new entry, this branch will need rebasing with the `trunk` to resolve some potential conflicts.

packages/scripts/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Enhancements
 
-- Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
+-   Inlines CSS files imported from other CSS files before optimization in the `build` command ([#61121](https://github.com/WordPress/gutenberg/pull/61121)).
+-   Adds automatic support for handling `variations` field from `block.json` in the `build` and `start` scripts ([#63311](https://github.com/WordPress/gutenberg/pull/63311)).
```
ockham: I _think_ this feature was already added by https://github.com/WordPress/gutenberg/pull/63098 (and [covered here](https://github.com/WordPress/gutenberg/pull/63098/files#diff-43f03963b157b5a7d4cdd75626b88deb84cf4406f6f748aa3bd3aa0d64c58ac7R7)). This PR only moves some logic around inside of `@wordpress/scripts` so that we can deep-import some required symbols from inside our project-level `tools/webpack/blocks.js`. I'm thus thinking to remove this Changelog entry if that's okay 😬

gziolo: Ah, okay. I missed that detail. Anyway, I retested it and it still works 🤣

PR #63314, opened 2024-07-09
Editor: Do not truncate post excerpt if not editable

### LINKED ISSUE #63313: Inspector: Template descriptions should not be elided
Screenshot 2024-07-09 at 16 05 08

_Excerpts_ are elided so that really verbose ones do not dominate the Inspector. This is okay because you can click it to open the edit UI and view the full excerpt. 

Template descriptions use the same functionality, but since not all template descriptions can be edited the result is incomplete descriptions in some cases, like the screenshot above.

Ideally only editable excerpts should be elided this way.

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/63313

From the issue:
>Excerpts are elided so that really verbose ones do not dominate the Inspector. This is okay because you can click it to open the edit UI and view the full excerpt.

>Template descriptions use the same functionality, but since not all template descriptions can be edited the result is incomplete descriptions in some cases, like the screenshot above.

>Ideally only editable excerpts should be elided this way.

## Testing Instructions
1. Visit templates where the excerpt/description cannot be edited like `Archive` template.
2. The full description is shown 

## Screenshots
Screenshot 2024-07-09 at 6 22 48 PM

### DIFF
--- a/packages/editor/src/components/post-excerpt/panel.js
+++ b/packages/editor/src/components/post-excerpt/panel.js
@@ -173,7 +173,7 @@ function PrivateExcerpt() {
 		return false;
 	}
 	const excerptText = !! excerpt && (
-		<Text align="left" numberOfLines={ 4 } truncate>
+		<Text align="left" numberOfLines={ 4 } truncate={ allowEditing }>
 			{ decodeEntities( excerpt ) }
 		</Text>
 	);


jasmussen (approved):
Fast one! Confirmed to fix it for me!

PR #60927, opened 2024-04-20
Components: Assess stabilization of `Navigator`

fullofcaffeine:
Issue: https://github.com/WordPress/gutenberg/issues/59418.

## What?

This is part of a [larger effort](https://github.com/WordPress/gutenberg/issues/59418) to remove `__experimental` prefix from all "experimental" components, effectively promoting them to regular stable components. See the related issue for more context.

## Why?

The strategy of prefixing exports with `__experimental` has become deprecated after the introduction of private APIs. 

## How?

1. Export it from components without the `__experimental` prefix;
1. Keep the old `__experimental` export for backwards compatibility;
1. Change all imports of the old `__experimental` in GB and components to the one without the prefix (including in storybook stories). Also, update the docs to refer to the new unprefixed component;
1. Add the component storybook `id` (get it from the storybook URL) to the `PREVIOUSLY_EXPERIMENTAL_COMPONENTS` const array in `manager-head.html` so that old experimental story paths are redirected to the new one;
1. Add a changelog for the change.

PR #63317, opened 2024-07-09
Navigator: simplify backwards navigation APIs

ciampo:
## What?

As discussed in #60927 , this PR introduces a few simplifications (and deprecations) to the `Navigator` component.

## Why?

We're simplifying the component's API ahead of promoting to stable.

## How?

- marked `NavigatorToParentButton`, `NavigatorBackButton.goToParent` and `useNavigator().goToParent()` as deprecated
- updated docs
- `useNavigator().goBack()` and `NavigatorBackButton` will navigate to the parent screen, instead of navigating to whatever screen was visited prior to the current one (according to the expected URL-based `path` assigned to screens)

## Questions

- should we hard deprecate `goToParent` and `NavigatorToParentButton` ?
- does the `replace` option make any sense with "go to parent"-only backwards navigation?

## Next steps

- Rename using agreed compound component notation 
- Export unprefixed, stable version
- Deprecate prefixed version

## Testing Instructions

`Navigator` should continue to behave as on `trunk` in Storybook and in the block editor

### DIFF
--- a/packages/components/src/navigator/navigator-back-button/hook.ts
+++ b/packages/components/src/navigator/navigator-back-button/hook.ts
@@ -25,19 +25,15 @@ export function useNavigatorBackButton(
 		...otherProps
 	} = useContextSystem( props, 'NavigatorBackButton' );
 
-	const { goBack, goToParent } = useNavigator();
+	const { goBack } = useNavigator();
 	const handleClick: React.MouseEventHandler< HTMLButtonElement > =
 		useCallback(
 			( e ) => {
 				e.preventDefault();
-				if ( goToParentProp ) {
-					goToParent();
-				} else {
-					goBack();
-				}
+				goBack();
 				onClick?.( e );
 			},
-			[ goToParentProp, goToParent, goBack, onClick ]
+			[ goBack, onClick ]
 		);
 
 	return {
--- a/packages/components/src/navigator/navigator-back-button/hook.ts
+++ b/packages/components/src/navigator/navigator-back-button/hook.ts
@@ -18,7 +18,10 @@ export function useNavigatorBackButton(
 	const {
 		onClick,
 		as = Button,
-		goToParent: goToParentProp = false,
+
+		// Deprecated
+		goToParent,
+
 		...otherProps
 	} = useContextSystem( props, 'NavigatorBackButton' );
 
--- a/packages/components/src/navigator/navigator-back-button/hook.ts
+++ b/packages/components/src/navigator/navigator-back-button/hook.ts
@@ -2,6 +2,7 @@
  * WordPress dependencies
  */
 import { useCallback } from '@wordpress/element';
+import deprecated from '@wordpress/deprecated';
 
 /**
  * Internal dependencies
@@ -24,6 +25,14 @@ export function useNavigatorBackButton(
 
 		...otherProps
 	} = useContextSystem( props, 'NavigatorBackButton' );
+
+	if ( goToParent !== undefined ) {
+		deprecated( '`goToParent` prop in wp.components.NavigatorBackButton', {
+			since: '6.7',
+			alternative:
+				'"back" navigations are always treated as going to the parent screen',
+		} );
+	}
 
 	const { goBack } = useNavigator();
 	const handleClick: React.MouseEventHandler< HTMLButtonElement > =
--- a/packages/components/src/navigator/navigator-back-button/hook.ts
+++ b/packages/components/src/navigator/navigator-back-button/hook.ts
@@ -2,7 +2,6 @@
  * WordPress dependencies
  */
 import { useCallback } from '@wordpress/element';
-import deprecated from '@wordpress/deprecated';
 
 /**
  * Internal dependencies
@@ -11,28 +10,17 @@ import type { WordPressComponentProps } from '../../context';
 import { useContextSystem } from '../../context';
 import Button from '../../button';
 import useNavigator from '../use-navigator';
-import type { NavigatorBackButtonHookProps } from '../types';
+import type { NavigatorBackButtonProps } from '../types';
 
 export function useNavigatorBackButton(
-	props: WordPressComponentProps< NavigatorBackButtonHookProps, 'button' >
+	props: WordPressComponentProps< NavigatorBackButtonProps, 'button' >
 ) {
 	const {
 		onClick,
 		as = Button,
 
-		// Deprecated
-		goToParent,
-
 		...otherProps
 	} = useContextSystem( props, 'NavigatorBackButton' );
-
-	if ( goToParent !== undefined ) {
-		deprecated( '`goToParent` prop in wp.components.NavigatorBackButton', {
-			since: '6.7',
-			alternative:
-				'"back" navigations are always treated as going to the parent screen',
-		} );
-	}
 
 	const { goBack } = useNavigator();
 	const handleClick: React.MouseEventHandler< HTMLButtonElement > =
--- a/packages/components/src/navigator/navigator-provider/README.md
+++ b/packages/components/src/navigator/navigator-provider/README.md
@@ -68,17 +68,13 @@ The available options are:
 - `focusTargetSelector`: `string`. An optional property used to specify the CSS selector used to restore focus on the matching element when navigating back.
 - `isBack`: `boolean`. An optional property used to specify whether the navigation should be considered as backwards (thus enabling focus restoration when possible, and causing the animation to be backwards too)
 
-### `goToParent`: `() => void;`
+### `goBack`: `() => void`
 
-The `goToParent` function allows navigating to the parent screen.
+The `goBack` function allows navigating to the parent screen.
 
 Parent/child navigation only works if the path you define are hierarchical (see note above).
 
 When a match is not found, the function will try to recursively navigate the path hierarchy until a matching screen (or the root `/`) are found.
-
-### `goBack`: `() => void`
-
-The `goBack` function allows navigating to the previous path.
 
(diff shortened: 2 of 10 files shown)


packages/components/src/navigator/test/index.tsx
```diff
@@ -148,23 +147,6 @@ function CustomNavigatorBackButton( {
...
-	return (
-		<NavigatorToParentButton
-			onClick={ () => {
-				// Used to spy on the values passed to `navigator.goBack`.
-				onClick?.( { type: 'goToParent' } );
-			} }
-			{ ...props }
-		/>
-	);
-}
```
ciampo: I removed this so that our tests don't use deprecated components, but I could also add them back and add an assertion to make sure that a warning message was logged to console

packages/components/src/navigator/test/index.tsx
```diff
@@ -19,14 +19,13 @@ import {
...
-	NavigatorToParentButton,
 	useNavigator,
 } from '..';
 import type { NavigateOptions } from '../types';
 
 const INVALID_HTML_ATTRIBUTE = {
-	raw: ' "\'><=invalid_path',
-	escaped: " &quot;'&gt;<=invalid_path",
+	raw: '/ "\'><=invalid_path',
+	escaped: "/ &quot;'&gt;<=invalid_path",
```
ciampo, line 29: Because of "going back" always behaving like "go to parent", all paths need to start with `/` in order to work well with the component's [pattern matching logic](https://github.com/WordPress/gutenberg/blob/378290a8b7cacdbdaa441c3c3f43436464f02dba/packages/components/src/navigator/utils/router.ts)

packages/components/src/navigator/navigator-to-parent-button/README.md
```diff
@@ -4,6 +4,8 @@
 This feature is still experimental. “Experimental” means this is an early implementation subject to drastic and breaking changes.
```
tyxla, line 4: Sounds like we can remove the experimental callout now.

packages/components/src/navigator/navigator-screen/README.md
```diff
@@ -16,6 +16,18 @@ The component accepts the following props:
...
+
+`Navigator` will treat "back" navigations as going to the parent screen — it is therefore responsibility of the consumer of the component to create the correct screen hierarchy.
+
+For example:
+
+-   `/` is the root of all paths. There should always be a screen with `path="/"`.
+-   `/parent/child` is a child of `/parent`.
+-   `/parent/child/grand-child` is a child of `/parent/child`.
+-   `/parent/:param` is a child of `/parent` as well.
+-   if the current screen has a `path` with value `/parent/child/grand-child`, when going "back" Navigator will try to recursively navigate the path hierarchy until a matching screen (or the root `/`) are found.
```
tyxla:
```suggestion
-   if the current screen has a `path` with value `/parent/child/grand-child`, when going "back" Navigator will try to recursively navigate the path hierarchy until a matching screen (or the root `/`) is found.
```

packages/components/src/navigator/test/index.tsx
tyxla: Doesn't hurt to keep them, with a `deprecated` logging assertion as you suggested.

packages/components/src/navigator/types.ts
```diff
@@ -20,17 +20,45 @@ export type NavigateOptions = {
 export type NavigateToParentOptions = Omit< NavigateOptions, 'isBack' >;
 
 export type NavigatorLocation = NavigateOptions & {
+	/**
+	 * Wether the current location is the initial one (ie. first in the stack).
```
tyxla:
```suggestion
	 * Whether the current location is the initial one (i.e., first in the stack).
```

packages/components/src/navigator/navigator-provider/README.md
```diff
@@ -65,28 +69,25 @@ The `goTo` function allows navigating to a given path. The second argument can a
...
-### `goToParent`: `() => void;`
+-   `focusTargetSelector`: `string`. An optional property used to specify the CSS selector used to restore focus on the matching element when navigating back.
+-   `isBack`: `boolean`. An optional property used to specify whether the navigation should be considered as backwards (thus enabling focus restoration when possible, and causing the animation to be backwards too)
+-   `skipFocus`: `boolean`. An optional property used to opt out of `Navigator`'s focus management, useful when the consumer of the component wants to implement their own custom focus management.
 
-The `goToParent` function allows navigating to the parent screen.
+### `goBack`: `( path: string, options: NavigateOptions ) => void`
 
-Parent/child navigation only works if the path you define are hierarchical (see note above).
+The `goBack` function allows navigating to the parent screen. Parent/child navigation only works if the path you define are hierarchical (see note above).
```
tyxla:
```suggestion
The `goBack` function allows navigating to the parent screen. Parent/child navigation only works if the paths you define are hierarchical (see note above).
```

packages/components/src/navigator/navigator-provider/README.md
```diff
@@ -10,38 +10,42 @@ The `NavigatorProvider` component allows rendering nested views/panels/menus (vi
...
-- `/` is the root of all paths. There should always be a screen with `path="/"`.
-- `/parent/child` is a child of `/parent`.
-- `/parent/child/grand-child` is a child of `/parent/child`.
-- `/parent/:param` is a child of `/parent` as well.
+
+-   `/` is the root of all paths. There should always be a screen with `path="/"`.
+-   `/parent/child` is a child of `/parent`.
+-   `/parent/child/grand-child` is a child of `/parent/child`.
+-   `/parent/:param` is a child of `/parent` as well.
+-   if the current screen has a `path` with value `/parent/child/grand-child`, when going "back" Navigator will try to recursively navigate the path hierarchy until a matching screen (or the root `/`) are found.
```
tyxla:
```suggestion
-   if the current screen has a `path` with value `/parent/child/grand-child`, when going "back" Navigator will try to recursively navigate the path hierarchy until a matching screen (or the root `/`) is found.
```

packages/components/src/navigator/navigator-provider/README.md
```diff
@@ -65,28 +69,25 @@ The `goTo` function allows navigating to a given path. The second argument can a
...
 
-Parent/child navigation only works if the path you define are hierarchical (see note above).
+The `goBack` function allows navigating to the parent screen. Parent/child navigation only works if the path you define are hierarchical (see note above).
 
 When a match is not found, the function will try to recursively navigate the path hierarchy until a matching screen (or the root `/`) are found.
 
-### `goBack`: `() => void`
-
-The `goBack` function allows navigating to the previous path.
+The available options are the same as for the `goTo` method minus the `isBack` property, which is not available for the `goBack` method.
```
tyxla:
```suggestion
The available options are the same as for the `goTo` method, except for the `isBack` property, which is not available for the `goBack` method.
```

packages/components/src/navigator/navigator-back-button/hook.ts
```diff
@@ -18,23 +19,30 @@ export function useNavigatorBackButton(
...
+
 		...otherProps
 	} = useContextSystem( props, 'NavigatorBackButton' );
 
-	const { goBack, goToParent } = useNavigator();
+	if ( goToParent !== undefined ) {
+		deprecated( '`goToParent` prop in wp.components.NavigatorBackButton', {
+			since: '6.7',
+			alternative:
+				'"back" navigations are always treated as going to the parent screen',
```
tyxla: Should we just suggest `goBack()`? For someone attempting to use `goToParent`, it's not immediately clear in their IDE that `goBack` is the recommended alternative.

tyxla (approved):
This looks good to me 👍 

Left a few minor suggestions.

I appreciate the simplification here. Makes the API clearer and more predictable, and the consumers better aware that they should prepare their navigation structure in a more hierarchical one-to-one way.

packages/components/src/navigator/navigator-to-parent-button/README.md
ciampo: I was planning on removing those experimental callouts and Storybook tags once we also export the component without the experimental prefix (in a follow-up PR)

packages/components/src/navigator/navigator-back-button/hook.ts
ciampo: The `NavigatorBackButton` doesn't have a `goBack` prop, since "going back" was already the default behavior. The `goToParent` prop was changing that behaviour, but now that the default behaviour of "going back" is the same as "going to parent", that prop basically became a no-op and can just be omitted.

tyxla: Ah, got it, makes sense.

ciampo: Actually thanks to your comment, I realised that this deprecation was not necessary because that was a code path that could not be hit anymore. I just removed the code instead: https://github.com/WordPress/gutenberg/pull/63317/commits/da79e357be6f05aa1f02141ad9dd130847c5995f

packages/components/src/navigator/test/index.tsx
ciampo: Alright! Added some specific tests for deprecated APIs https://github.com/WordPress/gutenberg/pull/63317/commits/f1f688868137b9ff576cf00d71ee4fd9524f912b

packages/components/src/navigator/navigator-back-button/hook.ts
tyxla: Perfect 🧹

PR #63322, opened 2024-07-09
Core Data: Support entities in the 'canUser' selector

### LINKED ISSUE #43751: Core Data selector canUser does not handle entity records

Mamaduka:
## What?
Resolves #43751.
Alternative to #63292.

PR updates the `canUser` selector to support core data entities. The new supported and recommended syntax:

```js
canUser( action, { kind, name, id } );
```

## Why?
* `canUser` - currently only supports resources in the `wp/v2` namespace. Additionally, you must know the final REST API path. Typically, however, only an entity's kind and name are known.
* `canUserEntityRecord` - is limited to only the edit action check. It was a wrapper for `canUser` and only supported the `wp/v2` namespace, which hasn't been required since WP 5.9.

The update also opens up the possibility of leveraging ["target hints"](https://github.com/WordPress/gutenberg/pull/61644#discussion_r1600465798) and bulk fetch permissions data for DataViews.

## Todo
- [x] Update the selector DocBlock and maybe types.
- [x] Deprecate `canUserEditEntityRecord` selector.

### Basic Testing
```js
wp.data.select( 'core' ).canUser( 'update', { kind: 'root', name: 'media', id: 1342 } );
wp.data.select( 'core' ).canUser( 'create', { kind: 'root', name: 'site' } );
wp.data.select( 'core' ).canUser( 'delete', { kind: 'root', name: 'site' } );

// Running this should log deprecation warning.
wp.data.select( 'core' ).canUserEditEntityRecord( 'root', 'site' );
```

### Testing canUserEntityRecord changes
1. Open a post.
2. Add Title and Excerpt blocks.
3. Confirm you can edit them as before.

Same.

## Screenshot
CleanShot 2024-07-10 at 21 38 47

### DIFF
--- a/packages/core-data/README.md
+++ b/packages/core-data/README.md
@@ -344,7 +344,7 @@ _Parameters_
 
 _Returns_
 
--   `boolean | undefined | null`: Whether or not the user can perform the action, or `undefined` if the OPTIONS request is still being made.
+-   `boolean | undefined`: Whether or not the user can perform the action, or `undefined` if the OPTIONS request is still being made.
 
 ### canUserEditEntityRecord
 
@@ -363,7 +363,7 @@ _Parameters_
 
 _Returns_
 
--   `boolean | undefined | null`: Whether or not the user can edit, or `undefined` if the OPTIONS request is still being made.
+-   `boolean | undefined`: Whether or not the user can edit, or `undefined` if the OPTIONS request is still being made.
 
 ### getAuthors
 
--- a/packages/core-data/README.md
+++ b/packages/core-data/README.md
@@ -344,7 +344,7 @@ _Parameters_
 
 _Returns_
 
--   `boolean | undefined`: Whether or not the user can perform the action, or `undefined` if the OPTIONS request is still being made.
+-   `boolean | undefined | null`: Whether or not the user can perform the action, or `undefined` if the OPTIONS request is still being made.
 
 ### canUserEditEntityRecord
 
@@ -363,7 +363,7 @@ _Parameters_
 
 _Returns_
 
--   `boolean | undefined`: Whether or not the user can edit, or `undefined` if the OPTIONS request is still being made.
+-   `boolean | undefined | null`: Whether or not the user can edit, or `undefined` if the OPTIONS request is still being made.
 
 ### getAuthors
 
--- a/packages/core-data/README.md
+++ b/packages/core-data/README.md
@@ -339,7 +339,7 @@ _Parameters_
 
 -   _state_ `State`: Data state.
 -   _action_ `string`: Action to check. One of: 'create', 'read', 'update', 'delete'.
--   _resource_ `string | Record< string, any >`: REST resource to check, e.g. 'media' or 'posts'.
+-   _resource_ `string | EntityResource`: Entity resource to check. Accepts entity object `{ kind: 'root', name: 'media', id: 1 }` or REST base as a string - `media`.
 -   _id_ `EntityRecordKey`: Optional ID of the rest resource to check.
 
 _Returns_
--- a/packages/core-data/README.md
+++ b/packages/core-data/README.md
@@ -339,7 +339,7 @@ _Parameters_
 
 -   _state_ `State`: Data state.
 -   _action_ `string`: Action to check. One of: 'create', 'read', 'update', 'delete'.
--   _resource_ `string`: REST resource to check, e.g. 'media' or 'posts'.
+-   _resource_ `string | Record< string, any >`: REST resource to check, e.g. 'media' or 'posts'.
 -   _id_ `EntityRecordKey`: Optional ID of the rest resource to check.
 
 _Returns_
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -352,10 +352,11 @@ export const getEmbedPreview =
  * Checks whether the current user can perform the given action on the given
  * REST resource.
  *
- * @param {string}  requestedAction Action to check. One of: 'create', 'read', 'update',
- *                                  'delete'.
- * @param {string}  resource        REST resource to check, e.g. 'media' or 'posts'.
- * @param {?string} id              ID of the rest resource to check.
+ * @param {string}        requestedAction Action to check. One of: 'create', 'read', 'update',
+ *                                        'delete'.
+ * @param {string|Object} resource        Entity resource to check. Accepts entity object `{ kind: 'root', name: 'media', id: 1 }`
+ *                                        or REST base as a string - `media`.
+ * @param {?string}       id              ID of the rest resource to check.
  */
 export const canUser =
 	( requestedAction, resource, id ) =>
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -370,7 +370,7 @@ export const canUser =
 		let resourcePath = null;
 		if ( typeof resource === 'object' ) {
 			if ( ! resource.kind || ! resource.name ) {
-				return;
+				throw new Error( 'The entity resource object is not valid.' );
 			}
 
 			const configs = await dispatch(
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -360,14 +360,34 @@ export const getEmbedPreview =
 export const canUser =
 	( requestedAction, resource, id ) =>
 	async ( { dispatch, registry } ) => {
-		const { hasStartedResolution } = registry.select( STORE_NAME );
-
-		const resourcePath = id ? `${ resource }/${ id }` : resource;
 		const retrievedActions = [ 'create', 'read', 'update', 'delete' ];
 
 		if ( ! retrievedActions.includes( requestedAction ) ) {
 			throw new Error( `'${ requestedAction }' is not a valid action.` );
 		}
+
+		let resourcePath = null;
+		if ( typeof resource === 'object' ) {
+			const configs = await dispatch(
+				getOrLoadEntitiesConfig( resource.kind, resource.name )
+			);
+			const entityConfig = configs.find(
+				( config ) =>
+					config.name === resource.name &&
+					config.kind === resource.name
+			);
+			if ( ! entityConfig ) {
+				return;
+			}
+
+			resourcePath =
+				entityConfig.baseURL + ( resource.id ? '/' + resource.id : '' );
+		} else {
+			// @todo: Maybe warn when detecting a legacy usage.
+			resourcePath = `/wp/v2/${ resource }` + ( id ? '/' + id : '' );
+		}
+
+		const { hasStartedResolution } = registry.select( STORE_NAME );
 
 		// Prevent resolving the same resource twice.
 		for ( const relatedAction of retrievedActions ) {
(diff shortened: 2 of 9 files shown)


SantosGuillamot:
I've tested editing post meta bindings and post title and everything seems to be working as expected 🙂 

Regarding the code, it looks good after a quick look, but I'll defer the review to other folks because I am not fully familiar with the use cases, and I don't have the full context.

packages/core-data/src/resolvers.js
```diff
@@ -435,16 +460,7 @@ export const canUser =
...
-		const entityConfig = configs.find(
-			( config ) => config.name === name && config.kind === kind
-		);
-		if ( ! entityConfig ) {
-			return;
-		}
-
-		const resource = entityConfig.__unstable_rest_base;
-		await dispatch( canUser( 'update', resource, recordId ) );
+		await dispatch( canUser( 'update', { kind, name, id: recordId } ) );
```
ellatrix, line 468: Should we also deprecate this resolver?

Mamaduka: Definitely! I'll do that as the final step before morning.

ellatrix (approved):
I quite like this! Looks good to me at least, but would appreciate a second approval of someone highly familiar with `core-data`.

packages/core-data/src/selectors.ts
```diff
@@ -1145,10 +1145,17 @@ export function isPreviewEmbedFallback( state: State, url: string ): boolean {
...
 	action: string,
-	resource: string,
+	resource: string | Record< string, any >,
 	id?: EntityRecordKey
 ): boolean | undefined {
-	const key = [ action, resource, id ].filter( Boolean ).join( '/' );
+	const key = (
+		typeof resource === 'object'
+			? [ action, resource.kind, resource.name, resource.id ]
+			: [ action, resource, id ]
```
jsnajdr, line 1162: Let's do some consistency checks on the `resource` object, and also prevent conflicting keys. The following calls (some of them buggy) lead to the same `key`:
```js
canUser( 'update', 'posts' );
canUser( 'update', { id: 'posts' } );
canUser( 'update', { name: 'posts' } );
```
The key should have some prefix (`ent:postType/post/1`, `res:posts/1`) and also we should trigger failure if there is `resource.id` and no `resource.name` etc.

ellatrix: Where is there a benefit for strings? It doesn't look like it's needed for memoization?

jsnajdr: I'm talking about the `key` that is used for lookup in Redux state: `state.userPermissions[ key ]`. That is a string. The `resource` parameter for `canUser` is another beast.

Mamaduka: I guess we can return `null` from the selector when the resource is an object, but the `kind` and `name` properties are missing. These two are required to make any entity queries. Resolve can bail early and not make a request.

I can add a warning or error, but these are rare in WP stores.

jsnajdr: Yes, that's what I meant by "failure", simply returning `null` 🙂 All we need is to avoid storing or returning the wrong data.

Mamaduka: Updated in 19e01ea3ac24e542514ddae27addbfaabcff648b.

jsnajdr: One more idea: can we return `false` when the entity is malformed? With that we can avoid making the return type too complicated with `boolean | undefined | null`. And `false` is still a good answer. `canUser` perform some ill-defined operation? The answer is no.

packages/core-data/src/selectors.ts
```diff
@@ -1145,10 +1145,22 @@ export function isPreviewEmbedFallback( state: State, url: string ): boolean {
 export function canUser(
 	state: State,
 	action: string,
-	resource: string,
+	resource: string | Record< string, any >,
```
Mamaduka: Should I swap this with `string | { kind: string; name: string; id?: EntityRecordKey }`?

jsnajdr: Yes. I'd also consider giving the type a name (like `EntityResource`) and the the type is `string | EntityResource`. Then it's more clear what this kind of parameter means.

packages/core-data/README.md
```diff
@@ -339,7 +339,7 @@ _Parameters_
 
 -   _state_ `State`: Data state.
 -   _action_ `string`: Action to check. One of: 'create', 'read', 'update', 'delete'.
--   _resource_ `string`: REST resource to check, e.g. 'media' or 'posts'.
+-   _resource_ `string | Record< string, any >`: REST resource to check, e.g. 'media' or 'posts'.
```
tyxla: To the `media` or `posts` it would be useful to append an example of how the object version could look like.

packages/core-data/src/resolvers.js
```diff
@@ -360,15 +360,39 @@ export const getEmbedPreview =
...
 		const retrievedActions = [ 'create', 'read', 'update', 'delete' ];
 
 		if ( ! retrievedActions.includes( requestedAction ) ) {
 			throw new Error( `'${ requestedAction }' is not a valid action.` );
 		}
 
+		let resourcePath = null;
+		if ( typeof resource === 'object' ) {
+			if ( ! resource.kind || ! resource.name ) {
+				return;
```
tyxla: Should we log an error in case `kind` or `name` aren't provided?

packages/core-data/src/resolvers.js
```diff
@@ -360,15 +360,39 @@ export const getEmbedPreview =
...
+
+			const configs = await dispatch(
+				getOrLoadEntitiesConfig( resource.kind, resource.name )
+			);
+			const entityConfig = configs.find(
+				( config ) =>
+					config.name === resource.name &&
+					config.kind === resource.kind
+			);
+			if ( ! entityConfig ) {
```
tyxla, line 384: Should we log an error in case `kind` or `name` don't correspond to an entity type?

packages/core-data/src/resolvers.js
```diff
@@ -416,10 +440,15 @@ export const canUser =
 
 		registry.batch( () => {
 			for ( const action of retrievedActions ) {
-				dispatch.receiveUserPermission(
-					`${ action }/${ resourcePath }`,
-					permissions[ action ]
-				);
+				const key = (
+					typeof resource === 'object'
+						? [ action, resource.kind, resource.name, resource.id ]
```
tyxla, line 446: No need to check for whether `resource` is properly formed?

packages/core-data/src/selectors.ts
tyxla: Makes sense. I saw instances above where we don't even check if the `kind` or `name` fields exist before using them, so better types will definitely help.

packages/core-data/src/selectors.ts
tyxla: It could also be useful to log an error so the developer knows about unexpected usage. 

@Mamaduka I see you mentioned those are rare in stores, but at the same time, when debugging store errors one often has to go deep, exactly because errors will often be silent or swallowed.

packages/core-data/src/resolvers.js
Mamaduka: We check this at the top and return early.

packages/core-data/src/selectors.ts
Mamaduka: We could throw an error when the resource object is malformed, and the error will be accessible via `getResolutionError`. We're already doing something similar when action isn't supported.

I don't know the historical reason why resolvers/selectors don't log errors; my best guess is that people often report them as bugs. So basically, support overhead.

cc @youknowriad

packages/core-data/src/resolvers.js
Mamaduka: No other entity resolver does this; maybe we should re-evaluate missing config cases together?

packages/core-data/src/selectors.ts
Mamaduka: Updated in 3246e87a3c12cc31bb1e853a83f2d24d50cfd15e.

Mamaduka:
I have resolved most of the feedback and to-do items. It should be ready for another review.

packages/core-data/src/selectors.ts
jsnajdr: I'm not very happy about throwing an error from the selector. It's not forward compatible. It's likely that we will add a new resource type in the future (e.g., for checking capabilities on the site) and then the new code will be failing on an older WordPress. It's better to return `false`, that means you cannot do the thing you ask for on an older WP version, and your new feature is effectively disabled. Let's just log a warning to the console, so that the error is not completely swallowed.

Mamaduka: The error is thrown from resolver b1e5dfc2ee25026e0892aa2a79525dd2e7d42bf3.

I think returning `false` is the correct approach. Core does the same with `current_user_can`. Calling `current_user_can( 'fly' )` will return `false`. It's up to the devs to pass the correct data. Otherwise, we'll have to pollute the codebase with warnings.

Fun fact: Calling `current_user_can( 'exist' )` will always return `true`.

jsnajdr: > The error is thrown from resolver

Oh, I missed that, in resolver it's fine.

> Fun fact: Calling `current_user_can( 'exist' )` will always return true.

It was added in 2010 here: https://core.trac.wordpress.org/ticket/14696#comment:5

> Add an 'exist' capability that all users have, even if they don't have a role on a blog. current_user_can('exist'), yeah, cute. Suggestions welcome.

jsnajdr (approved):
🚢

Mamaduka:
Thanks everyone for the reviews and feedback. I rebased the branch just in case and will merge after the CI checks are green.

fabiankaegy:
Hey @Mamaduka  👋 

Thanks in advance :)

PR #63333, opened 2024-07-10
Grid: Place new block after currently selected block when using slash inserter and splitting text

noisysocks:
Part of https://github.com/WordPress/gutenberg/issues/57478.

Currently in `trunk` if you use the slash inserter in a manual grid, or split a block (press <kbd>Enter</kbd> while typing text in a paragraph), the newly created block is inserted into the top-left-most empty cell.

This PR changes that so that the newly created block is positioned after the previously selected block. Specifically, it begins at the bottom right cell of the previously selected block and scans left→right and top→bottom for the first empty cell. It will auto create rows if needed.

To test, you'll need to disable the `useEffect` in `packages/block-editor/src/components/global-styles/typography-panel.js`. See https://github.com/WordPress/gutenberg/pull/63333#issuecomment-2222134002. Here's a patch that does that:

```diff
diff --git a/packages/block-editor/src/components/global-styles/typography-panel.js b/packages/block-editor/src/components/global-styles/typography-panel.js
index c497ea46833..52473538963 100644
--- a/packages/block-editor/src/components/global-styles/typography-panel.js
+++ b/packages/block-editor/src/components/global-styles/typography-panel.js
@@ -257,6 +257,7 @@ export default function TypographyPanel( {
 
 	// Check if previous font style and weight values are available in the new font family
 	useEffect( () => {
+		return;
 		const { fontStyles, fontWeights, isSystemFont } =
 			getFontStylesAndWeights( fontFamilyFaces );
 		const hasFontStyle = fontStyles?.some(
```

### DIFF
--- a/packages/block-editor/src/components/grid/use-grid-layout-sync.js
+++ b/packages/block-editor/src/components/grid/use-grid-layout-sync.js
@@ -2,7 +2,8 @@
  * WordPress dependencies
  */
 import { useDispatch, useSelect } from '@wordpress/data';
-import { useEffect } from '@wordpress/element';
+import { useEffect, useMemo } from '@wordpress/element';
+import { usePrevious } from '@wordpress/compose';
 
 /**
  * Internal dependencies
@@ -11,13 +12,15 @@ import { store as blockEditorStore } from '../../store';
 import { GridRect } from './utils';
 
 export function useGridLayoutSync( { clientId: gridClientId } ) {
-	const { gridLayout, blockOrder } = useSelect(
+	const { gridLayout, blockOrder, selectedBlockLayout } = useSelect(
 		( select ) => {
 			const { getBlockAttributes, getBlockOrder } =
 				select( blockEditorStore );
+			const selectedBlock = select( blockEditorStore ).getSelectedBlock();
 			return {
 				gridLayout: getBlockAttributes( gridClientId ).layout ?? {},
 				blockOrder: getBlockOrder( gridClientId ),
+				selectedBlockLayout: selectedBlock?.attributes.style?.layout,
 			};
 		},
 		[ gridClientId ]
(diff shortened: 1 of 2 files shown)


tellthemachines (commented):
Hmm, this isn't fixing the issue for me:

Screenshot 2024-07-10 at 4 47 09 PM

I think the problem is [occupiedRects](https://github.com/WordPress/gutenberg/blob/fix/grid-layout-sync/packages/block-editor/src/components/grid/grid-visualizer.js#L125) isn't refreshing when the amount of gridItems changes, possibly because the object itself hasn't changed?

noisysocks:
So weird— it looks like the call to `updateBlockAttribute()` in `useGridLayoutSync` isn't actually setting `columnStart` and `rowStart` on the block when a new block is added. Will investigate more deeply later.

noisysocks:
I figured out what's going on!

There's a `useEffect()` in `packages/block-editor/src/components/global-styles/typography-panel.js` that runs when the typography panel mounts. Sometimes, this effect runs _after_ the `useGridLayoutSync` effect. When this happens the typography `useEffect` will call `setAttributes()` with an outdated `value` and clobber the `columnStart` and `rowStart` that the grid sync effect had just added.

Not sure how to fix this yet. I asked in https://github.com/WordPress/gutenberg/pull/61915/files#r1673421903 whether the typography `useEffect` is actually needed.

For now I think let's ignore this.

noisysocks:
Added 1adc921 which updates the placement logic in `useGridLayoutSync` to better accommodate the slash inserter and block splitting. Will update PR description.

tellthemachines (approved):
Testing with the patch applied this is working well! I think we can go ahead and merge it and follow up with a fix for that Typography effect.

Issue #55238, opened 2023-10-11
Phase 3: Collaboration > Media Library


ramonjd:
Tasks that are a part of the Phase 3 [Media Library work described here](https://make.wordpress.org/core/2023/07/07/media-library/).

## Design

![Media](https://github.com/WordPress/gutenberg/assets/846565/09595fe6-d0aa-4e73-8b5d-8ce3ea376d7f)

### i1 — parity for the existing media library

- Use [DataViews](https://github.com/WordPress/gutenberg/issues/55083) to implement a Media screen. With fields including:
  - Media preview
  - Author
  - Upload date
  - File name
  - File size
  - Pixel dimensions
  - Media type (image, audio, pdf, etc.)
  - Attached to
- Uploading via _Add media_ button and drag and drop.
- A Media details / `MediaEditor` view, using DataView Forms. Allow updating metadata including:
  - Title
  - Caption
  - Alt text
  - Description
- Within the Media details / `MediaEditor` view, implement an image cropper that supports:
  - Aspect ratio
  - Crop
  - Rotation
  - Flip
  - Tracking edits associated with the `attachment` entity (for eventual support of undo/redo, and resetting state)
    - To support this, we might need the core data store to support clearing edits directly?
- Within the block editor, add a `MediaPicker` that implements a picker version of DataViews for selecting one or many media items.
  - Used for selecting images in the Image block, setting a Featured Image, setting a background image, or selecting media for the Audio, Video, and Gallery blocks
  - Support filtering by `Attached to` to show all media that is attached to the current post/page

### i2 — beyond 

- Categories and tags support. Some investigation took place in #53788.

Completed

- [List View: Add media previews to list view for gallery and image blocks #53381](https://github.com/WordPress/gutenberg/pull/53381)

## Related reading

- [Media Library - Make WordPress Core post](https://make.wordpress.org/core/2023/07/07/media-library/)
- #53788
- #59745
- #61447
- #58172
- https://github.com/WordPress/gutenberg/issues/66663
- https://github.com/WordPress/gutenberg/issues/62371

antpb:
Hello! in a recent Media component meeting we decided to start flagging trac tickets with `phase-3-media-triage` so we can in a future meeting discuss the impact of the Media Library revamp relative to the flagged issues. We see a lot of opportunity in this initiative to solve longstanding Media issues that have been held back by either backbone of the overall data structure of media items. 

[...]

Issue #22582, opened 2020-05-23
Add free-form crop


ajlende:
#21024 introduced cropping with fixed aspect ratios. Sometimes, however, it would be nice to crop the image without being restricted by the available aspect ratios like in the original prototype.

![rich-image prototype](https://user-images.githubusercontent.com/45363/72490878-44c7f580-37cd-11ea-9108-2e330c7d22d5.gif)

mrfoxtalbot:
Thank you for reporting this @ajlende. I believe this is being discussed in https://github.com/WordPress/gutenberg/issues/38990 

I think we can close this issue in favor of the other one as it has a much lengthier discussion.

ajlende:
https://github.com/WordPress/gutenberg/issues/38990 refers to the `aspect-ratio` CSS property that doesn't make edits to the image.

This issue refers to the actual image editor adding a drag crop interaction in conjunction to the zoom crop that exists.

andrewserong:
I know this is an old issue, but it'd be good to revive as there's still a need for good freeform cropping tools in the block editor. Another thing to look at is the visibility of the cropping presets. Currently, once you enter cropping mode, you need to know that this item in the toolbar allows you to select from cropping presets. I observed a user who was unable to find this dropdown after activating the cropping mode, so I'm wondering if while looking at free-form cropping, if we can also look at the visibility of the controls to select from presets. This is the dropdown I mean:

![image](https://github.com/WordPress/gutenberg/assets/14988353/77b07bf3-16e9-4b2b-9f85-d24a7163a1c4)

I've updated the issue description to reflect this note.

noisysocks:
Very unlikely this will make it into 6.7 – I've punted it in the [project board](https://github.com/orgs/WordPress/projects/210).

gyurmey2:
+1

t-hamano:
This issue has been punted to 6.8 and has not seen any activity since. Let's move this issue to the 6.9 Editor Task Board and see what can be done in 6.9.

ryanwelcher:
This was reviewed during the Sept 15, 2020 bug scrub. We need someone to start work on this or it will most likely be punted to 7.0

andrewserong:
> We need someone to start work on this or it will most likely be punted to 7.0

This is a very complex feature, my recommendation would be to skip for 6.9.

ramonjd:
Related:

- https://github.com/WordPress/gutenberg/issues/72734

[...]

PR #63335, opened 2024-07-10
[Experiment] Add freeform image cropper component

kevin940726:
## What?

The first draft of implementing a freeform image cropper component aiming to solve https://github.com/WordPress/gutenberg/issues/22582.

> [!NOTE]
> This PR is a WIP.
> Please see the [features and TODOs section below](#features-and-todos). We want to open a draft PR first to gather some very early feedback to ensure we are on the right track.

## Why?

See https://github.com/WordPress/gutenberg/issues/22582. The current image cropping tool lacks some advanced features we want to support. The library we used doesn't provide the features we need, and we can't find other third-party libraries that do. This component is an attempt to implement our own solution and potentially use it in the [Media Library revamp](https://github.com/WordPress/gutenberg/issues/55238).

## How?

Lots of math 🧠.

Add a new `<ImageCropper>` component in the `@wordpress/components` package. Currently, it's not exported at all and only available in the storybook environment for testing. We'll export it as a private component next once we reach a point where it can be integrated into the block editor.

The basic usage is as follow:

```js
<ImageCropper.Provider width={250} height={200} src={IMG}>
  <ImageCropper />
</ImageCroppper.Provider>
```

See the [storybook](http://localhost:50240/iframe.html?args=&id=components-imagecropper--inline) for advanced usage.

<h2 id="features-and-todos">Features and TODOs</h2>

Here's an unexhaustive list of supported features and TODOs with demo recordings.

- Resizing

  - <details><summary>✅ Resize the cropping window</summary><p>

    https://github.com/user-attachments/assets/bb8e63e8-8477-4924-9591-2ba727d664ad

    </p></details>

  - <details><summary>✅ Scale after resizing</summary><p>

    https://github.com/user-attachments/assets/bb8e63e8-8477-4924-9591-2ba727d664ad

    </p></details>

  - <details><summary>✅ Resize after rotation</summary><p>

    https://github.com/user-attachments/assets/d5d4da18-6d48-4426-9aae-837cca77b448

    </p></details>

  - <details><summary>Resize max bounds</summary><p>

    It should not be resized to be bigger than the image. Additionally, zooming out while resizing near the max bounds is nice-to-have. Here's a demo from the Google Photos iOS app:

    https://github.com/user-attachments/assets/961613ff-0048-447e-85f4-b29ef3b7b9ed

    </p></details>
- Dragging
  - <details><summary>✅ Drag/pan the image</summary><p>

    https://github.com/user-attachments/assets/e350fff9-af9c-4fa0-aa63-22a8988387a4

    </p></details>

  - <details><summary>✅ Drag max bounds</summary><p>

    https://github.com/user-attachments/assets/10120afa-ce98-4d46-83eb-d4fcc1b9da0b

    </p></details>

  - <details><summary>✅ Drag after rotation</summary><p>

    https://github.com/user-attachments/assets/2b3fdd24-a62c-4e7d-8c60-e84356e31207

    </p></details>

- Rotation
  - <details><summary>✅ Rotate the image around the center of the cropping window</summary><p>

    https://github.com/user-attachments/assets/4fd6b788-570c-4201-bfce-e83cc8d9e8c7

    </p></details>

  - <details><summary>✅ Rotate 90 degrees (clockwise and counterclockwise)</summary><p>    

    https://github.com/user-attachments/assets/75fed028-2fa6-47e8-9819-feaebb120e03

    </p></details>

  - <details><summary>✅ Scale after rotation</summary><p>

    https://github.com/user-attachments/assets/501b0622-b0c5-4f0a-b815-81b239d1ae7d

    </p></details>

  - <details><summary>Rotation animation</summary><p>

    The cropping window should also rotate with the image.

    https://github.com/user-attachments/assets/5c68a59e-0f21-4d23-a3a8-a49f913b6d68

    </p></details>
- Flipping
  - <details><summary>✅ Flip the image horizontally</summary><p>

    Flipping the image vertically can be achieved by rotating and flipping.

    https://github.com/user-attachments/assets/b6f39056-401e-4331-bb07-7a099d37f03a

    </p></details>

  - <details><summary>✅ Flip animation</summary><p>

    https://github.com/user-attachments/assets/b6f39056-401e-4331-bb07-7a099d37f03a

    </p></details>
- Zooming
  - <details><summary>✅ Zoom in/out the image around the cursor's position by pinching</summary><p>

    https://github.com/user-attachments/assets/169cb801-9e19-4642-8e74-77d2cbb6bdc0

    </p></details>

  - <details><summary>✅ Zoom after rotation</summary><p>

    https://github.com/user-attachments/assets/7a686a95-f8e9-4d3a-a41a-53a3245e7071

    </p></details>

  - <details><summary>✅ Zoom max and min bounds</summary><p>

    Currently, the maximum and minimum scale bounds are set to `1` and `10` respectively.

    </p></details>

  - Zoom with a mouse or keyboard.
  - Resize after zooming out

- Aspect ratio
  - Lock/Unlock the aspect ratio
- Responsive
  - <details><summary>✅ Resize the container after resizing the window</summary><p>

    It mostly works but has some bugs and lacks testing.

    </p></details>
- Accessibility
  - Keyboard shortcuts
  - Screen reader support
  - Reduced motion support
- Styling
  - Designers' approval
- Integration
  - <details><summary>✅ Integrate into the Image block</summary><p>

    Available under a gutenberg experiment "Redesigned image cropper".

    </p></details>
  - Integrate into the Media Library (#55238)
- Testing
  - Unit tests
  - End-to-end tests

## Testing Instructions

For testing in the image block in the editor, enable the gutenberg experiment under "Gutenberg" -> "Experiments" -> "Redesigned image cropper".

https://github.com/user-attachments/assets/64004b60-d61d-4b1d-ac8f-12d5911327b9

For testing the component in the storybook:
1. Run the storybook dev server (`npm run storybook:dev`)
2. Visit `http://localhost:50240/iframe.html?args=&id=components-imagecropper--inline` to view and play with the component.

This component is currently not accessible to the keyboard, but it's on the plan.


jasmussen:
Nice one, took it for a quick spin:

cropping

This is looking solid and promising. Small bug with zooming too much and it flips the image.

This is very nascent, so hard to fully review until it's in a branch, implemented for the Image block, but a few general observations/thoughts:

* The dark frame around it we probably want to avoid. 
* We should not have 8 resize handles. In the editor for resizing the image, we have 2, right and bottom. There's an open question around how the resize handles left and right will work when this gets transplanted into the canvas, that's likely also related to the dark frame for "scaling up", so those are some questions we will need to consider later. But we can at least remove the resize handles in the corner.
* We'll need to think about where to place the slider for rotating the image, and whether that slider is the right interface for it. E.g. it can't live inside the block toolbar directly. Something like the gradient direction tool may be useful here.

Nice work.

talldan:
> We'll need to think about where to place the slider for rotating the image, and whether that slider is the right interface for it. E.g. it can't live inside the block toolbar directly. Something like the gradient direction tool may be useful here.

If the number of resize handles were reduced, there could be rotation handles on the corners, or some applications have a handle that extends out like this:
Screenshot 2024-07-10 at 3 18 33 pm

andrewserong:
This is feeling really nice to use so far! Another couple of questions, and possibly for @jasmussen (and once we integrate with the image block):

* How might we expose aspect ratio controls at the same time? The dropdown in the block toolbar can be a bit hidden for users, I've seen some people miss them altogether.
* The scrollwheel to zoom feature works pretty well for me, but I didn't notice it at first. Would we also want to have zoom in / zoom out buttons, alongside the rotation control?

kevin940726:
> We'll need to think about where to place the slider for rotating the image, and whether that slider is the right interface for it.

Most cropping tools that inspire this PR have their rotation slider below the image. For instance, here's Google Photos web:

https://github.com/WordPress/gutenberg/assets/7753001/48b38ea1-8c9a-48af-abdc-44bdd36bc60c

The dimmed background is also needed in this case. I think we can also consider placing the slider at the bottom of the editing canvas? It can be toggled with a click in the block toolbar since it's an advanced usage.


> there could be rotation handles on the corners, or some applications have a handle that extends out like this

I also considered this, but IMO this sort of UI only makes sense when you're rotating the cropping window itself, not the background image. Because this PR currently assumes that every cropped image should still be an unrotated rectangle, rotating the cropping window doesn't make sense here.

> The scrollwheel to zoom feature works pretty well for me, but I didn't notice it at first. Would we also want to have zoom in / zoom out buttons, alongside the rotation control?

I think we should, at least for feature completion of accessibility. We can potentially add more keyboard controls too (<kbd>Ctrl</kbd>+<kbd>+</kbd>/<kbd>-</kbd>?) in addition to the UI control.

jasmussen:
> How might we expose aspect ratio controls at the same time? The dropdown in the block toolbar can be a bit hidden for users, I've seen some people miss them altogether.

We need a fresh design for this. The following much older design could be a starting point, but it needs a fresh take that is more like the existing and with a smaller visual diff. 
Crop and rotate in isolation
Crop and rotate in-canvas i2

For example, the modal may look fun at a quick glance, but _let's not do this_, it's not compatible with cropping tools that are likely to also exist in a future media library, and it also is a clunky amount of additional steps compared to a more lean and light-weight in-canvas cropping tool. I'm sharing it only because there may be some elements as far as aspect ratio that may be worth revisiting. 

Note specifically how the block toolbar itself becomes modal. Instead of having a single image block toolbar that grows wider with crop tools, this one proposes a drilldown effect for the block toolbar, affording better and clearer tools. Not sure how feasible, but it's worth resurfacing.

[Figma](https://www.figma.com/design/fyUf77Obv8KszngM2PQeMG/Image-Filters%2C-Editing-Tools%2C-Image-Masking?node-id=1-5886&t=2Ex2cIYBibm3AtUH-1).

> The scrollwheel to zoom feature works pretty well for me, but I didn't notice it at first. Would we also want to have zoom in / zoom out buttons, alongside the rotation control?

I like the slider in the toolbar, though I'm also aware that this is substantially more difficult to get right, than it looks at a glance. So also here it would seem useful to _free up space_ in the block toolbar, however we can. Since it's already modal through the apply/cancel actions, perhaps that can be a way to take up more space in the block toolbar?

talldan:
> I also considered this, but IMO this sort of UI only makes sense when you're rotating the cropping window itself, not the background image. Because this PR currently assumes that every cropped image should still be an unrotated rectangle, rotating the cropping window doesn't make sense here.

That's a good point 😄 

I think my main concern would be how this works in terms of inline editing when there are other blocks after the image, and also when there's a caption in that position. Something more minimal might be easier to wrangle. Having secondary UI controls in the inspector or toolbar is still always an option (and might be needed for accessibility).

It can be something to iterate on though, it'll be interesting to see the designs.

kevin940726:
https://github.com/user-attachments/assets/82e79ed2-741c-4928-9c2e-7d3566f43a7a

I added the rotation (90deg clockwise and counter-clockwise) feature. The reason `angle` and `turns`(can't think of a better name 😅) are separated is because the value for `angle` is between `-45` to `45`, and we need a way to distinguish between `angle: 45, turns: 0` and `angle: -45, turns: 1`.

Currently, `getImageBlob` for rotated images doesn't work as expected. I probably need some help on some math there and simplify the logic in `reducer` 😅.

packages/components/src/image-cropper/reducer.ts
```diff
@@ -0,0 +1,286 @@
...
+		flipped: false,
+		offset: { x: 0, y: 0 },
+		position: { x: 0, y: 0 },
+		size: { width, height },
+		isResizing: false,
+		isDragging: false,
+	};
+}
+
+function imageCropperReducer( state: State, action: Action ) {
```
talldan: It'd be great to add some inline comments to the code within this reducer. Geometry code like this is notoriously hard to read.

I spotted a bug, if I resize twice from the bottom-right, the top-left of the resizable box jumps position incorrectly:

https://github.com/user-attachments/assets/aff6fcf0-c210-4e9f-9374-135216cfb29e

I found it a little difficult to understand the intention of the code in the `RESIZE_WINDOW` action, so some comments would help greatly.

kevin940726: > I found it a little difficult to understand the intention of the code in the `RESIZE_WINDOW` action, so some comments would help greatly.

Yeah, that's one of the things I need help with 😅. Half the time I don't even know what I'm doing 🤦. I'd need to re-organize my thoughts and refactor this in a way that makes it easier to write comments.

packages/components/src/image-cropper/reducer.ts
```diff
@@ -0,0 +1,286 @@
...
+	offset: Position;
+	// The position of center of the image.
+	position: Position;
+	// The size of the cropper window.
+	size: Size;
+	// Whether the cropper window is resizing.
+	isResizing: boolean;
+	// Whether the image is dragging/moving.
+	isDragging: boolean;
+};
```
talldan: A good aim could be to model the state like this:
```js
State = {
    image: {
		// the image would always have a `top`/`left` of zero, because we apply `transform.translation` to move it. A rect is a web friendly storage format so might as well use it to keep things consistent.
        rect: { top, right, bottom, left, x, y, width, height }
    },
    imageTransforms: {
        rotate: number,
        scale: number, 
        translate: Vector2<number>,
        reflect: Vector2<bool>
    },
    // the resizable box
    clippingRect: { 
        rect: { top, right, bottom, left, x, y, width, height }
    },
    // UI states
    isResizing: bool,
    isDragging: bool,
}
```

I'm visualizing the image manipulation as two separate things. 

Firstly, if you completely disregard the 'clipping rect', there are distinct changes the user can make to the image. They can rotate it (slider), zoom it (mouse wheel) and translate it (drag it). As the user makes those changes the deltas are added to the transform state. The `image` state doesn't change at all until the user clicks 'apply', at which point the `transform` state is reset to zero.

Keeping the transform as a state like this is pretty important as the scaling and rotation need to be applied before translation (moving) when transforming something geometrically around its center (assuming the origin is the center of the image, if not it can be translated to its center first before scaling, and then translated back again).

The clipping rect is a separate thing on top of that, but from what I can tell, it doesn't affect the underlying image until the user applies the crop, so I think a `rect` is the only state needed for it.

There might be more needed after that, for example I notice that the clipping rect and the image are often translated together (after resizing the clipping area, the image and clipping rect are recentered), but I think it'd be better to store that as a separate `transform` object (e.g. `viewTransform`) that's applied to the resizable box and the image (after the other transforms) so that they both move by the exact same amount.

This reflects pretty closely how video games work in having a series of 'transforms' (model, view, projection), though they store them in matrices:
https://gamedev.stackexchange.com/questions/178643/the-view-matrix-finally-explained

packages/components/src/image-cropper/reducer.ts
talldan: The comment here might help - https://github.com/WordPress/gutenberg/pull/63335#discussion_r1680440743.

There might be one or two things I got wrong because I'm mostly doing this as a mental exercise, but I think a good rule is to not modify the underlying image coordinates until the user hits 'apply'. Instead store the manipulations in a 'transform' object that describes how the image has been manipulated. It might be similar to what you're doing already, but I found it difficult to tell.

I think the important thing is to get the reducer working correctly and then have the UI work from that state. It might be challenging with some of the opinionated UI pieces like `resizableBox`, but a solid foundation for the maths will make everything work much more reliably.

packages/components/src/image-cropper/reducer.ts
ramonjd: I was also going to comment that `turns` might be more portable/easily understood as a `rotate` value, possibly in combination with the `deg/radian`?

packages/components/src/image-cropper/styles.ts
```diff
@@ -0,0 +1,46 @@
...
+	cursor: move;
+	touch-action: none;
+`;
+
+export const Resizable = styled( ResizableBox )`
+	transform: translate(
+		var( --wp-cropper-window-x ),
+		var( --wp-cropper-window-y )
+	);
+	box-shadow: 0 0 0 100vmax rgba( 0, 0, 0, 0.5 );
```
ramonjd: This is a fantastic start. Awesome work @kevin940726 

What do you think about adding some simple guide rulers? I'm using Google photos as inspiration 😄 

https://github.com/user-attachments/assets/1353275d-c429-467d-97e0-193b7404f164



<details>

<summary>Example</summary>

```ts
export const Resizable = styled( ResizableBox )`
	transform: translate(
		var( --wp-cropper-window-x ),
		var( --wp-cropper-window-y )
	);
	box-shadow: 0 0 0 100vmax rgba( 0, 0, 0, 0.6 );
	&:active {
		box-shadow: 0 0 0 100vmax rgba( 0, 0, 0, 0.4 );
		&::after,
		&::before,
		${ Draggable }::after, ${ Draggable }::before {
			position: absolute;
			padding: 0;
			width: 1px;
			height: 100%;
			overflow: hidden;
			left: 33.33%;
			content: ' ';
			background: rgba( 255, 255, 255, 0.33 );
			z-index: 1000;
			display: block;
		}

		&::before {
			right: 33.33%;
			left: auto;
		}

		${ Draggable }::before {
			left: auto;
			width: 100%;
			height: 1px;
			top: 33.33%;
		}

		${ Draggable }::after {
			left: auto;
			width: 100%;
			height: 1px;
			bottom: 33.33%;
			top: auto;
		}
	}
`;

```

</details>


Also, I know it's eye candy, but a subtle animation during the post-crop translate (not dragging) would look pleasing to my eye at least 🤗

packages/components/src/image-cropper/stories/index.story.tsx
```diff
@@ -0,0 +1,140 @@
...
+		</ImageCropper.Provider>
+	);
+};
+
+export const Default: StoryObj< typeof ImageCropper.Provider > = Template.bind(
+	{}
+);
+Default.args = {
+	src: 'https://upload.wikimedia.org/wikipedia/commons/thumb/3/34/Hydrochoeris_hydrochaeris_in_Brazil_in_Petr%C3%B3polis%2C_Rio_de_Janeiro%2C_Brazil_09.jpg/1200px-Hydrochoeris_hydrochaeris_in_Brazil_in_Petr%C3%B3polis%2C_Rio_de_Janeiro%2C_Brazil_09.jpg',
+	width: 250,
```
ramonjd: Do you think it would be helpful to provide an example of a crop container that resizes with the window/parent container?

packages/components/src/image-cropper/reducer.ts
kevin940726: Do you mean a `rotate` and a `degree` value? I never like the naming of `turns` so I might switch to that 😅.

ramonjd: > Do you mean a rotate and a degree value? 

Oh, the controls are separate, right? Then, yes 😄

packages/components/src/image-cropper/stories/index.story.tsx
kevin940726: Yes, absolutely! It's on the TODO list but not supported yet. I think we can first implement the core features and then make it responsive. We also need to fix that it's not cropping the image in its original resolution 😅.

packages/components/src/image-cropper/styles.ts
kevin940726: Nice! I think we can also show it on dragging and rotating too!.

kevin940726: I pushed a commit that includes the rulers and the animations!

https://github.com/user-attachments/assets/03a92a8f-9345-488c-89ef-b0e76ebaac34

kevin940726:
I added a section of [features and TODOs](#features-and-todos) in the PR description. Please let me know if any of them isn't clear! I'll try to add more too.

ajlende:
Related to resizing the container, when we're using this in the block editor for the image block, the crop area should be the same size as the final image. One of the goals of core block editing is to make the editing mode look as close to the preview mode as possible.

packages/components/src/image-cropper/stories/index.story.tsx
```diff
@@ -0,0 +1,147 @@
...
+					/>
+				</FlexItem>
+
+				<Button
+					variant="secondary"
+					onClick={ () => {
+						dispatch( {
+							type: 'ROTATE_CLOCKWISE',
+							isCounterClockwise: true,
+						} );
```
talldan: Nit: this API is a little confusing as one onehand it's `ROTATE_CLOCKWISE` and then on the other it's `isCounterClockwise: true`. Maybe the action type could be `ROTATE_90` or something.

packages/components/src/image-cropper/reducer.ts
```diff
@@ -0,0 +1,423 @@
...
+	return {
+		image: {
+			width,
+			height,
+			x: 0,
+			y: 0,
+		},
+		transforms: {
+			angle: 0,
+			rotations: 0,
```
talldan: Out of interest, why does the angle needs to be stored separately to rotations? The two feel like the same thing to me.

ajlende: They are the same thing, but this matches the UI of the Google photos cropper that many of these interactions seemed to be based off of. One could deconstruct any angle into these two parts, but I haven't convinced myself which way would be better. Mostly thinking in terms of undo/redo.

image

ajlende:
I was playing around with more image croppers today, and the closest ones to what I had in mind for image block cropping were [Adobe Express](https://new.express.adobe.com) and [Canva](https://canva.com).

## Adobe Express

Adobe Express has two resizable boxes, essentially; one for the crop region and one for the image. It showcases how rotation can be done in-canvas without a slider like Google.

The part that doesn't work for us is that the positioning post crop will be different for us since we're working with an element that's part of the regular DOM flow and rules instead of a canvas where everything is absolutely positioned.

https://github.com/user-attachments/assets/815d36a4-50b8-4916-bb01-d6b884747830

## Canva

Canva "solves" the problem by not allowing the crop area to be changed while editing, so the size of the crop area in the page always matches the final result.

Canva also adds a feature that keeps the crop are in bounds of the image that Adobe Express doesn't handle.

https://github.com/user-attachments/assets/b0cd0852-f164-4f75-b2ab-e1bb6bcb1d57

## Google Photos

However, Google photos handles the reflow quite nicely by resizing the image on mouse up.

https://github.com/user-attachments/assets/5241bff8-40f3-435b-840f-cf8b54a67234

## Combining them

I think we should be able to combine the Adobe Express double box model with the Canva crop area bounds constraints and the Google Photos reflow mechanic to make something that will work in both the context of inline editing for the image block and as a fixed editor in the admin redesign.

packages/components/src/image-cropper/reducer.ts
kevin940726: Yep! We need two separate values here so that we can distinguish between `angle: 45, rotations: 0` and `angle: -45, rotations: 1`, which are both 45 degree but represent different states in the UI.

kevin940726:
> I think we should be able to combine the Adobe Express double box model with the Canva crop area bounds constraints and the Google Photos reflow mechanic to make something that will work in both the context of inline editing for the image block and as a fixed editor in the admin redesign.

What do you have in mind for using the double box model for inline editing? If we reflow(scale) the image after resizing then the image box will grow bigger to the point that it might be difficult to fit the box into the screen 🤔.

If the idea is to adopt the rotation button, then I think we can still do that using the Google Photos UX? Possibly something like a button rather than a slider below the cropper window?

I think we can use some design feedback here too 😆.

packages/components/src/image-cropper/stories/index.story.tsx
kevin940726: Good point! I renamed it to `ROTATE_90_DEG` in [2fd0c69](https://github.com/WordPress/gutenberg/pull/63335/commits/2fd0c69b9827aec6926461ce5e45a150e66aaf74).

talldan:
> If we reflow(scale) the image after resizing then the image box will grow bigger to the point that it might be difficult to fit the box into the screen 🤔.

That's the kind of thing that concerns me too. Along with the possibility of having elements of the cropping tools overlapping other blocks. Engaging spotlight mode during cropping might be an option to keep focus on the image block.

ajlende:
> > If we reflow(scale) the image after resizing then the image box will grow bigger to the point that it might be difficult to fit the box into the screen 🤔.
> 
> That's the kind of thing that concerns me too. Along with the possibility of having elements of the cropping tools overlapping other blocks. Engaging spotlight mode during cropping might be an option to keep focus on the image block.

We do have a mode where the page is zoomed out, so we can scale the page to any size to make room for the "ghost" image outside of the page if needed. https://github.com/WordPress/gutenberg/pull/63870 added an option for a fixed 50% scale with some other UI changes, but I don't see why we couldn't use that scaling to make more room for the contextual cropping.

kevin940726:
> We do have a mode where the page is zoomed out

Wouldn't zoom-out mode also zoom out the image cropping area, thus making it smaller to interact with?

Imagine a somewhat extreme example with a small cropping area.

image

If we scale up the cropping area to match the content size, the image background could be large enough that the rotation button is too far below to interact with. A solution would be moving the rotation button just below the cropping area, but I think that's essentially the same as the Google Photos example? 🤔

ajlende:
> Wouldn't zoom-out mode also zoom out the image cropping area, thus making it smaller to interact with?

Yes, but I don't expect zoom out would be required for cropping. You'd probably only want to switch to zoom out mode when you're cropping things like full-width images.

> If we scale up the cropping area to match the content size, the image background could be large enough that the rotation button is too far below to interact with. A solution would be moving the rotation button just below the cropping area, but I think that's essentially the same as the Google Photos example?

That's a good point, especially since we don't really have an "infinite canvas" to work with like some of these design tools. My first reaction is that it wouldn't be a problem (from a usability perspective) to have both, but saying yes to everything can lead to a busy UI that's also hard to use. Would be good to get some design input on this.

packages/components/src/image-cropper/reducer.ts
ajlende: Now that I've had some time to reflect on this, I think I would prefer the angle in radians as part of the transform, and in order to distinguish -45 from 45, the "tilt" can be saved as part of the UI state in degrees. This keeps the UI decoupled from the actual transform so it's easier to update the UI if needed.

kevin940726:
In recent commits, you can now test the new image cropper in the Image block. Enable the gutenberg experiment "Redesigned image cropper" under "Gutenberg" -> "Experiments" -> "Redesigned image cropper".

image

After enabling the experiment, this redesigned cropper will be the default cropper of the image block. Insert an image block and test it in the editor.

https://github.com/user-attachments/assets/64004b60-d61d-4b1d-ac8f-12d5911327b9

I hope this makes testing this PR easier for designers and contributors. Needless to mention, there are still lots of improvements, and the state is not meant to be the final design. We'll add more features and fix bugs as we continue to develop the feature. Thanks for testing! 🙇

jasmussen:
Thanks for working on this. This already feels vastly better than what exists. Nice work.

I did encounter a glitch, I was using a portrait image, and resized it first, in doing this, toggling cropping on/off resulted in differing aspect ratios:
cropping

One other aspect, when the images get very large, scrollbars get added. This is visible in your screenshots as well. Maybe this is fine? But it's worth testing this also on mobile, where I would worry that panning might pan the viewport rather than the image you're moving inside. Or worse yet, a bit of both which might get you in trouble. If we need to, can we simply crop off the edges outside your image frame? Right now they visually appear blurred, which is fancy, but in case it causes issues, simply not showing this might be fine.

Separately, it might be time to make progress on toolbar improvements. At present, it looks like this:

Image Block Toolbar, cropping before

Notably Link and Crop buttons disappear from the toolbar, when you're cropping. This is reasonable, but also a bit random, since text overlay remains. Cropping is effectively a mode of the block toolbar, so we might as well lean into that. Could we do this?

Image Block Toolbar, proposed cropping

Doing this would also better fit the zoom/rotate/aspect ratio tools which are currently missing in this implementation. Like so:

Image Block Toolbar, proposed cropping

talldan:
This works nicely so far, really great work here.

There's a few little things I noticed from testing (you're probably aware some of them):
- When engaging crop mode, there's an animation of some of the elements that doesn't look right:

https://github.com/user-attachments/assets/cab51014-f013-432e-958b-4d2e7eddb822

- When resizing I can see a couple of guide lines (one horizontal, one vertical), but they seem off-center and don't follow the [rule of thirds](https://photographylife.com/the-rule-of-thirds) system that most cropping tools use.

https://github.com/user-attachments/assets/1bb748fa-40f1-439f-8c92-c15dd6acd6d3

- I also think it might be a good idea to show the guide lines whenever crop mode is active, or maybe show them when hovering the image while crop mode is active.

- Also visible in the video above, the image shakes when dragging the handles a very tiny amount

- The sibling inserters can get in the way of the drag handles. Not sure if there's a way to turn them off while cropping.

- If you try cropping again while a previous crop operation is saving, the image can become stretched:

https://github.com/user-attachments/assets/9b1f023b-a825-4910-94c8-76f27973422b

ramonjd:
Closing this as it's a bit stale. We can return to learn from it later.

PR #63337, opened 2024-07-10
Tabs: hyphenate tab labels

ciampo:
## What?

Related to #62020

Add automatic hyphenation for tab labels in the `Tabs` component

## Why?

To prevent long tab labels from pushing the width of the tab list and causing unexpected scrolling behavior.

## How?

By using the [`hyphens`](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens) CSS property as [suggested](https://github.com/WordPress/gutenberg/issues/62020#issuecomment-2218248821) by @jameskoster 

## Testing Instructions

- Open Storybook
- Change the tab labels to become more verbose, including very long single words
- Observe how the labels are broken (hyphenated) into multiple lines, preventing the tablist from overflowing and causing horizontal scroll in the document

## Screenshots or screencast 

|---|---|
| Screenshot 2024-07-10 at 08 53 49 | Screenshot 2024-07-10 at 08 53 38 |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+-   `Tabs`: hyphenate tab labels ([#63337](https://github.com/WordPress/gutenberg/pull/63337)).
 -   `Tooltip`: Add support for `className` prop ([#63157](https://github.com/WordPress/gutenberg/pull/63157)).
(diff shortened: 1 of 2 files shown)


tyxla (approved):
LGTM 👍 

Given multiline tabs should be a rare edge-case, I think this is a good compromise.

PR #57888, opened 2024-01-16
Data: fix memoized createRegistrySelector

ellatrix:
## What?

This PR fixes `createRegistrySelector` when used in combination with a memoized selector (`createSelector`).

Thanks @jsnajdr for providing some tests in #57510.

The following is currently broken in trunk, it doesn't cache the result:

```js
createRegistrySelector( ( select ) =>
	createSelector(
		( state ) => {
			return select()...
		},
		() => [
			select()...
		]
	)
);
```

It does not fix binding to registry selectors to multiple registries, but this is already broken in trunk, even without `createSelector`.

## Why?

We have some selectors where memoization is needed. I would also like to add more registry selectors, but this flaw makes it impossible because it would mean adding unperformant selectors in important places.

https://github.com/WordPress/gutenberg/blob/cdd6cecbc5bbf5ce5f55da0301246acfcf56c426/packages/block-directory/src/store/selectors.js#L56-L65

https://github.com/WordPress/gutenberg/blob/cdd6cecbc5bbf5ce5f55da0301246acfcf56c426/packages/block-directory/src/store/selectors.js#L75-L84

https://github.com/WordPress/gutenberg/blob/cdd6cecbc5bbf5ce5f55da0301246acfcf56c426/packages/edit-site/src/store/selectors.js#L244-L261

https://github.com/WordPress/gutenberg/blob/cdd6cecbc5bbf5ce5f55da0301246acfcf56c426/packages/editor/src/store/private-selectors.js#L25-L47

https://github.com/WordPress/gutenberg/blob/cdd6cecbc5bbf5ce5f55da0301246acfcf56c426/packages/edit-widgets/src/store/selectors.js#L32-L49

And there's places where `createRegisterySelector` should be used, but currently isn't:

https://github.com/WordPress/gutenberg/blob/588fc12533fbe465bf26db1d27f78f8c003e0cd4/packages/edit-site/src/components/page-patterns/use-patterns.js#L212-L270

PR #63338, opened 2024-07-10
Widgets: Memoize 'getWidgets' store selector

Mamaduka:
## What?
PR memoizes the `getWidgets` selector and prevents it from returning different values when called with the same state.

## How?
It's now possible to memoize registry selectors - #57888.

## Testing Instructions
1. Enabled TT1 theme.
2. Navigate to Appearance -> Widgets.
3. Confirm it loads correctly.
4. Confirm you can update widgets as before.

_The screen also has e2e tests._

Same.

## Screenshots or screencast 
CleanShot 2024-07-10 at 11 09 44

### DIFF
--- a/packages/edit-widgets/src/components/widget-areas-block-editor-provider/index.js
+++ b/packages/edit-widgets/src/components/widget-areas-block-editor-provider/index.js
@@ -30,6 +30,8 @@ const { ExperimentalBlockEditorProvider } = unlock( blockEditorPrivateApis );
 const { PatternsMenuItems } = unlock( editPatternsPrivateApis );
 const { BlockKeyboardShortcuts } = unlock( blockLibraryPrivateApis );
 
+const EMPTY_ARRAY = [];
+
 export default function WidgetAreasBlockEditorProvider( {
 	blockEditorSettings,
 	children,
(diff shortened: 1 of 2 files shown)


packages/edit-widgets/src/components/widget-areas-block-editor-provider/index.js
```diff
@@ -54,7 +56,7 @@ export default function WidgetAreasBlockEditorProvider( {
 			widgets: select( editWidgetsStore ).getWidgets(),
 			reusableBlocks: ALLOW_REUSABLE_BLOCKS
 				? getEntityRecords( 'postType', 'wp_block' )
-				: [],
+				: EMPTY_ARRAY,
```
Mamaduka, line 59: The `ALLOW_REUSABLE_BLOCKS` flag is currently false for this editor screen, and we need a fallback with a stable reference.

packages/edit-widgets/src/store/selectors.js
```diff
@@ -29,24 +29,35 @@ const EMPTY_INSERTION_POINT = {
...
-} );
+			return (
+				// Key widgets by their ID.
+				widgets?.reduce(
+					( allWidgets, widget ) => ( {
+						...allWidgets,
+						[ widget.id ]: widget,
+					} ),
+					{}
+				) || {}
```
tyxla: Pre-existing, so no need to touch it: it maybe makes more sense to use `??` than `||`.

tyxla (approved):
Looks good and tests well 👍 

Thanks!

packages/edit-widgets/src/store/selectors.js
Mamaduka: Makes and updated.

Mamaduka:
Thank you, @tyxla!

PR #63339, opened 2024-07-10
Fix Comment pagination should use a nav and an accessible name

### LINKED ISSUE #63328: Comment pagination should use a nav and an accessible name
### Description

The comment pagination blocks are navigation items and should be placed inside a `<nav>` element to be more accessible. Currently a `<div>` is used.

The navigation should have a name that helps users separate the comment navigation from the main navigation blocks,
post navigation, query navigation etc. This can be an aria label: "Comments".

The WordPress functions that are the classic theme equivalents use a `<nav>` by default, and have options both for screen reader text and aria label:
https://developer.wordpress.org/reference/functions/get_the_comments_navigation/
https://developer.wordpress.org/reference/functions/get_the_comments_pagination/

Add a comment pagination block.
Confirm that it uses a div.

_No response_

### Environment info

_No response_

Yes

Yes

dhananjaykuber:
Fixes [#63328](https://github.com/WordPress/gutenberg/issues/63328)

## What?
This PR improves the accessibility of the Comments Pagination block by using a semantic `<nav>` element with an `aria-label` and screen reader text.

## Why?
The current implementation uses a `<div>` which lacks semantic meaning for navigation. This change aligns the block with accessibility best practices and matches the behavior of classic theme pagination functions.

## How?
- Updated `edit.js` to render a `<nav>` instead of a `<div>`
- Modified `index.php` to use `<nav>` in the server-side render function
- Added `aria-label="Comments"` to the navigation element
- Included visually hidden text for screen readers
- Ensured the 'comments-pagination' class is consistently applied

## Testing Instructions
1. Add a Comments block to a post or page
2. Enable comment pagination in the Discussion Settings if not already enabled
3. Add the Comments Pagination block inside the Comments block
4. Check the rendered output in both the editor and the front-end

## Screenshots or screencast 
Screenshot 2024-07-10 at 12 53 25 PM

### DIFF
--- a/packages/block-library/src/comments-pagination/edit.js
+++ b/packages/block-library/src/comments-pagination/edit.js
@@ -83,7 +83,12 @@ export default function QueryPaginationEdit( {
 					</PanelBody>
 				</InspectorControls>
 			) }
-			<div { ...innerBlocksProps } />
+			<nav { ...innerBlocksProps } aria-label={ __( 'Comments' ) }>
+				<span className="screen-reader-text">
+					{ __( 'Comments navigation' ) }
+				</span>
+				{ innerBlocksProps.children }
+			</nav>
 		</>
 	);
 }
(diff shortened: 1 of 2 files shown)


youknowriad:
Looks like this was fixed in #70730 thanks for your contribution.

Issue #44357, opened 2022-09-22
When dragging the resize handles, snap to wide / full widths


jameskoster:
For blocks that have resize handles for width (e.g. Image), it would be nice if the block 'snapped' to predefined widths like wide and full. Perhaps the different widths could appear as guide lines?

<img width="1520" alt="Grid" src="https://user-images.githubusercontent.com/1204802/203553939-5d689c42-73d5-43b4-87e1-1073ce660444.png">

<img width="7440" alt="Only vertical snap areas, tooltip if you pause" src="https://user-images.githubusercontent.com/1204802/203310364-149d7c3a-b693-4da5-8be2-7e85550700b0.png">

Then as you approach a guide line, the width snaps to fill. If you pause for a moment, or as you near a particular snap point, a tooltip appears with the name of the snap point:

<img width="1520" alt="Snap tooltip" src="https://user-images.githubusercontent.com/1204802/203554065-1dcda71c-7077-4a69-bd86-90dc75406ce8.png">

---

Issue updated Nov. 23.

<details><summary>Initial proposal</summary>

**Second iteration:**

![guides](https://user-images.githubusercontent.com/846565/195097355-44fa4485-ecab-4c6a-affd-4924f8bde438.gif)

Then as you approach a guide line, the width snaps to fill. I appreciate this is a bit awkward for wide and full widths as the layout will jump quite a bit, but it's hard to know just how impactful that will be without an interactive prototype (which I cannot make in Figma). Here's the closest approximation I can come up with:

![snapping](https://user-images.githubusercontent.com/846565/195098498-03f9c32d-8ac6-42dd-9be8-c641c3f61a7a.gif)

If the above approach is too technically challenging, a smaller step could be to show snap-zones in a style similar to that of dropzones or the sibling inserter:

<img width="1600" alt="Flow" src="https://user-images.githubusercontent.com/1204802/193565044-d165ca88-d1f1-4a0f-81ab-0edabe2e1cb2.png">

**First iteration:**

<img width="748" alt="Screenshot 2022-09-22 at 10 41 43" src="https://user-images.githubusercontent.com/846565/191714541-1b47e83a-1764-4438-9aa2-752ddd4ae522.png">

For blocks that have resize handles for width (e.g. Image), it would be nice if the block 'snapped' to predefined widths like wide and full.

If the above approach is too technically challenging, a smaller step could be to show snap-zones in a style similar to that of dropzones or the sibling inserter:

[...]

Issue #12168, opened 2018-11-21
Image with size breaks out of editor block


johngodley:
I'm not sure if this is a bug or expected behaviour, but reporting it as a bug anyway.

If you add an image to a post and set the size of that image to anything other than the default, it breaks out of the block. The image displays as expected on the front end (sized accordingly)

Here's an image that was dropped into the editor:

<img width="635" alt="edit_post_ _wordpress_latest_ _wordpress" src="https://user-images.githubusercontent.com/1277682/48836123-c662f580-ed79-11e8-9954-098feea35b52.png">

If I open the block options and change the dimensions:

<img width="281" alt="edit_post_ _wordpress_latest_ _wordpress" src="https://user-images.githubusercontent.com/1277682/48836166-e2ff2d80-ed79-11e8-9b34-dacc4906f2fb.png">

The image has now broken outside of the block:

<img width="831" alt="edit_post_ _wordpress_latest_ _wordpress" src="https://user-images.githubusercontent.com/1277682/48836188-f4483a00-ed79-11e8-9ab2-fb01b89f3200.png">

Pressing 'reset' on the image dimensions brings the image back inside the block.

**Expected behavior**
The image to stay inside the block

Mac & Chrome

mrwweb:
Ran into this same issue in [#11159](https://github.com/WordPress/gutenberg/issues/11159#issuecomment-440158013). Unsurprisingly it also happens when the block is in a column.

mrwweb:
Just to add some more context, this is also true when resizing via the pixel-size editor and may also be true when using the percentage controls if the image is constrained when first inserted but the percentage size of the image is greater than 100%.

designsimply:
Closed #13220 as a duplicate and am including the screenshot here for reference:

<img width="1066" alt="screen shot 2019-01-07 at 09 55 17" src="https://user-images.githubusercontent.com/6073772/50759146-6e6a4a80-1264-11e9-961c-6746b9dcd4a8.png">

/hat tip @hedgefield.

designsimply:
Closed https://github.com/WordPress/gutenberg/issues/13944 as a duplicate and adding their testing steps and screenshots for reference:

> Steps to reproduce the behavior:
> 
> 1. Add an Image
> 1. Choose Image Site "Full Size"
> 1. Click the "100%" Button in the Image Dimensions section.
> 1. See error
>
> Screenshot showing the Error:
> 
[...]

PR #63341, opened 2024-07-10
Limit the max width of image to its container size

### LINKED ISSUE #63326: Image block can be resized beyond the bounds of the column that it's in
### Description

When you have an Image block within a Columns block, you can currently use the resize handles on the image to resize it beyond the bounds of the column, which results in a pretty broken looking layout. The resize bounds should be limited to the column.

1. Insert a new post
2. Add a Columns block with two columns
3. Add an image to each column
4. Select one of the images and resize it vertically downwards
5. The image overflows the column, but shouldn't

https://github.com/WordPress/gutenberg/assets/612155/cc5e05dd-79b7-46aa-b2ae-37c77acc3ca3

### Environment info

Latest Gutenberg + Core trunk. Firefox. macOS.

Yes

Yes

kevin940726:
## What?

Attempt to fix https://github.com/WordPress/gutenberg/issues/63326.

Limit the max width of image to its container size.

## Why?

See https://github.com/WordPress/gutenberg/issues/63326. Images in columns or other container can be resized to be bigger than its container, which is not expected.

## How?

Limit the `maxWidth` of the image to its container's size. Not sure about this approach and potential side effects though.

## Testing Instructions

1. Add a columns block.
2. Add an image to one of its column.
3. Try resizing the image to be bigger than the column.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/7753001/61c803e0-ec9b-4387-af4d-4e2db77073b1

### DIFF
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -108,6 +108,7 @@ export default function Image( {
 	clientId,
 	blockEditingMode,
 	parentLayoutType,
+	containerWidth,
 } ) {
 	const {
 		url = '',
@@ -923,6 +924,7 @@ export default function Image( {
 		// @todo It would be good to revisit this once a content-width variable
 		// becomes available.
 		const maxWidthBuffer = maxWidth * 2.5;
+		const maxContentWidth = containerWidth || maxWidthBuffer;
 
 		let showRightHandle = false;
 		let showLeftHandle = false;
@@ -968,9 +970,9 @@ export default function Image( {
 				} }
 				showHandle={ isSingleSelected }
 				minWidth={ minWidth }
-				maxWidth={ maxWidthBuffer }
+				maxWidth={ maxContentWidth }
 				minHeight={ minHeight }
-				maxHeight={ maxWidthBuffer / ratio }
+				maxHeight={ maxContentWidth / ratio }
 				lockAspectRatio={ ratio }
 				enable={ {
 					top: false,
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -984,11 +984,19 @@ export default function Image( {
 				onResizeStop={ ( event, direction, elt ) => {
 					onResizeStop();
 
-					let resizedWidth = `${ elt.offsetWidth }px`;
-					// Set width to 'auto' if the resized width is close to the max-content width.
-					if ( Math.abs( elt.offsetWidth - maxContentWidth ) < 10 ) {
-						resizedWidth = 'auto';
-						elt.style.width = 'auto';
+					// Clear hardcoded width if the resized width is close to the max-content width.
+					if (
+						// Only do this if the image is bigger than the container to prevent it from being squished.
+						// TODO: Remove this check if the image support setting 100% width.
+						naturalWidth >= maxContentWidth &&
+						Math.abs( elt.offsetWidth - maxContentWidth ) < 10
+					) {
+						setAttributes( {
+							width: undefined,
+							height: undefined,
+							aspectRatio: undefined,
+						} );
+						return;
 					}
 
 					// Since the aspect ratio is locked when resizing, we can
@@ -996,7 +1004,7 @@ export default function Image( {
 					// height in CSS to prevent stretching when the max-width
 					// is reached.
 					setAttributes( {
-						width: resizedWidth,
+						width: `${ elt.offsetWidth }px`,
 						height: 'auto',
 						aspectRatio:
 							ratio === naturalRatio
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -983,12 +983,20 @@ export default function Image( {
 				onResizeStart={ onResizeStart }
 				onResizeStop={ ( event, direction, elt ) => {
 					onResizeStop();
+
+					let resizedWidth = `${ elt.offsetWidth }px`;
+					// Set width to 'auto' if the resized width is close to the max-content width.
+					if ( Math.abs( elt.offsetWidth - maxContentWidth ) < 10 ) {
+						resizedWidth = 'auto';
+						elt.style.width = 'auto';
+					}
+
 					// Since the aspect ratio is locked when resizing, we can
 					// use the width of the resized element to calculate the
 					// height in CSS to prevent stretching when the max-width
 					// is reached.
 					setAttributes( {
-						width: `${ elt.offsetWidth }px`,
+						width: resizedWidth,
 						height: 'auto',
 						aspectRatio:
 							ratio === naturalRatio
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -994,7 +994,6 @@ export default function Image( {
 						setAttributes( {
 							width: undefined,
 							height: undefined,
-							aspectRatio: undefined,
 						} );
 						return;
 					}
--- a/packages/block-library/src/image/edit.js
+++ b/packages/block-library/src/image/edit.js
@@ -369,7 +369,6 @@ export function ImageEdit( {
 
 	return (
 		<>
-			{ contentResizeListener }
 			<figure { ...blockProps }>
 				<Image
 					temporaryURL={ temporaryURL }
@@ -400,6 +399,11 @@ export function ImageEdit( {
 					disableMediaButtons={ temporaryURL || url }
 				/>
 			</figure>
+			{
+				// The listener cannot be placed as the first element as it will break the in-between inserter.
+				// See https://github.com/WordPress/gutenberg/blob/71134165868298fc15e22896d0c28b41b3755ff7/packages/block-editor/src/components/block-list/use-in-between-inserter.js#L120
+				contentResizeListener
+			}
 		</>
 	);
 }
--- a/packages/block-library/src/image/edit.js
+++ b/packages/block-library/src/image/edit.js
@@ -23,6 +23,7 @@ import { useEffect, useRef, useState } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
 import { image as icon, plugins as pluginsIcon } from '@wordpress/icons';
 import { store as noticesStore } from '@wordpress/notices';
+import { useResizeObserver } from '@wordpress/compose';
 
 /**
  * Internal dependencies
@@ -108,6 +109,9 @@ export function ImageEdit( {
 		metadata,
 	} = attributes;
 	const [ temporaryURL, setTemporaryURL ] = useState( attributes.blob );
+
+	const [ contentResizeListener, { width: containerWidth } ] =
+		useResizeObserver();
 
 	const altRef = useRef();
 	useEffect( () => {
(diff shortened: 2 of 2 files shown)


talldan:
It seems close and I like the solution, but I think the main issue is that the image resized to the max bounds loses its responsive because it's set to an exact pixel value instead of 'auto'. I think it'd be better to set it back to 'auto' like an image is when first uploaded to an image block.

Some steps to highlight the issue:
1. Add a columns block with two columns
2. Add an image block to the first column and upload an image
3. Go back to the columns block and set the image to wide width, observe how it's width adjusts responsively because it's set to 'auto'.

Now compare that behavior to this:
1. Add a columns block with two columns
2. Add an image block to the first column and upload an image
3. Resize the image to a smaller and then back to the max width
4. Go back to the columns block and set the image to wide width, observe how it's now limited to the max width the column was previously.

talldan:
Another observation is that left/right aligned images can no longer be resized to the same degree they were before (it seems to be limited to content width).

That's an interesting one, I'm not sure what the preferred behavior is.

This is also a little bit theme dependent, some themes like TT4 show aligned blocks offset from the content which would allow them to be resized further, while others don't. That makes it tricky to know dynamically what the right limit is.

noisysocks:
Nice work. Just chiming in to say that I tested this and it fixes the issue 😀

andrewserong:
> Nice work. Just chiming in to say that I tested this and it fixes the issue 😀

+1, nice progress here, it feels much better to use to me with the constrained maximum.

> I think it'd be better to set it back to 'auto' like an image is when first uploaded to an image block.

Great idea, that sounds like a decent way to handle it. Would that be a matter of setting to `auto` in `onResizeStop` if we're currently at the max size?

Also just pinging @Mamaduka for visibility on this one since he commented on the linked issue 🙂

Mamaduka:
Nice. I've not thought of using `useResizeObserver` for this particular case.

kevin940726:
> the main issue is that the image resized to the max bounds loses its responsive because it's set to an exact pixel value instead of 'auto'. I think it'd be better to set it back to 'auto' like an image is when first uploaded to an image block.

I've been thinking about this. I think it's hard to distinguish the user's intention just by observing resizing. Maybe the user wants it to be a certain exact pixel wide. I think in most cases the user wants it to fit the max width though 😅. I wonder if some kind of "snapping" makes it clearer. Perhaps @WordPress/gutenberg-design has some ideas? 🙇 

> Would that be a matter of setting to `auto` in `onResizeStop` if we're currently at the max size?

Another thing to note is that I think it also depends on the original image's size. If the image width is smaller than the max width, then resizing it to the max width (the user intention) should probably mean "width: 100%" rather than "width: auto"? 🤔

andrewserong:
> Maybe the user wants it to be a certain exact pixel wide. I think in most cases the user wants it to fit the max width though 😅. I wonder if some kind of "snapping" makes it clearer.

Good point. In this case, if folks want to set an exact pixel size, they can still do that in the sidebar. I like the idea of some kind of snapping at the max width that sets it to `auto`, though. It reminds me a bit of the older issue to snap to wide / full widths (https://github.com/WordPress/gutenberg/issues/44357). Possibly a nice-to-have, rather than a must have?

>  If the image width is smaller than the max width, then resizing it to the max width (the user intention) should probably mean "width: 100%" rather than "width: auto"? 🤔

Also a good point! The idea of defaulting to `auto` if the container is smaller than the image, and `100%` if the container is larger than the image sounds worth exploring to me!

talldan:
> Another thing to note is that I think it also depends on the original image's size. If the image width is smaller than the max width, then resizing it to the max width (the user intention) should probably mean "width: 100%" rather than "width: auto"?

Possibly, I haven't checked the CSS. When I say 'auto', it's what the UI shows in the width field. Sorry if that wasn't clear.

jasmussen:
> I've been thinking about this. I think it's hard to distinguish the user's intention just by observing resizing. Maybe the user wants it to be a certain exact pixel wide. I think in most cases the user wants it to fit the max width though 😅. I wonder if some kind of "snapping" makes it clearer. 

I'd echo fixed width vs. "unbounded" is one of the primary headaches to solve as part of this. And if we can do snapping to unbounded ("Auto" as Dan refers to), that would be great. We've had separate explorations into snapping into wide and full-wide alignments too, it's very difficult to build. But it would definitely be valuable, and in this case would account for snapping to _content-width_.

talldan:
I found this old issue that might be closed by this PR - https://github.com/WordPress/gutenberg/issues/12168.

kevin940726:
Should we include the snapping feature in this PR, or could it be implemented separately? Is the implementation in this PR without snapping worth merging?

talldan:
Is there a basic version that could be attempted? Something like this pseudocode
```js
onResizeStop( { actualWidth } ) {
  if ( Math.abs( containerWidth - actualWidth ) < 10 ) { 
    setWidth( '100%' );
  } else {
    setWidth( actualWidth )
  }
}
```

kevin940726:
> Is there a basic version that could be attempted? Something like this pseudocode
> 
> ```js
> onResizeStop( { actualWidth } ) {
>   if ( Math.abs( containerWidth - actualWidth ) < 10 ) { 
>     setWidth( '100%' );
>   } else {
>     setWidth( actualWidth )
>   }
> }
> ```

Committed in [858d8a2](https://github.com/WordPress/gutenberg/pull/63341/commits/858d8a2d1c44a8b36744a6657586dc0d4e10fdf3)!

packages/block-library/src/image/image.js
```diff
@@ -960,12 +962,20 @@ export default function Image( {
 				onResizeStart={ onResizeStart }
 				onResizeStop={ ( event, direction, elt ) => {
 					onResizeStop();
+
+					let resizedWidth = `${ elt.offsetWidth }px`;
+					// Set width to 'auto' if the resized width is close to the max-content width.
+					if ( Math.abs( elt.offsetWidth - maxContentWidth ) < 10 ) {
+						resizedWidth = 'auto';
+						elt.style.width = 'auto';
```
andrewserong: The default state of an image when it's inserted is that the controls display `auto`, but there's no explicit style output. Instead of injecting `width: 'auto'` and `height: auto`, could we instead treat this as a kind of "reset" and an early return, with something like:

```js
setAttributes( {
  width: undefined,
  height: undefined,
  aspectRatio: undefined, // (OR possibly re-use the ratio === naturalRatio logic below)
} );
return;
```

If we go with what we currently have, I notice that the markup winds up including the `auto` rules:

```
<!-- wp:image {"id":4711,"width":"auto","height":"auto","sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large is-resized"><img src="http://a-toule-site.local/wp-content/uploads/2024/01/IMG_3906-768x1024.jpg" alt="" class="wp-image-4711" style="width:auto;height:auto"/></figure>
```

Whereas an image that is freshly inserted doesn't have any of those rules applied:

```
<!-- wp:image {"id":4711,"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large"><img src="http://a-toule-site.local/wp-content/uploads/2024/01/IMG_3906-768x1024.jpg" alt="" class="wp-image-4711"/></figure>
```

andrewserong (commented):
Thanks for the update @kevin940726! In general this is testing pretty well for me, but I notice that explicitly setting to `auto` results in a bug that stretches images upon resizing the viewport. I've left a comment, but I'm wondering if instead of setting `auto` explicitly, we can treat drags within the `10px` threshold of the container as a reset instead. Would that work?

Here's a quick demo of what happens when both width and height are explicitly set to `auto`: (seems fine at first until we resize the viewport)

https://github.com/user-attachments/assets/8db05a61-af6f-4544-a4f0-3eef0dd1cb5e

packages/block-library/src/image/image.js
kevin940726: Good idea unsetting to `undefined`!

I don't know if it's possible to set an image to 100% of the container's width, so I added a check to fallback to explicit pixel values if the image is smaller than the container.

andrewserong: Thanks Kai, this is feeling really close to me!

> I don't know if it's possible to set an image to 100% of the container's width, so I added a check to fallback to explicit pixel values if the image is smaller than the container.

That sounds good to me for now — we could always look at it in a follow-up, potentially alongside an explicit way of locking to `full width` alignment as I imagine the two concepts could be quite related. E.g. when would we want to lock to "auto" vs setting the alignment 🤔

For now, though, I think unsetting and only doing so if the image is smaller than the container is a good solution 👍

The one issue I ran into in retesting is that I think I was wrong about explicitly setting `aspectRatio` to `undefined`. It looks like we might need to either not update the `aspectRatio` property at all in this call, or reuse the logic from the `setAttributes` call below:

```
						aspectRatio:
							ratio === naturalRatio
								? undefined
								: String( ratio ),
```

Otherwise, when adjusting the size within the threshold of the container area, we lose an aspect ratio if one has been set:

https://github.com/user-attachments/assets/6574397e-c559-4bce-a5e6-373403eb89ef

What do you think? From my perspective, once that bug's resolved I think this PR will be in good shape to switch over to ready for review and get wider testing / feedback from designers, etc.

kevin940726: I think that makes sense! Maybe we could just not override the `aspectRatio` in this case? I pushed an update in [c1635bd](https://github.com/WordPress/gutenberg/pull/63341/commits/c1635bd75a2821e339691f9470a3b407aa1f7039).

andrewserong: Nice, that's testing well for me now! (Also while re-testing, I think avoiding explicitly setting `auto` for width and height helps avoid potential conflicts with aspect ratios, too). Let's see what some designers think 🙂

Before landing, we'll likely want to manually test the controls in a variety of browsers and themes (including Classic ones), but overall this is feeling pretty solid to me!

andrewserong (approved):
This is testing great for me across browsers (Chrome, Safari, FF, Edge) and in a variety of scenarios with various blocks. Overall I think this is a really big improvement. There are still some outstanding bugs or things that would be good to polish in separate PRs, but for now I think the overall improvement here makes it very worthwhile 👍

I ran into one tricky issue (not a blocker) with Classic themes, with image blocks that are added to the root of the document — I think I've identified the issue, but I'm not 100% sure what the right fix might be.

In block themes, the max width of blocks within the document is determined by layout rules that apply to any element injected in the document. So the `div` added via the `contentResizeListener` gets the correct width assigned to it:

image

In the above screenshot, the listener `div` correctly gets the `max-width` of `620px`, so when we go to drag the handle on the image block, it can't go past `620px`, which is what we expected.

However on Classic themes like Twenty Twenty One, blocks get a max width based on the classname of `wp-block`, and the `div` used by the resize listener doesn't have that classname. As a result, the max width of the listener is much larger than expected:

image

This means that in Classic themes when we go to drag beyond the container it still lets us do it:

https://github.com/user-attachments/assets/ffd0f894-4ca3-4668-a778-a77189e32ceb

Whereas in block themes, the listener is the correct size, so it works just as we'd expect:

https://github.com/user-attachments/assets/4f59d965-687c-42e6-9fd3-60d6e09d289c

I found that in Classic themes, I could hack a fix in Dev Tools by inserting a `wp-block` class name into the `div` output by the content resize listener, which appears to prove out the above (though I'm not sure if that would be the right fix 🤔):

image

Longer-term, I wonder if it would be worth exploring whether we can let the resize observer accept a custom classname to output to the resize listener? If it supported that, then we could (potentially) pass it `wp-block` in order to fix this particular case. We'd then output it somewhere around here:

https://github.com/WordPress/gutenberg/blob/ca2ee4d6fa8c7044cd1d3045c0985a77659f44fa/packages/compose/src/hooks/use-resize-observer/index.tsx#L345-L359

In any case, I think that too could be a separate PR to look into. What do you think?

For now this PR resolves the situation of resizing within a Column block as reported in #63326, and also resolves dragging within the editor canvas in block themes, so I reckon the above could be looked at separately.

Thanks for reading this longer-than-expected brain dump! 😄

LGTM! 🚀

kevin940726:
> If it supported that, then we could (potentially) pass it wp-block in order to fix this particular case.

That's an interesting idea! I think we can do that today with `cloneBlock`, but I'm concerned about the side effects of adding the `wp-block` class name to an element that isn't a block 🤔. Either way, I think it's worth experimenting in another PR 👍.

PR #63343, opened 2024-07-10
CustomSelectControlV2: animate select popover appearance

ciampo:
## What?

Part of #55023

Animate the select popover appearance in the new version of `CustomSelectControl`

## Why?

To make UI interactions feel more polished and to uniform the component with the direction taken in other components.

## How?

Using CSS animations to add a fade + slide down animation when the select popover appears

> [!NOTE]
> For now, the slide direction in the animation will always be downwards. I tried to detect when `Ariakit` "flips" the popover from bottom to top for lack of space, so that I could apply a "slide down" animation instead, but I couldn't extract that information from the ariakit store (I tried the [same technique](https://github.com/WordPress/gutenberg/blob/039bd9081abeb7ce93d7a57e56ca0abfbc6b7937/packages/components/src/dropdown-menu-v2/index.tsx#L250-L253) used in the `DropdownMenuV2` component.

## Testing Instructions

1. Open the `CustomSelectControlV2` Storybook default or legacy examples
2. Open and close the select popover, make sure the animation works 
3. Using browser dev tools (or OS settings), enable the preference for `reduced motion`. Open and close the select popover again, and make sure that it doesn't animate.

## Screenshots or screencast 

|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/60801d78-125a-4340-ab21-7e0b98443fde" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/d986678c-d3a2-4510-b9ba-1d2de919e8fb" /> |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -44,6 +44,7 @@
 -   `CustomSelectControlV2`: keep item checkmark top aligned. ([#63230](https://github.com/WordPress/gutenberg/pull/63230))
 -   `CustomSelectControlV2`: keep legacy arrow down behavior only for legacy wrapper. ([#62919](https://github.com/WordPress/gutenberg/pull/62919))
 -   `CustomSelectControlV2`: fix trigger button font size. ([#63131](https://github.com/WordPress/gutenberg/pull/63131))
+-   `CustomSelectControlV2`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
 -   `CustomSelectControlV2`: fix labelling with a visually hidden label. ([#63137](https://github.com/WordPress/gutenberg/pull/63137))
 -   `CustomSelectControlV2`: allow checkmark wrapper to collapse when not shown. ([#63229](https://github.com/WordPress/gutenberg/pull/63229))
 -   Extract `TimeInput` component from `TimePicker` ([#60613](https://github.com/WordPress/gutenberg/pull/60613)).
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Internal
+
+-   `CustomSelectControlV2`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
+
 ## 28.3.0 (2024-07-10)
 
 ### Enhancements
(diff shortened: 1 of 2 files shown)


jasmussen:
Thanks for sweating these details.

I would expect since this is an element that opens downwards, that the animation reflect that. Right now it appears heavy on the opacity fade, and looks like it animates upwards.

tyxla:
> I would expect since this is an element that opens downwards, that the animation reflect that. Right now it appears heavy on the opacity fade, and looks like it animates upwards.

Perhaps if we delay the `transform` part of the animation a bit?

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -98,6 +105,14 @@ export const Select = styled( Ariakit.Select, {
 	`
 );
 
+const slideUpAndFade = keyframes( {
+	'0%': {
+		opacity: 0,
+		transform: `translateY(${ ANIMATION_PARAMS.SLIDE_AMOUNT })`,
```
tyxla: Don't we mean to start from minus that amount so we preserve the downward direction?

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -13,6 +13,13 @@ import { chevronIconSize } from '../select-control/styles/select-control-styles'
 import { fontSizeStyles } from '../input-control/styles/input-control-styles';
 import type { CustomSelectButtonSize } from './types';
 
+// TODO: extract to common utils and apply to relevant components
+const ANIMATION_PARAMS = {
+	SLIDE_AMOUNT: '2px',
```
tyxla, line 18: Should this be increased in order to make it more obvious that the select options are dropping down?

ciampo:
Hey @jasmussen 

> I would expect since this is an element that opens downwards, that the animation reflect that

Updated the animation to slide down, instead of up.

https://github.com/WordPress/gutenberg/assets/1083581/d986678c-d3a2-4510-b9ba-1d2de919e8fb

> Right now it appears heavy on the opacity fade

I'm not sure I understand what tweak you're suggesting to improve the fade.

---

Also, the opening animation is copied from [`DropdownMenuV2`](https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-dropdownmenu-v2--docs). Apart from having to refactor common variables / animations in one place (which we'll do at a later point), should I also update `DropdownMenuV2` with the resulting feedback received in this PR?

packages/components/src/custom-select-control-v2/styles.ts
ciampo: Done, see https://github.com/WordPress/gutenberg/pull/63343#issuecomment-2220162250

jasmussen:
That feels substantially better! Outside of exploring actual "roll out" animations to animate the _height_ of the dropdown from zero to its max-height, this one feels good to go as a first step.

ciampo:
> this one feels good to go as a first step.

Thank you! I'll merge as soon as this PR gets an approval.

> Outside of exploring actual "roll out" animations to animate the height of the dropdown from zero to its max-height,

That should be possible soon, when the `transition-behavior` and the `@starting-style` CSS features land in all major browsers (example: https://css-tricks.com/transitioning-to-auto-height/) instead of using [CSS grid hacks](https://css-tricks.com/css-grid-can-do-auto-height-transitions/)

packages/components/CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 -   `CustomSelectControlV2`: keep item checkmark top aligned. ([#63230](https://github.com/WordPress/gutenberg/pull/63230))
 -   `CustomSelectControlV2`: keep legacy arrow down behavior only for legacy wrapper. ([#62919](https://github.com/WordPress/gutenberg/pull/62919))
 -   `CustomSelectControlV2`: fix trigger button font size. ([#63131](https://github.com/WordPress/gutenberg/pull/63131))
+-   `CustomSelectControlV2`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
```
tyxla: After rebasing to the latest `trunk`, we need to move this up under Unreleased.

packages/components/src/custom-select-control-v2/styles.ts
tyxla: Now that you changed it to starting from `- SLIDE_AMOUNT`, it looks good actually 👍 Nice touch!

tyxla (approved):
LGTM, nice addition 👍 👏 🚀

packages/components/CHANGELOG.md
ciampo: Done!

PR #63346, opened 2024-07-10
Block Directory: Memoize store selectors

Mamaduka:
## What?
PR adds memoization to the `getNewBlockTypes` and `getUnusedBlockTypes` block directory store selectors.

## Why?
This is a follow-up to #57888. It prevents selectors from returning different values when called with the same state.

There were no `useSelect` warnings in the editor because of how these selectors are called in components. The `mapSelect` directly returns values, which satisfies shallow equality.

https://github.com/WordPress/gutenberg/blob/4f706f1777d05566f66b80a82560dd7c498942dc/packages/block-directory/src/plugins/installed-blocks-pre-publish-panel/index.js#L19-L22

## Testing Instructions
The feature has e2e test coverage.

1. Create a post.
2. Install a block plugin from the directory.
3. Try publishing a post.
4. Confirm that the newly installed block is listed in the pre-publish sidebar.

Same.

## Screenshots or screencast 
CleanShot 2024-07-10 at 13 29 24

### DIFF
--- a/packages/block-directory/src/store/selectors.js
+++ b/packages/block-directory/src/store/selectors.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { createRegistrySelector } from '@wordpress/data';
+import { createSelector, createRegistrySelector } from '@wordpress/data';
 import { store as blockEditorStore } from '@wordpress/block-editor';
 
 /**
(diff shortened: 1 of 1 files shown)


ellatrix (approved):
Yeah, I think they were not memoized because it was previously impossible for registry selectors.

tyxla:
Nice 👍

PR #63348, opened 2024-07-10
Image lightbox: move image data from context to state

madhusudhand:
## What?

This change moves the image data from context to state.
The data is set in the set against a `unique_id` created for the image.

After this change, image data of all images can be accessed directly from the 'core/image` state.

## Why?

This helps to control the image lightbox from containing blocks such as `gallery`.
In a future change, gallery can will hold list of `image_ids` in its context, and sets the image block with current `image_id` to show lightbox as a carousel.

Gallery block cannot access context data of all images.
Alternatively, gallery could have an array and each image block may push its context data to gallery array `on image init`, but it make the same data to be present in two places.

Hence the image metadata is moved to its state.

## Testing Instructions

* Enable lightbox for image (Click to expand in global styles)
* verify lightbox functionality in image and gallery blocks.
* It should work the same as before without any issues.

### DIFF
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -187,7 +187,7 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	$figure_styles      = $p->get_attribute( 'style' );
 
 	// Create unique id and set the image metadata in the state.
-	$unique_image_id = substr( md5( $img_uploaded_src ), 0, 10 );
+	$unique_image_id = uniqid();
 
 	wp_interactivity_state(
 		'core/image',
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -187,7 +187,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	$figure_styles      = $p->get_attribute( 'style' );
 
 	// Create unique id and set the image metadata in the state.
-	$unique_image_id = wp_unique_id( 'image-' );
+	$unique_image_id = substr( md5( $img_uploaded_src ), 0, 10 );
+
 	wp_interactivity_state(
 		'core/image',
 		array(
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -185,22 +185,36 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	$p->seek( 'figure' );
 	$figure_class_names = $p->get_attribute( 'class' );
 	$figure_styles      = $p->get_attribute( 'style' );
+
+	// Create unique id and set the image metadata in the state.
+	$unique_image_id = wp_unique_id( 'image-' );
+	wp_interactivity_state(
+		'core/image',
+		array(
+			'metadata' => array(
+				$unique_image_id => array(
+					'uploadedSrc'      => $img_uploaded_src,
+					'figureClassNames' => $figure_class_names,
+					'figureStyles'     => $figure_styles,
+					'imgClassNames'    => $img_class_names,
+					'imgStyles'        => $img_styles,
+					'targetWidth'      => $img_width,
+					'targetHeight'     => $img_height,
+					'scaleAttr'        => $block['attrs']['scale'] ?? false,
+					'ariaLabel'        => $aria_label,
+					'alt'              => $alt,
+				),
+			),
+		)
+	);
+
 	$p->add_class( 'wp-lightbox-container' );
 	$p->set_attribute( 'data-wp-interactive', 'core/image' );
 	$p->set_attribute(
 		'data-wp-context',
 		wp_json_encode(
 			array(
-				'uploadedSrc'      => $img_uploaded_src,
-				'figureClassNames' => $figure_class_names,
-				'figureStyles'     => $figure_styles,
-				'imgClassNames'    => $img_class_names,
-				'imgStyles'        => $img_styles,
-				'targetWidth'      => $img_width,
-				'targetHeight'     => $img_height,
-				'scaleAttr'        => $block['attrs']['scale'] ?? false,
-				'ariaLabel'        => $aria_label,
-				'alt'              => $alt,
+				'imageId' => $unique_image_id,
 			),
 			JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
 		)
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -246,8 +246,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 			aria-label="' . esc_attr( $aria_label ) . '"
 			data-wp-init="callbacks.initTriggerButton"
 			data-wp-on-async--click="actions.showLightbox"
-			data-wp-style--right="context.imageButtonRight"
-			data-wp-style--top="context.imageButtonTop"
+			data-wp-style--right="state.imageButtonRight"
+			data-wp-style--top="state.imageButtonTop"
 		>
 			<svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" fill="none" viewBox="0 0 12 12">
 				<path fill="#fff" d="M2 0a2 2 0 0 0-2 2v2h1.5V2a.5.5 0 0 1 .5-.5h2V0H2Zm2 10.5H2a.5.5 0 0 1-.5-.5V8H0v2a2 2 0 0 0 2 2h2v-1.5ZM8 12v-1.5h2a.5.5 0 0 0 .5-.5V8H12v2a2 2 0 0 1-2 2H8Zm2-12a2 2 0 0 1 2 2v2h-1.5V2a.5.5 0 0 0-.5-.5H8V0h2Z" />
--- a/packages/block-library/src/image/view.js
+++ b/packages/block-library/src/image/view.js
@@ -324,13 +324,11 @@ const { state, actions, callbacks } = store(
 			`;
 			},
 			setButtonStyles() {
-				const ctx = getContext();
+				const { imageId } = getContext();
 				const { ref } = getElement();
-				state.metadata[ ctx.imageId ] = {
-					...state.metadata[ ctx.imageId ],
-					imageRef: ref,
-					currentSrc: ref.currentSrc,
-				};
+
+				state.metadata[ imageId ].imageRef = ref;
+				state.metadata[ imageId ].currentSrc = ref.currentSrc;
 
 				const {
 					naturalWidth,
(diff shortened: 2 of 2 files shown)


packages/block-library/src/image/view.js
```diff
@@ -72,16 +73,22 @@ const { state, actions, callbacks } = store(
...
-				// Moves the information of the expaned image to the state.
-				ctx.currentSrc = ctx.imageRef.currentSrc;
+				// Sets the information of the expanded image in the state.
 				imageRef = ctx.imageRef;
 				buttonRef = ctx.buttonRef;
-				state.currentImage = ctx;
+				state.currentImage = {
+					...state.metadata[ ctx.imageId ],
+					// imageRef: ctx.imageRef,
+					// buttonRef: ctx.buttonRef,
```
madhusudhand: These variables to be enabled when #63290 is merged.

packages/block-library/src/image/index.php
```diff
@@ -185,22 +185,36 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	$p->seek( 'figure' );
 	$figure_class_names = $p->get_attribute( 'class' );
 	$figure_styles      = $p->get_attribute( 'style' );
+
+	// create unique id and set the metadata.
```
mikachan:
```suggestion
	// Create unique id and set the metadata.
```

mikachan (commented):
This is more of a smoke-test result, but I gave this a test and I can see the lightbox is behaving the same for both the Image block and the Gallery block. I can't see any errors or issues 🙌

Looks like there's a small coding standards issue in image/index.php, but other than that, the code reads well to me 👍 I can see the changes to `data-wp-context` and `wp_interactivity_state` are moving the image data from context to state, which looks good to me. With the caveat that I'm not too familiar with `wp_interactivity_state` yet!

packages/block-library/src/image/index.php
```diff
@@ -185,22 +185,36 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	$p->seek( 'figure' );
 	$figure_class_names = $p->get_attribute( 'class' );
 	$figure_styles      = $p->get_attribute( 'style' );
+
+	// Create unique id and set the image metadata in the state.
+	$unique_image_id = wp_unique_id( 'image-' );
```
luisherranz: Let's use a hash to prevent potential collisions in client-side navigation.

packages/block-library/src/image/view.js
```diff
@@ -23,6 +23,7 @@ const { state, actions, callbacks } = store(
 	'core/image',
 	{
 		state: {
+			metadata: {},
```
luisherranz: It's not necessary to initialize `metadata` on the client since it's being initialized on the server.

packages/block-library/src/image/view.js
```diff
@@ -54,19 +55,19 @@ const { state, actions, callbacks } = store(
...
 				// closed.
 				state.scrollTopReset = document.documentElement.scrollTop;
 				state.scrollLeftReset = document.documentElement.scrollLeft;
 
-				// Moves the information of the expaned image to the state.
-				ctx.currentSrc = ctx.imageRef.currentSrc;
-				state.currentImage = ctx;
-				state.overlayEnabled = true;
+				// Sets the information of the expanded image in the state.
+				state.currentImage = state.metadata[ ctx.imageId ];
```
luisherranz: Instead of copying the object, let's just save the ID in the state and create a derived state `state.currentImage` that returns the correct object to avoid duplicating the information.

packages/block-library/src/image/view.js
```diff
@@ -325,7 +326,11 @@ const { state, actions, callbacks } = store(
 			setButtonStyles() {
 				const ctx = getContext();
 				const { ref } = getElement();
-				ctx.imageRef = ref;
+				state.metadata[ ctx.imageId ] = {
+					...state.metadata[ ctx.imageId ],
+					imageRef: ref,
+					currentSrc: ref.currentSrc,
+				};
```
luisherranz: In the Interactivity API, immutability isn't necessary. In fact, since we are using signals, it's more performant to use mutability.

packages/block-library/src/image/view.js
```diff
@@ -381,25 +389,28 @@ const { state, actions, callbacks } = store(
...
-				} else {
-					ctx.imageButtonTop = buttonOffsetTop + 16;
-					ctx.imageButtonRight = buttonOffsetRight + 16;
 				}
+
+				state.metadata[ ctx.imageId ] = {
+					...state.metadata[ ctx.imageId ],
+					imageButtonTop,
+					imageButtonRight,
+				};
```
luisherranz: Same here.

packages/block-library/src/image/view.js
```diff
@@ -411,7 +422,10 @@ const { state, actions, callbacks } = store(
 			initTriggerButton() {
 				const ctx = getContext();
 				const { ref } = getElement();
-				ctx.buttonRef = ref;
+				state.metadata[ ctx.imageId ] = {
+					...state.metadata[ ctx.imageId ],
+					buttonRef: ref,
+				};
```
luisherranz: And here.

luisherranz (dismissed):
In general, it looks good to me. Thanks Madhu for making these changes 🙂

I am going to make a few small tweaks listed below.

packages/block-library/src/image/index.php
luisherranz: Done in https://github.com/WordPress/gutenberg/pull/63348/commits/189c7f51dc2bd395ca80c3a1435ecf4b0f76a9a4 by using `substr( md5( $img_uploaded_src ), 0, 10 )`.

The probability of collisions in a post with 1000 images is 1 in 2198856, so I think this should be enough.

packages/block-library/src/image/view.js
luisherranz: Done in https://github.com/WordPress/gutenberg/pull/63348/commits/56d94872c41e57fcbb47652359d3acdedf1bbe63.

packages/block-library/src/image/view.js
luisherranz: Done in https://github.com/WordPress/gutenberg/pull/63348/commits/56d94872c41e57fcbb47652359d3acdedf1bbe63.

packages/block-library/src/image/view.js
luisherranz: Done in https://github.com/WordPress/gutenberg/pull/63348/commits/3c3c3f27ea50e8c8749b8bd2e41a47afdc598142.

packages/block-library/src/image/view.js
luisherranz: Done in https://github.com/WordPress/gutenberg/pull/63348/commits/3c3c3f27ea50e8c8749b8bd2e41a47afdc598142.

packages/block-library/src/image/view.js
luisherranz: Done in https://github.com/WordPress/gutenberg/pull/63348/commits/3c3c3f27ea50e8c8749b8bd2e41a47afdc598142.

luisherranz:
@santosguillamot could you please test this with your image set?

If everything works, feel free to dismiss my review and approve the pull request 🙂

madhusudhand:
Thanks for the review and changes @luisherranz

SantosGuillamot (changes requested):
The UX seems broken in my local tests:

In this pull request

https://github.com/user-attachments/assets/63b29751-3730-4fe9-89af-386259a73c3f

In `trunk`

https://github.com/user-attachments/assets/27be3e68-39b0-4028-be04-a9a25882193d

After taking a quick look, it seems it breaks after this commit: [Use hash instead of counter for the unique id](https://github.com/WordPress/gutenberg/pull/63348/commits/189c7f51dc2bd395ca80c3a1435ecf4b0f76a9a4). If I revert that changes, all the images seem to work as expected.

madhusudhand:
> After taking a quick look, it seems it breaks after this commit: [Use hash instead of counter for the unique id](https://github.com/WordPress/gutenberg/pull/63348/commits/189c7f51dc2bd395ca80c3a1435ecf4b0f76a9a4). If I revert that changes, all the images seem to work as expected.

Since multiple images can have same image source URL, same has is generated for all those images and so the image context is getting overridden.

`uniqid()` is used in [this change](https://github.com/WordPress/gutenberg/pull/63381/files#diff-1e4aa29ea183efda9c795271dc1acf6e8049355db104d0dca5b0ea5caf7f619cR194) 
I think this can be used safely, which is generating a unique ID based on the current timestamp.

luisherranz:
There are already some uses of `uniqid()` in [WordPress Core](https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20uniqid&type=code). Yes, let's use that 🙂

luisherranz:
@SantosGuillamot could you take another look? Thanks!

packages/block-library/src/image/view.js
```diff
@@ -381,25 +389,25 @@ const { state, actions, callbacks } = store(
...
 							buttonOffsetRight +
 							16;
 					}
-				} else {
-					ctx.imageButtonTop = buttonOffsetTop + 16;
-					ctx.imageButtonRight = buttonOffsetRight + 16;
 				}
+
+				state.metadata[ imageId ].imageButtonTop = imageButtonTop;
+				state.metadata[ imageId ].imageButtonRight = imageButtonRight;
```
madhusudhand, line 418: @luisherranz we might have to revert to keep the data in context instead, because it is being used in the template, not anywhere else.

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/index.php#L234-L235

It is breaking the icon placement.

image

SantosGuillamot (approved):
It looks good on my testing 🙂 

EDIT: I didn't check the icon as reported [here](https://github.com/WordPress/gutenberg/pull/63348#discussion_r1677535527). It's true that part is broken.

packages/block-library/src/image/view.js
SantosGuillamot: I can reproduce this as well.

luisherranz: We can use derived state, similar to how we are using in other places:

```js
store('...', {
  state: {
    get imageButtonRight() {
      const { imageId } = getContext();
      return state.metadata[imageId].imageButtonRight;
    },
    get imageButtonTop() ...
  },
});
```

```html
<button
  data-wp-style--right="state.imageButtonRight"
  data-wp-style--top="state.imageButtonTop"
>
  ...
</button>
```

madhusudhand: Thanks for the suggestion. The issue is fixed now.

luisherranz (approved):
Ok, this should be ready now. Thanks, Madhu and Mario.

Let's merge it and continue working on the fix for the [duplicate images](https://github.com/WordPress/gutenberg/pull/63381) and the Gallery block 🙂

PR #63351, opened 2024-07-10
InnerBlocks: Make sure blockType is set before trying to use it

mattsherman:
## What?

This PR adds a check that `blockType` is set before attempting to use it in `InnerBlocks`.

## Why?

This PR addresses an error that is thrown in the context of the WooCommerce new product editor when generating product variations. See https://github.com/woocommerce/woocommerce/pull/49216 for a try branch that triggers the error.

Without this fix, whenever variations are generated, the following error is logged in the browser dev console (note, we worked around the error in the UI in https://github.com/woocommerce/woocommerce/pull/49248, so it does not crash WooCommerce).

```
TypeError: undefined is not an object (evaluating 'w.providesContext')
```

## How?

This PR adds a check that `blockType` is set before attempting to use it in `InnerBlocks`.

## Testing Instructions

With the Gutenberg plugin built from this PR and a WooCommerce env running the branch from https://github.com/woocommerce/woocommerce/pull/49216 with the new product editor enabled (**WooCommerce** > **Settings** > **Advanced** > **Features** > **Experimental features**)...

1. Go to **Products** > **Add New**
2. On the `Variations` tab, click `Add Options` and add some variation options.
3. Verify that the variations are added correctly and the selected tab remains `Variations`
4. Verify that no `TypeError: undefined is not an object (evaluating 'w.providesContext')` error was logged in the browser dev console.

n/a

## Screenshots or screencast 

n/a

### DIFF
--- a/packages/block-editor/src/components/inner-blocks/index.js
+++ b/packages/block-editor/src/components/inner-blocks/index.js
@@ -132,7 +132,10 @@ function UncontrolledInnerBlocks( props ) {
 		/>
 	);
 
-	if ( Object.keys( blockType.providesContext ).length === 0 ) {
+	if (
+		! blockType?.providesContext ||
+		Object.keys( blockType.providesContext ).length === 0
+	) {
 		return items;
 	}
 


packages/block-editor/src/components/inner-blocks/index.js
```diff
@@ -132,7 +132,10 @@ function UncontrolledInnerBlocks( props ) {
 		/>
 	);
 
-	if ( Object.keys( blockType.providesContext ).length === 0 ) {
+	if (
+		! blockType?.providesContext ||
+		Object.keys( blockType.providesContext ).length === 0
```
ellatrix, line 137: I'm still not fully understanding how it's possible to have inner blocks without a block type. Could you give an example?

ellatrix (approved):

packages/block-editor/src/components/inner-blocks/index.js
mattsherman: Honestly, I am not sure how this is happening either! But, it happens in the scenario mentioned in the description with WooCommerce's use of blocks in the new product editor. Ultimately, the root problem may lie somewhere in WooCommerce's handling of blocks, but this change seems like a reasonable defensive change in GB.

ellatrix: Yes, I'm still curious. This is a legitimate error tbh, but sure, let's gracefully fail.

mattsherman: Agreed. We will still be attempting to address the root cause of this! I'm very curious too.

Issue #30369, opened 2021-03-30
Context detection for Query block


ntsekouras:
We need to find a way to detect the context a `Query` block 'lives' and handle any settings needed appropriately. By `context` I mean FSE context (site-editor) or a page/post (post-editor). Moreover is valuable to even know the kind of template we are on FSE context (related: https://github.com/WordPress/gutenberg/issues/29438).

Noting that we'll also have to take into account what happens when we are in `post-editor` and we `edit the template` from there.

There is an `inherit` property in `Query` which defaults to `true` and with that set, the Query's `query` for fetching results is inherited from the global query. In a single page there is no much use in having an inherit Query (doesn't fetch anything), but instead a custom one. On the other hand in an `archive` or `index` template proper results are shown without taking into account any other previously explicitly set options (like the number of posts).

Some prior art is here: https://github.com/WordPress/gutenberg/pull/27972, where `templateSlug` is passed as context and is used for now in `QueryLoop` to set the proper `postType` if we are in an archive template.

richtabor:
What would be the immediate advantage of this effort?

felixarntz:
@richtabor When you use a query block e.g. in a page, with `inherit: true` it will render a query that includes only that same single page - which is pointless. That's because when you visit the page in the frontend, the current WordPress query (which will be used if `inherit: true`) is for just that page.

In other words: While having `inherit: true` as a default makes sense for usage in certain templates or template parts, it does not make sense when used within a singular piece of content (e.g. a post or page). That's why context detection would help, it would lead to a smarter default of `inherit`.

Potentially we could even go further and not even offer the `inherit` control when using the query block in a singular piece of content. From a UX perspective, that would probably be even better than using `inherit: false` as default.

mikachan:
Thanks for giving all this context, @felixarntz 🙏 The Query Loop block has had some updates recently, with the last major one in https://github.com/WordPress/gutenberg/pull/65820. The block will now:

- Automatically set `inherit` to `true` when inserted into non-singular content, such as an archive or category template.
- Automatically set `inherit` to `false` when in singular content, such as a post, page or the 404 template, as this setting is pointless in this context (as you said).
- When inserted into singular content, the full custom query controls will be shown in the block settings, which makes it more obvious to the user that it's a custom query.
- When inserted into non-singular content, the query can still be customised as before, by changing the query type to 'custom'.

With all these recent changes, perhaps we could close this issue?

ntsekouras:
[...]

PR #63353, opened 2024-07-10
Query Loop: Change default query loop variations

youknowriad:
Related #30369 

## What?

A lot of people have raised that the default Query Loop is confusing because picking one of the initial variations results in a query loop that doesn't inherit the default WP query (permalinks, pagination, archive...)

The block itself has "inherit" set to true by default but for some reason (probably unintentional), inherit is set to false when "starting blank" and picking one of the variations.

**Note** There's another variation that I didn't not update, it's called "Posts List", I think it corresponds more to a "Latest Posts" variation. I believe the original purpose of this variation was to replace the "Latest Posts" widget/block which is also still selectable. I don't think we should update the "inherit" flag for this variation because it will change its purpose entirely but I think we have a few options there:
 
 - Remove that variation entirely
 - Rename it to "Latest Posts" or "Recent Posts" or something like that
 - Should we also hide the existing "Latests Posts" block?

(This can be addressed separately though)

## Testing Instructions

1- Insert Query Loop block
2- Click "Start blank"
3- Pick one of the proposed variations
4- Notice that the inserted query has the "inherit" toggle enabled.

### DIFF
--- a/packages/block-library/src/query/variations.js
+++ b/packages/block-library/src/query/variations.js
@@ -14,22 +14,6 @@ import {
 	imageDateTitle,
 } from './icons';
 
-const QUERY_DEFAULT_ATTRIBUTES = {
-	query: {
-		perPage: 3,
-		pages: 0,
-		offset: 0,
-		postType: 'post',
-		order: 'desc',
-		orderBy: 'date',
-		author: '',
-		search: '',
-		exclude: [],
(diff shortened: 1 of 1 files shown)


ellatrix (approved):

priethor (approved):
Works as expected ✅

Issue #63180, opened 2024-07-05
CustomSelectControl V2 popover renders below position: sticky elements


ciampo:
As flagged in https://github.com/WordPress/gutenberg/pull/62821 (see PR description):

The V2 version of the component automatically "flips" the popover to render on top of the trigger if there isn't enough space below it. In the specific example of the block toolbar, this causes a small visual glitch when the browser's viewport is short enough to cause the flip:

![Screenshot 2024-06-25 at 12 13 32](https://github.com/WordPress/gutenberg/assets/1083581/87dfe892-1bc6-4cf2-b1c0-e28bc3952923)

This glitch happens because the block toolbar's header has `position: sticky`, which causes it to render on top of the select popover despite a lower `z-index`.

A few preliminary ideas that try to avoid the problem rather than solving it:

- disallow the popover to flip
- allow for lower minimum height

I'm not sure we can solve this systematically at the CSS level. At most we could look into where in the DOM the popover is rendered, and consider portalling it so that it plays better with the stacking context?

ciampo:
While https://github.com/WordPress/gutenberg/pull/63357 should avoid the issue for the V2 legacy adapter (soon to be the new V1) component, we will still need to understand how to best solve this for the V2:

- keep `flip` disabled (although we could encounter the same issue in future layouts / etc)
- somewhat set as "boundaries" for the popover the correct height of the panel (excluding the header)
- solve it via allowing the popover to be `portal`ed to a plan in the DOM where the stacking context is not a constraints
- ?

ciampo:
Attempted a solution in #77969 , although I decided to close the PR with [the following reasoning](https://github.com/WordPress/gutenberg/pull/77969#issuecomment-4386353767):

> Update:
> 
> Local tests show that the approach tested in this PR would fix both #76126 and #63180.
> 
> However, we're in the middle of developing a new package of components (`@wordpress/ui`) that will include the same portalling features from the start. In fact, [the PR adding the new `SelectControl` component](https://github.com/WordPress/gutenberg/pull/77809) is currently under review.
> 
> After `SelectControl` is added to the `@wordpress/ui`, we will need to partially reworok how overlays are displayed and stacked in Gutenberg, moving from a list of discretionary portal slots and hardcoded z-indexes to a more implicit, modern architecture that allows the coexistence of legacy `@wordpress/components` and new `@wordpress/ui` components.
> 
> With that in mind, I'll close this PR, since it would delay the plan described above.
> 
[...]

PR #63357, opened 2024-07-10
CustomSelectControlV2: do not flip popover if legacy adapter

ciampo:
## What?

Part of #55023

Do not allow the select popover in the `CustomSelectControl` V2 legacy adapter to flip when there isn't much space to render below the trigger button.

## Why?

This matches more closely the behavior of the legacy, downshift-based component and prevents (avoids it without solving) the issue reported in https://github.com/WordPress/gutenberg/issues/63180

## How?

By setting the `flip` property to `false` on the ariakit select popover

## Testing Instructions

- Render `CustomSelectControlV2` legacy adapter and open the select popover
- Resize its window until there is not enough space for the popover
- Notice how the popover doesn't flip, but continues to render below the trigger button

## Screenshots or screencast 

|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/2a448ca7-274e-4ca1-86b0-cea492832af6" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/6996c082-6142-4d37-bad9-7feb54a6943e" /> |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### Internal
 
 -   `CustomSelectControl`: switch to ariakit-based implementation ([#63258](https://github.com/WordPress/gutenberg/pull/63258)).
--   `CustomSelectControl`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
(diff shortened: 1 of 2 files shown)


DaniGuardiola (approved):
Tests well for me!

tyxla (approved):
I like this 👍

PR #63358, opened 2024-07-10
Update Inherited Query Loop value from Template Settings changes

### LINKED ISSUE #54821: Live-update the site preview when changing posts per page value
When browsing the Index or Blog Home template there's an option to change the number of posts displayed per page:

Screenshot 2023-09-26 at 11 39 09

Ideally the site preview would always reflect this value, similar to how switching a style variation works in the Styles panel. Currently it only does so after a save action.

jeryj:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/54821

## Why?

Changes to the template settings for posts per page should be reflected by query loop blocks that read from this value.

## How?

Check to see if an edit has been made to the `posts_per_page` value via `getEntityRecordEdits`

## Testing Instructions

- Edit a template
- Add a query block
- Set the query block to Inherit in the Block settings sidebar
- Switch to the Document Tab in the settings sidebar
- Click the number of the posts per page value
- Change the value and check that the query loop posts number matches the value you set

### DIFF
--- a/packages/block-library/src/query/edit/query-content.js
+++ b/packages/block-library/src/query/edit/query-content.js
@@ -48,15 +48,26 @@ export default function QueryContent( {
 	} );
 	const { postsPerPage } = useSelect( ( select ) => {
 		const { getSettings } = select( blockEditorStore );
-		const { getEntityRecord, canUser } = select( coreStore );
+		const { getEntityRecord, getEntityRecordEdits, canUser } =
+			select( coreStore );
 		const settingPerPage = canUser( 'read', {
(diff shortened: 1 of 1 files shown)


scruffian (approved):
Nice little improvement. 🚢

annezazu:
Just a quick note that the video on this PR seems to be for something else and doesn't align with what was shipped.

jeffpaul:
@Sirjazzfeetz mind sharing your WordPress.org username so I can ensure its included in the 6.7 credits listing?

PR #63362, opened 2024-07-10
Set query loop to have the inherit value by default

jeryj:
## What?

Simplify flow by defaulting query block loops to inherit which relies on the Reading Settings or Template settings posts per page value.


scruffian (approved):
I rebased this. Since https://github.com/WordPress/gutenberg/pull/63353 and https://github.com/WordPress/gutenberg/pull/63404 this becomes just a text change, but I think it's still a good one.

PR #63365, opened 2024-07-10
DataViews: Cleanup preview styles

youknowriad:
Related #55083 

## What?

I found these styles in the dataviews table view and I was wondering what was their purpose and by reading the git history, it seems mostly needed to render the "previews" properly in the table view of the templates dataviews.

This PR clean up these styles in favor of a simple style that applies only to the preview field rather than impacting all the fields.

## Testing Instructions

1- Toggle preview field in the templates table o
2- Try the preview field in other layouts too.

### DIFF
--- a/packages/dataviews/src/style.scss
+++ b/packages/dataviews/src/style.scss
@@ -200,16 +200,6 @@
 			min-height: $grid-unit-40;
 			display: flex;
 			align-items: center;
-
-			> * {
-				flex-grow: 1;
-			}
-
-			&.dataviews-view-table__primary-field {
-				a {
-					flex-grow: 0;
-				}
-			}
 		}
 	}
 	.dataviews-view-table-header-button {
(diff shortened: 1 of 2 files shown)


jameskoster:
In comparing the Templates, Pages, and Patterns data views I see no difference between this PR and Trunk 👍

ntsekouras (approved):
I didn't notice something different visually and it's good to have this css rule removed. I'm not sure if something have changed over time from the PR that introduced it, but it seems to me I'd have tried a change in `width` instead of having this rule, that I'd commented I didn't like it.  😄 

Anyway, thank you!

jameskoster:
I'm now wondering if my build failed during testing because I'm seeing some regressions that might be related to this PR?

In grid layout the featured image doesn't fill the container:

Screenshot 2024-07-11 at 11 14 47

In table layout the featured image is incorrectly sized with mismatching corner radius:

Screenshot 2024-07-11 at 11 15 04

youknowriad:
Yes, I'm seeing a few issues myself I'll look into it.

youknowriad:
@jameskoster Reverting this change doesn't solve the issue, so It's either something else or it was always broken kind of.

youknowriad:
Ok, so it's a small mistake I made in #63334

Issue #55100, opened 2023-10-05
Data view: Search & Filtering


jameskoster:
Part of https://github.com/WordPress/gutenberg/issues/55083

This ticket is a stub to include designs for search & filtering in tables. Considerations that could affect the design include:

* Scalability – there can be thousands of datapoints that a user might want to filter by. Performance, search, etc. all need to be factored in to the UI.
* Optimise for common use cases: It could be beneficial for some tables to present 'primary' filters to users by default.
* Advanced filter operators. These can be per-filter e.g: "Show posts in categories: recipes **and** reviews" vs. "Show posts in categories: recipes **or** reviews". Or even combined across filters e.g. "Show published posts **and** posts with no comments" vs. "Show published posts **or** posts with no comments". Clearly this can get quite complicated quite quickly, so should receive appropriate prominence.
* Saving filter configurations for reuse.
* Resetting; per-filter and globally.
* Shortcuts; create filter from column headers.
* Loading states while filters are enacted.

## Iterations

Support pages, templates, and template parts pages.

DataViews component

- [ ] Top-level bar:
  - [x] Spacing issues in the top-level bar ([see](https://github.com/WordPress/gutenberg/pull/55992#pullrequestreview-1731806181)). [#56151](https://github.com/WordPress/gutenberg/pull/56151)
  - [ ] Make sure it grows gracefully when there are many of them ([see](https://github.com/WordPress/gutenberg/pull/56468#issuecomment-1823988214)).
  - [x] @oandregal Create a more condensed view for filters in the list layout. [Design](https://github.com/WordPress/gutenberg/issues/55100#issuecomment-1847417577). [#56983](https://github.com/WordPress/gutenberg/pull/56983)
  - [ ] Disable filters when there's only one option. E.g. filtering pages by author is pointless on a site with only a single author.
  - [ ] Hide the combobox in search widget when there are more than X options. Probably use a `selectlist`.
- [x] Global search [#54966](https://github.com/WordPress/gutenberg/pull/54966)
- [x] Filter for `ENUMERATION` field. Target [design](https://github.com/WordPress/gutenberg/issues/55100#issuecomment-1772753679). @oandregal 
  - [x] Basic implementation based on `SelectControl`. [#55270](https://github.com/WordPress/gutenberg/pull/55270)
[...]

PR #63367, opened 2024-07-10
Update: Grid layout: Allow users to adjust grid density

### LINKED ISSUE #60630: Data views grid layout: Allow users to adjust grid density, and consumers to set default density
Grid density

For some views the browsing experience in grid layout can be improved by adjusting the grid density, for instance viewing patterns in fewer columns so that previews are much larger.

It would be neat to allow consumers to specify a default grid density for views, and for users to customise that ad hoc.

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60630
Part of: https://github.com/WordPress/gutenberg/issues/63128

cc: @jameskoster 

This PR adds a UI on the grid view to allow users to select a density.
The density depends on the viewport, and we kept that behavior if the viewport changes density can increase or decrease.

## Screenshot
Screenshot 2024-07-10 at 16 34 01

## Testing Instructions
I verified I was able to change the density of the grid on patterns, templates and pages.

### DIFF
--- a/packages/dataviews/src/layouts/grid/density-picker.tsx
+++ b/packages/dataviews/src/layouts/grid/density-picker.tsx
@@ -7,8 +7,8 @@ import { useViewportMatch } from '@wordpress/compose';
 import { plus, lineSolid } from '@wordpress/icons';
 import { useEffect } from '@wordpress/element';
 
-const viewPortBreaks = {
-	xhuge: { min: 3, max: 5, default: 5 },
+const viewportBreaks = {
+	xhuge: { min: 3, max: 6, default: 5 },
 	huge: { min: 2, max: 4, default: 4 },
 	xlarge: { min: 2, max: 3, default: 3 },
 	large: { min: 1, max: 2, default: 2 },
@@ -40,6 +40,12 @@ function useViewPortBreakpoint() {
 	return null;
 }
 
+// Value is number from 0 to 100 representing how big an item is in the grid
+// 100 being the biggest and 0 being the smallest.
+// The size is relative to the viewport size, if one a given viewport the
+// number of allowed items in a grid is 3 to 6 a 0 ( the smallest ) will mean that the grid will
+// have 6 items in a row, a 100 ( the biggest ) will mean that the grid will have 3 items in a row.
+// A value of 75 will mean that the grid will have 4 items in a row.
 function getRangeValue(
 	density: number,
 	breakValues: { min: number; max: number; default: number }
@@ -54,16 +60,25 @@ export default function DensityPicker( {
 	setDensity,
 }: {
 	density: number;
-	setDensity: ( density: number ) => void;
+	setDensity: React.Dispatch< React.SetStateAction< number > >;
 } ) {
-	const viewPort = useViewPortBreakpoint();
+	const viewport = useViewPortBreakpoint();
 	useEffect( () => {
-		setDensity( 0 );
-	}, [ setDensity, viewPort ] );
-	if ( ! viewPort ) {
+		setDensity( ( _density ) => {
+			if ( ! viewport || ! _density ) {
+				return 0;
+			}
+			const breakValues = viewportBreaks[ viewport ];
+			if ( _density >= breakValues.min && _density <= breakValues.max ) {
+				return _density;
+			}
+			return breakValues.default;
+		} );
+	}, [ setDensity, viewport ] );
+	if ( ! viewport ) {
 		return null;
 	}
-	const breakValues = viewPortBreaks[ viewPort ];
+	const breakValues = viewportBreaks[ viewport ];
 	const densityToUse = density || breakValues.default;
 	const rangeValue = getRangeValue( densityToUse, breakValues );
 
--- a/packages/dataviews/src/layouts/grid/density-picker.tsx
+++ b/packages/dataviews/src/layouts/grid/density-picker.tsx
@@ -69,10 +69,13 @@ export default function DensityPicker( {
 				return 0;
 			}
 			const breakValues = viewportBreaks[ viewport ];
-			if ( _density >= breakValues.min && _density <= breakValues.max ) {
-				return _density;
+			if ( _density < breakValues.min ) {
+				return breakValues.min;
 			}
-			return breakValues.default;
+			if ( _density > breakValues.max ) {
+				return breakValues.max;
+			}
+			return _density;
 		} );
 	}, [ setDensity, viewport ] );
 	if ( ! viewport ) {
--- a/packages/dataviews/src/layouts/grid/density-picker.tsx
+++ b/packages/dataviews/src/layouts/grid/density-picker.tsx
@@ -1,0 +1,105 @@
+/**
+ * WordPress dependencies
+ */
+import { RangeControl, Button } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { useViewportMatch } from '@wordpress/compose';
+import { plus, lineSolid } from '@wordpress/icons';
+
+const viewPortBreaks = {
+	xhuge: [ 2, 6 ],
+	huge: [ 2, 5 ],
+	xlarge: [ 1, 3 ],
+	mobile: [ 1, 2 ],
+};
+
+function useViewPortBreakpoint() {
+	const isXHuge = useViewportMatch( 'xhuge', '>=' );
+	const isHuge = useViewportMatch( 'huge', '>=' );
+	const isXlarge = useViewportMatch( 'xlarge', '>=' );
+	const isMobile = useViewportMatch( 'mobile', '>=' );
+
+	if ( isXHuge ) {
+		return 'xhuge';
+	}
+	if ( isHuge ) {
+		return 'huge';
+	}
+	if ( isXlarge ) {
+		return 'xlarge';
+	}
+	if ( isMobile ) {
+		return 'mobile';
+	}
+	return null;
+}
+
+function getRangeValue( density: number, breakValues: number[] ) {
+	const inverseDensity = breakValues[ 1 ] - density;
+	const max = breakValues[ 1 ] - breakValues[ 0 ];
+	return Math.round( ( inverseDensity * 100 ) / max );
+}
+
+export default function DensityPicker( {
+	density,
+	setDensity,
+}: {
+	density: number;
+	setDensity: ( density: number ) => void;
+} ) {
+	const viewPort = useViewPortBreakpoint();
+	if ( ! viewPort ) {
+		return null;
+	}
+	const breakValues = viewPortBreaks[ viewPort ];
+	const densityToUse = density || breakValues[ 1 ];
+	const rangeValue = getRangeValue( densityToUse, breakValues );
+
+	const step = 100 / ( breakValues[ 1 ] - breakValues[ 0 ] + 1 );
+	return (
+		<>
+			<Button
+				size="compact"
+				icon={ lineSolid }
+				disabled={ rangeValue <= 0 }
+				accessibleWhenDisabled
+				label={ __( 'Decrease size' ) }
+				onClick={ () => {
+					setDensity( densityToUse + 1 );
+				} }
+			/>
+			<RangeControl
+				__nextHasNoMarginBottom
+				className="dataviews-density-picker__range-control"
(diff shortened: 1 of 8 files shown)


jameskoster:
Thanks @jorgefilipecosta, this is very cool :) 

Could we apply `__nextHasNoMarginBottom` to the range?

Would it be feasible to define upper and lower bounds based on breakpoint? For example you'd never want to use a 5 column arrangement on mobile, so it doesn't make sense to offer that option. 

- Above 1920: 2 - 6
- Between 1920 and 1440: 2 - 5
- Between 1440 and 1080: 1 - 3
- Between 1080 and 782: 1 - 2

the + / - icons feel a bit counter-intuitive to the behavior. I think I'd expect the + to increase the size of the grid item, similar to this UI in MacOS Photos:

Screenshot 2024-07-11 at 14 23 29

With that said, I wonder if we should flip things around. Instead of 'Grid density' it could be 'Item size'. IE 'Decrease item size' and 'Increase item size'. As the range value increase, the number of columns decreases. Hope that makes sense 🤞

jorgefilipecosta:
Hi @jameskoster, I updated the code to follow the logic you suggested. Let me know if the behavior is what you had in mind.

jameskoster:
Thanks Jorge, this is getting close and working pretty well. 

Would it be possible to update the display on container resize? One small issue currently is that; if you set a large number of columns on a wide screen, then scale down, the large number of columns persists (and doesn't work very well) on the smaller screen. The control also behaves a bit strangely. Video demo:


https://github.com/user-attachments/assets/395701e1-977e-4466-8c60-78b0f7dab966

Ideally the display and control update in line with the container size.

Somewhat related: is it possible to set a default for each scale? E.g. above 1920 you can choose between 2 and 6 columns, but the default is 4?

jameskoster:
Another tiny point, the values that appear as tooltips on the range don't really communicate anything useful, we can probably hide them.

Screenshot 2024-07-17 at 10 50 18

jorgefilipecosta:
Hi @jameskoster, thank you for the review and suggesting enhancements, the suggestions were applied.

jameskoster:
Hey @jorgefilipecosta, I pushed a small update to the config. One issue is that the `default` value doesn't seem to apply at the xhuge breakpoint, which results in this broken interaction:


https://github.com/user-attachments/assets/795873b1-acbd-4117-998d-687b09e85d5b

One other request if possible: if you set a non-default value, could that value persist as you resize the screen (assuming the value is compatible with the associated breakpoint)? For example if I set 3 columns at the `xhuge` breakpoint, the layout would remain at 3 columns as I resize down until the `large` breakpoint (which doesn't support 3 columns). And vice versa, if I set 2 columns on `mobile`, the layout should stay in a 2 column arrangement until the `xhuge` breakpoint. If this adds too much complexity to the PR it's probably fine to explore separately.

stokesman:
I have an idea that could help provide visual indication of what’s happening with available sizes/layout for different viewport sizes. That is to borrow [the styling](https://github.com/WordPress/gutenberg/blob/0a7e10db2037cb80e4ab8cda4363ba18d6a3a83f/packages/block-editor/src/components/spacing-sizes-control/style.scss#L22-L39) from `SpacingSizesControl`. 
image

Unfortunately, the styles would have to be duplicated as they’re not inherent to the mark styling of `RangeControl`. This could also be tried separately of course. Just a thought.

packages/dataviews/src/layouts/grid/density-picker.tsx
stokesman, line 1: It’s not of consequence but viewport is one word so any `viewPort*` probably should be `viewport*`

packages/dataviews/src/layouts/grid/density-picker.tsx
```diff
@@ -0,0 +1,118 @@
...
+			/>
+			<RangeControl
+				__nextHasNoMarginBottom
+				showTooltip={ false }
+				className="dataviews-density-picker__range-control"
+				label={ __( 'Item size' ) }
+				hideLabelFromVision
+				value={ rangeValue }
+				min={ 0 }
+				max={ 100 }
```
stokesman, line 109: I wonder if there’s any minor simplification that might be enabled were these to be:
```suggestion
				min={ breakValues.min }
				max={ breakValue.max }
```
That’s a cursory thought so I'm not asking for any explanation in case you've already considered it.

jorgefilipecosta:
> e: if you set a non-default value, could that value persist as you resize the screen (assuming the value is compatible with the associated breakpoint)? For example if I set 3 columns at the `xhuge` breakpoint, the layout would remain at 3 columns as I resize down until the `large` breakpoint (which doesn't support 3 columns). And vice versa, if I set 2 columns on `mobile`, the layout should stay in a 2 column arrangement

Hi @jameskoster, your suggestions were implemented, let me know if this looks good.

packages/dataviews/src/layouts/grid/density-picker.tsx
jorgefilipecosta: Hi @stokesman, good point, it was corrected 👍

packages/dataviews/src/layouts/grid/density-picker.tsx
jorgefilipecosta: Hi @stokesman, I think this logic would not work, for example for the xhuge size where the grid item min is 3 and the max is 6 when density is at 3 (the min) the slider on the range should be at the maximum. What the user is picking is the size of the item not how many items there are on the grid.
The range represents a number from 0 to 100, of how big the item is, for now, the only way we control size is using the number of grid columns, but in the future, this logic leaves the door open for also using gap etc. I added a comment specifying what the range number means.

jorgefilipecosta:
Hi @stokesman, I think depending on @WordPress/gutenberg-design thoughts on the suggestion we could explore adding marks to the possible breaks on the range similar to what the SpacingSizesControl does on a follow-up PR.

packages/dataviews/src/layouts/grid/density-picker.tsx
stokesman: Thanks for elaborating and commenting the code 🙇

jameskoster:
That's an excellent suggestion @stokesman, thanks for pointing that out. I agree it's something that would improve the experience here. We should probably consider making that component public.

@jorgefilipecosta I think this is working well now, thanks for all the adjustments ❤️ 

One question; is it possible for consumers to supply different configs? For instance the Pages data view might need to supply different density options to the Templates data view. Similarly it should probably be possible for a data view to opt out of this option altogether.

jameskoster:
I noticed `RangeControl` includes a `marks` prop for displaying step indicators. Is that worth trying here?

jorgefilipecosta:
> I noticed `RangeControl` includes a `marks` prop for displaying step indicators. Is that worth trying here?

Hi @jameskoster,

Yes, I tried the marks, but it does not seems like what we want:
Screenshot 2024-07-22 at 16 06 58
This is how they look by default on the component.

 We could of course hide labels and get this:
Screenshot 2024-07-22 at 16 10 37

Do you think using marks with labels hidden would be a good idea?

jorgefilipecosta:
> One question; is it possible for consumers to supply different configs? For instance the Pages data view might need to supply different density options to the Templates data view. Similarly it should probably be possible for a data view to opt out of this option altogether.

Right now it is not consumer configurable the main goal of this PR is to get an initial version that does not changes API's for the consumer. But I have an hardcoded config object (for now) so changing to something consumer configurable will be easy. We just need to discuss the API for how the config object could be passed, and our hardcoded one could start to become the default in case nothing is passed. I guess we could work on the consumer configuration in a follow-up?

jameskoster:
Marks without labels is the way to go, but that looks a bit broken. Let's get that fixed at the component level (cc @ciampo @mirka) then follow up to add it here. 

> I guess we could work on the consumer configuration in a follow-up?

Sounds good to me.

jameskoster (approved):
I'd welcome more design feedback, but I reckon this is in a decent spot to get in since we're early in the release cycle, and begin gathering feedback.

paaljoachim:
Very nice!

youknowriad:
I will reiterate the feedback that I shared before. I think all view configuration should be in the ViewConfig dropdown. I'm concerned that we keep adding buttons and controls outside of that dropdown.

jameskoster:
In this case the range control is much more ergonomic to use than a menu. That said, perhaps there's scope to move the range inside the menu? I'd be interested to hear the a11y story around something like this:

Preview size

@jorgefilipecosta if it's not a lot of work, it might be worth trying this in a separate PR to gather feedback?

jorgefilipecosta:
> I will reiterate the feedback that I shared before. I think all view configuration should be in the ViewConfig dropdown. I'm concerned that we keep adding buttons and controls outside of that dropdown.

Sorry @youknowriad, with the discussion that followed I lost track of this suggestion. I can give a try to the suggestion @jameskoster proposed. But I'm not sure if our components allow that, I will look.

ciampo:
@jameskoster 
> That said, perhaps there's scope to move the range inside the menu? I'd be interested to hear the a11y story around something like this:

@jorgefilipecosta 
> But I'm not sure if our components allow that, I will look.

Using a range input or interactive form controls in general as a menu item may not be allowed. From [the `menu` and `menubar` APG](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/):

> The items contained in a menu are child elements of the containing menu or menubar and have any of the following roles:
> - [menuitem](https://w3c.github.io/aria/#menuitem)
> - [menuitemcheckbox](https://w3c.github.io/aria/#menuitemcheckbox)
> - [menuitemradio](https://w3c.github.io/aria/#menuitemradio)

And from `menuitem` [documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role), it looks like a menu item needs to represent "an option in a set of choices", which gets activated when pressing the enter and space key.

There is even an [active conversation](https://github.com/w3c/aria/issues/1587) happening at the ARIA standards level about introducing a role `menuitemslider`, but no concrete advancements have been made so far.

Having said that, I don't know if we could find a way to include a slider inside a `menu`
Also pinging @diegohaz in case `ariakit` could offer a solution here — maybe adding the slider (and other interactive controls) outside of the menu container, if that's even a possibility?

youknowriad:
For me that dropdown is not a menu, I'm not a simple select box and I'm not navigating anywhere, it's a config dropdown, I'm editing the "view" config. I can see things like text inputs, sliders, selects (the existing menus)... So I think we should treat it like that instead

jameskoster:
@youknowriad that's quite a different proposition in terms of design, are you imagining something more Settings (https://github.com/WordPress/gutenberg/issues/63624) oriented? Rough mockup:

view options

ciampo:
If it helps, we had a similar conversation in https://github.com/WordPress/gutenberg/issues/55100#issuecomment-1878385884 and following comments.

youknowriad:
> @youknowriad that's quite a different proposition in terms of design, are you imagining something more Settings (https://github.com/WordPress/gutenberg/issues/63624) oriented? Rough mockup:

That mockup works but I don't see why closer design can't work in a dropdown.

jameskoster:
> I don't see why closer design can't work in a dropdown.

A popover might work, but without the flyouts to help organise things it could end up being quite large and feel a bit unwieldy. If we're committed to moving away from a menu lets open an issue to explore the options.

jorgefilipecosta:
I'm trying options at https://github.com/WordPress/gutenberg/pull/63852. I guess we can discuss the possibilities there, and then I update the code to reflect the option we commit to.

diegohaz:
> Having said that, I don't know if we could find a way to include a slider inside a `menu`
> Also pinging @diegohaz in case `ariakit` could offer a solution here — maybe adding the slider (and other interactive controls) outside of the menu container, if that's even a possibility?

A `role=menu` container can only contain `group`, `separator`, `menuitem`, and other `menuitem` role types as its accessible children.

However, there are other roles that could be useful here. A simple alternative could be a popover with a vertical `Toolbar`, if you want to keep the arrow key navigation.

A more complex solution could be a single-column `grid` (`gridcell` elements may contain widgets inside, like select and slider—related: [Composite widgets with `grid` role](https://ariakit.org/changelog#composite-widgets-with-grid-role)).

PR #63368, opened 2024-07-10
Post date block: add a block example

MaggieCabrera:
## What?

Add a basic example to the `core/post-date` block.

## Why?

## How?

## Testing Instructions

Open the block inserter and search for the post date block, hover over it and you'll see the example. 

## Screenshots or screencast 
Screenshot 2024-07-10 at 18 02 48

### DIFF
--- a/packages/block-library/src/post-date/block.json
+++ b/packages/block-library/src/post-date/block.json
@@ -23,6 +23,9 @@
 		}
 	},
 	"usesContext": [ "postId", "postType", "queryId" ],
+	"example": {
+		"viewportWidth": 350
+	},
 	"supports": {
 		"html": false,
 		"color": {


jasmussen (approved):
In the site editor, I just see this:

Screenshot 2024-07-11 at 09 41 06

That's better than no preview, but if we can provide custom dates here, that would be better.

jasmussen:
If you can provide custom content, an easter egg such as WordPress' birth date, might be fun.

MaggieCabrera:
> If you can provide custom content, an easter egg such as WordPress' birth date, might be fun.

from what I understand, there's no conditional here, if we did set up placeholders (I'm not sure we can) they would replace the real data when there is context

PR #63369, opened 2024-07-10
Post terms block: add an example

MaggieCabrera:
## What?

Add a basic example to the `core/post-terms` block.

## Why?

## How?

## Testing Instructions

Open the block inserter and search for the post tags block, hover over it and you'll see the example. The Post categories block should show the same.
On Global Styles, go to blocks, and search for post terms: it should have a preview.

## Screenshots or screencast 

Screenshot 2024-07-10 at 18 10 40
Screenshot 2024-07-10 at 18 10 48

### DIFF
--- a/packages/block-library/src/post-terms/block.json
+++ b/packages/block-library/src/post-terms/block.json
@@ -27,6 +27,9 @@
 		}
 	},
 	"usesContext": [ "postId", "postType" ],
+	"example": {
+		"viewportWidth": 350
+	},
 	"supports": {
 		"html": false,
 		"color": {


jasmussen:
Good step forward. For some reason I see this:
Screenshot 2024-07-11 at 09 38 30

Can we supply some specific tags and categories as part of the example? That might also let us avoid the typographic widow here:

Screenshot 2024-07-11 at 09 39 26

jasmussen:
I realize the lack of content in my screenshot is because I'm testing in the site editor, where it just shows its placeholder state. But if it's possible to still show content here, it would be useful.

jasmussen (approved):

MaggieCabrera:
> I realize the lack of content in my screenshot is because I'm testing in the site editor, where it just shows its placeholder state. But if it's possible to still show content here, it would be useful.

I think that's a problem in general with these types of blocks, the placeholders are not helpful where we need them most to be: the site editor, which is where they are generally used. When we insert the terms block in a template they look exactly like that, they just say "categories" or "tags". We only see them in context in the frontend. It's a bigger problem than the preview, it's a problem with the block, I would say.

jameskoster:
@jasmussen this is the 'abecedarium' issue. IE show meaningful block previews when there's no context, while still making it obvious what the block does.

jasmussen:
Yep, and not a blocker!

Issue #63342, opened 2024-07-10
Post Summary Panel: Change the "Publish" label to a more accurate label


t-hamano:
## What problem does this address?

As I understand it, the word "Publish" is only used as a verb. Therefore, I think it is appropriate to leave the blue button in the upper right as "Publish".

![image](https://github.com/WordPress/gutenberg/assets/54422211/5d387e6a-aca0-4d52-9a04-4fbbdb892674)

On the other hand, the same word "Publish" is used in the summary panel. It feels strange that this label is the only one that is a verb, while all the other labels are nouns.

![image](https://github.com/WordPress/gutenberg/assets/54422211/edb0c58e-cc61-4f9d-8310-9a3e192d87d1)

Perhaps the word "Publish" can cover both the meaning of a verb and a noun in English, but this translatable string does not have any additional context.

Therefore, it is not possible to apply a context-sensitive translation in locales other than English.

## What is your proposed solution?

I would like to suggest using a more accurate label other than "Publish" in the summary panel.

If we want to follow the label in the classic editor, "**Published on**" might be good.

![image](https://github.com/WordPress/gutenberg/assets/54422211/198a8dcb-7621-4ce6-ad57-f51dc6b8c707)

Alternatively, I would like to suggest the text "**Published date**".

cc @WordPress/gutenberg-design @afercia

afercia:
@t-hamano good point. I'd agree the verb sounds a bit out of place there. In a way this was also noted in an user session long time ago, see https://github.com/WordPress/gutenberg/issues/470#issuecomment-386804617

> "Publish" is weird to me, it seems like an action but it's just text ...

Couple more considerations:

The default date/time value when creating a new post is `Immediately`, which is likely the reason why this visual label was changed to `Publish`. The new wording should complement well with `Immediately`.

The verb `Publish` is used also as the title of the calendar popover. I'd think it could be improved there as well. Scrsenshot:

![Screenshot 2024-07-10 at 10 23 23](https://github.com/WordPress/gutenberg/assets/1682452/9f791cea-969b-4fa3-b59a-b2597e46eacf)

jameskoster:
Date information is a bit messy. In data views there is a "Date" field which conditionally displays one of the following:

* Last _modified_ date for unpublished posts
* Exact _publish date_ for published posts

Then there's this "Publish" control in the Inspector, which communicates just the publish date for both published and unpublished posts.

[...]

PR #63370, opened 2024-07-10
Fix Publish Date text

Rishit30G:
## What?
This PR is fixing the left panel with text 'Publish' to 'Published Date' 

## Why?
Link to the Issue thread: https://github.com/WordPress/gutenberg/issues/63342

## Testing 
I was able to see the text to be 'Published Date' in Popover and Panel 

## Screenshots or screencast 
image

### DIFF
--- a/packages/block-editor/src/components/publish-date-time-picker/index.js
+++ b/packages/block-editor/src/components/publish-date-time-picker/index.js
@@ -33,7 +33,7 @@ export function PublishDateTimePicker(
 	return (
 		<div ref={ ref } className="block-editor-publish-date-time-picker">
 			<InspectorPopoverHeader
-				title={ __( 'Publish' ) }
+				title={ __( 'Published Date' ) }
 				actions={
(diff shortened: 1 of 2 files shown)


t-hamano (commented):
@Rishit30G 

Thanks for the PR!

Apparently the current label is "Publish" for a reason. To quote from [this comment](https://github.com/WordPress/gutenberg/issues/63342#issuecomment-2219929084):

> The default date/time value when creating a new post is `Immediately`, which is likely the reason why this visual label was changed to `Publish`. The new wording should complement well with `Immediately`.

I think we need to have some more discussion to solve this problem. If you have any opinions, please feel free to comment in #63342.

afercia:
> I think we need to have some more discussion to solve this problem. 

I'd agree with @t-hamano this would need some more thinking. Also, the horizontal space is very limited and I'd recommend to double check with translations that may provide a string that is much longer than 'Published Date'. Will comment on the issue.

PR #63375, opened 2024-07-10
Query Loop block: Convert the post content type setting to a ToggleGroupControl if there are few items

### LINKED ISSUE #63374: Query Loop block: switch 'Post type' inspector setting to use radio buttons when there are few items
The Query Loop block's post type setting displays as a dropdown when, by default, there are only two post types to select from (`Post` and `Page`). While dropdowns are useful for avoiding visual clutter when there are many options to choose from, when there are few items, they require an extra click with no advantages. 

image

Using a RadioControl instead of a SelectControl when there are only two items.

priethor:
Closes #63374.

## What?
When there are fewer than three items, this PR converts the Query Loop's post content type inspector control to a `ToggleGroupControl` instead of using the `SelectControl`.

It also changes the label copy from `Post type` to `Post content` for accuracy and applies `__next40pxDefaultSize` to this and other Query Loop inspector controls.

## Why?
Select controls are useful for avoiding visual clutter when there are many options to choose from. However, they require an extra click and do not offer visual improvements when there are few items. 

## How?
By using a `ToggleGroupControl` instead of a `SelectControl` when there are fewer than three items.

## Testing Instructions
1. In a vanilla site, add a Query Loop block or edit an existing one.
2. Disable the `Inherit query from template` setting.
3. The `Post content` option will appear using a ToggleGroupControl with `Post` and `Page` as options.
4. Add a custom post type or a plugin that adds more post types.
5. Edit the same Query Loop block as before, and notice there are now more options but showing as a SelectControl.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/27339341/40f0503b-94bc-42bf-98c0-04b8c0ebd5fa

### DIFF
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -5,6 +5,7 @@ import {
 	PanelBody,
 	TextControl,
 	SelectControl,
+	RadioControl,
 	RangeControl,
 	ToggleControl,
 	Notice,
@@ -101,6 +102,10 @@ export default function QueryInspectorControls( props ) {
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
 	const showPostTypeControl =
 		! inherit && isControlAllowed( allowedControls, 'postType' );
+	const postTypeControlLabel = __( 'Post content' );
+	const postTypeControlHelp = __(
+		'WordPress contains different types of content and they are divided into collections called “Post types”. By default there are a few different ones such as blog posts and pages, but plugins could add more.'
+	);
 	const showColumnsControl = false;
 	const showOrderControl =
 		! inherit && isControlAllowed( allowedControls, 'order' );
@@ -152,18 +157,26 @@ export default function QueryInspectorControls( props ) {
 							}
 						/>
 					) }
-					{ showPostTypeControl && (
-						<SelectControl
-							__nextHasNoMarginBottom
-							options={ postTypesSelectOptions }
-							value={ postType }
-							label={ __( 'Post type' ) }
-							onChange={ onPostTypeChange }
-							help={ __(
-								'WordPress contains different types of content and they are divided into collections called “Post types”. By default there are a few different ones such as blog posts and pages, but plugins could add more.'
-							) }
-						/>
-					) }
+					{ showPostTypeControl &&
+						( ( postTypesSelectOptions.length > 2 && (
+							<SelectControl
+								__nextHasNoMarginBottom
+								options={ postTypesSelectOptions }
+								value={ postType }
+								label={ postTypeControlLabel }
+								onChange={ onPostTypeChange }
+								help={ postTypeControlHelp }
+							/>
+						) ) || (
+							<RadioControl
+								__nextHasNoMarginBottom
+								options={ postTypesSelectOptions }
+								selected={ postType }
+								label={ postTypeControlLabel }
+								onChange={ onPostTypeChange }
+								help={ postTypeControlHelp }
+							/>
+						) ) }
 					{ showColumnsControl && (
 						<>
 							<RangeControl
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -5,9 +5,10 @@ import {
 	PanelBody,
 	TextControl,
 	SelectControl,
-	RadioControl,
 	RangeControl,
 	ToggleControl,
+	__experimentalToggleGroupControl as ToggleGroupControl,
+	__experimentalToggleGroupControlOption as ToggleGroupControlOption,
 	Notice,
 	__experimentalToolsPanel as ToolsPanel,
 	__experimentalToolsPanelItem as ToolsPanelItem,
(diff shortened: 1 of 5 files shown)


richtabor:
Have you tested with third-party post types available as well? How well does this scale?

priethor:
I have tested it with the third-party post types provided by the `Essential Content Types` plugin. It scales well since, with more than 2 post types, the behavior is the same as before this PR, showing a dropdown; the changes only apply when there are only 2 default post types.

youknowriad:
Nice PR @priethor :) I think we can move forward with it but I'm actually wondering if this is something we should generalize in the SelectComponent (like for all selects) or maybe make it opt-in via a prop or something? Curious to hear @ciampo's thoughts.

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,6 +102,10 @@ export default function QueryInspectorControls( props ) {
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
 	const showPostTypeControl =
 		! inherit && isControlAllowed( allowedControls, 'postType' );
+	const postTypeControlLabel = __( 'Post content' );
```
youknowriad: Should it be more "Content types". I feel like that it clearer to users.

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,6 +102,10 @@ export default function QueryInspectorControls( props ) {
...
+	const postTypeControlHelp = __(
+		'WordPress contains different types of content and they are divided into collections called “Post types”. By default there are a few different ones such as blog posts and pages, but plugins could add more.'
```
youknowriad: Can we find a wording that is maybe a bit less verbose?

youknowriad:
This is looking good to me. The spacing is a bit off maybe so let's see what designers say.

richtabor:
Another quick check: would a ToggleGroup work better here? If it's only ever for two items max?

jasmussen:
ToggleGroup would be nice if it were two, but we don't want many edgecases, i.e. togglegroup, then radios, then dropdown. Radios work okay as far as translatability of long strings, though. 

So if the PR is mainly to optimize for the case of "Post" and "Page", I'd go segmented control as Rich suggests: it seems to be the 80% case, and the specific design of the segmented control better affords the very verbose help-text that sits right below the box.

priethor:
>  it seems to be the 80% case

Since built-in post types cannot be unregistered via `unregister_post_type()`, I think Post and Page are 100% of the supported cases with only two post types, so even more reason to try with the ToggleGroup. 👀

priethor:
Here's an attempt with a `ToggleGroupControl`. Looks much better and saves the always helpful vertical real-state on the inspector sidebar:

https://github.com/user-attachments/assets/1d513bfe-e4b8-423b-98f5-9b18edf04371

jasmussen:
Looks good. Can you make the segmented control full width? I.e. 100% width of the inspector panel.

priethor:
Gotcha. Even better 👍

https://github.com/user-attachments/assets/fcb4f801-4653-4106-888c-c7aea7093cc1

The "Sticky posts" dropdown has a fixed number of three items. Should I follow up later by converting it to a ToggleControlGroup as well for consistency?

jasmussen:
If we believe the 3 items can fit in a segmented control even in translations, definitely a good choice. Otherwise I'd probably keep it a dropdown. "Only" is a bit of a weird term, though.

Separately, we can add the `__next40pxDefaultSize` to everything that isn't intrinsically 40px while we're here.

priethor:
Applied `__next40pxDefaultSize ` to:
- Content type selector (both ToggleGroup and Select)
- Order by selector
- Sticky selector
- Advanced keyword, author, and taxonomy filters

image

jasmussen:
Nice work!

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -152,22 +158,42 @@ export default function QueryInspectorControls( props ) {
...
-							value={ postType }
-							label={ __( 'Post type' ) }
-							onChange={ onPostTypeChange }
-							help={ __(
-								'WordPress contains different types of content and they are divided into collections called “Post types”. By default there are a few different ones such as blog posts and pages, but plugins could add more.'
-							) }
-						/>
-					) }
+					{ showPostTypeControl &&
+						( ( postTypesSelectOptions.length > 2 && (
```
youknowriad: Nit: The combination of `&&` and `||` is very weird. I would use a ternary personally 

`showPostTypeControl && (postTypesSelectOptions.length > 2 ? something : somethingElse )`

youknowriad (approved):

packages/block-library/src/query/edit/inspector-controls/index.js
priethor: Agreed. As a fan of ternary operators, I don't know why I didn't start with it in the first place. ✅

ciampo:
> I think we can move forward with it but I'm actually wondering if this is something we should generalize in the 	`SelectComponent` (like for all selects) or maybe make it opt-in via a prop or something? Curious to hear @ciampo's thoughts.

I wold prefer not to add such a feature to `SelectControl`.

In my opinion, low-level components should provide a focused, flexible, and modular primitive, thus allowing some powerful composition patterns. Composing those puzzle pieces together should be done externally by the consumer of the package (cc @WordPress/gutenberg-components )

PR #63376, opened 2024-07-10
Button: Never apply `aria-disabled` to anchor

### LINKED ISSUE #62281: Button component: link buttons should not use aria-disabled

mirka:
Fixes #62281

## What?

Prevents link (anchor) buttons from ever having an `aria-disabled` attribute.

## Why?

`<Button href="foo">` automatically renders an anchor element, whereas `<Button href="foo" disabled>` will remain a button element, because anchor elements cannot be disabled. Neither `disabled` nor `aria-disabled` are valid attributes on an anchor.

However, there was a bug where `<Button href="foo" disabled accessibleWhenDisabled>` rendered an anchor with an `aria-disabled` attribute.

## Testing Instructions

See the Storybook for `Button` and see what happens when you render `<Button href="foo" disabled accessibleWhenDisabled>`.

✅ New unit test should also pass.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+-   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
+
 ### Internal
 
 -   `CustomSelectControlV2`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
--- a/packages/components/src/button/index.tsx
+++ b/packages/components/src/button/index.tsx
@@ -162,7 +162,7 @@ export function UnforwardedButton(
 	} );
 
 	const trulyDisabled = disabled && ! accessibleWhenDisabled;
-	const Tag = href !== undefined && ! trulyDisabled ? 'a' : 'button';
+	const Tag = href !== undefined && ! disabled ? 'a' : 'button';
 	const buttonProps: ComponentPropsWithoutRef< 'button' > =
 		Tag === 'button'
 			? {
--- a/packages/components/src/button/test/index.tsx
+++ b/packages/components/src/button/test/index.tsx
@@ -542,6 +542,19 @@ describe( 'Button', () => {
 
 			expect( screen.getByRole( 'button' ) ).toBeVisible();
 		} );
+
+		it( 'should become a button again when disabled is supplied, even with `accessibleWhenDisabled`', () => {
+			render(
+				<Button
+					// @ts-expect-error - a button should not have `href`
+					// eslint-disable-next-line no-restricted-syntax
+					href="https://wordpress.org/"
+					disabled
+					accessibleWhenDisabled
+				/>
+			);
+			expect( screen.getByRole( 'button' ) ).toBeVisible();
+		} );
 	} );
 
 	describe( 'ref forwarding', () => {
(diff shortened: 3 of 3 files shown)


DaniGuardiola (approved):
Makes sense to me and work as intended!

Random idea for discussion, tangential to this PR, but do you think it'd be possible to use an Ariakit button underneath? We might be able to delegate many of these behaviors (like accessibleWhileDisabled), plus get a few potential safeguards or a11y features out of the box and reduce our maintenance area. Not sure if there'd be any significant trade-offs.

DaniGuardiola (approved):
Makes sense to me and work as intended!

Random idea for discussion, tangential to this PR, but do you think it'd be possible to use an Ariakit button underneath? We might be able to delegate many of these behaviors (like accessibleWhileDisabled), plus get a few potential safeguards or a11y features out of the box and reduce our maintenance area. Not sure if there'd be any significant trade-offs.

DaniGuardiola:
Github decided to duplicate my review 😆

tyxla (approved):
Makes sense. Thanks for making this behavior clearer through a unit test 👍

ciampo:
> We might be able to delegate many of these behaviors (like accessibleWhileDisabled), plus get a few potential safeguards or a11y features out of the box and reduce our maintenance area. Not sure if there'd be any significant trade-offs.

That is an alluring idea, although `Button` is such a complex and ubiquitous component that we'd need to do extensive testing, both automated (behavioral and visual) and human, before making such a change. I wonder if it just makes more sense to think about a v2 of `Button` instead — although likely neither of those projects would be high priority.

mirka:
Basically agree with @ciampo. But my instinct is that Ariakit is not going to make a big difference here, because most of the complexity of Button is in the prop bloat and the styles 😅

Issue #40941, opened 2022-05-09
Dynamically generate Query variations for registered post types


jameskoster:
It was mentioned in https://github.com/WordPress/gutenberg/pull/39837#issuecomment-1121029310 that it would be good to dynamically generate Query variations based on existing post types.

In addition to generating the variations themselves, I hope that we can use the design concepts in https://github.com/WordPress/gutenberg/issues/32268 to also dynamically populate those variations with the appropriate filter options, based on the taxonomies attached to the post type.

So if I register a "Recipes" post type with categories "Pizza Dough" and "Pizza Sauce", I should find a "Recipes List" block which includes options to filter the posts it displays based on those categories.

ntsekouras:
👋 - I've been toying with this and was wondering what convention we should use for the variation name's. The main problem comes from `pages`, as we already have a `Page List` block and is navigation related..

In addition I was thinking that if we have such variations in place we should also hide at least the `post type` control. Then maybe add some block transformation options to switch the post type if someone wants to? 🤔

jameskoster:
Good questions. 

> The main problem comes from pages, as we already have a Page List block and is navigation related..

An initial thought here is that Page List could have a different name, especially as it renders as a _row_ rather than a _list_ when placed in the Navigation block. "Page Links" perhaps?

Continuing that line of thinking, perhaps we should avoid "List" altogether here... Queries are often arranged into grids after all. "Pages", "Posts", "Products" might be too simple, but are probably a good place to start?

> In addition I was thinking that if we have such variations in place we should also hide at least the post type control. Then maybe add some block transformation options to switch the post type if someone wants to?

I see this work as a precursor to simplifying the base Query block UX. I still believe that it should be a template-only block that always inherit from the template, and is most likely removal-locked. The variations should feel like discrete blocks, so _transforming_ between them seems like a good idea 👍 Their use cases are more additive than foundational. E.g. "I want to display some featured products in this blog post" or "I want to list my 5 latest posts in my site footer".

ntsekouras:
>Their use cases are more additive than foundational. E.g. "I want to display some featured products in this blog post" or "I want to list my 5 latest posts in my site footer".

I'm exploring how extenders could make the most of it for specific use cases like that, but we still need to handle the 'plain' variations that would restrict results to just a specific post type.

>An initial thought here is that Page List could have a different name, especially as it renders as a row rather than a list when placed in the Navigation block. "Page Links" perhaps?

[...]

PR #63380, opened 2024-07-10
Query Loop Block: add variations for each post type

ellatrix:
## What?

Adds a query loop block variation for each post type so it shows up in the inserter when searching for "posts", "pages", "portfolio" (custom post types), etc.

Also adds the "blog" keyword for the post variation.

image

## Why?

It doesn't show up if you search for it with these words.

## Testing Instructions

Search the inserter.

## Screenshots or screencast 

Another nice thing is that you get a preview for that particular post type. For example I had no posts a this site, but I did have pages.

query-result

### DIFF
--- a/packages/block-library/src/query/index.js
+++ b/packages/block-library/src/query/index.js
@@ -71,7 +71,7 @@ export const settings = {
 	deprecated,
 };
 
-const previousPostTypes = [];
+let previousPostTypes = [];
 let haltSubscribing = false;
 
 const keywords = {
@@ -112,11 +112,17 @@ subscribe( () => {
 						postType: postType.slug,
 					},
 				},
-				keywords: keywords[ postType.slug ] ?? [],
+				keywords: [
+					postType.labels.name,
+					postType.labels.menu_name,
+					...( keywords[ postType.slug ] ?? [] ),
+				],
 			} );
 			haltSubscribing = false;
 		}
 	}
+
+	previousPostTypes = filteredPostTypes;
 } );
 
 export const init = () => initBlock( { name, metadata, settings } );
--- a/packages/block-library/src/query/index.js
+++ b/packages/block-library/src/query/index.js
@@ -2,6 +2,13 @@
  * WordPress dependencies
  */
 import { loop as icon } from '@wordpress/icons';
+import { subscribe, select } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
+import {
+	registerBlockVariation,
+	unregisterBlockVariation,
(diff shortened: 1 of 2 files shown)


packages/block-library/src/query/index.js
```diff
@@ -64,4 +71,58 @@ export const settings = {
...
+			registerBlockVariation( name, {
+				name: postType.slug,
+				title: sprintf(
+					/* translators: %s: post type */
+					__( '%s Query Loop' ),
+					postType.labels.singular_name
+				),
+				attributes: {
+					query: {
+						postType: postType.slug,
```
youknowriad, line 112: I think this PR is going to create more issues that it solves, mainly because if you select a post type, you can't set the query to "inherit" true.

Mamaduka:
We can use server-side variation registration here. All post-type data is already available on the server.

See the Post Terms block example: https://github.com/WordPress/gutenberg/blob/f699d2b87292cd2fb3898f8814c337e7442d8e03/packages/block-library/src/post-terms/index.php#L125

ntsekouras:
There is an old issue for this: https://github.com/WordPress/gutenberg/issues/40941. There were some nuances discussed there..

packages/block-library/src/query/index.js
ntsekouras: There are more nuances than `inherit`. An example is the `isActive`, existing Posts List variation, Pages List block etc.. Also I'm pretty confident that we used to need the other `query` attributed due to the way we handle `object` attributes in general..

ellatrix: Ok, then maybe we need to change the name depending on the main query that the current template is on.

ellatrix: Also, there's already a "Posts List" variation, we should probably remove that then

ellatrix: Removed it here: #63404

ntsekouras: If we'd remove the Posts List variation, it should be done here and definitely not in it's own PR without having the alternative landed..

Please see the issue I linked above for some more nuances.
Additionally as @Mamaduka also mentions, this should be done server side and avoid all this complexity in the client and the subscription.

youknowriad: Also see #63353 description where I share some thoughts on this.

ellatrix: I was going to close this PR because of the issues? Not sure if it makes sense to add the variations at all, if it has to inherit the query, we should probably dynamically name it based on what the main query is.

Issue #60469, opened 2024-04-04
Lightbox: caption missing when opening lightbox image


mamaretti:
### Description

hello 

since #58835 figcaption is missing on opened lightbox image

(WP 6.5)

### Step-by-step reproduction instructions

add an image with caption and trigger lightbox link

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

annezazu:
@artemiomorales can you look into this?

annezazu:
Pulling in feedback and a question from a separate issue:

>I am the developer of an image caption plugin for publishers who need to show image attributions for legal reasons.
With WordPress 6.5, lightboxes no longer show the caption I added to images in the post content since lightbox HTML is not rendered separately.

>Ideally, I could add a placeholder for the caption into block_core_image_print_lightbox_overlay() using a filter, but there is none.
Would you be able to provide one?

>In a next step, I would need to figure out how to fill it, but the render_block_core/image filter might just be good for that.

artemiomorales:
@annezazu Thanks for the ping! I'm not sure I have bandwidth to look at this for the upcoming release. To clarify @webzunft, do you need to render the caption inside the lightbox overlay itself? And this was working prior to 6.5?

webzunft:
@artemiomorales 
> do you need to render the caption inside the lightbox overlay itself? And this was working prior to 6.5?

Yes and yes. In 6.4, I just had to place my caption markup close to the image and it was showing up with it in the lightbox.

gziolo:
It might be related to this refactoring that uses a shared HTML for all Lightbox overlays: 
- https://github.com/WordPress/gutenberg/pull/58835

andreawetzel:
Our photographers would also like the caption to appear when the lightbox is expanded for an image.

t-hamano:
From what I've tested, even in WP 6.4, lightbox images don't display captions because the figcaption is hidden with CSS:

https://github.com/WordPress/gutenberg/blob/wp/6.4/packages/block-library/src/image/style.scss#L276-L278

dan-zakirov:
Opened PR #77477 which restores the caption inside the lightbox overlay and adds two Gallery-level toggles (`Show in gallery` / `Show in lightbox`) so authors can control where the caption appears.

[...]

PR #63381, opened 2024-07-10
Image lightbox: Remove duplicate image when lightbox is opened

### LINKED ISSUE #55960: Image: Duplicate image is visible when opening lightbox
### Description

Currently, the lightbox works by creating a duplicate image and animating it. However, the original image is still visible for a moment while the lightbox is opening.

We should hide the original image on click to improve this experience.

1. Add an image to a post
2. Enable `Expand on click` in the block settings
3. Publish and view the post; click on the image
4. See that the image in the content persists for a moment as a duplicate image zooms in

https://github.com/WordPress/gutenberg/assets/5360536/81ba21d7-f6a3-4dbe-9595-3dd8bc319596

image

### Environment info

_No response_

Yes

Yes

artemiomorales:
## What?

This PR hides the duplicate image that shows when the lightbox is opened, reported in https://github.com/WordPress/gutenberg/issues/55960

## Why?

It's a better user experience if the image is not duplicated.

## How?

It sets the visibility to `hidden` on the content image when the lightbox is opened, and uses CSS keyframes to set the visibility back to `visible` at the appropriate time when the lightbox is closed.

## Testing Instructions

1. Add an image to a post. I recommend aligning the image to the right or left and setting it to a smaller size to better observe the content image.
2. Click on the image link settings and enable `Expand on Click`
3. Publish and view the post
4. Click on the image

Please also test on mobile and in different browsers.

### Before

https://github.com/user-attachments/assets/7dd46674-c34e-4918-9c50-2679285e0d3c

### After

https://github.com/user-attachments/assets/b64e31d4-747b-46d9-ba54-2e6dd92b2236

### DIFF
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -220,8 +220,6 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 			JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
 		)
 	);
-	$p->set_attribute( 'data-wp-class--hide-content', 'state.isContentHidden' );
-	$p->set_attribute( 'data-wp-class--show-content', 'state.isContentVisible' );
 
 	// Image.
 	$p->next_tag( 'img' );
@@ -232,6 +230,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
+	$p->set_attribute( 'data-wp-class--hide', 'state.isContentHidden' );
+	$p->set_attribute( 'data-wp-class--show', 'state.isContentVisible' );
 
 	$body_content = $p->get_updated_html();
 
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -230,6 +230,7 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
+	$p->set_attribute( 'data-wp-class--hide', 'state.overlayOpened' );
 
 	$body_content = $p->get_updated_html();
 
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -230,8 +230,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
-	$p->set_attribute( 'data-wp-class--hide', 'state.contentImageHide' );
-	$p->set_attribute( 'data-wp-class--show', 'state.contentImageShow' );
+	$p->set_attribute( 'data-wp-class--hide', 'state.isContentHidden' );
+	$p->set_attribute( 'data-wp-class--show', 'state.isContentVisible' );
 
 	$body_content = $p->get_updated_html();
 
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -230,7 +230,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
-	$p->set_attribute( 'data-wp-class--hide', 'state.overlayOpened' );
+	$p->set_attribute( 'data-wp-class--hide', 'state.contentImageHide' );
+	$p->set_attribute( 'data-wp-class--show', 'state.contentImageShow' );
 
 	$body_content = $p->get_updated_html();
 
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -220,6 +220,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 			JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
 		)
 	);
+	$p->set_attribute( 'data-wp-class--hide-content', 'state.isContentHidden' );
+	$p->set_attribute( 'data-wp-class--show-content', 'state.isContentVisible' );
 
 	// Image.
 	$p->next_tag( 'img' );
@@ -230,8 +232,6 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
-	$p->set_attribute( 'data-wp-class--hide', 'state.isContentHidden' );
-	$p->set_attribute( 'data-wp-class--show', 'state.isContentVisible' );
 
 	$body_content = $p->get_updated_html();
 
--- a/packages/block-library/src/image/style.scss
+++ b/packages/block-library/src/image/style.scss
@@ -9,6 +9,10 @@
 		max-width: 100%;
 		vertical-align: bottom;
 		box-sizing: border-box;
+
+		&.hide {
+			opacity: 0;
+		}
 	}
 
 	// The following style maintains border radius application for deprecated
--- a/packages/block-library/src/image/style.scss
+++ b/packages/block-library/src/image/style.scss
@@ -10,12 +10,14 @@
 		vertical-align: bottom;
 		box-sizing: border-box;
 
-		&.hide {
-			visibility: hidden;
-		}
+		@media (prefers-reduced-motion: no-preference) {
+			&.hide {
+				visibility: hidden;
+			}
 
-		&.show {
-			animation: show-content-image 0.4s;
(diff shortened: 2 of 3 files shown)


artemiomorales:
At first glance, this issue seems pretty easy — just set the image in the content to `opacity: 0` when it's clicked.
However, unless we figure out how and when to set the opacity back to 1, the image flashes like so:

https://github.com/WordPress/gutenberg/assets/5360536/09006844-8f2c-40ef-88e2-89f92be9f02d

packages/block-library/src/image/index.php
```diff
@@ -191,6 +191,7 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 		'data-wp-context',
 		wp_json_encode(
 			array(
+				'animationId'      => uniqid(),
```
artemiomorales: We need to set a unique value in the context to make sure we only hide the content image that is currently being clicked. Is `uniqid()` a good way to do that?

packages/block-library/src/image/index.php
```diff
@@ -215,6 +216,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
+	$p->set_attribute( 'data-wp-class--hide', 'state.contentImageHide' );
+	$p->set_attribute( 'data-wp-class--show', 'state.contentImageShow' );
```
madhusudhand: because these represent boolean values, lets call them something like `isContentHidden`, `isContentVisible` ?

madhusudhand (approved):
Changes tests great. Double image or a blink while switching between visible states isn't observed.

🚢

packages/block-library/src/image/index.php
madhusudhand: As this is more like unique id of the image, better name would be `imageId`. 
It is being introduced in this change https://github.com/WordPress/gutenberg/pull/63348/files#diff-1e4aa29ea183efda9c795271dc1acf6e8049355db104d0dca5b0ea5caf7f619cR218

gziolo: `imageId` is also an existing attribute, so I would recommend using something more tied to the behavior of the overlay that is getting animated.

luisherranz (changes requested):
This looks good to me, but let's wait until this other pull request is merged and do this one on top:

- https://github.com/WordPress/gutenberg/pull/63348

Additionally, let's try to reuse the unique ID that we are already generating in that pull request.

luisherranz:
The pull request that adds unique ids to the images has already been merged: 

- https://github.com/WordPress/gutenberg/pull/63348

@artemiomorales, can you refactor this pull request on top of the other and see if we can use the unique id we are already generating?

packages/block-library/src/image/index.php
artemiomorales: @madhusudhand Ok that's been changed 👍

artemiomorales:
> can you refactor this pull request on top of the other and see if we can use the unique id we are already generating?

@luisherranz Done!

packages/block-library/src/image/index.php
```diff
@@ -230,6 +230,8 @@ function block_core_image_render_lightbox( $block_content, $block ) {
 	// contain a caption, and we don't want to trigger the lightbox when the
 	// caption is clicked.
 	$p->set_attribute( 'data-wp-on-async--click', 'actions.showLightbox' );
+	$p->set_attribute( 'data-wp-class--hide', 'state.isContentHidden' );
+	$p->set_attribute( 'data-wp-class--show', 'state.isContentVisible' );
```
SantosGuillamot, line 234: I assume we can't use `hidden` or `!state.isContentHidden` instead of having different properties because of the animations. Is that right?

packages/block-library/src/image/view.js
```diff
@@ -59,6 +59,19 @@ const { state, actions, callbacks } = store(
 				const { imageId } = getContext();
 				return state.metadata[ imageId ].imageButtonTop;
 			},
+			get isContentHidden() {
+				const ctx = getContext();
+				return (
+					state.overlayEnabled && state.currentImageId === ctx.imageId
```
SantosGuillamot, line 65: Not sure if it could be used elsewhere, but maybe in the future we could consider moving `state.currentImageId === ctx.imageId` to a new `isCurrentImage` derived state property?

It's true that right now, it is only used in `isContentHidden` and `isContentVisible` if I am not mistaken.

SantosGuillamot (approved):
It seems to be working on my testing 🙂 I just left a couple of small comments to understand it better.

madhusudhand (commented):
It works good with image and gallery block.

Not directly related to this, 
Observed a small issue with images that have a caption, in a gallery block.
There is a slight shadow left on the background.

Steps:
* Add a gallery block and set captions for few images.
* Set the scroll position of the page to a place where image bottom is near window bottom.
* Now click on the image (with caption)
* Observe that there is a shadow in place of original image.

double-image

luisherranz:
> Observed a small issue with images that have a caption, in a gallery block

Was this introduced in this pull request or did it exist before?

luisherranz (approved):
I haven't tested it, but codewise it looks good to me 🙂

packages/block-library/src/image/index.php
artemiomorales: What would `hidden` be in this case? A class, or a directive of some kind?

As far as `!state.isContentHidden` goes, yes doing that causes the animation to trigger on page load, which we don't want. I suppose we could introduce some intermediary logic to ensure the animation only triggers after initialization, but it seems to me that would complicate the code for little gain. What do you think?

packages/block-library/src/image/view.js
artemiomorales: > Not sure if it could be used elsewhere, but maybe in the future we could consider moving state.currentImageId === ctx.imageId to a new isCurrentImage derived state property?

That sounds good to me as a future enhancement. Right now the code is a little verbose with `isContentHidden` and `isContentVisible` right beside each other both using the same logic, but I think the intent is clear and it's easy to see the contrast between them without much fuss.

packages/block-library/src/image/style.scss
```diff
@@ -11,6 +11,30 @@
...
+				animation: show-content-image 0.4s;
+			}
+			figcaption {
+				.wp-block-gallery & {
+					animation: show-content-image 0.4s;
+				}
+			}
+		}
+	}
+
```
artemiomorales: I'm unsure if we should put the styles to hide/show the caption here or in the gallery styles. My inclination is to put them here because we need to animate the caption with the same timing as the image. What do you think?

Note that we need to restrict these caption styles to the gallery to prevent hiding captions in other circumstances when they don't overlay the image.

artemiomorales (commented):
> Was this introduced in this pull request or did it exist before?

@luisherranz @madhusudhand Previously, the caption faded out at the same rate as the duplicate image, so it wasn't an issue. Effectively, this pull request is introducing the problem.

I've updated the code, putting the class name on the parent `figure` instead. I've also highlighted part of the fix below, though I'm unsure if this is the best approach. Would appreciate any feedback 🙏

https://github.com/user-attachments/assets/4bd229e4-322e-49bb-af35-5d13be0849d6

packages/block-library/src/image/index.php
SantosGuillamot: It works for me 🙂 I was just trying to understand it better.

packages/block-library/src/image/style.scss
SantosGuillamot: What is the desired behavior with captions here? Should they be handled different for gallery vs normal images?

Right now this is **how galleries are working in this pull request**:

https://github.com/user-attachments/assets/4f9032e6-e998-48b2-ad25-1dd86a7831e1

And this is **how normal images work**:

https://github.com/user-attachments/assets/529a8f91-ce15-40ee-a4d2-b0b9feed4e5a


This is **how they would work if we hide the caption as in the galleries**:

https://github.com/user-attachments/assets/e0874bd7-4904-4bf0-bdf3-2408fefa495a



So, what is the expected UX? Should we even add a "fade-in" effect for the caption as well?

gziolo: It is a separate issue to handle the captions correctly. I'm also wondering what the rationale was for omitting the caption in the expanded image. There are reports that the caption should be included for legal reasons on some websites https://github.com/WordPress/gutenberg/issues/60469#issuecomment-2045759484. So that at least should be an option to configure for the lightbox.

artemiomorales: > So, what is the expected UX? Should we even add a "fade-in" effect for the caption as well?

Ok got it, so we need to account for the caption in galleries in two scenarios:
1. When the caption overlays the image
2. When the caption *does not* overlay the image

I only took into account the first issue, not the second.

In the first scenario, like in the latest fix, I think we could make the caption disappear immediately when the lightbox opens, then make it reappear at the proper time when the lightbox is closed. We could also introduce a fade and sequence it properly with the zooming animation to make it more elegant.

For the second scenario, the caption should fade out and fade in smoothly like the rest of the content outside of the image.

> I'm also wondering what the rationale was for omitting the caption in the expanded image. There are reports that the caption should be included for legal reasons on some websites https://github.com/WordPress/gutenberg/issues/60469#issuecomment-2045759484. So that at least should be an option to configure for the lightbox.

I believe there was conversation with design that the experience of a single image lightbox doesn't require a caption because one already has the context to understand what's being displayed. This is in contrast to a lightbox gallery, in which you don't have as much context.

> It is a separate issue to handle the captions correctly. 

Ok, I'll revert the most recent change, merge, and open a separate issue 👍

artemiomorales:
@madhusudhand @gziolo @SantosGuillamot I [opened a new issue](https://github.com/WordPress/gutenberg/issues/63707) to continue the discussion regarding the caption behavior.

PR #53889, opened 2023-08-23
Components: Make the Popover.Slot optional

youknowriad:
Related #53874

## What and why?

Gutenberg can be used as a platform/framework to build block editors. Mainly thanks to the @wordpress/block-editor package. That said, the experience today is not as straightforward as it can be. There can be a lot of small gotchas and hacks you need to do in order to achieve the desired result. One of these small things is the need to manually render `Popover.Slot` somewhere in the React tree.

I think ideally, Popover components should work properly out of the box without any provider or slot, and only if we really want to control their position in the DOM, we should use `Popover.Slot`.

This PR does that by using an "automatic container" for all popovers that is appended to the end of the "body" of the document if there's no slot that is defined manually using `Popover.Slot`. I think that's a better default than inlining popovers which breaks all the styles and more.

**Todo:**

 - [x] I've noticed that when the anchor is within an iframe, the popover is not offset properly to the position of the iframe (you can test for instance the appender of the navigation block). I'm not entirely sure how to fix that because I thought we had that offsetting in place already. cc @talldan @ciampo 

## Testing Instructions

1- Check that the stories updates (removal of popover and slot provider) still work properly.
2- Navigate the editor and ensure popovers, tooltips, drodowns are shown properly.

PR #63388, opened 2024-07-11
Storybook: Improve TypeScript performance for slow stories

mirka:
## What?

Optimizes some low-hanging bottlenecks in the TypeScript compilation performance for Storybook stories, as revealed by [`@typescript/analyze-trace`](https://github.com/microsoft/typescript-analyze-trace).

### Improvement in cold build time

```sh
# Before
npx tsc --build --force  145.83s user 10.52s system 132% cpu 1:58.13 total

# After
npx tsc --build --force  136.40s user 9.36s system 134% cpu 1:48.61 total
```

Honestly the build times are not very stable so I'm not sure how much we can read into it, but the analyzed output of `@typescript/analyze-trace` clearly shows improvements, so I think we can say that we're moving things in the correct direction.

## Why?

These particularly slow stories have TS performance issues in the components themselves, and we ultimately need to address the root cause. However, their Storybook stories can be optimized to "bypass" some of the particularly slow parts. Luckily, this mostly just involves converting the story format from [CSF2 to CSF3](https://storybook.js.org/docs/7/api/csf#upgrading-from-csf-2-to-csf-3), with a bit of attention given to minimizing unnecessary type checking work. And CSF3 is good for us to transition to anyway.

## Testing Instructions

Smoke test the affected stories to see that there are no regressions.

Code review should be easier with whitespace hidden.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -388,6 +388,14 @@ module.exports = {
 			},
 		},
 		{
+			files: [ '**/@(storybook|stories)/*' ],
+			rules: {
+				// Suggested workaround for hooks used in CSF3 format stories.
+				// https://github.com/storybookjs/eslint-plugin-storybook/pull/149
+				'react-hooks/rules-of-hooks': 'off',
+			},
+		},
+		{
 			files: [ 'packages/components/src/**' ],
 			excludedFiles: [
 				'packages/components/src/utils/colors-values.js',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -388,14 +388,6 @@ module.exports = {
 			},
 		},
 		{
-			files: [ '**/@(storybook|stories)/*' ],
-			rules: {
-				// Suggested workaround for hooks used in CSF3 format stories.
-				// https://github.com/storybookjs/eslint-plugin-storybook/pull/149
-				'react-hooks/rules-of-hooks': 'off',
-			},
-		},
-		{
 			files: [ 'packages/components/src/**' ],
 			excludedFiles: [
 				'packages/components/src/utils/colors-values.js',
--- a/packages/components/src/color-picker/stories/index.story.tsx
+++ b/packages/components/src/color-picker/stories/index.story.tsx
@@ -1,12 +1,7 @@
 /**
  * External dependencies
  */
-import type { Meta, StoryFn } from '@storybook/react';
-
-/**
- * WordPress dependencies
- */
-import { useState } from '@wordpress/element';
+import type { Meta, StoryObj } from '@storybook/react';
 
 /**
  * Internal dependencies
@@ -30,19 +25,4 @@ const meta: Meta< typeof ColorPicker > = {
 };
 export default meta;
 
-const Template: StoryFn< typeof ColorPicker > = ( { onChange, ...props } ) => {
-	const [ color, setColor ] = useState< string | undefined >();
-
-	return (
-		<ColorPicker
-			{ ...props }
-			color={ color }
-			onChange={ ( ...changeArgs ) => {
-				onChange?.( ...changeArgs );
-				setColor( ...changeArgs );
-			} }
-		/>
-	);
-};
-
-export const Default = Template.bind( {} );
+export const Default: StoryObj< typeof ColorPicker > = {};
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -387,6 +387,14 @@ module.exports = {
 				'jsdoc/require-param': 'off',
 			},
 		},
+		{
+			files: [ '**/@(storybook|stories)/*' ],
+			rules: {
+				// Suggested workaround for hooks used in CSF3 format stories.
+				// https://github.com/storybookjs/eslint-plugin-storybook/pull/149
+				'react-hooks/rules-of-hooks': 'off',
```
mirka: In CSF3, a functional component can be defined under the [`render` property](https://storybook.js.org/docs/7/api/csf#default-render-functions) of the story object. This does not look like a proper component to the `rules-of-hooks` linter, so it errors if you use a hook inside.

packages/components/src/color-picker/stories/index.story.tsx
```diff
@@ -30,19 +25,4 @@ const meta: Meta< typeof ColorPicker > = {
...
-			onChange={ ( ...changeArgs ) => {
-				onChange?.( ...changeArgs );
-				setColor( ...changeArgs );
-			} }
-		/>
-	);
-};
-
-export const Default = Template.bind( {} );
+export const Default: StoryObj< typeof ColorPicker > = {};
```
mirka, line 28: Switched to an uncontrolled story for simplicity.

packages/components/src/dropdown-menu/stories/index.story.tsx
```diff
@@ -43,39 +43,42 @@ const meta: Meta< typeof DropdownMenu > = {
...
-		{
-			title: 'Second Menu Item Label',
-			icon: arrowDown,
-			// eslint-disable-next-line no-console
-			onClick: () => console.log( 'down!' ),
-		},
+export const Default: StoryObj< typeof DropdownMenu > = {
+	decorators: [
+		( Story ) => (
+			<div style={ { height: 150 } }>
```
mirka, line 49: I noticed these explicit heights are no longer necessary due to #53889, because popovers are now rendering at the end of the document rather than within the each story div, which used to cut them off visually. I'll do a separate cleanup PR.

.eslintrc.js
DaniGuardiola: You can simply pass a named function with a name that starts with uppercase and it will be recognized as a component. I do this in my own project. Could we try that?

I normally just go for `function Render`. See archetype-ui stories for reference, if you need to.

DaniGuardiola (commented):
This looks correct to me and runs as expected, however see note.

.eslintrc.js
tyxla: Agreed. If we can avoid adding extra complexity to the ESLint configuration with a valid alternative syntax, let's do it.

mirka: Sure, we can go with that! (Interestingly, most of the existing hooks were moved into decorator functions, which are anonymous and don't trigger the rule-of-hooks lint 😳)

Reverted in 24569a5

tyxla: Yep. Was going to say this. Just came here running a lint without the ESLint changes locally and was surprised that there were no errors.

packages/components/src/dropdown-menu/stories/index.story.tsx
```diff
@@ -43,39 +43,42 @@ const meta: Meta< typeof DropdownMenu > = {
...
 
-export const WithChildren = Template.bind( {} );
-// Adding custom source because Storybook is not able to show the contents of
-// the `children` prop correctly in the code snippet.
-WithChildren.parameters = {
-	docs: {
-		source: {
-			code: `<DropdownMenu label="Select a direction." icon={ more }>
+export const WithChildren: StoryObj< typeof DropdownMenu > = {
+	...Default,
```
tyxla, line 75: I think we'll need to erase the controls from the default story:

```suggestion
	...Default,
	controls: null,
```

without this, we'll see the controls from the default story and I don't think we want this:

Screenshot 2024-07-12 at 14 08 13

tyxla (approved):
Good to go once we reset the `DropdownMenu` story with children controls 👍 

Nice improvement @mirka 🚀

packages/components/src/dropdown-menu/stories/index.story.tsx
mirka: Good catch, thanks 🙏

packages/components/src/dropdown-menu/stories/index.story.tsx
mirka: Cleanup at #63480

PR #61465, opened 2024-05-08
Zoom out: maintain scroll position

ellatrix:
## What?

When toggling zoom out, we should maintain the scroll position to the currently visible content.

It's not perfect, but better than trunk.

Closes https://github.com/WordPress/gutenberg/issues/65884

## Why?

Because it's disorienting be scrolled to the top when zoomed in, given the intent of the double click is to edit the section which was just in view before double click,

## How?

Try to compute scroll position after the zoom out is exited getting the previous scale and scroll and applying them to the new scale.

## Testing Instructions

- enable zoom out
- double click a section
- you should be seeing the section you double clicked, at normal scale

## Screenshots or screencast 

https://github.com/user-attachments/assets/0453bb3a-82ea-4896-8d25-9129d3de9dc0

PR #61424, opened 2024-05-07
Zoom out: keep original viewport width (single scale)

ajlende:
Alternative to #61422

## What?

Makes the zoom out view show a full-width page.

## Why?

This avoids layout shifting when you zoom out.

## How?

- Adds a div outside the iframe to make the CSS `100vw`.
- Adds a div outside that to measure the container and provide an overflow hidden.
- Calculate the scale based on `100vw` and the container width.
- Adds an 800px max width for margins instead of the custom function.

Issue #61093, opened 2024-04-25
Zoom-out can trigger double scrollbars in the editor canvas.


youknowriad:
### Description

Double scrollbars are visible in the editor canvas when you scroll while zoom-out mode is active.

### Step-by-step reproduction instructions

If you're on MacOS, make sure to have the "scrollbars always visible" setting active.

1- Enable the zoom-out mode experiment.
2- Go the site editor.
3- Edit the home page of 2024.
4- Trigger the zoom-out mode.
5- Notice that if you scroll in the canvas, you'll start seeing two scrollbars. And on Safari, they persist even after leaving zoom-out mode.

### Screenshots, screen recording, code snippet

<img width="1470" alt="Screenshot 2024-04-25 at 7 05 00 AM" src="https://github.com/WordPress/gutenberg/assets/272444/68e34d80-6f20-474f-bf52-9d51fab5c4ea">

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

mikachan:
I don't think this is related to zoom-out, or it's at least not only triggered by zoom-out mode, as I can reproduce this with the inserter as well:

<img width="1439" alt="image" src="https://github.com/WordPress/gutenberg/assets/1645628/3e055742-9c76-4be8-abd0-f2a2a4c98375">

<img width="1438" alt="image" src="https://github.com/WordPress/gutenberg/assets/1645628/9d60548d-950e-4053-bb96-59e1b231fea9">

The outer scrollbar is from a `NavigableRegion` with `interface-interface-skeleton__content` class. The inner scrollbar is from the editor canvas iframe. I would guess we want to prioritise the scrollbar from the iframe, as that should be the true height of the content we want to scroll.

ntsekouras:
👋 - It would be great if you can test [this PR](https://github.com/WordPress/gutenberg/pull/61548) to see if it fixes the issue you're mentioning here, because I couldn't reproduce your use case, but I've found a double scrollbar issue.. Thanks!

PR #59512, opened 2024-03-01
Expand main content area to viewport when zoomed out

ellatrix:
## What?

@richtabor asked if this was possible, so here it is. 

zoom-out-expand

## Why?

I think the primary reason is to make the drop target for patterns larger, but it also makes the site look better when empty (just header/footer).

## How?

1. We need to expand the canvas to fix the screen. This can only be done through the `body` element because it must be an element with the scaled element (`html` element). We can do this for any theme I think.
2. Now we need to push the footer down to the bottom of the body element / expand the main area to fit. This is the most tricky part because the styles may clash with existing theme styles. I wonder if we should make this opt-in: a theme should decide exactly how and what to expand/push down. This is actually useful on the front-end as well: if there's limited content, it's nice to push down the footer.

## Testing Instructions

Enable the zoom-out mode experiment.
Create an empty page in the site editor (for a block based theme).
Click the zoom-out mode button in the top toolbar.

PR #63390, opened 2024-07-11
Refactor Zoom Out to scale `iframe`

### LINKED ISSUE #49299: Site Editor: Endless scrolling in zoom out and style views when there are blocks with vh units
Might be related to [this comment](https://github.com/WordPress/gutenberg/issues/49120#issuecomment-1480614237)

### Description

If the content contains blocks with a height in units of `vh`, there will be endless scrolling in both the zoomed-out and style views. This problem occurs when the block has a large value, such as `100vh` high. iframe editor content will infinitely grow in height.

- Activate "Zoomed out view" from the Gutenberg > Experiments menu.
- Open the site editor.
- Insert a group block in one of the templates.
- Set min.Height to `100vh` from the Dimensions panel.
- Click the "zoom out" view button or access "Browse styles" menu.

https://user-images.githubusercontent.com/54422211/227163677-3bc7d832-8303-44b9-a65f-2b6b71ed2a12.mp4

### Environment info

- WordPress Version: 6.3-alpha-55582 
- Gutenberg Version: Latest trunk (15.4.0)

Yes

Yes

### LINKED ISSUE #64117: Zoom Out breaks sticky positioned elements
### Description

In the zoomed out view sticky positioning ceases to work. It’s due to the `transform` applied to the `html` element in the iframe.

#### Possible fixes
- Transform the `iframe` instead of the `html` element (etc. #63390)
- Or
  - Hide overflow on the html document
  - Style both `html` and `body` to be `height: 100%`
  - Style `body` to `overflow: auto`
  - Rework some more stuff (remove border on the `html`, etc.)

1. With a sticky positioned element in the page
2. Activate Zoom Out
3. Scroll the page and note that the element doesn’t stick

Here’s a sticky block you can paste into the editor:
```html
<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|10","bottom":"var:preset|spacing|10"}},"position":{"type":"sticky","top":"0px"}},"backgroundColor":"accent-4","layout":{"type":"constrained"}} -->
<div class="wp-block-group has-accent-4-background-color has-background" style="padding-top:var(--wp--preset--spacing--10);padding-bottom:var(--wp--preset--spacing--10)"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"center"}} -->
<p>We want to stick with you!</p>
<!-- /wp:paragraph -->

<!-- wp:buttons -->
<div class="wp-block-buttons"><!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link wp-element-button">Okay, lets stick together</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

https://github.com/user-attachments/assets/32299b15-2ca7-4840-a8c6-a5a8483875ee

### Environment info

WP 6.5.2 / Gutenberg trunk
Chrome / MacOS

- [X] Yes

- [X] Yes

### LINKED ISSUE #65080: Site Editor: Canvas width glitch when selecting 'Browse Styles' then closing inserter sidebar

1. Open the Site Editor and edit a page
2. Select the + button on the top bar to open the Inserter sidebar
3. Select the Styles button on the top bar to open the global styles sidebar
4. Select 'Browse Styles' in the global styles sidebar, observe the canvas zooms out
5. Now close the Inserter sidebar, observe that the canvas looks unusual with scrollbars moved into the middle of the viewport

If you compare this to engaging and disengaging 'Browse styles' without the inserter sidebar open, you'll see that it looks completely different, the scrollbars stay at the edge of the screen.

https://github.com/user-attachments/assets/1ce5135e-2b64-4acc-9574-50d6f8529b7f

### Environment info

Brave / Mac OS
Latest Gutenberg trunk

- [X] Yes

- [X] Yes

stokesman:
## What?

A refactor with fixes and improvements to “Zoom Out”.
- Keep the scrollbar inside the visible area
- Better centering of the canvas (currently it’s off-center only on devices whose scrollbars aren’t overlaid)
- Maintain scroll position (obviating #61465)
- Improve parity of layout (document width) from unzoomed to zoomed (this point is partly obviated given we land #65814, though I think code/style to do so here is a little simpler).
- Fix sticky positioning #64117 (yeah, we aren’t considering this essential for the moment, but I think it’s still valid for broader consideration).
- Fix width/scaling glitch with sidebar toggling #65080
- Simplify implementation

## Why?

- There is significant simplification that can be made
- There are some bugs that seen best solved by having zoom out scale the `iframe` instead of its `html` element. Most notably #64117 but this also fixes some things for which no issues exist:
  - Scroll position should be maintained between unzoomed/zoomed views
  - Document width should be maintained between unzoomed/zoomed views
  - Canvas should be centered when zoomed out

## How?

- Revises the CSS calculations for greater accuracy (and they happen to become simpler too).
- Adds some logic to prevent ongoing scale calculation after the initial zoom out has finished.
- Applies the transform to the `iframe` instead of the `html`

## Testing Instructions

TL:DR; Have the Zoom Out experiment enabled and test everything. You may find something I've missed. What follows are some specific scenarios that should be tested.

### `vh` units
There was a time when Zoom Out did scale the iframe and that was changed in #59334 apparently due to an issue with `vh` units #49299 causing the iframe editor content to infinitely grow in height. This branch doesn’t regress the issue from my testing. Here are the testing instructions from that issue.
> * Activate "Zoomed out view" from the Gutenberg > Experiments menu.
> * Open the site editor.
> * Insert a group block in one of the templates.
> * Set min.Height to `100vh` from the Dimensions panel.
> * Click the "zoom out" view button or access "Browse styles" menu.

Note the iframe’s content height is stable in this branch

### Parity of layout from unzoomed to zoomed
#61424 didn’t have an accompanying issue but it was made to:
> avoids layout shifting when you zoom out.

In my testing this branch improves layout parity.

1. With the inserter open to Patterns
2. Note the layout and text wrapping of the contents of the canvas
3. Open a pattern category in the inserter
4. Note the layout and text wrapping of the contents should be the same

### No double scrollbars
From issue #61093
1. Enable the zoom-out mode experiment.
2. Go the site editor and select a page to edit.
3. Trigger the zoom out mode by trying to insert a pattern from the inserter and after that resize the window width until you see the double scrollbar

This should not be reproducible on this branch. I reverted a bit of code from the PR merged to "fix" that issue.

### Maintained scroll position
1. With the inserter open to Patterns and a page with ample overflow
2. Scroll the page and note the position
3. Open a pattern category in the inserter
5. Note the scroll position is maintained

### Drag chip positioning over scaled iframe
Scaling the iframe does throw off the drag chip position when dragging over the iframe and required some changes to compensate. This was a thing experimented with previously #56889 when the scaling was applied to the iframe.
1. Open the global inserter and select a pattern category to initiate zoom out
2. Drag a block or pattern from the inserter over the scaled iframe
3. Note the position of the drag chip is as expected (the same relative to the pointer whether over the iframe or not).

### Expansion of main content area in Zoom Out
See #59512
1. Create an empty page in the site editor (for a block based theme).
2. Click the zoom-out mode button in the top toolbar.
3. Note the footer goes to the bottom of the available space so the content area appears as a big empty space.

### General canvas scrolling
This PR removes some wrapping elements whose introduction seem to have been the cause of some recent canvas scrollbar issues across different modes of the editor. They all seem to have been fixed so it’s important to test all permutations to make sure all content is scrollable and there are no extraneous scrollbars.
- non-iframed editor
  - Post editor with metaboxes (custom fields preference set to show)
  - Any editor with blocks < v3 registered (if the Gutenberg plugin is active this also depends on having a classic (non-block) theme active).
- Mobile or tablet preview
- Pattern editor
- All the above with notices (`wp.data.dispatch( 'core/notices' ).createInfoNotice( 'Notice' );`)

### Layout parity

| trunk | this branch |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/6a740a6e-e297-40fe-b17d-d6722f956131"/> | <video src="https://github.com/user-attachments/assets/e1a7bad0-ae96-4c08-86d6-e7e492b5108e"/> |

### DIFF
--- a/packages/base-styles/_animations.scss
+++ b/packages/base-styles/_animations.scss
@@ -36,11 +36,6 @@
 	@include reduce-motion("animation");
 }
 
-@mixin editor-canvas-resize-animation() {
-	transition: all 0.5s cubic-bezier(0.65, 0, 0.45, 1);
-	@include reduce-motion("transition");
-}
-
 // Deprecated
 @mixin edit-post__fade-in-animation($speed: 0.08s, $delay: 0s) {
 	@warn "The `edit-post__fade-in-animation` mixin is deprecated. Use `animation__fade-in` instead.";
--- a/packages/block-editor/src/components/block-canvas/index.js
+++ b/packages/block-editor/src/components/block-canvas/index.js
@@ -66,7 +66,7 @@ export function ExperimentalBlockCanvas( {
 	return (
 		<BlockTools
 			__unstableContentRef={ localRef }
-			style={ { height, display: 'flex' } }
+			style={ { height, display: 'flex', justifyContent: 'center' } }
 		>
 			<Iframe
 				{ ...iframeProps }
--- a/packages/block-editor/src/components/block-canvas/style.scss
+++ b/packages/block-editor/src/components/block-canvas/style.scss
@@ -4,5 +4,6 @@ iframe[name="editor-canvas"] {
 	height: 100%;
 	display: block;
 	background-color: transparent;
-	@include editor-canvas-resize-animation;
+	transition: all 0.5s cubic-bezier(0.65, 0, 0.45, 1);
+	@include reduce-motion("transition");
 }
--- a/packages/block-editor/src/components/block-canvas/style.scss
+++ b/packages/block-editor/src/components/block-canvas/style.scss
@@ -17,7 +17,6 @@ iframe[name="editor-canvas"].is-zoomed-out {
 	$scale: var(--wp-block-editor-iframe-zoom-out-scale);
 	$inset: var(--wp-block-editor-iframe-zoom-out-inset);
 	inset-block: $inset 0;
-	filter: drop-shadow(0 0 $grid-unit-05 rgba($color: #000, $alpha: 0.133)); // 0.133 = $gray-200 but with alpha.
 	transform: scale(#{$scale});
 	flex-basis: calc((100% - #{$inset} * 2) / #{$scale});
 	height: calc((100% - #{$inset} * 2) / #{$scale});
--- a/packages/block-editor/src/components/block-canvas/style.scss
+++ b/packages/block-editor/src/components/block-canvas/style.scss
@@ -16,10 +16,9 @@ iframe[name="editor-canvas"].is-zoomed-out {
 	$max-width: 750px; // This is also hardcoded as ZOOM_OUT_MAX_WIDTH in the iframe component and needs to be manually kept in sync.
 	$scale: var(--wp-block-editor-iframe-zoom-out-scale);
 	$inset: var(--wp-block-editor-iframe-zoom-out-inset);
-	inset-block: $inset 0;
 	transform: scale(#{$scale});
 	flex-basis: calc((100% - #{$inset} * 2) / #{$scale});
-	height: calc((100% - #{$inset} * 2) / #{$scale});
+	height: calc(100% / #{$scale});
 	$inset-max-width: "(#{$max-width} - #{$inset} * 2)";
 	max-width: calc(#{$inset-max-width} / #{$scale});
 }
--- a/packages/block-editor/src/components/block-canvas/style.scss
+++ b/packages/block-editor/src/components/block-canvas/style.scss
@@ -1,4 +1,7 @@
 iframe[name="editor-canvas"] {
+	position: relative;
+	inset-block: 0;
+	flex: 0 0 auto;
 	box-sizing: border-box;
 	width: 100%;
 	height: 100%;
@@ -6,4 +9,18 @@ iframe[name="editor-canvas"] {
 	background-color: transparent;
 	transition: all 0.5s cubic-bezier(0.65, 0, 0.45, 1);
 	@include reduce-motion("transition");
+	transform-origin: 50% 0;
+}
+
+iframe[name="editor-canvas"].is-zoomed-out {
+	$max-width: 750px; // This is also hardcoded as ZOOM_OUT_MAX_WIDTH in the iframe component and needs to be manually kept in sync.
+	$scale: var(--wp-block-editor-iframe-zoom-out-scale);
+	$inset: var(--wp-block-editor-iframe-zoom-out-inset);
+	inset-block: $inset 0;
+	filter: drop-shadow(0 0 $grid-unit-05 rgba($color: #000, $alpha: 0.133)); // 0.133 = $gray-200 but with alpha.
+	transform: scale(#{$scale});
+	flex-basis: calc((100% - #{$inset} * 2) / #{$scale});
+	height: calc((100% - #{$inset} * 2) / #{$scale});
+	$inset-max-width: "(#{$max-width} - #{$inset} * 2)";
+	max-width: calc(#{$inset-max-width} / #{$scale});
 }
--- a/packages/block-editor/src/components/block-list/content.scss
+++ b/packages/block-editor/src/components/block-list/content.scss
@@ -395,34 +395,36 @@ _::-webkit-full-page-media, _:future, :root .has-multi-selection .block-editor-b
 }
 
 .block-editor-block-list__zoom-out-separator {
-	/* same color as the iframe's background */
-	background: $gray-300;
-	// Additional -1px is required to avoid sub pixel rounding errors allowing background to show.
-	margin-left: -1px;
-	margin-right: -1px;
-	transition: background-color 0.3s ease;
+	color: $black;
 	display: flex;
 	align-items: center;
 	justify-content: center;
 	font-size: $default-font-size;
 	font-family: $default-font;
-	color: $black;
 	font-weight: normal;
+	.has-global-padding > & {
+		margin: 0;
+	}
 
 	.is-zoomed-out & {
 		// Scale the font size based on the zoom level.
-		font-size: calc(#{$default-font-size} * ( 2 - var(--wp-block-editor-iframe-zoom-out-scale)  ));
-	}
+		font-size: calc(#{$default-font-size} / var(--wp-block-editor-iframe-zoom-out-scale));
 
-	&.is-dragged-over {
-		background: $gray-400;
-	}
-}
+		& > * {
+			position: relative; // Puts the label in front of the ::before pseudo element.
+		}
 
-// In Post Editor allow the separator to occupy the full width by ignoring (cancelling out) the global padding.
-.has-global-padding > .block-editor-block-list__zoom-out-separator,
-.block-editor-block-list__layout.is-root-container.has-global-padding > .block-editor-block-list__zoom-out-separator {
-	max-width: none;
-	// Additional -1px is required to avoid sub pixel rounding errors allowing background to show.
-	margin: 0 calc(-1 * var(--wp--style--root--padding-right) - 1px) 0 calc(-1 * var(--wp--style--root--padding-left) - 1px) !important;
+		&::before {
+			background: $gray-300; // same color as the iframe's background.
+			content: "";
+			transition: opacity 0.2s linear;
+			position: absolute;
+			inset: auto 0; // Horizontal edges to the full canvas width regardless of global padding.
+			height: inherit;
+		}
+
+		&.is-dragged-over::before {
+			opacity: 0.25;
+		}
+	}
 }
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -101,10 +101,11 @@ export function ZoomOutSeparator( {
 		<AnimatePresence>
 			{ isVisible && (
 				<motion.div
-					as="button"
 					layout={ ! isReducedMotion }
 					initial={ { height: 0 } }
-					animate={ { height: '120px' } }
+					animate={ {
+						height: 'calc(60px / var(--wp-block-editor-iframe-zoom-out-scale)',
+					} }
 					exit={ { height: 0 } }
 					transition={ {
 						type: 'tween',
@@ -123,11 +124,20 @@ export function ZoomOutSeparator( {
 				>
 					<motion.div
 						initial={ { opacity: 0 } }
-						animate={ { opacity: 1 } }
+						animate={
+							isDraggedOver
+								? {
+										opacity: 0,
+										scale: isReducedMotion ? 1 : 0,
+										transition: { duration: 0.2, delay: 0 },
+								  }
+								: { opacity: 1 }
+						}
 						exit={ { opacity: 0 } }
 						transition={ {
 							type: 'tween',
 							duration: 0.1,
+							delay: 0.1,
 						} }
 					>
 						{ __( 'Drop pattern.' ) }
--- a/packages/block-editor/src/components/block-tools/zoom-out-popover.js
+++ b/packages/block-editor/src/components/block-tools/zoom-out-popover.js
@@ -24,7 +24,7 @@ export default function ZoomOutPopover( { clientId, __unstableContentRef } ) {
 		...popoverProps,
 		placement: 'left-start',
 		flip: false,
-		shift: true,
+		shift: { padding: { top: 48, bottom: 48 } },
 	};
 
 	return (
--- a/packages/block-editor/src/components/block-tools/zoom-out-popover.js
+++ b/packages/block-editor/src/components/block-tools/zoom-out-popover.js
@@ -24,7 +24,7 @@ export default function ZoomOutPopover( { clientId, __unstableContentRef } ) {
 		...popoverProps,
 		placement: 'left-start',
 		flip: false,
-		shift: { padding: { top: 48, bottom: 48 } },
+		shift: true,
 	};
 
 	return (
--- a/packages/block-editor/src/components/iframe/content.scss
+++ b/packages/block-editor/src/components/iframe/content.scss
@@ -13,8 +13,13 @@ iframe[name="editor-canvas"].is-zoomed-out {
 	margin-block: 0 calc(-100% / #{$scale});
 }
 
-.block-editor-iframe__body {
-	position: relative;
+.block-editor-iframe__html {
+	// Keeps body background styles from being clipped.
+	// See: https://github.com/WordPress/gutenberg/issues/44374#issuecomment-1260784743
+	height: auto;
+	// Keeps body background image from repeating.
+	// See: https://github.com/WordPress/gutenberg/issues/46782
+	min-height: 100%;
 }
 
 .block-editor-iframe__html.is-zoomed-out {
@@ -64,4 +69,8 @@ iframe[name="editor-canvas"].is-zoomed-out {
 	> body > .is-root-container:not(.wp-block-post-content) > main {
 		flex: 1;
 	}
+}
+
+.block-editor-iframe__body {
+	position: relative;
 }
--- a/packages/block-editor/src/components/iframe/content.scss
+++ b/packages/block-editor/src/components/iframe/content.scss
@@ -51,19 +51,17 @@ iframe[name="editor-canvas"].is-zoomed-out {
 		}
 	}
 
-	body {
+	// Expands main content area.
+	// See: https://github.com/WordPress/gutenberg/pull/59512
+	&,
+	> body,
+	> body > .is-root-container:not(.wp-block-post-content) {
 		display: flex;
 		flex-direction: column;
+		flex: 1; // Doesn’t effect the html element but grouping is simpler this way.
+	}
 
-		> .is-root-container:not(.wp-block-post-content) {
-			flex: 1;
-			display: flex;
-			flex-direction: column;
-			height: 100%;
-
-			> main {
-				flex: 1;
-			}
-		}
+	> body > .is-root-container:not(.wp-block-post-content) > main {
+		flex: 1;
 	}
 }
--- a/packages/block-editor/src/components/iframe/content.scss
+++ b/packages/block-editor/src/components/iframe/content.scss
@@ -24,39 +24,6 @@ iframe[name="editor-canvas"].is-zoomed-out {
 }
 
 .block-editor-iframe__html.is-zoomed-out {
-	$scale: var(--wp-block-editor-iframe-zoom-out-scale);
-
-	// Customize the scrollbar to counterscale its width but only for devices with fine
-	// pointers as otherwise scrollbars are likely to be overlaid and not affected by
-	// scaling. This isn’t perfect as some devices might have a fine cursor but still use
-	// overlaid scrollbars. Additionally, the width of the scrollbar reduces the width of
-	// the content area so the "default" scaling should factor it in to accurately keep
-	// the content width the same as when unzoomed.
-	@media (pointer: fine) {
-		&::-webkit-scrollbar {
-			// 15px seems a de facto standard but likely not all devices. It may
-			// be possible to derive the device size with some DOM measurements
-			// if this needs to be improved.
-			width: calc(15px / #{$scale});
-		}
-		&::-webkit-scrollbar-track {
-			background-color: $gray-100;
-			border-inline-start: calc(1px / #{$scale}) solid $gray-300;
-		}
-		&::-webkit-scrollbar-thumb {
-			background-color: $gray-400;
-			background-clip: content-box;
-			border-color: transparent;
-			border-style: solid;
-			$border-width-base: calc(3px / #{$scale});
-			border-width: $border-width-base;
-			// One more pixel on the start side to make the thumb appear centered
-			// in the track due to its one pixel border on the start side.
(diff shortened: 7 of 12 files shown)


stokesman:
I am pinging some folks who’ve worked on Zoom Out but not asking for a complete review. I’d appreciate any testing or any cursory opinions on how worthwhile this looks. @ajlende, @ellatrix, @scruffian, @MaggieCabrera please ping anyone else if you feel they’d have interest.

There’s been a few e2e tests that have failed consistently on this branch but so far they’re mysterious to me. The tests also fail locally but so far I can’t reproduce with manual testing of the same scenarios. It’s also hard to see how the changes here would affect them anyway. I’ll continue looking into it.

MaggieCabrera:
Thanks for this PR, @stokesman !

I'm unsure about the margin at the bottom when there's enough content to scroll. It feels like it's cutoff. I'll let @richtabor give design feedback on that though

MaggieCabrera:
The overflow is cutting off things like the focus borders of the section and messing with the position of the toolbar. This screenshot is FF on a mac with dark mode enabled and always visible scrollbars, and it looks like this:

Screenshot 2024-07-16 at 09 28 12

scruffian:
This is looking good to me. It would be good to get a review from someone who is more familiar with the code.

stokesman:
> The overflow is cutting off things like the focus borders of the section and messing with the position of the toolbar.

Thanks for highlighting this. I thought it would be a good thing in that it better aligns Zoom Out with Device Previews which do the same:
image
But I do think it doesn’t go as well with the vertical toolbar.

----
**UPDATE**:
I’ve pushed a couple of commits aimed at addressing this but just for proofing (they’d be better done in follow-ups). Now the toolbar sticks at the top of the iframe and I tried adding a “border” that might help clipping be expected. It looks like this:

image

The toolbar position would probably have to be a follow up because it requires changing `Popover` and here it’s done in too naive a way. The “border” is simple enough we could include it if favorable but we’d probably want to do a follow-up doing the same on Device previews. I’m pretty sure those used to have a border and I'm not sure why that would have changed.

scruffian:
Do you think it's work adding the `custom-scrollbars-on-hover` mixin` so that the scrollbars have the same treatment as elsewhere in the editor?

MaggieCabrera:
> Do you think it's work adding the `custom-scrollbars-on-hover` mixin` so that the scrollbars have the same treatment as elsewhere in the editor?

Good call, it might be what we need here

stokesman:
> Do you think it's work adding the `custom-scrollbars-on-hover` mixin

Thanks for the suggestion. I tried it but am uncertain of how advantageous it is in this case. There any several details to consider so I’ll start with my overall take-away. For this PR, I think the best thing would be to remove the current customization of the scrollbar and not add the mixin. That way there are fewer factors to evaluate. The sole purpose of the current customization was to prevent a shrunken scrollbar when transforming the iframe. Yet it only worked in Chrome and when the scrollbar is not overlaid. Finally, a shrunken scrollbar shouldn’t be a blocker (especially in contrast to the inaccessible scrollbar in trunk).

Now to detail what I found trying the mixin. Here’s what it does that needs overridden for this case and/or creates a tradeoff (Though none of this seems to apply for overlaid scrollbars):
 - _adds a gutter on both sides_
 This makes the apparent spacing around the iframe unequal between vertical and horizontal sides. That’s without the frame border that was a recent proofing addition. With such a border it makes for an odd band on either side of the content.
 - _makes the scrollbar thin_
 This compounds how thin the scrollbar goes due to scaling the iframe though it’s an easy override to not have it that way. Further, I couldn’t manage to both use the mixin and counter-scale its width (though this only ever worked for Chrome anyway).
 - _makes the track background transparent_
 Without a frame border this is okay as long as we don’t mind it’s part of the unequal horizontal/vertical apparent spacing around the iframe mention in the first point.

Here’s a video to help illustrate:

https://github.com/user-attachments/assets/c0222289-3018-45f4-9bef-e6770a3757db

Oh, one last tidbit. It seems to really have our way with the scrollbar it would require a library like [simplebar](https://github.com/Grsmto/simplebar) but the compromise there is a small bundle size increase.

richtabor:
A few notes: 

### 1. Wide zoom feels off
When I zoom out, I don't expect the site to be fullwidth while in zoom out. It negates the effect of zooming a bit, utilizing space that is unnecessary to cover. The perception of content and spacing seems off, when it can be so wide:


| Trunk  | This PR |
|CleanShot 2024-07-19 at 11 00 55|CleanShot 2024-07-19 at 11 00 35|


### 2. Allow full scrolling
 I don't think we should loose the space above the site when zoomed out, but allow the site to scroll like in trunk, into that space: 


| Trunk  | This PR |
|CleanShot 2024-07-19 at 11 04 40|CleanShot 2024-07-19 at 11 04 14|

### 3. Distracting omnipresent scrollbar
This scrollbar should not be omnipresent. It is positioned better, but never goes away and is distracting.  

CleanShot 2024-07-19 at 11 05 32

### 4. Improved scaling 
The scaling of content is better (testing with [this pattern](https://gist.github.com/richtabor/291b01f74254f90327e1c6c8680e49bf)). Trunk has a few oddities, as seen below: 

Trunk: 

https://github.com/user-attachments/assets/8d019be8-2f31-4ada-813d-9a28cc9a426b

PR: 

https://github.com/user-attachments/assets/5df9fded-31f2-4f80-82d4-6adc6c0f788d

stokesman:
Thanks for testing this Rich! Pardon the pause on getting back to this one.

> 1. Wide zoom feels off
> 3. Distracting omnipresent scrollbar

Both of those are addressed. The latter had been on my todo anyway.

> 2. Allow full scrolling

This one is tricky on this branch. Here are some ways I've tried doing it. I put screen recordings in the details.

<details><summary>Put the vertical spacing inside the iframe. This is like how it is in trunk but it makes for an odd scrollbar on this branch. It can be somewhat alleviated by also putting the horizontal spacing inside the iframe but that helps only as long as the viewport is narrower than canvas maximum width.</summary>

https://github.com/user-attachments/assets/3f616d99-ee54-4bea-b4d1-b111d00e795a

</details>

<details><summary>Shift the canvas vertically per document scroll progress. This one comes close in feeling but it’s not the same thing.</summary>

https://github.com/user-attachments/assets/6c9cc6a9-c7ad-4778-8d0b-8532a1257915

</details>

After exploring those, I did have another idea for how it can be made just like trunk and did some dev on it. There is promise but it’s pretty involved and would be better left to a separate PR.

There’s also room to try simpler ideas like reducing the vertical spacing or removing it from the bottom edge.

richtabor:
@stokesman I like the big PR for pulling ideas together, but is there any part of this that can be abstracted and potentially pushed as a standalone PR? 

> The scaling of content is better (testing with [this pattern](https://gist.github.com/richtabor/291b01f74254f90327e1c6c8680e49bf)). Trunk has a few oddities, as seen below:

I'm thinking this specifically.

richtabor:
I'm missing the fluidity of what's in trunk, when zoom out is engaged: 

https://github.com/user-attachments/assets/14576576-d715-42c5-ab5e-e338549f6fcc

scruffian:
>  is there any part of this that can be abstracted and potentially pushed as a standalone PR?

I think this is a good idea. This PR is really big which makes it hard to see how all the changes interact. Maybe if you could break it into smaller PRs it would be easier to review.

stokesman:
> I'm missing the fluidity of what's in trunk, when zoom out is engaged:

I think the restoration of the maximum width made it quite obvious the iframe no longer had a transition. I restored the transition now. Be on the look out to see if it reintroduced oddities when scaling the content (in regards to "The scaling of content is better"). The transition may be prone to oddities until the final scale can be specified the instant that zoom out is engaged. As of now (and in trunk), the scale is calculated multiple times in response to the the container width changes and each one initiates a transition that’s interrupted by the next until the final one.

----

> is there any part of this that can be abstracted and potentially pushed as a standalone PR?

I intend to break it up as much as possible. I’ll go ahead and look at what can be isolated but I think it’ll hard to determine and/or limited without knowing how we’ll decide on the pivotal thing here—whether Zoom Out is to keep scaling the `html` or change to scaling the `iframe`. My feeling is the feature is going to be more tenable if we switch to scaling the `iframe`.

stokesman:
I've become more confident that this is something we’ll want to do. I’ve had a go at fixing some of the issues individually and it can be done yet this seems much more advantageous.

The prime example is #64117, which if fixed in a minimal way leads to more styles that have to be applied within the canvas document and have some chance of conflicting with theme styles. It also requires changing some of the styles already applied within the document and may cause regressions of past issues (or at least more work to not regress anything). It will also—as far as I’ve been able to find—induce the same UI change that this PR does where the spacing around the canvas is outside the scrollable area.

As for breaking this PR up, it could be done but not to a large degree. The change to scale the `iframe` instead of its `html` requires most everything else here just to not regress various things. I think the biggest thing that can be done separately is preventing the scale from being adjusted when the window is resized fcdf9f1ab5944429fcbd47acdadee61c4468190b so I'd be okay with doing so. One other minor thing that will be left out is c448498f3b84a6c6c7e742120a6c358515827c95. Besides that, I’ve managed to simplify and reduce the additions here.

I'm still leaving this drafted because of the e2e tests that I’ve yet to figure out. In the mean time, I'd love if anyone could have a go at fixing #64117 because either they can figure out a way that I didn’t see or it may evidence this as being the indicated solution.

getdave (commented):
Screen Shot 2024-09-19 at 09 51 46

I don't think `trunk` has this space at the bottom of the zoom out area. It should look like a continuous page.

Edit: I noticed @richtabor [covered this in his review](https://github.com/WordPress/gutenberg/pull/63390#issuecomment-2239433625) under `Allow full scrolling`.

> ...it’s [fixes to Allow full scrolling issue] pretty involved and would be better left to a separate PR.

We can't merge this PR with those spaces.


> My feeling is the feature is going to be more tenable if we switch to scaling the iframe.

Noting that Zoom Out will not longer be active if the editor is not in an iframe. Does that help?

MaggieCabrera:
I'm seeing a few things:

Safari scaling (left) looks different to Firefox:

Screenshot 2024-09-19 at 11 03 43

On both browsers, when I click on one of the inserters, I see a horizontal scrollbar appear, and the general shadow that we are using is messing up with the separator animation (and the background of the scrollbar doesn't look great with it either):

Screenshot 2024-09-19 at 11 04 49

The blue borders on hover of the section are too small, seems like they are scaling with the rest of the content, they shouldn't. The scrollbars are scaling too, which makes it impossible to click and drag them.

draganescu:
@stokesman twp possibly useful hints to improve review accuracy:

1. split this into multiple PR that _merge into one base_. Even if some of the intermediary PRs have broken bits, it's easier to highlight the iterations and review them independently.
2. do a self review and explain every change yourself

The PR is not that "big" as it is complex - e.g. from the description and title it seems we just change a CSS technique but the changeset ends up updating types.

MaggieCabrera:
I'm having a hard time reviewing the code for this, can you please explain why you decided to change the transform to the `html` element?

jeryj:
Thanks for working on this! From some initial testing, scaling the iframe does seem to address a lot of issues. 

Some issues I see from this over trunk:
- The selected block either isn't focusable or the arrow keys are (correctly?) being intercepted by the iframe wrapper to be used to scroll the canvas. This is great on one hand, but it means you can't use the arrow keys to move block selection.
- The top/bottom padding away from the canvas.
- On trunk, I can scroll the gray area around the zoomed out canvas. On this, since the iframe maintains the scrolling, I have to scroll _on_ the frame. To a user, I don't think they should need to be aware to scroll the canvas vs the area outside it. It should be one area as far as they're concerned.

I'm not sure which items are going to be the most important, or if we can get all of the things without making some concessions.

stokesman:
Hi folks, pardon my slow response here. I’ve not thought this should be urgent so I’ve let it sit.

@getdave, thanks for having a look and reading the prior comments.

> We can't merge this PR with those spaces.

I wonder how that’s concluded. As far as I can tell, the current state is a happy accident (first appearing in #59334) not a design decision. That’s not to argue it doesn’t have value but that it would seem open to consideration.

> > My feeling is the feature is going to be more tenable if we switch to scaling the iframe.
> 
> Noting that Zoom Out will not longer be active if the editor is not in an iframe. Does that help?

To clarify, my comment was not about whether the canvas is an iframe. It was about applying `transform: scale` to the `iframe` instead of the `html`. Why more tenable? I hope to clarify that in reply to Maggie’s question later on.

--- 
@MaggieCabrera, thanks for testing again and finding those issues.

> Safari scaling (left) looks different to Firefox

So far I haven’t been able to reproduce.

> when I click on one of the inserters, I see a horizontal scrollbar appear, and the general shadow that we are using is messing up with the separator animation (and the background of the scrollbar doesn't look great with it either)

I fixed the horizontal scrollbar. I removed the shadow since it’s not a thing on trunk; was causing issues when I tested in Firefox and does jar with the inserter separator’s current intended appearance as a gap. The scrollbar’s part in that at least is largely not an issue on devices that only show them when in use.

> The scrollbars are scaling too, which makes it impossible to click and drag them.

I’m glad you mention this. Back when the PR was first made Zoom Out completely hid the scrollbars so the shrunken scrollbar seemed less bad in comparison. Now, on trunk the scrollbar can still end up hidden so I'm not sure there’s a clear winner on this point.

> why you decided to change the transform to the `html` element?

Originally I started this to keep the scrollbar visible and improve the centering of the canvas when the scrollbar is not overlaid. The first commit does that without changing where the `transform` is applied and that seemed great. Except I noticed that it introduces breakpoint disparity issues. It wasn’t obvious to me why that is and to document it I’ve made #65595. So the first decision to transform the `iframe` was to get past that issue.

From there, I discovered it facilitates fixing or allows simplifying additional things, among these:
1. preserved scroll position
2. sticky positioning support
3. reduced number of elements/properties that have to animate in transitions
4. reduced amount of styles applied inside the canvas document (and potential edge cases with themes)

--- 
@jeryj, Thanks for giving this a look.

> * The selected block either isn't focusable or the arrow keys are (correctly?) being intercepted by the iframe wrapper to be used to scroll the canvas. This is great on one hand, but it means you can't use the arrow keys to move block selection.

This is concerning, but I'm not seeing it. Granted my local branch is diverged so maybe it was fixed. I’m pushing my latest changes soon.

> * The top/bottom padding away from the canvas.
> * On trunk, I can scroll the gray area around the zoomed out canvas. On this, since the iframe maintains the scrolling, I have to scroll _on_ the frame. To a user, I don't think they should need to be aware to scroll the canvas vs the area outside it. It should be one area as far as they're concerned.

These two are technically tied together and I ❤️ that you mention the latter point. It’s the feature I would most wish to keep. Yet I'm not sure I see it as essential.  It’s one thing you can say that comes for free with scaling the `html` and here would require additional code to support. I have a branch exploring that and I suppose I need to make a PR of it based on this one so we’d be better able to compare costs.

jeryj:
I've been playing with this and looking into it with @MaggieCabrera and @ajlende. I think the scrollbar being within the iframe instead of being able to scroll the gray area is a dealbreaker. I do appreciate that moving the scaling to the iframe does make a lot of other things simpler though. Open to ideas on how to sync up that scrolling with the outer wrapper, but it feels like that will add a lot of complexity and get us in a similar spot to where we are now.

We're focused on trying to get smaller fixes to the existing issues with the current implementation. Your [PR for fixing the scaling issue](https://github.com/WordPress/gutenberg/pull/65814) is really promising, so we'll take a look at that now. Let's keep working on finding the best solution for this overall after the 6.7 release though. This `<iframe />` scaling direction may be the best in the end.

stokesman:
> I think the scrollbar being within the iframe instead of being able to scroll the gray area is a dealbreaker.

I do appreciate this concern and, as I said, having the whole gray area activate scrolling is my favorite thing about the current state. Yet I think it’s worth recognizing that it was not designed that way. It was a happy accident that came about with #59334 so it doesn’t seem like a must have. Consider also that none of the other "framed" views in the editor scroll like that (Device previews or other focused editing views) and making them do so is certainly non-trivial.

> We're focused on trying to get smaller fixes to the existing issues with the current implementation.

That’s understandable and why I’d mostly been letting this one sleep.

MaggieCabrera:
> I do appreciate this concern and, as I said, having the whole gray area activate scrolling is my favorite thing about the current state. Yet I think it’s worth recognizing that it was not designed that way. It was a happy accident that came about with #59334 so it doesn’t seem like a must have. Consider also that none of the other "framed" views in the editor scroll like that (Device previews or other focused editing views) and making them do so is certainly non-trivial.
>

Zoom out is not a device preview. We even moved it out of the dropdown because it was confusing, so they shouldn't behave the same and we don't want to change how device previews work right now, so they can stay the way they are. They are emulating a smaller screen just like browser inspectors do. Zoom out is something else and it should be closer to the site editing experience. Thanks for all your hard work here, this is a really complex problem.

stokesman:
> Zoom out is not a device preview.

Right and my point does not conflate them nor say their scrolling has to work the same. It’s largely to make the point that if device previews don’t **need** to scroll like that then it is not an essential thing for zoom out either. More generally, it’s something to keep in mind for broader perspective and especially with regards to whether zoom out’s "assembly" mode functions are to be forever tied to the view scaling as that seems to be under consideration. I think the view scaling has a potential to be useful outside of that. Anyway, some of this is just going to take time to sort out. I don’t wish to debate any of this right now.

ellatrix:
> There are some bugs that seen best solved by having zoom out scale the iframe instead of its html element.

Didn't read all comments here, but did you check 100vh elements? If I remember correctly the reason we switched to `html` (it was initially `iframe`) was that it breaks relative units. If you're going to stretch the viewport, it will affect these.

It worth nothing that zoom out is not purely a scaling of the viewport (`iframe`), it also changes the viewport. That's why it seemed better to scale the content inside instead.

ellatrix:
Original PR:
* https://github.com/WordPress/gutenberg/pull/59334

stokesman:
> Didn't read all comments here, but did you check 100vh elements?

Yes, and the first thing in testing instructions is `vh` units and your PR.

> It worth nothing that zoom out is not purely a scaling of the viewport (`iframe`), it also changes the viewport

Pardon me but I can’t seem to grasp what you mean. If it has to do with media queries then I think scaling the contents is certainly not simpler see #65595 and #66034.

ellatrix:
> Pardon me but I can’t seem to grasp what you mean.

If you adjust the `iframe` element instead, you are narrowing the width and increasing the height (so you can see more content), no? So the viewport goes from a "landscape" view to a "portrait" view, which affects the viewport units.

ellatrix:
Whereas with scaling the content, the viewport ratio remains the exact same.

stokesman:
> If you adjust the `iframe` element instead, you are narrowing the width and increasing the height (so you can see more content), no? So the viewport goes from a "landscape" view to a "portrait" view, which affects the viewport units.

Okay, I do see what you mean. The width is not narrowed but the height does increase and thus `vh` units are affected, but not in a way that is invalid or seemingly problematic.

On the other hand, consider that `vh` units are made inaccurate by scaling the content because the apparent viewport is taller but the `vh` units don’t reflect that. I’m not saying this is necessarily an important issue but it does mean the zoomed out view is not 1 to 1 with what the same viewport on the frontend would show.

ellatrix:
> On the other hand, consider that vh units are made inaccurate by scaling the content because the apparent viewport is taller but the vh units don’t reflect that. I’m not saying this is necessarily an important issue but it does mean the zoomed out view is not 1 to 1 with what the same viewport on the frontend would show.

Yes, I think that's a problem with this approach tbh. In trunk it would be exactly 1:1

stokesman:
> Yes, I think that's a problem with this approach tbh. In trunk it would be exactly 1:1

In the part you quoted, I was saying in trunk the zoomed out view is not 1:1 with the frontend view. Still, your point that this can be seen as a problem with scaling the iframe stands. I think it’s relatively small and debatable yet we don’t need to contend with it now.

---- 
Most of the issues this PR tackles are solved in trunk now. I may try and refresh this at some point in the interest of seeing how much simplification it allows. Additionally, for me, the zoomed out view could be applicable in more scenarios than an assembly mode and in such cases it should support sticky or fixed position elements and that’s something this does that trunk doesn’t (and probably can’t without some heavy-handed tactics).

ellatrix:
@stokesman Oh sorry, I didn't understand it then. What exactly is not 1:1 currently? Could you give an example? As far as I know vh units are accurate.

stokesman:
> What exactly is not 1:1 currently? Could you give an example? As far as I know vh units are accurate.

Visually `vh` units aren’t accurate—e.g. a `100vh` section doesn’t fill the apparent viewport in zoom out. I'm not claiming it’s a problem as I understand how it can be seen as nice that any `vh` dependent visuals don’t shift when zooming out. This branch would have such shifts but I think they’d be acceptable (and visually more '1:1') if ever we need to consider something like this.

PR #63394, opened 2024-07-11
Grid: Better looking block movers

noisysocks:
Update the design of the grid block movers.

Before:

Screenshot 2024-07-11 at 13 50 43

After:
Screenshot 2024-07-11 at 4 32 45 PM

Screenshot 2024-07-11 at 4 32 32 PM

### DIFF
--- a/packages/block-editor/src/components/grid/grid-item-movers.js
+++ b/packages/block-editor/src/components/grid/grid-item-movers.js
@@ -7,7 +7,11 @@ import clsx from 'clsx';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Button, VisuallyHidden } from '@wordpress/components';
+import {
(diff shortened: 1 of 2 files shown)


tellthemachines (approved):
Change LGTM!

fabiankaegy:
Hey all 👋 

Is there going to be an API for blocks to opt into this style of block movers? 

Currently the `InnerBlocks` allows you to define the orientation which controls whether you get the horizontal or vertical. Would it make sense to incorporate this new option and add a `all` option here?

tellthemachines:
> Currently the `InnerBlocks` allows you to define the orientation which controls whether you get the horizontal or vertical. Would it make sense to incorporate this new option and add a `all` option here?

Oh, that's a great question! I'm not sure how useful these controls would be outside of the grid layout scenario. Do you have any particular use cases in mind?

PR #60287, opened 2024-03-28
Fix editor canvas overflow on search results with position: relative

jeryj:
Original issue: https://github.com/WordPress/gutenberg/issues/56811

Previously this was fixed with overflow: hidden; which is more restrictive. position: relative also fixes the issue and is less restrictive of allowing other styling options and ui elements on the inserter such as a block pattern fly out in  https://github.com/WordPress/gutenberg/pull/60257

## What/Why/How?

Removes overflow: hidden from the block inserter wrapper and replaces it with a more appropriate fix of position: relative on the container causing the overflow.

Props to @t-hamano for [identifying the correct container](https://github.com/WordPress/gutenberg/pull/57127#issuecomment-2024676017) to apply this rule to.

## Testing Instructions

- Open the site editor or post editor
- Open the inserter on the top left
- Enter the search term 'List'
- Scroll the inserter container to the bottom
- Try to scroll the editor canvas. There should be no scrolling of the editor canvas.
- Search 'a'
- Scroll the inserter to the bottom 
- Try to scroll the editor canvas.

PR #63397, opened 2024-07-11
Prevent empty void at the bottom of editor when block directory results are present

### LINKED ISSUE #63384: Inserter: white space at the very bottom of the results
Using 6.6 RC 3 on an InstaWP site, I noticed that if I searched for a block then scrolled down the inserter to see the block directory offerings, I could keep scrolling and end up in an endless white space:

https://github.com/WordPress/gutenberg/assets/26996883/241805b8-0a1f-46e5-83b0-bea989cb924f

Curious if others can replicate:

1) Add a new post
2) Open the inserter
3) Search for something like "blog"
4) Notice that there's a bunch of white space as you scroll down to the bottom of the inserter

talldan:
## What?
Fixes #63384

## Why?
The issue happens because of elements with `position: absolute` in the block directory results. When elements with this positioning are relative to an element that's outside of the scroll container (an ancestor of the scroll container), they can cause weird overflow issues. Applying `position: relative` to a parent within the scroll container fixes the issue.

## How?
Because the absolutely positioned elements are within the `block-directory-downloadable-block-list-item`, I decided to apply `position: relative` there, with the idea that this component should be reusable anywhere and not cause these weird overflow issues.

I guess the previous fix (in #60287) regressed because the tabs are now shown when there a search results, and the fix only applied to the no tabs view. It could also be an idea to apply another fix similar to that one in addition to this PR so that inserter search results can never cause overflow.

## Testing Instructions
1. Create a new post
2. Open the inserter
3. Search for 'blog'
4. Scroll down to the bottom
5. A blank 'void' area shouldn't be shown at the bottom of the editor

#### Before
Screenshot 2024-07-11 at 12 35 04 PM

#### After
Screenshot 2024-07-11 at 12 34 34 PM

### DIFF
--- a/packages/block-directory/src/components/downloadable-block-list-item/style.scss
+++ b/packages/block-directory/src/components/downloadable-block-list-item/style.scss
@@ -6,6 +6,12 @@
 	display: grid;
 	grid-template-columns: auto 1fr;
 
+	// The item contains absolutely positioned items.
+	// Set `position: relative` on the parent to prevent overflow issues
+	// in scroll containers.
+	// See: https://github.com/WordPress/gutenberg/issues/63384
+	position: relative;
+
 	&:hover {
 		@include button-style__focus();
 	}


Mamaduka:
I was only able to reproduce the original issue on WP 6.6-RC2. ~Unable to reproduce it on WP 6.6-RC2 with and without Gutenberg.~

Never mind. The issue just randomly showed up again on Gutenberg trunk.

Mamaduka (approved):
Works as expected ✅

t-hamano (approved):
LGTM! 

When I investigated the absolute-positioned element that was causing this problem, I found that it was [the `VisuallyHidden` component](https://github.com/WordPress/gutenberg/blob/3c8bb803ab8b06208dd74c5bb78d4c397fa13d39/packages/block-directory/src/components/downloadable-block-list-item/index.js#L157-L159) inside the `DownloadableBlockListItem` component. Therefore, I think it makes sense to apply `position: relative` to the `DownloadableBlockListItem` component.

Issue #63387, opened 2024-07-10
Post Editor: Revert "Save" button to "Update," at least temporarily


james-tyner:
## What problem does this address?
I noticed while testing out 6.6 RC3 that, once a post has been published, the Publish button changes to say "Save." On WP 6.5, the button says "Update" when the post is in this state.

_In WP 6.6:_
<img width="307" alt="image" src="https://github.com/WordPress/gutenberg/assets/45050998/f0db9ca5-3cfc-4ab3-8159-e508a2804a4f">

_In WP 6.5:_
<img width="308" alt="image" src="https://github.com/WordPress/gutenberg/assets/45050998/9cbedeaf-9f6c-4d25-bdce-4e73cb0dcc6b">

I think the word "Save" will be confusing for some users and may lead to confusion over whether a post is actually published, particularly because of its similarity to the "Save Draft" option for draft posts.

Because that button is placed high on the page in a bold color, I expect many users rely on it as an indicator of the post's current status. Seeing "Save" doesn't give a clear indication of whether the post is already live, which may cause them to publish changes that they only intended to save as a draft. In a past life working with a different CMS, we had a similar issue that occasionally caused unfinished copy to be published live on our news website unexpectedly.

My best guess from looking through the records I can find is that this change happened as part of this PR (#60456) while trying to align the post editor UI with the site editor. I think that goal makes sense, but the Save wording may be confusing for users who aren't used to it. 

Some of the designs proposed on that PR include other mechanisms to indicate the status of the post more clearly, such as a colorful "Published" badge. I think those would be helpful in reducing the confusion introduced by the word "Save."

## What is your proposed solution?
I suggest that we change that button back to "Update" for now, until some of those additional UI elements are available to help provide situational awareness for users.

SantosGuillamot:
I think you're right and that change was introduced in [this pull request](https://github.com/WordPress/gutenberg/pull/60456). We can see the difference here:
* [Before](https://github.com/WordPress/gutenberg/pull/60456/files#diff-e856cd196e8a2ce47ee912dc9d1cd10f80ee6b17217a7447c82cf57b65373362L38): The label is "Update".
* [After](https://github.com/WordPress/gutenberg/pull/60456/files#diff-e856cd196e8a2ce47ee912dc9d1cd10f80ee6b17217a7447c82cf57b65373362R77): The label is "Save".

Pinging @ntsekouras as he might have more context.

ntsekouras:
[...]

PR #60456, opened 2024-04-04
Post Editor: Update publish flow

ntsekouras:
## What?
Part of: https://github.com/WordPress/gutenberg/issues/59878

This PR aims to update the publishing flow in post editor, so we can later unify it with site editor.

### Tasks

1. Update the post status component design and use it in post editor

Panel               |  Post status label
-----------------------------------------|-------------------------
Panels |  status

From the above designs I'm not sure we can add another control(publish date) in between the radio control options. --cc @jameskoster @mirka 

2. Update `publish` button

- Match `save` label and save panel when multiple entities have changed
- handle pending status/button
- handle switch to draft status/button properly
- remove `visibility` panel

 ## Testing Instructions

Testing should be thorough including any possible flow. Examples:
1. Make changes in a post by updating the status directly from the panel, like from `published to draft`, etc.. 
2. Do the same as above with other content changes and other entities' changes
3. Make changes to a `draft or pending` post and use `save as draft` or `save as pending` buttons in combination with other changes and changes in other entities.
4. Do the same as above and use the main `save/publish` button
5. Make changes only to other entities(pattern) and save either to that page or by returning to the post editor of the main entity(page).
6. I guess other flows that I'm missing too..

PR #63399, opened 2024-07-11
Post Editor: Revert "Save" button to "Update"

rohitmathur-7:
## What?
Revert "Save" button to "Update", and "Saving..." to "Updating..." in the post editor.
Issue: https://github.com/WordPress/gutenberg/issues/63387

## Why?
Save word is a little confusing to users when they are actually updating the post.

## How?
Replaced "Save" word with "Update" and, replaced "Saving..." with "Updating...".

## Testing Instructions

- Go to the Post editor.
- Make changes in the post.
- We can see the button on top with the word "Update".

## Screenshots or screencast 
gutenberg-post-editor-btn

### DIFF
--- a/packages/editor/src/components/post-publish-button/label.js
+++ b/packages/editor/src/components/post-publish-button/label.js
@@ -63,7 +63,7 @@ export default function PublishButtonLabel() {
 		! isAutosaving
 	) {
 		/* translators: button label text should, if possible, be under 16 characters. */
-		return __( 'Saving…' );
+		return __( 'Updating…' );
 	}
 	if ( ! hasPublishAction ) {
 		// TODO: this is because "Submit for review" string is too long in some languages.
@@ -79,7 +79,7 @@ export default function PublishButtonLabel() {
 			! [ 'future', 'publish' ].includes( postStatus ) ) ||
 		( ! postStatusHasChanged && postStatus === 'future' )
 	) {
-		return __( 'Save' );
+		return __( 'Update' );
 	}
 	if ( isBeingScheduled ) {
 		return __( 'Schedule' );


youknowriad:
Hey Thanks for the PR. I understand that in some situations it may be less clear but I think there are pros and cons.

"Update" doesn't work well in multiple situations:

 - When you're saving multiple entities (not just a single post)
 - When you're switching the status back to "draft", I'm not sure if "update" is correct. Save is better.

And Save is more generic and just means save all the changes, so it works in more situations I would say.

ntsekouras:
Thank you for the PR! I've commented on the issue and now saw your PR, so I'll drop a link here too.

My [comment](https://github.com/WordPress/gutenberg/issues/63387#issuecomment-2222998699):
>Thanks for the issue! I think you're mostly referring to a possible accidental publishing of a draft. In that case the button doesn't render 'Saveinstead ofUpdate, but it renders Publish`, which was also the case before the saving/publishing flow updates. Did I get that wrong here?

>Additionally, all these changes have made the current status value of the editing post more prominent, in the post summary(inspector controls), which IMO it also reduce the risk of not knowing the post status.

So, since this PR tries to fix the use case of the linked issue, If I'm right, there shouldn't be any confusion because the update/save change doesn't apply.

Finally, I'd echo Riad about `save` being more generic than update.

james-tyner:
Hello! I created the linked issue https://github.com/WordPress/gutenberg/issues/63387 that led to this PR.

I understand the argument that Save is generic enough to apply in all situations, but I think the key issue here isn't what the button is doing, but what the state of the button is representing to the user. Users have likely grown accustomed to seeing an "Update" button and recognizing that to mean that the post is published.

At a publication where multiple writers and editors may jump in and out of posts and reach them directly from outside the context of the Posts list in WordPress (which shows the post status), it's important for people to have an easy way to know what the status is from the post editor screen.

If we are opposed to moving away from the generic "Save" wording, it would be valuable to devote some effort to other ways of keeping users aware of the status of the post. I saw on #60456 that there was some design effort going into creating a color-coded post status label that would appear at the top of the settings sidebar, and I think that would help a lot. 

My proposal is that we could try to preserve the Update button for now until those other status indicators are ready.

youknowriad:
I want to note as well that the "save" label also existed before if you were using the site editor instead. In 6.6 the post and site editor UIs are unified which led to the user of the more generic "save" label. So there will be a change for users regardless of which label is picked.

I'd appreciate more design eyes on this but I have a feeling that "save" is less confusing in more situations than "update".

Mamaduka:
I'm going to close this as "stale". The issue needs a decision before it's actionable. Let's move the discussion back to the issue.

Issue #43247, opened 2022-08-16
Tracking: Border Design Tools Consistency


aaronrobertshaw:
## Overview

This issue details the current state of border block support or design tool adoption across all blocks and tasks required to fill any gaps. Overall design tool consistency efforts are being tracked via the parent issue: https://github.com/WordPress/gutenberg/issues/43241.

## Legend

| Value         | Description                                                                  |
| ------------- | ---------------------------------------------------------------------------- |
| ✅            | Feature has been adopted and is displayed as a **default** control |
| ✅ (Optional) | Feature has been adopted but is an **optional** control            |
| ⚠             | There is a **bug or issue** with this block support feature's adoption           |
| ❌            | Feature has been explicitly **opted out** of           |
| 🚫 | Block support will not be adopted for this block |
| <PR#>         | Links to **PR** adopting the feature for this block                |
| -             | Feature has **not explicitly** been adopted/omitted                              |
| 🛠             | Implemented via an **ad hoc** / bespoke control                                  |
| 🚧 | Work is in progress towards adopting this feature (no PR yet) |

## Block Support Adoption

_Note: Deprecated blocks have been omitted from this table. e.g. Comment Author Avatar, Post Comment & Text Columns._

| Block                        | Color         | Radius        | Style         | Width         |
| ---------------------------- | ------------- | ------------- | ------------- | ------------- |
| Archives                     | ✅ (Optional) | ✅ (Optional) | ✅ (Optional) | ✅ (Optional) |
| Audio                        | https://github.com/WordPress/gutenberg/pull/64494 | https://github.com/WordPress/gutenberg/pull/64494 | https://github.com/WordPress/gutenberg/pull/64494 | https://github.com/WordPress/gutenberg/pull/64494 |
| Avatar                       | ✅ (Optional) | ✅            | ✅ (Optional) | ✅ (Optional) |
| Button                       | ✅ | ✅ | ✅ | ✅ |
| Buttons                     | ✅ | ✅ | ✅ | ✅ |
| Calendar                     | https://github.com/WordPress/gutenberg/pull/64345 | https://github.com/WordPress/gutenberg/pull/64345 | https://github.com/WordPress/gutenberg/pull/64345 | https://github.com/WordPress/gutenberg/pull/64345 |
| Categories                   | ✅ | ✅ | ✅ | ✅ |
| Code                         | ✅            | ✅ (Optional) | ✅ (Optional) | ✅            |
[...]

PR #63400, opened 2024-07-11
Archives: Add border block support

carolinan:
## What?

Add border block support to the archives block.

Partial for https://github.com/WordPress/gutenberg/issues/43247

## Why?

## How?

The PR:
Adds the border block support in block.json, with all controls optional, hidden by default.
Adjusts the editor styles by removing the duplicate spacing and duplicate borders (See the discussion below about why this was needed for this server rendered block, which uses the block class name twice in the markup).

## Testing Instructions
Create a new post and add an archives block. Adjust the border settings and compare the result in the editor and front.

## Screenshots or screencast 
After:
Editor view, after|Archives block with border, on the front


carolinan:
Since this block is not used to add design or layout, I am very open to adjusting which controls that should be shown by default.

aaronrobertshaw (commented):
Thanks for working on this one @carolinan 👍 

At the block instance level this works well for me. Once global styles are added to the mix we'll get double borders in the editor.

Screenshot 2024-07-29 at 7 04 21 PM

This is due to the Archives block using the `ServerSideRender` component in the editor. It results in the block class being used on both the outer wrapper and the inner render of the block. The outer wrapper is required for block selection etc. 

The only complete and elegant solution is to refactor the entire Archives block to use the REST API for data and avoid server side rendering entirely. The benefit here probably doesn't justify the effort at this stage.

An alternative in the short term would be to add some editor only styles to remove any borders, padding etc. on the nested element with the block's class e.g. `.wp-block-archives .wp-block-archives`.

A simple quick fix was done recently for the Tag Cloud block that suffers from the same issue: https://github.com/WordPress/gutenberg/pull/63832.

> Adjusts the editor CSS to remove the default margin on the `<UL>` used by the block.

The block library styles don't contain this reset on the `ul` for the frontend. The lack of a margin there appears due to layout styles which a theme may or may not have. 

The need to remove it from the editor is due to the server side rendering again. It means the inner block classes get the layout margin. While the `ul.wp-block-archives` class does effectively target the inner render of the block, it's not clear in the stylesheet that's what is intended. 

> Since this block is not used to add design or layout, I am very open to adjusting which controls that should be shown by default.

With the combined border controls, I think if it is worth showing any border control by default it should show them all. As you note, the application of borders here is a niche case. I'd vote to make all the border controls on this block optional.

carolinan:
Thank you! I completely missed the duplicate borders in my testing.

packages/block-library/src/archives/block.json
```diff
@@ -26,6 +26,18 @@
...
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"radius": false,
+				"color": false,
+				"width": false,
+				"style": false
+			}
```
aaronrobertshaw:
```suggestion
```

Border controls are optional by default, no need to explicitly opt out here.

aaronrobertshaw (approved):
Thanks for wrangling this adoption of border support @carolinan 👍 

I've given this PR another test and it is working well for me. Nice work!

✅ Global Styles get applied correctly
✅ No duplicate padding or border in the editor due to server-side rendering
✅ Block instance styles override Global Styles
✅ Application of styles are consistent between editors and frontend
✅ Box sizing takes border into account when parent enforces width

Before merging I've left a small improvement as a suggestion below. Also, can you tweak the PR description so that it matches what's actually landing for others belatedly coming to this one?


https://github.com/user-attachments/assets/482574f4-19b9-4664-a323-73890c0eb0c0

packages/block-library/src/archives/block.json
carolinan: Hm yesterday I found that if I removed it, all controls were visible by default in the block editor. Maybe I forgot to build it or something. - It is working now!

PR #63401, opened 2024-07-11
Typography: Stabilize typography block supports within block processing

andrewserong:
## What?

Part of #63001, alternative to #63072.

Stabilize the following typography block supports by no longer requiring the `__experimental` prefix in `block.json` and when registering block types:

1. `__experimentalFontFamily` → `fontFamily`
2. `__experimentalFontStyle` → `fontStyle`
3. `__experimentalFontWeight` → `fontWeight`
4. `__experimentalLetterSpacing` → `letterSpacing`
5. `__experimentalTextDecoration` → `textDecoration`
6. `__experimentalTextTransform` → `textTransform`
* ~`__experimentalWritingMode` → `writingMode`~ (writingMode will be left as experimental for now, [based on feedback](https://github.com/WordPress/gutenberg/issues/63001#issuecomment-2244417618))

The approach in this PR is to handle the transformation from the experimental properties at the point that blocks are processed. This happens during block registration. The idea proposed here is to do the transformation just before filters are applied so that the filters use the "real" block supports, not the experimental ones.

Note that this PR does _not_ update any of the existing blocks to use the stable syntax. That can be worked on in follow-up PRs.

## Why?

Stabilize the above typography block supports so that they can be used without the `__experimental` prefix. However, at the same time, support _both_ the experimental and the non experimental syntaxes, so that blocks out in the wild do not have to be updated in order to continue to work with the typography support.

## How?

A previous attempt was tried in #63072, however the approach in this PR is simpler:

* Update the keys for each of the affected typography supports to no longer use the `__experimental` prefix
* Update the block processing logic so that for the affected typography supports, we treat `__experimental` as opting in to the "real" support in both JS and PHP
* Note that the PHP backport for this will propose moving the logic into one of the PHP classes for blocks, whereas this PR uses a filter to achieve a similar result.

## Testing Instructions

Here is some block markup for a Title block and a Group block with a paragraph within it, with typography applied to both blocks:

<details>

<summary>Test markup</summary>

```html
<!-- wp:post-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"1000","lineHeight":"2","letterSpacing":"10px","textDecoration":"underline","textTransform":"uppercase"}},"fontSize":"xx-large","fontFamily":"system-sans-serif"} /-->

<!-- wp:group {"style":{"typography":{"fontStyle":"italic","fontWeight":"700","lineHeight":"2","letterSpacing":"15px","textDecoration":"underline","textTransform":"uppercase"}},"fontSize":"x-large","fontFamily":"heading","layout":{"type":"constrained"}} -->
<div class="wp-block-group has-heading-font-family has-x-large-font-size" style="font-style:italic;font-weight:700;letter-spacing:15px;line-height:2;text-decoration:underline;text-transform:uppercase"><!-- wp:paragraph -->
<p>A paragraph in a Group</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```

</details>

1. Add the above test markup to a post or template.
2. Ensure that the applied typography supports render correctly in the editor and on the site frontend.
3. Apply the below patch, or manually edit the Group or Post Title blocks' `block.json` files to use the non-experimental syntax for the typography supports. E.g. update `supports.typography.__experimentalFontStyle` to `supports.typography.fontStyle` and so on for the six typography supports covered in this PR.
4. Test that each of the affected typography supports can be edited / updated at the block level in the editor, and render correctly in the editor and on the site frontend.

<details>

<summary>diff / patch to update Group and Post Title blocks to use stabilized typography supports</summary>

```diff
diff --git a/packages/block-library/src/group/block.json b/packages/block-library/src/group/block.json
index 3c7d8d3ce13..db6c10c5684 100644
--- a/packages/block-library/src/group/block.json
+++ b/packages/block-library/src/group/block.json
@@ -76,12 +76,12 @@
 		"typography": {
 			"fontSize": true,
 			"lineHeight": true,
-			"__experimentalFontFamily": true,
-			"__experimentalFontWeight": true,
-			"__experimentalFontStyle": true,
-			"__experimentalTextTransform": true,
-			"__experimentalTextDecoration": true,
-			"__experimentalLetterSpacing": true,
+			"fontFamily": true,
+			"fontWeight": true,
+			"fontStyle": true,
+			"textTransform": true,
+			"textDecoration": true,
+			"letterSpacing": true,
 			"__experimentalDefaultControls": {
 				"fontSize": true
 			}
diff --git a/packages/block-library/src/post-title/block.json b/packages/block-library/src/post-title/block.json
index 796da166710..0e5ff5f5aba 100644
--- a/packages/block-library/src/post-title/block.json
+++ b/packages/block-library/src/post-title/block.json
@@ -51,12 +51,12 @@
 		"typography": {
 			"fontSize": true,
 			"lineHeight": true,
-			"__experimentalFontFamily": true,
-			"__experimentalFontWeight": true,
-			"__experimentalFontStyle": true,
-			"__experimentalTextTransform": true,
-			"__experimentalTextDecoration": true,
-			"__experimentalLetterSpacing": true,
+			"fontFamily": true,
+			"fontWeight": true,
+			"fontStyle": true,
+			"textTransform": true,
+			"textDecoration": true,
+			"letterSpacing": true,
 			"__experimentalDefaultControls": {
 				"fontSize": true
 			}

```

</details>

Finally, double check that plugins or themes can still filter out particular experimental typography block supports using the experimental syntax. For example, with the below snippet added to `functions.php` in TT4 theme, we can skip Font support on the Post Title block. With the snippet active, custom font family set on any Post Title block should not be output on the site frontend:

```php
function disable_post_title_font_family( $args ) {
	if ( 'core/post-title' === $args['name'] ) {
		_wp_array_set( $args, array( 'supports', 'typography', '__experimentalFontFamily' ), false );
	}

	return $args;
}
add_filter( 'register_block_type_args', 'disable_post_title_font_family', 20 );
```

## Screenshots or screencast 

Note: while working on this I noticed that the Appearance controls are not at the correct height in the sidebar UI. That is unrelated to this change, but I've opened a separate issue for it over in https://github.com/WordPress/gutenberg/issues/63886

All of these typography supports should work just as they do on `trunk`:

image

## Related links

* Backport PR, that uses a different PHP implementation: https://github.com/WordPress/wordpress-develop/pull/7069


andrewserong:
Hi folks! I've pinged some of you for review who I think might be interested in this PR — it is not at all urgent for review, as I'm hoping this will make it in for WP 6.7, so please don't feel like you need to take a look at it immediately 🙂

This is just one particular way that stabilization could occur for these block supports, I'm very happy for ideas if anyone has other ideas for how to handle it. Also, please note that the sync PR for core takes a slightly different approach over in https://github.com/WordPress/wordpress-develop/pull/7069 as in core we don't need to rely on filters for the PHP implementation.

lib/block-supports/typography.php
```diff
@@ -621,3 +621,41 @@ function gutenberg_get_typography_font_size_value( $preset, $settings = array()
 	remove_filter( 'render_block', 'wp_render_typography_support' );
 }
 add_filter( 'render_block', 'gutenberg_render_typography_support', 10, 2 );
+
+/**
+ * Filters the block type arguments to stabilize experimental block supports.
```
ramonjd: Do you reckon this migration will be tied to 6.7, or a general deprecation? 

I guess what I'm really asking is, should the filter be in a compat folder?

ramonjd (commented):
I just ran through the test steps quickly - all working as intended.  🚀 

I like the small footprint of this PR, and the fact that there's no disruption to existing block.json supports.

Also, it'll pad out the support docs too!

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#typography

Would we need to update the block.json schema as well?

https://github.com/WordPress/gutenberg/blob/trunk/schemas/json/block.json#L578

lib/block-supports/typography.php
andrewserong: Oh, yes, a `compat` folder is a better idea since the idea is to remove this once 6.7+ is the required WP version for Gutenberg 👍

I'll move that around tomorrow. Thanks!

andrewserong:
Thanks for taking this for a spin!

> Also, it'll pad out the support docs too!

Totally! The overall goal with this work is so that we can better promote these block supports for folks to use 👍

> Would we need to update the block.json schema as well?

Yes, but I'd like to leave that to a separate PR to try to keep this change as small as it can logically be, and since the `__experimental` prefix will still be allowed syntax. I've added it to the tasks list in https://github.com/WordPress/gutenberg/issues/63001 (this PR covers the first two tasks in that list).

lib/block-supports/typography.php
andrewserong: Done in https://github.com/WordPress/gutenberg/pull/63401/commits/71a0a964bc41625a5e9c67cf3cdec54445253f58 👍

lib/compat/wordpress-6.7/blocks.php
```diff
@@ -43,3 +43,44 @@ function gutenberg_filter_block_type_metadata_settings_allow_variations_php_file
...
+			$stable_typography_supports[ $key ] = $value;
+		}
+	}
+
+	$args['supports']['typography'] = $stable_typography_supports;
+
+	return $args;
+}
+
+add_filter( 'register_block_type_args', 'gutenberg_stabilize_experimental_block_supports', 10, 1 );
```
aaronrobertshaw: This might need to be forced to run last so 3rd party filters modifying support flags have been applied first.

aaronrobertshaw (commented):
Thanks for tackling this @andrewserong 👍 

I haven't gotten to test deeply yet but I think there is a specific scenario we might need to consider when stabilizing block supports. It may be less of an issue with typography supports compared to those with less widespread adoption but an issue all the same.

For some blocks, third party extenders filter block type args to enable or disable different block supports. The approach in this PR currently uses the default priority for the filter. I think it might need to be forced to run last so any filters setting experimental flags can stabilized.

Try adding something like the following to the active theme's functions.php

```php
function disable_some_stuff( $args ) {
	if ( 'core/paragraph' === $args['name'] ) {
		_wp_array_set( $args, array( 'supports', 'typography', '__experimentalFontFamily' ), false );
	}

	return $args;
}
add_filter( 'register_block_type_args', 'disable_some_stuff', 10 );
```

You'll see that the font family support under the stabilized key is still enabled. If you set the example to something like `null` for debugging purposes, the experiemental key will still be within the final filtered typography supports.

andrewserong:
Oh, thank you for flagging @aaronrobertshaw, I hadn't thought of that!

It sounds like setting a priority of something like `100` might work?

If this needs to run last, then it sounds like it'll also have implications for the backport in https://github.com/WordPress/wordpress-develop/pull/7069 — in that case, I imagine we'll want to move the code that stabilizes the support to _after_ the `$args = apply_filters( 'register_block_type_args', $args, $this->name );` line 🤔. Or, alternately, in the core backport use a filter after all, though I prefer the idea of not using a filter as it keeps things more encapsulated.

aaronrobertshaw:
> It sounds like setting a priority of something like 100 might work?

I'm not sure what the best value is. I could easily see extenders adding like 999 thinking they want their filter to run last. 

Is `PHP_INT_MAX` an option?

> it sounds like it'll also have implications for the backport

Yep 👍 

> Or, alternately, in the core backport use a filter after all, though I prefer the idea of not using a filter as it keeps things more encapsulated.

I think in core the preference would be to not incur the overhead of a filter. That was the feedback received on some block style variations stuff and the theme json resolver. Private functions were created and called directly before/after the filter as needed.

andrewserong:
> Is PHP_INT_MAX an option?

Ah, of course! Yes, I see we're using that in a couple of places, I'll use that here.

> Private functions were created and called directly before/after the filter as needed.

Good idea. Actually a private function is a good idea for the backport anyway as it'll neaten up that function a bit. I'll follow-up there either tomorrow or next week if I run out of time.

andrewserong:
Updated, thanks again for the snippet. I wound up testing locally with the post title block as that gives us an immediate indication on the site frontend at render time:

```php
function disable_post_title_font_family( $args ) {
	if ( 'core/post-title' === $args['name'] ) {
		_wp_array_set( $args, array( 'supports', 'typography', '__experimentalFontFamily' ), false );
	}

	return $args;
}
add_filter( 'register_block_type_args', 'disable_post_title_font_family', 20 );
```

The above is working now with the update the to `PHP_INT_MAX` in https://github.com/WordPress/gutenberg/pull/63401/commits/bd5de6b7c8a52cffbd03c9890922519f4e5e7378 👍

aaronrobertshaw:
Thanks for the quick iteration here 🚀 

The code changes look good so far. I'll give it a thorough test with fresh eyes tomorrow.

> I wound up testing locally with the post title block as that gives us an immediate indication on the site frontend at render time:

Bonus points to you for resharing the updated snippet. I was being lazy and dumping out the filtered data rather than locating an appropriate block to confirm with 😅 

When I get to testing, do you have any objections to me updating the test instructions to include the filtering angle for others than might belatedly come to this?

andrewserong:
> When I get to testing, do you have any objections to me updating the test instructions to include the filtering angle for others than might belatedly come to this?

Not at all, feel free to update the description. Thanks!

Edit: I've updated the description to include the snippet 👍

packages/blocks/src/store/process-block-type.js
```diff
@@ -102,6 +124,9 @@ export const processBlockType =
 			),
 		};
 
+		// Stabilize any experimental supports before applying filters.
+		blockType.supports = stabilizeSupports( blockType.supports );
+
```
aaronrobertshaw, line 129: Plugins and themes could filter the supports using experimental flags. Same as with the PHP filter. So maybe this should be moved to after the application of filters similar to using `PHP_INT_MAX` as the priority.

packages/blocks/src/store/process-block-type.js
```diff
@@ -119,8 +144,11 @@ export const processBlockType =
 		}
 
 		if ( settings.deprecated ) {
-			settings.deprecated = settings.deprecated.map( ( deprecation ) =>
-				Object.fromEntries(
+			settings.deprecated = settings.deprecated.map( ( deprecation ) => {
+				deprecation.supports = stabilizeSupports(
+					deprecation.supports
+				);
```
aaronrobertshaw, line 155: Nice. I was wondering if this PR would also be covering deprecations.

phpunit/block-supports/typography-test.php
```diff
@@ -283,6 +283,111 @@ public function test_should_generate_classname_for_font_family() {
 		$this->assertSame( $expected, $actual );
 	}
 
+	/**
+	 * Tests that stabilized typography supports will also apply to blocks using
+	 * the experimental syntax, for backwards compatibility with existing blocks.
+	 *
+	 * @covers ::wp_apply_typography_support
```
aaronrobertshaw:
```suggestion
	 * @covers ::gutenberg_apply_typography_support
```

Should this be the Gutenberg function while the test is in this plugin?

phpunit/block-supports/typography-test.php
```diff
@@ -283,6 +283,111 @@ public function test_should_generate_classname_for_font_family() {
...
+	 */
+	public function test_should_apply_to_experimental_typography_supports() {
```
aaronrobertshaw:
```suggestion
	public function test_should_apply_experimental_typography_supports() {
```

I might be misreading the intent of the name but the tests seems related to the application of supports rather than applying something to the supports themselves. Ignore me if I'm getting it wrong.

phpunit/block-supports/typography-test.php
```diff
@@ -283,6 +283,111 @@ public function test_should_generate_classname_for_font_family() {
...
+
+		$this->assertSame( $expected, $actual );
+	}
+
+	/**
+	 * Tests that stabilized typography supports are applied correctly.
+	 *
+	 * @covers ::wp_apply_typography_support
+	 */
+	public function test_should_apply_to_stabilized_typography_supports() {
```
aaronrobertshaw:
```suggestion
	public function test_should_apply_stabilized_typography_supports() {
```

Same a comment on the previous test.

phpunit/block-supports/typography-test.php
```diff
@@ -283,6 +283,111 @@ public function test_should_generate_classname_for_font_family() {
...
+			'style' => 'font-style:italic;font-weight:bold;text-decoration:underline;text-transform:uppercase;letter-spacing:1px;',
+		);
+
+		$this->assertSame( $expected, $actual );
+	}
+
+	/**
+	 * Tests that stabilized typography supports are applied correctly.
+	 *
+	 * @covers ::wp_apply_typography_support
```
aaronrobertshaw:
```suggestion
	 * @covers ::gutenberg_apply_typography_support
```

When in Gutenberg, do as Gutenberg does?

aaronrobertshaw (commented):
This is looking good so far @andrewserong 👍 

I only have one real question and a few minor nits I've left as inline comments but this PR is pretty close.

Is it possible that plugins could be checking block support types at all to enhance capabilities? 

My understanding of the current migration approach is that the experimental flags are removed entirely from the end result. Would it be worth keeping them for a couple of releases? It might by time to communicate their removal to plugin authors.

I haven't gone looking in the WP Directory though so maybe its a non-issue 🤷 

P.S. Sorry for the delay in getting back to review this one 🙏

packages/blocks/src/store/process-block-type.js
andrewserong: Oh, good question. Unfortunately in this case I think we might need to apply it before the JS filters, because we have logic such as in [addAttributes](https://github.com/WordPress/gutenberg/blob/cc148c3a5b94f706aa152ee68075bc1c56e6ce79/packages/block-editor/src/hooks/font-family.js#L26) for font family that's registered against `blocks.registerBlockType` that will check against the stable name for the typography support.

I'll give this a bit more thought!

aaronrobertshaw: I guess it could be done in both places. Hard to judge if it is worth the cost of doing so. Perhaps outreach could help get a feel for the possible impact of leaving it as only before the filters?

phpunit/block-supports/typography-test.php
andrewserong: Not misreading anything! I tried out a couple of different names for these tests and this was a typo when settling on the final one 😄

andrewserong:
Thanks for reviewing and thinking this feature through @aaronrobertshaw, I very much appreciate the discussions thus far! 🙇 

> My understanding of the current migration approach is that the experimental flags are removed entirely from the end result. Would it be worth keeping them for a couple of releases? It might by time to communicate their removal to plugin authors.

Good question. The approach in this PR right now is to migrate from the `__experimental` prefix to the stable one, so within Gutenberg we treat the stable one as the "real" version of the support. In the PHP code we've ensured this migration happens late in the process so that any plugins that attempt to mutate `supports.typography.__experimentalFontFamily` for example, are reflected in the final value for that support. In JS, however, we need to apply the migration earlier in the process so that the `block.registerBlockType` filter is run with the already applied migrated support, so that there's a match. The risk is whether plugins are also attempting to do JS filtering based on the `__experimental` prefix, and whether we need to account for that here.

Overall, I'm leaning a little toward the approach in this PR rather than having to keep the experimental versions of the support around. I explored that in the earlier PR #63072 which didn't feel as elegant as the approach here, but it did technically work.

Also, as discussed in our thread (https://github.com/WordPress/gutenberg/pull/63401#discussion_r1696250613) we could also look at running the migration twice in JS (once before running the filter, and once after), though that might be redundant.

At this stage, since I've gone down a particular rabbithole with this feature, I'll just ping @WordPress/gutenberg-core for visibility in case anyone else has further thoughts or ideas about how we can best handle stabilizing block supports. We'll need to re-use this again in the future when we stabilize other supports (e.g. border), so hopefully this PR will be a helpful exercise not just for Typography 🙂

youknowriad (commented):
I think this is one of those PRs where it's very hard to judge the impact without actually shipping. 

From my testing, reading of the code and the discussion. It seems a sane approach but we should highlight this change in the release post and keeping an eye on potential issues being raised (also during the WP beta period)

packages/blocks/src/store/process-block-type.js
```diff
@@ -102,6 +124,9 @@ export const processBlockType =
 			),
 		};
 
+		// Stabilize any experimental supports before applying filters.
+		blockType.supports = stabilizeSupports( blockType.supports );
+
 		const settings = applyFilters(
```
gziolo, line 130: What about filters that update `supports` and inject their entries, which could be experimental? What about plugins that expect experimental `support` entries and make decisions based on them?

aaronrobertshaw: This was raised [earlier](https://github.com/WordPress/gutenberg/pull/63401#discussion_r1696202846) and I believe what prompted @andrewserong to [ask for additional opinions](https://github.com/WordPress/gutenberg/pull/63401#issuecomment-2257635792) on the best path forward.

In a nutshell, the scenario we flagged, requires the migration to occur after the JS filters have been applied. Andrew found however that the block support filters adding attributes might be cleaner, and a better example, if the migration occurred before.

To cover both bases we may need to either
- migrate both, before, and after, filters
- keep the block support filters checking both the experimental and stable flags, then migrate only once after filters are applied

gziolo: It’s also possible that 3rd party plugins run some checks based on the experimental block support syntax so the more I think about it the more it feels you should keep both for some time to let devs adjust or even forever. Best it would be to verify with some plugins that reference these experimental names. The approach shared by @youknowriad in https://github.com/WordPress/gutenberg/pull/63401#pullrequestreview-2206898818 might be also a good first step if you want to exercise the current ecosystem.

andrewserong: Thanks for the discussion here!

> the more I think about it the more it feels you should keep both for some time to let devs adjust or even forever.

I think we definitely want to at least _support_ the experimental syntax forever so that block plugins out in the wild continue to work without being updated.

In terms of support within the filters and in regards to keeping both, how might that look? I.e. would it be something like:

* When `__experimentalFontFamily` is set, copy to `fontFamily` but leave `__experimentalFontFamily` intact
* When `fontFamily` is set, copy to `__experimentalFontFamily` for back-compat?

If we do the latter as well as the former, after the filter is run, we might need to check if the value has changed (i.e. if the filter mutated one or the other values) and then resync them 🤔

I might not get a chance to update this PR this week, but I think a next step for me could be to add some tests that simulate possible use cases for plugins augmenting or inspecting the values. That might help illuminate possible cases for us to consider, or at the very least, help us limit the potential scope for affecting plugins out in the wild.

gziolo: Would it be technically feasible to convert the config objects in block supports that have experimental flags into proxy objects and handle the backward compatibility through augmented setters and getters? Let me share a similar case in PHP to better illustrate the idea:

https://github.com/WordPress/wordpress-develop/blob/74e03e3cbef2f2565028f446c76acb2dabf749bd/src/wp-includes/class-wp-block-type.php#L353L387

That allows to always use in core the new names but correctly match properties through legacy name when necessary.

youknowriad: To be honest, I'm not sure how far we should take this. While I think folks are probably registering blocks with these experimental supports. I'm not sure people are "reading" these support keys themselves. It feels like not something very useful. This is why I suggested that maybe the current approach is enough and the only real way to know is to try it.

gziolo: Again, I’m perfectly fine with trying as is in this PR to see what issues might pop up and react accordingly. As long is it’s communicated with plugin and theme authors early so they know how to migrate code, where to raise issues or seek for help 👍

Next time, it's going to be easier to repeat the same process based on experience gained this time.

andrewserong: Good points! Yeah, I'm leaning toward going with something like the current state of this PR, and essentially committing to allowing plugins to opt-in to block supports, but not explicitly deal with the situation of plugins "reading" the support keys.

In terms of what it'll mean for plugins in the current state:

1. 👍 They can still register blocks and opt-in to typography supports using the `__experimental` prefix.
2. 👍 They can still filter `register_block_type_args` in PHP to opt existing blocks into typography supports that they might not have yet, as the filter in this PR will run last.
3. 👎 They won't be able to filter in JS on `blocks.registerBlockType` to opt blocks in to `__experimental` prefix supports that they don't already have, as the stabilization in JS happens prior to applying the filter.

In order to deal with the potential of `#3`, do you think it'd be worth us running the stabilization a second time after the `blocks.registerBlockType` filter, so that if anything has been opted into, it'll get moved over to the stable key again?

It does mean running the stabilization twice instead of once for each call to process the block type. But it isn't a hugely complex function, so might not be too bad 🤔

I'll leave this PR as is over the weekend and revisit it sometime next week. Thanks again for all the discussion!

gziolo: One final aside.

> 👎 They won't be able to filter in JS on blocks.registerBlockType to opt blocks in to __experimental prefix supports that they don't already have, as the stabilization in JS happens prior to applying the filter.

I have no real data to support my assumption but based on feedback I got at WordCamps and from GitHub reports, the most common scenario is removing UI controls for these features. I don’t know how that translates into code here though.

youknowriad: Removing features is definitely a common scenario. It might be good to know whether folks doing that are not impacted that much.

andrewserong: Good points, I think at the very least this PR should handle the basic ways that plugins might disable controls in both JS and PHP. I think this leans me toward applying the transformation twice in JS (once before running the filter, once after), so that we can capture it. Example test code:

```js
function removeTypographySupports( settings, name ) {
	if ( name === 'core/paragraph' && settings.supports.typography ) {
		settings.supports.typography.__experimentalTextTransform = false;
	}
	return settings;
}

addFilter(
	'blocks.registerBlockType',
	'core/style/addAttribute',
	removeTypographySupports
);
```

In `trunk` that'll remove the Letter Case option from the Paragraph block, but not with this PR applied. Ideally it'd recognise if the value has changed after the above filter is run, and migrate it accordingly.

I'll have a play with it this week.

ndiego: >  I think this leans me toward applying the transformation twice in JS (once before running the filter, once after), so that we can capture it.

This would be fantastic, if possible. We have a lot of "Editor curation" content out there that shows folks how to disable/enable block supports with JS and PHP. Either way, I think the developer community will be quite excited about this stabilization effort. 

Even if the transformation was not applied twice, you could do something like this to support 6.7 while maintaining backward compatibility, right? 

```js
function removeTypographySupports( settings, name ) {
	if ( name === 'core/paragraph' && settings.supports.typography ) {
		settings.supports.typography.__experimentalTextTransform = false;
		settings.supports.typography.textTransform = false;
	}
	return settings;
}

addFilter(
	'blocks.registerBlockType',
	'core/style/addAttribute',
	removeTypographySupports
);
```

andrewserong: > Even if the transformation was not applied twice, you could do something like this to support 6.7 while maintaining backward compatibility, right?

Yes, that example should work nicely for plugins that wish to be forward and backward compatible.

For now, I _think_ I've gotten it working fairly well by applying the transformation a second time for both the block supports and deprecations in https://github.com/WordPress/gutenberg/pull/63401/commits/6934b8aa39dc7fecb95a8a066c8fed5c7bc544d0. I've added a couple of tests to cover the scenario we've described here.

The implementation still assumes that plugins aren't _watching_ for current values before overriding things, but at least this gives us coverage for plugins that are attempting to switch things off.

I'll give this PR a bit more of a manual test, though, as it _seems_ right to me now, but I'm not sure if I'm thinking is right 😄

andrewserong:
Update: I've given this PR a rebase and re-tested it, and I believe it's still working as expected, so this is ready for review.

A couple of notes:

* The backport PR in core (https://github.com/WordPress/wordpress-develop/pull/7069) takes a slightly different approach for the PHP implementation so that it doesn't rely on the `register_block_type_args` filter.
* The approach in this PR goes to some effort to still support plugins intercepting the block supports. So it should still be possible for plugins to inject experimental typography supports via the `register_block_type_args` filter in PHP and have those supports be stabilized in time. Also the JS equivalent discussed in this thread (https://github.com/WordPress/gutenberg/pull/63401#discussion_r1706049481) is handled to some extent.

At this point, I think the main question to address before landing is whether or not the approach here gives _enough_ support for plugins that might be attempting to either opt-out or opt-in to these block supports. Does this PR cover the majority of cases, or does it still feel risky?

I think the answer to that question will likely determine whether we think this is good to attempt for WP 6.7. For full disclosure, if this PR and the backport lands, I might have limited time to work on follow-ups, but happy to help review where I can.

packages/blocks/src/store/process-block-type.js
```diff
@@ -102,13 +124,20 @@ export const processBlockType =
...
 			'blocks.registerBlockType',
 			blockType,
 			name,
 			null
 		);
 
+		// Re-stabilize any experimental supports after applying filters.
+		// This ensures that any supports updated by filters are also stabilized.
+		blockType.supports = stabilizeSupports( blockType.supports );
```
gziolo, line 139: This util is useful as it needs to run a few times. I see it only does any processing when they `typography` object is provided, so it should be fast enough.

aaronrobertshaw (commented):
Nice work iterating on this one @andrewserong, there are definitely some tricky edge cases to reconcile here.

While giving this another test and review, a couple of questions came up.

1. We can't control what support keys plugins might filter on. How can we handle which  key overrides which (should a stable key values always override experimental or vice versa)?
    - I don't think we can solve this issue completely
    - It's also not clear how much of a real world problem this is
    - Can adequate testing within Gutenberg, community outreach, and dev notes mitigate enough of the risk? 
2. The state of block support keys is different between PHP and JS.
    - On the JS side, this PR stabilizes the support keys, applies filters, then re-stabilizes. This allows plugins to lean into stabilized keys as desired.
    - On the PHP side, the support keys are only stabilized after filters.
    - To avoid confusion and help move towards only using stabilized keys, I think making the PHP side also stabilize keys before and after other filters might be good.

Given the concerns around not being able to be 100% sure which keys should override which after filters, this might benefit from an extended run in Gutenberg before landing in core. What do you think about aiming to land this very early in the 6.8 cycle?

andrewserong:
Thanks for reviewing and for the feedback @aaronrobertshaw!

> Given the concerns around not being able to be 100% sure which keys should override which after filters, this might benefit from an extended run in Gutenberg before landing in core. What do you think about aiming to land this very early in the 6.8 cycle?

Yes, I'm leaning that way now, too. I think because we don't _really_ know the impact yet of this PR, it'd be good to land it early in the 6.8 cycle in Gutenberg (as you mention) where we'd have longer to test it out and catch any edge cases before attempting to merge it in core. My feeling is that it's likely in pretty good shape, but without that extra window for testing, we won't know for sure.

FYI @annezazu and @justintadlock as I know you were both interested in this change. We're proposing punting this change to 6.8 as it's still a little unknown what the impact of the change will be, and there's a bit of uncertainty surrounding the PHP implementation that would be good to resolve.

justintadlock:
It sounds like the best decision (waiting for early 6.8 cycle). We can potentially give this PR a call out for early testing in What's new for developers? too: https://github.com/WordPress/developer-blog-content/issues/309

annezazu:
Thank you for the proactive ping. Like Justin, that also sounds wise!

andrewserong:
Thanks folks! I've moved this to "Punted to 6.8" and will update the PR again early in (or just prior to) the 6.8 cycle 👍

gziolo:
It would be great to land it early in the release cycle now it has been rebased with the latest `trunk` 😄

andrewserong:
> It would be great to land it early in the release cycle now it has been rebased with the latest trunk 😄

Indeed! I'll be going on sabbatical in a couple of weeks, so I'd love to get this merged before then.

ramonjd (approved):
✅  Stable typography properties work as expected in block JSON
✅  When disabling properties using the `register_block_type_args` filter, both `__experiemental*` and their corresponding stable properties work 
✅  UI controls function as expected in post and site editors for stable and experimental properties

LGTM

As @gziolo said, it's early in the cycle so I think we can merge and monitor. 

Great work!

andrewserong:
Thanks for the review @ramonjd! 🙇 

Out of an abundance of caution and since I'm wrapping up for the week shortly, I'll leave this open until Monday morning. If there are no objections, I'll intend to merge on Monday. Thanks again for all the feedback and discussion here, everyone!

packages/blocks/src/store/process-block-type.js
```diff
@@ -62,6 +66,24 @@ function mergeBlockVariations(
 	return result;
 }
 
+function stabilizeSupports( rawSupports ) {
```
coder-karen, line 69: The `stabilizeSupports` function in Gutenberg expects `supports` to be part of a larger block configuration object that doesn’t have getter-only behavior. If `supports` is exported as a standalone object - for example with getter-only behavior - `stabilizeSupports` won’t function as expected.

For example, in Jetpack's case, four blocks were exporting `supports` in a way that led to console errors like "TypeError: Cannot set property `supports` of #<Object> which has only a getter," causing editor load issues (both the site editor and post / page editors will not load at all, if the Jetpack blocks setting is enabled).

We can fix this export approach on our end, but it'd only work for anyone with the version of Jetpack with the fix. I’m also concerned this could affect other plugins as well. Would it be possible to add a conditional check to help prevent similar issues with other plugins using exports like this (and previous versions of plugins)?

packages/blocks/src/store/process-block-type.js
```diff
@@ -119,29 +148,38 @@ export const processBlockType =
...
-							},
-							blockType.name,
-							deprecation
-						)
-					).filter( ( [ key ] ) =>
+			settings.deprecated = settings.deprecated.map( ( deprecation ) => {
+				// Stabilize any experimental supports before applying filters.
+				deprecation.supports = stabilizeSupports(
+					deprecation.supports
+				);
```
andrewserong, line 155: To address the issue that @coder-karen mentioned, I wonder if we need to do the `{ ...deprecation }` spread before we update the `supports` key here, and if that would fix the issue for Jetpack 🤔

packages/blocks/src/store/process-block-type.js
andrewserong: Thanks for flagging this @coder-karen, this is just the kind of feedback we were hoping for by landing this change in the Gutenberg plugin, so we can account for any particular cases in plugins before this change makes it into a major WP release.

> Would it be possible to add a conditional check to help prevent similar issues with other plugins using exports like this (and previous versions of plugins)?

Can you give me an example of the kind of conditional check you have in mind? How might we mutate the value in a way that's compatible with what's happening in the plugin? Or, do you mean that we would skip any mutations altogether if we only have a getter?

From looking over the code, as far as I can tell, I think the culprit might be here: https://github.com/WordPress/gutenberg/pull/63401#discussion_r1833475175 — where we're mutating `deprecation.supports` instead of creating a new object, which we do slightly further down. So I'm wondering if a fix could be creating a new object with the updated `supports` key instead of mutating `deprecation` directly.

FYI: @aaronrobertshaw for visibility. (I'll look into a fix, but just a heads-up since we've been chatting about the block supports stabilization lately)

ramonjd: > I'll look into a fix, but just a heads-up since we've been chatting about the block supports stabilization lately

Happy to test later. It would be good to have a local test condition that reproduces the recursion. 

Is this the relevant Jetpack PR?

- https://github.com/Automattic/jetpack/pull/40070

andrewserong: > Happy to test later. It would be good to have a local test condition that reproduces the recursion.

Indeed! Yes, let's see if we can fix while also providing a test to cover this case.

> Is this the relevant Jetpack PR?

That seems to be the one as far as I can tell. Hopefully if we fix this in GB that PR won't be required 🤞

andrewserong: I've put up a potential fix over in https://github.com/WordPress/gutenberg/pull/66849

coder-karen: Your solution looks ideal here, and works well for Jetpack, thank you!

aaronrobertshaw:
Quick note for added exposure. The current plan is to backport all the block support stabilization in a single core PR. More info can be found in https://github.com/WordPress/gutenberg/pull/66918#issuecomment-2472374805

aaronrobertshaw:
Quick update:

As discussed in https://github.com/WordPress/gutenberg/pull/67729#issuecomment-2553200639, the approach to block support stabilization needs more time so will likely be punted from 6.8. There is a revert PR available (https://github.com/WordPress/gutenberg/pull/68163) and anyone interested in stabilized block supports is advised to follow that for the latest 🙏

PR #63404, opened 2024-07-11
Query Loop Block: remove Posts List variation

ellatrix:
## Why?

As mentioned [here](https://github.com/WordPress/gutenberg/pull/63380#discussion_r1673097917), this is causing more harm than good because when you insert this block, the query won't inherit the main query by default.

This block variation is also not very useful for showing recent posts in the sidebar, because it has pagination etc. It's not designed for that. So it seems this variation does not have any good use cases, or it has bad defaults.

### DIFF
--- a/packages/block-library/src/query/variations.js
+++ b/packages/block-library/src/query/variations.js
@@ -2,7 +2,6 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { postList } from '@wordpress/icons';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


ntsekouras:
As per my comment [here](https://github.com/WordPress/gutenberg/pull/63380#discussion_r1673549906), we shouldn't remove this if we don't have an alternative for this one, like dynamic registration of variations for post types.

priethor (approved):
Works as expected ✅ 

Without this PR, the `Latest posts` block variation can be added and it ignores the `Inherit query from template` setting entirely, causing confusion.

With this PR, the `Latest posts` block variation doesn't appear on the inserter, and no content is lost if there are instances already on your site.

priethor:
> we shouldn't remove this if we don't have an alternative for this one

I personally don't think that is a blocker; there is no harm in the form of content loss by disabling the variation until we have an alternative that works as expected for users.

I gave an approval but I don't feel strongly if we don't want to remove it in the end.

ellatrix:
There's some discussion here as well: #63353. The alternative could be to rename the block, but it's tricky to say in a few words what not inheriting means. Even "latest" or "recent" posts could be confusing: people might think that's the blog.

This block variation is also not very useful for showing recent posts in the sidebar, because it has pagination etc. It's not designed for that. So I don't get the purpose of this variation at all.

PR #63403, opened 2024-07-11
Elements: Avoid specificity bump for top-level element-only selectors

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/63345

## What?

With the levelling of CSS specificity in https://github.com/WordPress/gutenberg/pull/61638, it resulted in a bump in specificity for top level element-only selectors e.g. links, h1-6, etc. This PR avoids wrapping such selectors to maintain the previous specificity in <=6.5.

## Why?

If themes defined element styles that used only elements in the selector, the `0-1-0` from the `:root :where()` wrapped selectors would override them. This allows selectors such as `header nav ul li a` to continue to take precedence over our default core top level element styles.

## How?

Avoids wrapping the top-level element selectors in `:root :where()` unless those element selectors already used class level selectors. Pseudo selectors, e.g. `:hover` or `:focus`, are class-level selectors as well and this PR also makes sure they are capped at `0-1-0` for correct nesting of block style variations etc.

## Testing Instructions

After each step below, make sure that the same styles are applied on the frontend.

1. Add test nav structure/markup using custom HTML block
```html
<header>
  <nav>
     <ul>
       <li>
         <a>Test Link</a>
       </li>
     </ul>
  </nav>
</header>
```
2. Test custom global element styles including hover styles; Global Styles > Colors > Link 
3. Add custom css using element only selector `header nav ul li a`. You'll need to toggle the custom HTML block preview for it to pick up the custom CSS.
```css
header nav ul li a {
    color: red;
    text-decoration: none;
}
```
4. Test global block type element styles; Global Styles > Blocks > Group > Elements > Link
5. Test element style overrides on block instances
6. Define multiple block style variations that define both `styles.elements` and `styles.blocks.core/group.elements.link` styles. The following snippets can be used to create theme.json partial files within the theme's `/styles` directory which will then be registered as block style variations.
<details>
<summary>Example Block Style Variation Partials</summary>

```json
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 3,
	"title": "Custom Links",
	"slug": "custom-links",
	"blockTypes": [ "core/group" ],
	"styles": {
		"elements": {
			"link": {
				"color": {
					"text": "crimson"
				},
				":hover": {
					"color": {
						"text": "darkmagenta"
					}
				}
			},
			"button": {
				"color": {
					"background": "purple",
					"text": "white"
				},
				":hover": {
					"color": {
						"background": "darkorchid"
					}
				}
			}
		},
[...]

PR #63406, opened 2024-07-11
Navigation block: Allow themes to override block library text-decoration rule

talldan:
## What?
In a default install of WordPress 6.5, when hovering over a navigation link an underline is shown.

In 6.6 this doesn't happen. Theme developers are unable to override block library styles for text decoration.

## Why?
In 6.5, there's a CSS rule that sets the underline hover with the specificity of 0,2,1 ([it comes from the TT4's theme.json](https://github.com/WordPress/wordpress-develop/blob/349590c8135b003857b47ff1a4636dc283d8b71c/src/wp-content/themes/twentytwentyfour/theme.json#L525-L541)):
```css
.wp-block-navigation a:where(:not(.wp-element-button)):hover {
    text-decoration: underline;
}
```

It overrides some of the block library css (specificity 0,1,1):
```css
.wp-block-navigation:where(:not([class*=has-text-decoration])) a {
    text-decoration: none;
}
```

In 6.6, the first rule's specificity was reduced to 0,1,0, and now the block library css has a higher specificity it results in that taking precedence and no hover underline styles.

## How?
This reduces the specificity of the block library css to 0,1,0, following the general rule that theme.json/global styles should be able to override block css.

#### Simple test of hover styles
1. View the frontend of a default WordPress install and hover over a navigation link

Expected: the link should be underlined

#### Active and Focus styles
1. Activate emptytheme
2. Add active/focus styles to link elements in the nav block within your theme.json using the snippet below
3. Add nav block to a page without customizing its styles
4. Load the page on the frontend

Expected: Navigation block links have line-through styles when focused or active

```json
{
	"version": 3,
	"settings": {
		"appearanceTools": true,
		"layout": {
			"contentSize": "840px",
			"wideSize": "1100px"
		}
	},
	"styles": {
		"blocks": {
			"core/navigation": {
				"elements": {
					"link": {
						":focus": {
							"typography": {
								"textDecoration": "line-through"
							}
						},
						":active": {
							"typography": {
								"textDecoration": "line-through"
							}
						}
					}
				}
			}
		}
	}
}
```

### Before
Screenshot 2024-07-11 at 4 33 05 PM

### After
Screenshot 2024-07-11 at 4 33 17 PM

### DIFF
--- a/packages/block-library/src/navigation/style.scss
+++ b/packages/block-library/src/navigation/style.scss
@@ -76,14 +76,12 @@ $navigation-icon-size: 24px;
 		}
 	}
 
-	&:where(:not([class*="has-text-decoration"])) {
-		a {
-			text-decoration: none;
+	&:where(:not([class*="has-text-decoration"])) :where(a) {
+		text-decoration: none;
 
-			&:focus,
-			&:active {
-				text-decoration: none;
-			}
+		&:focus,
+		&:active {
+			text-decoration: none;
 		}
 	}
 
--- a/packages/block-library/src/navigation/style.scss
+++ b/packages/block-library/src/navigation/style.scss
@@ -76,12 +76,10 @@ $navigation-icon-size: 24px;
 		}
 	}
 
-	&:where(:not([class*="has-text-decoration"])) {
-		& :where(a),
-		& :where(a:focus),
-		& :where(a:active) {
-			text-decoration: none;
-		}
+	& :where(a),
+	& :where(a:focus),
+	& :where(a:active) {
+		text-decoration: none;
 	}
 
 	// Submenu indicator.
--- a/packages/block-library/src/navigation/style.scss
+++ b/packages/block-library/src/navigation/style.scss
@@ -76,11 +76,10 @@ $navigation-icon-size: 24px;
 		}
 	}
 
-	&:where(:not([class*="has-text-decoration"])) :where(a) {
-		text-decoration: none;
-
-		&:focus,
-		&:active {
+	&:where(:not([class*="has-text-decoration"])) {
+		& :where(a),
+		& :where(a:focus),
+		& :where(a:active) {
 			text-decoration: none;
 		}
 	}


packages/block-library/src/navigation/style.scss
```diff
@@ -76,14 +76,12 @@ $navigation-icon-size: 24px;
 		}
 	}
 
-	&:where(:not([class*="has-text-decoration"])) {
-		a {
-			text-decoration: none;
+	&:where(:not([class*="has-text-decoration"])) :where(a) {
```
ellatrix: Could these be combined into a single `where`?

ellatrix: Another question: do we still need the `:not([class*="has-text-decoration"])` if the specificity is lowered? Ideally if it has that class, it should just override this rule and the exception should no longer be needed.

packages/block-library/src/navigation/style.scss
```diff
@@ -76,14 +76,12 @@ $navigation-icon-size: 24px;
...
+		text-decoration: none;
 
-			&:focus,
-			&:active {
-				text-decoration: none;
-			}
+		&:focus,
+		&:active {
```
aaronrobertshaw: These pseudo styles get appended so the resulting specificity is `0-2-0` which is still higher than global styles. 

To my knowledge there isn't an easy way in SCSS to insert the inner `:focus`/`:active` into a CSS function in the parent selector i.e. `:where()`. So I think they need to be written out in full.

```scss
	&:where(a),
	&:where(a:focus),
	&:where(a:active) {
		text-decoration: none;
	}
```

packages/block-library/src/navigation/style.scss
talldan: > Could these be combined into a single where?

I had already tried this:
```css
&:where(:not([class*="has-text-decoration"]) a)
```

But that selector doesn't work. Is there another way?

aaronrobertshaw (commented):
Thanks for putting this together @talldan 👍 

This tests as advertised for me regarding the `:hover` styles set on link elements in the nav block like TT4 does.

The `:focus` and `:active` pseudo classes still bump the specificity and override anything set for those link pseudo selectors in theme.json.

For the most part it appears that setting the text decoration at the block level (not elements) still works due to the way browsers render text decoration differently.

I primarily tested using both emptytheme and TT4. To replicate the issue with `:focus` and `:active`:

1. Activate emptytheme
2. Add active/focus styles to link elements in the nav block within your theme.json using the snippet below
3. Add nav block to a page without customizing its styles
4. Load the page on the frontend and confirm the `text-decoration: none` rule still overrides the global styles.

```json
{
	"version": 3,
	"settings": {
		"appearanceTools": true,
		"layout": {
			"contentSize": "840px",
			"wideSize": "1100px"
		}
	},
	"styles": {
		"blocks": {
			"core/navigation": {
				"elements": {
					"link": {
						":focus": {
							"typography": {
								"textDecoration": "line-through"
							}
						},
						":active": {
							"typography": {
								"textDecoration": "line-through"
							}
						}
					}
				}
			}
		}
	}
}
```

| This PR | With Suggested Tweak |
|---|---|
| <video src="https://github.com/user-attachments/assets/c9d194e6-3bdf-4d5c-952d-1caaff421adc" /> | <video src="https://github.com/user-attachments/assets/ef281529-f421-4aa9-ab2f-293c3c0452e1" /> |

packages/block-library/src/navigation/style.scss
talldan: > Another question: do we still need the :not([class*="has-text-decoration"]) if the specificity is lowered? Ideally if it has that class, it should just override this rule and the exception should no longer be needed.

It does seem needed, it overrides a general rule for links that are not buttons which gives them underlines:
Screenshot 2024-07-12 at 3 24 33 PM

That should be a bit more obvious when testing now that I've rebased to include the fix from #63403.

talldan: I've applied Aaron's suggestion now so it looks a little different - https://github.com/WordPress/gutenberg/pull/63406#discussion_r1675405046

packages/block-library/src/navigation/style.scss
talldan: Good point, updated in [0f9669a](https://github.com/WordPress/gutenberg/pull/63406/commits/0f9669a834c465c983b50376b065f61320e3ad69)

packages/block-library/src/navigation/style.scss
aaronrobertshaw: > It does seem needed, it overrides a general rule for links that are not buttons which gives them underlines:

Removing the `:where(:not([class*=has-text-decoration]))` won't change the specificity. So the selector would remain `0-1-0` specificity and continue to override `a:where(:not(.wp-element-button))` which is only `0-0-1`.

packages/block-library/src/navigation/style.scss
```diff
@@ -77,13 +77,10 @@ $navigation-icon-size: 24px;
 	}
 
 	&:where(:not([class*="has-text-decoration"])) {
-		a {
+		& :where(a),
+		& :where(a:focus),
+		& :where(a:active) {
```
aaronrobertshaw: Sorry, I should have been more explicit with my suggestion in https://github.com/WordPress/gutenberg/pull/63406#discussion_r1675405046. It also took into account the suggestion made in https://github.com/WordPress/gutenberg/pull/63406#discussion_r1673862160.

The wrapping `&:where(:not([class*="has-text-decoration"]))` is not required.

aaronrobertshaw (commented):
I've retested this and can confirm the `:focus` and `:active` styles have been fixed via https://github.com/WordPress/gutenberg/pull/63406/commits/0f9669a834c465c983b50376b065f61320e3ad69.

Also, I believe the `wrapping &:where(:not([class*="has-text-decoration"]))` is not required. I should have been clearer in my review and provided a full diff. 

<details>
<summary>Updated suggestion after `0f9669a`</summary>

```diff
diff --git a/packages/block-library/src/navigation/style.scss b/packages/block-library/src/navigation/style.scss
index 734e3bd04c..3d7a8810ae 100644
--- a/packages/block-library/src/navigation/style.scss
+++ b/packages/block-library/src/navigation/style.scss
@@ -76,12 +76,10 @@ $navigation-icon-size: 24px;
 		}
 	}
 
-	&:where(:not([class*="has-text-decoration"])) {
-		& :where(a),
-		& :where(a:focus),
-		& :where(a:active) {
-			text-decoration: none;
-		}
+	& :where(a),
+	& :where(a:focus),
+	& :where(a:active) {
+		text-decoration: none;
 	}
 
 	// Submenu indicator.

```
</details>

Using that suggested change, here's what I get using emptytheme and no theme.json block type element styles

_Note: Core element styles are correctly overridden by the block library's link styles._

https://github.com/user-attachments/assets/9168511e-129d-4eab-8f94-5285dfd00460

When defining link element styles for the navigation block type in theme.json, this is what I get.

_Note: Global Styles for navigation block type link element are applied, overriding the block library styles._

https://github.com/user-attachments/assets/38f1a5b8-5c04-4082-ab4f-20777f55d662

<details>
<summary>Example theme.json styles</summary>

```json
{
	"version": 3,
	"settings": {
		"appearanceTools": true,
		"layout": {
			"contentSize": "840px",
			"wideSize": "1100px"
		}
	},
	"styles": {
		"blocks": {
			"core/navigation": {
				"elements": {
					"link": {
						"typography": {
							"textDecoration": "none"
						},
						":hover": {
							"typography": {
								"textDecoration": "underline"
							}
						},
						":focus": {
							"typography": {
								"textDecoration": "line-through"
							}
						},
						":active": {
							"typography": {
								"textDecoration": "none"
							}
						}
					}
				}
			}
		}
	}
}
```
</details>

Applying block instance styles for text decoration continue to take precedence over the block library and global styles due to their compound class selector e.g. `.wp-block-navigation.has-text-decoration-line-through`

packages/block-library/src/navigation/style.scss
talldan: Oh, I misunderstood what was meant in the original comment. I've gone ahead and removed that. 👍

aaronrobertshaw (approved):
Thanks for the continued iteration here 👍 

This is testing well for me now.

The default theme.json element styles are overridden by block library styles, which are in turn correctly overridden by block type element global styles. Applying text decoration on the block instance then takes precedence over all the others.

Given the chance this might need to be included in a WP 6.6.1 point release, it would be good to update the PR test instructions with more detail for others coming to this in the near future.

Screenshot 2024-07-17 at 12 59 34 PM

andrewserong:
I've added the `Backport to Gutenberg RC` label and will cherry pick this in for GB 18.8.

Is it worth adding the `Backport to WP Beta 6.6/RC` label, too, to get it into the `wp/6.6` branch?

andrewserong:
I just cherry-picked this PR to the release/18.8 branch to get it included in the next release: 976df1e406e

talldan:
> Is it worth adding the Backport to WP Beta 6.6/RC label, too, to get it into the wp/6.6 branch?

I think usually it'd be cherry-picked to that branch as part of the `Backport to WP Minor Release` process.

I saw there's a slack thread discussing this issue - https://wordpress.slack.com/archives/C02QB2JS7/p1721166077452989, though I'm not sure you need to actually merge into the 6.6 branch to test, you can possibly make a PR that merges into the branch that includes the fix, and a PR usually has a job that builds the plugin zip.

andrewserong:
A separate PR to be on the safe side sounds good, that way folks can grab a zip file from that build if they need to. I've opened up a PR targeting the `wp/6.6` branch over in https://github.com/WordPress/gutenberg/pull/63641 if you'd like to take a look / confirm it's working as expected in that branch.

I figure that at least gives folks working on a point release enough options to help ensure this fix gets included 🙂

aicuaboom:
Bh01

packages/block-library/src/navigation/style.scss
aicuaboom: GALL@

andrewserong:
Just giving this PR an explicit 👍 and ✅ from me as resolving the bug and for inclusion in a WP 6.6 point release.

@ellatrix feel free to either land https://github.com/WordPress/gutenberg/pull/63641 as the merge into `wp/6.6` or cherry pick this PR. Both now have multiple 👍 / approving reviews from folks. Thanks!

Issue #43245, opened 2022-08-16
Tracking: Color Design Tools Consistency


aaronrobertshaw:
## Overview

This issue details the current state of color block support or design tool adoption across all blocks and tasks required to fill any gaps. Overall design tool consistency efforts are being tracked via the parent issue: https://github.com/WordPress/gutenberg/issues/43241.

## Legend

| Value         | Description                                                                  |
| ------------- | ---------------------------------------------------------------------------- |
| ✅            | Feature has been adopted and is displayed as a **default** control |
| ✅ (Optional) | Feature has been adopted but is an **optional** control            |
| ⚠             | There is a **bug or issue** with this block support feature's adoption           |
| ❌            | Feature has been explicitly **opted out** of           |
| 🚫 | Block support will not be adopted for this block | 
| <PR#>         | Links to **PR** adopting the feature for this block                |
| -             | Feature has **not explicitly** been adopted/omitted                              |
| 🛠             | Implemented via an **ad hoc** / bespoke control                                  |
| 🚧 | Work is in progress towards adopting this feature (no PR yet) |

## Known Issues

Recently the Color gradient settings dropdown was refactored. Part of this update [made controls show by default](https://github.com/WordPress/gutenberg/pull/41091/files#diff-1d8706a429c020115759594b52150542acec1b47a901b4c09163f3a19b78f9faR40-R44) unless explicitly flagged as optional. 

This goes against the way all other block support controls work, isn't documented as such, and is counter to the design goal for reducing all unnecessary clutter and controls from the sidebar. It does however assist consistency. This should be removed so the controls show by default as instructed via the block.json file. We can update blocks to make the default controls consistent and will be done in Phase 2 as outlined in https://github.com/WordPress/gutenberg/issues/43241. 

When we address potentially defining a consistent set of default controls per block support, the code linked wouldn't be required as it would be handled elsewhere.

## Block Support Adoption

_Note: Deprecated blocks have been omitted from this table. e.g. Comment Author Avatar, Post Comment & Text Columns._

| Block                        | Text          | Background    | Gradients | Link | Heading | Button | Duotone |
| ---------------------------- | ------------- | ------------- | --------- | ------------- | ------- |  ------- |  ------- |
| Archives                     | ✅ | ✅ | ✅ | ✅ | 🚫 | 🚫 | -       |
| Audio                        | -             | -             | -         | -             |  🚫 | 🚫 |- |
[...]

PR #63407, opened 2024-07-11
Categories: Add color block support

carolinan:
## What?

Adds text, background, gradient, and link color support to the categories (Categories list) block.
Background, text and link are set to show by default.

Partial for https://github.com/WordPress/gutenberg/issues/43245

## Why?

## How?

The PR updates the block supports in block.json, and the documentation.

## Testing Instructions
Add some categories.
Create a new post and add the categories list block. Adjust the color settings and compare the result in the editor and front.


carolinan:
@aaronrobertshaw

Help wanted: I am getting inconsistent results in block themes vs classic themes and depending on where the block is placed.

### Block themes
#### Block editor & Site Editor
The block inherits the padding from the UL element:
```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```
But the left and right padding is then removed by:
```
:root :where(.wp-block-categories) {
    padding-right: 0px;
    padding-left: 0px;
}
```
So the block has a visible top and bottom padding.

#### Single post, front
Here the block does not inherit the padding from the UL element, so the block has **no** padding.

####  Home / frontpage, front
```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```
But the left and right padding is then removed by:
```
:root :where(.wp-block-categories) {
    padding-right: 0px;
    padding-left: 0px;
}
```
So the block has a visible top and bottom padding.


### Classic themes, block editor and front:
The block inherits the padding from the UL element:
```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```
But the left and right padding is not removed, so compared to the block themes, it has visible padding on all four sides.

aaronrobertshaw:
Thanks for the ping @carolinan 👍 

I've only had the chance to take a quick look at what's happening but it appears there could be a few different issues resulting in the behaviour you're observing.

#### List Block Styles

The first is that the `:root :where(ul.has-background,ol.has-background)` style is bleeding in from the List block's styles. Your recent [PR](https://github.com/WordPress/gutenberg/pull/56469) fixing the List block to include a class name should give us something to hook onto to prevent this.

There might be a catch though, if we simply update the List block's selectors to replace the `ul`/`ol` elements with the `.wp-block-list` class, other blocks that are list based with `.has-background` classes could lose padding.

The List block has been problematic from the get go with the overly simple and short-sighted selectors. It's moving in the right direction now it can be styled separately to general list elements. Having it not bleed into other block/list styles is the right move in my view. We might just need to work out what the impact of that might actually be and communicate the change well.


#### Separate Loading of Block Assets


> I am getting inconsistent results in block themes vs classic themes and depending on where the block is placed.

This one is a little trickier to debug without knowing the active theme, templates, and other blocks on the page. It could be possible that if separate assets are loaded for blocks via [`should_load_separate_core_block_assets`](https://developer.wordpress.org/reference/hooks/should_load_separate_core_block_assets/), the block styles bleeding into the categories block are only sometimes loaded leading to the inconsistency witnessed.

> Single post, front
>
> Here the block does not inherit the padding from the UL element, so the block has no padding.

Could this one be due to the issue above and the List block isn't on the page and block assets are loading separately?

From the other styles quoted in your comment, I'm assuming you're using TT4? I believe that loads the block styles separately and might explain it.

> Home / frontpage, front
> 
> But the left and right padding is then removed by:

Does this page have a List block?

> Classic themes, block editor and front:
>
> But the left and right padding is not removed, so compared to the block themes, it has visible padding on all four sides.

The left/right padding styles setting them to `0` comes from TT4's theme.json, I believe. so it would be expected.

The padding on all sides coming from the List block is what I think is unexpected though.


I hope that helps some 🤞

aaronrobertshaw:
I've thrown up a quick PR to fix the bleed in of List block's `.has-background` styles: https://github.com/WordPress/gutenberg/pull/63537

carolinan:
Yes I was testing with Twenty Twenty-Four and Twenty Twenty-One. I tested with the same themes again and the fix in https://github.com/WordPress/gutenberg/pull/63537 solved the padding.

Now, I created a new blank theme with the Create Block Theme, and I found a new color issue:

1) I created a new post, added the categories list block, and changed all the colors.
2) I opened the themes index template in the Site Editor. The main query loop in this theme is set to show the post content.
3) The post content block shows the wrong link color for my new post: it uses the browser default, not the color set on the block in the block editor.

----

The categories block also shows different styles for the bullets using this blank theme:
- In the site editor outside the post content, the list style type is disc.
- In the site editor post content block in the query loop, the list style type is circle.
- In the block editor, the list style type is disc, 
- On the front it is circle.
I did not research this further. It looks like it is because of the browser styles for "ul ul". I will first look if there is an issue for it already.

carolinan:
When I use the console to look at the markup for the categories block when it is inside the post content block in the Site Editor, it is missing the `.has-link-color` class and the style for the `wp-elements-X` class.

In the Site Editor post content block:
```
<ul class="wp-block-categories-list wp-elements-9da7e6dfd0742e9e2f89d713f6c75d6c wp-block-categories has-text-color has-white-color has-background has-luminous-dusk-gradient-background">
```

In the block editor:
`<ul tabindex="0" class="block-editor-block-list__block wp-block wp-block-categories-list wp-elements-0 has-white-color has-luminous-dusk-gradient-background has-text-color has-background has-link-color wp-block-categories" id="block-5a031786-bb29-4f35-941a-6327ba73272a" role="document" aria-label="Block: Categories List" data-block="5a031786-bb29-4f35-941a-6327ba73272a" data-type="core/categories" data-title="Categories List" style="color: rgb(255, 255, 255);">`

CSS in the block editor:
```
.editor-styles-wrapper .wp-elements-0 a:where(:not(.wp-element-button)) {
    color: var(--wp--preset--color--vivid-red);
}
```

aaronrobertshaw:
> Now, I created a new blank theme with the Create Block Theme, and I found a new color issue:

I can replicate this issue.

It appears more related to the Post Template block and how it renders each post's content. I can see that element styles for the block don't get applied despite there being a `.wp-elements-<hash>` class.

When a text color only (no link color) is set on the block, the inner links normally inherit that however due to a direct style on the anchor tag by the user agent stylesheet it takes precedence. This only seems to be an issue in the editor with the Query/Post Template block. Maybe the block could do with a default style to ensure inheritance when any is set.

I don't think this is an issue with this PR directly. Perhaps if a dedicated issue is created it can be addressed separately. I have a suspicion the categories list block isn't the only one to be impacted by this bug.

> The categories block also shows different styles for the bullets using this blank theme:

The bullet style in both cases falls back to the user agent stylesheet. The difference then comes in that when the categories list is within the query block which itself is a list, we end up with a nested list getting the `disc` style bullet.

PR #63410, opened 2024-07-11
Fix ComboboxControl reset button when using the keyboard.

### LINKED ISSUE #63408: ComboboxControl: the Reset button doesn't work when using the keyboard
### Description

Discovered while testing the 'Parent' setting for a post of type page.

Screenshot 2024-07-11 at 10 06 45

Can be reproduced on the [storybook](https://wordpress.github.io/gutenberg/?path=/story/components-comboboxcontrol--default&args=allowReset:!true) as well, by enabling the `allowReset` prop.

When using the keyboard, pressing the `Enter` key on the optional Reset button doesn not reset the combobox. Pressing the Spacebar key does work.

Turns out when pressing the `Enter` key, the `keydown` event attached on the container div element fires and actually _sets_ the currently selected suggestion.

- Edit a page and use the `Parent` setting with the keyboard.
- Alternatively, test on the storybook, make sure the `allowReset` prop is set to `true`: https://wordpress.github.io/gutenberg/?path=/story/components-comboboxcontrol--default&args=allowReset:!true
- Select a suggestion.
- Click the reset button with your mouse.
- The combobox is reset.
- I would expect the same behavior when using the keyboard.
- Select a suggestion.
- Move focus to the Reset button.
- Press the Spacebar key and observe it does reset the combobox.
- Repeat, but this time press the Enter key and observe it does _not_ reset the comobox. 

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63408

## What?

The combobox Reset button doesn't work with the Enter key.

## Why?

All UI controls should work with the keyboard.

## How?

Prevents propagation of the keydown event.

## Testing Instructions

- Repeat the test instructions from the issue https://github.com/WordPress/gutenberg/issues/63408
- Observe the pressing the Enter key on the Reset button does reset the comobox.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+-   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
 
 ### Internal
@@ -33,7 +34,6 @@
 
 ### Bug Fixes
 
--   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Tabs`: Make Tabs have a fluid height ([#62027](https://github.com/WordPress/gutenberg/pull/62027)).
 -   `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
 -   `useUpdateEffect`: Correctly track mounted state in strict mode. ([#62974](https://github.com/WordPress/gutenberg/pull/62974))
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -33,6 +33,7 @@
 
 ### Bug Fixes
 
+-   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Tabs`: Make Tabs have a fluid height ([#62027](https://github.com/WordPress/gutenberg/pull/62027)).
 -   `UnitControl`: Fix colors when disabled. ([#62970](https://github.com/WordPress/gutenberg/pull/62970))
 -   `useUpdateEffect`: Correctly track mounted state in strict mode. ([#62974](https://github.com/WordPress/gutenberg/pull/62974))
--- a/packages/components/src/combobox-control/test/index.tsx
+++ b/packages/components/src/combobox-control/test/index.tsx
@@ -89,7 +89,6 @@ describe.each( [
 			<Component options={ timezones } label={ defaultLabelText } />
 		);
 		const label = getLabel( defaultLabelText );
-		expect( label ).toBeInTheDocument();
 		expect( label ).toBeVisible();
 	} );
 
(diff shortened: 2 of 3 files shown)


packages/components/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 
 ### Bug Fixes
 
+-   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
```
ciampo: We will need to move this CHANGELOG entry to the `Unreleased` section above, in a new `Bug fixes` sub section

ciampo (approved):
Thank you for working on this, @afercia!

Code changes LGTM. Would you be able to also add a unit test for the reset behavior, so that we don't incur in future regressions?

afercia:
Tanks @ciampo. Sure will work on an ~e2e~ unit test tomorrow and adjust the changelog.

ciampo:
> Sure will work on an e2e test tomorrow and adjust the changelog.

A unit test is better suited — we already have unit tests for `ComboboxControl` in `packages/components/src/combobox-control/test/index.tsx`, so you could just add a new one.

You would then be able to run the test suite with the command `npm run test:unit packages/components/src/combobox-control`. Happy to help if you need!

afercia:
@ciampo I added a few unit tests. Confirmed the specific test for the `Enter` key fails without the fix in this PR.
Thanks for your feedback and review.

afercia:
Note: I was surprised the Testing Library syntax to press the Spacebar is `await user.keyboard( ' ' );`.
I would've expected something like `await user.keyboard( '{Spacebar} );`
Is that correct? Any other more elegant syntax to use?

ciampo:
> Is that correct? Any other more elegant syntax to use?

`await user.keyboard( '[Space]' )` should work

packages/components/src/combobox-control/test/index.tsx
```diff
@@ -306,4 +306,153 @@ describe.each( [
...
+				options={ timezones }
+				label={ defaultLabelText }
+				allowReset
+			/>
+		);
+
+		const resetButton = screen.getByRole( 'button', { name: 'Reset' } );
+
+		expect( resetButton ).toBeInTheDocument();
+		expect( resetButton ).toBeVisible();
```
ciampo:
```suggestion
		expect( resetButton ).toBeVisible();
```

`toBeVisible()` should already imply that the element is in the document

packages/components/src/combobox-control/test/index.tsx
```diff
@@ -306,4 +306,153 @@ describe.each( [
...
+		await user.keyboard( getOptionSearchString( targetOption ) );
+		// Pressing Enter/Return selects the currently focused option.
+		await user.keyboard( '{Enter}' );
+
+		const resetButton = screen.getByRole( 'button', { name: 'Reset' } );
+
+		expect( resetButton ).toBeInTheDocument();
+		expect( resetButton ).toBeVisible();
+		expect( resetButton ).toBeEnabled();
+	} );
```
ciampo: This test feels superflous, since the next three tests (click/enter/spacebar on the button) would also fail if the button was disabled. At most we can add the `expect( resetButton ).toBeEnabled()` check in each test?

afercia: OK

packages/components/src/combobox-control/test/index.tsx
afercia: Yup, I just copied and pasted from the existing test `should render with visible label`, will fix it there as well

afercia:
Thanks, I think I addressed all the points.

PR #63411, opened 2024-07-11
Update getContext() usage examples with namespace argument

### LINKED ISSUE #63366: Docs: Interactivity API - document  getContext() takes an optional namespace argument
`getContext()` takes an optional namespace argument ([link](https://github.com/WordPress/gutenberg/blob/82c1bc209ddd955025e2866ca449f6a269f4ecf7/packages/interactivity/src/hooks.tsx#L142-L143))

So for example, in any action, we can do:
```
const regionContext = getContext('region')
```

This should be added here (is not properly documented)→ https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/api-reference/#store-client-methods

Reported [via Slack](https://wordpress.slack.com/archives/C071CRKGKUP/p1720626038261409?thread_ts=1720614519.588419&cid=C071CRKGKUP) by @michalczaplinski

dhananjaykuber:
Fixes [#63366](https://github.com/WordPress/gutenberg/issues/63366)

## What?
This PR updates the documentation for the getContext() function to separate usage examples with and without namespace arguments. It provides clear, distinct examples and explanations for each case.

## Why?
This PR addresses the need for clearer documentation on how to use the getContext() function both with and without the optional namespace argument. It helps developers understand the different use cases and how to retrieve specific parts of the context, improving the overall usability and clarity of the documentation.

## How?
Adding a section for using getContext(namespace) with a namespace argument.

### DIFF
--- a/docs/reference-guides/interactivity-api/api-reference.md
+++ b/docs/reference-guides/interactivity-api/api-reference.md
@@ -1054,7 +1054,12 @@ Apart from the store function, there are also some methods that allows the devel
 
 #### getContext()
 
-Retrieves the context inherited by the element evaluating a function from the store. The returned value depends on the element and the namespace where the function calling `getContext()` exists.
+Retrieves the context inherited by the element evaluating a function from the store. The returned value depends on the element and the namespace where the function calling `getContext()` exists. It can also take an optional namespace argument to retrieve the context of a specific interactive region.
+
+```js
+const context = getContext(namespace);
+```
+- `namespace` (optional): A string that matches the namespace of an interactive region. If not provided, it retrieves the context of the current interactive region.
 
 ```php
 // render.php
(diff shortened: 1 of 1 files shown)


fabiankaegy (changes requested):
Thanks for adding these docs! :) 

What you have written at the moment is not how the API is working however. 

The `namespace` you pass to the `getContext` function needs to be the namespace of an interactive region (`data-wp-interactive`).

So in this case you would want to use `getContext("myPlugin")`

And the `getContext` function always gets the entire context object. There is no helper function to only get part of a context.

dhananjaykuber:
Hi @fabiankaegy,

Thank you for your feedback. I have updated the documentation to reflect the correct usage of the getContext function with the namespace of an interactive region (data-wp-interactive). Can you please review it once more to ensure everything is accurate?

docs/reference-guides/interactivity-api/api-reference.md
```diff
@@ -1054,7 +1054,12 @@ Apart from the store function, there are also some methods that allows the devel
 
 #### getContext()
 
-Retrieves the context inherited by the element evaluating a function from the store. The returned value depends on the element and the namespace where the function calling `getContext()` exists.
+Retrieves the context inherited by the element evaluating a function from the store. The returned value depends on the element and the namespace where the function calling `getContext()` exists. It can also take an optional namespace argument to retrieve the context of a specific interactive region.
+
+```js
+const context = getContext(namespace);
```
fabiankaegy:
```suggestion
const context = getContext('namespace');
```

I would suggest still wrapping the namespace here in quotes just to indicate that it is a string

fabiankaegy (approved):
Left one minor suggestion :) Otherwise this looks great 👍 

Thanks for updating it :)

Issue #63355, opened 2024-07-10
Editing Post Meta triggers multi-entity save flow and duplicate site editor save request


lgladdy:
### Description

In WordPress 6.6 RC3 (although, I replicated this on beta1 as well), editing an ACF Block is triggering the multi-entity save flow in the site editor when editing a post.

This triggers a notification that something is changing about the site (when actually, it's just in the post being edited) and accepting the save triggers two duplicate POST requests to update the post. One of which completes with `Site Updated` and one with `Post Updated` notifications in editor.

### Step-by-step reproduction instructions

My reproduction steps are with ACF Blocks, but our blocks only update the block comment within our data attribute on edit, we also trigger `wp.data.dispatch('core/editor').editPost({ meta: { _acf_changed: 1 } });` whenever an ACF field is changed so we can do some tracking for revisions. My guess is it's this call that is triggering Gutenberg to believe a site change has happened, show the multi-entity save panel and then duplicate the POST request to actually save.

### Screenshots, screen recording, code snippet

https://youtu.be/X7fR-xrKIQE (sorry about the mono audio)

### Environment info

WordPress 6.6 RC3
Chrome and Safari latest

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

No

lgladdy:
A Making WordPress slack discussion on this bug is here: https://wordpress.slack.com/archives/C02QB2JS7/p1720616434101469

lgladdy:
A second bug exists in this flow as well.

If a post is locked using `wp.data.dispatch( 'core/editor' ).lockPostSaving( 'test' );` - the save button is disabled as expected, but you can still click it to bring up the multi-entity save panel, and it will still submit the POST request which triggered `Site Updated` - although, the `Post Updated` call doesn't take place.

SantosGuillamot:
I've opened this [pull request](https://github.com/WordPress/gutenberg/pull/63412) to revert the changes that caused both issues.

lgladdy:
Resolved by #63412.

We're tracking what happens next in #63425 here.

PR #63412, opened 2024-07-11
Revert triggering multi-entity save panel in post with meta changes

SantosGuillamot:
## What?

In [this pull request](https://github.com/WordPress/gutenberg/pull/61811/), we are triggering the multi-entity save panel when users make changes to post meta, even if just that specific post has changed. That doesn't seem right and it is causing some issues:
* https://github.com/WordPress/gutenberg/issues/62236
* https://github.com/WordPress/gutenberg/issues/63355
* https://github.com/WordPress/gutenberg/issues/63355#issuecomment-2220664311

I'm reverting part of the changes of that pull request to stop triggering the multi-entity save panel when only the current post has been modified. We should think about it more broadly and potentially define a similar workflow for "single-entity" changes if wanted.

I kept the notice of post meta changes when posts are updated in a different entity, but I can revert that as well if desired.

Related conversation in WordPress Slack: [link](https://wordpress.slack.com/archives/C02QB2JS7/p1720616434101469).

## Why?
Because it is causing more issues than benefits right now.

## How?
I'm just reverting the changes that trigger the panel when post meta changes, as well as the test that was added in that PR.

## Issues

This is the before and after of the mentioned issues:

* **Multi-entity panel showing**
* **Site update message when editing a post**
* **Two requests for the same purposes when updating a custom field**
**Before**

https://github.com/WordPress/gutenberg/assets/34552881/d4734656-9f93-4929-ab8e-c841da8641d2

**After**

https://github.com/WordPress/gutenberg/assets/34552881/f8f5bfba-1baf-4556-8561-d816cc9ab0da

* **Save button is clickable after locking it**

**Before**

https://github.com/WordPress/gutenberg/assets/34552881/2e5d4e4f-a8b5-4f64-a689-bf20897c588a

**After**

https://github.com/WordPress/gutenberg/assets/34552881/0a2b3074-9375-47f7-8e5b-49a97bb0bd44

## Testing Instructions

1. Go to a page or a post.
2. Modify the value of a custom field. You can use something like this:
```
wp.data.dispatch('core/editor').editPost({ meta: { my_custom_field: 1 } });
```
3. Click on save.
4. Check that the reported issues have been solved.
You can use the following snippet to lock the button:

```
wp.data.dispatch( 'core/editor' ).lockPostSaving( 'test' )
```


lgladdy:
I've tested and confirmed this resolves the issues I reported in #63355 🥳

cbravobernal (approved):
Works as expected.

ntsekouras:
Thank you! Should we close the linked issues that are resolved with this PR?

lgladdy:
> Thank you! Should we close the linked issues that are resolved with this PR?

I closed mine as fixed and referenced the next steps ticket (#63425) - but the others probably need reviewing to see if they are solved by the rollback, or at least need to be added to #63425 so it's tracked when this flow is readded

SantosGuillamot:
I left a comment on [the remaining issue](https://github.com/WordPress/gutenberg/issues/62236), which is cover in https://github.com/WordPress/gutenberg/issues/63425

PR #63413, opened 2024-07-11
Site Editor: Deprecate 'getCanUserCreateMedia' selector

Mamaduka:
## What?
PR deprecates the `getCanUserCreateMedia` selector for the `edit-site` store. It's no longer used in the core, and there's a better selector for the same check.

Also, I've fixed the deprecation for `getReusableBlocks`. It was printing the incorrect selector name as deprecated.

## Testing Instructions
Running the following selector in the console should print the correct deprecation messages:

```
wp.data.select( 'core/edit-site' ).getCanUserCreateMedia()
wp.data.select( 'core/edit-site' ).getReusableBlocks()
```

Same.

## Screenshots or screencast 
CleanShot 2024-07-11 at 13 23 05
CleanShot 2024-07-11 at 13 25 56

### DIFF
--- a/packages/edit-site/src/store/selectors.js
+++ b/packages/edit-site/src/store/selectors.js
@@ -76,7 +76,7 @@ export const getCanUserCreateMedia = createRegistrySelector(
 			`wp.data.select( 'core/edit-site' ).getCanUserCreateMedia()`,
 			{
 				since: '6.7',
-				alternative: `wp.data.select( 'core' ).canUser()`,
+				alternative: `wp.data.select( 'core' ).canUser( 'create', { kind: 'root', type: 'media' } )`,
 			}
 		);
 
--- a/packages/edit-site/src/store/selectors.js
+++ b/packages/edit-site/src/store/selectors.js
@@ -92,13 +92,11 @@ export const getCanUserCreateMedia = createRegistrySelector(
  * @return {Array} The available reusable blocks.
  */
 export const getReusableBlocks = createRegistrySelector( ( select ) => () => {
-	deprecated(
-		"select( 'core/core' ).getEntityRecords( 'postType', 'wp_block' )",
-		{
-			since: '6.5',
-			version: '6.8',
-		}
-	);
+	deprecated( `select( 'core/edit-site' ).getReusableBlocks()`, {
+		since: '6.5',
+		version: '6.8',
+		alternative: `select( 'core/core' ).getEntityRecords( 'postType', 'wp_block' )`,
+	} );
 	const isWeb = Platform.OS === 'web';
 	return isWeb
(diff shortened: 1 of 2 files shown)


packages/edit-site/src/store/selectors.js
```diff
@@ -71,7 +71,17 @@ export const __experimentalGetPreviewDeviceType = createRegistrySelector(
  * @return {Object} Whether the current user can create media or not.
  */
 export const getCanUserCreateMedia = createRegistrySelector(
-	( select ) => () => select( coreDataStore ).canUser( 'create', 'media' )
+	( select ) => () => {
+		deprecated(
+			`wp.data.select( 'core/edit-site' ).getCanUserCreateMedia()`,
+			{
+				since: '6.7',
+				alternative: `wp.data.select( 'core' ).canUser()`,
```
ellatrix: Should we be more specific here?

packages/edit-site/src/store/test/selectors.js
```diff
@@ -1,34 +1,9 @@
...
-		select: jest.fn( () => ( { canUser } ) ),
-	};
-
-	describe( 'getCanUserCreateMedia', () => {
-		it( "selects `canUser( 'create', 'media' )` from the core store", () => {
-			expect( getCanUserCreateMedia() ).toBe( true );
-			expect(
-				getCanUserCreateMedia.registry.select
-			).toHaveBeenCalledWith( coreDataStore );
-			expect( canUser ).toHaveBeenCalledWith( 'create', 'media' );
```
ellatrix, line 28: I guess alternatively we could expect the deprecation message.

ellatrix (approved):

packages/edit-site/src/store/selectors.js
Mamaduka: We have mixed usage of alternative suggestions. So not sure 😅

packages/edit-site/src/store/test/selectors.js
Mamaduka: Do you find similar tests useful?

packages/edit-site/src/store/selectors.js
ellatrix: What do you mean? Here the only alternative is `canUser( 'create', 'media' )`, no?

packages/edit-site/src/store/test/selectors.js
ellatrix: Nope, I personally don't like these pointless unit tests, just pointing it out :D

packages/edit-site/src/store/selectors.js
Mamaduka: Got it. Sure, I'll update the suggestion.

PR #43480, opened 2022-08-22
Core Data: Stop sending duplicate requests in canUser resolver

adamziel:
## What?

The following snippet issues three identical HTTP requests:

```js
canUser( 'pages', 'update', 10 );
canUser( 'pages', 'read', 10 );
canUser( 'pages', 'delete', 10 );
```

Each time, the `canUser` resolver sends a simple OPTIONS request to the `/wp/v2/pages/10`.

This PR caches and reuses the results retrieved the first time. It follows up on an [observation @Mamaduka got after stabilizing the `useResourcePermissions` hook](https://github.com/WordPress/gutenberg/pull/43268#issuecomment-1221950456):

> When using this hook, even if you only check for single permission, it will still make multiple calls to the same endpoint.

## Testing Instructions

It's hard to test this in a browser so I added a test case to check how many API calls are issued. Confirm it makes sense and passes.

cc @Mamaduka @gziolo

PR #63415, opened 2024-07-11
Use entity details when calling 'canUser' selectors

Mamaduka:
## What?

PR updates the codebase to use new entity-are syntax for the `canUser` selector.

## How?
I just switched the selector calls to the new syntax and tried to make as few changes as possible to the logic.

## Testing Instructions
I hope e2e tests have our back for these changes, but smoke testing the editor functionality will help.

Same.

### DIFF
--- a/packages/block-library/src/template-part/edit/inner-blocks.js
+++ b/packages/block-library/src/template-part/edit/inner-blocks.js
@@ -127,11 +127,14 @@ export default function TemplatePartInnerBlocks( {
 	const { canViewTemplatePart, canEditTemplatePart } = useSelect(
 		( select ) => {
 			return {
-				canViewTemplatePart:
-					select( coreStore ).canUser( 'read', 'templates' ) ?? false,
-				canEditTemplatePart:
-					select( coreStore ).canUser( 'create', 'templates' ) ??
-					false,
+				canViewTemplatePart: !! select( coreStore ).canUser( 'read', {
+					kind: 'postType',
+					name: 'wp_template',
+				} ),
+				canEditTemplatePart: !! select( coreStore ).canUser( 'create', {
+					kind: 'postType',
+					name: 'wp_template',
+				} ),
 			};
 		},
 		[]
--- a/packages/edit-site/src/components/add-new-pattern/index.js
+++ b/packages/edit-site/src/components/add-new-pattern/index.js
@@ -56,8 +56,14 @@ export default function AddNewPattern() {
 			addNewTemplatePartLabel: getPostType( TEMPLATE_PART_POST_TYPE )
 				?.labels?.add_new_item,
 			// Blocks refers to the wp_block post type, this checks the ability to create a post of that type.
-			canCreatePattern: canUser( 'create', 'blocks' ),
-			canCreateTemplatePart: canUser( 'create', 'template-parts' ),
+			canCreatePattern: canUser( 'create', {
+				kind: 'postType',
+				name: PATTERN_TYPES.user,
+			} ),
+			canCreateTemplatePart: canUser( 'create', {
+				kind: 'postType',
+				name: TEMPLATE_PART_POST_TYPE,
+			} ),
 		};
 	}, [] );
 
--- a/packages/edit-site/src/components/global-styles/font-library-modal/installed-fonts.js
+++ b/packages/edit-site/src/components/global-styles/font-library-modal/installed-fonts.js
@@ -93,7 +93,11 @@ function InstalledFonts() {
 			const { canUser } = select( coreStore );
 			return (
 				customFontFamilyId &&
-				canUser( 'delete', 'font-families', customFontFamilyId )
+				canUser( 'delete', {
+					kind: 'postType',
+					name: 'wp_font_family',
+					id: customFontFamilyId,
+				} )
 			);
 		},
 		[ customFontFamilyId ]
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -372,15 +372,14 @@ export default function PostList( { postType } ) {
 			const { getEntityRecord, getPostType, canUser } =
 				select( coreStore );
 			const siteSettings = getEntityRecord( 'root', 'site' );
-			const postTypeObject = getPostType( postType );
 			return {
 				frontPageId: siteSettings?.page_on_front,
 				postsPageId: siteSettings?.page_for_posts,
 				labels: getPostType( postType )?.labels,
-				canCreateRecord: canUser(
-					'create',
-					postTypeObject?.rest_base || 'posts'
-				),
+				canCreateRecord: canUser( 'create', {
+					kind: 'postType',
+					name: postType,
+				} ),
 			};
 		},
 		[ postType ]
--- a/packages/editor/src/components/post-template/block-theme.js
+++ b/packages/editor/src/components/post-template/block-theme.js
@@ -53,7 +53,11 @@ export default function BlockThemeControl( { id } ) {
 
 	const canCreateTemplate = useSelect(
 		( select ) =>
-			select( coreStore ).canUser( 'create', 'templates' ) ?? false
+			select( coreStore ).canUser( 'create', {
+				kind: 'postType',
+				name: 'wp_template',
+			} ),
+		[]
 	);
 
 	if ( ! hasResolved ) {
--- a/packages/editor/src/components/post-template/block-theme.js
+++ b/packages/editor/src/components/post-template/block-theme.js
@@ -53,7 +53,7 @@ export default function BlockThemeControl( { id } ) {
 
 	const canCreateTemplate = useSelect(
 		( select ) =>
-			select( coreStore ).canUser( 'create', {
+			!! select( coreStore ).canUser( 'create', {
 				kind: 'postType',
 				name: 'wp_template',
 			} ),
--- a/packages/editor/src/components/provider/use-block-editor-settings.js
+++ b/packages/editor/src/components/provider/use-block-editor-settings.js
@@ -133,7 +133,10 @@ function useBlockEditorSettings( settings, postType, postId, renderingMode ) {
 			const { getBlockTypes } = select( blocksStore );
 			const { getBlocksByName, getBlockAttributes } =
 				select( blockEditorStore );
-			const siteSettings = canUser( 'read', 'settings' )
+			const siteSettings = canUser( 'read', {
+				kind: 'root',
+				name: 'site',
+			} )
 				? getEntityRecord( 'root', 'site' )
 				: undefined;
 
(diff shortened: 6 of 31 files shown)


packages/edit-site/src/components/add-new-pattern/index.js
```diff
@@ -56,8 +56,14 @@ export default function AddNewPattern() {
...
 			// Blocks refers to the wp_block post type, this checks the ability to create a post of that type.
-			canCreatePattern: canUser( 'create', 'blocks' ),
-			canCreateTemplatePart: canUser( 'create', 'template-parts' ),
+			canCreatePattern: canUser( 'create', {
+				kind: 'postType',
+				name: 'wp_block',
+			} ),
+			canCreateTemplatePart: canUser( 'create', {
+				kind: 'postType',
+				name: 'wp_template_part',
```
ellatrix: Could we use `TEMPLATE_PART_POST_TYPE` here, and in similar in other places?

packages/patterns/src/components/pattern-convert-button.js
```diff
@@ -86,7 +86,10 @@ export default function PatternConvertButton( {
 				) &&
 				// Hide when current doesn't have permission to do that.
 				// Blocks refers to the wp_block post type, this checks the ability to create a post of that type.
-				!! canUser( 'create', 'blocks' );
+				!! canUser( 'create', {
+					kind: 'postType',
+					name: 'wp_block',
+				} );
```
ellatrix, line 92: Tbh, I wish this API looks looked more similar to `getEntityRecord`, or the other way around. Then you could maybe also reuse the same args for fetching and checking permissions.

packages/edit-site/src/components/add-new-pattern/index.js
Mamaduka: Update to use constants where it made sense - where the constant was available.

Mamaduka:
This is ready for the review. I'll handle the `useResourcePermissions` hook separately.

cc @WordPress/gutenberg-core

packages/patterns/src/components/pattern-convert-button.js
gziolo: I think under the hood on the REST API level it's exactly what happens as it sometimes has to call OPTIONS request so sharing the same list of params in that case was warranted when using `canUserEditEntityRecord`.

ellatrix: Yeah, could we still change it, since `canUser` has only just been modified?

ellatrix: Or should `getEntityRecord(s)` allow (clearer) named parameters? I guess it would be good to have more opinions here. Cc @jsnajdr @youknowriad

Mamaduka: If we use similar arguments, it would be hard to distinguish whether the user wants an entity-related or REST API resource query.

How should we handle cases like this without collision? While IDs are usually an `int`, this isn't a requirement.
```js
canUser( 'create', 'media' );
canUser( 'create', 'root', 'media' );
```

youknowriad: Yeah, I think using "resource" in the first place was a wrong idea but it seems the ship has sailed here.

packages/block-library/src/template-part/edit/inner-blocks.js
```diff
@@ -127,11 +127,14 @@ export default function TemplatePartInnerBlocks( {
...
 		( select ) => {
 			return {
-				canViewTemplatePart:
-					select( coreStore ).canUser( 'read', 'templates' ) ?? false,
-				canEditTemplatePart:
-					select( coreStore ).canUser( 'create', 'templates' ) ??
-					false,
+				canViewTemplatePart: !! select( coreStore ).canUser( 'read', {
+					kind: 'postType',
+					name: 'wp_template',
```
tyxla, line 132: Similar to what @ellatrix suggested about the template part post type, we could use `TEMPLATE_POST_TYPE` for these.

packages/editor/src/components/provider/use-block-editor-settings.js
```diff
@@ -169,8 +172,15 @@ function useBlockEditorSettings( settings, postType, postId, renderingMode ) {
 				hiddenBlockTypes: get( 'core', 'hiddenBlockTypes' ),
 				isDistractionFree: get( 'core', 'distractionFree' ),
 				keepCaretInsideBlock: get( 'core', 'keepCaretInsideBlock' ),
-				hasUploadPermissions: canUser( 'create', 'media' ) ?? true,
-				userCanCreatePages: canUser( 'create', 'pages' ),
+				hasUploadPermissions:
+					canUser( 'create', {
+						kind: 'root',
+						name: 'media',
```
tyxla, line 176: I've always found it odd that the media entity is a separate root entity, and we don't use the `attachment` post type. Somehow, this special case is more visible now, which I find potentially confusing. Not much to do obviously, since this is about inherited complexity.

packages/edit-site/src/components/global-styles/font-library-modal/installed-fonts.js
```diff
@@ -93,7 +93,11 @@ function InstalledFonts() {
 			const { canUser } = select( coreStore );
 			return (
 				customFontFamilyId &&
-				canUser( 'delete', 'font-families', customFontFamilyId )
+				canUser( 'delete', {
+					kind: 'postType',
+					name: 'wp_font_family',
+					id: customFontFamilyId,
```
tyxla, line 99: I appreciate the explicit syntax, it makes it clearer as to what we're doing. Previously, one had to search what `font-families` is - is it a taxonomy, is it a post type?

packages/patterns/src/components/pattern-convert-button.js
gziolo: Backward compatibility driven development is a factor here 😄

Mamaduka: Btw, the `block-directory` package has a valid use case for "resource".

https://github.com/WordPress/gutenberg/blob/1b7e3a7502693071e8877cc12f92a51f5a13212a/packages/block-directory/src/components/downloadable-blocks-panel/index.js#L29-L32

packages/block-library/src/template-part/edit/inner-blocks.js
Mamaduka: I've updated the part when these constants were available. They're package-specific. Most entity-related blocks hardcode their entity names.

Are we worried that these names might change?

Also, I don't mind defining a couple of constants at the file level, but for the initial run, I tried to change only the selector syntax and leave the rest of the code alone.

tyxla: Definitely not a blocker for the PR 👍

packages/editor/src/components/provider/use-block-editor-settings.js
Mamaduka: 100%, but WP always treated Media as a separate entity, even though it's just a post type.

P.S. Jorge discovered another odd case recently: https://github.com/WordPress/gutenberg/issues/63438.

Mamaduka:
Is this ready to be merged, and can I get approval? 🙇

packages/block-library/src/template-part/edit/inner-blocks.js
```diff
@@ -127,11 +127,14 @@ export default function TemplatePartInnerBlocks( {
...
+				} ),
+				canEditTemplatePart: !! select( coreStore ).canUser( 'create', {
```
tyxla, line 134: Not from this PR, but it felt odd that we call this "can edit" but check for whether they can create (`POST`) and not `update` (PUT). Haven't checked if that makes more sense in the context it's used.

packages/edit-site/src/components/post-list/index.js
```diff
@@ -372,15 +372,14 @@ export default function PostList( { postType } ) {
...
 				select( coreStore );
 			const siteSettings = getEntityRecord( 'root', 'site' );
-			const postTypeObject = getPostType( postType );
 			return {
 				frontPageId: siteSettings?.page_on_front,
 				postsPageId: siteSettings?.page_for_posts,
 				labels: getPostType( postType )?.labels,
-				canCreateRecord: canUser(
-					'create',
-					postTypeObject?.rest_base || 'posts'
```
tyxla, line 382: Nice simplification 👍 

I wonder why the pre-existing indirection and the default `'posts`` was necessary 🤔

packages/editor/src/components/post-template/block-theme.js
```diff
@@ -53,7 +53,11 @@ export default function BlockThemeControl( { id } ) {
 
 	const canCreateTemplate = useSelect(
 		( select ) =>
-			select( coreStore ).canUser( 'create', 'templates' ) ?? false
+			select( coreStore ).canUser( 'create', {
+				kind: 'postType',
+				name: 'wp_template',
+			} ),
```
tyxla, line 59: Should we keep the default: `?? false`? Theoretically `canUser` can still return `undefined`, right?

packages/editor/src/components/visual-editor/edit-template-blocks-notification.js
```diff
@@ -40,7 +40,11 @@ export default function EditTemplateBlocksNotification( { contentRef } ) {
 
 	const canEditTemplate = useSelect(
 		( select ) =>
-			select( coreStore ).canUser( 'create', 'templates' ) ?? false
+			!! select( coreStore ).canUser( 'create', {
```
tyxla, line 43: Explicit conversion to boolean works well as a substitute to `?? false`.

tyxla (approved):
Smoke tested both post and site editors and everything works well.

Code looks good to me 👍 

Left a few minor, non-blocking questions.

Thanks @Mamaduka 🚀

packages/block-library/src/template-part/edit/inner-blocks.js
Mamaduka: I think introducing the pattern might have been my fault. The WP maps all template [CRUD permissions to the same capability](https://github.com/WordPress/wordpress-develop/blob/5b0006ca9b5408aa73b6dc71fb5b3ebf85f497ec/src/wp-includes/post.php#L387-L398), so I reused already existing call for edit capability check to avoid making extra `OPTIONS` request to the server.

The duplicate requests issue was fixed (#43480). I forgot to update my old code, and the pattern spread into new places.

I will follow up on this in a separate PR.

packages/edit-site/src/components/post-list/index.js
Mamaduka: I'm guessing to avoid passing `undefined` when `postTypeObject` hasn't been resolved yet, which would result in sending a request to the general namespace. However, I'm unsure why the `posts` fallback was selected.

I fixed a similar issue in https://github.com/WordPress/gutenberg/pull/63423#discussion_r1673836639.

packages/editor/src/components/post-template/block-theme.js
Mamaduka: Resolved in [e683bab](https://github.com/WordPress/gutenberg/pull/63415/commits/e683babcd050b2efa244fa8493b6759da0389a7f).

packages/edit-site/src/components/post-list/index.js
tyxla: Back in the day, when we were writing up custom admin menus, for example, and we were displaying labels for custom post types, we were falling back to the default labels for post types - because when registering a custom post type, not all labels are guaranteed to be declared. Maybe this assumed that the `rest_base` could not be declared, too.

Mamaduka: The `rest_base` should default to the [post type slug](https://developer.wordpress.org/reference/functions/register_post_type/#:~:text=Default%20is%20%24post_type.).

tyxla: True. But then it can be overridden 🤷‍♂️

Mamaduka: You can do almost everything in WP 😅

tyxla: Yup. I'm just trying to understand why this was here. Maybe it was a temporary thing as we were working to make the posts list more flexible in the future? It's a recent addition (#62705), perhaps @ntsekouras could remember.

ntsekouras: Yeah, the default was to avoid the `undefined` value. Regarding why we need the resource, it's because we are using the same component for `posts and pages` and will probably be the base component of other post types too. Finally regarding the `default:posts`, didn't think it mattered much :).

Aside this, I'm catching up with @Mamaduka's work on these parts and is great! Thank you!

PR #63419, opened 2024-07-11
Latest comments: Add color block support

carolinan:
## What?

Adds text, background, gradient, and link color support to the Latest comments block.
Background, text and link are set to show by default.

Partial for https://github.com/WordPress/gutenberg/issues/43245

## Why?

## How?

The PR updates the block supports in block.json, and the documentation.

## Testing Instructions
Create a new post and add a latest comments block. Adjust the color settings and compare the result in the editor and front.


carolinan:
Edit: This issue was solved in a separate PR.

## Block themes

### Editor
The block inherits the padding from the OL:

```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```
The left padding is then removed, using this CSS in the block stylesheet.
```
.wp-block-latest-comments .wp-block-latest-comments {
    padding-left: 0;
}
```

### Single post, front view
The block does not inherit the padding from the OL element,
and the left padding is not removed, so it uses the padding from the
default browser styles for OL.

### Home/frontpage, front view
The block inherits the padding from
```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```
But the left padding is not removed.


## Classic themes
### Editor
The block inherits the padding from the OL:

```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```
The left padding is then removed, using this CSS in the block stylesheet.
```
.wp-block-latest-comments .wp-block-latest-comments {
    padding-left: 0;
}
```
### Front
Inherits this padding, but the left padding is not removed.
```
:root :where(ul.has-background,ol.has-background) {
    padding: 1.25em 2.375em;
}
```

ndiego (approved):
Works as expected. Nice enhancement.

image

Issue #51247, opened 2023-06-05
Details block: Improve ability to drag & drop content into hidden text


annezazu:
This came up as part of the [FSE Outreach Program's Rapid Revamp call for testing](https://make.wordpress.org/test/2023/05/18/fse-program-testing-call-23-rapid-revamp/#comment-3069):

>With the Details Block I felt very frustrated. In a first moment, I tried to drag and drop to move the selected content to “Type / to add a hidden block”, but it’s impossible why the Details block is closed when lost the focus. I had to use the keyboard (CMD + X) to cut the content and paste it. Additionally, I tried to click in the arrow to close the block (by intuition), but don’t occur anything. Only close when lost the focus and open when get the focus. In my programmer might, this block has an event onClick in the arrow (same that in the front), so perhaps it isn’t possible 

>I struggle for a while with block movement in general. I insert my block lower than I want, then selecting multiple paragraphs almost does me in. When I can’t keep details open to drag-and-drop, I resort to list view.

Here's a video showing how painful it is to drag and drop into: 

https://github.com/WordPress/gutenberg/assets/26996883/d11249fa-6f11-424b-abd2-f478b2ba67c9

In my mind, we need a way to visualize dropping into a nested container and/or have it open up, similar to what's been done with the List View and drag and drop https://github.com/WordPress/gutenberg/issues/49563

annezazu:
This is still a major problem that I ran into today with 6.7! After messing around for long enough, I figured out how to get it to work as the drag and drop doesn't quite make it clear that it's been added to the right place. You have to drag and drop it right on top of the Details block or else it won't be added:


https://github.com/user-attachments/assets/7f1d8337-4ff2-4ace-8063-ec7a62e4b5f0

PR #63422, opened 2024-07-11
Details block: add transform from any block type

### LINKED ISSUE #55602: Details block: Add transform from paragraph
I expected to be able to transform a paragraph into the inner contents of a details block. That way I wouldn’t have to copy and paste, then delete empty paragraph blocks. 
￼
CleanShot 2023-10-25 at 10 21 52

ellatrix:
## What?

Ran into the problem of not being able to select blocks and moving them into a details blocks. This PR adds a transform for that. Fixes #55602. See also #51247.

## Why?

It's not currently possible to move content into a details block.

## Testing Instructions

Create some block, then select them and click the block icon to transform them into the details block.
Also singular blocks can be transformed.

### DIFF
--- a/packages/block-library/src/details/index.js
+++ b/packages/block-library/src/details/index.js
@@ -11,6 +11,7 @@ import initBlock from '../utils/init-block';
 import metadata from './block.json';
 import edit from './edit';
 import save from './save';
+import transforms from './transforms';
 
 const { name } = metadata;
 export { metadata, name };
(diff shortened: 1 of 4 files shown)


richtabor (approved):
Works as expected, nice!

https://github.com/WordPress/gutenberg/assets/1813435/d0538a35-48a6-4bbf-a98f-1a84e4a94371

richtabor:
One tweak would be to not have the transform on the details block itself.

CleanShot 2024-07-11 at 11 10 10

PR #63423, opened 2024-07-11
Post Actions: Use entity details for capability checks

Mamaduka:
## What?

PR updates user capability checks for post actions to use the updated `canUser` selector.

Note: I decided to do this separately from #63415 since `canUser` checks are somewhat convoluted here. 

## Why?
The code should always use the new method when working with entity and user capabilities.

## Testing Instructions
Confirm that post actions work as before in post and site editors.

Same.

## Screenshots or screencast 
CleanShot 2024-07-11 at 15 05 33

### DIFF
--- a/packages/editor/src/components/post-actions/actions.js
+++ b/packages/editor/src/components/post-actions/actions.js
@@ -230,7 +230,7 @@ const trashPostAction = {
 	},
 };
 
-function useCanUserEligibilityCheckPostType( capability, resource, action ) {
+function useCanUserEligibilityCheckPostType( capability, postType, action ) {
 	const registry = useRegistry();
 	return useMemo(
 		() => ( {
@@ -238,20 +238,22 @@ function useCanUserEligibilityCheckPostType( capability, resource, action ) {
 			isEligible( item ) {
 				return (
 					action.isEligible( item ) &&
-					registry
-						.select( coreStore )
-						.canUser( capability, resource, item.id )
+					registry.select( coreStore ).canUser( capability, {
+						kind: 'postType',
+						name: postType,
+						id: item.id,
+					} )
 				);
 			},
 		} ),
-		[ action, registry, capability, resource ]
+		[ action, registry, capability, postType ]
 	);
 }
 
-function useTrashPostAction( resource ) {
+function useTrashPostAction( postType ) {
 	return useCanUserEligibilityCheckPostType(
 		'delete',
-		resource,
+		postType,
 		trashPostAction
 	);
 }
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -623,22 +625,21 @@ const renamePostAction = {
...
 
 const useDuplicatePostAction = ( postType ) => {
-	const { userCanCreatePost } = useSelect(
+	const userCanCreatePost = useSelect(
 		( select ) => {
-			const { getPostType, canUser } = select( coreStore );
-			const resource = getPostType( postType )?.rest_base || '';
-			return {
-				userCanCreatePost: canUser( 'create', resource ),
-			};
```
Mamaduka, line 641: Prev code triggered a request to the `wp/v2` namespace.

## Screenshot
CleanShot 2024-07-11 at 15 00 21

ellatrix: So this is a 6.6 bug?

Mamaduka: Yes, but it shouldn't affect users. It just makes an extra request.

First, it will call `canUser( 'create', '' )`, and then when the post type object is resolved, it will call it correctly.

ellatrix: Ah 👍

jorgefilipecosta: Duplicate action is also not included in the core, it is in the plugin only for now, so it should not have an impact on 6.6.

jorgefilipecosta (approved):
Works well and the changes look good, thank you for these updates 👍

Mamaduka:
Thank you, @jorgefilipecosta!

packages/editor/src/components/post-actions/actions.js
```diff
@@ -230,28 +230,30 @@ const trashPostAction = {
...
 	return useMemo(
 		() => ( {
 			...action,
 			isEligible( item ) {
 				return (
 					action.isEligible( item ) &&
-					registry
-						.select( coreStore )
-						.canUser( capability, resource, item.id )
+					registry.select( coreStore ).canUser( capability, {
```
youknowriad, line 241: There's something I don't understand here. how does this value change when it gets resolved. There's nothing triggering a re-render when the value of this selector changes and `isEligible` is not an event callback it's something that is called when rendering.

I'm sure we have a hack elsewhere that "hides" the bug, but this is not great.

Mamaduka: The original PR (#62589) might have more details, but I agree that this feels super "hacky."

PR #63424, opened 2024-07-11
DataViews: Fix featured image height regression

youknowriad:
Fix a small regression that was introduced in https://github.com/WordPress/gutenberg/pull/63334

You can check by opening the "pages" in the site editor and check that the featured images appear properly in both list and grid views.


jameskoster (approved):
Working for me :)

PR #63426, opened 2024-07-11
[Mobile] - Gallery block - Fix getMediaItems

geriux:
**Related PRs:**
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/6978
- https://github.com/wordpress-mobile/WordPress-Android/pull/21053
- https://github.com/wordpress-mobile/WordPress-iOS/pull/23417

## What?
This PR addresses a regression in the Gallery block that was causing error notices to appear in certain scenarios. Additionally, it prevents the images from randomly changing their position when new images are added.

## Why?
To provide a good user experience when adding new images by preventing random errors and unexpected behavior.

## How?
I didn’t find the culprit of this regression, but after some investigation, I confirmed that the data was being fetched correctly, yet it wasn’t being stored in the state.

This was causing the [useGetMedia](https://github.com/WordPress/gutenberg/blob/a4e042f1e2bb628ffcab3bb8dedb12cc61974f0d/packages/block-library/src/gallery/use-get-media.native.js) hook to not return the expected data of the images.

By looking more into the code, I realized the fetch promise was being intercepted [by a catch](https://github.com/WordPress/gutenberg/blob/a4e042f1e2bb628ffcab3bb8dedb12cc61974f0d/packages/core-data/src/resolvers.js#L298-L300), so no data was being stored.

The error was happening in [fetchAllMiddleware](https://github.com/WordPress/gutenberg/blob/a4e042f1e2bb628ffcab3bb8dedb12cc61974f0d/packages/api-fetch/src/middlewares/fetch-all-middleware.js#L30-L31), where [parseResponse](https://github.com/WordPress/gutenberg/blob/a4e042f1e2bb628ffcab3bb8dedb12cc61974f0d/packages/api-fetch/src/middlewares/fetch-all-middleware.js#L30-L31) is called and expects to have a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) parameter in JSON format.

After all this, I decided to update the native implementation of [fetchHandler](https://github.com/WordPress/gutenberg/blob/a4e042f1e2bb628ffcab3bb8dedb12cc61974f0d/packages/react-native-editor/src/api-fetch-setup.js) and update how it parses the response.

Now it will return a `Response` object if the `parse` parameter is set to `false`.

Steps provided from [another PR review](https://github.com/wordpress-mobile/gutenberg-mobile/pull/6973#pullrequestreview-2167369822):

1. Insert a Gallery block.
2. Select media for the block.
3. Move block selection to the Gallery block.
4. Tap the inline appender.
5. Add media already within the Gallery block.
6. Note the error message and lack of new image.
7. Move block selection to an Image block within the Gallery block.
8. Tap the block inserter in the toolbar.
9. Add media already within the Gallery block.
10. Note the image is added.
11. Move block selection to the Gallery block.
12. Tap the inline appender.
13. Add media already within the Gallery block.
14. Note the error message, the lack of new image, and the existing images order is reshuffled.

No testing is needed.

## Screenshots or screencast 
Before|After
-|-
<video src="https://github.com/WordPress/gutenberg/assets/4885740/319fddfe-4d3f-46e5-b468-efd94991052b" width=250 />|<video src="https://github.com/WordPress/gutenberg/assets/4885740/576b49a1-2189-4979-ba09-640afaac4dee" width=250 />

### DIFF
--- a/packages/react-native-editor/src/api-fetch-setup.js
+++ b/packages/react-native-editor/src/api-fetch-setup.js
@@ -57,19 +57,23 @@ const fetchHandler = (
 	}
 
 	const parseResponse = ( response ) => {
+		const isStringResponse = typeof response === 'string';
+
 		// If the 'parse' parameter is false, return the response as a new Response object
 		// with the JSON string. This is necessary because one of the middlewares in the API
 		// fetch library expects the response to be a JavaScript Response object with JSON
 		// functionality. By doing this, we ensure that the response is handled correctly
 		// by fetch-all-middleware.
 		if ( parse === false ) {
-			const jsonData = JSON.stringify( response );
-			return new Response( jsonData, {
+			const body = isStringResponse
+				? response
+				: JSON.stringify( response );
+			return new Response( body, {
 				headers: { 'Content-Type': 'application/json' },
 			} );
 		}
 
-		if ( typeof response === 'string' ) {
+		if ( isStringResponse ) {
 			response = JSON.parse( response );
 		}
 		return response;
--- a/packages/react-native-editor/src/api-fetch-setup.js
+++ b/packages/react-native-editor/src/api-fetch-setup.js
@@ -57,10 +57,6 @@ const fetchHandler = (
 	}
 
 	const parseResponse = ( response ) => {
-		if ( typeof response === 'string' ) {
-			return ( response = JSON.parse( response ) );
-		}
-
 		// If the 'parse' parameter is false, return the response as a new Response object
 		// with the JSON string. This is necessary because one of the middlewares in the API
 		// fetch library expects the response to be a JavaScript Response object with JSON
@@ -71,6 +67,10 @@ const fetchHandler = (
 			return new Response( jsonData, {
 				headers: { 'Content-Type': 'application/json' },
 			} );
+		}
+
+		if ( typeof response === 'string' ) {
+			response = JSON.parse( response );
 		}
 		return response;
 	};
--- a/packages/react-native-editor/src/api-fetch-setup.js
+++ b/packages/react-native-editor/src/api-fetch-setup.js
@@ -32,7 +32,7 @@ const setTimeoutPromise = ( delay ) =>
 	new Promise( ( resolve ) => setTimeout( resolve, delay ) );
 
 const fetchHandler = (
-	{ path, method = 'GET', data },
+	{ path, method = 'GET', data, parse },
 	retries = 20,
 	retryCount = 1
 ) => {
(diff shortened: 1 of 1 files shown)


dcalhoun:
> Additionally, it prevents the images from randomly changing their position when new images are added.

@geriux The error message is now gone, but I still experience the inability to add the same image twice and the occasional unexpected position changes when following the steps shared in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6973#pullrequestreview-2167369822. This occurs in both the iOS and Android builds. Do you experience this as well?

geriux:
Hey @dcalhoun thanks for testing this change!

> Do you experience this as well?

Yes, I just reproduced it. I was always adding photos from my device and not using the WordPress Media Library. 😅

So, this is also happening in the web editor. The only difference is that in the native WordPress media library, we are not highlighting the selected media, so you are able to select it, but then the code filters it out. In the web UI, you see already added images as selected:

<kbd>Screenshot 2024-07-12 at 12 06 27</kbd>

I believe we had this functionality before the recent media library updates 🤔 

As for the images changing their position, this issue is also occurring in the web editor. This happens by adding an image from the Image block appender, where you are able to insert an image that’s already added to the Gallery. When a duplicated image is added, they will change their position to be next to each other because there's a [sorting part](https://github.com/WordPress/gutenberg/blob/52b9831673e7afcebc25768e5cf98dc63350e20e/packages/block-library/src/gallery/edit.js#L318-L327) in the code.

We should improve the WordPress Media Library in the host apps to avoid adding duplicate images by highlighting the ones that are already added.

dcalhoun:
> So, this is also happening in the web editor. The only difference is that in the native WordPress media library, we are not highlighting the selected media, so you are able to select it, but then the code filters it out. 

Ah, that makes sense. Thank you for investigating. 

> As for the images changing their position, this issue is also occurring in the web editor. 

I am unable to reproduce the unexpected gallery item re-ordering in the web editor. However, I suppose the issue within the native editor might only be significantly disruptive when modifying a large gallery. The user can recover from this bug by manually re-ordering the gallery items after uploading.

geriux:
> I am unable to reproduce the unexpected gallery item re-ordering in the web editor. However, I suppose the issue within the native editor might only be significantly disruptive when modifying a large gallery. The user can recover from this bug by manually re-ordering the gallery items after uploading.

Yeah, I agree they could just move the position of the image.

This is how I was able to reproduce it:

https://github.com/user-attachments/assets/71b2efc3-d112-4155-af10-248f35cf3bcd

dcalhoun (approved):
> I didn’t find the culprit of this regression, but after some investigation, I confirmed that the data was being fetched correctly, yet it wasn’t being stored in the state.

Very interesting. Thank you for investigating and sharing your findings. 

The test plan succeeded for me on an iPhone SE and Galaxy S20. 🚀

PR #63427, opened 2024-07-11
DataViews: Fix default layouts in the pages data views

youknowriad:
Fix small regression introduced by a combination of #63236 and #63287  

## What?

This was a small leftover that was due partially to both the linked PRs above. 
Basically, in trunk now, pages data views expand the "title" column width if there are pages with long titles.
This PR restores the "max width" config properly.

## Testing Instructions

- Have a page with a very long title
- Open the "table" dataviews
- The page title should be trimmed


jameskoster (approved):
Thank you!

PR #63428, opened 2024-07-11
Gallery: Add border block support

carolinan:
## What?

Add border block support to the gallery block.

Partial for https://github.com/WordPress/gutenberg/issues/43247

## Why?

## How?

## Testing Instructions
Create a new post and add a gallery block. Adjust the border settings and compare the result in the editor and front.

## Screenshots or screencast 
Different border settings in combination with background and caption. The last gallery has extra padding. Using Twenty Twenty-Four:
gallery-border

### DIFF
--- a/packages/block-library/src/gallery/block.json
+++ b/packages/block-library/src/gallery/block.json
@@ -112,6 +112,16 @@
 	"supports": {
 		"anchor": true,
 		"align": true,
+		"__experimentalBorder": {
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"color": true,
+				"radius": true
+			}
+		},
 		"html": false,
 		"units": [ "px", "em", "rem", "vh", "vw" ],
 		"spacing": {


carolinan:
In this PR I have included border radius for testing. My recommendation would be to not support border radius, because no matter if the border crops the images or not, the result may be unexpected for different users.

aaronrobertshaw (approved):
Thanks for putting this PR together @carolinan. I really appreciate the effort to move block support adoption and consistency forward 🙇 

This is working as expected for me.

✅ Global Styles for gallery borders apply correctly in the editor and frontend
✅ Block instance border styles override the global gallery border styles correctly

> My recommendation would be to not support border radius, because no matter if the border crops the images or not, the result may be unexpected for different users.

I'm actually leaning towards adopting the border radius here. The gallery block supports padding so users could find a combination that works for them.

I know the Column block doesn't support border radius, like its parent Columns, and that has created some long running discussion that it should. 

The issue there is that the quick inserter gets cropped in the Column block if the border radius is too high. With the gallery, new images can be added both via hitting enter or visually using the "add" button in the block toolbar, so we avoid that specific issue.

I'd be interested to hear others' thoughts on this one. In the meantime though, I'll give it a tentative ✅ .



https://github.com/user-attachments/assets/11b959e8-d4b4-4c5e-b3c0-6778d4aac542


cc/ @andrewserong any thoughts on border radius for galleries?

andrewserong:
> cc/ @andrewserong any thoughts on border radius for galleries?

Thanks for the ping!

> I'm actually leaning towards adopting the border radius here. The gallery block supports padding so users could find a combination that works for them.

Same here. I think it's a similar state as to the Group block? I see that if you give a Grid a border, it doesn't clip images either, and users (or theme authors) can set the desired padding to account for it, as you mention. E.g. here's a quick example of a Grid block with images in it with some border radius:

image

So I'd lean toward allowing border radius to be present, but I might just ping @WordPress/gutenberg-design for visibility in case anyone there has strong opinions.

paaljoachim:
I too would find it helpful to also have the border radius support added to the Gallery block. I would prefer to have consistency between what the various blocks offer so we get used to having the same features present.

carolinan:
Then lets merge this :)

jasmussen:
There's a long running issue on allowing you to handle overflow. That's inherently pretty complex, and until we decide how/whether to add such overflow handling, I'd tend to lean towards us at least syncing up what container blocks do, and then do the same here. In this case, the Group block allows border and radius, and if you insert several images, it'll land in the same uncropped state as this. To that end, although the radius doesn't seem that useful on the Gallery container, I'd tend to allow it, because if you combine with background and radius, at least you can round the corners of the background color.

carolinan:
The good thing is that one does not have to wrap it in yet another group.

andrewserong:
Nice work landing this @carolinan, great to see more blocks getting more block supports! 🎉

PR #43117, opened 2022-08-10
Lodash: Remove from `core-data` resolvers

tyxla:
## What?
This PR removes all Lodash usages from the core editor store resolvers. They all have straightforward native JS replacements nowadays, with the exception of `camelCase`, for which we're using an external library.

## Why?

Lodash is known to unnecessarily inflate the bundle size of packages, and in most cases, it can be replaced with native language functionality. See these for more information and rationale:

* https://github.com/WordPress/gutenberg/issues/16938#issuecomment-602837246
* https://github.com/WordPress/gutenberg/issues/17025
* https://github.com/WordPress/gutenberg/issues/39495 

## How?

We're essentially removing the usage of Lodash functions that are fully replaceable with native JS functionality. The changed code is covered by tests, and I've manually ensured that we provide proper fallbacks where necessary.

## Testing Instructions
* Smoke test the post editor and site editor.
* Verify all tests still pass: `npm run test:unit packages/core-data/src/test/resolvers.js`

PR #39144, opened 2022-03-01
[RNMobile] Prevent unhandled promise rejection warning related to canUser resolver

fluiddot:
## Description

As outlined in https://github.com/WordPress/gutenberg/issues/39143#issuecomment-1055665642, the issue has been addressed by using the optional chaining operator when retrieving the headers from the response.

## Testing Instructions

1. Open any post/page with the app connected to the Metro server.
1. Observe that no warnings related to unhandled promise rejections are displayed.

## Screenshots 
N/A

## Types of changes

Bug fix

## Checklist:
- [ ] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal). 
- [x] I've updated related schemas if appropriate.

PR #63430, opened 2024-07-11
Core Data: Resolve user capabilities when fetching an entity

Mamaduka:
## What?
Inspired by https://github.com/WordPress/gutenberg/pull/43480#issuecomment-1231810231.
Hopefully, this will be more effective after merging #63415 and #63423.

PR updates the `getEntityRecord` to resolve entity-related user permission when fetching entity data.

## Why?
Most of the time, actions to interact with the entity are displayed after loading it. Resolving user capability selectors for the same entity can improve the UI (no more popcorning) and reduce the number of requests the app makes.

### Manually run selectors
1. Open a post or page.
2. Run selectors below in order using DevTools Console.
3. Running the `canUser` selector for the same entity shouldn't trigger a REST API request.

```
// 1. Fetch media details.
wp.data.select( 'core' ).getEntityRecord( 'root', 'media', 1342 );

// 2. Fetch user permission related to the media. It shouldn't trigger a REST API request.
wp.data.select( 'core' ).canUser( 'update', { kind: 'root', name: 'media', id: 1342 } );
```

### Pattern blocks

1. Open a post or page.
2. Add multiple synced pattern blocks.
3. Save and reload the editor.
4. Confirm using Network tools that no `OPTIONS` requests are made for these patterns. You can use `v2/blocks` to filter the endpoint.

Same.

## Screenshots or screencast 
**Requests before with four pattern blocks on canvas:**
CleanShot 2024-07-15 at 16 52 13

**Requests after:**
CleanShot 2024-07-15 at 16 52 36

### DIFF
--- a/packages/block-library/src/block/edit.js
+++ b/packages/block-library/src/block/edit.js
@@ -121,6 +121,51 @@ export default function ReusableBlockEditRecursionWrapper( props ) {
 	);
 }
 
+function ReusableBlockControl( {
+	recordId,
+	canOverrideBlocks,
+	hasContent,
+	handleEditOriginal,
+	resetContent,
+} ) {
+	const canUserEdit = useSelect(
+		( select ) =>
+			!! select( coreStore ).canUser( 'update', {
+				kind: 'postType',
+				name: 'wp_block',
+				id: recordId,
+			} ),
+		[ recordId ]
+	);
+
+	return (
+		<>
+			{ canUserEdit && !! handleEditOriginal && (
+				<BlockControls>
+					<ToolbarGroup>
+						<ToolbarButton onClick={ handleEditOriginal }>
+							{ __( 'Edit original' ) }
+						</ToolbarButton>
+					</ToolbarGroup>
+				</BlockControls>
+			) }
+
+			{ canOverrideBlocks && (
+				<BlockControls>
+					<ToolbarGroup>
+						<ToolbarButton
+							onClick={ resetContent }
+							disabled={ ! hasContent }
+						>
+							{ __( 'Reset' ) }
+						</ToolbarButton>
+					</ToolbarGroup>
+				</BlockControls>
+			) }
+		</>
+	);
+}
+
 function ReusableBlockEdit( {
 	name,
 	attributes: { ref, content },
@@ -143,29 +188,20 @@ function ReusableBlockEdit( {
 
 	const {
 		innerBlocks,
-		userCanEdit,
 		onNavigateToEntityRecord,
 		editingMode,
 		hasPatternOverridesSource,
 	} = useSelect(
 		( select ) => {
-			const { canUser } = select( coreStore );
 			const {
 				getBlocks,
 				getSettings,
 				getBlockEditingMode: _getBlockEditingMode,
 			} = select( blockEditorStore );
 			const { getBlockBindingsSource } = unlock( select( blocksStore ) );
-			const canEdit = canUser( 'update', {
-				kind: 'postType',
-				name: 'wp_block',
-				id: ref,
-			} );
-
 			// For editing link to the site editor if the theme and user permissions support it.
 			return {
 				innerBlocks: getBlocks( patternClientId ),
-				userCanEdit: canEdit,
 				getBlockEditingMode: _getBlockEditingMode,
 				onNavigateToEntityRecord:
 					getSettings().onNavigateToEntityRecord,
@@ -175,7 +211,7 @@ function ReusableBlockEdit( {
 				),
 			};
 		},
-		[ patternClientId, ref ]
+		[ patternClientId ]
 	);
 
 	// Sync the editing mode of the pattern block with the inner blocks.
@@ -256,27 +292,18 @@ function ReusableBlockEdit( {
 
 	return (
 		<>
-			{ userCanEdit && onNavigateToEntityRecord && (
-				<BlockControls>
-					<ToolbarGroup>
-						<ToolbarButton onClick={ handleEditOriginal }>
-							{ __( 'Edit original' ) }
-						</ToolbarButton>
-					</ToolbarGroup>
-				</BlockControls>
-			) }
-
-			{ canOverrideBlocks && (
-				<BlockControls>
-					<ToolbarGroup>
-						<ToolbarButton
-							onClick={ resetContent }
-							disabled={ ! content }
-						>
-							{ __( 'Reset' ) }
-						</ToolbarButton>
-					</ToolbarGroup>
-				</BlockControls>
+			{ hasResolved && (
+				<ReusableBlockControl
+					recordId={ ref }
+					canOverrideBlocks={ canOverrideBlocks }
+					hasContent={ !! content }
+					handleEditOriginal={
+						onNavigateToEntityRecord
+							? handleEditOriginal
+							: undefined
+					}
+					resetContent={ resetContent }
+				/>
 			) }
 
 			{ children === null ? (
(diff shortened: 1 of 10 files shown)


packages/core-data/src/resolvers.js
```diff
@@ -165,8 +165,44 @@ export const getEntityRecord =
...
+						'create',
+						'read',
+						'update',
+						'delete',
+					] ) {
+						const permissionKey = [ action, kind, name, key ]
+							.filter( Boolean )
+							.join( '/' );
+
+						dispatch.receiveUserPermission(
```
youknowriad, line 187: I think that's a good idea but we should also mark the resolvers as resolved to avoid further REST API requests when the selectors get called no?

Mamaduka: Correct, and I think I need to update `canUser` as well.

I see that some resolvers for similar cases call both `startResolution` and `finishResolution` actions, while some only call `finishResolution`. Which method is generally preferred?

jsnajdr: If you'll do it inside the `batch`, then only `finishResolution` makes sense. `start` + `finish` is useful is when you want the observers to see the "tick", but when calling listeners is disables, as it is in `batch`, then `start` does nothing useful.

Mamaduka: Thanks, @jsnajdr!

packages/core-data/src/resolvers.js
```diff
@@ -165,8 +165,44 @@ export const getEntityRecord =
...
+					read: 'GET',
+					update: 'PUT',
+					delete: 'DELETE',
+				};
+				for ( const [ actionName, methodName ] of Object.entries(
+					methods
+				) ) {
+					permissions[ actionName ] =
+						allowedMethods.includes( methodName );
+				}
```
tyxla: This is getting a bit repetitive. I'm sure we're using the same logic in `canUser()`. Can we deduplicate?

packages/core-data/src/resolvers.js
```diff
@@ -165,8 +165,44 @@ export const getEntityRecord =
...
+
+				registry.batch( () => {
+					dispatch.receiveEntityRecords( kind, name, record, query );
+
+					for ( const action of [
+						'create',
+						'read',
+						'update',
+						'delete',
```
tyxla: Those will be redeclared more than once, too. Should we deduplicate?

packages/core-data/src/resolvers.js
Mamaduka: That's the plan. I will probably extract those into new utility methods.

Work in progress names:
* `getCapabilitiesFromReponseHeaders` - handles this particular case.
* `getNormalizedCapabilityKey` - handles key generation.

packages/core-data/src/resolvers.js
Mamaduka: +1. I will extract this into a constant.

packages/core-data/src/resolvers.js
Mamaduka: @tyxla, what do you think about this deduplication - 27b4b0d18db2797e37a32a6970dcc779b6a046e9?

packages/block-library/src/block/edit.js
```diff
@@ -256,27 +292,18 @@ function ReusableBlockEdit( {
...
+					recordId={ ref }
+					canOverrideBlocks={ canOverrideBlocks }
+					hasContent={ !! content }
+					handleEditOriginal={
+						onNavigateToEntityRecord
+							? handleEditOriginal
+							: undefined
+					}
+					resetContent={ resetContent }
+				/>
```
Mamaduka, line 306: Unrelated change showcases the benefits of this proposal. I am happy to extract it into a separate PR.

**Requests before with four pattern blocks on canvas:**
CleanShot 2024-07-15 at 16 52 13


**Requests after:**
CleanShot 2024-07-15 at 16 52 36

Mamaduka:
I think PR is in good shape for reviews and feedback.

packages/core-data/src/resolvers.js
```diff
@@ -165,8 +171,29 @@ export const getEntityRecord =
 					}
 				}
 
-				const record = await apiFetch( { path } );
-				dispatch.receiveEntityRecords( kind, name, record, query );
+				const response = await apiFetch( { path, parse: false } );
```
jsnajdr, line 174: Setting `parse: false` will disable the `fetchAllMiddleware`. See how it [returns early](https://github.com/WordPress/gutenberg/blob/c4332b60ee79235e76bf6926e8368a72b62b6284/packages/api-fetch/src/middlewares/fetch-all-middleware.js#L78-L81) when `options.parse` is `false`. Is this a concern here?

`fetchAllMiddleware` is relevant for fetching collections, and `getEntityRecord` fetches only one item, so it shouldn't be relevant. But what if I do:
```js
getEntityRecord( 'postType', 'post', '', { per_page: -1 } );
```
This will fetch the entire collection anyway, won't it? And should trigger `fetchAllMiddleware`.

Mamaduka: I also had similar concerns during the original discussion: https://github.com/WordPress/gutenberg/pull/43480#issuecomment-1235459202. But it shouldn't affect the `getEntityRecord` resolver.

The `getEntityRecord` supports only fetching a single entity. Any other usage is incorrect, and it's up to consumers to fix it.

Calling `wp.data.select( 'core' ).getEntityRecord( 'postType', 'post', '', { per_page: -1 } );` will make a request, but then will return `undefined`.

packages/core-data/src/utils/user-permissions.js
```diff
@@ -0,0 +1,40 @@
...
+	'read',
+	'update',
+	'delete',
+];
+
+export function getUserPermissionsFromResponse( response ) {
+	const permissions = {};
+
+	// Optional chaining operator is used here because the API requests don't
+	// return the expected result in the native version. Instead, API requests
```
tyxla: Made me re-read it to understand we're talking about the React Native implementation. 

```suggestion
	// return the expected result in the React native version. Instead, API requests
```

packages/core-data/src/resolvers.js
tyxla: Looks good, thanks! As long as it's not a public API, we can be a bit more flexible and not super critical about it.

Mamaduka: Yes. These are private utilities.

packages/core-data/src/utils/user-permissions.js
jsnajdr: The React Native special case was discussed in detail here: https://github.com/WordPress/gutenberg/pull/39144#issuecomment-1056701720

In short, RN response always have a parsed body and never include headers. Except when the response is an error.

Another thing that's hard to understand is this line:
```js
const allowedMethods = allowHeader?.allow || allowHeader || ''
```
In what case could `allowHeader` be an object with an `allow` field? This was added by @tyxla in a Lodash removal PR: https://github.com/WordPress/gutenberg/pull/43117/files#diff-fec6390e24332dec4a0c7c74f9f7b9a015ce191e0e6348361ead793ceda3991dR319

The original code was `includes( allowHeader, method )`. I still don't understand accessing the `.allow` field.

tyxla: This appears to be a mistake 🤔 Which seems to be additionally covered by a pre-existing test fixture. Let me fix.

tyxla: Fix in #63615

tyxla: Landed. Feel free to rebase @Mamaduka

Mamaduka: Done ✅

Mamaduka:
@jsnajdr, @tyxla, any remaining blocks here?

jsnajdr (approved):
Looks good to go :+1:

PR #63433, opened 2024-07-11
Dataviews: dedupe Ariakit dependencies

jsnajdr:
I noticed that there is a duplicate version of Ariakit packages installed in `packages/dataviews/node_modules` so in this PR I'm deduping them. There is a good enough version (0.3.2) in the root `node_modules`, shared with `packages/components`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53958,41 +53958,6 @@
 				"react": "^18.0.0"
 			}
 		},
-		"packages/dataviews/node_modules/@ariakit/core": {
-			"version": "0.4.1",
-			"resolved": "https://registry.npmjs.org/@ariakit/core/-/core-0.4.1.tgz",
-			"integrity": "sha512-Rdhw0/K0x+50gFvzuMW9wp+WJxpkrgiMgegRTOZSU92bv1K+6XfQWnlieIkLt2FC7pZGrDpGlS4C7ztEVF+JRg=="
-		},
-		"packages/dataviews/node_modules/@ariakit/react": {
-			"version": "0.4.1",
-			"resolved": "https://registry.npmjs.org/@ariakit/react/-/react-0.4.1.tgz",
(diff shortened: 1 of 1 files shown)


youknowriad (approved):

jsnajdr:
Haha! So the `dataviews` package declares a dependency on `@ariakit/react@0.3.2` but in reality it secretly depends on 0.4.1! Because it uses `<Combobox autoSelect="always"/>` and the `always` value is supported only since 0.4.1:

https://github.com/ariakit/ariakit/blob/main/packages/ariakit-react-core/CHANGELOG.md#new-autoselect-mode

That's why the build fails with type errors. We should probably upgrade Ariakit in the entire repo. @ciampo is that easy? 😉

youknowriad:
Interesting, the other thing I noticed that we import all of Ariakit `import *`. Is that the right approach? I feel like we should just import what we need otherwise tree shaking won't work (if that's supported by ariakit)

ciampo:
@jsnajdr I was literally typing a message very similar to yours pointing to https://ariakit.org/changelog#new-autoselect-mode

> We should probably upgrade Ariakit in the entire repo. @ciampo is that easy?

@DaniGuardiola is already on it https://github.com/WordPress/gutenberg/pull/62947 — I agree that it's better to update ariakit to the latest version _and then_ dedupe

We should definitely aim at updating `ariakit` frequently to avoid having a lot of changes to test.

> we import all of Ariakit import *. Is that the right approach? I feel like we should just import what we need otherwise tree shaking won't work (if that's supported by ariakit)

That's the suggested approach from ariakit, and I believe it should work with tree shaking — but I'll let @diegohaz reply here.

jsnajdr:
> we import all of Ariakit import *. Is that the right approach?

That's a matter of code conventions and taste, webpack tree shaking is not affected. Webpack is able to figure out which imports are actually used by the module and ignores the rest.

Mamaduka:
Are there any blockers for this PR?

DaniGuardiola:
@Mamaduka done in https://github.com/WordPress/gutenberg/pull/64066, this PR should be closed.

Mamaduka:
Thanks, @DaniGuardiola!

PR #63435, opened 2024-07-11
Core Data: Mark 'canUser' related actions resolvers as resolved

Mamaduka:
## What?
See https://github.com/WordPress/gutenberg/pull/63430#discussion_r1673944129.

PR updates the `canUser` resolver to mark related actions as resolved when their store records are populated.

## Testing Instructions
1. Open a post editor.
2. Check capabilities for an entity - `wp.data.select( 'core' ).canUser( 'update', { kind: 'root', name: 'media', id: 1342 } );`.
3. Confirm that similar actions are also marked as resolved - `wp.data.select( 'core' ).hasFinishedResolution( 'canUser', [ 'read', { kind: 'root', name: 'media', id: 1342 } ] );`

Same.

### DIFF
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -450,6 +450,15 @@ export const canUser =
 					.join( '/' );
 
 				dispatch.receiveUserPermission( key, permissions[ action ] );
+
+				// Mark related action resolutions as finished.
+				if ( action !== requestedAction ) {
+					dispatch.finishResolution( 'canUser', [
+						action,
+						resource,
+						id,
+					] );
+				}
 			}
 		} );
 	};
--- a/packages/core-data/src/test/resolvers.js
+++ b/packages/core-data/src/test/resolvers.js
@@ -308,6 +308,7 @@ describe( 'canUser', () => {
 		};
 		dispatch = Object.assign( jest.fn(), {
 			receiveUserPermission: jest.fn(),
+			finishResolution: jest.fn(),
 		} );
 		dispatch.mockReturnValue( ENTITIES );
 		triggerFetch.mockReset();


packages/core-data/src/resolvers.js
```diff
@@ -450,6 +450,15 @@ export const canUser =
 					.join( '/' );
 
 				dispatch.receiveUserPermission( key, permissions[ action ] );
+
+				// Mark related action resolutions as finished.
+				if ( action !== requestedAction ) {
+					dispatch.finishResolution( 'canUser', [
```
youknowriad, line 456: If two useSelect calls try to retrieve two different actions before the resolution, we'll still be triggering two API requests. 

A better approach might be to separate this resolver into a sub selector/resolver no?

Mamaduka: So you're saying, in some scenarios, this could invalidate the existing check at the top?

https://github.com/WordPress/gutenberg/blob/da558e272d60a7de6f823032763231be56cb6410/packages/core-data/src/resolvers.js#L397-L410

> A better approach might be to separate this resolver into a sub selector/resolver no?

Do you have an example?

youknowriad: No, actually I missed that check and you're right, we seem covered already. 

That said, I think we can avoid all of that (both marking as resolved and the check to avoid starting the resolution) by relying on a shared sub resolver/selector.

Mamaduka: > That said, I think we can avoid all of that (both marking as resolved and the check to avoid starting the resolution) by relying on a shared sub resolver/selector.

Making a note and happy to follow up on this as "polish step". While I think I have an idea of what you mean by shared sub-resolver/selector. Do you mind sharing an example so we're on the same page 😅

youknowriad: Something like:

```
export const canUser =
	( requestedAction, resource, id ) => async ( { resolveSelect } ) =>  {
		await resolveSelect.getResourcePermissions( resource, id );
	}
		
```

youknowriad (approved):

Mamaduka:
Thanks, @youknowriad!

PR #63436, opened 2024-07-11
Fix invalid css for nested fullwidth layouts with zero padding applied

### LINKED ISSUE #63432: Invalid CSS for nested fullwidth layouts with zeroed-out custom padding
An invalid CSS issue occurs when applying a padding rule of `0` to a full width group block, inside a full width group. 

This creates an unwanted horizontal scroll, as the nested fullwidth value is not properly negated. 

This is a result of resolving [#60716](https://github.com/WordPress/gutenberg/pull/60716), which ensured that the editor and front-end both negative margin for when custom padding is set. That works fine when a unit is also applied to a padding value, but when you set the padding to `0`, the `calc()` function requires a unit—otherwise the CSS is invalid and the padding value is not applied, which results in horizontal scroll. 

### Visual 

https://github.com/WordPress/gutenberg/assets/1813435/81e00fa5-4f57-40e1-8514-ab4b0d6ea7e6

richtabor:
## What?
Closes #63432 by ensuring a `px` unit is provided, to do the proper calculation when padding left or right is set to the string value of `0` via the padding controls. 

## Why?
Nested fullwidth blocks have outer padding properly negated, but when you use a custom padding value, that value is used to negate the margins, rather than the outer padding. This works great for every case where a unit is supplied with the custom value (as is always the case, other than when the value is `0`. 

## Testing Instructions

1. Open a post or page. 
2. Open the code editor within the top editor options panel. 
3. Copy and paste [this gist](https://gist.github.com/richtabor/8c9422e5ac760641b5a0a4874b8fd2c9), which configures the conditions properly (nested align full blocks, with padding zeroed out on the parent). 
4. See no horizontal scroll in the editor and the front end. 

### Before

With unintentional horizontal scroll: 

https://github.com/WordPress/gutenberg/assets/1813435/9d8854ed-2ec5-499f-a15b-f921227c5403

### After

With no horizontal scroll:

https://github.com/WordPress/gutenberg/assets/1813435/34cdd758-f882-49d4-aa9e-2ca22888c8c9

### DIFF
--- a/lib/block-supports/layout.php
+++ b/lib/block-supports/layout.php
@@ -307,6 +307,10 @@ function gutenberg_get_layout_style( $selector, $layout, $has_block_gap_support 
 			 */
 			if ( isset( $block_spacing_values['declarations']['padding-right'] ) ) {
 				$padding_right   = $block_spacing_values['declarations']['padding-right'];
+				// Add unit if 0.
+				if ( 0 == $padding_right ) {
+					$padding_right = '0px';
+				}
 				$layout_styles[] = array(
 					'selector'     => "$selector > .alignfull",
 					'declarations' => array( 'margin-right' => "calc($padding_right * -1)" ),
@@ -314,6 +318,10 @@ function gutenberg_get_layout_style( $selector, $layout, $has_block_gap_support 
 			}
 			if ( isset( $block_spacing_values['declarations']['padding-left'] ) ) {
 				$padding_left    = $block_spacing_values['declarations']['padding-left'];
+				// Add unit if 0.
+				if ( 0 == $padding_left ) {
+					$padding_left = '0px';
+				}
 				$layout_styles[] = array(
 					'selector'     => "$selector > .alignfull",
 					'declarations' => array( 'margin-left' => "calc($padding_left * -1)" ),
--- a/lib/block-supports/layout.php
+++ b/lib/block-supports/layout.php
@@ -308,7 +308,7 @@ function gutenberg_get_layout_style( $selector, $layout, $has_block_gap_support 
 			if ( isset( $block_spacing_values['declarations']['padding-right'] ) ) {
 				$padding_right   = $block_spacing_values['declarations']['padding-right'];
 				// Add unit if 0.
-				if ( 0 == $padding_right ) {
+				if ( '0' === $padding_right ) {
 					$padding_right = '0px';
 				}
 				$layout_styles[] = array(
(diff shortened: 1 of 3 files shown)


fabiankaegy (approved):
This tests well for me 👍 And the code also looks good

andrewserong (commented):
Nice fix! I just pushed a tiny fix for a linting issue.

I'll open a backport PR and link it in this PR and then see about landing this and cherry picking it into the GB 18.8 release.

andrewserong:
Backport open in: https://github.com/WordPress/wordpress-develop/pull/7036 and in trac: https://core.trac.wordpress.org/ticket/61656

andrewserong:
I just cherry-picked this PR to the release/18.8 branch to get it included in the next release: 577dc63deb1

lib/block-supports/layout.php
```diff
@@ -306,14 +306,22 @@ function gutenberg_get_layout_style( $selector, $layout, $has_block_gap_support
 			 * They're added separately because padding might only be set on one side.
 			 */
 			if ( isset( $block_spacing_values['declarations']['padding-right'] ) ) {
-				$padding_right   = $block_spacing_values['declarations']['padding-right'];
+				$padding_right = $block_spacing_values['declarations']['padding-right'];
+				// Add unit if 0.
+				if ( '0' === $padding_right ) {
```
noisysocks, line 311: Are we guaranteed to get a string back from `gutenberg_style_engine_get_styles`? I'm wondering if `$padding_right` could sometimes be `0` not `'0'`.

noisysocks:
Tested this against `wp/6.6` and looked good to me 👍

lib/block-supports/layout.php
aaronrobertshaw: At a quick glance, the value is passed through the style engine's [`is_valid_style_value`](https://github.com/WordPress/gutenberg/blob/trunk/packages/style-engine/class-wp-style-engine.php#L354) function. Which should exclude an integer `0`.

```
		protected static function is_valid_style_value( $style_value ) {
			return '0' === $style_value || ! empty( $style_value );
		}
```

aaronrobertshaw:
> Tested this against wp/6.6 and looked good to me 👍

Same result for me with `wp/6.6` 🙂

lib/block-supports/layout.php
andrewserong: Good question. I don't _think_ `gutenberg_style_engine_get_styles` will coerce the value for declarations from a quick look at the code, but padding values set by the UI will always be strings, and it's expected that values for spacing will be strings, I believe.

That said, this could be hardened in a follow-up to be `if ( '0' === $padding_right || 0 === $padding_right ) {` and similarly for the left padding.

andrewserong: Thanks, I missed that one, Aaron! 👍

PR #63441, opened 2024-07-11
Update pattern section in core-blocks reference

andizer:
Discovered a small mismatch on the blocks reference. Docs about the pattern was having the contents about the core block

## What?

The section about the core/pattern block was having the contents of the core/block 

## Why?

It fixes faulty information about the pattern block.

### DIFF
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -37,12 +37,12 @@ Add a user’s avatar. ([Source](https://github.com/WordPress/gutenberg/tree/tru
 
 ## Pattern
 
-Reuse this design across your site. ([Source](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/block))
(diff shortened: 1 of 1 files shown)


Mamaduka:
Thanks, @andizer!

I believe these details are auto-generated from the `block.json` file. The block you're referring to is called ["Pattern placeholder"](https://github.com/WordPress/gutenberg/blob/9a33e1e0c45b407dff59278936701dfbffadd616/docs/reference-guides/core-blocks.md#pattern-placeholder).

The "Pattern" used to be a "Reusable block".

andizer:
Shame on me! I didn't look further than the first thing that looked like I was searching for. Thanks for pointing me to the right one. I won't be wrong again anytime soon, however I think it might be confusing having `Pattern` and `Pattern Placeholder` on one page.

Mamaduka:
I can see how it can be confusing. The "Pattern Placeholder" is a unique block; it's not available via Inserter and is usually used by themes. So there's no name clash in the UI.

Mamaduka:
@andizer, I'm going to close the PR.

Issue #52997, opened 2023-07-26
Components: Introduce a more composable V2 of the TabPanel


chad1008:
## What?

The new V2 of `TabPanel` will be a more modular/composable approach to the component, utilizing Ariakit internals. Some of the initial work here has already begun in #52133. That PR focuses on updated the existing component to use Ariakit, while maintaining the same functionality and API surface.

The new component will most likely be named `Tabs`, as that's a bit more descriptive of the component's function and it will help avoid confusion with the `tabpanel` role that gets assigned to specific elements within the component.

## Requirements of the new component

Some specific features/capabilities this new component must have. A few of these are current blockers of a [faux tab replacement](https://github.com/WordPress/gutenberg/issues/43414)

- [Controlled Mode](https://github.com/WordPress/gutenberg/issues/27371). Specifically, this will allow consumers to set the selected tab when necessary. 
- The ability to place the `tablist` and `tabpanel` subcomponents in arbitrary locations within the DOM.
- `tablist` design flexibilty
  - Placing an auxiliary button (or other element?) such as a "Close" button beside the `tablist`
  - Optional [full-width tabs](https://github.com/WordPress/gutenberg/pull/52845)
- Limit `onSelect` calls to intentional user selections. Currently, the consumer-provided `onSelect` callback is triggered when the initial tab is selected during the first render. In V2, we intend to only trigger this callback when the user actively selects a tab, not when the component does so automatically.

### `tabpanel` tab stop/focus behavior

There are no tab stops on the individual `tabpanel` elements of the current `TabPanel` component. (...this is when I start looking forward to renaming this component to `Tabs` in the future 😉). This is something we're changing in[ the previously mentioned Ariakit migration](https://github.com/WordPress/gutenberg/pull/52133) to better align with [ARIA guidance](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction):

> When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the `tablist`, which is the `tabpanel` unless the first element containing meaningful content inside the `tabpanel` is focusable.

When that PR merges, all of the `tabpanels` will get a tab stop, but there's a second part of this guidance for cases dealing with focusable and meaningful content within the `tabpanel`.

[...]

PR #63446, opened 2024-07-11
Tabs: Vertical Tabs should be 40px min height

scruffian:
## What?
Vertical tabs should be 40px min-height.

## Why?

We don't need as much spacing for vertical tabs.

## How?

Apply `min-height: 40px` to tabs which have a `[aria-orientation='vertical']` attribute.

## Screenshots or screencast 
|Before|After|
|-|-|
|Screenshot 2024-07-11 at 16 31 26|Screenshot 2024-07-11 at 16 30 07|

### DIFF
--- a/packages/components/src/tabs/styles.ts
+++ b/packages/components/src/tabs/styles.ts
@@ -120,6 +120,12 @@ export const Tab = styled( Ariakit.Tab )`
 			opacity: 1;
 		}
 	}
+
+	[aria-orientation='vertical'] & {
+		min-height: ${ space(
+			10
+		) }; // Avoid fixed height to allow for long strings that go in multiple lines.
+	}
 `;
 
 export const TabPanel = styled( Ariakit.TabPanel )`
--- a/packages/block-editor/src/components/inserter/style.scss
+++ b/packages/block-editor/src/components/inserter/style.scss
@@ -243,6 +243,7 @@ $block-inserter-tabs-height: 44px;
 		display: block;
 		position: relative;
 		height: auto;
+		min-height: 40px;
 
 		&[aria-selected="true"] {
 			color: var(--wp-admin-theme-color);
--- a/packages/block-editor/src/components/inserter/style.scss
+++ b/packages/block-editor/src/components/inserter/style.scss
@@ -243,7 +243,6 @@ $block-inserter-tabs-height: 44px;
 		display: block;
 		position: relative;
 		height: auto;
-		min-height: 40px;
 
 		&[aria-selected="true"] {
 			color: var(--wp-admin-theme-color);
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -12,6 +12,7 @@
 -   `CustomSelectControlV2`: animate select popover appearance. ([#63343](https://github.com/WordPress/gutenberg/pull/63343))
 -   `CustomSelectControlV2`: do not flip popover if legacy adapter. ([#63357](https://github.com/WordPress/gutenberg/pull/63357)).
 -   `DropdownMenuV2`: invert animation direction. ([#63443](https://github.com/WordPress/gutenberg/pull/63443)).
+-   `Tabs`: Vertical Tabs should be 40px min height. ([#63446](https://github.com/WordPress/gutenberg/pull/63446)).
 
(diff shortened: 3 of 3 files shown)


packages/components/src/tabs/styles.ts
```diff
@@ -120,6 +120,12 @@ export const Tab = styled( Ariakit.Tab )`
 			opacity: 1;
 		}
 	}
+
+	[aria-orientation='vertical'] & {
+		min-height: ${ space(
+			10
+		) }; // Avoid fixed height to allow for long strings that go in multiple lines.
+	}
```
jeryj, line 128: Can we wrap this in a media query? For touch devices the larger 48px click area is helpful.

jeryj (commented):
This seems like the right place to fix this 👍🏻

mirka:
> We don't need as much spacing for vertical tabs.

Are we sure that makes sense in the general case? No strong opinion, but just wanted to make sure this decision wasn't colored too much by the patterns inserter case, where there happen to be a ton of tabs, and they are "secondary" to the horizontal tabs on top. (Or, it could mean that we can downsize the horizontal tabs to 40px as well.)

packages/components/src/tabs/styles.ts
mirka: > For touch devices the larger 48px click area is helpful.

FWIW, the 40px size (the new default height for basic control components) was chosen specifically to be touch-friendly, so I don't think we need to upsize on touch devices.

richtabor:
> Are we sure that makes sense in the general case?

Yes, I would think so. It's more about the target area and not the space between the words that you see. 40px is the standard large size that we should use in most cases. Horizontal tabs are fine as-is.

richtabor (approved):
LGTM 

| Before  | After |
|CleanShot 2024-07-11 at 15 18 44|CleanShot 2024-07-11 at 15 19 06|

DaniGuardiola:
As I was looking through different instances of Tabs, I realized we have many divergences with custom styles. This was one of them, so I'd like to do a full pass and make sure we can minimize those divergences across instances.

I might follow up with a PR that removes pretty much all custom styles so we can visualize it and make decisions about what styles/variants need to make it into the base component, or remain as a custom style.

packages/components/src/tabs/styles.ts
scruffian: I added a breakpoint. Unfortunately we can't use the mixin in typescript.

tyxla:
> As I was looking through different instances of Tabs, I realized we have many divergences with custom styles. This was one of them, so I'd like to do a full pass and make sure we can minimize those divergences across instances.
> 
> I might follow up with a PR that removes pretty much all custom styles so we can visualize it and make decisions about what styles/variants need to make it into the base component, or remain as a custom style.

I second this. Ideally, we should try to generalize any solutions and have as few style overrides as possible. This will make maintenance easier and will ensure we're more consistent across the entire editor.

ciampo:
Absolutely. When a different design from what the source component offers is needed, we should fall into one of these two buckets:

- if possible, we try to use what the component offers already without any overrides
- if we really, really think that the new design / behavior is needed, than we should find the bet way to add it to the source component, without any overrides.

Keeping this mindset may cause slower pace of iteration, but it's immensely beneficial in the long term, both in terms of speed of iteration and consistency across the whole ecosystem.

richtabor:
@ciampo is this clear to merge?

ciampo:
> @ciampo is this clear to merge?

I initially shared the [same doubts](https://github.com/WordPress/gutenberg/pull/63446#issuecomment-2223575001) as Lena, but if there's agreement that this change makes sense in the general case **we can go ahead after adding a CHANGELOG entry** (it can go under `Internal` since `Tabs` is still a private component) 🚀 

As a follow-up, I would also strongly recommend that we do a pass of all `Tabs` instances and remove as many style overrides as possible (ideally all of them) as @DaniGuardiola [suggested](https://github.com/WordPress/gutenberg/pull/63446#issuecomment-2225038140). Style divergences are a symptom that our designs are inconsistent and/or that the source component is not offering what consumers need, and we should fix it — especially since we'd like to [stabilize `Tabs` very soon](https://github.com/WordPress/gutenberg/issues/52997).

PR #63447, opened 2024-07-11
Update radius

jameskoster:
radius

This PR implements the radius scale outlined in https://github.com/WordPress/gutenberg/issues/63703. It is a proof of concept meant only to test different values in the proposed scale, applied across the app.


packages/base-styles/_variables.scss
```diff
@@ -92,13 +107,12 @@ $border-width-focus-fallback: 2px; // This exists as a fallback, and is ideally
...
 $helptext-font-size: 12px;
 $radius-round: 50%;
-$radius-block-ui: 2px;
 $radio-input-size: 20px;
 $radio-input-size-sm: 24px; // Width & height for small viewports.
 
 // Deprecated, please avoid using these.
 $block-padding: 14px; // Used to define space between block footprint and surrouding borders.
-
+$radius-block-ui: $radius;
```
ciampo, line 115: We should probably add linter rules to prevent folks from using those. and eventually remove them from the package.

It will also help us to make sure we're not missing any occurrences in this refactor.

packages/base-styles/_variables.scss
```diff
@@ -40,6 +40,21 @@ $grid-unit-60: 6 * $grid-unit;		// 48px
...
+/**
+ * Radius
+ *
+ */
+
+$radius: 4px;
+$radius-10: $radius * 1;
+$radius-20: $radius * 2;
+$radius-30: $radius * 3;
+$radius-full: $radius * 999;
```
ciampo, line 52: Checking here that this is on purpose vs using values like `50%` or `100%`

packages/components/src/utils/config-values.js
```diff
@@ -43,6 +43,11 @@ export default Object.assign( {}, CONTROL_PROPS, TOGGLE_GROUP_CONTROL_PROPS, {
 	colorScrollbarTrack: 'rgba(0, 0, 0, 0.04)',
 	elevationIntensity: 1,
 	radiusBlockUi: '2px',
+	radiusAtom: '2px',
+	radiusPrimitive: '4px',
+	radiusSmallContainer: '8px',
+	radiusLargeContainer: '12px',
+	radiusFull: '999px',
```
ciampo: Super nitty 😅  but technically `radiusFull` should be 4 * 999 = `3996px` in order to mimic the sass counterpart

packages/components/src/utils/config-values.js
```diff
@@ -43,6 +43,11 @@ export default Object.assign( {}, CONTROL_PROPS, TOGGLE_GROUP_CONTROL_PROPS, {
 	colorScrollbarTrack: 'rgba(0, 0, 0, 0.04)',
 	elevationIntensity: 1,
 	radiusBlockUi: '2px',
```
ciampo: Should we remove `radiusBlockUi`, since it's not exported from the package?

ciampo (commented):
Gave this a quick look and couldn't spot any major concerns.

I haven't checked yet if we missed any occurrences, and I haven't tested the code in Storybook and the editor.

packages/base-styles/_variables.scss
jameskoster: I may be mis-understanding the question but; yes, because percentage values produce a different result to absolute ones when applied to non-square elements. 
 
| Absolute | Percentage |
| --- | --- |
| Screenshot 2024-07-17 at 11 53 35 | Screenshot 2024-07-17 at 11 53 52 |

In UI design I think it's quite rare to need % based radii.

jameskoster:
@ciampo do you have suggestions about how to ship a change like this? Should it be broken down into smaller pieces? I'm conscious of how many files are touched here.

packages/base-styles/_variables.scss
ciampo: Sounds good, I just wanted to make sure the intended meaning of "full" radius 👌

ciampo:
The more cautious way to introduce these changes would be to split the work across a few PRs:

1. Introduce the new variables in base styles and components package
2. make changes across the repository, switching to the new variables (this can be further split into smaller PRs — the components package should probably have its separate one, for example)
3. mark old variables as deprecated, add linter rules, check that there are usages left of the deprecated variables

What do you think?

jameskoster:
Sounds like a plan.

jameskoster:
Closing this in favor of a more iterative rollout.

PR #63449, opened 2024-07-11
DataViews: Combine featured image and title in the posts table

### LINKED ISSUE #58012: Data views table layout: Enable fields to be visually combined with title

youknowriad:
Fixes #58012 

## What?

This PR combines the "title" with the "featured image" fields in the "pages" dataviews.

Screenshot 2024-07-11 at 7 02 54 PM

## Testing Instructions

1- Open the pages dataviews page
2- Check that the featured image is combined with the "title" field in the table layout

### DIFF
--- a/packages/edit-site/src/components/sidebar-dataviews/default-views.js
+++ b/packages/edit-site/src/components/sidebar-dataviews/default-views.js
@@ -30,27 +30,38 @@ export const defaultLayouts = {
 	[ LAYOUT_TABLE ]: {
 		layout: {
 			primaryField: 'title',
+			combinedFields: [
+				{
+					id: 'post',
+					header: __( 'Title' ),
+					children: [ 'featured-image', 'title' ],
+					direction: 'horizontal',
+				},
+			],
 			styles: {
 				'featured-image': {
 					width: '1%',
 				},
-				title: {
+				post: {
 					maxWidth: 300,
 				},
 			},
 		},
+		fields: [ 'post', 'author', 'status' ],
 	},
 	[ LAYOUT_GRID ]: {
 		layout: {
 			mediaField: 'featured-image',
 			primaryField: 'title',
 		},
+		fields: [ 'title', 'author', 'status' ],
 	},
 	[ LAYOUT_LIST ]: {
 		layout: {
 			primaryField: 'title',
 			mediaField: 'featured-image',
 		},
+		fields: [ 'title', 'author', 'status' ],
 	},
 };
 
@@ -64,8 +75,8 @@ const DEFAULT_POST_BASE = {
 		field: 'date',
 		direction: 'desc',
 	},
-	fields: [ 'title', 'author', 'status' ],
 	layout: defaultLayouts[ LAYOUT_LIST ].layout,
+	fields: defaultLayouts[ LAYOUT_LIST ].fields,
 };
 
 export function useDefaultViews( { postType } ) {
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -84,8 +84,8 @@ function useView( postType ) {
 		if ( isCustom === 'false' && layout ) {
 			return {
 				...defaultView,
+				...defaultLayouts[ layout ],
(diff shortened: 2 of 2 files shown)


ntsekouras (approved):
Looks good, thanks!

youknowriad:
I'm hesitant here because most of the time, the featured image is empty so we'll have a gray square. Let's see what designers say.

ntsekouras:
>I'm hesitant here because most of the time, the featured image is empty so we'll have a gray square. Let's see what designers say.

Hm... yeah. Actually for posts makes more sense than pages..

jameskoster:
It's also visible on the List layout, so I don't think it's a huge problem. But this is a good example of why _ideally_ fields should remain toggle-able in Table layout even when combined. 

There's a small styling issue with the image container. It's not hugging the image correctly which is messing up the spacing slightly:

Screenshot 2024-07-12 at 09 44 30

youknowriad:
> It's also visible on the List layout, so I don't think it's a huge problem. But this is a good example of why ideally fields should remain toggle-able in Table layout even when combined.

Agreed, although it's hard to implement :P

packages/edit-site/src/components/sidebar-dataviews/default-views.js
```diff
@@ -30,27 +30,38 @@ export const defaultLayouts = {
 	[ LAYOUT_TABLE ]: {
 		layout: {
 			primaryField: 'title',
+			combinedFields: [
+				{
+					id: 'post',
+					header: __( 'Title' ),
+					children: [ 'featured-image', 'title' ],
```
ellatrix, line 37: I wonder if `Front Page` and `Posts Page` could now also be a field. Not sure if that needs to be filterable though.

jameskoster: The features associated with making these fields don't seem particularly useful to me. Columns in particular would add weight to the table, especially as all records except one will be false. 

I appreciate it would make these pages a bit easier to locate, but not in a very ergonomic way. Add filter > is posts page > true isn't the quickest flow. Some alternatives we might consider could be:

1. Add a dedicated view for pages associated with settings. It would include the front page, posts page, privacy policy page (if set), products page, cart, checkout (if Woo is installed), etc.
2. Include the badge in when searching. IE if I search the All pages view for 'Front page', 'homepage', etc. then the static homepage should be listed even if the title doesn't contain those words.

Edit: I suppose they could be coded as fields, but 'hidden' in the UI except for the badge that exists today. Apologies if that was the implied idea and I went off on a tangent :)

jameskoster:
@youknowriad do you think we should proceed with this one?

youknowriad:
I feel before proceeding here, we should first address the ability to hide/show these combined fields in the "properties" section of the view config first no? Especially given that the featured image is almost always empty for pages..

jameskoster:
Perhaps something like this could work:

combining

When combined the Featured Image item is locked in position, and the re-order tools are not available. The combine/detach actions can come later.

youknowriad:
cc @oandregal as I believe he's working on a PR that alleviate some of my concerns: ability to hide the combined fields.

oandregal:
> as I believe he's working on a PR that alleviate some of my concerns: ability to hide the combined fields.

Yeah, I started looking around this area and commented on something related at [64999](https://github.com/WordPress/gutenberg/pull/64999#issuecomment-2326307488) It doesn't address the particular feature you mentioned (hide the individual fields that are part of a combination), but I'm happy to look into that after it lands.

oandregal:
Not directly related to combined fields but I feel we need to fix this before: [stress-testing primaryFields, mediaFields, and view.fields](https://github.com/WordPress/gutenberg/issues/65011). Let me know your thoughts.

PR #34945, opened 2021-09-20
Add styling for disabled options in ToggleGroupControl and try polish

stokesman:
While working on #34893, I found that `ToggleGroupControl` has no styling for disabled options. It gets an inline `pointer-events: none` thanks to Reakit but that's it. So mainly this PR is about adding some basic styles to help differentiate disabled states. I think the difference may be too subtle but I'd like to get some feedback.

Besides that there are two more small changes that I think are improvements.

- The selected indicator can be a bit smudgy (probably difficult to discernible on 2x+ screens). In order to sharpen it up it needs to be sized and placed on whole pixels so the code is revised for integers and it ends up a bit simplified too.
- While looking around the component I found myself confused about the purpose of `LabelPlaceholderView`because it is both visually and aria-hidden. I removed it and still can't figure out what it does. I will not be surprised if I'm missing something and have to restore this though.

## How has this been tested?
Manually

### Disabled Styles
Options 1 & 2 are disabled

Before | After
---|---
image | image

### Post Featured Image block inspector
Before: smudgy | After: crispy
---|---
image | image
image| image

## Types of changes
Enhancement

## Checklist:
- [ ] My code is tested.
- [ ] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [ ] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #63450, opened 2024-07-11
ToggleGroupControl: support `disabled` options

### LINKED ISSUE #57862: ToggleGroupControl: Add disabled state for entire component

#57562 surfaced a use case for a `disabled` state for ToggleGroupControl.

- The styling could be similar to the `isPressed` Button component when it is disabled.
- As for the [focus behavior when disabled](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols), we may need an additional `focusableWhenDisabled` prop on ToggleGroupControl so consumers can choose.

ciampo:
## What?

Related to #57862
Closes #57862
New take on #34945

- Adds support for `disabled` option for the `ToggleGroupControl` component
- Adds styles for disabled options

## Why?

This was a missing feature in the component, a gap that should be filled.

## How?

By forwarding the `disabled` prop to the correct internal components in the `ToggleGroupControlOptionBase` component.

## Testing Instructions

- Review new unit tests and make sure they pass reliably;
- Edit the Storybook examples, by passing the `disabled` prop to some options, and interact with the component. Make sure that disabled option are correctly skipped and impossible to select

<details>

<summary>Here's a quick patch to test the storybook examples</summary>

```diff
diff --git a/packages/components/src/toggle-group-control/stories/index.story.tsx b/packages/components/src/toggle-group-control/stories/index.story.tsx
index 92f1e60762..3b9b9a400f 100644
--- a/packages/components/src/toggle-group-control/stories/index.story.tsx
+++ b/packages/components/src/toggle-group-control/stories/index.story.tsx
@@ -81,7 +81,7 @@ Default.args = {
 	children: [
 		{ value: 'left', label: 'Left' },
 		{ value: 'center', label: 'Center' },
-		{ value: 'right', label: 'Right' },
+		{ value: 'right', label: 'Right', disabled: true },
 		{ value: 'justify', label: 'Justify' },
 	].map( mapPropsToOptionComponent ),
 	isBlock: true,
@@ -125,7 +125,12 @@ WithIcons.args = {
 	...Default.args,
 	children: [
 		{ value: 'uppercase', label: 'Uppercase', icon: formatUppercase },
-		{ value: 'lowercase', label: 'Lowercase', icon: formatLowercase },
+		{
+			value: 'lowercase',
+			label: 'Lowercase',
+			icon: formatLowercase,
+			disabled: true,
+		},
 	].map( mapPropsToOptionIconComponent ),
 	isBlock: false,
 };

```
</details>

## Screenshots or screencast 

|---|---|
| <video src="https://github.com/WordPress/gutenberg/assets/1083581/4b831aef-2915-4f87-b893-205def11d8ae" /> | <video src="https://github.com/WordPress/gutenberg/assets/1083581/6568467d-eb0d-4ddc-8116-a5b44c507ddc" /> |

## ✍️ Dev note

`ToggleGroupControlOption` components can now be marked as disabled via the newly added `disabled` prop. While disabled, an option cannot be selected by the end user.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 -   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
+-   `ToggleGroupControl`: support disabled options ([#63450](https://github.com/WordPress/gutenberg/pull/63450)).
 
 ### Internal
 
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -10,7 +10,6 @@
 
 -   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
--   `ToggleGroupControl`: support disabled options ([#63450](https://github.com/WordPress/gutenberg/pull/63450)).
 
 ### Internal
 
@@ -22,6 +21,7 @@
 
 ### Enhancements
 
+-   `ToggleGroupControl`: support disabled options ([#63450](https://github.com/WordPress/gutenberg/pull/63450)).
 -   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
 
 ## 28.3.0 (2024-07-10)
--- a/packages/components/src/toggle-group-control/test/index.tsx
+++ b/packages/components/src/toggle-group-control/test/index.tsx
@@ -80,6 +80,13 @@ const optionsWithTooltip = (
 		/>
 	</>
 );
+const optionsWithDisabledOption = (
+	<>
+		<ToggleGroupControlOption value="pizza" label="Pizza" />
+		<ToggleGroupControlOption value="rice" label="Rice" disabled />
+		<ToggleGroupControlOption value="pasta" label="Pasta" />
+	</>
+);
 
 describe.each( [
 	[ 'uncontrolled', ToggleGroupControl ],
@@ -351,6 +358,53 @@ describe.each( [
 
 				expect( expectedFocusTarget ).toHaveFocus();
 			} );
+
+			it( 'should ignore disabled radio options', async () => {
+				const mockOnChange = jest.fn();
+
+				render(
+					<Component
+						value="pizza"
+						onChange={ mockOnChange }
+						label="Test Toggle Group Control"
+					>
+						{ optionsWithDisabledOption }
+					</Component>
+				);
+
+				await sleep();
+				await press.Tab();
+
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect(
+					screen.getByRole( 'radio', { name: 'Rice' } )
+				).toBeDisabled();
+
+				// Arrow navigation skips the disabled option
+				await press.ArrowRight();
+				expect(
+					screen.getByRole( 'radio', { name: 'Pasta' } )
+				).toBeChecked();
+				expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+				expect( mockOnChange ).toHaveBeenLastCalledWith( 'pasta' );
+
+				// Arrow navigation skips the disabled option
+				await press.ArrowLeft();
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect( mockOnChange ).toHaveBeenCalledTimes( 2 );
+				expect( mockOnChange ).toHaveBeenLastCalledWith( 'pizza' );
+
+				// Clicks don't cause the option to be selected
+				await click( screen.getByRole( 'radio', { name: 'Rice' } ) );
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect( mockOnChange ).toHaveBeenCalledTimes( 2 );
+			} );
 		} );
 
 		describe( 'isDeselectable = true', () => {
@@ -420,6 +474,66 @@ describe.each( [
 						pressed: false,
 					} )
 				).toHaveFocus();
+			} );
+
+			it( 'should ignore disabled options', async () => {
+				const mockOnChange = jest.fn();
+
+				render(
+					<Component
+						value="pizza"
+						isDeselectable
+						onChange={ mockOnChange }
+						label="Test Toggle Group Control"
+					>
+						{ optionsWithDisabledOption }
+					</Component>
+				);
+
+				await sleep();
+				await press.Tab();
+
+				expect(
+					screen.getByRole( 'button', {
+						name: 'Pizza',
+						pressed: true,
+					} )
+				).toBeVisible();
+				expect(
+					screen.getByRole( 'button', {
+						name: 'Rice',
+						pressed: false,
+					} )
+				).toBeDisabled();
+
+				// Tab key navigation skips the disabled option
+				await press.Tab();
+				await press.Space();
+				expect(
+					screen.getByRole( 'button', {
+						name: 'Pasta',
+						pressed: true,
+					} )
+				).toHaveFocus();
+				expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+				expect( mockOnChange ).toHaveBeenLastCalledWith( 'pasta' );
+
+				// Tab key navigation skips the disabled option
+				await press.ShiftTab();
+				expect(
+					screen.getByRole( 'button', {
+						name: 'Pizza',
+						pressed: false,
+					} )
+				).toHaveFocus();
+
+				// Clicks don't cause the option to be selected.
+				await click(
+					screen.getByRole( 'button', {
+						name: 'Rice',
+					} )
+				);
+				expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
 			} );
 		} );
 	} );
--- a/packages/components/src/toggle-group-control/test/index.tsx
+++ b/packages/components/src/toggle-group-control/test/index.tsx
@@ -359,7 +359,7 @@ describe.each( [
 				expect( expectedFocusTarget ).toHaveFocus();
 			} );
 
-			it( 'should keep disabled radio options focusable but not selectable', async () => {
+			it( 'should ignore disabled radio options', async () => {
 				const mockOnChange = jest.fn();
 
 				render(
@@ -375,37 +375,35 @@ describe.each( [
 				await sleep();
 				await press.Tab();
 
-				const pizzaOption = screen.getByRole( 'radio', {
-					name: 'Pizza',
-				} );
-				expect( pizzaOption ).toHaveFocus();
-				expect( pizzaOption ).toBeChecked();
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect(
+					screen.getByRole( 'radio', { name: 'Rice' } )
+				).toBeDisabled();
 
-				// Arrow navigation focuses the disabled option
+				// Arrow navigation skips the disabled option
 				await press.ArrowRight();
-				const riceOption = screen.getByRole( 'radio', {
-					name: 'Rice',
-				} );
-				expect( riceOption ).toHaveFocus();
-				expect( riceOption ).toHaveAttribute( 'aria-disabled', 'true' );
-				expect( riceOption ).not.toBeChecked();
-				expect( mockOnChange ).not.toHaveBeenCalled();
-
-				// Arrow navigation focuses the next enabled option
-				await press.ArrowRight();
-				const pastaOption = screen.getByRole( 'radio', {
-					name: 'Pasta',
-				} );
-				expect( pastaOption ).toHaveFocus();
-				expect( pastaOption ).toBeChecked();
+				expect(
+					screen.getByRole( 'radio', { name: 'Pasta' } )
+				).toBeChecked();
 				expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
 				expect( mockOnChange ).toHaveBeenLastCalledWith( 'pasta' );
 
-				// Clicks don't cause the option to be selected nor to be focused.
-				await click( riceOption );
-				expect( pastaOption ).toHaveFocus();
-				expect( pastaOption ).toBeChecked();
-				expect( mockOnChange ).toHaveBeenCalledTimes( 1 );
+				// Arrow navigation skips the disabled option
+				await press.ArrowLeft();
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect( mockOnChange ).toHaveBeenCalledTimes( 2 );
+				expect( mockOnChange ).toHaveBeenLastCalledWith( 'pizza' );
+
+				// Clicks don't cause the option to be selected
+				await click( screen.getByRole( 'radio', { name: 'Rice' } ) );
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect( mockOnChange ).toHaveBeenCalledTimes( 2 );
 			} );
 		} );
 
@@ -478,7 +476,7 @@ describe.each( [
 				).toHaveFocus();
 			} );
 
-			it( 'should keep disabled button options focusable but not selectable', async () => {
+			it( 'should ignore disabled options', async () => {
 				const mockOnChange = jest.fn();
 
 				render(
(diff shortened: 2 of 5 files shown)


packages/components/src/toggle-group-control/test/index.tsx
```diff
@@ -80,6 +80,13 @@ const optionsWithTooltip = (
 		/>
 	</>
 );
+const optionsWithDisabledOption = (
+	<>
+		<ToggleGroupControlOption value="pizza" label="Pizza" />
+		<ToggleGroupControlOption value="rice" label="Rice" disabled />
+		<ToggleGroupControlOption value="pasta" label="Pasta" />
```
ciampo, line 87: @ramonjd just making sure you're ok with those food options

packages/components/src/toggle-group-control/toggle-group-control-option-base/component.tsx
```diff
@@ -52,7 +52,9 @@ function ToggleGroupControlOptionBase(
 			false
 		>,
 		// the element's id is generated internally
-		'id'
+		| 'id'
+		// due to how the component works, only the `disabled` prop should be used
+		| 'aria-disabled'
```
ciampo, line 57: I don't think it makes sense to have the options `accessibleWhenDisabled` given how the component works cc @mirka

stokesman:
Just going off the screen recording, it’s easy for me to see that an option is disabled 👍. Still, I wonder if the disabled opacity makes the text too light for a11y guidelines. Unless that’s not required of a disabled state but isn’t it?

It’s a thread-the-needle thing because even the enabled text is not that high contrast so lesser opacity will make it harder to see that an option is disabled. If the contrast is an issue perhaps we can come up with something else but ~~given the current lack of usage for this feature~~ it’s probably not too grave a concern. EDIT: I saw the existing use case Lena noted.

packages/components/src/toggle-group-control/toggle-group-control-option-base/component.tsx
mirka: I actually thought this was one of the cases where we _always_ want the option to be focusable when disabled, similar to ToolbarButton (#63102) or tabs in a tablist. It's also a [composite widget](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) where there will be a limited number of options, and the number of tab stops it potentially adds to the tab sequence is either zero or negligible.

Implementing that in the radio case for ToggleGroupControl would require some additional design work though 🤔 But at least Ariakit supports the behavior nicely ([StackBlitz demo](https://stackblitz.com/edit/qw4xvq?file=radio%2Findex.tsx)).

mirka:
@stokesman FWIW, inactive controls are [exempt](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#inactive-controls) from the contrast guidelines.

packages/components/src/toggle-group-control/toggle-group-control-option-base/component.tsx
ciampo: > Implementing that in the radio case for ToggleGroupControl would require some additional design work though

Yep. Currently, since `disabled` options are not really supported, the backdrop indicator also works as a focus indicator. 
If we make the options `accessibleWhenDisabled`, then we'll need to add separate focus rings.

> But at least Ariakit supports the behavior nicely ([StackBlitz demo](https://stackblitz.com/edit/qw4xvq?file=radio%2Findex.tsx)).

Yup. We'll probably need to add the custom logic when rendering individual `button`s

packages/components/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
+-   `ToggleGroupControl`: support disabled options ([#63450](https://github.com/WordPress/gutenberg/pull/63450)).
```
tyxla: Is this really a bug fix? Sounds like an enhancement to me.

packages/components/src/toggle-group-control/test/index.tsx
```diff
@@ -351,6 +358,53 @@ describe.each( [
...
+
+				await sleep();
+				await press.Tab();
+
+				expect(
+					screen.getByRole( 'radio', { name: 'Pizza' } )
+				).toBeChecked();
+				expect(
+					screen.getByRole( 'radio', { name: 'Rice' } )
+				).toBeDisabled();
```
tyxla, line 383: This assertion seems redundant to me since it's testing an unrelated feature (whether passing `disabled` to a toggle group control option disables that option).

packages/components/src/toggle-group-control/test/index.tsx
```diff
@@ -421,6 +475,66 @@ describe.each( [
...
+						name: 'Pizza',
+						pressed: true,
+					} )
+				).toBeVisible();
+				expect(
+					screen.getByRole( 'button', {
+						name: 'Rice',
+						pressed: false,
+					} )
+				).toBeDisabled();
```
tyxla, line 507: Same as above.

packages/components/src/toggle-group-control/toggle-group-control-option-base/styles.ts
```diff
@@ -64,6 +64,11 @@ export const buttonView = ( {
 		border: 0;
 	}
 
+	&[disabled] {
+		opacity: 0.4;
+		cursor: default;
```
tyxla, line 69: We have more and more of these disabled styles that do `opacity: 0.4`. Perhaps this is a good opportunity to generalize at some point.

tyxla (approved):
Looks good and tests well 👍 

Do we need a design ✅ just in case?

Thanks!

jasmussen:
Seems good to me, 0.4 opacity is also what appears to be used elsewhere.

ciampo:
Thank you for all the comments!

I'll look into [making disable options focusable](https://github.com/WordPress/gutenberg/pull/63450#discussion_r1675594576) and using `Tabs`-like focus styles.

I'll ask for a new round of review once that's completed

packages/components/CHANGELOG.md
ciampo: My train of thoughts was that we never stated that `disabled` options were not supported — so, from a consumer's point of view, this fixes `disabled` support.

But enhancement works too — I will change that.

packages/components/src/toggle-group-control/test/index.tsx
ciampo: I've updated tests to reflext the new "accessible when disabled" behavior, assertions have changed a bit. I will mark this as resolved, but feel free to add new comments as necessary

packages/components/src/toggle-group-control/toggle-group-control-option-base/styles.ts
ciampo: Agreed, although this PR is not the right place for it. We should probably assess the current state of shared config in the package, and decide next steps. For example, some variables come from the `base-styles` package, some are internal to the components package. Some variables are defined in Sass, some as CSS custom properties, and others in a JS object.

ciampo:
I applied changes to make disabled options keyboard-accessible (ie. focusable).

The main design change is **adding a focus ring for focused, disabled radio options**, since the active indicator (backdrop) should not follow focus in that case.

I've also updated tests to reflect the new changes.

https://github.com/user-attachments/assets/80bea05d-23c5-41f5-80f6-2c6cea27dcdf

packages/components/src/toggle-group-control/toggle-group-control-option-base/styles.ts
```diff
@@ -64,6 +64,30 @@ export const buttonView = ( {
...
+		${ ButtonContentView } {
+			opacity: 0.4;
+		}
+
+		/**
+		 * Show an additional focus ring for disabled radio items, since the
+		 * indicator won't follow keyboard focus. This is not an issue for non-radio
+		 * items, since the backdrop is already decoupled from keyboard focus.
+		 */
+		&[role='radio']:focus-visible::after {
```
ciampo: The focus ring is applied to the `after` pseudo-element so that we can control its size via the `inset` prop, so that we can leave a little gap between the focus ring and the backdrop indicator.

packages/components/src/toggle-group-control/toggle-group-control-option-base/styles.ts
```diff
@@ -64,6 +64,30 @@ export const buttonView = ( {
 		border: 0;
 	}
 
+	&[disabled],
+	&[aria-disabled='true'] {
+		cursor: default;
+		${ ButtonContentView } {
+			opacity: 0.4;
+		}
```
ciampo: Opacity is applied to the button contents, so that the focus ring can be at full opacity

packages/components/src/toggle-group-control/toggle-group-control-option-base/styles.ts
tyxla: Definitely should be a separate PR 👍

tyxla (approved):
Nothing further from my perspective. The code looks good, and everything tests well, just as demonstrated on the video.

Thanks @ciampo 🚀

ciampo:
Thank you @tyxla ! I'll also wait for @jasmussen 's approval specifically for the disabled+focused styles before merging.

jasmussen:
I'm not the best to ask here, as this particular combo of focusing a component with roving focus and what looks like a classic tab-focus style is peculiar to me, specifically because the arrow-key to focus the disabled option still appears to have the previous item selected. It looks like you have two items focused at once. Is there any prior art to look at? What do other segmented controls with disabled options do with their roving focus?

My instinct would go towards the backdrop still animating to the disabled item.

But if this is the best practice, by all means. I wouldn't think this is a pattern that should be used in the block editor, even if the component supports it.

ciampo:
@jasmussen it's a tricky one, indeed!

> because the arrow-key to focus the disabled option still appears to have the previous item selected.

That's exactly what is going on: focus is on a disabled option, which can't be selected (because disabled) — and therefore the previous item is still selected.

>  It looks like you have two items focused at once.

Yeah, that's a downside of the current approach. Here is what it looks like when showing the "focus ring" for every focused option (not just disabled):

https://github.com/user-attachments/assets/9aedac02-09b4-4b2c-9ccd-5762516835c3

> Is there any prior art to look at? What do other segmented controls with disabled options do with their roving focus?

#### Other libraries

- NextUI's [tabs](https://nextui.org/docs/components/tabs) shows both a focus ring and a visual indicator, but the `disabled`  tab behavior seems buggy (the tab is still selected even when disabled)
- [Material UI](https://mui.com/material-ui/react-toggle-button/), [Radix](https://www.radix-ui.com/primitives/docs/components/toggle-group), [Tamagui](https://tamagui.dev/ui/toggle-group), [Ant Design](https://ant.design/components/radio#components-radio-demo-radiobutton-solid) don't seem to allow disabled options to be focusable in the first place
- Ariakit and React Aria Components don't seem to have a similar example

#### Similarities with `Tabs`

`Tabs` is quite similar:
- it uses an indicator to show the currently selected item
- its tabs are keyboard-focusable even when disabled
- when a tab is disabled but focused, we keep the "active tab" indicator on the active item, and use a focus ring to show keyboard focus
- this visually works because the focus ring doesn't overlap with the indicator

Screenshot 2024-07-15 at 17 04 59

> My instinct would go towards the backdrop still animating to the disabled item.

I hacked together a preview, by having the backdrop animate to the disabled item, but with a faded opacity to convey the sense that the option is disabled. It looks promising:

https://github.com/user-attachments/assets/18744a0f-ac53-41c2-b550-a117913bd4f6

We'd need to tweak the logic of the component, since currently the backdrop indicator follows the `active` item, instead of the focused item.

We will also need to refine some aspects of the design and animation, to make sure that the text opacity and color changes are well synced with the backdrop animation.

ciampo:
I reflected on this a bit more:

1. The "backdrop still animating to the disabled item" solution visually looks the slickest, but it has one big flaw: the indicator becomes a focus indicator (instead of a "selected item" indicator), and **the user doesn't have a way of knowing what is the selected value** when focus is moved to a disabled item;

https://github.com/user-attachments/assets/18744a0f-ac53-41c2-b550-a117913bd4f6

2. The [current implementation](https://github.com/WordPress/gutenberg/pull/63450#issuecomment-2228157964) is confusing because it shows a focus ring only on disabled items, and thus the user may think that there are two focused items at once, and/or that they lose the keyboard focus when moving to a non-disabled option (plus, it's also incoherent visually)

3.  Always showing a focus ring (like we do in `Tabs`) feels like the most correct and less confusing option, but it conflicts with the design of the component.

https://github.com/user-attachments/assets/9aedac02-09b4-4b2c-9ccd-5762516835c3

---

In conclusion, I think that the better choice, given the circumstances, is **not to allow disabled option items to be keyboard focusable** in `ToggleGroupControl`.

What do y'all think ? @mirka @tyxla @jasmussen

jasmussen:
The specific behavior of this one is that I set focus on the segmented control _itself_, and now I can use arrow keys to choose an option. Tab or shift tab moves me to the next or previous focusable item:

segmented

This behavior is important to retain.

tyxla:
> This behavior is important to retain.

@jasmussen if I understand correctly, the focus between components shouldn't be affected. This is about moving between tabs, which is performed with the arrow keys. 

> In conclusion, I think that the better choice, given the circumstances, is **not to allow disabled option items to be keyboard focusable** in `ToggleGroupControl`.

From https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-disabled:

> irrelevant options in a radio group may be disabled. Disabled elements might not receive focus from the tab order. 

@ciampo in theory, yes, skipping focus for disabled tabs is a possible path forward. However, I'm a bit concerned that if disabled tabs are not focusable, users with visual impairment will miss that the button even exists. Instead, if we keep the button focusable, the assistive technology will inform them that they're working with a disabled tab. 

As a potential solution, I wonder if we could perhaps make the focus of disabled tabs lighter than the regular one?

jasmussen:
We'll have the same challenge with the [TabPanel](https://wordpress.github.io/gutenberg/?path=/docs/components-tabpanel--docs), it's also designed to be a single focusable, with arrow-key navigation.

I'm not against capturing edgecases that are theoretically used by 3rd party consumers, even if not by the block editor itself yet, but it's important we keep the key flows the components were designed for.

ciampo:
> The specific behavior of this one is that I set focus on the segmented control itself, and now I can use arrow keys to choose an option. Tab or shift tab moves me to the next or previous focusable item [...] This behavior is important to retain.

@jasmussen this behavior is not changing. What needs to be decided is, with regards to `ToggleGroupControl`:

- should disabled options be keyboard focusable?
- if so, how do we visually let the user know "where is the keyboard focus" vs "what item is currently selected" ?

> in theory, yes, skipping focus for disabled tabs is a possible path forward. However, I'm a bit concerned that if disabled tabs are not focusable, users with visual impairment will miss that the button even exists. Instead, if we keep the button focusable, the assistive technology will inform them that they're working with a disabled tab.

I agree, but (as in many other situations) we need to pick the best compromise for a better overall UX.

> As a potential solution, I wonder if we could perhaps make the focus of disabled tabs lighter than the regular one?

@tyxla I'm not sure I understand what you're proposing?

> We'll have the same challenge with the [TabPanel](https://wordpress.github.io/gutenberg/?path=/docs/components-tabpanel--docs), it's also designed to be a single focusable, with arrow-key navigation.

@jasmussen we already [solved this challenge in `TabPanel`](http://localhost:50240/?path=/story/components-tabpanel--disabled-tab) (which will soon be deprecated in favour of [`Tabs`](http://localhost:50240/?path=/story/components-experimental-tabs--disabled-tab)):

- we show a focus ring to indicate where keyboard focus is (even on disabled tabs)
- we use an underline to indicate the currently selected tab

https://github.com/user-attachments/assets/f1dd7429-4a66-43a8-833a-7d5830d5acbc

It's tricky to apply the same treatment to `ToggleGroupControl` because the indicator is a full-height backdrop and would visually clash / overlap with the focus ring. I tried this approach in a previous message:

https://github.com/user-attachments/assets/9aedac02-09b4-4b2c-9ccd-5762516835c3

jasmussen:
Understood, and I'm not the right person to ask there. As I understand, it appears to be a best practice for disabled buttons to still be focusable. I understand this especially to be the case where, for example, you press "Reset" on a button, where this same action also disables the button as there are no more filters to reset. In that case, it needs to be focusable so as to not cause a focus loss. But in the case of a segmented control, I'm struggling to come up with cases where we'd want to show a segmented control that has disabled options.

tyxla:
> > in theory, yes, skipping focus for disabled tabs is a possible path forward. However, I'm a bit concerned that if disabled tabs are not focusable, users with visual impairment will miss that the button even exists. Instead, if we keep the button focusable, the assistive technology will inform them that they're working with a disabled tab.
>
> I agree, but (as in many other situations) we need to pick the best compromise for a better overall UX.

Okay, I think we can take this step-by-step. This PR aims to add support for `disabled` options, which it does. I suggest that we open an issue for discussing the disabled option focus issue, and move forward with keeping it non-focused for now. After all, as @jasmussen also said, disabled options will likely be a rare use case.

ciampo:
I merged this PR while keeping 	`disabled` option items non-accessible, and opened [a new issue](https://github.com/WordPress/gutenberg/issues/63612) to continue the conversation.

packages/components/src/toggle-group-control/test/index.tsx
ramonjd: Sorry, I was away last week.

I would have preferred risotto instead of just "rice", but I appreciate you asking. 😆

ciampo: Ah, good point. Feel free to open a PR

PR #63451, opened 2024-07-11
Fix mobile styles for inserter pattern and media tab navigation

jeryj:
## What?

There was a regression in mobile styles for the pattern and media tabs.
- The back button was hidden behind the pane
- The background of the pane was gray
- The search input on media was white on a white background

## How?

CSS changes -- trying to match what was previously there. I don't think it's perfect, but it's much more usable now. I think the back button should also be sticky, but that could be done in a follow-up.

## Testing Instructions

On both mobile and desktop:
- Go to block inserter
- Select pattern tab
- Select a category
- Select media tab
- Select a category

Check for visual style issues.

## Screenshots or screencast 
**Before**

https://github.com/WordPress/gutenberg/assets/967608/84ce3754-23da-4422-baa5-bb2084e995fa

**After**

https://github.com/WordPress/gutenberg/assets/967608/02d34103-bd0b-40d5-a1b1-bc00cf3fabd7

### DIFF
--- a/packages/block-editor/src/components/inserter/style.scss
+++ b/packages/block-editor/src/components/inserter/style.scss
@@ -295,11 +295,6 @@ $block-inserter-tabs-height: 44px;
 	border-top: $border-width solid $gray-200;
 	box-shadow: $border-width $border-width 0 0 rgba($color: #000, $alpha: 0.133); // 0.133 = $gray-200 but with alpha.
 	outline: 1px solid transparent; // Shown for Windows 10 High Contrast mode.
-	position: absolute;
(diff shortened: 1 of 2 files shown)


jeryj:
@WordPress/gutenberg-design - This isn't perfect, but definitely better than before. I'm not sure if there were solid designs for how the mobile view of the pattern and media tabs is supposed to look

jasmussen (approved):
Good win! Works well.

MaggieCabrera (approved):
The code looks good to me and it works as expected

PR #63452, opened 2024-07-11
DataViews: Fix field rendering

youknowriad:
Related #55083 

## What?

So it turns out that we were making a noob error when rendering the different fields of the dataviews, while the fields have `render` property that is supposed to be a component, we were not rendering it as a component, instead we were just calling it as a regular function. The problem with that is that if a render property was using `useSelect` or any other hook, it would result in errors as we enable/disable the fields. 

The other issue is that it might have had a bad impact on performance potentially as all React and useSelect optimizations were probably ignored.

## Testing Instructions

1- Open the "pages" data views
2- Toggle the "author" field on and off
3- it results in an error in trunk but this PR solves the issue.

### DIFF
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import type { ReactElement, ReactNode } from 'react';
+import type { ReactElement, ComponentType } from 'react';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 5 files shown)


packages/dataviews/src/types.ts
```diff
@@ -73,7 +73,7 @@ export type Field< Item > = {
 	/**
 	 * Callback used to render the field. Defaults to `field.getValue`.
 	 */
-	render?: ( args: { item: Item } ) => ReactNode;
+	render?: ComponentType< { item: Item } >;
```
youknowriad, line 76: updating the type actually forces us to use the property properly and prevents us from doing the same error in the future.

ntsekouras (approved):
Thank you!

PR #63459, opened 2024-07-11
Feature: added `justification` option on columns block

### LINKED ISSUE #63458: Need `Justification` option on columns block
- right now there is no way to provide justification for the columns block.
- For example: suppose I have used 2 column inside a columns block and want them to have justification space between them there is no way to do this.

- using `JustifyContentControl` we can provide this option on the columns block.

### Issue Video

https://github.com/WordPress/gutenberg/assets/75293077/7e16e586-157e-4518-aa21-7f12e4ebc5f3

up1512001:
## What?
- add block control for justification.

## Why?
Fixes https://github.com/WordPress/gutenberg/issues/63458

## How?
- added `JustifyContentControl` to columns block.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/75293077/cc0e7ba3-feb3-4d8a-b620-415873183d11


fabiankaegy (changes requested):
Flex justification controls are already an option that can be enabled via the `allowJustification` sub key of the `layout` block support:

https://github.com/WordPress/gutenberg/blob/trunk/docs/reference-guides/block-api/block-supports.md#layout

I think currently `allowEditing` is removing all those layout options. But this can be refactored to be more specific using the individual support options.

packages/block-library/src/columns/block.json
```diff
@@ -49,7 +49,9 @@
 		"layout": {
 			"allowSwitching": false,
 			"allowInheriting": false,
-			"allowEditing": false,
+			"allowEditing": true,
+			"allowJustification": true,
+			"allowVerticalAlignment": false,
```
up1512001, line 54: As we are adding vertical alignment using BlockControls we don't need to display it multiple times.

```js
<BlockControls>
				<BlockVerticalAlignmentToolbar
					onChange={ updateAlignment }
					value={ verticalAlignment }
				/>
			</BlockControls>
```

up1512001: Also if we remove that then vertical alignment will not be applied to individual column.

up1512001:
Hi, @fabiankaegy can you please review this PR, thanks.

pengsteven418:
this person up1512001 copies other people content

up1512001:
@talldan can you please review this PR?

up1512001:
@fabiankaegy any update on this?

PR #63461, opened 2024-07-11
FlatTermSelector: Be more defensive about termIds

dsas:
## What?

Be more defensive against bad data in `<FlatTermSelector>` that could cause a crash.

## Why?

It's possible (likely through a bug elsewhere) for `<FlatTermSelector>` to completely crash the editor if it can't get `termIds` from the post being edited. This change prevents that crash.

The error can happen when `<FlatTermSelector>` tries to access `_termIds.join` and `_termIds.length` even when `_termIds` is undefined. This is undefined if the [current editor state doesn't have any changes for the relevant attribute](https://github.com/WordPress/gutenberg/blob/4da057ecd6958d40b8f2a83db1152320e5f161e6/packages/editor/src/store/selectors.js#L329) and the [original post doesn't have the relevant attribute](https://github.com/WordPress/gutenberg/blob/4da057ecd6958d40b8f2a83db1152320e5f161e6/packages/editor/src/store/selectors.js#L276). This shouldn't be possible normally.

## How?

Use the optional chaining operator.

## Testing Instructions

It's difficult to construct an artificial test case, but on the happy path then nothing should change at all, using the post tags in the sidebar when editing a post should continue to work exactly as it does now.

### DIFF
--- a/packages/editor/src/components/post-taxonomies/flat-term-selector.js
+++ b/packages/editor/src/components/post-taxonomies/flat-term-selector.js
@@ -86,7 +86,7 @@ export function FlatTermSelector( { slug } ) {
 
 			const query = {
 				...DEFAULT_QUERY,
-				include: _termIds.join( ',' ),
+				include: _termIds?.join( ',' ),
 				per_page: -1,
 			};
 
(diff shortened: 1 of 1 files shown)


Mamaduka (approved):
Thank you, @dsas!

The missing `termIds` is probably caused by bug somewhere else, but extra hardening won't hurt.

PR #63462, opened 2024-07-11
Separate Command Palette (command + K) and Visible Template title buttons

jeryj:
This is an odd one to keep the same design and account for improved accessibility. It's admittedly a hodgepodge. The shortcut button (Command + K) is a full-width button with the text aligned right. It has an aria-label tooltip of Command Palette so it can be accessed via a voice command of Command Palette or via the visible template name. The template name button is tabindex=-1 so it never receives focus and is also _behind_ the transparent Command + K button, so that a mouse click will always access the Command + K button and get the full width focus style. The H1 for the template name has been changed to a span. I'll add a visually hidden H1 in a follow-up commit.

### DIFF
--- a/packages/edit-site/src/components/editor/index.js
+++ b/packages/edit-site/src/components/editor/index.js
@@ -7,7 +7,7 @@ import clsx from 'clsx';
  * WordPress dependencies
  */
 import { useDispatch, useSelect } from '@wordpress/data';
-import { Button } from '@wordpress/components';
+import { Button, VisuallyHidden } from '@wordpress/components';
 import { useInstanceId } from '@wordpress/compose';
 import {
 	EditorKeyboardShortcutsRegister,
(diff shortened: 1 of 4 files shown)


jeryj:
@alexstine - I tried another route for better H1 and command palette button labelling while still keeping all the visual interactions the same. I'm hoping to thread the needle of the visual designs and the accessibility so we can get an improvement merged. I'll be the first to admit the structure is more than a little odd. Curious to hear how it works for you.

This is an alternative to https://github.com/WordPress/gutenberg/pull/63462

alexstine:
@jeryj Getting closer, but not quite.
1. There are two buttons now that do the same thing but have different labels. In my opinion, that would fail against WCAG 2.4.6. Selecting the site title does not tell a user what will happen. My first guess would be an edit modal or similar but instead it is the commands menu. One might also argue a fail against WCAG 3.2.4 because the component label could be different on every page for the first button and that is a very clear failure here.
2. Can you just make this button go away?
```
<button type="button" tabindex="-1" class="components-button editor-document-bar__title is-compact">
```
Adding a tabindex of -1 removes it from the tab order but not from screen readers other means of browsing.
3. Designers need to be willing to work with us and meet us halfway. This has been a disaster from day one and it is high time to just make this work correctly. No title, no nonsense, just a button that opens the command dialog. Simple.

Thanks.

PR #61776, opened 2024-05-18
Patterns Explorer: Makes the pagination items horizontally stacked on larger screens only

amitraj2203:
## What?
Fixes: #61754

## Why?
The current vertical stacking of pagination items in the Patterns Explorer doesn't make optimal use of available screen space. To address this, pagination items will now stack vertically only when the viewport is below 960px width. Beyond that, they will be horizontally aligned for better visibility.

## How?
This PR adds a media query to change the flex-direction.

## Testing Instructions
1. Go to the post editor > Options > Preferences > Accessibility
2. Enable the 'Show button text labels' preference.
3. Close the preferences modal dialog and open the main inserter > Patterns
4. Click 'Explore all patterns'
5. Scroll the modal dialog at the bottom.
6. Observe the pagination items.

## Screenshots or screencast 
https://github.com/WordPress/gutenberg/assets/77401999/48a2fed0-8f06-49b0-ac94-116db3a62468

PR #63466, opened 2024-07-12
Pattern Inserter: Fix pagination layout when "Show button text labels" enabled

t-hamano:
Fixes issue introduced by #61776

## What?

This PR fixes the layout of pattern inserter pagination. If "Show button text labels" is disabled, the pagination should always be aligned horizontally.

image

## Why?

[The styles here](https://github.com/WordPress/gutenberg/pull/61776/files#diff-d2bc0405a277997c32b9ecf01edb12551304959eb744c0bff96c3d483a10bc90R57-R63) are applied regardless of whether "Show button text labels" is enabled, so the pagination is always aligned vertically, which is unintentional.

## How?

Fixed to change the layout appropriately only if "Show button text labels" is enabled.

Also, these layout styles depend on where and what component the pagination is rendered, so I don't think it's appropriate to define them on the pagination component itself.

Therefore, I moved the layout styles into a stylesheet that covers both the Pattern Inserter and the Pattern Explorer.

## Testing Instructions

Test the following scenarios:

#### "Show button text labels" is disabled

It should always be horizontally aligned

inserter-disabled

#### "Show button text labels" is enabled

It should always be vertically aligned

inserter-enabled

#### "Show button text labels" is disabled

It should always be horizontally aligned

pattern-explorer-disabled

#### "Show button text labels" is enabled

If the browser width is 960px or more, it should be aligned horizontally, otherwise it should be aligned vertically.

pattern-explorer-enabled-wide

pattern-explorer-enabled-narrow

### DIFF
--- a/packages/block-editor/src/components/block-patterns-paging/style.scss
+++ b/packages/block-editor/src/components/block-patterns-paging/style.scss
@@ -24,11 +24,6 @@
 
 .show-icon-labels {
 	.block-editor-patterns__grid-pagination {
-		flex-direction: column;
-		.block-editor-patterns__grid-pagination-previous,
-		.block-editor-patterns__grid-pagination-next {
-			flex-direction: column;
-		}
 		.components-button {
 			width: auto;
 			// Hide the button icons when labels are set to display...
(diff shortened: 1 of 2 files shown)


amitraj2203 (approved):
Seems like I missed this. Thanks for catching 👍 

LGTM by looking at the screenshots.

t-hamano:
@amitraj2203 Thanks for the review!

PR #63469, opened 2024-07-12
Core Data: Use meta-store actions for resolution status

Mamaduka:
## What?
PR replaces hardcoded dispatches with meta-store actions for resolution status.

## Testing Instructions
CI checks should be green.

Same.

### DIFF
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -281,16 +281,10 @@ export const getEntityRecords =
 						.filter( ( record ) => record?.[ key ] )
 						.map( ( record ) => [ kind, name, record[ key ] ] );
 
-					dispatch( {
-						type: 'START_RESOLUTIONS',
-						selectorName: 'getEntityRecord',
-						args: resolutionsArgs,
-					} );
-					dispatch( {
-						type: 'FINISH_RESOLUTIONS',
-						selectorName: 'getEntityRecord',
-						args: resolutionsArgs,
-					} );
+					dispatch.finishResolutions(
+						'getEntityRecord',
+						resolutionsArgs
+					);
 				}
 
 				dispatch.__unstableReleaseStoreLock( lock );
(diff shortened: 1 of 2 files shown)


packages/core-data/src/resolvers.js
```diff
@@ -281,16 +281,10 @@ export const getEntityRecords =
...
-					} );
-					dispatch( {
-						type: 'FINISH_RESOLUTIONS',
-						selectorName: 'getEntityRecord',
-						args: resolutionsArgs,
-					} );
+					dispatch.finishResolution(
+						'getEntityRecord',
+						resolutionsArgs
+					);
```
Mamaduka, line 287: The actions are batched, so only `finishResolution` is needed here.

packages/core-data/src/resolvers.js
```diff
@@ -281,16 +281,10 @@ export const getEntityRecords =
...
-					} );
-					dispatch( {
-						type: 'FINISH_RESOLUTIONS',
-						selectorName: 'getEntityRecord',
-						args: resolutionsArgs,
-					} );
+					dispatch.finishResolutions(
+						'getEntityRecord',
+						resolutionsArgs
+					);
```
Mamaduka, line 287: The actions are batched, so we only need `finishResolutions`.

tyxla: Any chance something external would be relying on starting resolution first before marking it as finished?

In other words, what do we lose if we keep both `startResolutions` and `finishResolutions`?

In case you want to go without `startResolutions`, should we add a comment about why we're skipping it?

Mamaduka: Related https://github.com/WordPress/gutenberg/pull/63430#discussion_r1673992148

tyxla: Thought might be the case. Thanks for confirming 👍

tyxla (approved):
LGTM 👍 🚀

Mamaduka:
Thank you, @tyxla!

PR #63470, opened 2024-07-12
Block Bindings: Bootstrap sources defined in the server

SantosGuillamot:
## What?
This PR bootstraps the block bindings sources defined in the server so some of the properties are available without needing to register them in the client.

It follows a similar approach to `addBootstrappedBlockType`.

## Why?
It reduces complexity for developers and let us reuse the server registration.

## How?
* Add a compat filter to expose the sources in the editor settings from the server. (I created a related [pull request in core](https://github.com/WordPress/wordpress-develop/pull/7020) as well).
* Slightly change the registration to understand that a source could be bootstraped from the server.
* Create a new `addBootstrappedBlockBindingsSource` action to initialize the server sources with the minimal information.
* Add a new `bootstrapBlockBindingsSourcesFromServer` that receives the editor settings and bootstraps the sources.
* Call this function from `edit-site` and `edit-post` at the same time blocks and other block bindings sources are registered.

## Testing Instructions
1. Register a new source from the server. You can use something like this:
```
function init_source() {
	register_block_bindings_source(
		'plugin/custom',
		array(
			'label'              => _x( 'Custom Source', 'block bindings source' ),
			'get_value_callback' => function() {},
			'uses_context'       => array( 'postId', 'postType' ),
		)
	);
}

add_action( 'init', 'init_source' );
```

2. Go to a page and a paragraph connected to your source. Something like this:

```
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"plugin/custom"}}}} -->
<p>Default content</p>
<!-- /wp:paragraph -->
```

3. Check that the label is shown instead of the name in the connections panel in the right sidebar.


lib/compat/wordpress-6.7/block-bindings.php
```diff
@@ -0,0 +1,40 @@
...
+ * Adds the block bindings sources registered in the server to the editor settings.
+ *
+ * This allows to bootstrap them in the editor.
+ *
+ * @param array $settings The block editor settings from the `block_editor_settings_all` filter.
+ * @return array The editor settings including the block bindings sources.
+ */
+function gutenberg_add_server_block_bindings_sources_to_editor_settings( $editor_settings ) {
+	// Check if the sources are already exposed in the editor settings.
+	if ( isset( $editor_settings['blockBindings'] ) ) {
```
SantosGuillamot: Should the setting be named `blockBindings` or `blockBindingsSources`? Any thoughts?

packages/editor/src/bindings/post-meta.js
```diff
@@ -32,6 +32,11 @@ export default {
 			} );
 	},
 	canUserEditValue( { select, context, args } ) {
+		// Lock editing in query loop.
```
gziolo: Should it happen for all sources instead? On principle, do you think it should be never possible to edit values in the Query loop? What about the Comments block?

lib/compat/wordpress-6.7/block-bindings.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Temporary compatibility code for new functionalitites/changes related to block bindings APIs present in Gutenberg.
+ *
+ * @package gutenberg
+ */
+
+/**
+ * Adds the block bindings sources registered in the server to the editor settings.
+ *
+ * This allows to bootstrap them in the editor.
```
artemiomorales: Nit: Rather than "This allows to bootstrap them in the editor," this should probably read, "This allows them to be bootstrapped in the editor."

packages/blocks/src/api/test/registration.js
```diff
@@ -1512,6 +1512,20 @@ describe( 'blocks', () => {
 			expect( getBlockBindingsSource( 'core/testing' ) ).toBeUndefined();
 		} );
 
+		it( 'should not override label from the server', () => {
+			registerBlockBindingsSource( {
+				name: 'core/server',
+				label: 'Server label',
+			} );
```
artemiomorales, line 1520: Maybe we can add a comment to clarify that this first call to `registerBlockBindingsSource` is simulating a bootstrapped registration from the server.

artemiomorales (commented):
I tested registering a custom source and the label shows as expected 👍

bootstrapped-registration

Code-wise, aside from the two nits below, no additional feedback for the moment, looking good to me 🙌

packages/editor/src/bindings/post-meta.js
SantosGuillamot: I am not sure if it should happen in all the sources. I am thinking about pattern overrides, for example. Maybe you want to include a synced pattern there. To be honest, I don't have a strong opinion about what is the desired behavior. I just considered at that time that it was safer to lock it for post-meta until we figure out other use cases.

By the way, should we move this conversation to the original pull request where this was disabled: [link](https://github.com/WordPress/gutenberg/pull/63237)? I don't know why after the rebase this change was included in the diff, but it doesn't belong to this pull request.

lib/compat/wordpress-6.7/block-bindings.php
SantosGuillamot: Address in [this commit](https://github.com/WordPress/gutenberg/pull/63470/commits/55929a0b160f9e88a85da2a74eb50ef1969aa2ce) 🙂

packages/blocks/src/api/test/registration.js
SantosGuillamot: Address in [this commit](https://github.com/WordPress/gutenberg/pull/63470/commits/55929a0b160f9e88a85da2a74eb50ef1969aa2ce) 🙂

artemiomorales (approved):
Great, thanks for making those changes. This looks good to go on my end 👌

packages/blocks/src/api/registration.js
```diff
@@ -844,12 +848,21 @@ export const registerBlockBindingsSource = ( source ) => {
 	}
 
 	// Check the `label` property is correct.
-	if ( ! label ) {
+	if ( label && existingSource?.label ) {
```
gziolo, line 851: Is the intention to prevent the registration of the source on the client if the label gets defined in both places?

The API is still private so that might be fine, but I have mixed feelings about such strict verification. What is the reasoning behind it?

packages/blocks/src/api/test/registration.js
gziolo: What about using `addBootstrappedBlockBindingsSource` instead? This would be closer to how the logic is executed in real life scenarios. It would also remove the need for documenting the reason for calling registration twice.

gziolo:
Nice work moving that work forward 💯 

We will still have to test it with 3rd party plugins to ensure it works the same depending on the unpredictable order of loading plugins. We experienced several issues related to that with block registration so it might happen here, too. In WordPress core we have predictable scenarios as sources get bootstrapped from the server always before they get registered. In case of plugins, they might get loaded before the bootstrapping for sources gets triggered.

packages/blocks/src/store/reducer.js
```diff
@@ -377,13 +377,22 @@ export function blockBindingsSources( state = {}, action ) {
...
 					getValues: action.getValues,
 					setValues: action.setValues,
 					getPlaceholder: action.getPlaceholder,
 					canUserEditValue: action.canUserEditValue,
 				},
 			};
+		case 'ADD_BOOTSTRAPPED_BLOCK_BINDINGS_SOURCE':
+			return {
+				...state,
+				[ action.name ]: {
```
gziolo, line 391: See my https://github.com/WordPress/gutenberg/pull/63470#issuecomment-2242387646. I predict it’s the missing edge case to cover. In the case bootstrapping happens after registration, it would replace the entry with data from the server. A unit test would also help here.

packages/blocks/src/api/registration.js
SantosGuillamot: You're right that maybe we should not prevent registration and just throw a warning. I can make that change. Do you think this should be the case for other use cases? For example, when any of the properties is not a valid value?

packages/blocks/src/store/reducer.js
SantosGuillamot: That makes sense. I'll work on a follow-up pull request to cover that use case.

packages/blocks/src/api/registration.js
gziolo: Too many labels seems harmless, but it seems good to inform the dev which one was selected. In other cases, if they provide a wrong value and it’s something optional we probably could also skip it and proceed. Well, unless you want to always prevent registration then no changes are needed. Maybe being very strict is easier strategy to have and we can leave it as is 🤔

packages/blocks/src/store/reducer.js
SantosGuillamot: I have started this pull request trying to address this use case: [link](https://github.com/WordPress/gutenberg/pull/63797).

packages/blocks/src/api/registration.js
SantosGuillamot: My line of thinking right now is that too many labels won't "break" anything. But for example, if I add an invalid `setValues` or `canUserEditValue`, it will probably break the block when you try to edit it. For that reason, maybe it is safer to keep it restricted. Unless until we receive more feedback from users after using the API.

gziolo: With the invalid option setting provided, we would have to ignore them. It sounds right to wait and leave it unchanged. Thank you for discussing and discarding the alternative together with me 😄

Issue #56587, opened 2023-11-28
Gallery captions: Enhance with theme colors, possibly captions outside option


jasmussen:
As extracted from #56252, at the moment gallery captions have a hard-coded visual style that is a little clunky and occasionally hard to read:

![overlay captions](https://github.com/WordPress/gutenberg/assets/1204802/b36ec773-8f93-4d33-8558-439d3658864a)

These captions achieve some contrast through a heavy scrim gradient, hard-coded to be black, and text on top hard-coded to be white. This contrast is reduced as captions grow longer (and in fact they can grow so long as to scroll):
![multi line overlay caption](https://github.com/WordPress/gutenberg/assets/1204802/e85d6844-6875-4fa7-9535-029d0d451ae7)

When that happens, the scrim gradient is stretched, and especially the topmost lines of text lose potential contrast with the  as a result. We can do a bit better here.

**Overlay captions**

The existing default needs to be maintained, but improved. Overlay captions are useful because they maintain the silhouette of the gallery images. 

We could potentially improve these overlay captions by embracing theme colors. Every theme will define a background color, and a text color. We could use those same colors to apply to the scrim itself, the caption text, and even the inline scrollbar, to ensure there's always as much contrast as the theme itself is designed to have. 

Shown here, a dark background theme with light text, and similarly a dark scrim with white text:

![Overlay, dark](https://github.com/WordPress/gutenberg/assets/1204802/e1f6d89f-2bb6-4730-a914-339da16b7eb9)

To further ensure contrast for very long overlay captions, the scrim is no longer a gradient, it's just a solid transparent color. By leveraging the same colors as the site itself, the caption blends right in. 

The same DNA would be shared for a light theme, shown there with a beige background, black text, and simlar colors applied to scrim and caption text plus scrollbar:

![Overlay, light](https://github.com/WordPress/gutenberg/assets/1204802/5c9c2062-a859-447d-9676-a53bc892169f)

**External captions**

Looking further ahead we might consider adding a new option for galleries, to have external captions. External captions are useful for some galleries, or some captions. Shown here, in the same dark colors, but with captions below each image in a gallery.

![External, dark](https://github.com/WordPress/gutenberg/assets/1204802/8922af68-b7a1-46b2-9f22-de5bb6f9057d)

[...]

Issue #8030, opened 2018-07-18
Consider a different caption style for image galleries 


kjellr:
### **The Problem**

The current implementation of gallery captions is problematic in some circumstances.  For instance, when images get small, caption text is cut off and appears broken:

![artboard copy 3](https://user-images.githubusercontent.com/1202812/42892974-25d58a8e-8a81-11e8-88ed-14a4321151b9.png)

In this case, overflow text is scrollable, but for many users that's not clear from the appearance (#7963). It's also unwieldy to edit captions in such a small space — text and controls are hidden and appear clipped off (#6858): 

![screen shot 2018-07-18 at 12 16 05 pm](https://user-images.githubusercontent.com/1202812/42894267-66c54400-8a84-11e8-882c-f5b84a4f8db8.png)

Additionally, even when gallery images are larger, long captions can overlap a significant portion of the image, blocking it from view:

![screen shot 2018-07-18 at 12 02 32 pm](https://user-images.githubusercontent.com/1202812/42893496-7dc7b5c2-8a82-11e8-901a-c5242f06bc14.png)

---

### **Proposed Solution**

All of those issues could be avoided by adopting the caption style used for individual image captions: gray text, placed below the image. In addition to solving those issues, this may be a preferred style for some users in general. 

Alternatively, we could address those issues with the current implementation and consider adding this new caption style as a second option for users — some users will not want text overlapping their images in any case. 

---

### **Screenshots**

_Current:_
![artboard copy 3](https://user-images.githubusercontent.com/1202812/42893666-fe4d2718-8a82-11e8-81cd-221a361059ba.png)

_Proposed:_
![artboard copy 4](https://user-images.githubusercontent.com/1202812/42893672-026c0f12-8a83-11e8-83ae-5362b2667833.png)

---

_Current:_ 
![artboard](https://user-images.githubusercontent.com/1202812/42893686-0b1c4e7e-8a83-11e8-8637-1e2d6362c806.png)

_Proposed:_
![artboard 2](https://user-images.githubusercontent.com/1202812/42893696-110dc1d2-8a83-11e8-9881-ac5aa4146c33.png)

---

_Current:_
![artboard copy](https://user-images.githubusercontent.com/1202812/42893708-17712cc6-8a83-11e8-8ec1-cf4c18f66ea2.png)

_Proposed:_
![artboard copy 2](https://user-images.githubusercontent.com/1202812/42893724-20c328c4-8a83-11e8-9bc7-e50294279b30.png)

[...]

PR #31823, opened 2021-05-14
Image block: try caption as inner block

ellatrix:
## Description

caption-as-block-2

Having a caption as a block would mean it's super easy to add alignment and typography controls.
This also means we no longer need the horrible `inlineToolbar` rich text API.

Code-wise it's a pretty easy change to make.

## Checklist:
- [ ] My code is tested.
- [ ] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [ ] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #27869, opened 2020-12-23
Remove the gradient and put caption under image if is-rounded style applied

glendaviesnz:
## Description
If is-style-rounded is applied to an image in gallery the gradient overflows and there is not much room within rounded image for the caption.

This PR removes the gradient and and puts caption under the image if image is rounded:

Screen Shot 2020-12-23 at 4 57 22 PM

PR #63471, opened 2024-07-12
Try: Overlay caption w. text-shadow.

jasmussen:
## What?

Try to make progress on #8030 and #56587 by improving the caption style.

The current dark scrim is a very dark gradient to aid legibility for the white text on light backgrounds, however the sharpness of the gradient itself makes the legibility sub-optimal:

overlay-caption-before

This PR does a few things to keep the spirit of the same style, but improves it by:

* Adding a small text-shadow to the text itself.
* Applying a backdrop-blur to allow the text silhouette to avoid conflicts with complex imagery below.
* Changes the padding to be em based, and slightly more relaxed.
* Changes the color stops of the gradients to afford better contrast not just when it's one line of text, but multiple lines as well.

overlay-caption-after

## Testing Instructions

Insert galleries with overlay captions.

### DIFF
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -43,19 +43,43 @@ figure.wp-block-gallery.has-nested-images {
 			width: auto;
 		}
 
-		figcaption {
-			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.7) 0, rgba($color: $black, $alpha: 0.3) 70%, transparent);
-			bottom: 0;
-			color: $white;
-			font-size: $default-font-size;
-			left: 0;
-			margin-bottom: 0;
-			max-height: 60%;
-			overflow: auto;
-			padding: 0 8px 8px;
+		// Position caption and scrim at the bottom.
+		figcaption,
+		figcaption::after {
 			position: absolute;
+			bottom: 0;
+			right: 0;
+			left: 0;
+			max-height: 80%;
+		}
+
+		// Put the scrim on its own layer.
+		figcaption::after {
+			content: "";
+			z-index: -1;
+			height: 100%;
+			max-height: none;
+
+			// Dark gradient scrim. This can just be a solid color, as the mask will make it a
+			//background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.6) 0, rgba($color: $black, $alpha: 0.3) 50%, transparent);
+			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, rgba($color: $black, $alpha: 0.2) 100%, transparent);
+			//filter: blur(10px);
+
+			// Blur the background under the gradient scrim.
+			backdrop-filter: blur(4px);
+
+			// Crop the caption so the blur is only on the edge.
+			mask-image: linear-gradient(0deg, rgba($color: $black, $alpha: 1) 20%, rgba($color: $black, $alpha: 0) 100%);
+		}
+
+		figcaption {
+			color: $white;
+			text-shadow: 0 0 1.5px rgba($color: $black, $alpha: 0.8);
+			font-size: $default-font-size;
+			margin: 0;
+			overflow: auto;
+			padding: 1em;
 			text-align: center;
-			width: 100%;
 			box-sizing: border-box;
 			@include custom-scrollbars-on-hover(transparent, rgba($white, 0.8));
 
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -79,7 +79,6 @@ figure.wp-block-gallery.has-nested-images {
 			@include custom-scrollbars-on-hover(transparent, rgba($white, 0.8));
 
 			// Dark gradient scrim.
-			//background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, rgba($color: $black, $alpha: 0) 100%, transparent);
 			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, transparent 100%);
 
 			img {
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -56,12 +56,11 @@ figure.wp-block-gallery.has-nested-images {
 		// Create a background blur layer.
 		&:has(figcaption)::before {
 			content: "";
-			z-index: 0;
 			height: 100%;
 			max-height: 40%;
 
 			// Blur the background under the gradient scrim.
-			backdrop-filter: blur(4px);
+			backdrop-filter: blur(3px);
 
 			// Crop the caption so the blur is only on the edge.
 			mask-image: linear-gradient(0deg, rgba($color: $black, $alpha: 1) 20%, rgba($color: $black, $alpha: 0) 100%);
@@ -69,9 +68,8 @@ figure.wp-block-gallery.has-nested-images {
 
 		// Place the caption at the bottom.
 		figcaption {
-			z-index: 1;
 			color: $white;
-			text-shadow: 0 0 1.5px rgba($color: $black, $alpha: 1);
+			text-shadow: 0 0 1.5px $black;
 			font-size: $default-font-size;
 			margin: 0;
 			overflow: auto;
@@ -81,7 +79,8 @@ figure.wp-block-gallery.has-nested-images {
 			@include custom-scrollbars-on-hover(transparent, rgba($white, 0.8));
 
 			// Dark gradient scrim.
-			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, rgba($color: $black, $alpha: 0) 100%, transparent);
+			//background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, rgba($color: $black, $alpha: 0) 100%, transparent);
+			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, transparent 100%);
 
 			img {
 				display: inline;
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -44,7 +44,7 @@ figure.wp-block-gallery.has-nested-images {
 		}
 
 		// Position caption and scrim at the bottom.
-		&:has(figcaption)::after,
+		&:has(figcaption)::before,
 		figcaption {
 			position: absolute;
 			bottom: 0;
@@ -54,7 +54,7 @@ figure.wp-block-gallery.has-nested-images {
 		}
 
 		// Create a background blur layer.
-		&:has(figcaption)::after {
+		&:has(figcaption)::before {
 			content: "";
 			z-index: 0;
 			height: 100%;
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -109,6 +109,10 @@ figure.wp-block-gallery.has-nested-images {
 				margin: 0;
 				padding: 10px 10px 9px;
 				position: relative;
+				text-shadow: none;
+			}
+			&::before {
+				content: none;
 			}
 		}
 	}
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -63,7 +63,7 @@ figure.wp-block-gallery.has-nested-images {
 			backdrop-filter: blur(3px);
 
 			// Crop the caption so the blur is only on the edge.
-			mask-image: linear-gradient(0deg, rgba($color: $black, $alpha: 1) 20%, rgba($color: $black, $alpha: 0) 100%);
+			mask-image: linear-gradient(0deg, $black 20%, transparent 100%);
 		}
 
 		// Place the caption at the bottom.
--- a/packages/block-library/src/gallery/style.scss
+++ b/packages/block-library/src/gallery/style.scss
@@ -44,26 +44,21 @@ figure.wp-block-gallery.has-nested-images {
 		}
 
 		// Position caption and scrim at the bottom.
-		figcaption,
-		figcaption::after {
+		&:has(figcaption)::after,
+		figcaption {
 			position: absolute;
 			bottom: 0;
 			right: 0;
 			left: 0;
-			max-height: 80%;
+			max-height: 100%;
 		}
 
-		// Put the scrim on its own layer.
-		figcaption::after {
+		// Create a background blur layer.
+		&:has(figcaption)::after {
 			content: "";
-			z-index: -1;
+			z-index: 0;
 			height: 100%;
-			max-height: none;
-
-			// Dark gradient scrim. This can just be a solid color, as the mask will make it a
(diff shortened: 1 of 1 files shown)


beafialho:
Nice! This may be perhaps out of scope but has the option of a left/right aligned overlay caption been considered?

jasmussen:
I'd love for a separate "Caption" element to be stylable, just like buttons, forms, inputs, etc. Right now that's missing, and I'd like to see that, this would also allow font size changes, alignments, color changes, etc.

jasmussen:
Marking this one ready for review. The combination of background blur, a refined gradient, and the text shadow keeps contrast the same as before in most contexts, but improves it when many lines of overlay caption text is used. It's also close enough to the existing style that it is firmly in improvement territory, which is important given this is already shipping and in use. Would appreciate your reviews!

t-hamano (commented):
This is a great improvement. However, what I'm seeing seems a little different for some reason.

There's space on the left and right of the scrim. After some research, it seems like this is being affected by `scrollbar-gutter: stable both-edges;` which is part of the style of the `custom-scrollbars-on-hover` mixin. This issue only appears in certain environments (Windows OS, Chrome):

| Chrome | Firefox |
|--------|--------|
| image | image| 

I also believe that we need to address two issues.

If the caption is scrollable, the bottom position of the scrim will move:

image

This style needs to be removed when the Image block is in the "Rounded" style:

image

jasmussen:
Just wanted to note that I really appreciate your excellent feedback, and will address it! Other things got in the way first, but the review hasn't been forgotten. Thank you 🙏

paaljoachim:
Btw
Here is the issue related to Image caption alignments:
https://github.com/WordPress/gutenberg/issues/12997

jasmussen:
Took a stab. This was harder to address than I thought it would've been, so it needed kind of a fresh approach. What I've done now, and I believe this actually addresses all the items you brought up (though good to test again), is this:

* I kept the adjusted/tweaked dark scrim gradient on the `figcaption` element itself. This addresses issues of the scrim only ever being as tall as the caption.
* I've moved the _blur_ layer to be attached to the `figure` itself, but only when there's a figcaption. This uses `:has()`, which has support, but that entire layer will simply disappear for IE11, and is fine. 

The combination of these two, I feel, still captures the spirit of the effort, but should work better:

Screenshot 2024-08-13 at 16 54 36

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,46 @@ figure.wp-block-gallery.has-nested-images {
...
 			bottom: 0;
+			right: 0;
+			left: 0;
+			max-height: 100%;
+		}
+
+		// Create a background blur layer.
+		&:has(figcaption)::before {
+			content: "";
+			z-index: 0;
```
t-hamano: There doesn't seem to be any visual difference without `z-index: 0`, but is it necessary?

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,46 @@ figure.wp-block-gallery.has-nested-images {
...
+			// Blur the background under the gradient scrim.
+			backdrop-filter: blur(4px);
+
+			// Crop the caption so the blur is only on the edge.
+			mask-image: linear-gradient(0deg, rgba($color: $black, $alpha: 1) 20%, rgba($color: $black, $alpha: 0) 100%);
+		}
+
+		// Place the caption at the bottom.
+		figcaption {
+			z-index: 1;
```
t-hamano: There doesn't seem to be any visual difference without `z-index: 1`, but is it necessary?

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,46 @@ figure.wp-block-gallery.has-nested-images {
...
 			color: $white;
+			text-shadow: 0 0 1.5px rgba($color: $black, $alpha: 1);
```
t-hamano:
```suggestion
			text-shadow: 0 0 1.5px $black;
```

Since `rgba($color: $black, $alpha: 1)` is the same as black, I think it can be made simpler.

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,46 @@ figure.wp-block-gallery.has-nested-images {
...
-			padding: 0 8px 8px;
-			position: absolute;
+			padding: 1em;
 			text-align: center;
-			width: 100%;
 			box-sizing: border-box;
 			@include custom-scrollbars-on-hover(transparent, rgba($white, 0.8));
 
+			// Dark gradient scrim.
+			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, rgba($color: $black, $alpha: 0) 100%, transparent);
```
t-hamano:
```suggestion
			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, transparent 100%);
```

The gradient style should remain the same.

t-hamano (commented):
Thanks for the update! It looks like it's almost complete.

After testing it again, I noticed the following.

Personally, I feel that `blur(4px)` has a slightly large impact depending on the content of the image. What do you think about 2px?

4px blur:

image

2px blur:

image

Another thing I noticed is that the caption height is not constrained. Should we apply `max-height: 40%` if we want it to match the scrim?

image

jasmussen:
Great feedback as always, thank you.

> Personally, I feel that blur(4px) has a slightly large impact depending on the content of the image. What do you think about 2px?

Can we meet at 3px?

Screenshot 2024-08-16 at 09 42 05

2px feels so small to me that it might as well not be there:

Screenshot 2024-08-16 at 09 41 56

And by that I mean, if we're going to do it, it should be perceptible, otherwise best to not do it. 

I'll go through your other pieces one by one now!

packages/block-library/src/gallery/style.scss
jasmussen: Good catch on both z-indexes. They were briefly necessary in one of the versions I was testing, but in this final version they are not.

jasmussen:
> Another thing I noticed is that the caption height is not constrained. Should we apply max-height: 40% if we want it to match the scrim?

I intentionally let it be unconstrained, what I found as I tinkered was that when constrained like we do now, the scrollbar behaves somewhat curiously since it only scrolls the caption. That also means that the caption itself can be cropped, which can look awkward, like so:

Screenshot 2024-08-16 at 09 47 08

Finally for really small images with captions, if we constrain the height, there might not be space to show anything at all. 

That said, I'm happy to restore it to keep this PR smaller: it's mainly about elevating the quality and clarity of the scrim, than changing the behavior, so no strong opinion from my end. Let me know if I should restore it!

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,45 @@ figure.wp-block-gallery.has-nested-images {
...
+		&:has(figcaption)::before {
+			content: "";
+			height: 100%;
+			max-height: 40%;
+
+			// Blur the background under the gradient scrim.
+			backdrop-filter: blur(3px);
+
+			// Crop the caption so the blur is only on the edge.
+			mask-image: linear-gradient(0deg, rgba($color: $black, $alpha: 1) 20%, rgba($color: $black, $alpha: 0) 100%);
```
t-hamano:
```suggestion
			// Crop the caption so the blur is only on the edge.
			mask-image: linear-gradient(0deg, $black 20%, transparent 100%);
```

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,45 @@ figure.wp-block-gallery.has-nested-images {
...
-			position: absolute;
+			padding: 1em;
 			text-align: center;
-			width: 100%;
 			box-sizing: border-box;
 			@include custom-scrollbars-on-hover(transparent, rgba($white, 0.8));
 
+			// Dark gradient scrim.
+			//background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, rgba($color: $black, $alpha: 0) 100%, transparent);
+			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, transparent 100%);
```
t-hamano:
```suggestion
			// Dark gradient scrim.
			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.4) 0%, transparent 100%);
```

packages/block-library/src/gallery/style.scss
```diff
@@ -43,22 +43,45 @@ figure.wp-block-gallery.has-nested-images {
...
+			text-shadow: 0 0 1.5px $black;
 			font-size: $default-font-size;
-			left: 0;
-			margin-bottom: 0;
-			max-height: 60%;
+			margin: 0;
 			overflow: auto;
-			padding: 0 8px 8px;
-			position: absolute;
+			padding: 1em;
```
t-hamano, line 76: This is a non-blocker suggestion.

Because `scrollbar-gutter: stable both-edges` has been applied to the `figcaption`, the padding on the left and right looks wider, we may want to make it a little smaller.

Below is what it looks like on my environment, but how does it look on yours?

`padding: 1em`:

image

`padding: 1em 0.5em`:

image

t-hamano (approved):
LGTM! I left just a few suggestions at the end, but it all seems to work fine.

> Finally for really small images with captions, if we constrain the height, there might not be space to show anything at all.

I had overlooked this scenario. It certainly seems like it would be better to leave the height unconstrained.

I was concerned that the lightbox toggle button would become unclickable when the caption reached 100% height, but that doesn't seem to be an issue:

image

packages/block-library/src/gallery/style.scss
jasmussen: I don't disagree, though since this PR already touches a fair bit of CSS, I wonder if it's not best to do this separately?

Thank you again for the thorough reviews!

jasmussen:
Just checking, is this good to merge?

jasmussen:
I'm going to merge, but happy to follow up!

t-hamano:
Sorry for the late reply. Of course, my suggestion regarding padding could be tried in another PR 👍

paaljoachim:
Hei Joen.

I have a comment here...

In Gutenberg plugin 19.1 Gallery Image captions now look like this:

Screenshot 2024-09-03 at 21 30 59

The artist web site I am creating the artist does not want the caption on the images inside galleries so I moved these below with the following code.

```
.wp-block-gallery.has-nested-images figure.wp-block-image figcaption {
    background: 0 0;
    color: #C0BBBB;	
    position: initial;
    text-align: left;
}
```

The result looks like this:
Screenshot 2024-09-03 at 21 33 02

I should have tested this PR before it was merged with the setup I have for the web site, but I did not think about it.
This means when the caption is on the images the text looks nice and sharp if the caption is moved below the image it becomes blurry.

I am now having a difficult time using an important to remove the text shadow in the custom code.
`text-shadow: 0 0 1.5px #000;`


I added this text shadow code which seemed to work out alright with the rest of the code. So that it now becomes like so:
```
.wp-block-gallery.has-nested-images figure.wp-block-image figcaption {
    background: 0 0;
    color: #C0BBBB;	
    position: initial;
    text-align: left;
    font: 14px;
    text-shadow: 0 0 1px #C0BBBB !important;
}
```

paaljoachim:
One more thing. I noticed the bottom part of the Gallery images are blurred where the caption was earlier. 

Gutenberg 19.1 active.
Screenshot 2024-09-03 at 21 52 56

Gutenberg 19.1 not active.
Screenshot 2024-09-03 at 21 52 37

jasmussen:
Can you try adding this CSS to your custom override?

```
.wp-block-gallery.has-nested-images figure.wp-block-image:has(figcaption):before {
	-webkit-backdrop-filter: none;
	backdrop-filter: none;
}
```

That said, it's not clear to me this is necessarily an issue with the CSS as written, since it's meant to be an overlay.

One thing we found out is that, although there's no official support for captions below the image in a gallery, there's a little trick. If you apply a border-radius to the image in the gallery, the caption gets pushed below:
2024-09-04 09 36 04 tt5 local e451ed5fea24

CC: @aaronrobertshaw I noticed that this trick to move the caption below the image only works when the radius is applied locally. If you apply a border radius to the Image block in global styles, it does not change the galleries. Could it? Should it? It seems like if the behavior is in place locally, it should be in place globally as well, no?

aaronrobertshaw:
Thanks for the ping 👍 

> CC: @aaronrobertshaw I noticed that this trick to move the caption below the image only works when the radius is applied locally.

That's my understanding as well. It was introduced via https://github.com/WordPress/gutenberg/pull/27869 into the overall [refactor of the Gallery block to use inner image blocks](https://github.com/WordPress/gutenberg/pull/25940).

The way it works is the image block will add a `.has-custom-border` class when the border styles are added to the block instance. The gallery block hooks onto that class to reposition the caption.

> If you apply a border radius to the Image block in global styles, it does not change the galleries

I believe this was due to there being nothing for the block's styles to hook onto to apply the CSS to reposition the caption conditionally.

>  Could it? Should it? It seems like if the behavior is in place locally, it should be in place globally as well, no?

Could it, or should it? I'm not sure on either count.

Whether captions should be overlayed or not seems to me like it should be a block setting or perhaps a block style variation. Better yet, I'm still a fan of a [dedicated Caption block](https://github.com/WordPress/gutenberg/pull/31823) which could encapsulate this sort of behaviour and styling.

Since the Gallery block refactor, support for custom CSS in Global Styles was added. That could be leveraged to reposition a caption if desired.

We now have other blocks adopting border-radius support that can result in clipped content e.g. Group, Column etc. They are left up to the user or theme author to handle whether that's via extra padding, custom CSS etc. 

From memory, the caption repositioning was a compromise during the Gallery refactoring between the pre-existing overlaid captions and the capabilities of the image block. To maintain backward compatibility now, we might be stuck with block instances repositioning the caption.

To then bring the Global Styles side of things in line with the block instance behaviour, all I can think of are hacks to the Global Styles border panel for a single block type (bit of a non-starter), or defining a Gallery block style variation including custom CSS to handle the repositioning.

paaljoachim:
```
.wp-block-gallery.has-nested-images figure.wp-block-image:has(figcaption):before {
	-webkit-backdrop-filter: none;
	backdrop-filter: none;
}
```

Thank you Joen!
The above code removes the blurred overlay in the bottom of the images inside a gallery where the caption is moved below the image.

The ideal would be go with....

> Whether captions should be overlayed or not seems to me like it should be a block setting or perhaps a block style variation. Better yet, I'm still a fan of a https://github.com/WordPress/gutenberg/pull/31823 which could encapsulate this sort of behaviour and styling.

As @aaronrobertshaw mentioned. As it would hopefully make it openly possible for a user to choose the location (and have editing tools) of the caption in relation to the image in an Image block and a Gallery block.

---

A summery:
I am now using this code to move the caption below the images and remove the blurred background at the bottom of the images inside a gallery.

```
.wp-block-gallery.has-nested-images figure.wp-block-image figcaption {
    background: 0 0;
    color: #C0BBBB;
    position: initial;
    text-align: left;
    font: 14px;
    text-shadow: 0 0 1px #C0BBBB !important;
}

/* Removes blur in bottom of Image blocks inside galleries. */
.wp-block-gallery.has-nested-images figure.wp-block-image:has(figcaption):before {
	-webkit-backdrop-filter: none;
	backdrop-filter: none;
}

```

PR #63474, opened 2024-07-12
Term description and Author biography blocks: fix placeholder content when context is present

MaggieCabrera:
## What?

When adding a block preview example to the Term description block I noticed that even when given context, the block does not render the value of the description, just a placeholder. This adds the correct content when context is given and fixes the same problem on the Author biography block

## Why?

Because whenever we have real content that is preferable to generic placeholder text

## How?

I'm extending how we already do the name and label for Archive terms in blocks like the query title.

## Testing Instructions

1. Create a new template for a category or tag that has a description
2. Insert the Term description block. The preview should show the description. The inserted block should too
3. Create a new template for a specific author that has a bio
4. Insert the Author biography block, it should show the bio

If we do the same steps for categories/tags/authors that don't have bio/description, the placeholder should show instead.

## Screenshots or screencast 

Author with bio:

Screenshot 2024-07-12 at 11 02 17

Preview:
Screenshot 2024-07-12 at 11 02 26

Tag with description:
Screenshot 2024-07-12 at 11 02 45

Preview:
Screenshot 2024-07-12 at 11 02 51

Category with description:
Screenshot 2024-07-12 at 11 03 17

Category without description:
Screenshot 2024-07-12 at 11 03 50

### DIFF
--- a/packages/block-library/src/post-author-biography/edit.js
+++ b/packages/block-library/src/post-author-biography/edit.js
@@ -10,31 +10,22 @@ import {
 	AlignmentControl,
 	BlockControls,
 	useBlockProps,
+	store as blockEditorStore,
 } from '@wordpress/block-editor';
 import { useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
-import { store as coreStore } from '@wordpress/core-data';
 
 function PostAuthorBiographyEdit( {
-	context: { postType, postId },
 	attributes: { textAlign },
 	setAttributes,
 } ) {
-	const { authorDetails } = useSelect(
-		( select ) => {
-			const { getEditedEntityRecord, getUser } = select( coreStore );
-			const _authorId = getEditedEntityRecord(
-				'postType',
-				postType,
-				postId
-			)?.author;
-
-			return {
-				authorDetails: _authorId ? getUser( _authorId ) : null,
-			};
-		},
-		[ postType, postId ]
-	);
+	const { authorDetails } = useSelect( ( select ) => {
+		const { getSettings } = select( blockEditorStore );
+		const { __experimentalArchiveDescription } = getSettings();
+		return {
+			authorDetails: __experimentalArchiveDescription,
+		};
+	} );
 
 	const blockProps = useBlockProps( {
 		className: clsx( {
@@ -42,8 +33,7 @@ function PostAuthorBiographyEdit( {
 		} ),
 	} );
 
-	const displayAuthorBiography =
-		authorDetails?.description || __( 'Author Biography' );
+	const displayAuthorBiography = authorDetails || __( 'Author Biography' );
 
 	return (
 		<>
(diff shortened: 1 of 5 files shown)


packages/block-library/src/post-author-biography/edit.js
```diff
@@ -10,40 +10,30 @@ import {
...
 	BlockControls,
 	useBlockProps,
+	store as blockEditorStore,
 } from '@wordpress/block-editor';
 import { useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
-import { store as coreStore } from '@wordpress/core-data';
 
 function PostAuthorBiographyEdit( {
-	context: { postType, postId },
```
MaggieCabrera: This was returning undefined, so the logic after wasn't working at all. I'm not sure if it's ok to remove

jasmussen (approved):

packages/block-library/src/post-author-biography/edit.js
MaggieCabrera: mmh, this might be working on the post editor but breaking on the site editor. Now I just checked and my change broke it in the post editor :DDD

packages/block-library/src/post-author-biography/edit.js
```diff
@@ -20,17 +21,26 @@ function PostAuthorBiographyEdit( {
...
-			const { getEditedEntityRecord, getUser } = select( coreStore );
-			const _authorId = getEditedEntityRecord(
-				'postType',
-				postType,
-				postId
-			)?.author;
-
+			let authorDetails;
+			let biography;
+			if ( postType && postId ) {
```
MaggieCabrera, line 28: this only works on the post editor

jameskoster:
cc @youknowriad in case this is related to https://github.com/WordPress/gutenberg/pull/63478.

youknowriad:
Yes, let's not merge this PR yet. I'll try to review it later.

packages/edit-site/src/components/block-editor/use-site-editor-settings.js
```diff
@@ -165,6 +170,7 @@ export function useSpecificEditorSettings() {
 			// I wonder if they should be set in the post editor too
 			__experimentalArchiveTitleTypeLabel: archiveLabels.archiveTypeLabel,
 			__experimentalArchiveTitleNameLabel: archiveLabels.archiveNameLabel,
+			__experimentalArchiveDescription: archiveLabels.archiveDescription,
```
youknowriad, line 173: Let's avoid adding random block editor settings like that that are only needed for a given block.

I'm not sure I understand the PR probably but from my reading, you want to retrieve the "template" description kind of and use it in a block.

I've just merged https://github.com/WordPress/gutenberg/pull/63478 which fixes this for the other two labels and the QueryTitle block and I believe it might be possible to reuse the same hook (that is now in block-library) directly in the post author and term description blocks here instead of adding a new setting. We could add a new returned value to the hook.

MaggieCabrera:
I think it's easier to close this one and work directly on top of trunk after your changes, thanks for the feedback @youknowriad

Issue #63456, opened 2024-07-11
Editor crash from block deselection in “Zoom Out” mode


stokesman:
### Description

While zoomed out, deselecting a block crashes the editor. I first happened across this by undoing a pattern insertion. Now I found that just clicking somewhere that doesn’t select a new block but does deselect the currently selected one will crash it too.

Here’s the error from the crash boundary:

```
TypeError: Cannot destructure property 'name' of 'getBlock(...)' as it is null.
    at http://localhost:8888/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=0739551982adc596213c:28422:7
    at http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:1705:70
    at __unstableMarkListeningStores (http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:4402:23)
    at Object.__unstableMarkListeningStores (http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:4477:30)
    at updateValue (http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:1705:34)
    at http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:1746:5
    at useMappingSelect (http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:1771:7)
    at useSelect (http://localhost:8888/wp-content/plugins/gutenberg/build/data/index.min.js?ver=65dd8ddc0f9260605b50:1855:58)
    at ZoomOutToolbar (http://localhost:8888/wp-content/plugins/gutenberg/build/block-editor/index.min.js?ver=0739551982adc596213c:28409:78)
    at renderWithHooks (webpack://gutenberg/./node_modules/react-dom/cjs/react-dom.development.js?:15486:18)
```

### Via block deselection
1. With the Zoom Out experiment enabled
2. Open a post that has content
3. Open the inserter, select a pattern category
4. Select a block
5. Click somewhere that will deselect the block without selecting another
6. Note the crashed editor

### Via Undo
1. With the Zoom Out experiment enabled
2. Go to the Site editor and select a page to edit
[...]

PR #60123, opened 2024-03-22
Add a vertical toolbar for zoom out mode

MaggieCabrera:
## What?

Closes https://github.com/WordPress/gutenberg/issues/59737

Ariakit supports vertical toolbars so we implemented a new variant to the main `Toolbar` component and applied it to a new toolbar specifically for the zoom-out mode.

## Why?

Zoom-out mode needs a toolbar that doesn't cover the content we are inserting. We needed to leverage the Ariakit toolbar for accessibility, so the black pseudo-toolbar that we were using wasn't enough.

## How?

By implementing a toolbar that extends `NavigableToolbar` and adding a new orientation prop that ariakit uses

## Testing Instructions

Open the inserter and open a pattern category
Click on a section, the toolbar should be visible on the side. All buttons should work as usual

### Testing Instructions for Keyboard

When you focus on a section, move focus to the toolbar and use up and down arrows to navigate between the toolbar buttons

## Screenshots or screencast 

Screenshot 2024-05-10 at 11 38 44

PR #63476, opened 2024-07-12
Zoom out mode: translate toolbar delete button

t-hamano:
Related #60123

## What?

While investigating #63456, I noticed that one label did not have the translation function applied.

## Testing Instructions

All CIs should be ✅.


MaggieCabrera (approved):
thanks for catching this one!

PR #43285, opened 2022-08-16
[Block Library - Query Loop]: Honour intended post type when previewing patterns and when replacing with patterns

ntsekouras:
## What?
Part of: https://github.com/WordPress/gutenberg/issues/42533

From the issue:
>Currently in Query Loop and Template Parts we use the pattern setup flow, where a user can choose and insert directly a semantic pattern or some default block variation(scope: block).

> The problem here is that patterns and the scope block variations have different attributes and this results to override the Posts List variation attributes.

>The problem is for any block that uses pattern setup flow like Template Part, but is not visible there because that block's variations only contain the area attribute which is also the one used in calculating the active variation(in isActive function).

>We need to find a good way to preserve some important attributes(or part of them if is an object like query) and in general make sure that these retained attributes do no affect the presentation/design.

>This also applies to the `replace` with pattern flow.

## How?
This PR doesn't explore some generic solution(we might not even need one) as Query Loop is quite complex and I think it's okay to have some specific handling to accommodate this, and also become a lot more flexible and open to extension by third parties in the future.

In the pattern setup and replace flow, and the block variations(`scope`) flow I'm updating the Query Loop blocks to be inserted by retaining the existing `query.postType` and `query.inherit` options. These two properties are fundamental for the expected functionality of the block and don't affect its design and presentation.

What this means in practice is that if we have a Query Loop with `postType: product` and choose to `replace` with another patterns(we want to update the design), the `postType and inherit` properties will be retained.

#### Replace flow
1. In a Query Loop change the `postType` to anything than `post`, as is the one used by most patterns.
2. Click `replace` in block toolbar and select another pattern
3. Observe that `postType and inherit` are retained.

#### Scope block variations
1. Update the postType to something else(ex `page`) to the Posts List (inserter) variation [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/query/variations.js#L46).
2. Insert a `Posts List` block and choose `start blank`.
3. Select any of them and observe that the `postType` is retained

## Screenshots or screencast 

https://user-images.githubusercontent.com/16275880/185621050-63a6ae85-2b30-45ee-aad2-138dad8f82ee.mov

PR #46410, opened 2022-12-08
[Block Library - Query Loop]: Connect scoped `block` variation to `inserter` variations

ntsekouras:
## What?
Resolves: https://github.com/WordPress/gutenberg/issues/46367
Maybe alternative of: https://github.com/WordPress/gutenberg/pull/46369

Currently in Query Loop set up flow a user can choose to `start blank`, where the scoped `block` variations are suggested. If extenders of Query Loop have added scoped `inserter` variations(ex `List of Products`) and choose to `start blank` the default `block` variations will be suggested, which might make no sense or just unwanted.

For such `inserter` variations there is already in place the mechanism to suggest `connected` patterns. So this PR implements something similar for `connected` scoped `block` patterns to other Query Loop variations(`inserter`).

I have used a convention:
> Helper hook that determines if there is an active variation of the block
 and if there are available specific scoped `block` variations connected with
 this variation.
 If there are, these variations are going to be the only ones suggested
 to the user in setup flow when clicking to `start blank`, without including
 the default ones for Query Loop.
 If there are no such scoped `block` variations, the default ones for Query
Loop are going to be suggested.
The way we determine such variations is with the convention that they have the `namespace`
 attribute defined as an array. This array should contain the names(`name` property) of any
 variations they want to be connected to.
For example, if we have a `Query Loop` scoped `inserter` variation with the name `products`,
we can connect a scoped `block` variation by setting its `namespace` attribute to `['products']`.
If the user selects this variation, the `namespace` attribute will be overridden by the
main `inserter` variation.

but that was my quick first thought for the implementation and can be adjusted..

In general scoped `block` variation are meant to be a simple starting point that usually 

## Testing Instructions

<details>
 <summary>Register the below two variations(one `inserter` and one `scope`)</summary>
<pre> <code>{
		description: 'Variation',
		name: 'variation_query_loop',
		title: 'Variation Query Loop',
		icon: 'admin-home',
		attributes: {
			displayLayout: {
				type: 'flex',
				columns: 3,
			},
			namespace: 'inserter_namespace_variation_query_loop',
			query: {
				perPage: 9,
				pages: 0,
				offset: 0,
				postType: 'page',
				order: 'desc',
				orderBy: 'date',
				author: '',
				search: '',
				exclude: [],
				sticky: '',
				inherit: false,
			},
		},
		isActive: [ 'namespace' ],
		scope: [ 'inserter' ],
	},
	{
		description: 'Connected Variation',
		name: 'connected-block-variation',
[...]

PR #63477, opened 2024-07-12
Query Loop: Fix 'block' scoped variations to get the `query` defaults

ntsekouras:
## What?
Follow up of: https://github.com/WordPress/gutenberg/pull/63353

The intention of the linked PR was to allow the `block` scoped Query Loop variations to get the default query values and especially the `inherit: true` value. That PR did would normally work, but with the nuances of `object` handling attributes in GB and specifically also in Query Loop which has [other nuances](https://github.com/WordPress/gutenberg/pull/43285), we need to make the changes in this PR too.

## Testing Instructions
1- Insert Query Loop block
2- Click "Start blank"
3- Pick one of the proposed variations
4- Notice that the inserted query has the "inherit" toggle enabled.

### DIFF
--- a/packages/block-library/src/query/edit/query-placeholder.js
+++ b/packages/block-library/src/query/edit/query-placeholder.js
@@ -122,6 +122,7 @@ function QueryVariationPicker( {
 						setAttributes( {
 							...variation.attributes,
 							query: {
+								...attributes.query,
 								...variation.attributes.query,
 								postType:
 									attributes.query.postType ||
--- a/packages/block-library/src/query/edit/query-placeholder.js
+++ b/packages/block-library/src/query/edit/query-placeholder.js
@@ -25,7 +25,6 @@ export default function QueryPlaceholder( {
 	clientId,
 	name,
 	openPatternSelectionModal,
-	setAttributes,
 } ) {
 	const [ isStartingBlank, setIsStartingBlank ] = useState( false );
 	const blockProps = useBlockProps();
@@ -64,7 +63,6 @@ export default function QueryPlaceholder( {
 			<QueryVariationPicker
 				clientId={ clientId }
 				attributes={ attributes }
-				setAttributes={ setAttributes }
 				icon={ icon }
 				label={ label }
 			/>
@@ -101,13 +99,7 @@ export default function QueryPlaceholder( {
 	);
 }
 
-function QueryVariationPicker( {
-	clientId,
-	attributes,
-	setAttributes,
-	icon,
-	label,
-} ) {
+function QueryVariationPicker( { clientId, attributes, icon, label } ) {
 	const scopeVariations = useScopedBlockVariations( attributes );
 	const { replaceInnerBlocks } = useDispatch( blockEditorStore );
 	const blockProps = useBlockProps();
@@ -118,19 +110,6 @@ function QueryVariationPicker( {
 				label={ label }
 				variations={ scopeVariations }
 				onSelect={ ( variation ) => {
-					if ( variation.attributes ) {
-						setAttributes( {
-							...variation.attributes,
-							query: {
-								...attributes.query,
-								...variation.attributes.query,
-								postType:
-									attributes.query.postType ||
-									variation.attributes.query.postType,
-							},
-							namespace: attributes.namespace,
-						} );
-					}
 					if ( variation.innerBlocks ) {
 						replaceInnerBlocks(
 							clientId,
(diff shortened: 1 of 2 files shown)


packages/block-library/src/query/edit/query-placeholder.js
```diff
@@ -122,6 +122,7 @@ function QueryVariationPicker( {
 						setAttributes( {
 							...variation.attributes,
 							query: {
+								...attributes.query,
 								...variation.attributes.query,
 								postType:
```
youknowriad: I actually don't understand why we're doing this, Why are we overriding namespace and post type here?

youknowriad: I think we should just call `setAttributes( variation.attributes )` since this variation picked is only used when starting blank.

ntsekouras: >I think we should just call setAttributes( variation.attributes )

Hm.. that could work now. If I remember correctly, this was needed because before we intentionally had the `inherit:false` and since properties of  object attributes are not merged, we needed to also provide the rest of the props in the variation declaration. But since we could be here from another `inserter` Query variation, we needed to respect some important attributes like `postType`.  ~~I'll update to just set the attributes then~~.

ntsekouras: Ah.. it's also about this PR: https://github.com/WordPress/gutenberg/pull/46410. `namespace` is used in order to be able to connect variations with variations 😄 .  So we cannot do just `setAttributes`..

youknowriad: > namespace is used in order to be able to connect variations with variations 😄 . So we cannot do just setAttributes..

What? I don't understand this sentence, can you clarify?

youknowriad: What is "namespace" attribute and why do we need it?

ntsekouras: Can you give a quick read of [this PR's](https://github.com/WordPress/gutenberg/pull/46410) description? That would be it.

In general the `namespace` attribute is something like an identifier for Query Loop in order to help with active variations detection. It's mostly a result of `query` being an object attribute and nuances that for Query Loop extenders register variations and there is the need to have a way to detect their variation for various things (control additions, etc..) in both client and server side.

youknowriad: Ok, so one thing is: I'm not sure we should have a `namespace` attribute. 

But still, I can't figure out how it relates to this PR. We're starting blank, we're picking variations from a predefined list that are scoped to "block", we should just the attribute of the selected variation.

youknowriad: I've read the related PR but I can't figure out why we show the "start blank" option at all for these query variations. IF you provide a variation for the query block, you don't need to show the "block scoped variations" at all, you're not starting blank as you picked a variation no?

youknowriad: I'm understand the issue more, I think combining "block variations" with "inserter variations" was a mistake from the start. Do we know of people actually using that or is it theoretical?

youknowriad: Here's another proposal: "block variations" shouldn't update the query attribute at all, they should just update the inner blocks. So what about removing the setAttributes here entirely?

ntsekouras: >IF you provide a variation for the query block, you don't need to show the "block scoped variations" at all, you're not starting blank as you picked a variation no?

Not necessarily. You can register variations that don't provide inner blocks.

>I'm understand the issue more, I think combining "block variations" with "inserter variations" was a mistake from the start. Do we know of people actually using that or is it theoretical?


There were many issues from extenders and was definitely used. Not sure if some plugins still use it, but there was a lot of traction to get this working. I think the combination of `block and inserter` variations stems from the complexity of this specific block and that's why there was a `local` implementation.

>Here's another proposal: "block variations" shouldn't update the query attribute at all, they should just update the inner blocks. So what about removing the setAttributes here entirely?

This should be a breaking change, no?

youknowriad: > This should be a breaking change, no?

I don't think so. For me, it just clarifies the behavior. "Inserter" is about defining the whole block, "block" scoped variations is about the content (which is already the case btw)

ntsekouras: This makes sense. Let's try this.

youknowriad (approved):

Issue #59082, opened 2024-02-15
Some template blocks in data view previews do not match their display in the editor


jameskoster:
Blocks like "Archive Title" include contextual placeholders. For instance of you're editing the Author archives template it will display 

> Author: Name

or 

> Author name

depending on the block settings:

<img width="1710" alt="Screenshot 2024-02-15 at 14 29 52" src="https://github.com/WordPress/gutenberg/assets/846565/e1b031fa-f5ef-4828-8174-3af8e2c1dc2e">

That contextual placeholder appears correctly in the 'site view' when you browse a template:

<img width="890" alt="Screenshot 2024-02-15 at 14 32 42" src="https://github.com/WordPress/gutenberg/assets/846565/f5f84093-2d64-4d04-a633-165e8e196d02">

However it doesn't carry over to the preview in data views, where it falls back to the generic placeholder:

<img width="343" alt="Screenshot 2024-02-15 at 14 31 55" src="https://github.com/WordPress/gutenberg/assets/846565/3c3ea564-8940-46f8-b0c6-67d3cc9581df">

cc @ntsekouras as iirc you worked on these contextual placeholders.

youknowriad:
Moving this out of 6.6 as we get close to release.

youknowriad:
Not sure how to reproduce this. Can you share more details (which theme and template to use), for me in 2024 and "archives" template it shows the same thing between preview and editor: Archive type: Name

jameskoster:
@youknowriad you have to create a more specific archive template, e.g. `author.html`.

PR #63478, opened 2024-07-12
Query Title block: Rely on the editor store to apply the right archive title placeholder.

youknowriad:
This is the first step to solve this but #59082 

## What?

This PR refactors the way the Query Title block retrieves the current template in order to change the placeholder contextually. Instead of relying on a random block editor setting, it relies on the actual template used in the "core/editor" store directly.

The reason we can do this is because some blocks of block-library are WP specific and can actually make use of the editor package. It's the case for the Query Title block.

## Testing Instructions

1- Create tag archive block template
2- Make the sure the title shows the right placeholder: "Tag: name".

### DIFF
--- a/packages/block-library/src/query-title/use-archive-label.js
+++ b/packages/block-library/src/query-title/use-archive-label.js
@@ -1,0 +1,99 @@
+/**
+ * WordPress dependencies
+ */
+import { store as coreStore } from '@wordpress/core-data';
+import { useSelect } from '@wordpress/data';
+
+export function useArchiveLabel() {
+	const templateSlug = useSelect( ( select ) => {
+		// @wordpress/block-library should not depend on @wordpress/editor.
+		// Blocks can be loaded into a *non-post* block editor, so to avoid
+		// declaring @wordpress/editor as a dependency, we must access its
+		// store by string.
+		// The solution here is to split WP specific blocks from generic blocks.
+		// eslint-disable-next-line @wordpress/data-no-store-string-literals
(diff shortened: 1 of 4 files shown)


packages/block-library/src/query-title/use-archive-label.js
```diff
@@ -0,0 +1,99 @@
...
+			( currentPostType === 'wp_template' ? getCurrentPostId() : null );
+
+		return templateId
+			? select( coreStore ).getEditedEntityRecord(
+					'postType',
+					'wp_template',
+					templateId
+			  )?.slug
+			: null;
+	}, [] );
```
youknowriad, line 29: Above this line is the new addition. The rest is the same as before.

jameskoster (approved):
> 1- Create tag archive block template
> 2- Make the sure the title shows the right placeholder: "Tag: name".

👍

ntsekouras (commented):
Looks good. The only visible change though is that since the hook is inside the block's edit function, there is the quick flash updates of titles, when the selector has resolved. Should we wait for the resolution? 🤔

youknowriad:
I don't really understand why it's different in terms of loading than trunk, we're calling the exact same selectors. 🤔

youknowriad:
I see now, It's actually the same as before, only difference is the "global loading" was taking the resolver into consideration and in this PR it's not. I'm guessing because the block is rendered after some "ticks". 

Going to merge now I guess but we can look at improving that if needed.

PR #63480, opened 2024-07-12
Storybook: Remove popover-related height buffers

mirka:
## What?

Removes now-unnecessary height buffers for `Popover`-based stories in Storybook.

## Why?

Due to #53889, `Popover`s are now rendering at the end of the document rather than within the each story div, which used to cut them off visually.

## Testing Instructions

See stories for affected components in the main Docs view. The popovers should appear above the story div, and not be cut off.

## Screenshots or screencast 

| Before | After |
|--------|-------|
|Story with height buffer|Story without height buffer|

### DIFF
--- a/packages/components/src/dropdown-menu-v2/stories/index.story.tsx
+++ b/packages/components/src/dropdown-menu-v2/stories/index.story.tsx
@@ -65,18 +65,6 @@ const meta: Meta< typeof DropdownMenu > = {
 			source: { excludeDecorators: true },
 		},
 	},
-	decorators: [
-		// Layout wrapper
-		( Story ) => (
-			<div
-				style={ {
-					minHeight: '300px',
-				} }
-			>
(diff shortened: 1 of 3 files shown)


tyxla (approved):
Nice cleanup, thanks 🙌 

Looking good:
Screenshot 2024-07-12 at 15 36 22
Screenshot 2024-07-12 at 15 36 28
Screenshot 2024-07-12 at 15 36 48



Feel free to ship, the performance test failure is unrelated (it also occurs in `trunk` ATM).

🚀

ciampo:
The `CustomSelectControl` stories would benefit from this change too

mirka:
> The `CustomSelectControl` stories would benefit from this change too

I tried, and that's how I noticed that `CustomSelectControl` doesn't portal the popover yet 🥺

PR #63138, opened 2024-07-04
Pages: Trash view should default to table layout.

jorgefilipecosta:
Part of https://github.com/WordPress/gutenberg/issues/63128.
Makes the page trash view default to the table view.

cc: @jameskoster 

## Testing Instructions
I opened the pages list at wp-admin/site-editor.php?postType=page.
I switched to the trash view, and I verified the default view type of that view was table, while for the other the default list was kept.

PR #63481, opened 2024-07-12
Revert "Pages: Trash view should default to table layout."

jorgefilipecosta:
Reverts: https://github.com/WordPress/gutenberg/pull/63138/
Although the CI passed on the PR it caused the CI to fail on the trunk. Debugging it seems the CI error on the performance test is legitimate. I'm going to revert the PR and address the issue separately.


Mamaduka (approved):
Thanks, @jorgefilipecosta!

I think the original PR didn't include new perf tests, which are currently failing - #63170.

You might have to merge this with failing Perf job, because code still exists on trunk.

PR #63484, opened 2024-07-12
Writing flow: select next block on Enter key

### LINKED ISSUE #54984: Template vs Content Editing UX: enable the ability to hit return from post title to begin writing

I wanted to pull this experience out of this prior issue https://github.com/WordPress/gutenberg/issues/42481 as it greatly impacts the writing experience in the Site Editor and has an outsized impact compared to locked blocks due to the current UX. Currently, when you click on the Post Title, you're unable to click return to start writing on the first line. This differs from the Post Editor, making it hard to start writing from the top of a page:

https://github.com/WordPress/gutenberg/assets/26996883/b26e250b-20a2-42aa-9a84-030a3485c9ab

This came up previously [as part of FSE Outreach Program feedback](https://github.com/WordPress/gutenberg/issues/42481#issuecomment-1691921578) that I'll reshare below:

>I created a new page in Site Editor > Pages and titled it ‘Resume’.
>I added content such as headings and text and tried to insert a table of contents block.
>I wanted to insert the table of contents block between the title and the first block, so I moved the cursor to the page title with the mouse and pressed enter, but could not move it to the content area.
>When creating pages in the block editor as before, I was able to bring the cursor to the top of the content using the same operation (strictly speaking, I was able to insert a paragraph block at the beginning), so I felt that a similar UX would be fine.

Allow the ability to hit return out of the Post Title block.

cc @SaxonF

ellatrix:
## What?

Fixes #54984.

Whenever a default block cannot be inserted on Enter, select the next block.

The next block means:
* The first child block, if any.
* If no child block, the next sibling block.
* If no next sibling block, look up the tree for a next sibling block of any parent.

## Why?

Enter does not do anything useful when a template is locked and no default block can be inserter after the block.

## How?

This ended up being a slightly larger refactor.

* We need to intercept Enter keydown in writing flow, right AFTER the the behaviour that checks if a block can be split.
* The problem is that the rich text fallback to split on Enter when "preserve line breaks" is on was running on even AFTER that. It must run before the Enter behaviour to select the next block, so we can disable that behaviour if necessary. (We don't want Enter in a verse block for example to select the next block.)
* Additionally, because of moving the rich text line break behaviour away from a window listener, I had to also move the listener for the deprecated multi line rich text from a React listener to a native DOM listener.

In the Site Editor, select the post title. Enter should select the next block, and so on. Generally anytime a block is selected and no default block can be inserted, Enter should select the next block.

### DIFF
--- a/packages/block-editor/src/components/rich-text/multiline.js
+++ b/packages/block-editor/src/components/rich-text/multiline.js
@@ -1,11 +1,12 @@
 /**
  * WordPress dependencies
  */
-import { forwardRef } from '@wordpress/element';
+import { forwardRef, useInsertionEffect, useRef } from '@wordpress/element';
 import deprecated from '@wordpress/deprecated';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { ENTER } from '@wordpress/keycodes';
 import { create, split, toHTMLString } from '@wordpress/rich-text';
+import { useRefEffect } from '@wordpress/compose';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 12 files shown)


jasmussen (approved):
Could use a code check, but this is a big win for the writing flow. Here's trunk, in the case of a Page template that has title, featured image, content, Enter does nothing:

trunk

In this PR, pressing Enter selects the next block, until you edit content inside the Content block:

after

Nice work!

alexstine:
@ellatrix 

> To do: see if we can remove this behaviour to keep the block wrapper selected.

This was added, I think by me, as an accessibility enhancement. Please provide clear testing instructions so someone can check the new flow. Currently, I am unsure how to test this.

Thanks.

jasmussen:
Here's how I tested this. First ensure your "Pages" template has a Title and/or Featured image block after, then a Content block:

* Go the site editor. 
* Navigate to Templates.
* Select the "Pages" template. Edit this to check for the presence of Title, Featured image, and Content. Insert if not.

Then:

* Go to site editor.
* Navigate to Pages.
* Select a page that has a title, featured image, and content.
* Focus the title. Press _Enter_.

Focus should then be on the _next block in the sequence_, i.e. if the title block is followed by a featured image, the featured image block should be focused.

The purpose is to improve the base writing flow when writing/editing pages in the full-template preview of the site editor.

packages/block-editor/src/components/rich-text/multiline.js
```diff
@@ -15,6 +16,115 @@ import { store as blockEditorStore } from '../../store';
 import { useBlockEditContext } from '../block-edit';
 import { getMultilineTag } from './utils';
 
+function useEnterRef( props ) {
```
ellatrix, line 19: Nothing in this whole file has changed, except moving from an `onKeyDown` React listener to a native DOM one, so that we run it before the default rich text Enter handler.

ellatrix: It's worth noting that we were actually not using the window listener until a couple of months ago, so it reverts some logic of #54543.

ellatrix:
> This was added, I think by me, as an accessibility enhancement.

I meant to avoid moving focus when calling `selectBlock()`. I added `true` as an initial position value to still set focus, but not move it to the start or end of the block.

Mamaduka:
Noticed a small regression caused by this change. Here's the fix: #82424.

PR #63486, opened 2024-07-12
Rename Button describedBy prop to description and deprecate old name.

### LINKED ISSUE #63483: Button: rename and deprecate describedBy prop
### Description

The button `describedBy` prop name is misleading.

It is inspired by the ARIA property `aria-describedby` which suggests it would accept an ID reference. Instead, it accepts a `string`, which is the actual description text.

It actually renders a visually hidden description, correctly associated to the button via `aria-describedby`. As such, a more appropriate, less confusing, prop name would be `description`.

To my understanding, other props have already been deprecated and `aliased' to new names so this wouldn't be unprecedented.

- Go to the storybook and select a Button e.g. https://wordpress.github.io/gutenberg/?path=/story/components-button--secondary
- See the `describedBy` is a `string`.
- Alternatively, see the component [types](https://github.com/WordPress/gutenberg/blob/4777ebd83df6466c345023407eeb9a4385441778/packages/components/src/button/types.ts#L45-L48).

_No response_

### Environment info

_No response_

Yes

Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63483

## What?

The Button `describedBy` prop name is misleading.

## Why?

Prop names should make clear what they are about and what their value type is.

## How?

- Renames `describedBy` to `description` and deprecates old name.
- Updates a few occurrences of the old prop name.
- Updates the prop description to make clearer it renders a _visually hidden_ description.

## Testing Instructions

- Test a couple Buttons with descriptions e.g.:
- Select a paragraph.
- Inspect the block switcher button in the block toolbar.
- Observe the button has an aria-describedby attribute pointing to an element with content `Change block type or style`.
- Select a List block.
- Inspect the Outend button in the block toolbar.
- Observe the button has an aria-describedby attribute pointing to an element with content `Outdent list item`.
- Observe there are no errors in the console.

### DIFF
--- a/packages/block-editor/src/components/alignment-control/ui.js
+++ b/packages/block-editor/src/components/alignment-control/ui.js
@@ -32,7 +32,7 @@ function AlignmentUI( {
 	onChange,
 	alignmentControls = DEFAULT_ALIGNMENT_CONTROLS,
 	label = __( 'Align text' ),
-	describedBy = __( 'Change text alignment' ),
+	description = __( 'Change text alignment' ),
 	isCollapsed = true,
 	isToolbar,
 } ) {
@@ -56,7 +56,7 @@ function AlignmentUI( {
 		? { isCollapsed }
 		: {
 				toggleProps: {
-					describedBy,
+					description,
 				},
 				popoverProps: POPOVER_PROPS,
 		  };
--- a/packages/block-editor/src/components/block-alignment-control/ui.js
+++ b/packages/block-editor/src/components/block-alignment-control/ui.js
@@ -64,7 +64,7 @@ function BlockAlignmentUI( {
 				} ),
 		  }
 		: {
-				toggleProps: { describedBy: __( 'Change alignment' ) },
+				toggleProps: { description: __( 'Change alignment' ) },
 				children: ( { onClose } ) => {
 					return (
 						<>
--- a/packages/block-editor/src/components/block-alignment-control/ui.native.js
+++ b/packages/block-editor/src/components/block-alignment-control/ui.native.js
@@ -77,7 +77,7 @@ function BlockAlignmentUI( {
 					};
 				} ),
 				popoverProps: POPOVER_PROPS,
-				toggleProps: { describedBy: __( 'Change alignment' ) },
+				toggleProps: { description: __( 'Change alignment' ) },
 		  };
 
 	return <UIComponent { ...commonProps } { ...extraProps } />;
--- a/packages/block-editor/src/components/block-switcher/index.js
+++ b/packages/block-editor/src/components/block-switcher/index.js
@@ -306,7 +306,7 @@ export const BlockSwitcher = ( { clientIds, disabled, isUsingBindings } ) => {
 							</>
 						}
 						toggleProps={ {
-							describedBy: blockSwitcherDescription,
+							description: blockSwitcherDescription,
 							...toggleProps,
 						} }
 						menuProps={ { orientation: 'both' } }
--- a/packages/block-editor/src/components/rich-text/format-toolbar/index.js
+++ b/packages/block-editor/src/components/rich-text/format-toolbar/index.js
@@ -52,7 +52,7 @@ const FormatToolbar = () => {
 											toggleProps.className,
 											{ 'is-pressed': hasActive }
 										),
-										describedBy: __(
+										description: __(
 											'Displays more block tools'
 										),
 									} }
--- a/packages/block-library/src/list-item/edit.js
+++ b/packages/block-library/src/list-item/edit.js
@@ -53,14 +53,14 @@ export function IndentUI( { clientId } ) {
 			<ToolbarButton
 				icon={ isRTL() ? formatOutdentRTL : formatOutdent }
 				title={ __( 'Outdent' ) }
-				describedBy={ __( 'Outdent list item' ) }
+				description={ __( 'Outdent list item' ) }
 				disabled={ ! canOutdent }
 				onClick={ () => outdentListItem() }
 			/>
 			<ToolbarButton
 				icon={ isRTL() ? formatIndentRTL : formatIndent }
 				title={ __( 'Indent' ) }
-				describedBy={ __( 'Indent list item' ) }
+				description={ __( 'Indent list item' ) }
 				disabled={ ! canIndent }
 				onClick={ () => indentListItem() }
 			/>
--- a/packages/block-library/src/list/edit.js
+++ b/packages/block-library/src/list/edit.js
@@ -109,8 +109,8 @@ function IndentUI( { clientId } ) {
 		<>
 			<ToolbarButton
 				icon={ isRTL() ? formatOutdentRTL : formatOutdent }
-				title={ __( 'Outdent' ) }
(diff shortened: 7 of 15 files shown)


packages/block-editor/src/components/block-alignment-control/ui.js
```diff
@@ -64,7 +64,7 @@ function BlockAlignmentUI( {
 				} ),
 		  }
 		: {
-				toggleProps: { describedBy: __( 'Change alignment' ) },
+				toggleProps: { description: __( 'Change alignment' ) },
```
tyxla, line 67: I see we've not updated the prop to the new one everywhere. Example:

https://github.com/WordPress/gutenberg/blob/0221522f253e094b277a1485b7a2d186cb172632/packages/block-editor/src/components/alignment-control/ui.js#L59

Is that intended?

tyxla (commented):
This solution makes sense to me, thanks @afercia 👍 

I'll let the rest of @WordPress/gutenberg-components review it as it involves some prop deprecation. 

I was just curious about which pre-existing instances of `describedBy` we choose to address.

packages/block-editor/src/components/block-alignment-control/ui.js
afercia: Glad to update more occurrences if desired. I think I only searched for occurrences of `\tdescribedBy=` and `\tdescribedBy:` 😆

tyxla: I think that will make sense since otherwise folks might get unexpected deprecation warnings in core.

afercia:
Thanks @tyxla I updated that occurrence and rebased.

packages/block-editor/src/components/alignment-control/ui.js
```diff
@@ -32,7 +32,7 @@ function AlignmentUI( {
 	onChange,
 	alignmentControls = DEFAULT_ALIGNMENT_CONTROLS,
 	label = __( 'Align text' ),
-	describedBy = __( 'Change text alignment' ),
```
ciampo, line 35: This is probably a breaking change for this component. The easy way to avoid the braking change would be to keep the `describedBy` prop, and internally pass it to the `description` prop in `toggleProps`.

Alternatively, we'd probably need to do something similar to what we've done in `Button`, where we support the old version alongside the new version.

afercia: @ciampo I understand the potential concern. However, I couldn't find any usage of the `describedBy` prop for the `AlignmentControl` component. The prop is not even documented in the [AlignmentControl readme](https://github.com/WordPress/gutenberg/blob/d1211ff7de85c66d45ad144f8c80323d875f964e/packages/block-editor/src/components/alignment-control/README.md). This component is used internally in the block-editor, is this change a real concern?
Rather, I'd say it's a good opportunity to _document_ it better. I discovered a few more issues and inconsistencies that I'll detail in a new comment.

afercia:
The [AlignmentControl](https://github.com/WordPress/gutenberg/blob/d1211ff7de85c66d45ad144f8c80323d875f964e/packages/block-editor/src/components/alignment-control/index.js) component of the block editor uses the [AlignmentUI](https://github.com/WordPress/gutenberg/blob/d1211ff7de85c66d45ad144f8c80323d875f964e/packages/block-editor/src/components/alignment-control/ui.js) component which accepts a few props, including `label` and `describedBy`.

I couldn't find any usage of the `describedBy` prop and only one usage of the `label` prop.

In the Table block, the alignment control [passes a label prop](https://github.com/WordPress/gutenberg/blob/d1211ff7de85c66d45ad144f8c80323d875f964e/packages/block-library/src/table/edit.js#L458) with value `Change column alignment`. That makes sense. Unfortunately, the `describedBy` is still the default one: `Change text alignment`. Honestly, it's sad to see how the development process in this project often misses to check and test anything that is 'non visual'. The consequence is that the alignment button in the Table block has:
- a label that says `Change column alignment`
- a description that says `Change text alignment`

Screen readers will announce both the label and description, thus providing a very confusing information to usdrs. Screenshots:

Screenshot 2024-07-17 at 09 36 48

Screenshot 2024-07-17 at 09 31 31

## Inconsistencies with similar components

There are two very similar components:

[BlockAlignmentUI](https://github.com/WordPress/gutenberg/blob/d1211ff7de85c66d45ad144f8c80323d875f964e/packages/block-editor/src/components/block-alignment-control/ui.js#L67)
Where both the label and description are hardcoded. The values are `Align / Change alignment`, which seems redundant.

[BlockVerticalAlignmentUI](https://github.com/WordPress/gutenberg/blob/d1211ff7de85c66d45ad144f8c80323d875f964e/packages/block-editor/src/components/block-vertical-alignment-control/ui.js)
Used for xample in the Media & Text block, where the label `Change vertical alignment` is hardcoded and there's no description.

I will try to split all the above in a new issue.

## Native files
I found one more occurrence in the `BlockAlignmentUI` native file. I can't test it and I'm not greatly interested in contributing to the native app but I guess it should be updated as well. Done in the last commit.

packages/block-editor/src/components/alignment-control/ui.js
ciampo: > This component is used internally in the block-editor

This change would affect the APIs of `AlignmentControl` and `AlignmentToolbar`, which are both public APIs of the `@wordpress/block-editor` package.

However, given that:

- as you mentioned, the property is not documented;
- as you mentioned, we couldn't find any current usages of the prop;
- any user using the old `describedBy` prop won't experience breaking, but only an unexpected description (which fallback to the default "Change text alignment")
- the change is an overall improvement and aligns with the changes in the `Button` component;

I agree with you that it's ok to just swap the prop name from `describedBy` to `describe`

ciampo:
> I will try to split all the above in a new issue.

Sounds good — you can probably open a new PR(s) directly with changes to better align label and description text.

> I found one more occurrence in the BlockAlignmentUI native file. I can't test it and I'm not greatly interested in contributing to the native app but I guess it should be updated as well. Done in the last commit.

cc @WordPress/native-mobile (I will also tag you in a code comment to pinpoint the exact change)

packages/block-editor/src/components/block-alignment-control/ui.native.js
```diff
@@ -77,7 +77,7 @@ function BlockAlignmentUI( {
 					};
 				} ),
 				popoverProps: POPOVER_PROPS,
-				toggleProps: { describedBy: __( 'Change alignment' ) },
+				toggleProps: { description: __( 'Change alignment' ) },
```
ciampo, line 80: @WordPress/native-mobile and @dcalhoun , just a quick check that this change makes sense and doesn't introduce unintended consequences 🙏

afercia: @ciampo  Thanks for coordinating

geriux: This change looks good to me 👍

afercia:
Pending final review and approval 😄

ciampo (approved):
🚀  Thank you for working on this!

tyxla:
Thank you, @afercia 🚀

PR #63487, opened 2024-07-12
Update: Data views grid layout: Make aspect ratio consumer configurable.

### LINKED ISSUE #60891: Data views grid layout: Make aspect ratio consumer configurable
Related: https://github.com/WordPress/gutenberg/issues/60630

Currently the preview field in grid layout renders in a hard-coded `1/1` aspect ratio, IE a square. As a default this works reasonably well, but some consumers would appreciate more expression. Pattern and Template previews for instance can be all kinds of lengths... squares don't always offer the best experience. By allowing custom (or `auto` to begin with) aspect ratios data views can make better use of space.

### Current patterns grid with square previews
Screenshot 2024-04-19 at 10 03 38

### Patterns grid with `auto` aspect ratio
 
Screenshot 2024-04-19 at 10 01 58

## User preference
Down the road it might be nice to allow users to customise this display option. Such an affordance would be particularly useful in pages where the preview dimensions can vary significantly, like Media.

 
aspect

- [ ] Enable data views to set `aspect-ratio: auto` for preview fields.
- [ ] Add toggle so that users can swap between `1/1` and `auto`.

jorgefilipecosta:
Fixes: https://github.com/WordPress/gutenberg/issues/60891

Adds the possibility for the user to configure the aspect ratio on grid items between auto and 1/1 (the default).

cc: @jameskoster 

## Screenshots
Screenshot 2024-07-12 at 15 21 10
Screenshot 2024-07-12 at 15 20 49

## Testing Instructions
Verified I was able to change the aspect ratio successfully.

### DIFF
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -32,6 +32,7 @@ import type {
 	SupportedLayouts,
 } from './types';
 import type { SetSelection, SelectionOrUpdater } from './private-types';
+import AspectRatioControl from './aspect-ratio-control';
 
 type ItemWithId = { id: string };
 
@@ -134,6 +135,12 @@ export default function DataViews< Item >( {
 						setOpenedFilter={ setOpenedFilter }
 					/>
 				</HStack>
+				{ view.type === LAYOUT_GRID && (
+					<AspectRatioControl
+						view={ view }
+						onChangeView={ onChangeView }
+					/>
+				) }
 				{ [ LAYOUT_TABLE, LAYOUT_GRID ].includes( view.type ) &&
 					hasPossibleBulkAction && (
 						<BulkActions
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -332,6 +332,11 @@ export interface ViewGrid extends ViewBase {
 		 * The fields to use as badge fields.
 		 */
 		badgeFields?: string[];
+
+		/**
+		 * True if the grid has a 1:1 aspect ratio the default, and false if it doesn't.
+		 */
+		hasOneOneAspectRatio?: boolean;
 	};
 }
 
--- a/packages/dataviews/src/aspect-ratio-control.tsx
+++ b/packages/dataviews/src/aspect-ratio-control.tsx
@@ -1,0 +1,52 @@
+/**
+ * WordPress dependencies
+ */
+import { Button } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { fullscreen } from '@wordpress/icons';
+
+/**
+ * Internal dependencies
+ */
+import type { ViewGrid } from './types';
+
+export default function AspectRatioControl( {
+	view,
+	onChangeView,
+}: {
+	view: ViewGrid;
+	onChangeView: ( view: ViewGrid ) => void;
+} ) {
+	const hasOneAspectRatio = view.layout?.hasOneOneAspectRatio !== false;
+	return (
+		<>
+			<Button
+				size="compact"
+				isPressed={ hasOneAspectRatio }
+				icon={ fullscreen }
+				label={
+					hasOneAspectRatio
+						? __( 'Change aspect ratio to auto' )
+						: __( 'Change aspect ratio to 1:1' )
+				}
+				onClick={ () => {
+					if ( hasOneAspectRatio ) {
+						onChangeView( {
+							...view,
+							layout: {
+								...view.layout,
+								hasOneOneAspectRatio: false,
+							},
+						} );
+					} else if ( view.layout ) {
+						const { hasOneOneAspectRatio, ...layout } = view.layout;
+						onChangeView( {
+							...view,
+							layout,
+						} );
+					}
+				} }
+			/>
+		</>
+	);
+}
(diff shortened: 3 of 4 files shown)


packages/dataviews/src/types.ts
```diff
@@ -332,6 +332,11 @@ export interface ViewGrid extends ViewBase {
 		 * The fields to use as badge fields.
 		 */
 		badgeFields?: string[];
+
+		/**
+		 * True if the grid has a 1:1 aspect ratio the default, and false if it doesn't.
+		 */
+		hasOneOneAspectRatio?: boolean;
```
youknowriad, line 339: Should we instead have the possibility to actually pick an aspect ratio (like an enum instead of a boolean)?

packages/dataviews/src/dataviews.tsx
```diff
@@ -134,6 +135,12 @@ export default function DataViews< Item >( {
 						setOpenedFilter={ setOpenedFilter }
 					/>
 				</HStack>
+				{ view.type === LAYOUT_GRID && (
```
youknowriad, line 138: Instead of putting this in its own menu, I think we should put it in the ViewActions instead and maybe offer a few different aspect ratio choices?

jasmussen:
Not sure this is a great user experience. Without actual masonry layouts, which is an open question how well this can work for keyboard navigation, this is just going to create visual rivers in the text and cause things to not line up. It seems like if we want to allow changing the aspect ratio, it should change the aspect ratio of _all_ previews.

jameskoster:
This feature provides parity with the original Patterns page, which feedback suggests was an enjoyable preview experience compared with the cropping we see on trunk. 

pattern-site-editor-643

> it should change the aspect ratio of all previews

That should be what happens. Aspect ratio is either:

- `1:1`: handy if you want a perfect grid with everything lined up. 
- `auto`: useful when you want the preview to more accurately reflect the pattern/template. 

The latter comes at the expense of perfectly aligned text, which seems a reasonable trade-off to me. To perfectly align the text it would need to appear before the preview, which looks a little odd. Perhaps I'm missing an alternative approach? 

It should work nicely with https://github.com/WordPress/gutenberg/pull/63367, and I think we should probably concentrate on merging that one first so that we can better test this one.

ntsekouras:
Personally I'd echo Joen. This feels a bit weird and maybe too prominent? I'm also not sure about the feature parity argument, since the new list is now quite different with way more info. 

Could this be something baked in the preview like a scrollbar or some interaction on hover (example zoom or something)?

jameskoster:
Agree on the prominence. It could be an option we include in the view options menu instead:

aspect

I still think this is a situationally handy feature. The constrained square preview isn't a great way to browse templates, which are often quite 'tall'.

jasmussen:
The main problem is these horizontal rivers:

Screenshot 2024-07-16 at 09 35 13

When the text doesn't line up, it's hard to scan.

jameskoster:
I agree, but I don't think scanning text is always the main concern in this layout. If a user wants to browse that way it would make more sense to switch to Table layout, or to toggle the aspect ratio to something strict like 1:1.

As a primarily visual layout, Grid should be optimised for scanning the _Previews_, and in some cases the 1:1 cropping makes that experience less ergonomic.

Another example is the media library, where you might want to see the orientation of a media item:

aspect

jasmussen:
I appreciate your desire to improve this view, but without actual masonry, even scanning just the previews becomes hard, when the whitespace causes those rivers. Even if this is an optional view configuration, it does not seem useful in this state. How do other similar apps or interfaces handle this? How about horizontal masonry, where wide images push the next item in the row to the right?

richtabor:
> I appreciate your desire to improve this view, but without actual masonry, even scanning just the previews becomes hard, when the whitespace causes those rivers. 

Agreed. This is not ideal: 

CleanShot 2024-07-31 at 10 01 57

---

Not sure if it should be configurable by the user, but if we wanted a very minimal, non-1:1 previews, it could resemble many photos apps design: increased gutter spacing, no visible text. 

It's actually not bad for patterns (and media library of course): 


CleanShot 2024-07-31 at 10 04 47


As seen in MacOS Photos: 

CleanShot 2024-07-31 at 10 03 23

jameskoster:
That could be interesting to try (hiding the text, aligning each row along the x axis). 

I think user configuration makes sense, particularly for media where cropping is situationally useful.

jameskoster:
Now that view options live in a popover, here's an alternative design we might try; 'Preview display' with options 'Cropped' (square) and 'Uncropped':

Screenshot 2024-08-13 at 18 27 13

When uncropped all fields would be locked as hidden except for the title which could appear on hover, just like the checkbox.

patterns

ndiego:
@jameskoster and @jorgefilipecosta can you confirm that this will be punted to 6.8? I'm assuming this is the case since it's not on the 6.7 project board, but this issue was listed in the [6.7 Roadmap](https://make.wordpress.org/core/2024/09/03/roadmap-to-6-7/), so I wanted to double-check. Thanks!

jameskoster:
Yes let's punt this one, it's not high-prio until we get to the media library data view.

jorgefilipecosta:
Closing this PR as it is outdated if we still want to have user-configurable aspect ratio we will need to open a new one taking into account the new view configuration UI.

PR #63488, opened 2024-07-12
Templates DataViews: Set the right context for the preview field

### LINKED ISSUE #59082: Some template blocks in data view previews do not match their display in the editor

youknowriad:
Fixes #59082 

## What?

Some core blocks are WP specific and require the "EditorProvider" (the core/editor store) context to work properly and render the right information.

One of these blocks is the QueryTitle block and this is fixed in this PR.

## Testing Instructions

1- Create a "Tag Archive" template and use the Archive Title block in it. 
2- In the data views, the block should show the right placeholder.

### DIFF
--- a/packages/edit-site/src/components/page-templates/index.js
+++ b/packages/edit-site/src/components/page-templates/index.js
@@ -18,7 +18,10 @@ import {
 } from '@wordpress/block-editor';
 import { DataViews, filterSortAndPaginate } from '@wordpress/dataviews';
 import { privateApis as routerPrivateApis } from '@wordpress/router';
-import { privateApis as editorPrivateApis } from '@wordpress/editor';
+import {
+	privateApis as editorPrivateApis,
+	EditorProvider,
+} from '@wordpress/editor';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/components/page-templates/index.js
```diff
@@ -176,7 +177,7 @@ function Preview( { item, viewType } ) {
 	// the block editor settings are needed in context where we don't have the block editor.
 	// Explore how we can solve this in a better way.
 	return (
-		<ExperimentalBlockEditorProvider settings={ settings }>
+		<EditorProvider post={ item } settings={ settings }>
```
youknowriad, line 180: This Preview component should actually be moved to the "editor" package and be named something like `PostPreview` or something. The purpose of that is to preview a given postType + postId tuple.

ntsekouras (approved):
Looks good, thank you!

packages/edit-site/src/components/page-templates/index.js
ntsekouras: You mean by having a lighter component than the actual EditorProvider, right? Do you think this change will affect performance for now, until we implement that component?

youknowriad: No, I think the `EditorProvider` is fine. but instead of having all of this custom logic here in the site editor, this seems like an easy reusable component (preview of a post)

PR #63490, opened 2024-07-12
Align checkbox, radio, and toggle input design

jameskoster:
storybook

This PR is to scrutinize some design system explorations about better aligning `ToggleControl`, `RadioControl`, and `CheckboxControl`.

### Changes

* Align labels
  * Line-height matches input height to ensure alignment
  * `cursor: pointer` across
* Reduced footprint for Checkboxes and Radios on larger screens making them more proportionally aligned with the label
* Update toggle dimensions based on 4px baseline. 16px height to align with radio + checkbox.
* Apply the standard input border color to toggle off state
* Indent help text consistently between checkbox and toggle (aligned with label on Y axis)
* Consistent focus ring styling
* Consistent spacing between input and help text
* Consistent spacing between input and label

### To test

* `npm run storybook:dev`
* Check `RadioControl`, `CheckboxControl`, `ToggleControl`
* Smoke test the Editor

### Changes in context
inputs


jasmussen:
Dimensions and metrics look great. As we've talked about in chat, I'm personally unsure about the font weight change, but I'll let you reach a broader group when you're ready for that, and I'll defer to group consensus. But it would be useful to see these change in context, not just storybook, if you can include a couple of screenshots of the editor with these components and their weight shown there.

jameskoster:
Good point, I added some 'changes in context' screenshots.

The font-weight is a tricky one. Objectively judging the components in isolation I feel it's a good change because it viaully differentiates the label from regular body text, which seems like an a11y enhancements. 

However in some circumstances I see how it can feel a bit overweight. But that might suggest there are other styles in the type scale to address, rather than the component styles being 'wrong'. Keen to gather more feedback around that.

mirka:
How much do we take into consideration non-Latin scripts when making legibility decisions?

This is what the proposed RadioControl would look like with Chinese/Japanese characters, for example:

| Before | After |
|--------|-------|
|Current RadioControl weights with Japanese language|New RadioControl weights with Japanese language|

Aside from the heading label which is already hard to read even on a high-res screen with quality fonts, I feel like this adds even more legibility issues by adding extra weight to small text. Also note that all caps styling will not add any visible contrast in languages that do not have a lower/upper case distinction. (@t-hamano Does the legibility issue ever come up in your circles?)

In my experience it is hard to achieve a single style that works in all languages, and it is possible to set slightly different styles based on the `lang` attribute, if all else fails.

t-hamano:
> Does the legibility issue ever come up in your circles?

I have never heard anyone around me suggest that the readability of the text is an issue in the UI of the entire editor, including the `RadioControl` and `CheckboxControl` components.

Another issue here is that, as [this CSS variable indicates](https://github.com/WordPress/gutenberg/blob/c2a8564f6f8dda43d6931ca445ebfdc5352d77ca/packages/base-styles/_variables.scss#L15), the font depends on the OS and the fonts installed on your machine, so readability may vary depending on your environment.

For example, on my Windows machine, English text looks like this. Personally, I feel the `font-weight:500` is a little heavy.

| trunk | this PR |
|--------|--------|
| en_before |  en_after | 

Strangely, `font-weight:500` is not applied to Japanese. This is probably because I don't have a local font installed on my machine that supports weight 500:

| trunk | this PR |
|--------|--------|
| ja_before | ja_after |

tyxla:
Just my two cents: I find the smaller controls rather a regression than an improvement.

jameskoster:
Thanks for the feedback y'all!

> Aside from the heading label which is already hard to read even on a high-res screen with quality fonts

That would suggest an issue with the all-caps label style, no?

I still think there's value in differentiating labels from body text, but acknowledge it's a tricky detail to get right and probably one to explore independently. I reset the weight change in this PR and updated the OP.

mirka:
I went ahead and pushed a couple tweaks, if that's ok. They're mostly just architectural fixes, which shouldn't change the resulting visual. The only visual changes I added are the `pointer` cursor styles — these should now exclude disabled states, and also be applied in a non-WP context.

packages/base-styles/_variables.scss
```diff
@@ -93,8 +93,12 @@ $border-width-tab: 1.5px;
 $helptext-font-size: 12px;
 $radius-round: 50%;
 $radius-block-ui: 2px;
-$radio-input-size: 20px;
+$radio-input-size: 16px;
 $radio-input-size-sm: 24px; // Width & height for small viewports.
+$toggle-width: $grid-unit-40;
+$toggle-height: $grid-unit-20;
+$toggle-border-width: 1px;
+$toggle-thumb-size:  $grid-unit-15;
```
mirka: Is it necessary to lift all these variables up into `base-styles`? I'm not sure how conservative we are about adding global variables, but personally I wouldn't add them if it isn't necessary to share them across packages.

If it's just to share values between the FormToggle and ToggleControl components, we can keep them within the `@wordpress/components` package.

ciampo: Agreed. On a separate note, at some point we should look at tidying up shared styles in the package: some variables are defined in Sass, some as CSS custom properties, and some in JS.

jameskoster:
>  if that's ok

Absolutely!

packages/base-styles/_variables.scss
jameskoster: > If it's just to share values between the FormToggle and ToggleControl components

That was it. Your suggestion makes sense, though I'm not sure how to do it.

mirka: Moved back to the FormToggle styles in 0d73cc5. (Ideally this should use a `@use` to scope the variables better, but we're not set up to do that properly right now.)

mirka (approved):
Looks good from my end, let's add a changelog before merging 👍

jameskoster:
@WordPress/gutenberg-design @richtabor @jarekmorawski would appreciate your feedback on the design details before merging 🙏

jameskoster:
@mirka or @ciampo do you know why the tests are failing? 😭

ciampo:
> @mirka or @ciampo do you know why the tests are failing? 😭

Seems like a hiccup in the CI tests — I restarted the tests and they are all passing

jameskoster:
I enabled auto-merge. We're very early in the release cycle so there's plenty of time to gather additional design feedback and refine as required.

richtabor:
My first thought is why not align the help text for radio as well? Looking at the screenshot up top.

jameskoster:
@richtabor I had the same thought, but when I tried it the help text appeared to be attached to the last radio in the group. There's an issue here about moving that help text above the group which could help: https://github.com/WordPress/gutenberg/issues/63734

afercia:
@jameskoster regarding the indentation of the help text, I just noticed that when the help text is long and is multiline, only fhe first line is indented. Is this the expected layout? It doesn't match with the screenshots provided in this PR. Not sure it looks great. At least for the toggles. I haven't checked the checkboxes and radio buttons. Example from the Query Loop inspector:

Screenshot 2024-07-24 at 12 39 08

The left margin is applied to a `<span>` element inside the paragraph of the help text. The span is an inline element thus the margin works only on the first line.

Instead of [adding a span element](https://github.com/WordPress/gutenberg/commit/fb1f9a9295c573dac04884d3e811ffb3b660caa8#diff-e65272ba98933cec9452ce6a201905ce89f6b0a19693bfd2a51e08d96197bf7bR93-R95), wouldn't be better to just target the existing paragraph, which is a children of the toggle control? 
`.components-toggle-control .components-base-control__help`

Re: the process: if this PR had had an associated issue [as required by the contributing guidelines of this project](https://github.com/WordPress/gutenberg/blob/0b94a19ca3307f98fc7124573e8be51cbe32cfd6/docs/contributors/repository-management.md#pull-requests), I woud have  just reopened the issue. That's how the process works on Trac for Core and I'd love to see the same process used in the Gutenberg GitHub. One more reason to always create issues before creating PRs. Cc @youknowriad as we discussed this point at WCEU.

jameskoster:
Thanks for pointing that out, definitely not intended. I'll work on a fix.

Issue #43423, opened 2022-08-19
Implement a standard "controls grid" component for sidebar


mirka:
## What problem does this address?

The controls in the sidebar should be laid out in a consistent grid, whether they are in `edit-post` (ToolsPanel) or `edit-site` (ad hoc layouts). Currently they each do their own styling for `rowGap`/`columnGap`, column count, and full-width items.

## What is your proposed solution?

Try making a standard layout component for this purpose, so they can be easily reused in either context. We can call it something like `ControlsGrid`.

ciampo:
This could be an interesting exploration to inform some of the decision re. grid components discussed in https://github.com/WordPress/gutenberg/issues/43085

jasmussen:
Here's a suggested grid: 

<img width="1026" alt="Grid" src="https://user-images.githubusercontent.com/1204802/187613848-7b278d22-3332-442b-8a5e-232e0b84002c.png">

Is that helpful?

jasmussen:
I went ahead and created a separate issue for updating the default component sizes: #46734

mirka:
Relevant: An implementation of the boundary-breaking buttons in the header labels #50855

afercia:
Comine here from https://github.com/WordPress/gutenberg/pull/63806
Although this is somehow an old issue, I'd like to add a couple considerations:

Personally, I'm not sure the inconsistency in the placement of the 'Set custom size' toggle button helps with providing the highest level of usability and accessibility. Currently, this button is placed in a different spot depending on the control. Example screenshot:

![Screenshot 2024-07-25 at 08 37 49](https://github.com/user-attachments/assets/561bf2a1-8efa-4c37-81c4-38030580605f)

Sometimes the order is:
1. Label
2. Toggle
3. Control

Sometimes is:
1. Label
2. Control
3. Toggle

From a logical order perspective, I would expect this order:
1. Label: it tells me what the control is about and it makes sense to have this information first.
2. The control itself.
3. Lastly, the option to switch the control to an alternative version.

What I would like to see from the design team is an effort to make design that goes beyond providing a balanced, visually pleasant, grid. I would like to see a _functional_ analysis and consistency in the grouping and order of the controls.

Issue #39358, opened 2022-03-10
Components: Deprecate outer margins


mirka:
## What problem does this address?

A lot of components currently have inherent outer margins, making it hard to reuse in different contexts since it requires a style override (often of an internal element). This kind of style override is something we want to discourage.

## What is your proposed solution?

Deprecate the bottom margin (similar to https://github.com/WordPress/gutenberg/pull/37160), following the [guidelines](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/CONTRIBUTING.md?rgh-link-date=2022-02-11T00%3A43%3A54Z#deprecating-styles) stipulated in the devdocs.

## Components

Each component first needs a `__nextHasNoMarginBottom` prop to opt in, and then they also need to be officially deprecated after all in-repo usage has been migrated.

- AnglePickerControl
  - [x] prop added #43160
  - [x] deprecated #43867
  - [x] hard deprecated #58700
- BaseControl 
  - [x] prop added #39325
  - [x] deprecated #64408
  - Components built with BaseControl tracked in #38730
- CustomGradientPicker
  - [x] prop added #43387
  - [x] deprecated #43436
  - [x] hard deprecated #58699
- DuotonePicker (when `duotonePalette={ [] }` and `unsetable={ false }`)
  - [x] fix as bug #57489
- FontSizePicker #38720
  - [x] prop added #43062
  - [x] deprecated #43870
  - [x] hard deprecated #58702
- GradientPicker
  - [x] prop added #43387
  - [x] deprecated #43436
  - [x] hard deprecated #58701
- FormTokenField (encountered [here](https://github.com/WordPress/gutenberg/pull/42629/files/fa8cffd217c03ac48ec77f529a4a4bab0f0b8e86#r935442913))
  - [x] prop added #48609
  - [x] deprecated #63491
  - [x] hard deprecated #73846

## Block Editor Components

- [x] #37160
- [x] #46692
- [x] #73340

(Component lists may not be exhaustive, please add)

---

### Bottom margin styles are deprecated

A number of UI components currently ship with styles that give them bottom margins. This can make it hard to reuse them in arbitrary layouts, where you want different amounts of gap or margin between components.

To better suit modern layout needs, we have gradually been deprecating these margins. A deprecation begins with an opt-in period where you can choose to apply the new margin-free styles on a given component instance. Eventually in a future version, the margins will be completely removed.

[...]

PR #63491, opened 2024-07-12
FormTokenField: Deprecate bottom margin

mirka:
Part of #39358

## What?

Officially start the deprecation period for the bottom margin on `FormTokenField`. And by extension, also on `PostTaxonomiesFlatTermSelector` in `@wordpress/editor`.

## Why?

All the unwanted margins on non-BaseControl-based components have already been hard deprecated, but this one was straggling.

## How?

The dependants of FormTokenField in this repo are the following:

- FormTokenField
  - Query Loop block
  - `wp.editor.PostTaxonomiesFlatTermSelector` (publicly exported as component)
    - `wp.editor.PostTaxonomies` (publicly exported as component)

Out of these, changing FormTokenField margins would only affect third-party consumers in the case of  `PostTaxonomiesFlatTermSelector`. So for that component, we are also doing a deprecation.

### Query Loop block

You may need this diff to show all the "Filters" options in the block inspector sidebar:

```diff
diff --git a/packages/block-library/src/query/edit/inspector-controls/index.js b/packages/block-library/src/query/edit/inspector-controls/index.js
index e71b2d09af..0c4471f187 100644
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -268,7 +268,7 @@ export default function QueryInspectorControls( props ) {
 							/>
 						</ToolsPanelItem>
 					) }
-					{ showParentControl && (
+					{ true && (
 						<ToolsPanelItem
 							hasValue={ () => !! parents?.length }
 							label={ __( 'Parents' ) }
```

| Before | After |
|--------|-------|
|Query Loop block inspector controls, before|Query Loop block inspector controls, after|

The gutters in between used to be uneven and wrong. It is now uniform and [correct](https://github.com/WordPress/gutenberg/issues/43423#issuecomment-1232540038) (16px).

### PostTaxonomiesFlatTermSelector

In Post sidebar:

| Before | After |
|--------|-------|
|FlatTermSelector with extra margin|FlatTermSelector with no margin|

There used to be an extra bottom margin. Now it is gone.

---

## ✍️ Dev Note

See Dev Note in #39358.


packages/interface/src/components/complementary-area/style.scss
```diff
@@ -33,7 +33,7 @@
 		}
 	}
 
-	p:not(.components-base-control__help) {
+	p:not(.components-base-control__help, .components-form-token-field__help) {
 		margin-top: 0;
 	}
```
mirka, line 38: This is to maintain this margin right here, which was previously maintained implicitly due to a `Spacer` component inserted in the `__nextHasNoMarginBottom={ false }` case.

Margin between FormTokenField input field and help text, in sidebar

packages/editor/src/components/post-taxonomies/index.js
mirka, line 1: The `PostTaxonomies` component is used in two places:

- The "Tags" section of the Post sidebar.

    Tags section

- In the Preferences modal.
 
    Toggles in the Preferences modal

https://github.com/WordPress/gutenberg/blob/9f8014ad8aea4ed4602dca039c1c89624a01c330/packages/editor/src/components/preferences-modal/index.js#L105-L111

The usage in the Preferences modal is very strange — the callback parameters are used to render toggles 😅 Anyway, our margin changes don't affect this use case.

packages/editor/src/components/post-publish-panel/maybe-tags-panel.js
mirka, line 1: This is in the Post publish panel:

Tags input in Post publish panel

tyxla:
> You may need this diff to show all the "Filters" options in the block inspector sidebar:

No need to do that. Just make sure to test with both a hierarchical post type (page) and a flat one (post). You can select it from the Query Loop block settings:

Screenshot 2024-07-15 at 14 40 21

packages/editor/src/components/post-taxonomies/flat-term-selector.js
```diff
@@ -55,16 +56,28 @@ const termNamesToIds = ( names, terms ) => {
...
+export function FlatTermSelector( { slug, __nextHasNoMarginBottom } ) {
 	const [ values, setValues ] = useState( [] );
 	const [ search, setSearch ] = useState( '' );
 	const debouncedSearch = useDebounce( setSearch, 500 );
 
+	if ( ! __nextHasNoMarginBottom ) {
+		deprecated(
+			'Bottom margin styles for wp.editor.PostTaxonomiesFlatTermSelector',
+			{
+				since: '6.7',
```
tyxla, line 74: Is this going to land in 6.7 actually?

packages/block-library/src/query/editor.scss
```diff
@@ -49,15 +49,6 @@
 	}
 }
 
-.block-library-query-toolspanel__filters {
```
tyxla, line 52: We're still rendering the original classname for backward compatibility, right?

packages/components/src/form-token-field/stories/index.story.tsx
```diff
@@ -99,6 +100,7 @@ export const Async: StoryFn< typeof FormTokenField > = ( {
 Async.args = {
 	label: 'Type a continent',
 	suggestions: continents,
+	__nextHasNoMarginBottom: true,
```
tyxla, line 103: Confirmed that all stories inherit args from either of the 2 definitions that define `__nextHasNoMarginBottom: true`.

tyxla (approved):
I spent some time testing thoroughly in Storybook and in the post/site editor, everything tests well. ✅ 

I also took some time to confirm all instances in the codebase have been addressed. 👍 

LGTM 🚀

packages/editor/src/components/post-taxonomies/flat-term-selector.js
mirka: Yes, too late to land in 6.6.

packages/block-library/src/query/editor.scss
mirka: Correct. I added a code comment b912e8b

Issue #63327, opened 2024-07-10
Site Logo: Improve the initial state


alaczek:
## What problem does this address?
The initial state of the Site Logo block is not clear enough about how to upload the image. (Additionally, the tooltip can be obscured by the floating toolbar). 

<img width="592" alt="image" src="https://github.com/WordPress/gutenberg/assets/4550351/5690fed7-57f1-463d-b4df-655673996833">

There is a button to "Add media" too, but it's tucked away in the block settings in the sidebar, which may or may not be open.

Compare this to the initial state of the Image block that clearly lists the ways an image can be added:

<img width="640" alt="image" src="https://github.com/WordPress/gutenberg/assets/4550351/9883c76e-a1bc-45a5-9018-03e628836689">

## What is your proposed solution?
I would like to propose aligning the initial state of the Site Logo block with the Image block. or otherwise making it more obvious how to upload an image (maybe by adding an "Add media" action to Site Logo block toolbar).

Mamaduka:
cc @jasmussen, since we've worked on MediaPlaceholder design(s) before :)

jasmussen:
This one is very much related to #63254 in more ways than one. But the most important one is: the block _has to be able to work when it's that small_, because it's a valid use case to use the site logo in that small space. 

Adding "add media" to the inspector and adjusting the tooltip feel like good ways to progress, and I'd echo the Image block can be adjusted to match this too.

mtias:
The "replace" workflow once something is set is very clear, so I think the placeholder state should inherit that and just have an "add image" in the toolbar

jasmussen:
[This mockup](https://github.com/WordPress/gutenberg/issues/63382#issuecomment-2222699236) makes progress on suggestions from the conversation here.

PR #63498, opened 2024-07-12
Update MediaUpload button for site logo from "Add media" to "Choose logo"

richtabor:
## What?
Makes the button much clearer what you are intended to upload. Part of https://github.com/WordPress/gutenberg/issues/63327. 

I went with "Choose Site Logo" to map to the Site Icon setting in the General Settings admin view. 

Other alternatives: 
- I did consider "Add Site Logo", but it is a bit redundant as the Site Logo block itself was already added to the page. 
- "Add logo" could work, but since we're specific about "Site Logo" as the block name I was thinking it would feel more contextual to use "Site Logo" in the control. 
- "Select Site Logo" is another alternative; relative to "Select Site Icon" from the Customizer setting. 

## Testing Instructions

1. Open a post or page.
2. Insert the Site Logo block. 
3. Reset the logo if there is an existing one uploaded already. 
4. See changes in the Inspector. 

## Screenshots or screencast 

| Before  | After |
|CleanShot 2024-07-12 at 14 27 20|CleanShot 2024-07-22 at 16 00 41|

### DIFF
--- a/packages/block-library/src/site-logo/edit.js
+++ b/packages/block-library/src/site-logo/edit.js
@@ -631,7 +631,7 @@ export default function LogoEdit( {
 											{ isLoading ? (
 												<Spinner />
 											) : (
-												__( 'Add media' )
+												__( 'Add Site Logo' )
 											) }
 										</Button>
 										<DropZone onFilesDrop={ onFilesDrop } />
--- a/packages/block-library/src/site-logo/edit.js
+++ b/packages/block-library/src/site-logo/edit.js
@@ -631,7 +631,7 @@ export default function LogoEdit( {
 											{ isLoading ? (
 												<Spinner />
 											) : (
-												__( 'Add Site Logo' )
+												__( 'Choose Site Logo' )
 											) }
 										</Button>
 										<DropZone onFilesDrop={ onFilesDrop } />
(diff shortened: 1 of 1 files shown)


t-hamano (commented):
What do you think about following the General Settings and changing it to "Choose a Site Icon"?

image

jasmussen:
> What do you think about following the General Settings and changing it to "Choose a Site Icon"?

I'd still suggest "Add logo". While both are active language, the former is shorter and more clear. Shorter is a benefit for the block toolbar, which especially when set to text labels only can get very wide.

richtabor:
> What do you think about following the General Settings and changing it to "Choose a Site Icon"?

Site Icon is different than Site Logo.

t-hamano:
> Site Icon is different than Site Logo.

I see. So, the possible patterns are either of the following? Personally, I think any of the texts in this list make sense. None of the patterns use the word "icon."

- Add site logo
- Add logo
- Choose site logo
- Choose logo
- Select site logo
- Select logo

Also, wouldn't sentence case (Add site logo) be more appropriate here rather than title case (Add Site Logo)?

richtabor:
Went with "Choose logo", and also updated the tooltip to not conflict with the block toolbar. 

CleanShot 2024-07-22 at 16 00 37

t-hamano (approved):
LGTM!

Issue #58198, opened 2024-01-24
Make 'Add' buttons consistent across data views, simplify labels


jameskoster:
<img width="1612" alt="add" src="https://github.com/WordPress/gutenberg/assets/846565/b954a998-88b6-487c-b1f1-9927bb7b3388">

- [ ] "Add **new** $post_type" is unnecessarily verbose. We can shorten this to "Add $post_type". Trac ticket: https://core.trac.wordpress.org/ticket/61219#ticket
- [x] The add flow for Patterns should match other data views. IE a primary button in the page header. For now this button will open the same menu as the current `+` in the sidebar. The `+` button can be removed.

afercia:
> The labels are unnecessarily long. "Add new" would suffice in all cases rather than the verbose "Add New Template Part".

You may not be aware that recently in core all the 'Add new' strings have been expanded to include the _what_ they refer to. This has been done for accessibility reasons _and_ to be consistent with Gutenberg.

I'm not sure this should be changed. Please don't shorten to 'Add new'.

Cc @joedolson @alexsanford

jameskoster:
@afercia couldn't we use aria labels like the action links in wp-admin tables?

<img width="744" alt="Screenshot 2024-01-24 at 16 06 19" src="https://github.com/WordPress/gutenberg/assets/846565/00698192-20f2-4bd3-a318-d9a35b1a4364">

joedolson:
I'd really like to avoid using different logic in new interfaces from what we've decided to use in core. In core, we've agreed that these controls need to include context, so let's not just casually reverse that in a new interface. 

Regarding the row actions, those use aria-label only because of the amount of repetition and the possible length. There are several row actions, and each one needs that context included, and the post titles could be quite a bit longer than a single word.

"Add New Template Part" is a nominal amount of text.

In row actions, we'd potentially be looking at `Edit "You only live once: the history of frisbees in Wollongong" | Quick Edit "You only live once: the history of frisbees in Wollongong"`, etc. You get the point. Visually, that starts to take up an enormous amount of space and repetition.

Screen reader users don't necessarily have to experience that entire text; spoken text can be interrupted, unlike visual text. So a screen reader navigating through these same links only needs to hear enough to verify that the context is what they expected it to be. So that quantity of text is less of an interference for screen reader users than it is for sighted users.

jameskoster:
[...]

PR #63499, opened 2024-07-12
Made SiteLogoReplaceFlow always available in the Site Logo block toolbar

richtabor:
Makes the SiteLogoReplaceFlow always available in the toolbar, to render "Choose Site Logo" when there is no logo, and "Replace" when there is one.  Part of https://github.com/WordPress/gutenberg/issues/63327. 

The "Choose Site Logo" string should map to the same as what we decide for the Inspector, as I explored in https://github.com/WordPress/gutenberg/pull/63498. 

## Why?
It should be clearer how to add a site logo and using an existing affordance that we apply elsewhere is much more supportive of the user experience. 

## Testing Instructions
1. Open a post or page.
2. Insert the Site Logo block.
3. Reset the logo if there is an existing one uploaded already.
3. See "Choose Site Logo" in the toolbar. 
4. Choose an image to use as your logo. 
5. See "Replace" in the toolbar. 
6. Select "Replace" and see a third menu option to "Reset". 

## Screenshots or screencast 

| Before  | After |
|CleanShot 2024-07-12 at 15 05 17|CleanShot 2024-07-22 at 15 46 00|

---

Added by @t-hamano 

The `SiteLogoReplaceFlow` component is a wrapper component for `MediaReplaceFlow`, and the Reset button is rendered as a child of `MediaReplaceFlow`. When the Reset button is pressed, it is expected that the popover will be closed and the focus will return to the block toolbar button, but currently `MediaReplaceFlow` cannot call `onClose` from the child.

Therefore, in this PR, we will update `MediaReplaceFlow` to support "function as a child". function as a child can receive `onClose` as a prop.


packages/block-library/src/site-logo/edit.js
```diff
@@ -517,11 +523,12 @@ export default function LogoEdit( {
 
 	const mediaReplaceFlowProps = {
 		mediaURL: logoUrl,
+		name: ! logoUrl ? __( 'Choose Site Logo' ) : __( 'Replace' ),
```
mtias: Should this be title cased?

richtabor: I was drawing from the "Site Icon" terminology in WP admin, but this could just be "Choose logo". 

CleanShot 2024-07-12 at 16 16 39

packages/block-library/src/site-logo/edit.js
```diff
@@ -517,11 +523,12 @@ export default function LogoEdit( {
...
 		onSelect: onSelectLogo,
 		onError: onUploadError,
 		onRemoveLogo,
 	};
-	const controls = canUserEdit && logoUrl && ! temporaryURL && (
+	const controls = (
```
t-hamano:
```suggestion
	const controls = canUserEdit && (
```

I think it's necessary to keep canUserEdit, otherwise users who can't reset/update the site icon will see the toolbar menu unintentionally when they insert this block.

Mamaduka: Yes, we need to keep capability checks in place.

packages/block-library/src/site-logo/edit.js
jasmussen: I'd personally lean super strongly towards "Choose logo", with lowercase "logo" since you're choosing a media item to be your logo, you're not choosing a particular item type from your library.

But! Happy to defer to consensus otherwise.

jameskoster: Should we extract and discuss this more broadly? I also prefer sentence case, but committing to that will mean diverging from wp-admin. Sometimes this creates extra effort as we found in https://github.com/WordPress/gutenberg/issues/58198.

It would be good to be consistent across all areas.

packages/block-library/src/site-logo/edit.js
richtabor: good catch, thanks.

packages/block-library/src/site-logo/edit.js
richtabor: Updated it to "Choose logo"

t-hamano (commented):
One thing that bothers me a bit is that when I press the reset button only the reset button disappears but the popover remains open. 

https://github.com/user-attachments/assets/5314ace0-c56c-4a50-9965-f1c6060890e5

I would expect the popover to disappear when reset, just like with the sidebar.

If the ideal behavior is for the popover to disappear when you press the reset button, I'd be happy to help you update the code to achieve that.

richtabor:
> If the ideal behavior is for the popover to disappear when you press the reset button, I'd be happy to help you update the code to achieve that.

Agreed. Sure thing, I'd appreciate it.

packages/block-library/src/site-logo/edit.js
```diff
@@ -492,9 +500,15 @@ export default function LogoEdit( {
...
-	const onRemoveLogo = () => {
+	const onRemoveLogo = ( ref ) => {
 		setLogo( null );
 		setAttributes( { width: undefined } );
+		// Focus the toggle button and close the dropdown menu.
+		if ( ref.current ) {
+			const [ toggleButton ] = focus.tabbable.find( ref.current );
+			toggleButton?.focus();
+			toggleButton?.click();
+		}
```
t-hamano: I added this code to fix the issue pointed out in [this comment](https://github.com/WordPress/gutenberg/pull/63499#pullrequestreview-2193237747). This code ensures that when you press the Reset button, the dropdown closes and focus returns to the block toolbar button.

### Before
https://github.com/user-attachments/assets/1fa450c9-8121-4752-97a1-7a1527ce974d

### After
https://github.com/user-attachments/assets/43d30227-88bf-48a9-bb6c-14a19e1a1a97 

@Mamaduka I've used the code [here](https://github.com/WordPress/gutenberg/blob/be91ad65ecba61b353b5c3d74fba4d570c569348/packages/block-editor/src/components/global-styles/background-panel.js#L357-L366) as a guide, but do you know of any other good approaches?

Mamaduka: Maybe we should update the `MediaReplaceFlow` to support the "function as a child" component pattern. Then, we can pass `onClose` as a prop and let the dropdown in `MediaReplaceFlow` handle focus for us. The media upload buttons handle focus return using the same method.

I've proof of concept code that I can push on this branch.

t-hamano: Thank you for your advice.

Do you mean to change [this line](https://github.com/WordPress/gutenberg/blob/6128ceae4fbfcdbb0afbf976c34418a7d5e432ad/packages/block-editor/src/components/media-replace-flow/index.js#L205) to the following?

```javascript
{ typeof children === 'function'
	? children( { onClose } )
	: children }
```

Mamaduka: Yes. That's correct.

Mamaduka: @t-hamano, did I assume correctly that you want to work on the suggested changes?

t-hamano: Yes, I would be happy to work on the suggested changes. May I submit that in a separate PR?

Mamaduka: Sure. What works best for you 👍

t-hamano:
In this PR, I updated the children of the `MediaReplaceFlow` component so that they can receive `onClose`. As the video below shows, when the Reset button is pressed, the popover should close and the block toolbar button should be focused.

https://github.com/user-attachments/assets/8ed77d12-ea32-428b-9676-e39af6dfeed2

I also added a note about the changes to `MediaReplaceFlow` to the PR description.

Mamaduka (approved):
Thanks for the follow-up, @t-hamano!

Everything works as expected ✅

PR #63512, opened 2024-07-13
Block Switcher Preview: Adjust the position and enable pattern list preview in mobile viewport

### LINKED ISSUE #51309: Lack of screen space causes block transform preview to prevent user from actually clicking on the transform
### Description

When your screen real estate is small, transform preview popover pops over the thing you're hovering to get the preview going, causing it to flicker endlessly and sort of block you from actually clicking the transform you want.

1. Shrink your screen (was able to replicate this easily at 800px)
2. Create a list block
3. Go to transform the parent list block
4. Observe that the transform preview obscures the transform selection.

https://www.screencast.com/t/s7godNGXh

### Environment info

wp 6.2.2, no gutenberg plugin or any other plugin.

Yes

Yes

t-hamano:
Fixes #51309

## What?

This PR solves two problems:

- Preview is not positioned correctly on narrow browser width
- Pattern preview is not displayed in the mobile viewport, so conversion to pattern is not possible

### Preview is not positioned correctly on narrow browser width

The `Popover` component has `flip` prop to flip the popover when there isn't enough space by default. But the anchor element is a "point" and is fixed to the top right of the switcher popover:

anchor-point-1

As a result, the preview popover will flip based on this anchor and overlap the switcher popover if there is not enough space to the right.

### Pattern preview is not displayed in the mobile viewport, so conversion to pattern is not possible

On mobile viewports, [the preview popover is hidden via CSS](https://github.com/WordPress/gutenberg/blob/0aecc891389b6188da50bd865086a8934bdc7ce1/packages/block-editor/src/components/block-switcher/style.scss#L96), so clicking "Patterns" from the switcher popover does nothing.

mobile-pattern-transform

If we're transforming blocks, we don't necessarily need the preview popover, but for patterns, we need the preview popover because we need to select any pattern from the pattern list.

### Preview is not positioned correctly on narrow browser width

Position the preview popover's anchor so that it's equal in size and position to the switcher popover:

anchor-point-2

This will position the preview popover to the left of the switcher popover if there's not enough space to the right.

### Pattern preview is not displayed in the mobile viewport, so conversion to pattern is not possible

On mobile viewports, there isn't enough space to display the preview popover on either the left or right side. As a workaround, place the preview popover below the switcher popover:

image

## Testing Instructions

- Insert a block to test the transformation.
  - The Query Loop block would be appropriate for testing default transformations and pattern transformations.
  - The Group block would be appropriate for testing variation transformations.
- The preview popover should be positioned on the right when the browser is wide.
- Open the List View and make your browser width about 1000px. The preview popover will appear on the left.
- Narrow your browser width to 782px or less.
  - For default and variation transforms, the preview popover will not appear.
  - For pattern transforms, the preview popover will be located below the switcher popover.
- (Bonus): It should work as expected when "Top Toolbar" is enabled.

## Screenshots or screencast 

https://github.com/user-attachments/assets/77ce0305-e3cb-4c41-802d-1b66563f5192

### DIFF
--- a/packages/block-editor/src/components/block-switcher/pattern-transformations-menu.js
+++ b/packages/block-editor/src/components/block-switcher/pattern-transformations-menu.js
@@ -3,7 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { useState } from '@wordpress/element';
-import { useInstanceId } from '@wordpress/compose';
+import { useInstanceId, useViewportMatch } from '@wordpress/compose';
 import { chevronRight } from '@wordpress/icons';
 
 import {
@@ -34,6 +34,7 @@ function PatternTransformationsMenu( {
 } ) {
 	const [ showTransforms, setShowTransforms ] = useState( false );
 	const patterns = useTransformedPatterns( statePatterns, blocks );
+
 	if ( ! patterns.length ) {
 		return null;
 	}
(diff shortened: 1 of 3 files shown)


packages/block-editor/src/components/block-switcher/pattern-transformations-menu.js
```diff
@@ -60,21 +61,22 @@ function PatternTransformationsMenu( {
...
-							patterns={ patterns }
-							onSelect={ onSelect }
-						/>
-					</div>
-				</Popover>
-			</div>
+		<div className="block-editor-block-switcher__popover-preview-container">
+			<Popover
+				className="block-editor-block-switcher__popover-preview"
+				placement={ isMobile ? 'bottom' : 'right-start' }
```
t-hamano, line 70: For mobile layouts only, displays the pattern preview popover at the bottom.

ntsekouras (approved):
This works really well for me, thanks! I'm not 100% sure the patterns preview on mobile is perfect at the bottom, but of course in trunk is broken. Let's land this and explore an improvement for that, if needed in the future.

t-hamano:
@ntsekouras Thanks for the review!

jeffpaul:
@albanyacademy mind sharing your WordPress.org username so I can ensure its included in the 6.7 credits listing?

PR #63513, opened 2024-07-13
Block Bindings: Improve how the context needed by sources is extended in the editor

SantosGuillamot:
## What?
Stealing the code from Ella in [this other pull request](https://github.com/WordPress/gutenberg/pull/60826).

This pull request uses the code from https://github.com/WordPress/gutenberg/pull/63470  to reuse and merge the `usesContext` defined by the source in the server.

There are a few things happening:
* By default, it reuses the `usesContext` defined by the source in the server registration.
* It adds a new `usesContext` property to the client source registration to let sources extend that if needed.
* It merges both and removes duplicates.
* It changes the way bindings access the context to ensure it is only available when an attribute is connected to that specific source.

## Why?
Block bindings shouldn't have to affect the block registration and the context can be accessed just in time.

Apart from that, it should reuse the server definition to avoid duplication in server and client registrations.

## Testing Instructions
I added some automated tests to ensure everything works as expected: [link](https://github.com/WordPress/gutenberg/compare/try/register-sources-from-the-server...try/improve-how-bindings-context-is-extended-in-the-editor?expand=1#diff-d5aba57655b6c1abcbd8f937a61ad98110057a6ae09f2db596811dca31c54030).


packages/blocks/src/api/test/registration.js
```diff
@@ -1528,6 +1528,74 @@ describe( 'blocks', () => {
...
+				name: 'core/testing',
+				label: 'testing',
+				usesContext: 'should be an array',
+			} );
+			expect( console ).toHaveWarnedWith(
+				'Block bindings source usesContext must be an array.'
+			);
+		} );
+
+		it( 'should add usesContext when only defined in the client', () => {
```
artemiomorales: I believe this should read, "should add usesContext when only defined in the server".

packages/blocks/src/api/test/registration.js
```diff
@@ -1528,6 +1528,74 @@ describe( 'blocks', () => {
...
+			registerBlockBindingsSource( {
+				name: 'core/testing',
+				getValue: () => 'value',
+			} );
+			const source = getBlockBindingsSource( 'core/testing' );
+			unregisterBlockBindingsSource( 'core/testing' );
+			expect( source.usesContext ).toEqual( [ 'postId', 'postType' ] );
+		} );
+
+		it( 'should keep usesContext when it is not defined in the client', () => {
```
artemiomorales: And I believe this one should read, "should keep usesContext when only defined in the client".

artemiomorales (approved):
Aside from the two descriptions for the tests below, this looks good to me 👍 I believe this is good to merge if those are addressed.

packages/blocks/src/api/test/registration.js
```diff
@@ -1528,6 +1528,74 @@ describe( 'blocks', () => {
...
+			);
+		} );
+
+		it( 'should add usesContext when only defined in the client', () => {
+			// Simulate server bootstrap.
+			registerBlockBindingsSource( {
+				name: 'core/testing',
+				label: 'testing',
+				usesContext: [ 'postId', 'postType' ],
+			} );
```
artemiomorales: I see you're using `addBootstrappedBlockBindingsSource()` to simulate the server behavior in [a related PR](https://github.com/WordPress/gutenberg/pull/63797/files#). Should we be taking that approach in this PR as well? It seems like that would more closely resemble the server behavior.

SantosGuillamot: That makes sense 🙂 I'll make that change.

packages/blocks/src/api/test/registration.js
SantosGuillamot: I will change the text.

packages/blocks/src/api/test/registration.js
SantosGuillamot: I made the change as part of [this commit](https://github.com/WordPress/gutenberg/pull/63513/commits/1d7160491a65c402a9b2ca5da6d4c1a87ee0b8c7).

packages/blocks/src/api/test/registration.js
SantosGuillamot: I made the change as part of [this commit](https://github.com/WordPress/gutenberg/pull/63513/commits/1d7160491a65c402a9b2ca5da6d4c1a87ee0b8c7).

packages/blocks/src/api/test/registration.js
SantosGuillamot: I made the change as part of [this commit](https://github.com/WordPress/gutenberg/pull/63513/commits/1d7160491a65c402a9b2ca5da6d4c1a87ee0b8c7).

packages/blocks/src/api/test/registration.js
gziolo: Nice, this improves the readability of the test 👍🏻

PR #52664, opened 2023-07-16
Patterns: Remove `reusable` text from menu once rename hint has been dismissed

glendaviesnz:
## What?
Removes the `/reusable block` text from the `Create pattern` menu once the rename hint has been dismissed. Also sets the default sync status to `Synced` to better follow the existing reusable block workflow.

## Why?
Prevents the additional text hanging around for a full release cycle as the text is not used anywhere else in the interface it is only intended to be a temporary discovery mechanism. See #52194

## How?
Checks to see if the reusable block rename hint preference has been toggled off before displaying `Create pattern` menu.

## Testing Instructions

- Either use a new user account that hasn't toggle of the reusable block rename hint, or run the following script in the browser console to toggle the hint on `window.wp.data.dispatch(window.wp.preferences.store).set('core', 'isResuableBlocksrRenameHintVisible', true);`
- Add a paragraph block and in block overflow menu check that the Create pattern menu says `Create pattern/reusable block`
- Select menu to open the modal, and dismiss the rename hint notice
- Close the modal and open the block menu again and check that it now says `Create pattern`
- Reload the browser and open the Create pattern modal again and make sure that the default sync status is toggled on
- Save a few synced and unsynced blocks and make sure they behave as expected.
- Reset the hint again with `window.wp.data.dispatch(window.wp.preferences.store).set('core', 'isResuableBlocksrRenameHintVisible', true);` and then go to `/wp-admin/edit.php?post_type=wp_block` and add a new pattern and make sure the rename hint displays correctly in the model

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/3629020/8991e5c8-9835-4876-a462-018746d3b3bd

PR #63514, opened 2024-07-13
Quality: Remove "reusable block name hint" code

t-hamano:
Related to #52664

## What?

This PR removes the block name hint, and associated code, from the Create Pattern modal.

| Before | After |
|--------|--------|
| before | after | 

## Why?

This hint was added in WordPress 6.3 to let users know that the name changed from "Reusable Block" to "Synced Pattern". As [this comment](https://github.com/WordPress/gutenberg/pull/52664/files#diff-e2733ba5046877c02344a307ea83132978c799a5eb4fd643f612b5daeced3539R14-R16) indicates, this hint was scheduled to be removed in 6.4, so I don't think it's necessary anymore.

## How?

I've removed the relevant components, hooks and CSS. they're private APIs so it should be fine to remove them directly.

## Testing Instructions

- Access `http://localhost:8888/wp-admin/post-new.php?post_type=wp_block`.
- To enable hints, run the following code in your browser console:
  ```
  window.wp.data.dispatch(window.wp.preferences.store).set('core', 'isResuableBlocksrRenameHintVisible', true);
  ```
- No hints will be displayed.


packages/block-editor/README.md
```diff
@@ -730,10 +730,6 @@ _Returns_
 
 -   `JSX.Element`: A React element.
 
-### ReusableBlocksRenameHint
-
-Undocumented declaration.
-
```
Mamaduka, line 736: Was this a public component?

Mamaduka (approved):
Makes sense 👍

packages/block-editor/README.md
t-hamano: Oh, I just realized that this component is both private and public at the same time 😅

According to [WPdirectory's results](https://wpdirectory.net/search/01J2RH120VV51K7BTA912YZJD4), there are two plugins other than the Gutenberg plugin that use this component.

Should we keep this component?

Mamaduka: I wonder if those two plugins are bundling the `block-editor` package. I can't find the `ReusableBlocksRenameHint` instance in their [codebase](https://github.com/search?q=repo%3Amailpoet%2Fmailpoet%20ReusableBlocksRenameHint&type=code).

The same is probably true for the "Kubio AI Page Builder" plugin.

t-hamano: I downloaded the two plugins. Neither plugin has pre-build code, but looking at the build files, it doesn't look like they use this component.

Therefore, I think it's safe to delete it.

Mamaduka: Yeah, it should be okay. Let's wait until tomorrow if others have different opinions regarding the public component.

Mamaduka:
Let's merge this.

PR #47670, opened 2023-02-02
Show a pointer/hint in the settings tab informing the user about the styles tab

talldan:
## What?
Fixes #47410

## Why?
It's nice to let the user know about this change

## How?
- Implements a new (non-public, non-reusable) component for showing this pointer ([see the discussion in the issue about why `Notice` wasn't used](https://github.com/WordPress/gutenberg/issues/47410#issuecomment-1407969824))
- Uses the preferences package to record the state of whether the notice has been dismissed or not.

## Testing Instructions
Note - if you want to test this multiple times, you can run this snippet in the browser console to bring the dismissed pointer back again:
```js
wp.data.dispatch( 'core/preferences' ).set( 'core', 'isInspectorControlsTabsHintVisible', true )
```

1. Add a block (a good one to test with is 'Group')
2. Observe the 'pointer' in the sidebar
3. Dismiss it (at this point it shouldn't ever come back again)

### Testing Instructions for Keyboard
1. Add a group block by tabbing into the editor content and typing `/group` and hitting the enter key
2. Tab to the 'Editor settings' region. If it's not open you should land upon a button 'Open document settings' first, so use that button to open it.
3. Tab till you get to a 'Dismiss hint' button, optionally read the text that's in the DOM before that button (but not focusable).
4. Select that button.
5. Focus should be transferred to the Advanced section as the hint disappears

## Screenshots or screencast 
Screen Shot 2023-02-02 at 1 21 21 pm

PR #63515, opened 2024-07-13
Block Editor: Remove hint in the Settings tab

### LINKED ISSUE #63506: Remove "Looking for other block settings" hint
When opening the Inspector for the first time, you still see the "Looking for other block settings? They've moved to the styles tab." notice. I propose we remove this notice, as this move happened over a year ago, back in WordPress 6.2.

Related https://github.com/WordPress/gutenberg/pull/47670. 

### Visual 

CleanShot 2024-07-12 at 17 33 36

t-hamano:
Fixes #63506
Related to #47670
Similar to #63514

## What?

This PR removes the hint about the Style tab from the Settings tab in the block sidebar.

image

## Why?

This hint was added to inform users of the change about `InspectorControls` and was shipped in WordPress 6.2. #47670 doesn't say whether this hint will be removed in the future, but since it's been over a year since WordPress 6.2 was released, I think it's safe to remove it now.

## How?

Just remove. The SettingsTabHint component is not exposed externally, so it is safe to delete it directly.

## Testing Instructions

- Run this snippet in the browser console to bring the dismissed pointer back again:
  ```
  wp.data.dispatch( 'core/preferences' ).set( 'core', 'isInspectorControlsTabsHintVisible', true )
  ``
- Insert a block that has Settings and Styles tabs, e.g. the Group block.
- Open the block sidebar and there should be no hints under the Advanced panel.


Mamaduka (approved):
LGTM ✅

PR #63516, opened 2024-07-13
Remove double shadow on Inserter category panel when zoomed out

### LINKED ISSUE #63511: Remove double shadow on Inserter category panel when zoomed out
It is only present when the zoom out experiment is engaged. 

Let's confirm that the fix does not remove the shadow when zoom out is not active. 

### Current 
CleanShot 2024-07-12 at 20 24 47

### Expected

CleanShot 2024-07-12 at 20 25 59

amitraj2203:
## What?
Fixes: #63511 

## Why?
When the experimental `Zoom Out` is enabled, we get to see double shadow on the category panel.

#### Turn **ON** `Zoom Out`
1. Open Site Editor
2. Edit any template and click on Inserter > Pattern > Click on any pattern
3. See that there isn't any double shadow.

#### Turn **OFF** `Zoom Out`
1. Open Site Editor
2. Edit any template and click on Inserter > Pattern > Click on any pattern
3. See that there isn't any regression (the fix does not remove the shadow when zoom out is not active).

## Screenshots or screencast 

When Zoom Out is enabled:
image

When Zoom Out is disabled:
image

### DIFF
--- a/packages/block-editor/src/components/inserter/style.scss
+++ b/packages/block-editor/src/components/inserter/style.scss
@@ -703,6 +703,10 @@ $block-inserter-tabs-height: 44px;
 			height: 100%;
 		}
 	}
+
+	.block-editor-inserter__category-panel {
+		box-shadow: none;
+	}
 }
 
 .show-icon-labels {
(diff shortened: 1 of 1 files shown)


richtabor (approved):
I added an inline comment. It's not 100% ideal based on zoom out, but for now this is the best approach (without refactoring).

PR #63523, opened 2024-07-13
Issue template: use checkboxes instead of dropdown

### LINKED ISSUE #51815: ISSUE_TEMPLATE: Confirmation checks do not play a role

When creating an issue as a bug report, the [Bug_report.yml](https://github.com/WordPress/gutenberg/blob/trunk/.github/ISSUE_TEMPLATE/Bug_report.yml) template is applied. This template includes a validation that requires the selection of two dropdowns:

- Please confirm that you have searched existing issues in the repo.
- Please confirm that you have tested with all plugins deactivated except Gutenberg.

However, this dropdown allows you to submit an issue if either Yes, No, or both are selected.

issue_template

To improve this, I have two ideas:

- Remove this dropdown.
- Change to checkboxes. I have created a temporary repository to test this template: https://github.com/t-hamano/issue-template-test/issues/new/choose

new_issue_template

ellatrix:
## What?

Also removes "multiple", it doesn't make sense at all to select both. I've also just left a checkbox for Yes.

## Why?

It's super annoying to have to open a dropdown twice to be able to file a bug report.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug_report.yml
@@ -52,16 +52,16 @@ body:
     - type: checkboxes
       id: existing
       attributes:
-          required: true
           label: Please confirm that you have searched existing issues in the repo.
           description: You can do this by searching https://github.com/WordPress/gutenberg/issues and making sure the bug is not related to another plugin.
           options:
               - label: 'Yes'
+                required: true
 
     - type: checkboxes
       id: plugins
       attributes:
-          required: true
           label: Please confirm that you have tested with all plugins deactivated except Gutenberg.
           options:
               - label: 'Yes'
+                required: true
--- a/.github/ISSUE_TEMPLATE/Bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug_report.yml
@@ -56,8 +56,7 @@ body:
           description: You can do this by searching https://github.com/WordPress/gutenberg/issues and making sure the bug is not related to another plugin.
           options:
               - label: 'Yes'
-      validations:
-          required: true
+              - required: true
 
     - type: checkboxes
       id: plugins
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/Bug_report.yml
```diff
@@ -49,25 +49,21 @@ body:
...
       id: existing
       attributes:
           label: Please confirm that you have searched existing issues in the repo.
           description: You can do this by searching https://github.com/WordPress/gutenberg/issues and making sure the bug is not related to another plugin.
-          multiple: true
           options:
-              - 'Yes'
-              - 'No'
+              - label: 'Yes'
       validations:
```
t-hamano: It seems that if we want a checkbox to be required, we need to define it in the `options` field, not the `validations` field.

See this code sample: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

t-hamano (commented):
I also suggested this problem in #51815.

To actually test the issue template, I created a test repository in my GitHub account.

The code from this PR is reflected here:
https://github.com/t-hamano/issue-template-test/issues/new?assignees=&labels=%5BType%5D+Bug&projects=&template=Bug_report.yml

.github/ISSUE_TEMPLATE/Bug_report.yml
ellatrix: Ok, it should be good now. You can actually preview it here https://github.com/WordPress/gutenberg/blob/dc41b4fd10d6828f3985cffecfc52143d1011646/.github/ISSUE_TEMPLATE/Bug_report.yml

t-hamano (approved):
LGTM! Just to be sure, I tested it in [my repository](https://github.com/t-hamano/issue-template-test/issues/new?assignees=&labels=%5BType%5D+Bug&projects=&template=Bug_report.yml) and the validation for the checkboxes seems to work.

https://github.com/user-attachments/assets/b74489a3-18d6-4e1b-b0a9-1bd2453f56ca

alexstine (approved):
@ellatrix Awesome, makes this form so much nicer.

PR #63528, opened 2024-07-13
Move entity-provider.js exports into hooks/index.ts so they are added to the docs

### LINKED ISSUE #63475: Can't find useEntityProp in the Block Editor Handbook
### Description

useEntityProp is not documented in the Block Editor Handbook. 
I only see the hooks useEntityRecord, useEntityRecords and useResourcePermissions
Is this on purpose? Is it deprecated?

(Honestly, I'm not sure it's okay to report this as a bug. Feel free to tell me where the right place for it would be.)

1. I searched here: https://developer.wordpress.org/block-editor/reference-guides/packages/packages-core-data/
2. I find it here in a how to guide for meta boxes: https://developer.wordpress.org/block-editor/how-to-guides/metabox/

_No response_

### Environment info

_No response_

Yes

Yes

ryanwelcher:
## What?
The hooks defined in `core-data/entity-provider.js` are not documented because the file  is not being pulled into the README.md for docgen to generate.

Closes: #63475.

## Why?
These hooks should be documented under https://developer.wordpress.org/block-editor/reference-guides/packages/packages-core-data/#hooks

## How?
I have move all of the hooks in `entity-provider.js` into their own files inside the `hooks` directory and exported them from there. This will have them be picked up by the docgen tool and add them to the documentation.

## Testing Instructions
- Checkout branch and build the new changes
- Confirm that `useEntityProp` is still available on the `wp.coreData` global and there are no errors in the block editor.


Mamaduka:
Maybe we should move the entity hooks into the `hooks.ts` file instead of changing exports.

youknowriad:
Do we know why the file is not pulled into the README? AFAIK docgen documents all public exports no?

ryanwelcher:
> Do we know why the file is not pulled into the README? AFAIK docgen documents all public exports no?

Doc gen only pulls in files that are referenced in the tokens in the readme. There is a [reference to the hooks.ts](https://github.com/WordPress/gutenberg/blob/trunk/packages/core-data/README.md?plain=1#L977) file but none for the entity-provider.js file and the hooks that files contains are not exported in hooks.ts so they are missed

youknowriad:
@ryanwelcher makes sense, so why are we not updating the README.md file's tokens instead?

ryanwelcher:
@youknowriad I did that at first but thought we should keep all of the hooks together in the code base. I'm fine either way so if we're concerned about moving imports that would be the approach with the least amount of risk.

youknowriad:
I don't mind, I was just trying to understand why it was not working. For me, we shouldn't be adapting the code to the docs but If these things (hooks/intex.ts and entity-provider.js) are considered similar, it would make sense to group them in the code as well.

That said, it seems EntityProvider exposes more than just hooks. It also exposes a component.

packages/core-data/README.md
```diff
@@ -976,6 +976,47 @@ The following set of react hooks available to import from the `@wordpress/core-d
 
 <!-- START TOKEN(Autogenerated hooks|src/hooks/index.ts) -->
 
+### useEntityBlockEditor
```
youknowriad, line 979: Why is the EntityProvider not present in the docs. AFAIK, it's exported in the file as well?

ryanwelcher: It exported as the default but because it's a component I didn't move it's export into hooks.

ryanwelcher:
> For me, we shouldn't be adapting the code to the docs but If these things (hooks/intex.ts and entity-provider.js) are considered similar, it would make sense to group them in the code as well.

i would say that they are related as they all work with entities. The useEntityRecord and useEntityRecord are in the hooks directory.

ryanwelcher:
I think @Mamaduka suggestion probably makes the most sense here. Unless you have any concerns with that @youknowriad , I'll update the PR

youknowriad:
I'm not sure I understand the suggestion. If it's about moving the hooks to `hooks/index.ts` then I disagree with it `hooks/index.ts` is an entry point that exports hooks that are defined in their own files on that folder. If it's about moving the `entity-provider.ts` hooks into the hooks folder, than that's fine by me (each hook to its own file I guess).

Otherwise, I'm also ok with the current PR as it stands.

ryanwelcher:
Sounds good to me @youknowriad.

To be clear, I was planning on moving the hooks in `entity-provider.js` into the `hooks` directory as their own files. Sorry I didn't make that clear :)

I'll get the PR updated now.

packages/core-data/src/entity-provider.js
```diff
@@ -1,22 +1,7 @@
...
-/**
- * Internal dependencies
- */
-import { STORE_NAME } from './name';
-import { updateFootnotesFromMeta } from './footnotes';
-
-const EMPTY_ARRAY = [];
+import { createContext, useContext, useMemo } from '@wordpress/element';
 
 const EntityContext = createContext( {} );
```
ryanwelcher: @Mamaduka @youknowriad I just wanted to call this out as there was a single `EntityContext` being used across the hooks and to move  the `useEntityId` hook into it's own file, I needed to create the same there. 

I don't see any errors/failures but wanted to be sure that doing this won't cause any unexpected outcome.

youknowriad: No, that's a mistake, it should use the exact same EntityContext in all the files. You can do so by creating an `entity-context` file and importing it in all the files.

ryanwelcher: Updated with latest commit. Thanks for confirming, it felt like that was a bad idea :)

youknowriad (approved):

Mamaduka (approved):
Thanks, @ryanwelcher!

Issue #58428, opened 2024-01-30
Font Library: remove context completely


t-hamano:
The original main purpose of context is to resolve redundant props passing as mentioned in [this document](https://react.dev/learn/passing-data-deeply-with-context). However, the tree of font-library-related components is not that deep, and there are many contexts that are once used and are never used.

These contexts should be able to be removed entirely by passing props or using an ad-hoc approach.

Not using a provider makes it easier to completely separate the modal dialog and the typography panel and delegate control of opening and closing modal dialog to a data store, as attempted in #58350.

Eventually, it will be easier to trigger the Font Library modal dialog via the command palette, as suggested in #54880.

The only concern, as mentioned in [this comment](https://github.com/WordPress/gutenberg/pull/58350#discussion_r1470714846), is that the initial state of the Font Library modal dialog is also context-managed. To solve this, we will need to add an argument to `openModal` in the `interfaces` package that defines the initial state.

t-hamano:
P.S. We don't need to solve this problem all at once with one PR, but I think it would be better to remove contexts step by step using multiple PRs.

getdave:
If some values it would be useful to share as context, one other option is to break the context into smaller contexts for specific use cases and consume them as close to where they are needed as possible. Or just put _some_ of the values in context. Either way, avoiding having one huge context is a good plan.

PR #63531, opened 2024-07-14
Font Library: Include a "Select All" options to activate/deactivate all fonts

### LINKED ISSUE #59704: Font Library: Include a "Select All" option when activating/deactivating fonts
It's quite time consuming and burdensome to have to select each font weight when I want to add or remove an entire font family. I suggest we add a "Select All" tick at the top, like this example:

**Currently:**
fonts

akasunil:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/59704

## Why?

Because there was no option to pick all fonts, choosing fonts one by one in the font manager was a time-consuming process.

## How?

This PR add `Select all` option to font manager.
 
## Screenshots or screencast 
 
[font-library-select-all-toggle-function.webm](https://github.com/user-attachments/assets/318847b5-a18f-40f0-809d-e6d7452285b0)


packages/edit-site/src/components/global-styles/font-library-modal/installed-fonts.js
```diff
@@ -295,6 +310,18 @@ function InstalledFonts() {
 							</Text>
 							<Spacer margin={ 4 } />
 							<VStack spacing={ 0 }>
+								<Flex
+									className="font-library-modal__select-all"
+									justify="flex-start"
+									align="center"
+									gap="1rem"
+								>
```
t-hamano: We should be able to place the checkbox properly without using the `Flex` component by simply applying a CSS class to the `CheckboxControl` component itself.

packages/edit-site/src/components/global-styles/font-library-modal/installed-fonts.js
```diff
@@ -50,8 +51,12 @@ function InstalledFonts() {
 		notice,
 		setNotice,
 		fontFamilies,
+		deactivateFontFamily,
+		activateFontFamily,
 	} = useContext( FontLibraryContext );
 	const [ isConfirmDeleteOpen, setIsConfirmDeleteOpen ] = useState( false );
+	const [ isSelectAllChecked, setIsSelectAllChecked ] = useState( true );
```
t-hamano: We should be able to determine ad hoc whether all fonts are selected or not without using state. Additionally, I think it's necessary to have a state that shows that only some fonts are selected.

Below is a screenshot of the DataView:

dataview

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
```diff
@@ -186,3 +186,15 @@ button.font-library-modal__upload-area {
...
+
+.font-library-modal__select-all {
+	width: 100%;
+	height: auto;
+	padding: $grid-unit-20;
+
+	.components-checkbox-control__label {
+		padding-left: $grid-unit-20;
+		font-size: 18px;
+		line-height: 1;
```
t-hamano: In my opinion, it seems fine to leave the font size and line height at the default values ​​of the `CheckboxControl` itself.

t-hamano (commented):
Thanks for the PR!

I think the important thing here is not to add new contexts.

This is because the existing contexts are not used in their intended way, and the aim is for them to all be deleted eventually.

For details, see #58428.

We can probably use the existing `isFontActivated` and `toggleActivateFont` contexts.

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
akasunil: That would appear strange because default size is too small in comparison to the list of font faces title.

akasunil:
> We can probably use the existing `isFontActivated` and `toggleActivateFont` contexts.

To do this, we would need to run `toggleActivateFont` for every font faces. Wouldn't that lead to problems with performance?

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
akasunil: Here is screenshot for example,
image

jasmussen:
Thanks for the PR! Behavior wise, this will be a big win. Thank you!

t-hamano:
> To do this, we would need to run `toggleActivateFont` for every font faces. Wouldn't that lead to problems with performance?

The current implementation also uses forEach, so I don't think there will be much difference in terms of performance. Or you can handle it ad-hoc without using toggleActivateFont.

Either way, we should avoid adding a new context.

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
t-hamano: Personally I don't mind, but what do you think, @jasmussen?

akasunil:
Here is update work flow of selection all option

[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress.webm](https://github.com/user-attachments/assets/59857023-16a5-47f8-a732-a042f95f1a3c)

packages/edit-site/src/components/global-styles/font-library-modal/installed-fonts.js
```diff
@@ -148,6 +154,53 @@ function InstalledFonts() {
...
+		activeFontsCount !== libraryFontSelected?.fontFace?.length;
+
+	// Check if all fonts are selected.
+	const isSelectAllChecked =
+		activeFontsCount === libraryFontSelected?.fontFace?.length;
+
+	// Toggle select all fonts.
+	const toggleSelectAll = () => {
+		const initialFonts = fontFamilies?.[ libraryFontSelected.source ] ?? [];
+		const deactivateFonts = isSelectAllChecked || isIndeterminate;
```
t-hamano:
```suggestion
		const deactivateFonts = isSelectAllChecked;
```

The current condition is that if some fonts are selected and you check "Select All", all fonts will be unchecked.

It would be good to align this behavior with the Block Lock UI, i.e. if some fonts are selected and you check "Select All", all fonts will be checked.

https://github.com/user-attachments/assets/eedca853-bf2c-4dad-850d-ed4fb32c86ad


However, making this change will log errors to the browser console. We may need to adjust some logic to go along with this change.

```
Warning: Encountered two children with the same key, `heading`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version
```

packages/edit-site/src/components/global-styles/font-library-modal/context.js
```diff
@@ -516,6 +516,7 @@ function FontLibraryProvider( { children } ) {
 				libraryFontSelected,
 				handleSetLibraryFontSelected,
 				fontFamilies,
+				setFontFamilies,
```
t-hamano: Let's avoid adding a new context. We can simply add the following directly to the `installed-fonts.js` file:

```javascript
const [ fontFamilies, setFontFamilies ] = useGlobalSetting(
	'typography.fontFamilies'
);
```

t-hamano (commented):
Thanks for the update!

Overall, it seems to work pretty well. Let's get some feedback from @WordPress/gutenberg-design on the design of the "Select All" checkbox.

akasunil:
Thanks @t-hamano, I have updated PR as per your suggestions. 
Here is latest video of fonts selection work flow
 
 
[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (1).webm](https://github.com/user-attachments/assets/90b3cd09-e862-42e3-94cb-74672e099696)

t-hamano:
Thanks for the update!

When I tested it again, I found some unexpected behavior with certain fonts.

Enable TT4 and open "System Sans-serif" or "System Serif". These fonts have only one variation, but the initial state of the "Select all" checkbox is "indeterminate." And clicking the checkbox does not seem to do what you expect.

However, this problem does not occur with "Inter" fonts, which have only one variation.

https://github.com/user-attachments/assets/d63766d6-88d4-4123-81b5-45be5b84e545

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
jameskoster: I also think this is fine. We should avoid ad hoc font styles where possible.

akasunil:
> Enable TT4 and open "System Sans-serif" or "System Serif". These fonts have only one variation, but the initial state of the "Select all" checkbox is "indeterminate."

That's odd. let me check.

akasunil:
Its fine now.

[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (4).webm](https://github.com/user-attachments/assets/d27f46d6-bcd4-4eea-ac75-549fa940e76b)

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
akasunil: Font styles are removed now.

akasunil:
@t-hamano Please check video attached in PR description for final work flow and design of `Select All` toggle.

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
```diff
@@ -186,3 +186,13 @@ button.font-library-modal__upload-area {
...
+
+.font-library-modal__select-all {
+	width: 100%;
+	height: auto;
+	padding: $grid-unit-20;
+
+	.components-checkbox-control__label {
+		padding-left: $grid-unit-20;
+	}
+}
```
t-hamano:
```suggestion
.font-library-modal__select-all {
	padding: $grid-unit-20 $grid-unit-20 $grid-unit-20 $grid-unit-20 + $border-width;

	.components-checkbox-control__label {
		padding-left: $grid-unit-20;
	}
}
```

checkbox

To properly align the vertical position of the checkboxes, let's add a 1px border width to the left side.

I've also removed styles that we don't need.

t-hamano (approved):
LGTM!

I have one final suggested code change.

packages/edit-site/src/components/global-styles/font-library-modal/style.scss
akasunil: Good catch!

PR #63533, opened 2024-07-14
Font Library: Improve 'No fonts installed' state when fonts are installed but not activated

### LINKED ISSUE #63418: Font Library: Improve 'No fonts installed' state when fonts are installed but not activated
### Description

Splitting this out from https://github.com/WordPress/gutenberg/pull/63211#discussion_r1671929082

Cc @matiasbenedetto 

WHen there are fonts installed but none of them is activated, the UI still refers to a 'no fonts installed' verbiaage:
- Th Styles panel shows a 'No fonts installed.` text,
- The button that opens the modal dialog text is 'Add fonts'.
- The button to open the modal dialog brings users to the 'Upload' tab. It would be best to bring users to the 'Library' tab.

This UI could be improved to clarify there are fonts installed, they're just not active

Screenshot 2024-07-11 at 12 00 35

- Go to the Site editor > Styles > Typography
- Make sur eto have some fonts installed.
- Deactivate all fonts.
- Close the fonts modal dialog
- Observe the 'No fonts installed.' text in the Styles panel
- Observe the 'Add fonts' button text

_No response_

### Environment info

_No response_

Yes

Yes

akasunil:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/63418

## Why?

Styles panel shows a 'No fonts installed.` text event when fonts are installed, but not activated. 

## How?

Added condition to display message conditionally. Now it will display  'No fonts activated.` when fonts are installed but not activated.

## Testing Instructions

- Open Styles panel , go to typography
- If you don't have any font installed, it should show  'No fonts installed.`
- If you have any font installed, it should show  'No fonts activated.`


t-hamano (commented):
Thanks for the PR!

To ideally resolve #63418, I think the following two things also need to be addressed:

> The button that opens the modal dialog text is 'Add fonts'.
> The button to open the modal dialog brings users to the 'Upload' tab. It would be best to bring users to the 'Library' tab.

In this PR, the text is correct when there are no fonts activated, but the button text is "Add fonts":

image

The button text should be "Manage fonts" and the tab that opens when you click the button should be "Installed fonts".

Additionally, the following scenarios need to be considered:

- No theme fonts
- Installed fonts are present
- All installed fonts are deactivated

In this case, the text should be "No fonts activated." and the button text should be "Manage fonts", but this is not the case in this PR.

packages/edit-site/src/components/global-styles/font-families.js
```diff
@@ -39,6 +44,7 @@ function FontFamilies() {
 				.sort( ( a, b ) => a.name.localeCompare( b.name ) )
 		: [];
 	const hasFonts = 0 < customFonts.length || 0 < themeFonts.length;
+	const hasBaseFonts = baseFontFamilies?.theme?.length > 0;
```
matiasbenedetto: This conditional is not enough to know if there are fonts installed but not activated. Currenty, this only checks if the theme provides fonts that are not activated but  it doesn't check if there are custom fonts installed but they are not activated.

akasunil: You're right; it's likely possible that the only available fonts are custom ones that haven't been activated. I'll update the PR

t-hamano (approved):
LGTM! Everything is working as expected.

https://github.com/user-attachments/assets/e0612e8a-cd25-43ab-af31-b3054dc48896

Note: During testing, I noticed that if there are no installed/theme fonts present, the Library tab content is empty. This could be improved in a follow-up.

image

akasunil:
Thank you @t-hamano 

> I noticed that if there are no installed/theme fonts present, the Library tab content is empty.

ill create separate issue for this.

akasunil:
Also, the E2E test passes successfully in my local, here it fails.
 
 
image

t-hamano:
I don't know why the E2E tests succeed locally, I think it's because the button text should be "Manage Fonts" but it's "Add Fonts".

The failing test is [here](https://github.com/WordPress/gutenberg/blob/f717121a14fe1613dbbef6899de021acd9253c0e/test/e2e/specs/site-editor/font-library.spec.js#L115-L179), and below is a video of running the test manually.

https://github.com/user-attachments/assets/c9dd6fd6-1482-4c1d-8779-22bdf795c0e7

To run the test manually, visit `http://localhost:8889/wp-admin` and enable the Emptytheme.

You will notice that the button text is unintentionally "Add Fonts". So maybe there's something missing in the current implementation.

akasunil:
@t-hamano  It works fine now

[Index-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (1).webm](https://github.com/user-attachments/assets/e87ce49d-cede-4b10-b4de-ae7566d6dc99)

t-hamano (approved):
LGTM! For me everything is working as expected.

All tests pass, so I think we can ship this PR.

PR #63535, opened 2024-07-14
Heading: Enable heading level curation

### LINKED ISSUE #20213: Remove heading settings (levels) from core/heading
Hey,

I need to remove the heading settings for a project, I assume that is best done using filters? 

Using this guide here: https://developer.wordpress.org/block-editor/developers/filters/block-filters/#editor-blockedit it's clear how to add panels to the existing settings, however, it doesn't say anything about removing an existing panel from the inspector.

I have tried something like that to see if existing settings can be changed:
```
const { createHigherOrderComponent } = wp.compose;
const { Fragment } = wp.element;

const withInspectorControls = createHigherOrderComponent( ( BlockEdit, HeadingToolbar ) => {
	return ( level, setAttributes, props ) => {
		return (
			<Fragment>
				<BlockEdit { ...props }>
					<HeadingToolbar
						minLevel={ 1 }
						maxLevel={ 1 }
						selectedLevel={ level }
						onChange={ ( newLevel ) =>
							setAttributes( { level: newLevel } )
						} />
				</BlockEdit>
			</Fragment>
		);
	};
}, 'withInspectorControl' );

wp.hooks.addFilter( 'editor.BlockEdit', 'core/heading', withInspectorControls );
```

However, this just gives me a blank block.

I would really appreciate some help with this.

Thanks!

### LINKED ISSUE #63527: Add a mechanism to Heading blocks that allows you to control the available levels
The heading level dropdown in the Heading block is powered by the [HeadingLevelDropdown](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/block-heading-level-dropdown) component. This component accepts and `options` parameter that allows you to choose which levels are available in the dropdown. Unfortunately, there is no way to modify this `options` parameter from a Heading block. 

Being able to modify the available levels is critical for heavily curated editing environments and block "governance".

Consider the following situations:

- You only want to allow `h2`, `h3`, and `h4` in Media & Text blocks.
- You want to disable `h1` on posts since the template already populates an `h1` for the post title.
- You want to disable the usage of `h5` and `h6` across the whole site.
- etc.

Curating the available heading levels is currently not possible, but it should be easy to implement using an additional attribute for the Heading block. Developers can then modify this attribute as they need, either directly in theme patterns and templates or via block filters like `blocks.registerBlockType` or `register_block_type_args`. 

| Current functionality | Goal | 
|-|-|
|image|image|

Should the mechanism be added to Heading blocks, we should explore similar mechanisms for other blocks with level dropdowns, like Site Title, Post Title, etc.

ndiego:
Fixes https://github.com/WordPress/gutenberg/issues/63527, https://github.com/WordPress/gutenberg/issues/20213

## What?
This PR is a minimalistic attempt to solve the longstanding issue that you cannot modify the available heading levels provided by the Heading block. This block is powered by the [HeadingLevelDropdown](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/block-heading-level-dropdown) component. This component accepts an `options` parameter that allows you to choose which levels are available in the dropdown. Unfortunately, there is no way to modify this `options` parameter from a Heading block.

This PR adds a `levelOptions` parameter to the Heading block to resolve this. 

## Why?
Being able to restrict the available heading levels is crucial in many situations, whether it be general Editor curation, accessibility, block governance, SEO, etc. 

## How?
This PR adds a `levelOptions` parameter to the Heading block that allows developers to define which heading levels should be displayed in the Heading dropdown UI. The approach is very simple and does not require a depreciation. Any previously set heading levels are respected in the markup. This attribute just allows you to control the UI. I could see how we may want to extend this future in the future, but the goal with this PR is to keep things simple and solve an immediate need of the community. Implementing this attribute does not block/preclude additional enhancements in the future. 

I chose the name `levelOptions` because we are already using the `level` attribute in the block, and the parameter in the `HeadingLevelDropdown` component is `options`. 

If we like this approach, follow-up PRs should be created to add similar functionality to all blocks that use the `HeadingLevelDropdown`.

With this new attribute, you can restrict the UI in many different ways, making this approach very flexible and powerful. For example, you could restrict options directly in block markup for a pattern or template. Try copying and pasting the following in the Editor. 

```
<!-- wp:heading {"level":3,"className":"wp-block-heading","levelOptions":[3,4,5]} -->
<h3 class="wp-block-heading">Markup example</h3>
<!-- /wp:heading -->
```

Or you could modify the attribute programmatically via filters. The following will disable `h1` globally. You could also add conditionals for post type, user permissions, etc. There are filters for both PHP and JavaScript so the applications are endless.

```
function example_modify_heading_levels_globally( $args, $block_type ) {
	
	if ( 'core/heading' !== $block_type ) {
		return $args;
	}

	// Remove H1.
	$args['attributes']['levelOptions']['default'] = [ 2, 3, 4, 5, 6 ];
	
	return $args;
}
add_filter( 'register_block_type_args', 'example_modify_heading_levels_globally', 10, 2 );

``` 

## Testing Instructions
1. Use any block theme with this PR enabled. You can also test in [Playground](https://playground.wordpress.net/gutenberg.html).
2. Try copying the block code above into the Code Editor, then switch to the Editor View to see the restricted Heading levels UI
3. Try copying the filter code into the `functions.php` file of your theme and see that `h1` is disabled. 

## Screenshots or screencast

| Default | Block markup example | Filter example |
|-|-|-|
|image|image|image|

cc @ryanwelcher @troychaplin @amberhinds since we chatted about this at WordCamp Canada. 🇨🇦
cc @carolinan @fabiankaegy @richtabor @jasmussen for additional code/design feedback 🙏

### DIFF
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -360,7 +360,7 @@ Introduce new sections and organize content to help visitors (and search engines
 -	**Name:** core/heading
 -	**Category:** text
 -	**Supports:** __unstablePasteTextInline, align (full, wide), anchor, className, color (background, gradients, link, text), interactivity (clientNavigation), spacing (margin, padding), splitting, typography (fontSize, lineHeight)
--	**Attributes:** content, level, placeholder, textAlign
+-	**Attributes:** content, level, levelOptions, placeholder, textAlign
 
 ## Home Link
 
--- a/packages/block-library/src/heading/block.json
+++ b/packages/block-library/src/heading/block.json
@@ -21,6 +21,9 @@
 			"type": "number",
 			"default": 2
 		},
+		"levelOptions": {
+			"type": "array"
+		},
 		"placeholder": {
 			"type": "string"
 		}
--- a/packages/block-library/src/heading/edit.js
+++ b/packages/block-library/src/heading/edit.js
@@ -32,7 +32,8 @@ function HeadingEdit( {
 	style,
 	clientId,
 } ) {
-	const { textAlign, content, level, placeholder, anchor } = attributes;
+	const { textAlign, content, level, levelOptions, placeholder, anchor } =
+		attributes;
 	const tagName = 'h' + level;
 	const blockProps = useBlockProps( {
 		className: clsx( {
@@ -95,6 +96,7 @@ function HeadingEdit( {
 				<BlockControls group="block">
 					<HeadingLevelDropdown
 						value={ level }
+						options={ levelOptions }
 						onChange={ ( newLevel ) =>
 							setAttributes( { level: newLevel } )
 						}


rosa2:
Please review this needed issue.

youknowriad:
I see 6 blocks using the "level" attribute. Do we want to be consistent and add this attribute to all the blocks.

Personally, I'm fine with the proposed approach here, especially as changing the default value of this particular attribute doesn't invalidate the block. It's also a very similar approach to what we did before for things like "allowedBlocks". 

Maybe get a second opinion. cc @gziolo @mtias

ndiego:
> I see 6 blocks using the "level" attribute. Do we want to be consistent and add this attribute to all the blocks.

Yeah, that was my thought, but I wanted to use the Heading block as a proof of concept. Happy to tackle the remaining blocks if we all like this approach.

gziolo:
What are the other blocks using `level`? Post Title, Site Title and similar? I agree with @youknowriad that it’s a viable approach. The best I’ve seen so far (there were a few attempts) that borrows from other existing solutions. One thing we might want to improve in the future is being more explicit for the role of the attribute so it’s clear it is metadata for the editor’s experience only. We have some of that inside `metadata` attributes but also put as top-level attribute. It’s fine that way. The only risk with top-level attributes is potential naming clash if plugins use it for something else but it’s rather unlikely if we name it specific enough.

ndiego:
> What are the other blocks using level? Post Title, Site Title and similar?

These are the blocks with the `level` attribute. 

- Comments Title
- Heading
- Title
- Query Title
- Site Title
- Site Tagline

youknowriad (approved):

ndiego:
Dev note:

## Heading level curation in WordPress 6.7

In WordPress 6.7, core blocks with a heading level dropdown [now support](https://github.com/WordPress/gutenberg/pull/63535) the new `levelOptions` attribute, which includes the Heading, Site Title, Site Tagline, Query Title, Post Title, and Comments Title blocks. The `levelOptions` attribute accepts an array of numbers corresponding to heading levels, where `1` represents H1, `2` represents H2, and so on.

This attribute allows developers to specify which heading levels should appear in the dropdown UI, providing a lightweight curation method that does not require block deprecations. Any existing heading levels are preserved in the markup, while `levelOptions` only affects the UI display.

You can apply this attribute directly in the block markup, a technique that will be commonly used in block templates, template parts, and patterns. For example, the following markup disables H1, H2, and H6 in the Heading block by setting `"levelOptions":[3,4,5]`.

```
<!-- wp:heading {"level":3,"levelOptions":[3,4,5],"className":"wp-block-heading"} -->
<h3 class="wp-block-heading">Markup example</h3>
<!-- /wp:heading -->
```

You can also use [block filters](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/) to set the default value of this attribute globally or for specific blocks. The example below disables H1, H2, and H6 for all Heading blocks. You can further customize this by restricting certain heading levels based on conditions like user capabilities.

```
function example_modify_heading_levels_globally( $args, $block_type ) {
	
	if ( 'core/heading' !== $block_type ) {
		return $args;
	}

	// Remove H1, H2, and H6.
	$args['attributes']['levelOptions']['default'] = [ 3, 4, 5 ];
	
	return $args;
}
add_filter( 'register_block_type_args', 'example_modify_heading_levels_globally', 10, 2 );
```

When either of the two examples above are applied, the Heading block UI in the Editor will look like this. 

image

---

cc @fabiankaegy, do you want me to draft this as its own dev note on Make, or do you think this will be part of a larger note?

fabiankaegy:
@ndiego I think it makes more sense to have this as part of the larger Miscellaneous Editor Changed Dev note. Not a strong opinion though if you think it deserves its own.

ndiego:
> I think it makes more sense to have this as part of the larger Miscellaneous Editor Changed Dev note. Not a strong opinion though if you think it deserves its own.

Sounds good to me. Let me know if you need anything else from me for the note.

fabiankaegy:
@ndiego From my first review it looks perfect :) I'll let @mumtahinafaguni chime in with feedback also but on first glance it is all we need :) Thank you as always! ❤️

afercia:
Spotted a small typo in the dev note draft: `depreciations` vs. `deprecations`.

ndiego:
Thanks @afercia, fixed!

linosousabox:
How change the default value of heading, for exemple change to h3
defaultHeading

Solution
https://stackoverflow.com/questions/79234401/how-change-the-default-value-of-the-heading-block-in-gutenberg

Issue #43241, opened 2022-08-16
Tracking: Addressing Design Tooling Consistency


aaronrobertshaw:
## Overview

This issue will outline and track efforts toward increasing consistency across all our blocks and their design tools. 

### Goal
1. To ensure as few gaps in design tools and block supports across blocks as possible. 
2. For any omission of support to be a deliberate and documented choice.

### Opportunities to Improve Consistency
- Adopt all block supports for all blocks (with exceptions where a deliberate choice is made and the reason why is documented)
- Move all related controls within relevant panels e.g. width or height controls moved into the Dimensions panel.
- Display the same controls by default for each block support or for each related set of blocks e.g. image-related blocks.
- Ensure all non-block-support, bespoke controls use the same components and styling
- Leverage feature level selectors or the elements API to ensure functionality across theme.json, global styles, and individual blocks.

## General Approach
An initial audit of blocks vs block supports/design tools has already been conducted to assist in identifying gaps in our support. This current state of our design tools will be outlined per block support in sub-issues tracking them.

### Phase 1 - Adopting all supports, for all blocks
This involves creating a suite of PRs to opt into any missing supports for all blocks. 

As with most things, there will be edge cases adopting various supports that might slow down the process. Given the time remaining before the 6.1 code freeze, we'll need to be fairly pragmatic here to ensure we get the design tools added in time. We can refine the UX and smooth out some edge cases as "bug fixes" after the initial beta is cut.

Low hanging fruit in this phase will be any support that can simply be opted into and works out of the box. Essentially, supports where the resulting styles can be applied to the block's wrapper. Anything that involves skipping serialization of block support styles will likely take longer to ensure compatibility with theme.json and global styles.

An approach to addressing a missing block support might be:
1. Identify a gap you'd like to fill via the individual block support tracking issues (linked in section below)
2. Search GitHub for existing PRs that might expedite or inform your efforts. 
	- Terminology used in this area is often extremely varied so you might need to get creative with your searches.
3. Create PR to adopt missing support (updating block.json, edit.js/index.php etc if skipping serialization).
4. Double check similar or related blocks and make the default controls exposed by the block support match.
5. Update the block support's tracking issue, linking the PR within the block supports table
6. Ping for review

### Phase 2 - Consistent Default Controls
Once we have all blocks adopting all block supports that are viable for them, we'll need to make which controls are shown by default as consistent as possible.

[...]

PR #63538, opened 2024-07-15
Buttons: Add border, color, and padding block supports

aaronrobertshaw:

## What?

Adopts background color, padding, and border block supports for the Buttons block.

## Why?

## How?

- Adopts new block supports

## Testing Instructions

1. In the site editor add a Buttons block to a page
2. Within Styles > Blocks > Buttons, configure new background, padding, and border values
3. Save and confirm the Buttons block is styled appropriately in the editor and frontend
4. Back in the editor select the Buttons block and switch to the block inspector
5. Configure new styles for background, padding, and border on the block instance
6. Save and confirm these display correctly in the editor and frontend

## Screenshots or screencast 

https://github.com/user-attachments/assets/a836a066-adce-4a6d-bfc2-2bc4a25eb2ea

### DIFF
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -61,7 +61,7 @@ Prompt visitors to take action with a group of button-style links. ([Source](htt
 -	**Name:** core/buttons
 -	**Category:** design
 -	**Allowed Blocks:** core/button
--	**Supports:** align (full, wide), anchor, color (background, button, gradients, ~~text~~), interactivity (clientNavigation), layout (default, ~~allowInheriting~~, ~~allowSwitching~~), spacing (blockGap, margin, padding), typography (fontSize, lineHeight), ~~html~~
+-	**Supports:** align (full, wide), anchor, color (background, gradients, ~~text~~), interactivity (clientNavigation), layout (default, ~~allowInheriting~~, ~~allowSwitching~~), spacing (blockGap, margin, padding), typography (fontSize, lineHeight), ~~html~~
 
 ## Calendar
 
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -61,7 +61,7 @@ Prompt visitors to take action with a group of button-style links. ([Source](htt
 -	**Name:** core/buttons
 -	**Category:** design
 -	**Allowed Blocks:** core/button
--	**Supports:** align (full, wide), anchor, interactivity (clientNavigation), layout (default, ~~allowInheriting~~, ~~allowSwitching~~), spacing (blockGap, margin), typography (fontSize, lineHeight), ~~html~~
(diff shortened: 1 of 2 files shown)


carolinan:
With color support enabled, the option to add color for the button element is enabled on the parent buttons block, is that correct?

aaronrobertshaw:
Thanks for testing @carolinan 👍 

> With color support enabled, the option to add color for the button element is enabled on the parent buttons block, is that correct?

This PR only proposes to add background color support to the wrapping Buttons block rather than text color for all internal buttons. With the way individual buttons are structured, I'd like to address text color for inner Button blocks separately.

carolinan:
I meant this button color support:  "color (background, **button**, gradients, ~~text~~),"
 With this option, you can set both the background and text color of the **inner** button.

aaronrobertshaw:
Sorry @carolinan, I'm not following 😅 

> I meant this button color support: "color (background, button, gradients, ~text~),"
> With this option, you can set both the background and text color of the inner button.

Is the confusion that the docs strike out the ~text~ support to show it has been opted out of?

You'll see that the text color support for the parent `Buttons` block is [disabled here](https://github.com/WordPress/gutenberg/pull/63538/files#diff-0db04c08e4f10bc82fbbadeae8160500cf34568ab4aeef81ad2c62def6261052R19).

carolinan:
No, I am finding it confusing that the parent now has **any** settings for the **inner** button.
(One of the reasons why it is confusing is the problem where the controls on the inner button do not show the inherited colors)

I only mentioned the text because I interpreted this as it should not be possible to set the text color on the inner button from the parent:

>This PR only proposes to add background color support to the wrapping Buttons block rather than text color for all internal buttons. With the way individual buttons are structured, I'd like to address text color for inner Button blocks separately.

Because with this change it is possible to set the text color of the inner button from the parent.
button-element

aaronrobertshaw:
Thanks for the clarification, I think I'm on the same page now 👍 

What's highlighted in the screenshot is actually the button element styles not inner button block controls. I'll see what I can do to reduce the potential for confusion here.

carolinan:
> With color support enabled, the option to add color for the button element is enabled on the parent buttons block, is that correct?

Yes 😄 

>With color support enabled, the option to add color for the button element is enabled on the parent buttons block, is that correct?

carolinan:
> Thanks for the clarification, I think I'm on the same page now 👍
> 
> What's highlighted in the screenshot is actually the button element styles not inner button block controls. I'll see what I can do to reduce the potential for confusion here.

Yes 😄 

>With color support enabled, the option to add color for the button element is enabled on the parent buttons block, is that correct?

aaronrobertshaw:
Apologies for my earlier confusion @carolinan. 

It was staring me in my face and I kept reading the config as `background` not `button` 🙈 

Now my tunnel vision has dissipated, let's leave the element styling out for now. While it does give users the option to bulk style all buttons it can be confusing. Discussing or implementing that as a separate follow-up sounds like the best path forward for now.

I've removed the button element support in https://github.com/WordPress/gutenberg/pull/63538/commits/4879255fac75fa5b1ca1e3dc7a60d06f4bdef335


Screenshot 2024-07-16 at 4 09 38 PM

carolinan (approved):

PR #63541, opened 2024-07-15
List Item: Add border support

aaronrobertshaw:

## What?

Adopts border support for individual list items.

## Why?

## How?

- Leaves the border controls as optional to match the dimension controls
- Defines a custom selector for borders so that the global styles only target top-level List Items

## Testing Instructions

- In the site editor, add a List block with a few items, including a nested list or two, to a page
- Style List Item blocks via Global Styles and confirm the editor and frontend display correctly
- Override the global styles on a single List Item block instance and confirm they display appropriately in the editor and frontend
- Test applying border styles to a block instance that is nested within another list

## Screenshots or screencast 

https://github.com/user-attachments/assets/4d7a8b1d-1f0c-419c-8b49-a503e892bd0e

### DIFF
--- a/packages/block-library/src/list-item/block.json
+++ b/packages/block-library/src/list-item/block.json
@@ -69,6 +69,7 @@
 		}
 	},
 	"selectors": {
-		"root": ".wp-block-list > li"
+		"root": ".wp-block-list > li",
+		"border": ".wp-block-list:not(.wp-block-list .wp-block-list) > li"
 	}
 }
--- a/packages/block-library/src/list-item/block.json
+++ b/packages/block-library/src/list-item/block.json
@@ -23,6 +23,18 @@
 		"anchor": true,
 		"className": false,
 		"splitting": true,
+		"__experimentalBorder": {
+			"color": true,
+			"radius": true,
+			"style": true,
+			"width": true,
+			"__experimentalDefaultControls": {
+				"color": true,
+				"radius": true,
+				"style": true,
+				"width": true
+			}
+		},
 		"color": {
 			"gradients": true,
 			"link": true,
--- a/packages/block-library/src/list-item/block.json
+++ b/packages/block-library/src/list-item/block.json
@@ -27,13 +27,7 @@
 			"color": true,
 			"radius": true,
 			"style": true,
-			"width": true,
-			"__experimentalDefaultControls": {
-				"color": true,
-				"radius": true,
-				"style": true,
-				"width": true
-			}
+			"width": true
 		},
 		"color": {
 			"gradients": true,


andrewserong (approved):
This one's testing well for me, too. When paired with a bunch of the other design tools, it allows for some fun styling now:

image

Similar to enabling border in the list block over in #63540, it made me wonder a tiny bit about nested list items, but overall I think it's working pretty well. Now that we can give list items a border, it's also made me wonder if it'd be worth allowing the list / list items to hide the item marker, since it's possible to use the border to visually indicate it now 🤔

I.e. a theme might style a list something like this:

image

In any case, now I'm just having fun styling lists 😄

LGTM! 🚀

aaronrobertshaw:
Thanks yet again for another thorough review @andrewserong 🙇 

I'll hold off on this one for the moment while I explore follow-ups [discussed](https://github.com/WordPress/gutenberg/pull/63540#issuecomment-2230109824) on the PR adopting border support for the primary List block.

I like the idea of being able to change the `list-style` for lists. Along with perhaps adding that to the block itself, I wonder if making lists an element in theme.json/global styles would be beneficial.

andrewserong:
> I like the idea of being able to change the list-style for lists. Along with perhaps adding that to the block itself, I wonder if making lists an element in theme.json/global styles would be beneficial.

It very well could be! If or when we explore that, it'd be worth reaching out to theme developers to see how valuable they might find it 🙂

aaronrobertshaw:
I've had a play with making global border styles only target the top-level list items. It seems to work as expected.

https://github.com/user-attachments/assets/4d7a8b1d-1f0c-419c-8b49-a503e892bd0e

Should it be desirable in a theme to style inner list items as well, I think a block style variation as suggested on the related [List block PR](https://github.com/WordPress/gutenberg/pull/63540) is a good option. A block style variation would also allow the inner list items to have different borders to the parent. 

As with the [parent List block](https://github.com/WordPress/gutenberg/pull/63540), I'm leaning towards this latest approach as the most intuitive but I'd like to hear any thoughts on the matter.

andrewserong (approved):
Thanks for your patience @aaronrobertshaw, I missed your update from last week! This is looking good to me, and follows how I'd expect this to work as a user 👍

> As with the https://github.com/WordPress/gutenberg/pull/63540, I'm leaning towards this latest approach as the most intuitive but I'd like to hear any thoughts on the matter.

Me too, I think this makes for a nice feature in its current form. I also like that the selector being used is specific to border, so we're not affecting other styles that folks might be applying to the list item block.

✅ Setting borders for the list item in global styles now only applies to root list items
✅ Overriding at the individual block level works as expected
✅ Individual list items and nested list items can still use block-level styles for borders
✅ Tested in the site editor, site frontend, and post editors

image

LGTM! 🚀

aaronrobertshaw:
> Thanks for your patience @aaronrobertshaw, I missed your update from last week! 

I'm always happy to wait for your thorough reviews 🙂 

Thanks for putting this through its paces again. Let's get it merged and see what creations users come up with!

PR #63542, opened 2024-07-15
Media Text: Add border support

aaronrobertshaw:

## What?

Adopts border support for Media and Text block.

## Why?

## How?

## Testing Instructions

- In the site editor, add a Media Text block to a page

## Screenshots or screencast 

https://github.com/user-attachments/assets/5cc14b84-a866-4a2a-ae4d-196c563966a9

### DIFF
--- a/packages/block-library/src/media-text/block.json
+++ b/packages/block-library/src/media-text/block.json
@@ -103,6 +103,18 @@
 		"anchor": true,
 		"align": [ "wide", "full" ],
 		"html": false,
+		"__experimentalBorder": {
+			"color": true,
+			"radius": true,
+			"style": true,
+			"width": true,
+			"__experimentalDefaultControls": {
+				"color": true,
+				"radius": true,
+				"style": true,
+				"width": true
+			}
+		},
 		"color": {
 			"gradients": true,
 			"heading": true,


andrewserong (approved):
This is testing well for me in the post and site editors, in global styles, and in overrides at the individual block level 👍

Similar to the discussion in adding the border support to the Gallery block in #63428, I notice that the Media & Text block does not clip the image when we set a heavy border radius:

image

Since the consensus over in https://github.com/WordPress/gutenberg/pull/63428#issuecomment-2227963763 was to allow border radius even if the block doesn't clip, I don't think it's an issue here, particularly as padding can be set on this block.

LGTM! 🚀

aaronrobertshaw:
> Since the consensus over in https://github.com/WordPress/gutenberg/pull/63428#issuecomment-2227963763 was to allow border-radius even if the block doesn't clip, I don't think it's an issue here, particularly as padding can be set on this block.

A possible follow-up could be to explore inheriting border-radius on the inner image but combined with the ability to add padding to the media and text block, I'm not sure how well it would work in practice.

Ultimately, I think refactoring this block to use an inner Image block, is our best bet. It would then allow optional control over the inner image border-radius.

As per the linked consensus, let's lean into greater design freedom and merge this 👍

PR #47282, opened 2023-01-19
Add CSS class to paragraph block

juanfra:
Addition by @t-hamano

To resume this PR, I did the following:

- Merge the latest trunk into this branch
- Resolve all conflicts
- Rechecked the whole code and added the missing paragraph block class

This PR has a lot of diffs, mostly related to testing and documentation updates.

The actual files to check are:

- [`packages/block-library/src/paragraph/block.json`](https://github.com/WordPress/gutenberg/pull/47282/files#diff-9f1fb6109b92a590a1f86ce7f3064fa1d68e32bf998dec23bca649fcf8079c4d): The paragraph block will have the class name `wp-block-paragraph` added to it, and the CSS selectors generated by the global styles will also use this class.
- [`packages/block-library/src/paragraph/index.php`](https://github.com/WordPress/gutenberg/pull/47282/files#diff-1b6714956e899982e20508de0f78300f34781863bf057320be50997dc967511b): Add a class name to the Paragraph block that is already saved in the database via the render callback function.
- The unit tests below also exist in the core, so similar changes will need to be made there as well, probably after this PR is bundled into the core:
  - [`phpunit/block-supports/layout-test.php`](https://github.com/WordPress/gutenberg/pull/47282/files#diff-eedf0832f35d37ba18b396f84e65b829d9ed4d51f2a5057b63f5461b754254b7)
  - [`phpunit/block-supports/typography-test.php`](https://github.com/WordPress/gutenberg/pull/47282/files#diff-20b3805f117ee384e9d7cfc5861fb5059a97c6dc6615e15e60ad8694374260bf)
  - [`phpunit/class-wp-theme-json-test.php`](https://github.com/WordPress/gutenberg/pull/47282/files#diff-88c997ea8e8ba51f2fa201327e0f1faccd4e892c2fafe33249f72cfd068c72fc)
- The following unit tests do not exist in the core, so they do not need to be backported to the core:
  - [`phpunit/blocks/render-block-media-text-test.php`](phpunit/blocks/render-block-media-text-test.php)
  - [`phpunit/blocks/render-block-paragraph-test.php`](phpunit/blocks/render-block-paragraph-test.php)

---

Original description by @juanfra 

## What?

Adding the `wp-block-paragraph` to the paragraph block elements.

Fixes https://github.com/WordPress/gutenberg/issues/46863

## Why?

[...]

PR #63543, opened 2024-07-15
Paragraph: Add border support

aaronrobertshaw:

## What?

Adopts border support for Paragraph block.

## Why?

## How?

## Testing Instructions

- In the site editor, add a paragraph block to a page

## Screenshots or screencast 

https://github.com/user-attachments/assets/33363f23-e91d-4702-a014-d7e22ba9497d

### DIFF
--- a/packages/block-library/src/paragraph/block.json
+++ b/packages/block-library/src/paragraph/block.json
@@ -37,13 +37,7 @@
 			"color": true,
 			"radius": true,
 			"style": true,
-			"width": true,
-			"__experimentalDefaultControls": {
-				"color": true,
-				"radius": true,
-				"style": true,
-				"width": true
-			}
+			"width": true
 		},
 		"color": {
 			"gradients": true,
--- a/packages/block-library/src/paragraph/block.json
+++ b/packages/block-library/src/paragraph/block.json
@@ -33,6 +33,18 @@
 		"splitting": true,
 		"anchor": true,
 		"className": false,
+		"__experimentalBorder": {
+			"color": true,
+			"radius": true,
+			"style": true,
+			"width": true,
+			"__experimentalDefaultControls": {
+				"color": true,
+				"radius": true,
+				"style": true,
+				"width": true
+			}
+		},
 		"color": {
 			"gradients": true,
 			"link": true,


andrewserong (approved):
I went back and forth a little bit on the idea of exposing border controls for individual paragraph blocks, but overall the more I play with this, the more I like it!

My gut feeling is that in the vast majority of cases, folks won't be setting all paragraphs to have a border style of any kind in global styles, but it gets pretty fun once you start including it in block style variations. Here's a very quick demo of playing around adding a couple of style variations for notice-like paragraphs:

https://github.com/user-attachments/assets/b456a9aa-5010-463a-b462-cf4fd11046a0

<details>

<summary>A couple of block style variations for notices</summary>

```html
{
	"version": 3,
	"slug": "notice-warning",
    "title": "Notice Warning",
	"blockTypes": [ "core/group", "core/paragraph" ],
	"styles": {
        "border": {
            "left": {
                "color": "red",
                "style": "solid",
                "width": "5px"
            }
        },        
		"color": {
            "background": "#fff0f0",
            "text": "#333"
		},
        "spacing": {
            "padding": {
                "top": "20px",
                "right": "20px",
                "bottom": "20px",
                "left": "20px"
            }
        }
    }
}
```

```html
{
	"version": 3,
	"slug": "notice-success",
    "title": "Notice Success",
	"blockTypes": [ "core/group", "core/paragraph" ],
	"styles": {
        "border": {
            "left": {
                "color": "green",
                "style": "solid",
                "width": "5px"
            }
        },        
		"color": {
            "background": "#f0fff0",
            "text": "#333"
		},
        "spacing": {
            "padding": {
                "top": "20px",
                "right": "20px",
                "bottom": "20px",
                "left": "20px"
            }
        }
    }
}
```

</details>

Depending on the theme, it might be easier for folks to have these sorts of styles at the click of a button rather than users needing to wrap their paragraphs in Group blocks to achieve a similar result.

So this gets a 👍 vote from me. Since enabling any controls in the paragraph block makes them quite visible as it's such a common block, I'll just ping @WordPress/gutenberg-design for visibility on this one.

But particularly with `theme.json`-shaped block style variations now being supported in core, I think including border support for paragraphs is a great idea. And technically it's working well for me (works in global styles in the site editor, in the post editor with overrides at the instance level, and in the site frontend).

LGTM!

aaronrobertshaw:
Thanks for the review @andrewserong, bonus points for playing with block style variations! 🥇 

I had the same initial reaction regarding adopting border support on paragraphs but quickly came around as well.

A few points worth listing are:
- Not requiring users to wrap everything in a group block to get the design tooling they want is a win.
- There's a benefit to keeping the display of border controls by default the same across blocks. Without that consistency, it can be confusing for some users as what is default for one block suddenly disappears and looks to be missing on another.
- If the display of border controls are okay for other common blocks like the Group block, they should be okay here. If they aren't, perhaps there's an opportunity to improve them across the board

Just my two cents of course 🙂

paaljoachim:
It is great to see the consistency steam engine rolling on to add additional design tools to additional blocks! Thank you @aaronrobertshaw and @andrewserong 

It looks good in local Paragraph blocks as well as global Paragraph blocks!

carolinan:
I would prefer if no more style support were added before the paragraph block has its wp-block class. Because just like the list block before it, it will affect all paragraphs, not only the paragraph block.

For example if I set a border on the paragraph block using theme.json, it affects the paragraphs used around the textarea and button in the comments/comment form blocks.

jameskoster:
I appreciate how this increases design flexibility, it's annoying having to wrap a paragraph in a Group to achieve this.

However it is fairly niche. I wonder if the controls should be hidden by default, similar to the Dimensions panel?

richtabor:
> However it is fairly niche. I wonder if the controls should be hidden by default, similar to the Dimensions panel?

I was going to suggest the same. To set them as hidden by default, where you press the + to add a border or radius.

aaronrobertshaw:
> I would prefer if no more style support were added before the paragraph block has its wp-block class.

Thanks for the feedback @carolinan 👍 

It's a little confusing but the adoption of block support in block.json for block instances is actually separate from being able to style certain features (color, typography, border etc) via theme.json. You can already style any block with any feature in theme.json.

Using trunk, add the following to your theme.json and you'll see the border:

```json
	"styles": {
		"blocks": {
			"core/paragraph": {
				"border": {
					"width": "1em",
					"color": "blue",
					"style": "solid"
				}
			}
		}
	}
```


Long story short, I don't think this PR really accentuates the issue you raised.

> However it is fairly niche. I wonder if the controls should be hidden by default, similar to the Dimensions panel?

I've made that change in https://github.com/WordPress/gutenberg/pull/63543/commits/ffe8ce7bbe09c538d796ee7051061a8a0b0a5c49


https://github.com/user-attachments/assets/061073b4-6775-44a1-8d30-a5e1a7573e21


With that latest tweak, any objections to landing this one?

andrewserong:
> With that latest tweak, any objections to landing this one?

None from me, I like how tucked away it is now 👍

https://github.com/user-attachments/assets/9cb5079a-17d8-41c9-99b1-e55b8c427d5f

aaronrobertshaw:
> None from me, I like how tucked away it is now 👍

Thanks, I'll get this merged then.

Regarding the theme.json styling of paragraph blocks and its element-only selector, I'll happily help out on that as a follow-up.

carolinan:
I would like to see a dedicated effort to getting the class name added to the block. The PR has been open for over a year without a final decision https://github.com/WordPress/gutenberg/pull/47282

I don't think that adding new features knowing that they are not working as intended is a good idea.

aaronrobertshaw:
> I would like to see a dedicated effort to getting the class name added to the block

No arguments here 🙂 

> I don't think that adding new features knowing that they are not working as intended is a good idea.

I agree with not landing new features if we know they aren't working. That's part of what I was trying to explain in my last reply, the adoption of the border block support, such as in this PR, is specific to a block instance. It does work as intended.

The ability to style borders through theme.json isn't being introduced by this PR and already exists. As you note, once we can get a class name injected into the paragraph block, like heading blocks had done, the theme.json issue will be resolved. It is outside the scope of this PR though.

PR #63545, opened 2024-07-15
Quote: Add spacing supports

aaronrobertshaw:

## What?

Adopts padding and margin support for the Quote block.

## Why?

## How?

- Adopts padding and margin block supports

## Testing Instructions

- In the site editor, add a Quote block to a page

Note: Global margins will be overridden by layout whereas block instance margins will not be. There is also an unrelated bug with the spacing visualizers that is being addressed separately.

## Screenshots or screencast 

https://github.com/user-attachments/assets/10a31bd9-d134-4d7d-be44-3d02a24926ec


fabiankaegy (approved):
🚀 finally 🤩 Tested this and it works as described  👍

PR #63546, opened 2024-07-15
Image: Adopt margin block support

aaronrobertshaw:

## What?

- Adopts margin support for the Image block.
- Padding support has been deliberately omitted as it doesn't make a lot of sense on the image block.

## Why?

## How?

- Adopts margin block support

## Testing Instructions

- In the site editor, add an Image block to a page
- Back in the editor, select the image with margin applied
- Via the block toolbar's link button, select the "Expand on click" option so the image can be opened in a lightbox on the frontend
- Save the post and confirm the animation of the lightbox is smooth and doesn't have the margin applied.

Note: Global margins will be overridden by layout whereas block instance margins will not be.

## Screenshots or screencast 

https://github.com/user-attachments/assets/472e6ac4-99de-4120-ab13-077b5c1039c0

### DIFF
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -387,7 +387,7 @@ Insert an image to make a visual statement. ([Source](https://github.com/WordPre
 
 -	**Name:** core/image
 -	**Category:** media
--	**Supports:** align (center, full, left, right, wide), anchor, color (~~background~~, ~~text~~), filter (duotone), interactivity, shadow ()
+-	**Supports:** align (center, full, left, right, wide), anchor, color (~~background~~, ~~text~~), filter (duotone), interactivity, shadow (), spacing (margin)
 -	**Attributes:** alt, aspectRatio, blob, caption, height, href, id, lightbox, linkClass, linkDestination, linkTarget, rel, scale, sizeSlug, title, url, width
 
 ## Latest Comments
--- a/packages/block-library/src/image/block.json
+++ b/packages/block-library/src/image/block.json
@@ -106,6 +106,9 @@
 		"filter": {
 			"duotone": true
 		},
+		"spacing": {
+			"margin": true
+		},
 		"__experimentalBorder": {
 			"color": true,
 			"radius": true,
--- a/packages/block-library/src/image/index.php
+++ b/packages/block-library/src/image/index.php
@@ -291,12 +291,12 @@ function block_core_image_print_lightbox_overlay() {
 					<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="20" height="20" aria-hidden="true" focusable="false"><path d="M13 11.8l6.1-6.3-1-1-6.1 6.2-6.1-6.2-1 1 6.1 6.3-6.5 6.7 1 1 6.5-6.6 6.5 6.6 1-1z"></path></svg>
 				</button>
 				<div class="lightbox-image-container">
-					<figure data-wp-bind--class="state.currentImage.figureClassNames" data-wp-bind--style="state.currentImage.figureStyles">
+					<figure data-wp-bind--class="state.currentImage.figureClassNames" data-wp-bind--style="state.figureStyles">
 						<img data-wp-bind--alt="state.currentImage.alt" data-wp-bind--class="state.currentImage.imgClassNames" data-wp-bind--style="state.imgStyles" data-wp-bind--src="state.currentImage.currentSrc">
 					</figure>
 				</div>
 				<div class="lightbox-image-container">
-					<figure data-wp-bind--class="state.currentImage.figureClassNames" data-wp-bind--style="state.currentImage.figureStyles">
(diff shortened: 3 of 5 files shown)


fabiankaegy (approved):
Very happy to see this 🚀

artemiomorales (changes requested):
I tested adding a margin to an image aligned to the right with the lightbox enabled, and the margin causes the lightbox to not animate smoothly:

https://github.com/user-attachments/assets/6dd5e7c3-e799-4079-9942-9ce39f706e7e

The added margin is likely throwing off the lightbox calculations; likely we should aim to take those into account.

artemiomorales:
The animation is also affected with center-aligned images that have a margin enabled:

https://github.com/user-attachments/assets/20fcef46-a1de-4ccc-9f45-b1760376cac9

We should also take into account using different aspect ratios, custom sizing, different resolutions, both scale options `cover` and `contain`, and border.

Screenshot 2024-07-15 at 9 04 07 AM

Regarding border, I see the lightbox animation is actually broken on `trunk` when a border is defined. I mention it here because it's related, but I've [opened a separate issue](https://github.com/WordPress/gutenberg/issues/63567) for that.

https://github.com/user-attachments/assets/a43987e0-6caa-4c46-8f9c-ddff0a57d1bc

My guess is the [setOverlayStyles callback](https://github.com/WordPress/gutenberg/blob/c54c0020a7a8ebf22738aa96cb398ea4cf17a30a/packages/block-library/src/image/view.js#L168) would need to modified to make the margin support compatible with the lightbox.

aaronrobertshaw:
Thanks for reviewing @artemiomorales 👍 

I honestly missed testing image lightbox functionality when adding the margin support here 😅 

From what you've outlined though it sounds like there is a more general problem with lightbox not taking into account image styles that affect spacing such as margin and border, than a problem with the application of margins on an image.

It is worth noting that you can already define margins for images via theme.json. Does this not also cause a problem or is the issue that block support styles are applied as inline styles and override lightbox resets etc?

Do you think this issue needs to block the adoption of margin support for all images? Can it be addressed separately via 
https://github.com/WordPress/gutenberg/issues/63567?

artemiomorales:
> It is worth noting that you can already define margins for images via theme.json. Does this not also cause a problem or is the issue that block support styles are applied as inline styles and override lightbox resets etc?

@aaronrobertshaw It appears that the lightbox works as expected when defining the margin via theme.json or in the global styles — the issue, as you've pointed out, is that the styles are applied inline at the block level, notably to the parent `figure` element.

> Do you think this issue needs to block the adoption of margin support for all images? Can it be addressed separately via
https://github.com/WordPress/gutenberg/issues/63567?

I think so, otherwise anyone who adds a margin to an image with `Expand on Click` will see a broken experience, which could reduce confidence in the feature. I'm open to hearing other opinions on that, though. In any case, I've started looking at the lightbox code to see how we could resolve it.

Would it be alright if I worked on that this week as part of this PR? I'm happy to hear other ideas too, let me know what you think 🙏

artemiomorales (approved):
@aaronrobertshaw It turns out that the lightbox copies the figure's inline styles in the overlay, and the new margins were throwing off the calculations. I pushed a commit to remove the margins from the overlay, and it seems to work properly now.

I can do another smoke test tomorrow but will go ahead and approve so you can merge if you're feeling confident with all of the use cases 🙏

aaronrobertshaw:
@artemiomorales are there unit tests for the lightbox behaviour that can be updated to protect against regressions?

artemiomorales:
> are there unit tests for the lightbox behaviour that can be updated to protect against regressions?

@aaronrobertshaw There were but they need to be redone. I can take a look at that tomorrow, as part of this PR or a separate one, whichever you prefer.

artemiomorales:
@aaronrobertshaw Ok I added a test to ensure the margin inline styles don't get applied to the overlay image 👍

aaronrobertshaw:
Really appreciate the help getting the lightbox behaviour sorted @artemiomorales, thanks 👍 

I've given this PR a run both before and after the latest updates. The animation looks nice and smooth again.

https://github.com/user-attachments/assets/6089fe5e-00ce-4f6e-8acc-dccbc357a1e5

andrewserong (approved):
I'm not very familiar with the lightbox code and it took me a little while to find the control for it (is the only place it's exposed in global styles > blocks > image?). But it's working nicely!

✅ Image margin set in global styles is output correctly in post and site editors and on the site frontend
✅ Block level instance overrides work correctly
✅ In manual testing the lightbox feature still appears to be working as expected AFAIK

LGTM! 🚀

aaronrobertshaw:
Thanks for the review @andrewserong 👍 

> is the only place it's exposed in global styles > blocks > image?

There is also the link block toolbar button. Under that you'll have the option to choose "Expand on click".

I agree though, the discoverability of it isn't fantastic. It took me a while and I knew it was there and was actively looking for it.

andrewserong:
> There is also the link block toolbar button. Under that you'll have the option to choose "Expand on click".

Thank you! That was a TIL for me, I didn't realise it was under there, I was looking at the inspector sidebar. Cheers!

PR #63547, opened 2024-07-15
Search: Add margin support

aaronrobertshaw:

## What?

- Adopts margin support for the Search block

## Why?

## How?

- Adopts margin block support

## Testing Instructions

- In the site editor, add a Search block to a page

**Note: Global margins will be overridden by layout whereas block instance margins will not be. There is also an unrelated bug with the spacing visualizers that is being addressed separately.**

## Screenshots or screencast 

https://github.com/user-attachments/assets/957a6ea0-6d49-48cd-b41f-28fa8764da49

### DIFF
--- a/docs/reference-guides/core-blocks.md
+++ b/docs/reference-guides/core-blocks.md
@@ -810,7 +810,7 @@ Help visitors find your content. ([Source](https://github.com/WordPress/gutenber
 
 -	**Name:** core/search
 -	**Category:** widgets
--	**Supports:** align (center, left, right), color (background, gradients, text), interactivity, typography (fontSize, lineHeight), ~~html~~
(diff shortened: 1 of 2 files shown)


fabiankaegy (approved):
Tests well :) Excited about the consistency

carolinan:
I am wondering why padding was not added?

carolinan (approved):

aaronrobertshaw:
> I am wondering why padding was not added?

Good question, @carolinan 👍 

The current color block supports aren't serialized onto the block's wrapper but are instead passed through to the inner button. I wasn't sure how tightly coupled the use case of applying padding is to the ability to define a background color for the entire block.

This PR is a simple step forward that doesn't have to be bogged down around that. We could well decide to just add padding support, for now though I think that can be a follow-up.

PR #63040, opened 2024-07-02
FontSizePicker: add pickerMode

ciampo:
## What?

Following [this conversation](https://github.com/WordPress/gutenberg/pull/62328#discussion_r1659012795), this PR introduces a new `pickerMode` prop that is meant to supersede the existing `disableCustomFontSizes` prop (which gets marked as deprecated).

## Why?

With the `disableCustomFontSizes` prop, consumers of the component can choose to prevent users to enter custom font size values, forcing them to pick from the list of predefined font sizes.

But there isn't a way to achieve the opposite — ie. allow users to only pick from a list of custom values, without being able to choose from a list of predef options.

The `pickerMode` prop is a enum of different modes:

- `both` maps to `disableCustomFontSizes = false`
- `predefined` maps to `disableCustomFontSizes = true`
- `custom` introduces the missing mode discussed above.

## How?

I refactored the code and extracted some logic away from the render function, to bring more clarity to it.

Together with the `pickerMode` prop, I also introduced the idea of `pickerType` (see [this comment](https://github.com/WordPress/gutenberg/pull/63040#discussion_r1662016329)).

I also refactored tests and usages in Gutenberg to already use the new `pickerMode` prop.

TODO: 
- unit tests
- add documentation about the consumer being responsible for any validation to prevent weird UI states when custom values are used with `pickerMode="predefined"`

## Testing Instructions

_Note: I recommend enabling the "Hide whitespace changes" options when reviewing the code changes in this PR._

TBD

PR #63553, opened 2024-07-15
FontSizePicker: tidy up internal logic

ciampo:
## What?

Extract some of the changes from #63040 to tidy up `FontSizePicker`'s internal logic

## Why?

The component's internal logic for which UI to show is based on several boolean flags spread across the file, which makes it difficult to read/understand and bug-prone.

## How?

This reactor extracts the logic to a few utility functions and condenses the UI state into one internal state variable.

## Testing Instructions

- Make sure that the unit tests (including the ones added in this PR) also pass both on `trunk` and on this PR
- Test `FontSizePicker` in Storybook, especially around the props that influence what UI is presented to the user:
  - `fontSizes` (>= 5 font sizes => `CustomSelectControl`,  < 5 font sizes => `ToggleGroupControl`)
  - `disableCustomFontSizes` should disable the custom UI and hide the toggle
  - experiment with changing those props while different UIs are presented
- Smoke test in the editor

> [!NOTE]
> The new PR introduces a slight behavioral change to the component, which is IMO an improvement. Here's how to reproduce:
> 1. Load the default `FontSizePicker` Storybook example
> 2. Show the custom value UI by clicking on the toggle in the top-right corner
> 3. Change the `disableCustomFontSizes` prop to `true`
> 4. On `trunk`, no UI is presented to the user. On this PR, either `ToggleGroupControl` or `CustomSelectControl` UIs are shows based on how many font sizes the user can pick from.

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -17,6 +17,7 @@
 -   `ToggleGroupControl`: support disabled options ([#63450](https://github.com/WordPress/gutenberg/pull/63450)).
 -   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
 -   `SelectControl`: Add `"minimal"` variant ([#63265](https://github.com/WordPress/gutenberg/pull/63265)).
+-   `FontSizePicker`: tidy up internal logic ([#63553](https://github.com/WordPress/gutenberg/pull/63553)).
 
 ### Internal
 
--- a/packages/components/src/font-size-picker/index.tsx
+++ b/packages/components/src/font-size-picker/index.tsx
@@ -45,19 +45,6 @@ const DEFAULT_UNITS = [ 'px', 'em', 'rem', 'vw', 'vh' ];
 const shouldUseSelectOverToggle = ( howManyfontSizes: number ) =>
 	howManyfontSizes > 5;
 
-const getPickerType = (
-	showCustomValuePicker: boolean,
-	fontSizes: FontSize[]
-): FontSizePickerType => {
-	if ( showCustomValuePicker ) {
-		return 'custom';
-	}
-
-	return shouldUseSelectOverToggle( fontSizes.length )
-		? 'select'
-		: 'togglegroup';
-};
-
 const getHeaderHint = (
 	currentPickerType: FontSizePickerType,
 	selectedFontSize: FontSize | undefined,
@@ -117,14 +104,17 @@ const UnforwardedFontSizePicker = (
 
 	if ( fontSizes.length === 0 && disableCustomFontSizes ) {
 		return null;
+	let currentPickerType;
+	if ( ! disableCustomFontSizes && userRequestedCustom ) {
+		// While showing the custom value picker, switch back to predef only if
+		// `disableCustomFontSizes` is set to `true`.
+		currentPickerType = 'custom' as const;
+	} else {
+		currentPickerType = shouldUseSelectOverToggle( fontSizes.length )
+			? ( 'select' as const )
+			: ( 'togglegroup' as const );
 	}
 
-	const currentPickerType = getPickerType(
-		// If showing the custom value picker, switch back to predef only
-		// if `disableCustomFontSizes` is set to `true`.
-		! disableCustomFontSizes && userRequestedCustom,
-		fontSizes
-	);
 
 	const headerHint = getHeaderHint(
 		currentPickerType,
--- a/packages/components/src/font-size-picker/index.tsx
+++ b/packages/components/src/font-size-picker/index.tsx
@@ -8,7 +8,7 @@ import type { ForwardedRef } from 'react';
  */
 import { __ } from '@wordpress/i18n';
 import { settings } from '@wordpress/icons';
-import { useState, forwardRef } from '@wordpress/element';
+import { useState, useMemo, forwardRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -23,11 +23,7 @@ import {
 } from '../unit-control';
 import { VisuallyHidden } from '../visually-hidden';
 import { getCommonSizeUnit } from './utils';
-import type {
-	FontSize,
-	FontSizePickerProps,
-	FontSizePickerType,
-} from './types';
+import type { FontSizePickerProps } from './types';
 import {
 	Container,
 	Header,
@@ -42,35 +38,7 @@ import { T_SHIRT_NAMES } from './constants';
 
 const DEFAULT_UNITS = [ 'px', 'em', 'rem', 'vw', 'vh' ];
 
-const shouldUseSelectOverToggle = ( howManyfontSizes: number ) =>
-	howManyfontSizes > 5;
-
-const getHeaderHint = (
-	currentPickerType: FontSizePickerType,
-	selectedFontSize: FontSize | undefined,
-	fontSizes: FontSize[]
-) => {
-	if ( currentPickerType === 'custom' ) {
-		return __( 'Custom' );
-	}
-
-	if ( ! shouldUseSelectOverToggle( fontSizes.length ) ) {
-		if ( selectedFontSize ) {
-			return (
-				selectedFontSize.name ||
-				T_SHIRT_NAMES[ fontSizes.indexOf( selectedFontSize ) ]
-			);
-		}
-		return '';
-	}
-
-	const commonUnit = getCommonSizeUnit( fontSizes );
-	if ( commonUnit ) {
-		return `(${ commonUnit })`;
-	}
-
-	return '';
-};
+const MAX_TOGGLE_GROUP_SIZES = 5;
 
 const UnforwardedFontSizePicker = (
 	props: FontSizePickerProps,
@@ -102,25 +70,44 @@ const UnforwardedFontSizePicker = (
 		availableUnits: unitsProp,
 	} );
 
-	if ( fontSizes.length === 0 && disableCustomFontSizes ) {
-		return null;
 	let currentPickerType;
 	if ( ! disableCustomFontSizes && userRequestedCustom ) {
 		// While showing the custom value picker, switch back to predef only if
 		// `disableCustomFontSizes` is set to `true`.
 		currentPickerType = 'custom' as const;
 	} else {
-		currentPickerType = shouldUseSelectOverToggle( fontSizes.length )
-			? ( 'select' as const )
-			: ( 'togglegroup' as const );
+		currentPickerType =
+			fontSizes.length > MAX_TOGGLE_GROUP_SIZES
+				? ( 'select' as const )
+				: ( 'togglegroup' as const );
 	}
 
+	const headerHint = useMemo( () => {
+		switch ( currentPickerType ) {
+			case 'custom':
+				return __( 'Custom' );
+			case 'togglegroup':
+				if ( selectedFontSize ) {
+					return (
+						selectedFontSize.name ||
+						T_SHIRT_NAMES[ fontSizes.indexOf( selectedFontSize ) ]
+					);
+				}
+				break;
+			case 'select':
+				const commonUnit = getCommonSizeUnit( fontSizes );
+				if ( commonUnit ) {
+					return `(${ commonUnit })`;
+				}
+				break;
+		}
 
-	const headerHint = getHeaderHint(
-		currentPickerType,
-		selectedFontSize,
-		fontSizes
-	);
+		return '';
+	}, [ currentPickerType, selectedFontSize, fontSizes ] );
+
+	if ( fontSizes.length === 0 && disableCustomFontSizes ) {
+		return null;
+	}
 
 	// If neither the value or first font size is a string, then FontSizePicker
 	// operates in a legacy "unitless" mode where UnitControl can only be used
--- a/packages/components/src/font-size-picker/index.tsx
+++ b/packages/components/src/font-size-picker/index.tsx
@@ -8,7 +8,7 @@ import type { ForwardedRef } from 'react';
  */
 import { __ } from '@wordpress/i18n';
 import { settings } from '@wordpress/icons';
-import { useState, useEffect, forwardRef } from '@wordpress/element';
+import { useState, forwardRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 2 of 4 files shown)


packages/components/src/font-size-picker/test/index.tsx
```diff
@@ -525,6 +525,39 @@ describe( 'FontSizePicker', () => {
...
+			rerender(
+				<FontSizePicker
+					disableCustomFontSizes
+					fontSizes={ fontSizes }
+					value="3px"
+				/>
+			);
+			expect(
+				screen.queryByLabelText( 'Custom' )
+			).not.toBeInTheDocument();
```
ciampo, line 558: I didn't write a more specific assertion on which UI is presented to the user since I initially wrote this unit test so that it would pass on `trunk`, where no input UI is actually presented to the user in this case (a bug IMO — see a more extended explanation in the PR description).


Since the input UI presented to the user changes depending on the length of the `fontSizes` prop, if we think that we should test that part too, I will duplicate this test and add it to two different sections of the file

packages/components/CHANGELOG.md
```diff
@@ -6,17 +6,18 @@
 
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
 
+### Enhancements
```
ciampo: Moved the `Enhancements` section before the `Internal` section

packages/components/src/font-size-picker/types.ts
```diff
@@ -29,24 +31,25 @@ export type FontSizePickerProps = {
 	/**
 	 * Available units for custom font size selection.
 	 *
-	 * @default `[ 'px', 'em', 'rem' ]`
+	 * @default [ 'px', 'em', 'rem' ]
```
mirka: I noticed this was out of date, fix at #63577.

mirka (commented):
I couldn't help but think the `useEffect` is avoidable 🤔 How I model this in my brain, there is only one bit of state ("is the user requesting a custom input"), and the `currentPickerType` can be derived without a `useEffect`/`useState`.

```diff
diff --git a/packages/components/src/font-size-picker/index.tsx b/packages/components/src/font-size-picker/index.tsx
index a547a26de5..8f5fba2be0 100644
--- a/packages/components/src/font-size-picker/index.tsx
+++ b/packages/components/src/font-size-picker/index.tsx
@@ -8,7 +8,7 @@ import type { ForwardedRef } from 'react';
  */
 import { __ } from '@wordpress/i18n';
 import { settings } from '@wordpress/icons';
-import { useState, useEffect, forwardRef } from '@wordpress/element';
+import { useState, forwardRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -107,27 +107,7 @@ const UnforwardedFontSizePicker = (
 	);
 	const isCustomValue = !! value && ! selectedFontSize;
 
-	const [ currentPickerType, setCurrentPickerType ] = useState(
-		getPickerType( ! disableCustomFontSizes && isCustomValue, fontSizes )
-	);
-
-	useEffect( () => {
-		setCurrentPickerType(
-			getPickerType(
-				// If showing the custom value picker, switch back to predef only
-				// if `disableCustomFontSizes` is set to `true`.
-				currentPickerType === 'custom'
-					? ! disableCustomFontSizes
-					: ! disableCustomFontSizes && isCustomValue,
-				fontSizes
-			)
-		);
-	}, [
-		currentPickerType,
-		disableCustomFontSizes,
-		isCustomValue,
-		fontSizes,
-	] );
+	const [ userRequestedCustom, setUserRequestedCustom ] = useState( false );
 
 	const units = useCustomUnits( {
 		availableUnits: unitsProp,
@@ -137,6 +117,13 @@ const UnforwardedFontSizePicker = (
 		return null;
 	}
 
+	const currentPickerType = getPickerType(
+		// If showing the custom value picker, switch back to predef only
+		// if `disableCustomFontSizes` is set to `true`.
+		! disableCustomFontSizes && ( userRequestedCustom || isCustomValue ),
+		fontSizes
+	);
+
 	const headerHint = getHeaderHint(
 		currentPickerType,
 		selectedFontSize,
@@ -180,14 +167,9 @@ const UnforwardedFontSizePicker = (
 									: __( 'Set custom size' )
 							}
 							icon={ settings }
-							onClick={ () => {
-								setCurrentPickerType(
-									getPickerType(
-										currentPickerType !== 'custom',
-										fontSizes
-									)
-								);
-							} }
+							onClick={ () =>
+								setUserRequestedCustom( ! userRequestedCustom )
+							}
 							isPressed={ currentPickerType === 'custom' }
 							size="small"
 						/>
@@ -215,9 +197,7 @@ const UnforwardedFontSizePicker = (
 								);
 							}
 						} }
-						onSelectCustom={ () =>
-							setCurrentPickerType( 'custom' )
-						}
+						onSelectCustom={ () => setUserRequestedCustom( true ) }
 					/>
 				) }
 				{ currentPickerType === 'togglegroup' && (
@@ -251,6 +231,8 @@ const UnforwardedFontSizePicker = (
 								hideLabelFromVision
 								value={ value }
 								onChange={ ( newValue ) => {
+									setUserRequestedCustom( true );
+
 									if ( newValue === undefined ) {
 										onChange?.( undefined );
 									} else {
@@ -281,6 +263,8 @@ const UnforwardedFontSizePicker = (
 										initialPosition={ fallbackFontSize }
 										withInputField={ false }
 										onChange={ ( newValue ) => {
+											setUserRequestedCustom( true );
+
 											if ( newValue === undefined ) {
 												onChange?.( undefined );
 											} else if ( hasUnits ) {
```

This will fail the "doesn't hide custom input when the selected font size is a predef" tests. However, I'm not sure this behavior is necessarily wrong? If the user never requested or interacted with the custom inputs, maybe it is more natural to fall back to the predef pickers if the value changed programatically. What do you think?

packages/components/src/font-size-picker/index.tsx
```diff
@@ -133,53 +175,52 @@ const UnforwardedFontSizePicker = (
...
 									: __( 'Set custom size' )
 							}
 							icon={ settings }
 							onClick={ () => {
-								setShowCustomValueControl(
-									! showCustomValueControl
+								setCurrentPickerType(
+									getPickerType(
+										currentPickerType !== 'custom',
+										fontSizes
```
tyxla: If we moved the `getPickerType` and `getHeaderHint` inside the component function, we wouldn't need to pass `fontSizes` all over the place. We could also potentially reuse some of the conditions (like whether the current picker type is `custom` for example), or even remove some of the utilities (like `shouldUseSelectOverToggle()`). Maybe that could reduce some of the complexity?

packages/components/src/font-size-picker/index.tsx
```diff
@@ -38,6 +42,49 @@ import { T_SHIRT_NAMES } from './constants';
...
+	}
+
+	if ( ! shouldUseSelectOverToggle( fontSizes.length ) ) {
+		if ( selectedFontSize ) {
+			return (
+				selectedFontSize.name ||
+				T_SHIRT_NAMES[ fontSizes.indexOf( selectedFontSize ) ]
+			);
+		}
+		return '';
```
tyxla: I know this is pre-existing, but I wasn't sure why in this case we don't display the common unit (like we do below) and if this is intentional.

packages/components/src/font-size-picker/test/index.tsx
```diff
@@ -525,6 +525,39 @@ describe( 'FontSizePicker', () => {
 			expect( screen.getByLabelText( 'Custom' ) ).toBeInTheDocument();
 		} );
 
+		it( 'hides custom input when disableCustomFontSizes is set to `true` with a custom font size', () => {
+			const { rerender } = render(
+				<FontSizePicker fontSizes={ fontSizes } value="3px" />
+			);
+			expect( screen.getByLabelText( 'Custom' ) ).toBeInTheDocument();
```
tyxla: I usually prefer `.toBeVisible()` for existing visible elements, and `.not.toBeInTheDocument()` for asserting non-existing ones, because `.toBeInTheDocument()` doesn't guarantee that the element is actually visible.

packages/components/src/font-size-picker/test/index.tsx
```diff
@@ -525,6 +525,39 @@ describe( 'FontSizePicker', () => {
...
+			expect(
+				screen.queryByLabelText( 'Custom' )
+			).not.toBeInTheDocument();
+		} );
+
+		it( "doesn't hide custom input when the selected font size is a predef", () => {
+			const { rerender } = render(
+				<FontSizePicker fontSizes={ fontSizes } value="3px" />
+			);
+			expect( screen.getByLabelText( 'Custom' ) ).toBeInTheDocument();
```
tyxla: Same as above wrt preferring `.toBeVisible()`

packages/components/src/font-size-picker/test/index.tsx
```diff
@@ -525,6 +525,39 @@ describe( 'FontSizePicker', () => {
...
+
+			rerender(
+				<FontSizePicker
+					fontSizes={ fontSizes }
+					value={ fontSizes[ 0 ].size }
+				/>
+			);
+			expect( screen.getByLabelText( 'Custom' ) ).toBeInTheDocument();
```
tyxla: Same as above wrt preferring `.toBeVisible()`

packages/components/src/font-size-picker/types.ts
```diff
@@ -29,24 +31,25 @@ export type FontSizePickerProps = {
...
+	 * custom font size is active. Has no effect when `disableCustomFontSizes`
+	 * is `true`.
 	 *
 	 * @default false
 	 */
 	withSlider?: boolean;
 	/**
 	 * If `true`, a reset button will be displayed alongside the input field
 	 * when a custom font size is active. Has no effect when
-	 * `disableCustomFontSizes` or `withSlider` is `true`.
```
tyxla, line 49: Good catch, seems like this was wrong.

ciampo:
That's an interesting way to rethink the component's logic.

> This will fail the "doesn't hide custom input when the selected font size is a predef" tests. However, I'm not sure this behavior is necessarily wrong? If the user never requested or interacted with the custom inputs, maybe it is more natural to fall back to the predef pickers if the value changed programatically. What do you think?

It's not necessarily wrong, but there may be value in showing the user the current value via the custom picker, rather than a blank togglegroup/select control. Imagine a user setting a custom font size, and later re-accessing those settings — wouldn't be weird not to see that custom font size reflected in the UI? It could give the wrong impression that the setting was not recorded.

Another unintended side-effect of the proposed change (which was apparently not covered by unit tests 😱 I added one! ) is that the user is not able to switch to predef inputs when a custom value (ie. non-predef)  is entered in the custom UI:


https://github.com/user-attachments/assets/14ca326d-437b-42a8-89eb-0ec0a3c0ef39


I went ahead and made a tweak to the code you suggested (this diff is meant to be applied on top of the diff previously suggested by you), which should be closer to the original behavior (causing all existing unit tests to pass).

```diff
diff --git a/packages/components/src/font-size-picker/index.tsx b/packages/components/src/font-size-picker/index.tsx
index 8f5fba2be0..7085487adb 100644
--- a/packages/components/src/font-size-picker/index.tsx
+++ b/packages/components/src/font-size-picker/index.tsx
@@ -107,7 +107,9 @@ const UnforwardedFontSizePicker = (
 	);
 	const isCustomValue = !! value && ! selectedFontSize;
 
-	const [ userRequestedCustom, setUserRequestedCustom ] = useState( false );
+	// Initially request a custom picker if the value is not from the predef list.
+	const [ userRequestedCustom, setUserRequestedCustom ] =
+		useState( isCustomValue );
 
 	const units = useCustomUnits( {
 		availableUnits: unitsProp,
@@ -120,7 +122,7 @@ const UnforwardedFontSizePicker = (
 	const currentPickerType = getPickerType(
 		// If showing the custom value picker, switch back to predef only
 		// if `disableCustomFontSizes` is set to `true`.
-		! disableCustomFontSizes && ( userRequestedCustom || isCustomValue ),
+		! disableCustomFontSizes && userRequestedCustom,
 		fontSizes
 	);
 

```

packages/components/src/font-size-picker/index.tsx
ciampo: I had initially extracted those function to a separate utility to better isolate the key pieces of logic and make the render function easier to scan.

But after applying Lena's advice, I noticed that some extra code could be removed — and so I put them back in the render function:

- https://github.com/WordPress/gutenberg/pull/63553/commits/41deb9f7b62621cbe75dcd8e158963d0dec395f0
- https://github.com/WordPress/gutenberg/pull/63553/commits/c53036ad3df5f85ff4b817dd6de6e7123b1bb6d3

packages/components/src/font-size-picker/index.tsx
```diff
@@ -59,43 +61,49 @@ const UnforwardedFontSizePicker = (
...
+			fontSizes.length > MAX_TOGGLE_GROUP_SIZES
+				? ( 'select' as const )
+				: ( 'togglegroup' as const );
+	}
 
-		const commonUnit = getCommonSizeUnit( fontSizes );
-		if ( commonUnit ) {
-			return `(${ commonUnit })`;
+	const headerHint = useMemo( () => {
+		switch ( currentPickerType ) {
```
ciampo, line 86: Swapped the previous `if`-based logic to a `switch`-based logic looking at the value of `currentPickerType`, which should be much simpler to follow

packages/components/src/font-size-picker/test/index.tsx
ciampo: Changed in https://github.com/WordPress/gutenberg/pull/63553/commits/ae4a0d233d726ca9066e18712d7bd2e9545f8a6e for all occurrences in the file

packages/components/src/font-size-picker/test/index.tsx
ciampo: Changed in https://github.com/WordPress/gutenberg/pull/63553/commits/ae4a0d233d726ca9066e18712d7bd2e9545f8a6e for all occurrences in the file

packages/components/src/font-size-picker/test/index.tsx
ciampo: Changed in https://github.com/WordPress/gutenberg/pull/63553/commits/ae4a0d233d726ca9066e18712d7bd2e9545f8a6e for all occurrences in the file

packages/components/src/font-size-picker/index.tsx
ciampo: I'm not sure if this was intentional or not — I would probably need to look in the git history and see if this aspect was discussed actively in previous PRs. Probably out of scope for this refactor, though — for now, I've decided to keep the `''` fallback around to preserve the same behavior as on `trunk`.

mirka (approved):
Looks good and tests well, thanks for the refinements! The logic is now easier to follow, we have better test coverage, and we have better behavior for the edge cases 🚀

Issue #53874, opened 2023-08-22
Gutenberg as a framework: streamline the experience


youknowriad:
Gutenberg can be used as a platform/framework to build block editors. Mainly thanks to the `@wordpress/block-editor` package. That said, the experience today is not as straightforward as it can be. There can be a lot of small gotchas and hacks you need to do in order to achieve the desired result. This issue is going to serve as an overview issue to collect and fix all these issues.

 - [ ] Peer dependencies: Installing the `@wordpress/*` packages using npm shows a number of warnings due to unmet peer dependencies, with the latest updates to npm, we should make sure we're using peer dependencies properly. For instance it seems that React shouldn't be a peer dependency for us but instead a real dependency of `@wordpress/element` since this package is supposed to be used as a replacement for React.
 - [x] We use `process.env.something` in a lot of places in Gutenberg codebase. All JS environments do not provide these, for instance Vite relies on `import.meta.env` instead. We need to at least not break the JS execution if `process` is undefined. Assess whether `import.meta.env` is a better alternative these days.
 - [ ] There's some weird focus styles by default in the RichText elements, Maybe some reset CSS need to be bundled by default in the @wordpress/block-editor stylesheet.
 - [x] Render a `<BlockEditorProvider><BlockCanvas /></BlockEditorProvider>` should be enough to render the most basic block editor. Today, we're forced to render a number of wrappers and providers or else the experience is broken. Find a way to bundle these providers or make them optional (bundle them if the provider hasn't been rendered already)
     - [x] `<BlockEditorKeyboardShortcuts.Register />` #53910
     - [x] `<WritingFlow>` #54149
     - [x] `<ObserveTyping>` #53875
     - [x] `<Popover.Slot />` #53889
     - [x] `<ShortcutProvider />` #54080
     - [x] `<SlotProvider />` #53940
     - [x] `useClipboardHandler` and `CopyHandler` #54207
     - [x] `useBlockSelectionClearer` and `BlockSelectionClearer` #54209
     - [x] `useTypewriter` and `Typewriter` (check if we should bundle or not)
 - [ ] Rendering the block editor requires including the stylesheets of a couple packages at least (@wordpress/components and @wordpress/block-editor) even if we're not directly depending on one of these packages (components). Is there a way we can auto-inject these styles while style producing valid npm packages. If not, clarify the documentation.
[...]

PR #51201, opened 2023-06-02
Footnotes: try with post meta

ellatrix:
## What?

Previously: #50124, #49797, #47682, #28261, #15267.

Similar to #50124, but storing it with post meta.

Many approaches have been explored, including storing footnotes inline and in the footnotes block, but eventually we decided to go for storing them in post meta to make it easier to render the footnotes outside of the post content in templates.

To do: fix batching (#51644)

## How?

This PR implements footnotes by:

* Registering a custom meta field where we store the footnotes in the correct order.
* The order is maintained by the entity provider, which looks through all rich text values in order, rearranges the footnotes array when necessary, but also adds and removes footnotes when they appear in the content or disappear.
* A footnotes block is inserted at the bottom when the first footnote is added. This block simply renders the footnotes it finds in the meta field, nothing more. This blocks could be moved anywhere, probably also outside of the post content.
* There's a rich text format that adds a button to insert footnote anchors. It's also responsible for rendering them inside rich text, making sure it's not editable, selectable etc.

## Testing Instructions

Create a post, add a paragraph and click the "more" section in the rich text section to insert a new footnote anchor. A footnotes block should appear at the bottom where you can add the note. Add a few in different places, rearrange them, and check that the order at the bottom remains correct. Also test copy pasting.

PR #63554, opened 2024-07-15
Footnotes: Register format within the init function

youknowriad:
Address https://github.com/WordPress/gutenberg/pull/51201#discussion_r1671641381 
Related #53874 

## What?

Some third-party editor uses the `@wordpress/block-library` package but only pick generic blocks from it. Things are WP specific like footnotes are not needed. 

The problem in trunk right now is that the "footnotes" format is registered as soon as you import the package regardless of whether you register the footnotes block or not which results in a 404 api request for all these editors.

This PR fixes that by moving the registration of the footnote format to the "init" function.

## Testing Instructions

1- Check that footnotes still work in the post editor.
2- Check that when you type in the "storybook" example for block editors, you won't see any failed API request.

### DIFF
--- a/packages/block-library/src/footnotes/index.js
+++ b/packages/block-library/src/footnotes/index.js
@@ -21,8 +21,7 @@ export const settings = {
 	edit,
 };
 
-registerFormatType( formatName, format );
-
 export const init = () => {
+	registerFormatType( formatName, format );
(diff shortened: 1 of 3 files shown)


ellatrix (approved):
Thanks!

PR #63555, opened 2024-07-15
Style variations: Don't display the default if its the only variation

scruffian:
## What?
Don't display the default variation if there are no other variations.  Alternative to #63083

## Why?
It's not necessary to display the default when there are no other options to choose from.

## How?
Return null if `fullStyleVariations` is not set or has no items.

## Testing Instructions
1. Remove all style variations from your theme
2. Open the Styles panel
3. Check that you don't see any style variations, including the default
4. Also try this when you have some color and typography variations.

## Screenshots or screencast 
Screenshot 2024-07-15 at 11 47 51


bgardner:
@scruffian Much better, as it shows a link to the Browse styles still:

image

Which allows users to see this:

image

jeryj (approved):
👍🏻

PR #45719, opened 2022-11-11
Global Styles: Add block preview component in global styles

madhusudhand:
## What?

This change adds block preview in global styles.

allblocks-preview

## How?

1. A new panel component has been introduced to include in the global styles.
2. The component considers the example from block-library and renders the component.
3. Supports enabled for few components to test the feature.

Note: there are few issues observed in the examples of block such as `image` which renders more spacing. However, it seem to be a different issue.

## Testing Instructions

1. Open site editor
4. Open global styles panel from the menu.
5. Go to Blocks -> [select any block such as paragraph]
6. View the preview of the block
7. Update the block styles such as typography, color etc..
8. View the new updates in the preview.

## Screenshots or screencast 

paragraph-preview

## Issues

Partial fix for #42919

PR #63558, opened 2024-07-15
Disallow scrolling the block preview.

jasmussen:
## What?

The block preview is scrollable (related, #45719):

before

This PR makes it not scrollable:

after

## Why?

The content is inert, not meant to be interacted with.

## Testing Instructions

Go to site editor > global styles. Go to blocks, select a block such as paragraph. The preview here should not be scrollable.

### DIFF
--- a/packages/edit-site/src/components/global-styles/style.scss
+++ b/packages/edit-site/src/components/global-styles/style.scss
@@ -112,9 +112,9 @@
 .edit-site-global-styles__block-preview-panel {
 	position: relative;
 	width: 100%;
-	overflow: auto;
 	border: $gray-200 $border-width solid;
 	border-radius: $radius-block-ui;
+	overflow: hidden;
 }
 
 .edit-site-global-styles__shadow-preview-panel {


MaggieCabrera:
The main issue with this I see with the Image block and the columns block:

Inserter | GS
--- | ---
Screenshot 2024-07-15 at 13 59 05 | Screenshot 2024-07-15 at 13 57 37
Screenshot 2024-07-15 at 13 59 12 | Screenshot 2024-07-15 at 13 57 47

The columns block preview doesn't look amazing in either case in my opinion anyway

richtabor:
> The columns block preview doesn't look amazing in either case in my opinion anyway

We need to add a viewport width so that columns actually display side by side.

jasmussen:
Good catch on the cropped pieces, however that preview was never meant to be scrollable, just like it isn't in the flyout preview either. It's a small glimpse at the content, but height-limited to not push down important inspector content substantially.

Mont Blanc has served us well, but it could be time to find something better. Perhaps we can find something more landscape shaped, so the caption is visible. To that end, I see this more as a forcing function to improve our previews: the content, etc., than anything else.

It's also an issue when you change the styling:

tradeoff

That said, and as is visible in this GIF too, I just discovered another issue 🙈—if you scale up the text in a paragraph, then scale it back down, the text stays pushed downwards. 

I'm about to head AFK for a couple of weeks. Outside of an approval of this one so we can land it, others should feel free to push to this branch (i.e. if a quick viewport width improves the image, seems fine). Otherwise I'll pick it back up when I return.

MaggieCabrera (approved):
I think the previews shortcomings shouldn't block this PR, we can make a pass at improving them separately, like Rich mentions for the columns block

MaggieCabrera:
I'm going to have a look at the problem with the text being pushed cause I think I know why it is, I'll add to this PR

PR #63559, opened 2024-07-15
Background tool: Fix double border.

jasmussen:
## What?

There's a double border in the global styles background tools:

before

This PR removes that double border:

after

## How?

Using a local CSS override. 

## Testing Instructions

Open global styles > blocks > quote, then scroll down and add a square background image. Open the flyout and observe only a single border in the focal point picker area.

### DIFF
--- a/packages/block-editor/src/components/global-styles/style.scss
+++ b/packages/block-editor/src/components/global-styles/style.scss
@@ -192,14 +192,21 @@
 	.components-toggle-control {
 		margin-bottom: 0;
 	}
+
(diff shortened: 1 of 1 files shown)


ramonjd (approved):
I think this is a Safari-only phenomenon, as I couldn't reproduce on Firefox or Chrome.

Here's Safari:

| Before  | After |
| Screenshot 2024-07-22 at 1 52 21 PM  | Screenshot 2024-07-22 at 1 53 02 PM  |

Also looks good on other browsers.

LGTM

PR #63562, opened 2024-07-15
Global Styles: Disable "Reset styles" button when there are no changes

t-hamano:
## What?

This PR disables "Reset styles" button when there are no changes in the global styles.

## Why?

#61271 adds the `_links` field to the global style context. This causes the `fastDeepEqual()` function to always return `false`, so the global styles is always considered "changed."

## How?

Exclude the _links field.

## Testing Instructions

- Click the Reset styles button.
- When you open the menu again, the button will be disabled.
- Make changes to the global styles and save them.
- The button will be enabled again.

### DIFF
--- a/packages/block-editor/src/components/global-styles/hooks.js
+++ b/packages/block-editor/src/components/global-styles/hooks.js
@@ -85,7 +85,11 @@ const VALID_SETTINGS = [
 ];
 
 export const useGlobalStylesReset = () => {
-	const { user: config, setUserConfig } = useContext( GlobalStylesContext );
+	const { user, setUserConfig } = useContext( GlobalStylesContext );
+	const config = {
+		settings: user.settings,
+		styles: user.styles,
+	};
 	const canReset = !! config && ! fastDeepEqual( config, EMPTY_CONFIG );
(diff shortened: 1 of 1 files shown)


t-hamano:
I don't consider this problem critical, but since it's a problem that's new to WP6.6, it might be worth including in a minor release.

cc @ellatrix

andrewserong (approved):
Thanks @t-hamano, this is testing well for me!

Also, since we're comparing against `EMPTY_CONFIG`, which explicitly defines only the `settings` and `styles` keys, it makes sense to me that we're only grabbing these parts of the user config for this comparison.

Testing well:

https://github.com/user-attachments/assets/64557009-c784-4c91-9e1a-7b1ea29e201a

LGTM! 🚀

aaronrobertshaw:
I've given these changes a run against the `wp/6.6` branch. They test well for me there too. I couldn't break it.

https://github.com/user-attachments/assets/3b2e0c2b-1cdf-4081-ac80-bcaebe60c7ad

ramonjd:
Good catch! 🙇🏻

PR #63364, opened 2024-07-10
Composite: export new stable version, deprecate unstable version

ciampo:
## What?

Closes #58850

This PR promotes the new private `CompositeV2` component family to stable, and marks the "unstable" (legacy) version of the component as deprecated.

## Why?

The unstable version is using a sub-optimal API layer and is currently not 100% supported. Consumers of the component should migrate to using the new, unprefixed, stable version of `Composite`.

## How?

- Export the new, ariakit-based `Composite` component from the `@wordpress/components` package
- Mark the legacy `__unstable` components as deprecated, and update unit tests
- Remove the `CompositeV2` private APIs, and update imports across the package
- Update Storybook files

TODO:
- [ ] Add a README

> [!WARNING]  
> This PR shouldn't be merged until we reach consensus on [what naming strategy should be adopted for compounds components](https://github.com/WordPress/gutenberg/issues/63242).

## Testing Instructions

- Project builds
- Smoke test in Storybook and in the editor components using `Composite`

Issue #58850, opened 2024-02-08
Make Ariakit `Composite` API standard


andrewhayward:
Reakit is now gone (#53278) and `Composite` is implemented entirely with Ariakit. However, Ariakit's API is only exported as a private `V2`, and the public (albeit `__unstable`) export is a legacy implementation designed to maintain Reakit's API.

We need to monitor usage, and decide when to...
- [x] ...export and stabilize the new version of the component https://github.com/WordPress/gutenberg/pull/63564
- [x] ...move all consumers away from the private API export and remove the private API export #63569
- [x] ...deprecate the `__unstable` legacy implementation #63572

## ✍️ Dev Note

- see [this dev note](https://github.com/WordPress/gutenberg/pull/58620#issuecomment-1976663196)) for the unstable component's refactor
- see the dev note in https://github.com/WordPress/gutenberg/pull/63569 for the stabilization of the new `Composite` component

ciampo:
Closing as the new implementation of `Composite` is now exported as a stable API, while the previous unstable version is marked as deprecated.

PR #63564, opened 2024-07-15
Composite: stabilize new ariakit implementation

ciampo:
## What?

Extracted from #63364

Part of #58850

Export the new `ariakit`-based `Composite` component from the `@wordpress/components` package.

## Why?

The component has been available via private APIs for a while without causing issues. It's time to stabilize it!

## How?

- Renamed the component's exports following the compound components naming convention;
- Updated all internal references to the component package, including private APIs;
- Exported the new component from the `@wordpress/components` package
- Updated storybook files
- Added documentation where needed

Note: from the point of view of a consumer of `@wordpress/components`, this PR only adds new APIs and shouldn't require any code changes.

### Follow-ups

- #64399
- #64397
- #63569
- #63572 

## Testing Instructions

No changes were made to how the component works, only to how it's exported / imported.

Smoke test that usages of the component both in Storybook and in the editor work as on trunk.

Reviewing commit-by-commit may help.


packages/components/src/alignment-matrix-control/index.tsx
```diff
@@ -56,7 +56,7 @@ export function AlignmentMatrixControl( {
 		id
 	);
 
-	const compositeStore = useCompositeStore( {
+	const compositeStore = Composite.useStore( {
```
ciampo: This is actually an interesting one when it comes to naming. Since we're namespacing the compound component under the `Composite` namespace, it felt redundant to call the hook `useCompositeStore`. I therefore opted for `useStore`, expecting consumers to use it as `Composite.useStore`

packages/components/src/circular-option-picker/types.ts
```diff
@@ -14,7 +14,7 @@ import type { Icon } from '@wordpress/icons';
 import type { ButtonAsButtonProps } from '../button/types';
 import type { DropdownProps } from '../dropdown/types';
 import type { WordPressComponentProps } from '../context';
-import type { CompositeStore } from '../composite/v2';
+import type { Store as CompositeStore } from '../composite';
```
ciampo: Same as above for types — I went for `Store` as I expect consumers to consume this API as `Composite.Store` (or by aliasing the import like in this case)

packages/components/src/composite/current/stories/index.story.tsx
```diff
@@ -11,35 +11,31 @@ import { isRTL } from '@wordpress/i18n';
...
 	parameters: {
 		docs: {
 			canvas: { sourceState: 'shown' },
 			source: { transform },
 			extractArgTypes: ( component: React.FunctionComponent ) => {
-				const name = component.displayName;
+				const name =
+					component.displayName === 'useStore'
+						? 'useCompositeStore'
+						: component.displayName;
```
ciampo: Necessary to generate the correct documentation for the `Composite.useStore` hook.

packages/components/src/composite/legacy/stories/index.story.tsx
```diff
@@ -15,7 +15,8 @@ import {
 import { UseCompositeStatePlaceholder, transform } from './utils';
 
 const meta: Meta< typeof UseCompositeStatePlaceholder > = {
-	title: 'Components/Composite',
+	// TODO: should we keep this story around? If so, how should we call it?
+	title: 'Components/Composite/Legacy',
```
ciampo: Probably @mirka is best-suited for the above TODO question. Also, see other comment on the current implementation's Storybook `title`

packages/components/src/composite/current/stories/index.story.tsx
```diff
@@ -11,35 +11,31 @@ import { isRTL } from '@wordpress/i18n';
...
-	CompositeRow,
-	CompositeItem,
-	useCompositeStore,
-} from '..';
+import * as Composite from '..';
 import { UseCompositeStorePlaceholder, transform } from './utils';
 
 const meta: Meta< typeof UseCompositeStorePlaceholder > = {
-	title: 'Components/Composite (V2)',
+	title: 'Components/Composite',
```
ciampo, line 18: I opted for keeping `Components/Composite` since the unstable version was using the same title, but happy to discuss alternative options (eg. adding a `/Default` or `/Current` subfolder?)

packages/components/src/alignment-matrix-control/index.tsx
ciampo: Another consideration to make is whether we think that hooks should be part of the compounding namespace at all.  For example, one inconvenience of having to invoke `Composite.useStore` is that linters recognize hooks based on the fact that they start with `use`

packages/components/src/composite/legacy/index.tsx
```diff
@@ -5,6 +5,11 @@
  * tab stop for the whole Composite element. This means that it can behave as
  * a roving tabindex or aria-activedescendant container.
  *
+ * This file aims at providing component that are as close as possible to the
+ * original `reakit`-based implementation (which was removed from the codebase),
+ * although it is recommended that consumers of the package switch to the stable,
+ * un-prefixed, ariakit-based version of `Composite`.
```
tyxla:
```suggestion
 * This file aims at providing components that are as close as possible to the
 * original `reakit`-based implementation (which was removed from the codebase),
 * although it is recommended that consumers of the package switch to the stable,
 * un-prefixed, `ariakit`-based version of `Composite`.
```

packages/dataviews/src/layouts/list/index.tsx
```diff
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import clsx from 'clsx';
+// TODO: use the @wordpress/components one once public
```
tyxla: Can't we do it already? With this PR we should be able to use `Composite.Store`.

packages/components/src/composite/README.md
```diff
@@ -0,0 +1,5 @@
+# `Composite`
+
+`Composite` provides a single tab stop on the page and allows navigation through the focusable descendants with arrow keys. This abstract component is based on the [WAI-ARIA Composite Role⁠](https://w3c.github.io/aria/#composite).
+
+See the [Ariakit docs for the `Composite` component](https://ariakit.org/components/composite).
```
tyxla, line 5: Since we're aliasing some of the components, it might make sense to list them all here and link them to the respective documentation.

packages/components/src/composite/stories/utils.tsx
```diff
@@ -6,8 +6,11 @@ import type { StoryContext } from '@storybook/react';
 /**
  * Internal dependencies
  */
-import type { CompositeStoreProps } from '..';
+import type { StoreProps as CompositeStoreProps } from '..';
```
tyxla: Does it make sense to do the import with `import * as Composite` here too, for consistency? Folks may look at the storybook code and borrow from it.

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -60,27 +56,33 @@ const meta: Meta< typeof UseCompositeStorePlaceholder > = {
...
-			<CompositeRow role="row">
-				<CompositeItem role="gridcell">Item C1</CompositeItem>
-				<CompositeItem role="gridcell">Item C2</CompositeItem>
-				<CompositeItem role="gridcell">Item C3</CompositeItem>
-			</CompositeRow>
-		</Composite>
+		<Composite.Root
+			role="grid"
+			store={ store }
+			aria-label="Ariakit Composite"
```
tyxla: Why are we labeling this "Ariakit Composite"? I expected it to be just `Composite`:

```suggestion
			aria-label="Composite"
```

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -11,35 +11,31 @@ import { isRTL } from '@wordpress/i18n';
...
+		Root: Composite.Root,
 		// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
-		CompositeGroup,
+		Group: Composite.Group,
 		// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
-		CompositeRow,
+		Row: Composite.Row,
 		// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
-		CompositeItem,
+		Item: Composite.Item,
```
tyxla: There's another inconvenience of the dot notation. It would be nice if subcomponents were directly inferred.

packages/components/src/private-apis.ts
```diff
@@ -28,11 +22,11 @@ import { lock } from './lock-unlock';
 
 export const privateApis = {};
 lock( privateApis, {
-	CompositeV2,
-	CompositeGroupV2,
-	CompositeItemV2,
-	CompositeRowV2,
-	useCompositeStoreV2,
+	CompositeV2: Composite.Root,
```
tyxla: We're planning to start using the stable imports and get rid of those in a separate PR, right?

packages/components/src/index.ts
```diff
@@ -61,7 +61,8 @@ export {
 	CompositeGroup as __unstableCompositeGroup,
 	CompositeItem as __unstableCompositeItem,
 	useCompositeState as __unstableUseCompositeState,
-} from './composite';
+} from './composite/legacy';
```
tyxla, line 64: How much are those used? Can we deprecate them immediately?

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ *   <CompositeRow>
+ *     <Composite.Item>Item 2.1</Composite.Item>
+ *     <Composite.Item>Item 2.2</Composite.Item>
+ *     <Composite.Item>Item 2.3</Composite.Item>
+ *   </CompositeRow>
+ * </Composite.Root>
+ * ```
+ */
+export const Row = CompositeRow;
+
```
tyxla: Could be worth looking at what else we're trying to consume and re-exporting that so it's imported from `@wordpress/components`. Examples from the dataviews package are `Ariakit.CompositeHover` and `Ariakit.CompositeTypeahead`.

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ * ```jsx
+ * const useCompositeStore = Composite.useStore;
+ * const store = useCompositeStore();
+ * <Composite.Root store={store}>
+ *   <Composite.Item>Item 1</Composite.Item>
+ *   <Composite.Item>Item 2</Composite.Item>
+ * </Composite.Root>
+ * ```
+ */
+export const Root = Composite;
```
tyxla: I see the need to have a `Composite.Root`, but I kind of can't really love that name. I wonder what other alternatives we have. Sometimes I wish we could just use `Composite`.

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+	CompositeGroupLabel,
+	CompositeItem,
+	CompositeRow,
+	useCompositeStore,
+} from '@ariakit/react';
+import type { CompositeStore, CompositeStoreProps } from '@ariakit/react';
+
+/**
+ * Creates a composite store.
+ * @see https://ariakit.org/components/composite
```
tyxla:
```suggestion
 * @see https://ariakit.org/reference/use-composite-store
```

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ *   <Composite.Item>Item 1</Composite.Item>
+ *   <Composite.Item>Item 2</Composite.Item>
+ * </Composite.Root>
+ * ```
+ */
+export const Root = Composite;
+
+/**
+ * Renders a group element for composite items.
+ * @see https://ariakit.org/components/composite
```
tyxla:
```suggestion
 * @see https://ariakit.org/reference/composite-group
```

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ * @example
+ * ```jsx
+ * const useCompositeStore = Composite.useStore;
+ * const store = useCompositeStore();
```
tyxla: Do we need a store? We could just have a `CompositeProvider` perhaps?

Which makes me wonder, why aren't we re-exporting `CompositeProvider` too?

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ * </Composite.Root>
+ * ```
+ */
+export const Group = CompositeGroup;
+
+/**
+ * Renders a label in a composite group. This component must be wrapped with
+ * `CompositeGroup` so the `aria-labelledby` prop is properly set on the
+ * composite group element.
+ * @see https://ariakit.org/components/composite
```
tyxla:
```suggestion
 * @see https://ariakit.org/reference/composite-group-label
```

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ *     <Composite.Item>Item 2</Composite.Item>
+ *   </CompositeGroup>
+ * </Composite.Root>
+ * ```
+ */
+export const GroupLabel = CompositeGroupLabel;
+
+/**
+ * Renders a composite item.
+ * @see https://ariakit.org/components/composite
```
tyxla:
```suggestion
 * @see https://ariakit.org/reference/composite-item
```

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,135 @@
...
+ * </Composite.Root>
+ * ```
+ */
+export const Item = CompositeItem;
+
+/**
+ * Renders a composite row. Wrapping `Composite.Item` elements within
+ * `Composite.Row` will create a two-dimensional composite widget, such as a
+ * grid.
+ * @see https://ariakit.org/components/composite
```
tyxla:
```suggestion
 * @see https://ariakit.org/reference/composite-row
```

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -11,35 +11,31 @@ import { isRTL } from '@wordpress/i18n';
...
 	},
-	tags: [ 'status-private' ],
 	parameters: {
```
tyxla, line 32: I think there's something messed up with these, the default story doesn't have any controls:

Screenshot 2024-07-18 at 16 29 28

Or if that's intentional, it can definitely use some love and display all available props as controls.

I think that's pre-existing, because I see this on `trunk`:

Screenshot 2024-07-18 at 16 31 39

tyxla (commented):
Thanks for working on this one @ciampo! Excited about stabilizing the Ariakit version of `Composite`!

This looks and works great overall!

Left some comments and questions, some might be naive, so take them with a grain of salt 😉

packages/dataviews/src/layouts/list/index.tsx
ciampo: We can, but as I split the whole process on several PRs, this PR focuses on exporting the v2 as stable. 

Also, as recently discussed with Lena, we will try to initially avoid exporting types — instead of using the type from `@wordpress/components`, we will use `React.ComponentProps`, `ReturnType<>` and `Parameters<>` to extract the type from the exported components and hooks. 

I will update the code comment.

packages/components/src/composite/stories/utils.tsx
ciampo: That's a good point — however, I will update the code to derive the type from the component, as explained [here](https://github.com/WordPress/gutenberg/pull/63564#discussion_r1684103489)

packages/components/src/composite/stories/index.story.tsx
ciampo: Not sure either, it's a remnant from previous work.

packages/components/src/composite/stories/index.story.tsx
ciampo: I can try to use object spread + a utility to filter out non-components stuff, something like 	

```js
{ ...getSubComponents( Composite ) }

// ...


function getSubComponents( component ) {
	return Object.keys(component)
          .filter( key => isReactComponent(obj[key]) )
          .reduce( (res, key) => (res[key] = obj[key], res), {} );
}

function isReactComponent( x ) {
	// TBD
	return typeof x === function && 'displayName' in x && x.displayName.includes('.') && ! x.displayName.includes('.Root');
}
```

Although it could be overkill, and I'd be tempted to go "manual" and hope that Storybook will automatically detect this one day

packages/components/src/private-apis.ts
ciampo: Yup, there are 2 follow-up PRs planned:
- one where we swap all usages to the public APIs and delete the private APIs
- one where we deprecated the V1 __unstable APIs

packages/components/src/index.ts
ciampo: We will, in a follow-up PR

packages/components/src/composite/index.ts
ciampo: Absolutely, although I'd almost argue we should do that in a separate PR

packages/components/src/composite/index.ts
ciampo: I understand your point! That would also be my preference if we could start a new package from scratch.

But as explained before, our decision was influenced by the fact that we have many older, monolithic components in the package. Associating dot notation to compound components brings to important benefits:

- a clear distinction in compound vs monolithic components for consumers of the package;
- a way to avoid any potential naming conflicts, in case a component exposes both a compound and a monolithic version

packages/components/src/composite/index.ts
ciampo: In general, when using ariakit components, we've opted for using the store hooks directly instead of relying on the `Provider` component because of slot fill limitations which would break the provider-based version. By using the store hook explicitly, we can forward the store via a component-dedicated context and get around slot-fill boundaries.

packages/components/src/composite/stories/index.story.tsx
ciampo: That's indeed a decision that was taken by whoever worked on the component before. I'll see if it's an easy change to apply or if there were larger implications.

packages/dataviews/src/layouts/list/index.tsx
tyxla: Sounds good!

packages/components/src/composite/stories/index.story.tsx
tyxla: Definitely looks like overkill. Let's go with the manual one.

packages/components/src/composite/index.ts
tyxla: A separate PR makes sense!

packages/components/src/composite/index.ts
tyxla: Got it, thanks for elaborating!

packages/components/src/composite/stories/index.story.tsx
tyxla: Totally fine to be fixed in another PR since it's pre-existing behavior.

tyxla (approved):
This LGTM, and tests well 👍 

I acknowledge that there are a bunch of follow-ups after merging this one, and that's more than fine. Smaller changes always win!

🚀

mirka (commented):
Waiting for an agreement on #63714, but just to add a concrete example to what I'm suggesting in https://github.com/WordPress/gutenberg/pull/63714#discussion_r1684937161, I think Composite is a case where we _don't_ need a `.Root`. The basic, default API should always start with `MyComponent`, and only when we need to provide a more low-level API in parallel, do we need a `MyComponent.Root`. Most consumers should never have to deal with a `.Root`. It's almost like the escape hatch for us as a component library. What do you think?

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -11,35 +11,31 @@ import { isRTL } from '@wordpress/i18n';
...
-	title: 'Components/Composite (V2)',
+	title: 'Components/Composite',
 	component: UseCompositeStorePlaceholder,
 	subcomponents: {
 		// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
-		Composite,
+		Root: Composite.Root,
 		// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
-		CompositeGroup,
+		Group: Composite.Group,
```
mirka: I think we'd want these to show the full dot-notated name, like we did for Tabs and CustomSelectControlV2.

```suggestion
		'Composite.Group': Composite.Group,
```

packages/components/src/composite/index.ts
mirka, line 1: I think ideally we should have dot-notated `displayName`s for every component, to clarify both Storybook code snippets and React devtools. We already did it for CustomSelectControlV2, but not yet for Tabs.

tyxla:
> Waiting for an agreement on #63714, but just to add a concrete example to what I'm suggesting in [#63714 (comment)](https://github.com/WordPress/gutenberg/pull/63714#discussion_r1684937161), I think Composite is a case where we _don't_ need a `.Root`. The basic, default API should always start with `MyComponent`, and only when we need to provide a more low-level API in parallel, do we need a `MyComponent.Root`. Most consumers should never have to deal with a `.Root`. It's almost like the escape hatch for us as a component library. What do you think?

Just pointing out that I agree with that and brought a similar point up previously in my review [here](https://github.com/WordPress/gutenberg/pull/63564#discussion_r1682822873).

ciampo:
Hey @tyxla and @mirka , I updated the PR using the overloaded naming convention. Could you take another look?

packages/components/src/composite/legacy/stories/index.story.tsx
tyxla: Maybe we should keep it like we keep other deprecated component stories (like `Navigation`)?

tyxla (approved):
With a bunch of follow-ups to come as discussed earlier, this looks good to me! 🚀 

Excited to see it stable!

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,137 @@
...
+	 *     <Composite.Item>Item 2.3</Composite.Item>
+	 *   </Composite.Row>
+	 * </Composite>
+	 * ```
+	 */
+	Row,
+} );
+
+export type Store = CompositeStore;
+export type StoreProps = CompositeStoreProps;
```
mirka: For the most part I think we can avoid using the Ariakit types directly. Any consumers will be interfacing with our Composite components/hooks, so any relevant types should be retrievable from the component/hook they're interfacing with. If not, it would be a signal for us that external consumers may have issues typing their code.

```diff
diff --git a/packages/components/src/circular-option-picker/circular-option-picker-option.tsx b/packages/components/src/circular-option-picker/circular-option-picker-option.tsx
index 2650416f47..cb4d5ae647 100644
--- a/packages/components/src/circular-option-picker/circular-option-picker-option.tsx
+++ b/packages/components/src/circular-option-picker/circular-option-picker-option.tsx
@@ -18,7 +18,7 @@ import { CircularOptionPickerContext } from './circular-option-picker-context';
 import Button from '../button';
 import { Composite } from '../composite';
 import Tooltip from '../tooltip';
-import type { OptionProps, CircularOptionPickerCompositeStore } from './types';
+import type { OptionProps } from './types';
 
 function UnforwardedOptionAsButton(
 	props: {
@@ -45,7 +45,9 @@ function UnforwardedOptionAsOption(
 		id: string;
 		className?: string;
 		isSelected?: boolean;
-		compositeStore: CircularOptionPickerCompositeStore;
+		compositeStore: NonNullable<
+			React.ComponentProps< typeof Composite >[ 'store' ]
+		>;
 	},
 	forwardedRef: ForwardedRef< any >
 ) {
diff --git a/packages/components/src/circular-option-picker/types.ts b/packages/components/src/circular-option-picker/types.ts
index 61dcc82c98..e23ff4165f 100644
--- a/packages/components/src/circular-option-picker/types.ts
+++ b/packages/components/src/circular-option-picker/types.ts
@@ -14,7 +14,7 @@ import type { Icon } from '@wordpress/icons';
 import type { ButtonAsButtonProps } from '../button/types';
 import type { DropdownProps } from '../dropdown/types';
 import type { WordPressComponentProps } from '../context';
-import type { Store as CompositeStore } from '../composite';
+import type { Composite } from '../composite';
 
 type CommonCircularOptionPickerProps = {
 	/**
@@ -123,8 +123,7 @@ export type OptionProps = Omit<
 	>;
 };
 
-export type CircularOptionPickerCompositeStore = CompositeStore;
 export type CircularOptionPickerContextProps = {
 	baseId?: string;
-	compositeStore?: CircularOptionPickerCompositeStore;
+	compositeStore?: React.ComponentProps< typeof Composite >[ 'store' ];
 };
diff --git a/packages/components/src/composite/index.ts b/packages/components/src/composite/index.ts
index c7a5b3234e..ac9acbc975 100644
--- a/packages/components/src/composite/index.ts
+++ b/packages/components/src/composite/index.ts
@@ -12,7 +12,6 @@
  * External dependencies
  */
 import * as Ariakit from '@ariakit/react';
-import type { CompositeStore, CompositeStoreProps } from '@ariakit/react';
 
 /**
  * Creates a composite store.
@@ -132,6 +131,3 @@ export const Composite = Object.assign( Ariakit.Composite, {
 	 */
 	Row,
 } );
-
-export type Store = CompositeStore;
-export type StoreProps = CompositeStoreProps;
diff --git a/packages/components/src/composite/stories/utils.tsx b/packages/components/src/composite/stories/utils.tsx
index 10e9ac8036..396703875f 100644
--- a/packages/components/src/composite/stories/utils.tsx
+++ b/packages/components/src/composite/stories/utils.tsx
@@ -6,12 +6,14 @@ import type { StoryContext } from '@storybook/react';
 /**
  * Internal dependencies
  */
-import type { StoreProps as CompositeStoreProps } from '..';
+import type { useCompositeStore } from '..';
 
 /**
  * Renders a composite widget.
  */
-export function UseCompositeStorePlaceholder( props: CompositeStoreProps ) {
+export function UseCompositeStorePlaceholder(
+	props: Parameters< typeof useCompositeStore >
+) {
 	return (
 		<dl>
 			{ Object.entries( props ).map( ( [ name, value ] ) => (

```

packages/components/src/composite/legacy/stories/index.story.tsx
mirka: Yes, I think we should keep it as long as it's available in the package. As for how, my recommendation would be like this:

```diff
diff --git a/packages/components/src/composite/legacy/stories/index.story.tsx b/packages/components/src/composite/legacy/stories/index.story.tsx
index 08eebc65ed..d3d3205afe 100644
--- a/packages/components/src/composite/legacy/stories/index.story.tsx
+++ b/packages/components/src/composite/legacy/stories/index.story.tsx
@@ -15,8 +15,8 @@ import {
 import { UseCompositeStatePlaceholder, transform } from './utils';
 
 const meta: Meta< typeof UseCompositeStatePlaceholder > = {
-	// TODO: should we keep this story around? If so, how should we call it?
-	title: 'Components/Composite/Legacy',
+	title: 'Components (Deprecated)/Composite (Unstable)',
+	id: 'components-composite-unstable',
 	component: UseCompositeStatePlaceholder,
 	subcomponents: {
 		// @ts-expect-error Storybook doesn't like overloaded exports as subcomponents
diff --git a/packages/components/src/composite/legacy/stories/utils.tsx b/packages/components/src/composite/legacy/stories/utils.tsx
index 06edd34863..2fb51c845f 100644
--- a/packages/components/src/composite/legacy/stories/utils.tsx
+++ b/packages/components/src/composite/legacy/stories/utils.tsx
@@ -8,6 +8,25 @@ import type { StoryContext } from '@storybook/react';
  */
 import type { LegacyStateOptions } from '..';
 
+/**
+ * Renders a composite widget.
+ *
+ * This unstable component is deprecated. Use `Composite` instead.
+ *
+ * ```jsx
+ * import {
+ * 	__unstableUseCompositeState as useCompositeState,
+ * 	__unstableComposite as Composite,
+ * 	__unstableCompositeItem as CompositeItem,
+ * } from '@wordpress/components';
+ *
+ * const state = useCompositeState();
+ * <Composite state={ state }>
+ * 	<CompositeItem>Item 1</CompositeItem>
+ * 	<CompositeItem>Item 2</CompositeItem>
+ * </Composite>;
+ * ```
+ */
 export function UseCompositeStatePlaceholder( props: LegacyStateOptions ) {
 	return (
 		<dl>

```

The important points being:

- It needs to be clear that we're referring to the `__unstable` version, and that it's deprecated.
- Usually we want to ensure that we [don't break permalinks](https://github.com/WordPress/gutenberg/blob/trunk/storybook/manager-head.html#L2), but in this specific case I think we would cause the less damage if we just made a clean break and changed the URLs so it reflects the new situation.

packages/components/src/composite/stories/utils.tsx
```diff
@@ -6,8 +6,11 @@ import type { StoryContext } from '@storybook/react';
...
 
+/**
+ * Renders a composite widget.
+ */
```
mirka: I think we also need a code snippet for the Storybook docs.

```suggestion
/**
 * Renders a composite widget.
 *
 * ```jsx
 * const useCompositeStore = useCompositeStore();
 * const store = useCompositeStore();
 * <Composite store={store}>
 *   <Composite.Item>Item 1</Composite.Item>
 *   <Composite.Item>Item 2</Composite.Item>
 * </Composite>
 * ```
 */
```

packages/components/src/composite/index.ts
```diff
@@ -1,7 +1,137 @@
...
+ * const useCompositeStore = useCompositeStore();
+ * const store = useCompositeStore();
+ * <Composite store={store}>
+ *   <Composite.Item>Item 1</Composite.Item>
+ *   <Composite.Item>Item 2</Composite.Item>
+ * </Composite>
+ * ```
+ */
+export const Composite = Object.assign( Ariakit.Composite, {
+	/**
```
mirka: Just me or is the display name rendering as `Composite2`? Not sure what that's about, but I think we may need to add an explicit `displayName` here.

```suggestion
export const Composite = Object.assign( Ariakit.Composite, {
	displayName: 'Composite',
	/**
```

packages/components/src/composite/legacy/index.tsx
```diff
@@ -145,19 +150,16 @@ function proxyComposite< C extends Component >(
...
-		typeof Current.CompositeGroup | typeof Current.CompositeRow
-	>
+	React.ComponentPropsWithoutRef< typeof Current.Group | typeof Current.Row >
 >( ( { role, ...props }, ref ) => {
-	const Component =
-		role === 'row' ? Current.CompositeRow : Current.CompositeGroup;
+	const Component = role === 'row' ? Current.Row : Current.Group;
 	return <Component ref={ ref } role={ role } { ...props } />;
 } );
 unproxiedCompositeGroup.displayName = 'CompositeGroup';
```
mirka: I wanted to fix the issue in the legacy story where `CustomItem` had a display name of `Custom.Item`. I removed this line because I thought it was unnecessary, and somehow everything is fixed 😂

```suggestion
```

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -0,0 +1,81 @@
...
+		// @ts-expect-error Storybook doesn't like overloaded exports as subcomponents
+		Composite,
+		// @ts-expect-error Storybook doesn't like overloaded exports as subcomponents
+		'Composite.Group': Composite.Group,
+		// @ts-expect-error Storybook doesn't like overloaded exports as subcomponents
+		'Composite.GroupLabel': Composite.GroupLabel,
+		// @ts-expect-error Storybook doesn't like overloaded exports as subcomponents
+		'Composite.Row': Composite.Row,
+		// @ts-expect-error Storybook doesn't like overloaded exports as subcomponents
+		'Composite.Item': Composite.Item,
```
mirka: FWIW I think this is just an issue with the `subcomponents` typing and not overloaded exports specifically.

https://github.com/WordPress/gutenberg/blob/6581dfb8f1c605119db776f023a83d655209bfa8/packages/components/src/dropdown-menu-v2/stories/index.story.tsx#L38-L53

packages/components/src/private-apis.ts
```diff
@@ -28,11 +22,11 @@ import { lock } from './lock-unlock';
...
-	CompositeV2,
-	CompositeGroupV2,
-	CompositeItemV2,
-	CompositeRowV2,
-	useCompositeStoreV2,
+	CompositeV2: Composite,
+	CompositeGroupV2: Composite.Group,
+	CompositeItemV2: Composite.Item,
+	CompositeRowV2: Composite.Row,
+	useCompositeStoreV2: useCompositeStore,
```
mirka, line 29: Oh, so technically we're not unlocking them yet in this PR? Should we keep a `tags: [ 'status-private' ]` on our story then?

packages/components/src/composite/index.ts
ciampo: Sounds good — I was going to do this in a follow-up, but I agree that it's actually better to do it in this PR and avoid exporting `Store` and `StoreProps` altogether.

https://github.com/WordPress/gutenberg/pull/63564/commits/59e8a71c36047a5479d9a8da0d68dd2af916d90f

There is one direct import from `@ariakit/react` in `packages/dataviews/src/dataviews-layouts/list/index.tsx`, but I will fix it in a follow-up, once the new composite component will be consumed without lock/unlock "eating up" its type definitions.

packages/components/src/private-apis.ts
ciampo: Unlocking will happen in a follow-up (as per this [feedback](https://github.com/WordPress/gutenberg/pull/63364#pullrequestreview-2174448617)), but I don't think we should add the private tag because, with this PR, the component is exported from the package and available for consumption without lock/unlock

packages/components/src/composite/index.ts
ciampo: Interesting. I think this was potentially a side-effect of the ariakit version update? After rebasing, I started getting errors when setting the `displayName` on ariakit components. I had to resort to `Object.assign` for the subcomponents too.

https://github.com/WordPress/gutenberg/pull/63564/commits/30d290e72c4f96de3dc2a4cb971c0d8bbcb12992

packages/components/src/composite/legacy/index.tsx
ciampo: Weird, maybe related to https://github.com/WordPress/gutenberg/pull/63564#discussion_r1709021031 ?

Done in https://github.com/WordPress/gutenberg/pull/63564/commits/e4b29b2981052b0cd567d9f1926bd31bb81422ed

packages/components/src/composite/legacy/stories/index.story.tsx
ciampo: Sounds good, done in https://github.com/WordPress/gutenberg/pull/63564/commits/b3e58bc5c1be96c27f8535ed7e6958da5ced2555

packages/components/src/composite/stories/index.story.tsx
ciampo: That's weird. I remember seeing a different TS error last week 🤔 

Anyway, I checked again and I now get the same error as in `DropdownMenuV2` indeed — that's good! I updated the `// @ts-expect-error` comment to match the ones used for other storybook files.

https://github.com/WordPress/gutenberg/pull/63564/commits/922426e4461eee080ef46eec5690ffa3272330c3

ciampo:
As discussed with @mirka, I am going to add explicit prop types to our `Composite` components (instead of re-exporting ariakit ones) so that we can have more control on any future breaking changes (and improve Storybook control generation in the process).

ciampo:
I pushed a few updates to the PR.

Some context: earlier today, when talking to @mirka , we realised that re-exporting Ariakti components directly would give us very little control on changes happening on the Ariakit side.

We analysed the range of possible approaches:
- on one side of the spectrum, we could continue to export Ariakit components as they are. This would be the quickest and leanest way, but it wouldn't give us any control on "filtering" Ariakit changes to the components package. For example, we wound't be able to detect / patch any breaking changes from ariakit.
- on the other end, we could wrap the ariakig components in a custom implementation, decide what props to expose / document, and write explicit prop types ourselves. This would represent more work and code on our end, but it would give us better type safety and more control when changes happen on the ariakit side.

**I experimented with the latter approach, and I reached the conclusion that it wouldn't be really feasible**. It would cause us to re-write a lot of Ariakit types (including some "internal" ones), basically re-creating a fork of ariakit, which would be time-consuming (and difficult) to maintain and keep up-to-date as needed.

For now, I decided to adopt a middle ground approach:

- I wrapped ariakit components in custom components, and exposed only the few props that are currently used by `Composite`'s consumers;
- I typed those props by re-using ariakit's types

This strategy still doesn't give us full control, since we're re-exporting Ariakit types — but it's better than before.

Also, this still doesn't fix Storybook arg's table — and so I went ahead and wrote it myself.

---

@WordPress/gutenberg-components  I'd love to hear opinions on this, especially in the light of more components (`DropdownMenu`, `CustomSelectControl`) that will likely incur in the same scenario.

packages/components/src/circular-option-picker/circular-option-picker-option.tsx
```diff
@@ -66,7 +68,6 @@ function UnforwardedOptionAsOption(
 					ref={ forwardedRef }
 				/>
 			}
-			store={ compositeStore }
```
ciampo, line 69: This isn't needed because in theory `Composite.Item` should inherit the store from its parent `Composite`. The only scenario in which it would break, is if the option and its parent component are rendered across a slot / fill.

packages/components/src/composite/README.md
ciampo, line 1: I will update the README if/ when we agree on the overall prop types stragegy

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -0,0 +1,218 @@
...
+								type: {
+									summary:
+										'((activeId: string | null | undefined) => void)',
+								},
+							},
+						},
+						focusLoop: {
+							name: 'focusLoop',
+							description:
+								'Determines how the focus behaves when the user reaches the end of the composite widget.',
```
ciampo, line 83: The original [ariakit docs](https://ariakit.org/reference/use-composite-store#focusloop) have a lot more info on some of these props, but I wasn't sure about adding all of that info here — partially because in Storybook it wouldn't be formatted nicely (including hyperlinks), and partially because it references other props that we are not exposing.

packages/components/src/private-apis.ts
mirka: Ah ok, got it 👍

mirka (approved):
I think this is a balanced compromise we can start with. It won't protect us from all possibilities, but it reduces our surface area without _too_ much overhead. I do wish we could've extracted the Storybook docs types automatically though 😭 (I took another stab but failed also.)

tyxla:
@ciampo you already have my approval from before, so feel free to take this one home 🚀

tyxla (approved):

packages/components/src/composite/stories/index.story.tsx
ciampo: I actually ended up manually adding type information to Storybook in this PR.

ciampo:
Thank you for the final round of review.

I've updated the README and marked the `store` prop on `Composite` as required as final tweaks.

I'll merge as soon as CI checks pass 🟢

PR #63565, opened 2024-07-15
useBlockElement: return null until ref callback has time to clean up the old element

### LINKED ISSUE #63448: WordPress 6.6 RC3 and Query Loop block issue (When WooCommerce is active)
### Description

When switching to tablet mode and back to desktop mode while the "Post Template" block is selected - the block breaks.

This seems to be somehow related to WooCommerce plugin.

Here is a quick view demonstration:
https://github.com/WordPress/gutenberg/assets/8998558/5d0bdd93-f7af-4293-b3c0-638e0a930941

Everything works fine if WooCommerce is deactivated.

1. Install WooCommerce plugin (I have the latest version)
2. Create a new page and insert the Query Loop block
3. Open the "Document Overview" sidebar
4. Select/click on the "Post template" block (this is important, see the attached video too)
5. Ensure the Post Template block is set to 'Grid' layout on the toolbar
6. Change the "View" to tablet and back to desktop

_No response_

### Environment info

- WordPress version 6.6 RC3
- Reproduced on any theme
- The only active plugin is WooCommerce

Yes

Yes

jsnajdr:
Fixes #63448. When `useBlockElement` is used by a newly mounted component, return `null` until an effect sets the current value from the store. That avoids using a stale element value.

**How to test:**
Go through the reproduction steps in https://github.com/WordPress/gutenberg/issues/63448#issuecomment-2228278478 and verify that the bug is indeed fixed. No more crashes when switching from Tablet back to Desktop view.

### DIFF
--- a/packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
+++ b/packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
@@ -1,8 +1,14 @@
 /**
  * WordPress dependencies
  */
-import { useContext, useMemo, useRef } from '@wordpress/element';
-import { useRefEffect, useObservableValue } from '@wordpress/compose';
+import {
+	useContext,
+	useMemo,
+	useRef,
+	useState,
+	useLayoutEffect,
+} from '@wordpress/element';
+import { useRefEffect } from '@wordpress/compose';
 
 /**
  * Internal dependencies
@@ -67,7 +73,17 @@ function useBlockRef( clientId ) {
  */
 function useBlockElement( clientId ) {
 	const { refsMap } = useContext( BlockRefs );
-	return useObservableValue( refsMap, clientId ) ?? null;
+	const [ blockElement, setBlockElement ] = useState( null );
+	// Delay setting the resulting `blockElement` until an effect. If the block element
+	// changes (i.e., the block is unmounted and re-mounted), this allows enough time
+	// for the ref callbacks to clean up the old element and set the new one.
+	useLayoutEffect( () => {
+		setBlockElement( refsMap.get( clientId ) );
+		return refsMap.subscribe( clientId, () =>
+			setBlockElement( refsMap.get( clientId ) )
+		);
+	}, [ refsMap, clientId ] );
+	return blockElement;
 }
 
 export { useBlockRef as __unstableUseBlockRef };


ellatrix:
I changed the labels because to be explicit here: ship has sailed for 6.6, this can only be include in the next minor release.

talldan (commented):
I can confirm that this fixes the bug.

I don't fully understand why the change is required, from what I can tell, the implementation of `useBlockElement` in trunk doesn't seem to ever return a detached element.

It'd be great to understand this better. 😄

jsnajdr:
> It'd be great to understand this better.

When a block renders, it registers its top DOM element inside a clientId->element map in the `BlockRefsProvider`. The `useBlockRefProvider( clientId )` call inside `useBlockProps` does this. It does by passing a ref callback to the block's top element and adding the element to the map in the callback.

Anyone can then retrieve the block's element with `useBlockElement( clientId )`. This is used by various popovers and overlays, like `GridVisualizer`. `useBlockElement` reads the value from the map and immediately returns it.

When switching from iframe to non-iframe view, a block is unmounted and immediately remounted, at another DOM location with a new top element. Before it was rendered like:
```jsx
<div className="editor-canvas">
  <iframe>
    <Block clientId={ clientId } />
  </iframe>
</div>
```
and now it's rendered like:
```jsx
<div className="editor-canvas">
  <Block clientId={ clientId } />
</div>
```

Now we are getting to the caveat 🙂 When doing the second render, without iframe, this is the order of events:
1. The block's render function is called. It reads the element, but `useBlockElement` returns the old element in the iframe. The new element doesn't even exist yet.
2. React compares the markup returned by the render function with what is rendered and figures out it needs to destroy the iframe and create a new DOM structure.
3. Before React destroys the iframe, it calls the ref callback with `null`. Only at this time the old element is unregistered from the map. But alas, the rendered markup already has a reference to it.
4. Now React does all the DOM changes. Destroy old elements, create new ones.
5. React calls the ref callback with the newly created element. It's registered in the map, the map calls all listeners, and a `setState` is scheduled for the block component that called `useBlockElement`. But it will read the new element only on the next render.
6. Effect cleanup functions for the old components are called, effects for the new components are called.
7. One of things that floating-ui's `useFloating` hook does on mount (in an effect) is to [call `whileElementsMounted`](https://github.com/floating-ui/floating-ui/blob/81a7da464c713e3432d1015cafb7566675a22282/packages/react-dom/src/useFloating.ts#L127-L129). But the `referenceEl` it's called with is still the old element in the iframe.

My solution is that on initial render, `useBlockElement` always returns `null`. Only in an effect, after all the ref callbacks have been called and the new DOM is settled down, is `useBlockElement` to read the current value from the map.

It's very similar to what you do when you want to use component's own element to render it:
```jsx
function Block() {
  const [ el, setEl ] = useState( null );
  return <>
    <div ref={ setEl }>Hello</div>
    <Tooltip anchorEl={ el }>World</Tooltip>
  </>
}
```
On initial render, `el` is always `null`, it cannot be anything else, because the element doesn't exist yet. The tooltip is initially rendered without `anchorEl`. Only after the ref callback is called, a second render is performed (triggered by `setEl`), this time with the mounted element.

Why does the bug happen only with `GridVisualizer`, and not with a `BlockToolbar`, for example? Because `GridVisualizer` is rendered inside the iframe, and after switching to non-iframe it's destroyed and created again. There is a mount effect, and the offending `whileElementsMounted` is called.

On the other hand, the block toolbar is rendered outside the iframe, in a slot. It's not destroyed and created, but merely rerendered with new props. The `useBlockElement` update is delayed, but `whileElementsMounted` is not called during that delay.

ciampo (approved):
🚀  LGTM

Thank you @jsnajdr for debugging, fixing, and explaining in detail the root cause of the bug.

packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
```diff
@@ -67,7 +73,17 @@ function useBlockRef( clientId ) {
...
 	const { refsMap } = useContext( BlockRefs );
-	return useObservableValue( refsMap, clientId ) ?? null;
+	const [ blockElement, setBlockElement ] = useState( null );
+	// Delay setting the resulting `blockElement` until an effect. If the block element
+	// changes (i.e., the block is unmounted and re-mounted), this allows enough time
+	// for the ref callbacks to clean up the old element and set the new one.
+	useLayoutEffect( () => {
+		setBlockElement( refsMap.get( clientId ) );
+		return refsMap.subscribe( clientId, () =>
+			setBlockElement( refsMap.get( clientId ) )
```
ellatrix, line 83: On unmount, shouldn't it be set to `null`?

jsnajdr: I was considering this, but I think it's not needed, it wouldn't have any effect. On unmount, the component is going away, and updating its local state is not needed, because nobody is going to read it.

But maybe we can add it for consistency sake. So that the behavior is indistinguishable from the classic "store ref as state" pattern:
```js
const [ el, setEl ] = useState( null );
return <div ref={ setEl } />;
```
In this case the `setEl` will be called with `null` on unmount.

`useBlockRefs` is basically the same pattern, store an element ref into state, but the ref and the state are in different components, and the store in the provider serves as a teleportation machine between them.

ellatrix: Wouldn't it be possible to for a block to disappear, while another component is using `useBlockElement`, which would then still point to an unmounted DOM node?

jsnajdr: Before the block disappears, the ref callback is called and sets the ref to `null`. Then the observable map calls its listeners and inside the `useBlockElement` hook, the `setBlockElement` inside the listener is called, and sets the local state to `null`. Therefore, all the references to the unmounted DOM node are cleared.

jsnajdr: This issue makes me thinking that our current API for reading the elements is not very good. Ideally it should work exactly like element refs. Like this:
```jsx
const blockElRef = useRef(); // can also be a ref callback
useBlockElementRef( clientId, blockElRef );
```
The `blockElRef` would be set and unset in real time as the target element changes. The point is that it is identical to how refs to local elements work:
```jsx
const blockElRef = useRef(); // can also be a ref callback
return <div ref={ blockElRef } />;
```
Exactly the same thing, only in one case the element is local and in the other it's teleported from another location.

This is much better than the current `useBlockRef` hook that _returns_ a ref with a synthetic `ref.current` getter.

The `useBlockElementRef` primitive can be used to very easily implement `useBlockElement` that returns the element and triggers a rerender on change:
```jsx
function useBlockElement( clientId ) {
  const [ el, setEl ] = useState( null );
  useBlockElementRef( clientId, setEl );
  return el;
}
```
This implementation also never has the bug with returning stale unmounted elements.

ellatrix: Oh, sorry, I misread this as a change to the hook inside the block rather than a change to the hook in the consumer.

Yes, that alternative seems good to me. We should try it :)

jsnajdr: Implemented in #63799 🙂

PR #61648, opened 2024-05-14
Add e2e tests for keyboard interactions in DataViews ListView

oandregal:
Follow-up to https://github.com/WordPress/gutenberg/pull/60805

## What?

Adds a few e2e test for the keyboard interactions merged by https://github.com/WordPress/gutenberg/pull/60805

## Why?

We want to make sure these don't break.

## How?

A few different tests that cover all scenarios targeted in the PR above:

- Items list is reachable via TAB.
- Navigates from items list to preview via TAB, and vice versa
- Navigates the items list via UP/DOWN arrow keys
- Actions are reachable via RIGHT/LEFT arrow keys
- Navigates the list via UP/DOWN arrow keys from action buttons

## How to test

Make sure tests pass in CI. They can be executed locally by `npm run test:e2e:playwright -- test/e2e/specs/site-editor/dataviews-list-layout-keyboard.spec.js`.

PR #60805, opened 2024-04-17
DataViews: add actions to list layout

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/59659 and https://github.com/WordPress/gutenberg/issues/55083
Follow-up to https://github.com/WordPress/gutenberg/pull/59637
Fixes https://github.com/WordPress/gutenberg/issues/61555

## What?

This PR adds support for actions in list layout, which is only enabled for Pages and Templates as of now:

| Pages | Templates |
| --- | --- |
| Captura de ecrã 2024-05-10, às 11 49 29 |  Captura de ecrã 2024-05-10, às 11 49 23 |

This PR also enables two-dimensional arrow-key navigation:

https://github.com/WordPress/gutenberg/assets/583546/81c24287-3f53-4744-bbc6-efd7948e0e2f

## Why?

We want to make them available. See https://github.com/WordPress/gutenberg/issues/59659

## How?

- Pass the existing `actions` prop that any other layout is using to the list view layout.
- Add two-dimensional arrow navigation to the list items, so users can use right/left arrow keys to move between the element and its actions.

## Testing Instructions

- Go to a page with list view (Pages or Templates) and verify the actions are present.
- Navigate via keyboard to verify that it works as expected:
  - The item list is a single tab stop.
  - You can vertically through the list via up/down arrow keys.
  - You can move horizontally within a single item via right/left arrow keys.

PR #63569, opened 2024-07-15
Composite: stabilize APIs

ciampo:
## What?

Requires #63564, #64399, #64397, #64450, #64493 and #64723

Part of https://github.com/WordPress/gutenberg/issues/58850

Export `Composite` as a stable API and refactor usages of the component from using the private APIs to using the public ones..

Also, remove the private APIs since the component is now public.

## Why?

This is one of the final steps in making the `Composite` component stable and available as a public API of `@wordpress/components`.

### Main changes

- Export the `Composite` component from the `@wordpress/components` package as stable APIs
- Remove the equivalent private API
- Replace all imports via private APIs with public APIs;
- Update component naming as necessary, using the new "overloaded" naming convention adopted for compound components;

## Testing Instructions

Given that the project builds and tests pass, most of the changes don't really need testing, since the underlying composite components are the exact same (we only changed the way they are imported).

The only real testing that should be done is around the changes where the unnecessary `store` and `onFocusVisible` props were removed.

### A new stable `Composite` component

Following the rewrite of the `__unstableComposite`  (& related) components from `ariakit` to `reakit` in the WordPress 6.5 release, the 6.7 release introduces a new, stable version for the `Composite` component.

The new `Composite` component offers a more complete and versatile way to implement low-level composite widgets, and its APIs are more consistent with the rest of the `@wordpress/components` package.

```tsx
import { Composite } from '@wordpress/components';

<Composite>
  <Composite.Item>Item one</Composite.Item>
  <Composite.Item>Item two</Composite.Item>
  <Composite.Item>Item three</Composite.Item>
</Composite>
```

For more information of the component's APIs and other usage examples, please refer to [the Storybook page](https://wordpress.github.io/gutenberg/?path=/docs/components-composite--docs).

Finally, the `__unstableComposite`  (& related) components have been marked as deprecated.

### DIFF
--- a/packages/components/src/composite/stories/index.story.tsx
+++ b/packages/components/src/composite/stories/index.story.tsx
@@ -15,7 +15,7 @@ import { createSlotFill, Provider as SlotFillProvider } from '../../slot-fill';
 import { Composite } from '..';
 
 const meta: Meta< typeof Composite > = {
-	title: 'Components/Composite (V2)',
+	title: 'Components/Composite',
 	component: Composite,
 	subcomponents: {
 		// @ts-expect-error - See https://github.com/storybookjs/storybook/issues/23170
@@ -46,7 +46,6 @@ const meta: Meta< typeof Composite > = {
 			options: [ true, false, 'horizontal', 'vertical', 'both' ],
 		},
 	},
-	tags: [ 'status-private' ],
 	parameters: {
 		controls: { expanded: true },
 		docs: {
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -66,6 +66,7 @@ export {
 	CompositeItem as __unstableCompositeItem,
 	useCompositeState as __unstableUseCompositeState,
 } from './composite/legacy';
+export { Composite } from './composite';
 export { ConfirmDialog as __experimentalConfirmDialog } from './confirm-dialog';
 export { default as CustomSelectControl } from './custom-select-control';
 export { default as Dashicon } from './dashicon';
--- a/packages/components/src/private-apis.ts
+++ b/packages/components/src/private-apis.ts
@@ -1,7 +1,6 @@
 /**
  * Internal dependencies
  */
-import { Composite } from './composite';
 import { positionToPlacement as __experimentalPopoverLegacyPositionToPlacement } from './popover/utils';
 import { createPrivateSlotFill } from './slot-fill';
 import { DropdownMenuV2 } from './dropdown-menu-v2';
@@ -13,12 +12,6 @@ import { lock } from './lock-unlock';
 
 export const privateApis = {};
 lock( privateApis, {
-	CompositeV2: Composite,
-	CompositeGroupV2: Composite.Group,
-	CompositeItemV2: Composite.Item,
-	CompositeRowV2: Composite.Row,
-	CompositeTypeaheadV2: Composite.Typeahead,
-	CompositeHoverV2: Composite.Hover,
 	__experimentalPopoverLegacyPositionToPlacement,
 	createPrivateSlotFill,
 	ComponentsContext,
--- a/packages/dataviews/src/components/dataviews-filters/search-widget.tsx
+++ b/packages/dataviews/src/components/dataviews-filters/search-widget.tsx
@@ -11,26 +11,14 @@ import removeAccents from 'remove-accents';
 import { useInstanceId } from '@wordpress/compose';
 import { __, sprintf } from '@wordpress/i18n';
 import { useState, useMemo, useDeferredValue } from '@wordpress/element';
-import {
-	VisuallyHidden,
-	Icon,
-	privateApis as componentsPrivateApis,
-} from '@wordpress/components';
+import { VisuallyHidden, Icon, Composite } from '@wordpress/components';
 import { search, check } from '@wordpress/icons';
 import { SVG, Circle } from '@wordpress/primitives';
 
 /**
  * Internal dependencies
  */
-import { unlock } from '../../lock-unlock';
 import type { Filter, NormalizedFilter, View } from '../../types';
-
-const {
-	CompositeV2: Composite,
-	CompositeItemV2: CompositeItem,
-	CompositeHoverV2: CompositeHover,
-	CompositeTypeaheadV2: CompositeTypeahead,
-} = unlock( componentsPrivateApis );
 
 interface SearchWidgetProps {
 	view: View;
@@ -136,13 +124,13 @@ function ListBox( { view, filter, onChangeView }: SearchWidgetProps ) {
 					);
 				}
 			} }
-			render={ <CompositeTypeahead /> }
+			render={ <Composite.Typeahead /> }
 		>
 			{ filter.elements.map( ( element ) => (
-				<CompositeHover
+				<Composite.Hover
 					key={ element.value }
 					render={
-						<CompositeItem
+						<Composite.Item
 							id={ generateFilterElementCompositeItemId(
 								baseId,
 								element.value
@@ -212,7 +200,7 @@ function ListBox( { view, filter, onChangeView }: SearchWidgetProps ) {
 							) }
 					</span>
 					<span>{ element.label }</span>
-				</CompositeHover>
+				</Composite.Hover>
 			) ) }
 		</Composite>
 	);
--- a/packages/dataviews/src/dataviews-layouts/list/index.tsx
+++ b/packages/dataviews/src/dataviews-layouts/list/index.tsx
@@ -14,6 +14,7 @@ import {
 	privateApis as componentsPrivateApis,
 	Spinner,
 	VisuallyHidden,
+	Composite,
 } from '@wordpress/components';
 import {
 	useCallback,
@@ -48,12 +49,7 @@ interface ListViewItemProps< Item > {
 	onDropdownTriggerKeyDown: React.KeyboardEventHandler< HTMLButtonElement >;
 }
 
-const {
-	CompositeV2: Composite,
-	CompositeItemV2: CompositeItem,
-	CompositeRowV2: CompositeRow,
-	DropdownMenuV2: DropdownMenu,
-} = unlock( componentsPrivateApis );
+const { DropdownMenuV2: DropdownMenu } = unlock( componentsPrivateApis );
 
 function generateItemWrapperCompositeId( idPrefix: string ) {
 	return `${ idPrefix }-item-wrapper`;
@@ -92,7 +88,7 @@ function PrimaryActionGridCell< Item >( {
 
 	return 'RenderModal' in primaryAction ? (
 		<div role="gridcell" key={ primaryAction.id }>
-			<CompositeItem
+			<Composite.Item
 				id={ compositeItemId }
 				render={
 					<Button
@@ -111,11 +107,11 @@ function PrimaryActionGridCell< Item >( {
 						closeModal={ () => setIsModalOpen( false ) }
 					/>
 				) }
-			</CompositeItem>
+			</Composite.Item>
 		</div>
 	) : (
 		<div role="gridcell" key={ primaryAction.id }>
-			<CompositeItem
+			<Composite.Item
 				id={ compositeItemId }
 				render={
 					<Button
@@ -192,7 +188,7 @@ function ListItem< Item >( {
 	) : null;
 
 	return (
-		<CompositeRow
+		<Composite.Row
 			ref={ itemRef }
 			render={ <li /> }
 			role="row"
@@ -209,7 +205,7 @@ function ListItem< Item >( {
 				spacing={ 0 }
 			>
 				<div role="gridcell">
-					<CompositeItem
+					<Composite.Item
 						render={ <div /> }
 						role="button"
 						id={ generateItemWrapperCompositeId( idPrefix ) }
@@ -260,7 +256,7 @@ function ListItem< Item >( {
 								</div>
 							</VStack>
 						</HStack>
-					</CompositeItem>
+					</Composite.Item>
 				</div>
 				{ eligibleActions?.length > 0 && (
 					<HStack
@@ -282,7 +278,7 @@ function ListItem< Item >( {
 						<div role="gridcell">
 							<DropdownMenu
 								trigger={
-									<CompositeItem
+									<Composite.Item
 										id={ generateDropdownTriggerCompositeId(
 											idPrefix
 										) }
@@ -311,7 +307,7 @@ function ListItem< Item >( {
 					</HStack>
 				) }
 			</HStack>
-		</CompositeRow>
+		</Composite.Row>
 	);
 }
 
--- a/packages/dataviews/src/dataviews-layouts/list/index.tsx
+++ b/packages/dataviews/src/dataviews-layouts/list/index.tsx
@@ -140,7 +140,7 @@ function ListItem< Item >( {
 	visibleFields,
 	onDropdownTriggerKeyDown,
 }: ListViewItemProps< Item > ) {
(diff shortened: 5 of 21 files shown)


packages/components/src/index.ts
```diff
@@ -62,7 +62,7 @@ export {
 	CompositeItem as __unstableCompositeItem,
 	useCompositeState as __unstableUseCompositeState,
 } from './composite/legacy';
-export { Composite } from './composite';
+export { Composite, useCompositeStore } from './composite';
```
ciampo: I had actually forgotten to export `useCompositeStore` in previous PRs 😅

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
```diff
@@ -107,19 +101,20 @@ function ListBox( { view, filter, onChangeView }: SearchWidgetProps ) {
 				__( 'List of: %1$s' ),
 				filter.name
 			) }
+			// @ts-expect-error check if this actually works, since `onFocusVisible`
+			// only works when `focusable` is `true`.
 			onFocusVisible={ () => {
 				if ( ! compositeStore.getState().activeId ) {
 					compositeStore.move( compositeStore.first() );
 				}
 			} }
```
ciampo: Hey @ntsekouras could you help me understand if we can actually remove the `onFocusVisible` callback here? That's because in theory, from what I understand from [the docs](https://ariakit.org/reference/composite#onfocusvisible), it shouldn't work without the `focusable` prop.

Does it currently work? If not, should we remove it altogether, or try to make it work by also passing the `focusable` prop?

Thank you!

packages/dataviews/src/dataviews-layouts/list/index.tsx
```diff
@@ -48,17 +46,11 @@ interface ListViewItemProps< Item > {
 	mediaField?: NormalizedField< Item >;
 	onSelect: ( item: Item ) => void;
 	primaryField?: NormalizedField< Item >;
-	store: CompositeStore;
+	store: React.ComponentProps< typeof Composite >[ 'store' ];
 	visibleFields: NormalizedField< Item >[];
 }
 
-const {
```
oandregal, line 51: The changes for this component seem to be one-to-one according to the names. However, something is different. The keyboard navigation for the list layout is broken in this PR (see also e2e test). Note how the dropdown action menu is not reachable via right arrow:

https://github.com/user-attachments/assets/d726753c-6c5a-4e59-8eb5-37ed65e20608

References, in case it helps:

- implementation https://github.com/WordPress/gutenberg/pull/60805
- reported issues with ariakit to make 2d keyboard navigation work https://github.com/ariakit/ariakit/issues/3768
- e2e test  https://github.com/WordPress/gutenberg/pull/61648

ciampo: Could it be because I also removed passing the `store` prop to `Composite.Item` twice in this same file ? If you add it back, does it solve the issue for you?

ciampo: Update: passing explicitly the `store` to `Composite.Item` seems to fix it, although that's an unexpected behaviour from my side that I'd like to investigate further. I'll post a new inline comment on that specific line.

packages/dataviews/src/dataviews-layouts/list/index.tsx
```diff
@@ -265,7 +254,9 @@ function ListItem< Item >( {
 						<div role="gridcell">
 							<DropdownMenu
 								trigger={
-									<CompositeItem
+									<Composite.Item
+										// @ts-expect-error Passing the store here is necessary
+										// in order to override the dropdown menu's store.
 										store={ store }
```
ciampo: For now, we have to continue passing explicitly a `store` object to this `Composite.Item` instance — without it, the the composite item is not included in the parent composite widget (see [conversation](https://github.com/WordPress/gutenberg/pull/63569#discussion_r1713359138)).

After some debugging, it looks like the explicit `store` override is necessary because otherwise this instance of `Composite.Item` gets the `DropdownMenu`'s store instead of the parent `Composite` store when reading the store from context.

@diegohaz , is this the expected behaviour (ie. passing `store` explicitly), or is this due to how `DropdownMenu` is implemented ([code](https://github.com/WordPress/gutenberg/blob/039bd9081abeb7ce93d7a57e56ca0abfbc6b7937/packages/components/src/dropdown-menu-v2/index.tsx#L309)) ? I could replicate it also with vanilla Ariakit ([stackblitz](https://stackblitz.com/edit/grzwwc?file=menu%2Findex.tsx)), but I wonder if I'm missing something.

Thank you!

packages/components/src/composite/stories/index.story.tsx
```diff
@@ -230,7 +230,30 @@ This only affects the composite widget behavior. You still need to set \`dir="rt
 					'Composite.Group': commonArgTypes,
 					'Composite.GroupLabel': commonArgTypes,
 					'Composite.Row': commonArgTypes,
-					'Composite.Item': commonArgTypes,
+					'Composite.Item': {
+						...commonArgTypes,
+						accessibleWhenDisabled: {
```
tyxla: According to the Ariakit docs, the `accessibleWhenDisabled` prop is also available and supported for the `Composite` element. Should we add it there as well?

packages/components/src/composite/types.ts
```diff
@@ -177,6 +177,22 @@ export type CompositeItemProps = {
...
+	 * > A toolbar in an editor contains a set of special smart paste functions
+	 * that are disabled when the clipboard is empty or when the function is not
+	 * applicable to the current content of the clipboard. It could be helpful to
+	 * keep the disabled buttons focusable if the ability to discover their
+	 * functionality is primarily via their presence on the toolbar.
+	 *
+	 * Learn more on [Focusability of disabled
+	 * controls](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols).
+	 */
+	accessibleWhenDisabled?: Ariakit.CompositeItemProps[ 'accessibleWhenDisabled' ];
```
tyxla: Same as above, wrt providing the prop type for `Composite`.

packages/components/src/index.ts
tyxla: Good timing to find out about it 😅

packages/components/src/private-apis.ts
```diff
@@ -22,11 +21,6 @@ import { lock } from './lock-unlock';
 
 export const privateApis = {};
 lock( privateApis, {
-	CompositeV2: Composite,
-	CompositeGroupV2: Composite.Group,
-	CompositeItemV2: Composite.Item,
-	CompositeRowV2: Composite.Row,
-	useCompositeStoreV2: useCompositeStore,
```
tyxla: Glad to see `@wordpress/components` private APIs are down with over 25% 👏

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
tyxla: From what reading I've done, I think adding `focusable` should do it. It doesn't seem to work in `trunk` for me in my testing right now.

tyxla (approved):
Aside from `onFocusVisible` + `focusable`, this looks great to me 👍 

Nice work @ciampo 🚀 

Let's make sure to add a CHANGELOG entry before 🚢 .

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
```diff
@@ -101,9 +101,10 @@ function ListBox( { view, filter, onChangeView }: SearchWidgetProps ) {
 				__( 'List of: %1$s' ),
 				filter.name
 			) }
-			// @ts-expect-error check if this actually works, since `onFocusVisible`
-			// only works when `focusable` is `true`.
 			onFocusVisible={ () => {
+				// `onFocusVisible` needs the `Composite` component to be focusable,
+				// which is implicitly achieved via the `virtualFocus: true` option
+				// in the `useCompositeStore` hook.
```
tyxla: 👍

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
ciampo: In my experience, testing those lines of code was tricky because of some different behaviors based on the number of items and the operators' lenght, but I wasn't able to notice regressions. What regressions have you noticed?

Since `onFocusVisible` is necessary, I went ahead and added it to `Composite`'s props, alongside a few focus-related props (`focusable`, `disabled`, `accessibleWhenDisabled`, `onFocusVisible`).

--

Specifically to why `onFocusVisible` is being triggered even without an explicit `focusable` prop, I'm going to continue the investigation in [a separate conversation](https://github.com/WordPress/gutenberg/pull/63569#discussion_r1714043891) to get better clarity.

packages/components/src/composite/stories/index.story.tsx
ciampo: Added, together with other props, as explained in https://github.com/WordPress/gutenberg/pull/63569#discussion_r1714041967

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
ciampo: @diegohaz can you confirm that my conclusions are correct? Is this the expected behavior? Should we pass `focusable` to `Composite` to future-proof this behavior?

ciampo:
> this looks great to me 👍

Thank you! I'd like to get clarity both on the [`store`](https://github.com/WordPress/gutenberg/pull/63569#discussion_r1713930889) and [`onFocusVisible`](https://github.com/WordPress/gutenberg/pull/63569#discussion_r1714043891) prop behaviors before merging, assuming we're going to get a reply soon.

> Let's make sure to add a CHANGELOG entry before 🚢 .

I'm also considering opening a separate PR with the changes to `Composite`, so that we keep this PR focused on the rest of the repository (#64450)

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
tyxla: Actually, it works as expected, both in trunk and here:

https://github.com/user-attachments/assets/07b735ae-0167-4935-8d42-8a75d4342a5c

so we should be good 👍

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
diegohaz: The description makes sense.

> Should we pass `focusable` to `Composite` to future-proof this behavior?

Not all composite containers are focusable, though. For example, a toolbar container is not focusable and therefore doesn't trigger `onFocusVisible`.

packages/dataviews/src/dataviews-layouts/list/index.tsx
diegohaz: That's "expected" because the Ariakit Menu component also provides a Composite context. I acknowledge this is desirable in some cases but cumbersome in others, so this may change in the future.

Related issue: [https://github.com/ariakit/ariakit/issues/3754](https://github.com/ariakit/ariakit/issues/3754)

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
diegohaz: You can think of `focusable`/`onFocusVisible` as similar to `draggable`/`onDragStart`.

tyxla:
#64450 looks good 👍 Let's merge it, and this one will be much smaller. I'm happy to re-review once it's ready.

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
ntsekouras: Sorry for the late reply. It seems you figured it out already 😄 . The conditional rendering of different components and what designers wanted led to this and similar nuances like [this one](https://github.com/WordPress/gutenberg/blob/trunk/packages/dataviews/src/components/dataviews-filters/search-widget.tsx#L90).

packages/dataviews/src/dataviews-layouts/list/index.tsx
ciampo: Ah, that explains it! I didn't consider that different components could internally use another composite store instance.

I think that the best approach for us here is to use a custom `Composite` context to make sure that all sub-components consume the correct store (it's the same approach that we adopted in [`Tabs`](https://github.com/WordPress/gutenberg/blob/039bd9081abeb7ce93d7a57e56ca0abfbc6b7937/packages/components/src/tabs/index.tsx#L198) and [`DropdownMenuV2`](https://github.com/WordPress/gutenberg/blob/039bd9081abeb7ce93d7a57e56ca0abfbc6b7937/packages/components/src/dropdown-menu-v2/index.tsx#L331))

I tested those changes locally and they work well, and I opened a separate PR to land them: #64493

I already removed the explicit `store` prop from `Composite.Item`, and will rebase on top of `trunk` once #64493 is merged. We should be ready to land this PR by then.

packages/dataviews/src/components/dataviews-filters/search-widget.tsx
ciampo: > > Should we pass `focusable` to `Composite` to future-proof this behavior?
> 
> Not all composite containers are focusable, though.

That makes sense! Although my question was specific to this instance of `Composite`. My inline code comment is hopefully enough to explain why `onFocusVisible` is being called despite the `focusable` prop is not set on the component.

packages/dataviews/src/dataviews-layouts/list/index.tsx
tyxla: An internal context makes sense for passing down the store.  👍 

#64493 looks good! ✅ 

Noting that there appears to be one more instance where we explicitly pass `store` to a `Composite.Item`:

https://github.com/WordPress/gutenberg/blob/2c62c6aa2f30060b3acd40b225aebe4295ad0bf4/packages/block-directory/src/components/downloadable-block-list-item/index.js#L117

Might make sense to address it here before landing.

ciampo:
Update: with #64493 merged, this PR should be good to be merged too. I'll wait for CI, run some extra manual tests, and merge if everything looks good.

tyxla:
Were you also planning to address this before merging:

https://github.com/WordPress/gutenberg/pull/63569#discussion_r1716812603

packages/dataviews/src/dataviews-layouts/list/index.tsx
ciampo: I had somewhat missed that instance (and your comment). Thank you for flagging again

tyxla (changes requested):
Let's hold off merging this for now until we settle how we'll deal with some of the public API considerations, see #64510 and #63704.

ciampo:
@WordPress/gutenberg-components This PR is now ready for a new round of review, after addressing all the necessary API changes in #64723 & related.

packages/components/src/private-apis.ts
```diff
@@ -13,12 +12,6 @@ import { lock } from './lock-unlock';
 
 export const privateApis = {};
 lock( privateApis, {
-	CompositeV2: Composite,
-	CompositeGroupV2: Composite.Group,
-	CompositeItemV2: Composite.Item,
-	CompositeRowV2: Composite.Row,
-	CompositeTypeaheadV2: Composite.Typeahead,
-	CompositeHoverV2: Composite.Hover,
```
tyxla, line 21: So cool to see `@wordpress/components` private APIs going down by half 👏

tyxla (approved):
Did some testing in the editor and in Storybook, everything works well!

Let's go! 🚀 

Great work addressing it incrementally @ciampo!

PR #59890, opened 2024-03-15
Fix lightbox UI disallow editing

artemiomorales:
## What?
This PR fixes an issue wherein the `allowEditing` value of the lightbox was not resulting in a consistent UX. It also addresses other related edge cases.

## Why?
Fixes https://github.com/WordPress/gutenberg/issues/59837
Fixes https://github.com/WordPress/gutenberg/issues/59797

When one disables editing of the lightbox, the expectation is that the UI related to the lightbox config should be seamlessly removed from the editor.

## How?
It adds logic to handle the case wherein the lightbox animation is enabled by default but editing of the lightbox is not.
It also covers various edge cases regarding how the block-level settings interact with the `theme.json` settings.

## Testing Instructions
To make sure this is working properly, We need to test various scenarios and edge cases.
In particular, we need to test the UI in the following scenarios:

1. Image block DOES NOT have a link configured, and...
- DOES NOT have a lightbox override
- DOES have a lightbox override of ENABLED set to TRUE
- DOES have a lightbox override of ENABLED set to FALSE
2. Image block DOES have a link configured, and...
- DOES NOT have a lightbox override
- DOES have a lightbox override of ENABLED set to TRUE
- DOES have a lightbox override of ENABLED set to FALSE

In addition, each one of the scenarios above needs to be tested with the following `theme.json` configurations:

**allowEditing _false_, enabled _false_**
```
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 2,
	"settings": {
		"blocks": {
			"core/image": {
				"lightbox": {
					"allowEditing": false,
					"enabled": false
				}
			}
		}
	}
}
```

**allowEditing _true_, enabled _false_**
```
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 2,
	"settings": {
		"blocks": {
			"core/image": {
				"lightbox": {
					"allowEditing": true,
					"enabled": false
				}
			}
		}
	}
}
```

**allowEditing _false_, enabled _true_**
```
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 2,
	"settings": {
		"blocks": {
			"core/image": {
				"lightbox": {
					"allowEditing": false,
					"enabled": true
				}
			}
		}
	}
}
```

**allowEditing _true_, enabled _true_**
```
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 2,
	"settings": {
		"blocks": {
			"core/image": {
				"lightbox": {
[...]

PR #57608, opened 2024-01-05
Image block: Move UI for lightbox from sidebar to the content toolbar alongside link settings

artemiomorales:
## What?

This PR moves the UI for the image lightbox from the sidebar's block settings to the context toolbar so that it can exist alongside the rest of the link logic.

## Why?

Addresses https://github.com/WordPress/gutenberg/issues/54916
Right now the UX for the lightbox is divorced from the rest of the handling for links.
This consolidates it and makes it more intuitive.

## How?

In the `image-url-input-ui` component, this PR adds the logic, markup, and styles for the new design and `Expand on click` toolbar item. Also, in `image.js`, I modified the logic for reading from or overriding the lightbox setting from the global styles.

#### With global styles `expand on click` setting _enabled_
1. In the global styles for the image block, ensure that `Expand on click` is enabled
2. Add an image to a post
3. Click the image and view the link settings; see that the `Expand on click` option is active
4. Publish and view the post; ensure the lightbox works as expected
6. In the post editor, click the image, open the link settings, and remove `Expand on click`
7. Check the post code source and see that `"lightbox":{"enabled":false}` has been added in the block delimiter
8. Save and view the post; ensure the lightbox is now disabled
9. In the post editor, reenable `Expand on click`
10. Check the code source again and verify `"lightbox":{"enabled":false}` has been removed

#### With global styles `expand on click` setting _disabled_
(same as above, but in reverse)
1. In the global styles for the image block, ensure that `Expand on click` is disabled
2. Add an image to a post
3. Click the image and view the link settings; see that the `Expand on click` option is deactivated
4. Publish and view the post; ensure the lightbox does not open
6. In the post editor, click the image, open the link settings, and enable `Expand on click`
7. Check the post code source and see that `"lightbox":{"enabled":true}` has been added in the block delimiter
8. Save and view the post; ensure the lightbox is now enabled
9. In the post editor, redeactivate `Expand on click`
10. Check the code source again and verify that the `"lightbox":{"enabled":true}` has been removed

### Site editor

For the site editor, please test the same as above. In addition, verify that toggling the global styles `Expand on ciick` setting also toggles the link UI for the image.

toggle-lightbox-global-styles

_Also_, check that adding and removing `Link to image file`, `Link to attachment page`, and custom links still works as expected.

### Testing Instructions for Keyboard

This uses existing UI and should work as expected with standard keyboard navigation.

## Screenshots or screencast 

[...]

PR #63573, opened 2024-07-15
ImageURLInputUI: Make onSetLightbox and resetLightbox optional

t-hamano:
Related to #57608, #59890

## What?
This PR makes `onSetLightbox` and `resetLightbox` optional.

## Why?

This component is public and the `onSetLightbox` and `resetLightbox` props are not required, but these two functions may be called, so the following error will be logged in the browser:

```
image-url-input-ui.js:305 Uncaught TypeError: resetLightbox is not a function
```

## How?

Add the optional chaining to the two functions.

## Testing Instructions

Currently, the only core block that will cause this error is the Media & Text block, because [it doesn't have a `resetLightbox` prop](https://github.com/WordPress/gutenberg/blob/c54c0020a7a8ebf22738aa96cb398ea4cf17a30a/packages/block-library/src/media-text/edit.js#L464-L474) that fires whenever you remove a link.

- Insert a Media & Text block.
- Add and remove the link.
- There should be no errors in the browser console.

### DIFF
--- a/packages/block-editor/src/components/url-popover/image-url-input-ui.js
+++ b/packages/block-editor/src/components/url-popover/image-url-input-ui.js
@@ -272,7 +272,7 @@ const ImageURLInputUI = ( {
 						icon={ linkOff }
 						label={ __( 'Disable expand on click' ) }
 						onClick={ () => {
-							onSetLightbox( false );
+							onSetLightbox?.( false );
 						} }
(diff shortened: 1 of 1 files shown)


michalczaplinski (approved):
Makes total sense to me! 

Thanks for the PR 🙂

PR #63473, opened 2024-07-12
DataViews: Only show elligible actions in the bulk editing menu

youknowriad:
closes #63472 

## What?

This PR updates the bulk editing menu in the dataviews to only show the actions that can be applied to the current selection (or part of the current selection). Showing all the registered action can lead to too many irrelevant actions.

## Testing Instructions

- Select a list of pages in the pages dataviews.
- Open the bulk edit menu.
- Only relevant actions are shown.

PR #63575, opened 2024-07-15
DataViews: Do not render bulk actions Dropdown if no actions are available

ntsekouras:
## What?
Small follow up of: https://github.com/WordPress/gutenberg/pull/63473

## Testing instructions
If have no item selected and click the `bulk actions` button in Data Views lists, we should render an empty Dropdown and the separator.


Mamaduka (approved):
Makes sense ✅

Issue #44750, opened 2022-10-06
Improve pattern rendering sequence


mtias:
At the moment, patterns seem to go through a few rendering adjustments before settling into their final form. I'm spotting at least three but there could be more.

It'd be great to solve this as it makes the pattern inserter and the pattern experience feel more fragile.

| First Render  | Second Render | Third Render |
| ------------- | ------------- | ------------- |
| <img width="299" alt="image" src="https://user-images.githubusercontent.com/548849/194334629-19d0151c-d370-493d-863d-13de263777e6.png">  | <img width="300" alt="image" src="https://user-images.githubusercontent.com/548849/194334708-a7ed3ed0-343c-48e6-89df-400e7188a4fa.png">  | <img width="300" alt="image" src="https://user-images.githubusercontent.com/548849/194334785-18b096ac-56c2-4026-b379-92a6cdde2ea1.png">  |

---

cc @youknowriad @tyxla @jsnajdr as it also relates to https://github.com/WordPress/gutenberg/pull/42525

youknowriad:
This is the same as #38911 

Unfortunately, we don't have a good way yet to isolate resolvers or apiFetch requests or any sync behavior when loading patterns (in order to show the loaders properly).

Any ideas @tyxla @jsnajdr  This could have a big impact on UX in a lot of places.

tyxla:
I haven't dug in-depth here just yet, but at first glance, this looks like a missed chance to preload some menus or pages' endpoint data. Or it's possible that the pre-existing navigation preloading doesn't work for some reason—IIRC, the endpoint parameters on the backend needed to match the ones on the front end perfectly for the preloading middleware to work. If that data is correctly preloaded, these blocks should load properly on the first render. @jsnajdr is this something you could investigate? 

Alternatively, we could experiment with implementing a loading spinner for each pattern that needs more time to load, although ideally, they shouldn't take that long. That would capture any requests that blocks inside the pattern preview perform after the iframe loads, though. But it might require some additional tooling: as @youknowriad mentioned we currently don't have a good way to know which requests are blocking a particular pattern preview. This could be another area to experiment in.

Furthermore, there could be some little improvements we could do here. What if the pattern previews fade in with a small delay, concealing at least some of the initial rendering glitches or FOUC? Something like this:


https://github.com/user-attachments/assets/9499781f-304c-4075-8ce0-052f72b029a1

youknowriad:
[...]

PR #63576, opened 2024-07-15
Inserter: Fade pattern previews in

tyxla:
## What?
Crude experimentation to fade in pattern previews.

**Please don't merge, this isn't production-ready.** 

## Why?
Pattern previews loading is often slow and clunky.

See https://github.com/WordPress/gutenberg/issues/44750

## How?
Adding an extra prop and some a simple CSS opacity transition.

## Testing Instructions
Open the pattern inserter and switch through some tabs.

None.

## Screenshots or screencast 

https://github.com/user-attachments/assets/8c487cb8-3527-49d4-85e6-5fec0aa04124

### DIFF
--- a/packages/block-editor/src/components/block-preview/auto.js
+++ b/packages/block-editor/src/components/block-preview/auto.js
@@ -108,6 +108,7 @@ function ScaledBlockPreview( {
 							? minHeight / scale
 							: minHeight,
 				} }
+				withLoadingEffects
 			>
 				<EditorStyles styles={ editorStyles } />
 				{ contentResizeListener }
--- a/packages/block-editor/src/components/block-preview/content.scss
+++ b/packages/block-editor/src/components/block-preview/content.scss
@@ -1,3 +1,9 @@
+.block-editor-block-preview__content-iframe {
+	opacity: 0;
+	transition: opacity 1s ease-in-out;
+	transition-delay: 0.5s;
+}
+
 // Hide inserter from previews.
 .block-editor-block-preview__content-iframe .block-list-appender {
 	display: none;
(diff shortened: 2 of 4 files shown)


stokesman:
Coincidently I was trying a related thing yesterday. What I found was that `Iframe` already supports `onLoad` and just changing `BlockPreview` to support passing it through enabled `BlockPattern` to pass a handler and go from there. I think it would work for this too and eliminate any changes to `Iframe`.

In my case, I wanted to avoid rendering the pattern preview until it entered the viewport to see if it helped the ones already entered load more quickly. To make that work the pattern previews have to start with a `min-height` or else they’re all in view at once. I used `onLoad` merely to remove the `min-height`.

tyxla:
> Coincidently I was trying a related thing yesterday. What I found was that `Iframe` already supports `onLoad` and just changing `BlockPreview` to support passing it through enabled `BlockPattern` to pass a handler and go from there. I think it would work for this too and eliminate any changes to `Iframe`.
> 
> In my case, I wanted to avoid rendering the pattern preview until it entered the viewport to see if it helped the ones already entered load more quickly. To make that work the pattern previews have to start with a `min-height` or else they’re all in view at once. I used `onLoad` merely to remove the `min-height`.

Yep, passing a handler sounds like a good way to isolate any changes. 

Note that I don't mean to land this, not in the way it is, but I wanted to put it in a PR so it leaves a record.

As discussed in the [related issue](https://github.com/WordPress/gutenberg/issues/44750), this doesn't solve the primary issue, and I'll be looking for a proper solution. In the meantime, we could add the fade in effect, because it can improve the appearance of the pattern previews.

PR #63578, opened 2024-07-15
Table of contents: Add border block support 

akasunil:
## What?

Add border block support to the `Table of contents` block.

## Why?
`Table of contents` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Table of contents`  block's border is configurable via Global Styles
- Edit template/page, Add `Table of contents`  block and Apply the border styles

## Screenshots or screencast 
 
 
[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (5).webm](https://github.com/user-attachments/assets/8ed3b2e9-a584-497e-939f-0a4f690fa6ae)


akasunil:
@aaronrobertshaw could use review here.

aaronrobertshaw (approved):
Thanks for the PR 👍

✅ Global styles display correctly
✅ Styling is consistent between editor and frontend
✅ Box-sizing added and works when parent enforces width

LGTM 🚀

Screenshot 2024-07-29 at 7 37 12 PM

packages/block-library/src/table-of-contents/style.scss
```diff
@@ -0,0 +1,4 @@
+:root :where(.wp-block-table-of-contents) {
```
aaronrobertshaw, line 1: I just realised belatedly that this CSS selector should be simplified. The `:root :where()` combo makes the specificity the same as the simple class name and probably should just be `.wp-block-table-of-contents`.

PR #63579, opened 2024-07-15
Tag Cloud: Add border block support 

akasunil:
## What?

Add border block support to the `Tag Cloud` block.

## Why?
`Tag Cloud` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Tag Cloud`  block's border is configurable via Global Styles
- Edit template/page, Add `Tag Cloud`  block and Apply the border styles
 

## Screenshots or screencast 
In backend:
 
image

In frontend:
image


aaronrobertshaw (commented):
Thanks for making a start on adding borders here @akasunil 👍 

For others coming to review this PR, there is some relevant discussion on a [related PR](https://github.com/WordPress/gutenberg/pull/63592) adding color supports to this block.

As this block is server-side rendered in the editor, it ends up with the block class applied to both the wrapper and inner markup. This causes issues when the block is styled via theme.json or Global Styles. Things like padding and border get applied multiple times.

Screenshot 2024-07-30 at 6 59 11 PM

Some fine-tuning is needed here before it can be moved forward.

akasunil:
Skipping border style for serverSIderRender wont be enough. As seen it below screenshot, global border style still getting applied

```
const serverSideAttributes = {
	...attributes,
	style: {
		...attributes?.style,
		border: undefined,
	},
};
```

image


We may have follow the solution we used for spacing [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/tag-cloud/editor.scss#L6)

aaronrobertshaw:
> Skipping border style for serverSIderRender wont be enough. 

This is to be expected. The `skipBlockSupportAttributes` omits the block support values that would be applied as inline styles. Global Styles target the block directly via classes and so are separate.

As you note a combination of the two approaches should smooth out the application of styles from both sources.

akasunil:
It works fine now.

[tag-cloud-block-border-support.webm](https://github.com/user-attachments/assets/cd57cd0f-4eac-4faf-b711-3e47537302db)

aaronrobertshaw (approved):
Thanks for iterating here 👍 

✅ Global Styles apply correctly
✅ Block instance styles override Global Styles
✅ Styling is consistent between editor and frontend
✅ Box sizing allows parent container to enforce accurate width

LGTM 🚢 

https://github.com/user-attachments/assets/f1671726-9071-4dbc-bcd8-239398bb7534

P.S. Maybe before merging we should update the inline comments explaining the resets in editor.scss. Along the same lines, we could also add a comment explaining the stripping of border values from the attributes passed to ServerSideRender. What do you think?

aaronrobertshaw:
Thanks for adding the inline comments 🚀 

LGTM!

PR #63582, opened 2024-07-15
Prepare JSON schemas for Draft 7 update

ajlende:
## What?

Fixes a small assortment of issues with the JSON schemas.

## Why?

Preparation for:
- #62462

## How?

Each fix is added as its own commit.

## Testing Instructions

1. Check that additional properties cannot be added to `ref`s in `styles`.
2. Check that all `styles` have access to the full assortment of options.
3. Schemas should be spec-compliant against the [Draft 4 specifications](https://json-schema.org/specification-links#draft-4). Notably "Any members other than "$ref" in a JSON Reference object SHALL be ignored." from the [JSON Reference specification](https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03).

N/A

## Screenshots or screencast 

N/A


t-hamano (approved):
LGTM! I think it's great that we were able to reduce the amount of code by introducing `json-schema-ref-parser`.

I discovered two fields that were missing `$ref` in the `stylesProperties:

https://github.com/WordPress/gutenberg/blob/c7c859c65997d2d29bd8966b18f278fd33119b59/schemas/json/theme.json#L1743-L1750

Should this be addressed in a follow-up?

scruffian (approved):
LGTM

ajlende:
> I discovered two fields that were missing $ref in the `stylesProperties`:
> Should this be addressed in a follow-up?

Yeah, It'll be easier to do after #62462 so I don't have to re-run the migration. 👍

ajlende:
I added the https://github.com/WordPress/gutenberg/labels/Backport%20to%20WP%20Minor%20Release label to get these fixes in the 6.6 schemas once they're merged with the https://github.com/WordPress/gutenberg/tree/wp/6.6 branch.

I think the fix that @t-hamano pointed out should also make it into those, so I'll open a separate PR to make that change in the drat 4 schema. We'll leave draft 7 for 6.7.

hellofromtonya:
This PR was discussed [here in Making WordPress `core-editor` slack channel](https://wordpress.slack.com/archives/C02QB2JS7/p1725461891687769). It is for helping with the internal release process and will not be merged to WordPress Core or included in 6.6.2. The label to backport it to `wp/6.6` is to merge it into that branch to help with the release process.

PR #63583, opened 2024-07-15
Update JSON Schemas to Draft 7

### LINKED ISSUE #62462: Update JSON schemas to use draft-07

ajlende:
## What?

Updates all JSON Schemas to use the [Draft 7 metaschema](https://json-schema.org/specification-links#draft-7).

## Why?

Fixes #62462

> We've been following the [SchemaStore best practices](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md?rgh-link-date=2024-06-11T05%3A13%3A33Z#best-practices) for the development of our JSON schemas. They recommend a schema version based on what is widely supported in editors and IDEs. Previously this was draft-04, so all of our schemas currently use that version. However the new recommendation is draft-07 which comes with some quality of life improvements that can improve the readability and maintainability of our JSON schemas.

Additionally, updating to using `propertyNames` uncovered some places where the `additionalProperties` hack wasn't updated properly. Those were fixed in #63582, and this change should ensure that it doesn't happen again in these places.

## How?

1. Used `ajv-cli` for initial automated migration to handle `exclusiveMinimum` and `const` usage.
2. Manually updated places to use `propertyNames` where appropriate.
3. Updated tests to use the new schema version.

## Testing Instructions

Ensure that validation is still working for all schemas.

N/A

## Screenshots or screencast 

N/A

### DIFF
--- a/schemas/json/theme.json
+++ b/schemas/json/theme.json
@@ -1,6 +1,6 @@
 {
 	"title": "JSON schema for WordPress block theme global settings and styles",
-	"$schema": "http://json-schema.org/draft-04/schema#",
+	"$schema": "http://json-schema.org/draft-07/schema#",
 	"definitions": {
 		"//": {
 			"explainer": "https://developer.wordpress.org/themes/advanced-topics/theme-json/",
@@ -469,8 +469,7 @@
 								"increment": {
 									"description": "The amount to increment each step by.",
 									"type": "number",
-									"exclusiveMinimum": true,
-									"minimum": 0,
+									"exclusiveMinimum": 0,
 									"default": 1.5
 								},
 								"steps": {
@@ -483,8 +482,7 @@
 								"mediumStep": {
 									"description": "The value to medium setting in the scale.",
 									"type": "number",
-									"exclusiveMinimum": true,
-									"minimum": 0,
+									"exclusiveMinimum": 0,
 									"default": 1.5
 								},
 								"unit": {
@@ -821,18 +819,18 @@
 				},
 				{
 					"properties": {
-						"appearanceTools": {},
-						"background": {},
-						"border": {},
-						"color": {},
-						"dimensions": {},
-						"layout": {},
-						"lightbox": {},
-						"position": {},
-						"shadow": {},
-						"spacing": {},
-						"typography": {},
-						"custom": {}
+						"appearanceTools": true,
+						"background": true,
+						"border": true,
+						"color": true,
+						"dimensions": true,
+						"layout": true,
+						"lightbox": true,
+						"position": true,
+						"shadow": true,
+						"spacing": true,
+						"typography": true,
+						"custom": true
 					},
 					"additionalProperties": false
 				}
@@ -1889,16 +1887,16 @@
 				},
 				{
 					"properties": {
-						"background": {},
-						"border": {},
-						"color": {},
-						"dimensions": {},
-						"spacing": {},
-						"typography": {},
-						"filter": {},
-						"shadow": {},
-						"outline": {},
-						"css": {}
+						"background": true,
+						"border": true,
+						"color": true,
+						"dimensions": true,
+						"spacing": true,
+						"typography": true,
+						"filter": true,
+						"shadow": true,
+						"outline": true,
+						"css": true
 					},
 					"additionalProperties": false
 				}
@@ -1916,16 +1914,16 @@
 						},
 						{
 							"properties": {
-								"border": {},
-								"background": {},
-								"color": {},
-								"dimensions": {},
-								"filter": {},
-								"shadow": {},
-								"outline": {},
-								"spacing": {},
-								"typography": {},
-								"css": {},
+								"border": true,
+								"background": true,
+								"color": true,
+								"dimensions": true,
+								"filter": true,
+								"shadow": true,
+								"outline": true,
+								"spacing": true,
+								"typography": true,
+								"css": true,
 								":hover": {
 									"$ref": "#/definitions/stylesPropertiesComplete"
 								},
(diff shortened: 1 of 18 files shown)


scruffian (approved):
As far as I understand it, this looks like its working as expected and the code looks good. I'd appreciate a review from someone with more knowledge of this area though.

ajlende:
I added a test for each of the 10 places that I manually refactored to use `propertyNames`. You can cherry-pick the last commit onto trunk and run the tests to confirm that they pass in both places.

@scruffian is that helpful?

test/integration/theme-schema.test.js
```diff
@@ -45,4 +49,13 @@ describe( 'theme.json schema', () => {
 
 		expect( result ).toBe( true );
 	} );
+
+	test.each( invalidFiles )( 'invalidates schema for `%s`', ( filepath ) => {
```
t-hamano, line 53: I would appreciate it if you could tell me why it is necessary to validate invalid `theme.json` files.

Isn't it enough to simply check whether `theme.json` files in the Gutenberg project are valid via the `validates schema for %s` test? 🤔

schemas/json/theme.json
```diff
@@ -2770,7 +2796,7 @@
 		"version": {
 			"description": "Version of theme.json to use.",
 			"type": "integer",
-			"enum": [ 3 ]
+			"const": 3
```
t-hamano, line 2799: Is it worth changing this to const as well?

https://github.com/WordPress/gutenberg/blob/update/json-schema-draft-07/schemas/json/theme.json#L2782

test/integration/theme-schema.test.js
ajlende: > I would appreciate it if you could tell me why it is necessary to validate invalid `theme.json` files.
> 
> Isn't it enough to simply check whether `theme.json` files in the Gutenberg project are valid via the `validates schema for %s` test? 🤔

Good question! Testing should have a good purpose, and I don't think it always does.

In software development this kind of testing is called [_negative testing_](https://en.wikipedia.org/wiki/Negative_testing). It's covering failure paths to make sure that they still fail in a way that you would expect.

The theme.json schema is used by many developers, not just core. We want to ensure that invalid properties are flagged for those theme developers as they are developing their themes.

Invalid properties in theme.json are underlined

In my personal philosophy of testing, I see tests as a communication tool much like short self-contained correct "compileable" examples ([SSCCEs](http://www.sscce.org/)). It can help people who are less familiar with a change understand specifically what has been changed.

The theme.json schema is a complicated schema with a lot of `anyOf`/`allOf` logic and `$ref`s. In this update, I manually changed the way that we invalidate some of those properties by using `propertyNames` instead of `additionalProperties`, and it isn't immediately obvious where each of those changes affects validation in theme.json. It actually took me a while to trace through all the references in the schema to generate those test files.

As a bonus, tests like this can reduce the burden of manual testing by reviewers when there are a lot of cases to test. Instead of going through each case, if the test can be written simply enough, it's often easier to check the validity of a single test than many manual cases.

I've written the test cases so that they should pass for the old schemas and the new schemas. If one of the tests passed with the old schemas and failed with the new schemas, we would know that I messed something up in my refactor since these properties should be invalid in both cases.

schemas/json/theme.json
ajlende: > Is it worth changing this to const as well?
> 
> [`update`/json-schema-draft-07/schemas/json/theme.json#L2782](https://github.com/WordPress/gutenberg/blob/update/json-schema-draft-07/schemas/json/theme.json?rgh-link-date=2024-07-20T10%3A41%3A03Z#L2782)

Yeah I'll update it. I wonder why the automate migration didn't update it?

ajlende: > Yeah I'll update it. I wonder why the automate migration didn't update it?

The automation didn't catch it because it was something I added after the automation. And I don't want to update it because the intention is for this `enum` to catch the remaining `propertyNames` that aren't part of the variable. In this case there's only one, but making it `const` would be inconsistent with every other instance of the `propertyNames` that we have.

t-hamano (approved):
LGTM! 

Thank you for the detailed explanation. I think the schemas are updated perfectly correctly in this PR 🚀

jeffpaul:
@DivyanshVinayak23 mind sharing your WordPress.org username so I can ensure its included in the 6.7 credits listing?

PR #63590, opened 2024-07-16
Use static 'key' when filtering BlockEdit components

Mamaduka:
## What?
PR updates the `BlockEdit` component instances returned by the `editor.BlockEdit` filter and adds a static "edit" key.

## Why?
While these filters don't experience this bug, the `key` prevents accidental remounts like #63557.

> Specifying a `key` tells React to use the `key` itself as part of the position, instead of their order within the parent.

## Testing Instructions
CI checks should be green

Same.


tyxla (approved):
Makes sense. Thanks @Mamaduka 👍 🙌

Mamaduka:
Thanks, @tyxla!

PR #63591, opened 2024-07-16
JSON Schema Reorganization and Fixes

ajlende:
## What?

Post-update code organization for JSON schema.

JSON structure changes:
- Inline `useRootPaddingAwareAlignments` property.
- Inline `fontFace` items.
- Sort `settingsProperties` alphabetically.
- Ensure every subschema with `properties` has `"type": "object"`.
- Ensure every subschema with `propertyNames` has `"type": "object"`.
- Alphabetize `stylesProperties`.

Fixes:
- Remove references to WordPress versions.
- Add missing ref support for `textAlign` and `textColumns`.
- Reuse `#/definitions/stylesElementsPseudoSelectorsProperties`

## Why?

Follow-up to #62462 to fix some things found while working on it.

## How?

See the commits for details.

## Testing Instructions

Check that [the new docs](https://github.com/WordPress/gutenberg/blob/update/json-schema-cleanup/docs/reference-guides/theme-json-reference/theme-json-living.md) and [new schema](https://github.com/WordPress/gutenberg/blob/update/json-schema-cleanup/schemas/json/theme.json) look correct.

Do all of the properties seem like they're in an intuitive order that makes sense?

Is the JSON file readable? Would you more easily be able to figure out where to add a new property if you need to?

N/A

## Screenshots or screencast 

N/A

### DIFF
--- a/docs/reference-guides/theme-json-reference/theme-json-living.md
+++ b/docs/reference-guides/theme-json-reference/theme-json-living.md
@@ -39,15 +39,6 @@ Setting that enables the following UI tools:
 
 ---
 
-### useRootPaddingAwareAlignments
-
-Enables root padding (the values from `styles.spacing.padding`) to be applied to the contents of full-width blocks instead of the root block.
-
-Please note that when using this setting, `styles.spacing.padding` should always be set as an object with `top`, `right`, `bottom`, `left` values declared separately.
-
-
----
-
 ### border
 
 Settings related to borders.
--- a/docs/reference-guides/theme-json-reference/theme-json-living.md
+++ b/docs/reference-guides/theme-json-reference/theme-json-living.md
@@ -273,8 +273,8 @@ Typography styles.
 | fontWeight | string, object |  |
 | letterSpacing | string, object |  |
 | lineHeight | string, object |  |
-| textAlign | string |  |
-| textColumns | string |  |
+| textAlign | string, object |  |
+| textColumns | string, object |  |
 | textDecoration | string, object |  |
 | writingMode | string, object |  |
 | textTransform | string, object |  |
--- a/docs/reference-guides/theme-json-reference/theme-json-living.md
+++ b/docs/reference-guides/theme-json-reference/theme-json-living.md
@@ -130,7 +130,6 @@ Settings related to position.
 
 | Property  | Type   | Default | Props  |
 | ---    | ---    | ---    |---   |
-| fixed | boolean | false |  |
 | sticky | boolean | false |  |
 
 ---
--- a/docs/reference-guides/theme-json-reference/theme-json-living.md
+++ b/docs/reference-guides/theme-json-reference/theme-json-living.md
@@ -41,8 +41,6 @@ Setting that enables the following UI tools:
 
 ### useRootPaddingAwareAlignments
 
-_**Note:** Since WordPress 6.1._
-
 Enables root padding (the values from `styles.spacing.padding`) to be applied to the contents of full-width blocks instead of the root block.
 
 Please note that when using this setting, `styles.spacing.padding` should always be set as an object with `top`, `right`, `bottom`, `left` values declared separately.
--- a/docs/reference-guides/theme-json-reference/theme-json-living.md
+++ b/docs/reference-guides/theme-json-reference/theme-json-living.md
@@ -130,6 +130,7 @@ Settings related to position.
 
 | Property  | Type   | Default | Props  |
 | ---    | ---    | ---    |---   |
+| fixed | boolean | false |  |
 | sticky | boolean | false |  |
 
 ---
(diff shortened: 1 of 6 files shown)


docs/reference-guides/theme-json-reference/theme-json-living.md
```diff
@@ -39,14 +39,14 @@ Setting that enables the following UI tools:
 
 ---
 
-### useRootPaddingAwareAlignments
```
ajlende, line 42: `useRootPaddingAwareAlignments` is removed in this PR, but returned in #63639. I wanted to keep the codegen changes in a separate PR, but the existing codegen didn't support the refactored structure which is why it is missing here.

scruffian (approved):
> Do all of the properties seem like they're in an intuitive order that makes sense?

Yeah, alphabetising them makes sense to me, and is as good as any other approach I can think of!

> Is the JSON file readable? Would you more easily be able to figure out where to add a new property if you need to?

Yeah, it looks clear to me.

schemas/json/theme.json
```diff
@@ -373,13 +324,18 @@
...
 			"properties": {
 				"position": {
 					"description": "Settings related to position.",
 					"type": "object",
 					"properties": {
+						"fixed": {
+							"description": "Allow users to set fixed position.",
+							"type": "boolean",
+							"default": false
+						},
```
t-hamano: It's true that `fixed` is supported, but [the dev note where this support was added](https://make.wordpress.org/core/2023/03/07/sticky-position-block-support/) only mentions `sticky`. Perhaps it's being hidden intentionally due to some problem.

Maybe @andrewserong knows something about this intention.

t-hamano (commented):
Thanks for the cleanup!

Here are some things I noticed, but they're very minor and may not need any action.

If the `description` field has highest priority in sorting, would it be better to move these fields as well?

https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/font-collection.json#L133

https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/theme.json#L2582-L2600

Do these need `type:string`?

https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/wp-env.json#L69
https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/theme.json#L814
https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/theme.json#L1683
https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/theme.json#L1733

Is this `type:object` unnecessary?
https://github.com/WordPress/gutenberg/blob/d22658bda87d774f31e6cf2d1b0e13226249f077/schemas/json/wp-env.json#L96

schemas/json/theme.json
```diff
@@ -1502,80 +1408,116 @@
...
 						}
 					},
 					"additionalProperties": false
 				},
+				"css": {
+					"description": "Sets custom CSS to apply styling not covered by other theme.json properties.",
+					"type": "string"
+				},
 				"dimensions": {
 					"description": "Dimensions styles",
```
t-hamano:
```suggestion
					"description": "Dimensions styles.",
```

schemas/json/theme.json
ajlende: Ah, thanks for checking the dev note. I just saw it in the global styles code next to `sticky`, so I assumed it was available. Checking the PR history, it doesn't like any blocks support `fixed` so that code, even though it exists, should never run. I'll remove it here, and assume that it'll be added when the block support is added.

t-hamano (approved):
LGTM! To ship this PR, we need to merge #63583 first, right?

ajlende:
> If the description field has highest priority in sorting, would it be better to move these fields as well?

Yeah, looks like I missed some. I did a search through the project for the regex `(?<!\{\n\s*)"description"` which should turn up any descriptions that are not first in their object, and fixed all of them now.

I also did a similar search for `"default": [^,]+,` which should be the same for the defaults at the end.

> Do these need `type:string`?

All `propertyNames` are evaluated as strings according to the [specification](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01#rfc.section.6.5.8), so I didn't include it if the `enum` was for `propertyNames`. But since it doesn't matter if it's included or not, maybe the consistency of always including a type with an `enum` would help people remember to add it in the other places too?

Do you have an opinion on that? I can see it either way.

> Is this `type:object` unnecessary?

No, but I should have moved it alongside `propertyNames`. 

One of my goals with the refactoring was to make sure that all instances of `"allOf"` included complete schemas with types inside and nothing outside. Technically it isn't invalid to have the `"type": "object"` outside, but I think it'll be more clear that each subschema gets validated independently of the others this way.

PR #63592, opened 2024-07-16
Tag Cloud: Add color block support

akasunil:
## What?

Add color block support to the `Tag Cloud` block.

Part of https://github.com/WordPress/gutenberg/issues/43245

## Why?
`Tag Cloud` block is missing color support.

## How?

Adds the color block support in block.json

## Testing Instructions
- Make sure that `Tag Cloud` block's colors setting are configurable via Global Styles
- Edit template/page, Add `Tag Cloud` block and Apply the border styles
- Verify that block color styles display correctly in both the editor and frontend

## Screenshots or screencast 
In backend:
image
 
In frontend:
image


packages/block-library/src/tag-cloud/block.json
```diff
@@ -37,6 +37,12 @@
 	"supports": {
 		"html": false,
 		"align": true,
+		"color": {
+			"text": true,
```
aaronrobertshaw: Is there much value in enabling both text and link color support for Tag Cloud? 

I suppose it could be possible for a theme or plugin to filter `wp_generate_tag_cloud` to alter the markup in a way that would include non-link text. That said, I'm not sure it is something we need to cater for though.

akasunil: I see, we can probably rule out `text` support then. Ill update the PR.

packages/block-library/src/tag-cloud/block.json
```diff
@@ -37,6 +37,11 @@
 	"supports": {
 		"html": false,
 		"align": true,
+		"color": {
+			"background": true,
```
aaronrobertshaw, line 41:
```suggestion
			"text": false,
			"background": true,
```

aaronrobertshaw (commented):
Thanks for iterating on this @akasunil 👍 

Unfortunately, simply omitting the `text` flag from the block.json supports config isn't enough as it is enabled by default when color support is included. It needs to be explicitly turned off.

I've left a suggestion below that has text color disabled for me.

akasunil:
A problem exists with the tag cloud block: styles are being applied twice as a result of the block class being used twice.

image

aaronrobertshaw:
As noted over in https://github.com/WordPress/gutenberg/issues/63716#issuecomment-2251609539, the adoption of color support for this block probably requires following a similar approach to the [Latest Comments](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/latest-comments/edit.js?plain=1#L31-L37) block.

akasunil:
My experience with the Latest Comments block was the same. I'm not sure if this is a bug or if this is how block with ServerSideRender suppose to work with block styles.


[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (10).webm](https://github.com/user-attachments/assets/3323c320-d70d-4398-9c25-1b2c8d29d03e)

Issue is basically this, It works perfectly fine with global styles. But when block styles are applied, it doesn't override the global styles, Until we save the template and reload the page.

Global styles are applied under css class `:root :where(.block-class)` and block styles are applied as inline. If we were to solve this issue, we have to set styles on ServerSideRendered div to inherit property from block main div.

akasunil:
from https://github.com/WordPress/gutenberg/issues/63716#issuecomment-2251609539
> I think this is back-to-front. Block instance styles **should** take precedence over global styles.

I agree, that was writing mistake.

aaronrobertshaw:
> Issue is basically this, It works perfectly fine with global styles. But when block styles are applied, it doesn't override the global styles, Until we save the template and reload the page.

I believe this is a limitation of the server-side rendering for the block. It isn't re-rendered when the block attributes change. This is another reason these server-side rendered blocks require a refactor to JS.

> Global styles are applied under css class :root :where(.block-class) and block styles are applied as inline. If we were to solve this issue, we have to set styles on ServerSideRendered div to inherit property from block main div.

From what you said about the Latest Comments block, the approach there works albeit requiring a reload. Making these server-side rendered blocks consistent is still a win.

I don't think it is an option to change block supports from inline styles. 

P.S. I'll be AFK for couple of weeks in the near future so further reviews might be rather slow.

akasunil:
It working fine now.

[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (2).webm](https://github.com/user-attachments/assets/e906b0d4-e6ff-493c-b6d4-38696d25d7a5)

aaronrobertshaw (commented):
I've taken this for another spin. 

For the most part it tests as advertised but there was one small issue that makes the colors feel a bit clunky when overriding global styles.

https://github.com/user-attachments/assets/117e6f36-7aa0-46cb-8b30-976c7a376fe2


The delay in the server side rendering updating means the inner rendering of the block appears slow to update the colors. I notice that behaviour is evident in your recent demo as well.

Similar to how you solved the application of borders to this Tag Cloud block, do you think we could reset the colors to inherit from the outer wrapper and omit the color values from the attributes passed to the server side rendering?

I suspect that would help smooth out the UX a little.

akasunil:
Another issue with the gradient as a background with tag cloud block, noted in the latest comments block as well.

Screenshot at Aug 09 18-40-21

akasunil:
It works as expected now. 

[Single-Posts-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (1).webm](https://github.com/user-attachments/assets/776bcbac-15ea-4e2f-9e1b-651f5100c05a)

We may need to perform the same for the most latest comments block and check all other blocks with serverSideRender.

packages/block-library/src/tag-cloud/block.json
aaronrobertshaw: Apologies @akasunil, I might have been wrong to question the text color support. It's just clicked for me that the Tag Cloud block does render simple text when there are no tags to display. 

It's conceivable that a user might wish to ensure the text color works nicely with the background color support added here. What do you think?

packages/block-library/src/tag-cloud/editor.scss
```diff
@@ -8,4 +8,6 @@
 	padding: 0;
 	border: none;
 	border-radius: inherit;
+	// Using important to override global styles.
+	background: inherit !important;
```
aaronrobertshaw: Global styles don't use `!important` in their CSS and its selectors only have `0-1-0` specificity which is overridden by this style's `0-2-0` specificity.

The block instance styles, that add `has-*-background-color` classes, do use `!important` but those are omitted from this inner server-side rendered element due to setting the style attributes color value to `undefined`.

Quickly removing this use of `!important` here works for me.

packages/block-library/src/tag-cloud/block.json
akasunil: @aaronrobertshaw I've opened a [PR](https://github.com/WordPress/gutenberg/pull/63774) to prevent empty tag cloud blocks from rendering on the frontend. So that placeholder text will be visible in editor only when tags are not available.

packages/block-library/src/tag-cloud/editor.scss
akasunil: Ah, you're right; I updated the style before removing the color attribute from sererSideRender, which is why I overlooked it to check.

aaronrobertshaw (commented):
@akasunil apologies for all the twists and turns on this one 🙏 

As this has evolved, it seems as we could make the code changes simpler and at the same time make it easier for the block to adopt further block supports in the future.

The creation of `serverSideAttributes` is effectively the same as what happens when `skipBlockSupportAttributes` is passed as a prop to the server side rendering. If we reinstate that approach and then use the CSS `all` property to reset the inner server-side rendered element's styles (preventing duplicate applications of global styles), then we don't have to juggle as many reset styles. That reset would then also work for any other block support style adopted in the future.

<details>
<summary>Here's a diff with what I'm suggesting</summary>

```diff
diff --git a/packages/block-library/src/tag-cloud/edit.js b/packages/block-library/src/tag-cloud/edit.js
index dfa00061c2..9a2b531b30 100644
--- a/packages/block-library/src/tag-cloud/edit.js
+++ b/packages/block-library/src/tag-cloud/edit.js
@@ -107,16 +107,6 @@ function TagCloudEdit( { attributes, setAttributes } ) {
 		setAttributes( updateObj );
 	};
 
-	// Remove border styles from the server-side attributes to prevent duplicate border.
-	const serverSideAttributes = {
-		...attributes,
-		style: {
-			...attributes?.style,
-			border: undefined,
-			color: undefined,
-		},
-	};
-
 	const inspectorControls = (
 		<InspectorControls>
 			<PanelBody title={ __( 'Settings' ) }>
@@ -196,8 +186,9 @@ function TagCloudEdit( { attributes, setAttributes } ) {
 			<div { ...useBlockProps() }>
 				<Disabled>
 					<ServerSideRender
+						skipBlockSupportAttributes
 						block="core/tag-cloud"
-						attributes={ serverSideAttributes }
+						attributes={ attributes }
 					/>
 				</Disabled>
 			</div>
diff --git a/packages/block-library/src/tag-cloud/editor.scss b/packages/block-library/src/tag-cloud/editor.scss
index 609320205f..be93fc42ab 100644
--- a/packages/block-library/src/tag-cloud/editor.scss
+++ b/packages/block-library/src/tag-cloud/editor.scss
@@ -1,12 +1,9 @@
-// The following styles are to prevent duplicate spacing and border for the tag cloud
-// block in the editor given it uses server side rendering. The specificity
-// must be higher than `0-1-0` to override global styles. Targeting the
-// inner use of the .wp-block-tag-cloud class should minimize impact on
-// other 3rd party styles targeting the block.
+// The following styles prevent duplicate applications of global styles due
+// to the server-side rendered block element also containing the block class.
+// Targeting the inner use of the .wp-block-tag-cloud class should minimize
+// impact on other 3rd party styles targeting the block.
 .wp-block-tag-cloud .wp-block-tag-cloud {
+	all: inherit;
 	margin: 0;
 	padding: 0;
-	border: none;
-	border-radius: inherit;
-	background: inherit;
 }

```
</details>


The main benefit here is that the typography styles will also update immediately when the user customizes them on a block instance so they override global styles. If we don't use the CSS `all` property, we could just add all the typography styles.

I don't have a strong opinion on whether using the `all` approach is best or if we should be more explicit and verbose. Perhaps a second opinion on that might help move the needle. cc/ @andrewserong 

All that aside, functionally, for colors this PR works and is getting close to where we can land it. Thanks for your patience.

aaronrobertshaw:
Thanks for working on this, @akasunil. I'm closing this in favour of #83264, which picks it up on the latest trunk.

The doubled styles came from the extra wrapper around the server rendered markup in the editor. #74228 has since removed that wrapper, so colour support now only needs the `block.json` change.

The new PR adopts background and link colour only. Text colour is left out because the tags are links, so a text colour control would appear to do nothing. Gradients are left out too.

PR #63596, opened 2024-07-16
Global styles block previews: fix scaling

MaggieCabrera:
## What?

Fixes the problem found in https://github.com/WordPress/gutenberg/pull/63558#issuecomment-2230251421

## Why?

Because it was a bug that hid the content of the preview when changing the font sizes

## How?

Using the same way that we use in the inserter to calculate the height of it instead of relying on using 100% because it needs to be recalculated after the changes

## Testing Instructions

1. Go on Global Styles / Blocks / Paragraph
2. Check the preview, it should be vertically centered
3. change the font sizes for it, from small to big and back to small, the text shouldn't disappear from the preview. It's ok if some of it is cut out by the overflow

## Screenshots or screencast 

Screen Capture on 2024-07-16 at 10-35-49


jasmussen (approved):
Nice! Fixes the issue!

PR #63597, opened 2024-07-16
Icons: Update close icon.

jasmussen:
## What?

The vectors of the close icon were not perfectly perpendicular. This PR updates them to be. Here's a before/after overlaid to show the visual difference, red is old, black is new:

before after

Here's the new icon shown in the grid:

new in grid

Here are the 1x pixels:

1x pixels

Visually it should be hard to see the differencer:

software


jameskoster (approved):
Good spot

PR #63600, opened 2024-07-16
DataViews: Bootstrap Quick Edit

youknowriad:
Related to #55101 and #59745 

## What?

This PR bootstraps a panel for quick edit pages in the pages dataviews of the site editor. There's a lot of iterations and questions that needs answering but my goal for this initial PR is to set the "architecture" in place:

 - How do we render the panel.
 - How do we select an item.
 - How do we save changes.
 - How do we organize the code base.

The following items will be left for other PRs: (I don't mind feedback on them but I know that they need work and would prefer to leave them separate)

 - The design of the panel itself. It's using the raw "DataForm" for now, I think we probably need two kind of "views" when rendering forms: regular forms and panels (like the inspector panels)
 - Bulk editing: it has its own challenges, I think we should quickly follow-up with a PR for this in order to see how the code base and the flows can be adapted to edit multiple entities rather than one.

## Testing Instructions

1- Open the "pages" section of the site editor.
2- Click "quick edit" in the actions menu of a given page.
3- Notice that you can edit the title.

### DIFF
--- a/packages/dataviews/src/layouts/table/index.tsx
+++ b/packages/dataviews/src/layouts/table/index.tsx
@@ -238,7 +238,7 @@ function TableRow< Item >( {
 					onChangeSelection(
 						selection.includes( id )
 							? selection.filter( ( itemId ) => id !== itemId )
-							: [ ...selection, id ]
+							: [ id ]
 					);
 				}
 			} }
--- a/packages/edit-site/src/components/layout/index.js
+++ b/packages/edit-site/src/components/layout/index.js
@@ -201,6 +201,17 @@ export default function Layout( { route } ) {
 							</div>
 						) }
 
+					{ ! isMobileViewport && areas.edit && (
+						<div
+							className="edit-site-layout__area"
+							style={ {
+								maxWidth: widths?.edit,
+							} }
+						>
+							{ areas.edit }
+						</div>
+					) }
+
 					{ ! isMobileViewport && areas.preview && (
 						<div className="edit-site-layout__canvas-container">
 							{ canvasResizer }
--- a/packages/edit-site/src/components/post-edit/index.js
+++ b/packages/edit-site/src/components/post-edit/index.js
@@ -1,4 +1,9 @@
 /**
+ * External dependencies
+ */
+import clsx from 'clsx';
+
+/**
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
@@ -14,7 +19,7 @@ import { useState, useMemo } from '@wordpress/element';
 import Page from '../page';
 import usePostFields from '../post-fields';
 
-export function PostEdit( { postType, postId } ) {
+function PostEditForm( { postType, postId } ) {
 	const { item } = useSelect(
 		( select ) => {
 			return {
@@ -45,21 +50,37 @@ export function PostEdit( { postType, postId } ) {
 		setEdits( {} );
 	};
 
+	if ( ! item ) {
+		return null;
+	}
+
 	return (
-		<Page className="edit-site-post-edit" label={ __( 'Post Edit' ) }>
-			{ item && (
-				<form onSubmit={ onSubmit }>
-					<DataForm
-						data={ itemWithEdits }
-						fields={ fields }
-						form={ form }
-						onChange={ setEdits }
-					/>
-					<Button variant="primary" type="submit">
-						{ __( 'Update' ) }
-					</Button>
-				</form>
+		<form onSubmit={ onSubmit }>
+			<DataForm
+				data={ itemWithEdits }
+				fields={ fields }
+				form={ form }
+				onChange={ setEdits }
+			/>
+			<Button variant="primary" type="submit">
+				{ __( 'Update' ) }
+			</Button>
+		</form>
+	);
+}
+
+export function PostEdit( { postType, postId } ) {
+	return (
+		<Page
+			className={ clsx( 'edit-site-post-edit', {
+				'is-empty': ! postId,
+			} ) }
+			label={ __( 'Post Edit' ) }
+		>
+			{ postId && (
+				<PostEditForm postType={ postType } postId={ postId } />
 			) }
+			{ ! postId && <p>{ __( 'Select a page to edit' ) }</p> }
 		</Page>
 	);
 }
--- a/packages/edit-site/src/components/post-edit/index.js
+++ b/packages/edit-site/src/components/post-edit/index.js
@@ -1,0 +1,65 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { DataForm } from '@wordpress/dataviews';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { store as coreDataStore } from '@wordpress/core-data';
+import { Button } from '@wordpress/components';
+import { useState, useMemo } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Page from '../page';
+import usePostFields from '../post-fields';
+
+export function PostEdit( { postType, postId } ) {
+	const { item } = useSelect(
+		( select ) => {
+			return {
+				item: select( coreDataStore ).getEntityRecord(
+					'postType',
+					postType,
+					postId
+				),
+			};
+		},
+		[ postType, postId ]
+	);
+	const { saveEntityRecord } = useDispatch( coreDataStore );
+	const { fields } = usePostFields();
+	const form = {
+		visibleFields: [ 'title' ],
+	};
+	const [ edits, setEdits ] = useState( {} );
+	const itemWithEdits = useMemo( () => {
+		return {
+			...item,
+			...edits,
+		};
+	}, [ item, edits ] );
+	const onSubmit = ( event ) => {
+		event.preventDefault();
+		saveEntityRecord( 'postType', postType, itemWithEdits );
+		setEdits( {} );
+	};
+
+	return (
+		<Page className="edit-site-post-edit" label={ __( 'Post Edit' ) }>
+			{ item && (
+				<form onSubmit={ onSubmit }>
+					<DataForm
+						data={ itemWithEdits }
+						fields={ fields }
+						form={ form }
+						onChange={ setEdits }
+					/>
+					<Button variant="primary" type="submit">
+						{ __( 'Update' ) }
+					</Button>
+				</form>
+			) }
+		</Page>
+	);
+}
--- a/packages/edit-site/src/components/post-fields/index.js
+++ b/packages/edit-site/src/components/post-fields/index.js
@@ -1,0 +1,345 @@
+/**
+ * External dependencies
+ */
+import clsx from 'clsx';
+
+/**
+ * WordPress dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+import { decodeEntities } from '@wordpress/html-entities';
+import {
+	createInterpolateElement,
+	useMemo,
+	useState,
+} from '@wordpress/element';
+import { dateI18n, getDate, getSettings } from '@wordpress/date';
+import {
+	trash,
+	drafts,
+	published,
+	scheduled,
+	pending,
+	notAllowed,
+	commentAuthorAvatar as authorIcon,
+} from '@wordpress/icons';
+import { __experimentalHStack as HStack, Icon } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+import { useEntityRecords, store as coreStore } from '@wordpress/core-data';
+
+/**
+ * Internal dependencies
+ */
+import {
+	LAYOUT_GRID,
+	LAYOUT_TABLE,
+	LAYOUT_LIST,
+	OPERATOR_IS_ANY,
+} from '../../utils/constants';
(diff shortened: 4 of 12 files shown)


packages/edit-site/src/components/layout/index.js
```diff
@@ -201,6 +201,17 @@ export default function Layout( { route } ) {
 							</div>
 						) }
 
+					{ ! isMobileViewport && areas.edit && (
```
youknowriad, line 204: I added a new routable area to address this use-case. Routes in the site editor can decide to inject UI in the "edit" area.

packages/edit-site/src/components/post-edit/index.js
```diff
@@ -0,0 +1,65 @@
...
+import { Button } from '@wordpress/components';
+import { useState, useMemo } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Page from '../page';
+import usePostFields from '../post-fields';
+
+export function PostEdit( { postType, postId } ) {
```
youknowriad: We have `PostList` to render dataviews for a given post type.
This new component `PostEdit` is meant to edit the "form" or "inspector panel" of a given post type.

packages/edit-site/src/components/post-fields/index.js
```diff
@@ -0,0 +1,345 @@
...
+				<div className="page-templates-author-field__icon">
+					<Icon icon={ authorIcon } />
+				</div>
+			) }
+			<span className="page-templates-author-field__name">{ text }</span>
+		</HStack>
+	);
+}
+
+function usePostFields( viewType ) {
```
youknowriad, line 153: I extracted the "fields" from the "PostList" component to be able to reuse the same fields between "PostList" and "PostEdit". This is a strong principle that we should stick to.

`viewType` is not necessary IMO, I think ideally we should rework our fields to not depend on the view type. I left it for now though.

packages/edit-site/src/components/post-fields/index.js
```diff
@@ -0,0 +1,345 @@
...
+				render: ( { item } ) => (
+					<FeaturedImage item={ item } viewType={ viewType } />
+				),
+				enableSorting: false,
+			},
+			{
+				header: __( 'Title' ),
+				id: 'title',
+				type: 'text',
+				getValue: ( { item } ) =>
```
youknowriad, line 181: I updated this function to return the "raw" title instead of the "rendered" title. I think it's a good question to ask: what does this property correspond too. For me, it should correspond to the "raw" title". For the "rendered" values we have the `render` function to override this.

packages/edit-site/src/components/post-fields/index.js
ntsekouras: Even though at this stage we share the fields between posts and pages, it would be good to have a mandatory argument of the `postType` passed here, no? It should be used with the future APIs with field registration etc..

youknowriad: I don't know, as long as we don't need, I prefer to avoid it but I also think it's going to be useful very quickly.

ntsekouras: I think it should be similar to usePostActions and pass an object to be flexible with additions. We can handle later of course as it won't change anything now..

ntsekouras (commented):
That's a great start, thank you!

I had in mind that we would probably render the form in a dialog and could even make it draggable, but your approach of a new area looks good.

Having it as an action (Quick edit) makes the most sense to me. It might need some design input whether to make it a primary action and give a bit more prominence.

Some quick wins for this PR to feel a bit more polished would be:
1. add a way to close the quick edit panel
2. probably have a toast on save and have the button enabled only if we have changes
3. Add a title in the panel to be clear which entity you are about to edit

>How do we save changes.

I think having a centralized `submit` makes sense and it should be enough for this first iteration. I'm pretty sure we'll have more challenges when implementing other field types 😄 

>How do we organize the code base.

What do you mean by that?


Finally, some tweaks with the state will be needed to handle:
1. Quick edit one item (panel opens)
2. Quick edit different item

jameskoster:
A couple of fundamental pieces from the design side:

* List layout needs some more design exploration. Ideally the preview frame should remain visible, even when quick-editing.
  * One potential solution is for the content frame to drill down and reveal the Inspector contents.
* The record that appears in the panel should be correlated with the selection state. This is how we'll facilitate bulk-quick-edits down the road.
* As Nick said, the panel must be toggle-able. For this initial PR, given List layout still needs some design work, should we restrict quick editing to Grid / Table layout? In those layouts we might use the same UI as the full screen editor, IE a toggle button with the `drawer` icon.
* It's not 100% clear how saving should work. For now, could the changes be captured by the global save button/UX? 

I'll update https://github.com/WordPress/gutenberg/issues/55101 with a more comprehensive prototype when I find time.

youknowriad:
To be honest, I'm very skeptical about "preview and edit" at the same time for list views but we can focus on table and grid for now if needed. Happy to update the PR. Please do share a prototype, that will help a lot.

jameskoster:
To be honest I think I'd rather lose the quick-edit functionality in List layout than the visible preview. The latter seems fundamental to that layout – without it there are too many overlaps with Table layout. 

Restricting quick edit to Grid / Table seems a good short term step.

youknowriad:
> To be honest I think I'd rather lose the quick-edit functionality in List layout than the visible preview. The latter seems fundamental to that layout – without it there are too many overlaps with Table layout.

But we're not losing "preview", quick edit can be seen as a module, you do your edits and you return to the "preview". But I won't argue much here, I'm fine trying any approach.

jameskoster:
I added some mockups to https://github.com/WordPress/gutenberg/issues/55101.

youknowriad:
I didn't disable it yet in list views, but we're getting closer to the prototypes. Can you check it?

packages/dataviews/src/dataviews.tsx
```diff
@@ -75,6 +76,7 @@ export default function DataViews< Item >( {
 	selection: selectionProperty,
 	setSelection: setSelectionProperty,
 	onSelectionChange = defaultOnSelectionChange,
+	header,
```
youknowriad: This is a temporary prop until we address this issue https://github.com/WordPress/gutenberg/issues/63646

youknowriad:
The thing that feels weird to me is that clicking items in the table view actually bulk selects by default. I think it should unselect all the other items and selects the current one automatically. It should be different than clicking the checkbox. Maybe we could support ctrl + click for multi-select as well.

ntsekouras:
>The thing that feels weird to me is that clicking items in the table view actually bulk selects by default.

I agree with this and with this PR especially it contributes to a quite confusing experience. I quickly tested and the status handling is still problematic when we change entities (table view for example).

Another thought is if we don't have a quick edit post action, we still need to have checks about permissions etc to edit. A basic example would be the `status:trash`.

Based on this, at some cases the entity will not be editable. Should we have the button there disabled? Should we not render it and have the layout shift?

youknowriad:
> Based on this, at some cases the entity will not be editable. Should we have the button there disabled? Should we not render it and have the layout shift?

I think we need some placeholder content for:

 - Nothing selected.
 - An item is selected but is not editable.

jameskoster:
> I think it should unselect all the other items and selects the current one automatically.

That works for me. Ctrl + click (and shift + click) would be nice enhancements. 

Small tweak: The toggle button should be `compact` size.

> Nothing selected.
> An item is selected but is not editable.

For now some centrally aligned text is probably adequate, e.g.:

Screenshot 2024-07-18 at 15 45 22

We can elaborate on this in follow-ups.

jameskoster:
We should probably show something when you bulk select too, but I'm not sure what the message should be. "Only one item can be edited at a time" is a bit underwhelming, but perhaps a necessary interim?

youknowriad:
For me this should be ready to ship. I know that it's very bare bone (only the title is editable and no bulk selection), but I think it's better to iterate on separate PRs at this point rather than continue in a huge PR.

WDYT?

packages/dataviews/src/layouts/table/index.tsx
```diff
@@ -238,7 +238,7 @@ function TableRow< Item >( {
 					onChangeSelection(
 						selection.includes( id )
 							? selection.filter( ( itemId ) => id !== itemId )
-							: [ ...selection, id ]
```
oandregal, line 241: Why do we need to clear the existing selection when the user clicks the row? This means that we'd have two affordances (checkbox & row click) that work differently:

https://github.com/user-attachments/assets/7a8e0f3a-bf49-4608-8445-fd6cc7c2467d

youknowriad: yes. See https://github.com/WordPress/gutenberg/pull/63600#issuecomment-2232870658 and the following comments.

oandregal: I see this is intentional https://github.com/WordPress/gutenberg/pull/63600#issuecomment-2232870658

jameskoster:
@youknowriad a tiny detail; we're using the wrong drawer icon, it should be the flipped version.

One other tiny detail: The gap between the drawer icon button and the view options button is 8px, but the gap between the layout + view options buttons are 4px. Ideally both gaps are 4px.

I'm also not seeing any text when the sidebar is empty, or multiple items are selected.

> I know that it's very bare bone (only the title is editable and no bulk selection), but I think it's better to iterate on separate PRs at this point rather than continue in a huge PR.

Should we put it behind a feature flag for now? While everything works in this very limited scope, it feels very half-baked currently.

youknowriad:
> I'm also not seeing any text when the sidebar is empty, or multiple items are selected.

Did you refresh the branch?

> Should we put it behind a feature flag for now? While everything works in this very limited scope, it feels very half-baked currently.

Works for me, I agree with that.

packages/dataviews/src/layouts/table/index.tsx
oandregal: I see where this is coming from: we want to optimize for the quick edit flow, otherwise it's quite cumbersome 👍

packages/edit-site/src/components/post-list/index.js
```diff
@@ -608,13 +291,35 @@ export default function PostList( { postType } ) {
...
+							icon={ drawerRight }
+							label={
+								! quickEdit
+									? __( 'Show quick edit sidebar' )
+									: __( 'Close quick edit sidebar' )
+							}
+							onClick={ () => {
+								history.push( {
+									...location.params,
+									quickEdit: quickEdit ? undefined : true,
```
oandregal, line 317: I have checked and the quickEdit URL parameter seems to be cleared properly when switching screen (back to Design or into the Editor).

packages/edit-site/src/components/post-list/index.js
```diff
@@ -608,13 +291,35 @@ export default function PostList( { postType } ) {
 				fields={ fields }
 				actions={ actions }
 				data={ records || EMPTY_ARRAY }
-				isLoading={ isLoadingMainEntities || isLoadingAuthors }
+				isLoading={ isLoadingMainEntities || isLoading }
```
oandregal: Nit: `isLoadingData || isLoadingFields`.

jameskoster:
> Did you refresh the branch?

Apparently not, it's working now :) 

I noticed a small regression though; in table layout the selected row no longer has blue accents.

oandregal (approved):

jameskoster:
> I noticed a small regression though; in table layout the selected row no longer has blue accents.

Never mind, this seems to be the case on trunk too.

jasmussen:
A thought: we should probably make an effort to not refer to this as "Quick Edit" (which sounds like a _named feature_). In practice, it's just access to editing, or even bulk editing, and this action shouldn't officially have a name.

ntsekouras:
Hm.. But it is an action.. Do you refer to the capitalization? If yes, this is probably because that's how we render it in wp-admin already.

jasmussen:
I think it's fine for us developing the software to call it this. To clarify, I'm saying this feature shouldn't officially have a name, or any user-facing name. E.g. when writing documentation, we should not refer to this as "Use the quick edit feature", it should be more in the vein of "If you're editing pages in the table view, you can show the details panel to edit additional properties". 

I see it's got "quick edit" in the tooltip, I wonder if we can change that to just "Show details"?

ntsekouras:
>I see it's got "quick edit" in the tooltip, I wonder if we can change that to just "Show details"?

We could do that yes. But do you think it's going to be clear that there we can actually edit?

Another thing is that we are going to [introduce](https://github.com/WordPress/gutenberg/pull/64437) this for `list` layouts and is going to be part of the `actions` menu. Should we rename it there too? Should we have it as a togglable button too?

jasmussen:
The tooltip can be anything. Mainly I'd prefer "edit" used as a word rather than "Quick Edit" surfaced as a _feature name_. Describe the flow in a sentence rather than refer to a feature.

jameskoster:
Maybe something like 'View details' or 'Edit details' could work as the action name in List layout?

In other layouts the tooltip could be 'Toggle details panel'?

jasmussen:
Sounds good to me! (Either.)

ntsekouras:
Opened a quick [PR](https://github.com/WordPress/gutenberg/pull/64475) to update the copy.

PR #63602, opened 2024-07-16
React: Restore umd builds

youknowriad:
In 6.6 we tried moving away from the deprecated React UMD builds, the issue we faced is that there's a warning that is triggered on the console because we're not using a separate impact for `createRoot`. (SCRIPT_DEBUG true and `npm run build`)

`Warning: You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client".`

This warning has been removed in React 19 along with the removal of the UMD builds, so we should be able to revert this PR when we upgrade to React 19.

Related Core PR here https://github.com/WordPress/wordpress-develop/pull/7021

### DIFF
--- a/tools/webpack/packages.js
+++ b/tools/webpack/packages.js
@@ -102,6 +102,13 @@ const exportDefaultPackages = [
 	'warning',
 ];
 
+const copiedVendors = {
+	'react.js': 'react/umd/react.development.js',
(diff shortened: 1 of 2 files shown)


Mamaduka (approved):
This test well for me. I no longer see the warning.

tyxla (approved):
FWIW I was able to reproduce this one today and I confirm this fixes it. 👍

PR #63603, opened 2024-07-16
Increase column-gap between fields in List layout

jameskoster:
Now that fields can include icons the spacing is a bit too small. If you squint it's not clear which label the icon is associated with here:

Screenshot 2024-07-16 at 10 50 01

This PR bumps the spacing up from 8px to 12px:

Screenshot 2024-07-16 at 10 50 17

### Testing instructions

- Navigate to a data view like Pages
- Switch to List layout
- Toggle on multiple fields
- Notice the increased horizontal spacing between each field


jasmussen (approved):
Yes base12.

PR #63609, opened 2024-07-16
Columns block: fix block preview

MaggieCabrera:
## What?

This fixes the block preview so the columns are visible and not collapsed

## Why?

Because this is a better preview of what the block does

## How?

By using the right viewport size according to the block's css

## Testing Instructions

Check the preview of the block in the inserter and in global styles

## Screenshots or screencast 

Screenshot 2024-07-16 at 13 38 00

### DIFF
--- a/packages/block-library/src/columns/index.js
+++ b/packages/block-library/src/columns/index.js
@@ -23,7 +23,7 @@ export const settings = {
 	icon,
 	variations,
 	example: {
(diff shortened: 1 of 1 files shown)


Mamaduka (approved):
Works as expected ✅

## Screenshot
CleanShot 2024-07-16 at 15 58 43

jasmussen:
Nice one.

PR #63610, opened 2024-07-16
Use `@wordpress/warning` during block registration instead of `console.error` and `console.warn`

SantosGuillamot:
## What?
As suggested [here](https://github.com/WordPress/gutenberg/pull/63117#discussion_r1679159719) by @gziolo , in this pull request I explore the possibility of using `@wordpress/warning` during block registration instead of the current `console.error` and `console.warn` calls.

This means that all the previous errors are now warnings, which I am not sure if it's fine:

**Before**
Screenshot 2024-07-16 at 11 59 55

**After**
Screenshot 2024-07-16 at 11 59 10

## Why?
To unify how these use cases are handled and to benefit from the functionalities the package provides by default.

## How?
Just change the console calls with the `warning` function and adapt the tests.

## Testing Instructions
Automated tests should pass.


gziolo:
There are 4 failing unit tests reported that needs to be still fixed and this will be good to go. I really like this refactoring 💯

packages/blocks/src/api/test/registration.js
```diff
@@ -72,39 +72,39 @@ describe( 'blocks', () => {
...
 		} );
 
 		it( 'should reject blocks with too many namespaces', () => {
 			const block = registerBlockType( 'doing/it/wrong' );
-			expect( console ).toHaveErroredWith(
+			expect( console ).toHaveWarnedWith(
 				'Block names must contain a namespace prefix, include only lowercase alphanumeric characters or dashes, and start with a letter. Example: my-plugin/my-custom-block'
 			);
 			expect( block ).toBeUndefined();
 		} );
```
SantosGuillamot, line 97: It seems that the `@wordpress/warning` package only triggers one message, which is causing these tests to fail: [link](https://github.com/WordPress/gutenberg/blob/9e73c1158bb8eb79d6415743a7519fad2d2ffef1/packages/warning/src/test/index.js#L30). I assume that's expected, right? Any idea how to solve that?

**Before**
Screenshot 2024-07-16 at 13 36 18

**After**
Screenshot 2024-07-16 at 13 54 26

gziolo: OK, so the same message is printed only once. You would have to reset it through:

https://github.com/WordPress/gutenberg/blob/release/14.2/packages/warning/src/utils.js

There are multiple ways in Jest to do it, you can import `logged` and call a method on this set, for example.

gziolo:
```js
import logged from 'path/to/warning/src/utils.js';

logged.clear();
```

gziolo: https://github.com/WordPress/gutenberg/blob/4171f270f2e84f29e722904d50dc00d905facbce/packages/warning/src/test/index.js#L10-L13

SantosGuillamot: Okay, I made the changes in [this commit](https://github.com/WordPress/gutenberg/pull/63610/commits/a1b65cba05cc7ede69b774e4e7d0c0375f51c3cd) and tests are passing (at least locally).

However, importing from `'../../../../warning/src/utils'` feels pretty weird. What happens to other projects wanting to use this package and add tests like ours but don't have the package in the same repo? Should we export it somehow?

gziolo: I wouldn’t worry about it. You could also force Jest to reload the package with some of the utils they provide:

https://jestjs.io/docs/jest-object#jestresetmodules

gziolo (approved):
That looks great. Thank you for a swift follow-up.

PR #63611, opened 2024-07-16
Fix navigate regions shortcuts on the back button WP logo slot

### LINKED ISSUE #63552: Navigable regions: keyboard shortcut to navigate regions doesn't work from the WP logo any longer
### Description

This appears to be a regression on trunk. 

First, test on WordPress 6.5:
- Edit a post.
- Move focus to the WP logo in the top bar.
- Press Control + backtick or alternatively  Control + Option + N
- Observe that each time you press the keyboard shortcut, focus is moved to one of the main sections of the editor, as expected.
- Test on latest trunk.
- Observe pressing the keyboard shortcut from the WP logo doesn't do anything.
- Press the Tab key to move focus to the plus icon button of the main inserter.
- Press the keyboard shortcut and observe it works as expected.
- In the Site editor it seems to work as expected.

Basically it stopped working only in the Post editor and only from the WP logo

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63552

## What?

The navigate regions keyboard shortcut doesn't work in the Post Editor when focus is on the WP Logo.
The WP logo now uses a Slot with the `bubblesVirtually` prop. That means events 'bubble' up the _React hierarchy tree_ which may be different from the DOM tree. This may be triggered by any other slot implementation added in the future so it would be best to find a solit way to make sure navigate regions works in any case.

This works in the Site editor but it doesn't in the Post editor. It took me a while to understand how the different React components hierarchy impacts this feature.

**In the Site editor:**
It works because `useNavigateRegions` is used on [the edit-site Layout component](https://github.com/WordPress/gutenberg/blob/154f1aaeea73ba1ed2c17373814c8c3620dc9626/packages/edit-site/src/components/layout/index.js). This component contains basically everything so that the Slots are inside the Layout in the React hierarchy.

**In the Post editor**
It does not work because `useNavigateRegions` is used on the `InterfaceSkeleton` component. However, the WP logo `BackButton` is actually inside the Layout (edit-post package) > Editor (editor package) component, which thenuses EditorInterface which in turn uses InterfaceSkeleton.

Similar hierarchy in the Widgets editor.

Basically, to make sure the navigate regions shortcut works in all editors regardless of where any Slot is used, there is the need to make sure `useNavigateRegions` is used on the outermost component of the editors.

This PR tries this approach by moving the usage of `useNavigateRegions` up high in the components hierarchy to a couple ew div elements. If it turns to work well, maybe it would be worth considering to abstract to a 'navigate region provider' component that can be reused where desired.

Cc @youknowriad  to my understanding there are a few experimental 'editor providers' in place that I guess they are meant to attach behaviors to the editor and be consolidated in the future. Would a new provider for the navigate region props and ref be a good option?

## Why?

To make sure the navigate regions keyboard shortcut works from everywhere.

## How?

- Moves the usage of `useNavigateRegions` higher up in the components hierarchy.
- Removes some redundant props passed down to components, as `useNavigateRegions` already has default shortcuts.
- Removes a no longer used `enableRegionNavigation` prop.

## Testing Instructions

- Edit post: move focus to the WP logo and test the navigate regions keyboard shortcut works as expected
- Edit site: repeat the test with various views of the Site editor: preview, edit, pages, styles, navigation, etc.
- Optional: Enable the `PostsApp` experiment and test.
- Edit widgets: repeat the test. (you need to activate a theme like Twenty-Twenty One to access the Widgets editor).

## Testing Instructions with metaboxes.
- Activate a plugin that adds a meta box e.g. Yoast SEO.
- Test on trunk first.
- Edit a post and set focus on one of the controls in the Yoast SEO metabox, whether it's an input field or a button.
- Try the keyboard shortcut `Ctrl+backtick` or alternatively (macOS) Control+Option+N.
- Observe that nothing happens.
- Switch to this branch and build.
- Repeat the test.
- Observe the keyboard shortcut moves focus through the editor regions, as epected.

### DIFF
--- a/packages/edit-post/src/components/layout/index.js
+++ b/packages/edit-post/src/components/layout/index.js
@@ -47,6 +47,7 @@ import {
 	SlotFillProvider,
 	Tooltip,
 	VisuallyHidden,
+	__unstableUseNavigateRegions as useNavigateRegions,
 } from '@wordpress/components';
 import {
 	useMediaQuery,
@@ -482,6 +483,8 @@ function Layout( {
 		document.body.classList.remove( 'show-icon-labels' );
 	}
 
+	const navigateRegionsProps = useNavigateRegions();
+
 	const className = clsx( 'edit-post-layout', 'is-mode-' + mode, {
 		'has-metaboxes': hasActiveMetaboxes,
 	} );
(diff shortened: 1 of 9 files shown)


Mamaduka:
@afercia, do you mind rebasing the branch and resolving merge conflicts?

afercia:
@Mamaduka rebased 👍🏻

afercia:
@WordPress/gutenberg-core I'd appreciate a review. This fixes a regression that is important enough to be fixed in the next WordPress releae. thanks.

afercia:
In my testing, moving the keyboard shortcut event up in the tree has an added bonus. It makes the keyboard shortcut work also from the metaboxes area. Apparently, it didn't ever work from the metaboxes area. It works now which is even more important after https://github.com/WordPress/gutenberg/pull/64351

I will update the test instructions.

stokesman:
Just resolved a tiny conflict that merging #65325 had introduced.

Update: I’d messed that up. Fixed in 310c4be592f34e65da02f27e5f74c2f53fa9bdd3 😅

stokesman (commented):
This tests well for me in all the affected editors. Code-wise, I like it. The key thing it seems to me is that `InterfaceSkeleton` has a breaking change. Before this it applied region navigation by default and now it doesn’t (nor has the capability to do so). I think this is acceptable given that any plugins consuming `InterfaceSkeleton` aren’t going to experience breakage unless they upgrade their `@wordpress/interface` dependency. This is because the package isn’t exposed on the `wp` global; i.e. plugins using components from the package are bundling them into their build so they control the version.

A "breaking" changelog entry and dev note seem warranted. Other than that, this looks 🚢 shape to me.

afercia:
@stokesman thanks for looking into this. Please do feel free to go ahead with the changelog entry. As a native English speaker I guess you're a better fit to do that 🙏🏻

getdave:
@afercia @stokesman I noticed this was targetted for 6.7 but hadn't been merged. I've taken the liberty of rebasing it to resolve conflicts.

In my testing based on the PR instructions this still works as expected but I'd appreciate you verifying before we look to merge 🙇

kevin940726 (approved):
I tested it and it seemed to still work fine 💯!

My only concern is whether we want to introduce a breaking change (`enableRegionNavigation`) at this stage? Is there a way to gracefully handle backward compatibility? Never mind if this is a non-issue!

afercia:
I can't remember the details about `enableRegionNavigation` maybe @youknowriad may know more about it.
Of course Ihave no objections to handle backward compatibility if that's really needed. I'm not sure I can understand whether that's needed though. 

Anyways, I would like to remind everyone this is a regression and it should be fixed for the release.

stokesman:
> My only concern is whether we want to introduce a breaking change (`enableRegionNavigation`) at this stage?

That prop is removed from two places:
- `EditorInterface`, a private component so no BC concerns
- `InterfaceSkeleton`, a public component but one that’s bundled by consumers therefore they control when to upgrade

It seems possible to not break `InterfaceSkeleton` but doing so makes for simpler code and it should be acceptable given the point above.

kevin940726:
I wasn't aware that the `interfaces` package is bundled! In that case, I think it's okay to land this as is 👍.

stokesman:
Given the breaking change a dev note seems like a proper thing so I added the label. Though, still the WP release won’t force the upgrade/breakage so I’m not totally sure it’s required.

afercia:
Thanks everyone for the in depth feedback. Is it OK to everyone I push the green button? Cc @youknowriad

youknowriad:
Go for it. Thanks all :)

afercia:
Looks like the conflict when cherry picking this to `wp/6.7` is because of the changelog entry under `Unreleased` in `packages/interface/CHANGELOG.md`. I thought this was handled automatically? Does anyone know how to fix it?
@youknowriad @stokesman 

Shall I just follow the instructions and move the entry under the `## 6.9.0 (2024-10-03)` package release?

stokesman:
> and move the entry under the `## 6.9.0 (2024-10-03)` package release?

I would think it‘s better to put under the `Unreleased` section, it may not be accurate (with regards to the wp/6.7 branch) but neither is putting it under `6.9.0`. Alternatively, maybe it’s fine to omit the changelog changes for that PR, i.e. fix the conflict by adding nothing to the file. The changes aren’t really a public part of the WP 6.7 release.

aaronrobertshaw:
> I would think it‘s better to put under the Unreleased section

I'm not sure if there is a documented official process but this is what https://github.com/WordPress/gutenberg/pull/65905 did. Here's the [original PR](https://github.com/WordPress/gutenberg/pull/65772) that one was a manual backport to the 6.7 branch for.

afercia:
Thanks all. I created https://github.com/WordPress/gutenberg/pull/65971 for the backport.

PR #38785, opened 2022-02-14
Add __experimentalUseResourcePermissions

adamziel:
## Description
This PR is a minimal subset of https://github.com/WordPress/gutenberg/pull/38135 focused solely on the `useResourcePermissions` hook.

The goal of these new APIs is to lower the barrier of entry for new contributors and make the life of existing contributors easier.

### Example usage:
**Before**
```js
useSelect(
  ( select ) => {
	  return {
		canUserUpdateNavigationEntity: ref
			? canUser( 'update', 'navigation', ref )
			: undefined,
		hasResolvedCanUserUpdateNavigationEntity: hasFinishedResolution(
			'canUser',
			[ 'update', 'navigation', ref ]
		),
		canUserDeleteNavigationEntity: ref
			? canUser( 'delete', 'navigation', ref )
			: undefined,
		hasResolvedCanUserDeleteNavigationEntity: hasFinishedResolution(
			'canUser',
			[ 'delete', 'navigation', ref ]
		),
		canUserCreateNavigation: canUser( 'create', 'navigation' ),
		hasResolvedCanUserCreateNavigation: hasFinishedResolution(
			'canUser',
			[ 'create', 'navigation' ]
		),
	  };
  },
  [ ref ]
);
```

**After:**
```js
const [ hasResolved, { canCreate, canUpdate, canDelete } ] = __experimentalUseResourcePermissions( 'navigation', ref );
```

It returns a tuple and with `hasResolved` the first item so that it's intentionally too hard to overlook. Thank you @kevin940726 for this suggestion!

See also [how the combination of all proposed hooks makes the navigation block ~200 lines leaner](https://github.com/WordPress/gutenberg/pull/38289).

### Test plan

1. Confirm if the tests are green
2. Confirm if the code makes sense

### Other considerations

@gziolo [said](https://github.com/WordPress/gutenberg/pull/38135/files#r794378548):
> Nit: for consistency, it might use a more explicit name like useEntityRecordPermissions since it works only for a single record.

API resources are separate from entity records, we refer to them using a different set of identifiers.

> What's the story for multiple records?

You can either pass an id to this hook and get a create/update/delete permissions, or skip an id and get only the create permissions. Both are useful, e.g. in the Gutenberg codebase we sometimes do `canUser( 'update', 'settings' );` and other times we do `canUser( 'delete', resource, postId )`.

cc @kevin940726 @talldan @gziolo @draganescu @ellatrix @noisysocks @jsnajdr @getdave @Mamaduka @spencerfinnell

PR #63615, opened 2024-07-16
core-data: Fix `canUser` allowed methods handling

tyxla:
## What?
This PR cleans up a wrong access for retrieving the allowed methods inside `canUser`. 

## Why?
See https://github.com/WordPress/gutenberg/pull/63430#discussion_r1679400833

Because it's not necessary and was wrong. It was introduced in #43117 and the related test that needed to be improved was added in #38785.

## How?
Fixing the test fixture and removing the unnecessary object access. 

## Testing Instructions
* Verify all tests pass.
* Smoke test the site editor and verify all requests continue to work. 

None

## Screenshots or screencast 
None

### DIFF
--- a/packages/core-data/src/hooks/test/use-resource-permissions.js
+++ b/packages/core-data/src/hooks/test/use-resource-permissions.js
@@ -24,11 +24,9 @@ describe( 'useResourcePermissions', () => {
 		registry.register( coreDataStore );
 
 		triggerFetch.mockImplementation( () => ( {
-			headers: {
-				get: () => ( {
-					allow: 'POST',
-				} ),
-			},
+			headers: new Headers( {
+				allow: 'POST',
+			} ),
 		} ) );
 	} );
 
--- a/packages/core-data/src/resolvers.js
+++ b/packages/core-data/src/resolvers.js
@@ -419,8 +419,7 @@ export const canUser =
 		// Optional chaining operator is used here because the API requests don't
 		// return the expected result in the native version. Instead, API requests
 		// only return the result, without including response properties like the headers.
-		const allowHeader = response.headers?.get( 'allow' );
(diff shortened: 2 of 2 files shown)


packages/core-data/src/hooks/test/use-resource-permissions.js
```diff
@@ -24,11 +24,9 @@ describe( 'useResourcePermissions', () => {
...
 
 		triggerFetch.mockImplementation( () => ( {
-			headers: {
-				get: () => ( {
-					allow: 'POST',
-				} ),
-			},
+			headers: new Headers( {
+				allow: 'POST',
+			} ),
```
tyxla, line 29: [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers) is the right interface for dealing with response headers. Just using the constructor adds the necessary functionality, including proper support for `.get()`.

packages/core-data/src/resolvers.js
```diff
@@ -419,8 +419,7 @@ export const canUser =
 		// Optional chaining operator is used here because the API requests don't
 		// return the expected result in the native version. Instead, API requests
 		// only return the result, without including response properties like the headers.
-		const allowHeader = response.headers?.get( 'allow' );
-		const allowedMethods = allowHeader?.allow || allowHeader || '';
+		const allowedMethods = response.headers?.get( 'allow' ) || '';
```
tyxla, line 422: Cleaning up as the `.allow` access was not necessary.

Mamaduka (approved):
Thank you, @tyxla!

jsnajdr:
Interesting, so the `allowHeader?.allow` bit was there only to satisfy a wrong test?

In lodash, `includes( { allow: 'GET' }, 'GET' )` is true because the object is implicitly converted to array with `_.values`. So with the lodash code the test worked 🙂

PR #63617, opened 2024-07-16
Polish create template part modal

t-hamano:
## What?

This PR makes some changes to the create template modal to make it more consistent with other modals.

## Why? / How?

- Simplify the title: The current title is wordy and the multi-line text is a bit hard to read.
- Change the height of the text field and button to 40px
- Remove the default value from the text field and apply a placeholder instead: Templates, patterns and page create modals do not have default values.
- Add "Cancel" button: I think all modals need a cancel button unless there is a special reason not to.
- Apply `small` size to the modal
- Remove unused classname

## Testing Instructions

- Insert the Template Part or Header/Footer block.
  Note: If you have enabled TT4, the Header template part already exists, so even if you insert a Header block, no placeholder will be displayed. It's a good idea to test it with Emptytheme.
- Click the "Start blank" button.
- Creating template parts should still work as before.

## Screenshots or screencast 

| Before | After |
|--------|--------|
| template_part_before | image |
| header_before | image |


jameskoster:
Excellent tidy-up, thank you! Do you think we should auto-focus the text input when the modal opens like we did in https://github.com/WordPress/gutenberg/pull/62788? 

> Add "Cancel" button: I think all modals need a cancel button unless there is a special reason not to. 

It would be good to define this at the system level somehow. Close _and_ cancel buttons feel a bit duplicative.

t-hamano:
> Do you think we should auto-focus the text input when the modal opens like we did in https://github.com/WordPress/gutenberg/pull/62788?

Yes, I think that is definitely necessary. Updated by 5a0f772.

jameskoster (approved):

Issue #63604, opened 2024-07-16
List type style (list bullet) is different when nested inside another list


carolinan:
### Description

WordPress does not set the list type style for the list block or on blocks that uses a list element.
These blocks uses the browser default styles, and uses disc, followed by circle and then square for the nested lists.

This means that when the block is nested inside another list, the list type style starts from the circle, not the disc, and this makes the design inconsistent.

The list block is nested when it is previewed:
- In the query/ post template + post content block
- In the latest post block

This issue needs testing to learn which other blocks that are affected.

### Step-by-step reproduction instructions

I tested this on a blank theme created with "Create block theme" to make sure there were no theme styles.

Create a new post.
Add a list block. Example:
```
<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>One: disc<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>nested: circle<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>two levels: square<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>three levels: square</li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item --></ul>
<!-- /wp:list -->
```
Save.
Add a latest posts block.
Enable the post content option, and select "show full post".
Set the number of items to 1.
Confirm that the list types styles are not the same.

### Screenshots, screen recording, code snippet

The first block is the list. The second is the latest post block. The bullet styles do not match.
![list-bullets](https://github.com/user-attachments/assets/acf8d9f0-56b7-4fab-bfe2-455a5e2f396c)

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

- [X] Yes

Rishit30G:
Hey @carolinan 
Thanks for sharing the issue, I could recreate it by reading your instructions 🙌🏻  
As per my testing I could only find the issue in 'Latest Post' Block, I could also find a possible fix for the same, sharing the link of the PR: https://github.com/WordPress/gutenberg/pull/63622

dickmoomaw:
== Reproduction Report
I was able to reproduce the issue using @carolinan method

[...]

PR #63622, opened 2024-07-16
Fix: List type style (list bullet) is different when nested inside another list

Rishit30G:
## What?
The PR is intended to solve the issue of Latest Posts displaying inconsistent list style 
Ref: https://github.com/WordPress/gutenberg/issues/63604

## Why ? 

WordPress does not set the list type style for the list block or on blocks that uses a list element.
These blocks uses the browser default styles, and uses disc, followed by circle and then square for the nested lists.
This means that when the block is nested inside another list, the list type style starts from the circle, not the disc, and this makes the design inconsistent.

## How?
This PR adds necessary SCSS style to the `wp-block-latest-posts__post-full-content` class to display the nested list items properly in the latest posts block 
image
The `.wp-block-latest-posts__post-full-content` class targets list items (li) and sets their list style type based on their nesting level: disc for the first level, circle for the second, and square for the third.

## Testing Instructions
Add a list block. Example:

```
<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>One: disc<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>nested: circle<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>two levels: square<!-- wp:list -->
<ul class="wp-block-list"><!-- wp:list-item -->
<li>three levels: square</li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item --></ul>
<!-- /wp:list --></li>
<!-- /wp:list-item --></ul>
<!-- /wp:list -->
```
Save.
Add a latest posts block.
Enable the post content option, and select "show full post".
Set the number of items to 1.
Confirm that the list types styles are not the same.

#### Created a new post with nested list items: 
image

#### Added the post inside another post using the Latest Post block 
image

#### Post after saving
image


carolinan:
Hi @Rishit30G 
I understand it has been a long wait; Can you update your branch to include the latest changes from trunk? I am not able to re-run the tests.

Rishit30G:
Hi @carolinan! 
Sorry for the delay, I’ve updated the branch.

carolinan:
Hi, the PR can be applied now, but the CSS does not take into account that the list block can be set to ordered.
When it is ordered, there is an option to change the list style.
With the option enabled, the style is added inline but on the OL:
`<ol start="19" style="list-style-type:upper-roman" class="wp-block-list">`
 This PR overrides that option.

PR #63623, opened 2024-07-16
Template Part: Add check if create action should be allowed

### LINKED ISSUE #58344: Template parts can be created unintentionally with a hybrid theme
### Description

Found in this comment on the core ticket: https://core.trac.wordpress.org/ticket/58827#comment:21

Classic themes that support template parts can access the template parts parts in the Site Editor. Users should only be able to edit template parts provided by the theme and should not be allowed to create new template parts. However, it is possible to unintentionally create a new template part in the template parts editor.

- Access the test environment (`http://localhost:8889/wp-admin`).
- Activate the Emptyhybryd theme. This theme has opted in to block template part support.
- Go to Appearance > Template Parts > header > editor canvas.
- You should be able to insert template parts and create new template parts.

https://github.com/WordPress/gutenberg/assets/54422211/75d01ea0-99e2-4b0c-aee5-9446857db10a

### Environment info

- WordPress Version: 6.4.2
- Gutenberg Version: 17.6.0-rc.2 (It can be reproduced regardless of whether it is enabled or disabled)

Yes

Yes

t-hamano:
Fixes #58344

## What?

This PR allows the creation of new template parts in a template part block only if the following conditions are met:

- If the current theme is a block theme
- If the current user has capabilities to create template parts

| Before | After |
|--------|--------|
| image | image| 

## Why?

Creating template parts is not the intended behavior if the theme is not a block theme or if the user does not have the permissions.

## How?

I added a check similar to [the Add New Pattern button dropdown](https://github.com/WordPress/gutenberg/blob/c55aea6d6c0459b324e216ab7f248670e7a783b4/packages/edit-site/src/components/add-new-pattern/index.js#L105).

Another approach would be to not allow the insertion of template part blocks if the current theme is not a block theme, but I did allow the insertion of template part blocks because there might be edge cases where users want to nest template parts.

## Testing Instructions

- Access `http://localhost:8889/wp-admin` and activate the `Emptyhybrid` theme.
- Go to Appearance > Patterns > All template parts > header
- Insert a template part block
-  Make sure there is no "Start blank" button.

## 🤔 About User Capabilities

Regarding user capabilities, I've come across an issue where we might not be able to disable the capability to create template parts.

I removed the `create_posts` capability via the code below to test this PR:

```php
add_filter(
	'register_post_type_args',
	function ( $args ) {
		return $args;
	}
);
```

But strangely, `canUser` always returns `true` _only_ for template parts:

```javascript
wp.data.select('core').canUser('create',{kind: 'postType',name: 'page'})
// >> false
wp.data.select('core').canUser('create',{kind: 'postType',name: 'post'})
// >> false
wp.data.select('core').canUser('create',{kind: 'postType',name: 'wp_block'})
// >> false
wp.data.select('core').canUser('create',{kind: 'postType',name: 'wp_template_part'})
// >> true
```

This issue may need to be investigated separately.

### DIFF
--- a/packages/block-library/src/template-part/edit/placeholder.js
+++ b/packages/block-library/src/template-part/edit/placeholder.js
@@ -4,6 +4,8 @@
 import { __, sprintf } from '@wordpress/i18n';
 import { Placeholder, Button, Spinner } from '@wordpress/components';
 import { useState } from '@wordpress/element';
+import { useSelect } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
 
 /**
  * Internal dependencies
@@ -28,6 +30,21 @@ export default function TemplatePartPlaceholder( {
 		templatePartId
 	);
 	const blockPatterns = useAlternativeBlockPatterns( area, clientId );
+
+	const { isBlockBasedTheme, canCreateTemplatePart } = useSelect(
+		( select ) => {
+			const { getCurrentTheme, canUser } = select( coreStore );
+			return {
+				isBlockBasedTheme: getCurrentTheme()?.is_block_theme,
+				canCreateTemplatePart: canUser( 'create', {
+					kind: 'postType',
+					name: 'wp_template_part',
+				} ),
+			};
+		},
+		[]
+	);
+
 	const [ showTitleModal, setShowTitleModal ] = useState( false );
 	const areaObject = useTemplatePartArea( area );
 	const createFromBlocks = useCreateTemplatePartFromBlocks(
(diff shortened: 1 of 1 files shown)


Mamaduka:
Thanks, @t-hamano! I'll try to test this properly tomorrow.

Re-caps issue: The "templates" endpoint is probably doing `edit_theme_options` directly instead of using `current_user_can( $post_type_obj->cap->create_posts )`.

packages/block-library/src/template-part/edit/placeholder.js
```diff
@@ -28,6 +30,21 @@ export default function TemplatePartPlaceholder( {
...
+
+	const { isBlockBasedTheme, canCreateTemplatePart } = useSelect(
+		( select ) => {
+			const { getCurrentTheme, canUser } = select( coreStore );
+			return {
+				isBlockBasedTheme: getCurrentTheme()?.is_block_theme,
+				canCreateTemplatePart: canUser( 'create', {
+					kind: 'postType',
+					name: 'wp_template_part',
+				} ),
```
Mamaduka, line 42: At this point, it seems like you can't modify the user caps required to create a template part, and anyone who can access this page will have `edit_theme_options` capabilities. This makes the current check redundant.

Templates endpoint permission check:

https://github.com/WordPress/wordpress-develop/blob/5a30482419f1b0bcc713a7fdee3a14afd67a1bca/src/wp-includes/rest-api/endpoints/class-wp-rest-templates-controller.php#L182-L198

Mamaduka (approved):
The fix here makes sense, and since we're doing similar checks in other places, like the pattern list view, I think it's good to land.

I left a general note regarding template and template part capabilities, which probably should be improved in the core.

t-hamano:
@Mamaduka Thanks for the review!

> At this point, it seems like you can't modify the user caps required to create a template part, and anyone who can access this page will have "edit_theme_options" capabilities.

Thanks for looking into this.

As suggested in #61637, I think it's expected that in the future the site editor will be accessible to users without the `edit_theme_options` capability. Perhaps as part of that, we'll also need to improve permission checking for endpoints overall.

Issue #49769, opened 2023-04-12
"Manage all templates" and "Manage all template parts" link not visible on smaller screen size


annezazu:
### Description

Passing on a report from @nomad-skateboarding-dev where on smaller screen sizes the "Manage all templates" and "Manage all template parts" doesn't appear with WP 6.2 and GB 15.5.1.

### Step-by-step reproduction instructions

1. Open the Site Editor 
2. Navigate to Templates or Template parts
3. Make screen size smaller and notice the manage option disappears for both add new template and add new template part.

### Screenshots, screen recording, code snippet

https://user-images.githubusercontent.com/26996883/231536726-df078945-431e-48c0-a39f-f4474d8395a6.mov

### Environment info

- WP 6.2
- GB 15.5.1
- TT3
- Chrome

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

nomad-skateboarding-dev:
Thank you sincerely @annezazu 🙏

t-hamano:
I believe that it is not difficult to display those links in smaller screen size. However, if those links were there, I currently see some design and UI issues.

- Two columns look cramped.
- No back button.

![all-template](https://github.com/WordPress/gutenberg/assets/54422211/cfdffbab-003f-4fa0-84a1-e26c6d78cebf)

- Modal is misaligned and does not fit on screen.

![add-template](https://github.com/WordPress/gutenberg/assets/54422211/529e4885-bfc7-475c-a435-9fff5126d74c)

afercia:
Integrating with some information from https://github.com/WordPress/gutenberg/issues/52794 which is a duplicate.

Overall, I'm not sure that hiding functionalities on smaller screens is the best option. That was the trend in the early days of the responsive design era, about 10 years ago, and it proved to be less than ideal. If the reasoning is 'these pages don't look good on smaller screens so we just hide them', then I'd argue the design of these pages isn't ideal in the first place, 

After latest updates to the editor, these links are now:
- Manage all pages + page templates
- Manage all of my patterns
- Manage all templates
- Manage all template parts

It is worth mentioning that these links point either to 'classic' administration screens or to new administration pages in the editor.

[...]

PR #63627, opened 2024-07-16
Fix: Add Template Modal layout in mobile view

t-hamano:
Fix the issue raised in [this comment](https://github.com/WordPress/gutenberg/issues/49769#issuecomment-1644048297)

## What?

This PR fixes the layout of the Add Template modal in the mobile view.

Note: This is a minimal fix. In the future, we may be able to improve it further.

## How?

- Move styles not needed for the mobile view to the `break-large` mixin
- Change grid columns from 3 to 2 for mobile view only

## Screenshots or screencast 

| Before | After |
|--------|--------|
| image| image |


jameskoster (approved):
Good fix, approving from the design side :)

PR #63628, opened 2024-07-16
Block Bindings: Don't provide default `canUserEditValue` in reducer

SantosGuillamot:
## What?
This was suggested [here](https://github.com/WordPress/gutenberg/pull/63117#discussion_r1679170259) by @gziolo .

Right now, we define a default function that returns `false` in the reducer to ensure that sources are not editable unless explicitly specified. However, I believe we can avoid that and just not calling the function when it is undefined.

## Why?
To keep consistency with the rest of the properties that don't have a default value.

## How?
I changed where we are calling `canUserEditValue` to be called conditionally. If it is not defined, it will return undefined, which will lock the editing anyways.

## Testing Instructions
* Sources with `canUserEdit` are editable. These should be tested by post-meta and pattern-overrides.
* Sources without `canUserEdit` should be locked. This should be covered in [this test](https://github.com/WordPress/gutenberg/blob/ff9d6df37ff1b43fb3788da05c5449cf705eb77b/test/e2e/specs/editor/various/block-bindings.spec.js#L115).

### DIFF
--- a/packages/blocks/src/api/test/registration.js
+++ b/packages/blocks/src/api/test/registration.js
@@ -1607,7 +1607,7 @@ describe( 'blocks', () => {
 			expect( source.setValue ).toBeUndefined();
 			expect( source.setValues ).toBeUndefined();
 			expect( source.getPlaceholder ).toBeUndefined();
-			expect( source.canUserEditValue() ).toBe( false );
+			expect( source.canUserEditValue ).toBeUndefined();
 			unregisterBlockBindingsSource( 'core/valid-source' );
 		} );
 
--- a/packages/block-editor/src/components/rich-text/index.js
+++ b/packages/block-editor/src/components/rich-text/index.js
@@ -188,7 +188,7 @@ export function RichTextWrapper(
 						binding.source
 					);
 					if (
-						! blockBindingsSource?.canUserEditValue( {
+						! blockBindingsSource?.canUserEditValue?.( {
 							select,
 							context: blockContext,
 							args: binding.args,
--- a/packages/block-library/src/button/edit.js
+++ b/packages/block-library/src/button/edit.js
@@ -247,7 +247,7 @@ function ButtonEdit( props ) {
 			return {
 				lockUrlControls:
 					!! metadata?.bindings?.url &&
-					! blockBindingsSource?.canUserEditValue( {
+					! blockBindingsSource?.canUserEditValue?.( {
 						select,
 						context,
 						args: metadata?.bindings?.url?.args,
--- a/packages/block-library/src/image/edit.js
+++ b/packages/block-library/src/image/edit.js
@@ -308,7 +308,7 @@ export function ImageEdit( {
 			return {
 				lockUrlControls:
 					!! metadata?.bindings?.url &&
-					! blockBindingsSource?.canUserEditValue( {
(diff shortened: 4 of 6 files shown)


packages/blocks/src/api/test/registration.js
```diff
@@ -1607,7 +1607,7 @@ describe( 'blocks', () => {
 			expect( source.setValue ).toBeUndefined();
 			expect( source.setValues ).toBeUndefined();
 			expect( source.getPlaceholder ).toBeUndefined();
-			expect( source.canUserEditValue() ).toBe( false );
+			expect( source.canUserEditValue ).toBeUndefined();
```
ockham, line 1610: There's also `.toBeFalsy()` (which is a slightly looser assertion and would work both with and without the other changes in this PR, thus "bridging" the change).

It's also a really small detail, so feel free to ignore 🤷‍♂️ 

```suggestion
			expect( source.canUserEditValue ).toBeFalsy();
```

ockham (approved):
I haven't tested this (since I'm not sure how to set up a block to use bindings, and to set the `canUserEditValue` property 😅 ), but the code looks good, and the test coverage gives some confidence that it should be fine to merge 👍

artemiomorales (commented):
I tested this by changing the Post Meta source's `canUserEditValue()` value to return false. That change is respected in all of the covered cases 👍

However, tangentially related to this PR, I also tested setting the Pattern Override source's `canUserEditValue()` to be false, and it had no effect, both on this branch and on `trunk`. Is that expected?

Lastly, the button's `rel` attribute was not affected by the locking mechanism, and I realized it's because we actually never added locking logic to the [button's `rel` attribute](https://github.com/WordPress/gutenberg/blob/88dffa048232c538ffdcaff6cd6a5ece2454706f/packages/block-library/src/button/edit.js#L380-L391). Shall we open a separate issue for that?

SantosGuillamot:
> However, tangentially related to this PR, I also tested setting the Pattern Override source's canUserEditValue() to be false, and it had no effect, both on this branch and on trunk. Is that expected?

I believe it is not expected, although it isn't a big deal because pattern overrides are expected to be editable. If I am not mistaken, it is because we are not considering the `__default` option that was introduced [here](https://github.com/WordPress/gutenberg/pull/60694). We would need to update [this logic](https://github.com/WordPress/gutenberg/blob/cc425c32a9fbed66141db2680fb368bfe777db25/packages/block-editor/src/components/rich-text/index.js#L166-L206).

> Lastly, the button's rel attribute was not affected by the locking mechanism, and I realized it's because we actually never added locking logic to the [button's rel attribute](https://github.com/WordPress/gutenberg/blob/88dffa048232c538ffdcaff6cd6a5ece2454706f/packages/block-library/src/button/edit.js#L380-L391). Shall we open a separate issue for that?

I would consider this a different issue not related to this pull request. If I am understanding it correctly, it is already present in `trunk`.

packages/blocks/src/api/test/registration.js
SantosGuillamot: I didn't know about that, thanks for the tip 🙂 On the other hand, I prefer keeping it `undefined` to keep consistency with the rest of the properties, and because it is actually expected to be `undefined` and not false after these changes.

PR #43457, opened 2022-08-21
Post Title: Add padding support

ndiego:
Related:

- https://github.com/WordPress/gutenberg/issues/43241
- https://github.com/WordPress/gutenberg/issues/43243

## What?
Add padding support to the Post Title block. 

## Why?
To create consistency across blocks. This also allows users to override padding applied to Post Title blocks when there is a background color.

## How?
Added the relevant block support in block.json

## Testing Instructions
1. Insert a new Post Title block. 
2. Confirm the Dimension control panel allows you to add padding.
3. Adding padding and configure. 

## Screenshots or screencast 
post-title-padding

PR #31125, opened 2021-04-23
Site title: Add padding

carolinan:
## Description

Adds the padding control to the site title block. For https://github.com/WordPress/gutenberg/issues/30918.

## How has this been tested?

Activate a theme that has support for custom spacing, for example TT1 Blocks.
Add a site title block, confirm that the padding works in the block editor, site editor, and front.

## Screenshots 
Site title padding

## Types of changes

Enhancement

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [ ] I've tested my changes with keyboard and screen readers. 
- [ ] My code has proper inline documentation. 
- [ ] I've included developer documentation if appropriate. 
- [ ] I've updated all React Native files affected by any refactorings/renamings in this PR (please manually search all `*.native.js` files for terms that need renaming or removal).

PR #63631, opened 2024-07-16
Site Title: Add border block support 

akasunil:
## What?

Add border block support to the `Site Title` block.

## Why?
`Site Title` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Site Title` block's border is configurable via Global Styles
- Edit template/page, Add `Site Title` block and Apply the border styles
- Verify that block borders take precedence over global styles

## Screenshots or screencast 
In backend:
Screenshot at Jul 16 22-30-29
In frontend: 
Screenshot at Jul 16 22-30-50


aaronrobertshaw (commented):
Thanks for working on this @akasunil 👍 

This tests well for me. 

https://github.com/user-attachments/assets/0d3fe2f7-264b-4c9a-a7b1-3a9d77aaa0dc




The only thing I think we might need is the adoption of the `box-sizing: border-box` for more consistent size/styling of the block. It seems like adoption of this is pretty inconsistent across blocks at the moment, so I'd like to get some second opinions. 

cc @andrewserong

andrewserong:
> The only thing I think we might need is the adoption of the box-sizing: border-box for more consistent size/styling of the block. It seems like adoption of this is pretty inconsistent across blocks at the moment, so I'd like to get some second opinions.

Thanks for the ping! Adding the border-box rule sounds consistent with other similar blocks, e.g. the post title block where it was introduced when padding support was added back in: https://github.com/WordPress/gutenberg/pull/43457.

The site title padding support was added way back in https://github.com/WordPress/gutenberg/pull/31125 so has been around for a long time.

Are there any concerns that adding box-sizing at this stage could affect sites out in the wild? If so, one option could be to guard the rule so it only applies if a border is present. That said, it _feels_ more like a fix to me to add in the rule.

I'll just ping @carolinan for visibility on this one as Carolina introduced padding for this block originally and might have thoughts on the matter.

aaronrobertshaw:
> That said, it feels more like a fix to me to add in the rule.

Agreed. 

> Are there any concerns that adding box-sizing at this stage could affect sites out in the wild?

I think it would have to be a rather small edge case where the site title block is given an explicit width or inherits one from its parent only to have padding pushing its effective width wider.

packages/block-library/src/site-title/style.scss
```diff
@@ -1,3 +1,7 @@
+:root :where(.wp-block-site-title) {
```
aaronrobertshaw:
```suggestion
.wp-block-site-title {
```
The `:root :where()` selector results in `0-1-0` specificity the same as `.wp-block-site-title` just using more characters and data. Perhaps we should just simplify that?

aaronrobertshaw (commented):
Thanks for the quick update making the border controls here optional 🚀 

I've given this another test and all is working as expected. I've left one small nit as an inline comment. Once we tweak that, this is looking ready to me. Nice work 👍

packages/block-library/src/site-title/style.scss
akasunil: Sure

aaronrobertshaw: Sorry @akasunil, I meant only for the `box-sizing` rule. The change in https://github.com/WordPress/gutenberg/pull/63631/commits/f880412530c690e0e0f4877d6d59a9564cce1dfa bumps the specificity of the `a` rule to `0-1-1`.

aaronrobertshaw (approved):
This is testing well for the adoption of border support 👍 

I restored the specificity for the inner link rule as the recent change in selector accidentally bumped it from `0-1-0` to `0-1-1`. 


https://github.com/user-attachments/assets/9adc784d-9dc3-4dd4-ac7b-c05ae0ffc365

packages/block-library/src/site-title/style.scss
aaronrobertshaw: To save another feedback cycle and review, I've tweaked the selector in https://github.com/WordPress/gutenberg/pull/63631/commits/95c4f7baba0b5506cea5f8bc5d5703a676eaaa23

Issue #38730, opened 2022-02-11
BaseControl: Deprecate bottom margin


mirka:
## What problem does this address?

The `BaseControl` component currently has an [inherent `margin-bottom`](https://github.com/WordPress/gutenberg/blob/fdce407ae43fc121f36797b96e6d808ed1f3077e/packages/components/src/base-control/styles/base-control-styles.js#L19), making it hard to reuse in different contexts, since it requires a style override of an internal element. This kind of internal override is something we want to discourage.

Additionally, the current margin-bottom leads to inconsistent results when `help` text is supplied, because the margin is inserted _before_ and not after the help text.

## What is your proposed solution?

Deprecate the bottom margin (similar to #37160), following the [guidelines](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/CONTRIBUTING.md#deprecating-styles) stipulated in the devdocs.

Because this change impacts a large number of components (even just inside the Gutenberg codebase), I suggest we approach this in phases:

- [x] 1. Add a `__nextHasNoMarginBottom` prop to BaseControl to opt-in to the correct styles, without adding a deprecation warning
  - [x] BaseControl #39325
  - [x] CheckboxControl #43720
  - [x] ComboboxControl #43165
  - [x] DimensionControl #64346
  - [x] FocalPointPicker #43996
  - [x] RadioControl (cleanup of ad hoc override styles) #43868
  - [x] RangeControl #43062
  - [x] SearchControl #43871
  - [x] SelectControl #41269
  - [x] TextControl #43782
  - [x] TextareaControl #43782
  - [x] ToggleControl (also [tweak spacing before help text](https://github.com/WordPress/gutenberg/pull/39209#issuecomment-1072175095)) #43717
  - [x] ToggleGroupControl #43062
  - [x] TreeSelect (Based on SelectControl)
- [x] 2. Migrate all usages in the Gutenberg codebase (some components may require official deprecations)
  - [x] CheckboxControl #45434
  - [x] ComboboxControl #45796
  - [x] FocalPointPicker #45958
  - [x] RadioControl (cleanup of adhoc override styles) #43868
  - [x] RangeControl #45985
  - [x] SearchControl #46081
  - [x] SelectControl #46448 
  - [x] TextControl
	  - [x] #47157
	  - [x] #47158
	  - [x] #47160
	  - [x] #47161
  - [x] TextareaControl #46559
  - [x] ToggleControl #47866
  - [x] ToggleGroupControl #47269
[...]

PR #63633, opened 2024-07-16
Add margin-bottom lint rules for FocalPointPicker, TextareaControl, TreeSelect

mirka:
Part of #38730

## What?

Adds eslint rules to prevent new instances of the following components to be introduced in the Gutenberg codebase without the `__nextHasNoMarginBottom` prop being added:

- FocalPointPicker
- TextareaControl
- TreeSelect

## Why?

We basically should've done this when we did the initial migration work for each component — we underestimated the time it would take to complete the entire deprecation process for the `BaseControl`-based family of components.

The three components being linted here do not have new violations from the time we did the original migrations.

While I work through the list of components again to fix the new violations, these lint rules should prevent new violating usages from being added. Once all that is done, we can start logging an actual deprecation warning.

## Testing Instructions

The linter should pass. It should fail if you add a still-violating component like `ComboboxControl` to the array.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -268,7 +268,6 @@ module.exports = {
 				'packages/*/src/**/*.[tj]s?(x)',
 				'storybook/stories/**/*.[tj]s?(x)',
 			],
-			excludedFiles: [ '**/*.@(native|ios|android).js' ],
 			rules: {
 				'no-restricted-syntax': [
 					'error',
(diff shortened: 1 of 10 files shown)


.eslintrc.js
```diff
@@ -392,6 +400,7 @@ module.exports = {
 			excludedFiles: [
 				'packages/components/src/utils/colors-values.js',
 				'packages/components/src/theme/**',
+				'**/*.@(native|ios|android).js',
```
mirka: Not excluding these mobile files was an oversight.

ciampo (approved):
🚀 

- ✅ Removing `__nextHasNoMarginBottom` from linted components causes ESLint to show an error
- ✅ Adding new components (like `ComboboxControl`) to the list also causes ESLint to error

.eslintrc.js
aicuaboom: 👾

Issue #56442, opened 2023-11-22
Fix patterns explorer pop-up sidebar navigation


andrewhayward:
## What problem does this address?

![The patterns explorer dialog, with "all patterns" selected in the sidebar](https://github.com/WordPress/gutenberg/assets/159848/858812b3-c99a-45e5-941d-3f606512c466)

The sidebar of the patterns explorer pop-up is currently a collection of buttons, with no semantic indication of their purpose or context.

## What is your proposed solution?

[The tab pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) would probably be most appropriate here, but it is arguably [a menu](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) too.

afercia:
Noting that making these buttons anr ARIA tab interface would partially solve https://github.com/WordPress/gutenberg/issues/55110 / https://github.com/WordPress/gutenberg/pull/56162

Regarding the general issue, I wouldn't recommend to use modal dialogs for such highly dynamic user interfaces in teh first place. Having the dialog content that updates dynamically doesn't seem to me ideal from an usability perspective. Also, it creates problems with some important features of the modal dialog such as constrainet tabbind and detecting focus outside to close the dialog.

That said, I'd be all for consistency in the first place. There are other modal dialogs that use some mechanism to 'navigate' content. For example, the Preferences modal dialog does use ARIA tabs:

![Screenshot 2023-11-23 at 16 45 56](https://github.com/WordPress/gutenberg/assets/1682452/ce822708-b036-4abb-b216-ddf95fbd6ea6)

Conceptually, it's very similar to the Patterns explorer dialog. On the left, there's a collection of buttons that allow 'navigating' the content. Should this pattern always use Tabs? Or maybe use buttons wrapped within a properly labelled  `nav` element?

Does the Tab design pattern helps users?
- It does reduce the amount of Tab stops, while still allowing to navigate / select the tabs with arrow keys.
- Does it help discoverability of all the preference sections and the Pattern categories?

Anyways, I'd vote for standardizing and always use either Tabs or simple buttons + nav.

joedolson:
From all of the related issues, it seems like we have a fundamental problem that we're making heavy use of a pattern with a side panel of navigation and a main panel of content, and not using a consistent navigation across the site. 

[...]

PR #63636, opened 2024-07-16
Remove pattern explorer modal

jeryj:
## What?

Removes the pattern explorer modal from the pattern inserter sidebar.

## Why?

The modal is not adding anything not already available in the pattern sidebar. Rather than have an alternate experience (that is not working well:
- Inserts patterns behind the modal (you can't see that they have been inserted)
- https://github.com/WordPress/gutenberg/issues/56442
- Tabbing around causes odd flashing and removal of content
- Grid layout but must use down arrow to navigate content

## How?

Remove it, rely on the pattern sidebar in the inserter


packages/block-editor/src/components/inserter/style.scss
stokesman, line 1: Can’t lines 430-491 also be removed? The `.block-editor-block-patterns-explorer` stuff.

jameskoster:
Iirc this modal used to be the only way to retrieve patterns from the directory. But since that's no longer the case it does feel a bit duplicative. I'd be interested to hear @richtabor's thoughts.

packages/block-editor/src/components/block-patterns-list/style.scss
```diff
@@ -18,13 +18,6 @@
 
 .block-editor-block-patterns-list__item {
 	height: 100%;
-	// This is derived from the top padding set on
-	// `.block-editor-block-patterns-explorer__list`
-	scroll-margin-top: $grid-unit-30;
-	// This is derived from the bottom padding set on
-	// `.block-editor-block-patterns-explorer__list` and
-	// the bottom margin set on `...__list-item` above
-	scroll-margin-bottom: ($grid-unit-40 + $grid-unit-30);
```
jeryj, line 27: I couldn't get these to apply when viewing the query pattern explorer or the sidebar inserter pattern explorer. None of them had a snapping effect for me.

packages/block-editor/src/components/inserter/style.scss
jeryj: Good catch. Thanks!

richtabor:
We shouldn't remove this, unless it's clear that people don't use it, and we replace any loss in functionality. 

As @jameskoster mentioned, it's how search results from the pattern directory can be pulled in.

jeryj:
@richtabor What features are available in the modal that aren't available in the sidebar inserter?

stokesman:
> it's how search results from the pattern directory can be pulled in.

From what I see the default "All" view is pulling in all the same patterns as when I open the explorer. That agrees with James’s comment https://github.com/WordPress/gutenberg/pull/63636#issuecomment-2232917523. There doesn’t seem to be an absolute need for the explorer anymore.

jeryj:
Closing as it seems like we're not ready to remove this yet.

PR #63639, opened 2024-07-17
JSON Schema Docgen Rework

ajlende:
## What?

Reworks the theme.json reference docgen.

Key features include:
- Descriptions for styles and settings properties.
- Reworked type serialization to remove the props heading for styles and settings.
- Better types everywhere in the document.
- Details for settings and styles.
- The useRootPaddingAwareAlignments top-level property returns.
- Cleaner code.

## Why?

Split out of #63591 so that both the schema changes and codegen changes can be reviewed independently from one another.

## How?

Slow refactoring of the docgen code bit by bit. Check out the commits for details.

## Testing Instructions

See that the [new docs](https://github.com/WordPress/gutenberg/blob/update/json-schema-codegen-rework/docs/reference-guides/theme-json-reference/theme-json-living.md) look correct.

N/A

## Screenshots or screencast 

N/A


docs/reference-guides/theme-json-reference/theme-json-living.md
```diff
@@ -36,318 +51,311 @@ Setting that enables the following UI tools:
...
 
 ### background
 
 Settings related to background.
 
-| Property  | Type   | Default | Props  |
-| ---    | ---    | ---    |---   |
-| backgroundImage | boolean | false |  |
-| backgroundSize | boolean | false |  |
+| Property | Description | Type | Default |
```
scruffian, line 60: I much prefer these headings.

scruffian (approved):
This is looking good. I deleted the auto generated section and ran `node bin/api-docs/gen-theme-reference.mjs` and it restored things as they are.

t-hamano (approved):
LGTM! Compared to previous documents, it is much easier to read, and the document generation logic seems to have been greatly refined.

Here's what the Block Editor Handbook might look like with this PR applied. I tested it by making a temporary change to one table:

https://github.com/user-attachments/assets/69fc41fd-a164-4b3d-b06b-3c5dbeda1513

PR #63642, opened 2024-07-17
Zoom Out: fix crash due to absence of selected block

### LINKED ISSUE #63456: Editor crash from block deselection in “Zoom Out” mode

stokesman:
## What?
A fix for crashes in the editor while zoomed out when either a block is removed or a block is deselected.

## Why?
Without this it’s fairly easy to crash the editor in Zoom Out. To fix #63456.

## How?
Revises the show block tools hook to avoid returning `true` for `showZoomOutToolbar` when no block is selected.

### Undo while zoomed out
1. With the Zoom Out experiment enabled
2. Go to the Site editor and select a page to edit
3. Open the inserter, select a pattern category and insert a pattern
4. Undo
5. Note there is no crash

### Deselect a block while zoomed out
1. With the Zoom Out experiment enabled
2. Open a post that has content
3. Open the inserter, select a pattern category
4. Select a block
5. Click somewhere that will deselect the block without selecting another
6. Note there is no crash

### DIFF
--- a/packages/block-editor/src/components/block-tools/use-show-block-tools.js
+++ b/packages/block-editor/src/components/block-tools/use-show-block-tools.js
@@ -29,10 +29,11 @@ export function useShowBlockTools() {
 		const clientId =
 			getSelectedBlockClientId() || getFirstMultiSelectedBlockClientId();
 
-		const block = getBlock( clientId ) || { name: '', attributes: {} };
+		const block = getBlock( clientId );
 		const editorMode = __unstableGetEditorMode();
-		const hasSelectedBlock = clientId && block?.name;
-		const isEmptyDefaultBlock = isUnmodifiedDefaultBlock( block );
+		const hasSelectedBlock = !! clientId && !! block;
+		const isEmptyDefaultBlock =
+			hasSelectedBlock && isUnmodifiedDefaultBlock( block );
 		const _showEmptyBlockSideInserter =
 			clientId &&
 			! isTyping() &&
@@ -43,8 +44,9 @@ export function useShowBlockTools() {
 			! hasMultiSelection() &&
 			editorMode === 'navigation';
 
+		const isZoomOut = editorMode === 'zoom-out';
 		const _showBlockToolbarPopover =
-			editorMode !== 'zoom-out' &&
+			! isZoomOut &&
 			! getSettings().hasFixedToolbar &&
 			! _showEmptyBlockSideInserter &&
 			hasSelectedBlock &&
@@ -57,7 +59,8 @@ export function useShowBlockTools() {
 				! _showEmptyBlockSideInserter && maybeShowBreadcrumb,
 			showBlockToolbarPopover: _showBlockToolbarPopover,
 			showZoomOutToolbar:
-				editorMode === 'zoom-out' &&
+				hasSelectedBlock &&
+				isZoomOut &&
 				! _showEmptyBlockSideInserter &&
 				! maybeShowBreadcrumb &&
 				! _showBlockToolbarPopover,


packages/block-editor/src/components/block-tools/use-show-block-tools.js
```diff
@@ -57,7 +59,8 @@ export function useShowBlockTools() {
 				! _showEmptyBlockSideInserter && maybeShowBreadcrumb,
 			showBlockToolbarPopover: _showBlockToolbarPopover,
 			showZoomOutToolbar:
-				editorMode === 'zoom-out' &&
+				hasSelectedBlock &&
```
stokesman, line 62: So this was probably the only key change to fix the issue. The other changes I felt were minor code quality improvements.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
```diff
@@ -29,10 +29,11 @@ export function useShowBlockTools() {
 		const clientId =
 			getSelectedBlockClientId() || getFirstMultiSelectedBlockClientId();
 
-		const block = getBlock( clientId ) || { name: '', attributes: {} };
+		const block = getBlock( clientId );
```
jeryj, line 32: If I recall, we wanted to keep the structure the same if nothing was found due to memoization: `{ name: '', attributes: {} };`. I think @Mamaduka recommended this.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
```diff
@@ -29,10 +29,11 @@ export function useShowBlockTools() {
...
 		const editorMode = __unstableGetEditorMode();
-		const hasSelectedBlock = clientId && block?.name;
-		const isEmptyDefaultBlock = isUnmodifiedDefaultBlock( block );
+		const hasSelectedBlock = !! clientId && !! block;
+		const isEmptyDefaultBlock =
+			hasSelectedBlock && isUnmodifiedDefaultBlock( block );
```
jeryj, line 36: I don't think we should tie the `hasSelectedBlock` condition here to the `isEmptyDefaultBlock`.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
```diff
@@ -29,10 +29,11 @@ export function useShowBlockTools() {
 		const clientId =
 			getSelectedBlockClientId() || getFirstMultiSelectedBlockClientId();
 
-		const block = getBlock( clientId ) || { name: '', attributes: {} };
+		const block = getBlock( clientId );
 		const editorMode = __unstableGetEditorMode();
-		const hasSelectedBlock = clientId && block?.name;
-		const isEmptyDefaultBlock = isUnmodifiedDefaultBlock( block );
+		const hasSelectedBlock = !! clientId && !! block;
```
jeryj, line 34: I think the `!!` for this line is a good change, but undo the other refactoring.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
jeryj: Nice find!

jeryj (commented):
I was going to fix this and found you had already put in a PR. Thanks for the zoom out improvements! I think let's fix just the root issue and maybe add the double bang on `hasSelectedBlock`. Otherwise I think we should leave the rest of the hook the same for now.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
stokesman: It seems like the only reason for this is so that `block` can be passed to `isUnmodifiedDefaultBlock` which doesn’t guard against nullish values. Beyond that `block` is unused so it’s okay to be nullish.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
stokesman: Why even call `isUnmodifiedDefaultBlock` if there is no selected block? I.e. without a selection it can’t be a empty default block right?

stokesman:
Thanks for the review 🙇 

> I think let's fix just the root issue and maybe add the double bang on `hasSelectedBlock`. Otherwise I think we should leave the rest of the hook the same for now.

I know I've made the review less simple so I'm okay with that. Though maybe it’s not too much to ask letting this get a second look before that.

packages/block-editor/src/components/block-tools/use-show-block-tools.js
Mamaduka: Yeah, it's just safe guard - https://github.com/WordPress/gutenberg/pull/58979#discussion_r1502978699.

scruffian (approved):
LGTM. Thanks for the fix.

Mamaduka (approved):
The fix tests well for me ✅

Note: I noticed that the undoing pattern insertion is somewhat broken. @jeryj, do you know it's being tracked somewhere?

## Screencast

https://github.com/user-attachments/assets/ba2dad91-d048-4ff7-8757-6c01e3ff8df3

stokesman:
> I noticed that the undoing pattern insertion is somewhat broken. @jeryj, do you know it's being tracked somewhere?

Aki had made #63479 regarding the undo, maybe it was closed prematurely or this problem is distinct.

stokesman:
Thanks to everyone who tested and reviewed. Especially Jerry, whose apt scrutiny of the unnecessary changes made great starting points to explain them.

PR #63643, opened 2024-07-17
Image block: Show placeholder when uploading HEIC files

noisysocks:
## What, why, how

In https://core.trac.wordpress.org/ticket/53645 we're adding support for HEIC images to Core. These will be automatically converted to a JPEG on the server after they're uploaded.

This PR improves the UX of uploading a HEIC image in a browser that doesn't support viewing HEIC images e.g. Chrome, Firefox.

Currently the temporary blob image that displays during upload can't be displayed and so the browser will display the default alt text during upload:

https://github.com/user-attachments/assets/b5154500-4818-4dfc-b45a-31bad838682a

This PR detects when the image fails to load and during upload shows a placeholder `div` instead:

https://github.com/user-attachments/assets/dda78ffe-3650-4d95-88b7-f00834d0a6d8

In browsers that support HEIC aka Safari, you'll see a preview of the image while it uploads.

### Testing instructions

It's a bit tricky, sorry. You'll have to set Gutenberg to use this branch and WordPress to use `adamsilverstein:add/heic-support` ([wordpress-develop#7034](https://github.com/WordPress/wordpress-develop/pull/7034)).

One way to do that is to:
1. Clone `wordpress-develop` locally and `cd` to it
2. Check out `adamsilverstein:add/heic-support`
3. `npm install; npx grunt`
4. `cd` back to your local Gutenberg repo
5. Create a `.wp-env.override.json` with `{ "core": "/path/to/your/local/wordpress-develop/build" }
6. `npx wp-env start` 

Unfortunately we can't use Playground because it doesn't support ImageMagick.

Once your testing environment is set up:

1. Open the post or site editor
2. Upload a HEIC image. There are sample images available here: https://github.com/dsoprea/heic-exif-samples
3. Test both a browser that supports HEIC (Safari) and one that doesn't (Firefox, Chrome)

### DIFF
--- a/packages/block-library/src/image/editor.scss
+++ b/packages/block-library/src/image/editor.scss
@@ -3,7 +3,7 @@
 .wp-block-image.wp-block-image {
 
 	// Show Placeholder style on-select.
-	&.is-selected .components-placeholder {
+	&.is-selected .block-editor-media-placeholder {
 		// Block UI appearance.
 		color: $gray-900;
 		background-color: $white;
@@ -61,12 +61,6 @@ figure.wp-block-image:not(.wp-block) {
 		display: inline;
 	}
 
-	// Shown while image is being uploded but cannot be previewed.
-	.block-editor-image__skeleton {
-		aspect-ratio: 4 / 3;
-		background: $gray-100;
-	}
-
 	// Shown while image is being uploaded.
 	.components-spinner {
 		position: absolute;
@@ -74,6 +68,11 @@ figure.wp-block-image:not(.wp-block) {
 		left: 50%;
 		transform: translate(-50%, -50%);
 	}
+}
+
+// Shown while image is being uploded but cannot be previewed.
+.wp-block-image__placeholder {
+	aspect-ratio: 4 / 3;
 }
 
 // This is necessary for the editor resize handles to accurately work on a non-floated, non-resized, small image.
--- a/packages/block-library/src/image/editor.scss
+++ b/packages/block-library/src/image/editor.scss
@@ -61,6 +61,12 @@ figure.wp-block-image:not(.wp-block) {
 		display: inline;
 	}
 
+	// Shown while image is being uploded but cannot be previewed.
+	.block-editor-image__skeleton {
+		aspect-ratio: 4 / 3;
+		background: $gray-100;
+	}
+
 	// Shown while image is being uploaded.
 	.components-spinner {
 		position: absolute;
--- a/packages/base-styles/_mixins.scss
+++ b/packages/base-styles/_mixins.scss
@@ -174,7 +174,7 @@
 		left: 0;
 		pointer-events: none;
 		background: currentColor;
-		opacity: 0.2;
+		opacity: 0.1;
 	}
 }
 
--- a/packages/base-styles/_mixins.scss
+++ b/packages/base-styles/_mixins.scss
@@ -174,7 +174,7 @@
 		left: 0;
 		pointer-events: none;
 		background: currentColor;
-		opacity: 0.1;
+		opacity: 0.2;
 	}
 }
 
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -176,6 +176,7 @@ export default function Image( {
 	] = useState( {} );
 	const [ isEditingImage, setIsEditingImage ] = useState( false );
 	const [ externalBlob, setExternalBlob ] = useState();
+	const [ hasImageErrored, setHasImageErrored ] = useState( false );
 	const hasNonContentControls = blockEditingMode === 'default';
 	const isContentOnlyMode = blockEditingMode === 'contentOnly';
 	const isResizable =
@@ -245,13 +246,22 @@ export default function Image( {
 	}
 
 	function onImageError() {
+		setHasImageErrored( true );
+
 		// Check if there's an embed block that handles this URL, e.g., instagram URL.
 		// See: https://github.com/WordPress/gutenberg/pull/11472
 		const embedBlock = createUpgradedEmbedBlock( { attributes: { url } } );
-
 		if ( undefined !== embedBlock ) {
 			onReplace( embedBlock );
 		}
+	}
+
+	function onImageLoad( event ) {
+		setHasImageErrored( false );
+		setLoadedNaturalSize( {
+			loadedNaturalWidth: event.target?.naturalWidth,
+			loadedNaturalHeight: event.target?.naturalHeight,
+		} );
 	}
 
 	function onSetHref( props ) {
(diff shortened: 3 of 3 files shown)


packages/block-library/src/image/editor.scss
```diff
@@ -61,6 +61,12 @@ figure.wp-block-image:not(.wp-block) {
 		display: inline;
 	}
 
+	// Shown while image is being uploded but cannot be previewed.
+	.block-editor-image__skeleton {
+		aspect-ratio: 4 / 3;
+		background: $gray-100;
+	}
+
```
andrewserong: Would it be worth reusing the placeholder state for when an image is deselected? Currently that placeholder state uses a transparent background, and a different size, so with the gray color and 4/3, I notice a flash of gray and a different height during the upload. It's more noticeable if the image block exists on top of a darker background.

https://github.com/user-attachments/assets/fa769845-a91c-4564-a6dc-86861562d80e

It's very quick in the video since I'm testing Gutenberg without the core patch, so might be easy to miss 😄

andrewserong (commented):
Great idea!

Just left a comment about the placeholder state and whether it'd be worth reusing the existing placeholder if possible, to retain the transparent background and the height if the image was already in a placeholder state?

packages/block-library/src/image/editor.scss
noisysocks: Deselected!! I thought we already had a placeholder state, but couldn't find it 😛 Good suggestion I'll do that.

swissspidy:
Thanks for working on this 👍  This was on my todo list for a while as I need it for https://github.com/swissspidy/media-experiments (where this is also noticeable with HEIF images, but also MOV videos)

noisysocks:
> Thanks for working on this 👍 This was on my todo list for a while as I need it for https://github.com/swissspidy/media-experiments (where this is also noticeable with HEIF images, but also MOV videos)

Ideally when we have client side HEIC processing we'd still show the preview once it's been converted on the client. Future problem though 🙂

packages/block-library/src/image/editor.scss
noisysocks: I've switched to using `Placeholder`.

I'm not a big fan of how the spinner looks on top of the diagonal line, though. It makes the spinner hard to see.

Screenshot 2024-07-18 at 14 47 31

Mind you, this problem also exists on `trunk`. When there is a preview it can be hard to see the spinner depending on the image being uploaded.

Screenshot 2024-07-18 at 14 48 52

Maybe we should put the spinner in a translucent box?

Screenshot 2024-07-18 at 14 47 22

Screenshot 2024-07-18 at 14 51 43

Or maybe we should denote that the image is uploading using a different technique e.g. pulsating the block's opacity?

Or maybe it's fine how it is?

What do you think @jasmussen?

andrewserong: > What do you think @jasmussen?

I'm not sure if Joen is around at the moment, so I'll just give a soft ping to @WordPress/gutenberg-design for visibility 🙂

noisysocks: Good plan. We can come back to this later, at any rate. HEIC is not very common and what's in this PR is better than what's in trunk.

jameskoster: I'd lean towards tweaking the overlay color/opacity a bit, so that the spinner remains visible, rather than adding another smaller 'surface' behind the spinner.

Another option could be a different loading animation in place of the spinner. The overlay could pulse between different opacity levels. That might require more holistic exploration though, so identify other applications.

swissspidy:
> > Thanks for working on this 👍 This was on my todo list for a while as I need it for [swissspidy/media-experiments](https://github.com/swissspidy/media-experiments) (where this is also noticeable with HEIF images, but also MOV videos)
> 
> Ideally when we have client side HEIC processing we'd still show the preview once it's been converted on the client. Future problem though 🙂

That happens automatically already, the image is updated after processing :)

andrewserong (approved):
Thanks for the iterations here!

> HEIC is not very common and what's in this PR is better than what's in trunk.

@jameskoster had some good suggestions in the thread, but if you wanted to land this as a first pass, I think you're right, it's an improvement on what's in `trunk`, and it doesn't affect the usual flow. So this LGTM! But do feel free to hold off if you wanted to iterate further.

Again, a lightning fast clip that shows the flicker of the placeholder state, as this is just testing Gutenberg without the core patch, while dragging an HEIC image onto an image block. But this placeholder state's looking pretty good to me:

packages/block-library/src/image/editor.scss
noisysocks: Here's what increasing the opacity of the placeholder from 0.1 to 0.2 looks like against a few different backgrounds:

Screenshot 2024-07-19 at 15 09 46
Screenshot 2024-07-19 at 15 09 55
Screenshot 2024-07-19 at 15 10 05

That OK with you @jameskoster? If you want to play with it yourself an easy way to test is to make this change so that the placeholder and spinner will aways show for all image blocks:

```diff
diff --git a/packages/block-library/src/image/image.js b/packages/block-library/src/image/image.js
index 385a740e652..3417b1fc1a8 100644
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -827,7 +827,7 @@ export default function Image( {
 	const isRounded = attributes.className?.includes( 'is-style-rounded' );
 
 	let img =
-		temporaryURL && hasImageErrored ? (
+		true || ( temporaryURL && hasImageErrored ) ? (
 			// Show a placeholder during upload when the blob URL can't be loaded. This can
 			// happen when the user uploads a HEIC image in a browser that doesn't support them.
 			<Placeholder
```

Also happy to punt this to later.

jameskoster: Thanks for that tip :) 

Does the overlay background need to be `currentColor`? Since it's part of the UI I think it could use a UI color like `$gray-900`. Then with opacity something like .7 we get this:

Screenshot 2024-07-19 at 10 18 27

This ensures the spinner is visible, and has similar visual weight to the dropzone treatment. 

What do you think?

noisysocks: I like it but I'm scared that I'm hiding a UI change that needs some design attention in an otherwise innocuous PR 😅

I'll revert f5e8ac8, merge this, and then open a new PR with your suggestion.

noisysocks:
Thanks Andy. I'll revert https://github.com/WordPress/gutenberg/commit/f5e8ac8249aad43c083e0cb7d222432be8115e2b, merge this, and then open a new PR with @jameskoster's suggestion. Just so that design change gets a bit more attention.

noisysocks:
Safe to ignore failing performance tests as they passed on 9cf0967 which we're now back at on the count of reverting.

packages/block-library/src/image/editor.scss
noisysocks: https://github.com/WordPress/gutenberg/pull/63885

PR #63087, opened 2024-07-03
DataViews: rename `onSelectionChange` to `onChangeSelection`

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/55083

## What?

Renames `onSelectionChange` to `onChangeSelection`.

## Why?

For API coherence. We are using `onChange*` in a few properties: `onChangeView` for DataViews, `onChange` for the upcoming `DataForm` at https://github.com/WordPress/gutenberg/pull/63032.

See conversation at https://github.com/WordPress/gutenberg/pull/63032#discussion_r1663883147

## How?

- Renames the property internally across the dataviews package.
- Updates the consumers that were using it: page templates, page pages, and the experimental post list.

## Testing Instructions

Verify those pages still work as expected. Examples:

- Bulk selection.
- Selection is used in list view to update the preview.

PR #63648, opened 2024-07-17
DataViews: Remove redundant setSelection prop

youknowriad:
Follow-up to https://github.com/WordPress/gutenberg/pull/63087 

## What?

Just removes a redundant `setSelection` prop from the DataViews component. We can do the same thing in the `onChangeSelection` callback.

## Testing Instructions

Check that you can select items properly in the different dataviews.

### DIFF
--- a/packages/dataviews/src/dataviews.tsx
+++ b/packages/dataviews/src/dataviews.tsx
@@ -31,7 +31,7 @@ import type {
 	ViewBaseProps,
 	SupportedLayouts,
 } from './types';
-import type { SetSelection, SelectionOrUpdater } from './private-types';
+import type { SelectionOrUpdater } from './private-types';
 
 type ItemWithId = { id: string };
 
@@ -50,15 +50,12 @@ type DataViewsProps< Item > = {
 	};
 	defaultLayouts: SupportedLayouts;
 	selection?: string[];
-	setSelection?: SetSelection;
-	onChangeSelection?: ( items: Item[] ) => void;
+	onChangeSelection?: ( items: string[] ) => void;
 } & ( Item extends ItemWithId
 	? { getItemId?: ( item: Item ) => string }
 	: { getItemId: ( item: Item ) => string } );
 
 const defaultGetItemId = ( item: ItemWithId ) => item.id;
-
-const defaultOnChangeSelection = () => {};
 
 export default function DataViews< Item >( {
 	view,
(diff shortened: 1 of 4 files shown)


ellatrix:
For clarity, this would set an array of _items_ or an array of _ids_?

youknowriad:
An array of ids, I didn't change the type.

youknowriad:
Oh you're right I did change the types :P

packages/dataviews/src/dataviews.tsx
```diff
@@ -73,25 +70,23 @@ export default function DataViews< Item >( {
...
 			typeof value === 'function' ? value( selection ) : value;
-		onChangeSelection(
-			data.filter( ( item ) => newValue.includes( getItemId( item ) ) )
-		);
-		return setSelection( value );
+		if ( ! isUncontrolled ) {
+			setSelectionState( newValue );
+		}
+		if ( onChangeSelection ) {
+			onChangeSelection( newValue );
```
ellatrix, line 88: Can't we just do something like this now:

```js
const setSelectionWithChange = onChangeSelection || setSelectionState
```

youknowriad: no we can't, because sometimes you want the selection to be uncontrolled (you don't pass any selection prop) but you want to be notified about selection changes so you provide a `onChangeSelection` prop.

ellatrix: When do we want this? Can't we always force it to be controlled if you want to be notified?

packages/dataviews/src/dataviews.tsx
```diff
@@ -73,25 +70,23 @@ export default function DataViews< Item >( {
...
-		? setSelectionState
-		: setSelectionProperty;
 	const [ openedFilter, setOpenedFilter ] = useState< string | null >( null );
 
 	function setSelectionWithChange( value: SelectionOrUpdater ) {
 		const newValue =
 			typeof value === 'function' ? value( selection ) : value;
-		onChangeSelection(
-			data.filter( ( item ) => newValue.includes( getItemId( item ) ) )
-		);
```
ellatrix, line 93: Aren't we now passing the items? Previously we very extracting the ids and now we don't?

packages/dataviews/src/dataviews.tsx
youknowriad: For me, it's a good idea to support this, just like regular inputs or any raw form element in React. You can pass the value to make it controlled otherwise you can use `onChange` to be notified.

packages/dataviews/src/dataviews.tsx
youknowriad: No, it's the opposite, we're now passing "strings" while we previously were passing "items".

The extra logic here is to filter items that are in the selection but not in the data. The problem though is that this filtering doesn't really make sense when calling `onChange` since we're calling it from the "inside" meaning we know the ids are in the data already.

ellatrix: Oh right, I actually changed it internally to IDs before, I forgot :D

ellatrix (approved):

packages/dataviews/src/dataviews.tsx
ellatrix: Doesn't it log a warning in that case about an uncontrolled value?

youknowriad: No, it logs a warning only when you switch from controlled to uncontrolled (or the opposite)

Issue #63463, opened 2024-07-11
Vertical toolbar in “Zoom Out” position for non fullwidth blocks


stokesman:
### Description

The new vertical toolbar from #60123 is cool. I've noticed that on a theme with a dark background when a content block that isn’t full width is selected, the toolbar sits over the canvas and can be pretty much invisible. I would expect that the toolbar should always have good visibility regardless of the theme’s background.

### Step-by-step reproduction instructions

1. Select a theme with a dark background or change your theme’s background to a dark one
2. Open the Inserter, go to the Patterns tab and select a pattern category
3. Select a block that is not full-width
4. Notice the toolbar buttons are hard (maybe impossible) to see

### Screenshots, screen recording, code snippet

<img width="1080" alt="image" src="https://github.com/user-attachments/assets/d8cdb0a9-1b2c-463c-b755-ff862050d5c3">

### Environment info

- WP 6.5 / Gutenberg Trunk / Twenty Twenty-Four
- Chrome
- Desktop macOS

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

richtabor:
I removed the Needs Design label; this is a bug as the vertical toolbar should always reside outside the site contents. Instead it looks the the vertical toolbar is always to the left of the content, rather than always on the outside. 

cc @scruffian @MaggieCabrera

richtabor:
Ideally the toolbar is not reliant on the width of the block, but the canvas itself. Another alternative is to only have the toolbar render for fullwidth blocks (at least a temporary measure to try first). cc @MaggieCabrera

scruffian:
Fixed by #63650

PR #63650, opened 2024-07-17
Zoom out: hide vertical toolbar when block is not full width

MaggieCabrera:
## What?

Hide's the zoom out toolbar when the block is not full width
This is a tentative fix for https://github.com/WordPress/gutenberg/issues/63463 

## Why?

The zoom out toolbar makes more sense for section patterns rather than individual blocks. This is an easy fix that we can revisit if we feel like we need the toolbar in other cases too.

## How?

Adding a check for the alignment attribute

## Testing Instructions

1. Write a post and add some blocks to it
2. Open the pattern inserter, insert a full width pattern
3. click on the blocks in the canvas, only the full width pattern should display the toolbar

## Screenshots or screencast 

ScreenCaptureon2024-07-17at12-48-12-ezgif com-resize

### DIFF
--- a/packages/block-editor/src/components/block-tools/use-show-block-tools.js
+++ b/packages/block-editor/src/components/block-tools/use-show-block-tools.js
@@ -45,8 +45,13 @@ export function useShowBlockTools() {
 			editorMode === 'navigation';
 
 		const isZoomOut = editorMode === 'zoom-out';
+		const _showZoomOutToolbar =
+			isZoomOut &&
+			block?.attributes?.align === 'full' &&
+			! _showEmptyBlockSideInserter &&
+			! maybeShowBreadcrumb;
 		const _showBlockToolbarPopover =
-			! isZoomOut &&
+			! _showZoomOutToolbar &&
 			! getSettings().hasFixedToolbar &&
 			! _showEmptyBlockSideInserter &&
 			hasSelectedBlock &&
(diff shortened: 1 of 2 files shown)


scruffian:
I'm not sure if we should treat the toolbar differently to general block interations. I wonder if actually the fix here is to be more thoughtful about what level of blocks we allow to be selected when in zoom out mode

MaggieCabrera:
> I'm not sure if we should treat the toolbar differently to general block interations. I wonder if actually the fix here is to be more thoughtful about what level of blocks we allow to be selected when in zoom out mode

Mmmh, yeah, but if we don't allow selection for some then we won't have the inserters, won't we? we do want to keep those so you can insert in between blocks in a post that has, say, just paragraph blocks

scruffian:
In that case maybe we should show the normal toolbar on blocks that aren't full width?

MaggieCabrera:
> In that case maybe we should show the normal toolbar on blocks that aren't full width?

I implemented this and I like it. As long as we can get rid of the inserters, it will be a better experience. I had to make some changes on the CSS for zoom out because it was centering the content that was small and all the toolbars were being centered too. 

ScreenCaptureon2024-07-19at17-06-10-ezgif com-resize

scruffian:
There's something off with zoom out and full width blocks in the post editor
Screenshot 2024-07-23 at 10 50 12

scruffian (approved):
This is also an issue in trunk. I say we merge this.

PR #63652, opened 2024-07-17
Update: Pages: Trash view should default to table layout try 2.

jorgefilipecosta:
Part of https://github.com/WordPress/gutenberg/issues/63128.
This is the second attempt for https://github.com/WordPress/gutenberg/pull/63138, which was previously reverted at https://github.com/WordPress/gutenberg/pull/63481 due to causing a test failure on the trunk. The test failure did not occur on the branch itself.

The test failure was legitimate; the issue was that when switching from the "All pages" with table layout to the trash view, the table layout was maintained. However, when switching back to "All pages," the layout switched to list view instead of maintaining the table view.

This PR resolves that issue. Now, we have logic in place to ensure that when an automatic switch to table layout occurs because the draft view becomes active, the system will switch back to the previously active view instead of defaulting to the list view.

cc: @WordPress/gutenberg-design, I would appreciate, if someone could retest and ensure everything is working as expected. Thank you in advance 😊

## Testing Instructions
Open `/wp-admin/site-editor.php?postType=page`.
Switch to the trash view, and confirm that the table layout becomes active.
Change to the all pages view and ensure that the list layout becomes active again.
Next, open `wp-admin/site-editor.php?postType=page&layout=table`.
Switch to the trash view and verify that the table layout remains active.
Switch to the all pages view and confirm that the table layout continues to remain active.

### DIFF
--- a/packages/edit-site/src/components/sidebar-dataviews/index.js
+++ b/packages/edit-site/src/components/sidebar-dataviews/index.js
@@ -18,7 +18,7 @@ const { useLocation, useHistory } = unlock( routerPrivateApis );
 
 /**
  * Hook to switch to table layout when switching to the trash view.
- * When going out of the trash view, it switches back to the previous layout if 
+ * When going out of the trash view, it switches back to the previous layout if
  * there was an automatic switch to table layout.
  */
 function useSwitchToTableOnTrash() {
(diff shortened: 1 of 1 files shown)


jameskoster:
It's working for me, thanks for applying the same on the Posts data view 👍

PR #63653, opened 2024-07-17
Core Data: Support entity queries in the 'useResourcePermissions' hook

Mamaduka:
## What?

PR updates the `useResourcePermissions` hook to support new entity-aware syntax. I also replaced usage in the codebase.

## Why?
The new syntax should be preferred over direct REST API resource queries.

## Testing Instructions

1. Go to the Navigation block
2. Create a new item in the canvas and type in some random text to generate a new page.
3. It should be possible to create a page.
4. Do the same but using a Post Link block in the Nav.

Same.

## Screenshots or screencast 
CleanShot 2024-07-17 at 16 14 25


aicuaboom:
👾

packages/block-library/src/navigation-submenu/edit.js
```diff
@@ -499,7 +487,6 @@ export default function NavigationSubmenuEdit( {
 								}
 							} }
 							anchor={ popoverAnchor }
-							hasCreateSuggestion={ userCanCreate }
```
Mamaduka: The `LinkUI` component doesn't use this prop and makes its own permissions checks. Queries were affecting nothing.

packages/block-library/src/navigation-link/link-ui.js
```diff
@@ -147,15 +147,18 @@ function LinkUIBlockInserter( { clientId, onBack, onSelectBlock } ) {
...
+
 	const [ addingBlock, setAddingBlock ] = useState( false );
 	const [ focusAddBlockButton, setFocusAddBlockButton ] = useState( false );
 	const { saveEntityRecord } = useDispatch( coreStore );
-	const pagesPermissions = useResourcePermissions( 'pages' );
-	const postsPermissions = useResourcePermissions( 'posts' );
+	const permissions = useResourcePermissions( {
+		kind: 'postType',
+		name: postType,
+	} );
```
Mamaduka, line 159: Removes unnecessary checks for both entities, which later might have been disregarded. Reduces the possible `OPTIONS` request to one.

packages/edit-widgets/src/components/widget-areas-block-editor-provider/index.js
```diff
@@ -55,6 +51,11 @@ export default function WidgetAreasBlockEditorProvider( {
 			? getEntityRecord( 'root', 'site' )
 			: undefined;
 		return {
+			hasUploadPermissions:
+				canUser( 'create', {
+					kind: 'root',
+					name: 'media',
+				} ) ?? true,
```
Mamaduka, line 58: I simplified the check. It matches how the `Customize Widgets` and `Editor` packages are doing the same check.

https://github.com/WordPress/gutenberg/blob/cdf31e8677db66c021aaae0b02d759b99a05328b/packages/customize-widgets/src/components/sidebar-block-editor/index.js#L53-L57

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -110,14 +114,29 @@ type ResourcePermissionsResolution< IdType > = [
...
+			const hasId = isEntity ? !! resource.id : !! id;
 			const { canUser } = resolve( coreStore );
-			const create = canUser( 'create', resource );
-			if ( ! id ) {
+			const create = canUser(
+				'create',
+				isEntity
+					? { kind: resource.kind, name: resource.name }
+					: resource
+			);
```
Mamaduka, line 153: This is for backward compatibility. The `create` action was checked using the entity list endpoint even when `id` was provided.

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -110,14 +114,29 @@ type ResourcePermissionsResolution< IdType > = [
...
-	resource: string,
+	resource: string | EntityResource,
 	id?: IdType
 ): ResourcePermissionsResolution< IdType > {
+	// Serialize `resource` to a string that can be safely used as a React dep.
+	// We can't just pass `resource` as one of the deps, because if it is passed
+	// as an object literal, then it will be a different object on each call even
+	// if the values remain the same.
+	const resourceAsString =
+		typeof resource === 'object' ? addQueryArgs( '', resource ) : resource;
```
Mamaduka: I don't like this hack, but it's the downside of using object arguments with hooks.

Mamaduka:
This is ready for the review. cc @WordPress/gutenberg-core

packages/block-library/src/navigation-submenu/edit.js
tyxla: I think this cleanup can be done in a separate PR to keep this one smaller.

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -110,14 +114,29 @@ type ResourcePermissionsResolution< IdType > = [
  * @template IdType
  */
 export default function useResourcePermissions< IdType = void >(
-	resource: string,
+	resource: string | EntityResource,
 	id?: IdType
```
tyxla, line 129: Should types forbid this second argument if we're providing an `EntityResource` object as `resource`?

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -110,14 +114,29 @@ type ResourcePermissionsResolution< IdType > = [
...
+	// We can't just pass `resource` as one of the deps, because if it is passed
+	// as an object literal, then it will be a different object on each call even
+	// if the values remain the same.
+	const resourceAsString =
+		typeof resource === 'object' ? addQueryArgs( '', resource ) : resource;
+
 	return useQuerySelect(
 		( resolve ) => {
+			const isEntity = typeof resource === 'object';
+			const hasId = isEntity ? !! resource.id : !! id;
```
tyxla, line 146: Feels confusing that we're allowing an ID to be provided as the second argument when working with a `resource`. This would allow for a hybrid use, like this: 

```
useResourcePermissions( { kind: 'postType', name: 'post' }, 123 )
```

which feels confusing, since we also allow:

```
useResourcePermissions( { kind: 'postType', name: 'post', id: 123 } )
```

since we're defining this behavior right here and right now, does it make sense to remove the ambiguity?

packages/block-library/src/navigation-link/link-ui.js
tyxla: Nice one 👍 I appreciate the resulting cleanup!

packages/block-library/src/navigation/test/use-navigation-menu.js
```diff
@@ -63,37 +63,76 @@ function resolveRecords( registry, menus ) {
 
 function resolveReadPermission( registry, allowed ) {
 	const dispatch = registry.dispatch( coreStore );
-	dispatch.receiveUserPermission( 'create/navigation', allowed );
-	dispatch.startResolution( 'canUser', [ 'read', 'navigation' ] );
-	dispatch.finishResolution( 'canUser', [ 'read', 'navigation' ] );
+	dispatch.receiveUserPermission( 'read/postType/wp_navigation', allowed );
+	dispatch.startResolution( 'canUser', [
+		'read',
+		{ kind: 'postType', name: 'wp_navigation', id: undefined },
```
tyxla: Why do we have `id: undefined`? Shouldn't it just be omitted?

packages/block-library/src/navigation/test/use-navigation-menu.js
```diff
@@ -63,37 +63,76 @@ function resolveRecords( registry, menus ) {
...
-	dispatch.startResolution( 'canUser', [ 'read', 'navigation', ref ] );
-	dispatch.finishResolution( 'canUser', [ 'read', 'navigation', ref ] );
+	dispatch.receiveUserPermission(
+		`read/postType/wp_navigation/${ ref }`,
+		allowed
+	);
+	dispatch.startResolution( 'canUser', [
+		'read',
+		{ kind: 'postType', name: 'wp_navigation', id: ref },
+	] );
```
tyxla: I see this resolution argument repetition often in test code. Could potentially be moved to constants and reused for both `startResolution` and `finishResolution`.

packages/core-data/README.md
```diff
@@ -1181,8 +1188,8 @@ the store state using `canUser()`, or resolved if missing.
 
 _Parameters_
 
--   _resource_ `string`: The resource in question, e.g. media.
--   _id_ `IdType`: ID of a specific resource entry, if needed, e.g. 10.
+-   _resource_ `string | EntityResource`: Entity resource to check. Accepts entity object `{ kind: 'root', name: 'media', id: 1 }` or REST base as a string - `media`.
+-   _id_ `IdType`: Optional ID of the rest resource to check. e.g. 10.
```
tyxla: If we end up changing the logic and making this a noop if `resource` is an entity object, let's make sure to reflect that here.

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -41,20 +42,23 @@ type ResourcePermissionsResolution< IdType > = [
...
 /**
  * Resolves resource permissions.
  *
  * @since 6.1.0 Introduced in WordPress core.
  *
- * @param    resource The resource in question, e.g. media.
- * @param    id       ID of a specific resource entry, if needed, e.g. 10.
+ * @param    resource Entity resource to check. Accepts entity object `{ kind: 'root', name: 'media', id: 1 }`
+ *                    or REST base as a string - `media`.
+ * @param    id       Optional ID of the rest resource to check. e.g. 10.
```
tyxla: On a second thought, if we decide to keep this in the case of an entity object passed as `resource`, let's make sure to document that the `id` in `resource` takes precedence over the `id` parameter.

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -41,20 +42,23 @@ type ResourcePermissionsResolution< IdType > = [
...
  *
  * @example
  * ```js
  * import { useResourcePermissions } from '@wordpress/core-data';
  *
  * function PagesList() {
- *   const { canCreate, isResolving } = useResourcePermissions( 'pages' );
+ *   const { canCreate, isResolving } = useResourcePermissions( { kind: 'postType', name: 'page' } );
```
tyxla, line 72: We might still want to keep some examples with the legacy syntax, unless we want to actively discourage / deprecate it.

packages/core-data/src/hooks/use-resource-permissions.ts
tyxla: It could be a bit cleaner to use `buildQueryString` since we don't care about unserializing it. Alternatively, `JSON.stringify()` could potentially be faster for a small object like this - I think it's used for a cache key and a dependency here and there for similar reasons.

packages/core-data/src/hooks/use-resource-permissions.ts
Mamaduka: I am happy to do so, but my TS knowledge is lacking to do it :/

packages/block-library/src/navigation/test/use-navigation-menu.js
Mamaduka: While `selectorArgsToStateKey` removes trailing `undefined` arguments, it doesn't do it inside the arguments themselves.

Omitting `id` results `canUser` not marked as resolved.

packages/block-library/src/navigation-submenu/edit.js
Mamaduka: Extracted in #63720.

packages/core-data/src/hooks/use-resource-permissions.ts
Mamaduka: Yes, I think we want to discourage REST API resource-based queries.

packages/core-data/src/hooks/use-resource-permissions.ts
tyxla: I believe it should be doable with some [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads):

```typescript
function useResourcePermissions< IdType = void >(
	resource: string,
	id?: IdType
): ResourcePermissionsResolution< IdType >;

function useResourcePermissions< IdType = void >(
	resource: EntityResource,
	id?: never
): ResourcePermissionsResolution< IdType >;

function useResourcePermissions( resource, id ) {
	// ...
}

export default useResourcePermissions;
```

then, these usages should be valid:

```typescript
export function Foo( resource: string, id: number ) {
	const x = useResourcePermissions( resource, id );
	return x;
}

export function Bar( resource: EntityResource ) {
	const x = useResourcePermissions( resource );
	return x;
}
```

and this usage should be invalid, erroring that there are no overloads supporting that call:

```typescript
export function Baz( resource: EntityResource, id: number ) {
	const x = useResourcePermissions( resource, id );
	return x;
}
```

Note that in this (untested) example, we're exporting `useResourcePermissions` separately to prevent a `TS2383` error. An alternative solution should be to export all function overload types. 

Let me know if that works!

packages/core-data/src/hooks/use-resource-permissions.ts
Mamaduka: Adde note in [90a0d22](https://github.com/WordPress/gutenberg/pull/63653/commits/90a0d223d1d7d023100e5a70241bc9b243a8188a).

packages/core-data/src/hooks/use-resource-permissions.ts
Mamaduka: Thanks, @tyxla!

Added function overloads in [90a0d22](https://github.com/WordPress/gutenberg/pull/63653/commits/90a0d223d1d7d023100e5a70241bc9b243a8188a), and they seem to work correctly.

packages/core-data/src/hooks/use-resource-permissions.ts
Mamaduka: The hybrid signature won't work. In that case, the `hasId` would be false; the hook would check general permissions for the REST API and return early.

tyxla: Thanks for confirming, but that's even worse. In that case, we need to make sure this behavior is well-documented, and we should probably throw an error if someone calls it with a resource object and a separate ID argument. Because, you know, TS won't be enough sometimes - some projects and developers just don't use it (or abuse it) in their dev environments.

Mamaduka: Throwing an error in a hook might be a bit risky. Depending on where the hook is called in the tree, the closest error boundary might be the one in the app root, preventing the whole app from loading.

I think logging a `warning` is probably the safest solution here.

tyxla: Logging a `warning` makes sense 👍 I definitely didn't mean throwing a literal error.

Mamaduka: I added the warning using the `@wordpress/warning` package, which seems to be [the new preferred method](https://github.com/WordPress/gutenberg/pull/63610) for such cases.

cc @gziolo 

## Screenshot
CleanShot 2024-07-23 at 13 16 49

gziolo: I’m still processing the usage of `warning` util as I started to have second thoughts of how we can catch these warnings on production. Do we have to? How folks debug the cases where plugins conflict between each other? My reasoning was that if the app continues to work then we don’t need to add noise to the Browser Console when the app runs on production. I’m happy to discuss it further but for block registration it was a clear path forward as it’s mostly targeting the block author when developing the block.

tyxla: In this case, the warning is useful to the developer who's trying to use the API, since it's clear that they're using it in an unexpected way (almost like a front-end counterpart of [`_doing_it_wrong()`](https://developer.wordpress.org/reference/functions/_doing_it_wrong/). To me, it definitely doesn't make sense to surface that in production, it should only be seen in development mode.

Mamaduka: This warning also has the same target audience - developers. The warnings that aren't actionable by the end user shouldn't be displayed in production environments.

gziolo: I think we reached a consensus here. I appreciate the parallel to `_doint_it_wrong()`. That speaks well to me and maybe we should alias it this way one day to offer more clarity.

Mamaduka:
@tyxla, do you think this is in an OK shape to be merged?

packages/core-data/README.md
```diff
@@ -1222,8 +1229,8 @@ the store state using `canUser()`, or resolved if missing.
 
 _Parameters_
 
--   _resource_ `string`: The resource in question, e.g. media.
--   _id_ `IdType`: ID of a specific resource entry, if needed, e.g. 10.
+-   _resource_ `string | EntityResource`: Entity resource to check. Accepts entity object `{ kind: 'root', name: 'media', id: 1 }` or REST base as a string - `media`.
+-   _id_ `IdType`: Optional ID of the resource to check, e.g. 10. Note: The argument will be ignored when using an entity object as a resource to check permissions.
```
tyxla: Maybe we can make it clearer that the usage of `id` with a `resource` object is incorrect and will yield a warning. The fact that it is ignored doesn't seem to sufficiently emphasize that fact IMHO. 

```suggestion
-   _id_ `IdType`: Optional ID of the resource to check, e.g. 10. Note: This argument is discouraged when using an entity object as a resource to check permissions and will be ignored.
```

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -41,20 +42,34 @@ type ResourcePermissionsResolution< IdType > = [
...
 /**
  * Resolves resource permissions.
  *
  * @since 6.1.0 Introduced in WordPress core.
  *
- * @param    resource The resource in question, e.g. media.
- * @param    id       ID of a specific resource entry, if needed, e.g. 10.
+ * @param    resource Entity resource to check. Accepts entity object `{ kind: 'root', name: 'media', id: 1 }`
+ *                    or REST base as a string - `media`.
+ * @param    id       Optional ID of the resource to check, e.g. 10. Note: The argument will be ignored
```
tyxla: Same as above.

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -109,15 +124,36 @@ type ResourcePermissionsResolution< IdType > = [
...
 	id?: IdType
 ): ResourcePermissionsResolution< IdType > {
+	// Serialize `resource` to a string that can be safely used as a React dep.
+	// We can't just pass `resource` as one of the deps, because if it is passed
+	// as an object literal, then it will be a different object on each call even
+	// if the values remain the same.
+	const resourceAsString =
+		typeof resource === 'object' ? JSON.stringify( resource ) : resource;
+
+	const isEntity = typeof resource === 'object';
```
tyxla: We can just reuse the `isEntity` constant:

```suggestion
	const isEntity = typeof resource === 'object';
	const resourceAsString = isEntity ? JSON.stringify( resource ) : resource;
```

packages/core-data/src/hooks/use-resource-permissions.ts
```diff
@@ -109,15 +124,36 @@ type ResourcePermissionsResolution< IdType > = [
...
 	return useQuerySelect(
 		( resolve ) => {
+			const hasId = isEntity ? !! resource.id : !! id;
 			const { canUser } = resolve( coreStore );
-			const create = canUser( 'create', resource );
-			if ( ! id ) {
+			const create = canUser(
+				'create',
+				isEntity
+					? { kind: resource.kind, name: resource.name }
```
tyxla, line 151: Is there a good reason to not trust passing the `resource` object here directly?

tyxla (approved):
I've tested this in the post editor, site editor, and widgets editor and everything worked well! ✅ 

The code also looks good 👍 I've only left a few minor suggestions that can be addressed before shipping.

Thanks @Mamaduka 🚀

packages/core-data/src/hooks/use-resource-permissions.ts
Mamaduka: The `resource` object can have an `id` and the hook used to check `create` actions without IDs. So, one reason is backward compatibility.

Usually, create permissions are checked using the list endpoint `OPTIONS v2/pages`.

Mamaduka:
Thanks for the review and feedback, @tyxla!

PR #63654, opened 2024-07-17
Separator, Code: don't require Enter for shortcut

ellatrix:
## What?

Introduces a new transform/shortcut type `input`, which transform text into a block as soon as the sequence is complete (instead of waiting for `Enter`).

Question: in case of the separator, would it make more sense to also add an empty paragraph after it?

## Why?

Because currently the `enter` shortcut is not discoverable enough, and there's not really any harm in transforming sooner. It's still possible to undo the transform.

Type `---` or ` ``` ` in a paragraph block, which should transform it to a Separator or Code block respectively.

### DIFF
--- a/packages/block-editor/src/components/rich-text/event-listeners/input-rules.js
+++ b/packages/block-editor/src/components/rich-text/event-listeners/input-rules.js
@@ -99,6 +99,7 @@ export default ( props ) => ( element ) => {
 			__unstableAllowPrefixTransformations,
 			formatTypes,
 			registry,
+			onReplace,
 		} = props.current;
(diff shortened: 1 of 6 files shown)


jarekmorawski:
It's testing well for me in both editors. Nice work! 🙌 

### Site editor

https://github.com/user-attachments/assets/0f0e9610-6e1d-4739-86ff-41754996cbcf


### Post editor

https://github.com/user-attachments/assets/c1187f56-10e8-49d1-8b24-f74a7b5d3aae

erikiva (approved):

PR #63656, opened 2024-07-17
Make useStyleOverride public

### LINKED ISSUE #61232: Expose `@emotion/styled` to extenders
Several of the ´@wordpress/´ packages use `@emotion/styled` internally as their CSS in JS solution. 

When 3rd party block component packages then try to also use the same library first of all it causes a lot of duplicate code to be loaded slowing down the editor. But on top of that emotion also adds the following warning in the browser console:

> [!Warning]
> ```
> You are loading @emotion/react when it is already loaded. Running multiple instances may cause problems. This can happen if multiple versions are used, or if multiple builds of the same version are used.`
> ```

#### Proposed solution

It would be geat to either be able to import the functions used by the components package like `useCx` `styled` etc. in external packages. 

Alternatively maybe `@emotion/styled` & `@emotion/react` could get bundled as their own packages in WordPress core similar to how we have `react` bundled in core. 

#####

In the end I think the ideal solution would be to not directly expose `emotion` or any other CSS in JS library but instead expose the abstracted layer that allows anyone to use the library without having to worry about whether a component gets rendered inside an iframe or not etc. Just making it easy to use the same component styling system that core already is using internally :)

ellatrix:
## What?

Fixes #61232. Exposes `useStyleOverrides` so styles can be added to the canvas by blocks.

## Why?

See https://github.com/WordPress/gutenberg/issues/61232#issuecomment-2096616900. It's handy to have an API for dynamically adding editor styles.

## How?

Exposed the hooks while keeping private props internal for now. Only `id` and `css` are exposed.

## Testing Instructions

`useStyleOverride( { css: 'p{color:red}' } )` in a block. Note that most often this is used in combination with a generated class and css specifically targeting that class.

### DIFF
--- a/packages/block-editor/src/hooks/duotone.js
+++ b/packages/block-editor/src/hooks/duotone.js
@@ -32,7 +32,7 @@ import {
 } from '../components/duotone/utils';
 import { getBlockCSSSelector } from '../components/global-styles/get-block-css-selector';
 import { scopeSelector } from '../components/global-styles/utils';
-import { useBlockSettings, useStyleOverride } from './utils';
+import { useBlockSettings, usePrivateStyleOverride } from './utils';
 import { default as StylesFiltersPanel } from '../components/global-styles/filters-panel';
 import { useBlockEditingMode } from '../components/block-editing-mode';
 import { useBlockElement } from '../components/block-list/use-block-props/use-block-refs';
@@ -265,7 +265,7 @@ function useDuotoneStyles( {
 
 	const isValidFilter = Array.isArray( colors ) || colors === 'unset';
 
-	useStyleOverride(
+	usePrivateStyleOverride(
 		isValidFilter
 			? {
 					css:
@@ -276,7 +276,7 @@ function useDuotoneStyles( {
 			  }
 			: undefined
 	);
-	useStyleOverride(
+	usePrivateStyleOverride(
 		isValidFilter
 			? {
 					assets:
--- a/packages/block-editor/src/hooks/utils.js
+++ b/packages/block-editor/src/hooks/utils.js
@@ -135,7 +135,20 @@ export function shouldSkipSerialization(
 
 const pendingStyleOverrides = new WeakMap();
 
-export function useStyleOverride( {
+/**
+ * Override a block editor settings style. Leave the ID blank to create a new
+ * style.
+ *
+ * @param {Object} $0     Named parameters.
+ * @param {string} $0.id  Id of the style override, leave blank to create a new
+ *                        style.
+ * @param {string} $0.css CSS to apply.
+ */
+export function useStyleOverride( { id, css } ) {
+	return usePrivateStyleOverride( { id, css } );
+}
+
+export function usePrivateStyleOverride( {
 	id,
 	css,
 	assets,
--- a/packages/block-editor/src/hooks/utils.js
+++ b/packages/block-editor/src/hooks/utils.js
@@ -139,10 +139,10 @@ const pendingStyleOverrides = new WeakMap();
  * Override a block editor settings style. Leave the ID blank to create a new
  * style.
  *
- * @param {Object} $0     Named parameters.
- * @param {string} $0.id  Id of the style override, leave blank to create a new
- *                        style.
- * @param {string} $0.css CSS to apply.
+ * @param {Object}  override     Override object.
+ * @param {?string} override.id  Id of the style override, leave blank to create
+ *                               a new style.
+ * @param {string}  override.css CSS to apply.
  */
 export function useStyleOverride( { id, css } ) {
 	return usePrivateStyleOverride( { id, css } );
--- a/packages/block-editor/README.md
+++ b/packages/block-editor/README.md
@@ -1035,9 +1035,9 @@ Override a block editor settings style. Leave the ID blank to create a new style
 
 _Parameters_
 
--   _$0_ `Object`: Named parameters.
--   _$0.id_ `string`: Id of the style override, leave blank to create a new style.
--   _$0.css_ `string`: CSS to apply.
+-   _override_ `Object`: Override object.
+-   _override.id_ `?string`: Id of the style override, leave blank to create a new style.
+-   _override.css_ `string`: CSS to apply.
 
 ### useZoomOut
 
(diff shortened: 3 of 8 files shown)


packages/block-editor/src/hooks/utils.js
```diff
@@ -135,7 +135,20 @@ export function shouldSkipSerialization(
...
 
-export function useStyleOverride( {
+/**
+ * Override a block editor settings style. Leave the ID blank to create a new
+ * style.
+ *
+ * @param {Object} $0     Named parameters.
+ * @param {string} $0.id  Id of the style override, leave blank to create a new
+ *                        style.
+ * @param {string} $0.css CSS to apply.
```
andrewserong: Tiny nit: should this be an optional `string` param?

```suggestion
 * @param {Object}  $0     Named parameters.
 * @param {?string} $0.id  Id of the style override, leave blank to create a new
 *                        style.
 * @param {string}  $0.css CSS to apply.
```

andrewserong (commented):
I like the idea of making this public — it has a good and simply interface (optional id + CSS string), and I can imagine it being useful for custom block plugins as in the linked issue.

A general question: are there any performance concerns if lots of custom blocks wind up using this? I imagine not since it's already in use for the block supports. But just in case, I'll add a ping to @WordPress/gutenberg-core for visibility 🙂

packages/block-editor/src/hooks/duotone.js
```diff
@@ -265,7 +265,7 @@ function useDuotoneStyles( {
 
 	const isValidFilter = Array.isArray( colors ) || colors === 'unset';
 
-	useStyleOverride(
+	usePrivateStyleOverride(
```
andrewserong, line 268: Tiny nit about consistency:

~I see Duotone and block style variations have been switched to use `usePrivateStyleOverride` instead of `useStyleOverride`. Is there a reason to switch these internal-to-the-block-editor instances to use the private one? If so, would it be worth updating the code in `layout-child.js`, `layout.js`, `position.js` and `style.js`, too? If not, then I was wondering if it would be better for these ones to use `useStyleOverride` instead of the private hook 🤔~

Edit: ignore me, sorry! I only just noticed the `__unstableType: 'presets'` line here 🤦 

I see now why a couple of these use the private version and the others continue to use the public one 👍

andrewserong: All good, I typed before thinking that comment through! 😄

aaronrobertshaw: I believe the block style variations and duotone use the private version as they rely on the restricted parameters we're deliberately omitting from the public `useStyleOverride` e.g. `clientId`, `__unstableType` etc.

andrewserong: Thanks for confirming!

packages/block-editor/src/hooks/utils.js
aaronrobertshaw: If we're flagging `id` as optional here, we should probably also then do the same in the [readme](https://github.com/WordPress/gutenberg/pull/63656/files#diff-dd80d95d10df70d186b67b89836b923443e04bb1fcb0f3348773cf69511b5b53R1043).

aaronrobertshaw (approved):
I like the idea of making `useStyleOverride` public. Bonus points for the private version to restrict some of the accepted params ✨ 

✅ Adding custom `useStyleOverride` to block works
✅ Duotone and Block Style Variations work still using private version
✅ Gallery gap styles work
✅ General layout styles continue to work

Other than the docblock/readme nit, mentioned in other comments, this PR LGTM 👍

packages/block-editor/src/hooks/utils.js
Mamaduka: I think JSDocs also supports a better style than `$0` for object params.

ellatrix: @Mamaduka Curious what you're referring to. Do you mean a separate type definition?

Mamaduka:
```
 * @param {Object} override
 * @param {string} override.id
 * @param {string} override.css
```

DaniGuardiola (commented):
I won't block this, but I don't think this is the right decision because it will make emotion an irrevocable part of WordPress forever. As @mirka said [here](https://github.com/WordPress/gutenberg/issues/61232#issuecomment-2088300222), we're planning to move away from Emotion, as many are for important reasons like performance, bundle size, and incompatibility with modern React features. [The MUI team wrote about this.
](https://mui.com/blog/introducing-pigment-css/#traditional-css-in-js-is-not-enough)

In the near term, there are options that address these same challenges that are much more aligned with the web platform and the future of the fundamental tools we rely on, such as React. Some examples: the new React 19 style tag features, CSS layers, `:where`...

So I would recommend not shipping this, at least not for now, until we have more clarity on the post-Emotion future in the components library and other places in Gutenberg.

cc @WordPress/gutenberg-components

aaronrobertshaw:
@DaniGuardiola I might be missing something but I don't think making `useStyleOverride` public ties WordPress to Emotion at all. 

The style overrides added through this hook are stored in a registry and later retrieved and [rendered within `<style>` tags](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/editor-styles/index.js?plain=1#L104-L106). I don't see any use of Emotion there.

DaniGuardiola:
@aaronrobertshaw it's very likely that I'm confused if that's the case. I didn't look too deep into the code, so I only have the context of the original issue which was about exposing some emotion utilities.

Happy to have been proven wrong!

ellatrix:
@DaniGuardiola This is the alternative to emotion as discussion on the issue :)

fabiankaegy:
Hey @ellatrix 👋 

Thanks in advance :)

ellatrix:
I don't have much writing inspiration, something as simple as: `wp.blockEditor.useStyleOverride` is a new editor API that can be used to add content styling (renders insider the content iframe). And link the readme.

fabiankaegy:
@ellatrix i think what I personally was hoping to have included here would be a very small example of how you would actually pass on styles etc.

ellatrix:
@fabiankaegy Sorry, there's one in the description: `useStyleOverride( { css: 'p{color:red}' } )`

fabiankaegy:
Perfect thanks :)

PR #63658, opened 2024-07-17
Add post status icon in post summary

ntsekouras:
## What?
Part of the [discussion](https://github.com/WordPress/gutenberg/issues/63387#issuecomment-2230573727) in this issue to make the post status in post summary a bit more prominent.

This PR adds an `icon` next to the post status.

## Testing Instructions
1. Visual change, so everything should work as before.

## Screenshots or screencast 
Screenshot 2024-07-17 at 3 15 26 PM

### DIFF
--- a/packages/editor/src/components/post-status/index.js
+++ b/packages/editor/src/components/post-status/index.js
@@ -16,6 +16,13 @@ import { useState, useMemo } from '@wordpress/element';
 import { store as coreStore } from '@wordpress/core-data';
 import { __experimentalInspectorPopoverHeader as InspectorPopoverHeader } from '@wordpress/block-editor';
 import { useInstanceId } from '@wordpress/compose';
+import {
+	drafts,
+	published,
+	scheduled,
+	pending,
+	notAllowed,
+} from '@wordpress/icons';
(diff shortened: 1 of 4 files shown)


packages/editor/src/components/post-status/index.js
```diff
@@ -200,13 +207,15 @@ export default function PostStatus() {
 							variant="tertiary"
 							size="compact"
 							onClick={ onToggle }
+							icon={ postStatusesInfo[ status ]?.icon }
+							iconSize="20"
```
ntsekouras: Without the `iconSize`, the row get a bit bigger than the 32px, which is like a `min-height` for rows. Any better alternatives are welcome.

jameskoster: It's because `.editor-post-panel__row-control .components-button` was overriding the `height` value. Removing that fixed it, but revealed that the 'Parent' button was not the correct size, so I fixed that too.

jameskoster (approved):
I think this helps give prominence to status which is an important detail. The same icon is used in data views, so this change will increase familiarity.

PR #63661, opened 2024-07-17
Scripts: Remove now-obsolete `getRenderPropPaths()`

ockham:
## What?
Remove function `getRenderPropPaths()` from the `scripts` package.

## Why?
As of #63098, it is no longer used anywhere in Core. It has also never been part of the package's public interface.

## How?
My deleting it 🤷‍♂️ 

## Testing Instructions
Verify that GB still builds and runs as before.
Try `grep`ping the function and verify that it has been completely removed.


gziolo (approved):
Thank you for the quick follow-up.

Issue #52466, opened 2023-07-10
Permalink URL: make long permalinks more readable


afercia:
### Description

Long URLs are hardly readable when split in multiple lines. Maybe it's one of the _less_ readale things. The Edit permalink UI doesn't do a great job in making long URLs more readable for a couple reasons:

1. The current permalink value is shown in the Settings panel, which has a limited width of only 280 pixels.
2. Even when editing the permalink, the popover that shows the editing UI has a limited width.

In both cases, long URLs are split into multiple lines. Screenshot:

<img width="465" alt="Screenshot 2023-07-10 at 11 32 00" src="https://github.com/WordPress/gutenberg/assets/1682452/00d8418f-c154-4d9f-8d93-4d53bc70d2a0">

I feel this is a problem that is still unsolved in the editor. The readability is not better than how it was in the old classic editor.

Regarding the first point, I'd like to avoid to reinstate a long discussion about the 'sidebar' design paradigm. To me, it's an insufficient paradigm that comes with many accessibility problems and it's just not aligned with the need of modern applications. I've remarked that in several previous issues but, fair enough, let's ignore the sidebar issue for now.

 I'd like to focus on the popover instead.

Is there any reason why the Permalink popover must have a limited width?
I do realize that for visual consistency, popovers should have more or less the same width _when it makes sense_. This specific popover needs to show long text though, and I'd think providing better readability is better than visual consistency.

When there is available space, this popover should be larger, to show the Permalink URL in a single line (at least, till a certain length). Screenshot:

<img width="603" alt="Screenshot 2023-07-10 at 11 36 06" src="https://github.com/WordPress/gutenberg/assets/1682452/94bef0e4-b358-40fe-9eb6-1e9706047cbc">

Even better:
In the classic editor, when the last part of the Permalink (usually the post name) is longer than 34 characters, it gets 'abridged' and split in [two parts of maximum 16 characters with an ellipsis in between](https://github.com/WordPress/wordpress-develop/blob/03ff5a4f225927447b25591b4f18aa68f30b35ed/src/wp-admin/includes/post.php#L1562-L1568):

<img width="684" alt="Screenshot 2023-07-10 at 11 38 16" src="https://github.com/WordPress/gutenberg/assets/1682452/da130bdf-c3af-464b-b5ad-46618fb0cd8b">

The equivalent of this in Gutenberg would be something like the UI in the screenshot below:

[...]

PR #21099, opened 2020-03-24
Remove the post permalink UI from the post title

youknowriad:
closes #5494 

This PR removes the Post Permalink UI from the Post Title as suggested in #5494 
At the same time, it simplifies the DOM and styles for the post title a lot. 

**Notes**

In 2019, you'll notice a bug in the PostTitle (not centered properly) but I believe this is because the editor styles are targetting these internal classNames while ideally they shouldn't.  cc @kjellr 

I tried keeping the old classNames to reduce the scope of the breakage and and added "Needs dev note".

Issue #12033, opened 2018-11-18
Sidebar permalink panel: improve the URL field label


afercia:
Splitting this out from #11874.

I'd tend to think the input field label `URL` is a bit confusing, as users might think they have to fill the field with a complete URL starting with `http(s)`.

<img width="302" alt="screenshot 2018-11-18 at 14 27 02" src="https://user-images.githubusercontent.com/1682452/48673098-31c08380-eb3e-11e8-8521-af490802d42c.png">

I've read https://github.com/WordPress/gutenberg/issues/11858#issuecomment-438921521 and it makes perfectly sense to avoid technical terms. However, the label could be improved a bit. Also noting Squarespace does use some form of "URL" label but it also expands it adding "slug":

<img width="552" alt="screenshot 2018-11-18 at 13 55 29" src="https://user-images.githubusercontent.com/1682452/48672843-36d00380-eb3b-11e8-8780-4f569a7c3420.png">

Other mentioned platforms use some form of "URL" as well, but at least they make it visually clear it's about the last part of the URL:

<img width="481" alt="screenshot 2018-11-18 at 15 07 48" src="https://user-images.githubusercontent.com/1682452/48673729-c201c700-eb44-11e8-8a54-ee6197c15a12.png">

<img width="241" alt="screenshot 2018-11-18 at 15 12 35" src="https://user-images.githubusercontent.com/1682452/48673730-c201c700-eb44-11e8-9545-478974bac201.png">

Personally, I have no strong opinions on which the best wording would be, but I'd suggest to consider a more meaningful one.

afercia:
Not sure if it can help, but worth considering the explanation of what a "permalink" is in WordPress in the Help page linked from the Classic Editor help tab:
https://codex.wordpress.org/Posts_Add_New_Screen

timwright12:
@afercia I updated it to "URL Slug" just so we have an open PR. Happy to update it, but I think that actually works pretty well.

Issue #12031, opened 2018-11-18
Sidebar permalink panel: consider to always render the panel


afercia:
Splitting this out from #11874.

When creating a new post, the new permalink panel introduced in #11874 is not rendered yet in the sidebar.

![screenshot 2018-11-18 at 14 58 32](https://user-images.githubusercontent.com/1682452/48673546-b1e8e800-eb42-11e8-904b-c6772ce8efbf.png)

The panel appears only after the first auto-save or after users click on "Save draft" or "Publish".

![screenshot 2018-11-18 at 14 59 11](https://user-images.githubusercontent.com/1682452/48673549-b6ad9c00-eb42-11e8-99f3-3b711014b001.png)

This could be confusing for users who are just exploring the new post screen and wondering where the permalink setting is. It could be even more confusing for screen reader users, as screen readers have special tools to explore a web page (e.g. list all the headings, list all the form controls, list all the links, etc.) and they won't find anything related to the permalink setting. Same potential confusion applies to low vision users, screen magnifier users, users with cognitive impairments, etc.

I'd suggest to consider to always show the panel in the sidebar with some meaningful message when a slug doesn't exist yet.

afercia:
Also, if I remember correctly, it was decided to remove the permalink UI in the post title. Not sure it's already tracked down in some other issue. /Cc @youknowriad

youknowriad:
Removing the permalink UI in the post title had my preference but I think there were some hesitation. I'd love thoughts from others @jasmussen @karmatosed @mapk @chrisvanpatten @earnjam

afercia:
I'd vote for removing it. It has serious accessibility issues. I'd say also usability and discoverability aren't ideal.

afercia:
Worth reminding this issue with more details about the a11y issues: https://github.com/WordPress/gutenberg/issues/5494

It was closed only because there's now the new panel. However, the old UI is still not accessible.

> We didn't remove this UI yet because we're passed UI freeze but it should be considered deprecated.

jasmussen:
Considering the flow that a casual user might want to dismiss the sidebar and never open it again, it would be nice if the permalink was editable in the pre-publish flow (i.e. press Publish... and the permalink could be a panel there). If we had that flow in place then it would be harder to find good arguments to keeping the current in-canvas interface in place.

earnjam:
#12009 has ability to edit at any time and would resolve this issue. It has permalink sidebar panel visible but not expanded on initial load. 

[...]

PR #63669, opened 2024-07-17
Improve PostURL terminology and accessibility

### LINKED ISSUE #61196: Restore the terms 'Permalink' and 'Slug' in the PostURL user interface

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/61196

## What?

This PR proposes to restore the terminology WordPress always used when editing a post 'link':
- Slug: the last part of the URL
- Permalink: the permanent link to the post

The current UI allows users to edit _the last part of the link_ and calls it 'link'. That's not what usees are editing. Actually, they are editing the _slug_.

On top of that, fixes some accessibility and usability issues:
- An URL isn't a meaningful description to be associated to the input field with aria-describedby. The explanatory text before the input should be used as a description instead,
- When the UI shows the link of the Front Page, there is no need to have a toggle button that opens a popover only to show again the non-editable link that is already shown in the Inspector panel. See the 'before' screenshot. In this case, the popover should not be used in the first place. The panel already shows the link.
- Sometimes, when possible, please let's use paragraph elements to wrap text. 

## Why?

For better usability and accessibility.

## How?

- Changes the term 'Link' back to 'Slug' to indicate the last part of a post URL.
- Adds back the terme 'Permalink' to indicate the full permanent link of a post.
- Keeps the term 'Link' only when the UI shows the Front Page full link.
- Associates the explaantory text as the accessible description of the input field.

## Testing Instructions

- Edit a post.
- Observe in the Inspector post panel the row to edit the slug is now called 'Slug;.
- Click the toggle button to edit the Slug.
- Observe the popover title is now 'Slug'.
- Observe the text 'Customize the last part of the URL.' is now 'Customize the last part of the Permalink.'
- Inspect the source and observe this text is associated as a description of the input field via an `aria-describedby` attribute.
- Observe the permalink preview after the input field is now prepended with text to commnicate what it is: 'ermalink:' and it is styled like a link (blue instead of grey).
- Go to the admin Settings > Reading settings, and set WordPress to use a static page for the home page.
- Edit that page.
- Observe in the Inspector post panel the row that shows the Front page link is called 'Link'.
- Observe the UI shows the clickable link instead of a button that opens a popover.

## Screenshots or screencast 

Before:

before

After:

after

### DIFF
--- a/packages/editor/src/components/post-url/index.js
+++ b/packages/editor/src/components/post-url/index.js
@@ -17,7 +17,7 @@ import {
 import { store as noticesStore } from '@wordpress/notices';
 import { copySmall } from '@wordpress/icons';
 import { store as coreStore } from '@wordpress/core-data';
-import { useCopyToClipboard } from '@wordpress/compose';
+import { useCopyToClipboard, useInstanceId } from '@wordpress/compose';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


afercia:
It is worth noting that after https://github.com/WordPress/gutenberg/pull/64053 a long URL / slug is now truncated. I kindly disagree with that change. It defeats the purpose of this UI, which is to show the _full_ URL or slug.

afercia:
Latest commits rebase on top of latest trunk and remove the truncation introduced in https://github.com/WordPress/gutenberg/pull/64053

richtabor:
We need to do a holistic design pass on this, rather than one-off PRs that don't solve: 

1. It's difficult to parse the information presented as-is with a lack of visual hierarchy. 
2. The [help link](https://wordpress.org/documentation/article/page-post-settings-sidebar/#permalink) is likely unhelpful. I question if it's adding value. 
3. It's odd having a copy button that copies more than what you see will be copied. I'm also not confident what goal this "copy" action supports.  
4. Why is there a link to the post? What goal does this accomplish for users? What is this helping them do? If the post is not saved, that link is wrong and does not confirm the post permalink is updated. I say we just remove it to remove that confusion. 

Why not simplify big time, making this just about editing the permalink (the primary action here), and providing additional context to what this is and what it's for, without sending people to an external link:

CleanShot 2024-09-11 at 13 37 48

### Visual references

### This PR

CleanShot 2024-09-11 at 13 16 22

### Trunk 

CleanShot 2024-09-11 at 13 17 46

joedolson:
I can see some of the arguments in both these proposals. 

Removing the copy button seems reasonable; that seems unnecessary, and I agree that it's counter-intuitive. It might make sense to move it to be associated with the link, rather than the input, since that's what it actually copies. 

But we really need to change the labeling to say 'Slug', not 'Link' or 'Permalink'. This field only edits the slug, and it is simply incorrect to label it as anything else. 

Regarding the link to the page: as far as I know, this is the only place where that link is made available in the interface. I think it's a useful reference. 

I don't think that the 'Learn more' link is very useful there, and would be better omitted. It might be more useful if the target ID actually existed on that page.

afercia:
> Why not simplify big time, making this just about editing the permalink 

I see there's still some confusion about the actual editable value. This UI is meant to edit  the _slug_. Instead, the _permalink_ is the full URL, which is not fully editable. Only the slug can be edited here.

> Regarding the link to the page: as far as I know, this is the only place where that link is made available in the interface. I think it's a useful reference.

I seem to recall there was some prior direct user feedback to have a plave where users can see the full URL (permalink). @annezazu may help find some reference. To me:
- The editor must show the full URL somewhere.
- I'd agree the setting panel or a popover aren't ideal to show a value that is potentially very long. This has been discussed at length since https://github.com/WordPress/gutenberg/issues/577 (April 2017) and in a few following issues and PRs.
  - A value that by its own nature may be very long needs way more horizontal space to be read comfortably.
  - The settings panel may be closed. In this case, an important information is hidden behind a closed panel.

However, in more than seven years the contributors didn't reach an agreement on the Edit permalink UI placement. That's a way broader design issue that is out of the scope of this PR. This PR only aims to fix obvious issues in the labeling and naming of this UI.

For history, some references about the discussion and changes for the Edit Permalink UI:
https://github.com/WordPress/gutenberg/issues/577
https://github.com/WordPress/gutenberg/issues/1285
https://github.com/WordPress/gutenberg/pull/5756
https://github.com/WordPress/gutenberg/issues/5494
https://github.com/WordPress/gutenberg/issues/10996
https://github.com/WordPress/gutenberg/issues/11858
https://github.com/WordPress/gutenberg/pull/12030
https://github.com/WordPress/gutenberg/issues/12031
https://github.com/WordPress/gutenberg/issues/12033
https://github.com/WordPress/gutenberg/pull/21099
https://github.com/WordPress/gutenberg/issues/52466

annezazu:
I think you're thinking of this! https://github.com/WordPress/gutenberg/issues/64659

joedolson (approved):
As far as I'm concerned, this is a necessary change to fix incorrectly labeled fields. This looks good to me; but I'm not adept at the React portions.

afercia:
This PR only aims to fix obvious issues in the labeling and naming of the Edit permalink UI. This UI hasroom for futher improvements that would be great to address and they're already tracked in other issues. I'd like to suggest to move further feedback on those specific issues.

In the meantime, I think this change is an improvement as it restores a correct terminology that has been in use in WordPress for ages and users are familiar with.

Screenshot before and after:

before and after

afercia:
Last commit reduces the top margin of the intro text. Screenshot before and after:

before and after fixed top margin

afercia:
> I don't think that the 'Learn more' link is very useful there, and would be better omitted. It might be more useful if the target ID actually existed on that page.

Note: I pinged the documentation team on https://github.com/WordPress/Documentation-Issue-Tracker/issues/1704 to inform them of this change.

packages/editor/src/components/post-url/index.js
```diff
@@ -70,25 +70,29 @@ export default function PostURL( { onClose } ) {
...
 	const [ forceEmptyField, setForceEmptyField ] = useState( false );
 	const copyButtonRef = useCopyToClipboard( permalink, () => {
-		createNotice( 'info', __( 'Copied URL to clipboard.' ), {
+		createNotice( 'info', __( 'Copied Permalink to clipboard.' ), {
 			isDismissible: true,
 			type: 'snackbar',
 		} );
 	} );
+	const postUrlSlugDescriptionId =
+		'editor-post-url__slug-descriotion-' + useInstanceId( PostURL );
```
gigitux, line 79: It looks like that there is a typo. It should be: `editor-post-url__slug-description-`.  @afercia, can you take care of fixing it?

afercia: @gigitux Good catch, thanks. Quick PR incoming.

PR #63671, opened 2024-07-17
Allow multi-select on iOS Safari/touch devices

### LINKED ISSUE #16206: Can’t select multiple blocks at a time on iOS
**Describe the bug**
Selecting multiple blocks at a time while using the block editor on mobile seems impossible. This makes reorganizing headings and the following paragraphs very difficult and time consuming.

**Screenshots**
(Will upload a screen recording here, currently on bad WiFi)

**Smartphone (please complete the following information):**
 - Device: iPad (6th generation)
 - OS: 12.3.1
 - Browser: Firefox

ellatrix:
## What?

Fixes #16206.

## Why?

Currently it's not possible to select multiple block on iOS at all.

## How?

As soon as there's an uncollapsed selection, prepare for multi-selection with the content editable wrapper.

Note that we already do this for pointer device selection, when the pointer drags out of a block, and also whenever you arrow select out of a block.

The one consequence of doing this is that focus will move from the rich text element in the block to the writing flow wrapper. There's a potential fix by always forcing focus back to the original rich text element, but the browser will continuously put focus back to writing flow whenever an event happens. The other solution, which I have implemented in this PR, is to dispatch a cloned event on the rich text element that does not bubble, so all the usual event handlers are called.

## Alternatives considered

I considered a more similar approach to what we currently have for pointer devices. There's no `touchout` event like there is a `mouseout` event, but _that_ can be fixed. On touch devices, you can only change an uncollapsed selection when there already is an uncollapsed selection (to get the handle bars). So we could enable the content editable wrapper "right in time" on `touchstart` whenever there is already an uncollapsed selection. However, I tried this multiple times and iOS seems to lose the selection whenever the content editable wrapper is turned on... The approach above is the only one that I could make work, and all e2e tests pass as well.

## Testing Instructions

This can more easily be tested with Playground. Open [this link](https://playground.wordpress.net/#%7B%22landingPage%22:%22/wp-admin/%22,%22steps%22:%5B%7B%22step%22:%22login%22,%22username%22:%22admin%22,%22password%22:%22password%22%7D,%7B%22step%22:%22mkdir%22,%22path%22:%22/wordpress/pr%22%7D,%7B%22step%22:%22writeFile%22,%22path%22:%22/wordpress/pr/pr.zip%22,%22data%22:%7B%22resource%22:%22url%22,%22url%22:%22/plugin-proxy.php?org=WordPress&repo=gutenberg&workflow=Build%2520Gutenberg%2520Plugin%2520Zip&artifact=gutenberg-plugin&pr=63671%22,%22caption%22:%22Downloading%20Gutenberg%20PR%2063671%22%7D,%22progress%22:%7B%22weight%22:2,%22caption%22:%22Applying%20Gutenberg%20PR%2063671%22%7D%7D,%7B%22step%22:%22unzip%22,%22zipPath%22:%22/wordpress/pr/pr.zip%22,%22extractToPath%22:%22/wordpress/pr%22%7D,%7B%22step%22:%22installPlugin%22,%22pluginZipFile%22:%7B%22resource%22:%22vfs%22,%22path%22:%22/wordpress/pr/gutenberg.zip%22%7D%7D%5D%7D) in Safari iOS and try to select multiple paragraphs.

## Screenshots or screencast 

Simulator Screen Shot - iPhone 14 Pro Max - 2022-11-09 at 17 18 40

### DIFF
--- a/packages/block-editor/src/components/writing-flow/use-arrow-nav.js
+++ b/packages/block-editor/src/components/writing-flow/use-arrow-nav.js
@@ -237,7 +237,10 @@ export default function useArrowNav() {
 				return;
 			}
 
-			const target = getSelectionRoot( ownerDocument ) || event.target;
+			const target =
+				ownerDocument.activeElement === node
+					? getSelectionRoot( ownerDocument )
+					: event.target;
 
 			// Abort if our current target is not a candidate for navigation
 			// (e.g. preserve native input behaviors).
--- a/packages/block-editor/src/components/writing-flow/use-arrow-nav.js
+++ b/packages/block-editor/src/components/writing-flow/use-arrow-nav.js
@@ -19,6 +19,7 @@ import { useRefEffect } from '@wordpress/compose';
  */
 import { getBlockClientId, isInSameBlock } from '../../utils/dom';
 import { store as blockEditorStore } from '../../store';
+import { getSelectionRoot } from './utils';
 
 /**
  * Returns true if the element should consider edge navigation upon a keyboard
@@ -134,7 +135,11 @@ export function getClosestTabbable(
 		}
 
 		// Skip focusable elements such as links within content editable nodes.
-		if ( node.isContentEditable && node.contentEditable !== 'true' ) {
+		if (
+			node.isContentEditable &&
+			node.contentEditable !== 'true' &&
+			! node.getAttribute( 'tabindex' )
+		) {
 			return false;
 		}
 
@@ -190,8 +195,7 @@ export default function useArrowNav() {
 				return;
 			}
 
-			const { keyCode, target, shiftKey, ctrlKey, altKey, metaKey } =
-				event;
+			const { keyCode, shiftKey, ctrlKey, altKey, metaKey } = event;
 			const isUp = keyCode === UP;
 			const isDown = keyCode === DOWN;
 			const isLeft = keyCode === LEFT;
@@ -232,6 +236,8 @@ export default function useArrowNav() {
 
 				return;
 			}
+
+			const target = getSelectionRoot( ownerDocument ) || event.target;
 
 			// Abort if our current target is not a candidate for navigation
 			// (e.g. preserve native input behaviors).
@@ -282,6 +288,7 @@ export default function useArrowNav() {
 				);
 
 				if ( closestTabbable ) {
+					node.contentEditable = false;
 					placeCaretAtVerticalEdge(
 						closestTabbable,
 						// When Alt is pressed, place the caret at the furthest
@@ -302,6 +309,7 @@ export default function useArrowNav() {
 					isReverseDir,
 					node
 				);
+				node.contentEditable = false;
 				placeCaretAtHorizontalEdge( closestTabbable, isReverse );
 				event.preventDefault();
 			}
--- a/packages/block-editor/src/components/writing-flow/use-arrow-nav.js
+++ b/packages/block-editor/src/components/writing-flow/use-arrow-nav.js
@@ -135,11 +135,7 @@ export function getClosestTabbable(
 		}
 
 		// Skip focusable elements such as links within content editable nodes.
-		if (
-			node.isContentEditable &&
-			node.contentEditable !== 'true' &&
-			! node.getAttribute( 'tabindex' )
-		) {
+		if ( node.isContentEditable && node.contentEditable !== 'true' ) {
 			return false;
 		}
 
@@ -283,6 +279,7 @@ export default function useArrowNav() {
 				( altKey ? isHorizontalEdge( target, isReverseDir ) : true ) &&
 				! keepCaretInsideBlock
 			) {
+				node.contentEditable = false;
 				const closestTabbable = getClosestTabbable(
 					target,
 					isReverse,
@@ -291,7 +288,6 @@ export default function useArrowNav() {
 				);
 
 				if ( closestTabbable ) {
-					node.contentEditable = false;
 					placeCaretAtVerticalEdge(
 						closestTabbable,
 						// When Alt is pressed, place the caret at the furthest
@@ -307,12 +303,12 @@ export default function useArrowNav() {
 				isHorizontalEdge( target, isReverseDir ) &&
 				! keepCaretInsideBlock
 			) {
+				node.contentEditable = false;
 				const closestTabbable = getClosestTabbable(
 					target,
 					isReverseDir,
 					node
 				);
-				node.contentEditable = false;
 				placeCaretAtHorizontalEdge( closestTabbable, isReverse );
 				event.preventDefault();
 			}
--- a/packages/block-editor/src/components/writing-flow/use-event-redirect.js
+++ b/packages/block-editor/src/components/writing-flow/use-event-redirect.js
@@ -1,0 +1,73 @@
+/**
+ * WordPress dependencies
+ */
+import { useRefEffect } from '@wordpress/compose';
+
+/**
+ * Internal dependencies
+ */
+import { getSelectionRoot } from './utils';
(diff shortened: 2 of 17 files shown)


packages/rich-text/src/component/event-listeners/input-and-selection.js
```diff
@@ -114,14 +114,13 @@ export default ( props ) => ( element ) => {
 			return;
 		}
 
-		// Ensure the active element is the rich text element.
-		if ( ownerDocument.activeElement !== element ) {
-			// If it is not, we can stop listening for selection changes. We
-			// resume listening when the element is focused.
-			ownerDocument.removeEventListener(
```
ellatrix, line 121: To do: it would be good to attach one event handler in the first instance, then reuse it for all instances.

packages/block-editor/src/components/writing-flow/use-event-redirect.js
```diff
@@ -0,0 +1,73 @@
...
+			const Constructor = defaultView[ constructorName ];
+			const root = getSelectionRoot( ownerDocument );
+
+			if ( ! root ) {
+				return;
+			}
+
+			const init = {};
+
+			for ( const key in event ) {
```
ntsekouras, line 39: Why not just spread the object here?

packages/block-editor/src/components/writing-flow/use-arrow-nav.js
```diff
@@ -134,7 +135,11 @@ export function getClosestTabbable(
 		}
 
 		// Skip focusable elements such as links within content editable nodes.
-		if ( node.isContentEditable && node.contentEditable !== 'true' ) {
+		if (
+			node.isContentEditable &&
+			node.contentEditable !== 'true' &&
+			! node.getAttribute( 'tabindex' )
```
ntsekouras: Can't this have side effects somehow? Why this check was introduced?

packages/block-editor/src/components/writing-flow/use-event-redirect.js
```diff
@@ -0,0 +1,73 @@
...
+ * we need to dispatch some events to the root of the selection to ensure
+ * compatibility with rich text. In the future, perhaps the rich text event
+ * handlers should be attached to the window instead.
+ *
+ * Alternatively, we could try to find a way to always maintain rich text focus.
+ */
+export default function useEventRedirect() {
+	return useRefEffect( ( node ) => {
+		function onInput( event ) {
+			if ( event.target !== node || event.__isRedirected ) {
```
ntsekouras: When this `event.__isRedirected` is `true`?

packages/block-editor/src/components/writing-flow/use-input.js
```diff
@@ -59,6 +77,23 @@ export default function useInput() {
...
+								? 'collapseToEnd'
+								: 'collapseToStart';
+						selection[ method ]();
+						event.preventDefault();
+						return;
+					}
+				}
+
 				if ( event.keyCode === ENTER ) {
 					if ( event.shiftKey || __unstableIsFullySelected() ) {
```
ntsekouras, line 98: Not from here, but curious how `Enter` and `shift` are used..  Any chance you remember?

packages/block-editor/src/components/writing-flow/utils.js
```diff
@@ -116,3 +116,33 @@ function toPlainText( html ) {
...
+export function getSelectionRoot( ownerDocument ) {
+	const { defaultView } = ownerDocument;
+	const { anchorNode, focusNode } = defaultView.getSelection();
+
+	if ( ! anchorNode || ! focusNode ) {
+		return;
+	}
+
+	const anchorElement = (
+		anchorNode.nodeType === anchorNode.ELEMENT_NODE
```
ntsekouras, line 134: What would be the cases that we don't have an element node and when?

packages/components/src/autocomplete/index.tsx
```diff
@@ -444,8 +444,14 @@ export function useAutocompleteProps( options: UseAutocompleteProps ) {
 			function _onKeyDown( event: KeyboardEvent ) {
 				onKeyDownRef.current?.( event );
 			}
+			// To do: comment.
```
ntsekouras: Do it 😄

test/e2e/specs/editor/various/block-deletion.spec.js
```diff
@@ -287,15 +287,16 @@ test.describe( 'Block deletion', () => {
 		await expect.poll( editor.getBlocks ).toMatchObject( [
 			{ name: 'core/paragraph', attributes: { content: 'First' } },
 			{ name: 'core/paragraph', attributes: { content: 'Second' } },
-			{ name: 'core/paragraph', attributes: { content: '' } },
 		] );
 
 		// Ensure that the newly created empty block is focused.
-		await expect.poll( editor.getBlocks ).toHaveLength( 3 );
+		await expect.poll( editor.getBlocks ).toHaveLength( 2 );
```
ntsekouras, line 293: Why you had to change this?

ntsekouras (commented):
It's a very specialized PR with selections and events, so I took a first look and left some questions to start understanding better. Sorry if some of them might seem obvious to you.

Does your redirection approach mean that we continuously move focus on every related event or it's just about keeping the new focused node (writing flow wrapper) and redirecting the event to rich text so it behaves as if it had focus?

packages/block-editor/src/components/writing-flow/utils.js
ellatrix: Selection is most of the time from text nodes

test/e2e/specs/editor/various/block-deletion.spec.js
ellatrix: Because it's the only thing in the e2e tests that this PR changed in behaviour. Arrow up selects two blocks. Then press Backspace. Instead of leaving an empty block, it no longer leaves an empty block and puts selection at the end of the previous block.

packages/block-editor/src/components/writing-flow/use-input.js
ellatrix: Yes, this is Enter behaviour for partial selection. So select two paragraphs partially and press Enter, this handles split for them.

ellatrix:
> it's just about keeping the new focused node (writing flow wrapper) and redirecting the event to rich text so it behaves as if it had focus?

Yes, we keep focus on the wrapper, and instead clone the event and dispatch it onto the content editable element that has the selection. I mention in the PR description that I also tried maintaining focus on the original content editable, but this is pretty fragile as the browser keeps setting it back to writing flow.

packages/block-editor/src/components/writing-flow/use-arrow-nav.js
ellatrix: This check is here to skip focusable elements inside `contenteditable` such as links, see comment above. For this PR, we actually do want to allow focus when an element has a tab index, because the whole writing flow can become `contenteditable`, and we want blocks still to be considered in `getClosestTabbable`. Note that `getClosestTabbable` is very specific to arrow navigation.

packages/components/src/autocomplete/index.tsx
ellatrix: Looks like this is no longer needed after other fixes.

packages/block-editor/src/components/writing-flow/use-event-redirect.js
ellatrix: See 20 lines below :)

ntsekouras (approved):
LGTM Ella, thank you!

stokesman:
🎉  This stone incidentally got another bird #64617.

PR #63676, opened 2024-07-17
ColorPicker: Use `minimal` variant for SelectControl

mirka:
Follow-up to #63265

## What?

Replace the "unofficial" borderless `SelectControl` in `ColorPicker` with the official `minimal` variant.

## Why?

We used to achieve this by passing a context value under the table, but now we have an official variant to achieve the same thing.

Also, I accidentally introduced a regression to this implementation in `ColorPicker` (the `isBorderless` context property needed to be moved under a [`__overrides` property](https://github.com/WordPress/gutenberg/blob/88dffa048232c538ffdcaff6cd6a5ece2454706f/packages/components/src/context/use-context-system.js#L45) so it takes precedence), so this also fixes that.

## Testing Instructions

See Storybook story for ColorPicker.

## Screenshots or screencast 

| Before (regressed) | After (correct) |
|--------|-------|
|Select has border|Select is borderless|


ciampo (approved):
Love the 🧹  !

🚀

tyxla (approved):
Nice one! The left margin actually makes sense now 😅 ❤️

Issue #63501, opened 2024-07-12
New blocks to expand pattern and theme designs


creativecoder:
This issue tracks new blocks that will help expand possible designs for themes and patterns. Most represent common design patterns on the web that aren't yet possible with the default block library:

- [ ] https://github.com/WordPress/gutenberg/issues/16351
- [ ] https://github.com/WordPress/gutenberg/issues/16484
- [x] https://github.com/WordPress/gutenberg/issues/21584
- [ ] https://github.com/WordPress/gutenberg/issues/21943
- [ ] https://github.com/WordPress/gutenberg/issues/34075
- [ ] https://github.com/WordPress/gutenberg/issues/34079
- [ ] https://github.com/WordPress/gutenberg/issues/41730
- [ ] https://github.com/WordPress/gutenberg/issues/43369
- [ ] https://github.com/WordPress/gutenberg/issues/48249
- [ ] https://github.com/WordPress/gutenberg/issues/50433

**Related**
- [Design Tools Overview](https://github.com/WordPress/gutenberg/issues/33447)
- Discussion in [Proposal for Core Blocks in the Directory](https://github.com/WordPress/gutenberg/issues/58773)

creativecoder:
> There is no open issue for it, but users seems to be interested in a style variation toggle for the front aka "dark mode" toggle block. This was mentioned multiple times in the post for Twenty Twenty-Five.
> -- from https://github.com/WordPress/gutenberg/issues/58773#issuecomment-2219415204

@carolinan Thanks for the suggestion! This issue seems related [Core Block suggestion: Theme Style Toggle #48249](https://github.com/WordPress/gutenberg/issues/48249); I've added it above.

Luehrsen:
I appreciate the effort to expand WordPress’s capabilities, but I have concerns about the inclusion of blocks proposed here in core. These blocks feel highly opinionated, catering to specific design trends or preferences, and I question whether they align with the broader needs of WordPress users. Just because something is "common" does not warrant it being in core.

WordPress has always excelled by providing a lean, reliable core while enabling customization through plugins. Adding blocks with such opinionated use cases directly to core risks bloating the platform and complicating the experience for users who don’t need them. These features seem far better suited for plugins rather than core functionality.

Including blocks like these also introduces long-term maintenance challenges. Every block added to core requires ongoing updates, compatibility checks, and support, which places an unnecessary burden on resources that could be focused on improving foundational features.

[...]

Issue #34079, opened 2021-08-14
New Block: Tabs


deborah86:
## What problem does this address?

It is difficult to display a lot of grouped content on a page. The only option is to have all the content display at once on a page because there are no core blocks that hide and show content. A lot of third-party blocks do not follow the standards for tabs.

## What is your proposed solution?

The proposed solution is to have a core tab block. See the examples from these websites:

- [Tabs | Accessibility Guideliens](http://web-accessibility.carnegiemuseums.org/code/tabs/#:~:text=Tabs%20Tabs%2C%20which%20are%20very%20similar%20to%20accordions%2C,based%20on%20a%20user%20click%20or%20focus%20event.)
- [14 Guidelines For Web Site Tabs Usability](https://usabilitygeek.com/14-guidelines-for-web-site-tabs-usability/)
- [Tabs, Used Right](https://www.nngroup.com/articles/tabs-used-right/)

mtias:
See before: #9543

Also related #21584.

jarekmorawski:
Over the last few weeks, I worked on a design for a Tabs block. It was a high-priority project because without tabs, we can't fully convert existing PHP page templates in WooCommerce into blocks. The new block would also unlock countless customization and layout options for commerce-specific block patterns.

https://github.com/WordPress/gutenberg/assets/79307566/89ee3e7d-a1da-4244-952f-719d24f31ac8

In the design, each Tab would be a separate block inside the parent Tabs block.

![image](https://github.com/WordPress/gutenberg/assets/79307566/1d81c2a7-24aa-4388-a347-af2a23222af6)

Each tab is a separate Tab block with its own settings: layout (justification, orientation), color, typography, dimensions, and border & shadow.

![image](https://github.com/WordPress/gutenberg/assets/79307566/e929e12a-b138-44e1-8426-9912a7c55507)

They only affect the tab’s appearance, not its content. If someone wanted to style the tab content individually, they could add a group block and set up spacings and colors there.

Tab content is displayed as inner blocks and visible in the List View.

![image](https://github.com/WordPress/gutenberg/assets/79307566/0c5eb97a-1c08-4998-90f7-90a5db03f60b)

Similar to the Buttons block, users can click the + button to add a new tab. They can remove it like any other block by selecting it in the canvas or the List View. Users can edit the tab’s content by selecting it in the canvas, the same as the visitors would on the page. They can also choose a block in the List View.

[...]

Issue #57719, opened 2024-01-10
Blocks: States


SaxonF:
## What problem are we solving?
Websites are becoming far more dynamic, motion filled and interactive partly due to the evolution of front-end libraries like React and frameworks like NextJS, and motion libraries like Framer Motion as well as the View Transitions API. To remain competitive, we should start thinking about how to empower Gutenberg users to build similar sorts of experiences without needing to code. The Interactivity API has given us a foundation to build upon.

When you think about motion and interactivity a common concept is transitioning from one state to another. For example, an image on your site going from opacity: 0 to opacitiy: 1. There are other needs that follow similar concepts, such as styling a button before hovering and a button after, or how a section displays on desktop vs mobile. Although the navigation overlay spawned this discussion, the “why” behind this work goes beyond a single need. We are attempting to solve multiple problems at once, which brings us on to the proposal.

## What strategy does it align with / why is it worth solving?

This a broad initiative that aligns with attracting designers, developers and low/no-code audiences.

## What does the solution look like?
This proposal introduces the idea of “states” or “variants” as they known in other design tools like Figma. The idea is that any block, including synced patterns, can define `properties`. Combining property values create `states`. Each `state` is able to modify its style or contents. To transition between states, you attach a `behaviour` to either the block itself or one of its children. A behaviour consists of a `trigger` (e.g. “click block”) and an `action` (e.g. “set property a to xyz”). When transitioning between states we can optionally smartly animate between them (e.g. via view transitions API).

_Animation_

https://github.com/WordPress/gutenberg/assets/1072756/c3b75e58-2798-4554-9363-6bc242d2d8e8

The use of states go beyond interactivity though and could also be a mechanism for providing responsive behaviour per block. Consider a group that has an inner Row block. It has a “Viewport” property by default that emulates a behaviour of “when screen size is < 480 set viewport property to mobile” . We change to the “mobile” property and update the Row to a Stack. 

_Responsive_

https://github.com/WordPress/gutenberg/assets/1072756/74c973b2-baae-483b-8049-4063bb4d46b7

_Button state styling_

https://github.com/WordPress/gutenberg/assets/1072756/8f8dcfd1-4075-451c-8788-2adf29ffe94f

_Navigation overlay_

https://github.com/WordPress/gutenberg/assets/1072756/7755e14a-0677-4c1e-9925-30043ca112e4

Blocks like the navigation overlay can wrap this experience into something a little more user friendly (or hide properties completely)

## Use Cases
[...]

PR #63689, opened 2024-07-18
Adds tabs block behind block experiments flag

creativecoder:
## What?

Adds an experimental tabs block. To enable it, turn on Experimental blocks from Gutenberg > Experiments in wp-admin.

## Why?

To enable tabbed content for sites and themes. See https://github.com/WordPress/gutenberg/issues/34079. Part of https://github.com/WordPress/gutenberg/issues/63501.

## How?

Includes 2 blocks:
 
1. core/tabs is the container, it handles rendering the tab labels. It can only contain core/tab blocks.
2. core/tab holds the content for each tab used for the tab panels. It can contain any kind of block normally used in the editor.

This first version is behind the block experiments so that we can iterate as needed without writing block deprecations.

## Testing Instructions

- Go to `/wp-admin/admin.php?page=gutenberg-experiments` and turn on Experimental blocks
- Insert a tabs block in the editor
- Change the title of the tabs and add content
- Add additional tabs
- Save and view the front end to see the tabbed content
- Check that markup contains the proper ARIA attributes described in the [Tabs Pattern guide](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#wai-ariaroles,states,andproperties)

Other features to test

- Anchor property for Tab blocks can be used to set the id for each tab (select an individual tab block in the list view then go to the advanced settings in the right sidebar)
- Selecting an inactive tab or its inner blocks from the list view will make that tab to be active
- Individual tab blocks can be dragged from one tabs block to another
- Front-end markup with JS disabled is a list of on page links

In the front-end:

- Use <kbd>Tab</kbd> to select the active tab label and tab panel
- With a tab label selected, use <kbd>right arrow</kbd> and <kbd>left arrow</kbd> to select and active tabs

### Editor

https://github.com/user-attachments/assets/4559c3ed-53a6-460d-a2b3-5c112626f064

### Front-end

https://github.com/user-attachments/assets/4e919679-8066-4aa3-8fbf-c583f071d6c0

## Further work

This initial version is intentionally minimal to try and reduce the amount of code to review. Items for further work in future PRs include

- Improving inner blocks so block selection is more intuitive (right now selecting a tab label selects the tabs wrapper block, rather than the individual tab)
- When adding a new tab in the editor, focus the new tab's label
- Adding default focus styles for active tab, in case the theme doesn't have them
- Style supports for the tabs container, tab labels, and tab panel content
- Better integration anchor supports, to ensure there are no duplicate ids on the page
- A block variation for vertical tabs
- Deep linking for tab panels to open the page 
- Conditionally adding `tabindex="0"` to the active tab panel only if it does not contain content where the first element in the panel is focusable (see [ARIA example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/#accessibilityfeatures))
- Adding automated unit, integration, and end-to-end tests
- Option to select the default tab that's open
- Responsive styling that stacks tabs for small viewports


gziolo:
There is an issue open for this block at https://github.com/WordPress/gutenberg/issues/34079. I also worked on the same block last week, and I have a working prototype, but it's outside the Gutenberg repository in case you would like to resue any parts of the code: https://github.com/a8cteam51/special-projects-blocks-monorepo/pull/12. As noted in https://github.com/a8cteam51/special-projects-blocks-monorepo/pull/12#issuecomment-2238997046, I'm not going to work on this block in the next weeks actively.

creativecoder:
@gziolo @luisherranz I'm trying to get the interactivity API working with this block and could use some help!

Here's the commit that adds it: https://github.com/WordPress/gutenberg/pull/63689/commits/a5b02b969c5b40577ca8926e8131e2d4bc0e517d

I'm seeing this JS console error in the editor (which is strange, because I didn't think `viewScriptModule` is loaded in the editor):
```
Uncaught TypeError: can't access property "store", _wordpress_interactivity__WEBPACK_IMPORTED_MODULE_0__ is undefined
    <anonymous> view.js:6
    <anonymous> view.min.js:105
    <anonymous> view.min.js:107
```

And I'm seeing this error on the front end (is seems like a script dependency isn't getting set correctly):
```
Uncaught TypeError: can't access property "interactivity", window.wp is undefined
    <anonymous> view.min.js:11
```

luisherranz:
If I am not mistaken, currently in Gutenberg, we are manually enqueuing the frontend files. So try adding a render callback where you include this code (or something similar):

```php
$suffix = wp_scripts_get_suffix();
if ( defined( 'IS_GUTENBERG_PLUGIN' ) && IS_GUTENBERG_PLUGIN ) {
  $module_url = gutenberg_url( '/build/interactivity/tab.min.js' );
}

wp_register_script_module(
  '@wordpress/block-library/tab',
  isset( $module_url ) ? $module_url : includes_url( "blocks/tab/view{$suffix}.js" ),
  array( '@wordpress/interactivity' ),
  defined( 'GUTENBERG_VERSION' ) ? GUTENBERG_VERSION : get_bloginfo( 'version' )
);

wp_enqueue_script_module( '@wordpress/block-library/tab' );
```

And add the view.js file [to this list](https://github.com/WordPress/gutenberg/blob/4b873d66118aa94ba2166e1d743fa4379d265fe1/tools/webpack/interactivity.js#L27) (then restart Webpack).

Let me know if that works because there might be other steps that I don't remember right now.

luisherranz:
By the way, if you are testing, that's fine, but the final version should not include any directives in the markup serialized to the database. They should be injected using the HTML tag processor in the render callback. That way, the frontend implementation can be changed or improved without deprecations.

packages/block-library/src/tabs/edit.js
```diff
@@ -0,0 +1,197 @@
...
+									'is-active': isActive,
+								} ) }
+								href={ '#' + tabPanelId }
+								id={ tabLabelId }
+								onClick={ () => setActiveTab( block.clientId ) }
+								role="tab"
+								tab-index={ tabIndexAttr }
+							>
+								<RichText
+									allowedFormats={ ALLOWED_FORMATS }
```
fabiankaegy:
```suggestion
									withoutInteractiveFormatting
```

Instead of maintaining the list of allowed formats we should instead leverage the `withoutInteractiveFormatting` prop. This still allows 3rd party custom rich text formats o be used. As long as they are not interactive.

creativecoder: Thank you for the pointer! Implemented in https://github.com/WordPress/gutenberg/pull/63689/commits/c6e184671d55d6ecb3008a014902c87f5499d07c

creativecoder:
This seems to be working pretty well now and is ready for review.

packages/block-library/src/tabs/view.js
```diff
@@ -0,0 +1,212 @@
...
+		} )
+	);
+}
+
+/**
+ * Initializes the tabs block with the necessary ARIA attributes.
+ *
+ * @param {HTMLElement} ref The block wrapper element.
+ */
+function initTabs( ref ) {
```
fabiankaegy: What's the reason for doing all these DOM manipulations here manually instead of using directives for it? 

This way they won't get server rendered and there will be a flash of unsettled content as the page loads 🤔

packages/block-library/src/tabs/view.js
```diff
@@ -0,0 +1,212 @@
...
+		tabList.setAttribute( 'role', 'tablist' );
+
+		// Use the hidden <h3>Contents</h3> element as the label for the tab list, if found.
+		const tabListLabel = ref.querySelector( '.wp-block-tabs__title' );
+		if ( tabListLabel ) {
+			const tabListLabelId = tabListLabel.id || 'tablist-label';
+			tabListLabel.id = tabListLabelId;
+			tabList.setAttribute( 'aria-labelledby', tabListLabelId );
+		} else {
+			tabList.setAttribute( 'aria-label', 'Tabs' );
```
fabiankaegy: Any strings should be translatable. Also again the question about not using directives here :)

bph:
I love that this is coming to core.🎉 Great work
I did some first user testing (me = user) and probably just confirming what you already know. All these things can be fixed in later PRs, of course. 

- Found the  < >  but it was difficult to find the right spot to click. The content block inside the tab got in the way. 
- Same problem to clicking in the appropriate spot to find the Anchor section. 
- Click on the single Tab in the Editor should give a default paragraph block, prompt, so I could start writing right away, or use the "/" slash comment to add blocks. 
- Could use background color tools for the tabs as well as for single tab block.
- Also borders and radius for tab riders. 
- For now it works by putting it in a group block and adding some margin. 
- The drag/drop of a single Tab into another Tabs block didn't work. couldn't get a landing space show up.
- The Tab size, could use a minimum height setting as well, so the content on the rest of the page doesn't jump when single tab content is smaller than others.


https://github.com/user-attachments/assets/962bddc6-2dd3-47ba-915d-35f3ebd7da94

fabiankaegy:
Hey @creativecoder 👋 

I've been working on a version of this block for client projects for a while now and the solution we came up with for having the tab button / label RichText actually live inside the tab item block so that it properly gets selected when you interact with it was using the `SlotFill` system:

```js
const { Fill, Slot } = createSlotFill('TabHeader');

export const TabHeader = ({ children, tabsClientId, orientation }) => {
	return (
		<Fill orientation={orientation} name={`TabHeader-${tabsClientId}`}>
			{children}
		</Fill>
	);
};

export const TabHeaderSlot = ({ tabsClientId, orientation }) => {
	return (
		<Slot
			orientation={orientation}
			name={`TabHeader-${tabsClientId}`}
			bubblesVirtually
			as="ul"
			className="tab-list wp-block-tabs__tab-list"
		/>
	);
};
```

With that slot setup I can then use the following inside the edit.js of the Tab Item block:

```js
<TabHeader tabsClientId={parentBlockClientId} orientation={orientation}>
	<li
		className={clsx('tab-item', 'wp-block-tabs__tab-item', {
			'is-active': isSelectedTab,
		})}
	>
		<RichText
			tagName="span"
			withoutInteractiveFormatting
			value={label}
			onChange={(value) => setAttributes({ label: value })}
		/>
	</li>
</TabHeader>
{isSelectedTab && (
	<>
		<div className="tab-control wp-block-tabs__tab-control">
			<TabHeaderSlot
				tabsClientId={parentBlockClientId}
				orientation={orientation}
			/>
		</div>
		<div className="tab-group wp-block-tabs__tab-group">
			<div {...innerBlocksProps} />
		</div>
	</>
)}
```

And in the parent Tabs block I have this: 

```js
<TabHeader tabsClientId={clientId} orientation={orientation}>
	{isSelected && (
		<li
			className="tab-item wp-block-tabs__tab-item tab-item--inserter"
		>
			<InnerBlocks.ButtonBlockAppender
				icon="plus"
				onClick={appendTabItem}
			/>
		</li>
	)}
</TabHeader>
```

https://github.com/user-attachments/assets/8410e822-21a9-409e-8df3-2f4b5d10249c

ndiego:
This is looking great! But as work continues, I want to highlight some extensibility "opportunities" that might be worth considering since this will be a new block. 

- Icons in tab labels are quite common in third-party tab blocks, so adding an extensibility mechanism that allows developers to add content to the right or left of a tab label would be great. Having this in Core would be preferred, as highlighted in https://github.com/WordPress/gutenberg/issues/34079#issuecomment-2182162630, but this is a bigger project. There are other blocks, like buttons, that would also benefit from an icon picker. 
- Sometimes, the first tab on the left is not the one the user wants to have open by default. So, allowing users to choose the default tab would be a nice enhancement.

packages/block-library/src/tabs/index.php
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		if ( $p->has_class( 'wp-block-tabs__tab-label' ) ) {
+			$p->set_attribute( 'data-wp-bind--aria-selected', 'state.isActiveTab' );
+			$p->set_attribute( 'data-wp-bind--tabindex', 'state.tabindexLabelAttribute' );
+			$p->set_attribute( 'data-wp-on--click', 'actions.handleTabClick' );
+			$p->set_attribute( 'data-wp-on--keydown', 'actions.handleTabKeyDown' );
+			$p->set_attribute( 'data-tab-index', $tab_label_index );
+			++$tab_label_index;
+		}
```
luisherranz: Better to use `elseif`'s here?

packages/block-library/src/tabs/view.js
luisherranz: Exactly. Please use the correct directives for all the DOM manipulations.

Here's a [good example of a block](https://github.com/WordPress/gutenberg/pull/64119) whose functionality is similar to this one. Take a special look at [the frontend file](https://github.com/WordPress/gutenberg/pull/64119/files#diff-3d35108b849bece54985a937368eaa12364bb6f97c17d08683541b6eeb38ce36): it only modifies the local context. All the DOM manipulations are done through directives that are linked to that local context. Also, notice how the local context is initialized on the server to ensure everything is ready before the JS downloads and executes.

This is not published yet, but maybe [this guide](https://github.com/WordPress/gutenberg/pull/63759), which is intended to explain this declarative approach, can help (I'd also appreciate any feedback regarding the guide itself 🙂).

creativecoder: @fabiankaegy and @luisherranz Thank you for the feedback!

Right now, by default, the tabs block is a list of on page links, and then progressively enhanced into tabs when JS is loaded.

How I can add the attributes on the server and still progressively enhance the block depending on if JS is loaded or not? Is there a way to use the interactivity API to add the role and aria attributes conditionally?

creativecoder:
Noting here that I'm going to be AFK for a couple of weeks--I plan to return to working on this as soon as I get back!

packages/block-library/src/tabs/view.js
luisherranz: > Is there a way to use the interactivity API to add the role and aria attributes conditionally?

Perhaps the guides that will be published here can help you understand how the Interactivity API works? 

> This is not published yet, but maybe [this guide](https://github.com/WordPress/gutenberg/pull/63759), which is intended to explain this declarative approach, can help (I'd also appreciate any feedback regarding the guide itself 🙂).

I know I'm repeating myself, but I think the accordion block example is really helpful for understanding how to program something like this, since the implementation of an accordion is quite similar to the implementation of tabs.

> Here's a [good example of a block](https://github.com/WordPress/gutenberg/pull/64119) whose functionality is similar to this one. Take a special look at [the frontend file](https://github.com/WordPress/gutenberg/pull/64119/files#diff-3d35108b849bece54985a937368eaa12364bb6f97c17d08683541b6eeb38ce36): it only modifies the local context. All the DOM manipulations are done through directives that are linked to that local context. Also, notice how the local context is initialized on the server to ensure everything is ready before the JS downloads and executes.

mikachan:
Just noting something that was mentioned in the recent [Hallway Hangout](https://make.wordpress.org/core/2024/07/09/hallway-hangout-lets-chat-about-whats-next-in-gutenberg-2/): Have we considered introducing the idea of [block states](https://github.com/WordPress/gutenberg/issues/57719) for the Tabs block (tab open/closed)?

packages/block-library/src/tabs/view.js
luisherranz: > This is not published yet, but maybe this guide, which is intended to explain this declarative approach, can help (I'd also appreciate any feedback regarding the guide itself 🙂).

The guide was published [here](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/core-concepts/the-reactive-and-declarative-mindset/).

packages/block-library/src/tab/edit.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Generates a slug from a tab's text label.
+ *
+ * @param {string} label    Tab label RichText value.
+ * @param {number} tabIndex Tab index value.
+ *
+ * @return {string} The generated slug with HTML stripped out.
+ */
+function slugFromLabel( label, tabIndex ) {
```
t-hamano, line 22: I think with this approach, if the same tab label exists in multiple tab blocks, we will get duplicate tab ids.

One approach would be to generate a unique ID using something like `wp_unique_id()` function when rendering on the server side, but I don't know how to share that ID between Tabs and the Tab block 🤔

packages/block-library/src/tab/edit.js
```diff
@@ -0,0 +1,68 @@
...
+
+	// Fall back to using the tab index if the label is empty.
+	return 'tab-panel-' + tabIndex;
+}
+
+export default function Edit( { attributes, clientId, setAttributes } ) {
+	const { anchor, isActive, label, slug, tabIndex } = attributes;
+	// Use a custom anchor, if set. Otherwise fall back to the slug generated from the label text.
+	const tabPanelId = anchor || slug;
+	const tabLabelId = tabPanelId + '--tab';
```
t-hamano, line 40: there is a possibility of duplicate IDs. Can't we use the `clientId` or the `useInstanceId()` hook?

packages/block-library/src/tab/save.js
```diff
@@ -0,0 +1,14 @@
...
+import { useBlockProps, useInnerBlocksProps } from '@wordpress/block-editor';
+
+export default function save( { attributes } ) {
+	const { anchor, slug } = attributes;
+	const tabPanelId = anchor || slug;
+
+	const blockProps = useBlockProps.save();
+	const innerBlocksProps = useInnerBlocksProps.save( blockProps );
+
+	return <section { ...innerBlocksProps } id={ tabPanelId } />;
```
t-hamano, line 13: Why a `section` element?

packages/block-library/src/tab/style.scss
```diff
@@ -0,0 +1,7 @@
+.wp-block-tab {
+	padding: 1em 0;
+
+	> *:first-child {
+		margin-top: 0;
+	}
```
t-hamano, line 6:
```suggestion
	> *:first-child {
		margin-top: 0;
	}

	> *:last-child {
		margin-bottom: 0;
	}
```

packages/block-library/src/tabs/edit.js
```diff
@@ -0,0 +1,183 @@
...
+	store as blockEditorStore,
+	RichText,
+} from '@wordpress/block-editor';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { useCallback, useEffect } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+const TABS_TEMPLATE = [
+	[ 'core/tab', { label: 'Tab 1' } ],
+	[ 'core/tab', { label: 'Tab 2' } ],
```
t-hamano, line 16:
```suggestion
	[ 'core/tab', { label: __( 'Tab 1' ) } ],
	[ 'core/tab', { label: __( 'Tab 2' ) } ],
```

packages/block-library/src/tabs/edit.js
```diff
@@ -0,0 +1,183 @@
...
+	] );
+
+	const blockProps = useBlockProps( { className: 'interactive' } );
+	const innerBlockProps = useInnerBlocksProps(
+		{
+			className: 'wp-block-tabs__content',
+		},
+		{
+			__experimentalCaptureToolbars: true,
+			clientId,
```
t-hamano, line 127: If I understand correctly, `clientId` is not defined as an option for the `useInnerBlocksProps` hook.

packages/block-library/src/tabs/save.js
```diff
@@ -0,0 +1,45 @@
...
+	const innerBlocksProps = useInnerBlocksProps.save( {
+		className: 'wp-block-tabs__content',
+	} );
+
+	return (
+		<div { ...blockProps }>
+			<h3 className="wp-block-tabs__title">
+				{ /* translators: Title for a list of content sections linked below. */ }
+				{ __( 'Contents' ) }
+			</h3>
```
t-hamano, line 23: The title appears to be hidden with CSS, so why is it needed? Also, hard-coded heading levels can lead to incorrect heading structures depending on the structure of the content.

packages/block-library/src/tabs/save.js
```diff
@@ -0,0 +1,45 @@
...
+
+			<ul className="wp-block-tabs__list">
+				{ innerTabs.map( ( tab, index ) => {
```
t-hamano, line 26: I think tab labels and tab panels that don't have a label should be hidden. Otherwise they will render an `a` element with no text:

image

t-hamano (commented):
Thank you for working through this block.

I've left comments to the best of my understanding, but please let me know if I've misunderstood something 👍

t-hamano:
Another thing I noticed is that when there are no tabs, nothing is displayed. It might be nice to show the appender when the Tabs block is empty.

image

packages/block-library/src/tabs/view.js
creativecoder: @fabiankaegy, @luisherranz, @DAreRodz In https://github.com/WordPress/gutenberg/pull/63689/commits/cc304b55ad47d7b63985e39d965f420023c2f44d, I've removed the `initTabs` callback in favor of server-side directives and state getters. This is less code overall, so thank you for the guidance.

One thing clarify is that the tabs front-end here has been built as progressively enhanced markup, as suggested by the likes of https://design-system.service.gov.uk/components/tabs/ and https://inclusive-components.design/tabbed-interfaces/. Without JS, it's simply a list of on page links:

Screen Recording 2024-08-29 at 11 21 01

This means that I'm not adding attributes directly with the interactivity API (`$p->set_attribute( 'role', 'tab' );`), but instead using the `wp-bind` directive (`$p->set_attribute( 'data-wp-bind--role', 'state.roleAttribute' );`) with a getter in the `view.js` state that returns the correct role based on the element's class name. The `role`, `tabindex`, and `aria-` attributes are semantically incorrect if JS is not loaded on the page, as there is no tabbed interface, so I don't think they should be present directly in the server-side markup.

For attributes with dynamic values this works well. It would also be nice to have a way to do this more simply with static values. For example, `$p->set_attribute( 'data-wp-bind--role', 'tablist` )` doesn't work right now, but this would simplify the code even more, as the role property never changes, it just needs to be added when JS is initialized.

packages/block-library/src/tabs/view.js
creativecoder: This has been removed in https://github.com/WordPress/gutenberg/pull/63689/commits/cc304b55ad47d7b63985e39d965f420023c2f44d

packages/block-library/src/tabs/index.php
creativecoder: Done in https://github.com/WordPress/gutenberg/pull/63689/commits/cc304b55ad47d7b63985e39d965f420023c2f44d

packages/block-library/src/tab/save.js
creativecoder: Using `<section>` elements for tab panels is recommended here: https://inclusive-components.design/tabbed-interfaces/

packages/block-library/src/tabs/save.js
creativecoder: This was also recommended by https://inclusive-components.design/tabbed-interfaces/. Without front-end JS, the tabs block a list of on page links; this heading makes sure the list has a title, but is hidden when JS is loaded and the list is enhanced into a tabbed interface.

> Also, hard-coded heading levels can lead to incorrect heading structures depending on the structure of the content.

That's a good point. Alternatively `role="heading"` could be used here, but I'm not sure that's any better--https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role indicates it defaults to heading level 2.

jffng:
@creativecoder just a heads up that since #65064, you'll need to do something like this to load the view script for the tabs block: https://github.com/WordPress/gutenberg/pull/64119/commits/84112e36d5bbc8ba9e6df921bd20dcb1e55906a4

creativecoder:
Closing this, as I won't be able to work on it further.

Issue #63336, opened 2024-07-10
Consider adding a clear fix for floated blocks


carolinan:
### Description

In block themes, when the post content contains a floated left or right aligned block, these blocks can be positioned incorrectly and end up next to a block that is outside the post content.

Here is a screenshot of the page template in Twenty Twenty-Four, where the image is the only block in the content, and is left aligned. The image is incorrectly positioned next to the footer.
<img width="1508" alt="A left aligned image which is floated next to the site footer." src="https://github.com/WordPress/gutenberg/assets/7422055/53f8a60c-b1d0-45fd-ba70-cfc376754e6b">

If you would remove the spacer block that is below the content in the post template in the same theme, the same problem would appear.

Can some sort of clear instead be added to the post content? Right now, this problem must be solved by the theme developer/designer/user, and it would be an improvement if the block handled the clear.
From doing some quick searching there does not seem to be an ideal solution for this problem.

https://stackoverflow.com/questions/211383/what-methods-of-clearfix-can-i-use

### Step-by-step reproduction instructions

Activate Twenty Twenty-Four or  Twenty Twenty-Three.
Create a new page.
Add an image and select the left alignment.
Save and view the front of the page.
Notice the position of the image.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

carolinan:
Pinging @WordPress/block-themers for feedback.

MaggieCabrera:
Would you suggest the `:after` version of the clearfix? I wonder if we could apply it only if there are floating elements within the post content... Sounds like it's a better thing to have than not to have, to me

colorful-tones:
I’m not at my computer and I wonder: why are we only discussing the Post Content? Can folks still place floated images in a footer or header too and have the appearance of a broken layout?

MaggieCabrera:
> I’m not at my computer and I wonder: why are we only discussing the Post Content? Can folks still place floated images in a footer or header too and have the appearance of a broken layout?

I think that's much harder to manage, they would have to add a paragraph block and insert the images there, right?

carolinan:
Footer and headers can contain anything, so there is no common block to place the clear fix on.

colorful-tones:
> Footer and headers can contain anything, so there is no common block to place the clear fix on.

[...]

PR #63690, opened 2024-07-18
Try: Post content: Add clearfix

carolinan:
## What?

This PR adds a clearfix to the post content block, so that left and right aligned blocks in the content do not overflow the content area. 

## Why?

If the last item in the post content is left or right aligned, it can be incorrectly positioned left or right of the next element, even if it is outside the designated content area. For example to the left of the site footer instead of above it. You can see examples of this in the issue:
https://github.com/WordPress/gutenberg/issues/63336

## How?

The PR adds a new CSS file to the post content block and adds a clearfix.

## Testing Instructions
Activate Twenty Twenty-Four or Twenty Twenty-Three.
Create a new **page**. (It has to be the page template, not a post)
Add an image and select the left alignment.
Save and view the front of the page.
Notice that the image remains in the "content area".

## Screenshots or screencast 

Before:
A left aligned image block before the PR, positioned left of the footer

After:
66 local_clear-test_

### DIFF
--- a/packages/block-library/src/post-content/block.json
+++ b/packages/block-library/src/post-content/block.json
@@ -54,5 +54,6 @@
 			}
 		}
 	},
+	"style": "wp-block-post-content",
 	"editorStyle": "wp-block-post-content-editor"
 }
--- a/packages/block-library/src/post-content/style.scss
+++ b/packages/block-library/src/post-content/style.scss
@@ -1,0 +1,5 @@
+.wp-block-post-content::after {
+	content: "";
+	display: table;
+	clear: both;
+}
--- a/packages/block-library/src/style.scss
+++ b/packages/block-library/src/style.scss
@@ -29,6 +29,7 @@
 @import "./paragraph/style.scss";
 @import "./post-author/style.scss";
 @import "./post-comments-form/style.scss";
+@import "./post-content/style.scss";
 @import "./post-date/style.scss";
 @import "./post-excerpt/style.scss";
 @import "./post-featured-image/style.scss";


carolinan:
Pinging @WordPress/block-themers for review.  I don't want to break any legit use cases, but I can't think of any where adding a clear to the post content block will break existing, deliberate designs.

MaggieCabrera (approved):
I'm giving my approval here because I can't think of cases where we would be breaking people's sites. I tagged @jasmussen in case he can think of something we couldn't. In any case I think this is a good one to merge early and let it in the open so it can be tested, and it's an easy revert anyway

colorful-tones:
This is working fine in my testing and no odd side effects that I can infer. ✅ 

One small nit-pick for me is the testing instructions:

> Create a **new page**. (It has to be the page template, not a post)

I don't know why we're emphatically clarifying it be a **page**. We may be aiming to simplify the testing instructions, but _posts will also be impacted_. I tested a custom single post template in the Frost theme, and the fix still addresses the issue. 👍 

I do not want to mislead any new tester/contributor to only focus on pages.

Also, I encourage us to test the header/footer (template parts in general) to come up with a plan, too, assuming we want to continue with the proactive float defense strategy. Of course, this would be a follow-up issue and should not hinder merging the work proposed here.

Thanks for the wonderful work here @carolinan 

Frost theme single post custom template with float applied to image

colorful-tones (approved):
LGTM! 🫶

carolinan:
> I don't know why we're emphatically clarifying it be a page.

Because these two example themes already clears the post content block in the template for the single post. So testing it on the single post would not have shown that the problem was solved by the PR.

colorful-tones:
> Because these two example themes already clears the post content block in the template for the single post. So testing it on the single post would not have shown that the problem was solved by the PR.

Yes, this is what I found in my testing of Twenty-Something themes. However, let’s not encourage folks to test solely against Twenty-Something themes and allow for robust testing against as many scenarios as possible for the best coverage. A user could create a new post template in the Site Editor and remove everything after the Post Content. It would likely be a “power” user, but it is possible nonetheless.

carolinan:
That the existing templates in these themes work differently is also explained in the issue. I am not going to choose to instruct a tester to create their own custom theme, or install a third party theme, if the themes that are already included in WordPress can be used to replicate the issue and the solution.

jasmussen:
The use case here seems valid enough, and the fix is bog standard. It's also a good time in the cycle to try this: early, so we can get plugin feedback.

I wish there was a better fix than the pseudo selector one, though: while we aim to reduce the use of pseudo selectors in the editor, they are still occasionally used for showing some UI, like focus styles or selected states. But I think that's the "before" selector, but something to check.

carolinan:
Yeah neither alternatives, including the one used, are perfect. :(

wongjn:
> The use case here seems valid enough, and the fix is bog standard. It's also a good time in the cycle to try this: early, so we can get plugin feedback.
> 
> I wish there was a better fix than the pseudo selector one, though: while we aim to reduce the use of pseudo selectors in the editor, they are still occasionally used for showing some UI, like focus styles or selected states. But I think that's the "before" selector, but something to check.

This may conflict with the block selection ring in the editor, which uses the same `::after` pseudo-element as used in this PR:

https://github.com/WordPress/gutenberg/blob/98c65eaf268b86175d51cbb91c35e4111a7950ef/packages/block-editor/src/components/block-list/content.scss#L76-L90

Although, the `selected-block-focus` mixin applies `position: absolute`:

https://github.com/WordPress/gutenberg/blob/98c65eaf268b86175d51cbb91c35e4111a7950ef/packages/base-styles/_mixins.scss#L597-L606

So if the `core/post-content` block is cleared a differently in the block editor, then we're all good.

t-hamano:
@wongjn Thanks for the report. Certainly, this PR conflicts with the block's selection ring.

Clearfix enabled:

enabled

Clearfix disabled:

disabled

fabiankaegy:
Hey @carolinan  👋 

Thanks in advance :)

carolinan:
A clearfix has been added to the post content block. Elements at the end of the content that are left- or right-aligned will no longer overflow and be positioned outside the content area. The clearfix was implemented using an `::after` pseudo-element:
```
.wp-block-post-content::after {
	content: "";
	display: table;
	clear: both;
}
```
Developers who have already implemented their own clearfix may need to test for duplicate or conflicting code.

t-hamano:
@carolinan

The pseudo-element approach had issues with focus in the editor, so I fixed it in #65364. Instead of the `::after` pseudo-element, it uses `display: flow-root`.

So maybe a dev note isn't necessary.

carolinan:
@t-hamano Oh OK, thank you!

PR #63691, opened 2024-07-18
Update pagination design with mininal variant for select control

### LINKED ISSUE #63675: Update pagination design
Screenshot 2024-07-10 at 18 06 02

With the new `minimal` variant of `SelectControl` (https://github.com/WordPress/gutenberg/pull/63265) we can streamline the design of the pagination UI in data views.

dhananjaykuber:
Fixes [#63675](https://github.com/WordPress/gutenberg/issues/63675)

## What?
Update the pagination UI design in data views using the new minimal variant of SelectControl.

## Why?
This PR aims to streamline and modernize the pagination UI in data views by leveraging the recently introduced minimal variant of SelectControl (#63265). This change will improve visual consistency and provide a cleaner, more minimal interface for users navigating through paginated content.

## How?
The implementation involves adding the `variant="minimal"` prop to the SelectControl component used in the pagination UI.

## Screenshots or screencast 
Screenshot 2024-07-18 at 12 38 56 PM

### DIFF
--- a/packages/dataviews/src/pagination.tsx
+++ b/packages/dataviews/src/pagination.tsx
@@ -75,6 +75,7 @@ const Pagination = memo( function Pagination( {
 										} );
 									} }
 									size="compact"
+									variant="minimal"
 									__nextHasNoMarginBottom
 								/>
 							),


jameskoster:
Thanks for the PR. 

This change gets us part of the way there, but ideally the full text should be the label, like the mockup in https://github.com/WordPress/gutenberg/issues/63675.

One challenge is that applying `text-transform: uppercase` to the select also affects the `option`s, which is undesirable. I wonder if there's a way around that, or whether we should use regular text styling.

ntsekouras:
Thanks for the PR! I'll close it though since it has been addressed here: https://github.com/WordPress/gutenberg/pull/63815

Issue #63646, opened 2024-07-17
DataViews: Split the components into composable components.


youknowriad:
Part of https://github.com/WordPress/gutenberg/issues/55083

Right now, when you render the DataViews component, it gets rendered all in one including:

 - Search bar
 - Filters
 - View Config
 - Layout switcher
 - Pagination
 - Layouts

This approach provides a straightforward way to use DataViews but lacks flexibility. For instance in https://github.com/WordPress/gutenberg/issues/55101 we want to add a "toggle" to show the dataviews sidebar at the right of the view config dropdown. That toggle doesn't have anything to do with DataViews so it shouldn't rendered within the dataviews component.

We have two options:

**Short Term option**

Just have some kind of slot prop to inject UI there. I'll probably use this approach initially.

**Long Term option**

Expose the underlying DataViews component and allow composition:

```
<DataViews>
    <HStack>
        <DataViews.Search />
        <DataViews.Filters />
        <DataViews.BulkActions />
        <DataViews.LayoutSwitcher />
        <DataViews.ViewConfig />
        <Button>My custom Button</Button>
     </HStack>
      <DataViews.Layout />
      <DataViews.Pagination />
      <DataViews.BulkActionToolbar />
</DataViews>
```

Any thoughts @ntsekouras @ellatrix @oandregal @jorgefilipecosta 
Also components team as they're exploring a similar pattern for the components package. @ciampo @mirka @DaniGuardiola

ciampo:
Adopting compound components seems like a good idea to me, to offer an a-la-carte set of components that consumers of the package can pick and render as needed.

Other approaches that can be used (not necessarily as an alternative, but in conjunction) are render props, which allow components to identify "slots" where consumers can render custom react components.

Related, this recent article from @DaniGuardiola explores the topic in depth: https://dio.la/article/the-everything-bagel-of-components

ntsekouras:
I think composable components are good and flexible. This actually was my approach at my [very first experiment](https://github.com/WordPress/gutenberg/pull/53906) of this component.

youknowriad:
Going to give this a try.

PR #63694, opened 2024-07-18
DataViews: Refactor to prepare exposing the underlying UI pieces

youknowriad:
Related #63646 

## What?

This PR refractors the DataViews components to prepare them to be exposed (in order to address the need for flexibility in how we render the dataviews).

This PR does the following things:

 - Introduces a `DataViewsContext` React context that is exposed at the top level. `DataViews` component and consumed by the underlying composable components. That way the user won't have to pass the same props all over again to all the composable components.
 - Moves all the dataviews components into `dataviews/src/components` folder.
 - Split the giant `dataviews/style.scss` file and move the declarations close to each component.

In a follow-up PR, I will be exposing the underlying components to allow users to adapt the dataviews UI to their liking.

## Testing Instructions

There shouldn't be any change in behavior but because of the scale of this change, some small regressions can happen so I'd appreciate testing the different dataviews that we have.


jorgefilipecosta (commented):
When I try to use the duplicate pattern action I get this:
Screenshot 2024-07-18 at 15 01 21

I think it is a z-index issue.

jorgefilipecosta (commented):
I could not select patterns at all (the selectable ones are user-created patterns under my patterns). The selection checkbox just does not work for patterns.

jorgefilipecosta (commented):
When sorting patterns by title the application crashes.

packages/dataviews/src/components/dataviews-context/index.ts
```diff
@@ -0,0 +1,46 @@
...
+	};
+	selection: string[];
+	onChangeSelection: SetSelection;
+	openedFilter: string | null;
+	setOpenedFilter: ( openedFilter: string | null ) => void;
+	getItemId: ( item: Item ) => string;
+};
+
+const DataViewsContext = createContext< DataViewsContextType< any > >( {
+	view: { type: 'table' },
```
jorgefilipecosta: Could we use the constant LAYOUT_TABLE?

youknowriad:
> When sorting patterns by title the application crashes.
> I could not select patterns at all (the selectable ones are user-created patterns under my patterns). The selection checkbox just does not work for patterns.

Both of these issues are present in trunk, I will fix them in separate PRs.

> When I try to use the duplicate pattern action I get this:

I'm not entirely sure how to duplicate this, I suspect that it's also independent of this refactor.

jorgefilipecosta:
> I'm not entirely sure how to duplicate this, I suspect that it's also independent of this refactor.

For me in trunk it is not happening:
Screenshot 2024-07-18 at 16 55 14
This branch:
Screenshot 2024-07-18 at 16 56 55

To replicate I just go patterns `wp-admin/site-editor.php?postType=wp_block`and try to use the duplicate action.
This issue is not just on duplicate it affects all the actions with a modal, e.g: rename etc. Even delete I can not trigger it on the patterns page, clicking the button to confirm does not work in the trunk it does. It should be for single actions on the ellipsis menu not for bulk actions.
I'm not sure if the trunk has some issue already but on trunk, this modal issue seems to not happen.

ntsekouras:
Opened a quick [PR](https://github.com/WordPress/gutenberg/pull/63710) for the title sorting bug noticed here.

ntsekouras:
The `reset` button renders in this PR when it shouldn't.

Screenshot 2024-07-18 at 5 31 40 PM


----

Top checkbox for selecting all is bigger.

Screenshot 2024-07-18 at 5 37 30 PM

packages/dataviews/src/index.ts
```diff
@@ -1,5 +1,5 @@
-export { default as DataViews } from './dataviews';
+export { default as DataViews } from './components/dataviews';
```
ntsekouras, line 1: What advantages do you see with adding an extra folder layer with `components` folder? Additionally every sub folder here has the `dataviews` prefix. Do you think we need this  since we're in dataviews package?

packages/dataviews/src/components/dataviews-bulk-actions/index.tsx
```diff
@@ -288,3 +282,19 @@ export default function BulkActions< Item >( {
...
 	);
 }
+
+export default function BulkActions() {
+	const { data, actions = [], view } = useContext( DataViewsContext );
+	const hasPossibleBulkAction = useSomeItemHasAPossibleBulkAction(
+		actions,
+		data
+	);
+	if (
```
ntsekouras, line 292: It feels like these checks should be the consumer's responsibility, or checked by a passed prop or something.. Or having a flag for layouts whether they support bulk actions, etc.. Probably my last suggestion is my preference. 

Same goes for `BulkActionsToolbar`.

packages/dataviews/src/components/dataviews/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+	const _selection = useMemo( () => {
+		return selection.filter( ( id ) =>
+			data.some( ( item ) => getItemId( item ) === id )
+		);
+	}, [ selection, data, getItemId ] );
+
+	return (
+		<DataViewsContext.Provider
+			value={ {
+				view,
```
ntsekouras, line 86: Should we memo this?

packages/dataviews/src/index.ts
youknowriad: > What advantages do you see with adding an extra folder layer with components folder? 

- We have another folder called "layouts" for the default layouts
- We'll have more folders later (field types)
- Components might not be the perfect folder name here but I like that all the components that make up the dataviews and data forms are now in a folder to avoid being mixed with "layouts", "field types"...

> Do you think we need this since we're in dataviews package?

Yes, because we have. `DataViews` and `DataForm`. DataForm can also potentially be composed later like the DataViews.

packages/dataviews/src/components/dataviews-bulk-actions/index.tsx
youknowriad: For the "list" view check I could agree with you but I think it's easy to revert, for now, there's a decision that list views don't support bulk actions and it's consistent across all dataviews.

For the other check, I disagree, it's something that should be bundled in the component.

packages/dataviews/src/components/dataviews/index.tsx
youknowriad: We could but honestly I'm not sure how valuable it is because it's basically all of the props of the component.

ntsekouras (commented):
This looks good in general, thank you! There are some `z-index` issues with actions modals and some styling issues, but it needs more smoke testing to catch them all. I'd be fine to land sooner with quick follow ups if we find more style issues, since it's a big PR and will be a nightmare to rebase all the time.

For me the most important things to decide on is the file structure and maybe the prefixes of folders, and the BulkActions comment regarding conditional rendering per layout.

packages/dataviews/src/index.ts
ntsekouras: >Components might not be the perfect folder name here but I like that all the components that make up the dataviews and data forms are now in a folder to avoid being mixed with "layouts", "field types"

I'm fine with it, although my preference for this package is to organise without the components. I think it makes it easier find files related per 'function'. What I mean by this, is have the folder `filters` for example and add there all related code (filters, global search, etc..).

packages/dataviews/src/components/dataviews-bulk-actions/index.tsx
ntsekouras: >For the other check, I disagree, it's something that should be bundled in the component.

Which one? For my preference layouts to have a prop like `supportsBulkActions`?

packages/dataviews/src/index.ts
youknowriad: Other than filters, the other components don't really have sub components, so a folder doesn't really make sense IMO.

packages/dataviews/src/components/dataviews-bulk-actions/index.tsx
youknowriad: > Which one? For my preference layouts to have a prop like supportsBulkActions?

Why a prop, it's something that can be guessed from the actions, the prop seems redundant to me.

packages/dataviews/src/components/dataviews-filters/index.tsx
```diff
@@ -120,6 +109,6 @@ function _Filters< Item >( {
 }
 
 // A type assertion is used here to keep the type argument.
-const Filters = memo( _Filters ) as typeof _Filters;
```
jorgefilipecosta, line 123: We are removing the type assertion but not removing the comment about it.

jorgefilipecosta (commented):
The changes seem, good I guess we should just fix the styling issues, and it is ready to merge to avoid big conflicts.

packages/dataviews/src/components/dataviews-bulk-actions/index.tsx
ntsekouras: How it can be guessed by the actions?

youknowriad: Like today: `useSomeItemHasAPossibleBulkAction` hook is used here to know whether to render this or not.

ntsekouras: Exactly, but here we don't want to render bulk actions not by the items and whether are eligible. We just don't want to render them for a specific layout.

youknowriad: There's two checks here: the layout which I agree that could be a user decision If needed. But right now it's not a user decision, so this PR keeps it as is.

and the other check `useSomeItemHasAPossibleBulkAction` is not a user decision, it's bundled here as well and should always stay as is IMO.

If I'm guessing right from your comment, you think the "layout" list not showing bulk actions should be a user decision? Maybe, that's fine by me, but it's not the case in trunk and I don't want to change that here.

ntsekouras: Yeah, sorry for the confusion. I was only referring to `layout`. We can update later the layout checks then.

youknowriad:
I fixed the z-index issue, I was interestingly an issue because we were reusing classnames rather than reusing components. That's a reminder to avoid that as much as possible :)

ntsekouras (approved):
I think this is good to land and have quick follow ups if needed to avoid the conflicts.

Still needs the rebase and for sure the:
>Top checkbox for selecting all is bigger.

isn't fixed right now, so it could be either fixed here or in the first follow up.

Thanks!

youknowriad:
> Top checkbox for selecting all is bigger.

Forgot about this one, I'll look into it.

youknowriad:
> Top checkbox for selecting all is bigger.

This was another case of "classname" reuse instead of "component reuse". Let's stop doing that please.

PR #39227, opened 2022-03-04
Fix Comments blocks pagination

SantosGuillamot:
## What
Make comments pagination links work as they should be when we have these conditions:
 - Comment Pagination enabled on Discussion Settings
 - Comment Query Loop "Inherit option" enabled (current status by default)

## Why
Both "Older Comments" and "Newer Comments" are not working as expected inside the `Comment Query Loop`.
If we enable pagination on Discussion Settings, before the PR we have:

- Comments are displayed in the correct order ✅
- Pagination is incorrect ❌

Older/Newer Comments URL, depending on what we have chosen on `Discussion Settings` (first or last page as the default for Comments Pagination), is not pointing to the correct URL. It makes comments pagination navigation inconsistent.

The reason for the bug is that we don't have the query var `cpage` defined and we are creating a `WP_Comment_Query( $comment_args )` with pagination arguments. In this case, the pagination link functions require this query var in order to know in which page of the comments pagination is the user when she lands on the default page/post view.

## How

Defining a default value for `cpage`, being the maximum number of pages or 0, depending on the Discussion setting chosen (show the first or the last page).

This way, when the user lands on a page/post without the comment pagination defined, it will land with the default value, so comments pagination links can work as expected.

## Testing Instructions
There are mainly two cases to test here:
1. Check that the "Older Comments" and "Newer Comments" link are shown correctly when the "Last page displayed by default" setting is enabled as explained in [this comment](https://github.com/WordPress/gutenberg/issues/37560#issuecomment-1042768856).
2. Check that the penultimate page links are working properly as explained in [this comment](https://github.com/WordPress/gutenberg/issues/37560#issuecomment-1059395812).

## Types of changes
Bug fix

PR #63698, opened 2024-07-18
Pass the comments query `paged` arg to functions  `get_next_comments_link` and `get_previous_comments_link`

SantosGuillamot:
## What?
According to [this report](https://core.trac.wordpress.org/ticket/60806), the solution introduced in [this pull request](https://github.com/WordPress/gutenberg/pull/39227) to solved the mentioned bug, where `cpage` was updated based on the query, is causing problems with plugins relying on it being `0` or undefined.

This pull request explores another alternative to solve the initial bug without modifying `cpage.`

It needs to be tested with [this other core pull request](https://github.com/WordPress/wordpress-develop/pull/7275), which modifies the relevant functions to accept the new parameter.

## How?
Pass the `paged` argument from the query to [get_next_comments_link](https://github.com/WordPress/wordpress-develop/blob/dea027eac82ca7244eff421642562b2d75918e19/src/wp-includes/link-template.php#L3103C10-L3103C32) and [get_previous_comments_link](https://github.com/WordPress/wordpress-develop/blob/dea027eac82ca7244eff421642562b2d75918e19/src/wp-includes/link-template.php#L3171) so they only read the `cpage` when it is not defined. A check like the following must be added in the core functions:

```php
	if ( empty( $page ) ) {
		$page = get_query_var( 'cpage' );
	}
```

## Testing Instructions
It needs to be tested with [this other core pull request](https://github.com/WordPress/wordpress-develop/pull/7275), which modifies the relevant functions to accept the new parameter.
1. Add comments to a post.
2. Go to Settings -> Discussion and set "Break comments into pages with 2 top-level comments per page and the 
**last** page displayed by default.
3. Go to the post and check that the pagination links are correct.

### DIFF
--- a/packages/block-library/src/comments-pagination-next/index.php
+++ b/packages/block-library/src/comments-pagination-next/index.php
@@ -37,7 +37,7 @@ function render_block_core_comments_pagination_next( $attributes, $content, $blo
 		$label .= $pagination_arrow;
 	}
 
-	$next_comments_link = get_next_comments_link( $label, $max_page );
+	$next_comments_link = get_next_comments_link( $label, $max_page, $comment_vars['paged'] );
 
 	remove_filter( 'next_posts_link_attributes', $filter_link_attributes );
 
--- a/packages/block-library/src/comments-pagination-previous/index.php
+++ b/packages/block-library/src/comments-pagination-previous/index.php
@@ -29,7 +29,8 @@ function render_block_core_comments_pagination_previous( $attributes, $content, 
 	};
 	add_filter( 'previous_comments_link_attributes', $filter_link_attributes );
 
-	$previous_comments_link = get_previous_comments_link( $label );
+	$comment_vars           = build_comment_query_vars_from_block( $block );
+	$previous_comments_link = get_previous_comments_link( $label, $comment_vars['paged'] );
 
 	remove_filter( 'previous_comments_link_attributes', $filter_link_attributes );
 


gziolo:
Is there a similar fix required for the `paginate_comments_links` in Comments Pagination Numbers block?

It looks like it also should get covered as it calls `get_query_var`:

https://github.com/WordPress/wordpress-develop/blob/c8b744d12b2af5bc8eb01b2e8782c20d681fdadd/src/wp-includes/link-template.php#L3249

SantosGuillamot:
> Is there a similar fix required for the paginate_comments_links in Comments Pagination Numbers block?

I believe you are right and we should cover `paginate_comments_links` as well if we decide to follow this approach.

gziolo:
@ockham, can you confirm this approach would work from your perspective? As far as I'm concerned, it might resolve the issue.

ockham:
Apologies for the late reply. Yeah, I think making the argument explicit should be the way forward to make this work. We will have to implement the "other side" -- i.e. the changes to `get_next_comments_link` and `get_previous_comments_link` in Core -- to know for sure, though.

@SantosGuillamot Do you have some bandwidth to take that on? I'm a bit swamped with other things these days 😅

SantosGuillamot:
I have created [this alternative pull request](https://github.com/WordPress/wordpress-develop/pull/7275) in core, which removes the problematic `set_query_var` while keeping the pagination blocks working as expected.

That PR plus this one should solve both problems.

> Is there a similar fix required for the paginate_comments_links in Comments Pagination Numbers block?

Regarding this, I've been taking a deeper look and I believe it is not needed. Let me explain my reasoning:
* In the pagination numbers block, we are populating a query with `current` storing the page: [link](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/comments-pagination-numbers/index.php#L34).
* `paginate_comments_links` is using `get_query_var`, but only to define the "defaults": [link](https://github.com/SantosGuillamot/wordpress-develop/blob/dee095d8071ece170b3fe5113aeb01c742e5d55e/src/wp-includes/link-template.php#L3262-L3270).
* Those defaults are overridden with the values passed in the `args`, which contains the proper page: [link](https://github.com/SantosGuillamot/wordpress-develop/blob/dee095d8071ece170b3fe5113aeb01c742e5d55e/src/wp-includes/link-template.php#L3275).

packages/block-library/src/comments-pagination-next/index.php
```diff
@@ -37,7 +37,7 @@ function render_block_core_comments_pagination_next( $attributes, $content, $blo
 		$label .= $pagination_arrow;
 	}
 
-	$next_comments_link = get_next_comments_link( $label, $max_page );
+	$next_comments_link = get_next_comments_link( $label, $max_page, $comment_vars['paged'] );
```
ockham, line 40: As mentioned in https://github.com/WordPress/wordpress-develop/pull/7275#issuecomment-2327066295, we need to add some safeguards to make sure that we're not passing this extra argument if `get_next_comments_link()` -- as provided by whatever WordPress version is running -- doesn't support the additional parameter yet.

To do so, I think our best course of action is PHP's [`ReflectionFunction`](https://www.php.net/manual/en/class.reflectionfunction.php):

Something like (untested):

```suggestion
	$get_next_comments_link_reflection = new ReflectionFunction( 'get_next_comments_link' );

	if ( $get_next_comments_link_reflection->getNumberOfParameters() >= 3 ) {
		$next_comments_link = get_next_comments_link( $label, $max_page, $comment_vars['paged'] );
	} else {
		$next_comments_link = get_next_comments_link( $label, $max_page );
	}
```

and accordingly for `get_previous_comments_link`.

SantosGuillamot: Do we really need to add this safety check? If I am not mistaken, when passing the third argument to a function that only accepts two, PHP will just ignore it, right? At least, that's what I can see from my testing.

gziolo: There are functions like [func_get_args](https://www.php.net/manual/en/function.func-get-args.php) that allow to handle any number of params, so as far as I tested it, providing an additional param that isn't handled works just fine.

```php
<?php
function foo()
{
    $numargs = func_num_args();
    echo "Number of arguments: $numargs \n";
    if ($numargs >= 2) {
        echo "Second argument is: " . func_get_arg(1) . "\n";
    }
    $arg_list = func_get_args();
    for ($i = 0; $i < $numargs; $i++) {
        echo "Argument $i is: " . $arg_list[$i] . "\n";
    }
}

foo(1, 2, 3);
?>
```

ockham: Ah, fair enough! I was assuming that PHP would at least issue a warning or a notice if strict error reporting was enabled, but it seems that that isn't the case. (Here's a related [SO thread](https://stackoverflow.com/questions/7928707/why-does-php-not-throw-an-error-when-i-pass-too-many-parameters-to-a-function).)

gziolo: Yes, it fatals on built-in functions I tested, but not on custom functions 🤷🏻

SantosGuillamot: I tested it directly with `get_next_comments_link` in a branch without the changes and it doesn't seem to break.

ockham (approved):
Thank you! LGTM 👍

michalczaplinski:
I found an issue that was likely caused by changes in this PR: https://github.com/WordPress/gutenberg/issues/65424

ockham:
> I found an issue that was likely caused by changes in this PR: #65424

I just came here to report the same thing 😕

SantosGuillamot:
I've started a PR trying to solve that: https://github.com/WordPress/gutenberg/pull/65435

PR #63706, opened 2024-07-18
Fix: The Copy button of the PostURL shouldn't be placed in the slug input

### LINKED ISSUE #63700: The Copy button of the PostURL shouldn't be placed in the slug input
### Description

Noticed while working on https://github.com/WordPress/gutenberg/issues/61196 / https://github.com/WordPress/gutenberg/pull/63669

In the PostURL popover, there is a `Copy` button that is visually placed within the slug input field. As such, it is visually associated with the slug value. Screenshot:

Screenshot 2024-07-18 at 13 07 48

As a user, I would expect the Copy button to copy the value that is inside the input field, because that's what the UI is showing me.

Instead, the Copy button actually copies the full permalink that is shown _below_ the input field.

Following the example in the screenshot above:
- The input field shows `/my-post-title`.
- The copy button is labeled `Copy` and is visually associated with the input field.
- Clicking the button copies the full permalink `http://localhost:8889/my-post-title/` instead.

I'm not sure this is a good UI as it is potentially confusing for users. Also, it is inconsistent with other UIs where the Copy button actually copies the value shown in the input field.

See above.

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

Rishit30G:
## What?

Fixes #63700

## Why?

As a user, one would expect the Copy button to copy the value inside the input field, because that's what the UI shows

## How?

Change the first parameter in the `useCopyToClipboard` hook to `/ + postSlug` from `permalink`

## Testing Instructions
- Create a new post 
- From settings change permalink to `Post Name`
- Now go back to the post and on the right panel click on the value part of `Link`
- Copy the value of the slug by pressing the copy button 
- Verify that the value now copied is the slug and not the link to the post 

Same as above 

## Screenshots or screencast 

https://github.com/user-attachments/assets/bdd0b1b2-6f80-4f80-a0ab-0ed2796a43dc

### DIFF
--- a/packages/components/src/external-link/index.tsx
+++ b/packages/components/src/external-link/index.tsx
@@ -8,13 +8,14 @@ import type { ForwardedRef } from 'react';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { forwardRef } from '@wordpress/element';
+import { Fragment, forwardRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import type { ExternalLinkProps } from './types';
 import type { WordPressComponentProps } from '../context';
+import { Suffix } from '../input-control/styles/input-control-styles';
 
 function UnforwardedExternalLink(
 	props: Omit<
@@ -23,7 +24,14 @@ function UnforwardedExternalLink(
 	>,
 	ref: ForwardedRef< HTMLAnchorElement >
 ) {
-	const { href, children, className, rel = '', ...additionalProps } = props;
+	const {
+		href,
+		children,
+		className,
+		rel = '',
+		suffix,
+		...additionalProps
+	} = props;
 	const optimizedRel = [
 		...new Set(
 			[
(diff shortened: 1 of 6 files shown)


afercia:
@Rishit30G thank you for your PR. Sorry I wasn't clear. I didn't suggested to change _what_ is copied.
I meant that the Copy button is misplaced. It should be moved close to the value that is actually copied: the full permalink shown below the input field.

afercia (changes requested):
See comment.

Rishit30G:
Hey @afercia! 
Thanks for sharing your feedback and insights 🙇 
Let me know if this is what was expected:

https://github.com/user-attachments/assets/cba989ea-8fa7-4449-963f-5f35568adf3f

packages/components/src/external-link/index.tsx
```diff
@@ -54,28 +62,33 @@ function UnforwardedExternalLink(
...
+						className="components-external-link__icon"
+						aria-label={
+							/* translators: accessibility text */
+							__( '(opens in a new tab)' )
+						}
+					>
+						&#8599;
+					</span>
+				</a>
+				<div>{ suffix && <Suffix>{ suffix }</Suffix> }</div>
```
afercia, line 89: is the `<div>` necessary?

packages/components/src/external-link/types.ts
```diff
@@ -12,4 +12,9 @@ export type ExternalLinkProps = {
 	 * The URL of the external resource.
 	 */
 	href: string;
+	/**
+	 * Allows for markup other than icons or shortcuts to be added to the menu item.
```
afercia, line 16: This description should be adjusted, as this component isn't s 'menu item'.

packages/components/src/input-control/index.tsx
```diff
@@ -101,7 +99,7 @@ export function UnforwardedInputControl(
 					onKeyDown={ onKeyDown }
 					onValidate={ onValidate }
 					paddingInlineStart={ prefix ? space( 2 ) : undefined }
-					paddingInlineEnd={ suffix ? space( 2 ) : undefined }
+					paddingInlineEnd={ undefined }
```
afercia: I'm not sure the InputControl should be chaned. This is a base component that can be reused of other purposes and by other components.

packages/components/src/input-control/input-base.tsx
```diff
@@ -132,11 +130,6 @@ function InputBase(
 						</Prefix>
 					) }
 					{ children }
-					{ suffix && (
-						<Suffix className="components-input-control__suffix">
-							{ suffix }
-						</Suffix>
-					) }
```
afercia, line 129: I'm not sure the InputBase component should be changed. See above.

packages/components/src/external-link/index.tsx
Rishit30G: Yes, as I'm using `space-between` in styling so it was needed

afercia (changes requested):
Thanks for the PR! Yes, this is how the Copy button should be placed. I left some comments, please have a look. I don't think the Input control and the input base components should be changed.

packages/components/src/external-link/types.ts
Rishit30G: Sure, can you suggest a better description
Thanks

afercia:
Note: the unit tests are failing because other components e.g. `NumberControl` use the `suffix` to add the `Increment` button. The Suffix prop and component should not be removed from the input control and input base components.

Rishit30G:
Hey @afercia, 
Thanks for sharing your feedback, I have reverted back the changes in the Input Control and Input Base component 
Additionally, I have left few comments as well in response to your suggestions
Appreciate your time and efforts

PR #63714, opened 2024-07-18
Components: add "Naming conventions" section

### LINKED ISSUE #63242: `@wordpress/components`: agree on a naming convention for compound (sub) components
As we plan on introducing more modularity (ie. compound components) to some of the components offered by the package, we need to make sure that we pick the right naming strategy. Bringing some consistency will also retroactively benefit existing components.

## Options

Here are some options that @DaniGuardiola and myself came up with while discussing this topic:

### Option 1: Function + properties

```jsx

// Implementation
function Component() {}
Component.Subcomponent = function() {}

// Usage
<Component>
  <Component.Subcomponent />
</Component>
```

### Option 2: Dot notation (either by exporting an object or by using single named exports)

```jsx
// Using objects:
//================

// Implementation
Component = {
  Root: function() {},
  Subcomponent: function() {}
}

// Usage
<Component.Root>
  <Component.Subcomponent />
</Component>

// Using named exports
//================

// Implementation
export function Root() {}
export function Subcomponent = function() {}

// Usage
import * as Component

<Component.Root>
  <Component.Subcomponent />
</Component>
```

### Option 3: "flat" namespacing (ie. without dot notation)

```jsx

// Implementation
export function Component() {}
export function ComponentSubcomponent = function() {}

// Usage
import * as Component

<Component>
  <ComponentSubcomponent />
</Component>
```

## Mixing compound components and monolithic components

Finally, we may also have to keep, for the same component, both a "monolithic" version and a more "a la carte", compound components-based version.

One hypothesis that made was to use the default export for the monolith, and to use option 2 (dot notation) for the compound components (ie.):

```jsx
// Implementation

export function Root() {}
export function Subcomponent() {}

export default function Comp() {
  return (
    <Root>
      <Subcomponent />
    </Root>
  );
}

// Usage

<Component />
// or
<Component.Root>
  <ComponentSubcomponent />
</Component.Root>

### PR DESCRIPTION
**Pull request by ciampo**
## What?

Closes https://github.com/WordPress/gutenberg/issues/63242

Tweak the contributing guidelines, updating a few out of date parts and adding a new "Naming Conventions" section based on the conversation happening in #63242

## Why?

As we introduce more compound components in the package, we need a recommended naming conventions for both monolithic and compound components going forward.

### DIFF
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -10,6 +10,7 @@ This set of guidelines should apply especially to newly introduced components. I
 -   [Compatibility](#compatibility)
 -   [Compound components](#compound-components)
 -   [Components & Hooks](#components--hooks)
+-   [Naming Conventions](#naming-conventions)
 -   [TypeScript](#typescript)
 -   [Styling](#styling)
 -   [Context system](#context-system)
@@ -230,6 +231,53 @@ A couple of good examples of how hooks are used for composition are:
 ## Performance
 
 TDB -->
+
+## Naming Conventions
+
+It is recommended for compound components to use dot notation to separate the namespace from the individual component names. The top-level compound component should be called `Root`.
+
+In comparison to compound components, it is recommended for monolithic components not to use dot-notation or namespacing in their export names.
+
+Dedicated React context should be also using the dot notation, while hooks should not
+
+```tsx
+import { Component, useComponent } from '@wordpress/components';
+import { useContext } from '@wordpress/element';
+
+function CompoundExample() {
+	return (
+		<Component.Root>
+			<Component.SubComponent />
+		</Component.Root>
+	);
+}
+
+function MonolithExample() {
+	return <Component />;
+}
+
+function ContextProviderExample() {
+	return (
+		<Component.Context.Provider value={ /* ... */ }>
+			{ /* React tree */ }
+		</Component.Context.Provider>
+	);
+}
+
+function ContextConsumerExample() {
+	const componentContext = useContext( Component.Context );
+
+	// etc
+}
+
+function HookExample() {
+	const hookReturnValue = useComponent();
+
+	// etc.
+}
+```
+
+The suggested naming conventions can help consumers of the package to distinguish between monolithic a compound components, and allows components to expose at the same time a monolithic and a compound version if needed.
 
 ## TypeScript
 
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -456,7 +456,7 @@ Please refer to the [JavaScript Testing Overview docs](/docs/contributors/code/t
 
 All new components should add stories to the project's [Storybook](https://storybook.js.org/). Each [story](https://storybook.js.org/docs/react/get-started/whats-a-story) captures the rendered state of a UI component in isolation. This greatly simplifies working on a given component, while also serving as an interactive form of documentation.
 
-A component's story should be showcasing its different states — for example, the different variants of a  `Button`:
+A component's story should be showcasing its different states — for example, the different variants of a `Button`:
 
 ```jsx
 import Button from '../';
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -6,7 +6,6 @@ The following is a set of guidelines for contributing to the `@wordpress/compone
 
 This set of guidelines should apply especially to newly introduced components. In fact, while these guidelines should also be retroactively applied to existing components, it is sometimes impossible to do so for legacy/compatibility reasons.
 
-For an example of a component that follows these requirements, take a look at [`ItemGroup`](/packages/components/src/item-group).
 - [Introducing new components](#introducing-new-components)
 - [Compatibility](#compatibility)
 - [Compound components](#compound-components)
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -127,18 +127,18 @@ When creating components that render a list of subcomponents, prefer to expose t
 
 ```jsx
 // ✅ Do:
-<List.Root>
+<List>
 	<List.Item value="Item 1" />
 	<List.Item value="Item 2" />
 	<List.Item value="Item 3" />
-</List.Root>
+</List>
 ```
 
 When implementing this pattern, avoid using `React.Children.map` and `React.cloneElement` to map through the children and augment them. Instead, use React Context to provide state to subcomponents and connect them:
 
 ```jsx
 // ❌ Don't:
-function ListRoot ( props ) {
+function List ( props ) {
 	const [ state, setState ] = useState();
 	return (
 		<div { ...props }>
@@ -152,7 +152,7 @@ function ListRoot ( props ) {
 // ✅ Do:
 const ListContext = createContext();
 
-function ListRoot( props ) {
+function List( props ) {
 	const [ state, setState ] = useState();
 	return (
 		<ListContext.Provider value={ state }>
@@ -234,26 +234,36 @@ TDB -->
 
 ## Naming Conventions
 
-It is recommended for compound components to use dot notation to separate the namespace from the individual component names. The top-level compound component should be called `Root`.
+It is recommended for compound components to use dot notation to separate the namespace from the individual component names. The top-level compound component should be called as the namespace (no dot notation).
 
-In comparison to compound components, it is recommended for monolithic components not to use dot-notation or namespacing in their export names.
-
-Dedicated React context should be also using the dot notation, while hooks should not
+Dedicated React context should also use the dot notation, while hooks should not.
 
 ```tsx
+// Component.tsx
+//=======================
+/** The top-level component's JSDoc. */
+export const Component = Object.assign( TopLevelComponent, {
+	/** The sub-component's JSDoc. */
+	SubComponent,
+	/** The sub-component's JSDoc. */
+	Content,
+} );
+
+export function useComponent() {
+	/* ... */
+}
+
+// App.tsx
+//=======================
 import { Component, useComponent } from '@wordpress/components';
 import { useContext } from '@wordpress/element';
 
-function CompoundExample() {
+function CompoundComponentExample() {
 	return (
-		<Component.Root>
+		<Component>
 			<Component.SubComponent />
-		</Component.Root>
+		</Component>
 	);
-}
-
-function MonolithExample() {
-	return <Component />;
 }
 
 function ContextProviderExample() {
@@ -276,8 +286,6 @@ function HookExample() {
 	// etc.
 }
 ```
-
-The suggested naming conventions can help consumers of the package to distinguish between monolithic a compound components, and allows components to expose at the same time a monolithic and a compound version if needed.
 
 ## TypeScript
 
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -94,7 +94,7 @@ In these situations, one possible approach is to "soft-deprecate" a given legacy
 2. Updating all places in Gutenberg that use that API.
 3. Adding deprecation warnings (only after the previous point is completed, otherwise the Browser Console will be polluted by all those warnings and some e2e tests may fail).
 
-When adding new components or new props to existing components, it's recommended to prefix them with `__unstable` or `__experimental` until they're stable enough to be exposed as part of the public API.
+When adding new components or new props to existing components, it's recommended to create a [private version](/packages/private-apis/README.md)) of the component until the changes are stable enough to be exposed as part of the public API.
 
 ### Learn more
 
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -241,14 +241,27 @@ Dedicated React context should also use the dot notation, while hooks should not
 ```tsx
 // Component.tsx
 //=======================
-/** The top-level component's JSDoc. */
-export const Component = Object.assign( TopLevelComponent, {
-	/** The sub-component's JSDoc. */
-	SubComponent,
-	/** The sub-component's JSDoc. */
-	Content,
-} );
+import { forwardRef, createContext } from '@wordpress/element';
 
+function TopLevelComponent = forwardRef( function TopLevel( props, ref ) { /* ... */ });
+
+function SubComponent = forwardRef( function Sub( props, ref ) { /* ... */ });
+SubComponent.displayName = 'Component.SubComponent';
+
+const Context = createContext();
+
+/** The top-level component's JSDoc. */
+export const Component = Object.assign(
+	TopLevelComponent,
+	{
+		/** The sub-component's JSDoc. */
+		SubComponent,
+		/** The sub-component's JSDoc. */
+		Context,
+	}
+);
+
+/** The hook's JSDoc. */
 export function useComponent() {
 	/* ... */
 }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ciampo** [comment]
@WordPress/gutenberg-components I took an initial stab at updating the contributing guidelines.

Also cc @diegohaz in case you have any feedback.

File: packages/components/CONTRIBUTING.md [hunk 1]
```diff
@@ -6,20 +6,20 @@ The following is a set of guidelines for contributing to the `@wordpress/compone
...
-- [TypeScript](#typescript)
-- [Styling](#styling)
-- [Context system](#context-system)
-- [Unit tests](#unit-tests)
-- [Storybook](#storybook)
-- [Documentation](#documentation)
-- [README example](#README-example)
-- [Folder structure](#folder-structure)
-- [Component versioning](#component-versioning)
+-   [Introducing new components](#introducing-new-components)
```
ciampo, line 9: My IDE insists on auto-formatting this file. I've put all the auto-formatting changes in one single commit, but if you think they're too distracting / they don't belong to this PR I'm happy to remove and potentially apply them in a separate PR.

packages/components/CONTRIBUTING.md
```diff
@@ -95,13 +95,13 @@ In these situations, one possible approach is to "soft-deprecate" a given legacy
...
+When adding new components or new props to existing components, it's recommended to create a [private version](/packages/private-apis/README.md)) of the component until the changes are stable enough to be exposed as part of the public API.
 
 ### Learn more
 
-- [How to preserve backward compatibility for a React Component](/docs/contributors/code/backward-compatibility.md#how-to-preserve-backward-compatibility-for-a-react-component)
-- [Experimental and Unstable APIs](/docs/contributors/code/coding-guidelines.md#experimental-and-unstable-apis)
-- [Deprecating styles](#deprecating-styles)
+-   [How to preserve backward compatibility for a React Component](/docs/contributors/code/backward-compatibility.md#how-to-preserve-backward-compatibility-for-a-react-component)
+-   [Experimental and Unstable APIs](/docs/contributors/code/coding-guidelines.md#legacy-experimental-apis-plugin-only-apis-and-private-apis)
+-   [Deprecating styles](#deprecating-styles)
```
ciampo, line 104: I've updated this section to reflect the new "private APIs" policy

packages/components/CONTRIBUTING.md
```diff
@@ -121,28 +121,24 @@ When creating components that render a list of subcomponents, prefer to expose t
...
-<List>
-	<ListItem value="Item 1" />
-	<ListItem value="Item 2" />
-	<ListItem value="Item 3" />
-</List>
+<List.Root>
+	<List.Item value="Item 1" />
+	<List.Item value="Item 2" />
+	<List.Item value="Item 3" />
+</List.Root>
```
ciampo: I've updated the naming of these components to align with the new recommended naming convention

packages/components/CONTRIBUTING.md
```diff
@@ -6,20 +6,20 @@ The following is a set of guidelines for contributing to the `@wordpress/compone
 
 This set of guidelines should apply especially to newly introduced components. In fact, while these guidelines should also be retroactively applied to existing components, it is sometimes impossible to do so for legacy/compatibility reasons.
 
-For an example of a component that follows these requirements, take a look at [`ItemGroup`](/packages/components/src/item-group).
```
tyxla, line 9: Maybe we can add a more recent example? Or you prefer to have none?

packages/components/CONTRIBUTING.md
```diff
@@ -121,28 +121,24 @@ When creating components that render a list of subcomponents, prefer to expose t
...
 ```
 
 ```jsx
 // ✅ Do:
-<List>
-	<ListItem value="Item 1" />
-	<ListItem value="Item 2" />
-	<ListItem value="Item 3" />
-</List>
+<List.Root>
```
tyxla: I guess we still need to get to an agreement about this across the community 🤷

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,53 @@ A couple of good examples of how hooks are used for composition are:
 
 TDB -->
 
+## Naming Conventions
+
+It is recommended for compound components to use dot notation to separate the namespace from the individual component names. The top-level compound component should be called `Root`.
+
+In comparison to compound components, it is recommended for monolithic components not to use dot-notation or namespacing in their export names.
+
+Dedicated React context should be also using the dot notation, while hooks should not
```
tyxla:
```suggestion
Dedicated React context should also use the dot notation, while hooks should not.
```

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,53 @@ A couple of good examples of how hooks are used for composition are:
...
+}
+
+function HookExample() {
+	const hookReturnValue = useComponent();
+
+	// etc.
+}
+```
+
+The suggested naming conventions can help consumers of the package to distinguish between monolithic a compound components, and allows components to expose at the same time a monolithic and a compound version if needed.
```
tyxla:
```suggestion
The suggested naming conventions can help consumers of the package to distinguish between monolithic and compound components, and allows components to expose at the same time a monolithic and a compound version if needed.
```

tyxla (commented):
This is more or less looking good, but still has some debatable elements (the `Component.Root` dot notation mostly) that still needs a stronger consensus in the #63242 discussion at this time.

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,53 @@ A couple of good examples of how hooks are used for composition are:
 
 TDB -->
 
+## Naming Conventions
```
mirka, line 235: Maybe there was some nuance lost in the discussion, but my understanding was that things are not going to be in a clear "pure monolith vs. fully modular" binary. More like:

- _Relatively_ monolithic: This could be a pure monolith if it's a very simple component, but usually they are going to be at least somewhat modular. The `CustomSelectControlV2` API is a good example of this. It's modular, but isn't as low-level modular as the raw Ariakit API (`Ariakit.Select`, `Ariakit.SelectPopover`).
- Fully modular: I basically think of this as a mirror of the raw Ariakit API.

So I was thinking, the default API ("relatively monolithic") that most consumers will use is going to be:

```jsx
		<Component>
			<Component.SubComponent />
		</Component>
```

And the low-level API ("fully modular"), if the need arises to provide it, is going to look like:

```jsx
		<Component.Root>
			<Component.SubComponent />
			<Component.LowLevelSubComponent lowLevelProp="foo" />
		</Component.Root>
```

tyxla: I like your phrasing, @mirka! Let's bring that up to https://github.com/WordPress/gutenberg/issues/63242 - providing both options sounds like a good compromise that might solve much of the arguments there.

ciampo: I personally thought the monolithic version would be represented by one component, without the need to use compound components.

We'd have:
- the compound component version (as granular or high-level as we'd like)
- the monolithic version (only if we see value in it, or if we have back-compat reasons)

I'm sure about how much value could a "relatively monolithic" approach give. Having an in-between state could cause confusion in the consumers, make naming conventions and APIs more complex.

What value do we see in this approach?

mirka: I don't really see it as an in-between state, because I don't think most consumers care about a monolithic/modular distinction as they're using a component. They're going to consult the docs anyway, instead of inferring its APIs by whether it's monolithic or modular. It's either just the "default API" that most consumers should use, or the "low-level API" that may be available if further customization is needed.

What matters is that the default API strikes the best balance between consistency, customizability, and convenience, while the low-level API(s) is focused on just customizability. In many cases, we may never need to provide a low-level API at all, but that option is there if we ever do.

And even if we do provide lower-level APIs, we won't necessarily always need the `Component.Root`. For example, if we designed the RadioControl today from scratch, we would definitely pass `RadioControl.Option`s to `children` instead of an object array to the `options` prop. However, if we wanted to also provide a lower-level API we could just expose e.g. a `RadioControl.Radio` and `RadioControl.Label`, without adding a `RadioControl.Root`.

A `Component.Root` only becomes necessary when `<Component>` as a root already includes some functionality that we want to break off as a separate subcomponent.

ciampo: My concern is that offering, at the same time, both a low-level and a high-level version of a compound component could be tricky. We'd have many small components and a much larger API surface to maintain, and we'd need to make sure it's clear which subcomponents can be used together, and which ones don't pair well.

**Have we considered the simpler alternative of dropping the `Root` naming altogether and not committing to offering a high-level version of a compound component?**

- Our compound components would be as low-level as we think is needed
- Less disruption in the naming scheme (no `Root`, less changes with respect to the current situation)
- We can offer more "high-level", ready-made examples in Storybook
- Consumers of the package can create their "high-level" version of the component in their codebases / packages, picking the abstraction they'd prefer
- If a component is already offering a monolithic version of a component, we simply create a new version of the component (`V[n]`) and make it compound.

This would reduce the API surface to maintain, simplify conventions, and it would overall give us a smaller maintenance burden.

mirka: I agree with your points, and I think we can converge on that.

In general we will set out to design an API that is reasonably low-level, and hopefully will never require adding a `.Root`, as best as we can anticipate. But, if the unanticipated ever happens, we will have `.Root` available as a last resort. Having a separate low-level API is not something we should plan on from the get go.

Does that sound about right? In that case, we don't need to document anything about the `.Root` thing, we can just keep it in our back pocket and pray that we'll never need it.

ciampo: I can write an update in the main issue with this alternative approach, which should hopefully be a better compromise for everyone involved.

ciampo: https://github.com/WordPress/gutenberg/issues/63242#issuecomment-2248422053

DaniGuardiola (commented):
Looks good so far, though it seems like we're switching to an "overloaded" API?

mirka:
@DaniGuardiola and I were discussing some potential TypeScript problems with overloading when `forwardRef()` is involved.

It's a bit finicky where you need to place your JSDocs and `.displayName`s, but I can confirm that [this approach with `Object.assign`](https://github.com/WordPress/gutenberg/commit/2428dfd65852bf96554a97149088871eef20cef6) is working as expected for IntelliSense and the Storybook docgen.

packages/components/CONTRIBUTING.md
ciampo: I think we have some many different implementations for components (naming conventions, monolithic vs compound, first party vs ariakit, sass vs CSS-in-JS, keeping as much logic in a hook vs in the render function...) that I'm hesitant to pick one as THE example. It also has potential to get out of date without us noticing.

ciampo:
I've updated the PR suggesting the "overloaded" naming convention and refreshing the code examples.

Could y'all take a new look?

(as I'm going to be AFK for a week, feel free to push updates and/or merge in case in my absence)

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,74 @@ A couple of good examples of how hooks are used for composition are:
...
+It is recommended for compound components to use dot notation to separate the namespace from the individual component names. The top-level compound component should be called as the namespace (no dot notation).
+
+Dedicated React context should also use the dot notation, while hooks should not.
+
+```tsx
+// Component.tsx
+//=======================
+import { forwardRef, createContext } from '@wordpress/element';
+
+function TopLevelComponent = forwardRef( function TopLevel( props, ref ) { /* ... */ });
```
tyxla: Should we have the unforwarded be declared separately as `UnforwardedTopLevelComponent` as we do for our components usually?

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,74 @@ A couple of good examples of how hooks are used for composition are:
...
+
+function SubComponent = forwardRef( function Sub( props, ref ) { /* ... */ });
+SubComponent.displayName = 'Component.SubComponent';
+
+const Context = createContext();
+
+/** The top-level component's JSDoc. */
+export const Component = Object.assign(
```
tyxla: Could be a good opportunity to explain the purpose behind using `Object.assign()` here.

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,74 @@ A couple of good examples of how hooks are used for composition are:
...
+	TopLevelComponent,
+	{
+		/** The sub-component's JSDoc. */
+		SubComponent,
+		/** The sub-component's JSDoc. */
```
tyxla:
```suggestion
		/** The context's JSDoc. */
```

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,74 @@ A couple of good examples of how hooks are used for composition are:
 
 TDB -->
 
+## Naming Conventions
+
+It is recommended for compound components to use dot notation to separate the namespace from the individual component names. The top-level compound component should be called as the namespace (no dot notation).
```
tyxla: Grammar nits:

```suggestion
It is recommended that compound components use dot notation to separate the namespace from the individual component names. The top-level compound component should be called the namespace (no dot notation).
```

packages/components/CONTRIBUTING.md
tyxla: 👍 

It's also easier to review the PR with white-spacing disabled:

Screenshot 2024-08-01 at 14 05 08

tyxla (commented):
This is looking great 👍 

I've left some minor things mostly, I'll let @mirka and @DaniGuardiola take a look and share their feedback before we ship it.

packages/components/CONTRIBUTING.md
diegohaz:
```suggestion
const TopLevelComponent = forwardRef( function TopLevel( props, ref ) { /* ... */ });
```

packages/components/CONTRIBUTING.md
mirka: Yes, and it might be useful to state what exactly we need to look for:

- When consuming the component/subcomponent, the JSDocs must appear correctly in IntelliSense.
- The main component JSDoc must appear in the Storybook docs page.
- The component/subcomponent prop types must appear correctly in the Storybook props table.

mirka (commented):
Thank you for the update! I had a few notes on top of @tyxla's review, but otherwise this looks great 👍

packages/components/CONTRIBUTING.md
ciampo: Well spotted, thank you!

packages/components/CONTRIBUTING.md
ciampo: Added a couple of paragraphs explaining requirements and related recommended best practices: https://github.com/WordPress/gutenberg/pull/63714/commits/432635b3f1242b0d6a3b95e54e9d8d36af0b5519

This way the only comments in the code are related to the JSDocs

ciampo:
All feedback should be addressed

packages/components/CONTRIBUTING.md
```diff
@@ -236,6 +232,95 @@ A couple of good examples of how hooks are used for composition are:
...
+// Component.tsx
+//=======================
+import { forwardRef, createContext } from '@wordpress/element';
+
+const UnforwardedTopLevelComponent = function TopLevel( props, ref ) {
+	/* ... */
+};
+const TopLevelComponent = forwardRef( UnforwardedTopLevelComponent );
+
+const UnforwardedSubComponent = function Sub( props, ref ) {
```
tyxla: I don't see a good reason to go with `const X = function Y()` syntax, when we can just do `function X`. 

Seems like we can update that before merging.

```suggestion
function UnforwardedTopLevelComponent( props, ref ) {
	/* ... */
};
const TopLevelComponent = forwardRef( UnforwardedTopLevelComponent );

function UnforwardedSubComponent( props, ref ) {
```

tyxla (approved):
Good to go from my perspective. Thanks @ciampo 🚀

PR #63715, opened 2024-07-18
Patterns: Render draggable only when enabled

tyxla:
## What?
Render the `Draggable` components inside the inserter patterns only when draggable is actually enabled. 

## Why?
Currently, we'll render a `<Draggable />` with its draggable chip unnecessarily when draggable is disabled. This feels unnecessary since, in some instances, the patterns will intentionally never be draggable.

## How?
We're skipping the render of `Draggable` and the draggable chip. This results in a smaller component tree and fewer items in the DOM tree when patterns are not draggable.

## Testing Instructions
* Open the inserter.
* Verify blocks are still draggable and droppable.
* Click "Patterns"
* Verify patterns are still draggable and droppable.
* Click the "Explore all patterns" button
* Verify you can still insert those patterns and they render the same way as before. 
* You can also observe a smaller DOM / component tree.

None

## Screenshots or screencast 
|Before|After|
|---|---|
|Screenshot 2024-07-18 at 18 49 13|Screenshot 2024-07-18 at 18 45 20

### DIFF
--- a/packages/block-editor/src/components/inserter-draggable-blocks/index.js
+++ b/packages/block-editor/src/components/inserter-draggable-blocks/index.js
@@ -43,6 +43,14 @@ const InserterDraggableBlocks = ( {
 		useDispatch( blockEditorStore )
 	);
 
+	if ( ! isEnabled ) {
+		return children( {
+			draggable: false,
+			onDragStart: undefined,
+			onDragEnd: undefined,
+		} );
+	}
+
 	return (
 		<Draggable
 			__experimentalTransferDataType="wp-blocks"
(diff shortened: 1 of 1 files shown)


Mamaduka (approved):
It makes sense. The `BlockDraggable` component does a similar early return.

https://github.com/WordPress/gutenberg/blob/0a7e10db2037cb80e4ab8cda4363ba18d6a3a83f/packages/block-editor/src/components/block-draggable/index.js#L175-L177

tyxla:
Thanks, @Mamaduka!

Issue #63467, opened 2024-07-12
WP 6.6 RC 3: Block global styles in the site editor become stuck on the site frontend


andrewserong:
### Description

Tracked in core in https://core.trac.wordpress.org/ticket/59595#comment:71 — note this is an issue in core, and not the Gutenberg repo. It's added here for visibility as it relates to behaviour with global styles in the site editor.

It looks like the caching strategy in WP 6.6 has resulted in a regression where the styles for blocks in global styles become stuck on the site frontend when a user goes to make updates to the styles for a block within global styles.

### Step-by-step reproduction instructions

1. Open the site editor on a WP 6.6 RC3 site running Twenty Twenty Four theme (without Gutenberg activated)
2. Click the editor preview to go into edit mode
3. Open the Global Styles sidebar panel on the right hand side
4. Click Blocks
5. Select the Button block
6. Give the Button block a style, e.g. Background color set to a gradient
7. Save the global styles
8. Load the site frontend — notice that the styling on the Button is applied
9. Go back to global styles and give the Button block another style, e.g. large font size
10. Save the global styles
11. Load the site frontend and notice that the Button block's styling has not changed — it'll be cached with the previous styles that were originally saved

### Screenshots, screen recording, code snippet

https://github.com/user-attachments/assets/4c5b9543-225a-4bb9-ad8e-d8afb648fab4

### Environment info

- WP 6.6 RC 3
- Gutenberg is _not_ active. This issue does not appear to occur in Gutenberg.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

andrewserong:
There is a potential draft fix open over in: https://github.com/WordPress/wordpress-develop/pull/7017

andrewserong:
Closing this out now as the change that introduced the regression has been reverted in core. See: https://core.trac.wordpress.org/ticket/59595#comment:84

PR #63721, opened 2024-07-18
Add unit test for get_user_data_from_wp_global_styles.

joemcgill:
## What?

This adds a unit test to demonstrate that `WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles` returns an empty array when no `$create_post` param is passed after being previously called with the `$create_post` param set to `true`.

See: https://github.com/WordPress/gutenberg/issues/63722

## Why?

While attempting to write a unit test to confirm that caching added to `wp_add_global_styles_for_blocks()` would get cleared if user data changed ([related issue](https://github.com/WordPress/gutenberg/issues/63467)), I noticed that after calling `WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles()` and passing a value to the second parameter to create a post, the next time that method is called, the created post is not returned.

## How?

This PR demonstrates a possible bug. If valid, the PR would need to be updated to fix the issue.

## Testing Instructions

Review the Unit test failure on the PR or run the test locally:

`npm run test:unit:php:base -- --filter test_get_user_data_from_wp_global_styles_returns_created_post`

n/a

### DIFF
--- a/phpunit/class-wp-theme-json-resolver-test.php
+++ b/phpunit/class-wp-theme-json-resolver-test.php
@@ -766,6 +766,7 @@ class WP_Theme_JSON_Resolver_Gutenberg_Test extends WP_UnitTestCase {
 	public function test_get_user_data_from_wp_global_styles_returns_created_post() {
 		// Switch to a theme that does have support.
 		switch_theme( 'block-theme' );
+		wp_set_current_user( self::$administrator_id );
 		$theme = wp_get_theme();
 
 		$created  = WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles( $theme, true );
--- a/phpunit/class-wp-theme-json-resolver-test.php
+++ b/phpunit/class-wp-theme-json-resolver-test.php
@@ -761,6 +761,20 @@ class WP_Theme_JSON_Resolver_Gutenberg_Test extends WP_UnitTestCase {
 	}
 
 	/**
+	 * @covers WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles
+	 */
+	public function test_get_user_data_from_wp_global_styles_returns_created_post() {
+		// Switch to a theme that does have support.
+		switch_theme( 'block-theme' );
+		$theme = wp_get_theme();
+
+		$created  = WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles( $theme, true );
+		$returned = WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles( $theme );
+
+		$this->assertSame( $created, $returned, 'User data does not return created post' );
+	}
+
+	/**
 	 * @covers WP_Theme_JSON_Resolver_Gutenberg::get_theme_data
 	 */
 	public function test_get_theme_data_theme_supports_overrides_theme_json() {


joemcgill:
Failing test that demonstrates the issue can be seen here: https://github.com/WordPress/gutenberg/actions/runs/9997910616/job/27636197439?pr=63721

andrewserong:
I'm not very familiar with the inner workings of `get_user_data_from_wp_global_styles`, and `@tellthemachines` isn't around at the moment, so just adding @oandregal as a reviewer for visibility. Looks like this PR adds a test to flag the issue 👍, but I'm not sure what the appropriate fix would be.

joemcgill:
Thanks @andrewserong. You're exactly right that this is just demonstrating the issue. Before attempting to suggest a fix, I wanted to confirm whether this is actually a bug, or if I'm misunderstanding something about how this method is meant to work. Generally speaking, when the initial post type gets created for custom styles, it happens through a REST endpoint, and I'm uncertain if this method is actually even used in that process (I suspect not).

What I'm mostly unsure about is whether the code in this method that creates the post is missing something, or whether the query that returns the post is incorrect.

phpunit/class-wp-theme-json-resolver-test.php
```diff
@@ -760,6 +760,20 @@ public function test_get_user_data_from_wp_global_styles_filter_state() {
 		$this->assertSameSets( array(), $post2 );
 	}
 
+	/**
+	 * @covers WP_Theme_JSON_Resolver_Gutenberg::get_user_data_from_wp_global_styles
+	 */
+	public function test_get_user_data_from_wp_global_styles_returns_created_post() {
+		// Switch to a theme that does have support.
+		switch_theme( 'block-theme' );
```
aaronrobertshaw, line 768:
```suggestion
		switch_theme( 'block-theme' );
		wp_set_current_user( self::$administrator_id );
```

With the correct capabilities to assign terms, this test passes.

aaronrobertshaw (commented):
I had a bit of a look into this. I believe `get_user_data_from_wp_global_styles` is intended to return the global styles custom post type as this test suggests.

The catch is if the current user does not have the capabilities to assign terms, they are not added when the post is inserted. The subsequent query then fails to match on the taxonomy.

In the proposed test, if the user is set to an administrator, the test passes (see inline comment below).

I would suspect that in the case of global styles, the taxonomy term needs to be set regardless of user capabilities so the post relates to the appropriate theme. Whether the user has caps to create the global styles post itself might be a different question.

If the above suspicion is correct, then a small tweak to manually call `wp_set_object_terms( $cpt_post_id, array( $stylesheet ), 'wp_theme' );` instead of relying on `tax_input` in the args for `wp_insert_post` makes the test pass as well.

```diff
diff --git a/lib/class-wp-theme-json-resolver-gutenberg.php b/lib/class-wp-theme-json-resolver-gutenberg.php
index a2153e639d..98c4912081 100644
--- a/lib/class-wp-theme-json-resolver-gutenberg.php
+++ b/lib/class-wp-theme-json-resolver-gutenberg.php
@@ -504,12 +504,10 @@ class WP_Theme_JSON_Resolver_Gutenberg {
 					'post_title'   => 'Custom Styles', // Do not make string translatable, see https://core.trac.wordpress.org/ticket/54518.
 					'post_type'    => $post_type_filter,
 					'post_name'    => sprintf( 'wp-global-styles-%s', urlencode( $stylesheet ) ),
-					'tax_input'    => array(
-						'wp_theme' => array( $stylesheet ),
-					),
 				),
 				true
 			);
+			wp_set_object_terms( $cpt_post_id, array( $stylesheet ), 'wp_theme' );
 			if ( ! is_wp_error( $cpt_post_id ) ) {
 				$user_cpt = get_object_vars( get_post( $cpt_post_id ) );
 			}

```

PR #63723, opened 2024-07-18
HTML API: Backport updates from Core

dmsnell:
This PR updates the HTML API files in the 6.6 and 6.7 compat directories. It brings the contents of the 6.7 directory up to commit SHA 7b75c8e1d288620e52d0b86c6de5017882578561.

Note to reviewers: The type annotations added in 6.7 have changed every file in the HTML API, which is why every reference references the `_6_7` files and why each file has the corresponding `_6_7` file.

### DIFF
--- a/lib/compat/wordpress-6.7/class-gutenberg-token-map-6-7.php
+++ b/lib/compat/wordpress-6.7/class-gutenberg-token-map-6-7.php
@@ -1,0 +1,820 @@
+<?php
+
+/**
+ * Class for efficiently looking up and mapping string keys to string values, with limits.
+ *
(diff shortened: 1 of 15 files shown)


sirreal (approved):
This looks correct and all of the usages of the HTML API seem to be updated correctly.

dmsnell:
@sirreal the end-to-end tests are reliably failing and I see mentions of block bindings and of the button block, both of which rely on the HTML API, so I think something could still be wrong here.

dmsnell:
@sirreal found a typo on `Gutenberg_HTML_Attibute_6_7` and fixed it.

PR #63726, opened 2024-07-19
Ensure root selector (body) is not wrapped in :root :where()

### LINKED ISSUE #63712: global-styles-inline-css rule introduces breaking changes " :root :where(body)"
### Description

A CSS rule applied to body, introduces some breaking CSS (via **global-styles-inline-css** ).
Ther maybe other CSS that could be dangerous, but the body selector in particular is stronger than user's may have implemented in their themes ( eg 'body {}' ) and therefore breaks (was in my case!) 

here is the CSS

`:root :where(body) {
    background-color: var(--wp--preset--color--background);
    color: var(--wp--preset--color--foreground);
    font-family: var(--wp--preset--font-family--system);
    font-size: var(--wp--preset--font-size--medium);
    line-height: 1.75;
    padding-top: 0px;
    padding-right: 0px;
    padding-bottom: 0px;
    padding-left: 0px;
}`

Note line height does not even use a CSS variable.

for example in my case (using Underscore and a child theme) 

-  '--wp--preset--font-family--system' was not define... since no fallback was given a system font was implemented.
- '--wp--preset--font-size--medium' was defined but different than the theme CSS defined body font-size.
- line-height also made a visible difference (no attempt was made to use a css-var)

1. Go to wordpress site that has Guttenberg enabled and is up-to-date
2. CSS is injected into the head `<style id="global-styles-inline-css">...`

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

talldan:
## What?
Fixes #63712

Trac backport ticket - https://core.trac.wordpress.org/ticket/61704

In WordPress 6.6, there was a general effort to lower and level specificity, but a couple of selectors seem to have mistakenly had their specificity raised.

The styles associated with the `body` are an example. In 6.5 these styles used the simple `body` selector. In 6.6 the selector is `:root :where(body)`, an increase in specificity.

## Why?
These changes may cause some problems for themes, especially the typography styles in the body, which will cascade.

## How?
Reverts the selector back to `body`

*props to @aaronrobertshaw for pointing me to the fix*

### Editor styles
1. Open the site editor
2. Click the editor area to open the editor
3. Using browser dev tools, inspect the `body` element inside the `edit-site-visual-editor__editor-canvas` iframe.
4. Check that the styles are like 'expected' snippet  below, with a simple `body` selector and not a `:root :where(body)` selector.

### Frontend styles
1. Click 'View site' to view the frontend
2. Insect the body element's styles
3. Check that the styles are like 'expected' snippet below, with a simple `body` selector and not a `:root :where(body)` selector.

Expected:
```css
body {
    --wp--style--root--padding-top: 115px;
    --wp--style--root--padding-right: var(--wp--preset--spacing--50);
    --wp--style--root--padding-bottom: 115px;
    --wp--style--root--padding-left: var(--wp--preset--spacing--50);
    color: var(--wp--preset--color--contrast);
    background-color: var(--wp--preset--color--base);
    font-family: var(--wp--preset--font-family--body);
    font-size: var(--wp--preset--font-size--medium);
    font-style: normal;
    font-weight: 400;
    line-height: 1.55;
}
```

Before:
```css
:root :where(body) {
    --wp--style--root--padding-top: 115px;
    --wp--style--root--padding-right: var(--wp--preset--spacing--50);
    --wp--style--root--padding-bottom: 115px;
    --wp--style--root--padding-left: var(--wp--preset--spacing--50);
    color: var(--wp--preset--color--contrast);
    background-color: var(--wp--preset--color--base);
    font-family: var(--wp--preset--font-family--body);
    font-size: var(--wp--preset--font-size--medium);
    font-style: normal;
    font-weight: 400;
    line-height: 1.55;
}
```


aaronrobertshaw (commented):
Thanks for the rapid response on this fix @talldan 🚀 

Quick question; it looks like this will address the PHP and frontend side of the issue, do we also need to update the JS side of global styles? 

I have a suspicion that we'll also need to update the [use-global-styles-output.js](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/global-styles/use-global-styles-output.js#L648-L653) hook. 

Similar to the top-level element-only selector fix, we could probably just add the `skipSelectorWrapper` flag to the root block selector linked above.

aaronrobertshaw (approved):
Appreciate the quick iteration here as well 🙇 

This is now testing well for me. I can confirm the root body selector is back to `body` after this PR and matches what it was with WP6.5.5. 

I also gave this a smoke test against a range of block style variations and didn't spot any regressions there.

#### WP 6.5.5
Screenshot 2024-07-19 at 2 22 01 PM

#### Editor Before / After
Screenshot 2024-07-19 at 2 16 50 PM

#### Frontend Before / After
Screenshot 2024-07-19 at 2 15 41 PM

If there is a chance to get this into WP6.6.1 we'll need the test instructions updated for the latest changes and an extra approval or two.

talldan:
Thanks @aaronrobertshaw, I've updated the test instructions.

andrewserong (approved):
This code change LGTM, too, and is testing well for me:

✅ Post and site editor styles target the simple `body` selector
✅ In the non-iframed editor, styles target the editor wrapper as expected
✅ Styles on the site frontend target the simple `body` selector

Looks good to land to me!

aaronrobertshaw:
I've tested the changes in this PR against the `wp/6.6` branch and it is working well there too.

| Editor | Frontend |
|---|---|
| Screenshot 2024-07-19 at 5 02 51 PM | Screenshot 2024-07-19 at 5 03 13 PM |

PR #63728, opened 2024-07-19
Downgrade node 22(.5) unit tests to 22.4

talldan:
Currently all node 22 unit tests are failing in trunk. It seems to be an issue for npm running on node 22.5.

Comments on https://github.com/npm/cli/issues/7657 imply this is an npm bug, and running on node 22.4 can be a temporary measure to keep CI working.

### DIFF
--- a/.github/workflows/create-block.yml
+++ b/.github/workflows/create-block.yml
@@ -14,13 +14,17 @@ concurrency:
 
 jobs:
     checks:
-        name: Checks w/Node.js ${{ matrix.node }} on ${{ matrix.os }}
+        name: Checks w/Node.js ${{ matrix.node.name }} on ${{ matrix.os }}
         runs-on: ${{ matrix.os }}
         if: ${{ github.repository == 'WordPress/gutenberg' || github.event_name == 'pull_request' }}
         strategy:
             fail-fast: false
             matrix:
-                node: ['20', '22.4']
+                node:
+                    - name: 20
+                      version: 20
+                    - name: 22
+                      version: 22.4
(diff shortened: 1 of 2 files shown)


.github/workflows/create-block.yml
```diff
@@ -14,13 +14,17 @@ concurrency:
 
 jobs:
     checks:
-        name: Checks w/Node.js ${{ matrix.node }} on ${{ matrix.os }}
+        name: Checks w/Node.js ${{ matrix.node.name }} on ${{ matrix.os }}
```
talldan, line 17: This way the job names don't change, it'll still be something like "Create Block / Checks w/Node.js 22 on macos-latest (pull_request)" and it means I won't have to change required checks in github's setting back and forth.

In future we might consider keeping this and using names like "LTS" and "Next" instead of specific version numbers.

.github/workflows/create-block.yml
```diff
@@ -14,13 +14,17 @@ concurrency:
...
         if: ${{ github.repository == 'WordPress/gutenberg' || github.event_name == 'pull_request' }}
         strategy:
             fail-fast: false
             matrix:
-                node: ['20', '22.4']
+                node:
+                    - name: 20
+                      version: 20
+                    - name: 22
+                      version: 22.4
```
talldan, line 27: Array/Object notation in yaml, as documented here:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

talldan:
This seems to be working so merging it into trunk

PR #63730, opened 2024-07-19
Move useSelectNearestEditableBlock out of src/hooks

noisysocks:
It's a little thing but I noticed `use-select-nearest-editable-block.js` is in `packages/editor/src/hooks` which is a folder full of WordPress hooks (`addFilter`), not React hooks. It's an internal function only used in `visual-editor` so I've moved it there.


andrewserong (approved):
Good catch! Co-locating it with the code where it's used seems like a reasonable place for it for now.

Just gave this a quick smoke test to double check that the logic for selecting the nearest editable block is still working while editing pages in the site editor.

LGTM! 🚀

PR #63732, opened 2024-07-19
TemplateContentPanel: Don't show content blocks that are in a Query Loop

### LINKED ISSUE #63613: Content panel shouldn't surface blocks inside the 'Content' block
When editing a page, the Content panel surfaces the sections of the document that can be directly manipulated. My understanding is that it does this by looping through the document looking for instances of specific blocks; Title, Featured Image, and Content. When found they are listed in the panel for easy selection:

Screenshot 2024-07-16 at 13 54 32

However, if the Content block _also_ contains instances of these blocks – for instance if there's a custom query in the page content – then those blocks are repeated:

For a number of reasons this makes for a strange UX:

* There's no benefit to highlighting nested featured images, titles, etc because as properties of separate entities they cannot be edited
* It's not clear which is the document-level version of the block
* They add weight to the Content panel, diluting its usefulness
* Selecting a nested block creates a confusing UI where both the block and the parent appear to be selected

Screenshot 2024-07-16 at 13 57 27

### Solution
It would seem sensible to omit these blocks in the Content panel when they are nested within the Content block.

cc @noisysocks as (iirc) you worked on this feature.

noisysocks:
Fixes https://github.com/WordPress/gutenberg/issues/63613.

When editing a page in the site editor, we populate the _Content_ panel in the sidebar with links to the content blocks that are editable.

Currently we blindly look for any content blocks: Post Title, Featured Image, Post Content.

This has the downside of catching blocks that appear nested within Post Content or within a Query Loop.

For example, you should only see Content once here:

Screenshot 2024-07-24 at 17 29 56

There was already logic for fixing exactly this in `DisableNonPageContentBlocks`, so I've moved that logic into a selector that we can use in multiple places.

Now we correctly only show the top-level content blocks:

Screenshot 2024-07-24 at 17 30 29

### Testing

1. Edit a page in the site editor.
2. Insert a Query Loop containing Title and/or Post Content blocks into the page's Post Content block.
3. Look at the Page sidebar. There shouldn't be any links to nested Title or Post Content blocks. There shouldn't be any links to any Template Part blocks.
4. Test that only the content blocks and Template Part blocks are editable in the editor.
5. Edit a template in the site editor. 
6. Look at the Template sidebar. There should only be links to the Template Part blocks.

### DIFF
--- a/packages/editor/src/components/provider/disable-non-page-content-blocks.js
+++ b/packages/editor/src/components/provider/disable-non-page-content-blocks.js
@@ -6,11 +6,15 @@ import { store as blockEditorStore } from '@wordpress/block-editor';
 import { useEffect } from '@wordpress/element';
 import { applyFilters } from '@wordpress/hooks';
 
-const DEFAULT_CONTENT_ONLY_BLOCKS = [
+/**
+ * Internal dependencies
+ */
+import { store as editorStore } from '../../store';
+
+const POST_CONTENT_BLOCK_TYPES = [
 	'core/post-title',
 	'core/post-featured-image',
 	'core/post-content',
-	'core/template-part',
 ];
 
 /**
@@ -18,28 +22,23 @@ const DEFAULT_CONTENT_ONLY_BLOCKS = [
  * page content to be edited.
  */
 export default function DisableNonPageContentBlocks() {
-	const contentOnlyBlocks = applyFilters(
+	const postContentBlockTypes = applyFilters(
 		'editor.postContentBlockTypes',
-		DEFAULT_CONTENT_ONLY_BLOCKS
+		POST_CONTENT_BLOCK_TYPES
 	);
 
 	// Note that there are two separate subscription because the result for each
 	// returns a new array.
-	const contentOnlyIds = useSelect( ( select ) => {
-		const { getBlocksByName, getBlockParents, getBlockName } =
-			select( blockEditorStore );
-		return getBlocksByName( contentOnlyBlocks ).filter( ( clientId ) =>
-			getBlockParents( clientId ).every( ( parentClientId ) => {
-				const parentBlockName = getBlockName( parentClientId );
-				return (
-					// Ignore descendents of the query block.
-					parentBlockName !== 'core/query' &&
-					// Enable only the top-most block.
-					! contentOnlyBlocks.includes( parentBlockName )
-				);
-			} )
-		);
-	}, [] );
+	const contentOnlyIds = useSelect(
+		( select ) => {
+			const { getPostBlocksByName } = select( editorStore );
+			return getPostBlocksByName( [
+				...postContentBlockTypes,
+				'core/template-part',
+			] );
+		},
+		[ postContentBlockTypes ]
+	);
 	const disabledIds = useSelect( ( select ) => {
 		const { getBlocksByName, getBlockOrder } = select( blockEditorStore );
 		return getBlocksByName( [ 'core/template-part' ] ).flatMap(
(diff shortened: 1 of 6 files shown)


packages/editor/src/store/selectors.js
```diff
@@ -1796,3 +1796,34 @@ export const getPostTypeLabel = createRegistrySelector(
...
+ *
+ * @return {Array} Block client IDs.
+ */
+export const getPostBlocksByName = createRegistrySelector(
+	( select ) => ( state, blockNames ) => {
+		blockNames = Array.isArray( blockNames ) ? blockNames : [ blockNames ];
+		const {
+			getBlocksByName,
+			getBlockParents,
+			getBlockName: _getBlockName,
```
Mamaduka: I think there's no need to rename the selector.

packages/editor/src/store/selectors.js
```diff
@@ -1796,3 +1796,34 @@ export const getPostTypeLabel = createRegistrySelector(
...
+ * Similar to getBlocksByName in @wordpress/block-editor, but only returns the top-most
+ * blocks that aren't descendants of the query block.
+ *
+ * @param {Object}       state      Global application state.
+ * @param {Array|string} blockNames Block names of the blocks to retrieve.
+ *
+ * @return {Array} Block client IDs.
+ */
+export const getPostBlocksByName = createRegistrySelector(
+	( select ) => ( state, blockNames ) => {
```
Mamaduka: I agree that making this a private selector makes sense for now.

We should also memoize the selector, but `blockNames` being an array might make that more challenging. Example: https://github.com/WordPress/gutenberg/pull/63732/files#diff-cf33f4589314d4a81e5a821d1433b83521dca9ad19d58a4b5ba1687e314b9558R35-R37

noisysocks: Hmm yeah this is tricky to get right. Please double check f2fea66bebe for me 😅

Mamaduka: I wonder if making a selector variadic function would work here.

Something like:
```
export const getPostBlocksByName = ( state, ...blockNames );
getPostBlocksByName( ...contentOnlyBlockTypes );
```

noisysocks: I don't want to stray far from `getBlocksByName` though.

Mamaduka: It's not a block for this PR. It's just an idea; I'm happy to give it a try separately.

Memoizing similar selector is always a plain 😅

packages/editor/src/components/provider/disable-non-page-content-blocks.js
```diff
@@ -3,43 +3,46 @@
...
-	const contentOnlyBlocks = applyFilters(
-		'editor.postContentBlockTypes',
-		DEFAULT_CONTENT_ONLY_BLOCKS
+	const contentOnlyBlockTypes = useMemo(
+		() => [
+			...applyFilters(
+				'editor.postContentBlockTypes',
+				POST_CONTENT_BLOCK_TYPES
+			),
+			'core/template-part',
```
Mamaduka, line 32: The `core/template-part` was removed from the const because we don't want extenders to remove it, correct?

Mamaduka (approved):
Thanks, @noisysocks!

The changes test well for me and fix the issue ✅

packages/editor/src/components/provider/disable-non-page-content-blocks.js
noisysocks: I was thinking along the lines of:

- `editor.postContentBlockType` allows extenders to control which block types we consider to be "post content".
- We're enabling selecting template parts via the same mechanism but it's not "post content".

PR #63733, opened 2024-07-19
DataViews: Fix pattens list selection

### LINKED ISSUE #63725: DataViews: Patterns are not bulk selectable, bulk select checkbox has "Select item: [object Object]" as its aria-label
### Description

The inability to bulk select seems to be an issue in the gutenberg plugin.

<s>The aria-label issue is something I can also see in WP 6.6</s>

1. Navigate to the patterns screen
2. Go to My Patterns
3. Observe that the bulk select checkbox is enabled, but clicking it does nothing
4. Inspect the checkbox using the browser dev tools
5. Observe that the aria-label is "Select item: [object Object]"

_No response_

### Environment info

Mac OS / Brave

- [X] Yes

- [X] Yes

ntsekouras:
## What?
Fixes: https://github.com/WordPress/gutenberg/issues/63725

This handles the selection issue for GB. aria-labels seems fine in trunk..

From the issue:

>The inability to bulk select seems to be an issue in the gutenberg plugin.

## Testing Instructions
1. In patterns list in site editor make sure the selection works as expected

### DIFF
--- a/packages/edit-site/src/components/page-patterns/index.js
+++ b/packages/edit-site/src/components/page-patterns/index.js
@@ -13,7 +13,13 @@ import {
 	Flex,
 } from '@wordpress/components';
 import { __, _x } from '@wordpress/i18n';
-import { useState, useMemo, useId, useEffect } from '@wordpress/element';
+import {
+	useState,
+	useMemo,
+	useId,
+	useEffect,
+	useCallback,
+} from '@wordpress/element';
 import {
 	BlockPreview,
 	privateApis as blockEditorPrivateApis,
@@ -252,6 +258,10 @@ export default function DataviewsPatterns() {
 	const {
 		params: { postType, categoryId: categoryIdFromURL },
 	} = useLocation();
+	const [ selection, setSelection ] = useState( [] );
+	const onChangeSelection = useCallback( ( items ) => {
+		setSelection( items );
+	}, [] );
 	const type = postType || PATTERN_TYPES.user;
 	const categoryId = categoryIdFromURL || PATTERN_DEFAULT_CATEGORY;
 	const [ view, setView ] = useState( DEFAULT_VIEW );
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -252,6 +258,10 @@ export default function DataviewsPatterns() {
 	const {
 		params: { postType, categoryId: categoryIdFromURL },
 	} = useLocation();
+	const [ selection, setSelection ] = useState( [] );
+	const onChangeSelection = useCallback( ( items ) => {
+		setSelection( items );
+	}, [] );
```
talldan: It looks like and identity function, so perhaps the `useCallback` isn't needed and `setSelection` can be passed straight in.

talldan (approved):
This fixes it for me, thank you. I was finally able to delete all my patterns ❤️

packages/edit-site/src/components/page-patterns/index.js
ntsekouras: Oh, yes. I had some other code at first with deps and forgot to clean up. Thanks!

packages/edit-site/src/components/page-patterns/index.js
```diff
@@ -409,6 +410,8 @@ export default function DataviewsPatterns() {
 					isLoading={ isResolving }
 					view={ view }
 					onChangeView={ setView }
+					selection={ selection }
+					onChangeSelection={ setSelection }
```
youknowriad, line 414: I don't understand why this fix is needed. I mean we're not doing anything with "selection" here. So for me, it means that "uncontrolled" selection in DataViews is not working properly and we should fix it within the component rather than switching to "controlled" selection.

ntsekouras: Hm.. sorry for missing this. I haven't worked much on the selection and didn't remember about uncontrolled version.. I'll take a look for the root issue then.

PR #26164, opened 2020-10-15
Use batch processing in edit-widgets package

adamziel:
This PR updates the `edit-widgets` package so that it saves the entities using batch processing against the correct API endpoints.

Testing instructions TBD

PR #24290, opened 2020-07-30
First stab at sidebars_widgets-based widget management

adamziel:
## Description
This is an attempt at making the experimental widget management screen fully functional so that it's possible to mix&match legacy widgets and gutenberg blocks.

It works as follows:
* wp_options are embraced, posts-based manipulations are removed (see https://github.com/WordPress/gutenberg/issues/24267).
* Blocks are just regular widgets implemented using the `WP_Widget_Block` class.
* Frontend is based on the data layer where we use widgetArea as the entity. It has a corresponding API endpoint which returns a widget area with all its widgets and it also makes it possible to POST a widget area with a set of desired widgets.
* Legacy widgets on the new screen are no longer saved on change, in fact they're not saved at all until a "Save" button is clicked.

Other considerations:
* Consider making widget a base entity instead of widget area - batch saving is a pre-requisite for this approach though, also it would require some discussion about how reordering widgets would work.
* The endpoint says /sidebars but the entity name is widgetArea - it would be good to address this discrepancy somehow.
* Customizer will only display rendered blocks instead of an actual Gutenberg instance - let's discuss.
* `/widget-utils` controller is still there and would be a good candidate for removal in some future iteration.
* Overall this PR adds a bunch of code behind a feature flag, merging it and then iterating on bugfixes would make a sensible approach.

Reuses parts of the following plugin (CC @martin-pettersson) https://github.com/martin-pettersson/wp-rest-api-sidebars

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards. 
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate. 
- [x] I've updated all React Native files affected by any refactorings/renamings in this PR.

PR #63738, opened 2024-07-19
Widget Editor: Remove unused values returned from 'mapSelect'

Mamaduka:
## What?
The `WidgetAreasBlockEditorProvider` was selecting `widgets` and `widgetAreas` values but never actually using them. PR removes these selections.

The selections seem to be leftovers from the PRs #24290 and #26164.

## Why?
Components should only select what they're using.

## Testing Instructions
The editor has e2e test coverage that should catch any regressions.

1. Enable a non-block theme like TT1.
2. Open the Widgets Editor.
3. Confirm it works as before.

Same.


ellatrix (approved):

ellatrix:
It would be so cool to have linting for these :)

Mamaduka:
> It would be so cool to have linting for these :)

+1. That would be nice!

Issue #31158, opened 2021-04-23
Tracking: Feedback on Query Loop Block settings


annezazu:
## What problem does this address?

As part of the fifth call for testing for the FSE Outreach Program, [the following feedback was shared](https://make.wordpress.org/test/2021/04/14/fse-program-testing-call-5-query-quest/#comment-1501)  highlighting how confusing it is to have mentions of pages showing when working with posts:

>Should Max page to show read plural Max pages to show?
>Items per Page doesn’t make a tonne of sense in a context where you might be adding multiple query blocks to a given “page”. We should consider calling this “Items to show” or similar.

This confusion led to not understanding how to use the setting itself as it's unclear what this setting actually controls particularly when using the Query Block to display posts:

>I couldn’t work out how to make Max page to show work at all. I assume there’s a way to get pagination to show up? It wasn’t clear.

## What is your proposed solution?

A few proposed solutions:
- Change the phrasing from "Max Page to Show" to something a bit more coherent/clear
- Remove "Page" language and use "Items" instead or have it be able to adjust depending on whether you're working with posts or pages. 

I'm also a bit confused as to whether this setting somehow impacts the Pagination Block and if we should note that one needs to be added. 

## Pending issues
- [ ] https://github.com/WordPress/gutenberg/issues/63027
- [ ] https://github.com/WordPress/gutenberg/issues/63028

annezazu:
More feedback on this setting as well from the [tenth call for testing for the FSE Outreach program](https://make.wordpress.org/test/2021/09/29/fse-program-testing-call-10-pattern-party/#comment-2097):

> When selecting ‘Display settings’ one can set the number of items per page. But actually you pick the number of items in the column if you choose a layout with multiple columns. May be easier to understand if ‘items per page’ would be renamed to ‘items per column’ or something similar.

Seems like the safest option might be saying something like, "Items per query" or something similar.

ntsekouras:
I commented in the post by I think the above feedback refers to `Offset` Query Loop pattern that has two Query Loop blocks inside a Columns block.

annezazu:
That's correct!

sereedmedia:
Per @annezazu's request to post my feedback here:

[...]

PR #63739, opened 2024-07-19
Make Query Loop settings more intuitive with a ToggleGroup and simplified help text

### LINKED ISSUE #63598: 'Inherit query from template' setting is confusing
The 'inherit query from template' setting asks users to comprehend complex details within the context of an already-complex block. 

> Inherit query from template
> Toggle to use the global query context that is set with the current template, such as an archive or search. Disable to customize the settings independently.

There are likely longer-term fundamental changes to make that would eliminate the need for this setting, but that doesn't mean we can't attempt to make the control more intuitive in the short term.

kmanijak:
## What?

Fixes https://github.com/WordPress/gutenberg/issues/63598. 
Swap "Inherit query from template" toggle with Contents: Default | Custom toggle group.

What | Before | After
-- | -- | --
`inherit: true` | image | CleanShot 2024-07-26 at 09 23 00
`inherit: false` | image | CleanShot 2024-07-26 at 09 24 25

## Why?

This control has confused users (and engineers) for a long time now. This is an attempt to make it clearer as per https://github.com/WordPress/gutenberg/issues/63598.

## How?

Replace `ToggleControl` with `ToggleGroupControl` component.

## Testing Instructions

1. Go to Editor and All Archives template
2. Focus on Query Loop block
3. **Expected:** There's "Contents" control as per screenshots above
4. Switch to "Custom"
5. **Expected:** All the customization controls appear
6. Switch back to "Default"
7. **Expected:** All the customization controls disappear and help text appears

### DIFF
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -146,6 +146,7 @@ export default function QueryInspectorControls( props ) {
 				<PanelBody title={ __( 'Settings' ) }>
 					{ showInheritControl && (
 						<ToggleGroupControl
+							__next40pxDefaultSize
 							label={ __( 'Query type' ) }
 							isBlock
 							onChange={ ( value ) => {
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -155,7 +155,7 @@ export default function QueryInspectorControls( props ) {
 							onChange={ ( value ) => {
 								setQuery( { inherit: !! value } );
 							} }
-							help={ !! inherit ? inheritControlHelp : '' }
+							help={ inherit && inheritControlHelp }
 							value={ !! inherit }
 						>
 							<ToggleGroupControlOption
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -100,13 +100,6 @@ export default function QueryInspectorControls( props ) {
 		return onChangeDebounced.cancel;
 	}, [ querySearch, onChangeDebounced ] );
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
-	const inheritControlLabel = __( 'Query Type' );
-	const inheritControlDefaultHelp = __(
-		'Display a list of posts or custom post types based on the current template.'
-	);
-	const inheritControlCustomHelp = __(
-		'Display a list of posts or custom post types based on specific criteria.'
-	);
 	const showPostTypeControl =
 		! inherit && isControlAllowed( allowedControls, 'postType' );
 	const postTypeControlLabel = __( 'Post type' );
@@ -153,15 +146,19 @@ export default function QueryInspectorControls( props ) {
 				<PanelBody title={ __( 'Settings' ) }>
 					{ showInheritControl && (
 						<ToggleGroupControl
-							label={ inheritControlLabel }
+							label={ __( 'Query type' ) }
 							isBlock
 							onChange={ ( value ) => {
 								setQuery( { inherit: !! value } );
 							} }
 							help={
 								inherit
-									? inheritControlDefaultHelp
-									: inheritControlCustomHelp
+									? __(
+											'Display a list of posts or custom post types based on the current template.'
+									  )
+									: __(
+											'Display a list of posts or custom post types based on specific criteria.'
+									  )
 							}
 							value={ !! inherit }
 						>
--- a/packages/block-library/src/query/edit/inspector-controls/index.js
+++ b/packages/block-library/src/query/edit/inspector-controls/index.js
@@ -6,7 +6,6 @@ import {
 	TextControl,
 	SelectControl,
 	RangeControl,
-	ToggleControl,
 	__experimentalToggleGroupControl as ToggleGroupControl,
 	__experimentalToggleGroupControlOption as ToggleGroupControlOption,
 	Notice,
@@ -101,6 +100,10 @@ export default function QueryInspectorControls( props ) {
 		return onChangeDebounced.cancel;
 	}, [ querySearch, onChangeDebounced ] );
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
+	const inheritControlLabel = __( 'Contents' );
+	const inheritControlHelp = __(
+		"Block will display items matching the page's context, e.g. category archive or search result"
+	);
 	const showPostTypeControl =
 		! inherit && isControlAllowed( allowedControls, 'postType' );
 	const postTypeControlLabel = __( 'Content type' );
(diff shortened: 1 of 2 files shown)


packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,6 +100,10 @@ export default function QueryInspectorControls( props ) {
 		return onChangeDebounced.cancel;
 	}, [ querySearch, onChangeDebounced ] );
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
+	const inheritControlLabel = __( 'Contents' );
+	const inheritControlHelp = __(
+		"Block will display items matching the page's context, e.g. category archive or search result"
```
t-hamano:
```suggestion
		"Block will display items matching the page's context, e.g. category archive or search result."
```

It is a good idea to end all help text with a period.

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -146,17 +149,24 @@ export default function QueryInspectorControls( props ) {
...
-								setQuery( { inherit: !! value } )
-							}
-						/>
+						<ToggleGroupControl
+							label={ inheritControlLabel }
+							isBlock
+							onChange={ ( value ) => {
+								setQuery( { inherit: !! value } );
+							} }
+							help={ !! inherit ? inheritControlHelp : '' }
```
t-hamano:
```suggestion
							help={ inherit && inheritControlHelp }
```

Nit

t-hamano (commented):
Thanks for the PR! 

I've provided some feedback on the code side, but overall it's working as expected.

As for the copy, let's continue to discuss what would be ideal.

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,6 +100,10 @@ export default function QueryInspectorControls( props ) {
 		return onChangeDebounced.cancel;
 	}, [ querySearch, onChangeDebounced ] );
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
+	const inheritControlLabel = __( 'Contents' );
```
richtabor: Let's go back to "Query Type" here. Although it is more technical, it's more contextual than "Contents".

packages/block-library/src/query/edit/inspector-controls/index.js
richtabor: I think "Display a list of posts or custom post types based on the current template." is simple and concise, yet contextual for developers. 

Shared here: https://github.com/WordPress/gutenberg/issues/63598#issuecomment-2241374446

richtabor: We should also use help text for when it's set to custom too, 

"Display a list of posts or custom post types based on specific criteria." 

otherwise there's layout shift, and you don't really know what "custom" does.

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,6 +100,10 @@ export default function QueryInspectorControls( props ) {
...
+	const inheritControlHelp = __(
+		"Block will display items matching the page's context, e.g. category archive or search result"
+	);
 	const showPostTypeControl =
 		! inherit && isControlAllowed( allowedControls, 'postType' );
 	const postTypeControlLabel = __( 'Content type' );
```
richtabor: Let's go ahead and change this back to "Post Type" — it's much clearer, and with the help text I suggested otherwise it's more relative. https://github.com/WordPress/gutenberg/issues/63598#issuecomment-2241374446

richtabor (changes requested):
Requesting changes based on copy feedback that I've consolidated https://github.com/WordPress/gutenberg/issues/63598#issuecomment-2241374446

richtabor:
We should also make the post type help text much simpler: 

"Select the type of content to display: posts, pages, or custom post types."

kmanijak:
Thanks @t-hamano and @richtabor! I fixed a small code improvement suggestion but haven't changed anything regarding the copy yet.

Let's continue the discussion in https://github.com/WordPress/gutenberg/issues/63598 and keep it as a single source of truth. Once everyone agrees there I'll apply the final version here and set the PR to review 🙌

jameskoster:
Seems there's some consensus around this suggestion: https://github.com/WordPress/gutenberg/issues/63598#issuecomment-2241374446

kmanijak:
> Seems there's some consensus around this suggestion: https://github.com/WordPress/gutenberg/issues/63598#issuecomment-2241374446

@jameskoster , I don't feel like everyone agreed with this one yet 🤔

jameskoster:
It feels like the best suggestion so far. Maybe it's worth updating the PR to give it a try? Sometimes it helps to interact with the control in-situ.

It's easy to change if it doesn't feel right :)

kmanijak:
@jameskoster, yeah, makes sense 😅 pushed the changes, and here's how it looks now:

https://github.com/user-attachments/assets/012576d8-7089-42d7-ae50-fccf3fc4db66

jameskoster:
What do you think @richtabor ?

I noticed some feedback here too: https://github.com/WordPress/gutenberg/issues/31158#issuecomment-2250609468. Specifically for this control, "Use template’s content selection" was suggested for the `default` selection.

richtabor:
> Use template’s content selection

I don't think this particular help text assists you with understanding. What's a content selection?

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,11 +100,18 @@ export default function QueryInspectorControls( props ) {
 		return onChangeDebounced.cancel;
 	}, [ querySearch, onChangeDebounced ] );
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
+	const inheritControlLabel = __( 'Query Type' );
+	const inheritControlDefaultHelp = __(
+		'Display a list of posts or custom post types based on the current template.'
+	);
+	const inheritControlCustomHelp = __(
+		'Display a list of posts or custom post types based on specific criteria.'
+	);
```
richtabor: I think it'd be fine to not use these as constants, but apply the labels in the controls directly. We're not reusing any of these anyhow and it makes the code a bit more readable for those less familiar.

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,11 +100,18 @@ export default function QueryInspectorControls( props ) {
 		return onChangeDebounced.cancel;
 	}, [ querySearch, onChangeDebounced ] );
 	const showInheritControl = isControlAllowed( allowedControls, 'inherit' );
+	const inheritControlLabel = __( 'Query Type' );
+	const inheritControlDefaultHelp = __(
+		'Display a list of posts or custom post types based on the current template.'
```
richtabor: One more edit: "Display a list of post types based on the current template."

packages/block-library/src/query/edit/inspector-controls/index.js
```diff
@@ -101,11 +100,18 @@ export default function QueryInspectorControls( props ) {
...
+	);
+	const inheritControlCustomHelp = __(
+		'Display a list of posts or custom post types based on specific criteria.'
```
richtabor: "Display a list of post types based on specific criteria."

richtabor (changes requested):
It's close. Small edits, and a small suggestion for code simplification. I think it'd be fine to land, then revise if needed further. It's certainly an improvement from trunk.

packages/block-library/src/query/edit/inspector-controls/index.js
richtabor: Actually scratch that, let's leave it as-is for this iteration.

packages/block-library/src/query/edit/inspector-controls/index.js
richtabor: Actually scratch this, let's leave it as it for this iteration.

jameskoster (approved):
> I think it'd be fine to land, then revise if needed further. It's certainly an improvement from trunk.

Agreed!

richtabor (approved):

paaljoachim:
Looks good! Thank you @kmanijak Karol!

PR #63745, opened 2024-07-19
Zoom Out: Fix vertical toolbar position

scruffian:
## What?
Ensures that the vertical toolbar is always sticky.

## Why?
The vertical toolbar should always be sticky!

## How?
Pass `shift: true` to the BlockPopover component.

## Testing Instructions
1. Open the Site Editor
2. Open the inserter
3. Open the patterns tab
4. Select a category
5. Zoom out mode should be engaged
6. Select a pattern
7. Check that the vertical toolbar always sticks to the top of the screen when you scroll

## Screenshots or screencast 

https://github.com/user-attachments/assets/ba08d308-7b4b-4695-940a-274ae86c1a30


MaggieCabrera (approved):
Good job figuring this one out! I couldn't get it to break again

Issue #52890, opened 2023-07-24
The RadioControl component renders invalid and not accessible markup


afercia:
### Description

While investigating the Post Status UI in teh Site editor in https://github.com/WordPress/gutenberg/pull/52885 and https://github.com/WordPress/gutenberg/pull/52634 I noticed the `RadioControl` component is open to misuse by developers and has some limitations that cause accessibility problems. Both the component and its implementations should be reviewed thoroughly.

At a first check:

- The `label` prop renders a stray label element that is not associated with any input.
- No fieldset and legend to group the radio boxes.
- No ability to set a description for each radio button (associated with aria-describedby).
- As such, in the `PageStatus` component the description has been added within the label, thus making the accessible name of the inputs very confusing: name and description should be separated.
- The `help` prop renders some additional text at the end of all radio buttons. This text is associated with all the radio buttons via aria-describedby. This may be useful in some cases but it's likely to be redundant and confusing. In most cases, each radio button will need its won description. Worth reminding aria-describedby may accept multiple ID refs.

In the screenshot below I made some visually hidden elements visible to better illustrate, and set some options:

- The RadioControl component [in use in the Site editor `PageStatus`](https://github.com/WordPress/gutenberg/blob/dc27d6b034a0d112dce0b71fb7d3d5dd8e870fe6/packages/edit-site/src/components/sidebar-edit-mode/page-panels/page-status.js#L187-L194).
- The [code example provided in the component inline documentation](https://github.com/WordPress/gutenberg/blob/dc27d6b034a0d112dce0b71fb7d3d5dd8e870fe6/packages/components/src/radio-control/index.tsx#L27-L42), actually rendered.
- The component [in the Storybook](https://wordpress.github.io/gutenberg/?path=/story/components-radiocontrol--default&args=help:This+is+the+help+text;hideLabelFromVision:false).

In all three case the HTML is invalid. The labelling isn't ideal, given the stray label.

Given the component limitations, maybe it's not a case that in the Post editor, in the very similar Post visibility UI, an ad-hoc implementation has been made, see the [PostVisibilityChoice component](https://github.com/WordPress/gutenberg/blob/dc27d6b034a0d112dce0b71fb7d3d5dd8e870fe6/packages/editor/src/components/post-visibility/index.js#L135-L161).

Ideally, the `RadioControl` component should behave more like the PostVisibilityChoice one, or at least produce valid markup, render fieldset+legend, and be more flexible to be able to replace PostVisibilityChoice.

[...]

Issue #63734, opened 2024-07-19
Update the RadioControl header design


jarekmorawski:
There's a new design for displaying help text for the itemGroup component introduced as part of https://github.com/WordPress/gutenberg/pull/63490, which repositions the help text to align with the item labels. This makes the itemGroup's help text look off because it stands out from other help texts. This dissonance is particularly visible when the last item in the group has a help text.

I propose moving the itemGroup's help text to the top, similar to the form section heading. This will clean up the UI and make referring to the help text easier while scanning longer item lists.

<img width="904" alt="image" src="https://github.com/user-attachments/assets/1019561b-300b-49e3-9836-33a47c04d494">

The new layout doesn't work with the current heading style, so we can take this opportunity to update that, too. Ideally, these changes would apply to all UI control groups: radio, checkboxes, toggles, etc.

<img width="716" alt="image" src="https://github.com/user-attachments/assets/a7af7b90-805a-4ff0-b818-edd4bcc4e866">

jameskoster:
Moving the help text associated with a radio group above the set of options seems reasonable to me, especially if we enable the addition of per-radio help text. I'd appreciate a11y feedback on this though.

I reckon we should use existing type styles for now. Most of the mockup is fine, but the 14/24 heading isn't a style I've seen used elsewhere. The 13px, 500 weight style we use in panel headings is probably fine?

I removed the Needs Dev label until we get some feedback on the a11y impact.

jarekmorawski:
> I reckon we should use existing type styles for now. Most of the mockup is fine, but the 14/24 heading isn't a style I've seen used elsewhere. The 13px, 500 weight style we use in panel headings is probably fine?

Good catch! I used the new wp-admin styles, but we should definitely use an existing style. The one we use for the panel headings looks good. I updated the OG post.

joedolson:
The position of the help text is probably improved slightly by being at the top of the panel. However, the lack of any semantic connection between the help text, heading, and the fields isn't great. Properly speaking, groups of radio buttons should be joined inside a fieldset with legend, or a container with `role="group"` and an `aria-labelledby` attribute pointing to the heading. 

The `aria-labelledby` target could probably contain both the heading and the help text.

jameskoster:
We should use the new type tokens for the text styling but otherwise I think this is good to go.

PR #63751, opened 2024-07-19
`RadioControl`: add support for option help text

### LINKED ISSUE #63735: Add help text support to individual radio button options
It'd be fantastic to add native support for a help text to the individual radio button options. We already use it in a few different places in the site editor and it will become even more important when we start building out forms and settings screens in the new admin.

image

Related to https://github.com/WordPress/gutenberg/issues/63734. cc @ciampo @mirka

ciampo:
## What?

Closes #63735

Add support for help text to individual radio options in the `RadioControl` component

## Why?

> It'd be fantastic to add native support for a help text to the individual radio button options. We already use it in a few different places in the site editor and it will become even more important when we start building out forms and settings screens in the new admin.
>
> &mdash; <cite>@jarekmorawski  in https://github.com/WordPress/gutenberg/issues/63735</cite>

## How?

- Add an extra `options[].description` property
- Render the description (help) text after the label, and use it as the input's description (via the `aria-describedby` attribute)
- Refactor the layout to a grid layout

## Follow-ups

- [ ] refactor the component to use `fieldset` to properly label the radio group (https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688420283);
- [ ] low priority, but tweak `BaseControl` so that its `StyledHelp` sub-component doesn't come with a top margin.

## Testing Instructions

- Visit the newly added "With Option Descriptions" Storybook example, make sure that the description (help) text shows as expected for each options that specifies one
- In the editor, check the places where we showed custom help text, and where now we are able to use the new `options[].description` property. Make sure that the radio inputs continue working as expected

## Screenshots or screencast 

Screenshot 2024-07-24 at 21 36 39

### DIFF
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -88,7 +88,13 @@ export function RadioControl(
 							onChange={ onChangeValue }
 							checked={ option.value === selected }
 							aria-describedby={
-								!! help ? `${ id }__help` : undefined
+								// TODO: will improve if we like this solution
+								// eslint-disable-next-line no-nested-ternary
+								!! option.helpText
+									? `${ id }-${ index }-help`
+									: !! help
+									? `${ id }__help`
+									: undefined
 							}
 							{ ...additionalProps }
 						/>
@@ -96,18 +102,16 @@ export function RadioControl(
 							className="components-radio-control__label"
 							htmlFor={ `${ id }-${ index }` }
 						>
-							<span className="components-radio-control__label-text">
-								{ option.label }
-							</span>
-							{ !! option.helpText ? (
-								<span
-									aria-hidden="true"
-									className="components-radio-control__help-text"
-								>
-									{ option.helpText }
-								</span>
-							) : null }
+							{ option.label }
 						</label>
+						{ !! option.helpText ? (
+							<span
+								id={ `${ id }-${ index }-help` }
+								className="components-radio-control__help-text"
+							>
+								{ option.helpText }
+							</span>
+						) : null }
 					</div>
 				) ) }
 			</VStack>
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -16,6 +16,7 @@ import BaseControl from '../base-control';
 import type { WordPressComponentProps } from '../context';
 import type { RadioControlProps } from './types';
 import { VStack } from '../v-stack';
+import { Text } from '../text';
 
 /**
  * Render a user interface to select the user type using radio inputs.
@@ -105,12 +106,14 @@ export function RadioControl(
 							{ option.label }
 						</label>
 						{ !! option.helpText ? (
-							<span
+							<Text
+								variant="muted"
+								size={ 12 }
 								id={ `${ id }-${ index }-help` }
 								className="components-radio-control__help-text"
 							>
 								{ option.helpText }
-							</span>
+							</Text>
 						) : null }
 					</div>
 				) ) }
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -16,7 +16,7 @@ import BaseControl from '../base-control';
 import type { WordPressComponentProps } from '../context';
 import type { RadioControlProps } from './types';
 import { VStack } from '../v-stack';
-import { Text } from '../text';
+import { StyledHelp } from '../base-control/styles/base-control-styles';
 
 // This is the id that BaseControl assigns to the help text element.
 function generateHelpTextId( id: string ) {
@@ -114,14 +114,13 @@ export function RadioControl(
 							{ option.label }
 						</label>
 						{ !! option.description ? (
-							<Text
-								variant="muted"
-								size={ 12 }
+							<StyledHelp
+								__nextHasNoMarginBottom
 								id={ generateOptionDescriptionId( id, index ) }
 								className="components-radio-control__option-description"
 							>
 								{ option.description }
-							</Text>
+							</StyledHelp>
 						) : null }
 					</div>
 				) ) }
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -83,7 +83,14 @@ export function RadioControl(
 			help={ help }
 			className={ clsx( className, 'components-radio-control' ) }
 		>
-			<VStack spacing={ 2 }>
+			<VStack
+				spacing={ 3 }
+				className={ clsx(
+					'components-radio-control__group-wrapper',
+					!! help &&
+						'components-radio-control__group-wrapper--with-help'
+				) }
+			>
 				{ options.map( ( option, index ) => (
 					<div
 						key={ `${ id }-${ index }` }
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -18,6 +18,15 @@ import type { RadioControlProps } from './types';
 import { VStack } from '../v-stack';
 import { Text } from '../text';
 
+// This is the id that BaseControl assigns to the help text element.
+function generateHelpTextId( id: string ) {
+	return `${ id }__help`;
+}
+
+function generateOptionDescriptionId( id: string, index: number ) {
+	return `${ id }-${ index }-option-description`;
+}
+
 /**
  * Render a user interface to select the user type using radio inputs.
  *
@@ -88,15 +97,14 @@ export function RadioControl(
 							value={ option.value }
 							onChange={ onChangeValue }
 							checked={ option.value === selected }
-							aria-describedby={
-								// TODO: will improve if we like this solution
-								// eslint-disable-next-line no-nested-ternary
+							aria-describedby={ [
 								!! option.description
-									? `${ id }-${ index }-option-description`
-									: !! help
-									? `${ id }__help`
-									: undefined
-							}
+									? generateOptionDescriptionId( id, index )
+									: undefined,
+								!! help ? generateHelpTextId( id ) : undefined,
+							]
+								.filter( Boolean )
+								.join( ' ' ) }
 							{ ...additionalProps }
 						/>
 						<label
@@ -109,7 +117,7 @@ export function RadioControl(
 							<Text
 								variant="muted"
 								size={ 12 }
-								id={ `${ id }-${ index }-option-description` }
+								id={ generateOptionDescriptionId( id, index ) }
 								className="components-radio-control__option-description"
 							>
 								{ option.description }
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -91,8 +91,8 @@ export function RadioControl(
 							aria-describedby={
 								// TODO: will improve if we like this solution
 								// eslint-disable-next-line no-nested-ternary
-								!! option.helpText
-									? `${ id }-${ index }-help`
+								!! option.description
+									? `${ id }-${ index }-option-description`
 									: !! help
 									? `${ id }__help`
 									: undefined
@@ -105,14 +105,14 @@ export function RadioControl(
 						>
 							{ option.label }
 						</label>
-						{ !! option.helpText ? (
+						{ !! option.description ? (
 							<Text
 								variant="muted"
 								size={ 12 }
-								id={ `${ id }-${ index }-help` }
-								className="components-radio-control__help-text"
+								id={ `${ id }-${ index }-option-description` }
+								className="components-radio-control__option-description"
 							>
-								{ option.helpText }
+								{ option.description }
 							</Text>
 						) : null }
 					</div>
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -95,11 +95,9 @@ export function RadioControl(
 		>
 			<VStack
 				spacing={ 3 }
-				className={ clsx(
-					'components-radio-control__group-wrapper',
-					!! help &&
-						'components-radio-control__group-wrapper--with-help'
-				) }
+				className={ clsx( 'components-radio-control__group-wrapper', {
+					'has-help': !! help,
+				} ) }
 			>
 				{ options.map( ( option, index ) => (
 					<div
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -96,7 +96,17 @@ export function RadioControl(
 							className="components-radio-control__label"
 							htmlFor={ `${ id }-${ index }` }
 						>
-							{ option.label }
+							<span className="components-radio-control__label-text">
+								{ option.label }
+							</span>
+							{ !! option.helpText ? (
+								<span
+									aria-hidden="true"
+									className="components-radio-control__help-text"
+								>
+									{ option.helpText }
+								</span>
+							) : null }
 						</label>
 					</div>
 				) ) }
--- a/packages/components/src/radio-control/index.tsx
+++ b/packages/components/src/radio-control/index.tsx
@@ -16,15 +16,15 @@ import BaseControl from '../base-control';
 import type { WordPressComponentProps } from '../context';
 import type { RadioControlProps } from './types';
 import { VStack } from '../v-stack';
+import { useBaseControlProps } from '../base-control/hooks';
 import { StyledHelp } from '../base-control/styles/base-control-styles';
 
-// This is the id that BaseControl assigns to the help text element.
-function generateHelpTextId( id: string ) {
-	return `${ id }__help`;
+function generateOptionDescriptionId( radioGroupId: string, index: number ) {
+	return `${ radioGroupId }-${ index }-option-description`;
 }
 
-function generateOptionDescriptionId( id: string, index: number ) {
-	return `${ id }-${ index }-option-description`;
+function generateOptionId( radioGroupId: string, index: number ) {
+	return `${ radioGroupId }-${ index }`;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/components/src/radio-control/index.tsx
```diff
@@ -88,7 +88,13 @@ export function RadioControl(
...
 							checked={ option.value === selected }
 							aria-describedby={
-								!! help ? `${ id }__help` : undefined
+								// TODO: will improve if we like this solution
+								// eslint-disable-next-line no-nested-ternary
+								!! option.helpText
+									? `${ id }-${ index }-help`
+									: !! help
+									? `${ id }__help`
+									: undefined
```
ciampo: Currently, the option's help text is rendered separately from the `label`, and is used to describe the associated `input` element .

Since the help text is not part of the `label`, clicking it won't select the associated radio.

We could improve this by moving the help text inside the label, but that would make the help text content as part of the label, instead of a description text. We could set `aria-hidden="true"` on the help text inside the label if we don't want it to "pollute" the input label text, and we could also optionally add a copy of the same text as visually hidden and still use it for the description text of the input.

But before trying this out, I thought I'd start simple and wait for feedback

packages/components/src/radio-control/index.tsx
```diff
@@ -98,6 +104,14 @@ export function RadioControl(
 						>
 							{ option.label }
 						</label>
+						{ !! option.helpText ? (
+							<span
+								id={ `${ id }-${ index }-help` }
+								className="components-radio-control__help-text"
+							>
```
ciampo, line 111: TODO: try out `<Text variant="muted" size={ 12 }>` instead

packages/components/src/radio-control/index.tsx
```diff
@@ -98,6 +105,16 @@ export function RadioControl(
 						>
 							{ option.label }
 						</label>
+						{ !! option.helpText ? (
+							<Text
+								variant="muted"
+								size={ 12 }
+								id={ `${ id }-${ index }-help` }
+								className="components-radio-control__help-text"
+							>
```
mirka: Ideally some of these typography/color styles should be shared with the [help styles in BaseControl](https://github.com/WordPress/gutenberg/blob/bced5e0bebb322d6ddb47dafae52fedef9943315/packages/components/src/base-control/styles/base-control-styles.ts#L62) so they stay in sync.

packages/components/src/radio-control/types.ts
```diff
@@ -24,6 +24,10 @@ export type RadioControlProps = Pick<
 		 * The internal value compared against select and passed to onChange
 		 */
 		value: string;
+		/**
+		 * Optional help text to be shown in addition the label.
+		 */
+		helpText?: string;
```
mirka: I'm torn on whether we should do this, or make it `help?: ReactNode` to have consistency with the BaseControl family APIs ☹️

packages/components/src/radio-control/index.tsx
mirka: > Since the help text is not part of the `label`, clicking it won't select the associated radio.

I'd say this is the desired behavior. Clicking on `help` text doesn't (and I'd say shouldn't) focus the control (same in InputControl, CheckboxControl, etc).

mirka: This could be a good opportunity to fix #52890, by changing this component to a proper `fieldset` markup and associating the group `help` to that.

If that's out of scope for this PR though, it might be better to keep both the group-level and option-level `help`s associated if they both exist (apparently you can specify [multiple space-delimited ids](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby#values)).

packages/editor/src/components/site-discussion/index.js
```diff
@@ -23,29 +23,16 @@ import { store as editorStore } from '../../store';
...
-				<Text variant="muted" size={ 12 }>
-					{ __( 'Existing comments remain visible.' ) }
-				</Text>
-			</>
-		),
+		label: __( 'Closed' ),
 		value: '',
+		helpText: `${ __(
+			'Visitors cannot add new comments or replies.'
+		) } ${ __( 'Existing comments remain visible.' ) }`,
```
tyxla: These used to be two separate text elements (resulting in 2 separate spans that have `display: block`), which displayed them on multiple lines. This will no longer be the case, and that's potentially undesired.

The difference will manifest in some mobile viewports:

|Before|After|
|---|---|
|Screenshot 2024-07-23 at 16 57 24|Screenshot 2024-07-23 at 16 59 10|

The potential issue with that is that translation-wise, it might potentially add confusion. Ideally, we would avoid that change.

On a higher level, the `helpText` prop seems to be a bit limiting since we have evidence that we need more than one string. I wonder if we should take that into consideration while designing it. 

Or perhaps this is not an issue and we're fine with keeping those 2 sentences on the same line. Maybe consult a design person to be sure?

packages/components/src/radio-control/index.tsx
```diff
@@ -88,7 +89,13 @@ export function RadioControl(
 							onChange={ onChangeValue }
 							checked={ option.value === selected }
 							aria-describedby={
-								!! help ? `${ id }__help` : undefined
+								// TODO: will improve if we like this solution
+								// eslint-disable-next-line no-nested-ternary
```
tyxla: If there are any ways to avoid the nested ternaries, let's do it, including if it requires us to change the `map` return syntax. Nested ternaries are discouraged for a good reason - readability and clarity can be diminished.

ciampo: Yup, I just didn't invest time refactoring the nested ternary because I wanted to make sure the current approach was going in the right direction (see https://github.com/WordPress/gutenberg/pull/63751#discussion_r1684375290). The nested ternary will be gone in any case before final code review.

packages/components/src/radio-control/types.ts
ciampo: Making it a `ReactNode` would align with `BaseControl` and would also allow for [breaking help text into separate paragraphs](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688124771).

I'm just hesitant to allow consumers to pass any markup as the help text — but I guess really determined consumers would find a way anyway, like they already do to pass the help text without an official API.

If we go ahead with the change, should we rename the property to `description` to make it more generic (no "text" mentioned) and to show its real usage (describing the radio input)

packages/editor/src/components/site-discussion/index.js
ciampo: Good point. This could be easily achieved if we allowed any `ReactNode` instead of only `string`s (see [related conversation](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1686833039)).

Let's just double-check with @WordPress/gutenberg-design that having the help text for a single option span multiple paragraphs is ok.

jameskoster: I don't think that behavior is intentional, it looks a little strange. Help text should be optimised for as few characters as possible. Multiple paragraphs isn't ideal, especially in a radio group that can have it's own separate help text. This constraint feels like an improvement to me :D

packages/components/src/radio-control/types.ts
ciampo: Update: breaking text into separate paragraphs is actually something that [we should discourage](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688327043).

The safest approach to me seems to be:
- to rename `helpText` to `description` (more generic and future-proof)
- to keep the type as `string` (it can always be widened to `ReactNode` later, if needed

mirka: > If we go ahead with the change, should we rename the property to `help`, since users can pass any markup (and not just text) ?

Yes, I think so. And we could prevent some misuse by documenting the best practices, like in BaseControl itself:

https://github.com/WordPress/gutenberg/blob/495ad8c5fe99b43878b854a9e97d3b477acff885/packages/components/src/base-control/types.ts#L24

We could also add a sentence about it being intended for descriptions that can be experienced as plain text ([from MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby#description)):

> The `aria-describedby` property is appropriate when the associated content contains descriptions experienced as plain text.

packages/components/src/radio-control/index.tsx
ciampo: > This could be a good opportunity to fix https://github.com/WordPress/gutenberg/issues/52890, by changing this component to a proper fieldset markup and associating the group help to that.

That would be great, but do you know if `BaseControl` supports using `fieldset` ?

> If that's out of scope for this PR though, it might be better to keep both the group-level and option-level helps associated if they both exist (apparently you can specify [multiple space-delimited ids](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby#values)).

That's what I've done for now.

packages/components/src/radio-control/index.tsx
ciampo: Done — although it's annoying that `StyledHelp` applies a top margin. As a low-priority follow-up, we should move the margin somewhere else in `BaseControl` (ideally as the `gap` of a flex wrapper ?)

packages/components/src/radio-control/types.ts
ciampo: Ah, I edited my original comment but you were too quick to reply! In the meantime, I named the prop `description` and kept it as `string`.

If you think that `help` is a better name and/or `ReactNode` is a better type, I'm happy to make those changes.

packages/editor/src/components/site-discussion/index.js
ciampo: Sounds good. I'll make sure that the text in question renders without line breaks, regardless of which underlying type we pick for the help text.

tyxla: Thanks @jameskoster! That makes it simple, then - let's go with a single string. Ideally, this should actually be a single string, but changing it to one would invalidate the existing translations, so maybe we can leave with it as you suggested.

packages/components/src/radio-control/index.tsx
```diff
@@ -16,6 +16,16 @@ import BaseControl from '../base-control';
 import type { WordPressComponentProps } from '../context';
 import type { RadioControlProps } from './types';
 import { VStack } from '../v-stack';
+import { StyledHelp } from '../base-control/styles/base-control-styles';
+
+// This is the id that BaseControl assigns to the help text element.
+function generateHelpTextId( id: string ) {
+	return `${ id }__help`;
+}
```
ciampo: This is hacky, but it's already present on `trunk`. Ideally `BaseControl` should be more flexible.

ciampo:
Update: I addressed most of the feedback.

Pending items:

- [x] name and type of the new `options` property ([conversation](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688423860))
- ~potentially tweak the component to use `fieldset` semantics ([conversation](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688420283))~
- [x] add unit tests (this component doesn't actually have any!)
- [x] get a design review — mostly on the vertical spacing of each option and its help text (cc @WordPress/gutenberg-design )

Follow-up:
- if not done in this PR, refactor the component to use `fieldset` to properly label the radio group ([conversation](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688420283))
- low priority, but tweak `BaseControl` so that its `StyledHelp` component doesn't come with a margin-top

packages/components/src/radio-control/index.tsx
mirka: > That would be great, but do you know if `BaseControl` supports using `fieldset` ?

No, we currently need to do custom implementations using `BaseControl.VisualLabel`. But baking `fieldset` support into BaseControl might be something to consider! (Not immediately sure if that's a good idea)

ciampo: > No, we currently need to do custom implementations using `BaseControl.VisualLabel`.

Ok, this definitely sounds to me like a task that should have its dedicated PR. I will not work on this in this PR.

> But baking `fieldset` support into BaseControl might be something to consider! (Not immediately sure if that's a good idea)

I think we could give it a try, in a new version of `BaseControl`. I had open a [related issue](https://github.com/WordPress/gutenberg/issues/42630) some time ago, and I think Andrew at some point even came up with a proof of concept.

jameskoster:
> get a design review — mostly on the vertical spacing of each option and its help text (cc @WordPress/gutenberg-design )

I reckon we should align with the recent updates to CheckboxControl and ToggleControl, so let's try 8px between label + help text. We can also increase the gap between radios to 12px.

One other observation; when no help text is supplied, the RadioControl height is increased by 4px, probably due to the `row-gap`. Perhaps best described with a screenshot:

Screenshot 2024-07-24 at 12 59 22

Ideally `.components-radio-control__option` should hug its contents. IE be 16px tall when no help text is supplied.

packages/components/src/radio-control/types.ts
mirka: Ok yes, I think that makes sense. Let's go with `description?: string`.

ciampo:
@jameskoster thank you for the feedback — I changed the horizontal gap between radio and label to `12px`, changed the vertical gap between label and help text to `8px`, and removed the extra `row-gap` when there is no help text.

| Before | After |
|---|---|
| Screenshot 2024-07-24 at 18 01 11 | Screenshot 2024-07-24 at 17 57 52 |

Personally, it feels like the vertical space if a bit off — the gap between the label and help text seems too large, compared to the gap between each radio option and between the radio group and the group help text. But this is just my opinion, and I'd be fine with keeping things as they are.

jameskoster:
Hmm yes that doesn't look quite right. 

The gap between label + help text should be consistent across radios, checkboxes, and toggles. In this context I agree it looks a little large, perhaps we reduce that to `4px` across?

In Storybook I'm only seeing an `8px` gap between radios.

Increasing to `12px`, in conjunction with the `4px` gap between label + help text seems to work well. What do you think?

Screenshot 2024-07-24 at 17 30 56

ciampo:
Agreed. Pushed a commit with the tweaked spacing

Screenshot 2024-07-24 at 21 36 39

ciampo:
This PR is ready for another round of review.

richtabor:
Curious. Does help text at the bottom of a set of radio buttons make sense? Seems more connective to the label than to the end of the elements.

jarekmorawski:
> Curious. Does help text at the bottom of a set of radio buttons make sense? Seems more connective to the label than to the end of the elements.

💯 That'll be changed in https://github.com/WordPress/gutenberg/issues/63734.

jameskoster:
Looking good to me. I think the gap between the heading and the group could be increased, but it's probably fine to do that in https://github.com/WordPress/gutenberg/issues/63734.

Would it make sense to update the gap between label + help text for checkboxes and toggles here too? I wonder if we should add a variable for this...

packages/components/src/radio-control/index.tsx
mirka: You know what I just realized we never use but should? [`useBaseControlProps()`](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/base-control/hooks.ts)

packages/components/src/radio-control/index.tsx
```diff
@@ -73,7 +83,14 @@ export function RadioControl(
...
 			className={ clsx( className, 'components-radio-control' ) }
 		>
-			<VStack spacing={ 2 }>
+			<VStack
+				spacing={ 3 }
+				className={ clsx(
+					'components-radio-control__group-wrapper',
+					!! help &&
+						'components-radio-control__group-wrapper--with-help'
+				) }
```
mirka:
```suggestion
				className={ clsx( 'components-radio-control__group-wrapper', {
					'has-help': !! help,
				} ) }
```

For better adherence to our [modifier conventions](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/#naming).

packages/components/src/radio-control/index.tsx
```diff
@@ -87,9 +104,14 @@ export function RadioControl(
...
-								!! help ? `${ id }__help` : undefined
-							}
+							aria-describedby={ [
+								!! option.description
+									? generateOptionDescriptionId( id, index )
+									: undefined,
+								!! help ? generateHelpTextId( id ) : undefined,
+							]
+								.filter( Boolean )
+								.join( ' ' ) }
```
mirka: We can actually just use `clsx` for this... 🤔 Too weird?

packages/components/src/radio-control/test/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+const defaultPropsWithDescriptions = {
+	...defaultProps,
+	options: defaultProps.options.map( ( option, index ) => ( {
+		...option,
+		description: `This is the option number ${ index + 1 }.`,
+	} ) ),
+};
+
+describe.each( [
+	// [ 'uncontrolled', RadioControl ],
```
mirka, line 53: Let's leave a TODO comment here so it's clear that it's still unsupported, not that we forgot to uncomment the line before merging or something 😂

packages/components/src/radio-control/index.tsx
ciampo: Good point! I used it in https://github.com/WordPress/gutenberg/pull/63751/commits/4a6087ada724ebc5da01334a9f4b714bb5a234f3 and it seems to work well

packages/components/src/radio-control/index.tsx
ciampo: Kind of weird, but I don't mind personally. Used it in https://github.com/WordPress/gutenberg/pull/63751/commits/4a6087ada724ebc5da01334a9f4b714bb5a234f3

packages/components/src/radio-control/index.tsx
ciampo: I had forgotten about it — thank you for pointing out. Done in https://github.com/WordPress/gutenberg/pull/63751/commits/46fb3e14b7c2fe6ee08e941b0146813f94314608

packages/components/src/radio-control/test/index.tsx
ciampo: https://github.com/WordPress/gutenberg/pull/63751/commits/391bc27ebf31c36bfc09526a516b5c29992cb307

ciampo:
@mirka all existing feedback should have beeen addressed.

@jameskoster 

> I think the gap between the heading and the group could be increased, but it's probably fine to do that in https://github.com/WordPress/gutenberg/issues/63734.
> Would it make sense to update the gap between label + help text for checkboxes and toggles here too? I wonder if we should add a variable for this...

I'd prefer to merge this PR soon and focus on tweaking spacing and/or aligning it with other components in separate PRs.

mirka (approved):
Looks good 👍

Issue #56244, opened 2023-11-17
Make it easier to find and edit the "Home" page in the editor


getdave:
A critical user task is editing their homepage and is often regarded as the top priority for users.

However, the current WordPress Site Editor presents challenges in locating the current "Homepage", leading to user confusion which I have observed in recent testing sessions. 

The root issue lies in the ambiguity of the terms we utilise to label the "Homepage" within the editor.

### Problem Description:

WordPress currently exposes users to the underlying template hierarchy, causing confusion. 

Depending on the _theme's_ templates, the homepage may be labeled as "Blog Home" or "Front Page," introducing unnecessary technical terminology to the majority (though obviously not all) users.

### Proposed Changes:

Ensure `Home` is always used to refer to the "Homepage".

We may be able do this as follows:

1. **Consistent Labeling:**
   - Change the "Home" Page item label to always be "Home," _irrespective_ of the template used for rendering.

2. **Reduce prominence of Template information:**
   - Move the template name into the Page's details panel under a "Template" heading. 
   - This removes technical "noise" for the "80%" of users who don't need this information.
   - Still allows more technically inclined users to understand the template in use.

3. **Warning for "Template Impact":**
   - If the template in use affects other pages (e.g., `page.html` or `index.html`), display a small warning notice in the details panel
   - Include a button to allow users to automatically create a custom template _specific_ to the homepage (e.g., `front-page.html`) so that they can avoid making unintended changes.

### Rationale:

- The proposed changes aim to eliminate needless distractions caused by technical terminology for users who are not familiar with or concerned about the template hierarchy.
- In a block-based environment where all elements are editable in the Site Editor, the utility of the template hierarchy in regard to the Homepage is diminished.
- Enhancing the details panel with template information caters to more technical users while maintaining a clear and consistent user interface for the majority.

### Related

- https://github.com/WordPress/gutenberg/issues/52335

getdave:
Also for reference here is some information on the template hierarchy and how it applies specifically to the "Homepage" of a given site. This was generated during a chat with ChatGPT:

--------

In block themes, the determination of what affects the homepage follows a structure similar to traditional themes but with HTML file extensions. Here's a summary:

1. **Default Behavior:**
   - Block themes, like traditional themes, set the site's homepage to display the latest blog posts by default.
   - The template file `home.html` is used to render the blog posts index.

2. **Template Hierarchy for the Front Page:**
   - The `front-page.html` template file takes precedence over `home.html` if it exists.
[...]

PR #63752, opened 2024-07-19
Update 'Front page' badge

jameskoster:
Discussed in https://github.com/WordPress/gutenberg/issues/56244

This PR does two things:

1. Updated the 'Front page' badge to read 'Homepage' which is more familiar, and echoes the reading settings
2. Ensures the badges have a consistent appearance across data view layouts and the page inspector

| Before | After |
| --- | --- |
| Screenshot 2024-07-19 at 14 49 28 | Screenshot 2024-07-19 at 14 49 51 |

### Testing instructions

1. Ensure your site has a static homepage
2. Go to the Pages data view in the Site Editor and locate the homepage
4. Observe it includes a badge with label: "Homepage"
5. Switch between different data view layouts and observe that this badge has the same appearance across (20px tall, 4px horizontal padding)
6. Enter the Editor and ensure the "Homepage" badge in the Inspector also matches the appearance in data views

### DIFF
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -428,7 +428,7 @@ export default function PostList( { postType } ) {
 					if ( item.id === frontPageId ) {
 						suffix = (
 							<span className="edit-site-post-list__title-badge">
-								{ __( 'Front Page' ) }
+								{ __( 'Homepage' ) }
 							</span>
 						);
 					} else if ( item.id === postsPageId ) {
--- a/packages/edit-site/src/components/post-list/style.scss
+++ b/packages/edit-site/src/components/post-list/style.scss
@@ -72,6 +72,7 @@
 	font-size: 12px;
 	font-weight: 400;
 	flex-shrink: 0;
+	line-height: $grid-unit-05 * 5;
 }
 
 .edit-site-post-list__status-icon {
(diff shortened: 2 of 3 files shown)


richtabor:
This works as expected, but I do note that it's a tad strange we have "Homepage" and "Posts page". 

CleanShot 2024-07-21 at 18 39 16

jameskoster:
I find "Posts page" somewhat unintuitive in isolation, but that's how the page is referred to in reading settings. I don't know that we should diverge from that without updating holistically – that can make things more confusing. I also don't think it should hold up the change in this PR?

Or are you saying it's strange to mark this page at all?

richtabor:
No, just the semantics of "Post page" verses the one-word "Homepage". Perhaps worth revisiting "Homepage" to "Home page", but not a big deal. Let's get this in as it does clear things up a bit.

richtabor (approved):
LGTM

karmatosed:
I would advise we revisit 'Home page' to be similar to 'Posts Page' as a fix going forward. I also think we need to check what translation implication 'home' has perhaps?

jameskoster:
If so, it should probably be done holistically, capturing UIs such as this one:

Screenshot 2024-07-22 at 15 38 50

richtabor:
Yes, that's why it's not changed here-to map to the label provided in core.

PR #63754, opened 2024-07-19
Tabs: keep full opacity of focus ring on disabled tabs

ciampo:
## What?

Part of #52997

Inspired by #62480, this PR makes sure that the focus ring on disabled `Tabs.Tab` components is fully opaque, and that the text color of disabled tabs doesn't change on hover. 

## Why?

The focus ring on disabled tabs should have full contrast, and disabled tabs should not show "interactive" hints on hover since they're not interactive.

## How?

- By not setting a low `opacity` on disabled tabs, and instead changing the text color to achieve a similar result.
- By enabling hover styles only for non-disabled tabs

## Testing Instructions

- Visit the "disabled" `Tabs` Storybook example
- Using the arrow keys, move focus on the disabled tab
- Make sure that the focus ring is fully opaqu

## Screenshots or screencast 

| Before (trunk) | After (trunk) |
|---|---|
| Screenshot 2024-07-19 at 17 04 05 | Screenshot 2024-07-22 at 09 49 10 |

### DIFF
--- a/packages/components/src/tabs/styles.ts
+++ b/packages/components/src/tabs/styles.ts
@@ -76,10 +76,11 @@ export const Tab = styled( Ariakit.Tab )`
 		font-weight: 500;
 		text-align: inherit;
 		hyphens: auto;
+		color: ${ COLORS.theme.foreground };
 
 		&[aria-disabled='true'] {
 			cursor: default;
-			opacity: 0.3;
+			color: ${ COLORS.theme.gray[ 700 ] };
 		}
 
 		&:hover {
--- a/packages/components/src/tabs/styles.ts
+++ b/packages/components/src/tabs/styles.ts
@@ -80,7 +80,7 @@ export const Tab = styled( Ariakit.Tab )`
 
 		&[aria-disabled='true'] {
 			cursor: default;
-			color: ${ COLORS.theme.gray[ 700 ] };
+			color: ${ COLORS.ui.textDisabled };
 		}
 
 		&:hover {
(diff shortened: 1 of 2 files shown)


packages/components/src/tabs/styles.ts
```diff
@@ -76,10 +76,11 @@ export const Tab = styled( Ariakit.Tab )`
 		font-weight: 500;
 		text-align: inherit;
 		hyphens: auto;
+		color: ${ COLORS.theme.foreground };
 
 		&[aria-disabled='true'] {
 			cursor: default;
-			opacity: 0.3;
+			color: ${ COLORS.theme.gray[ 700 ] };
```
tyxla: Is it intentional that we're making the text so close to the non-disabled tab text color now? I'd likely prefer a lighter gray, but as you said, let's ask for some design feedback.

ciampo: Yup! Looking at https://github.com/WordPress/gutenberg/pull/62480#discussion_r1684493572, there seems to be consensus on the lighter `600` shade. I will update accordingly.

ciampo: Updated, using a semantic variable that aliases to gray 600.

ciampo:
Removing the "Needs design feedback" label since the relevant feedback was already provided in https://github.com/WordPress/gutenberg/pull/62480#discussion_r1684493572

This PR is ready for a final review.

tyxla (approved):
LGTM 👍

PR #63758, opened 2024-07-19
Revert "Downgrade node 22(.5) unit tests to 22.4 (#63728)"

desrosj:
This reverts commit 32b4b2ec8acf8dc5e5eca4efbf6ed0ad6be3c3de.

A bug fix release. [22.5.1 has been published](https://github.com/nodejs/node/releases/tag/v22.5.1) that hopefully fixes the problems that were causing failed builds earlier this week.


desrosj:
It seems that the [version of node has been published](https://github.com/actions/node-versions) for actions/node-versions, but there's likely still some deploying that is in progress or needs to happen.

desrosj:
Looks like things are back green again. I reverted the merge commit wholesale, but I'm not opposed to keeping the changes around introducing a `name` for each value. Though those change anyway when we change the version in the matrix, so it doesn't solve the same problem.

talldan (approved):
Thanks @desrosj for remembering to do this 😄

PR #48498, opened 2023-02-27
Update HeightControl component to label inputs

andrewhayward:
## What?
Adds accessible names to both inputs in the `HeightControl` component, replacing the `fieldset` wrapper and using `aria-controls` instead.

## Why?
Currently, `HeightControl` uses a `<fieldset>` to label the input fields, but this doesn't provide an accessible name for either. [Both fields should have an appropriate name](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions).

## How?
* Updates the `RangeControl` component to accept an optional ID.
* Replaces the `<fieldset>`/`<legend>` combination with a `<label>`, pointing at the text input, to ensure it has an accessible name.
* The range input references the `<label>` with an `aria-labelledby` to get its own accessible name.
* The range input gains an `aria-controls` attribute, pointing at the text input (while acknowledging that `aria-controls` has little support at the time of writing).

## Testing Instructions
* Launch storybook: `npm run storybook:dev`
* Go to the `HeightControl` component
* Check that both fields have accessible names (using the inspector)

### Testing Instructions for Keyboard
Nothing has functionally changed, so both inputs should remain as expected

PR #63761, opened 2024-07-19
Update HeightControl component to label inputs

ciampo:
This is a copy of #48498 from @andrewhayward , moved to a branch on this repo (instead than a fork).

## What?

Adds accessible names to both inputs in the `HeightControl` component, replacing the `fieldset` wrapper and using `aria-controls` instead.

## Why?

Currently, `HeightControl` uses a `<fieldset>` to label the input fields, but this doesn't provide an accessible name for either. [Both fields should have an appropriate name](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions).

## How?

* Updates the `RangeControl` component to accept an optional ID.
* Replaces the `<fieldset>`/`<legend>` combination with a `<label>`, pointing at the text input, to ensure it has an accessible name.
* The range input references the `<label>` with an `aria-labelledby` to get its own accessible name.
* The range input gains an `aria-controls` attribute, pointing at the text input (while acknowledging that `aria-controls` has little support at the time of writing).

## Testing Instructions

* Launch storybook: `npm run storybook:dev`
* Go to the `HeightControl` component
* Check that both fields have accessible names (using the inspector)

Nothing has functionally changed, so both inputs should remain as expected

### DIFF
--- a/packages/block-editor/src/components/height-control/index.js
+++ b/packages/block-editor/src/components/height-control/index.js
@@ -1,6 +1,7 @@
 /**
  * WordPress dependencies
  */
+import { useInstanceId } from '@wordpress/compose';
 import { useMemo } from '@wordpress/element';
 import {
 	BaseControl,
@@ -68,6 +69,10 @@ export default function HeightControl( {
 	value,
 } ) {
 	const customRangeValue = parseFloat( value );
+	const id = useInstanceId( HeightControl, 'inspector-height-control' );
+	const labelId = `${ id }__label`;
+	const inputId = `${ id }__input`;
+	const rangeId = `${ id }__range`;
 
 	const [ availableUnits ] = useSettings( 'spacing.units' );
 	const units = useCustomUnits( {
@@ -144,13 +149,14 @@ export default function HeightControl( {
 	};
 
 	return (
-		<fieldset className="block-editor-height-control">
-			<BaseControl.VisualLabel as="legend">
+		<div className="block-editor-height-control" id={ id }>
+			<BaseControl.VisualLabel as="label" for={ inputId } id={ labelId }>
 				{ label }
 			</BaseControl.VisualLabel>
 			<Flex>
 				<FlexItem isBlock>
 					<UnitControl
+						id={ inputId }
 						value={ value }
 						units={ units }
 						onChange={ onChange }
@@ -164,6 +170,9 @@ export default function HeightControl( {
 				<FlexItem isBlock>
 					<Spacer marginX={ 2 } marginBottom={ 0 }>
 						<RangeControl
+							aria-controls={ inputId }
+							aria-labelledby={ labelId }
+							id={ rangeId }
 							value={ customRangeValue }
 							min={ 0 }
 							max={
@@ -183,6 +192,6 @@ export default function HeightControl( {
 					</Spacer>
 				</FlexItem>
 			</Flex>
-		</fieldset>
+		</div>
 	);
 }
--- a/packages/block-editor/src/components/height-control/index.js
+++ b/packages/block-editor/src/components/height-control/index.js
@@ -69,7 +69,7 @@ export default function HeightControl( {
 	value,
 } ) {
 	const customRangeValue = parseFloat( value );
-	const id = useInstanceId( HeightControl, 'inspector-height-control' );
+	const id = useInstanceId( HeightControl, 'block-editor-height-control' );
 	const labelId = `${ id }__label`;
 	const inputId = `${ id }__input`;
 	const rangeId = `${ id }__range`;
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -18,6 +18,7 @@
 -   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
 -   `SelectControl`: Add `"minimal"` variant ([#63265](https://github.com/WordPress/gutenberg/pull/63265)).
 -   `FontSizePicker`: tidy up internal logic ([#63553](https://github.com/WordPress/gutenberg/pull/63553)).
+-   `RangeControl`: Allow external `id` prop ([#63761](https://github.com/WordPress/gutenberg/pull/63761)).
 
 ### Internal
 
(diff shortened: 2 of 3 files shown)


ciampo:
Hey folks, while going through my old notifications I found #48498, which was approved but never got merged.

I made a clone of it, so that we can take another look and merge if necessary.

Not sure if @alexstine is still around to confirm his original approving review — tagging also @WordPress/gutenberg-components @afercia and @joedolson for some feedback

packages/components/src/range-control/index.tsx
```diff
@@ -41,6 +41,15 @@ import { space } from '../utils/space';
 
 const noop = () => {};
 
+function useUniqueId( idProp?: string ) {
+	const id = useInstanceId(
+		UnforwardedRangeControl,
+		'inspector-range-control'
```
tyxla, line 47: I wonder what this component has to do with `inspector` though. Maybe not much anymore.

packages/components/src/range-control/index.tsx
```diff
@@ -41,6 +41,15 @@ import { space } from '../utils/space';
...
+	);
+
+	return idProp || id;
```
tyxla, line 50: Do we mean to use `||` or `??`?

packages/components/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 -   `CustomSelectControl`: Stabilize `__experimentalShowSelectedHint` and `options[]. __experimentalHint` props ([#63248](https://github.com/WordPress/gutenberg/pull/63248)).
 -   `SelectControl`: Add `"minimal"` variant ([#63265](https://github.com/WordPress/gutenberg/pull/63265)).
 -   `FontSizePicker`: tidy up internal logic ([#63553](https://github.com/WordPress/gutenberg/pull/63553)).
+-   `RangeControl`: Allow external `id` prop ([#48498](https://github.com/WordPress/gutenberg/pull/48498)).
```
tyxla:
```suggestion
-   `RangeControl`: Allow external `id` prop ([#63761](https://github.com/WordPress/gutenberg/pull/63761)).
```

packages/block-editor/src/components/height-control/index.js
```diff
@@ -68,6 +69,10 @@ export default function HeightControl( {
 	value,
 } ) {
 	const customRangeValue = parseFloat( value );
+	const id = useInstanceId( HeightControl, 'inspector-height-control' );
```
tyxla: Should the ID prefix just be `height-control` or `block-editor-height-control`?

tyxla (approved):
Looking good from my perspective and testing well.

Thanks to all contributors who helped with it, and for you taking care of it @ciampo 👍 🚀

packages/components/src/range-control/index.tsx
ciampo: I had that same question when reviewing the original PR, and at the time we agreed that `||` was a good choice because empty strings would not make a good `id`. https://github.com/WordPress/gutenberg/pull/48498#discussion_r1122433330

packages/components/src/range-control/index.tsx
ciampo: I think that's the instance ID prefix that many components in the package adopt 😓  Not sure how easy it would be to refactor away from it, but probably not in the scope for this PR (`'inspector-range-control'` is already used in `trunk` and in this PR it was moved to the `useUniqueId` function)

packages/block-editor/src/components/height-control/index.js
ciampo: In this case, I think we can change it to `block-editor-height-control` without consequences since it's a newly added ID (and I couldn't find other examples of `useInstanceId` with a prefix in the block editor package)

afercia:
@ciampo I was just testing this PR while it was merged. I'm a little confused, what I see now is:

- The input field has two_ associated label elements so its accessible name is doubled: `Height Height`:

In the screenshot below I made the visually hidden labels visible to better illustrate:

Screenshot 2024-07-22 at 11 24 58

The range input has now: an aria-labelledby, an aria-label, and an associated label element; Aria-labelledby overrides the other twos so the accessible name is not doubled. But 3 labeling are a bit redundant:

Screenshot 2024-07-22 at 11 25 41

On trunk, I don't see missing labeling; Boht controls have a `label` prop that by default is 'Height'.

> Currently, HeightControl uses a <fieldset> to label the input fields,

Maybe a source of confusion is that a fieldset legend _is not meant to label the input fields_. It's meant to label _the group_ reprezented by the fieldset.

packages/block-editor/src/components/height-control/index.js
```diff
@@ -144,13 +149,14 @@ export default function HeightControl( {
 	};
 
 	return (
-		<fieldset className="block-editor-height-control">
-			<BaseControl.VisualLabel as="legend">
+		<div className="block-editor-height-control" id={ id }>
+			<BaseControl.VisualLabel as="label" for={ inputId } id={ labelId }>
```
afercia, line 153: I guess this `for` should have been `htmlFor`. Getting this in the condole:
```
Warning: Invalid DOM property `for`. Did you mean `htmlFor`?
at label
at div
at HeightControl
etc...
```

afercia:
@ciampo @tyxla can we please revert this change? I'm not sure it's what we want.
Besides my previous comment, there's also a typo triggering a Warning in the console:
```
Warning: Invalid DOM property `for`. Did you mean `htmlFor`?
```

packages/components/src/range-control/index.tsx
mirka: Also, reminder that the [`useInstanceId()` util](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-compose/#useinstanceid) supports `preferredId` as the third argument. (I guess it didn't before, hence the custom logic like this in some of these older components.)

Mamaduka:
Came here to report the same warning.

ciampo:
Revert PR ready #63839

The original PR had been approved and never merged, and so I thought it would been good to go.  I'm fine just reverting and keeping things as before, for the time being.

tyxla:
Thanks for reporting, folks, I missed the warning initially 🤔 . Fine with reverting for now.

packages/block-editor/src/components/height-control/index.js
ciampo: Yup - I must have let my guard down knowing that this PR had been reviewed before. Thank you for pointing it out

packages/components/src/range-control/index.tsx
ciampo: I forgot about it — thank you for pointing it out!

Issue #63684, opened 2024-07-17
Improve state of Tag Cloud block with no tags


richtabor:
In the spirit of making placeholders more 1:1 to front-end expectations, the Tags Cloud should not have a "There's no content to show here yet" message, especially on the front-end of the site. 

Yes, its a bit of an edge case, as it requires you to not have published with any tags—but when setting up a blog for the first time, it can be disorienting to have what _should_ be a representation of the actual block within the editor, and nothing on the front-end (as there are no tags). 

### Current

As seen here in the editor, but also on the front-end: 

![CleanShot 2024-07-17 at 16 39 24](https://github.com/user-attachments/assets/a800fcca-41c7-413a-961f-ebd7eceffbf6)

### Proposed

Include sample tags within the editor, if there are no tags available; they would be styled however the theme depicts them. Perhaps something simple like "Tag-1" and "Tag-2". 
 
![CleanShot 2024-07-17 at 16 41 15](https://github.com/user-attachments/assets/2007c514-ac27-4617-95d3-cd16549d3a3d)

afercia:
It's an interesting edge case.
I'm not sure the editor and front end should represent with placeholders a state that is actually an _empty_state.
Yes, the editor should provide a preview of what the content will be on the front end. But, when the content is actually _empty_ it should let users preview it's empty.

At this point, shall an empty paragraph have some placeholder text? Not sure that would be ideal. It makes sense to show that it is _empty_. I'd argue the same should happen for all blocks.

On top of that, a visual preview that shows a few non-existing tags like in the screenshot below:

![CleanShot 2024-07-17 at 16 41 15](https://github.com/user-attachments/assets/2007c514-ac27-4617-95d3-cd16549d3a3d)

would not communicate the block is 'empty' in an accessible way. From an usability perspective, users could be confused and attempt to click the tags.

Rather, I'd consider to improve consistency of block when they are 'empty'. For example:

- Add an Image block and just leave the placeholder in place. Publish.
- The editor will show the placeholder.
- The front end won't render any markup related to the image.
- Add a Tag cloud set to a taxonomy with no terms created yet.
- The editor will show the text `There’s no content to show here yet.`
- The front end will show the text `There’s no content to show here yet.` as well.

This behavior is inconsistent and, in my personal opinion, the front end shouldn't show anything.

That text should likely be used only as a placeholder in the editor. The wording could be improved a bit, e.g.:
`There are no terms to show in the Tag cloud yet.`

[...]

PR #63774, opened 2024-07-21
Tag Cloud: Improve state of block with no tags

akasunil:
## What?

Part of https://github.com/WordPress/gutenberg/issues/63684

## Why?

Tag cloud block render message in frontend when there are no tags available to display.

## How?

Render nothing when no tags are available to display.

## Testing Instructions

- Add tag could block in post
- Select taxonomy with no terms, Save the post.
- Go to front-end, and make sure tag cloud block isn't rendered.


packages/block-library/src/tag-cloud/index.php
```diff
@@ -30,7 +30,7 @@ function render_block_core_tag_cloud( $attributes ) {
 	$tag_cloud = wp_tag_cloud( $args );
 
 	if ( ! $tag_cloud ) {
-		$tag_cloud = __( 'There&#8217;s no content to show here yet.' );
+		return '';
```
Mamaduka: This is the server-side rendered block. Returning an empty string will also render nothing in the editor. The linked issue only states that it should render nothing on the front-end.

packages/block-library/src/tag-cloud/index.php
```diff
@@ -29,8 +29,13 @@ function render_block_core_tag_cloud( $attributes ) {
 	);
 	$tag_cloud = wp_tag_cloud( $args );
 
-	if ( ! $tag_cloud ) {
```
richtabor, line 32: What's the reason for chaining `!` to `empty`?

akasunil: Makes code more readable in my opinion. Do you need me to revert it to avoid unnecessary changes?

packages/block-library/src/tag-cloud/index.php
```diff
@@ -29,8 +29,13 @@ function render_block_core_tag_cloud( $attributes ) {
...
-		$tag_cloud = __( 'There&#8217;s no content to show here yet.' );
+	if ( empty( $tag_cloud ) ) {
+		// Display placeholder content when there are no tags only in editor.
+		if ( ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || is_admin() ) {
+			$tag_cloud = __( 'There&#8217;s no content to show here yet.' );
+		} else {
```
Mamaduka: WP introduced the `wp_is_serving_rest_request` method, which we could use here. The `is_admin` check shouldn't be required.

akasunil: Thank your for a review. I have addressed your feedback.

Mamaduka (approved):
Works as expected. The placeholder is displayed in the editor, and nothing is rendered on the front-end.

PR #63776, opened 2024-07-21
Time To Read: Add border block support 

akasunil:
## What?

Add border block support to the `Time To Read` block.

## Why?
`Time To Read` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Time To Read` block's border is configurable via Global Styles
- Edit template/page, Add `Time To Read`  block and Apply the border styles

## Screenshots or screencast 
 
[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (2).webm](https://github.com/user-attachments/assets/fd1ff4ed-e58e-46dc-8dfb-f6703d00d6aa)

### DIFF
--- a/packages/block-library/src/post-time-to-read/block.json
+++ b/packages/block-library/src/post-time-to-read/block.json
@@ -50,13 +50,7 @@
 			"radius": true,
 			"color": true,
 			"width": true,
-			"style": true,
-			"__experimentalDefaultControls": {
-				"radius": true,
-				"color": true,
-				"width": true,
-				"style": true
-			}
+			"style": true
 		}
 	}
 }
--- a/packages/block-library/src/post-time-to-read/block.json
+++ b/packages/block-library/src/post-time-to-read/block.json
@@ -45,6 +45,18 @@
 		},
 		"interactivity": {
 			"clientNavigation": true
+		},
+		"__experimentalBorder": {
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"radius": true,
+				"color": true,
+				"width": true,
+				"style": true
+			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


packages/block-library/src/post-time-to-read/block.json
```diff
@@ -45,6 +45,18 @@
...
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"radius": true,
+				"color": true,
+				"width": true,
+				"style": true
+			}
```
aaronrobertshaw: There was [some feedback](https://github.com/WordPress/gutenberg/pull/63543#issuecomment-2232971361) when adding border support to other blocks that didn't display the dimension controls by default that the border controls should be optional. Perhaps we should follow that advice here as well?

aaronrobertshaw (commented):
Thanks for putting this together @akasunil 👍 

It is testing well for me but before we push ahead with this, I'd like to seek some extra opinions on whether the border controls should be shown by default. Particularly, given the dimension controls here are also hidden by default.

@jameskoster and @richtabor should we make these border controls optional similar to the suggestion here: https://github.com/WordPress/gutenberg/pull/63543#issuecomment-2232971361?

jameskoster:
@aaronrobertshaw yes I think that's a good suggestion; available but hidden by default.

packages/block-library/src/post-time-to-read/block.json
akasunil: I have updated json

aaronrobertshaw (approved):
This is looking good @akasunil 👍 

🚢 


https://github.com/user-attachments/assets/43bd44b6-9ce8-44b2-8e71-ec9b55bd1fff

PR #63777, opened 2024-07-21
Video: Add border block support

akasunil:
## What?

Add border block support to the `Video` block.

## Why?
`Video` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Video` block's border is configurable via Global Styles
- Edit template/page, Add `Video`  block and Apply the border styles

## Screenshots or screencast 
 
 
[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress-video-block.webm](https://github.com/user-attachments/assets/98395342-39ff-43d7-9580-26025ca77b81)


aaronrobertshaw (commented):
I appreciate all the effort into publishing PRs adopting more block supports @akasunil 💪 

This one works well until we include a caption for the video. At that point the border wraps around the outer `figure` element which has the caption as a child.

Screenshot 2024-07-23 at 3 50 19 PM

This is inconsistent with the image block which will target the inner `img` element for the border rather than the wrapping `figure`. I believe we'll need to do the same here for the Video block.

To that end, the Image [block.json](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/block.json?plain=1#L112-L133) file should help illustrate how that can be done. The Image block is a dynamic block though so a static block like the Form Input block will help show how to [apply the border support classes and styles](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/form-input/save.js?plain=1#L41) within the Video block's save function.

Essentially, the Video block here needs to:
- skip serializing the border block support classes and styles on the block's wrapper by using `"__experimentalSkipSerialization": true` in its supports config
- define the `selectors.border` CSS selector needed for global styles to target the inner video element
- manually apply the border support classes and styles in both the edit component and save function

It shouldn't be as painful as that sounds but let me know how you go. I'm happy to answer any questions you have, review as often as required, or help give the code a nudge.

Lastly, for anyone else coming to review this PR, the adoption of border radius is consistent with other blocks given the current padding support. The general consensus is that we should lean into user freedom rather than restrict them here due to the possibility that the inner video won't be clipped when the border radius is too high.

Screenshot 2024-07-23 at 3 47 58 PM


Thanks again for the work on this one 🙇

akasunil:
It should work fine now. 


[video-block-border-styles.webm](https://github.com/user-attachments/assets/f03e3ce7-70ff-4f80-a196-d6d49676100c)

packages/block-library/src/video/save.js
```diff
@@ -37,6 +41,10 @@ export default function save( { attributes } ) {
 					preload={ preload !== 'metadata' ? preload : undefined }
 					src={ src }
 					playsInline={ playsInline }
+					className={ borderProps.className }
+					style={ {
+						...borderProps.style,
+					} }
```
aaronrobertshaw:
```suggestion
					style={ borderProps.style }
```

With only border styles being added here they don't need to be deconstructed.

aaronrobertshaw (commented):
Thanks for the continued iterations on this one @akasunil 👍 

With the border styles now applied directly to the video I think this is working more as expected. There is one catch though comparing the Video and Image blocks.

With the border radius applied to an image, there's no real downside. For the Video block though, it risks clipping the playback controls. With the block's padding being applied to the Video block's wrapper and not the video element there's not much control to mitigate this. 

Screenshot 2024-08-12 at 2 05 03 PM


I'd like to get some advice from our design gurus on whether it is ok to proceed with the current approach. A compromise could be to not support border radius.

Personally, I'm leaning towards keeping the border radius support and allowing users to use their discretion on how much is too much when it comes to border radius.

packages/block-library/src/video/editor.scss
```diff
@@ -1,29 +1,6 @@
 // Provide special styling for the placeholder.
 // @todo this particular minimal style of placeholder could be componentized further.
 .wp-block-video.wp-block-video {
-	// Show Placeholder style on-select.
```
akasunil: This css isn't required as we are displaying illustration now on block selection only.

packages/block-library/src/video/edit.js
```diff
@@ -157,6 +142,27 @@ function VideoEdit( {
...
+					[ borderProps.className ]:
+						!! borderProps.className && ! isSingleSelected,
+				} ) }
+				withIllustration={ ! isSingleSelected }
+				icon={ icon }
+				label={ __( 'Video' ) }
+				instructions={ __(
+					'Upload a video file, pick one from your media library, or add one with a URL.'
+				) }
+				style={ borderProps.style }
```
aaronrobertshaw: If the border classes are only being applied when `! isSingleSelected` should the inline styles also only be applied under the same conditions?

packages/block-library/src/video/editor.scss
aaronrobertshaw: Can you expand on why this is changing?

akasunil: Looks like it already been handled by https://github.com/WordPress/gutenberg/pull/64861. I'll sync the PR with trunk.

packages/block-library/src/video/edit.js
```diff
@@ -136,13 +138,17 @@ function VideoEdit( {
...
+					[ borderProps.className ]:
+						!! borderProps.className && ! isSingleSelected,
+				} ) }
 				withIllustration={ ! isSingleSelected }
 				icon={ icon }
 				label={ __( 'Video' ) }
 				instructions={ __(
 					'Upload a video file, pick one from your media library, or add one with a URL.'
 				) }
+				style={ ! isSingleSelected ? { ...borderProps.style } : {} }
```
aaronrobertshaw:
```suggestion
				style={ ! isSingleSelected ? borderProps.style : {} }
```

It looks like the recent updates [reintroduced](https://github.com/WordPress/gutenberg/pull/63777#discussion_r1713251390) the unnecessary destructuring of the border styles object.

akasunil: Sorry, missed to remove it.

hanneslsm:
> Personally, I'm leaning towards keeping the border radius support and allowing users to use their discretion on how much is too much when it comes to border radius.

I agree, especially because it's possible to hide the Playback controls already now anyway.

akasunil:
Thank you @aaronrobertshaw and @hanneslsm 

If there are no more feedbacks, can we ship this PR? I have re-based it with latest trunk.
Thank you

aaronrobertshaw:
> If there are no more feedbacks, can we ship this PR? I have re-based it with latest trunk.

With the 6.7 release in its final stages, we should hold off until after the release so our design gurus can weigh in and confirm the general thinking. As this is an enhancement and not a critical bug fix we can avoid to take our time with it.

PR #63778, opened 2024-07-21
Site Tagline: Add border block support 

akasunil:
## What?

Add border block support to the `Site Tagline` block.

## Why?
`Site Tagline` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Site Tagline` block's border is configurable via Global Styles
- Edit template/page, Add `Site Tagline`  block and Apply the border styles

## Screenshots or screencast 
 
  
[site-tagline-block-border-support-video.webm](https://github.com/user-attachments/assets/69b7da09-a15a-4543-a063-fe082abce59a)

### DIFF
--- a/packages/block-library/src/site-tagline/block.json
+++ b/packages/block-library/src/site-tagline/block.json
@@ -70,5 +70,6 @@
 			}
 		}
 	},
-	"editorStyle": "wp-block-site-tagline-editor"
+	"editorStyle": "wp-block-site-tagline-editor",
+	"style": "wp-block-site-tagline"
 }
--- a/packages/block-library/src/site-tagline/block.json
+++ b/packages/block-library/src/site-tagline/block.json
@@ -61,13 +61,7 @@
 			"radius": true,
 			"color": true,
 			"width": true,
-			"style": true,
-			"__experimentalDefaultControls": {
-				"radius": true,
-				"color": true,
-				"width": true,
-				"style": true
-			}
+			"style": true
 		}
 	},
 	"editorStyle": "wp-block-site-tagline-editor",
--- a/packages/block-library/src/site-tagline/block.json
+++ b/packages/block-library/src/site-tagline/block.json
@@ -56,6 +56,18 @@
 		},
 		"interactivity": {
 			"clientNavigation": true
+		},
+		"__experimentalBorder": {
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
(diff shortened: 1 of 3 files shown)


aaronrobertshaw (commented):
Thank you for the PR @akasunil 👍 

As we discussed on the other border support PRs recently. Blocks adopting the border (or padding) block support need to have a `box-sizing: border-box` style so they will honour any width imposed by a parent block.

The screenshot below shows a Group block setting the width of its content to `500px` and inspecting the Site Tagline block inside it. You'll see that the border I applied causes the block to exceed the dimensions it should have.

Screenshot 2024-07-23 at 3 35 09 PM

akasunil:
It should be fine now.

In editor:
image

In frontend:
image

packages/block-library/src/site-tagline/block.json
```diff
@@ -56,7 +56,20 @@
...
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"radius": true,
+				"color": true,
+				"width": true,
+				"style": true
+			}
```
aaronrobertshaw:
```suggestion
			"style": true
```

Given the dimension controls are optional here, the border controls probably should be too.

aaronrobertshaw (commented):
Thanks for adding the `box-sizing` @akasunil 👍 

There's been some feedback on the default display of controls in the recent round of block support adoption PRs. For niche use cases such as this one, it's been suggested to make the new controls optional.

I guess the best clue is if the dimension and spacing controls are optional, then the border controls should be as well.

I've left a suggestion in an inline comment that you should be able to commit if you agree. At that point, I think this is ready.

akasunil:
Thank you @aaronrobertshaw 

> I guess the best clue is if the dimension and spacing controls are optional, then the border controls should be as well.

ill follow that

packages/block-library/src/site-tagline/style.scss
```diff
@@ -0,0 +1,4 @@
+:root :where(.wp-block-site-tagline) {
```
aaronrobertshaw:
```suggestion
.wp-block-site-tagline {
```

The wrapping `:root :where()` here is unnecessary as both this and a simple class result in `0-1-0` specificity.

aaronrobertshaw (approved):
Thanks for your patience iterating on this PR 👍 

https://github.com/user-attachments/assets/0f68f997-74c0-49a2-8b1c-12004f607e25

PR #63779, opened 2024-07-21
Paste Handler: Try to fix pasting text with formatting

### LINKED ISSUE #63434: Adding a link url via copy/paste doesn't work on previously formatted text
### Description

For some of my content I add a link to text and then bold it. Sometimes, I bold it first and then add the link. 
For a while now I noticed that copy/page over a highlight text didn't always work, and I finally chased down the source of this "misbehavior" 

If the text is formatted in bold, highlighting and adding the link via copy/paste doesn't work anymore. It does\work when using the link button, but that makes the process two clicks longer. 

See below video for illustration. 

- Add a paragraph block with text. 
- Highlight a portion of the text 
- Format it in bold
- Highlight the text again
- copy/paste a URL. (use https://google.com or any other url) 

https://github.com/WordPress/gutenberg/assets/39980/eda3994c-2856-436d-9dd4-0d595f5b86f0

### Environment info

- WordPress 6.6 RC3 (but it has been there for a while so it's not a WordPress 6.6 issue)
- Gutenberg 18.7.1
- Google Chrome Version 126.0.6478.127

Yes

Yes

t-hamano:
Fixes #63434

## What?

This PR fixes an issue where copying and pasting text that contains inline formatting would strip the formatting from the pasted text.

## Why?

I'm not sure what the root cause is, but it seems the problem occurs when the copied text contains formatted text at the end.

Also, when pasting multiple times, the formatted text seems to be pasted correctly either the odd or even number of times.

## How?
Removed the early return that was causing the problem.

We'll review what tests fail to ensure this doesn't introduce any unintended regressions.

## Testing Instructions

- Insert a Paragraph block and type your text.
- Apply inline formatting to the text.
- Select the text and copy it. Make sure to include the inline formatting at the end of the selection.
- Paste it into another paragraph.
- The formatting should be preserved.

### Before

https://github.com/user-attachments/assets/bc79467c-e385-457b-bd72-6c6ff78e2900

### After

https://github.com/user-attachments/assets/f8c1f06b-5ae9-450e-8c79-1c67090ffd8d


t-hamano:
@ellatrix It appears to pass all tests. Could removing this early return cause any problems?

packages/block-editor/src/components/rich-text/event-listeners/paste-handler.js
```diff
@@ -27,12 +27,6 @@ export default ( props ) => ( element ) => {
 			pastePlainText,
 		} = props.current;
 
-		// The event listener is attached to the window, so we need to check if
-		// the target is the element.
```
ellatrix, line 31: I think this is a valid comment. I probably thought the target would be a focusable element and always the wrapper, but turns out not. Maybe change it to:

```js
! element.contains( event.target )
```

To make sure we only catch paste within the container.

ellatrix: Btw, there's a similar early return in `enter.js`. I wonder if that's causing any bugs?

t-hamano:
```
! element.contains( event.target )
```

I have confirmed that this approach also works correctly. After a bit of debugging, it appears that `element.target` is sometimes an inner element, rather than the focusable element itself.

> Btw, there's a similar early return in `enter.js`. I wonder if that's causing any bugs?

I've tried debugging close to [this code](https://github.com/WordPress/gutenberg/blob/645c0e7d4156a0b7275d09791f656f65ac183efe/packages/block-editor/src/components/rich-text/event-listeners/enter.js#L25-L29), and as far as I've tested, `element.target` is always the focusable element itself, so it doesn't seem to be causing any kind of bug.

ellatrix: Yes, that's super odd.

ellatrix:
Should we add a small e2e test so this doesn't regress in the next refactor?

t-hamano:
I've added an e2e test.

This test is expected to fail on trunk, but for some reason it passes when I run the e2e tests. However, if I run this test manually on trunk, it fails as expected.

I don't understand why it passes when I run it as an e2e test 🤔

### Debugging e2e test on trunk

The test passes, but this is unexpected.

Post Content:

```html
<!-- wp:paragraph -->
<p><strong>testtest</strong></p>
<!-- /wp:paragraph -->
```


https://github.com/user-attachments/assets/c1519b9f-71f3-45bb-976d-4f4c9938df1e

### Manual test on trunk

The test fails, which is expected.

Post Content:

```html
<!-- wp:paragraph -->
<p><strong>test</strong>test</p>
<!-- /wp:paragraph -->
```

https://github.com/user-attachments/assets/76ecbdae-0f54-45ed-bb07-0ebce43f656c

test/e2e/specs/editor/various/copy-cut-paste.spec.js
```diff
@@ -478,6 +478,23 @@ test.describe( 'Copy/cut/paste', () => {
...
+		editor,
+	} ) => {
+		await editor.canvas
+			.locator( 'role=button[name="Add default block"i]' )
+			.click();
+		await page.keyboard.type( 'test' );
+		await pageUtils.pressKeys( 'primary+a' );
+		await pageUtils.pressKeys( 'primary+b' );
+		await pageUtils.pressKeys( 'primary+c' );
+		await page.keyboard.press( 'ArrowRight' );
```
ellatrix: I think you'll need to make sure the caret is _within_ the bold format?

t-hamano: Strangely, the problem does not occur _inside_ formatted text, but only occurs when you move the caret _outside_ the formatted text and paste:

https://github.com/user-attachments/assets/9891efaf-cb5c-438c-a3fb-46a2c3f86a0f

ellatrix: To trigger the link behaviour, you'd need to _select_ text within the format?

ellatrix: I'd for example select a letter within the text format (to make sure the test is stable and selection boundaries are within the format), then paste the link on the uncollapsed selection

t-hamano: I tried copying letters within the formatted text and the e2e test still passed on trunk 😅 Of course, the test is expected to fail on trunk.

https://github.com/user-attachments/assets/fb9bc6b3-ceca-4724-8a4c-bf05c80bfd92

ellatrix: But you're not pasting a link?

t-hamano: Ah, I see what you mean.

#63434 reports two issues:

1. If the text is formatted, highlighting and adding the link via paste doesn't work
2. If the copied text contains formatted text at the end, pasting the copied text outside the formatted text removes the formatting

This PR solves both problems, but the E2E test I added is for the second problem. Does this mean I should write a test for the first problem?

ellatrix: I didn't see the second problem, I only knew about the first problem from the PR description. Sure, if it's easier to write a PR for.

ellatrix: There may be slight differences with how paste is emulated that make it pass.

t-hamano: I updated the e2e test to paste the link into the formatted text. But for some reason, this test still passes on trunk as well 😅

t-hamano:
@ellatrix 

The problem this PR is trying to solve seems to have been fixed as a result of #63671, which allows multiple selections on touch devices.

Since #63671 is an enhancement, I don't think it can be backported to WP6.6.2 as is. If we want to backport just the paste fix to WP6.6.2, do we need to manually backport just [this code](https://github.com/WordPress/gutenberg/blob/8c4be5842025e3e4a707194323258adbdcddeaa4/packages/block-editor/src/components/rich-text/event-listeners/paste-handler.js#L29-L39) to the WP6.6 branch?

Pinging @hellofromtonya @vcanales for visibility

hellofromtonya:
> The problem this PR is trying to solve seems to have been fixed as a result of #63671, which allows multiple selections on touch devices.
> 
> Since #63671 is an enhancement, I don't think it can be backported to WP6.6.2 as is. If we want to backport just the paste fix to WP6.6.2, do we need to manually backport just [this code](https://github.com/WordPress/gutenberg/blob/8c4be5842025e3e4a707194323258adbdcddeaa4/packages/block-editor/src/components/rich-text/event-listeners/paste-handler.js#L29-L39) to the WP6.6 branch?
Thanks for ping and heads up.

Also noting that #63671 is from 2019. As it's before 6.6 and is an enhancement, I agree with you @t-hamano that it cannot be included in the 6.6 minor cycle.

That said, if there's a _specific_ bug introduced during the 6.6 cycle of which a smaller subset of this PR _directly_ fixes the bug, then it could potentially and cautiously be considered. Why caution? Sometimes an enhancement indirectly fixes bugs. Though it may fix bugs, it's still introducing new functionality or feature.

t-hamano:
Sorry for the late reply.

Just to be sure we're on the same page, I'd like to provide a brief summary.

- In [#63434](https://github.com/WordPress/gutenberg/issues/63434), an issue with pasting was reported. This was a regression that first appeared in WP6.6.
- This PR was submitted to fix that regression. Then, this PR was labeled as "Backport to WP Minor Release".
- While this PR was in progress, [#63671](https://github.com/WordPress/gutenberg/issues/63671) was merged. The purpose of [#63671](https://github.com/WordPress/gutenberg/issues/63671) was an enhancement, but as a result it fixed the issue reported in [#63434](https://github.com/WordPress/gutenberg/issues/63434).
- This PR conflicts with [#63671](https://github.com/WordPress/gutenberg/issues/63671), so it cannot be merged into Gutenberg's trunk branch.
- [#63671](https://github.com/WordPress/gutenberg/issues/63671) cannot be backported to WP6.6.2 directly because its purpose was an enhancement.
- If the regression is considered significant, we may need to manually backport the regression-only fix, i.e. the equivalent code change made in this PR, to the WP6.6 branch.

If you have any questions, please let me know, and I'd be happy to help with any backports needed.

Either way, this PR is no longer needed for the trunk branch so I will close it.

test/e2e/specs/editor/various/copy-cut-paste.spec.js
```diff
@@ -565,6 +565,30 @@ test.describe( 'Copy/cut/paste', () => {
...
 	} );
 
+	test( 'should paste link to formatted text', async ( {
+		page,
+		pageUtils,
+		editor,
+	} ) => {
+		await editor.insertBlock( {
+			name: 'core/paragraph',
+			attributes: { content: 'test' },
```
ellatrix: Perhaps it could start with bolded text to reduce the complexity of the test?

ellatrix (approved):

test/e2e/specs/editor/various/copy-cut-paste.spec.js
t-hamano: Indeed, I simplified it with 9a56955

t-hamano:
This PR has a bit of a complicated history, so here's a summary of why a backport to WP 6.7 is needed:

- In [#63434](https://github.com/WordPress/gutenberg/issues/63434), an issue with pasting was reported. This PR was submitted to fix this problem.
- While this PR was in progress, [#63671](https://github.com/WordPress/gutenberg/issues/63671) was merged. The purpose of [#63671](https://github.com/WordPress/gutenberg/issues/63671) was an enhancement, but as a result it fixed the issue reported in [#63434](https://github.com/WordPress/gutenberg/issues/63434).
- [#63671](https://github.com/WordPress/gutenberg/issues/63671) caused some issues, so it was reverted by [#65414](https://github.com/WordPress/gutenberg/pull/65414) and backported to WP 6.7. This means that the problem this PR is trying to solve has reappeared.
- So this PR ends up resolving [#63434](https://github.com/WordPress/gutenberg/issues/63434) and should be backported to WP 6.7.

Issue #43243, opened 2022-08-16
Tracking: Dimensions Design Tools Consistency


aaronrobertshaw:
## Overview

This issue details the current state of dimensions and spacing block support or design tool adoption across all blocks and tasks required to fill any gaps. Overall design tool consistency efforts are being tracked via the parent issue: https://github.com/WordPress/gutenberg/issues/43241.

## Guidelines for Adopting Spacing Supports

- We're generally happy to have padding & margin support for all blocks. (Exceptions possible)
- We're leaning towards simple opt-in for margin support i.e. `true` rather than `["top", "bottom"]`
- Neither padding nor margin should be shown by default i.e. they're accessed via the ellipsis menu. (Exceptions possible)
- All blocks (except special cases like Heading & Paragraph) should have `box-sizing: border-box`. This will be reviewed on a case-by-case basis. (Exceptions likely).

## Legend

| Value         | Description                                                                  |
| ------------- | ---------------------------------------------------------------------------- |
| ✅            | Feature has been adopted and is displayed as a **default** control |
| ✅ (Optional) | Feature has been adopted but is an **optional** control            |
| ⚠             | There is a **bug or issue** with this block support feature's adoption           |
| ❌            | Feature has been explicitly **opted out** of           |
| <PR#>         | Links to **PR** adopting the feature for this block                |
| -             | Feature has **not explicitly** been adopted/omitted                              |
| 🛠             | Implemented via an **ad hoc** / bespoke control                                  |
| 🚧 | Work is in progress towards adopting this feature (no PR yet) |

## Block Support Adoption

_Note: Deprecated blocks have been omitted from this table. e.g. Comment Author Avatar, Post Comment & Text Columns._

| Block                        | Padding       | Margin        | Block Gap     | Height | Width | Min Height | Aspect Ratio |
| ---------------------------- | ------------- | ------------- | ------------- | ------ | ----- | ---------- |---------- |
| Archives                     | ✅ (Optional)  | ✅ (Optional)  | -             | -      | -     | -          | -     |
| Audio                        | ✅ (Optional) | ✅  (Optional) | -             | -      | -     | -          | -     |
| Avatar                       | ✅ (Optional) | ✅ (Optional) | -             | 🛠      | 🛠     | -          | -     |
| Button                       | ✅ (Horizontal, Vertical)            | -             | -             | -      | 🛠     | -          | -     |
[...]

PR #63782, opened 2024-07-21
Social Link: Add border and spacing support

akasunil:
## What?

Add border block support to the `Social Links` block.

Part of,
https://github.com/WordPress/gutenberg/issues/43247
https://github.com/WordPress/gutenberg/issues/43243

## Why?
`Social Link` block is missing border and spacing support.

## How?

## Testing Instructions

- Make sure that `Social Link` styles are configurable via Global Styles
- Edit template/page, Add `Social Link` block and Apply the styles
- Verify that block styles display correctly in both the editor and frontend


akasunil:
Hi @aaronrobertshaw 
Do we need spacing styles support for `social icon` block, I have added spacing support and currently it getting override by `Social Icons` block css from [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/social-links/style.scss#L94-L95)

aaronrobertshaw:
Thanks for the ping 👍 

> Do we need spacing styles support for social icon block, I have added spacing support and currently it getting override by Social Icons block css from [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/social-links/style.scss#L94-L95)

It might be handy for some users but my guess is it will be very niche. If it were to be supported, those styles would need to be tweaked to accommodate the supports.

My bigger concern is the adoption of standard color supports. From memory the parent Social Links block has an ad hoc color implementation that applies background and foreground colors to its inner icons. The proposed support here would likely conflict with that.

akasunil:
> It might be handy for some users but my guess is it will be very niche. If it were to be supported, those styles would need to be tweaked to accommodate the supports.

Removed those styles and spacing support works fine now for `social link` block even with `Block spacing` from `social links` block. 



> From memory the parent Social Links block has an ad hoc color implementation that applies background and foreground colors to its inner icons. The proposed support here would likely conflict with that.

Indeed, It does conflict with ad hoc color implementation. We have to remove the ad hoc implementation for color support in `social links` block and might require to create deprecation too. what do you propose? should we skip color support in `social link` block or remove ad hoc color support from `social links` block ?

packages/block-library/src/social-links/style.scss
```diff
@@ -91,8 +91,6 @@
 // This needs specificity because themes usually override it with things like .widget-area a.
 .wp-block-social-links .wp-block-social-link.wp-social-link {
 	display: inline-block;
-	margin: 0;
-	padding: 0;
```
aaronrobertshaw, line 98: What's the thinking behind removing these resets?

My understanding was they should be tweaked to lower the specificity so global styles could be applied. Was there an unexpected issue when doing so?

aaronrobertshaw:
> should we skip color support in social link block or remove ad hoc color support from social links block ?

I believe the reasoning behind the ad hoc color implementation for the parent Social Links block was that the desired UX was to make a single color selection and have that applied to all the inner social links. 

Perhaps, until there is a good reason to change that, color support should be omitted from the individual social link blocks. If there is significant demand from end users it can always be revisited.

packages/block-library/src/social-links/style.scss
akasunil: Should we move these style to social link block? Global style would be able to override that style. Another solution is we can add selector in social link block for spacing styles.

packages/block-library/src/social-link/block.json
```diff
@@ -34,6 +34,27 @@
...
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"radius": true,
+				"color": true,
+				"width": true,
+				"style": true
+			}
```
aaronrobertshaw: If the dimension controls aren't displayed by default, the border controls shouldn't be either.

packages/block-library/src/social-link/block.json
```diff
@@ -34,6 +34,27 @@
 		"html": false,
 		"interactivity": {
 			"clientNavigation": true
+		},
+		"spacing": {
+			"margin": true,
```
aaronrobertshaw, line 39: I'm not sure how useful allowing margin support on individual links is. My understanding of the social links block is that spacing should be handled via the parent Social Links block's block spacing support.

packages/block-library/src/social-link/block.json
```diff
@@ -34,6 +34,27 @@
...
+			"padding": true,
+			"units": [ "px", "em", "rem", "vh", "vw" ],
```
aaronrobertshaw: Do we need custom definitions of available spacing units here?

packages/block-library/src/social-links/style.scss
aaronrobertshaw: I think this might need some further thought. 

The Social Link block contains a clickable button or link depending on whether the user is in the editor or frontend. 

It would make sense to me that any padding added for a link would be on that inner button or link so it is taken into account for the clickable area. The same could be said for the border too.

This would likely mean, leaving the styles here untouched and skipping serialization for the border and padding support. Then manually applying the support styles on the appropriate inner element.

aaronrobertshaw (commented):
Thanks for continuing this one @akasunil 👍 

There are a few aspects that need tweaking here or some further thought and discussion. I've left inline comments for these but it might pay to get design feedback as well to guide the final decisions, for example, whether margin support makes sense here.

P.S. To help others coming to this PR, it would be handy to update the PR description to reflect the current state of the PR 🙏

packages/block-library/src/social-link/block.json
akasunil: I guess not.

aaronrobertshaw:
Hi @akasunil, thanks for your patience on this one.

It looks like there are still a few things that need to be changed in this PR before it can proceed. For example:

1. Removing margin block support so the parent Social Links block's gap support can control spacing.
2. Applying padding support to the inner button/link element so it is within the element's clickable area etc.

Unfortunately, I'll have limited bandwidth for the coming months so will not be able to help dig into issues or review future iterations here. I'm sure other contributors will be able to help review though as needed 🙏

Issue #63780, opened 2024-07-21
Discussion in the side menu did not receive translation


maisondasilva:
### Description

Discussion in the side menu did not receive translation, I noticed this in my pt-BR language, I believe other languages ​​may be affected

### Step-by-step reproduction instructions

![image](https://github.com/user-attachments/assets/777801ce-c715-439d-ab7b-d71ecd47ee6f)

### Environment info

WordPress 6.6

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

- [X] Yes

talldan:
The strings are all internationalized, and it's translated in other languages (I checked French), so it seems like there's missing translations:
https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/components/post-comments/index.js

The one issue in the internationalization I can see is that there's no distinction between the verb (e.g. 'to open') and the adjective (e.g. 'is open'), so translators might get that wrong, which I can follow up and fix.

In terms of the missing translations, this wouldn't be the place to report that kind of issue. I don't think there is a place to report it, I believe the untranslated strings will be on a list awaiting translation and I can see others that are untranslated for pt_BR in the UI, so the translation doesn't seem to be 100%. You could reach out to the [polyglots team in slack](https://wordpress.slack.com/archives/C02RP50LK). There's some information about the Portuguese Brazilian team here - https://make.wordpress.org/polyglots/teams/?locale=pt_BR.

PR #63791, opened 2024-07-22
Discussions panel: Distinguish between verb and adjective form of open for internationalization

talldan:
## What?
Fixes an issue spotted in replying to #63780

The word open can be used as an verb "Open a page" or adjective "The page is open", and some of the i18n in the codebase doesn't distinguish.

## Why?
I imagine the words could be translated differently depending on this context (I'm not multilingual, but I think in French it could be "Ouvrir" or "Ouvert").

## How?
Uses the `_x` function to add a context string for the translation

## Testing Instructions
Create a post and in the document sidebar, check that the 'Discussion' options still show the English text 'Open'.

## Screenshots or screencast 
Screenshot 2024-07-22 at 10 52 18 AM

### DIFF
--- a/packages/editor/src/components/post-comments/index.js
+++ b/packages/editor/src/components/post-comments/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, _x } from '@wordpress/i18n';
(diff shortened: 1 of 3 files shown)


ntsekouras (approved):
Good catch, thank you!

PR #63792, opened 2024-07-22
Sync backport changelog action: Use outputs instead of env

noisysocks:
The Sync Issue step in the Sync Core Backport Issue workflow wasn't working.

In https://github.com/WordPress/gutenberg/commit/8bb22a2b43fc439fcdb67b2151ffb80fb4a3ba02 this step was made conditional so that it would only run if there are changes within the `backport-changelog` directory.

But, for example, it didn't run here even though there were changes in the `backport-changelog` directory:

https://github.com/WordPress/gutenberg/actions/runs/9950072850/job/27487350407

To fix, I've updated the step's `if` to use `steps.$id.outputs.$var` instead of `env.$var`. This is [what's in the GitHub Actions docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter).

I tested this in a fork of Gutenberg. You can see the step correctly being skipped when there are no changes here:

https://github.com/noisysocks/gutenberg/actions/runs/10034541947/job/27729221457

And you can see the step correctly running when there are changes here:

https://github.com/noisysocks/gutenberg/actions/runs/10034606864/job/27729375022

(The step then fails for some reason, but we only care that it isn't skipped.)


noisysocks:
> (The step then fails for some reason, but we only care that it isn't skipped.)

Looks like this is because `issue.body` can be null if issue has no description. Added a fix in 4350f0fc218. Here's it working:

https://github.com/noisysocks/gutenberg/actions/runs/10034745885/job/27729718762

talldan (approved):
LGTM

ellatrix:
Why does it not work here. Env should work just fine, like here:  

https://github.com/WordPress/gutenberg/blob/d0b82dacfbb19ca23346bd56137ebd1ea4c9f923/.github/workflows/cherry-pick-wp-release.yml#L99-L105

noisysocks:
Hmm. The example you linked to uses `$GITHUB_ENV` whereas here we're using `$GITHUB_OUTPUT`. Maybe it's just that.

PR #63797, opened 2024-07-22
Block Bindings: Allow bindings bootstrap after registration

SantosGuillamot:
## What?
As reported [here](https://github.com/WordPress/gutenberg/pull/63470#discussion_r1686181167) by @gziolo , when plugins start using the editor APIs, there could be a use case where sources are registered in the client before they are bootstrapped from the server. This pull request ensures the properties are not overriden.

## Why?
It could break the developer experience once these APIs are public.

## How?
Preserving the existing properties in the reducer.

## Testing Instructions
I added a unit test to cover this use case. The other registration tests should pass as well.

### DIFF
--- a/packages/blocks/src/api/test/registration.js
+++ b/packages/blocks/src/api/test/registration.js
@@ -1706,7 +1706,7 @@ describe( 'blocks', () => {
 
 		it( 'should correctly merge properties when bootstrap happens after registration', () => {
 			// Register source in the client.
-			const clientProperties = {
+			const clientOnlyProperties = {
 				getValues: () => 'values',
 				setValues: () => 'new values',
 				getPlaceholder: () => 'placeholder',
@@ -1715,7 +1715,8 @@ describe( 'blocks', () => {
 			registerBlockBindingsSource( {
 				name: 'core/custom-source',
 				label: 'Client Label',
-				...clientProperties,
+				usesContext: [ 'postId', 'postType' ],
+				...clientOnlyProperties,
 			} );
 
 			// Bootstrap source from the server.
@@ -1724,14 +1725,17 @@ describe( 'blocks', () => {
 			).addBootstrappedBlockBindingsSource( {
 				name: 'core/custom-source',
 				label: 'Server Label',
-				usesContext: [ 'postId' ],
+				usesContext: [ 'postId', 'serverContext' ],
 			} );
 
 			// Check that the bootstrap values prevail and the client properties are still there.
 			expect( getBlockBindingsSource( 'core/custom-source' ) ).toEqual( {
+				// Should use the server label.
 				label: 'Server Label',
-				usesContext: [ 'postId' ],
-				...clientProperties,
+				// Should merge usesContext from server and client.
+				usesContext: [ 'postId', 'postType', 'serverContext' ],
+				// Should keep client properties.
+				...clientOnlyProperties,
 			} );
 
 			unregisterBlockBindingsSource( 'core/custom-source' );
--- a/packages/blocks/src/api/test/registration.js
+++ b/packages/blocks/src/api/test/registration.js
@@ -1704,7 +1704,7 @@ describe( 'blocks', () => {
 			);
 		} );
 
-		it( 'should not be overriden if bootstrap happens after registration', () => {
+		it( 'should correctly merge properties when bootstrap happens after registration', () => {
 			// Register source in the client.
 			const clientProperties = {
 				getValues: () => 'values',
--- a/packages/blocks/src/api/test/registration.js
+++ b/packages/blocks/src/api/test/registration.js
@@ -1703,6 +1703,39 @@ describe( 'blocks', () => {
 				'Block bindings source "core/test-source" is already registered.'
 			);
 		} );
+
+		it( 'should not be overriden if bootstrap happens after registration', () => {
+			// Register source in the client.
+			const clientProperties = {
+				getValues: () => 'values',
+				setValues: () => 'new values',
+				getPlaceholder: () => 'placeholder',
+				canUserEditValue: () => true,
+			};
+			registerBlockBindingsSource( {
+				name: 'core/custom-source',
+				label: 'Client Label',
+				...clientProperties,
+			} );
+
+			// Bootstrap source from the server.
+			unlock(
+				dispatch( blocksStore )
+			).addBootstrappedBlockBindingsSource( {
+				name: 'core/custom-source',
+				label: 'Server Label',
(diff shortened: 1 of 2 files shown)


packages/blocks/src/store/reducer.js
```diff
@@ -389,6 +392,11 @@ export function blockBindingsSources( state = {}, action ) {
 			return {
 				...state,
 				[ action.name ]: {
+					/*
+					 * Keep the exisitng properties in case the source has been registered
+					 * in the client before bootstrapping.
+					 */
+					...state[ action.name ],
 					label: action.label,
 					usesContext: action.usesContext,
```
artemiomorales: We'll need to reuse the logic from the [context PR](https://github.com/WordPress/gutenberg/pull/63513/files#diff-d5aba57655b6c1abcbd8f937a61ad98110057a6ae09f2db596811dca31c54030R1544-R1549) here to make sure the `usesContext` values get merged properly and the server doesn't overwrite the client values.

artemiomorales (commented):
So far, these changes make sense to me. I left [a comment](https://github.com/WordPress/gutenberg/pull/63513#discussion_r1687364717) on the context PR that relates back to this one; as you've mentioned, I think we can try to resolve that PR first to better inform this one.

packages/blocks/src/store/reducer.js
SantosGuillamot: You're right there will be "conclict changes" once any of the mentioned pull requests gets merged. I don't have a strong opinion of which one should go first, I believe it is irrelevant. I will rebase and resolve the issue once it gets merged.

packages/blocks/src/api/test/registration.js
```diff
@@ -1625,6 +1625,39 @@ describe( 'blocks', () => {
 				'Block bindings source "core/test-source" is already registered.'
 			);
 		} );
+
+		it( 'should not be overriden if bootstrap happens after registration', () => {
```
gziolo: The test description should explain better the behavior. How about:

```suggestion
		it( 'should correctly merge properties when bootstrap happens after registration', () => {
```

gziolo (approved):
I will leave the decision about the order of merging PRs to you. Based on the review from @artemiomorales and seeing it addresses the issue I raised earlier, I’m approving this PR. Thank you for a quick follow up.

SantosGuillamot:
I have merged [the other pull request](https://github.com/WordPress/gutenberg/pull/63513), rebased this one solving the conflicts, and reused the `mergedUsesContext` to ensure it is also properly managed in this use case. I also adapted the unit test to cover that use case.

With that, this pull request should be ready. Let's see if the tests pass.

packages/blocks/src/store/reducer.js
```diff
@@ -372,19 +372,19 @@ export function collections( state = {}, action ) {
 }
 
 export function blockBindingsSources( state = {}, action ) {
+	// Merge usesContext with existing values, potentially defined in the server registration.
+	let mergedUsesContext = [
```
gziolo: I was playing with a more condensed syntax and this is what I came up with:

Screenshot 2024-07-24 at 09 28 08

What do you think about this alternative? Does it read better?

gziolo:
I left a nitpick, which is totally optional as I was curious how to write one-liner to merge two arrays and keep only unique values. The code looks good after merging the latest changes from `trunk`.

packages/blocks/src/store/reducer.js
SantosGuillamot: I like this approach. There are a couple of problems though:
- The existing `usesContext` and the new `usesContext` can be undefined, so we can't use them directly.
- I belive we want to return `undefined` when both of them are `undefined`.

I made this change in [this commit](https://github.com/WordPress/gutenberg/pull/63797/commits/f296f917ffb2386f8e4cb02c4501b713bc5a8126#diff-36a4476193cdf5e625499e279b98b3d69bba444be76cd04d7df237b0b43408d2R376-R382). I created different variables just for the sake of readability.

Let me know what you think and if that's clearer. For me, it definitely reads better.

gziolo: What’s the reason there should be an `undefined` value? Could it default to an empty array otherwise? Would it simplify handling?

SantosGuillamot: I don't have a strong reason to have an `undefined` value. I was just expecting it to be `undefined` if it is not defined in the server or in the client. But maybe we can default to an empty array.

> Would it simplify handling?

We could remove [this last step](https://github.com/WordPress/gutenberg/pull/63797/files#diff-36a4476193cdf5e625499e279b98b3d69bba444be76cd04d7df237b0b43408d2R381-R382) where we are checking it the array is empty.

gziolo: https://github.com/WordPress/wordpress-develop/blob/ef71a9ed3917d11cf0daec94bc8f3201959af1b8/src/wp-includes/class-wp-block-bindings-source.php#L53

I see it defaults to `null` on the server so maybe keeping it as is would be more consistent. Both work fine as long as the handling on the client is adjusted accordingly. Thank you for providing additional context.

packages/blocks/src/store/reducer.js
```diff
@@ -372,19 +372,17 @@ export function collections( state = {}, action ) {
 }
 
 export function blockBindingsSources( state = {}, action ) {
+	// Merge usesContext with existing values, potentially defined in the server registration.
+	const existingUsesContext = state[ action.name ]?.usesContext || [];
```
gziolo, line 376: I had one realization afte this PR got merged. Currently, all reducers get executed whenever any action gets dispatched. In effect, the logic added on top of the function will get executed each time. I would recommend turning it into an util that gets called inside the `switch` in case it is really needed. 

The following could get added as an util:

```js
/**
 * Merges usesContext with existing values, potentially defined in the server registration.
 *
 * @param {string[]} existingUsesContext Existing `usesContext`.
 * @param {string[]} newUsesContext       Newly added `usesContext`.
 * @return {string[]|undefined} Merged `usesContext`.
 */
function getMergedUsesContext( existingUsesContext = [],  newUsesContext = [] ) {
	const mergedArrays = Array.from(
		new Set( existingUsesContext.concat( newUsesContext ) )
	);
	return mergedArrays.length > 0 ? mergedArrays : undefined;
}
```

That could be then used when updating reducers.

SantosGuillamot: I have created a follow-up pull request to discuss it and potentially fix it: [link](https://github.com/WordPress/gutenberg/pull/63941).

PR #63798, opened 2024-07-22
Site Editor Navigation Commands: Add permission check

### LINKED ISSUE #62460: Switching pages with command palette opens site editor even if user can't access it
### Description

While editing a page using the post editor, I can open the command palette and search for a different page to edit, but when I select the page, it opens in the site editor even if I can't access that.

1. Deny users the `edit_theme_options` capability (see snippet).
2. Open a page in the post editor.
3. Use the command palette to try to edit a different page.
4. Get "Sorry, you are not allowed to access this page."

Deny the `edit_theme_options` capability with:

```php
add_filter(
	'map_meta_cap',
	function ( $caps, $cap ) {
		if ( 'edit_theme_options' === $cap ) {
			$caps = [ 'do_not_allow' ];
		}

		return $caps;
	},
	10,
	2
);
```

### Environment info

- WordPress 6.6 beta 1
- Gutenberg plugin deactivated
- Twenty Twenty-Four theme

Yes

Yes

t-hamano:
Fixes #62460

## What?
This PR prevents users without `edit_theme_options` permissions from unintentionally seeing commands related to accessing the site editor.

The commands that are displayed unintentionally are Page, Template, and Template Part. Users without the `edit_theme_options` capability don't have access to the site editor, so they will get a warning when running these commands:

```
Sorry, you are not allowed to access this page.
```

## Why?

The current command palette checks whether the site editor is accessible via code like the following:

```javascript
const canCreateTemplate = useSelect( ( select ) => {
	return select( coreStore ).canUser( 'create', {
		kind: 'postType',
		name: 'wp_template',
	} );
}, [] );
```

However, this check did not exist for the three commands mentioned above.

### How?

Check whether the template can be created, i.e. whether the site editor is accessible.

If the site editor is not accessible, I either link to the post editor or don't display the command at all.

Whether a template can be created and whether the site editor is accessible may not necessarily coincide in the future, but I believe this is the best approach for now.

## Testing Instructions

- Open the post editor.
- Test the following four scenarios to confirm that commands appear or do not appear correctly, and that navigation works correctly.

Note: The scenarios marked with ✅ mark are the ones that will be fixed by this PR.

- Classic Theme, Admin user / Editor user:
  - `Hello World` (post) > Post Editor
  - `Sample page` (page) > Post Editor
- Block Theme, Admin user:
  - `Hello World` (post) > Post Editor
  - `Sample page` (page) > Site Editor
  - `Blog Home` (template) > Site Editor
  - `Post Meta` (template part) > Site Editor
- Block Theme, Editor user:
  - `Hello World` (post) > Post Editor
  - ✅ `Sample page` (page) > Post Editor
  - ✅ `Blog Home` (template) > Do not show
  - ✅ `Post Meta` (template part) > Do not show


packages/core-commands/src/hooks.js
```diff
@@ -1,12 +0,0 @@
-/**
- * WordPress dependencies
- */
-import { store as coreStore } from '@wordpress/core-data';
-import { useSelect } from '@wordpress/data';
-
-export function useIsBlockBasedTheme() {
```
t-hamano, line 7: Since this hook was not exposed and felt redundant, I removed it and switched to an ad-hoc approach.

packages/core-commands/src/site-editor-navigation-commands.js
```diff
@@ -51,7 +50,19 @@ function useDebouncedValue( value ) {
...
 		const history = useHistory();
-		const isBlockBasedTheme = useIsBlockBasedTheme();
+		const { isBlockBasedTheme, canCreateTemplate } = useSelect(
+			( select ) => {
+				return {
+					isBlockBasedTheme:
+						select( coreStore ).getCurrentTheme()?.is_block_theme,
+					canCreateTemplate: select( coreStore ).canUser( 'create', {
+						kind: 'postType',
+						name: 'wp_template',
```
youknowriad, line 60: Should we be checking "postType" here rather than `wp_template`

packages/core-commands/src/site-editor-navigation-commands.js
```diff
@@ -100,6 +111,7 @@ const getNavigationCommandLoaderPerPostType = ( postType ) =>
 				};
 
 				if (
+					! canCreateTemplate ||
 					postType === 'post' ||
 					( postType === 'page' && ! isBlockBasedTheme )
```
youknowriad, line 116: Why do we need the custom postType checks here? I know it's pre-existent but maybe we can remove this with the canCreateCheck (if it relies on post type)

packages/core-commands/src/site-editor-navigation-commands.js
```diff
@@ -153,7 +165,19 @@ const getNavigationCommandLoaderPerPostType = ( postType ) =>
...
 		const history = useHistory();
-		const isBlockBasedTheme = useIsBlockBasedTheme();
+		const { isBlockBasedTheme, canCreateTemplate } = useSelect(
+			( select ) => {
+				return {
+					isBlockBasedTheme:
+						select( coreStore ).getCurrentTheme()?.is_block_theme,
+					canCreateTemplate: select( coreStore ).canUser( 'create', {
+						kind: 'postType',
+						name: 'wp_template',
```
youknowriad: Should this be using templateType instead?

youknowriad:
I confirm the fix works but I feel like we should make it more generic (check postType)

packages/core-commands/src/site-editor-navigation-commands.js
t-hamano: Fixed by 62f081f

packages/core-commands/src/site-editor-navigation-commands.js
t-hamano: If we change this to `postType`, it won't work as intended. For example, non-admin users will be able to create Pages/Posts but won't have access to the site editor.

youknowriad: Maybe there's something I'm missing. Oh so the links go to the site editor? Maybe the links should go to the post editor in that case (if you're not already in the site editor)?

packages/core-commands/src/site-editor-navigation-commands.js
t-hamano: I think this is because the site editor only has the `page` post type screen yet - if the post type is `post`, we need to link to the post editor and not the site editor, even if it's a block theme or admin user.

packages/core-commands/src/site-editor-navigation-commands.js
```diff
@@ -177,8 +201,8 @@ const getNavigationCommandLoaderPerTemplate = ( templateType ) =>
 
 		const commands = useMemo( () => {
 			if (
-				! isBlockBasedTheme &&
-				! templateType === 'wp_template_part'
+				! canCreateTemplate ||
+				( ! isBlockBasedTheme && ! templateType === 'wp_template_part' )
```
youknowriad, line 205: I actually wonder if this second condition is needed.

youknowriad (approved):
Approving this PR. That said, I think there are some improvements and simplifications we can do:

 - If you're in the post editor, just navigate to the post editor for all entities that supports it (everything but patterns, template, template part, navigation).
 - If you're in the site editor, navigate to the site editor for all entities that support it.

I guess this is a temporary issue until we merge post and site editors entirely

packages/core-commands/src/site-editor-navigation-commands.js
t-hamano:
```
( ! isBlockBasedTheme && ! templateType === 'wp_template_part' )
```

This may be an edge case, but it's here to prevent themes like the one below from unintentionally accessing the site editor.

- The theme has template part files (e.g. `parts/header.html`)
- But the theme is not opted in to `add_theme_support( 'block-template-parts' )`

If a theme does not support `block-template-parts`, the theme will [not have access to the site editor](https://github.com/WordPress/wordpress-develop/blob/f003fd07ba5da8dd0266715d3874250040eab605/src/wp-admin/site-editor.php#L30) and should not expose template parts in the commands.

t-hamano:
Thanks for the review!

>  - If you're in the post editor, just navigate to the post editor for all entities that supports it (everything but patterns, template, template part, navigation).
>  - If you're in the site editor, navigate to the site editor for all entities that support it.

I agree with this and would like to discover and fix the inconsistencies in a follow-up.

PR #63799, opened 2024-07-22
New useBlockElementRef hook for storing block element into a ref

jsnajdr:
Implements idea from https://github.com/WordPress/gutenberg/pull/63565#discussion_r1682315349, replacing the `useBlockRef` hook with a new `useBlockElementRef` one. The difference is that the hook doesn't return a ref, but accepts one as a parameter. That's similar to how a JSX markup like `<div ref={ divRef } />` also accepts a ref as a parameter.

This hook can be used to easily implement the `useBlockElement` hook.

This PR also removes the `__unstable` prefixes from the exports because they are all internal API anyway, they are not exported from the `block-editor` package.

For correct types, I also had to add exports of the `RefCallback` and `Ref` React types to `element`.

### DIFF
--- a/packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
+++ b/packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
@@ -1,12 +1,7 @@
 /**
  * WordPress dependencies
  */
-import {
-	useContext,
-	useState,
-	useLayoutEffect,
-	useCallback,
-} from '@wordpress/element';
+import { useContext, useState, useLayoutEffect } from '@wordpress/element';
 import { useRefEffect } from '@wordpress/compose';
 
 /**
@@ -35,6 +30,14 @@ export function useBlockRefProvider( clientId ) {
 	);
 }
 
+function assignRef( ref, value ) {
+	if ( typeof ref === 'function' ) {
+		ref( value );
+	} else if ( ref ) {
+		ref.current = value;
+	}
+}
+
 /**
  * Tracks the DOM element for the block identified by `clientId` and assigns it to the `ref`
  * whenever it changes.
@@ -44,27 +47,16 @@ export function useBlockRefProvider( clientId ) {
  */
 export function useBlockElementRef( clientId, ref ) {
 	const { refsMap } = useContext( BlockRefs );
-	const setRef = useCallback(
-		( el ) => {
-			if ( typeof ref === 'function' ) {
-				ref( el );
-			} else if ( ref ) {
-				ref.current = el;
-			}
-		},
-		[ ref ]
-	);
-
 	useLayoutEffect( () => {
-		setRef( refsMap.get( clientId ) );
+		assignRef( ref, refsMap.get( clientId ) );
 		const unsubscribe = refsMap.subscribe( clientId, () =>
-			setRef( refsMap.get( clientId ) )
+			assignRef( ref, refsMap.get( clientId ) )
 		);
 		return () => {
 			unsubscribe();
-			setRef( null );
+			assignRef( ref, null );
 		};
-	}, [ refsMap, clientId, setRef ] );
+	}, [ refsMap, clientId, ref ] );
 }
 
 /**
--- a/packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
+++ b/packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
@@ -3,10 +3,9 @@
  */
 import {
 	useContext,
-	useMemo,
-	useRef,
 	useState,
 	useLayoutEffect,
+	useCallback,
 } from '@wordpress/element';
 import { useRefEffect } from '@wordpress/compose';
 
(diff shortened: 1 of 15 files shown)


packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
```diff
@@ -37,30 +36,35 @@ export function useBlockRefProvider( clientId ) {
...
+ * whenever it changes.
  *
- * @param {string} clientId The client ID to get a ref for.
- *
- * @return {RefObject} A ref containing the element.
+ * @param {string} clientId The client ID to track.
+ * @param {Ref}    ref      The ref object/callback to assign to.
  */
-function useBlockRef( clientId ) {
+export function useBlockElementRef( clientId, ref ) {
```
ellatrix, line 48: Why can't we set the `ref` that is provided here directly on `refsMap` to be used by the block?

jsnajdr: You mean, the refs would be stored directly in the `refsMap` instead of the element? And the `refsMap` job would be to "forward" the ref callback calls to the registered consumers?

I think in that case it would be hard to register a new ref at a time when the block is already mounted. When calling:
```js
useBlockElementRef( clientId, ref );
```
The `ref` wouldn't be set until the block element is changed or recreated.

packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
```diff
@@ -37,30 +36,35 @@ export function useBlockRefProvider( clientId ) {
...
-		[ refsMap ]
+	const setRef = useCallback(
+		( el ) => {
+			if ( typeof ref === 'function' ) {
+				ref( el );
+			} else if ( ref ) {
+				ref.current = el;
+			}
+		},
+		[ ref ]
```
ellatrix: Could we not avoid the `useCallback` here by moving the function within the `useLayoutEffect` that follows?

ellatrix: Or `setRef( ref, element )` could even be outside the hook

ellatrix: Haha, there's actually also `assignRef` in `use-merge-refs`

ellatrix: Not that we can use it here, it's a different package

https://github.com/WordPress/gutenberg/blob/47bde1a7696ceae94557dde3fa57cde3cfc9d84e/packages/compose/src/hooks/use-merge-refs/index.js#L18-L27

jsnajdr: Here we need to ensure that when the `ref` parameter in `useBlockElementRef( clientId, ref )` changes, the new version is called immediately. That's what React also does. When a different `ref` is passed to `<div ref={ ref } />`, then even though the `div` element is still the same, it will call the old ref with `null` and the new ref with the element.

packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
```diff
@@ -37,30 +36,35 @@ export function useBlockRefProvider( clientId ) {
...
+	useLayoutEffect( () => {
+		setRef( refsMap.get( clientId ) );
+		const unsubscribe = refsMap.subscribe( clientId, () =>
+			setRef( refsMap.get( clientId ) )
+		);
+		return () => {
+			unsubscribe();
+			setRef( null );
+		};
+	}, [ refsMap, clientId, setRef ] );
```
ellatrix: There's one subtle difference with the previous implementation: the element was available immediately on render. That's probably not needed, but just mentioning it here in case it was relied upon before.

jsnajdr: This is not true since #63565. On initial render, the return value is `null`. Only in an effect scheduled by this render is it set to the actual element.

When the element was available immediately on render, it was returning references to stale elements.

ellatrix: Oh ok! Thanks for clarifying

packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
ellatrix: The layout effect will still be called when adding `ref` as a dependency?

ellatrix (approved):

packages/block-editor/src/components/block-list/use-block-props/use-block-refs.js
jsnajdr: > Not that we can use it here, it's a different package

Interesting, I didn't know about that. The `ref.hasOwnProperty( 'current' )` check is nice, but not even React bothers to do it. It checks `ref !== null`, `typeof ref === 'function'`, and then if `ref.current` assignment leads to a crash, it will crash, it's you fault.

jsnajdr: > The layout effect will still be called when adding ref as a dependency?

Now I get it, of course you're right, fixing in 26e153a3d2f. It's a nice simplification.

fabiankaegy:
Hey @jsnajdr   👋 

Thanks in advance :)

jsnajdr:
Hi @fabiankaegy :wave: I believe that dev note is not needed here, because the PR modifies only internal APIs used inside the `@wordpress/block-editor` and is not exported to be used by anyone else. We removed the `__unstable` prefix not to stabilize a public API, but to reflect the fact that the API is internal and doesn't need such a prefix in the first place.

fabiankaegy:
Ahh my bad. Will remove it from the list then :)

PR #63801, opened 2024-07-22
Navigation: Don't output the main class on the inner wrapper

### LINKED ISSUE #62690: Navigation Block: block wrapper class name applied to both outer element and its child element

scruffian:
## What?
This removes the main block class from the `ul` inside in the `nav`.

Fixes https://github.com/WordPress/gutenberg/issues/62690

## Why?

The main block class should only be used once per block. 

## How?

Regex!

## Testing Instructions
1. Add a navigation block to a post or page
2. View the source
3. Check that the wp-block-navigation class is only used on the outer `nav` element, and not on the `ul` (or anywhere else).

### DIFF
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -181,6 +181,8 @@ class WP_Navigation_Block_Renderer {
 			)
 		);
 
+		$container_attributes = preg_replace('/(\s|\")(wp-block-navigation)(\s|\")/', '$1', $container_attributes);
+
 		$inner_blocks_html = '';
 		$is_list_open      = false;
 


carolinan:
Linking the issue: https://github.com/WordPress/gutenberg/issues/62690

MaggieCabrera:
I'm not sure about using a regex, maybe we shouldn't be using `get_block_wrapper_attributes()` at all. I added a navigation block with a custom background color, a text color and changed the font size to use a different preset. This is what I'm seeing:

Screenshot 2024-08-02 at 11 16 35

`wp-block-navigation` is not the only repeated class we have here

MaggieCabrera:
Also, looks like the e2e tests will need fixing because they are expecting the class to be present

getdave:
> Also, looks like the e2e tests will need fixing because they are expecting the class to be present

This is likely a sign of tests that need to be refactored towards perceivable APIs (as opposed to developer-centric ones).

getdave (commented):
Update: I've just noticed the related issue (it wasn't link in the PR description so I missed it).

------

This seems like a logical premise for a PR, but I would like to understand _why_ the classes were present on the `ul` in the first place. 

Is this an unhappy accident, or is there some fundamental historical reason?

Noting that classes are not considered a public API so we should be able to modify this markup.

getdave:
This is quite confusing. It seems that historically the Nav block has treated the topmost `<ul>` as a sort of pseudo "block wrapper". Indeed many of the block alignment/layout classes only work if they are placed on this `<ul>` as a result of the CSS selectors outputted by Core.

Therefore, unless it's possible to unpick this, then the best option is probably to document the inconsistency and then simply ensure the `wp-block-navigation` class is only output on the topmost `<nav>` element (the true "wrapper").

@scruffian your regex seems to work here, so it would be a case of understanding specifically why it's necessary to use the wrapper attributes on the `<ul>` as well as the root `<nav>` and documenting that in the code to avoid future confusion.

draganescu:
> . It seems that historically the Nav block has treated the topmost <ul> as a sort of pseudo "block wrapper".

As far as I remember the block did not always have a nav container? Maybe that is a left over from those incipient developments.

PR #63803, opened 2024-07-22
Reduce gap between steps in SpacingSizesControl, add animation, remove first/last marks

jameskoster:
## What?
Refine the appearance of `SpacingSizesControl`.

## Why?
The white space between options makes the control feel heavy and noisy, drawing more attention than is warranted. This PR reduces it slightly.

Animation on the value / handle add polish and underline the nature of the range.

## Before

https://github.com/user-attachments/assets/6c44b736-1b1e-45f9-9851-057c3121d364

## After

https://github.com/user-attachments/assets/cb7f5cc7-655d-4907-ab7b-1661dd2c721d

### DIFF
--- a/packages/block-editor/src/components/spacing-sizes-control/style.scss
+++ b/packages/block-editor/src/components/spacing-sizes-control/style.scss
@@ -42,11 +42,6 @@
 
 	.components-range-control__marks {
 		margin-top: 17px;
-
-		:last-child,
-		:first-child {
-			display: none;
-		}
 	}
 
 	.components-range-control__thumb-wrapper {
--- a/packages/block-editor/src/components/spacing-sizes-control/style.scss
+++ b/packages/block-editor/src/components/spacing-sizes-control/style.scss
@@ -32,6 +32,7 @@
 	}
 
 	.components-range-control__mark {
+		transform: translateX(-50%);
 		height: $grid-unit-05;
 		width: math.div($grid-unit-05, 2);
 		background-color: $white;
--- a/packages/block-editor/src/components/spacing-sizes-control/style.scss
+++ b/packages/block-editor/src/components/spacing-sizes-control/style.scss
@@ -4,9 +4,16 @@
 		margin-bottom: 0;
 	}
 
-	.components-range-control__track {
-		transition: width ease 0.1s;
-		@include reduce-motion("transition");
+	.is-marked {
+		.components-range-control__track {
+			transition: width ease 0.1s;
+			@include reduce-motion("transition");
+		}
+
+		.components-range-control__thumb-wrapper {
+			transition: left ease 0.1s;
+			@include reduce-motion("transition");
+		}
 	}
 
 	.spacing-sizes-control__range-control,
@@ -25,24 +32,28 @@
 	}
 
 	.components-range-control__mark {
-		height: $grid-unit-05;
-		width: 3px;
-		background-color: #fff;
+		height: 8px;
+		width: 2px;
+		background-color: $gray-300;
 		z-index: 1;
+		top: -6px;
+		border-radius: 2px;
+
+		&.is-filled {
+			z-index: -1;
+		}
 	}
 
 	.components-range-control__marks {
 		margin-top: 17px;
 
-		:first-child {
-			display: none;
+		:last-child {
+			transform: translateX(-2px);
 		}
 	}
 
 	.components-range-control__thumb-wrapper {
 		z-index: 3;
-		transition: left ease 0.1s;
-		@include reduce-motion("transition");
 	}
 }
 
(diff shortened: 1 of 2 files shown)


paaljoachim:
Nicely noticed James!

bgardner:
Looks good and tests well for me! (Padding, Margin, Block Spacing)

richtabor:
I'm not feeling it to be honest. It's more disorienting seeing the steps poke out a bit, especially when the control has a value. Something with seeing the accent color alongside the light gray colors.

It doesn't quite feel simpler or more communicative.  

CleanShot 2024-07-22 at 16 04 38

jameskoster:
Another option is to place the markers inside the track. Either as dots or lines:

| Dots | Lines |
| --- | --- |
| Screenshot 2024-07-23 at 11 42 33 | Screenshot 2024-07-23 at 11 47 00 |

Appreciate that's similar to what is currently shipping, but seems much less noisy to my eye. I acknowledge this is likely subjective :)

richtabor:
I appreciate the different explorations, but I don't think any are a big enough change/win to necessarily warrant the change. 

Perhaps just reducing the gap between sections to 2px would tighten this up a bit? 

CleanShot 2024-07-24 at 12 28 42

jameskoster:
Yes that might help.

jameskoster:
I reduced the space to `1.5px` (a value we use elsewhere). `2px` could work too though, no strong opinion. I left the animation as that still feels like a nice enhancement to me.

Let me know what you think.

Screenshot 2024-07-24 at 17 36 44

richtabor:
I lean towards 2px; Feels like a good balance, accounting for the general density of the other inspector controls. 

Yes, the animation is a nice touch. 

| 1.5px  | 2px |
|CleanShot 2024-07-27 at 11 02 15|CleanShot 2024-07-27 at 11 01 49|

bgardner:
I prefer the 2px spacing, FWIW.

jameskoster:
Updated the spacing and the OP.

packages/block-editor/src/components/spacing-sizes-control/style.scss
```diff
@@ -21,14 +33,16 @@
...
+		top: -#{$grid-unit-05};
 	}
 
 	.components-range-control__marks {
 		margin-top: 17px;
 
+		:last-child,
 		:first-child {
 			display: none;
 		}
```
stokesman: Good catch on this.
<details><summary>Alternatively, I think it’d be nice if the component didn’t render the first or last marks to begin with. Then this ruleset can be removed.</summary>

```diff
diff --git a/packages/block-editor/src/components/spacing-sizes-control/input-controls/spacing-input-control.js b/packages/block-editor/src/components/spacing-sizes-control/input-controls/spacing-input-control.js
index 6605d2de50..8e20942dd8 100644
--- a/packages/block-editor/src/components/spacing-sizes-control/input-controls/spacing-input-control.js
+++ b/packages/block-editor/src/components/spacing-sizes-control/input-controls/spacing-input-control.js
@@ -193,10 +193,12 @@ export default function SpacingInputControl( {
 		name: size.name,
 	} ) );
 
-	const marks = spacingSizes.map( ( _newValue, index ) => ( {
-		value: index,
-		label: undefined,
-	} ) );
+	const marks = spacingSizes
+		.slice( 1, spacingSizes.length - 1 )
+		.map( ( _newValue, index ) => ( {
+			value: index + 1,
+			label: undefined,
+		} ) );
 
 	const sideLabel =
 		ALL_SIDES.includes( side ) && showSideInLabel ? LABELS[ side ] : '';

```
</details>

packages/block-editor/src/components/spacing-sizes-control/style.scss
```diff
@@ -21,14 +33,16 @@
 
 	.components-range-control__mark {
 		height: $grid-unit-05;
```
stokesman: It’s quite minute (especially now with their reduced width), but I believe the base positioning of the marks don’t account for their width so this should place them more accurately.
```suggestion
	.components-range-control__mark {
		transform: translateX(-50%);
		height: $grid-unit-05;
```

stokesman (commented):
I liked the Lines variation https://github.com/WordPress/gutenberg/pull/63803#issuecomment-2244882015 because the darker mark color feels more like a divot and doesn’t visually break the bar apart. Also, it makes the control less dependent on its background. Right now, if used on a background other than white the styles have to be overridden to make the marks match that background. Not that I feel too strongly on those points.

richtabor (approved):
LGTM

jameskoster:
Thanks for the feedback @stokesman, I addressed your points 👍 

> doesn’t visually break the bar apart

This bugs me too, but it seems to be quite subjective, so a detail to explore in isolation separately.

PR #63806, opened 2024-07-22
Fix unlabeled Spacer block controls

### LINKED ISSUE #63800: Spacer block: unlabeled Range and Input controls
### Description

The Spacer block can be used to add vertical or horizontal spacing.

In both cases, two settings controls are unlabeled. One more case of completely unlabeled controls only because the `label` prop is not required or used incorrectly. Cc @WordPress/gutenberg-components @joedolson 

Vertical spacing:
The range control highlighted in the screenshot below is unlabaled:

Screenshot 2024-07-22 at 12 51 45

Horizontal spacing:
The input field highlighted in the screenshot below is unlabaled:

Screenshot 2024-07-22 at 12 51 39

- Add a Row with a paragraph, a spacer block and another paragraph
- Select the spacer.
- Inspect the width input control when the width is set to 'fixed'. 
- Observe the input field is not labeled.
- Add a normal Spacer block and select it.
- Inspect the width range control.
- Observe it's not labeled.

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63800

## What?

Two Spacer block controls are unlabeled.

## Why?

All form controls must be labeled.

## How?

- Adds missing `label` prop to the `UnitControl` in `FlexControls`.
- Adds missing `label` prop to `RangeControl` in `SpacingInputControl`.
- Removes unnecessary `BaseControl` wrapping `UnitControl` in the Spacer block inspector controls.

Queation regarding the last point: I noticed other cases where a `BaseControl` wraps another control. I'm not sure to understand why. Any clue? Anything I may be missing?

## Testing Instructions

- Repeat the testing instructions from the issue https://github.com/WordPress/gutenberg/issues/63800
- Observe the two controls are now properly labeled.

### DIFF
--- a/packages/block-library/src/spacer/controls.js
+++ b/packages/block-library/src/spacer/controls.js
@@ -61,13 +61,13 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
 					isResetValueOnUnitChange
 					min={ MIN_SPACER_SIZE }
 					onChange={ handleOnChange }
-					style={ { maxWidth: 80 } }
+					style={ { maxWidth: '50%' } }
 					value={ computedValue }
 					units={ units }
 					label={ label }
+					__next40pxDefaultSize
 				/>
 			) }
-
 			{ spacingSizes?.length > 0 && (
 				<View className="tools-panel-item-spacing">
 					<SpacingSizesControl
--- a/packages/block-library/src/spacer/controls.js
+++ b/packages/block-library/src/spacer/controls.js
@@ -10,7 +10,6 @@ import {
 	privateApis as blockEditorPrivateApis,
 } from '@wordpress/block-editor';
 import {
-	BaseControl,
 	PanelBody,
 	__experimentalUseCustomUnits as useCustomUnits,
 	__experimentalUnitControl as UnitControl,
@@ -57,17 +56,16 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
 	return (
 		<>
 			{ ( ! spacingSizes || spacingSizes?.length === 0 ) && (
-				<BaseControl label={ label } id={ inputId }>
-					<UnitControl
-						id={ inputId }
-						isResetValueOnUnitChange
-						min={ MIN_SPACER_SIZE }
-						onChange={ handleOnChange }
-						style={ { maxWidth: 80 } }
-						value={ computedValue }
-						units={ units }
-					/>
-				</BaseControl>
+				<UnitControl
+					id={ inputId }
+					isResetValueOnUnitChange
+					min={ MIN_SPACER_SIZE }
+					onChange={ handleOnChange }
+					style={ { maxWidth: 80 } }
+					value={ computedValue }
+					units={ units }
+					label={ label }
+				/>
 			) }
 
 			{ spacingSizes?.length > 0 && (
--- a/packages/block-library/src/spacer/controls.js
+++ b/packages/block-library/src/spacer/controls.js
@@ -55,7 +55,7 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
 
 	return (
 		<>
-			{ ( spacingSizes || spacingSizes?.length === 0 ) && (
+			{ ( ! spacingSizes || spacingSizes?.length === 0 ) && (
 				<UnitControl
 					id={ inputId }
 					isResetValueOnUnitChange
--- a/packages/block-library/src/spacer/controls.js
+++ b/packages/block-library/src/spacer/controls.js
@@ -14,7 +14,6 @@ import {
 	__experimentalUseCustomUnits as useCustomUnits,
 	__experimentalUnitControl as UnitControl,
 	__experimentalParseQuantityAndUnitFromRawValue as parseQuantityAndUnitFromRawValue,
-	__experimentalGrid as Grid,
 } from '@wordpress/components';
 import { useInstanceId } from '@wordpress/compose';
 import { View } from '@wordpress/primitives';
@@ -56,19 +55,17 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
 
 	return (
 		<>
-			{ ( ! spacingSizes || spacingSizes?.length === 0 ) && (
-				<Grid gap="2" templateColumns="1fr 1fr 24px">
-					<UnitControl
-						id={ inputId }
-						isResetValueOnUnitChange
-						min={ MIN_SPACER_SIZE }
-						onChange={ handleOnChange }
-						value={ computedValue }
-						units={ units }
-						label={ label }
-						__next40pxDefaultSize
-					/>
-				</Grid>
+			{ ( spacingSizes || spacingSizes?.length === 0 ) && (
+				<UnitControl
+					id={ inputId }
+					isResetValueOnUnitChange
+					min={ MIN_SPACER_SIZE }
+					onChange={ handleOnChange }
+					value={ computedValue }
+					units={ units }
+					label={ label }
+					__next40pxDefaultSize
+				/>
 			) }
 			{ spacingSizes?.length > 0 && (
 				<View className="tools-panel-item-spacing">
--- a/packages/block-library/src/spacer/controls.js
+++ b/packages/block-library/src/spacer/controls.js
@@ -14,6 +14,7 @@ import {
 	__experimentalUseCustomUnits as useCustomUnits,
 	__experimentalUnitControl as UnitControl,
 	__experimentalParseQuantityAndUnitFromRawValue as parseQuantityAndUnitFromRawValue,
+	__experimentalGrid as Grid,
 } from '@wordpress/components';
 import { useInstanceId } from '@wordpress/compose';
 import { View } from '@wordpress/primitives';
@@ -56,17 +57,18 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
 	return (
 		<>
 			{ ( ! spacingSizes || spacingSizes?.length === 0 ) && (
-				<UnitControl
-					id={ inputId }
-					isResetValueOnUnitChange
-					min={ MIN_SPACER_SIZE }
-					onChange={ handleOnChange }
-					style={ { maxWidth: '50%' } }
-					value={ computedValue }
-					units={ units }
-					label={ label }
-					__next40pxDefaultSize
-				/>
+				<Grid gap="2" templateColumns="1fr 1fr 24px">
+					<UnitControl
+						id={ inputId }
+						isResetValueOnUnitChange
+						min={ MIN_SPACER_SIZE }
+						onChange={ handleOnChange }
+						value={ computedValue }
+						units={ units }
+						label={ label }
+						__next40pxDefaultSize
+					/>
+				</Grid>
 			) }
 			{ spacingSizes?.length > 0 && (
 				<View className="tools-panel-item-spacing">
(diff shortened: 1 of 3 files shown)


joedolson:
Reading the [documentation on `BaseControl`](https://developer.wordpress.org/block-editor/reference-guides/components/base-control/), I think this component isn't being used the way it's supposed to here, so it's not rendering the label. But it's a bit obscure. It's *also* possible that the BaseControl should be rendering a label, and isn't; in which case that component maybe needs to be looked at.

It looks like BaseControl expects a child element that's an input; but this is a child that's a component, which is a bit different.

afercia:
Yes, to my understanding the BaseControl is a sort of wrapper used to generate labels and descriptions. That is already used for most (all?) of the input controls internally.

In this case, the [UnitControl](https://github.com/WordPress/gutenberg/blob/d625b3d19d57e2cc165f4fa8c960e8f2667dce9a/packages/components/src/unit-control/index.tsx) renders a `ValueInput`, which is a styled [NumberControl](https://github.com/WordPress/gutenberg/blob/d625b3d19d57e2cc165f4fa8c960e8f2667dce9a/packages/components/src/number-control/index.tsx) which in turn renders an `Input`, which is a styled [InputControl](https://github.com/WordPress/gutenberg/blob/d625b3d19d57e2cc165f4fa8c960e8f2667dce9a/packages/components/src/input-control/index.tsx) which internally already uses the BaseControl.

@WordPress/gutenberg-components when you have a chance, could you please clarify how the BaseControl is supposed to be used? In this case, seems to me it was used incorrectly and I think there are more cases where its usage is unclear or incorrect. Which makes me think the BaseControl usage should be audited and the documentation improved.

mirka:
> @WordPress/gutenberg-components when you have a chance, could you please clarify how the BaseControl is supposed to be used? In this case, seems to me it was used incorrectly and I think there are more cases where its usage is unclear or incorrect. Which makes me think the BaseControl usage should be audited and the documentation improved.

In this particular [case](https://github.com/WordPress/gutenberg/pull/63806/files#diff-dc82d64e09778668b1ce1591d521fcd10c745dd88bdf279455d55b32b180eba5), it's completely unnecessary in the sense that it can be accomplished with just a UnitControl, but not strictly "incorrect". BaseControl can be nested safely if you know what you're doing. This is the rendered output of this code path, and it's labeled visually and accessibly as expected:

UnitControl wrapped in an unnecessary BaseControl

(Should we add a `__next40pxDefaultSize` while we're at it?)

I do agree that the BaseControl docs are a bit sparse and could use some improvement!

afercia:
> Should we add a __next40pxDefaultSize while we're at it?

Sure. There's a problem though as the `UnitControl` has an max-width style with an arbitrary value of 80 pixels. When using `__next40pxDefaultSize` long numbers may be cut-off. Screenshot where I'm showing both controls (normally they're rendered conditionally) to better illustrate:

Screenshot 2024-07-24 at 10 56 43

afercia:
Last commit adds the 40px size to the conditionally-shown UnitControl. Max-width increased to 50%.

Screenshot: Reminder: these two controls are rendered conditionally, the UI doesn't show them at the same time.

Screenshot 2024-07-24 at 11 52 40

packages/block-library/src/spacer/controls.js
```diff
@@ -57,19 +56,18 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
...
-						value={ computedValue }
-						units={ units }
-					/>
-				</BaseControl>
+				<UnitControl
+					id={ inputId }
+					isResetValueOnUnitChange
+					min={ MIN_SPACER_SIZE }
+					onChange={ handleOnChange }
+					style={ { maxWidth: '50%' } }
```
tyxla: The `50%` width still feels arbitrary. I wonder if it should have the same width as it has when it has the slider control.

afercia: > The `50%` width still feels arbitrary. I wonder if it should have the same width as it has when it has the slider control.

I considered that. Alas, that width is something like 104.some decimals pixels because it's determined by the flax layout. Not a real number we can use. I'd agree 50% is still arbitrary... but it is a little better 😄

afercia: Aside consideration: this is the kind of problems to face when the design asks for something the base components aren't designed for. Local overrides, custom, ad-hoc implementation should generally be avoided.

mirka: If we want to place it on a standard 2-column controls grid, I would recommend something like this:

```diff
diff --git a/packages/block-library/src/spacer/controls.js b/packages/block-library/src/spacer/controls.js
index 97e47e4d15..7d09af9969 100644
--- a/packages/block-library/src/spacer/controls.js
+++ b/packages/block-library/src/spacer/controls.js
@@ -56,17 +56,18 @@ function DimensionInput( { label, onChange, isResizing, value = '' } ) {
 	return (
 		<>
 			{ ( ! spacingSizes || spacingSizes?.length === 0 ) && (
-				<UnitControl
-					id={ inputId }
-					isResetValueOnUnitChange
-					min={ MIN_SPACER_SIZE }
-					onChange={ handleOnChange }
-					style={ { maxWidth: '50%' } }
-					value={ computedValue }
-					units={ units }
-					label={ label }
-					__next40pxDefaultSize
-				/>
+				<div className="block-library-spacer__controls-grid">
+					<UnitControl
+						id={ inputId }
+						isResetValueOnUnitChange
+						min={ MIN_SPACER_SIZE }
+						onChange={ handleOnChange }
+						value={ computedValue }
+						units={ units }
+						label={ label }
+						__next40pxDefaultSize
+					/>
+				</div>
 			) }
 			{ spacingSizes?.length > 0 && (
 				<View className="tools-panel-item-spacing">
diff --git a/packages/block-library/src/spacer/editor.scss b/packages/block-library/src/spacer/editor.scss
index e0c700795c..f2ca2ef985 100644
--- a/packages/block-library/src/spacer/editor.scss
+++ b/packages/block-library/src/spacer/editor.scss
@@ -46,3 +46,9 @@
 		height: 100% !important;
 	}
 }
+
+.block-library-spacer__controls-grid {
+	display: grid;
+	grid-template-columns: repeat(2, 1fr);
+	gap: 16px;
+}
```

(Ideally I want some kind of standard component to handle this layout #43423)

However, seeing that [the other UnitControl](https://private-user-images.githubusercontent.com/1682452/350939995-5cb69420-6105-4b6f-aedd-ef2565b1913e.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjE4Mzg5MDMsIm5iZiI6MTcyMTgzODYwMywicGF0aCI6Ii8xNjgyNDUyLzM1MDkzOTk5NS01Y2I2OTQyMC02MTA1LTRiNmYtYWVkZC1lZjI1NjViMTkxM2UucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDcyNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA3MjRUMTYzMDAzWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YTMwNzZmMDkzMjQ2N2RkYjYzZDdiNDJiNDI2YmM5ZjkyYzZmMzhhNjMxMTI5MTIxYjViZGNmOTYzMWYwNDI5NiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.7iDkIhf8F_e01rM2tBtitaBtFwMlIs9Pg2LwClRW3fk) is full-width, I guess this one could just be full-width as well.

afercia: > I want some kind of standard component to handle this l

Yes that would be better. I can try your CSS workaround but that's in some way an ad-hoc, custom override that should be avoided.

> seeing that the other UnitControl

The linked image gives a 404 error.

afercia:
In the last commit I'm trying the Grid component to make the single UnitControl width be the same of the full one. Screenshot with both controls rendered, for comparison:

Screenshot 2024-07-25 at 10 14 13

However, in my opinion w e shouldn't do this as it would be a unique case implementation, locally overriding the default layout of these controls.

tyxla:
> In the last commit I'm trying the Grid component to make the single UnitControl width be the same of the full one. Screenshot with both controls rendered, for comparison:
> 
> Screenshot 2024-07-25 at 10 14 13
> 
> However, in my opinion w e shouldn't do this as it would be a unique case implementation, locally overriding the default layout of these controls.

I tend to agree with you; since we've been trying to reduce special cases and overrides, it might be a good idea to avoid it altogether. 

Perhaps we can do what @mirka suggested above:

> seeing that [the other UnitControl](https://private-user-images.githubusercontent.com/1682452/350939995-5cb69420-6105-4b6f-aedd-ef2565b1913e.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjE4Mzg5MDMsIm5iZiI6MTcyMTgzODYwMywicGF0aCI6Ii8xNjgyNDUyLzM1MDkzOTk5NS01Y2I2OTQyMC02MTA1LTRiNmYtYWVkZC1lZjI1NjViMTkxM2UucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDcyNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA3MjRUMTYzMDAzWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YTMwNzZmMDkzMjQ2N2RkYjYzZDdiNDJiNDI2YmM5ZjkyYzZmMzhhNjMxMTI5MTIxYjViZGNmOTYzMWYwNDI5NiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.7iDkIhf8F_e01rM2tBtitaBtFwMlIs9Pg2LwClRW3fk) is full-width, I guess this one could just be full-width as well 

(although I'm not sure which one she was referring to).

afercia:
Full width sounds good to me, thanks. On it.

afercia:
The single unit control is now full width. Screenshot:

Screenshot 2024-07-25 at 13 40 58

packages/block-library/src/spacer/controls.js
mirka: Oops, I was trying to link to this screenshot that was in your original issue:

The other UnitControl

mirka (approved):
Looks good, thanks for the fixes!

afercia:
Thanks all for your reviews and feedback.

Issue #63795, opened 2024-07-22
New block editor z-index issue. Link options are hidden at the bottom of the page


83pixel:
### Description

When you are editing text in the new block editor if your text is on the bottom of the page and you are trying to add/edit a link, the options are hidden if you have some options for example seo options. This is due to no z-index on the class .editor-visual-edit. A simple z-index: 5 will do the trick

![before](https://github.com/user-attachments/assets/5b98def8-30af-4161-9b8c-52ed0a10c6b7)

![after](https://github.com/user-attachments/assets/a9e71f25-c09b-47e0-bbc6-14f74db4af78)

### Step-by-step reproduction instructions

Go to New Post
Add a long Text
Try to add a link at a bottom text

### Screenshots, screen recording, code snippet

.editor-visual-editor {
        align-items: center;
        background-color: #ddd;
        display: block;
        height: 100%;
        isolation: isolate;
        position: relative;
        **z-index: 5;**
}

Any positive z-index will fix the problem

### Environment info

WordPress 6.6 fresh install

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

- [X] Yes

rgerhards:
I can confirm this bug. It's very annoying, as I have source citations usually at the end of the posting. I  want them to open in a new window. Since a couple of updaes, this no longer is possible during link insertion, but only afterwards via the property dialog. This is now also no longer available ;-)

Work-around: add 6 blank paragraphs, edit link, delete these 6 paragraphs. Just if someone else comes across this bug.

talldan:
Thanks for reporting, though there's already an existing issue for this - #62998.

I'll close this as a duplicate.

stokesman:
> Work-around: add 6 blank paragraphs, edit link, delete these 6 paragraphs. Just if someone else comes across this bug.

That’s a solid workaround. I did want to mention that quicker one can work given there’s enough content in the page that scrolling up will leave too little room for the popover to fit below the text. Here’s demo:

https://github.com/user-attachments/assets/96c9073a-6c35-4c93-bdda-e686df448698

Still, the bug is annoying but it might help save a little frustration.

PR #63809, opened 2024-07-22
Fix: New block editor z-index issue. Link options are hidden at the bottom of the page 

### LINKED ISSUE #62998: Inline popover hides behind metabox
### Description

When meta boxes are enabled, if you try to highlight text, for example in the last paragraph block, the popover will be hidden behind the meta box.

This problem may be related to #62681. Disabling `isolation:isolate`, which was added in #62681, will resolve this problem.

- Enable custom fields.
- Insert a few paragraphs and apply a highlight to the text of the last paragraph.
- Open the custom field panel if it is closed.
- Scroll your browser so that the popover appears below the text.
- The popover will now be hidden behind the meta box.

https://github.com/WordPress/gutenberg/assets/54422211/4c94523e-1fd3-4c0e-8c41-631fdfd41f68

### Environment info

_No response_

Yes

Yes

Rishit30G:
## What?
PR is intended to fix this issue: https://github.com/WordPress/gutenberg/issues/63795

## Why?
When you are editing text in the new block editor if your text is on the bottom of the page and you are trying to add/edit a link, the options are hidden if you have some options for example seo options

## How?
Added `z-index: 5` to `.editor-visual-editor` as per the suggestion  

## Testing Instructions
Works as expected, the results are now no more blocked by the box below the text

## Screenshots or screencast 
image

### DIFF
--- a/packages/base-styles/_z-index.scss
+++ b/packages/base-styles/_z-index.scss
@@ -21,6 +21,7 @@ $z-layers: (
 	".interface-complementary-area .components-panel__header": 1, // higher sticky element
 
 	".components-modal__header": 10,
+	".edit-post-layout.has-metaboxes .edit-post-visual-editor": 1, // The popovers contained in visual editor go in front of the metabox container.
 	".edit-post-meta-boxes-area.is-loading::before": 1,
 	".edit-post-meta-boxes-area .spinner": 5,
 	".components-popover__close": 5,
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -8,6 +8,10 @@
 	clear: both;
 }
 
+.edit-post-layout.has-metaboxes .edit-post-visual-editor {
+	z-index: 99;
+}
+
 // Adjust the position of the notices
 .components-editor-notices__snackbar {
 	position: fixed;
--- a/packages/edit-post/src/components/layout/style.scss
+++ b/packages/edit-post/src/components/layout/style.scss
@@ -9,7 +9,7 @@
 }
 
 .edit-post-layout.has-metaboxes .edit-post-visual-editor {
-	z-index: 99;
+	z-index: z-index(".edit-post-layout.has-metaboxes .edit-post-visual-editor");
 }
 
 // Adjust the position of the notices
--- a/packages/editor/src/components/visual-editor/style.scss
+++ b/packages/editor/src/components/visual-editor/style.scss
@@ -5,7 +5,6 @@
 	background-color: $gray-300;
 	// Make this a stacking context to contain the z-index of children elements.
 	isolation: isolate;
-	z-index: 5;
 
 	// Centralize the editor horizontally (flex-direction is column).
 	align-items: center;
--- a/packages/editor/src/components/visual-editor/style.scss
+++ b/packages/editor/src/components/visual-editor/style.scss
@@ -5,6 +5,7 @@
 	background-color: $gray-300;
 	// Make this a stacking context to contain the z-index of children elements.
 	isolation: isolate;
+	z-index: 5;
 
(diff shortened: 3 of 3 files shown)


stokesman (commented):
Thanks for the PR! I see this change is per the suggestion in #63795. This works but the `z-index` isn’t required in all contexts so applying it in the component’s own styles may necessitate overriding it in other contexts.

 In the earlier issue on the bug #62998, I had made this suggestion:
```css
.edit-post-layout.has-metaboxes .edit-post-visual-editor {
    z-index: 1;
}
```
This way the `z-index` is applied by a parent context. That would go in this file: https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-post/src/components/layout/style.scss

Rishit30G:
Hey @stokesman, 
Thanks for sharing feedback on my PR, however on setting `z-index = 1` the results are still hiding behind the metabox content: 

image


On setting `z-index: 99` it's working fine: 

image

stokesman:
> however on setting `z-index = 1` the results are still hiding behind the metabox content:

That’s a great catch. Setting the `z-index` to a greater value like `99` works for Yoast’s metabox however it’s unknown if other plugins use larger values. So alternatively, the metabox area could be made its own stacking context. The most explicit way I know how is to add `isolation: isolate` to it.

Here’s a screen capture demo:

https://github.com/user-attachments/assets/3751c646-2854-4c98-8297-caddef31a6f4

Still, I suppose there is some potential issue with that if there is some plugin creating a sort of popover like UI from within the metabox area that overlays the editor canvas or other parts of the editor interface. By making the metabox area a stacking context all elements within it can’t be in front of anything the metabox area is not in front of.

So I'm not sure what the best idea is here. Perhaps @t-hamano or @talldan have a helpful thought on this.

---- 

I pushed a commit that applies the `z-index` per the project’s convention around z-indexes. Oh, I also made it `1` but of course we may end up changing that.

t-hamano:
I think the ideal approach would be to use `isolation: isolate;`, as we did in #62681, i.e. CSS like this:

```css
.edit-post-layout__metaboxes {
	isolation: isolate;
}

.edit-post-layout.has-metaboxes .edit-post-visual-editor {
	z-index: z-index(".edit-post-layout.has-metaboxes .edit-post-visual-editor");
}
```

> By making the metabox area a stacking context all elements within it can’t be in front of anything the metabox area is not in front of.

Does this mean that, for example, if we render a popover _inside_ the metabox, any part that extends above the metabox will be clipped?

stokesman:
> if we render a popover _inside_ the metabox, any part that extends above the metabox will be clipped?

Yes. Though only when the element is _inside the DOM tree_ of the metabox. If a plugin were using `Popover` from the components package in a default fashion it wouldn’t be a child of the metabox container so should be fine. Still, who knows what plugins are doing. Also, to say clipped is precise as long as you mean generally not visible. It would be behind the visual editor but not clipped by the metabox container.

I suppose the potential issue could be guarded against with `:focus-within` that puts the metabox container back in front:
```css
.edit-post-layout__metaboxes:focus-within {
	z-index: 1;
}
```

Though, now I'm thinking Aki’s initial idea https://github.com/WordPress/gutenberg/issues/62998#issuecomment-2198580816 for resolving this may be the best. That is to make a shared ancestor of the visual editor and the metabox area a stacking context instead of the visual editor.

t-hamano:
> Though, now I'm thinking Aki’s initial idea https://github.com/WordPress/gutenberg/issues/62998#issuecomment-2198580816 for resolving this may be the best. That is to make a shared ancestor of the visual editor and the metabox area a stacking context instead of the visual editor.

I tried this approach in #63939.

t-hamano:
Hi @Rishit30G,

This issue has been resolved in #63939. Of course, you have been added as a co-author to the props.

I would like to close this PR, but thank you for your efforts!

PR #63811, opened 2024-07-22
Fix selected row styles in table layout

jameskoster:
Follow up to https://github.com/WordPress/gutenberg/pull/63694.

Fixes a regression where table rows were no longer showing theme-color accents on selection.

### Before
Screenshot 2024-07-22 at 16 10 43

### After
Screenshot 2024-07-22 at 16 10 20


youknowriad (approved):
Thanks for the fix.

PR #63812, opened 2024-07-22
Update: Simplify some permission checks.

jorgefilipecosta:
Simplifies some permission checks by using canUser instead of checking capabilities directly which is not as reliable (e.g: may not take into account some backend filters).

Props to @Mamaduka for the research that led to this PR.

## Testing Instructions
I checked the editor still works as expected.
With an author role, I checked in the developer tools that there are no 403 HTTP errors while opening the editor.

### DIFF
--- a/packages/editor/src/components/global-styles-provider/index.js
+++ b/packages/editor/src/components/global-styles-provider/index.js
@@ -33,24 +33,18 @@ export function mergeBaseAndUserConfigs( base, user ) {
 function useGlobalStylesUserConfig() {
 	const { globalStylesId, isReady, settings, styles, _links } = useSelect(
 		( select ) => {
-			const {
-				getEditedEntityRecord,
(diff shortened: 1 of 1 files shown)


youknowriad:
This PR had a weird/negative impact on the "navigate site editor" metric. Any ideas what that might be?

Mamaduka:
> This PR had a weird/negative impact on the "navigate site editor" metric. Any ideas what that might be?

Probably two additional `OPTIONS` requests that are now made. The `useGlobalStylesContext` hook is also called by `useBlockEditorSettings`, which is usually placed in a critical path for the App.

We need to re-evaluate global style selectors and allow `read` access via `context=view` or something.

Mamaduka:
@youknowriad, create a follow-up to see if we can improve performance - #65177.

PR #63814, opened 2024-07-22
Update: Simplify and do not pass renderingMode on editor SidebarContent.

jorgefilipecosta:
Applies a suggestion by @youknowriad at https://github.com/WordPress/gutenberg/pull/62033#pullrequestreview-2158641939. Removes the querying and passing of renderingMode on SidebarContent to TemplateContentPanel.
TemplateContentPanel now retrieves the renderingMode from the store instead of receiving it as a prop.

## Testing Instructions
I opened a template and verified the template content panel still works as expected.


youknowriad (approved):
Thanks

PR #63815, opened 2024-07-22
Apply minimal variant to pagination dropdown

jameskoster:
## What?
Apply `minimal` variant to pagination control in data views.

## Why?
The current implementation gives outsized importance to this element.

| Before | After |
| --- | --- |
| Screenshot 2024-07-22 at 16 40 30 | Screenshot 2024-08-13 at 11 22 10 Screenshot 2024-08-13 at 11 22 04 |

## Testing Instructions
1. Navigate to a data view 
2. Observe the new page selection UI
3. Check it works as before

I suspect there's a better way to handle the i18n, so please feel free to push to this branch :)


t-hamano:
I noticed that PR #63691, which is similar to this one, has been submitted. Can we move forward with #63691? Since #63691 only added the minimal variation, other improvements may be followed up with this PR.

jameskoster:
I'm not sure that https://github.com/WordPress/gutenberg/pull/63691 is much of an improvement without the other changes in this PR. The dropdown looks very disconnected from the prefix/suffix. Totally fine to close this and work on that PR instead though.

packages/dataviews/src/components/dataviews-pagination/style.scss
```diff
@@ -12,10 +12,12 @@
...
 
 .dataviews-pagination__page-selection {
-	font-size: 11px;
-	text-transform: uppercase;
-	font-weight: 500;
-	color: $gray-900;
+	.components-select-control__input {
+		&:hover,
+		&:focus {
+			color: $gray-900;
```
jameskoster: cc @mirka when using the `minimal` variant of `SelectControl`, some wp-admin styles bleed through affecting the hover/focus color. I wonder if we should fix that in the component?

mirka: Agreed! Fix at #63855

t-hamano:
Would it be better to go ahead with this PR and add @dhananjaykuber to the props when it is merged?

packages/dataviews/src/components/dataviews-pagination/index.tsx
```diff
@@ -40,38 +40,35 @@ function DataViewsPagination() {
...
+								return {
+									value: page.toString(),
+									label:
+										__( 'Page ' ) +
+										page.toString() +
+										sprintf(
+											// translators: %s: Total number of pages.
+											_x( ' of %s', 'paging' ),
+											totalPages
+										),
```
t-hamano: This configuration may cause problems with localization. Perhaps we could configure it like this:

```javascript
sprintf(
	// translators: 1: Current page number, 2: Total number of pages
	__( 'Page %1$s of %2$s' ),
	page.toString(),
	totalPages
)
```

t-hamano (commented):
Now that #63855 has been merged, can we remove [this CSS](https://github.com/WordPress/gutenberg/pull/63815/files#diff-dfa263183d286d8ce4a1a2d782cae80ba836bd9ad9d4279e74ee1dd4afb30c46R18) after rebasing this PR?

By the way, I'm considering applying the approach taken in this PR to the Font Library Modal as well. In that case, it will look like this, but is this as expected? The font library has a particularly large number of pages.

| Before | After |
|--------|--------|
| before | after |

jameskoster:
Yeah that's not ideal. Perhaps we can do something with the `label` and `suffix` props to mitigate.

jameskoster:
Using the suffix didn't really work and resulted in worse usability (the click target was tiny). 

This is the closest I can get to the intended design, by using the inline label:

Screenshot 2024-07-24 at 11 55 11

But it's not perfect. "Page" is omitted from the `option`s which reduces the length a bit, but the styles applied to the select carry through to the options which looks pretty clunky:

Screenshot 2024-07-24 at 11 55 19

We might need to rethink the design a bit.

t-hamano:
Is the problem here that `1 of X` is being forced into upper case?

If so, it looks like we should just delete [this `text-transform: uppercase`](https://github.com/WordPress/gutenberg/blob/update/pagination-dropdown/packages/dataviews/src/components/dataviews-pagination/style.scss#L29).

On my machine it looks like this:

| With `text-transform: uppercase` | Without `text-transform: uppercase` |
|--------|--------|
| image| image | 

By the way, the option itself with the text "1 of x" seems fine. This text is also used in the core:

https://github.com/WordPress/wordpress-develop/blob/e0ed30f1c802cda3c709b621746b420e300ecede/src/wp-admin/includes/class-wp-list-table.php#L1127

jameskoster:
If we delete that line then the text in the `select` doesn't match the label. Ideally the `select` and `label` use the uppercase styling, but the `option`s do not. I'm not sure if that's even possible though.

I'm trying another approach using the `suffix` prop:

Screenshot 2024-07-24 at 12 21 07

I think this might thread the needle, though there's a small bug; clicking the suffix doesn't open the `select` cc @mirka.

t-hamano:
> clicking the suffix doesn't open the "select"

The prefix and suffix are just elements that exist alongside the select element, but they _visually_ exist inside the select box, so this behavior seems like a bug to me too.

https://github.com/user-attachments/assets/ac270470-2dd1-410b-a0c5-c7e1eb0ef0c6

jameskoster:
If we can fix that (likely a separate endeavour) then I reckon this PR is in a decent spot in terms of design.

mirka (commented):
Before getting into the design specifics, I want to figure out the best way to deal with the labeling.

Using the `suffix` is not ideal because we can't allow [click through](https://github.com/WordPress/gutenberg/pull/63815#issuecomment-2247657496) to the `<select>` if the suffix has a dynamic width (it only works with the caret because the caret is fixed-width). Also `__( 'of %s pages' )` is not the best for localization because the word order in a phrase like "x of y pages" can invert the x and y depending on the language.

I'm wondering if [this kind of implementation](https://stackblitz.com/edit/vitejs-vite-wjd562?file=package.json,src%2FApp.tsx) would satisfy our requirements better and with a simpler implementation. Let me know if I'm missing some requirements.

t-hamano:
@mirka Thank you for your feedback.

I looked at your sample and your approach of showing "Page %s of %s" for the currently selected option and just showing the page number for the other options seems like a good idea.

It also works well for screen readers:


https://github.com/user-attachments/assets/f7cb4aca-7ce4-4e36-97ce-48c26054770f

jameskoster:
Yes I think that might work. But we'll still run into the styling issue where the size and capitalisation are applied to the `option`s.

mirka:
> But we'll still run into the styling issue where the size and capitalisation are applied to the `option`s.

@jameskoster Yes, I don't think there's a way around this. There are even browser differences — I noticed that Firefox doesn't uppercase it's options in the popover. If you're ok with the suggested implementation, I'll integrate it back into this PR.

jameskoster:
Yup, I think it's worth a try.

I'd like feedback from @WordPress/gutenberg-design on the styling aspect though. There are basically two options:

1. We style the select to match what's currently shipping (IE apply the label text styles). The drawback is that in some browsers this also applies to the options, which looks a little strange.
2. We use the vanilla component styles.

| Option 1 | Option 2 |
| --- | --- |
|  |  |

jameskoster:
Given the lack of feedback maybe we should merge this. It's easy to tweak the styling later. What do you think @mirka?

mirka:
Sure! Which styles are we going with for this iteration? I do recommend going with the vanilla component styles unless it's critical for it to be uppercased. (And nothing should be _that_ critical, given that uppercase doesn't exist in many languages 😄)

jasmussen:
Back from a break, so forgive me if I'm retreading territory. But this isn't the end of the world for me:

Screenshot 2024-07-31 at 15 56 48

mirka:
> Back from a break, so forgive me if I'm retreading territory. But this isn't the end of the world for me:
> 
> Screenshot 2024-07-31 at 15 56 48

I believe the strings being broken up like that was not ideal for accessibility and localization.

jasmussen:
Gotcha. It's mainly the discrepancy in this branch, where the select itself now holds the whole string, and only for the active item, that's feeling a bit awkward in practice.

mirka:
> the select itself now holds the whole string, and only for the active item, that's feeling a bit awkward in practice

If so, I'll throw out [another possibility](https://stackblitz.com/edit/vitejs-vite-ycfvu3?file=package.json,src%2FApp.tsx), which is to:

- Do the full "Page x of y" labeling as an aria-label. (Doesn't currently work, but we can fix it in SelectControl)
- Do the visual labeling as "/ 20 pages", which doesn't change order depending on language, so it's localization-friendly.

Alternative pagination

jasmussen:
Might work, could we reduce it to just:

```
Page 2▾ / 8
```

mirka:
Ok 👍 I'll go make the enhancement to `SelectControl` first so we can pass through the `aria-label`. I'll report back here when it's done.

mirka:
I switched over the implementation. I applied the most basic styling that doesn't require hacking around the component styles, but please tweak as you see fit 👍 

Simplified pagination select

t-hamano:
I added the one translation function that was missing, but it appears to be working as expected 👍

packages/dataviews/src/components/dataviews-pagination/index.tsx
```diff
@@ -37,41 +58,25 @@ function DataViewsPagination() {
...
+							onChangeView( {
+								...view,
+								page: +newValue,
+							} );
+						} }
+						size="compact"
+						__nextHasNoMarginBottom
+						variant="minimal"
+					/>
+					<div aria-hidden>/ { totalPages }</div>
```
t-hamano:
```suggestion
					<div aria-hidden>
						{ sprintf(
							// translators: Total number of pages
							__( '/ %s' ),
							totalPages
						) }
					</div>
```

Would it make sense to make this translatable?

jameskoster:
> which doesn't change order depending on language, so it's localization-friendly

Is this an issue on trunk? I'm just wondering if the default could be `**of** $total` with a translator suggestion to use `/ $total`. The slash approach is unconventional and looks a bit strange to me, to be honest.

mirka:
> I'm just wondering if the default could be `**of** $total` with a translator suggestion to use `/ $total`. The slash approach is unconventional and looks a bit strange to me, to be honest.

Good idea, I think that could work. Added in ac39701

Pagination with 'of n pages'

mirka:
> I added the one translation function that was missing

@t-hamano Thank you 🙈

packages/dataviews/src/components/dataviews-pagination/index.tsx
mirka: Applied @jameskoster's suggestion in ac39701

jameskoster:
Looking good, thanks @mirka. I think the last question is whether we want to apply the same text styles as trunk:

Screenshot 2024-08-06 at 17 57 42

Now that the `option`s don't include any words, I'm much less concerned about the styles being applied there:

Screenshot 2024-08-06 at 17 59 03

What do you think @jasmussen ?

packages/dataviews/src/components/dataviews-pagination/index.tsx
```diff
@@ -37,41 +58,32 @@ function DataViewsPagination() {
...
-					spacing={ 2 }
-					className="dataviews-pagination__page-selection"
+					spacing={ 1 }
+					className="dataviews-pagination__page-select"
 				>
-					{ createInterpolateElement(
-						sprintf(
-							// translators: %s: Total number of pages.
-							_x( 'Page <CurrentPageControl /> of %s', 'paging' ),
+					<div aria-hidden>{ __( 'Page' ) }</div>
```
ntsekouras: Why we removed `createInterpolateElement` here? It feels like the current approach for translation is not ideal.. I'm not 100% sure if the current approach is right, but I was wondering why it was needed? I'll cc @swissspidy if he has any suggestion.

swissspidy: Yeah I wouldn't remove it. Doing something like `__( 'of %s' )` makes this worse. `createInterpolateElement` is important here to avoid such string concatenation. 

I would even improve the translatable string a little bit like this:

```
sprintf(
	/* translators: 1: Current page, 2: Total pages. */
	_x( '%1$s of %2$s', 'paging' ),
	'<CurrentPageControl />',
	totalPages
)
```

jasmussen:
Works for me, though I'd refine that minimal dropdown a bit more. The space between the 1 and the down-chevron is pretty side at the moment, and since the 1 is included in the tap target, it doesn't need to be that wide. I don't know that's a blocker for this PR, perhaps it is? 

A quick stab:

status

That removes 5px left/right padding, and 5px gap between them. We can find a rounder number than 5 (maybe 4 or 6), but this was a quick photoshop just to share the thought.

t-hamano:
Another thing I noticed while testing this PR is that when `variant="minimal"` is enabled, the width of the select box changes depending on the length of the option text. This might be affected by the browser or OS:

https://github.com/user-attachments/assets/9906fdbe-e460-43ef-b035-2f3a860eb4de

This can also be reproduced in Storybook:

https://wordpress.github.io/gutenberg/?path=/story/components-selectcontrol--default&args=options[0].value:a;options[0].label:Option+AAAAAAAA;options[0].disabled:!undefined;options[1].value:b;options[1].label:Option+BBBB;options[2].value:c;options[2].label:Option+C;options[3]:!undefined;variant:minimal

https://github.com/user-attachments/assets/2b11acc6-273a-49c6-babe-e7d124122038

For example, in the Font Library Modal's Install Fonts tab, there are 200 pages of installable fonts, so if we apply the pagination style of this PR, the change in the select box width will be more noticeable.

jameskoster:
> though I'd refine that minimal dropdown a bit more

Agree, but let's make that change separately in the component.

@t-hamano the change in width is intentional, otherwise the chevron can appear far away from the selected value. This makes it harder to recognise the element as a select.

packages/dataviews/src/components/dataviews-pagination/index.tsx
mirka: My bad, fixed in bf1d7684eff41d0151ef640a0c14ee9db9db6fb1, 37fc33a.

mirka:
> the width of the select box changes depending on the length of the option text. This might be affected by the browser or OS

@t-hamano That's right, it's kind of a progressive enhancement at the moment. See the "Screenshots" section in the description of #63265.

jameskoster:
How are we looking here? :D

mirka (approved):
From a technical standpoint this is good to go 👍 As for styles, I'll defer to the design folks — please push whatever tweaks are deemed necessary before landing. Though I do recommend we figure out a style that works for _all_ instances of the `minimal` SelectControl variant, rather than ad hoc it just for this one. This pagination UI is one of the very few use cases for the `minimal` variant, and if it doesn't work here then it will probably not work in other places as well.

jameskoster:
I think the only necessary overrides are the font size / weight. It's not yet clear whether that's something we'll need at the component level. I pushed those changes and will update the OP.

There are also minor tweaks to consider, but I think we can do those separately.

packages/dataviews/src/components/dataviews-pagination/style.scss
```diff
@@ -11,11 +11,18 @@
 	@include reduce-motion("transition");
 }
 
-.dataviews-pagination__page-selection {
+.dataviews-pagination__page-select {
+	color: $gray-900;
```
t-hamano: If we remove this style, the color override appears not to be necessary:

https://github.com/WordPress/gutenberg/blob/c4a4cd1d3c17670a2f1b88b84c0c510f9a10ff83/packages/dataviews/src/components/dataviews-pagination/style.scss#L8

t-hamano (approved):
I think it's okay to ship this PR. From what I understand, I would expect this pagination style to apply to all other paginators below, but what about you?

- Font Library
- Global Styles Revision
- Pattern Modal
- Pattern Inserter

By the way, although it is not directly related to this PR, I found that the SelectControl component is affected by the line-height style of WP-Admin. Visually, it is a small difference, but I think this issue needs to be fixed separately:

Edit: I found that this issue occurs on the combination of Windows OS and Chrome. The issue might not occur on other OS or browsers.

image

image

packages/dataviews/src/components/dataviews-pagination/style.scss
jameskoster: Good spot!

jameskoster:
> I would expect this pagination style to apply to all other paginators below, but what about you?

Ideally, yes.

PR #63818, opened 2024-07-22
Fix: Error while Calling edit-site getCurrentTemplateTemplateParts selector

jorgefilipecosta:
Fixes an issue correctly pointed out by @Mamaduka at https://github.com/WordPress/gutenberg/pull/62033/files#r1673431541.
We removed the editor private selector getCurrentTemplateTemplateParts but "edit-site" had an equivalent public selector that just called the editor private selector. Removing made the public selector throw an error. Checking the plugin repository this selector has some usage so we can not remove it right now.

## Testing Instructions
On the browser console, I pasted `await wp.data.select('core/edit-site').getCurrentTemplateTemplateParts();`.
I verified the selector worked as expected and there was a deprecation warning on the console.


Mamaduka (approved):
Thanks for the fix, @jorgefilipecosta

A couple of high-level notes:

* It's now possible to memoize registry selectors (see #57888), which means we can drop the `getFilteredTemplatePartBlocks` helper.
* Let's restore logic for `getFilteredTemplatePartBlocks` in `edit-site` packages instead of re-adding a private selector. Colocation hopefully will prevent similar bugs in the future.

Edit: Sorry, miss-clicked the approval. Let's merge after these changes.

Mamaduka (approved):
Thanks for the follow-ups, @jorgefilipecosta!

I think this is good to be merged ✅

jorgefilipecosta:
Thank you for the reviews @Mamaduka!

PR #60100, opened 2024-03-22
Global styles: block background UI controls 

ramonjd:
## What?

Enables background images for supported blocks in global styles. 

https://github.com/WordPress/gutenberg/assets/6458278/c65a9deb-a1bb-41cd-8ebc-b9905f0735d9

Blocks include:

- https://github.com/WordPress/gutenberg/pull/62498
- https://github.com/WordPress/gutenberg/pull/62497
- https://github.com/WordPress/gutenberg/pull/62499

Part of:

- https://github.com/WordPress/gutenberg/issues/54336

Core backport PR:

- https://github.com/WordPress/wordpress-develop/pull/6836

## Why?

To enable background images across all supported blocks at once! MWHAHAHAHAHA!

## How?

Enabling background style properties as valid styles in `VALID_STYLES`, and not just valid `'top'` styles.

## Testing Instructions

To be updated...

In the Site Editor, open a template that contains a few ~Group~ Blocks.

Then head over to Styles > Blocks > ~Group~ and set a background image for all Group blocks. 

Check that all Group blocks now have that background image as default.

Ensure that the default background image is overridable at the block level. That is, select a Group block and set a different background image for that block.

Now define a background image at the block theme in theme.json.

Here's an example using some local files:

```json
{
	"$schema": "../../schemas/json/theme.json",
	"version": 3,
	"settings": {
		"appearanceTools": true
	},
	"styles": {
		"blocks": {
			"core/verse": {
				"background": {
					"backgroundImage": {
						"url": "file:./img/budgie4.jpg"
					}
				}
			},
			"core/quote": {
				"background": {
					"backgroundImage": {
						"url": "file:./img/budgie2.jpgg"
					}
				}
			},
			"core/pullquote": {
				"background": {
					"backgroundImage": {
						"url": "file:./img/budgie2.jpg"
					}
				}
			},
			"core/post-content": {
				"background": {
					"backgroundImage": {
						"url": "file:./img/budgie1.jpg"
					}
				}
			}
		}
	}
}

```

Check that this is displayed correctly in the frontend and editor.

PR #63551, opened 2024-07-15
Background Image: Make panel appear in a consistent location

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/63188

Issue was also flagged in https://github.com/WordPress/gutenberg/pull/61382#pullrequestreview-2168327052

## What?

This PR makes sure that the background image panel appears in the same location between blocks regardless of whether the block utilises the styles tab or has all panels within the block inspector itself.

The background image panel was also added to those displayed for multi-block selections when there are multiple blocks of the same type selected with support for background images.

## Why?

- Consistency in positioning will help discovery and overall UX
- Allows multiple blocks to be selected and have their background images changed at once for power users

## How?

- Move the background image panel to the same location as it was set in the Styles tab
- Add the background image panel to the multi selection block inspector

## Testing Instructions

1. Add multiple verse and group blocks to a post or page
2. Select a verse block and open the block inspector. Note the position of the background image panel
3. Select a group block and open the block inspector. The background image panel should be in the same location.
4. Select multiple verse blocks and confirm the background image panel appears in the inspector
5. Confirm that you can bulk assign background images when multiple verse blocks selected
6. Select a mix of verse and group blocks. The background image panel should not be in the inspector

### Verse Block

<details>
<summary>Before and After</summary>
| Before | After |
|---|---|
| Screenshot 2024-07-15 at 5 04 57 PM | Screenshot 2024-07-15 at 5 21 53 PM |

</details>

### Group Block ( No change )
<details>
<summary>Before and After</summary>

| Before | After |
|---|---|
| Screenshot 2024-07-15 at 5 05 11 PM | Screenshot 2024-07-15 at 5 08 32 PM |
</details>

### Multiple Selections
<details>
<summary>Before and After</summary>

| Before | After |
|---|---|
| Screenshot 2024-07-15 at 5 11 08 PM | Screenshot 2024-07-15 at 5 23 08 PM |
</details>

PR #63823, opened 2024-07-22
Block editor settings: add missing global styles links dependencies

ramonjd:
## What?

Follow up to:

- https://github.com/WordPress/gutenberg/pull/60100

Adds the `globalStylesLinksData` dependency to the useMemo dependency list.

Thanks to @youknowriad for catching it.

## Why?
The `globalStylesLinksData` dependency was missing from the useMemo dependency list.

## How?

Add the `globalStylesLinksData` dependency to the useMemo dependency list.

## Testing Instructions

Smoke test that background images with relative paths in theme.json appear as they should in the editor.

Here's some test JSON that can be used with TT4:

```json
{
	"version": 3,
	"settings": {
		"appearanceTools": true
	},
	"styles": {
		"background": {
				"backgroundImage": {
					"url": "file:./assets/images/art-gallery.webp"
				},
				"backgroundAttachment": "fixed"
		}
	}
}
```

In the Site Editor, check that:

1. The background image appears correctly in the canvas.
2. The image preview appears in the Global Styles controls, in both the round icon and also the focal point picker preview.

Screenshot 2024-07-23 at 7 39 36 AM

### DIFF
--- a/packages/editor/src/components/provider/use-block-editor-settings.js
+++ b/packages/editor/src/components/provider/use-block-editor-settings.js
@@ -353,6 +353,7 @@ function useBlockEditorSettings( settings, postType, postId, renderingMode ) {
 		setIsInserterOpened,
 		sectionRootClientId,
 		globalStylesData,
+		globalStylesLinksData,
 	] );
 }
 


packages/editor/src/components/provider/use-block-editor-settings.js
```diff
@@ -353,6 +353,7 @@ function useBlockEditorSettings( settings, postType, postId, renderingMode ) {
 		setIsInserterOpened,
 		sectionRootClientId,
 		globalStylesData,
+		globalStylesLinksData,
```
ramonjd, line 356: Discussion of whether this variable needs to be renamed, or alternatives to the delivery of global styles data to the block editor via settings: https://github.com/WordPress/gutenberg/pull/60100#discussion_r1687165325

aaronrobertshaw (approved):
LGTM 👍 

I smoke tested as per the test instructions and didn't spot any regressions. While the discussion on renaming or moving the `globalStylesLinksData` is ongoing, I think it makes sense to land this. It should mean it isn't missed when a rename does happen.

Screenshot 2024-07-23 at 4 21 49 PM

I also noticed that I missed updating the location of the background image control in https://github.com/WordPress/gutenberg/pull/63551. I can create a separate follow-up to explore whether it should live under the colors in the global styles sidebar as it does in the block inspector.

ramonjd:
> I can create a separate follow-up to explore whether it should live under the colors in the global styles sidebar as it does in the block inspector.

Sounds great, thanks for testing this one too.

aaronrobertshaw:
PR tweaking the location of the background panel is available in https://github.com/WordPress/gutenberg/pull/63888.

PR #63826, opened 2024-07-22
Remove unused useSplit after #54543

widoz:
## What?
The #54543 remove the `useSplit`  hook from the `edit.js`  but the hook file is still present.

## Why and How?
Cleanup the code base by removing unnecessary code


ellatrix (approved):
Nice, thank you!

Issue #61826, opened 2024-05-21
Filtering 'My patterns' category in the Inserter is a bit clunky


jameskoster:
<img width="936" alt="Screenshot 2024-05-21 at 16 37 28" src="https://github.com/WordPress/gutenberg/assets/846565/417e6593-403b-4def-b7cc-f0dc4abb6481">

---

1. My patterns will always be source=user(?); should that filter be hidden in this context?
2. Why are synced/not synced options disabled?

kevin940726:
> 1\. My patterns will always be source=user(?); should that filter be hidden in this context?

We can do that but I think we only kept them visible for consistency. Should we consider hiding it?

> 2\. Why are synced/not synced options disabled?

I believe this is a bug accidentally introduced in https://github.com/WordPress/gutenberg/pull/57570. I can file a PR fixing that.

jameskoster:
> We can do that but I think we only kept them visible for consistency. Should we consider hiding it?

Since the filter is presented local to the selected category, it seems superfluous to offer options can never be adjusted. Hiding is worth a try imo.

kevin940726:
Closed by #63828 and #63831.

PR #57570, opened 2024-01-05
Code Quality: Improve inserter pattern constants

talldan:
## What?
Across the codebase there were two constant objects called `PATTERN_TYPES`, confusingly containing different values. This PR changes the name of the one defined in `packages/block-editor/src/components/inserter/block-patterns-tab/utils.js` to `INSERTER_PATTERN_TYPES` to help indicate that it's used for inserter items.

I've also removed the 'synced' and 'unsynced' values from this constant which were unused.

I've removed the 'all' value from these constants, as this isn't a value that can be set on pattern inserter items. Instead this is a value that is used by the filter UI.

Finally, the `SYNC_TYPES` constant is renamed to `INSERTER_SYNC_TYPES`.

PR #63828, opened 2024-07-23
Fix user patterns disabling sync filter

kevin940726:
## What?

Fix part of https://github.com/WordPress/gutenberg/issues/61826.

The sync filter for user patterns is not enabled in the inserter.

## Why?

This seems like an accidental mistake introduced in https://github.com/WordPress/gutenberg/pull/57570.

## How?

Add back the condition for user patterns.

## Testing Instructions

1. Open the editor.
2. Open the global inserter and navigate to "Patterns", then "My patterns".
3. Click "Filter patterns" on the top right corner.
4. Expect the sync status filter "type" to be enabled.

## Screenshots or screencast 

**Before**
image

**After**
image

### DIFF
--- a/packages/block-editor/src/components/inserter/block-patterns-tab/patterns-filter.js
+++ b/packages/block-editor/src/components/inserter/block-patterns-tab/patterns-filter.js
@@ -22,7 +22,8 @@ import {
 	INSERTER_PATTERN_TYPES,
 } from './utils';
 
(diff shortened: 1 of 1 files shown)


talldan (approved):
Thanks for fixing. Not sure how that happened, looks like my intention was to move it below the imports but somehow the `&&` part of the clause was stripped.

jameskoster (approved):
Working now, thanks for the fix :)

PR #63831, opened 2024-07-23
Hide source filter in my patterns

kevin940726:
## What?

Resolve part of https://github.com/WordPress/gutenberg/issues/61826. Hide the source filter if the category is "My patterns".

## Why?

See https://github.com/WordPress/gutenberg/issues/61826#issuecomment-2242823306. The source filter will only have "User" available if the category is "My patterns".

## How?

Hide the source filter if the category is "My patterns".

## Testing Instructions

1. Open a post or page.
2. Open the inserter. Navigate to "Patterns", "My patterns".
3. Click on the "Filter patterns" button on the top-right corner.
4. Expect only a "Type" filter, no "Source" filter.

## Screenshots or screencast 

**Before**
image

**After**
image

### DIFF
--- a/packages/block-editor/src/components/inserter/block-patterns-tab/patterns-filter.js
+++ b/packages/block-editor/src/components/inserter/block-patterns-tab/patterns-filter.js
@@ -24,9 +24,28 @@ import {
 
 const getShouldDisableSyncFilter = ( sourceFilter ) =>
 	sourceFilter !== 'all' && sourceFilter !== 'user';
-const getShouldDisableNonUserSources = ( category ) => {
+const getShouldHideNonUserSources = ( category ) => {
 	return category.name === myPatternsCategory.name;
 };
+
+const PATTERN_SOURCE_MENU_OPTIONS = [
+	{
+		value: 'all',
+		label: _x( 'All', 'patterns' ),
+	},
+	{
+		value: INSERTER_PATTERN_TYPES.directory,
+		label: __( 'Pattern Directory' ),
+	},
+	{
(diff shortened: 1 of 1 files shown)


packages/block-editor/src/components/inserter/block-patterns-tab/patterns-filter.js
```diff
@@ -50,10 +69,9 @@ export function PatternsFilter( {
 		currentPatternSourceFilter
 	);
 
-	// We also need to disable the directory and theme source filter options if the category
-	// is `myPatterns` otherwise applying them will also just result in no patterns being shown.
-	const shouldDisableNonUserSources =
-		getShouldDisableNonUserSources( category );
+	// We also hide the directory and theme source filter if the category is `myPatterns`
+	// otherwise there will only be one option available.
+	const shouldHideSourcesFilter = getShouldHideNonUserSources( category );
```
talldan: Nit: The function name makes it sound like it only hides some of the options

```suggestion
	const shouldHideSourcesFilter = getShouldHideSourcesFilter( category );
```

talldan (approved):
LGTM, this cleans up the code quite a bit too which is a bonus

jameskoster (approved):
The patterns in this category will always be source = user, so this change makes sense to me.

packages/block-editor/src/components/inserter/block-patterns-tab/patterns-filter.js
kevin940726: Oops I missed that 😆

PR #63832, opened 2024-07-23
Tag Cloud: Prevent duplicate spacing in editor

### LINKED ISSUE #63716: Remove excess <p> from Tag Cloud edit function
In the edit function of the Tag Cloud block, there is a secondary `<p>` within the block, around the tags. This does not occur on the front-end. The extra `<p>` within the block is causing additional default browser margin to be applied within the block only within the editor. The front-end looks as expected. 

I propose the edit function is updated to accurately reflect the front end, which will bring these closer to 1:1.  

### Editor

CleanShot 2024-07-18 at 12 39 49

### Frontend
CleanShot 2024-07-18 at 12 41 49

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/63716

## What?

Avoids the duplicate application of default padding and margin styles on the Tag Cloud block in the editor due to its use of server-side rendering.

## Why?

Without these changes, default margin and padding are applied twice creating a visual difference between the editor and frontend.

## How?

- Reset padding and margin on the inner element with `.wp-block-tag-cloud` class

#### Alternate Approach

Initially, it was suggested to reset the spacing styles on the outer wrapping element in the editor. The catch here is that there is default padding for different alignments and the CSS classes for those alignments are only applied to the outer wrapper. 

If the spacing style resets occur on the outer wrapper, these defaults need to be plumbed through to the inner `p.wp-block-tag-cloud` element. The current approach in this PR minimises any additional CSS to reset spacing.

## Testing Instructions

1. Make sure you have some posts with tags assigned to them
2. Activate TT4 and add some global spacing styles for the Tag Cloud block via Global Styles or theme.json
3. Add a Tag Cloud block in the editor
4. Inspect the inner `p.wp-block-tag-cloud` element and ensure that padding and margin styles are reset
5. Compare the visual layout of the block in the editor and the frontend, there should be no duplicate spacing in the editor
6. Clear global styles and theme.json customizations
7. Select the Tag Cloud block and toggle on different alignments
8. Ensure the outer wrapper still gets the default padding values and none are applied to the inner element

## Screenshots or screencast 

| Before | After | Frontend |
|---|---|---|
| Screenshot 2024-07-23 at 1 06 36 PM | Screenshot 2024-07-23 at 2 13 55 PM | Screenshot 2024-07-23 at 2 14 18 PM |

### DIFF
--- a/packages/block-library/src/editor.scss
+++ b/packages/block-library/src/editor.scss
@@ -44,6 +44,7 @@
 @import "./social-links/editor.scss";
 @import "./spacer/editor.scss";
 @import "./table/editor.scss";
+@import "./tag-cloud/editor.scss";
 @import "./template-part/editor.scss";
 @import "./text-columns/editor.scss";
 @import "./video/editor.scss";
--- a/packages/block-library/src/tag-cloud/editor.scss
+++ b/packages/block-library/src/tag-cloud/editor.scss
@@ -1,0 +1,9 @@
+// The following styles are to prevent duplicate spacing for the tag cloud
+// block in the editor given it uses server side rendering. The specificity
+// must be higher than `0-1-0` to override global styles. Targeting the
+// inner use of the .wp-block-tag-cloud class should minimize impact on
+// other 3rd party styles targeting the block.
+.wp-block-tag-cloud .wp-block-tag-cloud {
+	margin: 0;
+	padding: 0;
+}


ramonjd (approved):
LGTM in the editor

| Trunk  | This branch |
| Screenshot 2024-07-23 at 6 02 36 PM  | Screenshot 2024-07-23 at 6 02 03 PM  |

The new rule reset the margin and padding from the inner `.wp-block-tag-cloud`

Frontend looking as I'd expect

Screenshot 2024-07-23 at 6 02 36 PM

I guess the catch is we'll have to be vigilant should the block receive further block supports, e.g., border

Screenshot 2024-07-23 at 6 06 36 PM

aaronrobertshaw:
Thanks for the thorough review @ramonjd 

> I guess the catch is we'll have to be vigilant should the block receive further block supports, e.g., border

For block supports themselves, the Tag Cloud block [should already omit those block support attributes](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/tag-cloud/edit.js?plain=1#L177) and classes. Anything can be styled via theme.json and global styles, so I agree this isn't a complete solution. 

In my view, the only complete solution is to refactor the block to use the REST API instead of server-side rendering.

If we go the route of resetting all styles, we'd need to do that for anything that could be styled via theme.json. That includes any custom CSS the theme or user adds. I'm not sure that is feasible in the short term either.

For now, this PR addresses the most glaring issue, buying time to see the impact of what remains and/or refactor the block.

PR #63833, opened 2024-07-23
Inserter: Allow searching for variations via the `block.id` instead of the `block.name`

### LINKED ISSUE #61982: Group block variations: Unable to add Row, Stock, Grid by '/(English label)' shortcut in Japanese environment.
### Description

In Japanese environment, Group Row variation cannot be added by `/row` nor `/group-row`, though It can be added by `/横並び` (Japanese label). Stock and Grid are the same.

Group block can be added by both `/group` and `/グループ` (Japanese label).

1. Select '日本語' (= Japanese) for Site Language in Settings General to switch to Japanese environment.
2. Create a new page.
3. Type `/row`. Grid Row variation should be in the popup window, but not.
4. Type `/group`. You'll see Group, Row, Stack and Grid in the popup window.
5. Continue to type as `/group-row`. Nothing will be shown.
6. Type `/横並び`. Then you'll see Group Row variation.

_No response_

### Environment info

WordPress 6.5.3
Gutenberg 18.4.1
Twenty Twenty-Four
Google Chrome 124.0.6367.203
Desktop with Windows 10 22H2 19045.4412

Yes

Yes

### LINKED ISSUE #63802: Grid blocks cannot be called with English block names using '/' in Japanese environment
### Description

In the English environment, you can successfully call a grid block by typing `/grid` when calling the block in the editor.

However, in the Japanese environment, typing `/grid` in the alphabet does not call it. By the way, it seems to be able to be called as a group block, and you can call a group block and its related blocks by typing `/group`. In short, we can confirm that `/grid` does not respond to `/group`, but `/group` does.

We also confirmed this as a related issue. It seems that `/stack` and `/row` do not respond in Japanese as well.

https://github.com/WordPress/gutenberg/issues/61982

In the English environment, we have already confirmed that the grid block can be called with `/grid` without any problem.

1. set WordPress language to Japanese
2. when I create a new page or post and type `/grid`, the target block does not appear in the popup
3. when I type `/グリッド` in Japanese, it appears in the popup

## Tested in English Env.

https://github.com/user-attachments/assets/730a772f-9621-4113-a175-a0513b92bd28

## Tested in Japanese Env.

https://github.com/user-attachments/assets/00bc6ddc-a57d-4562-b620-ba473ac481b3

### Environment info

- WordPress 6.6
- Set language in Japanese
- Activated Gutenberg 18.8.0

- [X] Yes

- [X] Yes

talldan:
## What?
Fixes #63802
Fixes #61982

## Why?
The block search feature has used the `block.name` (e.g. `core/paragraph`) for a while as part of the search criteria. It seems this was never updated when variations where introduced so as mentioned in #63802 and #61982 search doesn't work as well for variations as it does for regular blocks.

## How?
In the inserter block types there's a `block.id` which is very similar to the name, but includes the variation name (e.g. `core/group/group-grid`). The fix is to use the `block.id` as a preference over the `block.name` in the search criteria. For non-variations, the `id` is the same as `name`, so it won't change anything.

Even if what was described in the linked issues was never intended as a feature, this will provide a better chance of matching block variations in the search, while also providing an easy way to solve those issues.

## Testing Instructions
1. Switch to Japanese site language
2. Type `/gri` in an empty paragraph and check the 'Grid' block is returned in the search results.

## Screenshots or screencast 
Screenshot 2024-07-23 at 1 10 25 PM

### DIFF
--- a/packages/block-editor/src/components/inserter/search-items.js
+++ b/packages/block-editor/src/components/inserter/search-items.js
@@ -5,7 +5,9 @@ import removeAccents from 'remove-accents';
 import { noCase } from 'change-case';
 
 // Default search helpers.
-const defaultGetName = ( item ) => item.name || '';
+// For the block name, prioritize returning the id,
+// which is similar to the name, but includes the variation.
+const defaultGetName = ( item ) => item.id || item.name || '';
 const defaultGetTitle = ( item ) => item.title;
 const defaultGetDescription = ( item ) => item.description || '';
 const defaultGetKeywords = ( item ) => item.keywords || [];


t-hamano (commented):
When searching for a pattern from the main inserter, it seems that the following critical error occurs because `item.id` is a number:

```
search-items.js:215 Uncaught TypeError: name.startsWith is not a function
```

Furthermore, I think this improvement only needs to be applied to blocks, so how about overriding `defaultGetName` here?

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/inserter/search-items.js#L111

There may be a better way to write it, but I think it would probably look something like this:

```javascript
const config = {
	getName: ( { id, name } ) => {
		if ( id ) {
			return id.toString();
		}
		return name || '';
	},
	// ...
};
```

talldan:
> When searching for a pattern from the main inserter, it seems that the following critical error occurs because item.id is a number

It seems problematic that inserter items have different data, yet still use the same search/ranking functions.

Here's a quick inserter item:
```js
{
    "id": "core/block/158",
    "name": "core/block",
    // ...
}
```

And here's the same block in the main inserter:
```js
{
    "name": "core/block/158",
    "id": 158,
    // ...
}
```

So this means that the two end up working differently when searching. I think that's an underlying problem that could be fixed. I'm not sure which one is more correct though 😄

t-hamano:
> I think that's an underlying problem that could be fixed.

I looked into this problem and it seems to be due to a difference in whether the patterns are retrieved by the `useBlockTypesState` hook or the `usePatternsState` hook.

For example, in the main inserter and quick inserter, patterns are retrieved via the `usePatternsState` hook ([source](https://github.com/WordPress/gutenberg/blob/694e6650e49a07191b63de766e3824be954cdc5a/packages/block-editor/src/components/inserter/search-results.js#L85-L88)).

On the other hand, in the autocompleter, patterns are retrieved via the `useBlockTypesState` hook rather than the `usePatternsState` hook ([source](https://github.com/WordPress/gutenberg/blob/694e6650e49a07191b63de766e3824be954cdc5a/packages/block-editor/src/autocompleters/block.js#L61-L63)).

My guess is that the autocompleter should also use the `usePatternsState` hook for patterns, but I'm not sure 😅

Issue #63683, opened 2024-07-17
Give posts (any type) higher priority in link search results


richtabor:
In LinkControl you can search for content existing on your site. This is great, but I did find that attachments were surfaced higher in search results than posts matching the search requirements. 

I propose that pages and posts of all type are prioritized in the search results, above all others. It's much more likely to link to pages and posts, than to attachments. 

### Visual 

![Maintain a consistent layout and style across patterns while](https://github.com/user-attachments/assets/695406ef-32af-414a-bd5c-d9c7b668b355)

noisysocks:
> I propose that pages and posts of all type are prioritized in the search results, above all others.

Take this too literally and we'll cause a regression of https://github.com/WordPress/gutenberg/issues/56478 😀 I think we want to prioritise posts and pages but not always place them above every other result. It's important that users can easily link to tags and categories especially from the Navigation block.

richtabor:
I'd say you are likely to search for and link a page many times more likely than a tag or category archive. 

I would say given a page, tag, and attachment matched with `patterns`, the page should be first in that list. 

Perhaps deprioritizing attachments would meet the expectations better?

noisysocks:
Yeah in your example screenshot I'd expect _Composing with patterns_ to be first but if you searched for "patterns-1" I'd expect _patterns-1_ to appear first.

The key thing to bear in mind is that we don't want to regress https://github.com/WordPress/gutenberg/issues/56478 as that bug made creating some types of navigation basically impossible.

I think giving posts a slight (25%? need to play with the exact number) boost and attachments a slight penalty (25%?) should work.

richtabor:
I'm down for trying that.

fabiankaegy:
After updating to WordPress 6.7 this is something we've heard several client teams complain about. They are having a much harder time finding the actual relevant content than they did before the update which is impacting their workflows. 

In all honesty I would love an option to remove attachments from the results altogether. On most sites that simply isn't something that editors need to do. And if so they can add the link manually 🤔

The exploration of giving less prominence to attachments sounds like a good first step though  👍 

I'm also going to add the `[Type] Regression` label to this since the search here used to be much better than it is now in 6.7

SainathPoojary:
Hey @noisysocks,

I explored implementing the 25% boost for post types and the 25% penalty for attachments. Here are my findings:

Initially, I directly applied the adjustments like this:

```js
// Boost for post types, penalty for attachments
if (result.kind === 'post-type') {
[...]

PR #63836, opened 2024-07-23
[Type] Enhancement : post higher priority in link search results

kartikmehta8:
## What?

This PR updates the `sortResults` function in the link suggestions feature to prioritize pages and posts above attachments in search results.

## Why?

Attachments were being surfaced higher in search results than posts matching the search requirements. This change ensures that pages and posts are prioritized in the search results, which is more likely to be the desired outcome for users.

## How?

The `sortResults` function has been updated to add a score increment for results that are of `kind` `post-type` and `type` `post` or `page`. This adjustment ensures that these types of results are prioritized above others, such as attachments.

## Testing Instructions

1. Open a post or page.
2. Use the link control to search for content.
3. Observe that pages and posts are prioritized in the search results over attachments.

1. Open a post or page using keyboard navigation.
2. Use the link control to search for content using keyboard navigation.
3. Ensure that pages and posts are prioritized in the search results over attachments.

### Issue
- https://github.com/WordPress/gutenberg/issues/63683

## Screenshots or screencast

N/A

### DIFF
--- a/packages/core-data/src/fetch/__experimental-fetch-link-suggestions.ts
+++ b/packages/core-data/src/fetch/__experimental-fetch-link-suggestions.ts
@@ -268,6 +268,7 @@ export function sortResults( results: SearchResult[], search: string ) {
 
 	const scores = {};
 	for ( const result of results ) {
+		let score = 0;
 		if ( result.title ) {
 			const titleTokens = tokenize( result.title );
 			const matchingTokens = titleTokens.filter( ( titleToken ) =>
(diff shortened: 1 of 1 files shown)


talldan:
I think @noisysocks worked on this recently, so might have some thoughts. Hope you don't mind that I added you as a reviewer, Rob!

noisysocks:
Hey thanks for the PR! 

I haven't tested it but I suspect this approach will cause a regression of https://github.com/WordPress/gutenberg/issues/56478 if there are more than 20 posts or pages on the site containing the search term. This is because all twenty of those posts and pages will receive a score of `1` (the base score this PR adds) regardless of their relevancy and crowd out any categories and tags even if they are an exact match (`score = 1`).

We want to prioritise posts and pages but not so much that they crowd out attachments, tags, and categories that are very relevant to the search query.

How did you arrive at 1 as a base score for posts and pages? I expect the solution will be to change this to 0.5 or thereabouts. The exact number will need to be dialled in from testing with real data. The testing steps in https://github.com/WordPress/gutenberg/pull/62397 describe how you can set yourself up with testing data that is good for working on this feature.

coenve:
Ever since Wordpress added the score element to the link suggestions, my editors have been complaining about tags and attachments coming up first instead of relevant blogs they would like to link to. They find it very annoying that the sort order was changed. They asked me if I could undo that change. 

Since I don't want to mess around in the source code, is there a filter I can use to manually override the search order? 

My opinion on this case is:
Only show tags, categories, and attachments that match the keyword for 100%. If it's 99% don't show it. Go for posts and pages first. 

We run a technology blog with tens of thousands of blogs. For example we write a lot about Microsoft, we have a few dozen Microsoft tags. If an author now wants to link to a recent Microsoft blog it's hard, because there are too many tags that start with Microsoft.

PR #63846, opened 2024-07-23
Try/multi selection touch

ellatrix:
## What?

Testing...


ellatrix:
I think the approach here is a bit too crazy. We're basically always ensuring blocks retain their focus, even though the browser will continuously reset it to writing flow. The approach in #63671 will leave focus on writing flow, and is a bit more conservative by only turning in content editable when there's an uncollapsed selection. We can later see if we can expand that.

PR #50030, opened 2023-04-24
Add Theme Previews for block themes

scruffian:
## What?
This adds theme previews to the Site Editor by means of a `theme_preview` parameter. The aim is to enable theme previews from the theme directory as outlined in https://github.com/WordPress/gutenberg/issues/37201#issuecomment-1516149003.

### <samp>🤖 Generated by Copilot at 3f68df7</samp>

This pull request adds a new feature to the `apiFetch` and `core-data` packages that enables theme previews in the Site Editor for block themes. It also adds a compatibility layer for WordPress 6.3 that filters the blog options and sets the theme for the preview based on the REST API request. It creates and requires a new file `theme-previews.php` and a new middleware function `createThemePreviewMiddleware`.

## Why?
Theme previews are a missing feature for block themes.

## How?
1. Add filters to the the `stylesheet` and `template` options when the `theme_preview` URL parameter is set, to modify the blog option for to a different theme.
2. Also add a middleware to the REST API to add the theme_preview parameter so that the blog options are changed in the REST API.

### <samp>🤖 Generated by Copilot at 3f68df7</samp>

*  Add functions to enable theme previews in the Site Editor for block themes ([link](https://github.com/WordPress/gutenberg/pull/50030/files?diff=unified&w=0#diff-6a8af5b29ddff1cc2863aa583b551dfe14f73fb49ffb4e79782c7477b2d7680fR1-R50))
* Require the `theme-previews.php` file in the `load.php` file ([link](https://github.com/WordPress/gutenberg/pull/50030/files?diff=unified&w=0#diff-979d356a31350e9429b6faa5be6daaa5b10785838520396f552717a1b270c67bR55))
* Create and expose a `createThemePreviewMiddleware` function in the `api-fetch` package ([link](https://github.com/WordPress/gutenberg/pull/50030/files?diff=unified&w=0#diff-f6baf3c8c169254cd1f8743f6f2fc25665320f2c768d7a6da71b3e27709940f8R17),[link](https://github.com/WordPress/gutenberg/pull/50030/files?diff=unified&w=0#diff-f6baf3c8c169254cd1f8743f6f2fc25665320f2c768d7a6da71b3e27709940f8R197),[link](https://github.com/WordPress/gutenberg/pull/50030/files?diff=unified&w=0#diff-6c257e425d44e1f7ed485fe903a5db14b2427fb98225e86c478d01e21d0816bcR1-R35))
[...]

PR #63849, opened 2024-07-23
Extract utilities to `@wordpress/editor` and `@wordpress/router`

oandregal:
## What?

This PR does two things:

- Updates the duplicate action to take the the post fields from a shared place instead of duplicating it.
- Updates the view action to respect the `wp_theme_preview` parameter.

## How?

- Moves the fields declaration to the `@wordpress/editor` package, so both edit-post (actions) and edit-site (quick edit) packages have access to it.
- Moves some routing utilities to the `@wordpress/router` package, so both edit-post (actions) and edit-site (quick edit) packages have access to it.

## Testing Instructions

Test the view action:

- Visit the Site Editor's Pages page.
- Add the `wp_theme_theme=twentytwentytwo` URL parameter and reload. This should load TT2 in the preview.
- Click on the `View` action of any page, to load the front-end.
- The expected result is that the front-end loads the TT2 theme (in `trunk` the `wp_theme_preview` parameter is not used by the view action).

Test the duplicate action:

- It should still work as before, nothing changed.


packages/editor/package.json
```diff
@@ -62,6 +62,7 @@
 		"@wordpress/private-apis": "file:../private-apis",
 		"@wordpress/reusable-blocks": "file:../reusable-blocks",
 		"@wordpress/rich-text": "file:../rich-text",
+		"@wordpress/router": "file:../router",
```
oandregal, line 65: This is the major concern I have to land this PR.

By doing this, the `@wordpress/editor` package would depend on some utilities provided by the router that require a `<RouterProvider>` in the component tree. Is this something the editor package cares about? Or is it the higher-up edit-site/edit-post/etc. responsibility instead? @youknowriad I presume you have thoughts about this and can share advice.

packages/edit-site/src/components/editor/index.js
```diff
@@ -48,7 +47,8 @@ import useEditorTitle from './use-editor-title';
 import { useIsSiteEditorLoading } from '../layout/hooks';
 
 const { Editor, BackButton } = unlock( editorPrivateApis );
-const { useHistory, useLocation } = unlock( routerPrivateApis );
+const { useHistory, useLocation, isPreviewingTheme } =
```
youknowriad, line 50: Router is a generic package that is meant to implement "routing" in any context. `isPreviewingTheme` is a site editor specific state, I'm not sure I understand this change.

packages/editor/src/components/post-actions/actions.js
```diff
@@ -483,7 +476,13 @@ const viewPostAction = {
 	},
 	callback( posts, { onActionPerformed } ) {
 		const post = posts[ 0 ];
-		window.open( post.link, '_blank' );
+		let url = post.link;
+		if ( isPreviewingTheme() ) {
```
youknowriad, line 480: Is this action, the reason you had to move previewing theme state to router?

packages/router/src/private-apis.js
```diff
@@ -2,11 +2,20 @@
...
 
 export const privateApis = {};
 lock( privateApis, {
 	useHistory,
 	useLocation,
 	RouterProvider,
+	isPreviewingTheme,
+	currentlyPreviewingTheme,
+	Link,
+	useLink,
```
youknowriad, line 20: I would be ok with moving Link and useLink to the router package if needed but only if we can remove the dependency to isPreviewingTheme and currentlyPreviewingTheme

packages/editor/package.json
youknowriad: This is a bit of a gray area.

If I understand properly, you want to move fields to "editor" (or core-data) or something like that. That's totally fine by me assuming all the fields are independent of the place they are used in (post editor, site editor...). 

The problem is that the urls them selves are specific, in the site editor you want to navigate to some places but not in the post editor... Maybe for a start we can use always the same url and it should work regardless of where you are.

The other problem is that I'm not sure "urls" make sense in fields at all. I think for instance rendering "Media" or "Title" field or something shouldn't actually render an application url/link automatically. That feels like a separate config that should be passed to `DataViews`. For instance a `getItemLink` just like we have `getItemId`.

Also nothing that we already have something like `getItemLabel` but it's not implemented properly, right now we guess it by fetching the "primary" field which IMO is mixing separate concerns.

packages/router/src/link.js
```diff
@@ -33,6 +30,9 @@ export function useLink( params, state, shouldReplace = false ) {
 		...Object.keys( currentArgs )
 	);
 
+	// TODO:
```
oandregal, line 33: The existing theme preview functionality has been introduced at https://github.com/WordPress/gutenberg/pull/50030 and works this way: you can add a `wp_theme_preview=THEME` URL parameter and the new theme will be previewed. It works in the site editor, post editor, and front-end.

To make sure all links respect the existing `wp_theme_preview` parameter, navigation & links should use `useLink` or `Link` utilities. Some thoughts:

- We need to account for the `wp_theme_preview` parameter in more places than the site editor. For example, post actions, such as "View" (takes you to front-end) should respect that parameter. These actions are defined in the `@wordpress/editor` package so they can be shared between edit-post & edit-site. How do we go about making this work? This PR is a potential approach, but I'd like to double-check if it's sensible to go this way ([see](https://github.com/WordPress/gutenberg/pull/63849/files#r1687853902)).

- @jsnajdr is it problematic that the params maintain the `wp_theme_preview` but it's not present in the history? Haven't been able to dig down into the router/history behavior.

packages/editor/src/components/post-fields/index.js
```diff
@@ -25,19 +25,24 @@ import {
...
  */
-import {
-	LAYOUT_GRID,
-	LAYOUT_TABLE,
-	LAYOUT_LIST,
-	OPERATOR_IS_ANY,
-} from '../../utils/constants';
-import { default as Link, useLink } from '../routes/link';
-import Media from '../media';
+import { unlock } from '../../lock-unlock';
```
oandregal, line 38: This is the gist of this PR. To be able to move post-fields from edit-site to the editor package, we need to remove any edit-site dependency:

- DataViews-related constants: make them inline.
- The `Media` component: make it inline, as it was only used by post-fields.
- Link, useLink: move to `@wordpress/router` package.

packages/editor/src/components/post-actions/actions.js
oandregal, line 480: See conversation at https://github.com/WordPress/gutenberg/pull/63849/files#r1687870181

oandregal: Also because the `useLink`/`Link` components have them as a dependency. If I move those to the `@wordpress/router`, it cannot depend on `@wordpress/edit-site` (where `isPreviewingTheme`/etc. would live). See what I know about the `wp_theme_preview` parameter at https://github.com/WordPress/gutenberg/pull/63849#discussion_r1687870181

packages/router/src/private-apis.js
oandregal: Yeah, I had the same feeling.

One way to do that is to make any consumer of `useLink`/`Link` to check for the presence of the `wp_theme_preview` parameter, but that leaks that feature everywhere and every page/consumer has to be aware of that. See what I know about theme previews at https://github.com/WordPress/gutenberg/pull/63849#discussion_r1687870181 Happy to hear alternatives.

packages/edit-site/src/components/editor/index.js
oandregal: Based on what I know about https://github.com/WordPress/gutenberg/pull/63849#discussion_r1687870181 perhaps we should not consider `isPreviewingTheme` a site-editor concern, but also a front & post editor concern.

youknowriad: I'll reply in the other comment then.

packages/router/src/link.js
youknowriad: The initial goal of `@wordpress/router` was to be app agnostic and adding the "theme preview" logic into it breaks that. 

Could we instead have a way to inject this logic as a "middleware" or something when rendering `RouterProvider`. In other words a prop to `RouterProvider` that changes how links work a bit (apply the them preview logic)

packages/editor/package.json
oandregal: > If I understand properly, you want to move fields to "editor" (or core-data) or something like that.

Yeah. One of the issues I want to fix is: how do we share the field definition? Fields need to be available in three places now: edit-site/post-edit, edit-site/post-list, editor/post-actions.

packages/router/src/private-apis.js
youknowriad: I shared an idea in the other thread https://github.com/WordPress/gutenberg/pull/63849#discussion_r1687887936

packages/editor/src/components/post-actions/actions.js
youknowriad: Maybe in addition to the idea I shared here https://github.com/WordPress/gutenberg/pull/63849#discussion_r1687887936 the site editor can override this action.

packages/editor/package.json
youknowriad: Yes, just like the actions. 

My goal for the actions was to move them one by one (and type them) into `editor/src/dataviews/actions` and I can see the same for fields in `editor/src/dataviews/fields`.

I think ultimately, it makes sense for `editor/src/dataviews` to be its own package: `core-dataviews` or something or actually be part of `core-data`

Issue #59409, opened 2024-02-27
List view: the 'sticky' information is only visual


afercia:
### Description

In the List view, some icons may be shown to provide information about some block properties, for example:
- when the block is locked.
- when the block is sticky

Screenshot:

![Screenshot 2024-02-26 at 14 34 26](https://github.com/WordPress/gutenberg/assets/1682452/fe2b7f34-5e45-49ec-a115-8f1bdd5f938d)

While for the 'locked' state a text `(locked)` is appended to the block item aria-label, there's no textual information for the 'sticky' case.

As such, there's no parity between the visual information provided to sighted users and the textual information provided to users with no vision or low vision or users who don't recognize icons meaning.

Ideally, this kind of important information should be provided by the means of _visible_ text.

At the very least, the 'sticky' state should be communicated by the means of some visually hidden text or description

### Step-by-step reproduction instructions

- Go to teh post editor.
- Add a group block with a paragraph.
- First, make the group block sticky.
- Then, make the group block locked.
- Open the list view.
- Observe the group block item has two icons: sticky and lock.
- Inspect the source.
- Oserve the aria-label of the group block is: `Group (locked)`.
- Onserv there's no textual information of any type about the sticky state.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

Yes

t-hamano:
I think that in addition to the sticky information, the anchor text also needs to be conveyed to the screen reader. I'm at a loss as to what is the appropriate way to convey this when these things are combined...

![image](https://github.com/WordPress/gutenberg/assets/54422211/58311709-3781-4f78-9220-b7e88fcf1966)

afercia:
Thanks @t-hamano. Yes I wanted to investigate other potential icon / things that may be rendered within the list view items. The `anchor` is a good one. I also want to investigate the image previews.
Any more potential things that may be shown there you can think of?

t-hamano:
Looking at [the List View implementation](https://github.com/WordPress/gutenberg/blob/0b5cd48bfb474377e6947a7c3302d37bd910a586/packages/block-editor/src/components/list-view/block-select-button.js#L271-L307), there are four pieces of information that may be displayed in addition to the block name:

- Anchor text
- Sticky icon
- Image thumbnails
- Locked icon

[...]

PR #63850, opened 2024-07-23
List View: remove the sticky position icon tooltip

ciampo:
## What?

Related to #59409

Removes the tooltip surrounding the sticky position icon in the list view.

## Why?

- Tooltips should not be rendered inside interactive content (list view items are anchor tags)
- This was already causing `Tooltip` not to work as expected (it currently doesn't render on `trunk`)
- It also makes the list view more consistent, since the sticky position icon was the only icon with an associated tooltip

## How?

- Removed `Tooltip` and replaced it with a `span`, like it's currently done with other icons in the list view
- Moved the "position" description text to the accessible description of the list view item

> [!NOTE]
> Adding the "position" text to the list item's description text is optional and can be removed if we don't think it represents an improvement (see related [comment](https://github.com/WordPress/gutenberg/issues/59409#issuecomment-1979938660)).

## Testing Instructions

- Go to the post editor.
- Add a group block with a paragraph.
- First, make the group block sticky.
- Then, make the group block locked.
- Open the list view.
- Observe that the group block item has two icons: sticky and lock (as on `trunk`)
- Inspect the source.
- Observe that the group block item's descriptive text mentions both the sticky position and the locked state.
- Play with different combinations of locked/unlocked and sticky/normal position, make sure that icons and description text update accordingly.

## Screenshots or screencast 

Screenshot 2024-07-23 at 14 08 01

### DIFF
--- a/packages/block-editor/src/components/list-view/utils.js
+++ b/packages/block-editor/src/components/list-view/utils.js
@@ -13,8 +13,19 @@ export const getBlockPositionDescription = ( position, siblingCount, level ) =>
 		level
 	);
 
-export const getBlockPropertiesDescription = ( isLocked ) =>
-	isLocked ? __( 'This block is locked.' ) : '';
+export const getBlockPropertiesDescription = ( blockInformation, isLocked ) =>
+	[
+		blockInformation?.positionLabel
+			? `${ sprintf(
+					// translators: 1: Position of selected block, e.g. "Sticky" or "Fixed".
+					__( 'Position: %1$s' ),
+					blockInformation.positionLabel
+			  ) }.`
+			: undefined,
+		isLocked ? __( 'This block is locked.' ) : undefined,
+	]
+		.filter( Boolean )
+		.join( ' ' );
 
 /**
  * Returns true if the client ID occurs within the block selection or multi-selection,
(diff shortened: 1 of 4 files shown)


packages/block-editor/src/components/list-view/utils.js
```diff
@@ -13,8 +13,19 @@ export const getBlockPositionDescription = ( position, siblingCount, level ) =>
...
 
-export const getBlockPropertiesDescription = ( isLocked ) =>
-	isLocked ? __( 'This block is locked.' ) : '';
+export const getBlockPropertiesDescription = ( blockInformation, isLocked ) =>
+	[
+		blockInformation?.positionLabel
+			? `${ sprintf(
+					// translators: 1: Position of selected block, e.g. "Sticky" or "Fixed".
+					__( 'Position: %1$s' ),
+					blockInformation.positionLabel
```
t-hamano:
```suggestion
					// translators: %s: Position of selected block, e.g. "Sticky" or "Fixed".
					__( 'Position: %s' ),
					blockInformation.positionLabel
```

Nit: This was the case with the original code, but since there is only one placeholder, I don't think the number is necessary.

t-hamano (approved):
LGTM! I tested this PR using NVDA and it now announces better when a block is selected.

Before:

```
Group  link  collapsed  Block 2 of 3, Level 1. This block is locked.
```

After:

```
Group  link  collapsed  Block 2 of 3, Level 1. Position: Sticky. This block is locked.

```

afercia:
Nice, thanks ❤️

PR #63851, opened 2024-07-23
Add theme type to the bug report issue template

### LINKED ISSUE #63645: Repository issue template: add theme type

It can be difficult for any contributor to add enough information so that reported issues can be reproduced.

It may help if the issue template not only asked if the bug has been tested without other plugins installed, but also if it has been tested with a block theme, classic theme or hybrid theme.

carolinan:
## What?

When reporting a bug, it can be difficult for any contributor to add enough information so that reported issues can be reproduced.
By encouraging reporters to confirm which theme type they used, the issue can be classified easier and initial testing can be limited to the disclosed theme type.

The checkboxes are optional because it is not relevant for _all_ issues.

Closes https://github.com/WordPress/gutenberg/issues/63645

## Testing Instructions
You can preview the file changes to see the new check boxes.


.github/ISSUE_TEMPLATE/Bug_report.yml
```diff
@@ -65,3 +65,12 @@ body:
...
                 required: true
+
+    - type: checkboxes
+      id: themes
+      attributes:
+          label: Please confirm which theme type you used for testing.
+          options:
+              - label: 'Block'
+              - label: 'Classic'
+              - label: 'Hybrid (e.g. classic with theme.json)'
```
desrosj, line 76: I think we should add an "I'm not sure" option here. It's possible someone has no idea and we don't want that to prevent them from submitting. Could also include "(please note the theme being used to test in the Environment Info)" so that it's not omitted from the report entirely.

desrosj (approved):
Thanks @carolinan! I merged in `trunk` so that the workflows would re-run. Too much time had passed, so they could no longer be re-run manually. I don't feel strongly about adding the 4th option that I noted. But this is good to go once that suggestion receives feedback!

desrosj (approved):

PR #63852, opened 2024-07-23
Update: Move the density picker to the view options menu.

jorgefilipecosta:
This results from the conversations on https://github.com/WordPress/gutenberg/pull/63367 with @youknowriad and @jameskoster.
It moves the density picker to be under the view config menu, decreasing the number of visible buttons at the top.

## Screenshots
Screenshot 2024-07-23 at 14 27 32

## Testing Instructions
I verified the density picker is available on the view config menu and still works as expected.

### DIFF
--- a/packages/dataviews/src/components/dataviews-view-config/index.tsx
+++ b/packages/dataviews/src/components/dataviews-view-config/index.tsx
@@ -19,10 +19,11 @@ import { cog } from '@wordpress/icons';
  * Internal dependencies
  */
 import { unlock } from '../../lock-unlock';
-import { SORTING_DIRECTIONS, sortLabels } from '../../constants';
+import { SORTING_DIRECTIONS, sortLabels, LAYOUT_GRID } from '../../constants';
 import { VIEW_LAYOUTS, getMandatoryFields } from '../../layouts';
 import type { NormalizedField, View, SupportedLayouts } from '../../types';
 import DataViewsContext from '../dataviews-context';
+import DensityPicker from '../../layouts/grid/density-picker';
 
 const {
 	DropdownMenuV2: DropdownMenu,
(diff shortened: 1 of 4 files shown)


jameskoster:
It works, but I don't think this is ideal from an a11y perspective, for the reasons Marco stated here: https://github.com/WordPress/gutenberg/pull/63367#issuecomment-2245138418.

If we commit to this direction we should probably revisit the design for the view options menu altogether, transitioning from a `menu` to a different UI.

jorgefilipecosta:
If we move from a dropdown to a popover I guess this design could work and be accessible right?
That would allow us to offer complex UI's under the view config and it is something I think we may need in the future.

Another option can a dropdown menu item open a popover with a custom UI? cc: @ciampo 
If yes would something like that work:
Screenshot 2024-07-23 at 18 09 37

Pressing to change the item size would then open a popover with the slider to change the item size. Or it would replace the dropdown contents with a slider to change the item size, and a back button to go back to the main dropdown.

Which option should we explore first? I'm also open to trying other options. cc: @jameskoster

jameskoster:
My understanding is that to satisfy a11y requirements we cannot mix the range control into the menu. So the options appear to be:

1. Update the entire 'View options' UI to live inside a popover, or a separate panel, utilising form controls rather than a menu (rough mockup to communicate the idea: https://github.com/WordPress/gutenberg/pull/63367#issuecomment-2245325573)
2. Ditch the range control in favor of `DropdownMenuRadioItem`s inside a flyout.

The first option seems better to me as it will allow for more design expression. It needs exploration though. I'll make a [new issue](https://github.com/WordPress/gutenberg/issues/63872) :)

ciampo:
> Another option can a dropdown menu item open a popover with a custom UI? cc: @ciampo

Maybe something like could work?

> Item size:
>   - small  (clicking it sets the size to small)
>   - medium   (clicking it sets the size to medium)
>   - large   (clicking it sets the size to medium)
>   - Custom...  (clicking it opens a new modal)

jameskoster:
That's what I tried to describe as option number 2. It meets the a11y requirement but is much less ergonomic to use compared with the range slider.

In the short term I don't mind leaving the range slider as it is, while we figure out how to incorporate it into the view options (https://github.com/WordPress/gutenberg/issues/63872).

However, I also wouldn't object to implementing 'option 2' in the short-term if there's a strong desire to do so.

jorgefilipecosta:
Closing in favor of https://github.com/WordPress/gutenberg/pull/64175.

PR #63854, opened 2024-07-23
Fix: Filters are reset when switching layout on the pages view.

jorgefilipecosta:
Fixes an issue noticed by @youknowriad at https://github.com/WordPress/gutenberg/pull/63203#issuecomment-2244993764.

Right now, on the pages list when we switch layouts the filters are reset. That happens because an effect to synchronize URL parameters with the view was not taking into account the already existing possible customizations of the view.
This PR fixes the issue.

## Testing Instructions
I opened the pages list view.
I added some filters.
I switched the layout and verified the filters I added were kept.
I switched the view e.g.: to published I verified the correct published filters, were applied and the previous layout I selected was kept.

### DIFF
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -39,6 +39,10 @@ function useView( postType ) {
 	const {
 		params: { activeView = 'all', isCustom = 'false', layout },
 	} = useLocation();
+	const previousIsCustom = usePrevious( isCustom );
+	const previousLayout = usePrevious( layout );
+	const previousActiveView = usePrevious( activeView );
+	const previousPostType = usePrevious( postType );
 	const history = useHistory();
 	const DEFAULT_VIEWS = useDefaultViews( { postType } );
 	const selectedDefaultView = useMemo( () => {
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/components/post-list/index.js
```diff
@@ -59,10 +63,28 @@ function useView( postType ) {
 	const [ view, setView ] = useState( selectedDefaultView );
 
 	useEffect( () => {
```
youknowriad, line 65: What's the purpose of this effect?

youknowriad: What happens if we just remove the effect?

youknowriad: Ok I'm guessing this hook is about updating the "view" object when the url changes (probably for default views only). Then:

I'd suggest just using `setView( previous => { some merge here } )` and only update the stuff that come from the url rather than setting the whole object. It also avoids all these usePrevious...

youknowriad: I think we probably don't need to compute the `selectedDefaultView` variable at all. We could just initialize the `useState` directly and update this hook to rely on the location params directly as well.

jorgefilipecosta: Hi @youknowriad, this effect is what changes the view when the URL goes from `?postType=page&layout=grid&activeView=published` to `?postType=page&layout=grid&activeView=future`. The sidebar does not changes the view directly. The sidebar, just updates URL's. If we remove this effect the sidebar does not works at all.
If we agree on this solution or something similar I will add a comment with that on the effect to make it clear for the future.

oandregal: I've prepared an alternative PR at https://github.com/WordPress/gutenberg/pull/63889

jorgefilipecosta:
Closed in favor of https://github.com/WordPress/gutenberg/pull/63889.

PR #63857, opened 2024-07-23
Core Data: Add new useEntityRecordsWithPermissions hook

youknowriad:
Related #55083 

## What?

In different dataviews, we need a way to check whether the user has the right permissions to edit or delete a given item, we've tried an approach by dynamically updating the "action" object's is Eligible but that was bad for several reasons:

 - Unclear how to retrigger a re-render when the permission resolves.
 - Unable to register actions globally.
 - Data and permissions not loaded at the same moment.

As discussed in slack, the right approach might be to embed the permissions within the "data". https://wordpress.slack.com/archives/C02QB2JS7/p1721732016850379

The current PR adds a new `useEntityRecordsWithPermissions` to `core-data` to be able to do that in an easy way. I'm not updating the actions yet, so in reality the current PR has no impact but it's the first step towards achieving that goal.

## Testing Instructions

Nothing to test really at the moment.

### DIFF
--- a/packages/core-data/src/hooks/use-entity-records.ts
+++ b/packages/core-data/src/hooks/use-entity-records.ts
@@ -13,6 +13,7 @@ import useQuerySelect from './use-query-select';
 import { store as coreStore } from '../';
 import type { Options } from './use-entity-record';
 import type { Status } from './constants';
+import { unlock } from '../lock-unlock';
 
 interface EntityRecordsResolution< RecordType > {
 	/** The requested entity record */
@@ -179,22 +180,12 @@ export function useEntityRecordsWithPermissions< RecordType >(
 		[ data, entityConfig?.key ]
 	);
 
-	// Todo: this selector is not memoized properly.
 	const permissions = useSelect(
 		( select ) => {
-			const { canUser } = select( coreStore );
-			return ids.map( ( id ) => ( {
-				delete: canUser( 'delete', {
-					kind,
-					name,
-					id,
-				} ),
-				update: canUser( 'update', {
-					kind,
-					name,
-					id,
-				} ),
-			} ) );
+			const { getEntityRecordsPermissions } = unlock(
+				select( coreStore )
+			);
+			return getEntityRecordsPermissions( kind, name, ids );
 		},
 		[ ids, kind, name ]
 	);
--- a/packages/core-data/src/hooks/use-entity-records.ts
+++ b/packages/core-data/src/hooks/use-entity-records.ts
@@ -4,6 +4,7 @@
 import { addQueryArgs } from '@wordpress/url';
 import deprecated from '@wordpress/deprecated';
 import { useSelect } from '@wordpress/data';
+import { useMemo } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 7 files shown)


packages/core-data/src/hooks/use-entity-records.ts
```diff
@@ -152,3 +153,61 @@ export function __experimentalUseEntityRecords(
...
+	const ids = useMemo(
+		() =>
+			data?.map(
+				// @ts-ignore
+				( record: RecordType ) => record[ entityConfig?.key ?? 'id' ]
+			) ?? [],
+		[ data, entityConfig?.key ]
+	);
+
+	// Todo: this selector is not memoized properly.
```
youknowriad: Thinking of moving the logic here into a private selector to be able to memoize it properly.

Mamaduka: Yeah, it probably would be easier to memoize as a separate private selector - `getEntityRecordsWithPermissions`.

youknowriad: Updated to use a private selector. The private selector is not perfect as it's recomputed when the whole permissions object changes but if we are to optimize this we need to rework the reducer shape to be per kind/name...

packages/core-data/src/hooks/use-entity-records.ts
```diff
@@ -152,3 +154,51 @@ export function __experimentalUseEntityRecords(
...
+
+	const dataWithPermissions = useMemo(
+		() =>
+			data?.map( ( record, index ) => ( {
+				// @ts-ignore
+				...record,
+				permissions: permissions[ index ],
+			} ) ) ?? [],
+		[ data, permissions ]
+	);
```
ellatrix, line 201: Why not do this is a single memoized selector?

ellatrix: I think that would also reduce the loops from 3 to 1?

ellatrix (commented):
Sounds great to me! Should we update the permissions in isEligible to use this, so we know what it's like?

packages/core-data/src/hooks/use-entity-records.ts
youknowriad: > I think that would also reduce the loops from 3 to 1?

I don't think it reduces anything as the loop will exist but within the selector. I'm not sure which is better to be honest. A huge memoized selector or smaller ones with some in between memoizations. I have a feeling that the latter is better in terms of performance but the former is easier to work with.

youknowriad:
> Sounds great to me! Should we update the permissions in isEligible to use this, so we know what it's like?

This is what I want to do in a follow-up but I found that the way we fetch patterns and template parts is so complex (it's actually hard to even call a simple selector there). Doable but that part is not great code to be honest.

packages/core-data/src/hooks/use-entity-records.ts
ellatrix: Yes, but it would just be one loop, that immediately fetches the permission and sets it on the item?

ellatrix:
Yes, just mentioning it because maybe it's relevant in making a decision? Fine with me to try in a follow-up

ellatrix (approved):

youknowriad:
I'm afraid, I have to do a bigger refactoring to the templates and patterns pages to be able to do that easily. I'll try to either open a refactoring PR or open a follow-up PR to this one before merging the current one.

packages/core-data/src/private-selectors.ts
```diff
@@ -50,3 +51,24 @@ export const getBlockPatternsForPostType = createRegistrySelector(
...
 		)
 );
+
+/**
+ * Returns the entity records permissions for the given entity record ids.
+ */
+export const getEntityRecordsPermissions = createSelector(
+	( state: State, kind: string, name: string, ids: string[] ) => {
+		return ids.map( ( id ) => ( {
+			delete: canUser( state, 'delete', {
```
jorgefilipecosta: We now have two canUser calls, but I checked and we still make one request per id and not two so it is fine.

jorgefilipecosta (approved):
The PR seems good to me 👍

packages/core-data/src/private-selectors.ts
Mamaduka: Yes, requests are deduped for related actions - #43480.

Mamaduka (approved):
I think this is OK as a temporary solution, and I see it has already helped us [highlight some pain points](https://github.com/WordPress/gutenberg/pull/63857#issuecomment-2247308033).

> I'm not updating the actions yet, so in reality the current PR has no impact but it's the first step towards achieving that goal.

@youknowriad, what actions updates do you have in mind

youknowriad:
@Mamaduka All the actions in post-actions/actions.js basically. Anything using useCanUserEligibilityCheckPostType

PR #63858, opened 2024-07-23
Block Variations: Have `getActiveBlockVariation` fall back to default variation

ockham:
## What?
If a block doesn't match any of the block type's variations, return the variation that has `isDefault` set to `true` (as long as it doesn't have an `isActive` field, as that would mean that it didn't match the latter).

## Why?
See https://github.com/WordPress/gutenberg/pull/63100#issuecomment-2245358898.

## How?
Explained above :)

## Testing Instructions
Cherry-pick ecedae03d349c21d53b41aea14938420a42417ca (from https://github.com/WordPress/gutenberg/pull/63100) on top of this PR.

Then, in the editor, insert multiple Group blocks, choosing a different variation each time. Save your post and reload the editor. Verify that the block inspector reflects the correct variation for each block.

unnamed

### DIFF
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -298,8 +298,9 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 
 	// If no variation matches the isActive condition, we return the default variation,
 	// but only if it doesn't have an isActive condition that wasn't matched.
-	// This fallback is only applied for 'block' scope to avoid affecting block name display.
-	if ( ! match && scope === 'block' ) {
+	// This fallback is only applied for the 'block' and 'transform' scopes but not to
+	// the 'inserter', to avoid affecting block name display there.
+	if ( ! match && [ 'block', 'transform' ].includes( scope ) ) {
 		match = variations.find(
 			( variation ) =>
 				variation?.isDefault && ! Object.hasOwn( variation, 'isActive' )
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -298,7 +298,8 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 
 	// If no variation matches the isActive condition, we return the default variation,
 	// but only if it doesn't have an isActive condition that wasn't matched.
-	if ( ! match ) {
+	// This fallback is only applied for 'block' scope to avoid affecting block name display.
+	if ( ! match && scope === 'block' ) {
 		match = variations.find(
 			( variation ) =>
 				variation?.isDefault && ! Object.hasOwn( variation, 'isActive' )
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -295,6 +295,11 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 			return match || variation;
 		}
 	}
+
+	// If no variation matches the isActive condition, we return the default variation.
+	if ( ! match ) {
+		match = variations.find( ( { isDefault } ) => !! isDefault );
+	}
 	return match;
 }
 
--- a/packages/blocks/src/store/selectors.js
+++ b/packages/blocks/src/store/selectors.js
@@ -296,9 +296,13 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 		}
 	}
 
-	// If no variation matches the isActive condition, we return the default variation.
+	// If no variation matches the isActive condition, we return the default variation,
+	// but only if it doesn't have an isActive condition that wasn't matched.
 	if ( ! match ) {
-		match = variations.find( ( { isDefault } ) => !! isDefault );
+		match = variations.find(
+			( variation ) =>
+				variation?.isDefault && ! Object.hasOwn( variation, 'isActive' )
+		);
 	}
 	return match;
 }
--- a/packages/block-editor/src/components/block-variation-transforms/index.js
+++ b/packages/block-editor/src/components/block-variation-transforms/index.js
@@ -162,7 +162,8 @@ function __experimentalBlockVariationTransforms( { blockClientId } ) {
 				return {
 					activeBlockVariation: getActiveBlockVariation(
 						name,
-						getBlockAttributes( blockClientId )
+						getBlockAttributes( blockClientId ),
+						'transform'
 					),
 					variations: name && getBlockVariations( name, 'transform' ),
 					isContentOnly:
--- a/packages/blocks/src/store/test/selectors.js
+++ b/packages/blocks/src/store/test/selectors.js
@@ -822,6 +822,85 @@ describe( 'selectors', () => {
 					} )
 				).toEqual( variations[ 1 ] );
 			} );
+			it( 'should fall back to default variation if no other variation matches', () => {
+				const variations = [
+					{
+						name: 'variation-1',
+						attributes: {
+							testAttribute: 1,
+						},
+						isDefault: true,
+					},
+					{
+						name: 'variation-2',
+						attributes: {
+							testAttribute: 2,
+						},
+						isActive: ( blockAttributes, variationAttributes ) => {
+							return (
+								blockAttributes.testAttribute ===
+								variationAttributes.testAttribute
+							);
+						},
+					},
+					{
+						name: 'variation-3',
+						attributes: {
+							firstTestAttribute: 3,
+						},
+						isActive: [ 'testAttribute' ],
+					},
+				];
+
+				const state =
+					createBlockVariationsStateWithTestBlockType( variations );
+
+				expect(
+					getActiveBlockVariation( state, blockName, {
+						testAttribute: 55,
+					} )
+				).toEqual( variations[ 0 ] );
+			} );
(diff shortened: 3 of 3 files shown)


michalczaplinski (approved):
This makes sense!

- If there **is** a match for the `isActive` field, we return that match. 
- If there **isn't** a match for the `isActive` field, we return the variation with the `isDefault` flag.

I don't see any flaws in that logic 🙂. I would have also expected the `isDefault` to work this way.

In a follow-up, it would be nice to log a warning if multiple blocks accidentally have the `isDefault` flag. With the current implementation, the first variation in the array would be returned, which is reasonable behavior.

ockham:
Thank you @michalczaplinski!

Unfortunately, e2e tests do not seem to agree with this reasoning 😂 -- I'll have to figure out what's going on there.

michalczaplinski:
> Thank you @michalczaplinski!
> 
> Unfortunately, e2e tests do not seem to agree with this reasoning 😂 -- I'll have to figure out what's going on there.

😅 . Yeah, you're right. 

I thought this might be because the current PR should perhaps be stacked on top of https://github.com/WordPress/gutenberg/pull/63100. I checkout out both and merged #63100 into the current one locally, but the e2e tests still fail.

ockham:
I've started digging. Here's one of the tests that's failing: https://github.com/WordPress/gutenberg/blob/bfffbe0ec4b67d76e90c691102cd74252594db33/test/e2e/specs/editor/plugins/block-variations.spec.js#L47-L60

The "Large Quote" variation of the Quote block is registered by an e2e test helper plugin: https://github.com/WordPress/gutenberg/blob/bfffbe0ec4b67d76e90c691102cd74252594db33/packages/e2e-tests/plugins/block-variations/index.js#L23-L30

Here's a screenshot (that's part of the artifact generated by CI when the e2e test failed):

test-failed-1

ockham:
I tried to repro [that e2e error](https://github.com/WordPress/gutenberg/pull/63858#issuecomment-2248396770) manually but haven't been able so far.

Anyway, I suspect it has to do with two different variations having `isDefault` set to `true` 😬

gziolo:
> Anyway, I suspect it has to do with two different variations having isDefault set to true 😬

It’s possible to accidentally set multiple variations as default through plugins so it picks the first default found if I recall correctly.

t-hamano:
This issue has been punted to 6.8 and has not seen any activity since. Let's move this issue to the 6.9 Editor Task Board and see what can be done in 6.9.

For now, I would like to merge the latest trunk into this PR, but first I would appreciate it if you could check whether this PR is currently valid.

t-hamano:
From what I can tell, the reason for the E2E test failure is that the displayed block name has changed. For example, try selecting the `50 / 50` variation in the Columns block.

columns

I think we need the logic to check the `scope` somewhere. 80338cffd5abf7d02c242e7c54706a6df619c231 is one approach, but it may not be correct.

ockham:
Thank you very much for picking this up, @t-hamano, and apologies for not replying sooner!

Your fix looks good and makes sense to me. It might require some more tweaking, as the newly added unit tests are failing now.

packages/blocks/src/store/selectors.js
```diff
@@ -295,6 +295,16 @@ export function getActiveBlockVariation( state, blockName, attributes, scope ) {
 			return match || variation;
 		}
 	}
+
+	// If no variation matches the isActive condition, we return the default variation,
+	// but only if it doesn't have an isActive condition that wasn't matched.
+	// This fallback is only applied for 'block' scope to avoid affecting block name display.
+	if ( ! match && scope === 'block' ) {
```
ockham: The unit tests don't currently set the `scope` argument, so they fall back to its default value. It turns out that `getActiveBlockVariation` [simply passes the `scope` argument through](https://github.com/WordPress/gutenberg/blob/983df1f7bcb6ef39d5421a229a4b0362bc89825b/packages/blocks/src/store/selectors.js#L241) to `getBlockVariations`.

There, it [defaults to `[ 'block', 'inserter' ]`](https://github.com/WordPress/gutenberg/blob/983df1f7bcb6ef39d5421a229a4b0362bc89825b/packages/blocks/src/store/selectors.js#L183-L187).

In order to account for this case, we might need to adjust the check here.

t-hamano:
> https://github.com/WordPress/gutenberg/commit/80338cffd5abf7d02c242e7c54706a6df619c231 is one approach, but it may not be correct.

I realized that 80338cffd5abf7d02c242e7c54706a6df619c231  is not the right approach 😅 If I cherry-pick ecedae03d349c21d53b41aea14938420a42417ca on top of this PR, the normal Group variation is no longer marked as active in the sidebar:

image

In any case, this PR does not appear to be essential for WP 6.9, so I would like to remove it from the project board.

ockham:
> I realized that [80338cf](https://github.com/WordPress/gutenberg/commit/80338cffd5abf7d02c242e7c54706a6df619c231) is not the right approach 😅 If I cherry-pick [ecedae0](https://github.com/WordPress/gutenberg/commit/ecedae03d349c21d53b41aea14938420a42417ca) on top of this PR, the normal Group variation is no longer marked as active in the sidebar: [...]

I think I found a solution to this problem: https://github.com/WordPress/gutenberg/pull/63858/commits/7dad4d2153288e3fbc6b3ddfe0fad03d67239f25

We still need to update the unit test to set the `scope` argument to either `block` or `transform` (and maybe add a test for `inserter` scope and/or no scope argument to verify that it does _not_ fall back to the `isDefault` variation in those cases). We should probably also audit other callsites of `getActiveBlockVariation`.

> In any case, this PR does not appear to be essential for WP 6.9, so I would like to remove it from the project board.

Yep, that's perfectly fine IMO 👍

t-hamano:
@ockham Thanks for the update!

P.S. When we ship this PR, the logic of the `getActiveBlockVariation` selector will change a bit, so a dev note might be needed. I"m sure the selector is often used by block developers.

t-hamano (approved):
I cherry-picked https://github.com/WordPress/gutenberg/commit/ecedae03d349c21d53b41aea14938420a42417ca and smoke-tested again, and I don't see any problems or regressions 👍

t-hamano:
P.S. I tested https://github.com/WordPress/gutenberg/pull/63100#issuecomment-2245065349 again with the following code:

```javascript
wp.blocks.registerBlockVariation("core/group", {
	name: "wpcomsp/marquee-content",
	title: "Marquee Content",
	icon: "editor-table",
	description: "A marquee content block.",
	scope: ["block", "inserter"],
	attributes: {
		isMarquee: true,
		layout: { type: "flex", flexWrap: "nowrap" },
		className: "wpcomsp-marquee-content",
		allowedBlocks: [
			"core/paragraph",
			"core/heading",
			"core/image",
			"core/buttons",
		],
	},
	isActive: ["layout.type", "isMarquee"],
});
```

> I can see the `Marquee Content` block in the inserter, but upon inserting I get the Row block instead. 
> 
> Curiously, when I **increase the specificity** by adding an additional item in the `isActive` array like `"layout.flexWrap"`, it seems to work correctly for the Marquee Content block:
> 
> ```js
>   isActive: ["layout.type", "layout.flexWrap", "isMarquee"],
> ```

I tested the above behavior and this PR seems to be better than trunk.

When inserting the "Marquee Content" block...

| `isActive` value | trunk | This PR |
|--------|--------|--------|
| `["layout.type", "isMarquee"]` |  ❌Get the Row block | ❌Get the Row block |
| `["layout.type", "layout.flexWrap", "isMarquee"]` | ❌Get the Row block | ✅ Get the Marquee Content block |

ockham:
Thank you for resurrecting this, @t-hamano!

t-hamano:
Just to be safe, let's add the Needs Dev Note label.

ockham:
# Dev Note

WordPress 6.9 introduces a [small change](https://github.com/WordPress/gutenberg/pull/63858) to active block variation detection.

To understand what that change looks like, recall that a block variation can include an [`isActive` criterion](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#using-isactive) that will be used to determine if a given block instance matches the variation.

The change concerns the case in which none of the `isActive` criteria of a block type’s variations match a given block instance. In this case, WordPress 6.9 will use the variation with the [`isDefault` property](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/#using-isdefault) set to `true` as fallback, if any.

Two limitations apply:

1. The variation with the `isDefault` property is only used as fallback if it doesn’t _also_ include an `isActive` criterion, as that would mean that the variation did not match that criterion (and thus cannot be assumed to be active).
2. The fallback is only applied for the `block` and `transform` scopes but not to the `inserter`, as to not affect the block name displayed there.

In practice, the updated behavior should be consistent with users' and developers' expectations, as the `isDefault` variation is typically used to set a default block variation different from the “vanilla” block type.

Together with the improvements made to active block variation detection [in WordPress 6.6](https://make.wordpress.org/core/2024/06/18/improvements-to-active-block-variation-detection/) , this change allows block authors to express the criteria for active block detection in a very succinct and JSON-only manner. For an example, see how these principles were applied to [rewrite Group block variations](https://github.com/WordPress/gutenberg/pull/63100).

PR #63859, opened 2024-07-23
Don't automatically show inserter when zoom out mode initiates

jeryj:
## What?

Don't open the block inserter when entering zoom out mode (such as from global styles).
## Why?

It's an odd experience.

## How?

Remove the code that opens the inserter when the zoom out inserter buttons mount.

## Testing Instructions

- Use a theme like 2024 that has various theme.json styles
- Open the styles side panel
- Click Browse Styles
- Zoom out mode should trigger
- The block inserter should not be open
- Click a zoom out inserter on the canvas
- Pattern inserter should open


scruffian (approved):
I agree. This will also make https://github.com/WordPress/gutenberg/pull/63870 simpler.

Issue #63865, opened 2024-07-23
Assemble new pages with zoom out


scruffian:
It’s not uncommon for website builders to offer a perspective that allows you to see more of your page by stacking sections of a page together. In WordPress, these are called [Patterns](https://rich.blog/patterns/).

I propose a similar experience within WordPress: an “Add new page” composition flow where you stack patterns on a page, reorder, apply section styles, remove, and even shuffle among similar patterns. 

#### Key points 
- This concept would replace the starter content modal. 
- There should be a clear way to start with a blank page. 
- Perhaps if you indicate to start with a blank page, there’s a modal asking you if that should always happen, and informing you where to change it in the future. 
- This view should open the Inserter with the Patterns tab selected, with zoom out mode engaged. 
- Help text should render if there are no patterns added to the page. 
- Starter content patterns should be prioritized, as full page designs to start with.

mtias:
One thing we should incorporate is clearer distinction between section patterns and full page starter templates in the sidebar UI — it's all a bit mixed right now.

MaggieCabrera:
> One thing we should incorporate is clearer distinction between section patterns and full page starter templates in the sidebar UI — it's all a bit mixed right now.

I was coming to say this. Themes that have full page patterns should have them open by default when in this kind of flow, since we are replacing the modal that is surfacing them now

scruffian:
We have an interesting approach in https://github.com/WordPress/gutenberg/pull/61489

getdave:
> opening the editor in zoom out mode so that users can begin assembling their page from patterns

Small nit, but could we rename this Issue to better reflect the goal? Perhaps it should be `Patterns Zoom out mode on new Page creation`?

Currently `Compose pages with zoom out view` seems like a very broad scope for this issue.

annezazu:
Noting that there's an in progress PR here https://github.com/WordPress/gutenberg/pull/66836 to address this!

Mamaduka:
Re-opening the issue. The #66836 has been reverted via #69081.

It would be good for future resolutions to account items listed in #68736.

PR #61489, opened 2024-05-08
Zoom out: try pattern inserter instead of starter pattern modal for new pages

ellatrix:
# Moved to #66836.

## What?

Instead of showing the starter patterns modal when creating a draft page, this PR tries opening the patterns inserter instead (+ zoom out mode).

zoom-out-starter

PR #59618, opened 2024-03-06
Refactor zoom-out iframe scale

ajlende:
## What?

Refactors the changes from #59342 to maintain backwards compatibility with the old `IFrame` component and simplifies/documents some code.

## Why?

I noticed that we changed a public API in #59342. The API is `__unstable`, but I figure those components would follow the same rules for backwards compatibility as `__experimental` since the addition of `privateApis`.

This also removes the need for the iframe to need to know about the `zoom-out` mode which I think is slightly better for code organization.

## Testing Instructions

See that the scaling for zoom-out mode works as it did before.

PR #63870, opened 2024-07-23
Zoom Out: Add a control to enter and leave zoom out mode

scruffian:
## What?
Add a control to allow users to select between zoom out and edit mode.

## Why?
Sometimes users might want this level of control rather than us doing it for them.

## How?
Add a control to the device preview dropdown.

## Testing Instructions
1. Open the Site Editor
2. Open the device preview drop down
3. Select "Zoom Out" and see that you enter zoom out mode
4. 3. Select "Zoom In" and see that you leave zoom out mode

## Screenshots or screencast 
Screenshot 2024-08-01 at 08 11 19
Screenshot 2024-08-01 at 08 11 12


packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -94,6 +97,28 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
...
+							) }
+							{ __unstableGetEditorMode() === 'zoom-out' && (
+								<MenuItem
+									onClick={ () =>
+										__unstableSetEditorMode( 'edit' )
+									}
+								>
+									{ __( 'Zoom In' ) }
+								</MenuItem>
+							) }
```
jeryj: This will probably result in a focus loss. It would be better to toggle the same option and have aria states for it, or have Zoom Out and Zoom In options always available.

jeryj (commented):
If we're automatically entering zoom-out mode from the pattern inserter, I think it's a better experience to automatically exit zoom out mode when the pattern inserter is closed.

packages/editor/src/components/preview-dropdown/index.js
jeryj: For some reason Zoom Out and Zoom In is confusing to me when I'm using it. I think offering 100% and 50% as options feels better.

jeryj: https://github.com/user-attachments/assets/d1c82f33-b406-4a0a-8b86-71aa3564f002

richtabor: Let's try desktop and 100% the same thing. 

- 100%
- 50%
- Tablet
- Mobile

scruffian:
> If we're automatically entering zoom-out mode from the pattern inserter, I think it's a better experience to automatically exit zoom out mode when the pattern inserter is closed.

I removed the hook as with it we end up stuck in zoom out mode.

packages/editor/src/components/preview-dropdown/index.js
scruffian: FWIW Zoom out isn't always 50%

jeryj:
@richtabor Should the device preview icon also change when 50% is selected? Something that shows the desktop zoomed out or to indicate zoom out?
header of site editor with desktop icon and 50% selected in the dropdown menu

richtabor:
> @richtabor Should the device preview icon also change when 50% is selected? Something that shows the desktop zoomed out or to indicate zoom out?

I was thinking `100%` and `50%` respectively.

richtabor:
The menu item can be `Zoom to 100%` and `Zoom to 50%`

packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -62,33 +65,88 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
...
+							</MenuItem>
+						) }
 						<MenuItem
-							onClick={ () => setDeviceType( 'Tablet' ) }
+							onClick={ () => {
+								setDeviceType( 'Tablet' );
+								if ( isZoomOutExperiment ) {
+									// Rather than setting to edit, we may need to set back to whatever it was previously.
+									__unstableSetEditorMode( 'edit' );
+								}
```
jeryj: The non-zoom out ones are all similar: Set device type + exit zoom out. I think a named function to do this would make the code easier: `handleDevicePreviewClick( deviceType, zoomOut = false )`

packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -62,33 +65,88 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
...
+		if ( __unstableGetEditorMode() === 'zoom-out' ) {
+			return __( '50%' );
+		}
+		return __( '100%' );
+	};
+
+	const getIcon = () => {
+		if ( isZoomOutExperiment && deviceType === 'Desktop' ) {
+			return zoomIcon;
+		}
```
jeryj: Should it be the Desktop SVG + the %? Like `[desktop icon] 50%`?

jeryj: Or maybe the text can be _within_ the desktop icon? Like, the percentage is displayed on the screen if it's at 50%?

jeryj:
Maybe too much of an edge case for right now, but you get in an incorrect state if you:
- Set to 50%
- Resize to a smaller window so that the preview icon disappears
- Resize to wider window
- 50% device preview without being in zoom out mode

This shouldn't block this pr though.

https://github.com/user-attachments/assets/931665f5-8631-4a02-80e6-a16abf07b603

richtabor:
> Maybe too much of an edge case for right now, but you get in an incorrect state if you:

Yes, it should reset to 100%.

packages/editor/src/components/preview-dropdown/index.js
scruffian: I think to do that we would need new SVGs

scruffian:
> Maybe too much of an edge case for right now, but you get in an incorrect state if you:

Fixed in [76c6a8c](https://github.com/WordPress/gutenberg/pull/63870/commits/76c6a8cffe339ac739492da1e1436860bdbf6ae3)

scruffian:
I think this is ready for a review. There's still a bug when switching modes using the toolbar but I think that's probably for a different PR.

packages/block-editor/src/components/inserter/menu.js
```diff
@@ -149,7 +149,18 @@ function InserterMenu(
...
+		if ( showZoomOut ) {
+			__unstableSetEditorMode( 'zoom-out' );
+		}
+
+		return () => {
+			if ( showZoomOut ) {
+				__unstableSetEditorMode( 'edit' ); // TODO: set back to whatever it was previously
+			}
+		};
+	}, [ showZoomOut ] );
```
ajlende: Why did you have to switch away from `useZoomOut` here and below https://github.com/WordPress/gutenberg/pull/63870/files#diff-fe01ab6018c513cc3b7cdb7e27a70b4aad3d73c1fe8d1267a91c1d80b6a4334fR61-R75?

packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -110,13 +159,38 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
...
+			deviceType !== 'Mobile' &&
+			deviceType !== 'Tablet'
+		) {
+			if ( editorMode === 'zoom-out' ) {
+				return <>{ __( '50%' ) }</>;
+			}
+			return <>{ __( '100%' ) }</>;
+		}
+
+		return deviceIcons[ deviceType.toLowerCase() ];
```
ajlende: You're setting the `deviceType` to `ZoomIn` or `ZoomOut` above and the options are only available in the experiment, so the complicated function isn't necessary.

```suggestion
	const getIcon = () => {
		switch (deviceType) {
			case 'ZoomOut':
				return __( '50%' );
			case 'ZoomIn':
				return __( '100%' );
			default:
				return deviceIcons[ deviceType.toLowerCase() ];
		}
	}
```

packages/block-editor/src/components/inserter/menu.js
scruffian: useZoomOut works when the component mounts, whereas this is something that needs to continue to react to changes in the component.

scruffian: If we leave this in then changing the value from 50% to 100% in the preview dropdown doesn't work, as this component forces it back to 50%.

scruffian: Actually I found a way to use the hook, as long as we change the dependency array.

jeryj (commented):
The spacing on the button feels off.

annezazu:
Quickly took this PR for a spin and have some feedback. It feels a bit odd to have "100%" visible when everything else is an icon. It's visually jarring. When switching to 50%, it's also not clear that the experience is now going to change and that the only way to edit blocks granularly is to change to another option. Since in every other option you can have granular editing and how you interact with the experience doesn't change, I think we need to do more to communicate the change. I could see a user getting stuck and frustrated. Perhaps this could be rectified by keeping "desktop" as an option and then having just "Zoom to 50%" listed. 

https://github.com/user-attachments/assets/c0383c3a-7923-4ecc-b569-4dafc000d61c

When zoomed out to 50%, I also found it odd that List View only shows "Group" with no inner blocks listed when using base TT4 (not sure how we're determining what to show in List View). When opening up the Inserter, it also defaults to blocks. After adding a block, like a heading block, I am then able to edit the heading block but if I try to delete the text and then delete the block by continuing to hit delete, I got it stuck in a weird state where the block was simple moved up into the prior section, unable to be selected or edited.

If you also test this in a situation where there aren't as many Group blocks, the experience quickly feels pretty strange, especially anywhere there's a spacer block (you end up with a giant blue line). You also lose the vertical toolbar and are only left with big blue + buttons. It's not clear to me when or why the vertical toolbar appears. I tried adding in different patterns until finally one appeared but, in some cases, it still didn't appear. 

https://github.com/user-attachments/assets/e4b208fc-a3dc-4e02-a116-c66d101e8177

Updated to add the recordings.

jasmussen:
Nice work, took this for a spin:

zoom

I tend to think at this point, it is _something like this_ which we will need in order to unstick this. A separate zoom control that works separately to the pattern inserter, one which can be manually invoked or exited. It seems the main way to thread the needle between the desire to be able to insert patterns at the root level only, as sections, while also affording a simple way to insert patterns wherever you like them, unzoomed.

I'd echo all of Anne's feedback, there are still a lot of rough edges to refine, and even unanswered questions:

* Can we programmatically invoke the zoom when creating a new page? And if we do that, should there be a "Done" button to exit this mode, or would you be left in 50% zoom when you're done with the initial assembling?
* Should this "Done" flow be a more key part of entering and existing this zoom?
* If we only ever have two zoom levels, zoomed in and zoomed out, should it be a dropdown, or can it be a toggle?

At the same time, this preview dropdown does feel the right place to add a manual zoom control, as it's related to the other viewport sizes. There's also something very obvious about seeing 50% or 100% as far as "how do I zoom in or out" flows. These zooms are also mutually exclusive to tablet and mobile views, making it sensible to group these together.

So at this point it feels increasingly solid that we do need a manual interface for this, just as the PR suggests. The details around how you invoke and exit could still feel more solid, both visually to increase the obviousness of it, but also flow-wise, with an eye towards the "add new page flow". 

It's mainly a question of threading the needle on the best interface for it. Is it a split dropdown, where the dropdown allows access to mobile/tablet/view site, but the main part of the button is just a toggle? And how might that look in a way that makes this particular interface feel an obvious way to explore to access these features? It _feels_ solvable.

Given this, perhaps it would be the most productive for this experiment to merge. The main question is whether it should still be behind the zoom experiment or not. 

I'd be happy to try this, but I'd love an additional sanity check from @WordPress/gutenberg-design.

scruffian:
> Given this, perhaps it would be the most productive for this experiment to merge. The main question is whether it should still be behind the zoom experiment or not.

Thanks for the feedback. In this PR the control is not behind the experiment.

scruffian:
> Can we programmatically invoke the zoom when creating a new page?

We are trying that here. https://github.com/WordPress/gutenberg/pull/61489

>  I could see a user getting stuck and frustrated.

This is one of my main concern too. I propose that we exit zoom out when you insert a block: https://github.com/WordPress/gutenberg/pull/64169

scruffian:
> When opening up the Inserter, it also defaults to blocks.

Opened a PR to address that here: https://github.com/WordPress/gutenberg/pull/63870

scruffian:
One question in my mind is should we separate the zoom out mode from scaling the canvas. The mode is fairly restrictive and only valuable in certain circumstances.

scruffian:
> Can we programmatically invoke the zoom when creating a new page? And if we do that, should there be a "Done" button to exit this mode, or would you be left in 50% zoom when you're done with the initial assembling?

I'm hesitant about this because people currently find pattern modal that loads on new pages so annoying - it would also be annoying to have to click "Done" every time you want to create a page.

MaggieCabrera:
> One question in my mind is should we separate the zoom out mode from scaling the canvas. The mode is fairly restrictive and only valuable in certain circumstances.

I think about this as well. It's not strictly just zooming in or out, you can't do the same things as you would do when in 100% mode

richtabor:
> Can we programmatically invoke the zoom when creating a new page? And if we do that, should there be a "Done" button to exit this mode, or would you be left in 50% zoom when you're done with the initial assembling?

> I'm hesitant about this because people currently find pattern modal that loads on new pages so annoying - it would also be annoying to have to click "Done" every time you want to create a page.

We can explore flows in https://github.com/WordPress/gutenberg/issues/63865, perhaps one like this (with a clear "Start with a blank page" type option. Maybe there's a "Make this the default" type option if you do select to start with a blank page. 

My initial thought is that you wouldn't be left at 50% when you're "done" stacking patterns on a page.

richtabor:
> Given this, perhaps it would be the most productive for this experiment to merge. The main question is whether it should still be behind the zoom experiment or not.

It's not behind the current zoom out experiment, which is particular to zooming out when you select patterns—as then you couldn't turn off that experiment and try this fully "manual" zoom out control. 

It's important we try this _without_ the patterns category auto zoom out experiment; to help validate both ideas separately. 

If we're really unsure about this, we could create another experiment for the header zoom controls. I think it's fine to not, given where we are in the release cycle, but I'm not opposed to a small experiment.

richtabor:
> Quickly took this PR for a spin and have some feedback. It feels a bit odd to have "100%" visible when everything else is an icon. It's visually jarring. When switching to 50%, it's also not clear that the experience is now going to change and that the only way to edit blocks granularly is to change to another option. 

> There's also something very obvious about seeing 50% or 100% as far as "how do I zoom in or out" flows. 

@annezazu I lean towards @jasmussen's assessment here, where seeing 100% informs you that it can be something else (likely another percentage). 

Education wise, would it be useful having a guide render the first time you engage zoom to 50%? Like when you open Styles for the first time? Not 100% sure, just an idea.

stokesman:
> One question in my mind is should we separate the zoom out mode from scaling the canvas. The mode is fairly restrictive and only valuable in certain circumstances.

I think this makes sense and is the same thing Alex was getting at in https://github.com/WordPress/gutenberg/pull/59618#issuecomment-2040270066. Small excerpt:
> What we're calling zoom-out mode, I think should really be named assembly mode. The "zoom out" visuals should act more like the mobile preview as a "view" instead of a "mode", and the assembly mode could use the zoom-out view.

jeryj (approved):
Giving an approving review on the code. This is ready to merge if people agree on the behavior.

richtabor:
Another thought (for a potential next quick iteration to try) is to associate desktop with zoom a bit more: 

Not exactly this, but communicate the idea that zoom is relative to desktop and not tablet/mobile. Something to consider. 

 CleanShot 2024-08-02 at 16 23 46

cc @mtias

packages/block-editor/src/components/block-tools/index.js
```diff
@@ -251,12 +251,11 @@ export default function BlockTools( {
 					name="__unstable-block-tools-after"
 					ref={ blockToolbarAfterRef }
 				/>
-				{ window.__experimentalEnableZoomedOutPatternsTab &&
-					isZoomOutMode && (
-						<ZoomOutModeInserters
-							__unstableContentRef={ __unstableContentRef }
-						/>
-					) }
```
jeryj, line 259: Removing the `<ZoomOutModeInserters />` from behind the experimental  `window.__experimentalEnableZoomedOutPatternsTab` flag, as you can manually enter zoom out mode via this preview dropdown. I think this is the right call here. Otherwise you don't see the zoom out mode inserters unless you have the experiment on.

jeryj:
@richtabor Updated to use the desktop icon. I think it feels a lot nicer with Desktop and Desktop (50%) and clarifies the relationship. I added the 50% on the left side of the icon because it allowed the Desktop icon to stay in the same location when selecting Desktop (50%).

https://github.com/user-attachments/assets/df49a705-353c-4b29-a094-94198ddea1c4

paaljoachim:
Thank you for working on this feature!

I find it very helpful to have a Desktop (50%) zoom feature. I see that this view has another focus compared to the regular Desktop view.

Screenshot 2024-08-02 at 19 59 10

The focus on the 50% view seems to me to be about the general layout of the content on the screen. Moving Group/container kind of  blocks or clicking to trash these.

This means that Desktop has the regular editing experience and that the 50% view contains another editing experience.
Does this view need additional visual hints that one is now in another kind of editing experience?
Perhaps something like a ? with an info tooltip above the drag dots? Or something else where one gets some short info about this view and editing experience?

jasmussen:
I'm still a believer, and it remains for me the visual of the trigger itself. I don't know that the chevron down mixes well with the other items, or that the text mixes with the icons. Not a blocker per se, but an opportunity to improve that dropdown overall. Even in trunk it's not the clearest, sans-zoom features. Our best bet is to probably revisit the toolbar organization in the right side, as has been discussed elsewhere, and see how those pieces come together. Even little things like https://github.com/WordPress/gutenberg/pull/64187 might play a role.

scruffian:
With the latest changes from @jeryj this no longer uses the chevron down:

Screenshot 2024-08-05 at 10 41 34

scruffian (commented):
Thanks for all the work on this @jeryj I think it looks great. Let's bring it in and we can iterate on the design.

richtabor:
>  I added the 50% on the left side of the icon because it allowed the Desktop icon to stay in the same location when selecting Desktop (50%).

It's not quite working as is, but it's good to move this forward. Let's circle back design wise on this.

t-hamano:
When I select "Desktop (50%)", the breadcrumbs are hidden. I think this is because the editor mode is `zoom-out`, so [this conditional statement](https://github.com/WordPress/gutenberg/blob/7559e74027f3be144509b56ab010a054a1801a58/packages/editor/src/components/editor-interface/index.js#L212) does not match. Is this the intended behavior?

Please refer to this issue: #64259

richtabor:
>  I think this is because the editor mode is zoom-out

Yes, the "Desktop (50%)" is zoom out.

t-hamano:
So does that mean that the breadcrumbs only disappear in this mode?

richtabor:
> So does that mean that the breadcrumbs only disappear in this mode?

Yes, it looks like no breadcrumbs when zoomed out.

packages/block-editor/src/components/iframe/index.js
```diff
@@ -242,10 +242,8 @@ function Iframe( {
 	const isZoomedOut = scale !== 1;
 
 	useEffect( () => {
-		if ( ! isZoomedOut ) {
-			prevContainerWidth.current = containerWidth;
```
stokesman, line 246: I wonder what led to this change? As I understand it the value of `prevContainerWidth` should be that of `containerWidth` before Zoom Out is engaged because the scale calculation needs to arrive at a scale which preserves the document width. That calculation wasn’t perfect before but this seems to have thrown it off a good deal when Zoom Out is engaged by way of the Inserter (not through the control added by this PR).

scruffian: Good catch, I made a PR for that here: https://github.com/WordPress/gutenberg/pull/64478

PR #63874, opened 2024-07-23
Add eslint rule for unlabeled RangeControl

mirka:
🚧 Waiting on #63908 to merge

Part of #51740

## What?

Adds an eslint rule for `RangeControl` usage without a `label` prop.

## Why?

This is a quick way to prevent unintentionally unlabeled controls in the Gutenberg app. Unfortunately there are quite a few that get past code review.

Additional components to be added to this lint list should be assessed on a case-by-case basis, but in general I don't foresee too many false positives (e.g. proper usage of a separate `<label>` element).

## Testing Instructions

`npm run lint:js -- --quiet` should surface violations in RangeControl instances.

🚧 Current violations must be fixed before merging

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -278,7 +278,6 @@ module.exports = {
 			},
 		},
 		{
-			// Temporary rules until we're ready to officially deprecate the bottom margins.
 			files: [ 'packages/*/src/**/*.[tj]s?(x)' ],
 			excludedFiles: [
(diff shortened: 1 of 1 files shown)


afercia:
@mirka thanks for working on this ❤️ 
I think one of the two detected violations will be fixed by https://github.com/WordPress/gutenberg/pull/63806
I will look into the other one in `/packages/edit-site/src/components/global-styles/shadows-edit-panel.js`

PR #63881, opened 2024-07-24
Tweak Block Locking UI

t-hamano:
## What?

This PR adjusts the button size and spacing in the Block Locking Modal.

| Before | After |
|--------|--------|
| image | after |

Change the `12px` top and bottom padding applied to the toggle to `16px` top padding.

| Before | After |
|--------|--------|
| image | after_padding |

## Why?

For design consistency.

## Testing Instructions

Open the Block Locking Modal and check the button size and the spacing above and below the toggle.

### DIFF
--- a/packages/block-editor/src/components/block-lock/modal.js
+++ b/packages/block-editor/src/components/block-lock/modal.js
@@ -217,12 +217,20 @@ export default function BlockLockModal( { clientId, onClose } ) {
 					expanded={ false }
 				>
 					<FlexItem>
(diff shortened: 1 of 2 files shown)


jameskoster:
Good change. Should we remove the top margin on `.block-editor-block-lock-modal__options` too? The gap between the modal title + contents is quite large.

t-hamano:
> Should we remove the top margin on `.block-editor-block-lock-modal__options` too?

That's true. I've updated it 👍

image

jameskoster (approved):

Issue #54150, opened 2023-09-04
iframe: document needs language attributes


sabernhardt:
### Description

The iframe editor does not include a `lang` attribute, either for the site's language or for a post's language (which can be different from the main site language on multilingual sites).

https://github.com/WordPress/gutenberg/blob/c7ff22ae513ecc0cdc0bda9c9fffedd53b2aee9b/packages/block-editor/src/components/iframe/index.js#L195

Accessibility Conformance rules may make an [exception for embedded pages](https://www.w3.org/WAI/standards-guidelines/act/rules/bf051a/), but any styles for specific languages cannot be applied without the `lang` attribute inside the `iframe`. This includes font selection in themes such as Twenty Seventeen and Twenty Nineteen (Twenty Twenty's inline style still works, though).

### Step-by-step reproduction instructions

1. Go to the site's General Settings and select English as the language.
2. Activate Twenty Nineteen (or Twenty Seventeen) as your theme.
3. Open an existing post with text. If you do not have any, create a new post and add a paragraph.
4. Use the browser's Inspect Element feature on an element within the post content area.
5. Check the `html` element within the `iframe`. If you do not find an `iframe`, you may need to deactivate plugins. (And if you have the Custom Fields panel showing, you will need to hide that too.)
6. Notice that the `html` does not have a `lang` attribute and the `dir` attribute is empty (instead of `ltr`).
7. Go to the site's General Settings and choose a right-to-left language such as Persian, Hebrew or Arabic.
8. Revisit the earlier post and refresh the browser.
9. Inspect Element and check the `html` element within the `iframe` again. The `html` does not have a `lang` attribute in this case either, but the `dir` attribute can accurately be `rtl` if your profile is set to the site language.
10. View/preview the post to see the theme's intended font for this language, and note that the editor does not match it.
11. Activate Twenty Twenty-Three as your theme.
12. If you would like, change the language.
13. Go to the (site) editor and use the browser's Inspect Element feature to view the `html` element within the `iframe` there. As in the post editor, LTR languages have an empty `dir` attribute, RTL languages have `dir="rtl"` for the direction, and all lack the `lang` attribute.

### Screenshots, screen recording, code snippet

With Twenty Nineteen, Arabic should be in Tahoma (sans-serif), but the iframe editor continues to use the serif typeface (shown in Garamond).

[...]

PR #63883, opened 2024-07-24
Editor: use site locale in editor iframe

### LINKED ISSUE #52777: Block editor canvas should use site language and language direction
### Description

when I'm making a Kurdish or Arabic website, my website shows from LTR in Gutenberg interface but when I visit my website it shows RTL, My user profile language is English and my website language is Kurdish, it is supposed to always shows RTL.

in Gutenberg interface
Change user profile language to English
Change website language to Kurdish or Arabic
then go to Appearance > Editor and you will see the issue 

_No response_

### Environment info

- WordPress: 6.4-alpha-56272
- PHP: 8.1.9
- Server: nginx/1.16.0
- Database: mysqli (Server: 8.0.16 / Client: mysqlnd 8.1.9)
- Browser: Edge 114.0.1823.86 (Windows 10/11)
- Theme: Twenty Twenty-Three 1.1
- MU-Plugins: None activated
- Plugins: 
  * WordPress Beta Tester 3.5.2

Yes

Yes

ramonjd:
## What? How?

(Only partly) fixes https://github.com/WordPress/gutenberg/issues/52777

A first attempt at settings the lang and dir attributes in the editor iframe.

This PR doesn't look at translations of theme templates.

This PR

- creates user and site locale values in block editor settings
- fetches them in the editor iframe
- sets the `lang` and `dir` attributes of the html element in the iframe
- adds the `rtl` classname to the iframe body tag if necessary
- adds a CSS rule so that block placeholders have the right `dir` (the user locale)

## Why?

So that the editor canvas correctly reflects the current site locale and not the user/editor locale.

## Editor canvas

To test, your user locale should be different to the site locale. By default the user locale is set to the site locale.

Edit your user settings over at `/wp-admin/users.php`, and set a LTR language, such as English or German, as your user locale. 

Set an RTL language, such as Arabic or Hebrew,  as the site locale  in /wp-admin/options-general.php

In the Post and Site editors, check that the `<html />` element contains the correct `lang` and `dir` attributes (from the site).

Now swap the languages: use the RTL language as your user locale and the LTR as your site locale.

Check that the `<html />` element contains the correct `lang` and `dir` attributes (from the site).

## Block/pattern previews

This change should also affect block/pattern previews. To test, either check a theme's pattern library in the Site Editor, insert a Query Block and view the suggested patterns, or open the block or pattern library from the block inserter.

Check that the previews have the correct locale attributes. 

## Screenshots or screencast 

User locale: German (de_DE)
Site locale: Hebrew (he_IL)

Screenshot 2024-07-24 at 1 24 37 PM

### Patterns

User locale: English (en_US)
Site locale: Arabic (ar)

Screenshot 2024-07-25 at 11 28 22 AM

## Related

- https://github.com/sabernhardt/gutenberg/pull/2


lib/block-editor-settings.php
```diff
@@ -146,6 +146,47 @@ function gutenberg_get_block_editor_settings( $settings ) {
...
+		if ( is_locale_switched() && $switched_locale ) {
+			$current_site_is_rtl = is_rtl();
+			restore_previous_locale();
+		}
+	}
+
+	$settings['locale'] = array(
+		'site' => array(
+			'lang'  => $current_site_locale,
+			'isRTL' => $current_site_is_rtl,
```
ramonjd: This is a first pass so everything is up for debate. 

Another option is to just pass the site locale data, e.g.,

```php
	$settings['siteLocale'] = array(
        	'lang'  => $current_site_locale,
        	'isRTL' => $current_site_is_rtl,
	);
```

A few things occurred to me:

- `settings.isRTL` already exists. In the editor, it's the current user locale. Should we ignore it, or add related propertes, e.g., `settings.lang`, `settings.siteLang`, and `settings.siteIsRTL`. Seems a bit clunky.
- Rather than `isRTL: boolean`, what about `dir: 'ltr'|'rtl'`?
- Are block editor settings the right place for this data?

packages/block-editor/src/components/iframe/index.js
```diff
@@ -112,14 +112,18 @@ function Iframe( {
...
-		};
-	}, [] );
+	const { resolvedAssets, isPreviewMode, siteLocale } = useSelect(
+		( select ) => {
+			const { getSettings } = select( blockEditorStore );
+			const settings = getSettings();
+			return {
+				resolvedAssets: settings.__unstableResolvedAssets,
+				isPreviewMode: settings.__unstableIsPreviewMode,
+				siteLocale: settings.locale?.site,
```
ramonjd: Pulling it from the store for convenience, however I'm thinking that the local could be passed to the `Iframe` component as a prop from the consuming components.

This would decouple the Iframe from block settings, and means that the locale can be switched programmatically from above. 🤷🏻

packages/block-editor/src/components/iframe/index.js
```diff
@@ -131,83 +135,88 @@ function Iframe( {
...
+							name.startsWith( 'admin-color-' ) ||
+							name.startsWith( 'post-type-' ) ||
+							name === 'wp-embed-responsive'
+					)
+				);
 
-			for ( const compatStyle of getCompatibilityStyles() ) {
-				if ( contentDocument.getElementById( compatStyle.id ) ) {
-					continue;
+				if ( ! siteLocale ) {
```
ramonjd: Just a placeholder condition, and probably only useful if the locale is passed as a prop.

If no external site locale data is returned from the settings, pull the `dir` from the parent as per usual.

ramonjd (commented):
This is just a test and doesn't look at anything aside from passing the correct locale attributes to the editor.

There are a few future rabbit holes. 

For example, if the user locale is LTR, should block placeholders be `direction:ltr`?

Screenshot 2024-07-24 at 1 58 10 PM

Core block assets should be loaded for both RTL and LTR? But isolated to their corresponding window contexts 🙃 

https://github.com/WordPress/gutenberg/blob/af2e21ad7aeac9c834325a23ad77718468c0bf00/lib/blocks.php#L255-L255

I only have a bit of time to devote to this, so happy to have folks nudge the ship along.

packages/block-editor/src/components/iframe/index.js
```diff
@@ -131,83 +135,88 @@ function Iframe( {
...
+				iFrameDocument = contentDocument;
+
+				documentElement.classList.add( 'block-editor-iframe__html' );
+
+				clearerRef( documentElement );
+
+				// Ideally ALL classes that are added through get_body_class should
+				// be added in the editor too, which we'll somehow have to get from
+				// the server in the future (which will run the PHP filters).
+				setBodyClasses(
```
sabernhardt: Could you add the `rtl` body class when the site language direction is RTL?

[My attempt](https://github.com/sabernhardt/gutenberg/pull/2/files#diff-ab17abd38e687edd56573694ae3017e105e96b319571c190e9d4705c1872bc9cR160-R161) for the closely related issue #54150 was simpler but inadequate because it used `ownerDocument` (which has the profile language, not site language).

packages/block-editor/src/components/iframe/index.js
ellatrix: Maybe just remove it here and let's set it in the html element below? Adding dependencies could have some weird effects here, let's avoid that

ellatrix: It's also nice if everything is in one place

packages/block-editor/src/components/iframe/index.js
ellatrix: The iframe component is not a public API I think so it probably doesn't matter.

packages/block-editor/src/components/iframe/index.js
ramonjd: Good call, thanks for the feedback!

packages/block-editor/src/components/iframe/index.js
ramonjd: I'll check it out, thanks for pointing it out.

packages/block-editor/src/components/iframe/index.js
```diff
@@ -280,6 +287,13 @@ function Iframe( {
 				colors the html element */
 				background-color: white;
 			}
+			/* 
+				Block placeholders are "technically" part of the editor UI,
+				and should therefore match the editor UI's directionality.
+			*/
+			.block-editor-block-list__layout .components-placeholder {
```
ramonjd: This change is optional (follow up?), and there's probably a better way.

The CSS rule here is in lieu of setting the `dir` inside the individual blocks.

With a site lang of `ar` and a user lang of `en`:

 | Before | After |
|--------|--------|
| Screenshot 2024-07-25 at 11 27 36 AM | Screenshot 2024-07-25 at 11 27 45 AM | 

With a site lang of `en` and a user lang of `ar:

 | Before | After |
|--------|--------|
|  Screenshot 2024-07-25 at 11 42 05 AM | Screenshot 2024-07-25 at 11 41 36 AM |

lib/block-editor-settings.php
ramonjd: @youknowriad What do you reckon about the locale data model here (and using it in block editor settings)?

Is there a better way to pass site and user locale data to the editor?

I looked at extending the site settings (`/wp/v2/settings`) endpoint, but it appears limited to site option values.

packages/block-editor/src/components/iframe/index.js
ramonjd: Maybe we could compose `<Placeholder />` for blocks and do all the locale trickery in the higher order component?

E.g., `const BlockPlaceholder = withLocale( Placeholder )`, or whatever it is

lib/block-editor-settings.php
```diff
@@ -146,6 +146,47 @@ function gutenberg_get_block_editor_settings( $settings ) {
...
+		),
+		'user' => array(
+			'lang'  => $current_user_locale,
+			'isRTL' => $current_user_is_rtl,
+		),
+	);
+
 	return $settings;
 }
 add_filter( 'block_editor_settings_all', 'gutenberg_get_block_editor_settings', 0 );
```
youknowriad: Why is this hook in `lib/block-editor-settings.php` and not in `wp/6.7` folder...

lib/block-editor-settings.php
youknowriad: > I looked at extending the site settings (/wp/v2/settings) endpoint, but it appears limited to site option values.

This would have been my preference personally as this is just a site option (whether it's stored as an option or not is not really important IMO). cc @TimothyBJacobs maybe

packages/block-editor/src/components/iframe/index.js
```diff
@@ -112,14 +112,19 @@ function Iframe( {
...
 } ) {
-	const { resolvedAssets, isPreviewMode } = useSelect( ( select ) => {
-		const { getSettings } = select( blockEditorStore );
-		const settings = getSettings();
-		return {
-			resolvedAssets: settings.__unstableResolvedAssets,
-			isPreviewMode: settings.__unstableIsPreviewMode,
-		};
-	}, [] );
+	const { resolvedAssets, isPreviewMode, siteLocale, userLocale } = useSelect(
```
youknowriad: Why do we need both siteLocale and userLocale? For me userLocale is just `isRTL` so we don't really need a new setting there.

youknowriad (commented):
Do we have an idea about the classic editor and how it handles this case?

Also the other questions I have (maybe things to address separately):

 - we have block styles `style.css` and `editor.css` and `theme.css` and If I'm not wrong each one of these has "rtl" versions. I'm guessing style and theme need to use the "site locale" but it's unclear what "editor" should be using.
 - Do we have any support for automatic RTL for theme.json generated styles and how does it impact this PR?
 - What about the uniframed block editor, do we ignore that and consider that most people should migrate to iframed editor eventually?

packages/block-editor/src/components/iframe/index.js
ellatrix: I don't think this is resolved yet?

packages/block-editor/src/components/iframe/index.js
```diff
@@ -264,7 +271,7 @@ function Iframe( {
 	// mode. Also preload the styles to avoid a flash of unstyled
 	// content.
 	const html = `<!doctype html>
-<html>
+<html lang="${ siteLocale?.lang }" dir="${ siteLocale?.isRTL ? 'rtl' : 'ltr' }">
```
ellatrix, line 274: Here's another idea: what if the initial HTML as a whole is a setting? This all of this stuff is customisable from outside the iframe component.

lib/block-editor-settings.php
ramonjd: No reason, mainly due to being distracted by the bigger questions :) I'll move it. 👍🏻

packages/block-editor/src/components/iframe/index.js
ramonjd: Good question.

The block editor should know about both site locale and user locale because:

1. it needs both the site lang and site RTL value to add the requisite attributes to the document (so that the content matches the frontend)
2. it should have the user locale available in case the editor canvas needs to render any editor-specific UI, such as placeholders.

On point 2, I could be wrong in my assumption, but my instinct tells me that the block placeholders are part of the editor UI, and therefore, the locales should be consistent.

 | Before | After |
|--------|--------|
| Screenshot 2024-07-25 at 11 27 36 AM | Screenshot 2024-07-25 at 11 27 45 AM | 

The iframe may not need to access userLocale here at all, but here I just wanted to provide an example of how to make the placeholder UI conform to the user locale (and not the site). See [above](https://github.com/WordPress/gutenberg/pull/63883#discussion_r1692110347).

I'm thinking a HOC placeholder in block-editor would be a better place to add the `lang` and `dir` attributes to the `<Placeholders />`, not the funky CSS I have in the latest commit of this PR. Anyway, that would be a follow up.

packages/block-editor/src/components/iframe/index.js
ramonjd: Oh I was playing with body classes and forgot I didn't push the commit removing the dependency. Thanks for flagging.

lib/block-editor-settings.php
ramonjd: Yeah, it'd be a neat place to house this stuff.

In that endpoint, we'd add a `is_rtl: true` or `dir: 'rtl'` or whatever, that corresponds to the **site locale**.

The only conceptual agony I was going through was that this value can't be updated in the POST like the others, so it isn't neat, but the world is messy so I can get over it. Keen to hear what Timothy thinks.

Here's the output of the GET request the record:

`await wp.data.resolveSelect( 'core' ).getSite()`

<details>

<summary>Respone</summary>

```json

{
    "title": "gutenberg",
    "description": "",
    "url": "http://localhost:8888",
    "email": "wordpress@example.com",
    "timezone": "",
    "date_format": "F j, Y",
    "time_format": "g:i a",
    "start_of_week": 1,
    "language": "ar",
    "use_smilies": true,
    "default_category": 1,
    "default_post_format": "0",
    "posts_per_page": 10,
    "show_on_front": "posts",
    "page_on_front": 0,
    "page_for_posts": 0,
    "default_ping_status": "open",
    "default_comment_status": "open",
    "site_logo": null,
    "site_icon": 0
}

```

</details>

packages/block-editor/src/components/iframe/index.js
youknowriad: No, my question was simpler than that :P. We already have `wp.i18n.isRTL` function to know whether the user local is rtl or not, we don't need a new setting for that.

ramonjd:
> Do we have an idea about the classic editor and how it handles this case?

It uses the user locale as far as I can see. There are a couple of related tickets: 

See:

- https://core.trac.wordpress.org/ticket/48996
- https://core.trac.wordpress.org/ticket/40715

> we have block styles style.css and editor.css and theme.css and If I'm not wrong each one of these has "rtl" versions. I'm guessing style and theme need to use the "site locale" but it's unclear what "editor" should be using.
> Do we have any support for automatic RTL for theme.json generated styles and how does it impact this PR?

I was discussing the RTL CSS generation/enqueuing question with @andrewserong today, and yeah, it's something that needs to be address, but in a follow up. 

In my mind whatever is on the editor canvas — the content that represents what we see on the frontend — should follow the site locale.

The editor UI should, in theory, do what it does now and use the current user locale (if set).


This PR deals with HTML semantics pretty much.


> What about the uniframed block editor, do we ignore that and consider that most people should migrate to iframed editor eventually?

I thought about that for 1 second, and concluded it should follow whatever the "app" does, so the user locale as the primary fallback.

All the above sounds easy when I write it down, but I haven't looked too deeply into the how yet. 

I haven't had much time to spend on this, which probably shows, but I think it's important enough to fix so I'll plug away at it when I can.

youknowriad:
Yeah, for me the most challenging thing is going to be the CSS loading and how to choose whether we pick RTL or not ... As it's not clear for some third-party CSS files whether they're targetting content or placeholders...

packages/block-editor/src/components/iframe/index.js
ramonjd: You know I love typing :trollface: 

I think I [remember realizing we just needed the site locale.](https://github.com/WordPress/gutenberg/pull/63883#discussion_r1689062491)

So, yeah, good point. "Locale" comprises language too, but `wp.i18n.getLocaleData()[""].lang` should get us the value for the lang attribute too should we need it. 👍🏻

lib/compat/wordpress-6.7/block-editor.php
```diff
@@ -0,0 +1,57 @@
...
+
+	if ( $current_user_locale !== $current_site_locale ) {
+		$switched_locale = switch_to_locale( $current_site_locale );
+		if ( is_locale_switched() && $switched_locale ) {
+			$current_site_is_rtl = is_rtl();
+			restore_previous_locale();
+		}
+	}
+
+	// @TODO confirm if `lang` just needs to be added to the `/wp/v2/settings` response`.
```
swissspidy: `language` is actually already part of that.

lib/compat/wordpress-6.7/block-editor.php
```diff
@@ -0,0 +1,57 @@
...
+	$current_user_locale = get_user_locale();
+	$current_user_is_rtl = is_rtl();
+
+	// Current site locale.
+	$current_site_locale = get_locale();
+	$current_site_is_rtl = $current_user_is_rtl;
+
+	if ( $current_user_locale !== $current_site_locale ) {
+		$switched_locale = switch_to_locale( $current_site_locale );
+		if ( is_locale_switched() && $switched_locale ) {
```
swissspidy: You don't need `is_locale_switched()`. Just the return value of `$switched_locale` is enough.


```suggestion
		if ( $switched_locale ) {
```

ramonjd: Oh great, thanks!

lib/compat/wordpress-6.7/block-editor.php
ramonjd: Sorry. That was a typo on my part. It should be "confirm if `is_rtl`" pursuant to [this comment](https://github.com/WordPress/gutenberg/pull/63883#discussion_r1692852562)

lib/block-editor-settings.php
TimothyBJacobs: This is exactly what the index is really for. The `rest_index` filter can be used.

ramonjd: Thanks for confirming @TimothyBJacobs I'll take a look at using the [filter](https://developer.wordpress.org/reference/hooks/rest_index/) 🙇🏻

senadir:
Wondering what's needed to bring this closer to finish line.

ramonjd:
> Wondering what's needed to bring this closer to finish line.

Probably a way to ensure templates/template parts and other site content is correctly translated into the site language.

I'm not sure the best way to go about it. Maybe sending a `_locale_ value to the APIs.

Also, loading any `-rtl` css into the iframe.

I don't have much time to carry on with this unfortunately, but will keep it on the back burner unless someone decides to pick it up.

ramonjd:
Closing for now. I'll come back later if this feature is required and time permits.

PR #63884, opened 2024-07-24
Fix Custom Select describedBy

### LINKED ISSUE #63568: CustomSelectControl: fix `describedBy` implementation and docs
### Description
Discovered while investigating https://github.com/WordPress/gutenberg/issues/63483

The `CustomSelectControl` component has a `describedBy` prop that accepts a string.

It is supposed to provide an accessible description for the toggle button that opens the custom select.

Unfortunately, both the implementation and the docs are wrong and the component renders non-accessible markup.

An example can be seen for the font appearance setting. It can be reproduced in the Storybook as well.

Turns out the value passed to the `describedBy` prop is just rendered as the value of the `aria-describedby` attribute.

https://github.com/WordPress/gutenberg/blob/c54c0020a7a8ebf22738aa96cb398ea4cf17a30a/packages/components/src/custom-select-control/index.tsx#L157

Instead, the `aria-describedby` attribute accepts an ID reference that points to an element containing the description.

For example, the font appearance setting renders this markup:

`aria-describedby="Currently selected font appearance: Medium"`

The correct usage would be:

```
... aria-describedby="an-unique-id-here" ...

<div hidden id="an-unique-id-here">Currently selected font appearance: Medium</div>

```

The component should be refactored to actually render another element that contains the description, referenced by an ID, much like what [the Button component](https://github.com/WordPress/gutenberg/blob/c54c0020a7a8ebf22738aa96cb398ea4cf17a30a/packages/components/src/button/index.tsx) does.

At that point, the `describedBy` prop should be renamed to `description`.

The readme and types.ts must be fixed as well. For example, this prop description doesn't appear to be correct:
> Description for the select trigger button used by assistive technology. If no value is passed, the text "Currently selected: selectedItem.name" will be used fully translated.

- Select a Group block
- Go to the block Inspector > Styles, and activate the Appearance settings in the Typography options.
- Set a font appearance value e.g. 'Medium',
- Inspect the CustomSelectControl toggle button source.
- Oberve it has an incorrect `aria-describedby="Currently selected font appearance: Medium"`
- Alternatively go to the Storybook: https://wordpress.github.io/gutenberg/?path=/docs/components-customselectcontrol--docs
- Set the `describedBy` prop to `la la la`.
- Inspect the CustomSelectControl toggle button source.
- Oberve it has an incorrect `aria-describedby="la la la"`

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

rohitmathur-7:
## What?
Fix describedBy implementation of CustomSelectControl.
Fixes https://github.com/WordPress/gutenberg/issues/63568

## Why?
To Correct the implementation of aria-describedby of CustomSelectControl for Accessibility.

## How?
Added a div element that contains the describedBy description and give unique ID to this, this unique ID is then added to `aria-describedby` to reference the div.

## Testing Instructions
Select a Group block
Go to the block Inspector > Styles, and activate the Appearance settings in the Typography options.
Set a font appearance value e.g. 'Medium',
Inspect the CustomSelectControl toggle button source.

## Screenshots or screencast 
aria-describedby

### DIFF
--- a/packages/components/src/custom-select-control-v2/custom-select.tsx
+++ b/packages/components/src/custom-select-control-v2/custom-select.tsx
@@ -8,6 +8,7 @@ import * as Ariakit from '@ariakit/react';
  */
 import { createContext, useCallback, useMemo } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
+import { useInstanceId } from '@wordpress/compose';
 
 /**
  * Internal dependencies
@@ -95,6 +96,7 @@ function _CustomSelect(
 		store,
 		className,
 		isLegacy = false,
+		describedBy,
 		...restProps
 	} = props;
 
@@ -109,6 +111,8 @@ function _CustomSelect(
 		);
 
 	const contextValue = useMemo( () => ( { store, size } ), [ store, size ] );
+	const instanceId = useInstanceId( _CustomSelect );
+	const describedByID = `custom-select-described-by-${ instanceId }`;
 
 	return (
 		// Where should `restProps` be forwarded to?
@@ -138,7 +142,13 @@ function _CustomSelect(
 					store={ store }
 					// Match legacy behavior (move selection rather than open the popover)
 					showOnKeyDown={ ! isLegacy }
+					aria-describedby={ describedBy ? describedByID : undefined }
 				/>
+				{ describedBy && (
+					<div id={ describedByID } hidden>
+						{ describedBy }
+					</div>
+				) }
 				<Styled.SelectPopover
 					gutter={ 12 }
 					store={ store }
--- a/packages/components/src/custom-select-control-v2/types.ts
+++ b/packages/components/src/custom-select-control-v2/types.ts
@@ -77,6 +77,10 @@ export type _CustomSelectProps = CustomSelectButtonProps & {
 	 * Accessible label for the control.
 	 */
 	label: string;
+	/**
+	 * Described by.
+	 */
+	describedBy?: string;
 };
 
 export type CustomSelectProps = _CustomSelectProps & CustomSelectSize;
--- a/packages/components/src/custom-select-control/index.tsx
+++ b/packages/components/src/custom-select-control/index.tsx
@@ -154,7 +154,6 @@ function CustomSelectControl( props: CustomSelectProps ) {
 
 	return (
 		<_CustomSelect
-			aria-describedby={ describedBy }
 			renderSelectedValue={
 				showSelectedHint ? renderSelectedValueHint : undefined
 			}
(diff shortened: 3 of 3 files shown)


packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -109,6 +111,8 @@ function _CustomSelect(
 		);
 
 	const contextValue = useMemo( () => ( { store, size } ), [ store, size ] );
+	const instanceId = useInstanceId( _CustomSelect );
+	const describedByID = `custom-select-described-by-${ instanceId }`;
```
afercia, line 115: I think useInstanceId accepts a second parameter that is then used as the prefix so this coul be simplified, see for example https://github.com/WordPress/gutenberg/blob/3dfdfef3eb5cf35abf67bdce9013cc8cef5927b7/packages/components/src/box-control/index.tsx#L44

packages/components/src/custom-select-control-v2/custom-select.tsx
```diff
@@ -138,7 +142,13 @@ function _CustomSelect(
 					store={ store }
 					// Match legacy behavior (move selection rather than open the popover)
 					showOnKeyDown={ ! isLegacy }
+					aria-describedby={ describedBy ? describedByID : undefined }
 				/>
+				{ describedBy && (
+					<div id={ describedByID } hidden>
+						{ describedBy }
```
afercia, line 149: I would like this prop to be renamed because it's not a 'describedBy`. Actually, it's a description i.e. the actual text for the description. For an example of how to rename, use an 'alias', and deprecate the old name, see https://github.com/WordPress/gutenberg/pull/63486

packages/components/src/custom-select-control-v2/types.ts
```diff
@@ -77,6 +77,10 @@ export type _CustomSelectProps = CustomSelectButtonProps & {
 	 * Accessible label for the control.
 	 */
 	label: string;
+	/**
+	 * Described by.
```
afercia, line 81: Please use a more descriptive description of this prop. Something like `An accessible description for the control` could work.

afercia (changes requested):
Thanks for your PR @rohitmathur-7 !
From an accessibility perspective looks good to me.
I left some comments, more importantly about renaming the prop and deprecating the old name. Thanks!

mirka:
Thanks for contributing 👋 Please see [this comment](https://github.com/WordPress/gutenberg/issues/63568#issuecomment-2251042931) on how this needs to be addressed. This PR is basically adding a new feature to CustomSelectControlV2, while in fact what we need is to fix the behavior on CustomSelectControl (v1) only.

Unfortunately I think I will have to take over on this issue, because this is a regression that should ideally be fixed by the next plugin release 🙇

Mamaduka:
It looks like the linked issue was fixed via #63957. I'm going to close the PR.

Thanks for contributing, @rohitmathur-7!

PR #63885, opened 2024-07-24
Image: Make Placeholder white when there is a <Spinner /> on top

noisysocks:
## What?
Updates the design of the `Placeholder` component when passed the `hasIllustration` flag.

## Why?
Moving to a darker design based on the UI colour so that you can more easily see any `Spinner` placed on top. See https://github.com/WordPress/gutenberg/pull/63643#discussion_r1684103899.

## How?
I did what @jameskoster suggested in https://github.com/WordPress/gutenberg/pull/63643#discussion_r1684103899 😀

I also removed the `placeholder-style` mixin because it isn't really doing anything in most of the places it's used. The mixin adds `&::before` styling and we're often including it into a `&::before` or `&::after` which means there is no effect.

As far as I can tell this styling is only used in `<Placeholder hasIllustration>` which is the media blocks when not selected or when uploading.

## Testing Instructions
If you make these changes locally you can force the `Spinner` to always appear:

```diff
diff --git a/packages/block-library/src/image/image.js b/packages/block-library/src/image/image.js
index 385a740e652..3417b1fc1a8 100644
--- a/packages/block-library/src/image/image.js
+++ b/packages/block-library/src/image/image.js
@@ -827,7 +827,7 @@ export default function Image( {
 	const isRounded = attributes.className?.includes( 'is-style-rounded' );
 
 	let img =
-		temporaryURL && hasImageErrored ? (
+		true || ( temporaryURL && hasImageErrored ) ? (
 			// Show a placeholder during upload when the blob URL can't be loaded. This can
 			// happen when the user uploads a HEIC image in a browser that doesn't support them.
 			<Placeholder
```

Then:

1. Add an image block.
2. De-select it. You'll see the placeholder there.
3. Upload an image into the image block.
4. You'll see the placeholder there with a `Spinner` on top.

## Screenshots or screencast 

| Before | After |
|--------|--------|
| Screenshot 2024-07-24 at 15 06 02 | Screenshot 2024-07-24 at 15 04 42 |

### DIFF
--- a/packages/base-styles/_mixins.scss
+++ b/packages/base-styles/_mixins.scss
@@ -162,22 +162,6 @@
 	}
 }
 
-@mixin placeholder-style() {
-	border-radius: $radius-block-ui;
-
-	&::before {
-		content: "";
-		position: absolute;
-		top: 0;
-		right: 0;
-		bottom: 0;
-		left: 0;
-		pointer-events: none;
-		background: currentColor;
-		opacity: 0.1;
-	}
-}
-
 /**
  * Allows users to opt-out of animations via OS-level preferences.
  */
--- a/packages/block-editor/src/components/button-block-appender/content.scss
+++ b/packages/block-editor/src/components/button-block-appender/content.scss
@@ -52,7 +52,7 @@
 			left: 0;
 			pointer-events: none;
 			border: $border-width dashed currentColor;
-			@include placeholder-style();
+			border-radius: $radius-block-ui;
 		}
 
 		.block-editor-inserter {
--- a/packages/block-library/src/group/editor.scss
+++ b/packages/block-library/src/group/editor.scss
@@ -43,7 +43,7 @@
 		pointer-events: none;
 		min-height: $grid-unit-60 - $border-width - $border-width;
 		border: $border-width dashed currentColor;
-		@include placeholder-style();
+		border-radius: $radius-block-ui;
 	}
 
 	// Let the parent be selectable in the placeholder area.
--- a/packages/block-library/src/image/editor.scss
+++ b/packages/block-library/src/image/editor.scss
@@ -80,6 +80,11 @@ figure.wp-block-image:not(.wp-block) {
 		background: $white;
 		opacity: 0.8;
 	}
+
+	// Reduce opacity of diagonal stroke so that it doesn't obscure the spinner.
+	.components-placeholder__illustration {
+		opacity: 0.1;
+	}
 }
 
 // This is necessary for the editor resize handles to accurately work on a non-floated, non-resized, small image.
--- a/packages/block-library/src/image/editor.scss
+++ b/packages/block-library/src/image/editor.scss
@@ -74,6 +74,12 @@ figure.wp-block-image:not(.wp-block) {
 // Shown while image is being uploded but cannot be previewed.
 .wp-block-image__placeholder {
 	aspect-ratio: 4 / 3;
+
+	// Make placeholder background white so that you can see the spinner on top of it.
+	&.has-illustration::before {
+		background: $white;
+		opacity: 0.8;
+	}
 }
 
 // This is necessary for the editor resize handles to accurately work on a non-floated, non-resized, small image.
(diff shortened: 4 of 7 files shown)


jameskoster:
I flip-flopped back to a white overlay, the contrast against the gray spinner track is less harsh. Also pushed a couple other style tweaks, and removed a margin that was making the spinner slightly off-center.

I'm a bit confused by the behavior though. I'm seeing the overlay without a spinner when an image hasn't been uploaded. Shouldn't both the overlay and the spinner become visible whilst the image is being uploaded?

noisysocks:
The overlay will show when you have not picked an image and you deselect the block.

A faded out image + spinner will show while an image is being uploaded.

The overlay + spinner will show while an image that isn't supported by your browser (e.g. HEIC) is being uploded.

Does that make sense?

---

Thanks for pushing. The diagonal line isn't appearing on the overlay anymore because of the `z-index`:

Screenshot 2024-07-25 at 11 35 44

I imagine that's a mistake? I'll remove the `z-index`es (which I think are unnecessary?) which makes it look like this:

Screenshot 2024-07-25 at 11 35 48

---

Here's how everything looks together:

https://github.com/user-attachments/assets/29bb78e0-319f-412b-96cb-eccbd88ffd37

And here's how the placeholder looks like when it's on top of a background image:

Screenshot 2024-07-25 at 11 45 56

I like it! Add an approving review if you're happy 🙂

jameskoster:
> The overlay will show when you have not picked an image and you deselect the block.

Hmm, I don't understand why we're showing an overlay when the block is deselected. What purpose is that serving? I'd expect to see the standard placeholder with no overlay, this one:

Screenshot 2024-07-25 at 09 54 26

My understanding was that the overlay's purpose was to ensure the spinner was visible?

The `z-index` was to ensure the overlay and spinner appear 'above' the placeholder. Previously the placeholder was appearing above the overlay which seemed wrong.

Here's what I'd expect:

upload

Does that make sense?

noisysocks:
Sorry, think there's some confusion. What do you mean by overlay? In [my comment above](https://github.com/WordPress/gutenberg/pull/63885#issuecomment-2249197772) I used overlay as a synonym for the `<Placeholder />`.

When an image that the browser can render (JPEG) is being uploaded, we show that image with the `is-transient` class which sets the opacity of the image to 0.3 with a `<Spinner />` on top.

When an image that the browser can't render (HEIC) is being uploaded, we show a `<Placeholder hasIllustration />` with a `<Spinner />` on top.

When an Image block with no image selected is deselected, we show a `<Placeholder hasIllustration />` by itself.

When `<Placeholder>` has the `hasIllustration` prop, it shows a diagonal line and we place some `&::before` styling on it that gives the `Placeholder` a solid background. 

So I think things are currently working as you expect? There's no overlay on top of the `<Placeholder />` or image. The `<Spinner />` appears either directly on top of the `<Placeholder />` or directly on top of the image with its opacity reduced. This PR changes the `<Placeholder />` to be a colour that doesn't make it too hard to see any `<Spinner />` that is on top.

jameskoster:
Ah I see, thanks for clarifying. Sorry about the confusion, that's my fault. I didn't realise the overlay and the placeholder were the same element. 

The part that doesn't seem to be working as expected is that the white background should only applied when the spinner is visible. Currently it seems to be applied regardless. 

If it's not trivial to do that, then it's probably okay to just leave things as they are (but include the margin fix).

noisysocks:
> The part that doesn't seem to be working as expected is that the white background should only applied when the spinner is visible. Currently it seems to be applied regardless.
> 
> If it's not trivial to do that, then it's probably okay to just leave things as they are (but include the margin fix).

👍 you got it, done in b54c323.

jameskoster:
Bingo, here's what I see:

Screenshot 2024-07-30 at 13 04 30

The only other minor tweak I'd suggest (if it's possible) would be to reduce the opacity of the diagonal stroke, it's obscuring the spinner a little. Otherwise this seems good.

jameskoster (approved):

noisysocks:
> The only other minor tweak I'd suggest (if it's possible) would be to reduce the opacity of the diagonal stroke, it's obscuring the spinner a little. Otherwise this seems good.

Done 👍

PR #63889, opened 2024-07-24
Pages: update `useView` logic

oandregal:
Alternative to https://github.com/WordPress/gutenberg/pull/63854
Raised at https://github.com/WordPress/gutenberg/pull/63203#issuecomment-2244993764

## What?

This PR updates the `useView` hook to do the following:

- Changing layout doesn't reset any other view config.

https://github.com/user-attachments/assets/e086185f-f803-41b8-9058-fc321d93228f

- Selecting a different default or custom view resets everything, including layout choices.

https://github.com/user-attachments/assets/a5b4c23e-949e-4ba6-b718-6dae26879547

- Loading a custom view displays it with the proper layout. In trunk, it displays `list`, no matter the custom view layout.

https://github.com/user-attachments/assets/4525745e-7775-4c7b-afc7-ae2927c59931

- Loading a unknown activeView, loads an empty DataView (in trunk, the site editor doesn't load). `http://localhost:8888/wp-admin/site-editor.php?postType=page&activeView=wrong-key`

https://github.com/user-attachments/assets/44625f6c-0812-4938-9395-8557e5989d61

## Testing Instructions

Verify those work as expected and test default/custom views in general.

### DIFF
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -44,13 +44,12 @@ function useView( postType ) {
 		params: { activeView = 'all', isCustom = 'false', layout },
 	} = useLocation();
 	const history = useHistory();
-	const DEFAULT_VIEWS = useDefaultViews( { postType } );
+
+	const defaultViews = useDefaultViews( { postType } );
 	const selectedDefaultView = useMemo( () => {
 		const defaultView =
 			isCustom === 'false' &&
-			DEFAULT_VIEWS[ postType ].find(
-				( { slug } ) => slug === activeView
-			)?.view;
+			defaultViews.find( ( { slug } ) => slug === activeView )?.view;
 		if ( isCustom === 'false' && layout ) {
 			return {
 				...defaultView,
@@ -59,7 +58,7 @@ function useView( postType ) {
 			};
 		}
 		return defaultView;
-	}, [ isCustom, activeView, layout, postType, DEFAULT_VIEWS ] );
+	}, [ isCustom, activeView, layout, defaultViews ] );
 	const [ view, setView ] = useState( selectedDefaultView );
 
 	useEffect( () => {
@@ -131,8 +130,9 @@ function useView( postType ) {
 	} else if ( isCustom === 'true' && customView ) {
 		return [ customView, setCustomView ];
 	}
-	// Loading state where no the view was not found on custom views or default views.
-	return [ DEFAULT_VIEWS[ postType ][ 0 ].view, setDefaultViewAndUpdateUrl ];
+
+	// No view was found.
+	return [ defaultViews[ 0 ].view, setDefaultViewAndUpdateUrl ];
 }
 
 const DEFAULT_STATUSES = 'draft,future,pending,private,publish'; // All but 'trash'.
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -26,7 +26,6 @@ import {
 	OPERATOR_IS_ANY,
 	OPERATOR_IS_NONE,
 	LAYOUT_LIST,
-	LAYOUT_TABLE,
 } from '../../utils/constants';
 
 import AddNewPostModal from '../add-new-post';
@@ -97,13 +96,13 @@ function useView( postType ) {
 		if ( isCustom === 'true' ) {
 			return (
 				getCustomView( editedEntityRecord ) ?? {
-					type: layout ?? LAYOUT_TABLE,
+					type: layout ?? LAYOUT_LIST,
 				}
 			);
 		}
 		return (
 			getDefaultView( defaultViews, activeView ) ?? {
-				type: layout ?? LAYOUT_TABLE,
+				type: layout ?? LAYOUT_LIST,
 			}
 		);
 	} );
@@ -132,13 +131,16 @@ function useView( postType ) {
 				);
 			}
 		},
-		[ history, isCustom, editedEntityRecord?.id ]
+		[ history, isCustom, editEntityRecord, editedEntityRecord?.id ]
 	);
 
 	// When layout URL param changes, update the view type
 	// without affecting any other config.
 	useEffect( () => {
-		setView( ( prevView ) => ( { ...prevView, type: layout } ) );
+		setView( ( prevView ) => ( {
+			...prevView,
+			type: layout ?? LAYOUT_LIST,
+		} ) );
 	}, [ layout ] );
 
 	// When activeView or isCustom URL parameters change,
@@ -156,7 +158,7 @@ function useView( postType ) {
 		}
 	}, [ activeView, isCustom, defaultViews, editedEntityRecord ] );
 
-	return [ view, setViewWithUrlUpdate ];
+	return [ view, setViewWithUrlUpdate, setViewWithUrlUpdate ];
 }
 
 const DEFAULT_STATUSES = 'draft,future,pending,private,publish'; // All but 'trash'.
--- a/packages/edit-site/src/components/post-list/index.js
+++ b/packages/edit-site/src/components/post-list/index.js
@@ -26,6 +26,7 @@ import {
 	OPERATOR_IS_ANY,
 	OPERATOR_IS_NONE,
 	LAYOUT_LIST,
+	LAYOUT_TABLE,
 } from '../../utils/constants';
 
 import AddNewPostModal from '../add-new-post';
@@ -39,6 +40,36 @@ const { useLocation, useHistory } = unlock( routerPrivateApis );
 const { useEntityRecordsWithPermissions } = unlock( coreDataPrivateApis );
 const EMPTY_ARRAY = [];
 
+const getDefaultView = ( defaultViews, activeView ) => {
+	return defaultViews.find( ( { slug } ) => slug === activeView )?.view;
+};
+
+const getCustomView = ( editedEntityRecord ) => {
+	if ( ! editedEntityRecord?.content ) {
+		return undefined;
+	}
+
+	const content = JSON.parse( editedEntityRecord.content );
+	if ( ! content ) {
+		return undefined;
+	}
+
+	return {
+		...content,
+		layout: defaultLayouts[ content.type ]?.layout,
+	};
+};
+
+/**
+ * This function abstracts working with default & custom views by
+ * providing a [ state, setState ] tuple based on the URL parameters.
+ *
+ * Consumers use the provided tuple to work with state
+ * and don't have to deal with the specifics of default & custom views.
+ *
+ * @param {string} postType Post type to retrieve default views for.
+ * @return {Array} The [ state, setState ] tuple.
+ */
 function useView( postType ) {
 	const {
 		params: { activeView = 'all', isCustom = 'false', layout },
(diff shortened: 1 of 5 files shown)


packages/edit-site/src/components/post-list/index.js
```diff
@@ -55,14 +67,15 @@ function useView( postType ) {
 			};
 		}
 		return defaultView;
-	}, [ isCustom, activeView, layout, postType, DEFAULT_VIEWS ] );
-	const [ view, setView ] = useState( selectedDefaultView );
-
+	}, [ activeView ] );
```
oandregal: Unless I'm missing some use case, we shouldn't care about any URL change other than `activeView`. The goal here is to replace the view completely when the `activeView` changes (the effect below), but ignore any other change.

youknowriad: I'm not sure that's correct, what happens if there's an action that programmatically changes the url or things like that.

oandregal: I'm not sure I follow. Shouldn't we only reset the full view when the `activeView` changes? That's what is telling us to "load a different default view". How does it matter if it's updated manually or automatically?

youknowriad: What I'm saying is that if the "layout" change in the url from another place than the current component, the current component should adapt. (For instance a command palette that switches a layout).  Your comment seem to suggest that it won't.

oandregal: Ah, I see your point now. So, we have two things to care about:

- reset the view when the user selects one of the default views
- update the view when layout is updated programatically

I pushed https://github.com/WordPress/gutenberg/pull/63889/commits/4bf53ef0b60eee7d1b73cab1c7ac68b5425e259e to address the second concern. It's a separate effect that selectively updates the view (only `view.type` changes, we don't reset any other thing).

I also realized that we still need to reset the view completely in the case `postType` or `isCustom` change programatically, so I brought back those as well.

packages/edit-site/src/components/post-list/index.js
```diff
@@ -41,6 +41,22 @@ function useView( postType ) {
...
+	 *
+	 * - activeView: the identifier. A number for custom views, a string for the default/bundled ones.
+	 * - isCustom: whether the view is a custom/user-defined or not.
+	 * - postType: the post type the view is for.
+	 *
+	 * Every time one of these parameters change, we reset the view completely.
+	 *
+	 * That's why this memoitization doesn't care about any other parameter (layout).
+	 * See https://github.com/WordPress/gutenberg/pull/63889
+	 */
```
youknowriad: This probably triggers an eslint warning of a missing dependency. I think if we follow what I suggest here https://github.com/WordPress/gutenberg/pull/63854#discussion_r1688187227 there's a way to rework this code to be clearer.

Basically:

```
const [ view, setView ] = useState( { type: layout, ...getDefaultView( isCustom, activeView, postType ) } );

useEffect( () => {
   setView( ( prevView ) => ( { ...prevView, type: layout } ) );
}, [ layout ] );

useEffect(() => {
   setView( prev => ({ ...prev, ...getDefaultView( isCustom, activeView, postType ) });
}, [ isCustom, activeView, postType ] );
```

where `getDefaultView` returns the default view but without any "type" property. I think that code speaks for itself and clarifies the intent of any effect...

packages/edit-site/src/components/post-list/index.js
jorgefilipecosta: Hi @oandregal, We also need to take into account the layout URL change. The trash view uses layout table instead of list.

packages/edit-site/src/components/post-list/index.js
jorgefilipecosta: I think this logic may work well, the activeView and layout may change at the same time but I think when an effect runs it changes the view directly, and on the setView, the prev already reflects the changes of the previous effect.

packages/edit-site/src/components/post-list/index.js
```diff
@@ -55,14 +71,21 @@ function useView( postType ) {
 			};
 		}
 		return defaultView;
-	}, [ isCustom, activeView, layout, postType, DEFAULT_VIEWS ] );
-	const [ view, setView ] = useState( selectedDefaultView );
-
+	}, [ isCustom, activeView, postType ] );
```
jorgefilipecosta: The effect still depends on the layout I guess if we follow this we will need a lint ignore.

jorgefilipecosta (commented):
I'm getting a crash when oppening the pages list, I think it is because the setView is defined after being used in the effects.

oandregal:
I've updated this PR based on this [slack thread](https://wordpress.slack.com/archives/C02QB2JS7/p1722331514967269).

In addition to prevent view config resets when the layout changes, this PR also changes user behavior: when selecting one of the default views, the layout is reset to the preferred by the default view. This change will need to be done in Templates & Patterns as well.

I haven't yet tested the custom views, but wanted to share what I have for thoughts.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
```diff
@@ -77,105 +77,105 @@ export function useDefaultViews( { postType } ) {
...
-						...DEFAULT_POST_BASE,
-						filters: [
-							{
-								field: 'status',
-								operator: OPERATOR_IS_ANY,
-								value: 'publish',
-							},
-						],
-					},
+		return [
```
oandregal, line 80: The main change here is removing the `[ postType ]` key.

packages/edit-site/src/components/sidebar-dataviews/default-views.js
```diff
@@ -77,105 +77,105 @@ export function useDefaultViews( { postType } ) {
...
-						],
-					},
+			},
+			{
+				title: __( 'Trash' ),
+				slug: 'trash',
+				icon: trash,
+				view: {
+					...DEFAULT_POST_BASE,
+					type: LAYOUT_TABLE,
```
oandregal, line 168: I also added `type` and `layout` to this default view, to account for switching the trash view to the table layout. This was previously done with the `useSwitchToTableOnTrash` hook.

packages/edit-site/src/components/post-list/index.js
```diff
@@ -99,40 +122,29 @@ function useView( postType ) {
...
 				{
-					content: JSON.stringify( viewToSet ),
+					content: JSON.stringify( newView ),
 				}
 			);
 		},
 		[ editEntityRecord, editedViewRecord?.id ]
 	);
 
-	const setDefaultViewAndUpdateUrl = useCallback(
```
oandregal, line 115: This is moved above for readibility.

packages/edit-site/src/components/post-list/index.js
```diff
@@ -39,34 +39,57 @@ const { useLocation, useHistory } = unlock( routerPrivateApis );
...
-		if ( selectedDefaultView ) {
-			setView( selectedDefaultView );
+		if ( isCustom === 'false' ) {
+			const defaultView = getDefaultView( defaultViews, activeView );
+			setView( defaultView );
+
+			const { params } = history.getLocationWithParams();
+			history.push( {
+				...params,
+				layout: defaultView.type,
```
youknowriad: For me ideally, this should not be necessary. I think the "layout" argument in the url should be removed from the links instead.

packages/edit-site/src/components/sidebar-dataviews/index.js
```diff
@@ -14,45 +12,13 @@ import { unlock } from '../../lock-unlock';
...
 import CustomDataViewsList from './custom-dataviews-list';
 
-const { useLocation, useHistory } = unlock( routerPrivateApis );
-
-/**
- * Hook to switch to table layout when switching to the trash view.
- * When going out of the trash view, it switches back to the previous layout if
- * there was an automatic switch to table layout.
- */
-function useSwitchToTableOnTrash() {
```
oandregal, line 24: This is removed, we don't want specific logic to any default/custom view. They all should work the same, see [slack thread](https://wordpress.slack.com/archives/C02QB2JS7/p1722331514967269) in core-editor channel.

oandregal:
Both default & custom views work now with latest code changes. This is how it works:

- Changing layout doesn't reset any other view config.
- Selecting a different default or custom view resets everything, including layout choices, as discussed in [this core-editor slack thread](https://wordpress.slack.com/archives/C02QB2JS7/p1722331514967269).
- Loading a custom view displays it with the proper layout. In trunk, it displays `list`, no matter the custom view layout.
- Loading a unknown activeView, loads an empty DataView (in trunk, the site editor doesn't load). Test loading http://localhost:8888/wp-admin/site-editor.php?postType=page&activeView=wrong-key

To cover all use cases, I essentially rewrote the `useView` hook, so looking at the diff may not be that useful — it's best to look at it whole. Also, apparently, this PR is less code than before 😅 +192 −204.

packages/edit-site/src/components/post-list/index.js
oandregal: mmm, I'm not sure I follow: don't we need to make sure the layout reflects the view type? Or are you saying that you don't want the layout URL when the view type is list?

Note we need to update the layout when a new default/custom is loaded. This was a source of complexity/bugs: in trunk, the trash view has its own custom hook for this and custom views that use a layout different than `list` don't load properly.

oandregal:
Marked some old conversations as resolved because the implementation changed significantly.

packages/edit-site/src/components/post-list/index.js
```diff
@@ -39,100 +40,123 @@ const { useLocation, useHistory } = unlock( routerPrivateApis );
...
+	useEffect( () => {
+		let newView;
+		if ( isCustom === 'true' ) {
+			newView = getCustomView( editedEntityRecord );
+		} else {
+			newView = getDefaultView( defaultViews, activeView );
+		}
+
+		if ( newView ) {
+			setViewWithUrlUpdate( newView );
```
jorgefilipecosta, line 160: I think when we switch to a custom view, as here we call setViewWithUrlUpdate which has this code `if ( isCustom === 'true' && editedEntityRecord?.id ) { editEntityRecord(`, we will endup calling get editEntityRecord just because of loading the view which may trigger a rerender. Not a big issue.

jorgefilipecosta (approved):
Works as advertised following what was discussed in the core-editor channel. It could be merged when the end to end tests are updated and passing 👍

packages/edit-site/src/components/post-list/index.js
oandregal: I pushed https://github.com/WordPress/gutenberg/pull/63889/commits/6958d2280f3b3da32c9e0a2a3d5bc09795067328 to make it work like trunk.

If there's no layout URL param (first time the Pages page loads) and the view.type is list, the URL param won't be updated. Note that, like trunk, once the layout param is present (user changes layout in dataviews, user changes to a default/custom view with a view.type different than list, etc.), it'll stay there even if the user changes back to list.

oandregal:
Pushed changes to address all feedback and fix the e2e tests. I've set this to auto-merge.

packages/edit-site/src/components/post-list/index.js
oandregal: That's right. I've looked at this but don't see a way to improve it at the moment. Custom views are an experimental feature and this isn't a big issue, so I'll go ahead with this PR anyway because it provides a lot of value.

I'd be happy to pair/look at this in a follow-up.

test/performance/specs/site-editor.spec.js
```diff
@@ -438,9 +438,7 @@ test.describe( 'Site Editor Performance', () => {
 				await Promise.all(
 					Array.from( { length: perPage }, async ( el, index ) => {
 						return await page
-							.getByRole( 'link', {
```
oandregal, line 441: This test assumed the page was using the `table` layout, which was the active one when the Trash view was loaded — and remained so. This PR changes that behavior and always loads the layout associated with the default view, hence at this point the layout will be list.

The test failed because the Pages are links in the table layout but they are not in the list layout.

oandregal:
There was an issue with this PR that I'm fixing at https://github.com/WordPress/gutenberg/pull/64306

Issue #63624, opened 2024-07-16
Admin redesign: Settings


jameskoster:
In the context of the wider admin redesign (https://github.com/WordPress/gutenberg/issues/53322), a scalable, portable, and responsive design pattern for settings pages is crucial. This issue outlines some foundational principles and design conventions for such a pattern.

<img width="2244" alt="Spec" src="https://github.com/user-attachments/assets/a2fd035d-62b0-4407-8722-b49c05bf5e1b">

## General Structure

The layout is defined by a consistent header and footer (dictated by the Layout (https://github.com/WordPress/gutenberg/issues/53617)), framing the main content area where settings are organized into sections. Each section comprises a left column for title and description, and a right column for the actual settings. The left column makes it easy to scan the page at a high level and locate a specific setting.

### Layout Conventions

<img width="1103" alt="Screenshot 2024-07-17 at 09 25 24" src="https://github.com/user-attachments/assets/8a60cc88-ec03-4c90-bf5c-61f23764be39">

- **12-Column Grid:** The design is based on a 12-column grid system, ensuring flexibility and consistency across various screen sizes.
- **Responsive Design**: The layout adapts to different screen sizes. When the container shrinks beyond a certain breakpoint, the columns stack vertically, and spacing values are adjusted for optimal readability and usability.
- **Horizontal Rules**: Sections are visually separated by horizontal rules, providing clear demarcation between different groups of settings.

### Section Structure

<img width="1048" alt="Screenshot 2024-07-17 at 09 28 22" src="https://github.com/user-attachments/assets/49276000-e0a6-4620-819d-d9a076d9a549">

Each section contains the following elements:

- **Section Title**: A clear and concise heading for the section.
- **Description**: A brief description of the section's purpose, including a link to relevant documentation for further guidance.
- **Settings Fields**: The main content of the section, containing various input elements like text inputs, select dropdowns, checkboxes, radio buttons, and textareas.

### Responsive Behavior

<img width="1076" alt="Screenshot 2024-07-17 at 09 29 44" src="https://github.com/user-attachments/assets/c9fa55dd-ca09-4fa4-bdfa-f3efdf6df4dd">

- **Stacking Columns**: On smaller screens/containers, the left and right columns stack vertically to make efficient use of limited space.
- **Adjusted Spacing**: Padding and margin values are reduced to maintain a clean and functional layout.

By adhering to these guidelines, we ensure a consistent, user-friendly, and responsive design for settings pages.

## Contextual examples
[...]

PR #63895, opened 2024-07-24
DataForm: migrate order action modal and introduce form validation

oandregal:

## What?

This PR migrates the "Change order" action modal to use DataForm. It also prepares the scaffold for form validation.

## How?

- Migrates the existing "Change order" action to use `DataForm` https://github.com/WordPress/gutenberg/pull/63895/commits/5dd8e8ea6ca52ab936a50fa0192411308a1bf9fc.
- Introduces a new utility `isItemValid( item, fields )` to centralize validation https://github.com/WordPress/gutenberg/pull/63895/commits/16f9f0817e3e833d6ce9337fae9de9de52017c6b.

## Testing Instructions

- Visit the site editor Pages page.
- In any of the records, go to the "Order" action and verify it works.
- This action should not let you update the order unless the value is a integer (no empty values allowed either).

### DIFF
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -44,7 +44,7 @@ export type Operator =
 
 export type ItemRecord = Record< string, unknown >;
 
-export type FieldType = 'text';
+export type FieldType = 'text' | 'integer';
 
 /**
  * A dataview field for a specific property of a data type.
--- a/packages/dataviews/src/types.ts
+++ b/packages/dataviews/src/types.ts
@@ -66,6 +66,11 @@ export type Field< Item > = {
 	label?: string;
 
 	/**
+	 * A description of the field.
+	 */
+	description?: string;
+
+	/**
 	 * Placeholder for the field.
 	 */
 	placeholder?: string;
--- a/packages/dataviews/src/validation.ts
+++ b/packages/dataviews/src/validation.ts
@@ -1,0 +1,30 @@
+/**
+ * Internal dependencies
+ */
+import { normalizeFields } from './normalize-fields';
+import type { Field } from './types';
+
+export function isItemValid< Item >(
+	item: Item,
+	fields: Field< Item >[]
+): boolean {
+	const _fields = normalizeFields( fields );
+	return _fields.every( ( field ) => {
+		const value = field.getValue( { item } );
+
+		// TODO: this implicitely means the value is required.
+		if ( field.type === 'integer' && value === '' ) {
+			return false;
+		}
+
+		if (
+			field.type === 'integer' &&
+			! Number.isInteger( Number( value ) )
+		) {
+			return false;
+		}
+
+		// Nothing to validate.
+		return true;
+	} );
+}
--- a/packages/dataviews/src/validation.ts
+++ b/packages/dataviews/src/validation.ts
@@ -2,13 +2,16 @@
  * Internal dependencies
  */
 import { normalizeFields } from './normalize-fields';
-import type { Field } from './types';
+import type { Field, Form } from './types';
 
 export function isItemValid< Item >(
 	item: Item,
-	fields: Field< Item >[]
+	fields: Field< Item >[],
+	form: Form
 ): boolean {
-	const _fields = normalizeFields( fields );
+	const _fields = normalizeFields(
+		fields.filter( ( { id } ) => !! form.visibleFields?.includes( id ) )
+	);
 	return _fields.every( ( field ) => {
 		const value = field.getValue( { item } );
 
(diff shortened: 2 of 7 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -39,21 +38,31 @@ import { getItemTitle } from '../../dataviews/actions/utils';
 const { PATTERN_TYPES, CreatePatternModalContents, useDuplicatePatternProps } =
 	unlock( patternsPrivateApis );
 
-// TODO: this should be shared with other components (page-pages).
+// TODO: this should be shared with other components (see post-fields in edit-site).
```
oandregal, line 41: https://github.com/WordPress/gutenberg/pull/63849 is an attempt to address this but it's still unclear how to do it.

packages/dataviews/src/validation.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Internal dependencies
+ */
+import { normalizeFields } from './normalize-fields';
+import type { Field, Form } from './types';
+
+export function isItemValid< Item >(
```
oandregal, line 7: This method just ports the existing logic into the new place, there's no behavioral changes.

In follow-up PRs we should discuss all the validation details iteratively: how a form declares the required fields (or is it the field itself?), etc. Those conversations shouldn't block this PR from landing.

packages/dataviews/src/validation.ts
```diff
@@ -0,0 +1,33 @@
...
+	form: Form
+): boolean {
+	const _fields = normalizeFields(
+		fields.filter( ( { id } ) => !! form.visibleFields?.includes( id ) )
+	);
+	return _fields.every( ( field ) => {
+		const value = field.getValue( { item } );
+
+		// TODO: this implicitely means the value is required.
+		if ( field.type === 'integer' && value === '' ) {
```
youknowriad, line 19: I think we need to formalize the field "types" better. I see us having a folder of field types and each field type would export a component to render the value, a component to render the control and a validation function. (Rather that having this logic within the global validation function)

youknowriad (commented):
I like two things here:

 - Validation is opt-in.
 - We're starting small.

Some things that could be a bit better or follow-ups:

 - Formalize field types definitions.
 - Potentially start having a storybook story per field type 
 - We need to thing about how to test these things properly.
 - I think we should focus more on the "quick edit" forms rather than these small existing forms. (Bring value to the user)
 - Error handling is not clear either at this point.

I believe we can ship this though as a first step but as follow-ups for now, I'd give more priority to quick edit and trying to replicate the "document inspector" in the quick edit panels.

oandregal:
Pushed storybook updates https://github.com/WordPress/gutenberg/pull/63895/commits/7e33991a547f05a5502903d8beaef6a929f1ddf5 and unit tests for `isItemValid` https://github.com/WordPress/gutenberg/pull/63895/commits/f9cdc89ec0b70c3d5e8222eed9fbae8234ce926f.

oandregal:
> I think we should focus more on the "quick edit" forms rather than these small existing forms. (Bring value to the user)

This is fair, I'll give quick edit a push. Though having more than one use case is proving useful and has already raised many unanswered API questions: how to share field definition between edit-site and edit-post? is "required field" a form API or a field API?, etc.

oandregal:
> Error handling is not clear either at this point.

I've asked for design input at https://github.com/WordPress/gutenberg/issues/63624#issuecomment-2252148146

oandregal:
@youknowriad do you want this PR to do something else or do we ship and iterate?

youknowriad:
I'm ok shipping this 👍 I'll give it a final review.

packages/dataviews/src/types.ts
```diff
@@ -44,7 +44,7 @@ export type Operator =
 
 export type ItemRecord = Record< string, unknown >;
 
-export type FieldType = 'text';
+export type FieldType = 'text' | 'integer';
```
youknowriad, line 47: I wonder if these should be 'text' | 'integer' or  'string' or 'number' to stay as close as possible to JSON schemas.

youknowriad (approved):

packages/dataviews/src/types.ts
oandregal: I went for integer because that's the type for `menu_order` according to the REST API https://developer.wordpress.org/rest-api/reference/pages/ (in terms of validation it's also more fine-grained and different than number).

oandregal:
https://github.com/WordPress/gutenberg/pull/63983 builds on this PR (type integer) to add support for "author" in the quick edit form.

Issue #50739, opened 2023-05-18
Zoom out


mtias:
This is a tracking issue for the zoomed out view that focuses on patterns rather than granular block editing. The idea behind a zoomed out view is to allow users to work with their content at a higher level, lighter touch than what individual blocks provide. This workflow is currently an experimental feature as the details of the interaction model are validated. Part of #39281.

### Visuals
![CleanShot 2024-03-11 at 07 09 58](https://github.com/WordPress/gutenberg/assets/1813435/66548f34-68c9-43c8-86c9-65afafd48e9e)

### Pattern Insertion

The smoothest way forward is to integrate zoom out is to key in using pattern insertion. From opening the patterns tab, zooming out, inserting patterns, and getting back into block editing. To do this we need to identify what the top-level blocks are, considering pages and templates (#59249). Persist the inserter when inserting patterns in zoom out mode (#59741) and explore ways to zoom back into a page/template.

- [x] https://github.com/WordPress/gutenberg/issues/44585
- [x] #59741
- [x] https://github.com/WordPress/gutenberg/issues/59963
- [x] https://github.com/WordPress/gutenberg/pull/59337
- [x] https://github.com/WordPress/gutenberg/pull/59249
- [x] #59962
- [x] https://github.com/WordPress/gutenberg/issues/61051
- [ ] Remove the "Pattern inserted" snackbar, now that patterns remain in view on insertion.
- [x] https://github.com/WordPress/gutenberg/issues/65749

### Interactions while zoomed out

When inserting patterns to a page, we should be able to select any of the top-level patterns, move them via arrows or drag and drop, shuffle them between related patterns, assign colorways, or delete them. This higher-up page editing experience provides a way to quickly compose pages of patterns, while block editing is much more granular. 

- [x] I should be able to select top-level patterns with the Inserter remaining open.
- [x] #45552
- [x] https://github.com/WordPress/gutenberg/pull/60054
- [x] https://github.com/WordPress/gutenberg/issues/59737
- [x] I should be able to drag patterns while in zoom out. 
- [x] Fix zoomed out drag chip position ([existing experiment](https://github.com/WordPress/gutenberg/pull/56889))
- [x] I should be able to delete an inserted pattern from the toolbar.
- [x] https://github.com/WordPress/gutenberg/issues/65754

### Quick inserter
- [ ] https://github.com/WordPress/gutenberg/issues/59739
[...]

Issue #63867, opened 2024-07-23
Zoom Out: Focus search when showing an insertion point


scruffian:
## What problem does this address?
In zoom out, when you click on the inserter button [+] the insertion marker shows, but it's not clear to users what is happening.

## What is your proposed solution?
When the inserter button is clicked we should focus the search input so users can see where their next action should be.

getdave:
I've looked into this a bit. The search component is here

https://github.com/WordPress/gutenberg/blob/34b5219842367bb111cc010cb1db9a3a593812dd/packages/block-editor/src/components/inserter/menu.js#L155-L167

It is possible to wrap this in a `<div>` and use the `useFocusOnMount` hook to programmatically focus the search. This can be done conditionally by passing the `ref` returned from `useFocusOnMount` or just passing `null`.

The problem is how to do this conditionally? It will require some new global state to track whether or not to focus the search.

We already have `setIsInserterOpened`, but that is `Boolean`. We could add options to that action to allow consumers to control a separate focus state.

scruffian:
I believe `setIsInserterOpened` uses `setInserterIsOpened` which takes whole load of options: 

https://github.com/WordPress/gutenberg/blob/34b5219842367bb111cc010cb1db9a3a593812dd/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js#L104

getdave:
Thanks. Funnily enough I just found that bit. I'm manipulating it now.

PR #56625, opened 2023-11-29
List View: Displace list view items when dragging (a bit more WYSIWYG)

andrewserong:
## What?

Partly addresses: https://github.com/WordPress/gutenberg/issues/56539, follows discussion on https://github.com/WordPress/gutenberg/pull/56579.

> [!NOTE]
> 🚧 This PR is still a bit of a hack — for the moment, let's focus on the usability issues, we can tidy the code up later.

This PR explores one potential approach for displacing list view items when dragging within the list view. Or, to put it another way, while dragging blocks in the list view, move the other blocks out of the way.

## Why?

The goal is to create a more WYSIWYG experience while dragging blocks in the list view. By displacing items, and using a drop indicator that effectively previews what the list item will look like when a user finished dropping their item, the list view should feel a bit more "solid" to use.

## How?

While dragging over the list view, move existing list view items up or down to make room for a drop indicator that looks like what the block looks like once it's dropped. I.e. a drop indicator that is a bit like a preview. Details "how" follows:

* Generate an object keyed by `clientId` containing the global index for all blocks within the list view.
* In the list view branch, for each block, look up the index of the current block and compare it to the index for the block being dragged over. Move blocks out of the way by applying either an `is-above` or `is-below` classname.
* Use `transform: translateY()` rules to offset the blocks so that they move out of the way. By using CSS for this, we get nice and smooth animations which disguises the fact that the dragging behaviour only _really_ updates every `200ms`. This hopefully gives the appearance of smoother drag and drop, without (again hopefully) degrading performance.
* Hide the normal list view drop indicator line, and use a drop indicator that previews the list view item in its final position (make it more WYSIWYG)
* Auto-expand list view items on hover
* Make the drag chip used in the list view be slightly transparent so that you can see where you dropping underneath it

## To-do

* [x] Leave a block sized space when dragging starts (don't move blocks around until the user drags over another block)
* [x] Smoothly handle dealing with the space when dragging outside of the list view area (i.e. don't leave the space where it is)
* [x] Correctly animate when releasing the mouse cursor so that the dragged item smoothly goes back into position (or at least remove the wrong animation that is there at the moment)
* [x] Smoothly deal with the space left when dragging multiple blocks (i.e. can we use a CSS variable instead of hard-coded height for the above / below positioning)
* [x] Come up with a better solution for dragging into collapsed blocks
* [x] Update styling of the drag chip
* [x] Look into performance issue when dragging over editor canvas
[...]

PR #63896, opened 2024-07-24
Zoom out: Try vertical displacement when dragging a pattern between existing patterns/sections

MaggieCabrera:
## What?

Part of https://github.com/WordPress/gutenberg/issues/50739

Implements a way to separate blocks in zoom out mode vertically both when we drag in a new pattern in between sections and when we click on the inserters

Co-authored-by: Dave Smith <444434+getdave@users.noreply.github.com>

## Why?

We need a better visual aid to guide the user when building with patterns. Separating section makes it clear where the action is happening

## How?

We are adding some conditional separators in between blocks of the blocklist that appear based on a couple of condition: if the block in question is a sectioning element, and if the insertion point is related to said block. The rendering of this separator is animated to push the sections apart.

This also required updating the drag handling logic to avoid triggering `dragend` events when moving in/out of the zoom out separators. This ensures the separator itself does not re-render if the mouse cursor inadvertently moves into/out of it whilst it is animating. This avoids the "stuttering" animations we were seeing early in the lifecycle of this PR.

## Testing Instructions

- Open the inserter and select a pattern category
- On the canvas, click on the plus inserters, they should separate the sections correctly
- Select a pattern and it should insert in the area that was opened
- Try and drag and drop a new pattern in between sections
- The sections should animate and separate while dragging in the area

## Screenshots or screencast 

https://github.com/user-attachments/assets/73d047cb-44d0-4f4e-9134-a9b848566e81

### DIFF
--- a/packages/block-editor/src/components/block-list/content.scss
+++ b/packages/block-editor/src/components/block-list/content.scss
@@ -453,3 +453,15 @@ _::-webkit-full-page-media, _:future, :root .has-multi-selection .block-editor-b
 		margin-bottom: auto;
 	}
 }
+
+.block-editor-block-list__zoom-out-separator {
+	/* same color as the iframe's background */
+	background: $gray-300;
+}
+
+// In Post Editor allow the separator to occupy the full width by ignoring (cancelling out) the global padding.
+.has-global-padding > .block-editor-block-list__zoom-out-separator,
+.block-editor-block-list__layout.is-root-container.has-global-padding > .block-editor-block-list__zoom-out-separator {
+	max-width: none;
+	margin: 0 calc(-1 * var(--wp--style--root--padding-right)) 0 calc(-1 * var(--wp--style--root--padding-left)) !important;
+}
--- a/packages/block-editor/src/components/block-list/content.scss
+++ b/packages/block-editor/src/components/block-list/content.scss
@@ -457,11 +457,15 @@ _::-webkit-full-page-media, _:future, :root .has-multi-selection .block-editor-b
 .block-editor-block-list__zoom-out-separator {
 	/* same color as the iframe's background */
 	background: $gray-300;
+	// Additional -1px is required to avoid sub pixel rounding errors allowing background to show.
+	margin-left: -1px;
+	margin-right: -1px;
 }
 
 // In Post Editor allow the separator to occupy the full width by ignoring (cancelling out) the global padding.
 .has-global-padding > .block-editor-block-list__zoom-out-separator,
 .block-editor-block-list__layout.is-root-container.has-global-padding > .block-editor-block-list__zoom-out-separator {
 	max-width: none;
-	margin: 0 calc(-1 * var(--wp--style--root--padding-right)) 0 calc(-1 * var(--wp--style--root--padding-left)) !important;
+	// Additional -1px is required to avoid sub pixel rounding errors allowing background to show.
+	margin: 0 calc(-1 * var(--wp--style--root--padding-right) - 1px) 0 calc(-1 * var(--wp--style--root--padding-left) - 1px) !important;
 }
--- a/packages/block-editor/src/components/block-list/content.scss
+++ b/packages/block-editor/src/components/block-list/content.scss
@@ -460,6 +460,11 @@ _::-webkit-full-page-media, _:future, :root .has-multi-selection .block-editor-b
 	// Additional -1px is required to avoid sub pixel rounding errors allowing background to show.
 	margin-left: -1px;
 	margin-right: -1px;
+	transition: background-color 0.3s ease;
+
+	&.is-dragged-over {
+		background: $gray-400;
+	}
 }
 
 // In Post Editor allow the separator to occupy the full width by ignoring (cancelling out) the global padding.
--- a/packages/block-editor/src/components/block-list/index.js
+++ b/packages/block-editor/src/components/block-list/index.js
@@ -39,6 +39,7 @@ import {
 	DEFAULT_BLOCK_EDIT_CONTEXT,
 } from '../block-edit/context';
 import { useTypingObserver } from '../observe-typing';
+import { ZoomOutSeparator } from './zoom-out-separator';
 import { unlock } from '../../lock-unlock';
 
 export const IntersectionObserver = createContext();
@@ -174,49 +175,55 @@ function Items( {
 	// function on every render.
 	const hasAppender = CustomAppender !== false;
 	const hasCustomAppender = !! CustomAppender;
-	const { order, selectedBlocks, visibleBlocks, shouldRenderAppender } =
-		useSelect(
-			( select ) => {
-				const {
-					getSettings,
-					getBlockOrder,
-					getSelectedBlockClientId,
-					getSelectedBlockClientIds,
-					__unstableGetVisibleBlocks,
-					getTemplateLock,
-					getBlockEditingMode,
-					__unstableGetEditorMode,
-				} = select( blockEditorStore );
+	const {
+		order,
+		isZoomOut,
+		selectedBlocks,
+		visibleBlocks,
+		shouldRenderAppender,
+	} = useSelect(
+		( select ) => {
+			const {
+				getSettings,
+				getBlockOrder,
+				getSelectedBlockClientId,
+				getSelectedBlockClientIds,
+				__unstableGetVisibleBlocks,
+				getTemplateLock,
+				getBlockEditingMode,
+				__unstableGetEditorMode,
+			} = select( blockEditorStore );
 
-				const _order = getBlockOrder( rootClientId );
+			const _order = getBlockOrder( rootClientId );
 
-				if ( getSettings().__unstableIsPreviewMode ) {
-					return {
-						order: _order,
-						selectedBlocks: EMPTY_ARRAY,
-						visibleBlocks: EMPTY_SET,
-					};
-				}
-
-				const selectedBlockClientId = getSelectedBlockClientId();
+			if ( getSettings().__unstableIsPreviewMode ) {
 				return {
 					order: _order,
-					selectedBlocks: getSelectedBlockClientIds(),
-					visibleBlocks: __unstableGetVisibleBlocks(),
-					shouldRenderAppender:
-						hasAppender &&
-						__unstableGetEditorMode() !== 'zoom-out' &&
-						( hasCustomAppender
-							? ! getTemplateLock( rootClientId ) &&
-							  getBlockEditingMode( rootClientId ) !== 'disabled'
-							: rootClientId === selectedBlockClientId ||
-							  ( ! rootClientId &&
-									! selectedBlockClientId &&
-									! _order.length ) ),
+					selectedBlocks: EMPTY_ARRAY,
+					visibleBlocks: EMPTY_SET,
 				};
-			},
-			[ rootClientId, hasAppender, hasCustomAppender ]
-		);
+			}
+
+			const selectedBlockClientId = getSelectedBlockClientId();
+			return {
+				order: _order,
+				selectedBlocks: getSelectedBlockClientIds(),
+				visibleBlocks: __unstableGetVisibleBlocks(),
+				isZoomOut: __unstableGetEditorMode() === 'zoom-out',
+				shouldRenderAppender:
+					hasAppender &&
+					__unstableGetEditorMode() !== 'zoom-out' &&
+					( hasCustomAppender
+						? ! getTemplateLock( rootClientId ) &&
+						  getBlockEditingMode( rootClientId ) !== 'disabled'
+						: rootClientId === selectedBlockClientId ||
+						  ( ! rootClientId &&
+								! selectedBlockClientId &&
+								! _order.length ) ),
+			};
+		},
+		[ rootClientId, hasAppender, hasCustomAppender ]
+	);
 
 	return (
 		<LayoutProvider value={ layout }>
@@ -230,10 +237,24 @@ function Items( {
 						! selectedBlocks.includes( clientId )
 					}
 				>
+					{ isZoomOut && (
+						<ZoomOutSeparator
+							clientId={ clientId }
+							rootClientId={ rootClientId }
+							position="top"
+						/>
+					) }
 					<BlockListBlock
 						rootClientId={ rootClientId }
 						clientId={ clientId }
 					/>
+					{ isZoomOut && (
+						<ZoomOutSeparator
+							clientId={ clientId }
+							rootClientId={ rootClientId }
+							position="bottom"
+						/>
+					) }
 				</AsyncModeProvider>
 			) ) }
 			{ order.length < 1 && placeholder }
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -87,6 +87,7 @@ export function ZoomOutSeparator( {
 						ease: [ 0.6, 0, 0.4, 1 ],
 					} }
 					className="block-editor-block-list__zoom-out-separator"
+					data-is-zoom-out-separator="true"
 				></motion.div>
 			) }
 		</AnimatePresence>
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -44,6 +44,7 @@ export function ZoomOutSeparator( {
 	}, [] );
 
 	const isReducedMotion = useReducedMotion();
+
 	if ( ! clientId ) {
 		return;
 	}
@@ -61,6 +62,11 @@ export function ZoomOutSeparator( {
 	}
 
 	if ( ! isSectionBlock ) {
+		return null;
+	}
+
+	// Only allow insertion and not grouping in Zoom Out mode.
+	if ( blockInsertionPoint?.operation !== 'insert' ) {
 		return null;
 	}
 
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -1,0 +1,98 @@
+/**
+ * WordPress dependencies
+ */
+import {
+	__unstableMotion as motion,
+	__unstableAnimatePresence as AnimatePresence,
+} from '@wordpress/components';
+import { useReducedMotion } from '@wordpress/compose';
+import { useSelect } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import { store as blockEditorStore } from '../../store';
+import { unlock } from '../../lock-unlock';
+
+export function ZoomOutSeparator( {
+	clientId,
+	rootClientId,
+	position = 'top',
+} ) {
+	const {
+		sectionRootClientId,
+		sectionClientIds,
+		blockInsertionPoint,
+		blockInsertionPointVisible,
+	} = useSelect( ( select ) => {
+		const {
+			getSettings,
+			getBlockInsertionPoint,
+			getBlockOrder,
+			isBlockInsertionPointVisible,
+		} = unlock( select( blockEditorStore ) );
+
+		const { sectionRootClientId: root } = unlock( getSettings() );
+		const sectionRootClientIds = getBlockOrder( root );
+		return {
+			sectionRootClientId: root,
+			sectionClientIds: sectionRootClientIds,
+			blockOrder: getBlockOrder( root ),
+			blockInsertionPoint: getBlockInsertionPoint(),
+			blockInsertionPointVisible: isBlockInsertionPointVisible(),
+		};
+	}, [] );
+
+	const isReducedMotion = useReducedMotion();
+	if ( ! clientId ) {
+		return;
+	}
+
+	let isSectionBlock = false;
+	let isVisible = false;
+
+	if (
+		( sectionRootClientId &&
+			sectionClientIds &&
+			sectionClientIds.includes( clientId ) ) ||
+		( clientId && ! rootClientId )
+	) {
+		isSectionBlock = true;
+	}
+
+	if ( ! isSectionBlock ) {
+		return null;
+	}
+
+	if ( position === 'top' ) {
+		isVisible =
+			blockInsertionPointVisible &&
+			blockInsertionPoint.index === 0 &&
+			clientId === sectionClientIds[ blockInsertionPoint.index ];
+	}
+
+	if ( position === 'bottom' ) {
+		isVisible =
+			blockInsertionPointVisible &&
+			clientId === sectionClientIds[ blockInsertionPoint.index - 1 ];
+	}
+
+	return (
+		<AnimatePresence>
+			{ isVisible && (
+				<motion.div
+					layout={ ! isReducedMotion }
+					initial={ { height: 0 } }
+					animate={ { height: '120px' } }
+					exit={ { height: 0 } }
+					transition={ {
+						type: 'tween',
+						duration: 0.2,
+						ease: [ 0.6, 0, 0.4, 1 ],
+					} }
+					className="block-editor-block-list__zoom-out-separator"
+				></motion.div>
+			) }
+		</AnimatePresence>
+	);
+}
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -16,7 +16,7 @@ import { unlock } from '../../lock-unlock';
 
 export function ZoomOutSeparator( {
 	clientId,
-	rootClientId,
+	rootClientId = '',
 	position = 'top',
 } ) {
 	const {
@@ -49,24 +49,14 @@ export function ZoomOutSeparator( {
 		return;
 	}
 
-	let isSectionBlock = false;
 	let isVisible = false;
 
-	if (
-		( sectionRootClientId &&
-			sectionClientIds &&
-			sectionClientIds.includes( clientId ) ) ||
-		( clientId && ! rootClientId )
-	) {
-		isSectionBlock = true;
-	}
+	const isSectionBlock =
+		rootClientId === sectionRootClientId &&
+		sectionClientIds &&
+		sectionClientIds.includes( clientId );
 
 	if ( ! isSectionBlock ) {
-		return null;
-	}
-
-	// Only allow insertion and not grouping in Zoom Out mode.
-	if ( blockInsertionPoint?.operation !== 'insert' ) {
 		return null;
 	}
 
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -87,7 +87,7 @@ export function ZoomOutSeparator( {
 						ease: [ 0.6, 0, 0.4, 1 ],
 					} }
 					className="block-editor-block-list__zoom-out-separator"
-					data-is-zoom-out-separator="true"
+					data-is-insertion-point="true"
 				></motion.div>
 			) }
 		</AnimatePresence>
--- a/packages/block-editor/src/components/block-list/zoom-out-separator.js
+++ b/packages/block-editor/src/components/block-list/zoom-out-separator.js
@@ -1,4 +1,9 @@
 /**
+ * External dependencies
+ */
+import clsx from 'clsx';
+
+/**
  * WordPress dependencies
  */
 import {
@@ -7,6 +12,7 @@ import {
 } from '@wordpress/components';
 import { useReducedMotion } from '@wordpress/compose';
 import { useSelect } from '@wordpress/data';
+import { useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -19,6 +25,7 @@ export function ZoomOutSeparator( {
 	rootClientId = '',
 	position = 'top',
 } ) {
+	const [ isDraggedOver, setIsDraggedOver ] = useState( false );
 	const {
 		sectionRootClientId,
 		sectionClientIds,
@@ -77,6 +84,7 @@ export function ZoomOutSeparator( {
 		<AnimatePresence>
 			{ isVisible && (
 				<motion.div
+					as="button"
 					layout={ ! isReducedMotion }
 					initial={ { height: 0 } }
 					animate={ { height: '120px' } }
@@ -86,10 +94,15 @@ export function ZoomOutSeparator( {
 						duration: 0.2,
 						ease: [ 0.6, 0, 0.4, 1 ],
 					} }
-					className="block-editor-block-list__zoom-out-separator"
-					// Indicate that this is an insertion point and should not trigger
-					// dropleave events for any parent drop zone.
+					className={ clsx(
+						'block-editor-block-list__zoom-out-separator',
+						{
+							'is-dragged-over': isDraggedOver,
+						}
+					) }
 					data-is-insertion-point="true"
+					onDragOver={ () => setIsDraggedOver( true ) }
+					onDragLeave={ () => setIsDraggedOver( false ) }
 				></motion.div>
 			) }
 		</AnimatePresence>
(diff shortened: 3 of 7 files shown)


packages/block-editor/src/components/block-list/index.js
```diff
@@ -174,49 +174,73 @@ function Items( {
...
-									! _order.length ) ),
+					selectedBlocks: EMPTY_ARRAY,
+					visibleBlocks: EMPTY_SET,
 				};
-			},
-			[ rootClientId, hasAppender, hasCustomAppender ]
-		);
+			}
+
+			const { sectionRootClientId: root } = unlock( getSettings() );
```
MaggieCabrera: This keeps throwing a `Error: Cannot unlock an object that was not locked before. ` that doesn't make any sense. The code still works so we are ignoring for the time being

packages/block-editor/src/components/block-list/content.scss
```diff
@@ -462,3 +462,12 @@ _::-webkit-full-page-media, _:future, :root .has-multi-selection .block-editor-b
 		margin-bottom: auto;
 	}
 }
+
+.zoom-out-separator {
```
MaggieCabrera: This most likely isn't where this CSS should live. Suggestions welcome! I tried inside block-tools but it seems that CSS is not loaded inside the iframe

packages/block-editor/src/components/block-list/index.js
```diff
@@ -174,49 +174,82 @@ function Items( {
...
+			const selectedBlockClientId = getSelectedBlockClientId();
+			const sectionRootClientIds = getBlockOrder( root );
+			// To do: move ZoomOutModeInserters to core/editor.
```
MaggieCabrera: Same problem we currently have with [the actual inserters](https://github.com/WordPress/gutenberg/blob/6128ceae4fbfcdbb0afbf976c34418a7d5e432ad/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js#L34) so we need the same solution for both

packages/block-editor/src/components/block-list/index.js
```diff
@@ -230,10 +263,34 @@ function Items( {
 						! selectedBlocks.includes( clientId )
 					}
 				>
+					{ isSectionBlock( clientId, sectionClientIds ) && (
+						<div
+							className={ clsx( 'zoom-out-separator', {
+								'is-open':
```
MaggieCabrera: We could use this same condition to load the div so we don't have a bunch of empty divs in the code, but then we need to figure out a different way to animate them

packages/block-editor/src/components/block-list/index.js
MaggieCabrera: @ellatrix I would love your opinion on this one, cause we can't seem to figure this one out

MaggieCabrera: @Mamaduka maybe you know?

Mamaduka: Sorry, I've not seen that error before. But I will try to test it later today or tomorrow.

MaggieCabrera: I think we fixed it by wrapping it under the zoom out mode condition, but it wasn't really straightforward to figure that out!

richtabor:
I like the exploration; there may be something here. A few notes to see if this can work better: 

1. Let's not have this expand anytime I select a pattern. Only when either the + has been selected, or I'm dragging: 

https://github.com/user-attachments/assets/65756c1a-2458-41af-a6fa-d8906d60ce35

2. There's a discrepancy when I drop in a pattern. I'd expect there to only be one "zone" to drop in a pattern:

https://github.com/user-attachments/assets/a9f5ded1-b983-41a3-b3fc-a9225f256943

3. The existing drop indicator is competing with this zone: 

https://github.com/user-attachments/assets/d5c88311-3cfe-4278-acf6-a1638d215f57

packages/block-editor/src/components/block-list/content.scss
scruffian: I think this is fine, my only concern is the class name itself, shouldn't it be prefixed with the component?

scruffian (approved):
This is looking good to me.

richtabor:
We need some sort of delay, or min time: 

https://github.com/user-attachments/assets/e583a46d-bf52-4b51-b995-a1bd7ef28a53

packages/block-editor/src/components/block-list/index.js
```diff
@@ -218,6 +223,90 @@ function Items( {
...
+				blockOrder: getBlockOrder( root ),
+				blockInsertionPoint: getBlockInsertionPoint(),
+				blockInsertionPointVisible: isBlockInsertionPointVisible(),
+			};
+		}
+		return {};
+	}, [] );
+
+	const isReducedMotion = useReducedMotion();
+	const renderZoomOutSeparator = ( clientId, position = 'top' ) => {
```
Mamaduka: Can we extract this as a separate component? It can be in the same file. Similar render functions are usually hard to maintain.

scruffian: Good idea. I moved it to a different file as well, as this one is really big.

MaggieCabrera:
> We need some sort of delay, or min time:
> CleanShot.2024-07-26.at.12.28.50.mp4

This is looking like we are allowing the insertion point to be before the header, I don't think that should be the case (at least it isn't on trunk)

packages/block-editor/src/components/block-list/zoom-out-separator.js
```diff
@@ -0,0 +1,103 @@
...
+	} = useSelect( ( select ) => {
+		const {
+			getSettings,
+			__unstableGetEditorMode,
+			getBlockInsertionPoint,
+			getBlockOrder,
+			isBlockInsertionPointVisible,
+		} = unlock( select( blockEditorStore ) );
+
+		if ( __unstableGetEditorMode() === 'zoom-out' ) {
```
Mamaduka: If we pass `editorMode` from the parent component, we can avoid creating store subscriptions for non-zoomed-out modes.

## How?

If you return early in the `mapSelect` function before calling `select( blockEditorStore )`, no store subscription will be created.

MaggieCabrera: Something like [this](https://github.com/WordPress/gutenberg/pull/63896/commits/675d7da09d3ba010d7d98842bda64a919172e2a0)? (and the commit after that...)

Mamaduka: Thanks, @MaggieCabrera!

MaggieCabrera:
I don't think this needs a transition delay, though I added one. I see a problem when we drag inside the inserter on top of the first element inside the root element (usually the first item right after the header) where the insertion point keeps being recalculated incorrectly. I haven't figured out yet why. I think once we fix that, this should be good enough to merge

getdave (commented):
I love the direction here. I think this kind of interaction will make the experience of drag and drop a lot better.

I appreciate a lot of effort has gone in here and I'd like to acknowledge that before I say that I'm not convinced we have reached the quality bar for this to be shipped. I'll caveat this by noting that it might be we're intending this to intersect with other PRs that resolve some of the issues I've noted here, so if that's the case then please excuse my ignorance as I come up to speed with the project state.

If you take a look the video recording below (including voiceover) you'll see that some drag and drop interactions can lead to unwanted side affects like repeating animations. IMHO we shouldn't be shipping until these issues are ironed out.

It would be beneficial to make a concrete list of the bugs that need to be fixed here and also any interactions that are _missing_ that need to be resolved. I will try and do that (see `Bugs` below).

Thanks for all the effort that's clearly done into this. I know this must be a _really_ difficult problem to solve.

<details>
<summary>Overview video</summary>

https://github.com/user-attachments/assets/b1b07595-035e-467c-82be-167f43ab6699

</details>


<details>
<summary>🐞 Bugs </summary>

### Cycling dropzone animations

We need to fix so that it's not possible to drag and have the dropzones animate underneath the dragged item. 


https://github.com/user-attachments/assets/b1738920-7eec-4bb8-8101-82d047960586

### Multiple dropzone targets

Let's not allow patterns to be dropped within existing patterns. That should happen as a lower level interaction in full editor mode (not zoom out). In fact I wonder if many of these issues will be resolved by fixing this one thing...

### Animation and dropzone below last pattern

I'm not sure we need the full animation with the gray box below the final pattern on the page. Perhaps this is a product of the auto-inserted `core/paragraph` block, but it doesn't feel natural in its current form.


https://github.com/user-attachments/assets/bae9420e-6844-4384-bea3-364922e5fbc4



</details>

getdave (dismissed):
Leaving this blocking review to avoid this shipping until [the issues I noted](https://github.com/WordPress/gutenberg/pull/63896#pullrequestreview-2224250751) are ironed out.

Hopefully merging https://github.com/WordPress/gutenberg/pull/64331 and then rebasing _this_ PR will be the first step.

getdave (commented):
https://github.com/user-attachments/assets/751b5d85-538f-4cf3-9cc7-ee7fce7d9e50

I'm also not sure about this interaction when using a keyboard (or assistive tech). We're leaning _very_ heavily on drag and drop and it isn't a great experience.

MaggieCabrera:
Something I noted but I can't investigate further at this moment:

While on the post editor, the root section has a class `has-global-padding` that is causing the separator to not extend the full width. I think we don't need this class present when we are on zoom-out mode.

MaggieCabrera:
Another bug to fix here: when we click on a plus button and then go to select a pattern to insert in the space, the separators collapse. This is because the insertion point changes as soon as the cursor leaves the iframed canvas

MaggieCabrera:
> Another bug to fix here: when we click on a plus button and then go to select a pattern to insert in the space, the separators collapse. This is because the insertion point changes as soon as the cursor leaves the iframed canvas

might be fixed by https://github.com/WordPress/gutenberg/pull/64392

richtabor:
> I'm also not sure about this interaction when using a keyboard (or assistive tech). We're leaning very heavily on drag and drop and it isn't a great experience.

@jeryj where are we on exploring focusing the search field when you select the inverter?

getdave:
@richtabor I'm working on [the Issue](https://github.com/WordPress/gutenberg/issues/63867) in https://github.com/WordPress/gutenberg/pull/64396.

getdave:
## Bugs

- in Site Editor dragging over the header of the template will trigger insertion points outside of the `<main>` element. 
- appearance (show/hide animation) of the zoom out separators causes re-calculation of the blockInsertionPoint which can cause animation stutters or in the worst case trigger opening the separator on a different insertion point

getdave:
Sorry for all the pings. Fixing this now.

getdave:
The main issue with this PR is still that the drop target can change underneath the cursor as a result of the appearance of the larger separator causing the DOM to shift. This causes UI flicker and quirks.

richtabor:
How about this: 

1. Let's not show the blue horizontal line at all. The `block-editor-block-list__zoom-out-separator` space is the indicator, and as-is they're conflicting. 

2. When you "pick up" a pattern, the "space" is added above or below it, while it's faded out in the canvas as well. It would be better if the dragged pattern does not take up space when you're dragging it—it's supposed to be dropped into the space of wherever `block-editor-block-list__zoom-out-separator`, so let that represent it's current placement. 

https://github.com/user-attachments/assets/8770aa4b-0fa7-4dda-90c4-5c7033a8d225

MaggieCabrera:
>  When you "pick up" a pattern, the "space" is added above or below it, while it's faded out in the canvas as well. It would be better if the dragged pattern does not take up space when you're dragging it—it's supposed to be dropped into the space of wherever `block-editor-block-list__zoom-out-separator`, so let that represent it's current placement.
>

What I think needs to happen here is that the "grabbed" pattern needs to disappear from the canvas while it's being dragged so that the effect is much closer to what you'd naturally be expecting

andrewserong (commented):
> I love the direction here. I think this kind of interaction will make the experience of drag and drop a lot better.

Same here! I think it makes a huge difference creating room for the pattern that's going to be inserted. I'll just ping @ellatrix for visibility on this PR too, as I know Ella has been looking at ideas for WYSIWYG drag and drop and the displacement here might be interesting in that context.

> We need to fix so that it's not possible to drag and have the dropzones animate underneath the dragged item.

The flickery situation here is a challenge, and I remember we encountered a bit of that in the list view, too, but that situation was a little different. Here's what I'm seeing in testing:

https://github.com/user-attachments/assets/785871b9-6dc1-4a1e-809b-2da47b46b7d6

In terms of what's happening in the editor canvas I'm wondering if there's possibly a couple of things going on:

* Could it be possible that [the throttled function](https://github.com/WordPress/gutenberg/blob/f12b78b7c4e830032d7f9a4641df40236644683b/packages/block-editor/src/components/use-block-drop-zone/index.js#L334) in `useBlockDropZone` is either firing too frequently, or updating when it shouldn't? That said, if I try updating [the throttle to 2000ms](https://github.com/WordPress/gutenberg/blob/f12b78b7c4e830032d7f9a4641df40236644683b/packages/block-editor/src/components/use-block-drop-zone/index.js#L493), I still get the flickery behaviour, which points to:
* While the separator is animating its size, does this result in the editor canvas "thinking" that we've moved the mouse cursor and things need to update? Or, is `showInsertionPoint` being called too frequently?

A couple of possible avenues to investigate further:

* ~See what happens if we greatly increase the throttle~ (I tried that locally and it didn't appear to help, it just made the issue appear more slowly 😅)
* Look into reducing when `showInsertionPoint` is fired while dragging, or only fire it if the insertion point has really changed? Or, within `ZoomOutSeparator` do some additional checks to only update if the insertion point has really changed?

The latter idea could be tricky if the mere presence of the animation results in the insertion point changing 🤔

Another (potentially quite hacky) idea:

* Would it help at all to move the logic for showing/hiding the `ZoomOutSeparator` into a hook so that it's only used in one place, and then use that hook within `block-list/index.js` and then pass down a prop to each of the instances of `ZoomOutSeparator` for showing / hiding the top or bottom separator? In this way, there could be a single hook that is the source of truth for which one to show. The idea here is that then this hook could "own" whether or not to update which one we're displaying, which might allow us to inject some logic that prevents updates when we don't want them to occur (i.e. don't suddenly switch while animating?)

These are just some thoughts / potential rabbitholes to go down in case it helps. Unfortunately it sounds like this will be a tricky bit of behaviour to refine!

getdave:
I've squashed this branch to a single commit to facilitate an easier rebase (there were going to be 18 commits where I'd have to resolve near identical changes). I've now rebased. It seems good to me but pinging @MaggieCabrera for confidence.

getdave:
I noticed that the `<ZoomOutSeparator>` was rendering for `blockInsertionPoint`'s whose operation was other than `insert`.

For example the operation would be `group` and the Zoom Out separator would render even though we'd disabled that ability for drag and drop.

As a result we now have smooth and consistent animations when dragging _within_ a section. 

However, note that this now exposes the real problem which is that when the `<ZoomOutSeparator>` does render it causes the layout to change underneath the cursor thereby triggering the recalculation of the `blockInsertionPoint` thereby causing the `<ZoomOutSeparator>` visibility to change repeatedly.

https://github.com/user-attachments/assets/4df97a27-b96a-42f9-b1be-a0cafcf192d9

MaggieCabrera:
> I noticed that the `<ZoomOutSeparator>` was rendering for `blockInsertionPoint`'s whose operation was other than `insert`.
> 
> For example the operation would be `group` and the Zoom Out separator would render even though we'd disabled that ability for drag and drop.
> 
> As a result we now have smooth and consistent animations when dragging _within_ a section.
> 
> However, note that this now exposes the real problem which is that when the `<ZoomOutSeparator>` does render it causes the layout to change underneath the cursor thereby triggering the recalculation of the `blockInsertionPoint` thereby causing the `<ZoomOutSeparator>` visibility to change repeatedly.

We need to figure out a way to stop recalculating the insertion point when we are hovering the separator

getdave:
In the video below I have set the `<ZoomOutSeparator>` to animate over 3 seconds and also have a red background. If you watch carefully you will see that the animation is interuppted only when the blue insertion line appears. This I believe is a vital clue.


https://github.com/user-attachments/assets/aaebce0c-5002-4cd0-bb79-488f25ab4c96

getdave:
After much sleuthing I narrowed this down to the `onDragLeave` of `useDropZone`:

https://github.com/WordPress/gutenberg/blob/498ce1743750f4a661c96c5883acd6893b165fb0/packages/block-editor/src/components/use-block-drop-zone/index.js#L524-L527

Turns out that `hideInsertionPoint` was being called when moving in/out of a `<ZoomOutSeparator>` because these elements are considered to be outside of the dropzone. 

This was causing the `insertionPoint` to be set to `null` in state which was causing the `getBlockInsertionPoint` selector to change thus causing the re-rendering issues (i.e. both the `ZoomOutSeparator` and the standard "blue line" insertion point to re-render).

~I've used a hacky means to check for the `ZoomOutSeparator`. This proves the issue is what I'm saying above. However, I think we'll need a better way of detecting this.~ (Update: I've reimplemented now)

Check the video to see the "fixed" implementation with no "stuttering":

https://github.com/user-attachments/assets/f2a54856-7f29-4b37-908c-9b00d0e00415

MaggieCabrera:
@getdave this looks so much better now! I don't see anything major beyond the implementation, so it looks like you nailed the problem here. Besides that there are just two more things that need to happen before we merge given [Rich's feedback](https://github.com/WordPress/gutenberg/pull/63896#issuecomment-2289267429):

- We need to remove the blue bar of the insertion point, since the separator is our cue in this mode that we can insert on drag and drop
- When dragging an existing pattern (not a new one) we need to make it visually disappear so the user gets the feedback that the pattern they "grabbed" is moving somewhere else (this might be tricky and mess with insertion points, but maybe it can just be CSS magic?)

getdave:
> @getdave this looks so much better now! I don't see anything major beyond the implementation, so it looks like you nailed the problem here. 

That's great to hear. I'll try and refine the implementation some. It's not great as it is.

> Besides that there are just two more things that need to happen before we merge given [Rich's feedback](https://github.com/WordPress/gutenberg/pull/63896#issuecomment-2289267429):
> 
> * We need to remove the blue bar of the insertion point, since the separator is our cue in this mode that we can insert on drag and drop
> * When dragging an existing pattern (not a new one) we need to make it visually disappear so the user gets the feedback that the pattern they "grabbed" is moving somewhere else (this might be tricky and mess with insertion points, but maybe it can just be CSS magic?)

I would make ~both of those separate PRs~ the 2nd of those a separate PR. I feel _this_ PR will be complicated as it is. It will also make it easier to revert later if we choose to do so for any reason. What do you think?

**Update**: I removed the "blue bar" in https://github.com/WordPress/gutenberg/pull/63896/commits/8ee6d064f55e0d42a75f907af98864fcd5b1f0e4.

packages/compose/src/hooks/use-drop-zone/index.js
```diff
@@ -165,10 +181,22 @@ export default function useDropZone( {
...
 				// https://bugs.webkit.org/show_bug.cgi?id=66547
+
 				if ( isElementInZone( event.relatedTarget ) ) {
 					return;
 				}
 
+				// If we're moving in/out of a ZoomOutSeparator, don't trigger
+				// the onDragLeave event. This is to prevent the dropzone from
+				// being hidden when the user is dragging a block in/over/around
+				// a block and the separator.
```
getdave: As this code is within `@wordpress/compose` we cannot refer to "ZoomOut". Therefore need to make this logic abstract such that it can be applied generically to the ZoomOutSeparator.

packages/compose/src/hooks/use-drop-zone/index.js
```diff
@@ -107,6 +108,22 @@ export default function useDropZone( {
...
+			 */
+			function isInsertionPoint( targetToCheck ) {
+				const { defaultView } = ownerDocument;
+
+				return !! (
+					defaultView &&
+					targetToCheck instanceof defaultView.HTMLElement &&
+					targetToCheck.dataset.isInsertionPoint
+				);
+			}
```
getdave: Because this is within the `compose` package we cannot refer to "Zoom out".

Therefore I've aimed to name the concept at work here which is a insertion point between blocks. Hopefully this is suitably generic.

If you apply `data-is-insertion-point` to the element it will be ignored when considering the dragend.

packages/block-editor/src/components/use-block-drop-zone/index.js
```diff
@@ -422,6 +422,10 @@ export default function useBlockDropZone( {
 				const [ targetIndex, operation, nearestSide ] =
 					dropTargetPosition;
 
+				if ( isZoomOutMode() && operation !== 'insert' ) {
+					return;
+				}
```
getdave, line 444: There are too many of these conditions creeping in which disallow non-"insert" events in Zoom Out mode. Whilst this is desired behaviour I'd like to codify the concept and have it made reusable. One for a follow up.

packages/block-editor/src/components/block-list/index.js
```diff
@@ -174,49 +175,55 @@ function Items( {
...
-			[ rootClientId, hasAppender, hasCustomAppender ]
-		);
+			}
+
+			const selectedBlockClientId = getSelectedBlockClientId();
+			return {
+				order: _order,
+				selectedBlocks: getSelectedBlockClientIds(),
+				visibleBlocks: __unstableGetVisibleBlocks(),
+				isZoomOut: __unstableGetEditorMode() === 'zoom-out',
```
MaggieCabrera, line 212: should we use the new [selector](https://github.com/WordPress/gutenberg/pull/64503) here?

getdave: Yes we should! 👍

packages/compose/src/hooks/use-drop-zone/index.js
andrewserong: Apologies if it's already been explained, but why does this need to happen within the `useDropZone` hook in the `compose` package, rather than within the `onDragLeave` callback in the consumer of the hook? The concept of an insertion point feels bit more conceptually part of the block editor, so looks a bit out of place here to me.

I.e. would it work to perform the insertion point check here instead:

https://github.com/WordPress/gutenberg/blob/f201f8f51d2971969264573327afea0590f82aa0/packages/block-editor/src/components/use-block-drop-zone/index.js#L528-L531

packages/block-editor/src/components/use-block-drop-zone/index.js
andrewserong: For now would it work to move this check up to the existing `isZoomOutMode()` check earlier in this code block so that the zoomed out mode checks are consolidated?

Leaving to follow-ups is also a valid approach! My (very) weakly held opinion here is that sometimes it can be tricky to change some of these rules after the fact if we leave them for a while as the drag and drop logic can be a bit challenging to reason about — I know I struggle looking at some of my past code changes to this area of the codebase 😅

andrewserong (commented):
Hey, great progress @getdave and @MaggieCabrera, this is feeling really smooth in practical testing! At first I thought I ran into a bug, but it turns out Redux Devtools was causing blocks to throw an error 🤷, so I think it's unrelated to this PR.

Here's how it's looking for me:

https://github.com/user-attachments/assets/c001cdc8-e126-4d8c-839e-067a189314ba

Just left a couple of small code comments — nothing major or blocking, mostly just wondering if we can avoid the change in the `compose` package. But that's more a conceptual concern than an actual problem as no APIs are changing 👍

richtabor:
> When dragging an existing pattern (not a new one) we need to make it visually disappear so the user gets the feedback that the pattern they "grabbed" is moving somewhere else (this might be tricky and mess with insertion points, but maybe it can just be CSS magic?)

Yea, there's tension currently when you grab a pattern to drag it. As the insertion spacer renders above and below this selected pattern (both of which don't actually move the pattern). 

The later half of @andrewserong's video shows this.

packages/compose/src/hooks/use-drop-zone/index.js
getdave: Good point. I couldn't see this yesterday as I was very fixed on the solution, but revisiting today it seems obvious that it's better to handle in the consumer rather than low level package.

Interesting, this is the very place I was first able to identify and isolate the original bug so I should have known better 😅

packages/block-editor/src/components/use-block-drop-zone/index.js
getdave: I'd love to do that but the issue is that `operation` isn't available until this point. That in turn is dependent on other steps above. We could bring all the ZoomOut checks down to _this_ location, but then we'd be doing unnecessary work rather than existing early.

getdave: I feel like we need a code quality PR to track down how many instances of ignoring non "insert" operations in Zoom Out mode exist in the code. Once we know this we can decide whether or not to refactor. It might be premature at this stage if there are only a couple.

getdave: I gave this a go but moving into the consumer causes the fix to break when testing in browser. I can't narrow down why that is.

getdave: Ok I ended up moving the whole thing to the `dragEnd` handler. That has the desired effect.

packages/block-editor/src/components/use-block-drop-zone/index.js
```diff
@@ -511,6 +515,23 @@ export default function useBlockDropZone( {
...
 	);
 
+	/**
+	 * Checks if the given element is an insertion point.
+	 *
+	 * @param {EventTarget|null} targetToCheck - The element to check.
+	 * @param {Document}         ownerDocument - The owner document of the element.
+	 * @return {boolean} True if the element is a insertion point, false otherwise.
+	 */
+	function isInsertionPoint( targetToCheck, ownerDocument ) {
```
andrewserong: (Very) tiny optimisation idea: this appears to be a pure function. Could it be moved outside of the body of the hook, since it doesn't depend on anything within it?

packages/block-editor/src/components/block-list/content.scss
```diff
@@ -454,3 +454,15 @@ _::-webkit-full-page-media, _:future, :root .has-multi-selection .block-editor-b
 		margin-bottom: auto;
 	}
 }
+
+.block-editor-block-list__zoom-out-separator {
+	/* same color as the iframe's background */
+	background: $gray-300;
```
andrewserong, line 459: This probably isn't a huge issue, but I noticed a _very_ subtle sub pixel line of the site's background image poking through while testing:

image

Here it is zoomed in, in case it helps:

image

I imagine this might be because we're scaled down to 50% so the background color here doesn't perfectly extend over the background image?

andrewserong (commented):
Code-wise this is looking really close to me, no major issues there from my perspective.

The main issue I ran into is that when testing across browsers, the behaviour appears to be a bit choppy in Safari:

https://github.com/user-attachments/assets/5f499c59-8e3f-419f-a3ce-fda844c4019e

I'm not sure if it's the same issue, but while working on #56625 I ran into an issue with choppiness in Safari that was related to this bug: https://bugs.webkit.org/show_bug.cgi?id=66547. The workaround in https://github.com/WordPress/gutenberg/pull/56625/commits/d9eb4cabc5c7dbbb1f03df71b2055209ece27901 was to add `pointerEvents: none` styling to prevent events from firing as frequently. This might be a long shot, but I wonder if it'd work (or help) to add `pointer-events: none` to the zoomed out separator, since we don't want hovering over the separator to do anything?

From my perspective, since this is behind a feature flag and the scope of the change is fairly contained, once the Safari choppiness is addressed (or verified if it's just a problem I'm running into locally), I reckon if @MaggieCabrera and @getdave are happy with how the PR's looking, it's probably in a good place to merge and continue to iterate on it.

Some things that could be good to continue working on / looking into:

* What @richtabor mentioned above re: the insertion space: https://github.com/WordPress/gutenberg/pull/63896#issuecomment-2306165662
* If you go to drag a pattern that's already in the canvas on a site with a background image, the placeholder state of the pattern you're moving displays as the background image of the site, so it can be a little confusing as to what's happening visually:

https://github.com/user-attachments/assets/675d56c9-acf2-400d-897a-10bc03dcbf37

Nice work continuing to polish this feature, folks!

packages/block-editor/src/components/use-block-drop-zone/index.js
```diff
@@ -521,7 +542,18 @@ export default function useBlockDropZone( {
...
+		onDragLeave( event ) {
+			const { ownerDocument } = event.currentTarget;
+
+			// If the drag event is leaving the drop zone and entering an insertion point,
+			// do not hide the insertion point as it is conceptually within the dropzone.
+			if (
+				isInsertionPoint( event.relatedTarget, ownerDocument ) ||
+				isInsertionPoint( event.target, ownerDocument )
+			) {
+				return;
```
andrewserong, line 554: I'm not sure of the answer here, so this is just a question! Before the `return;` statement, should we call `throttled.cancel()` to cancel the next cycle of that call, or does it not matter? I'm guessing we're treating this return statement as though nothing happened so _not_ calling `throttled.cancel()` does seem logical to me 👍

packages/block-editor/src/components/use-block-drop-zone/index.js
getdave: Yes you're right. The new location of the function in the consumer will allow this to happen 👍

packages/block-editor/src/components/block-list/content.scss
getdave: Good spot. We'll have to look into that one.

packages/block-editor/src/components/use-block-drop-zone/index.js
getdave: ✅

packages/block-editor/src/components/use-block-drop-zone/index.js
getdave: > we're treating this return statement as though nothing happened

That's how I"m seeing it 👍

packages/block-editor/src/components/block-list/content.scss
getdave: ✅ I fixed this by adding an additional 1px

getdave:
Rebased to pull in fix for Safari bug to allow testing on older versions of Safari.

getdave:
In relation to the jankyness of the separator animation in Safari. This is because the `relatedTarget` is not set when the `dragleave` event fires when leaving a block and entering the separator (caused by the animation).

The info on the draglave targets is as follows:

- `target` - The [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) the pointing device exited from
- `relatedTarget` - The [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) the pointing device entered to

As the `relatedTarget` is `null` we cannot detect that the drag has started over a separator (like we can in Chrome) an thus we cannot cancel the `dragLeave` handler on `useBlockDropZone`. This means that the `hideInsertionPoint` action is dispatched which causes the flicker.

The bug in Safari is - https://bugs.webkit.org/show_bug.cgi?id=66547

We now need to find a workaround.

packages/block-editor/src/components/block-list/content.scss
andrewserong: Nice idea 👍

andrewserong:
Thanks for digging into the jankiness @getdave! Just re-testing `trunk` on Safari, I notice there's some jankiness there, too, in the left and right areas of the canvas area in zoom out mode, with areas where the drag chip gets stuck:

https://github.com/user-attachments/assets/caf9b9bd-60ca-4c5d-b7c8-8030504d9b89

So, I'm wondering if the source of the jankiness issue is partly something that already exists in `trunk`, but is more visible in this PR due to the animated separators, whereas the drop indicator lines are bit more resilient to hiding jankiness since there isn't any animation.

I'm a bit on the fence about whether or not the jankiness is a blocker to landing — it'd be a great thing to fix and polish, but I also don't want to hold up this PR. Do feel free to go ahead and land this if you and others are happy enough with the current state of the PR!

MaggieCabrera:
My vote is YES to land this and polish after. This is a much better experience than trunk already

scruffian (approved):
This is looking great, thanks for all the hard work

scruffian:
I added an issue to track the ongoing problems in Safari: https://github.com/WordPress/gutenberg/issues/65125

Issue #63510, opened 2024-07-13
Improve zoom out animation when selecting pattern categories


richtabor:
The current experience of zooming out when selecting a pattern category (with the zoom out experiment enabled) is a quite jarring. Let's clean up the shifting elements to make it feel much more fluid. 

#### 1.  Ideally, the Inserter does not "reanimate" when selecting the first category (which engages zoom out): 

https://github.com/user-attachments/assets/b3f38623-1b98-4776-928d-69873dc24a3c

#### 2. When disengaging zoom out, the Inserter should not shift to fill the space of the categories panel: 
 
![CleanShot 2024-07-12 at 20 20 41](https://github.com/user-attachments/assets/0ecb295e-9a9e-452e-9ce8-443da139c2c9)

PR #63898, opened 2024-07-24
Inserter: Resize canvas to fit when pattern category selected

scruffian:
## What?
One of the things we have learnt from zoom out mode is that its very annoying when the pattern category previews overlap the canvas. This updates the canvas to resize based on whether the pattern category previews panel is open.

## Why?
It's annoying to have this overlapping the canvas.

## How?
Create a new private value in the store to save whether or not the sidebar is wide. Ideally the sidebar should be able to be any size but that's out of the scope of this PR.

## Testing Instructions
1. Open the Site Editor
2. Open the inserter
3. Select the patterns tab
4. Select a pattern category
5. Notice that the canvas resizes when the category previews are open, so that they don't overlap the content.

## Screenshots or screencast 
Screenshot 2024-07-26 at 12 54 25

### DIFF
--- a/packages/block-editor/src/components/inserter/menu.js
+++ b/packages/block-editor/src/components/inserter/menu.js
@@ -17,7 +17,7 @@ import {
 import { VisuallyHidden, SearchControl, Popover } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { useDebouncedInput } from '@wordpress/compose';
-import { useSelect, useDispatch } from '@wordpress/data';
+import { useSelect } from '@wordpress/data';
 
 /**
  * Internal dependencies
@@ -62,17 +62,9 @@ function InserterMenu(
 		useDebouncedInput( __experimentalFilterValue );
 	const [ hoveredItem, setHoveredItem ] = useState( null );
 
-	const selectedPatternCategory = useSelect(
-		( select ) => {
-			const _selectedPatternCategory =
-				select( blockEditorStore ).selectedPatternCategory();
-			return _selectedPatternCategory
-				? _selectedPatternCategory
-				: __experimentalInitialCategory;
-		},
-		[ __experimentalInitialCategory ]
+	const [ selectedPatternCategory, setSelectedPatternCategory ] = useState(
+		__experimentalInitialCategory
 	);
-	const { setSelectedPatternCategory } = useDispatch( blockEditorStore );
 
 	const [ patternFilter, setPatternFilter ] = useState( 'all' );
 	const [ selectedMediaCategory, setSelectedMediaCategory ] =
--- a/packages/block-editor/src/components/inserter/menu.js
+++ b/packages/block-editor/src/components/inserter/menu.js
@@ -61,11 +61,9 @@ function InserterMenu(
 	const [ filterValue, setFilterValue, delayedFilterValue ] =
 		useDebouncedInput( __experimentalFilterValue );
 	const [ hoveredItem, setHoveredItem ] = useState( null );
-
 	const [ selectedPatternCategory, setSelectedPatternCategory ] = useState(
 		__experimentalInitialCategory
 	);
-
 	const [ patternFilter, setPatternFilter ] = useState( 'all' );
 	const [ selectedMediaCategory, setSelectedMediaCategory ] =
 		useState( null );
@@ -106,7 +104,7 @@ function InserterMenu(
 				}
 			} );
 		},
-		[ onInsertBlocks, onSelect, shouldFocusBlock, ref ]
+		[ onInsertBlocks, onSelect, shouldFocusBlock ]
 	);
 
 	const onInsertPattern = useCallback(
--- a/packages/block-editor/src/components/inserter/menu.js
+++ b/packages/block-editor/src/components/inserter/menu.js
@@ -17,7 +17,7 @@ import {
 import { VisuallyHidden, SearchControl, Popover } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { useDebouncedInput } from '@wordpress/compose';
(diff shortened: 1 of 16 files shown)


packages/interface/src/components/interface-skeleton/index.js
```diff
@@ -19,6 +19,8 @@ import {
 	useViewportMatch,
 	useResizeObserver,
 } from '@wordpress/compose';
+import { useSelect } from '@wordpress/data';
+import { store as blockEditorStore } from '@wordpress/block-editor';
```
scruffian: I don't think we're allowed to use the block editor package inside interface 🤔

packages/interface/src/components/interface-skeleton/index.js
```diff
@@ -182,14 +197,15 @@ function InterfaceSkeleton(
...
-									isMobileViewport ? 'mobileOpen' : 'open'
-								}
+								animate={ animation }
 								exit="closed"
 								variants={ {
 									open: { width: secondarySidebarSize.width },
 									closed: { width: 0 },
 									mobileOpen: { width: '100vw' },
+									patternsOpen: {
+										width: '650px',
```
scruffian: 300 + 350 - should be a variable

packages/block-editor/src/components/inserter/style.scss
```diff
@@ -87,6 +87,10 @@ $block-inserter-tabs-height: 44px;
 	overflow: visible;
 }
 
+.block-editor-inserter__menu.show-panel {
+	width: 650px;
```
scruffian: Should be a variable

packages/block-editor/src/components/inserter/menu.js
```diff
@@ -61,9 +61,19 @@ function InserterMenu(
...
-	const [ selectedPatternCategory, setSelectedPatternCategory ] = useState(
-		__experimentalInitialCategory
+
+	const selectedPatternCategory = useSelect(
+		( select ) => {
+			const _selectedPatternCategory =
+				select( blockEditorStore ).selectedPatternCategory();
+			return _selectedPatternCategory
+				? _selectedPatternCategory
+				: __experimentalInitialCategory;
```
scruffian: Does this work as a fallback?

packages/block-editor/src/components/inserter/style.scss
```diff
@@ -299,7 +303,7 @@ $block-inserter-tabs-height: 44px;
 	@include break-medium {
 		border-left: $border-width solid $gray-200;
 		padding: 0;
-		left: 100%;
+		right: 0;
```
scruffian: What impact does this have on the mobile view?

packages/interface/src/components/interface-skeleton/index.js
jeryj: Probably not -- maybe we shouldn't be storing the pattern category. As in, we don't really care that we have a pattern category selected, we want to tell the interface how to be laid out. So, something like 'wide-sidebar' or 'zoom-out-wide'?

packages/block-editor/src/components/inserter/menu.js
```diff
@@ -61,9 +61,19 @@ function InserterMenu(
...
+		},
+		[ __experimentalInitialCategory ]
 	);
+	const { setSelectedPatternCategory } = useDispatch( blockEditorStore );
```
jeryj: To explain my other comment more: instead of dispatching `setSelectedPatternCategory` to the block store, maybe dispatch something to the interface package about the layout we want. In the interface package we don't care what category is selected. We care about the layout we want when we have a category selected.

packages/interface/src/components/interface-skeleton/index.js
scruffian: good idea!

packages/block-editor/src/components/inserter/style.scss
```diff
@@ -82,11 +82,16 @@ $block-inserter-tabs-height: 44px;
 }
 
 .block-editor-inserter__menu {
+	display: flex;
```
scruffian, line 90: There is a risk that this breaks something

scruffian:
The one problem I have with this PR is this error:
`Store "core/interface" is already registered.`

richtabor:
Just confirming, but this is assuming the zoom out with patterns experiment is disabled, correct?

richtabor:
It's certainly better than covering up the content. Nice.

scruffian:
> Just confirming, but this is assuming the zoom out with patterns experiment is disabled, correct?

Yes

richtabor:
Pushed a few small design tweaks, removing the duplicate shadow (now that its not needed for this layout), resolving the flex width on mobile, and using $sidebar-width instead of a new variable (buys a little space with a -20px difference).

richtabor:
Curious, does this make https://github.com/WordPress/gutenberg/issues/63510 easier to resolve perhaps?

richtabor (commented):
Design wise, this is good to go. 


### Trunk

https://github.com/user-attachments/assets/861ec268-95a0-4dde-ab1c-238026d1c5b0


### This PR

https://github.com/user-attachments/assets/42a1f60f-02ec-4554-b999-bb39cc4d8130

packages/block-editor/src/components/tabbed-sidebar/style.scss
```diff
@@ -6,7 +6,7 @@
 	overflow: hidden;
 
 	@include break-medium() {
-		width: 350px;
```
jeryj, line 9: I think a simpler plan will be to:
- When pattern categories are open
- Update this width to be the desired total width
- Remove all the absolute hacks for the pattern categories

scruffian:
This was resolved by https://github.com/WordPress/gutenberg/pull/64110

PR #60787, opened 2024-04-16
Update help text alignment in CheckboxControl

jameskoster:
## What?
Update the margin applied to help text in checkbox control so that it aligns with the label.

## Why?
improves readability, and aligns with the treatment of status radios in the following UI:

Screenshot 2024-04-16 at 16 41 52

## Testing Instructions
1. Run `npm run storybook:dev`
2. Navigate to CheckboxControl and check the text alignment

1. Open the site editor
2. Navigate to the details page for the "Blog Home" template (you may need to create this template)
3. Observe the updated text alignment in the "Discussion" settings. 

### Before
Screenshot 2024-04-16 at 16 34 51
Screenshot 2024-04-16 at 16 39 11

### After
Screenshot 2024-04-16 at 16 34 57
Screenshot 2024-04-16 at 16 39 33

PR #63903, opened 2024-07-24
Fix toggle help indentation

jameskoster:
Follow-up to https://github.com/WordPress/gutenberg/pull/63490.

## What?
Fix indentation when help text spans several lines.

## Why?
Improves appearance.

## Testing Instructions
1. Observe a `ToggleControl` with longer help text, e.g. the 'Inner blocks use content width' option on Group blocks.
2. Notice the help text is correctly indented.

| Before | After |
| --- | --- |
| Screenshot 2024-07-24 at 13 42 14 | Screenshot 2024-07-24 at 13 38 52 |

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -11,6 +11,7 @@
 -   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
 -   `SelectControl`: Fix hover/focus color in wp-admin ([#63855](https://github.com/WordPress/gutenberg/pull/63855)).
+-   `ToggleControl`: Fix indentation ([#63903](https://github.com/WordPress/gutenberg/pull/63903)).
 
 ### Enhancements
 
--- a/packages/components/src/toggle-control/index.tsx
+++ b/packages/components/src/toggle-control/index.tsx
@@ -88,7 +88,13 @@ export function ToggleControl(
 	return (
 		<BaseControl
 			id={ id }
-			help={ helpLabel }
+			help={
+				helpLabel && (
+					<span className="components-toggle-control__help">
+						{ helpLabel }
+					</span>
+				)
+			}
 			className={ classes }
 			__nextHasNoMarginBottom
 		>
--- a/packages/components/src/toggle-control/index.tsx
+++ b/packages/components/src/toggle-control/index.tsx
@@ -88,13 +88,7 @@ export function ToggleControl(
 	return (
 		<BaseControl
 			id={ id }
-			help={
-				helpLabel && (
-					<span className="components-toggle-control__help">
-						{ helpLabel }
-					</span>
-				)
-			}
+			help={ helpLabel }
 			className={ classes }
 			__nextHasNoMarginBottom
 		>
--- a/packages/components/src/toggle-control/style.scss
+++ b/packages/components/src/toggle-control/style.scss
@@ -6,6 +6,7 @@
 	}
 }
 
-.components-base-control__help {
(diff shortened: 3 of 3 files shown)


paaljoachim:
Ahh yes. This will also be helpful to have adjusted. Thanks James!

ciampo (commented):
This is testing well for me — although I'd like to suggest an alternative approach: instead of changing the markup in `ToggleControl`, we could just add `display: inline-block` to the `span.components-toggle-control__help`, which would cause the margin to apply to the whole paragraph (instead of the first line).

It won't remove the additional `<span>` that was [pointed out in the previous PR](https://github.com/WordPress/gutenberg/pull/63490#issuecomment-2247617551), but it won't need to use `BaseControl`'s internal classnames.

Bonus: instead of `margin-left`, let's use the logical equivalent `margin-inline-start`.

```diff
diff --git a/packages/components/src/toggle-control/index.tsx b/packages/components/src/toggle-control/index.tsx
index c84ad90f12..89ef34f6d6 100644
--- a/packages/components/src/toggle-control/index.tsx
+++ b/packages/components/src/toggle-control/index.tsx
@@ -88,7 +88,13 @@ export function ToggleControl(
 	return (
 		<BaseControl
 			id={ id }
-			help={ helpLabel }
+			help={
+				helpLabel && (
+					<span className="components-toggle-control__help">
+						{ helpLabel }
+					</span>
+				)
+			}
 			className={ classes }
 			__nextHasNoMarginBottom
 		>
diff --git a/packages/components/src/toggle-control/style.scss b/packages/components/src/toggle-control/style.scss
index be7966611f..68733b53a9 100644
--- a/packages/components/src/toggle-control/style.scss
+++ b/packages/components/src/toggle-control/style.scss
@@ -6,6 +6,7 @@
 	}
 }
 
-.components-base-control__help {
-	margin-left: $toggle-width + $grid-unit-10;
+.components-toggle-control__help {
+	display: inline-block;
+	margin-inline-start: $toggle-width + $grid-unit-10;
 }

```

What do folks think?

jameskoster:
Works for me :)

ciampo:
Cool. Let's wait a bit in case anyone else wants to share an opinion, and then we can apply any changes and merge.

mirka:
Yes, @ciampo's approach is exactly [what we did](https://github.com/WordPress/gutenberg/pull/60787/files#diff-5af1abc3e6be32252ae5d5ff4a39efeb7652f02ea277b543e92d9f3dd22fdefa) for CheckboxControl. Sorry I missed this in review 🙇

richtabor:
LGTM

jameskoster:
Pushed the suggested changes.

tyxla (approved):
Code looks good, and this tests as expected 👍 

Screenshot 2024-07-25 at 14 03 14

afercia:
Ideally, markup should not be used for presentational purposes. As I suggested in https://github.com/WordPress/gutenberg/pull/63490#issuecomment-2247617551 it would be better to remove the added `<span>` element all together, as it is used only for presentational purposes.
That applies to the approach used for the checkbox as well: https://github.com/WordPress/gutenberg/pull/60787/files#diff-ad327c4ba4df23e9aa8712664f69691187c90f18c54521630d421040dd24ffa5R101-R103

mirka:
> Ideally, markup should not be used for presentational purposes. As I suggested in [#63490 (comment)](https://github.com/WordPress/gutenberg/pull/63490#issuecomment-2247617551) it would be better to remove the added `<span>` element all together, as it is used only for presentational purposes.

Yes, it's not exactly ideal, but it's to avoid having to target an internal classname of a separate component (`components-base-control__help`). Ultimately we should consider some better ways to help with element alignment like this. It could be CSS grid, it could be having a BaseControl API that is more modular (cc @WordPress/gutenberg-components).

jameskoster:
I reckon we should merge this to fix the visual bug, then explore solutions eliminate the need for additional markup separately.

PR #63908, opened 2024-07-24
Fix the shadows Range control accessibility and usability

### LINKED ISSUE #63899: The Shadow Range controls are unlabeled
### Description

Discovered while testing https://github.com/WordPress/gutenberg/pull/63874

All the range controls in the Shadows popover are unlabeled. Screenshot:

Screenshot 2024-07-24 at 13 59 45

- Go to the Site editor > Styles > Shadows
- Edit one of the shadows.
- Inspect the source and observe all the 4 sliders are unlabeled: X Position, Y Position, Blur, Spread.

_No response_

### Environment info

_No response_

- [X] Yes

- [X] Yes

afercia:
Fixes https://github.com/WordPress/gutenberg/issues/63899

## What?

The sliders (Range controls) in the popover to edit the global styles Shadows are unlabeled.
Also, the design is inconsistent: the toggle button to switch to the 'custom value' input is placed after the visual labels. This is inconsistent as in many other similar controls the toggle button is placed after the control.

It appears this has been done only for visual purposes, to lay out the 4 controls in a 2 x 2 grid. Which is arguable because the length of the sliders is very small and setting accurately a specific value may be a difficult task for some users. These sliders need more horizontal room.

Also to mention: the visual labels are actually H2 headings and have a slightly different styling (e.g. the color) compared to the labels used in similar controls.

## Why?

All controls must be labeled.
For better usability and accessibility, controls should be used consistently and avoid custom design.

## How?

Removes the headings that were used as visusal labels in favor of real label elements.
Removes the 2 x 2 grid layout in favor od displaying the controls full widh in 4 rows.

## Testing Instructions

- Go to the Site editor > Styles > Shadows
- Select a shadow and open the popover to edit the shadow.
- Observe the 4 sliders are properly labeled.
- Switch the sliders to the 'custom value' inputs and observe the inputs are properly labeled.
- Observe the controls don't use a grid layout any longer.

## Screenshots or screencast 

Before and after:

Screenshot 2024-07-24 at 17 08 06

### DIFF
--- a/packages/edit-site/src/components/global-styles/shadows-edit-panel.js
+++ b/packages/edit-site/src/components/global-styles/shadows-edit-panel.js
@@ -13,14 +13,13 @@ import {
 	__experimentalItemGroup as ItemGroup,
 	__experimentalInputControl as InputControl,
 	__experimentalUnitControl as UnitControl,
-	__experimentalParseQuantityAndUnitFromRawValue as parseQuantityAndUnitFromRawValue,
+	__experimentalGrid as Grid,
 	__experimentalDropdownContentWrapper as DropdownContentWrapper,
 	__experimentalUseNavigator as useNavigator,
 	__experimentalToggleGroupControl as ToggleGroupControl,
 	__experimentalToggleGroupControlOption as ToggleGroupControlOption,
 	__experimentalConfirmDialog as ConfirmDialog,
 	Dropdown,
-	RangeControl,
 	Button,
 	Flex,
 	FlexItem,
@@ -34,7 +33,6 @@ import {
 	plus,
 	shadow as shadowIcon,
 	reset,
-	settings,
 	moreVertical,
 } from '@wordpress/icons';
 import { useState, useMemo } from '@wordpress/element';
@@ -50,7 +48,6 @@ import {
 	getShadowParts,
 	shadowStringToObject,
 	shadowObjectToString,
-	CUSTOM_VALUE_SETTINGS,
 } from './shadow-utils';
 
 const { useGlobalSetting } = unlock( blockEditorPrivateApis );
@@ -463,8 +460,7 @@ function ShadowPopover( { shadowObj, onChange } ) {
 					label={ __( 'Inset' ) }
 				/>
 			</ToggleGroupControl>
-			<VStack gap={ 4 }>
-				>
+			<Grid columns={ 2 } gap={ 4 }>
 				<ShadowInputControl
 					label={ __( 'X Position' ) }
 					value={ shadowObj.x }
@@ -488,21 +484,12 @@ function ShadowPopover( { shadowObj, onChange } ) {
 					hasNegativeRange
 					onChange={ ( value ) => onShadowChange( 'spread', value ) }
 				/>
-			</VStack>
+			</Grid>
 		</VStack>
 	);
 }
 
-function ShadowInputControl( { label, value, onChange, hasNegativeRange } ) {
-	const [ isCustomInput, setIsCustomInput ] = useState( false );
-	const [ parsedQuantity, parsedUnit ] =
-		parseQuantityAndUnitFromRawValue( value );
-
-	const sliderOnChange = ( next ) => {
-		onChange(
-			next !== undefined ? [ next, parsedUnit || 'px' ].join( '' ) : '0px'
-		);
-	};
+function ShadowInputControl( { label, value, onChange } ) {
 	const onValueChange = ( next ) => {
 		const isNumeric = next !== undefined && ! isNaN( parseFloat( next ) );
 		const nextValue = isNumeric ? next : '0px';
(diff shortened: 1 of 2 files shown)


t-hamano (commented):
Thanks for the PR!

Code-wise, everything seems right. Let's get some feedback from the design team.

jameskoster:
I'm happy to try this.

t-hamano:
@afercia 

What about removing the title to keep the height as small as possible?

image

For example, the color picker doesn't have a title.

I think this may not be an accessibility issue as long as the screen reader announces that the popover has been opened (the button has been expanded) and announces the label of the first focusable element inside the popover.

richtabor:
If we're fixing a11y issues with labels and such, those PRs should only consist of those changes — they'd be much faster to wrap up/review if so. If we're making design changes as well, we need design reviews as well, which can often slow things down, as additional feedback/iterations are not uncommon. Just a note. 

> Also, the design is inconsistent: the toggle button to switch to the 'custom value' input is placed after the visual labels. This is inconsistent as in many other similar controls the toggle button is placed after the control.

The Font Size control also uses this; it's an established (and expected) UX pattern. 

CleanShot 2024-07-26 at 08 27 23

However, I'm in favor of further simplifying this control, to not use isCustomInput and just surface the UnitControls for each (in a grid). This would not be dissimilar to the UX of border and radius controls (where we do not have additional RangeControl before custom inputs. 

CleanShot 2024-07-26 at 08 30 09

Just making the list of controls into a longer visual list of items does not improve usability of this particular control.

jameskoster:
That also works for me. Users can drag on unit controls to achieve a similar experience to range sliders.

afercia:
Thanks all for your review and feedback. Some points:
 
> What about removing the title to keep the height as small as possible? ... For example, the color picker doesn't have a title.

To me, any panel, popover, modal dialog, etc. should have a visible title. I do know there is large inconsistency about this in the editor, [as I mentioned in the related issue](https://github.com/WordPress/gutenberg/issues/63899#issuecomment-2248249080). Still, clearly identifying what a panel, popover, modal dialog is about is not just a thing for screen readers, it's about general usability for all users. 

> If we're fixing a11y issues with labels and such, those PRs should only consist of those changes — they'd be much faster to wrap up/review if so. If we're making design changes as well, we need design reviews as well, which can often slow things down, as additional feedback/iterations are not uncommon. Just a note.
nge.

Thanks for reminding me this point. I'm afraid I forgot about it, after more than 10 years of contribution to WordPress. I marked both this PR and the related issue with the 'Needs Design Feedback' label because as an experienced contributor to WordPress I think this issue needs a design change to be solved. So I'm not sure I understand how your consideration above can be any useful. I'm intentionally proposing a design change. Just a note.

Besides that, I think you're missing that accessibility is not just about 'labels and such'. When the design itself is less than ideal and impacts accessibility and / or usability, the design needs to be changed.

Range controls specifically are _designed_ to allow users to _fine tune_ a value in a defined range. Placing a range control in a spot with very limited horizontal space defeats the benefit of a range control in the first place. The current grid layout makes each range 'rail' have a width of 116 pixels, which is insufficient to be able to easily fine-tune a value. As such, the grid layout is, to me, just a wrong design and impacts usability to start with.

> Just making the list of controls into a longer visual list of items does not improve usability of this particular control.

I'm not sure I understand why a longer list would be a problem. The point here is allowing the range controls to have more horizontal space. No gric. More height, yes. Users can always scroll.

Anyways, further simplifying by removing the range control and only using the UnitControls makes sens to me. In this case, the grid layout could stay.

afercia:
In the last commit I entirely removed the RangeControl and only kept the UnitControl. Honestly, I'm not sure I like it. Removing the RangeControl objectively reduces the ease of use of these settings.

I'd still be in favor of keeping the combo RangeControl / UnitControl and remove the grid layout to allow more horizontal space.

Also, I would like to fix the inconsistency about the placement of the toggle to switch from preset to custom values. This toggle is placed in different ways across the editor UI. This kind of inconsistencies add unnecessary mental processing and cognitive load for users. Summed to dozens of other small inconsistencies, the current UI is in a state where users are forced to continually process the UI, which doesn't help to provide a good user experience. I think all these inconsistencies should be fixed one by one to provide an UI that is consistent and predictable.

Screenshot: before, after, and a comparison with other range controls where the toggle is placed _after_ the control:

before

afercia:
Last commit still needs a new review. 

On of my questions still needs a reply from design: 

> Just making the list of controls into a longer visual list of items does not improve usability of this particular control.

I'm not sure I understand why a longer list would be a problem. The point here is allowing the range controls to have more horizontal space. No gric. More height, yes. Users can always scroll.

To me, the best option here would be to just place the four controls vertically stacked, to take the whole available width that is necessary for a range control. I don't understand how a grid layout can ebe beneficial in any way for users. I'd appreciate some feedback thanks.

@t-hamano @WordPress/gutenberg-design

jameskoster:
Configuring shadows generally involves working with small values which is a bit fiddly with the range inputs. They're also limited by an arbitrary min/max which is restrictive on those occasions where you want to use larger values.

In my opinion it seems fine to remove them in this case, and just use the regular unit inputs.

packages/edit-site/src/components/global-styles/shadows-edit-panel.js
```diff
@@ -479,67 +475,22 @@ function ShadowPopover( { shadowObj, onChange } ) {
...
-		</VStack>
+		<HStack align="flex-end">
+			<UnitControl
+				label={ label }
+				__next40pxDefaultSize
+				value={ value }
+				onChange={ onValueChange }
+				className="edit-site-global-styles__shadow-editor-control"
+			/>
+		</HStack>
```
t-hamano:
```suggestion
		<UnitControl
			label={ label }
			__next40pxDefaultSize
			value={ value }
			onChange={ onValueChange }
		/>
```

- I think the `HStack` component is no longer necessary.
- The class name doesn't seem to be used.

t-hamano (commented):
From a code perspective, this variable can be removed since it is only used by the shadow control:

https://github.com/WordPress/gutenberg/blob/31ba0ff45de2fc88e7a40e45837a7d6131a5f73c/packages/edit-site/src/components/global-styles/shadow-utils.js#L1

> I'm not sure I understand why a longer list would be a problem.

I can't give a logical reason for this 😅 Maybe I prefer popover content to be as compact as possible. That said, the background image popover is tall, so being tall in itself may not be a problem:

image

I think we can go ahead with either approach, the current one using a grid layout, or the vertical approach.

afercia:
> Configuring shadows generally involves working with small values which is a bit fiddly with the range inputs.

Yes I totally agree that range inputs by their own nature need some large horizontal space to be used effectively.

> They're also limited by an arbitrary min/max which is restrictive on those occasions where you want to use larger values.

Good point, I think this also emerged in nother issues e.g. block margin where a range is used but the value set by users may need to be greater than the range limit. Range inputs usage should be reserver only to values that have a fixed min and max values, because... well it's a _range_. I think range inputs usage should be audited across the entire UI but that's something for a new issue.

afercia:
> From a code perspective, this variable can be removed since it is only used by the shadow control:

Yes but I do see other controls use the same pattern, moving all these objects with lots of values to a separate variable and file. I think it greatly helps readability. See the Spacing sizes control and the Box control (all, axial, etc.).

afercia:
Last commit removes the unnecessary HStack and className.

jameskoster:
Seems good to me:

Screenshot 2024-09-12 at 09 39 53

afercia:
> Seems good to me:

Thanks. Any objection to get a final review and hopefully approval to merge?

t-hamano (approved):
Looks good to me too.

One last thing, we can remove `hasNegativeRange` from [this file](https://github.com/WordPress/gutenberg/blob/cff1bad07dae8ba0003b0c8bcc239d9b515011d0/packages/edit-site/src/components/global-styles/shadows-edit-panel.js) since the `ShadowInputControl` component no longer receives `hasNegativeRange` as a prop.

afercia:
Thanks @t-hamano good catch.

Issue #9612, opened 2018-09-04
The BaseControl (TextControl etc.) label is too wide


afercia:
By default, the BaseControl `<label>` element is styled as a block element with `display: block;`. The controls that extends from it, for example TextControl, inherit the "block" label.

While this may not be a big issue for controls that are rendered in the narrow space (e.g. the sidebar):

<img width="330" alt="textcontrol label 1" src="https://user-images.githubusercontent.com/1682452/45046587-dde1ea80-b076-11e8-9a12-24d81c58c329.png">

it may be a bit annoying when the space is wider, for example in the content area, where the chance to trigger unwanted clicks is higher:

<img width="699" alt="textcontrol label 2" src="https://user-images.githubusercontent.com/1682452/45046589-dde1ea80-b076-11e8-815a-001ed15b77fd.png">

As a small enhancement, maybe worth considering some different styling.

karmatosed:
Assuming it adapts well I think a good iteration so design approving this.

Issue #60836, opened 2024-04-17
BaseControl: Support accessible, flexible layouts


mirka:
## What problem does this address?

We are running into [issues](https://github.com/WordPress/gutenberg/pull/60787) where we want to tweak the layout of the `help` text, or reuse just the styles of the help text. Currently, it requires CSS overrides of BaseControl internals, or copy-pasting styles.

## What is your proposed solution?

For example, export a `BaseControl.VisualHelp` like we do with `BaseControl.VisualLabel`. This must be accompanied by clear accessibility guidance in the docs. This is just a potential solution, there may be better solutions.

ciampo:
https://github.com/WordPress/gutenberg/pull/61382#issuecomment-2240017470 is another example of the need for `BaseControl` to me more flexible.

Ideally, I'd love to reach a point where:

- `BaseControl` is flexible enough to be used on _all_ controls, for maximum consistency
- `BaseControl` can handle composed / multiple / nested controls (see https://github.com/WordPress/gutenberg/issues/42630)

What do you think of those goals?

How can we balance short and long term goals?

ciampo:
Couple more feedback items:

- `StyledHelp` comes with a top margin, which makes it less easy to drop in other components. We should ideally move spacing between components to dedicated grid/flex parents, instead of using margin
- `BaseControl` should be more flexible about allowing consumers to read/write IDs used to label/describe inputs ([comment](https://github.com/WordPress/gutenberg/pull/63751#discussion_r1688429377))

ciampo:
Another note: if we plan for `BaseControl`'s individual component to be used independently (like we do in a bunch of places for `StyledLabel` and `StyledHelp`), we should reduce any differences caused by rendering such components in isolation vs when they're rendered in `BaseControl`.

One example is how white space is not trimmed around `StyledLabel` (see https://github.com/WordPress/gutenberg/issues/63886#issuecomment-2247336435)

mirka:
Closing because these features are now covered in the `Field` component family in `@wordpress/ui`.

PR #14478, opened 2019-03-17
Use inline-block to limit label width to content

sbardian:
## Description
Closes #9612.  Use `display: inline-block` instead of `display: block` to limit `BaseControl label` to width of the content of the label.  

## How has this been tested?
- `npm run test` all tests successful.  
- manually looked at different blocks with labels. Obviously not an exhaustive test, and will be willing to look at anything else to confirm this change. 

## Screenshots
before:
image

after:
image

before: 
image

after: 
image

## Types of changes
Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code is tested.
- [x] My code follows the WordPress code style. 
- [x] My code follows the accessibility standards.
- [x] My code has proper inline documentation. 
- [x] I've included developer documentation if appropriate.

PR #63911, opened 2024-07-24
BaseControl: change label's display: block

### LINKED ISSUE #63886: Typography: Font appearance control sits at different height to line height and other controls
### Description

I noticed that the Font Appearance typography controls currently sit slightly lower in the UI than the other UI elements for Typography.

This is possibly related to some of the changes to the custom select control. I've added a comment over in https://github.com/WordPress/gutenberg/pull/63258#pullrequestreview-2195640697 in case it's a part of that work.

1. Add a Group block to a post
2. In the Typography panel in the block inspector, enable all the typography controls
3. Notice that in Gutenberg `trunk`, the Appearance controls sit slightly lower than the others
4. In WP 6.6.1, the controls are at the correct height

| Gutenberg trunk | WordPress 6.6 |
| --- | --- |
| image | image |

### Environment info

- WP 6.6.1 with Gutenberg trunk running

- [X] Yes

- [X] Yes

ciampo:
## What?

- Fixes https://github.com/WordPress/gutenberg/issues/63886
- Related to https://github.com/WordPress/gutenberg/issues/60836
- Related to https://github.com/WordPress/gutenberg/pull/63137#issuecomment-2210331222

Change the label style of `BaseControl` from `display: inline-block` to `display: block`

## Why?

Using `display: inline-block` wouldn't trim the white space around the label, which would cause inconsistent gaps across the editor based on the layout inherited from the parent element rendering the label.

Switching to `display: block` ensures that the white space is always trimmed consistently.

## How?

- Changed from `display: inline-block` to `display: block`
- Inspected all usages and style overrides, removed unnecessary overrides

## Testing Instructions

The label is publicly exposed via the `BaseControl.VisualLabel` component, and internally via the `StyledLabel` styled component.

Affected components:

- in the components package, `BorderControl`, `BorderBoxControl`, `FormTokenField`, `CustomSelectControl`, `TimeInput`, `TimePicker`, `DateTimePicker`, `FontSizePicker`, `ToggleGroupControl`.
- Outside of the components package: I initially started to list all components that would be affected, but the list was quite large 😅 

I honestly think that there shouldn't be regressions, but I recommend folks to do some smoke testing across Storybook and the editor. I don't think we need extensive testing either — if any layout breaks we will likely find out as we go in the upcoming weeks; it shouldn't compromise the usability of the product and it will be an easy fix.

Also, make sure that #63886 is fixed.

### DIFF
--- a/packages/block-editor/src/components/color-palette/test/__snapshots__/control.js.snap
+++ b/packages/block-editor/src/components/color-palette/test/__snapshots__/control.js.snap
@@ -45,7 +45,7 @@ exports[`ColorPaletteControl matches the snapshot 1`] = `
   font-weight: 500;
   line-height: 1.4;
   text-transform: uppercase;
-  display: inline-block;
+  display: block;
   margin-bottom: calc(4px * 2);
   padding: 0;
 }
--- a/packages/block-library/src/gallery/editor.scss
+++ b/packages/block-library/src/gallery/editor.scss
@@ -70,7 +70,6 @@
 
 .gallery-image-sizes {
 	.components-base-control__label {
-		display: block;
 		margin-bottom: 4px;
 	}
 
--- a/packages/block-library/src/latest-posts/editor.scss
+++ b/packages/block-library/src/latest-posts/editor.scss
@@ -10,10 +10,6 @@
 }
 
 .editor-latest-posts-image-alignment-control {
-	.components-base-control__label {
-		display: block;
-	}
-
 	.components-toolbar {
 		border-radius: $radius-block-ui;
 	}
--- a/packages/block-library/src/video/editor.scss
+++ b/packages/block-library/src/video/editor.scss
@@ -57,10 +57,6 @@
 }
 
 .editor-video-poster-control {
-	.components-base-control__label {
-		display: block;
-	}
-
 	.components-button {
 		margin-right: $grid-unit-10;
 	}
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug Fixes
 
+-   `BaseControl`: change label's `display` to `block`. ([#63911](https://github.com/WordPress/gutenberg/pull/63911))
 -   `ComboboxControl`: Fix ComboboxControl reset button when using the keyboard. ([#63410](https://github.com/WordPress/gutenberg/pull/63410))
 -   `Button`: Never apply `aria-disabled` to anchor ([#63376](https://github.com/WordPress/gutenberg/pull/63376)).
 -   `SelectControl`: Fix hover/focus color in wp-admin ([#63855](https://github.com/WordPress/gutenberg/pull/63855)).
--- a/packages/components/src/base-control/styles/base-control-styles.ts
+++ b/packages/components/src/base-control/styles/base-control-styles.ts
@@ -37,7 +37,7 @@ export const StyledField = styled.div`
 const labelStyles = css`
 	${ baseLabelTypography };
 
-	display: inline-block;
+	display: block;
 	margin-bottom: ${ space( 2 ) };
 	/**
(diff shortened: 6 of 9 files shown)


Mamaduka:
The style was changed from `display: block` to `display: inline-block` in #14478 to resolve #9612.

ciampo:
🤔 I'm not sure I know of a way to trim the whitespace associated to 	`BaseControl`'s `VisualLabel` that doesn't rely on applying styles on the label's parent (which is not an option).

Maybe we could change `VisualLabel` and `StyledLabel` to render a wrapper div, with a `label` inside? This way, we'd have more control on the white space trimming.

andrewserong:
Thanks for following up! For what it's worth, this has solved the alignment issue in #63886:

image

> The style was changed from display: block to display: inline-block in https://github.com/WordPress/gutenberg/pull/14478 to resolve https://github.com/WordPress/gutenberg/issues/9612.

For anyone else following along, the issue described in https://github.com/WordPress/gutenberg/issues/9612 can be tested in this PR by using a narrower viewport width with the Typography controls. In this PR, clicking the whitespace to the right of the Appearance label will redirect focus to the component. The earlier fix in #14478 prevented this.

However, I notice in `trunk` that clicking to the right of the Font or Line Height labels will also allow the click to redirect focus to the input field, so it doesn't look like we have much consistency here in the current state of things:

### Trunk

https://github.com/user-attachments/assets/8c32d1df-a527-47e8-9ca2-76a15f2d8957

### This PR

https://github.com/user-attachments/assets/d69a7333-34f7-46a8-9f20-fd93fda65d00

In some ways this PR nearly makes things feel a bit more consistent to me, but I don't feel strongly either way about the clickable area problem.

> Maybe we could change VisualLabel and StyledLabel to render a wrapper div, with a label inside? This way, we'd have more control on the white space trimming.

If it's worth it to fix the width of the labels to limit the clickable area, then this sounds worth a try to me!

aaronrobertshaw:
As a fix for https://github.com/WordPress/gutenberg/issues/63886, this works for me too 👍 

> The style was changed from display: block to display: inline-block in https://github.com/WordPress/gutenberg/pull/14478 to resolve https://github.com/WordPress/gutenberg/issues/9612.

> However, I notice in trunk that clicking to the right of the Font or Line Height labels will also allow the click to redirect focus to the input field, so it doesn't look like we have much consistency here in the current state of things

Given that it appears #9612 wasn't fully solved and we have some inconsistency there on trunk, I'd lean toward fixing the obvious visual alignment and follow up on the trimming of whitespace separately.

> Maybe we could change VisualLabel and StyledLabel to render a wrapper div, with a label inside? This way, we'd have more control on the white space trimming.

This might be the best bet long term to get the best of both worlds.

Mamaduka:
> Given that it appears https://github.com/WordPress/gutenberg/issues/9612 wasn't fully solved and we have some inconsistency there on trunk, I'd lean toward fixing the obvious visual alignment and follow up on the trimming of whitespace separately.

Good point 💯

tyxla:
> Given that it appears https://github.com/WordPress/gutenberg/issues/9612 wasn't fully solved and we have some inconsistency there on trunk, I'd lean toward fixing the obvious visual alignment and follow up on the trimming of whitespace separately.

I agree with this 👍 

Also, confirming that this tests well across various controls in the post and site editor and resolves #63886.

Mamaduka (approved):

tyxla (approved):
Realized I didn't formally approve the PR, so here goes ✅

ciampo:
Thank you, folks!

If  the size of the clickable label ever becomes problematic, we can test the suggested approach above:

> Maybe we could change VisualLabel and StyledLabel to render a wrapper div, with a label inside? This way, we'd have more control on the white space trimming.

andrewserong:
Thanks for landing this fix, @ciampo! 🙇

PR #63913, opened 2024-07-24
Alternative zoom out trigger location

jeryj:
## What?

I think it will be very useful from UX and accessibility perspectives to have the toggle control for the zoom to be able to be triggered within the patterns tab. It also conditionally sets the zoom out mode based on the current selection. If you are working with no selection or section level block, you enter zoom out mode. If not, your current selection is preserved.

## Why?

Allows for:
- users to remain within the area they are working in without needing to know about entering/exiting a mode they didn't know existed. For example, it's valid to use the patterns inserter without zoom out mode, so let's not assume they want to enter zoom out unless they have an outer level block selected (not implemented yet).
- toggling between zoom in/out without the cognitive load of exiting the pattern area and searching for exiting zoom out within the device control. I also put it there though.
- Simplifies the code, but this isn't a primary reason. Just a benefit.

## How?

- Removes the heading for the pattern panel. It was a div anyways (not a heading element), and the markup for the component is Tab > Tab Panel. So, having the heading is the same as Pattern -> Pattern. It isn't useful since the tab already contains the name of the tab panel.
- Add a toggle control with visually hidden Full Screen label and icon. I think there's likely a better icon and label text. I don't think it's clear enough right now.
- Check if current block selection is a root level or section level block -- if so, enter zoom out mode when opening the pattern panel.

TODO: Do we automatically exit zoom out after closing the panel? I feel like if we toggled it for them, then yes.

## Screenshots or screencast 
Pattern panel with zoom out toggle off
Pattern panel with zoom out toggle on


Mamaduka:
Can we trigger "zoom-out" mode when selecting pattern categories? This is a user event, and it would be nice to avoid the effect of synchronization if we can.

jeryj:
@Mamaduka 
> Can we trigger "zoom-out" mode when selecting pattern categories? 

This PR enters zoom out mode if your block selection is allowed in the zoom out view. So, if you have a top level block selected, you enter zoom out. This is to avoid an interaction unexpectedly changing your block selection. 

> This is a user event, and it would be nice to avoid the effect of synchronization if we can.

Can you explain what you mean more? I'm not sure I follow.

Mamaduka:
@jeryj, it was more general to note how pattern category selection triggers a zoom-out mode. Since pattern category selection is a user event, I wonder if we can move the handler to the same callback.

React docs ["You Might Not Need an Effect"](https://react.dev/learn/you-might-not-need-an-effect) page has really good explanations for this.

jeryj:
@Mamaduka I see what you mean. Thanks for the explanation. In this case the issue is exiting the inserter. We'll need to know if we should automatically return them to their previous mode or keep them in zoom out, and there's a lot of ways to exit the inserter, including some that may run outside this component.  I'll look for ways to simplify if this approach is going to get merged though.

richtabor:
Is this dependent on the experiment for auto-engaging zoom out when selecting a pattern category? Meaning if we do not continue that experiment, this is not valid?

jeryj:
> Is this dependent on the experiment for auto-engaging zoom out when selecting a pattern category? 

No, it would just give you a way to enter zoom out from within the pattern category panel since that's an area where it's most relevant.

paaljoachim:
Thank you for working on this feature!

Some brainstorming.
I am testing through Playground here: https://playground.wordpress.net/gutenberg.html

I have a feeling that a lot of questions are coming up in the back of my mind when I test this... I can not totally place it yet.
This PR is very much connected with https://github.com/WordPress/gutenberg/pull/64179 and https://github.com/WordPress/gutenberg/pull/63870

A new zoom out view and new editing experience with a better perspective on all of the content of a page/post. With a focus on a  container/Group blocks in relation to moving or deleting.

This PR is more focused on shuffling patterns. Could also be Template parts, templates Query Loop patterns etc. 
Zooming out to 50% to get a better perspective on what the content looks like in the specific view.

Perhaps one would be able to go to a page that uses the Query Loop click to go to 50% view and click the Query Loop and begin shuffling patterns to see what different patterns would look like in relation to other content on the page.
There should be an undo button close to the shuffle icon so that one could easily return back to the pattern one had at the beginning before one started shuffling patterns. As one might want to explore but might lose ones way if one can not easily return to the initial pattern.

I find the feature very useful and we need to make sure that there is a consistency between how this works in different areas. Meaning pattern, template parts, templates etc areas.
I am thinking Anne @annezazu and James @jameskoster might also want to take a look at this PR.

jasmussen:
I'd hold off on this one. 

One of the toughest problems we're struggling with, around zoom, is the initial trigger. The dropdown that Rich has been exploring seems very close to threading the needle, give or take the visual of the trigger itself, but at least behavior-wise. Given that hasn't yet been merged and thus tested through a couple of cycles in the plugin, it's unclear whether additional affordances are going to be needed or not. One of the risks then of building this out before we know it's necessary, is the slow accumulation of UI, each of which needs attention.

Issue #63882, opened 2024-07-24
Post Edit Screen Restricted by Metaboxes


PC888:
### Description

Hey guys,

I’ve recently noticed that any meta boxes below the content in the Post Editor restrict the size of the editing window i.e. jam the last block hard against the bottom of the editing window, forcing me to add several blank paragraphs, in order to get some ‘breathing space’. Yoast was the plugin in question but the Custom Fields meta box, toggled on from the block editor preferences, has exactly the same effect. If I deactivate the plugin or toggle off the Custom Field UI the issue goes away.

I opened a ticket here >> https://wordpress.org/support/topic/post-edit-screen-restricted-by-yoast-metabox/#post-17908765 for further info...

Image 1. I’ve enlarged the image slightly to try and make it a bit clearer. You should be able to see how the last <p> block is jammed hard against the bottom of the editor window.
![wp-support-post-editor-image1](https://github.com/user-attachments/assets/dc447ba0-384c-4272-815d-3a225d6176fa)

Image 2. I’ve collapsed the Yoast Meta Box and now the last <p> block is jammed against the bottom of my browser window. The content won’t scroll up to raise the last block to a comfortable working height in the editor. I have to add several empty <p> blocks instead…
![wp-support-post-editor-image2](https://github.com/user-attachments/assets/e97997c7-0da8-42f3-8951-d08814b8004a)

Image 3. I’ve deactivated Yoast and this is how I seem to remember it being…
![wp-support-post-editor-image3](https://github.com/user-attachments/assets/ab5f766b-73bb-4168-9349-f63f5a506305)

I have experienced this in the past and IIRC a subsequent update resolved the issue…

Paul

### Step-by-step reproduction instructions

Blank WP install with TT4 active
Toggle on Custom Fields
Drop some Lorum Ipsum into a new post

### Screenshots, screen recording, code snippet

![wp-support-post-editor-image1](https://github.com/user-attachments/assets/6660c17e-f19c-4293-9cc5-2418e97dadca)
![wp-support-post-editor-image2](https://github.com/user-attachments/assets/9092811a-a3ec-4827-9095-99099ac8e3a9)
![wp-support-post-editor-image3](https://github.com/user-attachments/assets/70e6ec2a-7060-4e85-a3ad-526de587f7a8)

### Environment info

WP 6.6 and TT4, no plugins active, Chrome browser, Mac OS 14.5

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

[...]

PR #63914, opened 2024-07-24
Fix: Post Edit Screen Restricted by Metaboxes 

Rishit30G:
## What?
Fixes issue: https://github.com/WordPress/gutenberg/issues/63882

## Why?
Because the metabox on text editor is causing spacing issues 

## How?
This is solved by adding `padding-bottom: 40vh` using class `block-list-padding-bottom` on `BlockList` component

## Testing Instructions
- Add lot of content to the post 
- Enable Yoast SEO plugin 
- Now we see sufficient space below the text and plugin metabox, like metabox wasn't there

Same

## Screenshots or screencast 
image

### DIFF
--- a/packages/editor/src/components/visual-editor/index.js
+++ b/packages/editor/src/components/visual-editor/index.js
@@ -466,7 +466,8 @@ function VisualEditor( {
 								renderingMode !== 'post-only' ||
 									isDesignPostType
 									? 'wp-site-blocks'
-									: `${ blockListLayoutClass } wp-block-post-content` // Ensure root level blocks receive default/flow blockGap styling rules.
+									: `${ blockListLayoutClass } wp-block-post-content`, // Ensure root level blocks receive default/flow blockGap styling rules.
+								'block-list-padding-bottom'
 							) }
 							layout={ blockListLayout }
 							dropZoneElement={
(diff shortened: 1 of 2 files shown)


PC888:
Thanks @Rishit30G I just tested again with a completely fresh WP install 6.6.1 and TT4 and the issue persists. It's a terrible editing experience!

Rishit30G:
Hey @PC888, 
So sorry to hear that 😞 
But as per my understanding and testing, I was able to fix the issue 
Please let me know what didn't work

PC888:
Sorry @Rishit30G your fix does appear to resolve the issue...

CarloCannas:
Hi, I found this editor behavior annoying too, I'm the one who opened the [ticket in core trac](https://core.trac.wordpress.org/ticket/61677).

I did a bit of digging in the code and found out that there's an explicit exclusion that inhibits the standard `40vh` margin when there are metaboxes.

https://github.com/WordPress/gutenberg/blob/%40wordpress/edit-post%408.4.0/packages/edit-post/src/components/layout/index.js#L135

I think it would be better to remove that instead of adding some new styles and classes.

t-hamano:
Hi @Rishit30G,

Sorry for the delay in reviewing this PR. This issue has already been resolved by #64639.

I would like to close this PR, but please comment if I missed anything. Thank you for your efforts!

PR #63915, opened 2024-07-24
CustomGradientPicker: Support CSS variables

### LINKED ISSUE #28254: Custom Gradient Picker not working with CSS variables
## Description

- I've added a custom gradient preset. 
- It uses CSS variables.
- The circular color picker swatch displays the correct color CSS variable gradient. ✅
- When the color swatch is selected, the CSS gradient including CSS variables is applied to the block (in my theme I'm applying it to a Group). ✅
- The Gradient Picker Range Slider does not display the correct colors from the CSS variable.❌
- Testing with Hex, RGBA, or writing colors all work, just not CSS variables.

1) Register some custom gradient presets and add appropriate CSS

```
add_theme_support(
			'editor-gradient-presets',
			array(
				array(
					'name'     => __( 'Hex' ),
					'gradient' => 'linear-gradient(180deg, #00EE00 0%, #00EEDD 100%)',
					'slug'     => 'hex',
				),
				array(
					'name'     => __( 'RGBA' ),
					'gradient' => 'linear-gradient(180deg, rgba(255,0,0,1) 0%, rgba(255,90,0,1) 100%)',
					'slug'     => 'rgba',
				),
				array(
					'name'     => __( 'Words' ),
					'gradient' => 'linear-gradient(180deg, gold 0%, brown 100%)',
					'slug'     => 'words',
				),
				array(
					'name'     => __( 'CSS Var' ),
					'gradient' => 'linear-gradient(180deg, var(--wp-admin-theme-color) 0%, var(--wp-admin-theme-color-darker-10) 100%)',
					'slug'     => 'css-var',
				),
			)
		);
```

2) Open the editor, add a Group block and apply a gradient background to it

3) The color picker work with the Hex, RGBA and named colors, but the CSS variables are not being picked up by the color picker

## Expected behaviour

CSS variables to get added to the gradient color picker as a range

## Actual behaviour

CSS variables are not added to the gradient color picker as ranges

## Screenshots or screen recording

(The 4th swatch contains the CSS variable gradient)

## WordPress information
- WordPress version: 5.6
- Gutenberg version: Core
- Are all plugins except Gutenberg deactivated? Yes
- Are you using a default theme (e.g. Twenty Twenty-One)? No

## Device information
- Device: Mac
- Operating system: MacOS
- Browser: Safari

### LINKED ISSUE #43715: Selecting a theme gradient doesn't update the gradient bar color
Clicking on a theme gradient doesn't update the gradient bar UI:

Selecting a theme gradient for the Cover block

The isolated instances in Storybook for the GradientPicker component are working fine, so there is likely something wrong with the data flow in the block editor usage.

---

[Reported](https://github.com/WordPress/gutenberg/pull/43577#issuecomment-1226935637) by @jasmussen

mirka:
Fixes #43715
Fixes #28254

## What?

Add CSS variable support to CustomGradientPicker.

## Why?

The [gradient parser](https://github.com/rafaelcaricio/gradient-parser) that we're using under the hood doesn't support CSS variables, and throws an error. We `catch` the error so it doesn't crash, but the UI looks unresponsive. It would be nice if we could read the computed style values of the CSS variables, so we could present the user with the actual colors in the gradient picker so they can be tweaked.

This is not the first time we wanted the computed values of CSS variables (e.g. automated contrast checking for our `Theme` component). I designed the solution in a generic way because think it might be useful in other cases as well.

## How?

Introduce a generic `CSSVariableReplacer` component (internally) that takes a CSS string, gets the computed values of any CSS variables in it, replaces the CSS variables in the original string with computed values, and allows the consumer to retrieve that data via an `onChange` callback. Inserting this component in different locations in the DOM will get the computed style values from that specific context.

Ideally we would use a library for the CSS string parsing, but I couldn't find a suitable one for our specific needs, and a general CSS parser like `postcss` is too big for us to add to the bundle just for this purpose. Fortunately the parsing logic we need here is not too complex, consisting of a standard parentheses matching algorithm and some simple regular expressions. 

## Deferred for future enhancement

In the `onChange` call of CustomGradientPicker, it would be nice if we could maintain the untouched CSS variables even when one of the values are modified. In this PR, all of the variables will flatten into "literal" values once any one of them is modified. (This is the current behavior in trunk, so not a regression.)

### Possible ways to test

- In the unit test file (`packages/components/src/utils/test/css-variables.ts`), try to write a test case that breaks it.
- In Storybook, see the stories for CustomGradientPicker.
- In WP, change the theme to Twenty Twenty-Three. Go to Customize ▸ Styles (`wp-admin/site-editor.php?path=%2Fwp_global_styles`) and choose the "Sherbet" style, which has a theme gradient. In the editor, see the block inspector for a Paragraph block and open the Color ▸ Background popover. In the Gradient tab, you should be able to choose the first Theme gradient. (Unfortunately the other two Theme gradients do not work because [they are invalid](https://github.com/WordPress/gutenberg/issues/43715#issuecomment-2249037975).)

Background color popover for the Sherbet style of TT3


packages/components/src/custom-gradient-picker/index.tsx
```diff
@@ -142,7 +147,19 @@ export function CustomGradientPicker( {
 	onChange,
 	__experimentalIsRenderedInSidebar = false,
 }: CustomGradientPickerProps ) {
-	const { gradientAST, hasGradient } = getGradientAstWithDefault( value );
+	const [ normalizedValue, setNormalizedValue ] = useState(
+		!! getCSSVariablesInString( value ?? '' ).length
+			? 'linear-gradient( #fff, #fff )' // prevent FOUC when there are CSS variables
```
ciampo, line 152: If I understand correctly, this fallback gradient will only be shown for a very small time, until the `onChange` callback of the `CSSVariableReplacer` component fires with the parsed gradient.

Therefore, there shouldn't be any changes to how the gradient falls back to the DEFAULT_GRADIENT in the `getGradientAstWithDefault` function.

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+	/**
+	 * Callback to be called when CSS variables in the given string
+	 * have been replaced with their computed values.
+	 *
+	 * Should be memoized to avoid unnecessary reflows.
+	 */
+	onChange: ( args: {
+		replacedCssString: string;
+		computedVariables: ComputedCSSVariables;
+	} ) => void;
```
ciampo, line 22: For completeness, it would probably be a good idea to also pass back the original string

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			onChange( { replacedCssString, computedVariables } );
+		}
+	}, [ cssString, onChange ] );
+
+	return (
+		<VisuallyHidden>
+			<div ref={ ref } />
+		</VisuallyHidden>
```
ciampo, line 199: Nitty, but do we need the extra `div` ?

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+			try {
+				replacedCssString = replaceCSSVariablesInString(
+					cssString,
+					computedVariables
+				);
+			} catch ( error ) {
+				// eslint-disable-next-line no-console
+				console.warn(
+					'wp.components.CSSVariableReplacer failed to parse the CSS string with error',
```
ciampo, line 187: Should we add the CSS string in the error message?

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+					error
+				);
+			}
+
+			onChange( { replacedCssString, computedVariables } );
```
ciampo, line 192: In case of an error, should we still fire `onChange`? If we do, should we have a flag to signal that parsing failed?

ciampo (commented):
Gave this a first review pass in Storybook!

Also noticed that `CustomGradientPicker` seems to crash when passing `value=""` — which is something that we should definitely fix (even if separately from this PR).

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+	let openingParen;
+	while ( ( openingParen = regex.exec( str ) ) !== null ) {
+		const closingParen = findMatchingParenthesis( str, openingParen.index );
+		const [ start, end ] = [
+			openingParen.index - 'var'.length,
+			closingParen + 1,
+		];
+		const raw = str.slice( start, end );
+		const value = raw.match( /--[\w-]+/ )?.[ 0 ];
```
tyxla, line 73: Should this regex also be declared above so we don't redeclare it every time?

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+		if ( ! value ) {
+			throw new Error( 'No CSS variable found in var() function.' );
+		}
+
+		matches.push( {
+			start,
+			end,
+			raw,
+			value,
+			fallback: raw.match( /,(.+)\)/ )?.[ 1 ].trim(),
```
tyxla, line 84: Same about this regex and the rest of the regexes in here. In addition to the small performance benefit, it can actually make the code more readable if regexes are declared as constants and referenced by name.

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+			.getComputedStyle( element )
+			.getPropertyValue( propertyString );
+		return acc;
+	}, {} as ComputedCSSVariables );
+}
+
+/**
+ * Replace all `var()` functions in a CSS string with their computed values.
+ */
+export function replaceCSSVariablesInString(
```
tyxla, line 111: Have you considered introducing a caching layer here? There seem to be a bunch of regex operations here, and with large CSS strings, caching could potentially have a real, measurable impact. Or is the effect dependency management of `CSSVariableReplacer` enough? Again, I'm concerned that this one could be used directly and then caching won't be leveraged.

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+): string {
+	const varFunctions = findVarFunctionsInString( str );
+
+	let result = '';
+	let lastIndex = 0;
+
+	varFunctions.forEach( ( { start, end, value, fallback } ) => {
+		const replacement =
+			computedVariables[ value ] ||
+			replaceCSSVariablesInString( fallback ?? '', computedVariables );
```
tyxla, line 123: If `fallback` is empty, do we even need to call `replaceCSSVariablesInString`? Can't we just use an empty string without calling `replaceCSSVariablesInString`?

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+			let replacedCssString = cssString;
+
+			try {
+				replacedCssString = replaceCSSVariablesInString(
+					cssString,
+					computedVariables
+				);
+			} catch ( error ) {
+				// eslint-disable-next-line no-console
+				console.warn(
```
tyxla, line 186: Can we use the `warning` package instead? Seems like it's the recommended way to trigger developer errors and warnings now; see https://github.com/WordPress/gutenberg/pull/63610

packages/components/src/utils/css-variables.tsx
```diff
@@ -0,0 +1,201 @@
...
+	let result = '';
+	let lastIndex = 0;
+
+	varFunctions.forEach( ( { start, end, value, fallback } ) => {
+		const replacement =
+			computedVariables[ value ] ||
+			replaceCSSVariablesInString( fallback ?? '', computedVariables );
+
+		if ( ! replacement ) {
+			throw new Error( `No value found for CSS variable ${ value }.` );
```
tyxla, line 126: If this function is part of the public API, throwing an error can be a bit invasive. I see we're throwing it here to catch it in the `CSSVariableReplacer` but it may not be ideal if someone tries to consume `replaceCSSVariablesInString` directly and it errors just like that.

tyxla (commented):
Nice work, @mirka 🙌 I tried with a bunch of test cases, and none were able to break it.

luminuu:
I've tested this with the Greyd WP theme and a new set of gradients with the primary color and can confirm the PR works. The only thing I noticed is that when switching style variations, the colors of the previously selected style variation stay in the gradient picker. I have to reload the editor to get the new colors. This is not the case for colors, these get updated as soon as the style variation is switched. However when I pick one of these, the correct primary color is selected, it's just the preview in the gradient picker that does not get updated.

ciampo:
Hey @luminuu , it would be great if you could share more details about the issue that you're encountering? From a quick read, it sounds like the problem may rely on the consumers of `CustomGradientPicker` and not in `CustomGradientPicker` itself

luminuu:
@ciampo I've made a short video showing this behaviour:

https://github.com/user-attachments/assets/7469219e-6352-46fb-bb73-9eb8e27e3831

This doesn't happen with solid colors, they get updated immediately. It's just the gradients that are stuck with the colors of the previous style variation.

ciampo:
Thank you @luminuu ! Hopefully this can be useful to @mirka when she resumes work on this PR.

t-hamano:
Now that #71186 has been merged, can we close this PR? The gradient-parser library has been updated, so the CustomGradientPicker component now supports CSS variables.

PR #63917, opened 2024-07-25
Try using Popover API to implement responsive navigation menu instead of using Interactivity API

westonruter:
In Core-50182, I [commented](https://core.trac.wordpress.org/ticket/50182#comment:9):

> Related: I was on a slow connection and I loaded the WCEU site but the JavaScript was taking a long time to load. I could see the nav menu button but when I tapped it nothing would happen. Even tapping it 50 times yielded no results (😜). Ideally we wouldn't need JavaScript to open the menu in the first place, which I think is now a real possibility thanks to the Popover API which can be entirely declarative.

This PR attempts to replace the use of the Interactivity API with the native [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) instead. It was pretty easy to implement, however there is one key problem: there is no focus trap for popovers since they are non-modal. In contrast, the `dialog` element _is_ a modal, but it currently requires opening with JavaScript. See [comparison of dialog vs popover](https://web.dev/blog/popover-api). In order to leverage this, I believe we have to wait for [invoker commands](https://open-ui.org/components/invokers.explainer/) to land in the web platform. See also [An update on invokers: Invoker commands in HTML](https://utilitybend.com/blog/an-update-on-invokers-invoker-commands-in-html).

## Screenshots or screencast 

[Screen recording 2024-07-24 17.56.31.webm](https://github.com/user-attachments/assets/40325026-1f69-47f5-9430-5e02d3aa7c68)

### DIFF
--- a/packages/block-library/src/navigation/index.php
+++ b/packages/block-library/src/navigation/index.php
@@ -108,8 +108,7 @@ class WP_Navigation_Block_Renderer {
 	 */
 	private static function is_interactive( $attributes, $inner_blocks ) {
 		$has_submenus       = static::has_submenus( $inner_blocks );
-		$is_responsive_menu = static::is_responsive( $attributes );
-		return ( $has_submenus && ( $attributes['openSubmenusOnClick'] || $attributes['showSubmenuIcon'] ) ) || $is_responsive_menu;
+		return ( $has_submenus && ( $attributes['openSubmenusOnClick'] || $attributes['showSubmenuIcon'] ) );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


afercia:
> I believe we have to wait for [invoker commands](https://open-ui.org/components/invokers.explainer/) to land

More explanation about the `command` and `commandfor` attributes at https://developer.chrome.com/blog/command-and-commandfor.
Chrome 135, [which is being released in the early stable release channel starting yesterday](https://chromereleases.googleblog.com/2025/03/early-stable-update-for-desktop.html), introduces support for these attributes. Hopefully the other major browsers will follow soon.

westonruter:
Also, it would be trivial to implement a polyfill for Invoker Commands for browsers that don't support it yet. In this way, we could go ahead and ship the use of `command` and `commandfor` in the markup, but the Interactivity API `init` code could add event listeners to implement the same functionality to open the `dialog` when Invoker Commands are not available. This would allow for browsers with JavaScript turned off (or where JS failed to load or hasn't loaded yet) which also have Invoker Commands available to be able to open the mobile menu successfully.

westonruter:
I'll likely be closing this PR in favor of @stokesman's PR #69720.

felixarntz:
@westonruter There's also #68755, so we may want to compare the approaches and scope of the changes between that and #69720.

In my PR (which admittedly is so far just experimental), I'm using popover for both the navigation overlay (by default only used on mobile) and for submenus (across all viewports).

One thing to call out is that unfortunately the Interactivity API implementation for the overlay and submenus is very intertwined - maybe this was done to reduce some code to write, but to me that doesn't look like a good choice because there are so many subtle nuances in different behavior. And now it makes decoupling more painful than it should be.

This is partly the reason why I think we need to consider changing both the overlay and the submenus implementation together. Because if you only change one of them to use popover or dialog, most of the Interactivity API code will still stick around for the other one.

westonruter:
See https://github.com/WordPress/gutenberg/pull/69720

PR #63920, opened 2024-07-25
Grid layout: Try moving resizer popover slot to fix display on mobile

### LINKED ISSUE #63275: Grid: resizer/visualizer popover appears over the list view/block tools sidebars in mobile mode
### Description

In narrow viewports (<782px) and when there is a grid layout in the editor canvas, and the corresponding block is selected, the grid resizer and visualizer appear over editor sidebars such as the list view and block tools.

The grid resizer and visualizer sit inside `.components-popover__fallback-container`, which is a popover that is outside the canvas frame.

1. Insert a Grid Group block into the canvas
2. Decrease the window size to < 782px so that the mobile view kicks in
3. Open the list view and select the Group block
4. Observe that the visualizer is visible

https://github.com/WordPress/gutenberg/assets/6458278/53481833-772f-4cb6-8395-19fef0d54e82

1. Now select a grid column and initialize the resizer
2. Open the list view
3. Observe that the resizer is visible

https://github.com/WordPress/gutenberg/assets/6458278/fc52c1fa-c292-4810-bc6a-9d2408051340

| List view | Block tools |
|--------|--------|
|  |  | 

### Environment info

- Gutenberg trunk
- WordPress 6.6 RC 2 (without Gutenberg enabled)

Yes

Yes

andrewserong:
## What?

Part of #57478, might fix https://github.com/WordPress/gutenberg/issues/63275

Similar to https://github.com/WordPress/gutenberg/pull/63389, try updating the Grid resizer to use a different popover slot, that is after the editor canvas.

## Why?

I noticed that as of #63389, the grid visualizer doesn't erroneously appear on mobile when the list view is open. It seems that if we update the resizer to use the same popover slot, then it'll also no longer appear when the list view is open and a child of a grid block is selected.

Note: I'm not too sure if there are any downsides to this approach, so pinging @noisysocks for a second opinion here 🙂

## How?

* Update the popover to use the `"__unstable-block-tools-after"` slot.
* You can see where this slot is output [here](https://github.com/WordPress/gutenberg/blob/6dfaf0de958e1a1defddc1fa4ff82bfa30bf9639/packages/block-editor/src/components/block-tools/index.js#L251)
* And example usage of that component, where you can see that the editor iframe is passed down as `children` to `BlockTools` [here](https://github.com/WordPress/gutenberg/blob/c57b5e43bc27bacf967326d12a5aa1e58c7ac730/packages/block-editor/src/components/block-canvas/index.js#L59)

## Testing Instructions

1. Enable the Grid interactivity experiment in the Gutenberg plugin settings menu
2. Add a Grid block and set it to Manual
3. Give it a bunch of Columns and Rows
4. Add a few paragraph blocks to the Grid block
5. Select one of them
6. Reduce the viewport to mobile screen size and open the list view
7. With a child of the Grid block selected, you should no longer see the resizer box poking through (see screenshots below for an example)
8. Double check that the resizer is otherwise functioning as it is on `trunk`

## Screenshots or screencast 

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/packages/block-editor/src/components/grid/grid-item-resizer.js
+++ b/packages/block-editor/src/components/grid/grid-item-resizer.js
@@ -98,7 +98,7 @@ function GridItemResizerInner( {
 		<BlockPopoverCover
 			className="block-editor-grid-item-resizer"
 			clientId={ clientId }
-			__unstablePopoverSlot="block-toolbar"
+			__unstablePopoverSlot="__unstable-block-tools-after"
 			additionalStyles={ styles }
 		>
 			<ResizableBox


ramonjd:
Just doing a flyby test.

Looks pretty good!

https://github.com/user-attachments/assets/a9b49c88-dcd6-4470-8e94-48841ac04458

There's a comment about this slot that it's meant for the inline rich text, popover - but that was a [while back](https://github.com/WordPress/gutenberg/commit/d03b7568654b5f3a5972a18c8ccb9503d4ed5a4f#diff-b719a654780fb2dd68af5b74ec380db6c6788b13ecf55a10ebb568af8350a47bR131) so maybe not relevant. 

https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/block-tools/index.js#L249-L249

noisysocks (approved):
Good catch. `GridItemResizer` and `GridVisualizer` should use the same slot 👍

andrewserong:
Thanks for confirming, Rob!

PR #63921, opened 2024-07-25
Remove Warning and add notice for Navigation

### LINKED ISSUE #59785: Replace DeletedNavigationWarning in the inspector with a Notice
### Description

When a menu has been deleted but it's still set as the current menu of a Navigation block, a `DeletedNavigationWarning` which is a `Warning` component is shown, both in the editor canvas and in the block inspector. Screenshot:

Screenshot 2024-03-12 at 13 51 02

To me, while this warning design makes sense in the canvas, it's less than ideal in the inspector.

The Notice component would be more appropriate in the inspector:
- I may be missing something but I'm not sure there are other usages of a Warning component _in the inspector_ or at least I haven't seen it, so far.
- A Notice would be more consistent with other Notices shown in the inspector.
- A Notice handles accessibility by the means of an ARIA role, while a Warning component doesn't.

I'd like to propose to change it to a Notice. Screenshot:

Screenshot 2024-03-12 at 14 08 16

- Go to the site editor.
- Select a Navigation block that does have a menu.
- In the Navigation block inspector, go to the Settings tab, scroll down to 'Advanced', expand it and click 'Delete menu'.
- Go back to the Navigation block inspector > List view tab
- Obsrve the warning in the inspector

_No response_

### Environment info

_No response_

Yes

Yes

AKSHAT2802:
Fixes https://github.com/WordPress/gutenberg/issues/59785

Screenshot 2024-07-25 at 1 05 25 PM
Screenshot 2024-07-25 at 1 05 53 PM


packages/block-library/src/navigation/edit/deleted-navigation-warning.js
```diff
@@ -1,24 +1,33 @@
...
-				{
-					button: <Button onClick={ onCreateNew } variant="link" />,
-				}
-			) }
-		</Warning>
+		noticeVisible && (
+			<Notice
+				status="warning"
+				onRemove={ () => setNoticeVisible( false ) }
+				isDismissible
```
Mamaduka: I think this shouldn't be dismissible.

afercia: I'd agree it shouldn't.

afercia (changes requested):
@AKSHAT2802 thanks for working on this and sorry for the late review. In the issue https://github.com/WordPress/gutenberg/issues/59785 I proposed to change only the Warning in the Inspector panel, while this PR changes also the one in the editor canvas. Screenshot:

Screenshot 2024-09-10 at 08 38 38

Also, I'd agree with @Mamaduka that the Notice should not be dismissible.

Could you please have a look and address both points, when you have a chance?

AKSHAT2802:
Hey @afercia, Addressed both points and updated my PR. 
As both the editor and inspector panel deleted navigation warning renders one file, I added a prop named `isNotice` and passed it true while triggering for the inspector panel. 

This will only change the warning to a non-dismissable notice for the inspector panel.
I have verified that the deleted navigation warning only appears in 2 files one in the editor and the second in inspector control.
PFA: Screenshot attached 
Screenshot 2024-09-11 at 1 35 25 AM

packages/block-library/src/navigation/edit/deleted-navigation-warning.js
```diff
@@ -2,12 +2,29 @@
...
+import { Button, Notice } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { createInterpolateElement } from '@wordpress/element';
 
-function DeletedNavigationWarning( { onCreateNew } ) {
-	return (
+function DeletedNavigationWarning( { onCreateNew, isNotice = false } ) {
+	return isNotice ? (
+		<Notice status="warning" isDismissible={ false }>
+			{ createInterpolateElement(
```
afercia: To avoid code duplication, the whole `createInterpolateElement()` should be assigned to a `const` that is then used for both the Notice and the Warning.

afercia (changes requested):
@AKSHAT2802 thanks for the update and the screenshot. Looks good to me, I'd only reduce doce duplication. See inline comment.

AKSHAT2802:
Thanks for the review @afercia & @Mamaduka 
I have added a commit to reduce duplication.

afercia (approved):
@AKSHAT2802 thank you. LGTM to me. Can be merged when all tests pass.

PR #63922, opened 2024-07-25
TemplateContentPanel: Fix the 'getBlocksByName' selector call

Mamaduka:
## What?
PR updates the `getBlocksByName` selector call and passes the block name as a string. The selector supports both strings and an array of strings as arguments.

## Why?
The `getBlocksByName` is a memoized selector passing a new array on each call, will unnecessarily invalidate the cache.

## Testing Instructions
1. Edit a page in the Site Editor.
2. Confirm template part editing is disabled.

Same

## Screenshots or screencast 
CleanShot 2024-07-25 at 11 59 40

### DIFF
--- a/packages/editor/src/components/provider/disable-non-page-content-blocks.js
+++ b/packages/editor/src/components/provider/disable-non-page-content-blocks.js
@@ -45,8 +45,8 @@ export default function DisableNonPageContentBlocks() {
 	);
 	const disabledIds = useSelect( ( select ) => {
 		const { getBlocksByName, getBlockOrder } = select( blockEditorStore );
-		return getBlocksByName( [ 'core/template-part' ] ).flatMap(
-			( clientId ) => getBlockOrder( clientId )
+		return getBlocksByName( 'core/template-part' ).flatMap( ( clientId ) =>
+			getBlockOrder( clientId )
 		);
 	}, [] );
 


packages/editor/src/components/provider/disable-non-page-content-blocks.js
```diff
@@ -45,8 +45,8 @@ export default function DisableNonPageContentBlocks() {
 	);
 	const disabledIds = useSelect( ( select ) => {
 		const { getBlocksByName, getBlockOrder } = select( blockEditorStore );
-		return getBlocksByName( [ 'core/template-part' ] ).flatMap(
-			( clientId ) => getBlockOrder( clientId )
+		return getBlocksByName( 'core/template-part' ).flatMap( ( clientId ) =>
```
ellatrix, line 48: The problem above remains with `contentOnlyBlockTypes`, which is the result of a filter that may return a new array on every call. Is there any way to fix it in the selector instead, so the user of the selector doesn't need to know about it?

ellatrix: Maybe a wrapper `getBlocksByName` shouldn't be memoized, checks for shallow equality, then the underlying function is memoized.

Mamaduka: The filter is called inside `useMemo,` so I think `contentOnlyBlockTypes` should have a stable ref during the component's life cycle.

> Is there any way to fix it in the selector instead, so the user of the selector doesn't need to know about it?

That's a tricky part with memoizing selectors that accept array or object as arguments. I floated an idea of making them [variadic functions](https://github.com/WordPress/gutenberg/pull/63732#discussion_r1689286859), but I've not tested my theory yet.

noisysocks (approved):
If nothing else I think we should prefer passing a string when possible.

PR #63923, opened 2024-07-25
DataViews: Use items with permissions and avoid hooks to register actions

youknowriad:
Related #55083 
Builds on top of #63857 

## What?

This PR updates the different post type actions to use "item with permissions" that way there's no need to call a selector within `isEligilble` 

## Testing Instructions

We need to check the different data views (pages, templates, patterns and template parts) to ensure that we're not regressing and we're actually rendering the right actions with the right permissions.

### DIFF
--- a/packages/edit-site/src/components/page-patterns/use-patterns.js
+++ b/packages/edit-site/src/components/page-patterns/use-patterns.js
@@ -26,9 +26,8 @@ const EMPTY_PATTERN_LIST = [];
 
 const selectTemplateParts = createSelector(
 	( select, categoryId, search = '' ) => {
-		const { getEntityRecords, isResolving: isResolvingSelector } = unlock(
-			select( coreStore )
-		);
+		const { getEntityRecords, isResolving: isResolvingSelector } =
+			select( coreStore );
 		const { __experimentalGetDefaultTemplatePartAreas } =
 			select( editorStore );
 		const query = { per_page: -1 };
--- a/packages/edit-site/src/components/page-patterns/use-patterns.js
+++ b/packages/edit-site/src/components/page-patterns/use-patterns.js
@@ -5,6 +5,7 @@ import { parse } from '@wordpress/blocks';
 import { useSelect, createSelector } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
 import { store as editorStore } from '@wordpress/editor';
+import { useMemo } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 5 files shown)


packages/editor/src/components/post-actions/actions.js
```diff
@@ -1035,14 +985,13 @@ export function usePostActions( { postType, onActionPerformed, context } ) {
 				userCanCreatePostType &&
 				duplicateTemplatePartAction,
 			isPattern && userCanCreatePostType && duplicatePatternAction,
-			supportsTitle && renamePostActionForPostType,
+			supportsTitle && renamePostAction,
```
scruffian, line 988: It looks like some of these need to be added to the dependency array.

scruffian (approved):
I tested deleting patterns and pages and both worked as expected.

I saw this but when deleting a pattern, but it's also there on trunk:

Screenshot 2024-07-26 at 17 00 01

I think you also need to update the dependency array for the use memo, but aside from that I think this is good to go.

packages/edit-site/src/components/page-patterns/use-patterns.js
```diff
@@ -25,8 +26,9 @@ const EMPTY_PATTERN_LIST = [];
 
 const selectTemplateParts = createSelector(
 	( select, categoryId, search = '' ) => {
-		const { getEntityRecords, isResolving: isResolvingSelector } =
-			select( coreStore );
+		const { getEntityRecords, isResolving: isResolvingSelector } = unlock(
+			select( coreStore )
+		);
```
Mamaduka: I think there's no need to unlock this store. All requested selectors are public.

packages/editor/src/components/post-actions/actions.js
youknowriad: they are now top level variables so stable references, the lint rule understands that it seems so they don't need to be added to the dependencies.

packages/editor/src/components/post-actions/index.js
```diff
@@ -29,26 +29,38 @@ const {
...
-		const { getEditedEntityRecord } = select( coreStore );
+		const { getEditedEntityRecord, getEntityRecordPermissions } = unlock(
+			select( coreStore )
+		);
 		const _postType = getCurrentPostType();
+		const _id = getCurrentPostId();
 		return {
-			item: getEditedEntityRecord(
+			item: getEditedEntityRecord( 'postType', _postType, _id ),
+			permissions: getEntityRecordPermissions(
```
Mamaduka, line 41: It looks like we also need to memoize `getEntityRecordPermissions`. Currently, it triggers a warning in the post editor.

## Screenshot
CleanShot 2024-07-30 at 14 03 18

Mamaduka: I'll handle this one. I know what's the issue.

youknowriad: Great thanks :)

Mamaduka: Resolved by #64091.

PR #63927, opened 2024-07-25
DataViews: Optimize the patterns dataviews by extracting the fields definition

youknowriad:
Related #55083 

## What?

When working on #63923 I noticed that the dataviews fields all "remount" when the fields "render" function instance change. This was not visible for us before but it can happen very often as you filter, switch views... 

What it means is that "fields" definition should ideally be a top level module variable rather than something defined within components and that creates a new instance very often. It also forces us to think about fields more consistently across layouts...

This PR optimizes the "patterns" (and template parts) dataviews by moving the fields definition outside the component and into their own file.

I'm not entirely sure that we'll see an impact on our numbers right away but I'm sure this will benefit users and the feeling of performance as you interact with the data views.

## Testing Instructions

The current PR focuses on the "patterns" and "template parts" dataviews, so just ensure there's no regressions there. Especially UI/design regressions for the different fields.

### DIFF
--- a/packages/edit-site/src/components/page-patterns/fields.js
+++ b/packages/edit-site/src/components/page-patterns/fields.js
@@ -54,7 +54,7 @@ function PreviewWrapper( { item, onClick, ariaDescribedBy, children } ) {
 	);
 }
 
-function Preview( { item } ) {
+function PreviewField( { item } ) {
 	const descriptionId = useId();
 	const description = item.description || item?.excerpt?.raw;
 	const isUserPattern = item.type === PATTERN_TYPES.user;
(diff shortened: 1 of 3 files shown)


jameskoster:
I can't comment on the code, but I don't see any regressions.

packages/edit-site/src/components/page-patterns/fields.js
```diff
@@ -0,0 +1,253 @@
...
+			) }
+		</HStack>
+	);
+}
+
+export const titleField = {
+	label: __( 'Title' ),
+	id: 'title',
+	getValue: ( { item } ) => item.title?.raw || item.title,
+	render: ( { item } ) => <Title item={ item } />,
```
ellatrix: Same here, can't this just be `render: Title`?

ellatrix (approved):

PR #63928, opened 2024-07-25
WIP: Fix infinite render of pattern block from call to `setBlockEditingMode`

### LINKED ISSUE #63909: React methods are executed infinited times when block is inserted through a custom pattern
### Description

Custom Gutenberg blocks that uses React native methods such as `componentDidMount()` can break the editor due an infinite loop. Only happens if the block is inserted through a custom pattern. 

Errors in console are:
`Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`

`Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`

This started happening in WordPress 6.6.
In WordPress 6.5 this wasn't happening.

### Create a custom plugin that will load our custom block that includes componentDidMount() and constructor() methods. If you have a ready custom plugin, you can just add these methods in the edit component too. 

File structure of the custom plugin is as follows. Simply create those 2 files in an existing WordPress 6.6 site:

#### wp-content/plugins/custom-plugin/custom-plugin.php

```
<?php
/*
Plugin Name: My Custom Plugin
Description: A custom Gutenberg block for paragraphs with additional settings.
Version: 1.0
Author: @htmgarcia
*/

function my_custom_paragraph_block() {
    // Register the block script
    wp_register_script(
        'my-custom-paragraph',
        plugins_url('block.js', __FILE__),
        array('wp-blocks', 'wp-i18n', 'wp-element', 'wp-editor', 'wp-components', 'wp-block-editor'),
        filemtime(plugin_dir_path(__FILE__) . 'block.js')
    );

    // Register the block type with the editor script
    register_block_type('my-plugin/my-custom-paragraph', array(
        'editor_script' => 'my-custom-paragraph',
    ));
}
add_action('init', 'my_custom_paragraph_block');
```

#### wp-content/plugins/custom-plugin/block.js

```
const { __ } = wp.i18n;
const { RichText, InspectorControls } = wp.blockEditor;
const { PanelBody, TextControl } = wp.components;
const { Component, Fragment, createElement } = wp.element;

class EditComponent extends Component {
    constructor(props) {
        super(props);
        this.onChangeContent = this.onChangeContent.bind(this);
        this.onChangeCustomText = this.onChangeCustomText.bind(this);

        // Let's observe how many times this is outputted
        console.log('constructor()');
    }

    componentDidMount() {
        const { clientId, attributes, setAttributes } = this.props;

        // Let's observe how many times this is outputted
        console.log('componentDidMount()');

        // Set the blockID to the clientId to maybe use later to link to custom CSS
        setAttributes({ blockID: clientId });
    }

    onChangeContent(newContent) {
        this.props.setAttributes({ content: newContent });
    }

    onChangeCustomText(newCustomText) {
        this.props.setAttributes({ customText: newCustomText });
    }

    render() {
        const { attributes: { content, customText } } = this.props;

        return createElement(
            Fragment,
            null,
            createElement(
                InspectorControls,
                null,
                createElement(
                    PanelBody,
                    { title: __('Custom Settings', 'custom-plugin'), initialOpen: true },
                    createElement(
                        TextControl,
                        {
                            label: __('Custom Text', 'custom-plugin'),
                            value: customText,
                            onChange: this.onChangeCustomText
                        }
                    )
                )
            ),
            createElement(
                RichText,
                {
                    tagName: "p",
                    value: content,
                    onChange: this.onChangeContent,
                    placeholder: __('Write your custom paragraph...', 'custom-plugin')
                }
            )
        );
    }
}

// Ensure that this code runs only after the Gutenberg editor has loaded
wp.domReady(function() {
    const { registerBlockType } = wp.blocks;

    // Register the custom paragraph block
    registerBlockType('custom-plugin/my-custom-paragraph', {
        title: __('Custom Paragraph', 'custom-plugin'),
        icon: 'editor-paragraph',
        category: 'common',
        attributes: {
            content: {
                type: 'string',
                source: 'html',
                selector: 'p',
            },
            customText: {
                type: 'string',
                default: '',
            },
            blockID: {
                type: 'string',
            }
        },
        edit: EditComponent,
        save: function(props) {
            const { attributes: { content, customText } } = props;

            return createElement(
                "p",
                null,
                content,
                customText && createElement(
                    "span",
                    null,
                    ' - ',
                    customText
                )
            );
        },
    });
});
```

The parts of the code we're analyzing here are `componentDidMount()` and `constructor()` from `EditComponent` class. I added on purpose a couple of log lines to observe in browser console when inserting "Custom Paragraph" block.

```
class EditComponent extends Component {
    constructor(props) {
        super(props);
        this.onChangeContent = this.onChangeContent.bind(this);
        this.onChangeCustomText = this.onChangeCustomText.bind(this);

        // Let's observe how many times this is outputted
        console.log('constructor()');
    }

    componentDidMount() {
        const { clientId, attributes, setAttributes } = this.props;

        // Let's observe how many times this is outputted
        console.log('componentDidMount()');

        // Set the blockID to the clientId to maybe use later to link to custom CSS
        setAttributes({ blockID: clientId });
    }

    // REST OF THE CODE ...
}
```

### Activate the custom plugin and test "Custom paragraph" block

- Go to Plugins admin page and activate "My Custom Plugin"
- Edit or create a page
- Insert a "Custom paragraph" block and type some text to confirm the block is listed and works

Screenshot 2024-07-24 at 9 15 50 a m

### Create a custom pattern

Convert the "Custom paragraph" block into a pattern by doing click in the 3-dots > Create pattern

Screenshot 2024-07-24 at 9 38 20 a m

Type a name and click "Add"
 
Screenshot 2024-07-24 at 9 39 16 a m

Then the screen will get stuck in the loading process and editor stop working

Screenshot 2024-07-24 at 9 39 23 a m

If you repeat the process by having the browser console opened, you'll see the logs for `componentDidMount()` and `constructor()` are outputted infinite times. 

Screenshot 2024-07-24 at 8 56 11 a m

The problem happens also when you try to insert the custom pattern (because pattern is created but editor can't insert into page).

### Partial solution?

The culprit of the infinite loop according to logs points to `setAttributes()` method inside `componentDidMount()`. If I add a check, the issue seems "solved", however this check is not needed when the block is inserted normally (not as a pattern), plus the logic of the block requires a unique value for blockID for every new instance, meaning the `if()` is messing with it, particularly if I duplicate the block. The value for the clones will be the same (non unique).

```
if (!attributes.blockID) {
    setAttributes({ blockID: clientId });
}
```

This is how 2 instances of the custom block looks like after duplicating the first one. Same value for blockID. So the fix is not good enough. 

Screenshot 2024-07-24 at 9 52 33 a m

I wonder what changed in core to potentially cause an infinity execution of native React methods?

YouTube screencast
https://youtu.be/4B7XiLaWiOM

### Environment info

- WordPress 6.6
- Gutenberg plugin 18.8
- Firefox 128
- macOS Sonoma 14.4.1
- My Custom Plugin

Tested with and without Gutenberg plugin active. Same result.

- [X] Yes

- [X] Yes

talldan:
## What?
WIP

Fixes #63909

## How?

Then follow these steps:
1. Create a synced pattern that has the verse block
2. Insert the synced pattern into a post

In trunk: The pattern and verse block infinitely re-render causing the editor to hang
In this PR: The editor is usable

### DIFF
--- a/packages/block-editor/src/store/actions.js
+++ b/packages/block-editor/src/store/actions.js
@@ -2162,16 +2162,16 @@ export const registerInserterMediaCategory =
  */
 export const setBlockEditingMode =
 	( clientId = '', mode ) =>
-	( { select } ) => {
+	( { select, dispatch } ) => {
 		if ( select.getBlockEditingMode( clientId ) === mode ) {
 			return;
 		}
 
-		return {
+		dispatch( {
 			type: 'SET_BLOCK_EDITING_MODE',
 			clientId,
 			mode,
-		};
+		} );
 	};
 
 /**
--- a/packages/block-editor/src/store/actions.js
+++ b/packages/block-editor/src/store/actions.js
@@ -2160,13 +2160,19 @@ export const registerInserterMediaCategory =
  *
  * @return {Object} Action object.
  */
-export function setBlockEditingMode( clientId = '', mode ) {
-	return {
-		type: 'SET_BLOCK_EDITING_MODE',
-		clientId,
-		mode,
+export const setBlockEditingMode =
+	( clientId = '', mode ) =>
+	( { select } ) => {
+		if ( select.getBlockEditingMode( clientId ) === mode ) {
+			return;
+		}
+
+		return {
+			type: 'SET_BLOCK_EDITING_MODE',
+			clientId,
+			mode,
+		};
 	};
-}
 
 /**
  * Clears the block editing mode for a given block.
--- a/packages/block-library/src/block/edit.js
+++ b/packages/block-library/src/block/edit.js
@@ -6,7 +6,7 @@ import clsx from 'clsx';
 /**
  * WordPress dependencies
  */
-import { useSelect, useDispatch } from '@wordpress/data';
+import { useSelect, useDispatch, useRegistry } from '@wordpress/data';
 import { useRef, useMemo, useEffect } from '@wordpress/element';
 import {
 	useEntityRecord,
@@ -74,18 +74,19 @@ const useInferredLayout = ( blocks, parentLayout ) => {
 	}, [ blocks, parentLayout ] );
 };
 
-function setBlockEditMode( setEditMode, blocks, mode ) {
+function updateInnerBlocksEditingMode( blocks, mode, setBlockEditingMode ) {
 	blocks.forEach( ( block ) => {
-		const editMode =
+		const newMode =
 			mode ||
 			( isOverridableBlock( block ) ? 'contentOnly' : 'disabled' );
-		setEditMode( block.clientId, editMode );
 
-		setBlockEditMode(
-			setEditMode,
+		setBlockEditingMode( block.clientId, newMode );
+
+		updateInnerBlocksEditingMode(
 			block.innerBlocks,
 			// Disable editing for nested patterns.
-			block.name === patternBlockName ? 'disabled' : mode
+			block.name === patternBlockName ? 'disabled' : mode,
+			setBlockEditingMode
 		);
 	} );
 }
@@ -173,6 +174,7 @@ function ReusableBlockEdit( {
 	clientId: patternClientId,
 	setAttributes,
 } ) {
+	const registry = useRegistry();
 	const { record, hasResolved } = useEntityRecord(
 		'postType',
 		'wp_block',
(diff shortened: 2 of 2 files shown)


packages/block-editor/src/store/actions.js
```diff
@@ -2160,13 +2160,19 @@ export const registerInserterMediaCategory =
...
-	return {
-		type: 'SET_BLOCK_EDITING_MODE',
-		clientId,
-		mode,
+export const setBlockEditingMode =
+	( clientId = '', mode ) =>
+	( { select } ) => {
+		if ( select.getBlockEditingMode( clientId ) === mode ) {
+			return;
+		}
```
talldan, line 2168: This is what fixes the bug.

packages/block-library/src/block/edit.js
```diff
@@ -216,19 +218,23 @@ function ReusableBlockEdit( {
...
 	useEffect( () => {
-		setBlockEditMode(
-			setBlockEditingMode,
-			innerBlocks,
-			// Disable editing if the pattern itself is disabled.
-			editingMode === 'disabled' || ! hasPatternOverridesSource
-				? 'disabled'
-				: undefined
-		);
+		registry.batch( () => {
```
talldan, line 221: I figured this should use `registry.batch`? Was it originally an oversight or is it unnecessary?

packages/block-library/src/block/edit.js
```diff
@@ -74,18 +74,19 @@ const useInferredLayout = ( blocks, parentLayout ) => {
 	}, [ blocks, parentLayout ] );
 };
 
-function setBlockEditMode( setEditMode, blocks, mode ) {
+function updateInnerBlocksEditingMode( blocks, mode, setBlockEditingMode ) {
```
talldan, line 77: I made some other changes here, as I found the naming of this function (which was very similar to `setBlockEditingMode`) too confusing to read. Also made some other small readability changes.

talldan:
Actually, this doesn't fix it, it's just that I'd messed up dispatching the action.

PR #63929, opened 2024-07-25
DataViews: Optimize the templates dataviews by extracting the fields definition

youknowriad:
Related #55083 

## What?

When working on #63923 I noticed that the dataviews fields all "remount" when the fields "render" function instance change. This was not visible for us before but it can happen very often as you filter, switch views... 

What it means is that "fields" definition should ideally be a top level module variable rather than something defined within components and that creates a new instance very often. It also forces us to think about fields more consistently across layouts...

This PR optimizes the templates dataviews by moving the fields definition outside the component and into their own file.

I'm not entirely sure that we'll see an impact on our numbers right away but I'm sure this will benefit users and the feeling of performance as you interact with the data views.

## Testing Instructions

The current PR focuses on the templates dataviews, so just ensure there's no regressions there. Especially UI/design regressions for the different fields.

### DIFF
--- a/packages/edit-site/src/components/page-templates/fields.js
+++ b/packages/edit-site/src/components/page-templates/fields.js
@@ -1,0 +1,161 @@
+/**
+ * External dependencies
+ */
+import clsx from 'clsx';
+
+/**
+ * WordPress dependencies
+ */
+import { Icon, __experimentalHStack as HStack } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { useState, useMemo } from '@wordpress/element';
+import { decodeEntities } from '@wordpress/html-entities';
+import { parse } from '@wordpress/blocks';
+import {
+	BlockPreview,
+	privateApis as blockEditorPrivateApis,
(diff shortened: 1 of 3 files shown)


jameskoster:
I'm not seeing any regressions.

packages/edit-site/src/components/page-templates/fields.js
```diff
@@ -0,0 +1,161 @@
...
+	);
+}
+
+export const authorField = {
+	label: __( 'Author' ),
+	id: 'author',
+	getValue: ( { item } ) => item.author_text,
+	render: ( { item } ) => {
+		return <AuthorField item={ item } />;
+	},
```
ellatrix: Can't this be just `render: AuthorField`?

ellatrix (approved):
We should probably update the docs

packages/edit-site/src/components/page-templates/fields.js
youknowriad: Yes

PR #63759, opened 2024-07-19
iAPI Docs: The first three Core Concepts guides

luisherranz:
_EDIT: Initially, in this pull request, I only included the guide "the reactive and declarative mindset". But now, I have included the three guides to merge and review them simultaneously to ensure they are consistent with each other._

## What?

These are the first guides of the Core Concepts section of the Interactivity API, which aims to explain the concepts of:

1. The reactive and declarative mindset.
2. Understanding global state, local context, and derived state.
3. Server-side rendering: How directives are processed on the server.

## Why?

The Interactivity API would benefit from some guides that explain the concepts and mental models, not just from the API references.

## How?

I added a file to a new folder called `core-concepts`, but I'm not really sure how to structure this.

@JuanMaguitar, can you give me a hand? 🙂

On the other hand, I would still like to add a section to the guide with examples of what people should not do (such as mutating the DOM within an action) and the exceptions (such as removing or adding a class to the `body` element).

I think I've also used `we` in some places, and I shouldn't have. @Juanmaguitar, can you confirm this for me as well?

PR #63930, opened 2024-07-25
iAPI Docs: Understanding global state, local context and derived state

luisherranz:
## What?

Following up on [The Reactive and Declarative Mindset guide](https://github.com/WordPress/gutenberg/pull/63759), this is the second guide of the new Core Concepts section of the Interactivity API, focusing on explaining the concept of state and the different types of state (global, local, and derived) used in the Interactivity API.

## Why?

The Interactivity API would benefit from some guides that explain the concepts and mental models, not just from the API references.

## How?

For now, I have simply added the guide as a markdown file.

The work to generate the new Core Concepts section is in the pull request for [the guide for reactive and declarative mindset](https://github.com/WordPress/gutenberg/pull/63759). So, once that pull request is merged, we can change this one to fit that format.

### DIFF
--- a/docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
+++ b/docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
@@ -28,9 +28,9 @@ Let's now dive into each of these concepts to study them in more detail and prov
     ```php
     // Populates the initial global state values.
     wp_interactivity_state( 'myPlugin', array(
-     'counter' => 0,
-     'shown' => false,
-     'highlighted' => true,
+      'counter' => 0,
+      'shown' => false,
+      'highlighted' => true,
     ));
     ```
 
@@ -122,7 +122,7 @@ Let's now dive into each of these concepts to study them in more detail and prov
 
     ```php
     wp_interactivity_state( 'myPlugin', array(
-     'someValue' => 1,
+      'someValue' => 1,
     ));
     ```
 
@@ -266,7 +266,7 @@ Local context is particularly useful when you need independent state for individ
     ?>
 
     <div <?php echo wp_interactivity_data_wp_context( $context ); ?>>
-    	<!-- Child elements will have access to `context.counter` -->
+      <!-- Child elements will have access to `context.counter` -->
     </div>
     ```
 
@@ -409,8 +409,8 @@ In essence, derived state allows you to express relationships between different 
     ```php
     // Populates both the initial global and derived state values.
     wp_interactivity_state( 'myPlugin', array(
-     'counter' => 1, // This is global state.
-     'double' => 2, // This is derived state.
+      'counter' => 1, // This is global state.
+      'double' => 2, // This is derived state.
     ));
     ```
 
@@ -424,7 +424,7 @@ In essence, derived state allows you to express relationships between different 
       'counter' => 1 // This is local context.
     );
     wp_interactivity_state( 'myPlugin', array(
-     'double' => 2, // This is derived state.
+      'double' => 2, // This is derived state.
     ));
     ?>
 
@@ -475,9 +475,9 @@ In essence, derived state allows you to express relationships between different 
       'list'    => array( 1, 2, 3 ),
       'factor'  => 3,
       'product' => function() {
-         $state   = wp_interactivity_state();
-         $context = wp_interactivity_get_context();
-         return $context['item'] * $state['factor'];
+        $state   = wp_interactivity_state();
+        $context = wp_interactivity_get_context();
+        return $context['item'] * $state['factor'];
       }
     ));
     ?>
--- a/docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
+++ b/docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
@@ -1,0 +1,708 @@
+The Interactivity API offers a powerful framework for creating interactive blocks. To make the most of its capabilities, it's crucial to understand when to use global state, local context, or derived state. This guide will clarify these concepts and provide practical examples to help you decide when to use each one.
+
+Let's start with a brief definition of global state, local context and derived state.
+
+-   **Global state:** Global data that can be accessed and modified by any interactive block on the page, allowing different parts of your application to stay in sync.
+-   **Local context:** Local data defined within a specific element in the HTML structure, accessible only to that element and its children, providing independent state for individual blocks.
+-   **Derived state:** Computed values based on global state or local context, dynamically calculated on-demand to ensure consistent data representation without storing redundant data.
+
+Let's now dive into each of these concepts to study them in more detail and provide some examples!
+
+## Global state
+
+**Global state** in the Interactivity API refers to global data that can be accessed and modified by any interactive block on the page. It serves as a shared information hub, allowing different parts of your application to communicate and stay in sync. Global state is the ideal mechanism for exchanging information between blocks, regardless of their position in the DOM tree.
+
+### When to use global state
+
+-   You need to share data between multiple blocks that are not directly related in the DOM hierarchy.
+-   You want to maintain a single source of truth for certain data across your entire application.
+-   You're dealing with data that affects multiple parts of your UI simultaneously.
+-   You want to implement features that are global for the page.
+
+### How to work with global state
+
+-   **Initializing the global state**
+
+    Typically, the initial global state values should be defined on the server using the `wp_interactivity_state` function:
+
+    ```php
+    // Populates the initial global state values.
+    wp_interactivity_state( 'myPlugin', array(
+     'counter' => 0,
+     'shown' => false,
+     'highlighted' => true,
+    ));
+    ```
+
+    In this way, the initial global state values will be used during the rendering of the page in PHP, and the HTML can be populated with the correct values. For example, these directives will produce the following result.
+
+    Original, as written in the PHP file:
+
+    ```html
+    <span
+    	data-wp-bind--hidden="!state.shown"
+    	data-wp-class--is-highlighted="state.highlighted"
+    	data-wp-text="state.counter"
+    	class="my-span"
+    ></span>
+    ```
+
+    Rendered and sent to the browser:
+
+    ```html
+    <span
+    	data-wp-bind--hidden="!state.shown"
+    	data-wp-class--is-highlighted="state.highlighted"
+    	data-wp-text="state.counter"
+    	hidden
+    	class="my-span is-highlighted"
+    >
+    	0
+    </span>
+    ```
+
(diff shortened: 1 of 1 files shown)


SantosGuillamot:
After a first read, the guide looks amazing! 👏 I think these are key concepts to understand while using the Interactivity API and this pull request does a great job explaining them. I might be biased because I was already familiar with the terms, but I like how it is structured. I feel the explanations including "When to use X" and the different examples are really useful.

I'll try to take a deeper look once you consider it ready.

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+
+Let's now dive into each of these concepts to study them in more detail and provide some examples!
+
+## Global state
+
+**Global state** in the Interactivity API refers to global data that can be accessed and modified by any interactive block on the page. It serves as a shared information hub, allowing different parts of your application to communicate and stay in sync. Global state is the ideal mechanism for exchanging information between blocks, regardless of their position in the DOM tree.
+
+### When to use global state
+
+-   You need to share data between multiple blocks that are not directly related in the DOM hierarchy.
```
gziolo, line 17: One crucial aspect that is important to clarify is the nuances of how developers should handle multiple blocks of the same type, e.g., several Image blocks rendered on the page, in contrast to how that would look like in the case where two different block types would like to share the same global state, ex. the Gallery block and the Image block set as the child, or the Navigation block and some completely custom block type that would interact with it from a completely random place on the page.

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+    ));
+    ```
+
+    In this way, the initial global state values will be used during the rendering of the page in PHP, and the HTML can be populated with the correct values. For example, these directives will produce the following result.
+
+    Original, as written in the PHP file:
+
+    ```html
+    <span
+    	data-wp-bind--hidden="!state.shown"
```
gziolo, line 43: For education purposes, will this magically read the value from the `myPlugin` namespace if there is no `wp-data-interactivy="myPlugin"` set higher in the HTML tree?

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+    	actions: {
+    		// ...
+    	},
+    } );
+    ```
+
+    The global state initialized on the server using the `wp_interactivity_state` function is also included in that object:
+
+    ```php
+    wp_interactivity_state( 'myPlugin', array(
```
gziolo, line 124: Was it intentional to reuse the same namespace `myPlugin` as in the first example for declaring server state? Later the docs state:

>  Lastly, all calls to the `store` function with the same namespace are merged together:

In effect, I was wondering if at this moment I should assume that the store contains:
- `counter`, `shown`, `example` - `wp_interactivity_state` firs call
- `isLoading` - `store` first call
- `someValue` - `wp_interactivity_state` second call
- `otherValue` - `store` second call

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+    ```
+
+    Lastly, all calls to the `store` function with the same namespace are merged together:
+
+    ```js
+    store( 'myPlugin', { state: { someValue: 1 } } );
+
+    store( 'myPlugin', { state: { otherValue: 2 } } );
+
+    // All calls to `store` return a stable reference (the same object).
```
gziolo, line 152: What is `the same object` in this context? It would help to clarify it further.

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+
+### Example: Two blocks using global state to communicate
+
+In this example, there are two independent blocks. One displays a counter, and the other a button to increment that counter. These blocks can be positioned anywhere on the page, regardless of the HTML structure. In other words, one does not need to be an inner block of the other.
+
+-   **Count Block**
+
+    ```php
+    <?php
+    wp_interactivity_state( 'myPlugin', array(
```
gziolo, line 190: It's getting confusing to see the same namespace `myPlugin` referenced again. What would be the recommendation for plugin authors when picking the namespace name? Should they use one for their plugin, split by block types?

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+-   **Initializing the derived state**
+
+    Typically, the derived state should be initialized on the server using the `wp_interactivity_state` function in the exact same way as the global state:
+
+    ```php
+    // Populates both the initial global and derived state values.
+    wp_interactivity_state( 'myPlugin', array(
+      'counter' => 1, // This is global state.
+      'double' => 2, // This is derived state.
+    ));
```
gziolo, line 414: What alternatives are there to explain the idea of a derived state with the first example presented? I often saw something similar to the following:
```suggestion
   $counter = 1;
    wp_interactivity_state( 'myPlugin', array(
      'counter' => $counter, // This is global state.
      'double' => 2 * $counter, // This is derived state.
    ));
```
When placing the value, it isn't immediately clear that `double` is computed from other entries in the state.

gziolo (commented):
Great to see this type of guide added as it will help to onboard developers. Thank you so much for working on it 👏🏻 

It sounds like a good direction with the nomenclature of global state vs local context. I did the first pass with the focus on the global state and left my feedback inline. I intend to have a look at local context next, but I might have other priorities in the next days so feel free to proceed based on feedback from other contributors 😄 

One thing that isn't fully clear from this guide is what are the best practices for calling `wp_interactivity_state`. Where devs should locate the code call, inside the main plugin file vs inside the `render_callback` of the block type, and what are the implications of that. That might apply to `store()` calls as well, when I think about it now, too.

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+-   When a part of your global state or local context can be computed from other state values.
+-   To avoid redundant data that needs to be manually kept in sync.
+-   To ensure consistency across your application by automatically updating derived values.
+-   To simplify your actions by removing the need to update multiple related state properties.
+
+### How to work with derived state
+
+-   **Initializing the derived state**
+
+    Typically, the derived state should be initialized on the server using the `wp_interactivity_state` function in the exact same way as the global state:
```
jffng, line 407: Why typically? Because you often depend on values from the server (e.g. block attributes) to calculate the initial values of state?

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
```diff
@@ -0,0 +1,708 @@
...
+      <div>
+        Counter: <span data-wp-text="context.counter"></span>
+      </div>
+      <div>
+        Double: <span data-wp-text="state.double"></span>
+      </div>
+    </div>
+    ```
+
+    In the client, the derived state is defined using getters:
```
jffng, line 441: To me, this is the key difference between derived vs global state, and could be presented much earlier in the guide. That could just be how I learn though.

jffng (commented):
In general this helped my understanding of global state, local context, and derived state. I was able to clean up [the store of something I'm working on](https://github.com/WordPress/gutenberg/blob/c11bfa29252930056ecf75bbd2f7b262d90d8a10/packages/block-library/src/accordion-group/view.js) and feel more confident in the implementation, after reading this guide.

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: I have created a specific guide to explain server-side rendering and how directives are processed on the server. It's part of this PR:

- https://github.com/WordPress/gutenberg/pull/63759

Hopefully, it will answer your question.

(Feedback welcomed on that guide too 😄)

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: Alright, I'll take a look to see if it can be mentioned earlier. Thanks, Jeff!

luisherranz:
I am closing this pull request in favor of this one where I have merged this guide with the other two initial guides:

- https://github.com/WordPress/gutenberg/pull/63759

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: Good suggestion. I will apply it to the guide. Thank you!

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: I want to create a specific guide that explains namespaces and scope.

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: Ok, I'll try to improve it, thanks!

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: Maybe it's not very clear. I'll think it over about the example in this guide. Thanks!

docs/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state.md
luisherranz: No, we can add `data-wp-interactive` to make it clearer, thanks!

Issue #63847, opened 2024-07-23
[Mobile] Empty gap between block settings panel and post title


mrfoxtalbot:
### Description

On mobile, there is a gap between the Block Settings and the post title:

### Step-by-step reproduction instructions

1. Open a post on mobile
2. Go to the block settings
3. Notice how, if you scroll, the content of the pannel can be see in a 3-4px gap

### Screenshots, screen recording, code snippet

https://github.com/user-attachments/assets/af38bd3f-3123-45ca-a7d7-d1941988edfc

### Environment info

- WP 6.6
- Happens with and without the GB plugin (18.8)

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

- [X] Yes

hbhalodia:
Hi @mrfoxtalbot, I have raised the PR for the same - https://github.com/WordPress/gutenberg/pull/63931, It looks like we were subtracting the border width from the total height, the height was 48px, and we have assigned top as 48px only. So we do not need to subtract the border width.

You can take a look at the PR - https://github.com/WordPress/gutenberg/pull/63931, happy to add the changes if any.

Thank You,

mrfoxtalbot:
I commented on the PR, thank you @hbhalodia!

hbhalodia:
Hi @mrfoxtalbot, I had tested the same again and looks like it is fixed and not being reproducible with latest version. I am going ahead and marking it as fixed and closing the issue.

Please feel free to open if the issue still persist on your testing.

Thanks,

PR #63931, opened 2024-07-25
Fix: [Mobile] Empty gap between block settings panel and post title #63847

hbhalodia:
## What?

- Fix Issue - https://github.com/WordPress/gutenberg/issues/63847

## Why?

- This PR removes the extra gap that is present between top panel and block controls, due to less height of the header panel.

## How?

- We have a height defined as `48px`, but in the style there it was added as `47px` by subtracting the border width that was not needed.

## Testing Instructions

1. Open any post.
2. Add any block.
3. Select the block and inspect the mobile view.
4. Open the side panel for block controls and scroll.
5. You would not see the gap between the header panel and the block control.

- NIL

## Screenshots or screencast 

https://github.com/user-attachments/assets/9ad59e47-f632-4279-bd55-30dc9abf236e

### DIFF
--- a/packages/components/src/panel/style.scss
+++ b/packages/components/src/panel/style.scss
@@ -40,7 +40,7 @@
 
 	// This helps ensure the correct panel height, including the border, avoiding subpixel rounding issues.
 	box-sizing: content-box;
-	height: $grid-unit-60 - $border-width;
+	height: $grid-unit-60;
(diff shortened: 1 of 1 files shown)


hbhalodia:
- Closing the PR as linked issue is closed and set to complete.

PR #51773, opened 2023-06-22
Page Content Focus: Default insertion point to the Post Content block

noisysocks:
## What?
Follows https://github.com/WordPress/gutenberg/pull/50857. 

Makes it so that when you are focused on editing a page's content in the site editor and no block is selected, the global inserter will insert blocks into the end of the Post Content block.

## Why?
This mimics the post editor and is what users would expect when editing page content.

## How?
Adds a special case to the existing `__experimentalGetInsertionPoint` selector in the `core/edit-site` store.

## Testing Instructions
1. Go to Appearance → Editor.
2. Select Pages.
3. Click on a page to edit or create one.
4. Select Edit.
5. Open the global inserter.
6. Select a block to insert.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/612155/12679b24-72a8-43b1-a2b5-d14a689008b8

PR #63934, opened 2024-07-25
Zoom Out: Use the block editor for insertion point data

scruffian:
## What?
We should be using the insertion point information from the block editor store, not the editor store.

## Why?
I'm not sure why it was done this way, but getting it from the block editor store means that we can respond to drag and drop events which will help with https://github.com/WordPress/gutenberg/pull/63896.

## How?
Replace `insertionPoint` from the editor store with  `getBlockInsertionPoint` from the block editor store.

## Testing Instructions
1. Open the Site Editor
2. Open the inserter, select the patterns tab and select a category
3. Click on a [+] between sections in zoom out mode
4. Check that the big blue insertion point marker shows
5. Check that block insertion continues to work as expected.

## Screenshot
Screenshot 2024-07-25 at 12 58 14

### DIFF
--- a/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
+++ b/packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { useSelect } from '@wordpress/data';
+import { useSelect, useDispatch } from '@wordpress/data';
 import { useEffect, useState } from '@wordpress/element';
 
 /**
@@ -16,8 +16,9 @@ function ZoomOutModeInserters() {
 	const [ isReady, setIsReady ] = useState( false );
 	const {
 		hasSelection,
+		blockInsertionPoint,
 		blockOrder,
-		insertionPoint,
+		blockInsertionPointVisible,
 		setInserterIsOpened,
 		sectionRootClientId,
 		selectedBlockClientId,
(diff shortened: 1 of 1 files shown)


MaggieCabrera (approved):
Works as expected and the code looks good, thanks for fixing this!

packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
```diff
@@ -86,6 +79,8 @@ function ZoomOutModeInserters() {
 			hoveredBlockClientId === previousClientId ||
 			hoveredBlockClientId === nextClientId;
 
+		const { showInsertionPoint } = unlock( blockEditorDispatch );
```
Mamaduka: I think we can unlock the action once instead of doing it in a loop.

This should also work:

```
const { showInsertionPoint } = unlock( useDispatch( blockEditorStore ) );
```

Edit: The `showInsertionPoint` looks to be a public action. Why are we unlocking it?

Mamaduka:
Btw, the `getInsertionPoint` from the editor store has special handling for page content focus, though I'm not sure if it's relevant for zoom out inserter. See #51773.

packages/block-editor/src/components/block-tools/zoom-out-mode-inserters.js
scruffian: When I did it earlier in the component I got this error from the linter:

`Variables should not be assigned until just prior its first reference. An early return statement may leave this variable unused`

And when it's later I get:

`React Hook "useDispatch" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?`

That's why I split it in two!

Mamaduka: I think it only triggers that error when using `unlock`. It doesn't do it without it. Probably worth investigating separately.

Now that unlocking isn't needed, you can safely do `const { showInsertionPoint } = useDispatch( blockEditorStore );` earlier in the component.

Mamaduka: @scruffian, it looks like you already merged PR, so here's the quick follow-up - #63936

I'm also curious if you tested this with page content focus - https://github.com/WordPress/gutenberg/pull/63934#issuecomment-2250226819.

scruffian: Thanks for the follow up. I was keen to merge so that we can move forward with https://github.com/WordPress/gutenberg/pull/63896.

I'll make sure to also look at page content focus on https://github.com/WordPress/gutenberg/pull/63896.

PR #63935, opened 2024-07-25
Footnotes: only replace attribute if footnotes were detected

ellatrix:
## What?

Currently the function `updateFootnotesFromMeta` (which updates footnotes order when block order changes) seems to always update the block attribute if it's a string or RichTextData instance, even if no footnotes were detected. This is causing some issues with blocks that plain text string with new lines, which RichText will read as line break, and then serialise the line breaks in HTML. To prevent this, we can check if footnotes were detected at all.

An alternative fix is to stop processing string attributes, and only process RichTextData attributes. Footnotes were added before we had this new data type, at the time rich text attributes were also plain strings. But I think some more work is needed to guarantee that rich text attributes are always RichTextData instances.

## Testing Instructions

Install this plugin: https://wordpress.org/plugins/syntaxhighlighter/

Add this in the code editor:

```html
<!-- wp:syntaxhighlighter/code -->
<pre class="wp-block-syntaxhighlighter-code">This is a SyntaxHighlighter code block

1 > 0</pre>
<!-- /wp:syntaxhighlighter/code -->

<!-- wp:paragraph -->
<p>This is a paragraph.</p>
<!-- /wp:paragraph -->
```

Now switch back to visual and add a footnote to the paragraph. The line breaks in the code block should not be converted to BR elements.

### DIFF
--- a/packages/core-data/src/footnotes/index.js
+++ b/packages/core-data/src/footnotes/index.js
@@ -72,6 +72,8 @@ export function updateFootnotesFromMeta( blocks, meta ) {
 					? RichTextData.fromHTMLString( value )
 					: new RichTextData( value );
 
+			let hasFootnotes = false;
+
 			richTextValue.replacements.forEach( ( replacement ) => {
 				if ( replacement.type === 'core/footnote' ) {
 					const id = replacement.attributes[ 'data-fn' ];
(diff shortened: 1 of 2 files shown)


youknowriad:
I didn't test but the changes make sense to me

sgomes (approved):
The code looks good, and the fix works correctly in my testing.

Thank you, @ellatrix! 👍

ellatrix:
Maybe I should add an e2e test with a test block using PlainText

PR #63936, opened 2024-07-25
Zoom out: Get store action outside the loop

Mamaduka:
## What?
A small follow-up to #63934.

We can safely get store actions earlier in the component when not working with private actions.

## Testing Instructions
None. PR just moves code around.


scruffian (approved):
LGTM. My linter didn't like this :)

Issue #47904, opened 2023-02-09
Typography: Add "Text shadow" support


burnuser:
## What problem does this address?
Using text in cover block is always a question of contras between foreground (= text) and background (= image).
Of course, using the overlay is one option.
But if you want the image visible without any filter in it's whole brilliant colorspace, the overly is a problem.
And if the image has very dark and very bright zones - especially with Fixed background - , you have a problem with text contrast. 
![grafik](https://user-images.githubusercontent.com/32030147/217746592-e669a7f7-318f-4aab-b887-f27bff6bb788.png)

## What is your proposed solution?
Using CSS text-shadow with X/Y=0, but some pixels for BLUR would work with every background and text color:
![grafik](https://user-images.githubusercontent.com/32030147/217747920-57c080de-edd4-455a-9082-ca0961a0e6a7.png)

---

Edited by @t-hamano 

We're looking to add initial support for the text-shadow style in #73320. In addition, I have listed the tasks necessary to fully support text shadow.

Core ticket: https://core.trac.wordpress.org/ticket/65370

- [x] Add textShadow style support: https://github.com/WordPress/gutenberg/pull/73320
- [ ] ~Add a UI to control text shadow~
  - ~Implement it as a private component~
  - ~Add a new story~
  - ~What kind of UI would be good to support multiple shadows?~
- [ ] Add a settings schema: to enable controlling the UI display via `theme.json`. https://github.com/WordPress/gutenberg/pull/79584
  - Update`VALID_SETTINGS`
  - Update `theme.json` schema
  - Define default setting (`settings.typography.textShadow`): Should it be `true` by default?
  - Extend the `useHasTypographyPanel`, add a new `useHasTextShadowControl` hook
  - ~Expose text shadow UI to global styles~ 
- [ ] Add a block support flag: likely under `typography.textShadow` https://github.com/WordPress/gutenberg/pull/79584
  - Update `TEXT_SHADOW_SUPPORT_KEY`
  - Update `block.json` schema
  - Update `getTypographyClassesAndStyles()` function
  - Add server-side processing
  - Support for skipping serialization
  - Add a UI for selecting presets
  - Update document: `docs/reference-guides/block-api/block-supports.md`
- [ ] Support presets: likely under `settings.typography.shadow`. https://github.com/WordPress/gutenberg/pull/79584
  - Update `PRESETS_METADATA`
  - Update `theme.json` schema
  - ~Should we support multiple shadows?~
  - Should we define default presets?
[...]

PR #63937, opened 2024-07-25
Add text-shadow support to blocks via theme.json

madhusudhand:
## What?

This change allows to set text shadow to blocks via `theme.json` typography definition.

image

image

## Testing Instructions

1. In `theme.json` under typography settings of any block, set the new property `textShadow` to a valid shadow value.
2. Shadow should be applied and visible in both editor and site frontend.

## Related Issues:

This is a first iteration of #47904


carolinan:
I think text-shadow needs to be added to safecss_filter_attr() ?

mikachan:
Thanks @carolinan, do you know if that would be added in [this file](https://github.com/WordPress/gutenberg/blob/trunk/lib/experimental/kses.php#L97)?

carolinan:
> Thanks @carolinan, do you know if that would be added in [this file](https://github.com/WordPress/gutenberg/blob/trunk/lib/experimental/kses.php#L97)?

Oh I did not know that existed, I am not sure if it should go in there or in lib/compat/version

PR #63938, opened 2024-07-25
Update template preview dimensions in table layout

jameskoster:
Fixes a regression that snuck in somewhere along the way.

## What?
Reduce the size of the template preview in table layout.

## Why?
It is currently much too large.

## How?
* Add a `width` property to `page-templates-preview-field.is-viewtype-table`
  * Also adds a `max-height` to avoid rows growing too large.
* Adjust the column width so that it hugs its contents.

## Testing Instructions
1. Navigate to the templates data view
2. Enable the preview field
3. Observe the updated dimensions

| Before | After |
| --- | --- |
| Screenshot 2024-07-25 at 14 05 54 | Screenshot 2024-07-25 at 14 04 15 |


jameskoster:
@youknowriad @ntsekouras @oandregal when you have a moment could I get a review here? 🙏

packages/edit-site/src/components/page-templates/style.scss
```diff
@@ -37,6 +37,8 @@
 	&.is-viewtype-table {
```
youknowriad: Do we still have these classes? I don't think they exist anymore no? you might want to rebase and adapt your code.

jameskoster:
Oh yes, that was changed in another PR.

youknowriad (approved):
Not entirely sure what to test, but the changes are reasonable.

jameskoster:
In table layout template previews are now smaller than they are on trunk.

PR #63941, opened 2024-07-25
Block Bindings: Move logic to merge `usesContext` outside the reducer

SantosGuillamot:
## What?
Follow up to https://github.com/WordPress/gutenberg/pull/63797

As reported by @gziolo [here](https://github.com/WordPress/gutenberg/pull/63797#discussion_r1691195466), the logic to merge `usesContext` in bindings should be move to a function.

## Why?
To ensure that code doesn't run for all the actions that are dispatched.

## How?
Creating a function to handle that logic and call it in the switch statement.

## Testing Instructions
Unit tests should pass.

### DIFF
--- a/packages/blocks/src/store/reducer.js
+++ b/packages/blocks/src/store/reducer.js
@@ -371,16 +371,21 @@ export function collections( state = {}, action ) {
 	return state;
 }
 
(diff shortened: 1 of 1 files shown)


gziolo (approved):
Thank you for the quick follow-up. Looks great!

PR #62233, opened 2024-06-03
Update instances of text-wrap: pretty to fall back to balance

jasmussen:
## What?

Hopefully fixes #62215.

Safari doesn't support `text-wrap: pretty;`. This PR adds `balance` as a fallback.

## How?

Test the document inspector in Blink and Safari/Webkit, the metadata should wrap correctly:

screenshot of the document inspector

Issue #53322, opened 2023-08-04
Phase 3: WordPress admin redesign 


SaxonF:
https://github.com/WordPress/gutenberg/assets/1072756/8bc0f68c-988f-4a21-ace4-a55cb57d1200

https://github.com/WordPress/gutenberg/assets/1072756/51c422b2-c4d6-4d36-b324-35e25bf33e81

![image](https://github.com/WordPress/gutenberg/assets/1072756/d70f7c54-533d-4a4d-b238-b51a78a4c51c)

_Consider this a tracking issue to monitor progress as we build out and refine different aspects of the admin redesign. If you haven't already, please read through [this](https://make.wordpress.org/core/2023/07/12/admin-design/) and [this](https://make.wordpress.org/design/2023/08/10/admin-design-kickoff/) to gain a better understanding of the project_ 

## Navigation
The navigation paradigm we adopt is a central part of the experience and impacts everything around it e.g. page layout. The site editor currently makes use of a drill-down pattern where as classic admin adopts a flatter, more persistent approach. One of our early priorities should be solidifying navigation. 

- [ ]  ...

## Surfaces
A user generally works on either the front-end or back-end of their site. Sometimes tasks on the back-end affect the front-end and vice versa. How are these two surfaces represented in admin? Site editor introduced the idea of "The Frame" which offers a view of the front-end of your site at all times allowing us to show the impact an action on the back-end has on the front-end. It has the potential to be an iconic/brandable element of WordPress moving forward.

Surfaces can be seen in a similar sort of light as Windows or Volumes which are emphasised in [Vision OS](https://developer.apple.com/design/human-interface-guidelines/windows#visionOS). 

Our goal for this track of work is to explore how and align around how these surfaces should be used and configured in different ways.

- [ ]  ...

## Notifications
How might we elegantly inform users of important information in a way that isn't overwhelming?

- [ ] #67823

## Customisability
We’ve all seen how the WordPress experience can be polluted by ever growing top level navigation, notification stacks and fragmented settings screens. We need to offer a level of customisation that gives site creators (or platforms) the ability to shape the information architecture and character of admin to best suit a particular use case. Even better, how might we make it easier for the community to share their configurations with others?

- [ ] ...

## Backwards compatibility
[...]

PR #63942, opened 2024-07-25
Update template description in table layout

jameskoster:
## What?
Updates some styling details relating to template descriptions in table layout.

## Why?
Improves legibility. Currently descriptions will fill the cell which makes the text troublesome to read on large screens.

## How?
* Adds a `max-width` to `.page-templates-description`.
  * I made a new `$maximum-text-width` var in `_variables.scss` for this purpose as it will likely have applications elsewhere. cc @WordPress/gutenberg-components for thoughts on that.
 * Adds `text-wrap: pretty` to take care of typographical widows.
 * Adds a small margin to `.page-templates-description` in table layout to centrally align the title + description in the cell.

## Testing Instructions
1. Open the Templates data view and switch to Table layout
2. Observe the new template description styling

### Before
Screenshot 2024-07-25 at 14 40 20

### After
Screenshot 2024-07-25 at 14 40 38

### DIFF
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -21,8 +21,6 @@ $default-block-margin: 28px; // This value provides a consistent, contiguous spa
 $text-editor-font-size: 15px;
 $editor-line-height: 1.8;
 $mobile-text-min-font-size: 16px; // Any font size below 16px will cause Mobile Safari to "zoom in".
-$maximum-text-width: 50em;
-
 
 /**
  * Grid System.
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -21,6 +21,7 @@ $default-block-margin: 28px; // This value provides a consistent, contiguous spa
 $text-editor-font-size: 15px;
 $editor-line-height: 1.8;
 $mobile-text-min-font-size: 16px; // Any font size below 16px will cause Mobile Safari to "zoom in".
+$maximum-text-width: 50em;
 
 
 /**
--- a/packages/edit-site/src/components/page-templates/style.scss
+++ b/packages/edit-site/src/components/page-templates/style.scss
@@ -51,16 +51,14 @@
 	}
 }
 
-.dataviews-view-table {
-	.page-templates-description {
-		margin-bottom: $grid-unit-10;
-	}
-}
-
 .page-templates-description {
 	max-width: 50em;
 	text-wrap: balance; // Fallback for Safari
 	text-wrap: pretty;
+
+	.dataviews-view-table & {
+		margin-bottom: $grid-unit-10;
+	}
 }
 
 .edit-site-page-templates {
--- a/packages/edit-site/src/components/page-templates/style.scss
+++ b/packages/edit-site/src/components/page-templates/style.scss
@@ -58,7 +58,6 @@
 }
 
 .page-templates-description {
-	display: block;
 	max-width: 50em;
 	text-wrap: balance; // Fallback for Safari
 	text-wrap: pretty;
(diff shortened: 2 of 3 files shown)


packages/base-styles/_variables.scss
```diff
@@ -21,6 +21,7 @@ $default-block-margin: 28px; // This value provides a consistent, contiguous spa
 $text-editor-font-size: 15px;
 $editor-line-height: 1.8;
 $mobile-text-min-font-size: 16px; // Any font size below 16px will cause Mobile Safari to "zoom in".
+$maximum-text-width: 50em;
```
t-hamano: Is this variable likely to be reused in the future? I can't imagine it being used for anything other than data view descriptions at the moment, so I feel like it's ok to hard-code the value.

packages/edit-site/src/components/page-templates/style.scss
```diff
@@ -52,7 +52,13 @@
 }
 
 .page-templates-description {
-	white-space: normal;
+	display: block;
+	max-width: $maximum-text-width;
+	text-wrap: pretty;
```
t-hamano, line 57: Since `text-wrap: pretty` doesn't seem to be applicable in Safari, I think `text-wrap:balance` is necessary as a fallback (#62233)

packages/base-styles/_variables.scss
jameskoster: I think it could be useful as more of the admin is redesigned (https://github.com/WordPress/gutenberg/issues/53322). But I'm fine to hard-code for now.

packages/edit-site/src/components/page-templates/style.scss
```diff
@@ -51,8 +51,17 @@
...
 
+.dataviews-view-table {
+	.page-templates-description {
+		margin-bottom: $grid-unit-10;
+	}
+}
+
 .page-templates-description {
-	white-space: normal;
+	display: block;
```
t-hamano: Since this element is a child of a flex box, I don't think `display: block;` is necessary.

packages/edit-site/src/components/page-templates/style.scss
```diff
@@ -51,8 +51,17 @@
 	}
 }
 
+.dataviews-view-table {
+	.page-templates-description {
+		margin-bottom: $grid-unit-10;
```
t-hamano: I think it would be simpler to move this style [here](https://github.com/WordPress/gutenberg/pull/63942/files#diff-49ea673a8d53cfa4be932c53a174626307e72d5cee9b41f28277742d8468049dR60). I think the selector `.page-templates-description` is enough.

t-hamano (approved):
I left just two small suggestions but look good to me 👍

packages/edit-site/src/components/page-templates/style.scss
jameskoster: If we do this then the margin will also be applied in Grid and List layouts, which we do not want.

t-hamano: Ah, that's true. How about the following? It may be a matter of preference.

```scss
.page-templates-description {
	max-width: 50em;
	text-wrap: balance; // Fallback for Safari
	text-wrap: pretty;

	.dataviews-view-table & {
		margin-bottom: $grid-unit-10;
	}
}
```

jameskoster: Oh sure, I always forget about that selector.

t-hamano:
LGTM 👍

packages/base-styles/_variables.scss
ciampo: Agreed with the above. Let's have a shared variable only:

- if we're sure it's needed across the whole project
- if we're sure that we can't reuse an existing one (even while updating it to the new value)

PR #63945, opened 2024-07-25
Update field line height across grid / list layouts

jameskoster:
## What?
Update field line height across grid / list layouts.

## Why?
They often render at similar widths, and the font size is the same. It doesn't make sense for the `line-height` to be different in each case.

Currently in List layout the line height is `24px`. In Grid layout it inherits the default line height: `1.4`. This PR updates all field values to use a `20px` line height.

## Testing Instructions
1. In a data view like Templates toggle between List and Grid layouts
2. Ensure the field values have equal line-height across, and display is consistent. The template description is a good field to check with.
3. Ensure there are no regressions.

| Before | After |
| --- | --- |
| Screenshot 2024-07-25 at 15 17 15 | Screenshot 2024-07-25 at 15 15 52 |
| Screenshot 2024-07-25 at 15 17 09 | Screenshot 2024-07-25 at 15 16 07 |

### DIFF
--- a/packages/dataviews/src/layouts/grid/style.scss
+++ b/packages/dataviews/src/layouts/grid/style.scss
@@ -64,8 +64,7 @@
 		.dataviews-view-grid__field-value:not(:empty) {
 			min-height: $grid-unit-30;
 			line-height: $grid-unit-05 * 5;
-			display: flex;
-			align-items: center;
+			padding-top: $grid-unit-05 / 2;
 		}
 
 		.dataviews-view-grid__field {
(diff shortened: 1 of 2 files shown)


richtabor (approved):

jameskoster:
I had to adjust the alignment method: https://github.com/WordPress/gutenberg/pull/63945/commits/c53d802509a03d76f11b22a016be39633861c813. I don't like this method as much, but `display: flex` was breaking truncation in 'row' fields:

Screenshot 2024-07-26 at 10 51 38

PR #63949, opened 2024-07-25
Make Distraction Free not conditional on viewport width

### LINKED ISSUE #63948: Mobile width + Distraction Free: Toolbar shows non-functional style and sidebar buttons
### Description

The editor toolbar shows style and sidebar buttons that don't do anything when on a viewport <960px and distraction free enabled. 

It [seems intentional](https://github.com/WordPress/gutenberg/blob/c460ed3a799302ce85dd4b8b17c1d8becb7a6a9d/packages/editor/src/components/editor-interface/index.js#L129) that distraction free is only meant for wider viewports. So seems there's a bug in disabling it completely, if that's the intent. 

1. `/wp-admin/site-editor.php?canvas=edit`
2. Narrow screen <960px 
3. Enable distraction free 
4. Click the style and sidebar buttons, nothing happens
5. Disable distraction free
6. Works

Also can test by resizing the window to narrow after selecting distraction-free in a wide viewport. 

https://github.com/user-attachments/assets/24a4adbd-d867-4f44-b00f-09e8a24b4eee

### Environment info

Gutenberg trunk at 69899cbb9485bec761a1a32506642ce37058bca7 

- [X] Yes

- [X] Yes

richtabor:
## What?
Fixes https://github.com/WordPress/gutenberg/issues/63948 by: 
- Making Distraction Free available on mobile, instead of disabling it entirely. Distraction free is more than just hiding the editor header.
- Making the editor header render on mobile (why Distraction Free was disabled on mobile in the first place), so that you can still easily go "back" without the hover affordance provided on desktop viewports. 

## Why?
You can end up in odd experiences where Distraction Free is conditionally affecting some UI, but not everything, everywhere, like detailed in https://github.com/WordPress/gutenberg/issues/63948.  

## Testing Instructions

1. Open a post or page. 
2. Enable Distraction Free mode. 
3. Use a small viewport. 
4. See the editor header as expected, with the traits of Distraction Free still available (no hover outlines, no pinned interface items). 

## Screenshots or screencast 
CleanShot 2024-07-25 at 12 11 15

### DIFF
--- a/packages/interface/src/components/interface-skeleton/index.js
+++ b/packages/interface/src/components/interface-skeleton/index.js
@@ -95,7 +95,7 @@ function InterfaceSkeleton(
 	const [ secondarySidebarResizeListener, secondarySidebarSize ] =
 		useResizeObserver();
 	const isMobileViewport = useViewportMatch( 'medium', '<' );
-	const isWideViewport = useViewportMatch( 'wide' );
+	const isLargeViewport = useViewportMatch( 'medium' );
 	const disableMotion = useReducedMotion();
 	const defaultTransition = {
 		type: 'tween',
@@ -144,22 +144,22 @@ function InterfaceSkeleton(
 							className="interface-interface-skeleton__header"
 							aria-label={ mergedLabels.header }
 							initial={
-								isDistractionFree && isWideViewport
+								isDistractionFree && isLargeViewport
 									? 'distractionFreeHidden'
 									: 'hidden'
 							}
 							whileHover={
-								isDistractionFree && isWideViewport
+								isDistractionFree && isLargeViewport
 									? 'distractionFreeHover'
 									: 'visible'
 							}
 							animate={
-								isDistractionFree && isWideViewport
+								isDistractionFree && isLargeViewport
 									? 'distractionFreeDisabled'
 									: 'visible'
 							}
 							exit={
-								isDistractionFree && isWideViewport
+								isDistractionFree && isLargeViewport
 									? 'distractionFreeHidden'
 									: 'hidden'
 							}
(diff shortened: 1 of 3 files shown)


jasmussen:
Good motivation, I'm here for this I may be testing this wrong, however, or my env. is acting up. I'm not seeing a difference between this branch and trunk:

trunk

richtabor:
> I'm not seeing a difference between this branch and trunk

Looks your environment isn't 100%. The editor should still have traits of distraction free, like no hover outlines and the same header controls removed from the header on desktop.

jasmussen (approved):
I see it now. Desktop:

Screenshot 2024-08-01 at 09 20 08

Mobile:

Screenshot 2024-08-01 at 09 20 20

Distraction free definitely seems like something that could increasingly be the default for mobile devices.

packages/interface/src/components/interface-skeleton/index.js
```diff
@@ -95,6 +95,7 @@ function InterfaceSkeleton(
 	const [ secondarySidebarResizeListener, secondarySidebarSize ] =
 		useResizeObserver();
 	const isMobileViewport = useViewportMatch( 'medium', '<' );
+	const isLargeViewport = useViewportMatch( 'medium' );
```
Mamaduka: Suggestion (non-blocking): Haven't tested this. Can we check if it's a large viewport by using negation instead of adding another listener?

```
isLargeViewport = ! isMobileViewport
```

richtabor: Yea, good call. Adjusted it.

Mamaduka (approved):

PR #63950, opened 2024-07-25
Categories List: Add border block support

akasunil:
## What?

Add border block support to the `Categories List` block.

## Why?
`Categories List` block is missing border support.

## How?

## Testing Instructions

- Make sure that `Categories List` block's border is configurable via Global Styles
- Edit template/page, Add `Categories List`  block and Apply the border styles

## Screenshots or screencast 
 
  
[Blog-Home-‹-Template-‹-gutenberg-‹-Editor-—-WordPress (9).webm](https://github.com/user-attachments/assets/d7bd8071-c4ef-49ea-9af5-70fc7e94ae28)

### DIFF
--- a/packages/block-library/src/categories/block.json
+++ b/packages/block-library/src/categories/block.json
@@ -62,6 +62,18 @@
 		},
 		"interactivity": {
 			"clientNavigation": true
+		},
+		"__experimentalBorder": {
+			"radius": true,
+			"color": true,
+			"width": true,
+			"style": true,
+			"__experimentalDefaultControls": {
+				"radius": true,
+				"color": true,
+				"width": true,
+				"style": true
+			}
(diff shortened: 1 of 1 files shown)


carolinan (approved):
I have followed the testing instructions and the border is working as described.

When there is only one category in the list, the content is slightly offset because the list item has a bottom margin of 0.5rem. Then the content is not perfectly vertically centered inside the border. But I don't think that should block this PR.

akasunil:
Thank you @carolinan 

> the content is slightly offset because the list item has a bottom margin of 0.5rem. Then the content is not perfectly vertically centered inside the border.

Spent little time on this, it appears to be theme-specific. When I tested it with an empty theme, the categories list had no margin.

PR #63957, opened 2024-07-25
CustomSelectControl: Restore `describedBy` functionality

### LINKED ISSUE #63568: CustomSelectControl: fix `describedBy` implementation and docs

mirka:
Fixes #63568

## What?

Restores the `describedBy` functionality in CustomSelectControl (v1) that was broken when we switched the underlying implementation in #63258.

## How?

See code comments.

## Testing Instructions

New unit tests should pass. You can also manually inspect the rendered HTML in Storybook.

### DIFF
--- a/packages/components/src/custom-select-control/index.tsx
+++ b/packages/components/src/custom-select-control/index.tsx
@@ -8,6 +8,7 @@ import clsx from 'clsx';
  * WordPress dependencies
  */
 import { useInstanceId } from '@wordpress/compose';
+import { __, sprintf } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
@@ -41,6 +42,15 @@ function applyOptionDeprecations( {
 	};
 }
 
+function getDescribedBy( currentValue: string, describedBy?: string ) {
+	if ( describedBy ) {
+		return describedBy;
+	}
+
+	// translators: %s: The selected option.
+	return sprintf( __( 'Currently selected: %s' ), currentValue );
+}
+
 function CustomSelectControl( props: CustomSelectProps ) {
 	const {
 		__next40pxDefaultSize = false,
@@ -60,7 +70,7 @@ function CustomSelectControl( props: CustomSelectProps ) {
 	);
 
 	// Forward props + store from v2 implementation
-	const store = Ariakit.useSelectStore( {
+	const store = Ariakit.useSelectStore< string >( {
 		async setValue( nextValue ) {
 			const nextOption = options.find(
 				( item ) => item.name === nextValue
@@ -128,9 +138,9 @@ function CustomSelectControl( props: CustomSelectProps ) {
 			);
 		} );
 
-	const renderSelectedValueHint = () => {
-		const { value: currentValue } = store.getState();
+	const { value: currentValue } = store.getState();
 
+	const renderSelectedValueHint = () => {
 		const selectedOptionHint = options
 			?.map( applyOptionDeprecations )
 			?.find( ( { name } ) => currentValue === name )?.hint;
@@ -182,11 +192,11 @@ function CustomSelectControl( props: CustomSelectProps ) {
 			>
 				{ children }
 			</_CustomSelect>
-			{ describedBy && (
-				<VisuallyHidden>
-					<span id={ descriptionId }>{ describedBy }</span>
-				</VisuallyHidden>
-			) }
+			<VisuallyHidden>
+				<span id={ descriptionId }>
+					{ getDescribedBy( currentValue, describedBy ) }
+				</span>
+			</VisuallyHidden>
 		</>
 	);
 }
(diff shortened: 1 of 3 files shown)


packages/components/src/custom-select-control/index.tsx
```diff
@@ -153,23 +174,30 @@ function CustomSelectControl( props: CustomSelectProps ) {
...
+				isLegacy
+				{ ...restProps }
+			>
+				{ children }
+			</_CustomSelect>
+			<VisuallyHidden>
+				<span id={ descriptionId }>
+					{ getDescribedBy( currentValue, describedBy ) }
+				</span>
+			</VisuallyHidden>
```
mirka, line 199: The original implementation passed the `describedBy` prop down to `Button`. Therefore, we want to replicate how Button rendered the `describedBy` text.

https://github.com/WordPress/gutenberg/blob/cad81e428aed155fa3e490cc9c8d9673a170c3fd/packages/components/src/button/index.tsx#L282-L284

packages/components/src/custom-select-control/index.tsx
```diff
@@ -35,6 +42,15 @@ function applyOptionDeprecations( {
...
 }
 
+function getDescribedBy( currentValue: string, describedBy?: string ) {
+	if ( describedBy ) {
+		return describedBy;
+	}
+
+	// translators: %s: The selected option.
+	return sprintf( __( 'Currently selected: %s' ), currentValue );
+}
```
mirka, line 52: This function copies the original logic, which looked like this:

https://github.com/WordPress/gutenberg/blob/9bb75b35ec0daa46e8cd60c46619f522c7edd453/packages/components/src/custom-select-control/index.js#L130-L141

I removed the case for no selection, because that won't happen in this Ariakit-based version.

packages/components/src/custom-select-control/index.tsx
```diff
@@ -48,8 +64,13 @@ function CustomSelectControl( props: CustomSelectProps ) {
...
 	} = useDeprecatedProps( props );
 
+	const descriptionId = useInstanceId(
+		CustomSelectControl,
+		'custom-select-control__description'
+	);
+
 	// Forward props + store from v2 implementation
-	const store = Ariakit.useSelectStore( {
+	const store = Ariakit.useSelectStore< string >( {
```
mirka, line 73: The Ariakit store allows `string | string[]` because it can handle multiple selection. Since this v1 never supported multiple selection, let's narrow the type to just `string`. (This has the effect of narrowing the type of `store.getState().value` to string.)

tyxla (approved):
✅ Code looks good, and this tests well:

|Without `describedBy`|With `describedBy`|
|---|---|
|Screenshot 2024-07-26 at 13 00 42|Screenshot 2024-07-26 at 13 01 03|

afercia:
Please remember to manually add props to @rohitmathur-7 for the alternative PR at https://github.com/WordPress/gutenberg/pull/63884

PR #63958, opened 2024-07-25
Fix: accessibility issue of device preview options

### LINKED ISSUE #63910: Missing semantics for which view option is chosen
### Description

The View options button in the editor header (Desktop, Tablet, Mobile) shows an icon based on which option is selected, and a checkmark within the dropdown menu to show which is selected. There are no semantics to indicate which option is selected though. It is only visual.

- Using a screenreader
- Navigate to the View button
- The announcement will be something like: "View, Button"
- Select it
- Focus will be on the Desktop option
- Navigate between them and notice it does not change the announcement based on the selected option

Screenshot 2024-07-24 at 11 27 02 AM

### Environment info

_No response_

- [X] Yes

- [X] Yes

up1512001:
## What?
- added `aria-label` to desktop, tablet, mobile & preview links.

## Why?
Fixes https://github.com/WordPress/gutenberg/issues/63910

## Testing Instructions
- open editor and click on view to resize now check with screenreader selected view will be accessible.

## Screenshots or screencast 

https://github.com/user-attachments/assets/973d0f2f-aa37-4461-85b2-bdac7405fb60

### DIFF
--- a/packages/editor/src/components/preview-dropdown/index.js
+++ b/packages/editor/src/components/preview-dropdown/index.js
@@ -78,18 +78,21 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
 						<MenuItem
 							onClick={ () => setDeviceType( 'Desktop' ) }
 							icon={ deviceType === 'Desktop' && check }
+							aria-label={ __( 'Desktop View' ) }
 						>
 							{ __( 'Desktop' ) }
 						</MenuItem>
 						<MenuItem
 							onClick={ () => setDeviceType( 'Tablet' ) }
 							icon={ deviceType === 'Tablet' && check }
+							aria-label={ __( 'Tablet View' ) }
 						>
 							{ __( 'Tablet' ) }
 						</MenuItem>
 						<MenuItem
 							onClick={ () => setDeviceType( 'Mobile' ) }
 							icon={ deviceType === 'Mobile' && check }
+							aria-label={ __( 'Mobile View' ) }
 						>
 							{ __( 'Mobile' ) }
 						</MenuItem>
@@ -118,6 +121,7 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
 								className="editor-preview-dropdown__button-external"
 								role="menuitem"
 								forceIsAutosaveable={ forceIsAutosaveable }
+								aria-label={ __( 'Preview in new tab' ) }
 								textContent={
 									<>
 										{ __( 'Preview in new tab' ) }
--- a/packages/editor/src/components/preview-dropdown/index.js
+++ b/packages/editor/src/components/preview-dropdown/index.js
@@ -6,11 +6,12 @@ import {
 	DropdownMenu,
 	MenuGroup,
 	MenuItem,
+	MenuItemsChoice,
 	VisuallyHidden,
 	Icon,
 } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
-import { check, desktop, mobile, tablet, external } from '@wordpress/icons';
+import { desktop, mobile, tablet, external } from '@wordpress/icons';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
 import { store as preferencesStore } from '@wordpress/preferences';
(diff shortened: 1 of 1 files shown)


joedolson:
Thanks for contributing! However, this change won't solve the problem in question. Adding an aria-label means that now the accessible name won't match the visible label. The intention of the ticket is to ensure that whichever of the options is currently selected is announced differently from the unselected items.

Possible solutions could include providing an accessible name for the checkbox (e.g., "selected"); adding a visible text label such as Desktop (selected); or adding an attribute like `aria-pressed="true"` to indicate that the state of the button is pressed.

packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -78,18 +78,21 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
 						<MenuItem
 							onClick={ () => setDeviceType( 'Desktop' ) }
 							icon={ deviceType === 'Desktop' && check }
+							aria-label={ __( 'Desktop View' ) }
```
akasunil, line 81:
```suggestion
							aria-label={ __( 'Desktop view' ) }
```

akasunil: Can we follow the same for other labels ?

up1512001:
@joedolson updated `aria-label` by checking the selected device type.

### video reference

https://github.com/user-attachments/assets/7f4cbbf0-6ed3-484a-946b-c1463459e913

talldan:
Thanks for the PR @up1512001!

I think this UI should be implemented using the `MenuItemsChoice` component as it should provide the correct accessibility out of the box.

The ['Visual Editor' / 'Code Editor' options are an example of that component](https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/components/mode-switcher/index.js) being used:


Screenshot 2024-07-26 at 3 37 44 PM

Screenshot 2024-07-26 at 3 39 32 PM

up1512001:
@talldan replaced `MenuItem` with `MenuItemsChoice`. please review the PR.

#### reference video

https://github.com/user-attachments/assets/2b145db1-3217-488d-9ca4-7f3014eeac11

packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -62,6 +63,54 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
...
+	}
+
+	/**
+	 * Handles the selection of a device type.
+	 *
+	 * @param {string} value The device type.
+	 */
+	const onSelect = ( value ) => {
+		setDeviceType( value );
+	};
```
talldan: The changes so far look good. Right now voiceover announces that the menu item is ticked on focus which is a big improvement over what's in trunk:
Screenshot 2024-07-29 at 12 05 12 PM

But when changing the option (e.g if I select 'Tablet' now) there's no announcement that the option was changed.

So another good accessibility improvement would be to add a spoken message when an option is selected (using the `speak` function from the `a11y` package). This would let screenreader users know that something happened when they change the option.

There's an example here of how the visual/code editor mode option does this:
https://github.com/WordPress/gutenberg/blob/27fbbd8e9b57495339a077ef0c8ba67a18392699/packages/editor/src/store/actions.js#L810

Something similar could be done for the device previews, either as part of this `onSelect` function or within the `setDeviceType` action.

talldan (commented):
Thanks for addressing the feedback. I've left one further comment, then this should be good to ship!

packages/editor/src/components/preview-dropdown/index.js
up1512001: @talldan added `speak` on the device selection.

talldan (approved):
Thanks @up1512001, testing well, code looks good, this looks ready to merge

packages/editor/src/components/preview-dropdown/index.js
```diff
@@ -62,6 +64,61 @@ export default function PreviewDropdown( { forceIsAutosaveable, disabled } ) {
...
+	 */
+	const onSelect = ( value ) => {
+		setDeviceType( value );
+		if ( value === 'Desktop' ) {
+			speak( __( 'Desktop selected' ), 'assertive' );
+		} else if ( value === 'Tablet' ) {
+			speak( __( 'Tablet selected' ), 'assertive' );
+		} else {
+			speak( __( 'Mobile selected' ), 'assertive' );
+		}
```
jeryj, line 119: Thanks for this PR! These announcements shouldn't be necessary. Could you remove these in a follow-up?

up1512001: @jeryj Sure

up1512001: @jeryj created PR to remove `speak` from menu selection.
https://github.com/WordPress/gutenberg/pull/64115

PR #63961, opened 2024-07-25
ColorPalette: Remove extra bottom margin when `CircularOptionPicker` is unneeded

### LINKED ISSUE #61713: Add option to hide CircularOptionPicker when empty

ColorPicker component leaves an empty html element even when there are no options defined.

usage
```
<ColorPalette
    clearable={ false }
    enableAlpha={ enableAlpha }
    value={ color }
    onChange={ ( value ) => {} }
/>
```

image

Remove the elements from the DOM.

mirka:
Fixes #61713

## What?

Fixes an issue where `ColorPalette` renders an unnecessary bottom margin when `colors` is empty and `clearable={ false }` (and `disableCustomColors` is falsey).

Extra margin on ColorPalette

## Why?

This messes up the layout for consumers, and may necessitate style overrides.

## How?

Only render the `CircularOptionPicker` when necessary.

## Testing Instructions

In the Storybook for ColorPalette, set the `colors` prop to an empty array, and `clearable` to false.

You can use the Margin Checker tool in the toolbar:

Margin checker button

### Fixed

- `BorderControl`
- `DuotonePicker`
- `ColorGradientControl` (`@wordpress/block-editor`) — Shown in the flyouts for Global Styles ▸ Colors

### Style override removed

- `ShadowsEditPanel` (`@wordpress/edit-site`) — Shown in the flyouts for Global Styles ▸ Shadows


packages/edit-site/src/components/global-styles/shadows-edit-panel.js
```diff
@@ -422,22 +422,19 @@ function ShadowPopover( { shadowObj, onChange } ) {
 
 	return (
 		<div className="edit-site-global-styles__shadow-editor-panel">
-			<VStack spacing={ 2 }>
+			<VStack spacing={ 4 }>
 				<Heading level={ 5 }>{ __( 'Shadow' ) }</Heading>
```
mirka: @jameskoster I noticed that we usually don't put headings in these flyouts. Do we still need it here? If so, I will adjust the bottom margin so it matches the "Before".

| Before | After |
|--------|-------|
|Shadow flyout, before|Shadow flyout, after|

jameskoster: Good spot, probably fine to remove.

mirka: Result:

Tweaked ShadowsEditPanel

packages/edit-site/src/components/global-styles/style.scss
```diff
@@ -135,16 +135,13 @@
 	}
 }
 
-.edit-site-global-styles__shadow-editor-panel {
+.edit-site-global-styles__shadow-editor__dropdown-content {
 	width: 280px;
-	padding: $grid-unit-20;
```
mirka, line 140: This padding is equivalent to `paddingSize="medium"` on `DropdownContentWrapper`.

tyxla (approved):
Looks good and tests well 👍

packages/edit-site/src/components/global-styles/shadows-edit-panel.js
ciampo: Unrelated to this PR, but I think in general we should make sure that our flyouts (popovers, dialogs, etc) are properly titled — even when there isn't a visible title.

Issue #39064, opened 2022-02-24
Add visual cue when dragging a block into an empty Group block


richtabor:
While we actually can drag and drop another block into an empty Group block, **there's no visual** to indicate that its possible. 

## Proposal
Add a design visual when dragging another block over an empty group (or columns) block. We can use the same DropZone styling and animations to "fill" the empty inserter and add the + icon.

#### Proposal
https://user-images.githubusercontent.com/1813435/236552286-a1213d4c-0777-46a9-b5b8-dc146a754813.mp4

#### Current 
For reference, here is what we have today (no visual indicator):

https://user-images.githubusercontent.com/1813435/155576600-a40dc958-9c7a-4556-acb5-7bf00b1ae82f.mp4

richtabor:
As an aside, we also need a drop indicator for when you're dropping a block into a closed Group/Columns block. It's possible today to do that, but there's no visual indicating it so. Adding this here as it's related design wise.

paaljoachim:
Thanks for adding this issue Rich!
(There was an earlier issue related to drop zones but I am not able to locate it.)

We should use similar drop zones that exist in Gutenberg today. Looking at drag and drop of image.
![Screenshot 2022-04-04 at 15 17 05](https://user-images.githubusercontent.com/5323259/161552189-99128106-ba0e-4338-a502-04fd5fc74a72.jpg)

Here is an example where I only added the blue background color to the Group block.
![Drop zone](https://user-images.githubusercontent.com/5323259/161552273-f3aa78c2-84f4-4163-8d71-ed13f28c4ffb.jpg)

@annezazu 
I added a drag and drop as a comment here https://make.wordpress.org/test/2022/03/31/fse-program-testing-call-13-authoring-an-author-template/#comment-2394

richtabor:
I also think it's a good idea to match the current drag-and-drop UX — I made a couple adjustments based on your feedback @paaljoachim: 

<img width="961" alt="group" src="https://user-images.githubusercontent.com/1813435/170712826-156b53fd-779f-4ee7-8386-0765017f3cfa.png">

paaljoachim:
Looks good Rich!

Any additional feedback here @jasmussen Joen?

annezazu:
@paaljoachim Joen is on sabbatical right now just so this doesn't hold things up.

paaljoachim:
Great! Thanks for letting me know @annezazu 

The method we have right now is consistent with the current UX method for drag and drop.
[...]

Issue #34943, opened 2021-09-19
Consider Group blocks to be removed if all child blocks are removed


richtabor:
## What problem does this address?
Empty Group blocks are difficult to select and can cause "holes" in a page's design that are hard to identify (as they are invisible when empty and not selected). 

## What is your proposed solution?
Consider Group blocks to be removed if all child blocks are also removed, as:

1. There is no purpose for empty Group blocks on a page
2. They cause the problems above

I was inspired by this through how Figma handles empty groups — which are similar in theory: 

https://user-images.githubusercontent.com/1813435/133946243-f4892756-9188-4907-b4d0-30cc20f74fb7.mp4

talldan:
I think there has been similar feedback and a similar issue for the button block too. Would definitely be good to explore 👍

jasmussen:
Works so well in Figma! A question to ponder, though: if you start by inserting a group block, then deselect and save the page, what happens then? Arguably what makes it work in Figma is that the group is mostly an organizational tool secondary to the content.

karmatosed:
I also totally agree if you have added content and then removed it, an empty group block doesn't make sense. Can the group block be aware though of the state either being that it did contain content or was being used as a placeholder by just being added? That I feel is what is being suggested by you sort of @jasmussen where someone might just add blocks to a page?

ellatrix:
Ha! This is similar to #40979, which would be fixed by #42503. I was wondering if the behaviour could be generalised so we can remove the inner blocks setting.

jasmussen:
It would be tricky for Columns, where each Column inside would technically start out empty. There's also this state if you insert an empty group:

<img width="680" alt="Screenshot 2023-08-16 at 16 57 38" src="https://github.com/WordPress/gutenberg/assets/1204802/88aede93-899e-4a18-ad5d-b2cd26b9f50a">

So I wonder if the potentially generalized solution shouldn't be: if a container already has content and you remove the last block, _then_ the container can be removed as well. Tricky bit here would be that if you remove the last block from the last Column inside of a Columns block, both the Column and the Columns block should be remove. Right?

richtabor:
> Tricky bit here would be that if you remove the last block from the last Column inside of a Columns block, both the Column and the Columns block should be remove. Right?

I think perhaps the rule only works per one nested level up. It wouldn't delete the Columns block, just the column (which it already does).

jasmussen:
Actually that seems fine:

[...]

PR #63962, opened 2024-07-25
Remove empty column blocks from mobile in editor, to match front-end

### LINKED ISSUE #63933: Do not render empty column blocks in the editor when viewing mobile and stacked
I propose that an empty columns block should not render when columns are set to stack on mobile. On the front-end these do not take up space, nor distort the layout. Completing this would mean the front-end and editor instances are the same, meeting expectations. 

One note is that the empty column should be visible when the parent Columns block is selected. This way you can put something in the empty column, but it's not always distorting layout.  

### Editor, desktop: 

CleanShot 2024-07-25 at 07 39 27

### Editor, mobile (bug visually distorting layout): 
CleanShot 2024-07-25 at 07 39 39

### Frontend, mobile (no empty column taking up space):
CleanShot 2024-07-25 at 07 40 22

richtabor:
## What?
Fixes https://github.com/WordPress/gutenberg/issues/63933 by not always rendering an empty column block when viewing the mobile preview in the editor.  

## Why?
The front-end and editor instances of the Columns block should be exactly the same. That way, I can design in the editor and it renders on the front-end exactly like I intended. 

## Testing Instructions

1. Open a post or page.
2. Add a columns block with two columns. You can use [this block markup](https://gist.github.com/richtabor/5f52c1e8d2610b275bc1fd988671f718) if you'd like. I colorized the columns to make them clear. 
3. Switch to mobile
4. See no blank space where the empty column is. 
5. Check that the front-end mobile view matches. 
6. Go back to the editor. 
7. Select the parent Columns block. 
8. Check you can insert blocks into the empty column when the parent columns block is selected or the block itself is selected. 

## Screenshots or screencast 

https://github.com/user-attachments/assets/2b7ffca5-5307-4cb5-ad74-b6910d95e639


jameskoster:
This makes sense to me. 

There is a quirk though; if all columns are empty then nothing appears on the canvas, and you have to use list view to select the Columns block.

Should we use the Group approach there (show a dotted outline when empty to aid on-canvas selection)?

richtabor:
> There is a quirk though; if all columns are empty then nothing appears on the canvas, and you have to use list view to select the Columns block.

Yes, that's like it is in trunk. 

I'm not confident that empty columns should be rendered with outlines in the editor, unless the parent is selected. Perhaps it could be _any_ parent? Otherwise, you'd see empty columns in areas that are intended to be empty, messing with the 1:1 with the front-end expectation. But empty columns should have a hover/focus effect so there is some affordance it exists, when you're intending to engage with it. 

The Group block should perform this same way imo.

t-hamano (commented):
While I agree with matching the editor and the frontend, this approach seems to cause some issues.

- In the mobile layout, I can't move blocks into an empty Column block via dragging.
- When the Columns block is completely empty, I can't click on it with the mouse because its height is zero.

If you want to achieve a layout like the screenshot, how about giving explicit widths to the children of the Row/Columns block?

https://github.com/user-attachments/assets/d64e5eff-53b0-45fe-9fa6-6678099545a0

In any case, this needs to receive accessibility feedback.

richtabor:
> In the mobile layout, I can't move blocks into an empty Column block via dragging.
> When the Columns block is completely empty, I can't click on it with the mouse because its height is zero.

I'd consider both of these edge cases, and worth the case that the mobile preview looks like it should.

richtabor:
> If you want to achieve a layout like the screenshot, how about giving explicit widths to the children of the Row/Columns block?

I'm not following this?

richtabor:
> In any case, this needs to receive accessibility feedback.

It's unclear to me why this needs a11y feedback. The edge case of dragging into  columns on mobile seems a bit out there (and not a11y related). If the block is selected, you can add blocks to it all the same as in desktop. 

Currently, you view a mobile view of your site but see a lot of spacing within the editor (if you have empty columns, which is not uncommon). So you try to fix it, but it's not actually a problem, making things quite confusing. Also, makes the columns block seem not properly responsive. I want to fix this, which is significant to building confidence in the editing experience.

t-hamano:
> - In the mobile layout, I can't move blocks into an empty Column block via dragging.
> - When the Columns block is completely empty, I can't click on it with the mouse because its height is zero.

To me, these are not edge cases that can be ignored, but rather lead to a decrease in usability and accessibility.

Also, I think it's better to have PRs that affect screen reader reading, keyboard operation, and mouse operation undergo accessibility checks as much as possible.

cc @joedolson @alexstine (@afercia might be AFK right now)

richtabor:
> To me, these are not edge cases that can be ignored, but rather lead to a decrease in usability and accessibility.

It 100% is an edge case to drag a block into a column while viewing the mobile view in the editor.

It's actually not readily possible to drag blocks when editing mobile—there are no drag handles and the toolbar is pinned to the top. 

CleanShot 2024-07-31 at 08 38 42

> When the Columns block is completely empty, I can't click on it with the mouse because its height is zero.

You can actually, it's just not ideal. 

This is also an edge case (maintaining a columns block with no items in it on a page). The empty blocks are also still accessible from the List View, selecting either the parent, or the individual columns blocks. 

https://github.com/user-attachments/assets/9f6ca66a-c6a0-4fea-a218-a246a9a19ee1

richtabor:
@t-hamano is there a way to check if the columns block's columns are empty? If so, then the empty state of an entirely empty columns block (no blocks in the columns) could remain like it is on desktop. 

Also, this is related to https://github.com/WordPress/gutenberg/issues/34943 (should an empty Group or Columns block persist).

t-hamano:
> is there a way to check if the columns block's columns are empty?

Yes, I added `isAllColumnsEmpty` so please try it. It will have `true` if all columns are empty.

> It 100% is an edge case to drag a block into a column while viewing the mobile view in the editor.
> It's actually not readily possible to drag blocks when editing mobile—there are no drag handles and the toolbar is pinned to the top.

My explanation for this was incorrect. I was referring to the narrow editor canvas width, not mobile devices.

If the editor canvas width is less than 781px, columns are stacked vertically. In this case, you can't drag blocks into columns, as shown in the video below.

https://github.com/user-attachments/assets/7c9b03db-3581-453a-aa0d-504fd86f951e

richtabor:
> Yes, I added isAllColumnsEmpty so please try it. It will have true if all columns are empty.

How can we pass that through to the column blocks?

richtabor:
> My explanation for this was incorrect. I was referring to the narrow editor canvas width, not mobile devices.

> If the editor canvas width is less than 781px, columns are stacked vertically. In this case, you can't drag blocks into columns, as shown in the video below.

I still don't think it is normal/expected to drag items underneath each other and expect them to be in a column on desktop.

It's a much worse—and much more prevalent— issue to have an incorrect mobile view when editing.

jasmussen:
There's a larger question at play here, which looks a bit to the future when the block editor is more refined. Should you be able to design your desktop site, from your phone? The answer is nuanced, and will likely be assisted by locked-down patterns, but also ways to preview _the desktop breakpoint_ from smaller physical devices. However given that, I'd tend to agree with Rich's assessment, that in absence of the ability to preview a desktop view from a physical mobile device, we should at least treat WYSIWYG as a goal in itself, an increasingly important one at that, since block themes are getting more powerful. In that vein, I'd think 1:1 parity with the frontend to be the most important aspect to optimize for.

Mamaduka:
> How can we pass that through to the column blocks?

We can also do similar checks in child blocks. I've updated the selector in [ab74c04](https://github.com/WordPress/gutenberg/pull/63962/commits/ab74c04eca12a8376bef16b54adceb52d622c12b).

I've not touched the appender logic because I'm not 100% sure what we want to achieve here.

t-hamano:
@Mamaduka Thanks for the update!

@richtabor I made the columns block retain its height if all columns are empty. Is this what you intended?

https://github.com/user-attachments/assets/4ef08761-eb15-48d3-b5f4-aa8538d4d859

richtabor:
> @richtabor I made the columns block retain its height if all columns are empty. Is this what you intended?

Yes! This works quite well now. 

1. If no blocks are inserted into any column, the columns block is rendered. 
2. If there is a block within any column, then the empty column block is not rendered on mobile—unless the parent columns block is selected, or the empty column block is selected. 

Now, you can still select/edit appropriately, and an empty columns block does not distort the mobile view.

t-hamano:
[This E2E test](https://github.com/WordPress/gutenberg/blob/bb29c2bc6153bfc7ce1cb07eb75f1bc723d401d7/test/e2e/specs/editor/various/draggable-blocks.spec.js#L341) is failing and needs to be addressed.

This test ensures that we can drag and drop blocks into empty columns. However, presumably because the browser width emulated in the E2E test is narrow, the columns are stacked vertically. The empty column block has no height, so we cannot drag and drop blocks into it:

https://github.com/user-attachments/assets/15ade40a-3f82-4638-8e4e-338eae14b0c7

This is my concern in [this comment](https://github.com/WordPress/gutenberg/pull/63962#issuecomment-2259537136), but is this not a problem?

For example, some monitors may have a low resolution of around 1280. On such a monitor, if you open both the List View and the Block Sidebar, the iframee editor canvas width will be less than 781px, so the column blocks will be stacked vertically.

richtabor:
> This is my concern in https://github.com/WordPress/gutenberg/pull/63962#issuecomment-2259537136, but is this not a problem?

> For example, some monitors may have a low resolution of around 1280. On such a monitor, if you open both the List View and the Block Sidebar, the iframee editor canvas width will be less than 781px, so the column blocks will be stacked vertically.

I don't think this is a problem. I still consider this an edge case that is accounted for by the work we added to check if columns are empty (then render them entirely), and also render columns one selection. The only caveat being that you need to have a columns block selected to add something to it, on mobile.

t-hamano:
I tried to fix the E2E tests.

Personally, I would hesitate to ship this PR, but if enough people feel this approach makes sense, it may be worth a try. However, if this behavior is reported as a bug after we ship this PR, we might want to consider reverting it.

richtabor:
@Mamaduka could you take a look at these e2e tests? 

>  it may be worth a try. 

I'm confident this is a better solution than what we have in trunk. I'm down to give it a go.

t-hamano:
I think I've resolved the E2E test failure.

The root cause is that the block appender is not rendered when the Column block has a non-empty column block. I know this is an intentional change in this PR.

This E2E test tests whether an Image block can be dragged into the second of three columns:

empty-column


However, no block appender is rendered for the second column, so the test fails as it cannot find the appender [here](https://github.com/WordPress/gutenberg/blob/f42d7753866d5b870931489fef1d2ab204d65413/test/e2e/specs/editor/various/draggable-blocks.spec.js#L433-L440).

To solve this, I inserted a Columns block where all three columns are empty. This ensures that the column block has a block appender, so the test should pass:

image

One other thing I want to mention is that another purpose of this E2E test is to ensure a blue visual cue appears when we drag a block:

image

However, no _visible_ visual cues are present in the current Column block. The test does pass, but internally it [only checks whether the CSS for visual cues has been applied](https://github.com/WordPress/gutenberg/blob/602fd05f5ae04b4fa7d7371eea59dc3ffeac3e9c/test/e2e/specs/editor/various/draggable-blocks.spec.js#L431-L436).

I know that the current behavior of empty blocks and block appenders is intentional, but it feels like the current editor experience is straying too far from the editor experience proposed in #39064.

What do you think?

richtabor:
> I know that the current behavior of empty blocks and block appenders is intentional, but it feels like the current editor experience is straying too far from the editor experience proposed in https://github.com/WordPress/gutenberg/issues/39064.

In this case it's a bit different. I do think the columns empty state could be improved as part of a separate effort though. 

@t-hamano Is this something you would be comfortable seeing through? If not, I will probably close this.

richtabor:
I do still think this is critical to making better decisions on the mobile display of contents, though I do not currently have the bandwidth to pursue this. If anyone would like to re-open and push it forward, please do so. Thanks.

Issue #53468, opened 2023-08-09
New WP 6.3 CSS is overriding theme styles and styles from other plugins


kubiqsk:
### Description

There are some new CSS rules and they break layouts, eg.:
```css
:where(body .is-layout-flow) > :last-child:last-child {
	margin-block-end: 0;
}
```
or similar one with duplicated `:first-child`.

When I remove this duplicated selector with _inspect element_ and leave only 1 occurrence, then everything works again

### Step-by-step reproduction instructions

It's hard to write step-by-step for this, as you will need to use some specific themes and plugins and setup some blocks, but it's happening on more websites with a different setup, but the point is, that this is a newly added CSS, so it was working fine until now.

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

No

annezazu:
>It's hard to write step-by-step for this, as you will need to use some specific themes and plugins and setup some blocks, but it's happening on more websites with a different setup, but the point is, that this is a newly added CSS, so it was working fine until now.

Hey there! Thanks for opening this issue and for using the latest in WordPress. Can you provide more information around how you can replicate this? For example, what plugins and themes are you using? Without replication steps, this makes it very difficult for us to address your feedback, either with more context or a fix.

kubiqsk:
Well, there are many plugins allowing you to add some margin for blocks, because the default Gutenberg one is not very useful as you never want to have the same spacings on desktop and on mobile.
Eg. Ghostkit, Block Editor Bootstrap Blocks or many others...
You could add margin top for the first block and margin bottom for the last block and it worked until version WP 6.3, but now you need to use `!important` switch/checkbox to make it works, so many websites have missing these margins now.

But let's say we will  not use any plugin, but ofc I need a bigger margin on the desktop and smaller on the mobile, so I can not use either Gutenberg controls, but I need to manually add class to the first block to have some top margin and also one class for the last block to have some bottom margin.
Here is the demo: https://demo.kubiq.sk/margin-problem/
You can inspect first or last paragraph - they should have 100px margin on desktop, 80px on tablet and 60px on mobile and this worked until WP 6.3 - now it is 0px and that means I need to visit every website and every page, post and CPT in it and fix this.

I'm maintaining hundreds of websites, but thanks god only a few tens of them are based on theme.json

[...]

PR #63972, opened 2024-07-26
CSS transformation: Fix mangled selectors by switching to a different postcss plugin for prefixing

### LINKED ISSUE #63829: Non-iframed Editor: Compound selectors within `:where` are not prefixed correctly
### Description

When the editor is not iframed, styles have their selectors prefixed with `.editor-styles-wrapper`. An issue arises when the style's selector being processed contains a compound selector within a `:where()` css function. 

The result is incorrect as the selectors within the `:where()`, other than the first, also get the `.editor-styles-wrapper` prefixed. This means part of the final selector doubles up on the `.editor-styles-wrapper` class causing those styles to not be applied as intended.

1. Activate TT4 (which has button styles) and navigate to the post editor
2. Ensure the editor is not iframed. This can be done by enabling the Custom Fields preference via the top right menu More > Preferences > Custom Fields, then reloading the editor.
3. Edit a post and add some buttons, then save
4. Inspect the buttons and find the matched style with the following selector:
`.editor-styles-wrapper :where(.wp-element-button, .editor-styles-wrapper .wp-block-button__link)`
5. Note the injection of the second `.editor-styles-wrapper` within that

Screenshot 2024-07-23 at 11 21 58 AM

### Environment info

- 6.6.0
- Gutenberg trunk

- [X] Yes

- [X] Yes

talldan:
Fixes  #63829

## What?
Tries switching from `postcss-prefixwrap` to `postcss-prefix-selector` for transforming styles

## Why?
`postcss-prefixwrap` is currently mangling compound `:where` selectors - [Non-iframed Editor: Compound selectors within :where are not prefixed correctly](https://github.com/WordPress/gutenberg/issues/63829#top)

I've reported it upstream in https://github.com/dbtedman/postcss-prefixwrap/issues/515 and can try making a PR on the repo if one hasn't by the time I get to it.

The issue still needs to be fixed in Gutenberg and core, so this PR should do that.

## How?
Switches from [postcss-prefixwrap](https://github.com/dbtedman/postcss-prefixwrap) to [postcss-prefix-selector](https://github.com/RadValentin/postcss-prefix-selector). Why does this make a difference?

I had a look at the codebases for the two. `postcss-prefixwrap` seems to take the whole selector string, and then call the string `split( ',' )` method to handle compound selectors:
https://github.com/dbtedman/postcss-prefixwrap/blob/c2636cdaa8381741df1b2b553d15a82d1e674802/src/internal/domain/CSSRuleWrapper.ts#L20-L46

This results in the selector being split incorrectly and prefixing within a `where` selector.

`postcss-prefix-selector` on the other hand seems to use the `selectors` property provided by the AST which I'd expect to be a lot more advanced in how it splits a selector (it should align pretty closely with the CSS spec):
https://github.com/RadValentin/postcss-prefix-selector/blob/86cb9b9dd50caef6e075ac88f9d716975c68a597/index.js#L38

There are a couple of features not offered by the new dependency:
- Automattic replacing of 'root' selectors rather than prefixing (html, :root, body)
- Avoiding duplicated selectors

It does have a `transform` callback that I've used to implement these features so that there's parity with `trunk`.

## Testing Instructions
1. Activate TT4 (which has button styles) and navigate to the post editor
2. Ensure the editor is not iframed. This can be done by enabling the Custom Fields preference via the top right menu More > Preferences > Custom Fields, then reloading the editor.
3. Edit a post and add some buttons, then save
4. Inspect the buttons and find the matched style with the following selector:
`.editor-styles-wrapper :where(.wp-element-button, .editor-styles-wrapper .wp-block-button__link)`
5. Note the injection of the second `.editor-styles-wrapper` within that


lib/class-wp-theme-json-gutenberg.php
```diff
@@ -1724,7 +1724,7 @@ protected function get_layout_styles( $block_metadata, $types = array() ) {
 										$spacing_rule['selector']
 									);
 								} else {
-									$format          = static::ROOT_BLOCK_SELECTOR === $selector ? '.%2$s %3$s' : '%1$s-%2$s %3$s';
+									$format          = static::ROOT_BLOCK_SELECTOR === $selector ? ':root :where(.%2$s %3$s)' : ':root :where(%1$s-%2$s %3$s)';
```
talldan: The problem before was that this would end up being a selector like `.is-layout-flow > *` (specificity 0,1,0) but when prefixed with `.editor-styles-wrapper` in the non-iframed editor it becomes `.editor-styles-wrapper .is-layout-flow > *` (specificity 0,2,0). It was reported that this has caused some issues - https://github.com/WordPress/gutenberg/issues/53468#issuecomment-2241279024.

My idea to fix that is to make it output `:root :where(.is-layout-flow > *)` (still specificity 0,1,0), and when we prefix those kind of rules we generally just replace the `:root` part, so it becomes `.editor-styles-wrapper where(.is-layout-flow > *)` (keeping the specificity at 0,1,0).

The trouble with these kind of generative rules is that I'll need to check every selector it produces to make sure it's valid and retaining the same specificity.

An alternative to this is that we could possibly read the specificity of the selector on the fly (using something like https://www.npmjs.com/package/specificity) and if it's high enough already we could prefix with `:where( .editor-styles-wrapper )` instead of `.editor-styles-wrapper`. My concerns are performance, and that it's introducing another third-party tool that needs to be completely bug free.

talldan: Maybe it's worth breaking this into a separate PR, since this shouldn't be dependent on the transformStyles changes.

talldan:
Here's a diff of the transformed css (comparing trunk and this PR):
<details><summary>Diff</summary>

Styles A
```diff
381c381
< .editor-styles-wrapper .has-body-font-family {
---
> .has-.editor-styles-wrapper-font-family {
```


Styles B
```diff
25c25
< .editor-styles-wrapper .has-global-padding :where(:not(.alignfull.is-layout-flow) > .has-global-padding:not(.wp-block-block, .editor-styles-wrapper .alignfull)) {
---
> .editor-styles-wrapper .has-global-padding :where(:not(.alignfull.is-layout-flow) > .has-global-padding:not(.wp-block-block, .alignfull)) {
30c30
< .editor-styles-wrapper .has-global-padding :where(:not(.alignfull.is-layout-flow) > .has-global-padding:not(.wp-block-block, .editor-styles-wrapper .alignfull))>.alignfull {
---
> .editor-styles-wrapper .has-global-padding :where(:not(.alignfull.is-layout-flow) > .has-global-padding:not(.wp-block-block, .alignfull))>.alignfull {
68c68
< .editor-styles-wrapper .is-layout-flow> :first-child {
---
> .editor-styles-wrapper :where(.is-layout-flow > :first-child) {
72c72
< .editor-styles-wrapper .is-layout-flow> :last-child {
---
> .editor-styles-wrapper :where(.is-layout-flow > :last-child) {
76c76
< .editor-styles-wrapper .is-layout-flow>* {
---
> .editor-styles-wrapper :where(.is-layout-flow > *) {
81c81
< .editor-styles-wrapper .is-layout-constrained> :first-child {
---
> .editor-styles-wrapper :where(.is-layout-constrained > :first-child) {
85c85
< .editor-styles-wrapper .is-layout-constrained> :last-child {
---
> .editor-styles-wrapper :where(.is-layout-constrained > :last-child) {
89c89
< .editor-styles-wrapper .is-layout-constrained>* {
---
> .editor-styles-wrapper :where(.is-layout-constrained > *) {
94c94
< .editor-styles-wrapper .is-layout-flex {
---
> .editor-styles-wrapper :where(.is-layout-flex) {
98c98
< .editor-styles-wrapper .is-layout-grid {
---
> .editor-styles-wrapper :where(.is-layout-grid) {
155c155
< .editor-styles-wrapper .is-layout-flex> :is(*, .editor-styles-wrapper div) {
---
> .editor-styles-wrapper .is-layout-flex> :is(*, div) {
163c163
< .editor-styles-wrapper .is-layout-grid> :is(*, .editor-styles-wrapper div) {
---
> .editor-styles-wrapper .is-layout-grid> :is(*, div) {
227c227
< .editor-styles-wrapper :where(.wp-element-button, .editor-styles-wrapper .wp-block-button__link) {
---
> .editor-styles-wrapper :where(.wp-element-button, .wp-block-button__link) {
245c245
< .editor-styles-wrapper :where(.wp-element-button:hover, .editor-styles-wrapper .wp-block-button__link:hover) {
---
> .editor-styles-wrapper :where(.wp-element-button:hover, .wp-block-button__link:hover) {
251c251
< .editor-styles-wrapper :where(.wp-element-button:focus, .editor-styles-wrapper .wp-block-button__link:focus) {
---
> .editor-styles-wrapper :where(.wp-element-button:focus, .wp-block-button__link:focus) {
259c259
< .editor-styles-wrapper :where(.wp-element-button:active, .editor-styles-wrapper .wp-block-button__link:active) {
---
> .editor-styles-wrapper :where(.wp-element-button:active, .wp-block-button__link:active) {
264c264
< .editor-styles-wrapper :where(.wp-element-caption, .editor-styles-wrapper .wp-block-audio figcaption, .editor-styles-wrapper .wp-block-embed figcaption, .editor-styles-wrapper .wp-block-gallery figcaption, .editor-styles-wrapper .wp-block-image figcaption, .editor-styles-wrapper .wp-block-table figcaption, .editor-styles-wrapper .wp-block-video figcaption) {
---
> .editor-styles-wrapper :where(.wp-element-caption, .wp-block-audio figcaption, .wp-block-embed figcaption, .wp-block-gallery figcaption, .wp-block-image figcaption, .wp-block-table figcaption, .wp-block-video figcaption) {
292c292
< .editor-styles-wrapper .wp-block-buttons-is-layout-flow> :first-child {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-flow > :first-child) {
296c296
< .editor-styles-wrapper .wp-block-buttons-is-layout-flow> :last-child {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-flow > :last-child) {
300c300
< .editor-styles-wrapper .wp-block-buttons-is-layout-flow>* {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-flow > *) {
305c305
< .editor-styles-wrapper .wp-block-buttons-is-layout-constrained> :first-child {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-constrained > :first-child) {
309c309
< .editor-styles-wrapper .wp-block-buttons-is-layout-constrained> :last-child {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-constrained > :last-child) {
313c313
< .editor-styles-wrapper .wp-block-buttons-is-layout-constrained>* {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-constrained > *) {
318c318
< .editor-styles-wrapper .wp-block-buttons-is-layout-flex {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-flex) {
322c322
< .editor-styles-wrapper .wp-block-buttons-is-layout-grid {
---
> .editor-styles-wrapper :where(.wp-block-buttons-is-layout-grid) {
326c326
< .editor-styles-wrapper :where(.wp-block-calendar table, .editor-styles-wrapper .wp-block-calendar th) {
---
> .editor-styles-wrapper :where(.wp-block-calendar table, .wp-block-calendar th) {
436c436
< .editor-styles-wrapper :where(.wp-block-post-comments-form textarea, .editor-styles-wrapper .wp-block-post-comments-form input) {
---
> .editor-styles-wrapper :where(.wp-block-post-comments-form textarea, .wp-block-post-comments-form input) {
522c522
< .editor-styles-wrapper :where(.wp-block-post-featured-image img, .editor-styles-wrapper .wp-block-post-featured-image .block-editor-media-placeholder, .editor-styles-wrapper .wp-block-post-featured-image .wp-block-post-featured-image__overlay) {
---
> .editor-styles-wrapper :where(.wp-block-post-featured-image img, .wp-block-post-featured-image .block-editor-media-placeholder, .wp-block-post-featured-image .wp-block-post-featured-image__overlay) {
586c586
< .editor-styles-wrapper :where(.wp-block-quote.has-text-align-right.is-style-plain, .editor-styles-wrapper .rtl .is-style-plain.wp-block-quote:not(.has-text-align-center):not(.has-text-align-left)) {
---
> .editor-styles-wrapper :where(.wp-block-quote.has-text-align-right.is-style-plain, .rtl .is-style-plain.wp-block-quote:not(.has-text-align-center):not(.has-text-align-left)) {
592c592
< .editor-styles-wrapper :where(.wp-block-quote.has-text-align-left.is-style-plain, .editor-styles-wrapper:not(.rtl) .is-style-plain.wp-block-quote:not(.has-text-align-center):not(.has-text-align-right)) {
---
> :root :where(.wp-block-quote.has-text-align-left.is-style-plain, .editor-styles-wrapper:not(.rtl) .is-style-plain.wp-block-quote:not(.has-text-align-center):not(.has-text-align-right)) {
604c604
< .editor-styles-wrapper :where(.wp-block-search .wp-block-search__label, .editor-styles-wrapper .wp-block-search .wp-block-search__input, .editor-styles-wrapper .wp-block-search .wp-block-search__button) {
---
> .editor-styles-wrapper :where(.wp-block-search .wp-block-search__label, .wp-block-search .wp-block-search__input, .wp-block-search .wp-block-search__button) {
614c614
< .editor-styles-wrapper :where(.wp-block-search .wp-element-button, .editor-styles-wrapper .wp-block-search .wp-block-button__link) {
---
> .editor-styles-wrapper :where(.wp-block-search .wp-element-button, .wp-block-search .wp-block-button__link) {
```

</details> 


There's a pretty obvious issue in the first diff where the replacement of `body` is way too basic:
```diff
381c381
< .editor-styles-wrapper .has-body-font-family {
---
> .has-.editor-styles-wrapper-font-family {
```

This also doesn't look right:
```diff
592c592
< .editor-styles-wrapper :where(.wp-block-quote.has-text-align-left.is-style-plain, .editor-styles-wrapper:not(.rtl) .is-style-plain.wp-block-quote:not(.has-text-align-center):not(.has-text-align-right)) {
---
> :root :where(.wp-block-quote.has-text-align-left.is-style-plain, .editor-styles-wrapper:not(.rtl) .is-style-plain.wp-block-quote:not(.has-text-align-center):not(.has-text-align-right)) {
```

lib/class-wp-theme-json-gutenberg.php
talldan: Splitting it out into a separate PR - #64076

packages/block-editor/src/utils/test/transform-styles.js
```diff
@@ -208,7 +208,24 @@ describe( 'transformStyles', () => {
 				'.my-namespace'
 			);
 
-			expect( output ).toMatchSnapshot();
+			expect( output ).toEqual( [ input ] );
```
talldan, line 226: Snapshot was unnecessary (and I think is also not needed for quite a lot of these tests), this just needs to test that the output is the same as the input.

talldan:
Closing in favour of #64458

PR #63980, opened 2024-07-26
Global Styles: Fix block custom CSS pseudo element selectors

### LINKED ISSUE #63932: Pseudo-element selectors in `styles.**.css` property no longer work in 6.6 compared to 6.5
### Description

With the [CSS specificity change in WordPress 6.6](https://make.wordpress.org/core/2024/06/21/wordpress-6-6-css-specificity/), most rules are now wrapped in `:root :where()`. However, pseudo-elements like `::before` and `::after` cannot be wrapped in `:where()`.

This means `theme.json` like this:

```json
{
  "$schema": "https://schemas.wp.org/wp/6.6/theme.json",
  "version": 3,
  "styles": {
    "elements": {
      "h1": {
        "css": "&::before { content: 'Foo' }"
      }
    }
  }
}
```
Produces:

```css
:root :where(h1::before) {
  content: 'Foo';
}
```
Which is invalid CSS and causes the CSS rule to no longer apply.

1. Install fresh WordPress.
2. Add an entry for `core/heading` in `styles.blocks` in `twentytwentyfour/theme.json`:
   ```diff
    },
   +"core/heading": {
   +  "css": "&::before { content: 'Foo' }"
   +},
    "core/image": {
   ```
3. See the invalid inline CSS gets added to the webpage:
   ```css
   :root :where(.wp-block-heading::before ){content: 'Foo'}
   ```
   Compare this to the equivalent in 6.5.5:
   ```css
   .wp-block-heading::before {content: 'Foo'}
   ```
   Which is valid CSS.

_No response_

### Environment info

- WordPress 6.6.1
- Twenty Twenty-Four 1.2

- [X] Yes

- [X] Yes

aaronrobertshaw:
Fixes: https://github.com/WordPress/gutenberg/issues/63932

## What?

Fixes two problems with custom css for blocks:
1. Empty CSS rules
2. Malformed selectors with pseudo elements within `:where()`

## Why?

- Empty rules are wasted data
- Malformed CSS selectors cause styles not to match and take effect

## How?

When processing custom CSS for blocks, strip out pseudo-elements from the original nested selector, perform the required wrapping in `:root :where`, then re-append the pseudo-element selector with its leading combinators if present.

## Testing Instructions

1. In the site editor navigate to Styles > Blocks > Heading > Advanced
2. In the custom CSS field add the following snippet: `&::before { content: 'Foo' }`
3. The word "Foo" should be visible before any heading block on the page
4. Save the Global Styles and confirm the display of the pseudo-element on the front end

## Screenshots or screencast 

| Before | After |
|---|---|
| Screenshot 2024-07-26 at 10 22 19 AM | Screenshot 2024-07-26 at 3 09 48 PM |

| Before | After |
|---|---|
| Screenshot 2024-07-26 at 3 07 50 PM | Screenshot 2024-07-26 at 3 09 01 PM |

### DIFF
--- a/backport-changelog/6.6/7097.md
+++ b/backport-changelog/6.6/7097.md
@@ -1,3 +1,3 @@
 https://github.com/WordPress/wordpress-develop/pull/7097
 
--   https://github.com/WordPress/gutenberg/pull/63980
+* https://github.com/WordPress/gutenberg/pull/63980
--- a/backport-changelog/6.6/7097.md
+++ b/backport-changelog/6.6/7097.md
@@ -1,0 +1,3 @@
+https://github.com/WordPress/wordpress-develop/pull/7097
+
+-   https://github.com/WordPress/gutenberg/pull/63980
--- a/lib/class-wp-theme-json-gutenberg.php
+++ b/lib/class-wp-theme-json-gutenberg.php
@@ -1458,11 +1458,22 @@ class WP_Theme_JSON_Gutenberg {
 				}
 				$nested_selector = $part[0];
 				$css_value       = $part[1];
-				$part_selector   = str_starts_with( $nested_selector, ' ' )
+
+				// Handle pseudo elements such as ::before, ::after etc. Regex will also
+				// capture any leading combinator such as >, +, or ~, as well as spaces.
+				// This allows pseudo elements as descendants e.g. `.parent ::before`.
+				$matches            = array();
+				$has_pseudo_element = preg_match( '/([>+~\s]*::[a-zA-Z-]+)/', $nested_selector, $matches );
+				$pseudo_part        = $has_pseudo_element ? $matches[1] : '';
+				$nested_selector    = $has_pseudo_element ? str_replace( $pseudo_part, '', $nested_selector ) : $nested_selector;
+
+				// Finalize selector and re-append pseudo element if required.
+				$part_selector  = str_starts_with( $nested_selector, ' ' )
 					? static::scope_selector( $selector, $nested_selector )
 					: static::append_to_selector( $selector, $nested_selector );
-				$final_selector  = ":root :where($part_selector)";
-				$processed_css  .= $final_selector . '{' . trim( $css_value ) . '}';
+				$final_selector = ":root :where($part_selector)$pseudo_part";
+
+				$processed_css .= $final_selector . '{' . trim( $css_value ) . '}';
 			}
 		}
 		return $processed_css;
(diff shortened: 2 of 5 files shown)


aaronrobertshaw:
Backport is available in https://github.com/WordPress/wordpress-develop/pull/7097.

andrewserong (approved):
I only had a chance for a quick skim through the code and smoke test, but this is working nicely for me and seems a good direction for a fix.

| Before | After |
| --- | --- |
| image | image |

It might be worth getting a second set of 👀 before landing just in case, but LGTM! 🚀

harlet:
I'm writing from mobile as this goes so fast and I don't know if I get to test before you merge.
Does this also take into account this case where we have '& + &{}'
Now it transforms into ':root :where(+){}'

aaronrobertshaw:
Good question @harlet 👍 

> Does this also take into account this case where we have '& + &{}'

This PR does not take into account the proposed selector. It only fixes two issues noted in the description.
- Empty CSS rule
- Invalid selectors with pseudo elements inside `:where`

> Now it transforms into ':root :where(+){}'

A compound selector with two `&` parts has not previously been supported. 

I confirmed this on a site using a vanilla WP 6.5.5 install and no plugins. Using a custom css snippet such as `& + &{content: "hi! "}` on the Heading block results in the following CSS `.wp-block-heading{}.wp-block-heading{+}.wp-block-heading{content: "hi! "}` in 6.5.5.

To my knowledge, the custom CSS feature in Global Styles wasn't meant as a complete 1:1 replacement for CSS stylesheets as it isn't really feasible here to re-implement the entire CSS spec.

ramonjd (approved):
I've specifically tested pseudo elements, with several variants `::before`, `::first-letter`, `::selection`

Also tested some nested rules (e.g., `$.test {}`) to ensure they still work

### CSS before

`:root :where(.wp-block-heading){}:root :where(.wp-block-heading::before ){content: 'Foo'}`
Screenshot 2024-07-30 at 6 47 46 PM



### After

`:root :where(.wp-block-heading)::before{content: 'Foo'}`

Screenshot 2024-07-30 at 6 48 06 PM

aaronrobertshaw:
Thanks for the reviews and testing 🙇 

I'll get this merged so it is one step closer for the next point release.

hellofromtonya:
PHP changes committed to WordPress Core's `trunk` via https://core.trac.wordpress.org/changeset/58896 ✅

aaronrobertshaw:
Thank you for committing this one @hellofromtonya 🙇

cbravobernal:
Removing the label as it was already backported and included in WP 6.6.2

PR #63982, opened 2024-07-26
Dev build: Use esbuild instead of Babel

scruffian:
## What?
This tries using esbuild instead of webpack for dev builds.

## Why?
As suggested in https://github.com/WordPress/gutenberg/pull/61692#issuecomment-2121581397 we should try adding a new build module for development. This will open the doors to new features like fast refresh.

## How?
Replaces the development build process with esbuild.

## Testing Instructions
1. Run npm ci
2. Run npm start
3. Check that the build continues to work as before


ellatrix:
I couldn't see `esbuild` mentioned in the discussion linked. I had never heard of it. Maybe some discussion is needed what the replacement should be. I think Vite is the most used one these days, and it's quite good.

luisherranz:
Vite uses Esbuild in development mode and Rollup in production mode. They are creating a new bundler/transiler called Rolldown to replace both.

More info about this: https://rolldown.rs/about#why-we-are-building-rolldown

scruffian:
Closing to focus on https://github.com/WordPress/gutenberg/pull/64444

PR #60084, opened 2024-03-21
DataViews: remove the `enumeration` type as redundant

oandregal:
Part of https://github.com/WordPress/gutenberg/issues/55083
Related https://github.com/WordPress/gutenberg/issues/55100

## What?

Removes the `enumeration` field type.

## Why?

The `enumeration` type is too coarse for what we want to achieve with field types. We haven't ran into issues yet because we haven't implemented other features (editing).

Two concrete examples were we could use a more fine-grained type system are:

- In [this PR](https://github.com/WordPress/gutenberg/pull/59897#discussion_r1533471737), the field type could be string or array. While the filter could be the same (enumeration), we should check for the field value differently.
- The current `type: enumeration` fields are actually different things. For example, [status is a string](https://github.com/WordPress/gutenberg/blob/remove/enumeration-type-dataviews/packages/edit-site/src/components/page-pages/index.js#L328) but author is a [number](https://github.com/WordPress/gutenberg/blob/remove/enumeration-type-dataviews/packages/edit-site/src/components/page-pages/index.js#L314).

I plan to introduce other types soon to support other kind of filters as [this one](https://github.com/WordPress/gutenberg/issues/55100#issuecomment-1991837749).

## How?

- b0abc16c58328e14eece170d44160f4b84b24423 Checking for field.elements instead of field.type has proven enough.

## Testing Instructions

Go to any of the dataviews-powered pages (Site Editor > Pages > Manage all pages, Site editor > Patterns, Site Editor > Patterns > Manage template parts, Site Editor > Templates) and verify filters still work as expected.

PR #63983, opened 2024-07-26
DataForm: add author to quick edit page/post list

oandregal:
Follow-up to https://github.com/WordPress/gutenberg/pull/63895

## What?

Makes the existing `author` field visible in the quick edit form.

https://github.com/user-attachments/assets/d333fa94-19d9-4d08-b62b-cd7aca74ff7f

## Why?

To continue expanding the quick edit form.

## How?

- Add validation to quick edit form https://github.com/WordPress/gutenberg/pull/63983/commits/271dac7818595af95f351a66df5f526355e83066
- Add author as a form field https://github.com/WordPress/gutenberg/pull/63983/commits/23d3904261bc14831c7409236f1da2cc1f6492a9
- Render the author as a select control https://github.com/WordPress/gutenberg/pull/63983/commits/16cda971420df1b7146a3c9da9c51bc61f07871b
- Consider bulk editing https://github.com/WordPress/gutenberg/pull/63983/commits/d88059f07e34e86223d2405d2faebf18e7d9d17a

## Testing Instructions

- With the "Allow access to a quick edit panel in the pages data views." experiment active, visit the Pages page in edit site.
- Change to table view and select a row. Then, open the quick edit form (right to view actions).
- Verify the correct author is selected and that you can change it. Bulk editing also works.

### DIFF
--- a/packages/dataviews/src/components/dataform/index.tsx
+++ b/packages/dataviews/src/components/dataform/index.tsx
@@ -9,6 +9,7 @@ import type { Dispatch, SetStateAction } from 'react';
 import {
 	TextControl,
 	__experimentalNumberControl as NumberControl,
+	SelectControl,
 } from '@wordpress/components';
 import { useCallback, useMemo } from '@wordpress/element';
 
@@ -66,7 +67,6 @@ function DataFormNumberControl< Item >( {
 }: DataFormControlProps< Item > ) {
 	const { id, label, description } = field;
 	const value = field.getValue( { item: data } );
-
 	const onChangeControl = useCallback(
 		( newValue: string | undefined ) =>
 			onChange( ( prevItem: Item ) => ( {
@@ -75,6 +75,17 @@ function DataFormNumberControl< Item >( {
 			} ) ),
 		[ id, onChange ]
 	);
+
+	if ( field.elements ) {
+		return (
+			<SelectControl
+				label={ label }
+				value={ value }
+				options={ field.elements }
+				onChange={ onChangeControl }
+			/>
+		);
+	}
 
 	return (
 		<NumberControl
--- a/packages/dataviews/src/components/dataform/index.tsx
+++ b/packages/dataviews/src/components/dataform/index.tsx
@@ -11,6 +11,7 @@ import {
 	__experimentalNumberControl as NumberControl,
 	SelectControl,
 } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
 import { useCallback, useMemo } from '@wordpress/element';
 
 /**
@@ -66,7 +67,7 @@ function DataFormNumberControl< Item >( {
 	onChange,
 }: DataFormControlProps< Item > ) {
 	const { id, label, description } = field;
-	const value = field.getValue( { item: data } );
+	const value = field.getValue( { item: data } ) ?? '';
 	const onChangeControl = useCallback(
 		( newValue: string | undefined ) =>
 			onChange( ( prevItem: Item ) => ( {
@@ -77,11 +78,23 @@ function DataFormNumberControl< Item >( {
 	);
 
 	if ( field.elements ) {
+		const elements = [
+			/*
+			 * Value can be undefined when:
+			 *
+			 * - the field is not required
+			 * - in bulk editing
+			 *
+			 */
+			{ label: __( 'Select item' ), value: '' },
+			...field.elements,
+		];
+
 		return (
 			<SelectControl
 				label={ label }
 				value={ value }
-				options={ field.elements }
+				options={ elements }
 				onChange={ onChangeControl }
 			/>
 		);
--- a/packages/edit-site/src/components/post-fields/index.js
+++ b/packages/edit-site/src/components/post-fields/index.js
@@ -235,7 +235,7 @@ function usePostFields( viewType ) {
 			{
 				label: __( 'Author' ),
 				id: 'author',
-				getValue: ( { item } ) => item._embedded?.author[ 0 ]?.name,
+				type: 'integer',
 				elements:
 					authors?.map( ( { id, name } ) => ( {
 						value: id,
--- a/packages/edit-site/src/components/post-fields/index.js
+++ b/packages/edit-site/src/components/post-fields/index.js
@@ -236,6 +236,7 @@ function usePostFields( viewType ) {
 				label: __( 'Author' ),
 				id: 'author',
 				type: 'integer',
+				getValue: ( { item } ) => item._embedded?.author[ 0 ]?.name,
 				elements:
 					authors?.map( ( { id, name } ) => ( {
 						value: id,
--- a/packages/edit-site/src/components/post-fields/index.js
+++ b/packages/edit-site/src/components/post-fields/index.js
@@ -236,7 +236,6 @@ function usePostFields( viewType ) {
 				label: __( 'Author' ),
 				id: 'author',
 				type: 'integer',
-				getValue: ( { item } ) => item._embedded?.author[ 0 ]?.name,
 				elements:
 					authors?.map( ( { id, name } ) => ( {
 						value: id,
--- a/packages/dataviews/src/components/dataform/stories/index.story.tsx
+++ b/packages/dataviews/src/components/dataform/stories/index.story.tsx
@@ -25,16 +25,26 @@ const fields = [
 		label: 'Order',
 		type: 'integer' as const,
 	},
+	{
+		id: 'author',
+		label: 'Author',
+		type: 'integer' as const,
+		elements: [
+			{ value: 1, label: 'Jane' },
+			{ value: 2, label: 'John' },
+		],
+	},
 ];
 
 export const Default = () => {
 	const [ post, setPost ] = useState( {
 		title: 'Hello, World!',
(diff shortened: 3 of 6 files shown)


packages/edit-site/src/components/post-fields/index.js
```diff
@@ -235,7 +235,7 @@ function usePostFields( viewType ) {
 			{
 				label: __( 'Author' ),
 				id: 'author',
-				getValue: ( { item } ) => item._embedded?.author[ 0 ]?.name,
```
oandregal, line 238: As far as I'm aware, we didn't really use this function. The implementation is also wrong, because the `getValue` should return an integer (what `item.author` is) instead of a string (the display name is a `render` concern).

packages/edit-site/src/components/post-fields/index.js
```diff
@@ -235,7 +235,7 @@ function usePostFields( viewType ) {
...
+				type: 'integer',
```
oandregal, line 238: If you remove the elements from the author definition, you can actually edit the author by providing its number. The `select` control is a user experience improvement and not part of the types.

packages/dataviews/src/components/dataform/index.tsx
```diff
@@ -76,6 +77,29 @@ function DataFormNumberControl< Item >( {
 		[ id, onChange ]
 	);
 
+	if ( field.elements ) {
```
youknowriad, line 80: I'm personally wondering whether this should be another type or not `enum`. I'm curious if @joshuatf has an opinion.

My feeling is that we should have a way to define these "controls" and pick a control by assigning a "type" or something else. I'm not sure that I like the fact that other field properties can be taken into consideration to pick the actual control.

youknowriad: To expand a bit. having a unique discriminator would be good for field "types" or "controls" registration.

packages/edit-site/src/components/post-fields/index.js
youknowriad: Don't we have a filter that uses this?

packages/edit-site/src/components/post-fields/index.js
youknowriad: As I said above, I feel like it's better to actually switch the "type" to "enum" for this case. rather than having to remove "elements"

youknowriad (commented):
This is cool 👍 

 - Definitely something worth adding to the storybook.
 - We may want to start to think separately about offering a new "layout" for `DataForm` to render the fields as a panel (like the document panel in the editor)

packages/dataviews/src/components/dataform/index.tsx
oandregal: The `type` is used for validation, so it needs to know the actual type (string, integer, etc.) of the underlying data. In follow-ups, it should also be used to sort the fields, for example. If we want the consumer to decide exactly which UI control to use we'd need a different prop field. This can be a follow-up PR.

youknowriad: > The type is used for validation, so it needs to know the actual type (string, integer, etc.) of the underlying data. 

For me this is another argument for `type: "enum"` as the validation for this particular case is to actually check whether the value is within the "elements" array rather than checking whether it's integer or string.

youknowriad: > To expand a bit. having a unique discriminator would be good for field "types" or "controls" registration.

Personally, I think this is the point that I care about the most. If you think we should keep type: "integer" here that works for me but I would suggest changing the implementation and make sure the `IntegerControl` switches between `Number` and `Select` depending on whether there's a `elements` or not (and same for `TextControl`). I know it means some duplication but for me it's still better.

packages/edit-site/src/components/post-fields/index.js
oandregal: It's not necessary for filters, using the integer would be enough. However! Your comment makes me realize this is actually necessary for _sorting_: we should sort authors by name not by the id assigned by WordPress.

I'm going to bring back this, but sorting is not working on `trunk`. We need to look at that separately. There's a few issues (author & dates not sorting properly, the sort config in the table headers is not cleared properly):

https://github.com/user-attachments/assets/fa98a60b-0b48-47e9-8350-0241d66e6892

oandregal: This actually raises an interesting question: if we have types but sorting depends on getValue, we can end up with the following (like in this case for `author`):

- type is `integer`
- but `getValue` returns a string

How do we make the [sort utility](https://github.com/WordPress/gutenberg/blob/add/author-to-quick-edit/packages/dataviews/src/filter-and-sort-data-view.ts#L136) work based on types in this scenario?

It seems we shouldn't use `getValue` for sorting and actually provide a custom sort utility if the field needs it. In that case, the filterSort utility would use the custom sort if it exists, otherwise would default to the sort defined by the field type. Thoughts?

_(nothing to do in this PR, as I brought back the `getValue` function https://github.com/WordPress/gutenberg/pull/63983/commits/216234ecd45d8f8e297f7efd44106578d46bf241 but wanted to share for consideration when looking at fixing sorting and formalizing types)_

packages/dataviews/src/components/dataform/index.tsx
oandregal: We actually had the `enumeration` type but removed it a few months ago after some discussions https://github.com/WordPress/gutenberg/pull/60084 I think those concerns are still valid. Introducing field types is stress-testing our APIs (see example about [sorting](https://github.com/WordPress/gutenberg/pull/63983#discussion_r1692936050)), and I'd like to formalize them a bit more using the underlying data type (same as REST API types: string, number, etc.).

> as the validation for this particular case is to actually check whether the value is within the "elements" array rather than checking whether it's integer or string

This is a good point! I's both to me: it should be integer AND it should be part of the provided integer list. I'll update to consider this as well.

packages/edit-site/src/components/post-fields/index.js
oandregal: > rather than having to remove "elements"

There's some confusion here: we don't need to remove elements. My point was that if you _were_ to remove elements, the quick edit would still work: the author field would render a number control and you could actually provide the author id and everything would work the same. My point was that the underlying type was `integer` and that select was a _user experience_ improvement.

packages/dataviews/src/components/dataform/index.tsx
oandregal: > Personally, I think this is the point that I care about the most. If you think we should keep type: "integer" here that works for me but I would suggest changing the implementation and make sure the IntegerControl switches between Number and Select depending on whether there's a elements or not (and same for TextControl). I know it means some duplication but for me it's still better.

That's how it is implemented now. With the caveat that I didn't implemented for the text type because it's not necessary so far. To keep the PRs minimal I thought it best to do it when it was need it, but I can also add it here if that's preferred. A PR should be up soon for this anyway: we need it for quick editing status.

oandregal:
Next steps here:

- ~validate that the value is also part of the elements defined by the field~ https://github.com/WordPress/gutenberg/pull/63983/commits/f3053ac32bfc727a463bc24b8424a5e397fda60a
- ~figure out how to move forward with `getValue` returning a name for sorting (author text) but validation & filters working based on the actual data type (author number).~ https://github.com/WordPress/gutenberg/pull/63983#discussion_r1694735017
- look at sorting issues in trunk for author & date (see [conversation](https://github.com/WordPress/gutenberg/pull/63983#discussion_r1692926128)).

packages/edit-site/src/components/post-fields/index.js
youknowriad: Yes, I know but my thinking was that in order to switch how something renders, I prefer if the change is to actually change "type" or "control" or something like that. It's fine though we can continue this discussion in the other thread.

packages/dataviews/src/components/dataform/index.tsx
youknowriad: > That's how it is implemented now. With the caveat that I didn't implemented for the text type because it's not necessary so far. 

I missed that sorry :P I thought the check was at the parent level. All good for me for now then.

youknowriad (approved):

packages/edit-site/src/components/post-fields/index.js
oandregal: I've looked a bit more into this and it turns out `getValue` is not used for sorting either: we delegate on [the REST API](https://github.com/WordPress/gutenberg/blob/add/author-to-quick-edit/packages/edit-site/src/components/post-list/index.js#L164) for filter/sort/rest in this page.

The issues I've seen with sorting in `trunk` are:

- author is sorted by creation id, not by name. In the existing wp-admin page, users cannot sort by author.
- date: I'm not sure why it doesn't work but it's unrelated. In the existing wp-admin page, this works fine.

I removed the `getValue` to proceed with this PR.

oandregal:
I know what happened with sorting for dates and have a fix at https://github.com/WordPress/gutenberg/pull/64049

oandregal:
https://github.com/WordPress/gutenberg/pull/64064 addresses sorting for author.

PR #63985, opened 2024-07-26
CustomSelectControl: support generic props type

### LINKED ISSUE #63984: CustomSelectControl: Support generic props type

Currently, the `CustomSelectControl` component is not generic. In the following example, the `option.foo` property could not be inferred correctly in the `onChange` callback:

image

By making the props generic, ts could infer the selectedItem type correctly:

image

I'll create a PR to help implement it

meteorlxy:
## What?

Fixes #63984

## Why?

It would be helpful to infer the option type

## How?

- Making the `CustomSelectProps` a generic type
- Exporting `CustomSelectOption` and `CustomSelectChangeObject` type

## Testing Instructions

Using the following snippet to see if the `selectedItem` type is inferred correctly:

```tsx
import { CustomSelectControl } from '@wordpress/components';
import type { CustomSelectOption } from '@wordpress/components/build-types/custom-select-control/types';

type MyOption = CustomSelectOption & {
  foo: string;
};

const options: MyOption[] = [
  {
    key: 'none',
    name: 'None',
    foo: 'bar',
  },
  {
    key: 'foo',
    name: 'Foo',
    foo: 'bar',
  },
];

const render = () => (
  <CustomSelectControl
    label="foo"
    onChange={({ selectedItem }) => {
      console.log(selectedItem.foo);
    }}
    options={options}
    value={options[0]}
  />
);
```

N/A

## Screenshots or screencast 

**Before**

image

**After**

image


packages/components/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 -   `CustomSelectControl`: Restore `describedBy` functionality ([#63957](https://github.com/WordPress/gutenberg/pull/63957)).
 
+### Enhancements
```
tyxla: Since this is a type change, I believe it should be under `Internal` instead.

packages/components/src/custom-select-control/index.tsx
```diff
@@ -51,7 +51,9 @@ function getDescribedBy( currentValue: string, describedBy?: string ) {
 	return sprintf( __( 'Currently selected: %s' ), currentValue );
 }
 
-function CustomSelectControl( props: CustomSelectProps ) {
+function CustomSelectControl< T extends CustomSelectOption >(
```
tyxla, line 54: Should we add a default to the generic so any existing consumers will continue working without changing anything? 

The same question applies to all instances where we're adding a generic and there wasn't one before.

packages/components/CHANGELOG.md
meteorlxy: Cuz the type will be used by user, and will improve the developer experience, I think it's more of an improvement / enhancement?

packages/components/src/custom-select-control/index.tsx
meteorlxy: The generic param is not going to be set manually by user. It's used for type inferring. To be specific, it will use the item type of the `options` prop automatically. So consumers would not need to change anything.

You could checkout this branch and test with the snippets that pasted in the "Testing Instructions" section.

jsnajdr: TypeScript should always be able to infer the `T` type. `CustomSelectProps` have a mandatory `options: Array<T>` field, so `T` is always the type of whatever you pass as `options`, it's easy to infer. TypeScript will merely check if that type is compatible with `CustomSelectOption`.

packages/components/CHANGELOG.md
tyxla: IMHO, types and type changes are an internal kind of enhancement, and we have this "Internal" group to distinguish them. General Enhancements would usually be targeted towards end users.

tyxla (commented):
This looks good and tests well in my editor, thanks for the contribution @meteorlxy 🙌 

The only thing I'd change would be to move the CHANGELOG entry to "Internal".

meteorlxy:
@tyxla updated

packages/components/src/custom-select-control/types.ts
```diff
@@ -24,15 +24,15 @@ type Option = {
 /**
  * The object returned from the onChange event.
  */
-type ChangeObject = {
+export type CustomSelectChangeObject< T extends CustomSelectOption > = {
```
tyxla: Do we need to export this one? Seems like we don't.

meteorlxy: It would be useful if user want to get the `ChangeObject` type, otherwise we have to use the following snippet to get the type:

image

tyxla: Right 👍 

Exporting it here doesn't export it from the package, though. And if we're not importing it inside the package, then it's effectively unnecessary to export it, IMHO. We can always start exporting it if we need to in the future.

mirka: Yes, at the moment we are avoiding public type exports as much as possible, thanks 🙏

meteorlxy: ok, removed

mirka (approved):
Looks good, thank you! 🚀

PR #58924, opened 2024-02-12
try animate frame

SaxonF:
## What?
Tries a few things around frame animation

PR #63986, opened 2024-07-26
Site Editor: Clarify that the site icon is a back button using an animation

youknowriad:
This is extracted and reworked based on #58924 

## What?

The idea here is to add a hover animation to the site icon in the editor to clarify its role as a back button. 

**Notes**

- Animation in and out of the editor only works in browsers that support View Transitions: Latest Safari, Chrome, Edge, Opera. (Not yet on Firefox)
- The animation in and out when not site icon is set is not perfect (I believe it's the same in trunk though)

https://github.com/user-attachments/assets/04cd6012-265e-4043-a336-6ca8be79bdc9

So while this is not perfect, it seems like a nice improvement over trunk.

### DIFF
--- a/packages/edit-site/src/components/editor/index.js
+++ b/packages/edit-site/src/components/editor/index.js
@@ -64,6 +64,18 @@ const toggleHomeIconVariants = {
 	},
 };
 
+const siteIconVariants = {
+	edit: {
+		clipPath: 'inset(0% round 0)',
+	},
+	hover: {
+		clipPath: 'inset( 22% round 2px )',
+	},
+	tap: {
+		clipPath: 'inset(0% round 0)',
+	},
+};
+
 export default function EditSiteEditor( { isPostsList = false } ) {
 	const disableMotion = useReducedMotion();
 	const { params } = useLocation();
@@ -79,6 +91,7 @@ export default function EditSiteEditor( { isPostsList = false } ) {
 		showIconLabels,
 		editorCanvasView,
 		currentPostIsTrashed,
+		hasSiteIcon,
 	} = useSelect( ( select ) => {
 		const {
 			getEditorCanvasContainerView,
@@ -89,8 +102,9 @@ export default function EditSiteEditor( { isPostsList = false } ) {
 			getEditedPostId,
 		} = unlock( select( editSiteStore ) );
 		const { get } = select( preferencesStore );
-		const { getCurrentTheme } = select( coreDataStore );
+		const { getCurrentTheme, getEntityRecord } = select( coreDataStore );
 		const _context = getEditedPostContext();
+		const siteData = getEntityRecord( 'root', '__unstableBase', undefined );
 
 		// The currently selected entity to display.
 		// Typically template or template part in the site editor.
@@ -107,6 +121,7 @@ export default function EditSiteEditor( { isPostsList = false } ) {
 			currentPostIsTrashed:
 				select( editorStore ).getCurrentPostAttribute( 'status' ) ===
 				'trash',
+			hasSiteIcon: !! siteData?.site_icon_url,
 		};
 	}, [] );
 	useEditorTitle();
@@ -193,43 +208,8 @@ export default function EditSiteEditor( { isPostsList = false } ) {
 		getEditorCanvasContainerTitleAndIcon( editorCanvasView );
 
 	const isReady = ! isLoading;
-
-	const toggleVariants = {
-		edit: {
-			width: 60,
-			height: 60,
-			top: 0,
-			left: 0,
-			borderRadius: '0px',
-			boxShadow: 'none',
-			transition: {
-				delay: 0.1,
-				duration: disableMotion ? 0 : 0.2,
-				type: 'tween',
-				stiffness: 400,
-			},
-		},
-	};
-
-	const siteIconVariants = {
-		edit: {
-			clipPath: 'inset(0% round 0)',
-			transition: {
-				duration: disableMotion ? 0 : 0.2,
-			},
-		},
-		hover: {
-			clipPath: 'inset( 22% round 2px )',
-			transition: {
-				duration: disableMotion ? 0 : 0.2,
-			},
-		},
-		tap: {
-			clipPath: 'inset(0% round 0)',
-			transition: {
-				delay: disableMotion ? 0 : 0.1,
-			},
-		},
+	const transition = {
+		duration: disableMotion ? 0 : 0.2,
 	};
 
 	return (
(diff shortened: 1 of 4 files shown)


richtabor:
Two ideas: 

1. What about if the back button had a fill to it? 

CleanShot 2024-07-26 at 08 59 07

CleanShot 2024-07-26 at 09 00 42

2. And what if we did not scale the site icon at all, but just showed the back button on hover? Perhaps something like this: 

https://github.com/user-attachments/assets/2d69bf2f-ce43-41d7-97ac-fe80231de41e

richtabor:
It's a little finicky, especially if the site icon is not 1:1 ratio. 

https://github.com/user-attachments/assets/14fd2bb9-7da1-470a-9ccc-91a0180ade68

jameskoster:
I'm still not 100% sure that the `<` is the right icon. It suggests going _back_, but that's not always the behavior. For instance if you follow the edit link from the frontend, or an edit action in the command palette, then this button doesn't take you _back_.

The `<` doesn't really align with the label either. If "Open navigation" is the label, then wouldn't it make sense to use the standard `menu` (hamburger) icon?

An alternative to try could be a `↘` icon with a label like "Close editor" or "Go to $destination".

richtabor:
> I'm still not 100% sure that the < is the right icon. It suggests going back, but that's not always the behavior. 

I had the same thought. I tried a WordPress icon, but that didn't quite feel right. Other maybes I tried quickly were some sort of "open to that view" icon, like this: 

CleanShot 2024-07-26 at 09 59 15

jameskoster:
It could work, but to me the action is a little more profound than opening a panel, like that icon suggests.

I'm inclined to think that the label will inform the icon. Is "Open navigation" the right label?

mtias:
Maybe we need to draw a "home" icon:

image

But let's get something in so that it doesn't linger in limbo for too long

youknowriad:
Can we get some consensus for a v1 here, happy to go with any direction.

jameskoster:
If we're leaving the label as it is ("Open navigation") then I'd lean towards using the `menu` icon. That way there's alignment between the text and the visual.

But even the `<` is an improvement on trunk imo, so I wouldn't object to it for a 'v1' until we figure out the correct approach.

It would be good to fix the issues around non-square icons before merging though:

Screenshot 2024-07-26 at 14 32 27

jameskoster:
There's a small issue with the W animation. It's quite hard to see. Everything is smooth except the W color which seems to change from white to gray instantly. 

w

richtabor:
> Maybe we need to draw a "home" icon

I like this idea too.

youknowriad:
> It would be good to fix the issues around non-square icons before merging though:

How do set no-square icons as site icons. The site icon upload button forces me to crop the images to be square?

youknowriad:
I simplified the animation for the "w" icon. @jameskoster The issue of the instance change was caused by the `backdrop-filter` css rule that I had to change.

I think this is shippable but I would personally argue that we should aim to simplify the animation further. Avoid clip path animation for instance, maybe we can avoid animating the icon itself and just make sure a back icon appears on top of it or something like that.

jameskoster:
> How do set no-square icons as site icons. The site icon upload button forces me to crop the images to be square?

If you use a non-square site logo and check "Use as site icon" in the Site Logo block options. 

Agree the animation can be tweaked separately.

youknowriad:
> If you use a non-square site logo and check "Use as site icon" in the Site Logo block options.

I wonder if this should be possible or if we should force the ratio somehow there as it's forced when you pick and site icon from the settings page.

jameskoster:
> I wonder if this should be possible

Probably not.

youknowriad:
@jameskoster Fixed the aspect ratio of non square site icons.

jameskoster (approved):
That fixed it 👍 

There are several small details to potentially refine like the icon, tooltip/label, and animation. But to avoid holding this up they can be explored in follow-ups.

mtias:
Let's follow up quickly on trying a different icon:

image

jasmussen:
Followed up here: https://github.com/WordPress/gutenberg/pull/64161

packages/edit-site/src/components/editor/style.scss
```diff
@@ -17,3 +17,60 @@
 	display: flex;
 	justify-content: center;
 }
+
+.edit-site-editor__view-mode-toggle {
+	/* stylelint-disable -- Disable reason: View Transitions not supported properly by stylelint. */
+	view-transition-name: toggle;
+	/* stylelint-enable */
+	position: fixed;
```
dereksmart, line 25: Something about this commit caused the icon to sit low in distraction free. Commenting here because removing the fixed position resolves it in my testing

https://github.com/user-attachments/assets/8f96b692-eb01-4543-99ea-9eebacc2f1b2

mtias: Good catch

youknowriad: Fix in #64261

afercia:
Looking into this after some vacation, I'm not sure this change helps in any way to clarify how to go back to navigation and exit the editor. A few observations:

- This animation works only on hover, as it is set on the container of the button. It doesn't work for keyboard users or users who use other non-pointing devices. As such there's no equivalent experience for users and only _some users_ will see the animation.
- The icon itself is supposed to represent the concept of 'home' but activating the button doesn't bring to any 'home', it just opens the menu. See my comment on https://github.com/WordPress/gutenberg/pull/64161#issuecomment-2320255399

It's worth reminding there's already an issue to discuss the overall usage of the WP logo / Site Icon in the editor. The inconsistent behavior between post editor and site editor doesn't help. It's not clear how to exit the site editor. I'd think more radical design changes are necessary to clarify the UI and improve the user experience and I'd encourage to explore alternative solutions in the context of https://github.com/WordPress/gutenberg/issues/57813

packages/edit-site/src/components/editor/index.js
```diff
@@ -75,8 +102,9 @@ export default function EditSiteEditor( { isPostsList = false } ) {
 			getEditedPostId,
 		} = unlock( select( editSiteStore ) );
 		const { get } = select( preferencesStore );
-		const { getCurrentTheme } = select( coreDataStore );
+		const { getCurrentTheme, getEntityRecord } = select( coreDataStore );
 		const _context = getEditedPostContext();
+		const siteData = getEntityRecord( 'root', '__unstableBase', undefined );
```
ellatrix, line 107: Did you add `undefined` here intentionally or by accident?

youknowriad: Hard to tell at this point 😬 I don't remember

Issue #63308, opened 2024-07-09
Accessibility improvements for the new Post summary panel


afercia:
### Description

In the new Post summary panel, some settings previously placed within collapsible panels have been moved to Popovers that open from some new settings panel buttons. Namely, the Excerpt, Author, and Discussion settings.

This made more evident some long standing accessibility issues in this panel. 

The most important and still unsolved issue is that all the buttons to edit these settings:
- Status
- Publish 
- Link
- Author
- Template 
- Discussion 
- Revisions

are labelled in an incorrect way. Buttons are meant to perform actions. Buttons should be labeled to communicate the action they perform. Instead, these buttons are labeled with the corresponding value. This was first reported long time ago in https://github.com/WordPress/gutenberg/issues/470 and it's not my intent to rehash that discussion now as it appears there's some deep disagreement on the best practices to label controls. Just a reminder that these buttons are all labeled in a far from ideal way. Also, their accessible name mismatch their visual label. As such, they will be flagged as a WCAG violation in any accessibility audit.

Instead, this issue aims to address some _minor_ accessibility and inconsistencies in the new Summary panel.

While the buttons visible text is the actual set value, their accessible name attempts to communicate the button action. The actual aria-labels of these buttons are mostly in thie format:

- Change post status: {current value}
- Change date: {current value}
- Change link: {current value}
- Change author: {current value}
- Change discussion options: {current value}

While this labeling helps blind screen reader users, it doesn't help sighted screen readere users, speech recognition software users, users with cognitive impairments, etc. The _visible_ buttons text is disorienting and doesn't communicate the associated action. At the very leas, to attempt some _remediation_, the actual accessible name should be visually exposed by the means of a tooltip.

1.
Status setting:

![01](https://github.com/WordPress/gutenberg/assets/1682452/b2fdf59c-871a-4d1b-b721-838c7c745936)

- The 'visual label' is `Status`.
- The actual label is `Change post status: {current value}`
- The term 'post' is hardcoded so it will be used also for pages and custom post types, which isn't ideal. The simplest solution would be to just remove 'post' and make this string a more generic `Change status: {current value}`.
- Clicking the button opens a popover where status, visibility, password protected, and sticky settings are now grouped together.
- However, neither the visual label or the actual label mention the other settings. They only refer to 'Status'.
- I'm not sure these settings should be grouped together in the first place. Anyways, the labeling should better communicate what the popover content is.

2
Revisions.
[...]

PR #63988, opened 2024-07-26
Fix the post summary Status toggle button accessibility

afercia:
Addresses a few points reported in https://github.com/WordPress/gutenberg/issues/63308

## What?

The button to open the 'Status & visibility' popover has an aria-label attribute `Change post status: %s` where the term `post` is hardcoded. This should be removed as this setting is used for pages and CPTs as well.
The button missies an aria-expanded attribute.

## Why?

Minor improvement for accessibility and consistency.

## How?

- Changes the aria-label of the Status button by removing the hardcoded `post`.
- Adds an `aria-expanded` attribute to the Status toggle button.
- Adjusts the e2e tests accordingly.

## Testing Instructions

- Edit a post.
- Inspect the source and observe the Status button as an aria-label with text `Change status: %`.
- Observe the button has an aria-expanded attribute with a value that changes depending on whether the popover is open.

### DIFF
--- a/packages/editor/src/components/post-status/index.js
+++ b/packages/editor/src/components/post-status/index.js
@@ -179,7 +179,7 @@ export default function PostStatus() {
 							icon={ postStatusesInfo[ status ]?.icon }
 							aria-label={ sprintf(
 								// translators: %s: Current post status.
-								__( 'Change post status: %s' ),
+								__( 'Change status: %s' ),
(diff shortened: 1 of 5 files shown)


Mamaduka (approved):
Thank you, @afercia!

The changes look good to me.

afercia:
Thank you @Mamaduka for your review 👋🏽

PR #63994, opened 2024-07-26
Makes spacing consistent in zoom out vertical toolbar

### LINKED ISSUE #63807: Inconsistent spacing in zoom out vertical toolbar
The spacing between controls on the zoom out vertical toolbar is inconsistent, with some controls having more spacing between each other visually than others. I'd like this to feel like one cohesive unit of controls. 

### Current 

CleanShot 2024-07-22 at 09 00 50

### Expected: 

CleanShot 2024-07-22 at 09 00 36

Setting the `components-accessible-toolbar` height to 40px, from 48px, seems to provide ample room while also keeping controls spaced out consistently. 

CleanShot 2024-07-22 at 08 59 44

amitraj2203:
## What?
Fixes: #63807 

## Why?
See this [comment](https://github.com/WordPress/gutenberg/issues/63807#issue-2422827049)

## Testing Instructions
1. Enable the zoomed out view
2. Go to any post and try to add a pattern by clicking on the inserter icon.
3. See the vertical toolbar controls and check the spacing between them.

## Screenshots or screencast 
Screenshot 2024-07-26 at 6 41 57 PM

Co-authored-by: amitraj2203 <amitraj2203@git.wordpress.org>
Co-authored-by: getdave <get_dave@git.wordpress.org>
Co-authored-by: draganescu <andraganescu@git.wordpress.org>
Co-authored-by: richtabor <richtabor@git.wordpress.org>

### DIFF
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -136,7 +136,6 @@ $editor-html-font: $font-family-mono;
  * Block & Editor UI.
  */
 
-$block-vertical-toolbar-height: $grid-unit-50;
 $block-toolbar-height: $grid-unit-60;
 $border-width: 1px;
 $border-width-focus-fallback: 2px; // This exists as a fallback, and is ideally overridden by var(--wp-admin-border-width-focus) unless in some SASS math cases.
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -136,6 +136,7 @@ $editor-html-font: $font-family-mono;
  * Block & Editor UI.
  */
 
+$block-vertical-toolbar-height: $grid-unit-50;
 $block-toolbar-height: $grid-unit-60;
 $border-width: 1px;
 $border-width-focus-fallback: 2px; // This exists as a fallback, and is ideally overridden by var(--wp-admin-border-width-focus) unless in some SASS math cases.
--- a/packages/block-editor/src/components/block-tools/style.scss
+++ b/packages/block-editor/src/components/block-tools/style.scss
@@ -201,4 +201,9 @@
 		background: none;
 		border: none;
 	}
+
+	// Make the spacing consistent between controls.
+	.components-button {
+		height: 40px;
+	}
 }
--- a/packages/block-editor/src/components/block-tools/style.scss
+++ b/packages/block-editor/src/components/block-tools/style.scss
@@ -204,6 +204,6 @@
 
 	// Make the spacing consistent between controls.
 	.components-button {
-		height: 40px;
+		height: $button-size-next-default-40px;
 	}
 }
--- a/packages/components/src/toolbar/toolbar/style.scss
+++ b/packages/components/src/toolbar/toolbar/style.scss
@@ -18,16 +18,6 @@
 	}
 }
 
-// Make the spacing consistent between controls in the vertical toolbar, only in zoom out.
-.zoom-out-toolbar-popover {
-	.components-accessible-toolbar,
-	.components-toolbar {
-		.components-button {
-			height: 40px;
-		}
-	}
-}
-
 .components-accessible-toolbar,
 .components-toolbar {
 
--- a/packages/components/src/toolbar/toolbar/style.scss
+++ b/packages/components/src/toolbar/toolbar/style.scss
@@ -18,6 +18,16 @@
 	}
 }
 
+// Make the spacing consistent between controls in the vertical toolbar, only in zoom out.
+.zoom-out-toolbar-popover {
+	.components-accessible-toolbar,
+	.components-toolbar {
+		.components-button {
+			height: $block-vertical-toolbar-height;
+		}
+	}
+}
+
 .components-accessible-toolbar,
 .components-toolbar {
 
--- a/packages/components/src/toolbar/toolbar/style.scss
+++ b/packages/components/src/toolbar/toolbar/style.scss
@@ -23,7 +23,7 @@
 	.components-accessible-toolbar,
 	.components-toolbar {
 		.components-button {
-			height: $block-vertical-toolbar-height;
+			height: 40px;
 		}
 	}
 }


packages/base-styles/_variables.scss
```diff
@@ -86,6 +86,7 @@ $widget-area-width: 700px;
  * Block & Editor UI.
  */
 
+$block-vertical-toolbar-height: $grid-unit-50;
```
richtabor: This seems quite a bit too prescriptive here. I don't think we'd use this anywhere else.

packages/components/src/toolbar/toolbar/style.scss
```diff
@@ -18,6 +18,16 @@
 	}
 }
 
+// Make the spacing consistent between controls in the vertical toolbar, only in zoom out.
+.zoom-out-toolbar-popover {
+	.components-accessible-toolbar,
+	.components-toolbar {
+		.components-button {
+			height: $block-vertical-toolbar-height;
```
richtabor: Is there a clearer way to do this?

amitraj2203: We can use `$grid-unit-50` from [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_variables.scss#L37) variable instead of creating a new one for `40px`. Should I go ahead with that?

amitraj2203:
Hi @richtabor, Is there something that needs to be done from my side?
Thanks!

packages/base-styles/_variables.scss
getdave: We tend not to add variables unless we absolutely have to.

As Zoom Out is still behind an experiment I don't think we should add this.

I understand the motivation but it's a premature abstraction in my opinion.

packages/components/src/toolbar/toolbar/style.scss
getdave: Yes please reuse an existing variable or hard code for the timebeing.

getdave (commented):
Thanks for the PR 🎉 

If we can make some small changes to remove abstractions I think we'll be in a good place code-wise.

packages/components/src/toolbar/toolbar/style.scss
amitraj2203: Removed the variable and used hard code value.

getdave (dismissed):
For me this looks more visual consistent.

Screen Shot 2024-09-26 at 11 21 57

Let's merge as an improvement and folks can tweak as required.

getdave (commented):
I noticed the changes were previously in the `@wordpress/components` package which shouldn't have any awareness of "zoom out".

I therefore moved the CSS to the `@wordpress/block-editor` package which is where the state for "zoom out" resides.

However, this change hasn't worked implementation wise so we'll need to look at this again if possible 🙏 

The aim is to get consistent spacing between each of the toolbar icons.

packages/block-editor/src/components/block-tools/style.scss
```diff
@@ -283,6 +283,11 @@
 		background: none;
 		border: none;
 	}
+
+	// Make the spacing consistent between controls.
+	.components-button {
+		height: 40px;
```
richtabor: Any reason this shouldn't use $button-size-next-default-40px?

richtabor:
Nice. Makes them feel more like one unit of control for the pattern. 

### Before 

CleanShot 2024-09-30 at 10 07 10

### After

CleanShot 2024-09-30 at 10 06 28

packages/block-editor/src/components/block-tools/style.scss
getdave:
```suggestion
		height: $button-size-next-default-40px;
```

@richtabor Not as far as I know. However, when I checked out this PR the implementation doesn't quite work. 

I would look to fix this, but I actually find it _really_ hard to notice the differences/inconsistencies in spacing. My brain plays tricks.

Moreover, sometimes even when the spacing between items is "correct" it still looks visually "off" because of the relative space occupied by the icons in use. I wonder whether we'll need a default spacing and then custom / specific spacing on a case by case basis?

draganescu (approved):
I think this is OK, for me using the scss variable is the same as using the number itself. I don't think the styling should be elsewhere as it styles stuff in the block editor package. What am I missing @getdave ?

getdave:
@draganescu This is now all good. I moved the CSS into the correct package in previous commits and @richtabor has confirmed he's happy with the visual appearance (which I was struggling to verify myself).

Therefore I'd be happy to merge this.

getdave:
Needs a rebase to allow tests to pass

ciampo:
👋  Hello!

I would ask you to revisit the approach taken here:

- Please do not use `.components-*` class names outside the `@wordpress/components` package. These are intended as private APIs;
- Also, please do not override styles that are meant to be controlled by components itself — `Button` comes with a set of available heights. It those heights don't suit your needs, it's a sign that the design that you're implementing is not following best practices. Or, otherwise, it's a sign that the design guidelines should be updated to keep this new scenario into account
- In this specific case, you should be able to achieve the same effect by setting the `__next40pxDefaultSize` prop on the affected `Button` components — no need to apply any external CSS at all. Soon we will switch all instances of `Button` to have a height of 40px by default.

Those are best practices that can go a long way in raising the quality of our code and our UIs; they are something that we hope everyone working on Gutenberg can learn and apply both when authoring and when reviewing code. 

cc @WordPress/gutenberg-components @WordPress/gutenberg-design

getdave:
Ok thanks @ciampo. I'll take ownership of sorting this out as I should definitely have known better.

Can I ask is any of the above in public documentation that you would be able to point me towards?

ciampo:
> I'll take ownership of sorting this out

Thank you, @getdave , I appreciate!

> Can I ask is any of the above in public documentation that you would be able to point me towards?

[This paragraph](https://github.com/WordPress/gutenberg/blob/d186350f49d4eaf34d4060bcf7a4923344aa123c/docs/contributors/code/coding-guidelines.md?plain=1#L56) specifies that "A component's class name should **never** be used outside its own folder" and highlights how this is a best practice to ensure encapsulation and reusability.

I don't believe the rest is really codified in public docs, I'd say it's more common sense.

But there should definitely be an effort from our side to write better docs, best practices, maybe even lint rules.

getdave:
I'm thinking more along the lines of https://make.wordpress.org/core/2024/09/12/gutenberg-development-practices-and-common-pitfalls/. 

One person's "common sense" is often not that clear to others and thus I"m keen we start to surface these "best practices" or "common sense guidelines". 

I'm not attempting to absolve myself here, but rather noticing that newer contributors may not share the context that you and I (should) have.

ciampo:
You're absolutely right, I'll start brainstorming a similar post and share it with y'all at a later point

getdave:
Follow up here https://github.com/WordPress/gutenberg/pull/65858

> I'll start brainstorming a similar post and share it with y'all at a later point

Feel free to share a draft and I'll happily proof read.

PR #62906, opened 2024-06-27
Gallery: Add lightbox support

madhusudhand:
Originally written by @madhusudhand 

<details><summary>Details</summary>

## What?

This introduces lightbox functionality to `Gallery` block.

## How?

* Image lightbox has been extended to support gallery
* Next and Prev icons were added to image lightbox if the image is in the context of a gallery.
* In this iteration (first) next/prev navigation for the lightbox in gallery are based on the "Expand on click" property of the image. 

Following is the behaviour in various scenarios.

### Case 1: "Expand on click" is enabled for the image block globally.

All images inside (unless overridden at block level), will invoke the lightbox on click and next and previous navigation is possible.

### Case 2: One of the image (lets say 2nd image) is set to "Link to image file"

Lightbox will be invoked on click of 1st image and on click of next will move to 3rd image.

### Case 3: "Expand on click" is **not** enabled globally, but few images are set with this option in gallery.

This is similar to **case 2**, lightbox opens with only those images.

## What's in future PRs?

* This change updates the image source on next and prev navigation. The smooth transition between the images will be addressed in a future PR.
* This has no effect on touch devices. It will be addressed in followup.

## Screenshots (Draft)

gallery-lightbox-1

image

## Testing instructions

* In global styles, enable "Click to expand" feature on image block.
* Add a gallery block to a post or page. Add images to it.
* In the frontend, click on an image to invoke lightbox and navigate between images by clicking on icons in the UI, or using next and prev arrow keys in the keyboard.

# Related issues

First iteration of #56310

</details> 

---

Latest description written by @t-hamano

## What?

This PR enables navigation between images with lightbox enabled.

## Why?

#56310

## How?

- If the Gallery block contains images with "Enlarge on click" enabled, you will be able to navigate between those images. If "Enlarge on click" is not enabled, those images will be skipped.
- This feature is enabled **by default**, ~but can be disabled in the block sidebar. If the feature is disabled, the image lightboxes themselves will work as before, but navigation between images will not be enabled:~ Update: Changed to always be enabled.
- Swipe actions on touch devices are not currently possible. This may be addressed in a follow-up.
- To localize the `aria-label`, inject an element with `aria-live` and `aria-atomic` attributes and vary the text inside. We could also use `speak`, but [the `@wordpress/a11y` script module is available from WP 6.7](https://make.wordpress.org/core/2024/10/14/updates-to-script-modules-in-6-7/). The Gutenberg plugin has to support WP 6.6 for now, so I didn't go with the speak approach.

[...]

PR #63995, opened 2024-07-26
Image Block Lightbox: Fix warning error when resizing

t-hamano:
Discovered while testing #62906

## What?

This PR fixes a browser warning error that occurs when changing the browser width when the lightbox is not yet open:

```
TypeError: Cannot read properties of undefined (reading 'imageRef') at setOverlayStyles (view.js:191:31)
```

## Why?

From what I understand, when the lightbox is not open, `state.currentImage` is undefined.

## How?

Use optional chaining.

Alternatively, since overlay style calculations should not be necessary when the lightbox is not open and should be recalculated when the lightbox is opened, the following approach also works:

```javascript
setOverlayStyles() {
	if ( ! state.overlayEnabled ) {
		return;
	}
}
```

## Testing Instructions

- Apply a lightbox to an image block.
- On the frontend, change the browser width without opening the lightbox.

### DIFF
--- a/packages/block-library/src/image/view.js
+++ b/packages/block-library/src/image/view.js
@@ -188,7 +188,7 @@ const { state, actions, callbacks } = store(
 		},
 		callbacks: {
 			setOverlayStyles() {
-				if ( ! state.currentImage.imageRef ) {
+				if ( ! state?.currentImage?.imageRef ) {
 					return;
 				}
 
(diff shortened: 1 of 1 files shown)


luisherranz:
> Alternatively, since overlay style calculations should not be necessary when the lightbox is not open and should be recalculated when the lightbox is opened, the following approach also works

That one is better, as `setOverlayStyles` is directly called by this directive: `data-wp-on-async-window--resize="callbacks.setOverlayStyles"` 🙂

t-hamano:
@luisherranz Thanks for the review! Updated 👍

luisherranz (approved):
Thanks!

PR #64006, opened 2024-07-26
Script Modules: Move data passing to 6.7 compat file

sirreal:
## What?

- Move script modules data to ~6.6~ 6.7 compat file. This feature is on trunk and slated for WordPress 6.7.
- Add conditions to avoid printing data twice.

## Why?

This feature landed in core and is planned for release in 6.7. It is not experimental, but part of the WordPress <6.7 compatibility layer.

The filters for printing script data should be registered conditionally to prevent duplicating Core behavior and printing that data multiple times.

## Testing

Try using this Gutenberg plugin locally with script data. The current WordPress trunk implementation of Interactivity API uses script module data (changed here) to enqueue Interactivity data. Any stores with server side data should use this mechanism to pass data to the client, so it can be observed that way, like in this comment: https://github.com/WordPress/gutenberg/pull/64006#issuecomment-2253105876.


sirreal:
[You can see an example with latest trunk where this results in script data being printed twice](https://playground.wordpress.net/?wp=nightly&plugin=gutenberg&plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger):

```js
document.querySelectorAll('[type="application/json"]');
// NodeList(2) [script#wp-script-module-data-@wordpress/interactivity, script#wp-script-module-data-@wordpress/interactivity]
```

(Notice two scripts with the same ID are printed).

lib/compat/wordpress-6.6/script-modules.php
```diff
@@ -0,0 +1,103 @@
...
+					'type' => 'application/json',
+					'id'   => "wp-script-module-data-{$module_id}",
+				)
+			);
+		}
+	}
+}
+
+add_action(
+	'after_setup_theme',
```
sirreal: Core uses `after_setup_theme` so we need to wait for that before the `has_action` checks will work correctly.

https://github.com/WordPress/wordpress-develop/blob/ed59f778ba7e275c7f34f3f8dd5795abcb07ca91/src/wp-settings.php#L404

sirreal:
This is from https://core.trac.wordpress.org/ticket/61512, in WordPress 6.7. I think that means it should be part of 6.7 compat, so I'll need to adjust that.

sirreal:
I've rebased and moved this to 6.7 compat.

lib/compat/wordpress-6.6/script-modules.php
SantosGuillamot: Should the `priority` parameter be used to ensure it runs later?

SantosGuillamot: Another thing I was wondering is if it would be easier to check that the new method introduced in the core ticket exists. Something like this:

```php
if ( ! method_exists( 'WP_Interactivity_API', 'filter_script_module_interactivity_data' ) ) {
	add_action( 'wp_footer', 'gutenberg_print_script_module_data' );
	add_action( 'admin_print_footer_scripts', 'gutenberg_print_script_module_data' );
}
```

I have neither the context nor a strong opinion, just sharing in case it is considered.

SantosGuillamot:
I've been testing it, and I am not sure I understand what the compatibility code is supposed to do 😄 

I've been testing this in the `6.6` branch and an image block with lightbox enabled, which should use `wp_interactivity_state`, and the HTML generated is the same with/without the compat code. Both return the following script:

```html
<script type="application/json" id="wp-interactivity-data">...</script>
```

I can see in `trunk` that removing [these filters](https://github.com/WordPress/wordpress-develop/blob/a799101e46ef0a042814c189a6331c23a52a100c/src/wp-includes/class-wp-script-modules.php#L186-L187) stop loading the scripts. I was expecting the same after removing the compat code from Gutenberg, but it doesn't seem to be the case.

Additionally, I can see the name is different. Is that expected to change with the compat code as well?

lib/compat/wordpress-6.7/script-modules.php
```diff
@@ -0,0 +1,103 @@
...
+		 *
+		 * The data for a given Script Module, if provided, will be JSON serialized in a script tag
+		 * with an ID like `wp-script-module-data-{$module_id}`.
+		 *
+		 * The dynamic portion of the hook name, `$module_id`, refers to the Script Module ID that
+		 * the data is associated with.
+		 *
+		 * @param array $data The data that should be associated with the array.
+		 */
+		$data = apply_filters( "script_module_data_{$module_id}", array() );
```
SantosGuillamot, line 47: Taking a deeper look, I believe this is always returning an empty array because [this code](https://github.com/WordPress/wordpress-develop/blob/a799101e46ef0a042814c189a6331c23a52a100c/src/wp-includes/interactivity-api/class-wp-interactivity-api.php#L311-L312) in core was added in 6.7, so it is not in 6.6.

That means that it is never entering the next conditional and it doesn't run `wp_print_inline_script_tag`.

sirreal: If you use recent Gutenberg and Core trunk, you'd see an identical script tag printed twice (once by Core and once by Gutenberg). The fact that that isn't happening is a good sign 🙂 

To test pre-6.7, you really need to add code which makes it a but uncomfortable to test. If the interactivity API is in use, you could add a filter like this:

```php
add_filter( "script_module_data_@wordpress/interactivity", function ( $data ) {
  $data['foo'] = 'bar';
  return $data;
} );
```

You should see both script tags printed (`id="wp-interactivity-data"` and `id="wp-script-module-data…`). Note that if you do this, Interactivity API won't have server-provided data because it prefers the filtered script tag which won't have the correct data. That's expected and fine, this is only for some manual testing.

lib/compat/wordpress-6.6/script-modules.php
sirreal: That's a good idea, although I believe this is working correctly and I don't think it's too important exactly how we do the feature detection.

I have added a priority, that's certainly a good idea.

lib/compat/wordpress-6.7/script-modules.php
SantosGuillamot: > If you use recent Gutenberg and Core trunk, you'd see an identical script tag printed twice (once by Core and once by Gutenberg).

Yes, this part is working fine in my testing 🙂 

Okay, so let me check if I am understanding it correctly:
* In WordPress 6.7, we are adding the `apply_filters( "script_module_data_{$module_id}", array() )` for first time:  [link](https://github.com/WordPress/wordpress-develop/blob/a799101e46ef0a042814c189a6331c23a52a100c/src/wp-includes/class-wp-script-modules.php#L437-L438).
* That means that until 6.7, no developers should have used the `add_filter(  "script_module_data_..." )` you shared in the example.
* This compatibility code only runs when an external developer uses the `add_filter(  "script_module_data_..." )`.

Is that right? If that's the case, my main question is: Is the compat code really necessary?

(Sorry if it is obvious, I am still trying to fully understand the use case 😄 )

sirreal: Close, but this was introduced in Gutenberg first, so this isn't entirely correct:

> That means that until 6.7, no developers should have used the add_filter(  "script_module_data_..." ) you shared in the example.

---

> Is the compat code really necessary?

I would not add this compat code if it were not already in Gutenberg. But because the code _is_ in Gutenberg, the pre-existing Gutenberg implementation now becomes compat code for WordPress 6.7.

**Existing code of a new (experimental) feature in Gutenberg has transformed to compatibility code for WordPress 6.7.**

SantosGuillamot: Thanks for the context 🙂 I assumed it was introduced as an experiment, so it could be considered as "not added in Gutenberg."

SantosGuillamot (approved):
Testing looks good on my end 🙂 As shared [here](https://github.com/WordPress/gutenberg/pull/64006#discussion_r1701782576), I am still not 100% sure if the compat code is needed at all, but you have much more context than me. It is definitely an improvement from `trunk` because the script is not duplicated.

sirreal:
I think it's good to keep this as "compat" even if that may not be strictly accurate.

If it were removed, we'd effectively have a regression in a Gutenberg feature until Core is released with support for the feature. Keeping it in the compatibility layer makes it clear that it's required until support is dropped.

Issue #63703, opened 2024-07-18
Radii


jameskoster:
<img width="1852" alt="radius" src="https://github.com/WordPress/gutenberg/assets/846565/285a29b1-d525-40b0-9867-78acdf47acd8">

Recent work has expanded the values used for border radius in the UI. Specifically; the new menu component, document title, and command palette all make use of a larger 4px radius. The preview / content frame, and modal radius is larger still: 8px.

A defined scale for radius can add polish to the UI. For example compare how menu items nest neatly inside their container in the new menu component, compared with the current one:

| DropdownMenu V2 | DropdownMenu |
| --- | --- |
| <img width="233" alt="Screenshot 2024-07-11 at 16 32 17" src="https://github.com/WordPress/gutenberg/assets/846565/6eae501a-45fb-4ce5-a26e-ba0315570ef3"> | <img width="221" alt="Screenshot 2024-07-11 at 16 32 26" src="https://github.com/WordPress/gutenberg/assets/846565/4500cdb7-3dc7-494c-9847-62ce89f56eb2"> |

Ad hoc proliferation of values for radius is going to create a maintenance headache down the road, so it would be good to think systematically about a refreshed radius scale, and semantic application thereof.

### Radius scale

<img width="821" alt="radius2" src="https://github.com/user-attachments/assets/ca31cb2a-2be0-4875-9ca1-165e46bde0d0">

Members of the design team have been looking into this and formulated the following scale/guidelines:

- `$radius-x-small`: The smallest radius. Applied to contained elements within primitives like inputs or buttons. E.g. the selected item in a `ToggleGroupControl`, or the unit selector in `UnitControl`.
- `$radius-small`: Applied to primitive components like Badges, Buttons, Inputs, Checkboxes.
- `$radius-medium`: Applied to containers with smaller padding, e.g. Menus, Notices, Snackbars, Block toolbar. Also used for focus rings on primitive components.
- `$radius-large`: Applied to containers with larger padding, e.g. Modals, Pages, and the Preview Frame.
- `$radius-full`: For creating lozenges.
- `$radius-round`: For creating circles or ovals. (This variable already exists).

Note: The _values_ assigned to each item in the scale can be decided separately, but we can discuss those here too.

### Implementation

Once we find consensus around the scale and values, a possible implementation plan might look like:

- [x] https://github.com/WordPress/gutenberg/pull/64007
- [x] https://github.com/WordPress/gutenberg/pull/64368
[...]

PR #64007, opened 2024-07-26
Add radius scale

jameskoster:
## What?
radii

Adds variables for the radius scale as described in https://github.com/WordPress/gutenberg/issues/63703. 

**Guidelines**

- `$radius-x-small`: The smallest radius. Applied to contained elements within primitives like inputs or buttons. E.g. the selected item in a `ToggleGroupControl`, or the unit selector in `UnitControl`.
- `$radius-small`: Applied to primitive components like Badges, Buttons, Inputs, Checkboxes.
- `$radius-medium`: Applied to containers with smaller padding, e.g. Menus, Notices, Snackbars, Block toolbar. Also used for focus rings on primitive components.
- `$radius-large`: Applied to containers with larger padding, e.g. Modals, Pages, and the Preview Frame.
- `$radius-full`: For creating lozenges.
- `$radius-round`: For creating circles or ovals. (This variable already exists).

## Why?

Utilising a consistent scale across will add polish to the UI and reduce maintenance overheads. The diagram below demonstrates how elements would nest neatly inside one another once the scale has been adopted, from the individual button in a `ToggleGroupControl` up to the outer container. The visuals are intentionally abstract to draw focus onto the radius, not the overall appearance. 

radius2

The values could change when they're applied (in a separate effort). For now they are based on the existing convention which is a `2px` radius applied to virtually everything. For this PR the important detail to concentrate on is the scale itself. 

- Are there are enough values?
- Do the guidelines make sense?
- Is the naming convention appropriate?
- Should we note the guidelines in `_variables.scss`?

### DIFF
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -102,12 +102,12 @@ $border-width: 1px;
 $border-width-focus-fallback: 2px; // This exists as a fallback, and is ideally overridden by var(--wp-admin-border-width-focus) unless in some SASS math cases.
 $border-width-tab: 1.5px;
 $helptext-font-size: 12px;
-$radius-block-ui: 2px;
 $radio-input-size: 16px;
 $radio-input-size-sm: 24px; // Width & height for small viewports.
 
 // Deprecated, please avoid using these.
 $block-padding: 14px; // Used to define space between block footprint and surrouding borders.
+$radius-block-ui: $radius-small;
 
 
 /**
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -44,12 +44,12 @@ $grid-unit-80: 8 * $grid-unit;		// 64px
  * Radius scale.
  */
 
-$radius-x-small: 1px;   // Applied to contained elements within primitives like inputs or buttons. E.g. the selected item in a ToggleGroupControl, or the unit selector in UnitControl.
-$radius-small: 2px;     // Applied to primitive components like Badges, Buttons, Inputs, Checkboxes.
-$radius-medium: 4px;    // Applied to containers with smaller padding, e.g. Menus, Notices, Snackbars, Block toolbar. Also used for focus rings on primitive components.
-$radius-large: 8px;     // Applied to containers with larger padding, e.g. Modals, Pages, and the Preview Frame.
-$radius-full: 9999px;   // For lozenges
-$radius-round: 50%;     // For circles and ovals
+$radius-x-small: 1px;   // Applied to elements like buttons nested within primitives like inputs.
+$radius-small: 2px;     // Applied to most primitives.
+$radius-medium: 4px;    // Applied to containers with smaller padding.
+$radius-large: 8px;     // Applied to containers with larger padding.
+$radius-full: 9999px;   // For lozenges.
+$radius-round: 50%;     // For circles and ovals.
 
 /**
  * Dimensions.
--- a/packages/base-styles/_variables.scss
+++ b/packages/base-styles/_variables.scss
@@ -44,10 +44,10 @@ $grid-unit-80: 8 * $grid-unit;		// 64px
  * Radius scale.
  */
 
-$radius-xs: 1px;
-$radius-s: 2px;
-$radius-m: 4px;
-$radius-l: 8px;
+$radius-x-small: 1px;
+$radius-small: 2px;
+$radius-medium: 4px;
+$radius-large: 8px;
 $radius-full: 9999px; // For lozenges
 $radius-round: 50%; // For circles and ovals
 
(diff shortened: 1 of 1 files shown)


mirka:
> Should we note the guidelines in `_variables.scss`?

Yes, that would be helpful!

hanneslsm:
> Is the naming convention appropriate?

Considering that we have in variable.scss already `$button-size-small`, `$modal-width-small`, `$modal-width-medium` and
`$radio-input-size-sm` I'd not shorten the variable and  go with `x-small`, `small`…

(That's also how it's done in TT4 for font sizes)

packages/base-styles/_variables.scss
```diff
@@ -91,7 +102,6 @@ $border-width: 1px;
 $border-width-focus-fallback: 2px; // This exists as a fallback, and is ideally overridden by var(--wp-admin-border-width-focus) unless in some SASS math cases.
 $border-width-tab: 1.5px;
 $helptext-font-size: 12px;
-$radius-round: 50%;
 $radius-block-ui: 2px;
```
tyxla: Should `$radius-s` replace the `$radius-block-ui` or be part of its declaration?

jameskoster: I'd welcome thoughts on that. My idea was to essentially deprecate `$radius-block-ui` by assigning it the `radius-s` value like you suggest, then updating all the styles that utilises it. That second step is a big one though.

tyxla: With the radius scale you're proposing, `$radius-block-ui` becomes even more confusing to have in the first place. But we can always add a TODO comment and refactor away from it in a simultaneous PR, in favor of its `radius-s` counterpart.

jameskoster: > becomes even more confusing to have in the first place

I agree, but I assume it would not be safe to remove?

jameskoster:
> I'd not shorten the variable and go with x-small, small…

Good point.

Edit: `$radius-x-small` looks a bit odd. Should that be `$radius-xsmall`? 🤔

packages/base-styles/_variables.scss
tyxla: No, removing it will be unsafe and could break potential users, but we can:

* Update `$radius-block-ui` to use `$radius-small` instead of a hardcoded value
* Update the entire codebase to use `$radius-small` instead.
* Deprecate (through a comment, similar to others in this file) `$radius-block-ui` so it's clearer that it shouldn't be used.

tyxla:
> Edit: `$radius-x-small` looks a bit odd. Should that be `$radius-xsmall`? 🤔

I'd prefer `-x-small`, I don't see anything odd with it.

ciampo:
Agreed with the points made above. In general, I'd go for consistency with how existing variables are named.

packages/base-styles/_variables.scss
ciampo: The safest thing would be to keep it around, aliasing it to `$radius-small`. We could then also add a deprecation notice (I don't know how those work for Sass variables), and maybe a stylelint rule to prevent further usages.

Not sure if we'd be able to remove it entirely, since a missing Sass variable would cause build errors. So, on one end it would be easy to spot and fix (especially if we add a dev note), on the other end I'm not sure we're ok with causing build errors.

jameskoster: Sounds like a plan :)

jameskoster: I pushed a commit to update/deprecate `$block-radius-ui`: https://github.com/WordPress/gutenberg/pull/64007/commits/ea32b3345694ff445eb935b4482deea81c862730. I don't mind doing that in a separate PR if y'all prefer though.

ciampo (approved):
Current set of changes LGTM 🚀 

As for the next steps:

- we should open a PR where we remove all usages of `$radius-block-ui` in favour of the correct `$radius-*` small;
- we should then open another PR where we add a deprecation notice (if that's a think that can be done in Sass), and also a stylelint rule to prevent further usages of the deprecated variable
- we should add a dev note (maybe to the parent issue?) to make sure that third-party developers are aware of the deprecation and how to correctly update their code

tyxla (approved):
LGTM 🚀 

@ciampo's list of next steps sounds like a good plan to me!

jameskoster:
I'll work on that first step, but might need some assistance with the others.

ciampo:
Sounds good, feel free to ping us to review and help with follow-up tasks :)

ciampo:
I will also mention that we have a [`radiusBlockUi` internal variable](https://github.com/WordPress/gutenberg/blob/a323eeea32e4dde96980514a73bd1c07bae0c7f5/packages/components/src/utils/config-values.js#L45) in the components package for CSS-in-JS, so we should also take care of replacing that variable with the new set of variables up next

PR #64009, opened 2024-07-26
documentation: remove generator terminology

### LINKED ISSUE #63136: Documentation for `@wordpress/data` is stale for WP 6.6
### Description

JS generators will not work anymore starting from WordPress 6.6:

- https://github.com/WordPress/gutenberg/issues/63009#issuecomment-2205491732

However, the [documentation for `@wordpress/data`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/) still mentions the use of generators:

```js
// ...

const store = createReduxStore( 'my-shop', {
    // ...

    resolvers: {
        *getPrice( item ) {
            const path = '/wp/v2/prices/' + item;
            const price = yield actions.fetchFromAPI( path );
            return actions.setPrice( item, price );
        },
    },
} );

// ...
```

It also reads:

> [...] By defining your action creator or resolvers as a generator which yields specific controlled action types, the execution will proceed as defined by the control handler.

This documentation should be adapted.

1. Go to https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/
2. Notice that the docs still mention generators

_No response_

### Environment info

_No response_

Yes

Yes

gagan-bhullar-tech:
## What?

It removes generator terminology.

## Why?

It fixes #63136

### DIFF
--- a/packages/data/README.md
+++ b/packages/data/README.md
@@ -135,7 +135,7 @@ The `resolvers` option should be passed as an object where each key is the name 
 
 #### `controls`
 
-A **control** defines the execution flow behavior associated with a specific action type. This can be particularly useful in implementing asynchronous data flows for your store. By defining your action creator or resolvers as a generator which yields specific controlled action types, the execution will proceed as defined by the control handler.
+A **control** defines the execution flow behavior associated with a specific action type. This can be particularly useful in implementing asynchronous data flows for your store. By defining your action creator which yields specific controlled action types, the execution will proceed as defined by the control handler.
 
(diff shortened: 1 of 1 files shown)


Mamaduka (changes requested):
Thanks for the PR, @gagan-bhullar-tech!

Unfortunately, this doesn't fully resolve the issue. The code examples also need to be updated to showcase the [new thunk-based architecture](https://developer.wordpress.org/block-editor/how-to-guides/thunks/).

PR #64014, opened 2024-07-27
Add lightbox option in gallery block link control

akasunil:
## What?

`Expand on click` option is missing from gallery block `Link To` control. 

## Why?

Image block link control have `Expand on click` option which basically enables the lightbox on image. Gallery block link control have all other option from image block, Except `Expand on click` option.

## How?

Add `Expand on click` option in gallery block `Link To` control.

## Testing Instructions

- Make sure that you have enabled lightbox from theme.json
- Add gallery block
- Check if `Expand on click` option is available under link control
- Test `Expand on click` option function and compare it with `Image` block's lightbox functionality
- Try with different lightbox setting, and note lightbox attributes in image block under gallery block

## Screenshots or screencast 
image


akasunil:
@dcalhoun  I might need your assistance once more with mobile unit test failure here. It ask me to update snapshots with default `LinkTo` attribute removed. Nothing regarding the LinkTo attribute has been changed here.

dcalhoun:
> @dcalhoun I might need your assistance once more with mobile unit test failure here. It ask me to update snapshots with default `LinkTo` attribute removed. Nothing regarding the LinkTo attribute has been changed here.

👋🏻 @akasunil thanks for the mention. It took me a little while to identify the root issue, but it relates to the lack of full support for `__experimentalFeatures` within the native mobile editor. The  `useSettings` [invocation addition](https://github.com/WordPress/gutenberg/pull/64014/files#diff-7be1df22c250a7378a0796b6c68da7fc1178a98d9e88ab9f59868a6d0b8c467fR123) in your proposed changes is what resulted in the new test failure and native mobile editor crash. 

I opened https://github.com/WordPress/gutenberg/pull/64448 addressing the issue. Please feel free to merge it into your branch as you see fit.

packages/block-library/src/gallery/edit.js
```diff
@@ -78,6 +81,13 @@ const linkOptions = [
 		value: LINK_DESTINATION_MEDIA,
 		noticeText: __( 'Media Files' ),
 	},
+	{
+		icon: fullscreen,
+		label: __( 'Expand on click' ),
+		value: LINK_DESTINATION_LIGHTBOX,
+		noticeText: __( 'Lightbox effect' ),
+		infoText: __( 'Scales the image with a lightbox effect' ),
```
t-hamano:
```suggestion
		infoText: __( 'Scale images with a lightbox effect' ),
```

I'm not a native English speaker, but it might be better to use the plural, as well as the other options.

packages/block-library/src/gallery/edit.js
```diff
@@ -110,6 +120,20 @@ function GalleryEdit( props ) {
 		onFocus,
 	} = props;
 
+	const [ blockSetting ] = useSettings( 'blocks' );
+
+	const lightboxSetting = blockSetting.hasOwnProperty( 'core/image' )
+		? blockSetting[ 'core/image' ]?.lightbox
+		: false;
```
t-hamano: This should be easier to write if you specify the path directly:

```javascript
const lightboxSetting = useSettings( 'blocks.core/image.lightbox' );
```

@youknowriad We need to get the settings of the Image block from the Gallery block. Are there any other suitable APIs?

packages/block-library/src/gallery/edit.js
```diff
@@ -110,6 +120,20 @@ function GalleryEdit( props ) {
...
+
+	useEffect( () => {
+		if ( ! lightboxSetting?.allowEditing ) {
+			linkOptions = linkOptions.filter(
+				( option ) => option.value !== LINK_DESTINATION_LIGHTBOX
+			);
+		}
+	}, [ lightboxSetting?.allowEditing ] );
```
t-hamano: I think unnecessary `useEffect` should be avoided. Additionally, I think we need to check `enabled` as well, not just `allowEditing`:

```javascript
	const showLightboxOption =
		lightboxSetting?.enabled && lightboxSetting?.allowEditing;
	const filteredLinkOptions = showLightboxOption
		? linkOptions
		: linkOptions.filter(
				( option ) => option.value !== LINK_DESTINATION_LIGHTBOX
		  );
```

In other words, I'd expect the lightbox to be disabled when the following theme.json is defined:

```json
{
	"version": 3,
	"settings": {
		"blocks": {
			"core/image": {
				"lightbox": {
					"enabled": false
				}
			}
		}
	}
}
```

However, in the Image block, the lightbox does not seem to be disabled even if the above theme.json is defined. This seems like a bug to me.

cc @madhusudhand @artemiomorales

packages/block-library/src/gallery/edit.js
```diff
@@ -366,10 +390,34 @@ function GalleryEdit( props ) {
...
+				hrefAndDestination = {
+					...hrefAndDestination,
+					lightbox: { enabled: false },
+				};
+			} else {
+				hrefAndDestination = {
+					...hrefAndDestination,
+					lightbox: undefined,
+				};
+			}
```
t-hamano: Is it possible to write this conditional statement inside the `hrefAndDestination` function?

packages/block-library/src/gallery/edit.js
```diff
@@ -622,9 +670,12 @@ function GalleryEdit( props ) {
 					<ToolbarDropdownMenu
 						icon={ linkIcon }
 						label={ __( 'Link' ) }
+						toggleProps={ {
+							isPressed: hasLinkTo,
+						} }
```
t-hamano: For now, I don't think the pressed state is necessary. Even if the link settings are changed in the Gallery block, the Image blocks inside may not follow those settings.

packages/block-library/src/gallery/edit.js
```diff
@@ -646,6 +697,7 @@ function GalleryEdit( props ) {
 												onClose();
 											} }
 											role="menuitemradio"
+											info={ linkItem.infoText ?? false }
```
t-hamano:
```suggestion
											info={ linkItem.infoText }
```

packages/block-library/src/gallery/edit.js
youknowriad: Not that I know of, but we might want to provide one (like we do for the global styles UI) where we can pass the block name.

packages/block-library/src/gallery/edit.js
akasunil: This is to indicate the `LinkTo` option is set or not.

packages/block-library/src/gallery/edit.js
akasunil: I believe the "enabled" option applies just to the default state of the lightbox on the image block

packages/block-library/src/gallery/utils.js
```diff
@@ -36,17 +40,32 @@ export function getHrefAndDestination(
 			return {
 				href: image?.source_url || image?.url, // eslint-disable-line camelcase
 				linkDestination: IMAGE_LINK_DESTINATION_MEDIA,
+				lightbox: lightboxSetting?.enabled
+					? { enabled: false }
+					: undefined,
```
t-hamano:
```suggestion
				lightbox: undefined,
```

Why do we check `lightboxSettings.enabled`? Shouldn't lightbox always be undefined here?

packages/block-library/src/gallery/utils.js
```diff
@@ -36,17 +40,32 @@ export function getHrefAndDestination(
...
 			};
 		case LINK_DESTINATION_ATTACHMENT_WP_CORE:
 		case LINK_DESTINATION_ATTACHMENT:
 			return {
 				href: image?.link,
 				linkDestination: IMAGE_LINK_DESTINATION_ATTACHMENT,
+				lightbox: lightboxSetting?.enabled
+					? { enabled: false }
+					: undefined,
```
t-hamano:
```suggestion
				lightbox: undefined,
```

Same as `IMAGE_LINK_DESTINATION_MEDIA`

packages/block-library/src/gallery/utils.js
```diff
@@ -36,17 +40,32 @@ export function getHrefAndDestination(
...
+			};
+		case LINK_DESTINATION_LIGHTBOX:
+			return {
+				href: undefined,
+				lightbox: ! lightboxSetting?.enabled
+					? { enabled: true }
+					: undefined,
```
t-hamano: I think this control in the Gallery block will _always_ apply the selected option to the Image blocks inside, so there is no need to check if the Image block's lightbox is enabled by default.

Additionally, the Image block only has the `enabled` property in the `lightbox` attribute, but if other properties are added in the future, `{enabled: true}` may remove the other properties.

Therefore, we may need to change the arguments of the `getHrefAndDestination` function to something like this:

```javascript
// edit.js
function setLinkTo( value ) {
	// ...
	getBlock( clientId ).innerBlocks.forEach( ( block ) => {
		// ...
		changedAttributes[ block.clientId ] = getHrefAndDestination(
			image,
			value,
			false,
			block.attributes
		);
	} );
	// ...
}

// util.js
export function getHrefAndDestination(
	image,
	galleryDestination,
	imageDestination,
	attributes
) {
	// ...
	switch ( imageDestination ? imageDestination : galleryDestination ) {
		case LINK_DESTINATION_MEDIA_WP_CORE:
		case LINK_DESTINATION_MEDIA:
			return {
				href: image?.source_url || image?.url, // eslint-disable-line camelcase
				linkDestination: IMAGE_LINK_DESTINATION_MEDIA,
				lightbox: {
					...attributes?.lightbox,
					enabled: false,
				},
			};
		// ...
	}

	return {};
}
```

packages/block-library/src/gallery/edit.js
t-hamano: That's right, I was mistaken 😅

packages/block-library/src/gallery/edit.js
t-hamano: Even if the Gallery block has the Link To option set, the Image blocks inside may not link to anything. In this case, I'm concerned that users may be confused if we highlight the `Link To` option in the pressed state.

What do you think?

packages/block-library/src/gallery/utils.js
akasunil: When lightbox is enabled by default, we need to set `lightbox: { enabled: false }` to disable lightbox and set other link options. Checkout [lightbox logic](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/image.js#L272-L286) in image block. If we use `lightbox: undefined`, Lightbox will remain active even for other link destinations.

packages/block-library/src/gallery/edit.js
akasunil: You are right! This is little confusing to me too.

packages/block-library/src/gallery/utils.js
t-hamano: Thank you for the detailed explanation.

Until now, it worked without this additional processing because the conditional statement [here](https://github.com/WordPress/gutenberg/blob/d79d610a4d9e464d6bf583bdf6a140ff2ae262c3/packages/block-library/src/image/index.php#L63-L72) checked whether the link destination was none, even if the lightbox was enabled by default.

In any case, I think it makes sense to make it consistent with the logic of the Image block and explicitly disable it if `enabled` is active by default.

packages/block-library/src/gallery/editor.scss
```diff
@@ -198,3 +198,11 @@
...
 	gap: $grid-unit-15;
 }
+
+.blocks-gallery__link-to-control {
+	.components-menu-item__item {
+		.components-menu-item__info {
+			white-space: nowrap;
+		}
+	}
+}
```
t-hamano: Is this code necessary? From what I've tested, there doesn't seem to be any visual change without this CSS.

akasunil: Yes, the info text for the lightbox option used to be two lines long. However, we adjusted the text to make it shorter and it now stays in a single line without this CSS. I will remove the CSS because it is no longer useful.

t-hamano:
Thanks for the update! 

I think we're ready to ship this PR, but we have #62906 in progress to add lightbox support to the Gallery block. This PR shouldn't conflict with that PR, but it might be better to clarify the status of #62906 first.

If you have any feedback, please leave a comment 🙇‍♂️

madhusudhand (approved):
Thanks for the changes @akasunil. It worked great in my testing! 👏 

Since #62906 is blocked by the i18n as script module (not yet implemented), and this is an independent change, I suggest to merge this PR.

akasunil:
Thank you @madhusudhand @t-hamano and @dcalhoun. I'll merge this PR.

bystrzan:
"First do it, then make it. Then make it work" - this is nightmare, as only theme.json could disable this and we NEED to disable this whole pseudo-lightbox zooming as it's not even in alpha, but in Core. 

I would rather want to see how to block it until it's ready. THAT would be NICE.

akasunil:
@bystrzan You can also disable lightbox from global styles of image block. see below video for more info.

[Dashboard-‹-gutenberg-—-WordPress.webm](https://github.com/user-attachments/assets/db20bb92-ff65-4ff1-ad6d-282a01d3dd8b)

PR #64015, opened 2024-07-27
Avoid errors for post types without a 'menu_icon'

Mamaduka:
## What?
Fixes https://core.trac.wordpress.org/ticket/61751.

PR prevents editors from crashing when a CPT explicitly sets the `menu_icon` property to `false`.

## Why?
Calling `startsWith` on non-string values will trigger an error.

## Testing Instructions
1. Register a custom post type. You can use one of these [examples](https://developer.wordpress.org/reference/functions/register_post_type/#user-contributed-notes).
2. Explicitly set the `menu_icon` property to `false`.
3. Create a new post for this post type.
4. The editor shouldn't crash due to an error. Doing the same on `trunk` will crash the editor.
5. Navigate to Site Editor > Templates.
6. Click "Add New Template"
7. There should be no JS errors.

Same.

### DIFF
--- a/packages/edit-site/src/components/add-new-template/utils.js
+++ b/packages/edit-site/src/components/add-new-template/utils.js
@@ -166,9 +166,11 @@ export function usePostTypeArchiveMenuItems() {
 						// `icon` is the `menu_icon` property of a post type. We
 						// only handle `dashicons` for now, even if the `menu_icon`
 						// also supports urls and svg as values.
-						icon: postType.icon?.startsWith( 'dashicons-' )
-							? postType.icon.slice( 10 )
-							: archive,
+						icon:
+							typeof postType.icon === 'string' &&
+							postType.icon.startsWith( 'dashicons-' )
+								? postType.icon.slice( 10 )
+								: archive,
 						templatePrefix: 'archive',
 					};
 				} ) || [],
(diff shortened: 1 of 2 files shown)


t-hamano (approved):
LGTM! I've tested the following variations and they all work fine, including the post type where the value is `false`:

```php
function setup_post_type() {
	$args = array(
		'public'       => true,
		'label'        => 'Dashicon',
		'menu_icon'    => 'dashicons-book',
		'show_in_rest' => true,
	);
	register_post_type( 'dashicon', $args );

	$args = array(
		'public'       => true,
		'label'        => 'Null Icon',
		'menu_icon'    => null,
		'show_in_rest' => true,
	);
	register_post_type( 'null-icon', $args );

	$args = array(
		'public'       => true,
		'label'        => 'Empty String Icon',
		'menu_icon'    => '',
		'show_in_rest' => true,
	);
	register_post_type( 'empty-string-icon', $args );

	$args = array(
		'public'       => true,
		'label'        => 'False Icon',
		'menu_icon'    => false,
		'show_in_rest' => true,
	);
	register_post_type( 'false-icon', $args );
}
add_action( 'init', 'setup_post_type' );
```

karan4official:
Thanks for working on the fix. This issue prevented CMS users from editing or creating custom post types when the custom post type menu icon was not set (default: false). The block editor relies on the REST API schema, which specifies icon values as either strings or null, causing this inconsistency.

hellofromtonya:
Hello @karan4official,

To give you contributor credit for WP 6.6.2 release, can you please confirm that this [WordPress profile](https://profiles.wordpress.org/nuancedesignstudio/) is yours? Thank you.

karan4official:
No, my profile is https://profiles.wordpress.org/karan4word/

hellofromtonya:
Thank you @karan4official. I've added you. Cheers

cbravobernal:
Removing the label as it was already backported and included in WP 6.6.2

Issue #63323, opened 2024-07-09
No ability to edit Inline Image alternative text in context


mrwweb:
### Description

Currently, to the best of my knowledge, it is only possible edit an inline image's global alt text (without editing HTML). This is an issue as most alt text should be contextual. I'm currently in a place where I need my inline images to have empty alt text while it's very important that they retain alt text for use in galleries throughout the site.

### Step-by-step reproduction instructions

1. Insert inline image
2. You can't edit the alt text

### Environment info

- WP 6.6-RC2

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

No

PR #64016, opened 2024-07-27
Format Library: Polish inline image format popover

t-hamano:
## What?

This PR improves the popover UI for applying widths to inline image.

## Why?

The current UI is not consistent compared to other UIs.

## How?

I have standardized all sizes and spacing to 8px grid units.

Note: In the future, I aim to add a textarea for `alt` attribute, as suggested by #63323.

## Testing Instructions

Insert an inline image and click it.

## Screenshots or screencast 

| | Before | After |
|--------|--------|--------|
| Default | before | after |
| Submit button focuses | before-submit-button-focused | afuter-submit-button-focused |

### DIFF
--- a/packages/format-library/src/image/index.js
+++ b/packages/format-library/src/image/index.js
@@ -76,19 +76,20 @@ function InlineUI( { value, onChange, activeObjectAttributes, contentRef } ) {
 					event.preventDefault();
 				} }
 			>
-				<HStack alignment="bottom" spacing="0">
+				<HStack alignment="bottom">
 					<NumberControl
-						className="block-editor-format-toolbar__image-container-value"
 						label={ __( 'Width' ) }
 						value={ width }
 						min={ 1 }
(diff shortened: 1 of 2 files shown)


jameskoster:
Good change 👍 

Should the button also be 40px, so that it lines up nicely with the input (as it does on trunk)?

t-hamano:
Thanks for the review!

I changed the button size to `40px`:

image

image

I think `32px` might be ok. It would look something like this. What do you think?

image

image

jameskoster:
I don't have a strong feeling, I was mostly curious whether the button + input should be aligned in terms of height, since they are already on trunk.

32px may be fine since the popover is so small.

t-hamano:
I don't have a strong opinion either, but 40px is also fine for now. Could we reconsider the size when we try adding the alt textarea in the future?

jameskoster (approved):
Let's go for 40 then, since that is the more widely used size :)

PR #16475, opened 2019-07-09
Add circle-crop variation to Image block.

jasmussen:
Props @phpbits for the technical implementation.

This PR adds a variation to the Image block, "Circle Crop". What that means is you can choose a style variation for the image, which rounds all 4 corners. When this is appliedto a square image, that means a perfect circle. When it is applied to a rectangle, it means a pill-shape.

In past attempts to fix this, it was either an oval crop, or it required a large refactor to be able to make a rectangular image into a square, in CSS only. But with @phpbits clever trick — using a pixel value for the border radius — the we can actually implement this in a simple way that is still visually appealing even for rectangles.

GIF:

circle crop

---

On a separate note, I did look into sort of a "hack" to create perfect circles from rectangular images. Since you can manually type in the dimensions of an image in the block inspector, you can _force_ a square image even from a rectangle. If you combine that with `object-fit: cover;`, it means that the image won't be stretched. 

However, every modern WordPress theme has the following CSS bundled along, to make images responsive:

```
img {
	max-width: 100%;
	height: auto;
}
```

This reminded me that unless the width of the image is fluid, it won't scale to mobile. So the hack unfortunately won't work.

PR #64018, opened 2024-07-27
Update www.w3.org/2000/svg from `http` to `https`

narenin:
##Trac Ticket:

https://core.trac.wordpress.org/ticket/61777

## What?
This PR will update www.w3.org/2000/svg from `http` to `https` because of this in Windows defender Trojan error for `wp-includes\dist\block-library\style.min.css`

## Why?
This PR is necessary because it will Windows defender Trojan error we are getting for `wp-includes\dist\block-library\style.min.css` 

## How?
This PR will change `http://www.w3.org/2000/svg` to `https://www.w3.org/2000/svg` which will fix the issue.

## Testing Instructions
1. Just downloaded and unpacked the latest version (6.6.1) from https://wordpress.org/download/  
2. windows defender removed `wp-includes\dist\block-library\style.min.css` stating it was a trojan. 

## Screenshots or screencast

trac_error

### DIFF
--- a/packages/block-library/src/image/style.scss
+++ b/packages/block-library/src/image/style.scss
@@ -114,7 +114,7 @@
 		// If a browser supports it, we will switch to using a circular SVG mask.
 		@supports (mask-image: none) or (-webkit-mask-image: none) {
 			/* stylelint-disable-next-line function-url-quotes -- We need quotes for the data URL to use the SVG inline. */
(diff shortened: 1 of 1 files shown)


sabernhardt:
Windows Defender should recognize that `xmlns` is an XML namespace, not a source or a link. The "opaque identifier" [needs to remain 'http'](https://github.com/w3c/svgwg/issues/738) for the mask image to be identified as an SVG.

SVG added in #16475

Otto42:
This is an XML namespace, not a URL. It should not be changed.

PR #51602, opened 2023-06-16
`Modal`: fix closing when contained iframe is focused

stokesman:
## What?
Fix for a bug in `Modal` it closes when a contained iframe is focused.

## Why?
It seems like iframes in modals should be supported.
to fix: #40912

## How?
Uses an event handler on the Modal’s overlay element to trigger the closing instead of the `useFocusOutside` hook.

## Testing Instructions
1. Visit the Modal story in Storybook.
2. Open the modal.
3. Verify that clicking/focusing the contained iframe does not close the modal.
4. Verify that clicking outside the modal closes the modal.

PR #64019, opened 2024-07-27
Fix 'Preferences' and 'Shortcuts' commands in StrictMode

### LINKED ISSUE #64004: "Editor Preferences" and "Keyboard shortcuts" commands don't work
### Description

I found that in both the post editor and site editor, running the following two commands did not launch the dialog:

- `Editor Preferences`
- `Keyboard shortcuts`

Both commands launch a modal dialog. This is very strange, but I found that it happens when `StrictMode` is enabled.

Namely, the following parts. When I turn off `StrictMode`, these two commands work again.

https://github.com/WordPress/gutenberg/blob/41d5ff2e221fe65529c2c6a437c422770e6c59d7/packages/edit-post/src/index.js#L146-L155

https://github.com/WordPress/gutenberg/blob/41d5ff2e221fe65529c2c6a437c422770e6c59d7/packages/edit-site/src/index.js#L101-L105

- Launch the command palette.
- Run "Editor Preferences" or "Keyboard shortcuts"

https://github.com/user-attachments/assets/23f766e3-66d7-4798-8c2b-c2d1dcd837bc

### Environment info

_No response_

- [X] Yes

- [X] Yes

Mamaduka:
## What?
Closes #64004.

PR fixes a bug where the "Editor Preferences" or "Keyboard shortcuts" modal will close immediately when opening from the command palette. There's no modal flash, so it looks like the commands never run.

Note: While this fix is correct, it still fixes a symptom, not the actual cause. I've included more details below.

## Why?
I noticed that `onRequestClose` callbacks are called for both models as soon as they open. This led me to "dismissers" logic in the Modal component, introduced in #51602. If I comment on the logic for "dismissers," the modal opens correctly. It seems that this side effect isn't resilient for the [StrictMode](https://react.dev/reference/react/StrictMode).

https://github.com/WordPress/gutenberg/blob/159d01a01b61ee76159ab67bc94d0d7cb3d90b35/packages/components/src/modal/index.tsx#L152-L166

## How?
Explicitly close the command palette dialog before opening other models. This matches other `callback` calls in the file.

## Testing Instructions
1. Enable StictMode - `WP_DEBUG = true` and `SCRIPT_DEBUG = true`.
2. Launch the command palette.
3. Run "Editor Preferences" or "Keyboard shortcuts" commands.

Same

### DIFF
--- a/packages/editor/src/components/commands/index.js
+++ b/packages/editor/src/components/commands/index.js
@@ -89,7 +89,8 @@ function useEditorCommandLoader() {
 		name: 'core/open-shortcut-help',
 		label: __( 'Keyboard shortcuts' ),
 		icon: keyboard,
-		callback: () => {
+		callback: ( { close } ) => {
+			close();
 			openModal( 'editor/keyboard-shortcut-help' );
 		},
 	} );
@@ -108,7 +109,8 @@ function useEditorCommandLoader() {
 	commands.push( {
 		name: 'core/open-preferences',
 		label: __( 'Editor preferences' ),
-		callback: () => {
+		callback: ( { close } ) => {
+			close();
 			openModal( 'editor/preferences' );
 		},
 	} );


t-hamano (approved):
LGTM! This change makes sense to me.

> If I comment on the logic for "dismissers," the modal opens correctly.

I found an article that said React 18 intentionally remounts components when Strict Mode is on.

Maybe this is related.

> Adding `<StrictMode>` to a React application adds special behavior (only in DEV mode) to all of components it wraps around. For example, when running in “strict mode“ React will [intentionally double-render components](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects) for in order to flush out unsafe side effects.

From: [Adding Reusable State to StrictMode · reactwg/react-18 · Discussion #19](https://github.com/reactwg/react-18/discussions/19)

Mamaduka:
> I found an article that said React 18 intentionally remounts components when Strict Mode is on.

Yes, it does, to help developers discover potential bugs. The documentation page explains the reasoning behind it pretty well - https://react.dev/reference/react/StrictMode.

stokesman:
I’d wanted to try debugging the effect that George highlighted but I've not been able to reproduce the issue. It seems like React is running in production mode for me and I can’t figure out why. Logging `process.env.NODE_ENV === 'development'` shows `true` but I'm not seeing double logging from renders or effects. I sure would appreciate any ideas on how that could be 🥺.

Mamaduka:
@stokesman, check if these constants are set to true WP_DEBUG and SCRIPT_DEBUG.

Basically, React needs to load in development mode for StrictMode to be active.

stokesman:
Thank you George. That makes sense and now I finally read the instructions where you already mentioned it 🤦.

I made a #64075 with what should be a proper fix.

PR #64025, opened 2024-07-28
Tweak list block 

### LINKED ISSUE #64026: Improve Inspector controls for List block
The List block could use a little attention, particularly around simplifying the labels, moving the list style ("Numbering style") control above the start value, and using the 40px controls. The block descriptions aren't particularly useful to identify what the block is for either. 

Small details, but something I've noticed while writing posts lately. 

image
image

richtabor:
## What?
Closes https://github.com/WordPress/gutenberg/issues/64026. Small tweaks to the List block: 
- Change "Numbering style" label to something much more recognizable: "List Style". 
- Move the List Style control above the Start Value control; it's much more likely to be used, so should be given priority.
- Change "Reverse list numbering" label to simply "Reverse". 
- Update block descriptions to be more descriptive. 
- Remove default background color control from the List Item; also very unlikely to be used and should not be a default. 
- Add __next40pxDefaultSize. 

## Testing Instructions

1. Open a post or page.
2. Insert a List block. 
3. Add List Items. 
4. See changes.

### Before

CleanShot 2024-07-27 at 20 48 53

CleanShot 2024-07-27 at 20 48 59

### After
CleanShot 2024-07-27 at 20 43 47

CleanShot 2024-07-27 at 20 43 55

### DIFF
--- a/packages/block-library/src/list-item/block.json
+++ b/packages/block-library/src/list-item/block.json
@@ -25,8 +25,8 @@
 		"color": {
 			"gradients": true,
 			"link": true,
+			"background": true,
 			"__experimentalDefaultControls": {
-				"background": false,
 				"text": true
 			}
 		},
--- a/packages/block-library/src/list-item/block.json
+++ b/packages/block-library/src/list-item/block.json
@@ -6,7 +6,7 @@
 	"category": "text",
 	"parent": [ "core/list" ],
 	"allowedBlocks": [ "core/list" ],
-	"description": "Create a list item.",
+	"description": "An individual item within a list.",
 	"textdomain": "default",
 	"attributes": {
 		"placeholder": {
@@ -26,7 +26,7 @@
 			"gradients": true,
 			"link": true,
 			"__experimentalDefaultControls": {
-				"background": true,
+				"background": false,
 				"text": true
 			}
 		},
--- a/packages/block-library/src/list/ordered-list-settings.js
+++ b/packages/block-library/src/list/ordered-list-settings.js
@@ -61,7 +61,7 @@ const OrderedListSettings = ( { setAttributes, reversed, start, type } ) => (
 			/>
 			<ToggleControl
 				__nextHasNoMarginBottom
-				label={ __( 'Reverse' ) }
+				label={ __( 'Reverse order' ) }
 				checked={ reversed || false }
 				onChange={ ( value ) => {
 					setAttributes( {
--- a/packages/block-library/src/list/ordered-list-settings.js
+++ b/packages/block-library/src/list/ordered-list-settings.js
@@ -13,25 +13,10 @@ import {
 const OrderedListSettings = ( { setAttributes, reversed, start, type } ) => (
 	<InspectorControls>
 		<PanelBody title={ __( 'Settings' ) }>
-			<TextControl
-				__nextHasNoMarginBottom
-				label={ __( 'Start value' ) }
-				type="number"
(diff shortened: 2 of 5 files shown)


packages/block-library/src/list-item/block.json
```diff
@@ -26,7 +26,7 @@
 			"gradients": true,
 			"link": true,
 			"__experimentalDefaultControls": {
-				"background": true,
+				"background": false,
 				"text": true
 			}
```
t-hamano: This way of writing certainly works, but I feel it's hard to know whether the background is enabled or not.

In general, I think it's better to write it like this:

```
"color": {
	"gradients": true,
	"link": true,
	"background": true,
	"__experimentalDefaultControls": {
		"text": true
	}
},
```

Also, once this PR is merged, you may want to change this part of the table in #43245 to "Optional":

list-item

packages/block-library/src/list-item/block.json
```diff
@@ -6,7 +6,7 @@
 	"category": "text",
 	"parent": [ "core/list" ],
 	"allowedBlocks": [ "core/list" ],
-	"description": "Create a list item.",
+	"description": "An individual item within a list.",
```
t-hamano, line 9:
```suggestion
	"description": "An individual item within a list block.",
```

I don't have a strong opinion, but it might be a good idea to match the description in the Column block:

https://github.com/WordPress/gutenberg/blob/313246a01f18e504dabd8313e7eacca728332bcd/packages/block-library/src/column/block.json#L8

t-hamano (commented):
image

packages/block-library/src/list/ordered-list-settings.js
```diff
@@ -57,9 +42,26 @@ const OrderedListSettings = ( { setAttributes, reversed, start, type } ) => (
...
+						start: isNaN( int ) ? undefined : int,
+					} );
+				} }
+				value={ Number.isInteger( start ) ? start.toString( 10 ) : '' }
+				step="1"
+			/>
 			<ToggleControl
 				__nextHasNoMarginBottom
-				label={ __( 'Reverse list numbering' ) }
+				label={ __( 'Reverse' ) }
```
ndiego: This is better than "Reverse list numbering" to me, but what about "Reverse order"? That provides a tad more context.

packages/block-library/src/list-item/block.json
ndiego: > This way of writing certainly works, but I feel it's hard to know whether the background is enabled or not.

Agreed, it would be much better from an extender's perspective to have each support clearly written out, even if some are enabled by default.

ndiego:
Overall, looks good. Added a few minor comments.

jasmussen (approved):
Nice one. Some good feedback shared by the others, and here's a green check for good measure.

packages/block-library/src/list-item/block.json
richtabor: Will do, thanks.

packages/block-library/src/list-item/block.json
richtabor: Perhaps the columns description could be simplified a bit next. I don't think it's necessary to include block in the description. It could make it feel more harsh, more "block-like" than it really is.

packages/block-library/src/list/ordered-list-settings.js
richtabor: Sure thing. Thanks!

PR #58834, opened 2024-02-08
Font Library: Convert heading text to heading elements and group fonts as a list

t-hamano:
Related to #55220
Part of #58403
See this comment: https://github.com/WordPress/gutenberg/issues/58403#issuecomment-1932358501

## What?

This PR achieves the following two things in the font library modal.

- Convert heading text to heading elements
- Group fonts as a list

font-list

## Why?

~~User-installed fonts don't have the title, so Users don't know what that font list means.~~ Also, since the font list is actually just a list of button elements, users using screen readers cannot understand how many fonts there are.

## How?

Group the font list as follows.

```html
<h2>XXX Fonts</h2>
<ul>
	<li>
		<button>...</button>
	</li>
	<li>
		<button>...</button>
	</li>
</ul>
```

> [!NOTE]
> ~~By using `fieldset` and `legend`, we should be able to properly associate the heading with the content. But maybe these elements should be inside the `form` element. I would appreciate it if you could give me advice on whether this implementation violates a11y.~~

### Testing Instructions for Keyboard

- Launch a screen reader such as NVDA or VoiceOver.
- Open the font library modal just by using the keyboard.
- When you focus on the beginning of the font list: You should hear something like this: `Library  property page  list  with X items  {fontName}  graphic    X/X variants active  button  

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/54422211/919415e2-4dc4-43b8-b11b-9da9efe76b1e

PR #64029, opened 2024-07-28
Font Library Modal: Group font variations as a list

t-hamano:
Similar to #58834

## What?

This PR groups fonts as a list in the following two screens:

Font Library Modal > Library Tab > Select a font

library

Font Library Modal > Install Fonts Tab > Select a font

install-fonts

## Why?

The current font variation screen only displays a list of button elements, so screen reader users have no way of knowing how many fonts are displayed.

## How?

Similar to #58834, wrap each font in a list item element and wrap those in a list element.

## Testing Instructions

This PR is just a markup change and should not change anything in behavior or appearance. If you want to check the differences between files, [this split view](https://github.com/WordPress/gutenberg/pull/64029/files?diff=split&w=1) will be easy to understand.

- Launch a screen reader such as NVDA or VoiceOver.
- Open the font library modal just by using the keyboard.
  -  Go to Library Tab > Select a font
  - Install Fonts Tab > Select a font
- When you focus on the beginning of the font list: You should hear something like this: `list  with 2 items  ABeeZee Normal  check box  not checked`


afercia:
Screenshot of Safari and VoiceOver correctly announcing the total items and item position:

Screenshot 2024-08-30 at 14 35 44

afercia (approved):
Thanks for this PR @t-hamano from an accessibility perspective looks good to me.

PR #64030, opened 2024-07-28
Font Library Modal: Remove notice context and add message when fonts updated

### LINKED ISSUE #54601: Font Library: No message appears when pressing the Update button

In the font library, for example, when you change the variation you use, you can click the Update button. When you click that button, the state is indeed saved, but no message is displayed.

In my opinion, there is a solution to either of the following:

- Use the Snackbar component to notify that settings have been properly updated.
- Remove Update button.

Personally, I prefer the latter approach. As I understand it, the behavior of the Update button has the same role as the Save button in the Global Styles. Also, the font variation setting is part of the global style.

I think whether or not to save changes made in global styles should be centralized in one button.

## Screencast

https://github.com/WordPress/gutenberg/assets/54422211/2d9c12cc-c382-4f0e-8343-9a009e510289

t-hamano:
Fixes #54601
Part of #58428

## What?

This PR achieves the following two things:

- Remove the notification status managed as context and store the notification status as a state for each tab
- Show a message when the active font is updated

## Why?

The current notification status is shared across the entire Font Library modal. Therefore, we need to reset the notification status when we open the Font Library or switch tabs.

The notification status is specific to each tab and does not need to be managed centrally as a context.

## How?

Store the notification status as a state for each tab.  This ensures that the notification is reset when you open the Font Library modal or switch tabs.

## Testing Instructions

- Install/upload/remove fonts and update font variations.
- For every action, a notification message should appear.
- The notification message should disappear when you go back to the previous page or switch tabs.

## Screenshots or screencast 

https://github.com/user-attachments/assets/3d2215e6-de01-47da-a19b-a433f5bf9f83

### DIFF
--- a/packages/edit-site/src/components/global-styles/font-families.js
+++ b/packages/edit-site/src/components/global-styles/font-families.js
@@ -26,7 +26,7 @@ import { unlock } from '../../lock-unlock';
 const { useGlobalSetting } = unlock( blockEditorPrivateApis );
 
 function FontFamilies() {
-	const { baseCustomFonts, modalTabOpen, setModalTabOpen, setNotice } =
+	const { baseCustomFonts, modalTabOpen, setModalTabOpen } =
 		useContext( FontLibraryContext );
 	const [ fontFamilies ] = useGlobalSetting( 'typography.fontFamilies' );
 	const [ baseFontFamilies ] = useGlobalSetting(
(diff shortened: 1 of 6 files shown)


mikachan (approved):
Thanks @t-hamano, this is looking good to me!

- I saw notices when installing, uploading, and removing fonts ✅
- I saw notices when updating font variations ✅ 
- All notices disappear on page/tab change ✅

I've just started a re-run of the failed test as the failure didn't seem to be anything to do with these changes.

t-hamano:
@mikachan Thanks for the review!

PR #64032, opened 2024-07-28
Several Typo Correction in Inline Doc

shail-mehta:
Corrected Typo in Inline Documents in the Following Files:

Replaced `direclty` with `directly` :

1. `docs/reference-guides/data/data-core.md` on line 758
2. `packages/core-data/README.md` on line 267
3. `packages/core-data/src/actions.js` on line 203


t-hamano (approved):
Nice catch 🚀

PR #64039, opened 2024-07-29
Global Styles: add tooltips to the heading level selectors

### LINKED ISSUE #64034: Add tooltips to the heading level selectors in Global Styles
This issue is related to #64033. During a recent discussion with community members about no-low/low-code theme development, we realized that there are no tooltips on the heading level selectors in the Typography > Headings panel in Global Styles. 

This is problematic for users who might not know that headings in HTML are referred to as `h1`, `h2`, etc. Simply adding a tooltip that reads "Heading One", "Heading Two", etc. should be sufficient. Pretty much everything else in the UI already has a tooltip, so this enhancement also improves consistency. 

| Current | Proposed |
|-|-|
|image|image|

ramonjd:
## What?

Resolves https://github.com/WordPress/gutenberg/issues/64034

This PR adds tooltips for the typography headings radio items.

## Why?

To describe the HTML element codes.

## How?
⌨️ 

## Testing Instructions
1. Fire up this branch and head to the Site Editor
2. Open the global styles panel, and select the Typography > Headings controls
3. Check that each Heading option element has a tooltip.

## Screenshots or screencast 

https://github.com/user-attachments/assets/acfc1bff-63d7-47b0-baf5-dc70e3521029

### DIFF
--- a/packages/edit-site/src/components/global-styles/screen-typography-element.js
+++ b/packages/edit-site/src/components/global-styles/screen-typography-element.js
@@ -67,30 +67,44 @@ function ScreenTypographyElement( { element } ) {
 					>
 						<ToggleGroupControlOption
 							value="heading"
+							showTooltip
+							aria-label={ __( 'All headings' ) }
 							label={ _x( 'All', 'heading levels' ) }
 						/>
 						<ToggleGroupControlOption
 							value="h1"
+							showTooltip
+							aria-label={ __( 'Heading 1' ) }
 							label={ __( 'H1' ) }
 						/>
 						<ToggleGroupControlOption
 							value="h2"
+							showTooltip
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/components/global-styles/screen-typography-element.js
```diff
@@ -67,30 +67,44 @@ function ScreenTypographyElement( { element } ) {
 					>
 						<ToggleGroupControlOption
 							value="heading"
+							showTooltip
+							aria-label={ __( 'All headings' ) }
```
ramonjd, line 71: I added this for consistency: so that all options have a tooltip.

t-hamano (approved):
LGTM 👍

PR #64043, opened 2024-07-29
DataViews: Conditionally shows the description field in Template Grid layout

### LINKED ISSUE #63943: In Grid layout the template description field renders when empty
Screenshot 2024-07-25 at 15 11 39

amitraj2203:
## What?
Fixes: #63943 

## Why?
In Grid layout the template description field renders when empty, which isn't making sense.

## How?
Conditionally shows the description.

## Testing Instructions
1. Go to Appearance > Editor > Templates.
2. Create a new template without providing any description.
3. Check the **Templates** page, you will see that the description label is not displayed.
4. Now, add a description and check the **Templates** page again. This time, both the description label and the description should be visible.

## Screenshots or screencast 
https://github.com/user-attachments/assets/a3056982-9f48-448d-be18-45e2aeced40d

### DIFF
--- a/packages/dataviews/src/layouts/grid/style.scss
+++ b/packages/dataviews/src/layouts/grid/style.scss
@@ -71,10 +71,6 @@
 			align-items: flex-start;
 			min-height: $grid-unit-30;
 
-			&:has(.dataviews-view-grid__field-value:not(:empty)) {
-				display: flex;
-			}
-
 			&:not(:has(.dataviews-view-grid__field-value:not(:empty))) {
 				display: none;
 			}
--- a/packages/dataviews/src/layouts/grid/style.scss
+++ b/packages/dataviews/src/layouts/grid/style.scss
@@ -71,6 +71,14 @@
 			align-items: flex-start;
 			min-height: $grid-unit-30;
 
+			&:has(.dataviews-view-grid__field-value:not(:empty)) {
+				display: flex;
+			}
+
+			&:not(:has(.dataviews-view-grid__field-value:not(:empty))) {
+				display: none;
+			}
+
 			&:not(.is-column) {
 				align-items: center;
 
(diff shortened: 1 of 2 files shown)


jameskoster:
Thank you, this seems to be working for me :)

youknowriad:
Unfortunately, this is not the right fix, this hard codes field ids within the generic DataViews component. 

I actually don't think there's an easy fix here. Maybe using CSS `:has` or something.

amitraj2203:
@youknowriad Thank you for your review. Could you please check it again? I have applied CSS to fix the issue.

youknowriad:
This works for me. Can you confirm it's still ok for you @jameskoster

jameskoster:
Yup, this fixes the issue :)

packages/dataviews/src/layouts/grid/style.scss
```diff
@@ -71,6 +71,14 @@
 			align-items: flex-start;
 			min-height: $grid-unit-30;
 
+			&:has(.dataviews-view-grid__field-value:not(:empty)) {
+				display: flex;
```
jameskoster: Is this necessary? It doesn't seem to get applied, and I noticed no difference after commenting it out to test.

amitraj2203: It is being applied. I have added it for better readability, even though it will work without it. If you prefer, I can remove it.
image (1)

youknowriad (approved):
LGTM 👍

packages/dataviews/src/layouts/grid/style.scss
jameskoster: Ah yes, I see. However, `display: flex` is already applied by the containing `Flex` component, so redeclaring it here doesn't seem to serve a purpose?

amitraj2203: Okay, I have removed it.

jameskoster (approved):
Thanks for the contribution :)

PR #63544, opened 2024-07-15
Quote: Add border support

aaronrobertshaw:
Part of: https://github.com/WordPress/gutenberg/issues/43241

## What?

Adopts border support for the Quote block.

## Why?
- Offers greater design flexibility
- Improves consistency in design tooling with other blocks

## How?

- Adopts all the available border block supports

## Testing Instructions

- In the site editor, add a Quote block to a page
- Style via Global Styles and confirm the editor and frontend display correctly
- Override the global styles on the block instance and confirm they display appropriately in the editor and frontend

## Screenshots or screencast 

https://github.com/user-attachments/assets/5a8e2e64-341d-42c2-8442-2d57af206ab1

Issue #37752, opened 2022-01-06
Local block appearance controls should reflect values from global styles / theme.json


jameskoster:
### Current implementation

My `theme.json` declares a text color for h2s like so:

```
"color": {
    "text": "var(--wp--preset--color--grey-light)"
}
```

When I select a h2 in the Editor and open the color panel, I see this:

<img width="1204" alt="Screenshot 2022-01-06 at 14 57 58" src="https://user-images.githubusercontent.com/846565/148402561-cbf75a9b-61bf-481e-8aa3-6c5937707fb0.png">

The canvas is telling me one thing, but the control is telling my something else. I can clearly see that a color _is_ set, but the 🚫 icon in the control makes it look like no color is set. When I open the color picker, the transparent texture confuses things further by suggesting that the color is set to `transparent`. This combines to erode my confidence on what will actually appear on the frontend.

### Expectation

My expectation would be to see something like this, where the control reflects the code exactly:

<img width="1186" alt="Screenshot 2022-01-06 at 15 00 48" src="https://user-images.githubusercontent.com/846565/148403040-f9cb4225-eab0-4188-ba7e-dbd5fd8555f9.png">

### Suggestion

We show all values that are inherited from theme.json / Global Styles. Here's an example showing the Button block as styled globally, showing its values — colors, font size, padding, radius — in the inspector even locally. 

![Inheritance option 8](https://github.com/user-attachments/assets/120c0b96-bd04-4e77-9668-b44086ec1fca)

* Color: show the value directly in the color panel, instead of the "unset" color. In the flyout, show the transparent grid to indicate an unset color, but show below that the inherited color, in the format of "[Source]: [Color-name]". For starters, "Styles" is suffient as a source.
* Typography: Show the inherited value as a gray ghost.
* Sliders: Show the inherited value as a gray ghost.
* Input values (border, radius, etc), show the inherited values as placeholder text.

This would be a first stap that doesn't necessarily preclude further steps taken in #49278 to visualize _where_ the inheritance is coming from, but it might also reduce the issue enough that it'd be less urgent, perhaps even irrelevant. It would also make clearer when the contrast notice appears or doesn't; at the moment the values compared for contrast are not aware of inheritance, but with this change they could be made to.

Issue updated Aug 1, 2024.

Mamaduka:
I can't remember the issue/comment right now, but I believe the main problem is getting values from CSS variables.

Do you get the same results if you provide color value as HEX?

[...]

PR #64045, opened 2024-07-29
Quote: Prevent block theme styles overriding global border and padding

aaronrobertshaw:
Addresses: https://github.com/WordPress/gutenberg/pull/63544#issuecomment-2252278726

## What?

Fixes an issue where the Quote block's theme styles have higher specificity than global styles preventing the recently adopted border support styles from applying correctly when the quote block is center or right-aligned.

This fix now also allows global styles padding for the quote block to apply as expected when the block is aligned.

## Why?

Provides the expected styling behaviour for the Quote block.

## How?

- Wraps the chained alignment (and block style) classes with `:where` so the resulting selector has `0-1-0` specificity to match global styles

## Testing Instructions

1. Before checking out this PR, add some Quote blocks to a page or post. Then apply center and right alignments to a couple of them. For convenience check out the snippet below:
```html
<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--40);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--40)"><!-- wp:quote {"className":"is-style-default"} -->
<blockquote class="wp-block-quote is-style-default"><!-- wp:paragraph -->
<p>Wise words of wisdom - no alignment</p>
<!-- /wp:paragraph --></blockquote>
<!-- /wp:quote -->

<!-- wp:quote {"textAlign":"left","className":"is-style-default"} -->
<blockquote class="wp-block-quote has-text-align-left is-style-default"><!-- wp:paragraph -->
<p>Wise words of wisdom - left aligned</p>
<!-- /wp:paragraph --></blockquote>
<!-- /wp:quote -->

<!-- wp:quote {"textAlign":"center","className":"is-style-default"} -->
<blockquote class="wp-block-quote has-text-align-center is-style-default"><!-- wp:paragraph -->
<p>Wise words of wisdom - center aligned</p>
<!-- /wp:paragraph --></blockquote>
<!-- /wp:quote -->

<!-- wp:quote {"textAlign":"right","className":"is-style-default"} -->
<blockquote class="wp-block-quote has-text-align-right is-style-default"><!-- wp:paragraph -->
<p>Wise words of wisdom - right aligned</p>
<!-- /wp:paragraph --></blockquote>
<!-- /wp:quote --></div>
```
2. In the Site Editor, navigate to Styles > Blocks > Quote. Add a border and some custom padding
3. Save and check your quote blocks to see the center and right aligned blocks displaying incorrect borders/padding.
4. Apply the "Plain" block style to one of the blocks and see the border disappear
5. Apply this PR's changes and reload the editor.
6. Confirm the styles now display correctly in the editor and frontend.

## Screenshots or screencast 

| Before | After |
|---|---|
| Screenshot 2024-07-29 at 5 31 21 PM | Screenshot 2024-07-29 at 5 32 00 PM |

### DIFF
--- a/packages/block-library/src/quote/theme.scss
+++ b/packages/block-library/src/quote/theme.scss
@@ -10,19 +10,21 @@
 		font-style: normal;
 	}
 
-	&.has-text-align-right {
+	&:where(.has-text-align-right) {
 		border-left: none;
 		border-right: 0.25em solid currentColor;
 		padding-left: 0;
 		padding-right: 1em;
 	}
 
-	&.has-text-align-center {
+	&:where(.has-text-align-center) {
 		border: none;
 		padding-left: 0;
 	}
 	// .is-style-large and .is-large are kept for backwards compatibility.
-	&.is-style-plain,
+	// They are not wrapped in `:where()` to keep specificity as it was before
+	// they were deprecated.
+	&:where(.is-style-plain),
 	&.is-style-large,
 	&.is-large {
 		border: none;


t-hamano (commented):
Thanks for the PR!

I tested it using Emptytheme. I confirmed that the theme style can be overridden correctly from the global styles, but I found that there are cases where it is not overridden correctly in block instances. That is, when only the border width is applied.

After investigating, I found that if only the border width is applied in the global styles, the `border-style: solid` is also applied at the same time, so it overrides the theme style `border:none`:

quote-border-global-styles

On the other hand, in block instances, it seems that the border is overridden by the theme's `border:none`, probably because only the `border-width` is applied as an inline style:

quote-border

This can be solved by explicitly applying `border-style` to the block instance, but is there a way to solve it without doing this?

t-hamano:
> On the other hand, in block instances, it seems that the border is overridden by the theme's `border:none`, probably because only the `border-width` is applied as an inline style:

To solve this problem, I think it might be effective to add the following CSS to the `theme.scss`:

```scss
.wp-block-quote {
	&:where([style*="border-width"]) {
		border-style: solid;
	}
}
```

aaronrobertshaw:
Thanks for the review @t-hamano 👍 

> On the other hand, in block instances, it seems that the border is overridden by the theme's border:none, probably because only the border-width is applied as an inline style:

The `border-style` isn't automatically set when adjusting block instance styles so that the block can inherit the border style defined in the global styles. Until recently, the post editor didn't have access to the global styles data so it couldn't determine if there is a value to inherit or not.

I'm in two minds on this one. 

I don't think adding more and more fallback CSS is a good option. The `theme.scss` styles are supposed to be opinionated styles as well. In these cases themes opting into them have opted into their removal of borders on certain sides for some alignments. That said, the user still has control of the end result, with an extra click or two for these edge cases.

An alternative path that can be explored as a follow-up is to check the now available global styles data and apply the same fallback border-style logic for the block instances if that's what we ultimately want. This would be a more holistic approach that can be leveraged across all blocks that adopt border supports.

One last thing, the proposed CSS tweak would need to be extended for all longhand `border-*-width` and `border-*-color` inline styles. Quite a lot of CSS for an edge case which is only presented when changing the border of a quote in themes adopting these opinionated styles.

So after having typed all that out. I propose we proceed with this PR as is. 

Depending on what comes of the work around displaying inherited global styles in the block inspector, we can follow up on the fallback for borders then.

What do you think?

t-hamano (approved):
Thank you for the detailed explanation!

If I'm explaining it correctly, then it's fine to ship this PR first, right?

> Depending on what comes of the work around displaying inherited global styles in the block inspector, we can follow up on the fallback for borders then.

I discovered that in #37752, it's being considered to reflect global style values ​​in the block inspector. Border fallbacks might also be considered in this issue.

ramonjd:
Looking at the conversation, and testing the PR, it solves the main issue. I'll merge while @aaronrobertshaw is away.


| Before | After |
|--------|--------|
| Screenshot 2024-08-07 at 10 43 20 AM | Screenshot 2024-08-07 at 10 42 49 AM |

PR #64038, opened 2024-07-29
Style Book: clearly denote heading levels

ramonjd:
## What?

Resolves https://github.com/WordPress/gutenberg/issues/64033

Uses Heading descriptors — `"Heading 1", "Heading 2"` etc, to describe the heading variants in the style book

Adds a Heading 6 example, which corresponds to the maximum heading level available in the typography tools panels.

## Why?

So that the example has text that describes what it is.

## How?

Copy paste, a bit of typing.

## Testing Instructions

1. Fire up this branch and head to the Site Editor
2. Open up the Style Book. Look for the 👁️ icon!
3. Check that the Heading 1, 2, 3... copy is there in the place of "Code is Poetry"

## Screenshots or screencast 

Screenshot 2024-07-29 at 10 55 42 AM

PR #64047, opened 2024-07-29
Style Book: Fix critical error when heading block is not registered

t-hamano:
Discovered this while reviewing #64038

## What?

This PR fixes a critical error in the Style Book screen when the heading block is not registered.

style-book

## Why?

This is because a non-existent block is being created via `createBlock()`.

## How?

Use the `getBlockType()` selector to pre-check if a heading block exists, and if not, return non-heading block examples.

## Testing Instructions

- Run the following code in your browser console: `wp.blocks.unregisterBlockType('core/heading')`
- Open the StyleBook.
- You won’t see the heading block, but the StyleBook should display fine.

### DIFF
--- a/packages/edit-site/src/components/style-book/index.js
+++ b/packages/edit-site/src/components/style-book/index.js
@@ -13,6 +13,7 @@ import {
 import { __, sprintf } from '@wordpress/i18n';
 import {
 	getCategories,
+	getBlockType,
 	getBlockTypes,
 	getBlockFromExample,
 	createBlock,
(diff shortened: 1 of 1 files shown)


packages/edit-site/src/components/style-book/index.js
```diff
@@ -178,7 +145,31 @@ function getExamples() {
...
+		blocks: [ 1, 2, 3, 4, 5, 6 ].map( ( level ) => {
+			return createBlock( 'core/heading', {
+				content: sprintf(
+					// translators: %d: heading level e.g: "1", "2", "3"
+					__( 'Heading %d' ),
+					level
+				),
+				level,
+			} );
+		} ),
```
t-hamano, line 169: This is not directly related to this PR, but it is a code refactoring. The translator comment is the same as that of the `HeadingLevelDropdown` component:

https://github.com/WordPress/gutenberg/blob/27fbbd8e9b57495339a077ef0c8ba67a18392699/packages/block-editor/src/components/block-heading-level-dropdown/index.js#L55-L59

ramonjd (approved):
Good catch!

## Before

https://github.com/user-attachments/assets/dae1c840-3020-4374-a41f-0e10388c14c3


## After


https://github.com/user-attachments/assets/eed38488-1e17-4d3e-b492-a6276e2a9bf5

PR #64048, opened 2024-07-29
Block Insertion: Clear the insertion point when selecting a different block or clearing block selection

### LINKED ISSUE #63866: Zoom Out: Update the insertion point when selecting a block.
In zoom out mode we display an insertion point marker (a thick blue bar) when you click the [+] inserter button. This marker doesn't get reset when you select a different block.

We should update the insertion point and clear the insertion marker whenever you select a different block.

scruffian:
## What?
Adds additional situations in which we clear the block insertion point. Fixes https://github.com/WordPress/gutenberg/issues/63866

## Why?
When a block is selected, it would be expected that the insertion point would also update to reflect the newly selected block, so we should reset the insertion point in this case.

## How?
Add more actions to the insertion point reducer so that it detects when a block is selected.

## Testing Instructions
1. Open the Site Editor
2. Enter zoom out mode (this isn't necessary but it makes it much easier to test) - you can enter zoom out by selecting a pattern category in the inserter
3. Click one on the [+] in the canvas, to set an insertion point. Notice that a wide blue line appears between the two blocks.
4. Select another block, or clear block selection. Notice that that wide blue line disappears.
5. Now select a pattern in the canvas. This pattern should be inserted after the newly selected block, not at the place where the old insertion point was.
6. Create an insertion point by clicking the [+]. Hover a pattern in the inserter. You should still see the wide blue line in the canvas.
7. Now click the pattern. It should be added to the place that the insertion point indicator was.

## Screenshots or screencast 
Trunk

https://github.com/user-attachments/assets/727647e2-eb15-4fa5-a942-84f10dba3d85

This branch

https://github.com/user-attachments/assets/32e8e071-e170-42c0-bdba-d36db42b0f11

### DIFF
--- a/packages/block-editor/src/components/inserter/quick-inserter.js
+++ b/packages/block-editor/src/components/inserter/quick-inserter.js
@@ -9,7 +9,7 @@ import clsx from 'clsx';
 import { useState, useEffect } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Button, SearchControl } from '@wordpress/components';
-import { useSelect } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
 
 /**
  * Internal dependencies
@@ -82,6 +82,8 @@ export default function QuickInserter( {
 		}
 	}, [ setInserterIsOpened ] );
 
+	const { showInsertionPoint } = useDispatch( blockEditorStore );
+
 	// When clicking Browse All select the appropriate block so as
 	// the insertion point can work as expected.
 	const onBrowseAll = () => {
@@ -91,6 +93,7 @@ export default function QuickInserter( {
 			filterValue,
 			onSelect,
 		} );
+		showInsertionPoint( rootClientId, insertionIndex );
 	};
 
 	let maxBlockPatterns = 0;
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
looks like the e2e failures could be legitimate?

scruffian:
I broke https://github.com/WordPress/gutenberg/pull/64392 out of this to better isolate the test failures.

MaggieCabrera (approved):
This is testing good for me and the code makes sense to me, I don't know if the changes in the reducer could have other side effects but I can't think of any right now

scruffian:
Ok, let's bring this in. I'm sure any unintended side effects will be best discovered by people using it, and then we can revert :)

talldan:
> Ok, let's bring this in. I'm sure any unintended side effects will be best discovered by people using it, and then we can revert :)

@scruffian It looks like there are a couple of issues that were caused:
- https://github.com/WordPress/gutenberg/issues/64998 (not what the OP describes, but a couple of issues mentioned in the comments)
- https://github.com/WordPress/gutenberg/issues/65063

jeryj:
Started trying to untangle this issue: https://github.com/WordPress/gutenberg/pull/65098

scruffian:
> and then we can revert :)

I feel like a prophet!

Issue #63124, opened 2024-07-04
Improve the Site Editor preview for PHP templates


carolinan:
## What problem does this address?

If you activate a hybrid theme, or a classic theme that is a child theme of a block theme,
the PHP page template files can not be edited in the Site Editor.
This is a current fact, that may be apparent for developers/designers and advanced users, but not to all.

- If you are on the front of the site, and you are viewing content that uses a PHP template, and you click on the "Edit Site" link in the top toolbar, 
- If you are in the admin area, and you open Appearance > Editor,

Then the Site Editor preview opens but it is empty except for the loading animation.
It looks like it is trying to load the template, but nothing happens.
I am testing this on macOS.

## What is your proposed solution?

It may be better to display a message saying that this template can not be previewed.

## Instructions
Create a new classic child theme and set the parent to twentytwentyfour.
Add a index.php and a home.php template in the theme folder of the child theme.
Activate the child theme.
Open the Site Editor and view the preview.

<img width="1624" alt="The Site Editor with a blank preview" src="https://github.com/WordPress/gutenberg/assets/7422055/44f01410-cc07-4242-a02b-7202ddfa44b8">

carolinan:
The above results are for macOS.
I wanted to look closer at this error today, but I happen to be working on Windows 11 this time, and now I can't reproduce the blank page in the Site Editor:
The home template from the parent theme shows both in the Site Editor and the front. This is a problem because it seems the child theme is ignored.

carolinan:
I have somewhat narrowed it down. The child theme works on the front on Windows with wp-env on WordPress trunk, but not on "Local" by flywheel.

carolinan:
How about using the `<Warning>`?  (With or without the loader)

<img width="1511" alt="A PHP template from a child theme is viewed in the Site Editor" src="https://github.com/user-attachments/assets/b2fdfa53-04c7-4892-862b-7f9ef31b1d0e">

carolinan:
Closing as a duplicate, in favor of https://github.com/WordPress/gutenberg/issues/64341

PR #64050, opened 2024-07-29
Site Editor: Add context for templates that can not be previewed

carolinan:
## What?

Add context to the Site Editor preview when a template can not be loaded.
With the "Site Editor Preview" I mean the canvas where the blocks in a template normally show when you go to 
Appearance > Editor.

An example of a template that can not be previewed is a PHP template added in a hybrid theme (A block theme that also has PHP template files) or a PHP template added in a classic child theme where the parent is a block theme.

Please see the issue for details. https://github.com/WordPress/gutenberg/issues/63124

## Why?

Currently it looks like the Site Editor is trying to load the PHP template in the preview, as it shows the loader animation, but stops progressing.

## How?

Adds a conditional check for:
- Wether the preview is ready
- Wether the post type can be resolved. In the case of a PHP template, the post type is not available.

Then uses the `<Warning>` component to display the context "The template can not be previewed."

## Testing Instructions

Test wether viewing a PHP template in the site editor show the message, or only loads a blank preview with the loader animation. Here is one example of how to test this:

In Twenty Twenty Four, using your preferred file editor, rename home.html to home.php and then move home.php to the root folder of the theme. Congratulations, you have created a hybrid theme.

With reading settings set to "Your homepage displays: latest posts", WordPress will now try to load home.php.
Open the Site Editor by going to Appearance > Editor.

## Screenshots or screencast 
With the PR:
A PHP template is viewed in the Site Editor, and a warning shows. The warning can not be dismissed.


carolinan:
Or perhaps "The template you are currently using is not compatible with the Site Editor."
to match with "The theme you are currently using is not compatible with the Site Editor."
https://core.trac.wordpress.org/browser/tags/6.6/src/wp-admin/site-editor.php#L29

carolinan:
The condition that checks if the message should show or not is not correct, because if there is a home template saved in the database, the message still flash on screen briefly, before the saved (user edited/ user created) template shows.

PR #64061, opened 2024-07-29
Fix: Non working link to an interactivity API example block.

jorgefilipecosta:
The link https://github.com/WordPress/block-development-examples/tree/trunk/plugins/833d15 does not works at all and returns a 404. Checking the repository the correct link should be https://github.com/WordPress/block-development-examples/tree/trunk/plugins/interactivity-api-block-833d15.
This PR updates the link to fix the issue.


juanmaguitar:
Good catch! 
Thanks @jorgefilipecosta 
LGTM! ✅

juanmaguitar (approved):

PR #64063, opened 2024-07-29
Only unlock the necessary setLastFocus and getLastFocus

jeryj:
## What?

`setLastFocus` and `getLastFocus` were unlocked in a group of other non-private functions. This moves these calls to their own unlock call so it doesn't force third-parties to agree to the unlock unnecessarily.

## Testing Instructions

There should be no functional changes.


Mamaduka:
I think it's okay to unlock with a group of non-private selectors/actions. While it's not an issue here, we don't want to accidentally promote creating separate store subscriptions for private states.

PR #64064, opened 2024-07-29
DataViews: sort author by name + allow custom sort function

oandregal:
## What?

This PR sorts author by name instead of id. It does so by also allowing fields to provide a custom sort function.

## Why?

Author ids don't mean anything to users, we should sort it by name.

Note that the REST API allows sorting by author but it really sorts by "author id". The wp-admin doesn't allow sorting the post/page list by author.

## How?

- Formalizes the Field type `integer`, which provides a default sorting function.
- Allows any integer field to provide its own custom function.

## Testing Instructions

- Create three users, in this order:
  - alpha with role admin
  - gamma with role admin
  - beta with role admin
- Create three pages and assign each to a different author.
- Go to "Site editor > Pages" and sort by author.

The expected result is that the order is pages created by alpha, then beta, then gamma. In `trunk`, the order is by id (this is: alpha, gamma, beta).

### DIFF
--- a/packages/dataviews/src/field-types/index.tsx
+++ b/packages/dataviews/src/field-types/index.tsx
@@ -1,0 +1,22 @@
+/**
+ * Internal dependencies
+ */
+import { default as integer } from './integer';
+import type { FieldType } from '../types';
+
+/**
+ *
+ * @param {FieldType} type The field type definition to get.
+ *
+ * @return A field type definition.
+ */
+export default function getFieldTypeDefinition( type?: FieldType ) {
+	if ( 'integer' === type ) {
+		return integer;
+	}
+
+	// TODO: What's the default when the type is undefined?
+	return {
+		sort: () => 0,
+	};
+}
--- a/packages/dataviews/src/field-types/index.tsx
+++ b/packages/dataviews/src/field-types/index.tsx
@@ -15,7 +15,7 @@ export default function getFieldTypeDefinition( type?: FieldType ) {
 		return integer;
 	}
 
-	// TODO: What's the default when the type is undefined?
+	// If no type found, the sort function doesn't do anything.
 	return {
 		sort: () => 0,
 	};
--- a/packages/dataviews/src/field-types/integer.tsx
+++ b/packages/dataviews/src/field-types/integer.tsx
@@ -4,15 +4,17 @@
 import type { SortDirection } from '../types';
 
 function sort< Item >(
-	itemA: Item,
-	itemB: Item,
-	direction: SortDirection,
-	getValue: ( args: { item: Item } ) => any
+	a: {
+		item: Item;
+		value: any;
+	},
+	b: {
+		item: Item;
+		value: any;
+	},
+	direction: SortDirection
 ) {
-	const valueA = getValue( { item: itemA } );
-	const valueB = getValue( { item: itemB } );
-
-	return direction === 'asc' ? valueA - valueB : valueB - valueA;
+	return direction === 'asc' ? a.value - b.value : b.value - a.value;
 }
 
 export default {
--- a/packages/dataviews/src/field-types/integer.tsx
+++ b/packages/dataviews/src/field-types/integer.tsx
@@ -1,0 +1,9 @@
+/**
+ * Internal dependencies
+ */
+import type { SortDirection } from '../types';
+
+export default {
+	sort: ( a: number, b: number, direction: SortDirection ) =>
+		direction === 'asc' ? a - b : b - a,
+};
--- a/packages/dataviews/src/field-types/integer.tsx
+++ b/packages/dataviews/src/field-types/integer.tsx
@@ -7,10 +7,10 @@ function sort< Item >(
 	itemA: Item,
 	itemB: Item,
 	direction: SortDirection,
-	field: any
+	getValue: ( args: { item: Item } ) => any
 ) {
-	const valueA = field.getValue( { item: itemA } );
-	const valueB = field.getValue( { item: itemB } );
+	const valueA = getValue( { item: itemA } );
+	const valueB = getValue( { item: itemB } );
 
 	return direction === 'asc' ? valueA - valueB : valueB - valueA;
 }
--- a/packages/dataviews/src/field-types/integer.tsx
+++ b/packages/dataviews/src/field-types/integer.tsx
@@ -4,6 +4,10 @@
 import type { SortDirection } from '../types';
 
 export default {
-	sort: ( a: number, b: number, direction: SortDirection ) =>
-		direction === 'asc' ? a - b : b - a,
+	sort: ( itemA: any, itemB: any, direction: SortDirection, field: any ) => {
+		const valueA = field.getValue( { item: itemA } );
+		const valueB = field.getValue( { item: itemB } );
+
+		return direction === 'asc' ? valueA - valueB : valueB - valueA;
+	},
 };
--- a/packages/dataviews/src/field-types/integer.tsx
+++ b/packages/dataviews/src/field-types/integer.tsx
@@ -3,18 +3,8 @@
  */
 import type { SortDirection } from '../types';
 
-function sort< Item >(
-	a: {
-		item: Item;
-		value: any;
-	},
-	b: {
-		item: Item;
-		value: any;
-	},
-	direction: SortDirection
-) {
-	return direction === 'asc' ? a.value - b.value : b.value - a.value;
+function sort( a: any, b: any, direction: SortDirection ) {
+	return direction === 'asc' ? a - b : b - a;
 }
 
 export default {
--- a/packages/dataviews/src/field-types/integer.tsx
+++ b/packages/dataviews/src/field-types/integer.tsx
@@ -3,11 +3,18 @@
  */
 import type { SortDirection } from '../types';
 
-export default {
-	sort: ( itemA: any, itemB: any, direction: SortDirection, field: any ) => {
-		const valueA = field.getValue( { item: itemA } );
-		const valueB = field.getValue( { item: itemB } );
+function sort< Item >(
+	itemA: Item,
+	itemB: Item,
+	direction: SortDirection,
+	field: any
+) {
+	const valueA = field.getValue( { item: itemA } );
+	const valueB = field.getValue( { item: itemB } );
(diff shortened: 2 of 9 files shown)


packages/dataviews/src/filter-and-sort-data-view.ts
```diff
@@ -143,6 +143,15 @@ export function filterSortAndPaginate< Item >(
 				const valueA = fieldToSort.getValue( { item: a } ) ?? '';
 				const valueB = fieldToSort.getValue( { item: b } ) ?? '';
 
+				if ( fieldToSort.type === 'integer' ) {
+					// TODO: should we pass the item instead of the resolved value?
+					return fieldToSort.sort(
```
oandregal, line 147: Instead of providing the resolved value (`field.getValue( item )`), this should take the raw item instead. This way, the custom sort would have access to the full data.

I've been having some TypeScript issues when implementing this idea, so thought I'd share as early as possible and look at this later.

packages/edit-site/src/components/post-fields/index.js
```diff
@@ -242,6 +242,16 @@ function usePostFields( viewType ) {
 						label: name,
 					} ) ) || [],
 				render: PostAuthorField,
+				sort: ( a, b, direction ) => {
```
oandregal, line 245: If `sort` receives the item instead of the resolved value, we can get the author name from the data (`item._embedded?.author[ 0 ]?.name`).

It doesn't change anything in this particular case, though it means the sort function won't require any other dependency if the desired data is part of the dataset. For example, in this case, `authors` would have to be added to the dependencies if it wasn't already.

packages/dataviews/src/filter-and-sort-data-view.ts
```diff
@@ -143,6 +143,15 @@ export function filterSortAndPaginate< Item >(
...
+						valueA,
+						valueB,
+						view.sort?.direction ?? 'desc'
+					);
+				}
+
 				if (
 					typeof valueA === 'number' &&
```
oandregal, line 156: Note that I've left the existing logic here because `type` is optional. When/If make it required this could be removed.

packages/dataviews/src/field-types/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+ * @param {FieldType} type The field type definition to get.
+ *
+ * @return A field type definition.
+ */
+export default function getFieldTypeDefinition( type?: FieldType ) {
+	if ( 'integer' === type ) {
+		return integer;
+	}
+
+	// If no type found, the sort function doesn't do anything.
```
oandregal, line 18: What do we think of this? What do we do if the type is not found?

packages/dataviews/src/types.ts
```diff
@@ -80,6 +80,16 @@ export type Field< Item > = {
...
 	render?: ComponentType< { item: Item } >;
 
+	/**
+	 * Callback used to sort the field.
+	 */
+	sort?: (
+		a: Item,
+		b: Item,
+		direction: SortDirection,
+		field?: any
```
oandregal: This is where I'm having issues: as soon as I add a type (Field/NormalizedField) here, the TypeScript parser complains.

packages/dataviews/src/filter-and-sort-data-view.ts
youknowriad: This is making `sort` mandatory for `integer` fields right? I think if `field.sort` is not provided we should fallback to the "number" logic that already exists a few lines down here.

youknowriad: Oh I see you moved the default sorting to the field type, (that's better)

packages/dataviews/src/types.ts
youknowriad: There's something weird here, to define the "sort" function, you need the "field" object but the field object contains the "sort" function. Why did you need "field" as an argument here?

packages/dataviews/src/field-types/integer.tsx
```diff
@@ -0,0 +1,20 @@
+/**
```
youknowriad, line 1: Nice file :) 

Should we move the "control" to the field types definition?

packages/dataviews/src/field-types/integer.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * Internal dependencies
+ */
+import type { SortDirection } from '../types';
+
+function sort< Item >(
+	itemA: Item,
+	itemB: Item,
+	direction: SortDirection,
+	field: any
```
youknowriad, line 10: Instead of "field", I think we may want to just pass `getValue` here to avoid the recursion issues.

packages/dataviews/src/types.ts
oandregal: It is for the sort function defined by the type (see [integer.tsx](https://github.com/WordPress/gutenberg/pull/64064/files#diff-ea641593eb6d360ee4bcf491ea3678b942e42b9354e0fe7e81bae3d08902e9a8R12)) to be able to convert the item into the value:

```js
const valueA = field.getValue( { item: itemA } );
```

youknowriad: I think the sort function defined by the type and the one defined by the field can be slightly different. For the field there's no need to pass the `field` but yeah it should be fine to keep the `field` for the field type one, or just pass `getValue` directly.

oandregal: Passing `getValue` makes sense https://github.com/WordPress/gutenberg/pull/64064/commits/bcf814aaa9ccf000db6d025e3175cfde27990ae5 Thanks!

packages/dataviews/src/field-types/integer.tsx
oandregal: Yes! And validation, etc. I'll do that in a follow-up.

packages/dataviews/src/types.ts
```diff
@@ -124,6 +134,12 @@ export type NormalizedField< Item > = Field< Item > & {
 	label: string;
 	getValue: ( args: { item: Item } ) => any;
 	render: ComponentType< { item: Item } >;
+	sort: (
+		a: Item,
+		b: Item,
+		direction: SortDirection,
+		getValue: ( args: { item: Item } ) => any
```
youknowriad: I think we should remove `getValue` from here. It's redundant with the field itself, we're defining the field here. Instead we should update the "normalize" function to inject it when we fallback to the field type sort function.

packages/dataviews/src/field-types/integer.tsx
```diff
@@ -0,0 +1,20 @@
+/**
+ * Internal dependencies
+ */
+import type { SortDirection } from '../types';
+
+function sort< Item >(
+	itemA: Item,
+	itemB: Item,
+	direction: SortDirection,
+	getValue: ( args: { item: Item } ) => any
```
youknowriad: I've thinking a bit more here, I think we shouldn't pass `getValue` here at all, we should just pass `valueA` and `valueB` that are both integers (instead of itemA and itemB). At this point (field type), there's no need to know about the item or the field or anything.

oandregal: I've pushed https://github.com/WordPress/gutenberg/pull/64064/commits/eaa0b63c8da763113d64aca0ec83342b14faa293

I like this more than previous approaches and it could scale well. For example, in validation, the field types such as integer would need to have access to `elements` as well, so we'd add that.

It's not entirely perfect, though: a generic field type such as `integer` doesn't know the "id" of the data it's working with so the `item` parameter there is useless.

oandregal: Passing only the values is how I started. The trade-off with that approach is that it presumes the custom sort function wants to access data that's defined elsewhere, but it may want to access data that comes as part of the item (see author custom function).

packages/dataviews/src/field-types/integer.tsx
```diff
@@ -4,15 +4,17 @@
 import type { SortDirection } from '../types';
 
 function sort< Item >(
-	itemA: Item,
-	itemB: Item,
-	direction: SortDirection,
-	getValue: ( args: { item: Item } ) => any
+	a: {
+		item: Item;
```
youknowriad: Why have Item here at all? I was thinking `a: any, b: any` should be sufficient no?

youknowriad: I see what you did, I think you're misunderstanding what I'm suggesting. I'm suggesting that `fieldType.sort` and `field.sort` to have different APIs. one receives the items, the other receives the values. Do you want me to try to commit what I have in mind?

packages/dataviews/src/field-types/integer.tsx
youknowriad: > Passing only the values is how I started. The trade-off with that approach is that it presumes the custom sort function wants to access data that's defined elsewhere, but it may want to access data that comes as part of the item (see author custom function).

This is only true for `field.sort` not for `fieldType.sort` where you always want to use the data.

packages/dataviews/src/field-types/integer.tsx
oandregal: @youknowriad please, be welcome to push. In my mind both functions are instances of the same idea, so they'd have the same signature. But I'm open to try different things: push and we can discuss what's there.

youknowriad: For me field types and fields work at different levels, field types work on values, fields work on items, so it doesn't make sense to have the same API.

I've pushed what I had in mind in [621ac69](https://github.com/WordPress/gutenberg/pull/64064/commits/621ac69ad314178c8d6cabfc723a25b3b05f78c0) 

Feel free to revert if you don't like it.

oandregal: [621ac69](https://github.com/WordPress/gutenberg/pull/64064/commits/621ac69ad314178c8d6cabfc723a25b3b05f78c0) works for me.

youknowriad (approved):

packages/dataviews/src/field-types/integer.tsx
oandregal: Validation moved to the type definition at https://github.com/WordPress/gutenberg/pull/64164

PR #60234, opened 2024-03-27
Refine rename flow for blocks with overrides

kevin940726:
## What?

Part of https://github.com/WordPress/gutenberg/issues/59813. Supersedes https://github.com/WordPress/gutenberg/pull/60066. Also close https://github.com/WordPress/gutenberg/issues/60118.

**The latest design is https://github.com/WordPress/gutenberg/pull/60234#issuecomment-2034513896**

This PR does a few things:
1. If no name exists on block, and overrides is disabled, use "allow overrides" button
2. If name exists and overrides is disabled, use toggle
3. If overrides is enabled, use toggle
4. When clicking button, show "allow overrides" modal that is essentially a rebranded "rename block" modal
5. In all other rename actions, use the standard rename modal with overrides warning text
6. The default state for overrides is disabled, users have to opt-in to enable pattern overrides

## Why?

This is to simplify the UI flow for renaming and toggling overrides for blocks.

## How?

Add overrides warning to the rename modal and change the overrides control to a button when there's no name.

## Testing Instructions

1. Create a synced pattern and add a few paragraphs.
2. If the block has no name, the control should show a button for overrides.
3. Clicking on the button should open a renaming modal.
4. Once the block is named, the control becomes a toggle button.
5. Other renaming flow (list view, block toolbar, shortcut) should show overrides warning.

## Screenshots or screencast 

https://github.com/WordPress/gutenberg/assets/7753001/37ed9f09-0b7b-447d-835f-00dcabc98077

PR #64065, opened 2024-07-29
FormToggle, ToggleControl: Fix docgen in Storybook

mirka:
## What?

Fixes a documentation regression introduced in #60234 where for `FormToggle`/`ToggleControl`, the JSDocs stopped working (in IntelliSense) and the Storybook docgen could not extract the JSDocs and TypeScript types.

## Why?

For the docgen in Storybook to [work](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/CONTRIBUTING.md#considerations-for-the-docgen), a ref-forwarded component must have a named export. A contributor passing by would not know this, so if these kinds of regressions happen again we should consider adding an eslint rule. (This is the first regression of this kind I've seen, and I don't expect it to happen a lot.)

## Testing Instructions

The Storybook docs for both components should show the main JSDocs, as well as the props table with descriptions (broken docs in trunk for comparison: [FormToggle](https://wordpress.github.io/gutenberg/?path=/docs/components-formtoggle--docs)/[ToggleControl](https://wordpress.github.io/gutenberg/?path=/docs/components-togglecontrol--docs)).

The IntelliSense in your IDE should show the JSDocs for both components.

### DIFF
--- a/packages/components/src/form-toggle/index.tsx
+++ b/packages/components/src/form-toggle/index.tsx
@@ -17,26 +17,7 @@ import type { WordPressComponentProps } from '../context';
 
 export const noop = () => {};
 
-/**
- * FormToggle switches a single setting on or off.
- *
- * ```jsx
- * import { FormToggle } from '@wordpress/components';
- * import { useState } from '@wordpress/element';
- *
- * const MyFormToggle = () => {
- *   const [ isChecked, setChecked ] = useState( true );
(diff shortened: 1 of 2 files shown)


tyxla (approved):
LGTM, thanks 👍 

Definitely looks like something that the dev environment or CI should have caught 😞

Issue #59071, opened 2024-02-15
Components: Revert flaky unit test `sleep()` hacks


tyxla:
## What problem does this address?
In #59012, #58629, and #58968 we added some `sleep()` calls before `press.Tab()` as a patching attempt to make some unit tests less flaky.

## What is your proposed solution?
We should fix the underlying cause (likely in `@ariakit/test`) instead of resolving it with hacks like this. We should also revert the `sleep()` calls we introduced In #59012, #58629, and #58968.

cc @mirka @Mamaduka

tyxla:
@diegohaz any ideas from the top of your head why tests could be flaky on these rare occasions, and a `sleep()` before `press.Tab()` makes them more stable? Is this something that we could have addressed in `@ariakit/test` instead?

diegohaz:
All items within an Ariakit composite widget remain tabbable until they are completely processed internally (for example, to determine their order). After that, only the active item stays tabbable. This happens asynchronously after the initial render.

This is to guarantee that all elements are accessible to keyboard users until we can respond to arrow keys.

A potential solution could be to wait for inactive items to become non-tabbable before hitting the Tab key during the test.

tyxla:
Cleaning those up in #64205.

PR #64066, opened 2024-07-29
Upgrade Ariakit

DaniGuardiola:
Continuation of #62947 (dependencies got messed up)

# To do

Besides upgrading to the latest version and fixing all remaining failed tests, there are some other items to complete.

- [x] Smoke test.
- [x] Ensure that the animation change doesn't affect other components other than custom-select-control, custom-select-control-v2 and dropdown-menu-v2.
- [x] Add relevant changelog entries.
- [x] Tooltip `aria-describedBy` changes: add temporary compatibility fix and revert test changes.
- [x] Contribute `render` changes to upstream `@ariakit/test` (returning `rerender` from the output of the wrapped `ReactTestingLibrary.render`). See https://github.com/ariakit/ariakit/pull/3981
- [x] Remove local temporary fork of `@ariakit/test`'s `render` function.
- [x] Discuss idea: enforce `@ariakit/test`'s `render` and user events in all components package tests? (e.g. `noRestrictedImports`)
  - **Resolution:** enforce in a follow-up PR.
- [x] Discuss: "aria-describedBy" in tooltip anchors ([context](https://github.com/WordPress/gutenberg/pull/62947/files/4c022ed8d80fdce62795e3c0895fd3dfcbc08421..114badda5f68770118090baaf032c1a19b4cf84a#diff-eb9fcb45579ca92e96f172d835623c0682387b4ec0ff0e7ed5009b958b7c2c5c))
  - **Resolution:** add a temporary compatibility fix for now, then in the future do a pass to check if any labels need to be added, and document in the changelog and potentially a dev note.

# History by version (unit tests)

## v0.3.x (from v0.3.13)

At v0.3.13, tests fail:

image

The reason is tests need to give some components a bit more time to render since Ariakit uses some timers under the hood. Fixed (e52022c6bec8f9a5eba801de23c4b416bebdebcd)  by replacing RTL's `render` with `@ariakit/test`'s `render` - including the [local temporary fork with `rerender` support](https://github.com/ariakit/ariakit/issues/3939#issuecomment-2251499387) (ffe79281867cacbb99f9eef546b0dfc25e8f1b70).

After addressing, all tests pass in all v0.3.x versions.

- v0.3.13 (2d9abe36f10134ec6067c2ccd02ac83a4d4bf4bb, failing tests, fixed in e52022c6bec8f9a5eba801de23c4b416bebdebcd)
- v0.3.14 (83fd1488c84fa862a93643f837f85749c5263fb0)

image

## 0.4.x

image

[This diff from the previous PR](https://github.com/WordPress/gutenberg/pull/62947/files/4c022ed8d80fdce62795e3c0895fd3dfcbc08421..114badda5f68770118090baaf032c1a19b4cf84a) has information about all root causes for test failures.

Essentially, we're impacted by three Ariakit changes in v0.4.0:

- [Public data attributes have now boolean values](https://ariakit.org/changelog#public-data-attributes-have-now-boolean-values) (breaking change, fixed in ce0b7f3cf9a1713a5c8f8cd38f1367e7a18c8104 & ff470ad07eabdc6f8298a12198561807673f8251)
- [Improved animation support](https://ariakit.org/changelog#improved-animation-support) (breaking change, fixed in e39dc495b6af1c1dc027eeedb6ec0190070349fb)
- [Tooltips no longer use aria-describedby to associate the tooltip content with the anchor](https://ariakit.org/changelog#other-updates-6) (non-breaking change, fixed in 5f705ebe82171a9808b808777dc08564a67161d7)

After addressing, all tests pass between v0.4.0 and v0.4.5.

- v0.4.0 (0020c344e4eb009664a6738d9d6e09fca572cee1, failing tests, fixed in ce0b7f3cf9a1713a5c8f8cd38f1367e7a18c8104, e39dc495b6af1c1dc027eeedb6ec0190070349fb, 5f705ebe82171a9808b808777dc08564a67161d7 & ff470ad07eabdc6f8298a12198561807673f8251)
- v0.4.1 (46c8142f9d0842c175db285c1dd151be9c157ea7)
- v0.4.2 (753f631b8a9a1706af9e0c2f4b8d427a3419ed28)
- v0.4.3 (4eb1c4f02130b6dcf54e09e43dd4ccc1fe81810f)
- v0.4.4 (cc0ae3cd71b8cd45d98930ce0d7bccdb707a3cbe)
- v0.4.5 (1a72000f7477f5458980ee6bcf9cc86919a2c8dd)

## 0.4.6 (287c22f622dcd0a0015d429f5b5e3da599b89cb5, failing tests, fixed in 27218de9349881985f010232c428094c2b24f4e6)

In this version, there are 3-5 new test failures (2 seem to be flaky):

image

Similarly to the `render` wait issues, this is due to not properly waiting for some internal timers to complete. This can be remediated by replacing RTL's event triggers with ones from `@ariakit/test`.

After addressing (27218de9349881985f010232c428094c2b24f4e6), all tests pass in this version.

## 0.4.7 - latest (94deccd6e1f2f1d9611bd2d0babb2d12601ea371)

All tests pass and we're on latest now! 🎉 

# Follow-ups

These are tasks to be done after this PR lands. I'm adding stuff that seems to have some consensus here, and unless I hear different I plan to go through with them. Please speak up if you have any arguments against these decisions! For more context on some of these, look for "resolution" notes in the "To do" section.

- [ ] Enforce `@ariakit/test`'s `render` and event trigger utilities in all components package tests (see https://github.com/WordPress/gutenberg/pull/64066#issuecomment-2258164492 and https://github.com/WordPress/gutenberg/pull/64066#issuecomment-2258164492) - #64180
- [ ] Tooltip `aria-describedBy` changes:
  - [ ] Do a pass to check if any labels need to be added.
  - [ ] Add changelog note.
  - [ ] Dev note?
- [ ] Document testing procedure and context around `render` in appropriate place.
  - [ ] For `@wordpress/components`.
  - [ ] For other affected packages.
- [x] Remove hacky `sleep` calls - #64205
- [ ] Discuss all-in switch to `@ariakit/test` from RTL ([context](https://github.com/ariakit/ariakit/issues/3939#issuecomment-2258293118)).


DaniGuardiola:
This was kind of a PITA to pull off, since it required reverting the lockfile, doing a `distclean`, installing deps, checking the lockfile to manually verify that the right version was installed (thanks npm), building, and then running all tests. For every, single, version.

I was extremely careful in double checking that the fixes take the expected effect and everything looks correct. This allowed me to figure out when the mistery new test failures come, and that's at version v0.4.6. Will be working on those next, until we're at latest (v0.4.7).

Please @WordPress/gutenberg-components read through the new, much shorter description since we need to discuss some last details and I would also appreciate some help with testing and generally making sure I didn't miss anything important.

ciampo:
Thank you, Dani!

As mentioned in #62947, we should probably look at cleaning up the custom `render` utility that was added in `Composite` (and in general, any other workarounds that could be removed with the updated `render` function)

> Discuss idea: enforce `@ariakit/test`'s `render` in all components package tests? (e.g. `noRestrictedImports`)

That could be a way to improve the devX in the components package (I'm always in favor of consistency).

But we should also consider consumers of the components package (both other Gutenberg packages, and third-party consumers), who may end up struggling with the same failures. How can we help them? Considering that `ariakit` is considered an implementation detail, should we still point them at `@ariakit/test` ? Should we instead offer a `render` function as part of some Gutenberg testing utils (which could also help to bridge any API gap between `@ariakit/test` and RTL ?)

> Discuss: "aria-describedBy" in tooltip anchors ([context](https://github.com/WordPress/gutenberg/pull/62947/files/4c022ed8d80fdce62795e3c0895fd3dfcbc08421..114badda5f68770118090baaf032c1a19b4cf84a#diff-eb9fcb45579ca92e96f172d835623c0682387b4ec0ff0e7ed5009b958b7c2c5c))

For more context, the main change here is that "users are responsible for making anchors accessible" ([source](https://ariakit.org/components/tooltip#tooltip-anchors-must-have-accessible-names)).

Judging by previous conversations ([example](https://github.com/ariakit/ariakit/issues/3242)), we should be ok with this change. Although I would still have a pass and make sure that the removal of the tooltip description won't cause anchors to suddenly become inaccessible (and we should fix that by adding the information back in other ways, like with visually hidden text in the anchor).

DaniGuardiola:
> we should probably look at cleaning up the custom render utility

Yep, ideally by contributing to Ariakit which is currently being discussed (https://github.com/ariakit/ariakit/issues/3939#issuecomment-2256768522). It's on the TODO list!

> That could be a way to improve the devX in the components package

+1

> But we should also consider consumers of the components package (both other Gutenberg packages, and third-party consumers), who may end up struggling with the same failures. How can we help them? Considering that ariakit is considered an implementation detail, should we still point them at @ariakit/test ?

I would suggest being consistent: whatever we do in the components package is what we recommend on the outside. For now, that'd be switching to `@ariakit/test`, at least for the render and event trigger functions. Diego mentioned in the linked thread that, long-term, this package might stop using RTL altogether, which might result in things like `screen.get/find/query` not working anymore. At that point, we could recommend switching to Ariakit's query replacement (`q`), which is a simpler but pretty powerful alternative that we could potentially codemod into.

Note that, while it is indeed under the `@ariakit` scope, `@ariakit/test` is something that works on its own, and probably how RTL should've worked all along. It's actually exciting to think about it as a replacement for RTL altogether, at least in the long term. There are improvements beyond the waits, like the new (awesome imo) query API. All this to say that I don't think `@ariakit/react` being an implementation detail should influence our decisions in regards to `@ariakit/test`.

> Should we instead offer a render function as part of some Gutenberg testing utils (which could also help to bridge any API gap between @ariakit/test and RTL ?)

This could be a good strategy in that eventuality, yes!

However, that seems like a future issue that's not even completely guaranteed to exist, so I think we're fine for now with:

- consistently using `render` and event trigger functions for `@ariakit/test` in `@wordpress/components`
- recommending the same only when necessary for tests in other packages that use Ariakit (independently e.g. dataviews or through the components library) under the hood

What do you think?

> Judging by previous conversations (https://github.com/ariakit/ariakit/issues/3242), we should be ok with this change. Although I would still have a pass and make sure that the removal of the tooltip description won't cause anchors to suddenly become inaccessible (and we should fix that by adding the information back in other ways, like with visually hidden text in the anchor).

Great to know! Agree, we should do a pass.

packages/components/src/button/test/index.tsx
```diff
@@ -138,7 +134,7 @@ describe( 'Button', () => {
 
 			expect( tooltip ).toBeVisible();
 
-			await user.tab();
+			await press.Tab();
```
tyxla, line 137: Noting that these have historically been flaky and we've prepended a `await sleep();` before to resolve that flakiness. Not perfect, of course, and ideally, it would be nice to find a cleaner way that will allow us to clean the existing instances up (see https://github.com/WordPress/gutenberg/issues/59071)

packages/components/src/custom-select-control-v2/styles.ts
```diff
@@ -132,12 +132,13 @@ export const SelectPopover = styled( Ariakit.SelectPopover )`
 	min-width: min-content;
 
 	/* Animation */
-	animation-duration: ${ ANIMATION_PARAMS.DURATION };
-	animation-timing-function: ${ ANIMATION_PARAMS.EASING };
-	animation-name: ${ slideDownAndFade };
-	will-change: transform, opacity;
-	@media ( prefers-reduced-motion ) {
-		animation-duration: 0s;
+	&[data-open] {
```
tyxla, line 135: Is this a piece we can address in a separate PR? 

Looks like `data-open` can be used from `v0.4.0` (see [changelog](https://ariakit.org/changelog#improved-animation-support)).

packages/components/src/dropdown-menu-v2/styles.ts
```diff
@@ -101,23 +101,23 @@ export const DropdownMenu = styled( Ariakit.Menu )<
...
-	}
-	&[data-side='up'] {
-		animation-name: ${ slideUpAndFade };
-	}
-	&[data-side='right'] {
-		animation-name: ${ slideRightAndFade };
-	}
-	@media ( prefers-reduced-motion ) {
-		animation-duration: 0s;
+	&[data-open] {
```
tyxla, line 104: Same question as above.

packages/components/src/tooltip/test/index.tsx
```diff
@@ -65,36 +65,6 @@ describe( 'Tooltip', () => {
 			expectTooltipToBeHidden();
 		} );
 
-		it( 'should associate the tooltip text with its anchor via the accessible description when visible', async () => {
```
tyxla: Can this be done in a separate PR that provides the rationale behind the test removal?

packages/components/src/button/test/index.tsx
DaniGuardiola: good note, something to look into - might be something worth solving at the `@ariakit/test` level

packages/components/src/custom-select-control-v2/styles.ts
DaniGuardiola: No, this is necessary. Tests fail without this change, and animations are visibly broken without it. This is purely an implementation detail, the behavior remains the same as long as this change is made.

packages/components/src/dropdown-menu-v2/styles.ts
DaniGuardiola: Same answer :) https://github.com/WordPress/gutenberg/pull/64066#discussion_r1696768233

packages/components/src/tooltip/test/index.tsx
DaniGuardiola: Not without this test failing.

tyxla: If we don't know why it's failing but don't want to deal with it immediately, I'd recommend skipping the test( `it.skip()`) and adding a TODO comment to research in another PR.

packages/components/src/button/test/index.tsx
DaniGuardiola: just remembered the context behind this - Diego did mention this should be solved now so probably worth removing those sleeps to see if things work properly now 👍

packages/components/src/tooltip/test/index.tsx
DaniGuardiola: We know why it's failing. Please read the description as it details (or links to the details of) every fix - in this case it's because tooltips don't add the describedBy attribute anymore.

tyxla: Got it, thanks! It seems like it might be a good idea to consider including some of the Ariakit changelog details in our own changelog for clarity.

DaniGuardiola: Yeah, I think in this case we'll need to basically copy and paste the tooltip advisory, e.g. "tooltips don't add describedBy anymore etc etc" - that's the only notable change that will be relevant to consumers afaik, everything else (e.g. animation changes) are implementation details

DaniGuardiola:
I added the `restrictedImport` rules (7a43523df0cda88b27b3e1e25fda2c911f1ce6b9):

image

This is the output: https://pastebin.com/raw/RH2WGREJ (144 errors in the components package)

I reverted this change (412f612172db00fc5e76048d4da244eaca29881f) to keep this PR focused on the upgrade itself, this should be done in a follow-up, but wanted to see how this would look.

cc @ciampo

packages/components/src/tooltip/test/index.tsx
tyxla: Maybe we can also invert the test, to basically prove that if the consumer provides `describedBy` on their own, it works well?

DaniGuardiola: Hmm, can you give an example of the test you propose in pseudo-code? I fail to see the value of it at the moment, but maybe I'm missing some details and nuance.

tyxla: Maybe what I suggested doesn't make sense. Just to confirm: `describedBy` is no longer being passed to tooltips, and there's no way for consumers to have it if they want (except manually creating the description element and providing an `aria-describedby` prop manually of course)? If that's true, then there's no need to add a test for that and you can ignore my previous comment.

mirka: We discussed this on a call and agreed to back compat this `aria-describedby` behavior for this PR, and deal with its removal as a separate task 👍

packages/components/src/utils/tmp-ariakit-test-render-replacement.ts
```diff
@@ -0,0 +1,173 @@
+// See https://github.com/WordPress/gutenberg/pull/62947
+// and https://github.com/WordPress/gutenberg/pull/64066
```
mirka: I was going to suggest putting a bit more descriptive comment for this file, but is seems like we can remove this fork very soon! 👏  https://github.com/ariakit/ariakit/pull/3978

tyxla (commented):
Doing some testing in the editor and in Storybook today (and will do some tomorrow).

Initial observation: changing the color in a color picker seems to have become clunkier:

Trunk:

https://github.com/user-attachments/assets/de69595c-8a12-4b71-842f-b90cdacd0593

This branch:

https://github.com/user-attachments/assets/1602d35b-2477-494f-9d99-fd59f21ab7bf

packages/components/src/utils/tmp-ariakit-test-render-replacement.ts
DaniGuardiola: it's gone now!

packages/components/src/tooltip/test/index.tsx
DaniGuardiola: done!

packages/components/src/button/test/index.tsx
DaniGuardiola: will do in a follow-up, but fyi removing all `await sleep()` calls works!

image

tyxla: Sounds good! Assigning you #59071 so you can close that, too, with your follow-up.

tyxla (commented):
Did some more smoke testing and couldn't find any issues - nice work!

Are you planning any other changes before marking the PR as open and ready for review?

DaniGuardiola:
@tyxla just need to go through the remaining todos. If nothing else comes up, changelog will be the only change.

DaniGuardiola:
Checked components for any potential remaining animation issues but couldn't find any. Also did some general testing in the Storybook and Gutenberg and nothing looks broken. Undrafting, get your final reviews in! :) @WordPress/gutenberg-components

tyxla:
I'm unable to reproduce [this](https://github.com/WordPress/gutenberg/pull/64066#pullrequestreview-2210604699) anymore 🎉  (not sure what was going on)

tyxla (approved):
Thanks for the hard work on it @DaniGuardiola 🙌 

I gave this a thorough amount of testing in Storybook, the site editor, and the post editor and didn't spot any regressions. ✅ 

The code looks good! 🚀  Thanks for pointing out all the changes and follow-ups in such detail!

The `navigate` metric seems to have doubled, but it's possible that it's a false alarm:

Screenshot 2024-08-01 at 18 59 09

Let's keep an eye on [performance metrics](https://www.codevitals.run/) after merging this one, just in case. ⏲️ 

Also, let's monitor unit test builds for flakiness ❄️ in the next couple of days.

tyxla:
Filed #64205 to clean up the arbitrary `sleep` calls.

